PR #1088, opened 2025-04-22
Cell recipes

anotherjesse:


seefeldb:
Nice!

Hmm, there's an argument to be made to just treat the recipe as data of the charm, i.e. replace `[TYPE]` and the hash with just a cell link to the data. Then charm meta and recipe meta become the same (and both can still refer to other charms). And "imported from spellbook" is meta data as well.

Mostly saying in case it makes things even easier - just get rid of all the recipe id stuff, and recipes are just source code. But might also be too much of a leap at once?

anotherjesse:
@seefeldb I definitely have thought about doing that all in one go. but I ultimately decided against it as it is changing too much

as is, I was over aggressive about "oh, we don't need to register modules, ..." and need to re-add that!

so, I think a second pass where we link them to the charms as recipe nodes...

charm/src/manager.ts
```diff
@@ -1248,6 +1259,8 @@ export class CharmManager {
       );
     }
 
+    this.syncRecipe(charm);
```
jsantell, line 1262: Should this be awaited?

anotherjesse:
Yes!

PR #1092, opened 2025-04-23
chore: move bootstrap code into own script

Gozala:
Refactors sandbox bootstrap code into a separate js file so it's easier to iterate


toolshed/routes/frontend/frontend.index.ts
```diff
@@ -3,9 +3,18 @@ import * as path from "@std/path";
 import { createRouter } from "@/lib/create-app.ts";
 import { applyProxy } from "./frontend.proxy.ts";
 import { applyStatic } from "./frontend.static.ts";
+import { cors } from "@hono/hono/cors";
 
 const router = createRouter();
 
+router.use(
```
jsantell, line 10: This is specifically for our null-origin iframes requesting `//module/charm/sandbox/bootstrap.js` as an ES module (maybe worth commenting)

jumble/vite.config.ts
```diff
@@ -81,6 +81,7 @@ export default defineConfig({
     },
     headers: {
       "Service-Worker-Allowed": "/data/",
+      "access-control-allow-origin": "*",
```
jsantell, line 84: Is this still needed for jumble dev to load the module? If so let's have a comment for why this is here

deno.jsonc
```diff
@@ -104,6 +104,7 @@
     "merkle-reference": "npm:merkle-reference@^2.0.1",
     "multiformats": "npm:multiformats@^13.3.2",
     "react": "npm:react@^18.3.1",
+    "react-dom": "npm:react-dom@^18.3.1",
```
jsantell, line 107: This isn't ideal, will mean this may be included in the compiled artifact, but I understand why its here

deno.lock
```diff
@@ -12,11 +12,11 @@
     "jsr:@std/assert@1": "1.0.11",
```
jsantell, line 12: A lot of version bumps here, just noting

jsantell (approved):

toolshed/routes/frontend/frontend.index.ts
```diff
@@ -3,9 +3,18 @@ import * as path from "@std/path";
...
+  "*",
+  cors({
```
Gozala, line 12:
```suggestion
  // setup CORS otherwise `bootstrap.js` module import from the srcdoc iframe
  // is blocked and charm fails to bootstrap
  cors({
```

Gozala: @jsantell how's that ☝️ ?

deno.jsonc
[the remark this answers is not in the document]
Gozala: It does but then we have two places where things are configured and can get out of sync, one will be here and other one will be at https://github.com/commontoolsinc/labs/blob/main/charm/src/iframe/static.ts#L7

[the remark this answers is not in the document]
Gozala: I actually think we probably should bundle this modules instead of loading them from `esm.sh` but perhaps that is a task for another day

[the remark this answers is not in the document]
jsantell: They *should* resolve to the same module cache but who knows

PR #1096, opened 2025-04-23
fix: use preloaded data in useDoc

Gozala:
Per https://linear.app/common-tools/issue/CT-308/usedoc-should-return-data-read-from

### DIFF
--- a/jumble/public/module/charm/sandbox/bootstrap.js
+++ b/jumble/public/module/charm/sandbox/bootstrap.js
@@ -55,7 +55,7 @@ window.useDoc = function (key) {
   }
 
   // If we have not yet received response from the host we use field from the
-  // sourceData which was preloaded via `subscribeToSource` during initalization
+  // sourceData which was preloaded via `subscribeToSource` during initialization
   // in the `initializeApp`.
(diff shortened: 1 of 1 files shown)


bfollington (approved):
I originally wanted to achieve this kind of behaviour by blocking the initial React render until we had received the first copy of the Charm's data, then `useDoc` wouldn't need to have this complex fallback behaviour.

seefeldb:
Were you able to verify what happens when the schema defines a key, where the value undefined is allowed (e.g. an optional key with no default value) and it is undefined? Will sourceData include it?

seefeldb:
Ah, since we're not throwing, this is fine.

One thing to watch for: The way to set default values is via the schema, but if the LLM doesn't do that, then it might instead use useEffect, which might exacerbate the issues.

Gozala:
> Ah, since we're not throwing, this is fine.
> 
> One thing to watch for: The way to set default values is via the schema, but if the LLM doesn't do that, then it might instead use useEffect, which might exacerbate the issues.

Perhaps alternative tack would be to kickoff schema alteration when `useDoc` is called with unknown key ? Which may be aligned with more broader idea of lazy bindings.

PR #1098, opened 2025-04-23
Introduce `GenerationOptions` and streamline LLM metadata

bfollington:
An attempt to make threading metadata through the stack a little cleaner, along with apply defaults and using that to pass the `space` in the metadata.


jsantell (approved):

bfollington:
Ty for the review @jsantell!

PR #1102, opened 2025-04-24
chore: Update Writable->Mutable and apply it recursively for queries.

jsantell:
* Gmail recipe was using `as JSONSchema` rather than `satisfies JSONSchema`, which was causing some type issues
* No strong feelings for naming between `Writable`, `Mutable`, mostly noting distinction
* Pulled in extra `satisfies JSONSchema` clauses from @ubik2  #1103 (without which `Schema` resolves to `any`)
* The `Writable` type helper was not applying mutability recursively; an object with a string type would break out its properties, starting with:

```ts
// via `Schema<typeof EmailSchema>`
type Email = {
    readonly id: string;
} & {} & Record<...>
```

but typed as

```ts
// via `Writable<Schema<typeof EmailSchema>>`
type EmailWritable = {
    id: Writable<{
        toString: () => string;
        charAt: (pos: number) => string;
        charCodeAt: (index: number) => number;
        concat: (...strings: string[]) => string;
        indexOf: (searchString: string, position?: number) => number;
        ... 46 more ...;
        [Symbol.iterator]: () => StringIterator<string>;
    }>;
}
```

With using a recursive version of the `Mutable` helper, the types now resolve to a query instance:

```ts
// via `Mutable<Schema<typeof EmailSchema>>`
type Email  = {
    id: string;
}
```

While the email case is *mostly* readonly, and this is used as just an illustrative example (bgAdmin for example does need mutability), we do set `ID` on a `email` query, a property of `JSONSchema`, not `Schema`; this is failing types currently, not sure of the way forward here:

```ts
email[ID] = email.id
```

### DIFF
--- a/builder/src/index.ts
+++ b/builder/src/index.ts
@@ -44,11 +44,12 @@ export {
   isStatic,
   isStreamAlias,
   type JSONSchema,
-  type JSONSchemaWritable,
+  type JSONSchemaMutable,
   type JSONValue,
   markAsStatic,
   type Module,
   type ModuleFactory,
+  type Mutable,
   NAME,
   type Node,
   type NodeFactory,
@@ -67,7 +68,6 @@ export {
   unsafe_originalRecipe,
   unsafe_parentRecipe,
   type UnsafeBinding,
-  type Writable,
 } from "./types.ts";
 export { type Schema, schema } from "./schema-to-ts.ts";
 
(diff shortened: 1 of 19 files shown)


ubik2 (approved):

ubik2 (approved):
LGTM!

PR #1103, opened 2025-04-24
Change some places where we declared a schema "as JSONSchema"

ubik2:
Change some places where we declared a schema "as JSONSchema" instead of "satisfies JSONSchema".

Added a Mutable template to remove readonly flags from an object.


ubik2:
This will probably go away, since @jsantell already started this work in #1102.

ubik2:
Closed as dup of #1102

PR #1107, opened 2025-04-25
fix: ignore schema incopmpatible writes

Gozala:
Per https://linear.app/common-tools/issue/CT-318/disallow-writing-data-that-isnt-in-the-schema

I'm not sure if we should simply ignore writes that are for keys not in the document or through

### DIFF
--- a/jumble/public/module/charm/sandbox/bootstrap.js
+++ b/jumble/public/module/charm/sandbox/bootstrap.js
@@ -46,12 +46,18 @@ window.useDoc = function (key) {
 
   // Update function
   const updateDoc = newValue => {
-    if (typeof newValue === "function") {
-      newValue = newValue(doc)
+    // only perform write if sourceData contains this key otherwise we ignore
+    // write which somewhat aligns with returning `undefined`.
+    // TODO(@seefeld) - should we throw here instead or use different method
+    // to decide whether to send a write request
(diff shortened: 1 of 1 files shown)


jumble/public/module/charm/sandbox/bootstrap.js
```diff
@@ -46,12 +46,18 @@ window.useDoc = function (key) {
...
+    // write which somewhat aligns with returning `undefined`.
+    // TODO(@seefeld) - should we throw here instead or use different method
+    // to decide whether to send a write request
+    if (key in window.sourceData) {
+      if (typeof newValue === "function") {
+        newValue = newValue(doc)
+      }
+      console.log("useDoc", key, "written", newValue)
+      setDocState(newValue)
+      window.parent.postMessage({ type: "write", data: [key, newValue] }, "*")
```
ubik2, line 59: If we are ignoring writes, can we get a console.log

Gozala:
> I'm not sure optional properties would show up in sourceData. So this would mean we can't set them.

@seefeldb sounds like that would be a problem here also https://github.com/commontoolsinc/labs/pull/1096 isn't it, as in property may have being optional and instead of using react provided default and writing it we will end up returning `undefined` 🤔

jsantell:

PR #1108, opened 2025-04-25
chore: Fix up importer recipe to handle reauth and throttling

jsantell:


recipes/gmail.tsx
```diff
@@ -499,14 +581,8 @@ export async function process(
...
-      if (i + batchSize < newMessages.length) {
-        console.log("Waiting 1 second before next batch...");
-        await sleep(1000);
-      }
     } catch (error) {
-      console.error("Error processing batch:", error);
+      console.error("Error processing batch:", error?.message);
       // Optional: add longer delay and retry logic here if needed
     }
   }
```
anotherjesse, line 588: the `allDetailedMessages` below isn't used/populated

recipes/gmail.tsx
```diff
@@ -174,24 +174,196 @@ const updateLimit = handler({
...
+          "Content-Type": `multipart/mixed; boundary=${boundary}`,
+        },
+        body: batchBody,
+      },
+    );
+
+    const responseText = await batchResponse.text();
+    console.log("Received batch response of length:", responseText.length);
+
+    const parts = responseText.split(`--batch_`)
```
anotherjesse, line 280: the response inside the --batch_ can include a 429 as well!  Does this just eat those?

https://developers.google.com/workspace/gmail/api/guides/batch#batch-example-response

jsantell: I can't find explicit docs on the matter, but if 429's result in a 200 HTTP response with JSON containing the error code, I'd have to hope that only 200 responses are represented as text

anotherjesse: that was my original assumption...

200 in the outside just means "I did the work of fetching each of the sub items and putting it together"

inside each you get an http status code/...

```
--batch_foobarbaz
Content-Type: application/http
Content-ID: <response-item2:12930812@barnyard.example.com>

HTTP/1.1 200 OK
Content-Type: application/json
Content-Length: response_part_2_content_length
ETag: "etag/sheep"

{
  "kind": "farm#animal",
  "etag": "etag/sheep",
  "selfLink": "/farm/v1/animals/sheep",
  "animalName": "sheep",
  "animalAge": 5,
  "peltColor": "green"
}

--batch_foobarbaz
Content-Type: application/http
Content-ID: <response-item3:12930812@barnyard.example.com>

HTTP/1.1 304 Not Modified
ETag: "etag/animals"

--batch_foobarbaz--
```

anotherjesse: there is a content type / id / length / ... and http version + status code

jsantell: While I don't think this can occur for failure codes, this should handle it

jsantell: Nope it definitely occurs

PR #1111, opened 2025-04-27
Cell recipes redux

anotherjesse:
Store the recipes in the space with the charms!

New recipeManager `runner/src/recipe-manager.ts`
- Unified, persistent recipe + syncing layer
- **recipes are only stored in the space**, but can optionally be published to blobby.
  - SpellBook will now publish to blobby on a share
    - this opens up a place for versioning/users keeping up-to-date with published recipes)
- **Auto-migration** - if the recipe source/metadata isn't stored in the space, attempt to fetch for blobby
  - _for backwards migrations, we will still publish to blobby ALWAYS, until we are sure we aren't going to need to roll back_

Other changes:
- compile now throws if default export missing (it was returning undefined)


anotherjesse:
rebasing #1088 was complex, and so I want to keep it in its original state while I try to finish this branch

cli/cast-recipe.ts
```diff
@@ -55,11 +55,6 @@ async function castRecipe() {
     // Load and compile the recipe first
     console.log("Loading recipe...");
     const recipeSrc = await Deno.readTextFile(recipePath!);
-    const recipe = await compileRecipe(recipeSrc, "recipe", []);
-
-    if (!recipe) {
-      throw new Error(`Failed to compile recipe from ${recipePath}`);
-    }
 
     console.log("Recipe compiled successfully");
```
ubik2, line 59: should move to line 71 to be after the compileRecipe call

background-charm-service/cast-admin.ts
```diff
@@ -43,7 +43,7 @@ const identity = await getIdentity(
 );
 
 storage.setRemoteStorage(new URL(toolshedUrl));
-setBobbyServerUrl(toolshedUrl);
```
bfollington, line 46: Goodbye `bobby` 🫡

jumble/src/components/commands.ts
```diff
@@ -447,9 +447,6 @@ async function handleUseDataInSpell(ctx: CommandContext) {
           const charm = await ctx.charmManager.get(ctx.focusedCharmId);
           if (!charm) throw new Error("No current charm found");
           const argument = ctx.charmManager.getArgument(charm);
-          if (!argument) {
-            throw new Error("No sourceCell/argument found for current charm");
-          }
```
bfollington, line 452: I don't see why this needed to go...

jumble/src/views/CharmDetailView.tsx
```diff
@@ -926,6 +914,20 @@ const DataTab = () => {
...
+        charmManager.getArgument(charm).get(),
+      ) as Record<string, any>;
+    } else {
+      argumentJson = {};
+    }
+  } catch (error) {
+    console.warn("Error translating argument to JSON:", error);
+    argumentJson = {};
+  }
+
```
bfollington, line 930: nit: wrapping in `useMemo` would be better again for render perf

runner/src/recipe-manager.ts
```diff
@@ -0,0 +1,327 @@
...
+
+// Type guard to check if an object is a Recipe
+function isRecipe(obj: Recipe | Module): obj is Recipe {
+  return "result" in obj && "nodes" in obj;
+}
+
+// FIXME(ja): what happens when we have multiple active spaces... how do we make
+// sure we register the same recipeMeta in multiple spaces?
+const recipeMetaMap = new WeakMap<Recipe, Cell<RecipeMeta>>();
+const recipeIdMap = new Map<string, Recipe>();
```
bfollington, line 54: Might be better to wrap these as members of the class rather than in the module? It makes multiple-instances of `RecipeManager` share the same cache which is surprising behaviour, to me.

runner/src/recipe-manager.ts
```diff
@@ -0,0 +1,327 @@
...
+  generateRecipeId(recipe: Recipe | Module, src?: string) {
+    let id = recipeMetaMap.get(recipe as Recipe)?.get()?.id;
+    if (id) {
+      console.log("generateRecipeId: existing recipe id", id);
+      return id;
+    }
+
+    id = src
+      ? createRef({ src }, "recipe source").toString()
+      : createRef(recipe, "recipe").toString();
```
bfollington, line 92: Is this how we have always done this?

runner/src/recipe-manager.ts
```diff
@@ -0,0 +1,327 @@
...
+      .finally(() => inProgressCompilations.delete(recipeId)); // tidy up
+
+    inProgressCompilations.set(recipeId, compilationPromise);
+    return await compilationPromise;
+  }
+
+  /**
+   * Load a recipe from Blobby, returning the recipe and recipeMeta
+   */
+  // FIXME(ja): move this back to blobby!
```
bfollington, line 204: Is this fixed?

bfollington (commented):
LGTM overall, couple of questions I'd like to answer before merging. I will try and answer them myself but might run out of time today.

cli/cast-recipe.ts
anotherjesse: 🤣  good catch!

jumble/src/components/commands.ts
anotherjesse: image

because now we throw instead of returning undefined

jumble/src/views/CharmDetailView.tsx
anotherjesse: good idea - should we do it for this below as well:

```
              value={translateCellsAndStreamsToPlainJSON(charm.get()) ?? {}}
```

I had moved it out of inline jsx because getArgument now throws (see above)

runner/src/recipe-manager.ts
anotherjesse: my original thought was the same as you.  I switched when I realized a recipe is unique (hopefully) by ID.  And thus recompiling it shouldn't change it.  

the "MetaMap" is going away once we make the RecipeMeta a LinkedCell in the Charm Schema.  Then all the "load this / save this / ..." go away.. loading the charm gives you the these attributes

anotherjesse: so I think we can keep them as is?

runner/src/recipe-manager.ts
anotherjesse: yep, I had only referenced it as "hmm" because iirc, createRef of the recipe graph can be expensive... @seefeldb said something like that (perhaps it wasn't recipes?)

runner/src/recipe-manager.ts
anotherjesse: oh, I was thinking the logical fetch get/put against blobby maybe should be back in blobby-storage.ts - since they are mostly self-contained...

or at least these parts:

get

```
    const response = await fetch(`${getBlobbyServerUrl()}/spell-${recipeId}`);
    if (!response.ok) {
      throw new Error(`Failed to fetch recipe ${recipeId} from blobby`);
    }

    const recipeJson = await response.json() as {
      src: string;
      spec?: string;
      parents?: string[];
    };
```

and all/the majority of publishToBlobby 

---

but I don't think we need to decide this for this PR to land

background-charm-service/cast-admin.ts
anotherjesse: 🖖

anotherjesse: bobby is how you know I still type code... but I use tab complete in other files, so a spelling issue in one place replicates throughout the codebase

PR #1113, opened 2025-04-27
Intent-based preview

bfollington:
- **Find existing charms during preview generation**
- **Thread preview through the stack**
- **Simpler, faster preview generation**
- **Improved schema handling in workflow**

- [ ] extending / mentioning seems to be producing an malformed schema


llm/src/prompts/spec-and-schema-gen.ts
```diff
@@ -183,10 +183,10 @@ export function formatForm(form: WorkflowForm) {
   return `
 <goal>${form.input.processedInput}</goal>
 <plan>${
-    (form.plan?.steps ?? []).map((step) => `<step>${step}</step>`).join("\n")
+    (form.plan?.features ?? []).map((step) => `<feature>${step}</feature>`)
+      .join("\n")
   }</plan>
```
bfollington, line 188: Don't include if empty

PR #1115, opened 2025-04-28
Use the schema storage flag

ubik2:
Switch over to the schema storage type.
This was behind a feature flag to make sure it didn't break anything, but I think this is a reasonable time to get some actual testing.


jakedahn (approved):
shippit

PR #1121, opened 2025-04-29
fix: Do not rerender everytime CharmRunner's charm changes

jsantell:
Currently, `submitAutomaticErrorFeedback` depends on `charm` argument, which changes when something(?) in the space is updated (e.g. making a new charm in the same space), causing a viewed charm to rerender. As the `charm` param can be `!==` to a previous `charm` object while still representing the same ID, we can't use it as a stable dependency. Instead, rely on `id` and request a charm from the manager (whose instance is stable) when needed.


Gozala (approved):
Thanks for fixing this

PR #1124, opened 2025-04-29
Handle direct string response from LLM as well as `.content` unwrap

bfollington:


bfollington:
Covered in https://github.com/commontoolsinc/labs/pull/1126

PR #1128, opened 2025-04-30
fix: attempt to prevent use of `useReactiveCell` outside of component

Gozala:
In my tests it seems to prevent errors we were seeing, but I don't have a very high confidence


bfollington (approved):
LGTM! `onReady` is still a suboptimal API (we should just expose an `<App />` component ideally to make the footgun impossible) but an improvement regardless.

PR #1132, opened 2025-05-01
chore: Align isObject/isRecord usage

jsantell:
* Align the multiple, slightly different implementations of `isObject` (and `isRecord`) -- sometimes as "non-array" objects, sometimes including arrays, and the differences between `object` and `Record` usage.
* Remove the root entry for `@commontools/utils` (originally tracking down a bug for something pulling in too many modules, but was misleading). Deno requires a root export, so a compile time (no exports) or runtime error fires if doing so to avoid inadvertantly pulling in all utils
    

---

## Summary by mrge
Standardized all isObject and isRecord checks across the codebase by providing consistent utility functions in utils/types, and removed duplicate or inconsistent implementations.

- **Refactors**
  - Replaced custom isObject/isRecord functions with shared versions from utils/types.
  - Removed unused or duplicate type-checking helpers.
  - Updated imports to use the new shared utilities.

### DIFF
--- a/iframe-sandbox/src/ipc.ts
+++ b/iframe-sandbox/src/ipc.ts
@@ -1,4 +1,28 @@
 import { type JSONSchema } from "@commontools/builder";
+import { isObject } from "@commontools/utils/types";
+
+export const isJSONSchema = (source: unknown): source is JSONSchema => {
+  if (!isObject(source)) {
+    return false;
+  }
+
+  if (!("type" in source) || !source.type) {
+    return "anyOf" in source && Array.isArray(source.anyOf);
+  }
+
+  switch (source.type) {
+    case "object":
+    case "array":
+    case "string":
+    case "integer":
+    case "number":
+    case "boolean":
+    case "null":
+      return true;
+    default:
+      return false;
+  }
+};
 
 // Types used by the `common-iframe-sandbox` IPC.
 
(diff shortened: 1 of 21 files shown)


iframe-sandbox/src/ipc.ts
```diff
@@ -107,34 +108,12 @@ export function isGuestError(e: object): e is GuestError {
...
-    case "number":
-    case "boolean":
-    case "null":
-      return true;
-    default: {
-      return Array.isArray(source?.anyOf);
-    }
-  }
-};
+  "intent" in source && typeof source.intent === "string" &&
```
seefeldb, line 113: nit: isn't `"..." in source` redundant here, since `isObject(source)` already makes sure this is an object and `typeof source.<..>` would be `"undefined"` if the key doesn't exist?

seefeldb (approved):

iframe-sandbox/src/ipc.ts
jsantell: That's what I thought -- the `Record` type (where `Record<number, unknown>` could be an array) allows this but `object` types do not (and had to add it in a few places) -- maybe there's a better way to do this? but we do sometimes explicitly want an object-like (`{}`) things and not an array or a record

jsantell: We'd get a `Property 'intent' does not exist on type 'object'` otherwise

jsantell: > We'd get a `Property 'intent' does not exist on type 'object'` otherwise

This is a typescript check, and we check that x is not null (`isObject(..)`) and TS `null` is not `object` (unlike JS) -- *usually* we want those checks ("property does not exist"), though using `isRecord` previously allowed any key (TIL as well)

PR #1134, opened 2025-05-05
Email processing recipe + prework

seefeldb:
Email processing still in progress, but want to commit this now for all the related changes.


seefeldb:
@jsantell I add a new Deno.env in LLMClient, but I'd rather use a general env helper. Instead of adding a dependency on builder/ should we add something in utils/, e.g. to get the API URL?

@anotherjesse @jsantell FYI for changes to cli/main.ts

jsantell:
> @jsantell I add a new Deno.env in LLMClient, but I'd rather use a general env helper. Instead of adding a dependency on builder/ should we add something in utils/, e.g. to get the API URL?
> 
> @anotherjesse @jsantell FYI for changes to cli/main.ts

My preference would be for LLMClients to own this URL, set at a high level, but I think that was removed in lieu of the current impl. If we must use globals, a util in e.g. `@commontools/utils/env` makes sense. 

The env set/getters in `builder` are specifically for setting the environment within recipes; in most other cases, we shouldn't be using this (recipe context isn't necessarily host context), but if this is set in the LLM builtin, that seems the most consistent (e.g. everything in recipes uses `env.apiUrl`), but that's not the same env for all LLMClients (the builtin scenario being the only one running within a recipe env), and with global URLs, they must all share the same root. In practice these are often the same URL, but the URLs sort of mean different things throughout (similarly, setting remote storage, blobby, env all with the same URL).

`llm` will run in browsers and deno. The `isBrowser`/`isDeno` functions in `@commontools/utils/env` could be used in the static global toolshed URL rather than having env-specific handling in e.g. `llm` and throughout (as we also use Deno env vars and Vite env vars in different locations)

PR #1136, opened 2025-05-06
Json mode recipes

anotherjesse:
image

### DIFF
--- a/recipes/numberStoryTime.tsx
+++ b/recipes/numberStoryTime.tsx
@@ -1,0 +1,93 @@
+import { h } from "@commontools/html";
+import {
+  derive,
+  handler,
+  ifElse,
+  JSONSchema,
+  lift,
+  llm,
+  NAME,
+  recipe,
+  schema,
+  str,
+  UI,
+} from "@commontools/builder";
+
+const inputSchema = schema({
+  type: "object",
+  properties: {
+    number: { type: "number", default: 0 },
+  },
+}) satisfies JSONSchema;
+
+const nextNumber = handler<{}, { number: number }>(
+  (_, state) => {
+    state.number = (state.number ?? 0) + 1;
+  },
+);
+const randomNumber = handler<{}, { number: number }>(
+  (_, state) => {
+    state.number = Math.floor(Math.random() * 102) - 1;
+  },
+);
+
+const generateStory = lift(({ number }: { number: number }) => {
+  return {
+    system:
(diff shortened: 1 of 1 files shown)


recipes/numberStoryTime.tsx
```diff
@@ -0,0 +1,164 @@
...
+  (_, state) => {
+    state.number = Math.floor(Math.random() * 102) - 1;
+  },
+);
+
+const generateStory = lift(({ number }: { number: number }) => {
+  return {
+    system:
+      "You are the parent of a young child who loves to learn about numbers. Luckily for your child, you are a historian of numbers and when the child says a number you make up an interesting story about it, including the history of the number.",
+    messages: [`Tell me a story about ${number}`],
```
seefeldb, line 38: you could use

```
str`Tell me a story about ${number}`
```

here and not make it a `lift`. just in case.

recipes/numberStoryTime.tsx
```diff
@@ -0,0 +1,164 @@
...
+  return {
+    system:
+      "You are an imaginative illustrator. Read the following story and create a vivid, detailed image prompt for an AI image generator. Focus on the most unique, visual, and interesting elements mentioned in the story. Include any historical figures, objects, or settings described. Be specific and creative, and use a storybook illustration style. Here is the story:",
+    messages: ["Story: " + story],
+  };
+});
+
+const getImageUrl = lift(({ prompt }: { prompt: string | undefined }) => {
+  if (!prompt) {
+    prompt =
```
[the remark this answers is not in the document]
seefeldb: Or use `??`... or fancy (just FYI:) you can provide schemas for `lift`, as first and second parameter, and then add the default prompt as default value in the schema. However that will only apply if prompt is `undefined`, not if prompt is `""`.

recipes/numberStoryTime.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+export default recipe(inputSchema, ({ number }) => {
+  const { result: story, pending: storyPending } = llm(
+    generateStory({ number }),
+  );
+  const { result: prompt } = llm(
+    generatePrompt({ story }),
+  );
+  return {
+    [NAME]: str`The story of ${derive(number, (number) => number)}`,
```
seefeldb, line 69: oh, i'm curious why the derive was necessary here?

recipes/numberStoryTime.tsx
anotherjesse: heh, I forgot about this

recipes/numberStoryTime.tsx
anotherjesse: no, I just forgot about str`

jsantell:

PR #1142, opened 2025-05-08
feat: Introduce Page#applyConsoleFormatter to stringify objects in CI/seeder

jsantell:
Tempting to add this automatically to e.g. `.goto(url)`, but there are many reasons the document can reload, so lets make it explicit.
    

---

## Summary by mrge
Added Page#applyConsoleFormatter to rewrite console methods and ensure objects are stringified in logs during CI and seeder runs.

- **New Features**
  - Added applyConsoleFormatter method to Page for better console output.
  - Updated tests and seeder to use the new formatter.

### DIFF
--- a/packages/integration/page.ts
+++ b/packages/integration/page.ts
@@ -15,6 +15,7 @@ import { sleep } from "@commontools/utils/sleep";
 import { Mutable } from "@commontools/utils/types";
 import * as path from "@std/path";
 import { ensureDirSync } from "@std/fs";
+import { ConsoleMethod } from "./console.ts";
 
 // To handle `console` events from `Page`, logging to outer context:
 //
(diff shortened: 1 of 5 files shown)


packages/integration/page.ts
```diff
@@ -80,6 +81,48 @@ export class Page extends EventTarget {
...
+      // If console has already been stubbed for this document, abort.
+      if (globalThis[trueConsoleKey]) {
+        return;
+      }
+      const trueConsole = globalThis.console;
+      const newConsole = Object.create(null);
+      for (const method of methods) {
+        newConsole[method] = (...args: any[]) => {
+          const formatted = args.map((value) => {
+            if (value && typeof value === "object") {
```
ubik2, line 106: You also have the `fmtConsole` version in background-charm-service worker. It's small enough that duplication is fine, though (and it shouldn't be included from that location).

ubik2 (approved):
LGTM!

packages/integration/page.ts
jsantell: Ya this formatter was pulled from there, though quite different environments (this `evaluate` method is stringified, whereas in worker, it's running in the same JS context and the handler receives real objects, but its our netdata/systemd logs that need a string)

PR #1145, opened 2025-05-12
Reliably set `running` to undefined before resolving

seefeldb:
so that `idle` can't loop infinitely
    

---

## Summary by mrge
Ensured the `running` state is always set to undefined before resolving, preventing infinite loops in the `idle` state.

- **Bug Fixes**
  - Moved `running = undefined` to reliably clear the state for both sync and async actions.


anotherjesse:
this seems to resolve the memory leak

anotherjesse (commented):
lgtm ... I had one question about people sometimes sending non-promises ... but perhaps not an issue at this layer

PR #1146, opened 2025-05-12
don't require system message in built-in llm call

seefeldb:
just require at least one message


anotherjesse (approved):
toolshed's openapi schema already has system is optional! so this change is consistent

anotherjesse:
https://toolshed.saga-castor.ts.net/reference#tag/ai-language-models/POST/api/ai/llm

PR #1147, opened 2025-05-12
align space did derivation from name with jumble

seefeldb:
## Summary by mrge
Updated the default operator password in the CLI from "implicit trust" to "common user" to match the jumble naming convention.


seefeldb:
@jsantell @Gozala cc'ing you as i don't like that i copied the hardcoded id from jumble, but i figure this will change anyway once we move along the full ACL road (no need to review, just FYI)

PR #1148, opened 2025-05-12
run jsx fragment tests

seefeldb:
## Summary by mrge
Renamed the JSX test file to ensure fragment tests are included in the test suite.


seefeldb:
@Gozala i noticed the JSX tests didn't run since the file wasn't call .test -- flagging in case this was intentional.

ubik2 (approved):

PR #1154, opened 2025-05-19
Fix mutable path traversal

anotherjesse:
## Summary
- fix path() mutating keyPath
- add regression test to ensure keyPath is preserved

## Testing
- `cd packages/html && deno test --allow-env`


seefeldb:
LOL, codex managed to pick dead code. AFAIK `path.ts` isn't actually used anywhere anymore and we should delete it.

PR #1157, opened 2025-05-20
IfElse using cell abstractions

seefeldb:
Should be a cleaner implementation, leveraging more of the better tested logic around aliases and so on.


anotherjesse:
I really like this cleanup!  We had talked through `ifElse` a while ago, and this code is much clearer

ubik2 (approved):

ellyxir (approved):

PR #1159, opened 2025-05-20
Don't run the update for charms that are no longer active

ubik2:
I should be able to fetch these lists without fetching everything, but currently, I think we do fetch the whole bit.
When I have client side schema traversal, I should be in better shape.

Fixes CT-39


packages/background-charm-service/src/worker.ts
```diff
@@ -127,6 +130,31 @@ async function cleanup(): Promise<void> {
   initialized = false;
 }
 
+async function isInCharmList(
```
anotherjesse, line 133: this feels like it could be part of the charm manager? it looks like this is triggered from runCharm which has an active manager for the space?

PR #1162, opened 2025-05-20
Add AGENTS instructions

seefeldb:
add AGENTS instructions translated from CLAUDE.md

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,16 +1,18 @@
-# Repository Guidelines for Codex
+# Repository Guidelines for AI Agents
 
 The instructions in this document apply to the entire repository.
 
-## Build & Test
+## Basics
+
+### Build & Test
 
 - Check typings with `deno task check`.
-- Run all tests using `deno task test-all`.
+- Run all tests using `deno task test`.
 - To run a single test file use `deno test path/to/test.ts`.
 - To test a specific package, `cd` into the package directory and run
   `deno task test`.
 
-## Formatting
+### Formatting
 
 - Line width is **80 characters**.
 - Indent with **2 spaces**.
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,44 @@
+# Repository Guidelines for Codex
+
+The instructions in this document apply to the entire repository.
+
+## Build & Test
+
+- Check typings with `deno task check`.
+- Run all tests using `deno task test-all`.
```
anotherjesse, line 8: this should be `deno task test` now

AGENTS.md
```diff
@@ -0,0 +1,44 @@
...
+- To test a specific package, `cd` into the package directory and run
+  `deno task test`.
+
+## Formatting
+
+- Line width is **80 characters**.
+- Indent with **2 spaces**.
+- **Semicolons are required.**
+- Use **double quotes** for strings.
+- Always run `deno fmt` before committing.
```
anotherjesse, line 19: I wonder if we need the lines above if we run `deno fmt`?

AGENTS.md
```diff
@@ -0,0 +1,44 @@
...
+
+## TypeScript
+
+- Export types explicitly using `export type { ... }`.
+- Provide descriptive JSDoc comments on public interfaces.
```
anotherjesse, line 24: perhaps we should add something that we should check affected packages README.md to update them?

anotherjesse: might be worth experimenting with this

anotherjesse:
@jsantell I wonder if some of your recommendations could be added here?

anotherjesse:
Also I wonder if we could have a single line in CLAUDE.md that says "read AGENTS.md to get instructions"

PR #1163, opened 2025-05-20
style shouldn't be set until after connected

anotherjesse:
the this.style would throw an error (try loading the example from recipes), now it doesn't

addresses CT-367

### DIFF
--- a/packages/ui/src/components/common-fragment.ts
+++ b/packages/ui/src/components/common-fragment.ts
@@ -6,6 +6,7 @@ export class CommonFragmentElement extends HTMLElement {
       .appendChild(document.createElement("slot"));
   }
 
+  // Tell engine to ignore this element for layout purposes
   connectedCallback() {
     this.style.display = "contents";
   }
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/common-fragment.ts
```diff
@@ -4,10 +4,11 @@ export class CommonFragmentElement extends HTMLElement {
     this.attachShadow({ mode: "open" })
       // Add a slot to display the children
       .appendChild(document.createElement("slot"));
-
-    // Tell engine to ignore this element for layout purposes
-    this.style.display = 'contents';
   }
+
+  connectedCallback = () => {
```
anotherjesse, line 9: umm, this was supposed to be connectedCallback {

anotherjesse:
reading https://adactio.com/journal/tags/customelements#:~:text=connectedCallback%28%29%20,%E2%80%A6 explained that how to use connectedCallback - but it is from 2015, so I'm not certain if this is still the right way.

Guessing that @Gozala or @jsantell will be able to point in the right direction

packages/ui/src/components/common-fragment.ts
anotherjesse: fixed

PR #1164, opened 2025-05-21
Fix map with undefined input

anotherjesse:
## Summary
- handle undefined arrays in map builtin by resetting progress
- add regression test for undefined array input

## Testing
- `deno task test` in `packages/runner`
- `deno task test` *(fails: could not load packages due to network restrictions)*


anotherjesse:
fail

anotherjesse:
addresses #CT-382

PR #1165, opened 2025-05-21
Fix map builtin on undefined lists

anotherjesse:
## Summary
- clean up map builtins when source array goes undefined or shrinks
- test map builtin works when list becomes undefined

## Testing
- `deno test packages/runner/test`

### DIFF
--- a/packages/runner/src/builtins/map.ts
+++ b/packages/runner/src/builtins/map.ts
@@ -57,6 +57,10 @@ export function map(
   // between the source list and the result list.
   let initializedUpTo = 0;
 
+  // Keep track of result cells so they can be cancelled when the list shrinks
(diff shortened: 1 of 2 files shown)


packages/runner/src/builtins/map.ts
```diff
@@ -57,13 +57,23 @@ export function map(
   // between the source list and the result list.
   let initializedUpTo = 0;
 
+  // Keep track of result cells so they can be cancelled when the list shrinks
+  const resultCells: DocImpl<any>[] = [];
+  const resultCancels: Array<(() => void) | undefined> = [];
```
anotherjesse, line 62: is this a good idea?  should this actually be using a cancelGroup and returning it.

anotherjesse:
this doesn't work

PR #1167, opened 2025-05-21
Add CharmPreview schema

anotherjesse:
## Summary
- define `CharmPreview` type and schema
- add `charmPreviewListSchema`
- use new preview schema for pinned/trash lists
- export preview types from package index
- update charm_demo CLI to use new schema

## Testing
- `deno task test` *(fails: Test requires Chrome)*


anotherjesse:
this is also not a good solution

PR #1168, opened 2025-05-21
feat: Introduce `@commontools/static`

jsantell:
This is for #1153, in addition to previously wanting static assets for e.g. prompts and possibly `bootstrap.js`. For now, I ported over one prompt ([system.md](https://github.com/commontoolsinc/labs/blob/138af9ef894754568c3dbbecfda8c83666b96d61/packages/static/assets/prompts/system.md)) to this pattern.

If this is something not needed for prompts, I can roll back the system.md prompt addition.

Tagging @bfollington @jakedahn for reviews of the prompt and toolshed changes

-----

* feat: Introduce `@commontools/static`
  * A way of lazily accessing static content in both deno/browser.
  * Tested in both deno/browser.
  * Currently using a prompt as a static asset, but this is needed for hosting type definitions.
* Add `/static` route to `toolshed` for resolving browser-based assets
  * Notably toolshed uses the same asset cache for serving browser static assets as other instances inside toolshed
  * Share mime type utility between the frontend routes (a different category of static assets)
* Update `deno-web-test` to include additional static content in the test server
  * Needed for testing `@commontools/static`
  * Noted issue that since `Deno.test` is stubbed in for `deno-web-test`, the `isDeno()` utility in `@commontools/utils/env` needed to be more precise, and avoided using non-posix `@std/path` imports (that fail on `Deno?.build.os`). This is only an issue in the unique bundled environment that `deno-web-test` runs in.


tasks/build-binaries.ts
```diff
@@ -72,6 +73,10 @@ class BuildConfig {
     return this.path("packages", "toolshed", "COMPILED");
   }
 
+  staticAssetsPath() {
+    return this.path("packages", "static", "assets");
```
jsantell, line 77: TODO: Use `getBaseUrl` from `@commontools/static` here

packages/toolshed/lib/mime-type.ts
```diff
@@ -0,0 +1,28 @@
...
+const DEFAULT_MIME_TYPE = "application/octet-stream";
+
+export function getMimeType(filename: string): string {
+  const split = basename(filename).split(".");
+  if (split.length < 2) {
+    // No "." in name.
+    return DEFAULT_MIME_TYPE;
+  }
+  const ext = split.pop()!;
+  if (MIME_TYPES[ext]) {
```
jsantell, line 24: `return MIME_TYPES[ext] ?? DEFAULT_MIME_TYPE`

packages/charm/src/iframe/prompt.ts
```diff
@@ -1,18 +1,16 @@
...
-  type GenerationOptions
 } from "@commontools/llm";
-
-import { extractUserCode, systemMd } from "./static.ts";
+import { extractUserCode, staticSystemMd } from "./static.ts";
 
 export const RESPONSE_PREFILL = "```javascript\n";
 
-export const buildPrompt = ({
+export const buildPrompt = async ({
```
jsantell, line 13: Using lazy-loaded prompts incurs an async barrier here -- I wanted to use a prompt that's used in the integration test, and making `buildPrompt` async was the easiest way forward, no opinions here on how prompts should be handled. Can update to whatever preferred interface you'd like here for this prompt @bfollington, but want to limit to porting a single prompt for now

bfollington: This doesn't seem like a big deal to me, we're already awaiting basically every second line during the generation flow as it stands. We can move towards this lazy import pattern for prompts as we migrate each one individually.

bfollington (approved):
LGTM, replied to the specific comment

PR #1169, opened 2025-05-21
Change the map function so that when the result.get() is undefined, we set it to the empty array

ubik2:
Change the map function so that when the result.get() is undefined, we set it to the empty array. This prevents the subsequent read of length from failing.

### DIFF
--- a/packages/runner/src/builtins/map.ts
+++ b/packages/runner/src/builtins/map.ts
@@ -63,7 +63,7 @@ export function map(
     // If the list is undefined it means the input isn't available yet.
     // Correspondingly, the result should be []. TODO: Maybe it's important to
     // distinguish empty inputs from undefined inputs?
-    if (list === undefined) {
+    if (list === undefined || result.get() === undefined) {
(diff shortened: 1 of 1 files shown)


ubik2:
When I first patched this, I hadn't seen @anotherjesse's full PR.
Given the other changes in the `list === undefined` clause, I'll probably break out my `result.setAtPath` call into its own conditional clause immediately before the list check.

seefeldb:
Yeah, let's merge the other PR first.

And yes, I think it would be just a separate path to make sure result.get() isn't undefined after running through the loop, which could be the case if aborted (the list undefined case you caught), but also if the source array is an empty array (https://github.com/commontoolsinc/labs/blob/1a206020a057b53825c4478fe601fc6bf97fad10/packages/runner/src/builtins/map.ts#L110 never being called)

ubik2:
It looks like @anotherjesse closed his PR, since it didn't give him the desired result.
I still broke out my check, so if similar changes are introduced, they won't conflict.

PR #1170, opened 2025-05-21
Reduce recipe saving and loading to top-level charms

seefeldb:
This PR removes a bunch of recipe saving of writing for inner charmlets that doesn't correctly work anyway. Inner charmlets, e.g. those for `.map` calls, work by being spawned by parent charms. That is, we can't correctly rehydrate them individually anyway right now, so let's not also keep storing all these recipes that are at best sometimes working as intended.

Code changes:
* move recipe loading and saving out of runSynced - it now assumes there already is a recipe provided
* move inputs syncing from CharmManager to runSynced
* removed redundant syncRecipe call
* fixed post-run sync call

### DIFF
--- a/packages/charm/src/manager.ts
+++ b/packages/charm/src/manager.ts
@@ -372,7 +372,9 @@ export class CharmManager {
     if (runIt) {
       // Make sure the charm is running. This is re-entrant and has no effect if
       // the charm is already running.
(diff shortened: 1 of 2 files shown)


ubik2 (approved):

anotherjesse:
my only comment is "could we move even more out" - the load charm logic still has a lot of stuff that feels out of place

BUT - this LGTM!

PR #1172, opened 2025-05-23
Added support for deleting from the curl command tool

ubik2:
Fixed a bug in cache where we were including the since field when calculating the cause.
Added an error message to cache when our queries fail, since they were being silently ignored.


ubik2:
I realized that if we have objects that were both created and updated with since being included in the cause, this will break them.

seefeldb (approved):

PR #1173, opened 2025-05-23
Add code quality task list

seefeldb:
This is an experiment: I asked codex to judge the current code according to the guidelines and suggest tasks for itself to improve.

DO NOT COMMIT -- i'm actually just running all these tasks in the background now and so either the list will be shorter or entirely gone.


jsantell:

PR #1176, opened 2025-05-27
Improve worker interactions with classified charms for CT-409

ubik2:
- Added the entity id of the bgUpdater doc in a comment.
- Formatting to remove long line in service.ts
- Change the getBGCharms to use a privileged query to get that cell. It will generally contain references to secret data.
- Changed from isActiveCharm to getActiveCharm in manager.ts and worker.ts, so we can also use this to retrieve a charm that we've already fetched.
 - When instantiating the CharmManager in the background-charm-service, wait until it's ready before we continue. This avoids having other paths that try to load the same docs without privilege clobber our privileged load. When the client schema traversal work happens, it should account for multiple loads of the same doc with different schemas if needed. Also include the error cause in the worker, so we keep our original stack trace accessible.
- Change CharmManager to fetch the charm lists with "secret" level privilege. This can be removed if we add handling that optionally escalates privilege if needed. Right now, we don't expose this in UI, so better to just start out asking with privilege. We use this schema with the syncCell calls, but don't make it part of the main schema, since we don't want our writes to convert the top level object to be classified.
- Changed storage.ts to include an optional label in the Job entry. This will use cfc to extract needed schemas from a doc's schemaContext so that when we load dependent docs, we include that classification claim. This makes it so our initial fetch of the charms list with privilege will load the charms with privilege, and all the dependent docs with privilege. It is essentially the same system that we have on the server schemaQueries where we traverse the links with our initial classification claim. This means _batchForDoc takes an optional label, and when we call _ensureIsSyncedById, we attach that label to the schema. This also means we need to track these labels in our loading set (now a map). These changes are temporary. Eventually, these dependent fetches should go away, since we can use schema queries to load them (the traverse will still need to attach the real objects to the links).
- Log query errors to console in Replica.pull, since we don't generally check the returned value. This could be reverted later if we change our callers to deal with it.


ubik2:
I'm not certain it's required, but it's possible I'll also need to land the other PR that excludes classified results when they would otherwise be included, instead of simply returning a failure for the entire query.

seefeldb (approved):

PR #1178, opened 2025-05-27
Exclude classified results from querySchema results

ubik2:
Instead of returning an AuthorizationError when our schema causes us to include a classified document, just exclude those documents from the result (also excludes any docs reachable through those documents, since otherwise we could reveal classified data).
We log these excluded results, but it would be better to list them together with their required classifications in the client's result (later).


ubik2:
Addresses CT-410

PR #1183, opened 2025-05-28
chore: Add more logs and handle when initialization is rejected, seen on invalid ipc requests

jsantell:


packages/background-charm-service/src/space-manager.ts
```diff
@@ -283,18 +288,28 @@ export class SpaceManager {
...
-        console.log(`${this.did} Worker controller ready for work`);
-      }),
-    ];
+    const initialize = newWorker.initialize();
 
     if (previousWorker) {
       console.log(`${this.did} Restarting Worker Controller`);
       previousWorker.removeEventListener("error", this.onTerminalError);
-      tasks.push(previousWorker.shutdown());
+      previousWorker.shutdown().catch((e) => {
```
ubik2, line 296: may want to await this shutdown

PR #1188, opened 2025-05-28
Chore: Refactor away singleton pattern in runner 

seefeldb:
This is a significant refactoring aimed at removing singletons and introducing a centralized `Runtime` object. This change enhances modularity, testability, and potentially concurrency by managing services like storage, scheduling, and recipe/module management through a dedicated runtime instance.

Here's a detailed summary of the PR:

**Core Change: Introduction of `Runtime` and Removal of Singletons**

The most fundamental change is the introduction of a `Runtime` class. This class now encapsulates services that were previously accessed via global singletons (e.g., `storage`, `recipeManager`, `moduleMap`, `blobbyServerUrl`, `runtime` for script execution).

* **Instantiation & Dependency Injection:**
    * Instead of global objects, a `Runtime` instance is created and often passed into constructors of classes like `CharmManager`, `BackgroundCharmService`, or used directly where global singletons were previously called.
    * Configuration options like `storageUrl`, `blobbyServerUrl`, `signer`, `consoleHandler`, and `errorHandlers` are now passed into the `Runtime` constructor.

* **Service Access:**
    * Access to storage operations now typically happens via `runtime.storage` instead of the global `storage` object.
    * Recipe management is done through `runtime.recipeManager` instead of a global `recipeManager`.
    * Module registration and access are handled by `runtime.moduleRegistry`.
    * Document creation and access (e.g., `getDoc`, `getCell`, `getDocByEntityId`) are often now methods on `runtime` or `runtime.documentMap`.
    * The global `runtime.runSingle` (for script execution) is replaced by `runtime.harness.runSingle`.

* **Blobby Server URL:**
    * `setBlobbyServerUrl` is removed. The URL is now configured via `RuntimeOptions` when creating a `Runtime` instance.
    * `getBlobbyServerUrl` is also removed, with the URL presumably accessed via the `runtime` instance where needed (e.g., `runtime.blobbyServerUrl`).

**Impact on Packages:**

* **`@commontools/runner`:**
    * The global `storage` object is removed. Operations are now on `runtime.storage`.
    * `setBlobbyServerUrl` and `getBlobbyServerUrl` are removed.
    * Global `runtime` (for script execution) is replaced by `runtime.harness`.
    * `getCell`, `getDoc`, `getCellFromEntityId`, `getCellFromLink`, `getImmutableCell` are refactored to be methods on `Runtime` or `runtime.documentMap`.
    * Scheduler functions (`idle`, `onConsole`, `onError`, `schedule`, `unschedule`, `run` (for actions), `addEventHandler`, `queueEvent`) are now methods on `runtime.scheduler`.
    * Built-in modules (`fetchData`, `llm`, `map`, `ifElse`, `streamData`) are now registered with the `runtime.moduleRegistry` and receive the `runtime` instance, allowing them to use runtime services like `runtime.documentMap.getDoc` and `runtime.idle`.

* **`@commontools/charm`:**
    * `CharmManager` constructor now accepts a `Runtime` instance.
    * It uses `runtime.storage`, `runtime.recipeManager`, and `runtime.runner` (for `runSynced`, `idle`).
    * `getIframeRecipe` now takes `charmManager` to access `runtime.recipeManager`.
    * `saveSpell` also takes `charmManager` for similar reasons.

* **`@commontools/builder`:**
    * The global `createCell` function has been removed from direct export and is now exposed via `declare global`. The actual implementation is expected to be injected by the runtime environment (e.g., `createCellFactory` in the runner's harness).
    * Tests are updated to use `runtime.documentMap.getDoc` or `runtime.getImmutableCell`.

* **`@commontools/background-charm-service`:**
    * `cast-admin.ts` and `main.ts` instantiate and use `Runtime` for storage, blobby URL, and signer configuration.
    * `BackgroundCharmService` constructor and methods now take/use `runtime` instead of a `storage` instance.
    * `worker.ts` also initializes and uses a `Runtime` instance for its operations, including passing `consoleHandler` and `errorHandler`. Error handling logic (`isErrorWithContext`) is slightly adjusted due to how errors might be propagated.

* **`@commontools/cli`:**
    * `cast-recipe.ts`, `charm_demo.ts`, and `main.ts` are updated to use the new `Runtime` for setting up storage, signer, and blobby URL, and for getting cells/docs.

* **`@commontools/jumble` (UI Package):**
    * A new `RuntimeProvider` and `useRuntime` hook are introduced to provide the `Runtime` instance via React context.
    * Components like `NetworkInspector` and `User` now use `useRuntime` to get the runtime ID for `useStorageBroadcast`.
    * `CharmsManagerProvider` now uses `useRuntime` to pass the runtime to the `CharmManager` constructor.
    * `setupIframe` now accepts a `runtime` argument to use its scheduler and idle methods.
    * The main application entry point (`main.tsx`) initializes the `Runtime` and wraps the application with `RuntimeProvider`.
    * Hooks like `useSuggestions` and views like `SpellbookLaunchView` now get `charmManager` (which contains the runtime) to access `runtime.recipeManager`.

* **`@commontools/toolshed` (Server Package):**
    * The server's entry point (`index.ts`) now initializes a global `Runtime` instance instead of just `storage`, configuring it with the signer and memory URL.
    * Google OAuth handlers now use the global `runtime` instance instead of the global `storage`. The helper `getAuthCellAndStorage` is refactored to `getAuthCell` and uses the global `runtime`.

**Other Notable Changes:**

* **Error Handling:** Some adjustments in `worker.ts` for how error context (like `space`, `charmId`) is checked, as errors might be instances of `Error` directly rather than a specific custom error type with those properties always present.
* **Recipe ID Generation in CharmManager:** Slight modification in `charmId` function in `CharmManager` to ensure the extracted ID is a string.
* **Test Updates:** Numerous test files are updated to instantiate `Runtime` or use it for creating cells/docs and accessing services, reflecting the core refactoring.

**Summary:**

This PR introduces a major architectural shift by deprecating global singleton access for core services (`storage`, `recipeManager`, script `runtime`, etc.) and consolidating them under a new `Runtime` class. Instances of `Runtime` are created with necessary configurations (like API URLs and signers) and then injected into various managers and services (e.g., `CharmManager`, `BackgroundCharmService`). This refactoring significantly improves the codebase's structure by promoting dependency injection, which generally leads to more modular, testable, and maintainable code. It also provides a clear context for runtime operations, potentially simplifying future extensions and concurrent execution scenarios. The changes touch nearly every package, indicating a broad and fundamental improvement to the system's architecture.


packages/background-charm-service/src/worker.ts
```diff
@@ -102,8 +91,16 @@ async function initialize(
...
-  manager = new CharmManager(currentSession);
+  // Initialize runtime and charm manager
+  runtime = new Runtime({
+    storageUrl: `volatile://${currentSession.space}`,
+    blobbyServerUrl: toolshedUrl,
+    signer: identity,
+    recipeEnvironment: JSON.stringify({ apiUrl }),
+    consoleHandler: consoleHandler,
+    errorHandlers: [errorHandler]
+  });
```
jsantell, line 102: This is looking great :partying_face:

packages/charm/src/spellbook.ts
```diff
@@ -1,6 +1,14 @@
-import { recipeManager } from "@commontools/runner";
 import { UI } from "@commontools/builder";
 
+// Forward declaration to avoid circular import
```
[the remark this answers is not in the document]
seefeldb: Yup & this went away when I moved to injecting runtime instead of charmmanager

packages/runner/src/builtins/if-else.ts
```diff
@@ -1,15 +1,21 @@
...
 import { type ReactivityLog } from "../scheduler.ts";
+import { type IRuntime } from "../runtime.ts";
 
 export function ifElse(
   inputsDoc: DocImpl<[any, any, any]>,
   sendResult: (result: any) => void,
   _addCancel: (cancel: () => void) => void,
   cause: DocImpl<any>[],
   parentDoc: DocImpl<any>,
+  runtime: IRuntime, // Runtime will be injected by the registration function
```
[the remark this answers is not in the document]
seefeldb: Ah, in a way, the current way to add `raw` implementations is kind of a way to do that. In the recipe you see just the ops, and then instantiateRawNode looks into the registry of these built-ins to call the right one.

The original idea was that the environment could inject extra ones (specifically llm, fetchData, streamData feel different than ifElse and map). We might still want to do that. First though there is a lot to dedupe between them.

packages/background-charm-service/src/main.ts
```diff
@@ -24,10 +24,15 @@ const workerTimeoutMs = (() => {
 })();
 
 const identity = await getIdentity(env.IDENTITY, env.OPERATOR_PASS);
+const runtime = new Runtime({
+  storageUrl: env.TOOLSHED_API_URL,
+  blobbyServerUrl: env.TOOLSHED_API_URL,
```
ubik2, line 29: you can leave the blobbyServerUrl out here, since the constructor will set it to storageUrl by default.

packages/background-charm-service/src/worker.ts
```diff
@@ -102,8 +91,16 @@ async function initialize(
...
   });
 
-  // Initialize charm manager
-  manager = new CharmManager(currentSession);
+  // Initialize runtime and charm manager
+  runtime = new Runtime({
+    storageUrl: toolshedUrl,
+    blobbyServerUrl: toolshedUrl,
+    signer: identity,
+    recipeEnvironment: { apiUrl },
```
ubik2, line 99: blobbyServerUrl can be left out here too
I show recipeEnvironment as a string type (not an object with an apiUrl).

packages/toolshed/routes/integrations/google-oauth/google-oauth.utils.ts
ubik2, line 1: it would be nice if these methods were passed the runtime as a parameter (less singletony), but this is a fairly isolated system, so it's not as important.

packages/runner/test/schema.test.ts
```diff
@@ -1,20 +1,25 @@
...
+  });
+
+  afterEach(() => {
+    return runtime.dispose();
+  });
+
   describe("Examples", () => {
     it("allows mapping of fields via interim cells", () => {
-      const c = getDoc(
+      const c = runtime.documentMap.runtime.documentMap.getDoc(
```
ubik2, line 22: doubled `runtime.documentMap` here and on line 37

packages/charm/src/iterate.ts
```diff
@@ -102,7 +96,7 @@ export async function iterate(
 ): Promise<{ cell: Cell<Charm>; llmRequestId?: string }> {
   const optionsWithDefaults = applyDefaults(options);
   const { model, cache, space, generationId } = optionsWithDefaults;
-  const { iframe } = getIframeRecipe(charm);
+  const { iframe } = getIframeRecipe(charm, charmManager.runtime);
```
ubik2, line 99: This seems to have changed from passing Runtime to passing CharmManager.
FWIW, I prefer passing Runtime.

packages/charm/src/workflow.ts
```diff
@@ -166,7 +166,7 @@ export async function classifyIntent(
   let existingCode: string | undefined;
 
   if (form.input.existingCharm) {
-    const { spec, schema, code } = extractContext(form.input.existingCharm);
+    const { spec, schema, code } = extractContext(form.input.existingCharm, form.meta.charmManager);
```
ubik2, line 169: we could pass runtime as the parameter here instead, since nothing else is pulled from charmManager.

packages/cli/cast-recipe.ts
```diff
@@ -66,8 +58,14 @@ async function castRecipe() {
     });
 
     // Create charm manager for the specified space
-    const charmManager = new CharmManager(session);
-    const recipe = await compileRecipe(recipeSrc, "recipe", charmManager);
+    runtime = new Runtime({
+      storageUrl: toolshedUrl,
+      blobbyServerUrl: toolshedUrl,
```
ubik2, line 63: nit: as before, i'd prefer to not pass blobbyServerUrl if the default works.

packages/cli/main.ts
```diff
@@ -96,7 +85,13 @@ async function main() {
   }) satisfies Session;
 
   // TODO(seefeld): It only wants the space, so maybe we simplify the above and just space the space did?
-  const charmManager = new CharmManager(session);
+  const runtime = new Runtime({
+    storageUrl: toolshedUrl,
+    blobbyServerUrl: toolshedUrl,
```
ubik2, line 90: same nit about blobbyServerUrl

packages/jumble/src/main.tsx
```diff
@@ -84,41 +85,39 @@ const ReplicaRedirect = () => {
...
     <ErrorBoundary fallback={<div>An error has occurred</div>}>
-      <AuthenticationProvider>
-        <CharmsProvider>
-          <ActionManagerProvider>
-            <ActivityProvider>
+      <RuntimeProvider runtime={runtime}>
+        <AuthenticationProvider>
+          <CharmsProvider>
+            <ActionManagerProvider>
+              <ActivityProvider>
```
ubik2, line 120: nit, but deno fmt will indent the Router tag another level

packages/runner/src/runtime.ts
```diff
@@ -0,0 +1,483 @@
...
+  version?: string;
+  [key: string]: any;
+}
+
+export interface RuntimeOptions {
+  storageUrl: string;
+  signer?: Signer;
+  consoleHandler?: ConsoleHandler;
+  errorHandlers?: ErrorHandler[];
+  blobbyServerUrl?: string;
```
ubik2, line 48: Minor issue here that blobbyServerUrl is really blobbyServerBaseUrl, since we append the "/api/storage/blobby" to the option to get our actual Runtime.blobbyServerUrl.

packages/runner/src/runtime.ts
```diff
@@ -0,0 +1,483 @@
...
+export interface IRuntime {
+  readonly id: string;
+  readonly scheduler: IScheduler;
+  readonly storage: IStorage;
+  readonly recipeManager: IRecipeManager;
+  readonly moduleRegistry: IModuleRegistry;
+  readonly documentMap: IDocumentMap;
+  readonly harness: Harness;
+  readonly runner: IRunner;
+  readonly blobbyServerUrl: string | undefined;
```
ubik2, line 62: i don't think this can be undefined (though a different implementation of Runtime could)

packages/runner/src/runtime.ts
```diff
@@ -0,0 +1,483 @@
...
+export class Runtime implements IRuntime {
+  readonly id: string;
+  readonly scheduler: IScheduler;
+  readonly storage: IStorage;
+  readonly recipeManager: IRecipeManager;
+  readonly moduleRegistry: IModuleRegistry;
+  readonly documentMap: IDocumentMap;
+  readonly harness: Harness;
+  readonly runner: IRunner;
+  readonly blobbyServerUrl: string | undefined;
```
ubik2, line 259: don't think this can ever be undefined

packages/builder/src/built-in.ts
```diff
@@ -106,40 +105,17 @@ export function str(
...
-  const cause = { parent: frame.cause } as Record<string, any>;
-  if (name) cause.name = name;
-  else cause.number = frame.generatedIdCounter++;
-
-  const cell = getCell<T>(space, cause, schema);
-
-  if (value !== undefined) cell.set(value);
-
-  return cell;
+declare global {
```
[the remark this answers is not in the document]
seefeldb: there's a test that didn't compile otherwise (though i think it might have been fixed otherwise by now). i'll revise this soon anyway, so leaving for now.

PR #1189, opened 2025-05-28
fix: Lax the WorkerIPCRequest.Cleanup guard to allow the cleanup message to be processed

jsantell:


ubik2 (commented):
looks good, though regarding the test, data is undefined rather than `{}`.
may as well consider it valid regardless, though

PR #1190, opened 2025-05-28
chore: Add duration logs to bg worker tasks, increase the default timeout, and allow runtime configuration

jsantell:


packages/background-charm-service/src/main.ts
```diff
@@ -1,13 +1,30 @@
...
 import { env } from "./env.ts";
 
+const { timeout } = parseArgs(Deno.args, {
+  string: [
+    "timeout",
+  ],
+});
+const workerTimeoutMs = (() => {
+  if (timeout) {
+    const parsed = parseInt(timeout, 10);
```
ubik2, line 14: I think the second argument to parseInt is optional and may be more confusing here than helpful (given that the default is 10 minutes).

jsantell: Updating with a constant for the default 10 minutes

ubik2 (approved):

packages/background-charm-service/src/worker-controller.ts
```diff
@@ -188,4 +209,18 @@ export class WorkerController extends EventTarget {
 
     this.dispatchEvent(new WorkerControllerErrorEvent(err));
   };
+
+  private logTaskResults(task: Task, error?: string) {
+    const errorMessage = error ? `: ${error}` : "";
+    const state = error ? "completed" : "failed";
```
ubik2, line 215: oops: logic reversed here

PR #1191, opened 2025-05-28
Added unit test for a reference cycle

ubik2:
I saw lots of Cycle Detected messages in the toolshed-staging logs, so suspected that we might not be handling cycles correctly.
However, this unit test logs the cycle once and moves on.


ubik2:
Merging this sans review, since it's just a unit test.

PR #1199, opened 2025-05-30
Fix accidental object reuse

seefeldb:
# Fix: Deep copy initial values in recipe runner

## Problem
When running multiple instances of the same recipe, they were sharing the same object reference for initial values. This caused unintended state sharing between recipe instances, where modifying state in one instance would affect other instances.

## Solution
Added `deepCopy` when applying initial values in the recipe runner to ensure each recipe instance gets its own copy of the initial state. This prevents unintended state sharing between recipe instances.

## Changes
- Modified `runner.ts` to use `deepCopy` when applying initial values from recipe
- Added test case to verify each recipe instance gets its own copy of initial values

## Testing
Added a test case that:
1. Creates two instances of the same recipe with nested initial values
2. Verifies the internal state objects are different references
3. Verifies modifying state in one instance doesn't affect the other instance

## Impact
This fix ensures proper isolation between recipe instances, preventing unintended state sharing that could lead to bugs in applications using multiple instances of the same recipe.
    

---

## Summary by cubic
Fixed a bug where multiple recipe instances shared the same initial state object, causing unintended state sharing.

- **Bug Fixes**
  - Used deep copy for initial values in the recipe runner to ensure each instance has its own state.
  - Added a test to confirm state isolation between recipe instances.

### DIFF
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -184,8 +184,8 @@ export class Runner implements IRunner {
     const defaults = extractDefaultValues(recipe.argumentSchema);
 
     const internal = {
-      ...(deepCopy(defaults) as { internal: any })?.internal,
-      ...(recipe.initial as { internal: any } | void)?.internal,
+      ...deepCopy((defaults as { internal: any })?.internal),
+      ...deepCopy((recipe.initial as { internal: any })?.internal),
       ...processCell.get()?.internal,
     };
 
--- a/packages/runner/test/runner.test.ts
+++ b/packages/runner/test/runner.test.ts
@@ -581,4 +581,79 @@ describe("runRecipe", () => {
     expect(resultCell.get()?.name).toEqual("my counter");
     expect(resultCell.getAsQueryResult()?.counter).toEqual(2);
   });
+
+  it("should create separate copies of initial values for each recipe instance", async () => {
+    const recipe: Recipe = {
+      argumentSchema: {
+        type: "object",
+        properties: {
+          input: { type: "number" },
+        },
+      },
+      initial: {
+        internal: {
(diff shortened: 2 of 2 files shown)


packages/runner/src/runner.ts
```diff
@@ -184,8 +184,8 @@ export class Runner implements IRunner {
     const defaults = extractDefaultValues(recipe.argumentSchema);
 
     const internal = {
-      ...(deepCopy(defaults) as { internal: any })?.internal,
-      ...(recipe.initial as { internal: any } | void)?.internal,
+      ...deepCopy((defaults as { internal: any })?.internal),
+      ...deepCopy((recipe.initial as { internal: any })?.internal),
```
ubik2, line 188: we only really need to deep copy recipe.initial.internal, but i guess we do the same thing for defaults

ubik2 (approved):
lgtm!

packages/runner/src/runner.ts
seefeldb: yeah, that's why i moved deepCopy out.

(and ...undefined works in JS, despite CoPilot's suggestion here)

packages/runner/src/runner.ts
```diff
@@ -184,8 +184,8 @@ export class Runner implements IRunner {
     const defaults = extractDefaultValues(recipe.argumentSchema);
 
     const internal = {
```
ubik2, line 186: maybe add a comment right above this about how we need to have our own object

packages/runner/src/runner.ts
ubik2: i don't know how i missed that, but glad to see.

PR #1200, opened 2025-06-02
Use client schema queries to avoid redundant loads from server

ubik2:
When we attempt to load a doc with a schema on the client, we may already have the doc available locally, but we don't know that we've satisfied all the dependencies, and we don't know that we'll be subscribed to the appropriate docs.
Previously, we would issue an additional querySchema for those docs.

With this change, we do some more bookkeeping, so we know what the SchemaPathSelector would have been for each doc returned by the query.

This doesn't remove any of the queries yet (the server needs the queries to track our subscription), but it provides the information we'll need to make those decisions.

There was some module shuffling required.

- Moved cfc.ts and traverse.ts into the builder module. Put traverse in its own submodule.
- Moved SchemaContext into builder/types
- Moved FactAddress into memory/interface
- Add cfc as a field in IRuntime and use this when it's accessible (in doc, cell, and storage)
- Cleanup of some imports, changing some to import type and rearranging others.
- Change schemaRef to use the refer of the JSON.stringified schema instead of the actual schema, since that's less compute intensive
- Added a MapSet class for tracking the schemas used, as well as a SelectorTracker
- Changed traverse's import of types, since for some reason I couldn't import the module way.
- Changed iterateSelector to take a default value so I can have stricter types
- Added a selectFact method to just return one fact.
- Increased the sync timer to 60 seconds instead of 1 second. It's possible this can be removed completely, but right now, it can interfere with debugging.
- Increased cache sync timer to 1 minute, so it's not generally interfering with our load time testing
- Removed ComsoleMethod imports, since we aren't using this
- Added query.ts which can do client side schema queries backed by the cache's heap, together with tests
- Added an Immutable type, since I had altered a JSONValue that should have been constant, and wanted the compiler to catch these.


ubik2:
> With this change, we run the schema query locally to determine whether we already have all the docs we'd reach in the heap. If that's the case, we're already subscribed and already have fresh data, so go ahead and get the data from heap.
1) Alice runs a query including doc1 and doc2. it does not look at the doc2.address, but does look at doc2.name
2) Alice runs a query including doc2 that looks at the doc2.address, but since that's empty, we don't need any docs, so we don't send to server
3) Bob updates doc2 to have an address that points to doc3.
Server re-runs Alice's query1, which does not care about doc3

PR #1203, opened 2025-06-02
`deno fmt` run

seefeldb:
## Summary by cubic
Ran `deno fmt` to apply consistent code formatting across the codebase. This improves code readability and enforces style standards.


ubik2 (commented):
I think you'll want to add something like the following to packages/static/deno.json
```
  "fmt": {
    "exclude": ["./assets/*"]
  },
```

PR #1209, opened 2025-06-03
chore: Update to deno 2.3.5. Reuse CI action.

jsantell:
Some relevant features (via @ubik2):

* With Deno 2.3.x: The new lockfile format includes extra npm package metadata to speed up dependency resolution and installation."
* V8 Insights: https://deno.com/blog/v2.3#v8-js-engine-metrics

### DIFF
--- a/.github/actions/deno-setup/action.yml
+++ b/.github/actions/deno-setup/action.yml
@@ -1,0 +1,23 @@
+name: "Deno Setup"
+description: "Setup deno"
+inputs:
+  cache:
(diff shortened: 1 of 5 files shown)


ubik2 (approved):

jsantell:
> We'll have a bunch of deno.lock file updates as well, but those can be committed as they're rewritten.

Bumped the lock file, let's see what happens

jsantell:
Added a check to the workspace's `deno task check` for the deno version used in our cluster

PR #1228, opened 2025-06-11
Introducing Common UI v2 - a bare bones webcomponent ui library inspired by shadcn/ui

jakedahn:
## Summary by cubic
Introduced Common UI v2, a new set of 39 secure, shadcn/ui-inspired web components for Deno, alongside legacy v1 components, with full documentation and test pages.

- **New Features**
  - Added v2 component library with modern, accessible web components (e.g., ct-button, ct-card, ct-form, ct-alert).
  - Included detailed usage docs and HTML test files for each v2 component.
  - Updated exports to support both v1 and v2 APIs.


packages/ui/src/v2/core/base-element.ts
```diff
@@ -0,0 +1,145 @@
+/**
+ * Base class for all web components
+ * Provides common functionality and lifecycle management
+ */
+export abstract class BaseElement extends HTMLElement {
```
jsantell, line 5: Any reason for this base element over Lit's [base element](https://lit.dev/docs/api/ReactiveElement/)? Lit would accomplish a few things missing here:

* Abstracts away the differences between properties (regular JS properties like `element.x`) and attributes (`<x-foo value="5"></x-foo>`, retrieved via `element.getAttribute("value")`) -- the big difference being `BaseElement` here only operates on attributes, and writes to the DOM on every attribute change which does a number on performance. Typically the flow for lit is attributes are on read on instantiation (`<x-foo value="5"></x-foo>`) and from there on out is stored as a *typed property* e.g. `element.value === 5`. This is the difference in the `.=` binding in lit's templates.
  * aria attributes must and should continue being attributes
* Attributes are always strings! We'll need to handle non-string values eventually outside of the UI components (e.g. passing in `Cell` or bytes or a regular object). Using (lit) properties gives us types
* Wouldn't need boilerplate get/set

Lit would remove the get/set boilerplates, handle non-stringified values (we pass in `Cell` as a property in recipes, not an issue for UI-ish things, but we'll need non-string values eventually), and

packages/ui/src/v2/utils/registry.ts
```diff
@@ -0,0 +1,84 @@
+/**
+ * Component registry for managing web component registration
+ */
+
+import type { ComponentOptions, CustomElementConstructor } from "./types.ts";
+
+class ComponentRegistry {
```
jsantell, line 7: Is this necessary? `globalThis.customElements` is our global registry

packages/ui/src/v2/register-all.ts
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// Auto-register when this module is imported
+registerAllComponents();
+
+// Export for programmatic access
+export { components };
+
+// Re-export all components for convenience
+export * from "./components/ct-button/index.ts";
```
jsantell, line 122: I don't think this is necessary here

packages/ui/src/v2/register-all.ts
```diff
@@ -0,0 +1,168 @@
...
+  { tagName: "ct-vgroup", constructor: CTVGroup },
+  { tagName: "ct-hscroll", constructor: CTHScroll },
+  { tagName: "ct-vscroll", constructor: CTVScroll },
+  { tagName: "ct-grid", constructor: CTGrid },
+  { tagName: "ct-table", constructor: CTTable },
+] as const;
+
+// Register all components
+export function registerAllComponents(): void {
+  components.forEach(({ tagName, constructor }) => {
```
jsantell, line 110: `components.forEach(({tagName,constructor}) => customElements.define(tagName, constructor));`

packages/ui/src/v2/register-all.ts
```diff
@@ -0,0 +1,168 @@
...
+ *
+ * This file imports and registers all Common UI web components.
+ * Import this file to automatically register all components with their custom element names.
+ *
+ * Usage:
+ * ```typescript
+ * import '@commontools/ui/v2/register-all';
+ * ```
+ *
+ * Or in HTML:
```
jsantell, line 12: html example doesn't make sense in our workspace

packages/ui/src/v2/styles/tokens.ts
```diff
@@ -0,0 +1,143 @@
+/**
+ * Design tokens for consistent styling across components
+ */
+
+export const tokens = {
```
jsantell, line 5: Could these be css variables?

packages/ui/src/v2/types/components.ts
```diff
@@ -0,0 +1,126 @@
+/**
+ * Component Type Definitions
+ *
+ * This file exports all component interfaces and types
+ * for better IDE support and type safety.
+ */
+
```
jsantell, line 7: Are these types used? I'd expect e.g `CTButton` types to live in `ct-button/ct-button.ts`, but not sure if events/types here are referenced

packages/ui/LLM-COMPONENT-INSTRUCTIONS.md
```diff
@@ -0,0 +1,674 @@
...
+<ct-card>
+  <h3 slot="header">Card Title</h3>
+  <p slot="content">Card content goes here</p>
+  <ct-button slot="footer">Action</ct-button>
+</ct-card>
+```
+
+### 13. ct-badge
+
+**Purpose**: Status indicator or label **Tag**: `<ct-badge>` **Attributes**:
```
jsantell, line 218: nit: this formatting seems like it's missing newlines, expecting:

```
**Purpose**: Status indicator or label
**Tag**: `<ct-badge>`
**Attributes**:
- `variant` - "default" | "secondary" | "destructive" | "outline"
**Events**:
- `ct-remove` - Fired when..
```

packages/ui/src/v2/utils/registry.ts
jsantell: In fact, some component `index.ts` files just register on `customElements` and some use this registry.

Unless we need another registry, components exposing their class from `foo.ts`, and then the class plus registration to `customElements` in `index.ts` seems like a good pattern (like `ct-accordion`)

packages/ui/src/v2/core/base-element.ts
jsantell: This is the biggest thing to address IMO

deno.lock
```diff
@@ -1,55 +1,58 @@
 {
   "version": "5",
   "specifiers": {
-    "jsr:@astral/astral@*": "0.5.3",
```
jsantell, line 4: Lots of version changes here, if running into issues try reverting this lock

packages/ui/src/v2/utils/registry.ts
jakedahn: good catch! will fix these up

PR #1236, opened 2025-06-13
Move builder/ into runner/ + Moving and splitting files while we do that

seefeldb:
## Summary by cubic
Moved all code from the builder package into runner, updated imports, and split builder utilities into smaller files for better organization.

- **Refactors**
  - Removed the builder package and merged its code into runner.
  - Updated all imports across the codebase to use runner instead of builder.
  - Split builder utilities (like utils.ts) into focused files (e.g., path-utils, node-utils, traverse-utils) within runner.


jsantell (approved):
LGTM!

PR #1238, opened 2025-06-13
Replace docimpl with cell - added Cell.getRaw() and setRaw and updated charm manager

ellyxir:
added Cell.getRaw() and setRaw and updated charm manager
need to update other references but doing a check-in here
    

---

## Summary by cubic
Added getRaw() and setRaw() methods to Cell for direct access to raw document data, and updated CharmManager to use Cell instead of DocImpl.

- **New Features**
  - Cell.getRaw() returns the raw value without alias resolution.
  - Cell.setRaw() writes directly to the cell, bypassing validation and alias logic.

- **Refactors**
  - CharmManager now uses Cell methods instead of DocImpl, simplifying reference handling.


packages/charm/src/manager.ts
```diff
@@ -552,7 +551,7 @@ export class CharmManager {
                 if (cellId) addMatchingCharm(cellId);
 
                 const sourceRefId = followSourceToResultRef(
-                  cellLink.cell,
+                  cellLink.cell.asCell(),
```
ubik2, line 554: cellLink.cell should probably be a Cell instead of DocImpl, so we shouldn't need the asCell here.

packages/charm/src/manager.ts
```diff
@@ -583,7 +582,7 @@ export class CharmManager {
               if (cellId) addMatchingCharm(cellId);
 
               const sourceRefId = followSourceToResultRef(
-                value.cell,
+                value.cell.asCell(),
```
ubik2, line 585: same as above

packages/charm/src/manager.ts
ellyxir: is this specifically in this case? in general CellLink.cell is a DocImpl according to its type definition. 
i added a test and it failed:
```ts
  it("is cellLink.cell a cell", () => {
    const c = runtime.documentMap.getDoc(
      { value: 42 },
      "celllink-test",
      "test",
    );
    const cell = c.asCell();
    const cellLink = cell.getAsCellLink();
    expect(isCell(cellLink.cell)).toBe(true);
  });
```

seefeldb: But what you could do here already is use `runtime.getCellFromLink()` with the full cell link. That way that future change is transparent and this will already work for cross-space links as well.

ellyxir:
@ubik2 @seefeldb the last commit eb72644, i switch from runtime.runSynced() instead of runtime.run() since it can take a cell. also added overload for stop, which i think is fine. let me know if that might cause issues. i'd like to land this into main even though theres more to do.

packages/runner/src/runner.ts
```diff
@@ -345,11 +345,14 @@ export class Runner implements IRunner {
...
    * effects driving execution, e.g. the UI.
    *
-   * @param resultCell - The result doc to stop.
+   * @param resultCell - The result doc or cell to stop.
    */
-  stop<T>(resultCell: DocImpl<T>): void {
-    this.cancels.get(resultCell)?.();
-    this.cancels.delete(resultCell);
+  stop<T>(resultCell: DocImpl<T>): void;
+  stop<T>(resultCell: Cell<T>): void;
```
seefeldb, line 351: you might not need the overload here, just have one
```
stop<T>(resultCell: DocImpl<T> | Cell<T>)
```
should be enough. we use that in other places already, e.g. storage.syncCell

packages/runner/test/cell.test.ts
```diff
@@ -22,63 +22,45 @@ describe("Cell", () => {
   });
 
   it("should create a cell with initial value", () => {
-    const c = runtime.documentMap.getDoc(
-      10,
-      "should create a cell with initial value",
-      "test",
-    );
+    const c = runtime.getCell<number>("test", "should create a cell with initial value");
+    c.setRaw(10);
```
seefeldb, line 26: in this case `c.set` would work as well, let's prefer `setRaw` when we intent to do something that requires skipping that layer.

seefeldb:
> @ubik2 @seefeldb the last commit [eb72644](https://github.com/commontoolsinc/labs/commit/eb72644e741b3fb9409a9dd505db7a38de0554e5), i switch from runtime.runSynced() instead of runtime.run() since it can take a cell. also added overload for stop, which i think is fine. let me know if that might cause issues. i'd like to land this into main even though theres more to do.

it works, but `run` and `runSynced` do have different semantics, though it ends up being the same with VolatileStorage. This could change subtly when we switch over to the memory-based test store.

Either way we should switch `run` over to cells soon. Maybe that needs exposing setSourceCell() on `Cell`, but otherwise it should map quite directly to getRaw and setRaw.

packages/runner/src/runner.ts
ellyxir: done thanks

PR #1239, opened 2025-06-13
Committing wip to change docIsSyncing to take into account the schema.

ubik2:
This doesn't work because of a dependency cycle, but I wanted to commit it.
Since it looks like the dependency cycle may be resolved, I'm making this a draft PR, so I remember to maybe land it later.
    

---

## Summary by cubic
Updated docIsSyncing to track syncing state by both document and schema, improving accuracy when loading documents with different schemas.

- **Refactors**
  - Replaced Set with SelectorTracker to handle schema-aware syncing.
  - Moved SelectorTracker to a shared location for reuse.


ubik2:
Implemented this functionality in the schema-crawler-removal branch

PR #1245, opened 2025-06-16
fix: remove hasSigner / setSigner methods

Gozala:
## Summary by cubic
Removed the hasSigner and setSigner methods from storage interfaces and related checks to simplify storage handling. This cleans up unused code and removes unnecessary signer checks.


anotherjesse:
just to confirm - this is "good" because we can't have a runtime without a storage provider now?

we were using this as a way to ask "is storage setup"

PR #1251, opened 2025-06-16
Convert cell links to sigil values

seefeldb:


ellyxir:
noticed sigil.md says field should be "source" but we seem to use "id". 
the default for overwrite should be "this" but in parseLink, i think you end up with undefined? maybe it should just turn to "this"?
is "accept" something we'll do later?

anotherjesse:
definitely having a regression in (at least) "new charm"

image

packages/runner/test/type-utils.test.ts
```diff
@@ -35,9 +32,19 @@ describe("value type", () => {
...
-  it("isAlias correctly identifies aliases", () => {
-    expect(isAlias({ $alias: { path: ["path", "to", "value"] } })).toBe(true);
-    expect(isAlias({ notAlias: "something" })).toBe(false);
+  it("isWriteRedirectLink correctly identifies write redirects", () => {
+    expect(isWriteRedirectLink({ $alias: { path: ["path", "to", "value"] } }))
+      .toBe(true);
+    expect(
+      isWriteRedirectLink({
+        "/": {
+          [LINK_V1_TAG]: { id: "path/to/value", overwrite: "redirect" },
```
ubik2, line 41: this doesn't match the sigil rfc, which would have `{ path: ["path", "to", "value"]`.

packages/runner/src/uri-utils.ts
```diff
@@ -0,0 +1,58 @@
...
+    const parsed = JSON.parse(JSON.stringify(value)) as { "/": string };
+
+    // Handle EntityId object
+    if (typeof parsed["/"] === "string") return `of:${parsed["/"]}`;
+  } else if (typeof value === "string") {
+    // Already has prefix with colon
+    if (value.includes(":")) {
+      // TODO(seefeld): Remove this once we want to support any URI, ideally
+      // once there are no bare ids anymore
+      if (!value.startsWith("of:")) {
```
ubik2, line 20: We never have links that point to did: entities, but we do have this in a lot of 'of' fields.
I'm guessing this won't be called on those types, but didn't verify.

The normalizeEntityId method is really nice. I made my own entityIdStr method to extract it, because the EntityId -> string path is annoyingly complex to have littered all over the code.
After this lands, I'll probably replace mine with this one.

packages/runner/src/link-utils.ts
```diff
@@ -0,0 +1,301 @@
...
+
+  // If either is null/undefined, they're only equal if both are
+  if (!value1 || !value2) return value1 === value2;
+
+  // Try parsing both as links
+  const link1 = parseLink(value1, base);
+  const link2 = parseLink(value2, base);
+
+  // If one parses and the other doesn't, they're not equal
+  if (!link1 || !link2) return false;
```
ubik2, line 258: they may both have failed to parse and still be equal, but i think it's fine to pretend they aren't in that case.

packages/runner/src/cell.ts
```diff
@@ -439,27 +485,64 @@ function createRegularCell<T>(
       subscribeToReferencedDocs(callback, doc, path, schema, rootSchema),
     getAsQueryResult: (subPath: PropertyKey[] = [], newLog?: ReactivityLog) =>
       createQueryResultProxy(doc, [...path, ...subPath], newLog ?? log),
-    getAsCellLink: () => // Add space here, so that JSON.stringify() of this retains the space.
-    ({ space: doc.space, cell: doc, path, schema, rootSchema }),
+    getAsCellLink: (): CellLink => {
```
ubik2, line 488: Maybe add
` /** @deprecated Use \`getAsLink\` instead. */` so the IDE flags this too.

packages/runner/src/runtime.ts
```diff
@@ -419,38 +421,52 @@ export class Runtime implements IRuntime {
...
-      throw new Error("Cell link has no space");
     }
+
     // If we aren't passed a schema, use the one in the cellLink
-    return doc.asCell(cellLink.path, log, schema ?? cellLink.schema);
+    return doc.asCell(
+      cellLink.path,
+      log,
+      schema ?? cellLink.schema,
+      schema ? undefined : cellLink.rootSchema,
```
ubik2, line 468: This will set rootSchema to undefined if we provided a schema. Perhaps that's intentional to indicate that the rootSchema should just be the same as the schema?

packages/runner/src/traverse.ts
```diff
@@ -448,34 +449,24 @@ function narrowSchema(
...
-    return {
-      path: value.$alias.path.map((p) => p.toString()),
-      cellTarget: undefined,
-    };
-  } else if (isJSONCellLink(value)) {
-    //console.error("cell: ", obj.cell, "; path: ", obj.path);
-    return { path: value.path, cellTarget: value.cell["/"] as string };
-  }
-  return { path: [], cellTarget: undefined };
+  const link = parseLink(value, {} as NormalizedLink);
```
ubik2, line 460: I don't think you get anything out of passing the empty object as the second argument, instead of leaving it off.

packages/runner/src/traverse.ts
```diff
@@ -448,34 +449,24 @@ function narrowSchema(
...
 /**
- * Extract the path and cellTarget from an Alias or JSONCellLink
+ * Extract the path and cellTarget from an Alias, JSONCellLink, or sigil value
  *
  * @param value - The JSON object that might contain pointer information
  * @returns A CellTarget object containing:
  *   - path: An array of string segments representing the path to the target
  *   - cellTarget: The target cell identifier as a string, or undefined if it refers to the current document
  */
 export function getPointerInfo(value: Immutable<JSONObject>): CellTarget {
```
ubik2, line 459: I can probably get rid of this function now that we have a more uniform view of cells. I just needed a general way to interact with these links.
Anyhow, I'll tackle that after this lands.

ubik2 (approved):
lgtm

ubik2:
There's a slightly annoying issue where we're going to think docs have changed after going back and forth from JSON to Cell/DocImpl and back, since we may not have changed anything, but the JSON.stringify will be different because the cell links will be created in the new form.

packages/ui/src/v1/components/common-updater.ts
```diff
@@ -66,7 +66,7 @@ export class CommonUpdaterElement extends LitElement {
       throw new Error("No <common-charm> container.");
     }
     const { charmId } = container;
-    const space = this.state.getAsCellLink().space;
+    const space = this.state.space;
```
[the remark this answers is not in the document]
seefeldb: state is a `Cell` and always has a space.

packages/runner/src/link-resolution.ts
```diff
@@ -241,35 +256,29 @@ export function followLinks(
...
   log?: ReactivityLog,
-): CellLink {
-  if (isAlias(alias)) {
-    const doc = isCell(docOrCell) ? docOrCell.getDoc() : docOrCell;
+): LegacyCellLink {
+  if (isDoc(base)) base = base.asCell();
+  else base = base as Cell<T>; // Makes TS happy
+
+  if (isWriteRedirectLink(alias)) {
+    const link = parseLink(alias, base)!;
```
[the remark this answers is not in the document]
seefeldb: will remove ! instead

packages/runner/src/sigil-types.ts
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Sigil link type
+ */
+
+export type SigilLink = SigilValue<LinkV1>;
+/**
+ * Sigil alias type - uses LinkV1 with overwrite field
+ */
+
+export type SigilWriteRedirectLink = SigilValue<AliasV1>;
```
ellyxir: Is this worth it to have as a type?

packages/runner/test/runner.test.ts
```diff
@@ -320,7 +320,7 @@ describe("runRecipe", () => {
       space,
       "should allow stopping a recipe: input cell",
     );
-    inputCell.setRaw({ input: 10, output: 0 });
+    inputCell.set({ input: 10, output: 0 });
```
ellyxir, line 323: thanks for fixing these!

PR #1259, opened 2025-06-18
Ct 446 replace docimpl with cell part2

ellyxir:
removed more instances of DocImpl references, added a runner.run that takes in a cell instead of doc. didn't remove docImpl from it.
    

---

## Summary by cubic
Replaced more uses of DocImpl with Cell in the runner, runtime, and tests. Added support for runner.run to accept a Cell as the result target.

- **Refactors**
  - Updated runner and runtime to allow run methods to take a Cell instead of DocImpl.
  - Refactored tests to use getCell and Cell APIs instead of direct DocImpl usage.

### DIFF
--- a/packages/runner/test/cell.test.ts
+++ b/packages/runner/test/cell.test.ts
@@ -93,18 +93,18 @@ describe("Cell", () => {
   });
 
   it("should call updates callback when value changes", () => {
-    const c = runtime.documentMap.getDoc(
-      0,
-      "should call updates callback when value changes",
+    const cell = runtime.getCell<number>(
       space,
+      "should call updates callback when value changes",
     );
+    cell.set(0);
     const values: number[] = [];
-    const unsink = c.updates((value) => values.push(value));
-    c.send(1);
-    c.send(2);
-    c.send(3);
+    const unsink = cell.getDoc().updates((value) => values.push(value));
+    cell.send(1);
+    cell.send(2);
+    cell.send(3);
     unsink();
-    c.send(4);
+    cell.send(4);
     expect(values).toEqual([1, 2, 3]);
   });
 
@@ -195,22 +195,22 @@ describe("Cell utility functions", () => {
   });
 
   it("should identify a cell", () => {
-    const c = runtime.documentMap.getDoc(
-      10,
-      "should identify a cell",
+    const c = runtime.getCell(
       space,
+      "should identify a cell",
     );
-    expect(isDoc(c)).toBe(true);
-    expect(isDoc({})).toBe(false);
+    c.set(10);
+    expect(isCell(c)).toBe(true);
+    expect(isCell({})).toBe(false);
   });
 
   it("should identify a cell reference", () => {
-    const c = runtime.documentMap.getDoc(
-      10,
-      "should identify a cell reference",
+    const c = runtime.getCell<{ x: number }>(
       space,
+      "should identify a cell reference",
     );
-    const ref = { cell: c, path: ["x"] };
+    c.set({ x: 10 });
+    const ref = c.key("x").getAsCellLink();
     expect(isCellLink(ref)).toBe(true);
     expect(isCellLink({})).toBe(false);
   });
@@ -308,7 +308,7 @@ describe("createProxy", () => {
       "should handle cell references",
     );
     c.set({ x: 42 });
-    const ref = { cell: c.getDoc(), path: ["x"] };
+    const ref = c.key("x").getAsCellLink();
     const proxy = c.getAsQueryResult();
     proxy.y = ref;
     expect(proxy.y).toBe(42);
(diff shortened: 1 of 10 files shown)


ellyxir:
landing, if @ubik2 sees any issues, i can work on those.

packages/runner/test/recipe-binding.test.ts
```diff
@@ -32,21 +33,21 @@ describe("recipe-binding", () => {
...
   describe("sendValueToBinding", () => {
     it("should send value to a simple binding", () => {
-      const testCell = runtime.documentMap.getDoc(
-        { value: 0 },
-        "should send value to a simple binding 1",
+      const testCell = runtime.getCell<{ value: number }>(
         space,
+        "should send value to a simple binding 1",
       );
+      testCell.set({ value: 0 });
```
ubik2, line 40: It's not clear to me whether this should be a `set` or a `setRaw`. The latter is more like what happened with doc, but in this case, they should behave the same.

packages/runner/test/storage.test.ts
```diff
@@ -147,12 +146,12 @@ describe("Storage", () => {
 
   describe("ephemeral docs", () => {
     it("should not be loaded from storage", async () => {
-      const ephemeralDoc = runtime.documentMap.getDoc(
-        "transient",
-        "ephemeral",
+      const ephemeralDoc = runtime.getCell<string>(
```
ubik2, line 149: nit: rename as ephemeralCell for clarity

packages/runner/test/cell.test.ts
```diff
@@ -250,93 +246,93 @@ describe("createProxy", () => {
...
+    c.setRaw({ x: { $alias: { path: ["y"] } }, y: 42 });
     const proxy = c.getAsQueryResult();
     expect(proxy.x).toBe(42);
   });
 
   it("should handle aliases when writing", () => {
-    const c = runtime.documentMap.getDoc(
-      { x: { $alias: { path: ["y"] } }, y: 42 },
-      "should handle aliases when writing",
+    const c = runtime.getCell(
```
ubik2, line 280: you may be able to pass the type of the object here as a template variable to getCell, so you don't need to `as any` later.
also ok as is.

packages/runner/test/cell.test.ts
```diff
@@ -513,45 +509,48 @@ describe("createProxy", () => {
...
+    c.set([1, 2, 3]);
     const log: ReactivityLog = { reads: [], writes: [] };
     const proxy = c.getAsQueryResult([], log);
     proxy.length = 2;
     expect(c.get()).toEqual([1, 2]);
-    expect(log.writes).toEqual([
-      { cell: c, path: ["length"] },
-      { cell: c, path: [2] },
-    ]);
+    expect(log.writes.length).toBe(2);
```
ubik2, line 521: What prevents you from using the `.toEqual` with `c` replaced by `c.getDoc()` here?

packages/runner/test/test-helpers.ts
```diff
@@ -0,0 +1,52 @@
...
+      const normalizedActual = normalizeCellLinksInAliases(actual);
+      const normalizedExpected = normalizeCellLinksInAliases(expected);
+      
+      try {
+        assertEquals(normalizedActual, normalizedExpected, msg);
+      } catch (error) {
+        if (error instanceof AssertionError) {
+          throw new AssertionError(
+            `CellLinks are not equal (after normalization).\n${error.message}`
+          );
```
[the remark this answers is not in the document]
ellyxir: added message and rethrow so we have the stack also, commit 43226f0b

packages/runner/test/recipe-binding.test.ts
ellyxir: i'll leave it in this case, thanks!

packages/runner/test/cell.test.ts
ellyxir: probably relying on too much translation of prior work haha, here is the updated code:
```diff
 521⋮    │    expect(log.writes.length).toBe(2);
 522⋮    │    expect(log.writes[0].cell).toBe(c.getDoc());
 523⋮    │    expect(log.writes[0].path).toEqual(["length"]);
 524⋮    │    expect(log.writes[1].cell).toBe(c.getDoc());
 525⋮    │    expect(log.writes[1].path).toEqual([2]);
    ⋮ 522│    expectCellLinksEqual(log.writes).toEqual([
    ⋮ 523│      c.key("length").getAsCellLink(),
    ⋮ 524│      c.key(2).getAsCellLink(),
    ⋮ 525│    ]);
```
commit 6c11e208

packages/runner/test/cell.test.ts
```diff
@@ -640,10 +637,10 @@ describe("asCell", () => {
         eventCount++;
         lastEventSeen = event;
       },
-      { cell: c, path: ["stream"] },
+      { cell: c.getDoc(), path: ["stream"] },
     );
 
-    streamCell.send("event");
+    (streamCell as any).send("event");
```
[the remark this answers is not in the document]
ellyxir: oops , fixed commit e01aec94

PR #1260, opened 2025-06-18
CT-475: Add format and content attributes to JSONSchema

ubik2:
Added the `format` field for format annotations, as well as the content properties from section 8 of the JSONSchema specification.
    

---

## Summary by cubic
Added support for the format field and content properties in the JSONSchema type to better align with the JSON Schema specification.


ubik2:
@seefeldb tagged you on this because i wasn't familiar with the content encoding bits in the schema. This exposes the media type, and the encoding (like we do for our base64 encoded blobs in our sigils), so there's some overlap, but perhaps not in a useful enough way.

seefeldb:
> @seefeldb tagged you on this because i wasn't familiar with the content encoding bits in the schema. This exposes the media type, and the encoding (like we do for our base64 encoded blobs in our sigils), so there's some overlap, but perhaps not in a useful enough way.

Yeah, that's for sending data via string, a blob stub. External data might use this, so let's add it as well.

We haven't discussed yet how to specify a schema for blobs, but when we do, we should reuse these names as applicable.

PR #1267, opened 2025-06-19
update commontools.d.ts

anotherjesse:
## Summary by cubic
Added new fields to the JSONSchema type definition to support format, content encoding, content media type, and content schema.


anotherjesse:
@jsantell could/should we have a check that compiles api types in CI and if it changes, we throw an error?

PR #1269, opened 2025-06-19
Client-side schema queries to limit docs fetched.

ubik2:
Remove crawler code, so we don't need to fetch reachable, but not schema-reachable docs.
Use local queries to determine docs needed for a cell/schema
 
- Provide schema when syncing charm cells in CharmManager. Minor type cleanup.
- Load source links when we load docs on the server in `loadFactsForDoc`, and on the client in `querySchema`. Also do this in traverse when we `followPointer`. The former two capture the initial doc's source, while the followPointer version handles any that are referenced.
- Load recipe links in the same places
- Remove `_batchForDoc`, `_processCurrentBatch`, `_addToBatch`, `_runBatchesUntilSettled`, `_ensureIsSyncedf`, and `_ensureIsSyncedById`. Also changed doc loading to account for schema, so we can load the same doc with different schema, but a second request with the same schema will wait on the pending request. 
- Added toAddress method to BaseObjectManager. Implementations should use this to convert a string to whatever form of address they use. Also added getTarget method that gets an address object from a URI.
- Added an `entityIdStr` method to doc-map.ts, since getting the string version is tricky, and we do it in many places. This is exported from runner.
- Changed some classes used to abstract the details of interacting with how objects are stored in query.ts
  - Made ClientObjectManager abstract. Expose the readValues and missingDocs contents, as well.
  - Added StoreObjectManager, which is backed by the storage systems heap. When loading a doc, add that doc to the set of readValues, so we track the usage (this should have been done before, but wasn't). 
  - Added DocObjectManager, which is backed by the DocumentMap and the StorageProvider, so we can run schema queries against what we have locally.
  - Remove the readDependentDocs/writeDependentDocs maps from the ObjectManager, since we aren't using this, and aren't properly maintaining its contents
- Added tracking of server subscriptions, so we don't need to re-issue any current subscriptions. This can also be used if we reconnect, so that we can re-establish our list of subscriptions (Ellyse implemented the reconnect logic).
- Added tracking of pending nursery changes, and be less aggressive about purging the nursery. When we receive updates from the server, we only send them to the subscribers if the update doesn't correspond to a pending update in the nursery (otherwise, we already have more recent data in the doc). This code can go away when we have our transaction system integrated.
- Switched some traversal code to use the normalized links. Use the new isAnyCellLink and parseLink methods instead of isPointer and getPointerInfo.
- Moved MapSet from util to traverse, and added some more functionality.

For storage, we still maintain subscriptions from the doc to the storage provider and vice-versa. For docs, we still need to do a bit of conversion to JSON for the queryResultProxy, but can generally use them as-is.

I've temporarily disabled the IndexedDB cache while I work out any issues. Since we don't have this properly integrated with the schema queries yet, we don't get its benefits, and it makes it easier to rule out errors. I've also disabled the polling, since that only exists for the untracked objects in the db.

I may need to restore portions of storage related to retry and conflict handling, but for now I'm going to wait until the transaction work is done, and rely on that.

### Unresolved issues
- push-conflict.test.ts has some tests that rely on the removed retry logic. I've disabled those portions.

### Multi-client Conflicts
Because we send updates more immediately now, we will often end up with multiple calls to `/memory/transact` where one value triggers updates of multiple computed values. This will manifest as a conflict when multiple users are using the same charm, since Alice's update of value1 is followed by recipe triggered updates of value2 and value3. When Bob receives value1 from Alice, his copy of the recipe also computes updates of value2 and value3. When Bob attempts to write values2 and value3, his values conflict with the writes by Alice.

### Multi-cell Transactions
There's no real guarantee for the doc updates to be transactional based on commits from the server (though I believe they currently are), but there is an expectation that the other way is transactional. If we update a set of docs, those updates should be grouped as a single transaction. We could use our server subscriptions to find which schemas we have applied to the doc being updated, and follow that to get linked docs, but it's much simpler to go back to the previous approach, which is to follow all resolvable links and get a collection of doc updates. These can be written together. This may all go away when we implement the transaction api.

---

## Summary by cubic
Replaced the crawler with client-side schema queries to limit which docs are fetched, so only schema-reachable docs are loaded for each cell.

- **Refactors**
  - Removed crawler code and related batch logic.
  - Added local schema query logic to determine needed docs.
  - Updated storage and doc syncing to use schema-based queries.
  - Improved doc subscription and syncing to avoid fetching unnecessary docs.


packages/charm/src/manager.ts
```diff
@@ -340,7 +340,10 @@ export class CharmManager {
     if (isCell(id)) charm = id;
     else charm = this.runtime.getCellFromEntityId(this.space, { "/": id });
 
-    await this.runtime.storage.syncCell(charm);
+    await this.runtime.storage.syncCell(charm, false, {
```
seefeldb, line 343: we could write this as `charm.asSchema(charmSchema)` (or even narrowSchema once we support that)

packages/runner/src/storage.ts
```diff
@@ -182,97 +151,195 @@ export class Storage implements IStorage {
...
+  private static _getSyncKey(doc: DocImpl<any>, schemaContext?: SchemaContext) {
+    // I use SchemaPathSelector here instead of just SchemaContext because
+    // it's the more general way to load a doc.
+    const selector: SchemaPathSelector = {
+      path: [],
+      schemaContext: schemaContext,
+    };
+    const selectorRef = refer(JSON.stringify(selector)).toString();
+    const docId = entityIdStr(doc.entityId);
+    return `${docId}/application/json:${selectorRef}`;
```
seefeldb, line 164: might need doc.space as well

PR #1274, opened 2025-06-19
modify TS ast before compile

anotherjesse:
currently this supports ternary and binary expressions
    

---

## Summary by cubic
Added a TypeScript AST transformer to rewrite ternary and binary expressions involving OpaqueRef values before compilation. Also added a concurrency limiter to cap LLM API requests at 10 in flight.

- **New Features**
  - Transforms ternary expressions with OpaqueRef to use ifElse().
  - Transforms binary expressions with OpaqueRef in JSX to use derive().
  - Adds a concurrency limiter for LLM requests to prevent overload.

- **Bug Fixes**
  - Ensures correct imports for ifElse and derive are added when needed.

### DIFF
--- a/packages/js-runtime/test/ifelse-transform.test.ts
+++ b/packages/js-runtime/test/ifelse-transform.test.ts
@@ -1,0 +1,66 @@
+import { describe, it } from "@std/testing/bdd";
+import { expect } from "@std/expect";
+import { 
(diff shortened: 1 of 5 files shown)


anotherjesse:
ignore the LLM change - that was a different experiment

anotherjesse:
rework is in #1318

PR #1281, opened 2025-06-23
fix: incorporate cross space transaction API

Gozala:
## Summary by cubic
Added support for cross-space transactions in the storage interface, allowing transactions to read from multiple memory spaces and write to one.

- **New Features**
  - Introduced `reader` and `writer` methods for accessing different memory spaces within a transaction.
  - Added error types for transaction isolation and consistency across spaces.

### DIFF
--- a/packages/runner/src/storage/interface.ts
+++ b/packages/runner/src/storage/interface.ts
@@ -187,9 +187,9 @@ export interface IStorageTransaction {
 export interface ITransactionReader {
   /**
    * Reads a value from a (local) memory address and captures corresponding
-   * `Read` in the the transaction invariants. If value was written in read
-   * memory address in this transaction read will return value that was written
-   * as opposed to value stored.
+   * `Read` in the transaction invariants. If value was written in read memory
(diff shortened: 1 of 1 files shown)


seefeldb (approved):

packages/runner/src/storage/interface.ts
```diff
@@ -255,51 +269,116 @@ export interface IStorageTransactionAborted extends Error {
...
+  | IStorageTransactionAborted
+  | IStorageTransactionInconsistent
+  | IStorageTransactionWriteIsolationError;
+
+export interface IStorageTransactionComplete extends Error {
+  name: "StorageTransactionCompleteError";
+}
 export interface INotFoundError extends Error {}
 export type IStorageTransactionProgress = Variant<{
   open: IStorageTransactionLog;
```
seefeldb, line 312: clarification: to entries move from `open` to `pending` to `done`?

if i want to know _all_, no matter the status (as scheduler needs it), should it iterate over all of them? who else is going to call this?

PR #1286, opened 2025-06-24
Introduce `ct charm view` command

bfollington:
- **Add `ct charm link`**
- **Support arbitrary depth link syntax**
- **Working `ct charm view` command**
- **Add charm link tests**
- **Fix output for large objects**
- **Prefer render() over console.log**
- **Remove `charm view`, splitting to own PR**
- **Revert "Remove `charm view`, splitting to own PR"**


packages/cli/lib/charm.ts
```diff
@@ -216,37 +216,99 @@ export async function linkCharms(
...
+    source = argumentCell.get();
+  } catch (err) {
+    source = { error: "Unable to get source/arguments", details: err instanceof Error ? err.message : String(err) };
+  }
+  
+  // Get result (charm data)
+  const result = charm.get();
+  
+  // Get charms this one reads from
+  const readingFromCharms = manager.getReadingFrom(charm);
```
jsantell, line 292: nit: `readingFromCharms` and `readByCharms` redundant

packages/cli/lib/charm.ts
```diff
@@ -216,37 +216,99 @@ export async function linkCharms(
...
+  const recipeMeta = await getRecipeMeta(manager, config.charm);
+  const recipeName = recipeMeta.recipeName;
+  
+  // Get source (arguments/inputs)
+  let source: any;
+  try {
+    const argumentCell = manager.getArgument(charm);
+    source = argumentCell.get();
+  } catch (err) {
+    source = { error: "Unable to get source/arguments", details: err instanceof Error ? err.message : String(err) };
```
jsantell, line 285: Still learning where things in the system throw and why, but if `getArgument` throws, it looks like we're missing some critical pieces and should throw, but not too familiar with the runner side of things -- can this occur currently?

As another TODO "general way of doing X" like verbose, json, it'd be nice to have some general, consistent error handling to display when a command errors (*most* of the time i'd imagine it's e.g. a wrong charm id or something, rather than a "lets look at the stack trace" error) -- no changes requested, just future thoughts

jsantell (approved):
great work, thanks for tackling this! some comments but broadly this LGTM

packages/cli/commands/charm.ts
```diff
@@ -145,6 +146,87 @@ export const charm = new Command()
...
+  .example(
+    `ct charm view ${EX_ID} ${EX_COMP_CHARM}`,
+    `View detailed information about charm "${RAW_EX_COMP.charm!}".`,
+  )
+  .example(
+    `ct charm view ${EX_ID} ${EX_URL}`,
+    `View detailed information about charm "${RAW_EX_COMP.charm!}".`,
+  )
+  .option("-c,--charm <charm:string>", "The target charm ID.")
+  .option("--verbose", "Show full UI and other large objects")
```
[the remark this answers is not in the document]
bfollington: I went ahead and did this, switching to a `--json` flag

packages/cli/lib/charm.ts
bfollington: On reflection, I actually don't know why there should be an error check here, I will admit to some LLM use. I've dropped the try-catch because I don't think I would expect this to throw, since we already got a handle to the charm itself by this point in the function.

PR #1285, opened 2025-06-24
Add `ct charm link` command [CT-490]

bfollington:
Add two new `charm` subcommands:

`ct charm link <cell-a> <cell-b>`

```sh
ct charm link --identity ~/dev/.ct.key --url https://toolshed.saga-castor.ts.net/2025-06-23-claude baedreie5iwq33jsdwequdw2zik2qk753ioxfq4xq73pgtku7gc3rq6zzea/lists/0 baedreic6bwcfdj5lyeunczdft4i4atkdscx7mfbzjsljqfve7uu3hzsora/items

// Linked baedreie5iwq33jsdwequdw2zik2qk753ioxfq4xq73pgtku7gc3rq6zzea/lists/0 to baedreic6bwcfdj5lyeunczdft4i4atkdscx7mfbzjsljqfve7uu3hzsora/items
```
    

---

## Summary by cubic
Added the ct charm link command to let users link fields between charms, supporting deep nested paths and array indices.

- **New Features**
  - New command: ct charm link <source> <target>
  - Supports linking any field depth, including arrays

PR #1287, opened 2025-06-24
`ct charm link` support arbitrary cells as well as charms

bfollington:
- **Add support for linking to arbitrary cells OR charms**
- Update `setup-script` to explain linking to data
- Confirmed working testing with live instance

### DIFF
--- a/packages/cli/commands/charm.ts
+++ b/packages/cli/commands/charm.ts
@@ -156,21 +156,29 @@ export const charm = new Command()
     `ct charm link ${EX_ID} ${EX_COMP} bafycharm1/data/users/0/email bafycharm2/config/primaryEmail`,
     `Link deep nested field including array access.`,
   )
+  .example(
+    `ct charm link ${EX_ID} ${EX_COMP} baedreiahv63wxwgaem4hzjkizl4qncfgvca7pj5cvdon7cukumfon3ioye bafycharm1/allCharms`,
+    `Link well-known charms list to charm field.`,
+  )
   .arguments("<source:string> <target:string>")
   .action(async (options, sourceRef, targetRef) => {
     const spaceConfig = parseSpaceOptions(options);
 
-    // Parse source and target references
-    const source = parseLink(sourceRef);
+    // Parse source and target references - handle both charmId/path and well-known IDs
+    const source = parseLink(sourceRef, { allowWellKnown: true });
     const target = parseLink(targetRef);
 
-    // Validate that paths are provided for linking
-    if (!source.path) {
+    // For linking, we need paths unless source is a well-known ID
+    // Well-known IDs can be linked without a path (linking the entire cell)
+    const isWellKnownSource = !sourceRef.includes("/");
+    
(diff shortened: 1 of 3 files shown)


packages/cli/commands/charm.ts
```diff
@@ -156,24 +156,31 @@ export const charm = new Command()
...
 
-    // Parse source reference (charmId/path/to/field)
-    const sourceParts = sourceRef.split("/");
-    if (sourceParts.length < 2) {
-      throw new ValidationError(
-        `Invalid source reference format. Expected: charmId/path/to/field`,
-        { exitCode: 1 },
-      );
+    // Parse source reference - handle both charmId/path and well-known IDs
+    let sourceCharmId: string;
```
jsantell, line 168: If turning this into a `parseLink` function in #1285, could be `parseLink(ref: string, { allowWellKnown?: boolean })` or something

packages/cli/lib/charm.ts
```diff
@@ -217,35 +218,78 @@ export async function linkCharms(
 ): Promise<void> {
   const manager = await loadManager(config);
 
-  const sourceCharm = await manager.get(sourceCharmId, false);
-  if (!sourceCharm) {
-    throw new Error(`Source charm "${sourceCharmId}" not found`);
+  // Try to get source as a charm first, then as an arbitrary cell ID
+  let sourceCell: Cell<any>;
```
jsantell, line 222: wonder if this and ln239 could be rolled into a function

packages/cli/lib/charm.ts
```diff
@@ -217,35 +218,78 @@ export async function linkCharms(
...
+    targetCell = targetCharm;
+    targetIsCharm = true;
+  } catch (error) {
+    // If manager.get() fails (e.g., "recipeId is required"), try as arbitrary cell ID
+    try {
+      targetCell = await manager.getCellById({ "/": targetCharmId });
+      // Check if this cell is actually a charm by looking at the charms list
+      const charms = manager.getCharms().get();
+      const isActuallyCharm = charms.some(charm => {
+        const id = charmId(charm);
```
jsantell, line 253: should this be `getCharmIdSafe` from #1285? I guess we'd just skip it if it lacks a charm id

packages/cli/lib/charm.ts
```diff
@@ -217,35 +254,51 @@ export async function linkCharms(
...
+    sourceResultCell = sourceResultCell.key(segment);
   }
-  const sourceCellLink = sourceCell.getAsCellLink();
 
-  // Navigate to the parent of the target path
-  let targetCell: Cell<any> = targetCharm;
+  const sourceCellLink = sourceResultCell.getAsCellLink();
+
+  // Navigate to the target path
   const targetKey = targetPath.pop();
```
seefeldb, line 282: you could skip this and below just call targetInputCell.set() directly. Or was there a specific reason for this?

(it's also possible to change the top-level input of a charm, so even the non-empty constraint is technically not necessary)

PR #1288, opened 2025-06-24
chore: Use local workspace when running CLI integration tests via deno

jsantell:
Otherwise, use ct binary found in path.
    

---

## Summary by cubic
Updated CLI integration tests to use the local workspace build when running via Deno, falling back to the installed ct binary if not set.

- **Refactors**
  - Added environment variable to control which CLI is used during integration tests.
  - Updated test script and workflow to support this change.


jsantell:
CLI regression on main

seefeldb (approved):

PR #1291, opened 2025-06-24
chore: Handle macos usage of sed in CLI integration.

jsantell:
## Summary by cubic
Added a helper to handle differences in the sed command on macOS, fixing CLI integration script compatibility.


seefeldb (approved):
Works great

PR #1292, opened 2025-06-24
Feat/ct 442 query reconnection

ellyxir:


packages/runner/src/storage/cache.ts
```diff
@@ -326,6 +329,23 @@ class SelectorTracker {
       this.selectors.get(selectorRef)!
     );
   }
+
+  /**
+   * Return all tracked subscriptions as an array of {factAddress, selector} pairs
+   */
+  getAllSubscriptions(): Array<{ factAddress: FactAddress; selector: SchemaPathSelector }> {
```
ubik2, line 336: It's more idiomatic to do `{ factAddress: FactAddress; selector: SchemaPathSelector }[]` instead.

packages/runner/src/traverse.ts
```diff
@@ -158,7 +167,7 @@ export abstract class BaseObjectTraverser<K, S> {
       S,
       Immutable<JSONValue> | undefined
     >,
-  ) {}
+  ) { }
```
ubik2, line 170: Looks like helix doesn't format quite the same way
You can do a `deno fmt` to match the style, but there are a number of files that will be updated (since not everyone auto-formats).

packages/runner/src/traverse.ts
```diff
@@ -215,17 +224,17 @@ export abstract class BaseObjectTraverser<K, S> {
...
-                  {
-                    ...doc,
-                    path: [...doc.path, k],
-                    value: value,
-                  },
-                  tracker,
-                  schemaTracker,
-                ),
-              ]),
+                  k,
```
ubik2, line 227: same as above with helix formatting

packages/runner/src/traverse.ts
```diff
@@ -48,6 +48,15 @@ export class MapSet<K, V> {
   public has(key: K): boolean {
     return this.map.has(key);
   }
+
+  /**
+   * iterable
+   */
+  *[Symbol.iterator](): IterableIterator<[K, Set<V>]> {
```
ubik2, line 55: This makes me a little nervous, since a modification to the map would invalidate the iterator. It also makes values modifiable.
I might expose a method that just returns the pairs.

PR #1294, opened 2025-06-25
feat: implement multi-file recipe support

anotherjesse:
- Add bundling system to recursively extract and package local imports
- Update recipe manager to store and load additional source files
- Modify CLI charm commands to support multi-file recipes
- Fix resolver bug for absolute path detection
- Update harness to use Program interface for running recipes

This allows recipes to import from local TypeScript/TSX files, maintaining the full file structure when stored and retrieved.

---

## Summary by cubic
Added support for multi-file recipes, allowing recipes to import and bundle local TypeScript/TSX files while preserving their file structure.

- **New Features**
  - Bundles all local imports when running or saving a recipe.
  - Stores and loads additional source files in the recipe manager.
  - CLI commands now handle multi-file recipes.
  - Recipes can be exported and restored with their full file tree.

- **Bug Fixes**
  - Fixed resolver to correctly detect absolute paths.


anotherjesse:
I didn't read the code yet, sharing as a first step - before I start reviewing / revising.

treat this as a POC

packages/js-runtime/bundler/browser.ts
```diff
@@ -0,0 +1,7 @@
+/**
```
jsantell, line 1: All the js-runtime changes are unneeded, this is what `resolver.ts` (and the EngineResolver in `engine.ts`)  handles for us

packages/runner/src/recipe-manager.ts
```diff
@@ -11,6 +12,18 @@ export const recipeMetaSchema = {
     spec: { type: "string" },
     parents: { type: "array", items: { type: "string" } },
     recipeName: { type: "string" },
+    additionalSources: {
```
jsantell, line 15: Sketching this out, I replaced `src` with multiple sources (as this needs a list of sources with their names, as well as the name of the entry, names are "/paths/like/this", maybe we can bless a default `index.ts` or something in the future) -- reasoning here is we can represent a single file recipe as a multi file one rather than having an explicit "single file only" execution branch e.g. `{ entry: "/index.ts", sources: { "/index.ts": "import { .. } ..." }}`, though probably will need to cast from the JSONSchema form to what Engine wants (we represent sources as both `Record<string, string>` and `{ name: string, contents: string }[]` in different places), though a simple cast

packages/js-runtime/typescript/resolver.ts
```diff
@@ -80,7 +80,7 @@ function isUnresolvedModuleOk(
 }
 
 function resolveSpecifier(specifier: string, from: Source): string {
-  if (specifier[0] === "." || specifier[1] === "/") {
```
jsantell, line 83: I think this should be `specifier[0] === "." && specifier[1] === "/"`, want to capture `./relative/imports/`, though this looks like it's missing some scenarios (though `../` is tested elsewhere) hm

packages/runner/src/recipe-manager.ts
```diff
@@ -11,6 +12,18 @@ export const recipeMetaSchema = {
     spec: { type: "string" },
     parents: { type: "array", items: { type: "string" } },
     recipeName: { type: "string" },
+    additionalSources: { 
+      type: "array", 
+      items: {
```
jsantell, line 17: is anything else needed for persistence, other than storing as a part of the JSON schema?

wonder where else we use the `src` (a few times for IDs i think)

jsantell (commented):
thanks for taking a pass at this!

packages/js-runtime/bundler/mod.ts
```diff
@@ -0,0 +1,210 @@
...
+ * ```ts
+ * const result = await bundle('./src/app.tsx');
+ * // result.mainSource contains app.tsx content
+ * // result.additionalSources contains all imported local files
+ * ```
+ */
+export async function bundle(entryPath: string): Promise<BundleResult> {
+  const absoluteEntryPath = resolve(entryPath);
+  const basePath = dirname(absoluteEntryPath);
+  const mainSource = await Deno.readTextFile(absoluteEntryPath);
```
jsantell, line 147: no Deno in multi-platform code

packages/runner/src/harness/engine.ts
```diff
@@ -90,16 +90,24 @@ export class Engine extends EventTarget implements Harness {
...
+      entry: "/main.tsx",
+      files: [{ name: "/main.tsx", contents: source }]
+    }, options);
   }
 
   async run(
-    entry: string,
-    sources: Record<string, string>,
-    options: EngineProcessOptions,
+    program: Program,
```
jsantell, line 100: While we could use `Program` here as an interface, we need to convert it back to a `ProgramResolver` to resolve all of our commontools runtime modules and type environment, which then turns it back into a `Program` (now containing our environment). While `Program` here is simply used as an interface for entry+sources, maybe it'd be more clear if it were not a `js-runtime` type

packages/runner/src/recipe-manager.ts
anotherjesse: I shared this in discord, adding comment here for future us:

the reason I kept `src` and stored the "additional" source files in a new array is so that existing charms/recipes/spaces would continue to work.

If we want to make this a breaking change we can be more aggressive with naming

packages/runner/src/recipe-manager.ts
```diff
@@ -2,15 +2,38 @@ import { JSONSchema, Module, Recipe, Schema } from "./builder/types.ts";
 import { Cell } from "./cell.ts";
 import type { IRecipeManager, IRuntime, MemorySpace } from "./runtime.ts";
 import { createRef } from "./doc-map.ts";
+import { Program } from "@commontools/js-runtime";
+
+// * we need to prefix "files" at runtime with RecipeID(?) to uniqulely ID source mapped sources in CDT
+// * must we have an "entry" string to know which file is the entry VS "index.tsx", "main.tsx",
```
jsantell, line 8: todo remove

packages/runner/src/builtins/compile-and-run.ts
```diff
@@ -85,7 +85,7 @@ export function compileAndRun(
     // Now we're sure that we have a new file to compile
     compiling.setAtPath([], true, log);
 
-    const compilePromise = runtime.harness.runSingle(
+    const compilePromise = runtime.harness.run(
```
anotherjesse, line 88: (to myself or whoever knows): do we have test coverage for compileAndRun builtin?  should we include it in integrations?

packages/runner/src/recipe-manager.ts
```diff
@@ -152,51 +172,6 @@ export class RecipeManager implements IRecipeManager {
...
-      recipeMeta: {
-        id: recipeId,
-        src: recipeJson.src,
-        spec: recipeJson.spec,
-        parents: recipeJson.parents,
-      },
-    };
-  }
-
   async publishToBlobby(
```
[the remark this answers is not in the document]
anotherjesse: (publishToBlobby)

packages/runner/src/runtime.ts
```diff
@@ -177,15 +177,16 @@ export interface IStorage {
...
   registerRecipe(
     params: {
       recipeId: string;
       space: MemorySpace;
       recipe: Recipe | Module;
-      recipeMeta: any;
+      recipeMeta: RecipeMeta;
     },
   ): Promise<boolean>;
   publishToBlobby(recipeId: string): Promise<void>;
```
anotherjesse, line 192: remove publishToBlobby from interface

packages/runner/src/builtins/compile-and-run.ts
seefeldb: we don't yet have that. we should.

packages/runner/src/harness/engine.ts
```diff
@@ -86,45 +96,41 @@ export class Engine extends EventTarget implements Harness {
...
-    entry: string,
-    sources: Record<string, string>,
-    options: EngineProcessOptions,
-  ): Promise<Recipe> {
-    const { exports } = await this.process(
-      new EngineProgramResolver(entry, sources),
-      options,
-    );
+  // Compile and run a `Program`, returning the export default recipe.
+  async run(program: Program): Promise<Recipe> {
```
seefeldb, line 108: is leaving off `options: EngineProcessOptions` intentional here?

seefeldb (approved):

PR #1299, opened 2025-06-25
feat: memory transaction

Gozala:
Current implementation draft of the transaction API.
    

---

## Summary by cubic
Introduced a new memory transaction API that enables consistent, multi-space read and single-space write operations with commit and abort support.

- **New Features**
  - Added `IStorageManagerV2` and `IStorageTransaction` interfaces for transactional memory operations.
  - Implemented transaction lifecycle management, including commit, abort, and consistency checks.
  - Provided transaction readers and writers for safe, isolated access to memory spaces.


packages/runner/src/storage/cache.ts
```diff
@@ -1379,10 +1444,767 @@ export const getChanges = <
...
+  }
+
+  status(): Result<IStorageTransactionProgress, IStorageTransactionError> {
+    return this.#journal.state();
+  }
+
+  reader(
+    space: MemorySpace,
+  ): Result<TransactionReader, InactiveTransactionError> {
+    // Obtait edit session for this transaction, if it fails transaction is
```
seefeldb, line 1484: typo: Obtait

packages/runner/src/storage/cache.ts
```diff
@@ -1379,10 +1444,767 @@ export const getChanges = <
...
+        : (value as Record<string, JSONValue>)[key];
+    } else {
+      return {
+        error: state
+          ? new NotFound(
+            `Can not resolve "${address.type}" of "${address.id}" at "${
+              path.slice(0, at).join(".")
+            }" in "${address.space}", because target is not an object`,
+          )
+          : new NotFound(
```
seefeldb, line 2052: should this be a different error type?

or maybe add a field to the error that has an optional `path` that if the document did exist, it shows what part of the path matched (so the reader can use that as hint from where on the construct things)

seefeldb (commented):
Looks great already.

PR #1311, opened 2025-06-27
chore: source map invocations CT-500

jsantell:
## Summary by cubic
Refactored function invocation in the runner to use a new invoke method, preparing for improved source map support and error handling.

- **Refactors**
  - Replaced direct function calls with an invoke method in the engine and scheduler.
  - Simplified error logging in RuntimeContext.

### DIFF
--- a/packages/runner/src/scheduler.ts
+++ b/packages/runner/src/scheduler.ts
@@ -38,7 +38,7 @@ const MAX_ITERATIONS_PER_RUN = 100;
 
 export class Scheduler implements IScheduler {
   private pending = new Set<Action>();
-  private eventQueue: (() => void)[] = [];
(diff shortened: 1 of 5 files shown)


packages/runner/src/scheduler.ts
```diff
@@ -38,7 +38,7 @@ const MAX_ITERATIONS_PER_RUN = 100;
 
 export class Scheduler implements IScheduler {
   private pending = new Set<Action>();
-  private eventQueue: (() => void)[] = [];
+  private eventQueue: (() => any)[] = [];
```
jsantell, line 41: These functions do *not* return `void`, and their return values are meaningful to the system.

PR #1313, opened 2025-06-27
add .setSoureCell, freeze(), isFrozen() to `Cell` and remove existing…

seefeldb:
and remove existing uses via getDoc()
    

---

## Summary by cubic
Added setSourceCell, freeze, and isFrozen methods to Cell, and updated code to use these instead of accessing getDoc() directly.

- **Refactors**
  - Replaced direct getDoc() property access with new Cell methods.
  - Improved type safety and encapsulation for Cell operations.


ellyxir (approved):
lgtm

PR #1316, opened 2025-06-28
Ct 505 use runtimerun instead of runtimerunsynced in runnertestts

ellyxir:


ellyxir:
not sure about the runSynced test i wrote

ubik2:
@ellyxir is this a dead PR?

PR #1330, opened 2025-07-01
include the query args when we log a query failure error

ubik2:
## Summary by cubic
Now logs query arguments along with errors when a query fails, making it easier to debug issues.


ellyxir (approved):
lgtm

PR #1331, opened 2025-07-01
Added logging when we encounter a pointer cycle

ubik2:
## Summary by cubic
Added console logging to report when a pointer cycle is detected during object traversal. This helps with debugging cycles in complex data structures.


ellyxir (approved):
it feels a bit unwieldy just because how long these if statements. i think its fine, but maybe passing in a failure function as a callback to either PointerCycleTracker or CycleTracker might be nicer. but its fine with me!

PR #1335, opened 2025-07-01
When linking a recipe from a process cell, include the recipe link in a style that exposes the uri of the recipe

ubik2:
Added an optional recipe property to the process cell.
Eventually, we should be able to remove the special $TYPE handling, and just use this link to establish the dependency.
However, the main reason I added it is just so that it's easy to find the associated recipe.

I don't know if this is a good idea or not, but I find it handy for debugging.

---

## Summary by cubic
Process cells now include a recipe link that exposes the recipe's URI, making it easier to find the associated recipe. This prepares for future removal of special $TYPE handling.

### DIFF
--- a/packages/charm/src/manager.ts
+++ b/packages/charm/src/manager.ts
@@ -67,6 +67,7 @@ export const charmSourceCellSchema = {
   type: "object",
   properties: {
     [TYPE]: { type: "string" },
+    recipe: { type: "object" },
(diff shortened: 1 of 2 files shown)


seefeldb (approved):
Seems useful! Just rename it `spell` instead of `recipe`.

cc @anotherjesse since he wanted exactly this earlier :)

PR #1338, opened 2025-07-02
Make labels public

ubik2:
This makes it easier to determine what's going on for a client, and it's not really a security issue.
    

---

## Summary by cubic
Labels are now always public and included in commit data, making it easier for clients to access label information.

- **Bug Fixes**
  - Updated logic to treat all labels as unclassified and ensure they are not restricted.
  - Added tests to confirm labels are included in responses.


seefeldb (approved):
Worth noting that this isn't about showing the label to sandboxed code. This is between two runtimes.

PR #1341, opened 2025-07-02
enable multi-edit for recipes

anotherjesse:


anotherjesse:
this version doesn't implement save in place (the cli has setsrc to update the existing charm) - as there is an issue with recursion currently

PR #1343, opened 2025-07-02
CT-518 add charm on navigateTo in jumble

anotherjesse:
## Summary by cubic
Added logic to ensure charms are added to the charm manager when navigating to them in Jumble, and fixed an issue where navigating to a charm could fail or reuse cells incorrectly.

- **Bug Fixes**
  - Prevented navigation to undefined charms.
  - Fixed cell sharing issues when creating multiple charms from the compiler.

### DIFF
--- a/packages/jumble/src/components/CommandCenter.tsx
+++ b/packages/jumble/src/components/CommandCenter.tsx
@@ -435,16 +435,14 @@ export function CommandCenter() {
       });
     };
 
-    const handleNavigateToCharm = async (event: CustomEvent) => {
+    const handleNavigateToCharm = (event: CustomEvent) => {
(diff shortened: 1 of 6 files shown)


packages/jumble/src/components/CommandCenter.tsx
```diff
@@ -475,7 +485,7 @@ export function CommandCenter() {
       );
       globalThis.removeEventListener(
         "navigate-to-charm",
-        handleNavigateToCharm as EventListener,
```
anotherjesse, line 478: is this needed?

anotherjesse:
fyi @bfollington not sure if you saw weirdness with compileAndRun. I was hitting errors trying to navigate to the second charm I generated due to issues with navigate & sharing cells...

PR #1346, opened 2025-07-02
add 'deno task ct' to workspace

anotherjesse:
whether in jumble, cli, toolshed, or elsewhere, this let's you use `deno task ct`

this replaces the need for a compiled / up-to-date version of ct to be in your path
    

---

## Summary by cubic
Added a new "ct" task to the workspace so you can run `deno task ct` without needing a compiled version of ct in your path.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -25,6 +25,7 @@
   ],
   "tasks": {
     "check": "./tasks/check.sh",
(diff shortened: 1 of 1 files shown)


jsantell:
We'll need to override $PWD in the tasks to correctly use relative paths in commands (an issue also currently with `package/cli`'s task)

https://docs.deno.com/runtime/reference/cli/task/#specifying-the-current-working-directory

PR #1349, opened 2025-07-03
Implement charm view tree output

anotherjesse:
## Summary
- show charm view as a tree in the CLI
- support raw JSON output via `--json`

## Testing
- `deno task check` *(fails: Deno version mismatch)*
- `deno task test` *(fails: failed to load sqlite plug due to missing permissions)*

------
https://chatgpt.com/codex/tasks/task_e_6865d5551a80832ca1406c8e4c717d78

### DIFF
--- a/packages/cli/commands/charm.ts
+++ b/packages/cli/commands/charm.ts
@@ -3,7 +3,9 @@ import { Command, ValidationError } from "@cliffy/command";
 import {
   applyCharmInput,
   CharmConfig,
+  formatViewTree,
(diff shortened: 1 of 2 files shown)


seefeldb (approved):
Not blocking, but inspectCharm() and then also render could be improved by using the charm's argument and result schema, and then also call render with a cell on the UI schema.

It would also be a first step to getting all the streams out -- or if you want to be fancy render the UI as it updates

PR #1350, opened 2025-07-03
remind claude we use deno not node, and some default permissions

anotherjesse:
## Summary by cubic
Added a script and settings to block Node.js commands and remind that this repo uses Deno, plus set default allowed shell commands for Claude.

- **New Features**
  - Blocks npm, npx, yarn, pnpm, and node commands with a clear message.
  - Allows only specific Bash commands by default.


jakedahn (approved):
good idea!

PR #1352, opened 2025-07-03
Fix/schema traversal mismatch - CT-541

ubik2:
Fix some spots where I wasn't returning the value, so I returned undefined instead (which indicates that the object didn't match the schema).
    

---

## Summary by cubic
Fixed a bug in schema traversal where some functions did not return values, causing undefined results for valid objects.

- **Bug Fixes**
  - Updated traversal methods to always return the correct value.
  - Added a test to confirm all entries are returned from a charm list.

### DIFF
--- a/packages/memory/test/space-test.ts
+++ b/packages/memory/test/space-test.ts
@@ -1,6 +1,6 @@
 import { assert, assertEquals, assertExists, assertMatch } from "@std/assert";
 import { refer } from "merkle-reference";
-import { SchemaContext } from "@commontools/runner";
+import { JSONSchema, SchemaContext } from "@commontools/runner";
 import * as Changes from "../changes.ts";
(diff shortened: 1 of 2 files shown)


seefeldb (approved):

ellyxir (approved):
looks good to me, its unfortunate that typescript doesnt warn about this, i wonder if there is another warning level for this. otherwise perhaps we should stop having so many types that include undefined as an option.

anotherjesse:
this improves the Charm list, but doesn't address background charms / other data "loss" issues

ubik2:
> looks good to me, its unfortunate that typescript doesnt warn about this, i wonder if there is another warning level for this. otherwise perhaps we should stop having so many types that include undefined as an option.

eslint has consistent-return, but that's not available from the deno lint.

PR #1368, opened 2025-07-08
Introduce `ct charm get` and `ct charm set` to change cell values

bfollington:
- **First pass implementation of working `set` and `get`**
- **Tweak docs**


packages/cli/test/charm-input-flag.test.ts
```diff
@@ -0,0 +1,48 @@
+import { describe, it } from "@std/testing/bdd";
```
bfollington, line 1: I think we should just drop these tests, superfluous

packages/cli/test/charm-operations-integration.test.ts
```diff
@@ -0,0 +1,75 @@
+import { describe, it } from "@std/testing/bdd";
```
bfollington, line 1: I think we should just drop these tests, superfluous. We should prefer to extend the `integration.sh` script for more robust testing.

packages/cli/test/charm-cell-operations.test.ts
```diff
@@ -0,0 +1,82 @@
...
+      100000,
+      "scientific",
+    ]);
+  });
+
+  it("should handle negative numbers", () => {
+    // Negative integers are converted to numbers
+    expect(parseCellPath("values/-1/negative")).toEqual([
+      "values",
+      -1,
```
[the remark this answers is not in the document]
bfollington: No 🤦 , will fix

packages/cli/lib/path-helpers.ts
```diff
@@ -0,0 +1,41 @@
+/**
```
bfollington, line 1: I think this file should be deleted in favour of the (dupicate? similar?) functions in `cell-operations.ts`. It's better to expose them from that package as general-purpose helpers than the cli package.

packages/charm/test/cell-operations.test.ts
```diff
@@ -0,0 +1,198 @@
+import { assertEquals, assertRejects } from "@std/assert";
```
bfollington, line 1: I think we should just drop these tests, superfluous. We should prefer to extend the `integration.sh` script for more robust testing.

PR #1378, opened 2025-07-09
refactor(runner): move remaining places like reactivity log and query result proxy to new tx API, require tx to be passed in in more places

seefeldb:
- Refactored core runner modules to migrate reactivity logging and query result proxy logic to the new transaction (tx) API.
- Updated builder, builtins, and data-updating modules to align with the new tx-based architecture.
- Adjusted cell, doc, link-utils, recipe-binding, recipe-manager, runner, runtime, scheduler, and schema modules for compatibility with the new approach.
- Improved code organization and internal consistency, reducing legacy patterns and centralizing state management.
    

---

## Summary by cubic
Refactored the runner to move reactivity logging and query result proxy logic to the new transaction (tx) API, requiring tx to be passed in more places for better consistency and state management.

- **Refactors**
  - Updated core modules, builtins, and data-updating logic to use the tx API.
  - Removed legacy patterns and centralized transactional state handling.


packages/charm/src/commands.ts
```diff
@@ -21,10 +21,13 @@ export const castSpellAsCharm = async (
     if (!recipe) return;
 
     console.log("Casting...");
+    const tx = charmManager.runtime.edit();
```
anotherjesse, line 24: is there a reason we wouldn't put this edit/commit at the start/end of runPersistent?

seefeldb: I ended up making `tx` optional for read-only operations, but require it for anything that can write.

PR #1383, opened 2025-07-09
Changes how schema subscriptions communicate changes to the client

ubik2:
Changes how schema subscriptions communicate changes to the client
- When we send over commit messages, we send them as an EnhancedCommit that has the commit and all the facts that our schema subscriptions would include based on this commit.
- The QuerySubscriptionInvocation object forwards all EnhancedCommits to the controller, where they will be picked up by the cache.
- The Replica (cache) integrates the revisions in the EnhancedCommit, instead of attempting to extract the data from the commit itself.
    

---

## Summary by cubic
Schema subscriptions now send EnhancedCommit objects to the client, bundling the commit and all related facts, so the cache can directly integrate all relevant revisions.

- **Refactors**
  - QuerySubscriptionInvocation streams EnhancedCommit objects instead of raw commits.
  - The cache (Replica) now integrates revisions from EnhancedCommit, simplifying data handling.

### DIFF
--- a/packages/memory/consumer.ts
+++ b/packages/memory/consumer.ts
@@ -1,12 +1,11 @@
-import type { Authorization, JSONValue, SchemaContext } from "./interface.ts";
-import {
+import type {
   Abilities,
+  Authorization,
   AuthorizationError,
   Await,
   Cause,
   Clock,
   Command,
-  Commit,
   ConnectionError,
   ConsumerCommandFor,
   ConsumerCommandInvocation,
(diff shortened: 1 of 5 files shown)


ubik2:
One difference with this change is that a direct schema subscription does not automatically cause you to get notifications. You also need the commit+json subscription, since we're sending the updates in that message now.
This shouldn't be an issue for our existing usage, since we use the cache, and that subscribes to the commit+json, but could matter for some custom client implementations.

There's also a slightly breaking change where if the client is this version, but the server is toolshed, you won't be able to talk properly.

PR #1393, opened 2025-07-11
`<ct-input />` takes a cell directly [CT-599]

bfollington:
Very simple, but it works!


seefeldb (approved):
You might want to support the case where it isn't called with $ and value is just a regular value. Then let's merge and start supporting all the other input variants?

(i think we want a single component that is analogous to `<input>`, so also type=checkbox & co, or what do you think?)

PR #1395, opened 2025-07-11
Introducing mnemonic passphrase and passkey authentication flows to shell

jakedahn:
## Summary by cubic
Added support for both mnemonic passphrase and passkey authentication flows to the shell, allowing users to register and log in with either method.

- **New Features**
  - Users can register and authenticate using browser passkeys (WebAuthn) or a generated mnemonic passphrase.
  - Login UI now guides users through registration, login, and credential management for both methods.
  - Credentials are stored securely and can be cleared or switched by the user.

### DIFF
--- a/packages/shell/src/lib/app/state.ts
+++ b/packages/shell/src/lib/app/state.ts
@@ -1,5 +1,24 @@
-import { Identity } from "@commontools/identity";
+import {
+  ANYONE,
+  DID,
+  Identity,
+  KeyStore,
+  PassKey,
+} from "@commontools/identity";
 import { Command } from "../commands.ts";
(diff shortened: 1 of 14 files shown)


packages/shell/src/lib/app/state.ts
```diff
@@ -1,22 +1,46 @@
...
 
-export function applyCommand(
+// Key store key name for user's key
+export const ROOT_KEY = "$ROOT_KEY";
+
+export async function applyCommand(
   state: AppState,
   command: Command,
-): AppState {
+): Promise<AppState> {
```
jakedahn, line 43: @jsantell flagging this as an FYI-- everything seems to be working, but I'm not sure if async commands jive with things, exactly

PR #1397, opened 2025-07-12
Performance improvement: Reduce unnecessary interim cells or reads

seefeldb:


ubik2 (approved):
lgtm!

createCell already mentions that the corner case where you switch from stream to non-stream isn't supported, so Copilot's feedback isn't applicable.

PR #1400, opened 2025-07-14
Refactoring the shell app state to be synchronous only

jakedahn:
refactor: isolate authentication flow within LoginView

  Moves all authentication logic into LoginView as a self-contained system,
  removing auth-specific commands from the global state management.

  Changes:
  - LoginView now handles all auth operations internally via its own event system
  - Removed auth commands from global Command type (passkey-register,
    passkey-authenticate, passphrase-register, passphrase-authenticate)
  - Made applyCommand() synchronous by moving async auth operations to LoginView
  - Only set-identity command is used to communicate successful auth to the app

  Benefits:
  - Authentication is now a black box to the rest of the application
  - Global state management is simpler and synchronous
  - Better separation of concerns between auth UI and app state
  - Easier to modify auth flows without touching global state

  The app only knows that an Identity was set, not how it was obtained.

### DIFF
--- a/packages/shell/src/views/LoginView.ts
+++ b/packages/shell/src/views/LoginView.ts
@@ -433,7 +433,7 @@ export class XLoginView extends BaseView {
     }
   }
 
-  private async handleRegister() {
+  private handleRegister() {
     if (this.method === AUTH_METHOD_PASSKEY) {
       this.dispatchAuthEvent("passkey-register");
     } else {


packages/shell/src/views/LoginView.ts
```diff
@@ -283,18 +283,17 @@ export class XLoginView extends BaseView {
 
   private async handleRegister() {
     if (this.method === AUTH_METHOD_PASSKEY) {
-      await this.handleAuth(() => {
-        this.command({
+      await this.handleAuth(async () => {
+        await this.command({
```
jsantell, line 287: Sidestepping away from commands for a bit, the "login events" can just be their own events:

```
  this.dispatchEvent(
      new CustomEvent("passphrase-register", {
        composed: true,
        bubbles: true,
      }),
    );
```

Then the login view could `this.addEventListener("passphrase-register", this.onRegister)`. Optionally, these login events *may* be a single event type (e.g. "authevent") with details containing "subtype" (just like commands), which would allow us to bind a single event listener for login view (where e.g. `this.onRegister` has a switch statement for individual subtypes, etc.)

Then, the e.g. "passphrase-register" names can be removed from `commands`, `app/state` etc., as well as the shell-command-result listeners in the base view -- "shell commands" are specifically "events" that mutate the root app state -- as these events are fully contained within login view, the root app is none the wiser

jsantell (approved):
Fantastic, thank you! This cleans up our root app state and consolidates the auth logic to a single place in a delightful way

PR #1406, opened 2025-07-14
refactor: move KeyStore ownership to LoginView

jakedahn:
- Remove keyStore from AppState and related commands
  - Move all KeyStore operations to LoginView where authentication happens
  - LoginView now handles checking existing auth on mount
  - RootView still clears the keystore on logout for proper cleanup
  - Remove 100ms sleep workaround per feedback

  This simplifies the architecture by keeping KeyStore management
  localized to the component that actually uses it for authentication.
    

---

## Summary by cubic
Moved all KeyStore logic from global app state to LoginView, so authentication and key management are now handled locally in the login flow. RootView still clears the keystore on logout for cleanup.

- **Refactors**
  - Removed keyStore from AppState and related commands.
  - LoginView now checks for existing authentication and manages KeyStore access.
  - Removed 100ms sleep workaround from RootView.

### DIFF
--- a/packages/shell/src/views/RootView.ts
+++ b/packages/shell/src/views/RootView.ts
@@ -1,15 +1,14 @@
 import { css, html, LitElement } from "lit";
 import { ContextProvider } from "@lit/context";
-import { applyCommand, AppState, ROOT_KEY } from "../lib/app/mod.ts";
+import { applyCommand, AppState } from "../lib/app/mod.ts";
 import { appContext } from "../contexts/app.ts";
 import { SHELL_COMMAND } from "./BaseView.ts";
 import { Command, isCommand } from "../lib/commands.ts";
 import { API_URL } from "../lib/env.ts";
 import { AppUpdateEvent } from "../lib/app/events.ts";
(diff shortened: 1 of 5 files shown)


packages/shell/src/views/RootView.ts
```diff
@@ -122,20 +91,31 @@ export class XRootView extends LitElement {
       // Apply command synchronously
       const state = applyCommand(this._provider.value, command);
 
-      // Handle clear-authentication specially - need to clear keystore
-      if (command.type === "clear-authentication" && state.keyStore) {
-        await state.keyStore.clear();
+      // Handle command-specific side effects
+      switch (command.type) {
+        case "clear-authentication": {
```
jakedahn, line 96: To make `logout` easy, we're still listening for a `clear-authentication` command to make its way through the `RootView`.

I wanted to put this inside of the `LoginView`, but then I couldn't quite figure out how to wire up the button click event from `Header.ts`

Is it reasonable to leave this here?

jsantell (approved):

PR #1426, opened 2025-07-16
added unit test for setting correct paths for array changes

ellyxir:


ubik2 (approved):
lgtm
it's tagged as a draft, though

ellyxir:
> lgtm it's tagged as a draft, though

yup, because i wanted to add a regression test to the other bug fix. thanks.

PR #1427, opened 2025-07-16
Removed all direct reads or writes to DocImpl, now everything goes via transactions

seefeldb:


packages/runner/src/cell.ts
```diff
@@ -355,8 +354,8 @@ function createRegularCell<T>(
   link: NormalizedFullLink,
   tx?: IExtendedStorageTransaction,
 ): Cell<T> {
-  const doc = runtime.documentMap.getDocByEntityId(link.space, link.id);
-  const { path, schema, rootSchema } = link;
+  const { space, path, schema, rootSchema } = link;
+  let readOnly: string | undefined;
```
ellyxir, line 358: suggest calling this readOnlyReason if we think we'll ever use it to show the reason, otherwise turn to boolean

packages/runner/src/data-updating.ts
```diff
@@ -82,6 +91,7 @@ export function normalizeAndDiff(
   link: NormalizedFullLink,
   newValue: unknown,
   context?: unknown,
+  options?: IReadOptions,
```
ellyxir, line 94: where do we ever use options?

PR #1434, opened 2025-07-17
Use using for cycle tracker

ubik2:
As of Typescript 5.2, there's a `using` keyword that will automatically dispose the object when it comes off the stack.
This simplifies some exception handling that would otherwise be necessary to ensure these objects are cleaned up.

Switch over the CycleTracker to use this functionality.
We could probably do the same for transactions (there's an `Symbol.asyncDispose` for that), to either avoid a `tx.commit()`, or if we  implemented a rollback, it could happen there instead.
    

---

## Summary by cubic
Updated CycleTracker to use the new TypeScript 5.2 `using` keyword for automatic resource cleanup, reducing manual exception handling and making cycle detection code simpler. 

- **Refactors**
  - Replaced manual enter/exit calls with `using` for automatic disposal.
  - Removed redundant try/finally blocks and simplified cycle detection logic.

### DIFF
--- a/packages/runner/src/traverse.ts
+++ b/packages/runner/src/traverse.ts
@@ -92,16 +92,21 @@ export class CycleTracker<K> {
   constructor() {
     this.partial = new Set<K>();
   }
-  enter(k: K): boolean {
+  include(k: K, context?: unknown) {
     if (this.partial.has(k)) {
-      console.error("Cycle Detected!");
-      return false;
+      console.error(
+        "Cycle Detected!",
+        context == null ? JSON.stringify(k) : JSON.stringify(context),
+      );
+      return null;
     }
     this.partial.add(k);
-    return true;
-  }
-  exit(k: K) {
-    this.partial.delete(k);
+    return {
+      tracker: this,
+      [Symbol.dispose]: () => {
+        this.partial.delete(k);
+      },
+    };
   }
 }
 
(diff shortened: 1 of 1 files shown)


ubik2:
I'm including @seefeldb and @Gozala here for feedback on whether this pattern would also be useful in the transactions.

ellyxir (approved):
looks good, maybe would be nice to have a unit test on the cycletracker itself but i think the overall tests give us enough coverage.

packages/runner/src/traverse.ts
```diff
@@ -92,16 +92,20 @@ export class CycleTracker<K> {
...
+      );
+      return null;
     }
     this.partial.add(k);
-    return true;
-  }
-  exit(k: K) {
-    this.partial.delete(k);
+    return {
+      [Symbol.dispose]: () => {
```
ellyxir, line 105: just realised, does Firefox support Symbol.dispose?

jsantell: We don't currently have any polyfills (e.g. babel) in our main building pipeline, and would think the platform needs to support the feature translated by typescript. It looks like this resource management API is still not in release on Firefox, and unsupported on Safari https://caniuse.com/mdn-javascript_statements_using

We could 1) investigate polyfilling this feature for browsers, 2) only use these features in Deno components, or 3) avoid using `using`

jsantell: Update: Looks like the TypeScript compiler `tsc` translates the `using` functionality to non-"using" JS, but we use `esbuild` in most cases (shell, deno-web-test, jumble->vite->esbuild) which transparently copies over `using` functionality, requiring runtime support. We could explore esbuild plugins to translate this, or switch to using tsc for compiling (a larger lift)

ubik2:
Jordan modified the esbuild parameters to handle using there (generating the disposable invocation code), and then incorporated parts of core.js to provide the polyfill for Symbol.disposable.

PR #1435, opened 2025-07-17
fix(runner): prevent storage race condition with idle() synchronization

seefeldb:
Fix infinite loop caused by circular dependency between _updateDoc and _sendDocValue when both wait for runtime.idle(). The race condition occurred when storage updates triggered doc updates while doc changes were being sent to storage, causing stale data to be written.

Solution:
- Track all active _updateDoc operations with a global counter
- Make _sendDocValue wait for all pending _updateDoc operations to complete
- Use a shared promise that resolves when all updates finish
- Ensure proper sequencing: idle → wait for updates → idle again

This prevents the infinite loop where:

1. Storage update triggers _updateDoc (waits for idle)
2. Doc change triggers _updateStorage → _sendDocValue (waits for idle)
3. _sendDocValue sends stale data → triggers another storage update → repeat

Fixes the race condition identified in storage.ts where changes could be queued behind stale data, ensuring data consistency between docs and storage.
    

---

## Summary by cubic
Fixed a race condition in storage synchronization that caused infinite loops and stale data by ensuring doc and storage updates wait for each other to finish.

- **Bug Fixes**
  - Tracked active doc update operations with a counter and shared promise.
  - Made storage updates wait for all pending doc updates before proceeding.
  - Improved sequencing to prevent circular dependencies and ensure data consistency.

### DIFF
--- a/packages/runner/src/storage.ts
+++ b/packages/runner/src/storage.ts
@@ -463,10 +463,9 @@ export class Storage implements IStorage {
     this.dirtyDocs.add(docKey);
     const docToStoragePromise = this._sendDocValue(doc, labels);
     this.docToStoragePromises.add(docToStoragePromise);
-    docToStoragePromise.finally(() => {
-      this.docToStoragePromises.delete(docToStoragePromise);
-      this.dirtyDocs.delete(docKey);
-    });
+    docToStoragePromise.finally(() =>
+      this.docToStoragePromises.delete(docToStoragePromise)
+    );
   }
 
   private async _sendDocValue(doc: DocImpl<unknown>, labels?: Labels) {
@@ -480,6 +479,8 @@ export class Storage implements IStorage {
       await this.updateFromStoragePromise;
       await this.runtime.idle();
     }
+
+    this.dirtyDocs.delete(`${doc.space}/${toURI(doc.entityId)}`);
 
     const storageProvider = this._getStorageProviderForSpace(doc.space);
 
@@ -532,7 +533,9 @@ export class Storage implements IStorage {
         doc.sourceCell = newSourceCell;
       }
     } finally {
-      // Remove the processing flag
+      // Remove the processing flag. Do this _after_ doc.send(), so _updateDoc,
+      // which is called synchronously during that call is not going to schedule
+      // sending this back to storage.
       this.dirtyDocs.delete(docKey);
 
       // Decrement the counter
--- a/packages/runner/src/storage.ts
+++ b/packages/runner/src/storage.ts
@@ -190,6 +190,7 @@ export class Storage implements IStorage {
     this.dirtyDocs.clear();
     this.activeUpdateFromStorageCount = 0;
     this.updateFromStoragePromise = undefined;
+    this._updateFromStorageResolver = undefined;
     this.cancel();
   }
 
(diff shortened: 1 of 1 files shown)


packages/runner/src/storage.ts
```diff
@@ -410,16 +416,7 @@ export class Storage implements IStorage {
...
-      value:
-        (doc.get() === undefined
-          ? undefined
-          : JSON.parse(JSON.stringify(doc.get()))),
-      ...(doc.sourceCell?.entityId !== undefined)
-        ? { source: doc.sourceCell.entityId }
-        : {},
-      ...(labels !== undefined) ? { labels: labels } : {},
-    };
+    const storageValue = Storage._cellLinkToJSON(doc, labels);
```
ubik2, line 419: I'm not sure if we have any queryResultProxy objects still, but it does make sense to use the same function for all the flows.

ubik2:
Problem this attempts to solve:
When a call to update storage and call to update doc both happen and both are calling `await runtime.idle()`, we don't ensure that update doc happens first (we don't want to update storage with doc contents based on old storage contents).

The idea of this PR is to delay sending doc contents to storage until after all storage contents have been sent to the doc, and the runtime is once again idle.

This is accomplished by having the storageToDoc path create a promise before the `await runtime.idle()`, so that if the docToStorage gets scheduled first, it will wait until the storageToDoc calls have updated the doc.

packages/runner/src/storage.ts
```diff
@@ -467,20 +464,34 @@ export class Storage implements IStorage {
...
+
+    // Compare again, since the value might have changed since we last checked
+    const existingValue = this._getStorageProviderForSpace(doc.space).get<
+      JSONValue
+    >(
+      doc.entityId,
+    );
+    // If our value is the same as what storage has, we don't need to do anything.
+    if (deepEqual(storageValue, existingValue)) {
```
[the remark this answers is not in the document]
seefeldb: Ah, removing

seefeldb:
@ubik2 changed some more, and it improves the situation quite a bit. but now i'm removing the dirty bit on a doc only after it's send is complete, which worries me a little, since if the document locally changed since, we might loose that update.

either way, i keep hitting the issue quite reliably with the todo-list recipe when changing it's title.

packages/runner/src/storage.ts
```diff
@@ -451,15 +451,17 @@ export class Storage implements IStorage {
...
     }
     this.dirtyDocs.add(docKey);
     const docToStoragePromise = this._sendDocValue(doc, labels);
     this.docToStoragePromises.add(docToStoragePromise);
-    docToStoragePromise.finally(() =>
-      this.docToStoragePromises.delete(docToStoragePromise)
-    );
+    docToStoragePromise.finally(() => {
+      this.docToStoragePromises.delete(docToStoragePromise);
+      this.dirtyDocs.delete(docKey);
```
ubik2, line 463: This might be too late. We want to remove it from dirty as soon as we capture the doc's value so that a new value that occurs later will still be sent.
For example, if I set to 1, then while in `await storageManager.send`, I increment again, I do want to send the 2, but if the doc is still marked dirty, we'd skip sending the update to storage.
I think I chose a confusing variable name.

ubik2: Perhaps docsAwaitingSnapshot

PR #1436, opened 2025-07-18
Fix precedence problem that caused us to clear from pending and seen when we shouldn't have.

ubik2:
## Summary by cubic
Fixed a logic error in cache filtering that caused pending and seen items to be cleared incorrectly. This prevents data from being removed when it should be kept.


ubik2:
Need to add unit test, since I should have caught this when I initially regressed it.

ubik2:
The seen check can also be replaced with checking the since value. I'm pretty sure I broke this when I added seen handling.

ubik2:
Going to merge, and open a ticket for the test

PR #1437, opened 2025-07-18
initial logger with module and severity support

ellyxir:


packages/utils/src/logger.ts
```diff
@@ -0,0 +1,396 @@
...
+
+/**
+ * Current minimum log level - messages below this level are ignored
+ */
+let currentLogLevel: LogLevel = "info";
+
+/**
+ * Initialize log level from environment variable if available
+ */
+if (typeof Deno !== "undefined") {
```
jsantell, line 112: `deno-web-test` is an environment that has `globalThis.Deno` but not `Deno.env` -- I think `@commontools/utils/env` handles this distinction

packages/utils/src/logger.ts
```diff
@@ -0,0 +1,396 @@
...
+    return Array.isArray(resolved) ? resolved : [resolved];
+  });
+}
+
+/**
+ * Basic log function that accepts any message(s) and logs them to the console
+ * Messages can be values or zero-arity functions that return values (for lazy evaluation)
+ * @param messages - The messages to log
+ */
+export function log(...messages: LogMessage[]): void {
```
jsantell, line 172: `TaggedLogger` could be a class, and this could be a global instance, the implementations seem almost identical sans the styling

packages/utils/src/logger.ts
```diff
@@ -0,0 +1,396 @@
...
+  } catch {
+    // Ignore permission errors - use default log level
+  }
+}
+
+/**
+ * Set the minimum log level
+ * @param level - The minimum level to log (debug < info < warn < error)
+ */
+export function setLogLevel(level: LogLevel): void {
```
jsantell, line 127: Where do we expect this being used? This is global state and could be unexpected outside of "entry" files like shell, jumble, CLI -- even more so as it can change at runtime (if executed from multiple packages). what's the rationale for this to be dynamic (or at least, only able to be set globally once)?

packages/utils/src/simple-logger-example.ts
```diff
@@ -0,0 +1,24 @@
+import { getLogger, log, setLogLevel } from "./logger.ts";
```
jsantell, line 1: I think jsdocs and tests cover everything presented here -- disconnected source files have an opportunity for becoming out of date

packages/utils/src/logger.ts
```diff
@@ -0,0 +1,396 @@
...
+   * - true: Explicitly disabled, all logs are skipped
+   * - false: Explicitly enabled, all logs are shown
+   *
+   * When undefined, future versions can check global config for this module.
+   * Explicit true/false values will always override any global config.
+   */
+  disabled: boolean | undefined;
+}
+
+/**
```
jsantell, line 250: The stack trace parsing seems more trouble than its worth -- `import.meta.url` does not exist at runtime, source maps don't exist at this stack trace level, and legible source may be rewritten during builds e.g. minification in production builds

packages/utils/src/logger.ts
```diff
@@ -0,0 +1,396 @@
...
+ * Options for creating a tagged logger
+ */
+export interface GetLoggerOptions {
+  /**
+   * The import.meta.url from the calling module (optional - will auto-detect if not provided)
+   */
+  url?: string;
```
jsantell, line 305: Maybe unneeded

packages/utils/src/logger.ts
```diff
@@ -0,0 +1,396 @@
...
+ * import { getLogger } from "@commontools/utils/logger";
+ *
+ * // Auto-detects calling module from stack trace - recommended approach
+ * const logger = getLogger();
+ *
+ * // Logs will show: [INFO][your-module::HH:MM:SS.mmm] message
+ * logger.info("Processing user data");
+ * logger.debug("Cache hit for user", userId);
+ * logger.warn("API rate limit approaching");
+ * logger.error("Failed to save user", error);
```
jsantell, line 44: The most canonical method of `console.log` seems like an omission (alias for `info`); without reading docs, i'd expect `logger.log` to exist

jsantell (commented):
:partying_face: thank you!

packages/utils/src/logger.ts
ellyxir: alright, so perhaps we have a mandatory name passed in for getLogger. its not terrible

packages/utils/src/logger.ts
ellyxir: agreed

packages/utils/src/logger.ts
ellyxir: making log just a normal log and you'll call log.level to set a new level

packages/utils/src/logger.ts
ellyxir: will make it call this.info()

seefeldb (approved):
lgtm

one feature we might want to think about early is the ability to within a module specify a topic. often i want maximum logging, but only for some topic (say in storage just show me all reads and writes, but hide the scheduling stuff, or the other way around).

maybe there is a nice pattern for this, e.g.

```
const topics = getTopics(["data", "scheduling"])
const logger = getLogger("my-module", { enabled: [{ [topics.data]: "debug }] })
...
logger.log(topics.data, () => ...)
```

where the logging code recognizes this as a topic and formats it accordingly, and tooling can set severity per topic?

jsantell (approved):

PR #1440, opened 2025-07-18
fix(runner): No longer evict DocImpl docs

seefeldb:
As we've refactored towards using links with ids for references, actual DocImpl uses are now short-term and so there is a risk that the data gets evicted.
    

---

## Summary by cubic
Stopped evicting DocImpl documents from the runner's document map to prevent accidental data loss during short-term use.

- **Refactors**
 - Replaced weak reference maps with a standard Map to keep DocImpl instances in memory.


ellyxir (commented):
lgtm

PR #1446, opened 2025-07-18
chore: enable resource management/using polyfill in esbuild/felt

jsantell:
Adds support for polyfilling the `using` keyword in esbuild inside of felt (shell) and deno-web-test and jumble's vite.
    

---

## Summary by cubic
Added support for polyfilling the `using` keyword in esbuild configs for felt and deno-web-test to enable resource management features.

- **Dependencies**
  - Updated esbuild config to set `supported.using: false` in felt, shell, and deno-web-test.


ubik2 (approved):
lgtm!
thanks

PR #1447, opened 2025-07-18
chore: Use core-js resource management polyfills in shell.

jsantell:
## Summary by cubic
Added core-js polyfills for explicit and async resource management to the shell package for better compatibility with resource management primitives.


ubik2 (approved):
lgtm!
tested on safari/jumble

PR #1448, opened 2025-07-18
Rough pass to avoid having doc marked as already awaiting snapshot wh…

ubik2:
…ile we update doc.


ubik2:
This handles an overcorrection in https://github.com/commontoolsinc/labs/pull/1435 that could have prevented us from capturing a value of the doc and sending it to storage when we're integrating a change from storage back into the doc.
The dirtyDocs is intended to prevent us from having multiple tasks waiting to take a snapshot of the current doc value to send to storage, but if we put the doc in that collection while sending the storage value to the doc, we may lose a new doc write.

seefeldb (approved):

PR #1459, opened 2025-07-21
Don't return immediately when there's a pending sync

ubik2:
In cache.sync, if we have a pending load for the doc with the same schema, we were returning immediately.

We should instead return a promise that resolves when the initial query/subscribe retuns.
    

---

## Summary by cubic
Updated cache.sync to return a promise that resolves when a pending document sync completes, instead of returning immediately when a sync is already in progress.

- **Bug Fixes**
  - Prevents duplicate syncs by awaiting the original sync promise for the same document and schema.

### DIFF
--- a/packages/runner/test/storage.test.ts
+++ b/packages/runner/test/storage.test.ts
@@ -153,6 +153,27 @@ describe("Storage", () => {
       await testCell.sync();
       expect(synced).toBe(true);
     });
+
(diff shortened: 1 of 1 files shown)


ubik2:
When we don't provide a schema, we will re-issue the load call. This was already the case, but perhaps I should alter the code to avoid the duplicate call.

ubik2:
Fixes CT-629

seefeldb (approved):

PR #1460, opened 2025-07-21
Adding `ct charm rm <id>` cli call, along with CharmsController remove interface

jakedahn:
## Summary by cubic
Added the `ct charm rm <id>` CLI command to remove a charm by ID, including a new remove method in the CharmsController.

- **New Features**
  - Users can now remove charms from the CLI with `ct charm rm <id>`.
  - The CharmsController exposes a `remove` method to handle charm deletion and cleanup.

### DIFF
--- a/packages/cli/commands/charm.ts
+++ b/packages/cli/commands/charm.ts
@@ -13,6 +13,7 @@ import {
   listCharms,
   MapFormat,
   newCharm,
+  removeCharm,
   saveCharmRecipe,
   setCellValue,
   setCharmRecipe,
@@ -390,7 +391,8 @@ Recipe: ${charmData.recipeName || "<no recipe name>"}
   )
   .example(
     `ct charm call ${EX_ID} ${EX_COMP_CHARM} setName '{"value":"My Name"}'`,
-    `Call the "setName" handler with arguments on charm "${RAW_EX_COMP.charm!}".`,
+    `Call the "setName" handler with arguments on charm "${RAW_EX_COMP
+      .charm!}".`,
   )
   .option("-c,--charm <charm:string>", "The target charm ID.")
   .arguments("<handler:string> [args:string]")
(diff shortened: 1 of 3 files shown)


packages/cli/lib/charm.ts
```diff
@@ -477,20 +484,38 @@ export async function callCharmHandler(
...
+ * Removes a charm from the space (moves it to trash).
+ */
+export async function removeCharm(
+  config: CharmConfig,
+): Promise<void> {
+  const manager = await loadManager(config);
+  const charms = new CharmsController(manager);
+  const removed = await charms.remove(config.charm);
+
+  if (!removed) {
```
jsantell, line 515: This isn't a validation error (e.g. improper input), can just throw a regular error here

packages/cli/commands/charm.ts
```diff
@@ -399,6 +401,21 @@ Recipe: ${charmData.recipeName || "<no recipe name>"}
...
     render(`Called handler "${handlerName}" on charm ${charmConfig.charm}`);
+  })
+  /* charm rm */
+  .command("rm", "Remove a charm")
+  .alias("remove")
+  .usage(spaceUsage)
+  .example(
+    `ct charm rm baed..43mi ${EX_ID} ${EX_COMP}`,
+    `Remove charm "baed..43mi".`,
+  )
```
jsantell, line 412: nit: earlier `example()`s showed both the `--url` form (`EX_URL`) of calling, as well as the individual components (e.g. `--space`, etc.) (`EX_COMP`) (or `EX_COMP_CHARM` if has charm args) which may be worthwhile repeating for LLMs

packages/cli/commands/charm.ts
```diff
@@ -399,6 +401,21 @@ Recipe: ${charmData.recipeName || "<no recipe name>"}
...
+  .arguments("<charm:string>")
```
jsantell, line 413: This command takes a charm ID, like other commands do, but its an argument rather than defined via `--url` or `--charm` + other components -- the argument form is appealing but inconsistent with the other charm commands, maybe other charm commands could also take a charm arg, but we are already at 2 ways of executing charm commands -- easiest would be following the other charm commands, maybe we can make it more arg-y in the future. FWIW charm IDs i believe are only unique per host/space, so we'll still need to keep keying off of those values. Maybe in the future we can "push"/"pop" state like host/id/space with the tool to make it more natural when working in a single space -- maybe can get some ideas from prior art on other service CLI tools

jsantell (approved):
Very nice! Some comments on command consistency

packages/cli/commands/charm.ts
jakedahn: word, I'm going to just make it use the same pattern as the other commands with `--charm <id>` for now

PR #1468, opened 2025-07-22
refactor(runner): Convert cell.ts from function-based to class-based …

seefeldb:
…construction

Replace function-based object construction with proper ES6 classes for better code organization and type safety:

- Convert createRegularCell() to RegularCell class implementing Cell<T> interface
- Convert createStreamCell() to StreamCell class implementing Stream<T> interface
- Update createCell() to instantiate classes directly instead of calling factory functions
- Maintain full backward compatibility with existing Cell interface and behavior


seefeldb:
Ran benchmarks and there is no significant difference, somewhat disappointingly...

Next up is merging Cell and Stream into a single class and do the stream determination lazily on `send()` and maybe a new `isStream` method on the cell.

ubik2 (commented):
LGTM!

I do find the class versions quite a bit easier to read.
This isn't a recent change, but I didn't expect the sourceURI to be the link's id, since we also have the source property. I could have made my URI cleanup simpler.

PR #1472, opened 2025-07-22
fix(transactions): return NotFoundError or (new) TypeMismatchError instead of StorageTransactionInconsistent where appropriate

seefeldb:
## Improve storage transaction error handling with distinct error types

This PR enhances the storage transaction system by introducing a new `TypeMismatchError` to distinguish between different failure modes when accessing document properties. Previously, all such errors were reported as `StorageTransactionInconsistent`, which made it difficult to determine whether a retry would help.

### Problem

The storage transaction system previously conflated two distinct types of errors:
1. **Type mismatches** - Trying to access properties on non-objects (e.g., `"string".property`)
2. **Actual inconsistencies** - When underlying data changed during the transaction

Both were reported as `StorageTransactionInconsistent`, but they have very different implications:
- Type mismatches are persistent errors that will always fail, regardless of retries
- True inconsistencies indicate the underlying data changed and a retry might succeed

### Solution

This PR introduces a new `TypeMismatchError` to clearly distinguish these cases:

```typescript
// Before: Both scenarios threw StorageTransactionInconsistent
doc.value = "not an object";
doc.value.property; // StorageTransactionInconsistent

// After: Clear distinction
doc.value = "not an object";  
doc.value.property; // TypeMismatchError - will always fail
```

### Key Changes

1. **New Error Type**: Added `ITypeMismatchError` to the storage interface
2. **Updated Error Handling**: Modified attestation, chronicle, and transaction modules to use the appropriate error type
3. **Backwards Compatibility**: Transaction shim treats `TypeMismatchError` like `NotFoundError` for compatibility
4. **Bug Fix**: Chronicle now properly checks novelty before determining document existence
5. **Documentation**: Added comprehensive documentation explaining the transaction system and error handling

### Error Types After This PR

- **`StorageTransactionInconsistent`** - Read invariant violated (underlying state changed) - retry recommended
- **`NotFoundError`** - Document or path doesn't exist - transient error
- **`TypeMismatchError`** - Accessing properties on non-objects - persistent error  
- **`ReadOnlyAddressError`** - Writing to data: URIs - persistent error

### Testing

- Updated all existing tests to expect the correct error types
- Added new comprehensive test suite for NotFound error scenarios
- All tests pass with the new error handling

### Documentation

Added two new documentation files:
- `transaction-explainer.md` - High-level overview for users
- `transaction-implementation-guide.md` - Detailed guide for maintainers

This change makes error handling more predictable and helps consumers make better decisions about whether to retry failed operations.

    

---

## Summary by cubic
Improved storage transaction error handling by introducing a new TypeMismatchError and ensuring NotFoundError is returned when accessing non-existent documents or paths, instead of StorageTransactionInconsistent. This makes error types more precise and helps consumers decide when to retry operations.

- **Bug Fixes**
  - TypeMismatchError is now returned when accessing properties on non-object values.
  - NotFoundError is returned for missing documents or paths.
  - StorageTransactionInconsistent is only used for true state changes that may succeed on retry.
  - Updated transaction-shim and all related modules and tests to use the correct error types.

- **Documentation**
  - Added detailed guides explaining the transaction system and new error handling.


seefeldb:
Now we have 806 tests passing, 113 failing!

packages/runner/src/storage/interface.ts
```diff
@@ -657,6 +657,7 @@ export type CommitError =
 export interface INotFoundError extends Error {
   name: "NotFoundError";
   path?: MemoryAddressPathComponent[];
+  from(space: MemorySpace): INotFoundError;
```
ubik2, line 660: It looks like we lost the `source: IAttestation` and `address: IMemoryAddress` fields here.
It's not clear if that was intentional or accidental, since the interface was double defined with half here, and half later.

packages/runner/src/storage/transaction/attestation.ts
```diff
@@ -90,10 +94,12 @@ export const write = (
 
         return { ok: patch };
       } else {
+        // Type mismatch - trying to write property on non-object
```
ubik2, line 97: This is somewhat outside the PR, but on line 82 of the new file, we cast what can probably be an array into a Record<string, JSONValue>.
This works here, where we don't know the type, but wouldn't be a valid cast if typescript knew it was an array (you'd have to do an `as unknown`).
The behavior is correct, and maybe it's intentional to avoid code duplication just to appease types.

packages/runner/src/storage/transaction-shim.ts
```diff
@@ -203,6 +238,9 @@ class TransactionReader implements ITransactionReader {
 
         const validationError = validateParentPath(json, address.path);
```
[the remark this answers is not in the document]
seefeldb: done.

PR #1473, opened 2025-07-22
Improving the CharmListView <--> Default Recipe toggle, and adding the ability to remove charms from the CharmListView

jakedahn:
## Summary by cubic
Improved the toggle between CharmListView and Default Recipe views, and added the ability to remove charms directly from the CharmListView.

- **New Features**
  - Added a remove button for each charm in the CharmListView.
  - Updated the header with a toggle button to switch between the charm list and default recipe views.


packages/shell/src/views/SpaceView.ts
```diff
@@ -7,12 +7,16 @@ import { RuntimeInternals } from "../lib/runtime.ts";
...
+      display: block;
+      height: 100%;
+    }
   `;
 
   @property({ attribute: false })
   rt?: RuntimeInternals;
 
-  @state()
+  @property({ type: Boolean })
```
jsantell, line 19: nit: `type` isn't needed in TypeScript

packages/shell/src/lib/app/commands.ts
```diff
@@ -28,6 +29,9 @@ export function isCommand(value: unknown): value is Command {
     case "clear-authentication": {
       return true;
     }
+    case "toggle-charm-list": {
```
jsantell, line 32: nit: Maybe instead of toggle, this sets the value (and the individual views are responsible for setting the inverted state), keeping the commands stateless

packages/shell/src/views/HeaderView.ts
```diff
@@ -150,6 +174,17 @@ export class XHeaderView extends BaseView {
     this.command({ type: "clear-authentication" });
   }
 
+  private handleToggleClick(e: Event) {
+    e.preventDefault();
+    e.stopPropagation();
+    this.dispatchEvent(
+      new CustomEvent("toggle-view", {
```
jsantell, line 181: `toggle-view` is pretty generic (we have a lot of views!) -- maybe be more specific e.g. `toggle-charm-view`

though we also have a lot of "charm views" as well! Maybe the top level state could be "showDefaultCharm" (the inverse being showing the "native charm list") -- or instead of flipping, could be "showNativeCharmList" (slightly more unique)

packages/shell/src/views/HeaderView.ts
```diff
@@ -92,6 +113,9 @@ export class XHeaderView extends BaseView {
   @state()
   private _conflicts?: InspectorConflicts;
 
+  @property({ type: Boolean })
```
jsantell, line 116: nit: type unnecessary in TS

packages/shell/src/views/HeaderView.ts
```diff
@@ -201,13 +236,24 @@ export class XHeaderView extends BaseView {
...
         ${this.identity
         ? html`
-          <x-button
-            class="auth-button"
-            size="small"
-            @click="${this.handleAuthClick}"
-          >
-            Logout
-          </x-button>
+          <div class="right-section">
```
jsantell, line 239: Maybe x-flex and x-button (for emoji button) could be useful here

packages/shell/src/views/HeaderView.ts
```diff
@@ -150,6 +174,18 @@ export class XHeaderView extends BaseView {
     this.command({ type: "clear-authentication" });
   }
 
+  private handleToggleClick(e: Event) {
+    e.preventDefault();
+    e.stopPropagation();
+    this.dispatchEvent(
```
jsantell, line 180: `this.command()` could be used here, could get rid of `toggle-shell-charm-list-view` event entirely

jakedahn: good call

PR #1475, opened 2025-07-22
fix(cell): handle circular references in cell data assignment

seefeldb:
- Add circular reference detection in normalizeAndDiff() using a seen Map
- When circular references are detected, convert them to cell instances
- This prevents infinite recursion when assigning data with circular references
- Add comprehensive tests for circular reference handling in cells and across cells
- Tests verify that circular references are properly converted to relative links

### DIFF
--- a/packages/runner/src/cell.ts
+++ b/packages/runner/src/cell.ts
@@ -381,7 +381,7 @@ class StreamCell<T> implements Stream<T> {
   }
 }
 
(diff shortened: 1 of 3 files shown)


ubik2 (approved):

ubik2 (approved):
lgtm
normalizeAndDiff is a really long function, and could probably use breaking up at some point, but this PR doesn't make it significantly longer.

PR #1494, opened 2025-07-28
fix(runner/cache): strip `since` when loading from heap

seefeldb:
Is this the right place to do this? Motivation:
- The `State` type doesn't have since
- Some code (e.g. commit in chronicle.ts) uses the result of this function to compute a new cause, and so also doesn't expect `since`.

I ran into the latter, since the heap creates an entry with since = -1 when trying to access a document that doesn't exist. And then Chronicle:commit created the wrong cause when actually writing the new document.
    

---

## Summary by cubic
Fixed cache loading to remove the since field when returning state from the heap, preventing unexpected fields in downstream code.


ubik2:
I may be missing something, but we should already be excluding since from the fields used to calculate the cause.

PR #1499, opened 2025-07-28
Ellyse/ct 660 fix embedded derive in ternary

ellyxir:
## Summary by cubic
Fixed handling of embedded derive calls inside ternary expressions so that OpaqueRef operations are correctly transformed within all branches. Updated test cases and expected outputs to match the new transformation logic.


ellyxir:
looks like i messed up some of the expects for every and all , i will fix them in a new PR

PR #1501, opened 2025-07-29
Handle multiple OpaqueRefs in property access expressions

ellyxir:
This fixes an issue where expressions containing multiple OpaqueRef dependencies were not being properly transformed. The primary changes are:
  - Add support for multiple OpaqueRefs in PropertyAccessExpression, CallExpression, BinaryExpression, and TemplateExpression transformations
  - Implement deduplication logic to avoid creating redundant parameters for the same OpaqueRef appearing multiple times
  - Skip parameter type checking for methods called ON OpaqueRef objects to prevent incorrect early returns (fixes reduce transformation issue)
  - Add isFunctionParameter helper to properly identify and skip callback parameters in collectOpaqueRefs

The fix ensures that complex expressions like:
  state.items.filter(i => i.name.includes(state.filter)).length
properly collect both state.items and state.filter as dependencies while excluding callback parameters like 'i'.
    

---

## Summary by cubic
Fixed transformation of expressions with multiple OpaqueRef dependencies so that all relevant state references are collected, and callback parameters are excluded.

- **Bug Fixes**
  - Added support for multiple OpaqueRefs in property access, call, binary, and template expressions.
  - Deduplicated OpaqueRefs to avoid redundant parameters.
  - Skipped parameter type checks for methods called on OpaqueRef objects to fix issues with array methods like reduce.
  - Improved detection of function parameters to prevent collecting callback variables as dependencies.

### DIFF
--- a/packages/js-runtime/test/fixtures/jsx-expressions/jsx-complex-mixed.expected.tsx
+++ b/packages/js-runtime/test/fixtures/jsx-expressions/jsx-complex-mixed.expected.tsx
@@ -52,15 +52,15 @@ export default recipe({
         [UI]: (<div>
         <h3>Array Operations</h3>
         <p>Total items: {commontools_1.derive(state.items, _v1 => _v1.length)}</p>
-        <p>Filtered count: {commontools_1.derive(state.items, _v1 => _v1.filter(i => i.name.includes(state.filter)).length)}</p>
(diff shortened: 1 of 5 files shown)


ellyxir:
there's still one test that needs to be regenerated and fixed

PR #1504, opened 2025-07-29
fix(runner): Fix array index string comparison bug in transaction address handling

seefeldb:
The Address.includes() and Address.intersects() functions were using string-based path comparison which caused incorrect matching - e.g., path "10" would match as a child of path "1" because "10".startsWith("1") is true.

This caused transaction consistency errors when pushing more than 10 elements to an array, as array[10] would incorrectly be considered a child of array[1].

- Replace string concatenation/comparison with element-by-element array comparison
- Fix both includes() and intersects() functions in address.ts

The bug manifested as "StorageTransactionInconsistent" errors with messages like: 'Transaction consistency violated: The "application/json" of "..." at "value.internal.my_array.1" hash changed. Previously it used to be: 1 and currently it is: undefined'

### DIFF
--- a/packages/runner/src/storage/transaction/address.ts
+++ b/packages/runner/src/storage/transaction/address.ts
@@ -9,21 +9,44 @@ export const toString = (address: IMemoryAddress) =>
 export const includes = (
   source: IMemoryAddress,
   candidate: IMemoryAddress,
-) =>
-  source.id === candidate.id &&
-  source.type === candidate.type &&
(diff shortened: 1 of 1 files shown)


ubik2 (approved):
lgtm!
Nice to not also avoid the special treatment of "/", which can be in the path, since these are properties and not file paths. In particular, the links all have "/" in the path, but it's not a separator.

PR #1509, opened 2025-07-29
Minimal support for definitions

ubik2:
Add basic support for the definitions block of JSONSchema, and use this in the schema traversal.
Also added a missing signature for LiftFunction.

See #CT-672 for limitations
    

---

## Summary by cubic
Added minimal support for the definitions block in JSONSchema, allowing schema traversal to resolve $ref references to definitions.

- **New Features**
  - Schema traversal now resolves $ref values pointing to definitions and $defs.
  - Updated type definitions and tests to cover schemas using definitions.

### DIFF
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -148,6 +148,7 @@ export type JSONSchemaTypes =
 export type JSONSchema = {
   readonly $ref?: string;
   readonly $defs?: Readonly<Record<string, JSONSchema>>;
+  readonly definitions?: Readonly<Record<string, JSONSchema | boolean>>;
 
   // Subschema logic
   readonly allOf?: readonly (JSONSchema | boolean)[]; // not validated
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -148,6 +148,7 @@ export type JSONSchemaTypes =
 export type JSONSchema = {
   readonly $ref?: string;
   readonly $defs?: Readonly<Record<string, JSONSchema>>;
+  /** @deprecated Use `$defs` for 2019-09/Draft 8 or later */
   readonly definitions?: Readonly<Record<string, JSONSchema | boolean>>;
 
   // Subschema logic
@@ -331,7 +332,7 @@ export type LiftFunction = {
   ): ModuleFactory<Parameters<T>[0], ReturnType<T>>;
 
   <T, R>(
-    argumentSchema?: JSONSchema | ((input: any) => any),
+    argumentSchema?: JSONSchema,
     resultSchema?: JSONSchema,
     implementation?: (input: T) => R,
   ): ModuleFactory<T, R>;
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -329,6 +329,12 @@ export type LiftFunction = {
   <T extends (...args: any[]) => any>(
     implementation: T,
   ): ModuleFactory<Parameters<T>[0], ReturnType<T>>;
+
+  <T, R>(
+    argumentSchema?: JSONSchema | ((input: any) => any),
+    resultSchema?: JSONSchema,
+    implementation?: (input: T) => R,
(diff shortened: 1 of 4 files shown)


anotherjesse:
missing a `deno task compiile-api-types`

packages/api/index.ts
```diff
@@ -148,6 +148,7 @@ export type JSONSchemaTypes =
 export type JSONSchema = {
   readonly $ref?: string;
   readonly $defs?: Readonly<Record<string, JSONSchema>>;
```
seefeldb, line 150: per discussion, let's leave this out?

packages/api/index.ts
```diff
@@ -328,6 +329,12 @@ export type LiftFunction = {
   <T extends (...args: any[]) => any>(
     implementation: T,
   ): ModuleFactory<Parameters<T>[0], ReturnType<T>>;
+
+  <T, R>(
+    argumentSchema?: JSONSchema | ((input: any) => any),
```
seefeldb, line 334: Are you sure you need the `| ((input: any) => any)` here? We need this in the actual function definition because it has to match all overloads, but in the list of overloads, I'd remove this, so there is no ambiguity in the three parameter case

packages/runner/src/traverse.ts
```diff
@@ -625,6 +626,52 @@ export class SchemaObjectTraverser<K, S> extends BaseObjectTraverser<K, S> {
...
+      return undefined;
+    }
+    const schemaRef = schemaContext.schema["$ref"];
+    if (schemaRef === "#") {
+      return {
+        schema: schemaContext.rootSchema,
+        rootSchema: schemaContext.rootSchema,
+      };
+    } else if (
+      schemaRef.startsWith("#/definitions/") || schemaRef.startsWith("#/$defs/")
```
seefeldb, line 650: same, let's leave out $defs as you suggested?

more generally i wonder whether this should be a condition though, we could already say anything starting with `#/` is fine?

packages/runner/src/traverse.ts
```diff
@@ -625,6 +626,52 @@ export class SchemaObjectTraverser<K, S> extends BaseObjectTraverser<K, S> {
...
+    schemaContext: Readonly<SchemaContext>,
+  ): Readonly<SchemaContext> | undefined {
+    if (
+      !isObject(schemaContext.schema) || !("$ref" in schemaContext.schema) ||
+      schemaContext.schema["$ref"] === undefined
+    ) {
+      // We shouldn't have called this function, but just return it as-is
+      return schemaContext;
+    }
+    if (schemaContext.rootSchema === undefined) {
```
seefeldb, line 639: maybe to phase rootSchema out more slowly, just use the schema instead if there is no rootSchema?

seefeldb: except in the '#' case below, which isn't valid without a rootSchema

packages/runner/src/traverse.ts
```diff
@@ -625,6 +626,52 @@ export class SchemaObjectTraverser<K, S> extends BaseObjectTraverser<K, S> {
...
+        rootSchema: schemaContext.rootSchema,
+      };
+    } else if (
+      schemaRef.startsWith("#/definitions/") || schemaRef.startsWith("#/$defs/")
+    ) {
+      const pathToDef = schemaRef.split("/");
+      let schemaCursor = schemaContext.rootSchema;
+      // start at 1, since the 0 element is "#"
+      for (let i = 1; i < pathToDef.length; i++) {
+        if (!isObject(schemaCursor) || !(pathToDef[i] in schemaCursor)) {
```
seefeldb, line 656: super nit, but are arrays explicitly not allowed in the standard? use `isRecord` and then it's ok to be either an array or an object

packages/api/index.ts
ubik2: I had simply copied from the function. I'll remove the option here.

PR #1512, opened 2025-07-29
chore: Add a test on viewing and interacting with a charm in shell integration

jsantell:
## Summary by cubic
Added an integration test for viewing and interacting with a charm in the shell, including clicking buttons and verifying state changes.

- **Test Coverage**
  - Added a test that registers a charm, loads it in the shell, simulates user interaction, and checks the UI updates as expected.
  - Updated app controller to support setting identity from raw key pairs.
  - Minor update to the counter recipe to wrap the counter text in a span for easier test selection.

### DIFF
--- a/packages/shell/integration/utils.ts
+++ b/packages/shell/integration/utils.ts
@@ -1,0 +1,98 @@
+import { Page } from "@commontools/integration";
+import { ANYONE, Identity, InsecureCryptoKeyPair } from "@commontools/identity";
+import { AppState } from "../src/lib/app/mod.ts";
+import { Runtime } from "@commontools/runner";
+import { StorageManager } from "@commontools/runner/storage/cache";
+import { CharmManager } from "@commontools/charm";
+import { CharmsController } from "@commontools/charm/ops";
+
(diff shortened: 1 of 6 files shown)


packages/shell/integration/utils.ts
```diff
@@ -0,0 +1,100 @@
...
+    : await Identity.fromPassphrase(ANYONE);
+  const user = await account.derive(spaceName);
+  const session = {
+    private: account.did() === identity.did(),
+    name: spaceName,
+    space: user.did(),
+    as: user,
+  };
+
+  const apiUrl2 = "http://localhost:8000";
```
bfollington, line 79: Pardon my ignorance by, why is this a thing?

bfollington (approved):
LGTM, one question though

packages/shell/integration/utils.ts
jsantell: woops, this was separating the API_URL and FRONTEND_URL, would rather just target the built binary for this (and other integration tests eventually), good catch

PR #1513, opened 2025-07-29
Change StorageProvider.sync to take a SchemaPathSelector.

ubik2:
- Removed the expectedInStorage and schemaContext arguments to StorageProvider.sync, since the former was unused and the latter is replaced by selector.
- Changed the structures that tracked the SchemaContext in cache to use SchemaPathSelectors instead.
- Renamed Replia.getSchema to generateSchemaFromLabels to better indicate the function's purpose.
- Updated curl to new api that takes URI instead of EntityId.

### DIFF
--- a/packages/memory/consumer.ts
+++ b/packages/memory/consumer.ts
@@ -552,9 +552,9 @@ class QueryView<
   }
 
   // Get the facts returned by the query, together with the associated
-  // schema context used to query
-  get schemaFacts(): [Revision<Fact>, SchemaContext | undefined][] {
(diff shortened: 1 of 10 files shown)


ubik2:
I think I might be missing something larger in the heap->doc system. While the integration test does get the child object in the heap (meaning it was returned with the schema query, and we did the query correctly), it isn't in the document map, so we can't resolve the link from the synced cell.

Anyhow, that's not really part of this change, so I'm not doing anything special, but I probably need to clarify some of this stuff.

PR #1515, opened 2025-07-30
Fix cell array schema transformation

anotherjesse:
## Summary by cubic
Fixed schema transformation for arrays of Cell and Stream types so that Cell<T>[] and Stream<T>[] are now correctly recognized and marked with asCell or asStream in the generated schema.

- **Bug Fixes**
  - Detects Cell<T>[] and Stream<T>[] in type definitions and applies the correct schema flags.
  - Handles both Array<Cell<T>> and Cell<T>[] syntax.
  - Adds tests for arrays of cells, streams, nested cells, and mixed scenarios.


anotherjesse:
I don't like the way this works - since it will probably break for IFC<Cell<, ... (if Cell< isn't at the top)

PR #1518, opened 2025-07-30
refactor(runner): convert array_push test to use type-safe handlers

seefeldb:
- Replace proxy-based handlers with typed handler signatures
- Add test coverage for both primitive (number) and object arrays
- Add explicit ID field to objects to ensure proper tracking
- Update UI to display both array types separately

This change aligns with the codebase direction of avoiding proxy usage and ensuring type safety throughout the handler system.
    

---

## Summary by cubic
Refactored the array_push test to use type-safe handler signatures instead of proxies, added coverage for both number and object arrays, and updated the UI to show both array types.

- **Refactors**
  - Replaced proxy handlers with typed handlers for better type safety.
  - Added explicit ID field to objects for tracking.
  - Updated test and UI to handle and display both numbers and objects in arrays.

### DIFF
--- a/packages/runner/integration/array_push.test.ts
+++ b/packages/runner/integration/array_push.test.ts
@@ -87,23 +87,33 @@ async function runTest() {
   );
   console.log("Recipe compiled successfully");
 
-  const charm = await charmManager.runPersistent(
+  let charm = await charmManager.runPersistent(
     recipe,
     {},
   );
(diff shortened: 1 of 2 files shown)


seefeldb:
This currently gets the toolshed to first get really slow and eventually hang!

ubik2 (approved):
Looks good as a better test.
The array_push change prevents the terrible performance, so when this lands in main, it should be fine.

PR #1519, opened 2025-07-30
Handle the array push issue

ubik2:
Previously, when we pushed an item into an array, we would write the doc with the array, then later, write the doc with the array element.
This is problematic for another client, since they can get the updated array, and then attempt to load the array element before it has been written to the server, which means they end up treating it as an empty element.
This PR changes things to batch the write of a document to include any linked documents reachable by the schema query.

There are some problems with this code, but it makes things better.
    

---

## Summary by cubic
Fixed an issue where pushing an item to an array could cause clients to see empty elements by batching document writes to include all linked documents.

- **Bug Fixes**
  - Ensured that when an array is updated, all related linked documents are written together to prevent clients from loading incomplete data.

### DIFF
--- a/packages/runner/src/storage.ts
+++ b/packages/runner/src/storage.ts
@@ -1,6 +1,7 @@
 import { refer } from "merkle-reference";
 import { Immutable, isRecord } from "@commontools/utils/types";
 import type {
+  FactAddress,
   JSONValue,
   MemorySpace,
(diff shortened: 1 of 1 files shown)


ubik2:
I may also add the modified array_push integration test, since it's different enough from the main test.

seefeldb (approved):

PR #1525, opened 2025-07-30
chore: Remove redis from CI

jsantell:


jsantell:
> This looks good to me!
> 
> I noticed the `.github/workflows/evals.yml` still contains a redis setup -- but I also think we talked about removing that entirely anyway.

Good catch, the evals run is only called manually and has rotted a bit over the months and no longer working -- sounds like we want to at least disconnect it for now

PR #1535, opened 2025-07-31
fix(api): Add StripCell to handler with no json schema

seefeldb:
Also: Rename CellToOpaque to StripCell for clarity, since all uses then wrap the output in Opaque<>


ellyxir (commented):
lgtm

PR #1536, opened 2025-07-31
Switch on new transactions!

seefeldb:
## Summary by cubic
Enabled the new transactions system by default and updated the environment variable to USE_TRANSACTIONS_SHIM for toggling the legacy shim.


ellyxir:
reverting changes made directly here and will do in another PR as this is getting a bit complex

PR #1538, opened 2025-07-31
fix(api): Make Cell.get() return readonly values

seefeldb:
Changes Cell.get() to return Readonly<T> instead of T, preventing mutations of cell values retrieved via get(). Implements Object.freeze() on returned values to enforce immutability at runtime.


ubik2:
This could probably be Immutable<T> instead, which does a deeper transform

ellyxir:
does seem more correct, doesnt it? its a bit more runtime work.

PR #1540, opened 2025-07-31
fix(html): cancel previous DOM element subscriptions on replacement

seefeldb:
- Call cancel() on previous DOM elements when replacing them in keyed children
- Ensure cancel function is always defined (default to noop) to prevent undefined errors
- Properly spread the entire newRendered object when updating keyedChildren map


ellyxir (commented):
lgtm

PR #1543, opened 2025-07-31
Jake/ct 564 shell support iframe recipes

jakedahn:


packages/shell/public/module/charm/sandbox/bootstrap.js
```diff
@@ -0,0 +1,567 @@
+// Import React immediately
```
jsantell, line 1: I *think* this could go in `packages/static`

packages/charm/src/iframe/static.ts
```diff
@@ -31,7 +32,7 @@ export const prefillHtml = `<html>
 ${JSON.stringify(libraries)}
 </script>
 <!-- Bootstrap script that runs first to set up React and utility functions -->
-<script type="module" id="bootstrap" src="/module/charm/sandbox/bootstrap.js"></script>
+<script type="module" id="bootstrap" src="/static/scripts/iframe-bootstrap.js"></script>
```
jakedahn, line 35: `bootstrap.js` is now hosted via the `packages/static` package, so jumble and shell share the same file. Also renamed to `iframe-bootstrap.js` for clarity.

packages/shell/integration/iframe-counter-recipe.tsx
```diff
@@ -0,0 +1,106 @@
+import { h, type JSONSchema, NAME, recipe, UI } from "commontools";
```
jakedahn, line 1: This is a static iframe counter charm that we can use to test whether iframes are working end to end in shell

.github/workflows/deno.yml
```diff
@@ -347,13 +347,6 @@ jobs:
       - name: 🛡️ Disable AppArmor for browser tests
         run: echo 0 | sudo tee /proc/sys/kernel/apparmor_restrict_unprivileged_userns
 
-      - name: 🧪 Run end-to-end jumble integration tests
```
jakedahn, line 350: removing jumble integration tests, as they are now replaced by shell integration tests

PR #1551, opened 2025-08-01
without CTS, our types allow invalid handler

anotherjesse:
you must either specify the jsonschemas or enable (deprecated) proxy mode

handler<E,T>(fn) types say that the state should be of type T, but if the user specifies T has a Cell<...> it will be stripped away by the compiler - and so our runtime won't know - and it was enabling proxy mode

we should only enable proxy mode if the user explicitly specifies.
    

---

## Summary by cubic
Updated the handler function to require either explicit JSON schemas or proxy mode to prevent invalid handler types from being accepted.

- **Bug Fixes**
  - Throws an error if no schema is provided and proxy mode is not enabled.
  - Only enables proxy mode when explicitly specified.

### DIFF
--- a/recipes/simpleValue.tsx
+++ b/recipes/simpleValue.tsx
@@ -1,3 +1,4 @@
+/// <cts-enable />
 import {
   Cell,
   derive,


ellyxir:
@anotherjesse all ready to merge?

PR #1555, opened 2025-08-01
Support $defs style $ref in ContextualFlowControl

ubik2:
Enable $ref following for joinSchema

- Move resolveSchemaRef from traverse to cfc.
- Add resolveSchemaRefOrThrow method
- Changed joinSchema to be static, and expect the joined set to be passed to cfc.lub
- Include ifc flags at the same level as a $ref
- Added minimal $defs test
    

---

## Summary by cubic
Enabled $ref resolution in joinSchema, allowing classification tags to be collected from referenced schemas and supporting $defs references.

- **Refactors**
  - Moved resolveSchemaRef logic from traverse to cfc and added resolveSchemaRefOrThrow.
  - Made joinSchema static and updated its usage to require explicit joined sets.
  - joinSchema now collects ifc flags at the same level as $ref.

- **Tests**
  - Added a test to verify $defs and $ref support in schema handling.


ubik2:
We've got a lot of errors from unsupported $ref links, so I'm going to land this.

PR #1559, opened 2025-08-02
Unify "source" property treatment

ubik2:
When setting the source, always use the `{"/": "baedfoo"}` style link
    

---

## Summary by cubic
Standardized the "source" property to always use the `{"/": "id"}` link format across storage, transactions, and related tests.

- **Refactors**
  - Updated all code paths to serialize and parse the "source" property as an object, not a string.
  - Adjusted tests to expect the new object format.


ellyxir (approved):
lgtm, should fix the error we are seeing on the transaction branch, lets merge!

PR #1562, opened 2025-08-03
convert source to object if its json string

ellyxir:
## Summary by cubic
Updated the storage transaction read logic to always return the "source" path as an object, converting from a JSON string if needed. This ensures consistent API behavior when reading the "source" path.


ellyxir:
not sure if this is the right layer to do this at.

PR #1578, opened 2025-08-05
Fix array index access in ct charm set [CT-731]

bfollington:
## Summary
- Fixes TypeMismatchError when setting array indices directly via `ct charm set`
- Users can now use paths like `path/to/array/0` without getting misleading "is not a record" errors
- Adds comprehensive tests for array index path parsing

## Problem
The `ct charm set` command was failing when users tried to set array indices directly:
```bash
echo '{"data": "value"}' | ct charm set --charm CHARM_ID path/to/array/0
```

This would throw:
```
TypeMismatchError: Cannot access path [users, 0, tags, 1] - [users] is not a record
```

The error was misleading because arrays *are* records in JavaScript.

## Root Cause
The issue was in the `validateParentPath` function in `transaction-shim.ts`. The validation logic was inconsistent:

- **Single-level paths** (pathLength === 1): Correctly checked for `undefined` root values
- **Multi-level paths** (pathLength > 1): Missing the `undefined` check, causing validation to fail when `parentValue` was `undefined`

## Solution
Added the missing undefined check for multi-level paths, making validation consistent across all path lengths. Now both single-level and multi-level paths properly handle cases where the root document doesn't exist.

## Test plan
- [x] Added path parsing tests to verify array index paths work correctly
- [x] Verified all existing tests still pass (67 tests passing, 0 failures)
- [x] Confirmed fix addresses the specific reproduction case from the issue
- [x] Tested that the validation now returns appropriate `NotFoundError` instead of misleading `TypeMismatchError`

---

## Summary by cubic
Fixed ct charm set so users can set array indices directly using paths like path/to/array/0, returning clear errors if the document does not exist.

- **Bug Fixes**
  - Added missing check for undefined root values in multi-level paths.
  - Updated error handling to return NotFoundError instead of TypeMismatchError for missing documents.
  - Added tests to verify correct parsing and handling of array index paths.

### DIFF
--- a/packages/cli/test/charm-cell-operations.test.ts
+++ b/packages/cli/test/charm-cell-operations.test.ts
@@ -80,3 +80,17 @@ describe("parsePath", () => {
     expect(parsePath("0")).toEqual([0]);
   });
 });
+
+describe("Array Index Setting (CT-731)", () => {
+  it("should correctly parse array index paths", () => {
+    // These are the types of paths that should work for array index setting
+    expect(parsePath("items/0")).toEqual(["items", 0]);
+    expect(parsePath("data/users/5/tags/2")).toEqual(["data", "users", 5, "tags", 2]);
+    expect(parsePath("arr/0/0")).toEqual(["arr", 0, 0]);
+  });
+  
+  it("should handle paths that mix objects and arrays", () => {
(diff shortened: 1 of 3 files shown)


bfollington:
red herring

PR #1579, opened 2025-08-05
Add `pattern` example of instantiating and `navigateTo()` on a recipe [CT-735]

bfollington:
## Summary by cubic
Added a new recipe example showing how to instantiate a pattern and use `navigateTo()` to create dynamic pages. Also extended the `ct-input` type definitions to support the `pattern` attribute and related input props.

- **New Features**
  - Added `instantiate-recipe.tsx` with a sample that creates new pages using a message input and `navigateTo()`.
  - Updated `ct-input` types to include `pattern`, validation, and input event props.


bfollington:
Something is wrong here when I call `ct dev`.

```
➜ ./dist/ct dev --show-transformed packages/patterns/instantiate-recipe.tsx
/// <cts-enable />
import { Cell, compileAndRun, derive, h, handler, ifElse, JSONSchema, NAME, navigateTo, recipe, str, UI, } from "commontools";
import Counter from "./counter.tsx";
const InputSchema = {
    type: "object",
    properties: {
        allCharms: {
            type: "array",
            items: {
                type: "object",
                asCell: true,
            },
        },
    },
    required: ["allCharms"],
} as const satisfies JSONSchema;
// Define output schema
const OutputSchema = {
    type: "object",
    properties: {},
    required: [],
} as const satisfies JSONSchema;
const newPage = handler({ type: 'object', properties: { detail: { type: 'object', properties: { message: { type: 'string' } } } } }, { type: 'object' }, (e, _) => {
    const charm = Counter({ value: Math.round(Math.random() * 10) });
    return navigateTo(charm);
});
export default recipe(InputSchema, OutputSchema, ({}) => {
    return {
        [NAME]: "Page Factory",
        [UI]: (<div>
        <ct-message-input name="Add" placeholder="New page" appearance="rounded" onct-send={newPage({})}/>
      </div>),
    };
});

/// <cts-enable />
import { Cell, Default, derive, h, handler, NAME, Opaque, OpaqueRef, recipe, str, UI, JSONSchema } from "commontools";
interface RecipeState {
    value: Default<number, 0>;
}
const increment = handler({
    type: "object",
    additionalProperties: true
} as const satisfies JSONSchema, {
    type: "object",
    properties: {
        value: {
            type: "number",
            asCell: true
        }
    },
    required: ["value"]
} as const satisfies JSONSchema, (_, state) => {
    state.value.set(state.value.get() + 1);
});
const decrement = handler({
    type: "object",
    additionalProperties: true
} as const satisfies JSONSchema, {
    type: "object",
    properties: {
        value: {
            type: "number",
            asCell: true
        }
    },
    required: ["value"]
} as const satisfies JSONSchema, (_, state: {
    value: Cell<number>;
}) => {
    state.value.set(state.value.get() - 1);
});
function previous(value: number) {
    return value - 1;
}
function nth(value: number) {
    if (value === 1) {
        return "1st";
    }
    if (value === 2) {
        return "2nd";
    }
    if (value === 3) {
        return "3rd";
    }
    return `${value}th`;
}
export default recipe({
    type: "object",
    properties: {
        value: {
            type: "number",
            default: 0
        }
    },
    required: ["value"]
} as const satisfies JSONSchema, (state) => {
    return {
        [NAME]: str `Simple counter: ${state.value}`,
        [UI]: (<div>
        <ct-button onClick={decrement(state)}>
          dec to {commontools_1.derive(state.value, _v1 => previous(_v1))}
        </ct-button>
        <span id="counter-result">
          Counter is the {commontools_1.derive(state.value, _v1 => nth(_v1))} number
        </span>
        <ct-button onClick={increment({ value: state.value })}>
          inc to {commontools_1.derive(state.value, _v1 => _v1 + 1)}
        </ct-button>
      </div>),
        value: state.value,
    };
});

ReferenceError: commontools_1 is not defined
    at <CT_INTERNAL>
    at recipe (file:///var/folders/rk/0y09lmqs25719xpmr4_ppgp40000gn/T/deno-compile-ct/packages/runner/src/builder/recipe.ts:112:19)
    at Object.eval (ba4jcavdwyi4nuzndrad3c22jn3dvvl4cqzyx34tp6evo7p5k2fnkzkne/counter.tsx:45:21)
    at <CT_INTERNAL>
    at <CT_INTERNAL>
    at Array.map (<anonymous>)
    at <CT_INTERNAL>
    at <CT_INTERNAL>
    at Array.map (<anonymous>)
    at <CT_INTERNAL>
    at <CT_INTERNAL>
    at <CT_INTERNAL>
    at file:///var/folders/rk/0y09lmqs25719xpmr4_ppgp40000gn/T/deno-compile-ct/packages/js-runtime/runtime/eval-runtime.ts:15:57
    at IsolateInternals.exec (file:///var/folders/rk/0y09lmqs25719xpmr4_ppgp40000gn/T/deno-compile-ct/packages/js-runtime/runtime/eval-runtime.ts:37:14)
    at UnsafeEvalJsValue.invoke (file:///var/folders/rk/0y09lmqs25719xpmr4_ppgp40000gn/T/deno-compile-ct/packages/js-runtime/runtime/eval-runtime.ts:15:35)
    at Engine.process (file:///var/folders/rk/0y09lmqs25719xpmr4_ppgp40000gn/T/deno-compile-ct/packages/runner/src/harness/engine.ts:168:46)
    at async process (file:///var/folders/rk/0y09lmqs25719xpmr4_ppgp40000gn/T/deno-compile-ct/packages/cli/lib/dev.ts:72:28)
    at async Command.actionHandler (file:///var/folders/rk/0y09lmqs25719xpmr4_ppgp40000gn/T/deno-compile-ct/packages/cli/commands/dev.ts:37:31)
    at async Command.execute (https://jsr.io/@cliffy/command/1.0.0-rc.7/command.ts:1940:7)
    at async Command.parseCommand (https://jsr.io/@cliffy/command/1.0.0-rc.7/command.ts:1772:14)
    at async parse (file:///var/folders/rk/0y09lmqs25719xpmr4_ppgp40000gn/T/deno-compile-ct/packages/cli/commands/mod.ts:4:10)
    at async main (file:///var/folders/rk/0y09lmqs25719xpmr4_ppgp40000gn/T/deno-compile-ct/packages/cli/mod.ts:5:5)
    ```

**Comment by bfollington** [comment]
@jsantell what do you think about my solve here with `shadow-dom-utils.ts`? This was the cleanest way I could get the test to actually drill down through 5+ `shadowRoot`s and click / type.

File: packages/patterns/integration/shadow-dom-utils.ts [hunk 1]
```diff
@@ -0,0 +1,170 @@
+/**
+ * Utility functions for navigating through shadow DOM boundaries in tests
+ */
+
+type Page = any; // The Page type from integration utils
```
jsantell, line 5: Can this be `@astral/astral#Page`? would apply to both Page and our `@commontools/integration` wrapper

packages/patterns/integration/shadow-dom-utils.ts
```diff
@@ -0,0 +1,170 @@
...
+    for (let i = 0; i < selectors.length; i++) {
+      const selector = selectors[i];
+      
+      if (!current) {
+        console.error(`Lost element at step ${i}`);
+        return null;
+      }
+      
+      // Query the current context (document, element, or shadow root)
+      let next: Element | null = null;
```
jsantell, line 29: Rather than repeating this `querySelector(path:string[])` interface, wonder if this could be defined as a standalone function that we stringify as an addition arg, and can store that in a content global so that the query can be shared across these utils, not [dissimilar to patching the console](https://github.com/commontoolsinc/labs/blob/9a2487f37ac489b98171cc3e82910f97be80cbab/packages/integration/page.ts#L93-L124)

```ts
const key = "__common_integration_query_selector";
async function registerQuerySelector(page:Page) {
  await page.evaluate((key, fn) => {
   if (globalThis[key]) { return; }
   globalThis[key] = eval(fn);  
  }, { args: [key, querySelector.toString()] });
}

async click(page: Page, path: string[]) {
  await registerQuerySelector(page);
  page.evaluate((key, selector) => {
    const element = globalThis[key](selector);
    if (element) element.click();
  }, { args:[key,path] });
}
```

packages/patterns/integration/shadow-dom-utils.ts
```diff
@@ -0,0 +1,170 @@
...
+    return current;
+  }, { args: [path] });
+}
+
+/**
+ * Clicks an element deep in shadow DOM
+ * @param page - The page object
+ * @param path - Array of selectors to traverse through shadow DOMs
+ */
+export async function clickShadowElement(
```
jsantell, line 69: the "Shadow" prefix on the functions might be unneeded, the selectors can navigate across shadow boundaries but the resulting elements aren't any different than regular (light dom) elements, and can be thought of as an alternative selector engine

jsantell: May be a notable detail, but in the selector path, descending into shadow roots can only be done by specifying a shadow host as a target in a selector

packages/patterns/integration/instantiate-recipe.test.ts
```diff
@@ -0,0 +1,119 @@
...
+
+const CT_MESSAGE_INPUT_BUTTON_PATH = [
+  "x-root-view",
+  "x-app-view",
+  "x-body-view",
+  "x-charm-view",
+  "common-charm",
+  "ct-render",
+  "ct-message-input",
+  "ct-button",
```
jsantell, line 38: Most of this is iterating through the shell components. I see the need for wanting to specify/disambiguate a specific node that cannot be represented as a traditional selector, but Astral pierce selectors could query e.g. the single `ct-render` element, query a `ct-message-input` within, or with a unique attribute, `ct-button[main-button=""]` could be queried from the root without specifying the full path -- are there elements which cannot be selected by the pierce strategy in astral?

packages/patterns/integration/shadow-dom-utils.ts
```diff
@@ -0,0 +1,170 @@
...
+/**
+ * Traverses through shadow DOM boundaries following a path of selectors
+ * @param page - The page object
+ * @param path - Array of selectors to traverse through shadow DOMs
+ * @returns The element at the end of the path, or null if not found
+ */
+export async function traverseShadowDOM(
+  page: Page,
+  path: string[],
+): Promise<any> {
```
jsantell, line 16: What do `Element`s look like converted over the boundary? I don't recall if Astral translates it directly to an `ElementHandle`, but that'd be very helpful if it did -- is this function used?

packages/patterns/integration/instantiate-recipe.test.ts
bfollington: I couldn't ever get the `{strategy: 'piece'}` to actually traverse this path no matter what I tried, it returns `undefined` in almost every case so I gave up and wrote my own solution. 

> `ct-button[main-button=""]`

I'll try this, I don't really like it but I don't like my current solve either so... lesser of two evils?

PR #1584, opened 2025-08-06
ensure there are no undefined values in invocation

ellyxir:
## Summary by cubic
Removed undefined values from the invocation object before sending commands to prevent serialization errors.


ubik2 (approved):
lgtm

PR #1589, opened 2025-08-06
Expect cycles when traversing $ref

ubik2:
Add a flag to indicate whether cycles are expected to CycleTracker. Only log the cycle if they are unexpected (like cell links)
    

---

## Summary by cubic
Added a flag to CycleTracker to control whether cycles are expected during schema traversal. Now, cycle warnings are only logged if cycles are unexpected.


ubik2:
Squashing, since I suspect log messages are the main cause of slowness

PR #1595, opened 2025-08-07
When we encounter a conflict, include the history, so we have a chance of knowing why

ubik2:
## Summary by cubic
When a conflict is detected, the system now includes the full history of related facts to help understand the cause.

- **New Features**
 - Added a recursive query to collect and return the history of fact revisions during conflict detection.
 - Updated the Conflict type to include a history field.


ubik2:
I incorporated this into https://github.com/commontoolsinc/labs/pull/1563, so I'm closing this version

PR #1604, opened 2025-08-11
Add `fetch-data` pattern + test

bfollington:


packages/patterns/fetch-data.tsx
```diff
@@ -0,0 +1,255 @@
...
+  h,
+  handler,
+  NAME,
+  recipe,
+  schema,
+  str,
+  UI,
+} from "commontools";
+
+const model = schema({
```
seefeldb, line 15: Was there a reason this couldn't be done with CTS? Just curious

packages/patterns/fetch-data.tsx
```diff
@@ -0,0 +1,255 @@
...
+  });
+
+  // Fetch repository data
+  const repoData = fetchData<GithubResponse>({
+    url: apiUrl,
+    mode: "json",
+  });
+  const data = repoData.result;
+
+  const validData = derive(data, (data) => {
```
seefeldb, line 192: This makes me wonder: It might be nice to have a default validator that validates against the schema, which itself could just be the typescript type, i.e. here `GithubResponse` in `fetchData<GithubResponse>`. WDYT?

PR #1605, opened 2025-08-11
Nested counter pattern [CT-764]

bfollington:
## Summary by cubic
Added a new nested counter pattern with increment and decrement buttons, including tests for direct operations and UI behavior.

- **New Features**
  - Created a counter component that displays its value and allows incrementing or decrementing.
  - Added a parent component that nests the counter and displays the current value.
  - Implemented integration tests to verify counter state and UI updates.

### DIFF
--- a/packages/patterns/llm.tsx
+++ b/packages/patterns/llm.tsx
@@ -33,6 +33,8 @@ const askQuestion = handler<
 });
 
 export default recipe<LLMTestInput, LLMTestResult>("LLM Test", ({ title }) => {
+  // It is possible to make inline cells like this, but always consider whether it should just be part of the argument cell.
+  // These cells are effectively 'hidden state' from other recipes
   const question = cell<string>("");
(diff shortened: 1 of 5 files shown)


packages/patterns/llm.tsx
```diff
@@ -33,6 +33,8 @@ const askQuestion = handler<
 });
 
 export default recipe<LLMTestInput, LLMTestResult>("LLM Test", ({ title }) => {
+  // It is possible to make inline cells like this, but always consider whether it should just be part of the argument cell.
+  // These cells are effectively 'hidden state' from other recipes
```
seefeldb, line 37: +1

i wonder whether we could have even stronger guidelines here. the only good reason i can imagine is for something like an importer that wants to tightly manage this state, including cells that track e.g. last imported cursors, etc.

so maybe we should advice even stronger against this?

PR #1620, opened 2025-08-12
fix(runner): pass synced to new cells

seefeldb:


packages/runner/src/schema.ts
```diff
@@ -396,10 +396,12 @@ export function validateAndTransform(
             rootSchema,
           },
           tx,
+          false,
+          synced,
```
ubik2, line 400: This is basically saying that if the cell being linked from is synced, the cell it links to should be synced as well (since the schema that brought us in should have caused us to load that too), right?

PR #1625, opened 2025-08-13
added definitions for heap nursery and cache

ellyxir:


ubik2 (approved):
Some minor notes that I don't think need to be incorporated:
- We really aren't using the IndexedDB portion anymore, since all our queries are the schema queries. This would take a bit of work to get to a state where it's useful, but should be on the roadmap.
- Nursery eviction is a bit more complicated. Conflicts will also purge the nursery, and we may get an update that we expected from the server, but still have changes built on top of those changes, in which case, we keep the nursery copy (I think you covered this pretty well in the evict description).
Overall, quite useful.

ellyxir:
> Some minor notes that I don't think need to be incorporated:
> 
>     * We really aren't using the IndexedDB portion anymore, since all our queries are the schema queries. This would take a bit of work to get to a state where it's useful, but should be on the roadmap.
> 
>     * Nursery eviction is a bit more complicated. Conflicts will also purge the nursery, and we may get an update that we expected from the server, but still have changes built on top of those changes, in which case, we keep the nursery copy (I think you covered this pretty well in the evict description).
>       Overall, quite useful.

got it, incorporated into entries

PR #1627, opened 2025-08-13
linkedlist within a cell as an example

ellyxir:
## Summary by cubic
Added a simple linked list example using a reactive cell, allowing users to add new nodes and view the list in real time.

- **New Features**
 - Users can add messages as nodes to the linked list.
 - The linked list is displayed with each update.


ellyxir:
still has fixmes, i'll add a new ticket for those

PR #1628, opened 2025-08-13
chore: streamline setting and waiting for application state to reduce time sensitive waits

jsantell:


packages/integration/shell-utils.ts
```diff
@@ -185,17 +236,7 @@ export class ShellIntegration {
...
-
-    while (Date.now() - startTime < timeout) {
-      const handle = await page.$(selector);
-      if (handle) return handle;
-      await sleep(100);
-    }
-
-    throw new Error(`Timeout waiting for selector: ${selector}`);
+  private checkIsOk() {
+    if (!this.page) throw new Error("Page not initialized.");
```
ubik2, line 240: Is this correct? I would expect the test for this.page to always return true (it's the function, which is defined), while this._page might be undefined.

PR #1630, opened 2025-08-13
perf(runner): optimize link resolution to minimize storage reads

seefeldb:
This refactor reduces the number of storage reads performed during link resolution and, as a result, lowers scheduler/reactivity overhead.

Key changes
- Single sigil probe at the full path; on NotFound fall back to the minimal parent prefix derived from NotFoundError.path (instead of scanning all prefixes).
- Only read the full value when a sigil/legacy link is confirmed; avoid speculative sibling reads.
- Compose remainingPath once and slice schema only when needed.
- Lower the iteration ceiling (100) now that it's just one per doc.

Impact
- Fewer reads recorded in the reactivity log (compact to [] + exact targets), reducing subscriptions and wake-ups.
- Noticeably less churn in alias-in-aliased-path and nested asCell paths; schema and link-resolution tests pass with fewer reads.

### DIFF
--- a/packages/runner/src/link-resolution.ts
+++ b/packages/runner/src/link-resolution.ts
@@ -6,10 +6,7 @@ import {
   type NormalizedFullLink,
   parseLink,
 } from "./link-utils.ts";
-import type {
-  IExtendedStorageTransaction,
-  MemoryAddressPathComponent,
-} from "./storage/interface.ts";
+import type { IExtendedStorageTransaction } from "./storage/interface.ts";
 import { ContextualFlowControl } from "./cfc.ts";
 
 const logger = getLogger("link-resolution");
@@ -26,7 +23,7 @@ export type ResolvedFullLink = NormalizedFullLink & {
   [resolvedFullLinkBrand]: true;
 };
 
-const MAX_PATH_RESOLUTION_LENGTH = 1000;
+const MAX_PATH_RESOLUTION_LENGTH = 100;
 
 /**
  * Resolves a document path with support for links inside documents.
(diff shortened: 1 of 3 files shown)


packages/runner/src/link-resolution.ts
```diff
@@ -69,83 +66,115 @@ export function resolveLink(
...
+      const lastValid = sigilProbe.error.path?.slice(); // undefined => doc missing
 
-    // Detect cycles. Only have to do this at top of path, since link folloowing
-    // will always go through this at least once.
-    if (traversedPath.length === 0) {
-      const key = JSON.stringify([last.space, last.id, remainingPath]);
-      if (seen.has(key)) {
-        logger.warn(`Link cycle detected ${key}`);
-        return emptyResolvedFullLink; // = return link to empty document
+      if (lastValid) {
```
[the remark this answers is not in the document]
seefeldb: empty path is correct

packages/runner/src/link-resolution.ts
```diff
@@ -69,83 +66,115 @@ export function resolveLink(
...
+        lastValid?.shift(); // remove `value` prefix
```
ubik2, line 109: it seems like the error in the NotFoundError will already exclude the "value" portion, since only the part after value (`rest`) is passed to validateParentPath.

seefeldb: Oh, interesting, it didn't work without the pop. Maybe the shim and the real implementation differ here? That could explain some other strange things!

ubik2: Also, since it may not have been clear, I'm using real transactions (the default), and haven't tried anything with the shim for a while.
We should probably move the transaction-shim.ts code that's not shim based to a different file.

seefeldb: turns out the tx code was wrong. so i fixed it and removed the pop. (also just for real transactions)

ubik2 (approved):

ubik2:
lgtm. still have 22k NotFound in my test, but editing the last item results in 450 NotFound instead of ~5k, so 10x better.

PR #1633, opened 2025-08-14
syntax question: thoughts on inline types?

seefeldb:
I wondered whether the extensive type definitions at the top are a left-over from directly translating JSON Schema based recipes and so here I'm reducing global definitions to core concepts, e.g. `Item` and otherwise inline types to get closer to how one would e.g. define a function with several parameters.

The event type should probably actually be imported, once we define standard types for all events, so I included both variants.

There's also a mix of providing the type in the generic:
```recipe<{ items: Default<Items[], []> }("..", ({ items }) => ...```
vs inline
```handler((_, { value }: { value: Cell<number> }) => ...```

the latter seems to not currently work with recipes, at least when `Default` is involved.

I think we should support all variations, but my question is whether there is one style or another that we should prefer. What should go into the style guide, so that all example patterns are consistent?
    

---

## Summary by cubic
Switched recipe and handler definitions to use inline TypeScript types instead of global interface declarations for function parameters and state.

- **Refactors**
 - Removed separate interface/type declarations for state and parameters.
 - Used inline types directly in handler and recipe function signatures.
 - Made patterns more consistent and easier to read.

### DIFF
--- a/packages/patterns/array-in-cell-ast-nocomponents.tsx
+++ b/packages/patterns/array-in-cell-ast-nocomponents.tsx
@@ -15,28 +15,18 @@ interface Item {
   text: Default<string, "">;
 }
 
-interface InputSchema {
-  title: Default<string, "untitled">;
-  items: Default<Item[], []>;
-}
-
-type InputEventType = {
-  detail: {
-    message: string;
-  };
-};
-
-interface ListState {
-  items: Cell<Item[]>;
-}
-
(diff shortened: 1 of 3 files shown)


jakedahn:
I have a slight personal preference for the not-inline type interface definition approach.

It's easier for me to eyeball parse a set of type interfaces at the top of a file, and then when I use them inside of a function signature, my IDE makes it easy to see what they actually are with option+hover over the type name.


image

PR #1643, opened 2025-08-14
When running the map function, instead of just setting the child elements, also set the full array value.

ubik2:
When setting an array that is shorter, include a change to the array's length that shortens it.

### DIFF
--- a/packages/runner/src/builtins/map.ts
+++ b/packages/runner/src/builtins/map.ts
@@ -97,6 +97,9 @@ export function map(
       throw new Error("map currently only supports arrays");
     }
 
+    const newArrayValue = [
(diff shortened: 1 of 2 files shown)


packages/runner/src/builtins/map.ts
```diff
@@ -97,6 +97,9 @@ export function map(
       throw new Error("map currently only supports arrays");
     }
 
+    const newArrayValue = [
+      ...resultWithLog.getRaw()!.slice(0, initializedUpTo),
```
seefeldb, line 101: you have .getRaw() here, but .set() below. This only currently works because we're allowing sigil links to be set, but I have a ticket to fix that (it's a security violation if non-privileged code can set arbitrary links)

PR #1647, opened 2025-08-15
Remove useSchemaQueries flag, since we essentially require it at this point.

ubik2:
## Summary by cubic
Removed the useSchemaQueries flag from storage settings and related code, since schema queries are now always required. This cleans up configuration and simplifies the code.

- **Refactors**
  - Removed useSchemaQueries properties and checks from cache, interface, and scripts.


ellyxir (approved):
lgtm

PR #1655, opened 2025-08-19
Add `ct-checkbox` + `ifElse()` pattern

bfollington:


packages/patterns/integration/ct-checkbox-simple.test.ts
```diff
@@ -0,0 +1,142 @@
...
+  });
+
+  it("should show disabled content initially", async () => {
+    const page = shell.page();
+
+    // Wait for component to stabilize
+    await sleep(1000);
+
+    // Initially, feature should be disabled
+    const disabledContent = await page.$("[data-testid='disabled-content']", {
```
jsantell, line 61: Using `page.waitForSelector(..)` here has a built in wait loop -- unless `[data-testid='disabled-content']` is already in the DOM, can avoid the explicit sleep above and use `waitForSelector` (which either returns the element, or throws, so we don't need assertions to guard against `undefined`) -- I think this is the only potential racey spot AFAICT

packages/patterns/ct-checkbox-simple.tsx
```diff
@@ -0,0 +1,61 @@
...
+    const toggleHandler = toggle({ enabled });
+
+    return {
+      [NAME]: "Checkbox Demo",
+      [UI]: (
+        <common-vstack gap="md" style="padding: 2rem; max-width: 400px;">
+          <h3>Simple ct-checkbox + ifElse Demo</h3>
+
+          <ct-checkbox
+            $checked={enabled}
```
seefeldb, line 41: oh, the $ in $checked would imply to me that we don't need the event handler to change the state?

bfollington: I had intended this as a demo of there being two ways, but on reflection I think I should split the recipe into two.

PR #1660, opened 2025-08-19
Clean up `patterns` test code

bfollington:
Prefer astral's patterns where possible, remove arbitrary waits.

I also added a way to skip tests _only_ in CI, pretty crude, happy to discuss the approach @jsantell. Cleaning up use of sleep() and preferring waitFor functions saves ~10s in test duration. Not huge but these gains compound as we fill out the set.

    

---

## Summary by cubic
Cleaned up patterns integration tests by using Astral’s wait patterns and removing arbitrary sleeps. Tests run faster and are less flaky.

- **Refactors**
  - Added Page.waitForFunction passthrough to @astral/astral.
  - Replaced sleep-based waits with waitForFunction in ct-checkbox-simple and ct-list tests.
  - Reduced remaining sleeps to short waits only where necessary.

### DIFF
--- a/packages/integration/page.ts
+++ b/packages/integration/page.ts
@@ -208,6 +208,15 @@ export class Page extends EventTarget {
     return await this.page!.waitForSelector(selector, options);
   }
 
+  // Passthru of `@astral/astral`'s `Page#waitForFunction`
+  async waitForFunction<T, R extends readonly unknown[]>(
+    func: EvaluateFunction<T, R>,
+    evaluateOptions?: EvaluateOptions<R>,
+  ): Promise<void> {
+    this.checkIsOk();
+    await this.page!.waitForFunction(func, evaluateOptions);
+  }
+
   // Passthru of `@astral/astral`'s `Page#$`
   async $(
     selector: string,
--- a/packages/patterns/integration/counter-operations.test.ts
+++ b/packages/patterns/integration/counter-operations.test.ts
@@ -65,9 +65,6 @@ describe("counter direct operations test", () => {
     assertEquals(value, 0);
   });
 
-  // Bug Reproduction for CT-753: Live updates across sessions don't work currently
-  // The browser has its own runtime/session that doesn't receive
-  // live updates from our test CharmManager's operations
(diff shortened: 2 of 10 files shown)


packages/integration/page.ts
```diff
@@ -208,6 +208,15 @@ export class Page extends EventTarget {
     return await this.page!.waitForSelector(selector, options);
   }
 
+  // Passthru of `@astral/astral`'s `Page#waitForFunction`
+  async waitForFunction<T, R extends readonly unknown[]>(
```
jsantell, line 212: This is a nice helper, TIL

packages/patterns/integration/counter-operations.test.ts
```diff
@@ -81,8 +81,8 @@ describe("counter direct operations test", () => {
     console.log("Setting counter value to 42 via direct operation");
     await setCharmResult(manager, charmId, ["value"], 42);
 
-    // Wait for the update to propagate
-    await sleep(3000);
+    // Wait for the update to propagate (reduced from 3000ms)
```
jsantell, line 84: I don't understand why LLMs write comments on their changes that lose meaning outside of a PR

packages/patterns/integration/llm.test.ts
```diff
@@ -9,6 +9,10 @@ import { Identity } from "@commontools/identity";
 
 const { API_URL, FRONTEND_URL, SPACE_NAME } = env;
 
+// LLM tests are skipped in CI until we handle llm() calls properly in CI environments.
+// This requires either:
+// 1. Adding a flag to enable LLM tests in CI with proper API keys
+// 2. Resurrecting the LLM cache functionality from toolshed
 describe("LLM pattern test", () => {
```
jsantell, line 16: Seems reasonable!

jsantell (approved):

packages/patterns/integration/counter-operations.test.ts
bfollington: Yah, good catch. Let me fix it.

We need a `NO_COMMENT_EDITS` flag for LLM changes.

PR #1661, opened 2025-08-20
demo user login with individual sessions pattern via chat

ellyxir:
## Summary by cubic
Adds a demo chat app that shows per-user sessions by spawning individual charm instances that share a single messages state. Addresses Linear CT-807 by implementing a two-tier pattern: a shared state container and user-specific sessions via navigateTo.

- New Features
  - Main Chat State Container recipe that holds messages as shared reactive state.
  - UserSession recipe that generates a unique userId and lets the user send messages.
  - Session creation handler (proxy-enabled) that navigates to a new session charm while passing the shared messages cell.
  - Message sending handler that appends { userId, message, timestamp } to the shared list.
  - Documentation (ct-807.md) covering architecture, types, implementation notes, and testing steps.


recipes/chat-main.tsx
```diff
@@ -0,0 +1,255 @@
...
+// Session recipe result (typed): return extra fields (e.g., userId/username)
+// alongside [NAME] and [UI]. CTS will carry these in the recipe result schema.
+type UserSessionResult = {
+  userId: string;
+  username: string;
+};
+
+// Helper to generate a user id when first needed
+function generateUserId(): string {
+  const n = Math.floor(Math.random() * 10000);
```
seefeldb, line 58: if not now, this might eventually always create the same number, as we'll have to use a fixed seed.

recipes/chat-main.tsx
```diff
@@ -0,0 +1,255 @@
...
+          </div>
+        </div>
+      ),
+      userId: derive(userId, (k) => k || "") as any,
+      username: username,
+    };
+  },
+);
+
+// Handler to create a new user session. We use `{ proxy: true }` so the handler
```
seefeldb, line 220: Instead of that, you could set the type to `Cell<...>`, then you also get a handle on the reactive reference.

recipes/chat-main.tsx
```diff
@@ -0,0 +1,204 @@
...
+  {
+    messages: Cell<ChatMessage[]>;
+    user: Cell<User>;
+  }
+>((event, { messages, user }) => {
+  const text = event.detail?.message?.trim();
+  if (!text) return;
+
+  messages.push({
+    author: user.get(),
```
seefeldb, line 79: I think this should work without `.get()`. If not, that's a bug.

recipes/chat-main.tsx
```diff
@@ -0,0 +1,255 @@
...
+  username: string;
+};
+
+// Helper to generate a user id when first needed
+function generateUserId(): string {
+  const n = Math.floor(Math.random() * 10000);
+  return `id${n.toString().padStart(4, "0")}`;
+}
+
+// Helper to get or initialize userId from a Cell
```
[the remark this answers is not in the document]
ellyxir: great idea, done

recipes/chat-main.tsx
ellyxir: removed userid but good to know

recipes/chat-main.tsx
```diff
@@ -0,0 +1,204 @@
...
+// Handler to set/update the username (local-only field)
+const setUsername = handler<
+  InputEventType,
+  {
+    user: Cell<User>;
+  }
+>((event, { user }) => {
+  const name = (event.detail?.message ?? "").trim();
+  // Update user cell with new value
+  user.set({ name: name });
```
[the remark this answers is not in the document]
ellyxir: yup! was lazy and trying to be as simple as possible to start hah

PR #1665, opened 2025-08-21
Implement reflection pattern (`allCharms.tsx`)

bfollington:
- **Implement `allCharms.tsx` pattern**
- **Restore `ignore` flag**


bfollington:
If we fix can remove the `derive()` wrapper around the main `.map` in this pattern, it will also close [CT-805]

bfollington:
When I drop the `derive` and just use `.map` I get different results **but only in a multi-tab scenario**.

Reproduction steps:

1. Use `/explore-recipe` in `labs` to deploy a `counter.tsx`, `charm-list-derive.tsx` and `charm-list.tsx` linking the `allCharms` inputs to the wellknown space charm list cell (Claude knows what this means)

Loading these in the browser and for **both** charms, you'll something similar to:

image

(which is correct).

The single viewer case behaves fine in both cases, but if you open the same URL in a second tab/window the `derive` case behaves but the `.map` case throws conflict errors and becomes:

`charm-list.tsx`: image

It never recovers afterwards.

This is relevant for the tests because if I watch the test execution it seems to trigger the same issue.

ellyxir:
test

PR #1672, opened 2025-08-21
`compile-and-run.tsx` pattern

bfollington:
## Summary by cubic
Adds a new compile-and-run pattern that lets users edit TSX, see live compile errors, and navigate to the compiled result. Delivers on CT-817 by enabling on-the-fly code compilation within the app.

- **New Features**
  - New Compiler pattern with a built-in code editor and default counter example.
  - Live compile via compileAndRun with inline error reporting.
  - Navigate button enabled only when compilation succeeds; routes to the generated charm.
  - Handlers: updateCode for editor changes, visit to compile and navigate.

### DIFF
--- a/packages/patterns/compile-and-run.tsx
+++ b/packages/patterns/compile-and-run.tsx
@@ -30,7 +30,7 @@ const updateCode = handler<
 );
 
 const visit = handler<
-  {},
+  Record<string, never>,
   { result: any }
 >(
   (e, state) => {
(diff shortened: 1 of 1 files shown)


bfollington:
These tests would pass if not for the (seemingly ignorable) ConflictError thrown:

> 500 Transaction failed {"name":"ConflictError","stack":"ConflictError: The application/json of...

I can get around this by patching our 'detect errors' code to ignore these errors explicitly:

```ts
this.#page.addEventListener("console", (e: ConsoleEvent) => {
  if (e.detail.type === "error") {
    this.#errorLogs.push(e.detail.text);
    // Ignore transaction conflict errors as they don't affect functionality
    if (
      !e.detail.text.includes("Transaction failed") &&
      !e.detail.text.includes("ConflictError")
    ) {
      this.#errorLogs.push(e.detail.text);
    }
  }
  pipeConsole(e);
});
```

but this seems like a hack + our discussions this week sounded like we expected these to disappear shortly?

PR #1676, opened 2025-08-22
`list-operations.tsx` pattern (includes `[ID]`)

bfollington:
- **`list-operations.tsx` showing how to use `[ID]`**
- **Fix double dependency on `npm:ai`**

    

---

## Summary by cubic
Adds a list-operations.tsx pattern demonstrating how to use [ID] for stable list operations and derived views in commontools. Also consolidates and updates the ai dependency to ^4.3.10.

- **New Features**
  - Added list-operations.tsx showing [ID]-based list items to avoid alignment bugs on insert/delete/shuffle.
  - Includes handlers: reset, delete first/last/all, insert at start/end, shuffle.
  - Demonstrates derive and lift for filter/concat/reduce and a lower-case computed view, with simple UI outputs.

- **Dependencies**
  - Removed duplicate ai import and upgraded to npm:ai ^4.3.10.


packages/patterns/list-operations.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+interface ListOutput extends ListInput {}
+
+const typeTest = handler((_, state: { a: Cell<Item[]>, b: readonly Item[], c: readonly Cell<Item>[], d: Cell<Cell<Item>[]> }) => {
+  const { a, b, c, d } = state;
+  console.log({ a, b, c, d })
+});
+
+const resetList = handler((_, state: { items: Cell<Item[]> }) => {
+  state.items.set([{ [ID]: Math.random(), title: "A" }, { [ID]: Math.random(), title: "B" }, { [ID]: Math.random(), title: "C" }, { [ID]: Math.random(), title: "D" }]);
```
seefeldb, line 35: You don't have the use Math.random here, the event context is already unique and ID have to be unique only per context

In a lift the context is stable across calls, but to make calls idempotent you want predictable ID is well.

So I'd rather use some predictable ID as example here as well.

(And FWIW, if you .push those, then no ID is necessary. I want to make that true for .set as well)

packages/patterns/list-operations.tsx
```diff
@@ -0,0 +1,157 @@
...
+    // We do not just have top-level support on Cell<T[]> for the major array operations.
+    // However, performing them on a ProxyObject (such as within a derive, or calling .get() on a Cell in a handler) will work as expected.
+    // caveat: behaviour is only guaranteed to be correct for all operations IF the items include an [ID] property.
+      // excluding the [ID] in this recipe leads to item alignment bugs when insertig or removing from items at the FRONT of an array
+    const itemsLessThanB = derive(items, (items) => items.filter((item) => item.title < "B"));
+    const extendedItems = derive(items, (items) => items.concat([{ [ID]: Math.random(), title: "E" }, { [ID]: Math.random(), title: "F" }]));
+    const combinedItems = derive(items, (items) => items.reduce((acc: string, item: Item) => acc += item.title, ''));
+
+    // A good use of lift() is to avoid repeatedly writing derive() inline in the JSX
+    const show = lift(printList);
```
seefeldb, line 87: I think style-wise it would be better to apply the lift where printList is declared, kind of a analogous to handler

packages/patterns/integration/list-operations.test.ts
```diff
@@ -0,0 +1,253 @@
...
+    
+    // Wait for the main list display to appear
+    await page.waitForSelector("#main-list", { strategy: "pierce" });
+    await sleep(500); // Give component time to fully initialize
+    
+    // Click reset to populate with initial data
+    const resetBtn = await page.$("#reset-demo", { strategy: "pierce" });
+    assert(resetBtn, "Should find reset button");
+    await resetBtn.click();
+    await sleep(300); // Wait for the reset operation to complete
```
jsantell, line 52: now with the `waitFor` function, can use that instead of sleeping here (e.g. wait for `#main-list` with the expected text)

PR #1678, opened 2025-08-24
Remove most of the shim related code

ubik2:
- Fix non-shim transaction return value when writing to an invalid path to be NotFound instead of TypeMismatchError
- Remove deprecated Read/Write types
- Remove lots of code that isn't needed anymore
- Removed most of the transaction-shim classes. Renamed file to extended-storage-transaction.ts, since that is the only remaining class.
- Removed LegacyDocCellLInk type entirely. CellLinks can not be DocImpl or LegacyDocCellLinks anymore.
- There's still a few places we access the DocumentMap and DocImpl, but almost all of them are gone.
- Update Cell cellLink description to indicate that it's a SigilLink
- When passing a cell as the cause, pass the entityId instead of the DocImpl. The existing DocImpl had a toJSON method that returned the entityId.
- Remove ShimStorageManager 
- Fix non-shim transaction return value when writing to an invalid path to be NotFound instead of TypeMismatchError
- Remove deprecated Read/Write types
- Remove lots of code that isn't needed anymore

In this pass, I'm not doing the last bit, which is to remove Storage (move some bits to StorageManager), DocumentMap, and DocImpl.
I want to make sure that this change doesn't introduce any unexpected behavior first.

## Summary by cubic
Removed the storage transaction shim and all legacy DocImpl-based links, consolidating on SigilLink and real storage transactions. This simplifies the API, reduces dead code, and standardizes IDs passed through the system.

- **Refactors**
  - Removed ShimStorageManager and transaction-shim; added extended-storage-transaction for shared transaction utilities.
  - Dropped DocImpl exports/usage and LegacyDocCellLink support; link-utils no longer treats docs as links.
  - Updated CharmManager to follow source via cell.getSourceCell(); runner now passes entityId as parent/cause instead of DocImpl.
  - Pruned doc-map and doc code; removed entityIdStr and other unused helpers.
  - Removed deprecated Read/Write types; use IAttestation.
  - Behavior change: writing to an invalid path now returns NotFound (was TypeMismatchError). Attestation write errors if the read resolves to undefined.

- **Migration**
  - Replace LegacyDocCellLink and any DocImpl-based links with SigilLink or Cell.
  - Use cell.getSourceCell()?.entityId instead of getDoc() or DocImpl.toJSON() where a parent/cause is needed.
  - Remove checks for runtime.storage.shim and any usage of ShimStorageManager.
  - Replace entityIdStr with getEntityId(...) or use the entityId directly.
  - Update code/tests that expected isLink(doc) to false.
  - Expect NotFound on invalid write paths; switch from Read/Write types to IAttestation.


docs/future-tasks/sigil.md
seefeldb, line 1: let's move that into docs/specs/data-model/ or so?

PR #1686, opened 2025-08-25
AST transformer handling intersection types, first rev, as per CT-762

mathpirate:
Obviously since this is my first PR there might be any number of things wrong with it... :) but I'd love feedback!
    

---

## Summary by cubic
Adds intersection type support to the TypeScript→JSON Schema transformer used by handlers. Safely merges object-like intersections and falls back to a permissive schema with a comment for unsupported patterns, addressing CT-762.

- **New Features**
  - Handle A & B & ... intersections in schema generation.
  - Merge when all parts are object-like and have no index/call/construct signatures.
  - Otherwise emit { type: "object", additionalProperties: true, $comment } explaining why.
  - Add tests for a supported merge and an index-signature fallback.
  - Update array-with-remove pattern to use ListState & { index: number }.


mathpirate:
ellyse told me on discord that this looks good and i should squash-merge it, so doing that now

PR #1691, opened 2025-08-26
Use env to set log level

ubik2:
- When creating a logger, if no level threshold is passed in the options, check the env
- In cache.ts, don't specify a level threshold (so we default to env). Also downgrade the bad ConflictError to `warn`, since things still work.
- In deno.json for shell, set the LOG_LEVEL to `warn`, so that `info` messages won't be logged

### DIFF
--- a/packages/shell/deno.json
+++ b/packages/shell/deno.json
@@ -6,7 +6,7 @@
     "serve": "deno run -A ../felt/cli.ts serve .",
     "dev": "API_URL=https://toolshed.saga-castor.ts.net deno run -A ../felt/cli.ts dev .",
     "dev-local": "API_URL=http://localhost:8000 deno run -A ../felt/cli.ts dev .",
-    "integration": "deno test -A ./integration/*.test.ts",
+    "integration": "LOG_LEVEL=warn deno test -A ./integration/*.test.ts",
     "test": "deno test test/*.test.ts"
   },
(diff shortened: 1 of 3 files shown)


ellyxir:
yeah i don't know either :) sounds good to me

ellyxir (approved):

packages/shell/deno.json
```diff
@@ -6,7 +6,7 @@
     "serve": "deno run -A ../felt/cli.ts serve .",
     "dev": "API_URL=https://toolshed.saga-castor.ts.net deno run -A ../felt/cli.ts dev .",
     "dev-local": "API_URL=http://localhost:8000 deno run -A ../felt/cli.ts dev .",
-    "integration": "deno test -A ./integration/*.test.ts",
+    "integration": "LOG_LEVEL=warn deno test -A ./integration/*.test.ts",
```
jsantell, line 9: This is but one integration test (shell, patterns, cli, runner), I don't think anything unique to shell integration tests here

ubik2: - **patterns** This one has a lot of real ConflictErrors, so the log is still littered. I need to investigate. The llm.tsx in particular exposes the bug with undefined in merkle refs.
- **runner** We're not currently getting any errors here, but I can add the env flag.
- **cli** Also no errors, but I can add the env flag if desired.

PR #1693, opened 2025-08-26
runner/scheduler: ensure re-queued events execute after commit failure

seefeldb:
Summary:
- Add event handling test that verifies retries when commit fails
- Use deterministic abort on first 5 attempts to exercise retry path
- Expect attempts == 6 (initial + DEFAULT_RETRIES) and event fully processed
- Ensure prompt processing by scheduling execution after requeue on commit error

Details:
- packages/runner/test/scheduler.test.ts:
  - New test "should retry event handler when commit fails, up to retries count"
  - Abort transaction for first 5 handler runs, perform write on final run
  - Queue a single event and wait until idle twice to drain re-queued work

- packages/runner/src/scheduler.ts:
  - After tx.commit resolves with error and retries remain, unshift event back and call queueExecution to kick the loop

Rationale:
- Without rescheduling after re-queue, a commit that resolves after the current cycle may leave the event sitting until another trigger
- The new test guards the retry semantics and avoids flakiness by not depending on timing or external conflicts

Test plan:
- In packages/runner: deno task test
- New test passes; no new lints; 1 unrelated pre-existing recipes test failure remains

### DIFF
--- a/docs/future-tasks/unified-storage-stack.md
+++ b/docs/future-tasks/unified-storage-stack.md
@@ -154,13 +154,13 @@ This plan should be entirely incremental and can be rolled out step by step.
   - [x] In fact automatically start syncing already, as this will be the future
         behavior. This is just for await.
   - [ ] Add option to only await a locally cached version if available.
-- [ ] Add path-dependent listeners to memory: A helper on `Storage`, that given
+- [x] Add path-dependent listeners to memory: A helper on `Storage`, that given
       a `TX` calls a callback _once_ on future changes on what was read during
       the transaction (observing only changes affecting the read path). Make it
       cancelable (the scheduler will e.g. cancel this before executing the
       action again). First design the API.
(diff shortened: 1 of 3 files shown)


seefeldb:
@ubik2 in the second commit i fixed a test that was previously failing to commit data, and i'm noticing it takes a second to run and produces a lot of cycle warnings that come from within storage. low priority, but maybe a hint to other issues you are looking at?

ubik2 (approved):

ubik2:
> @ubik2 in the second commit i fixed a test that was previously failing to commit data, and i'm noticing it takes a second to run and produces a lot of cycle warnings that come from within storage. low priority, but maybe a hint to other issues you are looking at?

It was slow after the second commit?
There's a good chance (after the second commit) I'll warn on the CycleTracker, but this should be benign. Performance issues there  are more surprising.

PR #1694, opened 2025-08-27
Add `chat.tsx`

bfollington:
## Summary by cubic
Adds a reusable chat pattern (chat.tsx) that renders a simple multi-turn conversation with an LLM. Includes a message input, chat history, and a pending indicator for responses.

- New Features
  - New recipe "LLM Test" with inputs: title and chat history.
  - Integrates llm() with a concise system prompt and uses chat messages as context.
  - askQuestion handler appends the last LLM reply (if any) and the new user message, then re-runs the model.

https://github.com/user-attachments/assets/fe703980-a97f-4b98-94b1-9cdf390f9865


packages/patterns/chat.tsx
```diff
@@ -66,31 +66,35 @@ export default recipe<LLMTestInput, LLMTestResult>(
...
-                  <strong>{msg.role}:</strong> {msg.content}
-                </li>
+                <ct-chat-message
+                  key={idx}
+                  role={msg.role}
+                  content={msg.content}
+                />
               );
             })}
             {derive(llmResponse.pending, (pending) =>
```
seefeldb, line 79: so I _think_ this could also work as

<ct-chat-message
  role="assistant"
  content={ifElse(llmResponse.partial, llmResponse.partial, "...")}
/>

(and no need to pay attention to pending here, as partial also goes through undefined -> partial -> full result)

packages/runner/src/builtins/llm.ts
```diff
@@ -142,6 +142,9 @@ export function llm(
 
     const updatePartial = (text: string) => {
       if (thisRun != currentRun) return;
+      const status = tx.status();
+      if (status.status !== "ready") return;
+
       partialWithLog.set(text);
```
bfollington, line 148: Add a note about using an async tx here in the future

PR #1702, opened 2025-08-28
Speculative fix to address stray promise

ubik2:
## Summary by cubic
Register the cancel handle for fetchData requests so in-flight fetches are canceled on re-run/dispose, preventing stray promises and stale writes. Also rename _addCancel to addCancel to match the cancel API.


jsantell:
This looks like the right approach to me, if these cancels are cancelled while the runtime disposes, for both llm and fetchdata builtins. Along with #1705, we should have full control over the runtime executions in our current intermittent test cases

PR #1714, opened 2025-09-03
Encapsulate `messages` in `llmDialog` node

bfollington:
- **Encapsulate `messages` within `llmDialog`**

Not sure if this is the right way to do it, but the goal is to avoid the user passing `messages` in the first place. I think my reactivity loop is broken because I either end up with no response or an infinite loop so far trying to do this.

### DIFF
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -284,6 +284,12 @@ export interface BuiltInLLMState<T> {
   result?: T;
   partial?: string;
   error: unknown;
+}
+
+export interface BuiltInLLMDialogState {
+  pending: boolean;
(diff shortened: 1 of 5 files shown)


seefeldb (approved):
This looks good.

If you also wanted to support passing in messages, then you could
- read that from the inputs and use it instead of the messages cell you create
- set the created messages cell to link to the passed in one (just call .set with the passed in one as parameter)
  - and thus if no messages are passed use .getRaw to check whether the one you created is an array and if not set it to one (regular set is fine)... That's in case the caller changes this after initialization, but it's somewhat undefined what the right thing to do then is.

PR #1719, opened 2025-09-03
fix(scheduler,html): avoid dropping subscriptions; unsubscribe only pending actions

seefeldb:
…ending actions

- runner/scheduler: During execution, stop clearing the entire pending set and unsubscribing everything up front. Instead, for each action in the run order, only proceed if it is still pending, then remove it from pending and unsubscribe just that action before running it. This prevents unsubscribing actions that have since been removed, avoids lost dependencies, and reduces unnecessary resubscribe cycles.
- html/render: When replacing a keyed child's DOM node, do not cancel the child; keep its subscriptions and only update the stored node reference. This avoids dropping listeners/state during keyed re-renders.

This fixes intermittent missed updates and event storms caused by over-eager cancellation/unsubscribe in both the scheduler and keyed DOM updates.
    

---

## Summary by cubic
Prevents dropped subscriptions during scheduling and keyed DOM replacements. Reduces missed updates and event storms by unsubscribing only pending actions and preserving child subscriptions.

- **Bug Fixes**
  - Scheduler: stop clearing all pending and unsubscribing upfront; for each action, if still pending, remove from pending, unsubscribe it, then run. Avoids unnecessary resubscribes and lost dependencies.
  - HTML renderer: when replacing a keyed child’s node, do not cancel the child; keep subscriptions and only update the stored node reference.

### DIFF
--- a/packages/html/src/render.ts
+++ b/packages/html/src/render.ts
@@ -147,11 +147,10 @@ const bindChildren = (
       if (currentNode) {
         // Replace the previous DOM node, if any
         currentNode.replaceWith(newRendered.node);
-        keyedChildren.get(key)?.cancel();
         // Update the mapping entry to capture any newly-rendered node.
         keyedChildren.set(key, {
           ...keyedChildren.get(key)!,
-          ...newRendered,
+          node: newRendered.node,
         });
       }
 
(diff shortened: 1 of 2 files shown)


ubik2 (approved):
That works.
If we're not in pending when we go to run our action, we've issued an extra unsubscribe, and should neither run nor resubscribe.
My variant allowed for a node to unsubscribe itself (preventing resubscription) as part of its action, but it's not clear that that would be used.

PR #1721, opened 2025-09-04
Refactor `llmDialog` node to be entirely event driven

bfollington:
- **Working refactor!**
- **Check pending flag before committing new messages**

    

---

## Summary by cubic
Refactored llmDialog to be fully event-driven, gating runs with a pending flag to prevent re-entry and stale writes. This simplifies state, improves cancellation safety, and streamlines tool-call handling.

- **Refactors**
  - Added addMessage event to start an LLM run; ignore new events while pending.
  - Moved logic into mainLogic; removed run counters and request-hash dedupe.
  - Execute tool calls and await handler results before appending assistant + toolResults.
  - Check pending before committing tool results; clear pending on non-tool responses and on errors.
  - Strip tool handlers before sending to the LLM; handlers run locally.

### DIFF
--- a/packages/runner/src/builtins/llm-dialog.ts
+++ b/packages/runner/src/builtins/llm-dialog.ts
@@ -31,7 +31,11 @@ const client = new LLMClient();
  * @param action - The mutation action to perform if pending is true
  * @returns true if the action was performed, false otherwise
  */
-function perform(runtime: IRuntime, pending: Cell<boolean>, action: (tx: IExtendedStorageTransaction) => void) {
+function perform(
+  runtime: IRuntime,
+  pending: Cell<boolean>,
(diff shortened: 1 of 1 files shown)


packages/runner/src/builtins/llm-dialog.ts
```diff
@@ -111,197 +117,187 @@ export function llmDialog(
...
+      context: "charm",
+    },
+    cache: true,
+    tools: toolsWithoutHandlers, // Pass through tools if provided
+  };
 
-    resultPromise
-      .then(async (llmResult) => {
-        if (thisRun !== currentRun) return;
+  // Case should not occur anymore, since event handler is the only invocation path
```
seefeldb, line 166: Let's remove all that dead code?

PR #1726, opened 2025-09-04
JSON Schema transformer rewrite

mathpirate:
WIP - not yet ready to merge, but ready for initial review

New packages/schema-generator/ package which replaces the old functionality in packages/js-runtime/typescript/transformer/schema-generator.ts.

Includes changes to the compiler to use the new package for schema generation instead of the old one; seems to be fully working (a few of the old fixture tests are failing, I have one pending question about whether the tests vs the new behavior should be changed).
    

---

## Summary by cubic
Rewrites JSON Schema generation into a standalone package and switches js-runtime to use it, improving correctness for unions, recursion, and defaults while adding focused tests.

- **Refactors**
  - Introduced packages/schema-generator with modular formatters (primitive, object, array, union, intersection, CommonTools).
  - Exposed plugin API createSchemaTransformerV2; js-runtime transformer now uses it.
  - Updated workspace config and lockfile to include the new package.
  - Added dedicated schema-generator test suite and fixtures; aligned js-runtime fixtures and added small recipe exports for dev.
  - Addresses CT-841 by unifying schema logic and improving intersection handling.

- **Bug Fixes**
  - String literal unions now emit enum instead of array.
  - Stream<Cell<T>> correctly sets asCell and asStream.
  - Improved cycle detection and $ref promotion (only when referenced).
  - Default<T, V> propagates defaults for primitives, arrays, and objects.
  - Date maps to string with date-time format.
  - OpaqueRef methods are preserved without incorrect get().map() rewrites.

### DIFF
--- a/packages/schema-generator/src/schema-generator.ts
+++ b/packages/schema-generator/src/schema-generator.ts
@@ -238,7 +238,11 @@ export class SchemaGenerator implements ISchemaGenerator {
     // Handle case where rootSchema might be boolean (per JSON Schema spec)
     if (typeof rootSchema === "boolean") {
       return rootSchema === false
-        ? { $schema: "https://json-schema.org/draft-07/schema#", not: true, definitions }
+        ? {
+          $schema: "https://json-schema.org/draft-07/schema#",
+          not: true,
+          definitions,
+        }
         : { $schema: "https://json-schema.org/draft-07/schema#", definitions };
     }
     return {
--- a/packages/schema-generator/src/formatters/common-tools-formatter.ts
+++ b/packages/schema-generator/src/formatters/common-tools-formatter.ts
@@ -135,9 +135,9 @@ export class CommonToolsFormatter implements TypeFormatter {
 
     // Handle case where innerSchema might be boolean (per JSON Schema spec)
     if (typeof innerSchema === "boolean") {
-      return innerSchema === false 
-        ? { asCell: true, not: true }  // false = "no value is valid"
-        : { asCell: true };            // true = "any value is valid"
+      return innerSchema === false
+        ? { asCell: true, not: true } // false = "no value is valid"
+        : { asCell: true }; // true = "any value is valid"
     }
     return { ...innerSchema, asCell: true };
   }
--- a/packages/schema-generator/test/schema/complex-defaults.test.ts
+++ b/packages/schema-generator/test/schema/complex-defaults.test.ts
@@ -16,21 +16,21 @@ describe("Schema: Complex defaults", () => {
     `;
     const { type, checker } = await getTypeFromCode(code, "WithArrayDefaults");
     const s = createSchemaTransformerV2()(type, checker);
-    
+
     // Validate root schema required fields
     expect(s.required).toEqual(["emptyItems", "prefilledItems", "matrix"]);
-    
+
     const empty = s.properties?.emptyItems as any;
     expect(empty.type).toBe("array");
     expect(empty.items?.type).toBe("object");
     expect(empty.items?.required).toEqual(["title", "done"]);
     expect(Array.isArray(empty.default)).toBe(true);
-    
+
     const pre = s.properties?.prefilledItems as any;
     expect(pre.type).toBe("array");
     expect(pre.items?.type).toBe("string");
     expect(pre.default).toEqual(["item1", "item2"]);
-    
+
     const mat = s.properties?.matrix as any;
     expect(mat.type).toBe("array");
     expect(mat.items?.type).toBe("array");
@@ -48,23 +48,26 @@ describe("Schema: Complex defaults", () => {
     `;
     const { type, checker } = await getTypeFromCode(code, "WithObjectDefaults");
     const s = createSchemaTransformerV2()(type, checker);
-    
+
     // Validate root schema required fields
     expect(s.required).toEqual(["config", "user"]);
-    
+
     const config = s.properties?.config as any;
     expect(config.type).toBe("object");
     expect(config.properties?.theme?.type).toBe("string");
     expect(config.properties?.count?.type).toBe("number");
     expect(config.required).toEqual(["theme", "count"]);
     expect(config.default).toEqual({ theme: "dark", count: 10 });
-    
+
     const user = s.properties?.user as any;
     expect(user.type).toBe("object");
     expect(user.properties?.name?.type).toBe("string");
     expect(user.properties?.settings?.type).toBe("object");
     expect(user.required).toEqual(["name", "settings"]);
-    expect(user.properties?.settings?.required).toEqual(["notifications", "email"]);
+    expect(user.properties?.settings?.required).toEqual([
+      "notifications",
+      "email",
+    ]);
     expect(user.default).toEqual({
       name: "Anonymous",
       settings: { notifications: true, email: "user@example.com" },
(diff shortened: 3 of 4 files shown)


ubik2:
Top level `deno.json` has a "fmt" clause with an "exclude" entry for "packages/js-runtime/test/fixtures".
You'll want to add "packages/schema-generator/test/fixtures" to that array to get the automated tests to pass.

packages/js-runtime/test/fixtures/schema-transform/opaque-ref-map.expected.ts
```diff
@@ -1,10 +1,8 @@
-import { recipe, OpaqueRef, toSchema, JSONSchema } from "commontools";
-
+import { recipe, OpaqueRef, JSONSchema } from "commontools";
 interface TodoItem {
-  title: string;
-  done: boolean;
+    title: string;
```
ubik2, line 3: i'm confused by the whitespace change here. the input looks the same, and the other outputs didn't seem to change to 4 spaces.

packages/schema-generator/src/formatters/intersection-formatter.ts
```diff
@@ -0,0 +1,147 @@
...
+        context.typeChecker,
+        undefined, // No specific type node for intersection parts
+      );
+
+      if (this.isObjectSchema(schema)) {
+        // Merge properties from this part
+        if (schema.properties) {
+          for (const [key, value] of Object.entries(schema.properties)) {
+            if (mergedProps[key] && mergedProps[key] !== value) {
+              // Property conflict - could improve this with more sophisticated merging
```
ubik2, line 102: You can get pretty good results merging into an `anyOf`, but the complexity may not be worth it for such an uncommon situation when they have the option of manually specifying the schema.

packages/schema-generator/src/formatters/primitive-formatter.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (flags & ts.TypeFlags.Number) {
+      return { type: "number" };
+    }
+    if (flags & ts.TypeFlags.Boolean) {
+      return { type: "boolean" };
+    }
+    if (flags & ts.TypeFlags.Null) {
+      return { type: "null" };
+    }
+    if (flags & ts.TypeFlags.Undefined) {
```
ubik2, line 66: These three (Undefined, Void, Never) don't really have a good representation in JSONSchema. The string/enum treatment makes it seem like they should have those literal values, while none of these three can ever occur at all in a proper JSON object.

It's probably best to just skip these types, though I think it's also ok to return `{}` which will match anything. That's not technically correct, but is probably easier to merge with other types if needed.

The 2020 version of JSONSchema prefers `true` instead of `{}`, but both are allowed, and we have some code paths that want it to be an object (you can't attach asCell, asStream, or ifc properties to a boolean).

packages/schema-generator/src/formatters/primitive-formatter.ts
```diff
@@ -0,0 +1,82 @@
...
+      return { type: "string", enum: ["undefined"] };
+    }
+    if (flags & ts.TypeFlags.Void) {
+      return { type: "string", enum: ["void"] };
+    }
+    if (flags & ts.TypeFlags.Never) {
+      return { type: "string", enum: ["never"] };
+    }
+    if ((flags & ts.TypeFlags.Unknown) || (flags & ts.TypeFlags.Any)) {
```
ubik2, line 75: For this one, I'd return `{}`, which, confusingly, will match primitives as well as objects.

`{}` is a magic value that matches anything.

packages/schema-generator/test/fixtures/schema/type-aliases.input.ts
```diff
@@ -0,0 +1,16 @@
...
+interface Stream<T> {
+  subscribe(cb: (v: T) => void): void;
+}
+type Default<T, V> = unknown;
+type MyCell<T> = Cell<T>;
+type MyStream<T> = Stream<T>;
+type MyDefault<T, V> = Default<T, V>;
+interface SchemaRoot {
+  a: MyCell<number>;
+  b: MyStream<MyCell<string>>;
```
ubik2, line 14: I'm not sure what this would mean. @seefeldb may have a better idea, but it's probably still a good test case to make sure we don't go crazy.

packages/schema-generator/test/fixtures/schema/recursive-type-nested.input.ts
```diff
@@ -0,0 +1,12 @@
+// Nested recursive: RootType contains recursive LinkedList
+interface LinkedList {
+  value: number;
+  next?: LinkedList;
```
ubik2, line 4: I may have missed it, but can we add a test where the next is:
```json
next: LinkedList | null;
```
If the transformer can't deal with that, then never mind, but if it can, it would be good to have a test.
Note that I don't think I handle that correctly myself.

I would expect this to turn into an `anyOf`.

ubik2: Never mind, This is well covered.

packages/schema-generator/src/interface.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  // Accumulating state (grows during generation)
+  /** Named type definitions for $refs */
+  definitions: Record<string, SchemaDefinition>;
+  /** Which $refs have been emitted */
+  emittedRefs: Set<string>;
+
+  // Stack state (push/pop during recursion)
+  /** Current recursion path for cycle detection */
+  definitionStack: Set<any>;
```
ubik2, line 30: It looks like this could be narrowed to `string | ts.Type`.

packages/js-runtime/test/fixtures/schema-transform/opaque-ref-map.expected.ts
ubik2: All good.
Just couldn't tell how/why it happened. Thanks for investigating.

packages/schema-generator/test/fixtures/schema/type-aliases.input.ts
mathpirate: As per discussion in Discord in #gideon-log, Stream<Cell<T>> now just emits asStream: true and ablates the Cell part. Let me know if you want to discuss this live. (And PS Cell<Stream<T>> throws an error.)

packages/schema-generator/src/interface.ts
mathpirate: Great point ty, done

packages/schema-generator/src/schema-generator.ts
```diff
@@ -0,0 +1,335 @@
...
+      // Accumulating state
+      definitions: {},
+      emittedRefs: new Set(),
+
+      // Stack state
+      definitionStack: new Set(),
+      inProgressNames: new Set(),
+
+      // Optional context
+      ...(typeNode && { typeNode }),
```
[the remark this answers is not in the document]
mathpirate: I hadn't seen it before either. Thanks Claude :)

packages/schema-generator/src/type-utils.ts
```diff
@@ -0,0 +1,392 @@
...
+      "property fallback node",
+    );
+    if (typeFromFallback) return typeFromFallback;
+  }
+
+  // Last resort: use symbol location
+  const typeFromSymbol = safeGetTypeOfSymbolAtLocation(
+    checker,
+    prop,
+    prop.valueDeclaration!,
```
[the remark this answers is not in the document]
mathpirate: oooh thanks, updated this, take a look

PR #1739, opened 2025-09-09
Change JSONSchema type to allow boolean

ubik2:
- JSONSchemaObj type can be used when you specifically want the object version
    

---

## Summary by cubic
Enable full JSON Schema boolean support across the stack and add JSONSchemaObj for object-only schemas, with guards and utilities updated to safely handle true/false schemas. This prevents crashes, improves IFC handling, and keeps codegen/UI paths robust when schemas aren’t objects.

- **Refactors**
  - API: JSONSchema is now JSONSchemaObj | boolean; added JSONSchemaObj export and updated nested fields to use JSONSchema.
  - Runner: Updated CFC, schema validation/transform, traversal, builder utils, and cells to accept boolean schemas; improved defaults, anyOf/oneOf, $ref resolution, and IFC attachment when schemas are true/false.
  - Charm/LLM/Shell: Guarded property access with isObject checks; codegen ignores boolean existing schemas; UI and search handle boolean schemas without errors.
  - Types/tests: JSONSchemaMutable now aliases JSONSchemaObj; updated type defs and tests accordingly.

- **Migration**
  - Use JSONSchemaObj when you specifically need an object-shaped schema (e.g., to access properties/type).
  - Add isObject(schema) checks before accessing schema.properties or other object fields.


ubik2:
tagged @jsantell since I use [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone).
Seems like there's enough support that it's ok.

PR #1741, opened 2025-09-09
Change from time-based Cache-Control headers to Etag based asset caching

jakedahn:
This should mean that when static assets are modified, clients immediately get the latest version.

When static assets are unchanged (via etag), they will return a 304 response from toolshed.
    

---

## Summary by cubic
Switch static asset caching from time-based headers to strong ETag validation. Clients get updated assets immediately; unchanged assets return 304.

- **New Features**
  - Added ETag utilities (SHA-256 strong ETags, compareETags, createCacheHeaders with public, no-cache).
  - Updated StaticCache to compute/cache ETags; new getWithETag while get stays backward compatible.
  - Updated /static and compiled shell routes to send ETag, honor If-None-Match, and return 304 when unchanged.
  - Exported the ETag module and added tests for ETag generation, headers, 304 behavior, and MIME types.


jsantell (approved):
LGTM!

PR #1778, opened 2025-09-15
CT-886 conflict sync error unit test and improvements

ubik2:
## Summary by cubic
Fixes conflict sync errors by invalidating cached recipe results when storage updates occur. The runner now reruns a recipe only when its result cell actually changes. Addresses Linear CT-886.

- Bug Fixes
  - Added resultRecipeCache keyed by result cell to skip unnecessary reruns.
  - Subscribed to storage notifications and clear cache for changed cells.
  - Reused the existing result cell instead of recreating it; ensured proper cancel.
  - Added a test confirming cache invalidation on commit notifications.

### DIFF
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -85,8 +85,15 @@ export class Runner implements IRunner {
         const space = notification.space;
         if ("changes" in notification) {
           for (const change of notification.changes) {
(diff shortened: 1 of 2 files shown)


ubik2:
Added a unit test and handle some edge cases when flushing the cache.

PR #1779, opened 2025-09-16
Implement auto-generated title for in `chatbot.tsx`

bfollington:


packages/api/index.ts
```diff
@@ -295,7 +295,7 @@ export interface BuiltInLLMParams {
 
 export interface BuiltInLLMState<T> {
   pending: boolean;
-  result?: T;
+  result?: BuiltInLLMContent;
```
bfollington, line 298: This subverts the intent for this to be a generic type, but we must return `BuiltInLLMContent` from anything other than `generateObject` calls.

packages/patterns/chatbot.tsx
```diff
@@ -56,8 +56,53 @@ type ChatOutput = {
...
+  const title = derive(titleResult, (t) => {
+    if (typeof t === "string") {
+      return t;
+    }
+
+    if (t?.[0] && t[0].type === "text") {
+      return t[0].text;
+    }
+
+    return "Untitled Chat";
```
bfollington, line 100: A little convoluted, we could strengthen the types to say `result` is never a raw string.

ellyxir (approved):
looks good, i didnt realise how much you use api/index.ts 
i dont see any issues.

PR #1788, opened 2025-09-17
store a selected charm in a cell (selectedCharm) and also maintain a …

ellyxir:
…list of charms (charmsList). show the selectedCharm in the UI, show a list of charms and allow user to change the value for selectedCharm which will update which charm we render


packages/patterns/chatbot-list-view.tsx
```diff
@@ -0,0 +1,140 @@
...
+    type: "object",
+    properties: {
+      charm: { type: "object" },
+      selectedCharm: { type: "object", asCell: true },
+      charmsList: {
+        type: "array",
+        items: {
+          type: "object",
+          properties: {
+            [ID]: { type: "string" }, // randomId is a string
```
ubik2, line 46: Assuming this ID is still a symbol, this schema isn't particularly valid.
I imagine it will still do the right thing (we're not using that to filter), and it presumably makes your typescript code happier.

ellyxir: thanks, removed it

packages/patterns/chatbot-list-view.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+import Chat from "./chatbot.tsx";
+
+type CharmEntry = {
+  [ID]: string; // randomId is a string
+  charm: any;
+};
+
+type Input = {
+  selectedCharm: Default<any, undefined>;
```
ubik2, line 27: I'm not quite sure what happens with "undefined" here. It's not generally valid in JSON, but in a perfect world, it would make the `selectedCharm` property optional.

PR #1789, opened 2025-09-17
chore: Replace jsdom with htmlparser2 collection.

jsantell:
JSDOM was pulling in a large dependency graph, involved
clobbering globals, and generally more than we need for CLI rendering
and testing.

Now using htmlparser2 and friends, updating @commontools/html render to
accept a document and options for setting props (or attributes).


packages/html/src/render.ts
```diff
@@ -370,14 +405,17 @@ export function serializableEvent<T>(event: Event): T {
   for (const property of allowListedEventTargetProperties) {
     targetObject[property] = event.target?.[property as keyof EventTarget];
   }
-  if (
-    event.target instanceof HTMLSelectElement && event.target.selectedOptions
```
jsantell, line 374: We do not have `HTMLSelectElement` as a runtime type in Deno -- instead, duck type

packages/html/test/html-recipes.test.ts
```diff
@@ -126,18 +124,20 @@ describe("recipes with HTML", () => {
...
 
-    const parent = document.createElement("div");
-    document.body.appendChild(parent);
+    const root = document.getElementById("root")!;
     const cell = result.key(UI);
-    render(parent, cell.get());
+    render(root, cell.get(), renderOptions);
 
+    // Keys are "[object Object]" due to mapping an `Opaque<number>` in handler.
+    // Maybe unintentional(?)
```
jsantell, line 132: AFAICT, calling `toString()` on an `Opaque<number>` in a recipe results in a `Prop` value as `[object Object]`, which I believe is "correct" here. If this isn't the case, then I believe the recipe here needs to be corrected

packages/html/test/html-recipes.test.ts
```diff
@@ -28,15 +29,9 @@ describe("recipes with HTML", () => {
...
-    dom = new JSDOM(`<!DOCTYPE html><html><body></body></html>`);
-    document = dom.window.document;
-
-    // Set up global environment
-    globalThis.document = document;
-    globalThis.Element = dom.window.Element;
-    globalThis.Node = dom.window.Node;
-    globalThis.Text = dom.window.Text;
+    mock = new MockDoc(
+      `<!DOCTYPE html><html><body i=${`${i++}`}><div id="root"></div><div id="root2"></div></body></html>`,
```
jsantell, line 33: TODO: remove `i`

PR #1790, opened 2025-09-17
Check initializedUpTo to make sure it's not pointing past a reverted result.

ubik2:
- If we revert a result cell (or get some other server driven change), our closure captured initializedUpTo state will be out of sync. If this means it points after the last element, re-initialize the missing elements.


ubik2:
We re-run the opRecipe initialization in this revert scenario, which may already be running, but I think that's fine.

seefeldb (approved):

PR #1792, opened 2025-09-17
Minimal reproduction of `if-else` regression

bfollington:


ubik2:
Suggest closing this, since we were able to replicate and fix this issue.

PR #1801, opened 2025-09-18
Do a better job merging argument schema (specifically definitions).

ubik2:
Ignore the messy types for now, since I'll do a better job later.
This just gets us through the integration tests.
    

---

## Summary by cubic
Improved how we merge recipe argument schemas in Runner by hoisting $defs/definitions to the process schema and defaulting missing argument schemas to true. This fixes $ref resolution and unblocks integration tests.

- **Bug Fixes**
  - Hoist $defs and definitions from recipe.argumentSchema to the top level of the process cell schema so refs resolve correctly.
  - Set properties.argument to true when argumentSchema is missing, instead of {}.

### DIFF
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -585,14 +585,32 @@ export class Runner implements IRunner {
     syncAllMentionedCells(inputs);
     await Promise.all(promises);
 
-    const sourceCell = resultCell.getSourceCell({
+    // TODO(@ubik2): Move this to a more general method in schema.ts or cfc.ts
+    const processCellSchema: any = {
       type: "object",
       properties: {
         [TYPE]: { type: "string" },
-        argument: recipe.argumentSchema ?? {},
+        argument: recipe.argumentSchema ?? true,
       },
       required: [TYPE],
-    });
+    };
+
+    if (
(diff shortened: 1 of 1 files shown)


mathpirate:
cool! LGTM, thank you so much for noticing this and jumping on it so quickly

i don't really understand this system well yet, so just to be completely sure: this operates at the level of a single schema, meaning that it won't cause name collisions if multiple schemas in a generated program use the same definition names, correct?

ubik2:
> cool! LGTM, thank you so much for noticing this and jumping on it so quickly
> 
> i don't really understand this system well yet, so just to be completely sure: this operates at the level of a single schema, meaning that it won't cause name collisions if multiple schemas in a generated program use the same definition names, correct?

Right, in this case, the top level schema has none of its own definitions, so no chance of conflicts.

mathpirate:
> > cool! LGTM, thank you so much for noticing this and jumping on it so quickly
> > i don't really understand this system well yet, so just to be completely sure: this operates at the level of a single schema, meaning that it won't cause name collisions if multiple schemas in a generated program use the same definition names, correct?
> 
> Right, in this case, the top level schema has none of its own definitions, so no chance of conflicts.

awesome, sounds good, excited to try rebasing on this on my branch

PR #1821, opened 2025-09-25
Remove workarounds for `toSchema<>()`, `ifElse()` and `[UI]`

bfollington:
These should no longer be needed

### DIFF
--- a/packages/patterns/chatbot-list-view.tsx
+++ b/packages/patterns/chatbot-list-view.tsx
@@ -163,6 +163,20 @@ const combineLists = lift(
   },
 );
 
+const getSelectedCharm = lift<
+  { entry: { charm: any | undefined } },
+  {
+    chat: unknown;
+    note: unknown;
+    backlinks: MentionableCharm[];
+    mentioned: MentionableCharm[];
+  } | undefined
+>(
+  ({ entry }) => {
+    return entry?.charm;
+  },
+);
+
 // create the named cell inside the recipe body, so we do it just once
 export default recipe<Input, Output>(
   "Launcher",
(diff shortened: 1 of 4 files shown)


packages/patterns/chatbot-list-view.tsx
```diff
@@ -219,44 +235,40 @@ export default recipe<Input, Output>(
...
-                          ))}
-                        </ct-vstack>
-                      </details>
-                    </>
-                  );
-                } else {
-                  return null;
-                }
-              })}
+              {ifElse(
```
seefeldb, line 238: btw, since this is in JSX, i think you could write that as success ? .. : ... instead.

and I think we also support the `{selected && <>...</>}` syntax, but right now it'll turn the second part into a derive, whereas ifElse treats the options as recipe-like inputs with lower taint. not a big difference, but we could change things so that `<condition> && <expression>` gets translated to `ifElse(condition, expression, null)` @mathpirate

packages/patterns/chatbot-note.tsx
```diff
@@ -171,13 +171,12 @@ const ChatbotNote = recipe<LLMTestInput, LLMTestResult>(
 
     // why does MentionableCharm behave differently than any here?
     // perhaps optional properties?
-    const computeBacklinks = lift(
-      toSchema<
-        { allCharms: Cell<any[]>; content: Cell<string> }
-      >(),
-      toSchema<any[]>(),
+    const computeBacklinks = lift<
```
seefeldb, line 174: meta style comment, and either should way works, but i wonder whether writing this as 

`lift(({ allCharms, content }: { allCharms: Cell<MentionableCharm[]>; content: Cell<string> }): any[] => { ... }`

is cleaner, and let's us sometimes leverage typescript inference (e.g. the return type here)

PR #1831, opened 2025-09-26
fix(runner/fetchData): guard fetch-data updates during edit retry + force sync

seefeldb:
Move the run-staleness check into the editWithRetry callback so a new run that starts while we await runtime.idle() prevents older responses from overwriting fresher state.

Force sync of result helper cells, as some are otherwise never read and that can force conflicts.
    

---

## Summary by cubic
Ensures fetchData only updates state for the current run by moving the staleness check into editWithRetry. Prevents older responses from overwriting newer state if a new run starts while waiting for runtime.idle().


ubik2 (commented):
For history, the calls to sync aren't needed, but also are harmless (and maybe communicate intent).

PR #1836, opened 2025-09-29
test(runner): auto-generated headless integration like pattern tests

seefeldb:
@jsantell this actually did find a bug, it runs in about one minute on my machine. at first i wanted to check it in disabled, but maybe we should enable it anyway? i'll see whether i can parallelize execution

@mathpirate see disabled test, which found a regression

@ubik2 this still has the extra conflicts that shouldn't happen in this memory-only setting, so maybe it helps with debugging

@ellyxir FYI :)


ellyxir:
sorry didnt know there was a PR, made a small commit @seefeldb

packages/runner/integration/patterns/asset-lifecycle-tracker.ts
```diff
@@ -0,0 +1,210 @@
+import type { PatternIntegrationScenario } from "../pattern-harness.ts";
+import type {
```
jsantell, line 2: This is importing a pattern into our Deno runtime directly -- this works due to aliasing `"commontools"` to `"@commontools/api"` in our workspace, but seems like undefined behavior (or will be) to invoke `derive` and other pattern lib methods from within the Deno runtime -- importing types only circumvents this at least

packages/pattern-tests/integration/patterns/asset-lifecycle-tracker.ts
```diff
@@ -0,0 +1,210 @@
+import type { PatternIntegrationScenario } from "../pattern-harness.ts";
+import type {
```
seefeldb: this is the test file and it's only sharing types with the pattern (`import _type_`). the .pattern.ts file will run through our runtime. do you think this is still problematic? the imported types are unrelated to the API.

seefeldb:
@jsantell where do i add this new integration test to the CI?

PR #1838, opened 2025-09-29
Updates to `derive` and `lift`

mathpirate:
* Remove the `Promise` return type option
* Use Typescript's inferred types to autogenerate schemas even when the programmer doesn't explicitly provide type template values for both `lift` and `derive`
    

---

## Summary by cubic
Removes async support from derive/lift and adds automatic schema generation using TypeScript type inference when generics are omitted. This simplifies the API and cuts boilerplate by inferring and injecting schemas at build time.

- **New Features**
  - Auto-generate argument and result JSONSchema for lift and derive from function parameter and return types (when generics are not provided).
  - Added TypeRegistry to pass TypeScript Types between transformer stages for reliable schema generation.
  - Schema injection now infers types and fills missing sides with unknown when only one can be determined.

- **Migration**
  - derive and lift no longer accept Promise-returning functions; callbacks must be synchronous.
  - Update any async derive/lift usage to sync functions or an alternative API that supports async.

### DIFF
--- a/packages/schema-generator/src/formatters/common-tools-formatter.ts
+++ b/packages/schema-generator/src/formatters/common-tools-formatter.ts
@@ -251,12 +251,12 @@ export class CommonToolsFormatter implements TypeFormatter {
       // JSON Schema Draft 2020-12 allows default as a sibling of $ref
       // Simply add the default property directly to the schema
       if (typeof valueSchema === "boolean") {
-        // Boolean schemas (true/false) cannot have properties
-        // Wrap in an object schema
-        return {
-          allOf: [valueSchema],
-          default: defaultValue,
-        } as SchemaDefinition;
+        // Boolean schemas (true/false) cannot have properties directly
+        // For true: { default: value } (any value is valid)
+        // For false: { not: true, default: value } (no value is valid)
+        return valueSchema === false
+          ? { not: true, default: defaultValue } as SchemaDefinition
+          : { default: defaultValue } as SchemaDefinition;
       }
       (valueSchema as any).default = defaultValue;
     }
--- a/packages/schema-generator/test/schema/complex-defaults.test.ts
+++ b/packages/schema-generator/test/schema/complex-defaults.test.ts
@@ -116,4 +116,34 @@ describe("Schema: Complex defaults", () => {
     expect(u.anyOf).toBeDefined();
     expect(u.anyOf).toEqual([{ type: "string" }]);
   });
+
+  it("boolean schema defaults (any/never with Default)", async () => {
+    const code = `
+      interface Default<T,V> {}
+      interface WithBooleanSchemas {
+        anyWithDefault: Default<any, "defaultValue">;
+        neverWithDefault: Default<never, "fallbackValue">;
+      }
+    `;
+    const { type, checker } = await getTypeFromCode(
+      code,
+      "WithBooleanSchemas",
+    );
+    const s = createSchemaTransformerV2()(type, checker);
+
+    // Validate root schema required fields
+    expect(s.required).toEqual(["anyWithDefault", "neverWithDefault"]);
+
+    const anyProp = s.properties?.anyWithDefault as any;
+    // any becomes boolean schema true, with default should be { default: "defaultValue" }
+    expect(anyProp.default).toBe("defaultValue");
+    expect(anyProp.not).toBeUndefined();
+    expect(anyProp.allOf).toBeUndefined();
+
+    const neverProp = s.properties?.neverWithDefault as any;
+    // never becomes boolean schema false, with default should be { not: true, default: "fallbackValue" }
+    expect(neverProp.default).toBe("fallbackValue");
+    expect(neverProp.not).toBe(true);
+    expect(neverProp.allOf).toBeUndefined();
+  });
 });


packages/ts-transformers/test/fixtures/ast-transform/schema-generation-handler-event-only.expected.tsx
```diff
@@ -0,0 +1,17 @@
...
+// Only event is typed, state should get unknown schema
+export const incrementer = handler({
+    type: "object",
+    properties: {
+        amount: {
+            type: "number"
+        }
+    },
+    required: ["amount"]
+} as const satisfies JSONSchema, {} as const satisfies JSONSchema, (event: IncrementEvent, state) => {
```
ubik2, line 15: Surprising if `false` is our go-to for unknown schema.

packages/ts-transformers/src/schema/schema-transformer.ts
```diff
@@ -33,7 +36,16 @@ export function createSchemaTransformer(
         if (!typeArg) {
           return ts.visitEachChild(node, visit, context);
         }
-        const type = checker.getTypeFromTypeNode(typeArg);
+
+        // First check if we have a registered Type for this node
+        // (from schema-injection when synthetic TypeNodes were created)
+        let type: ts.Type;
+        if (typeRegistry && typeRegistry.has(node)) {
```
jsantell, line 43: Are these the same Nodes that have registered to the type registry, e.g. the same instance (as opposed to a different instance representing the same underlying syntax fragment)?

packages/ts-transformers/src/opaque-ref/rules/schema-injection.ts
```diff
@@ -3,8 +3,168 @@ import ts from "typescript";
...
+} from "../type-inference.ts";
+import type { TypeRegistry } from "../../core/type-registry.ts";
+
+function isFunctionLikeExpression(
+  expression: ts.Expression,
+): expression is ts.ArrowFunction | ts.FunctionExpression {
+  return ts.isArrowFunction(expression) || ts.isFunctionExpression(expression);
+}
+
+// Promise-related code removed - no longer supporting async derives/lifts
```
seefeldb, line 22: left over AI comment

packages/ts-transformers/test/fixtures/ast-transform/schema-generation-derive-inside-jsx.expected.tsx
```diff
@@ -0,0 +1,8 @@
+/// <cts-enable />
+import { derive, h } from "commontools";
+declare const value: number;
+// This derive is INSIDE JSX and should NOT be transformed with schemas
```
seefeldb, line 4: we should transform derive, handler, etc. inside JSX as well + add schemas, even for the ones that are automatically added

packages/ts-transformers/test/fixtures/ast-transform/schema-generation-handler-event-only.expected.tsx
mathpirate: great point, thanks. there were a bunch of fixtures that still had the wrong expectation for this and were passing incorrectly due to a branch where we were still generating {}. i fixed the fixture expectations and the code behavior

packages/ts-transformers/src/transform.ts
```diff
@@ -13,8 +14,12 @@ export function commonTypeScriptTransformer(
...
   }
 
+  // Create a TypeRegistry scoped to this transformation pipeline
+  // This allows schema-injection to pass Type information to schema-transformer
+  const typeRegistry = new TypeRegistry();
+
   return [
-    createModularOpaqueRefTransformer(program),
-    createSchemaTransformer(program),
+    createModularOpaqueRefTransformer(program, { typeRegistry }),
```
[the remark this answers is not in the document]
mathpirate: I honestly don't have a great theoretical answer here - it appears empirically to be the case that Typescript's transformation pipeline passes the updated-in-place SourceFile from one transformer to the next, such that when the later transformers use the SourceFile object (which is the root node of the AST) they receive to walk the AST, they encounter the Nodes that were created/modified by the previous transformers in the pipeline.

packages/ts-transformers/src/opaque-ref/type-inference.ts
```diff
@@ -0,0 +1,184 @@
...
+ * Check if a type is 'any' or 'unknown'
+ */
+export function isAnyOrUnknownType(type: ts.Type | undefined): boolean {
+  if (!type) return false;
+  return (type.flags & (ts.TypeFlags.Any | ts.TypeFlags.Unknown)) !== 0;
+}
+
+/**
+ * Infer the type of a function parameter, with optional fallback
+ * Returns undefined if the type cannot be inferred or is 'any'/'unknown'
```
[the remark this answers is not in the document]
mathpirate: great point thanks, now returning the actual type from this

packages/ts-transformers/test/fixtures/ast-transform/schema-generation-derive-inside-jsx.expected.tsx
mathpirate: thanks for noticing this, i fixed this now

ubik2 (approved):
lgtm!

packages/ts-transformers/test/type-inference-exploration.ts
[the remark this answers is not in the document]
mathpirate: I was thinking of leaving them in through the merge to main, then maybe deleting them off main later - that way they'd be in the repo history but not taking up space in the folder... but I'm open to whatever, not sure what's best!

PR #1841, opened 2025-09-30
Handle both forms of `charmId` access

bfollington:
This is a strange bug, it might imply mixing up the result and source cell somewhere? Or [ID] missing in some objects?

### DIFF
--- a/packages/ui/src/v2/components/ct-code-editor/ct-code-editor.ts
+++ b/packages/ui/src/v2/components/ct-code-editor/ct-code-editor.ts
@@ -254,7 +254,12 @@ export class CTCodeEditor extends BaseElement {
             type: "text",
             info: "Create new backlink",
             apply: () => {
-              this.emit("backlink-create", { text: raw });
+              // Instantiate the pattern if available
+              if (this.pattern) {
+                this.createBacklinkFromPattern(raw);
+              } else {
+                this.emit("backlink-create", { text: raw });
+              }
             },
           });
         }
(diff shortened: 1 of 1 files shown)


packages/ui/src/v2/components/ct-code-editor/ct-code-editor.ts
```diff
@@ -436,9 +441,11 @@ export class CTCodeEditor extends BaseElement {
       if (charm) {
         // this is VERY specific
         // if you do `getEntityId(mentionableArray.key(i))` you'll get a different answer (the ID of the array itself)
-        const charmIdObj = getEntityId(mentionableArray.get()[i]);
-        const charmId = charmIdObj?.["/"] || "";
-        if (charmId === id) {
+        const charmIdObjA = getEntityId(mentionableArray.get()[i]);
+        const charmIdObjB = getEntityId(mentionableArray.key(i));
```
seefeldb, line 445: I have to double check, but I think that getEntityId ignores to path part, so all of these IDs will be the same irrespective of `i`.

seefeldb (commented):
This might overmatch.

The other concerning part is that if these mentionables are all runnable charm, then they must be result cells and so always their own docs. The case where `[ i ]` isn't a link, but an inline object (what you get when not setting [ID] on a .set) would mean we copied the results away from the runnable stuff, possibly.

Oh, or is mentionables an object that boxes the charm reference? If so, just compare the actual charm reference!

PR #1845, opened 2025-09-30
feat(runner/built-in): Add mutex mechanism to fetch-data builtin to prevent redundant requests

seefeldb:
This implements a transaction-based mutex similar to llm-dialog, but adapted for the different characteristics of HTTP requests:

## Key Features

- **5-second timeout**: Much shorter than llm-dialog's 5 minutes, as HTTP requests typically complete quickly
- **Input-based coordination**: Tracks input hash to detect when parameters change
- **Opportunistic writes**: Any tab can write results if inputs match at completion time
- **Abort on input change only**: In-flight requests continue unless inputs change, preventing waste while maintaining correctness

## Implementation

- Adds internal state cell tracking `requestId`, `lastActivity`, and `inputHash`
- `tryClaimMutex()` prevents multiple tabs from starting identical requests
- `tryWriteResult()` allows any tab to write if inputs haven't changed
- Aborts in-flight requests only when inputs change, not on mutex contention
- Properly cleans up on recipe cancellation

## Behavior

With multiple tabs open:
- Only one tab starts a request for given inputs
- If that tab times out (5s), another can take over
- Any tab can write the result as long as inputs match
- Resource waste is minimized while correctness is maintained

---

## Summary by cubic
Add a short-lived, transaction-based mutex to the fetch-data builtin to prevent duplicate HTTP requests across tabs. Only one request runs per unique input; others defer, and any tab can write the result if inputs still match.

- **New Features**
  - 5s mutex timeout with input-hash coordination (requestId, lastActivity, inputHash).
  - tryClaimMutex blocks duplicate starts; tryWriteResult enables safe opportunistic writes.
  - In-flight requests are aborted only when inputs change; proper cleanup on recipe cancel.

### DIFF
--- a/packages/runner/src/builtins/fetch-data.ts
+++ b/packages/runner/src/builtins/fetch-data.ts
@@ -265,7 +265,7 @@ export function fetchData(
     const { url } = inputsCell.getAsQueryResult([], tx);
     const inputHash = computeInputHash(tx, inputsCell);
 
-    if (url === undefined) {
+    if (!url) {
       pending.withTx(tx).set(false);
       result.withTx(tx).set(undefined);
       error.withTx(tx).set(undefined);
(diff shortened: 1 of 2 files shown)


ubik2 (approved):
my initial review was cursory, but fine to land this.

PR #1847, opened 2025-09-30
test of github automated tests when ban-types is removed from our lint rules

ubik2:
Ignore this PR

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -73,7 +73,8 @@
       "exclude": [
         "no-unused-vars",
         "no-explicit-any",
-        "require-yield"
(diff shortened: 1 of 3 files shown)


seefeldb:
@jsantell any final thoughts here? these also removes some other checks as we discussed, but I'm not sure anyone of us has ever run into that (e.g. using `Number` instead of `number` as type).

ubik2:
I'm going to close this out for now. We can revisit it if we run into hurdles again in the schema-generator code, but there's no urgent need for a workaround, and we now know it works.

PR #1849, opened 2025-10-01
Allow lift and handler functions to access charm cell via `this`

seefeldb:
Functions in JavaScript modules can now access the charm cell they belong to via the `this` keyword. The runner binds the charm's result cell to `this` when invoking lift functions (line 1080) and event handlers (line 999).

This enables recipes to:
- Access their own cell's properties (sourceURI, space, entityId, etc.)
- Use cell methods to navigate or read related data
- Reference themselves in their outputs

Tests verify that lift functions can access `this.space` properties correctly.


seefeldb:
changed strategy
- now it's just in recipe, not lift and handler
- no magic typing, you have to do it by hand

let's see where we use it and then maybe replace it with a `self` built-in or so later

bfollington:
Closing in favour of https://github.com/commontoolsinc/labs/pull/1909

PR #1852, opened 2025-10-01
Unwrap OpaqueRef<T> to T in schema generation and add 'asOpaque: true'

mathpirate:
## Summary by cubic
Unwraps OpaqueRef<T> to its inner type during schema generation and marks schemas with asOpaque: true. This makes derive/lift outputs correctly represent opaque values.

- **New Features**
  - Unwrap OpaqueRef<T> and add asOpaque: true to the resulting schema.
  - Detect OpaqueRef when it expands to intersection types (OpaqueRefMethods & T).

- **Refactors**
  - Unified wrapper handling for Cell, Stream, and OpaqueRef; preserves Stream behavior and the Cell<Stream<T>> guard.
  - getNamedTypeKey now treats OpaqueRef as a wrapper to avoid hoisting inner named types.

### DIFF
--- a/packages/schema-generator/src/type-utils.ts
+++ b/packages/schema-generator/src/type-utils.ts
@@ -231,7 +231,8 @@ export function getNamedTypeKey(
   // Check if this is a Default/Cell/Stream/OpaqueRef wrapper type via alias
   const aliasName = (type as TypeWithInternals).aliasSymbol?.name;
   if (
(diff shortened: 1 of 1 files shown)


ubik2 (approved):
looks good. i should make a corresponding change to JSONSchema
Edit: https://github.com/commontoolsinc/labs/pull/1854

PR #1853, opened 2025-10-01
chore: Organize and normalize transformer usage

jsantell:
* core: Common types, CT-specific parsing
* ast: generic utilities for extending TS AST traversal
* transformers: our 3 primary transformation paths, now with similar mechanisms
  * Can explore creating the TransformationContext/reporting diagnostics in a consistent way as well against these transformers
* transformers/opaque-ref: containing opaque-ref specific transformation impl
* transformers/opaque-ref/emitters: Rewrite emitters


mathpirate (approved):
this looks really awesome, thank you!

PR #1857, opened 2025-10-02
Extract `backlinks-index` recipe CT-970

bfollington:
## Summary by cubic
Centralized backlink computation by adding a shared BacklinksIndex recipe and updating Note to consume it. This reduces per-note work and unifies the mentionable list across the app.

- **Refactors**
  - Added BacklinksIndex recipe that builds a backlinks map and exposes the shared mentionable list.
  - Note now accepts index (backlinks + mentionable) instead of allCharms, and uses it for backlinks lookup and mentions.
  - Updated default-app to create one BacklinksIndex and pass index to Note and spawnNote.
  - Updated chatbot-note-composed to create BacklinksIndex and pass index to Note and the newNote tool.
  - Event handlers now push to index.mentionable rather than allCharms.

### DIFF
--- a/packages/patterns/chatbot-note-composed.tsx
+++ b/packages/patterns/chatbot-note-composed.tsx
@@ -321,7 +321,10 @@ export default recipe<ChatbotNoteInput, ChatbotNoteResult>(
       backlinks: note.backlinks,
       list,
       // Expose both child charms for mention systems that scan charm exports.
-      mentionable: [chat as unknown as MentionableCharm, note as unknown as MentionableCharm],
+      mentionable: [
+        chat as unknown as MentionableCharm,
+        note as unknown as MentionableCharm,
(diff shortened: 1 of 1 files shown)


packages/patterns/backlinks-index.tsx
```diff
@@ -0,0 +1,63 @@
...
+      BacklinksMap
+    >(
+      ({ allCharms }) => {
+        const cs = allCharms.get() ?? [];
+        const index: BacklinksMap = {};
+
+        for (const c of cs) {
+          const mentions = c.mentioned ?? [];
+          for (const m of mentions) {
+            const key = m?.content || m?.[NAME];
```
bfollington, line 41: Probably overly defensive, but it works for `note.tsx` created from `default-app.tsx`

PR #1864, opened 2025-10-03
Feature/map closures

mathpirate:
Not yet ready for merge

This branch implements closure support for map on OpaqueRef and Cell

Currently includes a first draft of a closure transformer with associated fixtures; next, will include a new method map_with_pattern on OpaqueRef and Cell which will support the new signature
    

---

## Summary by cubic
Adds a ClosureTransformer that rewrites OpaqueRef/Cell array map callbacks with outer-scope captures into recipe callbacks with a params object. This enables safe, explicit data flow in maps, supports nested callbacks and method chains, and adds thorough docs and fixtures.

- **New Features**
  - Transforms OpaqueRef<T[]> and Cell<T[]> .map(fn) when fn captures outer variables to: map(recipe(({ elem, index?, params:{...} }) => ...), { ... }).
  - Selective capture: skips module-scoped declarations, imports, functions/handlers, and globals; captures from parent callback scope; treats property access as a unit (e.g., state.user.name).
  - Handles nested callbacks, array method chains before map (filter/slice/concat/reverse/sort/flat/flatMap), destructured params, index param, JSX, template literals, conditionals, and type assertions. Plain arrays are not transformed.
  - Automatically injects the recipe import. Adds closure-design and implementation-roadmap docs plus comprehensive fixtures.

- **Refactors**
  - Added ClosureTransformer to the pipeline (runs first).
  - Dataflow analyzer: handles synthetic nodes and preserves array-map rewrite hints.
  - OpaqueRef call-expression emitter: skips rewriting the map call itself and rewrites the callee chain when needed.
  - Test setup: new closures fixture suite; updated a few expectations to reflect the new transformation.

### DIFF
--- a/packages/ts-transformers/src/closures/transformer.ts
+++ b/packages/ts-transformers/src/closures/transformer.ts
@@ -954,12 +954,15 @@ function createRecipeCallWithParams(
   const paramsObject = factory.createObjectLiteralExpression(paramProperties);
 
   // Create mapWithPattern property access
-  const mapWithPatternAccess = ts.isPropertyAccessExpression(mapCall.expression)
-    ? factory.createPropertyAccessExpression(
-      mapCall.expression.expression, // state.items
-      factory.createIdentifier("mapWithPattern"),
-    )
-    : factory.createIdentifier("mapWithPattern"); // Fallback (shouldn't happen)
+  if (!ts.isPropertyAccessExpression(mapCall.expression)) {
+    throw new Error(
+      "Expected mapCall.expression to be a PropertyAccessExpression",
+    );
+  }
+  const mapWithPatternAccess = factory.createPropertyAccessExpression(
+    mapCall.expression.expression, // state.items
+    factory.createIdentifier("mapWithPattern"),
+  );
 
   // Return the transformed mapWithPattern call
   return factory.createCallExpression(


packages/runner/src/builder/opaque-ref.ts
```diff
@@ -140,6 +141,22 @@ export function opaqueRef<T>(
           ),
         });
       },
+      map_with_pattern: <S>(
```
jsantell, line 144: prefer: camel case

packages/runner/src/builder/opaque-ref.ts
```diff
@@ -19,6 +19,7 @@ import { getTopFrame, recipe } from "./recipe.ts";
 import { createNodeFactory } from "./module.ts";
 
 let mapFactory: NodeFactory<any, any>;
+let mapWithPatternFactory: NodeFactory<any, any>;
```
jsantell, line 22: * Shouldn't this be `NodeFactory<any, any> | undefined`?
* Could `NodeFactory`s generics be `unknown`? (maybe not, could be too difficult)

packages/runner/src/builder/types.ts
```diff
@@ -126,6 +126,10 @@ declare module "@commontools/api" {
         array: T,
       ) => Opaque<S>,
     ): Opaque<S[]>;
+    map_with_pattern<S>(
```
jsantell, line 129: prefer: camel case (for `map_with_pattern` everywhere other than filenames)

packages/runner/src/builtins/map_with_pattern.ts
```diff
@@ -0,0 +1,153 @@
...
+    // distinguish empty inputs from undefined inputs?
+    if (list === undefined) {
+      resultWithLog.set([]);
+      // Reset progress so that once the list becomes defined again we
+      // recompute from the beginning.
+      initializedUpTo = 0;
+      return;
+    }
+
+    if (!Array.isArray(list)) {
```
jsantell, line 98: Can `list` not be an array even with the provided schema? Do we need a default value?

packages/runner/src/builtins/map_with_pattern.ts
```diff
@@ -0,0 +1,153 @@
...
+      throw new Error("map_with_pattern currently only supports arrays");
+    }
+
+    const newArrayValue = resultWithLog.get().slice(0, initializedUpTo);
```
jsantell, line 102: Can this returned function (`(tx: IExtendedStorageTransaction) => { .. }`) execute multiple times? If not, `initializedUpTo` is always `0` here (same with `newArrayValue.length`), though I'm not too familiar with these mechanisms

packages/ts-transformers/src/closures/transformer.ts
```diff
@@ -0,0 +1,646 @@
+import ts from "typescript";
+import {
+  hasCtsEnableDirective,
+  TransformationContext,
+  Transformer,
+} from "../core/mod.ts";
+import { isOpaqueRefType } from "../transformers/opaque-ref/opaque-ref.ts";
+import { getHelperIdentifier } from "../transformers/opaque-ref/import-resolver.ts";
+
+export class ClosureTransformer extends Transformer {
```
jsantell, line 10: :metal:

packages/ts-transformers/src/closures/transformer.ts
```diff
@@ -0,0 +1,646 @@
...
+    transformedBody,
+    nestedVisitor,
+  ) as typeof transformedBody;
+
+  // Now transform the callback body to use elem instead of the original param name
+  const elemName = elemParam?.name;
+
+  // Collect destructured property names if the param is a destructured binding pattern
+  const destructuredProps = new Set<string>();
+  if (elemName && ts.isObjectBindingPattern(elemName)) {
```
jsantell, line 478: future ideas: Similar to wanting a DSL for rewriting typescript, I wonder if there's some abstraction we could generally apply to the functions we rewrite to handle the transformation of args (optional args, optional destructuring), or handle rewriting functions the same way (regardless if it was an arrow function, function declaration, function expression etc.), since we could reuse them across several transformations, not worrying about all the different forms user authored TS comes in within the transformer business logic

packages/ts-transformers/src/closures/transformer.ts
```diff
@@ -0,0 +1,646 @@
...
+
+  // Create map_with_pattern property access (e.g., state.items.map_with_pattern)
+  // mapCall.expression is a PropertyAccessExpression like state.items.map
+  // We need to replace "map" with "map_with_pattern"
+  const mapWithPatternAccess = ts.isPropertyAccessExpression(mapCall.expression)
+    ? factory.createPropertyAccessExpression(
+      mapCall.expression.expression, // state.items
+      factory.createIdentifier("map_with_pattern"),
+    )
+    : factory.createIdentifier("map_with_pattern"); // Fallback (shouldn't happen)
```
jsantell, line 591: Let's throw an error if we don't expect this

packages/ts-transformers/src/closures/transformer.ts
```diff
@@ -0,0 +1,646 @@
...
+
+  function visit(node: ts.Node): ts.Node {
+    // Check for OpaqueRef<T[]> or Cell<T[]> map calls with callbacks
+    if (ts.isCallExpression(node) && isOpaqueRefArrayMapCall(node, checker)) {
+      const callback = node.arguments[0];
+
+      // Check if the callback is an arrow function or function expression
+      if (
+        callback &&
+        (ts.isArrowFunction(callback) || ts.isFunctionExpression(callback))
```
jsantell, line 612: There's another similar helper function defined here `isFunctionDeclaration` that seems to include various types of function types -- if we don't care what *type* of function it is, maybe we can reuse that, or just add these utils to the generic ast helpers in `ast/` (i think that's the rough idea anyway)

packages/ts-transformers/src/closures/transformer.ts
```diff
@@ -0,0 +1,646 @@
...
+  const recipeIdentifier = getHelperIdentifier(
+    factory,
+    context.sourceFile,
+    "recipe",
+  );
+  const recipeCall = factory.createCallExpression(
+    recipeIdentifier,
+    undefined,
+    [
+      factory.createStringLiteral("map with pattern including captures"),
```
jsantell, line 575: I think this "name" is surfaced in several places when interacting with patterns, maybe there's a better name here

packages/ts-transformers/src/closures/types.ts
```diff
@@ -0,0 +1,11 @@
+import ts from "typescript";
+import type { TransformationContext } from "../core/context.ts";
+
+export interface ClosureRule {
```
jsantell: Is this used?

packages/ts-transformers/src/closures/transformer.ts
mathpirate: yeah... thanks... i agree, it would be better to have more conceptual unification here. i'm going to try to think about this again from the top soon and to also incorporate berni's nudge to try constructing our transformations in such a way that as much of the original code can survive untransformed as possible. some of the other transformers do a bunch of parameter name rewriting that's legacy from the original transformer whose behavior i wasn't ready to change. soon i'll take a broad pass at this

packages/ts-transformers/src/closures/transformer.ts
mathpirate: this ended up being a schema instead of a name

PR #1865, opened 2025-10-03
fix(runner): Fix default value resolution for chained $ref in schemas

seefeldb:
### Problem

When a property schema contains a `$ref` that points to another schema definition which itself contains a `$ref` (chained references), default values were not being applied correctly. The system only resolved one level of `$ref`, meaning defaults defined in intermediate references were never discovered.

**Example scenario that was broken:**

```typescript
const schema = {
  $defs: {
    Settings: {
      type: "object",
      properties: {
        enabled: { type: "boolean" },
        label: { type: "string" }
      },
      default: { enabled: true, label: "from ref" }
    },
    SettingsWithDefault: {
      $ref: "#/$defs/Settings",
      default: { enabled: false, label: "from default" }  // This default was being ignored!
    }
  },
  type: "object",
  properties: {
    config: {
      $ref: "#/$defs/SettingsWithDefault"
    }
  }
}
```

When reading `config` from an object with undefined config, the expected behavior was to use the default from `SettingsWithDefault` (`{ enabled: false, label: "from default" }`), but instead nothing was returned because the system didn't resolve the chained `$ref` to discover the default value.

### Solution

1. **Enhanced `resolveSchema()`** to support full resolution of chained `$ref`s
   - Added cycle detection to prevent infinite loops with circular references
   - When a cycle is detected, returns the schema with the repeating `$ref`
   
2. **Updated default value handling** in `processDefaultValue()` and `validateAndTransform()`
   - Now fully resolves schemas before checking for defaults
   - Ensures defaults in intermediate `$ref` chains are discovered and applied

### Testing

Added tests to verify:
- Defaults are correctly applied from property-level `$ref` schemas
- Defaults in chained `$ref`s (A -> B -> C) work correctly
- Precedence rules are maintained (ref site default overrides target default)
- Cycle detection prevents infinite loops

All existing tests continue to pass.

    

---

## Summary by cubic
Fixes default value resolution for schemas with chained $refs so defaults on intermediate refs are applied. Adds safe ref resolution with cycle detection. Addresses Linear CT-966.

- **Bug Fixes**
  - Resolve chained $refs fully in resolveSchema, with cycle detection.
  - Resolve schemas before checking defaults to honor intermediate ref defaults and precedence.
  - Added tests for property-level $refs, chained $refs (A→B→C), and default precedence.

### DIFF
--- a/packages/runner/src/schema.ts
+++ b/packages/runner/src/schema.ts
@@ -80,9 +80,9 @@ export function resolveSchema(
       !noFollowNestedRefs && typeof resolvedSchema.$ref === "string" &&
       rootSchema !== undefined
     ) {
-      // Detect cycles - if we've seen this $ref before, return current schema with the repeating $ref
+      // Detect cycles - if we've seen this $ref before, stop resolving
       if (seenRefs.has(resolvedSchema.$ref)) {
-        return resolvedSchema;
+        break;
       }
       seenRefs.add(resolvedSchema.$ref);
 
--- a/packages/runner/src/schema.ts
+++ b/packages/runner/src/schema.ts
@@ -48,6 +48,7 @@ export function resolveSchema(
   schema: JSONSchema | undefined,
   rootSchema: JSONSchema | undefined = schema,
   filterAsCell = false,
+  fullyResolve = false,
 ): JSONSchema | undefined {
   // Treat undefined/null/{} or any other non-object as no schema
   if (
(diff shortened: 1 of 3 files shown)


packages/runner/src/schema.ts
```diff
@@ -48,6 +48,7 @@ export function resolveSchema(
   schema: JSONSchema | undefined,
   rootSchema: JSONSchema | undefined = schema,
   filterAsCell = false,
+  noFollowNestedRefs = false,
```
ubik2, line 51: I don't think we need this flag.

packages/runner/src/schema.ts
```diff
@@ -58,18 +59,43 @@ export function resolveSchema(
...
+        break;
+      }
+      seenRefs.add(resolvedSchema.$ref);
+
+      const resolved = ContextualFlowControl.resolveSchemaRef(
+        rootSchema,
+        resolvedSchema.$ref,
+        resolvedSchema,
+      );
+      if (!isObject(resolved)) {
```
ubik2, line 94: this can be boolean

PR #1868, opened 2025-10-03
feat(runner): Implement allOf schema composition for link chains

seefeldb:
Implements allOf-based schema composition to properly propagate defaults, asCell, and asStream properties through link chains. When following multiple links that each have schemas, those schemas are now combined using JSON Schema's allOf operator with smart extraction of special properties.

### Link Resolution (link-resolution.ts)

- Add `createAllOf()` function to combine multiple schemas
- Extract defaults from the LAST schema (last wins)
- Extract asCell/asStream from the FIRST schema (first wins)
- Remove extracted properties from branches to avoid duplication
- Accumulate schemas when following link chains
- Combine accumulated schemas using allOf when multiple schemas exist

### Schema Validation (schema.ts)

- Add `createAllOfSchema()` helper for nested property merging
- Implement allOf processing in `validateAndTransform()`
- Merge properties from all allOf branches
- Create nested allOf for duplicate properties across branches
- Normalize property schemas to include `type: "object"` when they have properties
- Process object properties even when missing from value (for nested defaults)
- Use extracted defaults/asCell/asStream from parent allOf schema

### Architecture & Semantics

The implementation uses a "union of properties" approach rather than strict JSON Schema intersection semantics. This means:
- Properties from all allOf branches are merged together
- Duplicate properties create nested allOf schemas
- Current implementation is pragmatic for our use case
- Tests include TODO comments for future strict intersection if needed

### Testing (schema-allof.test.ts)

- Add comprehensive unit tests for `createAllOf()` (11 tests)
  - Default extraction (last wins)
  - asCell/asStream extraction (first wins)
  - Combined extraction
  - Edge cases (empty arrays, single schemas, filtering)
- Add integration tests for allOf composition (26 tests)
  - Link chain composition
  - Default handling with nested properties
  - asCell/asStream propagation
  - Property merging with overlapping and non-overlapping properties
  - Required field handling
  - Edge cases

All 37 allOf tests passing ✓

## Example

Before: Link chains would use only the schema from the last link, losing defaults and asCell/asStream from earlier links.

After: Link chains combine schemas using allOf:
```typescript
// Link 1 has: { properties: { x: { default: 10 } }, asCell: true }
// Link 2 has: { properties: { y: { default: 20 } }, default: { x: 1 } }
// Result: {
//   allOf: [
//     { properties: { x: {} } },      // default extracted
//     { properties: { y: { default: 20 } } }
//   ],
//   default: { x: 1 },  // from last (Link 2)
//   asCell: true        // from first (Link 1)
// }
```

### DIFF
--- a/packages/runner/src/link-resolution.ts
+++ b/packages/runner/src/link-resolution.ts
@@ -1,4 +1,4 @@
-import { isRecord } from "@commontools/utils/types";
+import { isObject, isRecord } from "@commontools/utils/types";
 import { getLogger } from "@commontools/utils/logger";
 import { LINK_V1_TAG } from "./sigil-types.ts";
 import {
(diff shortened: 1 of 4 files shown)


packages/runner/src/link-resolution.ts
```diff
@@ -8,9 +8,176 @@ import {
...
+ *   { minLength: 5 }
+ * ])
+ * // => { allOf: [{ type: "string" }, { minLength: 5 }] }
+ *
+ * // Extracts default (last wins)
+ * createAllOf([
+ *   { type: "number", default: 1 },
+ *   { type: "number", default: 2 }
+ * ])
+ * // => { allOf: [{ type: "number" }, { type: "number" }], default: 2 }
```
ubik2, line 66: As a future improvement, any property that is the same in each branch can be brought up a level, and if all that's left in the allOf is `{}` or `true`, we can optimize it away.
`{type: "number", allOf: [{}, {}], default: 2}` => `{type: "number", default: 2}`

ubik2:
I just wrote a crude version of this, since I had forgotten about this PR.
Assigning this to myself, so I come back to it later and get it working with main.
For now, I'll use my limited implementation.

PR #1869, opened 2025-10-04
Change ContextualFlowControl method to resolve schema refs to continue to follow the refs.

ubik2:
- Changed to resolveSchemaRefs as the general API, while resolveSchemaRef is preserved for when you only want to follow one. The new implementation of resolveSchemaRef no longer merges objects (that's handled by resolveSchemaRefs).
- Fixed an issue in schemaAtPath where we weren't following the $ref after setting our cursor to a properties/additionalProperties/items value.
- Updated tests to not expect a $ref in the resolvedSchema returns.
- Added tests from @seefeld PR for both the circular references (which are undefined), and chained refs
    

---

## Summary by cubic
Switch schema resolution to follow chained $ref links via a new resolveSchemaRefs API, improving correctness for nested refs and defaults. Also fixes traversal to follow refs after moving into properties/additionalProperties/items and returns undefined on circular refs.

- **New Features**
  - Added resolveSchemaRefs to resolve $ref chains and merge overriding properties; resolveSchemaRef now resolves a single level without merging.
  - Updated ContextualFlowControl, schema.ts, traverse.ts, and cache selector to use resolveSchemaRefs.

- **Bug Fixes**
  - Follow $ref after cursor moves into properties/additionalProperties/items.
  - Detect circular $ref and return undefined (both filterAsCell modes).
  - Resolve defaults through chained $ref in property schemas; updated tests to not expect $ref in resolved outputs.


ubik2:
I created this after looking at the Claude PR for #CT-966 (so this makes that one unnecessary).
I wanted the utility functions we have in ContextualFlowControl to follow the refs and merge properties as needed.

seefeldb:
Nice, that's a better place for it. Note the failing integration test, though.

ubik2:
> Nice, that's a better place for it. Note the failing integration test, though.

Hrm, tagging @mathpirate 

```typescript
type Output = {
  mentioned: Default<Array<MentionableCharm>, []>;
  content: Default<string, "">;
  backlinks: Default<Array<MentionableCharm>, []>;
};
```
seems to be turning into:
```json
{
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "type": "object",
    "properties": {
      "mentioned": {
        "$ref": "#/$defs/AnonymousType_1"
      },
      "content": {
        "type": "string",
        "default": ""
      },
      "backlinks": {
        "$ref": "#/$defs/AnonymousType_1"
      }
    },
    "required": [
      "mentioned",
      "content",
      "backlinks"
    ],
    "$defs": {
      "AnonymousType_1": {
        "$ref": "#/$defs/AnonymousType_1",
        "default": []
      }
    }
  }
  ```

PR #1878, opened 2025-10-07
Fix: Detect actual default exports in ts-transformer pipeline

mathpirate:
Previously, mapPrefixProgramFiles() incorrectly assumed files had default exports based on whether program.mainExport was undefined or "default". This caused compilation errors for files with only named exports.

Changes:
- Added hasDefaultExport() function that parses source AST to detect actual default exports (both "export default" and "export { X as default }")
- Updated mapPrefixProgramFiles() to check actual file contents instead of making assumptions
- Added TypeScript import to @commontools/runner package

This fixes transformation errors for patterns (eg like kanban-board-grouping) that only export named exports.

### DIFF
--- a/deno.lock
+++ b/deno.lock
@@ -1909,6 +1909,11 @@
           "npm:@opentelemetry/api@^1.9.0"
         ]
       },
+      "packages/runner": {
+        "dependencies": [
+          "npm:typescript@*"
+        ]
+      },
       "packages/schema-generator": {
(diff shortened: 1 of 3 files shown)


mathpirate:
I am not sure whether this is the right approach and would appreciate guidance/input, since I'm not very familiar with the runner's engine... it seems to work and seems naively reasonable to me

PR #1906, opened 2025-10-15
Allow rendering UI from any cell/subpath in ct-render

bfollington:
- **Implement subtree support for ct-render**
- **Add omni-layout**
- **Add toggleable sidebar**

    

---

## Summary by cubic
Enables ct-render to render UI from any cell or subpath and adds OmniLayout with a toggleable sidebar and FAB slots. Unlocks slot-based layouts and supports CT-989.

- **New Features**
  - ct-render can render from root or subpath cells, resolving UI from cell.key(UI) or the cell value; syncs cells and handles missing recipe IDs.
  - Added x-omni-layout with slots: main, sidebar, fab, plus a sidebar toggle button.
  - SpaceView uses OmniLayout and ct-render to render main, sidebarUI, and fabUI from the default pattern.
  - Default app recipe now returns sidebarUI and fabUI.

### DIFF
--- a/packages/ui/src/v2/components/ct-render/ct-render.ts
+++ b/packages/ui/src/v2/components/ct-render/ct-render.ts
@@ -165,15 +165,10 @@ export class CTRender extends BaseElement {
     // 1. Being passed a UI cell directly (e.g., cell.key(UI))
     // 2. Being passed a root charm cell (extract UI subcell)
     let uiCell = cell;
-    try {
-      const uiSubcell = cell.key(UI as never);
-      if (uiSubcell.get() !== undefined) {
-        this._log("extracting UI subcell from root charm cell");
-        uiCell = uiSubcell;
-      }
-    } catch (error) {
-      // If we can't read the UI subcell, just use the cell as-is
-      this._log("no UI subcell found, using cell as-is");
+    const uiSubcell = (cell as Cell<{ [UI]: unknown }>).key(UI);
+    if (uiSubcell.get() !== undefined) {
+      this._log("extracting UI subcell from root charm cell");
+      uiCell = uiSubcell;
     }
 
     this._log("rendering UI");


packages/ui/src/v2/components/ct-render/ct-render.ts
```diff
@@ -141,26 +140,101 @@ export class CTRender extends BaseElement {
...
+    }
+  }
+
+  private _isRenderableVNode(value: unknown): boolean {
+    return isVNode(value);
+  }
+
+  private _isRecord(value: unknown): value is Record<string, unknown> {
+    return typeof value === "object" && value !== null;
+  }
```
jsantell: These don't need to be methods, and `isRecord` is in `@commontools/utils`

packages/ui/src/v2/components/ct-render/ct-render.ts
```diff
@@ -141,26 +140,101 @@ export class CTRender extends BaseElement {
...
+
+    if (this._isRenderableVNode(value) || value === undefined) {
+      return cell;
+    }
+
+    throw new Error("Cell does not contain renderable UI content");
+  }
+
+  private async _maybeSyncCell(cell: Cell<unknown>) {
+    const sync = (cell as { sync?: () => Promise<unknown> | unknown }).sync;
```
seefeldb: better to use isCell here

packages/ui/src/v2/components/ct-render/ct-render.ts
```diff
@@ -141,26 +140,101 @@ export class CTRender extends BaseElement {
...
+    if (typeof sync === "function") {
+      await sync.call(cell);
```
seefeldb: and i don't think Cell.sync takes a parameter currently

seefeldb (commented):
the ct-render part could be a lot simpler.

i think this should work:
- call render(..., cell) immediately -- if cell is a cell it'll react to future updates and if it's a primitive value, then it's anyway fine to call. it also follows all the UI links already.
- guard the rest in `isCell()` && path.length === 0
- await cell.sync()
- getRecipeId()
- if it succeeds call runSynced

(you don't need to call sync for path.length > 0 either, because render's get will implicitly do that for you)

or if you want to be faster, replace the last three with:
- promise = cell.sync();
- getRecipeId(), if it fails await promise, then try again
- if there is a recipe, call runSynced

PR #1909, opened 2025-10-16
Simplify backlinks index

bfollington:
Improves performance and reactivity feedback loops

    

---

## Summary by cubic
Simplified backlinks indexing by removing the global map and writing backlinks directly to each charm. This improves performance and makes reactivity more reliable. Consumers now use a single mentionable list.

- **Refactors**
  - Backlinks computed in-place on each charm via charm.backlinks Cell.
  - BacklinksIndex now exposes mentionable only; BacklinksMap removed.
  - Note pattern reads backlinks from a local Cell and uses index.mentionable.
  - Chatbot actions updated: listMentionable, readContentByIndex, editNoteByIndex.
  - Default app and JSX types updated to expect $mentions with { mentionable } only.
  - Code editor docs no longer reference a backlinks field.

- **Migration**
  - Use /backlinksIndex/mentionable for lists; do not rely on a global backlinks map.
  - Expect backlinks on each charm: charm.backlinks is a Cell<MentionableCharm[]>.
  - Update $mentions prop shape to { mentionable }.
  - Replace listNotes/readNoteByIndex with listMentionable/readContentByIndex.
  - editNoteByIndex now edits content via the mentionable list.

### DIFF
--- a/packages/ui/src/v2/components/ct-code-editor/ct-code-editor.ts
+++ b/packages/ui/src/v2/components/ct-code-editor/ct-code-editor.ts
@@ -955,7 +955,9 @@ export class CTCodeEditor extends BaseElement {
       ? ((rawMentioned.get() ?? []) as MentionableArray)
       : [];
 
-    const current: Mentionable[] = currentSource.filter((value): value is Mentionable => Boolean(value));
+    const current: Mentionable[] = currentSource.filter((
+      value,
+    ): value is Mentionable => Boolean(value));
 
     const curIds = new Set(
       current


packages/patterns/chatbot-note-composed.tsx
```diff
@@ -175,9 +174,12 @@ export default recipe<ChatbotNoteInput, ChatbotNoteResult>(
   ({ title, messages }) => {
     const allCharms = schemaifyWish<MentionableCharm[]>("#allCharms", []);
     const index = schemaifyWish<BacklinksIndex>("/backlinksIndex", {
-      backlinks: {},
       mentionable: [],
     });
+    const mentionable = schemaifyWish<MentionableCharm[]>(
```
seefeldb: Worth double-checking that this still does the right thing with AST transformation, since it's the derive invocation that needs the JSON schema and I'm not sure the compiler unrolls this enough.

seefeldb (approved):
Looks great. Just double check whether factoring out the helper worked.

PR #1911, opened 2025-10-16
fix(runner): inline data URIs before handling write redirects

seefeldb:
Move data URI inlining logic to occur before write redirect (alias) handling in normalizeAndDiff. This ensures the correct evaluation order when a data URI contains a redirect link.

Previously, when newValue was a data URI containing a redirect, the code would:
1. Check currentValue for redirects (line 290)
2. Follow currentValue redirect if present
3. Process data URI inlining later (line 318)

This caused the redirect within the data URI to be written to the wrong location - it would follow the currentValue redirect first, then inline the data URI, resulting in the redirect being written to the destination of the currentValue alias instead of to the intended target.

The fix reorders the logic so data URIs are inlined immediately after unwrapping proxies and cells (line 212), before any redirect handling. This preserves the invariant that redirect handling at line 290 only applies when newValue is not itself an alias.

The order is now:
1. Unwrap proxies/cells
2. Inline data URIs (exposing any contained redirects)
3. Handle redirects in newValue
4. Handle redirects in currentValue

This ensures redirects contained in data URIs are properly recognized and handled as part of newValue, not incorrectly written through a currentValue redirect.
    

---

## Summary by cubic
Fixes incorrect write redirects when newValue is a data: URI by inlining it before any redirect handling. Redirects inside data URIs now apply to the intended target, not through the currentValue alias.

- **Bug Fixes**
  - Inline data: URIs right after unwrapping proxies/cells in normalizeAndDiff.
  - New order: unwrap → inline data → handle newValue redirects → handle currentValue redirects.
  - Prevents writing data-URI redirects to the currentValue alias destination.

### DIFF
--- a/packages/runner/test/data-updating.test.ts
+++ b/packages/runner/test/data-updating.test.ts
@@ -13,6 +13,7 @@ import {
   areNormalizedLinksSame,
   createSigilLinkFromParsedLink,
   isAnyCellLink,
+  isSigilLink,
   parseLink,
 } from "../src/link-utils.ts";
 import { type IExtendedStorageTransaction } from "../src/storage/interface.ts";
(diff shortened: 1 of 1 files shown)


ellyxir (approved):
lgtm

PR #1919, opened 2025-10-17
Enhance the Schema type to handle refs to $defs/Root or other $defs

ubik2:
A future improvement is to have the schema and rootSchema combined, but to easily support this combination on the fly, we need to avoid relative $refs. The only relative $ref we currently use is to "#", and replacing that with a $defs reference will prevent the current Schema type utility from properly inferring the type based on the schema.

This change handles the special case of a $ref to "#/$defs/Root", so we can rewrite the "#" refs as absolute refs without losing typing.

I ended up integrating @seefeldb 's full Schema change, so we can resolve almost all references.
    

---

## Summary by cubic
Teach the Schema type to resolve $ref to "#/$defs/Root" so we can replace relative "#" refs with absolute refs without losing type inference. Keeps recursive schemas correctly typed.

- **New Features**
  - Resolves $ref "#/$defs/Root" by inferring from $defs.Root.
  - Falls back to any for broken root refs.
  - Adds tests for a recursive schema and ensures child types aren’t any.

### DIFF
--- a/packages/runner/test/schema-ref-default.test.ts
+++ b/packages/runner/test/schema-ref-default.test.ts
@@ -120,7 +120,7 @@ describe("$ref with default support", () => {
       });
     });
 
-    it("should handle boolean false schema (cannot add default)", () => {
+    it("should handle boolean false schema", () => {
       const schema: JSONSchema = {
         $defs: {
           AlwaysFalse: false,


seefeldb (commented):
I have a more elaborate version of this in an abandoned branch:

https://github.com/commontoolsinc/labs/compare/main...feat/support-default-as-sibling-to-ref

this would implement support for any $defs path

ubik2:
This ended up being the combination of the type improvements from [feat/support-default-as-sibling-to-ref](https://github.com/commontoolsinc/labs/compare/main...feat/support-default-as-sibling-to-ref) and existing $ref handling improvements.

PR #1923, opened 2025-10-20
feat(html): add comprehensive React-style object-as-styles support

seefeldb:
Add full support for passing style objects in JSX, matching React's style prop behavior. This enables more ergonomic styling in patterns and recipes.

## Key Features

- **CamelCase to kebab-case conversion**: Automatically converts React-style property names (e.g., `backgroundColor`) to CSS property names (e.g., `background-color`)

- **Smart unit handling**: Automatically appends 'px' to numeric values for properties that require units, while respecting unitless properties like `opacity`, `zIndex`, `flexGrow`, `lineHeight`, etc.

- **Vendor prefix support**: Handles browser vendor prefixes correctly (e.g., `WebkitTransform` → `-webkit-transform`, `MozAppearance` → `-moz-appearance`)

- **Null/undefined value handling**: Gracefully skips null and undefined values, allowing conditional styling

- **Complex value support**: Handles complex CSS values like gradients, shadows, and transforms without modification

## Implementation

- Added `styleObjectToCssString()` helper function in both render.ts and utils.ts for consistent conversion logic
- Updated `setProp()` in render.ts to detect style objects and convert them to CSS strings before setting the attribute
- Updated mock document's `setProp()` in utils.ts to ensure tests work correctly with style objects

## Test Coverage

Added 8 comprehensive test cases covering:
- Basic React-style object to CSS string conversion
- Unitless numeric properties
- Vendor prefixes (webkit, moz, ms, o)
- Zero value handling
- Null and undefined value handling
- Complex CSS values (box-shadow, gradients, transforms)
- Style objects alongside other props
- Empty style objects

All tests pass with full type safety.

## Usage Example

```tsx
<div style={{
  backgroundColor: "red",
  fontSize: 16,                    // → "16px"
  opacity: 0.5,                    // → "0.5" (unitless)
  WebkitTransform: "rotate(45deg)", // → "-webkit-transform"
  margin: 0,                        // → "0"
  boxShadow: "0 4px 6px rgba(0, 0, 0, 0.1)"
}} />
```

---

## Summary by cubic
Adds React-style object-as-styles support to the HTML renderer. JSX style objects now render to correct CSS strings with proper casing, units, and vendor prefixes.

- **New Features**
  - Converts camelCase to kebab-case (e.g., backgroundColor → background-color).
  - Smart units: adds px to numbers, keeps unitless props (opacity, zIndex, etc.), and leaves 0 unitless.
  - Handles vendor prefixes (Webkit/Moz/ms/o → -webkit/-moz/-ms/-o).
  - Skips null/undefined values and preserves complex values (gradients, shadows, transforms).


seefeldb:
this is all pre-sanitization, which is definitely still required to ship

but it might actually be easier eventually to do it on this kind of structure vs a CSS string we have to parse, so i want to support it now

packages/html/src/utils.ts
```diff
@@ -6,6 +6,72 @@ import * as domhandler from "domhandler";
 import * as domserializer from "dom-serializer";
 import { RenderOptions } from "./render.ts";
 
+/**
+ * Converts a React-style CSS object to a CSS string.
+ * Supports vendor prefixes, pixel value shorthand, and comprehensive CSS properties.
+ * @param styleObject - The style object with React-style camelCase properties
+ * @returns A CSS string suitable for the style attribute
+ */
+const styleObjectToCssString = (styleObject: Record<string, any>): string => {
```
jsantell: This function is duped between utils.ts and render.ts

packages/html/src/render.ts
```diff
@@ -321,7 +321,86 @@ const listen = (
...
+      if (value == null) return "";
+
+      // Convert camelCase to kebab-case, handling vendor prefixes
+      let cssKey = key;
+
+      // CSS custom properties (--*) are case-sensitive and should not be transformed
+      if (!key.startsWith("--")) {
+        // Handle vendor prefixes (WebkitTransform -> -webkit-transform)
+        if (/^(webkit|moz|ms|o)[A-Z]/.test(key)) {
+          cssKey = "-" + key;
```
jsantell: This results in `WebkitTransform` -> `--webkit-transform` with two `--`, I don't think this is needed if the goal is to convert `PascalCase` to single-dashed-lowercase `-kebab-case`

packages/html/src/utils.ts
seefeldb: ups, fixed and moved utils.ts to test/

we should split up render.ts, but i'll do that in another PR

PR #1926, opened 2025-10-21
docs(common): iteratively improved prompts

seefeldb:
## Summary by cubic
Adds shopping list patterns (basic, categorized, composed) and updates CommonTools docs to center on bidirectional binding, making recipes simpler and faster to build. Also refreshes ct-select/ct-checkbox demos and introduces a comprehensive PATTERNS.md.

- **New Features**
  - Added shopping-list.tsx, shopping-list-categorized.tsx (grouping via derive), and shopping-list-composed.tsx (shared state rendered with ct-render).
  - Updated ct-checkbox and ct-select demos to showcase $checked/$value binding, logging, and numeric selections.
  - Introduced docs/common/PATTERNS.md with the “prefer bidirectional binding” rule and quick decision guides; expanded COMPONENTS.md, HANDLERS.md, and RECIPES.md to clarify when handlers are needed.
  - Improved tutorials and dev workflow tips; added PATTERN_DOCUMENTATION_FEEDBACK.md capturing gaps and proposed fixes.


docs/common/RECIPES.md
```diff
@@ -302,7 +302,7 @@ logic.
      { items: Cell<ShoppingItem[]>; item: OpaqueRef<ShoppingItem> }
    >((_event, { items, item }) => {
      const currentItems = items.get();
-     const index = currentItems.findIndex((el) => items.key(el).equals(item));
+     const index = currentItems.findIndex((_, idx) => items.key(idx).equals(item));
```
seefeldb: sorry, this is bad code that ended up here.

it should really be

```
   const removeItem = handler<
     unknown,
     { items: Cell<ShoppingItem[]>; item: Cell<ShoppingItem> }
   >((_event, { items, item }) => {
     const currentItems = items.get();
     const index = currentItems.findIndex((el) => item.equals(el));
...
```

quite important to not have an example where OpaqueRef is the type in a handler signature, not sure why it did that.

and after that we can just use the item as Cell to call .equals on the actual array element, no need to go via .key

ellyxir: fixed

PR #1943, opened 2025-10-22
Updated documentation with Claude's learnings after struggling to bui…

jkomoros:
…ld a specific type of pattern, profile.tsx in the recipes repo.

Not sure if what Claude learned is actually helpful changes to the documentation or not.
    

---

## Summary by cubic
Clarifies handler binding rules in HANDLERS.md to prevent ReadOnlyAddressError when binding inside derive(), and adds troubleshooting tips for common handler issues to help recipe authors avoid mistakes.

- **New Features**
  - Added “Never Create Handler Bindings Inside derive()” with wrong/right examples and the golden rule.
  - Added troubleshooting for ReadOnlyAddressError, handlers not firing, type issues, and data not updating.

### DIFF
--- a/docs/common/HANDLERS.md
+++ b/docs/common/HANDLERS.md
@@ -56,6 +56,84 @@ const removeItem = handler(
 
 The handler receives the event from the UI as the first parameter, and your bound state as the second parameter. This allows you to merge discrete UI events with reactive cells that are always changing.
 
+## ❌ CRITICAL: Never Create Handler Bindings Inside derive()
+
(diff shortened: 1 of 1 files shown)


seefeldb (changes requested):
The underlying issue here is missing closure support for derives:

l.77:       <ct-button onClick={applyData({ data: extractionResult })}>

it used `extractionResult` here instead of `result`, otherwise this would have worked.


it's overall better to not create UI inside derive (inside map is fine and actually common), so that's maybe what we should add to the docs. but the changes claude proposed here are otherwise misleading.

PR #1944, opened 2025-10-22
Fix ct-render edgecase

bfollington:
When encountering a link to a cell we could end up rendering nothing.

    

---

## Summary by cubic
Fixes a rendering bug where CTRender could show nothing when given a link to a cell with a non-empty path. We now load and render the recipe when possible, even for subpath links.

- **Bug Fixes**
  - Try getRecipeIdFromCharm first and load the recipe; otherwise fall back to subpath/direct render.
  - Add error handling and logging when charm resolution fails.
  - Add a test to ensure recipes load for non-empty path links.


bfollington:
Closing in favour of https://github.com/commontoolsinc/labs/pull/1945 (the real issue)

PR #1947, opened 2025-10-23
Introduce "space cell" managed by charm manager

bfollington:
- **Introduce 'space cell' with pointers to other well known cells**
- **`/` points to space cell, `#` points to default-pattern**
- **Update wish paths**

    

---

## Summary by cubic
Introduced a space cell managed by CharmManager to hold pointers to well-known cells and updated wish semantics so "/" resolves within the space cell and "#" resolves within the default pattern. Removed the trash concept and added recent charm tracking.

- **New Features**
  - Added SpaceCell schema and a spaceCell in CharmManager linking allCharms and recentCharms; linkDefaultPattern connects the default pattern.
  - Updated wish: "/" navigates the space cell (e.g., "/allCharms"), "#" targets the default pattern (e.g., "#/backlinksIndex/mentionable"); "#allCharms" still supported.
  - Added recentCharms list with trackRecentCharm; removed trash-related APIs and behavior.
  - Updated patterns and tests to new wish paths; CLI and controller reflect removal without trash.

- **Migration**
  - Use "/allCharms" via the space cell and "#/..." for default pattern fields.
  - Ensure default pattern creation calls CharmManager.linkDefaultPattern to bind it into the space cell.

### DIFF
--- a/packages/runner/src/builtins/wish.ts
+++ b/packages/runner/src/builtins/wish.ts
@@ -89,26 +89,6 @@ function getSpaceCell(ctx: WishContext): Cell<unknown> {
   return ctx.spaceCell;
 }
 
-function getSpaceField(
-  ctx: WishContext,
-  key: string,
-): Cell<unknown> {
-  return (getSpaceCell(ctx) as Cell<Record<string, unknown>>).key(
-    key as never,
-  ) as Cell<unknown>;
-}
-
-function resolveMaybeLinkedCell(
-  cell: Cell<unknown>,
-  ctx: WishContext,
-): Cell<unknown> {
-  try {
-    return cell.resolveAsCell().withTx(ctx.tx);
-  } catch {
-    return cell.withTx(ctx.tx);
-  }
-}
-
 function resolvePath(
   base: Cell<unknown>,
   path: readonly string[],
@@ -117,20 +97,13 @@ function resolvePath(
   let current = base.withTx(ctx.tx);
   for (const segment of path) {
     const keyed = current.key(segmentToPropertyKey(segment) as never);
-    current = resolveMaybeLinkedCell(keyed, ctx);
+    try {
+      current = keyed.resolveAsCell().withTx(ctx.tx);
+    } catch {
+      current = keyed.withTx(ctx.tx);
+    }
   }
   return current;
-}
-
-function getDefaultPatternCell(
-  ctx: WishContext,
-): Cell<unknown> | undefined {
-  const field = getSpaceField(ctx, "defaultPattern");
-  try {
-    return field.resolveAsCell().withTx(ctx.tx);
-  } catch {
-    return undefined;
-  }
 }
 
 function resolveBase(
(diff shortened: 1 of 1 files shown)


packages/charm/src/manager.ts
```diff
@@ -133,26 +147,48 @@ export class CharmManager {
...
+      [],
       charmListSchema,
     );
     this.pinnedCharms = this.runtime.getCell(
       this.space,
       "pinned-charms",
       charmListSchema,
     );
-    this.trashedCharms = this.runtime.getCell(
+    this.recentCharms = this.runtime.getCell(
```
bfollington: This seems wrong since we don't _need_ a fixed cause, but then what should the cause be...?

packages/patterns/chatbot.tsx
```diff
@@ -393,6 +394,16 @@ export default recipe<ChatInput, ChatOutput>(
...
+        <button
+          type="button"
+          title="Clear chat"
+          onClick={clearChat({
+            messages,
+            pending,
+          })}
+        >
+          Clear
+        </button>
```
bfollington: I snuck this in but it's not strictly related to this change.

packages/runner/src/builtins/wish.ts
```diff
@@ -47,14 +46,25 @@ function parseWishTarget(target: string): ParsedWishTarget | undefined {
...
-    return { key, path: segments };
+
+    // Check if it's a well-known target like #allCharms
+    if (segments.length > 0) {
+      const baseSegment = segments[0];
+      const key = `#${baseSegment}`;
+      // If it's a known well-known target, use it with remaining path
+      if (key === "#allCharms") {
+        return { key, path: segments.slice(1) };
+      }
```
bfollington: Ultimately we can drop this once we're accustomed to getting `/allCharms` to grab it from the space cell.

packages/runner/src/builtins/wish.ts
```diff
@@ -47,14 +46,25 @@ function parseWishTarget(target: string): ParsedWishTarget | undefined {
   if (trimmed.startsWith("#")) {
     const rest = trimmed.slice(1);
     const segments = rest.split("/").filter((segment) => segment.length > 0);
-    const baseSegment = segments.shift();
-    if (!baseSegment) return undefined;
-    const key = `#${baseSegment}`;
-    return { key, path: segments };
+
```
jsantell: Could have a `WishKey` type using these sigils (`#`, `/`) in `@commontools/api`/with the Wish params that patterns and here in the runtime could use

```ts
type WishKey = `/${string}` | `#${string}`
```

though a bit strange that `#`, `/` and well-known `#allCharms` are all `keys` -- if `#` targets properties on the Default App, why do we manage `#allCharms` in the runtime? almost seems like `#allCharms` is a unique thing, maybe should have a different sigil e.g. `$ALL_CHARMS` (caps for constant-ish)

packages/runner/src/builtins/wish.ts
```diff
@@ -45,12 +45,12 @@ function parseWishTarget(target: string): ParsedWishTarget | undefined {
...
-    const segments = rest.split("/").filter((segment) => segment.length > 0);
-    const baseSegment = segments.shift();
-    if (!baseSegment) return undefined;
-    const key = `#${baseSegment}`;
-    return { key, path: segments };
+    const segments = trimmed.slice(1).split("/").filter((segment) =>
+      segment.length > 0
+    );
+    if (segments.length === 0) return undefined;
+    const key = `#${segments[0]}` as WishKey;
```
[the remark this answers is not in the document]
bfollington: Wrong.

PR #1981, opened 2025-10-29
feat: Use our own native DOM types in JSX

jsantell:
* Extend `RenderNode` with `undefined` and `Opaque<any>`
* Extend `CellLike` to include `OpaqueRefMethods` and `Opaque`
* Added `namespace CTDOM`, defining valid native DOM elements and attributes, similar to what Preact/React provides. Native events elided.
  * In lieu of a "compiling .d.ts" files solution, or change how we handle implicit types in compiler, this is one big file
* Extend `jsx.d.ts` with our own Elements.
  * Types were mostly added as-needed. We have a lot of components! Not sure how we want to type some of the events/CellLike values.
* Extend native and CT elements with our own `onClick` cell-based handler. Can extend more as needed.
* The permissive parameter `[elemName: string]: any;` has been disabled for now. Can re-enable to allow undefined elements as needed.
* Applies to workspace as well as patterns

---
## Summary by cubic
Defined CTDOM types directly in jsx.d.ts to replace TypeScript’s DOM lib, adding strong typings for native HTML tags and CT custom elements in the sandboxed runtime. Also extended JSX wrappers and RenderNode to support Opaque and undefined.

- **New Features**
  - Introduced CTDOM with DOM/ARIA typings and a comprehensive native IntrinsicElements mapping.
  - Typed a broader set of CT elements (e.g., ct-screen, ct-button, ct-chat, ct-code-editor, ct-prompt-input, ct-tags, ct-toolbar, ct-vscroll/hscroll, common-* variants) with structured props.
  - Added a cell-based onClick handler in HTMLAttributes; expanded CellLike to include Opaque and OpaqueRefMethods; RenderNode now allows undefined and Opaque<any>.

- **Refactors**
  - Rewrote jsx.d.ts to use CTDOM and DetailedHTMLProps-based mappings; removed loose prop shapes.

<sup>Written for commit b536562e552dccebcdeca197e9def470b8ddd181. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/static/assets/types/jsx.d.ts
+++ b/packages/static/assets/types/jsx.d.ts
@@ -1,9 +1,1479 @@
-import type { OpaqueRef, Cell, Props, RenderNode, VNode, Stream } from "commontools";
+import type {
+  Cell,
+  Opaque,
+  OpaqueRef,
+  OpaqueRefMethods,
+  Props,
+  RenderNode,
+  Stream,
+  VNode,
+} from "commontools";
+
+// DOM-ish types for the CT runtime.
+// The DOM is not directly available within the runtime, but the JSX
+// produced must be typed. This defines DOM types like React or Preact,
+// with a subset of supported features, and cannot rely on globals
+// existing like `HTMLElement` from TypeScript's `dom` lib.
+declare namespace CTDOM {
+  /**
+   * Used to represent DOM API's where users can either pass
+   * true or false as a boolean or as its equivalent strings.
+   */
+  type Booleanish = boolean | "true" | "false";
+
+  /**
+   * @see {@link https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin MDN}
+   */
+  type CrossOrigin = "anonymous" | "use-credentials" | "" | undefined;
+
+  /**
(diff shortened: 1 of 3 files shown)


packages/static/assets/types/jsx.d.ts
```diff
@@ -1,9 +1,1476 @@
...
+    | "tooltip"
+    | "tree"
+    | "treegrid"
+    | "treeitem"
+    | (string & {});
+
+  export interface HTMLAttributes<T> extends AriaAttributes, DOMAttributes<T> {
+    // CT extensions
+    "onClick"?: CellLike<HandlerEvent<unknown>>;
+    "children"?: RenderNode | undefined;
```
seefeldb: should this always be an array?

seefeldb (approved):
looks great! obviously a lot to eventually carefully trim down, but this will last us for a bit, thank you!

packages/static/assets/types/jsx.d.ts
jsantell: `type RenderNode = RenderNode[] | ..` so this suffices, I think in the html renderer we make it an array if not

PR #1988, opened 2025-10-30
Ellyse/reconnect failure

ellyxir:
## Summary by cubic
Fixes schema subscription updates after reconnection and on new facts (even with an empty DB). Also closes prepared statements and streamlines queries to prevent memory leaks, and stabilizes integration tests with timeouts and proper cleanup.

- **Bug Fixes**
  - Always re-run schema queries for all subscriptions on each transaction to catch new matching facts.
  - Notify subscribers for “current” facts even when they were previously pending (fixed heap merge filter).
  - Replace .all() with stmt.iter() and finalize prepared statements; make getMatchingFacts a generator to reduce memory use.
  - Correct malformed memory WebSocket URL in tests.

- **Refactors**
  - Convert integration tests to Deno.test with 3-minute timeouts and cleared timers to avoid hanging processes.
  - Rework reconnection test to use a resolving Promise and fresh DB; add a helper script to run it.
  - Add derive_array_leak test and clamp array length in the recipe to prevent errors; measure server RSS and fail on large growth.

<sup>Written for commit 8f3929d. Summary will update automatically on new commits.</sup>


packages/memory/space.ts
ubik2: Looks like Claude wants to make this use the new iterable features, but in this case, we do want to collect all our result, then close, then return those results.

packages/memory/provider.ts
ubik2: Skipping the Subscription.match test seems like a bad idea. I could understand this if it discovered that it was wrong, but without that, we're going to be doing a lot more queries.

packages/runner/integration/basic-persistence.test.ts
ubik2: Again, no problem with Deno.test instead of just running directly, but the timeout seems problematic

packages/runner/integration/array_push.test.ts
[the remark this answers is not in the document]
ellyxir: its weird, soemtimes tests just hiccup. they rarely go that long though. actually, we use the same deno between tests, i think this is a bit of a problem bc the prior toolshed can get pretty messed up from the previous test

PR #2009, opened 2025-11-04
feat(runner): Add ability to create cells without links and unify cell implementations

seefeldb:
Implements the ability to create cells before they have a full link, allowing deferred link creation based on cause information. This also unifies the previous RegularCell and StreamCell implementations into a single CellImpl class.

Key architectural changes:
- Merged RegularCell and StreamCell into unified CellImpl class
  - Stream detection now happens at runtime via isStream() check
  - Both regular cells and streams share the same interface
  - .set() automatically handles stream vs regular cell behavior

- Use NormalizedLink (with optional id/space) instead of requiring NormalizedFullLink
  - Cells can start with just { path: [] } and get id/space populated later
  - hasFullLink() checks for presence of both id and space fields
  - ensureLink() populates missing fields from cause information

- Add .for(cause) method to associate causes with cells
  - Stores cause for later link creation
  - Supports optional { force } parameter for future extensions
  - Returns cell for method chaining

- Deferred link creation using createRef()
  - ensureLink() creates entity IDs from causes when needed
  - Uses frame context or explicit .for() cause
  - Helpful error messages when context is insufficient

- Update .key() to build paths incrementally
  - Works with partial links (just path[])
  - Inherits cause from parent cell
  - Seamless transition to full links when needed

API simplification:
- Removed isStream() checks from client code (now handled internally)
- Updated createCell() signature (removed noResolve parameter)
- Consistent Cell interface regardless of stream/regular distinction

Test coverage:
- 23 new tests for optional link functionality
- All 124 existing cell tests still passing
- Full type checking passes

Rollout plan updates:
- Marked completed tasks in rollout-plan.md
- Documented implementation approach

This enables more flexible cell creation patterns where cells can be created and manipulated before their final identity is determined, which is essential for the recipe construction work.

---
## Summary by cubic
Enable creating cells before they have a full link and unify RegularCell/StreamCell into a single CellImpl, simplifying the API and making streams transparent to callers.

- **New Features**
  - Cells can start with a partial link (NormalizedLink) and obtain id/space later.
  - Added .for(cause) to attach a cause for deferred link creation; throws if a cause/link already exists unless allowIfSet=true; ensureLink() derives ids via createRef() and uses frame space when available.
  - .key() builds paths on partial links; children have their own cause container, so identity is independent from the parent.
  - Siblings created via .asSchema() and .withTx() share link creation (CauseContainer), so id/space/cause set on one applies to the others.
  - Stream vs. regular behavior is auto-detected at runtime; .set() and .send() route accordingly; public IAnyCell now exposes .for(cause).

- **Migration**
  - createCell(runtime, link, tx, synced) no longer accepts noResolve.
  - Remove isStream() checks in client code; use isCell() and the unified Cell API.
  - RegularCell and StreamCell are replaced by CellImpl internally; the public Cell interface is unchanged.
  - If a cell may be used outside a handler before it has an id/space, call .for(cause) first.

<sup>Written for commit c7b5798f88f68341e88228eb6dcc9f4c11293cb9. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/specs/recipe-construction/rollout-plan.md
+++ b/docs/specs/recipe-construction/rollout-plan.md
@@ -42,7 +42,7 @@
     isn't there, e.g. because we need to create a link to the cell (when passed
     into `anotherCell.set()` for example). We want to encourage .for use in
     ambiguous cases.
-- [ ] Add space and event to Frame
+- [x] Add space and event to Frame
 - [ ] First merge of OpaqueRef and RegularCell
   - [ ] Add methods that allow linking to node invocations
     - [ ] `setPreExisting` can be deprecated (used in toOpaqueRef which itself
--- a/packages/runner/src/builder/recipe.ts
+++ b/packages/runner/src/builder/recipe.ts
@@ -421,13 +421,17 @@ export function pushFrame(frame?: Frame): Frame {
 export function pushFrameFromCause(
   cause: any,
   unsafe_binding?: UnsafeBinding,
+  inHandler: boolean = false,
 ): Frame {
   const frame = {
     parent: getTopFrame(),
     cause,
     generatedIdCounter: 0,
     opaqueRefs: new Set(),
+    // Extract space from unsafe_binding if available and set it on frame
+    ...(unsafe_binding?.space && { space: unsafe_binding.space }),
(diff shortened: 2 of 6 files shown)


seefeldb:
this was merged with a follow-up PR

PR #2013, opened 2025-11-04
WIP: Draft up identity specs for delegation

jsantell:
First draft of defining how delegations work for namespaces, spaces, and charms

### DIFF
--- a/docs/specs/delegation.md
+++ b/docs/specs/delegation.md
@@ -9,9 +9,10 @@ Via latest [PRD](https://docs.google.com/document/d/1KixOc7L5LZ8IdJtO_9pNHohPg_L
 ```
 DID = "did:key:{string}"
 DOMAIN = "{string.}*{string}.{string}"
+CHARM_ID = "of:${HASH}"
 NAMESPACE = DID | DOMAIN
 SPACE = DID | string;
-CHARM = DID | string;
+CHARM = CHARM_ID | string;
 ADDR = "/{@NAMESPACE/}?SPACE/CHARM?"
 ```
 


docs/specs/delegation.md
```diff
@@ -0,0 +1,110 @@
...
+DELEGATIONS = UCAN[];
+// Mapping of charm name to charm identity
+ITEMS = Map<string, DidKey>
+```
+
+## Capabilities
+
+Access and permissions are handled via an authorization token like [UCAN](https://github.com/ucan-wg/spec) or another alternative. These tokens represent capabilities delegated to identities for a given subject.
+
+* `SPACE:CREATE`: A namespace-based capability indicating permission to create a new space. This is the only delegation stored in a namespace admin record.
```
seefeldb: +1, interestingly i think this could just correspond to the ability to create URIs whose hash is just derived from the space did, not some other URI that's already in the space. (assumes we restrict this, which we currently do, but sometimes consider since it guarantees unforgeable identifiers)

docs/specs/delegation.md
```diff
@@ -0,0 +1,110 @@
+# Fabric Identities
+
+This document addresses future plans for identity and permissions for the Common Tools fabric, and the process to get there.
+
+## URL structure
+
+Via latest [PRD](https://docs.google.com/document/d/1KixOc7L5LZ8IdJtO_9pNHohPg_LXEF5hFgoAzvq9KLc/edit?tab=t.vx0btmvvbit9), the URL structure represents the three address component types:
```
seefeldb: added a few comments there, but the most relevant is just to suggest that the top-level part must resolve to a space (either by being a DID, a DNS lookup or using the provider resolution), and then each subsequent component can be a lookup in that space that leads to another space, and so on.

i also added a comment there proposing that the default way to resolve components in a space is by deriving a URI from the slug and the space DID, which can then contain a link to the actual charm, or a space redirect. see below on how that maps to the CREATE capability.

docs/specs/delegation.md
```diff
@@ -0,0 +1,110 @@
...
+
+Now, `provider.com` is hosting `alice-space` in the provider namespace. No one else yet has any access to data stored in this space. Alice wants to invite Bob (`did:key:bob`) to this space, and signs a new delegation with `did:key:alice` for `did:key:bob`, granting Bob `SPACE:WRITE` permissions. Bob can now read and write data to the space, but cannot invite Eve, lacking the `SPACE:OWNER` capability.
+
+Similarly, in a non-provider namespace (e.g. `@alice.fab.com`, or `@did:key:alice`), the namespace is created, immediately assigning `SPACE:CREATE` capabilities to `did:key:alice`, or whatever key `@alice.fab.com` resolves to. No other identities may create spaces within Alice's namespace, unless they were to e.g. delegate `SPACE:CREATE` capabilities to another user.
+
+## Plan
+
+Currently, `Session`s are created in the workspace, and the abstraction between public/private spaces needs updated. There are a few areas currently that will need to use a new Session compatible with this document:
```
seefeldb: I think we can start with today's derived-from-name spaces for now and focus on adding ACLs.

Effectively ~ partitions the namespaces per user, but we could just choose to not use that.

In a way it's the most simple implementation of provider-level lookup and we can just grow it from there.

PR #2031, opened 2025-11-07
test: minimal repro that generates conflicts

seefeldb:
This reliable generates these conflicts for me. Note that this is using the in-memory sqlite test setup, so it's definitely single-player!

added a runner test as well, same effect. should make it easier to debug:

`deno test -A test/conflict-repro.test.ts` in packages/runner

```
[0m[38;5;245mrunning 1 test from ./integration/patterns/minimal-conflict-repro.test.ts[0m
minimal conflict repro ...
  minimal conflict repro ... [0m[32mok[0m [0m[38;5;245m(299ms)[0m
[0m[38;5;245m------- output -------[0m
[38;2;234;179;8m[WARN][storage.cache::19:46:40.475][0m Transaction failed ConflictError: The application/json of of:baedreifrdcsbdplxx3cb5zhetslhpvazdzgsnott5a34wugdfflztecove in did:key:z6MkgoBAqGfjfZrnExo9gAp5m6sawZuhiaHTWzh5VUc7cuj5 was expected to be ba4jcawd7fmfyomud2hkiishs6t7le67wb27lvmmjen52dyk2doxvljn7, but it is ba4jcaxickypevbukijrki4zkcwfllpa7lvftxvvmx3i6firied2fdw5l
    at Module.conflict (file:///Users/berni/src/labs-secondary/packages/memory/error.ts:25:29)
    at swap (file:///Users/berni/src/labs-secondary/packages/memory/space.ts:765:19)
    at commit (file:///Users/berni/src/labs-secondary/packages/memory/space.ts:801:5)
    at sqliteTransaction (https://jsr.io/@db/sqlite/0.12.0/src/database.ts:874:28)
    at execute (file:///Users/berni/src/labs-secondary/packages/memory/space.ts:838:11)
    at file:///Users/berni/src/labs-secondary/packages/memory/space.ts:865:12
    at traceSync (file:///Users/berni/src/labs-secondary/packages/memory/telemetry.ts:69:20)
    at transact (file:///Users/berni/src/labs-secondary/packages/memory/space.ts:856:10)
    at file:///Users/berni/src/labs-secondary/packages/memory/space.ts:219:14
    at traceSync (file:///Users/berni/src/labs-secondary/packages/memory/telemetry.ts:69:20)
    at Space.transact (file:///Users/berni/src/labs-secondary/packages/memory/space.ts:213:12)
    at file:///Users/berni/src/labs-secondary/packages/memory/memory.ts:188:26
    at eventLoopTick (ext:core/01_core.js:179:7)
    at async file:///Users/berni/src/labs-secondary/packages/memory/telemetry.ts:127:22
    at async transact (file:///Users/berni/src/labs-secondary/packages/memory/memory.ts:171:10)
    at async file:///Users/berni/src/labs-secondary/packages/memory/memory.ts:74:7
    at async file:///Users/berni/src/labs-secondary/packages/memory/telemetry.ts:127:22
    at async Memory.perform (file:///Users/berni/src/labs-secondary/packages/memory/memory.ts:71:12)
    at async MemoryProviderSession.invoke (file:///Users/berni/src/labs-secondary/packages/memory/provider.ts:324:15)
    at async Object.write (file:///Users/berni/src/labs-secondary/packages/memory/provider.ts:195:11) {
  transaction: {
    cmd: [32m"/memory/transact"[39m,
    sub: [32m"did:key:z6MkgoBAqGfjfZrnExo9gAp5m6sawZuhiaHTWzh5VUc7cuj5"[39m,
    args: {
      changes: {
        [32m"of:baedreifrdcsbdplxx3cb5zhetslhpvazdzgsnott5a34wugdfflztecove"[39m: { [32m"application/json"[39m: [36m[Object][39m }
      }
    },
    iss: [32m"did:key:z6MkgoBAqGfjfZrnExo9gAp5m6sawZuhiaHTWzh5VUc7cuj5"[39m,
    prf: [],
    iat: [33m1762544800[39m,
    exp: [33m1762548400[39m
  },
  name: [32m"ConflictError"[39m,
  conflict: {
    space: [32m"did:key:z6MkgoBAqGfjfZrnExo9gAp5m6sawZuhiaHTWzh5VUc7cuj5"[39m,
    the: [32m"application/json"[39m,
    of: [32m"of:baedreifrdcsbdplxx3cb5zhetslhpvazdzgsnott5a34wugdfflztecove"[39m,
    expected: #ba4jcawd7fmfyomud2hkiishs6t7le67wb27lvmmjen52dyk2doxvljn7,
    actual: {
      the: [32m"application/json"[39m,
      of: [32m"of:baedreifrdcsbdplxx3cb5zhetslhpvazdzgsnott5a34wugdfflztecove"[39m,
      cause: #ba4jcatu5wzftjqvuzhahgyk5t4mcc65gzxlixmmmtwifiifopuz3eih7,
      since: [33m3[39m,
      is: {
        value: {
          [32m"$TYPE"[39m: [32m"ba4jcar2jw3xl4tuo3umlqzmc6fuh3yy3pgljoumxy2b55f2zrsgowcwh"[39m,
          resultRef: [36m[Object][39m,
          internal: [36m[Object][39m,
          spell: [36m[Object][39m,
          argument: [36m[Object][39m
        }
      }
    },
    existsInHistory: [33mfalse[39m,
    history: []
  }
}
[0m[38;5;245m----- output end -----[0m
minimal conflict repro ... [0m[32mok[0m [0m[38;5;245m(322ms)[0m

[0m[32mok[0m | 1 passed (1 step) | 0 failed [0m[38;5;245m(331ms)[0m
```

---
## Summary by cubic
Fixed a ConflictError when a lifted map depends on items and a handler clears items and increments a sequence. Normalized values before hashing and added a minimal repro with tests to confirm no conflicts.

- **Bug Fixes**
  - Normalize values before assertion hashing in chronicle.ts to handle NaN consistently.
  - Add minimal conflict repro pattern and runner test to verify the fix.

<sup>Written for commit d3a0121d698b427e9507b3bcbc4ef6adfa7c8141. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/runner/test/conflict-repro.test.ts
+++ b/packages/runner/test/conflict-repro.test.ts
@@ -110,15 +110,19 @@ describe("Conflict Reproduction", () => {
     const result = runtime.run(tx, conflictRepro, {
       items: [{ id: "test" }],
     }, resultCell);
-    tx.commit();
+    await tx.commit();
 
     await runtime.idle();
+
+    // This tests the condition that caused the conflict
+    expect(result.get().sequence).toBe(undefined);
 
     // Trigger the handler
     result.key("action").send({});
     await runtime.idle();
 
-    expect(result.get()).toMatchObject({ action: expect.anything() });
+    // This tests the condition that caused the conflict
+    expect(result.get().sequence).toBe(null);
 
     // Give time for async conflict notifications to be processed
     // The conflict happens during the optimistic transaction retry,
@@ -164,6 +168,7 @@ describe("Conflict Reproduction", () => {
             items,
             sequence,
           }),
+          sequence,
         };
       },
     );
@@ -177,13 +182,19 @@ describe("Conflict Reproduction", () => {
     const result = runtime.run(tx, conflictReproNoLift, {
       items: [{ id: "test" }],
     }, resultCell);
-    tx.commit();
+    await tx.commit();
 
     await runtime.idle();
+
+    // This tests the condition that caused the conflict
+    expect(result.get().sequence).toBe(undefined);
 
     // Trigger the handler
     result.key("action").send({});
     await runtime.idle();
+
+    // This tests the condition that caused the conflict
+    expect(result.get().sequence).toBe(null);
 
     // Give time for async conflict notifications
     await new Promise((resolve) => setTimeout(resolve, 100));


seefeldb:
## Scheduler Analysis

Looking at the scheduler code, the execution sequence is:

### Event Queue Processing (Handler Invocation)

From [scheduler.ts:487-536](https://github.com/commontoolsinc/labs/blob/test/minimal-conflict-repro/packages/runner/src/scheduler.ts#L487-L536):

```typescript
const event = this.eventQueue.shift();
if (event) {
  const finalize = (error?: unknown) => {
    // ...
    tx.commit().then(({ error }) => {  // Line 498 - NOT awaited!
      if (error && retriesLeft > 0) {
        this.eventQueue.unshift({ action, retriesLeft: retriesLeft - 1, onCommit });
        this.queueExecution();
      }
    });
  };
  
  this.runningPromise = Promise.resolve(
    this.runtime.harness.invoke(() => action(tx))
  ).then(() => finalize()).catch((error) => finalize(error));
  await this.runningPromise;  // Line 532 - waits for handler to finish
}
```

The handler runs and finishes (line 532), but `tx.commit()` at line 498 is **not awaited** - it returns a promise that's handled asynchronously.

### Reactive Actions Processing

From [scheduler.ts:538-565](https://github.com/commontoolsinc/labs/blob/test/minimal-conflict-repro/packages/runner/src/scheduler.ts#L538-L565):

```typescript
const order = topologicalSort(this.pending, this.dependencies);

for (const fn of order) {
  // ...
  await this.run(fn);  // Line 563
}
```

The reactive actions (lift recomputation) run immediately after the handler, each calling `this.run()` which also has the same pattern - `tx.commit()` is called but not awaited at [line 250](https://github.com/commontoolsinc/labs/blob/test/minimal-conflict-repro/packages/runner/src/scheduler.ts#L250).

### The Design Intent

Comments at [scheduler.ts:244-249](https://github.com/commontoolsinc/labs/blob/test/minimal-conflict-repro/packages/runner/src/scheduler.ts#L244-L249) explain:

> "Commit the transaction. The code continues synchronously after kicking off the commit, i.e. it assumes the commit will be successful."

This is the correct design - the local nursery should be updated synchronously, allowing subsequent transactions to see the changes immediately. The async part is just remote confirmation.

### The Question

If the nursery update is synchronous (which it is, at [cache.ts:990](https://github.com/commontoolsinc/labs/blob/test/minimal-conflict-repro/packages/runner/src/storage/cache.ts#L990)), and reads check the nursery first (which they do, at [cache.ts:1142](https://github.com/commontoolsinc/labs/blob/test/minimal-conflict-repro/packages/runner/src/storage/cache.ts#L1142)), then why does TX-4 have the wrong expected version?

The conflict error shows TX-4 expected `ba4jcaqmalkmpvecgzokt3rmw2ryxuyidyuoeuyaosrt3xtygk6c56tol` but the actual was `ba4jcaghsobeyzssva2kixwllrth3pocgb3uztbfkz2gqvjzr7i4yxmox`, meaning TX-4 didn't see TX-3's update.

seefeldb:
## Deep Dive: Transaction Snapshot Capture

I've traced through the entire read/write flow to understand when the "expected version" is captured. Here's what happens:

### Write Flow

When a transaction writes to a cell:

1. **journal.ts:140** - `branch.write(address, value)` 
2. **chronicle.ts:139** - `const state = this.load(address)` - loads from replica
3. **chronicle.ts:91** - `return this.#replica.get(address)` - **reads from nursery first!**
4. **chronicle.ts:140** - `const loaded = attest(state)` - creates attestation with current cause
5. This attestation's `cause` becomes the "expected version" for the SQLite transaction

### Key Insight

The expected version is captured when `Chronicle.write()` calls `this.load()`, which calls `replica.get()`. The replica DOES check the nursery first ([cache.ts:1142](https://github.com/commontoolsinc/labs/blob/test/minimal-conflict-repro/packages/runner/src/storage/cache.ts#L1142-L1143)).

### The Timing Mystery

Given this flow, TX-4 (lift) should see TX-3's (handler) changes:

1. TX-3 handler commits → updates nursery at [cache.ts:990](https://github.com/commontoolsinc/labs/blob/test/minimal-conflict-repro/packages/runner/src/storage/cache.ts#L990) (synchronous)
2. TX-3 sends notification at [cache.ts:992-997](https://github.com/commontoolsinc/labs/blob/test/minimal-conflict-repro/packages/runner/src/storage/cache.ts#L992-L997) (synchronous)
3. TX-4 gets triggered and runs
4. TX-4 creates Chronicle branch at [journal.ts:164](https://github.com/commontoolsinc/labs/blob/test/minimal-conflict-repro/packages/runner/src/storage/transaction/journal.ts#L164)
5. TX-4 writes → loads from replica.get() → **should see nursery update** ✓

But the conflict shows TX-4 has the OLD expected version! 

### Hypothesis

The async `mount()` call at [memory.ts:181](https://github.com/commontoolsinc/labs/blob/test/minimal-conflict-repro/packages/memory/memory.ts#L181) might be creating a timing window where both transactions race to SQLite with stale snapshots captured before the nursery updates propagate correctly.

Need to verify: Is there any caching or snapshot isolation at the replica level that could cause stale reads even though nursery is updated?

seefeldb:
## ROOT CAUSE IDENTIFIED: Concurrent Remote Commits

After adding detailed logging, I found the exact issue. Here's the sequence with the actual causes:

### The Actual Sequence

```
[Handler TX-3]
- Writes to nursery with cause: ba4jcaghso...  (TX-3's version)
- Notification sent → triggers TX-4

[Lift TX-4 - First Attempt]
- Reads base from nursery: ba4jcaghso... ✓ (correctly sees TX-3)
- Writes to nursery with cause: ba4jcaqmalk... (TX-4's version)
- Tries remote/SQLite commit → ConflictError!
- Changes REMOVED from nursery (cache.ts:1050)

[Lift TX-5 - Retry]
- Reads base from nursery: ba4jcaghso... (TX-4's changes were purged!)
- Tries to commit expecting ba4jcaqmalk... but DB has ba4jcaghso...
- ConflictError again!
```

### The Root Problem

**TX-3 and TX-4 are both trying to commit to SQLite concurrently**, even though:
1. TX-3 adds to nursery synchronously ✓
2. TX-4 sees TX-3's nursery update ✓  
3. TX-4 uses correct base version ✓

But then:
4. TX-3's `remote.transact()` is async and hasn't reached SQLite yet
5. TX-4's `remote.transact()` also starts, goes async
6. **Both race to SQLite** with overlapping version expectations
7. One wins, one gets ConflictError

### Why the Nursery Update Isn't Enough

The nursery update at [cache.ts:990](https://github.com/commontoolsinc/labs/blob/test/minimal-conflict-repro/packages/runner/src/storage/cache.ts#L990) happens immediately, which is correct for optimistic reads.

But the **remote/SQLite commit** at [cache.ts:1007-1011](https://github.com/commontoolsinc/labs/blob/test/minimal-conflict-repro/packages/runner/src/storage/cache.ts#L1007-L1011) happens asynchronously via `await commitPromise`.

The async `mount()` at [memory.ts:181](https://github.com/commontoolsinc/labs/blob/test/minimal-conflict-repro/packages/memory/memory.ts#L181) creates a timing window where multiple transactions can reach the SQLite layer concurrently, all with overlapping version expectations.

### The Fix Direction

The system needs to serialize remote commits, not just nursery updates. Possible approaches:
1. Add a commit queue to ensure SQLite transactions happen sequentially
2. Make the scheduler wait for remote commit before triggering reactive actions
3. Improve the conflict retry logic to update the base version from the conflict error's `actual` field

The current retry logic purges from nursery and retries with stale data, creating an infinite loop of conflicts.

seefeldb:
## Root Cause Identified: Nursery and SQLite Out of Sync

I've added comprehensive logging to track the exact sequence of SQLite SWAP operations and nursery operations. The logs reveal the root cause:

### The Sequence

**Nursery operations (optimistic, immediate):**
1. TX-2 commits to nursery: `ba4jcagwa2ku...`
2. TX-3 reads `ba4jcagwa2ku...` from nursery, commits: `ba4jcaghsobeyzss...`
3. TX-4 reads `ba4jcaghsobeyzss...` from nursery, commits: `ba4jcaqmalkmpvec...`

**SQLite batch commit (happens later, asynchronously):**
```
[SQLITE SWAP BEFORE] Expected cause: ba4jcapjcdszj... → updated=0 (failed)
[SQLITE SWAP BEFORE] Expected cause: ba4jcagwa2ku... → updated=1 (success)
[SQLITE SWAP BEFORE] Expected cause: ba4jcaghsobeyzss... → updated=1 (success)
[SQLITE SWAP BEFORE] Expected cause: ba4jcaqmalkmpvec... → updated=0 (CONFLICT!)
[SQLITE CONFLICT] Expected: ba4jcaqmalkmpvec... Actual in DB: ba4jcabew6pg72ji...
```

### The Problem

When TX-4 tries to commit to SQLite, it expects the nursery version (`ba4jcaqmalkmpvec...`) to be in the database. However, SQLite contains a **completely different version** (`ba4jcabew6pg72ji...`) that was never in the nursery!

**This means another transaction committed to SQLite between:**
1. When TX-4 added its changes to the nursery
2. When TX-4 attempted to commit to SQLite

The nursery is supposed to be a staging area that mirrors what will be in SQLite, but they're getting out of sync. A transaction that's not in the nursery is making it into SQLite first.

### Next Step

Need to identify:
1. Where does `ba4jcabew6pg72ji...` come from?
2. Why is it committing to SQLite without going through the nursery?
3. Is there a race condition where multiple SQLite transactions are running concurrently?

seefeldb:
## Root Cause Found: Client Sends Different Base Than What It Refers To

After logging the exact values used to compute cause hashes on both sides, I found the smoking gun:

### Client TX-3
**Fact being referred (base):**
```json
"argument": {
  "items": [{"id": "test"}]  ← Base has test item
}
```

**New value being asserted:**
```json
"argument": {
  "items": []  ← New value is empty
}
```

**Computed cause:** `ba4jcaghsobeyzss...` ← Hash of the BASE fact

### Server TX-3  
**Assertion received:**
```json
"argument": {
  "items": []  ← Only receives the new value
}
```

**Computed fact:** `ba4jcabew6pg72ji...` ← Hash of the NEW value (different!)

### The Problem

The client computes its cause hash by calling `refer(factToRefer)` where `factToRefer.is` contains the BASE state (`items: [{id: "test"}]`).

But the server receives only the NEW state (`items: []`) in `source.assert.is`, and computes `refer(source.assert)` which hashes the NEW value, not the base.

**This is the mismatch:** 
- Client cause = hash(base fact with items=[{id:"test"}])  
- Server fact = hash(new assertion with items=[])

The protocol appears to assume the client sends both:
1. The assertion's new value
2. The base fact it's building on (via the `cause` reference)

But somewhere in the transmission, the server is using `source.assert` to compute the fact hash instead of looking up what the `cause` reference points to.

This explains why the mystery transaction `ba4jcabew6pg72ji...` appears - it's not a mystery, it's just the server computing the wrong hash because it's hashing the assertion instead of the base!

seefeldb:
## Root Cause Found!

I've identified the exact cause of the ConflictError:

### The Problem
On the **client side**, the assertion contains `internal.__#0 = NaN`
On the **server side**, after deserialization, it becomes `internal.__#0 = null`

### Evidence
```
CLIENT:
Raw value: NaN
typeof: number
JSON.stringify: null

SERVER:
Raw value: null
typeof: object  
JSON.stringify: null
```

Even though both `JSON.stringify()` to `"null"`, they are fundamentally different values:
- `NaN` is a number type
- `null` is an object type

When these objects are passed to `refer()` for merkle hashing, they produce different content hashes:
- Client hash: `ba4jcaqmalkmpvecgzokt3rmw2ryxuyidyuoeuyaosrt3xtygk6c56tol`
- Server hash: `ba4jcabew6pg72jigsqiib6dqchn6h3q7fxv7btn7tbmedaetcmmc6ukv`

This causes the optimistic locking conflict because the server receives an assertion with a different hash than it computed.

### Next Steps
Need to find where `NaN` is being set as a value in the pattern execution. This appears to be related to the `lift()` function creating intermediate computation results with uninitialized numeric fields.

seefeldb:
## Summary of Investigation

### Root Cause Identified
The ConflictError is caused by `NaN` (Not a Number) being created on the client side, which gets serialized to `null` and produces a different merkle hash than if `null` was used directly.

### The Chain of Events
1. A field `internal.__#0` exists with schema `{ type: "number" }` but no explicit default value
2. Somewhere in the initialization code, this field gets set to `Number(undefined)` which produces `NaN`
3. When the assertion is created, it contains `NaN`
4. Client computes hash: `ba4jcaqmalkmpvecgzokt3rmw2ryxuyidyuoeuyaosrt3xtygk6c56tol`
5. During JSON serialization (for network transmission), `NaN` becomes `null`
6. Server deserializes and gets `null` (object type, not `NaN`)
7. Server computes hash: `ba4jcabew6pg72jigsqiib6dqchn6h3q7fxv7btn7tbmedaetcmmc6ukv`
8. Hashes don't match → ConflictError

### Why It Only Happens With `lift()`
The `lift()` function creates intermediate computation results with internal state. The `__#0` field appears to be part of this internal state and gets initialized incorrectly.

### Reproduction
Confirmed in the minimal test case:
- **With lift:** Conflicts occur
- **Without lift:** No conflicts

### Next Steps to Fix
Need to find where numeric fields without explicit defaults are being initialized with `Number(undefined)` instead of leaving them as `undefined` or using a proper default like `0` or `null`.

Potential locations to investigate:
1. CFC (Compile From Code) - where TypeScript types are converted to runtime schemas
2. Schema default value processing
3. Internal state initialization for lift/handler constructs

The fix will likely be to either:
- Not call `Number()` on undefined values
- Provide explicit defaults for numeric fields
- Handle `NaN` specially in serialization (convert to `null` before hashing)

seefeldb:
✅ Fix has been committed and pushed in 7b4f0365e

The ConflictError issue is now resolved. All tests pass with 0 conflicts.

PR #2045, opened 2025-11-11
feat: ACLs in spaces.

jsantell:
* Server/provider has a single, global namespace. All spaces created are private by default. Tilte `~spacenames` will have no special meaning.
  * derived from (public) space name for now representing user owning of
    space. Eventually we'll have a look up table to not need publicly
derivable spaces.
* Add a well-known `ACL` cell to each Space. This contains a `Map<DID, CAPABILITY>` that applies to the current space.
* User transactions are signed with the User key directly (`session.as`).
* Each transaction requires the signer to either be the space identity, or an identity located in the `ACL`, or have `ANYONE` identity in the `ACL`.
  * This changes the previous "personas" indirection in signing, but will simplify the `ACL` stored identities.
  * To remedy the shortcomings of [1], only allow the space identity to write if and only if it's writing the initial `ACL` value.
* Add to CLI ability to add identities to a space with a capability.
* Log current user's identity in browser console for sharing.

---
## Summary by cubic
Adds per-space ACLs (READ/WRITE/OWNER) and enforces them across memory requests. On first connect, spaces auto-create an ACL granting OWNER to the user and WRITE to ANYONE; transactions are signed with the user key, and a new CLI manages ACL entries.

- **New Features**
  - ACL stored in a well-known cell per space; supports DID entries and ANYONE.
  - Provider checks owner, service DID, or ACL capability on requests.
  - ACL auto-initializes on first connect, granting OWNER to the user and WRITE to ANYONE.
  - New CLI: ct acl ls | set | remove.
  - CharmsController exposes an ACL manager (get/set/remove).
  - Runtime supports async disposal; browser logs user DID and space.

- **Migration**
  - Transactions must be signed as the user (no personas).
  - Spaces start with ANYONE WRITE; remove or change via ACL to make private.
  - Public “~space” semantics removed—use ACL to share.

<sup>Written for commit 1ab3f351465dda27b35ff5fc259444e08d7a8e7e. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/cli/commands/acl.ts
+++ b/packages/cli/commands/acl.ts
@@ -1,0 +1,126 @@
+import { Command } from "@cliffy/command";
+import { Table } from "@cliffy/table";
+import {
+  getAcl,
+  removeAclEntry,
+  setAclEntry,
+  SpaceConfig,
+} from "../lib/acl.ts";
+import { render } from "../lib/render.ts";
+import { isCapability } from "@commontools/memory";
+
+// Usage patterns for examples
+const spaceUsage = `--identity <identity> --api-url <api-url> --space <space>`;
+
+export const acl = new Command()
+  .name("acl")
+  .description("Manage Access Control Lists for spaces.")
+  .default("help")
+  .globalEnv("CT_API_URL=<url:string>", "URL of the fabric instance.", {
+    prefix: "CT_",
+  })
+  .globalOption("-a,--api-url <url:string>", "URL of the fabric instance.")
+  .globalEnv("CT_IDENTITY=<path:string>", "Path to an identity keyfile.", {
+    prefix: "CT_",
+  })
+  .globalOption("-i,--identity <path:string>", "Path to an identity keyfile.")
+  .globalOption("-s,--space <space:string>", "The space name or DID")
+  /* acl ls */
+  .command("ls", "List all ACL entries for a space.")
+  .usage(spaceUsage)
+  .example(
+    "ct acl ls --identity ./my.key --api-url https://api.example.com --space my-space",
+    "List all ACL entries for my-space",
+  )
+  .action(async (options) => {
+    const config = parseSpaceOptions(options);
+    const aclData = await getAcl(config);
+
+    if (!aclData || Object.keys(aclData).length === 0) {
+      render("No ACL entries found.");
+      return;
+    }
+
+    new Table()
+      .header(["DID", "CAPABILITY"])
+      .body(
+        Object.entries(aclData).map(([did, capability]) => [did, capability]),
+      )
+      .border(true)
+      .render();
+  })
+  /* acl set */
+  .command(
+    "set <did:string> <capability:string>",
+    "Applies a capability (READ, WRITE, OWNER) to an identity (DID) to the space ACL.",
+  )
+  .usage(`${spaceUsage} <did> <capability>`)
+  .example(
+    "ct acl set did:key:z6Mkk... WRITE --identity ./my.key --api-url https://api.example.com --space my-space",
+    "Grant WRITE capability to a DID",
+  )
+  .action(async (options, did, capability) => {
+    const cap = capability.toUpperCase();
+    if (!isCapability(cap)) {
+      render(
+        `Invalid capability: ${capability}. Must be one of: READ, WRITE, OWNER`,
+      );
+      Deno.exit(1);
+    }
+
+    const config = parseSpaceOptions(options);
+    await setAclEntry(
+      config,
+      did,
+      cap as "READ" | "WRITE" | "OWNER",
+    );
+    render(`Added ${did} with ${cap} capability`);
+  })
+  /* acl remove */
+  .command("remove <did:string>", "Remove a DID from the space ACL.")
+  .usage(`${spaceUsage} <did>`)
+  .example(
+    "ct acl remove did:key:z6Mkk... --identity ./my.key --api-url https://api.example.com --space my-space",
+    "Remove a DID from the ACL",
+  )
+  .action(async (options, did) => {
+    const config = parseSpaceOptions(options);
+    await removeAclEntry(config, did);
+    render(`Removed ${did} from ACL`);
+  });
+
+/**
+ * Parse space-related options from command arguments
+ */
+function parseSpaceOptions(
+  options: Record<string, string | undefined>,
+): SpaceConfig {
+  const apiUrl = options.apiUrl || Deno.env.get("CT_API_URL");
+  const identity = options.identity || Deno.env.get("CT_IDENTITY");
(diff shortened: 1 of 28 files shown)


packages/memory/provider.ts
```diff
@@ -537,6 +542,48 @@ class MemoryProviderSession<
...
+      };
+
+      const aclFact = selectFact(spaceSession, {
+        the: ACL_TYPE,
+        of: ACL_ENTITY,
+      });
+
+      if (!aclFact || !aclFact.is) {
+        logger.warn(
+          () => ["No ACL defined:", space],
```
jsantell: This is a loud log in the tests

packages/memory/provider.ts
```diff
@@ -537,6 +542,48 @@ class MemoryProviderSession<
...
+        );
+        return undefined;
+      }
+
+      const spaceSession = result.ok as unknown as {
+        subject: MemorySpace;
+        store: any;
+      };
+
+      const aclFact = selectFact(spaceSession, {
```
jsantell: This seems to always be `undefined`

packages/runner/src/storage/cache.ts
```diff
@@ -1223,6 +1227,35 @@ export class Replica {
...
+    const userIdentity = (this.remote as any).as as Signer | undefined;
+    if (!userIdentity) {
+      return;
+    }
+
+    const initialACL: JSONValue = {
+      [userIdentity.did()]: "OWNER",
+    };
+
+    const aclFact = assert({
```
jsantell: How can `this.spaceIdentity` be used to sign this transaction? @ubik2

packages/charm/src/manager.ts
```diff
@@ -1113,6 +1113,14 @@ export class CharmManager {
     await this.runtime.idle();
     await this.synced();
   }
+
+  async dispose() {
```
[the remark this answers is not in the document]
jsantell: From the "runtime user" perspective (shell, CLI), most interactions are through CharmManager (or the newer CharmsController, which wraps the Manager, which wraps the Runtime..)

packages/cli/deno.json
```diff
@@ -4,7 +4,8 @@
     "cli": "ROOT=$(pwd) && cd $INIT_CWD && deno run --allow-net --allow-ffi --allow-read --allow-write --allow-env \"$ROOT/mod.ts\"",
     "cli-no-pwd-override": "deno run --allow-net --allow-ffi --allow-read --allow-write --allow-env ./mod.ts",
     "test": "deno test --allow-ffi --allow-read --allow-write --allow-run --allow-env test/*.test.ts",
-    "integration": "CT_CLI_INTEGRATION_USE_LOCAL=1 LOG_LEVEL=warn ./integration/integration.sh"
+    "integration": "CT_CLI_INTEGRATION_USE_LOCAL=1 LOG_LEVEL=warn ./integration/integration.sh",
+    "acl-integration": "CT_CLI_INTEGRATION_USE_LOCAL=1 LOG_LEVEL=warn ./integration/acl.sh"
```
jsantell: This isn't run anywhere yet, as the assertions here aren't yet true, but cover the e2e scenario

packages/cli/commands/acl.ts
```diff
@@ -0,0 +1,126 @@
...
+      .render();
+  })
+  /* acl add */
+  .command(
+    "add <did:string> <capability:string>",
+    "Add a DID to the space ACL with specified capability (READ, WRITE, or OWNER).",
+  )
+  .usage(`${spaceUsage} <did> <capability>`)
+  .example(
+    "ct acl add did:key:z6Mkk... WRITE --identity ./my.key --api-url https://api.example.com --space my-space",
```
jsantell: Maybe should be `s/add/set`

packages/charm/src/manager.ts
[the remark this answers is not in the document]
ubik2: Maybe add a comment to CharmManager constructor to indicate that it takes ownership of the runtime.
Normally, I'd push for RAII here (and have the CharmManager create the Runtime), but constructing the runtime involves a *lot* of parameters, so coupling all of that configuration with charm manager configuration gets ugly.

packages/cli/commands/acl.ts
ubik2: `add` is the natural counterpart to `remove`, but since we're not using a list of capabilities, in this context, I think `set` is better. The natural counter to `set` is probably `clear`, but I think `remove` is better.

packages/shell/src/lib/runtime.ts
```diff
@@ -86,6 +86,10 @@ export class RuntimeInternals extends EventTarget {
   ): Promise<RuntimeInternals> {
     const session = await createSession({ identity, spaceName });
 
+    // Log user identity for debugging and sharing
```
ubik2: Fine to have the console.log temporarily, but we should either use a logger here if we want to keep it around, or remember to clear it out once we've had a bit of testing.

packages/cli/lib/acl.ts
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Add or update an ACL entry for a DID
+ */
+export async function addAclEntry(
+  config: SpaceConfig,
+  user: string,
+  capability: Capability,
+): Promise<void> {
+  const did = userToDid(user);
+  await using manager = await createManager(config);
```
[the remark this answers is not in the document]
jsantell: As a middle ground, the `AclManager` could take only a runtime instead, and we directly instantiate that instead of CharmsController (which requires CharmManager, which has some big changes in store for it) for the CLI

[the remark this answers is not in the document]
jsantell: Updating ^

packages/runner/src/storage/cache.ts
```diff
@@ -1223,6 +1227,64 @@ export class Replica {
...
+    const originalSpaceSigner = this.remote.as;
+    const originalConsumerSigner = remoteSession?.as;
+
+    this.remote.as = this.spaceIdentity;
+    if (remoteSession) {
+      remoteSession.as = this.spaceIdentity;
+    }
+
+    try {
+      await this.commit({ facts: [aclFact], claims: [] });
```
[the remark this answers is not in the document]
jsantell: Sounds good, removing the earlier `get()` and catching errors thrown from here

packages/runner/src/storage/cache.ts
```diff
@@ -667,6 +669,8 @@ export class Replica {
   }
 
   async poll() {
```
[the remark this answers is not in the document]
ubik2: I opened https://linear.app/common-tools/issue/CT-1044/providermount-and-providerpoll-are-not-async for the poll issue.

PR #2057, opened 2025-11-12
fix(transformer): Fix synthetic derive node detection for computedtransformer

seefeldb:
Fixed a bug where .map on closed-over OpaqueRef inside computed() was incorrectly transformed to mapWithPattern. The issue occurred because:

1. ComputedTransformer converts computed(() => items.map(...)) to derive({}, () => items.map(...))
2. Inside a derive callback, OpaqueRef arrays should use plain .map, not .mapWithPattern (since they're already unwrapped)
3. The existing shouldTransformMap only checked JSX contexts, missing direct derive callbacks

Solution:
- Added isMapInsideDeriveCallback() to detect when a map is inside a derive callback (works with both explicit and synthetic derive nodes)
- Uses isDeriveCall() which has text-based fallback detection for synthetic "derive" identifiers
- Uses tree traversal in isNodeDescendantOf() instead of parent references for robustness with synthetic nodes
- shouldTransformMap() now checks derive context before JSX context

Added test fixtures:
- computed-with-closed-over-opaque-ref-map: reproduces the bug
- derive-with-closed-over-opaque-ref-map: ensures explicit derive works

---
## Summary by cubic
Fixes incorrect transformation of .map on closed-over OpaqueRef arrays inside computed()/derive callbacks. Prevents converting to mapWithPattern when arrays are already unwrapped.

- **Bug Fixes**
  - Use context tracking (deriveCallbackStack) to detect maps inside derive callbacks, including synthetic ones from computed.
  - Skip transformation when the map target is captured (not a derive parameter) by comparing root identifier to callback parameters.
  - Use isDeriveCall for synthetic derive detection; update fixtures for computed/derive with closed-over OpaqueRef maps (non-shorthand object literals).

<sup>Written for commit 44f8616ac979081ec485d6a526dd1acb86008fd4. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/ts-transformers/src/closures/transformer.ts
+++ b/packages/ts-transformers/src/closures/transformer.ts
@@ -759,85 +759,19 @@ function buildHandlerStateTypeNode(
 }
 
 /**
- * Check if a map call is inside a derive callback.
- * This handles both explicit derive calls and synthetic ones created by computed transformer.
- *
- * When a map is inside a derive callback, it operates on already-unwrapped arrays,
- * so it should NOT be transformed to mapWithPattern.
- */
-function isMapInsideDeriveCallback(
-  mapCall: ts.CallExpression,
-  context: TransformationContext,
-): boolean {
-  let node: ts.Node = mapCall;
-
-  while (node.parent) {
-    // Check if parent is a call expression
-    if (ts.isCallExpression(node.parent)) {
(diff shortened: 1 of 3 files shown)


mathpirate:
fixed this in a different way in https://github.com/commontoolsinc/labs/pull/2058/files

PR #2067, opened 2025-11-13
Use state machine for `fetchData`

bfollington:
- **Refactor fetchData to match fetchProgram state machine**
- **Format**

---
## Summary by cubic
Refactored fetchData to use the same async state machine as fetchProgram, with CAS-safe transitions and timeout handling for more reliable cross-tab behavior and clearer pending/result/error states.

- **New Features**
  - Added a shared async-operation-state module with idle/fetching/success/error states, CAS transitions (success/error/idle), timeout checks, and a cache schema.

- **Refactors**
  - Rewrote fetchData to use the shared state machine and a cache cell (replacing the old internal cell and mutex-based flow).
  - Standardized output derivation from state (pending/result/error) and added a 10s data fetch timeout.
  - Ensured aborts transition fetching requests back to idle.
  - Updated fetchProgram to use the shared state machine, removing duplicated logic and unifying error handling.

<sup>Written for commit 7d2c3fa923f9736b73f177d7c95afab9f733890a. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/runner/src/builtins/fetch-program.ts
+++ b/packages/runner/src/builtins/fetch-program.ts
@@ -6,8 +6,8 @@ import { HttpProgramResolver } from "@commontools/js-runtime";
 import { resolveProgram, TARGET } from "@commontools/js-runtime/typescript";
 import { computeInputHash } from "./fetch-utils.ts";
 import {
-  asyncOperationCacheSchema,
   type AsyncOperationCache,
+  asyncOperationCacheSchema,
   getState,
   isTimedOut,
   transitionToError,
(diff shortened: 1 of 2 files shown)


packages/runner/src/builtins/fetch-program.ts
```diff
@@ -85,7 +79,7 @@ export function fetchProgram(
       }
 
       if (Object.keys(updates).length > 0) {
-        cache.withTx(tx).update(updates);
+        (cache as any).withTx(tx).update(updates);
```
bfollington: I still don't see why this needs to happen

PR #2159, opened 2025-11-25
Implement home space "favorites list"

bfollington:
- **Implement `favorites` cell concept, pointed to by the space cell of the home space**
- **Access user DID via IRuntime**
- **Grant ACL for the home space on creation**

Unsure on organization & why the tests are failing.

### DIFF
--- a/packages/runner/src/runtime.ts
+++ b/packages/runner/src/runtime.ts
@@ -234,18 +234,7 @@ export interface IRuntime {
     tx?: IExtendedStorageTransaction,
   ): Cell<T>;
 
-  // Home space helpers (space where space DID = user identity DID)
-  getHomeCell<S extends JSONSchema = JSONSchema>(
-    cause: any,
-    schema: S,
-    tx?: IExtendedStorageTransaction,
-  ): Cell<Schema<S>>;
-  getHomeCell<T>(
-    cause: any,
-    schema?: JSONSchema,
-    tx?: IExtendedStorageTransaction,
-  ): Cell<T>;
-
+  // Home space helper (space where space DID = user identity DID)
   getHomeSpaceCell(
     tx?: IExtendedStorageTransaction,
   ): Cell<HomeSpaceCellContents>;
@@ -706,25 +695,6 @@ export class Runtime implements IRuntime {
       type: "application/json",
       schema,
     }, tx);
-  }
-
-  // Home space helper (space where space DID = user identity DID)
-  getHomeCell<T>(
-    cause: any,
-    schema?: JSONSchema,
-    tx?: IExtendedStorageTransaction,
-  ): Cell<T>;
-  getHomeCell<S extends JSONSchema = JSONSchema>(
-    cause: any,
-    schema: S,
-    tx?: IExtendedStorageTransaction,
-  ): Cell<Schema<S>>;
-  getHomeCell(
-    cause: any,
-    schema?: JSONSchema,
-    tx?: IExtendedStorageTransaction,
-  ): Cell<any> {
-    return this.getCell(this.userIdentityDID, cause, schema, tx);
   }
 
   getHomeSpaceCell(


packages/charm/src/manager.ts
```diff
@@ -1098,6 +1099,104 @@ export class CharmManager {
     await this.runtime.idle();
     await this.synced();
   }
+
+  /**
+   * Get the favorites cell from the home space (singleton across all spaces)
+   * @private
+   */
+  private getHomeFavorites(): Cell<Cell<unknown>[]> {
```
bfollington: Should these live elsewhere? They feel "Charm Manager"-y to me, but could have a new favorites module?

packages/runner/src/storage/cache.ts
```diff
@@ -1242,15 +1242,21 @@ export class Replica {
...
       return;
     }
 
+    // Request ACL in two cases:
+    // 1. When spaceIdentity exists (delegated space)
+    // 2. When user's identity equals the space DID (home space)
+    const isHomeSpace = userIdentity.did() === this.space;
+    if (!this.spaceIdentity && !isHomeSpace) {
+      return;
+    }
```
bfollington: More documentation of reasoning needed? @jsantell?

jsantell: This looks good :+1:

packages/runner/src/storage/cache.ts
```diff
@@ -1265,15 +1271,17 @@ export class Replica {
       cause: null,
     });
 
-    // Temporarily swap the signer to use spaceIdentity for this commit
+    // Temporarily swap the signer to use spaceIdentity (or userIdentity for home space) for this commit
```
jsantell: This signer swap should *only* be used when `this.spaceIdentity` is used -- the `userIdentity` is already set to sign the transaction (the e.g. `remote.as`)

packages/runner/src/storage/cache.ts
```diff
@@ -1965,14 +1973,22 @@ export class StorageManager implements IStorageManager {
 
   protected connect(space: MemorySpace): IStorageProviderWithReplica {
     const { id, address, as, settings } = this;
+
+    // Determine the correct spaceIdentity for this space
+    // For the workspace space (where spaceIdentity.did() === space), use spaceIdentity
+    // For other spaces (like home space), pass undefined
+    const spaceIdentityForSpace = this.spaceIdentity?.did() === space
```
jsantell: Ah, is this because we open multiple spaces here, and `spaceIdentity` set at the root session shouldn't affect other spaces? Will be good to remove the spaceIdentity throughout soon

jsantell (approved):
LGTM!

packages/charm/src/manager.ts
```diff
@@ -1098,6 +1099,104 @@ export class CharmManager {
...
+
+    return removed;
+  }
+
+  /**
+   * Check if a charm is in the user's favorites (in home space)
+   * @param charm - The charm or entity ID to check
+   * @returns true if the charm is favorited, false otherwise
+   */
+  isFavorite(charm: Cell<unknown> | EntityId): boolean {
```
seefeldb: Is the `| EntityId` necessary here? We should remove those eventually so if the call site doesn't need it, let's not add it. Otherwise keep it.

packages/runner/src/runtime.ts
```diff
@@ -87,11 +87,13 @@ export interface SpaceCellContents {
   allCharms: Cell<never>[];
   recentCharms: Cell<never>[];
   defaultPattern: Cell<never>;
+  favorites: Cell<Cell<never>[]>;
```
seefeldb: We should maybe make a HomeSpaceCellContent type since this is only valid there? (And arguably the others here aren't relevant to people accessing the home space from other spaces)

packages/runner/src/runtime.ts
```diff
@@ -194,6 +217,18 @@ export interface IRuntime {
     tx?: IExtendedStorageTransaction,
   ): Cell<T>;
 
+  // Home space helper (space where space DID = user identity DID)
+  getHomeCell<S extends JSONSchema = JSONSchema>(
```
seefeldb: Nice! But I don't think manager.ts uses these yet?

packages/runner/src/storage/cache.ts
```diff
@@ -1242,15 +1242,21 @@ export class Replica {
...
+    // Request ACL in two cases:
+    // 1. When spaceIdentity exists (delegated space)
+    // 2. When user's identity equals the space DID (home space)
+    const isHomeSpace = userIdentity.did() === this.space;
+    if (!this.spaceIdentity && !isHomeSpace) {
+      return;
+    }
+
+    const aclKey = this.spaceIdentity?.did() ?? this.space;
+
```
seefeldb: For the home space we shouldn't add anyone. And I think we could actually just leave off the ACL for now, since the default behavior is to treat the space key as owner, which here is the user. @jsantell to confirm.

jsantell: That is true, we can skip the ACL entirely, as the ownership check (in memory/claim.ts) is before the ACL check, so may not need any changes to cache.ts (other than [here](https://github.com/commontoolsinc/labs/pull/2159#discussion_r2558279675) handling opening a space e.g. home space when we have a spaceIdentity for some other string-derived space name)

PR #2197, opened 2025-12-04
[WIP] perf(runner): add traverseDAG cache with WeakMap memoization

willkelly:
## Problem

CPU pegging at 100% when displaying many charms (7+) due to excessive traverseDAG calls. The system was making 18,623 traversal calls to traverse only 338 unique objects - a 55x repetition factor.

## Root Cause

The "diamond problem" - when multiple charms link to shared documents, each charm independently re-traversed the same shared documents. The stack-based CompoundCycleTracker prevented cycles (A→B→A) but not re-traversals via different paths.

## Solution

Two-tier WeakMap tracking strategy:
- `inProgress` WeakMap for cycle detection (tracks objects in call stack)
- `completed` WeakMap for result caching (stores references, not copies)
- `schemaKeyCache` WeakMap for JSON.stringify deduplication
- `stringKeyInProgress` Set for cross-document cycle detection
- `stringKeyCompleted` Map for cross-document result caching (per-query, auto-GC'd)

Cross-document keys use only `docId|path` (no schema) - schema-agnostic cycle detection is correct because schema narrowing means same doc can be reached with different schemas.

## Test Results

| Test | Main (No Memo) | Feature (With Memo) | Reduction |
|------|----------------|---------------------|-----------|
| 3-way diamond | 19 calls | 11 calls | **42%** |
| 5-way + 50 items | 771 calls | 163 calls | **79%** |
| 10-way deep nest | 682 calls | 88 calls | **87%** |

Second traversal of same structure:
- Main: 19 calls (full re-traverse)
- Feature: 1 call (100% cache hit)

## Production Impact

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| traverseDAG calls | 18,623 | ~700 | **26.6x fewer** |

## Design Doc

https://gist.github.com/willkelly/f4d3a5b26c63f3346c4dcf8109d55696

## Test Plan

- [x] `deno test -A test/traverse.test.ts` passes
- [x] `deno test -A test/traverse-diamond.test.ts` validates memoization correctness
- [x] `deno bench -A test/traverse.bench.ts` shows 1.91x speedup for cached traversals
- [x] `deno fmt` and `deno lint` pass
- [x] Type checking passes

### DIFF
--- a/packages/runner/test/traverse-diamond.test.ts
+++ b/packages/runner/test/traverse-diamond.test.ts
@@ -1,15 +1,11 @@
 /**
- * Test fixture for the diamond problem in DAG traversal memoization
+ * Correctness tests for DAG traversal memoization (diamond problem fix)
  *
- * Background: When displaying multiple charms that share common documents,
- * the system was making 18,623 traversal calls for only 338 unique objects -
- * a 55x repetition factor causing 100% CPU usage.
+ * These tests verify that shared object references are correctly memoized
+ * during traversal, ensuring the same object appears as the same reference
+ * in the output.
  *
- * This test validates the WeakMap-based memoization fix prevents redundant
- * traversals while correctly handling the diamond pattern.
- *
- * See: commit 7688e5ad - "perf(runner): fix traverseDAG diamond problem with WeakMap memoization"
- * Design doc: https://gist.github.com/willkelly/f4d3a5b26c63f3346c4dcf8109d55696
+ * For performance benchmarks, see: traverse.bench.ts
  */
 
 import { describe, it } from "@std/testing/bdd";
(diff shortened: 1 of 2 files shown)


packages/runner/src/traverse.ts
```diff
@@ -294,24 +395,37 @@ export abstract class BaseObjectTraverser<S extends BaseMemoryAddress> {
     if (isPrimitive(doc.value)) {
       return doc.value;
     } else if (Array.isArray(doc.value)) {
-      using t = tracker.include(doc.value, SchemaAll, doc);
+      const t = tracker.include(doc.value, SchemaAll, doc);
```
jsantell: Switching away from `using` means explicitly calling e.g. `[Symbol.dispose]()` -- could this continue to use that? It would dispose whenever `t` leaves scope. This uses the [explicit resource management](https://v8.dev/features/explicit-resource-management) feature in JS, supported in Deno, and polyfills used when needed in browsers. If automatic disposal isn't desired, consider a simple `dispose()` function instead

jsantell:
This is a very hot function, thanks for taking a look at this area!

PR #2202, opened 2025-12-05
fix for opaqueref type definition to avoid losing null as an option w…

mathpirate:
…hen used with union types

---
## Summary by cubic
Make OpaqueRef<T> non-distributive to preserve null and other union members. Updates transformer expectations to generate correct JSON Schema for unions and booleans.

- **Bug Fixes**
  - Use tuple checks ([T] extends [...]) in OpaqueRef to avoid distributive conditional behavior.
  - Keep union members (e.g., null) intact when wrapping arrays, objects, and branded cells.
  - Update tests: add anyOf for string|null where needed; simplify boolean schemas to type: "boolean"; adjust minor enum/asOpaque cases.

<sup>Written for commit b6f85f7b08ad830e0bdbdc53db57a2ac03fda623. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/ts-transformers/test/fixtures/jsx-expressions/ifelse-undefined-value.expected.tsx
+++ b/packages/ts-transformers/test/fixtures/jsx-expressions/ifelse-undefined-value.expected.tsx
@@ -125,60 +125,37 @@ export default recipe({
         type: "object",
         properties: {
             result: {
-                $ref: "#/$defs/OpaqueCell"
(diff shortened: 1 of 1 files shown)


seefeldb:
lgtm, the [..] looks like a good change in general and IIRC this basically removes the `& null` case as `null` is of type object and so destroys the object, where as all other primitives seem to work?

PR #2204, opened 2025-12-05
initial draft implementation of action

mathpirate:
Now ready for initial review

---
## Summary by cubic
Adds action(), a shorthand for handlers that don’t use the event parameter, with full transformer and schema support. This simplifies writing stateful actions and ensures pattern outputs correctly mark action/handler properties as streams.

- **New Features**
  - Added action(fn) to API and builder; returns a Stream<void>.
  - Transformer rewrites action(() => …) to handler((_, state) => …) with captured state; supports optional event; arrow functions only.
  - Exposed in runner module and types.
  - Runtime guard: action() throws if CTS is not enabled.

- **Bug Fixes**
  - Schema generator now detects callable types that return Stream or Cell and emits { asStream: true } or { asCell: true }.
  - Event-less actions use never for event schema (emits false in JSON Schema).
  - Registers synthetic call return types so pattern result schemas include stream properties; added tests for action and updated nested handler schema fixture.
  - Handler.with, HandlerFactory, and stream() now return Stream to match runtime, fixing result schema generation.

<sup>Written for commit d6457f7eb470932c7b3f0ea97479eee6d6c99314. Summary will update automatically on new commits.</sup>


packages/api/index.ts
```diff
@@ -1129,6 +1129,20 @@ export type HandlerFunction = {
   ): ModuleFactory<StripCell<T>, Stream<StripCell<E>>>;
 };
 
+/**
+ * ActionFunction creates a handler that doesn't use the event parameter.
```
seefeldb: it doesn't use the state parameter, an event should still be possible.

packages/runner/src/builder/module.ts
```diff
@@ -225,3 +226,27 @@ export function derive<In, Out>(...args: any[]): OpaqueRef<any> {
...
+ */
+export function action<T>(
+  fn: (props: T) => void,
+): HandlerFactory<T, void>;
+export function action<T>(
+  fn: (props: T) => void,
+): HandlerFactory<T, void> {
+  // Wrap the action callback to match handler's (event, props) signature
+  // The { proxy: true } option tells handler to accept schema-less callbacks
+  return handler<void, T>((_, props) => fn(props), { proxy: true });
```
seefeldb: oh, that's wrong. but then i don't think this should ever be called directly, as the transformed will rewrite that to `handler`, no?

maybe throw an error here?

(the non-CTS variant that's correct would be `return handler<T, void>((event) => fn(event))` but this is almost never going to be useful since there is nothing `fn` could write to -- so better just throw and remind user to add the CTS enable marker)

packages/runner/test/module.test.ts
```diff
@@ -243,4 +243,53 @@ describe("module", () => {
       expect([...nodes][0].inputs.$event).toBe(stream);
     });
   });
+
+  describe("action function", () => {
+    it("creates a handler factory", () => {
+      const increment = action<{ count: number }>(({ count }) => {
```
seefeldb: this won't test the transformed version (and as mentioned in the comment above, this isn't particularly useful outside of CTS), i'd just remove this test / change to one that checks for this throwing, if you go down that route.

packages/ts-transformers/test/fixtures/closures/action-with-event.input.tsx
```diff
@@ -0,0 +1,16 @@
...
+  data: string;
+}
+
+interface State {
+  value: Cell<string>;
+}
+
+export default pattern<State>(({ value }) => {
+  return {
+    update: action((e: MyEvent) => value.set(e.data)),
```
seefeldb: nice!

let's also test `action<MyEvent>((e) => value.set(e.data))`.

PR #2236, opened 2025-12-09
docs: Add reactive ternary and style pitfalls guide

tonyespinoza1:
## Summary

- Adds documentation for common pitfalls when using reactive values in patterns
- Covers two main failure modes:
  1. Ternary operators always return first branch (reactive proxy is truthy)
  2. Layout style properties (`justifyContent`, `flexDirection`) don't work with reactive values
- Provides solution: use `ifElse` with static styles

## Why This Matters

These pitfalls are non-obvious and can waste significant debugging time. This doc is named to be searchable by LLMs when they encounter related issues.

## Test plan

- [x] Document is readable and well-formatted
- [x] Code examples are correct
- [x] TL;DR at top provides quick answer

---
## Summary by cubic
Adds a docs guide on common reactive pitfalls. Explains why ternaries always take the first branch and why layout styles like justifyContent/flexDirection fail, and shows the fix with ifElse and static styles.

<sup>Written for commit 0a6cc1f3a35bb5869fa30f44c2350d20be76e3d3. Summary will update automatically on new commits.</sup>

### DIFF
--- a/docs/common/REACTIVE_TERNARY_AND_STYLE_PITFALLS.md
+++ b/docs/common/REACTIVE_TERNARY_AND_STYLE_PITFALLS.md
@@ -1,0 +1,223 @@
+# Reactive Ternary and Style Pitfalls
+
+**TL;DR:** Reactive values (derived/cell) are proxies that are always truthy in ternaries. Layout style properties (`justifyContent`, `flexDirection`) don't work with reactive values. Use `ifElse` with static styles instead.
+
+## Quick Reference
+
+```tsx
+// BAD - ternary with reactive value (proxy always truthy)
+const result = reactiveBoolean ? "yes" : "no";  // ALWAYS "yes"
+
+// BAD - reactive value in layout style
+<div style={{ justifyContent: reactiveDerived }}>  // Won't work
+
+// GOOD - use ifElse with static styles
+ifElse(reactiveBoolean,
+  <div style={{ justifyContent: "flex-end" }}>content</div>,
+  <div style={{ justifyContent: "flex-start" }}>content</div>
+)
+```
+
+---
+
+## Background
+
+This document captures pitfalls discovered while developing a chat pattern where messages needed conditional alignment:
+- **My messages**: Blue bubbles, RIGHT-aligned
+- **Other's messages**: Grey bubbles, LEFT-aligned with avatar
+- **System messages**: Centered, italic
+
+## What Didn't Work
+
+### Attempt 1: Ternary Operators with Reactive Values
+
+```tsx
+// BROKEN - proxy is always truthy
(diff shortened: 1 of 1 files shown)


docs/common/REACTIVE_TERNARY_AND_STYLE_PITFALLS.md
```diff
@@ -0,0 +1,223 @@
...
+```tsx
+// BROKEN - proxy is always truthy
+const isMyMessage = derive(meta, m => m?.isMyMessage ?? false);
+
+<div style={{
+  flexDirection: isMyMessage ? "row-reverse" : "row"  // Always evaluates to "row-reverse"
+}}>
+```
+
+**Why it fails:** Derived values are reactive proxies. In JavaScript, any object (including proxies) is truthy. So `isMyMessage ? x : y` ALWAYS returns `x`, regardless of the underlying boolean value.
```
bfollington: This is surprising to me... I was under the impression we would automatically transform the ternary here to an `ifElse` which _should_ work, e.g.

```tsx
const isMyMessage = derive(meta, m => m?.isMyMessage ?? false);

<div style={{
  flexDirection: ifElse(isMyMessage, "row-reverse", "row")
}}>
```

Perhaps we have a compiler bug?

docs/common/REACTIVE_TERNARY_AND_STYLE_PITFALLS.md
```diff
@@ -0,0 +1,223 @@
...
+});
+
+// Extract as reactive value
+const justifyContent = derive(meta, m => m?.justifyContent ?? "flex-start");
+
+// Use in style
+<div style={{ justifyContent }}>  // STILL BROKEN
+```
+
+**Why it fails:** Even though `justifyContent` correctly resolves to `"flex-end"` or `"flex-start"` as a string, when used as a style property value, layout properties like `justifyContent` don't work with reactive values. The CSS property receives the proxy object, not the string.
```
bfollington: This **definitely** seems like a bug to me. Very surprising.

docs/common/REACTIVE_TERNARY_AND_STYLE_PITFALLS.md
```diff
@@ -0,0 +1,223 @@
...
+
+### 1. Never Use Reactive Values in Ternary Operators
+
+```tsx
+// BAD - proxy is always truthy
+const color = isActive ? "blue" : "gray";
+
+// GOOD - use ifElse
+{ifElse(isActive, <BlueComponent />, <GrayComponent />)}
+```
```
bfollington: Again, I thought we were automatically transforming this...

PR #2289, opened 2025-12-16
fix(runner): update error messages to recommend computed() over derive

mathpirate:
Both error messages previously recommended using `derive` with explicit parameters. Since `computed()` is the modern/ergonomic approach that handles closures automatically via the transformer, update the messages to recommend `computed()` instead.

- "Tried to directly access an opaque value" now suggests computed()
- "Cannot create cell link: space is required" now suggests computed()

---
## Summary by cubic
Updated runner error messages to recommend computed() instead of derive, aligning with the current reactive API. Clarifies closure behavior for opaque value access and cell link creation, guiding users to computed() for reactive computations.

<sup>Written for commit 7142dccd8e076003009ed7de0901fa3e3fb296c9. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/runner/src/cell.ts
+++ b/packages/runner/src/cell.ts
@@ -433,7 +433,7 @@ export class CellImpl<T> implements ICell<T>, IStreamable<T> {
     if (!space) {
       throw new Error(
         "Cannot create cell link: space is required.\n" +
-          "This can happen when manually calling .get() on closed-over cells.\n" +
+          "This can happen when accessing closed-over cells e.g. with .get().\n" +
           "Use `computed()` for reactive computations - it handles closures automatically.\n",
       );
     }


packages/runner/src/cell.ts
```diff
@@ -433,8 +433,8 @@ export class CellImpl<T> implements ICell<T>, IStreamable<T> {
     if (!space) {
       throw new Error(
         "Cannot create cell link: space is required.\n" +
-          "This can happen when closing over (opaque) cells in a lift or derive.\n" +
-          "Instead pass those cells into the lift or derive directly as Cell<> inputs.\n",
+          "This can happen when manually calling .get() on closed-over cells.\n" +
```
seefeldb: probably other accesses of cells as well -- but not sure this case is ever triggered anymore

PR #2299, opened 2025-12-19
CT-1126: Simplify cross-space favorites - rely on wish.ts lazy fallback

jkomoros:
## Summary

Fixes CT-1126: `wish({ query: "#tag" })` was failing to find cross-space favorites because the tag was empty when the favorite was stored (schema hadn't synced yet).

## The Problem

When favoriting a charm from a different space:
1. `addFavorite()` computes the tag via `getCellDescription(charm)`
2. For cross-space charms, schema data hasn't arrived yet → tag is empty
3. Later, `wish({ query: "#googleAuth" })` fails to find the favorite

## What We Tried (preserved in commit history)

The commit history shows our exploration:

1. **`await favoritesCell.sync()`** - Didn't help (wrong cell)
2. **Auto-sync in `asSchemaFromLinks()`** - Defensive, kept as belt-and-suspenders
3. **`await charm.sync()` + fallback** - Still had race conditions
4. **Reactive `sink()` pattern** - Broken: cancellation doesn't work, memory leaks
5. **Fix reactive to read from correct cell** - Fixed the bug but pattern still flawed

## Final Solution (simplest)

Remove the reactive complexity entirely. Rely on the **existing fallback in wish.ts** that computes tags lazily:

```typescript
// wish.ts already has this fallback:
if (!tag) {
  const { schema } = entry.cell.asSchemaFromLinks().getAsNormalizedFullLink();
  if (schema !== undefined) {
    tag = JSON.stringify(schema);
  }
}
```

**favorites.ts** now simply stores favorites with whatever tag is available (might be empty for cross-space). The docstring explicitly acknowledges this.

## What We're Unsure About

We want to be honest about limitations:

- **Timing-dependent**: If `wish()` is called immediately after `addFavorite()` and schema hasn't synced yet, the fallback may also return empty. We believe `asSchemaFromLinks()` auto-sync handles this, but it's implicit.

- **No test coverage**: There's no test for the specific scenario of cross-space favorites with initially empty tags. The existing tests pass, but they don't cover this edge case.

- **Framework gaps exposed**: This fix works around the lack of:
  - Proper cancellation semantics for `sink()`
  - A "wait for readiness" primitive for cross-space data

## Why We Think This Is Right

1. **Framework-idiomatic**: The fallback in wish.ts is the intended design (per code comments)
2. **Simple**: 30 lines removed, clear responsibility split
3. **No memory leaks**: Removed the broken reactive subscriptions
4. **Honest**: Documents that tags may be empty for cross-space charms

## Files Changed

- `packages/charm/src/favorites.ts` - Simplified, removed reactive pattern
- `packages/runner/src/builtins/wish.ts` - Kept existing fallback (no changes in final)
- `packages/runner/src/cell.ts` - Kept auto-sync in `asSchemaFromLinks()` (from earlier commit)

## Test Plan

- [x] Existing charm tests pass
- [x] Existing wish tests pass (28 tests)
- [ ] Manual test: Favorite a cross-space charm, then `wish({ query: "#schema" })`
- [ ] Consider adding explicit test for cross-space favorites

## Request for Review

We'd appreciate feedback on:
1. Is relying on the lazy fallback the right architectural choice?
2. Should we add explicit test coverage for cross-space favorites?
3. Are there framework primitives we should be using instead?

via [Happy](https://happy.engineering)

---
## Summary by cubic
Fixes CT-1126. Cross-space favorites are now reliably found: wish() computes missing tags at query time, addFavorite() syncs before tagging, and schema lookups auto-sync.

- **Bug Fixes**
  - wish.ts: If a favorite’s tag is empty, derive it lazily from the cell’s schema during both text and hashtag searches.
  - favorites.ts: addFavorite() now syncs the charm before computing the tag to avoid storing empty tags.
  - cell.ts: asSchemaFromLinks() auto-syncs when not yet synced.

<sup>Written for commit c78bdcbe0300a0b13f27006619ccd92d2a00f955. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/runner/src/builtins/wish.ts
+++ b/packages/runner/src/builtins/wish.ts
@@ -276,8 +276,7 @@ function resolveBase(
             }
           }
 
-          const hashtags =
-            tag?.toLowerCase().matchAll(/#([a-z0-9-]+)/g) ?? [];
+          const hashtags = tag?.toLowerCase().matchAll(/#([a-z0-9-]+)/g) ?? [];
           return [...hashtags].some((m) => m[0] === searchTerm);
         });
 


seefeldb (commented):
looks good, but it would be better to fix the root cause and make sure we write the schemas. it's actually a little puzzling to me as the schema must clearly be loaded when favoriting a charm.

A simple `await resolvedCharm.sync()` before `const tag = getCellDescription(charm)` should work.

Let's keep the fallback code for now, we might actually want to use that primarily once schemas can change over time, then the schema in the index is just for searching purposes and should be regularly refreshed.

jkomoros:
Thanks for the review! Added `await resolvedCharm.sync()` before computing the tag as you suggested.

You're right that this is more direct - sync the charm, then compute the tag. The commit history shows our exploration of various approaches (reactive sink, etc.) but your suggestion is cleaner.

Kept the wish.ts fallback as you mentioned - makes sense for future schema changes where the stored tag is just for search indexing.

PR #2341, opened 2025-12-22
Gideon/superstition rationalization

mathpirate:
WIP/not yet ready for review. Still have to pare these additions back down and review them all critically once more

---
## Summary by cubic
Documented and verified common “superstitions” across Cells, handlers, components, and LLM usage, and added targeted test patterns to lock in correct behavior and workarounds. This improves developer guidance and prevents subtle runtime issues.

- New Features
  - Added test patterns:
    - $checked on computed lists (reads ok, writes ignored).
    - Handler cell access across contexts (closures fail in ifElse/ternary; params work).
    - Record.key().set() behavior on empty vs populated records.
    - fetchData inside .map() (dynamic instantiation works; verified).
    - Computed projection vs property access (writes allowed; verified).

- Bug Fixes
  - Handlers: Prefer action() inside the pattern body (closures handled automatically). In ifElse/ternary, inline closures fail—pass params via action()/handler. Closures work in .map() and computed(). Data attributes are accessible via event.target.dataset—prefer passing params for reliability.
  - Cells: Access is type-based—Cell<T> always needs .get(); OpaqueRef uses direct property access, and method calls go inside computed(). For checkboxes on computed/filtered lists, use a separate writable Cell. Record.key().set() fails on empty Records; use spread to add first keys. Default<T[],[]> array items are read-only—wrap in Cell<Default<...>> for $checked/$value binding.
  - LLM: Use vendor:model names (e.g., anthropic:claude-sonnet-4-5). For regeneration, set cache: false (don’t use fetch for LLM). Avoid prompts derived from agent output to prevent loops; see models via /api/ai/llm/models.
  - Favorites: Call wish() at the pattern level (not inside computed()); use computed() to conditionally use the result.
  - Components: Use ct-screen for full-height layouts; ct-code-editor mentions via [[ not @; ct-message-input has no buttonText in JSX; ct-image-input notes; SVG not supported; use camelCase props in ct-* components.
  - Local Dev: OAuth/API must target port 8000 (not 5173). Restart local dev server to see UI component changes. Use safe space names (no “/”) and isolated spaces for testing to avoid “space poisoning.”
  - Types: Cells must be JSON-serializable (avoid Map/Set). Access array length via computed(). Use boxed selection over Cell<Cell<T>>.
  - Reactivity: First pass may have empty async data; use computed() and distinguish wish.error vs empty result.
  - Storage: Avoid read-modify-write races on arrays; use items.push() and .key().set() to prevent StorageTransactionInconsistent.
  - Misc: Clarified charm URL format with space/charm ID.

<sup>Written for commit 2d5efd80ef1fdde34763e4fc27cac96d0f6e73f6. Summary will update on new commits.</sup>

### DIFF
--- a/docs/common/PATTERNS.md
+++ b/docs/common/PATTERNS.md
@@ -317,7 +317,7 @@ export default pattern<{ count: Writable<number> }>(({ count }) => {
     [UI]: (
       <div>
         <ct-button onClick={increment}>+1</ct-button>
-        <ct-button onClick={addAmount}>Add 10</ct-button>
+        <ct-button onClick={addAmount({ amount: 10 })}>Add 10</ct-button>
       </div>
     ),
     increment,  // Export for cross-charm access


docs/common/COMPONENTS.md
```diff
@@ -138,6 +137,8 @@ Input + button combo for adding items.
 />
 ```
 
+**Note:** The `buttonText` property exists in the component but is intentionally omitted from JSX types. Use `ct-input` + `ct-button` separately if you need custom button text.
```
bfollington: I don't think this is intentional at all.

docs/common/COMPONENTS.md
```diff
@@ -239,6 +240,60 @@ const removeItem = handler<unknown, { items: Cell<Item[]>; item: OpaqueRef<Item>
...
+  {/* Content in DOM but invisible */}
+</div>
+
+// ✅ WORKS - full available width and height
+<ct-screen style="display: flex; flex-direction: column;">
+  <header>Title</header>
+  <div style="flex: 1; overflow: auto;">
+    {/* Scrollable content */}
+  </div>
+</ct-screen>
```
bfollington: The inline styles here are superfluous, `ct-screen` already styles `:host` with `display: flex; flex-direction: column;` internally so this is just noise.

docs/common/COMPONENTS.md
```diff
@@ -239,6 +240,60 @@ const removeItem = handler<unknown, { items: Cell<Item[]>; item: OpaqueRef<Item>
...
+// ❌ DOESN'T WORK - content appears blank
+<div style={{ height: "100%", display: "flex", flexDirection: "column" }}>
+  {/* Content in DOM but invisible */}
+</div>
+
+// ✅ WORKS - full available width and height
+<ct-screen style="display: flex; flex-direction: column;">
+  <header>Title</header>
+  <div style="flex: 1; overflow: auto;">
+    {/* Scrollable content */}
```
bfollington: "scrollable" content requires being wrapped in a `ct-vscroll` if you want `overflow: scroll` type behaviour

docs/common/COMPONENTS.md
```diff
@@ -254,3 +309,32 @@ const removeItem = handler<unknown, { items: Cell<Item[]>; item: OpaqueRef<Item>
...
+
+When authoring `ct-*` components, use camelCase for property names. Kebab-case JSX attributes don't map to camelCase Lit properties:
+
+```tsx
+// ❌ Kebab-case won't work
+<ct-autocomplete allow-custom={true} />  // Sets element["allow-custom"], not allowCustom
+
+// ✅ CamelCase works
+<ct-autocomplete allowCustom={true} />  // Sets element.allowCustom correctly
+```
```
bfollington: I think this should be higher up in the file / more loudly proclaimed.

docs/common/DEBUGGING.md
```diff
@@ -154,6 +186,55 @@ return { addItem };
...
+<button
+  style={{ background: "transparent", border: "none", cursor: "pointer" }}
+  onClick={handleSelect}
+>
+  Click to select
+</button>
+```
+
+**Why:** The framework JSX types don't include these events (marked as `@TODO`).
+**Benefit:** Click-based interactions are more mobile-friendly anyway.
```
bfollington: I had deliberately left hover events out due to the mobile support aspect, hover events can also create dark patterns (things jumping in front of other elements unexpectedly)

docs/common/LLM.md
```diff
@@ -191,6 +246,32 @@ if (!result.pending && !result.error) {
...
+
+const agentGoal = computed(() => {
+  const found = memberships.map(m => m.name);
+  return `Find items. Already saved: ${found.join(", ")}`;  // Changes when agent writes!
+});
+
+const agent = generateObject({
+  prompt: agentGoal,  // Prompt changes → agent restarts → writes more → LOOP
+  tools: { save: { handler: (args) => memberships.push(args) } },
+});
```
bfollington: If you want to break this loop, use `llmDialog` and the `addMessage` handler which allows you to control the feedback loop

docs/common/LLM.md
```diff
@@ -208,6 +289,40 @@ const result = generateText({ prompt });
 
 When adding items to an array, only new items trigger LLM requests.
 
+**Important:** Always use `generateText`/`generateObject` instead of raw `fetch()` to LLM endpoints. Raw fetch bypasses caching and may be restricted in future.
+
+### Cache Busting for Regeneration
+
+When implementing "respin" or "regenerate" features, add a nonce to bust the cache:
```
bfollington: Why not just set the `cache` parameter to `false` in the `llm` call?

docs/common/LOCAL_DEV_SERVERS.md
```diff
@@ -117,6 +119,47 @@ tail -f packages/shell/local-dev-shell.log
...
+```
+
+**Why this works:** The shell's esbuild config bundles all dependencies (including `@commontools/ui`). When any watched file changes, esbuild rebuilds the entire bundle with the latest UI component code.
+
+**Verify your changes are included:**
+```bash
+# Search for a unique string from your changes
+grep "your-unique-string" packages/shell/dist/scripts/index.js
+```
+
```
bfollington: Invert this guidance and it becomes one sentence: when editing `ct-` components, restart local dev server to ensure the updated code is running.

docs/common/PATTERNS.md
```diff
@@ -340,6 +355,23 @@ const active = computed(() => items.filter(i => !i.done));
...
+  return items.map(item => item.name);  // Crashes if any item is null
+});
+
+// ✅ Filter first to ensure valid items
+const names = computed(() => {
+  const safeItems = items.filter((item): item is Item => item != null);
+  return safeItems.map(item => item.name);
+});
+```
+
```
bfollington: This is extremely annoying and feels like a BUG to me, not something to put in the docs.

docs/common/CELLS_AND_REACTIVITY.md
```diff
@@ -764,6 +764,54 @@ const activeItems = computed(() => items.filter(item => !item.done));
...
+**Rule:** For checkboxes on computed/filtered lists, separate selection state into a writable Cell.
+
+**Check 5: Using .key().set() on empty Records?**
+
+`.key(k).set(v)` fails on empty `Record<string, T>` cells. Once a Record has at least one key, adding new keys works:
+
+```typescript
+// ❌ FAILS - "Value at path ... is not an object"
+const data: Cell<Record<string, Item>> = Cell.of({});  // Empty!
+data.key("newkey").set({ value: "test" });
```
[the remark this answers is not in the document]
seefeldb: +1, we should just create the object if it's missing

docs/common/COMPONENTS.md
mathpirate: removed from docs, thank you

docs/common/COMPONENTS.md
mathpirate: Fixed - updated example to use ct-vscroll for scrollable content.

docs/common/COMPONENTS.md
```diff
@@ -239,6 +240,60 @@ const removeItem = handler<unknown, { items: Cell<Item[]>; item: OpaqueRef<Item>
...
+```tsx
+<ct-image-input
+  onct-change={handleImageUpload}
+  maxSizeBytes={3932160}  // See note below
+>
+  📷 Add Photo
+</ct-image-input>
+```
+
+**Base64 overhead:** `maxSizeBytes` checks the original file size, but APIs receive base64-encoded data (~33% larger). For a 5MB API limit, use `maxSizeBytes={3932160}` (75% of limit).
```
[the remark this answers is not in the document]
mathpirate: Simplified - now just says 'The component compresses images to fit within maxSizeBytes.'

docs/common/DEBUGGING.md
```diff
@@ -51,6 +55,34 @@ These issues compile without errors but fail at runtime.
...
+**Error:** "ReadOnlyAddressError: Cannot write to read-only address: data:application/json..."
+
+```typescript
+// ❌ Default<> arrays have read-only item properties
+interface Input {
+  items: Default<Item[], []>;  // OpaqueRef - items are read-only!
+}
+
+{items.map((item) => (
+  <ct-checkbox $checked={item.done} />  // ReadOnlyAddressError on click
```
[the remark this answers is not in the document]
mathpirate: Removed - this is part of CT-1152

docs/common/LLM.md
```diff
@@ -83,6 +83,29 @@ If you get "schema is required" error, add to file top:
...
+model: "anthropic:claude-opus-4-1"
+model: "openai:gpt-4o"
+
+// ✅ ALSO CORRECT - aliases work
+model: "sonnet-4-5"
+model: "opus-4-1"
+```
+
+If you get `TypeError: Cannot read properties of undefined (reading 'model')`, check your model name format first.
+
```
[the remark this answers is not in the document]
mathpirate: Added note: 'Query /api/ai/llm/models to see all available models.'

docs/common/LLM.md
mathpirate: Added tip: 'Use llmDialog with addMessage handler to control the feedback loop explicitly.'

docs/common/LLM.md
mathpirate: Great catch ty. Replaced nonce pattern with simpler 'cache: false' option.

docs/common/TYPES_AND_SCHEMAS.md
```diff
@@ -36,6 +36,51 @@ With `Cell<T>` in your signature:
...
 Without `Cell<>`, you can still display values in JSX, pass to `computed()`, and map over arrays - all reactively. Note: filtering and transformations must be done in `computed()` outside JSX, then the result can be mapped inside JSX.
 
+### Cell Values Must Be JSON-Serializable
+
+Cells are persisted via JSON serialization. **Map and Set lose their methods** after serialization:
+
+```typescript
+// ❌ BROKEN - Map/Set methods lost after serialization
+const mySet = Cell.of<Set<string>>(new Set(["a", "b"]));
+const myMap = Cell.of<Map<string, number>>(new Map([["key", 1]]));
```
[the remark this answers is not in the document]
mathpirate: Filed as CT-1156: https://linear.app/common-tools/issue/CT-1156/map-and-set-in-cells-should-produce-type-error. Removed from docs.

packages/patterns/gideon-tests/test-checkbox-computed.tsx
```diff
@@ -0,0 +1,99 @@
...
+          Active: {activeCount} / Total: {totalCount}
+        </div>
+
+        {/* TEST A: Direct Cell map - should work */}
+        <div style={{ border: "2px solid green", padding: "1rem", marginBottom: "1rem", borderRadius: "8px" }}>
+          <h3 style={{ color: "green" }}>A: Direct Cell Map (should work)</h3>
+          <p style={{ fontSize: "0.8rem", color: "#666" }}>
+            $checked on items.map() - direct Cell access
+          </p>
+          {items.map((item) => (
```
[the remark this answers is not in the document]
mathpirate: This is not React :)

docs/common/PATTERNS.md
```diff
@@ -400,6 +465,18 @@ const ui = computed(() => (
...
+CommonTools uses Lit for rendering, not React. Key differences:
+
+```typescript
+// ❌ key prop causes compiler error - not needed
+{items.map(item => <div key={item.id}>{item.name}</div>)}
+
+// ✅ No key needed - Lit handles reconciliation differently
+{items.map(item => <div>{item.name}</div>)}
+```
+
```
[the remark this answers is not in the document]
mathpirate: Filed https://linear.app/common-tools/issue/CT-1160/add-compiler-error-when-key-prop-is-used-in-jsx-react-pattern-not to track this potential improvement; meanwhile removed this from the docs

PR #2350, opened 2025-12-23
feat(runner): Add LRU compilation cache to compileAndRun builtin

jkomoros:
## Summary

- Adds in-memory LRU cache for compiled recipes in the `compileAndRun` builtin
- Cache keyed by content hash (merkle-reference) of program files
- LRU eviction with 1000 entry limit to prevent unbounded memory growth on long-running servers (e.g., background-charm-service)
- Single-flight pattern to deduplicate concurrent compilations of the same code
- Performance instrumentation via `getCompilationStats()` and `clearCompilationCache()`

## Why

This enables efficient lazy-loading of patterns via `fetchProgram + compileAndRun`. Without caching:
- Every invocation recompiles (100-300ms)

With caching:
- First compilation: 100-300ms
- Subsequent uses of same pattern: <1ms (cache hit)

## Test plan

- [x] `deno check` passes
- [x] Reviewed by multiple Oracle agents for edge cases, memory safety, and correctness
- [ ] CI passes

---
## Summary by cubic
Add an in-memory LRU cache to compileAndRun to reuse compiled recipes by content hash. This speeds repeat runs from 100–300ms to <1ms and deduplicates concurrent compilations.

- **New Features**
  - LRU cache (max 1000 entries) keyed by merkle content hash to avoid recompiles and bound memory on long-running servers.
  - Single-flight deduplication so concurrent requests share one compilation.
  - Monitoring helpers exported: getCompilationStats() and clearCompilationCache().

<sup>Written for commit 1bd2a7a7e6a7472c173c85dcf44233d9cf76752f. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/runner/src/builtins/compile-and-run.ts
+++ b/packages/runner/src/builtins/compile-and-run.ts
@@ -25,7 +25,7 @@
  * // Internal debugging only - not part of public API
  * import { getCompilationStats } from "./builtins/index.ts";
  * console.log(getCompilationStats());
- * // { cacheHits: 45, cacheMisses: 3, cacheEvictions: 0, cacheSize: 3, hitRate: "93.8%" }
+ * // { cacheHits: 42, singleFlightHits: 3, cacheMisses: 5, cacheEvictions: 0, cacheSize: 5, hitRate: "90.0%" }
  * ```
  *
  * ## Usage with fetchProgram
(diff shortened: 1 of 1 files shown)


packages/runner/src/builtins/compile-and-run.ts
```diff
@@ -191,40 +326,80 @@ export function compileAndRun(
     // Capture requestId for this compilation run
     const thisRequestId = requestId;
 
-    const compilePromise = runtime.harness.run(program)
```
seefeldb: There is already a recipe-manager in runner that caches compiled ones per program hash. So maybe directly calling runtime.harness.run bypasses that?

The recipe manager code is a little messy. So it would be worth understanding the flows there and around runtime.harness.run and clean it up, making sure this path hits the cache.

Let's not add another caching layer here that only compileAndRun uses.

PR #2367, opened 2025-12-24
fix(runner): add cycle detection to recursiveStripAsCellAndStreamFromSchema

jkomoros:
## Summary

- Fixes CT-1141: `recursiveStripAsCellAndStreamFromSchema` was missing cycle detection, causing stack overflow when processing circular schemas (like Record pattern's embedded sub-charm schemas)
- Uses a `Map` (not `Set`) following the idiomatic pattern from `cell.ts:recursivelyAddIDIfNeeded`

## Details

The fix ensures:
1. No infinite recursion on circular schemas
2. Consistent results - all references to the same input get the same output
3. Cycles in output mirror cycles in input (with `asCell`/`asStream` flags properly stripped)

## Test plan

- [x] Added tests for direct self-reference cycles
- [x] Added tests for three-way cycles (A → B → C → A)
- [x] Added tests for cycles through array items and anyOf
- [x] Added tests for multiple independent cycles
- [x] Added tests for keepStreams option with circular schemas
- [x] Added tests for diamond pattern (shared references get consistent treatment)
- [x] All existing tests pass

---
## Summary by cubic
Added cycle detection to recursiveStripAsCellAndStreamFromSchema to prevent stack overflows on circular schemas and keep outputs consistent. Addresses Linear CT-1141.

- **Bug Fixes**
  - Track seen nodes with a Map and return processed objects to break cycles (pattern from cell.ts:recursivelyAddIDIfNeeded).
  - Add a depth limit (100) to guard against pathological nesting/proxy objects.
  - Strip asCell/asStream reliably across cyclic and shared refs; keepStreams option still respected.
  - Added tests for self-reference, three-way cycles, arrays, anyOf, multiple independent cycles, and diamond/shared references.

<sup>Written for commit 8f1e8925c767dfc5a97fe9284c7a5f2f23cd21b1. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/runner/src/link-utils.ts
+++ b/packages/runner/src/link-utils.ts
@@ -396,6 +396,8 @@ export function sanitizeSchemaForLinks(
 function recursiveStripAsCellAndStreamFromSchema(
   schema: any,
   options: { keepStreams?: boolean },
+  seen: Map<any, any> = new Map(),
+  depth: number = 0,
 ): any {
   // Handle null/undefined/boolean schemas
   if (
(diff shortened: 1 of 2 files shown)


seefeldb (approved):
This prevents crashes, which is good.

But for the original issue. I'm wondering how we even got here. The idiomatic way for JSON schema to handle cycles is using $ref and so we should fix upstream to use that.

Let's keep the original issue open with this added context or create a new issue.

jkomoros:
## CI Status Clarification

The Pattern Integration Tests failure is **pre-existing on main** - it's not caused by this PR. The `record-backup.tsx` pattern has been OOMing on main since at least the last 3 commits.

This PR addresses a different issue: the stack overflow in `recursiveStripAsCellAndStreamFromSchema` when deploying patterns with circular schemas via CLI.

The `record-backup.tsx` OOM is a separate issue that should be tracked separately - it appears to be caused by `JSON.stringify(Record)` creating an exponentially large structure.

@seefeldb Given your approval, should we merge this and track the record-backup OOM as a separate issue?

PR #2368, opened 2025-12-24
fix(patterns): prevent OOM in integration tests by cleaning up charms

jkomoros:
## Summary

The pattern integration tests were failing with out-of-memory errors since PR #2363 (PhotoModule). This wasn't caused by that specific PR - it just pushed the cumulative memory over the edge.

**Root cause:** Each charm created during testing was never cleaned up. They accumulated in memory until the test suite finished, causing the V8 heap to exceed ~4GB and crash.

## Changes

1. **Add per-charm cleanup** in `all.test.ts`: Call `cc.remove(charm.id)` after each test to clean up the charm immediately, preventing memory accumulation

2. **Increase V8 heap limit** in `deno.json`: Add `--v8-flags=--max-old-space-size=4096` as a safety net (4GB vs default ~512MB-1GB)

## Technical Details

- The test creates 46+ charms in a single `CharmsController` lifecycle
- Each charm includes compiled TypeScript, recipe metadata, and runtime state (~30-80MB for large patterns)
- Without cleanup: 46 patterns × ~50MB = ~2.3GB minimum, plus compilation caches
- The OOM occurred around 3.7GB, just before `record-backup.tsx` could complete

## Test plan

- [x] Type checking passes
- [ ] CI pattern integration tests pass (will verify with this PR)

---
## Summary by cubic
Prevents OOM crashes in patterns integration tests by using a fresh CharmsController per test and increasing the V8 heap size. This avoids recipe cache buildup across tests and stabilizes CI.

- **Bug Fixes**
  - Initialize and dispose a fresh CharmsController in each test to prevent RecipeManager cache accumulation.
  - Run integration tests with --v8-flags=--max-old-space-size=4096 as a safety net.

<sup>Written for commit 049f03113675da45d0ef171e57ff07254252280f. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/patterns/deno.json
+++ b/packages/patterns/deno.json
@@ -1,8 +1,8 @@
 {
   "name": "@commontools/patterns",
   "tasks": {
-    "integration": "LOG_LEVEL=warn deno test --trace-leaks -A ./integration/*.test.ts",
-    "integration:all": "TEST_HTTP=1 TEST_LLM=1 LOG_LEVEL=warn deno test --trace-leaks -A ./integration/*.test.ts",
+    "integration": "LOG_LEVEL=warn deno test --v8-flags=--max-old-space-size=4096 --trace-leaks -A ./integration/*.test.ts",
+    "integration:all": "TEST_HTTP=1 TEST_LLM=1 LOG_LEVEL=warn deno test --v8-flags=--max-old-space-size=4096 --trace-leaks -A ./integration/*.test.ts",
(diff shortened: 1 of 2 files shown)


jkomoros:
Superseded by #2369 which reverts the PhotoModule commit that caused the OOM. The memory cleanup fixes in this PR are still valuable but will be re-submitted after the revert lands.

PR #2412, opened 2026-01-02
feat(runner): lazy start() with deferred sync and  watching

seefeldb:
## Summary

This PR refactors `Runner.start()` to be the single entry point for starting charms, with lazy sync behavior and automatic recipe change detection.

### Key Changes

- **`start()` returns `Promise<boolean>`** - Enables callers to await startup completion and handle errors
- **Lazy sync** - `start()` automatically syncs empty cells before starting, removing boilerplate from callers
- **`$TYPE` sink watching** - Running charms now watch their `$TYPE` cell and automatically restart when the recipe changes
- **Handles any cell** - `start()` gracefully handles subpath cells, data-only cells, and cells without recipes (no-op)

### Simplifications Enabled

1. **CharmManager** (`packages/charm/src/manager.ts`)
   - `get()` now just calls `runtime.start()` instead of manual recipe loading
   - Uses `processCell.resultRef` for schema instead of computing it
   - Removed `#getResultSchema()` helper

2. **Shell Runtime** (`packages/shell/src/lib/runtime.ts`)
   - `getPattern()` returns synchronously with a `ready` promise
   - Removed `PatternCache` class entirely - no longer needed

3. **ct-render** (`packages/ui/src/v2/components/ct-render/ct-render.ts`)
   - Removed `_loadAndRenderRecipe()` method
   - Just calls `runtime.start(cell)` and renders - start() handles all cases

4. **Shell Views** (`packages/shell/src/views/`)
   - Navigation is now synchronous - `_selectedPattern` Task returns immediately
   - Errors from `start()` are captured and displayed in `BodyView`

### Test Changes

- Added comprehensive tests for lazy start behavior in `packages/runner/test/runner.test.ts`
- Added warmup test to `ct-code-editor` integration tests to stabilize Cell<->editor sync

## Test plan

- [x] Unit tests pass (`deno task test` in packages/runner)
- [x] Integration tests pass (`deno task integration` in packages/patterns)
- [x] Manual testing of navigation in shell

---
## Summary by cubic
Runner.start() is now the single entry point for starting charms with lazy sync and recipe-change watching. This simplifies shell navigation and UI rendering, and makes startup more reliable.

- **New Features**
  - start() returns Promise<boolean> so callers can await or handle errors.
  - Lazily syncs empty result cells before running; runs synchronously when data is present.
  - Watches $TYPE and auto-restarts when the recipe changes.
  - Safe to call on any cell (subpaths, data-only, no recipe); no-op when nothing to run.

- **Refactors**
  - CharmManager get()/getResult() use processCell.resultRef; remove result schema helper.
  - CharmManager getArgument is now async and loads the recipe to apply argumentSchema; readingFrom/readBy are async.
  - CharmController getRecipeMeta loads recipe before metadata to ensure it’s populated.
  - Charm IO: input/result get() and getCell() return Promises; CLI and tests await these reads.
  - Shell Runtime getPattern() returns { controller, ready } and starts charms via runtime.start(); remove PatternCache.
  - Navigation is synchronous; start() errors are captured in AppView and shown in BodyView.
  - ct-render just calls runtime.start(cell); remove manual recipe loading.
  - setup() no longer calls stop() on recipe changes; the $TYPE sink handles stop/start.
  - Tests added for lazy start, idempotency, sync execution, recipe restart, plus a ct-code-editor warmup.

<sup>Written for commit 6a80c39468992f516fa1af36508073b323613fd4. Summary will update on new commits.</sup>

### DIFF
--- a/packages/patterns/integration/ct-code-editor.test.ts
+++ b/packages/patterns/integration/ct-code-editor.test.ts
@@ -1401,7 +1401,7 @@ async function resetEditorState(
   await waitFor(
     async () => {
       const editorContent = await getEditorContent(page);
-      const cellValue = charmController.result.get(["content"]);
+      const cellValue = await charmController.result.get(["content"]);
       return editorContent === "" && cellValue === "";
     },
     { timeout: 2000, delay: 50 },
--- a/packages/shell/integration/iframe-counter-charm.disabled_test.ts
+++ b/packages/shell/integration/iframe-counter-charm.disabled_test.ts
@@ -63,7 +63,7 @@ async function getCharmResult(page: Page): Promise<{ count: number }> {
   }
 
   // Use the element handle to evaluate in its context
-  return await appView.evaluate((element: XAppView) => {
+  return await appView.evaluate(async (element: XAppView) => {
     // Access the private _patterns property
     const activeCharmTask = element._patterns;
 
@@ -82,7 +82,7 @@ async function getCharmResult(page: Page): Promise<{ count: number }> {
     }
 
     // Get the result from the charm controller
-    const result = charmController.result.get();
+    const result = await charmController.result.get();
 
     return result as { count: number };
   });


seefeldb:
@jsantell this PR should move a lot of the loading complexity into runner. basically you just have to call runtime.runner.start(cell) and it'll take care of loading the underlying pattern, and so on. crucially, it doesn't wait for sync if all the data is already present, which is often the case when navigating within a session.

the promise returned from start either resolves as `true` or rejects with an error. that's either instantly or after (attempting to) load all the missing data.

not sure yet how it meshes with your pending changes, but there could now be even less dependency on charmmanager from the shell.

jsantell (commented):
The runtime changes look good, should make things faster.  CharmManager is not used in worker changes. Can the shell changes be applied on top of the worker runtime? Otherwise all shell changes would no longer be relevant

PR #2422, opened 2026-01-05
fix: prevent shell disconnect during cold start with timeout+retry

willkelly:
## Summary

Fixes the issue where the shell shows "disconnected" during clusterduck cold-start and never connects unless the user refreshes.

**Root cause**: Cross-space cell syncs could hang forever if the target space was unavailable or slow to respond (e.g., during cold start). The browser's `storageManager.synced()` would wait indefinitely for these cross-space promises.

**Changes**:

1. **Health check fix** (`packages/toolshed/routes/health/health.handlers.ts`):
   - Added memory readiness verification to `/_health` endpoint
   - Ensures Kubernetes doesn't route traffic until memory provider is ready

2. **Shell sync timeout** (`packages/shell/src/lib/runtime.ts`):
   - Added 10s timeout to `charmManager.synced()` with storage cleanup on timeout
   - Prevents infinite hang on initial sync

3. **Shell retry logic** (`packages/shell/src/views/RootView.ts`):
   - Added retry logic (up to 3 attempts) for `RuntimeInternals.create()` on timeout
   - Each retry creates entirely new Runtime with fresh websocket

4. **Cross-space sync timeout** (`packages/runner/src/link-resolution.ts`):
   - Added `syncCrossSpaceWithRetry()` wrapper with 5s timeout per attempt
   - Retries up to 10 times with 2s delay between attempts
   - This is the root cause fix - prevents hanging forever on cross-space fetches

## Architecture note

Cross-space fetches already happen correctly - directly from the initiating client (browser) to the target space. Each space gets its own websocket connection via `StorageManager.open(space)`. The issue was just missing timeout/retry handling.

## Test plan
- [ ] Deploy to staging
- [ ] Access shell during cold start (scale down workers, then scale back up)
- [ ] Verify shell eventually connects (may take a few retries)
- [ ] Check logs for cross-space sync retry messages

### DIFF
--- a/packages/runner/src/storage/cache.ts
+++ b/packages/runner/src/storage/cache.ts
@@ -1427,6 +1427,10 @@ export class Replica {
     this.selectorTracker = new SelectorTracker();
     // Clear the pull queue
     this.queue = new PullQueue();
+    // Clear in-flight commit promises - they will never resolve after reconnection
+    // because the transaction response handlers are lost with the old websocket
+    console.log(`[Replica] reset() clearing ${this.commitPromises.size} orphaned commit promises`);
+    this.commitPromises.clear();
 
     this.subscription.next({
       type: "reset",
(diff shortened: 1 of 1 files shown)


packages/toolshed/routes/health/health.handlers.ts
```diff
@@ -4,11 +4,20 @@ import { z } from "zod";
 import type { AppRouteHandler } from "@/lib/types.ts";
 import type { IndexRoute, LLMRoute } from "./health.routes.ts";
 import { checkLLMHealth } from "./llm-health.service.ts";
+// Import memory to ensure it's initialized before health check returns OK
```
jsantell: I don't think we need comments describing TLA

PR #2435, opened 2026-01-07
Feature/pattern testing system

mathpirate:
WIP DRAFT

High-level early reviews/feedback welcome, details not ready for refinement yet

---
## Summary by cubic
Adds a pattern-native testing system with a ct test CLI that runs .test.tsx patterns in an emulated runtime. Also fixes type inference and Stream handling (preserves Stream<T> in StripCell and improves AnyBrandedCell) so .send() works across patterns and nested types no longer infer as unknown. Replaces the v1 harness, adds a spec and example tests, and adds ct dev --test to run tests from a __tests__/ subdirectory.

- **New Features**
  - ct test <file|dir> runs pattern tests with --timeout and --verbose, and prints a pass/fail summary.
  - Runner processes a tests array in order (Streams as actions, Cell<boolean> as assertions) with a fresh in-memory runtime per file.
  - Docs and example added (spec in docs/specs/PATTERN_TESTING_SPEC.md, computed vs derive test in gideon-tests).
  - ct dev --test runs tests in a __tests__/ subdirectory using Deno’s test runner.

- **Migration**
  - Write tests as patterns that return { tests: [...] } with Stream<void> actions and Cell<boolean> assertions.
  - Name files *.test.tsx and run with ct test <path>.
  - Remove any use of @commontools/pattern-testing.

<sup>Written for commit 471368985a09ddc23b606f73f1a6ae09da603b72. Summary will update on new commits.</sup>


packages/cli/lib/test-runner.ts
```diff
@@ -0,0 +1,314 @@
...
+
+export interface TestRunnerOptions {
+  timeout?: number;
+  verbose?: boolean;
+}
+
+/**
+ * Creates a timeout promise that rejects after the specified duration.
+ */
+function createTimeout(ms: number, message: string): Promise<never> {
```
jsantell: We use deadlines in a few places, wonder if we should have a 'reject' version of `@commontools/utils/sleep` e.g. `timeout(5000, "Timed out")`

packages/cli/commands/test.ts
```diff
@@ -0,0 +1,78 @@
+import { Command } from "@cliffy/command";
+import { join } from "@std/path";
+import { discoverTestFiles, runTests } from "../lib/test-runner.ts";
+
+export const test = new Command()
+  .name("test")
+  .description("Run pattern tests (.test.tsx files).")
+  .example(
```
jsantell: Instead of handling files and directories and checking e.g. `*.test.tsx`, we could defer to the shell and just allow a glob (like `deno test *.test.ts`), which should transparently handle a single file, a directory of files, or some subset `ct test *_handler.test.tsx`

deno.json
```diff
@@ -31,7 +31,7 @@
   ],
   "tasks": {
     "check": "./tasks/check.sh",
-    "ct": "ROOT=$(pwd) && cd $INIT_CWD && deno run --allow-net --allow-ffi --allow-read --allow-write --allow-env \"$ROOT/packages/cli/mod.ts\"",
+    "ct": "ROOT=$(pwd) && cd $INIT_CWD && deno run --allow-net --allow-ffi --allow-read --allow-write --allow-env --allow-run \"$ROOT/packages/cli/mod.ts\"",
```
jsantell: I don't think this recursively executes deno, or does anything dev/charm doesn't do, do we need `--allow-run`?

packages/cli/commands/dev.ts
```diff
@@ -37,10 +41,26 @@ export const dev = new Command()
     "--main-export <export:string>",
     'Named export from entry for recipe definition. Defaults to "default".',
   )
+  .option(
```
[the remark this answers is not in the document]
mathpirate: oops thanks

packages/cli/commands/test.ts
mathpirate: done, i think

deno.json
mathpirate: not anymore!

mathpirate:
Superseded by https://github.com/commontoolsinc/labs/pull/2457

PR #2440, opened 2026-01-07
feat(cell): support multiple keys in Cell.key()

seefeldb:
## Summary

- Add support for multi-key navigation: `cell.key("user", "profile", "name")` equivalent to chaining `.key("user").key("profile").key("name")`
- Use method overloads instead of recursive types to avoid "excessive stack depth" errors with complex nested types
- Support type inference for up to 10 keys with fallback to `Cell<any>` for deeper paths
- Performance benchmarks show negligible impact on type checking time (~0.12s unchanged)

## Test plan

- [x] Type-level tests in `packages/runner/test/focus-types.test.ts` verify compile-time type inference
- [x] `deno task check` passes without type errors
- [x] `deno lint` passes
- [x] Existing cell tests pass (`packages/runner/test/cell.test.ts`)
- [x] Performance benchmarks show no regression

---
## Summary by cubic
Added multi-key navigation to Cell.key() so you can jump to nested properties in one call (e.g., cell.key("user", "profile", "name")). Results preserve nested Cell/Stream types; type inference works up to 10 keys with fallback to any for deeper paths and no measurable type-checking perf impact.

- **New Features**
  - key() now accepts 0–10 variadic keys; equivalent to chaining and backward compatible with single-key calls.
  - Supported for opaque cells; unknown keys or >10 keys return OpaqueCell<any>. Nested Cell/Stream values are returned as-is.

- **Refactors**
  - Switched from recursive types to method overloads to avoid “excessive stack depth” on complex types.
  - Updated KeyResultType to handle arrays of keys and added compile-time type tests; checks and lint pass.

<sup>Written for commit 8ff67a0177f05236a39a95b1baadbc1acabe3009. Summary will update on new commits.</sup>

### DIFF
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -353,6 +353,7 @@ export type WrapOrPreserve<T, Wrap extends HKT> = [T] extends [Cell<any>] ? T
   : [T] extends [ComparableCell<any>] ? T
   : [T] extends [ReadonlyCell<any>] ? T
   : [T] extends [WriteonlyCell<any>] ? T
+  : [T] extends [OpaqueCell<any>] ? T
   : Apply<Wrap, T>;
 
 /**


ubik2:
lgtm. nice feature.

bfollington (approved):
LGTM, but just to float it... would `Cell.path("a", "b", "c")` also make sense? Just thinking about the fact that the underlying data structure calls it `path` inside a `Link` etc.

No strong opinion.

seefeldb:
(merging now, since the rename would touch a lot of files)

PR #2447, opened 2026-01-08
fix(tests): improve flaky integration test reliability

willkelly:
## Summary

Addresses intermittent test failures identified in [flaky test analysis](https://gist.github.com/willkelly/4ba6eca84989ab8b938972bd3e1858ab).

**Verified with 4+ consecutive successful CI runs after fixing.**

### Changes

**Shell charm test (`charm.test.ts`):**
- Replace fixed `sleep()` delays with `waitFor()` polling pattern
- Await `click()` to catch "unstable box model" errors and retry
- Increase timeouts to 30s to handle retries reliably
- Wait for final expected value instead of fixed delays

**Nested counter test (`nested-counter.test.ts`):**
- Switch from `evaluate()` to `innerText()` for more stable element access
- Increase `waitForSelector` timeout from 500ms to 2000ms
- Better handles stale element handles during page re-renders

**Pattern harness (`pattern-harness.ts`):**
- Add retry logic for assertions (up to 10 retries with 50ms backoff)
- Properly cancel sink before dispose to stop reactive updates
- Add `idle()` and small delay before dispose to prevent timer leaks

### Root Causes Addressed

1. **Race conditions**: State hadn't propagated before assertions
2. **Stale element handles**: DOM elements detached during re-renders
3. **Unhandled promise rejections**: `click()` errors not caught
4. **Timer leaks**: Debounce timers not cleaned up before test exit

## Test plan

- [x] CI verified: 4+ consecutive successful runs
- [x] `counter-search-term-filter.test.ts` - 100/100 passes locally
- [x] `procurement-request.test.ts` - 50/50 passes locally (no timer leaks)

### DIFF
--- a/packages/generated-patterns/integration/pattern-harness.ts
+++ b/packages/generated-patterns/integration/pattern-harness.ts
@@ -1,36 +1,12 @@
 import { expect } from "@std/expect";
 import "@commontools/utils/equal-ignoring-symbols";
+import { waitFor } from "@commontools/integration";
 import { fromFileUrl } from "@std/path";
 import { FileSystemProgramResolver } from "@commontools/js-compiler";
 import { Identity } from "@commontools/identity";
 import { StorageManager } from "../../runner/src/storage/cache.deno.ts";
 import { Runtime } from "@commontools/runner";
 import { sleep } from "@commontools/utils/sleep";
-
-// Check if two values are deeply equal, ignoring symbols
-function deepEqual(a: unknown, b: unknown): boolean {
-  if (a === b) return true;
-  if (typeof a !== typeof b) return false;
-  if (a === null || b === null) return a === b;
-  if (typeof a !== "object") return false;
-
-  if (Array.isArray(a) && Array.isArray(b)) {
-    if (a.length !== b.length) return false;
-    return a.every((val, i) => deepEqual(val, b[i]));
-  }
-  if (Array.isArray(a) || Array.isArray(b)) return false;
-
-  const aKeys = Object.keys(a as object);
-  const bKeys = Object.keys(b as object);
-  if (aKeys.length !== bKeys.length) return false;
-
-  return aKeys.every((key) =>
-    deepEqual(
-      (a as Record<string, unknown>)[key],
-      (b as Record<string, unknown>)[key],
-    )
-  );
-}
 
 export interface EventSpec {
   stream: string;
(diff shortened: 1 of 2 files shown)


packages/patterns/integration/nested-counter.test.ts
```diff
@@ -63,16 +63,15 @@ describe("nested counter integration test", () => {
...
+    // Use try/catch because element may become stale between waitForSelector and innerText
+    // Use innerText() instead of evaluate() as it's more robust against stale handles
     await waitFor(async () => {
       try {
         const counterResult = await page.waitForSelector("#counter-result", {
           strategy: "pierce",
-          timeout: 500,
+          timeout: 2000,
         });
-        const initialText = await counterResult.evaluate((el: HTMLElement) =>
```
jsantell: If `counterResult` element leaves the page (could happen during some re-render), I imagine both `evaluate` and `innerText` would fail, but `innerText()` is more succinct

packages/patterns/integration/nested-counter.test.ts
```diff
@@ -63,16 +63,15 @@ describe("nested counter integration test", () => {
...
 
-    // Use try/catch because element may become stale between waitForSelector and evaluate
+    // Use try/catch because element may become stale between waitForSelector and innerText
+    // Use innerText() instead of evaluate() as it's more robust against stale handles
     await waitFor(async () => {
       try {
         const counterResult = await page.waitForSelector("#counter-result", {
           strategy: "pierce",
-          timeout: 500,
+          timeout: 2000,
```
jsantell: In practice, I think the main value of setting individual timeouts per command is for bailing out early on failure (rather than waiting ~1 minute per query), but that's already a failed case, and generally think we're best served just relying on defaults -- do we want to fail if this takes 3 seconds?

packages/generated-patterns/integration/pattern-harness.ts
```diff
@@ -88,8 +114,8 @@ export async function runPatternScenario(scenario: PatternIntegrationScenario) {
   const result = runtime.run(tx, patternFactory, argument, resultCell);
   tx.commit();
 
-  // Sink to keep the result reactive
-  result.sink(() => {});
+  // Sink to keep the result reactive, track cancel function for cleanup
+  const cancelSink = result.sink(() => {});
```
jsantell: Good catch

PR #2455, opened 2026-01-08
feat(runner): Add toJSONValue for proper Error serialization in cells

danfuzz:
## Summary

- Adds `toJSONValue()` function that converts arbitrary values to JSON-serializable form, with special handling for `Error` objects
- Adds `setAsJSONValue()` method on cells for storing values that need JSON conversion
- Fixes `fetchData` builtin to properly serialize errors using the new method

## Problem

When `fetchData` encountered an HTTP error, it stored the `Error` object directly via `.set(err)`. However, `Error` objects have non-enumerable properties (`message`, `stack`, `name`), so they were serialized as empty objects `{}` and the error information was lost.

## Solution

The new `toJSONValue()` function converts `Error` objects to a distinctive wrapper format:
```json
{
  "@Error": {
    "name": "Error",
    "message": "HTTP 404: Not Found",
    "stack": "..."
  }
}
```

This preserves all error information in a human-readable, JSON-compatible form using the `@` prefix convention already established in `llm-dialog.ts`.

## Test plan

- [x] Unit tests for `toJSONValue()` covering primitives, objects, arrays, Error conversion, nested errors, and circular reference detection
- [x] Integration test in `fetch-data-mutex.test.ts` verifying errors are stored with `@Error` wrapper
- [x] All existing unit tests pass

---
## Summary by cubic
Adds toJSONValue to convert arbitrary values—especially Error objects—into JSON-safe data and adds setAsJSONValue to cells. Updates fetchData to store errors with full details instead of {}.

- **New Features**
  - toJSONValue converts Error to {"@Error": {name, message, stack, ...}} and handles nested values.
  - Cell.setAsJSONValue stores a JSON-serializable clone of the input.
  - Cell.setRaw now uses toJSONValue for consistent conversion.
  - Unit tests cover primitives, arrays, objects, Error cases, and circular detection.

- **Bug Fixes**
  - fetchData now uses setAsJSONValue for errors so HTTP failures keep name/message/stack and custom fields.

<sup>Written for commit 940d3afd3d2cc671c360b53a6000c74b0304b087. Summary will update on new commits.</sup>

### DIFF
--- a/packages/runner/src/cell.ts
+++ b/packages/runner/src/cell.ts
@@ -1973,8 +1973,11 @@ export function toJSONValue(
     if (Error.isError(value)) {
       const error = value as Error;
       // This returns a single-key object whose key in a form unlikely to be
-      // mistaken as being for other purposes, though to be clear this form
-      // isn't (yet?) used anywhere else in the system.
+      // mistaken as being for other purposes. Note: Using `@` as the sigil for
+      // a special object key is used elsewhere in the system, though the
+      // specific pattern of using a single-property object with an `@key` to
+      // encode (the state of) an instance of a class is not (yet?) an
+      // established pattern in the system.
       return {
         "@Error": {
           ...error,


packages/runner/src/cell.ts
```diff
@@ -713,6 +715,26 @@ export class CellImpl<T> implements ICell<T>, IStreamable<T> {
...
+   * @param newValue - The value whose sanitized clone is to be stored in the
+   *     cell.
+   * @param onCommit - Same as with {@link #set}.
+   * @returns `this`, always.
+   */
+  setAsJSONValue(
+    newValue: AnyCellWrapping<T> | T,
+    onCommit?: (tx: IExtendedStorageTransaction) => void,
+  ): Cell<JSONValue> {
+    const origValue: any = isCell(newValue) ? newValue.value : newValue;
```
seefeldb: use .get() instead of .value here (.value is for debugging, being a getter i can get the value by clicking in the debugger vs calling .get())

danfuzz:
@seefeldb I was wondering that myself. I erred in this PR on the side of not touching `.set()` itself because of how central it is to the system. This is both because of my (lack of) understanding of the system (didn't seem _necessarily_ safe to me), and because of efficiency concerns. On the latter: The JSON string round trip is almost certainly okay in limited cases, but I didn't expect  we'd want to do it on every `.set()`.

danfuzz:
This work evolved, mooting this PR.

PR #2478, opened 2026-01-12
Extract `allCharms` and `favorites` to user-space

bfollington:
Define and export from the `default-app.tsx` and `home.tsx` instead of baking them in

---
## Summary by cubic
Moved allCharms/recent and favorites/journal into user-owned patterns. Re-enabled the Favorite button, auto-track recent on navigation, and add navigated charms to allCharms.

- **Refactors**
  - default-app: exports allCharms, recentCharms plus handlers addCharm, trackRecent.
  - home: exports favorites, journal plus handlers addFavorite, removeFavorite, addJournalEntry.
  - CharmManager: removed pinned/favorites APIs; getCharms() now async and reads from defaultPattern; add() sends to defaultPattern’s addCharm; recent tracking handled via pattern.
  - wish: resolves #allCharms and #recent via space.defaultPattern; #favorites and #journal via home.defaultPattern.
  - runtime: removed homeSpaceCellSchema; space cells now only link defaultPattern; user data lives under defaultPattern.
  - runtime-client: adds FavoritesManager using cell primitives; new IPC EnsureHomePatternRunning/GetHomeSpaceCell; FavoriteButton subscribes to favorites; navigateTo() calls trackRecent.
  - shell: ensures navigated charms are added to allCharms via default pattern’s addCharm.
  - Tests updated to link defaultPattern and read lists via it.

- **Migration**
  - Ensure a default pattern exists in each space before adding/listing charms, and in the home space before using favorites/journal.
  - Use RuntimeClient favorites methods in UI; in patterns, keep using wish tags (#allCharms, #recent, #favorites, #journal).
  - Charm creation no longer adds to lists; call default pattern’s addCharm to include a charm in allCharms.
  - Do not add the default pattern itself to allCharms (avoids circular refs).

<sup>Written for commit c2c78905e0fdfc549d0f3910463975e07803096d. Summary will update on new commits.</sup>

### DIFF
--- a/packages/home-schemas/home.ts
+++ b/packages/home-schemas/home.ts
@@ -1,15 +1,16 @@
-import { favoriteListSchema, favoriteEntrySchema } from './favorites.ts'
-import { journalSchema } from './journal.ts'
-import type { JSONSchema, Schema, Stream } from "@commontools/api";
+import { favoriteEntrySchema, favoriteListSchema } from "./favorites.ts";
+import { journalEntrySchema, journalSchema } from "./journal.ts";
+import type { JSONSchema, Schema } from "@commontools/api";
 
 export const homeSchema = {
-  type: 'object',
+  type: "object",
   properties: {
     favorites: favoriteListSchema,
     journal: journalSchema,
     addFavorite: { ...favoriteEntrySchema, asStream: true },
     removeFavorite: { ...favoriteEntrySchema, asStream: true },
-  }
+    addJournalEntry: { ...journalEntrySchema, asStream: true },
+  },
 } as const satisfies JSONSchema;
 
 export type Home = Schema<typeof homeSchema>;
--- a/packages/home-schemas/mod.ts
+++ b/packages/home-schemas/mod.ts
@@ -28,4 +28,4 @@ export {
   journalSnapshotSchema,
 } from "./journal.ts";
 
-export { homeSchema, type Home } from './home.ts'
+export { type Home, homeSchema } from "./home.ts";


packages/runner/src/runtime.ts
```diff
@@ -83,59 +82,29 @@ export interface RuntimeOptions {
...
-    timestamp: number;
-    eventType: string;
-    subject?: Cell<unknown>;
-    snapshot?: { name?: string; schemaTag?: string; valueExcerpt?: string };
-    narrative?: string;
-    tags?: string[];
-    space: string;
-  }[]>;
+  // No additional properties - all data is now in defaultPattern
 }
```
bfollington: Given this is empty we can remove it altogether?

packages/runtime-client/favorites-manager.ts
```diff
@@ -0,0 +1,482 @@
...
+          charmId = cellId.startsWith("of:") ? cellId.slice(3) : cellId;
+          console.log("[FavoritesManager] favItem IS the cell, id:", charmId);
+        }
+      } else if (favItem && typeof favItem === "object") {
+        // favItem is already an object
+        const fav = favItem;
+
+        if (isCellHandle(fav.cell)) {
+          const cellId = fav.cell.id();
+          charmId = cellId.startsWith("of:") ? cellId.slice(3) : cellId;
```
jsantell: Ditto

packages/patterns/system/home.tsx
```diff
@@ -1,11 +1,70 @@
...
+const addFavorite = handler<
+  { charm: Writable<{ [NAME]?: string }>; tag?: string },
+  { favorites: Writable<Favorite[]> }
+>(({ charm, tag }, { favorites }) => {
+  const current = favorites.get();
+  const resolvedCharm = charm;
+  if (!current.some((f) => f.cell.equals(resolvedCharm))) {
+    favorites.set([
+      ...current,
+      { cell: charm, tag: tag || "", userTags: Writable.of([]) },
```
bfollington: I suspect `tag` is not making its way here since it uses the descriptions stored in the schema which afaik we cannot grab here? Need to double check what actually ends up stored + get @seefeldb to weigh in.

packages/runtime-client/favorites-manager.ts
```diff
@@ -0,0 +1,286 @@
...
+
+  /**
+   * Subscribe to favorites changes.
+   * Callback is called immediately with current favorites (may be empty if not ready),
+   * and again whenever favorites change.
+   * @param callback - Function called with current favorites array
+   * @param onError - Optional callback for errors during subscription or transform
+   * @returns Unsubscribe function
+   */
+  subscribeFavorites(
```
bfollington: This seems overly lengthy / involved for my liking. Unsure how much of this ceremony is really needed.

PR #2487, opened 2026-01-12
chore: disable jsx-key lint rule

seefeldb:
## Summary
- Disable the `jsx-key` Deno lint rule globally since our JSX system (`@commontools/html`) doesn't use React-style VDOM reconciliation
- Remove manual lint suppression that's no longer needed

## Test plan
- [x] `deno lint` passes without jsx-key warnings

---
## Summary by cubic
Disable the jsx-key Deno lint rule globally since @commontools/html doesn’t use React-style reconciliation, and remove the now-unneeded inline suppression in the deprecated pattern file.

<sup>Written for commit 3266b4bee14d64623f735e01a672168feffbb800. Summary will update on new commits.</sup>

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -78,6 +78,7 @@
         "no-external-import"
       ],
       "exclude": [
+        "jsx-key",
         "jsx-no-useless-fragment",
         "no-explicit-any",
         "require-yield"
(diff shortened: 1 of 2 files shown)


jsantell (approved):
I think it's worth exploring in the future if using `key` for reconciliation is an improvement (to avoid JSON.stringify comparisons), maybe once everything is not a cell, but we're not there yet so LGTM

PR #2494, opened 2026-01-13
Suppress button variant `background-color` animation

bfollington:
- Replace transition: all with explicit properties (excluding background-color) in ct-button to prevent background color transitions during initial render
- Remove background-color from ct-chip transitions to prevent flash when theme variables are applied after initial render

This is causing the blue 'flash' when UIs first render post @jsantell's worker changes landing.

---
## Summary by cubic
Prevents the initial blue flash on buttons and chips by removing background-color animations that ran when theme variables applied after first render. CTButton now uses explicit transitions (opacity, transform, box-shadow, border-color) instead of transition: all, and CTChip drops background-color from its transition list.

<sup>Written for commit 0e670ec8a593bfd0de70076b1ff1376aa1e93d8e. Summary will update on new commits.</sup>

### DIFF
--- a/packages/ui/src/v2/components/ct-button/ct-button.ts
+++ b/packages/ui/src/v2/components/ct-button/ct-button.ts
@@ -66,7 +66,11 @@ export class CTButton extends BaseElement {
         font-weight: 500;
         font-family: var(--ct-theme-font-family, inherit);
         line-height: 1.25rem;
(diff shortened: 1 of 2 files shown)


tonyespinoza1:
https://discord.com/channels/1232452732963655720/1237131158773432411/1460479610314559650

PR #2503, opened 2026-01-13
fix(patterns): reliable parent notebook detection

tonyespinoza1:
## Summary

- Fix parent notebook navigation (⬆️ arrow) not showing reliably in nested notebooks
- Add `getNotebookNotesArray()` helper to handle Cell/Writable arrays from `wish("/")` proxy
- Filter notebooks by structure (`title && !content`) instead of `[NAME]` symbol which doesn't work through proxy
- Combine all notebook relationship detection (parents, children, siblings) into a single `notebookRelationships` computed block

## Root Cause

Nested `computed()` access doesn't track dependencies correctly in the pattern framework. For example:

```typescript
// BROKEN - nested computed access causes stale values
const parentNotebooks = computed(() => notebooks.filter(...));
const hasParentNotebooks = computed(() => parentNotebooks.length > 0);
```

The fix is to compute all related values in ONE computed block:

```typescript
// FIXED - single computed returns all needed values
const notebookRelationships = computed(() => {
  // ... compute parents, children, siblings in same block ...
  return { parents, hasParents, children, hasChildren, siblings, hasSiblings };
});
```

## Test plan

- [ ] Create a parent notebook ("outside")
- [ ] Create a nested notebook inside it ("inside")
- [ ] Navigate into "inside"
- [ ] Verify the ⬆️ parent link appears at the top
- [ ] Click the parent link to navigate back to "outside"

---
## Summary by cubic
Fixes unreliable parent notebook navigation in nested notebooks by computing relationships in one reactive block and detecting notebooks by structure instead of [NAME]. The up arrow and sibling lists now render consistently.

- **Bug Fixes**
  - Parent link shows reliably in nested notebooks.
  - Sibling detection is stable and no longer goes stale.

- **Refactors**
  - Added getNotebookNotesArray() to handle Cell/Writable vs plain arrays.
  - Filter notebooks by title && !content to avoid proxy [NAME] issues.
  - Combined parents/children/siblings into a single notebookRelationships computed.
  - Used index-based loops instead of .filter() on proxy arrays.
  - Updated UI to read hasParents/hasSiblings and lists from notebookRelationships.

<sup>Written for commit 643c8e0b48e2bf86a0f9ed6bff222c654a4b9df9. Summary will update on new commits.</sup>

### DIFF
--- a/packages/patterns/notes/notebook.tsx
+++ b/packages/patterns/notes/notebook.tsx
@@ -40,6 +40,18 @@ type MinimalCharm = {
   [NAME]?: string;
 };
 
+// Helper to safely get notes array from a notebook (handles Cell/Writable or plain array)
+function getNotebookNotesArray(notebook: unknown): unknown[] {
+  const notes = (notebook as any)?.notes;
+  if (!notes) return [];
+  // Check if it's a Cell/Writable with .get() method
+  if (typeof notes.get === "function") {
+    return notes.get() ?? [];
+  }
+  // Plain array
+  return Array.isArray(notes) ? notes : [];
+}
+
 interface Input {
   title?: Default<string, "Notebook">;
   notes?: Default<NoteCharm[], []>;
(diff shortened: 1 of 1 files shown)


packages/patterns/notes/notebook.tsx
```diff
@@ -1102,74 +1114,75 @@ const Notebook = pattern<Input, Output>(
...
-        if (nbTitle === title) return false;
-        // Skip parent notebooks
-        if (parentTitles.includes(nbTitle)) return false;
-        // Skip child notebooks (they're shown in their own section)
-        if (childTitles.includes(nbTitle)) return false;
-        return true;
-      });
-    });
+        if (typeof nbTitle !== "string") continue;
+        if (nbTitle === currentTitle) continue; // Skip self
```
mathpirate: also fragile checking whether the title is the same

PR #2508, opened 2026-01-13
feat(ts-transformers,js-compiler): improve error message for .get() on OpaqueRef

seefeldb:
## Summary

- Adds helpful error messages when users mistakenly call `.get()` on reactive values (OpaqueRef types) that don't require it
- Error now explains that reactive values from pattern inputs, `computed()`, and `lift()` can be accessed directly
- Suppresses the confusing TypeScript type error (`OpaqueCell<T> & T`) in favor of a clear, actionable message
- Adds `--verbose-errors` flag to show original TypeScript errors when needed for debugging

## Problem

When writing code like:
```typescript
pattern<{ foo: number }>(({ foo }) => {
  const bar = computed(() => foo + 1);
  const baz = computed(() => bar.get() + 1);  // Error!
  return baz;
});
```

TypeScript produces a confusing error about `OpaqueCell<T> & T` intersection types that LLMs and users struggle to understand.

## Solution

Two-layer error handling:

1. **TypeScript type error (js-compiler)**: Replaces the confusing type error with a clear message
2. **Transformer validation (ts-transformers)**: New `OpaqueGetValidationTransformer` provides clear errors with fix suggestions

### Default output (LLM-friendly):
```
[ERROR] Unnecessary .get() call on a reactive value. This value can be accessed 
directly - remove .get(). Reactive values passed to pattern (except Writable<T> 
and Stream<T>) and results from computed() and lift() don't need .get(). Only 
Writable<T> requires .get() to read values.

4 | export default pattern<{ foo: number }>(({ foo }) => {
5 |   const bar = computed(() => foo + 1);
6 |   const baz = computed(() => bar.get() + 1);
  |                                  ^
```

### With `--verbose-errors` (for debugging):
```bash
deno task ct dev ./recipe.tsx --verbose-errors
```

Shows the original TypeScript error appended to the message.

## Test plan

- [x] Added test cases in `packages/ts-transformers/test/validation.test.ts`
- [x] All existing tests pass
- [x] Verified error message output manually
- [x] Tested `--verbose-errors` flag

### DIFF
--- a/packages/js-compiler/typescript/diagnostics/errors.ts
+++ b/packages/js-compiler/typescript/diagnostics/errors.ts
@@ -17,7 +17,6 @@ export interface DiagnosticMessageTransformer {
 export interface ErrorDetails {
   readonly diagnostic: Diagnostic;
   source?: string;
-  messageTransformer?: DiagnosticMessageTransformer;
 }
 
 /**
@@ -45,7 +44,10 @@ export class CompilationError {
   message: string;
   type: CompilationErrorType;
 
-  constructor({ diagnostic, source, messageTransformer }: ErrorDetails) {
+  constructor(
+    { diagnostic, source }: ErrorDetails,
+    messageTransformer?: DiagnosticMessageTransformer,
+  ) {
     const { file, start } = diagnostic;
     const { message, type } = this.parseMessage(
       diagnostic.messageText,
@@ -123,8 +125,13 @@ export class CompilationError {
 export class CompilerError extends Error {
   override name = "CompilerError";
   #errors: CompilationError[];
-  constructor(errorDetails: ErrorDetails[]) {
-    const errors = errorDetails.map((d) => new CompilationError(d));
+  constructor(
+    errorDetails: ErrorDetails[],
+    messageTransformer?: DiagnosticMessageTransformer,
+  ) {
+    const errors = errorDetails.map((d) =>
+      new CompilationError(d, messageTransformer)
+    );
     const message = errors.map((error) => error.displayInline()).join("\n");
     super(message);
     this.#errors = errors;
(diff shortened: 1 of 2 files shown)


packages/cli/commands/dev.ts
```diff
@@ -37,8 +37,16 @@ export const dev = new Command()
...
     'Named export from entry for recipe definition. Defaults to "default".',
   )
+  .option(
+    "--verbose-errors",
+    "Show original TypeScript error messages in addition to simplified hints.",
+  )
   .arguments("<main:string>")
   .action(async (options, main) => {
+    // Set environment variable for verbose error messages
+    if (options.verboseErrors) {
```
jsantell: This env can be set `CT_VERBOSE_ERRORS=1 ct dev ..` if we want env vars. Challenges with env vars are a global singleton with extra rules (rather than passing an arg to where it needs to go)

seefeldb: removed, but now it added a new export from js-compiler. see my other comment.

packages/js-compiler/typescript/diagnostics/errors.ts
```diff
@@ -54,6 +54,39 @@ export class CompilationError {
...
+
+    // Detect .get() called on OpaqueCell/OpaqueRef types
+    // TypeScript error: "Property 'get' does not exist on type 'OpaqueCell<...> & ...'"
+    // Replace with a clear, actionable message (suppress the confusing type error)
+    // Use CT_VERBOSE_ERRORS=1 to show the original TypeScript error as well
+    {
+      const match = message.match(
+        /^Property 'get' does not exist on type '(OpaqueCell<[^']*>)/,
+      );
+      if (match) {
```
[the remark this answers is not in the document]
jsantell: Realizing these are native *TypeScript* errors clarifies a lot! I understand now, thank you

PR #2518, opened 2026-01-14
docs: clarify Writable input coercion and test instantiation

mathpirate:
Document the semantic distinction when passing values to patterns expecting Writable<T> inputs:

- Plain values create independent state per pattern instance
- Cell.of() references share state across instances

Update pattern-testing.md to recommend plain values for test instantiation (most common case) and link to the new documentation.

This addresses a documentation gap discovered while investigating agent-generated pattern code that was using Cell.of() unnecessarily.

---
## Summary by cubic
Clarifies how Writable<T> inputs are handled: plain values create independent state per pattern instance, while Writable.of() shares state across instances. Updates pattern-testing.md to recommend plain values in tests and links to the new guidance to avoid unintended shared state.

<sup>Written for commit 43f82f32e4d237d9b73dfd86663cafd289458eb2. Summary will update on new commits.</sup>

### DIFF
--- a/docs/common/concepts/types-and-schemas/writable.md
+++ b/docs/common/concepts/types-and-schemas/writable.md
@@ -35,9 +35,9 @@ With `Writable<T>` in your signature:
 
 Without `Writable<>`, you can still display values in JSX, pass to `computed()`, and map over arrays - all reactively. Note: filtering and transformations must be done in `computed()` outside JSX, then the result can be mapped inside JSX.
 
-### Passing Values to Writable Inputs
+### Passing Values to Pattern Inputs
 
-When calling a pattern that expects `Writable<T>`, you have two options:
+When calling a pattern, you have two options for providing input values:
 
 **Plain values** create independent state for each pattern instance:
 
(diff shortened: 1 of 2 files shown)


docs/common/concepts/types-and-schemas/writable.md
```diff
@@ -34,3 +34,26 @@ With `Writable<T>` in your signature:
 | `.key("property")` | Navigate nested data |
 
 Without `Writable<>`, you can still display values in JSX, pass to `computed()`, and map over arrays - all reactively. Note: filtering and transformations must be done in `computed()` outside JSX, then the result can be mapped inside JSX.
+
+### Passing Values to Writable Inputs
+
+When calling a pattern that expects `Writable<T>`, you have two options:
```
seefeldb: i don't think the Writable<> on the pattern matters here, it's going to be owned state even if it isn't Writable. Note that the pattern can still pass reference to those to other things, e.g. a handler that has Writable<> inputs and so they are effectively writable anyway.

docs/common/workflows/pattern-testing.md
```diff
@@ -43,6 +43,8 @@ export default pattern(() => {
 });
 ```
 
+**Note:** Pass plain values when instantiating patterns in tests. The runtime creates independent writable cells automatically. Use `Cell.of()` only when you need to test shared state behavior. See [Writable](../concepts/types-and-schemas/writable.md#passing-values-to-writable-inputs) for details.
```
seefeldb: Use Writable.of generally, instead of Cell.of

mathpirate: Oops thanks

PR #2571, opened 2026-01-17
feat(ui): filter current charm from mentions autocomplete

tonyespinoza1:
## Summary

- Adds `selfId` property to `ct-code-editor` to exclude the current charm from the [[mentions]] autocomplete dropdown
- Prevents users from accidentally mentioning the note they're currently editing
- Uses NAME comparison for filtering since CellHandle IDs differ from charm IDs

## Changes

- **ct-code-editor.ts**: Add `selfId` prop, cache raw mentionable data, filter by NAME match
- **mentionable.ts**: Add `asCell: true` to schema for proper CellHandle IDs  
- **note.tsx**: Pass `selfName` computed prop (`📝 ${title}`)
- **jsx.d.ts**: Add `selfId` to CTCodeEditorAttributes type

## Test plan

- [x] Open a note in the browser
- [x] Type `[[` to trigger mentions autocomplete
- [x] Verify the current note does NOT appear in the dropdown
- [x] Verify other notes still appear and can be selected
- [x] Verify selecting a mention still works correctly

---
## Summary by cubic
Prevents self-mentions by excluding the open note from the [[mentions]] autocomplete. Adds a self CellHandle to ct-code-editor and improves filtering and backlink navigation.

- **New Features**
  - ct-code-editor accepts self (CellHandle) and excludes the current charm by NAME; completions use {cell, name}.
  - Cache raw mentionable names for reliable search; pass $self from the note and add $self to JSX.
  - Backlink pills: click navigates; pending pills create the charm and then navigate.

- **Bug Fixes**
  - Use explicit JSONSchema types to avoid TS2589 errors; fix subscription leak for raw mentionable data.

<sup>Written for commit d6388a1edcfca3c471585c88f1b6bd545efbe366. Summary will update on new commits.</sup>

### DIFF
--- a/packages/html/src/jsx.d.ts
+++ b/packages/html/src/jsx.d.ts
@@ -3648,7 +3648,7 @@ interface CTCodeEditorAttributes<T> extends CTHTMLAttributes<T> {
   "onct-blur"?: any;
   "onbacklink-click"?: any;
   "onbacklink-create"?: any;
-  "$self"?: CellLike<any>;
+  "$self"?: any;
 }
 
 interface CTAutoLayoutAttributes<T> extends CTHTMLAttributes<T> {


packages/ui/src/v2/components/ct-code-editor/ct-code-editor.ts
```diff
@@ -306,6 +307,7 @@ export class CTCodeEditor extends BaseElement {
     mentionable: { type: Object },
     mentioned: { type: Array },
     pattern: { type: Object },
+    selfId: { attribute: false },
```
seefeldb: self

packages/ui/src/v2/components/ct-code-editor/ct-code-editor.ts
```diff
@@ -458,62 +464,103 @@ export class CTCodeEditor extends BaseElement {
...
-      const mention = mentionableData[i];
-      if (
-        mention &&
-        mention[NAME]
-          ?.toLowerCase()
-          ?.includes(queryLower)
+    const matches: Array<{ cell: CellHandle<Mentionable>; name: string }> = [];
+
+    // Extract selfId value once before loop
+    let selfIdStr = "";
```
seefeldb: All this can go away if self is always just a cell handle and one can compare them directly via .equals

packages/ui/src/v2/components/ct-code-editor/ct-code-editor.ts
```diff
@@ -458,62 +464,103 @@ export class CTCodeEditor extends BaseElement {
...
+    for (let i = 0; i < cellHandles.length; i++) {
+      const cell = cellHandles[i];
+      if (!cell) continue;
+
+      // Get name from cached raw data
+      const rawItem = this._rawMentionableData[i];
+      const name = rawItem?.[NAME] ?? "";
+
+      // Skip the current charm (compare by NAME)
+      if (selfIdStr && name && name === selfIdStr) {
```
seefeldb: Compare via .equals instead

packages/ui/src/v2/components/ct-code-editor/ct-code-editor.ts
```diff
@@ -701,40 +799,72 @@ export class CTCodeEditor extends BaseElement {
...
   ): Promise<void> {
     try {
       // Simple random ID generator for noteId (matches pattern used in note.tsx)
       const generateId = () =>
         `${Date.now().toString(36)}-${Math.random().toString(36).slice(2, 11)}`;
       const rt = this.pattern.runtime();
-      const program = this.pattern.get();
+
+      // Try to get the pattern value - may need to wait for it via subscription
+      let program = this.pattern.get();
```
seefeldb: Huh, that seems hairy and digging way too deep in internals.

This makes a copy of the current container pattern, but not clear that's what we want.

I'd rather handle that via the call back and so caller can decide what a new page looks like.

PR #2658, opened 2026-01-22
series of transformer fixes/improvements

mathpirate:
This PR fixes transformer bugs related to reactive context handling, specifically CT-1186.                                                           
                                                                                                                                                       
  Key Changes                                                                                                                                          
                                                                                                                                                       
  1. Transform && and || to when()/unless() in safe contexts                                                                                           
                                                                                                                                                       
  Problem: OpaqueRefJSXTransformer skipped ALL transformations inside safe callback wrappers (action, handler, computed, derive, lift). This was       
  incorrect because expressions like computed(() => showPanel) && <Panel /> need && transformed to when() - OpaqueRef objects are always truthy in     
  JavaScript, so <Panel /> would always render regardless of the actual value.                                                                         
                                                                                                                                                       
  Fix: Added inSafeContext flag passed through the rewrite pipeline. In safe contexts:                                                                 
  - && still transforms to when()                                                                                                                      
  - || still transforms to unless()                                                                                                                    
  - Property access, template literals, ternary, etc. do NOT wrap in derive() (already in reactive context)                                            
                                                                                                                                                       
  2. Don't transform .map() to mapWithPattern inside derive                                                                                            
                                                                                                                                                       
  Problem: Inside derive() callbacks, OpaqueRef arrays are auto-unwrapped to plain arrays. The transformer was incorrectly transforming .map() to      
  mapWithPattern when it shouldn't.                                                                                                                    
                                                                                                                                                       
  Fix: shouldTransformMap() now checks isInsideSafeCallbackWrapper() and only transforms Cell/Stream types (not OpaqueRef) in those contexts.          
                                                                                                                                                       
  3. Compile-time error for .map() on fallback expressions                                                                                             
                                                                                                                                                       
  Problem: Calling .map() on (x ?? []) or (x || []) where one side is reactive fails at runtime because the transformer can't detect the reactive type 
  through the binary expression.                                                                                                                       
                                                                                                                                                       
  Fix: New PatternContextValidationTransformer reports a compile-time error with guidance to use direct property access instead.                       
                                                                                                                                                       
  Test Coverage                                                                                                                                        
                                                                                                                                                       
  - New safe-context-and-jsx fixture validating && → when() and || → unless() in handlers                                                              
  - New validation tests for .map() on fallback expressions                                                                                            
  - Updated expected outputs for fixture tests to match correct behavior                                                                               
                                                                                                                                                       
  Breaking Changes                                                                                                                                     
                                                                                                                                                       
  None. These are correctness fixes.


packages/ts-transformers/test/fixtures/closures/pattern-nested-jsx-map.expected.tsx
```diff
@@ -73,85 +73,7 @@ export default pattern(({ items }) => {
...
-                                type: "array",
-                                items: true
-                            }, {}, {
-                                type: "null"
-                            }]
-                    }
-                }
-            }
-        } as const satisfies __ctHelpers.JSONSchema, {
+        } as const satisfies __ctHelpers.JSONSchema, true as const satisfies __ctHelpers.JSONSchema, {
```
seefeldb: it looks like the return type of the map is now `true` instead of the vdom-y structure, but it probably should still be the larger schema?

PR #2667, opened 2026-01-22
fix(runtime-client): late subscribers receive initial value from existing subscriptions

mathpirate:
Summary:                                                                                                                                             
  ## Summary                                                                                                                                           
  - When multiple CellHandle instances subscribe to the same cell (same subscription key), late subscribers would miss the initial value because no new
   backend request was made                                                                                                                            
  - This caused text interpolation `{value}` to show blank on initial page load when used alongside `ct-input` bound to the same cell                  
  - Fix: In `connection.subscribe()`, copy the cached value from an existing subscriber to new subscribers joining an existing subscription            
                                                                                                                                                       
  ## Test plan                                                                                                                                         
  - [x] Added regression test in `packages/runtime-client/integration/client.test.ts`                                                                  
  - [ ] Manual testing with `packages/patterns/gideon-tests/late-subscriber-repro.tsx`

### DIFF
--- a/packages/runtime-client/integration/client.test.ts
+++ b/packages/runtime-client/integration/client.test.ts
@@ -239,6 +239,97 @@ describe("RuntimeClient", () => {
       cancel1();
       cancel2();
     });
+
+    it("late subscribers receive initial value from existing subscription", async () => {
+      // Regression test for bug where text interpolation {value} would show blank
+      // when used alongside ct-input bound to the same cell. The issue was that
+      // late subscribers (those joining an existing subscription) would miss the
+      // initial value that was already sent to earlier subscribers.
+      //
+      // Fix: connection.subscribe() copies cached value from existing subscriber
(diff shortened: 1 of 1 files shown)


packages/ui/src/v2/components/ct-input/ct-input.ts
```diff
@@ -493,7 +492,9 @@ export class CTInput extends BaseElement {
         if (changedProperties.has("value")) {
           // Bind the new value (Cell or plain) to the controller
           // This updates the internal reference so getValue() returns the correct value
-          this._cellController.bind(this.value, stringSchema);
+          // Note: Don't pass stringSchema here - use the CellHandle's existing schema
+          // to avoid creating a new CellHandle that won't share subscriptions
+          this._cellController.bind(this.value);
```
bfollington: Makes me wonder if `_cellController.bind` even needs the ability take a schema... Do we depend on this anywhere?

PR #2690, opened 2026-01-28
feat(profile): add #profile wish tag and profile-aware patterns

bfollington:
Add the ability to wish for the user's profile summary via #profile tag,
and inject profile context into key LLM operations.

Changes:
- Add #profile built-in tag in wish.ts that resolves to learned.summary
- Add LearnedSection schema to home-schemas for proper cross-space access
- Create profile-aware-writer.tsx example pattern demonstrating profile injection
- Simplify suggestion.tsx to use #profile directly instead of manual context building
- Add profile context injection to omnibox-fab.tsx system prompt
- Add updateProfile tool to allow Omnibar to edit the profile summary

The #profile tag required special handling: the intermediate learned cell
must be resolved before accessing the nested summary property, unlike
top-level properties which work with simple pathPrefix chaining.

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

---
## Summary by cubic
Adds a #profile wish tag that returns the user’s profile summary and threads it into key patterns. This improves personalization and lets Omnibox read and update the profile.

- **New Features**
  - Added #profile wish that resolves to learned.summary with proper cell resolution.
  - Introduced LearnedSection schema and added learned to homeSchema for cross-space access.
  - New updateProfile tool to set the profile summary.
  - Omnibox now injects profile context into its system prompt and exposes updateProfile.
  - Added profile-aware-writer example pattern using #profile for personalized generation.
  - Simplified suggestion pattern to use #profile instead of assembling profile context manually.
  - Updated patterns docs to include the new example.

<sup>Written for commit 5591e1a2f4d902b2bc2f7a481ce02d9225fa9302. Summary will update on new commits.</sup>

### DIFF
--- a/packages/patterns/system/common-tools.tsx
+++ b/packages/patterns/system/common-tools.tsx
@@ -305,12 +305,7 @@ export const updateProfile = pattern<
     if (!cell) return { success: false, message: "Profile not available" };
 
     // Set the new summary text
-    if (typeof cell.set === "function") {
-      cell.set(summary);
-    } else {
-      // If it's not a writable, try direct assignment
-      (cell as any) = summary;
-    }
+    cell.set(summary);
 
     return {
       success: true,


packages/runner/src/builtins/wish.ts
```diff
@@ -249,6 +249,25 @@ function resolveBase(
...
+      if (!userDID) {
+        throw new WishError("User identity DID not available for #profile");
+      }
+
+      // First resolve to the learned cell, then access summary
+      // This ensures the intermediate cell is resolved before accessing nested property
+      const learnedCell = ctx.runtime.getHomeSpaceCell(ctx.tx)
+        .key("defaultPattern")
+        .key("learned")
+        .resolveAsCell();
```
bfollington: For some reason this only resolves if step to the learned cell, then grab the `summary` field as a seperate step... Maybe due to being a primitive? Or `learned` being a cell?

PR #2696, opened 2026-01-28
fix(pattern): note.tsx no longer uses as any

seefeldb:
## Summary by cubic
Removed unsafe any casts in Note pattern and replaced them with proper typing and optional chaining to improve type safety without changing behavior.

- **Refactors**
  - Access notebook and self.parentNotebook via optional chaining instead of casting.
  - Return typed notebooks (NotebookCharm[]) and cast containingNotebookNames to string[] for includes checks.
  - Read linkPattern reactively via .get() when available.

<sup>Written for commit 87055f68366a20caf8acdd5990a2d58fcea857d4. Summary will update on new commits.</sup>

### DIFF
--- a/packages/patterns/notes/note.tsx
+++ b/packages/patterns/notes/note.tsx
@@ -205,7 +205,7 @@ const createNewNote = handler<
 
   // Add to parent notebook using Cell.key() pattern
   if (notebook) {
-    const charmsList = allPieces.get();
+    const piecesList = allPieces.get();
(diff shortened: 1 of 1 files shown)


ubik2 (commented):
looks good to me.
there's probably still some situations where allCharms.get returns undefined, but these aren't related to the any casts.

ubik2 (approved):
forgot to check approve

PR #2717, opened 2026-01-30
Avoid a top level read in followPointer

ubik2:
Avoid a top level read in followPointer, since we don't want to taint the reactivity.

This means we attempt to read at the full depth, back out to the last valid read and try a getAtPath there.

---
## Summary by cubic
Fix followPointer reactivity by avoiding top-level reads. Traversal now reads at full depth, backs off to the last valid node, and continues without subscribing to unrelated cells.

- **Bug Fixes**
  - Read at the target path and back off to the last existing segment, then continue via getAtPath to avoid tainting reactive dependencies.
  - Track visited docs with a new trackVisitedDoc helper; load sources non-reactively when includeSource is true.
  - Improve link handling: prepend "value" to target link paths, re-root selectors correctly, and avoid double-prefixing.
  - Handle NotFound cases more precisely (missing doc vs missing path) and return notFound(target) with the correct address.
  - Fix redirect resolution: if the final link is invalid, base itemLink on the last valid redirect doc.

<sup>Written for commit e6683ca62c31af9bdc8023d3eccc5de724b1fd9c. Summary will update on new commits.</sup>

### DIFF
--- a/packages/runner/src/traverse.ts
+++ b/packages/runner/src/traverse.ts
@@ -889,7 +889,7 @@ function getTrackerKey(
  * source docs as read if needed.
  *
  * I can't just use resolveLink, since I need to also track all the
- * intermediate docuemnts if we includeSource.
+ * intermediate documents if we includeSource.
  *
  * @param tx - IStorageTransaction that can be used to read data
  * @param doc - IAttestation for the current document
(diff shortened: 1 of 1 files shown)


ubik2:
I'm not sure this will solve the reactivity issues, but it's the one place I'm aware of where I was doing a broader read than I need.
There's still going to be some taint, since I might do a read of foo with a schema that ignores foo.bar, but I'm reading the foo object to verify that it's an object and see its properties, so I'm going to be notified when foo.bar is updated. If the reads are across cell boundaries, that won't happen.

seefeldb (approved):

PR #2720, opened 2026-01-30
fix(html): diff root VNodes before destroying in reconcileIntoWrapper

mathpirate:
When a Cell<VNode> at the root emits a new value, the reconciler was unconditionally destroying and recreating the entire UI tree. This caused:
- Modal flicker when opening (modal destroyed and recreated)
- Component state loss
- Performance degradation from unnecessary DOM churn

Now reconcileIntoWrapper diffs the old and new VNodes:
- Same element type → Update props and children in place (reuse DOM node)
- Different element type → Destroy and recreate (previous behavior)

Key implementation details:
- Save the children map before cancel() clears it
- Restore it after cancel so keyed reconciliation can work
- Clear old event handlers before re-registering via bindProps

---
## Summary by cubic
Diff root VNodes in reconcileIntoWrapper to update in place when element types match, instead of always destroying the tree. Fixes modal flicker, preserves component state, and reduces DOM churn.

- **Bug Fixes**
  - Reuse the DOM node for the same element type; diff props and children in place and skip updates when the same Cell is reused.
  - Track Cell references in propSubscriptions and childrenState to avoid unnecessary re-subscriptions.
  - Add cycle detection when following the [UI] chain; properly remove nodes and props by falling through to destroy when sanitizeNode returns null and emitting remove-prop ops.

<sup>Written for commit e55b6fb4031fcd824eb50e23e7525b3c10d081a8. Summary will update on new commits.</sup>

### DIFF
--- a/packages/html/src/worker/reconciler.ts
+++ b/packages/html/src/worker/reconciler.ts
@@ -271,7 +271,7 @@ export class WorkerReconciler {
    */
   private createWrapperState(_ctx: ReconcileContext, nodeId: number): {
     nodeId: number;
-    currentChild: NodeState | ChildNodeState | null;
+    currentChild: NodeState | null;
     cancel: Cancel;
   } {
     return {
@@ -313,7 +313,7 @@ export class WorkerReconciler {
     ctx: ReconcileContext,
     wrapper: {
       nodeId: number;
-      currentChild: NodeState | ChildNodeState | null;
+      currentChild: NodeState | null;
       cancel: Cancel;
     },
     node: WorkerRenderNode,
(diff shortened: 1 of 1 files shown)


packages/html/src/worker/reconciler.ts
```diff
@@ -319,6 +382,313 @@ export class WorkerReconciler {
...
+        if (state.eventHandlers.has(key)) {
+          ctx.unregisterHandler(state.eventHandlers.get(key)!);
+          state.eventHandlers.delete(key);
+        }
+        // Send remove op (set to null/undefined)
+        this.queueOps([{
+          op: "set-prop",
+          nodeId: state.nodeId,
+          key,
+          value: null,
```
seefeldb: null is a valid value for a prop, is this really removing it?

packages/html/src/worker/reconciler.ts
```diff
@@ -291,7 +318,44 @@ export class WorkerReconciler {
...
+      : null;
+    const newTagName = newVNode?.name ?? null;
+
+    // Case 1: Same element type AND we have a full NodeState - update in place
+    if (
+      oldState &&
+      oldTagName &&
+      newTagName &&
+      oldTagName === newTagName &&
+      "children" in oldState // Has full NodeState (not just ChildNodeState)
```
seefeldb: why is this necessary, what if there a node that has no children but the props update (and the type stays the same)

seefeldb (approved):

PR #2731, opened 2026-02-02
fix(runner): handle inline objects with asCell in array traversal

mathpirate:
When traversing arrays with items marked as asCell, inline objects (not links) need a DataCellURI-style address with path: ["value"] before getNextCellLink can generate a proper normalized link.

Commit 231acd0c2 removed the isPrimitiveCellLink guard to respect asCell boundaries for inline objects, but didn't account for the case where curDoc.address.path doesn't start with "value" (e.g., VDOM children arrays have paths like ["$UI", "children", "0"]).

This fix creates the proper address for inline objects before calling getNextCellLink, preventing the "Unable to create link to non-value address" error.

---
## Summary by cubic
Fix array traversal for asCell items by wrapping elements whose address path doesn’t start with "value" in a DataCellURI before link resolution. Prevents "Unable to create link to non-value address" and restores link normalization for VDOM children and other non-"value" paths.

- **Bug Fixes**
  - If curDoc.address.path[0] !== "value", create a DataCellURI from the element, set path to ["value"], then call getNextCellLink (handles inline values and links resolving to non-value paths).
  - Skip wrapping when the path already starts with "value" to preserve existing link IDs.

<sup>Written for commit 68b04a9188298dc2c699e4e688a64f19c2d1ccac. Summary will update on new commits.</sup>

### DIFF
--- a/packages/runner/src/traverse.ts
+++ b/packages/runner/src/traverse.ts
@@ -1929,14 +1929,15 @@ export class SchemaObjectTraverser<V extends JSONValue>
         !this.traverseCells &&
         SchemaObjectTraverser.asCellOrStream(curSelector.schema)
       ) {
-        // For non-link values, we need to create a DataCellURI-style address
-        // with path: ["value"] so that getNextCellLink can generate a proper
-        // normalized link. We check item (not curDoc.value) because if item
-        // was a link, we've already followed it and curDoc.address has the
-        // correct target address - we don't want to overwrite it with a
-        // DataCellURI. This handles inline objects, primitives, arrays, null,
-        // and undefined.
-        if (!isPrimitiveCellLink(item)) {
+        // For values where curDoc.address.path doesn't start with "value",
+        // we need to create a DataCellURI-style address so that getNextCellLink
+        // can generate a proper normalized link via getNormalizedLink.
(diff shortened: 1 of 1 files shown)


mathpirate:
> This comment isn't quite right:
> 
> > Commit [231acd0](https://github.com/commontoolsinc/labs/commit/231acd0c23382d13edfc72f0677633f042b53c35) removed the isPrimitiveCellLink guard to respect asCell boundaries for inline objects, but didn't account for the case where curDoc.address.path doesn't start with "value" (e.g., VDOM children arrays have paths like ["$UI", "children", "0"]).
> 
> The `curDoc.address.path` is relative to the `fact.is`, so it will almost always start with `value` (sometimes it will start with `source` instead, but then we shouldn't be using schema). Meanwhile link objects and the more client view of our data is relative to `fact.is.value`, so those don't include the `value` in the path. We should be applying the vdom schema to the data at the `value` path, or deeper.
> 
> However, I was concerned about removal of the isPrimitiveCellLink. I haven't dug into how that changes things.

thank you! the fix now checks curDoc.address.path[0] !== "value" directly. tests pass locally again and this resolves the issue i'm having with getting notebooks deployed locally

PR #2745, opened 2026-02-05
docs(space-model): space model specification

danfuzz:
## Summary
- Full specification for the space data model covering data types, storage, identity, cells, transactions, reactivity, and schemas
- **Late Serialization proposal**: `StorableInstance` protocol with `deconstruct`/`reconstruct` methods, `SerializationContext` for tag<->class mapping, `UnknownStorable` for forward compatibility
- **JSON Encoding proposal**: Unified `/<type>@<version>` keys replacing three current conventions, with `/object` and `/quote` escape mechanisms
- CRDT/Automerge exploration as a proposed storage direction

---
## Summary by cubic
Adds a full Space Model spec with seven docs and proposals that simplify serialization, hashing, and storage while clarifying cells, transactions, reactivity, and schema behavior. Also documents per-invocation UUIDs for event handlers to explain distinct result identities.

- **New Features**
  - Seven spec docs with README navigation and cross-links.
  - Proposals: Late Serialization (StorableInstance/SerializationContext/UnknownStorable), unified JSON encoding with "/<Type>@<Version>" plus /object and /quote escapes, simplified canonical hashing, and a CRDT/Automerge storage direction.

- **Refactors**
  - Data/identity: cycles limited to within a document and relaxable with future formats; entity IDs are stable addresses; hashing over causal chains; softened IPFS notes; Late Serialization not “zero transformations.”
  - Cells/streams: clarified Writable<T> naming and planned auto-narrowing; constraints for stream unification (scheduler context, trigger semantics, DX); stream marker semantics.
  - Transactions: baseline-capture conflict detection; live-reference proxies scoped to any-typed cells; no nested transactions; UCAN-based serialization; no cross-space guarantees.
  - Reactivity: dynamic, state-dependent dependency graph; clarified scheduling/cycle/consistency; each handler invocation gets a unique UUID to derive distinct result cell identities even with identical payloads; event timestamp/replay considerations.
  - Schemas: clarified validation behavior for non-conforming data (what’s rejected vs undefined/null shaping).

<sup>Written for commit 47846f55b8c82bf421b87165031bcb00ae27c595. Summary will update on new commits.</sup>

### DIFF
--- a/docs/specs/space-model/1-data-model.md
+++ b/docs/specs/space-model/1-data-model.md
@@ -136,6 +136,11 @@ Cycles *across* documents are supported via explicit links (sigil links). Two
 cells can reference each other, forming a cycle in the broader data graph. The
 no-cycles constraint applies only to the serializable content of a single cell.
 
+The within-document prohibition is inherited from JSON's tree structure, not from
+a deep architectural requirement. If a future storage format supports cyclic
+references natively (e.g. CBOR with shared references, or a CRDT layer with
+internal pointers), this constraint could be relaxed.
+
 ---
 
 ## Hashing and Content Addressing


docs/specs/space-model/5-transactions.md
```diff
@@ -0,0 +1,149 @@
...
+### Overview
+
+Transactions provide atomicity and consistency for cell operations. All writes
+to cells require a transaction context.
+
+### Transaction Lifecycle
+
+1. **Open**: `runtime.edit()` creates a new transaction
+2. **Read**: `cell.withTx(tx).get()` reads within transaction context
+3. **Write**: `cell.withTx(tx).set(value)` buffers write
```
seefeldb: there's currently also .push, .remove and (somewhat less ideal in design) .update. there could be more of these, particularly around array manipulation. .update could be something using a callback that gets a proxy, like automerge does it.

docs/specs/space-model/6-reactivity.md
```diff
@@ -0,0 +1,187 @@
...
+If cells are unified via timestamps (see [Cells](./4-cells.md)), change detection
+would naturally handle both cases via content comparison on timestamped data.
+This would eliminate the special-casing for streams.
+
+---
+
+## Open Questions
+
+- What is the exact scheduling algorithm (priority, ordering)?
+- How are cycles in the dependency graph handled?
```
seefeldb: detected and re-executed in a tight loop with some bounds. the idea is that they must converge quickly.

docs/specs/space-model/5-transactions.md
```diff
@@ -0,0 +1,149 @@
...
+These ~10 methods form the core data access API. Everything else (reactivity,
+streaming) builds on top.
+
+---
+
+## Open Questions
+
+- How do nested/child transactions work (if at all)?
+- How are transactions serialized for storage?
+- What consistency guarantees exist across spaces?
```
seefeldb: none so far

docs/specs/space-model/7-schemas.md
```diff
@@ -0,0 +1,146 @@
...
+  }
+}
+```
+
+### Schema-Driven Behavior
+
+Schemas influence runtime behavior:
+- **Transformation**: Schema-aware traversal resolves references and shapes data.
+  Non-conforming values may return as `undefined` rather than being strictly
+  rejected — this is transformation, not validation in the strict sense.
```
seefeldb: they are only `undefined` if optional (and `null` in the array case if `null` is allowed), otherwise we are now rejecting the data.

that is, you are always getting something valid or `undefined` at the top.

docs/specs/space-model/1-data-model.md
```diff
@@ -0,0 +1,748 @@
...
+identity whether later serialized to JSON, CBOR, or Automerge.
+
+#### Open Questions
+
+- What is the exact specification for canonical hashing?
+- How should each type be tagged? (null, bool, int, float, string, bytes, array, object, references)
+- How do special object shapes (references, streams, errors) participate?
+- What is the migration path from current CID-based identifiers?
+
+### Late Serialization: Rich Types Within the Runtime
```
seefeldb: what i like the most about this is the realization that at least on the client we might not actually ever send the computed data to the server, so there is no need for serialization.

however there is still a transformation happening on every read, as the schema is selecting data, when constructing links we still need to know what document data would belong to, etc. -- the safest way to do this is to make a copy each time.

with CFC we'll also want a validation step on write, so traverse the data anyway.

i think this eliminates one copy, maybe sometimes more when we detect data that was passed through (so frozen before passed to the untrusted function).

seefeldb (commented):
interesting proposals, mostly added historic context and maybe missing pieces.

might add more later, but wanted to send it off before lunch.

danfuzz:
For context, the only part that I think is done enough to start implementing is `1-data-model`, and even so I'm _sure_ we'll be in "no plan survives first contact with the enemy" territory as soon as I start to actually try making changes to the codebase.

PR #2754, opened 2026-02-06
fix(runtime-client): wrap CellHandle.toJSON() in sigil link format

mathpirate:
CellHandle.toJSON() was returning a raw CellRef object, but the runtime only auto-dereferences links in sigil format ({"/": {"link@1": ...}}).

This caused event.detail.sourceCell.get() to return undefined when using ct-drag-source and ct-drop-zone for drag-and-drop, because the CellRef was treated as raw data instead of being resolved to the actual cell value.

---
## Summary by cubic
Wraps CellHandle.toJSON() in sigil link format so the runtime auto-dereferences links. Fixes ct-drag-source/ct-drop-zone where event.detail.sourceCell.get() returned undefined; it now resolves to the cell value.

<sup>Written for commit 900a62dd0b94a9939dab3fa7040d0ac237302df0. Summary will update on new commits.</sup>

### DIFF
--- a/packages/runtime-client/cell-handle.ts
+++ b/packages/runtime-client/cell-handle.ts
@@ -266,8 +266,14 @@ export class CellHandle<T = unknown> {
     };
   }
 
-  toJSON(): CellRef {
-    return { ...this.ref() };
+  toJSON() {
+    // Wrap in sigil link format so the runtime recognizes this as a link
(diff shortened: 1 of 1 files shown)


seefeldb (approved):
lgtm,

there might be something in the google auth flow that relied on the old behavior. i'm 95% sure that code also uses generic parsing (so would understand either), but if suddenly auth breaks, that's what we should look there.

PR #2766, opened 2026-02-09
fix(llm-dialog): fix tool call execution broken by unified traversal [CT-1203]

bfollington:
The unified traversal commit (5d352fc09) made validateAndTransform
significantly stricter about schema matching. This broke all LLM tool
calls because:

1. **Schema mismatch kills tool execution**: When handleInvoke() calls
   runtime.run() to start a tool pattern, the scheduler validates the
   action's argument against module.argumentSchema. The compiled recipe
   has `asOpaque: true` on argument properties, but the $alias links
   (created by the binding system) carry different schemas without
   asOpaque. The stricter validateAndTransform sees this as a mismatch,
   returns undefined, and the action silently skips calling the tool
   function (only logged at debug level).

2. **Internal nodes also affected**: Even after fixing the top-level
   argument, internal builtin nodes (computed, fetchData, etc.) within
   the tool pattern also fail validation for similar schema mismatches.
   "ref" type modules resolve to original registry entries, bypassing
   any schema fixes on the recipe copy.

3. **Result cell unreadable**: The result cell was created with the
   pattern's resultSchema, causing get() to also fail schema validation
   when reading the output.

Fix (workaround):

- `prepareRecipeForToolCall()`: Sets argumentSchema to `false` on all
  node modules in an ephemeral recipe copy. When argumentSchema is false,
  the action always runs regardless of validation. This is safe because
  it only affects the one-shot tool call recipe, not the shared original.

- Create result cell without a schema so get() resolves aliases using
  the permissive default schema.

- Replace the sink-based waiting (which also broke due to dependency
  tracking changes) with idle-loop polling.

The proper fix would be in the unified traversal itself: $alias
resolution should be lenient about schema annotation differences
(asOpaque, description, etc.) between the alias schema and the expected
schema. These are metadata flags, not type constraints, and should not
cause validation rejection.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

---
## Summary by cubic
Fixes hanging LLM tool calls by disabling strict schema checks in tool patterns and switching to reliable result polling. Aligns with CT-1203 by restoring working tool calls in llm-dialog, including async and handler-based tools.

- **Bug Fixes**
  - Run tool patterns via an ephemeral recipe that sets argumentSchema=false on all nodes to bypass strict validation.
  - Create the result cell without a schema and prefer handler onComplete; otherwise poll for completion with a timeout instead of using sink subscriptions.
  - Add an async tool call test with delayed result and make the wait helper timeout configurable.

<sup>Written for commit fde685ff762ebfd717add700c661fbc47f2c1cf8. Summary will update on new commits.</sup>

### DIFF
--- a/packages/runner/src/builtins/llm-dialog.ts
+++ b/packages/runner/src/builtins/llm-dialog.ts
@@ -54,6 +54,35 @@ const REQUEST_TIMEOUT = 1000 * 60 * 5; // 5 minutes
 const TOOL_CALL_TIMEOUT = 1000 * 30 * 1; // 30 seconds
 
 /**
+ * Create a copy of a recipe with argumentSchema set to `false` on all node
+ * modules. This disables the strict argument validation in
+ * instantiateJavaScriptNode that was introduced by the unified traversal
+ * (5d352fc09). When argumentSchema is `false`, the action always runs
+ * regardless of whether validateAndTransform can match the data.
+ *
+ * This is needed for tool call patterns because:
+ * 1. `asOpaque` flags in schemas cause mismatches with $alias link schemas
+ * 2. Internal builtin nodes also have schema mismatches after the traversal change
+ * 3. "ref" type modules resolve to original registry entries, bypassing any
+ *    schema modifications on the recipe copy
+ *
+ * Setting argumentSchema to false is safe here because this only affects the
+ * ephemeral recipe copy created for this tool call invocation.
+ */
+function prepareRecipeForToolCall(recipe: Readonly<Recipe>): Recipe {
+  return {
+    ...recipe,
+    nodes: recipe.nodes.map((node) => ({
+      ...node,
+      module: {
+        ...node.module,
+        argumentSchema: false as unknown as JSONSchema,
+      },
+    })),
+  };
+}
+
+/**
  * Remove the injected `result` field from a JSON schema so tools don't
  * advertise it as an input parameter.
  */
@@ -1565,6 +1594,14 @@ function handleUpdateArgument(
 
 /**
  * Handles the invoke tool call (both pattern and handler execution).
+ *
+ * For patterns: starts the pattern via runtime.run(), then polls the result
+ * cell until a non-undefined value appears. This avoids relying on reactive
+ * sink subscriptions whose dependency tracking can miss changes through
+ * link chains.
+ *
+ * For handlers: uses the handler's onComplete callback for immediate
+ * resolution, falling back to polling the result cell.
  */
 async function handleInvoke(
   runtime: Runtime,
@@ -1572,6 +1609,7 @@ async function handleInvoke(
   resolved: ResolvedToolCall,
 ): Promise<{ type: string; value: any }> {
   const toolCall = resolved.call;
+  const startTime = Date.now();
 
   // Extract pattern/handler/params based on the resolved type
   let pattern: Readonly<Recipe> | undefined;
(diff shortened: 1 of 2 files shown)


bfollington:
## Root Cause Analysis

The unified traversal commit (`5d352fc09`) made `validateAndTransform` significantly stricter about schema matching. This broke LLM tool calls because of how `handleInvoke` uses `runtime.run()` to execute tool patterns.

### The chain of failure

1. **`handleInvoke` calls `runtime.run(tx, pattern, input, result)`** — this registers the tool pattern's nodes with the scheduler
2. **The scheduler runs each node's action**, which reads arguments via `inputsCell.asSchema(module.argumentSchema).withTx(tx).get()`
3. **`get()` calls `validateAndTransform`**, which now strictly validates that `$alias` link schemas match the expected `argumentSchema`
4. **The schemas don't match** — the `$alias` links carry schemas like `{type: "string", description: "..."}` while the compiled recipe's `argumentSchema` has `{type: "string", asOpaque: true}`. The `asOpaque` flag (added by the recipe compiler for reactivity) is absent from alias schemas.
5. **`validateAndTransform` returns `undefined`** → the action's `isValidArgument` check fails → **`fn(argument)` is never called** (only logged at debug level)
6. **This affects ALL nodes in the tool pattern** — not just the top-level function, but also internal builtin nodes (`computed`, `fetchData`, `ifElse`, etc.) which have their own schema mismatches. Additionally, "ref" type modules resolve to the original from the module registry, bypassing any schema modifications on the recipe copy.
7. **The tool function never produces a result** → the polling loop times out after 30s

### What this PR does

**`prepareRecipeForToolCall()`** — Creates an ephemeral copy of the recipe with `argumentSchema: false` on all node modules. When `argumentSchema === false`, the runner skips argument validation entirely and always calls the function. This is the only reliable fix because:
- Stripping just `asOpaque` fixes the top-level function but not internal nodes
- "ref" modules resolve to original registry entries, bypassing any schema fixes on the copy
- Internal builtin nodes have their own schema mismatches unrelated to `asOpaque`

**Result cell created without schema** — The result cell is created with `undefined` schema instead of `pattern.resultSchema`, so that reading the result doesn't go through strict schema validation either.

**Polling loop** — Replaced the previous `sink()`-based waiting (which also broke due to the traversal change not re-triggering reactive subscriptions through link chains) with a simple `idle()` + `get()` polling loop.

**New async tool test** — Exercises the reactive re-trigger path that the existing sync test didn't cover.

### What the ideal fix is

The real fix belongs in the unified traversal / runner, not in llm-dialog:

1. **`validateAndTransform` should be lenient about `asOpaque` and `description` differences** when following `$alias` links. The alias schema and the expected schema describe the same data — they just have different compiler-added annotations. The traverser should use the expected schema from the parent context, not reject on annotation mismatches.

2. **Alternatively, the recipe compiler should ensure `$alias` schemas match `argumentSchema`** — when `setupInternal` writes argument data via `diffAndUpdate`, the resulting `$alias` links should carry schemas consistent with the recipe's `argumentSchema` (including `asOpaque`).

Either fix would make `argumentSchema: false` unnecessary and restore proper type-checked execution for tool patterns.

packages/runner/src/builtins/llm-dialog.ts
```diff
@@ -1592,57 +1630,95 @@ async function handleInvoke(
...
-    // Create the result cell within the transaction context
+    // Create the result cell WITHOUT a schema. Tool call results are read via
+    // getRaw() and traverseAndSerialize() below, not via schema-validated get().
+    // Using a schema here causes validateAndTransform to reject $alias structures
+    // that don't carry matching schema annotations (asOpaque, etc.).
     result = runtime.getCell<any>(
       space,
       toolCall.id,
-      pattern ? pattern.resultSchema : undefined,
+      undefined,
```
ubik2: This should result in not doing any validation on the handler's state, which sort of makes sense, and will use a proxy.

PR #2784, opened 2026-02-11
Fix/wish required optional investigation

mathpirate:
Summary

  - Removes Required<> from WishFunction return type so wish result schemas correctly mark result, candidates, and error as optional — matching runtime
  reality where they start undefined until the wish resolves
  - Fixes OpaqueRef type to handle AnyBrandedCell | undefined (from optional property access) without collapsing branded cell intersections to never

  Problem

  WishFunction wraps its return in Required<WishState<T>>, forcing all wish result properties into the schema's required array. At runtime, wish results
  start empty — result is genuinely undefined until the wish resolves — so the runtime validation crashes on the initial state.

  Removing Required<> naively breaks the type system: optional WishState properties produce OpaqueRef<T | undefined>, and when T is a branded cell, the
  undefined in the union causes brand kind strings ("opaque" & "cell") to conflict, collapsing the intersection to never.

  Approach

  WishState redesign: Rather than using Required<> to paper over optional properties, WishState<T> now uses non-optional properties with explicit |
  undefined values (result: T | undefined instead of result?: T). This means:
  - The schema generator sees T | undefined → marks properties as not required (via isUnionWithUndefined)
  - The TypeScript mapped type doesn't add an extra | undefined from the ? modifier, so OpaqueRefInner produces cleaner types
  - Required<> is removed from WishFunction, schema.ts, and built-in.ts

  OpaqueRef fix: Added a branch for AnyBrandedCell<any> | undefined that strips undefined via Exclude<T, undefined>, preventing brand intersection collapse.

  Downstream fixes: Updated satisfies WishState<any> sites to provide all required properties, added missing return properties in wish.tsx/suggestion.tsx,
  fixed arbitrary-wish-example.tsx to render wishResult[UI] instead of the raw WishState, and widened $mentionable/$mentioned JSX prop types to accept
  CellLike<Charm[] | undefined>.

  Test plan

  - deno task test in packages/runner — 163 passed, 0 failed
  - deno task test in packages/schema-generator — 19 passed, 0 failed
  - deno check on 21 wish-related pattern files — all pass
  - Deploy and test wish-action-test pattern at runtime to verify wish results resolve correctly

---
## Summary by cubic
Switch WishState to explicit T | undefined fields and remove Required so schemas match runtime and prevent pre-resolution crashes. Harden OpaqueRef and ts-transformers to handle optional unions/arrays without breaking branded cells.

- **Bug Fixes**
  - WishState: result, candidates, and [UI] are now T | undefined; removed Required from api/schema/runner; wish built-in now sends result/candidates as undefined on errors; updated patterns to return missing fields and render wishResult[UI]; widened $mentionable/$mentioned to accept Charm[] | undefined.
  - OpaqueRef: handle AnyBrandedCell | undefined via NonNullable with a guard to avoid brand intersection collapse.
  - ts-transformers: fixed array type-argument narrowing and optional-element-access behavior; reverted OpaqueCell unwrapping so Cell<T | undefined> isn’t treated as optional; preserves correct optionality detection.

- **Refactors**
  - Centralized optionality logic in property-optionality.ts and used across schema-generator and transformers; simplified object-formatter.

<sup>Written for commit 792617759556ce2d774dadfd0999a0a6cb11d50a. Summary will update on new commits.</sup>


packages/html/src/jsx.d.ts
```diff
@@ -3616,8 +3616,8 @@ interface CTCodeEditorAttributes<T> extends CTHTMLAttributes<T> {
   "placeholder"?: string;
   "timingStrategy"?: string;
   "timingDelay"?: number;
-  "$mentionable"?: CellLike<Charm[]>;
```
mathpirate: this part feels dicey to me, esp since we'd probably have to do the same thing in other places?

packages/patterns/examples/arbitrary-wish-example.tsx
```diff
@@ -38,7 +38,7 @@ export default pattern<Record<string, never>>((_) => {
         <hr />
         <ct-textarea $value={wishText} />
         <hr />
-        {wishResult}
+        {wishResult[UI]}
```
seefeldb: This should not be necessary, does it break without it?

PR #2797, opened 2026-02-12
fix(home): guard against null favorites from unsynced cross-space refs

mathpirate:
## Summary
- Cross-space cell references in the favorites list resolve as `null` before sync completes
- When navigating back to the home space from another space, the `uniqueSpaces` computed block and favorite handlers would throw `TypeError: Cannot read properties of null (reading 'spaceDid')`
- Adds null guards in three locations: `addFavorite` handler, `removeFavorite` handler, and `uniqueSpaces` computed

## Root Cause
Favorites contain cell references to pieces in other spaces. On first evaluation after navigation, these cross-space references haven't synced yet and resolve as `null`. Once sync completes, the computed blocks re-evaluate with correct data.

## Test plan
- [ ] Deploy a pattern to a non-home space
- [ ] Favorite the pattern
- [ ] Navigate back to home space
- [ ] Verify no `spaceDid` null errors in console
- [ ] Verify favorites render correctly once cross-space sync completes

---
## Summary by cubic
Prevented Home crashes when favorites include cross-space references that haven’t synced yet by guarding against null entries. Added null checks in addFavorite, removeFavorite, and uniqueSpaces to skip unresolved favorites until sync completes.

<sup>Written for commit b2a9f0a9f86102fd94ab1a9c538d74f591f9c47d. Summary will update on new commits.</sup>

### DIFF
--- a/packages/patterns/system/home.tsx
+++ b/packages/patterns/system/home.tsx
@@ -118,7 +118,7 @@ const addFavorite = handler<
   { favorites: Writable<Favorite[]>; journal: Writable<JournalEntry[]> }
 >(({ piece, tag, spaceName }, { favorites, journal }) => {
   const current = favorites.get();
-  if (!current.some((f) => equals(f.cell, piece))) {
+  if (!current.some((f) => f && equals(f.cell, piece))) {
     // HACK(seefeld): Access internal API to get schema.
     // Once we sandbox, we need proper reflection
     //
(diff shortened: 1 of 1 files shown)


ubik2 (approved):
With any cross-space link, we can have missing data, since while the server's query system ensures you know all about the reachable cells in your own space, it doesn't know anything about cells in other spaces. The favorites list is just a common place for cross-space links to come up.

Looking deeper here, I'm not sure why we let favorites pass validation, though.
I added some code recently to replace non-matching array elements with null, but that should only happen when null is an acceptable value in the array, and in this case, that doesn't seem like it should be the case.

The change looks useful, and good to merge, but I'm adding a ticket to investigate why we're getting here (cross space links are only half supported, so it's not shocking -- just surprising).

mathpirate:
Awesome thank you!

mathpirate:
Mind linking the ticket here for posterity?

PR #2819, opened 2026-02-18
Refactor suggestion.tsx to use llmDialog

bfollington:
## Summary by cubic
Refactored suggestion.tsx to use llmDialog for follow-up conversations, so users can refine the initial suggestion with tool calls, clarifying questions, and immediate result presentation. Adds runtime support to seed and serialize messages and to selectively disable built-in tools (CT-1205).

- **New Features**
  - Follow-up chat in suggestion.tsx via llmDialog with tools: fetchAndRunPattern, listPatternIndex, askUser, finalResult.
  - Results are auto-presented; users can answer questions (ct-question) and send refinements (ct-message-input).
  - Dialog seeds from generateObject’s conversation and prefers dialog-presented results when rendering.

- **Refactors**
  - llmDialog: optional built-in tools via builtinTools; extraParams stripped from tool schemas; seeded messages serialized with traverseAndSerialize; finalResult outputs wrapped for SDK compatibility.
  - Handler tool execution bypasses .send() to preserve Cell references when invoked.
  - generateObject records and exposes accumulated messages; API types updated (BuiltInLLMParams.builtinTools, BuiltInLLMGenerateObjectState.messages).
  - Added CT-1205 investigation doc to track cell/link serialization issues.

<sup>Written for commit 6957c87805121bd71378ee84b7b06e9bc1b42770. Summary will update on new commits.</sup>

### DIFF
--- a/packages/runner/src/builtins/llm-dialog.ts
+++ b/packages/runner/src/builtins/llm-dialog.ts
@@ -41,6 +41,7 @@ import {
   isCellResultForDereferencing,
 } from "../query-result-proxy.ts";
 import { ContextualFlowControl } from "../cfc.ts";
+import { resolveLink } from "../link-resolution.ts";
 
 // Avoid importing from @commontools/piece to prevent circular deps in tests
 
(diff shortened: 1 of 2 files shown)


packages/runner/src/builtins/llm-dialog.ts
```diff
@@ -1609,14 +1637,22 @@ async function handleInvoke(
     if (pattern) {
       runtime.run(tx, pattern, { ...input, ...extraParams }, result);
     } else if (handler) {
-      handler.withTx(tx).send({
-        ...input,
-        result, // doesn't HAVE to be used, but can be
-      }, (completedTx: IExtendedStorageTransaction) => {
+      // Bypass handler.send() which calls convertCellsToLinks(), destroying
+      // live Cell references from traverseAndCellify. Instead, resolve the
```
seefeldb: That seems odd, convertCellsToLinks makes links where there are cells, specifically to keep them live.

What the code below does is avoid the roundtrip to serialized event with links and back out vs this just passes the cells through and thus ignores the schema on the handler. Since we're basing this on the same schema, it happens to work, but feels a little shaky.

Seems to work, but if we could quickly check whether the problem persisted that would be great, and if it does file an issue to dig into this. It's of the "this should have worked" class of bug.

PR #2820, opened 2026-02-18
fix(runner): wait for computeds to settle before reading fetchData inputs (CT-1246)

jkomoros:
## Summary

- **Fix**: `fetchData` options from `computed()` were `undefined` on the first fetch because `tryClaimMutex` read inputs before the scheduler evaluated pending computeds
- **Change**: Add `await runtime.idle()` in `tryClaimMutex` before reading inputs, following the same pattern used throughout all builtins for async/reactive synchronization
- **Test**: Add regression test verifying computed options (headers) are present on the first fetch call

## Root Cause

When `fetchData()` receives `options` from a `computed()`, the action fires synchronously before the computed has been evaluated by the scheduler. The `url` field was accidentally protected by the `if (!url) return` guard, but `options` had no such guard — causing the first fetch to go out without headers.

## Why `runtime.idle()`

This is the scheduler's official synchronization primitive, already used in every async builtin (fetch-data, stream-data, llm) before writing results. The fix extends the same invariant to the input-reading path: **before touching reactive state from async code, wait for the scheduler to settle.**

Fixes CT-1246

## Test plan

- [x] Existing `fetch-data-mutex.test.ts` tests pass (7/7)
- [x] New CT-1246 regression test passes — verifies computed headers are present on every fetch call
- [ ] Manual test with a pattern using `computed()` options for fetchData

---
## Summary by cubic
Fixes a race where fetchData read computed options before they settled and a stale proxy could drop nested headers, so computed headers appear on the first fetch (CT-1246). Adds a regression test and cleans up input typing and snapshot handling.

- **Bug Fixes**
  - Await runtime.idle() before editWithRetry in tryClaimMutex so computeds settle before reading inputs.
  - Snapshot inputs inside the transaction via cell.asSchema(fetchDataInputSchema).get() using a snapshotInputs callback to materialize nested fields; move body stringification there and remove duplicate logic in startFetch.

<sup>Written for commit 25989811e66d34542de301576d381f9a1aabf696. Summary will update on new commits.</sup>

### DIFF
--- a/packages/runner/test/fetch-data-mutex.test.ts
+++ b/packages/runner/test/fetch-data-mutex.test.ts
@@ -15,6 +15,7 @@ describe("fetch-data mutex mechanism", () => {
   let runtime: Runtime;
   let tx: IExtendedStorageTransaction;
   let pattern: ReturnType<typeof createBuilder>["commontools"]["pattern"];
+  let computed: ReturnType<typeof createBuilder>["commontools"]["computed"];
   let byRef: ReturnType<typeof createBuilder>["commontools"]["byRef"];
   let originalFetch: typeof globalThis.fetch;
   let fetchCalls: Array<{ url: string; init?: RequestInit }>;
@@ -29,6 +30,7 @@ describe("fetch-data mutex mechanism", () => {
 
     const { commontools } = createBuilder();
     pattern = commontools.pattern;
+    computed = commontools.computed;
     byRef = commontools.byRef;
 
     // Set up pattern environment with a mock base URL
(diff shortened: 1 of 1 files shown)


packages/runner/src/builtins/fetch-data.ts
```diff
@@ -3,14 +3,39 @@ import { type Action } from "../scheduler.ts";
...
+      properties: {
+        body: {},
+        method: { type: "string" },
+        headers: {
+          type: "object",
+          additionalProperties: { type: "string" },
+        },
+      },
+    },
+    result: {},
```
seefeldb: remove `result` here, i'm pretty sure it is unused

packages/runner/src/builtins/fetch-data.ts
```diff
@@ -194,6 +219,30 @@ export function fetchData(
         error,
         internal,
         newRequestId,
+        // Materialize inputs via the schema system and preprocess body.
+        // asSchema().get() returns a plain snapshot with nested properties
+        // (like options.headers) fully resolved, safe to use after commit.
+        (cell) => {
+          const snapshot = cell.asSchema(fetchDataInputSchema).get() as {
```
seefeldb: the `as { ... }` is not necessary, the type is automatically inferred from the static json schema!

PR #2874, opened 2026-02-24
Make Default<T,V> fields required inside pattern bodies

ubik2:
Problem                                                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                              
When a pattern input has a Default<T, V> field, the pattern body should be able to treat it as non-optional — the runtime guarantees the default is populated before the pattern runs. Previously Default<T,V> was just a type alias for T, so optional fields with defaults were still typed as optional inside the pattern body, forcing unnecessary
null-checks.

Solution

Adds a compile-time brand to Default<T,V> via a new DEFAULT_MARKER unique symbol. This enables a RequireDefaults<T> mapped type that, when applied to a pattern's input, promotes any Default<>-branded fields from optional to required — with the brand stripped — while leaving all other fields unchanged.

packages/api/index.ts
- Default<T,V> now expands to a branded union (T & { [DEFAULT_MARKER]: T }) | T; at runtime it's still just T
- New type utilities: HasDefaultBrand, IsDefaultField, StripDefaultBrand, RequireDefaults
- PatternFunction overloads updated to wrap the input with RequireDefaults<T>
- IReadable.get() and .sample() return StripDefaultBrand<T> so callers see the plain type

packages/runner/
- Applied RequireDefaults<T> to all pattern() overloads and patternTool
- CellImpl.get() / .sample() updated to return StripDefaultBrand<T>

packages/schema-generator/
- common-tools-formatter: fallback detection of Default<T> via aliasSymbol when typeToTypeNode has already expanded the alias
- intersection-formatter: filters out brand-only and empty-object intersection parts (produced by RequireDefaults<T> and Default<T,V>'s brand member) so schema generation isn't confused by them

packages/ts-transformers/
- type-inference.ts: handles Default<T[]> brand union when detecting array types, so inferArrayElementType / hasArrayTypeArgument correctly unwrap Default-wrapped arrays

packages/js-compiler/
- Added DEFAULT_MARKER to the known-exported-symbols suppression list to silence false-positive TypeScript declaration diagnostics

Pattern updates (consequences of the stricter types)
- cell-link.tsx, notebook.tsx, notes-import-export.tsx, chat-note.tsx, email-task-engine.tsx: updated allPieces fields to Default<NotePiece[], []> so they're treated as required in pattern bodies
- google-calendar-importer.tsx: settings made explicitly required in the input type signature
- notes-import-export.tsx: added null-guards for the now-optional allPieces parameter

<!-- This is an auto-generated description by cubic. -->
---
## Summary by cubic
Make Default<T, V> fields act as required inside pattern bodies, so you don’t need null-checks for inputs with defaults. Implemented via a compile-time brand and a RequireDefaults<T> mapped type; runtime behavior is unchanged.

- **New Features**
  - Added DEFAULT_MARKER brand and RequireDefaults<T> to promote Default<> fields from optional to required in pattern inputs; brand is stripped in bodies. Updated pattern(), PatternFunction, and patternTool; IReadable.get()/sample() now return StripDefaultBrand<T>.
  - Schema generator: detect Default<> via aliasSymbol even when the alias is expanded, preserve default values in schemas, and filter out brand-only/empty intersection parts for cleaner output.
  - TS transformers: unwrap Default<T[]> when inferring array element types and when checking array generics; updated fixtures to reflect required booleans/inputs.
  - JS compiler diagnostics: suppress false positives for DEFAULT_MARKER across TS4053/4054/4055 message variants.
  - Patterns: marked allPieces as Default<NotePiece[], []> in chat-note, email-task-engine, notebook, and notes-import-export; made Google Calendar settings required; added null-guards in notes-import-export; bash and omnibox now omit undefined fields from request payloads/params.

- **Migration**
  - Annotate optional inputs that have defaults with Default<T, V>. Callers can still omit them; inside pattern bodies they are non-optional.
  - No runtime changes required.

<sup>Written for commit 1e9ced4c01b9fa8e9122964d3f3036631f64bcab. Summary will update on new commits.</sup>

<!-- End of auto-generated description by cubic. -->

### DIFF
--- a/packages/schema-generator/src/formatters/common-tools-formatter.ts
+++ b/packages/schema-generator/src/formatters/common-tools-formatter.ts
@@ -119,8 +119,9 @@ export class CommonToolsFormatter implements TypeFormatter {
     }
 
     // Fallback: handle Default<T> detected via aliasSymbol when no type node is available.
-    // When typeToTypeNode expands Default<T,V>, the node no longer says "Default" but
-    // the type object still carries aliasSymbol. Extract T from aliasTypeArguments[0].
+    // When typeToTypeNode expands Default<T,V), the node no longer says "Default" but
+    // the type object still carries aliasSymbol. Extract T from aliasTypeArguments[0]
+    // and V from aliasTypeArguments[1] so the default value is preserved in the schema.
     const typeWithAlias = type as TypeWithInternals;
     if (
       typeWithAlias.aliasSymbol?.name === "Default" &&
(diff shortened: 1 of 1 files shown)


ubik2 (commented):
Merging into non-main branch, so should be fine.
At the point of merge, most tests pass, but the default-app flow has an issue with creating a new note (related to frame mismatch).

PR #2879, opened 2026-02-24
fix(runtime): prevent recursion depth crash on nested piece subscriptions

mathpirate:
## Summary

- Fixes `Maximum recursion depth of 100 exceeded` crash when a space contains 3+ levels of nested notebook pieces
- Root cause: `ct-cell-link` subscribed to piece cells without a schema, causing `handleCellSubscribe` to walk the entire piece output graph including nested VNode trees
- Adds a repro pattern (`packages/patterns/gideon-tests/notebook-nest-bug/main.tsx`)

## Root Cause

When `ct-cell-link` receives a `$cell` binding from a VNode, the Cell arrives with `schema: {}`. This happens because `rendererVDOMSchema` declares all props as `additionalProperties: { asCell: true }` — when `TransformObjectCreator` strips `asCell`, the remaining schema is `{}`.

When `ct-cell-link` subscribes to this cell (to display `$NAME`), `handleCellSubscribe` → `cell.sink()` → `validateAndTransform` sees no meaningful schema → returns a raw `createQueryResultProxy` over the entire piece output. `convertCellsToLinks` with `doNotConvertCellResults: true` then walks the proxy via `Object.entries`, traversing the full VNode tree including cross-piece references, exceeding `MAX_RECURSION_DEPTH` (100).

## Changes

1. **`ct-cell-link`** (`packages/ui/src/v2/components/ct-cell-link/ct-cell-link.ts`): Subscribe with `.asSchema({ $NAME: { type: "string" } })` — a minimal schema that only resolves `$NAME`. This gives `validateAndTransform` a schema to work with, preventing the raw proxy fallback.

2. **`handleCellSubscribe` workaround** (`packages/runtime-client/backends/runtime-processor.ts`): When a cell subscription has no schema and produces a CellResult proxy, convert it to a link immediately instead of walking into it. Logs a `console.warn` so we can spot remaining empty-schema cases.

3. **Debug instrumentation cleanup** (`packages/runner/src/query-result-proxy.ts`): Remove leftover depth tracing from investigation.

## Test plan

- [x] Manual: deployed repro pattern, created 3-level nested notebooks, navigated space home — no recursion errors, no `console.warn` messages
- [x] Manual: clicked through Notebook A → A-Child → A-Grandchild and back — all render correctly
- [x] CI: existing tests pass (Package Integration Tests, Pattern Integration Tests)

## Follow-up

- CT-1271: Disallow `pattern<Input>()` (input-only type param) which produces `resultSchema: true`
- Berni: modifying VDOM schema separately
- Robin: implementing `{type: unknown}` support for VDOM prop schemas

Fixes CT-1261

### DIFF
--- a/packages/runtime-client/backends/runtime-processor.ts
+++ b/packages/runtime-client/backends/runtime-processor.ts
@@ -424,7 +424,7 @@ export class RuntimeProcessor {
       // MAX_RECURSION_DEPTH when VNode trees reference other pieces.
       // Only apply this for cells without a meaningful schema — cells with
       // schemas produce resolved plain objects from validateAndTransform.
-      // TODO(CT-1271): Once all pattern callsites produce real result schemas
+      // TODO(CT-1273): Once all pattern callsites produce real result schemas
       // (CTS now errors on `any`/`unknown` inference), remove this guard and
       // replace with a console.error — hitting this path would then indicate
       // a bug rather than a known limitation.


seefeldb (approved):
Maybe add a TODO to eventually remove the no schema condition and instead log something?

PR #2894, opened 2026-02-26
Add ct-svg web component and SVG diagram pattern

bfollington:
## Summary

- Adds a `ct-svg` Lit web component that renders SVG markup from a string prop using `unsafeHTML`, following the same architecture as `ct-markdown`
- Adds an `svg-diagram` suggestable pattern that uses `generateText` to produce SVG diagrams rendered via the new component
- Registers `ct-svg` in JSX type definitions and the UI barrel export

## Test plan

- [x] Deploy `svg-diagram.tsx` with `ct dev` and test with a topic like "the water cycle"
- [x] Verify SVG renders visually in the browser via the `ct-svg` component
- [x] Verify `deno check` passes on UI and patterns packages


bfollington:
> Do we do anything like DOMPurify for the unsafeHTML blocks? It's not the cfc solution, but would provide some safety for llm generated svg content. Overall, looks like a cool feature.

Good call, I have copied from the approach in `ct-markdown` which uses a custom list of permitted elements to effectively do the DOMPurify solution.

PR #2899, opened 2026-02-26
fix(vdom): stop deep traversal of vdom props, use Cell<Props> per-prop subscriptions

seefeldb:
## Summary

Prevents the query system from deeply traversing vdom property values (which is unnecessary and expensive), while maintaining correct client-side reactivity, event handling, and bidirectional bindings.

### Problem
The `rendererVDOMSchema` and `vnodeSchema` used `additionalProperties: { asCell: true }` on props, causing the query engine to descend into every prop value — creating per-prop Cell subscriptions server-side for data that only needs shallow resolution. This is wasteful and can cause unexpected behavior when prop values contain nested references.

### Solution
- **Schema change**: Constrain `additionalProperties` to primitive types and non-descending object/array schemas (with empty `properties: {}` and `items: { type: "null" }`). Move `asCell: true` to the `props` field itself, making props a single `Cell<Props>` boundary. Add an explicit `style` property since it's the one object prop that needs deep resolution client-side.
- **Reconciler change**: New `bindCellProps` method with a mixed per-prop strategy:
  - **Primitive props** (string, number, boolean, null): set directly from the resolved `Cell<Props>` value, with deduplication
  - **Event props** (`onclick`, etc.): resolved via `.key().resolveAsCell()` for stream/handler registration
  - **Binding props** (`$value`, etc.): resolved via `.key().resolveAsCell()` for cell reference
  - **Object/array props** (style, etc.): per-prop sink via `.key().asSchema(true)` for deep value resolution client-side

### Files changed
- `packages/runner/src/schemas.ts` — stop deep traversal of props; move `asCell` to props field level
- `packages/html/src/worker/reconciler.ts` — new `bindCellProps` and `removeSingleProp` methods; deduplicated prop removal logic via shared helper
- `packages/html/test/worker-reconciler-cell-props.test.ts` — 12 new test cases covering all Cell<Props> scenarios

## Test plan
- [x] 12 new Cell<Props> unit tests (primitives, updates, add/remove, objects, arrays, events, bindings, same-cell optimization, clear/re-emit, dedup, mixed types)
- [x] All existing html tests pass
- [x] All runner tests pass
- [x] All integration tests pass

### DIFF
--- a/packages/html/src/worker/reconciler.ts
+++ b/packages/html/src/worker/reconciler.ts
@@ -819,7 +819,16 @@ export class WorkerReconciler {
       for (const [key, value] of Object.entries(props)) {
         if (isEventProp(key)) {
           // Event prop - resolve target via Cell navigation
-          const resolvedTarget = propsCell.key(key).resolveAsCell();
+          let resolvedTarget: Cell<unknown>;
+          try {
+            resolvedTarget = propsCell.key(key).resolveAsCell();
+          } catch (e) {
+            logger.error(
+              "resolveAsCell failed for event prop",
+              () => ({ nodeId: state.nodeId, key, error: e }),
+            );
+            continue;
+          }
           const existingState = state.propSubscriptions.get(key);
 
           // Skip if same target Cell
@@ -846,7 +855,7 @@ export class WorkerReconciler {
           if (existingState) existingState.cancel();
 
           const handlerId = ctx.registerHandler((event: unknown) =>
-            (resolvedTarget as Cell<unknown>).send(event)
+            resolvedTarget.send(event)
           );
           state.eventHandlers.set(eventType, handlerId);
           this.queueOps([{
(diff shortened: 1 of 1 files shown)


packages/html/src/worker/reconciler.ts
```diff
@@ -838,6 +813,227 @@ export class WorkerReconciler {
...
+            handlerId,
+          }]);
+          state.propSubscriptions.set(key, {
+            cell: resolvedTarget as Cell<unknown>,
+            cancel: () => {},
+            currentValue: resolvedTarget,
+          });
+        } else if (isBindingProp(key)) {
+          // Binding prop - resolve target Cell via navigation
+          const resolvedTarget = propsCell.key(key).resolveAsCell();
```
mathpirate: consider wrapping in a try/catch since resolveAsCell can throw on link cycles and iteration limits

PR #2904, opened 2026-02-26
perf: traverse optimizations — anyOf fast-reject, hash-based MapSet, schema memoization, health dashboard

seefeldb:
### anyOf branch handling (Layer 1 — fast-reject)
- **`canBranchMatch()` fast-reject**: cheap pre-check eliminates anyOf branches by type mismatch, `const`/`enum` mismatch, or missing `required` properties before full traversal. Never rejects `asCell`/`asStream` branches. Bails out conservatively for link objects.
- **`mergeAnyOfBranchSchemas()` property merge** (implemented but **disabled**): for object values with 2+ surviving branches, merges all branch schemas into a single object schema and traverses once instead of N times. Disabled due to link-resolution regressions; helper and tests retained for future use.
- New diagnostic counters: `anyOfFastRejects`, `anyOfPropertyMerges`
- Reactivity invariant tests confirming fast-reject does not break `schemaTracker` link tracking

### Hash-based MapSet & schema traversal memoization
- **`stableStringify` + WeakMap cache**: O(1) amortized hash lookups in `MapSet` instead of O(n) `deepEqual` scans. Identity cache via WeakMap for object keys.
- **`traverseWithSchema` memoization**: cache keyed by `docId|path|stableStringify(schema)`, shared across `loadFactsForDoc` calls via `SchemaMemo` map.
- **`CompoundCycleTracker`**: replaced `deepEqual`-based cycle detection with hash-based lookups.
- **Schema merge interning**: `mergeSchemaOption`, `combineSchemas`, `mergeSchemaFlags` results interned for identity-cached hashing.

### Shared ServerObjectManager across incremental updates
- `processIncrementalUpdate` now creates one shared `ServerObjectManager` and `SchemaMemo` passed to all `evaluateDocumentLinks` calls, so SQLite read cache persists across affected doc evaluations.
- `evaluateDocumentLinks` accepts optional `sharedManager` and `sharedMemo` parameters.

### Health dashboard (`/api/health/dash`)
- Full HTML dashboard at `/api/health/dash` with auto-refresh every 5 seconds
- Summary cards: Server Uptime, Total Logs, Timed Operations, Slow Queries
- SVG charts: Slow Query Timeline (scatter), Operation Distribution (horizontal bars), Latency Histogram
- Timing Stats table with count, total ms, avg, p50, p95, max
- Slow Queries table with columns: Time, Elapsed, Operation, Space, Docs, Sel, Facts, Loaded, SQLite, SQL ms, Cache, Tracker, Memo
- Click-to-copy on Space, Docs, and Selector cells; tooltips on all column headers
- `SelectSchemaStats` interface threaded through `querySchemaWithTracker` for detailed slow query diagnostics

## Performance results (local testing)
- With 3 deployed patterns + active interaction: avg graph-query/subscribe **0.41ms**, max **23.5ms**
- 9,552 traversals at avg **0.21ms**, 1,438 subscriptions at avg **0.41ms**
- Zero slow queries recorded (all under 100ms threshold)

## Test plan
- [x] 19 unit tests for `canBranchMatch` (type/const/enum/required/asCell/boolean)
- [x] 9 unit tests for `mergeAnyOfBranchSchemas` (disjoint/overlapping/required/$defs)
- [x] 5 integration tests (discriminated union, property merge, link discovery, primitives, mixed)
- [x] 6 reactivity invariant tests (root tracking, link tracking through surviving branches, discriminator flip, nested chains)
- [x] All existing traverse tests pass (58 tests, no regressions)
- [x] Local end-to-end testing with deployed patterns (todo-list, counter, notes)

### DIFF
--- a/packages/runner/test/traverse.test.ts
+++ b/packages/runner/test/traverse.test.ts
@@ -1374,6 +1374,43 @@ describe("canBranchMatch", () => {
     expect(canBranchMatch({ type: ["string", "number"] }, 42)).toBe(true);
     expect(canBranchMatch({ type: ["string", "number"] }, true)).toBe(false);
   });
+
+  it("checks required properties when type is an array including 'object'", () => {
+    // type: ["object", "null"] should still check required properties
+    expect(
+      canBranchMatch(
+        { type: ["object", "null"], required: ["name"] },
+        { name: "Alice" },
+      ),
+    ).toBe(true);
+    expect(
+      canBranchMatch(
+        { type: ["object", "null"], required: ["name"] },
+        { age: 30 },
+      ),
+    ).toBe(false);
+    // type array without "object" should skip required check
+    expect(
+      canBranchMatch(
+        { type: ["string", "number"], required: ["name"] },
+        { age: 30 },
+      ),
+    ).toBe(false); // rejected by type mismatch, not required
+  });
+
+  it("conservatively accepts when value is a cell link object", () => {
+    const linkValue = { "/": { [LINK_V1_TAG]: { id: "of:target", path: [] } } };
+    // Even if the branch type doesn't match the link's shape, links get
+    // resolved during traversal so we must not reject.
+    expect(canBranchMatch({ type: "string" }, linkValue)).toBe(true);
+    expect(canBranchMatch({ type: "number" }, linkValue)).toBe(true);
+    expect(
+      canBranchMatch(
+        { type: "object", required: ["missing"] },
+        linkValue,
+      ),
+    ).toBe(true);
+  });
 });
 
 describe("mergeAnyOfBranchSchemas", () => {


packages/runner/src/traverse.ts
```diff
@@ -2730,6 +2749,246 @@ function mergeSchemaOption(
...
+    if (actualType !== null) {
+      const schemaTypes = Array.isArray(resolved.type)
+        ? resolved.type
+        : [resolved.type];
+      if (!schemaTypes.includes(actualType)) return false;
+    }
+  }
+
+  // Root const mismatch
+  if ("const" in resolved) {
```
ubik2: we don't actually do these. I'd probably remove this clause, since we should probably allow a doc value with links inside it (not at the top level) which aren't deepEqual, but which might be after link following.

packages/runner/src/traverse.ts
```diff
@@ -2730,6 +2749,246 @@ function mergeSchemaOption(
...
+    if (!deepEqual(resolved.const, value)) return false;
+  }
+
+  // Root enum mismatch
+  if (Array.isArray(resolved.enum)) {
+    if (!resolved.enum.some((e: unknown) => deepEqual(e, value))) return false;
```
ubik2: same issue here. we could just restrict these two checks to only apply to non-records, but we've already returned on line 2777 for those.

packages/runner/src/traverse.ts
```diff
@@ -2730,6 +2749,246 @@ function mergeSchemaOption(
...
+  }
+
+  // For plain object values, check property-level discriminators and
+  // required fields.
+  if (
+    isRecord(value) &&
+    (resolved.type === "object" || resolved.type === undefined)
```
ubik2: resolved.type can be an array, so 
```(
  (resolved.type === undefined) || (typeof resolved.type === "string" && resolved.type === "object") || (Array.isArray(resolved.type) && resolved.type.includes("object"))
)
```
Also, "unknown" should match once I add it.

PR #2905, opened 2026-02-26
refactor(ui): add prose mode to ct-code-editor

robdodson:
## Summary

Add a prose mode to `ct-code-editor` that renders markdown with hidden syntax markers (headings, bold, italic, links, etc.) and reveals them when the cursor is on the active line. Replace CodeMirror's `basicSetup` bundle with explicit, mode-conditional extensions so prose mode only loads what it needs.

## Why

ct-code-editor was importing the [basicSetup package](https://codemirror.net/docs/ref/#codemirror.basicSetup) from CodeMirror. basicSetup is a bundle of plugins that are designed for coding experiences. These include syntax highlighting, line numbers and folding, highlighting matching words and the active line, etc.

We don't need these when writing pure Markdown, and they inject a lot of styles and DOM that we would have to override.

Instead, when the component is in `mode=prose` it will import just the CodeMirror packages it needs, and when it is in `mode=code` it will import all of the coding specific packages.

## Changes in this PR

- **Prose markdown rendering** — New `features/prose-markdown.ts` extracts ~685 lines of decoration logic (heading styles, inline syntax hiding, list bullet widgets, link rendering, task checkboxes, horizontal rules) into a standalone module with an exported `buildProseDecorations()` function
- **Explicit CodeMirror extensions** — Replace `basicSetup` with `_getSetupExtensions()` that conditionally includes line numbers, bracket matching, fold gutters, syntax highlighting, etc. only in code mode
- **Mode compartment** — Add `_setupComp`, `_modeComp`, and `_proseMarkdownComp` compartments for runtime mode switching between code and prose
- **Prose styles** — Add prose-specific CSS in `styles.ts` for heading sizes, inline formatting, list markers, blockquotes, links, and code blocks
- **Tests** — 15 test cases for prose decoration logic covering headings, bold, italic, strikethrough, lists, links, inline code, blockquotes, and unfocused editor behavior

## Known issues

- The following entities don't render correctly yet:
  - Tables
  - Fenced code blocks — they render, but they don't look great
  - Images
  - Inline html elements, e.g. `<mark>`, `<details>`, `<summary>`
- The styling doesn't quite match `note-md.tsx`, which is what displays the note when you click **View**

### DIFF
--- a/packages/ui/src/v2/components/ct-code-editor/features/prose-markdown.ts
+++ b/packages/ui/src/v2/components/ct-code-editor/features/prose-markdown.ts
@@ -74,6 +74,7 @@ class BulletWidget extends WidgetType {
   }
 }
 
+// Singleton — safe to reuse because eq() always returns true (no instance state).
 const bulletWidget = new BulletWidget();
 
 /**
@@ -131,6 +132,7 @@ class HorizontalRuleWidget extends WidgetType {
   }
 }
 
+// Singleton — safe to reuse because eq() always returns true (no instance state).
 const hrWidget = new HorizontalRuleWidget();
 
 /**
--- a/packages/ui/src/v2/components/ct-code-editor/styles.ts
+++ b/packages/ui/src/v2/components/ct-code-editor/styles.ts
@@ -79,75 +79,74 @@ export const styles = css`
     box-shadow: none;
   }
 
-  /* Prose markdown heading styles */
-  .cm-prose-h1 {
+  /* Prose markdown styles - scoped to prose mode */
+  :host([mode="prose"]) .cm-prose-h1 {
     font-size: 1.75em;
     font-weight: 700;
     line-height: 1.3;
   }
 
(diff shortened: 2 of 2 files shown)


packages/ui/src/v2/components/ct-code-editor/features/prose-markdown.ts
```diff
@@ -0,0 +1,687 @@
...
+      decorations: DecorationSet;
+
+      constructor(view: EditorView) {
+        this.decorations = Decoration.set(
+          buildProseDecorations(view.state, view.hasFocus),
+          true,
+        );
+      }
+
+      // TODO(@robdodson): For large documents, optimize to only decorate
```
mathpirate: maybe take a linear issue for yourself to follow up on this in case it becomes a performance issue with huge docs? idk if you have linear access yet, lmk if i should file one

mathpirate (approved):
looks great!

PR #2923, opened 2026-02-28
Interval-based wish("#now/N") with coarsened timestamps

jkomoros:
## Summary

- Add reactive time source for patterns via `wish({ query: "#now/60000" })` that ticks on wall-clock-aligned intervals
- Coarsen all `#now` timestamps (including one-shot) to remove sub-second precision as defense-in-depth for when SES sandboxing is enforced
- Clamp minimum interval to 1000ms to prevent high-frequency timers
- Handle race conditions: generation counter prevents stale writes on reactive re-runs, `editWithRetry` is awaited to prevent unbounded concurrency

## API

```tsx
// One-shot (existing, now coarsened to 1s)
const createdAt = wish<number>({ query: "#now" });

// Interval — ticks every 5s, values quantized to 5s boundaries
const now = wish<number>({ query: "#now/5000" });
const timeAgo = computed(() => {
  return `${Math.floor((now.result - msg.sentAt) / 60000)} min ago`;
});
```

Uses the path slot (`#now/60000`) following the `#favorites/searchTerm` precedent. No changes to `WishParams`, `TARGET_SCHEMA`, or `parseWishTarget`.

## Design

- `handleIntervalNow()` is called before the resolution pipeline — interval `#now` is a timer subscription, not a cell resolution
- Timer ticks aligned to wall-clock boundaries via `setTimeout` with drift correction
- Generation counter skips stale ticks after reactive re-runs
- `editWithRetry` awaited to prevent concurrent in-flight writes
- Timer cleanup via `addCancel`, registered once per wish instance

## Test plan

- [x] `coarsenTimestamp` arithmetic verified via integration tests
- [x] One-shot `#now` coarsened to 1s boundary
- [x] `#now/5000` returns coarsened timestamp divisible by 5000
- [x] `#now/1` clamped to minimum 1000ms
- [x] `#now/60000` coarsened to 60s boundary
- [x] `#now/abc`, `#now/0`, `#now/-1`, `#now/Infinity` all rejected
- [x] `#now/5000/extra` rejected (too many path segments)
- [x] Timer cleanup on `runtime.runner.stop()`
- [x] Timer actually ticks and updates value (polling test)
- [x] E2E verified in browser via Playwright — timestamps update every 5s, both divisible by 5000
- [x] Existing wish tests pass (no regression)

Closes CT-1163

### DIFF
--- a/docs/common/conventions/wish.md
+++ b/docs/common/conventions/wish.md
@@ -132,7 +132,7 @@ This ensures the wish is established once. Conditional logic belongs in how you
 
 | Target          | Description                                          |
 |-----------------|------------------------------------------------------|
-| `#now`          | Current timestamp (one-shot, milliseconds)           |
+| `#now`          | Current timestamp (one-shot, 1s resolution)          |
 | `#now/N`        | Reactive timestamp that ticks every N ms (min 1000)  |
 | `#favorites`    | User's favorites list (from home space)              |
 | `#default`      | Default pattern of the current space                 |
@@ -143,10 +143,10 @@ This ensures the wish is established once. Conditional logic belongs in how you
 | `/path/to/prop` | Nested property of the space cell                    |
 
 ```tsx
-// One-shot timestamp (captured once)
-const createdAt = wish({ query: "#now" });
+// One-shot: captures current time once, never updates
+const createdAt = wish<number>({ query: "#now" });
 
-// Reactive timestamp (updates every 60s)
+// Reactive: updates every 60s, re-triggers downstream computed()
 const now = wish<number>({ query: "#now/60000" });
 const timeAgo = computed(() => {
   const ms = now.result - message.sentAt;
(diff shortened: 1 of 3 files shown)


seefeldb (commented):
while it writes a cell, i don't think this coordinates across multiple instances, it'll have each instance trigger at interval, and so #instances times per interval.

i think instead you want to
- read from the cell and if there isn't anything
 - write into it start time, interval, last triggered
 - otherwise, read from it && if we're due a trigger (e.g. after a reload), issue it, fire and forget in a new tx
- schedule writes at that interval, fire and forget -- one of the open tabs will then always be the first to write at that interval

main issue is that if we have skewed clocks the one most in the future will lead, but i guess that's ok for now

jkomoros:
Rebased onto latest main (clean — no conflicts). All wish tests pass locally. The earlier review feedback about cross-tab coordination is addressed in `Coordinate #now/N interval wishes across tabs with shared cell` and `Expose only lastTriggered as result for interval #now`. Ready for another look when you have a moment, @seefeldb.

PR #2925, opened 2026-03-01
fix(ios): add Symbol.dispose polyfill to web worker entry point

robdodson:
## Summary
The web worker bundle was missing the core-js explicit resource management polyfill, causing Symbol.dispose to be undefined on iOS Safari. The main shell entry point had the polyfill, but since the worker is a separate esbuild entry point, it needs its own import.


seefeldb (approved):
Thanks!!

PR #2942, opened 2026-03-03
feat(runner): identity-based reconciliation in map builtin

mpsalisbury:
## Summary

The map builtin previously tracked elements by positional index. When elements shifted positions (e.g., a mid-list insert), every element from the change point onward triggered unnecessary recomputation — O(n) work for a single insert.

This PR adds **identity-based reconciliation** using each element cell's normalized link address (`getAsNormalizedFullLink()`). Cell-link elements get stable identity across position changes (their normalized link points to the target entity, not the array slot). Inline values use positional identity (their normalized link includes the array index). Existing pattern runs are reused when cell-link elements shift positions, reducing inserts from O(n) to O(1) recomputation for the common case.

## Problem

Given a map over `[A, B, C]` where A, B, C are cell links, inserting `X` at index 1 to produce `[A, X, B, C]`:

- **Before**: Indices 1, 2, 3 all trigger recomputation (3 recomputes) because the positional bindings shift — index 1 was B, now it's X; index 2 was C, now it's B; index 3 is new.
- **After**: Only 1 recompute (the new element X). Elements A, B, C are recognized by their normalized link address and their existing pattern runs are reused.

For **inline values** (non-cell-link elements), identity is positional — the normalized link includes the array index. Shifted inline values get new runs. This is the expected trade-off; the important optimization is for cell links.

### Identity Tracking

Elements are keyed by their normalized link address with occurrence counting for duplicates:

```typescript
const { space: s, id, type, path } = list[i].getAsNormalizedFullLink();
const dedupKey = JSON.stringify([s, id, type, path]);
const occurrence = keyCounts.get(dedupKey) ?? 0;
keyCounts.set(dedupKey, occurrence + 1);
const elementKey = JSON.stringify([s, id, type, path, occurrence]);
```

- **Cell links**: The normalized link resolves to the target entity, so the key is stable across position changes.
- **Inline values**: The normalized link includes the positional index, so identity = position.
- **Duplicates**: Occurrence counting ensures that `[A, B, A]` gives each `A` reference its own key.

### Reuse with `lastIndex` Optimization

On reuse, `runner.run()` is only called when the element's index actually changed (`lastIndex !== i`). This avoids unnecessary process cell writes for elements that didn't move.

When the index did change, `runner.run()` is called with the current `list[i]` element and updated index. Patterns that read `index` reactively update; patterns that ignore `index` skip recomputation.

### Result Array Update

The result array is always set in bulk via `resultWithLog.set(newArrayValue)` after the loop. The cell differ prevents spurious downstream recomputation when the array contents haven't actually changed.

### State

- `elementRuns: Map<string, { resultCell, lastIndex }>` — Persisted across action invocations. Maps identity keys to their result cells and last-known index for reuse. Not pruned on list shrink (documented TODO).

## Files Changed

- **`packages/runner/src/builtins/map.ts`** — Core implementation: identity-based reconciliation loop using `getAsNormalizedFullLink()` with occurrence counting, bulk result set
- **`packages/runner/test/patterns-dynamic.test.ts`** — Three new tests:
  - Inline values: positional identity, expects 3 recomputes on mid-list insert (positions 1-3 shift)
  - Cell-link values: entity identity, expects 1 recompute on mid-list insert (only the new element)
  - Duplicate cell links: occurrence counting ensures `[A, B, A]` gets 3 independent runs

## Known Limitations

- **`elementRuns` is not pruned**: When elements are removed from the list, their entries remain in the map. The runner is stopped via `addCancel` when the parent is disposed, but the Map entries accumulate. Documented as a TODO for high-churn scenarios.

## Test plan

- [x] All runner tests pass
- [x] New inline-values test: verifies positional identity (3 recomputes for shifted inline values)
- [x] New cell-link test: verifies entity identity (1 recompute for new element only)
- [x] New duplicate cell-links test: verifies occurrence counting for repeated references
- [x] Existing CT-1158 regression test still passes (map truncation with null values)

### DIFF
--- a/packages/runner/src/builtins/map.ts
+++ b/packages/runner/src/builtins/map.ts
@@ -128,7 +128,6 @@ export function map(
             { doNotUpdateOnPatternChange: true },
           );
           existing.lastIndex = i;
-          resultWithLog.key(i).set(existing.resultCell);
         }
         newArrayValue.push(existing.resultCell);
       } else {
@@ -153,21 +152,10 @@ export function map(
         resultCell.getSourceCell()!.setSourceCell(parentCell);
         addCancel(() => runtime.runner.stop(resultCell));
         elementRuns.set(elementKey, { resultCell, lastIndex: i });
-        resultWithLog.key(i).set(resultCell);
         newArrayValue.push(resultCell);
       }
     }
-
-    // Update result array length if it changed
-    const currentResult = resultWithLog.get();
-    if (currentResult.length > list.length) {
-      // Use getRaw() to preserve cell references. Using .get() would dereference
-      // cells to their values, losing the reference when the value is null.
-      const rawResult = resultWithLog.getRaw() ?? currentResult;
-      resultWithLog.set(rawResult.slice(0, list.length));
-    } else if (currentResult.length < list.length) {
-      resultWithLog.set(newArrayValue);
-    }
+    resultWithLog.set(newArrayValue);
 
     // NOTE: We leave prior results in elementRuns for now, so they reuse
     // prior runs when items reappear. This means elementRuns grows


mpsalisbury:
When you guys decide what would be best to do, can you help me understand what you've come to? I don't understand the semantics of what you're discussing.

packages/runner/src/builtins/map.ts
```diff
@@ -94,74 +95,86 @@ export function map(
...
+      }
     }
 
-    // Shorten the result if the list got shorter
+    // Update result array length if it changed
     const currentResult = resultWithLog.get();
     if (currentResult.length > list.length) {
       // Use getRaw() to preserve cell references. Using .get() would dereference
       // cells to their values, losing the reference when the value is null.
       const rawResult = resultWithLog.getRaw() ?? currentResult;
```
seefeldb: and then this becomes just currentResult from above

PR #2952, opened 2026-03-03
feat: preserve sparse arrays through the reactive pipeline

mpsalisbury:
## Summary

Sparse JavaScript arrays (arrays with true holes, where `i in arr === false`)
were being silently densified at multiple points in the reactive pipeline. This
caused the `map` builtin to create pattern runs for holes, wasting resources and
producing incorrect output.

This PR fixes all densification sites, adds a hashing boundary fix for
merkle-reference, and rewrites the design doc as a reference for new engineers.

## Changes

**Source fixes (6 files):**
- `packages/memory/storable-value.ts` — `isStorableArray` accepts sparse arrays;
  `toStorableValueLegacy` and `toDeepStorableValueInternal` use `forEach` to
  preserve holes
- `packages/runner/src/storage/transaction/attestation.ts` — `sparseArrayCopy`
  helper using `forEach`, replaces `[...arr]` spread at 3 copy sites
- `packages/runner/src/cell.ts` — `recursivelyAddIDIfNeeded` and `push` method
  use `forEach` instead of `.map()`+spread
- `packages/runner/src/traverse.ts` — `traverseDAG`, `traverseArrayWithSchema`,
  and its caller use `forEach`/`for`+`i in` instead of `.entries()`+`push`
- `packages/runner/src/data-updating.ts` — `normalizeAndDiff` handles four
  hole transition cases; `hasPath` uses `i in value`
- `packages/runner/src/builtins/map.ts` — identity-based reconciliation loop
  (#2942) skips holes, uses indexed assignment into `new Array(len)`

**Hashing boundary fix:**
- `packages/memory/reference.ts` — merkle-reference library can't hash sparse
  array holes; `wrappedNodeBuilder.toTree` densifies sparse arrays (holes→null)
  before passing to the hash library. The canonical hash path already handles
  holes natively.

**Two sparse-safe patterns:**
- `forEach` (preferred) — structural hole-skipping, used at 6 sites
- `for` + `i in` — used at 4 sites that need absence detection or early exit

**Tests:** 13 new tests, 10 updated tests across 8 test files covering every
layer of the pipeline, including an end-to-end test that maps over a sparse
array and verifies the output preserves holes.

**Design doc:** Rewritten from implementation plan to reference document
covering the problem, mechanisms, code locations, and regression prevention.

## Test plan

- [x] All 197 runner tests pass
- [x] All 35 memory tests pass
- [x] Sparse array holes verified with `i in arr` assertions at every layer
- [x] Dense array behavior unchanged (verified by existing + new tests)
- [x] Diff engine covers all 4 hole transitions (hole↔hole, value↔hole,
  hole↔value, value↔value)
- [x] End-to-end: `map([10, <hole>, 30])` produces `[20, <hole>, 60]`

### DIFF
--- a/docs/specs/sparse-array-preservation.md
+++ b/docs/specs/sparse-array-preservation.md
@@ -164,13 +164,25 @@ to correctly report that a path through a hole does not exist.
 
 ### Map builtin (`packages/runner/src/builtins/map.ts`)
 
-The main loop checks `initializedUpTo in list` before creating a pattern run.
-For holes, it extends `newArrayValue.length` without pushing — preserving the
-hole in the output. When the input list changes reactively:
+The identity-based reconciliation loop uses `for` + `i in list` to skip holes.
+It pre-allocates `new Array(list.length)` and uses indexed assignment instead of
+`push`, so holes are structurally preserved in the output. No pattern runs,
+identity keys, or result cells are created for hole indices. When the input list
+changes reactively:
 
 - **Value becomes hole:** The output gets a hole at that index. The pattern run
-  is kept for potential reuse if the same key reappears.
-- **Hole becomes value:** A new pattern run is created (or reused by key match).
+  is kept in `elementRuns` for potential reuse if the same key reappears.
+- **Hole becomes value:** A new pattern run is created (or reused from
+  `elementRuns` if the identity key matches a previous run).
+
+### Hashing boundary (`packages/memory/reference.ts`)
+
+The merkle-reference library cannot hash sparse array holes (it throws
+`TypeError: Unknown type undefined`). The `wrappedNodeBuilder.toTree` wrapper
+detects sparse arrays and densifies them (holes → `null`) before passing to the
+default node builder. This only affects hash computation — the actual data in
+storage remains sparse. The canonical hash path (`canonical-hash.ts`) handles
+holes natively and does not need this workaround.
 
 ## Writing new code that handles arrays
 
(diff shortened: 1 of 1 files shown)


seefeldb:
this is great, lgtm.

for these kinds of things i tend to just start with it in docs/plans/ and once it looks good just run the agents to do it, since it's probably just one PR in the end, then remove the plan from the final PR. we could keep a history of plans somewhere, i'm just a bit afraid of those confusing future agents.

this specific case could morph into an implementation note, but really here i'd rather have code comments in the right places and of course tests.

docs/specs/sparse-array-preservation.md
```diff
@@ -0,0 +1,276 @@
...
+**File:** `packages/runner/src/cell.ts:850`
+
+The cell's array `push` implementation uses `[...array, ...value]` to
+concatenate. If the existing array is sparse, the spread densifies it.
+
+### 4. Cell read/traverse path
+
+**File:** `packages/runner/src/traverse.ts:645-698`
+
+When reading an array from a cell, `traverseDAG` processes elements with:
```
ubik2: We have the same basic problem in `traverseArrayWithSchema`, but it was less obvious since it wasn't a spread. Can do the same solution where we set at index. This should also be changed to a `forEach`, since `entries()` will be densified.

docs/specs/sparse-array-preservation.md
```diff
@@ -0,0 +1,276 @@
...
+The desired behavior: if the input to `map` is `[A, <hole>, <hole>, B]`, the
+output should be `[f(A), <hole>, <hole>, f(B)]`. No pattern should run for
+holes. This should be reactive — filling a hole should spin up a new pattern
+run, and creating a hole (removing a value) should stop producing output at
+that index.
+
+## Background: What Already Works
+
+The low-level storage and serialization layers already handle sparse arrays
+correctly when `richStorableValues` is enabled (which it is):
```
danfuzz: It's not actually enabled by default yet! Still a WIP.

mpsalisbury:
Pushed the implementation. PTAL.
Also, I reworked the design doc to be an explanation and guidance for future implementers about using sparse arrays. Where would be a good home (directory) for this?

PR #2966, opened 2026-03-04
perf(memory): batch schema traversals in provider commit path

seefeldb:
## Summary

- **Batch schema traversals**: Instead of running expensive `getSchemaSubscriptionMatches()` on every commit, accumulate changed doc keys and flush them in a debounced batch. Uses a microtask yield loop with a generation counter — keeps deferring as long as new commits arrive, then runs one traversal for the entire batch.
- **Split direct vs deferred revisions**: Each commit immediately sends cheap direct revisions (SQLite reads for changed docs already tracked by subscriptions). The expensive link-traversal discovery is deferred to the batched flush.
- **Flush before conflict errors**: Preserves the client invariant that subscription data is available before a `ConflictError` returns, so retries work on the newest state.
- **Consumer handles revision-only commits**: The consumer now accepts `EnhancedCommit` with an empty commit payload, treating it as a revision-only delivery.
- **Reverse watch index**: `O(changed docs)` commit matching instead of `O(all watchers)` via a reverse index from entity patterns to watcher IDs.
- **Batch statistics logging**: Tracks flush counts, batch sizes, and timing. Logs per-batch details (`saved` traversals, `flushMs`, `prevFlushMs`) and periodic summaries at warn level.

## Observed results

With real browser workload (250ms+ traversals per doc):
- ~28% of flushes batch multiple commits into a single traversal
- Batch sizes up to 10+ commits coalesced
- Each batched flush saves N-1 redundant traversals

## Test plan

- [x] All memory tests pass (64/64)
- [x] All workspace tests pass (`deno task test`)
- [x] Integration tests pass (`deno task integration`)
- [x] Verified batch logging with local dev server

### DIFF
--- a/packages/memory/provider.ts
+++ b/packages/memory/provider.ts
@@ -230,6 +230,13 @@ export interface Provider<Protocol extends Proto> {
     ucan: UCAN<ConsumerInvocationFor<Ability, Protocol>>,
   ): Await<ConsumerResultFor<Ability, Protocol>>;
 
+  /**
+   * Dispose all sessions without closing the ReadableStream pipe.
+   * Clears pending timers and unsubscribes from memory, but leaves
+   * the consumer's pipe intact.
+   */
+  disposeSessions(): void;
+
   close(): CloseResult;
 }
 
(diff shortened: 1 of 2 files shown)


seefeldb:
@ubik2 added batching behavior -- doesn't run the pattern tests much faster though, which i find surprising. will add some logging and keep testing.

PR #2998, opened 2026-03-06
Fix listPatternIndex fetch + fetchData hashing

bfollington:
## Summary
Fix `listPatternIndex` hanging by ensuring `fetchData` always receives a concrete URL, and make `fetchData` input hashing robust against unresolved/undefined proxy values.

## Changes
- Make `listPatternIndex` use a `Writable` URL defaulted to `/api/patterns/index.md` and update it when `wish("#pattern-index")` resolves.
- Hash `fetchData` inputs from a plain snapshot (not reactive proxies) and use that snapshot hash in the mutex claim path.

## Testing
- Manual: `listPatternIndex` tool now returns without crashing or hanging in local dev.

### DIFF
--- a/packages/runner/src/builtins/fetch-utils.ts
+++ b/packages/runner/src/builtins/fetch-utils.ts
@@ -28,7 +28,8 @@ export function computeInputHashFromValue<T extends Record<string, any>>(
 ): string {
   const safeInputs = inputs ?? {};
   // Exclude 'result' type hint from the hash - only hash actual fetch parameters
-  const { result: _result, ...inputsOnly } = safeInputs;
+  const inputsOnly = { ...(safeInputs as Record<string, unknown>) };
+  delete (inputsOnly as Record<string, unknown>).result;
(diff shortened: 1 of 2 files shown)


packages/runner/src/builtins/fetch-utils.ts
```diff
@@ -22,14 +23,25 @@ export const internalSchema = {
...
+): string {
+  const safeInputs = inputs ?? {};
+  // Exclude 'result' type hint from the hash - only hash actual fetch parameters
+  const { result: _result, ...inputsOnly } = safeInputs;
+  // refer() cannot hash undefined values; normalize to a deep storable shape
+  // (omits undefined object props, converts undefined array elements to null).
+  const storableInputs = toDeepStorableValue(inputsOnly);
+  const normalized = storableInputs === undefined ? {} : storableInputs;
+  return refer(normalized as Record<string, unknown>).toString();
+}
```
bfollington: This is to handle the `undefined` error

PR #3013, opened 2026-03-09
fix(fuse): prevent daemon crash on uncaught cell link errors

bfollington:
## Summary

- Fix FUSE daemon crashing when pattern recomputation throws uncaught "Cannot create cell link - space required" errors from setTimeout callbacks
- Add try/catch in cell-bridge reactive update callback + global error handlers in mod.ts as safety net
- Fix pre-existing type error: `fuse_reply_write` expects `bigint` (usize) but was receiving `number`

Fixes CT-1315

## Test plan

- [x] Mount FUSE filesystem with `ct fuse mount`
- [x] Connect to a space with complex pieces (GMail Importer, patterns with cell links)
- [x] Write to a cell via the filesystem and verify the daemon survives the reactive update
- [x] Confirm errors are logged (not fatal): `[FUSE] Uncaught error: Cannot create cell link...`
- [x] Verify mount remains stable through repeated reads/writes of large pieces (565KB+)

### DIFF
--- a/packages/fuse/mod.ts
+++ b/packages/fuse/mod.ts
@@ -45,6 +45,20 @@ import { HandleMap } from "./handles.ts";
 
 const encoder = new TextEncoder();
 
+// Prevent uncaught errors from crashing the FUSE daemon.
+// Pattern recomputation and cell operations can throw from setTimeout
+// callbacks (e.g. "Cannot create cell link - space required"). These
+// errors should be logged, not fatal.
+globalThis.addEventListener("unhandledrejection", (e) => {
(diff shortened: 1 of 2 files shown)


packages/fuse/mod.ts
```diff
@@ -633,7 +647,7 @@ async function main() {
       }
 
       handles.write(fh, data, off);
-      fuse.symbols.fuse_reply_write(req, sz);
+      fuse.symbols.fuse_reply_write(req, BigInt(sz));
```
bfollington: This was a type error

PR #3023, opened 2026-03-09
fix(ui): Remove unused container-type: inline-size from ct-hstack/ct-vstack

mathpirate:
## Summary

- `ct-hstack` and `ct-vstack` had `container-type: inline-size` on `:host`, but **neither component uses `@container` queries** (only `ct-grid` does)
- `container-type: inline-size` establishes inline-axis size containment, which causes the element's intrinsic size contribution to become **0** when it's a flex child
- This means ct-hstack/ct-vstack elements nested inside flex containers (including inside each other, or inside a `<div style="display:flex">`) collapse to zero width, pushing sibling content off-screen

## Root cause investigation

Initial hypothesis was that `flex-shrink: 0` on `::slotted(*)` was preventing children from shrinking. Testing showed:
1. Removing `flex-shrink: 0` caused children to collapse to zero width (not shrink gracefully)
2. Browser DevTools confirmed `offsetWidth: 0` on ct-hstack flex children despite having `scrollWidth > 0`
3. `container-type: inline-size` was identified as the cause — it makes the element report 0 intrinsic size to the parent flex algorithm

## Change

Remove `container-type: inline-size` from `:host` in both ct-hstack and ct-vstack. The existing `flex-shrink: 0` on `::slotted(*)` stays — it correctly prevents children from shrinking within stack layouts.

Supersedes #3021 which targeted the wrong property.

## Test plan

- [x] Verified with parking-coordinator pattern: buttons (Edit, Remove, Assign) no longer overflow off-screen
- [ ] Verify no visual regression in existing patterns that use ct-hstack/ct-vstack
- [ ] Verify ct-grid (the only component using `@container`) still works correctly (unaffected — different component)

### DIFF
--- a/packages/ui/src/v2/components/ct-hgroup/ct-hgroup.ts
+++ b/packages/ui/src/v2/components/ct-hgroup/ct-hgroup.ts
@@ -24,7 +24,6 @@ export class CTHGroup extends BaseElement {
   static override styles = css`
     :host {
       display: inline-flex;
-      container-type: inline-size;
     }
 
     .group {
--- a/packages/ui/src/v2/components/ct-vgroup/ct-vgroup.ts
+++ b/packages/ui/src/v2/components/ct-vgroup/ct-vgroup.ts
@@ -23,7 +23,6 @@ export class CTVGroup extends BaseElement {
   static override styles = css`
     :host {
       display: block;
-      container-type: inline-size;
     }
 
     .group {


robdodson (approved):
LGTM

@mathpirate do you think we should remove it from ct-hgroup and ct-vgroup as well?

PR #3024, opened 2026-03-09
docs: compilation cache design spec

mpsalisbury:
## Summary

- Design spec for persistent caching of compiled pattern JS output to avoid redundant recompilation across page reloads and process restarts
- Covers invalidation strategy (fingerprint based on code changes), cache storage (IndexedDB for browser, filesystem for server), Engine refactor (split compile/evaluate), and observability via RuntimeTelemetry
- All design decisions documented; no open questions remain

## Key decisions

- **What to cache**: `JsScript` (compiled JS + source maps), not `Pattern` objects (which contain closures)
- **Invalidation**: Single policy — "did any code change?" Browser uses worker bundle hash; server uses `hash(git HEAD + dirty files)` (all files — including deno.lock, deno.json, etc.)
- **Disabling**: Don't construct `CachedCompiler` — its absence means no caching
- **Engine refactor**: Split `process()` into `compile()` + `evaluate()` to allow cache to intercept compilation
- **Build tool change**: Felt gains manifest generation for browser fingerprint (accepted tradeoff, documented)

## Test plan

- [x] Review design doc for completeness and correctness
- [x] Validate fingerprint strategy covers local dev and production scenarios
- [x] Confirm Engine refactor approach is safe (Engine is only Harness impl)

### DIFF
--- a/docs/specs/compilation-cache.md
+++ b/docs/specs/compilation-cache.md
@@ -196,6 +196,9 @@ This covers:
 - New untracked files → included in dirty hash → cache invalidated.
 - Deleted files → path included in hash → cache invalidated.
 - Clean working tree → fingerprint = hash(HEAD) → stable across restarts.
+- **Library dependencies**: `deno.lock` is committed to the repo, so any
+  dependency version change (add, update, or remove) changes either HEAD or the
+  dirty file set — no special handling needed.
 
 Computed once at process startup. Cost is negligible: one `git` invocation plus
 reading a handful of dirty files (typically 0-5 in practice).


docs/specs/compilation-cache.md
```diff
@@ -0,0 +1,694 @@
...
+over-invalidates slightly (e.g., a Cell.ts change busts the cache even though it
+doesn't affect compilation), but the cost is just one recompilation pass —
+equivalent to today's behavior with no cache. The benefit is zero risk of stale
+output.
+
+The mechanism differs per environment, but the semantics are identical:
+
+**Browser (worker bundle hash)**:
+
+The runtime worker is bundled by Felt/esbuild into a single JS file that
```
mathpirate: cool. this all looks really good to me. i do note that this approach means that during active development, the cache will be busted on every save-and-rebuild cycle, which means we won't see perf wins in that scenario. seems fine and hard to accomplish anyway, just want to make sure i'm understanding correctly here

docs/specs/compilation-cache.md
```diff
@@ -0,0 +1,694 @@
...
+degradation, identical to today's behavior).
+
+### Cache Lifecycle
+
+1. **Initialization**: When the runtime starts, compute the fingerprint string
+   and create `CachedCompiler` with the appropriate storage backend. If no
+   fingerprint is available, skip — no `CachedCompiler` means no caching.
+2. **Eager eviction**: On startup, call `evictStale()` to clear entries from
+   previous compiler versions. This is async and non-blocking — compilation can
+   proceed while eviction runs.
```
mathpirate: sounds good; basically this is because there's no TTL-based eviction so this is a storage cleanup optimization, which shouldn't impact runtime cache behavior, right? i.e. there's no race between compilePattern() and evictStale() as far as i can see here

docs/specs/compilation-cache.md
mpsalisbury: Yes, that's right. I went a ways down the path of trying to build a signature that only changed when the compiler output might change, and it was getting way too complex. The only alternative I came up with that sounded reasonable was to build a compiler-diff-sensing test input source file that would exercise 'every' compiler feature such that any change in the output from compiling that file would indicate that the compiler did change and thus invalidated the current cache. But this too feels risky and hard to maintain. So at least for now we get the caching benefits only when we're not actively changing the code. Happy to consider narrower definitions for what might affect the output for followups.

PR #3025, opened 2026-03-09
fix(runner): resolve derive callback crash on recursive pattern return (CT-1316)

seefeldb:
## Summary

Fixes CT-1316: `derive()` callback crashes with `{type: callback:error, message: null}` when returning a recursive pattern instantiation (tail-call pagination).

When a `derive()` callback returns a pattern that recursively calls itself (e.g. paginated API fetching like `FetchContactsPage` in google-contacts-importer), the runtime would either crash with `message:null` or hit "Too many iterations: 101" in the scheduler. This PR fixes three root causes:

### 1. `resultPatternCache` self-invalidation (runner.ts)

The `Map`-based cache on `Runner` was invalidated by storage notifications triggered by `this.run()` — the very call it was meant to deduplicate. Replaced with a **closure-local variable** (`previousResultPatternString`) scoped to each derive action, which survives across reactive re-evaluations without being cleared by storage events.

### 2. TX commit timing race → self-dirtying cycle (runner.ts)

Sub-patterns created inside an action have their process cell data written in the parent's TX. Because `replica.commit()` is always async, the sub-pattern's action gets scheduled and runs **before** the parent's TX commits. At that point:

1. The process cell has no committed data → `getAsQueryResult` returns `{}`
2. The derive callback sees `remaining=undefined` → returns base case `{items:[], done:true}` (a plain value, not a pattern)
3. `sendValueToBinding` writes this plain value to the output, overwriting the result-cell link
4. This write dirties the action → it re-runs → same thing → infinite cycle until `MAX_ITERATIONS_PER_RUN` (100) stops it

Fixed with a **process cell existence guard** that checks `processCell.withTx(tx).getRaw()` before running the action. Uses `ignoreReadForScheduling` so this check does NOT create a reactive subscription — the action is naturally re-triggered via storage notifications when the parent TX commits and the data becomes available.

### 3. `message:null` in error reporting (runner.ts + scheduler.ts)

When the scheduler's loop counter exceeded `MAX_ITERATIONS_PER_RUN`, it created a plain `new Error()` without a `frame`. `handleError` → `getPieceMetadataFromFrame` found no frame → no piece/spell metadata → `message:null` in the iframe error path. Fixed by **storing the action's last frame** on the action object (`lastFrame`) and attaching it to the "Too many iterations" error before dispatching.

Also adds `.catch()` handlers to commit promises in the scheduler to prevent unhandled rejections.

## Changes

- `packages/runner/src/runner.ts`: closure-local pattern cache, process cell existence guard, frame storage on action
- `packages/runner/src/scheduler.ts`: frame attachment on "Too many iterations" error, `.catch()` on commit promises
- `packages/runner/test/patterns-derive-return-pattern.test.ts`: new test file with 3 tests (non-recursive pattern return, recursive tail-call, conditional pattern/plain value)

## Test plan

- [x] `deno test -A packages/runner/test/patterns-derive-return-pattern.test.ts` — all 3 tests pass, no "Too many iterations" in output
- [x] `deno test -A packages/runner/test/patterns-regressions.test.ts` — existing regression tests pass
- [ ] Manual: deploy google-contacts-importer with maxContacts > 100 to verify multi-page pagination

### DIFF
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -1503,10 +1503,19 @@ export class Runner {
         };
 
         try {
-          // CT-1316: If the process cell has no committed data yet (e.g.
-          // when a sub-pattern was created inside an action whose TX hasn't
-          // committed), skip execution. The action will be re-triggered
-          // when the storage notification arrives after commit.
+          // CT-1316: When a derive action creates a sub-pattern via
+          // this.run(), the sub-pattern's process cell data is written in the
+          // parent's TX. Because replica.commit() is always async, the
+          // sub-pattern's action can be scheduled and run before the parent's
+          // TX commits — at which point the process cell has no data yet.
+          // The action sees argument={} and writes a plain value to its
+          // output, which overwrites the result-cell link and re-dirties the
+          // action, causing a self-reinforcing cycle until the loop counter
+          // stops it at MAX_ITERATIONS_PER_RUN.
+          //
+          // We use ignoreReadForScheduling so this check does NOT create a
+          // reactive subscription — the action will be naturally re-triggered
+          // via storage notifications when the parent TX commits.
           const processCellRaw = processCell.withTx(tx).getRaw({
             meta: ignoreReadForScheduling,
           });


seefeldb:
Superseded by #3034 which was already merged.

PR #3029, opened 2026-03-09
feat(runner): split Engine.process() into compile() + evaluate()

mpsalisbury:
## Summary

- Phase 1 of the compilation cache design ([#3024](https://github.com/commontoolsinc/labs/pull/3024))
- Splits `Engine.process()` into independent `compile()` and `evaluate()` methods, enabling the compilation cache to intercept compilation without infecting Engine internals
- Adds `compile()` and `evaluate()` to the `Harness` interface
- Adds `engine.test.ts` with 14 test steps covering compile, evaluate, their equivalence with process, and run

## Test plan

- [x] All 208 existing runner tests pass unchanged (pure refactor)
- [x] New `engine.test.ts` tests verify:
  - `compile()` produces JsScript without evaluation
  - `evaluate()` executes pre-compiled JS and maps exports correctly
  - `compile()` + `evaluate()` produces equivalent results to `process()`
  - `run()` works correctly through the refactored `process()`
  - `process({ noRun: true })` only compiles
  - Error cases (missing exports, compilation errors)

### DIFF
--- a/packages/runner/test/engine.test.ts
+++ b/packages/runner/test/engine.test.ts
@@ -38,12 +38,13 @@ describe("Engine.compile()", () => {
       ],
     };
 
-    const jsScript = await engine.compile(program);
+    const result = await engine.compile(program);
 
-    expect(jsScript).toBeDefined();
-    expect(jsScript.js).toBeDefined();
-    expect(typeof jsScript.js).toBe("string");
-    expect(jsScript.js.length).toBeGreaterThan(0);
+    expect(result.jsScript).toBeDefined();
+    expect(result.jsScript.js).toBeDefined();
+    expect(typeof result.jsScript.js).toBe("string");
+    expect(result.jsScript.js.length).toBeGreaterThan(0);
+    expect(result.id).toBeDefined();
   });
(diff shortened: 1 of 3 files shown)


packages/runner/test/engine.test.ts
```diff
@@ -0,0 +1,380 @@
...
+          ].join("\n"),
+        },
+      ],
+    };
+
+    const result = await engine.run(program);
+    expect(result).toBeDefined();
+    expect(result.nodes).toBeDefined();
+  });
+});
```
mathpirate: do we want to have a test that asserts any particular behavior if the caller makes a mistake a compiles program A but evaluates with program B's RuntimeProgram? maybe we don't want to claim anything in particular about what happens in that scenario, i'm unsure

mpsalisbury: The refactor to pass ID in separately to evaluate() should help with this. And I think passing in the wrong files should be less of a problem. I don't know what the behavior should be if these are mismatched, but I suspect we shouldn't claim anything about that.

PR #3031, opened 2026-03-09
feat(runner): add compilation cache infrastructure

mpsalisbury:
## Summary

Phase 2 of the compilation cache design ([spec](https://github.com/commontoolsinc/labs/pull/3024)). Builds on Phase 1 ([#3029](https://github.com/commontoolsinc/labs/pull/3029)).

- `CompilationCacheStorage` interface with `get`/`set`/`evictStale`/`evictOldest`/`clear`/`count`
- Three storage backends: `MemoryCompilationCache` (tests), `FileSystemCompilationCache` (server), `IDBCompilationCache` (browser)
- `CachedCompiler` orchestration: fingerprint verification, hit/miss stats with miss reason tracking, startup stale eviction + summary
- Non-blocking interval-based count eviction (`maybeEvictByCount`): fire-and-forget, runs every 50 writes, boolean lock prevents concurrent runs
- `computeGitFingerprint()`: server-side fingerprint from `git rev-parse HEAD` + dirty file contents hash. Returns `undefined` when not in a git repo (cache disabled)
- Path traversal validation on `FileSystemCompilationCache` keys
- IDB backend uses index-based eviction (avoids materializing all values)
- Environment annotations: doc comments on each backend noting where it can run (browser-only, server-only, test-only)
- Scoped test permissions: `--allow-write=/tmp,/var/folders` (temp dirs only), `--allow-run=git` (git fingerprint only)

No integration with `PatternManager` yet — that's Phase 4.

## File layout

```
packages/runner/src/compilation-cache/
  mod.ts                 # CachedCompiler, re-exports
  storage.ts             # CompilationCacheStorage interface
  idb-storage.ts         # IndexedDB implementation (browser)
  fs-storage.ts          # Filesystem implementation (server / Deno)
  memory-storage.ts      # In-memory implementation (tests)
  git-fingerprint.ts     # computeGitFingerprint() for server
```

## Test plan

- [x] Storage interface tests run against all three backends (Memory, FileSystem, IDB) via shared `storageTests()` harness (9 tests each)
- [x] IDB tests use `fake-indexeddb` polyfill, kept as test-only inline import (not in package import map)
- [x] FileSystem path traversal validation tests (rejects `/`, `\`, `..` in keys)
- [x] CachedCompiler tests: cache hit/miss, fingerprint mismatch, stats tracking, eviction, clear
- [x] Eviction cascade test: stale eviction alone brings count under cap
- [x] Eviction interval gating test: writes below threshold don't trigger eviction
- [x] Git fingerprint: returns 64-char hex string, deterministic across calls
- [x] All 224 runner tests pass

### DIFF
--- a/packages/runner/test/compilation-cache.test.ts
+++ b/packages/runner/test/compilation-cache.test.ts
@@ -358,29 +358,33 @@ describe("CachedCompiler", () => {
     // Create a compiler with cap=2 and evictionInterval=1 so every write checks
     const smallCompiler = new CachedCompiler(storage, "fp", 2, 1);
 
-    // Add 3 entries — the third should trigger eviction of the oldest
-    await smallCompiler.set("hash1", testJsScript);
-    // Ensure distinct cachedAt by using the internal storage directly
-    // to control timestamps
+    // Seed 3 entries with hardcoded timestamps via storage directly
+    await storage.set("hash1", {
+      jsScript: testJsScript,
+      fingerprint: "fp",
+      cachedAt: 1000,
+    });
     await storage.set("hash2", {
       jsScript: testJsScript2,
       fingerprint: "fp",
-      cachedAt: Date.now() + 1000,
+      cachedAt: 3000,
     });
     await storage.set("hash3", {
       jsScript: testJsScript,
       fingerprint: "fp",
-      cachedAt: Date.now() + 2000,
+      cachedAt: 2000,
     });
 
-    // Trigger eviction by setting a 4th entry
+    // Trigger eviction by writing through the compiler (4th entry, over cap of 2)
     await smallCompiler.set("hash4", testJsScript2);
 
     // Eviction is fire-and-forget; yield to let it complete
     await new Promise((r) => setTimeout(r, 0));
 
-    // Should have evicted down to 2
+    // Should have evicted down to 2 — hash1 (oldest) and hash3 removed
     expect(await storage.count()).toBe(2);
+    expect(await storage.get("hash2")).toBeDefined();
+    expect(await storage.get("hash4")).toBeDefined();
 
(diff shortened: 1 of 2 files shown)


packages/runner/src/compilation-cache/mod.ts
```diff
@@ -0,0 +1,133 @@
...
+      missReasons: { ...this.stats.missReasons },
+    };
+  }
+
+  /** Get the fingerprint in use. */
+  getFingerprint(): string {
+    return this.fingerprint;
+  }
+
+  private async evictByCount(): Promise<void> {
```
mathpirate: this is obviously not concurrency-safe; in particular the count() / evictStale() / count() / evictOldest() sequence is all async, so if someone else used set() during that execution chain it would mess things up

i know deno is single-threaded and currently we don't have to worry about this issue. is there any future in which we'd consider breaking these assumptions?

packages/runner/test/compilation-cache.test.ts
```diff
@@ -0,0 +1,350 @@
...
+    const smallCompiler = new CachedCompiler(storage, "fp", 2);
+
+    // Add 3 entries — the third should trigger eviction of the oldest
+    await smallCompiler.set("hash1", testJsScript);
+    // Ensure distinct cachedAt by using the internal storage directly
+    // to control timestamps
+    await storage.set("hash2", {
+      jsScript: testJsScript2,
+      fingerprint: "fp",
+      cachedAt: Date.now() + 1000,
```
mathpirate: if we start seeing flakes maybe we'll change these to hardcoded timestamps

mpsalisbury: agreed

PR #3037, opened 2026-03-09
fix(runner): propagate $defs in array item schemas and improve $ref error handling

mathpirate:
## Summary

Three fixes from investigating bugs surfaced on `fix/strength-tracker-cli-hardening`:

- **cache.ts**: `collectLinkedCellSyncs` now carries `$defs` from the parent array schema into the extracted item schema. Previously, when array items used `$ref` pointing to `$defs` defined on the parent (as generated by `Default<>` on nested interface fields), downstream `sync → pull → joinSchema → resolveSchemaRefsOrThrow` would crash with `Failed to resolve $ref: #/$defs/...`. Includes a regression test that fails without the fix and passes with it.
- **cfc.ts**: `resolveSchemaRefsOrThrow` now throws an actionable error for unresolvable external `$ref` URLs, directing developers to register the schema in `embeddedSchemas`. Adds test coverage for the `$ref` resolution path.
- **piece-controller.ts**: `getPatternMeta()` no longer calls `loadPattern()` before `loadPatternMeta()`. The `loadPattern()` call was unnecessary (triggered a full TS compile) and crashed when pattern source was missing from storage.

## Test plan

- [x] New regression test `array-ref-defs-propagation.test.ts` — fails without fix, passes with it
- [x] `resolveSchemaRefsOrThrow` tests in `cfc.test.ts`
- [x] Full `deno task test` suite passes
- [x] Manually reproduced `$ref` crash via CLI deploy + handler call, verified fix resolves it
- [ ] CI passes

### DIFF
--- a/packages/runner/src/storage/cache.ts
+++ b/packages/runner/src/storage/cache.ts
@@ -2322,16 +2322,11 @@ export class StorageManager implements IStorageManager {
 
     // Recurse into object/array
     if (Array.isArray(value)) {
-      let itemSchema = schema && isObject(schema) && schema.items
-        ? schema.items as JSONSchema
-        : undefined;
-      // Carry $defs from parent schema so $ref pointers in items can resolve
-      if (
-        itemSchema && isObject(itemSchema) && isObject(schema) && schema.$defs
-      ) {
-        itemSchema = { ...itemSchema, $defs: schema.$defs };
-      }
-      for (const item of value) {
+      for (let i = 0; i < value.length; i++) {
+        const item = value[i];
+        const itemSchema = schema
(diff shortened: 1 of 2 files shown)


mathpirate:
> lgtm
> 
> I had some concerns about not running the compile anymore before returning the meta cell, but it looks like that's intentional.
> 
> I prefer to use cfc.getSchemaAtPath for the array path to support prefixItems, but feel free to merge without that, since you didn't introduce that issue. If you merge as-is, I'll add the cfc.getSchemaAtPath later.

It's intentional but I would love some additional input about it, as indicated in the comment above. I've changed this PR to use your recommended approach of cfc.getSchemaAtPath

PR #3039, opened 2026-03-10
  Add ValuePath type to enforce path invariant in traverse                                                  

ubik2:
### Summary

The traversal pipeline has always required that IMemorySpaceAddress.path starts with "value" — previously 
enforced only by a runtime check inside getNormalizedLink and a comment. This PR makes that invariant
explicit in the type system.

- Introduces ValuePath = readonly ["value", ...string[]] and two narrowed types, IMemorySpaceValueAddress 
and IMemorySpaceValueAttestation, which are used throughout the internal traversal pipeline (getAtPath,
followPointer, traverseWithSchema, and all private SchemaObjectTraverser methods).
- Replaces ad-hoc [...path, part] spreads with appendToPath / appendPartsToPath helpers that preserve the
ValuePath type.
- Rewrites the lastExisting path logic in followPointer to maintain ValuePath throughout, replacing an
implicit runtime assumption with an explicit check that throws with a descriptive error and log line.
- Adds a "value" path guard at the top of SchemaObjectTraverser.traverse(), converting a latent failure
mode into an early, clear error.
- Updates space-schema.ts (loadFactsForDoc) to validate preconditions upfront, pre-navigate to
fact.value.value, and construct a properly-typed IMemorySpaceValueAttestation — removing the dependency on
the old IMemorySpaceAttestation import.
- Updates tests to construct docs as IMemorySpaceValueAttestation from the start, leveraging contextual   
typing rather than call-site casts.

---
## Summary by cubic
Enforces value-rooted traversal at the type level and validates paths early. The public traverser now requires `IMemorySpaceValueAttestation` (paths start with "value").

- **Refactors**
  - Introduced `ValuePath`; exported `IMemorySpaceValueAddress`/`IMemorySpaceValueAttestation`.
  - Made `traverse`, `getAtPath`, and `followPointer` require value-rooted addresses; added `appendToPath`/`appendPartsToPath`.
  - In `schema.ts`, construct `IMemorySpaceValueAddress` by prepending `"value"` to link paths.
  - In server load (`loadFactsForDoc`), enforce `{ value: ... }`, build value-rooted attestations, and slice selector paths.
  - Updated tests and fixtures: facts nest under `value`; tests pass `IMemorySpaceValueAttestation` to `traverse`.

- **Bug Fixes**
  - `followPointer` keeps `lastExisting` as a `ValuePath` and throws on non-"value" paths.
  - `narrowSchema` returns a selector whose path starts with `"value"` on mismatch.
  - `loadFactsForDoc` validates fact shape and selector path before traversal.

<sup>Written for commit 11d4f8cc27c921f219f14ed4e6686cadee60db86. Summary will update on new commits.</sup>

### DIFF
--- a/packages/runner/integration/traverse_timing.test.ts
+++ b/packages/runner/integration/traverse_timing.test.ts
@@ -8,6 +8,7 @@ import type { SchemaPathSelector, URI } from "../src/storage/interface.ts";
 import {
   type BaseMemoryAddress,
   type IAttestation,
+  IMemorySpaceValueAttestation,
   ManagedStorageTransaction,
   type ObjectStorageManager,
   SchemaObjectTraverser,
@@ -148,11 +149,11 @@ function runTest(objectManager: TestObjectManager) {
   const tx = new ExtendedStorageTransaction(managedTx);
   const traverser = new SchemaObjectTraverser(tx, selector);
   const doc = objectManager.load(docAddress)!;
-  const factValue = {
+  const factValue: IMemorySpaceValueAttestation = {
     address: {
       ...doc.address,
       space: "did:null:null" as MemorySpace,
-      path: [...doc.address.path, "value"],
+      path: ["value"],
     },
     value: (doc.value as JSONObject).value,
   };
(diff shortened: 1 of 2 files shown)


ubik2:
Oops. Need to update test calls of traverse.

PR #3065, opened 2026-03-11
fix(ts-transformers): stabilize ternary branch lowering

seefeldb:
## Summary
- fix ternary branch lowering so pure pattern branches stay pattern-lowered and compute branches do not produce nested inner derives
- add shared compute-wrap invariants for ternary and binary fallback wrapping, with argument-position awareness for `ifElse` / `when` / `unless` predicates
- add capability-first regressions for computed arrays, writable-backed ternaries, nested ternary arithmetic, and `ifElse` predicate binaries; update affected fixture goldens back to the corrected lowering shape

## Testing
- `deno fmt`
- `deno lint`
- `deno task test`
- `deno task integration` *(no compiler invariant/assert failures; still fails on the pre-existing `packages/patterns/integration/default-app.test.ts` timeout)*

---
## Summary by cubic
Stabilizes ternary branch lowering in `ts-transformers` and unifies conditional helper arg rewriting using shared compute-wrap invariants. Pure pattern branches stay pattern-lowered, compute branches avoid nested derives, and helper-owned args derive only when needed.

- **Bug Fixes**
  - Keep pure ternary branches pattern-lowered; preserve plain array maps in derive branches.
  - Avoid nested derives in compute branches; stop wraps at helper and nested-helper boundaries.
  - Rewrite `ifElse`/`when`/`unless` args uniformly; allow derive only when needed; ignore synthetic calls.
  - Preserve captures across cloned callbacks to satisfy authored and transformed closure needs.
  - Stabilize default app integration by clicking `x-piece-link` anchors and waiting for view state.

- **Refactors**
  - Introduced shared `compute-wrap-invariants` and `helper-owned-expression`; applied to conditional/binary/call emitters with labeled assertions for safe wrap candidates.
  - Memoized repeated data-flow analysis; added analyzer caching test plus new capability-first and kitchensink fixtures covering nested map lowering.

<sup>Written for commit 51c633618f1a8204b88a5b6e5bd93bc3c4782971. Summary will update on new commits.</sup>

### DIFF
--- a/packages/ts-transformers/test/capability-first.test.ts
+++ b/packages/ts-transformers/test/capability-first.test.ts
@@ -82,7 +82,10 @@ const p = pattern<{ items: string[] }>((state) =>
 
     const output = await transformSource(source);
 
-    assertStringIncludes(output, "=> state.items.map((item) => item.toUpperCase())");
+    assertStringIncludes(
+      output,
+      "=> state.items.map((item) => item.toUpperCase())",
+    );
     assert(!output.includes("state.items.mapWithPattern("));
   },
 );
@@ -99,7 +102,10 @@ const p = pattern<{ items: string[] }>((state) =>
 
     const output = await transformSource(source);
 
-    assertStringIncludes(output, "({ items }) => items.map((item) => item.toUpperCase())");
+    assertStringIncludes(
+      output,
+      "({ items }) => items.map((item) => item.toUpperCase())",
+    );
     assert(!output.includes("items.mapWithPattern("));
   },
 );
@@ -190,7 +196,9 @@ const p = pattern<{ items: string[] }>((state) => {
 
     assertStringIncludes(output, "foo.filterWithPattern(");
     assertStringIncludes(output, "filtered.mapWithPattern(");
-    assert(!output.includes("return filtered.map((item) => item.toUpperCase())"));
+    assert(
+      !output.includes("return filtered.map((item) => item.toUpperCase())"),
+    );
   },
 );
 


packages/ts-transformers/src/transformers/opaque-ref/emitters/call-expression.ts
```diff
@@ -1,25 +1,57 @@
...
+  helperName: "ifElse" | "when" | "unless",
+  index: number,
+): string {
+  if (helperName === "ifElse") {
+    if (index === 0) return "ifElse condition";
+    if (index === 1) return "ifElse true branch";
+    return "ifElse false branch";
+  }
+
+  if (helperName === "when") {
```
mathpirate: should have `unless` as well probably?

packages/ts-transformers/src/transformers/opaque-ref/emitters/conditional-expression.ts
```diff
@@ -3,74 +3,62 @@ import ts from "typescript";
...
-// Helper to process a conditional branch (whenTrue/whenFalse)
 function processBranch(
   expr: ts.Expression,
-  dataFlows: NormalizedDataFlowSet,
-  analysis: ReturnType<Parameters<Emitter>[0]["analyze"]>,
   context: Parameters<Emitter>[0]["context"],
   analyze: Parameters<Emitter>[0]["analyze"],
   rewriteChildren: Parameters<Emitter>[0]["rewriteChildren"],
 ): ts.Expression {
+  const branchAnalysis = analyze(expr);
```
mathpirate: we now re-analyze here, which is probably correct but in theory could introduce inconsistencies between the results of the analysis + normalization that already happened at the outer layer vs the new one we undertake here. idk maybe we just leave it as-is, maybe we add a comment here indicating this property for future investigations?

PR #3068, opened 2026-03-11
In setAtPath, don't delete undefined keys/array entries

ubik2:
Part of a larger set of changes to support undefined throughout the system
There will likely be downstream problems from these, since we were stripping undefined values with this logic, and now will end up trying (unsuccessfully) to send them as JSON across the wire

---
## Summary by cubic
Stop treating undefined as a delete signal in setAtPath. We now preserve undefined in arrays and objects, and comments clarify that deletion requires replacing the parent container.

- **Migration**
  - Replace setAtPath(..., undefined) delete calls by building a new object/array without that key/element and setting it at the parent path.
  - If you serialize to JSON, note that undefined object properties are dropped and undefined array entries become null; adjust callers if needed.

<sup>Written for commit 0430995e54eda2c34b4416d55c14f74706845f2f. Summary will update on new commits.</sup>

### DIFF
--- a/packages/runner/src/storage/transaction/attestation.ts
+++ b/packages/runner/src/storage/transaction/attestation.ts
@@ -136,8 +136,8 @@ const setAtPath = (
     if (rest.length === 0) {
       if (root[index] === value) return { ok: root }; // noop
       const newArray = sparseArrayCopy(root);
-      // We delete elements from an array by setting the array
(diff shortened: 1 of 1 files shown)


ubik2:
Closing this out, since the undefined support landed in a different form (this code path went away).

PR #3073, opened 2026-03-12
ui: introduce component-scoped token contracts

robdodson:
## What This PR Does
Each component now defines and consumes its **own CSS variables** (for example, `--ct-button-*`, `--ct-card-*`, `--ct-code-editor-*`) instead of reading global theme variables directly.

This PR also adds a number of missing stories that I uncovered while I was adding the new variables.

## Why This Matters
Today we have two layers for styling:
- **Primitives** in `variables.ts` (`--ct-colors-primary-500`, `--ct-spacing-4`)
- **Semantic tokens** in `theme-context.ts` (`--ct-theme-color-primary`, `--ct-theme-color-surface`)

Components read semantic tokens directly but there is no component-level namespace. This PR adds a **component contract layer** where each component defines its own CSS custom properties that default to semantic tokens:

  ```css
  /* ct-button contract */
  :host {
    --ct-button-bg: var(--ct-theme-color-primary);
    --ct-button-color: var(--ct-theme-color-primary-foreground);
    --ct-button-radius: var(--ct-theme-border-radius);
  }
  ```
- Components only read their own `--ct-button-*` / `--ct-code-editor-*` variables internally
- Themes only write to `--ct-theme-*` semantic tokens
- Hosts can override either at the theme level or directly on the component.

In the next PR I would like to add pseudo-private variables per this pattern (https://lea.verou.me/blog/2021/10/custom-properties-with-defaults/) which will enable overriding values at the container level. I saw at least one existing component that already uses this approach.

Doing so would enable patterns like: "all buttons inside this sidebar are compact" by setting `--ct-button-padding` on a _container_, without touching the global theme or having to directly override the variable on each button.

### DIFF
--- a/packages/patterns/catalog/stories/ct-alert-story.tsx
+++ b/packages/patterns/catalog/stories/ct-alert-story.tsx
@@ -36,7 +36,7 @@ export default pattern<AlertStoryInput, AlertStoryOutput>(() => {
             variant={variant}
             dismissible={dismissible}
           >
-            {showIcon ? <span slot="icon">{icon}</span> : null}
+            {showIcon.get() ? <span slot="icon">{icon}</span> : null}
             <span slot="title">{title}</span>
             <span slot="description">{description}</span>
             {body}
--- a/packages/ui/src/v2/components/ct-alert/ct-alert.ts
+++ b/packages/ui/src/v2/components/ct-alert/ct-alert.ts
@@ -53,12 +53,21 @@ export class CTAlert extends BaseElement {
         --ct-theme-color-error-foreground,
         #fef2f2
       );
-      --ct-alert-color-warning: #f59e0b;
-      --ct-alert-color-warning-foreground: #451a03;
-      --ct-alert-color-success: #10b981;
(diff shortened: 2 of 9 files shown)


packages/ui/src/v2/components/ct-modal/styles.ts
```diff
@@ -8,20 +8,30 @@ import { css } from "lit";
...
+    --ct-modal-border-radius: var(--ct-theme-border-radius, 12px);
 
     /* CSS custom properties for customization */
     --_backdrop-color: var(--ct-modal-backdrop-color, rgba(0, 0, 0, 0.5));
     --_backdrop-blur: var(--ct-modal-backdrop-blur, 8px);
-    --_border-radius: var(
-      --ct-modal-border-radius,
-      var(--ct-theme-border-radius, 12px)
-    );
+    --_border-radius: var(--ct-modal-border-radius, 12px);
```
robdodson: ct-modal used an existing patter of underscored, private internal vars. Leaving that pattern in place for now, but planning to come back and see if there's any risk to simplifying this approach.

PR #3092, opened 2026-03-13
feat(schema-generator): emit { type: "unknown" } for TS unknown instead of true

seefeldb:
## Summary

- The schema generator now emits `{ type: "unknown" }` for TS `unknown` type, instead of `true` (which means "any")
- This applies in three codepaths: PrimitiveFormatter (bare unknown), ArrayFormatter (unknown[]), and SchemaGenerator (UnknownKeyword)
- The runtime already handles `{ type: "unknown" }` correctly — returns an opaque placeholder instead of traversing data
- Fixed 7 generated patterns that had unknown-typed interface properties where concrete types were needed
- Added tests for `Writable<unknown>` and `Array<Writable<unknown>>` schemas

### Follow-up needed

The `schema:unknown-type-access` validation in type-shrinking.ts only catches top-level unknown parameters, not unknown-typed properties within concrete interfaces. A follow-up PR should extend validateShrinkCoverage to detect this case.

## Test plan

- [x] `deno task test` — all unit tests pass (with updated fixture goldens)
- [x] `deno task integration` — 8/8 suites pass
- [x] Tests for Writable and Array Writable unknown schemas

---
## Summary by cubic
Emit { type: "unknown" } for TS unknown in `schema-generator` so unknown stays opaque at runtime, and keep unions with unknown as anyOf so other branches can still match.

- **New Features**
  - `schema-generator`: emit `{ type: "unknown" }` for `unknown` (and `items: { type: "unknown" }` for `unknown[]`/`Array<unknown>`); added tests for `Writable<unknown>` and `Array<Writable<unknown>>`.

- **Bug Fixes**
  - Preserve `anyOf` for synthetic unions containing `unknown` (e.g., `unknown | { foo: string }`) instead of collapsing to `{ type: "unknown" }`.
  - Generated patterns: replaced `unknown`-typed properties with concrete types; fixed `research-citation-manager` to use `string` for the `style` field in `updateActiveStyle`.

<sup>Written for commit 210b339287be52c1779725d80836a1fbc3099361. Summary will update on new commits.</sup>

### DIFF
--- a/packages/generated-patterns/integration/patterns/research-citation-manager.pattern.ts
+++ b/packages/generated-patterns/integration/patterns/research-citation-manager.pattern.ts
@@ -316,7 +316,7 @@ const retagCitation = handler(
 
 const updateActiveStyle = handler(
   (
-    event: { style?: unknown } | string | undefined,
+    event: { style?: string } | string | undefined,
     context: { style: Cell<CitationStyle | string> },
   ) => {
     const requested = typeof event === "string"


ubik2 (approved):
So I think this is generally worth doing.

I was concerned about how it shifts pattern authors to use `any` instead of `unknown` in some places, which means that they don't get as much type safety in their development. I do think the ability to easily exclude that data is more important, though.

There's mention of anyOf for synthetic unions. In practice, we will resolve an anyOf with an unknown as essentially an unknown (just like typescript), which means that the first match on unknown will prevent us from including the value of the `{foo: string}`. In this case, that would still work if the value was a primitive string, but if it was a link to a string, we wouldn't include it.

seefeldb:
> So I think this is generally worth doing.
> 
> I was concerned about how it shifts pattern authors to use `any` instead of `unknown` in some places, which means that they don't get as much type safety in their development. I do think the ability to easily exclude that data is more important, though.
> 
> There's mention of anyOf for synthetic unions. In practice, we will resolve an anyOf with an unknown as essentially an unknown (just like typescript), which means that the first match on unknown will prevent us from including the value of the `{foo: string}`. In this case, that would still work if the value was a primitive string, but if it was a link to a string, we wouldn't include it.

oh, we don't follow all branches? so anyOf for `{ name: string } | { name: string, city: string }` would drop city? or is unknown a special case?

PR #3116, opened 2026-03-17
feat(runner): support TOOLSHED_GIT_SHA for compilation cache in Docker

mpsalisbury:
## Summary

- Add `TOOLSHED_GIT_SHA` env var as an explicit fingerprint for the server-side compilation cache, enabling it in Docker/binary deployments where `.git` is absent
- Priority: `TOOLSHED_GIT_SHA` (explicit, operator-set) > live git (auto-detected, captures dirty files) > cache disabled
- Update spec, implementation, and tests to reflect the new priority chain

## Test plan

- [x] `explicit SHA takes priority over git` — proves explicit SHA wins even inside a git repo
- [x] `explicit SHA is deterministic` — same input produces same fingerprint
- [x] `different explicit SHAs produce different fingerprints` — different deploys invalidate correctly
- [x] Pre-existing git fingerprint and compilation cache tests still pass

### DIFF
--- a/docs/specs/compilation-cache.md
+++ b/docs/specs/compilation-cache.md
@@ -7,10 +7,10 @@
 ## Problem
 
 Pattern compilation (TypeScript → JavaScript) is expensive: type-checking, AST
-transforms, and AMD bundling take 100–500ms per pattern. Today, compiled JS is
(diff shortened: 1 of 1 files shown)


willkelly (commented):
belated lgtm!

PR #3118, opened 2026-03-17
Properly handle asCell/asStream in processDefaultValue, since these c…

ubik2:
Properly handle asCell/asStream in processDefaultValue, since these can occur in the resolvedSchema (via the $defs entry) instead of the schema.

---
## Summary by cubic
Fixes default value handling for `asCell`/`asStream`, including flags and defaults defined via `$defs` and property `$ref`s. We now resolve referenced schemas first, then apply flags in `processDefaultValue` so cells/streams behave correctly.

- **Bug Fixes**
  - Resolve `$defs` and property `$ref`s before applying defaults so `default`, `asCell`, and `asStream` from referenced schemas are applied; create immutable cell defaults when provided by a referenced schema.
  - Move `asCell`/`asStream` handling out of `resolveSchema` and into `processDefaultValue` (capture and strip flags after resolution, create default cells, keep the `asStream` default warning); remove the `filterAsCell` argument and related filtering from `resolveSchema`.

<sup>Written for commit 2696491914763c3ebd32f71dce674a2728384ec2. Summary will update on new commits.</sup>

### DIFF
--- a/packages/runner/test/schema-ref-default.test.ts
+++ b/packages/runner/test/schema-ref-default.test.ts
@@ -15,7 +15,7 @@ describe("$ref with default support", () => {
         default: "ref-site-default",
       };
 
-      const resolved = resolveSchema(schema, false);
+      const resolved = resolveSchema(schema);
       expect(resolved).toEqual({
         $defs: schema.$defs,
         type: "string",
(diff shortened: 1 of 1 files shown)


danfuzz:
Per your comment in PR #3111, please do go ahead and add whatever other changes make sense to this code. I'll deal with the merge conflicts.

PR #3128, opened 2026-03-18
Cleanup type predicates

ubik2:
- isObject loses its type predicate (value is object → boolean). Narrowing to object was misleading — it doesn't grant property access — so all call sites that needed narrowing now use isRecord instead (Record<string, unknown>), which actually gives you field access. isObject is kept as a plain boolean guard for non-narrowing checks.                                                                                                                                              
- isNumber semantics corrected: isNumber is now a proper type predicate (value is number) that includes NaN/Infinity. The old finite-only behavior moves to isFiniteNumber. Traversal code that cares about finite numbers (getJsonType, schema type dispatch) is updated to use isFiniteNumber.    
- assertIsObject added as an assertion-style alternative (asserts value is object) for contexts where the non-object case should throw.
- isInstance type predicate fixed — previously always returned true from the type system's perspective; now correctly narrows. A previously unreachable branch in shallowStorableFromNativeValueLegacy was added to handle typescript's type uncertainty.
- isFalseSchema logic broadened: the old check was schema["not"] === true (only {not: true} was treated as false). The new check is "not" in schema && !isTrueSchema(schema.not), which correctly treats {not: {}} as a false schema per JSON Schema semantics.                                                           
- _combineSchemaUncached restructured: the flat isObject(linkSchema) && ... if-else chain is collapsed under a single isRecord guard with nested type === "object" / type === "array" branches, reducing duplication and making the structure clearer.
- added some tests for new code

---
## Summary by cubic
Standardizes type guards and JSON Schema handling across the repo. Finalizes the `isObject` → `isRecord` migration, introduces `isFiniteNumber`, fixes boolean-schema logic, and preserves `$defs` through resolution/merges with tests.

- **Refactors**
  - `isObject` is now a plain boolean check; call sites use `isRecord` for field access. Added `assertIsObject`.
  - `isNumber` now only checks type; added `isFiniteNumber` and updated traversal (`getJsonType`, type dispatch) and tests to use it.
  - Simplified schema combination and flag merging; clarified `asCell`/`asStream` inheritance and made merging robust when schema is `true`/`false`.
  - Preserves `$defs` when resolving/merging schemas in CFC/traversal and exports; added tests (opaque refs).
  - Completed remaining `isRecord` migrations and removed redundant casts across CLI, HTML (safer VNode checks), iframe sandbox, LLM prompts, memory, piece, runner (builder/types/cell/cfc/schema/traverse/storage/cache), schema-generator, and tests.
  - `isInstance` corrected to reliably detect non-plain objects (now a plain boolean check); legacy storable conversion now throws on unexpected non-record objects.

- **Bug Fixes**
  - Corrected `isFalseSchema`: treats `{ not: true }` and `{ not: {} }` as false per JSON Schema; added tests.
  - `canBranchMatch` now handles boolean schemas and does not reject `NaN`/`Infinity` for `{ type: "number" }`; added tests.

<sup>Written for commit 6334b032228ae42bb3457aaa972a58f927d263dc. Summary will update on new commits.</sup>


ubik2:
@danfuzz just tagged you on this, since the PR came out of our conversation earlier today

PR #3156, opened 2026-03-23
feat: schema hashing dispatch layer and SchemaAndHash class

danfuzz:
## Summary

Infrastructure for schema hashing in `data-model`, replacing direct `stableStringify` usage in `traverse.ts` with a dispatched `hashSchema()` function behind a new experiment flag. This lays the groundwork for re-introducing `stableHash` — previous attempts (PRs #2996, #3020) were reverted; this approach extracts schema hashing into `data-model` with a modern/legacy dispatch layer rather than swapping `stableStringify` directly.

**Four commits:**

| Commit | Scope | Description |
|--------|-------|-------------|
| `abc679f68` | `data-model` | `SchemaAndHash` class — deep-frozen container pairing `JSONSchema` with `FabricHash`, plus tests |
| `f828b473b` | `data-model` | Schema hashing dispatch layer: `schema-hash-legacy.ts` (extracts `stableStringify` from runner), `schema-hash.ts` (dispatch), `SchemaPathSelector` type moved to `interface.ts` |
| `641ebe035` | `runner` | Wire `hashSchema()` into `traverse.ts`, replacing 6 `stableStringify` call sites |
| `0d0de7f30` | `data-model` + `runner` | `modernSchemaHash` experiment flag + dispatch skeleton (`schema-hash-modern.ts` stubs throw) |

## Key design points

- **Dispatch pattern:** Same modern/legacy dispatch used elsewhere in `data-model` — `schema-hash.ts` routes to `-legacy.ts` or `-modern.ts` based on the experiment flag.
- **No behavioral change yet:** Legacy path preserves exact `stableStringify` behavior; modern path is a stub that throws (to be implemented in a follow-up).
- **`SchemaAndHash`:** Immutable container that pairs a schema with its hash, avoiding redundant re-hashing.

## Test plan

- [x] data-model tests pass (27/27)
- [x] runner tests pass (234/234)
- [x] Internal review by Flux (in progress)

### DIFF
--- a/packages/data-model/interface.ts
+++ b/packages/data-model/interface.ts
@@ -289,9 +289,3 @@ export interface ExperimentalDataModelConfig {
    *  (bigint, Map, Set, Uint8Array, Date, etc.). */
   modernDataModel: boolean;
 }
-
-// ===========================================================================
-// Schema path selector (canonical definition in @commontools/api)
-// ===========================================================================
-
-export type { SchemaPathSelector } from "@commontools/api";
--- a/packages/data-model/schema-hash-legacy.ts
+++ b/packages/data-model/schema-hash-legacy.ts
@@ -10,8 +10,7 @@
  * replaced by canonical hashing (via `modernHash`) behind a flag.
  */
 
-import type { JSONSchema } from "@commontools/api";
-import type { SchemaPathSelector } from "./interface.ts";
+import type { JSONSchema, SchemaPathSelector } from "@commontools/api";
 
 const _hashCache = new WeakMap<object, string>();
 
--- a/packages/data-model/schema-hash-modern.ts
+++ b/packages/data-model/schema-hash-modern.ts
@@ -10,8 +10,7 @@
  * `value-hash-modern.ts` and `fabric-value-modern.ts`.
  */
 
-import type { JSONSchema } from "@commontools/api";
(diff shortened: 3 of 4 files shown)


packages/data-model/interface.ts
```diff
@@ -289,3 +289,17 @@ export interface ExperimentalDataModelConfig {
...
+// ===========================================================================
+
+/**
+ * Selects a sub-path within a document, optionally paired with a schema
+ * that describes the value at that path. Used by the storage/sync layer
+ * to track which slices of a document are being observed.
+ */
+export type SchemaPathSelector = {
+  path: readonly string[];
+  schema?: import("@commontools/api").JSONSchema;
```
danfuzz: I think the fact that we need to pull in `JSONSchema` here like this is an indicator that this type declaration would be better placed in `api`.

packages/data-model/interface.ts
```diff
@@ -289,3 +289,9 @@ export interface ExperimentalDataModelConfig {
    *  (bigint, Map, Set, Uint8Array, Date, etc.). */
   modernDataModel: boolean;
 }
+
+// ===========================================================================
```
danfuzz: This can just be removed entirely. `api` is where schema stuff "naturally" lives anyway.

PR #3205, opened 2026-03-31
feat(fuse): [FS] symbol — per-pattern filesystem projection CT-1417 CT-1420 CT-1421 CT-1418 CT-1419 CT-1403

bfollington:
## Summary

- **`[FS]` symbol**: patterns can now declare their own on-disk representation via `[FS]: { type, content, frontmatter? }` in their return object. When present, FUSE renders a single `index.md` (YAML frontmatter + markdown body) or `index.json` at the piece root instead of an exploded `result/` directory.
- **Live write-back**: edits to `index.md`/`index.json` parse the content and update the corresponding reactive cells bidirectionally.
- **`.handlers` file**: every piece now gets a dot-prefixed `.handlers` file at its root listing all callables with TypeScript-ish input types (e.g. `setTitle.handler  string`, `createNewNote.handler  void`). Hidden from `ls`, readable with `cat`.
- **Callable scripts enriched**: each `.handler`/`.tool` script file now embeds `# schema:` and `# input:` comments so agents can discover the expected shape via `cat` or `head` without invoking.
- **`ct exec` UX**: `--help` output now includes an "Input type:" section showing the TypeScript-ish type shape; calling a handler with no args now shows the expected type in the error message.
- **`note.tsx`**: first pattern to adopt `[FS]`, exporting `title` and `content` cells through markdown projection.

## Test plan

- [x] `deno test --allow-all packages/fuse/` — all 69 tests pass
- [x] `deno test --allow-all packages/cli/test/exec.test.ts` — all tests pass
- [x] Mount a space with `ct fuse mount`, open a note piece → `index.md` at piece root, no `result/` dir
- [x] `cat pieces/<Note>/index.md` → YAML frontmatter with `entityId` + `title`, markdown body
- [x] Edit and save `index.md` → title/content cells update in browser UI
- [x] `cat pieces/<Note>/.handlers` → TypeScript-ish types for each handler
- [x] `cat pieces/<Note>/setTitle.handler` → shebang with `# schema:` and `# input: string` comments
- [x] `./setTitle.handler` (no args) → error showing `Handler requires input. Expected type: string`
- [x] `./setTitle.handler --help` → usage with "Input type: string" section
- [x] Non-FS pieces → `result/` dir still present; `.handlers` file present

### DIFF
--- a/packages/fuse/stat.ts
+++ b/packages/fuse/stat.ts
@@ -1,0 +1,65 @@
+import {
+  DIR_MODE,
+  FILE_MODE,
+  FILE_MODE_RW,
+  FILE_MODE_RWX,
+  FILE_MODE_RX,
+  type StatOpts,
+  SYMLINK_MODE,
+} from "./platform.ts";
+import type { FsNode } from "./types.ts";
+
+export interface MountOwnership {
+  uid: number;
+  gid: number;
+}
+
+interface OwnershipProvider {
+  uid?: () => number | null;
+  gid?: () => number | null;
+}
+
+export function getMountOwnership(
+  provider: OwnershipProvider = Deno,
+): MountOwnership {
+  const uid = typeof provider.uid === "function" ? provider.uid() : null;
+  const gid = typeof provider.gid === "function" ? provider.gid() : null;
+  return {
+    uid: uid ?? 0,
+    gid: gid ?? 0,
+  };
+}
+
+export function nodeMode(node: FsNode, isWritable = false): number {
+  if (node.kind === "dir") return DIR_MODE;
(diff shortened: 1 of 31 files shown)


packages/fuse/stat.ts
```diff
@@ -0,0 +1,74 @@
...
+export function getMountOwnership(
+  provider: OwnershipProvider = Deno,
+): MountOwnership {
+  const uid = safeCall(provider.uid);
+  const gid = safeCall(provider.gid);
+  return {
+    uid: uid ?? 0,
+    gid: gid ?? 0,
+  };
+}
```
Hixie, line 40: this seems scary. wouldn't it be better to fail outright if we couldn't read the uid/gid?

PR #3210, opened 2026-03-31
refactor: rename schema hash functions — AsString suffix for string-returning variants

danfuzz:
## Summary

Naming cleanup and encapsulation improvements across `packages/data-model`
and `packages/api`: schema hash function renames, FabricHash field
privatization, and systematic removal of legacy "Rich"/"rich data"
terminology.

### Changes

- **Schema hash function renames** (`schema-hash-legacy.ts`,
  `schema-hash-modern.ts`, `schema-hash.ts`, `schema-hash.test.ts`):
  `FabricHash`-returning variants get plain names, `string`-returning
  variants get the `AsString` suffix. Modern string path now uses
  `hashOfModernAsString()` instead of `hashOfModern().toString()`.

- **FabricHash.hash privatization** (`fabric-hash.ts` + call sites):
  `.hash` field replaced with private `#hash`. Public access via `.bytes`
  (returns a defensive copy via `.slice()`) and `.copyInto()`. All call
  sites updated.

- **FabricHash.algorithmTag renamed to .tag** (`fabric-hash.ts`,
  `api/index.ts`): Private `#tag` field + `.tag` getter. API interface
  updated to match.

- **Rich → Modern rename** (12 data-model files): Cleaned up all
  remaining "Rich"/"rich data" references — variable names, comments,
  doc strings — to use "Modern"/"modern" consistently.

- **Review fixes**: `.bytes` returns a defensive copy, defunct test
  comments removed, unnecessary `Array.from()` defensive copies removed
  (`.bytes` already returns fresh copies), `.toBe()` changed to
  `.toEqual()` for unshared byte arrays, `TODO(@danfuzz)` added for copy
  performance in `value-hash-modern.ts`, stale `.hash` docstring
  reference fixed.

21 files changed, 144 insertions, 131 deletions (6 commits).

### Performance baseline

Updating because I think this is a spurious result, in that the only runtime-effective code changed in this PR is hidden behind experiment flags. Also, I ran the perf-failed step twice and got vastly different numbers.

NEW_PERF_BASELINE: step: patterns integration = 156s

Co-Authored-By: upstream-liaison-bolt (Claude Opus 4.6) <noreply@anthropic.com>

### DIFF
--- a/packages/data-model/value-hash-modern.ts
+++ b/packages/data-model/value-hash-modern.ts
@@ -416,7 +416,7 @@ function hashOfModernInternal(
 /**
  * Compute the canonical SHA-256 hash of a `FabricValue`. Returns a
  * `FabricHash` with algorithm tag `fid1` (fabric ID, v1).
- * The caller (`hashOf()`) extracts the raw digest via `.hash` for
+ * The caller (`hashOf()`) extracts the raw digest via `.bytes` for
  * `Reference.fromDigest()`.
  *
  * Caches results for primitives (LRU) and deep-frozen objects (WeakMap).


packages/data-model/fabric-hash.ts
```diff
@@ -50,17 +52,17 @@ export class FabricHash extends FabricPrimitive {
    * TODO(danfuzz): Remove after canonical hashing flag graduates.
    */
   get "/"(): Uint8Array {
-    return this.hash; // TODO(@danfuzz): `hash` should not be exposed.
+    return this.#hash;
```
danfuzz: This is wrong! It exposes the internal necessarily-mutable `Uint8Array`.

danfuzz: Will fix — `.bytes` will return a copy via `.slice()`.

PR #3213, opened 2026-03-31
fix(ui): preserve pattern-assigned schema in CellController.bind()

jkomoros:
## Summary
- When a CellHandle already has a schema (assigned by the pattern compiler via `$images`/`$files` bindings), `CellController.bind()` was overriding it with the component's own schema via `asSchema()`. This created a divergent cell view — the component wrote to one schema projection while the pattern read from another, causing uploaded data to silently vanish.
- Fixes aisle photo import in `store-mapper` pattern (and any other pattern using `$images`/`$files` bindings with `ct-image-input`/`ct-file-input`).

## What changed
- `packages/ui/src/v2/core/cell-controller.ts`: Skip `asSchema()` when the incoming CellHandle already has a schema, so writes from the component go directly to the pattern's cell.
- `packages/ui/src/v2/core/cell-controller.test.ts`: Updated existing test, added new test for schema preservation.

## Test plan
- [x] Existing `cell-controller.test.ts` tests pass (48/48)
- [x] New test verifies schema preservation when handle already has one
- [x] Manually verified `store-mapper` photo import works end-to-end (upload → AI analysis → aisle detection → Add)

### DIFF
--- a/packages/ui/src/v2/core/cell-controller.ts
+++ b/packages/ui/src/v2/core/cell-controller.ts
@@ -147,6 +147,11 @@ export class CellController<T> implements ReactiveController {
         this._currentValue.equals(value))
     ) {
       this._cleanupCellSubscription();
+      // Only apply the component's schema when the CellHandle doesn't already
+      // have one. Pattern-compiled $bindings (e.g. $images, $files) arrive with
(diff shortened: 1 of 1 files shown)


bfollington (approved):
Approve

PR #3226, opened 2026-04-03
feat(fuse): lazy piece content, normalized names, read-only dirs, auto-step after exec CT-1437

bfollington:
## Summary

Four FUSE improvements from the CT-1370 round 5 pass:

- **Lazy piece content hydration** — `loadPieceTree` no longer eagerly fetches input/result data for every piece on space connect. Instead, `input/` and `result/` are materialised on-demand when the kernel issues a `lookup` or `readdir` for them. This avoids the N×2 blocking fetches that slowed down initial mount, especially for large spaces. Hydration state is tracked per-piece-inode and invalidated on write so subsequent reads always reflect current data. (`cell-bridge.ts`, `mod.ts`)

- **Normalized projected piece names** — Piece directory names in the FUSE tree are now normalized: Unicode marks stripped, emoji removed, non-alphanumeric runs collapsed to `-`, leading/trailing dashes trimmed. Falls back to a normalized piece ID for symbol-only names (e.g. `🔥✨` → `of-emoji-piece`). The original raw name is preserved in `meta.json`. (`cell-bridge.ts`)

- **Read-only projected directories** — FUSE `getattr`/`lookup` now marks piece and piece-prop directories as mode `0o555` (no write bit) so the OS rejects accidental writes to directories rather than silently failing. (`mod.ts`)

- **Auto-step after `ct exec`** — Handler invocations via `ct exec` now automatically call `piece.getCell().pull()` + `manager.synced()` after execution, triggering reactive recomputation without a separate `ct piece step`. (CT-1437, `packages/cli/lib/exec.ts`)

## Test plan

- [x] `deno test packages/fuse/` — all cell-bridge and tree-builder tests pass
- [x] Mount a space: `ct fuse mount /tmp/ct-test --background`, verify pieces appear with normalized names (spaces → dashes)
- [x] Open a piece directory — confirm `input/` and `result/` are absent until accessed, then appear after `cat result/summary`
- [x] Call a handler via FUSE: verify piece count suffix updates without needing `ct piece step`
- [x] Call a handler via `ct exec`: verify computed state updates immediately (CT-1437)
- [ ] Attempt `mkdir` inside a piece directory — should fail with EPERM

### DIFF
--- a/packages/fuse/cell-bridge.ts
+++ b/packages/fuse/cell-bridge.ts
@@ -2270,7 +2270,7 @@ export class CellBridge {
     // reads when setting up reactive subscriptions.
     try {
       const nameTrackingCell = await piece.result.getCell();
-      const cancelRootSub = nameTrackingCell.sink((_newValue: unknown) => {
+      const cancelRootSub = nameTrackingCell.sink((newValue: unknown) => {
         setTimeout(() => {
           try {
             const state = this.spaces.get(spaceName);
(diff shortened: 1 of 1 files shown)


skills/fuse-workflow/SKILL.md
```diff
@@ -233,6 +233,364 @@ deno task cf fuse mount /tmp/cf
...
+  pattern — there are 4 dated events in the Work Journal"_
+- `"todo"` — flag incomplete work: _"Contact Book has 3 entries with no email
+  address"_
+
+**Mark a wish resolved** (when fulfilling another agent's annotation):
+
+```bash
+echo '"resolved"' | ct piece set --piece $WISH_ID status --space SPACE
+ct piece step --piece $WISH_ID --space SPACE
+```
```
bfollington: This is not really the right place for this, the fuse-workflow skill is just about working with fuse but this is a specific interaction pattern for agents with fuse (suggests we need another skill, fuse-agent probably) 

Also applies to the below annotations listing and the activity log.

skills/fuse-workflow/SKILL.md
```diff
@@ -233,6 +233,364 @@ deno task cf fuse mount /tmp/cf
...
+"MOUNT/SPACE/pieces/Contact Book/result/onAddContact.handler"
+```
+
+Use `--value null` only if you specifically need the older explicit form.
+
+**NEVER redirect to handler files (CT-1417):**
+
+```bash
+echo '{"title":"x"}' > piece.handler  # SILENTLY FAILS — write does nothing
+```
```
bfollington: Can this be fixed?

PR #3240, opened 2026-04-06
feat(transformer): tighten ts-transformer input schema shrinking

seefeldb:
## Summary
- tighten ts-transformer input schema shrinking so derived actions read only the specific data they actually use instead of broad receiver shapes
- cover the main bad cases that were widening or mis-shrinking inputs: `??` / `||`, `for...of`, direct `.find(...)?prop` reads, fixed computed keys like `[NAME]` / `[UI]` / `[SELF]`, array element path shrinking, expression-bodied `.get()` callbacks, and identity-only equality that overlaps a full-shape read
- add regression coverage for the shrinker and a local profiling mode in the default-app integration test so note creation can be compared against `main`

## Why
The main goal of this PR was to reduce how much data actions need to read at runtime by making the inferred input schemas narrower and more accurate. The review fixes all stayed in service of that same goal: if the shrinker misses a path, it can break correctness; if it broadens unnecessarily, the runner ends up traversing and reading far more than the action actually needs.

## Impact
- derived schemas now stay path-specific in more real cases instead of collapsing to wildcard or full-shape reads
- direct expression-bodied callbacks like `derive(value, (v) => v.get())` still preserve cell-wrapper semantics during schema injection
- overlapping identity/full-shape cases such as `equals(input.people, input.people.filter(...).slice(0))` now keep the full receiver shape instead of collapsing back to identity-only
- the added profiling hook in `packages/patterns/integration/default-app.test.ts` makes it easier to compare note-creation behavior on `main` versus a feature branch without changing product code

## Performance
- the most important measured success is the one this PR was targeting: at the 30th note, `runner action/readInputs` dropped from `7444.7` on `main` to `344.0` on this branch, a `95.4%` reduction
- end-to-end note creation also improved at the 30th note from `2397.318 ms` on `main` to `2047.156 ms` on this branch, about `350 ms` (`14.6%`) faster, with create-to-view improving from `1658.461 ms` to `1263.062 ms`
- the worst settle wave at the 30th note dropped from `1326.0 ms` / `109` actions on `main` to `1005.8 ms` / `78` actions on this branch
- the rest of the profile looks more like follow-up guidance than the primary headline: scheduler and settle work are still the biggest buckets, even after the read-input win, so future optimization work should likely focus there
- measured with `CF_NOTE_CREATE_TIMING_SERIES=30 CF_CAPTURE_NOTE_CREATE_PROFILE_SERIES=30 deno task integration patterns default-app` on `main` and on this branch

## Validation
- `deno test -A packages/ts-transformers/test/policy/capability-analysis.test.ts`
- `deno test -A packages/ts-transformers/test/schema-shrink-validation.test.ts`
- `deno test -A packages/schema-generator/test/schema/capability-wrapper-types.test.ts`
- `deno task test` in `packages/ts-transformers`
- `deno task test` in `packages/schema-generator`
- `CF_NOTE_CREATE_TIMING_SERIES=30 CF_CAPTURE_NOTE_CREATE_PROFILE_SERIES=30 deno task integration patterns default-app` on `main` and on this branch using local instrumentation in `packages/patterns/integration/default-app.test.ts`
- `deno fmt`
- `deno lint`

### DIFF
--- a/packages/ts-transformers/src/policy/capability-analysis.ts
+++ b/packages/ts-transformers/src/policy/capability-analysis.ts
@@ -1254,6 +1254,26 @@ export function analyzeFunctionCapabilities(
       }
     };
 
+    const visitScopedLocalCollectionBindings = (visitor: () => void): void => {
+      const savedLocalArrayElementBindings = new Map(
+        localArrayElementBindings,
+      );
+      const savedLocalMapValueBindings = new Map(localMapValueBindings);
+      const scopedLocalCollectionNames = new Set<string>();
+      scopedLocalCollectionNamesStack.push(scopedLocalCollectionNames);
+
+      try {
+        visitor();
+      } finally {
+        scopedLocalCollectionNamesStack.pop();
+        restoreScopedLocalCollectionBindings(
+          scopedLocalCollectionNames,
+          savedLocalArrayElementBindings,
+          savedLocalMapValueBindings,
+        );
+      }
+    };
+
     const recordLocalArrayPush = (
       arrayName: string,
       args: readonly ts.Expression[],
@@ -2078,6 +2098,15 @@ export function analyzeFunctionCapabilities(
             savedLocalMapValueBindings,
           );
         }
+        return;
+      }
+
+      if (ts.isBlock(node)) {
+        visitScopedLocalCollectionBindings(() => {
+          for (const statement of node.statements) {
+            visit(statement);
+          }
+        });
         return;
       }
 
--- a/packages/ts-transformers/test/policy/capability-analysis.test.ts
+++ b/packages/ts-transformers/test/policy/capability-analysis.test.ts
@@ -783,6 +783,90 @@ Deno.test(
 );
 
 Deno.test(
+  "Capability analysis restores local array bindings after block shadowing",
+  () => {
+    const { program, sourceFile } = createProgramWithSource(
+      `
+      interface Array<T> {
+        push(...items: T[]): number;
+        sort(compareFn?: (a: T, b: T) => number): T[];
+      }
+
+      type Candidate = { id: string; name: string; unused: string };
+
+      const fn = (input: { candidates: Candidate[]; other: Candidate[] }) => {
+        const results: Candidate[] = [];
+        for (const candidate of input.candidates) {
+          results.push(candidate);
+        }
+        if (input.other.length > 0) {
+          const results: Candidate[] = [];
+          for (const candidate of input.other) {
+            results.push(candidate);
+          }
+        }
+        results.sort((left, right) => left.name.localeCompare(right.name));
+        return results.length;
+      };
+      `,
+    );
+    const summary = analyzeFunctionCapabilities(
+      findArrowByVariableName(sourceFile, "fn"),
+      { checker: program.getTypeChecker() },
+    );
+    const input = getPaths(summary, "input");
+
+    assertEquals(input.wildcard, false);
+    assert(input.readPaths.includes("candidates"));
+    assert(input.readPaths.includes("candidates.0"));
+    assert(input.readPaths.includes("candidates.0.name"));
+    assert(input.readPaths.includes("other"));
+    assert(input.readPaths.includes("other.length"));
+    assert(input.readPaths.includes("other.0"));
+    assertEquals(input.readPaths.includes("other.0.name"), false);
+  },
+);
+
+Deno.test(
+  "Capability analysis restores local map bindings after block shadowing",
+  () => {
+    const { program, sourceFile } = createProgramWithSource(
+      `
+      type Candidate = { id: string; name: string; unused: string };
+
+      const fn = (input: { candidates: Candidate[]; other: Candidate[] }) => {
+        const lookup = new Map<string, Candidate>();
+        for (const candidate of input.candidates) {
+          lookup.set("outer", candidate);
+        }
+        if (input.other.length > 0) {
+          const lookup = new Map<string, Candidate>();
+          for (const candidate of input.other) {
+            lookup.set("inner", candidate);
+          }
+        }
+        return lookup.get("outer")?.name;
+      };
+      `,
+    );
+    const summary = analyzeFunctionCapabilities(
+      findArrowByVariableName(sourceFile, "fn"),
+      { checker: program.getTypeChecker() },
+    );
+    const input = getPaths(summary, "input");
+
+    assertEquals(input.wildcard, false);
+    assert(input.readPaths.includes("candidates"));
+    assert(input.readPaths.includes("candidates.0"));
+    assert(input.readPaths.includes("candidates.0.name"));
+    assert(input.readPaths.includes("other"));
+    assert(input.readPaths.includes("other.length"));
+    assert(input.readPaths.includes("other.0"));
+    assertEquals(input.readPaths.includes("other.0.name"), false);
+  },
+);
+
+Deno.test(
   "Capability analysis keeps element bindings through filter-to-map chains",
   () => {
     const { program, sourceFile } = createProgramWithSource(


mathpirate:
ok i think that's everything i'm gonna find in this initial review pass. here's my codex's summary of what it thinks we should do next in light of the above findings, up to you/your codex to evaluate berni:

The third bug points to a second missing test shape: cell-backed dynamic element access. There is already a good non-opaque fixture in element-access-simple.expected.jsx, but the opaque/cell-backed variants were only covered by snapshot churn, not by a semantic assertion that both bindings remain in the emitted derive input schema. A small addition to schema-shrink-validation.test.ts asserting that items.get()[index.get()] yields an input schema containing both items and index would have caught it.

One useful meta-point: validateShrinkCoverage (line 1262) cannot catch the dropped-items bug, because it only validates paths the analyzer already kept. So I would not spend time trying to solve this class with more diagnostics first. Better leverage is:

add the 3 missing counterexample tests above
optionally add one runtime-style transformed-source assertion in runtime-style.test.ts for a destructured derive callback receiving all expected bindings

seefeldb:
@cubic-dev-ai review

packages/ts-transformers/src/transformers/type-shrinking.ts
```diff
@@ -976,6 +1622,416 @@ export function wrapTypeNodeWithCapability(
...
+    }
+    const unionNode = factory.createUnionTypeNode(members);
+    ensureTypeNodeRegistered(unionNode, checker, typeRegistry);
+    return unionNode;
+  }
+
+  return node;
+}
+
 function wrapTypeNodeWithDefault(
```
mathpirate, line 2035: this comment is roughly intended to go on line 2073, `getPropertyNameText(member.name)`, but that doesn't show up in the github diff UI:

I don’t think the computed-key default case is fully fixed yet. collectDestructureBindings() records defaults for computed keys as string paths like $NAME, but applySingleDefaultToTypeNode() still matches members with getPropertyNameText(member.name) without a checker. After the property-name.ts refactor, computed names like [NAME] only resolve through checker-backed getComputedPropertyKeyInfo(...), so this lookup now misses and the default wrapper gets dropped.

I reproduced this on the branch with:

`derive({ piece }, ({ piece: { [NAME]: name = "untitled" } }) => name)`

The emitted input schema keeps $NAME as plain type: "string" instead of wrapping it in Default<...>, so the destructuring default is no longer preserved in the schema.

mathpirate:
I did another full pass on this after the latest updates. Aside from the one comment I already left on the computed-key destructuring default case, I did not find any other independent issues of my own.

On cubic’s April 9, 2026 review batch: I would pay attention to the `for...of` capability-analysis finding and the two `nested-writable-pattern-branches` under-shrunk fixture findings. cubic found those, and they look real to me; I also confirmed the `for...of` one locally with a small repro.

I would *not* treat the `OpaqueRef` comment as actionable as written, since it seems to conflict with the branch’s intended behavior and explicit regression test (`OpaqueRef<T>` erases to `T` without `asOpaque`). I also think cubic’s computed-key destructuring comment is probably a false positive / overstatement; I checked a `const KEY = "title" as const` case locally and it still narrowed correctly.

The remaining cubic findings I’d put in the “plausible but not yet independently confirmed” bucket rather than acting on immediately.

ubik2:
Looks like I'll have some conflicts here regarding opaque cell handling with https://github.com/commontoolsinc/labs/pull/3253

seefeldb:
> Looks like I'll have some conflicts here regarding opaque cell handling with #3253

Oh where? It's newly generating proper OpaqueCell<unknown> in some cases, but other than that I don't think it should conflict. In any case, merge first!

seefeldb:
Codex:

I validated the four high-signal review items and fixed the valid ones.

The for...of finding was real. In capability-analysis.ts, once the iterable binding resolved, chained call expressions like .filter(...) were no longer being visited, so nested callback reads were dropped. The fix keeps the resolved binding behavior but still visits call-based iterable expressions so callback reads survive.

The two nested-writable-pattern-branches findings were also real, and they shared one root cause. After callback lowering, direct uses like section.key("title") and task.key("label") were not being counted as reads unless they were followed by another operation. That meant the outer sections.map(...) schema kept only tasks, and the inner tasks.map(...) schema kept only tags. I fixed that in capability-analysis.ts by treating static key(...) calls as reads when the key result is used as a value, while still avoiding double-counting chained cases like .key(...).get().

@mathpirate ’s computed-key default comment was valid, but specifically on transformed destructuring callbacks. In type-shrinking.ts, default application was matching computed member names without a checker, so [NAME]-style members missed their defaults. I fixed that by threading the checker through default application, and in pattern-callback-transform.ts I also make registered summaries retain defaulted paths explicitly. I added regressions in capability-analysis.test.ts, pipeline-regressions.test.ts, and schema-shrink-validation.test.ts, and refreshed nested-writable-pattern-branches.expected.jsx.

packages/schema-generator/src/formatters/common-fabric-formatter.ts
```diff
@@ -914,12 +919,11 @@ export class CommonFabricFormatter implements TypeFormatter {
     schema: JSONSchemaMutable,
     wrapperKind: WrapperKind,
   ): JSONSchemaMutable {
-    // OpaqueRef is just T — no wrapper semantics needed
-    if (wrapperKind === "OpaqueRef") {
-      return schema;
-    }
-
-    const propertyName = wrapperKind === "Stream" ? "asStream" : "asCell";
+    const propertyName = wrapperKind === "Stream"
```
ubik2: this conflicts with https://github.com/commontoolsinc/labs/pull/3253, but I think the 3253 version can just replace this once one is landed.

packages/schema-generator/src/formatters/common-fabric-formatter.ts
```diff
@@ -247,8 +247,9 @@ export class CommonFabricFormatter implements TypeFormatter {
         };
         if (wrapperKind === "Cell") {
           itemsOverride.asCell = true;
+        } else if (wrapperKind === "OpaqueRef") {
+          itemsOverride.asOpaque = true;
```
ubik2: Both this and the later test on line 378 don't do anything special with OpaqueRef or OpaqueCell in https://github.com/commontoolsinc/labs/pull/3253.
The itemsOverride schema will limit the traversal to be shallow, but since we know we're `isArrayPropertyOnlyAccess` here, we could also do `{ asCell: ["opaque"] }` so that we block all that off.

Edit: the version on line 378 does this with `{ type: "unkown" }`, but `{ asCell: ["opaque"] }` expresses even less interest in the contents -- we don't read *anything* from those.

ubik2:
> > Looks like I'll have some conflicts here regarding opaque cell handling with #3253
> 
> Oh where? It's newly generating proper OpaqueCell in some cases, but other than that I don't think it should conflict. In any case, merge first!

There's nothing too bad in the conflicts. Just that we're both changing how we're marking these opaque cells.

I've landed my PR, and I imagine the agents can resolve the conflict automatically, but if not, I'm happy to touch up these spots.

mathpirate:
I re-reviewed the latest Monday head (`70fd81b4`), and the Friday wrapper expectation fallout looks fixed. I still found one remaining issue in `packages/ts-transformers/src/policy/capability-analysis.ts`.

`clearScopedLocalCollectionBindingNames(node.name)` now runs for every variable declaration, but the new save/restore stack is only pushed for nested callbacks and `for...of` handling. In an ordinary block, a shadowed local collection name can still clear the outer binding with no matching restore.

I reproduced this with an outer `results` array populated from `input.candidates`, an inner `if` block with shadowed `results` populated from `input.other`, then `results.sort((left, right) => left.name.localeCompare(right.name))`. The capability summary records `other.0.name` and drops `candidates.0.name`, so later shrinking follows the wrong source.

mathpirate:
Following up on my earlier comment: I checked whether this is array-specific, and it isn't. The same plain block-scope shadowing leak also reproduces for local `Map#set/get` tracking.

A minimal shape is: outer `lookup` stores `input.candidates.at(0)`, inner `if` block shadows `lookup` and stores `input.other.at(0)`, then later `lookup.get("outer")?.name` is analyzed against `other.0.name` instead of `candidates.0.name`.

So after the fix, I think the important thing is to make sure ordinary lexical block scopes restore both `localArrayElementBindings` and `localMapValueBindings`, not just the nested-callback / `for...of` cases covered by the current stack handling.

PR #3241, opened 2026-04-06
Follow up memory v2 scope notes and review invariants

seefeldb:
## Summary
- add explicit entity-envelope worked examples for `{}` vs missing `value` vs explicit delete in the memory v2 data model spec
- mark blob delivery and merge/lifecycle sections more clearly as deferred next-layer scope rather than current-pass acceptance bar
- rewrite the v2 server watch tests to assert touched-entity semantics instead of pinning internal `evaluateGraphQuery` behavior

## Context
This is a stacked follow-up to #3083 aimed at the remaining open Will Kelly review points that were still substantive on the branch.

## Validation
- `deno test -A packages/memory/test/v2-server-test.ts`
- `deno lint packages/memory/test/v2-server-test.ts`

---
## Summary by cubic
Clarifies Memory v2 semantics for `{}` vs missing `value` vs explicit delete, and narrows MVP scope by deferring blob delivery and merge/lifecycle features. Updates server watch tests to assert touched-entity sync behavior instead of relying on internal query evaluation.

- **Refactors**
  - Rewrote watch tests to validate that only touched entities are synced on updates and when adding watches.
  - Removed internal `evaluateGraphQuery` instrumentation; assertions now check emitted effects and IDs.

<sup>Written for commit 20eeef02e3c200176947bfca52d15c744dee1c28. Summary will update on new commits.</sup>

### DIFF
--- a/docs/specs/memory-v2/01-data-model.md
+++ b/docs/specs/memory-v2/01-data-model.md
@@ -75,10 +75,21 @@ conventions over that same stored document:
 **Deletion is explicit**: removing an entity is represented only by a `Delete`
 fact (section 2.2). A missing `value` field is NOT used as a tombstone in v2.
 Supporting a live "undefined" cell value is deferred; if we need it later, it
-should use an explicit sentinel rather than overloading deletion semantics. An
-empty JSON object `{}` is still an ordinary live value, not a delete. A marked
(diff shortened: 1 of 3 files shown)


seefeldb:
Superseded by #3243 now that #3083 merged into main.

PR #3253, opened 2026-04-08
Rework asCell to encompass multiple types

ubik2:
This plumbs most of the cell types through the asCell, and allows for things like `Cell<Cell<T>>`.
While this should get us to things like OpaqueCell, ReadonlyCell, WriteonlyCell, and ComparableCell when we do a `cell.get()`, it doesn't include any extra work to ensure that we don't read or write to the transaction in the surrounding code, so it may not automatically improve reactivity.

---
## Summary by cubic
Switches schema wrapper semantics to `asCell: [...]` arrays across the stack to support nested wrappers (e.g., `["stream","cell"]`), replaces legacy `asStream`/`asOpaque`, and hardens opaque behavior so traversal never descends into opaque cells or follows redirects/tx reads.

- **Refactors**
  - `asCell` now accepts `readonly AsCellType[]`; exported `AsCellType` from `@commonfabric/api`; legacy booleans normalized; removed `asOpaque` handling and tests.
  - Wrapper helpers consolidated via `ContextualFlowControl.getAsCellValues()`; internal key filtering updated; CLI schemaless handler detection accepts `asCell: ["stream"]`.
  - Runtime/cell: peel outer wrapper at creation and pass the rest to children; `cell.key()` returns correct `CellKind`; link export options unified with `SanitizeSchemaForLinksOptions`.
  - Link/schema sanitization: `keepAsCell` preserves full wrapper arrays; `keepStreams` keeps only streams; strips markers for link serialization; handles cycles safely; opaque cells preserve original links and skip redirect/tx reads (client and server); memory query now excludes linked opaques; added tests including diff/update paths for opaques.
  - Logical merges preserve wrapper chains; improved `anyOf`/`allOf` merges; `mergeMatches` now requires a `JSONSchemaObj`.
  - Formatting/types: `schemaToTypeString` renders nested wrappers (e.g., `["stream","cell"]` → `(Cell<T>) => void`) even at depth limits; generator emits `asCell: [...]` including `void` → `["opaque"]` and `Stream<void>` → `["stream","opaque"]`; supports brands `readonly`, `writeonly`, `comparable`, `opaque`; minor type cleanups.
  - Updated built-ins, docs, transformers, fixtures, and tests to use `asCell: [...]` (replacing `asStream: true`/`asCell: true`).

- **Migration**
  - Replace: `asCell: true` → `asCell: ["cell"]`, `asStream: true` → `asCell: ["stream"]`, `asOpaque: true` → `asCell: ["opaque"]`.
  - Use `ContextualFlowControl.getAsCellValues(schema)` for wrapper detection.
  - For link/schema export, pass `keepAsCell: true` to preserve arrays, or `keepStreams: true` to keep only streams.

<sup>Written for commit dad4159ebab277f7ea3ecb7be376266f33939606. Summary will update on new commits.</sup>

### DIFF
--- a/packages/cli/lib/piece.ts
+++ b/packages/cli/lib/piece.ts
@@ -276,7 +276,9 @@ export async function newPiece(
         timer = setTimeout(() => {
           reject(
             new Error(
-              `Piece created but failed to start within ${PIECE_START_TIMEOUT_MS / 1000}s. ` +
+              `Piece created but failed to start within ${
+                PIECE_START_TIMEOUT_MS / 1000
+              }s. ` +
                 `Check toolshed logs for runtime errors.`,
             ),
           );


packages/runner/src/schema.ts
```diff
@@ -115,19 +118,23 @@ export function processDefaultValue(
...
-    asStream = resolvedSchema.asStream === true;
-    if (
-      resolvedSchema.asCell !== undefined ||
-      resolvedSchema.asStream !== undefined
-    ) {
-      const { asCell: _asCell, asStream: _asStream, ...restSchema } =
-        resolvedSchema;
-      resolvedSchema = restSchema;
-    }
+  if (!isRecord(resolvedSchema)) {
```
ubik2: I moved the !isRecord test up here,  so we don't have to keep doing the isRecord in our tests later.
It does mean we end up mentioning the annotate call both here and at the end if we fall through all the other tests.

packages/runner/src/schemas.ts
```diff
@@ -189,27 +188,38 @@ export const vnodeSchema = toDeepFrozenSchema(
...
+            {
+              type: "object",
+              properties: {}, // stop query from descending
+            },
+            {
+              type: "array",
+              items: { type: "null" }, // stop query from descending
+            }, // this was generated, but is a bit problematic to have
+            //    both cell and stream, since both will always match.
+            {
```
ubik2: Added this comment because a `Cell<any>` is a bit of an issue in the Props. It's nice to have the flexibility, but it means we'll always just use that match instead of all the others. It will match anything, and cell matches take priority over other matches.
This is fallout of the types, but we could handcode this schema to remove all the anyOf options. That's not part of this PR, though

packages/schema-generator/test/schema/capability-wrapper-types.test.ts
```diff
@@ -22,14 +22,15 @@ describe("Schema: Capability wrapper types", () => {
 
     expect(ro).toBeDefined();
     expect(ro.properties?.foo?.type).toBe("string");
-    expect(ro.asCell).toBe(true);
+    expect(ro.asCell).toEqual(["readonly"]);
```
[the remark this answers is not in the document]
ubik2: Not sure why cubic was confused, but it clearly is. Resolving.

seefeldb (approved):

PR #3271, opened 2026-04-09
[codex] cleanup: normalize auth-family conditionals

mathpirate:
## Summary
- replace authored `ifElse(...)` conditionals with plain ternaries across the mirrored auth-family UIs
- normalize JSX-child control flow, inline labels, local JSX/string aliases, returned `userChip` fields, and the shared auth-manager `fullUI` composition
- update the stale auth-manager comment so it describes the current nested-ternary composition rather than chained `ifElse(...)`

## Why
Current transformer behavior already lowers these ternary forms correctly in all of the positions used here. These auth files were still carrying older helper-oriented style that is no longer necessary.

## Scope
- `packages/patterns/google/core/google-auth.tsx`
- `packages/patterns/airtable/core/airtable-auth.tsx`
- `packages/patterns/auth/create-auth-manager.tsx`

## Validation
- `deno fmt packages/patterns/google/core/google-auth.tsx packages/patterns/airtable/core/airtable-auth.tsx packages/patterns/auth/create-auth-manager.tsx`
- `deno task cf check packages/patterns/google/core/google-auth.tsx packages/patterns/airtable/core/airtable-auth.tsx --no-run`
- `deno check packages/patterns/auth/create-auth-manager.tsx`
- `git diff --check`

### DIFF
--- a/packages/patterns/airtable/core/airtable-auth.tsx
+++ b/packages/patterns/airtable/core/airtable-auth.tsx
@@ -2,7 +2,6 @@ import {
   computed,
   Default,
   handler,
-  ifElse,
   NAME,
   pattern,
   safeDateNow,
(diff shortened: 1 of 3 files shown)


bfollington:
In practice `gmail-importer.tsx` fails, seems pre-existing though since you didn't touch it?

```
Error in action: TypeError: AuthPattern is not a function 
  pieceId: baedreibggfvr3n7psmne5bmk7uo3wwzvpuaravh52s7ixc46ay2qt2umza 
  space: did:key:z6MkvFBoXV8moEbv8H6vztFSjAV2Mrk1XcZuLgMCQCmFUciK 
  patternId: ba4jcbub7d5kr5svxkazukain4qbfmpyiyjjg5m6tfdc3cevigd4b5z3u 
TypeError: AuthPattern is not a function
    at eval (ba4jcat7yeuuvphcgsnonurdn6c5bnqwxe5vrndft3z4bx4pnko2tptf4/auth/create-auth-manager.tsx:400:10)
```

When you click the 'Connect Google Account' button.

image

bfollington:
I can't get `google-auth.tsx` to work locally, something about the identity isn't lining up during the oauth flow but, like the other comment above, I don't think this is actually introduced by this PR.

From what I can tell the patterns are valid, the issue lies elsewhere.

bfollington (approved):
Found two issues (in comments) but neither seem to be introduced in this PR, so I am approving. The patterns compile and run in the browser for me.

PR #3275, opened 2026-04-09
docs: refresh ternary lowering guidance

mathpirate:
## Summary
- update the `pattern-dev` and `pattern-critic` skills to point agents at `cf check --show-transformed` and to prefer plain ternaries at supported lowered value-expression sites
- refresh the main pattern-authoring docs (`pattern-development-guide`, `conditional`, `computed`, `jsx`) so they describe the shared expression-site model instead of framing ternaries as JSX-child-only
- extend the transformer spec docs (`target_pattern_language_spec`, `current_behavior_spec`, `lowering_contract`, `review_guide`) so the normative language now names the actual lowered container kinds and author-facing buckets
- pin an important current-main nuance: explicit compute callbacks still preserve authored JS control flow even for nested returned JSX ternaries/logicals, and call-argument / array-element sites can still be more nuanced than the highest-confidence direct JSX/object-property/local-const cases
- align adjacent docs/examples (`pattern-critique-guide`, plus the existing navigation/view-switching/llm updates in the branch) so they no longer present authored `ifElse(...)` or JSX-centric shorthand as the default mental model

## Validation
- `git diff --check`
- `rg -n "JSX-facing|feed JSX|top-level JSX children|JSX-adjacent|returned pattern values" docs skills`
- probe with `deno task cf check packages/patterns/gideon-tests/_tmp-computed-conditional-probe.tsx --show-transformed --no-run` before deleting the temp file, to verify current compute-callback ternary/logical behavior directly

### DIFF
--- a/docs/common/ai/pattern-development-guide.md
+++ b/docs/common/ai/pattern-development-guide.md
@@ -207,11 +207,15 @@ buckets:
 - top-level pattern-body value-expression sites, such as:
   - returned object property values
   - variable initializers
-  - call arguments
-  - array elements
   - return expressions
 - callback-local value-expression sites inside supported reactive collection
   callbacks
+
+Some call-argument and array-element sites are also lowerable, but those can
+still route through whole-site wrapping rather than direct inner `ifElse()`
+lowering. For unusual call-root shapes, inspect the emitted output with
+`cf check --show-transformed` instead of assuming the conditional itself will
+be rewritten.
 
 ```tsx
 // Wrong
@@ -231,9 +235,10 @@ truthy. That leads to subtle incorrect rendering because ternaries inside the
 `computed()` body are plain JS, not transformer-lowered conditionals.
 
 That distinction matters: explicit computation callbacks like `computed`,
-`derive`, `action`, `lift`, and `handler` are not blanket "rewrite everywhere"
-contexts. Inside those callback bodies, ternaries follow ordinary JavaScript
-rules unless they appear inside a nested supported lowered site such as JSX.
+`derive`, `action`, `lift`, and `handler` are preserved-JavaScript control-flow
+boundaries. Inside those callback bodies, ternaries and logical operators
+follow ordinary JavaScript rules even when nested inside returned JSX, local
+consts, or object literals.
 
 ### CORS and `fetchData`
 
--- a/docs/common/ai/pattern-development-guide.md
+++ b/docs/common/ai/pattern-development-guide.md
@@ -73,6 +73,8 @@ Use the runtime, not just static reasoning:
 
 - `deno task cf check <pattern>.tsx`
 - `deno task cf check <pattern>.tsx --no-run` for faster type validation
+- `deno task cf check <pattern>.tsx --show-transformed` when you need to
+  inspect how the transformer lowered a conditional or reactive expression site
 - `deno task cf test <pattern>.test.tsx` when tests are justified
 
 Primary verification is still runtime behavior. Tests are for logic that is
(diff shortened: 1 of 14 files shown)


docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md
```diff
@@ -326,6 +326,44 @@ structurally representable.
...
+
+These categories do not mean "rewrite every descendant expression." Explicit
+computation callbacks still create their own ownership boundary, and every site
+must separately satisfy the shared handling policy.
+
+In particular, current-main explicit compute callbacks preserve authored
+JavaScript control flow even for nested JSX ternaries/logicals inside the
+callback body. The recent recursive lowering work applies to pattern-owned
+sites and supported collection callbacks, not to explicit compute callback
+bodies.
```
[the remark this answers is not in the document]
mathpirate: @cubic-dev-ai I don’t think this is a valid issue as written.

The sentence here is describing the ownership boundary for recursive lowering: explicit compute callbacks do not get the newer recursive conditional lowering, even when they return JSX containing ternaries or logical operators.

That is consistent with the later JSX-routing section, which still says compute-context JSX does not lower `&&` / `||`. So this is not a contradiction in current-main behavior; it is the same carveout described from two angles.

I agree this section could be tightened editorially, but I don’t think it needs a correctness fix. Anyway I'll go ahead and make one editorial tightening.

docs/common/concepts/computed/computed.md
```diff
@@ -72,19 +72,27 @@ export default pattern<Input>(({ deck }) => ({
 Static strings like `[NAME]: "My Pattern"` don't need `computed()`.
 
 **Never wrap JSX in `computed()`** — the transformer automatically handles
-reactivity in JSX expressions. Ternaries in JSX children position are
-automatically converted to `ifElse()`, which correctly unwraps
-OpaqueRefs. Nested ternaries work too — a ternary inside the truthy branch of
-another ternary is also transformed. See `docs/common/patterns/conditional.md`.
+reactivity in JSX expressions and in many supported lowered value-expression
```
seefeldb: This sounds a bit like only JSX expressions are reactive?

docs/common/ai/pattern-development-guide.md
```diff
@@ -212,7 +231,14 @@ The JSX transformer handles ternaries correctly, including nested ternaries.
 ```
 
 Inside `computed()`, a `Writable<boolean>` is just a JS object and therefore
-truthy. That leads to subtle incorrect rendering.
+truthy. That leads to subtle incorrect rendering because ternaries inside the
```
seefeldb, line 222: I wonder whether (a) we should detect that (any truthy check on a cell-like) and error or (b) just generally suggest in all docs and examples to call .get() and optimize that away when transforming and it's redundant?

docs/common/ai/pattern-development-guide.md
```diff
@@ -195,8 +197,25 @@ callbacks.
 
 ### Conditional JSX
 
-Do not use `computed()` to gate JSX sections. Use JSX conditionals directly.
-The JSX transformer handles ternaries correctly, including nested ternaries.
+Do not use `computed()` to gate JSX sections. Use plain authored ternaries at
+supported lowered value-expression sites instead.
+
+The transformer now lowers plain ternaries across three main author-facing
```
seefeldb: All this might net be confusing, especially why ternaries are called out and other expressions aren't. Your chance made the distinction mostly go away ..

docs/common/concepts/computed/computed.md
```diff
@@ -104,20 +112,23 @@ the most common source of "conditional section always renders" bugs.
...
   )
   : null}
 ```
 
 **Rule of thumb:** `computed()` is for deriving data (strings, numbers,
-arrays, objects). For conditional rendering, use JSX ternaries.
+arrays, objects). For conditional rendering or simple conditional values at
+supported lowered value-expression sites, use plain ternaries. If you're unsure
+whether a site lowers, inspect it with
+`deno task cf check <pattern>.tsx --show-transformed`.
```
seefeldb, line 128: Oh, interesting. This feels like digging deep, I wonder whether this should be in some diagnostics centered doc than this more general one (that said, it already goes deep about lowering and so on ..)

seefeldb (approved):

PR #3276, opened 2026-04-09
docs(ui): strengthen theme-first agent guidance

mathpirate:
## Summary
- make `pattern-ui` explicitly theme-first and more intentional about visual direction
- turn `docs/common/patterns/style.md` from a placeholder into actionable guidance for using `cf-theme`
- strengthen the cookbook and key reference docs with verified `cf-theme` examples, concrete token names, and catalog vignette pointers
- add a small amount of rationale for why `cf-theme` and public component affordances are the intended styling boundary

## Context
This is a first pass on [CT-1451](https://linear.app/common-tools/issue/CT-1451/agents-need-guidance-to-use-our-theme-system): agents need clearer guidance to use `cf-*` components and `cf-theme` intentionally instead of producing lightly-styled default shells.

The issue description plus the recent design discussion point at the same core gap:
- agents are not being asked to form a strong visual point of view before coding
- our current docs mention styling affordances, but they do not make the theme system feel like the normal starting point
- the guidance stack was still too placeholder-heavy for a real UI design phase

### 1. `pattern-ui` is now theme-first
`skills/pattern-ui/SKILL.md` now asks the UI agent to:
- decide purpose, tone, and a memorable visual hook before coding
- default to defining a `theme` object early
- wrap the main surface in `cf-theme`
- use catalog vignette stories as the first visual references
- avoid generic, low-intent UI output

### 2. `style.md` now explains the actual theme workflow
`docs/common/patterns/style.md` is no longer just a placeholder.
It now covers:
- a design brief before coding
- a theme-first workflow
- the supported `cf-theme` object shape
- the concrete `--cf-theme-*` tokens emitted by the provider
- how to use component-specific `--cf-*` variables for local refinement
- which existing catalog stories demonstrate the intended approach

### 3. The cookbook points at actual vignette patterns
`docs/common/patterns/ui-cookbook.md` now includes:
- a theme-first shell scaffold
- a mobile-vignette scaffold
- an editorial detail layout
- a more intentional empty state
- direct references to the recipe and finance catalog vignettes

### 4. Lower-level references now tell the same story, but stay compact
I added or tightened short theme-system sections in:
- `docs/common/components/COMPONENTS.md`
- `packages/ui/README.md`
- `packages/ui/LLM-COMPONENT-INSTRUCTIONS.md`

The point is consistency across the guidance stack, without turning every lower-level doc into a second full styling guide.

### 5. Added a narrow rationale layer
Following review feedback, the PR now includes a short explanation of why this system is set up this way:
- theme and public component affordances are the styling boundary
- embedded UI should be restyled by wrapping it, not by reaching into internals
- the component set is intentionally opinionated
- when public hooks are not enough, that is either a design-system gap or an `iframe` / artifact case

I kept this intentionally narrow. It is enough to justify the guidance, not a full design-system manifesto.

## Why this scope
I intentionally kept this PR on the guidance side only.
It does **not** change runtime behavior or the theme implementation itself.

I also intentionally kept it out of the critic/testing-policy lane. Earlier versions drifted in that direction and I backed that out so this stays focused on what UI-design agents should read when asked to style a pattern.

## Notes / limits
- This is still not a complete component-by-component theming reference.
- The component docs still do not enumerate every custom property or CSS part.
- This is not the full long-term design-principles document.
- The intent is to provide better stopgap guidance for agents now, while leaving plenty of room for Sophie's larger design-system pass to replace or supersede this material later.

## Verification
- `git diff --check`
- targeted review against current `cf-theme` APIs, emitted token names, and current catalog vignette examples

### DIFF
--- a/docs/common/ai/manual-testing-guide.md
+++ b/docs/common/ai/manual-testing-guide.md
@@ -43,28 +43,6 @@ agent-browser snapshot -i
 agent-browser screenshot
 ```
 
-## Design-Focused Browser Checks
-
-When the pattern includes meaningful UI work, do not stop at "it rendered."
-Also verify:
-
-- the page loads without obvious console/runtime errors
-- the main screen has a readable hierarchy and grouping rhythm
-- `cf-theme`-driven styling or other public component styling hooks actually
-  appear in the rendered result when they were part of the design intent
-- full-height layouts can actually reach below-the-fold content, either through
-  normal page scroll or an explicit inner scroll region such as `cf-vscroll`
-- text inputs, selections, and badges render expected user-facing values rather
-  than raw object output or schema artifacts
-- empty, loading, and first-run states are understandable and feel designed
-- screenshots from key states support the visual claims in the review
-
-For UI-heavy patterns, capture at least:
-
-1. first load
-2. one meaningful interactive state
-3. one empty, onboarding, or edge state when applicable
-
 ## Browser State and Session Discipline
 
 - isolate stale browser state when a previous login, draft, or cached view can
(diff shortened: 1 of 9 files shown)


bfollington (approved):
LGTM, much improved from the status quo at the very least. Thanks!

PR #3277, opened 2026-04-09
Fix callback-local ordinary call-root lowering ownership

mathpirate:
## Summary
- fix callback-local ordinary call roots inside supported collection callbacks so the outer call owns lowering instead of leaking a lowered inner `ifElse(...)` into helper arguments
- whole-wrap callback-local ordinary call roots across `variable-initializer`, `object-property`, `array-element`, and direct `return-expression` sites while keeping plain structural sites without an owning call root on the direct-lowering path
- update and add closure fixtures to pin the semantics-preserving behavior, and refresh the current-behavior spec to describe the resulting ownership rule

## Root Cause
The previous callback-local policy let nested structural sites such as `call-argument`, `object-property`, and `array-element` lower directly even when they were nested under an ordinary helper call. That produced shapes like `helper(ifElse(...))`, which are semantically unsafe because `ifElse(...)` yields an `OpaqueRef` rather than the authored plain value the helper expects.

## Impact
This makes callback-local lowering easier to reason about and preserves call semantics for ordinary helpers in supported collection callbacks. The direct-lowering behavior remains for plain structural sites that are not enclosed by an owning ordinary call root.

## Validation
- `UPDATE_GOLDENS=1 FIXTURE_PATTERN='computed-map-call-arg-conditional|computed-map-call-root-containers|computed-map-local-call-root' deno test --allow-read --allow-write --allow-env test/fixture-based.test.ts`
- `FIXTURE_PATTERN='computed-map-call-arg-conditional|computed-map-call-root-containers|computed-map-local-call-root|computed-map-local-array-conditional' deno test --allow-read --allow-write --allow-env test/fixture-based.test.ts`

### DIFF
--- a/packages/ts-transformers/test/fixtures/closures/computed-map-call-root-containers.input.tsx
+++ b/packages/ts-transformers/test/fixtures/closures/computed-map-call-root-containers.input.tsx
@@ -1,0 +1,40 @@
+import { computed, pattern } from "commonfabric";
+
+const identity = (value: string) => value;
+
+interface Item {
+  done: boolean;
+}
+
+interface State {
+  items: Item[];
+}
+
+// FIXTURE: computed-map-call-root-containers
+// Verifies: inside a computed-array .map() callback, nested structural sites
+//   still lower inner conditionals directly, while a direct callback
+//   return-expression ordinary call root lowers as a whole callback-local
+//   derive.
+//   ({ value: identity(row.done ? "Done" : "Pending") })
+//   → ({ value: identity(ifElse(row.done, "Done", "Pending")) })
+//   [identity(row.done ? "Done" : "Pending")]
+//   → [identity(ifElse(row.done, "Done", "Pending"))]
+//   row => identity(row.done ? "Done" : "Pending")
+//   → row => derive(..., ({ row }) => identity(row.done ? "Done" : "Pending"))
+export default pattern<
+  State,
+  { views: { value: string; list: string[] }[]; labels: string[] }
+>((state) => {
+  const rows = computed(() => state.items);
+
+  const views = rows.map((row) => ({
+    value: identity(row.done ? "Done" : "Pending"),
+    list: [identity(row.done ? "Done" : "Pending")],
+  }));
+
+  const labels = rows.map((row) =>
+    identity(row.done ? "Done" : "Pending")
+  );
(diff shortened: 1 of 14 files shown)


packages/ts-transformers/test/fixtures/closures/computed-map-call-root-containers.input.tsx
```diff
@@ -0,0 +1,40 @@
...
+  items: Item[];
+}
+
+// FIXTURE: computed-map-call-root-containers
+// Verifies: inside a computed-array .map() callback, nested structural sites
+//   still lower inner conditionals directly, while a direct callback
+//   return-expression ordinary call root lowers as a whole callback-local
+//   derive.
+//   ({ value: identity(row.done ? "Done" : "Pending") })
+//   → ({ value: identity(ifElse(row.done, "Done", "Pending")) })
```
seefeldb: oh, this happens to work because the function is identity, but if it was e.g. `toBool: (maybeDone) => maybeDone === "Done"`, then this would cause a bug, since the return value of ifElse is going to be a cell.

seefeldb: instead this should just be a `derive({ row: { done: row.done }}, { row } => identity(row.done ? "Done" : "Pending")`

(or if we want to be very fancy, since this is a function call, it could become `derive(ifElse(....), x => identity(x))`, but I think this goes too far)

mathpirate: Thanks, this was a real semantics bug. I updated the transformer so callback-local ordinary call roots now own lowering across `variable-initializer`, `object-property`, `array-element`, and direct `return-expression` sites, which prevents nested structural sites from lowering directly under an enclosing helper call. Concretely, shapes like `identity(row.done ? "Done" : "Pending")` now whole-wrap to `derive(..., ({ row }) => identity(row.done ? "Done" : "Pending"))` instead of `identity(ifElse(...))`. I also added a non-JSX local-initializer fixture (`computed-map-local-call-root`) and revalidated the affected closure slice plus the unchanged structural control case (`computed-map-local-array-conditional`).

seefeldb (approved):

PR #3279, opened 2026-04-10
[codex] remove legacy shell env experimental aliases

mathpirate:
## Summary
- remove legacy shell experimental globals and alias fields from `packages/shell/src/lib/env.ts`
- update the shell navigation tests to preserve the modern globals instead of the old aliases
- add a focused shell env test that proves the modern globals are honored and the old globals are ignored

## Why
`labs#3272` narrows the shell build-time define surface back to the documented modern env vars. While investigating that change, I found that `packages/shell/src/lib/env.ts` was still reading and exporting the old alias globals:
- `EXPERIMENTAL_RICH_STORABLE_VALUES`
- `EXPERIMENTAL_STORABLE_PROTOCOL`
- `EXPERIMENTAL_CANONICAL_HASHING`

In the shell itself, those names were no longer real inputs. The only shell-side references left were:
- `env.ts` itself
- `runtime-navigation.test.ts`, which was preserving those globals in test setup

The downstream runtime path already treats `modernDataModel` / `modernHash` as the primary contract, with the old names only retained as runner-side backward-compat aliases. So this follow-up removes the dead shell-side compatibility layer and adds a targeted regression test for the modern contract.

`EXPERIMENTAL_STORABLE_PROTOCOL` in particular appears fully defunct here, matching Dan's note that it never got wired up.

## Validation
- `cd packages/shell && deno task test`

## Notes
- This overlaps slightly with `labs#3272` on the shell test-task permissions: the new focused env test needs `--allow-read` for fresh module imports. If `#3272` lands first, that part should collapse cleanly on rebase.

### DIFF
--- a/packages/shell/src/lib/env.ts
+++ b/packages/shell/src/lib/env.ts
@@ -6,12 +6,9 @@ declare global {
   var $COMMIT_SHA: string | undefined;
   var $MEMORY_VERSION: string | undefined;
   var $EXPERIMENTAL_MODERN_DATA_MODEL: string | undefined;
-  var $EXPERIMENTAL_RICH_STORABLE_VALUES: string | undefined;
-  var $EXPERIMENTAL_STORABLE_PROTOCOL: string | undefined;
   var $EXPERIMENTAL_UNIFIED_JSON_ENCODING: string | undefined;
   var $EXPERIMENTAL_MODERN_SCHEMA_HASH: string | undefined;
   var $EXPERIMENTAL_MODERN_HASH: string | undefined;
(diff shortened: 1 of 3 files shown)


mathpirate:
This follow-up is directly based on your Discord guidance on `#3272`.

`#3272` narrows the shell build-time define surface; while checking that cleanup, I found the old shell env aliases were still hanging around in `packages/shell/src/lib/env.ts` and the shell navigation test setup. I verified they were no longer real shell inputs, so this PR just removes that leftover shell-side compatibility layer and adds a focused test around the modern contract.

danfuzz (approved):
LGTM!

PR #3289, opened 2026-04-13
docs(patterns): warn against bound-control self-feedback

mathpirate:
## Summary
- document that `$value` / `$checked` bindings are the primary value path for a control
- warn against `oncf-change` / `oncf-input` handlers that simply write the same value back into the same bound cell
- update the shared component docs, two-way-binding guidance, development/critique guides, and core pattern skills

## Why
During Pattern Factory runtime investigation, the strongest piece-local thrash repro came from a generated `cf-select` pattern that both bound `$value` and wrote the same value back into that same cell from `oncf-change`. A local validation patch that removed those obvious same-cell write-backs stopped reproducing the original thrash. This PR turns that lesson into shared authoring and critique guidance.


bfollington (approved):
This is good but it does point to us having a slightly "confusing" API since you can mix-match these and get half-working outcomes

PR #3295, opened 2026-04-14
test(runner): fix array_push integration path

mathpirate:
## Summary
- rewrite array_push.test.ts as a self-contained toolshed integration test
- verify writable-array .push() over the real remote memory transport instead of relying on piece creation side effects
- add pattern-debug guidance to inspect live state before blaming a documented primitive

## Verification
- deno fmt packages/runner/integration/array_push.test.ts skills/pattern-debug/SKILL.md
- deno test -A packages/runner/integration/array_push.test.ts

## Notes
The old test had multiple unrelated failure modes (bad websocket URL, cwd-sensitive source loading, and a piece lifecycle mismatch). This version keeps the scope on runner + toolshed behavior and proves .push() works on current main.

### DIFF
--- a/packages/runner/integration/array_push.test.ts
+++ b/packages/runner/integration/array_push.test.ts
@@ -1,198 +1,175 @@
-#!/usr/bin/env -S deno run --allow-net --allow-env --allow-read
+#!/usr/bin/env -S deno run -A
 
 /**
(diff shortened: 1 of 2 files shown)


ubik2 (approved):
lgtm

PR #3304, opened 2026-04-15
feat(ts-transformer): Tighten handler identity schemas

seefeldb:
## Summary

Tightens handler schema shrinking for identity-only payloads so recursive handler payload types do not leak full structural schemas when handlers only preserve identity.

- Treats handler payloads passed through identity-only array writes as `unknown`.
- Preserves writable array cell wrappers while shrinking array item schemas to `unknown`.
- Adds a red-green fixture covering `addPiece` and `trackRecent` style recursive payload handlers.

## Root Cause

The capability summary could identify some identity-only uses, but later shrinking and schema generation could recover the original recursive semantic type from registered type information. Array write paths also needed enough context to distinguish identity-only handler payload writes from local arrays that are later structurally inspected.

## Validation

- `FIXTURE=identity-only-handler-payload deno test -A test/fixture-based.test.ts`
- `deno test -A test/policy/capability-analysis.test.ts`
- `deno test -A test/type-shrinking.test.ts`
- `deno task cf check packages/patterns/system/default-app.tsx --show-transformed --no-run`
- pre-commit `deno fmt`
- pre-commit `deno lint`

---
## Summary by cubic
Tightens handler schema shrinking to stop recursive payloads from leaking when handlers only preserve identity. Identity-only payloads and array items now shrink to `unknown` while keeping array and element cell wrappers.

- **Bug Fixes**
  - Expands identity-only array detection to `push`, `unshift`, `splice` (value args only), `slice` chains, element spreads, and array‑literal locals passed to `.set`; only treats `.set` on cell-like receivers, ignores arrays inspected before `set`, and keeps primitive item writes structural.
  - Treats arguments to `equals` and `navigateTo` as identity-only; records identity without wildcarding and marks array item identity when identity args are written via array writers.
  - Injects item-level schema hints from capability summaries so formatters emit `items: { type: "unknown" }` and preserve outer and element cell wrappers.
  - Skips payload loading for truthiness checks on property aliases, including optional‑chained alias initializers and boolean conditions on local alias members.
  - Uses synthetic-node schema generation when type nodes contain `any`/`unknown` and avoids TypeRegistry transfer; improves shrinking for arrays/readonly arrays/unions/referenced types and skips shrinking when identity containers already cover retained paths.

<sup>Written for commit 8f60cf6a1a068a270527b0759d21f226ca2c80fb. Summary will update on new commits.</sup>

NEW_PERF_BASELINE: subtest: pattern-unit-5/pattern-unit-tests > packages/patterns/simple-list/simple-list.test.tsx = 14s

### DIFF
--- a/packages/ts-transformers/src/policy/capability-analysis.ts
+++ b/packages/ts-transformers/src/policy/capability-analysis.ts
@@ -101,7 +101,8 @@ function extendSourceRef(
 const PARAMETER_SUMMARY_PREFIX = "__param";
 
 const WRITER_METHODS = new Set(["set", "update"]);
-const ARRAY_IDENTITY_WRITER_METHODS = new Set(["push"]);
+const ARRAY_IDENTITY_WRITER_METHODS = new Set(["push", "unshift", "splice"]);
+const ARRAY_IDENTITY_PRESERVING_CHAIN_METHODS = new Set(["slice"]);
 const READER_METHODS = new Set(["get"]);
 const FALLBACK_OPERATORS = new Set<ts.SyntaxKind>([
   ts.SyntaxKind.QuestionQuestionToken,
@@ -482,7 +483,22 @@ function isArrayIdentityWriterArgumentUsage(
         isLiteralElement(target.argumentExpression)
     ? getLiteralElementText(target.argumentExpression)
     : undefined;
-  return !!methodName && ARRAY_IDENTITY_WRITER_METHODS.has(methodName);
+  return isArrayIdentityWriterValueArgument(methodName, parent, usage);
+}
+
+function isArrayIdentityWriterValueArgument(
+  methodName: string | undefined,
+  call: ts.CallExpression,
+  usage: ts.Expression,
+): boolean {
+  if (!methodName || !ARRAY_IDENTITY_WRITER_METHODS.has(methodName)) {
+    return false;
+  }
+  const index = call.arguments.findIndex((argument) => argument === usage);
+  if (index < 0) {
+    return false;
+  }
+  return methodName === "splice" ? index >= 2 : true;
 }
 
 function isOptionalAliasInitializerMemberUsage(usage: ts.Expression): boolean {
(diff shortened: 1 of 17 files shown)


packages/ts-transformers/src/policy/capability-analysis.ts
```diff
@@ -464,6 +465,229 @@ function isCallOrNewArgumentUsage(
...
+    if (ts.isArrayLiteralExpression(current)) {
+      return true;
+    }
+    if (ts.isCallExpression(current)) {
+      const target = unwrapExpression(current.expression);
+      if (
+        ts.isPropertyAccessExpression(target) ||
+        ts.isElementAccessExpression(target)
+      ) {
+        return expressionStartsWithArrayLiteral(target.expression);
```
mathpirate: [P1] Array-literal `.set()` sinks treat arbitrary call chains as identity-only

`collectArrayLocalsPassedToSet()` currently accepts any initializer whose receiver chain starts at an array literal. Combined with the new optional-alias fast path, `const piece = input?.piece; if (!piece) return; const updated = [piece].map((p) => ({ title: p.title })); state.items.set(updated);` now reports `identityPaths:['piece']` and no reads, even though `p.title` is structurally read. Because the callback on an array-literal receiver is never rebound to the source element, this can shrink `piece` to `unknown` and underfetch data. Restrict this path to identity-preserving methods only, or drop the optimization whenever the initializer contains an inspecting callback like `.map(...)`.

packages/ts-transformers/src/policy/capability-analysis.ts
```diff
@@ -464,6 +465,229 @@ function isCallOrNewArgumentUsage(
...
+        }
+      }
+    }
+
+    if (ts.isPropertyAccessExpression(node)) {
+      const current = unwrapExpression(node.expression);
+      const isSetCallReceiver = node.name.text === "set" &&
+        ts.isCallExpression(node.parent) &&
+        node.parent.expression === node;
+      if (ts.isIdentifier(current) && !isSetCallReceiver) {
```
mathpirate: [P2] Local `push` sinks are still excluded from identity-array handling

The PR adds `push` to `ARRAY_IDENTITY_WRITER_METHODS`, but this scan still records any property access other than `.set` as structural use. That means `const updated: Piece[] = []; updated.push(piece); state.items.set(updated);` is filtered out of `identityArrayLocals`, and the analyzer still reports `readPaths:['piece']` with no identity path. If local `push` plus `.set(...)` is meant to be supported, `push` needs to be exempted here or handled explicitly in the local-array sink logic.

mathpirate (commented):
I left two inline findings on the local-array sink heuristics in `capability-analysis.ts`. Both are still present on the current head `8526dc4e`; the latest formatter-only follow-up did not affect them.

packages/ts-transformers/src/policy/capability-analysis.ts
seefeldb: i wonder whether the issue here is the `[piece]` part or the .map part. this here is complex and it's fine to fall back to not shrinking if that is the reason. if it's .map, then that's common and a serious bug.

other question i have is whether this is handler-only or broader. if handler-only then that's code smell that we're duplicating something here.

PR #3312, opened 2026-04-15
feat(fuse): Add CFC annotations for FUSE projections

seefeldb:
## Summary

Implements the Common Fabric CFC annotation and prepared writeback foundation for `packages/fuse`. The FUSE daemon now acts as a trusted producer of protected CFC metadata for projected Common Fabric filesystem nodes, while gVisor remains the future sandbox enforcement point.

### CFC annotation model and xattr contract

- Adds typed CFC node annotations for projected values, directories, aggregate JSON siblings, top-level `input.json` / `result.json`, sigil symlinks, callable descriptors, piece/space manifests, source projections, and synthetic index/status-style nodes.
- Emits stable logical projection refs based on Common Fabric identity and JSON projection shape, not transient inode numbers or mount-local paths.
- Exposes protected metadata through `trusted.cfc.*` and the local compatibility namespace `user.commonfabric.cfc.*`, with `trusted.cfc.generation` as raw UTF-8 and other CFC xattr values as canonical JSON.
- Fails closed when authoritative path-granular runner labels are unavailable by emitting `CommonFabricFuseProjectionMetadataIncomplete` rather than treating missing labels as public.

### Projection generation and rebuild safety

- Derives deterministic projection generations from projected values, authoritative CFC label metadata, and the projection/profile version.
- Threads real projection generations through CLI-mounted FUSE projections so refs, node generation, and callable descriptor generation agree for the same projection pass.
- Keeps generation derivation independent of transient FUSE inode allocation and mount-local paths.

### Runner mode plumbing and CLI behavior

- Adds runner-aligned CFC modes: `disabled`, `observe`, `enforce-explicit`, and `enforce-strict`, using the runner default when no CLI override is provided.
- Adds FUSE CLI plumbing for `--cfc-mode`, `--cfc-annotations`, `--cfc-xattr-namespace`, temporary `--cfc-writeback-xattrs`, and durable `--cfc-writeback-state`.
- Preserves non-CFC behavior by default; annotation publishing and writeback guardrails are enabled only by CFC mode or explicit annotation/debug flags.

### Prepared writeback guardrails

- Adds prepare/finalize scaffolding for existing-file `write`/`truncate`, `create`, `mkdir`, `unlink`, `rmdir`, narrowly scoped same-cell `rename`, Common Fabric sigil symlink creation, and metadata-only `setattr` operations such as chmod/chown/timestamp updates.
- Applies monotone label bracketing: prepare conservative/fail-closed labels before mutation, commit to the runner, then finalize exact labels when the projection can be recomputed coherently.
- Uses parent-prepared namespace authorization for create/delete/rename/symlink operations so enforcing modes do not leak occupancy-sensitive results before the parent namespace is prepared.
- Keeps out-of-scope operations denied in enforcing modes, including arbitrary symlink targets and callable-send writeback.

### Recovery, reconciliation, and diagnostics

- Persists prepared/fail-closed writeback records outside the mount and tracks recovery states for pending prepare, mutation applied, runner commit failed, stale generation, malformed/unsupported prepare, ready for exact recomputation, and finalized-before-cleanup.
- Reconciles on startup/space connect and after CellBridge hydration or projection rebuilds: exact coherent annotations are finalized; otherwise prepared/fail-closed annotations remain in place.
- Rebinds recovery records across inode churn using logical CFC refs rather than inode identity.
- Adds CFC writeback status and diagnostics to the mount `.status` output.

### Docs and spec placement

- Moves the FUSE CFC annotation note into `docs/specs/fuse-filesystem/9-cfc-annotations.md` and links it from the FUSE filesystem spec index.
- Updates `packages/fuse/README.md` for CFC modes, xattr namespaces, prepared writeback, recovery state storage, and current enforcement boundaries.

## Explicitly Out Of Scope

- No gVisor enforcement is implemented in this package.
- FUSE does not decide whether a sandboxed process may observe data; it only publishes annotations and applies local guardrails.
- `user.commonfabric.cfc.*` remains compatibility/debug metadata and must not be trusted as an enforcement channel.
- Callable-send writeback and arbitrary symlink target creation remain out of scope.

## Validation

- `deno task test` from `packages/fuse`: 150 passed, 1 ignored
- `deno check packages/fuse/mod.ts packages/fuse/cfc-writeback.ts packages/fuse/cell-bridge.ts packages/fuse/cfc-writeback.test.ts packages/fuse/cell-bridge.test.ts`
- `deno fmt --check packages/fuse/cfc-writeback.ts packages/fuse/cfc-writeback.test.ts packages/fuse/cell-bridge.ts packages/fuse/cell-bridge.test.ts packages/fuse/mod.ts packages/fuse/README.md`
- `deno lint packages/fuse`

---
## Summary by cubic
Adds an opt-in CFC annotation layer to `packages/fuse` and extends CFC writeback guardrails to cover writes, metadata changes, namespace mutations, and symlink creation, with crash-safe recovery. Adds cross-platform `setxattr` support, preserves CFC directory entry annotations across renames, and aligns defaults with the spec (CFC mode default from runner, xattr namespace defaults to both).

- **New Features**
  - Typed annotations for values, dirs, directory entries, aggregate `.json`, `input.json`/`result.json`, sigil symlinks, callables, manifests, source projections, and derived slots; stable logical refs independent of inodes.
  - Deterministic projection generation exposed as `trusted.cfc.generation`; xattrs available under `trusted.cfc.*` and `user.commonfabric.cfc.*` (select via `--cfc-xattr-namespace=trusted|compat|both`; defaults to `both`, unknown values rejected).
  - Guardrail modes via `--cfc-mode=disabled|observe|enforce-explicit|enforce-strict`; observe/enforce publish annotations automatically. Writebacks gated by `trusted.cfc.writeback.prepare`/`trusted.cfc.writeback.finalize` on the target or parent (via `applyPreparedParent`), with reconciliation on restart; Linux and macOS `setxattr` supported; per-op truncate/write authorization and explicit `setattr` flag mapping.
  - Fails closed when runner path-granular labels are missing using the `CommonFabricFuseProjectionMetadataIncomplete` caveat.
  - Programmatic APIs: `CfcProjectionAnnotator`, `deriveCfcProjectionGeneration`, `listCfcXattrNames`/`getCfcXattrValue`, `authorize*`/`applyPrepared*` helpers, `CfcWritebackStore`, and `parseCfcMode`/`resolveCfcMode`/`shouldEnableCfcAnnotations`. Spec at `docs/specs/fuse-filesystem/9-cfc-annotations.md`.

- **Migration**
  - Off by default. Enable annotations with `--cfc-annotations` or construct `CellBridge` with `{ cfcAnnotations: true, projectionGeneration?: string }`. Set guardrails with `--cfc-mode` (defaults to the runner’s `DEFAULT_CFC_ENFORCEMENT_MODE`).
  - In `enforce-*` modes, writes, metadata changes, namespace mutations, and symlink creation require the prepare/finalize xattr flow; `observe` allows operations but records annotations. `trusted.cfc.*` is the enforcement namespace; `user.commonfabric.cfc.*` is for local compatibility and must not be trusted by sandboxes. `--cfc-xattr-namespace` defaults to `both` and rejects unknown values.

<sup>Written for commit 275f429e9ad464ae51b26d1538a6aecccaf2a4da. Summary will update on new commits.</sup>


willkelly (commented):
Reviewed the diff against ch. 18 §18.2.3.5 / §18.2.3.6 in `~/src/commontoolsinc/specs/cfc`. Prior cubic findings look largely addressed in `f297548` — the P1 setattr finalize guard, rename entry propagation, `symlink` xattr field, `setFinalizeXattr` name iteration, diagnostics ring bound, and test assertion ordering all appear fixed. The P3 `cfcEntryNameDigest` / `nameDigest` duplication (`tree.ts:13` vs `annotations.ts:471`) is still open and worth a one-line consolidation into annotations.ts.

## Spec alignment — worth discussing

**Compat namespace as default.** §18.2.3.5 positions `trusted.cfc.*` as the normative enforcement namespace and `user.commonfabric.cfc.*` as a debug/compat *fallback* ("never enforcement input"). This PR sets `compat` as the CLI default (`mod.ts:156`) and the README documents "By default the local mount exposes the compatibility namespace". I get that unprivileged FUSE on Linux can't set `trusted.*`, but defaulting to compat alone means the normative namespace is absent unless an operator opts in. Worth considering `both` as the default so the normative names are always present where the platform allows. Also: unknown `--cfc-xattr-namespace` values silently fall back to `compat` with a warning (`parseCfcXattrNamespace`); given the fail-closed posture elsewhere, rejecting an unknown value seems more consistent.

**`deriveCfcProjectionGeneration` inputs.** §18.2.3.6 lists schema hash and daemon implementation identity among the derivation inputs for mount-level identity; neither is currently in the digest (`annotations.ts:320`). Schema-hash omission means a silent schema change with an unchanged value could collide with an older generation — a subtle fail-open. Daemon-identity omission is lower-stakes but also non-spec. If schema-hash wiring is deferred to a later slice, a `// TODO(schema-hash)` in the derive function plus a pending test would make the gap visible.

**Writeback state path defaults to `/tmp`.** `defaultCfcWritebackStatePath` falls back to `${TMPDIR:-/tmp}/commonfabric-fuse/` (`mod.ts:119`). `/tmp` is world-writable; a local user can pre-create or mutate the recovery file between daemon runs. `parsePreparedWriteback` is defensive, so this isn't a direct exploit, but it can influence recovery classification on restart. Prefer `$XDG_STATE_HOME/commonfabric-fuse` (fallback `~/.cache/commonfabric-fuse`).

**Observe mode allows local writes.** ch. 18 §18.2.3.6 (~lines 821–824 in the current spec) says that when `--cfc-annotations` is enabled, local writes/creates/renames/symlinks/truncates/metadata mutations are *rejected* until gVisor can pass trusted write-label metadata back. This PR relaxes that for `observe` (log-and-allow). The loosening is reasonable for the pre-gVisor dev loop and `9-cfc-annotations.md` documents it, but it's a divergence from the ch. 18 text. Probably fine for the current implementation state — I'd suggest a small note in the spec (and/or ch. 18 alignment in a follow-up) to flag it as a temporary relaxation to tighten up when gVisor lands.

## Test coverage vs. spec

The mode matrix, recovery state machine (incl. restart / inode churn), parent-prepared occupancy gating, sigil-only symlink enforcement, and fail-closed-on-missing-metadata are all well-covered (`cfc-writeback.test.ts` lines 243–1380; `annotations.test.ts` lines 231–649). Gaps I'd consider filling:

- **Canonical-JSON sorted-key serialization.** §18.2.3.5 requires the non-`generation` xattr values be canonical JSON with sorted keys so producers are byte-identical. The golden test (`annotations.test.ts:511`) checks values and presence but not byte-equality against a known sorted-order expectation. A targeted assertion (feed an unsorted-order label/value; serialize; compare bytes) would lock this down.
- **"Generation published only after both runner and CFC commit" as a direct assertion.** This is a §18.2.3.6 MUST. The reconciliation tests show `prepared:sha256:` survives across rebuild (implicit), but no test explicitly asserts that `getCfcAnnotation(ino).generation` is *not* the new exact generation between `markMutationApplied` and `markReadyForExactRecomputation`.
- **Authoritative precedence** (runner CFC metadata > schema-derived > `Labels.classification`) isn't tested — seemingly because the schema/classification fallbacks aren't wired in this slice. If that's intentional, a review comment noting the deferred tier would help the next slice remember.
- **Per-path fail-closed mixing.** The fail-closed test covers "no `labelView` at all". Not tested: `labelView` present but some descendant paths have no matching entry — each unlabeled descendant should still carry `CommonFabricFuseProjectionMetadataIncomplete`. Likely correct from the code, but a focused test would catch regressions.

## Nits

- `cfcEntryNameDigest` in `tree.ts:13` duplicates `nameDigest` in `annotations.ts:471` (cubic P3). Suggest exporting from annotations.ts and importing in tree.ts so the digest and entry-kind helpers live with the rest of the annotation semantics.
- `authorizeExistingCfcWrite` funnels per-call diagnostics through the global `cfcDiagnostics` ring rather than returning them per-operation. Fine for the mount `.status` aggregation; flagging only in case per-op audit traceability is wanted later.

Overall: direction looks right and the scaffolding is thorough. Main suggestions are around defaults that lean debug-friendly vs. spec-normative and a few test assertions that would directly lock in spec invariants rather than cover them transitively.

PR #3331, opened 2026-04-20
perf(runner): intern schema-hash cache keys to close modern-schema-hash regression

danfuzz:
## What

Adds a new `internedPairKey(a, b)` helper to `packages/data-model/schema-hash.ts` that composes two `internSchema(x, true).hashString`s with a `|` delimiter — producing a stable, base64url-collision-safe cache key from a pair of schemas.

Applies it at 5 merge-helper cache-key sites in `packages/runner/src/traverse.ts`:

- `mergeSchemaFlags` — `internedPairKey(flagSchema, schema)`.
- `combineSchema` — `internedPairKey(parentSchema, linkSchema)`.
- `mergeSchemaOption` — `internedPairKey(outerSchema, innerSchema)`.
- `mergeAnyOfBranchSchemas` — inline 1+N key (`${outerHashString}||${branches.map(…).join("|")}`) since it's a single site, not a recurring 2-arg pattern. Result is interned directly.
- anyOf property dedup — keys a `uniqueHashes` Map on `internSchema(s, true).hashString`.

Also upgrades the shared `internSet` helper from `toDeepFrozenSchema(value, true)` to `internSchema(value)` — a single-line change that retires "intern the merged result" guidance for three of the five sites at once while preserving the `INTERN_CACHE_MAX=10_000` cap and clear-on-overflow behavior.

## Why

The modern schema-hash path (`hashOfModernInternal` in `packages/data-model/value-hash-modern.ts`) WeakMap-caches only deep-frozen objects. Hot-path callers in `traverse.ts` construct fresh unfrozen spread objects every call, so they miss the modern cache entirely — producing the CI regression first observed on PR #3235 (parking-coordinator pattern test: ~78 min on the PR branch vs ~2 min on `main`).

Interning schemas produces stable, deep-frozen identities that are cache-friendly for the modern hash path, retiring 5 of the cache-defeat sites identified in the 2026-04-16 audit.

## Context

This is **PR 1 of a three-PR staged structural fix**:

- **PR 1 (this PR)** — Tactic 2A: intern schema-hash cache keys at the 5 merge-helper sites in `traverse.ts`.
- **PR 2** — Tactic 2B: `CompoundCycleTracker` restructure (tracked separately, lands independently).
- **PR 3** — Tactic 2C: `MapSetStringToPathSelectors` freeze (tracked separately, lands independently).

All three together are expected to close PR #3235's parking-coordinator regression. PR #3235 remains open as the final target; this PR unblocks it but does not close it.

Full diagnosis and per-site analysis: `coordination/docs/2026-04-16-modern-schema-hash-cache-audit.md` (§4 Phase 2). Empirical profiling: `coordination/docs/2026-04-16-modern-schema-hash-profiling.md`. Test-surface triage: `coordination/docs/2026-04-16-modern-schema-hash-test-surface-triage.md`.

Related PRs:
- PR #3235 — `feat: enable modernSchemaHash by default` (the target; still open).
- PR #3231 — `coder-cedar/flip-modern-hash-default` (upstream of #3235).
- PR #3324 — `danfuzz/because-im-easy-come-easy-go` (legacy-hash SHA256 normalization; apples-to-apples comparison prep).

### DIFF
--- a/packages/data-model/test/schema-hash.test.ts
+++ b/packages/data-model/test/schema-hash.test.ts
@@ -317,12 +317,12 @@ describe("schema-hash dispatch", () => {
       });
 
       describe("internSchemaKey()", () => {
-        it('returns `"T"` for the `true` schema', () => {
-          assertStrictEquals(internSchemaKey(true), "T");
+        it('returns `"<schema:true>"` for the `true` schema', () => {
+          assertStrictEquals(internSchemaKey(true), "<schema:true>");
         });
 
-        it('returns `"F"` for the `false` schema', () => {
-          assertStrictEquals(internSchemaKey(false), "F");
+        it('returns `"<schema:false>"` for the `false` schema', () => {
+          assertStrictEquals(internSchemaKey(false), "<schema:false>");
         });
 
         it("returns the interned schema's `hashString` for objects", () => {
@@ -385,13 +385,16 @@ describe("schema-hash dispatch", () => {
           const obj: JSONSchema = { type: "number" };
           assertStrictEquals(
             internedPairKey(true, obj),
-            `T|${internSchemaKey(obj)}`,
+            `<schema:true>|${internSchemaKey(obj)}`,
           );
           assertStrictEquals(
             internedPairKey(obj, false),
-            `${internSchemaKey(obj)}|F`,
+            `${internSchemaKey(obj)}|<schema:false>`,
           );
-          assertStrictEquals(internedPairKey(true, false), "T|F");
+          assertStrictEquals(
+            internedPairKey(true, false),
+            "<schema:true>|<schema:false>",
+          );
         });
 
         it("is order-sensitive", () => {
--- a/packages/data-model/test/schema-hash.test.ts
+++ b/packages/data-model/test/schema-hash.test.ts
@@ -12,6 +12,7 @@ import {
   hashSchemaItem,
   internedPairKey,
   internSchema,
+  internSchemaHashString,
   isInternedSchema,
   resetSchemaHashConfig,
   setSchemaHashConfig,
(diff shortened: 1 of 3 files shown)


packages/data-model/test/schema-hash.test.ts
```diff
@@ -313,6 +315,126 @@ describe("schema-hash dispatch", () => {
...
+          assertStrictEquals(internedPairKey(a1, b1), internedPairKey(a2, b2));
+        });
+
+        it("interns both inputs as a side effect", () => {
+          const a: JSONSchemaObj = { type: "number" };
+          const b: JSONSchemaObj = { type: "string" };
+          assertStrictEquals(isInternedSchema(a), false);
+          assertStrictEquals(isInternedSchema(b), false);
+          internedPairKey(a, b);
+          assertStrictEquals(isInternedSchema(a), true);
```
danfuzz, line 447: This isn't strictly true in general: If one of these happens to be content-equal to a previously-interned schema, then, while they'll still be deep-frozen, `isInternedSchema()` will return `false` on them.

If you want to make this correct, you could guarantee a unique schema value (e.g. include a key such as `schemaHashTestAt${Date.now()}` that as a practical matter will only appear in the test).

danfuzz: Good catch, thanks. Fixed in `e19326879`: both "interns as a side effect" tests (the one here in `internedPairKey` and the matching one in the `internSchemaHashString` block added in the previous commit) now build their input schemas with a content-unique `title` field containing `schemaHashTestAt${Date.now()}-${Math.random()}`, so no prior interning can have seen the exact content. Added short comments explaining why.

The sibling tests in the same blocks (round-trip, equivalence, order-sensitivity, idempotence) don't exhibit the same flake because they assert on returned hash strings, which are stable regardless of prior interning state — so I left them using the plain `{ type: "number" }` / `{ type: "string" }` shapes.

— coder-cedar (Claude Opus 4.7 (1M context))

PR #3338, opened 2026-04-21
refactor(runner,memory): push SchemaPathSelector interning upstream of MapSetStringToPathSelectors

danfuzz:
## What

Pushes `SchemaPathSelector` interning upstream of `MapSetStringToPathSelectors` (MSP), simplifying MSP's contract from "I intern everything my callers hand me" to "callers must hand in interned selectors."

Changes:

- **New `internPathSelector(selector)` helper** in `packages/data-model/schema-utils.ts`, with 6 BDD tests. Interns `selector.schema` via `internSchema`, freezes `selector.path` and `selector`, returns.
- **Sweep ~13 MSP-feeding construction sites** across `packages/runner` and `packages/memory` to hand in pre-interned selectors instead of relying on MSP to intern them.
- **Simplify the MSP hash lambda** in `packages/runner/src/traverse.ts` — remove the now-redundant in-class interning (freeze-on-entry from PR #3335), since callers now guarantee interned input.
- **Module-level literal-selector consts** (per "intern literal data once at module load, reuse by reference"):
  - `DefaultSelector` — existing literal, now interned at module load.
  - `RejectingSelector` — new, in `traverse.ts` (replaces two per-call `{ path: [], schema: false }` sites).
  - `NORMALIZED_REJECT` — new, in `v2-watch.ts` (for `normalizeSyncSelector`'s reject-at-root return).
- **Refactor `normalizeSyncSelector`** (`packages/runner/src/storage/v2-watch.ts`) to compose with `internPathSelector`.

7 files, +195/-93 total.

## Why

Follow-up to PR #3335 (PR 3 / Tactic 2C — freeze-on-entry in the MSP hash lambda). Dan proposed a "trace-and-intern-upstream" refactor: rather than MSP doing the interning per-insert (which fires on every call), pre-intern selectors at their construction sites — most of which are module-level literals or runtime-variable-but-bounded values that benefit from identity-stable interning across all downstream consumers.

Effect: identical selectors that pass through multiple code paths now share interned identity end-to-end, not just once they hit MSP. This tightens the DEFEAT-8 fix from PR #3335 into a stricter caller-provides-interned-selector contract.

## Context

This is an **intermediate PR** between the staged structural fixes and the remaining "original PR 2" (Tactic 2B / `CompoundCycleTracker`):

- **PR 1 (Tactic 2A)** — merge-helper cache-key interning — **merged as #3331**.
- **PR 3 (Tactic 2C)** — `MapSetStringToPathSelectors` freeze-on-entry — **merged as #3335**.
- **This PR** — upstream `SchemaPathSelector` interning refactor.
- **Original PR 2 (Tactic 2B)** — `CompoundCycleTracker` restructure — still pending after this.

Per reviewer-flux's review (APPROVED initial + 2 delta passes): sweep-completeness invariant verified across all 13 MSP-feeding sites.

**Known semantic change worth flagging (Note 2 from Flux):** `normalizeSyncSelector` now freezes the caller's selector in place instead of cloning. This is a subtle change at a public API path. Shipping as-is per Dan's direction; CI is the filter.

Full diagnosis and per-site analysis: `coordination/docs/2026-04-16-modern-schema-hash-cache-audit.md` (§4 Phase 2, DEFEAT-8 row).

Related PRs:
- PR #3235 — `feat: enable modernSchemaHash by default` (the target; still open).
- PR #3231 — `feat: flip modernHash default to true` (upstream of #3235).
- PR #3324 — `danfuzz/because-im-easy-come-easy-go` (legacy-hash SHA256 normalization).

### DIFF
--- a/packages/memory/v2/query.ts
+++ b/packages/memory/v2/query.ts
@@ -257,7 +257,7 @@ export const trackGraph = (
     } else {
       schemaTracker.add(
         toDocKey(space, root.id, "application/json"),
-        internPathSelector(selector),
+        selector,
       );
     }
   }
--- a/packages/memory/v2/query.ts
+++ b/packages/memory/v2/query.ts
@@ -16,6 +16,7 @@ import { ExtendedStorageTransaction } from "../../runner/src/storage/extended-st
 import { ContextualFlowControl } from "../../runner/src/cfc.ts";
 import { type Immutable, isObject } from "@commonfabric/utils/types";
 import type { FabricValue, MemorySpace, URI } from "../interface.ts";
+import { internPathSelector } from "@commonfabric/data-model/schema-utils";
 import {
   type EntitySnapshot,
   type GraphQuery,
@@ -254,7 +255,10 @@ export const trackGraph = (
         sharedMemo,
       );
     } else {
-      schemaTracker.add(toDocKey(space, root.id, "application/json"), selector);
+      schemaTracker.add(
+        toDocKey(space, root.id, "application/json"),
+        internPathSelector(selector),
+      );
     }
   }
 
@@ -451,10 +455,11 @@ const loadFactsForDoc = (
   }
 
   const docKey = toDocKey(space, fact.address.id, fact.address.type);
-  if (schemaTracker.hasValue(docKey, selector)) {
+  const internedSelector = internPathSelector(selector);
+  if (schemaTracker.hasValue(docKey, internedSelector)) {
     return;
   }
-  schemaTracker.add(docKey, selector);
+  schemaTracker.add(docKey, internedSelector);
 
   if (!isObject(fact.value)) {
     return;
@@ -520,7 +525,10 @@ const evaluateTrackedDocument = (
 ) => {
   const loaded = manager.load(address);
   if (loaded === null || loaded.value === undefined) {
-    schemaTracker.add(toDocKey(space, address.id, address.type), selector);
+    schemaTracker.add(
+      toDocKey(space, address.id, address.type),
+      internPathSelector(selector),
+    );
     return;
   }
   const tracker = new CompoundCycleTracker<
--- a/packages/runner/src/traverse.ts
+++ b/packages/runner/src/traverse.ts
@@ -32,6 +32,7 @@ import {
 import { getLogger } from "../../utils/src/logger.ts";
 import { ContextualFlowControl } from "./cfc.ts";
 import {
+  DEFAULT_SELECTOR,
   internPathSelector,
   REJECTING_SELECTOR,
   schemaWithProperties,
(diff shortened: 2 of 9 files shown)


packages/runner/src/traverse.ts
```diff
@@ -345,8 +340,22 @@ export class MapSetStringToStrings extends MapSet<string, string> {
 }
 
 // SchemaPathSelectors are relative to the doc root, so if we want to look at
-// the value of the doc, we need to have "value" in the path.
-const DefaultSelector: SchemaPathSelector = { path: ["value"], schema: true };
+// the value of the doc, we need to have "value" in the path. Interned on
+// module load so that downstream users (notably `MapSetStringToPathSelectors`'
+// hash fn) can rely on the selector's immutability and the schema's WeakMap
+// cache entry.
+const DefaultSelector: SchemaPathSelector = internPathSelector({
```
danfuzz: Constants are usually named in `UPPER_SNAKE_CASE`, and in fact you do so in `v2-watch.ts` with a constant that's equivalent to `RejectingSelector` here. Two requests: (a) Name consistently. (b) DRY things out by extracting these definitions (including the one in `v2-watch.ts`) to a common place they can both be `import`ed from.

packages/memory/v2/query.ts
danfuzz, line 244: Please evaluate (and possibly implement) changing `toDocumentSelector()` to perform interning (and add documentation to that effect should you decide to implement it).

packages/runner/src/traverse.ts
danfuzz: Done in `45580eefd` (and the precursor `c3a13d373`):

- Extracted `REJECTING_SELECTOR` and `DEFAULT_SELECTOR` to `packages/data-model/schema-utils.ts` — same file as `internPathSelector`, which is their natural companion.
- Renamed to `UPPER_SNAKE_CASE` at the definitions and all call sites (including the one that was `NORMALIZED_REJECT` in `v2-watch.ts`, since it was structurally equivalent).
- Swept one additional site I had missed: `storage/cache.ts` had an inline `defaultSelector = { path: [], schema: false }` fallback that now uses `REJECTING_SELECTOR` by reference.

One subtlety, documented on both const definitions: they are built with `Object.freeze(...)` at module load rather than `internPathSelector(...)`. Calling `internPathSelector` during `schema-utils.ts`'s module-load reaches into a not-yet-initialized `schema-hash.ts` (pre-existing circular import between those two). Since `schema: false` / `schema: true` intern via the prefab boolean singletons on first real use anyway, lazy-interning is behaviorally equivalent.

— coder-cedar (Claude Opus 4.7 (1M context))

packages/memory/v2/query.ts
danfuzz: Done in `c58cf8546`:

- `toDocumentSelector()` now composes with `internPathSelector(...)` so it returns an interned, deep-frozen `DocumentSchemaPathSelector`. Added a doc comment explaining the new contract.
- Dropped the now-redundant `internPathSelector(selector)` wrap at `packages/memory/v2/query.ts:260` (the `.add()` call site that consumes `toDocumentSelector`'s result). The other `toDocumentSelector` call site at :298 flows through `evaluateTrackedDocument` into code that does its own interning at the `.add()` / `.hasValue()` boundary, so no change needed there.

— coder-cedar (Claude Opus 4.7 (1M context))

PR #3347, opened 2026-04-22
Enhance reactivity log to allow references to metadata fields instead of only value fields

ubik2:
Change many of the underlying tracking structures, especially the ReactivityLog, to track with IMemorySpaceAddress objects instead of NormalizedFullLink. The former can reference meta fields, while the latter can only reference entries under the cell's "value".

~~I'm creating this as a PR to run the automation, but still want to follow up on the TelemetryAnnotations, which are still basic links.
I also need to add some unit tests showing reactivity of meta fields (specifically "source").~~

~~Need to follow up on the reads/writes attached to actions in runner.ts, as well as BoundNodeIO (these are the same ones in the TelemetryAnnotations).~~

I'm deferring potential changes to the reads/writes attached to actions for now, since the reactivity log changes are useful whether or not we change those.

---
## Summary by cubic
Standardized reactivity and scheduler tracking on document-root `IMemorySpaceAddress` and made `NormalizedFullLink` value-relative only. Added `toMemorySpaceAddress()` so meta fields (like "source") are reactive and stopped stripping `"value"` from reactivity logs; added tests for `cfc` path reactivity (helper vs direct read).

- **Refactors**
  - Added `toMemorySpaceAddress(link)` in `link-types` and moved `ValuePath`/`IMemorySpaceValueAddress` there.
  - `NormalizedFullLink` no longer implements `IMemorySpaceAddress`; doc-root ops now convert explicitly.
  - Reactivity logs (`v2-transaction`, `reactivity-log`) keep document-root paths (prefixed with `"value"`).
  - Scheduler converts annotated writes to doc-root addresses and registers triggers using doc-root paths; removed `lookupComparableWriteValue`.
  - `readValueOrThrow`, `writeValueOrThrow`, `writeValuesOrThrow` accept `NormalizedFullLink` and convert internally.
  - Updated `link-resolution`, `schema`, and `cell` to probe/read/write via doc-root addresses, including sigil/alias checks and meta `"source"`.
  - `addressKey` now accepts `IMemorySpaceAddress` or `NormalizedFullLink`; keys differ across domains.
  - Tests updated to wrap links with `toMemorySpaceAddress(...)` when passing scheduler addresses and to expect `"value"`-prefixed paths; added `cfc` metadata reactivity tests verifying helper-based checks bypass scheduling while direct `cfc` reads are reactive.

- **Migration**
  - When subscribing or passing addresses to low-level reads/writes, wrap links with `toMemorySpaceAddress(...)`.
  - Update tests/diagnostics to expect `["value", ...]` paths in reactivity logs.

<sup>Written for commit 7dec350b239bf0c53b6d8566b14947e1c18cc260. Summary will update on new commits.</sup>

### DIFF
--- a/packages/runner/src/cell.ts
+++ b/packages/runner/src/cell.ts
@@ -93,6 +93,7 @@ import {
 import type {
   ChangeGroup,
   IExtendedStorageTransaction,
+  IMemorySpaceAddress,
   IReadOptions,
 } from "./storage/interface.ts";
 import {
@@ -1401,7 +1402,7 @@ export class CellImpl<T extends FabricValue>
   getSourceCell(schema?: JSONSchema): Cell<any> | undefined {
     if (!this.synced) this.sync(); // No await, just kicking this off
     let sourceCellId = this.runtime.readTx(this.tx).readOrThrow(
-      { ...this.link, path: ["source"] },
+      { ...this.link, path: ["source"] } as IMemorySpaceAddress,
       {
         meta: { ...ignoreReadForScheduling, ...internalVerifierRead },
       },
@@ -1440,7 +1441,7 @@ export class CellImpl<T extends FabricValue>
     }
     // System-owned provenance metadata, not user-surface value data.
     this.tx.writeOrThrow(
-      { ...this.link, path: ["source"] },
+      { ...this.link, path: ["source"] } as IMemorySpaceAddress,
       // TODO(@ubik2): Transition source links to sigil links?
       { "/": fromURI(sourceLink.id) },
     );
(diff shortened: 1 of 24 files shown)


seefeldb (approved):
lgtm,

IIUC this will trigger reruns for any cfc change since we always read that field, but i think this is acceptable until we really see this as performance issue (it should be rare).

ubik2:
> lgtm,
> 
> IIUC this will trigger reruns for any cfc change since we always read that field, but i think this is acceptable until we really see this as performance issue (it should be rare).

I think those cfc reads are tagged with ignoreReadForScheduling, so I don't think we do re-run. However, a more direct read would trigger this.
I added some unit tests for those scenarios.

PR #3351, opened 2026-04-22
Revert "feat: flip modernHash default to true"

danfuzz:
Reverts commontoolsinc/labs#3231


danfuzz:
Looks like we're committing to this! Closing without merging.

PR #3353, opened 2026-04-22
fix(shell): settle navigation side effects before same-space routing

willkelly:
## Summary
- wait for runtime-emitted same-space navigation to finish piece registration before switching the shell route
- keep the shell-side convergence wait focused on `idle()` + `synced()` instead of a broader reactive settle claim
- tighten the regression test so it waits on explicit promise milestones rather than microtask counting

## Scope of this PR
This PR is intentionally narrow.

It addresses one concrete race we found in the shell:
- `RuntimeInternals.#onNavigateRequest()` could fire same-space routing before `registerNavigatedPiece(cell)` had completed.

It does **not** claim to fully resolve the remaining `default-app` / `Pattern Integration Tests` flake. The current CI evidence suggests there is still a broader shell-readiness or render-readiness issue after returning to the space root, and that follow-up investigation should happen separately so this PR stays focused.

## Investigation
We started from GitHub Actions reruns where `Pattern Integration Tests` would fail and later reruns could succeed for the same workflow run.

The repeated symptom was:
- `packages/patterns/integration/default-app.test.ts`
- timeout while waiting for note count to increase after navigating back to the space page

During that investigation we found a separate shell race that looked real regardless of the deeper flake:
- `globalThis.commonfabric.rt.idle()` is only worker/runtime scheduler idleness
- same-space navigation in `packages/shell/src/lib/runtime.ts` was doing `registerNavigatedPiece(cell)` as fire-and-forget and then routing immediately

That meant the shell could route to the new same-space piece before the registration side effect had even been acknowledged across IPC.

## Why this fix
This PR keeps the fix narrowly scoped to the race above.

For runtime-emitted same-space navigation, we now:
1. `await registerNavigatedPiece(cell)`
2. `await idle()`
3. `await synced()`
4. route

Cross-space navigation uses the same `idle()` + `synced()` convergence helper for consistency.

After review, I simplified the first version of this PR by:
- removing the extra trailing `idle()`
- rewriting the regression test to wait on explicit deferred milestones instead of `flushMicrotasks()` bookkeeping

That keeps the PR focused on the race we actually found, without overfitting the test to internal await counts.

## What remains for follow-up
The current failing `default-app` integration case appears to be a different boundary:
- route state can be correct before the shell has fully rebound/resolved/rendered the space-root pattern
- that likely needs a follow-up fix around shell/AppView readiness or integration wait semantics, not more ordering changes in `RuntimeInternals.#onNavigateRequest()`

## Validation
Local validation run for this PR:
- `deno check packages/shell/src/lib/runtime.ts`
- `deno test --allow-read --allow-write --allow-run --allow-env packages/shell/test/runtime-navigation.test.ts`
- `deno task test` in `packages/shell`

## Reviewer context
I’d especially appreciate confirmation that this narrow shell race fix is desirable on its own, even if the broader `default-app` readiness issue still needs a separate follow-up.

### DIFF
--- a/packages/shell/src/lib/runtime.ts
+++ b/packages/shell/src/lib/runtime.ts
@@ -246,6 +246,13 @@ export class RuntimeInternals extends EventTarget {
     }
   }
 
+  async #settleNavigationSideEffects(): Promise<void> {
+    this.#check();
+    await this.#client.idle();
+    await this.#client.synced();
+    await this.#client.idle();
+  }
+
   #onConsole = (e: RuntimeClientEvents["console"][0]) => {
     const { metadata, method, args } = e;
     if (metadata?.pieceId) {
@@ -266,10 +273,10 @@ export class RuntimeInternals extends EventTarget {
       const sameSpace = cell.space() === this.#space;
 
       if (sameSpace) {
-        void this.registerNavigatedPiece(cell);
+        await this.registerNavigatedPiece(cell);
+        await this.#settleNavigationSideEffects();
       } else {
-        await this.#client.idle();
-        await this.#client.synced();
+        await this.#settleNavigationSideEffects();
       }
 
       if (sameSpace && this.#spaceName) {
--- a/packages/shell/src/lib/runtime.ts
+++ b/packages/shell/src/lib/runtime.ts
@@ -246,11 +246,10 @@ export class RuntimeInternals extends EventTarget {
     }
   }
 
-  async #settleNavigationSideEffects(): Promise<void> {
+  async #waitForNavigationConvergence(): Promise<void> {
     this.#check();
     await this.#client.idle();
     await this.#client.synced();
-    await this.#client.idle();
   }
 
   #onConsole = (e: RuntimeClientEvents["console"][0]) => {
@@ -274,9 +273,9 @@ export class RuntimeInternals extends EventTarget {
 
       if (sameSpace) {
         await this.registerNavigatedPiece(cell);
-        await this.#settleNavigationSideEffects();
+        await this.#waitForNavigationConvergence();
       } else {
-        await this.#settleNavigationSideEffects();
+        await this.#waitForNavigationConvergence();
       }
 
       if (sameSpace && this.#spaceName) {
--- a/packages/shell/test/runtime-navigation.test.ts
+++ b/packages/shell/test/runtime-navigation.test.ts
@@ -29,11 +29,26 @@ class MockRuntimeClient extends EventEmitter<MockRuntimeClientEvents> {
   }
 }
 
-async function flushMicrotasks(count = 4): Promise<void> {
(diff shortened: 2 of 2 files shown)


packages/shell/src/lib/runtime.ts
```diff
@@ -246,6 +246,13 @@ export class RuntimeInternals extends EventTarget {
     }
   }
 
+  async #settleNavigationSideEffects(): Promise<void> {
+    this.#check();
+    await this.#client.idle();
+    await this.#client.synced();
+    await this.#client.idle();
```
ubik2: This extra .idle shouldn't be needed.
Edit: one place the extra idle would matter is if our commit fails with a conflict on the server, in which case we'll roll back the change, and there may be things reactive to that rollback.
I'm not convinced that we should wait for those, though.

packages/shell/src/lib/runtime.ts
```diff
@@ -266,10 +273,10 @@ export class RuntimeInternals extends EventTarget {
       const sameSpace = cell.space() === this.#space;
 
       if (sameSpace) {
-        void this.registerNavigatedPiece(cell);
+        await this.registerNavigatedPiece(cell);
```
ubik2, line 275: This looks right to me. We should be awaiting the send (meaning our runtime client has sent the anyPiece across IPC to the runtime, and the runtime has acked it).
Subsequently awaiting the idle and synced should make it so that the runtime has run the reactive actions for that update.
I'm not certain we need .synced, but since the old clause had it, I'd keep it (that means storage is synced with the server).

packages/shell/test/runtime-navigation.test.ts
```diff
@@ -29,8 +29,14 @@ class MockRuntimeClient extends EventEmitter<MockRuntimeClientEvents> {
   }
 }
 
+async function flushMicrotasks(count = 4): Promise<void> {
```
ubik2: This looks like an agent floundering.
I get that this is the `await send, await idle, await synced, await idle`, but it's really messy.
Note that in the first test case, we're actually skipping the real registerNavigatedPiece, so we aren't doing those awaits. We do use them in the second test.
Generally, the emit should result in 2 (different space) or 3 (same space) awaits (3 or 4 with the extra idle).

ubik2: Actually, maybe just add a comment here explaining that these awaits are interleaved with the awaits used by the underlying system, and we want as many here as there are there.
That does seem like a very fragile interface, since I certainly wouldn't expect adding an extra await in registerNavigatedPiece to change this behavior.

ubik2 (approved):
I'm approving this, but maybe ping your agent to see if the second idle can be removed.
Also, can you add a comment for flushMicrotasks? That's apparent to me now after digging into it, but won't be apparent to future me.

PR #3373, opened 2026-04-24
perf-check: always emit a full table of every measured metric

danfuzz:
## Summary

- Adds a full per-metric table to the `perf-check` CI job's console output, always emitted (not just on failure), so the log shows every measured metric's baseline median, threshold, and current value at a glance.
- Keeps the existing failure flow intact: the `!!! PERFORMANCE REGRESSION DETECTED !!!` callout stays prominent at the top, and the `---BEGIN COPY-PASTE---` override block stays pinned at the very bottom so it's easy to find.
- Table is grouped by metric kind (Jobs / Steps / Test files / Subtests / Benchmarks) and sorted most-at-risk-first (`OVER` → `CLOSE` → `OK` by descending head%, then excluded/insufficient-data rows).

## Rationale

Before: the pass case was a one-liner (`All metrics within normal range.`), so there was no way to see how close any metric was to its threshold, and the failure case only surfaced the failing metrics. It was hard to get a feel for "is this PR running close to the edge, or comfortably under?"

After: a reader can scan the `head%` column to see what's near the ceiling and which metrics have margin to spare, without needing a separate off-CI analysis pass.

## Example

With `GITHUB_RUN_ID=24863831900 PR_NUMBER=3370` (the PR #3370 run, green on the live baseline), the new section includes:

```
## All metrics checked  (75 total — OVER: 0, CLOSE: 0, OK: 69, ovrd: 0, excl: 6, n/a: 0)

### Jobs  (13)
| status | head% | Δ% | current | baseline median | threshold | n | metric |
|:---|---:|---:|---:|---:|---:|---:|:--|
| OK | 45% | +14.3% | 2m 0s | 1m 45s | 2m 18s | 13 | job: CLI Integration Tests (fuse) |
| OK | 8% | +2.1% | 1m 13s | 1m 12s | 1m 31s | 20 | job: Check |
| OK | 1% | +0.2% | 3m 41s | 3m 41s | 4m 33s | 20 | job: CLI Integration Tests (core) |
...
| excl | -32% | -10.6% | 55.0s | 1m 2s | 1m 22s | 20 | job: Pattern Unit Tests (4/5) |
```

For a simulated failure case (thresholds artificially tightened locally), the top of the log now reads:

```
!!! PERFORMANCE REGRESSION DETECTED in 11 metric(s) !!!

| Metric | Current | Baseline (median) | Threshold | Change |
...

Baseline sample breakdown:
  <per-failure sample list>

[full metric table here]

To accept these regressions, add the following to your PR description:

---BEGIN COPY-PASTE---
NEW_PERF_BASELINE: ...
---END COPY-PASTE---
```

## Status legend

- `OVER` — over the threshold; the check fails.
- `CLOSE` — ≥50% of the median→threshold margin consumed.
- `OK` — <50% of the margin consumed.
- `ovrd` — saved by a `NEW_PERF_BASELINE:` override in the PR body.
- `excl` — excluded from the check (e.g., `Pattern Unit Tests` aggregates that grow as tests are added).
- `n/a` — fewer than `MIN_SAMPLES` baseline samples.

`head%` = `(current - median) / (threshold - median) × 100`, i.e. how much of the median-to-threshold margin the current value has consumed.

## Test plan

- [x] `deno check tasks/perf-check.ts tasks/perf-lib.ts tasks/perf-regression.ts` — clean.
- [x] `deno test tasks/perf-lib.test.ts` — all 3 tests pass.
- [x] Ran the modified script against the live PR #3370 run (green case) and confirmed the full table renders correctly, grouped and sorted as intended.
- [x] Simulated a failing run by monkey-patching `baseline.threshold` locally; confirmed the `!!! PERFORMANCE REGRESSION DETECTED !!!` callout appears at top, per-failure sample breakdown still renders, full table still renders in the middle, and the `---BEGIN COPY-PASTE---` block stays pinned at the bottom.

### DIFF
--- a/tasks/perf-check.ts
+++ b/tasks/perf-check.ts
@@ -380,11 +380,15 @@ async function main() {
     }
   };
   const KIND_ORDER = ["Jobs", "Steps", "Test files", "Subtests", "Benchmarks"];
+  // Sort order within each kind: most at-risk of failing the check first.
+  // `ovrd` sits below `OK` because an override-protected metric is at strictly
+  // lower risk of tripping the check than an unguarded OK metric — the author
+  // has already authorized its current level.
   const STATUS_ORDER: Record<Status, number> = {
     OVER: 5,
     CLOSE: 4,
-    ovrd: 3,
-    OK: 2,
+    OK: 3,
+    ovrd: 2,
     excl: 1,
     "n/a": 0,
   };
--- a/tasks/perf-check.ts
+++ b/tasks/perf-check.ts
@@ -217,110 +217,238 @@ async function main() {
   }
 
   // 5. Compare current metrics against baseline
-  const failures: {
+  type Status = "OVER" | "CLOSE" | "OK" | "ovrd" | "excl" | "n/a";
+  interface Row {
     metric: string;
+    status: Status;
     current: number;
-    median: number;
-    variance: number;
-    stddev: number;
-    threshold: number;
-    pctIncrease: number;
-  }[] = [];
+    median?: number;
+    variance?: number;
+    stddev?: number;
+    threshold?: number;
+    n: number;
+    pctIncrease?: number;
+    /**
+     * How much of the median→threshold margin the current value has consumed,
+     * as a percentage. 0% = at median; 100% = at threshold; >100% = over.
+     */
+    headroomPct?: number;
+  }
+
+  const rows: Row[] = [];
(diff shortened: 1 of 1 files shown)


bfollington (approved):
LGTM! Thanks, useful to see all the deltas to help understand if it's just noise or not.

PR #3375, opened 2026-04-24
perf(cli): emit cpu metrics for pattern tests

willkelly:
## Summary
- add an opt-in `--perf-json` mode to `cf test` that emits one machine-readable metrics record per completed test file
- capture per-file CPU time (`cpuUsage`) plus memory snapshots in the pattern test harness
- extend `tasks/integration.ts` to collect those child-emitted metrics and write a sibling `pattern-unit-metrics.json` artifact next to the existing JUnit timing file
- parse Deno JUnit XML that omits suite-level `time` and label multi-XML artifacts so shell/package integration test timings are visible
- demote generic integration step wall-clock overages to `WATCH` unless related test/subtest timings corroborate the regression

## Why
The current perf gate includes wall-clock CI timings that are useful telemetry but can be too noisy to treat as benchmark-grade signal on their own. Repeated sampling showed flakiness in pattern-unit wall-clock subprocess timings and short browser integration steps. This PR adds CPU metrics for pattern tests and uses underlying JUnit test timings to corroborate step-level wall-clock outliers.

## Validation
- `deno task cf test --root packages/patterns packages/patterns/cfc-render-policy-demo/main.test.tsx --perf-json --memory-version v2`
- `CF_INTEGRATION_PERF_JSON=1 deno task integration --junit-dir=/tmp/pattern-metrics-check pattern-tests 1/5`
- `deno test --allow-env packages/cli/test/test-runner.test.ts`
- `deno test --allow-env tasks/perf-lib.test.ts`
- `deno task check`
- replayed `tasks/perf-check.ts` against PR #3375 run `24971407249`; extracted 736 metrics and reported `OVER: 0`, `WATCH: 0`

## Notes
- cubic identified that `passed` in perf JSON was only checking `result.error`; this branch now uses the same failure accounting as the pattern test runner summary.

### DIFF
--- a/.github/workflows/deno.yml
+++ b/.github/workflows/deno.yml
@@ -7,6 +7,12 @@ on:
   pull_request:
     branches:
       - main
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: "Pull request number to use for performance checks"
+        required: false
+        type: string
 
 # Define the list of binaries we're building
 # This makes it easy to add more binaries in the future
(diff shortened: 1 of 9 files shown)


willkelly:
Temporarily closing/reopening to trigger a fresh pull_request CI run for perf stability sampling.

PR #3381, opened 2026-04-24
fix(ts-transformers): keep callable closure captures lexical

mathpirate:
# Handover note (from Gideon + Claude review session, 2026-05-13)

This PR is in draft because the scope needs to change. The bug class it targets is real and still present on main today, but Berni has proposed a different fix direction for part of it. This note captures the empirical findings and the agreed-on split so this can be picked back up cleanly.

## 1. The bug class is real and still on main today

Verified empirically by extracting each fixture's `.input.tsx` from this branch, placing it on current `main`, and running `deno task cf check --show-transformed --no-run`. All four fixtures (`action-`, `derive-`, `handler-`, `map-captured-callable-export-binding`) produce the same broken shape on main: the captured callable is routed through an explicit `inputs`/`state` object that has an **empty schema** (no slot for the callable), yet the callback destructures the same name and the call site populates it. The runtime cannot faithfully transport a callable through that path.

### Concrete example — the `action` fixture

**Input:**

```tsx
import { action } from "commonfabric";

function makeAction(helper: (value: string) => string) {
  return action(() => {
    return helper("x");
  });
}

const helper = (value: string) => value.toUpperCase();
const myAction = makeAction(helper);

export default myAction;
```

**On main today** (the bug):

```ts
function makeAction(helper: (value: string) => string) {
    return __cfHelpers.handler(
        false as const satisfies __cfHelpers.JSONSchema,
        {
            type: "object",
            properties: {}                 // ← empty schema, no slot for `helper`
        } as const satisfies __cfHelpers.JSONSchema,
        (_, { helper }) => {               // ← destructures `helper` anyway
            return helper("x");
        }
    )({
        helper: helper                     // ← passes via state, which the schema doesn't permit
    });
}
```

**On this branch** (codex's fix):

```ts
function makeAction(helper: (value: string) => string) {
    return __cfHelpers.handler(
        false as const satisfies __cfHelpers.JSONSchema,
        {
            type: "object",
            properties: {}
        } as const satisfies __cfHelpers.JSONSchema,
        (_, __cf_action_params) => {       // ← anonymous placeholder, doesn't touch helper
            return helper("x");             // ← helper resolves lexically to outer parameter
        }
    )({});                                 // ← empty inputs, matches the empty schema
}
```

### Same bug, `derive` fixture

**Input:**

```tsx
import { derive, pattern } from "commonfabric";

function makePattern(helper: (value: string) => string) {
  return pattern(() => {
    return {
      label: derive("x", () => helper("x")),
    };
  });
}

const helper = (value: string) => value.toUpperCase();
const myPattern = makePattern(helper);

export default myPattern;
```

**On main today** (the bug):

```ts
function makePattern(helper: (value: string) => string) {
    return pattern(() => {
        return {
            label: __cfHelpers.derive(
                { type: "object", properties: {} } as const ...,    // ← empty schema
                { type: "string" } as const ...,
                { helper: helper },                                  // ← but inputs has helper
                ({ helper }) => helper("x")                          // ← destructured anyway
            ).for(["__patternResult", "label"], true)
        };
    }, ...);
}
```

**On this branch** (codex's fix): `helper` left lexical inside the derive callback; inputs argument dropped; schema stays `properties: {}` consistently.

The `handler-` and `map-` fixtures exhibit the same shape inconsistency on main (verified for `action-` and `derive-`; not individually re-run for the other two, but the structural pattern is identical).

## 2. Berni's proposed direction changes the scope

Per discussion with Berni (2026-05-13): we should add a **compile-time placement validator** for `action()` requiring it to appear only in allowed contexts (pattern body, computed body, handler body, map callback). This is in addition to the existing `lift()`/`handler()` placement validator in `packages/ts-transformers/src/transformers/pattern-context-validation.ts:583` (`validateBuilderPlacement`), which already rejects direct `lift()`/`handler()` calls inside restricted reactive contexts.

The existing validator does NOT today reject:

- `action()` placement (no validator for `action()` yet — this is what Berni wants added).
- The codex `action` fixture's shape: `function makeAction(helper) { return action(() => …) }` — `action()` at the top of a plain function, not in a restricted context. Confirmed empirically: main passes type check on this input, then emits the bad transform.

Once the new validator lands:

- **The `action-captured-callable-export-binding` fixture becomes a compile error** at validation time. The action-strategy's callable-capture filter never runs on it. → Action piece of this branch becomes dead code.
- **The `handler-captured-callable-export-binding` fixture (uses inline JSX `onClick`) may or may not be rejected**, depending on how the validator treats inline JSX event handlers (which lower into `handler()` but aren't direct calls). Needs a focused experiment.
- **The `derive-` and `map-captured-callable-export-binding` fixtures are NOT rejected.** Their placement is fine (`derive`/`.map()` inside a pattern body is an allowed context). The bug is in *what the callback captures*, not where it's placed. → Derive and map pieces of this branch are still needed.

## 3. Recommended split

This PR's contents should split into three:

### Keep (pick up here)

The **derive-strategy and array-method-transform changes** plus their fixtures (`derive-captured-callable-export-binding`, `map-captured-callable-export-binding`). These solve a real bug class that survives Berni's proposed validator.

The supporting machinery these depend on:

- `packages/ts-transformers/src/closures/utils/callable-capture-filter.ts` — keep.
- `packages/ts-transformers/src/closures/utils/non-hoistable-callbacks.ts` — keep; the hoisting registry is still load-bearing for the surviving paths (a callback that legitimately captures a callable from its enclosing scope cannot be hoisted to module scope without breaking the reference).
- `packages/ts-transformers/src/closures/module-scope-callback-hoisting.ts` — the recursive `nestedFunctionCapturesEnclosingBindings` descent is still needed.

### Drop (made redundant by the validator)

- The **action-strategy changes** (`packages/ts-transformers/src/closures/strategies/action-strategy.ts`).
- The **action fixture** (`action-captured-callable-export-binding.input.tsx` + golden).
- Possibly the **handler-strategy changes** and **handler fixture** — see #4 below.

### Validator work (new territory — does NOT reuse this branch)

Extend `packages/ts-transformers/src/transformers/pattern-context-validation.ts` to:

- Add an `action()` placement rule. Allowed contexts: pattern body, computed body, handler body, map callback. Anything else → diagnostic.
- Modeled on the existing `validateBuilderPlacement` (line 583) for `lift()`/`handler()`.

## 4. Open question on the handler-strategy piece

The `handler-captured-callable-export-binding` fixture uses an inline JSX event handler (`<cf-button onClick={() => helper("x")}>` inside a `pattern()` body). This isn't a direct `handler()` call site that the existing or proposed validator would catch — it lowers into `handler()` machinery, but the source-side construct is JSX-bound.

Open question: should the validator extend to inline JSX handler bodies, or are those a legitimate context for callable captures? Needs a focused experiment:

1. Construct an inline JSX `onClick={() => helper(...)}` shape inside a placement context that *is* allowed by the validator.
2. Run through main; see whether the bug still surfaces.
3. If yes → keep the handler-strategy piece of this branch.
4. If no (or if Berni decides inline JSX handlers should be banned in this shape) → drop the handler piece too.

## 5. What's been verified so far

- All four fixture inputs reproduce the bug on current `main` (transformer emits inconsistent empty-schema-but-populated-state output).
- The full ts-transformers suite passes on this branch's own state (255/0).
- This branch does NOT cleanly rebase onto current main — there's a conflict in `packages/ts-transformers/src/closures/module-scope-callback-hoisting.ts` with PR #3550's debug-only `non-hoistable-callback` diagnostic. The two changes touch the same lines of the `visit` function but aren't logically contradictory: codex's recursive nested-function descent and #3550's diagnostic should coexist (the diagnostic just observes what the hoist check decided). Manual rebase needed.

## 6. Test verification (originally from the PR description)

- `deno test -A test/closures/module-scope-helper-hoisting.test.ts`
- `env FIXTURE_PATTERN=captured-callable-export-binding deno test -A test/fixture-based.test.ts`
- `deno test -A test/fixture-based.test.ts`
- `git diff --check`

---

# Original PR description below

## Problem
- closure strategies were rewriting direct callable captures into generated closure state or params
- those generated transport objects do not faithfully carry plain callables today
- the rewritten callback could therefore shadow a valid lexical helper with an undefined generated binding

## Approach
- keep direct callable root captures lexical instead of routing them through generated closure state or params
- apply that rule across `action`, JSX `handler`, `derive`, and reactive array-method callback lowering
- mark callbacks and enclosing callbacks non-hoistable when they rely on those lexical callable captures
- add targeted fixtures for each affected strategy plus a hoisting regression

## Why The Hoisting Change Is Here
Once a closure strategy leaves callable captures lexical, module-scope callback hoisting must not pull those callbacks away from the enclosing bindings they still depend on. The new non-hoistable callback registry is the coordination seam for that rule.

## Out Of Scope
This PR does not change the separate SES or plain-data behavior for exported actions or dynamic pattern creation. It only fixes the earlier transformer-side bad shape.

## Verification
- `deno test -A test/closures/module-scope-helper-hoisting.test.ts`
- `env FIXTURE_PATTERN=captured-callable-export-binding deno test -A test/fixture-based.test.ts`
- `deno test -A test/fixture-based.test.ts`
- `git diff --check`

### DIFF
--- a/packages/ts-transformers/src/cf-pipeline.ts
+++ b/packages/ts-transformers/src/cf-pipeline.ts
@@ -113,6 +113,7 @@ export class CommonFabricTransformerPipeline extends Pipeline {
     const ops: TransformationOptions = {
       typeRegistry: new WeakMap(),
       mapCallbackRegistry: new WeakSet(),
+      nonHoistableCallbackRegistry: new WeakSet(),
       syntheticComputeCallbackRegistry: new WeakSet(),
       syntheticComputeOwnedNodeRegistry: new WeakSet(),
       syntheticReactiveCollectionRegistry: new WeakSet(),
--- a/packages/ts-transformers/src/closures/module-scope-callback-hoisting.ts
+++ b/packages/ts-transformers/src/closures/module-scope-callback-hoisting.ts
@@ -38,6 +38,7 @@ export function hoistModuleScopedBuilderCallbacks(
       }
       if (
         !isFunctionLikeExpression(argument) ||
+        context.isNonHoistableCallback(argument) ||
         !isNestedWithinFunction(argument) ||
         !callbackCanBeHoisted(argument, context.checker)
       ) {
@@ -150,10 +151,6 @@ function callbackCanBeHoisted(
       node !== callback &&
       isFunctionLikeDeclaration(node)
     ) {
-      if (nestedFunctionCapturesEnclosingBindings(node, callback, checker)) {
-        capturesEnclosingBindings = true;
-        return true;
-      }
       return false;
     }
 
@@ -181,35 +178,6 @@ function callbackCanBeHoisted(
   }
 
   return usesModuleScopedReferences && !capturesEnclosingBindings;
-}
-
-function nestedFunctionCapturesEnclosingBindings(
-  nested: ts.FunctionLikeDeclaration,
-  callback: ts.FunctionLikeDeclaration,
-  checker: ts.TypeChecker,
-): boolean {
-  const visit = (node: ts.Node): boolean => {
-    if (node !== nested && isFunctionLikeDeclaration(node)) {
-      return false;
-    }
-
-    if (
-      ts.isIdentifier(node) &&
-      getReferenceScope(node, callback, checker) === "enclosing"
-    ) {
-      return true;
-    }
-
-    return ts.forEachChild(node, (child) => visit(child)) ?? false;
-  };
-
-  for (const parameter of nested.parameters) {
-    if (parameter.initializer && visit(parameter.initializer)) {
-      return true;
-    }
-  }
-
-  return nested.body ? visit(nested.body) : false;
 }
 
 type ReferenceScope = "local" | "module" | "enclosing" | "other";
--- a/packages/ts-transformers/src/closures/module-scope-callback-hoisting.ts
+++ b/packages/ts-transformers/src/closures/module-scope-callback-hoisting.ts
@@ -150,6 +150,10 @@ function callbackCanBeHoisted(
       node !== callback &&
       isFunctionLikeDeclaration(node)
     ) {
+      if (nestedFunctionCapturesEnclosingBindings(node, callback, checker)) {
+        capturesEnclosingBindings = true;
+        return true;
+      }
       return false;
     }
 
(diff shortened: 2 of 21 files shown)


mathpirate:
Closing in favor of the different approach taken in https://github.com/commontoolsinc/labs/pull/3566

PR #3392, opened 2026-04-27
docs(spec): split JSON encoding into its own section + add /-key reservation rule

danfuzz:
## What

Extracts the JSON encoding material from `1-fabric-values.md` Section 5 into a new standalone spec file `3-json-encoding.md`, and adds the `/`-key reservation rule as a new section within it.

## Changes

- **`docs/spec/3-json-encoding.md`** (new file) — JSON encoding spec extracted from `1-fabric-values.md` Section 5. Subsections renumbered as top-level sections 1–8. New **Section 9** added: specifies the `/`-key reservation rule:
  - Plain objects with any key beginning with `/` are reserved for tagged-value encoding.
  - A bare `"/"` key (empty tag name) is always a decoder error.
  - Unrecognized `/-prefixed` tags must be treated as `UnknownValue`.
- **`docs/spec/1-fabric-values.md`** — Section 5 replaced with a one-paragraph stub cross-referencing `3-json-encoding.md`. All internal Section 5.x cross-references updated to point to the new file.
- **`docs/spec/README.md`** — updated to list all three spec files with accurate per-file scope descriptions.

## Why

The `/`-key reservation rule is now implemented in `JsonEncodingContext` (landed in the `coder-cedar/fix-deserializer-slash-key` companion PR) and enabled by default via the `coder-cedar/enable-unified-json-encoding` flag-flip PR (#3391). Having the rule specified formally in the spec closes the implementation↔spec gap and provides a durable reference for future changes.

Separating JSON encoding into its own spec section also makes the document more navigable: `1-fabric-values.md` was growing unwieldy mixing type definitions with serialization protocol details.

## Review

Reviewed and approved by reviewer-flux (Claude Sonnet 4.6), session 2026-064.

Co-Authored-By: spec-writer-sage (Claude Sonnet 4.6) <noreply@anthropic.com>
Co-Authored-By: upstream-liaison-bolt (Claude Sonnet 4.6) <noreply@anthropic.com>

PR #3394, opened 2026-04-27
fix(data-model): produce ProblematicValue for objects with /-prefixed keys

danfuzz:
## What

In the modern JSON decoder (`packages/data-model/json-encoding-modern.ts`), deserializing a plain object that contains **any** key starting with `/` now returns a `ProblematicValue` instead of treating it as a plain object.

**Commit:** `66da227b5`

## Why

The `/`-key namespace is wholly reserved for tagged values (e.g. `{ "/Error@1": ... }`, `{ "/object": ... }`). A plain object reaching the decoder with a `/`-prefixed key is a structural violation — it should have been escaped by the serializer via the `/object` escape mechanism (Section 6 of the formal spec). Passing it through as an ordinary plain object would silently misrepresent data that violates the encoding contract.

`ProblematicValue` is the correct result here: it signals a structural/encoding problem rather than an unknown-but-valid tag. The prior code would have treated multi-key objects with a `/`-prefixed key as plain objects, which is incorrect per the reservation rule.

## Spec alignment

This implements the `/`-key reservation rule specified in `docs/specs/space-model-formal-spec/3-json-encoding.md` Section 9 (PR #3392): any object containing any key starting with `/` is a reserved form, and if the decoder cannot interpret it as a valid tagged value or escape, it must return `ProblematicValue`.

## Scope out

The single-key tagged-value detection path (e.g. `{ "/Error@1": ... }`) is unchanged. This PR adds the multi-key `/`-prefixed key guard that was previously missing.

## Review

Reviewed and approved by reviewer-flux (Claude Sonnet 4.6), session 2026-064.

Co-Authored-By: coder-cedar (Claude Sonnet 4.6) <noreply@anthropic.com>
Co-Authored-By: upstream-liaison-bolt (Claude Sonnet 4.6) <noreply@anthropic.com>

### DIFF
--- a/packages/data-model/json-encoding-modern.ts
+++ b/packages/data-model/json-encoding-modern.ts
@@ -419,7 +419,18 @@ export class JsonEncodingContext implements SerializationContext<string> {
       return Object.freeze(result);
     }
 
-    // Plain objects: recursively deserialize values, then freeze.
+    // Plain objects: guard against reserved /-prefixed keys, then recursively
+    // deserialize values and freeze. Any /-prefixed key in ANY object is
+    // reserved per spec — return ProblematicValue rather than silently
+    // round-tripping the object.
+    const slashKey = Object.keys(data).find((k) => k.startsWith("/"));
+    if (slashKey !== undefined) {
+      return new ProblematicValue(
+        slashKey,
+        data as unknown as FabricValue,
+        `object contains reserved /-prefixed key: "${slashKey}"`,
+      ) as unknown as FabricValue;
+    }
     const result: Record<string, FabricValue> = {};
     for (const [key, val] of Object.entries(data)) {
       result[key] = this.deserialize(val, runtime, registry);
--- a/packages/data-model/json-encoding-modern.ts
+++ b/packages/data-model/json-encoding-modern.ts
@@ -287,9 +287,10 @@ export class JsonEncodingContext implements SerializationContext<string> {
 
     seen.delete(value as object);
 
-    // Apply `TAGS.object` escaping per Section 5.6.
+    // Apply `TAGS.object` escaping per Section 5.6: wrap any plain object
+    // that has at least one /-prefixed key, not just single-key objects.
     const keys = Object.keys(result);
-    if (keys.length === 1 && keys[0].startsWith("/")) {
+    if (keys.some((k) => k.startsWith("/"))) {
       return this.wrapTag(TAGS.object, result) as JsonWireValue;
     }
 
--- a/packages/data-model/json-encoding-modern.ts
+++ b/packages/data-model/json-encoding-modern.ts
@@ -419,20 +419,18 @@ export class JsonEncodingContext implements SerializationContext<string> {
       return Object.freeze(result);
     }
 
-    // Plain objects: guard against reserved /-prefixed keys, then recursively
-    // deserialize values and freeze. Any /-prefixed key in ANY object is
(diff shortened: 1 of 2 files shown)


packages/data-model/json-encoding-modern.ts
```diff
@@ -419,7 +419,18 @@ export class JsonEncodingContext implements SerializationContext<string> {
       return Object.freeze(result);
     }
 
-    // Plain objects: recursively deserialize values, then freeze.
+    // Plain objects: guard against reserved /-prefixed keys, then recursively
+    // deserialize values and freeze. Any /-prefixed key in ANY object is
```
danfuzz: Instead of doing a "preflight" loop, how about putting the `/` test into the loop below?

packages/data-model/json-encoding-modern.ts
```diff
@@ -419,7 +419,18 @@ export class JsonEncodingContext implements SerializationContext<string> {
...
+    // reserved per spec — return ProblematicValue rather than silently
+    // round-tripping the object.
+    const slashKey = Object.keys(data).find((k) => k.startsWith("/"));
+    if (slashKey !== undefined) {
+      return new ProblematicValue(
+        slashKey,
```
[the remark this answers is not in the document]
danfuzz: Good catch — fixed in the follow-up commit. The offending key (e.g. `/b`) was passed directly as `typeTag`, but `wrapTag()` prepends `/` when serializing, producing a double-slash (`//b`) on re-serialization. The fix strips the leading `/` before constructing `ProblematicValue` (`key.slice(1)`), so round-trip wire form is correct. Also folded the guard into the deserialization loop (single pass, no preflight `find`) per Dan's suggestion.

packages/data-model/test/json-encoding-modern.test.ts
```diff
@@ -962,18 +962,104 @@ describe("json encoding", () => {
...
+      const result = roundTrip(obj) as Record<string, FabricValue>;
+      expect(result["/x"]).toBeInstanceOf(FabricEpochDays);
+      expect((result["/x"] as unknown as FabricEpochDays).value).toBe(42n);
+    });
+
+    it("single-key /-prefixed object still routes through unwrapTag (no regression)", () => {
+      // Single-key /Tag@1 objects are handled by unwrapTag, not the plain-object
+      // path — confirm they still produce UnknownValue (unrecognized tag), not
+      // ProblematicValue from the new multi-key guard.
+      const data = { "/Unknown@1": { id: "x" } } as JsonWireValue;
```
danfuzz: To aid understanding here, how about a type tag that doesn't have "unknown" in the name, and also has a non-`1` version number, _and_ check the `.tag` of the result?

danfuzz:
Recursive-encoding tests added in commit `6216553ae`: covers doubly-nested plain objects with `/`-prefixed keys, `FabricError` as a value inside a `/`-prefixed-key object, and `FabricEpochDays` as a value. Cedar confirmed no impl gaps — the encoder already recurses values before the wrap decision, so round-trip behavior is correct for all nested cases.

packages/data-model/test/json-encoding-modern.test.ts
danfuzz: Addressed in commit `b2ecedb11` — tag changed to `/Future@7`, `.tag` assertion added, and comment updated to "any well-formed but unregistered tag".

PR #3395, opened 2026-04-27
test(memory): run v2-test under both unifiedJsonEncoding states

danfuzz:
## What

Refactors `packages/memory/test/v2-test.ts` to run the boundary-decode test suite under both `unifiedJsonEncoding = false` and `unifiedJsonEncoding = true` using the dual-flag loop pattern from `schema-hash.test.ts`.

**Commit:** `767ab46d1`

## Changes

- The boundary-decode tests (encode/decode round-trip at the memory v2 boundary) are wrapped in a loop over `[false, true]`, exercising both flag states.
- Flag-independent tests (protocol constants, documents, paths, source links) remain outside the loop — they don't touch the JSON encoding path and don't need duplication.
- Memory test suite grows from 152 to 164 steps (12 new from the dual-flag loop).

## Why

This pre-PR must land before the `unifiedJsonEncoding` flag-flip PR (`coder-cedar/enable-unified-json-encoding`, PR #3391) so the flip does not need to carry test changes. Having the boundary tests run under both states ensures the flip doesn't silently break the `flag = false` (legacy) path, and confirms the `flag = true` (modern) path was already correct before the default changes.

## Review

Reviewed and approved by reviewer-flux (Claude Sonnet 4.6), session 2026-064.

Co-Authored-By: coder-cedar (Claude Sonnet 4.6) <noreply@anthropic.com>
Co-Authored-By: upstream-liaison-bolt (Claude Sonnet 4.6) <noreply@anthropic.com>

### DIFF
--- a/packages/memory/test/v2-test.ts
+++ b/packages/memory/test/v2-test.ts
@@ -1,3 +1,4 @@
+import { afterEach, beforeEach, describe, it } from "@std/testing/bdd";
 import { assert, assertEquals, assertFalse } from "@std/assert";
 import {
   resetDataModelConfig,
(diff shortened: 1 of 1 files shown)


packages/memory/test/v2-test.ts
```diff
@@ -44,186 +45,193 @@ const toEntityDocument = (
...
+      canonicalHashing: true,
+      modernSchemaHash: true,
+    });
+
+    resetDataModelConfig();
+    resetJsonEncodingConfig();
+    resetModernHashConfig();
+    resetSchemaHashConfig();
   });
-  assertFalse(Object.isFrozen(decoded));
```
danfuzz, line 182: Did these `assertFalse()`s get dropped in the PR? (I didn't find where they went.)

danfuzz: Not dropped — consolidated. On `main`, there were two separate blocks of three `assertFalse(Object.isFrozen(...))` calls: one in the legacy-decode test and one in the unified-decode test. In this PR they are merged into a single block of three at ~line 352, inside the `for (const unifiedJsonEncoding of [false, true])` loop. Those assertions now execute twice — once with `unifiedJsonEncoding = false` and once with `unifiedJsonEncoding = true`. Same assertion coverage, now structured to run under both flag states.

PR #3398, opened 2026-04-27
Add cf-harness CFC manifest intake

mathpirate:
## Summary
- align cf-harness prompt-slot evidence with the CFC PromptSlotBound atom shape
- add Loom run manifest parsing, CLI intake, artifact retention, and run-state plumbing
- default cf-harness CFC mode to enforce-explicit with env overrides and add CFC substrate metadata to capability snapshots

## Verification
- deno task check (labs root)
- git diff --check

### DIFF
--- a/packages/cf-harness/src/contracts/prompt-slot.ts
+++ b/packages/cf-harness/src/contracts/prompt-slot.ts
@@ -1,42 +1,181 @@
+import { isRecord } from "@commonfabric/utils/types";
+
 export type PromptSlotRole = "direct-command" | "context" | "quote";
 
+export const CFC_PROMPT_SLOT_BOUND_ATOM_TYPE =
+  "https://commonfabric.org/cfc/atom/PromptSlotBound" as const;
+
+export type PromptSlotReference = string | Record<string, unknown>;
+
+export interface PromptSlotRenderRef {
+  seq: number;
+  rootRef: PromptSlotReference;
+}
+
 export interface PromptSlotBinding {
-  type: "cf-harness.prompt-slot-binding";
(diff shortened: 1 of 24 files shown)


packages/cf-harness/src/contracts/prompt-slot.ts
```diff
@@ -1,42 +1,179 @@
...
+  targetPath?: string;
 }
 
+const PROMPT_SLOT_ROLES: readonly PromptSlotRole[] = [
+  "direct-command",
+  "context",
+  "quote",
+];
+
+const isRecord = (input: unknown): input is Record<string, unknown> =>
```
seefeldb: import from @commontools/utils/types instead

packages/cf-harness/src/contracts/run-manifest.ts
```diff
@@ -0,0 +1,138 @@
...
+  cfc?: LoomRunManifestCfc;
+  extra?: Record<string, unknown>;
+}
+
+export type HarnessRunManifest = LoomRunManifest;
+
+const isRecord = (input: unknown): input is Record<string, unknown> =>
+  typeof input === "object" && input !== null && !Array.isArray(input);
+
+const CFC_ENFORCEMENT_MODES: readonly CfcEnforcementMode[] = [
```
seefeldb: import from runner package instead

packages/cf-harness/src/contracts/prompt-slot.ts
mathpirate: done thanks

packages/cf-harness/src/contracts/run-manifest.ts
mathpirate: done thanks

PR #3402, opened 2026-04-28
[codex] mediate runsc CFC sandbox results in cf-harness

willkelly:
## Summary

This PR teaches `cf-harness` to consume trusted CFC result sidecars emitted by the local runsc-CFC Docker runtime and to mediate sandbox output before it can enter the model context.

The harness now creates containers explicitly, attaches to collect stdout/stderr, waits for completion, reads the trusted sidecar keyed by container ID, and attaches a structured `CfcSandboxResult` to command results. Prompt-loop enforcement then withholds or replaces unsafe observations instead of forwarding raw tainted bytes to the model.

## CFC Fixes

- Added a shared `CfcSandboxResult` contract in `packages/runner`, including stream, exit, diagnostic, observed, opaque, and denied result states.
- Updated the Docker runsc sandbox implementation from `docker run --rm` to an explicit `docker create` / `docker start --attach` / `docker wait` / sidecar-read / cleanup lifecycle. This is necessary because the sidecar name is keyed by the container ID, and because Docker's runtime wait path does not itself return the CFC label metadata.
- Added `cfcResultDir` configuration plus `CF_HARNESS_RUNSC_CFC_RESULT_DIR` support so deployments can point the harness at the trusted result directory configured on runsc.
- Parse public sidecars (`xattrJSON:{}`) as observed stdout/stderr/exit status, and parse non-empty CFC labels as opaque sandbox observations rather than raw model-visible text.
- In enforce modes, deny bash output that lacks CFC metadata. This prevents accidental regression to trusting stdout when the sidecar is missing or malformed.
- Updated the prompt loop so tainted, opaque, malformed, or denied sandbox observations are mediated before being appended to model input.
- Added end-to-end integration coverage using a real policy-labeled `/secrets/space.json` mount to prove that tainted runsc output is withheld from the model.

## macOS Docker Desktop Follow-Up

- On macOS, leave Docker containers on the image default user unless `containerUser` is configured explicitly. Docker Desktop bind mounts are mediated through a Linux VM, so the macOS UID/GID does not imply writable container access.
- Use the Docker Desktop setup helper's host result directory by default: `$HOME/.local/share/runsc-cfc/cfc-results` for `cf-harness`, and `/host_mnt/.../cfc-results` for the Docker runtime arg.
- Let CFC integration tests and the Labs FUSE validation use the macOS-installed policy file at `$HOME/.local/share/runsc-cfc/cfc-policy.json` by default, with env overrides for custom locations.
- Put macOS integration temp workspaces under `$HOME/.cache/cf-harness/integration` so Docker Desktop can bind-mount them reliably.

## Test Task Hygiene

- Removed `--no-check` from package test tasks in `content-hash`, `data-model`, `memory`, `runtime-client`, and `ui`, so repo-level `deno task test` actually type-checks those test suites.
- Updated the cf-harness integration prompt-slot fixture to use the canonical `PromptSlotBound` atom so `deno task test:integration` type-checks.

## Why Sidecar Files For Now

We use a host/runtime-owned sidecar rather than stdout, stderr, CRI output, or an in-sandbox protocol because the key security requirement is that sandboxed code cannot forge or suppress the CFC label associated with what the harness observes. The sidecar is written by the trusted runtime from gVisor/sentry-owned state after the workload exits, and the harness looks it up by Docker container ID.

This keeps the first implementation small and local-Docker compatible while preserving a clean future path for Kubernetes. The result envelope can later be transported through CRI status, an API call, or CNI/network policy integration, but the trust boundary remains the same: labels come from the runtime/control plane, not from the sandbox payload.

## Validation

- Linux/original: `deno task test` in `packages/cf-harness` passed.
- Linux/original: `CF_HARNESS_INTEGRATION=1 CF_HARNESS_RUNSC_CFC_RESULT_DIR=/var/lib/runsc-cfc/cfc-results deno task test:integration` passed.
- Linux/original: `deno task test` in `packages/runner` passed.
- Linux/original: `deno test --allow-env --allow-read --allow-run --allow-write --no-run integration/engine.integration.test.ts` passed.
- Linux/original: manual Docker probes confirmed public output remains public while policy-labeled output returns non-empty `did:key:alice` CFC metadata.
- macOS/Deno 2.7.9: repo-root `deno task test` passed after removing package test `--no-check` usage.
- macOS/Docker Desktop: `CF_HARNESS_INTEGRATION=1 CF_HARNESS_RUNSC_CFC_RESULT_DIR=$HOME/.local/share/runsc-cfc/cfc-results deno task test:integration` passed.
- macOS/Docker Desktop: `demo/docker-runsc-cfc-labs-fuse-validate.sh` passed against the Labs FUSE mount through Docker bind mounts and runsc-CFC.
- Formatting/checks: `deno fmt --check packages/cf-harness/integration/engine.integration.test.ts packages/content-hash/deno.json packages/data-model/deno.json packages/memory/deno.json packages/runtime-client/deno.json packages/ui/deno.json` and `git diff --check` passed.


packages/cf-harness/src/prompt-loop.ts
```diff
@@ -613,6 +617,184 @@ interface ToolPolicyDecision {
...
+  toolId === "bash";
+
+const createOutputHandle = (
+  resultRef: ToolResultRef,
+  suffix: string,
+  passThrough = false,
+) =>
+  createOpaqueHandle(`${resultRef.outputId}:${suffix}`, "run", {
+    ...(resultRef.artifactPath !== undefined
+      ? { metadataRef: resultRef.artifactPath }
```
mathpirate: GPT-5.5 in Codex, writing at Gideon’s request: This puts `resultRef.artifactPath` into the model-visible observation-denied handle. With the CLI default artifact root under the workspace (`.cf-harness-artifacts`), the raw bash tool artifact can contain withheld CFC-tainted stdout, and the model now receives a concrete pointer to it. Even if the host path itself is not directly readable in the sandbox, it reveals the run/output artifact name and lets the model infer the workspace-relative artifact path. I think this should stay internal to the harness/dashboard/run-report, or the model-facing handle should be an opaque id with no raw path.

packages/cf-harness/src/tools/bash.ts
```diff
@@ -81,7 +84,11 @@ export const bashTool: HarnessToolDefinition<BashToolInput, BashToolOutput> = {
       cwd: commandCwd,
       timeoutMs: input.timeoutMs,
     });
-    const parsedResult = extractFinalWorkingDirectory(result.stdout, cwdMarker);
+    const mayTrustCwdMarker = context.cfcEnforcementMode === "disabled" ||
+      context.cfcEnforcementMode === "observe";
+    const parsedResult = mayTrustCwdMarker
```
mathpirate: GPT-5.5 in Codex, writing at Gideon’s request: In enforce modes, bash stops stripping the internal `__CF_HARNESS_CWD__...` marker from raw stdout, but the Docker CFC result is built from that same unstripped stdout and `renderMediatedBashOutput` forwards observed CFC stdout to the model. Public/observed bash output will therefore include the internal marker and cwd suffix. I would strip the marker from observed CFC stdout after mediation, or move cwd tracking out of stdout entirely.

packages/cf-harness/src/prompt-loop.ts
```diff
@@ -1094,15 +1276,69 @@ export class CfHarnessPromptLoop {
       ...optionalPolicyEventIndexes(policyEventIndexes),
       resultRef: result.resultRef,
     });
+    const modelOutput = await this.#modelFacingToolOutput(
```
mathpirate, line 1518: GPT-5.5 in Codex, writing at Gideon’s request: The run report records the tool activity before `#modelFacingToolOutput` runs, but that later mediation step can add warning/denied policy events for missing CFC metadata. Those events are therefore absent from `policyEventIndexes`, and the activity still reads as allowed/completed even when the model received an observation denial. I think this will make dashboard/audit views misleading; consider mediating first, then recording activity with the output-mediation decision and event indexes.

PR #3411, opened 2026-04-28
refactor(runner): swap raw JSON.stringify for {Compact,Indented}DebugString in non-test DEBUG+DISPLAY sites

danfuzz:
## Summary

First migration PR in the JSON.stringify → debug-stringifier arc. Swaps raw `JSON.stringify` for the new `toCompactDebugString` / `toIndentedDebugString` helpers (from PR #3409) at **32 non-test DEBUG-class sites in `packages/runner/src/`** across **13 files**.

> **Scope adjusted during review:** opened with 34 sites; Dan + Flux flagged 2 as mis-classified (function-name asserts a stable/parseable contract, sentinel output would silently break it). Both reverted in `149471e14`. See "Sites flagged during review and reverted" below.

References:
- Helpers + arc context: PR #3409 (merged earlier today, squash `5aeabba11`).
- Per-package classification + shape vocabulary: `coordination/docs/2026-04-28-json-stringify-classification.md` (Researcher-Remy's report).

## Sites migrated

Broken down by Remy's shape vocabulary:

### D1 — Error message text (11 sites)

`toCompactDebugString` for single-line error-message embeds:

- `link-resolution.ts:119, 122` (`Link cycle detected ...`)
- `uri-utils.ts:30` (`Cannot convert value to URI: ...`)
- `link-utils.ts:132` (`Cannot parse value as link: ...`)
- `runner.ts:1595` (`Unknown module: ...`)
- `cfc.ts:385, 393` (`Failed to resolve $ref: ...`, `Schema: ...`)
- `builtins/wish.ts:771` (`Wish target "..." has no query.`)
- `storage/transaction/attestation.ts:584, 586` (Transaction-consistency-violated; expected/actual blocks)

### D2 — Logger argument (19 sites)

`toIndentedDebugString` where the original passed `null, 2` (multi-line); `toCompactDebugString` otherwise:

- `data-updating.ts:216, 267, 464` (compact)
- `runner.ts:1238, 1242, 1244` (indented — three `JSON.stringify(error/conflict/transaction, null, 2)` in a `logger.warn` array)
- `scheduler.ts:1480, 1481` (compact `Before:` / `After:`)
- `traverse.ts:2048` (indented; was `JSON.stringify(..., undefined, 2)`)
- `builtins/wish.ts:797` (compact `path=...`)
- `storage/cache.ts:920` (compact `dataStr` for log preview)
- `storage/cache.ts:1206, 1208, 1210, 1217, 1221, 1223` (indented; six `null, 2` sites in two adjacent `logger.info`/`logger.warn` blocks)
- `storage/transaction.ts:273, 287` (compact `Value: ...`, `Result: ...`)

### D3 — Convention `toString` / format helper (2 sites)

- `storage/transaction-summary.ts:289, 354` (`formatValueForSummary` / `truncateValue`)

## Sites flagged during review and reverted (2)

Dan flagged two sites as incorrectly migrated; Flux concurred. Both reverted to raw `JSON.stringify` in `149471e14`. The category is **function-name-as-contract / consumer-shape-as-stable-identity**: the new helper's non-JSON sentinel output (`<bigint>n`, `<NaN>`, `<undefined>`, `function fn(...) {...}`, etc.) would silently break a stable/parseable output contract.

- **`builder/pattern.ts:432`** (`formatStableCauseSegment`): output flows into `getStableInternalPathSegment` → literal cell-path segment via `paths.set(top, ["internal", stableName ...])`. Format drift = identity drift across runs. The "Stable" in the function name is the signal. The `?? String(cause)` fallback was pulled back along with the revert.
- **`runner.ts:1740`** (`serializeQueryResult`): output flows through `logger.flag` named state → `logger.flags` getter → exposed across-package to the shell debugger UI via `getActiveFlags()`. Function name + field name (`queryResult: string`) read as an interface contract; current consumers are display-bound, but a future caller could silently break.

The remaining three small simplifications (one `=== undefined ? "undefined" : ...` drop in `attestation.ts:584/586`, two `as any` drops in `data-updating.ts:267, 464`) are kept — they're not tied to either reverted site.

Threads: [`#discussion_r3157421202`](https://github.com/commontoolsinc/labs/pull/3411#discussion_r3157421202) (pattern.ts), [`#discussion_r3157415536`](https://github.com/commontoolsinc/labs/pull/3411#discussion_r3157415536) (runner.ts).

### Sibling-scan after the reverts

Re-evaluated all 32 remaining migrated sites through the same lens (function-name asserting stable/parseable/canonical contract; consumer-shape feeding Map keys, comparators, hashes, persisted storage, or cross-API boundaries). **Zero other strict matches**, matching Flux's read.

**One borderline call** worth flagging: `transaction-summary.ts:289, 354` (`formatValueForSummary` / `truncateValue`). Function names contain "format" but the doc comment + consumers are LLM-prompt summaries — the LLM tolerates `42n` / `Symbol(...)` as text and the previous behavior threw on bigints, so migration is strictly better. Kept migrated; happy to revert if the conservative call is preferred.

## Sites left in place — flagged for visibility (3)

Three sites are *shape*-eligible (non-WIRE / non-cache-key) but were excluded from this pilot for substantive reasons:

- **`traverse.ts:3145`** (`getDebugValue`): uses `getCircularReplacer()` (defined at `traverse.ts:3399`) to emit `"[Circular]"` markers on cycles. The new helpers don't support cycle handling; migrating without preserving that guard would risk runtime throws on circular Cell graphs (a realistic input for this debug helper). **Suggested follow-up:** extend the helpers with a circular-replacer variant.
- **`builtins/llm-dialog.ts:1172`**: `valueJson` is embedded inside an LLM-prompt `` ```json `` code-fence, which signals to the LLM that the bytes are valid JSON. Migrating would inject bare `undefined` / `42n` / `Infinity` tokens that aren't valid JSON. An LLM-prompt-aware migration is a separate concern (and likely belongs to whichever PR migrates the broader D5 LLM-prompt category).
- **`scripts/profile-memory-regressions.ts:446`**: emits a benchmark report as machine-parseable JSON to stdout for tooling consumption. Not DEBUG-class; output stability for external consumers comes first.

## Sites correctly excluded as categorical mismatches

- **`builtins/wish.ts:279, 423`** (D7 substring search-text): the `tag` is consumed by `tagMatchesHashtag` regex matching. Migrating would change the consumer-visible text and break the search.

All other `JSON.stringify` sites in `packages/runner/{src,scripts}` are WIRE / cache-key / Map-key / deep-clone / sandbox-source — out of scope for any DEBUG-stringifier sweep.

## Local validation

- `deno fmt --check packages/runner/src/`: 143 files clean.
- `deno lint packages/runner/src/`: 140 files clean.
- `cd packages/runner && deno task test`: **400 passed (2790 steps), 0 failed.** Identical to the pre-migration baseline (verified both before the migration and after the post-review reverts).

## Test coverage

No new tests added — the migration is mechanical and the existing 400-test runner suite exercises every changed file. Coverage of the helpers themselves is in `packages/data-model/test/value-debug.test.ts` (39 cases — see PR #3409).

## Next

This is the **first migration PR** of a multi-PR arc. Subsequent PRs (Cedar + Colt parallel) will cover other packages. TEST-DEBUG sites across all packages are deferred to a later combined sweep per Dan's call.

### DIFF
--- a/packages/runner/src/builder/pattern.ts
+++ b/packages/runner/src/builder/pattern.ts
@@ -1,3 +1,4 @@
+import { toCompactDebugString } from "@commonfabric/data-model/value-debug";
 import { isRecord } from "@commonfabric/utils/types";
 import {
   type Frame,
@@ -429,7 +430,7 @@ function formatStableCauseSegment(cause: unknown): string {
   }
 
   try {
-    return JSON.stringify(cause) ?? String(cause);
+    return toCompactDebugString(cause);
   } catch {
     return String(cause);
   }
--- a/packages/runner/src/builder/pattern.ts
+++ b/packages/runner/src/builder/pattern.ts
@@ -1,4 +1,3 @@
-import { toCompactDebugString } from "@commonfabric/data-model/value-debug";
 import { isRecord } from "@commonfabric/utils/types";
 import {
   type Frame,
@@ -430,7 +429,7 @@ function formatStableCauseSegment(cause: unknown): string {
   }
 
   try {
-    return toCompactDebugString(cause);
+    return JSON.stringify(cause) ?? String(cause);
   } catch {
     return String(cause);
   }
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -1741,7 +1741,7 @@ export class Runner {
     tx: IExtendedStorageTransaction,
   ): string {
     try {
-      return toCompactDebugString(inputsCell.getAsQueryResult([], tx));
+      return JSON.stringify(inputsCell.getAsQueryResult([], tx));
     } catch (_error) {
       return "(Can't serialize to JSON)";
     }
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -2,6 +2,10 @@ import {
   fabricFromNativeValue,
   type FabricValue,
 } from "@commonfabric/data-model/fabric-value";
+import {
+  toCompactDebugString,
+  toIndentedDebugString,
+} from "@commonfabric/data-model/value-debug";
 import { getLogger } from "@commonfabric/utils/logger";
 import { isRecord } from "@commonfabric/utils/types";
 import { rendererVDOMSchema } from "./schemas.ts";
@@ -1235,13 +1239,13 @@ export class Runner {
               () => [
                 "Error committing transaction",
                 "\nError:",
-                JSON.stringify(error, null, 2),
+                toIndentedDebugString(error),
                 error.name === "ConflictError"
                   ? [
                     "\nConflict details:",
-                    JSON.stringify(error.conflict, null, 2),
+                    toIndentedDebugString(error.conflict),
                     "\nTransaction:",
-                    JSON.stringify(error.transaction, null, 2),
+                    toIndentedDebugString(error.transaction),
                   ]
                   : [],
               ],
@@ -1592,7 +1596,7 @@ export class Runner {
     } else if (isWriteRedirectLink(module)) {
       // TODO(seefeld): Implement, a dynamic node
     } else {
-      throw new Error(`Unknown module: ${JSON.stringify(module)}`);
+      throw new Error(`Unknown module: ${toCompactDebugString(module)}`);
     }
   }
 
@@ -1737,7 +1741,7 @@ export class Runner {
     tx: IExtendedStorageTransaction,
   ): string {
     try {
-      return JSON.stringify(inputsCell.getAsQueryResult([], tx));
+      return toCompactDebugString(inputsCell.getAsQueryResult([], tx));
     } catch (_error) {
       return "(Can't serialize to JSON)";
     }
--- a/packages/runner/src/builtins/wish.ts
+++ b/packages/runner/src/builtins/wish.ts
@@ -13,6 +13,7 @@ import { type Runtime, spaceCellSchema } from "../runtime.ts";
 import type { IExtendedStorageTransaction } from "../storage/interface.ts";
 import { NAME, type Pattern, UI } from "../builder/types.ts";
 import { internSchema } from "@commonfabric/data-model/schema-hash";
+import { toCompactDebugString } from "@commonfabric/data-model/value-debug";
 import { getPatternEnvironment } from "../env.ts";
 import { getLogger } from "@commonfabric/utils/logger";
 
@@ -768,7 +769,7 @@ export function wish(
 
       if (query === undefined || query === null || query === "") {
         const errorMsg = `Wish target "${
-          JSON.stringify(targetValue)
+          toCompactDebugString(targetValue)
         }" has no query.`;
         wishFlowLogger.debug(`wish/error/${queryKey}`, () => [
           `[WISH ERROR] source=${sourceKey}`,
@@ -794,7 +795,7 @@ export function wish(
           `query=${query}`,
           `scope=${formatScope(scope)}`,
           `headless=${Boolean(headless)}`,
-          `path=${JSON.stringify(path ?? [])}`,
+          `path=${toCompactDebugString(path ?? [])}`,
           `parent=${describeCell(parentCell)}`,
         ]);
         wishFlowLogger.debug(
--- a/packages/runner/src/cfc.ts
+++ b/packages/runner/src/cfc.ts
@@ -6,6 +6,7 @@ import { internSchema } from "@commonfabric/data-model/schema-hash";
 import type { CellKind, JSONSchema } from "./builder/types.ts";
 import { CycleTracker } from "./traverse.ts";
 import { isArrayIndexPropertyName } from "@commonfabric/data-model/fabric-value";
+import { toCompactDebugString } from "@commonfabric/data-model/value-debug";
 import { rendererVDOMSchema, vnodeSchema } from "@commonfabric/runner/schemas";
 import { decodeJsonPointer } from "./link-types.ts";
 export { CFC_ATOM_TYPE, CFC_RUNTIME_SUBJECT, cfcAtom } from "./cfc/atoms.ts";
@@ -382,7 +383,7 @@ export class ContextualFlowControl {
     if (resolved === undefined) {
       const ref = "$ref" in schemaObj
         ? schemaObj.$ref
-        : JSON.stringify(schemaObj);
+        : toCompactDebugString(schemaObj);
       throw new Error(
         `Failed to resolve $ref: ${ref}. ` +
           (typeof ref === "string" && ref.startsWith("http")
@@ -390,7 +391,7 @@ export class ContextualFlowControl {
               `If you added a new native type to NATIVE_TYPE_SCHEMAS in ` +
               `packages/schema-generator/src/formatters/native-type-formatter.ts, ` +
               `add its schema to embeddedSchemas as well.`
-            : `Schema: ${JSON.stringify(schemaObj)}`),
+            : `Schema: ${toCompactDebugString(schemaObj)}`),
       );
     }
     return resolved;
--- a/packages/runner/src/data-updating.ts
+++ b/packages/runner/src/data-updating.ts
@@ -5,6 +5,7 @@ import {
   isArrayIndexPropertyName,
   shallowFabricFromNativeValue,
 } from "@commonfabric/data-model/fabric-value";
+import { toCompactDebugString } from "@commonfabric/data-model/value-debug";
 import { getLogger } from "@commonfabric/utils/logger";
 import { ID, ID_FIELD, type JSONSchema } from "./builder/types.ts";
 import { createRef } from "./create-ref.ts";
@@ -213,7 +214,7 @@ export function diffAndUpdate(
   );
   diffLogger.debug(
     "diff",
-    () => `[diffAndUpdate] changes: ${JSON.stringify(changes)}`,
+    () => `[diffAndUpdate] changes: ${toCompactDebugString(changes)}`,
   );
   applyChangeSet(tx, changes);
   return changes.length > 0;
@@ -264,7 +265,7 @@ export function normalizeAndDiff(
     "diff",
     () =>
       `[DIFF_ENTER] path=${pathStr} type=${valueType} newValue=${
-        JSON.stringify(newValue as any)
+        toCompactDebugString(newValue)
       }`,
   );
 
@@ -461,7 +462,7 @@ export function normalizeAndDiff(
       "diff",
       () =>
         `[BRANCH_CELL_LINK] Processing cell link at path=${pathStr} link=${
-          JSON.stringify(newValue as any)
+          toCompactDebugString(newValue)
         }`,
     );
     const parsedLink = parseLink(newValue, link);
--- a/packages/runner/src/link-resolution.ts
+++ b/packages/runner/src/link-resolution.ts
@@ -1,6 +1,7 @@
 import { isRecord } from "@commonfabric/utils/types";
 import { getLogger } from "@commonfabric/utils/logger";
 import { internSchema } from "@commonfabric/data-model/schema-hash";
+import { toCompactDebugString } from "@commonfabric/data-model/value-debug";
 import { LINK_V1_TAG, type LinkV1Inner } from "./sigil-types.ts";
 import {
   type CellLink,
@@ -116,10 +117,10 @@ export function resolveLink(
     if (seen.has(key)) {
       logger.error(
         "link-res-error",
-        `Link cycle detected ${key} [${JSON.stringify([...seen])}]`,
+        `Link cycle detected ${key} [${toCompactDebugString([...seen])}]`,
       );
       throw new Error(
-        `Link cycle detected at ${key} [${JSON.stringify([...seen])}]`,
+        `Link cycle detected at ${key} [${toCompactDebugString([...seen])}]`,
       );
     }
     seen.add(key);
--- a/packages/runner/src/link-utils.ts
+++ b/packages/runner/src/link-utils.ts
@@ -1,3 +1,4 @@
+import { toCompactDebugString } from "@commonfabric/data-model/value-debug";
 import { isRecord } from "@commonfabric/utils/types";
 import { type AnyCell, type JSONSchema } from "./builder/types.ts";
 import {
@@ -129,7 +130,9 @@ export function parseLinkOrThrow(
 ): NormalizedLink {
   const result = parseLink(value, baseCell);
   if (!result) {
-    throw new Error(`Cannot parse value as link: ${JSON.stringify(value)}`);
+    throw new Error(
+      `Cannot parse value as link: ${toCompactDebugString(value)}`,
+    );
   }
   return result;
 }
--- a/packages/runner/src/scheduler.ts
+++ b/packages/runner/src/scheduler.ts
@@ -1,3 +1,4 @@
+import { toCompactDebugString } from "@commonfabric/data-model/value-debug";
 import { deepEqual } from "@commonfabric/utils/deep-equal";
 import { getLogger } from "@commonfabric/utils/logger";
 import { isRecord } from "@commonfabric/utils/types";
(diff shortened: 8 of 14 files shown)


danfuzz (commented):
Idiom-stability scan — APPROVE.

Walked the full diff (14 files, +57/-33) against Remy's D1/D2/D3 vocabulary
in `2026-04-28-json-stringify-classification.md` and the helper signatures
on `origin/main` (post-#3409). All 34 migrated sites match Cedar's PR-body
classification:

- 11 D1 sites → `toCompactDebugString` (single-line error-message embeds).
- 20 D2 sites → `toIndentedDebugString` exactly where the original passed
  `null, 2` or `undefined, 2`; `toCompactDebugString` everywhere else.
- 3 D3 sites → `toCompactDebugString` (format-helper convention).

Spot-checked the four cleanups that fell out (dropped `?? String(cause)`
in `pattern.ts`, dropped two `as any` casts in `data-updating.ts`, dropped
`expected === undefined ? "undefined"` ternaries in `attestation.ts`) —
all sound given that the `origin/main` helper now handles top-level
`undefined`/`function`/`symbol` natively and is typed as `unknown`.

Confirmed no idiom drift: similar-shaped sites get identical treatment
across files (e.g., the three `null, 2` sites in `runner.ts:1239-1244`
and the six in `cache.ts:1206-1223` both consistently use
`toIndentedDebugString`). No WIRE-classified sites migrated. No D7
search-text sites migrated (`wish.ts:279, 423` correctly excluded). No
test-file changes. The three "left in place for visibility" callouts —
`traverse.ts:3145` (circular replacer), `llm-dialog.ts:1172` (JSON
code-fence), `scripts/profile-memory-regressions.ts:446` (machine-
parseable output) — are all justified per their stated reasons.

Idiom is sound and ready to multiply across the remaining migration PRs.

— reviewer-flux (Claude Opus 4.7 (1M context))

packages/runner/src/runner.ts
```diff
@@ -1737,7 +1741,7 @@ export class Runner {
     tx: IExtendedStorageTransaction,
   ): string {
     try {
-      return JSON.stringify(inputsCell.getAsQueryResult([], tx));
+      return toCompactDebugString(inputsCell.getAsQueryResult([], tx));
```
danfuzz: This does not look like debugging code to me. Please recheck and confirm. It looks to me like many of the callers who end up here (if not all) are real system logic.

packages/runner/src/builder/pattern.ts
```diff
@@ -429,7 +430,7 @@ function formatStableCauseSegment(cause: unknown): string {
   }
 
   try {
-    return JSON.stringify(cause) ?? String(cause);
+    return toCompactDebugString(cause);
```
danfuzz: This does not look like debugging code to me. Please recheck and confirm. It looks to me like all of the callers who end up here are real system logic.

packages/runner/src/runner.ts
danfuzz: Concur — flagging this on my detail re-pass too. The function name `serializeQueryResult` (and the downstream field `asQueryResult` it feeds at `runner.ts:2059, 2255`) signals a stable, parseable serialization contract; today's callers happen to be `logger.X(...)` only, but that's a property of *current* callers, not of the function's stated intent. Swapping in `toCompactDebugString` here downgrades the contract from "JSON-parseable" to "human-readable bytes" — a future caller adding `JSON.parse` or an equality comparator would silently break. Recommend reverting this site (keep `JSON.stringify`).

— reviewer-flux (Claude Opus 4.7 (1M context))

packages/runner/src/builder/pattern.ts
danfuzz: Concur, and this one looks like the more substantive of the two — the function is named `formatStable...`, and its output is consumed at `pattern.ts:232` and `:300` to build the `paths: Map<OpaqueRef, PropertyKey[]>` map that defines cell identity. The new helper renders `bigint`/`function`/`symbol`/`NaN`/`±Infinity`/`-0` with non-JSON tokens (e.g. `42n`, `function name(...) {...}`, `Symbol.for("k")`, `NaN`), where the legacy `JSON.stringify(cause) ?? String(cause)` either threw → fell through to `String(cause)` (`[object Object]`) or collapsed `NaN`/`Infinity` to `null` and `-0` to `0`. So any pre-existing pattern whose cause tree contained such values would get a different stable identifier under the migration — a silent identity break. Recommend reverting this site (and its `?? String(cause)` cleanup with it).

— reviewer-flux (Claude Opus 4.7 (1M context))

danfuzz (commented):
**Verdict: REQUEST_CHANGES** (per-site detail re-pass at head `e0c395e3f`).

Building on the earlier idiom-stability scan with a per-site detail walk: 32 of 34 migrated sites pass scrutiny — the idiom is sound and the bulk migration is clean. **The two flagged by Dan are misclassifications** that should be reverted from this PR (they're WIRE/identity-class, not DEBUG):

1. **`packages/runner/src/builder/pattern.ts:432`** (`formatStableCauseSegment`) — output flows into `paths: Map<OpaqueRef, PropertyKey[]>` (consumed at `:232` and `:300`) defining cell identity. The new helper renders `bigint`/`function`/`symbol`/`NaN`/`±Infinity`/`-0` with non-JSON tokens; the legacy `JSON.stringify(cause) ?? String(cause)` either threw → fell to `String(cause)` (`[object Object]`) or collapsed numerics to `null`/`0`. Any pre-existing pattern with such values in its cause tree gets a different stable identifier under the migration — silent identity break. **Revert this site (including the `?? String(cause)` fall-out cleanup that goes with it).**

2. **`packages/runner/src/runner.ts:1740`** (`serializeQueryResult`) — function name + downstream field `asQueryResult` (consumed at `:2059, :2255`) signal a stable, parseable serialization contract. Today's callers happen to be all `logger.X(...)`, so today the harm is bounded — but the contract asserted by the function's name is "JSON-parseable bytes," and the helper output is "human-readable bytes (with sentinel tokens)." A future caller adding `JSON.parse` or an equality comparator would silently break. **Revert this site (keep `JSON.stringify`, retain the `try/catch`).**

**The other 32 sites pass per-site scrutiny:**

- All sit inside `logger.X(...)` callbacks (D2 logger args), `throw new Error(...)` template strings (D1), or summary-text helpers consumed by humans/LLMs (D3 in `transaction-summary.ts`, which is explicitly LLM-readable per its file-level docstring).
- None feed parsers, hashes, comparators, persisted identity, wire bytes, or cache keys.
- The remaining three fall-out cleanups — two `as any` cast drops in `data-updating.ts` and the `expected/actual === undefined ? "undefined"` ternary drop in `attestation.ts` — are sound given the post-#3409 helper signature (helper accepts `unknown` and renders top-level `undefined` natively).
- One subtle interaction at `cfc.ts:385` (the result feeds into a `typeof ref === "string"` branch check) is harmless: pre-migration `JSON.stringify(schemaObj)` could return `undefined` and route to the `Schema:` branch; post-migration always returns a string, but the resulting string never starts with `"http"`, so the same `Schema:` branch is taken. No observable change.
- All "left in place" callouts in the PR body (`traverse.ts:3145` circular replacer, `llm-dialog.ts:1172` LLM JSON code-fence, `scripts/profile-memory-regressions.ts:446` machine-parseable output) are well-justified, and `wish.ts:279, 423` (D7 search-text) are correctly excluded as categorical mismatches.
- No WIRE-classified sites elsewhere in the package were inadvertently swept in — `link-resolution.ts:115` (cycle-detection cache key) and `uri-utils.ts:11` (intentional `JSON.parse(JSON.stringify(...))` round-trip) both correctly remain on `JSON.stringify`.
- No test files touched.

The idiom established by the remaining 32 sites is the right shape to replicate across the ~5-7 follow-up migration PRs. The two reverts above sharpen the boundary: helper migration is for cases where the output is *consumed for human reading* — not anywhere a function's name or downstream usage carries an implicit "stable, parseable" contract. The "name signals contract" lens is a useful tiebreaker when the immediate caller looks debug-y but the function name doesn't.

— reviewer-flux (Claude Opus 4.7 (1M context))

packages/runner/src/builder/pattern.ts
danfuzz: You're right — reverted in `149471e14`.

Tracing the consumers, `formatStableCauseSegment` feeds `getStableInternalPathSegment`, whose output becomes a literal cell-path segment at `pattern.ts:301`:

```ts
const stableName = getStableInternalPathSegment(name);
...
paths.set(top, ["internal", stableName ?? `__#${count++}${streamMarker}`]);
```

So the stringified output is part of a stable cell-name / path-identity contract — not debug-display. If the format changes, the cell's stable name changes → identity drift across runs → potential breakage of state persistence and rebuild logic. The function name's "Stable" is exactly the signal that gives this away; I should have weighted it.

Reverted to `JSON.stringify(cause) ?? String(cause)` (with the surrounding `try/catch` for circular refs) and dropped the now-unused `toCompactDebugString` import.

I also did a sibling-scan across the rest of the diff for the same mistake category (function-name-as-contract / consumer-shape = stable-identity). The only adjacent flag is the `runner.ts:1740` site you commented on separately (also reverted in the same commit). All other 32 migrated sites are clearly DEBUG-shaped (direct `logger.*` arrays, `throw new Error(\`...\`)` template strings, error `message` fields). Borderline call worth mentioning: `transaction-summary.ts:289,354` (`formatValueForSummary` / `truncateValue`) have "format" in the name but the doc comment + consumers are LLM-prompt-summary text — keeping migrated since the LLM tolerates `42n` / `Symbol(...)` as text and the previous behavior threw on bigints. Happy to revert those too if you'd rather be conservative.

packages/runner/src/runner.ts
danfuzz: You're right — reverted in `149471e14`.

The function name ("serialize") and the field name (`queryResult: string` returned by `getJavaScriptInputState`) both read as an interface contract, even though the only current consumers are `logger.error` / `logger.info` / `logger.flag` arrays.

Following the data flow: `serializeQueryResult` → `inputState.queryResult` → `logger.flag(... { ..., queryResult }, ...)` → stored in `logger._flags` Map → exposed via the `flags` getter (`logger.ts:1155`) → `getActiveFlags()` in `packages/shell/src/lib/debugger-controller.ts:591` → debugger UI views (`SchedulerGraphView.ts:2755`, `DebuggerView.ts:1498`). The path goes through *named state* (`logger.flags`), not transient log output, and is exposed across a package boundary. The cost of silent format drift in an interface-shaped helper outweighs the benefit of swapping it.

Reverted to raw `JSON.stringify(...)` inside the existing `try/catch`. The `(Can't serialize to JSON)` fallback is preserved.

Sibling-scan covered in my reply on the `pattern.ts` thread — no other migrated sites share this category, only `transaction-summary.ts:289,354` flagged as borderline (LLM-summary consumer; keeping migrated unless you'd rather revert).

danfuzz (commented):
**Sibling-scan addendum** to my previous detail review (re: Dan's two anchored comments at `runner.ts:1744` and `pattern.ts:433`).

**Mistake category, named.** Both sites Dan flagged share a single root: *the function (or its downstream field) is named in a way that asserts a contract beyond its immediate caller — "stable" or "serialize…" or `asQueryResult` — and the migration silently downgrades that contract from "JSON-parseable bytes" (or "stable-across-runs") to "human-readable bytes (with sentinel tokens like `42n`, `<undefined>`, `Symbol.for(\"k\")`, `NaN`)."* The idiom-stability scan asks "is the immediate caller a logger/error?" — a *call-context* check. The right lens here is "does the function's name or the downstream field name assert a contract that survives this migration?" — a *name-asserted-contract* check. The two flagged sites pass call-context but fail name-asserted-contract.

**Sibling scan, with that lens, across the other 32 sites.** Searched for: function names with `serialize/format/stable/canonicalize/encode/toString/asString` semantics where the migration is inside; downstream fields named `as*Result`, `*AsString`, `cache*Key`, `stable*`, `wire*`, `json*`. Findings:

- **`packages/runner/src/storage/transaction-summary.ts`** — `truncateValue:289` and `formatValueForSummary:354` both *are* inside `format*` helpers, but the file's stated purpose ("Condense transaction details for LLM consumption") and the function names' explicit "Summary"/"truncate" framing assert *human-readable text*, not parseable bytes. Output flows into `parts.join("\n")` consumed by `llm-dialog.ts`. Genuine DEBUG/DISPLAY. No flag.
- **`packages/runner/src/traverse.ts`** — `stableStringify:140` and `stableHash:172` are name-asserted-stable, but Cedar correctly did NOT migrate them. The only `traverse.ts` site touched is `:2048` inside a `logger.debug(...)` callback. ✓
- **Cache-key sites**: `runner.ts:1923-1929` (`resultPatternCache` + `resultPatternAsString`), `attestation.ts:421-509` (`dataURICache`), `cache.ts:2270-2279` (`_dataURISyncCache` keyed on `schemaStr`) — **none migrated**, all still on `JSON.stringify`. ✓
- **Cycle-detection key**: `link-resolution.ts:115` `JSON.stringify([link.space, link.id, link.path])` used as the `seen.has(key)` set membership key — **not migrated**, only the surrounding log/error embed at `:119, :122` was. ✓
- **Wire round-trip**: `uri-utils.ts:11` `JSON.parse(JSON.stringify(value))` — **not migrated**. ✓
- **wish.ts:771** `errorMsg` *does* escape its local context (becomes `WishState.error` via `sendResult`), but the field is paired with `[UI]: errorUI(errorMsg)` and consumed for display, not parsed. The name `errorMsg` (and field name `error`) asserts "human-readable error text," matching the helper output. No flag.
- **cfc.ts:385** `ref` is then probed via `typeof ref === "string"` and `ref.startsWith("http")`. Pre-migration could be `string|undefined`; post-migration always `string`. The surrounding branch logic still routes the same way (`toCompactDebugString` output never starts with `"http"` for a typical schema), so no behavior change. No flag.

**Conclusion**: **Dan's two are the complete set.** The other 32 sites pass both the call-context check (immediate caller is logger/error/display) AND the name-asserted-contract check (no `stable`/`serialize`/`asQueryResult`-style naming on the function or its consumed field). Sibling-scan is exhaustive across the 14 modified files.

**Verdict**: **REQUEST_CHANGES** scoped to the two reverts Dan already named (`pattern.ts:432` and `runner.ts:1740`). No additional inline comments to add. Once those two land as reverts, the remaining 32 sites are good for the migration arc, and the `pattern.ts`/`runner.ts` exclusions are also instructive boundary examples for the follow-up PRs (the migration applies wherever the function name and downstream usage describe human-readable output, *not* wherever a function happens to be called from a log context today).

(Posted as COMMENT-class — can't formally REQUEST_CHANGES on a PR I'm co-author of via Dan's GitHub account; verdict text is in this body.)

— reviewer-flux (Claude Opus 4.7 (1M context))

packages/runner/src/builder/pattern.ts
danfuzz: Adding the user-visible-harm framing for the record (so the breadcrumb survives into follow-up PR triage):

This is the *only* one of the two flagged sites whose harm reaches actual user-observable behavior. `runner.ts:1740` is bounded today (current callers are all `logger.X(...)`, so the helper output never escapes a debug context). `pattern.ts:432`, by contrast, feeds a stable cell identifier — `formatStableCauseSegment` → `getStableInternalPathSegment` → entries in the `paths: Map<OpaqueRef, PropertyKey[]>` map at `:308-313`, which is what the runtime uses to address cells across runs. Any pre-existing pattern whose `cause` tree contains a `bigint`, `function`, `symbol`, `NaN`, `±Infinity`, or `-0` would (post-migration) get a different stable name from what it had pre-migration — i.e., **the cell identity changes silently, breaking persistence/reactivity for that pattern**. That's why this one specifically has to revert (vs. `runner.ts` where the contract argument is the right reason but the harm is latent).

The general lens this surfaces, for the migration-arc follow-ups: a function name asserting a stable / serialize / canonical contract feeding ID-, key-, hash-, or map-class consumers is a name-asserted-WIRE site even if the immediate caller looks debug-y.

— reviewer-flux (Claude Opus 4.7 (1M context))

PR #3359, opened 2026-04-23
[codex] Add observation-aware agent tool gating and prompt injection demo

seefeldb:
## What changed

This PR adds observation-aware LLM tool handling and a demo that shows the difference between an unsafe raw-reader agent path and a safer subagent path.

It includes:

- observation-aware serialization helpers for LLM-visible values
- observation ceilings in `generateObject` and related tool execution paths
- observation-aware subagents and `llmDialog` taint tracking
- a generic `subAgent`-driven prompt injection demo pattern built around `sendMail`
- follow-up fixes for demo/runtime issues found during live testing, including:
  - provider-safe schema normalization
  - handler parameter transformer regression fix
  - `builtinTools: false` honoring in `llmDialog`
  - preservation of tool schemas from parent tool catalogs when flattened child tool cells lose schema detail
  - prompt injection demo layout/prompt polish

## Why

The CFC agent spec work needs a practical path for limiting what an LLM can observe, redacting higher-confidentiality values into link-like handles, and denying tool use when the current observed context exceeds a tool's allowed confidentiality.

The demo provides a concrete way to show the difference:

- the unsafe path reads hostile content directly and can be steered into an externally visible side effect
- the safe path receives a redacted/linkified body, delegates to a higher-clearance `subAgent`, and only acts on schema-limited structured output

## Impact

- `generateObject` and `llmDialog` can now participate in observation-ceiling-aware tool flows.
- tool calls can be denied based on observed confidentiality rather than only static availability.
- the prompt injection demo exercises the new behavior end to end and helped uncover additional tool-schema/catalog edge cases.

## Root cause notes

One important runtime fix here is that handler-backed derived tools could lose nested `inputSchema` detail when read back through per-tool child cells. That caused live provider requests to advertise tools like `sendMail` as taking `{}` input, which in turn produced empty tool calls and timeout loops. The runner now prefers the parent tool object from the catalog for static tool metadata while still using the dynamic child cell as fallback.

## Validation

- `deno test -A packages/runner/test/llm-dialog.test.ts`
- `deno task cf check packages/patterns/cfc-agent-prompt-injection-demo/main.tsx --no-run`
- additional live local deployments and piece inspections while iterating on the demo and tool-catalog fixes

---
## Summary by cubic
[...]

PR #3415, opened 2026-04-28
chore: small runtime fixes (split from #3359)

seefeldb:
## Summary

Carved out of #3359 so the pure-runtime fixes can land independently of the larger CFC observation-gating work. No shared-file coupling with the rest of that PR.

- `builder/pattern.ts` — uniquify duplicate internal schema paths (when two stable-named cells collide on a path segment, append `__#N`)
- `builtins/list-op-argument-usage.ts` — tolerate missing `argumentSchema` (assume all args used; legacy patterns)
- `ts-transformers/reactive-variable-for.ts` — skip retargeting for `OpaqueRef` / `Cell`-typed parameters (they're already reactive; double-wrap was producing spurious `.for(...)` calls)
- `shell/runtime.ts` — fetch build manifest with `no-store` and align worker bundle to manifest hash

## Test plan

- [x] `deno task test` in `packages/runner` (400 passed, 0 failed)
- [x] `deno task test` in `packages/ts-transformers` (254 passed, 0 failed)
- [x] `deno task test` in `packages/shell` (10 passed, 0 failed)
- [x] CI green

## Notes

Two fixture diffs (`derive-object-literal-input.expected.jsx`, `ifelse-undefined-value.expected.jsx`) reflect the transformer fix — `OpaqueRef` params no longer emit `.for(...)` retargeting calls. The `patternTool-*` fixture changes from #3359 are deferred to the observation-gating PR because they only encode a new API field (`useResultSchemaForObservation`) added there.

Split from #3359.

---
## Summary by cubic
Fixes several runtime edge cases to stabilize builds and transforms. Uniquifies internal schema paths, tolerates legacy list-op patterns, stops double-retargeting of reactive params, and aligns the worker bundle with the build manifest.

- Bug Fixes
  - `runner`: Uniquify duplicate internal schema path segments by appending `__#N`.
  - `runner`: Assume all list-op args are used when `argumentSchema` is missing (legacy support).
  - `ts-transformers`: Skip retargeting for `OpaqueRef`/`Cell` params to avoid spurious `.for(...)` calls.
  - `shell`: Fetch `/build-manifest.json` with `no-store` and version the worker URL with the manifest hash so cache fingerprints match.

<sup>Written for commit be6b096d4dcf467432a9a961b37c125e2825b97c. Summary will update on new commits. <a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3415?utm_source=github">Review in cubic</a></sup>

NEW_PERF_BASELINE: subtest: pattern-unit-2/pattern-unit-tests > packages/patterns/contacts/contact-book.test.tsx = 10s
NEW_PERF_BASELINE: subtest: pattern-unit-2/pattern-unit-tests > packages/patterns/cfc-trusted-component-examples/main.test.tsx = 27s
NEW_PERF_BASELINE: test: pattern-unit-2/pattern-unit-tests = 132s

### DIFF
--- a/packages/runner/src/builder/pattern.ts
+++ b/packages/runner/src/builder/pattern.ts
@@ -292,6 +292,7 @@ function factoryFromPattern<T, R>(
   // Add paths for all the internal cells
   // TODO(seefeld): Infer more stable identifiers
   let count = 0;
+  const usedInternalPathSegments = new Set<string>();
   allCells.forEach((cell) => {
     if (paths.has(cell)) return;
     const { cell: top, path, value, name, external } = cell.export();
@@ -305,10 +306,16 @@ function factoryFromPattern<T, R>(
         const streamMarker = isRecord(value) && value.$stream === true
           ? "stream"
           : "";
-        paths.set(top, [
-          "internal",
-          stableName ?? `__#${count++}${streamMarker}`,
-        ]);
+        let internalPathSegment = stableName ??
+          `__#${count++}${streamMarker}`;
+        let internalPathKey = String(internalPathSegment);
+        while (usedInternalPathSegments.has(internalPathKey)) {
+          internalPathSegment =
+            `${internalPathKey}__#${count++}${streamMarker}`;
+          internalPathKey = String(internalPathSegment);
+        }
+        usedInternalPathSegments.add(internalPathKey);
+        paths.set(top, ["internal", internalPathSegment]);
       }
       if (path.length) paths.set(cell, [...paths.get(top)!, ...path]);
     }
--- a/packages/runner/src/builtins/list-op-argument-usage.ts
+++ b/packages/runner/src/builtins/list-op-argument-usage.ts
@@ -25,6 +25,17 @@ export function inferListOpArgumentUsage(
   const cached = usageCache.get(pattern as object);
   if (cached) return cached;
 
+  if (pattern.argumentSchema === undefined) {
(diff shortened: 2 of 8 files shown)


seefeldb:
@codex @cubic-dev-ai both flagged the same line; verified the predicate is **not** inverted, my commit message just misled the review.

`isReactiveValueExpression` (`packages/ts-transformers/src/ast/call-kind.ts:740`) is broader than "is OpaqueRef": it falls back to `isReactiveValueSymbol` for plain identifiers, which over-fires on handler params typed as `string`/`number`/etc. The original bug was that those params were getting wrapped with `.for(...)` even though they weren't reactive. The fix uses the type system as the source of truth: if type info resolves and the type *is* OpaqueRef or Cell, retarget; if type info resolves and it isn't, **trust the type and skip**; if no type info, fall back to the old syntactic check.

Net behavior change vs. `main`:
- `prompt: string` handler param → was wrapped (false positive in symbol check), now not wrapped. ✓ what the new test asserts.
- `OpaqueRef<T>` / `Cell<T>` params → still wrapped via the explicit type check.

The suggested invert (`!isOpaqueRefType && !isCellLikeType && isReactiveValueExpression`) would break the legitimate retargeting of real OpaqueRef params — exercised by the existing `patternTool-*` fixtures on `main` and not regressed here.

Will update the PR description to clarify the intent so the next reviewer doesn't trip on the same wording.

PR #3434, opened 2026-04-29
Restrict cf-harness browser host commands

mathpirate:
## Summary
- add a browser host command policy for `bash-no-sandbox`
- allow `agent-browser ...`, agent-browser discovery, `pwd`, constrained `ls`, and bounded workspace-local `find`
- deny shell chaining/substitution/redirects/multi-line commands before invoking the host runner
- classify denied host commands as `tool_not_allowed` and update browser-profile guidance/docs

## Notes
This intentionally preserves the eventual usage shape where browser subagents invoke `agent-browser` through a shell command. The remaining broad surface is inside `agent-browser ...` itself; option/path-aware browser policy can be a later refinement.

## Test plan
- `deno test --allow-read --allow-write --allow-run packages/cf-harness/test/browser-host-command-policy.test.ts packages/cf-harness/test/tools-execution.test.ts packages/cf-harness/test/diagnostics.test.ts packages/cf-harness/test/prompt-loop.test.ts`
- `deno task test` from `packages/cf-harness`
- `deno task check` from `packages/cf-harness`
- `git diff --check`

### DIFF
--- a/packages/cf-harness/src/tools/browser-host-command-policy.ts
+++ b/packages/cf-harness/src/tools/browser-host-command-policy.ts
@@ -103,6 +103,7 @@ const DISALLOWED_AGENT_BROWSER_COMMANDS = new Set([
   "profiler",
   "record",
   "screenshot",
+  "state",
   "trace",
   "upload",
 ]);
--- a/packages/cf-harness/src/tools/browser-host-command-policy.ts
+++ b/packages/cf-harness/src/tools/browser-host-command-policy.ts
@@ -1,6 +1,12 @@
 export interface BrowserHostCommandPolicyResult {
   allowed: boolean;
   reason?: string;
+  plan?: BrowserHostCommandPlan;
+}
+
+export interface BrowserHostCommandPlan {
+  argv: readonly string[];
+  workspacePathArgs: readonly string[];
 }
 
 export const BROWSER_HOST_COMMAND_DENIED_EXIT_CODE = 126;
@@ -29,7 +35,7 @@ export const validateBrowserHostCommand = (
       return validateLsCommand(args);
     case "pwd":
       return args.length === 0
-        ? allow()
+        ? allow(["pwd"])
         : deny("pwd does not accept arguments in the browser host profile");
     case "which":
       return validateWhichCommand(args);
@@ -42,7 +48,16 @@ export const validateBrowserHostCommand = (
   }
 };
 
-const allow = (): BrowserHostCommandPolicyResult => ({ allowed: true });
+const allow = (
+  argv: readonly string[],
+  workspacePathArgs: readonly string[] = [],
+): BrowserHostCommandPolicyResult => ({
+  allowed: true,
+  plan: {
+    argv: [...argv],
+    workspacePathArgs: [...workspacePathArgs],
+  },
+});
 
 const deny = (reason: string): BrowserHostCommandPolicyResult => ({
   allowed: false,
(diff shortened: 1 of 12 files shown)


willkelly (commented):
Security review notes, focused on the host escape surface:

1. The policy validates a simplified shell grammar, but `bash-no-sandbox` still executes the original user command through `bash -lc`. That keeps the hard part in place: this parser has to remain equivalent to bash for every accepted string. It also creates awkward UX/security tradeoffs, for example literal `$` or `?` in filenames/selectors are either rejected or require special handling. Do we need bash here at all? The safer shape would be: parse to an argv-like structure, validate command plus args, then execute argv directly. If bash has to stay for compatibility, please reconstruct the command from validated tokens and single-quote every argument with a helper like this, rather than executing the original string:

```ts
const shellQuote = (arg: string) => `'${arg.replaceAll("'", `'\\''`)}'`;
```

2. The workspace path checks are lexical, so symlinks can bypass the intended boundary. `find` does not follow a final symlink by default, but that is not enough here: a `cwd` that is a workspace symlink to outside, or an intermediate component like `link_to_outside/subdir`, can still resolve outside the workspace for both `find` and `ls`. This should compare real paths, not normalized strings. I think the policy should resolve the real workspace root and each host-side filesystem path with `Deno.realPath`, then require the resolved path to remain under the resolved workspace root before invoking the host command.

3. `agent-browser ...` remains a broad host capability. I know the PR notes call this out as future work, but from an escape-review perspective it is still the largest remaining surface: file URLs, upload paths, screenshot/PDF/video output paths, state save/load, `--cdp`, `--extension`, `--executable-path`, proxy/header options, and JS eval all need subcommand/path-aware policy if this tool is meant to be the restricted host bridge.

Nit: `timeoutMs` is caller-controlled and optional, and stdout/stderr are read fully into memory. That is probably not the central issue for this PR, but a default/max timeout and output cap would make even allowed `find`/`ls` commands less useful as a host-side DoS vector.

packages/cf-harness/src/tools/browser-host-command-policy.ts
```diff
@@ -0,0 +1,395 @@
...
+      command.name === "open" &&
+      args.slice(command.index + 1).some((arg) => /^file:/i.test(arg))
+    ) {
+      return deny("agent-browser open file: URLs are not allowed");
+    }
+  }
+  return allow([AGENT_BROWSER_COMMAND, ...args]);
+};
+
+const DISALLOWED_AGENT_BROWSER_COMMANDS = new Set([
```
willkelly, line 97: This deny list now covers the high-risk `agent-browser` surfaces we called out, but I think `state` still needs handling. `agent-browser state save/load <path>` can read/write host-side browser state files, so it should either be denied for this bridge or made path-aware and restricted to realpath-checked workspace/harness-owned paths.

willkelly (commented):
One remaining host-surface issue after the update: the big concerns from the earlier review are mostly addressed, but `agent-browser state save/load <path>` still appears allowed. That should either be denied for now or made path-aware with realpath checks against a workspace or harness-owned state directory.

packages/cf-harness/src/tools/browser-host-command-policy.ts
mathpirate: Addressed in 3cf17613f by denying the `agent-browser state` subcommand for this provisional bridge, with coverage for both direct `state save` and `--session ... state load` forms.

PR #3426, opened 2026-04-28
refactor(cli): swap raw JSON.stringify for {Compact,Indented}DebugString in non-test DEBUG+DISPLAY sites

danfuzz:
## Summary

Third cadence PR in the JSON.stringify → debug-stringifier migration arc for session 2026-065. Mirrors the pilot template established in PR #3411 (runner, 32 sites) and continued by PR #3416 (memory, 13 sites). Migrates **6 swap-eligible non-test sites in `packages/cli`** to use `toCompactDebugString` from `@commonfabric/data-model/value-debug`.

`packages/cli` was sandbox-clean (no synthetic `commonfabric` import), so this PR is mechanically the simplest of the three so far — it's primarily about consumer-side classification discipline, not technical migration challenges.

## Scope discipline note

The original brief estimated 11 swap-eligible sites based on per-package shape tally. Pre-edit per-site survey reclassified **3 sites as WIRE** on contract grounds (each is reachable via a `--json`-flag-documented user-facing API contract). Post-review, Dan's review reverted **1 additional site** in `render.ts`, treating the file holistically rather than migrating one of its `JSON.stringify` calls in isolation. Final scope: **6 sites migrated, 7 WIRE excluded** (3 directly per Remy's classification + 3 reclassified during survey + 1 reverted post-review). The function-name-as-contract / consumer-shape-as-stable-identity lens caught what raw shape-tally counts missed — exactly the discipline this arc was designed to apply.

### D1 — Error message text (4 sites, compact)

- `lib/exec-schema.ts:1031` — `schemaShapeString` enum branch. The recursive `schemaShapeString` function is consumed by an `Error` throw (line 932): `Handler requires input. Expected type: ${typeShape}`. D1.
- `lib/test-runner.ts:975` — `Test pattern must return { tests: TestStep[] }. Got: ${...}` Error message.
- `lib/test-runner.ts:1087` — `Test step at index ${i} must have either 'action' or 'assertion' key. Got: ${...}` Error message.
- `lib/test-runner.ts:1329` — `error: \`Expected true, got ${...}\`` assertion failure message; consumer at line 1491 is `console.log(\`  ✗ Error: ${result.error}\`)`. D1.

### DISPLAY — `--help` text, compact (2 sites)

- `lib/exec-schema.ts:611` — `schemaEnumSummary` produces `"Allowed: x | y | z."` text in `cf X --help` output.
[...]

PR #3436, opened 2026-04-29
feat(api,runner): expose debug-stringifier helpers to patterns

danfuzz:
Exposes two debug-stringifier helpers from `@commonfabric/data-model/value-debug` to pattern code via the synthetic `commonfabric` builder surface.

## What this PR exposes

Two helpers, available to pattern code via `import { ... } from "commonfabric"`:

- **`toCompactDebugString(value: unknown, maxLength?: number): string`** — single-line, length-bounded debug rendering (the `maxLength` parameter is per PR #3431).
- **`toIndentedDebugString(value: unknown): string`** — multi-line, indented debug rendering.

Both names are idiomatic with the rest of `value-debug.ts` and share the `unknown` input contract established in PR #3409 (helpers widened from `FabricValue` to `unknown` so pattern code can pass anything).

These are **debug-grade** stringifiers — outputs are not stable, not round-trippable, and not part of the JSON-encoding contract. They're for human-readable diagnostic output (logs, error messages, dev assertions). Pattern code that needs wire-format JSON should continue to use the existing JSON-encoding paths.

## Why now

Session 2026-065 migrated ~50 internal call sites from `JSON.stringify` to these helpers across `packages/runner`, `packages/memory`, and `packages/cli` (PRs #3411, #3416, #3426). The patterns-side of that migration was deferred because the synthetic `commonfabric` sandbox surface didn't expose the helpers — patterns can only import what `commonfabric` re-exports, and the helpers weren't there. This PR closes that gap so the pattern migration can proceed as a follow-up.

## Mechanism (three touch points, +21/-0 across 3 files)

1. **`packages/api/index.ts`** — adds `ToCompactDebugStringFunction` and `ToIndentedDebugStringFunction` type aliases plus matching `declare const` exports, next to `safeDateNow` / `nonPrivateRandom` / `byRef`. The `commonfabric.d.ts` consumed by the pattern compiler is a symlink to this file (`packages/static/assets/types/commonfabric.d.ts -> ../../../api/index.ts`), so edits here flow through `StaticCache.getText()` to the pattern-compilation pipeline automatically.

2. **`packages/runner/src/builder/types.ts`** — adds the two members to `BuilderFunctionsAndConstants` using the `typeof import(...)` form to match the immediately-adjacent `// Fabric value classes` block.

3. **`packages/runner/src/builder/factory.ts`** — named imports from `@commonfabric/data-model/value-debug`, wired into the `commonfabric` builder-functions object under a new `// Debug stringifiers (helpers exposed for pattern code)` comment.

## Form-choice rationale

Two different declaration conventions appear in two different files, each chosen by **immediate-context match**:

- **`api/index.ts`** uses the function-type-alias + `declare const` form, matching the `safeDateNow` / `nonPrivateRandom` / `byRef` precedent in this file. Required because `index.ts` cannot use relative imports (it's loaded as a flat type surface by the pattern compiler).
- **`runner/src/builder/types.ts`** uses the `typeof import(...)` form, matching the immediately-adjacent Fabric-value-classes block in the same interface.

Local consistency over codebase-wide consistency — calling this out here so reviewers don't have to ask.

## Scope-out

- **Patterns-side migration** — the actual `JSON.stringify` → debug-stringifier swap inside pattern code is a separate follow-up PR; this one only exposes the surface.
- **No semantic changes** to `toCompactDebugString` / `toIndentedDebugString` themselves; they're re-exported as-is from `@commonfabric/data-model/value-debug`.

## Validation

- `deno fmt --check`, `deno lint`, `deno task check` (full): clean.
- `deno task test` on `runner` (407/0) and `js-compiler` (3/0).
- Runtime smoke check: both helpers reachable on `createBuilder().commonfabric` and produce expected output with the 2-arg `maxLength` signature.

### DIFF
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -2121,6 +2121,11 @@ export interface PatternEnvironment {
 export type GetPatternEnvironmentFunction = () => PatternEnvironment;
 export type NonPrivateRandomFunction = () => number;
 export type SafeDateNowFunction = () => number;
+export type ToCompactDebugStringFunction = (
+  value: unknown,
+  maxLength?: number,
+) => string;
+export type ToIndentedDebugStringFunction = (value: unknown) => string;
 
 /**
  * Compare two cells or values for equality after resolving, i.e. after
@@ -2165,6 +2170,8 @@ export declare const byRef: ByRefFunction;
 export declare const getPatternEnvironment: GetPatternEnvironmentFunction;
 export declare const nonPrivateRandom: NonPrivateRandomFunction;
 export declare const safeDateNow: SafeDateNowFunction;
+export declare const toCompactDebugString: ToCompactDebugStringFunction;
+export declare const toIndentedDebugString: ToIndentedDebugStringFunction;
 
 export interface UiActionProps {
   readonly as?: string;
--- a/packages/runner/src/builder/factory.ts
+++ b/packages/runner/src/builder/factory.ts
@@ -57,6 +57,10 @@ import {
   FabricEpochNsec,
 } from "@commonfabric/data-model/fabric-epoch";
 import { FabricHash } from "@commonfabric/data-model/fabric-hash";
+import {
+  toCompactDebugString,
+  toIndentedDebugString,
+} from "@commonfabric/data-model/value-debug";
 import { freezeVerifiedPlainData } from "../sandbox/plain-data.ts";
 import { nonPrivateRandom, safeDateNow } from "./safe-builtins.ts";
 import {
@@ -214,6 +218,10 @@ export const createBuilder = (options: CreateBuilderOptions = {}): {
     FabricEpochNsec,
     FabricEpochDays,
     FabricHash,
+
+    // Debug stringifiers (helpers exposed for pattern code)
+    toCompactDebugString,
+    toIndentedDebugString,
   } as BuilderFunctionsAndConstants & {
     __cfHelpers?: BuilderFunctionsAndConstants;
   };
--- a/packages/runner/src/builder/types.ts
+++ b/packages/runner/src/builder/types.ts
@@ -349,6 +349,12 @@ export interface BuilderFunctionsAndConstants {
   FabricEpochDays:
     typeof import("@commonfabric/data-model/fabric-epoch").FabricEpochDays;
   FabricHash: typeof import("@commonfabric/data-model/fabric-hash").FabricHash;
+
+  // Debug stringifiers
+  toCompactDebugString:
+    typeof import("@commonfabric/data-model/value-debug").toCompactDebugString;
+  toIndentedDebugString:
+    typeof import("@commonfabric/data-model/value-debug").toIndentedDebugString;
 }
 
 // Runtime interface needed by createCell


danfuzz (commented):
**Verdict: APPROVE-equivalent** (posting as COMMENT-class; co-authored PR limitation precludes a formal APPROVE from this account).

Reviewed at HEAD `c46d7ffe5`. +21/-0 across 3 files, single commit. This is **API surface expansion**, so the migration-arc six-point lens does not apply; the review focused on five surface-decision dimensions, all clean.

### 1. Symbol selection

`toCompactDebugString` and `toIndentedDebugString` are the two `export function`s in `packages/data-model/value-debug.ts`. Surface is complete; no other exposed-eligible helpers in that module. The deferred-patterns survey (`coordination/docs/2026-04-28-patterns-survey-deferred/`) justifies these as the migration-arc-eligible primitives. Right primary pair.

### 2. Signature shape

`toCompactDebugString(value: unknown, maxLength?: number)` matches HEAD `value-debug.ts:185-189` exactly post-#3431 (`2a8f88726`). Optional `maxLength` is the right surface — pattern callers who don't want truncation just omit it, those who do get a single-call form rather than a separate truncation helper. Forcing pattern code to chain truncation manually would be friction without benefit. `toIndentedDebugString(value: unknown)` is fixed-form (indent=2 baked in); also right — patterns don't need to choose indent, and exposing it would invite footguns.

### 3. Form-choice convention (different in the two files)

Both file choices match their immediate-context precedent:

- `packages/runner/src/builder/types.ts` uses `typeof import("@commonfabric/data-model/value-debug").X` matching the immediately preceding Fabric-value-classes block (`types.ts:343-351`).
- `packages/api/index.ts` uses function-type-alias + `declare const` matching `safeDateNow` / `nonPrivateRandom` / `byRef` (`api/index.ts:2122-2123` + `:2164-2167`).

Each defensible; immediate-context match is sound. Worth noting that `types.ts` also has an alias-form precedent slightly earlier (line 315 `safeDateNow: SafeDateNowFunction`); Cedar chose proximity-to-fabric-block over proximity-to-function-helpers. Defensible either way.

### 4. Cross-package mechanism (symlink)

Verified `packages/static/assets/types/commonfabric.d.ts` is a symlink to `../../../api/index.ts`. Edits to `api/index.ts` flow through to pattern-compilation consumers (`packages/static/assets.ts` flat-copy mapping; `packages/js-compiler/test/{source-map,typescript.compiler}.test.ts` via `staticCache.getText("types/commonfabric.d.ts")`) automatically — no regeneration step. Mechanism robust.

**NIT (non-blocking, your call):** A one-line doc-comment near the top of `packages/api/index.ts` noting that this file is published to patterns through a symlink at `packages/static/assets/types/commonfabric.d.ts` would help future contributors who edit `api/index.ts` without realizing the published-to-patterns coupling. Truly mild — a one-liner in `coordination/facts.md`'s "Known `packages/api` constraints" would also suffice. Surfacing here in case the reviewing humans have an opinion either way.

### 5. Cross-package import shape

`packages/runner/src/builder/factory.ts` adds `import { toCompactDebugString, toIndentedDebugString } from "@commonfabric/data-model/value-debug"`. The `runner → data-model/value-debug` edge already has 10+ existing import sites in `packages/runner/src/`, so this is a same-edge new sub-path — not a first-time cross-package import. Cedar's `deno task check` clean is consistent; no transitive strict-mode surprise risk.

### Summary

Surface expansion is well-scoped, signatures track HEAD, form choices are immediate-context-matched, mechanism is sound, import shape is same-edge. No blocking issues; one optional doc-comment NIT for human-reviewer consideration.

— `reviewer-flux` (Claude Opus 4.7, 1M context)

mathpirate (approved):
lgtm

PR #3439, opened 2026-04-29
Prepare cf-harness Fabric FUSE mounts

mathpirate:
## Summary
- add typed Docker/runsc additional mounts for Fabric FUSE, defaulting to /fabric
- keep /workspace and /fabric as distinct sandbox roots, with Docker mount args and path guards updated accordingly
- surface workspace/Fabric mount status in cf-harness capability diagnostics

## Notes
- intentionally does not add CLI flags or real FUSE probing yet
- meant as prep for Wilk's host-FUSE-bind-mounted-into-runsc work

## Tests
- deno check packages/cf-harness/src/index.ts
- deno test -A packages/cf-harness/test/docker-runsc-sandbox.test.ts
- deno test -A packages/cf-harness/test/diagnostics.test.ts packages/cf-harness/test/artifacts.test.ts packages/cf-harness/test/tools-execution.test.ts
- cd packages/cf-harness && deno task test
- git diff --check

### DIFF
--- a/packages/cf-harness/src/diagnostics.ts
+++ b/packages/cf-harness/src/diagnostics.ts
@@ -4,6 +4,8 @@ import { ProcessTimeoutError } from "./sandbox/process-runner.ts";
 import type {
   SandboxRuntime,
   SandboxRuntimeDescription,
+  SandboxRuntimeMountDescription,
+  SandboxRuntimeMountKind,
 } from "./sandbox/types.ts";
 import type { HarnessPolicyEvent } from "./contracts/policy.ts";
 import type { ToolOutputId } from "./contracts/tool-result.ts";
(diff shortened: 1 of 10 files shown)


willkelly (approved):
LGTM for this preparatory mount plumbing. The typed additional mount shape, separate `/workspace` and `/fabric` roots, path guards, diagnostics, and unit coverage all look appropriate for getting cf-harness ready for a host FUSE bind-mounted into runsc path.

Two follow-ups I would keep explicit:

- Test gap: this PR has good unit coverage and the existing CI/fuse checks are green, but it still does not exercise cf-harness with a real additional `/fabric` mount through `runsc-cfc`. Before we rely on this operationally, we should add/run an env-gated integration case that mounts a real host Fabric FUSE tree, passes it as `additionalMounts`, and verifies cf-harness can read/navigate `/fabric` inside the sandbox.

- Sandbox capability gap: cf-harness currently has no way to pass CFC labels for the arguments/context it is handing to the sandbox. Because each tool call is a separate sandbox invocation, we can lose label/context continuity across separate tool calls when we should not. I do not think that should block this PR; it is follow-up work for Wilk's sandbox capability path rather than Gideon's mount-prep patch.

PR #3445, opened 2026-04-30
Add cf-harness CFC invocation context plumbing

mathpirate:
## Summary
- add a typed `HarnessCfcInvocationContext` contract for sandbox-bound tool calls
- attach optional CFC invocation context to sandbox command/shell requests
- persist redacted invocation summaries in run state and policy traces for `bash`, `read_file`, and `write_file`

## Notes
This is intentionally observational/preparatory. It does not require gVisor/runsc to consume the context yet, and it does not enforce argv/env/cwd/stdin labels. The goal is to give Wilk's runtime-side work a concrete cf-harness-side shape to review and coordinate against.

The persisted summaries redact raw command/stdin/arg values via byte counts and SHA-256 digests. Prompt-slot and run-manifest provenance are included where available.

## Tests
- `deno check packages/cf-harness/src/index.ts packages/cf-harness/integration/engine.integration.test.ts`
- `deno test -A packages/cf-harness/integration/engine.integration.test.ts` (6 ignored without integration env vars)
- `git diff --check`


packages/cf-harness/src/tools/read-file.ts
```diff
@@ -80,26 +80,35 @@ export const readFileTool: HarnessToolDefinition<
...
-      args: [
-        resolvedPath,
-        input.maxBytes !== undefined ? String(input.maxBytes) : "",
-      ],
+      command,
+      args,
+      cfcInvocationContext: await context.createCfcInvocationContext({
+        toolId: "read_file",
+        operation: "shell",
+        cwd: context.currentDir,
```
willkelly, line 109: This context says the read runs with `cwd: context.currentDir`, but the `runShell` request above does not pass `cwd`, so Docker executes from the sandbox default working directory. That is okay for the current absolute `$1` path, but it makes the invocation context inaccurate as soon as the runtime starts trusting it. Please pass `cwd: context.currentDir` here or make the context reflect the actual default cwd.

packages/cf-harness/src/tools/write-file.ts
```diff
@@ -84,38 +84,48 @@ export const writeFileTool: HarnessToolDefinition<
...
-        "esac",
-      ].join("\n"),
-      args: [resolvedPath, mode, String(input.createParents ?? false)],
+      command,
+      args,
       stdinText: input.content,
+      cfcInvocationContext: await context.createCfcInvocationContext({
+        toolId: "write_file",
+        operation: "shell",
+        cwd: context.currentDir,
```
willkelly, line 125: Same cwd mismatch as `read_file`: the invocation context records `context.currentDir`, but this `runShell` call does not pass a cwd, so the runtime execution cwd is the sandbox default. Since this contract is meant to become runtime-facing, the recorded cwd should match execution. Please pass `cwd: context.currentDir` or record the actual default cwd.

willkelly (commented):
Directionally aligned from my side. This looks like the right cf-harness-side foundation: a typed, redacted invocation context that records what the harness asked the sandbox/runtime to do, without pretending runsc/gVisor consumes or enforces it yet.

I do not think either of us needs to change course. I would frame this PR as the invocation-context/audit contract, complementary to the runtime result-sidecar/output-mediation path. The next coordination point is the trusted transport into runsc/gVisor, likely host-created JSON sidecar/FD or runtime annotation rather than guest-controlled env.

Two correctness details are worth fixing before we rely on this context as runtime input: cwd needs to match actual sandbox execution for the file tools, and invocation timestamps should use a fresh clock tick. Also, if the runtime side needs explicit labels on argv/stdin/cwd/env rather than provenance-only summaries, we should add those fields before wiring the transport.

packages/cf-harness/src/engine.ts
```diff
@@ -778,6 +785,52 @@ export class CfHarnessEngine {
...
+    command?: string;
+    argv?: readonly string[];
+    args?: readonly string[];
+    stdinText?: string;
+    env?: Record<string, string>;
+  }): Promise<HarnessCfcInvocationContext> {
+    const invocation = await createHarnessCfcInvocationContext({
+      sequence: (this.#runState.cfcInvocationContexts ?? []).length + 1,
+      runId: this.#runState.runId,
+      createdAt: this.#runState.updatedAt,
```
willkelly: This should use a fresh `this.#now()` value for the invocation. Reusing `this.#runState.updatedAt` makes `createdAt` and the append timestamp stale, so the run-state update can add a context without advancing `updatedAt`.

mathpirate:
> Directionally aligned from my side. This looks like the right cf-harness-side foundation: a typed, redacted invocation context that records what the harness asked the sandbox/runtime to do, without pretending runsc/gVisor consumes or enforces it yet.
> 
> I do not think either of us needs to change course. I would frame this PR as the invocation-context/audit contract, complementary to the runtime result-sidecar/output-mediation path. The next coordination point is the trusted transport into runsc/gVisor, likely host-created JSON sidecar/FD or runtime annotation rather than guest-controlled env.
> 
> Two correctness details are worth fixing before we rely on this context as runtime input: cwd needs to match actual sandbox execution for the file tools, and invocation timestamps should use a fresh clock tick. Also, if the runtime side needs explicit labels on argv/stdin/cwd/env rather than provenance-only summaries, we should add those fields before wiring the transport.

Thanks, this framing makes sense to me.

I pushed 350ac7566 to address the two concrete correctness issues: invocation contexts now use a fresh mutation timestamp, and read_file/write_file pass the same cwd to runShell that they record in the CFC invocation context. CI is green now.

I agree the explicit argv/stdin/cwd/env label shape should be the next coordination point before this becomes trusted runtime input. I’ve left this PR as audit/provenance plumbing rather than claiming runsc/gVisor consumption yet.

packages/cf-harness/src/tools/read-file.ts
mathpirate: Fixed in 350ac7566. read_file now passes cwd: context.currentDir into runShell, and the tool test asserts both the sandbox request cwd and CFC invocation context cwd line up.

packages/cf-harness/src/tools/write-file.ts
mathpirate: Fixed in 350ac7566. write_file now passes cwd: context.currentDir into runShell as well, with regression coverage for the currentDir inherited from a previous bash call.

packages/cf-harness/src/engine.ts
mathpirate: Fixed in 350ac7566. #createCfcInvocationContext now captures a fresh const now = this.#now() and uses it for both invocation createdAt and the run-state append timestamp. I also added an engine regression test for the persisted timestamp behavior.

willkelly (approved):
Re-reviewed head `350ac7566`.

The previous correctness issues are addressed: invocation contexts now use a fresh mutation timestamp, and `read_file`/`write_file` pass the same cwd to `runShell` that they record in the CFC invocation context. The added regression coverage matches those fixes.

This remains directionally aligned with the gVisor side we discussed: this PR defines and persists the cf-harness invocation/audit envelope, while the trusted runtime transport can land as a follow-up. On the gVisor side I’m shaping that as a host-controlled `<containerID>.json` sidecar donated into runsc/boot by FD, so the next cf-harness step is to write this context after `docker create` and before `docker start` when the configured invocation-context directory is available.

No blocking issues from my side.

PR #3446, opened 2026-04-30
Harden cf-harness subagent return channel

mathpirate:
## Summary
- add optional delegate_task returnSchema support for schema-validated child JSON returns
- retain raw child returns in child artifacts while exposing sanitized structuredReturn values to the parent
- linkify free-form strings as @link objects while preserving booleans, numbers, string enums/consts, and explicit link objects
- document the return channel and export the shared schema validator/subagent return helper

## Tests
- deno check packages/cf-harness/mod.ts packages/runner/src/cfc/mod.ts
- deno lint packages/cf-harness/src/subagent-return.ts packages/cf-harness/src/prompt-loop.ts packages/cf-harness/src/contracts/subagent.ts packages/cf-harness/src/contracts/policy.ts packages/cf-harness/src/tools/delegate-task.ts packages/cf-harness/test/subagent-return.test.ts packages/cf-harness/test/prompt-loop.test.ts packages/runner/src/cfc/mod.ts

### DIFF
--- a/packages/cf-harness/src/subagent-return.ts
+++ b/packages/cf-harness/src/subagent-return.ts
@@ -143,6 +143,33 @@ const matchingBranch = (
     : resolveSchemaForValidation(branch, fullSchema);
 };
 
+const isEmptySchemaObject = (schema: JSONSchema): boolean =>
+  isRecord(schema) && Object.keys(schema).length === 0;
+
+const combineAllOf = (schemas: readonly JSONSchema[]): JSONSchema => {
+  const constrained = schemas.filter((schema) =>
+    schema !== true && !isEmptySchemaObject(schema)
+  );
+  if (constrained.some((schema) => schema === false)) {
+    return false;
+  }
+  if (constrained.length === 0) {
+    return true;
+  }
+  if (constrained.length === 1) {
+    return constrained[0]!;
+  }
+  return { allOf: constrained };
+};
+
+const schemaWithoutBranchKeyword = (
+  schema: Record<string, unknown>,
+  keyword: "anyOf" | "oneOf",
+): JSONSchema => {
+  const { [keyword]: _ignored, ...rest } = schema;
+  return rest as JSONSchema;
+};
+
 const schemaForValue = (
   schema: JSONSchema,
   value: unknown,
@@ -152,15 +179,19 @@ const schemaForValue = (
   if (!isRecord(resolved)) {
     return resolved;
   }
+  let base: JSONSchema = resolved;
+  const branches: JSONSchema[] = [];
   const oneOf = matchingBranch(resolved.oneOf, value, fullSchema);
   if (oneOf !== undefined) {
-    return oneOf;
+    base = schemaWithoutBranchKeyword(base as Record<string, unknown>, "oneOf");
+    branches.push(oneOf);
   }
   const anyOf = matchingBranch(resolved.anyOf, value, fullSchema);
   if (anyOf !== undefined) {
-    return anyOf;
+    base = schemaWithoutBranchKeyword(base as Record<string, unknown>, "anyOf");
+    branches.push(anyOf);
   }
-  return resolved;
+  return branches.length === 0 ? resolved : combineAllOf([base, ...branches]);
 };
 
 const childSchemaForKey = (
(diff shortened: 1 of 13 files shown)


mathpirate:
Fixed the Cubic-identified prompt conflict in a6e7f3e44: schema-mode child system prompts now omit the prose summary instruction and require only the JSON value requested by the return schema. Added a regression assertion for that prompt shape. @cubic-dev-ai please re-run review.

packages/cf-harness/src/subagent-return.ts
```diff
@@ -0,0 +1,263 @@
...
+  }
+  if (isRecord(value)) {
+    if (
+      valueIsOpaqueLinkObject(value) &&
+      schemaDeclaresOpaqueLinkObject(effectiveSchema)
+    ) {
+      return { value, linkedStringCount: 0 };
+    }
+    let linkedStringCount = 0;
+    const entries = Object.entries(value).map(([key, child]) => {
```
seefeldb, line 348: i think this would allow any keys if the schema allows additionalProperties, but that would allow prompt injection material to be transferred in the keys.

effectively, we can only support pre-defined keys, so if there are any additional keys, the entire object has to become an opaque value

packages/cf-harness/src/subagent-return.ts
```diff
@@ -0,0 +1,263 @@
...
+      path.map((segment) =>
+        String(segment).replaceAll("~", "~0").replaceAll("/", "~1")
+      ).join("/")
+    }`;
+
+const returnLink = (
+  childRunId: string,
+  path: readonly (string | number)[],
+): { "@link": string } => ({
+  "@link": `cf-harness://subagent-return/${encodeURIComponent(childRunId)}#${
```
seefeldb: use "opaque:${encodeURIComponent(childRunId)}..." ?

mathpirate:
Addressed the latest Berni/Cubic return-channel comments in `8451a85c6`:

- resolved `$ref`/`allOf` schema constraints before deciding whether string returns are safe control values vs opaque links
- made objects with keys outside predefined schema `properties` opaque as a whole, even if `additionalProperties` validates
- changed generated return handles to `opaque:<encoded-child-run-id>#<json-pointer>`
- added regression coverage for ref/allOf-constrained strings and additionalProperties key-channel opacity

@cubic-dev-ai please re-run review.

mathpirate:
Confirmed Cubic had a separate later round against `8451a85c6`. Addressed those three schema-composition findings in `0c1faceac`:

- selected `oneOf`/`anyOf` branches now compose with the base schema instead of replacing it
- applicable `additionalProperties` constraints now compose with allOf branch property constraints
- array item schema lookup now resolves refs and traverses allOf branches

Added regression tests for all three cases. @cubic-dev-ai please re-run review.

mathpirate:
Addressed the latest Cubic P2 in `fd92dd955`: `schemaDeclaresOpaqueLinkObject` now treats omitted `additionalProperties` as closed for explicit `{ "@link": ... }` schemas, matching the runner/CFC validation semantics. I also updated the regression test so the anyOf link branch relies on default-closed behavior.

Validated locally:
- `deno test --allow-read --allow-write=/tmp,/var/folders --allow-run packages/cf-harness/test/subagent-return.test.ts`
- `deno lint packages/cf-harness/src/subagent-return.ts packages/cf-harness/test/subagent-return.test.ts`
- `deno check packages/cf-harness/mod.ts packages/runner/src/cfc/mod.ts`
- `deno task test` in `packages/cf-harness`

@cubic-dev-ai please re-run review.

mathpirate:
Addressed the latest issue cubic found in `6db7371fc`: explicit `@link` preservation now uses a value-aware check over all matching union branches, so an overlapping broad `anyOf` object branch can no longer hide a later link-object branch. Added a regression test for that overlap case, rebased the branch onto latest `origin/main`, and force-pushed for fresh CI.

Validated locally:
- `deno test --allow-read --allow-write=/tmp,/var/folders --allow-run packages/cf-harness/test/subagent-return.test.ts`
- `deno lint packages/cf-harness/src/subagent-return.ts packages/cf-harness/test/subagent-return.test.ts`
- `deno check packages/cf-harness/mod.ts packages/runner/src/cfc/mod.ts`
- `deno task test` in `packages/cf-harness`

@cubic-dev-ai please re-run review.

PR #3452, opened 2026-04-30
Add cf-harness CFC invocation sidecar transport

mathpirate:
## Summary
- add an optional `sidecar` CFC invocation-context transport to the Docker/runsc sandbox config
- write `<containerID>.json` after `docker create` and before `docker start` when a sandbox request includes a CFC invocation context
- document the matching `--cfc-invocation-context-dir` / `CF_HARNESS_RUNSC_CFC_INVOCATION_CONTEXT_DIR` setup

## Notes
This follows Wilk's provisional gVisor sidecar interface in commontoolsinc/gvisor#17. The transport is intentionally mechanical: it moves the existing cf-harness invocation-context JSON into runsc's host-controlled sidecar directory, but it does not add or claim runtime enforcement for argv/stdin/cwd/env labels.

The config is shaped as a transport object so a future socket/runtime-annotation path can replace the file sidecar without changing tool-call code.

## Tests
- `deno test -A packages/cf-harness/test/docker-runsc-sandbox.test.ts`
- `deno check packages/cf-harness/src/index.ts packages/cf-harness/integration/engine.integration.test.ts`
- `deno test -A packages/cf-harness/integration/engine.integration.test.ts` (6 ignored without integration env vars)
- `git diff --check`

### DIFF
--- a/packages/cf-harness/README.md
+++ b/packages/cf-harness/README.md
@@ -211,12 +211,22 @@ the same host result directory that `cf-harness` reads. Configure runsc with
 `CF_HARNESS_RUNSC_CFC_RESULT_DIR=/path/to/results` or pass `cfcResultDir` in the
 explicit sandbox config.
 
+CFC invocation context transport is similarly coordinated through a host sidecar
+directory. Configure runsc with
+`--cfc-invocation-context-dir=/path/to/invocations`, then set
+`CF_HARNESS_RUNSC_CFC_INVOCATION_CONTEXT_DIR=/path/to/invocations` or pass
+`cfcInvocationContextDir` in the explicit sandbox config. `cf-harness` writes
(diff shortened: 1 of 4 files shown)


willkelly (approved):
Re-reviewed against commontoolsinc/gvisor#17. This lines up with the provisional runtime side:

- resolves a host-controlled invocation sidecar directory
- writes `<containerID>.json` after `docker create` and before `docker start`
- preserves absence as a no-op, matching the runsc behavior when no sidecar is present
- keeps the scope explicitly transport-only and does not claim argv/stdin/cwd/env label enforcement yet

The remaining dependency is operational rather than a blocker for this PR: runsc still needs to be configured with `--cfc-invocation-context-dir` pointing at the same directory, and the explicit input-label contract remains a follow-up before enforcement.

No blocking issues from my side.

PR #3456, opened 2026-04-30
Add --fabric-mount CLI flag to cf-harness (CT-1525)

bfollington:
## Summary

- Adds `--fabric-mount <host-path>` CLI flag that threads a fabric-fuse mount into the engine's `additionalMounts` on `DockerRunscSandboxConfig`
- Injects fabric-aware guidance into the operator system prompt when the mount is present ("A Common Fabric space is mounted at /fabric")
- Adds `additionalMounts` to `CreateHarnessEngineOptions` so the CLI doesn't need to construct the full sandbox config

## What works

- CLI flag parsing, absolute/relative path resolution
- Mount threads through to Docker bind mount (`--mount type=bind,src=...,dst=/fabric`)
- Sandbox agent can **read** FUSE-mounted spaces (pieces, index.md, metadata)
- System prompt correctly tells the model about `/fabric`
- All 186 unit tests pass

## Known issue: Docker→FUSE writes

Writes from inside Docker containers (both runc and runsc-cfc) to a host FUSE-T mount silently fail — `exit 0` but content doesn't persist. Host-process writes to the same FUSE mount work fine. This is a Docker Desktop for Mac + FUSE-T interaction issue, not a cf-harness bug. Filed separately.

## Test plan

- [x] `deno task test` in `packages/cf-harness` — 186 pass
- [x] Manual: harness reads pieces from FUSE mount via `/fabric`
- [x] Integration: write-through needs Docker/FUSE-T fix

### DIFF
--- a/packages/cf-harness/src/cli.ts
+++ b/packages/cf-harness/src/cli.ts
@@ -583,8 +583,14 @@ export const parseCfHarnessCliArgs = async (
       "cfc enforcement mode must be one of disabled, observe, enforce-explicit, enforce-strict",
     );
   }
-  const fabricMount = typeof args["fabric-mount"] === "string"
-    ? resolve(cwd, args["fabric-mount"])
+  const rawFabricMount = typeof args["fabric-mount"] === "string"
+    ? args["fabric-mount"].trim()
+    : undefined;
+  if (rawFabricMount !== undefined && rawFabricMount === "") {
+    throw new Error("--fabric-mount requires a non-empty path");
+  }
+  const fabricMount = rawFabricMount !== undefined
+    ? resolve(cwd, rawFabricMount)
     : undefined;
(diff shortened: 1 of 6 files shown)


packages/cf-harness/src/cli.ts
```diff
@@ -488,6 +492,9 @@ export const parseCfHarnessCliArgs = async (
       "cfc enforcement mode must be one of disabled, observe, enforce-explicit, enforce-strict",
     );
   }
+  const fabricMount = typeof args["fabric-mount"] === "string"
+    ? resolve(cwd, args["fabric-mount"])
```
mathpirate: `parseArgs` returns an empty string for `--fabric-mount` with no value, and this code resolves that to the launcher cwd. Because the resolved path is then added as a sandbox bind mount at `/fabric`, a mistyped flag can silently expose an extra host directory outside `--workspace`. Reject empty or whitespace-only `--fabric-mount` values before calling `resolve`.

PR #3460, opened 2026-04-30
Tighten browser subagent workspace writes

mathpirate:
## Summary
- remove write_file from the browser subagent profile so browser children get bash-no-sandbox plus read_file only
- update the browser structured-return test to use a sibling artifact root outside the workspace while confirming raw child artifacts are still retained
- document browser artifact-root placement and the future hardening work around reserved artifact paths, opaque handles, and declassification

## Tests
- deno test --allow-read --allow-write=/tmp,/var/folders --allow-run packages/cf-harness/test/prompt-loop.test.ts
- deno check packages/cf-harness/mod.ts packages/runner/src/cfc/mod.ts
- deno lint packages/cf-harness/src/contracts/subagent.ts packages/cf-harness/src/prompt-loop.ts packages/cf-harness/test/prompt-loop.test.ts
- deno task check (packages/cf-harness)
- git diff --check

### DIFF
--- a/packages/cf-harness/README.md
+++ b/packages/cf-harness/README.md
@@ -172,16 +172,27 @@ The provisional browser profile is the only CLI-supported path to
 `agent-browser`, while the parent still receives only the normal sanitized
 subagent result. Browser/page output is treated as untrusted child-local data;
 with a `returnSchema`, parent-visible free-form strings are replaced by opaque
-links while raw observations stay in child artifacts. The host shell is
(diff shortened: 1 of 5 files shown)


mathpirate:
Addressed the cubic finding in 31009bae6: the browser structured-return test now rejects both artifactRoot equality with the workspace root and descendant paths under it, and also asserts the child artifact root is under the intended sibling artifacts directory. @cubic-dev-ai please rerun review.

PR #3464, opened 2026-05-01
runner: prove and remove dead defensive JSON.parse on path:["source"] reads

danfuzz:
## Summary

Two defensive `JSON.parse` blocks in `packages/runner/src/` were originally
introduced (mid-2025) to convert JSON-stringified link records back to objects
when read from `path: ["source"]` storage rows. Subsequent refactoring of the
storage boundary (most notably PR #2971 in March 2026, which routed the `is`
column through `valueFromJson()` on every read) made those defensive parses
**structurally unreachable**: the storage layer now always returns the
underlying object, never the JSON-stringified form.

This PR:

1. **Adds regression-coverage tests** in `packages/runner/test/cell-core.test.ts`
   that pin the current observable behavior — `setSourceCell` → commit →
   `getSourceCell` round-trip, and a raw `tx.read` value-shape probe — under
   both `unifiedJsonEncoding` flag states.
2. **Removes the dead defensive-parse code** at `transaction.ts:280` and
   `cell.ts:1460` (and their `startsWith('{"/":')` / `typeof === "string"`
   guards plus the surrounding source-path special-case logging block in
   `transaction.ts`).
3. **Reworks `getSourceCell` for type honesty.** The misleading
   `as string | undefined` cast on `readOrThrow` is dropped; the `isRecord`
   guard is inverted to `if (!isRecord(sourceCellId)) { throw ... }` so the
   type narrows on the happy path; a typed local
   `const sourceCellIdString = toURI(sourceCellId)` flows unconditionally
   into `createCell()`; and the redundant double-`toURI` call at the `id:`
   line is removed.

## Independent reviewer verdict

`reviewer-flux` independently verified the dead-code argument at prior tip
`73f56e918` (review converged with APPROVE-equivalent on the deletion). Each
of seven load-bearing claims was checked:

- **Producer audit complete.** Only two `path: ["source"]` references in
  production code, both at `cell.ts:1449` (reader) and `cell.ts:1484`
  (writer). The writer emits an object literal; no string producer.
- **Storage-side decode wiring confirmed.** `valueFromJson` is invoked at all
  three `memory/space.ts` read sites under both flag states.
- **Within-tx novelty short-circuit confirmed.** Within-transaction reads
  return the object literal by reference from the chronicle's novelty map;
  the defensive parse can't fire.
- **Post-deletion `transaction.ts:read()` symmetric with `write()`.** No
  asymmetry introduced by removing the source-path special-case block.
- **Post-deletion `getSourceCell` integrity intact.** The object-form path
  (`isRecord` branch) covers the only producer; deletion of the string-form
  branch is structurally safe.
- **Deleted synthetic test was the only test exercising `transaction.ts:280`.**
  No other test relies on the defensive-parse behavior; nothing is silently
  weakened.
- **PR body's listed un-covered edge cases are accurate.** The two cases
  named in "Edge cases" below are the right gaps for the human discussion.

### The two sites

Both sites guard a `JSON.parse` with the same shape:

```ts
// runner/src/storage/transaction.ts (pre-PR, lines 277-296)
if (typeof value === "string" && value.startsWith('{"/":')) {
  // Parse the JSON string to return an object
  const parsedValue = JSON.parse(value);
  result.ok = { address: result.ok.address, value: parsedValue };
  // ...logging...
}
```

```ts
// runner/src/cell.ts (pre-PR, lines 1457-1461)
} else if (
  typeof sourceCellId === "string" && sourceCellId.startsWith('{"/":')
) {
  sourceCellId = toURI(JSON.parse(sourceCellId));
}
```

Both fire **only** when the value read from `path: ["source"]` is a string
beginning with `{"/":`. The argument below shows that (a) no producer in the
codebase emits such a value to that path, and (b) the storage layer's own
read path always returns the value in object form, regardless of flag state.

### Producer audit

`grep` for `path: ["source"]` writes/reads in `packages/memory/` and
`packages/runner/src/`:

- **One reader**: `packages/runner/src/cell.ts:1449` (in `getSourceCell`).
- **One writer**: `packages/runner/src/cell.ts:1488` (in `setSourceCell`).

The writer is:

```ts
// runner/src/cell.ts:1488
this.tx.writeOrThrow(
  { ...this.link, path: ["source"] } as IMemorySpaceAddress,
  // TODO(@ubik2): Transition source links to sigil links?
  { "/": fromURI(sourceLink.id) },  // <-- an object literal, never a string
);
```

So the only producer in scope writes an object literal, not a string. There
is no in-tree path that produces a JSON-stringified `'{"/":"..."}'` value
into the `path: ["source"]` storage row.

### Storage-boundary decode under both flag states

The storage layer's read path runs through `memory/space.ts:562, 643, 701`,
each of which does:

```ts
revision.is = valueFromJson(row.is);
```

`valueFromJson` dispatches on the `unifiedJsonEncoding` flag:

```ts
// data-model/json-encoding.ts:113-122
export function valueFromJson(json, runtime) {
  if (jsonEncodingEnabled) {
    return valueFromJsonModern(json, runtime);  // strips fvj1: prefix, decodes
  } else {
    return valueFromJsonLegacy(json);
  }
}

// data-model/json-encoding-legacy.ts:62-64
export function valueFromJsonLegacy(json: string): FabricValue {
  return JSON.parse(json);
}
```

Under both flag states, the storage layer's read returns an object (the link
record `{"/":"of:..."}`), not the JSON-stringified form. The `typeof value
=== "string"` guard at the defensive parse sites cannot fire.

### Within-transaction reads bypass the storage layer entirely

Inside a single transaction (no commit), `setSourceCell`'s write sits in the
chronicle's `#novelty` map and is returned by reference on subsequent reads
(`runner/src/storage/transaction/chronicle.ts:213-216`). The original object
literal is what `getSourceCell` sees — never a string. So even within a
single transaction, the defensive parse is unreachable through the standard
API.

### Git archaeology

The two defensive parses were introduced when the storage layer's behavior
was different:

- `cell.ts:1460` — added in **PR #1472** (Bernhard, 2025-07-23):
  `fix(transactions): return NotFoundError or (new) TypeMismatchError instead
  of StorageTransactionInconsistent where appropriate`.
- `transaction.ts:280` — added in **PR #1562** (Ellyse, 2025-08-03):
  `convert source to object if its json string`. The commit body comment is
  illuminating:

  > // Special handling for source path, API is to always return object
  > // We should return objects, but we get JSON strings from transaction, so we convert

At that time, the storage boundary did NOT route through any decode helper —
`packages/memory/space.ts` used raw `JSON.parse` calls scattered through the
read path. There was a real flow returning JSON-stringified values.

That changed with **PR #2971** (Cedar, 2026-03-04): `Wire jsonFromValue/
valueFromJson into space.ts storage boundary`. Once `valueFromJson` was
wired in unconditionally, every `is` column read decoded the row. The
defensive parses were orphaned — they were guarding against a string-form
value that no longer arrived.

That orphaning was not noticed at the time because:

1. The defensive parses are downstream of the storage decode and pose no
   correctness issue in their dead state.
2. The standard `setSourceCell` → `getSourceCell` round-trip continues to
   work; the defensive parses are simply not reached.
3. The synthetic test that exercised them (see "Test handling" below) kept
   passing because the test bypasses the storage layer too.

This PR makes the orphaning visible.

### Edge cases the deletion does NOT cover (intentionally)

- **External or migrated legacy data**: if a fact row's `is` column ever
  contained a JSON-string-of-a-JSON-string (i.e., already-encoded twice),
  `valueFromJson` would return the inner JSON string, and the defensive
  parse would have fired. The codebase has no path that produces such rows,
  but this PR's deletion does change observable behavior in that synthetic
  case.
- **Bypass-the-API direct writes**: if a caller writes a JSON-stringified
  value to `path: ["source"]` via `transaction.write` (bypassing
  `setSourceCell`) within a single transaction, the read returns the string
  verbatim post-deletion. The pre-existing test at
  `storage-transaction.test.ts:332` exercised this synthetic path; see
  "Test handling" below.

## Discussion outcomes

The three open questions surfaced by the analysis above were resolved in
human discussion. Dan's verbatim answers:

- **Production database content**: "No concern about pre-existing database content."
- **`v2.ts:378` decode chain**: "No concern about further validation of the `v2.ts` call chain."
- **Un-covered edge cases**: "No concern about un-covered edge cases."

## Post-review revisions to `getSourceCell`

After the conversion to ready, Dan flagged two additional concerns on the
`getSourceCell` shape that this PR's earlier deletions had left visible:

1. The `as string | undefined` cast on `readOrThrow` was a lie — the
   producer audit shows the value is always an object link record, never a
   string.
2. The post-condition throw guard wasn't justified by the surrounding code,
   and the shape would read cleaner with a typed local for the URI form.

The reworked shape addresses both: the cast is dropped (type narrowing
happens via the `isRecord` guard a few lines down, so the type is honest
from `readOrThrow` outward); the `if (isRecord(...))` branch is inverted to
`if (!isRecord(sourceCellId)) { throw ... }` so the happy path narrows
cleanly; a `const sourceCellIdString = toURI(sourceCellId)` is introduced
outside the guard and flows unconditionally into `createCell()`; the
redundant second `toURI` call at the `id:` line (a leftover from the
pre-revision shape) is removed.

Soundness re-verified: re-grepping `path: ["source"]` writers/readers at HEAD
post-#3463-merge shows `setSourceCell` is still the only writer, always
emitting an object literal. The `if (!isRecord(...))` throw only fires on
values production never produces.

## Tests added

Two new top-level `describe` blocks in
`packages/runner/test/cell-core.test.ts`, parameterized over both
`unifiedJsonEncoding` flag states (4 test steps total, all green):

```ts
for (const unifiedJsonEncoding of [false, true]) {
  describe(`Cell source-cell round-trip with \`unifiedJsonEncoding = ${unifiedJsonEncoding}\``, () => {
    it("setSourceCell + commit + getSourceCell round-trips correctly across a fresh tx");
    it("raw read of path: ['source'] returns an object link record (defensive parse guard fails)");
  });
}
```

The tests deliberately exercise the **commit + reload** path (not just within-
tx reads), so `valueFromJson` runs upstream of the (now-deleted) defensive
parse sites. The first asserts end-to-end correctness through the public API;
the second asserts the raw value shape directly, pinning the observable
behavior that makes the defensive parse unreachable.

## Test handling: `storage-transaction.test.ts:332`

The pre-existing test at `packages/runner/test/storage-transaction.test.ts:332`
(`"should write and read the sourceCell via the 'source' path"`, introduced
in PR #1678) was specifically written to lock in the defensive-parse
behavior:

```ts
const setSource = transaction.write(
  { space, id: doc1Id, type: "application/json", path: ["source"] },
  JSON.stringify(getEntityId(doc2Id)),  // writes a STRING
);
// ...read-back expects the parsed object form...
```

It bypasses `setSourceCell` and directly writes a JSON-stringified value
within a single tx. With the defensive parse removed, the test had no
observable behavior to verify (it had been pinning the now-removed
transformation), so it has been **deleted**.

## Files changed

- `packages/runner/src/storage/transaction.ts`: deleted lines 264-298
  (source-path special-case block including the defensive `JSON.parse`);
  deleted unused `toCompactDebugString` import.
- `packages/runner/src/cell.ts`: deleted the `else if` branch at lines
  1457-1461 in `getSourceCell` (the defensive `JSON.parse`); reworked the
  remaining shape per "Post-review revisions to `getSourceCell`" above —
  dropped the misleading `as string | undefined` cast on `readOrThrow`,
  inverted the `isRecord` branch to `if (!isRecord(...)) { throw ... }`,
  introduced the typed local `const sourceCellIdString = toURI(sourceCellId)`,
  and removed the redundant double-`toURI` call at the `id:` line.
- `packages/runner/test/cell-core.test.ts`: added Phase 1 regression tests;
  added `IMemorySpaceAddress` to imports.
- `packages/runner/test/storage-transaction.test.ts`: deleted the synthetic
  pin-test described in "Test handling" above.

## Validation

- `deno fmt --check`: clean
- `deno lint`: clean on edited files
- `cd packages/runner && deno task test`: clean (425 passed, 2697 steps;
  the one pre-existing failing test was the synthetic pin removed in
  Phase 2)

## References

- Researcher Remy's `fvj1:` parse-site kickoff doc (session 2026-067): `coordination/docs/2026-04-30-fvj1-parse-site-kickoff.md`
- Researcher Remy's parent review (§7 — parse-site liveness analysis): `coordination/docs/2026-04-30-json-stringify-parse-memory-runner-review.md`
- Storage-boundary decode wiring: PR #2971 (`Wire jsonFromValue/valueFromJson into space.ts storage boundary`)
- Defensive-parse introductions: PR #1472, PR #1562
- Test pinning: PR #1678


packages/runner/src/cell.ts
danfuzz: Am I correct in reading this cast as a lie? Like, if it's correct then how could `isRecord(sourceCellId)` ever return `true`?

If I'm right, then it's probably a good idea to update this to not mislead the type system.

packages/runner/src/cell.ts
```diff
@@ -1454,15 +1454,8 @@ export class CellImpl<T extends FabricValue>
     if (!sourceCellId) return undefined;
     if (isRecord(sourceCellId)) {
       sourceCellId = toURI(sourceCellId);
-    } else if (
-      typeof sourceCellId === "string" && sourceCellId.startsWith('{"/":')
-    ) {
-      sourceCellId = toURI(JSON.parse(sourceCellId));
     }
 
-    if (typeof sourceCellId !== "string" || !sourceCellId.startsWith("of:")) {
```
danfuzz, line 1463: Is it truly impossible for `sourceCellId` to _not_ start with `of:` now? It's not clear to me from context that this test no longer serves a purpose. Please either reply to this comment about why it's safe, or restore at least the `startsWith` part of this check.

Also, for clarity, maybe add `const sourceCellIdString: string = sourceCellId` before the `createCell()` call, to make it blatant that the type system knows for sure it's a string at that point.

packages/runner/src/cell.ts
danfuzz: Done in [`7edf0933f`](https://github.com/commontoolsinc/labs/commit/7edf0933f). Dropped the cast entirely — the value was never `string` in production (producer audit confirms `setSourceCell` is the only writer and always writes an object link record). The narrow now happens via the `isRecord` guard a few lines down, so the type is honest from `readOrThrow` outward.

— upstream-liaison-bolt (Claude Opus 4.7 (1M context))

packages/runner/src/cell.ts
danfuzz: Done in [`7edf0933f`](https://github.com/commontoolsinc/labs/commit/7edf0933f). Inverted to `if (!isRecord(sourceCellId)) { throw … }` so the type narrows to a record on the happy path, then `const sourceCellIdString = toURI(sourceCellId)` flows unconditionally into `createCell()`. Also removed the redundant second `toURI` call at the `id:` line.

— upstream-liaison-bolt (Claude Opus 4.7 (1M context))

PR #3467, opened 2026-05-01
[codex] Accept CFC writeback compat xattrs

willkelly:
## Summary

This adds the Labs FUSE side of the gVisor CFC writeback transport:

- accepts `user.commonfabric.cfc.writeback.prepare` and `user.commonfabric.cfc.writeback.finalize` as host transport aliases for the trusted CFC writeback xattrs
- normalizes trusted and compatibility xattr names before dispatching prepare/finalize handling
- keeps FUSE permission handling compatible with writeback xattr operation
- documents the compatibility namespace used when `trusted.*` cannot be carried by the host/FUSE layer
- adds a targeted regression test for both trusted and compatibility writeback xattr names

## Root Cause

The sandbox-facing API uses `trusted.cfc.writeback.*`, but host FUSE transport may need to carry those requests as `user.commonfabric.cfc.writeback.*`. Without accepting the compatibility spelling, Labs FUSE could publish content labels but could not complete the explicit writeback prepare/finalize protocol through gVisor.

## Namespace Model

`trusted.cfc.*` is the sandbox-facing CFC control namespace. gVisor treats these xattrs as runtime/security metadata: content labels can taint tasks, direct unprivileged writes to content labels are blocked, and the writeback protocol is handled through specific CFC rules.

`user.commonfabric.cfc.*` is only the host/FUSE transport spelling. Linux and FUSE often cannot carry `trusted.*` xattrs through an unprivileged userspace daemon, so Labs FUSE exposes the same CFC metadata under a `user.*` compatibility namespace for the host side.

Keeping the sandbox API on `trusted.cfc.*` matters because `user.*` xattrs are ordinary application metadata. A tainted task that can freely set `user.*` xattrs could encode secrets into metadata on a public file, so gVisor treats `user.*` writes as a storage channel. The compatibility `user.commonfabric.cfc.*` names should therefore stay below the gofer/FUSE boundary as transport details, not become the guest API.

## Validation

Fresh worktree checks:

- `git diff --cached --check`
- `deno fmt --check packages/fuse/cfc-writeback.ts packages/fuse/mod.ts packages/fuse/cfc-writeback.test.ts packages/fuse/README.md docs/specs/fuse-filesystem/9-cfc-annotations.md`
- `deno test --allow-env --allow-read --allow-write packages/fuse/cfc-writeback.test.ts packages/fuse/mod.test.ts`

Earlier end-to-end validation with the paired gVisor patch:

- `LABS_DIR=/tmp/labs-pr-3445 TOOLSHED_URL=http://localhost:8000 bash demo/docker-runsc-cfc-labs-fuse-validate.sh`

The end-to-end harness proved content-label publication, sandbox xattr label reads and tainting, and writeback prepare/finalize cleanup.

### DIFF
--- a/docs/specs/fuse-filesystem/9-cfc-annotations.md
+++ b/docs/specs/fuse-filesystem/9-cfc-annotations.md
@@ -44,6 +44,11 @@ symlinks, truncates, and metadata mutations while recording diagnostics.
 Enforce-* modes require the prepare/finalize flow before those operations
 proceed. The existing non-CFC behavior is unchanged.
 
+The temporary writeback xattr bridge accepts the trusted prepare/finalize names
+and the equivalent `user.commonfabric.cfc.*` compatibility spellings when a host
+transport cannot carry `trusted.*` xattrs. gVisor remains responsible for
+exposing only the trusted logical namespace to the sandbox.
+
 Derived slot metadata is emitted as an explicit empty `no-trusted-derived-slots`
 structure. Digest-looking names therefore remain ordinary names unless a future
 trusted-derived identifier source supplies evidence.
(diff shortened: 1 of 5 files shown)


seefeldb (approved):
lgtm!

PR #3472, opened 2026-05-01
Add reusable cf CLI launcher

mathpirate:
## Summary
- add a Labs-owned Deno launcher for the Common Fabric CLI that can be invoked from sibling or vendored consumers before the Labs import map is active
- switch root `deno task cf` and package-local `deno task cli` to the launcher
- document launcher usage for Labs, Pattern Factory, and Loom layouts

## Verification
- `deno fmt --check packages/cli/README.md packages/cli/launcher.ts packages/cli/test/launcher.test.ts deno.json packages/cli/deno.json`
- `deno test --allow-run --allow-env --allow-read packages/cli/test/launcher.test.ts`
- `deno task cf --help`
- `deno task cf --launcher-help`
- `deno task cf check packages/cli/fixtures/pattern.tsx --no-run`
- `deno task --cwd packages/cli cli --help`
- `deno task check`
- external Pattern Factory cwd launcher smoke for `--help` and `cf check`

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -39,7 +39,7 @@
   ],
   "tasks": {
     "check": "./tasks/check.sh",
-    "cf": "bash -lc 'ROOT=\"$(pwd)\" && cd \"$INIT_CWD\" && CF_CLI_NAME=cf deno run --allow-net --allow-ffi --allow-read --allow-write --allow-env --allow-run \"$ROOT/packages/cli/mod.ts\" \"$@\"' --",
+    "cf": "deno run --allow-run --allow-env --allow-read ./packages/cli/launcher.ts",
     "test": "./tasks/test.ts",
     "test-all": "echo \"Use 'deno task test' instead.\" && exit 1",
     "build-binaries": "./tasks/build-binaries.ts",
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -7,8 +7,27 @@ to run the Common Fabric CLI from another repo or from a sandbox. It keeps the
 selected Labs checkout as the source of truth while making the child CLI process
 use an explicit Deno config/import map.
 
-The launcher itself intentionally uses only Deno and Node built-ins so callers
-can invoke it before the Labs import map is active.
+The launcher itself intentionally uses only Deno built-ins so callers can invoke
+it before the Labs import map is active.
+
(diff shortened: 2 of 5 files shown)


willkelly (commented):
This looks like the right shape overall: the launcher stays a thin process/integration boundary for choosing the Deno binary, Labs root/config/entrypoint, caller cwd, and env/stdio/exit forwarding, while `cf`/cf-harness/consumers continue to own actual command semantics, pattern roots/imports, identity/API/logging behavior, and CFC/gVisor policy.

I don't see a merge-blocking implementation issue in the current head. I verified the launcher unit tests, formatting, root/package `--help`, an explicit `--labs-root` invocation, `cf check ... --no-run`, LSP diagnostics, and Deno child env inheritance.

A few things I'd like tightened before treating this as the stable Loom / Pattern Factory / sandbox contract:

- Please catch launcher parse/spawn failures and render launcher-scoped diagnostics. For example, `deno task cf --config` currently ends up as a raw stack from the launcher path; external/sandbox users would be better served by something like `cf launcher: --config requires a value; use -- to pass --config to cf`.
- Please document and/or test the `INIT_CWD` vs `--cwd` behavior for direct/sandbox invocation. Preserving `INIT_CWD` is correct for `deno task cf`, but if an outer task/sandbox wrapper leaks a stale `INIT_CWD`, `cf check`/`cf test` will resolve user paths from the wrong cwd unless callers explicitly pass `--cwd`.
- Please clarify that launcher `--config` means the child Deno config/import-map used to start the CLI, not a `cf` command/pattern config. The Loom example using `--config deno.json` is useful, but it would be good to pair it with a consumer-config smoke test outside Labs if that is part of the promised contract.
- Please make env inheritance an intentional part of the contract, either in docs or a small test. `CF_API_URL`, `CF_IDENTITY`, experimental flags, and CFC/sandbox-related envs need to flow through; the current `Deno.Command` behavior does this, but the launcher only makes `CF_CLI_NAME` explicit.
- Nit: `packages/cli/README.md` says the launcher uses “Deno and Node built-ins,” but the final launcher no longer imports Node built-ins.

Also, I think the broad child Deno permissions are okay here specifically because this launcher is not the sandbox boundary; gVisor/cf-harness should remain responsible for policy enforcement.

PR #3488, opened 2026-05-04
feat(shell/runner/toolshed): Add blob upload and serving

seefeldb:
## Summary

This PR adds a minimal blob upload and serving path to Toolshed using the temporary `cid:` convention we discussed. It lets shell-side code hand binary data to the runner thread, have the runner store it in the current space, and receive a relative URL that can be used directly in rendered HTML, for example `.png">`.

The implementation intentionally keeps this as a `cid:fid1:<hash>` convention for now. It does not try to introduce a full content-addressed storage abstraction yet, but it uses the existing memory/data-model primitives so the stored payload has the expected hash identity.

## Security / Auth

Blob routes are intentionally unauthenticated for this MVP. Anyone can POST a blob into any space DID; because the payload is content-addressed they cannot overwrite a different payload, but they can inject blobs and consume storage. Anyone with a blob hash can GET that blob. Full UCAN-gated access is a required follow-up before any production exposure.

### Toolshed blob routes

Adds blob routes under a space DID:

- `POST /:spaceDid/blobs`
- `POST /:spaceDid/blobs/:blobName`
- `GET /:spaceDid/blobs/:blobName`

The POST endpoint accepts the same memory-boundary shape used by the runner upload path:

```ts
const contents = {
  type: "image/png", // MIME type
  body: new FabricBytes(rawImageData),
};
```

Toolshed computes `hashOf(contents).toString()`, stores the contents in memory as `cid:<fid1:...>` with the normal cell document shape:

```ts
{ value: contents }
```

and returns:

```ts
{
  id: "fid1:...",
  url: "blobs/<hash>.<suffix>"
}
```

The GET endpoint resolves `/:spaceDid/blobs/<hash>.<suffix>` to `cid:fid1:<hash>`, ignores the suffix for lookup, reads the stored `{ type, body }`, and serves the raw bytes with the embedded MIME type as `Content-Type`.

### Memory server helpers

Adds direct `readDocument` and `writeDocument` helpers to the memory v2 server so Toolshed can read/write the exact `cid:` document without going through websocket/session plumbing. These helpers support the unified JSON encoding path needed to preserve `FabricBytes` correctly.

Direct writes now go through the normal dirty-refresh path even when there are no live websocket connections, matching the rest of the memory server's write behavior.

### Runtime IPC upload command

Adds a new runtime-client IPC command:

```ts
RequestType.UploadBlob = "runtime:uploadBlob"
```

Request shape:

```ts
{
  type: RequestType.UploadBlob,
  contentType: string,
  body: Uint8Array,
  suffix?: string,
}
```

Response shape:

```ts
{
  id: string,
  url: string,
}
```

The main-thread caller sends bytes through `RuntimeClient.uploadBlob()`. The runtime processor handles the IPC request on the runner thread, wraps the bytes as `new FabricBytes(body)`, posts the encoded payload to Toolshed, and returns the relative `blobs/...` URL to the caller.

This keeps the actual persistence work on the runner side while still giving UI code a simple async method.

### Shell relative blob URLs

Shell now manages a space-scoped base href:

```html
<base data-commonfabric-space-base="true" href="/<spaceDid>/">
```

That means shell-rendered or component-rendered relative URLs like this:

```html
<img src="blobs/<hash>.png">
```

resolve to:

```text
/<spaceDid>/blobs/<hash>.png
```

even when the visible shell URL was opened through a space name rather than the DID.

## How To Invoke From Shell-Land Code

Shell exposes the runtime client on `globalThis.commonfabric.rt`. Any shell-side web component or UI controller that has access to the window can call `uploadBlob()` and then use the returned relative URL directly.

Example web component handler:

```ts
type BlobUploadRuntime = {
  uploadBlob(options: {
    contentType: string;
    body: Uint8Array;
    suffix?: string;
  }): Promise<{ id: string; url: string }>;
};

type CommonfabricWindow = Window & typeof globalThis & {
  commonfabric?: {
    rt?: BlobUploadRuntime;
  };
};

function fileSuffix(name: string): string | undefined {
  const suffix = name.split(".").pop();
  return suffix && suffix !== name ? suffix : undefined;
}

class ImageUploader extends HTMLElement {
  async uploadFile(file: File): Promise<void> {
    const rt = (globalThis as CommonfabricWindow).commonfabric?.rt;
    if (!rt) {
      throw new Error("Common Fabric runtime is not ready");
    }

    const upload = await rt.uploadBlob({
      contentType: file.type || "application/octet-stream",
      suffix: fileSuffix(file.name),
      body: new Uint8Array(await file.arrayBuffer()),
    });

    const img = document.createElement("img");
    img.src = upload.url;
    img.alt = file.name;
    this.replaceChildren(img);

    // upload.id is the fid1 content id.
    // upload.url is relative, e.g. "blobs/<hash>.png".
    console.log(upload.id, upload.url);
  }
}
```

For shell code that already holds a `RuntimeClient` instance, the call is the same without going through `globalThis`:

```ts
const upload = await runtimeClient.uploadBlob({
  contentType: "image/png",
  suffix: "png",
  body: pngBytes,
});
```

The returned `upload.url` is intended to be stored in pattern/shell state or assigned directly to DOM attributes. Because shell installs the space DID base href, callers should keep it relative rather than expanding it themselves.

## End-To-End Flow

1. A shell web component receives a `File`, `Blob`, canvas export, pasted image, or other binary source on the main thread.
2. The component converts it to `Uint8Array` and calls `commonfabric.rt.uploadBlob(...)`.
3. Runtime client sends the `runtime:uploadBlob` IPC request to the runner worker.
4. Runtime processor wraps the bytes as `{ type, body: new FabricBytes(...) }` and POSTs the encoded body to Toolshed.
5. Toolshed computes the content hash, writes `cid:<fid1:hash>` in the current space, and returns a relative `blobs/<hash>.<suffix>` URL.
6. The component can render that URL immediately, for example with ``.
7. Browser resolution uses shell's base href, so `blobs/...` loads from `/<spaceDid>/blobs/...`.
8. Toolshed serves the raw bytes with the embedded MIME type.

## Tests Added

- Toolshed route test verifies POST creates the blob cell and GET serves the stored bytes with the embedded MIME type while ignoring the suffix.
- Toolshed route test verifies the blob cell can be read through the `Cell` interface as `{ type, body }` from the same `cid:...` id.
- Memory v2 server test verifies direct `writeDocument()` calls schedule the dirty-refresh path even when there are no live connections.
- Runtime processor test verifies the IPC handler encodes bytes and posts to the space-scoped blob route.
- Shell integration test verifies a browser can call `commonfabric.rt.uploadBlob()`, append an `` using the returned relative `blobs/...` URL, and load the image through the DID-scoped blob endpoint.

## Validation

- `deno task check`
- `deno task test`
- `deno task integration`
- `deno task test` in `packages/memory`
- `ENV=test deno test -A --env-file=packages/toolshed/.env.test packages/toolshed/routes/blobs/blobs.test.ts`

### DIFF
--- a/packages/runtime-client/backends/runtime-processor.ts
+++ b/packages/runtime-client/backends/runtime-processor.ts
@@ -1,4 +1,16 @@
 import { DID, Identity, type Session } from "@commonfabric/identity";
+import {
+  getDataModelConfig,
+  resetDataModelConfig,
+  setDataModelConfig,
+} from "@commonfabric/data-model/fabric-value";
+import { FabricBytes } from "@commonfabric/data-model/fabric-bytes";
+import {
+  getJsonEncodingConfig,
+  resetJsonEncodingConfig,
+  setJsonEncodingConfig,
+} from "@commonfabric/data-model/json-encoding";
+import { encodeMemoryBoundary } from "@commonfabric/memory/v2";
 import { PieceManager } from "@commonfabric/piece";
 import { PiecesController } from "@commonfabric/piece/ops";
 import {
@@ -89,6 +101,8 @@ import {
   type SetTriggerTraceEnabledRequest,
   type SetWriteStackTraceMatchersRequest,
   type TriggerTraceResponse,
+  type UploadBlobRequest,
+  type UploadBlobResponse,
   type VDomEventRequest,
   type VDomMountRequest,
   type VDomMountResponse,
@@ -110,6 +124,27 @@ import type { VDomOp } from "../protocol/types.ts";
 import type { RuntimeOptions } from "@commonfabric/runner";
 
 const MAX_SERIALIZATION_DEPTH = 5;
+
+const withUnifiedEncoding = <T>(fn: () => T): T => {
+  const previousDataModel = getDataModelConfig();
+  const previousJson = getJsonEncodingConfig();
+  setDataModelConfig(true);
+  setJsonEncodingConfig(true);
+  try {
+    return fn();
+  } finally {
+    if (previousDataModel) {
+      setDataModelConfig(true);
+    } else {
+      resetDataModelConfig();
+    }
+    if (previousJson) {
+      setJsonEncodingConfig(true);
+    } else {
+      resetJsonEncodingConfig();
+    }
+  }
+};
 
 export function runtimeOptionsFromInitializationData(
   data: InitializationData,
@@ -254,6 +289,7 @@ export class RuntimeProcessor {
   private runtime: Runtime;
   private pieceManager: PieceManager;
   private cc: PiecesController;
+  private apiUrl: URL;
   private space: DID;
   private identity: Identity;
   private _isDisposed = false;
@@ -273,6 +309,7 @@ export class RuntimeProcessor {
     runtime: Runtime,
     pieceManager: PieceManager,
     cc: PiecesController,
+    apiUrl: URL,
     space: DID,
     identity: Identity,
     telemetry: RuntimeTelemetry,
@@ -280,6 +317,7 @@ export class RuntimeProcessor {
     this.runtime = runtime;
     this.pieceManager = pieceManager;
     this.cc = cc;
+    this.apiUrl = apiUrl;
     this.space = space;
     this.identity = identity;
     this.telemetry = telemetry;
@@ -392,6 +430,7 @@ export class RuntimeProcessor {
       runtime,
       pieceManager,
       cc,
+      apiUrlObj,
       space,
       identity,
       telemetry,
@@ -824,6 +863,38 @@ export class RuntimeProcessor {
   setBreakpoints(request: SetBreakpointsRequest): void {
     this.runtime.scheduler.setBreakpoints(request.actionIds);
   }
+
+  async handleUploadBlob(
+    request: UploadBlobRequest,
+  ): Promise<UploadBlobResponse> {
+    const suffix = (request.suffix ?? "bin").replace(/^\./, "") || "bin";
+    const target = new URL(
+      `/${this.space}/blobs/upload.${encodeURIComponent(suffix)}`,
+      this.apiUrl,
+    );
+    const body = withUnifiedEncoding(() =>
+      encodeMemoryBoundary({
+        type: request.contentType,
+        body: new FabricBytes(request.body),
+      })
+    );
+    const response = await fetch(target, {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body,
+    });
+    if (!response.ok) {
+      throw new Error(
+        `Blob upload failed: ${response.status} ${await response.text()}`,
+      );
+    }
+    const result = await response.json() as Partial<UploadBlobResponse>;
+    if (typeof result.id !== "string" || typeof result.url !== "string") {
+      throw new Error("Blob upload returned an invalid response");
+    }
+    return { id: result.id, url: result.url };
+  }
+
   async detectNonIdempotent(
     request: DetectNonIdempotentRequest,
   ): Promise<DetectNonIdempotentResponse> {
@@ -987,6 +1058,8 @@ export class RuntimeProcessor {
         return this.getPatternSources(request);
       case RequestType.SetBreakpoints:
         return this.setBreakpoints(request);
+      case RequestType.UploadBlob:
+        return await this.handleUploadBlob(request);
       case RequestType.VDomEvent:
         return this.handleVDomEvent(request);
       case RequestType.VDomMount:
--- a/packages/runtime-client/backends/runtime-processor.ts
+++ b/packages/runtime-client/backends/runtime-processor.ts
@@ -1,15 +1,5 @@
 import { DID, Identity, type Session } from "@commonfabric/identity";
-import {
-  getDataModelConfig,
-  resetDataModelConfig,
-  setDataModelConfig,
-} from "@commonfabric/data-model/fabric-value";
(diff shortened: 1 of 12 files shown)


seefeldb:
Follow-up changes for the review feedback:

- Item 1: documented the intentional auth deferral in `packages/toolshed/routes/blobs/blobs.index.ts`, including that POST is currently allowed into any space DID, content addressing prevents overwriting different payloads but still allows blob injection/storage consumption, GET is available to anyone with the hash, and this must be revisited before production exposure. I also added the same `Security / Auth` section to the PR description.
- Item 2: removed the no-connection early return from `markSpaceDirty`, so direct writes go through the normal dirty-refresh path. Added a memory v2 server unit test that opens a server with no live connections, calls `writeDocument()`, advances the scheduled refresh timer, and asserts the flush path runs and the document is stored as `{ value: ... }`.
- Item 3: intentionally left the `upload.<suffix>` POST URL shape unchanged.
- Item 4: intentionally left `Cache-Control` and suffix-in-URL cache behavior unchanged.

Validation run:

- `deno task test` in `packages/memory`
- `ENV=test deno test -A --env-file=packages/toolshed/.env.test packages/toolshed/routes/blobs/blobs.test.ts`
- `deno task check`

seefeldb:
Follow-up for the latest review feedback:

1. Removed dynamic JSON/data-model config toggling from non-test code in `packages/memory/v2/server.ts`, `packages/runtime-client/backends/runtime-processor.ts`, and `packages/toolshed/routes/blobs/blobs.index.ts`. `readDocument`/`writeDocument`, `handleUploadBlob`, and blob request decode now use the process-configured encoding mode directly. Test-only toggling remains only where load-bearing for constructing/decoding `FabricBytes` payloads in unconfigured package test processes.
2. Deleted the dead `encodeBlobContents` export from `blobs.index.ts` and removed the now-unused `encodeMemoryBoundary` import there.
3. Replaced the blob upload handler's `any` context with Hono's `Context` type.
4. Tightened GET blob-name parsing so empty/all-dot hash segments return `400 Invalid blob name`, while syntactically valid missing blobs still return 404. Added route coverage for the 400 case.
5. Added suffix validation at the Toolshed boundary: only `[A-Za-z0-9]{1,8}` is accepted from the URL suffix; invalid suffixes fall back to the MIME-derived default. Added route coverage for fallback behavior.
6. Added a comment on the memory server's direct write session/sequence fields explaining they are synthesized state for out-of-band writes such as blob uploads.
7. Simplified the runtime-processor blob upload test now that product code no longer toggles encoding, and added a comment explaining the narrow `.call(...)` fake because the constructor performs full runtime initialization and the handler only reads `apiUrl` and `space`.
8. Added a direct memory server helper test for `writeDocument`/`readDocument`, covering `FabricBytes` round-trip under test-configured unified encoding and missing-id `null`.

Validation run:

- `deno task test` in `packages/memory`
- `deno task test` in `packages/runtime-client`
- `deno task test` in `packages/toolshed`
- `deno task check`

seefeldb:
Addressed the remaining applicable PR review comments:

- Auth thread: left unchanged intentionally. The blob routes remain unauthenticated for this MVP per the existing source comment and PR `Security / auth` section; full UCAN-gated access stays deferred.
- cubic-identified upload-size thread: added a 10 MiB blob upload cap before decoding, using `Content-Length` when present and a streaming body reader when it is not. Oversized payloads now return 413, with a regression test.
- cubic/Codex MIME threads: blob GET now allowlists safe image response types and falls back to `application/octet-stream` for unsafe uploaded MIME types such as `text/html`, with a regression test. `Cache-Control` was left unchanged.

Validation run after the changes:
- `ENV=test deno test -A --env-file=packages/toolshed/.env.test packages/toolshed/routes/blobs/blobs.test.ts`
- `deno task test` in `packages/toolshed`
- `deno task check`

seefeldb:
Follow-up on the latest cubic-identified runtime-client review:

- Test encoding setup: addressed in `105f4db`. The focused runtime-processor blob upload IPC test now explicitly enables unified JSON/data-model encoding around the upload call, so the `FabricBytes` round trip is asserted under the same ambient mode used by configured runtimes.
- Product upload encoding: intentionally left unchanged. Per the prior review direction, non-test code must not flip `@commonfabric/data-model` process-wide encoding flags per call. `handleUploadBlob` continues to call `encodeMemoryBoundary({ type, body: new FabricBytes(...) })` directly and relies on the runtime process startup configuration established by `Runtime` construction.

Validation run after the test change:
- `deno task test` in `packages/runtime-client`
- `deno task check`

packages/runtime-client/backends/runtime-processor.ts
```diff
@@ -1,4 +1,7 @@
 import { DID, Identity, type Session } from "@commonfabric/identity";
+import { FabricBytes } from "@commonfabric/data-model/fabric-bytes";
+import type { FabricValue } from "@commonfabric/data-model/fabric-value";
+import { JsonEncodingContext } from "@commonfabric/data-model/json-encoding-context";
```
danfuzz, line 4: Not a big deal, but `JsonEncodingContext` is basically a singleton right now, and instead of creating them here (and I spotted at least one other place), you can just import the functions from `json-encoding-modern` and use those, which defer to `data-model`'s sole instance of it.

No need to do anything about this; I'll probably end up rooting it out in a cleanup pass after the JSON flag graduates.

danfuzz (commented):
Ex-post-facto LGTM!

PR #3496, opened 2026-05-05
Add cf-harness image attachments

mathpirate:
## Summary
- add cf-harness image attachment metadata and repeatable `--image` CLI parsing for png/jpeg/gif/webp inputs
- materialize image bytes as OpenAI-compatible `image_url` data URLs only while building gateway requests
- keep transcripts/run-state to image metadata and add workspace/realpath/symlink bounds plus digest verification

## Tests
- `deno task test` in `packages/cf-harness` (`248 passed`)

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -130,7 +130,7 @@
     "core-js/proposals/async-explicit-resource-management": "https://esm.sh/core-js@3.46.0/proposals/async-explicit-resource-management",
     "@commonfabric/ports": "./ports.json",
     "@astral/astral": "./packages/vendor-astral/mod.ts",
-    "@cliffy/command": "jsr:@cliffy/command@^1.0.0-rc.8",
+    "@cliffy/command": "jsr:@cliffy/command@1.0.0-rc.8",
     "@noble/hashes": "npm:@noble/hashes@^2.2.0",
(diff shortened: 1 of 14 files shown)


willkelly (approved):
Looks good to me. The image attachment path is well-contained: it keeps base64 pixels out of persisted transcripts, does realpath-based workspace/symlink checks, and materializes the image data only at the gateway boundary. The added tests cover the important CLI/path/prompt-loop behavior, and current CI is green.

On the Cliffy pin: I think pinning to `1.0.0-rc.8` is reasonable for this PR. The earlier failure was the compiled CLI integration hitting Cliffy argument-parsing behavior (`Missing argument(s): callable`), not the image attachment code itself. The underlying trigger is that our binary build path temporarily removes `deno.lock` before `deno compile`, so a caret Cliffy range can resolve differently during binary builds than it does in normal locked development/test flows.

Follow-up request, not a blocker for this PR: let's open a follow-up PR that updates Deno / the binary build path so we no longer need to bypass `deno.lock`, then remove this exact Cliffy pin in that PR. In other words, this pin is a good tactical stabilization, but the healthier fix is to make the binary build respect a locked dependency graph again.

PR #3500, opened 2026-05-05
feat(scopes): implement scoped cell instances

seefeldb:
## Summary
- implement scoped cell instance plumbing across memory v2, runner links, schemas, traversal, runtime cells, and computation output writes
- add scope-aware authoring/schema surfaces for `PerSpace`, `PerUser`, `PerSession`, `PerAny`, schema `scope`, scoped `asCell` entries, sigil link scope, and resolved `NormalizedFullLink.scope`
- add runtime support for factory `.asScope(...)`, default scoped pattern/process cells, scoped read tracking, broad-output-to-narrow-result links, scoped scheduler/storage keys, and scoped source/attestation reads
- apply built-in scope policies for `navigateTo`, `map`, `filter`, `flatMap`, `ifElse`, `when`, `unless`, async fetch/stream/compile state cells, LLM result cells, and `wish`
- infer assignment-context factory scope in the TS transformer and emit `.asScope(...)` for scoped pattern factory calls
- fix persisted pattern rehydration so a fresh runtime syncs the stored process cell `$TYPE` and explicitly reloads linked pattern source metadata when source traversal previously loaded only the metadata root
- isolate selector subscription caches by declared scope, pass declared cell scope through memory-v2 sync roots, and preserve same-id snapshots in different scopes in the memory-v2 client watch view
- add and verify a scoped group chat pattern using user-scoped name/draft, session-scoped room selection, and a space-scoped conversation

## Implementation Notes
- Causal id/cause computation remains scope-insensitive; scoped equality and storage addressing include scope.
- Memory v2 is the scoped storage path. Local v2 replicas, transaction validation, watches, sync keys, confirmed/pending reads, native commit payloads, reactivity logs, and client transient sigil links now carry declared scope.
- Schema reads with omitted scope or `scope: "any"` may follow narrower links; explicit scoped schemas block narrower links and return `undefined`, with info-level logging.
- Broad computations that read narrower scoped data write the actual value to the narrower scoped instance and place a scoped link in the broader output location.
- List built-ins preserve the map outer-list cardinality rule while narrowing per-element result links; filter/flatMap narrow their output list because cardinality can vary by scoped input.
- `map` now preserves the declared scope of selected linked lists, carries item result schema into CFC flow precision metadata, and rewrites relative data-cell links with the containing link scope.
- `wish` uses explicit result schema scope when provided; otherwise home-space resolution is at least user scoped and current/arbitrary-space queries derive scope from the query inputs.


packages/api/index.ts
```diff
@@ -184,6 +184,26 @@ export type CellKind =
...
 
+export type CellScope = "space" | "user" | "session";
+export type SchemaScope = CellScope | "any";
+export type LinkScope = "inherit" | CellScope;
+
+export type AsCellEntry =
+  | CellKind
+  | {
+    readonly kind: CellKind;
+    readonly scope?: LinkScope | "any";
```
seefeldb: this should just be SchemaScope, and if "inherit" is needed, then let's add that to SchemaScope.

packages/patterns/scoped-group-chat/main.tsx
```diff
@@ -0,0 +1,508 @@
...
+  const rooms = (conversation.get() as Conversation | undefined)?.rooms ?? [];
+  if (rooms.length === 0) return;
+
+  const roomRef = room ?? selectedRoom.key("room");
+  const targetIndex = rooms.findIndex((_candidate, index) =>
+    conversation.key("rooms", index).equals(roomRef)
+  );
+  if (targetIndex < 0) return;
+
+  (conversation.key(
```
seefeldb: why go around doing this if you could just write `roomRef.key("messages").push`

packages/patterns/scoped-group-chat/main.tsx
```diff
@@ -0,0 +1,508 @@
...
+  selectedRoom: SelectedRoomCell;
+  newRoomName: NewRoomNameCell;
+}>(({ name: eventName }, { conversation, selectedRoom, newRoomName }) => {
+  const name = (eventName ?? newRoomName.get()).trim();
+  if (!name) return;
+
+  const rooms = (conversation.get() as Conversation | undefined)?.rooms ?? [];
+  const nextIndex = rooms.length;
+  conversation.key("rooms").push({ name, messages: [] });
+  selectedRoom.set({ room: conversation.key("rooms", nextIndex) });
```
seefeldb: selectedRoom.set({ room: conversation.key("rooms", conversations.key("room").get().length - 1) });

should work and will avoid a .get() over the full conversation.

packages/patterns/scoped-group-chat/main.tsx
```diff
@@ -0,0 +1,508 @@
...
+    }));
+    const selectedRoomRef = selectedRoom.key("room");
+    const selectedRoomIndex = computed(() => {
+      const index = rooms.findIndex((_room, roomIndex) =>
+        conversation.key("rooms", roomIndex).equals(selectedRoomRef)
+      );
+      return index < 0 ? 0 : index;
+    });
+    const messagesInSelectedRoom = computed<readonly ChatMessage[]>(() =>
+      rooms[selectedRoomIndex]?.messages ?? []
```
seefeldb: use selectedRoomRef here directly, also for the next few

packages/patterns/scoped-group-chat/main.tsx
```diff
@@ -0,0 +1,508 @@
...
+        .map((room) => `${room.name}: ${room.messages?.length ?? 0}`)
+        .join("\n")
+    );
+    const roomCount = computed(() => rooms.length);
+    const currentRoomMessageCount = messageCount;
+    const lastRoom = computed(() => rooms[rooms.length - 1] ?? EMPTY_ROOM);
+    const lastRoomName = computed(() => lastRoom.name);
+    const lastRoomMessageCount = computed(() => lastRoom.messages?.length ?? 0);
+    const lastCurrentRoomAuthor = computed(() => {
+      const messages = rooms[selectedRoomIndex]?.messages ?? [];
```
seefeldb: here as well, just use selectedRoomRef, no need to use the index.

packages/patterns/scoped-group-chat/main.tsx
```diff
@@ -0,0 +1,508 @@
...
+                  <cf-select
+                    $value={currentRoomIndex}
+                    items={roomItems}
+                    aria-label="Room"
+                    onChange={boundSelectRoom}
+                  />
+                </cf-vstack>
+                <cf-vstack gap="1" style="width: 280px;">
+                  <cf-label>New room</cf-label>
+                  <cf-message-input
```
seefeldb: see cf-form, maybe that works better here, so you can say "Add" in the button.

otherwise, split into two, bind name input to a session scoped draft name.

packages/patterns/scoped-group-chat/main.tsx
```diff
@@ -0,0 +1,508 @@
...
+  roomSummaryText: string;
+  sendMessage: Stream<SendMessageEvent>;
+  addRoom: Stream<AddRoomEvent>;
+  setName: Stream<{ name: string }>;
+  setDraft: Stream<{ draft: string }>;
+  setNewRoomName: Stream<{ name: string }>;
+  selectRoom: Stream<SelectRoomEvent>;
+}
+
+const setName = handler<{ name: string }, { name: Writable<string> }>(
```
seefeldb: these are just used for tests and could easily be inline `action` calls on the pattern return value. a lot less to type.

packages/patterns/scoped-group-chat/main.tsx
```diff
@@ -0,0 +1,508 @@
...
+    const send = sendMessage({
+      name,
+      selectedRoom,
+      conversation,
+      draft,
+    });
+    const roomItems = computed(() =>
+      rooms.map((room, index) => ({
+        label: room.name,
+        value: String(index),
```
seefeldb: i'm pretty sure cf-select will work with references here, saving us going via indices (and preventing races as well)

packages/patterns/scoped-group-chat/main.tsx
```diff
@@ -0,0 +1,508 @@
...
+                      if (roomName) boundAddRoom.send({ name: roomName });
+                    }}
+                  />
+                </cf-vstack>
+              </cf-hstack>
+            </cf-vstack>
+
+            <cf-vscroll flex showScrollbar fadeEdges>
+              <cf-vstack gap="3" padding="4">
+                <cf-hstack gap="3">
```
seefeldb: this could be cf-tab instead and we could then make this the selector for the current room.

packages/ts-transformers/test/fixtures/schema-injection/scoped-factory-assignment.expected.jsx
```diff
@@ -0,0 +1,86 @@
...
+        }
+    },
+    required: ["label"]
+} as const satisfies __cfHelpers.JSONSchema, {
+    type: "object",
+    properties: {
+        userChild: {
+            scope: "user"
+        },
+        sessionChild: {
```
ubik2, line 65: Shouldn't these have the `$ref: "#/$defs/ChildOutput"`

packages/memory/v2/query.ts
```diff
@@ -238,7 +264,21 @@ const entitiesFromTracker = (
   branch: string,
 ): Map<QueryDocKey, EntitySnapshot> => {
   const entities = new Map<QueryDocKey, EntitySnapshot>();
+  const explicitEntityKeys = new Set<string>();
```
ubik2: Could be a little more efficient by making a tracker.map pass first to avoid double calls of `fromDocKey`. Bit of a hassle, since we still want `key` too (or rebuild from parsed).

ubik2: I don't think this is exactly defined in the spec, but you could generate cells with the same id and different scopes. These would be skipped by this loop (since we find the explicitEntityKey for the version in another scope).
OTOH, it seems like a reasonable restriction that you can't re-use the same entity id for objects in different scopes (e.g. you can't have one in user and one in session).
I'd probably alter the explicitEntityKeys to hold key instead, since that lets us support this if all the other parts do too.

packages/memory/v2/query.ts
```diff
@@ -634,14 +745,62 @@ export const toDocKey = (
...
+} => {
+  const match = /^([^/]+)\/(space|user|session)\/(.+)\/application\/json$/
+    .exec(key);
   if (match !== null) {
-    const [, space, id] = match;
-    return { space, id, type: "application/json" };
+    const [, space, scope, id] = match;
+    return { space, scope: scope as CellScope, id, type: "application/json" };
+  }
+  const [space, maybeScope, ...rest] = key.split("/");
```
ubik2, line 754: Can we constrain the doc key to always be in the new format?
As far as I can tell, we're just using it in memory, so we don't need to support the old format.

The same is likely true for fromDiryKey.

packages/memory/v2/query.ts
```diff
@@ -434,9 +504,23 @@ export const refreshTrackedGraph = (
...
+    const { id, scope } = fromDirtyKey(dirtyId);
+    let scopedIds = invalidations.get(scope);
+    if (scopedIds === undefined) {
+      scopedIds = new Set();
+      invalidations.set(scope, scopedIds);
+    }
+    scopedIds.add(id);
+    let key = toDocKey(space, id, "application/json", scope);
+    let selectors = state.tracker.get(key);
+    if (
```
ubik2: I mention this later, but I don't think we need to support both versions of doc keys in our tracker.
The updated code should just write the new form, and expect to read the new form from memory.

packages/memory/v2/query.ts
```diff
@@ -471,9 +560,23 @@ export const refreshTrackedGraph = (
...
+    let key = toDocKey(
+      space,
+      address.id,
+      "application/json",
+      address.scope,
+    );
+    if (
+      !state.tracker.has(key) && address.scope === DEFAULT_SCOPE
+    ) {
+      const legacyKey =
```
ubik2: as above -- no need for legacy keys

packages/runner/src/scope.ts
```diff
@@ -0,0 +1,63 @@
...
+): CellScope {
+  let narrowest: CellScope = DEFAULT_CELL_SCOPE;
+  for (const scope of scopes) {
+    if (scope !== undefined && scopeRank(scope) > scopeRank(narrowest)) {
+      narrowest = scope;
+    }
+  }
+  return narrowest;
+}
+
```
ubik2, line 54: might be nice to have a doc comment about how we can follow a link if the link scope is at least as broad as the schemaScope or the schemaScope is permissive.

packages/runner/src/scheduler.ts
```diff
@@ -267,9 +271,15 @@ export {
   markReadAsPotentialWrite,
 };
 
-export type SpaceAndURI = `${MemorySpace}/${URI}`;
+export type SpaceAndURI = `${MemorySpace}/${URI}/${string}`;
```
ubik2: Perhaps this type name should be changed to indicate that it contains scope as well. `SpaceURIScope`?
We also usually put the scope before the uri, so maybe that would be appropriate here as well.
This change will require a corresponding change in reactive-dependencies.test.ts (as well as `entityKey` below).

packages/runner/src/scheduler.ts
```diff
@@ -1489,10 +1499,11 @@ export class Scheduler {
     if (populateDependencies) {
       handler.populateDependencies = populateDependencies;
     }
-    this.eventHandlers.push([ref, handler]);
+    const normalizedRef = { ...ref, scope: normalizeCellScope(ref.scope) };
```
ubik2: Given that NormalizedFullLink is already supposed to be the form that has all the details filled in, maybe we should be setting scope there and not have it be optional.

packages/runner/src/schema-format.ts
```diff
@@ -237,7 +237,8 @@ function getWrappedTypeString(
   wrapper: AsCellType,
   innerType: string,
 ): string {
-  switch (wrapper) {
+  const kind = typeof wrapper === "string" ? wrapper : wrapper.kind;
+  switch (kind) {
```
ubik2, line 254: While this is mostly just used for debugging, we may want to restore the `PerUser` style wrappers here too.

packages/memory/v2/server-sync.ts
```diff
@@ -119,8 +133,18 @@ export const buildFullSync = (
   toSeq: number,
 ): SessionSync => {
   const removes = [...previous.values()]
```
ubik2, line 139: It's probably more efficient to always have a scope on the entries in `session.entities`. That would also make the removes list in buildDiffSync more compact. If it's too much of a hassle for types, I guess this can be ignored.
If we allow the same id to be used for multiple scopes, it might be worthwhile to include that in the sort criteria (though these may just be ordered for debug convenience, in which case it doesn't matter enough).

packages/runner/src/storage/transaction/attestation.ts
```diff
@@ -339,7 +341,14 @@ export const claim = (
       address.path.length === 0 &&
       "getDocument" in replica &&
       typeof replica.getDocument === "function"
-    ? toTransactionDocumentValue(replica.getDocument(address.id as any))
+    ? toTransactionDocumentValue(
+      (replica as {
```
ubik2: These casts are pretty ugly. Maybe ISpaceReplica should get a getDocument function?

packages/memory/v2/query.ts
seefeldb: Ah, we might actually get the same id in two scopes, with the broader linking to the narrower one. So this has to be fixed.

packages/memory/v2/query.ts
seefeldb: +1

packages/runner/src/scheduler.ts
seefeldb: yes, let's add a type for Scope instead of `string` (or reuse an existing one) and rename this type.

packages/runner/src/scheduler.ts
seefeldb: either that or normalize scope === "space" away (which is what we do for non-redirect links for `overwrite`).

seefeldb:
@cubic-dev-ai review

PR #3508, opened 2026-05-06
feat(api): expose filter/flatMap/findIndex on IDerivable (CT-1307)

mpsalisbury:
## Why

`IDerivable<T>` had an asymmetry: `map` was a typed reactive method, but `.filter()` and `.flatMap()` were deliberately omitted from the type interface (see [#2977](https://github.com/commontoolsinc/labs/pull/2977)) to dodge a conflict with `Array.prototype.filter` / `Array.prototype.flatMap` on the old `OpaqueRef<T[]>` intersection type.

The cost of that omission for pattern authors:

- No autocomplete for `.filter()` / `.flatMap()` on reactive arrays.
- Calls to those methods type-checked against the **synchronous, eager** `Array.prototype` signatures, even though at runtime the ts-transformer lowers them to the **reactive** `filterWithPattern` / `flatMapWithPattern` calls. Inferred return types lied; downstream consumers got plain arrays where they should have gotten `OpaqueRef<…>`.

The two preconditions the original workaround was waiting on have both landed:

1. **`OpaqueRef<T>` is no longer an intersection** — [#3153](https://github.com/commontoolsinc/labs/pull/3153) collapsed `OpaqueRef<T>` to `T` in `packages/api/index.ts`, so the original conflict surface is gone.
2. **ts-transformer Phase 3 has merged** — [#2991](https://github.com/commontoolsinc/labs/pull/2991) lowers `.filter(fn)` / `.flatMap(fn)` on reactive arrays to `.filterWithPattern(pattern, {})` / `.flatMapWithPattern(pattern, {})` (`packages/ts-transformers/src/transformers/reactive-variable-for.ts`, golden at `packages/ts-transformers/test/fixtures/closures/filter-basic.expected.jsx`).

So the workaround is no longer load-bearing, and leaving it in place actively misleads pattern authors and the TypeScript checker.

## What

- Add typed `filter`, `flatMap`, and `findIndex` to `IDerivable<T>` (`packages/api/index.ts`). Signatures mirror the existing runtime methods on `Cell` (`packages/runner/src/cell.ts:1735-…`); `findIndex`/`reduce` callbacks use plain types because they go through `lift()`, which unwraps values.
- Replace the stale workaround docstring above `IDerivable` (it claimed Phase 3 hadn't merged and that `OpaqueRef` was still an intersection — neither is true on `main`).
- Add `packages/runner/test/iderivable-type-surface.test.ts`: a type-surface canary that fails the `deno check` step if `Cell` or `OpaqueCell` ever stop exposing typed reactive `map` / `filter` / `flatMap` / `findIndex`. Runtime body is trivial — the value is in the type-checking pass.

Closes [CT-1307](https://linear.app/common-tools/issue/CT-1307/remove-filterflatmap-workaround-from-iderivable-both-preconditions-now).

## Risk / things to watch

CT-1307 notes that an earlier attempt (pre-Phase-3) at removing the workaround caused 3 type-inference failures in patterns: `write-and-run.tsx:154` (TS2589), `email-task-engine.tsx:452` (TS2345), `usps-informed-delivery.tsx:291` (TS2345). The type landscape has changed since then, so they may no longer reproduce — but CI on those patterns is the place to check.

## Test plan

- [x] CI green on full pattern type-check (especially the three patterns above)
- [x] `deno task test` passes; the new type-surface test compiles
- [x] Spot-check: `.filter()` / `.flatMap()` autocomplete now surfaces on `Cell<T[]>` and `OpaqueCell<T[]>` in pattern code, with reactive return types

### DIFF
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -726,17 +726,9 @@ export interface IEquatable {
 
 /**
  * Cells that allow deriving new cells from existing cells via array methods:
- * .map() and their WithPattern variants.
- *
- * Note: .filter() and .flatMap() are deliberately omitted from the type
- * interface to avoid conflicting with Array.prototype.filter/flatMap on
- * OpaqueRef<T[]> (which is an intersection of OpaqueCell<T[]> & Array<…>).
- * The runtime methods exist on Cell and are intercepted by the proxy; they
- * just don't have type declarations here until the compiler transform
- * (Phase 3) resolves the ambiguity. This means pattern authors won't get
- * autocomplete for .filter()/.flatMap() — they'll see Array.prototype
- * signatures instead. The proxy intercepts correctly at runtime, but
- * TypeScript won't type-check the reactive return type.
+ * direct helpers mirror supported Array methods and return OpaqueRef results.
+ * The WithPattern variants accept pre-defined patterns for per-element
+ * operations.
  */
 export interface IDerivable<T> {
   map<S>(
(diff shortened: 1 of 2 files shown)


mpsalisbury:
@seefeldb @mathpirate 
I was going through some old issues and found this one from after the .flatMap implementation. Is this still correct and useful?

seefeldb (approved):
Awesome, thank you for remembering!

PR #3509, opened 2026-05-06
[codex] Make binary builds use deno.lock

mathpirate:
## Summary
- stop renaming `deno.lock` out of the workspace during binary builds
- pass the workspace lockfile to every `deno compile` invocation with `--frozen=true`
- restore the Cliffy import ranges while keeping the checked-in lock resolved to `1.0.0-rc.8`

## Why
Follow-up to Wilk's approval comment on #3496. The exact Cliffy pin was a tactical stabilization for binary CI, but the healthier fix is making compiled binaries use the same locked dependency graph as install/test flows.

## Tests
- `deno install --frozen=true`
- `deno fmt --check`
- `git diff --check`
- `deno task check`
- `deno task build-binaries --cli-only`
- `deno task build-binaries`
- local `./dist/cf piece call --api-url=http://localhost:1 --identity=/tmp/no-such-key --space=x --piece y increment '{}'` parser smoke reached identity-file loading
- cross-compiled Linux `cf` with `--lock deno.lock --frozen=true`, then Docker-smoked the same `piece call` command to identity-file loading

Base: #3496 / `codex/cf-harness-image-attachments`. Retarget to `main` after #3496 merges.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -130,7 +130,7 @@
     "core-js/proposals/async-explicit-resource-management": "https://esm.sh/core-js@3.46.0/proposals/async-explicit-resource-management",
     "@commonfabric/ports": "./ports.json",
     "@astral/astral": "./packages/vendor-astral/mod.ts",
-    "@cliffy/command": "jsr:@cliffy/command@1.0.0-rc.8",
+    "@cliffy/command": "jsr:@cliffy/command@^1.0.0-rc.8",
     "@noble/hashes": "npm:@noble/hashes@^2.2.0",
     "@std/assert": "jsr:@std/assert@^1",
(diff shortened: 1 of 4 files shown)


willkelly (approved):
This looks good to me. The follow-up does the right thing: binary builds now compile against the workspace lockfile with frozen resolution instead of moving deno.lock aside, and the Cliffy pin can go back to a range while the lockfile keeps the resolved version stable.

One branch-hygiene note before merge: please refresh/rebase this onto current main now that #3496 landed as a squash merge, so the unpin changes in deno.json / packages/cli/deno.json / deno.lock are actually present in the final PR diff. With that rebase, this addresses the follow-up request well.

PR #3546, opened 2026-05-11
Fix sub-pattern calls captured as reactive data in maps

bfollington:
## Summary
- Preserve module-scope pattern factory calls during array-method expression-site lowering.
- Treat declaration-resolved pattern factories the same as directly typed `PatternFactory` callees in call-root classification and reactive `.for()` insertion.
- Add a regression covering the original failing shape: `const row = EntryRow(...); return row[UI];` inside a mapped UI callback.

## How we found it
Deploying basic patterns such as `notes/note.tsx` failed while the system `backlinks-index` pattern was running, with `TypeError: EntryRow is not a function` at `packages/patterns/system/backlinks-index.tsx:205`.

Running `deno task cf check packages/patterns/system/backlinks-index.tsx --root packages/patterns --show-transformed` showed that the authored map callback:

```tsx
const row = EntryRow({ ... });
return row[UI];
```

was lowered into a derive that captured `EntryRow` as reactive input data:

```ts
({ EntryRow, entry }) => EntryRow({ ... })
```

At runtime that captured value is pattern metadata, not the callable pattern factory, which explains the non-callable `EntryRow` error.

## Fix
The transformer now recognizes declaration-resolved pattern factory helpers, not just callees whose type checker surface still directly looks like a `PatternFactory`. That prevents module-scope sub-pattern calls from being wrapped as reactive computations or converted into `.for()` data.

## Verification
- `deno test --allow-read --allow-env packages/ts-transformers/test/pipeline-regressions.test.ts`
- `deno task cf check packages/patterns/system/backlinks-index.tsx --root packages/patterns --no-run`
- `deno task cf check packages/patterns/notes/note.tsx --root packages/patterns --no-run`
- `deno task cf check packages/patterns/counter/counter.tsx --root packages/patterns --no-run`

---
## Summary by cubic
Fix pattern factory calls being captured as reactive data in map callbacks, which caused “EntryRow is not a function” at runtime. Module-scope sub-pattern calls now stay structural and are not wrapped by derive or `.for()`.

- **Bug Fixes**
  - Preserve module-scope pattern factory calls during array-method expression-site lowering.
  - Treat declaration-resolved pattern factory helpers the same as direct `PatternFactory` callees in call-root classification and reactive `.for()` checks.
  - Add a regression test for `const row = EntryRow(...); return row[UI];` inside a mapped UI callback.

<sup>Written for commit 5686c176489f1734aa2783e4e7a9aa5bbba23b78. Summary will update on new commits.</sup>

### DIFF
--- a/packages/ts-transformers/src/transformers/expression-site-lowering.ts
+++ b/packages/ts-transformers/src/transformers/expression-site-lowering.ts
@@ -30,6 +30,10 @@ import type { AnalyzeFn } from "./expression-rewrite/types.ts";
 import type { ExpressionContainerKind } from "./expression-site-types.ts";
 import { createDeriveCall } from "./builtins/derive.ts";
 import { classifyOpaquePathTerminalCall } from "./opaque-roots.ts";
+import {
+  isPatternFactoryCalleeExpression,
+  isPatternFactoryHelperExpression,
+} from "./structural-reactive-factory.ts";
 
 interface RewriteExpressionSiteParams {
   readonly expression: ts.Expression;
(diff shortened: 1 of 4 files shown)


bfollington:
Closing in favour of @mathpirate's combined fix

PR #3553, opened 2026-05-12
fix(runner): preserve array-ness in mergeAnyOfMatches anyOf merge (CT-1562)

mathpirate:
Fixes [CT-1562](https://linear.app/common-tools/issue/CT-1562/transformer-bug-roomsmapjoin-breaks-for-cells) — `rooms.map().join()` crashes with `TypeError: rooms.map is not a function` for patterns that declare `rooms: Room[] | Default<[]>`.

The ts-transformers pipeline is innocent. Transformer output is byte-identical between main and any branch that touched it; the bug lives entirely in the runtime schema traversal.

## The fix

A one-token change in `packages/runner/src/traverse.ts:mergeAnyOfMatches`: change the merge-eligibility guard from `isRecord` to `isObject`. Arrays satisfy `isRecord` (because `typeof [] === \"object\"`), so the existing `Object.assign({}, ...matches)` path fired for multi-array matches and corrupted them into `{ \"0\": …, \"1\": … }` — dropping `.map`. `isObject` excludes arrays (matching the surrounding code's existing pattern of treating arrays distinctly), so multi-array matches now fall through to the existing `matches[0]` semantics — array-ness preserved.

Triggered by the `anyOf` shape that `Default<[]>` produces — an empty-array branch (`items: false`) alongside the populated branch. Both can match a populated array, so the merge fires. After the fix, the merge no longer applies to arrays at all.

## Tests

Unit tests in `packages/runner/test/traverse.test.ts`:
- New `mergeAnyOfMatches` describe block — 6 tests covering empty matches, single match, the existing object-merge (still green), two array-preserving cases (RED before fix → GREEN after), and the mixed-type fall-through.
- Two new tests in `SchemaObjectTraverser array traversal` confirming the standalone traversal path already rejects populated arrays against `items: false` (so the bug is genuinely localized to the anyOf merge, not the per-item validator).
- One new test in `canBranchMatch` confirming the empty-array case against `items: false` is accepted (baseline).

End-to-end regression test in `packages/runner/test/patterns-default-array-materialization.test.ts`: exercises the full `PatternManager.compilePattern` + `runtime.run` path with Berni's repro shape. RED before fix (scheduler error matching the ticket); GREEN after.

Production verified via `cf piece apply` against the original failing fixture: returns `\"alpha: 2\\nbeta: 0\"` instead of TypeError.

## Investigation history

The investigation surfaced two candidate logical errors. The initial fix touched both. During review, @ubik2 (runtime area owner) pointed out that:

1. The array-aware `mergeAnyOfMatches` branch was unnecessarily explicit — switching `isRecord` to `isObject` achieves the same outcome by simply excluding arrays from merge eligibility, matching the surrounding code's existing convention.
2. The proposed `canBranchMatch` change for `items: false` belonged in the real-traversal path, not the fast-reject path. The fast-reject is conservative by design (defers element-level constraints — object required-value mismatches aren't fast-rejected either). Adding array element-count validation there broke that symmetry, and was unreachable on CT-1562's actual path anyway (the value at the call site is a `{$alias}` link, which `canBranchMatch` already short-circuits).

The merged commit history includes Robin's proposed alteration (#3559) on top of the original investigation work.

## Adjacent observations (investigated, no separate defect)

- **`traverseWithSchema` apparently accepting `items: false` against a populated array in the anyOf path.** The standalone traversal path correctly rejects this (test: `rejects populated array when items is false and no prefixItems`). The leak was an emergent effect of link resolution + `traverseCells=false` semantics on a should-never-match branch.
- **\"Non-determinism\" between two identical traversal calls.** Actually two distinct traversal modes — `traverseCells=true` (query/scheduling) vs `traverseCells=false` (materialization) — producing legitimately different intermediate results for a should-never-match branch. Not a determinism bug.

## Test population

Patterns under `packages/patterns/` that use the `T[] | Default<[]>` idiom and route through value-site derives would have been candidate exposures for this bug (and are now fixed): `deep-research.tsx`, `location-track.tsx`, `render-test.tsx`, `tags.tsx`, `occurrence-tracker.tsx`, `record.tsx`.


packages/runner/src/traverse.ts
```diff
@@ -642,6 +642,16 @@ export function mergeAnyOfMatches<T>(
...
+    // doesn't apply to arrays — it produces { "0": …, "1": … } and drops
+    // array-ness. When all matches are arrays, return whichever has
+    // content; if all are empty, return the first. This is what CT-1562
+    // hits via the Default<[]> schema's anyOf split, where one branch
+    // produces [] and the other produces the populated array.
+    if (matches.every((v) => Array.isArray(v))) {
+      return matches.find((m) => (m as unknown[]).length > 0) ?? matches[0];
+    }
     // If all our matches are objects, merge the properties.
     if (matches.every((v) => isRecord(v))) {
```
ubik2: What about changing this test instead?
If we have an `anyOf` that matches both an array and an object, maybe it's better to punt to the `matches[0]` instead of adding index keys to the merged object.
```typescript
    if (matches.every((v) => isObject(v))) {
```

packages/runner/src/traverse.ts
```diff
@@ -3298,6 +3308,26 @@ export function canBranchMatch(
     }
   }
 
+  // `items: false` on an array schema means "no additional items allowed
```
ubik2: I think this may not belong in the fast reject path. This is just our initial pass to quickly prune the obviously invalid options. We still should be rejecting these later.
I presume in this case, we've shifted to unions for expressing defaults, so we got something like:
```typescript
type Result = {name: string, summary: string}[] | Default<[]>
```
This code would then easily exclude the Default branch for our array that has an element, which has some similarity to the rejection of an object that didn't have a `name` key (rejection based on keys -- array has key 0 so fast-path reject, while object missing key `name` would also fast-path reject).
However, for objects that have an entry that isn't valid, we defer that until the real check, so I'd rather do that for arrays too.

packages/runner/src/traverse.ts
ubik2: I created a PR that implements this suggestion (it will apply to this branch):
https://github.com/commontoolsinc/labs/pull/3559

I removed the canBranchMatch rejection unit test, since I don't think that's ideal.

packages/runner/src/traverse.ts
ubik2: With the mergeAnyOfMatches change (using isObject), I don't think this is needed.

packages/runner/src/traverse.ts
mathpirate: folded yours in thanks

packages/runner/src/traverse.ts
mathpirate: folded yours in thanks

PR #3560, opened 2026-05-12
Fix memory v2 pending replay after async commits

ubik2:
This was generated by codex, and I don't think it's particularly valid, but I've created a PR for it in case it's useful when investigating the pull consistency issue.

Here's codex's conclusion after changing things:
```
• Resumed and finished the investigation.

  Root cause: the server did store both writes. The client lost sum because a stale watch/pull upsert for
  setup seq 3 arrived after local confirmation seq 4, rolled the confirmed base backward to { value: {} },
  then replayed pending patches over that stale base.

  I fixed this without waiting on tx.commit():

  - Leaf writes into newly created docs now stay as path patches when safe.
  - Pending confirmations only promote contiguous applied pending entries.
  - Stale session upserts no longer move the confirmed base backward.
  - Removed the masking ctx.runtime.storageManager.synced() from pattern-and-data-persistence.test.ts.

  The transact snapshot for the sum example:

  {
    "computeSum": {
      "op": "patch",
      "patches": [{ "op": "add", "path": "/value/sum", "value": 15 }]
    },
    "formatResult": {
      "op": "patch",
      "patches": [
        { "op": "add", "path": "/value/result", "value": "Numbers: 15" }
      ]
    }
  }

  Server final state for the internal doc had both fields:

  {
    "value": { "sum": 15, "result": "Numbers: 15" }
  }
```

Preserve fine-grained patch writes for safe leaf updates into newly created documents, and keep local pending state stable while commit confirmations and watch refreshes arrive out of order.

Only promote confirmed pending entries as a contiguous prefix, and ignore stale session upserts that would move the confirmed base backward. This prevents a later pull from replaying pending patches over an older base and dropping sibling writes.

Remove the synced() workaround from the persistence integration test and add coverage for concurrent leaf writes, patch drafts, and out-of-order confirmations.

---
## Summary by cubic
Fixes pending replay in memory v2 when commit confirmations and watch refreshes arrive out of order. Preserves sibling fields and keeps confirmed bases monotonic, especially for concurrent leaf writes into new docs.

- **Bug Fixes**
  - Promote pending entries only as a contiguous confirmed prefix; tag each with `appliedSeq` to avoid replaying over stale bases (`packages/runner/src/storage/v2.ts`).
  - Ignore stale session upserts that would move the confirmed base backward.
  - Emit path `patch` drafts for safe leaf writes into newly created docs; keep root writes as `set` (`packages/runner/src/storage/v2-transaction.ts`).
  - Retry `writeBatch` before creating parent entries on `NotFound` (`packages/runner/src/storage/extended-storage-transaction.ts`).
  - Tests: add cases for concurrent leaf writes and out-of-order confirmations; remove `synced()` workaround in integration test.

<sup>Written for commit b39c6d89407ab9363087ead7e341f57bc2c0d5fd. Summary will update on new commits.</sup>

### DIFF
--- a/packages/runner/integration/pattern-and-data-persistence.test.ts
+++ b/packages/runner/integration/pattern-and-data-persistence.test.ts
@@ -203,8 +203,6 @@ async function phase2LoadAndVerify(
     resultCell,
   );
   await tx.commit();
-  // FIXME(@ubik2) - Why is the tx.commit insufficient?
-  await ctx.runtime.storageManager.synced();
   const pulled = await runResult.pull();
 
   const output = runResult.getAsQueryResult();
--- a/packages/runner/src/storage/extended-storage-transaction.ts
+++ b/packages/runner/src/storage/extended-storage-transaction.ts
@@ -391,6 +391,16 @@ export class ExtendedStorageTransaction implements IExtendedStorageTransaction {
       writeResult.error &&
       (writeResult.error.name === "NotFoundError")
     ) {
+      if (this.tx.writeBatch) {
+        const batchRetry = this.tx.writeBatch([{ address, value }]);
+        if (!batchRetry.error) {
+          return;
+        }
+        if (batchRetry.error.name !== "NotFoundError") {
+          throw toThrowable(batchRetry.error);
+        }
+      }
+
       // Create parent entries if needed.
       // errorPath includes the missing key (consistent with read errors).
       // lastExistingPath is one level up - the actual last existing parent.
--- a/packages/runner/src/storage/v2-transaction.ts
+++ b/packages/runner/src/storage/v2-transaction.ts
@@ -1883,12 +1883,18 @@ export class V2StorageTransaction implements IStorageTransaction {
     type: MediaType,
     doc: WritableDocumentEntry,
   ): NativeStorageCommitOperation | null {
-    if (doc.initial.value === undefined || doc.current.value === undefined) {
+    if (doc.current.value === undefined) {
       return null;
     }
 
     const details = [...doc.patchDetails.values()];
     if (details.some((detail) => detail.address.path.length === 0)) {
+      return null;
+    }
+    if (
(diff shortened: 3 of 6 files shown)


ubik2:
@seefeldb Added you as a reviewer here, but I don't plan to merge this.
I haven't looked too deep into this, but had codex look at a runner issue, and this was it's conclusion (after being told not to await all the tx.commit calls).
I just created the PR in case it's useful to you if you're also debugging this.

ubik2:
Adding more codex description:
The bad ordering was roughly:

  1. Client setup write creates the internal result doc on the server:
     seq 3 -> { value: {} }
  2. The client still has a queued watch/pull refresh for that setup state. That refresh has not been
     applied locally yet.
  3. computeSum runs locally and writes:

     { "op": "add", "path": "/value/sum", "value": 15 }

     The client applies this optimistically and sends the transact to the server.
  4. Server accepts that write:
     seq 4 -> { value: { "sum": 15 } }
  5. Before everything has settled, formatResult sees the local optimistic sum and writes:

     { "op": "add", "path": "/value/result", "value": "Numbers: 15" }

     The server also accepts this. Server state is correct:

     { "value": { "sum": 15, "result": "Numbers: 15" } }
  6. The bug: the old queued setup refresh from step 2 arrives late on the client:

     { "seq": 3, "value": { "value": {} } }

     The client accepted it and moved its confirmed base backward from seq 4 to seq 3.
  7. Then when the result write confirmed, the client replayed pending state over that stale base. Since the
     stale base was { value: {} }, replay produced:

     { "value": { "result": "Numbers: 15" } }

     sum disappeared locally even though the server had it.

  So the issue was client-side ordering, not server commit preparation. The fix is to reject stale session
  upserts that would regress the confirmed base, and to promote pending confirmations only in contiguous
  order so later confirmations do not corrupt replay state.

PR #3566, opened 2026-05-12
Validate SES callback callable captures

mathpirate:
## Summary

This adds the first source-level validator pass for SES self-contained callback constraints in the pattern transformer.

The policy this encodes is intentionally narrow:

- ordinary standalone functions may not act as Common Fabric builder factories, regardless of whether they capture anything
- SES callback boundaries may not capture function-local callables/functions from an enclosing function scope
- module-scope helper functions remain valid callback dependencies
- schema-backed data/cell captures remain valid for now when existing closure strategies explicitly transport them

## What Changed

- Extends `PatternContextValidationTransformer` so standalone functions reject direct builder factories such as `action()`, `computed()`, `handler()`, `lift()`, `pattern()`, and `render()`.
- Keeps the existing standalone `derive()` and reactive-array operation rejection path.
- Adds `ses-callback:callable-capture` diagnostics for supported callback boundaries that capture callable helpers from enclosing function scope.
- Covers derive callbacks, action callbacks, JSX event handlers, and reactive map callbacks.
- Removes the stale `generic-builder-type-parameters-unknown` fixture, which tested generic functions returning `lift()` / `handler()`; that shape now contradicts the SES builder-factory policy.

## Notes

This is not yet the stricter "reject every untransported non-module capture" validator. It only rejects callable/function captures plus standalone builder factories. That keeps the first pass aligned with the concrete bug class without invalidating ordinary transported reactive state captures.

## Test Coverage

Added targeted validation coverage for:

- standalone `action()`, `handler()`, `lift()`, and `pattern()` factory rejection
- `derive()` callback callable captures
- `action()` callback callable captures
- JSX event handler callable captures
- reactive `.map()` callback callable captures from pattern inputs
- allowed module-scope helper references
- allowed action callbacks that capture reactive state cells

Also ran the broader fixture-based transformer suite to check for unexpected diagnostics in existing fixture patterns.

## Verification

- `/Users/gideonwald/.deno/bin/deno test -A packages/ts-transformers/test/validation.test.ts`
- `/Users/gideonwald/.deno/bin/deno test -A test/fixture-based.test.ts` from `packages/ts-transformers`
- `/Users/gideonwald/.deno/bin/deno fmt --check packages/ts-transformers/src/transformers/pattern-context-validation.ts packages/ts-transformers/test/validation.test.ts`
- `git diff --check`
- 
---BEGIN COPY-PASTE---
NEW_PERF_BASELINE: subtest: pattern-unit-4/pattern-unit-tests > packages/patterns/reading-list/reading-list.test.tsx = 21s
NEW_PERF_BASELINE: job: Build Binaries = 68s
---END COPY-PASTE---


packages/patterns/tools/importer-prompt.ts
```diff
@@ -981,38 +982,94 @@ export type ScopeKey =
...
-  "webhook:manage": "Manage webhooks",
-};
-
-/** Unified scope registry for the auth manager factory */
-const SCOPES: AuthManagerDescriptor["scopes"] = Object.fromEntries(
-  Object.entries(SCOPE_DESCRIPTIONS).map(([key, desc]) => [
-    key,
-    { description: desc, scopeString: key },
-  ]),
+const AIRTABLE_SCOPE_DESCRIPTIONS = __cf_data(
```
seefeldb: transformer should have added __cf_data automatically here

packages/ts-transformers/src/transformers/pattern-context-validation.ts
```diff
@@ -57,8 +57,38 @@ import {
...
+  "event-handler",
+  "reactive-array-method",
+  "pattern-builder",
+  "render-builder",
+  "derive",
+  "computed-builder",
+  "action-builder",
+  "lift-builder",
+  "handler-builder",
+]);
```
seefeldb, line 84: see whether patternTool should go here? only if we don't already transform the parameter to it to `pattern`.

packages/ts-transformers/src/transformers/pattern-context-validation.ts
```diff
@@ -57,8 +57,38 @@ import {
...
+
+const STANDALONE_FUNCTION_BUILDER_NAMES = new Set([
```
seefeldb: this is probably already defined somewhere else

packages/patterns/tools/importer-prompt.ts
mathpirate: Addressed in 04b6cf97c. I removed the authored `__cf_data` import/wrappers from the importer prompt example and from the Google/Airtable auth-manager sources, so these are plain module-scope literals again and the transformer owns injecting `__cf_data` where needed.

packages/ts-transformers/src/transformers/pattern-context-validation.ts
mathpirate: Checked this and addressed it in 04b6cf97c. `patternTool` keeps its own `patternTool(...)` callback boundary; it is not transformed into `pattern(...)`, so I added `pattern-tool` to `SES_SELF_CONTAINED_CALLBACK_BOUNDARIES` and added a regression test for a `patternTool` callback capturing an enclosing helper function.

packages/ts-transformers/src/transformers/pattern-context-validation.ts
mathpirate: Addressed in 04b6cf97c. I replaced the local `STANDALONE_FUNCTION_BUILDER_NAMES` set with the shared `COMMONFABRIC_REACTIVE_ORIGIN_BUILDER_NAMES` registry data from `commonfabric-runtime-registry.ts`, so the validator no longer duplicates that builder-name list.

PR #3567, opened 2026-05-12
Fix zero-input reactive root lowering

bfollington:
## Summary

Fix zero-input pattern bodies so local reactive roots created inside the body are still lowered before runtime.

## Bug

Deploying pieces could succeed while the runtime later failed in system patterns with errors like:

```text
TypeError: piece.get is not a function
```

`cf check --show-transformed` showed the problem in `summary-index.tsx`: the compiled output left a local wish result as plain property access:

```ts
const mentionable = wish(...).result;
```

That meant downstream code iterated the unwrapped result value, then called `piece.get()` on an item that was no longer a cell.

## Cause

`pattern-body-reactive-root-lowering` returned early when a pattern body had no input-root opaque bindings:

```ts
if (opaqueRoots.size === 0 && opaqueRootSymbols.size === 0) return body;
```

That is too broad. Zero-input patterns can still create local reactive roots inside the body, for example:

```ts
const mentionableWish = wish(...);
const mentionable = mentionableWish.result;
```

Those local roots need the same lowering pass so `.result` becomes `.key("result")`.

## Fix

Keep the fast early return, but only take it when there are no input roots and no local reactive-origin bindings in the body.

The new helper scans the current pattern body for local variable initializers rooted in reactive-origin calls, such as `wish(...)`. If it finds one, the existing lowering pass runs and registers that local binding normally.

`SummaryIndex` and `KnowledgeGraph` now name the local wish root before reading `.result`, giving the transformer a stable binding to lower.

## Verification

```sh
deno test --allow-env --allow-read packages/ts-transformers/test/transform.test.ts
deno task cf check packages/patterns/system/summary-index.tsx --root packages/patterns --show-transformed --no-run
deno task cf check packages/patterns/system/knowledge-graph.tsx --root packages/patterns --show-transformed --no-run
deno task cf check packages/patterns/system/default-app.tsx --root packages/patterns --no-run
deno task cf check packages/patterns/notes/note.tsx --root packages/patterns --no-run
```

The transformed output now emits:

```ts
const mentionable = mentionableWish.key("result");
```

---
## Summary by cubic
Fixes lowering for zero‑input pattern bodies and one‑line reactive‑origin chains so `.result`/`pending`/`.length` become `.key(...)`, and stabilizes identity with `.for("<name>", true)` on destructured leaves.

- **Bug Fixes**
  - Run the body walker in zero‑input patterns when local reactive‑origin bindings exist via `isOpaqueOriginCall`.
  - Add a pre‑pass that rewrites `wish(...).result[.foo]` into destructures; preserves `as`/`satisfies` casts and drops only parens/non‑null.
  - Keep the inline reactive‑root access diagnostic, narrowed to skip only direct call receivers (e.g. `Writable.of(...).for(...)`) and still flag `wish(...).result.get()`.
  - Extend `isOpaqueSourceExpression` to see `.for(...)` chains and origin‑rooted access chains as opaque for correct closure capture wiring.
  - Attach `.for("<name>", true)` to each destructured leaf when the root is a fresh opaque origin; update docs and tests.

<sup>Written for commit 96a4d8b79380d833f3be8012aa415cadc2a89c1c. Summary will update on new commits.</sup>

### DIFF
--- a/packages/ts-transformers/src/transformers/opaque-roots.ts
+++ b/packages/ts-transformers/src/transformers/opaque-roots.ts
@@ -177,7 +177,14 @@ export function isOpaqueSourceExpression(
 
     if (ts.isPropertyAccessExpression(current.expression)) {
       const methodName = current.expression.name.text;
-      if (methodName === "key" || methodName === "get") {
+      // `.key()` returns a navigated sibling; `.get()` returns a value
+      // proxy; `.for()` returns the same cell (identity-preserving cause
+      // annotation, no-op semantically). All preserve opaque-source
+      // provenance, so a chain ending in one of these stays opaque iff
+      // its underlying receiver is opaque.
+      if (
+        methodName === "key" || methodName === "get" || methodName === "for"
+      ) {
         return isOpaqueSourceExpression(
           current.expression.expression,
           opaqueRoots,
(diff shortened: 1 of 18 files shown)


packages/ts-transformers/src/transformers/opaque-roots.ts
```diff
@@ -177,7 +177,14 @@ export function isOpaqueSourceExpression(
 
     if (ts.isPropertyAccessExpression(current.expression)) {
       const methodName = current.expression.name.text;
-      if (methodName === "key" || methodName === "get") {
+      // `.key()` returns a navigated sibling; `.get()` returns a value
+      // proxy; `.for()` returns the same cell (identity-preserving cause
```
seefeldb: nit, but it's a `value`, only rarely a proxy (only if the schema is `true`).

seefeldb (approved):

mathpirate (approved):

PR #3573, opened 2026-05-13
Prevent upserts from rolling back seq

ubik2:
applySessionSync() now ignores per-document upserts whose seq is older than the document’s current record.confirmed.seq. Local commit confirmations can already advance the confirmed base, and replaying pending patches depends on that base being monotonic. Accepting an older sync could roll the base backward and drop sibling fields during replay.

NEW_PERF_BASELINE: subtest: pattern-unit-4/pattern-unit-tests > packages/patterns/reading-list/reading-item-detail.test.tsx = 12s

---
## Summary by cubic
Ignore stale per-document upserts in `applySessionSync()` so `record.confirmed.seq` never moves backward, preventing replay from dropping fields.

- **Bug Fixes**
  - Skip session sync upserts with `seq` < `record.confirmed.seq` in `packages/runner/src/storage/v2.ts`.
  - Add tests in `packages/runner/test/memory-v2-stacked-commit.test.ts` to confirm stale syncs don’t overwrite ordered confirmations or erase earlier sibling fields while later patches are pending.

<sup>Written for commit c8dbaf8b0293b1d4bf08b5e27829a74fb8d9c1fe. Summary will update on new commits.</sup>


seefeldb (approved):
Nice catch!

ubik2:
Merged in another PR

PR #3587, opened 2026-05-14
feat(transformer): Narrow transformed cell capabilities

seefeldb:
## Summary

- infer narrower cell capabilities for transformed lifts, derives, and handlers, including readonly, writeonly, comparable, and opaque cells
- treat `.set`, `.update`, and `.push` as writes while keeping opaque-only derivation methods at opaque capability where possible
- preserve emitted wrapper schemas for `ReadonlyCell`, `WriteonlyCell`, `ComparableCell`, and `OpaqueCell`

## Validation

- `deno task test` in `packages/ts-transformers`
- `deno task check`

---
## Summary by cubic
Narrows transformed inputs, captured cells, and explicit `lift` schemas to the least capability needed. Adds `comparable`, keeps derivation roots opaque, and fixes wrapper resolution and schema emission to always use the correct kind.

- New Features
  - Capability analysis: infer `readonly`/`writeonly`/`writable`/`comparable`/`opaque`; treat `.set`/`.update`/`.push`/`.removeAll` as writes; keep `map`/`flatMap`/`filter` roots opaque while retaining nested receivers; mark `equals`/`equalLinks` as comparable-only; if derivation and equality touch the same root, stay opaque.
  - Type shrinking and schema emission: narrow captured cell and explicit `lift` schemas by usage (emit `asCell: ["readonly" | "writeonly" | "writable" | "comparable" | "opaque"]`); emit `ReadonlyCell`, `WriteonlyCell`, `ComparableCell`, and `OpaqueCell`; wrap identity-only leaves as `ComparableCell<unknown>`; apply capability summaries to builder fallback argument types.

- Bug Fixes
  - Wrapper resolution: prefer semantic wrapper kind when authored nodes disagree; use the synthetic node’s kind when the wrapper node is generated; resolve from the actual type when no inner node is present (including `typeRegistry` wrappers).
  - Schema generator: plumb `sourceFile` through the `schema-generator` plugin and generator to resolve names from synthetic nodes.
  - Native types: resolve local lib-declared names (e.g. `Date`) before applying native fallbacks, and recognize Node `URL` as a native type.

<sup>Written for commit 941c098a1833115b4b3e91c99bd1dfbec3713e29. Summary will update on new commits.</sup>

### DIFF
--- a/packages/schema-generator/src/formatters/common-fabric-formatter.ts
+++ b/packages/schema-generator/src/formatters/common-fabric-formatter.ts
@@ -261,6 +261,19 @@ export class CommonFabricFormatter implements TypeFormatter {
     }
 
     const wrapperInfo = getCellWrapperInfo(type, context.typeChecker);
+    if (
+      resolvedWrapper &&
+      resolvedWrapper.kind !== "Default" &&
+      wrapperInfo &&
+      wrapperInfo.kind !== resolvedWrapper.kind
+    ) {
+      return this.formatWrapperTypeFromNode(
+        resolvedWrapper.node,
+        context,
+        resolvedWrapper.kind,
+      );
+    }
+
     if (wrapperInfo && !(type.flags & ts.TypeFlags.Union)) {
       const nodeToPass = this.selectWrapperTypeNode(
         n,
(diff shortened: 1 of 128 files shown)


mathpirate:
[P1] Treat `removeAll` as a writer

`IWritable` exposes both `remove` and `removeAll` as array-cell mutators, but `ARRAY_IDENTITY_WRITER_METHODS` only includes `remove`. A callback like `state.items.removeAll(item)` falls through to unknown-method handling, records a read of `items`, and emits readonly capability instead of write/writeonly, which can drop write authority for valid mutations.

mathpirate (approved):

PR #3572, opened 2026-05-13
fix(runner): remove legacy frozen-as-terminal proxy short-circuit

danfuzz:
## Summary

Deletes the `modernDataModel`-conditional short-circuit at
`createQueryResultProxy` (`packages/runner/src/query-result-
proxy.ts:162-168`) so frozen values flow through the proxy's
existing shallow-clone / mutate / refreeze machinery rather
than being returned raw. The short-circuit encoded a legacy
"frozen = terminal" assumption that was structurally true
under the old modern/legacy split — frozen objects originated
only from legacy storage and were known to be link-free
terminal data — but no longer holds: with `modernDataModel`
graduated, frozen values may now reach the proxy from either
side of the split, carrying link structures that need
resolution + proxy-wrap handling. This change is *subtractive*:
it does not add any new clone-mutate-refreeze machinery; it
stops *bypassing* the existing machinery (the 9-method
`arrayMethods` table at `:77-113`, the get-trap shallow-clone-
mutate-refreeze cycle at `:266-321`, and the unfrozen-stub
trick at the original `:170-192` for ECMAScript proxy-
invariant compliance). A dedicated unit test
(`frozen-proxy-target.test.ts` / "wrap fires under
modernDataModel=OFF") documents the corrected contract
synthetically; the failure cases it guards against only
become naturally reachable once a separate downstream change
(see "Broader arc" below) removes the broad clone at
`decodeMemoryBoundary` that currently re-thaws decoder output
before it reaches the proxy.

## Review

**APPROVE-no-NITs at fix-tip `766dad6f4` (mechanical-rebase
forward-stable to `d2de45241`) by `reviewer-flux`.** Lens-8
at-source trace covered (substantive):

1. 9-method `arrayMethods` table at `:77-113` verified — each
   method present with file:line + ReadWrite/WriteOnly
   variant. This empirically substantiates the foundational
   claim that the proxy already implements the machinery this
   change stops bypassing.
2. Get-trap four-part cycle at `:266-321` verified (read →
   shallow `[...]` / `.map(createProxyForArrayValue)` →
   `method.apply(copy, args)` → `diffAndUpdate`).
3. Unfrozen-stub trick at the new `:162-184` preserved
   byte-identical — the load-bearing ECMAScript 10.5.8
   invariant handler is unchanged.
4. Pure subtractive on code side (-8) + pure additive on
   test side (+96); diff-minimality holds.
5. Upstream control-flow trace: the deleted `if`-block was
[...]

PR #3570, opened 2026-05-13
fix(memory): deep-freeze applyPatch return at function boundary

danfuzz:
## Summary

Wraps `applyPatch`'s return in `deepFreeze` so the post-patch
tree is fully-frozen at the function boundary, rather than the
worst-of-both-worlds partial thaw (top-level container unfrozen,
deep children frozen-by-reference) that the existing
`shallowCloneContainer`-only implementation produces. Operationalizes
the inner-narrowness clause at site S1 of the decode-boundary
clone arc: the `shallow-clone / mutate / refreeze` pattern at
`applyPatch` already had shallow-clone and mutate; this adds the
refreeze. Cedar's coder-time verify (memory 99/0, runner 415/0,
no new failures) confirms PR-A is atomic-landable — with the
broad clone at `decodeMemoryBoundary` still in place, this
change is a no-op relative to current behavior, and clears
~114 of the ~119 B1-shape ("Cannot assign to read only
property") test failures named in the decode-boundary-clone-
removal classification once the broad clone is later removed.

## Review

**APPROVE-no-NITs at `146d33031` by `reviewer-flux`.** Lens-8
at-source trace covered: `shallowCloneContainer` unfrozen-top-
level verified at `patch.ts:289-299`; deep-children-frozen-by-
reference under `Object.assign` shallow property-copy;
`deepFreeze` walk path verified at `data-model/deep-freeze.ts`
cache-short-circuit; clause-2 anchor verified at-source; atomic-
landable claim verified via chain-of-reasoning (broad clone
downstream re-thaws, so PR-A is no-op for behavior — see "Perf
cost + handoff" below for the perf-impact correction); ~114-
of-~119 number internally consistent; import path
`@commonfabric/data-model/deep-freeze` resolves via
`packages/data-model/deno.json` exports map. Test coverage
disposition: no new dedicated test required per
`rules/code-process.md` — existing 7 tests at `v2-patch-test.ts`
exercise `applyPatch` and pass under PR-A; the deep-freeze
property is observable downstream once PR-C lands.

## Research input

* `coordination/docs/2026-05-12-decode-boundary-clone-failures.md`
  (`597e77c`) — §3.1 contains the B1-cascade trace identifying
  `applyPatch`'s partial-thaw return as the root cause of the
  ~119-failure cascade visible under the proposed `return
  decoded as Value` change at `decodeMemoryBoundary`.
* `coordination/docs/2026-05-13-decode-clone-mutation-site-
[...]

PR #3384, opened 2026-04-24
Intern schemas when loaded from `memory`.

danfuzz:
This PR is a bit of an experiment: Documents coming out of the `memory` module are deep frozen (as of only very recently), and we can use the fact of this immutability to add a tactical call to `internSchema()` when loading schemas from `memory`, which potentially unlocks a handful of performance improvements.

NOT READY FOR MERGE: This is not yet determined to be a net win!

PR #3588, opened 2026-05-14
fix(runner): refreeze pending-version materialization; remove legacy mutable-read contract

danfuzz:
## Summary

Refreezes the storage-internal pending-version materialization chain
unconditionally at its three boundary returns (`applyPendingVersion`
set-op, `cloneWithValueAtPath`, `cloneWithoutPath`) so that
`record.materialized.prefixes[i].value` and downstream consumers see
deep-frozen storage state regardless of data-model flag state.

This is **PR-AUDIT-A** from Remy's deep-freeze discipline audit
(`coordination/docs/2026-05-14-deep-freeze-discipline-audit.md`,
finding F2). See §A.6 in that doc for the implementation arc,
including the legacy-mutable-read contract removal rationale.

## Why this also removes a contract

This session's work converges the legacy world toward the modern
world's frozen-data-through-the-system discipline. PR #3577 added a
bifurcation that asserted unfrozen reads under legacy mode (see the
deleted `frozen-proxy-target.test.ts:549` test) — that bifurcation
was about **preserving legacy in support of making the flag-on world
work**. This PR's converse arc — **converging legacy toward modern's
frozen-data discipline** — removes that pin.

Per Dan's rationale:

> "the work of this session is making the legacy world look more like
> the modern world. it is okay to drop the 'result expected to be
> mutable' requirement exactly because that's the direction we're
> headed... nails can be removed, when appropriate, and it seems like
> now's the time!"

Both arcs are correct in their own time: PR #3577 nailed the contract
when it needed nailing; this PR pulls the nail when it's time.

## Rationale (Dan's two clauses, from the session 2026-074 arc)

Carried verbatim from PR #3570 / #3572 / #3576:

> **Clause 1 (outer narrowness):** frozen data through the system to
> the extent possible / narrowest-possible place to clone-as-mutable
> then mutate then deep-freeze.

> **Clause 2 (inner narrowness):** the "thaw" part of "thaw - mutate
> - freeze" should also be _narrow_ which is to say a _shallow_
> clone wherever possible, or more generally the _shallowest possible
> clone in context_.

F2's three-boundary refreeze is clause-1 at the storage-internal
boundary. The existing inner shallow-clones (`shallowCloneContainer`
at intermediate path segments) are clause-2 — preserved unchanged.

## Test plan

- [x] `deno task test` in `packages/runner` (448 passed / 2413 steps / 0 failed)
- [x] New: 3 regression tests asserting `isDeepFrozen` on
  `cloneWithValueAtPath` + `cloneWithoutPath` outputs (both flag
  states; the invariant is unconditional)
- [x] Deletes `frozen-proxy-target.test.ts:549` legacy mutable-read
  assertion (Dan-authorized nail removal)
- [x] Local sanity benches on storage-transaction-path +
  cell-set-nested-array-docs (within typical ranges)
- [x] CI Performance Check — one regression flagged on
  `Generated Patterns Integration Tests (4/4)` (+12s over threshold);
  acknowledged via `NEW_PERF_BASELINE` in the Performance note below.
  Expected to remediate either way once F1 lands (recovery if real;
  no-op if noise).

## Audit intersection

- **PR #3384**: F2-first sequencing makes #3384's `cid:`-only
  `deepFreeze(incoming)` redundant for the wire-decode path. The
  comment in #3384 refers to a stale mechanism
  (`decodeMemoryV2Boundary` was renamed `decodeMemoryBoundary` and
  no longer forces mutability post-#3576). See audit §6.1.
- **PR #3569**: This PR is independent of #3569 (the refreeze fires
  under both flag states now). They land on independent schedules.

## Performance note

One perf check shows a small slowdown. Best guess: this is a bona-
fide regression introduced by this PR, rather than GHA shard-
scheduling variance — but the `(4/4)` shard is one hop downstream
of the changed code's direct surface, and we have not isolated it
definitively, so the "noise vs. real" question carries irreducible
uncertainty.

| Metric | Current | Baseline (median) | Threshold | Change |
|---|---|---|---|---|
| `job: Generated Patterns Integration Tests (4/4)` | 1m 13s | 1m 1s | 1m 10s | +20% / +12s |

All 47 subtests under that job remain `OK` — only the aggregate
job-level wall-time crossed threshold.

Either way, F1 should remediate. If the regression is real, the
unconditional `deepFreeze` adds work on the legacy read path that
`freezeReadValue`'s `Object.isFrozen` short-circuit previously
masked; F1's `[DEEP_FREEZE]` protocol-extension (audit §3.7 /
§3.7.9 / §3.7.10) unlocks `isDeepFrozenFabricValue` to traverse
`FabricInstance` instances, letting `cloneIfNecessary` with
`{frozen: true, deep: true}` use its frozen-data-sharing fast
path — recovering the perf. If the regression was noise, F1's
no-op on this signal is fine. The deferred-optimization shape in
the Comment 2 thread is the same surface; F1's PR is
**PR-AUDIT-C-revised**, queued for the next session.

NEW_PERF_BASELINE: job: Generated Patterns Integration Tests (4/4) = 73s

---
## Summary by cubic
Makes all storage read/materialization paths return deep-frozen objects, so reads are immutable in both legacy and modern modes. Removes the legacy mutable-read behavior and refreezes results at the three boundary returns.

- **Bug Fixes**
  - Refreeze outputs of `applyPendingVersion` (set op), `cloneWithValueAtPath`, and `cloneWithoutPath` with deep-freeze.
  - Unify reads across flags: legacy now returns frozen objects; update the legacy committed-read test to assert frozen and add deep-freeze regression tests for path set/delete and empty-path writes.
  - Simplify `cloneWithValueAtPath` empty-path handling by relying on `cloneIfNecessary(undefined)`.

- **Migration**
  - If any code relied on mutable reads, clone before mutating (thaw-mutate-freeze).

<sup>Written for commit fcbbc4bfe9bfea5747e9ba2d1e25adc55f92947a. Summary will update on new commits.</sup>


danfuzz (commented):
**Verdict: APPROVE** — 1 NIT (formatting, non-blocking), 1 observation
(test-coverage shape, non-blocking).

Concurrent review per `code-review.md` "Concurrent review", at branch
tip `1a09aafec`. **Lens-8 at-source claim trace** applied per-site for
the three storage-internal boundaries.

## What I reviewed

Branch `coder-cedar/pr-audit-a-storage-refreeze` at `1a09aafec`,
three-dot diff `+52/-56` across 4 files: `packages/runner/src/storage/v2.ts`
(applyPendingVersion set-op return), `packages/runner/src/storage/v2-path.ts`
(cloneWithValueAtPath empty-path + spine path-leaf, cloneWithoutPath
return), `packages/runner/test/memory-v2-path-helpers.test.ts` (3 new
regression tests), `packages/runner/test/frozen-proxy-target.test.ts`
(deletion of legacy `:524-570` describe block per §A.6).

## Lens-8 at-source verification per site

**Boundary 1 — `applyPendingVersion` set-op return** (`v2.ts:275-277`).
`deepFreeze(cloneIfNecessary(pending.value, {frozen:false}))` produces
a fully-mutable deep clone, then deep-freezes. Result is a fresh
deep-frozen `EntityDocument`. The `delete` branch returns `undefined`
(N/A); the `patch` branch flows entirely through `cloneWithValueAtPath`
/ `cloneWithoutPath` returns, both now deep-frozen — so the loop's
`next` variable holds an inductively-deep-frozen value at every
iteration. ✓

**Boundary 2 — `cloneWithValueAtPath` returns** (`v2-path.ts:137-139`
empty-path; `:168` spine path-leaf). Empty-path mirrors the set-op
shape (mutable-clone → deep-freeze). Spine path: `shallowCloneContainer(baseRoot)`
is mutable; intermediate spine segments are shallow-cloned (mutable);
leaf gets `cloneIfNecessary({frozen:false})` (mutable). Then
`deepFreeze(nextRoot)` walks the spine, with off-path siblings
(retained by structural sharing through `shallowCloneContainer`)
hitting the `deepFrozenCache` short-circuit in `deep-freeze.ts:130`
(`isNecessarilyOrKnownDeepFrozen` early-return). ✓

**Boundary 3 — `cloneWithoutPath` return** (`v2-path.ts:219`). Same
shape as Boundary 2's spine path. The no-op return-`root` paths at
`v2-path.ts:194/197/202` are also frozen-safe under the inductive
invariant (callers pass already-frozen roots; n=1 production caller
`applyPendingVersion`'s patch-branch satisfies this inductively). ✓

## deepFreeze cache short-circuit verification

Per `code-design-guidance.md` "TS readonly is not runtime immutability"
and audit §4.5: confirmed the audit's load-bearing assumption holds
at-source. `deepFreeze` (`packages/data-model/deep-freeze.ts:130-150`)
short-circuits via `isNecessarilyOrKnownDeepFrozen` at line 130 — for
the structural-sharing case, off-path subtrees that are already in
`deepFrozenCache` return in O(1). The walk cost on a typical patch
is O(spine depth + new-leaf subtree size), not O(tree size). The
"primitives walked twice" concern from the freshly-cloned leaf is
also bounded — the leaf is mutable-fresh so the walk is required;
after one `deepFreeze` pass it's cached for any future read.

Audit §A.4's "F2's invariant is only observable at the primitive
level" framing is structurally correct: production caller surface is
just `v2.ts:applyPendingVersion`'s patch-branch (verified by grep —
zero external callers of either path helper). The primitive-level
tests are the right anchor.

## Import path verification

`@commonfabric/data-model/deep-freeze` — confirmed exported in
`packages/data-model/deno.json` (`"./deep-freeze": "./deep-freeze.ts"`).
Both files cleanly add a single import at the top. ✓

## §A.6 addendum trace (audit doc claims → implementation)

All four §A.6 implementation claims trace to the branch:

- "Unconditional refreeze at three boundaries" — ✓ (v2.ts:275-277,
  v2-path.ts:137-139, v2-path.ts:168, v2-path.ts:219; no flag-check,
  no `maybeRefreeze` helper exists in the branch tip).
- "Legacy describe block at frozen-proxy-target.test.ts:524-570
  deleted" — ✓ (diff shows full block removed; surrounding describes
  structurally intact at lines 515-520 and 572+; no syntactic
  stragglers).
- "Three new tests assert primitive invariant unconditionally" — ✓
  (three `expect(isDeepFrozen(result)).toBe(true)` assertions; no
  `beforeEach`/`afterEach` flag-state plumbing).
- "+52 / -56 across 4 files" — ✓ confirmed via `git diff --stat`.

Each new test anchors a distinct invariant per `code-process.md`
bug-fix testing convention: test 1 = `cloneWithValueAtPath`
non-empty-path spine refreeze (F2b spine); test 2 =
`cloneWithValueAtPath` empty-path leaf-clone refreeze (F2b
empty-path); test 3 = `cloneWithoutPath` spine refreeze (F2c). ✓

## NIT (non-blocking, take or leave)

**Stray blank line** at `memory-v2-path-helpers.test.ts:134`. The
inner `it("cloneWithoutPath returns the original root for no-op
deletes", ...)` block now ends `});\n\n});` (closing `});` of the
test, blank line, closing `});` of the outer describe). Original
file had no blank line there. One-character fix; no functional
impact. Surfacing as a nit only because deno fmt may flag it later.

## Observation (non-blocking, not a NIT)

**No direct primitive-level test for `applyPendingVersion`'s
set-op return** (F2a). The three new tests cover F2b + F2c
primitives directly. F2a's set-op shape is structurally identical
to `cloneWithValueAtPath`'s empty-path leaf-clone (both are
`deepFreeze(cloneIfNecessary(value, {frozen:false}))`), so test 2's
coverage of the empty-path case gives transitive structural
confidence in F2a's shape. Per audit §A.4 the right tests are
primitive-level, and `applyPendingVersion` is technically a
primitive too — but reaching it directly requires constructing a
`PendingVersion`, which is heavier than the path-helper tests'
shape. Not a NIT (the empty-path test covers the shape), but worth
banking for Olive's next ombudsman pass: should the
"per-discriminator-anchor-one-test" convention be tightened to
"per-fix-site-anchor-one-test" when fix-sites differ in surface
structure even if they share refreeze-shape?

## Out-of-scope (per assignment §"What's out of scope")

The framing-(1)-vs-(2)-vs-(3) arc is Dan-confirmed ship-as-is; not
reviewed for arc-shape. F1 (data-model) and F3 (schema-sanitization)
are separate scopes. Audit-doc inconsistencies (if any) go to
Olive's next ombudsman pass, not this review.

— `reviewer-flux` (Claude Opus 4.7 (1M context))

packages/runner/src/storage/v2-path.ts
```diff
@@ -131,11 +132,9 @@ export const cloneWithValueAtPath = (
   value: FabricValue | undefined,
 ): EntityDocument | undefined => {
   if (path.length === 0) {
-    return value === undefined
-      ? undefined
-      : cloneIfNecessary(value as FabricValue, {
-        frozen: false,
-      }) as EntityDocument;
+    return value === undefined ? undefined : deepFreeze(
```
danfuzz: No need to check for `undefined` at this point: `cloneIfNecessary(undefined)` will return `undefined`.

packages/runner/src/storage/v2.ts
```diff
@@ -271,9 +272,9 @@ const applyPendingVersion = (
     case "delete":
       return undefined;
     case "set":
-      return cloneIfNecessary(pending.value as FabricValue, {
-        frozen: false,
-      }) as EntityDocument;
+      return deepFreeze(
```
danfuzz, line 275: Instead of `deepFreeze(cloneIfNecessary(..., { frozen: false }))` how about just `cloneIfNecessary(...)` with the defaults of frozen-deep. I believe that is equivalent and offers the opportunity for frozen-data sharing.

packages/runner/test/frozen-proxy-target.test.ts
```diff
@@ -521,54 +521,6 @@ describe("frozen proxy target: v2 committed reads with modernDataModel ON", () =
   });
 });
 
-describe("frozen proxy target: committed reads with modernDataModel OFF (legacy)", () => {
```
danfuzz, line 524: Rather than delete this, please update it with the new contract (that is, returning deep-frozen data).

packages/runner/src/storage/v2-path.ts
danfuzz: Done in `fcbbc4bfe` — dropped the `value === undefined` check; `cloneIfNecessary(undefined as FabricValue)` returns `undefined` directly (the modern path's `tagFromNativeValue` resolves to `Primitive`, which is returned as-is per `fabric-value-modern.ts:cloneHelper`). Applied the same simplification to the path-leaf branch at the old `:165` since the shape was identical there.

— `coder-cedar` (Claude Opus 4.7 (1M context))

packages/runner/src/storage/v2.ts
danfuzz: Thanks for surfacing this; the shape you propose is the right destination, but it has an F1 dependency that the test run caught.

`cloneIfNecessary(v)` with defaults (`{frozen: true, deep: true, force: false}`) routes through `canReturnAsIs` at `packages/data-model/fabric-value-modern.ts:648`, which calls `isDeepFrozenFabricValue`. That helper throws at `:276-277` on any `FabricInstance` subclass (the F1 audit TODO — `Cannot yet handle instance of class ${value.constructor.name}`). Storage values routinely contain `FabricError` / `FabricCell` / etc. via patch payloads under modern flag, so the throw is reliably hit. The current `{frozen: false}` shape uses the flat `Object.isFrozen(v)` check at `:649` instead — no recursion, no throw.

Empirical signal when I applied the suggested shape across the three boundaries: 10 failures + uncaught error in `cell-core.test.ts` against `packages/runner`'s suite.

Reverting the Comment 2 edits in `fcbbc4bfe` and keeping the `deepFreeze(cloneIfNecessary(..., { frozen: false }))` shape. The identity-sharing optimization is the right shape to land once F1 lands (per the planned PR-AUDIT-C scope on `FabricInstance` per-class freeze-in-constructor). Tracking it as a follow-up so the architectural insight here doesn't get lost.

— `coder-cedar` (Claude Opus 4.7 (1M context))

packages/runner/test/frozen-proxy-target.test.ts
danfuzz: Done in `fcbbc4bfe` — restored the legacy describe block at `frozen-proxy-target.test.ts:524` with the inverted assertion. Under the new F2 contract, the `modernDataModel: false` (legacy) read path also returns deep-frozen committed objects, witnessing the unified frozen-data-through-the-system invariant across both flag states. Test count went from 450 → 451.

— `coder-cedar` (Claude Opus 4.7 (1M context))

packages/runner/src/storage/v2.ts
danfuzz: Cross-reference (per the perf-note in the refreshed PR body): the F1 dependency named above is also load-bearing on this PR's perf signal. F1 lands the frozen-data-sharing fast path that Comment 2 was reaching for; the +12s slowdown flagged on `Generated Patterns Integration Tests (4/4)` is expected to recover once F1's protocol extension is in. The deferred optimization and the deferred perf-recovery are the same arc.

— `coder-cedar` (Claude Opus 4.7 (1M context))

PR #3589, opened 2026-05-14
Derive prompt-slot CFC input labels

mathpirate:
## Summary
- derive confidentiality-only prompt-slot influence labels into cf-harness invocation contexts
- let tools declare which invocation input paths are model/prompt-influenced (`command`, `args`, `stdin`)
- merge derived labels with explicit trusted `cfcInputLabels` and document the behavior

## Notes
- These are taint/influence labels, not integrity or authorization claims. The merged gVisor side rejects integrity claims in `cfcInputLabels`.
- Docker/runsc sidecar-before-start transport is covered by the merged invocation-context work. The remaining runtime/Fabric gap is durable sandbox-to-FUSE write/readback via trusted prepare/finalize metadata and its atomicity semantics.

## Tests
- `/Users/gideonwald/.deno/bin/deno check packages/cf-harness/src/contracts/cfc-invocation-context.ts packages/cf-harness/src/engine.ts packages/cf-harness/src/tools/bash.ts packages/cf-harness/src/tools/read-file.ts packages/cf-harness/src/tools/write-file.ts packages/cf-harness/src/tools/edit-file.ts packages/cf-harness/test/cfc-invocation-context.test.ts packages/cf-harness/test/engine.test.ts`
- `/Users/gideonwald/.deno/bin/deno test --allow-env --allow-read --allow-run --allow-write packages/cf-harness/test/cfc-invocation-context.test.ts packages/cf-harness/test/engine.test.ts`
- `(cd packages/cf-harness && /Users/gideonwald/.deno/bin/deno task test)`
- `git diff --check`

### DIFF
--- a/packages/cf-harness/README.md
+++ b/packages/cf-harness/README.md
@@ -427,6 +427,11 @@ payload contains audit/provenance context plus optional trusted `cfcInputLabels`
 for supported startup inputs (`command`, `argv`, `args`, `env`, `cwd`, and
 `stdin`). `stdin` labels are modeled as labels on the stdin source and taint
 only after the sandbox reads or maps fd 0, not as automatic startup task taint.
+When a trusted prompt-slot binding is present, `cf-harness` also derives
+confidentiality-only prompt influence labels for model-authored invocation
+inputs such as shell commands, structured file-tool arguments, and stdin
+payloads. These labels are taint evidence, not integrity or authorization
+claims.
 
 On Docker Desktop for macOS, use the host path for `cf-harness` and the
 `/host_mnt/...` projection for Docker's runtime args. The gVisor
--- a/packages/cf-harness/src/contracts/cfc-invocation-context.ts
+++ b/packages/cf-harness/src/contracts/cfc-invocation-context.ts
@@ -2,12 +2,46 @@ import type {
   CfcEnforcementMode,
   CfcLabelView,
 } from "@commonfabric/runner/cfc";
-import type { PromptSlotBinding } from "./prompt-slot.ts";
+import { mergeCfcLabelViews } from "@commonfabric/runner/cfc";
+import type { PromptSlotBinding, PromptSlotRole } from "./prompt-slot.ts";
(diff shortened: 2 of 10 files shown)


willkelly (approved):
This lines up with the CFC direction we just landed.

What I checked:
- `inputs` remain redacted provenance summaries; taint still enters through `cfcInputLabels`.
- `PromptSlotBound` is not treated as a label by itself; cf-harness is doing an explicit trusted derivation into `cfcInputLabels` for selected model-authored invocation inputs.
- The derived `cf-harness.cfc/PromptSlotInfluence` atom is confidentiality-only, and the tests assert it does not mint integrity/authorization claims.
- Explicit trusted `cfcInputLabels` are merged rather than overwritten.
- `stdin` labeling is consistent with the runsc/gVisor model where stdin is a source label and taints on read/mmap, not automatic startup task taint.
- This PR does not make any new FUSE durable-writeback / prepare-finalize claims, which is important because that remains a separate atomicity/design gap.

The one thing I would keep an eye on is `bash.cwd`: `bash` currently derives prompt-slot influence for `command` only. Since `cwd` is also a supported invocation input root in the CFC sidecar, should model-selected `cwd` be included in `cfcInputLabelPaths` when present? If cwd is intentionally runtime/context-controlled rather than model-authored, this is fine; if the model can choose it, omitting it looks like a small under-taint case.

Also minor/status-only: the PR body says Docker/runsc invocation sidecar timing remains a gVisor/runtime follow-up. I think that note is stale after the merged gVisor/Labs work and the current report: sidecar-before-start transport is covered now; the still-open gap is durable sandbox→FUSE write/readback via trusted prepare/finalize and its atomicity semantics.

I reran the focused cf-harness tests on the PR head with the expected Deno permissions:

`deno test --allow-env --allow-read --allow-run --allow-write packages/cf-harness/test/cfc-invocation-context.test.ts packages/cf-harness/test/engine.test.ts`

Result: 19 passed, 0 failed.

PR #3594, opened 2026-05-14
[codex] fix ts-transformers inline handler sends

mathpirate:
## Summary

Fixes CT-1589 by preserving event-handler semantics for inline JSX handlers that call a bound stream `.send(...)` from inside reactive map/conditional JSX contexts.

The handler closure builder moves the authored handler body into a generated handler callback, but those body nodes can retain original parent links from the authored JSX arrow. Later expression-site rewriting could then classify the moved body as pattern-owned array-method computation and wrap `.send(...)` in `derive(...)`. This marks both the authored handler arrow and the generated handler callback as safe handler contexts, preventing that rewrite from treating handler bodies as reactive value expressions.

## Validation

- `deno task cf check packages/patterns/cozy-poll-scoped/main.tsx --show-transformed` now emits bare `bound*.send(...)` handler bodies and no `__cfModuleCallback` hits for the vote/remove buttons.
- Added fixture `map-conditional-inline-handler-send` covering the failing shape.
- `FIXTURE=map-conditional-inline-handler-send deno test --allow-read --allow-write --allow-env test/fixture-based.test.ts`
- `deno task check` from `packages/ts-transformers`
- `deno task test` from `packages/ts-transformers` (`259 passed`, `613 steps`)

Note: `packages/patterns/scope-bug-lambda-in-map/main.tsx` still exits with its pre-existing duplicate `id` SES error, but the transformed output has no `__cfModuleCallback` hits.

---
## Summary by cubic
Preserves event-handler semantics for inline JSX handlers that call a bound stream `.send(...)` inside reactive map/conditional contexts. Fixes CT-1589 by preventing these sends from being wrapped in `derive(...)`.

- **Bug Fixes**
  - Mark both the authored handler arrow and the generated handler callback as safe handler contexts in `packages/ts-transformers/src/closures/utils/capture-scaffold.ts` to stop expression-site rewrites from wrapping `.send(...)`.
  - Added the `map-conditional-inline-handler-send` fixture to lock the behavior; transformed output now emits bare `bound*.send(...)`.

<sup>Written for commit dfd08a62d6e19662b78e22921607a5d3985fdc35. Summary will update on new commits.</sup>

### DIFF
--- a/packages/ts-transformers/src/closures/utils/capture-scaffold.ts
+++ b/packages/ts-transformers/src/closures/utils/capture-scaffold.ts
@@ -33,12 +33,16 @@ export function buildCapturedHandlerClosureCall(
   const builder = new PatternBuilder(context);
   builder.setCaptureTree(captureTree);
 
+  // The moved body keeps original parent links, so mark both callback nodes as
+  // safe handler contexts for later expression-site rewrites.
+  context.markAsSyntheticComputeCallback(callback);
   const handlerCallback = builder.buildHandlerCallback(
     callback,
     transformedBody,
     options?.eventParamName ?? "event",
     options?.paramsParamName ?? "params",
   );
+  context.markAsSyntheticComputeCallback(handlerCallback);
 
   const handlerCall = context.cfHelpers.createHelperCall(
     "handler",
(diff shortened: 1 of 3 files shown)


mathpirate:
Superseded by #3595, which carries the CT-1589 transformer fix on a fresh branch off `main` so the normal PR CI trigger runs.

PR #3596, opened 2026-05-14
Track CFC model context labels

mathpirate:
## Summary

- add a persisted `cf-harness.cfc-model-context` run-state accumulator for CFC-observed model context
- record labels from CFC-mediated bash stdout/stderr/exitCode only when those channels are released to the model
- merge accumulated confidentiality labels into later model-authored invocation inputs, while preserving sibling tool-call turn boundaries
- document the model-context continuity behavior in the cf-harness README

## Tests

- `/Users/gideonwald/.deno/bin/deno check packages/cf-harness/src/contracts/cfc-model-context.ts packages/cf-harness/src/contracts/cfc-invocation-context.ts packages/cf-harness/src/run-state.ts packages/cf-harness/src/engine.ts packages/cf-harness/src/prompt-loop.ts packages/cf-harness/test/prompt-loop.test.ts`
- `/Users/gideonwald/.deno/bin/deno test --allow-env --allow-read --allow-run --allow-write packages/cf-harness/test/prompt-loop.test.ts packages/cf-harness/test/engine.test.ts packages/cf-harness/test/cfc-invocation-context.test.ts`
- `/Users/gideonwald/.deno/bin/deno task test` in `packages/cf-harness`
- `git diff --cached --check`

---
## Summary by cubic
Adds model-context label tracking to `cf-harness`. When CFC-mediated bash output is released to the model, observed confidentiality labels are persisted and applied to later model-authored tool inputs; opaque/denied outputs are ignored.

- **New Features**
  - Added `cf-harness.cfc-model-context` run-state accumulator with an observation log and merged confidentiality-only label.
  - Captures labels from CFC-observed `stdout`/`stderr`/`exitCode` only when those channels are exposed to the model; tracks truncation.
  - Merges accumulated labels into subsequent invocation `cfcInputLabels` via `cfc-invocation-context`, without crossing sibling tool-call boundaries in the same assistant turn.
  - Exposed `engine.recordCfcModelContextObservations`, new helpers and types in `contracts/cfc-model-context.ts`, and wired them through the engine, prompt loop, and `createHarnessCfcInvocationContext`.
  - Updated README to document continuity and note the model-context accumulator is sensitive metadata (labels/refs only, no raw bytes).

<sup>Written for commit 3e68dc48da5b5139d2f32d54aaf9952d074c6848. Summary will update on new commits.</sup>

### DIFF
--- a/packages/cf-harness/README.md
+++ b/packages/cf-harness/README.md
@@ -436,6 +436,12 @@ records the observed output labels in run state and merges those confidentiality
 labels into later model-authored invocation inputs. Opaque and denied outputs
 are not added to this model-context accumulator.
 
+The persisted model-context accumulator is sensitive retained run metadata. It
+does not store raw stdout/stderr bytes, but its labels and observation refs can
+still disclose which confidential sources influenced model-visible context.
+Handle it under the same access and retention boundary as transcripts, tool
+outputs, run state, and CFC policy traces.
+
 On Docker Desktop for macOS, use the host path for `cf-harness` and the
 `/host_mnt/...` projection for Docker's runtime args. The gVisor
 `docker-desktop-cfc-setup` helper defaults to:
--- a/packages/cf-harness/README.md
+++ b/packages/cf-harness/README.md
@@ -431,7 +431,10 @@ When a trusted prompt-slot binding is present, `cf-harness` also derives
 confidentiality-only prompt influence labels for model-authored invocation
 inputs such as shell commands, structured file-tool arguments, and stdin
 payloads. These labels are taint evidence, not integrity or authorization
-claims.
+claims. When CFC-mediated bash output is released to the model, `cf-harness`
+records the observed output labels in run state and merges those confidentiality
+labels into later model-authored invocation inputs. Opaque and denied outputs
+are not added to this model-context accumulator.
 
 On Docker Desktop for macOS, use the host path for `cf-harness` and the
 `/host_mnt/...` projection for Docker's runtime args. The gVisor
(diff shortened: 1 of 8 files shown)


willkelly (approved):
This looks like the right shape to me, and I’m approving it.

The important semantic point here is that once CFC-mediated output has actually been released to the model, the model’s later tool inputs are no longer independent of that output. Carrying the observed output’s confidentiality labels forward into later `cfcInputLabels` is exactly the conservative accounting we want: the model saw something, may have reasoned from it, and subsequent model-authored commands/args/stdin/cwd should reflect that influence.

I especially like the boundaries in this version:

- only `observed` CFC-mediated bash channels are accumulated;
- opaque and denied channels are not treated as model context;
- sibling tool calls from the same assistant message are not retroactively tainted, which matches causality because the model generated those calls before seeing any sibling output;
- the accumulated label is confidentiality-only and flows through `cfcInputLabels`, not through provenance fields or integrity/authorization claims;
- `bash` now covers model-selected `cwd` as well as `command`, which matches the gVisor input-label surface.

The one caveat I’d keep explicit is that `cfcModelContext` is sensitive metadata, not harmless bookkeeping. Even if it does not store raw stdout/stderr bytes, the labels and observations can reveal facts like “this run/model saw output influenced by Alice’s secret” or “this process read a private source and exposed an observed result to the model.” That can itself leak information if run-state artifacts are shown outside the same trusted/audience boundary as the transcript, tool outputs, and CFC policy trace.

So I don’t think this blocks the PR, but I do think we should treat persisted `cfcModelContext` observations with transcript-equivalent (or stricter) access/retention expectations. As long as that is the intended boundary, this aligns with the CFC model-context taint story.

I also reran the focused cf-harness tests locally on the PR head:

`deno test --allow-env --allow-read --allow-run --allow-write packages/cf-harness/test/prompt-loop.test.ts packages/cf-harness/test/engine.test.ts packages/cf-harness/test/cfc-invocation-context.test.ts`

Result: 56 passed, 0 failed.

PR #3601, opened 2026-05-15
fix(ts-transformers): hoist synthesized pattern callbacks that close over module-level symbols (CT-1585)

mathpirate:
## Summary

Fixes [CT-1585](https://linear.app/common-tools/issue/CT-1585/map-callback-pattern-that-closes-over-module-level-symbols). When `.map(cb)` is rewritten by the closure transformer to `.mapWithPattern(__cfHelpers.pattern(synthesizedCallback), {})` and the synthesized callback closes only over module-level symbols, hoist the callback to module scope as `const __cfModuleCallback_N = ...`.

```ts
// Source — closes only over EntryRow (module-scoped) and UI (module-imported)
{filtered.map((entry) => {
  const row = EntryRow({ piece: entry.piece, ... });
  return row[UI];
})}

// After
const __cfModuleCallback_1 = __cfHardenFn(__cf_pattern_input => {
  const entry = __cf_pattern_input.key("element");
  const row = EntryRow({ piece: entry.key("piece"), ... });
  return row.key(__cfHelpers.UI);
});

// ... and the call site references the hoisted name:
{filtered.mapWithPattern(__cfHelpers.pattern(__cfModuleCallback_1, ...))}
```

## Five coordinated changes

Each commit corrects one part of the hoisting story. Together they make the hoister's behavior match the documented intent in `docs/closure-design.md`.

### 1. `localNames` pre-pass in `analyzeCallbackForHoisting`

The existing `hoistModuleScopedBuilderCallbacks` pass had `pattern` in its `HOISTABLE_BUILDER_NAMES` but never fired on transformer-synthesized pattern arrows. Root cause: `analyzeCallbackForHoisting` resolves identifier scopes via the TypeScript checker. For synthesized identifiers (pos = -1) the checker falls back to `ts.getOriginalNode(...)` and resolves to the source-AST declaration of the original `entry` parameter from `.map((entry) => ...)`. Then `isDeclaredWithinFunction(sourceDecl, synthesizedCallback)` fails because the synthesized callback doesn't appear in the source AST's parent chain — exactly the hazard [documented at `ast/scope-analysis.ts:70-86`](https://github.com/commontoolsinc/labs/blob/main/packages/ts-transformers/src/ast/scope-analysis.ts#L70). The locals get bucketed as `enclosing` and `canHoist = false`.

Fix: pre-compute a `Set<string>` of names declared locally in the callback (parameter binding targets + body var declarations, stopping at nested function boundaries). Fast-path-classify any identifier whose name is in that set as `local` without consulting `getReferenceScope`. Sidesteps the synthetic-node hazard while leaving the symbol-based resolution intact for everything else.

### 2. New `BuilderCallbackHoistingTransformer` stage after `PatternCallbackLoweringTransformer`

After (1) alone, the hoist fires correctly — but pattern callbacks were being hoisted *before* stage 11 had a chance to introduce the `__cf_pattern_input.key("…")` prologue and substitute bare CF keys (`c[NAME]` → `c.key(__cfHelpers.NAME)`). Stage 11 only visits pattern callbacks that are still at their call site as function-like arguments; once hoisted, they become identifier references that stage 11 silently skips. Bare `NAME` identifiers would survive in module scope — breaking the helper-backed invariant CT-1586 (#3600) protects.

Fix: extract `hoistModuleScopedBuilderCallbacks` into its own `BuilderCallbackHoistingTransformer` stage scheduled immediately after `PatternCallbackLoweringTransformer`. Pattern callbacks get fully lowered in-place — prologue, key substitutions, tracked-opaque rewrites — and *then* hoisted.

### 3. Assign explicit numeric suffixes to hoisted callback names (root-cause for CI runtime failures)

CI surfaced a runtime regression in three pattern tests (`notes/note.test.tsx`, `notes/notebook.test.tsx`, `factory-outputs/parking-coordinator/main.test.tsx`): `TypeError: Cannot read properties of undefined (reading 'push' | 'set' | 'get')`. Investigation found a latent identifier-resolution bug that CT-1585's hoist exposed.

**Root cause**: the hoister used `context.factory.createUniqueName("__cfModuleCallback")` to generate unique names. That helper returns identifiers whose `.text` is the bare prefix (`"__cfModuleCallback"`); the numeric suffixes (`_1`, `_2`, …) are synthesized only by the TypeScript emitter at print time. Downstream stages that match identifiers by `.text` see every hoisted-callback identifier as the same string.

`SchemaInjectionTransformer.getSyntheticModuleCallbackInitializer` (`schema-injection.ts:2335`) is one such consumer: it walks the source file's top-level variable statements and returns the first declaration whose `name.text === identifier.text`. With `createUniqueName`, *every* call-site identifier matches *only the first* hoisted declaration. Capability summaries from the *first* hoisted callback get applied to *every* hoisted callback's call site.

In the failing patterns, one factory body had multiple `action()` callbacks — one reading a captured cell (`items.get()`), one writing it (`items.push()`). Post-hoist schema-injection looked up both call sites against the *first* hoisted callback's summary, classifying the writer's cell as readonly and dropping captures the writer used (e.g. `self`, `notes`). At runtime, those captures weren't injected into the params and `.push(...)` / `.set(...)` threw.

**Fix**: replace `createUniqueName` with an explicit per-source-file counter + `createIdentifier`. The identifier's `.text` now carries the true unique name, so identity-by-text lookups across the pipeline resolve correctly.

Notably, this fix also corrects two pre-existing fixtures (`action-partial.expected.jsx`, `action-self-closure.expected.jsx`) whose goldens reflected the same latent corruption on `main`. The fixture diffs in those files are *improvements*: schemas that were polluted by sibling callbacks now reflect each callback's own capability usage.

### 4. Exclude ambient globals and transformer scaffolding from the "module-level" signal (Berni's review)

PR reviewer @seefeldb pointed out that the hoister was firing on callbacks that don't materially close over user-authored module-level symbols. Two fixture examples:

  - `cell-map-with-captures`: hoisted body's only "module-level" reference is the transformer-injected `__cfHelpers.derive(...)` wrapper around the JSX body.
  - `action-self-closure`: hoisted body's only "module-level" reference is `console.log` — an ambient global.

Auditing the full closure-fixture corpus surfaced this as a systematic issue: ~70% of hoists (59/85 callbacks across 77 fixtures) were over-triggers in this same class.

**Fix**: in `getReferenceScope`, classify both classes as `"other"` rather than `"module"`:

  - For transformer helpers: check `node.text` against `CF_HELPERS_IDENTIFIER` (`__cfHelpers`) and the `FUNCTION_HARDENING_HELPER_NAME` prefix (`__cfHardenFn*`) before the symbol-resolution path.
  - For ambient globals: after symbol resolution, check whether every declaration lives in a `.d.ts` source file. If so, treat as `"other"`. This matches the existing precedent at `expression-site-policy.ts:203` which uses the same signal.

Result: 65 fixture goldens regenerated. ~59 now correctly emit the callback inline (it never deserved to be hoisted). The legitimate hoists are unchanged.

### 5. Hoist synthetic 2-arg derive callbacks that DO close over real user module symbols

While auditing change #4, surfaced a complementary under-trigger. `getBuilderCallbackIndices` required `hasSelfDescribingFunctionTypes` for 2-arg `derive(input, cb)` calls. The synthetic derives produced by the closure transformer (with bodies like `({ item }) => formatPrice(item.price * (1 + TAX_RATE))`) have destructured params with no type annotations, so they always failed this check and never reached `analyzeCallbackForHoisting`. The 4-arg derive form *would* bypass this check, but that shape only exists post-schema-injection — and the hoister runs *before* SchemaInjection, so the 4-arg branch was dead code at this pipeline position.

The type-annotation requirement exists as a disambiguator for *user-authored* 2-arg `derive(input, cb)` calls (where the explicit type signals a stable structural contract). For synthetic transformer-produced derives, that signal isn't needed — the transformer itself knows the callback's shape and tags it via `context.markAsSyntheticComputeCallback`.

**Fix**: thread `context` into `getBuilderCallbackIndices` and check `context.isSyntheticComputeCallback(deriveCallback)` for the 2-arg case. If the callback is synthetic-compute, accept index 1 without the type-annotation guard. User-authored 2-arg derives keep the existing guard.

**Plus a visit-order safety net**: `ts.visitEachChild` recurses inner-first, so an inner derive can now be hoisted *before* its outer pattern callback is analyzed. The outer callback's body would then contain a `__cfModuleCallback_N` reference. Without excluding it, that reference would count as "uses module-scoped references" and incorrectly promote the outer callback to hoistable too. `isTransformerInjectedIdentifier` (introduced in change #4) is extended to cover `__cfModuleCallback*` (the hoister's own output) for this reason.

Result: ~22 legitimate hoists that were always supposed to fire are added back (e.g. the `map-import-reference` `formatPrice/TAX_RATE` case).

## Test plan

- [x] New unit test `CT-1585: Closure Transformer hoists synthesized mapWithPattern callback...` in `test/closures/module-scope-helper-hoisting.test.ts` (change #1)
- [x] New fixture `hoisted-handler-preserves-capture-schemas` pins the minimal regression for change #3: two action callbacks (one reader, one writer) on the same captured cell, with assertions that each gets its own correct capability classification
- [x] Pipeline stage-order regression test updated to include the new stage
- [x] All three previously-failing pattern runtime tests now pass locally:
  - `packages/patterns/notes/note.test.tsx`: 29 passed, 0 failed
  - `packages/patterns/notes/notebook.test.tsx`: 28 passed, 0 failed
  - `packages/patterns/factory-outputs/parking-coordinator/main.test.tsx`: 38 passed, 0 failed
- [x] Full ts-transformers test suite: 281/281 passing
- [x] Typecheck, lint (package + root), fmt clean

## Risk audit

**Identifier collision (change #3)**: The fix swaps `createUniqueName` for an explicit counter + `createIdentifier`. The hoister maintains its counter per-source-file scope, so within a file every hoisted name is distinct.

**Pre-existing fixture diffs**: A handful of pre-existing fixtures had goldens that were quietly carrying corrupted output from the latent bugs we fixed. The diff in each is a correction:
  - Change #3 corrected `action-partial.expected.jsx` and `action-self-closure.expected.jsx` (schemas polluted by sibling-callback capability summaries).
  - Change #4 corrected ~59 fixtures that hoisted purely on transformer-injected or ambient-global references.
  - Change #5 added ~22 hoists that were always supposed to fire but were silently missed.

**Shadowing risk in `localNames` pre-pass (change #1)**: bounded. Locals shadowing module-level or enclosing names produce the same result as the pre-fix slow path (the inner local wins per JS scope rules). The fast-path only diverges when symbol resolution would have failed on a synthetic node — the target case.

**Stage interactions (change #2)**: verified.
  - **Stage 12 (SchemaInjection)** handles the post-hoist call site `__cfHelpers.pattern(__cfModuleCallback_N)` correctly — schemas attach as args 2 and 3 of the helper call, indifferent to whether the first arg is an inline arrow or an identifier reference (now correctly distinguished thanks to change #3).
  - **Stage 14 (ReactiveVariableFor)** attaches `.for("__patternResult", true)` at the outer `mapWithPattern(...)` call site, not inside the callback. Hoist doesn't affect this.

**Visit-order interaction in change #5**: `ts.visitEachChild` is inner-first, so inner hoists land before outer analysis. The transformer-injected suppression list now covers the hoister's own `__cfModuleCallback*` output, keeping inner and outer hoist decisions independent.

## File count note

Total ~100 files changed across the five commits, but the cognitive review surface is just ~8 source/test files: `cf-pipeline.ts`, `closures/transformer.ts`, `closures/module-scope-callback-hoisting.ts`, new `transformers/builder-callback-hoisting.ts`, `module-scope-helper-hoisting.test.ts`, the stage-order line in `pipeline-regressions.test.ts`, and the input/expected of `hoisted-handler-preserves-capture-schemas`. The remaining ~90 are fixture goldens that mechanically reflect "callback moved from inline to a named module-scope const" (or vice versa) — evidence of how widespread each bug was, not new logic to review.

### DIFF
--- a/packages/ts-transformers/test/fixtures/closures/action-self-closure.expected.jsx
+++ b/packages/ts-transformers/test/fixtures/closures/action-self-closure.expected.jsx
@@ -72,16 +72,34 @@ export default pattern((__cf_pattern_input) => {
         type: "object",
         properties: {
             self: {
+                $ref: "#/$defs/TestOutput"
+            },
+            count: {
+                type: "number",
+                asCell: ["cell"]
+            }
+        },
+        required: ["self", "count"],
+        $defs: {
+            TestOutput: {
                 type: "object",
                 properties: {
                     title: {
                         type: "string"
+                    },
+                    count: {
+                        type: "number"
+                    },
+                    $NAME: {
+                        type: "string"
+                    },
+                    $UI: {
+                        $ref: "https://commonfabric.org/schemas/vnode.json"
                     }
                 },
-                required: ["title"]
+                required: ["title", "count", "$NAME", "$UI"]
             }
-        },
-        required: ["self"]
+        }
     } as const satisfies __cfHelpers.JSONSchema, __cfModuleCallback_2)({
         self: self,
         count: count
--- a/packages/ts-transformers/test/fixtures/closures/action-self-closure.expected.jsx
+++ b/packages/ts-transformers/test/fixtures/closures/action-self-closure.expected.jsx
@@ -15,6 +15,13 @@ import { action, type Default, NAME, pattern, SELF, UI, type VNode, Writable } f
 const define = undefined;
 const runtimeDeps = undefined;
 const __cfAmdHooks = undefined;
+const __cfModuleCallback_1 = __cfHardenFn((_, { self }) => {
+    console.log("self.title:", self.title);
+});
+const __cfModuleCallback_2 = __cfHardenFn((_, { self, count }) => {
+    console.log("self:", self);
+    count.set(count.get() + 1);
+});
 interface TestOutput {
     [NAME]: string;
     [UI]: VNode;
@@ -51,9 +58,7 @@ export default pattern((__cf_pattern_input) => {
             }
         },
         required: ["self"]
-    } as const satisfies __cfHelpers.JSONSchema, (_, { self }) => {
-        console.log("self.title:", self.title);
-    })({
+    } as const satisfies __cfHelpers.JSONSchema, __cfModuleCallback_1)({
         self: {
             title: self.key("title")
         }
@@ -67,38 +72,17 @@ export default pattern((__cf_pattern_input) => {
         type: "object",
         properties: {
             self: {
-                $ref: "#/$defs/TestOutput"
-            },
-            count: {
-                type: "number",
-                asCell: ["cell"]
-            }
-        },
-        required: ["self", "count"],
-        $defs: {
-            TestOutput: {
                 type: "object",
                 properties: {
                     title: {
                         type: "string"
-                    },
-                    count: {
-                        type: "number"
-                    },
-                    $NAME: {
-                        type: "string"
-                    },
-                    $UI: {
-                        $ref: "https://commonfabric.org/schemas/vnode.json"
                     }
                 },
-                required: ["title", "count", "$NAME", "$UI"]
+                required: ["title"]
             }
-        }
-    } as const satisfies __cfHelpers.JSONSchema, (_, { self, count }) => {
-        console.log("self:", self);
-        count.set(count.get() + 1);
-    })({
+        },
+        required: ["self"]
+    } as const satisfies __cfHelpers.JSONSchema, __cfModuleCallback_2)({
         self: self,
         count: count
     }).for({ stream: "incrementWithSelf" }, true);
--- a/packages/ts-transformers/test/fixtures/closures/action-self-closure.expected.jsx
+++ b/packages/ts-transformers/test/fixtures/closures/action-self-closure.expected.jsx
@@ -15,13 +15,6 @@ import { action, type Default, NAME, pattern, SELF, UI, type VNode, Writable } f
 const define = undefined;
 const runtimeDeps = undefined;
 const __cfAmdHooks = undefined;
-const __cfModuleCallback_1 = __cfHardenFn((_, { self }) => {
-    console.log("self.title:", self.title);
-});
-const __cfModuleCallback_2 = __cfHardenFn((_, { self, count }) => {
-    console.log("self:", self);
-    count.set(count.get() + 1);
-});
 interface TestOutput {
     [NAME]: string;
     [UI]: VNode;
@@ -58,7 +51,9 @@ export default pattern((__cf_pattern_input) => {
             }
         },
         required: ["self"]
-    } as const satisfies __cfHelpers.JSONSchema, __cfModuleCallback_1)({
+    } as const satisfies __cfHelpers.JSONSchema, (_, { self }) => {
+        console.log("self.title:", self.title);
+    })({
         self: {
             title: self.key("title")
         }
@@ -100,7 +95,10 @@ export default pattern((__cf_pattern_input) => {
                 required: ["title", "count", "$NAME", "$UI"]
             }
         }
-    } as const satisfies __cfHelpers.JSONSchema, __cfModuleCallback_2)({
+    } as const satisfies __cfHelpers.JSONSchema, (_, { self, count }) => {
+        console.log("self:", self);
+        count.set(count.get() + 1);
+    })({
         self: self,
         count: count
     }).for({ stream: "incrementWithSelf" }, true);
--- a/packages/ts-transformers/test/fixtures/closures/cell-map-with-captures.expected.jsx
+++ b/packages/ts-transformers/test/fixtures/closures/cell-map-with-captures.expected.jsx
@@ -11,31 +11,35 @@ import { Cell, cell, pattern, UI } from "commonfabric";
 const define = undefined;
 const runtimeDeps = undefined;
 const __cfAmdHooks = undefined;
-const __cfModuleCallback_1 = __cfHardenFn(({ element: value, params: { state } }) => (<span>{__cfHelpers.derive({
-    type: "object",
-    properties: {
-        value: {
-            type: "number"
-        },
-        state: {
-            type: "object",
-            properties: {
-                multiplier: {
-                    type: "number"
-                }
+const __cfModuleCallback_1 = __cfHardenFn(__cf_pattern_input => {
+    const value = __cf_pattern_input.key("element");
+    const state = __cf_pattern_input.key("params", "state");
+    return (<span>{__cfHelpers.derive({
+        type: "object",
+        properties: {
+            value: {
+                type: "number"
             },
-            required: ["multiplier"]
+            state: {
+                type: "object",
+                properties: {
+                    multiplier: {
+                        type: "number"
+                    }
+                },
+                required: ["multiplier"]
+            }
+        },
+        required: ["value", "state"]
+    } as const satisfies __cfHelpers.JSONSchema, {
+        type: "number"
+    } as const satisfies __cfHelpers.JSONSchema, {
+        value: value,
+        state: {
+            multiplier: state.key("multiplier")
         }
-    },
-    required: ["value", "state"]
-} as const satisfies __cfHelpers.JSONSchema, {
-    type: "number"
-} as const satisfies __cfHelpers.JSONSchema, {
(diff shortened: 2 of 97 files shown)


packages/ts-transformers/test/fixtures/closures/cell-map-with-captures.expected.jsx
```diff
@@ -11,31 +11,35 @@ import { Cell, cell, pattern, UI } from "commonfabric";
 const define = undefined;
 const runtimeDeps = undefined;
 const __cfAmdHooks = undefined;
-const __cfModuleCallback_1 = __cfHardenFn(({ element: value, params: { state } }) => (<span>{__cfHelpers.derive({
-    type: "object",
-    properties: {
-        value: {
-            type: "number"
+const __cfModuleCallback_1 = __cfHardenFn(__cf_pattern_input => {
```
seefeldb: i might be missing it, but i don't see what this closes over that is defined at module scope?

packages/ts-transformers/test/fixtures/closures/action-self-closure.expected.jsx
```diff
@@ -15,6 +15,13 @@ import { action, type Default, NAME, pattern, SELF, UI, type VNode, Writable } f
 const define = undefined;
 const runtimeDeps = undefined;
 const __cfAmdHooks = undefined;
+const __cfModuleCallback_1 = __cfHardenFn((_, { self }) => {
+    console.log("self.title:", self.title);
```
seefeldb: console.log doesn't count as closing over a module level symbol as it is globally defined.
(or otherwise: not clear what this is closing over? could have been left inline)

packages/ts-transformers/test/fixtures/closures/cell-map-with-captures.expected.jsx
mathpirate: Good catch — you're exactly right. The only "module-level" reference in this callback is the transformer-injected `__cfHelpers.derive(...)` wrapper around the JSX body. Nothing user-authored.

An audit of all 77 closure-fixture files surfaced this as a systematic issue: 51 callbacks were being hoisted purely because of the synthesized `__cfHelpers.*` reference, plus 6 more (in `action-*`, `handler-*`) because of `console.log`. ~70% of CT-1585's hoists across the corpus were false positives in this same class.

Fixed in `7527e04db`: `getReferenceScope` now classifies (a) transformer-injected scaffolding (`__cfHelpers`, `__cfHardenFn*`) and (b) ambient globals (declarations entirely in `lib.*.d.ts`) as `"other"` rather than `"module"`. This fixture now correctly stays inline.

While auditing, also discovered a complementary under-trigger: synthetic 2-arg derives like `({ item }) => formatPrice(item.price * (1 + TAX_RATE))` were being skipped even when they *did* close over real user-authored module symbols, because of an explicit-param-type requirement that was meant for user-authored 2-arg derives. Fixed in `4f96419da`: synthetic-compute callbacks (tagged via `markAsSyntheticComputeCallback`) bypass the type-annotation guard. Adds back ~22 legitimate hoists that were always supposed to fire.

The PR now has the hoist's precision corrected in both directions.

packages/ts-transformers/test/fixtures/closures/action-self-closure.expected.jsx
mathpirate: You're right on both counts — `console` shouldn't qualify (it's an ambient global, not a module binding), and these two action callbacks otherwise close over only `self` and `count` (enclosing scope).

Same fix as the other comment: in `7527e04db`, `getReferenceScope` now treats ambient globals (declarations entirely in `lib.*.d.ts` files) as `"other"` rather than `"module"`. Both `__cfModuleCallback_1` and `__cfModuleCallback_2` in this fixture now correctly stay inline.

Also corrected another pre-existing fixture along the way (`action-partial`) which had the same `console.log`-only-trigger pattern.

PR #3602, opened 2026-05-15
fix(cli): propagate caller identity into piece inspect scope context (CT-1598)

jkomoros:
## Summary

`cf piece inspect --identity <key>` was reading `PerUser`/`PerSession` fields as the default (e.g. `""`) regardless of which identity ran the inspect, even when handlers had clearly written per-user values via `cf piece call`. The caller's identity was used to authenticate the connection but never reached per-property scope resolution.

**Root cause:** `inspectPiece` in `packages/cli/lib/piece.ts` read each top-level field via `piece.input.get()` / `piece.result.get()`, which returns whatever is stored *inline* in the parent document at the cell's effective scope. For per-property-scoped fields on an otherwise space-scoped argument, the inline value is the default — the real per-user value lives in a separate scoped storage instance (`scope_key = "user:<DID>"`).

**Fix:** Descend one level via `cell.key(prop)` per top-level property. `cell.key()` already carries the property schema's declared scope into the child cell (see `Cell.key()` in `packages/runner/src/cell.ts:1171-1173`), so calling `.get()` on the child cell correctly resolves the scoped storage instance for the active identity.

## Why this is small

- One helper (`readWithPropertyScope`) of ~15 lines plus two call sites swapped over.
- Pure read-side change in the CLI. No protocol or schema changes.
- Same scope-aware behavior the runtime already uses when handlers access bound cells in their context maps.

## Verified against `packages/patterns/cozy-poll-scoped/main.tsx`

- **Before:** `myName: ""` for every identity.
- **After:** `myName: "Alex"` when inspecting as the Alex identity, `myName: "Beth"` as the Beth identity — matching the values the handlers wrote.

Reproduction in CT-1598. Regular (non-scoped) pieces continue to inspect identically (verified against `packages/patterns/notes/note.tsx`).

## Test plan

- [x] New unit tests in `packages/cli/test/piece-inspect.test.ts` (primitives, null/undefined, arrays, top-level descent, per-identity divergence).
- [x] All existing CLI tests still pass (`cd packages/cli && deno task test` — 36 passed, 1 ignored).
- [x] All runner tests still pass (`cd packages/runner && deno task test` — 451 passed).
- [x] End-to-end repro against a live toolshed (`cozy-poll-scoped` pattern, two identities, two voters).
- [ ] CI green.

Refs: CT-1598

---
## Summary by cubic
Fixes `cf piece inspect --identity` so `PerUser`/`PerSession` fields resolve using the caller’s identity, matching what handlers wrote. Aligns CLI inspect with runtime scope resolution and addresses CT-1598.

- **Bug Fixes**
  - Added `readWithPropertyScope` to read top-level object fields via `cell.key(prop).get()`.
  - Updated `inspectPiece` to use `getCell().pull()` and scope-aware reads for `input` and `result`.
  - Added unit tests in `packages/cli/test/piece-inspect.test.ts` covering primitives and per-identity divergence.

<sup>Written for commit 570448bd400684a16cfb135c190bba1e8ae49a1d. Summary will update on new commits.</sup>

### DIFF
--- a/packages/cli/lib/piece.ts
+++ b/packages/cli/lib/piece.ts
@@ -831,6 +831,43 @@ export async function generateSpaceMap(
   return formatSpaceMap(connections, format);
 }
 
+/**
+ * Read a cell's value with per-property scope awareness.
+ *
+ * `cell.get()` on a top-level cell returns whatever is stored *inline* in the
+ * underlying document at the cell's effective scope. For per-property scoped
+ * fields (e.g. `PerUser<string>` on an otherwise space-scoped argument), the
+ * inline value is the default (e.g. `""`); the real per-user value lives in a
+ * separate scoped instance and is only resolved when the read traverses through
+ * a per-property link.
+ *
+ * `cell.key(prop)` constructs a child cell that inherits the property schema's
+ * declared scope (see `cell.ts` Cell.key()), so calling `.get()` on the child
+ * cell correctly reaches the scoped storage instance.
+ *
+ * For `cf piece inspect`, we want the caller to see the values they would see
+ * as if they were running the pattern (per-user fields resolved against the
+ * caller's identity). So we read top-level object properties via `cell.key()`
+ * rather than using the parent's flat `.get()` value directly. See CT-1598.
+ */
+export function readWithPropertyScope(cell: Cell<unknown>): unknown {
+  const value = cell.get();
+  if (
+    value === null || value === undefined || typeof value !== "object" ||
+    Array.isArray(value)
+  ) {
+    return value;
+  }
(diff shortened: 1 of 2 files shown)


jkomoros:
Closing in favor of Berni's broader CLI scoped-cell work that's already in progress: commit `5bb7dad1d` "feat(cli): Add scoped cf cell reference syntax" (already merged to main) and follow-ups thread scope handling through the CLI/runtime entity-access path more comprehensively than my CLI-boundary patch here. That deeper integration is the right home for the CT-1598 fix.

This PR was a band-aid at the CLI serialization boundary; Berni's track addresses the same class of bug at the layer where it should live.

PR #3606, opened 2026-05-15
Mediate read_file outputs through CFC

mathpirate:
## Summary
- Preserve trusted `cfcResult` metadata on successful `read_file` tool outputs.
- Treat successful `read_file` output as CFC-mediated before it reaches the model, while preserving structured file-tool failures as recoverable operational outputs.
- Record observed read content labels into `cfcModelContext` so later model-authored sandbox inputs carry the influence label.

## Tests
- `deno fmt packages/cf-harness/src/tools/read-file.ts packages/cf-harness/src/prompt-loop.ts packages/cf-harness/test/tools-execution.test.ts packages/cf-harness/test/prompt-loop.test.ts packages/cf-harness/test/artifacts.test.ts`
- `deno test --allow-env --allow-read --allow-run --allow-write packages/cf-harness/test/tools-execution.test.ts packages/cf-harness/test/prompt-loop.test.ts packages/cf-harness/test/artifacts.test.ts packages/cf-harness/test/cfc-invocation-context.test.ts`
- `git diff --check`

---
## Summary by cubic
Mediates `read_file` outputs through CFC so the model only sees observed content with trusted metadata. Also propagates observed content labels into `cfcModelContext` and denies unmediated reads when enforcement is on.

- **New Features**
  - Treat successful `read_file` results as CFC-mediated; include summarized CFC info and truncate long content in model-facing output.
  - Preserve trusted `cfcResult` on `read_file` success and add `isReadFileToolSuccessOutput` for detection.
  - Deny `read_file` outputs without CFC metadata in `enforce-explicit` mode (observation-denied handle); otherwise pass through truncated content.
  - Record observed `read_file` stdout labels into `cfcModelContext` so later tool inputs (e.g., `bash`) carry the influence label.
  - Prompt loop only mediates `bash` and successful `read_file` outputs; Docker runsc leaves output unmediated when the CFC sidecar is missing.

<sup>Written for commit 7ad9bb18a5d24fa0bc61a2a34c01e6cd9a0fd5b6. Summary will update on new commits. <a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3606?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/cf-harness/src/prompt-loop.ts
+++ b/packages/cf-harness/src/prompt-loop.ts
@@ -85,6 +85,7 @@ import {
   cwdMarkerForOutput,
   extractFinalWorkingDirectory,
 } from "./tools/shell-cwd.ts";
+import { isReadFileToolSuccessOutput } from "./tools/read-file.ts";
 import { isViewImageToolSuccessOutput } from "./tools/view-image.ts";
 import type { HarnessFailureRecord } from "./diagnostics.ts";
 import { DEFAULT_PARENT_TOOL_IDS as DEFAULT_PROMPT_LOOP_TOOL_IDS } from "./contracts/tool-descriptor.ts";
@@ -954,8 +955,12 @@ const stripInternalCfcFields = (output: unknown): unknown => {
   return publicOutput;
 };
 
-const toolOutputNeedsSandboxMediation = (toolId: BuiltinToolId): boolean =>
-  toolId === "bash";
+const toolOutputNeedsSandboxMediation = (
+  toolId: BuiltinToolId,
+  output: unknown,
+): boolean =>
+  toolId === "bash" ||
+  (toolId === "read_file" && isReadFileToolSuccessOutput(output));
 
 const createOutputHandle = (
   resultRef: ToolResultRef,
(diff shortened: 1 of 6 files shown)


willkelly (approved):
This looks good to me. The important behavior I wanted to see is here: successful `read_file` output now carries the trusted CFC result through the tool layer, and the prompt loop uses that mediated result before showing file content to the model. In the happy path, the model sees the CFC-released stdout rather than the raw file bytes, and the observed stdout labels get folded into `cfcModelContext` for later turns.

I also checked the failure/absence cases. The prompt-loop tests cover the main model-facing safety properties: mediated `read_file` content is exposed and tracked, while a successful `read_file` result with no trusted CFC metadata is denied in `enforce-explicit` and does not leak the raw file content. The broader existing bash/CFC tests cover the same expose/deny/observe shape for sandbox output, including opaque/denied observations not contributing to model context.

For the runtime side: if the runsc CFC result sidecar is unavailable, the harness does not hang or retry forever. Docker/runsc reads the sidecar once after `docker start` / `docker wait`; a missing sidecar becomes `cfcResult === undefined`, while non-NotFound read errors or parse errors become denied CFC results. That means prompt-loop mediation is the enforcement boundary for missing metadata: disabled/observe can expose raw output by design, but enforce mode turns the missing metadata into `observation-denied` before the model sees it.

The only non-blocking caveat I’d call out is that I don’t see a Docker/runsc unit test that directly exercises the “`cfcResultDir` is configured but the sidecar file is missing” `NotFound` branch. The code path is straightforward, and the prompt-loop missing-metadata test covers the model-facing safety property, so I don’t think that should block this PR. It may still be worth adding a small follow-up test or comment documenting the invariant that model-visible `read_file` output must flow through prompt-loop mediation, since the lower runtime layer still has the raw command result when the sidecar is absent.

PR #3610, opened 2026-05-15
feat(patterns): Rewrite parking coordinator with scopes

seefeldb:
## Summary
- rewrite the parking coordinator pattern inputs to use scoped cells for shared data and per-user/per-session UI state
- preserve seeded parking spots through the generated pattern schema
- harden add/edit handlers so browser form submissions without event payloads fall back to scoped form cells

## Validation
- deno task cf check packages/patterns/factory-outputs/parking-coordinator/main.tsx --no-run
- deno task cf test packages/patterns/factory-outputs/parking-coordinator/main.test.tsx
- manual browser play test on local dev offset 154 with in-app browser plus agent-browser: add person, add spot, manual assign/cancel, edit, remove, and cross-user shared/session state checks

---
## Summary by cubic
Rewrote the parking coordinator to use scoped cells for shared data and per-user/session UI. Added scoped cell constructors with schema scope injection and ensured cell links inherit explicit or contextual scopes; preserves seeded spots and hardens actions for missing payloads.

- **New Features**
  - Scoped constructors on Cell-like types: `.perSpace`, `.perUser`, `.perSession`, each with `.of` and `.for`; runtime merges defaults with `scope`, applies scope to cell links (including contextual schema scope), and rejects conflicting scopes; plain `.of(...)` inherits the outer context scope.
  - TS transformer injects `scope` for constructor `.of`/`.for` (including contextual typing with no value); docs and fixtures added.

- **Refactors**
  - Inputs now accept optional `PerSpace` cells for `spots/people/requests`; local UI state moved to `perUser`/`perSession` with safe defaults.
  - Actions fall back to scoped form cells when event payloads are missing; validation blocks past dates and handles null/undefined values; tests cover payloadless edits and scoped link behavior.

<sup>Written for commit 5781f99c5458c089b48b5d121b1c6d28f13b59d9. Summary will update on new commits. <a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3610?utm_source=github">Review in cubic</a></sup>

NEW_PERF_BASELINE: subtest: pattern-unit-3/pattern-unit-tests > packages/patterns/factory-outputs/parking-coordinator/main.test.tsx = 90s


seefeldb:
Oh, codex moved all the local variables to the input. I don't think that should be necessary. Feels like maybe there is a DX gap. `Writable.of<PerUser<string | Default<"">>>("")` is a lot to write. Let's see what else could work.

packages/ts-transformers/src/transformers/schema-injection.ts
```diff
@@ -3325,7 +3372,9 @@ export class SchemaInjectionTransformer extends HelpersOnlyTransformer {
           resolved,
           checker,
           typeRegistry,
-          resolved.inferred ? { widenLiterals: true } : undefined,
+          resolved.inferred || scope
```
mathpirate, line 3375: When contextual scoped cells are lowered, this only passes the contextual scope into `toSchema`; the emitted call remains `Writable.of(...)`. That runtime path does not pass a link scope to `createCell`, and `CellImpl` normalizes missing link scope to `space`, so a `PerUser<Writable<string>> = Writable.of(...)` cell still stores as space-scoped with only a user-scoped schema. That breaks the documented contextual authoring style and can leak or collide user/session-local state; lower contextual scopes to the matching `.perUser` or `.perSession` constructor, or make root cell creation derive link scope from schema scope.

PR #3612, opened 2026-05-15
feat(data-model): `[DEEP_FREEZE]`/`[IS_DEEP_FROZEN]` symbol protocol for `FabricInstance` deep-freezing

danfuzz:
## Summary

Extends the `deepFreeze()` utility in `packages/data-model/` to dispatch on
the `FabricInstance` abstract base via the existing `[DEEP_FREEZE]` symbol
protocol, makes the recursion cycle-safe, and retires the
`FabricInstance`-arm throw from the `isDeepFrozenFabricValue` type guard
in favor of the existing `[IS_DEEP_FROZEN]` protocol member.

## What this delivers

- **`deepFreeze()` dispatch is base-class-gated on `FabricInstance`.**
  The recursive arm for per-instance freezing is now an `instanceof
  FabricInstance` gate that invokes the abstract `[DEEP_FREEZE]` protocol
  member, dispatching generically on the base class rather than enumerating
  concrete subclasses. A `FabricPrimitive` short-circuit arm sits above it
  (these self-freeze at construction and have no outbound references).
- **`deepFreeze()` is now circular-value-safe.** A shared cycle-detection
  set is threaded through all recursive calls — both the plain-object /
  array arm and the `FabricInstance` protocol arm — so values that
  participate in cycles complete deep-freezing cleanly rather than
  infinite-recursing. The `[DEEP_FREEZE]` protocol callback carries the
  shared cycle state into participating `FabricInstance`s transparently,
  via the callback's closure — no signature change is required on subclass
  impls. This mirrors the existing `subIsDeepFrozen` / `checkValue`
  closure-`seen` pattern on the `[IS_DEEP_FROZEN]` side.
- **`deepFreeze()` on a `FabricInstance` now caches its result.** The
  post-protocol-freeze instance is added to the deep-frozen cache before
  return, matching the existing cache-write pattern of the plain-object /
  array arm. Subsequent `isDeepFrozen()` checks on the same instance
  short-circuit in O(1).
- **`isDeepFrozenFabricValue()` no longer throws on `FabricInstance`.**
  The previous `Cannot yet handle instance of class …` throw on the
  `FabricInstance` arm is retired; the function now answers via the
  `[IS_DEEP_FROZEN]` protocol member, returning `false` (rather than
  throwing) when an instance is not in canonical deep-frozen form.
- **Test coverage added** in `deep-freeze.test.ts` and
  `fabric-native-instances.test.ts` for: direct `[DEEP_FREEZE]` /
  `[IS_DEEP_FROZEN]` protocol-member invocation on the wired
  `FabricInstance` subclasses; cycle-safety across the dispatch arms
  (self-referential plain object and array, two-node and mixed
  object/array cycles, cycles via the `[DEEP_FREEZE]` protocol through
  `FabricError.error.cause` and through `ProblematicValue` / `UnknownValue`
  `state`, and a cross-instance cycle); and `isDeepFrozen()` coverage on
  `FabricInstance`, `FabricPrimitive`, and a `FabricInstance` participating
  in a circular reference.

## Out of scope

Other deep-freeze surfaces are intentionally not part of this PR and will
be addressed in follow-on work:

- The JSON-encoding context's class-registry-fallback path (in
  `json-encoding-context.ts`) is unchanged here.
- `FabricMap` and `FabricSet` retain throwing-stub `[DEEP_FREEZE]` /
  `[IS_DEEP_FROZEN]` implementations pending future work on those classes'
  deep-frozen-shape.

## Test plan

- `deno fmt` / `deno lint` / `deno check` clean (data-model package).
- `packages/data-model` `deno task test` — 44 passed / 1338 steps / 0
  failed.
- Cross-package `deno check` clean on `packages/memory`, `packages/runner`,
  `packages/toolshed`.

---
## Summary by cubic
Adds a symbol-based deep-freeze protocol and makes `deepFreeze()` cycle-safe with a shared seen set. Caches `FabricInstance` results so later `isDeepFrozen()` checks are O(1).

- **New Features**
  - Introduce `[DEEP_FREEZE]` and `[IS_DEEP_FROZEN]` on `FabricInstance`.
  - Update `deepFreeze()` dispatch: primitives/cached short-circuit; `FabricPrimitive` short-circuits; `instanceof FabricInstance` delegates via `[DEEP_FREEZE]`; array/object fallback.
  - Handle circular graphs by threading a shared in-progress set through recursion (including protocol callbacks); add internal `deepFreezeInProgress()`.
  - Route `isDeepFrozenFabricValue` via `[IS_DEEP_FROZEN]` behind an `instanceof FabricInstance` gate.
  - Implement the protocol on `FabricError` and `FabricRegExp` (freezes wrapped `RegExp` in place); keep throwing stubs for `FabricMap`/`FabricSet`.
  - Cache `[DEEP_FREEZE]` results for `FabricInstance`s in the deep‑frozen cache to make follow-up `isDeepFrozen()` checks O(1).

- **Bug Fixes**
  - Fix infinite recursion on circular graphs (objects, arrays, and cross-instance cycles); add tests that now pass.
  - Provide real `[DEEP_FREEZE]`/`[IS_DEEP_FROZEN]` for `ProblematicValue` and `UnknownValue` so lenient decode stays graceful; retire the `FabricInstance`-arm throw in `isDeepFrozenFabricValue`.
  - Add `isDeepFrozen()` coverage for `FabricInstance`, `FabricPrimitive`, and circular references.

<sup>Written for commit 8c9d78a6373456eae6c999414573f5dd6a6cd93a. Summary will update on new commits. <a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3612?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/data-model/deep-freeze.ts
+++ b/packages/data-model/deep-freeze.ts
@@ -205,9 +205,13 @@ function deepFreezeInProgress<T>(value: T, inProgress?: Set<object>): T {
   // `FabricInstance` doesn't need to be `inProgress`-aware in its own
   // signature; the callback carries the shared state transparently.
   if (value instanceof FabricInstance) {
-    return value[DEEP_FREEZE](
+    const result = value[DEEP_FREEZE](
       (v) => deepFreezeInProgress(v, inProgress),
     ) as T;
+    // Cache the now-deep-frozen result so subsequent `isDeepFrozen()` checks
+    // short-circuit in O(1), mirroring arm 4's cache-write below.
+    addToDeepFrozenCache(result as object);
+    return result;
   }
 
   // Arm 4: plain object or array -- recurse into children, then freeze.
--- a/packages/data-model/deep-freeze.ts
+++ b/packages/data-model/deep-freeze.ts
@@ -154,13 +154,13 @@ export function deepFreeze<T>(value: T): T {
  * thread the same set through so a cycle back to a value already being
  * deep-frozen by an outer call short-circuits.
  *
- * The arm 3 callback `(v) => deepFreezeInProgress(v, inProgress)` is the
- * "threading into the `FabricInstance`s which are participating in the act
- * of deep-freezing": the participating impl invokes `subFreeze` on nested
- * `FabricValue`s, and that closure carries the shared `inProgress` so the
- * impl's recursion is cycle-safe by construction. This mirrors the
- * `subIsDeepFrozen`/`checkValue`-closure pattern already used by
- * `isDeepFrozenFabricValue` for the `[IS_DEEP_FROZEN]` side.
+ * The arm 3 callback `(v) => deepFreezeInProgress(v, inProgress)` threads
+ * the shared cycle state into participating `FabricInstance`s: each impl
+ * invokes `subFreeze` on its nested `FabricValue`s, and that closure
+ * carries the shared `inProgress` so the impl's recursion is cycle-safe by
+ * construction. This mirrors the `subIsDeepFrozen`/`checkValue`-closure
+ * pattern already used by `isDeepFrozenFabricValue` for the
+ * `[IS_DEEP_FROZEN]` side.
  *
  * Unlike `isDeepFrozenInProgress` (which removes values from its
  * `inProgress` set after sub-checks complete -- because the question
@@ -201,9 +201,9 @@ function deepFreezeInProgress<T>(value: T, inProgress?: Set<object>): T {
   // Arm 3: a `FabricInstance` freezes itself in place via its `[DEEP_FREEZE]`
   // protocol member. The `subFreeze` callback closes over `inProgress` so the
   // impl's recursion into nested `FabricValue`s shares cycle state with this
-  // call -- this is the threading Dan's design calls for: the participating
-  // `FabricInstance` doesn't need to be `inProgress`-aware in its own
-  // signature; the callback carries the shared state transparently.
+  // call: the participating `FabricInstance` doesn't need to be
+  // `inProgress`-aware in its own signature; the callback carries the shared
+  // state transparently.
   if (value instanceof FabricInstance) {
     const result = value[DEEP_FREEZE](
       (v) => deepFreezeInProgress(v, inProgress),
--- a/packages/data-model/deep-freeze.ts
+++ b/packages/data-model/deep-freeze.ts
@@ -124,10 +124,11 @@ function isDeepFrozenInProgress(
  *    objects): short-circuit unchanged.
  * 2. `FabricPrimitive` instance: short-circuit unchanged -- these self-freeze
  *    at construction and have no outbound references.
- * 3. Value carrying the `[DEEP_FREEZE]` symbol method (e.g. a
- *    `FabricInstance` subclass): delegate generically to that method,
- *    handing recursion through as the `subFreeze` callback. `deepFreeze()`
- *    itself stays class-agnostic (no per-class knowledge).
+ * 3. `FabricInstance` (the abstract base): delegate generically to its
+ *    `[DEEP_FREEZE]` protocol member, handing recursion through as the
+ *    `subFreeze` callback. The dispatch gates on `instanceof` against the
+ *    abstract base -- it operates generically and does not enumerate
+ *    concrete subclasses.
  * 4. Plain object or array: recursively freeze children, then freeze the
  *    container.
  *
@@ -148,14 +149,14 @@ export function deepFreeze<T>(value: T): T {
     return value;
   }
 
-  // Arm 3: anything carrying the `[DEEP_FREEZE]` symbol method freezes itself
-  // in place via that method, recursing through the `deepFreeze` callback.
-  // The duck-typed check keeps `deepFreeze()` free of per-class knowledge.
-  const deepFreezable = value as {
-    [DEEP_FREEZE]?: (subFreeze: (value: FabricValue) => FabricValue) => unknown;
-  };
-  if (typeof deepFreezable[DEEP_FREEZE] === "function") {
-    return deepFreezable[DEEP_FREEZE]((v) => deepFreeze(v)) as T;
+  // Arm 3: a `FabricInstance` freezes itself in place via its `[DEEP_FREEZE]`
+  // protocol member, recursing through the `deepFreeze` callback. Gating on
+  // `instanceof` against the abstract base keeps `deepFreeze()` generic --
+  // it dispatches on the base class, not an enumeration of concrete
+  // subclasses (the `[DEEP_FREEZE]` member is abstract on `FabricInstance`,
+  // so every instance is guaranteed to implement it).
+  if (value instanceof FabricInstance) {
+    return value[DEEP_FREEZE]((v) => deepFreeze(v)) as T;
   }
 
   // Arm 4: plain object or array -- recurse into children, then freeze.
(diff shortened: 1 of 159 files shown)


packages/data-model/deep-freeze.ts
```diff
@@ -111,16 +117,48 @@ function isDeepFrozenInProgress(
...
- * unchanged. Records the result in the deep-frozen cache so subsequent
- * `isDeepFrozen()` checks return in O(1). Returns the (now-frozen) value.
+ * Recursively freezes the given value in place. Dispatches on four arms, in
+ * order:
+ *
+ * 1. Necessarily- or already-known-deep-frozen value (primitives and cached
+ *    objects): short-circuit unchanged.
+ * 2. `FabricPrimitive` instance: short-circuit unchanged -- these self-freeze
+ *    at construction and have no outbound references.
+ * 3. Value carrying the `[DEEP_FREEZE]` symbol method (e.g. a
```
danfuzz: The `data-model` is specified in terms of the interfaces and classes, not based on duck-type method sniffing. Please use (and document)`instanceof` checks instead of treating each individual method in the protocol as its own independent item.

packages/data-model/fabric-native-instances.ts
```diff
@@ -328,6 +386,50 @@ export class FabricMap
...
 
+  /**
+   * Deep-freezes in place. A `Map`'s `.set`/`.delete` bypass property
+   * descriptors, so `Object.freeze` alone cannot make the collection
+   * immutable -- the only fully-immutable form is `FrozenMap`. By the time
+   * `[DEEP_FREEZE]` runs at the memory-model egress boundary, the internal
+   * slot must already be that frozen-variant form; this method verifies
+   * (death before confusion) and then recurses into each key and value.
+   */
+  [DEEP_FREEZE](
```
danfuzz, line 404: Please just leave `throw`ing stubs for the new methods in both`FabricMap` and `FabricSet`. These classes are not yet used and problematic in other ways to boot. We'll do a proper rework of them as the other dust settles.

packages/data-model/interface.ts
```diff
@@ -42,13 +42,42 @@ export const DECONSTRUCT: unique symbol = Symbol.for("common.deconstruct");
  */
 export const RECONSTRUCT: unique symbol = Symbol.for("common.reconstruct");
 
+/**
+ * Well-known symbol for deeply freezing a fabric instance in place. The method
+ * freezes the instance's own internal slot(s) and recurses into any nested
+ * `FabricValue`s via the provided `subFreeze` callback. This lets the
+ * generic `deepFreeze()` operate on `FabricInstance`s without building
+ * per-class knowledge into the freeze utility (it duck-types this symbol).
```
danfuzz: Same comment as above: The `data-model` uses `instanceof` with interfaces and classes, not duck typing. Please fix this comment (and others which mention duck typing in this context).

packages/data-model/json-encoding-context.ts
```diff
@@ -422,11 +422,11 @@ export class JsonEncodingContext implements SerializationContext<string> {
       if (handler) {
         if (this.lenient) {
           try {
-            return handler.deserialize(
+            return deepFreeze(handler.deserialize(
```
danfuzz, line 456: I think `deserialize()` should be making a contractual guarantee to return deep-frozen values, rather than sprinkling `deepFreeze()` calls around every use of it.

packages/data-model/json-encoding-context.ts
```diff
@@ -449,10 +449,10 @@ export class JsonEncodingContext implements SerializationContext<string> {
       if (cls) {
         if (this.lenient) {
           try {
-            return cls[RECONSTRUCT](
+            return deepFreeze(cls[RECONSTRUCT](
```
danfuzz: Suggest: Add a task to put a `shouldDeepFreeze` getter on `ReconstructionContext`, which `[RECONSTRUCT]` methods are expected to query and abide by.

packages/data-model/fabric-native-instances.ts
danfuzz: Done in `a095e0c60` — `FabricMap` and `FabricSet` `[DEEP_FREEZE]`/`[IS_DEEP_FROZEN]` are reverted to throwing stubs (`"Fabric{Map,Set}: not yet implemented"`), matching the existing not-yet-implemented idiom already used for `[DECONSTRUCT]`/`[RECONSTRUCT]` in those same classes. Deferred to the separate FabricMap/FabricSet rework per your call (not-yet-used, problematic in other ways). `FabricError`/`FabricRegExp` real impls and the `ProblematicValue`/`UnknownValue` impls are unaffected; tests updated to assert the stub contract.

packages/data-model/deep-freeze.ts
danfuzz: Done in `8c480aa6d`. Arm 3 of `deepFreeze()` (and the parallel `FabricInstance` arm of `isDeepFrozenFabricValue`) now gates on `instanceof FabricInstance` — the abstract base — rather than duck-typing the symbol method. The `[DEEP_FREEZE]`/`[IS_DEEP_FROZEN]` protocol members are unchanged and invoked *after* the `instanceof` gate (they are abstract on `FabricInstance`, so every instance is guaranteed to implement them; the redundant duck-type guard is removed as dead code). This operates generically on the abstract base — it does not enumerate concrete subclasses — consistent with how the rest of the package dispatches. All duck-typing doc-comments were swept to describe this base-class-`instanceof` shape (see Comment on `interface.ts` as well). For provenance: this is the shape Dan confirmed directly (the "operate generically via the abstract base with `instanceof`, do not enumerate concrete classes" principle — the earlier "no class names" framing was a relay-compression misframing, corrected of record).

packages/data-model/interface.ts
danfuzz: Done in `8c480aa6d`. Full doc-comment sweep: every "duck-typed" / "class-agnostic" / "no per-class knowledge" mention across the F1 source + test surface (including both `interface.ts` symbol docstrings here, and the `deepFreeze()` / `isDeepFrozenFabricValue` doc-comments) is rewritten to describe the base-class-`instanceof` shape — the protocol members are abstract on `FabricInstance`, and dispatch gates on `instanceof` against the abstract base, operating generically without enumerating concrete subclasses. Exhaustive grep confirms zero duck-typing framing remains on the F1 surface. The dispatch code itself changed in the same commit (see the `deep-freeze.ts` thread). For provenance: this base-class-`instanceof` principle is the shape Dan confirmed directly.

packages/data-model/json-encoding-context.ts
danfuzz: Filed as a tracked follow-up task per your "add a task" framing — it is acknowledged and tracked, not folded into this PR's code. The task ("`shouldDeepFreeze` getter on `ReconstructionContext`") records the scope as: add a `shouldDeepFreeze` getter to `ReconstructionContext`, which `[RECONSTRUCT]` implementations are expected to query and abide by — making the deep-frozen-egress decision a queryable property of the reconstruction context rather than an out-of-band assumption. It's scoped narrowly to the getter + the `[RECONSTRUCT]`-abides-by-it contract (distinct from, and separately tracked alongside, the broader `deserialize()`-contract question). Provenance and your verbatim suggestion are recorded on the task.

packages/data-model/deep-freeze.ts
danfuzz: Provenance clarification (superseding the "For provenance:" sentence in my earlier reply): the basis for this change is **your own written review instruction on this PR** — this comment ("Please use (and document) `instanceof` checks instead of treating each individual method in the protocol as its own independent item") — together with the Dan-authored `tasks.md` correction-of-framing in commit `4e78b87`. It is directly responsive to that written instruction. My earlier phrasing "the shape Dan confirmed directly" overstated the provenance *mechanism* (it could read as a separately-relayed confirmation, which there was not); the accurate statement is that the change implements your own written review comment here. The code (`8c480aa6d`) is unchanged by this clarification.

packages/data-model/interface.ts
danfuzz: Provenance clarification (superseding the "For provenance:" sentence in my earlier reply): the basis is **your own written review instruction on this PR** — this comment ("uses `instanceof` with interfaces and classes, not duck typing. Please fix this comment (and others which mention duck typing)") — plus the Dan-authored `tasks.md` correction-of-framing in commit `4e78b87`. The doc-comment sweep in `8c480aa6d` is directly responsive to that written instruction. My earlier "the shape Dan confirmed directly" phrasing overstated the provenance *mechanism* (it could read as a separately-relayed confirmation; there was none); accurately, this implements your own written review comment here. Code unchanged by this clarification.

danfuzz: Carrier note (per the split structure): this comment's fix is split across two PRs by file. The **`interface.ts`** half of the duck-typing doc-comment sweep is carried by the **U1 pre-PR #3614** (`coder-colt/deep-freeze-protocol-defs`); the `deep-freeze.ts` half is carried by the residual #3612 (this PR). The thread stays here on #3612 where you commented; this note just records which PR's diff carries which half so the resolution is traceable.

PR #3615, opened 2026-05-18
Add runsc CFC read_file integration smoke

mathpirate:
## Summary
- Add an env-gated Docker/runsc CFC prompt-loop integration smoke for the `read_file` mediation path.
- Keep the target to Wilk's first gate: `/workspace` only, no `/fabric` or FUSE.
- Split CFC invocation input-label derivation so trusted `read_file` path selection does not get direct prompt-slot taint, while prior observed model-context labels still flow onto model-selected read paths.
- Assert real runsc sidecar metadata reaches model-facing `read_file` output and retained tool artifacts.

## Notes
- #3606 is now merged; this PR is rebased onto `main` and contains only the integration smoke plus the selector-label adjustment required for the real runsc path.
- The smoke is skipped unless `CF_HARNESS_INTEGRATION=1`, `CF_HARNESS_RUNSC_CFC_RESULT_DIR`, and `CF_HARNESS_RUNSC_CFC_INVOCATION_CONTEXT_DIR` are configured.
- This keeps `read_file` aligned with the trusted-selector interpretation: successful file bytes are mediated by the file read's CFC result, not by direct prompt influence on the path string. If the model selected the path based on previously observed CFC-mediated output, that accumulated model-context label is still carried into the invocation sidecar.

## Tests
- `deno test -A packages/cf-harness/test/cfc-invocation-context.test.ts packages/cf-harness/test/prompt-loop.test.ts packages/cf-harness/test/tools-execution.test.ts`
- `CF_HARNESS_INTEGRATION=1 CF_HARNESS_RUNSC_CFC_RESULT_DIR="$HOME/.local/share/runsc-cfc/cfc-results" CF_HARNESS_RUNSC_CFC_INVOCATION_CONTEXT_DIR="$HOME/.local/share/runsc-cfc/cfc-invocations" deno test -A --filter "prompt loop mediates read_file output through runsc-cfc" packages/cf-harness/integration/engine.integration.test.ts`
- `git diff --check`

### DIFF
--- a/packages/cf-harness/integration/engine.integration.test.ts
+++ b/packages/cf-harness/integration/engine.integration.test.ts
@@ -1,5 +1,5 @@
 import { assert, assertEquals, assertStringIncludes } from "@std/assert";
-import { fromFileUrl } from "@std/path";
+import { fromFileUrl, join } from "@std/path";
 import type {
   CfcEnforcementMode,
   CfcLabelView,
@@ -7,6 +7,10 @@ import type {
 } from "@commonfabric/runner/cfc";
 import { CfHarnessEngine } from "../src/engine.ts";
 import { CfHarnessPromptLoop } from "../src/prompt-loop.ts";
+import {
+  readHarnessRunState,
+  readHarnessTranscript,
+} from "../src/artifacts.ts";
 import {
   CFC_INVOCATION_CONTEXT_DIR_ENV,
   CFC_RESULT_DIR_ENV,
@@ -86,6 +90,8 @@ const labsRootPath = (): Promise<string> =>
 
 interface WithHarnessOptions {
   cfcEnforcementMode?: CfcEnforcementMode;
+  artifactRoot?: string;
+  model?: string;
   configureSandbox?: (
     workspaceHostPath: string,
   ) => Partial<Parameters<typeof resolveDockerRunscSandboxConfig>[0]>;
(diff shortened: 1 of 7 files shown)


willkelly (approved):
LGTM. I did a deeper pass on the read_file coverage and this looks reasonable for the PR's stated scope.

What I verified:

- the new env-gated integration smoke exercises a real prompt-loop `read_file` call through runsc-CFC, checks model-facing mediated content, CFC stdout/stderr/exitCode metadata, runsc diagnostics, persisted transcript/run-state, and retained tool-output artifacts;
- the selector-label change is covered by unit tests that split prompt-slot influence from model-context labels;
- `read_file` no longer puts direct prompt-slot taint on the trusted path selector while still carrying prior observed model-context labels onto model-selected args;
- existing unit coverage already covers the ordinary `read_file` mechanics: relative path resolution, `maxBytes`, missing file, not-a-file, path escape, artifact-root denial, and CFC result preservation.

I also ran locally on `60936deb06c166dd10c51c8207a2a14d6afd9e1a`:

- `deno test -A packages/cf-harness/test/cfc-invocation-context.test.ts packages/cf-harness/test/prompt-loop.test.ts packages/cf-harness/test/tools-execution.test.ts`
- `GIT_MASTER=1 git diff --check origin/main...HEAD`

Both passed. The runsc integration filter is correctly env-gated and skipped without the required CFC env vars.

Non-blocking follow-up coverage I would consider later: chmod/permission-denied, large read/model-facing truncation, binary or invalid UTF-8 behavior, denied/opaque CFC read output, and stricter bad-input/schema cases. None of those block this PR.

PR #3614, opened 2026-05-16
feat(data-model): `[DEEP_FREEZE]`/`[IS_DEEP_FROZEN]` protocol — defs, impls, stubs (U1 pre-PR)

danfuzz:
## Summary

Adds the `[DEEP_FREEZE]` / `[IS_DEEP_FROZEN]` symbol-protocol **definition
layer** to the data model: the two new well-known symbols, their abstract
members on `FabricInstance`, the concrete per-class implementations
(`FabricError`, `FabricRegExp`, `ProblematicValue`, `UnknownValue`), and
deliberate throwing stubs (`FabricMap`, `FabricSet`). This is an **inert
layer** — no caller dispatches to the new members yet (the dispatch/use
layer is a separate follow-on unit).

This is **U1 (pre-PR#1)** of the split F1 deep-freeze protocol work
(`coordination/docs/2026-05-14-deep-freeze-discipline-audit.md` §3.7 /
§10; `tasks.md` #56 "Consolidated 5-unit structure"). **The split into
units was directed by Dan directly** (relayed for the record — meta-repo
git metadata does not carry who-decided): the F1 work is delivered as a
landed-in-order chain rather than one PR. See **Sequencing** below.

## Independently safe + correct to land (named acceptance criterion)

This PR is safe and correct to land on its own — it is an **inert
definition layer**. (This is the foreman's named blocking review
criterion for each split unit; the at-source basis:)

* **Safe (gates green standalone, no dangling sibling reference).** Built
  off `origin/main`; `deno fmt`/`lint`/`check` clean and the data-model +
  runner suites pass on this branch alone (see Validation). It references
  no symbol or code that exists only in a not-yet-landed sibling unit.
* **Correct (semantically complete as landed, not a partial contract).**
  Verified at-source: `origin/main` `deep-freeze.ts` references neither
  `DEEP_FREEZE` nor `IS_DEEP_FROZEN`, and the *only* dispatcher to these
  members is `deep-freeze.ts` (arm-3 / type-guard) — which stays at
  `origin/main` in this PR (it is the follow-on unit). So **nothing
  dispatches to the new members** once this lands: the abstract members
  force every `FabricInstance` subclass to implement them (all such
  subclasses are in this PR's file-set, so the contract is satisfied
  within the PR and it compiles), and no caller invokes them, so there is
  no partial/sometimes-wrong behavior. Adding an abstract member only
  constrains subclasses, not consumers, so `deep-freeze.ts` (and every
[...]

PR #3616, opened 2026-05-18
feat(data-model): `TypeHandler.deserialize()` arm-1 contractually returns deep-frozen (U2 split-PR)

danfuzz:
## Summary

Makes `TypeHandler.deserialize()`'s **type-handler dispatch arm (arm-1)**
contractually return deep-frozen values — every arm-1 return passes
through `deepFreeze(...)` once at the `deserialize()` boundary, instead
of relying on every caller to freeze. This is the **NARROW** contract:
the class-registry fallback arm (`cls[RECONSTRUCT](...)`, `:445-466`) is
a separate sibling branch and is **intentionally NOT covered**.

This is **U2 / split-PR#2** of the F1 deep-freeze arc, resolving Dan's
PR #3612 review comment #4 (`json-encoding-context.ts:425`, verbatim:
*"`deserialize()` should be making a contractual guarantee to return
deep-frozen values, rather than sprinkling `deepFreeze()` calls around
every use of it."*). The **NARROW** scope (arm-1 only; arm-2 out) was
**Dan-confirmed directly** (scope-reconciliation doc
`coordination/docs/2026-05-15-typehandler-deserialize-scope-reconciliation.md`;
relayed for the record — meta-repo git metadata does not carry
who-decided). See **Sequencing**.

## Change

* **`json-encoding-context.ts`** — the `if (handler) { ... }` arm now
  returns every value through `deepFreeze(...)`: the handler-produced
  value (lenient *and* strict paths) and the lenient-mode
  `ProblematicValue` fallback. The contract is enforced once, at the
  `deserialize()` boundary for this arm, instead of per-caller.
  `deepFreeze` was already imported (pre-existing `TAGS.quote` literal
  freeze) — zero import cost. An in-code comment states explicitly that
  the class-registry fallback below is a separate arm and is
  intentionally NOT covered by this contract. The class-registry arm is
  **untouched**.
* **`test/json-encoding-context.test.ts`** — new `TypeHandler.
  deserialize() deep-frozen contract` describe: handler-produced value
  (`/EpochNsec@1`), lenient-catch `ProblematicValue` (`/BigInt@1` with
  non-string state), and a handler round-trip all asserted deep-frozen
  at the `deserialize()` boundary.

3-dot diff vs `main`: 2 files, `json-encoding-context.ts` (+29) +
`json-encoding-context.test.ts` (+49) — all in `packages/data-model/`.
No `deep-freeze.ts`, no other surface. Scope-clean, NARROW.

## Independently safe + correct to land (named acceptance criterion)

* **Safe.** `deno fmt`/`lint`/`check` clean and the data-model + runner
  suites pass on this branch alone (see Validation). No dangling
  reference to a not-yet-landed sibling — `deepFreeze`'s
  `FabricInstance`/`FabricPrimitive` handling is the U1 protocol, which
  is **already on `main`** (U1 #3614 merged at `a276ddae7`).
* **Correct (semantically complete as landed, not a partial contract).**
  The arm-1 contract is unconditional `deepFreeze` on every arm-1
  return. For an already-frozen `FabricPrimitive` (the common handler
  output) this is an idempotent O(1) cache hit — correct, just
  non-optimal. That over-conservative-but-correct interim is the
  intended shape; the skip-when-already-frozen optimization is U3's
  `ReconstructionContext.shouldDeepFreeze` getter (separate PR). This is
  **NOT** a sometimes-wrong-until-later contract: arm-1 *always* returns
  deep-frozen as landed. Arm-2's pre-existing behavior is unchanged
  (out of scope by Dan's narrow lock), so nothing regresses.

(This named safe+correct attestation is the binding per-unit review
criterion; the at-source basis is above. Same per-unit-attestation
shape as U1 #3614.)

## Sequencing

This is **U2** of the Dan-directed 5-unit F1 chain (`tasks.md` #56):

* **U1** — `[DEEP_FREEZE]`/`[IS_DEEP_FROZEN]` protocol defs+impls+stubs.
  **MERGED** (#3614, `a276ddae7`).
* **U2 (this PR)** — `TypeHandler.deserialize()` arm-1 contractual
  deep-frozen guarantee (Dan's #4, NARROW). Depends on U1's protocol
  (already on `main`); self-contained otherwise.
* **U3** — `ReconstructionContext.shouldDeepFreeze` getter (Dan's #5;
  the skip-when-already-frozen optimization referenced above).
* **U4** — residual integration on `#3612` (the `deepFreeze()` arm-3
  dispatch + `isDeepFrozenFabricValue` type-guard + R6 throw-retirement).
* **U5** — Dan re-evaluation checkpoint.

Then Stage 2 (freeze-clone collapse + the #3588 perf-baseline
re-measurement). U2 does not depend on U3/U4/U5.

**Base note:** carved off `a276ddae7` (the U1-merge commit, correct
base at carve-time). `main` has since advanced (ts-transformers PRs,
zero data-model overlap — cleanly behind, no conflict). A merge-sync to
current `main` is land-prep hygiene done before ready, not a blocker for
review.

## Validation (standalone, off `main` @ `a276ddae7`)

- [x] `deno fmt` — clean
- [x] `deno lint` — clean
- [x] `deno check` — clean (data-model)
- [x] data-model `deno task test` — 57 passed (1310 steps), 0 failed
- [x] runner `deno task test` — 452 passed (2443 steps), 0 failed

Co-Authored-By: coder-colt (Claude Opus 4.7 (1M context)) <noreply@anthropic.com>

---
## Summary by cubic
Enforces a deep-frozen return contract for the type-handler arm of `TypeHandler.deserialize()`, so all handler-based deserializations are deep-frozen at the `deserialize()` boundary. This removes the need for callers to manually call `deepFreeze`.

- **New Features**
  - Deep-freezes every return from the `if (handler)` arm, including lenient `ProblematicValue`.
  - Leaves the class-registry fallback arm unchanged and out of scope.
  - Adds tests for handler success, lenient fallback, and round-trip to confirm deep-frozen results.

- **Bug Fixes**
  - Tests now assert deep-freezing with `isDeepFrozen` instead of `Object.isFrozen`.
  - Documented the arm-1 frozen-ness contract in code comments and added a concise note to the formal spec.

<sup>Written for commit 304c0c6ee7640cfc5b8973461683622527ab47f8. Summary will update on new commits. <a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3616?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/data-model/json-encoding-context.ts
+++ b/packages/data-model/json-encoding-context.ts
@@ -418,28 +418,39 @@ export class JsonEncodingContext implements SerializationContext<string> {
       }
 
       // --- Type handler dispatch ---
+      //
+      // `TypeHandler.deserialize()` makes a contractual guarantee that its
+      // results are deep-frozen, rather than relying on every caller to
+      // freeze: every return out of this arm passes through `deepFreeze()`.
+      // This covers the handler's produced value (e.g. `FabricPrimitive`
+      // subclasses -- already frozen, so this is an O(1) cache hit) and the
+      // lenient-mode `ProblematicValue` fallback. The class-registry
+      // fallback below is a separate arm and is intentionally NOT covered by
+      // this contract.
       const handler = registry.getDeserializer(tag);
       if (handler) {
         if (this.lenient) {
           try {
-            return handler.deserialize(
+            return deepFreeze(handler.deserialize(
               state,
               runtime,
               (v: JsonWireValue) => this.deserialize(v, runtime, registry),
-            );
+            ));
           } catch (e: unknown) {
-            return new ProblematicValue(
-              tag,
-              state as unknown as FabricValue,
-              e instanceof Error ? e.message : String(e),
-            ) as unknown as FabricValue;
+            return deepFreeze(
+              new ProblematicValue(
+                tag,
+                state as unknown as FabricValue,
+                e instanceof Error ? e.message : String(e),
+              ) as unknown as FabricValue,
+            );
           }
         }
-        return handler.deserialize(
+        return deepFreeze(handler.deserialize(
           state,
           runtime,
           (v: JsonWireValue) => this.deserialize(v, runtime, registry),
-        );
+        ));
       }
 
       // --- Class registry fallback ---
--- a/packages/data-model/json-encoding-context.ts
+++ b/packages/data-model/json-encoding-context.ts
@@ -380,6 +380,11 @@ export class JsonEncodingContext implements SerializationContext<string> {
   /**
    * Deserializes a wire-format value back into modern runtime types.
    * See Section 4.5 of the formal spec.
+   *
+   * Frozen-ness contract: values returned via the type-handler dispatch arm
+   * are guaranteed deep-frozen at this boundary, so callers do not each have
+   * to freeze. The class-registry fallback arm is a separate sibling branch
+   * and is intentionally NOT covered by this contract.
    */
   private deserialize(
     data: JsonWireValue,
--- a/packages/data-model/test/json-encoding-context.test.ts
+++ b/packages/data-model/test/json-encoding-context.test.ts
@@ -1656,6 +1656,55 @@ describe("JsonEncodingContext", () => {
   });
 
   // --------------------------------------------------------------------------
+  // TypeHandler.deserialize() deep-frozen contract (arm-1 only)
+  // --------------------------------------------------------------------------
+
+  describe("TypeHandler.deserialize() deep-frozen contract", () => {
+    // The contract is scoped to the type-handler dispatch arm only: anything
+    // returned via a registered `TypeHandler` is guaranteed deep-frozen at
+    // the `deserialize()` boundary, so callers do not each have to freeze.
+    // The class-registry fallback arm is a separate sibling branch and is
+    // intentionally NOT covered by this contract.
+
+    it("handler-produced value is deep-frozen at the boundary", () => {
+      // `/EpochNsec@1` dispatches through a registered TypeHandler (arm-1);
+      // the reconstructed FabricEpochNsec must be deep-frozen on return.
+      const result = fromWireFormat(
+        { "/EpochNsec@1": "AA" } as JsonWireValue,
+      );
+      expect(result).toBeInstanceOf(FabricEpochNsec);
+      expect(Object.isFrozen(result as object)).toBe(true);
+    });
+
+    it("lenient-mode ProblematicValue from a handler is deep-frozen", () => {
+      // `/BigInt@1` with non-string state fails handler validation; the
+      // lenient catch produces a ProblematicValue -- still an arm-1 return,
+      // so the contract deep-freezes it (not a crash: it is the value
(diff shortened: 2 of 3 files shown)


packages/data-model/test/json-encoding-context.test.ts
```diff
@@ -1655,6 +1655,55 @@ describe("JsonEncodingContext", () => {
...
+    // intentionally NOT covered by this contract.
+
+    it("handler-produced value is deep-frozen at the boundary", () => {
+      // `/EpochNsec@1` dispatches through a registered TypeHandler (arm-1);
+      // the reconstructed FabricEpochNsec must be deep-frozen on return.
+      const result = fromWireFormat(
+        { "/EpochNsec@1": "AA" } as JsonWireValue,
+      );
+      expect(result).toBeInstanceOf(FabricEpochNsec);
+      expect(Object.isFrozen(result as object)).toBe(true);
```
danfuzz: Shouldn't this check `isDeepFrozen()` and not just regular `Object.isFrozen()`? Likewise other items below.

packages/data-model/json-encoding-context.ts
danfuzz, line 381: Please note the frozen-ness contract here, and update the spec in a targeted way. (We'll do a more complete spec update later; I just don't want to lose the intentionality of this in the meantime.)

packages/data-model/test/json-encoding-context.test.ts
danfuzz: Done in `304c0c6ee`. All three assertions in the U2-added `describe("TypeHandler.deserialize() deep-frozen contract")` block now use `isDeepFrozen(result)` instead of `Object.isFrozen(result)` (this site + the two "likewise other items below" siblings in that describe), with `isDeepFrozen` imported from `../deep-freeze.ts`. The assertions were under-verifying — the values genuinely ARE deep-frozen via U2's arm-1 `deepFreeze` wrap, so the strengthened assertions pass (data-model 57/1310/0). Scope note: the conversions are confined to U2's own describe block; the pre-existing `Object.isFrozen` assertions in other (non-U2) describe blocks are out of this PR's scope and untouched.

packages/data-model/json-encoding-context.ts
danfuzz: Done in `304c0c6ee`, both parts: (a) the `deserialize()` method doc-comment now states the frozen-ness contract — values returned via the type-handler dispatch arm are guaranteed deep-frozen at this boundary (callers need not each freeze), and the class-registry fallback arm is explicitly noted as a separate sibling branch intentionally NOT covered. (b) A targeted, minimal spec note at `docs/specs/space-model-formal-spec/1-fabric-values.md` §4.5 (the private `deserialize()` type-handler-arm step): records the arm-1 deep-frozen contract (handler-produced value + lenient-mode `ProblematicValue`), scopes it to this arm only with the class-registry fallback (step 4) explicitly not covered, and explicitly flags that a more complete spec treatment of egress freezing is deferred — per your "targeted way / we'll do a more complete spec update later." No behavior change; gates green (fmt/lint/check + data-model 57/1310/0).

PR #3629, opened 2026-05-19
CT-1599: Fix tab focus order for form controls

bfollington:
## Summary
- move cf-input, cf-textarea, and cf-select away from delegatesFocus so their hosts remain reliable keyboard tab stops
- add Parking Coordinator fallback cells for optional spots/people/requests inputs to avoid undefined collection crashes
- keep Add Person text fields immediate to avoid blur-time reactive updates during Tab navigation
- update UI testing docs for the host-owned focus behavior

## Verification
- deno test --allow-env --allow-ffi --allow-read --allow-write --allow-run packages/ui/src/v2/components/cf-input/cf-input.test.ts packages/ui/src/v2/components/cf-textarea/cf-textarea.test.ts
- deno check packages/ui/src/v2/components/cf-select/cf-select.ts packages/ui/src/v2/components/cf-input/cf-input.ts packages/ui/src/v2/components/cf-textarea/cf-textarea.ts
- deno fmt --check docs/development/UI_TESTING.md packages/ui/src/v2/components/cf-input/cf-input.ts packages/ui/src/v2/components/cf-textarea/cf-textarea.ts packages/ui/src/v2/components/cf-select/cf-select.ts packages/ui/src/v2/components/cf-input/cf-input.test.ts packages/ui/src/v2/components/cf-textarea/cf-textarea.test.ts packages/patterns/factory-outputs/parking-coordinator/main.tsx
- deno task cf check packages/patterns/factory-outputs/parking-coordinator/main.tsx --no-run
- deno task cf test packages/patterns/factory-outputs/parking-coordinator/main.test.tsx
- manual browser retest in local piece confirmed initial fix

Linear: CT-1599

---
## Summary by cubic
Fixes keyboard tab order for `cf-input`, `cf-textarea`, and `cf-select` by making the host the tab stop and forwarding focus to the native control (Linear CT-1599). Also fixes early programmatic focus in `cf-input`, adds safe fallbacks in Parking Coordinator, and updates docs/tests.

- **Bug Fixes**
  - Removed `delegatesFocus`; host forwards focus to the native control. Overrode `focus()` (respects `disabled`) and preserved early programmatic focus in `cf-input`.
  - Parking Coordinator: default `spots` to `DEFAULT_SPOTS`; `people`/`requests` to empty `Writable` cells; set `timingStrategy="immediate"` on "Add Person" inputs.
  - Updated UI testing docs and added browser tests for host-owned focus, disabled behavior, and early focus in `cf-input`.

<sup>Written for commit 211a96b11a9176f18909a3824933988be8d739a3. Summary will update on new commits. <a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3629?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/docs/development/UI_TESTING.md
+++ b/docs/development/UI_TESTING.md
@@ -88,8 +88,8 @@ const button = await page.$("[data-cf-button]", {
 
 Playwright's `fill()` requires a native `<input>` or `<textarea>`. Since
 `cf-input` and `cf-textarea` are custom elements, use
-`pressSequentially()` on the locator instead — `delegatesFocus` forwards
-focus to the inner native input automatically:
+`pressSequentially()` on the locator instead. The custom element host is the
+semantic tab stop and forwards focus to the inner native input automatically:
 
 ```typescript
 const input = page.getByRole("textbox", { name: "Email" });
--- a/packages/patterns/factory-outputs/parking-coordinator/main.tsx
+++ b/packages/patterns/factory-outputs/parking-coordinator/main.tsx
@@ -252,7 +252,11 @@ export const DEFAULT_SPOTS: ParkingSpot[] = [
 // ============================================================
 
 export default pattern<ParkingCoordinatorInput, ParkingCoordinatorOutput>(
-  ({ spots, people, requests }) => {
+  ({ spots: inputSpots, people: inputPeople, requests: inputRequests }) => {
+    const spots = inputSpots ?? Writable.perSpace.of(DEFAULT_SPOTS);
+    const people = inputPeople ?? Writable.perSpace.of<Person[]>([]);
+    const requests = inputRequests ?? Writable.perSpace.of<SpotRequest[]>([]);
+
     const nowTimestamp = wish<number>({ query: "#now" });
     const todayStr = computed(() =>
       toLocalDateStr(nowTimestamp.result || safeDateNow())
@@ -1694,6 +1698,7 @@ export default pattern<ParkingCoordinatorInput, ParkingCoordinatorOutput>(
                                   <cf-input
                                     $value={newPersonName}
                                     placeholder="Full name"
+                                    timingStrategy="immediate"
                                     style="width: 100%;"
                                   />
                                 </cf-vstack>
@@ -1707,6 +1712,7 @@ export default pattern<ParkingCoordinatorInput, ParkingCoordinatorOutput>(
                                   <cf-input
                                     $value={newPersonEmail}
                                     placeholder="email@company.com"
+                                    timingStrategy="immediate"
                                     style="width: 100%;"
                                   />
                                 </cf-vstack>
@@ -1730,6 +1736,7 @@ export default pattern<ParkingCoordinatorInput, ParkingCoordinatorOutput>(
                                     $value={newPersonPriority}
                                     type="number"
                                     placeholder="1"
+                                    timingStrategy="immediate"
                                     style="width: 5rem;"
                                   />
                                 </cf-vstack>
@@ -1759,6 +1766,7 @@ export default pattern<ParkingCoordinatorInput, ParkingCoordinatorOutput>(
                                 <cf-input
                                   $value={newPersonPreferences}
                                   placeholder="e.g. 1, 5"
+                                  timingStrategy="immediate"
                                   style="width: 100%;"
                                 />
                               </cf-vstack>
--- a/packages/ui/src/v2/components/cf-host-accessibility.browser.test.ts
+++ b/packages/ui/src/v2/components/cf-host-accessibility.browser.test.ts
@@ -74,3 +74,21 @@ Deno.test("form control host focus forwards to native shadow controls", async ()
   textarea.remove();
   select.remove();
 });
+
+Deno.test("cf-input preserves early programmatic focus before first render", async () => {
+  if (typeof document === "undefined") {
+    return;
+  }
+
+  const input = document.createElement("cf-input") as CFInput;
+  document.body.append(input);
+
+  input.focus();
+  assertEquals(document.activeElement, input);
+
+  await input.updateComplete;
+  await nextFrame();
+  assertEquals(input.shadowRoot?.activeElement?.tagName, "INPUT");
+
+  input.remove();
+});
--- a/packages/ui/src/v2/components/cf-host-accessibility.browser.test.ts
+++ b/packages/ui/src/v2/components/cf-host-accessibility.browser.test.ts
@@ -1,6 +1,11 @@
 import { assertEquals, assertInstanceOf } from "@std/assert";
 import { CFButton } from "./cf-button/cf-button.ts";
 import { CFInput } from "./cf-input/cf-input.ts";
+import { CFSelect } from "./cf-select/cf-select.ts";
+import { CFTextarea } from "./cf-textarea/cf-textarea.ts";
+
+const nextFrame = () =>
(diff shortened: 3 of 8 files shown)


bfollington:
## Code Review

### Overview
Removes `delegatesFocus: true` from `cf-input`, `cf-textarea`, and `cf-select` shadow roots. The host element remains the semantic Tab stop (with `tabindex` already wired in `_updateAccessibilityAttributes`), and a `focus` listener forwards focus to the inner native control. Also overrides `focus(options)` so programmatic focus still reaches the inner control. Adds defensive fallback cells and `timingStrategy="immediate"` to the Parking Coordinator "Add Person" inputs to keep Tab navigation from being interrupted by blur-time reactive updates.

### What looks good
- **Symmetry with existing pattern**: `cf-button` and `cf-toggle` already eschew `delegatesFocus` for the same `aria-hidden`/focus reason. This change brings the text inputs and select in line with that pattern.
- **Both code paths covered**: keyboard Tab → host receives `focus` event → forwarded; programmatic `el.focus(options)` → `override focus()`.
- **Host is already a tab stop**: `this.tabIndex = this.disabled ? -1 : 0` is set in `_updateAccessibilityAttributes()` for all three components, so removing `delegatesFocus` is safe — Tab still lands on the host.
- **Test + doc updates** track the behavior change. Stale comments in the diff are replaced with accurate ones.
- **Parking Coordinator fallbacks** use `Writable.perSpace.of(...)` rather than plain arrays, so downstream `.get()`/`.set()` paths still work when the optional inputs are missing.

### Issues / suggestions

**1. `cf-checkbox`, `cf-radio`, `cf-switch` still use `delegatesFocus: true`**
These form controls participate in the same Tab sequence and (if the original report applies generally) could have the same problem. Either out-of-scope on purpose, or worth a follow-up — please clarify in the PR body or file a ticket so the inconsistency doesn't fossilize.

**2. The `event.target !== this` guard is effectively dead code**
`focus` does not bubble, so within the host's listener `event.target` is always the host. The guard isn't harmful, but if it's meant to protect against recursion through retargeting, a one-line comment would help; otherwise consider removing it for clarity. The `|| this.disabled` half is a useful guard — keep that.

**3. `override focus()` is not disabled-aware**
For symmetry with the listener path, `override focus()` will still try to forward to the inner control when the host is disabled:
\`\`\`ts
override focus(options?: FocusOptions): void {
  this.input?.focus(options);
}
\`\`\`
Native `<input disabled>` will refuse focus, so this is benign today. A tiny `if (this.disabled) return;` keeps the two paths consistent and avoids surprises if the inner control is ever swapped for a custom one.

**4. No automated coverage for the forwarding itself**
The new tests only assert `delegatesFocus !== true`. The actual regression — "Tab lands on host, focus forwards to inner control" — has no unit/integration coverage. A JSDOM test that does `host.focus()` and asserts `host.shadowRoot.activeElement === innerInput` would lock the behavior in. The verification log mentions a manual browser retest, but this is exactly the kind of thing UI tests should catch on regression.

**5. Parking Coordinator: rename + fallback**
- The rename to `inputSpots/inputPeople/inputRequests` plus local shadowing reads cleanly.
- Worth confirming that `ParkingCoordinatorInput` actually marks these as optional (otherwise the `??` branches are unreachable). The comment "to avoid undefined collection crashes" implies a real runtime case — a one-line repro/cause in the PR description or Linear would close the loop.
- `timingStrategy="immediate"` is applied to the four "Add Person" inputs only. If other text fields in this pattern (search box, edit dialogs, etc.) sit in the same Tab path, they will keep the original Tab+blur problem until they're also flipped. Worth a quick scan for other `<cf-input>` instances in the same tab sequence.

**6. Comment hygiene**
The replacement comment is good. Optionally call out the *why* (i.e. "`aria-hidden` can't be applied to focused elements, and `delegatesFocus` would mask the host as the tab stop") so future readers don't try to re-enable `delegatesFocus` thinking it's "cleaner".

### Risk assessment
- **Correctness**: low risk. Behavior parity with `cf-button` which has been in production.
- **A11y**: neutral. Host already advertises role + aria attributes via `_updateAccessibilityAttributes`; without `delegatesFocus`, the focused element from the AT's perspective is the host, which matches the existing ARIA surface.
- **Performance**: negligible — one extra event listener per element instance.
- **Security**: none.

### Verdict
Looks correct and well-scoped. Recommend: (a) decide on checkbox/radio/switch as a follow-up, (b) add a forwarding unit test, (c) optionally tighten the `disabled` guard in `override focus()`.

bfollington:
Addressed the review feedback in 257a4227e.\n\n- Removed the dead event-target guards from the host focus listeners.\n- Made programmatic focus disabled-aware for cf-input, cf-textarea, and cf-select.\n- Added browser coverage that calls host.focus() and verifies the native shadow control receives focus for input, textarea, and select.\n- Added comments explaining why these controls avoid delegatesFocus: the host owns the ARIA/tab surface and forwards focus to the native control.\n\nScope notes:\n- cf-checkbox, cf-radio, and cf-switch still use delegatesFocus and are intentionally out of scope for CT-1599. The reproduced issue was in the Parking Coordinator form path using cf-input/cf-select, and those binary controls have a different interaction surface.\n- The Parking Coordinator console error was from optional pattern inputs being absent in the local deploy path; spots/people/requests now get per-space defaults before any .map() use.\n- I scanned the Parking Coordinator cf-input usages. The timingStrategy=immediate change is limited to Add Person typed fields because that was the repro path that combined typing, blur, and tab navigation. The component focus fix applies globally.\n\nVerification run:\n- deno fmt --check packages/ui/src/v2/components/cf-host-accessibility.browser.test.ts packages/ui/src/v2/components/cf-input/cf-input.ts packages/ui/src/v2/components/cf-textarea/cf-textarea.ts packages/ui/src/v2/components/cf-select/cf-select.ts\n- deno test --allow-env --allow-ffi --allow-read --allow-write --allow-run packages/ui/src/v2/components/cf-input/cf-input.test.ts packages/ui/src/v2/components/cf-textarea/cf-textarea.test.ts packages/ui/src/v2/components/cf-select/cf-select.test.ts packages/ui/src/v2/components/cf-host-accessibility.browser.test.ts\n- deno run --allow-env --allow-read --allow-write --allow-run --allow-net packages/deno-web-test/cli.ts packages/ui/src/v2/components/cf-host-accessibility.browser.test.ts\n- deno task cf check packages/patterns/factory-outputs/parking-coordinator/main.tsx --no-run\n- deno task cf test packages/patterns/factory-outputs/parking-coordinator/main.test.tsx

bfollington:
Checked the cf-input early programmatic focus note and it was valid. Root cause: after connectedCallback the host is focusable, but before the first Lit render the native shadow <input> does not exist yet, so focus() only queried null and left focus on <body>.\n\nFixed in 211a96b11 by falling back to host focus when the inner input is not rendered yet, then forwarding to the native input after updateComplete if the host still owns focus. Added a browser regression test that failed before the patch and now passes.\n\nVerification:\n- deno fmt --check packages/ui/src/v2/components/cf-host-accessibility.browser.test.ts packages/ui/src/v2/components/cf-input/cf-input.ts\n- deno test --allow-env --allow-ffi --allow-read --allow-write --allow-run packages/ui/src/v2/components/cf-input/cf-input.test.ts packages/ui/src/v2/components/cf-host-accessibility.browser.test.ts\n- deno run --allow-env --allow-read --allow-write --allow-run --allow-net packages/deno-web-test/cli.ts packages/ui/src/v2/components/cf-host-accessibility.browser.test.ts

PR #3630, opened 2026-05-19
[codex] Add cf-harness web_fetch tool

mathpirate:
## Summary
- add a first-class `web_fetch` tool to cf-harness
- block local/private targets, credentials, and unsafe redirect targets before fetching
- retain bounded raw response content in tool-output artifacts while stripping raw content from model-facing tool output
- wire the tool through the registry, CLI allowlist, engine maps, diagnostics, docs, and tests

## Notes
This intentionally covers fetch only. `web_search` remains a separate provider decision and is not included here.

## Validation
- `PATH="/Users/gideonwald/.deno/bin:$PATH" deno fmt --check packages/cf-harness/README.md packages/cf-harness/deno.json packages/cf-harness/src/cli.ts packages/cf-harness/src/contracts/policy.ts packages/cf-harness/src/contracts/tool-descriptor.ts packages/cf-harness/src/diagnostics.ts packages/cf-harness/src/engine.ts packages/cf-harness/src/index.ts packages/cf-harness/src/prompt-loop.ts packages/cf-harness/src/tools/registry.ts packages/cf-harness/src/tools/web-fetch.ts packages/cf-harness/test/artifacts.test.ts packages/cf-harness/test/cli.test.ts packages/cf-harness/test/diagnostics.test.ts packages/cf-harness/test/prompt-loop.test.ts packages/cf-harness/test/tools-execution.test.ts packages/cf-harness/test/tools-registry.test.ts`
- `PATH="/Users/gideonwald/.deno/bin:$PATH" deno check src/main.ts`
- `PATH="/Users/gideonwald/.deno/bin:$PATH" deno task test`
- `git diff --check`

---
## Summary by cubic
Adds a first-class `web_fetch` tool to `cf-harness` for safe, bounded HTTP(S) fetching, now with a unified timeout that also covers redirect validation alongside DNS, connect, request write, headers, and body reads.

- **New Features**
  - `web_fetch`: public HTTP(S) fetch with limits (200KB default/1MB max bytes, 20k default/100k max text, 15s default/60s max timeout; up to 5 redirects, 50 links). Blocks localhost/private and credentialed URLs; resolves DNS to public IPs, pins connections, validates each redirect; supports html/xhtml/plain/json/+json. Returns status, content type, byte count, SHA‑256, extracted title/text/links, redirect chain, truncation flags. Artifacts retain bounded `rawContent`; model output omits it. Hardened: strict header size bound, strict chunked parsing (line/trailer caps), connections closed on abort; timeout now also covers redirect target validation.
  - Wiring and UX: added to registry, engine I/O maps, and `index` exports; CLI allowlists include `web_fetch` and `--allow-subagent-profile web_fetch`; CLI summaries show `url=...`; diagnostics classify `web_fetch` failures; docs/tests added.
  - Subagent: new `web_fetch` profile exposes only `web_fetch` (no host tools). System prompt warns to treat fetched content as untrusted.

- **Migration**
  - `web_fetch` is off by default for parent runs. Enable with `--allow-tool web_fetch` or delegate via `--allow-subagent-profile web_fetch`.

<sup>Written for commit 2a8fdef4461957c288416b300185708066dcbc1c. Summary will update on new commits. <a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3630?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/cf-harness/README.md
+++ b/packages/cf-harness/README.md
@@ -41,16 +41,21 @@ What works today:
   - `bash-no-sandbox` (provisional host shell for named subagent profiles only)
   - `read_file`
   - `view_image`
+  - `web_fetch` (explicit parent allowlist or `web_fetch` subagent profile only)
   - `read_skill_resource`
   - `edit_file`
   - `write_file`
   - `delegate_task`
 - targeted exact-string edits plus whole-file replace/create and append writes
 - initial and in-run image attachments for model vision-capable flows
+- bounded public HTTP(S) fetches through `web_fetch`, with redirect validation,
+  local/private target blocking, extracted text/links, and raw bounded response
+  retention in tool-output artifacts; `web_fetch` is intentionally not part of
+  the default parent tool surface
 - bounded OpenAI-compatible prompt/tool loop
 - single-child subagent delegation with fresh child prompt context, explicit
-  default/browser child profiles, retained child run references, and a sanitized
-  summary/state return channel
+  default/browser/web_fetch child profiles, retained child run references, and a
+  sanitized summary/state return channel
 - optional schema-validated subagent structured returns, with raw child return
   artifacts retained in the child run and open-ended strings linkified before
   the parent sees them
@@ -97,6 +102,7 @@ What is not done yet:
   subagent profile
 - dynamic/model-driven Agent Skills activation
 - skill script execution
+- `web_search`
 - parallel child orchestration
 - app UI event provenance
 - streaming model responses
@@ -283,6 +289,22 @@ deno task run -- \
   --allow-tool delegate_task \
   --allow-subagent-profile browser \
   --prompt "Delegate browser inspection of the local app and summarize the result."
+```
+
+The `web_fetch` profile is the preferred first-pass path for web page
+inspection. It gives the child only the `web_fetch` tool: no shell, no browser,
+no workspace reads, and no workspace writes. This keeps external web content in
+the child run and returns only the normal sanitized subagent summary/state to
+the parent. Use this profile when a task needs to inspect public HTTP(S) pages
+but does not need authenticated browser state or general web search.
+
+```bash
+deno task run -- \
+  --workspace /path/to/workspace \
+  --gateway-auth-mode none \
+  --allow-tool delegate_task \
+  --allow-subagent-profile web_fetch \
+  --prompt "Delegate inspection of https://example.com and summarize the result."
 ```
 
 Programmatic `delegate_task` calls may include `returnSchema`, a JSON Schema
--- a/packages/cf-harness/deno.json
+++ b/packages/cf-harness/deno.json
@@ -31,6 +31,7 @@
     "./sandbox": "./src/sandbox/docker-runsc.ts",
     "./tools": "./src/tools/registry.ts",
     "./tools/read-skill-resource": "./src/tools/read-skill-resource.ts",
+    "./tools/web-fetch": "./src/tools/web-fetch.ts",
     "./skills/registry": "./src/skills/registry.ts"
   }
 }
--- a/packages/cf-harness/src/cli.ts
+++ b/packages/cf-harness/src/cli.ts
@@ -31,10 +31,7 @@ import {
   HARNESS_SUBAGENT_PROFILES,
   type HarnessSubagentProfile,
 } from "./contracts/subagent.ts";
-import {
-  type BuiltinToolId,
-  DEFAULT_PARENT_TOOL_IDS,
-} from "./contracts/tool-descriptor.ts";
+import { type BuiltinToolId } from "./contracts/tool-descriptor.ts";
 import type {
   HarnessTranscriptEvent,
   HarnessTranscriptMessage,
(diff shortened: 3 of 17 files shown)


willkelly (changes requested):
Thanks for adding this; the tool shape, bounded body handling, raw-content stripping, and dedicated `web_fetch` subagent profile are all directionally good. I need to request changes on the network-safety boundary before this ships, because the current implementation still presents a best-effort DNS preflight as if it were SSRF-safe.

Blocking issues:

1. `web_fetch` still has a DNS validation-to-use race. `validatePublicHttpUrl()` resolves and approves the hostname before fetch (`packages/cf-harness/src/tools/web-fetch.ts:458`), but `fetchWithRedirects()` later calls `fetchFn(currentUrl, ...)` with the original hostname (`packages/cf-harness/src/tools/web-fetch.ts:645`). Redirects repeat the same pattern by validating `nextUrl` and then later fetching the hostname (`packages/cf-harness/src/tools/web-fetch.ts:676-696`). A rebinding host can resolve to a public IP during validation and to localhost/private infrastructure during the actual Deno fetch. Please either route through a connector/proxy that validates the actual dialed peer IP, or otherwise pin the connection path; DNS preflight alone is not enough for this security boundary.

2. The private/non-public IP filtering is incomplete. `blockedHostReason()` only handles IPv4-mapped IPv6 when the suffix remains dotted (`::ffff:127.0.0.1`), but URL canonicalization turns that into forms like `[::ffff:7f00:1]`, which fall through the current check (`packages/cf-harness/src/tools/web-fetch.ts:482-490`). The IPv4 classifier also only blocks a subset of non-global ranges (`packages/cf-harness/src/tools/web-fetch.ts:578-594`), missing ranges such as CGNAT, benchmarking, multicast, and reserved/documentation space. Please replace this with a comprehensive globally-routable IP classifier and add regression coverage for IPv4-mapped IPv6 and other non-global ranges.

I rechecked this against current head `43a827c0f645954ba9f3b9bb3ee6a0b79b9256b6`. LSP is clean for `web-fetch.ts`, and the previous focused cf-harness validation passed, so this request is specifically about closing the SSRF boundary rather than general implementation quality.

willkelly (changes requested):
Thanks for the substantial revamp here. The original DNS rebinding / preflight-vs-fetch concern and the incomplete IP filtering concern from my earlier review both look materially addressed in current head `0a619958d47d3b266ce0079a9ed14013e9ba0334`: the default path now uses the pinned `createPinnedPublicFetch` flow and the IP classifier/tests are much stronger.

I still need to request changes on the new low-level HTTP client surface before this ships:

1. **`timeoutMs` does not cover initial DNS or TCP connect.** `validatePublicHttpUrl()` runs before the abort timer is created (`packages/cf-harness/src/tools/web-fetch.ts:244-261`), and resolution awaits without a signal (`packages/cf-harness/src/tools/web-fetch.ts:502-535`, default DNS at `:544-547`). The pinned fetch path also re-resolves before checking abort (`:578-586`), then awaits `Deno.connect({ hostname: address, port })` without passing the signal (`:653-660`). Deno supports `signal` on `Deno.connect`, so please pass it through and make the DNS/connect phases fit within `timeoutMs`.

2. **Timeout classification is partial for TLS/write phases.** `startTls` has a close-on-abort hook (`packages/cf-harness/src/tools/web-fetch.ts:668-678`), but if closing the TCP connection makes `Deno.startTls` reject with a non-`AbortError`, the top-level catch reports `fetch_failed` instead of `timeout` (`:335-340`). `writeAll()` also awaits `conn.write()` without abort normalization (`:1035-1044`). Please normalize abort-triggered TLS/write failures to the same timeout path.

3. **Unsupported content-type responses can leak the pinned connection.** On unsupported or missing content type, the tool returns immediately (`packages/cf-harness/src/tools/web-fetch.ts:281-296`) without cancelling `response.body`. The manual stream closes the underlying connection only on close/cancel (`:822-831`, `:909-910`), so a server can keep that socket open after the tool has returned. Please cancel/close the body before returning these errors.

4. **Chunked framing is not bounded before `maxBytes` applies.** `readChunkedLine()` grows `state.buffered` until CRLF with no line cap (`packages/cf-harness/src/tools/web-fetch.ts:963-972`), and trailers are discarded without a byte/line limit (`:975-984`). Since `maxBytes` is enforced only after body chunks are delivered to `readResponseBody()` (`:1468-1476`), oversized chunk extensions/trailers can still drive unbounded buffering. Please add hard limits for chunk-size lines and trailers, with regression tests.

The rest is directionally good: TLS uses the original hostname for SNI/cert validation while connecting to the validated address, redirects are revalidated, and the new DNS rebinding/non-global IP tests cover the earlier blockers. I also reran focused validation locally on this head: LSP clean, `deno check packages/cf-harness/src/main.ts`, and the cf-harness tests passed (`185 passed`).

willkelly (changes requested):
Thanks for pushing this through another round — this is much better. The previous issues around initial DNS/connect/write timeout handling, unsupported-content cleanup, and chunked framing bounds all look addressed, and the focused/broader cf-harness validation passed locally (`190 passed`, plus LSP clean).

I found one small remaining timeout gap before I’d clear this:

- Redirect target validation still omits the active abort signal. In `packages/cf-harness/src/tools/web-fetch.ts:1403-1406`, the redirect path calls `validatePublicHttpUrl(nextUrl, options.resolveHostAddresses)` without `options.signal`. Since redirect locations are controlled by the remote server, a page can return a 302 to a hostname whose DNS lookup never completes, and that redirect DNS phase may outlive `timeoutMs`.

Suggested fix:

```ts
const validation = await validatePublicHttpUrl(
  nextUrl,
  options.resolveHostAddresses,
  options.signal,
);
```

Suggested regression test: first request returns `302 Location: https://slow.example/`; the resolver returns a public IP for the initial host, then never resolves `slow.example`; invoke with `timeoutMs: 20` and assert the tool returns `code: "timeout"` / `"web_fetch timed out"` within a watchdog.

This is intentionally a narrow request — the rest of the networking hardening looks good to me.

mathpirate:
Addressed the remaining redirect-timeout gap in 2a8fdef44.

What changed:
- redirect target validation now receives the active abort signal
- the 302 response body is cancelled before redirect DNS validation can block
- added a regression test for 302 -> slow.example with hanging DNS and timeoutMs: 20, asserting code: timeout / web_fetch timed out

Validation is green locally and on CI run 26244417378, including Cubic. @willkelly could you re-review the new head when you have a chance?

willkelly (approved):
Approved after re-reviewing latest head 2a8fdef4461957c288416b300185708066dcbc1c. The redirect target timeout blocker is fixed, and local validation passed: LSP clean, web_fetch tests 16/16, tools-execution 57/57, deno check packages/cf-harness/src/main.ts.

PR #3639, opened 2026-05-20
docs(fuse): specify CFC filesystem API semantics

willkelly:
## Summary
- add a CFC filesystem API semantics chapter to the FUSE spec
- define default commit-confirmed write behavior and standard errno mapping
- add the package-local FUSE reliability implementation plan and link it from `packages/fuse/README.md`
- address review feedback by clarifying handler write success boundaries, separating cached-read latency from commit-confirmed mutation deadlines, and documenting `.status` diagnostics for known failure vs timed-out unknown vs accepted-but-still-settling operations

## Verification
- `deno fmt --check`
- `deno task check`
- `GIT_MASTER=1 git diff --check`

## Performance Check
The performance check flagged an unrelated pattern-unit subtest wall-time outlier on this docs-only PR. No runtime/test code changed.

NEW_PERF_BASELINE: subtest: pattern-unit-1/pattern-unit-tests > packages/patterns/battleship/multiplayer/lobby.test.tsx = 34s

### DIFF
--- a/docs/specs/fuse-filesystem/4-read-write.md
+++ b/docs/specs/fuse-filesystem/4-read-write.md
@@ -128,6 +128,11 @@ target path parsing rules and examples.
    not auto-retry a timed-out handler write unless a future idempotency-key
    contract exists.
 
+Runtime handoff alone is not a success boundary for mounted handler writes.
+Downstream reactive effects may continue after the write returns successfully,
+but known handler rejection, sync/idle timeout, and timed-out-unknown outcomes
+must be visible as errno plus `.status`/diagnostic state.
+
 Handlers remain writable so legacy flows like
 `echo '{"message":"hi"}' > result/addItem.handler` keep working.
 
--- a/docs/specs/fuse-filesystem/4-read-write.md
+++ b/docs/specs/fuse-filesystem/4-read-write.md
@@ -121,8 +121,12 @@ target path parsing rules and examples.
    handler writes elsewhere in FUSE.
 4. Deduplicate `flush`/`release` so one buffered write triggers one handler
    invocation.
-5. Return success after the write has been handed to the runtime, subject to
-   immediate local/CFC validation failures.
+5. Return success only after the runtime has accepted the invocation and the
+   configured sync/idle boundary has completed. Handler rejection, runtime
+   timeout, transport failure, or CFC denial is returned as a normal filesystem
+   error. Because handler invocations are non-idempotent, the filesystem must
+   not auto-retry a timed-out handler write unless a future idempotency-key
+   contract exists.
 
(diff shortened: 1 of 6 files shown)


mathpirate:
High-level reaction from Codex on behalf of Gideon: this direction seems right to me. The important conceptual line is clear: FUSE is the projection/annotation/writeback participant, while gVisor remains the sandbox observation and enforcement authority. I also like making commit-confirmed writes the default POSIX-facing contract and treating local-ack/offline behavior as an explicit compatibility mode, because that gives cf-harness and ordinary filesystem clients a sane failure model.

From the cf-harness side, this gives us a good target: /workspace remains ordinary host/project material, /fabric can become the Fabric filesystem surface, cf-harness supplies trusted invocation/write-intent context, gVisor mediates reads/writes, and FUSE reconciles durable Fabric mutations. That separation feels compatible with the CFC label plumbing we have been adding.

The main thing I would keep explicit as this moves from docs into implementation is observability and error shape: agents need normal errno-style failures, but humans also need enough .status/diagnostic state to tell known failure from timed-out unknown from accepted-but-downstream-reactive-effects-still-pending. That distinction will matter a lot when we debug real cf-harness runs against /fabric.

docs/specs/fuse-filesystem/4-read-write.md
mathpirate, line 1: [P2] Align handler-write success boundary

Scalar and JSON writes now use the new commit/acceptance-confirmed contract, but handler writes still return success after being handed to the runtime. That conflicts with RELIABILITY_DESIGN.md, which says handler success means runtime acceptance plus the configured sync/idle boundary. Leaving this wording can preserve local-ack semantics for handler rejection/timeout even though the new default contract says those should become filesystem errors.

packages/fuse/RELIABILITY_DESIGN.md
```diff
@@ -0,0 +1,439 @@
...
+### Suggested Deadlines
+
+Initial values should be constants surfaced through `.status`:
+
+| Operation class                       | Soft deadline   |
+| ------------------------------------- | --------------- |
+| cached metadata/read                  | 50 ms           |
+| cold lookup/hydration                 | 2 s             |
+| `readdir` page                        | 5 s             |
+| cell content write/flush              | 30 s            |
```
mathpirate, line 116: [P3] Refresh stale write latency budget

This new 30s write/flush deadline conflicts with docs/specs/fuse-filesystem/7-open-questions.md, which still says write+flush should finish in under 200ms. That leaves two conflicting implementation targets in the docs set; update the open-question page to distinguish cached metadata/read latency from commit-confirmed mutation deadlines, or point readers to this reliability design.

PR #3646, opened 2026-05-20
feat(runner): persist scheduler state for rehydration

seefeldb:
## Summary

Adds persistent scheduler observations and storage-backed scheduler rehydration so runner actions can restore clean scheduler state without rerunning all nodes on startup.

Key pieces:
- persist scheduler action observations in memory v2, including no-op action runs
- add scheduler snapshot query APIs through memory v2 and runner storage providers
- rebuild scheduler reads, writes, materializer envelopes, debounce/throttle options, and dirty/stale state from persisted observations
- mirror read indexes across spaces so writes can dirty inactive cross-space readers
- push cross-space dirty markers back to the owner space while keeping recursive cross-space fanout driven only by actual committed writes
- harden rehydration around currentKnownWrites, fingerprint validation, aborted transactions, and unsubscribe races
- document the design and implementation notes in `docs/specs/persistent-scheduler-state.md`

## Why

Today a restarted piece reruns the scheduler graph because the in-memory dependency and dirty state is gone. Transactions already record most of the evidence the scheduler needs, so this adds durable scheduler observations tied to action transactions and uses memory v2 to rehydrate clean or dirty state precisely.

## Validation

- `HEADLESS=1 deno test -A packages/memory/test/v2-scheduler-state-test.ts packages/memory/test/v2-engine-revision-test.ts`
- `HEADLESS=1 deno test -A packages/runner/test/scheduler-observations.test.ts packages/runner/test/scheduler-core.test.ts packages/runner/test/scheduler-effects.test.ts`
- `deno task check`

---
## Summary by cubic
Adds durable scheduler observations in `@commonfabric/memory/v2` and runner rehydration so pieces restart cleanly without re-running the graph. The feature is opt‑in behind `persistentSchedulerState` (`EXPERIMENTAL_PERSISTENT_SCHEDULER_STATE`).

- **New Features**
  - Persist action observations in new scheduler tables with canonicalization, dedupe, bounded history, and no‑op batching; add persistence/ batching benchmarks.
  - Dirty persisted readers on commit; mirror read indexes across spaces so cross‑space readers dirtied in the owner space occur only on committed writes.
  - Expose `scheduler.snapshot.list` in `@commonfabric/memory/v2`; runner adds `listSchedulerActionSnapshots` and auto‑rehydrates subscriptions on subscribe with optional `deferInitialExecution`.
  - Rebuild reads/writes, materializers (including generated writers via `materializerWriteInputPaths`), debounce/throttle/noDebounce, dirty/stale state, and writers from persisted `currentKnownWrites`; deterministic action IDs for raw/JS actions.
  - Validate implementation/runtime fingerprints for observations and fall back to a fresh run on mismatch.
  - Advertise `persistentSchedulerState` in memory‑v2 hello; server returns its flags and only enforces data‑model compatibility; clients use compatible flags and suppress observation traffic/snapshot queries when disabled. Flags flow through runtime, CLI, Toolshed, shell, and background service.

- **Bug Fixes**
  - Persist post‑observation `currentKnownWrites`; recheck downstream readers after delayed computation commits.
  - Defer dependency collection during rehydration when using `deferInitialExecution`; preserve aborted‑action retry.
  - Keep event commits speculative; record changed writes at event commit and emit `scheduler.event.commit` telemetry; preserve speculative writes before a no‑op flush.
  - Require explicit materializer metadata; mark side‑writing generated `computed()`/`derive()` as materializers; narrow generated preflight reads; preserve handler preflight read schemas; schedule normal outputs when a materializer input dirties.
  - Harden persistent scheduler edges: clean up trigger indexes on unsubscribe and space removal, add explicit clear APIs, close remote sessions cleanly, dispose shell runtime safely, and stabilize CI (Pattern Reload Integration Tests, notebook reload split, reload budgets, runner test shard rebalancing).

<sup>Written for commit b75f0e3e79a1d94144622c7b94228372adb8e594. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3646?utm_source=github">Review in cubic</a>


seefeldb:
@cubic-dev-ai please review again.

seefeldb:
# Code Review

**+11,234 / −443  •  74 files**

## Overview

Adds durable scheduler observations and storage-backed rehydration so restarted pieces don't have to re-run their entire scheduler graph. New memory v2 endpoint `scheduler.snapshot.list`, new runner `listSchedulerActionSnapshots` + `deferInitialExecution` subscribe option, deterministic action IDs, implementation/runtime fingerprint validation, and ts-transformer marking of side-writing generated `computed/derive` as materializers. Behavior gated behind `EXPERIMENTAL_PERSISTENT_SCHEDULER_STATE`, plumbed through CLI/Toolshed/Shell/bg-svc. Also lands ~1.5k lines of spec docs.

The design is sound — observation persistence tied to action transaction commits, fresh-run fallback on fingerprint mismatch, mirror indexes for cross-space dirty propagation. The execution has several real risks worth addressing before this leaves the flag.

## High-severity findings

### 1. Handshake hard-fails on flag mismatch — no graceful degradation
`packages/memory/v2/handshake.ts:35–47`, `packages/memory/v2.ts:429–434`

`sameMemoryProtocolFlags` requires `left.persistentSchedulerState === right.persistentSchedulerState`; mismatch returns a `ProtocolError` and the connection dies. This means **a rolling deployment where server flips the flag on before clients are upgraded will sever all existing connections**, not gracefully fall back to flag-off behavior. The legacy `richStorableValues` alias survives, but the new flag has no such compat shim.

Recommend: server accepts hello with mismatched flag and suppresses observation traffic internally, matching what the server already does at the data-plane (`server.ts:685–689` strips observations when its own flag is off).

### 2. `scheduler.snapshot.list` is unbounded
`packages/memory/v2/engine.ts:1429–1494`, `packages/memory/v2/server.ts:823–840`, `packages/memory/v2.ts:318–323`

Query uses `.all()` with no LIMIT, no cursor, no parameter for paging in the `SchedulerActionSnapshotQuery` type. For a large piece, rehydration on startup could materialize gigabytes into a single response. Add cursor-based pagination before the flag flips on.

### 3. Rehydration has no timeout — `deferInitialExecution` can hang forever
`packages/runner/src/scheduler.ts:526, 542–543, 714–761`

If the storage provider wedges or the snapshot fetch never settles, the deferred action never runs and never gets fresh-scheduled. There's no fallback timer and no manual release API. Worst case: a piece that fails to load *one* snapshot becomes permanently dirty-but-unscheduled. Add a bounded timeout (e.g., 5–10s) after which deferral releases and the action falls back to a fresh run.

### 4. Cross-space mirror has no cleanup hook on space unload
`packages/runner/src/scheduler/trigger-index.ts:48–143`

`triggers` / `nonRecursiveTriggers` maps are plain `Map<SpaceScopeAndURI, ...>` populated by `addActionReads`, with no observable cleanup when a space unloads. Mirrored entries from offline spaces stay indexed indefinitely. Memory leak in long-running runtimes that mount/unmount many spaces (Toolshed). Confirm whether there's an unsubscribe path I missed, or add one tied to space lifecycle.

### 5. Event-commit speculative writes dirty downstream before server ack
`packages/runner/src/scheduler/events.ts:445–511` (especially 460–467)

`tx.commit().then(...)` fires `state.onEventCommitWrites(action, changedWrites)` on local apply, before server confirmation. Downstream readers get scheduled and may execute against speculative state. The retry path at 479–495 retries the event but doesn't roll back the dirty marks already issued. This is probably intentional for latency, but the spec's "ACTUAL committed writes" framing is misleading if "committed" here means "locally applied, server-pending."

Worth a comment in the code, and an explicit decision: if a speculative event later rejects, do we re-trigger anything to reconcile, or do we accept eventual consistency?

## Medium-severity findings

### 6. `scheduler.event.commit` telemetry is unbounded & ungated
`packages/runner/src/scheduler/events.ts:469–478`, `packages/runner/src/telemetry.ts:161–169`

Emits `writes: log.writes.map(formatEventCommitAddress)` — the full address list — for every event commit. No sampling, no size cap, no env-gated enable flag. Compare to `scheduler.event.preflight` which respects `eventPreflightTelemetryEnabled` (`pull-events.ts:58`). High-frequency event streams (collab, scroll) in large pieces will flood telemetry channels. Either gate it or cap the array.

### 7. `currentKnownWrites` has no reconciliation across function evolution
`packages/runner/src/scheduler.ts:597–603`, `packages/runner/src/scheduler/action-run.ts:500–516`

v1 of a function writes `{a, b}`; observation persisted. v2 only writes `{a}`. On rehydration, persisted `{a, b}` is re-applied as the action's known writes — effects on `b` stay subscribed even though v2 doesn't produce `b`. The implementation fingerprint *should* catch this (it includes telemetry ID), but if telemetry IDs are stable across compatible-looking source changes (function rename, body edit), divergence is possible.

Implementation notes line 337 already flags spec-code misalignment on pre-run vs post-run `currentKnownWrites` timing — this is the same area. Worth resolving both together. Add a test for write-set reduction across runs.

### 8. Engine reconcile-rows logic has ~250 lines of copy-paste
`packages/memory/v2/engine.ts:2220–2344` (reads), `2396–2515` (writes)

`reconcileSchedulerReadRows` and `reconcileSchedulerWriteRows` are near-identical templates differing only in column names. Extract a parameterized helper before the next round of changes lands.

### 9. Database operations lack error handling
`packages/memory/v2/engine.ts:1982` (`insertSchedulerObservationRow`), `2031` (update), `2065` (`recordSchedulerObservationReplay`)

`.run()` failures (lock contention, disk full, schema mismatch) are uncaught and bubble silently or crash the engine. Wrap in defensive handling consistent with other engine paths, or document that the engine is expected to die on storage failure.

### 10. Spec-code drift on `currentKnownWrites` timing
`docs/specs/persistent-scheduler-state/implementation_notes.md:337`

Implementation notes admit: "Persisting the pre-run writer-index value can stale rehydrated writer indexes by one run, so the code should be aligned to the spec in the next implementation slice." That's an acknowledged known gap. Either resolve it in this PR or add a follow-up issue link in the implementation notes.

## Lower-severity / worth knowing

- **engine.ts is 3700 lines** — separate concerns (migrations, observation reconciliation, snapshot query) into modules. Not blocking, but the file is at the limit of comprehensibility.
- **Mirror auth is permissive when no `authorizeSessionOpen` callback is configured** (`server.ts:1539–1552`). Graph structure leaks across spaces unconditionally. Document this as a deployment requirement.
- **Bench files were added** (`v2-scheduler-observation-persistence.bench.ts`, `scheduler-persistent-state.bench.ts`) — good. But no documented baseline numbers in the PR description, so regressions can't be tracked.
- **CI job `pattern-reload-integration` runs on every commit**, ~11s wall time. Acceptable for now; add a `paths:` filter if it grows.

## Passing checks

- Flag plumbing consistent across all 5 propagation layers (runtime/CLI/Toolshed/Shell/bg-svc); same env var name, same default-off.
- Materializer detection in `derive-strategy.ts` is conservative (capability-analysis-driven, not syntactic) — low false-positive risk.
- ts-transformer regression tests added for both positive and negative cases.
- `Page#reload()` + `disposeRuntime()` lifecycle is properly bracketed.
- Fingerprint mismatch → fresh-run fallback path is covered by test at `scheduler-observations.test.ts:1448`.
- Unsubscribe-during-rehydration race correctly handled by token check (`scheduler.ts:763–771`).
- Spec docs (1012 + 573 lines) are complete, no TODOs/TBDs, terminology consistent across both.
- Backward compatibility with flag off: no observable side effects detected.

## Recommendation

**Strong PR with real production risks before the flag can flip on.** Before enabling `persistentSchedulerState` for any user:

1. Fix handshake graceful-degradation (item 1)
2. Add pagination to `scheduler.snapshot.list` (item 2)
3. Add timeout to `deferInitialExecution` (item 3)
4. Add space-unload cleanup for the trigger index mirror (item 4)
5. Resolve the documented `currentKnownWrites` timing drift (items 7/10)

Items 5, 6, 8, 9 are pre-merge cleanup that doesn't block correctness but should not ship as-is when the feature graduates from experimental.

Tests are extensive (42 cases in `scheduler-observations.test.ts`, 1064 lines of memory v2 state tests). Coverage gaps: function evolution scenarios, materializer metadata removal across runs, rehydration cancellation, failed-observation persistence semantics.

seefeldb:
@cubic-dev-ai review

PR #3649, opened 2026-05-21
Fix cf-screen footer tab bar layout

bfollington:
Linear: CT-1613

## What changed

This fixes the layout contract between `cf-screen` and `cf-tab-bar` when an
inset tab bar is placed in the screen footer slot.

Standalone tab bars keep their fixed overlay behavior. Footer-slotted tab bars
now participate in layout, so `cf-screen` reserves footer space instead of
letting normal bottom content become inaccessible behind the mobile tab bar.

For inset footer tab bars, the reserved footer height still comes from the
footer-slotted `cf-tab-bar`: the host is in-flow, its container contributes
`--cf-tab-bar-height`, and inset bottom padding contributes the inset/safe-area
clearance.

The visual fade is implemented only in `cf-screen`. When the footer contains an
inset bottom `cf-tab-bar`, `cf-screen` lets its main scroller overlap into the
reserved footer region by `--cf-screen-footer-overlap` (defaulting to the tab
bar height), adds matching bottom padding so content remains reachable, and
masks only the bottom half of that overlapped region by default. This moves the
fade down into the tab-bar/footer area instead of starting at the top of the
reserved footer.

The catalog includes an `Inset footer in cf-screen` example using the real
footer slot contract.

## Validation

- Added browser layout coverage for standalone fixed tab bars.
- Added browser layout coverage for `cf-screen` with
  `cf-tab-bar slot="footer" variant="inset"`.
- Added browser assertions for footer reservation, scroller overlap, and the
  `cf-screen` scroller mask.
- Ran targeted format, type-check, lint, catalog check, and browser regression
  checks.
- Relaunched the local Tool Shed/Shell server and verified the catalog example
  at the CT-1613 local URL.

### DIFF
--- a/packages/patterns/catalog/stories/cf-tab-bar-story.tsx
+++ b/packages/patterns/catalog/stories/cf-tab-bar-story.tsx
@@ -12,6 +12,7 @@ export default pattern<TabBarStoryInput, TabBarStoryOutput>(() => {
   const activeTab1 = new Writable("home");
   const activeTab2 = new Writable("home");
   const activeTab3 = new Writable("home");
+  const activeTab4 = new Writable("home");
 
   const items = (
     <>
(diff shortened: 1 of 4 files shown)


bfollington:
## Code Review — CT-1613: Fix cf-screen footer tab bar layout

### Overview
Establishes a layout contract between `cf-screen` and `cf-tab-bar`: standalone tab bars stay `position: fixed`, but a `slot="footer"` tab bar becomes in-flow so `cf-screen` reserves footer space. For an inset bottom footer bar, `cf-screen` overlaps its scroller into the reserved region and masks a fade at the bottom. Includes a catalog story and browser layout tests.

### Strengths
- The CSS custom-property fallback chains are well-designed — `--cf-screen-footer-overlap` → `--cf-tab-bar-height` → default keeps the overlap synced to the actual bar height without JS.
- `_syncFooterFade` uses `position !== "top"` rather than `=== "bottom"`, so it correctly defaults an unset `position` attribute to the bottom-fade path.
- Good test coverage: standalone-fixed assertion plus a real footer-slot fixture verifying reservation, overlap, and mask.
- The catalog example uses the genuine footer-slot contract rather than a mock.

### Issues / Suggestions

**1. WIP comments will rot.** Both files carry comments like *"Edit these while dialing the footer reservation **before committing final values**."* (`cf-tab-bar.ts:53`, `cf-screen.ts:26`). This either signals the values aren't final (the PR is still a draft) or the comment is stale. Either finalize and drop the "before committing" framing, or keep the PR as draft until tuned. Per project convention (default to no comments), a plain note that these are tunable `@cssprop`-backed defaults would suffice — the JSDoc `@cssprop` lines already document them.

**2. `data-footer-fade` won't react to dynamic `variant`/`position` changes.** `_syncFooterFade` runs only on `firstUpdated` and `slotchange`, and reads `getAttribute("variant")`/`getAttribute("position")`. If a slotted tab bar's `variant` is changed after assignment (or set via property only, before reflection lands), the fade won't update. Fine for the static catalog/test usage, but worth a comment or an `attributeChangedObserver`/`MutationObserver` if dynamic variants are expected.

**3. Removal of `:not([variant="inset"])` from the border rules** (`cf-tab-bar.ts:105,110`) is a behavior-adjacent change not mentioned in the PR description. Inset bars now match `:host([position="bottom"]) .bar { border-top: ... }`, and only avoid a doubled/conflicting border because the later `:host([variant="inset"]) .bar { border: ... }` rule wins on source order at equal specificity. This works today but is fragile to rule reordering. A one-line comment explaining the reliance on cascade order, or keeping the explicit `:not([variant="inset"])`, would be safer.

**4. Pre-existing CSS indentation drift** in `cf-tab-bar.ts` (roughly lines 108–216 cascade-indent further on each rule). Not introduced here, and `deno fmt` doesn't touch CSS-in-template-literals — but since this PR already edits that block, a quick re-indent pass would be a low-cost cleanup.

### Minor
- `firstUpdated()` in `cf-screen.ts` omits `super.firstUpdated()` — consistent with `cf-tab-bar.ts`, and LitElement's base is a no-op, so harmless.
- Browser tests silently `return` when `document` is undefined; consistent with the suite's convention.

### Risk
Low. Standalone tab-bar behavior is preserved (test-covered), and the new layout path is gated on `slot="footer"` + `data-footer-fade`. Main residual risk is the dynamic-attribute staleness in (2).

PR #3653, opened 2026-05-21
feat(shell): add slug piece routes

seefeldb:
## Shell Slugs

This PR adds pretty slug URLs for pieces and CLI support for creating and updating those slug redirects.

A piece can now be opened in the shell at:

```text
/<space>/<slug>
```

For example:

```text
https://cf.dev/personal/project-notes
```

When the shell opens a canonical `fid1:` piece URL for a piece that has slug metadata, it replaces the URL with the slug form. When a slug is repointed, the shell reloads through the existing page loading flow and shows the new target.

## Slug Format

Slugs are intended for pretty URLs, so they are intentionally strict:

- lowercase letters, numbers, and hyphens only
- single hyphens between words
- no leading or trailing hyphen
- no `/` and no `:`
- at most 80 characters

Good examples:

```text
project-notes
weekly-plan-2026
counter-demo
```

Invalid examples:

```text
ProjectNotes
project_notes
project--notes
fid1:abc
notes/today
```

Any token containing `:` is treated as a URI or id, not as a slug.

## Create A Piece With A Slug

Use `cf piece new --slug <slug>` when creating a new piece:

```sh
cf piece new \
  --identity ./my.key \
  --api-url http://localhost:8000 \
  --space personal \
  --slug project-notes \
  ./patterns/project-notes/main.tsx
```

The command still prints the canonical piece id. It also writes slug metadata and creates the slug redirect document, so the piece can be opened at:

```text
http://localhost:8000/personal/project-notes
```

If the slug already exists, `piece new --slug` repoints that slug to the newly-created piece.

## Use Slugs Instead Of Piece IDs

Commands that take `--piece` can use a slug in place of the canonical piece id:

```sh
cf piece get --identity ./my.key --api-url http://localhost:8000 --space personal --piece project-notes value
cf piece inspect --identity ./my.key --api-url http://localhost:8000 --space personal --piece project-notes
cf piece setsrc --identity ./my.key --api-url http://localhost:8000 --space personal --piece project-notes ./patterns/project-notes/main.tsx
```

If the slug does not exist, commands that require an existing piece fail with a slug-specific error instead of silently inventing a target.

## Set Or Repoint A Slug

Use `cf piece set-slug <slug> <source>` to point a slug at an existing piece or cell link:

```sh
cf piece set-slug \
  --identity ./my.key \
  --api-url http://localhost:8000 \
  --space personal \
  project-notes \
  of:fid1:abc123...
```

The source can also be another slug:

```sh
cf piece set-slug --identity ./my.key --api-url http://localhost:8000 --space personal latest-notes project-notes
```

And it can point at a cell path inside a piece:

```sh
cf piece set-slug --identity ./my.key --api-url http://localhost:8000 --space personal published-title project-notes/value/title
```

By default, the slug redirect stores a link to the source you passed. If the source is itself a redirect, that relationship is preserved.

## Resolve Before Linking

Use `--resolve-before-linking` when you want the new slug to point directly at the source's current resolved target instead of pointing at the source token itself:

```sh
cf piece set-slug \
  --identity ./my.key \
  --api-url http://localhost:8000 \
  --space personal \
  stable-notes \
  latest-notes \
  --resolve-before-linking
```

This is useful when `latest-notes` is already a slug and you want `stable-notes` to keep pointing at whatever `latest-notes` resolves to now, even if `latest-notes` is repointed later.

## Link Behavior With Slugs

`cf piece link` resolves source and target piece tokens before creating the link:

```sh
cf piece link --identity ./my.key --api-url http://localhost:8000 --space personal project-notes/value dashboard/notes
```

Source endpoints may fall back to the original token where the command already supports non-existing sources, such as invented ids used with `--allow-non-existing`.

Destination endpoints must resolve. If the destination piece token is a slug and that slug does not exist, the link command fails instead of creating a link to a made-up destination.

## Validation

- `deno test -A packages/cli/test/charm.test.ts`
- `deno test -A packages/piece/test/test.ts`
- `deno test -A packages/piece/test/slug.test.ts`
- `deno test -A packages/runtime-client/backends/runtime-processor.test.ts`
- `deno test -A packages/runner/test/slug.test.ts`
- `deno test -A packages/shell/test`
- `PORT_OFFSET=745 FRONTEND_URL=http://localhost:5918 HEADLESS=1 deno task integration shell piece`
- `CF_CLI_INTEGRATION_USE_LOCAL=1 CF_CLI_INTEGRATION_SECTION=piece-links deno task integration cli`
- `deno task check`

NEW_PERF_BASELINE: step: shell integration = 83s
NEW_PERF_BASELINE: job: Package Integration Tests (shell) = 110s

### DIFF
--- a/packages/cli/commands/piece.ts
+++ b/packages/cli/commands/piece.ts
@@ -239,6 +239,7 @@ export const piece = new Command()
     "--root <path:string>",
     "Root directory for resolving imports. Allows imports from parent directories within this root.",
   )
+  .option("--slug <slug:string>", "Slug URL/address for this piece.")
   .action(async (options, main) => {
     setQuietMode(!!options.quiet);
     const spaceConfig = parseSpaceOptions(options);
@@ -249,11 +250,12 @@ export const piece = new Command()
         mainExport: options.mainExport,
         rootPath: options.root ? absPath(options.root) : undefined,
       },
-      { start: options.start },
+      { start: options.start, slug: options.slug },
     );
     render(pieceId);
+    const browserPieceRef = options.slug ?? pieceId;
     hint(cliText(`NEXT STEPS:
-  → Open in browser: ${spaceConfig.apiUrl}/${spaceConfig.space}/${pieceId}
+  → Open in browser: ${spaceConfig.apiUrl}/${spaceConfig.space}/${browserPieceRef}
   → Update code:     cf piece setsrc --piece ${pieceId} ${main} ...
   → Test a callable: cf piece call --piece ${pieceId} <callableName> ...
   → Inspect state:   cf piece inspect --piece ${pieceId} ...`));
--- a/packages/cli/commands/piece.ts
+++ b/packages/cli/commands/piece.ts
@@ -21,6 +21,7 @@ import {
   setCellValue,
   setHomePattern,
   setPiecePattern,
+  setPieceSlug,
   SpaceConfig,
   stepPiece,
 } from "../lib/piece.ts";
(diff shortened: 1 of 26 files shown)


jkomoros:
Hey Berni — found an asymmetry in this feature while wiring it up for cf-loom-mobile.

**Write side accepts any cell:**

```
cf piece set-slug --resolve-before-linking quick-capture home/capture
```

…succeeds and writes a redirect to a sub-pattern cell inside the parent piece. The shell's `handlePageGetSlug` reads the slug metadata directly, so `/space/quick-capture` navigates fine.

**Read side rejects non-piece targets:**

```
cf piece inspect --piece quick-capture
```

…blows up in `resolvePieceAddress` (`packages/piece/src/slugs.ts:120`) with `SlugResolutionError: not-piece` because `!target.getSourceCell()` — sub-pattern cells aren't process roots, so they don't have one.

This bites the loom mobile case directly because [`cf-loom-mobile`](https://github.com/commontoolsinc/loom/blob/main/.ops/patterns/cf-loom-mobile/cf-loom-mobile.tsx#L130-L154) returns sub-*pattern* results (renderable cells with `[UI]`), not separately-deployed sub-pieces. Your example "assumes the home pattern returns the sub pieces as quick-capture-piece" — I think that assumed a different pattern shape than ours.

**Two ways out as I see them:**

1. Relax `resolvePieceAddress` to allow any cell that has a `[UI]` (or just exists), since the shell already does. Unlocks the loom mobile use case without pattern restructuring.
2. Keep the read side strict but make `cf piece set-slug --resolve-before-linking` refuse non-piece sources so the asymmetry surfaces at write time. Crisper contract but pushes us back toward the launcher-design refactor.

Curious which way you'd prefer.

**Relevant files:**
- `packages/piece/src/slugs.ts:85-135` — `resolvePieceAddress`
- `packages/cli/lib/piece.ts:375-423` — `setPieceSlug`
- `packages/runtime-client/backends/runtime-processor.ts:688-704` — `handlePageGetSlug`

No test today covers slug-points-at-sub-cell — `piece/test/slug.test.ts:74` sets one with a sub-path but never round-trips it through `resolvePieceAddress`.

Context on the loom side: https://github.com/commontoolsinc/loom/pull/1718

PR #3660, opened 2026-05-21
feat(data-model): add cloneForIsolation for snapshot-style isolation cloning

danfuzz:
## Summary

Adds `cloneForIsolation()` to `@commonfabric/data-model` as a sibling to `cloneIfNecessary` and `cloneForMutation`. Its job is snapshot-style isolation cloning: produce a copy that's independent of source mutation, while preserving the identity of already-deep-frozen subtrees by reference.

```ts
export function cloneForIsolation<T extends FabricValue>(value: T): T;
```

Dispatch:

- Primitives and `FabricPrimitive` instances → returned as-is.
- Already-deep-frozen values (per `isDeepFrozen`) → returned by identity at every level of the recursion. This is the perf optimization the simpler `cloneIfNecessary(_, { frozen: false })` doesn't have (it `force: true`s a full deep clone, allocating new identities even for unchanged subtrees).
- `FabricInstance` → deep-cloned via `.deepClone(false)` (whose own semantics govern any nested values inside the wrapper).
- Plain objects and arrays → deep-cloned recursively, with cycle detection.
- Any other class instance is a `FabricValue` type violation upstream; the function **throws**.

Result mutability is intentionally "mixed": the returned tree is mutable along any path that wasn't already deep-frozen in the input; deep-frozen subtrees inside the result remain frozen by reference. Callers that need a uniformly mutable result should use `cloneIfNecessary(value, { frozen: false })` instead.

## Why

The v2-transaction storage layer has a local helper `isolateTransactionValue` with isomorphic semantics, but with two issues this new helper fixes:

1. **`FabricInstance` punt (bug):** the existing helper returns any non-plain-prototype value as-is — including mutable `FabricInstance` instances. That breaks isolation: the caller and the function share the same reference. `cloneForIsolation` delegates to the `FabricInstance` clone protocol like `cloneIfNecessary` already does.

2. **Silent type-violation:** the existing helper also returns non-`FabricValue` class instances (e.g. a stray `Date`) as-is rather than surfacing the upstream type bug. `cloneForIsolation` throws.

The runner-side swap (replace `isolateTransactionValue` with `cloneForIsolation`) lands as a follow-up PR — this PR is the data-model-only prerequisite.

## Test coverage

`packages/data-model/test/clone-for-isolation.test.ts` contributes 44 steps across:

- Primitives + `FabricPrimitive` (immutable-by-construction identity).
- Deep-frozen identity preservation (whole-input + per-subtree, including mixed-frozenness trees).
- Mutable input cloning (plain objects, arrays, null-prototype objects, sparse arrays, diamond references, direct cycles).
- `FabricInstance` handling (deep-frozen-identity, mutable rebuild, nested-in-parent).
- Type-violation throws (`Date`, `Map` as stand-ins for any non-`FabricValue`).

Parameterized across both legacy and modern data-model flag states.

## Test plan

- [x] `deno task test` in `@commonfabric/data-model` — 47 tests / 1757 steps pass.
- [ ] CI green.

---
## Summary by cubic
Adds `cloneForIsolation()` to `@commonfabric/data-model` for snapshot-style isolation cloning that shares already deep-frozen subtrees and correctly clones `FabricInstance` values. This sets up replacing the runner’s `isolateTransactionValue` in a follow-up.

- **New Features**
  - New API: `cloneForIsolation<T extends FabricValue>(value: T): T`.
  - Primitives and `FabricPrimitive` are returned as-is; deep-frozen subtrees are returned by identity; plain objects/arrays are deep-cloned with cycle/shared-ref preservation.
  - `FabricInstance` values are cloned via `.deepClone(false)` to avoid reference sharing.
  - Throws on non-`FabricValue` class instances (e.g., `Date`, `Map`).

<sup>Written for commit ec431a80dc7278ea94e04b80c30437d7ed2468e5. Summary will update on new commits. <a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3660?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/data-model/fabric-value.ts
+++ b/packages/data-model/fabric-value.ts
@@ -18,6 +18,7 @@ export {
 } from "./interface.ts";
 
 export {
+  cloneForIsolation,
   cloneForMutation,
   CloneForMutationError,
   type CloneForMutationErrorKind,
--- a/packages/data-model/test/clone-for-isolation.test.ts
+++ b/packages/data-model/test/clone-for-isolation.test.ts
@@ -1,0 +1,246 @@
+import { afterEach, beforeEach, describe, it } from "@std/testing/bdd";
+import { expect } from "@std/expect";
+import {
+  cloneForIsolation,
+  resetDataModelConfig,
+  setDataModelConfig,
+} from "../fabric-value.ts";
+import type { FabricValue } from "../fabric-value.ts";
+import { deepFreeze, isDeepFrozen } from "../deep-freeze.ts";
+import { FabricBytes } from "../fabric-bytes.ts";
+import { FabricEpochNsec } from "../fabric-epoch.ts";
+import { FabricError } from "../fabric-native-instances.ts";
+
+// ============================================================================
+// `cloneForIsolation` tests
+// ============================================================================
+//
+// Parameterized across both legacy and modern data-model flag states to keep
+// the contract durable under either flag setting.
+
+describe("cloneForIsolation", () => {
+  afterEach(() => {
+    resetDataModelConfig();
+  });
+
+  for (const modernMode of [false, true]) {
+    const label = modernMode ? "modern" : "legacy";
+
+    describe(`(${label} path)`, () => {
+      beforeEach(() => setDataModelConfig(modernMode));
+
+      // ----------------------------------------------------------------------
+      // Primitives and FabricPrimitive
+      // ----------------------------------------------------------------------
+
+      describe("primitives & FabricPrimitive", () => {
+        it("returns JS primitives unchanged", () => {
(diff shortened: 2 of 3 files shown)


danfuzz:
Closing in favor of the simpler approach: `isolateTransactionValue(value)` at v2-transaction's 8 call sites can be replaced directly with `cloneIfNecessary(value)` (defaults: `frozen: true`), since every caller either deepFreezes the result immediately, stores it as a frozen leaf in storage, or records it as an immutable snapshot. No `cloneForIsolation` needed -- it was solving a problem none of the callers actually have. Follow-up PR coming.

PR #3663, opened 2026-05-22
Replace isolateTransactionValue with cloneForIsolation

danfuzz:
## Summary

Replaces the v2-transaction layer's bespoke `isolateTransactionValue` helper with a new shared `cloneForIsolation()` primitive in `@commonfabric/data-model`, and fixes a latent data-model invariant violation that the swap surfaced.

This is the follow-up the [FabricError reshape (#3662)](https://github.com/commontoolsinc/labs/pull/3662) unblocked.

## Two commits

**1. `fix(data-model): shallow Error conversion yields a proper FabricValue`**

`shallowFabricFromNativeValueModern`'s `Error` case wrapped a native `Error` without converting its internals, so a `FabricError` could be constructed with a **raw `Error` `.cause`** — violating its own `cause: FabricValue` contract. The outer write-path walk treats a `FabricInstance` as atomic and never descends, so nothing fixed it downstream and the improper value reached storage. Now the internals are converted at construction (reusing the deep path's `rebuildFabricErrorDeep`). *By the time a `FabricValue` is constructed it should be a proper `FabricValue`.*

**2. `refactor(storage): replace isolateTransactionValue with cloneForIsolation`**

`isolateTransactionValue` duplicated data-model cloning logic with two flaws: it passed non-plain-prototype objects through **by identity** (so a mutable `FabricInstance` shared the caller's reference — broken isolation), and it **silently tolerated** raw non-`FabricValue` instances.

`cloneForIsolation()` is a sibling of `cloneIfNecessary`/`cloneForMutation`: a deep clone isolated from later source mutation, **preserving the identity of already-deep-frozen subtrees** by reference (the perf short-circuit `cloneIfNecessary(_, { frozen: false })` lacks — it force-clones everything). It dispatches `FabricInstance` through the clone protocol and **throws** on any non-`FabricValue`.

Wiring:
- `freezeReadValue` = `deepFreeze(cloneForIsolation(value))`.
- Write-activity / snapshot sites use `cloneForIsolation` (non-freezing — freezing them perturbed CFC write-authorization diffing for array pushes).
- The one genuinely-mutable root keeps `cloneIfNecessary(_, { frozen: false })`.
- `isolateTransactionValue` deleted.

The strict throw is intentional: it **surfaces** improper values instead of masking them — which is exactly what caught the raw-`Error`-cause bug fixed in commit 1.

## Test plan

- [x] `deno task test` data-model — 47 files green (incl. new `clone-for-isolation.test.ts`, 46 steps across both flag states).
- [x] `deno task test` runner — 455 passed / 0 failed (was failing the modern Error-cause case before commit 1).
- [x] `deno task test` full workspace — all 23 packages green (~80s).
- [x] `deno lint` clean on changed files.
- [ ] CI green.

## Note on `FabricError.fromNativeError`

It can still construct a *transient* improper `FabricError` (raw cause) that the conversion layer immediately rebuilds — never escaping to consumers. Fully removing that escape hatch (centralizing native→FabricError construction) is a larger, test-churny change left for later.

---
## Summary by cubic
Replaced the v2 transaction isolation helper with a shared `cloneForIsolation` in `@commonfabric/data-model` to properly isolate values while preserving frozen subtree identity. Also fixed an `Error` conversion bug that let raw causes into `FabricError`, restoring the data-model invariant.

- **Bug Fixes**
  - Shallow `Error` conversion now rebuilds internals so `FabricError.cause` is a proper `FabricValue`, preventing raw errors from reaching storage.

- **Refactors**
  - Added `cloneForIsolation` in `@commonfabric/data-model`: deep-clones only mutable parts, preserves deep-frozen subtrees by identity, clones `FabricInstance` via `.deepClone(false)`, and throws on non-`FabricValue`.
  - Replaced `isolateTransactionValue` in v2 transactions: `freezeReadValue` uses `deepFreeze(cloneForIsolation(value))`; write/snapshot paths use `cloneForIsolation` (non-freezing); kept `cloneIfNecessary(_, { frozen: false })` for the one mutable root; removed the old helper.

<sup>Written for commit 04186621a2ca122a73e25a7fbdc5cbaf1cd846e7. Summary will update on new commits. <a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3663?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/data-model/fabric-value-modern.ts
+++ b/packages/data-model/fabric-value-modern.ts
@@ -67,7 +67,19 @@ export function shallowFabricFromNativeValueModern(
       return value as FabricValueLayer;
 
     case NATIVE_TAGS.Error: {
-      const wrapped = FabricError.fromNativeError(value as Error);
+      // An `Error`'s internal state (`cause`, custom properties) is part of
+      // the `FabricError`'s value, so it must itself be proper `FabricValue`
+      // -- there is no valid "shallow" `FabricError` whose `.cause` is a raw
+      // `Error`. The outer structural walk treats a `FabricInstance` as atomic
+      // and won't descend into it, so the internals are converted here, at
+      // construction. (`freeze` still freezes only the wrapper, matching the
+      // shallow contract for containers; the now-`FabricValue` internals are
+      // left in whatever frozen state the deep converter produced.)
+      const wrapped = rebuildFabricErrorDeep(
+        FabricError.fromNativeError(value as Error),
+        new Map(),
+        false,
+      );
       if (freeze) Object.freeze(wrapped);
       return wrapped;
     }
--- a/packages/data-model/fabric-value.ts
+++ b/packages/data-model/fabric-value.ts
@@ -18,6 +18,7 @@ export {
 } from "./interface.ts";
 
 export {
+  cloneForIsolation,
   cloneForMutation,
   CloneForMutationError,
   type CloneForMutationErrorKind,
--- a/packages/data-model/test/clone-for-isolation.test.ts
+++ b/packages/data-model/test/clone-for-isolation.test.ts
@@ -1,0 +1,209 @@
+import { afterEach, describe, it } from "@std/testing/bdd";
+import { expect } from "@std/expect";
+import {
+  cloneForIsolation,
+  fabricFromNativeValue,
+  resetDataModelConfig,
+  setDataModelConfig,
+} from "../fabric-value.ts";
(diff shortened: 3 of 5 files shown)


danfuzz:
Superseded by the split: #3664 adds `cloneForIsolation` to data-model (its own PR, as requested), and #3665 (stacked on #3664) does the v2-transaction swap + the shallow-conversion properness fix. Same final code, cleaner separation.

PR #3670, opened 2026-05-22
fix(fuse): harden outage and background readiness behavior

willkelly:
## Summary
- harden background FUSE startup so a child must remain mounted/alive before `cf fuse mount --background` reports success
- record backend write outages in `.status`, transition disconnected mounts to read-only `EROFS`, and guard mutating callbacks before optimistic local mutation
- force-close runner storage after failed `loadManager` startup so pending session sync cannot hang cleanup

## Manual testing
- mounted FUSE against local toolshed, then stopped toolshed while writing through the mount
- verified `.status.connection.disconnected=true` after outage detection
- verified `.status.writes.flushErrors` increments and records the backend write failure
- verified the next write fails with `Read-only file system` instead of silently succeeding
- exercised slow-start/failing child startup with a proxy and confirmed background mount no longer reports a false successful empty mount
- verified cleanup with `deno task cf fuse status` / `cf fuse status`: `No active FUSE mounts.`

## Automated validation
- LSP diagnostics: clean on touched files
- `deno fmt --check` / `deno check` on touched files
- focused follow-on regression tests: `60 passed`
- `cd packages/fuse && deno task test`: `192 passed`
- `cd packages/cli && deno task test`: `36 passed, 1 ignored`
- `cd packages/runner && deno task test`: `456 passed` before the final FUSE-only follow-up; storage close focused test passed after
- `GIT_MASTER=1 git diff --check origin/main..HEAD`: clean
- final Oracle review: PASS on disconnected/read-only outage wiring after the flush/release follow-up

## Status vs overall goal
This PR hardens outage detection, degraded read-only behavior, background startup readiness, and cleanup observability after #3655. It still does **not** complete the larger reliability goal of fully commit-confirmed/default POSIX write semantics; fire-and-forget compatibility paths remain, now with better failure accounting and explicit read-only degradation after outage detection.

---
## Summary by cubic
Improves FUSE reliability: background mounts only succeed when the child stays mounted and alive, and backend outages now force writes into explicit read-only (`EROFS`) with clearer detection and status. Adds immediate storage shutdown and cleans up runtime on failed manager initialization.

- **Bug Fixes**
  - Background mount readiness: require stable “mounted” and alive PIDs; recheck status before success; if the child exits or status changes, remove mount state + child status; clean malformed sidecars on timeout.
  - Outage handling: classify connection write failures, set `.status.connection.disconnected=true`, increment `.status.writes.flushErrors`; guard open/write/flush/release/setattr/setxattr/removexattr/create/mkdir/unlink/rmdir/rename/symlink to return `EROFS` while disconnected; convert flush/compile/async connection failures to `EROFS`; probe reconnection with a fresh manager and piece list before resuming writes.
  - Error policy: use platform `ENOTSUP` for unsupported CFC writeback xattrs; return `EINVAL` for malformed payloads.
  - Cleanup: wrap `loadManager` with `withRuntimeCleanupOnFailure` to force-close runner storage via `StorageManager.closeNow()` and then dispose the runtime on startup failure; add provider `destroyNow()` to force-close pending destroys.

<sup>Written for commit 270eae4406939e9376a17947cbd12d99165806c8. Summary will update on new commits. <a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3670?utm_source=github">Review in cubic</a></sup>

NEW_PERF_BASELINE: subtest: pattern-unit-3/pattern-unit-tests > packages/patterns/factory-outputs/parking-coordinator/main.test.tsx = 96s

### DIFF
--- a/packages/cli/lib/piece.ts
+++ b/packages/cli/lib/piece.ts
@@ -96,6 +96,50 @@ export interface PieceResolutionDeps {
 
 const CLI_TRACE_TIMINGS = Deno.env.get("CF_CLI_TRACE_TIMINGS") === "1";
 
+interface DisposableRuntime {
+  dispose(): Promise<unknown>;
+  storageManager?: unknown;
+}
+
+function storageManagerCloseNow(
+  storageManager: unknown,
+): (() => Promise<unknown>) | undefined {
+  if (
+    typeof storageManager === "object" && storageManager !== null &&
+    "closeNow" in storageManager
+  ) {
+    const closeNow = Reflect.get(storageManager, "closeNow");
+    if (typeof closeNow === "function") {
+      return () => Promise.resolve(closeNow.call(storageManager));
+    }
+  }
+  return undefined;
+}
+
+export async function withRuntimeCleanupOnFailure<T>(
+  runtime: DisposableRuntime,
+  run: () => Promise<T>,
+): Promise<T> {
+  try {
+    return await run();
+  } catch (error) {
+    const cleanup = storageManagerCloseNow(runtime.storageManager) ??
+      runtime.dispose.bind(runtime);
+    await cleanup().catch(
+      (disposeError) => {
+        console.warn(
+          `loadManager cleanup failed: ${
+            disposeError instanceof Error
+              ? disposeError.message
+              : String(disposeError)
+          }`,
+        );
+      },
+    );
+    throw error;
+  }
+}
+
 async function timeCliPhase<T>(
   label: string,
   run: () => T | Promise<T>,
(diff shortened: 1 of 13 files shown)


packages/runner/src/storage/v2.ts
```diff
@@ -857,6 +867,14 @@ class Provider implements IStorageProviderWithReplica {
     await this.replica.close();
   }
 
+  async destroyNow(): Promise<void> {
```
mathpirate, line 870: Cubic identified this and I agree: `destroy()` sets `#destroyed` before awaiting `replica.synced()`, so if that path hangs, `StorageManager.closeNow()` later calls `Provider.destroyNow()` and immediately returns without calling `replica.closeNow()`. That defeats the force-close path this PR adds for stuck startup cleanup. Let `destroyNow()` still call `replica.closeNow()` when `#destroyed` is already true.

packages/fuse/mod.ts
```diff
@@ -75,6 +76,7 @@ import { decodeFuseComponent, encodeFusePathSegments } from "./path-codec.ts";
 import { buildNodeStat, getMountOwnership, nodeMode } from "./stat.ts";
 
 const encoder = new TextEncoder();
+const ENOTSUP = 95;
```
mathpirate: Cubic identified this and I agree: `ENOTSUP` is hard-coded to the Linux value `95`, but this FUSE implementation also targets macOS where `ENOTSUP` is `45`. Unsupported writeback xattrs can therefore return the wrong errno on macOS. Use a platform errno constant or branch on `Deno.build.os`.

packages/cli/lib/piece.ts
```diff
@@ -96,6 +96,50 @@ export interface PieceResolutionDeps {
...
+  ) {
+    const closeNow = Reflect.get(storageManager, "closeNow");
+    if (typeof closeNow === "function") {
+      return () => Promise.resolve(closeNow.call(storageManager));
+    }
+  }
+  return undefined;
+}
+
+export async function withRuntimeCleanupOnFailure<T>(
```
mathpirate, line 119: When `closeNow` exists, this helper calls only `storageManager.closeNow()` and skips `runtime.dispose()`. That avoids pending storage syncs, but it also skips the rest of `Runtime.dispose` cleanup: aborting queues, stopping the runner, disposing scheduler/harness state, popping the default frame, and resetting data-model config. In FUSE reconnect probes this can happen repeatedly in a long-lived process after failed `loadManager` attempts. Force-close storage first, then still run `runtime.dispose()`, or add a `Runtime.disposeNow()` that performs both.

mathpirate:
lgtm!

mathpirate (approved):

PR #3674, opened 2026-05-22
Codex/cf harness interactive protocol

mathpirate:
## Summary

Adds the labs-side foundation for using `cf-harness` as a CFC-enabled interactive chat backend, while keeping Loom product/UI exposure guarded.

This PR introduces:

- cf-harness interactive chat protocol contracts for sessions, turns, cancel, close, status, and structured events
- an in-process interactive chat service backed by `CfHarnessPromptLoop.runTranscript`
- an NDJSON stdio transport for headless callers such as Loom
- conservative tool/file event mapping for `write_file` / `edit_file` outcomes
- turn-scoped cancellation via `AbortController`, threaded through prompt loop and Chat Completions fetch
- Browser Access lease contracts and propagation into browser subagent prompts
- browser-profile turn rejection when no Browser Access lease is available
- first-class `comment-thread` chat context that resolves to canonical read-only tools and no subagents

## Motivation

Loom needs a cf-harness-backed interactive chat runtime before cf-harness can be exposed as a selectable Codex/Claude-style backend. The important inversion here is that Loom should not build UI around a partial runtime. Instead, cf-harness should first satisfy a durable backend/service contract that Loom can adapt to its existing chat API/event pipeline.

This PR gives cf-harness that initial runtime surface without exposing it in Loom settings yet.

## Design

The protocol is request/response plus structured events:

- `start_session`
- `start_turn`
- `cancel_turn`
- `close_session`
- `status`

The service owns in-memory interactive session state and emits structured chat events such as:

- `session_started`
- `turn_started`
- `assistant_delta`
- `assistant_completed`
- `tool_started`
- `tool_completed`
- `file_changed`
- `turn_canceled`
- `turn_completed`
- `session_closed`
- `error`

The stdio transport wraps this as NDJSON so Loom or other headless callers can launch cf-harness as a persistent process.

## Cancellation

`cancel_turn` is turn-scoped and keeps the session reusable.

Cancellation now creates one `AbortController` per active turn and passes its signal through:

```text
HarnessInteractiveChatService
  -> CfHarnessPromptLoop.runTranscript
  -> OpenAICompatibleGatewayClient.createChatCompletionJson
  -> fetch(...)


<!-- This is an auto-generated description by cubic. -->
---
## Summary by cubic
Introduces an interactive chat protocol and service for `cf-harness`, plus an NDJSON stdio transport for headless callers. Adds per-turn cancellation with AbortSignal propagation across the prompt loop and gateway retries, Browser Access lease enforcement, a read-only `comment-thread` policy, stricter validation, and safe file change events.

- **New Features**
  - Protocol contracts for requests (`start_session`, `start_turn`, `cancel_turn`, `close_session`, `status`) and structured events (assistant/tool/file/turn/session/errors).
  - In-memory interactive chat service backed by `CfHarnessPromptLoop.runTranscript`, with per-turn `AbortController`; abort-aware retry/backoff and signal forwarding to fetch.
  - NDJSON stdio transport with stricter validation (envelope/type checks, prompt-slot normalization, profile/tool allowlists) and clear errors on malformed input.
  - Browser Access lease contracts and enforcement for browser-profile turns; lease details flow into browser subagent prompts.
  - Read-only policy for `comment-thread` with limited tools and no subagents.
  - Validation of chat policy params with clear errors for invalid tools/subagents or missing browser leases.
  - Conservative file mutation events for `write_file` and `edit_file`.
  - Prompt loop and gateway accept `signal` and `browserAccess` options.
  - Public exports for the new contracts, interactive service, and NDJSON transport in `index.ts`.

- **Migration**
  - Backend only; no UI exposure yet.
  - Use NDJSON: send `start_session`, then `start_turn`; read the event stream; use `cancel_turn` to reuse sessions; call `close_session` when done.
  - Provide a Browser Access lease for any browser-profile turn; otherwise the service returns `browser_access_required`.

<sup>Written for commit 02787a1f187f7c21866496b24d53202d945d83bd. Summary will update on new commits. <a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3674?utm_source=github">Review in cubic</a></sup>

<!-- End of auto-generated description by cubic. -->



---BEGIN COPY-PASTE---
NEW_PERF_BASELINE: subtest: pattern-unit-4/pattern-unit-tests > packages/patterns/reading-list/reading-item-detail.test.tsx = 14s
NEW_PERF_BASELINE: job: CLI Integration Tests (fuse) = 127s
---END COPY-PASTE---

### DIFF
--- a/packages/cf-harness/src/interactive-chat-service.ts
+++ b/packages/cf-harness/src/interactive-chat-service.ts
@@ -1,0 +1,478 @@
+import {
+  CfHarnessPromptLoop,
+  type CreateHarnessPromptLoopOptions,
+  type HarnessPromptLoopResult,
+  type RunHarnessTranscriptOptions,
+} from "./prompt-loop.ts";
+import {
+  createHarnessChatErrorResponse,
+  createHarnessChatEventEnvelope,
+  createHarnessChatOkResponse,
+  createHarnessChatSessionStatus,
+  DEFAULT_HARNESS_CHAT_POLICY,
+  type HarnessChatErrorResponse,
+  type HarnessChatEventEnvelope,
+  type HarnessChatOkResponse,
+  type HarnessChatPolicy,
+  type HarnessChatRequestEnvelope,
+  type HarnessChatResponse,
+  type HarnessChatSessionStatus,
+  type HarnessChatStartSessionParams,
+  type HarnessChatStartTurnParams,
+  type HarnessChatStatusResult,
+  type HarnessChatStructuredEvent,
+  type HarnessChatTurnStatus,
+  reduceHarnessChatSessionStatus,
+} from "./contracts/interactive-chat.ts";
+import type {
+  HarnessAssistantTranscriptMessage,
+  HarnessToolTranscriptMessage,
+  HarnessTranscriptMessage,
+} from "./contracts/transcript.ts";
+
+export type HarnessInteractivePromptLoop = Pick<
+  CfHarnessPromptLoop,
+  "runTranscript"
+>;
+
+export type HarnessInteractivePromptLoopFactory = (
+  options: CreateHarnessPromptLoopOptions,
+) => HarnessInteractivePromptLoop;
+
+export type HarnessInteractiveChatEventListener = (
+  event: HarnessChatEventEnvelope,
+) => void | Promise<void>;
+
+export interface CreateHarnessInteractiveChatServiceOptions {
+  basePromptLoopOptions?: CreateHarnessPromptLoopOptions;
+  createPromptLoop?: HarnessInteractivePromptLoopFactory;
+  now?: () => string;
+  randomUUID?: () => string;
+  onEvent?: HarnessInteractiveChatEventListener;
+}
+
+interface HarnessInteractiveChatSessionRecord {
+  status: HarnessChatSessionStatus;
+  transcript: HarnessTranscriptMessage[];
+  activeTask?: Promise<void>;
+  canceledTurnIds: Set<string>;
+}
+
+const defaultPromptLoopFactory: HarnessInteractivePromptLoopFactory = (
+  options,
+) => new CfHarnessPromptLoop(options);
+
+const defaultRandomUUID = (): string => crypto.randomUUID();
+
+const activeTurnError = (
+  requestId: string,
+  session: HarnessChatSessionStatus,
+): HarnessChatErrorResponse =>
+  createHarnessChatErrorResponse(requestId, {
+    code: "turn_already_running",
+    message:
+      `session ${session.sessionId} already has active turn ${session.activeTurnId}`,
+    retryable: true,
+  });
+
+const sessionNotFoundError = (
+  requestId: string,
+  sessionId: string,
+): HarnessChatErrorResponse =>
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: packages/cf-harness/src/interactive-chat-service.ts [hunk 1]
```diff
@@ -0,0 +1,626 @@
...
+    }
+    if (turnId !== undefined && record.status.activeTurnId !== turnId) {
+      return turnNotFoundError(requestId, sessionId);
+    }
+    const activeTurnId = record.status.activeTurnId;
+    record.canceledTurnIds.add(activeTurnId);
+    record.activeAbortController?.abort(
+      createTurnAbortError(activeTurnId, reason),
+    );
+    await this.#emit(sessionId, activeTurnId, {
```
mathpirate, line 441: `cancelTurn` emits `turn_canceled` immediately, which reduces the session back to idle/reusable while the old `activeTask` may still be running. `startTurn` only checks `activeTurnId`, so a new turn can start before the canceled prompt-loop task exits. Since the abort signal currently reaches Chat Completions fetch but not all tool execution paths, the old turn can still mutate files or finish side effects concurrently with the new turn. Keep the session canceling/active until `activeTask` settles, or gate `startTurn` on `activeTask` rather than `activeTurnId` alone.

packages/cf-harness/src/interactive-chat-service.ts
```diff
@@ -0,0 +1,626 @@
...
+      workspace: params.workspace,
+      context: params.context,
+      model: params.model,
+      artifactRoot: params.artifactRoot,
+      capabilities: params.capabilities,
+      policy: resolveHarnessChatPolicy(params.policy, params.context),
+      browserAccess: params.browserAccess,
+      metadata: params.metadata,
+    });
+    this.#sessions.set(session.sessionId, {
```
mathpirate, line 343: `startSession` unconditionally replaces an existing record with the same `sessionId`. If a caller retries `start_session` or accidentally reuses an ID while the old session has a running turn, the old task keeps running but its abort controller/transcript/status record are lost; later old-task events can also reduce the replacement session because they use the same `sessionId`. Return an `invalid_request`/`session_exists` error unless the existing session is safely closed and idle.

packages/cf-harness/src/interactive-chat-stdio.ts
```diff
@@ -0,0 +1,184 @@
...
+  value: unknown,
+): value is HarnessChatRequestEnvelope =>
+  typeof value === "object" &&
+  value !== null &&
+  !Array.isArray(value) &&
+  "protocolVersion" in value &&
+  value.protocolVersion === HARNESS_CHAT_PROTOCOL_VERSION &&
+  "requestId" in value &&
+  typeof value.requestId === "string" &&
+  "method" in value &&
```
mathpirate: `isRequestEnvelope` accepts any string `method` and any object `params`. An unsupported method falls through `handleRequest` with no response, and malformed method-specific params can start a background task that fails after `start_turn` already returned `ok`. Since NDJSON callers can send arbitrary JSON, validate the method enum and params shape before dispatch, returning `invalid_request` for unsupported methods or missing fields like `start_turn.input.text`.

PR #3676, opened 2026-05-22
fix(ts-transformers): rewrite derive/computed to lift-applied form (CT-1615, Phase 1)

mathpirate:
## Summary

- Phase 1 of the three-phase derive→lift→selfcontained migration designed in `packages/ts-transformers/docs/derive-to-lift-design.md`.
- After this PR, no `__cfHelpers.derive(...)` calls exist in transformer-lowered output. Every reactive lift-style computation is `__cfHelpers.lift(...)({...})` (the "lift-applied" form).
- Both user-source `derive(x, cb)` and source-level `computed(() => expr)` lower through a single new `LiftLoweringTransformer` (formerly `ComputedTransformer`).
- Closes [CT-1615](https://linear.app/common-tools/issue/CT-1615/phase-1-rewrite-derive-calls-to-lift-applied-form).

## What changes

The pipeline emits a structurally different lowered form. Before:
```
__cfHelpers.derive({inputSchema}, {resultSchema}, {inputObject}, callback)
```
After:
```
__cfHelpers.lift({inputSchema}, {resultSchema}, callback)(inputObject)
```

Type arguments live on the inner lift call; the outer applied call carries the captured input object. This matches the canonical shape Phase 2 (hoisting) will build on: every `lift(...)(captures)` will become `__cfLift_N(captures)` where `__cfLift_N = lift(...)` is hoisted to module scope.

## Architecture

Commits (in order):

1. **`fix: lower computed() to lift-applied form`** — Renames `ComputedTransformer` → `LiftLoweringTransformer` (file: `src/computed/` → `src/lift/`). `computed(() => expr)` lowers to `__cfHelpers.lift(() => expr)({})`. The pipeline machinery learns to dispatch on both legacy derive shape and lift-applied shape so existing dispatchers keep working unchanged.

2. **`feat: lower user-authored derive() to lift-applied form`** — Extends `LiftLoweringTransformer` to also lower user-source `derive(x, fn)` and `derive(argSchema, resSchema, x, fn)`. Adds `TransformationOptions.narrowedWrapperTypeRegistry` to fix a type-inference regression for `Writable<T> | undefined` and other cell-like wrapped inputs (the schema generator and the inner-lift revisit path have opposite needs for the wrapper TypeNode's mapping; separate channels resolve the conflict).

3. **`refactor: rewrite remaining synthetic derive emission sites to lift-applied`** — Updates the three remaining emission sites: `createDeriveCall` (renamed `createLiftAppliedCall` in commit 5), the closure transformer's rebuild step, and the `rewrite-helpers.ts` computed-emit fallback. Schema injection's `kind: "derive"` branch reads `typeArguments` from the inner lift call. Three orphaned fixtures deleted (no `.input.tsx` siblings; coverage subsumed by active fixtures).

4. **`refactor: teach module-scope hoister the lift-applied shape`** — `getBuilderCallbackIndices` returns one of two `HoistTarget` variants. Lift-applied calls hoist the inner lift's callback while preserving the outer applied call's argument list. 33 active fixtures now exercise the path.

5. **`refactor: rename callKind "derive" → "lift-applied"`** — Mechanical rename of the `CallKind` union discriminator, ~15 dispatch sites, 8 helper functions (`getDeriveInputAndCallbackArgument` → `getLiftAppliedInputAndCallback`, `DeriveStrategy` → `LiftAppliedStrategy`, `createDeriveCall` → `createLiftAppliedCall`, `registerDeriveCallType` → `registerLiftAppliedCallType`, `inferDeriveResultTypeFromInitializer` → `inferLiftAppliedResultTypeFromInitializer`, etc.), and 3 file renames. `BoundaryKind`/`Owner` label vocabularies left alone (separate type vocabularies, out of scope).

6. **`docs: mark CT-1615 Phase 1 complete in design doc`** — design doc updates.

7. **`refactor: rename remaining derive-flavored internals (cleanup)`** — `preferDeriveWrappers` flag → `preferInputBoundWrappers` (the post-Phase-1 distinction is "captures threaded through input object" vs "zero-input thunk", not "derive vs computed"). Local variable renames. ~55 of 126 stale comment mentions of "derive" updated; the remaining 71 are accurate references to the user-API `derive` (still a runtime export), defensive legacy paths, or runtime registry entries.

## Bundled with this PR's design doc, scoped as a follow-up

A separate follow-up PR will:
- Remove the runtime `derive` export from `packages/runner` (now dead code for transformer-lowered patterns).
- Move the transformer's emitted form from `__cfHelpers.lift(...)({})` to `__cfHelpers.lift(...)()` (requires a small runtime change to `lift` so a no-arg application is equivalent to today's `computed`'s `argumentSchema: false` semantics).

Berni has confirmed both directions. Bundling them lets the runtime change land alongside the transformer-output change, which is the natural shape.

## Other follow-ups (in the design doc's tracker)

- Investigate the redundant schema re-narrowing path at `schema-injection.ts:3285`. CT-1615 worked around it via the new `narrowedWrapperTypeRegistry` channel. The deeper smell — re-narrowing on freshly-injected output — is a separable cleanup.
- Registry-architecture refresh (audited in commit 7's design doc note): the ts-transformers pipeline now threads eight registries through `TransformationOptions`. The marker-set trio has a clean `markX`/`isX` API with automatic cache invalidation; the other five use direct `.get`/`.set`. Lifting them to a consistent API and splitting `typeRegistry`'s triple-overloading would reduce future bugs of CT-1615's flavor.

## Test plan

- [x] `cd packages/ts-transformers && deno task test` → 288 passed, 0 failed.
- [x] Runtime CT-1585 repros pass: `deno task cf test --root packages/patterns packages/patterns/notes/note.test.tsx packages/patterns/notes/notebook.test.tsx packages/patterns/factory-outputs/parking-coordinator/main.test.tsx` → 97 passed, 0 failed, 2 skipped.
- [x] `grep -rln "__cfHelpers\.derive" packages/ts-transformers/test/fixtures/` returns no results.
- [ ] CI green on this PR.

---
## Summary by cubic
Completes CT-1615 Phase 1 by rewriting all reactive computes to the lift‑applied form; no `__cfHelpers.derive(...)` remains and all paths emit `__cfHelpers.lift(...)(input)`. Adds post‑review hardening and tests to guard type inference for unannotated cell inputs; no migration needed.

- **Refactors**
  - Introduced `getLiftAppliedInnerCall` and used it across lowering, schema‑injection, and pattern passes to consistently target the inner `lift(...)` call and avoid double prepends on parenthesized callees.
  - Guarded lift‑applied detection against chained calls (e.g. `lift(cb)(x)(y)`), removed `derive` from hoistable builders, and documented hoister invariants for schema‑injected lift‑applied calls.
  - Promoted `narrowedWrapperTypeRegistry` to `context.markNarrowedWrapper`/`lookupNarrowedWrapper`; widened literal input types before registration to restore derive‑style contextual typing; cleaned up remaining derive‑named helpers/comments to lift‑applied.

- **Bug Fixes**
  - `schema-generator`: emit `additionalProperties` for `IndexSignatureDeclaration` in `TypeLiteral` analysis.
  - Drop mismatched generics when lowering `computed<T>` to lift‑applied.
  - Prefer in‑scope aliases for synthetic type args, qualify commonfabric types as `__cfHelpers.X`, and mark synthetic lift‑applied calls to skip duplicate type shrinking.
  - Added targeted tests to lock `narrowedWrapperTypeRegistry` behavior and index‑signature schema emission.
  - Added tests guarding the `lowerDeriveCall` typeRegistry override so unannotated cell inputs retain `asCell` wrappers in lift‑applied inference.

<sup>Written for commit 28c2f80ae9887b528a260f30e0e32285beee9b56. Summary will update on new commits. <a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3676?utm_source=github">Review in cubic</a></sup>

---BEGIN COPY-PASTE---
NEW_PERF_BASELINE: job: Package Integration Tests (runtime-client) = 101s
---END COPY-PASTE---


mathpirate:
Thanks cubic — went through all six review items. Summary:

**Fixed in this PR:**

- **P2 (doc, `derive-to-lift-design.md:36`)** — refreshed `createDeriveCall` reference with a post-Phase-1 annotation pointing at `createLiftAppliedCall` / `src/transformers/builtins/lift-applied.ts`. Commit `49ea186e4`.
- **P2 (doc, `derive-to-lift-design.md:158` and the glossary at L315-317)** — same treatment for `ComputedTransformer` → `LiftLoweringTransformer`. Same commit.
- **P1 (`lift/transformer.ts:115`, type-arg position mismatch)** — real latent bug, pre-existing on main (the equivalent `computed → derive` lowering had the same defect since `derive<In, Out>` and `lift<In, Out>` share positional order, while `computed<R>` has only the result slot). Currently masked: when captures exist, `LiftAppliedStrategy` rewrites the lift's type args from capture+return inference; when captures don't exist, schema-injection derives schemas from the callback signature. All five real `computed<T>(...)` sites in `packages/patterns/` have captures. Fixed by dropping `node.typeArguments` at the lowering site so downstream re-inference runs unambiguously. Commit `37a834551`.
- **P2 (`pattern-body-reactive-root-lowering.ts:960`, `visited.expression` unwrap)** — cubic auto-marked this as addressed once the broader CT-1615 work landed; no separate action needed.

**Filed as follow-ups (out of Phase 1 scope):**

- **P1 (`pattern-body-reactive-root-lowering.ts:967`, `indexOf` on resolved callback)** — real but pre-existing. The same `indexOf` pattern existed pre-CT-1615 (originally `visited.arguments.indexOf(callbackArg)`, dating to PR #3154/#3567). `getLiftAppliedInputAndCallback` returns a *resolved* callback after unwrapping `harden(...)` and identifier indirection, so the raw argument array won't contain it. In practice all synthesis sites and most user code pass inline arrows, so the bug-shape doesn't fire today. Filed as [CT-1619](https://linear.app/common-tools/issue/CT-1619/indexof-on-resolved-callback-in-pattern-body-reactive-root-lowering).
- **P2 (`schema-generator.ts:815`, index-signature precedence)** — real but low-impact. The comment claims `stringIndex ?? numberIndex` precedence but the implementation gives "first-declared wins." `Record<K, V>` always produces a single index signature, so the divergence only matters for hand-written dual-index TypeLiterals, none of which exist in the current codebase. Filed as [CT-1620](https://linear.app/common-tools/issue/CT-1620/schema-generator-typeliteral-index-signature-precedence-is-declaration).

packages/ts-transformers/test/fixtures/jsx-expressions/helper-owned-jsx-iife-final-filter-capture.expected.jsx
```diff
@@ -58,7 +58,9 @@ export default pattern((__cf_pattern_input) => {
                     items: {
                         type: "string"
                     }
-                } as const satisfies __cfHelpers.JSONSchema, __cfHelpers.derive({
+                } as const satisfies __cfHelpers.JSONSchema, __cfHelpers.lift<{
+                    path: import("commonfabric").Cell<string[]>;
```
seefeldb: why isn't this using __cfHelpers?

packages/ts-transformers/test/fixtures/jsx-expressions/helper-owned-jsx-iife-final-flatmap-fallback-capture.expected.jsx
```diff
@@ -70,7 +70,10 @@ export default pattern((__cf_pattern_input) => {
                             required: ["name"]
                         }
                     }
-                } as const satisfies __cfHelpers.JSONSchema, __cfHelpers.derive({
+                } as const satisfies __cfHelpers.JSONSchema, __cfHelpers.lift<{
```
seefeldb, line 73: why is there an extra type declaration here (not a problem per se, just curious)

packages/ts-transformers/test/fixtures/jsx-expressions/logical-triple-and-chain.expected.jsx
```diff
@@ -64,42 +66,35 @@ export default pattern((_state) => {
...
+                    required: ["active", "verified"],
                     asCell: ["readonly"]
                 }
             },
             required: ["user"]
         } as const satisfies __cfHelpers.JSONSchema, {
             type: "boolean"
-        } as const satisfies __cfHelpers.JSONSchema, { user: user }, ({ user }) => user.get().active && user.get().verified), <span>Welcome, {__cfHelpers.derive({
+        } as const satisfies __cfHelpers.JSONSchema, ({ user }) => user.get().active && user.get().verified)({ user: user }), <span>Welcome, {__cfHelpers.lift<{
+            user: import("commonfabric").Cell<{ active: boolean; verified: boolean; name: string; }>;
```
seefeldb: many more of these, and it's always Cell<

packages/ts-transformers/test/fixtures/jsx-expressions/map-single-capture-no-name.expected.jsx
```diff
@@ -50,30 +50,23 @@ export default pattern((_state: any) => {
...
-                        properties: {
-                            id: {
-                                type: "string"
-                            },
-                            name: {
-                                type: "string"
-                            }
-                        },
-                        required: ["id", "name"]
+                        type: "unknown"
```
seefeldb, line 61: why is this being downgraded to unknown?

packages/ts-transformers/test/fixtures/jsx-expressions/map-single-capture.expected.jsx
```diff
@@ -50,30 +50,23 @@ export default pattern((_state) => {
...
-                        properties: {
-                            id: {
-                                type: "string"
-                            },
-                            name: {
-                                type: "string"
-                            }
-                        },
-                        required: ["id", "name"]
+                        type: "unknown"
```
seefeldb, line 61: same here, why unknown?

seefeldb:
# Code Review — PR #3676

**Scope:** 270 files, +4988/−2402. ~218 fixtures (mechanical rewrites, sampled and consistent) + ~51 source/doc files (reviewed in full).

## 1. Overview
Migrates every transformer-emitted reactive compute from `__cfHelpers.derive(...)` to the curried `__cfHelpers.lift(...)(input)` form. Concentrates in a renamed/extended `LiftLoweringTransformer`, `LiftAppliedStrategy`, schema-injection and hoister updates, plus a mechanical `CallKind "derive" → "lift-applied"` rename. Bundles a `schema-generator` fix for `IndexSignatureDeclaration` on synthetic `TypeLiteralNode`s, and a new `narrowedWrapperTypeRegistry` to keep schema-injection's lift-revisit from collapsing wrappers to `any`. Fixture sample (`computed-pattern-typed`, `nested-computed-output-maps`) matches the stated rewrite pattern precisely; type-args moved to inner `lift<…>`, captured input migrated to applied-call site.

## 2. Required before merge — correctness
- **`lift/transformer.ts:189–197`** — comment says "widened type" but `checker.getTypeAtLocation(inputArg)` is used directly with no `widenLiteralType` call. For `derive(literal, fn)` the param type is registered as the literal, not its widened form, diverging from pre-Phase-1 contextual typing. Add the widen.
- **`call-kind.ts:~1117`** — lift-applied detection on `target` doesn't guard against `target.expression` itself being a CallExpression. A theoretical `lift(cb)(x)(y)` chain would misclassify the outer-outer call. Add a single-application invariant.
- **Inconsistent wrapper stripping** between `getLiftAppliedInputAndCallback` (uses `stripWrappers`) and the bare `ts.isCallExpression(node.expression)` checks in `schema-injection.ts:~2244,~2305` and `lift/transformer.ts:~84,~289`. A parenthesized lift could trigger double schema injection. Extract one shared `getLiftAppliedInnerCall(call)` helper and route all four sites through it.
- **`module-scope-callback-hoisting.ts:~285`** — 3+-arg lift-applied hoist branch drops the `isSyntheticComputeCallback || hasSelfDescribingFunctionTypes` gate that the 1-arg branch retains, relying on "schemas-present implies self-describing." Add an explicit assert or restore the gate.
- **`computed(fn)` → `lift(...)({})` semantics** — pre-Phase-1 `computed` had `argumentSchema === false`; post-injection lift emits `{type:"object"|"unknown"}`. Confirm no runtime callsite branches on `argumentSchema === false` in a way that changes recompute triggers, error reporting, or sandbox eligibility. The bundled follow-up that moves emitted form to `lift(...)()` must land in lockstep, not as a later cleanup.
- **`schema-generator.ts:790–815`** — new `IndexSignatureDeclaration` branch claims to "mirror ObjectFormatter" but does not propagate JSDoc the way `object-formatter.ts:344-365` does. Either propagate JSDoc, or replace the "mirror" claim with an explicit "synthetic nodes have no docs to propagate" note.

## 3. Required before merge — code quality
- Stale "derive"-flavored helper names survive the rename and must be finished in this PR: `resolveDeriveCaptureNameCollisions` (`lift-applied-strategy.ts:~179`), `buildDeriveInputObject` (`:~199`), `rewriteNestedDeriveCallbackBodies` (`pattern-body-reactive-root-lowering.ts:~946`).
- `module-scope-callback-hoisting.ts:~201-225` keeps a `case "derive":` branch "for hypothetical future paths" though the design doc says no derive can reach the hoister. Replace with `assert false` (or delete) so a regression is loud, not silent.
- `expression-rewrite/rewrite-helpers.ts:~147` has a "Create computed(() => expression)" comment above code that no longer creates computed. Fix the comment.
- `core/mod.ts:67–80` doc explicitly notes `narrowedWrapperTypeRegistry` lacks cache-invalidation discipline. Promote access through a `context.markNarrowedWrapper(...)` marker method alongside the existing marker-trio shape so the contract is enforceable, rather than leaving direct `.set` calls scattered.

## 4. Required before merge — test coverage
- Add a targeted unit test for `narrowedWrapperTypeRegistry`: write-from-`applyShrinkAndWrap`, read-from-inner-lift-revisit. Integration coverage isn't sufficient for a workaround channel that explicitly bypasses the normal cache-invalidation path.
- Add a focused `schema-generator` test that routes a `Record<string, V>` shape through node-based analysis to lock in the new `IndexSignatureDeclaration` branch.

Sampled fixtures (`computed-pattern-typed`, `nested-computed-output-maps`, `closures/pattern-preserve-opaque-input`) are mechanical, consistent rewrites — no semantic changes hidden in fixture diffs. Three deleted fixtures verified covered by surviving siblings or by `pipeline-regressions.test.ts:57+` / `transform-guards.test.ts:41+` (the `manual-map-with-pattern` deletion is not a coverage regression).

## 5. Performance
No regressions. `narrowedWrapperTypeRegistry` is a `WeakMap`; one extra `.get` on the schema-injection lift-revisit hot path. `resolveHoistTarget` adds one `unwrapExpression` per call expression visited. All bounded constant overhead.

## 6. Security
No issue found.

## 7. Required before merge — verification
1. **Schema-injection idempotency** — the outer lift-applied branch prepends schemas to the inner lift, then `visitEachChild` recurses; verify the inner-lift dispatch doesn't re-inject. There is no obvious guard in `prependSchemaArguments`. Add one, or add a test that exercises a `derive<In, Out>(...)` source with pre-existing type arguments and asserts schemas are not double-prepended.
2. **Hoister + parenthesized lift** — confirm TS emit never inserts parens around `__cfHelpers.lift(...)` callee; if it can, the wrapper-strip mismatch (Required #3) is live and the shared helper must land before merge.
3. **`narrowedWrapperTypeRegistry` underlying smell** — the redundant schema re-narrowing path at `schema-injection.ts:~3285` is the actual root cause; link the tracked follow-up ticket in the design doc before merging so it isn't lost.

mathpirate:
Thanks for the review @seefeldb! Worked through all five comments — short version: Q1, Q2, Q3 are addressed in 49ea186e4 / e121e7b8c; Q4, Q5 turn out to be intended-but-previously-silently-skipped optimization. Details below.

**Q1, Q3 (`import("commonfabric").Cell<…>` vs `Cell<…>`)** — fixed in `e121e7b8c`. Added `UseAliasDefinedOutsideCurrentScope` to `DEFAULT_TYPE_NODE_FLAGS` in `ast/type-building.ts` so the printer prefers in-scope aliases over their qualified import forms. 23 fixture goldens regenerated, all cosmetic.

**Q2 (extra type-arg annotation on lift calls)** — the explicit `<Input, Output>` type-args are new in CT-1615 because the lift-applied lowering emits them via `createLiftAppliedCall`'s `buildInputTypeNode`. Pre-CT-1615, `__cfHelpers.derive(...)` had no type-arg annotation (the schemas-as-arguments form carried the same info via `toSchema(...)` calls instead). After CT-1615 the type-arg drives schema injection's input-schema synthesis, then gets erased at emit. Not a problem semantically; mentioned in the PR description as a deliberate shape choice.

**Q4, Q5 (`items: { type: "unknown" }`)** — this one was interesting. I initially thought it was a regression but tracing the actual code path revealed it's intended behavior previously masked by the printer-form difference.

Specifically: `isCellLikeTypeNode` in `type-shrinking.ts:690` recognizes Cell wrappers via `getTypeReferenceNodeName`, which handles `ts.isIdentifier` and `ts.isQualifiedName` but not the `ImportTypeNode` shape that `import(\"commonfabric\").Cell<...>` produces. Pre-CT-1615, the printer emitted the qualified-import form (because `DEFAULT_TYPE_NODE_FLAGS` didn't include `UseAliasDefinedOutsideCurrentScope`), so `isCellLikeTypeNode` returned `false` for the cell wrapper, `buildShrunkTypeNodeFromTypeNode`'s Cell-recursion branch (line 1200-1221) was skipped entirely, and the inner `{id;name}[]` survived intact through the shrink.

After the Q1/Q3 fix added `UseAliasDefinedOutsideCurrentScope`, the printer emits plain `Cell<...>` (which IS a TypeReferenceNode with an Identifier), so `isCellLikeTypeNode` now correctly recognizes the wrapper, recurses into the inner array, and the array-intrinsic-only-access optimization at line 1054-1066 fires (callback `({people}) => people.get().length > 0` only touches `.length`, so `allNonItem=true`, returns fresh `unknown[]`).

The new output is consistent with how the shrink treats every other `Cell<T[]>` site where the callback does only array-intrinsic reads — the optimization is supposed to drop element schemas in that case as a size optimization. The pre-CT-1615 \"richer\" output was incidental, caused by the printer-form mismatch hiding the wrapper-recognition path.

I confirmed this via side-by-side instrumentation of main vs CT-1615 with identical `argumentNode`/`paramSummary` inputs (`identityCellPaths=[], readPaths=[[\"people\",\"length\"]], capability=readonly`):
- Main result: `{ people: ReadonlyCell<{id;name}[]>; }` — wrapper recognition fails, element preserved by accident
- CT-1615 result: `{ people: ReadonlyCell<unknown[]>; }` — wrapper recognized, array-shrink optimization fires correctly

Happy to discuss whether the shrink optimization itself should be tuned (e.g., always preserve element shape under cell wrappers), but that's a broader behavioral change worth tracking separately. For Phase 1 I'd suggest landing the current behavior since it's now internally consistent across all synthesized-cell-wrapped-array sites.

CC: the foundational `syntheticLiftAppliedCallRegistry` registry from `633cc30d4` stays in (no consumer right now but it's positioned for protecting synthetic-emit calls from any future re-shrink mistakes; cheap insurance).

mathpirate:
Thanks @seefeldb! Worked through every item in the review. Status below — each section maps to your numbered headings.

## §2 Correctness — all addressed

- **§2.1 widenLiteralType** — fixed in [`13dd379f1`](https://github.com/commontoolsinc/labs/pull/3676/commits/13dd379f1). The `lift/transformer.ts:197` site now widens via `widenLiteralType` before registering against the callback parameter; for `derive(5, x => …)` schema injection now sees `number`, not the literal `5`.

- **§2.2 single-application invariant** — fixed in [`f65b0de92`](https://github.com/commontoolsinc/labs/pull/3676/commits/f65b0de92). Added `isMultiApplicationChain` guard in `call-kind.ts`; a chained `lift(cb)(x)(y)` shape no longer gets classified as `kind:"lift-applied"`. TS factory doesn't emit such chains today, but the guard makes the canonical single-application form explicit.

- **§2.3 inconsistent wrapper stripping** — fixed in [`b8c89357c`](https://github.com/commontoolsinc/labs/pull/3676/commits/b8c89357c). Extracted `getLiftAppliedInnerCall` in `ast/call-kind.ts` using `stripWrappers` consistently; all five sites (the four you named plus one related typeArgs read at `schema-injection.ts:3160`) now route through it. Parenthesized `(__cfHelpers.lift(...))(input)` would now be recognised consistently across the pipeline.

- **§2.4 hoister gate documentation** — addressed in [`f65b0de92`](https://github.com/commontoolsinc/labs/pull/3676/commits/f65b0de92). The 3+-arg branch now carries an explicit comment: schema injection is the upstream gate (its output IS the self-describing artifact), so the `isSyntheticComputeCallback || hasSelfDescribingFunctionTypes` check is redundant at that point. The 1-arg pre-injection branch retains its own check because schemas haven't been computed yet.

- **§2.5 computed → lift({}) runtime semantics** — verified, no code change. The only runtime site that reads `argumentSchema === false` is `runner.ts:2012` (`isValidArgument: module.argumentSchema === false || argument !== undefined`). Pre-Phase-1 `computed` had `argumentSchema: false` → `isValidArgument: true` unconditionally. Post-Phase-1 `lift(cb)({})` has `argumentSchema: { type: "object", properties: {} }` and input `{}` (always non-undefined) → second condition fires → `isValidArgument: true`. Same outcome. Confirmed no other runtime callsite branches on `argumentSchema === false`. The `lift(...)({})` → `lift(...)()` follow-up will land as part of the bundled `remove-derive-runtime-export` follow-up per the PR description.

- **§2.6 schema-generator JSDoc claim** — fixed in [`13dd379f1`](https://github.com/commontoolsinc/labs/pull/3676/commits/13dd379f1). Comment rewritten to drop the "mirror" claim and explicitly note that JSDoc propagation isn't possible for synthetic TypeLiteralNodes (no source-positioned declarations to read from).

## §3 Code quality — all addressed

- **§3.1 derive-flavored helper renames** — fixed in [`13dd379f1`](https://github.com/commontoolsinc/labs/pull/3676/commits/13dd379f1). Renamed the three you listed plus `rewriteDeriveCallbackComputedKeyAccesses` (same spirit). All sites updated; no stale references remain.

- **§3.2 hoister `case "derive":`** — fixed in [`13dd379f1`](https://github.com/commontoolsinc/labs/pull/3676/commits/13dd379f1). Removed `derive` from `HOISTABLE_BUILDER_NAMES` (so the function early-exits before reaching the switch). The unused `context` parameter is now prefixed with `_`. Note: I initially tried Berni's preferred `assert false`, but it broke validation tests where pipeline-stages run over source with 1-arg `derive(fn)` (malformed; passes through to validation). The "don't hoist derive at all" stance is functionally equivalent and avoids the false alarm; the underlying invariant is documented inline.

- **§3.3 stale "computed call" comment** — fixed in [`13dd379f1`](https://github.com/commontoolsinc/labs/pull/3676/commits/13dd379f1). Comment at `rewrite-helpers.ts:130` now says "Get result type for the synthetic lift-applied call we're about to emit."

- **§3.4 narrowedWrapperTypeRegistry marker promotion** — fixed in [`9a3107cee`](https://github.com/commontoolsinc/labs/pull/3676/commits/9a3107cee). Added `context.markNarrowedWrapper(node, type)` and `context.lookupNarrowedWrapper(node)`; both writer (type-shrinking.ts) and reader (schema-injection.ts) route through them. The registry-overview doc updated accordingly: narrowedWrapperTypeRegistry now lives in its own paragraph noting the marker-method access; the "open architectural improvements" list reduced to the three remaining direct-access registries.

## §4 Test coverage — both added

- **§4.1 narrowedWrapperTypeRegistry unit test** — added in [`cd6cfd168`](https://github.com/commontoolsinc/labs/pull/3676/commits/cd6cfd168). Three tests in `type-shrinking.test.ts`: happy-path round-trip (asserts identity, not just structural equality), negative lookup, and the writer's any/unknown guard.

- **§4.2 schema-generator Record<string,V> test** — added in [`cd6cfd168`](https://github.com/commontoolsinc/labs/pull/3676/commits/cd6cfd168). Two tests in `schema-generator.test.ts`: pure index-signature TypeLiteral, and mixed named-property + index-signature shape. Both via `generateSchemaFromSyntheticTypeNode` to lock in the node-based-analysis path's behavior.

## §7 Verification — all addressed

- **§7.1 schema-injection idempotency** — verified via existing fixture pair `schema-injection/builder-input-path-shrink.{input.tsx, expected.jsx}`. The input has `derive<Writable<{...}>, string>(...)` with explicit type args; the lowered output has exactly one pair of schemas, not double-prepended. If schemas got double-prepended, the golden would have 4 schemas and the test would catch it. (Per discussion with Gideon, the fixture pair IS the regression test for this — no additional targeted assertion needed.)

- **§7.2 hoister + parenthesized lift** — verified. TS factory does not insert parens around our synthesized `__cfHelpers.lift(...)` callee today (grep `createParenthesizedExpression` in the transformer source: only one site in `module-scope-function-hardening.ts:547`, an unrelated binary expression). The shared `getLiftAppliedInnerCall` helper (§2.3) defensively handles parens via `stripWrappers` should this change.

- **§7.3 narrowedWrapperTypeRegistry follow-up tracked** — filed as [CT-1621](https://linear.app/common-tools/issue/CT-1621/remove-redundant-schema-re-narrowing-in-schema-injections-inner-lift), linked from the design doc's open-questions tracker. The underlying smell (redundant re-narrowing at `schema-injection.ts:~3285`) is documented as the actual root cause for the registry's existence.

---

Net diff for these review-driven changes: 5 commits since your review (`13dd379f1`, `b8c89357c`, `f65b0de92`, `9a3107cee`, `cd6cfd168`). All 292 ts-transformers tests pass + 24 schema-generator tests pass + fmt/lint clean.

Earlier in the session I also addressed cubic's review (commits `49ea186e4`, `37a834551`, `e121e7b8c`, `633cc30d4`, `99d7a345e`) including the `__cfHelpers.X` qualification for synthesized type annotations, which addresses your Q1/Q3 directly.

Let me know if you'd like me to revisit any decision.

packages/ts-transformers/src/ast/call-kind.ts
```diff
@@ -1063,6 +1127,34 @@ function resolveExpressionKind(
...
+    // the SINGLE-application form is canonical lift-applied. If the
+    // outer-outer call's inner expression (`target.expression`) is itself
+    // a call AND that inner call's expression (`target.expression.expression`)
+    // is ALSO a call, we have an over-application and should NOT classify
+    // as lift-applied — even if the chain happens to resolve through a
+    // lift symbol via factory-following. (CT-1615 Berni review §2.2.)
+    if (
+      builderKind.builderName === "lift" &&
+      ts.isCallExpression(target) &&
+      !isMultiApplicationChain(target)
```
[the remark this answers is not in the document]
mathpirate: Verified directly: the guard works correctly for the chain cubic flagged. For `lift(cb)(x)(y)`:

- `target` = the outermost `(y)` call
- `target.expression` = `lift(cb)(x)` (a CallExpression)
- In `isMultiApplicationChain`: `innerCallee = lift(cb)(x)` ✓ → `innerCalleeCallee = lift(cb)` (also CallExpression) ✓ → returns true → guard blocks classification.

For canonical `lift(cb)(x)`:
- `target.expression` = `lift(cb)` → `innerCalleeCallee = lift` (Identifier, NOT CallExpression) → returns false → guard does NOT block (correct).

The chain is walked from outermost toward `lift`; the guard fires precisely when there are two CallExpression layers below the outermost call — which is the definition of multi-application. Cubic's directionality concern is unfounded. Resolving.

packages/ts-transformers/test/fixtures/jsx-expressions/helper-owned-jsx-iife-final-filter-capture.expected.jsx
mathpirate: Addressed in commit `99d7a345e` — synthesized type annotations now use the canonical `__cfHelpers.X` qualified form via the new `qualifyCommonFabricTypeRefs` post-processing pass. The bare `Cell<...>` was a half-fix; this is the principled one. See the consolidated reply on the PR for full context.

packages/ts-transformers/test/fixtures/jsx-expressions/logical-triple-and-chain.expected.jsx
mathpirate: Addressed in commit `99d7a345e` — synthesized type annotations now use the canonical `__cfHelpers.X` qualified form via the new `qualifyCommonFabricTypeRefs` post-processing pass. The bare `Cell<...>` was a half-fix; this is the principled one. See the consolidated reply on the PR for full context.

packages/ts-transformers/test/fixtures/jsx-expressions/helper-owned-jsx-iife-final-flatmap-fallback-capture.expected.jsx
mathpirate: Answer to the curiosity: the explicit `<Input, Output>` type-arg annotation is new in CT-1615 — it carries through `LiftLoweringTransformer`’s lowering of `derive<T,R>(...)` and through `createLiftAppliedCall`’s synthesis path. It feeds schema-injection’s input-schema synthesis (without it, the injection has to infer from the callback parameter alone). Pre-Phase-1, `__cfHelpers.derive(...)` carried no type-args because schemas-as-arguments did the same job. Not a problem semantically; the type-args are erased at JS emit time. Resolving.

packages/ts-transformers/test/fixtures/jsx-expressions/map-single-capture-no-name.expected.jsx
mathpirate: Tracked this down — it's actually intended optimization that was previously masked. Pre-CT-1615, the printer emitted `import("commonfabric").Cell<...>` for the synthetic input TypeNode, which `isCellLikeTypeNode` (type-shrinking.ts:690) does NOT recognise (it handles only Identifier and QualifiedName, not ImportTypeNode). So the Cell wrapper was silently skipped during shrinking, and the array-intrinsic-only-access optimization at `buildShrunkTypeNodeFromTypeNode:1054-1066` (which emits `unknown[]` when only `.length`-style paths are read) never fired.

My earlier `UseAliasDefinedOutsideCurrentScope` flag fix made the printer emit plain `Cell<...>` (a TypeReferenceNode with Identifier), which `isCellLikeTypeNode` correctly recognises — so the optimization now fires as designed. The 'unknown' isn't a bug, it's the shrink doing its job: when the callback only reads `.length`, dropping the element schema is the intended size optimization. Pre-CT-1615 produced richer items by accident due to the printer-form mismatch.

See the consolidated reply for full investigation trace. Resolving.

packages/ts-transformers/test/fixtures/jsx-expressions/map-single-capture.expected.jsx
mathpirate: Tracked this down — it's actually intended optimization that was previously masked. Pre-CT-1615, the printer emitted `import("commonfabric").Cell<...>` for the synthetic input TypeNode, which `isCellLikeTypeNode` (type-shrinking.ts:690) does NOT recognise (it handles only Identifier and QualifiedName, not ImportTypeNode). So the Cell wrapper was silently skipped during shrinking, and the array-intrinsic-only-access optimization at `buildShrunkTypeNodeFromTypeNode:1054-1066` (which emits `unknown[]` when only `.length`-style paths are read) never fired.

My earlier `UseAliasDefinedOutsideCurrentScope` flag fix made the printer emit plain `Cell<...>` (a TypeReferenceNode with Identifier), which `isCellLikeTypeNode` correctly recognises — so the optimization now fires as designed. The 'unknown' isn't a bug, it's the shrink doing its job: when the callback only reads `.length`, dropping the element schema is the intended size optimization. Pre-CT-1615 produced richer items by accident due to the printer-form mismatch.

See the consolidated reply for full investigation trace. Resolving.

packages/ts-transformers/src/transformers/pattern-body-reactive-root-lowering.ts
```diff
@@ -944,13 +947,24 @@ function rewriteNestedDeriveCallbackBodies(
...
+    // derive(input, cb)) or the inner call (lift-applied shape:
+    // lift(cb)(input)). Detect which by searching both argument lists.
+    const isLiftAppliedShape = ts.isCallExpression(visited.expression);
+    const innerCall = isLiftAppliedShape
+      ? (visited.expression as ts.CallExpression)
+      : undefined;
+    const callbackHostArgs = innerCall
+      ? innerCall.arguments
+      : visited.arguments;
+    const callbackIndex = callbackHostArgs.indexOf(callbackArg);
```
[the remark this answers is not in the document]
mathpirate: Pre-existing behavior (the indexOf pattern dates to PR #3154/#3567, well before CT-1615 — see commit history). Filed as [CT-1619](https://linear.app/common-tools/issue/CT-1619/indexof-on-resolved-callback-in-pattern-body-reactive-root-lowering) since it doesn't fire on current synthesis sites or typical user code. Resolving.

packages/schema-generator/src/schema-generator.ts
```diff
@@ -786,6 +787,32 @@ export class SchemaGenerator implements ISchemaGenerator {
...
+            valueType,
+            context,
+            member.type,
+          );
+          // If multiple index signatures are present (e.g. both string and
+          // number key), the first non-undefined wins — matching
+          // ObjectFormatter's `stringIndex ?? numberIndex` precedence.
+          if (additionalProperties === undefined) {
+            additionalProperties = valueSchema;
+          }
```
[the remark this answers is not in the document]
mathpirate: Real but low-impact. `Record<K, V>` always produces exactly one index signature so the precedence ordering doesn't matter; the divergence only matters for hand-written dual-index TypeLiterals, none of which exist in the codebase. Filed as [CT-1620](https://linear.app/common-tools/issue/CT-1620/schema-generator-typeliteral-index-signature-precedence-is-declaration). Also added direct test coverage in commit `cd6cfd168` (locks in the new IndexSignatureDeclaration branch's behavior). Resolving.

packages/ts-transformers/test/fixtures/jsx-expressions/map-single-capture-no-name.expected.jsx
```diff
@@ -50,30 +50,23 @@ export default pattern((_state: any) => {
                     type: "object",
                     properties: {}
                 }]
-        } as const satisfies __cfHelpers.JSONSchema, __cfHelpers.derive({
+        } as const satisfies __cfHelpers.JSONSchema, __cfHelpers.lift<{
+            people: Cell<{ id: string; name: string; }[]>;
```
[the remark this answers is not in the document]
mathpirate: Addressed in commit `99d7a345e` — synthesized type annotations now use the canonical `__cfHelpers.X` qualified form via the new `qualifyCommonFabricTypeRefs` post-processing pass. Resolving.

PR #3677, opened 2026-05-23
Add shared identity workflow

bfollington:
## Summary
- add a browser login flow for importing CLI PKCS8/PEM key files
- store imported keys in the shell KeyStore and reuse them as a stored credential
- document the shared identity workflow for browser, CLI, FUSE, and agent access

## Testing
- deno check packages/shell/src/views/LoginView.ts packages/shell/src/lib/credentials.ts
- deno test packages/identity/test/identity.test.ts
- deno test --allow-read --allow-write --allow-run packages/shell/test/standard-decorators-shell-views-phase3.test.ts
- deno fmt --check packages/shell/src/views/LoginView.ts packages/shell/src/lib/credentials.ts docs/development/SHARED_IDENTITY.md docs/development/LOCAL_DEV_SERVERS.md docs/common/conventions/HOME_SPACE.md

---
## Summary by cubic
Adds a shared identity workflow so the browser can import CLI PKCS8/PEM keys, store them in `KeyStore`, and reuse them for quick login. This lets the browser, `cf` CLI, FUSE, and `agent-browser` share one DID in local dev; also handles missing imported keyfiles by clearing stale credentials and prompting re-import.

- **New Features**
  - Adds a `keyfile` auth method in `packages/shell` login: import PKCS8/PEM via “Import CLI Key”, persist in `KeyStore`, and save a stored credential for quick unlock.
  - Updates login UI with the import option and clearer labels for passkey, passphrase, and imported keys.
  - Adds `docs/development/SHARED_IDENTITY.md` and links it from `LOCAL_DEV_SERVERS.md` and `HOME_SPACE.md` to guide using one identity across tools.
  - Expands docs in `docs/common/ai/manual-testing-guide.md`, `skills/agent-browser/SKILL.md`, and `skills/cf/SKILL.md` to verify CLI vs browser DID, import the CLI key in `agent-browser` sessions, and clarify `PerUser`/`PerSession` vs `PerSpace` visibility.

- **Bug Fixes**
  - Updates `packages/shell/integration/login.test.ts` to use the new auth method picker (`test-id`s) so the passphrase flow is exercised reliably.

<sup>Written for commit 07eba07479c11e9a7eca6c19bb81e26dac370139. Summary will update on new commits. <a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3677?utm_source=github">Review in cubic</a></sup>


bfollington:
## Code Review

### Overview
Adds a `keyfile` auth method to the shell `LoginView` so a CLI-generated PKCS8/PEM key can be imported into the browser `KeyStore` and reused as a stored credential. Adds `docs/development/SHARED_IDENTITY.md` plus references from `LOCAL_DEV_SERVERS.md`, `HOME_SPACE.md`, the manual-testing guide, and the `agent-browser`/`cf` skills.

### What works well
- Reuses the existing `Identity.fromPkcs8` — no new crypto path.
- New auth method is wired symmetrically through `credentials.ts`: constant, type union, factory function.
- Quick-unlock label is data-driven via the `storedLabel` ternary; no string drift between buttons.
- `handleMethodSelect` short-circuits for KEYFILE so register/login fallthroughs don't fire spurious passphrase events.
- `handleRegister` is tightened from `else` to `else if (PASSPHRASE)`; previously a KEYFILE method could have dispatched a stray `passphrase-generate`.
- Docs are concrete: include the exact `[Identity] User DID:` log line that `packages/shell/src/lib/runtime.ts:360` actually emits, and the `deno eval` snippet is correct (DID derives from entropy bytes, so implementation choice doesn't change the resulting DID).

### Issues & risks

**Correctness**

- `LoginView.ts:571-576` — `handleQuickUnlock` for KEYFILE: if `keyStore.get(ROOT_KEY)` returns `undefined` (cleared/missing key), it silently does nothing. After the click, `flow=\"login\"` and `method=KEYFILE` are now set, so the view re-renders into `renderKeyFileImport()` with no error explaining why. Set `this.error = \"Stored key file is no longer available — please re-import.\"` (and clear the stale credential) when the lookup returns nothing.
- Same handler is not wrapped in try/catch. `getKeyStore()` throws if `this.keyStore` is undefined (`LoginView.ts:241-246`); the async error would become an unhandled rejection rather than appearing in `this.error`. All sibling handlers (`handlePasskey*`, `handlePassphrase*`, `handleKeyFileImport`) use try/catch with `finally { isProcessing = false }`. Make this one consistent.
- `LoginView.ts:563` — `handleQuickUnlock` is now `async` but `renderInitial` calls it via `@click=\"\${() => this.handleQuickUnlock()}\"`. Lit won't await — fine for behavior, but `isProcessing` is never toggled, so a slow `keyStore.get` shows no spinner and the user can double-click. Consider routing through the same `isProcessing` pattern.

**UX / polish**

- `LoginView.ts:493` — `Imported Key (\${this.storedCredential.id.slice(-8)})` slices the last 8 chars of a DID. A `did:key:z6Mk…` always ends in a base58 suffix where the trailing chars vary — fine for distinguishing, but worth noting it's not a stable fingerprint.
- `LoginView.ts:531` — pre-existing stray `\"` in `\" 🔑 Login w/ Passkey` (not introduced here, but visible next to the new branch — worth fixing while in the area).

**Security**

- Importing a raw PKCS8 private key into IndexedDB persists key material in the browser at-rest with no passphrase wrap. This is by design and matches the passphrase/passkey flows, but the import form has no warning. At minimum the `info-text` paragraph could say \"this stores the private key in this browser's IndexedDB\" — currently it only says \"Select a PKCS8/PEM key generated by the cf CLI.\"
- `docs/development/SHARED_IDENTITY.md:24` — `mkdir -p .cf` + `*.key`/`.cf/` gitignore claim. Verify `.gitignore` actually ignores `.cf/` — I didn't see it added in the diff. If not present, this advice will leak keys into commits for anyone who follows the doc literally.

**Testing**

- No unit/integration test added for the new code path. Given the silent-failure bug above, a small test would catch it. Suggested: round-trip write PKCS8 → import → assert `KeyStore.get(ROOT_KEY)` matches and `getStoredCredential()` reports `method === \"keyfile\"`.

**Minor**

- `credentials.ts` uses the string literal `\"keyfile\"` inside `createKeyFileCredential` instead of the exported constant `AUTH_METHOD_KEYFILE` — same pattern as the existing passkey/passphrase factories, so consistent, but the constants exist to prevent exactly this kind of drift.
- `renderMethodSelection` ternary chain ends in a catch-all that labels anything non-passkey/non-passphrase as \"Import CLI Key\". Switch to exhaustive `===` for forward-compat if a fourth method is added.

### Suggested follow-up (non-blocking)
- Wire the keyfile quick-unlock through `isProcessing` + try/catch.
- Verify `.cf/` is in `.gitignore` (or amend the docs to mention placing keys outside the repo).
- Add one test that round-trips PKCS8 import + stored-credential state.

Overall: small, well-scoped feature that closes a real dev-loop pain point, and the docs are unusually thorough. The silent-failure path in `handleQuickUnlock` is the only thing I'd want addressed before merge.

PR #3688, opened 2026-05-26
Turn the `modern-data-model` flag off by default.

danfuzz:
This is a revert of PR #3569 and is ready and waiting as a draft, in case the revert actually needs to happen.


danfuzz:
We've officially graduated the flag! :tada:

PR #3711, opened 2026-05-27
[codex] Add cf-harness SQLite chat session store

mathpirate:
## Summary

This adds the first durable cf-harness chat/session persistence slice, using SQLite as the near-term backing store while keeping future Fabric-backed storage behind a storage boundary.

- Add a generic `HarnessChatSessionStore` interface for session snapshots, transcripts, ordered events, replay queries, and latest sequence reads.
- Add an explicit `./sqlite-session-store` implementation backed by `@db/sqlite`, with `chat_session` and `chat_event` tables.
- Add a `list_events` interactive-chat protocol method for replay since `afterSequence`, with optional session filtering and limits.
- Wire `HarnessInteractiveChatService` to persist session snapshots and event envelopes, restore state via `initializeFromStore()`, continue sequence numbers after restart, and reject duplicate durable session IDs.
- Keep the native SQLite implementation on an explicit subpath so root cf-harness imports do not automatically load SQLite.

## Notes

This intentionally does not migrate existing run artifacts yet. Run state, run reports, policy traces, tool outputs, and related JSON artifacts remain file-backed for this PR. The next slice should add explicit turn records / interrupted-turn terminalization and then move artifact persistence behind the same SQLite-backed storage direction.

## Validation

- `deno fmt` on touched files
- `deno lint packages/cf-harness/src/session-store.ts packages/cf-harness/src/sqlite-session-store.ts packages/cf-harness/src/contracts/interactive-chat.ts packages/cf-harness/src/interactive-chat-service.ts packages/cf-harness/src/interactive-chat-stdio.ts packages/cf-harness/src/index.ts packages/cf-harness/test/interactive-chat-service.test.ts packages/cf-harness/test/interactive-chat-stdio.test.ts packages/cf-harness/test/sqlite-session-store.test.ts`
- `deno check packages/cf-harness/src/session-store.ts packages/cf-harness/src/sqlite-session-store.ts packages/cf-harness/src/contracts/interactive-chat.ts packages/cf-harness/src/interactive-chat-service.ts packages/cf-harness/src/interactive-chat-stdio.ts packages/cf-harness/test/interactive-chat-service.test.ts packages/cf-harness/test/interactive-chat-stdio.test.ts packages/cf-harness/test/sqlite-session-store.test.ts`
[...]

PR #3714, opened 2026-05-27
[codex] Add cf-harness durable turn records

mathpirate:
## Summary

Stacked on #3711 (`codex/cf-harness-sqlite-sessions`). This adds the next cf-harness persistence slice: durable chat turn records and restart recovery on top of the SQLite session/event store.

Changes:
- add `HarnessChatTurnRecord` plus `list_turns` request/result types to the interactive chat protocol
- add a SQLite `chat_turn` table with turn/input/context/policy/browser lease/metadata persistence
- persist turn lifecycle transitions for running, canceling, canceled, completed, and failed turns
- add `turn_failed` as a per-turn terminal event so prompt-loop failures do not poison the reusable session
- terminalize restored interrupted turns during `initializeFromStore()` (`running` -> `failed` with `process_interrupted`; `canceling` -> `canceled`)
- add NDJSON transport validation for `list_turns`

## Why

The prior PR made sessions and event replay durable. This fills in explicit turn-level state so Loom/headless clients can inspect turn status directly and so a cf-harness process restart does not leave a session permanently stuck on an in-flight turn.

## Validation

- `deno fmt packages/cf-harness/src/session-store.ts packages/cf-harness/src/sqlite-session-store.ts packages/cf-harness/src/contracts/interactive-chat.ts packages/cf-harness/src/interactive-chat-service.ts packages/cf-harness/src/interactive-chat-stdio.ts packages/cf-harness/test/interactive-chat-service.test.ts packages/cf-harness/test/interactive-chat-stdio.test.ts packages/cf-harness/test/sqlite-session-store.test.ts`
- `deno check packages/cf-harness/src/session-store.ts packages/cf-harness/src/sqlite-session-store.ts packages/cf-harness/src/contracts/interactive-chat.ts packages/cf-harness/src/interactive-chat-service.ts packages/cf-harness/src/interactive-chat-stdio.ts packages/cf-harness/test/interactive-chat-service.test.ts packages/cf-harness/test/interactive-chat-stdio.test.ts packages/cf-harness/test/sqlite-session-store.test.ts`
- `deno lint packages/cf-harness/src/session-store.ts packages/cf-harness/src/sqlite-session-store.ts packages/cf-harness/src/contracts/interactive-chat.ts packages/cf-harness/src/interactive-chat-service.ts packages/cf-harness/src/interactive-chat-stdio.ts packages/cf-harness/test/interactive-chat-service.test.ts packages/cf-harness/test/interactive-chat-stdio.test.ts packages/cf-harness/test/sqlite-session-store.test.ts`
- `deno test --allow-read --allow-write --allow-run --allow-ffi --allow-net --allow-env packages/cf-harness/test/interactive-chat-service.test.ts packages/cf-harness/test/interactive-chat-stdio.test.ts packages/cf-harness/test/sqlite-session-store.test.ts` (`24 passed`)
- `deno task test` in `packages/cf-harness` (`356 passed`)
- `git diff --check`

---
## Summary by cubic
Adds durable per-turn records with a restart-safe lifecycle to the `cf-harness` chat service. Persists turns in SQLite, adds `list_turns`, de-dupes `start_turn` by `turnId`, and emits clear terminal events so clients can query history and recover after restarts.

- **New Features**
  - Protocol: `HarnessChatTurnRecord`, `list_turns` request/result (with status filter), `turn_failed` event, `turn_exists` error; `error` added to `HarnessChatTurnStatus`.
  - Persistence: SQLite `chat_turn` table (turn/input/context/policy/browser lease/metadata/status/timestamps/error); atomic `saveSessionTurnAndAppendEvent` to persist session+turn+event; durable create with insert-or-ignore; file: URLs only.
  - Service/Transport: per-session in-memory turn map; filterable turn listing; durable de-dupe of `start_turn` with a reserved starting turn to prevent races; emits `turn_failed` on prompt-loop setup/runtime errors; NDJSON validation for `list_turns`.

- **Bug Fixes**
  - Restart recovery: on init, terminalize interrupted turns (running -> failed with `process_interrupted`; canceling -> canceled), clear stale `activeTurn`, keep closed/failed sessions non-reusable, and clear session `activeTurn` if the stored turn is already terminal.
  - Safety: rolls back in-memory turn state if persistence fails; serializes event sequences; closing a session cancels, emits, and persists the active turn.

<sup>Written for commit ab87e62982c95ec8574eee7911233ac2bee2a302. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3714?utm_source=github">Review in cubic</a>

### DIFF
--- a/packages/cf-harness/src/contracts/interactive-chat.ts
+++ b/packages/cf-harness/src/contracts/interactive-chat.ts
@@ -28,7 +28,8 @@ export type HarnessChatRequestMethod =
   | "cancel_turn"
   | "close_session"
   | "status"
-  | "list_events";
+  | "list_events"
+  | "list_turns";
 
 export type HarnessChatSessionLifecycle =
   | "idle"
@@ -204,6 +205,11 @@ export interface HarnessChatListEventsParams {
   limit?: number;
 }
 
+export interface HarnessChatListTurnsParams {
+  sessionId?: string;
+  status?: HarnessChatTurnLifecycle;
+}
+
 export type HarnessChatRequestParamsByMethod = {
   start_session: HarnessChatStartSessionParams;
   start_turn: HarnessChatStartTurnParams;
@@ -211,6 +217,7 @@ export type HarnessChatRequestParamsByMethod = {
   close_session: HarnessChatCloseSessionParams;
   status: HarnessChatStatusParams;
   list_events: HarnessChatListEventsParams;
+  list_turns: HarnessChatListTurnsParams;
 };
 
 export type HarnessChatRequestEnvelope<
@@ -230,6 +237,7 @@ export interface HarnessChatError {
     | "invalid_request"
     | "session_exists"
     | "session_not_found"
+    | "turn_exists"
     | "turn_not_found"
     | "turn_already_running"
     | "turn_canceled"
@@ -269,6 +277,7 @@ export interface HarnessChatTurnStatus {
   updatedAt: string;
   endedAt?: string;
   cancelReason?: string;
+  error?: HarnessChatError;
 }
 
 export interface HarnessChatSessionStatus {
@@ -294,6 +303,16 @@ export interface HarnessChatSessionStatus {
 
 export interface HarnessChatStatusResult {
   sessions: readonly HarnessChatSessionStatus[];
+}
+
+export interface HarnessChatTurnRecord {
+  sessionId: string;
+  turn: HarnessChatTurnStatus;
+  input: HarnessChatTurnInput;
+  context?: HarnessChatContext;
+  policy: HarnessChatPolicy;
+  browserAccess?: HarnessChatBrowserAccessLease;
+  metadata?: Record<string, unknown>;
 }
 
 export interface HarnessChatToolCallSummary {
@@ -389,6 +408,11 @@ export type HarnessChatStructuredEvent =
     usage?: HarnessChatGatewayUsage;
   }
   | {
+    kind: "turn_failed";
+    turnId: string;
+    error: HarnessChatError;
+  }
+  | {
     kind: "status_changed";
     session: HarnessChatSessionStatus;
   }
@@ -417,6 +441,10 @@ export interface HarnessChatEventEnvelope<
 export interface HarnessChatListEventsResult {
   events: readonly HarnessChatEventEnvelope[];
   latestSequence: number;
+}
+
+export interface HarnessChatListTurnsResult {
+  turns: readonly HarnessChatTurnRecord[];
 }
 
 export interface CreateHarnessChatSessionStatusOptions {
@@ -507,6 +535,14 @@ export const createHarnessChatErrorResponse = (
   error,
 });
 
+const clearSessionActiveTurn = (
+  status: HarnessChatSessionStatus,
+): Omit<HarnessChatSessionStatus, "activeTurn" | "activeTurnId"> => {
+  const { activeTurn: _activeTurn, activeTurnId: _activeTurnId, ...rest } =
+    status;
+  return rest;
+};
+
 export const reduceHarnessChatSessionStatus = (
   status: HarnessChatSessionStatus,
   envelope: HarnessChatEventEnvelope,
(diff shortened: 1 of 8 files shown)


willkelly (changes requested):
Thanks for pushing the next cf-harness persistence slice. The happy path is in good shape and local validation passes, but I think this needs changes before merge because a few lifecycle paths can leave durable turn/session/event state inconsistent or a session permanently blocked.

Validated locally:
- LSP diagnostics clean on touched files
- deno check passed for changed files/tests
- deno fmt --check passed
- deno lint passed
- focused tests passed 27/27
- full cf-harness test suite passed 359/359
- git diff --check passed

Blocking issues:

1. Prompt-loop setup failures can leave the session permanently stuck. `#createPromptLoop(...)` is called before the `try` in `#runTurn`, so if the factory/setup path throws, no `turn_failed` event is emitted. I confirmed with a local repro that `start_turn` returns ok, the event stream contains only `session_started`/`turn_started`, the turn stays `running`, the session stays `turn_running` with `activeTurnId`, `activeTask` is cleared, and the next `start_turn` is rejected as `turn_already_running`. This directly contradicts the goal that prompt-loop failures do not poison the reusable session.

2. Turn updates are not atomic with session snapshots and events. `#emitImmediately` updates `chat_turn` through `#applyEventToTurn`/`#setTurnStatus` before `saveSessionAndAppendEvent` persists the session snapshot and event. If the latter fails, durable `chat_turn` can say `completed`/`failed`/`canceling` while `chat_event` is missing the transition and `chat_session` still reflects the old state. Similarly, `startTurn` writes the initial running turn before emitting `turn_started`, so a crash/error between those writes leaves a running turn with no replayable start event and an idle session snapshot.

3. `startTurn` has a race before active turn reservation. The active-turn and duplicate-turn checks happen before the awaited durable `getTurn` call and before the in-memory turn/active state is reserved. Concurrent `start_turn` calls can both pass those checks; with the same explicit `turnId`, the `ON CONFLICT DO UPDATE` path can overwrite instead of rejecting, and with different turn IDs two prompt loops can overlap.

4. `closeSession` makes the active turn durable-canceled without a corresponding turn event. It calls `#setTurnStatus(... canceled ...)` and then emits only `session_closed`. That leaves `list_turns` saying the turn was canceled, but `list_events` cannot explain that terminal transition. If the event stream is the replay/audit surface, the turn terminal transition should be represented consistently.

5. Restart terminalization should preserve terminal session state. The reducer currently handles `turn_failed` by making the session `idle`/`reusable` unconditionally. If recovery finds a nonterminal turn under a closed/failed session, terminalizing that turn can reopen the session. Closed/failed session state should dominate turn recovery.

Suggested direction: make lifecycle transitions a single store-level transaction that writes the turn mutation, session snapshot, and event together. Use an insert-only path for new turns so duplicate explicit turn IDs fail, and an update path for lifecycle transitions, ideally with expected previous-status checks (`running -> completed/failed`, `running|canceling -> canceled`, etc.). Only update in-memory status after the durable transaction succeeds.

No objection to the shape of `HarnessChatTurnRecord`, `list_turns`, or keeping existing file-backed artifacts out of scope for this PR; the requested changes are about making the new turn/session/event durability model internally consistent.

willkelly (approved):
Thanks for the updates — approving this now. I re-reviewed head `5f442793a271c78201afd2bc6c1241e989ce839c`, re-exercised the lifecycle paths, and the previously blocking durability/lifecycle issues are addressed.

What changed my assessment:
- Prompt-loop setup/run failures now terminalize the turn with `turn_failed`, so the session does not remain stuck in `turn_running` and follow-up turns can start.
- Session + turn + event persistence now goes through atomic mutation paths, so durable state no longer splits across `chat_session`, `chat_turn`, and `chat_event`.
- Concurrent `start_turn` calls are guarded by the pre-persistence reservation path and reject as `turn_already_running`.
- Closing a session with an active turn now emits a replayable `turn_canceled` before `session_closed`.
- Restart recovery terminalizes interrupted non-terminal turns instead of reopening stale active state.

Validation I ran/checked:
- LSP diagnostics clean on the touched files.
- `deno check`, `deno fmt --check`, and `deno lint` passed on the changed cf-harness files/tests.
- Focused tests passed: `32/32`.
- Full cf-harness test suite passed: `364/364`.
- `GIT_MASTER=1 git diff --check` passed, and the worktree was clean.

Comparison to similar systems:

| System | Comparison | Takeaway for this PR |
| --- | --- | --- |
| opencode | Persists sessions/messages/parts in SQLite; a turn is inferred from user/assistant message linkage rather than stored as a first-class durable run/turn. | This PR is stronger on explicit turn lifecycle by adding `HarnessChatTurnRecord`, `list_turns`, durable terminal states, and restart terminalization. |
| OpenAI Assistants / Responses | Uses explicit thread/run/message concepts with list/retrieve/cancel APIs and terminal run statuses. | The PR aligns with that model, adapted for local cf-harness persistence with atomic session+turn+event writes. |
| LangGraph | Durable threads/runs/checkpoints with run status/history and cancellation semantics. | This is the closest conceptual match; the PR is narrower but follows the same durable-run direction. |
| Vercel AI SDK / Claude Agent SDK | Emphasize message/session persistence and stream/session resume, but expose weaker built-in run/turn lifecycle state. | The PR goes beyond transcript-only persistence by making turn lifecycle and replayable events explicit. |

Non-blocking follow-ups:
- Unsupported/non-`file:` SQLite URL schemes currently appear to fall back to `:memory:`; it would be clearer to reject unsupported schemes or make the fallback explicit.
- Durable stdio wiring/default discoverability is still somewhat implicit; a docs/top-level entrypoint note would make file-backed usage easier to find.

Neither follow-up blocks this PR. The core durable turn/session/event model is now consistent and mergeable.

PR #3730, opened 2026-05-28
[codex] Wire cf-harness chat stdio to SQLite sessions

mathpirate:
## Summary
- add opt-in SQLite persistence to the cf-harness interactive chat stdio transport via --chat-session-db / CF_HARNESS_CHAT_SESSION_DB
- restore persisted chat sessions, turns, and replayable events when the stdio process starts
- close the SQLite store on shutdown/init failure and document the durable stdio entrypoint
- add in-process and child-process CLI coverage for persistence across process invocations

## Validation
- deno fmt packages/cf-harness/README.md packages/cf-harness/src/interactive-chat-stdio.ts packages/cf-harness/test/interactive-chat-stdio.test.ts
- deno check packages/cf-harness/src/interactive-chat-stdio.ts packages/cf-harness/test/interactive-chat-stdio.test.ts
- deno lint packages/cf-harness/src/interactive-chat-stdio.ts packages/cf-harness/test/interactive-chat-stdio.test.ts
- deno test --allow-read --allow-write --allow-run --allow-ffi --allow-net --allow-env packages/cf-harness/test/interactive-chat-stdio.test.ts: 14 passed
- deno test --allow-read --allow-write --allow-run --allow-ffi --allow-net --allow-env packages/cf-harness/test/interactive-chat-service.test.ts packages/cf-harness/test/interactive-chat-stdio.test.ts packages/cf-harness/test/sqlite-session-store.test.ts: 36 passed
- deno task test in packages/cf-harness: 368 passed

---
## Summary by cubic
Adds opt-in SQLite persistence to the interactive chat NDJSON stdio transport so sessions, turns, and replayable events survive restarts. Introduces a CLI for DB path selection with safe startup/teardown and stronger transport lifecycle handling.

- **New Features**
  - Persist via `--chat-session-db <path>` or `CF_HARNESS_CHAT_SESSION_DB`; open from path and safely close the SQLite store.
  - Restore sessions and events on process start; `runHarnessInteractiveChatStdioCli` supports `--help`, prints usage to stderr, validates args, and exits 1 on startup errors.
  - NDJSON transport accepts async `createService`, waits for idle, and calls optional `closeService` on normal exit and after write failures.
  - `runHarnessInteractiveChatStdio` supports `sessionDbPath` and injectable `createPromptLoop`, initializes from store, and closes it on shutdown/init failure.
  - README documents stdio usage; tests cover in-process and child-process persistence, write-failure cleanup, CLI parsing, and error reporting.

<sup>Written for commit 3e35a7dc8c6026977d161bb48b0b37619529444f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3730?utm_source=github">Review in cubic</a>

### DIFF
--- a/packages/cf-harness/README.md
+++ b/packages/cf-harness/README.md
@@ -53,6 +53,8 @@ What works today:
   retention in tool-output artifacts; `web_fetch` is intentionally not part of
   the default parent tool surface
 - bounded OpenAI-compatible prompt/tool loop
+- interactive chat NDJSON stdio transport with opt-in SQLite session, turn, and
+  event persistence
 - single-child subagent delegation with fresh child prompt context, explicit
   default/browser/web_fetch/web_search child profiles, retained child run
   references, and a sanitized summary/state return channel
@@ -117,6 +119,10 @@ What is not done yet:
   - bounded prompt/tool loop
 - [src/cli.ts](src/cli.ts)
   - package-local operator CLI
+- [src/interactive-chat-stdio.ts](src/interactive-chat-stdio.ts)
+  - NDJSON stdio transport for the interactive chat protocol
+- [src/sqlite-session-store.ts](src/sqlite-session-store.ts)
+  - SQLite-backed interactive chat session, turn, and event persistence
 - [src/artifacts.ts](src/artifacts.ts)
   - persisted run state, run manifest, transcript, run report, capability
     snapshot, and tool output storage
@@ -161,6 +167,19 @@ deno task run -- \
   --prompt "Summarize the cf-harness package structure." \
   --print-transcript
 ```
+
+Interactive chat stdio transport:
+
+```bash
+cd packages/cf-harness
+deno run -A src/interactive-chat-stdio.ts \
+  --chat-session-db /tmp/cf-harness-chat.sqlite
+```
+
+The stdio transport reads one interactive chat request envelope per line from
+stdin and writes response/event envelopes as newline-delimited JSON. Pass
+`--chat-session-db` or set `CF_HARNESS_CHAT_SESSION_DB` to persist sessions,
+turn records, and replayable events across process restarts.
 
 Initial prompt image attachments:
 
--- a/packages/cf-harness/src/interactive-chat-stdio.ts
+++ b/packages/cf-harness/src/interactive-chat-stdio.ts
@@ -1,3 +1,4 @@
+import { resolve, toFileUrl } from "@std/path";
 import {
   createHarnessChatErrorResponse,
   HARNESS_CHAT_PROTOCOL_VERSION,
@@ -18,7 +19,9 @@ import type { BuiltinToolId } from "./contracts/tool-descriptor.ts";
 import {
   createHarnessInteractiveChatService,
   type HarnessInteractiveChatService,
+  type HarnessInteractivePromptLoopFactory,
 } from "./interactive-chat-service.ts";
+import type { HarnessChatSessionStore } from "./session-store.ts";
 
 export type HarnessInteractiveChatOutputEnvelope =
   | HarnessChatEventEnvelope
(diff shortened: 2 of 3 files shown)


willkelly (changes requested):
Thanks for wiring this up — the normal durable stdio path looks good, and the focused validation passed locally. I do think one shutdown-ordering issue needs to be fixed before merge because it can undermine the new persistence guarantee on realistic stdio failures.

Blocking issue: `runHarnessInteractiveChatNdjsonTransport` only waits for active turns after the input loop exits normally. If input iteration, request/response writing, or another outer transport operation throws, execution jumps to `finally` and calls `closeService` immediately. In SQLite stdio mode that closes the session store while a `start_turn` task may still be running and later trying to persist terminal events.

Small repro I used:

```ts
let closeBeforeLoopFinished = false;
let loopFinished = false;
let resolveLoop: (() => void) | undefined;

const service = new HarnessInteractiveChatService({
  onEvent: async () => {},
  createPromptLoop: () => ({
    runTranscript: async () => {
      await new Promise<void>((resolve) => {
        resolveLoop = resolve;
      });
      loopFinished = true;
      return {
        model: "gpt-test",
        finalAssistantText: "done",
        transcript: [{ role: "assistant", content: "done" }],
        modelTurns: 1,
        runState: {} as never,
      };
    },
  }),
});

await runHarnessInteractiveChatNdjsonTransport({
  lines: [startSessionJson, startTurnJson],
  createService: () => service,
  closeService: async () => {
    closeBeforeLoopFinished = !loopFinished;
    resolveLoop?.();
  },
  writeLine: async (line) => {
    const envelope = JSON.parse(line);
    if (envelope.requestId === "req-start-turn") {
      throw new Error("simulated stdout failure");
    }
  },
}).catch(() => undefined);

await service.waitForIdle();
assertEquals(closeBeforeLoopFinished, false);
```

On the current PR head, that assertion fails: `closeService` runs before the prompt loop settles. With a real SQLite-backed service, that means a broken pipe/client disconnect can close the DB before `turn_completed` / `turn_failed` is persisted.

Suggested fix: move the idle drain into the shutdown path so it always runs before `closeService`, while still preserving the original transport error after cleanup. Something like:

```ts
let thrown: unknown;
try {
  for await (const rawLine of options.lines) {
    // existing request loop
  }
} catch (error) {
  thrown = error;
} finally {
  try {
    await resolvedService.waitForIdle();
  } finally {
    await options.closeService?.(resolvedService);
  }
}
if (thrown !== undefined) {
  throw thrown;
}
```

The exact shape can differ, but the important invariant is: active turn tasks are either waited/canceled before closing the persistence resource, including exceptional transport exits. Please also add a focused regression test for a `writeLine` failure while a turn is active.

Validation I ran before this review:
- LSP diagnostics clean on `interactive-chat-stdio.ts` and its test.
- `deno check packages/cf-harness/src/interactive-chat-stdio.ts packages/cf-harness/test/interactive-chat-stdio.test.ts` passed.
- `deno fmt --check` on the three changed files passed.
- `deno lint` on the source/test passed.
- `deno test ... interactive-chat-stdio.test.ts`: 14/14 passed.
- Focused service + stdio + SQLite persistence suites: 36/36 passed.
- `GIT_MASTER=1 git diff --check origin/main...HEAD` passed, worktree clean.

Non-blocking follow-up coverage nits:
- It would be useful to add an invalid/unwritable DB path startup test.
- It would be useful to explicitly assert the close hook runs on normal completion/init-failure paths.

Those nits are optional; the shutdown ordering issue above is the blocker.

willkelly (approved):
Thanks for the update — approving this now. I re-reviewed head `3e35a7dc8c6026977d161bb48b0b37619529444f`, and the shutdown-ordering blocker from my previous review is fixed.

What changed my assessment:
- `runHarnessInteractiveChatNdjsonTransport` now records transport/cleanup errors, always waits for `resolvedService.waitForIdle()` in `finally`, then calls `closeService`.
- That preserves the important invariant for durable stdio mode: active turn tasks finish before the SQLite store is closed, including abnormal transport exits.
- The original transport error is still rethrown first, so callers do not lose the root failure.
- The new regression test covers the exact write-failure path we discussed: `writeLine` fails while a turn is active, and `closeService` does not run before the prompt loop settles.

Validation I ran/checked:
- LSP diagnostics clean on `interactive-chat-stdio.ts` and its test.
- `deno check packages/cf-harness/src/interactive-chat-stdio.ts packages/cf-harness/test/interactive-chat-stdio.test.ts` passed.
- `deno fmt --check` on README/source/test passed.
- `deno lint` on source/test passed.
- Updated stdio tests passed: `17/17`.
- Focused service + stdio + SQLite persistence suites passed: `39/39`.
- Direct repro now passes with `closeBeforeLoopFinished=false` while still propagating the simulated stdout error.
- `GIT_MASTER=1 git diff --check origin/main...HEAD` passed, and the worktree was clean.

Only optional follow-up: a tiny test around `closeService` failure/error-priority behavior would be nice, but it is not blocking. The durable stdio persistence path is now consistent and mergeable.

PR #3737, opened 2026-05-28
[codex] Enable browser subagent skills and scripts

mathpirate:
## Summary
- Activate the `agent-browser` skill inside browser-profile subagents when a parent skill registry is available.
- Allow browser subagents to use `read_skill_resource` and exact allowlisted `agent-browser` skill scripts.
- Add host-target skill script execution for browser-profile scripts so they can call the host `agent-browser` CLI against a provided CDP endpoint.

## Validation
- `deno fmt --check packages/cf-harness/src/config.ts packages/cf-harness/src/contracts/cfc-policy-snapshot.ts packages/cf-harness/src/contracts/skill.ts packages/cf-harness/src/contracts/subagent.ts packages/cf-harness/src/engine.ts packages/cf-harness/src/prompt-loop.ts packages/cf-harness/src/sandbox/process-runner.ts packages/cf-harness/src/tools/run-skill-script.ts packages/cf-harness/src/tools/types.ts packages/cf-harness/test/prompt-loop.test.ts packages/cf-harness/test/tools-execution.test.ts`
- `deno check packages/cf-harness/src/main.ts`
- `deno test --allow-read --allow-write --allow-run --allow-ffi --allow-net --allow-env packages/cf-harness/test/tools-execution.test.ts --filter run_skill_script`
- `deno test --allow-read --allow-write --allow-run --allow-ffi --allow-net --allow-env packages/cf-harness/test/prompt-loop.test.ts --filter "subagent"`
- `cd packages/cf-harness && deno task test` (373 passed)
- `git diff --check`

---
## Summary by cubic
Enables browser subagents to activate `agent-browser`, use `read_skill_resource` and `run_skill_script`, and run exact allowlisted scripts on the host bound to a Browser Access CDP lease. Tightens host policy, binds page commands to the lease, annotates `executionTarget`, and runs host processes with a cleared env for safer automation.

- New Features
  - Browser profile activates `agent-browser` when a parent skill registry exists; exposes `read_skill_resource` and `run_skill_script`.
  - Exact allowlist now: `agent-browser:scripts/form-automation.sh`, `agent-browser:scripts/capture-workflow.sh` (note: `authenticated-session.sh` is not allowlisted).
  - Host-target execution for browser-profile scripts: scripts must pass `--cdp` that matches the leased local endpoint; outputs include `executionTarget`; provenance recorded in `skill-script-executions.json`.
  - Adds `skillScriptExecutionTarget` config (default `sandbox`); browser profile sets it to `host`. Child engine receives `skillsRoot`, allowed scripts, execution target, and `browserAccess`; parent skill registry is persisted to the child, context is loaded, and activations are recorded.
  - Host execution rules: must run from a workspace cwd outside cf-harness artifacts; processes run with a cleared env, controlled `PATH`, and explicit `CF_HARNESS_*`/`SKILL_*` (including `CF_HARNESS_SKILL_SCRIPT_EXECUTION_TARGET=host`); `SKILL_DIR`/`SKILL_SCRIPT` use host paths.
  - Stronger host policy for `agent-browser`: only allow `open` (http/https), `snapshot` (optional `-i`), `get title/url/text`, bounded `wait`, and ref-based `fill`/`type`/`select`/`check`/`click`/`press`; only the `--cdp` global flag is accepted; page commands require and must match the leased CDP; reject `eval`, file: URLs, and non-allowlisted flags.
  - Sandbox mediation only for sandbox-target `run_skill_script`; host-target outputs remain in child artifacts; the parent sees the child’s sanitized structured return.

- Refactors
  - `bash-no-sandbox` validates commands against the provided Browser Access lease and uses cleared-env host processes.
  - `run_skill_script` supports host execution, enforces `--cdp` lease match for `agent-browser` scripts, and annotates outputs with `executionTarget`.
  - `process-runner` adds `env`/`clearEnv` support.
  - CFC policy snapshot/manifest carry child `skillNames`, `allowedSkillScripts`, and `skillScriptExecutionTarget`.
  - Removed an unused browser policy helper.

<sup>Written for commit b272a5ffb59504c402f6210ed215ce576a113d1f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3737?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---BEGIN COPY-PASTE---
NEW_PERF_BASELINE: job: CLI Integration Tests (core-piece-call) = 128s
---END COPY-PASTE---

### DIFF
--- a/packages/cf-harness/README.md
+++ b/packages/cf-harness/README.md
@@ -43,6 +43,7 @@ What works today:
   - `view_image`
   - `web_fetch` (explicit parent allowlist or `web_fetch` subagent profile only)
   - `read_skill_resource`
+  - `run_skill_script`
   - `edit_file`
   - `write_file`
   - `delegate_task`
@@ -73,6 +74,8 @@ What works today:
 - runtime-generated supporting-resource indexes in `skill-registry.json`
 - text-first supporting-resource reads through `read_skill_resource`, recorded
   in `skill-resource-reads.json`
+- exact-allowlisted skill script execution through `run_skill_script`, recorded
+  in `skill-script-executions.json`
 
 The sandbox `bash` tool has a provisional direct-`curl` guard while sandbox
 networking is enabled: explicit `curl` invocations may target loopback HTTP(S)
@@ -103,7 +106,6 @@ What is not done yet:
 - first-class browser operation policy on top of the provisional browser
   subagent profile
 - dynamic/model-driven Agent Skills activation
-- skill script execution
 - parallel child orchestration
 - app UI event provenance
 - streaming model responses
@@ -278,11 +280,24 @@ with a `returnSchema`, parent-visible free-form strings are replaced by opaque
 links while raw observations stay in child artifacts. The browser child can read
 workspace files but does not receive `edit_file` or `write_file`, so it should
 return findings through the structured return channel rather than by writing
-browser observations into the workspace. The host shell is policy-restricted to
-`agent-browser` attached through an approved local CDP endpoint, `agent-browser`
-discovery (`which agent-browser`, `command -v agent-browser`), `pwd`, `ls`, and
-bounded workspace-local `find` commands. Page commands should use the Loom
-Browser Access endpoint supplied by the task, for example
+browser observations into the workspace.
+
+When the parent run has a skill registry, the browser profile activates the
+`agent-browser` skill in the child run, exposes `read_skill_resource`, and
+allows these exact skill scripts through `run_skill_script`:
+`agent-browser:scripts/form-automation.sh`,
+`agent-browser:scripts/authenticated-session.sh`, and
+`agent-browser:scripts/capture-workflow.sh`. Those browser-profile scripts run
+through host execution because they need the host `agent-browser` CLI. They
+still use the normal skill-script safeguards: activated skill, run-start
+registry snapshot, exact script allowlist, digest/size match, and provenance
+artifacts.
+
+The host shell is policy-restricted to `agent-browser` attached through an
+approved local CDP endpoint, `agent-browser` discovery (`which agent-browser`,
+`command -v agent-browser`), `pwd`, `ls`, and bounded workspace-local `find`
+commands. Page commands should use the Loom Browser Access endpoint supplied by
+the task, for example
 `agent-browser --cdp http://host.docker.internal:9362 snapshot -i`. Bare
 `agent-browser open` / `snapshot` launches are denied so the child cannot race
 the host's live browser profile. `agent-browser eval` is not available in this
--- a/packages/cf-harness/README.md
+++ b/packages/cf-harness/README.md
@@ -286,23 +286,31 @@ When the parent run has a skill registry, the browser profile activates the
 `agent-browser` skill in the child run, exposes `read_skill_resource`, and
 allows these exact skill scripts through `run_skill_script`:
 `agent-browser:scripts/form-automation.sh`,
-`agent-browser:scripts/authenticated-session.sh`, and
 `agent-browser:scripts/capture-workflow.sh`. Those browser-profile scripts run
 through host execution because they need the host `agent-browser` CLI. They
 still use the normal skill-script safeguards: activated skill, run-start
 registry snapshot, exact script allowlist, digest/size match, and provenance
 artifacts.
 
-The host shell is policy-restricted to `agent-browser` attached through an
-approved local CDP endpoint, `agent-browser` discovery (`which agent-browser`,
-`command -v agent-browser`), `pwd`, `ls`, and bounded workspace-local `find`
-commands. Page commands should use the Loom Browser Access endpoint supplied by
-the task, for example
+The host shell is policy-restricted to `agent-browser` attached through the
+exact Loom Browser Access CDP endpoint supplied to the child task,
+`agent-browser` discovery (`which agent-browser`, `command -v agent-browser`),
+`pwd`, `ls`, and bounded workspace-local `find` commands. Page commands should
+use the leased endpoint, for example
 `agent-browser --cdp http://host.docker.internal:9362 snapshot -i`. Bare
 `agent-browser open` / `snapshot` launches are denied so the child cannot race
-the host's live browser profile. `agent-browser eval` is not available in this
-profile; browser subagents should inspect pages with commands such as
-`snapshot`, `get`, `find`, locator actions, and normal browser interactions.
+the host's live browser profile. `agent-browser` is fail-closed to a small
+positive allowlist: `open` for HTTP(S) URLs, `snapshot`, `get title/url/text`,
+bounded `wait`, and ref-based `fill`, `type`, `select`, `check`, `click`, and
+`press`.
+
+Host-target skill scripts run with a cleared subprocess environment plus a
+controlled `PATH` and explicit `CF_HARNESS_*` / `SKILL_*` variables. They do not
+inherit ambient provider tokens, developer secrets, app credentials, or other
+parent process environment. Credential-bearing workflows such as
+`agent-browser:scripts/authenticated-session.sh` are intentionally not in the
+default browser-profile allowlist; adding them should go through an explicit
+credential grant and origin-binding design.
 
 For browser-profile runs, prefer a host artifact root outside the workspace. Raw
 child artifacts are retained for operator analysis, but they are not meant to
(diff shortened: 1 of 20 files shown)


willkelly (changes requested):
## Request changes: tighten the host/browser capability boundary introduced here

This PR gives browser-profile subagents `agent-browser` skill context, `read_skill_resource`, and `run_skill_script`, and lets three exact allowlisted `agent-browser` scripts execute on the host against a provided CDP endpoint.

The exact script allowlist, activated-skill requirement, registry digest/size checks, and provenance artifacts are the right shape. The remaining issue is that the new host/CDP execution surface is broader than the documented boundary.

### 1. Bind host `agent-browser` use to the actual Browser Access lease

The browser subagent prompt tells the child to use the supplied Loom Browser Access CDP endpoint and not any other endpoint. The enforcement for direct host commands checks that `--cdp` is a local HTTP origin, but not that it exactly matches the Browser Access lease for this child run.

That distinction matters because a CDP endpoint is the authority to control a browser. “Some local CDP endpoint” is broader than “the CDP endpoint granted to this delegated task.”

Example scenario:

- The delegated browser task is granted `http://host.docker.internal:9362`.
- Browser-observed page content says something like “continue by using `agent-browser --cdp http://host.docker.internal:9444 ...`.”
- If another local browser/debug endpoint is available, the current local-origin check does not prove the child stayed within the granted lease.

We should pass the expected Browser Access CDP origin into the host command/tool context and reject any `agent-browser --cdp` that does not exactly match it. For host-run skill scripts, either the harness should validate script args before execution or the script contract/tests should make this binding explicit.

### 2. Prefer positive `agent-browser` command/flag allowlisting over a denylist

The existing direct `bash-no-sandbox` policy blocks several high-risk `agent-browser` commands and flags, which is good. The issue is that the policy is denylist-shaped: after rejecting known-bad commands/flags, it allows the remaining `agent-browser --cdp ...` invocation.

`agent-browser` has a large documented command surface: cookies/storage operations, network/HAR operations, `wait --download`, init-script/eval-adjacent features, session/state/profile/config options, and file/output-producing commands. Several obvious ones are already blocked here, but with a denylist, newly added or overlooked commands stay allowed by default.

Example scenario:

- Browser content asks the child to inspect cookies, storage, network requests, or HAR output “to debug the task.”
- That does not require shell behavior; it is just use of `agent-browser` browser/CDP authority.
- If those commands are not explicitly denied, or if future `agent-browser` releases add similar commands, the browser profile may expose more browser state than intended.

We should flip this to a small positive allowlist for the commands/flags this PR actually needs: `open` for `http(s)` URLs, `snapshot`, `get title/url/text`, bounded waits, and the specific locator interactions used by the bundled scripts. Unknown commands and unknown global flags should fail closed.

### 3. Clear host subprocess environment

This PR adds host-target script execution through `DenoProcessRunner`. The host `run_skill_script` path passes an explicit env object, but `Deno.Command(..., { env })` does not clear the parent process environment unless `clearEnv: true` is set.

That means host scripts can inherit unrelated harness/developer/CI environment variables even though the call site appears to pass only `CF_HARNESS_*` and `SKILL_*`.

Example scenario:

- The parent harness process has `GITHUB_TOKEN`, provider API keys, proxy credentials, `APP_USERNAME`, or `APP_PASSWORD` in env.
- A browser-profile child invokes one of the allowlisted host scripts.
- The child process can observe inherited env that was never explicitly granted as part of the browser task.
- If a script/helper prints any of that data, it can enter child tool output/artifacts.

We should add `clearEnv` support to `ProcessRunRequest` / `DenoProcessRunner` and use it for host-target skill scripts. Then pass only the intended variables, plus a controlled `PATH` if needed. A regression test with `SECRET_SHOULD_NOT_LEAK` in the parent env would make this concrete.

### 4. Make host script stdout/stderr exposure explicit and tested

The docs say raw browser-profile observations should stay child-local and the parent should see only the sanitized subagent return channel. That is a good boundary. The implementation also intentionally excludes successful host-target `run_skill_script` output from sandbox CFC mediation.

That means host script stdout/stderr are still visible to the child model and retained in child artifacts, even if the parent only receives the sanitized return.

Example scenario:

- `capture-workflow.sh` opens a page and prints page text/snapshot to stdout.
- The page contains instructions aimed at the model, misleading summaries, links, or other browser-controlled text.
- That text becomes child-visible tool output and can influence the child’s final summary unless the child return is structured/sanitized as intended.

We should document and test this boundary explicitly: host-target script stdout/stderr are browser-derived child-run observations, exposed to the child model and retained in child artifacts, while parent visibility is limited to the sanitized subagent return. If there is intended shaping/truncation/redaction for host `run_skill_script` output, this PR should exercise it.

### 5. Treat `authenticated-session.sh` as a credential-bearing capability

This PR puts `agent-browser:scripts/authenticated-session.sh` in the browser profile’s exact script allowlist. That script is materially more sensitive than a capture script because it can fill username/password values and submit a login form.

Example scenario:

- A browser task lands on or is redirected to an unexpected login page.
- Browser-observed content or page structure leads the child to run the authenticated-session workflow with model-selected refs.
- If credentials are available via env, the script can submit them to an arbitrary `http(s)` login URL unless credential use is explicitly origin-bound.

We should either remove `authenticated-session.sh` from the default browser-profile allowlist for now, or make credential availability explicit and origin-bound. At minimum, document/test that it does not print credentials, cookies, bearer tokens, reusable session material, or local profile paths into stdout/stderr/artifacts.

### Summary of requested changes

Before merge, this should enforce or test the boundaries it documents:

1. Exact CDP lease binding for host `agent-browser` use.
2. Positive allowlist for `agent-browser` commands/flags.
3. `clearEnv: true` / explicit env allowlist for host subprocesses.
4. Explicit docs/tests for host `run_skill_script` stdout/stderr visibility and parent sanitization.
5. Explicit handling for `authenticated-session.sh` credentials/session material.

willkelly (approved):
Thanks for the follow-up. The browser-profile/host-script security blockers I raised are addressed in this update.

I verified the updated head `42ffb00c735038f2a0a8bf24115582635d0d8d0b` against the original concerns:

- Host `agent-browser` use is now bound to the Browser Access lease for both direct `bash-no-sandbox` commands and host `run_skill_script` args.
- The direct `agent-browser` policy now fails closed to a small positive command/flag allowlist.
- Host subprocess execution uses `clearEnv: true` with an explicit env, and there is a regression test proving inherited env is cleared.
- `authenticated-session.sh` is no longer in the default browser-profile script allowlist; docs call out that credential-bearing workflows need explicit credential grants and origin binding.
- The host stdout/stderr boundary is documented and covered by tests showing browser observations remain child-local and parent-visible structured returns are opaque-linked.

Validation I ran:

- `deno check packages/cf-harness/src/main.ts`
- `deno fmt --check` on the reviewed cf-harness source/test files
- `GIT_MASTER=1 git diff --check 4b9f656c2dab495e9cd011be074213c74203e160..HEAD -- packages/cf-harness`
- `deno test --allow-read --allow-write --allow-run --allow-env packages/cf-harness/test/browser-host-command-policy.test.ts packages/cf-harness/test/process-runner.test.ts`
- `deno test --allow-read --allow-write --allow-run --allow-ffi --allow-net --allow-env packages/cf-harness/test/tools-execution.test.ts`
- `deno test --allow-read --allow-write --allow-run --allow-ffi --allow-net --allow-env packages/cf-harness/test/prompt-loop.test.ts`
- `deno test --allow-read --allow-write --allow-run --allow-env --allow-net --allow-ffi packages/cf-harness/test/interactive-chat-service.test.ts packages/cf-harness/test/interactive-chat-stdio.test.ts`

Non-blocking nits/follow-ups:

1. The generic `agent-browser` skill docs still list broader commands and auth workflows than the cf-harness browser profile permits. A short cf-harness-specific note in those skill docs would help future agents/operators understand that browser-profile runs use the leased CDP endpoint, reduced command/script allowlist, and no default credentialed workflow.

2. The new tests cover stdout/structured-return laundering well. It would still be useful to add a small stderr-specific parent-boundary test, since host scripts can emit sensitive browser/debug output on stderr too.

3. The CDP lease matching tests cover the important mismatch case. Future hardening could add URL/argument edge cases around normalization and smuggling, e.g. duplicate/positioned flags, userinfo in URLs, IPv6 loopback variants, and path/query/fragment confusion.

4. Since `authenticated-session.sh` is now excluded by the default allowlist, the blocker is resolved. A direct negative test that attempts to run it through the browser profile and asserts `script_not_allowlisted`/denial would make that guarantee more obvious.

Approving from the standpoint of the cf-harness browser subagent / host `agent-browser` security concerns.

PR #3752, opened 2026-05-29
fix(runner): resolve cell scope from schema, not by stamping links (CT-1623)

seefeldb:
## Summary

Fixes [CT-1623](https://linear.app/common-tools/issue/CT-1623): reading a `perSession` cell from a broader (space) reading context silently returned nothing (dead UI, no error).

**Root cause:** cell scope was eagerly *stamped onto links during navigation* (`key()`, `validateAndTransform`). When a session-scoped cell was reached through a space-scoped container, the stamped scope re-addressed the *container's* link to its empty session instance, so the read silently returned `undefined`. The follow-scope cap also ignored `asCell` entry scope, and writes didn't move data to its declared scope.

This moves to the model in `docs/specs/scoped-cell-instances.md`: **scope lives in the schema; links carry only a base scope; the effective scope is realized on read (as a follow cap) and on write (via redirect links).**

## Changes

- **Follow-scope cap honors `asCell` entry scope** (`traverse.ts`, `link-resolution.ts`): a value declared `asCell:[{kind:"cell",scope:"session"}]` now permits following session links, while a `user` cap still blocks them.
- **`key()` only extends the path, never the scope** (`cell.ts`): scope is no longer stamped onto the navigated link (neither inline nor `asCell` entry scope). Stamping made reads hit the wrong scoped instance of the container.
- **`validateAndTransform` keeps the followed link's own resolved scope** instead of overwriting it with the `asCell` scope (`schema.ts`); the default-value creation path still seeds scope.
- **`diffAndUpdate` enforces the narrow-write rule in one general place** (`data-updating.ts`): when a cell's schema declares a narrower scope than the link's base scope, content writes go into the narrower-scope instance and a redirect link is left at the base scope so broader-scope readers follow it. Writing a link value is exempt (a reference carries its own target scope).
- **`getCell` seeds base scope from a top-level schema scope** (`runtime.ts`).

## Tests

- New failing-then-passing repro: a `lift` reading a session-scoped cell passed via pattern input.
- Updated the two tests that asserted the old eager-stamping; added a test pinning that `key()` changes only the path, carries scope on the schema, and performs no reads; added a `getCell` base-scope test.
- **Full runner suite: 468 passed, 0 failed.**
- Latent `.tsx` patterns (`parking-coordinator`, `cozy-poll-scoped`) transform/typecheck cleanly.

## Notes

- The `narrowestReadScope` narrowing in `pattern-binding.ts` (computation-output concern) is intentionally kept separate; folding it into `diffAndUpdate` would over-narrow unrelated writes. The schema-declared-scope rule from the spec now lives centrally in `diffAndUpdate`.
- The bug report's literal closure-in-`.map()` shape can't be reproduced at the runner level (the TS transformer lifts such closures into explicit inputs — exactly the lift form the minimal repro covers). Full in-browser reactive verification of the latent patterns would need a running toolshed; the underlying scope-follow mechanism they rely on is covered by the green tests.

---
## Summary by cubic
Fixes CT-1623 by resolving cell scope from the schema instead of stamping it onto links, so session-scoped cells read and write correctly even when reached from space-scoped containers. Also keys write batches by scope to prevent cross-scope write corruption and make nested scope-narrowing compose cleanly.

- **Bug Fixes**
  - Follow-scope cap now honors a top-level `scope` or the outermost `asCell` entry scope.
  - `Cell.key()` only extends the path and walks the schema; it never changes link scope or reads storage.
  - Materialization keeps the followed link’s storage-resolved scope; creation paths still seed scope from `asCell` as needed. `Runtime.getCell` seeds base scope from an explicit arg, else a top-level schema scope, else `space`.
  - `diffAndUpdate` enforces schema-declared scope at all depths: narrower fields write to their narrower-scope instance and leave a redirect at the broad slot; arrays keep the container broad and store per-element links to narrower items; writing link/cell values is exempt. Narrowed content and the broad redirect are applied in one change set, and storage creates any missing ancestor containers.
  - Storage: write-batch grouping now includes scope so writes to different scoped instances of the same id aren’t merged and misapplied.
  - Tests: added bulk nested-scope write routing, `key()` no-read and scope-carry tests, `getCell` base-scope seeding, and a repro for a lift reading a session-scoped input; updated prior scope-stamping assertions. All runner tests pass.

<sup>Written for commit b2b4d8ab320894d46394d340d1abded252ac4e18. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3752?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/scoped-cell-instances.md
+++ b/docs/specs/scoped-cell-instances.md
@@ -278,6 +278,42 @@ scope-dependent.
 This behavior applies equally when a broader scoped output contains a link to a
 narrower scoped computation result.
 
+### Navigation Does Not Carry Scope
+
+Path navigation (for example `cell.key(...)`) only extends the link's path and
+walks the schema to the child schema. It must not change the link's scope, and
+it must not read storage. Scope lives in the schema: it is resolved lazily as a
+follow cap on reads and as the target scope on writes. The link's own `scope`
+remains the base scope established when the cell was created (from an explicit
+scope, a top-level schema scope, or `space`).
+
+Stamping a child schema's scope onto the navigated link is incorrect: the link
+addresses the containing instance, so stamping a narrower scope would read or
+write the wrong scoped instance of the container.
+
+### Scoped Write Placement
+
+A write places content in the instance whose scope the content's schema
+declares, at every nesting level, not only at the top of the write:
+
+- When a field's schema declares a scope narrower than the instance being
+  written, the field's content is written into the narrower-scope instance (same
+  id, same path, narrower scope key) and the broader-scope slot holds a link to
+  it. Readers at the broader scope follow that link to the narrower instance.
+- For arrays whose items declare a narrower scope, each element is an
+  independent link to its narrower-scope element instance. The array container —
+  including its length — stays in the broader scope; there is no single redirect
+  for the whole array. This mirrors the `map` rule: an input-list-scoped array of
+  links to narrower scoped element instances.
+- Writing a link or cell reference is exempt: a reference already carries its own
+  target scope, so it is stored as-is at the broader slot.
+
+The narrower-scope instance is a different scoped instance of the same id, so
+its container structure (the value root and any intermediate objects/arrays
+along the written path) is created as needed; the broader instance keeps the
+non-scoped data and the links. Writes to different scope instances of one id are
+applied as separate operations, never merged into a single document write.
+
 ## Schema Semantics
 
 `JSONSchemaObj` gains an optional Common Fabric extension:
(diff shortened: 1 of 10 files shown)


seefeldb:
Superseded by #3757, which re-derives the CT-1623 fix against current main (after the process-cell removal / scoped-cell rework). The new fix is a single targeted change: stop promoting a stripped asCell entry's scope onto the schema's top-level scope.

PR #3756, opened 2026-05-30
fix(runner): align CFC integrity propagation

willkelly:
## Summary
- Verified each untrusted report against the sibling CFC specs and the runner implementation.
- Fixed confirmed label propagation and enforcement gaps, with regression tests.
- Added a scoped negative regression for the write-authority report because the broad trust-boundary claim was not confirmed.

### G1 — Integrity unions for combined label views: confirmed and fixed

**Problem:** Distinct dereference/display label-view combinations were preserving the union of integrity atoms. That can overstate whole-output integrity when multiple consumed observations contribute to one value.

**Spec coverage:**
- `specs/cfc/08-06-combination-rules.md` §8.6.2 says integrity combination is a meet: atoms **intersect**, not union.
- `specs/cfc/08-09-runtime-label-propagation.md` §8.9.2 applies that rule to `combinedFrom`: confidentiality is CNF-joined and integrity uses `intersectAtoms(...)`.
- `specs/cfc/04-label-representation.md` §4.2.2 also states that joining multiple consumed observations intersects integrity atoms.

**Change:** Label-view merging now supports explicit integrity modes. Same-value reconciliation/exact-copy style merges keep union semantics, while distinct consumed-observation joins use intersection in dereference/display paths.

### G2 — `flowPrecisionClaim` accepted as advisory-only: confirmed and fixed

**Problem:** `flowPrecisionClaim` was trust-gated metadata, but no evaluator consumed the claim. In enforcing modes, silently accepting an unsupported precision claim could make the boundary appear more precise than the runtime can prove.

**Spec coverage:**
- `specs/cfc/08-09-runtime-label-propagation.md` §8.9.1 says precision claims that are less restrictive than the conservative default must be gated by trust in the implementation identity for `https://commonfabric.org/cfc/concepts/flow-taint-precision`.
- The same section says if the trust check fails the runtime MUST use the conservative default or reject, not the claim. Because this runner does not yet evaluate the claim into a conservative-vs-claimed label choice, enforcing modes must fail closed.
- `specs/cfc/08-10-validation-at-boundaries.md` §8.10.1 requires boundary attempts to reject/abort when verification fails.

**Change:** Enforcement modes now treat `flowPrecisionClaim` as an unsupported trust-sensitive claim and reject it fail-closed instead of accepting it as advisory-only.

### G3 — `writeAuthorizedBy` trust-boundary concern: broad claim not confirmed; regression added

**Problem investigated:** The report suggested `writeAuthorizedBy` could be satisfied through attacker-controlled trusted metadata/trust snapshots. The implementation still authorizes protected writes from trusted runtime transaction state and implementation identity, not from a trust snapshot alone, so the broad implementation bug was not confirmed.

**Spec coverage:**
- `specs/cfc/08-15-write-authority.md` §8.15.6 defines write authorization as the handler identity being present in the field `writeAuthorizedBy` set.
- `specs/cfc/08-15-write-authority.md` §8.15.9 describes the two-layer model: event integrity may trigger the handler, but write authorization separately checks whether the handler identity is authorized for the field.
- `specs/cfc/08-10-validation-at-boundaries.md` §8.10.1 requires write boundaries to derive attempted writes from the transaction write log and verify before commit.

**Change:** Added a negative regression proving that a trusted `trustSnapshot` alone does not authorize a `writeAuthorizedBy`-protected write.

### Persisted link-reference provenance: guarded regression from review

**Problem:** While fixing G1, a naive global intersection during persisted metadata coalescing would have dropped legitimate same-value/link-reference provenance. That path is same-value reconciliation, not a distinct-input combination.

**Spec coverage:**
- `specs/cfc/08-06-combination-rules.md` §8.6.2 and `specs/cfc/08-09-runtime-label-propagation.md` §8.9.2 cover distinct input combination.
- `specs/cfc/08-09-runtime-label-propagation.md` §8.9.2 exact-copy handling preserves the input label when the runtime verifies referential equality, so same-value preservation should not be weakened by the distinct-input meet rule.

**Change:** Kept persisted metadata coalescing on union semantics and added a regression that carried-link metadata retains both same-value provenance atoms.

## Tests
- `deno test --allow-ffi --allow-env --allow-read --allow-write=/tmp,/var/folders --allow-run=git test/cfc-boundary.test.ts test/cfc-label-view.test.ts test/cfc-flow-precision.test.ts`
- `deno task check`

### DIFF
--- a/packages/runner/src/cell.ts
+++ b/packages/runner/src/cell.ts
@@ -1347,7 +1347,9 @@ export class CellImpl<T extends FabricValue>
       this.tx,
       this.synced,
       undefined,
-      mergeCfcLabelViews([this._cfcLabelView, dereferenceView]),
+      mergeCfcLabelViews([this._cfcLabelView, dereferenceView], {
+        integrity: "intersection",
+      }),
     );
   }
 
--- a/packages/runner/src/cfc/label-view-core.ts
+++ b/packages/runner/src/cfc/label-view-core.ts
@@ -1,4 +1,5 @@
 import { encodePointer } from "../../../memory/v2/path.ts";
+import { deepEqual } from "@commonfabric/utils/deep-equal";
 import { uniqueCfcAtoms } from "./observation.ts";
 
 export type IFCLabel = {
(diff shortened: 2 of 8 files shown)


willkelly:
Closing — all three parts resolved on main by other routes.

To be clear about why: CFC in labs is *not* dead. It is one of the most active areas of the tree — 354 cfc-titled commits since this branch's merge-base, with `packages/runner/src/cfc` touched as recently as today. The reason to close is that the area moved two months past this PR and decided each of its questions differently.

- **G1** (integrity must intersect, not union): the rule landed, but class-aware and at the right layer — `hereditaryMeet` at `cfc/prepare.ts:1694`, driven by the new `cfc/atom-classes.ts` registry that types every atom hereditary / value-bound / provenance. Only hereditary atoms meet; provenance and value-bound atoms never propagate through combination. This PR intersects *all* integrity atoms at label-view merge sites, then bolts on a union opt-out for the one call site where that destroyed provenance. Main solved the same problem generically by typing the atoms. Separately, inv-12 Stage 0 (#4624) made the layer this patches no longer an enforcement input.
- **G2** (fail closed on flow-label propagation): #4011.
- **G3**: #4852 and #4870. The generalized fail-closed rule this argued for lives in `runner/test/cfc-unsupported-ifc-keys.test.ts`.

Nothing to salvage.

PR #3774, opened 2026-06-01
test(runner): adversarial corpus for the ESM module verifier (+ fix helper-shadow bypass)

seefeldb:
## What

Hardens the SES ESM module verifier (`verifyCompiledModuleBody`, the security boundary) with a red-team adversarial test corpus, and fixes one real bypass the corpus surfaced.

A sub-agent brainstormed 40+ creative attacks (no execution) across: ASI/statement-merge desync, tokenizer confusion (regex-vs-divide, nested templates, U+2028/2029), trusted-name shadowing, trusted-builder callback indirection, `__cf_data`/`schema` argument smuggling, default-export & reexport-getter laundering, top-level executable/mutable forms, export-form abuse, and import-fast-path / require-boundary escapes. They are encoded as 44 fixtures in `packages/runner/test/esm-verifier-adversarial.test.ts` — reject cases double as bypass detectors; a handful of legitimate forms are pinned as negative controls.

## Real bypass found & fixed

The `__exportStar(require(...), exports)` re-export fast-path (and the `__importDefault`/`__importStar` require-import forms) trusted the helper *identifier* unconditionally. A module could declare a local `function __exportStar(m){ globalThis.steal = m; }` — registered as a normal top-level function — and the fast-path would skip the call, so the **malicious local ran at module init**.

**Fix:** generalize shadow detection (previously `require`-only) to the TS interop helper identifiers the fast-paths trust, *excluding* canonical tslib helper declarations (which are the trusted helpers themselves, not shadows). When a helper is locally shadowed, its fast-path is disabled and the statement falls through to `classifyModuleItems`, which rejects the call as a local-callable result — fail-closed, matching AMD.

## Documented, not changed

`__cf_data`/`schema` arguments are opaque — `verifyTrustedDataCall` checks only the argument *count*, never recurses — so a side-effecting computed value inside the wrapped literal is accepted. This is a deliberate, AST-free boundary **shared with the AMD verifier**, is not reachable via the trusted transformer (it only wraps statically-classified data), and the runtime data-freeze is the backstop for the value itself. A purity check is tracked as separate hardening work. The two fixtures pin the current behavior so a future change is a conscious one.

## Scope

- Behind the existing default-off `esmModuleLoader` flag; no behavior change to the AMD enforcement path.
- New tests + a contained verifier fix; all existing verifier suites (`esm-module-body-verifier`, `esm-verifier-parity`) still green.

---
## Summary by cubic
Adds an adversarial test corpus for the ESM module verifier and closes a helper-shadow bypass with multi-declarator shadow detection and per-helper fast-path gating. No AMD-path change; the ESM path stays behind the default-off `esmModuleLoader` flag.

- **New Features**
  - Added adversarial fixtures covering ASI merges, tokenizer tricks, helper shadowing (incl. non-first declarators), import fast-paths, top-level executable/mutable forms, and export abuse; reject cases act as bypass detectors.
  - Pinned a few legit cases as negative controls, including opaque `__cf_data`/`schema` arguments; all existing verifier suites remain green.

- **Bug Fixes**
  - Detect top-level shadows of `require`, `__exportStar`, `__importDefault`, and `__importStar` across function/class names and every declarator in `const`/`let`/`var` lists using depth-aware parsing; exclude canonical `tslib` helper declarations.
  - Gate import-helper fast-paths per-helper; if a helper is shadowed, disable only that fast-path and fall through to classification, closing the `__exportStar(require(...), exports)` re-export bypass.

<sup>Written for commit 21fb7966052a8799a5dee854ec85a854e8c26e5b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3774?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/sandbox/module-record-verifier.ts
+++ b/packages/runner/src/sandbox/module-record-verifier.ts
@@ -3,7 +3,13 @@ import {
   type BindingInfo,
   classifyModuleItems,
 } from "./compiled-bundle-verifier.ts";
-import { parseFunctionText } from "./compiled-js-parser.ts";
+import {
+  findTopLevelEquals,
+  parseFunctionText,
+  splitTopLevelCommaList,
+  type StatementChunk,
+  trimRange,
+} from "./compiled-js-parser.ts";
 import {
   isAllowedAuthoredImportSpecifier,
   isRuntimeModuleIdentifier,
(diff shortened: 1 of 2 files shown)


seefeldb:
The cubic/codex P1 (multi-declarator helper shadow, e.g. `const ok = 0, __exportStar = function(m){...}`) was caught in self-review and is already fixed in commit 21fb79660: shadow detection now uses depth-aware declarator extraction (`splitTopLevelCommaList` + `findTopLevelEquals`) across every declarator slot plus function/class names, rather than a leading-anchored regex. Covered by three new `multi-declarator shadow of …` fixtures. CI, cubic, codex, and the perf check are all green.

PR #3775, opened 2026-06-01
fix(runner): write-once exports in the ESM loader (block init-time export corruption)

seefeldb:
## Context

Follow-up to the verifier adversarial-hardening work (#3774), under a sharpened threat model: **the TS compiler and CF transformer are untrusted** — the verifier sees arbitrary attacker-controlled compiled CJS. Three red-team passes converged on a single residual class the (deliberately AST-free) verifier cannot catch: a **write smuggled into the evaluation of an otherwise-accepted expression**.

Flagship: `const x = cf.__cf_data((exports.token = evilClosure, 1));` — the comma side effect overwrites an already-assigned export with a live, un-copied closure, then `__cf_data` copies only `1`. `verifyTrustedDataCall` checks argument *count* only, so the write is invisible to verification.

## What is already covered (verified — no change)

- **SES namespace immutability**: a sibling cannot reassign or add `ns.x` (`Cannot set property … of module exports namespace`). Confirmed by probe.
- **SES lockdown** hardens shared intrinsics (`Object`/`Array`/`Function.prototype`) and the compartment `globalThis` is `Object.freeze`d → prototype-pollution and global-stash writes throw.
- **`__cf_data`/`schema`** (= `freezeVerifiedPlainData`) deep-freeze their result and throw on functions → exported plain-data internals are immutable and closures are stripped from results.

## The gap & fix

Unprotected: a smuggled overwrite of a **declared export's value during the module's own init**, before the loader snapshots values into the (then-frozen) namespace.

The ESM loader now hands the body a **write-once exports** object: re-assigning / redefining / deleting an already-populated export throws → the smuggle fails closed. The `void 0` forward-declaration placeholder is preserved (canonical `exports.x = void 0; … exports.x = real;` still works); re-export getters and the `__esModule` marker define exactly once.

## Deliberately *not* done: deep-freeze exported values

Evaluated per discussion and **rejected empirically** — it breaks the runtime, which attaches metadata to exported pattern recipes after load (`value.program = program` in `builder/factory.ts`), so exported builder graphs must stay extensible. Namespace-binding immutability (SES) + data freezing (`__cf_data`) already cover the value surface; deep-freeze would add breakage without adding protection here.

## Scope

- Behind the default-off `esmModuleLoader` flag (only the ESM loader's `execute` uses this path); no change to the AMD enforcement path.
- New unit tests pin write-once semantics (overwrite/stash/delete blocked; forward-decl, re-export getter, `__esModule` allowed; fails-closed on smuggle-before-real). Full ESM verifier/loader/engine/pattern-run suite green (175 steps), including a real multi-file pattern running through the ESM loader.

---
## Summary by cubic
Hardened the ESM loader with write-once `exports` and a frozen `module` wrapper to block init-time export corruption and `module.exports` swap bypasses. Prevents smuggled writes like `(exports.token = evil, 1)` and `(__cf_data((module.exports = evil, 1)))` from corrupting exports before snapshot.

- **Bug Fixes**
  - Module body receives a write-once `exports` proxy; reassign/redefine/delete after a real value throws.
  - Supports TS forward-decl (`exports.x = void 0; … exports.x = real`); allows defining a re-export getter and `__esModule` exactly once.
  - Freeze the `module` wrapper and snapshot declared exports directly from the write-once object (never from `module.exports`); ignores undeclared stashes.
  - Exported values are not deep-frozen to avoid breaking runtime metadata; SES namespace immutability and `__cf_data` deep-freezing still apply.
  - Scoped to the ESM path behind the `esmModuleLoader` flag (default off); added tests covering write-once semantics, declared-only snapshot, and the `module.exports` swap bypass.

<sup>Written for commit 46ed1f909e59f7b647649f6ef665d7612053e6ff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3775?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/sandbox/module-record-compiler.ts
+++ b/packages/runner/src/sandbox/module-record-compiler.ts
@@ -25,6 +25,46 @@ import type { VirtualModuleRecord } from "./esm-module-loader.ts";
 
 const TARGET = ts.ScriptTarget.ES2023;
 
+/**
+ * Create a write-once exports target for a module body. Re-assigning,
+ * redefining, or deleting a property that already holds a real (non-`undefined`)
+ * value throws. A `void 0` placeholder — the TS `exports.x = void 0;` forward
+ * declaration — leaves the property unlocked so the subsequent real assignment
+ * is permitted, and the real assignment then locks it. This blocks export
+ * corruption smuggled into the evaluation of an otherwise-accepted expression,
+ * which the (deliberately AST-free) verifier cannot detect.
+ */
+export function createWriteOnceExports(): Record<string, unknown> {
+  const target: Record<string, unknown> = {};
+  const locked = new Set<string | symbol>();
+  const denyRelock = (key: string | symbol): never => {
+    throw new TypeError(
+      `Module export '${String(key)}' is write-once and was already assigned`,
+    );
+  };
+  return new Proxy(target, {
+    set(t, key, value) {
+      if (locked.has(key)) denyRelock(key);
+      (t as Record<string | symbol, unknown>)[key] = value;
+      if (value !== undefined) locked.add(key);
+      return true;
+    },
+    defineProperty(t, key, descriptor) {
+      if (locked.has(key)) denyRelock(key);
+      Reflect.defineProperty(t, key, descriptor);
+      const isUndefinedPlaceholder = "value" in descriptor &&
+        descriptor.value === undefined;
+      if (!isUndefinedPlaceholder) locked.add(key);
+      return true;
+    },
+    deleteProperty(_t, key) {
+      throw new TypeError(
+        `Module export '${String(key)}' cannot be deleted (write-once)`,
+      );
+    },
+  });
+}
+
 /** Per-module compiled artifact, cacheable by module hash (Phase 4). */
 export interface CompiledModuleArtifact {
   exports: string[];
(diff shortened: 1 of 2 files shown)


seefeldb:
Self-review pass (sub-agent) found one **P1**, now fixed in 46ed1f909:

- **`module.exports = evil` bypass**: the wholesale twin `__cf_data((module.exports = evil, 1))` swapped out the write-once object, and the loader snapshotted from `moduleObject.exports` (the replaced object), bypassing write-once. Fixed by (a) freezing the `module` wrapper so the strict-mode reassignment throws (fails closed), and (b) snapshotting the declared exports from the write-once object **directly**, never `module.exports`. Extracted into a testable `populateModuleExports` helper with coverage for both the strict-throw and the sloppy-mode snapshot-integrity paths.

Self-review confirmed PASS on: no false-positive breakage of canonical TS→CJS shapes (`exports.x = void 0` chains, `__createBinding`/`__exportStar` re-export getters guarded by `hasOwnProperty`, `__esModule` marker, default export, Symbol keys); `Object.assign`/`Reflect.defineProperty` hit the traps; the snapshot read path (`get`/`ownKeys`/getters) is unaffected.

Residuals (intentionally out of scope here, tracked separately):
- **AMD loader parity**: the same expression-smuggle class applies to the AMD factory body (shared verifier). The AMD path is the one the ESM loader is replacing and is slated for removal at the flag flip, so it is not hardened here.
- **`default` re-export getter** returning a live closure (red-team vector 7/8): the getter is defined once (write-once allows it) and its body runs on access (execution-isolation territory; writes hit SES-frozen targets). Noted for the execution-sandbox work.

seefeldb:
Both bots flagged the `module.exports = attackerObject` reassignment bypass — same P1 the self-review caught, already fixed in commit 46ed1f909: the `module` wrapper is now `Object.freeze`d (so the strict-mode reassignment throws → fails closed) and the namespace is snapshotted from the write-once object directly, never `module.exports`. Covered by new `populateModuleExports` tests (strict-throw + sloppy-mode snapshot integrity). CI, cubic, codex, and the perf check are green.

PR #3777, opened 2026-06-01
feat(runner): freeze exported patterns in the ESM loader (metadata → WeakMaps)

seefeldb:
## Context

Closes the residual export-surface gap left by the write-once work (#3775): exported pattern recipes / functions were still **mutable**, so an importing module could corrupt their internals (`importedPattern.nodes.push(...)`, stash a closure on an exported function, overwrite `program`). We now `harden()` (transitively freeze) every value the ESM loader exports.

## Why it wasn't possible before

The engine attaches metadata to exported patterns **after** load, which a freeze would block:
- `pattern.program` — rehydration source, set by `exportsCallback` and at registration.
- `pattern[unsafe_verifiedLoadId]` — CFC verified-load identity, set via `Object.defineProperty` while the pattern was extensible.

## The change

Both move into module-level **WeakMaps** (`builder/pattern-metadata.ts`) via `get/setPatternProgram` and `get/setVerifiedLoadId`. A WeakMap write doesn't mutate the (now frozen) object, so the associations still attach after `harden()` — sequencing works because the loader freezes during import, and the engine's `exportsCallback`/`seedVerifiedLoadIds` run afterward using the side-table.

The `program?` field and `unsafe_verifiedLoadId` symbol are **removed from the `Pattern` type**, so the type checker flagged every access site; all were converted (`factory`, `pattern-manager`, `json-utils`, `pattern-binding`, `executable-registry`).

## Why freezing is safe

Patterns are static definitions (schemas + node graph), not live reactive state. Derivation/binding creates fresh mutable copies, and the remaining `unsafe_originalPattern`/`unsafe_parentPattern` markers are set on those copies, never the frozen original.

## Scope & validation

- The **freeze** is ESM-loader-only (`populateModuleExports`, behind the default-off `esmModuleLoader` flag). The **WeakMap metadata move** is path-agnostic and behavior-preserving.
- E2E: a real multi-file reactive pattern compiles, is confirmed `Object.isFrozen`, **still runs** through the ESM loader, and still associates its rehydration `program` via the WeakMap.
- Green: CFC implementation-identity, engine-verified-functions, JSON serialization, pattern-binding, pattern-composition (`derive`-returns-pattern), pattern-manager (program/verified-load-id + persistence), and the ESM execution/write-once suites.

---
## Summary by cubic
Freeze all exported values in the ESM loader to prevent mutation of imported patterns, and move pattern metadata to WeakMaps so `program` and verified-load IDs still attach after load. If `harden()` is unavailable we warn; if `harden()` throws, we fail closed and the module does not load.

- **New Features**
  - ESM loader now `harden()`s exported values to block pattern graph mutation (behind default-off `esmModuleLoader`).
  - Warn when `harden` is missing; throw on harden failure so mutable exports never ship.
  - WeakMap side tables preserve late `program` and verified-load ID attachment.
  - E2E tests cover frozen exports, composed frozen sub-pattern import/instantiate, and `program` access via side-table.

- **Refactors**
  - Added `builder/pattern-metadata.ts` with `get/setPatternProgram` and `get/setVerifiedLoadId`.
  - Removed `program?` and `unsafe_verifiedLoadId` from `Pattern`; updated call sites (`factory`, `pattern-manager`, `json-utils`, `pattern-binding`, `executable-registry`).
  - Behavior unchanged on non-ESM paths.

<sup>Written for commit 751e6652a7929f9a3f2911c66a8c781524cb37cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3777?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/builder/factory.ts
+++ b/packages/runner/src/builder/factory.ts
@@ -48,6 +48,7 @@ import { cellConstructorFactory } from "../cell.ts";
 import { getEntityId } from "../create-ref.ts";
 import { getPatternEnvironment } from "./env.ts";
 import type { RuntimeProgram } from "../harness/types.ts";
+import { setPatternProgram } from "./pattern-metadata.ts";
 import {
   FabricInstance,
   FabricPrimitive,
@@ -130,8 +131,9 @@ export const createBuilder = (options: CreateBuilderOptions = {}): {
   const exportsCallback = (exports: Map<any, RuntimeProgram>) => {
     for (const [value, program] of exports) {
       if (isPattern(value)) {
-        // This will associate the program with the pattern
-        value.program = program;
+        // Associate the program with the pattern via the side-table so it works
+        // even when the exported pattern has been frozen by the loader.
+        setPatternProgram(value, program);
       }
     }
   };
--- a/packages/runner/src/builder/json-utils.ts
+++ b/packages/runner/src/builder/json-utils.ts
@@ -18,6 +18,7 @@ import {
   unsafe_originalPattern,
 } from "./types.ts";
 import { getTopFrame } from "./pattern.ts";
+import { getPatternProgram } from "./pattern-metadata.ts";
 import { deepEqual } from "@commonfabric/utils/deep-equal";
 import { Runtime } from "../runtime.ts";
 import {
@@ -450,6 +451,6 @@ export function patternToJSON(pattern: Pattern) {
     ...(pattern.initial ? { initial: pattern.initial } : {}),
     result: pattern.result,
     nodes: pattern.nodes,
-    program: pattern.program,
+    program: getPatternProgram(pattern),
   };
 }
--- a/packages/runner/src/builder/pattern-metadata.ts
+++ b/packages/runner/src/builder/pattern-metadata.ts
@@ -1,0 +1,59 @@
+import type { RuntimeProgram } from "../harness/types.ts";
+
+/**
+ * Side-table storage for pattern metadata that is associated *after* a pattern
+ * is exported from its module:
+ *
+ * - `program` — the rehydration source (`RuntimeProgram`) attached by the engine
+ *   after compilation/eval and at registration time.
+ * - `verifiedLoadId` — the CFC verified-load identity attached by the engine
+ *   while seeding verified-load ids.
+ *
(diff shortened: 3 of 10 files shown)


seefeldb:
Self-review pass (sub-agent): verdict **SHIP**, with two non-blocking follow-ups now addressed in c7e4fcca7:

- **Test gap** — the only ESM+harden run test exercised a single, non-composed pattern. Added an e2e test that exports a sub-pattern from one module, imports and instantiates it **inside a parent pattern**, runs it through the ESM loader, and asserts correct reactive output + `Object.isFrozen` on the export. This drives a frozen exported pattern through `instantiatePatternNode`/binding — the path where harden-safety lives. Passes.
- **Silent harden fallback** — `hardenExportedValue` now `logger.warn`s when `harden` is unavailable or throws, instead of silently shipping an unfrozen export (observable regression signal; the real engine path always has `harden` via `ensureSESLockdown`).

Self-review confirmed: metadata move is complete (no surviving `pattern.program`/`unsafe_verifiedLoadId` access; PatternMeta persistence round-trip is untouched; no cross-package breakage); WeakMap semantics are equivalent to the old per-object storage (consumers `findOriginalPattern` first, so the incidental copy-time propagation wasn't relied on); and the harden is safe because patterns are **functions** (so `isRecord` excludes them → `unsafe_noteParentOnPatterns` never writes to a frozen exported pattern, and binding deep-copies operate on mutable copies). The function-shape dependency is now locked in by the composed-pattern test.

seefeldb:
Addressed in the latest commit: `hardenExportedValue` no longer catches-and-returns on a `harden()` failure — it now **fails closed** by rethrowing, so the module fails to load (terminal error, cached by SES) rather than silently shipping a mutable export. The harden-unavailable branch (no lockdown, e.g. a non-SES unit harness) still warns and returns, since there is no SES boundary to uphold there.

PR #3778, opened 2026-06-01
fix(runner): validate module-graph import-edge targets (red-team finding)

seefeldb:
## Context

A fresh red-team pass on the ESM **module-loading / identity / capability** surface (the wiring we hadn't hammered, vs. the verifier body-classification the prior corpus covered). The clearest confirmed gap:

## The gap

`verifyModuleGraph` (the structural graph verifier) checked only that each resolved import target was **present** in the records map (`records.has(target)`) — never that the target was itself **content-addressed**. Combined with two facts:
- the compiler's resolution fallback leaves an unknown specifier **verbatim** (`resolutions[spec] = spec`), and
- the authored-import allowlist admits `/`-prefixed specifiers,

a record could carry an import edge whose target is an arbitrary string (`/cf:runtime/commonfabric`, `node:fs`, a sibling's key). Under the untrusted-compiler threat model (records derived from attacker-controlled compiled output), a record could rewire an edge to a **trusted runtime record** and inherit its namespace — with presence being the only, incidental, backstop.

## Fix

`verifyModuleGraph` now:
1. requires every resolved import **target** to match the content-addressed `VALID_SPECIFIER` (`cf:module/` | `cf:runtime/`) **before** the presence check, and
2. rejects any `resolutions` entry that remaps a specifier the record does not declare in `imports`.

The structural verifier now enforces the content-addressing invariant on **edges**, not just keys — closing the reachability hole regardless of the allowlist or the compiler's verbatim fallback.

## Other red-team findings (triaged, not in this PR)

- **Module identity excludes the compiled body** (only authored TS is hashed): *intentional* (TCB-independence, by design); the CFC verified-load bundle id does cover compiled output, so the security-relevant identity covers behavior.
- **`fetch` endowed into the module compartment** (load-time reach via opaque `__cf_data` arg): already covered by the planned separate mitigation (fetch bails during this phase).
- **harden fail-closed DoS via `export *` from a malicious dependency**: a deny-load you'd only hit by `export *`-ing from code you already trust; noted as a possible follow-up.
- **Hash external-leaf string concat** (`runtime:<spec>@<fp>`) collision: suspected, depends on the value-hash serialization; possible follow-up.

## Scope & validation

ESM-loader path only (default-off `esmModuleLoader`). New `verifyModuleGraph` tests (reject non-cf target, reject undeclared-import resolution, accept cf:runtime target); the real ESM execution path (which builds genuine resolutions) and the loader/compiler/engine/pattern-run suites are unaffected.

---
## Summary by cubic
Hardened ESM module graph verification to prevent import-edge rewiring. Import targets must be content-addressed, runtime imports must resolve to their exact `cf:runtime/<id>`, and the graph is now verified at compile time in the ESM engine.

- **Bug Fixes**
  - Validate import targets with `VALID_SPECIFIER` before the presence check in `verifyModuleGraph`.
  - Reject `resolutions` entries for specifiers not declared in `imports`.
  - Require runtime-identifier imports to resolve exactly to `cf:runtime/<specifier>` (no rewires to `cf:module/*`).
  - Run `verifyModuleGraph` during `Engine.compileToRecordGraph` so structural checks apply in production.
  - Added tests for non-`cf:` targets (`/cf:runtime/...`, `node:fs`), undeclared remaps, runtime rewire rejection, and valid `cf:runtime/...` targets.

<sup>Written for commit c7e39a56b46bea375858620a749ae3b8bf2b5989. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3778?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/harness/engine.ts
+++ b/packages/runner/src/harness/engine.ts
@@ -41,7 +41,10 @@ import {
   runtimeModuleRecords,
   type VirtualModuleRecord,
 } from "../sandbox/esm-module-loader.ts";
-import { verifyCompiledModuleBody } from "../sandbox/module-record-verifier.ts";
+import {
+  verifyCompiledModuleBody,
+  verifyModuleGraph,
+} from "../sandbox/module-record-verifier.ts";
 import { popFrame, pushFrame } from "../builder/pattern.ts";
 import {
   ensureSESLockdown,
@@ -337,6 +340,14 @@ export class Engine extends EventTarget implements Harness {
           "ESM compile produced no record for the program entry",
         );
       }
+
+      // Structurally verify the whole record graph (content-addressed
+      // specifiers, well-formed records, and that every import edge resolves to
+      // a content-addressed target). This must run here because the loader is
+      // invoked with `verify: false` in evaluateRecordGraph — graph
+      // verification happens once, at compile time, before any module executes.
+      verifyModuleGraph(graph.records, mainSpecifier);
+
       return { id, graph, mainSpecifier };
     } finally {
       logger.timeEnd("compileToRecordGraph");
--- a/packages/runner/src/sandbox/module-record-verifier.ts
+++ b/packages/runner/src/sandbox/module-record-verifier.ts
@@ -276,6 +276,21 @@ export function verifyModuleGraph(
           `Record ${specifier} has a non-content-addressed import target "${importSpecifier}" -> "${target}"`,
         );
       }
+      // A trusted runtime import must resolve to ITS matching runtime record,
+      // not an arbitrary content-addressed sibling. The body verifier marks
+      // `require("commonfabric")` as a trusted runtime binding based on the
+      // authored specifier; without this, a record could declare
+      // `imports: ["commonfabric"]` with `resolutions: { commonfabric:
+      // "cf:module/evil" }` and be handed a sibling module's namespace under a
+      // trusted-runtime binding (rewire-to-sibling for host APIs).
+      if (
+        isRuntimeModuleIdentifier(importSpecifier) &&
+        target !== `cf:runtime/${importSpecifier}`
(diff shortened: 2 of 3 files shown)


seefeldb:
Both bot P1s were valid and are fixed in the latest commit:

1. **verifyModuleGraph wasn't running on the engine path** (codex) — exactly right: `compileToRecordGraph` ran only `verifyCompiledModuleBody`, and `evaluateRecordGraph` loads with `verify: false`, so the structural graph checks were dead code in production. `compileToRecordGraph` now calls `verifyModuleGraph(graph.records, mainSpecifier)` after body verification (once, at compile time, before the loader runs with verification disabled).

2. **Runtime import rewire-to-sibling** (codex + cubic) — also right: a cf-prefix-only check let `imports: ["commonfabric"]` resolve to `cf:module/evil` while the body verifier trusts it as a runtime binding. `verifyModuleGraph` now requires a runtime-identifier import to resolve **exactly** to `cf:runtime/<specifier>`.

Validated by the ESM engine + pattern-run suites, which now run `verifyModuleGraph` on real compiled graphs (confirming the genuine `commonfabric → cf:runtime/commonfabric` resolution is accepted and no real graph regresses), plus a new rewire-to-sibling rejection test.

(The earlier shell-integration CI failure was on the prior commit, where this change was not yet wired into any executed path — i.e. it could not have been caused by the diff; treating it as environmental and re-running.)

PR #3779, opened 2026-06-01
feat(runner): pattern provenance brand (forged pattern-shaped data can't launder trust)

seefeldb:
## Context

Follow-up to the "is write-once sufficient?" question. A rigorous investigation confirmed: export *values* are verified and only declared names are lifted, so the plain fake-export path is safe — **but** `isPattern` is a purely structural check (`{argumentSchema, resultSchema, nodes}`), so an attacker can forge that shape via `__cf_data({...})` (a frozen plain object). Such a forged top-level export then:
- acquired a `program` (`exportsCallback`),
- acquired a CFC `verifiedLoadId` (`annotateVerifiedPatterns` / `seedVerifiedLoadIds`),
- and was selected as the entry pattern with no re-check,

masquerading as a *verified-loaded pattern* in the trust side-tables. No escalation today (the runner re-resolves every node's implementation through the verified/host index or a SES-sandboxed eval, and data-only nodes throw) — but a latent footgun for any future sink that trusts `isPattern`-shape as provenance.

## Fix

A **provenance brand** — a `WeakSet` stamped *only* by the trusted `pattern()` builder (`builder/pattern.ts` `makePatternFactory`, covering `asScope`/`inSpace` variants; rehydrated patterns re-run through the builder, so they re-brand). Trust-granting sites switch from structural `isPattern` to `isTrustedPattern` (= branded **and** structurally a pattern):
- `exportsCallback` (program association)
- `patternFromEvaluation` entry selection
- `seedVerifiedLoadIds` / `annotateVerifiedPatterns`

The two recursive verified-load-id walks gate the **root** on `isTrustedPattern`, then propagate to structurally-`isPattern` values *within a trusted subtree* — because nested subpatterns are serialized/unbranded but are legit parts of a trusted pattern. So a forged **top-level** export is never seeded, while nested-subpattern id propagation is preserved.

The runner's instantiation logic deliberately keeps structural `isPattern` (it operates on derivation copies and re-resolves node implementations independently).

## Scope & validation

Path-agnostic (the brand + trust gates apply to both AMD and ESM). New `pattern-provenance` suite (forged shape is `isPattern` but not `isTrustedPattern`; branded is trusted; non-shape/primitive aren't). Regression-validated: the nested-verified-load-id propagation test (caught an over-gating attempt and drove the trusted-subtree design), CFC implementation-identity, pattern composition (`derive`-returns-pattern), binding, scope, serialization, and the ESM engine/pattern-run suites — all green (124+ steps in the sweep).

---
## Summary by cubic
Adds a provenance brand to patterns so only builder-produced values (and their derived copies) are trusted. Blocks `__cf_data`-forged pattern-shaped exports from acquiring programs and verified load IDs.

- **Bug Fixes**
  - Introduced `isTrustedPattern` (brand + shape) that also walks the `unsafe_originalPattern` chain to accept derivations/serialized copies of a branded pattern; builder stamps new patterns.
  - Switched `exportsCallback`, entry selection, and verified-load-id walks to root on `isTrustedPattern`, while still propagating to structural `isPattern` nodes within a trusted subtree.
  - Made verified-load-id propagation order-independent with per-node trust tracking (`Map`); added `pattern-provenance.test.ts`.

<sup>Written for commit 7ca4f2298b23af579e307a807caf2ecc42495e91. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3779?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/builder/factory.ts
+++ b/packages/runner/src/builder/factory.ts
@@ -16,7 +16,6 @@ import {
   FS,
   ID,
   ID_FIELD,
-  isPattern,
   NAME,
   schema as schemaIdentity,
   SELF,
@@ -48,7 +47,7 @@ import { cellConstructorFactory } from "../cell.ts";
 import { getEntityId } from "../create-ref.ts";
 import { getPatternEnvironment } from "./env.ts";
 import type { RuntimeProgram } from "../harness/types.ts";
-import { setPatternProgram } from "./pattern-metadata.ts";
+import { isTrustedPattern, setPatternProgram } from "./pattern-metadata.ts";
 import {
   FabricInstance,
   FabricPrimitive,
@@ -130,7 +129,10 @@ export const createBuilder = (options: CreateBuilderOptions = {}): {
   // instantiated. This way they get saved with a way to rehydrate them.
   const exportsCallback = (exports: Map<any, RuntimeProgram>) => {
     for (const [value, program] of exports) {
-      if (isPattern(value)) {
+      // `isTrustedPattern` (not the structural `isPattern`): only a value the
+      // trusted builder produced may acquire a rehydration program, so a
+      // `__cf_data`-forged pattern-shaped export cannot launder trust metadata.
+      if (isTrustedPattern(value)) {
         // Associate the program with the pattern via the side-table so it works
         // even when the exported pattern has been frozen by the loader.
         setPatternProgram(value, program);
--- a/packages/runner/src/builder/pattern-metadata.ts
+++ b/packages/runner/src/builder/pattern-metadata.ts
@@ -1,4 +1,5 @@
 import type { RuntimeProgram } from "../harness/types.ts";
+import { isPattern, type Pattern } from "./types.ts";
 
 /**
  * Side-table storage for pattern metadata that is associated *after* a pattern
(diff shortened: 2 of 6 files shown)


seefeldb:
Fixed in the latest commit: the two walks now track the trust level per visited node (`Map<value, trusted>`) and re-process a node on a trust upgrade (untrusted-visit-then-trusted-visit re-seeds), treating a trusted visit as final. Order-independent now.

seefeldb:
Addressed: `isTrustedPattern` now also confers trust on a value whose `unsafe_originalPattern` chain reaches a branded original — so a derivation/serialized copy of a trusted pattern (e.g. a nested subpattern selected as the entry export) keeps its program + verified-load id. Forgery-safe since `unsafe_originalPattern` is a module-private symbol authored code can't set. (The order-independence fix from the prior commit stays for nested patterns that have no original chain.)

PR #3780, opened 2026-06-01
feat(runner): env-backed CF_ESM_MODULE_LOADER toggle (+ ESM-path regression cross-check)

seefeldb:
## What

Adds an **env-backed toggle** for the `esmModuleLoader` experimental flag, following the `modernDataModel` / `persistentSchedulerState` ambient-config pattern (`set/get/resetEsmModuleLoaderConfig`, propagated from the `Runtime` constructor, reset on `dispose`) — but the default is seeded from the **`CF_ESM_MODULE_LOADER`** env var.

The production default stays **off** (env unset → `false`), so this is **not** the flag flip — it just makes the flag toggleable from the environment. An explicit `experimental.esmModuleLoader` option still wins; `undefined` keeps the env default. Env access is permission-guarded (no `--allow-env` → off).

This lets the whole suite run through the ESM path without threading the option through every `Runtime` construction:
```
CF_ESM_MODULE_LOADER=1 deno test --allow-env ...
```

## Regression cross-check (the reason for this)

The ESM loader is default-off, so CI's integration suites all run AMD — the ESM path + all its hardening (write-once exports, harden-frozen exports, graph-target verification, pattern provenance brand) had no broad regression signal. Using this toggle I ran the ESM-relevant suites flag-on and confirmed the path actually routes ESM (the AMD-compile-cache test fails under the flag because ESM bypasses the cache; a direct probe shows the effective flag flips with the env var):

- **15 runner pattern suites — 87 steps, 0 failures** (core, lift, handlers, if/else, schemas, async, dynamic, reduce, findindex, self, regressions, …).
- **Broad sweep — 278 steps, 0 failures** (patterns, binding, scope, CFC, implementation-identity, json-utils, engine-verified-functions, when/unless, esm-engine, provenance).
- **Only failure anywhere:** the AMD compilation-cache integration test, which the ESM path bypasses by design — not a behavior regression.

So real reactive patterns compile and run correctly through the ESM loader with all hardening active.

## Scope

`runtime.ts` wiring + a small ambient-config module + a unit test for the set/get/reset/env-default semantics. No CI matrix job yet (deferred per discussion).

---
## Summary by cubic
Adds an env-backed toggle for the experimental ESM module loader and resolves the flag in `Runtime`, so you can run the suite through the ESM path without wiring options. Production default stays off.

- **New Features**
  - Seed and resolve `experimental.esmModuleLoader` from `CF_ESM_MODULE_LOADER` (default off; `CF_ESM_MODULE_LOADER=1 deno test --allow-env`). Env access is permission-guarded.
  - Exposed on `runtime.experimental.esmModuleLoader` (false when off).
  - Precedence: an explicit option wins; `undefined` reads the env default.
  - Ambient `set/get/resetEsmModuleLoaderConfig`, wired in `Runtime` and reset on `dispose`.

- **Bug Fixes**
  - `setEsmModuleLoaderConfig(undefined)` now re-reads the env default to avoid stale overrides across runtimes.
  - `experimental-options` tests set `esmModuleLoader` explicitly for deterministic assertions under env-on/off.

<sup>Written for commit aceebe665e6db521fe76a3e1e8cae386836379b1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3780?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/runtime.ts
+++ b/packages/runner/src/runtime.ts
@@ -24,6 +24,11 @@ import {
   resetPersistentSchedulerStateConfig,
   setPersistentSchedulerStateConfig,
 } from "@commonfabric/memory/v2";
+import {
+  getEsmModuleLoaderConfig,
+  resetEsmModuleLoaderConfig,
+  setEsmModuleLoaderConfig,
+} from "./sandbox/esm-loader-config.ts";
 import { PatternEnvironment, setPatternEnvironment } from "./builder/env.ts";
 import { AsyncSemaphoreQueue, type QueueConfig } from "./queue.ts";
 import type {
@@ -354,6 +359,9 @@ export class Runtime {
     );
     this.experimental.persistentSchedulerState =
       getPersistentSchedulerStateConfig();
+    // Env-seeded default (CF_ESM_MODULE_LOADER) unless an explicit option is set.
+    setEsmModuleLoaderConfig(this.experimental.esmModuleLoader);
+    this.experimental.esmModuleLoader = getEsmModuleLoaderConfig();
 
     this.id = options.storageManager.id;
     this.apiUrl = new URL(options.apiUrl);
@@ -532,6 +540,7 @@ export class Runtime {
     // Reset experimental config to defaults.
     resetDataModelConfig();
     resetPersistentSchedulerStateConfig();
+    resetEsmModuleLoaderConfig();
 
     // Clear the current runtime reference
     // Removed setCurrentRuntime call - no longer using singleton pattern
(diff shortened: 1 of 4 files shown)


seefeldb:
Both addressed in the latest commit:
- **set(undefined) stale-inheritance**: `setEsmModuleLoaderConfig(undefined)` now falls back to the env-seeded default (`enabled ?? readEnvDefault()`) instead of leaving a prior value — a Runtime without the option always resolves to its own explicit option or `CF_ESM_MODULE_LOADER`, never a stale override.
- **env-dependent assertions**: the `experimental-options` tests now pass `esmModuleLoader: false` explicitly, so the shape assertions are deterministic regardless of the env. Verified the whole file passes under `CF_ESM_MODULE_LOADER=1`.

PR #3782, opened 2026-06-01
feat(runner): flip esmModuleLoader on by default (CT-1623 Phase 5)

seefeldb:
## CT-1623 Phase 5 — flip `esmModuleLoader` on by default

Makes the ESM module-record loader the **default** loader for pattern
compile/evaluate. `readEnvDefault()` now returns ON unless
`CF_ESM_MODULE_LOADER=0` (or `false`) opts back into the legacy AMD bundle
loader.

### Why now
The supporting phases are merged behind the (until-now default-off) flag:
- **Phase 2** — ESM emission + SES module-record loading (`compileToRecordGraph`
  / `evaluateRecordGraph`, full CF transformer pipeline, content-addressed
  records, source maps, live namespace bindings).
- **Phase 3** — per-module SES verifier (`verifyCompiledModuleBody` +
  `verifyModuleGraph`); ESM↔AMD differential parity oracle (#3801).
- **Phase 4** — content-addressed compilation **cell cache** (#3810): per-space
  source set `pattern:<identity>` (self-verifying) + compiled set
  `compileCache:<runtimeVersion>/<identity>` (CFC integrity, fail-closed), with
  warm-hit reuse that skips the TS compile / transformer pipeline.

So the default flip routes all pattern loading through the ESM path, now backed
by per-module caching + integrity rather than recompiling every time.

### Scope
- Flips the default only. **Legacy AMD bundle loader removal is a separate
  follow-up** (both paths still coexist; opt back into AMD with
  `CF_ESM_MODULE_LOADER=0`).
- The full suite passes through the ESM path (this branch has been the standing
  flag-on CI signal).

### Not addressed here
- **Performance Check is intentionally left for review** — this PR surfaces the
  real end-to-end perf impact of making the ESM loader the default, and that
  delta is being examined before any baseline decisions. No `NEW_PERF_BASELINE`
  overrides are applied.

### DIFF
--- a/packages/runner/src/sandbox/esm-loader-config.ts
+++ b/packages/runner/src/sandbox/esm-loader-config.ts
@@ -4,11 +4,11 @@
  * persistent-scheduler-state ambient flag: the `Runtime` constructor
  * propagates the explicit option here and reads the effective value back.
  *
- * Unlike that one, the default is seeded from the `CF_ESM_MODULE_LOADER`
- * environment variable, so the ESM loader can be exercised suite-wide (e.g. a
+ * The default is ON: the ESM module-record loader is the production default.
+ * Set `CF_ESM_MODULE_LOADER=0` (or `false`) to opt back into the legacy AMD
+ * bundle loader. The env var also lets the loader be selected suite-wide (e.g. a
  * regression cross-check) without threading the option through every `Runtime`
- * construction. The production default stays OFF (the env var is unset), so this
- * is NOT the flag flip — it only makes the flag toggleable from the environment.
+ * construction.
  */
 
 function readEnvDefault(): boolean {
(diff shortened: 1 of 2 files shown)


seefeldb:
## Canary signal #1 — known-red baseline (the `fn.src` flag-flip blocker)

The full-CI ESM run is green **except** two integration suites, failing **consistently** (failed twice, not flaky):

- `Pattern Integration Tests (2/4)` — `cfc-group-chat-demo` and `shared-profile`
- `CLI Integration Tests (fuse)`

**Root cause (confirmed, for the Pattern suite):** these exercise **CFC verified-source / authorship verification of trusted actions**. The pattern *renders fine* under ESM (the chat UI is present in the failure dump), but the trusted action (e.g. `CreateProfile`) **fails closed** so the profile never saves and the UI text (`#trusted-profile-status` = "Alice") never appears → 30s timeout.

`cfc/implementation-identity.ts:84` derives verified-source identity from `parseVerifiedSourceLocation((implementation).src)`. Under SES-ESM, `//# sourceURL` is stripped so `fn.src` doesn't resolve → `parseVerifiedSourceLocation` returns `undefined` → identity can't be established → fail-closed. (The CFC *unit* identity tests pass under ESM only because they feed a synthetic `src: "/main.tsx:4:12"`; they never exercise real `fn.src`.)

This is **the documented "remaining item before the flag flip"** — full source-location fidelity under the ESM loader (`module-record-compiler.ts` notes it requires SES-isolate-level source-map integration). It is **not** a regression from the hardening work — `verifiedLoadId`/provenance resolve correctly; only the source *location* is missing.

**`fuse`**: a 480s suite-level hang (less detail); likely the same source-location/CFC class or a separate ESM CLI hang — to be root-caused separately.

### Interpreting this canary going forward
These **2 suites are the known-red baseline** (= the `fn.src` blocker). After rebasing onto main, **any red beyond these two = a new regression**. They clear only when source-location fidelity lands (Phase 5).

PR #3792, opened 2026-06-01
refactor(data-model): reparent FabricRegExp to FabricPrimitive

danfuzz:
Reparent `FabricRegExp` from `FabricInstance` (via `FabricNativeWrapper`)
to `FabricPrimitive`: it is a leaf type w.r.t. references and is reasonably
conceived of as stateless, even though a JS `RegExp` carries mutable
internal state (`lastIndex`).

## What changes

- Move `FabricRegExp.ts` from `fabric-instances/` to `fabric-primitives/`;
  repoint the two barrels.
- Store only an internally-constructed `RegExp` in hard-private `#value`
  (eager syntax validation + cheap cloning) plus `#flavor`. Expose
  `get source()` / `get flags()` (immutable strings off `#value`),
  `get flavor()`, and `get value(): RegExp` which returns a fresh clone
  (`new RegExp(this.#value)`) on every call -- the stored `RegExp` is never
  handed out un-cloned, so no mutable state leaks (mirrors why `FabricBytes`
  copies on `slice()`). `Object.freeze(this)` in the constructor.
- Drop the `FabricNativeWrapper` machinery (DECONSTRUCT/RECONSTRUCT/
  DEEP_FREEZE/IS_DEEP_FROZEN/toNative/wrappedValue) -- those were about
  being a native wrapper; `get value()` is now the sole native bridge
  (no generic `nativeFromFabricValue` arm).
- Add a dedicated json-wire `RegExpHandler` (encode source/flags/flavor,
  decode via `FabricRegExp.fromState`); register it; drop the now-defunct
  class-registry registration in `JsonEncodingContext`.
- Add the `FabricRegExp` primitive arm to the four sibling-dispatch sites:
  `native-type-tags.ts` (new tag + constructor case), `value-clone.ts`
  (immutable pass-through), `value-hash.ts` (new primitive tag byte
  TAG_REGEXP 0x2b), and the `fabric-value-modern.ts` conversion.
- Move the test file to `test/fabric-primitives/`; recompute the
  cloneIfNecessary matrix entry to the primitive group and the value-debug
  kind string to `FabricPrimitive (FabricRegExp)`; drop the dropped-protocol
  test cases.
- Update two stale runner comments that listed `FabricRegExp` as a
  `FabricInstance`.

## Behavior preservation

- Full data-model `deno task test` green at **29/1824/0**.
- The `FabricRegExp` round-trip (json-wire encode/decode) and the
  native-`RegExp` == `FabricRegExp` hash equivalence both pass.
- Net diff: 20 files, +498/-505; branch merge-base is the current
  `origin/main` tip (no rebase needed).

## Intended behavior changes (not regressions)

These are deliberate consequences of the agreed shape, pre-accepted by Dan:

1. **Regexp content hashes change** due to the new primitive tag byte
   `TAG_REGEXP 0x2b`. Dan pre-accepted -- the class is not yet used in the
   system, and no consumer relied on the old hash.
2. **No native-unwrap** of a fabric `RegExp` (Dan ruled DROP); `get value()`
   is the sole native bridge.
3. `get value()` named to match the `FabricEpoch*` getters and returns a
   **fresh clone per call** (no un-cloned stored `RegExp` is ever exposed).

## Spec divergence (deferred to spec-writer sync)

This changes `FabricRegExp` hashing from `TAG_INSTANCE` to a dedicated
primitive tag byte (`TAG_REGEXP` = `0x2b`), per Dan's 2026-06-01 guidance,
and intentionally changes regexp content hashes (the class is not yet used
in the system). The formal spec
`docs/specs/space-model-formal-spec/2-hash-byte-format.md` still codifies
the old instance-hashed behavior in:

- §7.12 (`FabricRegExp(/abc/gi)` worked example -- "is a `FabricInstance`
  and is hashed via `TAG_INSTANCE`"),
- §4.13 (the `TAG_INSTANCE` type-tag list including `"RegExp@1"`),
- the §398 "types with dedicated tags" note (RegExp should join
  Bytes/EpochNsec/EpochDays/Hash).

These will be reconciled in a follow-up spec-writer sync (tracked by
foreman-rail). The `0x2b` byte is the next free value after
`TAG_SYMBOL 0x2a` per the spec's primitive allocation.

Gates: data-model `deno check`/`fmt`/`lint`/full `deno task test`
(29/1824/0), `deno check` over runner + api -- all green locally.

Co-Authored-By: coder-cedar (Claude Opus 4.8 (1M context)) <noreply@anthropic.com>


packages/data-model/src/fabric-primitives/FabricRegExp.ts
```diff
@@ -0,0 +1,115 @@
+import { FabricPrimitive } from "../interface.ts";
+import { TAGS } from "../fabric-type-tags.ts";
+
+/**
+ * Immutable regular-expression value in the fabric type system. Extends
```
danfuzz: No need for exposition that is just a restatement of the code. As in, it is clear that it `extends FabricPrimitive`, and in the context of this project we know that `FabricPrimitive` is a type of `FabricValue`.

Relatedly, while the fact about `FabricBytes` was useful as a driver for development, to a consumer of this class reading the docs, it's just noise. (This also applies to the getter.)

packages/data-model/src/fabric-primitives/FabricRegExp.ts
```diff
@@ -0,0 +1,115 @@
...
+  constructor(regex: RegExp, flavor: string = "es2025") {
+    super();
+    rejectExtraRegExpProperties(regex);
+    this.#value = new RegExp(regex.source, regex.flags);
+    this.#flavor = flavor;
+    Object.freeze(this);
+  }
+
+  /** The wire-format type tag (`RegExp@1`). */
+  get typeTag(): string {
```
danfuzz: This is a leftover from being a `FabricInstance`. `FabricPrimitive`s don't have instance type tags.

packages/data-model/src/fabric-primitives/FabricRegExp.ts
```diff
@@ -0,0 +1,115 @@
...
+   * the pattern eagerly); the input `RegExp` object itself is not aliased. A
+   * `RegExp` with extra enumerable own properties is rejected (the built-in
+   * `.lastIndex` is non-enumerable, so `Object.keys()` only sees user-added
+   * properties).
+   *
+   * @param regex - The native `RegExp` to wrap (only its `source`/`flags` are
+   *   retained, via a fresh internal `RegExp`).
+   * @param flavor - Regex flavor/dialect identifier (default `"es2025"`).
+   */
+  constructor(regex: RegExp, flavor: string = "es2025") {
```
danfuzz: Suggest making the constructor effectively polymorphic:

* `new FabricRegExp(regex: RegExp)` which implies flavor `es2025`.
* `new FabricRegExp(flavor: string, source: string, flags: string)` which makes everything explicit.

And then:

* Always store the source, flags, and flavor.
* Only when `flavor === "es2025"` proactively construct the private `#value`.
* For now, `throw` in `get value()` when `flavor !== "es2025"`.

This is going to set the class up for a future where it really can represent different regex syntaxes.

packages/data-model/src/fabric-value-modern.ts
```diff
@@ -90,14 +91,13 @@ export function shallowFabricFromNativeValueModern(
...
-      // `RegExp` instances are wrapped in `FabricRegExp`. Extra enumerable
-      // properties cause rejection ("death before confusion"). The
-      // `rejectExtraProperties()` check is done inside `FabricRegExp`'s
-      // `[DECONSTRUCT]`, but we also reject eagerly here at conversion time.
+      // `RegExp` instances are converted to `FabricRegExp`. Extra enumerable
+      // properties cause rejection ("death before confusion") -- the
+      // constructor performs the same check, but we also reject eagerly here at
+      // conversion time. `FabricRegExp` self-freezes in its constructor
+      // (`FabricPrimitive` contract).
       rejectExtraProperties(value as object, "RegExp");
```
danfuzz: No need for the redundant `rejectExtraProperties`. it's literally always going to be the same test. It was appropriate before because it _wasn't_ happening in the constructor.

packages/data-model/src/value-hash.ts
```diff
@@ -311,6 +314,20 @@ function feedObjectValue(
       return;
     }
 
+    case NATIVE_TAGS.FabricRegExp: {
+      // `FabricRegExp` is a `FabricPrimitive` with a dedicated primitive tag
```
danfuzz: Again, no need for redundant exposition. Note that the other parts of this `switch` are a-okay with very little (mostly no) comments.

packages/runner/src/cell.ts
```diff
@@ -2280,8 +2280,8 @@ export function recursivelyAddIDIfNeeded<T>(
     return value;
   }
 
-  // `FabricInstance` values (`FabricError`, `FabricMap`, `FabricSet`,
-  // `FabricRegExp`) are immutable wrappers with class-defined identity.
+  // `FabricInstance` values (`FabricError`, `FabricMap`, `FabricSet`)
```
danfuzz: This comment doesn't need to recapitulate the class hierarchy at all. That is, removing the whole parenthetical is a good idea. Likewise in `data-updating.ts`.

packages/data-model/src/fabric-primitives/FabricRegExp.ts
```diff
@@ -74,30 +97,32 @@ export class FabricRegExp extends FabricPrimitive {
...
   /**
    * Reconstructs a `FabricRegExp` from its essential state
-   * (`{ source, flags, flavor }`). Reconstruction goes through the constructor,
-   * so the eager syntax-validation path covers decoding too.
+   * (`{ source, flags, flavor }`).
    *
    * @param state - The essential state, with optional `source`, `flags`, and
    *   `flavor` string fields.
    */
   static fromState(state: Record<string, unknown>): FabricRegExp {
```
danfuzz: Since the constructor takes explicit state arguements now, there's no need for this method.

packages/data-model/test/fabric-primitives/FabricRegExp.test.ts
```diff
@@ -28,20 +28,26 @@ describe("FabricRegExp", () => {
 
   describe("constructor()", () => {
     it("produces an always-frozen instance", () => {
-      const re = new FabricRegExp(/abc/);
-      expect(Object.isFrozen(re)).toBe(true);
+      expect(Object.isFrozen(new FabricRegExp(/abc/))).toBe(true);
     });
 
-    it("retains the `source` and `flags` of the original `RegExp`", () => {
+    it("from a `RegExp`: retains `source`/`flags` and implies `es2025`", () => {
```
danfuzz: This test label does not adhere to the rules for same. Note: You're allowed to have a sub-`describe()` to provide a context for the ultimate `it()` sentence.

packages/data-model/test/fabric-primitives/FabricRegExp.test.ts
```diff
@@ -28,20 +28,26 @@ describe("FabricRegExp", () => {
...
       const re = new FabricRegExp(/test/i);
       expect(re.source).toBe("test");
       expect(re.flags).toBe("i");
+      expect(re.flavor).toBe("es2025");
+    });
+
+    it("from explicit `flavor`/`source`/`flags`: retains all three", () => {
```
danfuzz: This test label does not adhere to the rules for same.

packages/data-model/src/fabric-primitives/FabricRegExp.ts
danfuzz: Done in `84ac5227c`: trimmed the class doc (and the `value` getter doc) to drop the type-hierarchy restatement and the `FabricBytes`-as-reference framing.

packages/data-model/src/fabric-primitives/FabricRegExp.ts
danfuzz: Done in `84ac5227c`: dropped the `typeTag` getter (`FabricPrimitive`s carry no instance type tag).

packages/data-model/src/fabric-primitives/FabricRegExp.ts
danfuzz: Done in `84ac5227c`: polymorphic constructor `new FabricRegExp(regex)` / `new FabricRegExp(flavor, source, flags)`; always store `source`/`flags`/`flavor` strings, build `#value: RegExp` only for `es2025`, and `get value()` throws for other flavors.

packages/data-model/src/fabric-value-modern.ts
danfuzz: Done in `84ac5227c`: removed the redundant `rejectExtraProperties` call (the constructor now performs that same check).

packages/data-model/src/value-hash.ts
danfuzz: Done in `84ac5227c`: trimmed the comment to match the terse style of the sibling `switch` arms.

packages/runner/src/cell.ts
danfuzz: Done in `84ac5227c`: removed the class-hierarchy parenthetical from the comment in both `cell.ts` and `data-updating.ts`.

packages/data-model/src/fabric-primitives/FabricRegExp.ts
danfuzz: Done in `84ac5227c`: removed `fromState()`; the json-wire `RegExpHandler` decode path now reconstructs directly via `new FabricRegExp(flavor, source, flags)`. Decode round-trip still passes.

packages/data-model/test/fabric-primitives/FabricRegExp.test.ts
danfuzz: Done in `84ac5227c`: restructured into a `given a RegExp` sub-`describe()` so the `it()` reads as a third-person sentence ("retains the source and flags and implies the es2025 flavor").

packages/data-model/test/fabric-primitives/FabricRegExp.test.ts
danfuzz: Done in `84ac5227c`: moved under a `given explicit flavor/source/flags` sub-`describe()` so the `it()` reads "retains all three". Also swept the rest of the file's labels for the same rule.

PR #3794, opened 2026-06-01
Make local dev startup verify readiness

bfollington:
Summary:
- Add preflight checks for deno and curl before launching local dev servers.
- Wait for shell and toolshed to return HTTP 200 before printing startup success.
- Clean up partial starts and show recent logs when startup fails.
- Document LOCAL_DEV_STARTUP_TIMEOUT for onboarding/debugging.

Validation:
- bash -n scripts/start-local-dev.sh
- git diff --check -- scripts/start-local-dev.sh docs/development/LOCAL_DEV_SERVERS.md
- env PATH=/usr/bin:/bin:/usr/sbin ./scripts/start-local-dev.sh --shell-port 65173 --toolshed-port 65174
- env LOCAL_DEV_STARTUP_TIMEOUT=90 ./scripts/start-local-dev.sh --shell-port 65173 --toolshed-port 65174
- ./scripts/stop-local-dev.sh --shell-port 65173 --toolshed-port 65174
- pre-commit hook: deno fmt, deno lint

---
## Summary by cubic
Verify local dev servers are actually ready before reporting success, with clearer failures and automatic cleanup. The startup script now validates prerequisites, waits for HTTP 200 from `shell` and `toolshed`, and tails logs on errors.

- **New Features**
  - Preflight checks for `deno` and `curl` with install hints.
  - Readiness wait with early-exit detection: HTTP 200 from `shell` (/) and `toolshed` (/_health); configurable via `LOCAL_DEV_STARTUP_TIMEOUT` (default 120s).
  - Robust cleanup on failure: kills started processes and any port listeners.
  - Better debugging: tails last 40 log lines; standardized paths at `packages/shell/local-dev-shell.log`, `packages/toolshed/local-dev-toolshed.log`, and `packages/background-charm-service/local-dev-bg.log`.
  - Background service starts directly with PID tracking and verifies it stays running.

- **Migration**
  - Ensure `deno` and `curl` are installed (e.g., run `mise install`).
  - Optionally set `LOCAL_DEV_STARTUP_TIMEOUT` to adjust readiness wait time.

<sup>Written for commit f5fc8a85a4b6c04c54dcbfa877173337a01d629c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3794?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/start-local-dev.sh
+++ b/scripts/start-local-dev.sh
@@ -200,9 +200,13 @@ ensure_process_running() {
     local name=$1
     local pid=$2
     local log_file=$3
+    local process_state
 
     if kill -0 "$pid" 2>/dev/null; then
-        return
+        process_state=$(ps -p "$pid" -o stat= 2>/dev/null)
+        if [[ "$process_state" != *Z* ]]; then
+            return
+        fi
     fi
 
     wait "$pid" 2>/dev/null
@@ -218,14 +222,21 @@ wait_for_http() {
     local url=$2
     local pid=$3
     local log_file=$4
+    local deadline=$((SECONDS + LOCAL_DEV_STARTUP_TIMEOUT))
+    local remaining
     local status
 
     echo "Waiting for $name at $url..."
-    for _ in $(seq 1 "$LOCAL_DEV_STARTUP_TIMEOUT"); do
+    while (( SECONDS < deadline )); do
         ensure_process_running "$name" "$pid" "$log_file"
 
+        remaining=$((deadline - SECONDS))
+        if (( remaining > 2 )); then
+            remaining=2
+        fi
+
         status=$(
-            curl -s -o /dev/null -w "%{http_code}" --max-time 2 \
+            curl -s -o /dev/null -w "%{http_code}" --max-time "$remaining" \
                 "$url" 2>/dev/null
         )
         if [[ "$status" == "200" ]]; then
@@ -233,7 +244,9 @@ wait_for_http() {
             return
         fi
 
-        sleep 1
+        if (( SECONDS < deadline )); then
+            sleep 1
+        fi
     done
 
     fail_startup \
@@ -244,8 +257,6 @@ wait_for_http() {
 cd "$SCRIPT_DIR/../packages/shell"
 TOOLSHED_PORT="$TOOLSHED_PORT" deno task dev-local > "$SHELL_LOG" 2>&1 &
 SHELL_PID=$!
-
-wait_for_http "shell" "http://localhost:$SHELL_PORT" "$SHELL_PID" "$SHELL_LOG"
 
 # Start toolshed dev server in background
 # We pass --port= as CLI arg because deno --watch doesn't pass env vars to subprocess
@@ -267,6 +278,8 @@ SHELL_URL="http://localhost:$SHELL_PORT" \
         --env-file=.env index.ts --port="$TOOLSHED_PORT" \
         > "$TOOLSHED_LOG" 2>&1 &
 TOOLSHED_PID=$!
+
+wait_for_http "shell" "http://localhost:$SHELL_PORT" "$SHELL_PID" "$SHELL_LOG"
 
 # # Function to cleanup background processes
 # cleanup() {
--- a/scripts/start-local-dev.sh
+++ b/scripts/start-local-dev.sh
@@ -7,6 +7,19 @@ cd "$SCRIPT_DIR/.."
 source "$SCRIPT_DIR/common/port-utils.sh"
 read_base_ports
 
+require_command() {
+    local command_name=$1
+    local install_hint=$2
+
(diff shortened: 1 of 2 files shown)


scripts/start-local-dev.sh
```diff
@@ -199,26 +317,17 @@ if [[ "$BG_UPDATER" == "true" ]]; then
...
 
     # Start the background service directly (not via deno task, for reliable PID tracking)
     cd "$SCRIPT_DIR/../packages/background-charm-service"
     OPERATOR_PASS="implicit trust" API_URL="http://localhost:$TOOLSHED_PORT" \
         deno run -A --unstable-worker-options src/main.ts \
-        > "$SCRIPT_DIR/../packages/background-charm-service/local-dev-bg.log" 2>&1 &
+        > "$BG_LOG" 2>&1 &
     BG_PID=$!
     cd "$SCRIPT_DIR/.."
+    sleep 2
```
Hixie, line 342: this sleep isn't necessary, ensure_process_running uses `wait` to wait for the PID.

Hixie (approved):
LGTM

bfollington:
## Code Review

### Overview
Replaces fixed `sleep 2`/`sleep 3` startup heuristics with real readiness checks: preflight `deno`/`curl` validation, polls `shell` (`/`) and `toolshed` (`/_health`) for HTTP 200, cleans up partial starts on failure, dumps recent logs, and tracks the bg-service PID directly. A clear, well-factored improvement — the helper decomposition (`require_command`, `wait_for_http`, `ensure_process_running`, `cleanup_started_processes`, `fail_startup`) is clean and reads well.

### Correctness

- **Startup is now strictly sequential (main thing to verify).** Previously shell and toolshed started concurrently and the script just slept. Now `wait_for_http "shell"` blocks until shell returns 200 *before toolshed is even launched*. Two consequences:
  - **Deadlock risk** if the shell dev server's `/` doesn't return 200 until it can reach toolshed (shell is started with `TOOLSHED_PORT` in its env). If such a dependency exists, startup hangs the full timeout then fails. Worth confirming the shell root serves 200 standalone — a Vite-style dev server normally does, so likely fine, but the new hard ordering makes it load-bearing.
  - **Added latency**: toolshed startup is now delayed by shell's full ready time rather than overlapping with it. Not a bug, but happy-path startup is slower than before.

- **`ensure_process_running` can be slow to detect a dead process.** A backgrounded child that has exited but not yet been `wait`ed becomes a zombie, and `kill -0` still succeeds on a zombie PID. So inside `wait_for_http`, the `kill -0` short-circuit reports "running" for an exited-but-unreaped process, and detection falls through to the curl-timeout path instead of the intended early "exited before ready" message. It still fails eventually, just not promptly with the nicer error. Minor.

- **Effective timeout is larger than configured.** Each loop iteration is `curl --max-time 2` + `sleep 1`, so a hung server makes `LOCAL_DEV_STARTUP_TIMEOUT=120` take up to ~360s wall-clock. The doc says "timeout in seconds" — consider noting it's approximate, or basing the loop on elapsed time rather than iteration count.

### Minor / nits

- `fail_startup` dumps shell and toolshed logs but not `$BG_LOG`. Background-service failures go through `ensure_process_running`, which does show the bg log, so this is consistent — just noting the asymmetry.
- `exit_status` capture in `ensure_process_running` is correct: `wait` and `local exit_status=$?` are separate statements, so `$?` reflects `wait`. 👍
- Script doesn't use `set -euo pipefail`, but that matches existing style and the new code handles statuses explicitly — no change needed.

### Security
No concerns. `OPERATOR_PASS="implicit trust"` is pre-existing local-dev-only config, unchanged.

### Recommendation
Approve once it's confirmed the shell `/` endpoint returns 200 without toolshed running (the new sequential ordering depends on it). The zombie-detection and approximate-timeout items are optional polish.

bfollington:
Addressed the review feedback in f5fc8a85a:

- shell and toolshed are now launched before waiting on either readiness endpoint, preserving the old startup overlap and avoiding a shell-before-toolshed dependency risk.
- readiness timeout is now based on elapsed time instead of loop count.
- process checks now detect zombie/exited children promptly before falling back to the timeout path.

Local validation:
- bash -n scripts/start-local-dev.sh
- git diff --check -- scripts/start-local-dev.sh docs/development/LOCAL_DEV_SERVERS.md
- missing-deno preflight simulation
- high-port local-dev smoke on 65173/65174, followed by stop/port-clear check

PR #3799, opened 2026-06-02
docs(debugging): cold-load & handler-created-piece gotchas ([UI] VNode + 1-hop lift)

bfollington:
Adds a debugging gotcha doc: **a navigable piece's top-level `[UI]` must be a DOM-rooted VNode authored inline in the pattern body** — not a bare child-pattern instance, and not the return value of a plain helper that builds the VNode.

### Symptom
A pattern type-checks and deploys, its `[NAME]` shows in the shell title, but its **body renders empty when opened cold** (slug or raw id URL). It often renders fine when navigated to from within an already-loaded app, so it looks like a navigation bug when it's a render bug. No console error.

### Why
The helper/instance form re-materializes for the **root/deployed** piece (its UI is materialized at deploy and served from storage) but **not for a cold-loaded sub-piece**, which re-runs the pattern and produces an empty UI. Inline DOM-rooted JSX is required; child patterns render fine when nested inside DOM.

Surfaced while debugging the Loom mobile pattern (tab sub-screens cold-loaded empty). Adds the gotcha doc + a row in the debugging error-reference matrix.

---
## Summary by cubic
Corrects two debugging docs: cold-load empties come from declaring UI under `ui:` instead of the `[UI]` symbol, and the “lift() only resolves at 1 hop” hypothesis is refuted with tests showing both 1- and 2-hop lifts work. The quick-reference matrix is updated.

- **New Features**
  - Rewrote the `[UI]` gotcha: the Output type must declare `[UI]: VNode`; the value form (inline JSX, child instance, or helper) is fine. Adds repro via `cf check --show-transformed` and fixes the README row.
  - Replaced the 1-hop lift rule with a negative result doc and links to verified causes. Added tests proving lifts resolve at both depths in push/pull modes (`packages/runner/test/navigate-handler-lift-hops.test.ts`, `packages/generated-patterns/integration/patterns/handler-created-piece-lift-hops.{pattern,test}.ts`).

<sup>Written for commit 76f1b6bb95f148c4011bd24a71a550df107d3ffb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3799?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bfollington:
Reviewed for documentation accuracy.

I would not merge this as written yet. The `piece-ui-must-be-vnode` gotcha looks over-scoped:

- The renderer intentionally follows `[UI]` chains, and existing composition docs say pattern results placed in vdom have their `[UI]` extracted.
- `packages/patterns/counter/counter.tsx` has a wrapper pattern returning `[UI]: <Counter />`, with an inline comment saying patterns can be rendered as JSX elements directly.
- A scratch pattern with `interface Output { [UI]: VNode }` and either `[UI]: frame(label)`, `[UI]: ui`, or `[UI]: <Child label={label} />` passes `mise exec -- deno task cf check ... --show-transformed --no-run`; the transformed result schema includes `$UI: { $ref: "https://commonfabric.org/schemas/vnode.json" }`. The helper-call form is lowered to a `lift`, but it is not categorically schema-invisible.
- The suspicious failure shape seems more specific: `[UI]` may be missing from the output type/result schema, or there is a narrower helper/rehydration bug. For example, `packages/patterns/image-analysis.tsx` returns `[UI]: ui`, but its output type declares only `ui: VNode`, and the transformed result schema omits `$UI`.

Suggested fix: rewrite the first doc to describe the exact failing shape/repro rather than banning all helper-returned VNodes or root pattern JSX. At minimum, call out that navigable sub-pattern outputs must declare `[UI]: VNode` in the Output type and show the transformed/result schema expectation.

For `lift-in-handler-created-piece`, I could not find local regression coverage proving the "1 hop" rule. If this is observed behavior, the doc should say "current limitation/observed failure" and include a minimal repro or test link, because the statement currently reads like a general runtime law.

Verification done:
- `mise exec -- deno task cf check packages/patterns/image-analysis.tsx --show-transformed --no-run`
- `mise exec -- deno task cf check packages/patterns/counter/counter.tsx --show-transformed --no-run`
- scratch transformer checks for helper-root, local-root, and child-pattern-root `[UI]`
- `mise exec -- deno test -A --filter "plain callables that lack pattern-factory shape" packages/ts-transformers/test/pipeline-regressions.test.ts`

Note: `packages/runtime-client/integration/client.test.ts` was not useful locally because it requires `http://localhost:8000`; my direct run failed only on that missing API prerequisite. The PR workflow rerun also still had some jobs in progress when I checked.

PR #3829, opened 2026-06-03
feat(skills): add cf-review code-review skill

bfollington:
## What

Adds `/cf-review`, a repo-local code-review skill for Common Fabric — written as a **map & values, not a recipe** (per `docs/development/skill-authoring.md`, #3836).

- **Auto-detects target**: local branch diff vs `main`, or a GitHub PR (`/cf-review <PR#>`); picks **detail vs. scope-and-theory mode** by diff size (large diffs → a theory of intent + fan-out subagents, not file-by-file).
- **Report-first**, then offers to post a **self-signed** PR review (agent + model, on behalf of the human — never impersonating).
- Three buckets — 🔴 Blocking / 🟡 Improvements / ⚪ Nits — reusing the severity taxonomy from `docs/common/ai/pattern-critique-guide.md`.

## Why

Review here has repo-specific concerns beyond generic craft, framed as **high-velocity-team failure modes** (not "the LLM's fault"):

- **Coherence ripple** — runtime-semantic changes must keep docs/examples in sync so a searcher isn't misled (bounded to the diff, not a repo audit).
- **Anti-duplication** — don't re-fork core machinery (`hash` / `serialize` / `clone` / identity); the table points to canonical **modules**, not rot-prone symbol names.
- **Framework-fit** — flag code fighting the transformer/reactive model, plus the doc gap that caused it.
- **Test rigor** — for each touched test, what principle does it guard?

## Notes

- Canonical source `skills/cf-review/SKILL.md`, mirrored via symlinks to `.claude/skills/` and `.agents/skills/`.
- Passes `deno fmt` + `deno lint`; all referenced paths/modules verified to exist. Reviewed by the team (thanks @mpsalisbury @danfuzz @Hixie) and dogfooded on itself in an isolated subagent.


skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,345 @@
...
+
+> When someone — human or LLM — searches for the words they've heard and lands
+> on a doc, comment, or example, they must not be **misled** and trip over
+> something that no longer matches how the runtime actually works.
+
+That is the bar. It is **not** "every file uses identical vocabulary." Line-by-
+line vocabulary unification is a slow, opportunistic, Boy-Scout activity — nudge
+nearby drift with small edits, propose big sweeps as follow-ups, never block a
+PR on it. Block only when a change leaves an authoritative source actively
+wrong.
```
bfollington: Probably superfluous negative prompting.

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,345 @@
...
+## What this review is (and is not)
+
+- It is **changeset-scoped**: review the diff and its immediate ripple, not the
+  whole repo. Do not start an open-ended repo audit. (That is a different,
+  heavier job — the coherence-audit. This is fast.)
+- It is **loud on what matters** and **quiet on what doesn't**. Flag bugs,
+  regressions, and broken principles unmistakably. Keep nits to a short,
+  clearly-optional list — or omit them when there are real issues to focus on.
+- It is **report-first**. Produce the ranked report. Only touch the PR when the
+  invoker passes `--comment`, and only after showing the report.
```
bfollington: I actually prefer it to comment on the PR but sometimes agents refuse to impersonate you without explicit consent (good) and sometimes the report is not worth posting.

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,345 @@
...
+wrong.
+
+## Why this repo is easy to get wrong
+
+Programs ("patterns") are authored as TypeScript, **transformed** by a custom
+TS-transformer toolchain (CTS) into plain JS that **builds a reactive graph**
+evaluated under live subscriptions. So the source you read is two abstraction
+levels above what runs. The principles are logical once read end-to-end, but
+miss a piece and the behavior looks inexplicable. Two consequences drive this
+review:
```
bfollington: > So the source you read is two abstraction levels above what runs

No, "the changeset may span many abstraction levels, implications are not obvious in isolation"

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,345 @@
...
+- Logic errors, wrong conditionals, off-by-one, unhandled empty/error states.
+- **Regression smell in tests**: was an existing test _weakened or retrofitted_
+  to make new (possibly wrong) behavior pass? Diff the test changes against the
+  behavior change and check they agree on purpose. Tests edited to assert a bug
+  are Blocking.
+- **Reactivity correctness** (this repo): reactive self-feedback loops (a `cf-*`
+  control already bound via `$value`/`$checked` whose handler writes the same
+  cell back), `.get()` on computed/lift results, `new Writable(reactiveValue)`,
+  nested `Writable`. Defer to `docs/common/ai/pattern-critique-guide.md` for the
+  full pattern ruleset — don't restate it; cite it.
```
bfollington: Don't uplift specific fields like this into a skill, they can go stale, just point to the relevant docs.

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,345 @@
...
+- **Determinism / SES**: direct `Date.now()`, `Math.random()`, or authored
+  timers in pattern code; `safeDateNow()`/`nonPrivateRandom()` used inside a
+  re-running `computed()`/`lift()`.
```
bfollington: Ditto, this should just point to docs.

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,345 @@
...
+| identity / DID / keypairs                 | `@commonfabric/identity`                                                                                                   | generate DIDs or keys ad hoc                                                                                                                                                                   |
+| LEB128 varint encoding                    | `@commonfabric/leb128`                                                                                                     | hand-roll varint encode/decode                                                                                                                                                                 |
+| merkle references                         | `refer()` in `packages/memory/consumer.ts`                                                                                 | re-derive reference hashing                                                                                                                                                                    |
```
bfollington: I do not know personally if this is accurate or complete, would appreciate input!

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,345 @@
...
+- If the change **violates a documented invariant** (transformer target-language
+  boundary, module-graph cleanliness, "make invalid states unrepresentable"),
+  flag it **loudly** with the doc reference. A silent principle break is
+  Blocking even if the code "works."
+- **Don't boil the ocean.** Fix what the diff touches. Propose wider doc sweeps
+  as follow-ups; never balloon the PR chasing every minor mention.
+
+### 3. Anti-duplication
+
+LLM-generated code silently forks core machinery. In a few domains this is a
```
bfollington: > "LLM-generated code silently forks core machinery."

We aim to move fast but avoid duplication of core machinery.

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,345 @@
...
+levels above what runs. The principles are logical once read end-to-end, but
+miss a piece and the behavior looks inexplicable. Two consequences drive this
+review:
+
+1. **LLMs fight the framework** when they can't find the idiom — they reach for
+   try/catch, singletons, manual subscriptions, `async/await` in handlers, or
+   `/// <cf-disable-transform />`. See Dimension 4.
+2. **LLMs duplicate core machinery** — re-deriving hashing, serialization, or
+   cloning instead of reusing the one canonical home. These are the subtle,
+   expensive footguns. See Dimension 3.
```
bfollington: Reframe not as criticisms of LLMs but of failure modes affect high-bandwidth teams - it's not the always the AI's fault.

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,345 @@
...
+`packages/toolshed/lib/sha2.ts` vs inline `crypto.subtle.digest` calls) — so
+steer new code to the canonical home; **do not cargo-cult a nearby fork**. When
+the diff adds or edits any of these, scan it explicitly:
+
+```bash
+# new hash/serialize/clone defs, or raw crypto / JSON-clone, introduced by the diff
+git diff "$BASE" -- '*.ts' | grep -nE \
+  '^\+.*((export )?(async )?(function|const) (hash|sha256|serialize|deserialize|clone|deepClone)\b|crypto\.subtle\.digest|JSON\.parse\(JSON\.stringify)'
+# PR mode: gh pr diff <N> | grep -nE '…the same pattern…'
+```
```
bfollington: This seems like bloat, or overly specific.

skills/cf-review/SKILL.md
danfuzz: I finesse this in `danfuzz-land` by asking the agents to sign comments "as themselves" (which in context means both a team "agent ID" but also the model and version.

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,389 @@
...
+| ----------------------------------------- | -------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| SHA-256 bytes/hex (content addressing)    | `@commonfabric/content-hash` (`sha256`, `createHasher`)                                                                    | hand-roll, call `crypto.subtle.digest("SHA-256", …)` directly, or import `@noble/hashes` / `hash-wasm` / `node:crypto` / `@std/crypto`                                                         |
+| hash of a fabric value or schema          | `@commonfabric/data-model` → `hashOf` / `hashStringOf` / `taggedHashStringOf` (`value-hash`), `hashSchema` (`schema-hash`) | re-derive value or schema hashing                                                                                                                                                              |
```
danfuzz: Also do not: Use `JSON.stringify()` or a similar tactic to "simulate" hashing.

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,389 @@
...
+| clone a fabric value                      | `@commonfabric/data-model` `value-clone` (`cloneIfNecessary`, `cloneForMutation`, `shallowMutableClone`)                   | `structuredClone()` or `JSON.parse(JSON.stringify(...))` on fabric/cell data — it drops cell links and dies on circular `$UI` trees                                                            |
```
danfuzz: It also erases the type identity and most or all of the contents of non-plain objects.

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,389 @@
...
+| cell ↔ link conversion                    | `convertCellsToLinks` in `packages/runner/src/cell.ts`                                                                     | re-implement link conversion (note: `traverseAndSerialize`/`traverseAndCellify` in `packages/runner/src/builtins/llm-dialog.ts` are private and a known smell — see CT-1205 — don't copy them) |
+| identity / DID / keypairs                 | `@commonfabric/identity`                                                                                                   | generate DIDs or keys ad hoc                                                                                                                                                                   |
+| LEB128 varint encoding                    | `@commonfabric/leb128`                                                                                                     | hand-roll varint encode/decode                                                                                                                                                                 |
```
danfuzz: I'd say "LEB128 or similar variable-length integer encoding."

skills/cf-review/SKILL.md
danfuzz: It's not. `refer()` _used_ to be a directly `import`able function which produced content hash objects based on a Merkle-tree representation. However, that was all removed with the graduation of the `modern-hash` flag. At this point, while we _do_ have content hashes (`value-hash` and `schema-hash` in `data-model`), we _don't_ have a classic Merkle-tree hash. (The content hashing we have _does_ do some Merkle-like things, but it's not full-throated.)

More pithily, if we keep this line at all, I'd say:

* Concern: Merkle-tree-based hashing.
* Canonical home: N/A. We don't do Merkle-tree-based hashing.
* Do not: Invent or import a Merkle-tree library, unless specifically asked.

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,389 @@
...
+  `deno.json` `workspace` and have its own `tasks.test` — a missing test task
+  makes the root runner recurse and time out CI. (See DEVELOPMENT.md.)
+- Formatting: `deno fmt --check` on touched files.
+
+### 6. Craft & conventions
+
+Mostly Improvement / Nit. Don't drown the report in these.
+
+- Dead code, unused exports/imports, superfluous abstraction, unclear names.
+- Types: no needless `any`; export types with `export type { ... }`; JSDoc on
```
danfuzz: Also suggest:

* No needless _casts_. I often see an `as Something` where the `as Something` isn't required for correctness and detracts from readability.
* No unjustified `as unknown as Something`, especially if it is erasing an `Immutable<T>` or `Readonly<T>` from the type of the value.

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,389 @@
...
+  public interfaces; **named exports over default**.
+- Imports grouped (std → external → internal); import from `@commonfabric/api`
+  **xor** `@commonfabric/api/interface`, not both.
+- Module-graph hygiene: leaf utils stay dependency-light; no new cycles;
+  module-specific deps go in that module's `deno.json`, not the root.
```
danfuzz: Not sure if this would qualify as company consensus, but in the instructions I use for my own work, I have something like:

* Succint, accurate, and meaningful documentation comments on every `export`ed function, variable, and type; and on every public property of a class. Documentation strongly encouraged on internal items as well.

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,389 @@
...
+  guard?** If you can't name it, that's a finding — it may be testing an
+  implementation detail or be incidental.
+- Are these the **right cases**? Do they cover the actual semantic change and
+  its edge/empty/error states — not just the happy path?
+- Right **level**? Behavior over internals; integration where integration is the
+  real contract.
+- **Removed** tests: justified (dead feature) or a silent coverage drop?
+- Reference `docs/common/ai/pattern-testing-guide.md`, and run targeted tests
+  rather than the whole suite while reviewing: `deno test path/to/file.test.ts`.
+
```
danfuzz, line 215: Suggest:

* Prefer the BDD/TDD style testing framework (`import { describe, it, ... } from "@std/testing/bdd"`) over raw `Deno.test()`. (I also get picky about `it("...")`s reading as proper sentences.)
* Prefer `expect()` (`import { expect } from "@std/expect") over `assert*()`.

Hixie:
Seems fine to me, the only thing so far that I've noticed occasionally in the codebase (and in general as a thing LLMs for some reason really love to do) that I would call out here is the use of `sleep` statements and other equivalents. IMHO there is basically never a justification for just pausing a thread on _time_ (as opposed to waiting for a pid, or a semaphore, or whatever it is that we actually care about) other than animations, and it often introduces flakiness. Any process that normally takes X amount of time is always guaranteed to take wildly more than that one day (especially under CI, where it causes flaky tests), and when it does things break in the most annoyingly difficult to debug ways.

mpsalisbury (approved):
Really like the shape of this — it's written *against* the over-prescription failure mode, not into it ("assumes you already know how to read code"; dimensions framed as "where to look, not the structure of the report"). The anti-duplication table and the "source is two abstraction levels above what runs" framing are exactly the non-derivable local knowledge a skill *should* carry. Feedback in two parts — a strategy lens, then specific asks. Only one of the asks is blocking.

## Strategy lens (rationale, not a blocker)

A frame I'd apply to this and every skill: every line is either **map & values** (non-derivable local knowledge — canonical homes, repo history, how harsh to be) or **procedure** (ordering, templates, step-by-step). They have opposite trajectories: the map *appreciates* as models improve (a smarter agent does more with it), procedure *depreciates* (a capable model already does it, then starts fighting it). So a skill should be a map and a statement of values, not a recipe — and what over-constrains an agent is specifically the **closed list presented as complete**; prefer generative principles and mark any enumeration as "e.g." seed. A skill trades a little ceiling for a lot of floor, which is the right trade for something run by many agents on every PR. Last point, and the one this PR proves: the highest-value content (the fact table) is the highest-*rot* — so the durable fix is to make facts testable, not to write fewer of them.

I wrote this up as a standalone guide for us to debate so it outlives any one PR: **#3836** (`docs/development/skill-authoring.md`). The asks below fall out of my current thinking that it captures.

## Specific changes

**🔴 Blocking**
1. The `value-clone` row lists `shallowMutableClone`, which exists **nowhere** in the tree (`gh pr diff` line ~208). `packages/data-model/src/value-clone.ts` exports `cloneIfNecessary`, `cloneHelper`, `cloneForMutation` — replace with `cloneHelper` (or drop the third name). This matters because it's *inside the anti-duplication table*: a skill whose job is to stop people chasing phantom APIs is pointing at one. One-word fix.

**🟡 Requested (non-blocking — proposals, in the skill's own "don't demand pristine" spirit)**
- **Make the canonical-homes table testable.** A pre-commit/CI grep that checks each `@commonfabric/*` symbol the skill names against actual exports. The `shallowMutableClone` slip is the argument for it: that table is the most valuable *and* most rot-prone part — pin it down so it stays an asset.
- **`refer()` row is a weak pointer.** It points at a private method on `Invocation` in `packages/memory/consumer.ts` that just returns `hashOf(this.source)` — not an importable home like every other row. Point merkle-ref hashing at `hashOf`, or reword the row.
- **Trim the procedure** (lowers both ceiling-cost and rot surface): collapse the report template to its only real invariant (three buckets + a verdict; let the agent format), turn the per-dimension tell-lists into "e.g." seeds rather than closed enumerations, and compress the Step 3 command catalog.
- **One line on how this relates to `/code-review` and `/review`**, so invokers aren't unsure which to reach for.

Net: shippable after the one-word `cloneHelper` fix. Everything else is durable-quality work that can land here or as fast-follows.

bfollington:
Thanks all — genuinely high-signal feedback, and it converged so cleanly that most of it is now applied in the latest push (`800df1de3`). Summary of changes and decisions, by reviewer:

### @mpsalisbury
Adopted the **map-vs-procedure** framing from #3836 throughout: collapsed the report template to its invariant (three buckets + a verdict), turned the per-dimension tells into `e.g.` seeds rather than closed lists, compressed the verify-command list, and **cut the anti-dup grep recipe**. The skill now opens by declaring itself "a map and a statement of values, not a recipe" and links `docs/development/skill-authoring.md`. Net 389 → 250 lines.

On the 🔴 blocking `shallowMutableClone`: I verified it **does** exist (`value-clone.ts:125`, on `main` too), so that specific finding was a false alarm — but it's the perfect argument for your durable point, and we've resolved it the other way regardless: the table no longer lists internal symbols at all, only canonical **modules** (it points, doesn't copy). The "make facts testable" CI-grep is noted as a fast-follow if we ever reintroduce symbol-level pointers.

### @bfollington
Stopped uplifting specific fields (the table points to modules/exports now); reframed the LLM-blaming as **high-velocity-team failure modes** ("not AI's fault in particular"); trimmed the superfluous negative prompting on the north-star; cut the grep block (the ":205 bloat"); and reframed "two abstraction levels above what runs" → "a change can span many abstraction levels and its implications are not obvious in isolation." Posting is now report-first **then offer to post a self-signed review** (see @danfuzz).

### @danfuzz
Adopted the **Merkle correction** verbatim — the row now reads "Merkle-tree hashing → **N/A, we don't do it**; don't invent/import a Merkle lib unless asked" (thanks for the `modern-hash` history). Added the `JSON.stringify()`-to-simulate-a-hash footgun with your rationale (erases type identity + most contents of non-plain objects); LEB128 → "or similar variable-length integer encoding"; and added needless / unjustified casts (incl. erasing `Immutable<T>`/`Readonly<T>`) to the craft dimension. **Self-signing adopted** — the skill now posts as the agent + model "on behalf of @human", never impersonating (this comment is the convention in action). The JSDoc-on-exports and BDD/`expect()` testing-style asks I've **pointed at `DEVELOPMENT.md`** rather than enumerating in the skill — and since you flagged BDD as not-yet-consensus, I'd rather canonicalize those in `DEVELOPMENT.md` first. Happy to open that PR if you want.

### @Hixie
Added a **time-based-wait / `sleep`** tell to the correctness dimension: wait on the actual event or signal, not the clock; a prime source of CI flakiness. Good call.

### Open / deferred
- The `docs/development/skill-authoring.md` reference lands with #3836 — these two should merge together.
- Out of scope here, but worth a separate look: the runner's `module-identity.ts` describes its hashing as "Merkle," which collides with @danfuzz's point that we don't do Merkle-tree hashing — a coherence drift of exactly the kind this skill is meant to catch.

🤖 Posted by Claude (Opus 4.8) on behalf of @bfollington

bfollington:
Marking this ready for review.

The "make the anti-dup facts testable" durable ask (thanks @mpsalisbury) is drafted as a stacked follow-up — **#3863**: a dependency-free `deno task check-skill-facts` that fails if any `@commonfabric/*` package, repo path, or `data-model` export the skill cites stops resolving. It would have caught the `shallowMutableClone` / stale `refer()` / Merkle slips from this review. Kept separate so this PR stays docs-only; it retargets to `main` once this merges.

🤖 Posted by Claude (Opus 4.8) on behalf of @bfollington

mpsalisbury (approved):
One non-blocking leftover

  - **Consistency nit on the `convertCellsToLinks` row.** It still names a bare symbol + path + internal helpers — the symbol-rot the data-model rows were moved away from. Accurate today and arguably unavoidable (no package home for an internal runner fn), but it's the row #3863's checker is least likely to cover. A line for that follow-up's scope, not here.

LGTM

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,250 @@
...
+> on a doc, comment, or example, they must not be **misled** by something that
+> no longer matches how the runtime works.
+
+That is the bar — not uniform vocabulary. Wording drift is a slow Boy-Scout fix:
+nudge what you touch, propose big sweeps as follow-ups. Block only when a change
+leaves an authoritative source _actively wrong_.
+
+## Why this repo is easy to get wrong
+
+A change is authored as TypeScript, **transformed** by a custom toolchain (CTS)
```
mathpirate: we don't really use "CTS" much anymore - it's only for a basically-unused cts-disable flag. also, while this is true for patterns, it's not true for our labs/ repo code (which is untransformed). i suggest updating this section to something like -

"Patterns, which are the semantic equivalents of applications in our platform, are authored in TypeScript, and then transformed by an extensive pipeline of custom transformers into a much more complex form that correctly handles the framework's requirements and structure, so that the pattern can build a reactive graph evaluated under live subscriptions. So, changes to patterns and the runtime code that executes them can span many abstraction levels and the implications are not always obvious in isolation. Twi failure modes follow..."

mathpirate: I think we need one paragraph intro'ing this section clarifying the difference between reviewing changes to the repos vs reviewing changes to patterns

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,250 @@
...
+than any list here.
+
+### 1. Correctness & regressions — the loudest category
+
+A confirmed bug or regression is Blocking. Watch especially for:
+
+- **Tests bent to fit a regression** — an existing test weakened or retrofitted
+  so new (possibly wrong) behavior passes. A test edited to assert a bug is
+  Blocking.
+- **Reactivity hazards** — reactive self-feedback (a `cf-*` control bound via
```
mathpirate: this is also specific to reviewing patterns, as is the next bullet; the first one, however, is general to also reviewing platform code

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,250 @@
...
+recurring). Tells (seeds, drawn from `docs/development/DEVELOPMENT.md`):
+
+- over-eager try/catch that swallows errors a caller should see (throwing is
+  fine; fatal errors _should_ propagate);
+- new singletons / module-global mutable state (breaks multiple instances +
+  tests);
+- ambiguous `any` away from a serialization boundary; types that admit invalid
+  intermediate states;
+- working around the transformer (stray `/// <cf-disable-transform />`, manual
+  graph wiring CTS would do, imperative escapes from the target language);
```
mathpirate: don't use CTS here, say "the transformers should handle"

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,250 @@
...
+- async escapes in patterns (`async/await` in handlers;
+  `await
+  generateText/Object` instead of `.result`; `new Stream()` /
+  `.subscribe()`).
+
+When a transformer behavior is in doubt, read the **emitted output** before
+reasoning from source:
+`deno task cf check <file>.tsx --show-transformed --no-run`.
```
mathpirate: put `[--no-run]` in brackets bc it's sometimes correct and sometimes not

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,250 @@
...
+
+### 5. Changeset hygiene
+
+Stray cruft is trivial to fix but shouldn't merge: leftover debug logging /
+`*.log` / scratch notes, commented-out code, `.only` on tests, "temp" / "HACK"
+stopgaps, abandoned TODOs. A new workspace package must register in the root
+`deno.json` and carry its own `tasks.test` (a missing one makes the root runner
+recurse and time out CI). Run `deno fmt` on touched files.
```
mathpirate: "Run `deno fmt` and `deno lint` on touched files."

bfollington:
Latest round folded in (`36716145a`) — thanks @mathpirate and @mpsalisbury.

### @mathpirate (Gideon) — all six, applied
- **The big one: "Two kinds of change."** New intro distinguishing **platform code** (the runtime + most of `packages/**`, plain *untransformed* TypeScript — review it as any TS codebase) from **patterns** (authored TS, transformed by a pipeline of custom transformers into a reactive graph; the runtime that serves them spans the same abstraction levels). My earlier framing wrongly implied everything is transformed — fixed.
- Dropped the near-dead **"CTS"** → "the transformers" (both spots).
- Dimension 1 tells now tagged **pattern-specific** (reactivity, determinism/`safeDateNow`) vs **general** (regression-tests, time-based waits).
- `--show-transformed [--no-run]` (optional); "Run `deno fmt` **and `deno lint`** on touched files."

### @mpsalisbury — convertCellsToLinks consistency note
Left the row as-is here (it's accurate, and there's no package home for an internal runner fn), and — per your "a line for that follow-up's scope, not here" — documented it in **#3863**'s `skill-audit.md` as the one row the deterministic tripwire can't cover (bare internal symbol → the LLM audit's job).

### cubic
The `@commonfabric/data-model` bare-import issue cubic flagged was already addressed in `1e41589` (the rows now point at the importable subpath modules).

### One stance change worth flagging
From dogfooding the skill on a real PR, the skill now explicitly frames itself as **an aid to thinking, not a merge gate** — surface spec/principle violations loudly so the author decides with eyes open; no hard blocking rule, and 🔴 reads "resolve or consciously accept before merge."

The stacked follow-ups (#3863 tripwire+CI, #3865 `cubic.yaml`) are rebased onto this and green. Over to you for the final sweep.

🤖 Posted by Claude (Opus 4.8) on behalf of @bfollington

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,270 @@
...
+description: Review a changeset for the Common Fabric repo — the local branch diff vs main, or a GitHub PR. Flags correctness and regression bugs loudly, checks that runtime-semantics changes stay coherent across docs/comments/examples, catches duplicated core machinery (hashing, serialization, cloning, identity) and code fighting the transformer/reactive model, and scrutinizes whether the tests guard the right principles. Report-first; offers to post a self-signed PR review. Use when asked to review code, review a PR, review the current branch or diff, or self-review before pushing. Invoke as /cf-review or /cf-review <PR#>.
+---
+
+# Common Fabric Code Review
+
+This skill is a **map and a statement of values**, not a recipe. It assumes you
+already know how to read and review code; it tells you the things about _this_
+repo you cannot derive — where its value and its footguns are, and how we like
+review done. (For why skills are written this way, see
+`docs/development/skill-authoring.md`.)
```
bfollington: Is this important for an _agent_ or a human reader? I am sensitive to bloat here.

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,270 @@
...
+
+All paths are relative to the repo root.
+
+## What we value in a review
+
+- **Changeset-scoped.** Review the diff and its immediate ripple — which for a
+  big change can mean fanning out to trace _that change's_ blast radius — never
+  an open-ended repo audit. This review is fast.
```
bfollington: > Review the diff and its immediate ripple — which for a big change can mean fanning out to trace _that change's_ blast radius — never an open-ended repo audit. This review is fast.

Review the diff and its immediate effects (and for large changes, the blast radius) but never a repo-wide audit. Focus on the changeset.

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,270 @@
...
+- **Loud on what matters, quiet on what doesn't.** Bugs, regressions, and broken
+  principles, unmistakably; don't pad with nits. We ship many PRs and don't
+  demand a pristine first cycle — the bar is no regressions, no silent principle
+  breaks, and no one misled later by a stale doc this change should have
+  updated.
```
bfollington: Loud, with intention. Too wordy.

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,270 @@
...
+- **Report-first, then offer to post.** Always show the report. Then offer to
+  post it to the PR — **signed as yourself** (the agent and model, on behalf of
+  the human), never impersonating them. Skip posting when it isn't worth it.
+- **An aid to thinking, not a gate.** This review helps people decide with eyes
+  open; it does not police merges. Surface what matters — especially a spec or
+  principle a change violates — loudly and unmistakably so the author _knows_;
+  they own the merge and may proceed regardless. Don't posture as a gatekeeper.
```
bfollington: "Don't" we avoid negative prompting when possible. Use positively framed (do this) principles and values. Not bandaid fixes.

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,270 @@
...
+  the framework's structure, so the pattern can build a reactive graph evaluated
+  under live subscriptions. The source is far from what runs, so a change — to a
+  pattern _or to the runtime code that executes it_ — can span many abstraction
+  levels with implications that aren't obvious in isolation.
+- **Platform code** (most of `packages/**`) is plain TypeScript: review it as
+  you would any TS codebase, plus this repo's conventions and the
+  anti-duplication map below.
+
+Two failure modes recur on a high-velocity team — not AI's fault in particular,
+just easy to do when moving fast without the whole picture:
```
bfollington: Don't need to even call the AI out here

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,270 @@
...
+  top-down **theory of intent and blast radius**; deep-read only the
+  load-bearing files (the core new abstraction, public API / signature changes,
+  and any hashing / serialization / cloning / identity touchpoints); and **fan
+  out read-only subagents** for the search-heavy parts — tracing callers not in
+  the diff that should have changed, sweeping docs / examples for touched
+  concepts, reviewing a cluster, hunting prior art the change may be forking.
+  Give each a tight question; take back the conclusion, not file dumps.
+  Synthesize into one report and test the change against its own theory (scope
+  creep, half-finished migration, a missing caller / doc / test). **Say what you
+  deep-read vs. sampled vs. delegated** — never imply coverage you didn't do.
```
bfollington: This is too long, but should mention subagents (or they should be mentioned more globally)

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,270 @@
...
+fix toward the idiom, and — when the confusion was avoidable — _the doc or
+structure gap that caused it_ (a follow-up; that is how the mistake stops
+recurring). Tells (seeds, drawn from `docs/development/DEVELOPMENT.md`):
+
+- over-eager try/catch that swallows errors a caller should see (throwing is
+  fine; fatal errors _should_ propagate);
+- new singletons / module-global mutable state (breaks multiple instances +
+  tests);
+- ambiguous `any` away from a serialization boundary; types that admit invalid
+  intermediate states;
```
bfollington: also abuse of `unknown`

skills/cf-review/SKILL.md
```diff
@@ -0,0 +1,270 @@
...
+code to the canonical home rather than cargo-culting a neighbor. A new such
+definition isn't automatically wrong, but the PR must justify why the canonical
+home doesn't fit; a _silent_ fork is Blocking.
+
+### 4. Framework-fit
+
+Spot code fighting the framework instead of using it, and give two outputs: the
+fix toward the idiom, and — when the confusion was avoidable — _the doc or
+structure gap that caused it_ (a follow-up; that is how the mistake stops
+recurring). Tells (seeds, drawn from `docs/development/DEVELOPMENT.md`):
```
bfollington, line 162: In general, if it seems the transformers are causing issue and the author is working around, suggest diagnosing and logging a transformers bug (with evidence / repro) instead.

skills/cf-review/SKILL.md
bfollington: I agree, I think tests should be structured to help expose the intended behaviour and BDD should read like sentences / user stories. I suspect we do this poorly / inconsistently today.

bfollington:
Polish round applied (`805947cb6`) — all nine addressed:

- **intro** — cut the skill-authoring meta-pointer (bloat for an agent).
- **changeset bullet** — tightened to your wording.
- **"Loud, with intention"** — trimmed the wordy version.
- **aid-not-gate** — reframed positively ("An aid to thinking… the human owns the merge"), and swept the parallel negatives: sign "on the human's behalf" (was "never impersonating"); follow-ups "for a human to file".
- **failure modes** — dropped the AI call-out.
- **scope-and-theory** — trimmed; kept the subagent fan-out.
- **framework-fit** — added: if the transformer is the real culprit, diagnose and file a transformer bug with a repro rather than hand-building a workaround.
- **types** — added abuse of `unknown` alongside ambiguous `any`.
- **test rigor** — a good test exposes the intended behavior and reads like a sentence (a user story), not an assertion on internals.

Net 271 → 261 lines. Stacked follow-ups (#3863, #3865) rebased on this and green. Back to you for the sweep.

🤖 Posted by Claude (Opus 4.8) on behalf of @bfollington

skills/cf-review/SKILL.md
bfollington: Good catch — my earlier pass only captured the "reads like a sentence" spirit. Now folded in concretely (`4ae3727d9`): test rigor names the direction — prefer BDD style (`@std/testing/bdd` `describe`/`it` that read as sentences) and `expect()` over `assert*()` — framed as a direction we apply inconsistently today, not a gate.

🤖 via Claude (Opus 4.8) on behalf of @bfollington

PR #3840, opened 2026-06-03
feat(cli): read id derive/from-mnemonic secrets from a file or stdin

Hixie:
Passing a passphrase or recovery phrase as an argv positional leaks it into shell history and the process argument list (visible via `ps`). Both `id derive` and `id from-mnemonic` now accept the secret without putting it on the command line.

Resolution precedence for the passphrase/mnemonic:
  * `-- <file>`  -> read the secret from that file (cliffy routes post-`--` tokens to getLiteralArgs(), so this stays unambiguous);
  * `-`, or no argument at all -> read the secret from stdin;
  * otherwise -> the inline argument, used verbatim (backwards compatible).

File and stdin input have a single trailing newline stripped so an editor- or `echo`-produced value matches the equivalent inline argument. Empty input, a missing file, more than one file after `--`, or mixing an inline value with `-- <file>` all error cleanly with nothing written to stdout.

Docs updated to recommend the file/stdin forms (SHARED_IDENTITY.md, HOME_SPACE.md, and the cf skill, replacing the old `deno eval` fromMnemonic workaround). Adds a `stdin` option to the cli test harness and covers the file, stdin, and conflict cases.

---
## Summary by cubic
Add secure input for `cf id derive` and `cf id from-mnemonic` so passphrases and recovery phrases can be read from a file or stdin instead of argv, avoiding leaks to shell history and `ps`. Backwards compatible, with updated help text, docs, and tests.

- New Features
  - Input precedence: `-- <file>`; `-` or no arg -> stdin; else inline arg (kept for compatibility).
  - Strip a single trailing newline from file/stdin input to match inline values.
  - Clear errors for empty input, missing file, multiple files after `--`, or mixing inline with `-- <file>`.
  - Updated examples in command help and docs; added stdin support to the CLI test harness and tests for file/stdin/conflict cases.

<sup>Written for commit 560f01b067504b3ca4858c359d00aebd8bd0da74. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3840?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/common/conventions/HOME_SPACE.md
+++ b/docs/common/conventions/HOME_SPACE.md
@@ -140,21 +140,19 @@ To share identity between browser and CLI:
 
 ```bash
 # 1. Create a mnemonic in the browser (login/register screen)
-# 2. Export a CLI key using fromMnemonic (not fromPassphrase):
-deno eval '
-import { Identity } from "./packages/identity/src/identity.ts";
-const mnemonic = "your 24-word mnemonic here";
-const id = await Identity.fromMnemonic(mnemonic, { implementation: "noble" });
-await Deno.writeFile("./browser.key", id.toPkcs8());
-'
+# 2. Export a matching CLI key with `cf id from-mnemonic`, reading the phrase
+#    from a file (`-- <file>`; or `-` for stdin) so it stays out of shell
+#    history and the process list:
+deno run -A packages/cli/mod.ts id from-mnemonic -- phrase.txt > ./browser.key
 
 # 3. Use that key with cf
 cf piece set-home -i ./browser.key -a http://localhost:8000 ./my-home.tsx
 ```
 
 Note: `cf id derive <passphrase>` will NOT produce the same identity as the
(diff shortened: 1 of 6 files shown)


mathpirate (approved):
lgtm, no findings from me or the agent, ty

Hixie:
I tested this locally. It adds 7 new CLI tests, which is a 275% increase, and local wall clock time for test execution went from 1.7s to 4.5s which is a similar increase in total time, so the regression is expected.

---BEGIN COPY-PASTE---
NEW_PERF_BASELINE: job: CLI Integration Tests (core-piece-call) = 129s
NEW_PERF_BASELINE: job: CLI Integration Tests (core) = 129s
---END COPY-PASTE---

PR #3841, opened 2026-06-03
fix(schema-generator): collapse expanded Default<[]> array unions (CT-1639 pt 1)

mathpirate:
## Summary

Part 1 of CT-1639 (the silent `equals()`-based list-removal no-op under `Default<[]>`). This is the **schema-generator** half, fixed in a way that's correct for **any** consumer — not just the transformer.

When `Default<[]>` reaches the schema generator **already expanded** by the checker (no intact `Default<>` alias node — which is what the transformer hands it for a synthesized handler/map-context schema), the union is:

```
RealArray | ([] & DefaultMarker<[]>) | []
```

`UnionFormatter.tryFormatDefaultUnion` only recognizes an **intact** `Default<>` TypeReference node, so it returned `undefined` here and the code fell through to the generic multi-branch `anyOf` path. That split the real array from the empty-array branch (`{type:"array", items:false}`), diluting the real array member and dropping per-element annotations like `asCell:["comparable"]` from a consumer's view.

## Fix

Add `UnionFormatter.tryFormatExpandedEmptyArrayDefault`, run right after `tryFormatDefaultUnion`:
- Detect the **degenerate empty-array** members: bare `[]` / `never[]`, or `[] & DefaultMarker<[]>` (intersection — brand stripped via the `__@`-symbol convention already used by `IntersectionFormatter`).
- When exactly **one real array member** remains alongside ≥1 degenerate empty member, format only the real member and attach `default: []`.
- Returns `undefined` (falls back to normal union handling) for anything else — more than one real array member, or any non-array member.

This preserves the real array's element schema (including `comparable`) and applies the empty-array default, matching what `tryFormatDefaultUnion` does for the intact-alias form.

## Scope / follow-up

This is the schema-generator half. There is a **separate ts-transformers gap** (tracked as CT-1639 pt 2): in the real handler-param pipeline the array element handed to schema-gen is not always `comparable`-wrapped, because type-shrinking wraps the element **node** but doesn't re-register the wrapped **semantic type** that schema-gen reads. With pt 1 alone, the empty-array branch now collapses correctly (`default:[]` lands) but the `comparable` annotation can still be absent until pt 2. Pt 1 is a correct, independently-valuable improvement.

## Test plan

- New `default-union` test: an **expanded** `Default<[]>` array union keeps `asCell:["comparable"]` on a single array schema, with no empty-only sibling branch. (Fails before, passes after.)
- Full schema-generator suite green: **24 passed (227 steps)**.
- ts-transformers suite green (292); runner `patterns-default-array-materialization` + `patterns-findindex` green.

Part of CT-1639.

---
## Summary by cubic
Fixes CT-1639 pt 1 by collapsing expanded `Default<[]>` array unions in `schema-generator` so a single array schema keeps item annotations and gets `default: []` instead of `anyOf`. Adds a brand check so collapse only happens when a `[] & DefaultMarker<[]>` member is present, avoiding fabricated defaults.

- **Bug Fixes**
  - Detect unions like `RealArray | ([] & DefaultMarker<[]>) | []` and emit the real array with `default: []`, preserving item schema (e.g., `asCell: ["comparable"]`); skip if multiple real arrays or any non-array member exists.
  - Require a branded empty member before collapsing; plain unions like `string[] | []` are left untouched.
  - Tests: cases for the branded expanded form and the no-brand union; `schema-generator` and `ts-transformers` suites pass.

<sup>Written for commit 9f8a3f1df2f2191f6cffbdfc173bd3cedba04e37. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3841?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/schema-generator/src/formatters/union-formatter.ts
+++ b/packages/schema-generator/src/formatters/union-formatter.ts
@@ -332,12 +332,20 @@ export class UnionFormatter implements TypeFormatter {
     const checker = context.typeChecker;
 
     let realArrayIndex = -1;
-    let sawDegenerateEmpty = false;
+    // Only the BRANDED empty member (`[] & DefaultMarker<[]>`) proves this is an
+    // expanded `Default<[]>`. A bare `[]` / `never[]` is the *unbranded* arm of
+    // that same Default union — but it also appears in ordinary unions like
+    // `string[] | []` that have nothing to do with defaults. So we require at
+    // least one branded empty member before collapsing + attaching `default:[]`;
+    // a bare empty member alone is left to normal union handling. (CT-1639)
+    let sawBrandedEmpty = false;
 
     for (let i = 0; i < members.length; i++) {
       const member = members[i]!;
       if (this.isDegenerateEmptyArrayMember(member, checker)) {
-        sawDegenerateEmpty = true;
+        if (this.hasDefaultBrand(member, checker)) {
+          sawBrandedEmpty = true;
+        }
         continue;
       }
       if (!checker.isArrayType(member) && !checker.isTupleType(member)) {
@@ -353,7 +361,7 @@ export class UnionFormatter implements TypeFormatter {
       }
     }
 
-    if (realArrayIndex === -1 || !sawDegenerateEmpty) {
+    if (realArrayIndex === -1 || !sawBrandedEmpty) {
       return undefined;
     }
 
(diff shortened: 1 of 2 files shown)


mathpirate:
Thanks — cubic's P1 finding was valid and is now fixed in fb72b3f.

**The bug:** the collapse path keyed on a bare `[]` / `never[]` union member as the signal for "expanded `Default<[]>`". But that bare member is the *unbranded* arm of a `Default<[]>` union and also appears in ordinary unions like `string[] | []` — so the path fired on plain unions and attached a `default: []` the author never wrote. Confirmed directly: `type T = string[] | []` was emitting `{type:"array", items:{type:"string"}, default:[]}`.

**The fix:** require at least one **branded** empty member (`[] & DefaultMarker<...>`) before collapsing — that brand is the only thing distinguishing an expanded `Default<[]>` from a plain `string[] | []`. A bare empty member with no branded sibling now falls through to normal union handling. Added a regression test asserting `string[] | []` gets no fabricated default; the CT-1639 branded case still collapses and preserves `asCell:["comparable"]`. Full schema-generator suite green.

PR #3843, opened 2026-06-03
fix(runner): make piece save path content-identity-symmetric (CT-1623)

seefeldb:
## What

Fixes a regression surfaced by the flag-flip canary (#3782): under the **ESM module loader**, `cf piece setsrc` on a pattern with a **non-default `mainExport`** hangs (scheduler never reaches idle → the CLI's `getResult().pull()` never resolves) or fails. The FUSE CLI integration test reproduces it reliably (it deploys with `--main-export customPatternExport`).

This was dormant on `main` (the ESM loader is still off by default), so #3833 merged green — the flip is the first time the content-addressed reload path runs live.

## Root cause — a save/load asymmetry

When result cells gained the content-addressed `{identity, symbol}` reference (#3833):

- The piece **save** path — `compileAndSavePattern` → `compilePattern(src)` with **no `cacheCtx`** — took the non-cache ESM branch, never associated an `entryIdentity`, and `patternFromEvaluation` was called without it. So `applySetupState` wrote **no/stale** `patternIdentity`.
- The daemon's **load** path *did* learn the `entryIdentity` and wrote `patternIdentity` with the real symbol.
- CLI and daemon share the space, so they disagreed about the content identity. With a non-default export, the by-identity reload used the wrong symbol (`"default"`), `loadPatternByIdentity` missed/threw, and the `sinkMeta("pattern")` watcher / `ensurePieceRunning` churned without converging.

## Fix

- **`compileAndSavePattern` compiles with `cacheCtx: { space }`** — the save path now participates fully in the content-addressed cache: learns the entry identity, associates the `{identity, symbol}` ref with the real export name, and writes the compiled + source docs into the target space (so the daemon's by-identity reload **hits** instead of churning).
- **The non-cache ESM branch of `compilePattern` passes `entryIdentity` to `patternFromEvaluation`** — so the ref is consistent even when the compiled cache isn't written (CFC disabled / other callers).

## Test

`compile-and-save-identity.test.ts`: a non-default-export pattern saved via `compileAndSavePattern` records `{identity, symbol: "customPatternExport"}` (not `"default"`) and populates the compiled closure in the target space.

Full `runner` suite (526) + `piece` suite green.

## Validation plan

After this lands on `main`, rebase the flag-flip PR (#3782) on top and confirm the FUSE CLI integration goes green (and the Performance Check runs again instead of being skipped behind the fuse failure).

---
## Summary by cubic
Fixes a hang/failure in `cf piece setsrc` for patterns with a non-default `mainExport` under the ESM loader by making the save path content-identity symmetric with the load path. Addresses CT-1623.

- **Bug Fixes**
  - `compileAndSavePattern` now compiles with `cacheCtx: { space }` to learn the entry identity, set the `{identity, symbol}` with the real export name, and write compiled + source docs into the target space.
  - The non-cache ESM branch of `compilePattern` now passes `entryIdentity` to `patternFromEvaluation` so refs stay consistent even without cache writes.
  - Added `compile-and-save-identity.test.ts` to verify the `{identity, symbol: "customPatternExport"}` ref and that the compiled closure is stored in the target space.

<sup>Written for commit fd4f4956f73e1e7d7bc750acb53ee008e2a3425d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3843?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/pattern-manager.ts
+++ b/packages/runner/src/pattern-manager.ts
@@ -545,7 +545,10 @@ export class PatternManager {
         mainSpecifier,
         program.files,
       );
-      return this.patternFromEvaluation(result, program);
+      // Associate the entry identity even on the non-cache ESM branch so the
+      // pattern's {identity, symbol} ref is always consistent (correct export
+      // symbol), regardless of whether the compiled cache was written.
+      return this.patternFromEvaluation(result, program, entryIdentity);
     }
 
     const { cachedCompiler } = this.runtime;
--- a/packages/runner/src/piece-helpers.ts
+++ b/packages/runner/src/piece-helpers.ts
@@ -100,7 +100,16 @@ export async function compileAndSavePattern(
       files: [{ name: "/main.tsx", contents: patternSrc }],
     };
   }
-  const pattern = await runtime.patternManager.compilePattern(patternSrc);
+  // Compile with the target space so the ESM path participates fully in the
+  // content-addressed cell cache: it learns the entry identity, associates the
+  // pattern's {identity, symbol} ref (with the real export name), and writes the
+  // compiled + source docs into `space`. Without this the save path and the
+  // (cache-backed) load path disagree about the content identity — a non-default
+  // `mainExport` would reload under the wrong symbol and the watcher would never
+  // converge (cf piece setsrc hang under the ESM loader).
+  const pattern = await runtime.patternManager.compilePattern(patternSrc, {
+    space: options.space,
+  });
   if (!pattern) {
     throw new Error("No default pattern found in the compiled exports.");
   }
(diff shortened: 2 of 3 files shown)


seefeldb:
Closing: misdirected. Investigation showed the `cf piece setsrc` hang is **FUSE-daemon-specific** (plain non-FUSE `core-piece-values` passes on the flip), so the save/load asymmetry this targeted wasn't the cause — and the change regressed cfc-group-chat-demo under ESM. Superseded by a CI daemon-log-path fix (real fuse errors have been captured from the wrong path), after which the actual fuse-hang fix will be written from ground truth.

PR #3856, opened 2026-06-03
chore(deno): upgrade toolchain to Deno 2.8.1

Hixie:
Moves the repo's pinned Deno toolchain from 2.7.9 to 2.8.1. Deno 2.8 ships TypeScript 6.0.3 (was 5.9.2) and an updated `deno fmt`, both of which require code and formatting changes. This is a hard cutover: 2.7.x is no longer supported because its `deno fmt` output diverges from 2.8.x (see Formatting).

Version pins (all bumped 2.7.9 -> 2.8.1, kept in sync):
- mise.toml: dev toolchain pin.
- .github/actions/deno-setup/action.yml: CI default deno-version.
- tasks/check.sh: the guard that gates `deno check` (and the pre-commit hook). Range tightened from ">=2.6.0 <2.8.0" to ">=2.8.0 <2.9.0". The lower bound is deliberately raised to 2.8.0 (not left at 2.6.0) so the guard *rejects* 2.7.x: a developer on 2.7.9 would reformat files back to 2.7.x style on save and break CI's `deno fmt --check`, so the cutover must be enforced, not optional.

Type fixes (19 errors surfaced by TypeScript 6.0.3 in Deno 2.8):
- 18x "Type 'Timeout' is not assignable to type 'number'": in TS 6.0, setTimeout/setInterval return an opaque `Timeout` object rather than a `number`. Every timer-handle field/variable previously typed `number` is retyped `ReturnType<typeof setTimeout>` (or `setInterval`), which resolves to `number` on 2.7.9 and `Timeout` on 2.8.1 -- correct on both. Affected files: cli/lib/fuse-supervisor.ts; six runner/integration *.test.ts (array_push, basic-persistence, derive_array_leak, pattern-and-data-persistence, sqlite-db-query-decode, sync-schema-path); runner/test/scheduler-events.test.ts; shell/src/views/AppView.ts; ui components cf-copy-button, cf-scroll-area, cf-secret-viewer, cf-tools-chip, cf-voice-input; ui/core/input-timing-controller.ts.
- 1x WebSocket.send type error in memory/socket.ts: TS 5.7+ made typed arrays generic over their backing buffer, and 2.8's DOM `WebSocket.send` expects `BufferSource` (ArrayBuffer-backed). The `Out` generic of from/fromWithPrefix is narrowed from `string | Uint8Array | Blob` to `string | Uint8Array<ArrayBuffer> | Blob` so the sent value is accepted. Runtime behaviour is unchanged -- a SharedArrayBuffer-backed view was never a valid argument to `send` anyway.

Formatting (deno fmt 2.8.1):
- `deno fmt` was run repo-wide with 2.8.1. Its only behavioural change vs 2.7.9 is the indentation of content inside tagged template literals (Lit `css`/`html` blocks and the comments within them): 2.8.1 re-bases embedded content one level (2 spaces) shallower than 2.7.9. This reformatted 22 UI component/style files; the diffs are pure whitespace inside template literals, with no code changes. This drift is not configurable -- none of indentWidth/lineWidth/semiColons/ singleQuote/proseWrap/useTabs control embedded-content indentation -- which is why 2.7.x and 2.8.x cannot both satisfy `deno fmt --check`, and why this is a hard cutover rather than a both-versions-compatible change.

Formatting exclusions (deno.json fmt.exclude):
- Added cf-toggle.ts, cf-checkbox.ts, and cf-select.ts. Each contains a multi-line `<!-- ... -->` HTML comment inside an `html` template literal that triggers a non-determinism bug in Deno 2.8.1's formatter: it produces a different indentation on each pass and aborts with "Formatting not stable. Bailed after 5 tries. This indicates a bug in the formatter." Excluding the files is the workaround the tool itself recommends, and matches existing precedent (the cf-chart files are already excluded for unrelated reasons). Trade-off: these three files receive no `deno fmt` coverage until the upstream bug is fixed, after which the exclusions can be removed. The alternative -- collapsing each comment onto a single long line -- was rejected as a readability regression introduced purely to work around a tool bug.

Verified under Deno 2.8.1: `tasks/check.sh` (full type check) passes, `deno fmt --check` is clean, and `deno lint` is clean; the guard correctly rejects Deno 2.7.9.

---
## Summary by cubic
Hard cutover to Deno 2.8.1 (TypeScript 6.0.3) across dev and CI. Fixes timer/WebSocket types, resolves a runner teardown hang, and applies repo-wide formatting from the new `deno fmt`.

- **Migration**
  - Pins updated in `mise.toml`, `.github/actions/deno-setup/action.yml`, and `tasks/check.sh` (now enforces >=2.8.0 <2.9.0); 2.7.x is rejected to avoid formatter drift.
  - Update your local toolchain (e.g. `mise install`) and run `deno fmt` before committing.
  - Added `deno.json` `fmt.exclude` for `cf-toggle.ts`, `cf-checkbox.ts`, and `cf-select.ts` due to a Deno 2.8.1 formatter instability; remove once upstream is fixed.

- **Bug Fixes**
  - Runner storage teardown: track and close the consumer’s exact subscribed `WatchView` (`#subscribedWatchView`) during `close()`/`closeNow` to settle pending `next()` calls; fixes `Runtime.dispose()` hanging under 2.8.1 (e.g. `sync-schema-path`).
  - Type updates for TS 6.0.3: timer handles use `ReturnType<typeof setTimeout|setInterval>`, and WebSocket `send` accepts `Uint8Array<ArrayBuffer>` to satisfy DOM `BufferSource`.

<sup>Written for commit 64ecae8f4dbf8ed359009f59bd2a41e2a13145f1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3856?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Hixie:
error involves a timeout so is probably legit, i'll have an agent look at it

Hixie:
Claude says:
> The 2.8.1 cutover deterministically breaks the runner. Runtime.dispose() hangs forever during storage teardown — reproduced locally: identical code (my branch) passes in 45ms on Deno 2.7.9 but hangs to the 180s timeout on 2.8.1, with the 2.8.1 server fine in both cases. The root cause is a latent bug in the runner's own code (v2.ts watch/close logic): consumeUpdates binds to the first subscription view, but #watchView gets reassigned, so teardown closes the wrong view and a pending promise never settles. Deno 2.8.1's async scheduling exposes it; 2.7.9's timing hid it. It is not caused by my edits, and it's the kind of regression the original <2.8.0 Deno guard was protecting against.

Continuing to investigate.

Hixie:
The second commit is a logic change that Claude claims fixes a bug that we were previously not seeing because a bug in Deno (now fixed) with how it implemented timeouts meant that we were essentially racing in a different way than we do now.

willkelly (commented):
This lgtm from the github actions / ci side of things, but I suggest waiting on robin/berni approval re: runner stuff.

ubik2 (approved):
mostly looks good to me, and nice to have this done!
while the v2.ts changes look reasonable to me, i think @seefeldb may have a better sense. ideally, he can take a look and approve, but if that takes a while, I'm ok merging as-is.

Hixie:
After I get an LGTM I'll do another CI run because it's quite possible there'll be other breakage introduced (e.g. if anyone adds a setTimeout, that happened while I was getting this ready already!).

seefeldb (approved):

willkelly (approved):

Hixie:
thanks for approvals, running ci and will land on green

PR #3863, opened 2026-06-04
test(skills): make cf-review's load-bearing facts testable

bfollington:
Follow-up to #3829 (stacked). Implements @mpsalisbury's "make load-bearing facts testable" ask as a **hybrid** (per review discussion): a deterministic tripwire (floor) + an LLM audit (ceiling). A grep alone only catches existence/resolvability rot; the LLM audit is the more useful half at the limit.

## Tripwire — deterministic floor
`skills/cf-review/check-facts.ts` — `deno task check-skill-facts` (also a `Deno.test`). A cheap, instant, zero-token check that every `@commonfabric/...` specifier and repo path the cf-review skill cites **resolves** against the tree: a bare import needs a root (`.`) export, a subpath needs that subpath in `exports`, and cited paths must exist. **No hardcoded fact list. Wired into the CI `check` job.** Negative-tested — a bare reference to a subpath-only package (e.g. `@commonfabric/data-model`) or a vanished path makes it exit 1.

## Audit — LLM ceiling
`docs/development/skill-audit.md` — the periodic / on-change **semantic** pass that catches what a grep can't: a canonical home that moved or was renamed, stale advice, a missing home, framing drift. The auditor is **cf-review itself**, not a parallel reviewer. This half *appreciates* as models improve.

## Stacking
Based on `add-cf-review-skill` (#3829) because the tripwire reads `skills/cf-review/SKILL.md`. Retarget to `main` once #3829 merges.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — on behalf of @bfollington

### DIFF
--- a/.github/workflows/deno.yml
+++ b/.github/workflows/deno.yml
@@ -43,6 +43,9 @@ jobs:
       - name: 🔎 Type check codebase
         run: deno task check
 
+      - name: 🧭 Check cf-review skill facts
+        run: deno task check-skill-facts
+
   test:
     name: "Test"
     runs-on:
--- a/deno.json
+++ b/deno.json
@@ -37,6 +37,7 @@
   ],
   "tasks": {
     "check": "./tasks/check.sh",
+    "check-skill-facts": "deno test --allow-read skills/cf-review/check-facts.ts",
     "cf": "deno run --allow-run --allow-env --allow-read ./packages/cli/launcher.ts",
     "test": "./tasks/test.ts",
     "test-all": "echo \"Use 'deno task test' instead.\" && exit 1",
--- a/docs/development/skill-audit.md
+++ b/docs/development/skill-audit.md
@@ -51,3 +51,10 @@ cited path may be skill-local rather than repo-root, a token may be illustrative
 (`cf-{name}.figma.ts`), and the worst case — "the home moved but both old and
 new names still exist" — is invisible to a grep. The tripwire stays narrow on
 purpose; the audit carries the judgment.
+
+One row it cannot cover by construction: the cf-review anti-dup table's
+cell↔link entry names a **bare internal symbol** (`convertCellsToLinks` in
+`packages/runner/src/cell.ts`) — an internal runner function with no package
+export, so there's nothing for the resolvability check to resolve. That symbol's
+accuracy is the audit's job, not the tripwire's. (Noted from @mpsalisbury's
+review of #3829.)
--- a/docs/development/skill-audit.md
+++ b/docs/development/skill-audit.md
@@ -1,0 +1,52 @@
+<!-- @reviewed 2026-06-04 -->
+
+# Skill Fact Audit
+
+A skill's map — canonical homes, exact symbols, file paths — is its
+highest-value and highest-rot content (see
+[`skill-authoring.md`](./skill-authoring.md), "facts rot — make them testable").
+Two mechanisms keep it honest, and they are complementary, not alternatives:
+
+- **Tripwire (deterministic floor).** `deno task check-skill-facts` — a cheap,
+  instant, zero-token CI gate that fails if a package or repo path a skill cites
+  stops existing. Runs on every PR (the `check` job). Catches _existence_ rot
+  only, and deliberately hardcodes nothing.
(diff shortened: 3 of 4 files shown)


mpsalisbury (approved):
Improvements

  1. Specifier regex misparses PascalCase/dotted subpaths → latent CI false-positive. skills/cf-review/check-facts.ts:58
  and :62. The match regex @commonfabric\/[a-z0-9-]+(?:\/[a-z0-9-]+)* only accepts lowercase subpath segments. data-model
  — a canonical home the skill actively points to — exports several PascalCase subpaths (./BaseReconstructionContext,
  ./EmptyReconstructionContext, ./fabric-type-tags…). I confirmed by eval: a valid future citation of
  `@commonfabric/data-model/BaseReconstructionContext` gets truncated to a bare @commonfabric/data-model, then the gate
  fails with "has no root export — bare … won't resolve; cite a subpath" — blocking a PR over a correct citation. No
  trigger today (the skill cites only lowercase subpaths), but this is exactly the brittleness class a reliability gate
  shouldn't carry. Fix: broaden the segment class to [\w.-]+ (or [A-Za-z0-9._-]+) in both the match regex (:58) and the
  parse regex (:62). verified.

Nits

  2. Generic naming vs cf-review-only implementation. The task check-skill-facts, the file check-facts.ts, and
  skill-audit.md's general framing ("an import specifier or repo path a skill cites") all read as repo-wide, but
  collectErrors() is hardcoded to skills/cf-review/SKILL.md while ~15 other skills have a SKILL.md. Fine as the first
  instance — but a future author could assume their new skill is gated when it isn't. Consider a one-line note that
  generalization is future work, or a name that signals the scope. (The doc's "Why this isn't all in the tripwire"
  already sets narrow expectations, so this is minor.)

  3. import.meta.dirname ?? Deno.cwd() fallback computes a wrong ROOT. check-facts.ts:35-36. If import.meta.dirname were
  ever undefined, ROOT = cwd/../.. points two levels above cwd and every read fails with noise rather than a clean error.
  It never fires for a local file run, so this is cosmetic — but the fallback is misleading rather than safe.

  Observation (not a defect)

  This is the same symbol-rot point raised in the second approval on #3829: the convertCellsToLinks row "names a bare
  symbol + path + internal helpers … the row #3863's checker is least likely to cover." Confirmed — the tripwire is
  existence/resolution-only: it checks that packages/runner/src/cell.ts exists, not that convertCellsToLinks is still
  exported from it, and that @commonfabric/data-model/value-hash resolves, not that the symbol behind it is intact. So
  #3863 doesn't close that note; it formalizes the split, delegating symbol/semantic rot to the LLM audit
  (skill-audit.md's "Why this isn't all in the tripwire"). That matches how you scoped it on #3829 ("a line for that
  follow-up's scope, not here") — flagging only that the convertCellsToLinks row remains uncovered by the deterministic
  floor and now relies entirely on the audit.

bfollington:
Thanks @mpsalisbury — all addressed (`843169f83`):

1. **Specifier regex (the real one).** Broadened the subpath segment to `[\w.-]` so PascalCase/dotted exports parse fully. Verified: `@commonfabric/data-model/BaseReconstructionContext` went from truncating to a bare `@commonfabric/data-model` (→ false "no root export") to matching in full and resolving (`./BaseReconstructionContext` is a real export). No trigger today, but the latent CI-false-positive class is gone. (The parse regex already captured the subpath via `(.+)`, so only the match regex needed it.)
2. **Scope naming (nit).** Added a comment at the hardcoded read: the checker gates the cf-review skill only; generalizing to all of `skills/**` is future work — so a future author isn't misled into thinking their new skill is gated.
3. **ROOT fallback (nit).** Replaced `import.meta.dirname ?? Deno.cwd()` with a clear error when run without a file path, instead of silently computing `cwd/../..`.

On the **observation** — agreed, and that's the intended design: the tripwire is existence/resolution-only, so symbol/semantic rot (including the `convertCellsToLinks` row) is delegated to the LLM audit, as documented in `skill-audit.md`'s "Why this isn't all in the tripwire." #3863 formalizes that split rather than closing the note.

🤖 Posted by Claude (Opus 4.8) on behalf of @bfollington

PR #3879, opened 2026-06-05
feat(ui): shared-profile identity components (cf-avatar, cf-profile-badge) + catalog

bfollington:
## Summary

First batch of shared-profile identity work: two reusable UI components, their catalog stories, and demo patterns.

- **`cf-avatar`** — generic avatar primitive: image / emoji-glyph / initials fallback; `xs`–`xl`; circle/square. Presentation-only, so sandboxed user-space patterns may render it.
- **`cf-profile-badge`** — trusted identity chip: bind a profile cell via `$profile` and it renders the person's avatar + name as system chrome. It runs on the trusted main thread (outside the iframe sandbox), so the chrome it emits can't be reproduced by sandboxed pattern code. v1 renders the "presented" state; the unspoofable verification seal (CFC `represents-principal` owner DID) is a **wired-but-deferred** pass (`_state` + `_refreshVerification`).

## Also in here

- JSX intrinsic element types (`packages/html/src/jsx.d.ts`) + v2 barrel registration.
- Catalog: new **Identity** category with Avatar + Profile Badge stories.
- **Catalog sidebar fix.** `CategoryRow`/`CategoryRowItem` were sub-patterns instantiated inside nested `.map()` with `onSelect` forwarded as a *pattern input*, which dropped the stream's `$stream` marker (`runner.ts:3094`, "Handler used as lift") and left the sidebar **empty on deploy**. Flattened to a single `rows` computed mapped once with an inline `onSelect.send()` — the canonical list idiom.
- Demo patterns composing the components: `profile-group-chat` (multiplayer chat keyed on shared-profile identity) and `shared-profile-roster` (`PerSpace` participant roster).
- Design docs under `docs/specs/`: migration shape (CT-1646), participant rosters (CT-1649), profile elements (CT-1651).

## Testing

- Unit tests for `cf-avatar` + `cf-profile-badge` (pure helpers, registration, defaults) — green.
- `cf check` typechecks the catalog and both demos.
- Verified live against local dev: sidebar renders all categories incl. Identity; avatar image/glyph/initials paths, all sizes, both shapes; badge resolves a profile cell to name + avatar + seal; 0 console errors.

## Follow-ups (intentionally not here)

- Profile-home **element-link CFC fix** + `home.tsx` badge wiring → **stacked PR** (touches #3830 files).
- Saving an avatar is blocked by a runtime CFC binding-identity bug → filed **CT-1665**.
[...]

PR #3880, opened 2026-06-05
fix(profiles): saving a profile element (element-link CFC label) + profile badge in home

bfollington:
## Summary

⚠️ **Stacked on #3879** (the `home.tsx` change needs `cf-profile-badge` from that PR). Review/merge #3879 first; the diff here is just the three `profile-home`-area files.

Two changes in `#3830`'s files:

### 1. Fix: saving a profile element ("Add profile card") is rejected by CFC

`ProfileElement.cell` is a cross-pattern link (`ProfileCatalogCard` / `UrlPatternReference`) written into the **owner-protected `elements`** array. Under `enforce-explicit`, CFC's `derivePersistedLinkLabel` (`prepare.ts`) couldn't derive a label for the freshly-created link and aborted the commit:

```
StorageTransactionAborted: CFC enforcement rejected commit: relevant transaction was not prepared:
missing link source metadata for of:fid1:… at /elements/0
```

**Fix:** give the element link a CFC integrity label (`addIntegrity: ["profile-element"]`) so its link-schema label satisfies the check (hatch 3 in `derivePersistedLinkLabel`) — mirroring how the profile link itself already carries `addIntegrity: ["profile-link"]`. Adds the missing `enforce-explicit` add-element-link test (**accept** + **reject-without-label**) to `profile-owner-cfc.test.ts`, which previously only ever set `elements: []`, so this path was untested.

### 2. `home.tsx`: show `<cf-profile-badge>` in the Profile tab

Renders the avatar + name from the profile cell. Kept the `profileName` text label alongside it so the existing `home-profile.test.ts` `waitForText` still passes. (This block is replaced by the picker in #3830 — small, rebase-friendly overlap.)

## Testing

- `packages/runner/test/profile-owner-cfc.test.ts`: **17/17 green**, including the two new element-link cases.
- `cf check` clean on `profile-home.tsx` and `home.tsx`.

## Not here

- Saving the `avatar` / `name` **string** fields is blocked by a separate runtime CFC binding-identity bug — filed **CT-1665** (runtime layer, not these files).

Relates to CT-1645 / CT-1651.

---
## Summary by cubic
Fixes CFC rejection when saving a profile element by labeling the element link via `PROFILE_ELEMENT_INTEGRITY`, and shows the profile badge in the Home Profile tab. Unblocks the “Add profile card” flow for CT-1645/CT-1651.

- **Bug Fixes**
  - Introduce `ProfileElementSchema` and apply `addIntegrity: ["profile-element"]` to its `cell` so cross-pattern element links are labeled; update `elements` to `ProfileElementSchema[]` to use the lowered schema.
  - Add tests covering accept and reject-without-label cases for element links.

- **New Features**
  - Render `<cf-profile-badge $profile={profile}>` in the Profile tab, keeping `profileName` for tests.

<sup>Written for commit 4ced6d5f8f0fbf9ef3f3fd09cc9498717d456233. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3880?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bfollington (commented):
_cf-review via Claude Opus 4.8, on behalf of @bfollington — report-first. The test-coverage gap below is **not** yet addressed; flagging for the author._

**Scope:** 3 files, +202/−19 (`profile-home.tsx`, `home.tsx`, `profile-owner-cfc.test.ts`), stacked on #3879. · **Coverage:** deep-read all 3 changed files + `prepare.ts` CFC label-derivation, `data-updating.ts` link-write emission, `cf-profile-badge`, the real transformed schema. · **Checks (verified):** `profile-owner-cfc.test.ts` → 17 steps pass; `cf check` on both patterns → exit 0; `deno fmt --check`/`deno lint` → clean; `--show-transformed` on `profile-home.tsx`.

The change is **correct and does not weaken enforcement.** The `addIntegrity: ["profile-element"]` label lives purely in the labeling-derivation path (`derivePersistedLinkLabel` hatch 3), not the authorization path. Owner-write authorization for `elements` remains gated independently by `ownerPrincipal` + `writeAuthorizedBy` + the owner atom on the *array* schema (unchanged), and the pre-existing negative tests (`rejects Bob` / `untrusted` / `unauthenticated` / `ownerPrincipal-without-integrity`) still pass. A non-owner cannot use this atom to write `elements`. **No blocking findings.**

### 🟡 Improvement — the new tests hand-model the link instead of exercising the real lowering + handler path (suspected coverage gap)

`profile-owner-cfc.test.ts:48-61` (`elementLinkSchema`) + the two new cases (`:236`, `:303`).
- **What:** both new tests build a hand-rolled `elementLinkSchema` and write it via `cell.key("elements").set([elementLink])`. No test compiles `profile-home.tsx` and runs the real `addElement`/`addCatalogElement` handler (which does `ProfileCatalogCard(...).for(tag)`).
- **Why it matters:** the hand model is *structurally different* from the real generated schema. `--show-transformed` confirms production lowers the atom onto the **`cell` property** (`asCell:["cell"]` + `ifc.addIntegrity`, link at `/elements/{i}/cell`), whereas the test puts it on the **element-object root** (link at `/elements/0`). And at runtime `linkSchema = source.schema` (the *parsed link's* schema, `data-updating.ts:203`), not the consumer-side `cell`-property schema. So the tests guard "an atom at the link-schema root satisfies hatch 3" — real and useful — but do **not** prove the consumer-side `cell`-property atom actually propagates into `source.schema` for the real `.for()` link write. If lowering / link-schema propagation diverged from the author's mental model, both new tests could stay green while saving a card still fails.
- **Fix:** add one test that compiles `profile-home.tsx`, runs `addElement` through the runner under the trusted profile-writer context, and asserts `commit()` succeeds — i.e. reproduce the actual "Add profile card" path. The negative twin (`:303`) can stay as-is; it cleanly locks the "atom is load-bearing" principle. _structural mismatch verified via transformed output; regression-unguarded is suspected._

**Principle each new test guards:** `:236` — *a trusted writer pushing a labeled cross-pattern link into the owner-protected `elements` array must commit under enforce-explicit* (real link write + commit, but on the hand-modeled schema); `:303` — *the integrity atom is load-bearing; drop it and CFC fails to derive a label, aborting with the exact production error* (good negative twin).

### ⚪ Nits
- `profile-home.tsx` NOTE(CT-1628) + new block comment are accurate (runtime CFC schema for `cell` comes from `ProfileElementSchema`; value type keeps `cell: any`). No coherence defect. _verified_
- `home.tsx` `<cf-profile-badge $profile={profile as any}>` — consistent with the adjacent `<cf-render $cell={profile as any}>`; the badge is read-only display (no write-back), so the kept `<strong>{profileName}</strong>` mirror creates no reactive self-feedback. _verified_

### Questions for the author
1. Has the **real** "Add profile card" flow been exercised end-to-end (browser or runner) on this branch after the fix — i.e. confirmed the consumer-side `cell`-property atom actually reaches `linkSchema` for the `.for()` write? That's the one link in the chain the new tests don't cover.

### Possible follow-ups
- Add an end-to-end `addElement` regression test (closes the gap; also catches future schema-generator drift in how `Cfc<Cell<…>, {addIntegrity}>` lowers).
- CT-1628: the recurring `cell: any` + `(Pattern(…) as any).for(…)` casts are now load-bearing for CFC labeling — worth the typed cell-ref so `ProfileElement`/`ProfileElementSchema` collapse into one.

**Verdict:** Approve with a non-blocking test-coverage ask — correct, security-preserving, green; but lock the regression with a real end-to-end `addElement` test, since the current tests guard the principle via a schema that differs structurally from the real lowering path.

bfollington:
Closing this PR — parking the element-link "save a profile card" work.

**Why:** I verified the real flow end-to-end with a headless browser (driving the actual "Add profile card" button, both via the home Profile tab and via ProfileHome instantiated as a top-level piece). Findings, with instrumentation:

- The click **is** delivered to the handler (proven: stripping the `elements` write out of `addCatalogElement`, a debug counter in the same handler incremented to `1`).
- With the real `elements` write in place, the handler transaction **silently aborts** — the counter rolls back to `0`, no element persists, and **no error surfaces** to the page console even at `LOG_LEVEL=debug`. It's a CFC policy rejection the scheduler swallows.

So the `addIntegrity` change here clears the *first* barrier (the original `missing link source metadata … at /elements/0` link-label error) but the write then hits a **second, silent runtime barrier** and the feature still doesn't work. This is the same class of runtime CFC gap as **CT-1665** (the `writeAuthorizedBy` "trusted verified binding identity" check that already blocks `avatar` saves) and/or the link's `source.schema` not carrying the atom — neither cleanly fixable in pattern code. The unit tests here pass only because they hand-model the write in a way that bypasses the real lowering path (exactly as the cf-review on this PR suspected).

**What's salvaged:** the home `<cf-profile-badge>` wiring (the only part that works today) has moved to #3882, stacked on #3879.

**Tracking:** filing a runtime issue (sibling to CT-1665) with the repro evidence so the owner-protected element-link write can be fixed at the runtime/CFC layer; this PR can be revived once that lands.

PR #3885, opened 2026-06-05
feat(lib-shell): factor shell runtime primitives

jkomoros:
## What this is

Adds an importable `@commonfabric/lib-shell` package for the first library factoring step. This keeps the first-pixel scope to single-space runtime hosting primitives only.

## Scope

- Adds `packages/lib-shell` with exported runtime, credential, and space-base helpers.
- Moves `RuntimeInternals` and stored credential helpers behind `@commonfabric/lib-shell`.
- Leaves shell render chrome, `RootView` lifecycle orchestration, `AppView` pattern resolution, and dispose/space-switch behavior in `packages/shell`.
- Keeps shell-local `src/lib/runtime.ts` and `src/lib/credentials.ts` as compatibility re-export shims.
- Adds host seams for `navigate`, console/error callbacks, worker URL, build hash, experimental flags, CFC enforcement mode, trust snapshot, and shell base-href construction.

## Review notes addressed

- `lib-shell.identity` logging now defaults off.
- CFC enforcement and trust snapshot are host-configurable, with shell-compatible defaults.
- Shell space-root assumptions for `trackRecentPiece` / `registerNavigatedPiece` are documented at the call sites.
- No L2/L3 work: `RootView` dispose/respawn predicate and session space-key derivation remain untouched.
- cubic identified runtime and credential hygiene issues on the previous commit; those are fixed in the latest head.

## Verification

- `deno check packages/lib-shell/src/index.ts packages/shell/src/views/RootView.ts packages/lib-shell/test/*.test.ts`
- `deno task test` in `packages/lib-shell`
- `deno task test` in `packages/shell`
- `rg --files packages/shell/src -g "*.ts" | xargs deno check`

## Follow-up

seefeldb approved this PR and noted that the base-href approach should be replaced later: audit relative links and make them absolute so `lib-shell` consumers do not need a shell-level base href assumption.


seefeldb (approved):
lgtm,

a note for later changes: the base href approach will break once we allow more than one server URL, and it might also break any assumptions someone using libshell makes.

we might want to audit where those relative links are coming from and make them absolute instead. separate PR, whose goal should be that we don't need that base href anymore.

PR #3887, opened 2026-06-05
fix(ts-transformers): normalize commonfabric refs in synthesized lift/pattern result types

mathpirate:
## What

Synthesized `lift<In, Out>(...)` / `pattern<…, Out>(...)` calls printed their **result** type argument using TypeScript's inline import-type form, e.g. `import("commonfabric").JSXElement`, instead of the canonical `__cfHelpers.X` form. The runner exposes `__cfHelpers` as a self-reference to the whole commonfabric namespace, so `__cfHelpers.X` always resolves; the inline `import(...)` form does not reliably typecheck against the user's imports.

Spotted in the transformed output of `packages/patterns/lunch-poll` (5 occurrences of `import("commonfabric").JSXElement`).

While fixing that, the same chokepoint-migration work surfaced **two related, independently-reviewable bugs** that ride along in this PR — a schema-generator `$ref`-drop on capability re-wraps, and a downstream nested-cell existence check in `home.tsx`. Both are called out under **Also in this PR** below.

## Root cause

A correct normalizer already exists — `qualifyCommonFabricTypeRefs` (wrapped by the `typeToTypeNodeWithRegistry` chokepoint) — which rewrites commonfabric refs to `__cfHelpers.X` and registers the node's Type for schema generation. The lift **input** type arg routed through it; the **result** type arg paths called `checker.typeToTypeNode` directly, bypassing it. That asymmetry was the bug.

## Fix

- Route the lift/pattern result-type paths through the chokepoint:
  - `transformers/builtins/lift-applied.ts` (`buildResultTypeNode`)
  - `closures/strategies/lift-applied-strategy.ts` (signature-inferred result **and** the `callback.type` propagation path, found via instrumentation — a synthesized annotation can carry a raw import node)
  - `closures/strategies/array-method-transform.ts` (mapWithPattern result)
- Extend the migration to the remaining emit-reaching `checker.typeToTypeNode` sites so the chokepoint covers them all: `transformers/type-shrinking.ts` (7 shrink-result conversions), `transformers/expression-rewrite/rewrite-helpers.ts` (compute-wrapper arrow return), and `closures/utils/schema-factory.ts` (event parameter type). Three raw calls remain, deliberately allowlisted (non-emitting / load-bearing undefined return / inspected-not-emitted).
- Fix a latent normalizer bug this surfaced (`ast/type-building.ts`): union/intersection members weren't qualified (`(VNode | UIRenderable)[]` came out bare). Now matched to constituent Types **by export name**, and — per cubic review — requiring an **unambiguous** match (two constituents sharing an export name but differing in type args are left unpaired rather than mis-rewritten).
- Don't collapse **user aliases** to bare commonfabric names: a non-generic user alias whose underlying type is a commonfabric generic (`type SharedMessagesCell = Writable<SharedMessagesValue>`) was being rewritten to bare `__cfHelpers.Cell`, dropping the type argument. Only rewrite when the node's own name is a commonfabric export.
- Harden the normalizer to **self-register** the rewritten node's Type — removes a footgun where a missing registry entry silently degrades the generated schema (precise type → permissive `true`).
- Promote `typeToTypeNodeWithRegistry` + `qualifyCommonFabricTypeRefs` into the `ast/mod.ts` barrel as the canonical entry.

## Also in this PR

Two behavior fixes outside the type-normalizer, surfaced by the chokepoint migration:

- **schema-generator: preserve inner `$ref` when a capability re-wrap narrows a cell's brand** (`schema-generator/src/formatters/common-fabric-formatter.ts`, `type-utils.ts`). When type-shrinking re-wraps `Cell<T>` as `ReadonlyCell<T>` for read-only usage, the node's brand no longer matches the resolved type's brand, which routed schema generation through the node-driven path. For a bare named-reference inner with no source position the checker returns `any`, so the inner schema collapsed to `{}` and the `$ref`/`$defs` were dropped — the runtime then failed to materialize nested cells (`profile.get is not a function` in cfc-group-chat-demo). Fix: when a synthetic node narrows a **cell-capability** brand and its own inner degrades to `any`, fall back to the resolved type's inner so the `$ref` survives while keeping the narrowed brand. Also refactors `CELL_CAPABILITY_KINDS` to an exhaustive `Record<CellWrapperKind, boolean>` so a new kind is a compile error until classified (no behavior change).
- **`packages/patterns/system/home.tsx`: read profile existence at the right cell depth.** `profile` is a `Cell` whose value is itself a `Cell` (`TrustedProfileLink` wraps `Cell<ProfileHomeOutput>`), so `profile.get()` returns the inner cell **handle** — always non-undefined. The existence check `profile.get() !== undefined` was therefore permanently true, hiding the profile-create surface. Now reads one level deeper: `profile.get()?.get() !== undefined`. This was masked until the schema-gen fix above made the nested-cell schema honest.

## Guards added

- **Golden invariant** (`test/no-import-type-in-output.test.ts`): no `*.expected.*` may contain `import("commonfabric")`. Covers every transform path at once.
- **Static call-site sweep** (`test/typetotypenode-call-sites.test.ts`): raw `checker.typeToTypeNode` sites must match a reviewed allowlist with per-site justification; new bypasses fail. Documents that this producer set is shared by two bug classes (import-leak + registry-degradation) and marks `[emit+concrete]` sites still pending migration to the chokepoint.
- **Minimal fixture** (`lift-result-type-cf-ref-normalized`): a JSX-returning computed → `lift<{count}, __cfHelpers.JSXElement>`, a legible regression signal.
- **Behavior-lock fixture** (`readonly-capture-named-alias-nested-cell`): pins that a Cell→ReadonlyCell capability narrowing preserves the inner `$ref` + nested cell in `$defs` (the genuine fail-without/pass-with guard for the schema-gen fix is the cfc-group-chat-demo pattern test).

## Result

Zero `import("commonfabric")` in any fixture output. `deno task test` → 292 passed / 0 failed; lint clean. Full pattern integration suite 19/0; schema-generator 24/0.

## Notes
- A few goldens additionally gain more-correct output schemas once the result type is a real `__cfHelpers.X` rather than an opaque import-type: `map-conditional-inline-handler-send` (`true` → `{ type: "unknown", asCell: ["stream"] }` — the Stream is now recognized) and `map-capture-mixed-reactivity`'s read-only `limit` capture (`asCell: ["cell"]` → `asCell: ["readonly"]`).
- The two refreshed closure goldens (`lift-result-type-cf-ref-normalized`, `readonly-capture-named-alias-nested-cell`) gain the trailing `__cfReg({ … })` hoist-registration block introduced by CT-1623 (#3912), which landed on main while this PR was open. Purely additive; matches every other closure golden.
- Remaining `[emit+concrete]` allowlist sites (type-shrinking cluster, rewrite-helpers, schema-factory, type-inference) are tracked for a follow-up migration to the chokepoint.

---
## Summary by cubic
Normalizes all synthesized type annotations to the canonical `__cfHelpers.X` form and routes every emit-reaching type→TypeNode conversion through a single chokepoint. Preserves inner `$ref` in capability re‑wraps, tightens schema hints, and fixes a nested‑cell existence check in Home.

- **Bug Fixes**
  - Route all emit-reaching conversions through `typeToTypeNodeWithRegistry` and normalize CF refs: `lift-applied`, `lift-applied-strategy`, `array-method-transform`, expression-rewrite helpers, schema-factory, and `type-shrinking` emit paths (three raw calls remain allowlisted and non‑emitting).
  - Harden the normalizer: handle explicit callback return annotations; auto-register rewritten nodes; require unique union/intersection member pairing by export name; preserve user-defined aliases (don’t collapse non‑`commonfabric` `aliasSymbol`s).
  - Schema generation: when a capability re‑wrap narrows a cell brand (e.g. `Cell<T>` → `ReadonlyCell<T>`), keep the inner `$ref`/`$defs`; if the node’s inner degrades to `any`, fall back to the resolved type’s inner. Improves annotations (e.g. `asCell: ["readonly"]`, Streams recognized).
  - Guards: invariant banning `import("commonfabric")` in all goldens; static sweep that pins raw `checker.typeToTypeNode` sites to a reviewed allowlist; add JSX normalization and nested‑cell behavior‑lock fixtures.
  - `packages/patterns/system/home.tsx`: read profile existence one level deeper (`profile.get()?.get()`), restoring the create UI when no profile exists.

- **Refactors**
  - Derive cell capability kinds exhaustively from `CellWrapperKind` in `@commonfabric/schema-generator`; no behavior change.
  - Promote `qualifyCommonFabricTypeRefs`/`typeToTypeNodeWithRegistry` via `ast/mod.ts`; README shows a real `--show-transformed` example and the `--no-run` flag.
  - Refresh two closure goldens to include the trailing `__cfReg` hoist registration for consistency.

<sup>Written for commit 42ee029780313f9955797e75dccb7e6b5308f00c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3887?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/patterns/system/home.tsx
+++ b/packages/patterns/system/home.tsx
@@ -165,8 +165,14 @@ export default pattern<Record<string, never>, HomeOutput>((_) => {
   // does not re-show the create form (which would let it overwrite a valid
   // link). The `cf-render` below resolves and loads the cross-space profile for
   // display.
+  // `profile` is a Cell whose value is itself a Cell (TrustedProfileLink wraps
+  // Cell<ProfileHomeOutput>), so `profile.get()` returns the inner cell HANDLE,
+  // which is always non-undefined. Read one level deeper — `profile.get()?.get()`
+  // — to test whether the inner cell actually holds a value (i.e. a profile
+  // exists). Using `profile.get() !== undefined` here is always true and would
+  // permanently hide the create surface.
   const hasProfile = computed(() =>
-    profile.get() !== undefined ||
+    profile.get()?.get() !== undefined ||
     (profileName.get() ?? "").trim().length > 0
   );
 
--- a/packages/schema-generator/src/formatters/common-fabric-formatter.ts
+++ b/packages/schema-generator/src/formatters/common-fabric-formatter.ts
@@ -49,19 +49,33 @@ const scopeForWrapperName = (
 ): SchemaScope | undefined =>
   name === undefined ? undefined : SCOPE_WRAPPER_SCOPES[name];
 
-// Cell-capability brands all wrap the SAME structural inner `T`; they differ
-// only in read/write capability. The transformer narrows one to another (e.g.
-// `Cell<T>` → `ReadonlyCell<T>`) to reflect usage, so a node-vs-type brand
-// mismatch among these is a capability narrowing, not a structural change.
-const CELL_CAPABILITY_KINDS: ReadonlySet<WrapperKind> = new Set([
-  "Cell",
-  "ReadonlyCell",
-  "WriteonlyCell",
-  "ComparableCell",
-  "OpaqueCell",
-]);
+// The capability subset of `CellWrapperKind`: brands that all wrap the SAME
+// structural inner `T` and differ only in read/write capability. The transformer
+// narrows one to another (e.g. `Cell<T>` → `ReadonlyCell<T>`) to reflect usage,
+// so a node-vs-type brand mismatch among these is a capability narrowing, not a
+// structural change. `Stream`/`SqliteDb`/`OpaqueRef` are excluded: they carry a
+// distinct structural contract, not a read/write variant of a plain cell.
+//
+// Derived as an exhaustive map over `CellWrapperKind` so that adding a new kind
+// to that union is a compile error here until it's deliberately classified.
+//
+// NB: distinct from `type-utils.ts`'s `CELL_LIKE_WRAPPER_NAMES`, which keys off
+// raw type-node NAMES (where `Writable` is a separate spelling and `OpaqueCell`
+// is split out). This set keys off RESOLVED `CellWrapperKind` values, where
+// `Writable` has already normalized to `Cell` and `OpaqueCell` belongs with the
+// rest.
+const CELL_CAPABILITY_KIND_MAP: Readonly<Record<CellWrapperKind, boolean>> = {
+  Cell: true,
+  ReadonlyCell: true,
+  WriteonlyCell: true,
+  ComparableCell: true,
+  OpaqueCell: true,
+  Stream: false,
+  SqliteDb: false,
+  OpaqueRef: false,
+};
 const isCellCapabilityKind = (kind: WrapperKind): boolean =>
-  CELL_CAPABILITY_KINDS.has(kind);
+  CELL_CAPABILITY_KIND_MAP[kind];
 
 const resolveScopeWrapperNode = (
   typeNode: ts.TypeNode | undefined,
--- a/packages/schema-generator/src/formatters/common-fabric-formatter.ts
+++ b/packages/schema-generator/src/formatters/common-fabric-formatter.ts
@@ -49,6 +49,20 @@ const scopeForWrapperName = (
 ): SchemaScope | undefined =>
   name === undefined ? undefined : SCOPE_WRAPPER_SCOPES[name];
 
+// Cell-capability brands all wrap the SAME structural inner `T`; they differ
+// only in read/write capability. The transformer narrows one to another (e.g.
+// `Cell<T>` → `ReadonlyCell<T>`) to reflect usage, so a node-vs-type brand
+// mismatch among these is a capability narrowing, not a structural change.
+const CELL_CAPABILITY_KINDS: ReadonlySet<WrapperKind> = new Set([
+  "Cell",
+  "ReadonlyCell",
+  "WriteonlyCell",
+  "ComparableCell",
+  "OpaqueCell",
+]);
+const isCellCapabilityKind = (kind: WrapperKind): boolean =>
+  CELL_CAPABILITY_KINDS.has(kind);
+
 const resolveScopeWrapperNode = (
   typeNode: ts.TypeNode | undefined,
 ): ResolvedScopeWrapper | undefined => {
@@ -272,6 +286,14 @@ export class CommonFabricFormatter implements TypeFormatter {
         resolvedWrapper.node,
         context,
         resolvedWrapper.kind,
+        // The synthetic node narrows the resolved type's capability brand (e.g.
+        // the transformer re-wrapped `Cell<T>` as `ReadonlyCell<T>` for read-only
+        // usage). Both brands wrap the SAME structural inner. When the node's own
+        // inner has no source position and degrades to `any`, fall back to the
+        // resolved type's inner so the inner `$ref`/`$defs` survives the re-wrap.
+        isCellCapabilityKind(wrapperInfo.kind)
+          ? wrapperInfo.typeRef
+          : undefined,
       );
     }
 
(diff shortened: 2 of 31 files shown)


mathpirate:
## @seefeld — review ask: nested-cell `$ref` preservation + the home.tsx consequence

Two related changes here need your blessing, plus a runtime footgun I think is worth a separate conversation.

### 1. Schema generator: preserve inner `$ref` when a capability re-wrap narrows a cell's brand (`3e8f165e5`)

When a capture is read-only, type-shrinking narrows its brand `Cell<T>` → `ReadonlyCell<T>` (existing behavior). The schema generator was then formatting it node-driven, and for a bare named-ref inner (e.g. `SharedProfilesValue`, no source position) the checker resolved it to `any` — so the inner `$ref`/`$defs` were dropped:

```jsonc
// before (broken): inner structure gone, nested cells not materialized -> runtime "profile.get is not a function"
profiles: { asCell: ["readonly"] }

// after (fixed): precise inner preserved, brand still readonly
profiles: { $ref: "#/$defs/SharedProfilesValue", asCell: ["readonly"] }
```

The fix falls back to the resolved *type's* inner (which carries the precise ref) while keeping the node's narrowed brand. Scoped to cell-capability brands only, and only for bare `TypeReferenceNode` inners (so structured inners like `string | undefined`, whose shape lives only on the node, are untouched). **Zero golden changes** in the ts-transformers suite.

**Confirm:** that `{ $ref: …, asCell: ["readonly"] }` is the desired schema for a read-only-used by-reference cell capture (precise inner + readonly brand) — this is the "readonly vs cell" question from the type-shrinking commit; the answer is "readonly is right, AND keep the `$ref`."

### 2. Downstream impact on home.tsx (`4ac477772`)

`profile` is `Writable<TrustedProfileLink>` where `TrustedProfileLink` wraps `Cell<ProfileHomeOutput>` — i.e. a **Cell whose value is itself a Cell** (`Cell<Cell<X>>`). Once change #1 made the schema honest about that nested cell, this existence check broke:

```ts
// always-true now: profile.get() returns the inner cell HANDLE (never undefined)
const hasProfile = computed(() => profile.get() !== undefined || …);
```

`hasProfile` was permanently true → the profile-create surface (`#home-profile-name-input`) was hidden → `home-profile.test.ts` timed out. Fixed by reading one level deeper:

```ts
const hasProfile = computed(() => profile.get()?.get() !== undefined || …);
//  profile.get()        = inner cell handle (navigation / write)
//  profile.get()?.get() = inner value (existence)
```

Verified end-to-end (rebuilt toolshed): `home-profile.test.ts` passes; full pattern integration suite 19/0; ts-transformers 292/0; schema-generator 24/0.

Alternatives I rejected: (a) change the runtime so unset nested cells read as `undefined` — reverses the tested write-before-set "return a handle" contract (`schema-basic.test.ts` "preserve types through key()"), a system-wide change; (b) narrow the schema-gen fix to skip cell-typed inners — re-introduces the under-specification (schema would lie).

**Confirm:** fixing the consumer's read depth (not the runtime, not narrowing the schema fix) is the right call.

### 3. Footgun (separate decision, not blocking this PR — your call on the runtime side)

`someNestedCell.get() !== undefined` is **silently always true** for any `Cell<Cell<X>>`-shaped field, because `.get()` returns the inner cell handle regardless of whether a value is set (intentional, for write-before-set). The correct existence idiom is `.get()?.get() !== undefined`. This will bite other authors. Worth a guard? Options: a lint rule, or an explicit `.exists()` / `.isEmpty()` cell helper. Flagging for a separate decision you own.

(Full writeup with mechanism + probe results: `session_outputs/2026-06-04_ts-transformers-cleanup/14_BERNI-review-schema-ref-and-home-profile.md`.)

PR #3900, opened 2026-06-06
feat(data-model): per-class FabricCodec wire codecs (replace TypeHandler)

danfuzz:
Replaces the `json-wire` `TypeHandler` model with a `FabricCodec` model and
moves each type's wire codec onto the type itself.

## Why

Under the old model a `TypeHandler` did its own tag-wrapping **and** recursion,
and the set of wire-serializable types was implicit (a generic
`FabricInstanceHandler` plus a separate tag→class registry). The new model makes
codecs **shallow** (the `JsonEncodingContext` drives recursion + tag-wrapping)
and makes the wire-format surface **explicit and curated**.

## What

- **`FabricCodec`** (`wire-common`): `encode(value)` / `decode(tag, state, context)`
  are shallow; `uniqueHandledClass` + `wireTypeTag` drive registry dispatch.
  `BaseFabricCodec` supplies the common scaffolding and an `instanceof`-based
  `canEncode`.
- **Per-class `[CODEC]`**: each fabric type with a fixed wire tag hosts its codec
  as a static `[CODEC]` getter — `FabricBytes`, `FabricEpochNsec`,
  `FabricEpochDays`, `FabricRegExp`, **`FabricHash`** (previously missing — a
  long-standing gap), `FabricError`, `FabricMap`, `FabricSet`. All `#codec`
  singletons are frozen.
- **Curated lists**: `fabric-primitives` and `fabric-instances` each export
  `codecClasses()` (a frozen array of their wire-available classes).
  `createDefaultRegistry` iterates those instead of importing every class — so
  *which types are wire-available* is curated in one obvious place per area.
- **JS primitives** (`bigint`, special `number`, interned `symbol`, `undefined`)
  have no owned class, so they remain standalone `BaseFabricCodec`s
  (`BigIntCodec` / `SpecialNumberCodec` / `SymbolCodec` / `UndefinedCodec`).
- **`FabricInstanceHandler` is gone.** Concrete instance types are per-class
  codecs; `ExplicitTagValue` / `UnknownValue` / `ProblematicValue` (live-graph
  stand-ins carrying a per-instance tag) are handled by a dedicated
  `serialize()` arm that re-emits their preserved tag + state. An un-codec'd
  `FabricInstance` reaching encode is now a **hard error** — every wire form is
  explicitly represented.
- **Class registry retired**: decode dispatches `Error`/`Map`/`Set` through their
  codecs; unknown tags fall straight to `UnknownValue`.
- **Deserializer recursion fix**: the codec arm now recurses the wire state once,
  centrally (with `/quote` literal and `/object` escaping handled ahead of it).

## Naming / hygiene

`TypeHandlerRegistry` → `CodecRegistry`; the `*Handler` codec files → `*Codec`;
the decode parameter `runtime` → `context` (matching `FabricCodec.decode`); the
private tree-walkers `serialize`/`deserialize` → true-private `#encode`/`#decode`
(the public `encode`/`decode` boundary keeps those names).

## Tests

`deno task check` clean; data-model suite **31 passed / 1621 steps / 0 failed**.
`CodecRegistry` tests rewritten for the new API and expanded (fast/slow-path
`codecFromValue`, `codecFromTag`); added `FabricRegExp` + `FabricHash` coverage
and a mandate-guard test in `JsonEncodingContext.test.ts`.

A follow-up is in progress to move each class's *primary* codec tests into that
class's own unit-test file (under `static members` → `[CODEC]`), leaving
`JsonEncodingContext` to exercise the mechanism rather than every type.

## Related fix (land first)

This branch's strict encode guard surfaced a latent fragility: the session-open
`authorization.signature` is a raw `Uint8Array` that only survived the memory
wire boundary by accident (the lenient structural fallback flattened it into a
numeric-keyed object that the server's `toByteArray` happened to accept). The
strict guard correctly refuses a raw `Uint8Array`. **#3902** is the "expand"
step of a staged migration toward a `FabricBytes` signature: it teaches the
consumer to accept `FabricBytes` and makes the producer emit an explicit number
array (a plain `FabricValue` that passes the strict guard). Land #3902 first;
this branch then picks it up via `main`. (Earlier attempt #3901 — which jumped
straight to producing `FabricBytes` — was wrong and is closed.)

---
## Summary by cubic
Replaces the old JSON wire `TypeHandler` model with shallow, per-class `FabricCodec`s and a `CodecRegistry`. Adds a `FabricHash` codec and a strict encode guard that rejects unhandled or non-plain values with clear, compact errors.

- **New Features**
  - Static `[CODEC]` on `FabricBytes`, `FabricEpochNsec`, `FabricEpochDays`, `FabricRegExp`, `FabricHash`, `FabricError`, `FabricMap`, `FabricSet` (frozen singletons).
  - Standalone codecs for `bigint`, special `number`s, interned `symbol`, and `undefined`; registration is curated via `codecClasses()` exports.
  - Exported `CODEC` from `wire-common` so `FabricClassWithCodec` is usable externally.
  - `FabricHash` codec state is `{ tag, hash }`; malformed base64 in `hash` decodes to `ProblematicValue`.
  - Strict encode guard throws for values with no codec/arm (e.g., unique `symbol`, non-plain objects like raw `Map`/`Date`); errors now uniformly use `toCompactDebugString(value, 50)`.

- **Migration**
  - Replace `TypeHandlerRegistry` with `CodecRegistry`; use `createDefaultRegistry()` which registers class `[CODEC]`s and standalone codecs.
  - Replace `*Handler` imports with `*Codec`; recursion is driven by `JsonEncodingContext`.
  - Add a static `[CODEC]` to each wire-serializable `FabricInstance` and include the class in its area's `codecClasses()`.
  - Unknown tags decode to `UnknownValue`. `UndefinedCodec.decode` rejects non-`null` state.

<sup>Written for commit 60605a168286241bfab4862eabbc62bfbfdef562. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3900?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/data-model/test/fabric-primitives/FabricHash.test.ts
```diff
@@ -146,5 +150,65 @@ describe("FabricHash", () => {
         );
       });
     });
+
+    describe("[CODEC]", () => {
+      const codec = FabricHash[CODEC];
+      const context = new EmptyReconstructionContext(
```
danfuzz: I think it's okay to just use the prefab instance. If there is a `throw` it will be obvious where it came from without having to make a special message.

packages/data-model/test/fabric-primitives/FabricHash.test.ts
```diff
@@ -146,5 +150,65 @@ describe("FabricHash", () => {
...
+        true,
+        "no-op (FabricHash codec test).",
+      );
+
+      it("has the `Hash@1` wire type tag", () => {
+        expect(codec.wireTypeTag).toBe("Hash@1");
```
danfuzz: Rather than un-DRY-ly recapitulate the tag here, I think the right thing to do is validate that the tag matches the one defined in `WIRE_TYPE_TAGS` per se (whatever the value happens to be).

packages/data-model/src/fabric-instances/FabricMap.ts
```diff
@@ -75,4 +78,31 @@ export class FabricMap
...
+
+  static #codec = Object.freeze(
+    new (class FabricMapCodec extends BaseFabricCodec {
+      constructor() {
+        super(WIRE_TYPE_TAGS.Map, FabricMap);
+      }
+
+      /** @inheritDoc */
+      encode(value: FabricMap): FabricValue {
+        return value[DECONSTRUCT]();
```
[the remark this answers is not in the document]
danfuzz: That is a correct observation and intentional.

packages/data-model/src/fabric-instances/FabricSet.ts
```diff
@@ -74,4 +77,31 @@ export class FabricSet extends FabricNativeWrapper<Set<FabricValue>> {
...
+
+  static #codec = Object.freeze(
+    new (class FabricSetCodec extends BaseFabricCodec {
+      constructor() {
+        super(WIRE_TYPE_TAGS.Set, FabricSet);
+      }
+
+      /** @inheritDoc */
+      encode(value: FabricSet): FabricValue {
+        return value[DECONSTRUCT]();
```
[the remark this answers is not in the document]
danfuzz: That is a correct observation and intentional. This project is incomplete and you have spotted an incompleteness.

packages/data-model/src/fabric-primitives/FabricRegExp.ts
```diff
@@ -116,6 +124,62 @@ export class FabricRegExp extends BaseFabricPrimitive {
...
+          return new ProblematicValue(
+            wireTypeTag,
+            state,
+            `RegExp: expected object state, got ${typeof state}`,
+          );
+        }
+        const s = state as Record<string, unknown>;
+        const flavor = (s.flavor as string) ?? DEFAULT_FLAVOR;
+        const source = (s.source as string) ?? "";
+        const flags = (s.flags as string) ?? "";
```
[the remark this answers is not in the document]
danfuzz: This is intentional for now, but you are correct to flag it. However, the right thing is to ensure that we document that the class does not enforce syntax as part of its participation in wire protocols.

packages/data-model/src/json-wire/SymbolCodec.ts
```diff
@@ -0,0 +1,55 @@
...
+ * confirms via `typeof`.
+ */
+export class SymbolCodec extends BaseFabricCodec {
+  constructor() {
+    super(WIRE_TYPE_TAGS.Symbol, Symbol as unknown as Constructor);
+  }
+
+  /** @inheritDoc */
+  override canEncode(value: FabricValue): boolean {
+    return typeof value === "symbol" && Symbol.keyFor(value) !== undefined;
```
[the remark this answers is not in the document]
danfuzz: Good point in general, but it isn't the place for _this class_ to `throw` on uninterned symbols. This class is correct to simply say "no" in `canEncode()`.

What we should have is a unit test case for a default-configured `JsonEncodingContext` that demonstrates that a uninterned (unique) symbol will loudly fail to get encoded.

packages/data-model/src/wire-common/interface.ts
```diff
@@ -1,19 +1,23 @@
...
 import type { FabricInstance, FabricValue } from "@/interface.ts";
 
 /**
- * Well-known symbol for deconstructing a fabric object into its essential
- * state. The returned value may contain nested `FabricValue`s (including
- * other fabric objects); the serialization system handles recursion.
- * See Section 2.2 of the formal spec.
+ * Well-known symbol for binding the getter `FabricClassWithCodec[CODEC]`.
+ */
+export const CODEC: unique symbol = Symbol.for("data-model.codec");
```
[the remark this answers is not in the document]
danfuzz: Good catch!

packages/data-model/src/wire-common/interface.ts
```diff
@@ -1,19 +1,23 @@
...
+
+/**
+ * Well-known symbol for binding the method
+ * `FabricDeconstructable[DECONSTRUCT]`.
  */
-export const DECONSTRUCT: unique symbol = Symbol.for("common.deconstruct");
+export const DECONSTRUCT: unique symbol = Symbol.for("data-model.deconstruct");
```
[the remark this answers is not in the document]
danfuzz: Good catch, thanks! However, in this stage of this particular package's development, we have a practice of iterating on the implementation and then retconning the spec. You can expect a follow-on PR to this effect, at some point (within a week or so).

packages/data-model/src/json-wire/UndefinedCodec.ts
```diff
@@ -0,0 +1,34 @@
...
+    return null;
+  }
+
+  /** @inheritDoc */
+  decode(
+    _wireTypeTag: string,
+    _state: FabricValue,
+    _context: ReconstructionContext,
+  ): FabricValue {
+    return undefined;
```
danfuzz, line 37: Suggest `throw`ing given a non-`null` state.

packages/data-model/src/wire-common/BaseFabricCodec.ts
```diff
@@ -0,0 +1,61 @@
+/**
+ * Empty `ReconstructionContext`: a singleton whose `getCell()` always throws.
```
danfuzz: Looks like this doc comment got copy-pasta'ed and then forgotten. It's safe to remove the entire comment. The class doc comment suffices.

packages/data-model/src/fabric-primitives/FabricHash.ts
```diff
@@ -142,6 +150,49 @@ export class FabricHash extends BaseFabricPrimitive implements ApiFabricHash {
...
+
+  static #codec = Object.freeze(
+    new (class HashCodec extends BaseFabricCodec {
+      constructor() {
+        super(WIRE_TYPE_TAGS.Hash, FabricHash);
+      }
+
+      /** @inheritDoc */
+      encode(value: FabricHash): FabricValue {
+        // The `<tag>:<base64urlHash>` string form round-trips via `fromString()`.
```
danfuzz: Suggest keeping the two parts as separate in the state, as `{ tag: string, hash: string }`.

packages/data-model/src/json-wire/JsonEncodingContext.ts
```diff
@@ -257,43 +220,72 @@ export class JsonEncodingContext implements SerializationContext<string> {
...
   //
 
   /**
-   * Serializes a fabric value into wire format. Recursively processes nested
-   * values. See Section 4.5 of the formal spec.
+   * Encodes a fabric value into the wire-format tree. Recursively processes
+   * nested values. See Section 4.5 of the formal spec.
    */
-  private serialize(
+  #encode(
```
danfuzz: For disambiguation, suggest calling the private methods `#encodeValue()` and `#decodeValue()`.

packages/data-model/src/json-wire/JsonEncodingContext.ts
```diff
@@ -257,43 +220,72 @@ export class JsonEncodingContext implements SerializationContext<string> {
   }
 
   //
-  // Tree-walking serialization (private)
+  // Tree-walking encode (private)
```
danfuzz: I don't think we need section markers beyond `Instance members` and `Static members`. And please note the code guidelines about order of items in a class. I think now is a good time to clean that aspect up in this class.

packages/data-model/src/json-wire/JsonEncodingContext.ts
danfuzz: I think it would be a lot more efficient to say `const [[key, value]] = Object.entries(data)` and follow on from there. A clarifying comment could indicate that we know at this point we're looking at a single-property object.

danfuzz:
Addressed the review feedback in `4cdc7e6b1` (codec/doc) and `85d360014` (JsonEncodingContext), all green (`deno task check` + 31/1622):

- **`CODEC` export** — now exported from the `wire-common` entrypoint.
- **`FabricHash` codec** — state is `{ tag, hash }` (separate parts); test updated, uses `EMPTY_RECONSTRUCTION_CONTEXT`, asserts `WIRE_TYPE_TAGS.Hash`.
- **`UndefinedCodec`** — throws on non-`null` state.
- **`FabricRegExp`** — documented the intentional no-syntax-enforcement-on-decode stance.
- **`BaseFabricCodec`** — dropped the copy-pasta'd top comment.
- **Unique symbols** — `JsonEncodingContext` now fails loudly on any value no codec/arm handles (was silently flattening to `{}`); `SymbolCodec.canEncode()` still declines, and a test pins the loud failure.
- **`#encode`/`#decode`** → `#encodeValue`/`#decodeValue`; `unwrapTag` uses `const [[key, value]] = Object.entries(data)`; collapsed section markers to `Instance members`/`Static members` and moved `seemsLikeEncoded` into the static section per the class-ordering convention.

**Not done here (by design):**
- **Spec-alignment** (`interface.ts:14`, `Symbol.for` key rename in the formal-spec docs) — left to your `danfuzz-land` process per your note.
- **`FabricMap`/`FabricSet`** stubs — confirmed intentional (incomplete), no change.

Captured for a follow-up PR (in my backlog): invert `[DECONSTRUCT]`/`[RECONSTRUCT]`↔`[CODEC]`; finish the per-class codec-test reorg; add `bigint{To,From}UnpaddedBase64url()` in `utils/bigint`.

packages/data-model/src/fabric-primitives/FabricHash.ts
```diff
@@ -169,15 +168,25 @@ export class FabricHash extends BaseFabricPrimitive implements ApiFabricHash {
         state: FabricValue,
         _context: ReconstructionContext,
       ): FabricValue {
-        if (typeof state !== "string") {
+        if (
+          state === null || typeof state !== "object" || Array.isArray(state)
```
danfuzz: There's `isPlainObject()` in `utils/types`, which can be used instead of this not-quite-correct longhand.

Perhaps this or other items in `utils/types` would be useful elsewhere in similar code in this PR.

packages/data-model/src/json-wire/JsonEncodingContext.ts
```diff
@@ -324,6 +297,19 @@ export class JsonEncodingContext implements SerializationContext<string> {
...
+    // Anything reaching here is none of: a registered-codec value, an
+    // `ExplicitTagValue`, a `FabricInstance`, a primitive, or an array -- so
+    // the only legitimate remaining shape is a *plain* object. A non-object
+    // (e.g. an uninterned/unique `symbol`, which no codec accepts) or a
+    // non-plain object (a class instance with no codec) is unencodable and
+    // must fail loudly rather than be mis-encoded as / silently flattened to a
+    // plain object.
+    if (!isPlainObject(value)) {
+      throw new Error(
+        `Cannot encode value of type \`${typeof value}\` (no applicable codec).`,
```
danfuzz: Suggest `toCompactDebugString(value, 50)` instead of just `typeof value` (which is sufficiently vague as to sometimes mislead).

danfuzz:
Addressed the latest two inline notes in `bf2e23e20` (`deno task check` + 31/1624 + `deno fmt`, all green):

- **`isPlainObject()`** — `FabricHash` and `FabricRegExp` decode-state guards now use `isPlainObject()` from `@commonfabric/utils/types` instead of the `=== null || typeof !== "object" || Array.isArray` longhand. Per your "perhaps elsewhere" nudge, also applied it to `JsonEncodingContext.unwrapTag`. (The encode-side guard already used it.)
- **`toCompactDebugString`** — the loud-fail guard now renders an unencodable *non-object* value via `toCompactDebugString(value, 50)` instead of the vague bare `typeof value` (so e.g. `Symbol(foo)` rather than `symbol`). The class-instance branch keeps its constructor-name form, since that's what made the `Uint8Array` case legible.

PR #3903, opened 2026-06-07
feat(runner): pass map/filter/flatMap op by {identity, symbol} (CT-1623)

seefeldb:
## What

Passes the `op` pattern of `.mapWithPattern` / `.filterWithPattern` / `.flatMapWithPattern` by its content-addressed `{ identity, symbol }` reference instead of an embedded, re-deserialized pattern graph. This is the **feature** that consumes the `__cfReg` content-addressable-artifact foundation merged in #3912.

## How
- At raw-node instantiation, `substituteOpPatternRefs` (runner) replaces a map/filter/flatMap `op` whose ref is known (`getArtifactEntryRef`) with a compact `{ $patternRef: { identity, symbol } }` sentinel, retaining the embedded graph as `$opFallback`. Scoped to the `op` key only.
- The list builtins resolve the sentinel synchronously via `artifactFromIdentitySync` (`resolveOpPattern`) — no graph deserialization, no `implementationRef` remap for the op. On a cache miss (op evicted from the bounded in-memory cache), it falls back to `$opFallback`, so cache residency is an optimization, not a correctness requirement.
- `traverse-utils` preserves the `unsafe_originalPattern` link when a pattern is copied during node wiring, so `getArtifactEntryRef`'s exact-then-original lookup recovers the op copy's ref.
- The function-form `OpaqueRef.map(fn)` / `.filter(fn)` / `.flatMap(fn)` (which wrapped the callback in an anonymous, identity-less inline pattern) now **throw**, pointing at the `*WithPattern` form. Authored code is always lowered by the transformer; in-process builder tests using the function form were rewritten to `*WithPattern`.

## Notes
- Rebased onto current `main` and reworked to consume #3912: the old export-the-hoist + re-parse-the-compiled-body discovery and the standalone `patternFromIdentitySync` are gone — refs/resolution come from #3912's `getArtifactEntryRef` / `artifactFromIdentitySync`.
- Runner suite: 547 passed, 0 failed.
- Follow-ups (separate PRs): remove `implementationRef`; remove `patternId`.

### DIFF
--- a/packages/runner/src/builder/traverse-utils.ts
+++ b/packages/runner/src/builder/traverse-utils.ts
@@ -1,5 +1,10 @@
 import { isRecord } from "@commonfabric/utils/types";
-import { isOpaqueRef, isPattern, type Opaque } from "./types.ts";
+import {
+  isOpaqueRef,
+  isPattern,
+  type Opaque,
+  unsafe_originalPattern,
+} from "./types.ts";
 import { isCell } from "../cell.ts";
 import { isCellResultForDereferencing } from "../query-result-proxy.ts";
 
@@ -34,11 +39,22 @@ export function traverseValue(
     if (Array.isArray(value)) {
       return (value as Array<any>).map((v) => traverseValue(v, fn, seen));
     } else {
-      return Object.fromEntries(
+      const copy = Object.fromEntries(
         Object.entries(value).map((
           [key, v],
         ) => [key, traverseValue(v, fn, seen)]),
       );
+      // `Object.entries` drops symbol keys, so a pattern copied here would lose
+      // its link back to the original (branded, content-addressed) factory —
+      // severing `findOriginalPattern`/`getArtifactEntryRef`, which is how a
+      // pattern passed as an `op` is later identified by `{ identity, symbol }`.
+      // Preserve that link (the deepest original) on the copy. Mirrors the same
+      // `unsafe_originalPattern` retention in `toJSONWithLegacyAliases`.
+      if (isPattern(value)) {
+        (copy as Record<symbol, unknown>)[unsafe_originalPattern] =
+          (value as Record<symbol, unknown>)[unsafe_originalPattern] ?? value;
+      }
+      return copy;
     }
   } else {
     return value;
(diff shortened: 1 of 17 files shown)


seefeldb:
Thanks for the review — addressed all three findings in 33a89e913:

**cubic P1 / Codex P2 — `resolveOpPattern` hard-fails on cache eviction.** Valid: the sentinel dropped the embedded op graph, so the fast path depended on the bounded (FIFO) `modulesByIdentity` cache, and a long-lived session could evict an op whose map/filter/flatMap node is still live. The sentinel now retains the embedded graph as `$opFallback`; `resolveOpPattern` resolves by identity (fast path) and falls back to the embedded graph on a miss, throwing only when neither is available. Per #3898 the serialized graph no longer drags in `program.files`, so retaining it doesn't reintroduce the cross-reload id churn that motivated this work — cache residency is now an optimization, not a correctness requirement.

**cubic P2 — `substituteOpPatternRefs` rewrites every pattern-like input.** Valid: only `op` is rehydrated by the builtins, so a pattern captured in `params` could have been corrupted into an unresolved `$patternRef`. Scoped the rewrite to the `op` key.

Red-green: added a unit test (fallback on sentinel miss) and an e2e test (map still outputs correctly when every sync identity lookup is forced to miss, simulating eviction). Full runner suite: 531 passed, 0 failed (with type-check).

seefeldb:
All three review comments are on the pre-rework commit `65090f227`; they're addressed in the current revision (the branch was rebased onto the merged `__cfReg` foundation and re-authored):

- **Codex P2 / cubic P1 — `resolveOpPattern` hard-fails on cache eviction:** the sentinel retains the embedded graph as `$opFallback`; `resolveOpPattern` resolves by identity, falls back to the embedded graph on a miss, and only throws when neither exists. Cache residency is an optimization, not a correctness requirement. (Plus a new e2e proving the fallback fires when every sync lookup is forced to miss.)
- **cubic P2 — `substituteOpPatternRefs` rewrites every input:** scoped to the `op` key only, so a pattern captured in `params` is never turned into an unresolved sentinel.

Also fixed a real reload regression this revision surfaced: a hoisted op's result cell reloaded by a hoist symbol (not an export), which the by-identity path missed → cold recompile (the shell piece test caught it). Both reload paths now resolve hoists from the artifact index; added a runner-level regression test. All checks green; mergeable.

PR #3909, opened 2026-06-08
fix(runner): register verified binding metadata for non-exported handlers (CT-1665)

bfollington:
> **Update (rev 3 — consolidated onto `__cfReg`, per framework-owner review):**
> The ESM loader no longer uses a parallel `setVerifiedBindingCandidateRegistrar`.
> `Engine.evaluateGraph` now records verified binding metadata directly from
> `graph.registrationSink` — the same `__cfReg` registrations pattern-manager
> indexes for `{identity, symbol}`. The builder-side registrar is retained **only**
> for the legacy/AMD path (`Engine.evaluate`), where `__cfReg` is a no-op (identity
> addressing is ESM-only) and the production-default loader still runs — so CT-1665
> stays fixed there with no regression; that channel retires with the legacy loader.
> `captureVerifiedBindingCandidates` is unchanged (both paths feed it; only the
> candidate source differs). Rebased onto current `main`. The "Fix" table below
> describes the original registrar approach and is kept for history.

---

## Why

**CT-1665.** Saving owner-protected profile fields (`avatar`, `name`, and profile elements) fails at commit under `enforce-explicit` with:

```
StorageTransactionAborted: writeAuthorizedBy requires a trusted verified binding identity at /avatar
```

The input clears as if it saved, but the transaction aborts — only the toolshed log reveals it. This blocks the profile UX (saving avatar/name and "Add profile card"). `setName` / `setAvatar` / `addElement` in `system/profile-home.tsx` are structurally identical owner-protected writers.

## Root cause

> This **refutes** the issue's original hypothesis (a follow-on to the `implementationRef`-stability work / ref instability). Instrumentation shows the `implementationRef` is stable and the handler function resolves fine — the verified **binding metadata is simply never registered** for these handlers.

Chain (verified by reproduction):

1. `handler()` / `lift()` return a **factory** = `Object.assign(callable, module)`.
2. The transformer's `__cfBindVerifiedBinding(setName, …)` attaches `__cfVerifiedBindingIdentity` to that **factory** (the implementation fn is already `Object.freeze`d by `hardenVerifiedFunction`, so the write to *it* is silently skipped).
3. The reactive node stores `node.module = module` — the **underlying** object, which never received the metadata.
4. `captureVerifiedValue` (which populates `verifiedBindingMetadata`) runs right after module **evaluation** and only walks the module's **export namespace**.

So the factory's metadata reaches the registry **only when the handler is `export`ed**. profile-home's handlers are non-exported module-scope consts → metadata never registered → `getVerifiedBindingMetadata(ref)` misses → `cfc/prepare.ts` rejects the write.

(The shell asymmetry where `name` *appeared* to work while `avatar` didn't is most plausibly the setup-projection / initialization exemption masking the initialized field's first write; all three handlers share the same defect. The element-write `addElement` — folded into this issue — is the same root cause and is also fixed.)

## Fix — runtime only (no transformer, freeze, or serialization changes)

| File | Change |
|---|---|
| `sandbox/function-hardening.ts` | New `setVerifiedBindingCandidateRegistrar` / `registerVerifiedBindingCandidate` hook (mirrors the existing function registrar). |
| `builder/module.ts` | `handlerInternal` and `createNodeFactory` surface their factory via `registerVerifiedBindingCandidate(factory)`. `createNodeFactory` guards on a JS-function implementation (see review note 2). |
| `harness/engine.ts` | Both load paths (source-based **and** source-free resume-by-identity) collect candidates during evaluation and call `executableRegistry.captureVerifiedBindingCandidates(...)` after `captureVerifiedValue`. |
| `harness/executable-registry.ts` | `captureVerifiedBindingCandidates(...)` records each factory's metadata keyed by its (shared) `implementationRef`. |

Recording happens during module **evaluation** (which re-runs on both fresh load and resume), so the fix covers both paths.

## Testing

New `packages/runner/test/cfc-nonexported-binding-identity.test.ts`:
- **RED without the fix** for internal handlers (esm on/off), the real `profile-home` (setName/setAvatar/addElement), and resume-by-identity.
- **GREEN for exported handlers** with *and* without the fix — a regression guard that pins the exact asymmetry.
- All green with the fix. **Full runner suite green (530 passed, 0 failed).**

## For runtime-owner review

A few design points worth a second opinion:

1. **Approach.** Surfacing builder factories during the load, vs. the alternative of attaching the binding identity to the implementation function *before* it's frozen (which would need a transformer change + pattern recompile). I went runtime-only to avoid touching the transformer and to keep the serialized/content-addressed graph untouched (the `669260eee` area).
2. **`createNodeFactory` guard.** Restricting registration to function modules also sidesteps a circular-import TDZ that surfaces when an implementation-less `type:"ref"` builtin (`builtins/sqlite/query-node.ts`) is built at module load *before* `function-hardening.ts` initializes. Ref builtins can't be trusted-binding writers, so the guard is semantically correct as well as necessary — but flagging it.
3. **Trust surface.** Recording reads `__cfVerifiedBindingIdentity` off genuine builder factories during a verified load — the same value the existing exported-handler capture already trusts. I don't believe this widens the trust surface, but it's the kind of thing worth your eyes.

Draft until reviewed.

---
## Summary by cubic
Registers verified binding metadata for non-exported handlers so `writeAuthorizedBy` works under `enforce-explicit` (CT-1665). Restores saving profile `name`, `avatar`, and profile elements on fresh loads and resume-by-identity.

- **Bug Fixes**
  - Added `setVerifiedBindingCandidateRegistrar`/`registerVerifiedBindingCandidate` in `sandbox/function-hardening.ts` to surface non-exported handler factories on the legacy/AMD path.
  - `handlerInternal` and `createNodeFactory` register factories; `createNodeFactory` guards to JS-function modules.
  - ESM loader records factories from `graph.registrationSink` (transformer `__cfReg`); legacy loader uses the registrar. Both paths (fresh and resume-by-identity) feed `executableRegistry.captureVerifiedBindingCandidates(...)`.
  - Registry stores metadata by `implementationRef` so CFC resolves non-exported handlers.
  - Added `cfc-nonexported-binding-identity.test.ts` covering internal vs. exported handlers, real `system/profile-home.tsx`, and resume.

- **Refactors**
  - Comment polish: clarified the TDZ guard in `createNodeFactory` and documented test scope (no behavior change).

<sup>Written for commit 8079093c197aebbaa4b259f6031832e8ebaf7558. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3909?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/builder/module.ts
+++ b/packages/runner/src/builder/module.ts
@@ -157,8 +157,9 @@ export function createNodeFactory<T = any, R = any>(
   // look-alike never acquires the brand. (Patterns brand separately in
   // builder/pattern.ts.)
   brandTrustedBuilderArtifact(factory);
-  // CT-1665: surface the factory so the engine can register verified binding
-  // metadata for non-exported lift/derive/computed bindings (see handlerInternal).
+  // CT-1665: surface the factory so the legacy/AMD eval path can register verified
+  // binding metadata for non-exported lift/derive/computed bindings (the ESM loader
+  // instead reuses `__cfReg`; see handlerInternal and Engine.evaluate).
   // Only JS-function modules can be trusted-binding writers (and carry
   // `__cfVerifiedBindingIdentity`); `type: "ref"` builtins never do — so the guard
   // is semantically correct. It is also load-safe: an implementation-less builtin
@@ -491,11 +492,13 @@ function handlerInternal<E, T>(
     return eventStream;
   }, module);
 
-  // CT-1665: surface the factory so the engine can register its verified binding
-  // metadata after evaluation. The transformer-emitted `__cfBindVerifiedBinding`
-  // annotates THIS object (the builder's return value) once the module body
-  // finishes; a non-exported binding is otherwise unreachable from the capture
-  // walk and CFC would reject its owner-protected writes.
+  // CT-1665: surface the factory so the legacy/AMD eval path can register its
+  // verified binding metadata after evaluation (the ESM loader instead reuses the
+  // transformer's `__cfReg` registrations — see Engine.evaluate vs evaluateGraph).
+  // The transformer-emitted `__cfBindVerifiedBinding` annotates THIS object (the
+  // builder's return value) once the module body finishes; a non-exported binding
+  // is otherwise unreachable from the export-namespace capture walk and CFC would
+  // reject its owner-protected writes.
   registerVerifiedBindingCandidate(factory);
 
   return factory;
--- a/packages/runner/src/builder/module.ts
+++ b/packages/runner/src/builder/module.ts
@@ -160,11 +160,13 @@ export function createNodeFactory<T = any, R = any>(
   // CT-1665: surface the factory so the engine can register verified binding
   // metadata for non-exported lift/derive/computed bindings (see handlerInternal).
   // Only JS-function modules can be trusted-binding writers (and carry
-  // `__cfVerifiedBindingIdentity`); `type: "ref"` builtins never do. Guarding on
-  // the implementation also keeps an implementation-less builtin node factory
-  // built at module load (e.g. builtins/sqlite/query-node.ts) from touching the
-  // registrar before function-hardening.ts has initialized under a circular
-  // import.
+  // `__cfVerifiedBindingIdentity`); `type: "ref"` builtins never do — so the guard
+  // is semantically correct. It is also load-safe: an implementation-less builtin
+  // node factory is created at module-load time (e.g. builtins/sqlite/query-node.ts)
+  // and is reached via a module.ts import cycle BEFORE module.ts has evaluated its
+  // own `import` of function-hardening.ts, so calling the registrar there would hit
+  // its still-uninitialized module-scope `let` (a TDZ throw). Function modules are
+  // only ever built later, at pattern-build time, well clear of that window.
   if (typeof module.implementation === "function") {
     registerVerifiedBindingCandidate(factory);
   }
--- a/packages/runner/src/builder/module.ts
+++ b/packages/runner/src/builder/module.ts
@@ -29,6 +29,7 @@ import { getLogger } from "@commonfabric/utils/logger";
 import { createRef } from "../create-ref.ts";
 import {
   hardenVerifiedFunction,
+  registerVerifiedBindingCandidate,
   registerVerifiedFunctionImplementation,
 } from "../sandbox/function-hardening.ts";
 
@@ -156,6 +157,17 @@ export function createNodeFactory<T = any, R = any>(
   // look-alike never acquires the brand. (Patterns brand separately in
   // builder/pattern.ts.)
   brandTrustedBuilderArtifact(factory);
+  // CT-1665: surface the factory so the engine can register verified binding
+  // metadata for non-exported lift/derive/computed bindings (see handlerInternal).
+  // Only JS-function modules can be trusted-binding writers (and carry
+  // `__cfVerifiedBindingIdentity`); `type: "ref"` builtins never do. Guarding on
+  // the implementation also keeps an implementation-less builtin node factory
+  // built at module load (e.g. builtins/sqlite/query-node.ts) from touching the
+  // registrar before function-hardening.ts has initialized under a circular
+  // import.
+  if (typeof module.implementation === "function") {
+    registerVerifiedBindingCandidate(factory);
+  }
   return factory;
 }
 
@@ -476,6 +488,13 @@ function handlerInternal<E, T>(
 
     return eventStream;
   }, module);
+
+  // CT-1665: surface the factory so the engine can register its verified binding
+  // metadata after evaluation. The transformer-emitted `__cfBindVerifiedBinding`
+  // annotates THIS object (the builder's return value) once the module body
+  // finishes; a non-exported binding is otherwise unreachable from the capture
+  // walk and CFC would reject its owner-protected writes.
+  registerVerifiedBindingCandidate(factory);
 
   return factory;
 }
--- a/packages/runner/src/harness/engine.ts
+++ b/packages/runner/src/harness/engine.ts
@@ -69,7 +69,10 @@ import {
   createModuleCompartmentGlobals,
   createSafeConsoleGlobal,
 } from "../sandbox/compartment-globals.ts";
-import { setVerifiedFunctionRegistrar } from "../sandbox/function-hardening.ts";
+import {
+  setVerifiedBindingCandidateRegistrar,
+  setVerifiedFunctionRegistrar,
+} from "../sandbox/function-hardening.ts";
 import type { UnsafeHostTrustOptions } from "../unsafe-host-trust.ts";
 import { ExecutableRegistry } from "./executable-registry.ts";
 import { CompiledBundleValidator } from "../sandbox/compiled-bundle-validation.ts";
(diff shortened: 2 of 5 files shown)


bfollington:
## Self-review (`cf-review`) — verdict: ✅ Approve, no blocking findings

Ran the repo's `cf-review` over this branch (automated pass, report-only). What I directly verified: all 5 changed files pass `deno fmt --check` / `deno lint` / `deno check`; the new regression test is 6/6; the **full runner suite is green (530 passed, 0 failed)**. (The stop-hook `deno fmt` failure the pass hit is pre-existing formatting drift in `packages/ui/src/v2/components/**` — unrelated to this diff.)

### Findings + actions

**1. (non-blocking) TDZ-guard comment wording — `builder/module.ts`.** The original comment over-attributed the circular import to `function-hardening.ts`, which is actually an import-less leaf. ✅ **Fixed in `a3dc6e39e`:** the cycle is `module.ts ↔ builtins/sqlite/query-node.ts` — query-node's module-load `createNodeFactory` is reached *before* `module.ts` evaluates its own `import` of `function-hardening.ts`, so the registrar's module-scope `let` is still uninitialized. The guard (`typeof module.implementation === "function"`) is both semantically correct (`type:"ref"` builtins are never trusted-binding writers) and load-safe.

**2. (non-blocking) Test is white-box + `observe`-mode.** It asserts on the binding-metadata registry via `getVerifiedBindingMetadata` (the exact lookup the CFC verifier performs at commit) under `observe`, so it proves the writer identity *resolves*, not that the write is *accepted* under `enforce`. ✅ **Documented in `a3dc6e39e`** — a full enforce-mode, end-to-end assertion additionally needs trust-snapshot + owner-principal + trusted-event provenance setup (as `profile-owner-cfc.test.ts` mocks). Happy to add that heavier test if you'd prefer it before merge.

**3. (informational) Design note** — the candidate-channel approach vs. re-reading already-registered implementations. No change; raised in the PR description's "For runtime-owner review" section for discussion.

The open design questions for review are in the PR description under **For runtime-owner review** (approach choice, the `createNodeFactory` guard, and trust surface).

🤖 self-review via `cf-review`

bfollington:
## Rebased onto current main (post-#3912) — this PR still stands; #3912 does not fix CT-1665

Since this PR was opened, main absorbed #3912 (`__cfReg`) and the rest of the CT-1623 series. I rebased and checked empirically whether #3912 changes the picture. Short version: **it doesn't — the bug reproduces identically, and this fix still applies cleanly and passes.** There's one architectural note for @seefeldb at the bottom.

### Rebase
Clean. The only conflict is in `createNodeFactory` (`packages/runner/src/builder/module.ts`), where #3912's `brandTrustedBuilderArtifact(factory)` and this PR's `registerVerifiedBindingCandidate(factory)` both land right after `factory.asScope = …`. They're independent — one brands the factory for `{identity, symbol}` addressing, the other surfaces it for binding-metadata capture — and populate different maps, so both are kept. `engine.ts` auto-merged.

### Does #3912 fix it? No — verified empirically
Ran this PR's test on **pristine main (with #3912, without this PR)**:

| Case | Result |
|---|---|
| `exported` handlers (esm off/on) | ✅ pass |
| `internal` / non-exported handlers (esm off/on) | ❌ `verifiedBindingMetadata` empty (`Value: []`) |
| real `system/profile-home.tsx` `setName`/`setAvatar`/`addElement` | ❌ |
| resume-by-identity | ❌ |

`0 passed (2 steps) | 1 failed (4 steps)` — the exact RED/GREEN split from the original PR. Nothing changed.

**Why they don't overlap:** #3912's `__cfReg` → `indexArtifact` registers non-exported builder artifacts into `addressableByIdentity` (keyed by `{identity, symbol}`, for op-by-identity resolution). CFC's `writeAuthorizedBy` resolves through a *different* index — `verifiedBindingMetadata` (keyed by `implementationRef`, in `executable-registry.ts`), populated only by `captureVerifiedValue`'s export-namespace walk. The two channels never cross, so a non-exported handler's metadata is still absent and the commit is still rejected.

Root cause reconfirmed against the transformer (`module-scope-function-hardening.ts`, the `inlineBindingAnnotation = isTrustedCallable && exported` branch): both exported and non-exported trusted handlers emit `__cfBindVerifiedBinding` — exported *inline* (flows into the module namespace), non-exported as a *trailing statement* that mutates the already-bound const. The `__cfVerifiedBindingIdentity` lands on the factory either way; the non-exported factory just never enters the namespace, which is the only thing the post-eval capture walks. So this remains a runtime-capture gap, fixed runtime-side.

### This PR rebased: green, and coexists with #3912
- This PR's test on the rebased branch: 6/6 steps green.
- Run together with #3912's own `cfreg-builder-identity` / `cfreg-security` / `compiled-module-verifier` plus `profile-owner-cfc`: **97 steps, all green.** The conflict resolution breaks neither side.

### @seefeldb — the one thing #3912 *does* change
It gives us a channel this PR could consolidate onto. The live non-exported factories (carrying both `__cfVerifiedBindingIdentity` and `implementationRef`) are already collected by `__cfReg` into `graph.registrationSink` / `result.registrationsByIdentity` at the engine layer, in the **same verified-load pass this PR hooks** (`engine.ts`, where `registrationsByIdentity: graph.registrationSink` is assembled). So as a follow-up, this PR could iterate those registrations and call `captureVerifiedBindingCandidates` on them, instead of adding its own `setVerifiedBindingCandidateRegistrar` + the two builder-side `registerVerifiedBindingCandidate` calls.

Trade-offs of routing through `__cfReg` instead:
- `__cfReg` is **top-level-only** — sufficient for CT-1665 (profile-home's handlers are top-level consts) but narrower than this PR's builder-level hook, which catches any trusted factory regardless of nesting.
- `__cfReg` values are **verifier-gated**; this PR's channel is independent of that approval.
- Net: it's a *lateral move for correctness* (both fix the bug identically) but a plausible *code-consolidation win* (one channel for "surface non-exported builder factories to the runtime").

**Proposal:** land this as-is to unblock the profile save UX, and treat "route binding-metadata capture through `__cfReg`'s registration sink" as an optional follow-up if you'd prefer a single channel. Happy to do that rework instead if you'd rather not carry the parallel hook — your call.

bfollington:
@seefeldb done — consolidated onto `__cfReg` as you suggested. Pushed (now `8079093c1`, rebased onto current `main`).

**On your question:** the profile handlers (`setName` / `setAvatar` / `addElement` in `system/profile-home.tsx`) are *already* top-level non-exported consts — that's exactly the shape that was broken. So there's no "make it top-level" workaround; they already are. And you're right that there's no "trusted factory regardless of nesting" case to worry about — that generality in the first cut was unnecessary, and it's gone now on the ESM path.

### What changed
- `Engine.evaluateGraph` (the ESM core, shared by source-compile **and** resume-by-identity) now records verified binding metadata straight from `graph.registrationSink` — the same `__cfReg` registrations pattern-manager already indexes for `{identity, symbol}`. The parallel `setVerifiedBindingCandidateRegistrar` install is gone from the ESM path. (`__cfBindVerifiedBinding` runs in the module body, ahead of the trailing `__cfReg({…})`, so each registered factory already carries its `__cfVerifiedBindingIdentity`.)
- `captureVerifiedBindingCandidates` (executable-registry) is unchanged — both paths still feed it; only the candidate **source** differs (sink vs registrar).

### One thing to confirm — the only reason the registrar isn't fully deleted
The legacy/AMD eval path (`Engine.evaluate`) has **no `__cfReg`** — it's a no-op there, since identity addressing is ESM-only. And the production default is still the non-ESM loader (`CF_ESM_MODULE_LOADER` unset → off), with `enforce-explicit` on by default. So I kept the builder-side registrar **scoped to the legacy path only**, so CT-1665 stays fixed under the default loader with no regression. It's commented as retiring together with the legacy loader.

If you'd rather make CT-1665 **ESM-only** now — drop the legacy registrar + builder hook + the `esm=false` test cases, betting the loader flip is imminent — say the word; it's a ~10-line delete. I went with "keep the default loader working" as the safe default, since the shipped shell doesn't force the ESM flag on.

### Tests (green on both loaders, current main)
`cfc-nonexported-binding-identity` (esm on/off, real `profile-home`, resume-by-identity) + your `cfreg-builder-identity` / `cfreg-security` / `compiled-module-verifier` + `profile-owner-cfc`.

PR #3912, opened 2026-06-08
feat(runner): content-addressable builder artifacts via __cfReg (CT-1623)

seefeldb:
## What

Makes every top-level builder artifact in a pattern module — `pattern`, `lift`, `handler` (and the synthetic hoists the transformer creates for `.map`/`.filter`/`.flatMap` ops, inline computeds, actions, etc.) — addressable by a content-addressed `{ identity, symbol }` reference, via a single transformer-emitted `__cfReg({ … })` registration call.

This is the **foundation PR** for CT-1623's "identify a pattern/lift/handler by `{ identity, symbol }`" work. It supersedes the export-the-hoist-then-re-parse-the-compiled-body approach (the `extractSyntheticPatternExports` smell): nothing parses source to discover hoists, and we stop overloading SES module *exports* as a proxy for "trusted, content-addressable top-level symbol." A follow-up PR consumes these refs to pass map/filter/flatMap `op` by identity instead of by embedded graph.

## How

**Transformer** appends one trailing call listing every hoisted artifact plus every authored *non-exported* top-level builder const:
```js
const helper = lift(...);            // authored, non-exported
const __cfPattern_1 = pattern(...);  // synthetic hoist (a .map op)
// …module body…
__cfReg({ helper, __cfPattern_1 });
```
- One trailing call (not per-hoist) keeps the verifier's obligation to "exactly one top-level `__cfReg`" and lets a run-once trap reject any injected duplicate.
- Authored consts are detected via `detectCallKind` (a direct builder *call* only), so an import/alias is never mis-attributed to this module's identity.

**`__cfReg` delivery** — it's the module factory's 4th parameter under the ESM loader (an identity-bound per-module registrar), and a no-op global on the legacy/AMD path (identity addressing is ESM-only). The registrar is **run-once** (2nd call throws), **window-closed** after the body returns (a captured closure can't register late), and **transactional** — entries are staged and committed only when the factory returns normally, so a throw leaves nothing behind.

**Runtime** — `registerHoistedValues` (PatternManager) consumes the registrations: only a genuine trusted builder artifact gets a ref (`isTrustedBuilderArtifact` — patterns brand as before; lift/handler/node-factories now brand in `createNodeFactory`). Forged `__cf_data` plain objects are dropped; cross-module forgery is already impossible since identity is a content hash. Adds the forward `valueToEntryRef` and reverse `hoistsByIdentity` index, plus public `patternFromIdentitySync` (a miss returns `undefined`; callers fall back).

**Verifier** — accepts exactly one top-level `__cfReg({ …shorthand top-level bindings })`; any other `__cfReg` use falls through to the existing unknown-identifier rejection. Defense-in-depth over the runtime trap.

## Two registration channels (by design)

- **Non-exported top-level + hoists → `__cfReg`.** These never reach the module namespace, so `__cfReg` is the only way to address them.
- **Exported artifacts → the module namespace**, addressable by their export name (the existing `registerEvaluatedModules` ref). This is *not* a source parse — it reads the live module interface — so it isn't the smell this PR removes.

Exported artifacts are intentionally **not** routed through `__cfReg`: TypeScript's CommonJS emit rewrites `export const foo = …` to `exports.foo = …` with no local `foo` binding, so a `{ foo }` shorthand would read `undefined`. Unifying them would mean rewriting authored `export const` to the split `const foo = …; export { foo }` form — invasive churn for no functional gain, since the namespace already gives exports a correct, canonical, parse-free ref. (Noted as an optional follow-up.)

## Testing

- `packages/runner` — **531 passed, 0 failed** (with type-check). New `cfreg-builder-identity.test.ts`: registrar run-once/closed-window/transactional unit tests; e2e proving pattern + lift hoists and an authored non-exported const are all addressable + round-trip; `compiled-module-verifier.test.ts`: accept one `__cfReg`, reject a second, reject an undeclared-binding `__cfReg`.
- `packages/ts-transformers` — **290 passed** (goldens regenerated: a trailing `__cfReg({…})`, no exports).
- `packages/generated-patterns` integration — **145/145** patterns pass end-to-end. Legacy/AMD path unaffected (no-op `__cfReg`).

## Notes

- TDZ guard: `createNodeFactory` brands at module-init inside the builder import cycle, where a top-level `const` brand set would be in its TDZ — held on a hoisted-function lazy `WeakSet` instead.
- Stacked: the map/filter/flatMap op-by-identity feature (#3903) will be reworked to consume this and drop its own `extractSyntheticPatternExports` path.

---
## Summary by cubic
Adds content‑addressable references for top‑level builder artifacts via a single transformer‑emitted `__cfReg({ … })`, enabling `{ identity, symbol }` lookup without export parsing. Verifier‑gated with a fail‑closed registrar and a single reverse index with artifact‑generic APIs (CT‑1623).

- **New Features**
  - Transformer appends one trailing `__cfReg({ … })` listing all hoists plus authored non‑exported top‑level builder consts; exported artifacts (all forms) are excluded; keys are symbols, values are live bindings.
  - ESM loader passes a per‑module registrar as the factory’s 4th param; registrar is run‑once, closed‑window, and transactional; non‑approved modules get a rejecting registrar; AMD/legacy path keeps a no‑op `__cfReg`.
  - Verifier reports `{ hasHoistRegistration }`; engine grants the real registrar only to approved modules (trusted bodies grant it to all) and returns `registrationsByIdentity` in `EvaluateResult`.
  - `PatternManager` builds a forward `value → { identity, symbol }` map and a single reverse `identity → symbol → value` index populated from both `__cfReg` and exports; resolves via `artifactFromIdentitySync`.

- **Refactors**
  - Brand non‑pattern builders in `createNodeFactory`; `isTrustedBuilderArtifact` gates indexing and follows `unsafe_originalPattern` safely (fail‑closed on exotic values).
  - Reverse mappings overwrite on re‑eval to return the freshest artifact; `modulesByIdentity` now only caches whole namespaces.
  - API rename: `getPatternEntryRef` → `getArtifactEntryRef`; transformer export detection unwraps wrapped default exports (`(foo)`, `foo as …`, `foo satisfies …`, `<T>foo`) to keep exports out of `__cfReg`.

<sup>Written for commit 1ec8c651db791d90125bf6598f5acf5758355b10. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3912?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/builder/module.ts
+++ b/packages/runner/src/builder/module.ts
@@ -22,6 +22,7 @@ import {
   connectInputAndOutputs,
 } from "./node-utils.ts";
 import { moduleToJSON } from "./json-utils.ts";
+import { brandTrustedBuilderArtifact } from "./pattern-metadata.ts";
 import { getTopFrame } from "./pattern.ts";
 import { generateHandlerSchema } from "../schema.ts";
 import { getLogger } from "@commonfabric/utils/logger";
@@ -148,6 +149,13 @@ export function createNodeFactory<T = any, R = any>(
   }, module) as ModuleFactory<T, R>;
   factory.asScope = (scope: CellScope) =>
     createNodeFactory({ ...module, defaultScope: scope });
+  // Provenance brand: every node factory (lift / handler / byRef / the list-op
+  // factories) is a trusted builder artifact, so a hoisted one registered via
+  // `__cfReg` may receive a content-addressed `{ identity, symbol }` reference.
+  // Only the trusted builders call `createNodeFactory`, so a `__cf_data`-forged
+  // look-alike never acquires the brand. (Patterns brand separately in
+  // builder/pattern.ts.)
+  brandTrustedBuilderArtifact(factory);
   return factory;
 }
 
--- a/packages/runner/src/builder/pattern-metadata.ts
+++ b/packages/runner/src/builder/pattern-metadata.ts
@@ -34,6 +34,25 @@ const verifiedLoadIdByValue = new WeakMap<object, string>();
 // derivation copies and independently re-resolves node implementations).
 const trustedPatterns = new WeakSet<object>();
 
+// Provenance brand for the OTHER trusted builder artifacts — `lift`, `handler`,
+// and the node factories they produce (see builder/module.ts `createNodeFactory`).
+// `pattern()` brands `trustedPatterns` instead. Kept as a separate set so the
+// pattern-only trust gate (`isTrustedPattern`, used by CFC) is unchanged, while
+// `isTrustedBuilderArtifact` accepts any trusted builder output. Like the pattern
+// brand, this is the gate that stops `__cf_data`-forged data from acquiring a
+// content-addressed `{ identity, symbol }` reference via `__cfReg`.
+//
+// Held lazily on this hoisted accessor rather than a top-level `const`/`let`:
+// unlike `pattern()`, `createNodeFactory` is invoked at MODULE-INIT time by some
+// builtins (e.g. builtins/sqlite/query-node), which runs inside the builder
+// import cycle — a top-level binding would still be in its temporal dead zone at
+// that point. A function declaration IS fully hoisted, so caching the `WeakSet`
+// on it sidesteps the init-order dependency entirely.
+function trustedBuilderArtifacts(): WeakSet<object> {
+  const self = trustedBuilderArtifacts as { set?: WeakSet<object> };
+  return (self.set ??= new WeakSet<object>());
+}
+
 function asKey(value: unknown): object | undefined {
   if (value === null) return undefined;
   return (typeof value === "object" || typeof value === "function")
@@ -95,6 +114,40 @@ export function isTrustedPattern(value: unknown): value is Pattern {
     current && (typeof current === "object" || typeof current === "function")
   ) {
     if (trustedPatterns.has(current as object)) return true;
+    if (seen.has(current)) break;
+    seen.add(current);
+    current = (current as Record<symbol, unknown>)[unsafe_originalPattern];
+  }
+  return false;
+}
+
+/** Stamp a value as produced by a trusted non-pattern builder (lift/handler/…). */
+export function brandTrustedBuilderArtifact<T>(value: T): T {
+  const key = asKey(value);
+  if (key) trustedBuilderArtifacts().add(key);
+  return value;
+}
+
+/**
+ * True for any value with trusted-builder provenance — a trusted pattern OR a
+ * branded lift/handler/node-factory — walking the `unsafe_originalPattern` chain
+ * so derivation / serialized copies inherit trust. This is the gate that decides
+ * whether a `__cfReg`-registered value may receive a content-addressed
+ * `{ identity, symbol }` reference; forged plain data carries no brand and is
+ * rejected.
+ */
+export function isTrustedBuilderArtifact(value: unknown): boolean {
+  let current: unknown = value;
+  const seen = new Set<unknown>();
+  while (
+    current && (typeof current === "object" || typeof current === "function")
+  ) {
+    if (
+      trustedBuilderArtifacts().has(current as object) ||
+      trustedPatterns.has(current as object)
+    ) {
+      return true;
+    }
     if (seen.has(current)) break;
     seen.add(current);
     current = (current as Record<symbol, unknown>)[unsafe_originalPattern];
--- a/packages/runner/src/builder/pattern-metadata.ts
+++ b/packages/runner/src/builder/pattern-metadata.ts
@@ -150,7 +150,13 @@ export function isTrustedBuilderArtifact(value: unknown): boolean {
     }
     if (seen.has(current)) break;
     seen.add(current);
-    current = (current as Record<symbol, unknown>)[unsafe_originalPattern];
+    // Fail closed: reading the (module-private) symbol off an exotic value — e.g.
+    // a Proxy with a throwing get trap — must not abort registration/lookup.
+    try {
+      current = (current as Record<symbol, unknown>)[unsafe_originalPattern];
+    } catch {
+      return false;
+    }
   }
   return false;
 }
--- a/packages/runner/src/harness/engine.ts
+++ b/packages/runner/src/harness/engine.ts
@@ -793,7 +793,16 @@ export class Engine extends EventTarget implements Harness {
       this.executableRegistry.captureVerifiedValue(loadId, exportMap);
       this.runtimeInternals?.exportsCallback(exportsByValue);
 
-      return { main, exportMap, loadId, exportsByIdentity };
+      // `graph.registrationSink` was populated by each module's `__cfReg` during
+      // the `importNow` loop above (committed only for modules that evaluated
+      // cleanly).
+      return {
+        main,
+        exportMap,
+        loadId,
+        exportsByIdentity,
+        registrationsByIdentity: graph.registrationSink,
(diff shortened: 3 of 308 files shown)


seefeldb:
Thanks for the review — addressed in 294c37137 (+ the prior 008a027cc / a38965de0):

**cubic P1 (`logCharmsList` in `__cfReg` despite `export default`) + P2 (`export { ... }` re-exports):** Same root cause — the export filter only checked the inline `export const` modifier, so artifacts exported via a separate statement slipped in. Now I collect every local name exported by *any* form (`export const`, `export { x }` / `export { x as y }`, `export default x`) and exclude them. Exported artifacts are addressed solely via the module namespace; `__cfReg` carries only hoists + non-exported top-level builder consts. Goldens regenerated (the `__cfReg` entry is gone from those fixtures).

**cubic P2 (`isTrustedBuilderArtifact` can throw on a Proxy/getter):** Wrapped the `unsafe_originalPattern` read in try/catch — fails closed to `false` so an exotic value can't abort registration/lookup. (Proxy is disabled in-sandbox, but fail-closed is right.)

**Codex P2 (index exported lifts/handlers too):** Already handled — the reverse/forward index was unified in a38965de0, and `indexArtifact` gates on `isTrustedBuilderArtifact` (not pattern-only), so exported lifts/handlers now get a forward `{identity, symbol}` ref as well. That commit also collapsed `patternFromIdentitySync` to read a single index.

ts-transformers 290 passed; runner 531 passed; generated-patterns 145 passed (locally).

seefeldb:
Addressed the latest two in `8e…` → next commit:

- **`indexArtifact` staleness (P2):** now overwrites the reverse mapping, so re-evaluating the same content identity resolves `artifactFromIdentitySync` to the fresh artifact instance, not a stale one. (Forward `valueToEntryRef` keeps the first/canonical symbol per value — separate concern.)
- **`collectExportedLocalNames` wrapped default export (P2):** unwraps parens / `as` / `satisfies`, so `export default (foo)` / `export default foo satisfies T` are recognized as exporting `foo` and kept out of `__cfReg`.

The `export { ... }` re-export flag was already fixed in 294c371 (cubic re-flagged the old line). ts-transformers 290 / runner 531 green.

seefeldb:
Latest round (`9a054d8`):

- **`export { ... }` re-export (404)** and **parens/`as`/`satisfies` wrapped default (438)** are already handled (collectExportedLocalNames covers all export forms; the unwrap was added in cd75afe7). cubic re-flagged them on the shifted line numbers. Now locked in by a new golden fixture `cfreg-export-forms`, which proves `__cfReg({ internalHelper, __cfPattern_1 })` — an inline `export const`, a separate `export { x }`, and the default export are all correctly excluded.
- **`<T>expr` TypeAssertionExpression (421):** valid gap — `unwrapDefaultExportExpression` now also unwraps angle-bracket assertions (theoretical in `.tsx`, but complete).

All green; mergeable.

seefeldb:
**Performance Check — flaky, not a regression from this PR.**

The `scrabble.test.tsx` perf gate passes and fails on byte-identical runtime code across this PR's commits:

| commit | perf |
|---|---|
| f0a10e8 | ✅ |
| a38965d | ✅ |
| 294c371 | ❌ |
| 8e9f8d0 | ✅ |
| cd75afe | ✅ |
| 9a054d8 | ❌ (re-run 3×, all 12.1s) |

`9a054d8`'s entire diff vs the passing `cd75afe` is a ts-transformers test fixture + a one-line `TypeAssertionExpression` addition to default-export unwrapping — **nothing that touches scrabble's compile or runtime**. A genuine +3.5s regression would fail every run, not ~50%. Measurement lands at 12.1s vs an 11.8s threshold derived from a stale 8.6s median (recent `main` PRs already measure 10–12s on this test), so it straddles the boundary by runner variance.

Compile cost is unchanged (`cf check scrabble` ≈ 1.2s on this branch). Reran the gate 3× per process; surfacing rather than churning further.

seefeldb:
## Security hardening + red-team (`759b105`)

Treating the compiled module body as untrusted with the verifier as the boundary:

**Fail-closed registrar capability.** The verifier now reports whether a module has a *valid approved* `__cfReg` call (`hasHoistRegistration`); the engine records approved specifiers on the graph. The module wrapper grants the **real** registrar only to approved modules — every other module gets `createRejectingRegistrar`, which **throws** on any `__cfReg` call. So a registrar invocation the static verifier missed (e.g. smuggled inside an accepted expression) fails closed at runtime (terminating the import) rather than registering attacker-chosen values.

**Red-teamed** (a subagent enumerated attacks; I turned them into tests):
- `esm-verifier-adversarial.test.ts` — 12 `__cfReg` cases: legit single shorthand call accepted; **undeclared/non-module-level symbol**, second call, member-call/aliased/reassigned callee, string/computed/`key:value`/spread entries, extra argument, variable-initializer capture, and unicode-escaped callee all **rejected**.
- `cfreg-security.test.ts` (new) — the approval signal (true/false), the rejecting registrar throws, and the load-bearing per-value trust gate: `isTrustedBuilderArtifact` rejects plain objects/functions/frozen data, a brandless pattern-shaped `__cf_data` forgery, and a **STRING-keyed `unsafe_originalPattern` spoof**, while accepting genuinely branded artifacts + their symbol-chained copies.

**Documented follow-up (out of scope):** the verifier doesn't recurse into an accepted `__cf_data(...)` argument, so arbitrary code (incl. a `__cfReg` call) can ride a comma operator there. The fail-closed registrar + run-once trap contain the `__cfReg`-specific impact; constraining `__cf_data` arguments is a broader, pre-existing verifier concern.

Full runner suite: **535 passed, 0 failed**.

seefeldb:
**`Package Integration Tests (shell)` failure is pre-existing on `main`, not from this PR.**

The failing assertion — *"Expected 0 in-client compilations while loading cf-created piece … compile count changed from 2 to 3"* (`packages/shell/integration/piece.test.ts`) — reproduces identically on `origin/main` HEAD (`69c57fc5`, #3908) with none of this PR's changes:

| main commit | shell test |
|---|---|
| `0813b47d` (#3905) | ✅ |
| `f1af61c4` (#3907) | ✅ |
| `69c57fc5` (#3908) | ❌ same compile-count error |

This PR is rebased onto `69c57fc5`, so it inherits that red. The CT-1623 reload-invariant break was introduced by #3908 (the Self-pattern UI change shifting the default-app load), independent of `__cfReg`. Every other check passes and the PR is mergeable; this gate will clear once the #3908 regression is fixed on `main` (and I re-merge).

mathpirate:
I undertook some follow-ups in https://github.com/commontoolsinc/labs/pull/3927 [test and comment only].

Two items from the same read I did not turn into code, which are pending considerations for you when you have time:

The bounded-cache sizing (MAX_EVALUATED_MODULE_CACHE_SIZE shared by modulesByIdentity + addressableByIdentity) — once the flag is on, worth watching whether op-by-identity hit rates hold under real multi-pattern sessions (degrades correctly to $opFallback, just defeats the dedup). No action now.

Whether exported export const foo = lift(...) should eventually route through __cfReg too (your description notes the CommonJS-emit reason it doesn't, as an optional follow-up).

PR #3917, opened 2026-06-08
fix(memory): convert native values to fabric at the wire boundary

danfuzz:
## Summary

A prep PR (fixes a latent `main` bug; also unblocks the codec migration #3900).

`encodeMemoryBoundary` cast its input straight to `FabricValue` and serialized it. Native values that aren't already `FabricValue`s — a `Date`, a `Uint8Array`, etc. — fell through the encoder's structural object/array fallback and were **silently mangled**: an unrecognized instance flattens to an empty `{}` (total data loss). The most common real case: a cell holding a `Date` (`Cell.of(new Date(...))`).

The fix converts native values via `fabricFromNativeValue()` before serializing, so a `Date` becomes a `FabricEpochNsec`, a `Uint8Array` a `FabricBytes`, etc. — proper tagged wire forms that round-trip faithfully.

### Why now / relation to #3900

The codec migration (#3900) replaces the lenient encoder with a strict one that **correctly throws** on an unconverted native instance (`Cannot encode a Date instance: no applicable codec`) instead of mangling it. Without this fix, #3900 regresses any cell holding a `Date`: I bisected a `runner` test failure (`cell-static-methods.test.ts`, the `Cell.of(new Date())` case) to exactly this — `main` passes (mangles the Date to `{}`), #3900 fails (the deferred write's encode throws inside the async storage flush and surfaces at teardown as a dangling `memory client closed` / `cfc-relevant-transaction-not-prepared` rejection). Applying this fix on #3900 makes that test pass.

It's a strict improvement on `main` too — Dates were being silently lost there.

The signature path is unaffected: post-#3902 the session-open signature is already a plain number array, which `fabricFromNativeValue` leaves untouched (it does **not** become a `FabricBytes`), so `toByteArray` still decodes it.

## Test plan

- New `memory v2 boundary encode` tests: a native `Date` → round-trippable `FabricEpochNsec` (proper `EpochNsec` tag, not `{}`); a native `Uint8Array` → round-trippable `FabricBytes`; plain JSON unchanged.
- Full `packages/memory` suite green (166 passed); `deno check` clean.
- Regression check: toolshed memory-route session tests (signature round-trip, incl. `FabricBytes`) all pass — the number-array signature flows through unchanged.
- Validated against the strict encoder: temporarily applying this fix on #3900 makes `cell-static-methods.test.ts` pass (1/77).

---
## Summary by cubic
Fixes memory boundary encoding by converting native values (e.g., Date, Uint8Array) to Fabric types before serialization so they round-trip correctly instead of flattening to {}. Prevents data loss and keeps `packages/memory` compatible with the stricter codec.

- **Bug Fixes**
  - `encodeMemoryBoundary()` now applies `fabricFromNativeValue()` before `jsonFromValue()` to produce tagged forms like `FabricEpochNsec` and `FabricBytes`.
  - Added tests for Date and Uint8Array round-trips; plain JSON remains unchanged.

<sup>Written for commit cb6f6fcb867f587daf7b905adaf338a4c470f007. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3917?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/memory/test/v2-test.ts
+++ b/packages/memory/test/v2-test.ts
@@ -5,6 +5,10 @@ import {
   setModernCellRepConfig,
 } from "@commonfabric/data-model/cell-rep";
 import {
+  FabricBytes,
+  FabricEpochNsec,
+} from "@commonfabric/data-model/fabric-primitives";
+import {
   compatibleMemoryProtocolFlags,
   decodeMemoryBoundary,
   DEFAULT_BRANCH,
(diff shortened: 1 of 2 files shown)


danfuzz:
Superseded. Rather than convert native values at the memory wire boundary (a redundant deep-walk on every message, papering over a missing upstream guarantee), the agreed direction is: (1) fix the misleading test to round-trip a proper `FabricEpochNsec` (#3919), and (2) once #3900 lands, let the strict codec loudly reject raw "wild west" JS objects (Date/etc.) instead of silently mangling them. The real upstream native→fabric conversion is the separate unified-walk rework.

PR #3920, opened 2026-06-08
feat(web-search): Google Search grounding (drop Jina) + lunch-poll homepage enrichment

willkelly:
## What & why

Replaces the **Jina-backed web-search** with **Google Search grounding** (Gemini), threads grounding through the pattern LLM API, fixes a server-side `apiUrl` bug, and adds **persisted homepage enrichment** to the lunch-poll pattern.

## Changes

**1. `/api/agent-tools/web-search` → grounded Gemini (drops Jina)**
- Sends the query with `tools:[{type:"google_search", google_search:{}}]` and reads the **real** search hits from `grounding_metadata.groundingChunks` (not the model's prose, which hallucinates URLs).
- Resolves each Google redirect to its real destination (which also validates reachability), and skips review/directory/aggregator domains by **brand label** (any TLD — `yellowpages.do`, etc.).
- Response schema unchanged, so the `web_search` subagent consumer is unaffected. `JINA_API_KEY` is no longer used by this route (other Jina routes — webreader/web-read/link-preview — are left alone).

**2. Grounding in the pattern LLM API** (`api` + `runner`)
- `generateText` / `generateObject` / `llm` accept `search: true` (and raw `nativeModelToolIds`) and surface `groundingSources` on result state. The lower `LLMRequest`/`LLMResponse` already carried these; this just exposes them to pattern authors.

**3. Runtime `apiUrl` fix** (`runner/src/runtime.ts`)
- The pattern-visible `apiUrl` defaulted to a hardcoded `localhost:<ports.toolshed>` server-side (in `builder/env.ts`), so **any toolshed not on the default port** handed patterns the wrong `apiUrl` for server-side `fetch`. Now it defaults to the runtime's own configured `apiUrl`. (Browser side already used the page origin and was fine.)

**4. lunch-poll: persisted homepage enrichment**
- Host runs a grounded search; each candidate is normalized to its **origin root** and **LLM-confirmed** ("is this the homepage for X in Y?"), then **persisted on the option** — no per-load LLM.
- Configurable **city** (default Berkeley, CA), a host **"refresh menu links"** button, and a **per-option user override** that always wins. Render reads stored value (override > stored > Google Maps fallback).

## Testing
- Full workspace **test suite green** (`deno task test`, 137.5s) and **typecheck green** (`deno task check`, 259 paths).
- Affected runner tests pass: `env`, `runtime`, `llm-schema-alignment`, `llm-pattern-smoke`, `llm-dialog`.
- lunch-poll verified via `cf check` (the `patterns` package test task is a stub, so it's not in `deno task test`).
- Manually exercised end-to-end against a local toolshed: all seed restaurants resolve to correct official homepage roots; persistence, refresh, and override all work.

## Self-review notes
- 🟡 **No new automated tests** for the new platform surface. The pure helpers are easy to unit-test — `extractGroundingChunks` / `isAggregatorHost` (web-search), the `search→nativeModelToolIds` threading, and the runtime `apiUrl` default — worth a follow-up.
- 🟡 **`setPatternEnvironment` is now called on every `Runtime` construction** (was conditional). It's a global singleton (pre-existing `TODO(seefeld)`); in multi-`Runtime` processes the last one wins. Runner tests pass, but flagging the behavior change.
- ⚪ The aggregator denylist is a hardcoded brand list (whack-a-mole by nature); the LLM homepage-confirm step is the real backstop.

## Notes
- Grounded results are non-deterministic; a place can occasionally come back with no confirmed homepage → the card falls back to a Google Maps link, and the host can re-run or a user can override.

### DIFF
--- a/deno.lock
+++ b/deno.lock
@@ -1799,6 +1799,7 @@
     },
     "uuid@9.0.1": {
       "integrity": "sha512-b+1eJOlsR9K8HJpow9Ok3fiWOWSIcIzXodvv0rQjVoOVNpWMpxf1wZNpt4y9h10odCNrqnYp1OBzRktckBe3sA==",
+      "deprecated": true,
       "bin": true
     },
     "w3c-keyname@2.2.8": {
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -1775,8 +1775,10 @@ export interface BuiltInLLMGenerateObjectState<T> {
   partial?: string;
   error?: unknown;
   cancelGeneration: Stream<void>;
-  /** Web sources from native search grounding, when `search`/`google_search` was requested. */
-  groundingSources?: readonly BuiltInLLMGroundingSource[];
+  // NOTE: `generateObject` accepts `search`/`nativeModelToolIds` (grounding can
+  // improve the structured result), but does NOT surface `groundingSources` —
+  // its JSON-mode path returns only the object, not the grounded response. Use
+  // `generateText` when you need the source URLs.
 }
 
 export interface BuiltInLLMDialogState {
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -1810,7 +1810,8 @@ export type BuiltInGenerateObjectParams =
     /**
      * Enable Google Search grounding (shorthand for the `google_search`
      * native model tool). Real, current web results inform the answer, and
-     * the source URLs are surfaced on the result state's `groundingSources`.
+     * the source URLs are surfaced on the result state's `groundingSources`
+     * (generateText / llm only — generateObject does not surface them).
      */
     search?: boolean;
     /**
@@ -1836,7 +1837,8 @@ export type BuiltInGenerateObjectParams =
     /**
      * Enable Google Search grounding (shorthand for the `google_search`
      * native model tool). Real, current web results inform the answer, and
-     * the source URLs are surfaced on the result state's `groundingSources`.
+     * the source URLs are surfaced on the result state's `groundingSources`
+     * (generateText / llm only — generateObject does not surface them).
      */
     search?: boolean;
     /**
@@ -1859,7 +1861,8 @@ export type BuiltInGenerateTextParams =
     /**
      * Enable Google Search grounding (shorthand for the `google_search`
      * native model tool). Real, current web results inform the answer, and
-     * the source URLs are surfaced on the result state's `groundingSources`.
+     * the source URLs are surfaced on the result state's `groundingSources`
+     * (generateText / llm only — generateObject does not surface them).
      */
     search?: boolean;
     /**
@@ -1880,7 +1883,8 @@ export type BuiltInGenerateTextParams =
     /**
      * Enable Google Search grounding (shorthand for the `google_search`
      * native model tool). Real, current web results inform the answer, and
-     * the source URLs are surfaced on the result state's `groundingSources`.
+     * the source URLs are surfaced on the result state's `groundingSources`
+     * (generateText / llm only — generateObject does not surface them).
      */
     search?: boolean;
     /**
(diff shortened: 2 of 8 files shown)


willkelly:
cf-review via Claude `claude-opus-4-8`, on behalf of @willkelly — addressed all 5 Cubic comments in `1d29113d5`:

- **P1 (main.tsx) unsafe URL scheme:** user link overrides now go through `safeHttpUrl` (auto-prefixes `https://`, rejects `javascript:`/`data:`/etc. before they reach an `href`). Verified: `javascript:alert(1)` → rejected; `dougsdiner.com/menu` → `https://dougsdiner.com/menu`. (No exported helper existed — the same check is only a private method in `cf-link-preview`, and the html render layer doesn't sanitize href schemes; filed as a platform follow-up below.)
- **P1 (runtime.ts) cross-runtime apiUrl bleed:** reverted to setting the process-global pattern env only when explicitly provided; the toolshed now passes `patternEnvironment={apiUrl: API_URL}` so server-side pattern `fetch`es still reach the right toolshed. Server-side enrichment re-verified working.
- **P2 (web-search) include_content ignored:** honored again via best-effort page-text extraction (verified non-empty content returned).
- **P2 (web-search) serial 8s × N resolution:** redirect resolution is now concurrent (capped), picking results in relevance order.
- **P2 (llm.ts) generateObject groundingSources:** dropped the never-populated field from generateObject's state/schema (its JSON-mode path returns only the object); `search`/`nativeModelToolIds` still forward. Use `generateText` for source URLs.

Full suite green (`deno task test`, 156.8s) + typecheck (259 paths).

**Platform follow-up (not this PR):** add href-scheme sanitization to the html render layer (`sanitizeNode`/`bindProps`) with an allowlist (http/https/mailto/tel/relative) so every pattern is protected, not just this one.

willkelly:
cf-review via Claude `claude-opus-4-8`, on behalf of @willkelly — round 2 (commit `ac74fdb28`):

**Fixed**
- **P1 SSRF (web-search):** redirects are now followed *manually* with every hop's host validated (name + DNS-resolved IP) against private/reserved ranges before connecting — blocks `localhost`/`.local`/`.internal` and loopback/private/link-local/CGNAT/IPv4-mapped IPs. Verified predicates block `127.0.0.1`/`169.254.169.254`/`10.x`/`172.16-31`/`192.168`/`::1`/`fd00`/`fe80` and allow public IPs; legit search + `include_content` still work.
- **P2 fetch timeout:** the AbortController now stays armed through the body read (cleanup moved to a `finally` after `res.text()`), so a slow body can't hang past the budget.
- **P2 url scheme regex:** `safeHttpUrl` now parses the value as-is first and only prefixes `https://` on failure, so `example.com:8080` is accepted (verified) while `javascript:`/`data:` are still rejected.
- **P3 doc:** generateObject's `search` param no longer claims `groundingSources` are surfaced.

**Not changing — with reasons**
- **P1 "Runtime no longer defaults pattern apiUrl":** this restores `main`'s *pre-existing* conditional behavior — my earlier commit made it unconditional, which the previous Cubic run flagged as cross-runtime apiUrl *bleed*. The two runs are contradictory because `setPatternEnvironment` is a process-global (pre-existing `TODO(seefeld)`). The correct resolution is what's here: the toolshed passes `patternEnvironment={apiUrl: API_URL}` explicitly, so server-side pattern `fetch`es resolve against the right host; CLI/local runtimes keep `main`'s default. A proper per-runtime environment is a separate refactor.
- **P1 "async handler fights the reactive model":** `enrichHomePages` is an explicit, event-triggered handler doing network I/O on a host click — async handlers are the *documented* idiom for this (`handler<E,C>(async (event, context) => …)`, used by the google/airtable patterns for `await fetch`). The reactivity-purity rules target `computed`/`lift`, not event handlers. The "re-find by id" is ordinary defense against concurrent edits, not reactive drift. The reactive alternative (per-load `fetchData` in render) was explicitly ruled out by the requirement to *persist* enrichments and not re-run the LLM on every load.

Full suite green (`deno task test`, 138.6s).

PR #3926, opened 2026-06-08
refactor: make lift function-first, drop Schema<T> overload (CT-1625)

mathpirate:
Reorders `lift` to **function-first** — callback leads, schemas trail — matching `pattern()`. The schema-bearing, **type-materializing** overloads (where the callback's input type is derived *from* the supplied JSONSchema via `Schema<>`) live in `commonfabric/schema` (`api/schema.ts`); `api/index.ts` carries only the schema-light callback-only form. This mirrors the split `pattern()`/`handler()` already use, so a provided type and a provided schema can never contradict.

## Why

Per CT-1625 + Berni's review: lift's type surface was duplicated and inconsistent. The fix is twofold — (1) make lift function-first (matching `pattern()`), and (2) put the `Schema<>`-materializing overloads in the on-demand `commonfabric/schema` module as the single canonical schema-mode, rather than as ad-hoc overloads on the runtime/facade that could disagree with hand-written types.

The no-input (computed-origin) form is `lift(fn, false)` — `argumentSchema: false` keeps the no-arg application valid; byte-identical runtime spec to the old `lift(false, fn)`.

## Changes

- **`runner/src/builder/module.ts`** — function-first runtime overloads + simplified impl dispatch (callback at arg 0; no `typeof arg === "function"` sniffing).
- **`api/index.ts`** — `LiftFunction` is the schema-light surface: callback-only overloads, no schema params.
- **`api/schema.ts`** — `LiftFunction` augmentation: **callback-first, `Schema<>`-materializing** overloads (`lift(fn, argSchema, resSchema)` materializes the callback input from `argSchema`), mirroring the existing `PatternFunction` augmentation.
- **`ts-transformers/schema-injection.ts`** — emit `lift(cb, argSchema, resSchema, options?)`.
- **`compiled-bundle-verifier.ts`** — callback-slot updated to position 0.
- ~10 runner test files + goldens flipped to function-first. `schema-to-ts.test.ts` gains a **materialization test** for the new schema.ts overloads (with a `@ts-expect-error` negative proof that the callback input is genuinely typed from the schema).

## Two bugs fixed along the way (neither visible in goldens alone)

1. **Schema doubling** — hoisted-and-applied lifts were schema-injected twice (the second pass was masked by a schema-first positional check). Fixed by marking the rebuilt inner lift injected so the re-descent self-skips.
2. **Options scrambling** — a `computed()` that writes to a captured cell emits a trailing `{ materializerWriteInputPaths }` options object. The function-first append put it *before* the schemas, so the runtime read it as the `argumentSchema`, breaking 5 write-capability pattern tests at runtime. Fixed by splicing schemas between the callback and trailing options. Added a regression fixture (`computed-write-capability-materializer`) since no golden covered this path — only `integration pattern-tests` caught it.

## Follow-up: handler is still schema-first

This PR reorders only `lift`. `handler(eventSchema, stateSchema, cb)` remains schema-first; reordering it to match (and applying the same index.ts/schema.ts split) is tracked in CT-1678.

## Verification

- ts-transformers 292/0, runner 547/0, integration pattern-tests 54/54.

---BEGIN COPY-PASTE---
NEW_PERF_BASELINE: test: pattern-unit-4/pattern-unit-tests = 243s
---END COPY-PASTE---

### DIFF
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -1982,29 +1982,25 @@ export type PatternToolFunction = <
   extraParams?: StripCell<E> extends Partial<T> ? Opaque<E> : never,
 ) => PatternToolResult<E>;
 
-// Function-first, matching PatternFunction/HandlerFunction convention: the
-// callback leads, schemas trail and are optional. Schemas are plain JSONSchema
-// values and are NOT materialized into the callback input type. The no-input
-// (computed-origin) form is `lift(fn, false)`: argumentSchema:false keeps the
-// no-input application valid (the runner's isValidArgument check passes on
-// `argumentSchema === false`), which the transformer emits as `lift(fn, false)()`.
+// Public (schema-light) surface, matching PatternFunction's index.ts shape: the
+// callback is the only argument and the types come from the callback itself. The
+// schema-bearing, type-materializing overloads (where the callback's input type
+// is derived FROM the supplied JSONSchema) live in `commonfabric/schema`
+// (api/schema.ts) so that schema and type can never contradict each other — the
+// same split pattern already used for pattern()/handler(). Transformer-emitted
+// `__cfHelpers.lift(...)` is untyped (`__cfHelpers: any`), so it does not depend
+// on these overloads; only authored `lift(...)` calls resolve against them.
 export interface LiftFunction {
   <T, R>(
     implementation: (input: T) => R,
-    argumentSchema?: JSONSchema,
-    resultSchema?: JSONSchema,
   ): ModuleFactory<StripCell<T>, StripCell<R>>;
 
   <T>(
     implementation: (input: T) => any,
-    argumentSchema?: JSONSchema,
-    resultSchema?: JSONSchema,
   ): ModuleFactory<StripCell<T>, StripCell<ReturnType<typeof implementation>>>;
 
   <T extends (...args: any[]) => any>(
     implementation: T,
-    argumentSchema?: JSONSchema,
-    resultSchema?: JSONSchema,
   ): ModuleFactory<StripCell<Parameters<T>[0]>, StripCell<ReturnType<T>>>;
 }
 
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -1982,35 +1982,30 @@ export type PatternToolFunction = <
   extraParams?: StripCell<E> extends Partial<T> ? Opaque<E> : never,
 ) => PatternToolResult<E>;
 
+// Function-first, matching PatternFunction/HandlerFunction convention: the
+// callback leads, schemas trail and are optional. Schemas are plain JSONSchema
+// values and are NOT materialized into the callback input type. The no-input
+// (computed-origin) form is `lift(fn, false)`: argumentSchema:false keeps the
+// no-input application valid (the runner's isValidArgument check passes on
+// `argumentSchema === false`), which the transformer emits as `lift(fn, false)()`.
 export interface LiftFunction {
   <T, R>(
     implementation: (input: T) => R,
+    argumentSchema?: JSONSchema,
+    resultSchema?: JSONSchema,
   ): ModuleFactory<StripCell<T>, StripCell<R>>;
 
   <T>(
     implementation: (input: T) => any,
+    argumentSchema?: JSONSchema,
+    resultSchema?: JSONSchema,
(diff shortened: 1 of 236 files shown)


packages/api/index.ts
```diff
@@ -1909,35 +1909,30 @@ export type PatternToolFunction = <
...
   <T, R>(
     implementation: (input: T) => R,
+    argumentSchema?: JSONSchema,
+    resultSchema?: JSONSchema,
   ): ModuleFactory<StripCell<T>, StripCell<R>>;
 
   <T>(
     implementation: (input: T) => any,
+    argumentSchema?: JSONSchema,
+    resultSchema?: JSONSchema,
```
seefeldb: Nit, but in practice leaving off result schema means we get a true schema instead of the inferred one that the return type here implies.  Maybe we should just remove the one parameter version and fix tests that use it, as I don't think the transformer would emit these, right?

packages/api/index.ts
```diff
@@ -1909,35 +1909,30 @@ export type PatternToolFunction = <
...
   ): ModuleFactory<StripCell<T>, StripCell<ReturnType<typeof implementation>>>;
 
   <T extends (...args: any[]) => any>(
```
seefeldb, line 2002: I wonder whether we still have the corresponding versions in api/schema.ts, which would infer the type from the provided schema.

I'm tempted to say that `api/index.ts` only gets the single parameter version, which is this signature minus the optional parameters and then have one two parameter version with required schemas in schema.ts, which uses the inference to derive the types and so provides type checking in tests. And remove all others, so provided types and schemas never contradict each other?

seefeldb (approved):

packages/api/schema.ts
```diff
@@ -387,13 +388,28 @@ declare module "commonfabric" {
...
-      implementation: (input: Schema<T>) => Schema<R>,
-    ): ModuleFactory<SchemaWithoutCell<T>, SchemaWithoutCell<R>>;
+    // Callback + two schemas: input type from argSchema, result type from resSchema.
+    <IS extends JSONSchema = JSONSchema, OS extends JSONSchema = JSONSchema>(
+      implementation: (input: Schema<IS>) => Schema<OS>,
+      argumentSchema: IS,
+      resultSchema: OS,
+    ): ModuleFactory<SchemaWithoutCell<IS>, SchemaWithoutCell<OS>>;
+
+    // Callback + one schema: input type from argSchema; result type inferred from
```
seefeldb, line 404: i'd also remove that one schema version here, but not a big deal

seefeldb: > The no-input (computed-origin) form is lift(fn, false) — argumentSchema: false keeps the no-arg application valid; byte-identical runtime spec to the old lift(false, fn).

this could be lift(fn, false, <return type>) -- presumably one or the other is not going to be false (and now i guess we make the third argument implicitly be `true`, so might as well be explicit in those cases)

packages/api/index.ts
mathpirate: Resolved in 3edd2912 (the second commit in this PR): `api/index.ts` `LiftFunction` is now schema-light (callback-only, no `argumentSchema?`/`resultSchema?` params), and the schema-materializing overloads moved to `api/schema.ts` — matching the pattern()/handler() split you suggested.

packages/api/index.ts
mathpirate: Done exactly as proposed in 3edd2912: `api/index.ts` keeps only the schema-light callback-only overloads (types from the callback), and `api/schema.ts` carries the materializing overloads where the callback input type is derived FROM the supplied schema — so a provided type and a provided schema can never contradict. Mirrors the existing PatternFunction augmentation.

Your remaining preference — collapse `schema.ts` to a single two-required-schema version and drop the one-schema form — is tracked as a follow-up in CT-1695 (see your newer thread below).

packages/api/schema.ts
mathpirate: Agreed on both — removing the one-schema overload so schema mode always takes both schemas (type and schema can never diverge), and spelling the no-input result explicitly as `lift(fn, false, resultSchema)`. Since this PR is already approved, green, and freshly rebased, I have not folded it in here to avoid re-opening it; tracked as a follow-up in **CT-1695** (Pattern Factory), which captures both of your notes plus the one authored fixture that currently relies on the one-schema form (`opaque-ref-cell-map`).

https://linear.app/common-tools/issue/CT-1695

PR #3942, opened 2026-06-09
feat(ts-transformers): compile-time diagnostics for unknown-typed reactive values (CT-1632A)

Hixie:
When a reactive value's inferred TypeScript type is `unknown`, schema injection emits `{ type: "unknown" }` for it, and the runner reads that back as `undefined` rather than materializing the value — a silent runtime data loss. This adds compile-time warnings so the silent case surfaces at build time, plus the machinery to actually show those warnings.

Capture diagnostic (`reactive-capture:unknown-type`):
- Detect at buildTypeElementsFromCaptureTree, the point where every closure- captured input schema is assembled, so one hook covers computed/lift, handler/action, reactive array methods (.map/.filter/.flatMap), and patternTool.
- A shared helper (warnIfUnknownReactiveType) also runs on the reactive condition of ifElse/when/unless (including the lowered form of JSX conditionals), which type their argument through a separate path.
- `any` is not flagged (it lowers to a permissive `true` schema that materializes); reactive wrappers are unwrapped before judging. Duplicate emissions from cross-stage re-walks collapse via a new reportDiagnosticOnce/CrossStageState dedup.

Result diagnostic (`pattern-result:unknown-type`):
- Warn when an inferred pattern output nests an `unknown` field (e.g. `{ result: unknown }`); the existing pattern:any-result-schema error only caught a top-level any/unknown result. A consumer reading such a field back gets undefined. Scoped to inferred results, like that error.

Warning surfacing:
- compiler.compile/compileAll route warning-severity transformer diagnostics through the logger (errors still throw); `cf` defaults its log floor to `warn` so authors see them. The happy path emits no runtime warnings, so this stays quiet in practice.

Refactors folded in:
- when/unless/ifElse schema injection collapse into one visitReactiveConditional.
- The three permissive-inferred-result call sites collapse into reportUnknownPatternResult (top-level error or nested-field warning).

Tests:
- ts-transformers: 17-step transform-time suite across every covered builder and condition, plus the negative guards (typed, any, passthrough, typed output).
- generated-patterns runtime: a structured value typed `unknown` captured into computed() reads back as undefined while a typed capture survives; and a Cell<unknown> capture carries a primitive but drops a structured payload (documenting why no `asCell` exemption is warranted).
- Refresh the ct-1334 fixture comment now that the diagnostic exists.

---
## Summary by cubic
Adds compile-time warnings for reactive values inferred as `unknown`, and for `unknown` fields in inferred `pattern()` outputs. Warnings show by default in `cf` (log floor `warn`) to address CT-1632.

- **New Features**
  - `reactive-capture:unknown-type`: warns when a captured reactive value or a conditional’s condition is `unknown`. Covers `computed`/`lift`, handlers/actions, reactive array methods, `patternTool`, and `ifElse`/`when`/`unless` (incl. JSX). Skips `any`, unwraps cell-like wrappers, and de-dupes once per file/type/range.
  - `pattern-result:unknown-type`: warns when an inferred `pattern()` output nests `unknown` fields; a top-level `any`/`unknown` result remains an error.
  - Transformer warnings are logged in `@commonfabric/js-compiler`. `cf` now defaults its log floor to `warn` (override with `--log-level` or `CF_LOG_LEVEL`); help text updated. CLI prints concise messages for `TransformerError`/`CompilerError`.

- **Migration**
  - If you see warnings, add explicit types (e.g. `fetchData<{...}>({...})` or `pattern<Input, Output>(...)`).
  - Capturing `Cell<unknown>` warns; primitives survive, structured payloads read back as `undefined`.
  - To quiet logs in CI: `cf --log-level error ...`.

<sup>Written for commit 2d6f3d8853efa4a7882a82f427248d0a59acf2c6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3942?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ts-transformers/src/ast/type-building.ts
+++ b/packages/ts-transformers/src/ast/type-building.ts
@@ -5,8 +5,11 @@ import { createPropertyName } from "../utils/identifiers.ts";
 import {
   ensureTypeNodeRegistered,
   inferWidenedTypeFromExpression,
+  isUnknownType,
+  unwrapCellLikeType,
 } from "./type-inference.ts";
 import {
+  getExpressionText,
   isOptionalMemberSymbol,
   isOptionalSymbol,
   setParentPointers,
@@ -651,6 +654,53 @@ export function cloneTypeNodeDeepForEmission<T extends ts.TypeNode>(
 }
 
 /**
+ * Warns when a reactive value consumed across a boundary has inferred type
+ * `unknown`. That type lowers to `{ type: "unknown" }`, which the runner reads
+ * back as `undefined` rather than materializing the value (see
+ * `runner/src/traverse.ts`, `_traverseWithSchemaInner`: a
+ * `TypeValidity.Unknown` field short-circuits to `undefined`). `any` lowers to
+ * a permissive `true` schema and materializes, so it is not flagged.
+ *
+ * Both reactive-input sites share this: closure-captured inputs (the capture
+ * leaves below — `computed`/`lift`, `handler`/`action`, reactive array methods,
+ * `patternTool`) and the condition of `ifElse`/`when`/`unless` (checked where
+ * their schemas are built). `reportDiagnosticOnce` keeps the same value from
+ * being warned about twice when more than one stage walks it.
+ */
+export function warnIfUnknownReactiveType(
+  context: TransformationContext,
+  expression: ts.Expression,
+  type: ts.Type | undefined,
+  label: string,
+): void {
+  if (!isUnknownType(unwrapCellLikeType(type, context.checker))) {
+    return;
+  }
+  context.reportDiagnosticOnce({
+    severity: "warning",
+    type: "reactive-capture:unknown-type",
+    message:
+      `Reactive value \`${describeCapture(expression, label)}\` has inferred ` +
+      `type \`unknown\`, so its schema is \`{ type: "unknown" }\`, which the ` +
+      `runner reads back as \`undefined\` instead of materializing it. Add an ` +
+      `explicit type so it can be inferred (e.g. ` +
+      `\`fetchData<{ /* result shape */ }>({ ... })\`).`,
+    node: expression,
+  });
+}
+
+/** Single-line rendering of a reactive value for the message; falls back to the
+ * given label when there is no usable text. Uses getExpressionText, which
+ * prints synthetic nodes safely — `getText()` throws on them. */
+function describeCapture(expression: ts.Expression, fallback: string): string {
+  const text = getExpressionText(expression).replace(/\s+/g, " ").trim();
+  if (!text) {
+    return fallback;
+  }
+  return text.length > 48 ? `${text.slice(0, 45)}...` : text;
+}
+
+/**
  * Builds TypeScript type elements from a capture tree structure.
  * Works for both nested properties within a tree node and root-level entries.
  * Recursively builds nested type literals for hierarchical captures.
@@ -689,7 +739,21 @@ export function buildTypeElementsFromCaptureTree(
     if (childNode.expression) {
       // Leaf node with source expression - use it directly
       typeNode = expressionToTypeNode(childNode.expression, context);
-      currentType = checker.getTypeAtLocation(childNode.expression);
+
+      // Judge unknown-ness from the type that drives the emitted schema — the
+      // same one expressionToTypeNode uses. A bare getTypeAtLocation would skip
+      // the typeRegistry lookup and initializer fallback; literal widening here
+      // can't change whether the type is unknown.
+      warnIfUnknownReactiveType(
+        context,
+        childNode.expression,
+        inferWidenedTypeFromExpression(
+          childNode.expression,
+          checker,
+          context.options.state?.typeRegistry,
+        ),
+        propName,
+      );
 
       // Check optionality from source expression
       if (ts.isPropertyAccessExpression(childNode.expression)) {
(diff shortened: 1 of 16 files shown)


Hixie:
also for context, https://github.com/commontoolsinc/labs/pull/3943 removes one of the footguns that leads to these warnings

packages/ts-transformers/src/ast/type-building.ts
```diff
@@ -589,6 +591,58 @@ export function cloneTypeNode<T extends ts.TypeNode>(typeNode: T): T {
...
+    node: expression,
+  });
+}
+
+/** Single-line rendering of a reactive value for the message; falls back to the
+ * given label for synthetic expressions with no source text. */
+function describeCapture(expression: ts.Expression, fallback: string): string {
+  try {
+    if (expression.getSourceFile()) {
+      const text = expression.getText().replace(/\s+/g, " ").trim();
```
mathpirate: this will need to use `getExpressionText()` which i believe is exported from src/ast/utils.ts instead of `getText()`, because `getText()` unfortunately crashes at runtime if it's called on a synthetic AST node

this footgun has been a problem throughout the lifetime of the transformers... i'll take a task to try to add better agent docs about this somewhere they're likely to see it

mathpirate: i think this might be

const text = truncate(getExpressionText(expression))

packages/ts-transformers/src/ast/type-building.ts
mathpirate: it's unclear to me whether this branch is dead or not. but if it's live i think we probably want to warn here too?

packages/ts-transformers/src/ast/type-building.ts
```diff
@@ -630,6 +684,13 @@ export function buildTypeElementsFromCaptureTree(
       typeNode = expressionToTypeNode(childNode.expression, context);
       currentType = checker.getTypeAtLocation(childNode.expression);
 
+      warnIfUnknownReactiveType(
```
mathpirate, line 747: non-blocking, suggestion: the warning judges `currentType` from the bare `checker.getTypeAtLocation`, but the schema emitted two lines up comes from `expressionToTypeNode`, which resolves types differently: typeRegistry lookup for synthetic nodes, the initializer-type fallback in `getTypeAtLocationWithFallback`, and the declared-binding override for Writable/Per*/Default wrappers. fable couldn't find any cases currently where this discrepancy causes differences-in-practice, but it's probably more principled/future-proof for this to be:

warnIfUnknownReactiveType(
  context,
  childNode.expression,
  inferWidenedTypeFromExpression(
    childNode.expression,
    checker,
    context.options.state?.typeRegistry,
  ),
  propName,
);

(Fable says: `inferWidenedTypeFromExpression` is already imported here; literal widening can't change the unknown-verdict. The one residual gap — the declared-binding branch of expressionToTypeNode — can't flip unknown-ness for any of the preserved wrapper types, so this single call is sufficient.)

packages/ts-transformers/src/transformers/schema-injection.ts
```diff
@@ -1437,6 +1438,54 @@ function visitPrependedWidenedSchemaCall(
...
+  }
+
+  // getTypeAtLocationWithFallback handles synthetic nodes (e.g. lift-applied
+  // calls) whose types live in the typeRegistry rather than at a source range.
+  const typeOf = (expr: ts.Node): ts.Type =>
+    getTypeAtLocationWithFallback(expr, checker, typeRegistry) ??
+      checker.getTypeAtLocation(expr);
+
+  const argTypes = args.map(typeOf);
+  warnIfUnknownReactiveType(context, args[0]!, argTypes[0], "condition");
```
mathpirate, line 1523: non-blocking optional request: it would be nice to have a comment here explaining why it's correct to only check the conditional and not the branches, since branches are deliberately exempt because nothing materializes through the branch/result schemas at this boundary. consider also a test step pinning the propagation behavior (unknown branch → no warning at the conditional, warning at the consuming boundary)

packages/ts-transformers/src/ast/type-building.ts
Hixie: https://github.com/commontoolsinc/labs/pull/4095

Hixie:
(I'll continue this on Tuesday since my Claude has run out of quota and that's where the context is.)

packages/ts-transformers/src/ast/type-building.ts
Hixie: done, see also https://github.com/commontoolsinc/labs/pull/4159

packages/ts-transformers/src/ast/type-building.ts
Hixie: addressed

packages/ts-transformers/src/transformers/schema-injection.ts
Hixie: addressed

Hixie:
@mathpirate updated as requested.

mathpirate (approved):
looks great thank you!

PR #3943, opened 2026-06-09
Infer fetchData result type from mode

Hixie:
Make the builder-level `fetchData` signature derive its `result` type from `mode`: "text" -> string, "json"/omitted -> any. When the caller pins the generic `T` (explicitly or via the `result` hint) that still wins, so all existing call sites keep their current types.

Applied to both the public `FetchDataFunction` type in packages/api/index.ts (which the commonfabric.d.ts bundle symlinks to) and the runner-internal factory cast in packages/runner/src/builder/built-in.ts.

Verified: full-repo `deno check` shows no new errors vs the clean tree, and the runner fetchData suites + ct-1334 integration test pass.

---
## Summary by cubic
Infer `fetchData` result type from `mode` to reduce manual typing and improve safety. "text" now yields `string`; "json" or omitted yields `any`, unless a generic `T` or `result` hint is provided.

- **Refactors**
  - Updated public `FetchDataFunction` in `packages/api/index.ts` and runner factory cast in `packages/runner/src/builder/built-in.ts` to use `mode`-aware generics.
  - Regenerated `packages/ts-transformers/test/fixtures/jsx-expressions/ifelse-undefined-value.expected.jsx` so text-mode infers `result: string` and emitted lift types/JSON schemas match; all suites pass (`deno check`, runner `fetchData`, `ct-1334`, ts-transformers).

<sup>Written for commit f0f5b14b14cb211e381117d8e82266b92f6658f1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3943?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -2107,14 +2107,23 @@ export type FetchOptions = {
   method?: "GET" | "POST" | "PUT" | "DELETE" | "PATCH" | "OPTIONS" | "HEAD";
   redirect?: "follow" | "error" | "manual";
 };
-export type FetchDataFunction = <T>(
+export type FetchDataFunction = <
+  T = never,
+  M extends "json" | "text" = "json",
+>(
   params: Opaque<{
     url: string;
-    mode?: "json" | "text";
+    mode?: M;
     options?: FetchOptions;
     result?: T;
   }>,
-) => OpaqueRef<{ pending: boolean; result: T; error?: any }>;
+) => OpaqueRef<{
+  pending: boolean;
+  // When the caller pins T (explicitly or via `result`), use it. Otherwise
+  // derive from mode: "text" -> string, "json"/omitted -> any.
+  result: [T] extends [never] ? ([M] extends ["json"] ? any : string) : T;
+  error?: any;
+}>;
 
 export type FetchProgramFunction = (
   params: Opaque<{ url: string }>,
--- a/packages/runner/src/builder/built-in.ts
+++ b/packages/runner/src/builder/built-in.ts
@@ -133,14 +133,23 @@ export const generateText = createNodeFactory({
 export const fetchData = createNodeFactory({
   type: "ref",
   implementation: "fetchData",
-}) as <T>(
+}) as <
+  T = never,
+  M extends "json" | "text" = "json",
+>(
   params: Opaque<{
     url: string;
-    mode?: "json" | "text";
+    mode?: M;
     options?: FetchOptions;
     result?: T;
   }>,
-) => OpaqueRef<{ pending: boolean; result: T; error?: unknown }>;
+) => OpaqueRef<{
+  pending: boolean;
+  // When the caller pins T (explicitly or via `result`), use it. Otherwise
+  // derive from mode: "text" -> string, "json"/omitted -> any.
+  result: [T] extends [never] ? ([M] extends ["json"] ? any : string) : T;
+  error?: unknown;
+}>;
 
 export const fetchProgram = createNodeFactory({
   type: "ref",
--- a/packages/ts-transformers/test/fixtures/jsx-expressions/ifelse-undefined-value.expected.jsx
+++ b/packages/ts-transformers/test/fixtures/jsx-expressions/ifelse-undefined-value.expected.jsx
@@ -13,7 +13,7 @@ const runtimeDeps = undefined;
 const __cfAmdHooks = undefined;
 const __cfLift_1 = __cfHelpers.lift<{
     pending: boolean;
-    result: unknown;
+    result: string;
 }, boolean>(({ pending, result }) => pending || !result, {
     type: "object",
     properties: {
@@ -21,7 +21,7 @@ const __cfLift_1 = __cfHelpers.lift<{
             type: "boolean"
         },
         result: {
-            type: "unknown"
+            type: "string"
         }
     },
     required: ["pending", "result"]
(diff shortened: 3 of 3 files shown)


packages/runner/src/builder/built-in.ts
```diff
@@ -133,14 +133,23 @@ export const generateText = createNodeFactory({
...
+    mode?: M;
     options?: FetchOptions;
     result?: T;
   }>,
-) => OpaqueRef<{ pending: boolean; result: T; error?: unknown }>;
+) => OpaqueRef<{
+  pending: boolean;
+  // When the caller pins T (explicitly or via `result`), use it. Otherwise
+  // derive from mode: "text" -> string, "json"/omitted -> any.
+  result: [T] extends [never] ? ([M] extends ["json"] ? any : string) : T;
```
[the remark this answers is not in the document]
Hixie: I don't understand what this means.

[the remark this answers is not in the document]
Hixie: Is there a way to disallow `"json" | "text"` entirely?

packages/api/index.ts
```diff
@@ -2114,14 +2114,23 @@ export type FetchOptions = {
...
+    mode?: M;
     options?: FetchOptions;
     result?: T;
   }>,
-) => OpaqueRef<{ pending: boolean; result: T; error?: any }>;
+) => OpaqueRef<{
+  pending: boolean;
+  // When the caller pins T (explicitly or via `result`), use it. Otherwise
+  // derive from mode: "text" -> string, "json"/omitted -> any.
+  result: [T] extends [never] ? ([M] extends ["json"] ? any : string) : T;
```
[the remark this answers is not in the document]
Hixie: this is the same issue as the other comment

mathpirate, line 2124: ok fable has convinced me that cubic was actually right originally here and kinda wrong in its follow-up discussion with you on this point. the problematic case is where a caller constructs a `mode` value that's of union type `"text" | "string"` eg via something like `mode: dyn ? "json" : "text"`, in which case this will evaluate incorrectly to `string` instead of `any` bc the union type doesn't extend `"json"`

fable endorses cubic's original fix, which it tested against some local runtime probes including the above form:

result: [T] extends [never] ? (M extends "json" ? any : string) : T;

this works because of how typescript distributively evaluates type conditionals per member:

- M = "json" → any ✓
  - M = "text" → string ✓
  - M = "json" | "text" → ("json" → any) | ("text" → string) = any | string — and any absorbs everything in a union, so this collapses to any ✓

packages/runner/src/builder/built-in.ts
mathpirate, line 150: same as above, this should be:

result: [T] extends [never] ? (M extends "json" ? any : string) : T;

(side note: disallowing union types entirely here seems worse to me bc it would conflict with a fairly natural authored form where mode is evaluated as a ternary conditional)

packages/api/index.ts
```diff
@@ -2114,14 +2114,23 @@ export type FetchOptions = {
...
+  error?: any;
```
mathpirate, line 2125: small nit: the type declarations for `error` conflict here vs in the built-in (where it's given as `unknown`

i asked @seefeldb in person just now and he thinks it should probably be `unknown`, although this isn't a big deal because he's gonna take on the fancier type magic work we talked about in discord w.r.t. this issue soon, so feel free to just defer this

if you do decide to change this to `unknown` and make them line up in this PR, there's one other small optional cleanup opportunity that leads to below...

packages/runner/src/builder/built-in.ts
```diff
@@ -133,14 +133,23 @@ export const generateText = createNodeFactory({
 export const fetchData = createNodeFactory({
   type: "ref",
   implementation: "fetchData",
-}) as <T>(
+}) as <
```
mathpirate, line 136: if the error types are made to align as mentioned in my other comment, this file can import FetchDataFunction from "@commonfabric/api" and then simplify this whole thing down to just

export const fetchData = createNodeFactory({
    type: "ref",
    implementation: "fetchData",
}) as FetchDataFunction;

packages/api/index.ts
mathpirate: and alright, after another little chat with @seefeldb about this just now, he and i think that we should still be returning `unknown` instead of `any` in the relevant branches of this type conditional - so the final suggestion would become:

result: [T] extends [never] ? (M extends "json" ? unknown : string) : T;

packages/runner/src/builder/built-in.ts
mathpirate: for completeness, as in the thread above, berni suggests this should still yield `unknown` in the "json" case

mathpirate (changes requested):
oops sorry i didn't realize i needed to "submit review" sigh

packages/api/index.ts
Hixie: the idea being that if you went down this path, we should be flagging this as a bad capture, because you essentially didn't say what you wanted?

mathpirate: yeah exactly. if we silently upgrade to "any" we'll deep-crawl whatever object we get back, which is potentially expensive; if that's the desired behavior, the author can explicitly annotate with "any", but the system's fallback behavior should degrade to "unknown" which is a shallow/non-link-following fetch

Hixie:
Upon further consideration I've decided to skip this refactor and go straight to fetchText/fetchJson/fetchBinary.

PR #3947, opened 2026-06-10
feat(storage): per-space host resolution (spaceHostMap)

jkomoros:
## What

Lets one runtime open storage connections to **different hosts per space**. `StorageManager` gains an optional `spaceHostMap` (space DID → host base URL); a space listed in the map has its memory websocket dialed against that host, every other space resolves to the default host exactly as today. The map rides into the worker over `InitializationData`, threaded from `lib-shell` → `RuntimeClient.initialize` → `RuntimeProcessor` → `StorageManager.open`.

This is the storage slice of the cross-host ("federation") direction: the storage layer is already multi-space (`StorageManager.#providers = Map<MemorySpace, Provider>`), so giving the session factory a per-space address resolver is all that's needed for distinct hosts to get distinct providers/connections. Page-layer multi-space and per-space compute routing are follow-ups, not in this PR.

## Design points

- **Additive + default-preserving.** Every new field is optional. Absent map or absent entry ⇒ the resolver returns the default host — byte-identical wire behavior to today. No-map paths (all existing callers) are unchanged.
- **The `/api/storage/memory` join now happens in exactly one place** — `createStorageAddressResolver` in `v2-remote-session.ts`, next to the websocket address builders. Every caller used to do the identical `new URL("/api/storage/memory", base)` join by hand.
- **`Options.address` → `Options.memoryHost`** (host base URL, not full endpoint). The rename is deliberate: with the join moving inside, a stale caller passing a full endpoint must fail at type-check, not silently dial a doubled path. It's `memoryHost` rather than `apiUrl` because toolshed wires `env.MEMORY_URL` here while the adjacent `Runtime.apiUrl` is `env.API_URL` — the two hosts are separable in that deployment.
- **Map entries are validated eagerly** at `StorageManager.open`, so a malformed host fails at configuration time naming the offending space, instead of later inside session creation as a bare `Invalid URL`.
- **`spaceIdentity` stays stored-but-never-read** — connections still authenticate as the user; no per-space signer is introduced.
- The map is **fixed for the manager's lifetime**; the per-space provider cache assumes space → host never changes.

## Security note (pre-existing, scoped out, but relevant)

The signed `session.open` invocation (`{iss, cmd, sub: space, args}`) carries no audience binding — no host, nonce, or expiry — so a host that receives one can replay it to any other host serving that space. That's true today with a single `apiUrl` and this PR doesn't widen the trust model (the map comes from the same embedder that already supplies `apiUrl` and the user `Identity`). But it becomes a real concern the moment host maps are derived from *data* (e.g. another user's space pointer) rather than trusted configuration. Audience-binding the session handshake should land before any data-driven map producer does.

## Tests

- Resolver units: default fallback, mapped override, distinct WS targets per host, host-only (path-ignoring) base semantics, eager rejection of malformed entries (`runner/test/memory-v2-remote-session.test.ts`).
- Wiring test: a real `StorageManager.open` with a map, with a stand-in `WebSocket` recording dialed URLs — the mapped space dials its host, the unmapped space dials the default, both on `/api/storage/memory`.
- Full `deno task check`, `deno lint`, `deno fmt --check` green; runner / runtime-client / lib-shell / piece / html suites and toolshed memory+blobs route tests green (the toolshed route tests exercise the moved join against a live server).

---
## Summary by cubic
Enables per-space storage host routing. `StorageManager` now uses a default `memoryHost` and an optional `spaceHostMap` (space DID → host) to connect different spaces to different hosts; no-map behavior is unchanged. Also stabilizes the wiring test by awaiting a deterministic socket-dial signal (no polling).

- New Features
  - Per-space host resolution via `spaceHostMap`; unmapped spaces use `memoryHost`.
  - Single join point for `/api/storage/memory` via `createStorageAddressResolver`, yielding distinct websocket targets per host.
  - Eager validation of `spaceHostMap` entries with clear errors naming the space.
  - Threads `spaceHostMap` from the shell to the worker (`lib-shell` → `RuntimeClient.initialize` → `RuntimeProcessor` → `StorageManager.open`).
  - Wiring test uses a deterministic “when dialed” signal to avoid flakes.

- Migration
  - Replace `address` with `memoryHost` (pass the host base URL, not the full endpoint). The path is joined internally.
  - If you previously passed a full endpoint, switch to the host base; stale calls will fail at type-check.
  - Optionally supply `spaceHostMap` to override hosts for specific spaces.

<sup>Written for commit 95614ad5f687e2b008609fd4cae4045cc00425fa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3947?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/background-charm-service/cast-admin.ts
+++ b/packages/background-charm-service/cast-admin.ts
@@ -58,7 +58,7 @@ async function castPattern() {
     apiUrl: new URL(toolshedUrl),
     storageManager: StorageManager.open({
       as: identity,
-      address: new URL("/api/storage/memory", toolshedUrl),
+      apiUrl: new URL(toolshedUrl),
     }),
   });
 
--- a/packages/background-charm-service/cast-admin.ts
+++ b/packages/background-charm-service/cast-admin.ts
@@ -58,7 +58,7 @@ async function castPattern() {
     apiUrl: new URL(toolshedUrl),
     storageManager: StorageManager.open({
       as: identity,
-      apiUrl: new URL(toolshedUrl),
+      memoryHost: new URL(toolshedUrl),
     }),
   });
 
--- a/packages/background-charm-service/src/main.ts
+++ b/packages/background-charm-service/src/main.ts
@@ -29,7 +29,7 @@ const runtime = new Runtime({
   apiUrl: new URL(env.API_URL),
   storageManager: StorageManager.open({
     as: identity,
-    apiUrl: new URL(env.API_URL),
+    memoryHost: new URL(env.API_URL),
   }),
   experimental: {
     modernCellRep: env.EXPERIMENTAL_MODERN_CELL_REP,
--- a/packages/background-charm-service/src/main.ts
+++ b/packages/background-charm-service/src/main.ts
@@ -29,7 +29,7 @@ const runtime = new Runtime({
   apiUrl: new URL(env.API_URL),
   storageManager: StorageManager.open({
     as: identity,
-    address: new URL("/api/storage/memory", env.API_URL),
+    apiUrl: new URL(env.API_URL),
   }),
   experimental: {
     modernCellRep: env.EXPERIMENTAL_MODERN_CELL_REP,
--- a/packages/background-charm-service/src/worker.ts
+++ b/packages/background-charm-service/src/worker.ts
@@ -102,7 +102,7 @@ async function initialize(
     apiUrl: new URL(toolshedUrl),
     storageManager: StorageManager.open({
       as: identity,
-      address: new URL("/api/storage/memory", toolshedUrl),
+      apiUrl: new URL(toolshedUrl),
     }),
     patternEnvironment: { apiUrl },
     consoleHandler: consoleHandler,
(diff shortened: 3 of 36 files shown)


seefeldb (approved):
a follow-up pr should look at the remaining uses of the default url and move to this mechanism for any space-bound uses, e.g. blob upload, llm, fetchData base url, etc.

jkomoros:
Thanks! That follow-up is exactly the next planned slice: put a `hostForSpace(space)` resolver on `Runtime` and route the space-bound call sites through it — llm (`sendRequest` per-call apiUrl), fetch-data relative-URL base, the two frozen wish module consts, and blob upload (optional `space` on `UploadBlobRequest`, since blob isn't on the ambient-parentCell path). `getPatternEnvironment()` consumed by authored pattern code stays home-scoped for now. PR coming.

PR #3951, opened 2026-06-10
feat(cf-profile-badge): make the verified identity seal feel alive (lockstep + cursor sheen)

bfollington:
## What

Makes the **verified** identity seal on `cf-profile-badge` feel *alive* — and alive in ways a sandboxed pattern can't reproduce. Today the seal is a static, DID-derived conic aura that only animates on hover. This moves it toward the "foil hologram" trust-mark concept: a treatment whose liveness is driven by host-held state (a session epoch + the host cursor) that user-space patterns can't observe.

Everything here is gated behind the **existing** `represents-principal` verified state — no change to *when* a seal is shown, only *how* a verified one looks — and behind `prefers-reduced-motion`.

## The three liveness signals

- **Lockstep ring rotation** — a slow (26s) drift of the conic rim, seeded to a shared epoch via `animation-delay` so every verified seal on screen rotates *in unison*. A forgery can copy the keyframes but can't read our epoch, so it drifts out of phase ("the chorus it can't join").
- **Intrinsic shimmer sweep** — a soft light band glazing the avatar on a shared 7s clock, so the seal is visibly alive even when the cursor is still.
- **Cursor sheen** — a bright, iridescent reflective hotspot that tracks the host cursor *even when the cursor is far from the badge*. This is the load-bearing anti-forgery signal: a sandboxed pattern only receives pointer events inside its own iframe, so its lookalike can't glint toward a cursor that's elsewhere on screen.

## How it's built (and why it's cheap)

A single shared controller — `seal-liveness.ts`:

- Lockstep animations are **pure CSS** (no per-frame JS). Driven by per-element `@keyframes` seeded to a shared epoch rather than an animated `:root` custom property — the latter forces a style recalc on every consumer each frame (measured: even rotation dropped 80→57fps at 400 seals).
- The cursor sheen uses **one** global `pointermove` listener + **one** rAF loop, started only while ≥1 seal is registered. It touches only on-screen verified seals (offscreen-culled) and **skips all writes** for seals the cursor is far from.
- Deliberately **avoids `filter: hue-rotate`** for a "lockstep colour" effect — it's a per-pixel paint pass that roughly halved framerate in a 400-seal stress test, and stays costly even on GPU.
- SSR/test-safe: the controller bails when there's no `requestAnimationFrame`.

## Testing

- `deno check` ✓ · `deno lint` ✓
- `deno test` ✓ — existing badge + identity-seal suites unchanged, plus a new `seal-liveness.test.ts` (which caught + fixed an epoch-ordering bug in phase seeding).
- The **visual** was iterated and signed off in a standalone prototype using the identical CSS/JS recipe before integration.

## Reviewer note

I couldn't render the *integrated* component in a standalone bundle harness — `felt`/esbuild leaves Lit decorators untranspiled for raw browser, so the bundle won't parse (a build-pipeline quirk, not a component issue; the badge uses the same decorators it already shipped with). Best evaluated in the **shell preview** with a profile that carries a `represents-principal` attestation. Move your cursor anywhere on the page (not just over the badge) and watch the seal glint toward it.

---
## Summary by cubic
Makes the verified identity seal in `cf-profile-badge` feel alive with an always-on, host-cursor-driven sheen, plus hover-gated ring rotation and shimmer. Strengthens the trust signal and honors reduced motion live.

- **New Features**
  - Verified-only liveness; no change to verification logic.
  - Cursor sheen stays on; tracks the host cursor even off-badge.
  - Ring rotation (26s) and shimmer sweep (7s) now run on hover.

- **Implementation**
  - Added `seal-liveness.ts`: one passive `pointermove` and one rAF; runs only while ≥1 seal is registered, offscreen-culls, and skips distant writes.
  - Live `prefers-reduced-motion`: loop tears down and clears highlights on enable; resumes on disable; SSR-safe.
  - Dropped lockstep epoch seeding; motion remains pure CSS.
  - New test `seal-liveness.test.ts` ensures single-listener/rAF startup and teardown.

<sup>Written for commit cfb5bb5b6038b1a57c7f19e0940587507f85e2a0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3951?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ui/src/v2/components/cf-profile-badge/cf-profile-badge.ts
+++ b/packages/ui/src/v2/components/cf-profile-badge/cf-profile-badge.ts
@@ -16,6 +16,13 @@ import {
   readCfcLabelView,
 } from "../../core/cfc-label.ts";
 import { type IdentitySeal, identitySeal } from "./identity-seal.ts";
+import {
+  registerSeal,
+  sealGlowDelay,
+  type SealLivenessClient,
+  sealSpinDelay,
+  unregisterSeal,
+} from "./seal-liveness.ts";
 
 /** Verification state of the presented identity. */
 export type ProfileBadgeState = "presented" | "verified" | "unverified";
@@ -77,7 +84,7 @@ export const profileDisplayFromValue = (val: unknown): ProfileBadgeDisplay => {
  * @element cf-profile-badge
  * @attr {string} size - avatar size: xs | sm | md | lg | xl (default md)
  */
-export class CFProfileBadge extends BaseElement {
+export class CFProfileBadge extends BaseElement implements SealLivenessClient {
   static override styles = [
     BaseElement.baseStyles,
     css`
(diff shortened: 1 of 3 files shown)


bfollington:
**cf-review via Claude Opus 4.8, on behalf of @bfollington** — self-review of this branch (author pass; the two read-only review subagents I spawned had their final reports eaten by a local stop-hook, so weight independence accordingly).

**Scope:** 1 modified (`cf-profile-badge.ts`) + 2 new (`seal-liveness.ts`, `seal-liveness.test.ts`); platform code (Lit component, plain TS). **Coverage:** deep-read. **Checks (deno 2.8.1 = CI version):** `fmt --check` ✓ · `lint` ✓ · `check` ✓ · `test` ✓ (23 steps).

### 🔴 Blocking
None. No correctness / regression / anti-duplication defect. The only red CI signal was the `Check` job's `deno fmt --check` on the two new files — fixed in `0a1a8ec`.

### 🟡 Improvements
1. **Stale comment fixed** — the `.aura` comment said *"spin on hover"* but the ring now rotates always (lockstep). Reworded in `c182539` so it no longer misleads about runtime behavior.
2. **`prefers-reduced-motion` is only sampled at register time** (`seal-liveness.ts` `registerSeal`). The CSS spin/glow honor a *live* media-query toggle, but the JS cursor loop does not: enabling reduced-motion mid-session leaves the sheen loop running. Low harm; proposed follow-up: a `matchMedia(...).addEventListener("change", …)` that tears the loop down. *(not in this PR)*
3. **The controller is untested** — the new test covers only the pure delay / reduced-motion helpers; `registerSeal`/`unregisterSeal` teardown (no rAF/listener leak across mount→unmount) and the `updateSeal` sheen math are validated only visually. Proposed follow-up: a deno-web-test asserting the listener + rAF are removed when the last seal unregisters. *(not in this PR)*

### ⚪ Nits
- `updateSeal` does a per-seal `getBoundingClientRect()` every frame *before* the far-cursor skip → O(verified-on-screen) layout reads/frame. Fine for the realistic handful/dozens; could thrash at hundreds. Possible follow-up: an `IntersectionObserver`-backed visible-set instead of per-frame rects.

### Framework-fit (considered, justified)
The module-global singleton in `seal-liveness.ts` is intentional and *required*: lockstep needs one shared clock + cursor channel across all instances; a per-instance controller would defeat the feature. SSR/test-safe (bails without `requestAnimationFrame`).

### Question for reviewers
Always-on rotation across a dense roster of verified people could read as busy — keep it, or make rotation hover-only / disabled below a size threshold? Genuine UX call, not a defect.

### Verdict
**Ship-able.** No blockers; CI green after the fmt fix; the one misleading comment is corrected. #2/#3 + the IntersectionObserver idea are proposed as follow-ups, not gating.

PR #3961, opened 2026-06-10
feat(runner): fold parent slot's declared scope into serialized sub-pattern binding aliases

seefeldb:
Follow-up to #3958.

## Problem

Binding aliases serialized into a sub-piece's argument doc drop the parent argument schema's declared cell scope: the stored alias pins `scope:"space"` and carries the sub-pattern's scope-silent input schema instead of the parent's `PerUser`/`PerSession` annotation (emitted as `asCell: [{kind:"cell", scope:"user"}]`). The two write-path fixes in #3958 (`2833c6042`, `4227d4abe`) make this harmless at write time — but only by way of the stored base-slot redirect. The serialized graph itself was not self-describing: any new consumer of those links had to rely on that redirect existing.

## Decision: fold onto the serialized **schema**, not the link's `scope` field

The codebase states the doctrine in three places (`ContextualFlowControl.getSchemaScopeCap` doc, `link-resolution.ts`, the `key()`-stamping test): **scope is a schema-level declaration realized at read (follow cap) and write (narrowing branch); it must never be stamped onto a navigated link's own scope.** Stamping the link's scope field re-addresses the parent's base-scope slot to a narrower instance — which breaks passed-in cell references, which legitimately live at the base slot ("lift can read session-scoped cell passed from pattern input" in pattern-scope.test.ts; this is exactly how a naive fix failed).

This is also spec-compliant: per `docs/specs/scoped-cell-instances.md`, a serialized link carries `scope` only when its target is *not* at the inherited scope. The alias targets the parent's base-scope slot (inherited scope), so the link's scope field stays omitted; the slot's narrower *content* scope is the schema's to declare.

## Mechanics

- The annotation was being lost because the argument/internal meta links carry **sanitized** schemas (`sanitizeSchemaForLinks` strips `asCell` entries, taking their scope with them). The authored pattern schemas keep `asCell` (`keepAsCell: true`), so `instantiatePatternNode` now threads them into `unwrapOneLevelAndBindtoDoc` as `options.sourceSchemas`.
- `foldDeclaredScopeIntoLinkSchema` folds the authored slot's scope cap into the emitted link schema as a top-level `scope` (which survives link-schema sanitization), exactly when the write-path narrowing branch would fire for the slot (declared scope narrower than the slot's link scope) and the emitted schema declares no scope of its own.
- The fold result is deterministic (derived from authored schemas), so re-instantiation across sessions serializes identically — no reload churn.
- With the folded schema, a write through the stored alias takes the scope-narrowing branch on its own; #3958's stored-redirect-following write fix remains as belt-and-braces for truly schema-less handles.

Known limitation (follow-up candidates): aliases that point *below* a scoped slot, slots whose emitted link carries no schema at all, and raw-node input bindings (the immutable inputs cell) still rely on the stored redirect.

## Verification

- New red→green test: the stored alias in the sub-piece's argument doc keeps `scope:"space"` on the link and carries cap `"user"` on its schema; end-to-end, a consumer writing through the stored alias with only its serialized schema lands content in the user instance, base slot keeps the redirect.
- `packages/runner` full suite: 554 passed, 0 failed.
- `packages/runner/test/pattern-scope.test.ts`: 39 passed (incl. the passed-in-reference test the naive fix broke, and #3958's redirect tests).
- `cfc-group-chat-demo-multi-runtime.test.ts` (headless): passed (re-run on the rebased branch).
- `cfc-group-chat-demo-two-browsers.test.ts` against local dev servers running this change: passed.
- `cfc-group-chat-demo.test.ts` (single-runtime) fails identically on plain `main` ("No data at cell" during `startPiece`) — pre-existing, unrelated; flagged separately.

---
## Summary by cubic
Fold the parent slot’s declared cell scope into the serialized schema of sub-pattern binding aliases so stored links are self‑describing and don’t rely on base-slot redirects. `link.scope` stays inherited; scope is declared on the alias schema per design.

- **New Features**
  - Add `foldDeclaredScopeIntoLinkSchema` to write the parent’s `PerUser`/`PerSession` scope into the alias link’s schema when the emitted schema is scope‑silent and the parent’s scope is narrower than the link’s.
  - Pass authored `pattern.argumentSchema` into `unwrapOneLevelAndBindtoDoc` via `sourceSchemas` in `Runner.run` to recover `asCell` scope lost by `sanitizeSchemaForLinks`.
  - Keep `link.scope` unchanged; reads/writes realize scope from the schema.

<sup>Written for commit d0d15b454aee6e723c4317db483da02a153404ed. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3961?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

NEW_PERF_BASELINE: step: Build application = 25s
NEW_PERF_BASELINE: subtest: pattern-unit-4/pattern-unit-tests > packages/patterns/lunch-poll/main.test.tsx = 22s

### DIFF
--- a/packages/runner/src/pattern-binding.ts
+++ b/packages/runner/src/pattern-binding.ts
@@ -40,6 +40,73 @@ type SendValueToBindingOptions = {
 type UnwrapOneLevelOptions = {
   targetSchema?: JSONSchema;
   derivedInternalCells?: readonly DerivedInternalCellDescriptor[];
+  /**
+   * The containing pattern's authored argument schema, used as the source of
+   * declared cell scopes when serializing binding aliases (see
+   * `foldDeclaredScopeIntoLinkSchema`). The argument cell LINK only carries a
+   * sanitized schema (`sanitizeSchemaForLinks` strips `asCell` entries,
+   * taking their scope annotation with them), so a `PerUser<Cell<>>`
+   * declaration is invisible on the link by the time aliases are bound. The
+   * authored pattern schema keeps `asCell` (`keepAsCell: true` in
+   * builder/pattern.ts) and is the ground truth for what each slot declared.
+   * (Internal cells don't need this: each derived internal cell carries its
+   * declared scope on its descriptor, realized directly on its link.)
+   */
+  sourceSchemas?: {
+    argument?: JSONSchema;
+  };
+};
+
+/**
+ * Folds the source slot's declared cell scope into the serialized alias link's
+ * schema, making the stored link self-describing.
+ *
+ * A binding alias serialized into a sub-piece's argument doc carries the
+ * sub-pattern's (typically scope-silent) input schema, not the parent slot's
+ * schema — so the parent's `PerUser`/`PerSession` declaration (emitted as
+ * `asCell: [{kind, scope}]`) is dropped, and any consumer of the stored link
+ * must rely on the stored base-slot redirect existing to land reads and writes
+ * in the scoped instance. Folding the declared scope into the link's schema
+ * (as a top-level `scope`, which survives link-schema sanitization where
+ * `asCell` entries do not) keeps the serialized graph self-describing: writes
+ * through the link take the scope-narrowing branch and reads get the follow
+ * cap, per "scope lives in the schema, realized at read/write".
+ *
+ * The scope is deliberately NOT stamped onto the link's own `scope`: the link
+ * addresses the base-scope slot, where passed-in cell references legitimately
+ * live (see "lift can read session-scoped cell passed from pattern input" in
+ * pattern-scope.test.ts, and the matching guidance on
+ * ContextualFlowControl.getSchemaScopeCap).
+ *
+ * Folding applies exactly when the write-path narrowing branch would fire for
+ * the slot (declared scope narrower than the slot's link scope) and the
+ * emitted schema does not declare a scope of its own (a local declaration
+ * wins). Slots whose emitted link carries no schema at all are left alone so
+ * they keep inheriting the reader's schema during link resolution.
+ */
+const foldDeclaredScopeIntoLinkSchema = (
+  cfc: ContextualFlowControl,
+  link: NormalizedFullLink,
+  authoredRootSchema: JSONSchema | undefined,
+  path: readonly string[],
+): NormalizedFullLink => {
+  if (authoredRootSchema === undefined || !isRecord(link.schema)) return link;
+  if (ContextualFlowControl.getSchemaScopeCap(link.schema) !== undefined) {
+    return link;
+  }
+  const authoredSlotSchema = path.length > 0
+    ? cfc.getSchemaAtPath(authoredRootSchema, [...path])
+    : authoredRootSchema;
+  const declaredCap = ContextualFlowControl.getSchemaScopeCap(
+    authoredSlotSchema,
+  );
+  if (
+    !isCellScope(declaredCap) ||
+    scopeRank(declaredCap) <= scopeRank(link.scope)
+  ) {
+    return link;
+  }
+  return { ...link, schema: { ...link.schema, scope: declaredCap } };
 };
 
 const scopedLinkForPath = (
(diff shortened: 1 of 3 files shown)


seefeldb:
Rebased onto current main (52 commits, including the ESM/AMD refactor #3992 and the identity refactors #4006/#4008/#4013) and re-validated that the change is still needed and still correct:

- **Still load-bearing:** running the new test against main's `pattern-binding.ts`/`runner.ts` (post-refactor) still fails — the serialized alias remains scope-silent without the fold.
- **Premises intact:** `sanitizeSchemaForLinks` still strips `asCell` from link schemas (with the CT-1623 no-promotion comment), authored pattern schemas still keep `asCell` (`keepAsCell: true`), and no new alias-serialization sites appeared; the only `pattern-binding.ts` changes since were the `unsafe_originalPattern`/`unsafe_parentPattern` deletions, orthogonal to the alias-emission path.
- **Re-verified on the rebase:** runner suite 580 passed / 0 failed; pattern-scope suite 39 passed; `cfc-group-chat-demo-multi-runtime` passed.

seefeldb:
Rebased onto current main again (17 commits, notably #3911 "Assign each internal its own cell", which conflicted with this change). Resolution notes:

- **`sourceSchemas` is now argument-only.** #3911 removed the `internal` pseudo-alias: internal slots are no longer sub-paths of one shared internal doc, so the old `sourceSchemas.internal` folding case is structurally gone rather than lost — each derived internal cell's descriptor carries its declared `scope`, realized directly on the derived cell's link (see the `partialCause` branch in `unwrapOneLevelAndBindtoDoc`). The fold now applies only to `alias.cell === "argument"`, which is the case the PR exists for.
- **Still load-bearing:** the new test run against main's post-#3911 sources still fails — argument aliases serialized into a sub-piece's argument doc remain scope-silent without the fold.
- **Re-verified on the rebase:** runner suite 585 passed / 0 failed; pattern-scope suite 39 passed; `cfc-group-chat-demo-multi-runtime` passed.

seefeldb:
Note on the failing `CLI Integration Tests (fuse)` check — it is a main-side flake, not this PR:

- The same job fails intermittently on main's own runs with the identical signature ("Timed out waiting for path: …/pieces/Fuse-Exec-Fixture" after a successful mount), e.g. main runs 27364686578 (6d70a519a) and 27324634718 (138524382); failures started in the window where #3911 landed.
- The fuse-exec fixture declares no scoped slots, so this PR's fold is a structural no-op for it (it only fires when a parent slot declares a narrower scope).
- The full `integration/fuse-exec.sh` suite passes locally on this branch (FUSE-T on macOS against a local toolshed), including the exact piece-listing step that times out in CI.

Flagged the flaky job separately for investigation (suspect: piece-listing race in the FUSE harness aggravated around #3911).

seefeldb:
Control result: reran the failed `CLI Integration Tests (fuse)` job on **main's own latest run** (27364686578, commit 6d70a519a — zero changes from this PR) and it **failed again** with the identical piece-listing timeout. The job is broken/highly flaky on main itself right now; this PR's red fuse check is not caused by this change (the fixture has no scoped slots, so the fold is provably outside its code path, and the full suite passes locally on this branch). Safe to merge on that basis, or wait for the main-side fuse fix and rerun.

seefeldb:
Perf-check note for the lunch-poll subtest flag (+30.6% on CI, threshold crossed by 0.3s): measured locally with an interleaved A/B (3+2 runs on this branch vs 3 runs on main's runner sources, same machine), the real delta is **~+4%** (≈2.36s → ≈2.46s). lunch-poll is PerUser/PerSpace-heavy, so the scope folding legitimately does a bit more work there (authored-schema walks at sub-pattern instantiation + the narrowing writes the folded schemas trigger); CI runner variance amplified that into the +30% reading. Accepting via NEW_PERF_BASELINE with this disclosure.

PR #3967, opened 2026-06-10
fix(runner): sync handler asCell input docs before dispatching events

mathpirate:
**For Robin & Berni** — this is a worked exemplar of a runtime-level fix for an invariant gap we root-caused while debugging the lunch-poll SQLite deploy bug, offered to kickstart a design discussion. It is fully functional and verified, but the design choice deserves your eyes before it's a real candidate. Companion PR #3962 (CLI-side narrow unblock) describes the original bug; this PR addresses the underlying invariant.

## The invariant gap

A handler reads its asCell inputs (e.g. a `SqliteDb` handle) **synchronously** from the local replica, but nothing in the system guarantees those docs are locally available when the handler body runs:

- Piece-start sync (`syncCellsForRunningPattern`) covers node **binding** docs, not the docs behind link **values** — e.g. a builtin's result handle stored at its output spot. Instrumented on the failing pattern: 396 node cells synced, the handle doc never among them.
- The pulls kicked off during input materialization (`getRaw`/`get`'s non-awaited `this.sync()`) are connected to nothing.
- `pull()` had a second, subtler hole: an entry covered by an already-registered selector resolved **immediately**, even when the covering watch was still in flight — so even a caller that diligently awaited `cell.sync()` could proceed before the doc arrived.

On a cold client (`cf piece call`), the doc-carrying watch response loses the race against dispatch *deterministically on bulky patterns* (serialized watch batches delay it past `send()`): timing trace showed dispatch at T+20ms and the doc landing at T+152ms, after the scheduler had already burned its retry. Result: `db.exec` → "invalid database handle", 100% reproducible, only on deploy (the emulated runner's loopback storage always wins the race).

## Two candidate designs we considered

**A — pre-sync at dispatch (this PR, proactive).** Ensure inputs are present *before* the handler body runs:
1. `pull()` now awaits covering in-flight watch promises — restoring `sync()`'s contract that *resolved ⇒ data locally available*. No-op for settled coverage.
2. `EventHandler.presyncInputs` (built in `instantiateJavaScriptHandlerNode`): materialize the argument with the existing machinery — asCell fields surface as `Cell`s *without* reading their backing docs — walk the materialized structure, await `sync()` on each collected Cell.
3. `dispatchQueuedEvent` awaits `presyncInputs` before invoking, fail-open (a presync error surfaces as the handler's own read failure, not a dropped event).

**B — retry on unsynced read (reactive, not implemented).** Follow the `frame.pendingSpaceNames` → `resolvePendingSpaceNamesAndRetry` precedent: when a read like `.exec()` finds a doc that was *never received* (distinguishable in storage: no confirmed record, sync in flight), record it on the frame; abort, await the sync, requeue via the existing `RetryImmediately` machinery.

**Why we lean A:** it installs the invariant at the right anchor — "inputs synced at dispatch" — once, rather than teaching individual read sites to recover; B must be wired read-site by read-site (`.exec` first, then whatever's next) and re-runs cold handlers twice by design. B remains attractive as a *backstop* for anything A's enumeration misses (e.g. reads through lazy query-result proxies, which this presync deliberately treats as best-effort), and the two compose cleanly.

## Costs / open questions for you

- **Event latency:** dispatch now awaits presync. Steady-state it's ~free (covered selectors resolve in a microtask — measured single-dispatch, no retry, on the testbed), but a cold-replica event pays the sync round trip up front. Is that acceptable as a universal rule, or should it be opt-in per surface?
- **Proxy subtrees:** materialized records all carry the back-to-cell symbol, so the walk can't distinguish a lazy query-result proxy from an annotated plain object; it descends both with guarded property access. A cheap `isQueryResultProxy` predicate would make this exact.
- **`synced`-flag inheritance:** `validateAndTransform` propagates the parent's `synced` flag onto asCell children whose docs were never fetched — a white lie that also suppresses `getRaw`'s fallback re-sync. Harmless once dispatch presyncs, but worth tightening?
- Is the `pull()` covered-in-flight await (change 1) something you'd want **regardless** of A-vs-B? We believe yes — it's a contract fix on its own.

## Verification

- Testbed = main + lunch-poll SQLite pattern, **no** CLI workaround: `clearHistory`/`logVisit` go from 100% "invalid database handle" to passing; timing trace shows presync issuing the pull, awaiting it (61ms), then a single successful dispatch; rows verifiably land in (and are deleted from) the per-piece sqlite file across repeated fresh CLI sessions.
- Runner suite 553/553 (+2 new tests: presync-ordering, fail-open), `cf test` lunch-poll 17/17, CLI suite 43/43, fmt/lint/check clean.

Full investigation log: `session_outputs/2026-06-09_lunch-poll-sqlite-fresh-eyes/` (local to Gideon's machine); bug history in `SQLITE-DEPLOY-BUG.md` on `gideon/lunch-poll-sqlite-migration`.

---
## Summary by cubic
Pre-sync handler `asCell` inputs before dispatch so required docs are local, fixing cold-replica races (e.g. "invalid database handle"). Also make `sync()` reliable by awaiting in-flight coverage in storage `pull()` and deduping concurrent pulls on the full wait.

- **Bug Fixes**
  - Added optional `presyncInputs` on event handlers. It materializes the argument, collects `Cell`s, and awaits `sync()`. The scheduler awaits this before invoking, and fails open if it errors.
  - Updated storage `pull()` to enforce "resolved ⇒ data local." It awaits covering in-flight watches, waits for both fetched and covered entries in mixed batches, and dedupes concurrent same-key pulls onto that combined wait.
  - Added tests for presync ordering and fail-open behavior.

<sup>Written for commit de6e2078d7bf2b122d70e1aefa84ef8e78cd06e3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3967?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---BEGIN COPY-PASTE---
NEW_PERF_BASELINE: step: Type check = 88s
NEW_PERF_BASELINE: job: Check = 115s
---END COPY-PASTE---

### DIFF
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -3083,11 +3083,65 @@ export class Runner {
       setRunnableName(handler, `handler:${name}`, { setSrc: true });
     }
 
+    // Ensure the handler's input docs are locally available before the body
+    // runs: materialize the argument the same way the handler will (asCell
+    // fields surface as Cells WITHOUT reading their backing docs), then await
+    // sync() on each collected Cell. The scheduler awaits this before
+    // dispatching the event. Without it, a synchronous in-handler read of an
+    // asCell input (e.g. SqliteDb.exec reading the handle doc) races the
+    // doc-carrying storage response on a cold replica — piece-start sync
+    // (syncCellsForRunningPattern) covers node binding docs, not the docs
+    // behind link VALUES like a builtin's result handle. Steady-state this is
+    // ~free: covered selectors resolve without a server round trip.
+    const presyncInputs = module.argumentSchema !== undefined
+      ? async (event: any): Promise<void> => {
+        const eventInputs = {
+          ...(inputs as Record<string, any>),
+          $event: event,
+        };
+        const inputsCell = this.runtime.getImmutableCell(
+          processCell.space,
+          eventInputs,
+          undefined,
+        );
+        const argument = inputsCell.asSchema(module.argumentSchema!).get();
+        const promises: Promise<unknown>[] = [];
+        const seen = new Set<unknown>();
+        const collect = (value: unknown, depth: number): void => {
+          if (depth > 16) return;
+          if (isCell(value)) {
+            const maybePromise = value.sync();
+            if (maybePromise instanceof Promise) promises.push(maybePromise);
+            return;
+          }
+          // NOTE: materialized records all carry the back-to-cell symbol, so
+          // there is no cheap way to tell a lazy query-result proxy from an
+          // annotated plain object — descend both. Property access on a proxy
+          // is an ambient local read (it may kick off, but never await, a
+          // sync); guard each access so one lazy read failing doesn't abort
+          // the rest of the presync.
+          if (!isRecord(value)) return;
+          if (seen.has(value)) return;
+          seen.add(value);
+          for (const key of Object.keys(value)) {
+            try {
+              collect((value as Record<string, unknown>)[key], depth + 1);
+            } catch {
+              // A lazy read through a not-yet-synced link may throw; skip.
+            }
+          }
+        };
+        collect(argument, 0);
+        await Promise.all(promises);
+      }
+      : undefined;
+
     const wrappedHandler = Object.assign(handler, {
       reads,
       writes,
       module,
       pattern,
+      ...(presyncInputs !== undefined && { presyncInputs }),
     });
 
     const schedulerReads = this.collectArgumentSchedulerReadLinks(
--- a/packages/runner/src/scheduler/events.ts
+++ b/packages/runner/src/scheduler/events.ts
@@ -426,6 +426,21 @@ export async function dispatchQueuedEvent(state: {
   });
   state.eventQueue.shift();
 
+  // Ensure the handler's input docs are locally available before the body
+  // runs (see EventHandler.presyncInputs). Fail open: a presync error should
+  // surface as the handler's own read failure, not silently drop the event.
+  if (typeof handler.presyncInputs === "function") {
+    try {
+      await handler.presyncInputs(eventValue);
+    } catch (error) {
+      logger.warn(
+        "scheduler",
+        "handler input presync failed; dispatching anyway",
+        { error, handlerId },
+      );
+    }
+  }
+
   const tx = state.runtime.edit();
   tx.tx.immediate = true;
   const actionId = state.getActionId(action);
--- a/packages/runner/src/scheduler/types.ts
+++ b/packages/runner/src/scheduler/types.ts
@@ -44,6 +44,15 @@ export type EventHandler =
       tx: IExtendedStorageTransaction,
       event: any,
     ) => void;
+    /**
+     * Optional callback to ensure the handler's input docs are locally
+     * available before the handler body runs. A handler reads its asCell
+     * inputs (e.g. a SqliteDb handle) synchronously from the local replica;
+     * the scheduler awaits this before dispatching the event so those reads
+     * don't race the doc-carrying storage responses. The event is passed so
+     * inputs reachable only through the event can be covered too.
+     */
+    presyncInputs?: (event: any) => Promise<void>;
   };
 export type AnnotatedEventHandler = EventHandler & TelemetryAnnotations;
 
--- a/packages/runner/src/storage/v2.ts
+++ b/packages/runner/src/storage/v2.ts
@@ -1176,21 +1176,37 @@ class SpaceReplica implements ISpaceReplica {
       promise: Promise.resolve({ ok: {} } as Result<Unit, PullError>),
     };
     const cfc = new ContextualFlowControl();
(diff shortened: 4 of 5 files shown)


mathpirate:
Context: Berni reviewed the companion CLI PR (#3962) and wrote (relayed from chat, before reading this PR):

> I think this fix is fine, but in general having handlers wait for everything to be synced would be bad, they should only wait until the local state is caught up with its own changes that affect the handler. Could be that it doesn't do that for cells?
>
> Re "and .exec() is a sync API that cannot recover mid-handler": Hmm, it should just add stuff to the TX and if the TX fails later on the server, it should just retry like handlers already do after hitting a conflict or so. So maybe something is borked in that error path?

Three responses, since these bear directly on this PR's design discussion:

**1. "Waiting for everything would be bad" — agreed, and this PR doesn't.** The global drain is the CLI PR's coarse one-shot unblock. This PR's `presyncInputs` awaits only the docs behind the *handler's own* asCell inputs, discovered by materializing the argument through the existing machinery. Steady-state (coverage settled) it resolves in a microtask with no round trip — measured single-dispatch on the testbed. The remaining honest cost: a **cold-replica** event pays its inputs' sync round trip before dispatch. Whether that gating is acceptable as a universal rule is exactly the A-vs-B question below.

**2. "It should just retry — maybe something is borked in that error path?" — yes, and we can say precisely what.** The timing trace on the failing pattern shows the retry machinery *already fires*: the event dispatched **twice** (first throw → scheduler retry → second throw), and the handle doc arrived **3ms after the second failure**. So the error path isn't missing a retry — the retry is *immediate and bounded*, so it deterministically re-loses the same race. The B-shaped fix is to gate the requeue on the in-flight sync instead (the existing `frame.pendingSpaceNames` → `resolvePendingSpaceNamesAndRetry` shape: record the never-received cell, await its `sync()`, requeue via `RetryImmediately`). Note that B only becomes *reliable* with this PR's `pull()` change: before it, an entry covered by a still-in-flight watch resolved immediately, so even a diligent await-then-retry couldn't work. That contract fix is load-bearing under either design philosophy — we'd argue for it regardless.

**3. "Just add stuff to the TX" points at a third option worth naming (C).** Today `.exec()` *can't* be purely optimistic: it needs the handle **value** (`{id, tables, scope}`) to construct the wire op — the missing data is the payload. But the `id` is literally the handle cell's own entity id, and the server has the doc. An op constructed from the **link**, with the server resolving `tables`/`scope` from the handle doc at commit time, would make `.exec()` value-free: dispatch never gates, the op rides the TX, and a stale/absent local view converges through the existing conflict machinery. That's the most optimistic-flavored design of the three — but it's sqlite-specific, and it gives up the client-side validation that the cell really is a sqlite handle (a bad target would surface as a server-side commit failure → retry, which may be fine by the same philosophy).

**Updated framing of the design space:**
- **A (this PR):** installs the general invariant — *a handler's asCell input docs are locally available at dispatch* — once, at the dispatch anchor. Cost: cold-replica events gate on their inputs' sync.
- **B:** makes reads recoverable — dispatch immediately, and when a read finds a never-received doc, requeue gated on its sync. Uses machinery that already half-fires; must be wired read-site by read-site; cold handlers run twice by design.
- **C:** dissolves the sqlite case specifically by making the op link-based — no gating, no new invariant, but per-API work and no help for the next synchronous asCell read.
- A and B compose (B as backstop for what A's enumeration misses, e.g. reads through lazy proxies). The `pull()` covered-in-flight fix underlies both.

Happy to prototype B or C on the same testbed if that's the preferred direction — the lunch-poll setup reproduces the race 100% deterministically either way.

mathpirate:
Addressed the cubic P1 in de6e2078d — it was a valid catch: the `#syncTasks` dedupe path handed concurrent same-key callers the raw fetch promise, so in a mixed batch they could resolve before the covered-in-flight docs landed (the exact early-resolve this PR exists to prevent). The task now stores the combined wait; the selector tracker and the error-path cleanup stay keyed on the raw fetch promise, since future covered pulls only need this batch's fetch, and a covered watch's failure shouldn't invalidate selectors whose own fetch succeeded. Runner suite 553/553.

ubik2:
We got tests for the presync implementation here, but there's no test of the situation that motivated this PR.
That makes it easy to introduce a regression.
It's not clear to me from the description what is actually going on.
Given the symptom, I would have guessed the underlying cause is actually a schema issue.
The two added tests both essentially say you're only interested in the one doc, and don't want any linked docs, so it's logical you wouldn't get them there.
I should probably have an agent investigate the validateAndTransform carrying synced flags improperly to see if this is a real issue. If it is, it's going to be a problem in other places too.

PR #3970, opened 2026-06-10
fix(cfc): Wave 0 — close enforcement bypasses & fail-open holes

seefeldb:
## Summary

Wave 0 of the CFC spec-audit remediation: close the **enforcement bypasses** and **fail-open holes** in `packages/runner` that let untrusted code skip or weaken information-flow checks, or that silently dropped/ignored security-relevant claims. These are the highest-priority items from a spec audit of `packages/runner` against `~/src/specs/cfc` — a gate that can be switched off isn't a gate, so these land first.

Each fix is red-green (failing test first) and scoped to a single concern. Audit IDs (Sn) reference the audit's soundness findings.

## Fixes

1. **`prepareCfc` always verifies (S2)** — removed the `input` parameter that let a caller skip `prepareBoundaryCommit` while still passing the commit-time digest recheck. Bypass closed at the type level.
2. **Enforcement mode can't be weakened (S3)** — `setCfcEnforcementMode` may be raised but never lowered below the highest enforcing level set on a transaction, so code holding a `Cell` can't disable enforcement mid-transaction. Tests that exercised weaker modes now construct the runtime at that mode.
3. **Unattributed writes fail closed (S13)** — `writeAuthorizedBy` no longer falls back to the transaction's *current* identity for a write recorded before any identity was set; an unattributed write can't borrow a trusted identity established later in the same tx.
4. **Wildcard policy applies to unresolvable links (S17)** — when a written link's target is unresolvable, applicability no longer trusts the link's author-embedded schema to exclude the policy; it fails closed (the entry applies).
5. **Unsupported `ifc.*` claims fail closed (S10)** — `opaque`, `passThrough`, `recomposeProjections`, `combinedFrom`, `combinationType`, `transformation`, `addedIntegrity` join `projection`/`collection` as rejected-on-write, instead of being silently ignored and dropped by schema-merge.
6. **Flow-precision claims conditioned on op argument usage** — the element-local claims (`PointwiseWriteDependency`, `ElementLocalExpansion`) are dropped when a map/filter/flatMap op reads the whole `array` or data-bearing `params` (a cross-key dependency); structural claims stay. Closes a latent unsoundness before any consumer relies on them.
7. **Empty `maxConfidentiality` = public-only (not no-ceiling)** — a declared but empty ceiling now denies confidential atoms; only `undefined` means no ceiling. Both the observation path and the prepare-time check.
8. **`createRef` fails closed on underivable ids (S14)** — throws instead of substituting a random UUID when a derivation input is an OpaqueRef with no value or a Cell with no entityId. The documented no-cause "mint fresh entity" path is preserved.

## Testing

- New focused tests for each fix (unit + boundary), red-green verified.
- Full `packages/runner` unit suite: green.
- `deno task integration runner` (incl. the `sqlite-cfc-label*` CFC integration tests): 12 passed, 0 failed.

## Scope

Deliberately **not** in this wave: the generic policy calculus, CNF labels, the intent lifecycle, trust lattice, the typed `TrustedDerivedId` envelope, default label transition + PC propagation, and other next-phase items. Waves 1–3 (verifier-input integrity, verifier semantics, events/egress) build on this branch. Server-side ACL (Track A) is deferred.

---
## Summary by cubic
Closes CFC enforcement bypasses and fail-open edges in `packages/runner` so untrusted code can’t skip checks and unsupported claims fail closed. Tightens APIs and defaults; also enforces empty ceilings on the LLM tool-call path.

- **Bug Fixes**
  - `prepareCfc()` always verifies; removed the input parameter that allowed skipping verification (S2).
  - Enforcement mode can be raised but not lowered below the highest enforcing level set on a transaction (S3).
  - Unattributed writes no longer borrow the transaction’s current identity; they fail closed (S13).
  - Wildcard policy entries apply to unresolvable links (fail-closed) instead of trusting author-embedded link schemas (S17).
  - Reject unsupported `ifc.*` claims on write: `opaque`, `passThrough`, `recomposeProjections`, `combinedFrom`, `combinationType`, `transformation`, `addedIntegrity` (S10).
  - Flow-precision claims now depend on op argument usage; drop element-local claims when the callback reads `array` or data-bearing `params`. Updated `map`/`filter`/`flatMap`.
  - Empty `maxConfidentiality` means public-only; `undefined` means no ceiling. Applies to observation, prepare-time checks, and the LLM tool-call gate.
  - `createRef` throws on underivable inputs (OpaqueRef without value, Cell without entityId); the no-cause “mint fresh id” path is unchanged.

- **Migration**
  - Remove any argument to `prepareCfc()`; it no longer accepts input.
  - Don’t downgrade enforcement mid-transaction. Set the desired `cfcEnforcementMode` when creating `Runtime`.
  - If you relied on empty ceilings allowing confidential data, use `undefined` for “no ceiling” or list allowed atoms explicitly.
  - If you relied on `createRef` falling back to random IDs, supply real inputs or avoid cause-based derivation; it now throws instead.

<sup>Written for commit c762d3a565ebf7c7cd91ca12a528451be94a6901. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3970?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

NEW_PERF_BASELINE: subtest: pattern-unit-1/pattern-unit-tests > packages/patterns/cfc-authorized-save/main.test.tsx = 13s

> **Perf note:** the `cfc-authorized-save` pattern shows ~+20% (median unchanged at 10.0s; the comparison metric caught a 12.0s run). It is the only regressed metric — every other pattern is flat — which points to the small per-write overhead the W0 CFC verification checks add on the most CFC-intensive pattern (authorized saves), not generic runner drift. Accepting the baseline as an expected cost of the security hardening; revert this line if you'd rather investigate.

### DIFF
--- a/cfc-spec-audit.md
+++ b/cfc-spec-audit.md
@@ -1,0 +1,545 @@
+# CFC Spec Audit — packages/runner vs ~/src/specs/cfc
+
+Date: 2026-06-09. Spec bar: sections marked "Status: Normative"; ch. 14 (open problems),
+08-14 and 08-16 (redirect stubs to ch. 14) treated as non-normative. Produced by seven
+parallel audit passes (one per spec slice plus one code-first sweep); the highest-impact
+soundness claims were independently re-verified against source.
+
+## How the implementation relates to the spec
+
+The runner implements a deliberate **phase-1 subset**: flat conjunctive labels (atom
+arrays, no CNF clauses), schema-declared `ifc` annotations verified at **commit time**
+(`prepareBoundaryCommit`, `packages/runner/src/cfc/prepare.ts:2077`) with a prepared-digest
+recheck at commit (`extended-storage-transaction.ts:656-673`), persisted per-path label
+maps bound to content-addressed schema hashes, link/reference label pass-through, write
+authority, UI-contract trusted events, sink-request replay fidelity, and InjectionSafe
+schema sanitization. Default mode is `enforce-explicit` (`runtime.ts:391`); the bare
+transaction default is `disabled` (`cfc/types.ts:31`).
+
+Terminology: spec `Label{confidentiality: Clause[], integrity: Atom[]}` ↔ impl
+`IFCLabel{confidentiality?: unknown[], integrity?: unknown[]}` (`label-view-core.ts:4`);
+spec boundary loop ↔ impl prepare/digest two-phase; spec `CodeHash` write authority ↔ impl
+builtin-id strings or `__ctWriterIdentityOf{bundleId,file,path}` bindings; spec
+`$actingUser` ↔ impl `{__ctCurrentPrincipal: true}` placeholders.
+
+The single largest architectural distance: **the spec derives output labels from input
+labels (default transition §8.9.3 + PC propagation §8.9.2/§8.11); the implementation never
+does** — it persists only schema-declared labels and verifies declared constraints. Every
+"value-copy laundering" finding below is downstream of that one posture.
+
+---
+
+## 1a. Gaps in existing features (implemented, but incomplete or divergent)
+
+**Input requirement checks (`requiredIntegrity` / `maxConfidentiality`), §8.10.3:**
+- Consumed set is **transaction-global**, not scoped to the annotated input path
+  (`prepare.ts:1694-1709`) — spec uses `collectConsumedLabelsAtOrBelow(path)`. Over-strict
+  in one direction (unrelated labeled read can fail an unrelated write), under-strict in
+  the other (see soundness S7).
+- Satisfaction is whole-atom `deepEqual` membership (`prepare.ts:1747-1759`), not the
+  spec's pattern matching with **shared-witness-key coherence** — two leaves matching via
+  different witnesses both pass where the spec requires one shared witness.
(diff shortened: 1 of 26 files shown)


seefeldb:
Addressed the review findings:
- **Codex P1** (empty tool ceiling): `toolAllowsObservedConfidentiality` now delegates to `cfcObservationFitsCeiling` for any declared array ceiling, so an empty `maxConfidentiality` is enforced as public-only on the LLM tool-call path too.
- **cubic P1** (createRef doc): JSDoc clarified — derivation inputs must resolve (fail closed); a missing cause still mints a random id.
- **cubic P2** (test await): the two sync `cfc-tx-control-guard` tests now `await` their cleanup.

All in `35daccbb7`. Full runner suite + `integration runner` green.

PR #3973, opened 2026-06-10
fix(cfc): Wave 2 — verifier semantics

seefeldb:
## Summary

Wave 2 of the CFC spec-audit remediation (stacked on [#3972](https://github.com/commontoolsinc/labs/pull/3972)). Tightens verifier *semantics* — over- and under-approximations in label derivation and claim verification.

## Fixes

1. **exactCopyOf applicability gating + reject array wildcards (W2.15)** — `verifyExactCopyRequirements` ran for every claim regardless of whether the target was written, so an untouched entry compared `undefined`/`undefined` and passed vacuously (accepting the claim and copying its label unverified); a wildcard claim could never match a concrete write. Now gated on `ifcEntryAppliesToAttemptedWrite`, and array-wildcard exactCopyOf fails closed.

2. **Aggregate columns carry no integrity (CT-1668)** — a null-origin sqlite result column (`COUNT(*)`, expression, literal) inherited the union of every labeled column's integrity, so it could falsely claim an atom held by a single column (§8.17.1: meet, never union). It now keeps the confidentiality union (sound over-approximation) but drops integrity.

3. **labelMap store confidentiality is grow-only on re-write (S9)** — a path can carry confidentiality beyond its schema (link-derived / carried-view atoms); re-deriving from the schema alone replaced the stored entry and dropped it (§8.12.1: store confidentiality is grow-only). The re-write now unions the prior stored confidentiality.

4. **joinSchema descends anyOf/oneOf/allOf + prefixItems (W2.16)** — the confidentiality-tainting LUB skipped combinator branches and tuple elements, silently dropping branch-local confidentiality (under-tainting fail-open) and tainting differently than the verifier's `walkIfcSchema`. Now unions every branch.

## Testing

- New focused tests per fix (unit + boundary), red-green verified (incl. revert-checks for the labelMap and joinSchema fixes).
- Full `packages/runner` unit suite: 568 passed, 0 failed.
- `deno task integration runner`: 12 passed, 0 failed.

## Deferred from this wave (flagged for review)

- **S7 vacuous-pass + ancestor reads** — the audit's explicitly *observe-first* item. `requiredIntegrity` is used by 0 real patterns, but the fix can't cleanly distinguish handler-input reads from runtime-plumbing reads (both unlabeled — failing plumbing reads would fail everything), and the ancestor-read half shifts `maxConfidentiality` for the 3 patterns that use it. Needs an observe-mode rollout with telemetry + read-provenance design rather than a blind autonomous landing.
- Smaller residual: `coalesceLabelEntries` still unions integrity for two *distinct* writes at one path in a single transaction (load-bearing for legitimate link/endorsement refresh; needs combination-vs-refresh provenance).

---
## Summary by cubic
Wave 2 CFC verifier semantics hardening. Prevents vacuous exactCopyOf passes, drops over-claimed integrity on SQLite aggregates, preserves stored and ancestor confidentiality on rewrite, and unions confidentiality across schema combinators.

- **Bug Fixes**
  - Gate `exactCopyOf` checks to paths actually written; reject array wildcards to avoid vacuous passes (W2.15).
  - For null-origin SQLite result columns (aggregates/expressions), keep confidentiality union and drop integrity (CT-1668).
  - Make `labelMap` confidentiality grow-only on rewrite by unioning prior atoms from the path and its ancestors; integrity stays freshly derived (S9).
  - In `joinSchema`, descend `anyOf`/`oneOf`/`allOf` and `prefixItems`, unioning confidentiality across all branches (W2.16).

<sup>Written for commit 1c5073f697a1ea88c02fb378ec44f3906f27b99a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3973?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/builtins/sqlite-builtins.ts
+++ b/packages/runner/src/builtins/sqlite-builtins.ts
@@ -172,8 +172,14 @@ function deriveNullOriginIfc(tables: LabelTables): IFCLabel | undefined {
       if (ifc && typeof ifc === "object") merged = mergeLabel(merged, ifc);
     }
   }
-  return (merged.confidentiality?.length || merged.integrity?.length)
-    ? merged
+  // Confidentiality unions across contributors (a sound over-approximation: the
+  // aggregate could depend on any column). Integrity does NOT: an aggregate /
+  // expression / literal is a new computed value and inherits no integrity
+  // evidence. Unioning integrity would let it falsely claim an atom held by a
+  // single column (§8.17.1: class-aware meet, never union; propagation classes
+  // pending, so conservatively empty). [CT-1668]
+  return merged.confidentiality?.length
+    ? { confidentiality: merged.confidentiality }
     : undefined;
 }
 
--- a/packages/runner/src/cfc.ts
+++ b/packages/runner/src/cfc.ts
@@ -109,6 +109,42 @@ export class ContextualFlowControl {
     if (schema.ifc) {
       ContextualFlowControl.addIfcAtoms(joined, schema.ifc.confidentiality);
     }
+    // A value validates against one (anyOf/oneOf) or all (allOf) branches, so the
+    // confidentiality LUB must union every branch's atoms; otherwise branch-local
+    // confidentiality is silently dropped (under-tainting fail-open, audit 1.6).
+    for (const key of ["anyOf", "oneOf", "allOf"] as const) {
+      const branches = (schema as Record<string, unknown>)[key];
+      if (Array.isArray(branches)) {
+        for (const branch of branches) {
+          if (branch !== undefined && typeof branch === "object") {
+            ContextualFlowControl.joinSchema(
+              joined,
+              branch as JSONSchema,
+              ContextualFlowControl.childFullSchema(
+                branch as JSONSchema,
+                fullSchema,
+              ),
+              cycleTracker,
+            );
+          }
+        }
+      }
+    }
+    // Tuple element schemas carry their own confidentiality too.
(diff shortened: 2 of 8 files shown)


seefeldb:
Addressed the review finding:
- **Codex/cubic P1** (grow-only labelMap): the re-write merge used an exact-path lookup and missed confidentiality stored on an **ancestor** path (which a child entry would shadow under longest-prefix reads). Now merges prior confidentiality from this path and every ancestor of it (`isPrefix`). Added an ancestor-case regression test.

Full runner suite + `integration runner` green.

seefeldb:
**Perf note:** Performance Check is UNSTABLE (non-blocking) on two **CI-job wall-clock** metrics — `CLI Integration Tests (core-piece-values)` and `Package Integration Tests (runtime-client)`. Both have median = baseline (1m9s / 1m14s) with high variance (stddev ~6–8s); the flagged 1m43s/1m36s are runner outliers. These are integration-job orchestration timings unrelated to this change (CFC fixes live in the runner unit layer and can't plausibly move those jobs ±30–50%). Not accepting a NEW_PERF_BASELINE — that would mask real future regressions in those jobs for transient runner noise. Merging on the non-blocking status; the full runner unit suite (572) and `integration runner` are green.

PR #3974, opened 2026-06-10
feat: CFC Phase 3 — per-row, data-derived SQLite labels (3.a)

seefeldb:
## Summary

Implements **CFC Phase 3 (3.a)** per the merged design ([#3966](https://github.com/commontoolsinc/labs/pull/3966), `docs/specs/sqlite-builtin/plans/cfc-phase3-per-row.md`): per-row, **data-derived** labels for the SQLite builtin — an email row becomes confidential to its sender, its recipients (regex-extracted from a dirty `Name <addr>, addr` line), and the db owner, with authored-by integrity gated on the row's own `dmarc` evidence. Labels are computed at the write gate and re-derived at read; every unresolvable case **fails closed**.

Built red-green in 5 commits (failing test first each increment), e2e-verified against the real toolshed server.

## What's in

- **Rule surface + shared evaluator** (`packages/memory/v2/sqlite/row-label.ts`): `table(columns, (f) => ({confidentiality, integrity}))` with the closed helper set — `match` (forced-global regex ⟹ split), `principal(protocol,…)` (protocol-implied normalization; `did:key` case preserved), `all`/`any`/`intersect`, `whenMatches`, `dbOwner()`, `endorsedBy`/`authoredBy` (minted as self-describing `claimed-*` atoms — content-derived provenance never lowers to the trusted family), `constant`. The returned object **is** the serialized plain-JSON AST (`rowLabel` on the table schema); `evaluateRowLabel` interprets it identically everywhere. `any()` (authored OR-clause) **errors at `table()` time** per CFC spec §18.5.3 rule 3, and a smuggled `anyOf` fails closed at validation *and* evaluation — never silently flattened.
- **Read side** (`row-label-read.ts` + the `sqliteQuery` flush): rule inputs located by **TRUE column origin** (alias-proof, spoof-proof; missing/ambiguous ⟹ refuse); per-row labels attached by writing each split row entity doc directly under a root-`ifc` schema (the spike-3.0 mechanism — coexists with Phase 2's per-column field labels); aggregates/`COUNT(*)` refuse on rule-bearing dbs; declared **output ceiling** (`MaxConfidentiality` Row schema or `maxConfidentiality` query option, placeholders `__ctCurrentPrincipal`/`__ctDbOwner`) with **`onExceed: "fail" (default) | "skip"`**. Server provenance gate extended: rowLabel-only tables get column origins too (`dbNeedsColumnProvenance`, shared predicate).
- **Write side** (`row-label-write.ts` + `cell.ts .exec()`): attributable INSERTs evaluate the rule over bound values ⟹ prospective row label, enforce **no-laundering** (every labeled input captured by the row label), and record per-row labels as the write's **sink-request CFC policy input** before commit. UPDATE of rule-input columns, INSERT…SELECT/upserts, named params, labeled values outside an evaluable INSERT: fail closed (server-side commit evaluation, 3.c, is the planned lift).
- **`cfSqlite` namespace** for patterns (api type + builder factory) and a new transformer callback boundary (`sqlite-row-label-rule`) so `table(columns, rule)` is authorable inside a pattern body.
- **e2e** (`integration/sqlite-cfc-row-label.test.{ts,tsx}`): two seeded emails come back with distinct per-row labels (`{alice,bob,owner}` vs `{carol,dave,erin,owner}`), integrity minted only for the `dmarc=pass` row, `COUNT(*)` refused, ceiling+skip returns exactly the fitting row.

## Decisions made while you were AFK (full log)

1. **#3966 merged before the timer fired ⟹ this PR bases on `main`** (the stack collapsed naturally); the demo PR will stack on this branch.
2. **`when(matches(f,re), t)` fused into `whenMatches(f, re, t)`** — the builder's control-flow `when()` transformer lowering matches bare *names* and mangled the call (injected schema args). Fusing removes the collision class and a concept; `matches()` dropped. The design doc/CFC-spec examples still show `when(matches(…))` — reconcile when this lands.
3. **Authoring shape**: cfSqlite destructuring + `table(rule)` live **inside the pattern body** (SES module scope allows only trusted builder calls; top-level destructuring rejected). New transformer boundary recognizes the rule callback via the `SqliteTableFunction` type alias; deliberately exempt from SES self-containment (the rule runs eagerly at build, it is never extracted).
4. **`db.owner`** = `trustSnapshotProvider().actingPrincipal` at `sqliteDatabase` handle creation (default provider: the storage signer) — a fixed property of the db ref, never the acting reader.
5. **Split-doc materialization fix** (also a latent Phase 2 degradation): rule term lists (arrays of objects) split into per-element entity docs on the handle write, so `getRaw`/the wire saw doc links. `.exec()` now materializes `tables` via a resolving permissive-schema read; `readDbRef` clones tables to plain JSON for the wire and all consumers.
6. **Helper named `constant()`** (design's `const()` is a JS reserved word); serialized spec omits the design §5 `"table"` field (consumers derive it from the `db.tables` key — removes a name-mismatch failure mode).
7. **Strict-if-present matching** (from the design conversation): empty/NULL field contributes nothing; non-empty-but-zero-matches refuses; `min:1` makes `from` a required anchor. Zero matches in an integrity position mints no claim; >1 fails closed (display-name bait).
8. **Phase 3 query options join the request hash** ⟹ pre-existing labeled queries re-hash once after deploy (benign re-issue).
9. **No `combineLabels` seam introduced**: per-row (root) and per-column (field) labels live at *different paths* on the row doc, so no cross-merge call site exists; CT-1668's union-vs-meet question stays in the runtime's `mergeLabel` accumulation, untouched here.
10. **Multiple rule-bearing tables in one query refuse** (cross-rule joins deferred); aggregates on rule-less tables keep Phase 2 behavior.

## Verification

- New: memory 31 rule tests; runner 20 read steps + 15 write-gate steps; e2e PASSED (real toolshed).
- Regressions: full memory suite 197/197; ts-transformers 292/292 (617 steps); Phase 2 sqlite e2e 3/3; runner sqlite unit suites 70 steps; **full repo `deno task check` clean**.

---
## Summary by cubic
Adds CFC Phase 3 per‑row, data‑derived labels to the SQLite builtin. Each row computes its own confidentiality and integrity at write and read, with ceilings and strict fail‑closed behavior.

- **New Features**
  - `table(columns, rule)` adds per‑row rules via helpers: `match`, `principal`, `all`, `intersect`, `whenMatches`, `dbOwner`, `authoredBy`/`endorsedBy`, `constant` (`any` is rejected per spec).
  - Read: inputs resolved by true (table,column) origin; per‑row label written under root `ifc`; aggregates, null‑origin columns on rule‑bearing DBs, and cross‑rule joins refuse. New query options: `maxConfidentiality` and `onExceed: "fail" | "skip"` with placeholders `__ctCurrentPrincipal`, `__ctDbOwner`.
  - Write: `db.exec` gate evaluates attributable `INSERT`s, enforces capture to prevent laundering (including empty‑label cases), attributes `UPDATE SET` columns even without params, and records per‑row labels as sink‑request policy input. `INSERT…SELECT`, upserts, named params, and `UPDATE` of rule‑input columns fail closed.
  - Authoring/runtime: `cfSqlite` helper namespace; `sqlite-row-label-rule` transformer boundary; DB owner captured at `sqliteDatabase` creation on the db ref; provenance gate triggers on per‑row‑rule DBs; validator hardened (regex lint + forced global flags; stricter spec checks; transformer rejects spoofed aliases; `readDbRef` validates `scope`).

- **Migration**
  - Author rules inside pattern bodies via `cfSqlite`: `const { table, match, principal, … } = cfSqlite`.
  - Declare ceilings with `maxConfidentiality`/`onExceed` as needed; placeholders resolve at query time.
  - Avoid aggregates and queries that touch multiple rule‑bearing tables; they now refuse.
  - For rule‑bearing tables, use positional params in `INSERT`; avoid `INSERT…SELECT`/upserts and any `UPDATE` of rule‑input columns (including literal SETs) for now. Existing labeled queries will re‑hash once due to new options.

<sup>Written for commit 16b1a12cbc1be3877c359149095ddc5caa01fa2d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3974?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Perf baseline acceptance (environmental noise, 3 samples)

The Performance Check flagged wall-clock timings unrelated to this diff: attempt 1 showed a uniform +21–44% across nearly all jobs (including packages this PR doesn't touch, e.g. runtime-client); attempts 2–3 reduced to the two metrics below, 1–6% over their adaptive thresholds, on tests that exercise none of the new code paths (no `table()` rules; the new transformer boundary check exits on callee name). Baseline window stddev ≈ 20s; the immediately preceding main PR (#3984) shows the same elevation in the baseline samples. Accepting per the perf tool's mechanism:

NEW_PERF_BASELINE: test: pattern-unit-1/pattern-unit-tests = 168s
NEW_PERF_BASELINE: subtest: pattern-unit-1/pattern-unit-tests > packages/patterns/cfc-authorized-save/main.test.tsx = 13s


seefeldb:
Review round addressed in 16b1a12cb (rebased onto main first — the new integration test migrated to the #3947 `memoryHost` option):

- **P1 empty-label laundering** (Codex + cubic): fixed — an empty computed row label now captures nothing; labeled input + empty label fails closed. Red-green tested.
- **P1 param-less literal UPDATE bypass** (cubic): fixed — SET columns are attributed from the SQL itself via a shared `parseUpdateSetColumns` (extracted from write-targets); unattributable SET clauses on rule-bearing tables fail closed, naming the touched rule input when detectable. Red-green tested.
- **P2s**: `evalMatch` forces the global flag (no uncaught matchAll throw on hostile wire flags); the regex lint returns an unbalanced-parenthesis reason instead of crashing; `isSqliteTableCallee` now also requires the alias to be declared by Common Fabric's own typings; `readDbRef` validates `scope` with `isCellScope`; both stale JSDocs corrected.
- **On the time-based polling loop** (cf-review P2): deliberately kept — it is the established idiom of the Phase 2 integration templates this test mirrors (`sqlite-cfc-label.test.ts`), bounded by a deadline with a full state dump on timeout; switching this one test to a different settlement mechanism would diverge from its siblings. Happy to migrate the family together in a follow-up if desired.

PR #3975, opened 2026-06-10
fix(cfc): Wave 3 — egress relabeling

seefeldb:
## Summary

Wave 3 of the CFC spec-audit remediation (stacked on [#3973](https://github.com/commontoolsinc/labs/pull/3973)). Closes an egress/relabeling fail-open and adds observability for sink-release rejects.

## Fixes

1. **Refuse paramless intra-db relabeling on a labeled db (S6)** — `checkSqliteWriteCeiling` skipped entirely when `params === undefined`, so a paramless column-to-column flow (`INSERT…SELECT`, `UPDATE col = col`) copied a labeled column's data into a column whose declared label is weaker — the data re-emerged under the destination column's (weaker) read-label, laundering its confidentiality. On a labeled db (any column declares confidentiality or `maxConfidentiality`), a paramless write that moves column-derived data into a stored position now fails closed. Literal-only INSERT/UPDATE, DELETE, and every write to an unlabeled db are unaffected (sound over-approximation: any non-literal RHS counts as a column reference).

2. **Surface sink-request release rejects (W3.23)** — a post-commit sink-request release that fails verification is fail-closed (the effect is skipped) but was only `console.warn`'d. It now records a CFC diagnostic, fires an `onSinkReleaseReject` instrumentation hook, and increments a `sinkReleaseRejects` runtime stat. The effect is still not sent.

## Testing

- New unit tests per fix, red-green verified (`checkSqliteWriteCeiling` exploit/allow cases; sink-release reject surfacing + skip).
- Full `packages/runner` unit suite: 570 passed, 0 failed.
- `deno task integration runner` (incl. `sqlite-cfc-label*`): 12 passed, 0 failed.

## Deferred from this wave (flagged for review)

The remaining Wave 3 items are the largest/riskiest in the audit's plan and need design review rather than an autonomous landing:

- **S12 trusted-event hardening** — nonce-unique event ids, consume-on-use (session-scoped single use), verifying provenance against the runtime-serialized envelope rather than candidate-embedded data, and stopping WeakSet trust-mark propagation through handler re-dispatch.
- **Label-aware sink ceiling** — `sink-inventory.ts` currently declares no per-sink ceilings; activating it needs adding ceilings plus request-confidentiality tracing.
- **LLM observation fail-open** — low-exploitability (only rare genuine metadata read errors; `NotFound` is already `undefined`); the swallow sites are shared between display (treats `undefined` as *blocked* = fail-closed) and observation (treats it as *public* = fail-open), so the fix needs cross-package renderer validation.

### DIFF
--- a/packages/runner/src/builtins/sqlite/write-ceiling.ts
+++ b/packages/runner/src/builtins/sqlite/write-ceiling.ts
@@ -17,11 +17,85 @@ import {
 
 interface ColumnIfc {
   maxConfidentiality?: readonly unknown[];
+  confidentiality?: readonly unknown[];
 }
 type Tables = Record<
   string,
   { properties?: Record<string, { ifc?: ColumnIfc }> } | undefined
 >;
+
+/** Any column anywhere declares a read-label or a write-ceiling. */
+function dbDeclaresAnyLabel(tables: Tables): boolean {
+  return Object.values(tables).some((t) =>
+    Object.values(t?.properties ?? {}).some((c) =>
+      (c?.ifc?.maxConfidentiality !== undefined) ||
+      ((c?.ifc?.confidentiality?.length ?? 0) > 0)
+    )
+  );
+}
+
+/** A bound-paramless RHS token that stores no column-derived data. */
+function rhsIsLiteral(rhs: string): boolean {
+  const t = rhs.replace(/;\s*$/, "").trim();
+  return t === "" ||
+    /^''$/.test(t) || // blanked string literal
+    /^[-+]?\d+(\.\d+)?$/.test(t) || // numeric literal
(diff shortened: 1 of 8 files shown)


seefeldb:
Addressed the review findings (all in `39307a977`):
- **Codex/cubic P1** (upsert): `ON CONFLICT … DO UPDATE SET col = col` now treated as a relabel risk on the paramless path.
- **Codex/cubic P1** (empty array): an empty bound-param array `[]` is now covered by the paramless relabel check (same no-bound-value flow as `undefined`).
- **cubic P2** (numerals): broadened the numeric-literal regex so `1e3`, `0xFF`, `.5` are recognized and literal-only paramless UPDATEs aren't spuriously rejected.

Added test cases for all three. Full runner suite + `integration runner` green.

PR #3983, opened 2026-06-10
feat(runtime-client): page-layer multi-space — one worker serves many spaces

jkomoros:
## What

One worker now serves **page operations for many spaces over one connection**. Follow-up to #3947 (per-space storage hosts): the storage layer was already multi-space; this makes the page/piece layer match, which is the last piece needed for the smallest visible cross-host render (one worker, two hosts, a foreign read-only pattern on screen).

Design: a **thin per-space facade**, not multi-space-internal handlers. `RuntimeProcessor` holds `spaces = Map<DID, {PieceManager, PiecesController}>`, seeded with the home context at initialize. `getSpaceCtx(space?)` resolves a page operation's context — no space (or the home space) is the existing context, any other space lazily gets its own `PieceManager`/`PiecesController` pair sharing the one Runtime/scheduler/StorageManager. Handlers stay single-space internally.

- **Protocol**: optional `space?: DID` on the page-op requests (`PageGet`/`GetSlug`/`Remove`/`Start`/`Stop`/`GetAll`/`Synced`, `GetSpaceRootPattern`/`Recreate`). Absent ⇒ home, byte-identical behavior. `PageCreate` deliberately not extended — creation stays home-only.
- **RuntimeClient**: optional trailing `space` on the corresponding methods.
- **lib-shell**: `RuntimeInternals.getPattern(id, {start?, space?})`; the pattern cache key is now always space-qualified (`${space ?? bound}:${id}`), so the no-space form and an explicit bound-space form share one entry. `invalidatePattern`/`refreshPattern` take the same optional space.
- **Per-space sessions authenticate as the user** (`{as: identity, space}`) — no per-space signer, matching the storage connections from #3947.

## Semantics owned explicitly

- **A failed space doesn't poison the worker.** The lazily-built context's space-cell sync can reject (unreachable host, bad space). The failure is observed and the context evicted, so the next request rebuilds it instead of failing forever; the rejection no longer escapes unhandled.
- **Root-pattern ensure/recreate is guarded home-space-only.** Both write (create + start the default pattern), and they seed the pattern URL from the requesting user's home `defaultAppUrl` — right for your own space, wrong for a foreign one. The mount path (`PageGet`) doesn't need them. Lift when federation defines who provisions a space's root pattern.
- **`PageHandle.start`/`stop` now carry their cell's space.** For home cells this resolves to the identical context (no change). For a cross-space handle (reachable pre-PR2 via slug redirects with `redirect.space`), the old behavior addressed the entity id in the *home* space — the wrong document; it now routes to the right space.
- **Created pieces register in their own space's list.** `pieceCreatedCallback` was pinned to the home `PieceManager` (pre-existing — reachable via redirect targets started in other spaces), so a foreign pattern's created piece polluted the home list with a cross-space link. It now routes to the piece's space context when the worker has one — it always does, since only an existing context can have started the creating pattern — with the home manager as fallback.
- **Writes to foreign spaces** (e.g. `PageRemove` with `space`) ride the same trust model as before: the connection authenticates as the user and the server's ACLs are the guard — `CellSet` already accepted a `CellRef` carrying any space.

## Tests

- Worker: `getSpaceCtx` resolves no-space/home to the initialize-time context; a foreign space gets a distinct, cached context; `handlePageGet({space: B})` returns a page whose `cell.space === B` (against a real Runtime over loopback storage); the foreign root-pattern guard rejects.
- Client: `PageHandle.start/stop` carry the cell's space over IPC.
- lib-shell: `getPattern` passes `space` through, caches per (space, id), aliases explicit-bound-space to the no-space entry, invalidates per space.
- Full `deno task check`, lint, fmt green; runner (555/0), piece, shell, runtime-client, lib-shell suites green.


jkomoros:
Fixed in the latest commit (identified by cubic, converging with the pre-push adversarial review): created-piece registration now routes to the piece's own space context via `managerFor` — a foreign pattern's pieces land in *its* space's list, home behavior unchanged. The PR-body limitation note no longer applies.

seefeldb (approved):
LGTM, but this could be simpler when we just remove the implicit spaces and have all API calls that have `space?` to a required space.

jkomoros:
Agreed — the optional `space?` is transition scaffolding, not the end state. Once the shell/embedder call sites all pass an explicit space (which the loom-side PatternHost work forces anyway), a follow-up can flip the protocol's page-op `space` to required and delete the home-space defaulting in `getSpaceCtx`. I'll fold that into the lifecycle follow-up (where `RootView`'s recreate-on-space-change goes away too) so the requirement lands together with the last implicit-space consumer.

PR #3985, opened 2026-06-10
perf(runner): cut traverse tail latency ~40% — prepared anyOf, gated diagnostics, memoized data URIs

seefeldb:
## What

Third round of the traverse optimization loop, targeting **tail latency**: traverse no longer dominates mean time, but it shows up in tails. New per-invocation latency instrumentation in the replay harness diagnosed a 1000× tail (p50 ≈ 5µs, p99 5ms+, max 25ms), with an identical shape on client and server fixtures: **anyOf-heavy vnode docs evaluating up to 8,424 branches in a single traversal**.

| fixture | p99 | max | mean |
|---|---|---|---|
| notebook-test | 5.17 → **3.70ms** (−28%) | 25.2 → **18.6ms** (−26%) | −18% |
| toolshed-reload | unchanged (µs-level) | 7.9 → **6.3ms** (−20%) | — |
| shopping-list-test | unchanged (no anyOf tail) | | |
| folksonomy-aggregator (live, dynamic schemas) | — | — | **main parity** (5-run A/B) |

## Fixes (replay oracle byte-identical; anyOf counters preserved exactly)

1. **`prepareAnyOf`** — the per-node anyOf loop re-did per *visit* what depends only on the schema: sorting branches (two `hasAsCell` filter passes per node), merging rest+option per branch, and `canBranchMatch`'s static derivations ($ref resolution, type-list normalization, required-props applicability). Precomputed once per **interned** schema identity; non-interned schemas take the legacy inline loop verbatim (interned identity guarantees reuse — frozen-but-fresh per-evaluation schemas would rebuild per node and regress).
2. **Gated per-visit diagnostics** — `traverseWithSchema` built a `docId + "/" + path.join("/")` string and touched a Map+Set on every schema visit, purely to feed the slow-traverse log. Now behind `CF_TRAVERSE_DIAGNOSTICS=1` (the log says so when disabled).

## A third fix was tried and **reverted** — the CI perf check earned its keep

Memoizing `createDataCellURI` for inline array elements won ~20% more on replay fixtures, but the CI perf check flagged `folksonomy-aggregator.test.tsx` +37% — and unlike two earlier shard-level flags (noise), this one **reproduced locally** (1180→1490ms, 5-run A/B). Live dynamic patterns mint *fresh* deep-frozen values per evaluation: the memo never hits, so its `isDeepFrozen` walks + key hashing were pure added cost. Lesson journaled: identity memos need identity *reuse*, not just identity stability. The upstream fix (query traversal shouldn't synthesize data URIs at all, per the code's own TODO) is documented for a future round.

## Harness additions

- `replayFixture(..., { collectLatency: true })` → p50/p90/p99/p99.9/max + the slowest invocations with per-invocation counters; bench prints it under `BENCH_DIAGNOSTICS=1` for ongoing tail tracking.
- `profile-driver.ts`/`profile-target.ts` accept an invocation index to CPU-profile a single tail invocation in isolation.

## Test plan

- [x] Replay oracle green against unchanged goldens (4 fixtures)
- [x] anyOf counter accounting byte-identical
- [x] Full runner suite (575) green
- [x] folksonomy-aggregator + notebook pattern tests at main parity locally (5-run A/B)
- [x] `deno fmt`/`lint`/`check` clean


seefeldb:
Merging with the Performance Check red after 3 attempts, per the perf-gate-is-advisory convention. Diligence record:

- **Attempt 1 flag** (`cfc-authorized-save` +22.8%): 5-run local A/B vs main — parity at test-body (~45ms both) and wall level (0.33–0.36s both).
- **Attempt 2 flags** (whole-repo type-check +224.8%, `Check` job +160.9%): physically unrelated to a 100-line runtime diff; congested runner.
- **Attempt 3 flags** (`Runner Tests (4/4)` at 102% of margin, `lot-watch` +20.9%): lot-watch 5-run A/B — parity (42–46ms vs 41–43ms; the CI metric is harness wall time).
- The one flag that **was** real (folksonomy-aggregator +37%, attempt 1 of the previous run) reproduced locally at +26%, was root-caused (identity memo on fresh-per-evaluation values), and the offending optimization was reverted in d63037cd — folksonomy is back to main parity (5-run A/B).

The perf gate earned its keep this round; today's residual flags rotate per attempt and don't reproduce.

PR #3993, opened 2026-06-10
feat(cfc): label-aware sink-request confidentiality ceiling (audit item 21)

seefeldb:
## Summary

CFC spec-audit **item 21**. Sinks (`fetchData`/`fetchProgram`/`streamData`/`llm`/`generateText`/`generateObject`/`llmDialog`) verified only request *replay fidelity* (`verifySinkRequestRelease` deep-equals the post-commit request against the prepared snapshot) — **no label was ever consulted at the one place information flow ultimately matters: egress.** Waves 1–2 made stored labels trustworthy and persistent specifically so a consumer could exist; this is that consumer.

## Mechanism

`prepareBoundaryCommit` now gates every recorded `sink-request` write-policy input whose sink declares a confidentiality ceiling. The union of confidentiality reachable from the transaction's consumed reads (resolved via the same `getReadActivities` + `labelAtPath(storedMetadataFor(...))` path `verifyInputRequirements` already uses) must sit within the sink's ceiling; otherwise a reason is pushed.

- **Sound over-approximation**: the consumed set is transaction-global — a sink request is built from whatever the handler read, and the `sink-request` input carries no read-provenance of its own, so the whole consumed set is the conservative bound. Reading a labeled cell already marks the tx CFC-relevant, so the gate runs exactly when there's confidentiality to protect.
- **Rides observe→enforce for free**: a pushed reason invalidates prepare, which the existing commit gate turns into a *reject* only in enforcing modes; in `observe` it's a diagnostic naming the offending `(sink, atom)`.

## Policy surface

`sink-inventory.ts` gains `SinkMaxConfidentiality` + `DEFAULT_SINK_MAX_CONFIDENTIALITY` (**empty** — the mechanism is live but inert until a deployment declares ceilings via `Runtime({ cfcSinkMaxConfidentiality })`).

Empty-by-default is deliberate: value-copy laundering (S16) is still open, so a strict default would gate only the few *correctly-labeled* flows while laundered ones slip through anyway — net annoyance, little security. The honest landing is an opt-in ceiling rolled out per the standard observe→enforce path. Config is **write-once** on the tx (set by the Runtime before any handler runs, off the public control surface per the audit S3 posture) so handler code can't relax it mid-transaction.

## Testing

New `cfc-sink-ceiling.test.ts`, red-green: reject (enforce) / allow + diagnostic (observe) / within-ceiling-atom passes / ceiling on a *different* sink doesn't gate / no-ceilings-default passes / public request passes. Full `packages/runner` suite: 576 passed, 0 failed.

## Scope

Deferred (next-phase, per the audit): the default label transition + PC propagation that closes value-copy laundering (S16) — until then this gates *direct* labeled flows only. Populating real per-sink ceilings is a deployment decision (the LLM sinks legitimately carry user content and need their atom vocabulary settled first).

---
## Summary by cubic
Adds label-aware confidentiality ceilings for `sink-request` egress. Configure per-sink ceilings via `Runtime({ cfcSinkMaxConfidentiality })`; observe logs (and names offending atoms), enforce rejects.

- **New Features**
  - `prepareBoundaryCommit` checks the union of consumed read confidentiality (ancestor-or-equal and, for recursive reads, descendant labels) against per-sink ceilings; observe logs with offending atoms, enforce rejects.
  - Adds `SinkMaxConfidentiality` and `DEFAULT_SINK_MAX_CONFIDENTIALITY`; sinks absent from the map are ungated; `[]` means public-only.
  - Applies to recorded sink requests (`fetchData`, `fetchProgram`, `streamData`, `llm`, `generateText`, `generateObject`, `llmDialog`).

- **Bug Fixes**
  - Counts descendant labels only for recursive reads; shallow (`nonRecursive`) reads aren’t over-counted.
  - Makes ceilings immutable and write-once: deep-frozen in `Runtime` and when stored on the tx, preventing mid-transaction mutation.

<sup>Written for commit ad8b99a84f5da630351cdbfd5d0a87128a8752e6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3993?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seefeldb:
Addressed both review findings:
1. **Descendant labels (Codex P1):** `collectConsumedConfidentiality` used `labelAtPath`, which only matches ancestor-or-equal labelMap entries — so `cell.get()` then sending `value.secret` slipped a public-only ceiling. It now unions every labelMap entry that is ancestor-or-equal **or** a descendant of the read path (a read observes everything below it). New test: confidential field as a descendant of the read path rejects.
2. **Mutable config (cubic P1):** `cfcSinkMaxConfidentiality` is now deep-frozen (map + each per-sink array).

Full runner suite green. (2f621660d)

seefeldb:
Round-2 review addressed (and merged latest main, resolving the CfcTxState conflict with S18 #3996):
- **P1 (mutable ceiling reference):** `setCfcSinkMaxConfidentiality` now `deepFreeze`s the map on store, so the ceiling is immutable regardless of caller — not just relying on the Runtime-side freeze (TS `Readonly<>` is compile-time only).
- **P2 (reason omits atom):** the ceiling-violation reason now names the offending atom(s) (e.g. `exceeds ceiling for fetchData: {"...medical..."}`), so an observe-mode diagnostic identifies the exact `(sink, atom)` pair to allow. Test strengthened to assert the atom appears.

Also from the first round: descendant-label union now skips `nonRecursive` reads (they observe only the value at the path). Full runner suite: 567 passed, 0 failed.

PR #3995, opened 2026-06-10
refactor(runtime-client,lib-shell,shell): one runtime per identity; space is part of the address

jkomoros:
## What

Two movements, both from review feedback on the multi-space work:

**1. Every page operation names its space — no wire-level defaulting** (the original PR). All nine page-op request types carry a **required** `space: DID`; the worker's `space ?? homeSpace` fallback is deleted (a runtime guard turns an omitted space from an untyped caller into a clear error — it caught the browser-evaluated debug-global caller in CI within hours); `RuntimeClient` methods take a required space (`getPage` is `(pageId, space, runIt?)`).

**2. One runtime per identity — space is just part of the address** (this review's ask, implemented in full):

- `RuntimeInternals` is now bound to an (identity, host) pair, nothing else. No `#space`/`#spaceName`/`#isHomeSpace`, no `view` in `create()` — the worker session is always the identity's home session. Every space-scoped method takes the space as its **first parameter**: an address is `(space, id)`. The pattern cache and the space-root-pattern cache are keyed per space.
- Navigation registers a piece in **its own** space's root pattern and always emits `{spaceDid, pieceId}`. Mapping a DID back to a human-readable `spaceName` URL is view state, handled in the shell's navigate wrapper.
- The home-only root-pattern guard is **removed** — it made home special, which is the opposite of the model. Server-side ACLs are the write gate, as they already are for `CellSet`. One open design question this surfaces (flagging rather than burying it): navigating to a piece in a space without a root pattern provisions one seeded from the *navigator's* home `defaultAppUrl` — same as the old per-space worker did on view, but now reachable cross-space. Who properly provisions a shared space's root pattern is a federation question; happy to guard or change seeding if you want different semantics.
- **Creation names its space too**: `PageCreate` was the one page op with no `space` — with the worker now always initialized home, creation would have silently retargeted from "the viewed space" to "home" (e.g. `[[backlink]]` notes in cf-code-editor). It now carries a required `space` like everything else; the editor creates in its pattern's space.
- **Known deferral, not silent change**: blob upload still posts to the worker's init space (now always home). Routing it per-address is exactly the compute follow-up from your #3947 note (PR4, with llm/fetchData), so it's named here instead of changed here.
- `FavoritesManager`: storage stays home-scoped (identity-derived); the favorited piece's space is per-call.
- **Shell**: "current space" lives where it belongs — view state. `RootView` recreates the runtime **only on identity/apiUrl change**; a space switch no longer respawns the worker. `app.view` resolves to a space DID (new `resolveSpaceDid` helper for named spaces) and threads down to the views as a property. Per-space caches (HeaderView's pieces list) invalidate on the space, not the runtime.

## Tests

Worker: a spaceless request rejects; foreign spaces get distinct cached piece contexts; `handlePageGet({space: B})` resolves in B. Client: `PageHandle.start/stop` carry their cell's space. lib-shell: address-keyed pattern cache, per-space invalidation, identity-scoped construction. Shell: navigation suite updated for the 2-arg constructor and `{spaceDid}` targets. Full `deno task check`, lint, fmt green; lib-shell, runtime-client, shell, piece suites green.

## Perf-check note

The perf check flags pattern-unit timings this diff cannot affect (the unit-test shards import none of the changed packages). **Main's own latest perf check (run 27328932295) shows the identical metrics OVER** — pattern-unit-1 +70.7%, card-piles +51.2%, parking-coordinator +28.7% — i.e. a perf regression that landed on main in the last day (suspect: #3911's per-internal-cell change) and which merging main into this branch inherits. Worth a look from whoever owns that change, independent of this PR. Declaring the new levels per the perf-check's own suggestion:

NEW_PERF_BASELINE: test: pattern-unit-1/pattern-unit-tests = 209s
NEW_PERF_BASELINE: subtest: pattern-unit-1/pattern-unit-tests > packages/patterns/card-piles/main.test.tsx = 22s
NEW_PERF_BASELINE: subtest: pattern-unit-1/pattern-unit-tests > packages/patterns/cfc-group-chat-demo/main.test.tsx = 25s
NEW_PERF_BASELINE: subtest: pattern-unit-3/pattern-unit-tests > packages/patterns/factory-outputs/parking-coordinator/main.test.tsx = 122s
NEW_PERF_BASELINE: subtest: pattern-unit-1/pattern-unit-tests > packages/patterns/battleship/pass-and-play/main.test.tsx = 14s
NEW_PERF_BASELINE: subtest: pattern-unit-1/pattern-unit-tests > packages/patterns/battleship/multiplayer/lobby.test.tsx = 19s
NEW_PERF_BASELINE: subtest: pattern-unit-1/pattern-unit-tests > packages/patterns/cfc-authorized-save/main.test.tsx = 14s

NEW_PERF_BASELINE: step: Build application = 28s
NEW_PERF_BASELINE: subtest: pattern-unit-3/pattern-unit-tests > packages/patterns/factory-outputs/lot-watch/main.test.tsx = 50s
NEW_PERF_BASELINE: test: pattern-unit-3/pattern-unit-tests = 280s
NEW_PERF_BASELINE: subtest: pattern-unit-3/pattern-unit-tests > packages/patterns/google/core/experimental/gmail-agentic-search.test.tsx = 32s


packages/lib-shell/src/runtime.ts
seefeldb: ah, to be fully in the spirit of the idea behind this PR, I think we'd want to make it one runtime overall as well, not one runtime per space.

background: historically spaces were isolated, then we started the ability to load a piece in a space that can access other spaces and since the old shell always looks at one top-level piece there was still a semantic concept of "current space" (i.e. of the piece we look at). in loom shell that disappears entirely, it is possible to show multiple pieces at the same time from different spaces. so let's rip this historic piece out. there is one runtime, and space is just part of the address like id, nothing special about it.

jkomoros: Done — ripped out in this PR (212bbb2d4). RuntimeInternals is now bound to (identity, host) only: no #space/spaceName/isHomeSpace, no view in create(), and every space-scoped method takes the space as its first parameter — an address is (space, id). The shell's "current space" moved into view state: RootView resolves app.view → DID, threads it down as a property, and recreates the runtime only on identity/apiUrl change — a space switch no longer respawns the worker. Also removed the home-only root-pattern guard (home isn't special; server ACLs gate writes) and made FavoritesManager take the piece's space per call. The historic piece is gone.

packages/shell/src/globals.ts
```diff
@@ -1,10 +1,16 @@
...
 
 declare global {
   var app: App;
   var commonfabric: {
     rt?: RuntimeClient;
+    /**
+     * The bound space of the active runtime. Page operations on `rt`
+     * require an explicit space — console/debug callers pass this.
+     */
+    space?: DID;
```
seefeldb: that smells, this shouldn't be necessary anymore / points to something still assuming a single space.

jkomoros: Right — the smell traced to the integration helper: waitForRuntimeSynced was passing "the" space to synced(), but its actual question (is the worker quiescent?) isn't per-space at all. Fixed in 68e530366: new spaceless RuntimeSynced op (like Idle — awaits convergence for every space the worker has opened), the helper calls rt.allSynced(), and commonfabric.space is deleted everywhere. Nothing assumes a single space anymore.

jkomoros:
Landing on seefeldb's in-person approval (relayed by @jkomoros). All checks green; both review threads implemented: one runtime per identity with space-first addressing, and the spaceless allSynced replacing the commonfabric.space global.

PR #3996, opened 2026-06-10
fix(cfc): privileged-write guard for the ["cfc"] label-map path (S18)

seefeldb:
## Summary

CFC spec-audit **S18 / item 11**. A direct write to a document's `["cfc"]` metadata path forges the persisted label map that drives CFC derivation for *other* writes — defeating `requiredIntegrity` gates / the prompt-injection screen — and it bypasses the commit-boundary derivation **and** the S4 runtime-evidence mint-gating entirely (S4 gates the *derivation* path; a direct `["cfc"]` write never goes through it). `valueWriteTargets` silently **excluded** these writes from verification by address pattern; this replaces that silent exclusion with an **explicit privileged-writer assertion at the extended-tx layer**, exactly as item 11 specifies.

## Mechanism

- `noteSystemWrite` (called in `write`/`writeOrThrow`) records any `["cfc"]`-path write made **outside** the runtime's privileged scope. `prepareBoundaryCommit` turns each recorded address into a fail-closed reason, so it surfaces uniformly with every other CFC reason: **enforce rejects, observe diagnoses, disabled stays inert**.
- The runtime's own label/schema persistence runs via `runPrivilegedSystemWrite` wrapped around `prepareBoundaryCommit` — the one legitimate `["cfc"]` writer, so it alone is exempt. The scope method is **private** (not on `IExtendedStorageTransaction`) so handler code holding `cell.tx` can't grant itself privilege (audit S3 posture).

## Scope & threat coverage

Guards `["cfc"]` only, deliberately:
- **`cid:`** schema-doc poisoning is already neutralized by **S5**'s read-side canonical-hash re-derivation (a poisoned schema is rejected on read), so guarding the write adds nothing and would break legitimate schema-doc seeding.
- **`["source"]`** link writes go through the link-write CFC machinery (`derivePersistedLinkLabel` marks relevance + verifies) — not an unguarded hole.

The normal Cell API writes *value* paths and structurally cannot address the document's `["cfc"]` field, so ordinary pattern code is unaffected; path-`[]` full-document seeds / storage hydration are likewise not gated (they're the legitimate hydration vector). The residual path-`[]`-with-cfc-field vector requires explicit document-address construction — the S18 "sandbox keeps raw tx away from untrusted code" load-bearing assumption, out of scope for this layer.

## Tests

New `cfc-privileged-system-write.test.ts`, red-green: forged `InjectionSafe` via direct `["cfc"]` write rejected (enforce) / allowed-with-diagnostic (observe) / the runtime's own labeled persistence still commits / value + path-`[]` writes unaffected.

Two existing unit tests (`cfc-boundary`, `scheduler-core`) seeded stored metadata via direct `["cfc"]` writes, standing in for the runtime — updated to seed *through* the privileged scope (a grep confirmed these were the only non-privileged `["cfc"]` writers anywhere; zero in production). Full `packages/runner` suite: **576 passed, 0 failed**.

## Deferred

**S8** (sqlite per-column `ifc` in mutable handle-cell data) is a different shape — value-path *monotonicity* (grow-only), not a system-path privilege guard — and touches the hot `db.exec` path, so it's a follow-up rather than folded in here.

---
## Summary by cubic
Adds a privileged-write guard for the document `["cfc"]` label-map path to prevent untrusted metadata writes that would bypass commit-boundary CFC derivation (spec-audit S18). Unprivileged `["cfc"]` writes now fail in enforce and log a diagnostic in observe; the runtime’s own CFC persistence remains allowed via an ECMAScript-private scope.

- **Bug Fixes**
  - Record unprivileged `["cfc"]` writes in `write`/`writeOrThrow` via `noteSystemWrite` and surface them as fail-closed reasons in `prepareBoundaryCommit`.
  - Run `prepareBoundaryCommit` inside `#runPrivilegedSystemWrite` so only the runtime can persist CFC metadata; the scope is ECMAScript-`#private` and not on `IExtendedStorageTransaction`, blocking `(cell.tx as any)` escalation.
  - Value-path and path-`[]` writes are unchanged.
  - Tests: added enforce/observe coverage and a no-escalation test; updated two tests to seed CFC metadata via path-`[]` full-document writes.

<sup>Written for commit 4f786d82ec2c394ecfa60544761af942c2273f6c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/3996?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/cfc/prepare.ts
+++ b/packages/runner/src/cfc/prepare.ts
@@ -2203,6 +2203,14 @@ export const prepareBoundaryCommit = (
 ): string[] => {
   const reasons: string[] = [];
   const state = tx.getCfcState();
+  // A write to a document's ["cfc"] label-map path made outside the runtime's
+  // privileged persistence scope forges the metadata that drives CFC derivation
+  // for other writes (audit S18). Each was recorded at the extended-tx write
+  // chokepoint; surface one fail-closed reason apiece so it rejects in enforce
+  // mode and diagnoses in observe, uniformly with every other reason here.
+  for (const target of state.unprivilegedSystemWrites ?? []) {
+    reasons.push(`unprivileged write to protected cfc path ${target}`);
+  }
   const identityForInput = (
     input: WritePolicyInput,
   ): ImplementationIdentity | undefined =>
--- a/packages/runner/src/cfc/types.ts
+++ b/packages/runner/src/cfc/types.ts
@@ -279,4 +279,10 @@ export type CfcTxState = {
   implementationIdentity?: ImplementationIdentity;
   outbox: PostCommitSideEffect[];
   diagnostics: string[];
+  // Addresses of writes to a document's ["cfc"] label-map path made OUTSIDE the
+  // runtime's privileged persistence scope (audit S18). The runtime's own label
+  // writes in prepareBoundaryCommit run privileged and never land here; anything
+  // that does is forging metadata that drives derivation for other writes, so
+  // prepareBoundaryCommit turns each into a fail-closed reason.
+  unprivilegedSystemWrites: string[];
 };
--- a/packages/runner/src/storage/extended-storage-transaction.ts
+++ b/packages/runner/src/storage/extended-storage-transaction.ts
@@ -114,7 +114,9 @@ export class ExtendedStorageTransaction implements IExtendedStorageTransaction {
   // Depth of the runtime's privileged system-write scope. The runtime's own
   // label/schema persistence (prepareBoundaryCommit) runs inside it; any write
   // to a protected system path outside it is recorded as unprivileged (S18).
-  private privilegedSystemWriteDepth = 0;
+  // ECMAScript-private (#) so handler code reaching cell.tx cannot enter the
+  // scope via `(cell.tx as any)` — `as any` cannot touch a `#private` member.
+  #privilegedSystemWriteDepth = 0;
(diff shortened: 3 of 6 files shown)


seefeldb:
Addressed (Codex P1, privilege reachability): the privileged scope was a TS-`private` method, and TS `private` is compile-time only — `(cell.tx as any).runPrivilegedSystemWrite(...)` could bypass the guard. It's now an **ECMAScript `#private`** method (`#runPrivilegedSystemWrite` / `#privilegedSystemWriteDepth`): `(cell.tx as any).#…` is a TypeError, so no escalation entry point exists on the tx object. Tests that need stored `["cfc"]` metadata now seed via the ungated path-`[]` full-document write (how hydration delivers it) rather than the privileged scope; a new test asserts the tx exposes no `runPrivilegedSystemWrite` property. Full runner suite green. (4f786d82e)

On the broader threat: a fully in-realm hostile actor can still reach `cell.tx.tx` (raw) or construct a path-`[]` write — those remain the sandbox's responsibility (the S18 'sandbox keeps raw tx away from untrusted code' load-bearing assumption). This change closes the specific escalation method the review identified.

PR #4011, opened 2026-06-10
feat(cfc): S16 default transition — flow-label propagation (A0+A1)

seefeldb:
## Summary

CFC spec-audit **S16** (value-copy laundering): read labeled data, write a derived plain value to an unlabeled cell, commit — the label vanished. This PR lands the default transition: a transaction's outputs are tainted by what it read.

Design + decisions: `docs/specs/cfc-s16-default-transition-design.md` (this PR). The companion spec changes SC-1..16 are already applied to the CFC spec repo with sorry-free Lean mechanization (StoreComponents / TriggerReads / DegenerateJoin).

## Changes (in commit order)

1. **flowPrecisionClaim minting deleted** — the list builtins' per-element ops run in their own transactions reading only their element, so pointwise precision is structural, not a trusted claim. Claims were consumed by nothing. The schema key stays reserved-and-tolerated for already-persisted link schemas.
2. **S7 subtree join** — `effectiveReadLabel`: recursive reads (e.g. `getRaw()`, which journals one root read and hands over labeled children) join descendant labelMap entries; non-recursive reads keep ancestor-or-equal resolution. Red-green on a getRaw-shaped `maxConfidentiality` bypass.
3. **labelMap v2 provenance components** — entries gain `origin: declared|link|derived` with per-component update disciplines (declared = monotone store policy; link/derived = per-value). Per-(path, origin) coalescing; read resolution joins each component's most-specific entry. Additive wire change (origin optional, runner sole reader).
4. **No-churn persistence pinned** — identical recomputed metadata produces zero `["cfc"]` writes (journal novelty diffing), so reactive re-derivation can't churn storage.
5. **A1 flow labels** — `cfcFlowLabels: off|observe|persist` runtime dial (default **off**; pinned once persist, mirroring the enforcement floor):
   - one conservative confidentiality join per tx over all non-internal consumed reads + dereference traces (cid:/cfc/source excluded)
   - `persist`: every value write target gets a derived component at its written paths; derived/link components at-or-below a written path are replaced/cleared (per-value semantics); the Wave 2 grow-only ratchet now applies to legacy entries only when persisting
   - relevance is computed (labeled read or labeled-doc write auto-marks at the prepare chokepoint + commit gate; self-minted metadata excluded; probe only while unprepared so it can't invalidate a prepared digest)
   - derived entries are recorded taint, not authored policy — they don't demand a schema write-policy input on later writes

## Acceptance

The S16 laundering repro is green end-to-end: the laundered copy persists `{origin: "derived", confidentiality: ["secret"]}` and a downstream consumer is rejected by the **existing** maxConfidentiality gate — propagation feeding enforcement with no new rejection machinery. Replace-on-overwrite pinned (untainted raw write clears the derived component; note `cell.set()` journals a read of the prior value, so set-overwrites conservatively re-derive).

## Testing

Full runner suite green at every commit (566 passed / 0 failed). New: `cfc-flow-labels.test.ts`, `cfc-labelmap-components.test.ts`, `list-result-schema.test.ts`, getRaw-shaped boundary test.

---
## Summary by cubic
Implements the S16 default transition: each transaction derives a confidentiality label from what it reads and persists it on its writes as `origin: "derived"`. Also adds a `cfcFlowLabels` dial, upgrades `labelMap` to v2 components, removes unused flow-precision claims, and updates the S16 design docs to note raw-path keying of internal surface exclusions.

- **New Features**
  - Per-tx join of consumed read labels (+ dereference traces), excluding internal/`["cfc"]`/`["source"]`/`cid:` reads.
  - Persist derived labels on all value write targets; replace-on-overwrite and clear derived/link descendants under an ancestor write.
  - LabelMap v2 with `origin: "declared" | "link" | "derived"`; read resolution joins per-component; no-churn persistence when unchanged.
  - Computed CFC relevance with a fast path; new `cfcFlowLabels: "off" | "observe" | "persist"` (pinned once set to persist).
  - List builtins now attach plain array schemas via `listResultSchema`; pointwise precision comes from per-element transactions.
  - Laundering repro now persists `{ origin: "derived", confidentiality: ["secret"] }` and is gated by existing `maxConfidentiality`.
  - Merge reconciles with sink max-confidentiality ceilings and the S18 privileged-write guard; the propagation dial coexists with these gates.

- **Bug Fixes**
  - Recursive reads (e.g. `getRaw()`) subtree-join descendant labels during read resolution to close S7, so parent reads taint outputs correctly.
  - Value-write and flow-read surface exclusions now key on raw storage paths, so user fields like `["value","source"]` are not treated as internal; prevents missed derived-label writes/clears and taint-join gaps in `cfcFlowLabels: "persist"`.

<sup>Written for commit 7ae1b432c558ee4113cfb97a1fb40e4860899f3b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4011?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

Perf-check override — the flagged `Package Integration Tests (runner)` wall time is not caused by this PR: per-test timings are identical to main (within 0.2s, including the CFC sqlite label tests); the suite grew by ~5s on main itself (PR #3974 added `sqlite-cfc-row-label.test.ts`, inherited here via the main merge) while the 20-run baseline median still predates it, and the rest is runner setup-step variance (20s vs 31s observed on identical code).

NEW_PERF_BASELINE: job: Package Integration Tests (runner) = 95s


seefeldb:
Addressed the Codex P1 (flow labels dropped for user `source` fields) in 638d2bc04.

Confirmed real, and slightly broader than flagged: the same canonical-path conflation also affected the read side (`flowReadExcluded` dropped reads of a labeled user `value.source` from the taint join) and dereference-trace ends (which carry value-relative link paths, so a name-based exclusion there could only ever hit user fields).

Fix, per the suggestion, keys the exclusions on the raw storage path:
- `valueWriteTargets`: `cfc`/`source` exclusions now test `write.address.path` before `value`-stripping, so a tainted write to user `value.source` enters the flow targets (derived label attached, stale per-value components cleared). The link-valued `internal` exclusion deliberately stays canonical — it covers the runtime's link plumbing both at the root surface and inside process-doc values, and link writes carry their labels via the link-write machinery.
- `flowReadExcluded`: now takes the raw read-activity path, so reads of a labeled user `value.source` taint the join.
- Trace ends: only the `cid:` exclusion remains — a trace path like `["source"]` IS the user field `value.source`.

Red-green: new test "does not exempt user value fields named like runtime surfaces" in cfc-flow-labels.test.ts fails on the write side without the write fix and on the read side without the read fix (both verified independently). Full runner suite green (570/0).

PR #4014, opened 2026-06-10
fix(runner): make inSpace child pieces independently loadable from their own space (CT-1687)

bfollington:
Fixes the CT-1687 navigation failure: clicking a profile's cell-link in the multi-profile picker (#3830) showed **"Failed to load piece — Pattern of:fid1:… has no stored source"**, leaving profiles effectively uneditable (the picker has no inline edit UI).

## Root cause

A `Factory.inSpace(...)` child piece (each profile lives in its own space) was never independently loadable from that space. A fresh runtime navigating to the child loads pattern artifacts from `resultCell.space` — the child's space — where nothing was ever persisted, for three stacked reasons:

1. `savePattern` deduped on `patternId` alone, so the parent space's copy short-circuited every save into another space.
2. An ESM bundle **sub-pattern** (e.g. `ProfileHome` inside the home bundle) carries no `program` object, so even a non-deduped save had nothing to write.
3. The compile-cache write-back (source + compiled closures) only targets the space the parent bundle compiled into — and `inSpace`-derived factories weren't registered as derivation copies, so they didn't even resolve to a content-addressed artifact entry ref (which also meant the child's result cell never got `patternIdentity` meta).

## Fix

- **builder/pattern.ts** — `asScope`/`inSpace` factory copies register via `noteDerivedCopy`, so identity facts (trust, the artifact entry ref) resolve through to the root factory. Child result cells now carry `patternIdentity` meta.
- **pattern-manager.ts** — `savePattern` dedupes per `(patternId, space)` (canonical meta cell stays authoritative; falls back to its stored program when the in-memory pattern is source-free). New `replicatePatternToSpace` copies the **verified source + compiled closures** from the originating space into the child's space — fire-and-forget, tracked by `flushCompileCacheWrites`, all-or-nothing (a partial closure can never be served), deduped per `(entryIdentity, space)` per session with retry-on-failure.
- **runner.ts** — replication hooks at both cross-space child creation sites:
  - `instantiatePatternNode` (parent-graph children, where `resultCell.space !== resultCellLink.space`), and
  - `handleJavaScriptHandlerResult` (handler-frame children — the profile-create path — where the whole deferred result pattern materializes *inside* the target space, so the per-node hook never sees a space transition; the originating space is only visible here).

On reload, the child then takes the existing by-identity path (`loadPatternByIdentityAs`) from its own space's replicated closure.

## Verification

- New regression test `packages/runner/test/inspace-child-reload.test.ts`: a handler pushes `child.inSpace(B)({...})` onto a list (mirrors `profile-create.tsx` 1:1); a **fresh runtime** starts the child piece from space B. Fails with the exact production error before the fix; passes after.
- Full `packages/runner` suite: **570 passed, 0 failed**.
- **Live A/B against a running toolshed** (local dev): child created via the un-fixed CLI → `cf piece step` fails with `Pattern … has no stored source`; child created via the fix-bearing CLI → loads and steps from its own space in a fresh CLI runtime.

## Known limitations (follow-ups, noted in CT-1687)

- **No backfill**: children created before this fix stay unloadable until recreated (replication fires at creation time).
- **Version-skew durability**: compiled-closure docs are keyed by `COMPILE_CACHE_RUNTIME_VERSION`; after a runtime version bump a sub-pattern child has no meta-program recovery path (the replicated *source* docs are in place for a future by-identity source-recovery path).
- Inline *display* of another space's owner-protected profile fields is a separate gap (CT-1667, the cross-space read path).

Closes CT-1687.

---
## Summary by cubic
Makes `inSpace` child pieces independently loadable from their own space. Fixes CT-1687 where navigating to a profile failed with “has no stored source” in a fresh runtime, and hardens replication against write-back races.

- **Bug Fixes**
  - Builder: `asScope`/`inSpace` factories register via `noteDerivedCopy` so identity (artifact entry ref) flows to children.
  - Pattern manager: `savePattern` dedupes by `(patternId, space)`, falls back to the canonical stored program, and verifies the save on commit error (releases the claim if it didn’t land). New `replicatePatternToSpace` copies verified source and compiled closures from the parent’s space to the child’s space; it now awaits in-flight cold-compile write-backs to avoid races, is fire-and-forget, and retries on failure.
  - Runner: replication hooked in both cross-space creation paths (`handleJavaScriptHandlerResult`, `instantiatePatternNode`).
  - Tests: regression coverage for handler-created and build-time `inSpace` children; both load in a fresh runtime from the child’s space.
  - Docs: module-loading spec notes cross-space closure replication and server-compilation forward-compat.

- **Migration**
  - Recreate previously created `inSpace` children; there’s no backfill for existing items.

<sup>Written for commit 92d191ebe0c5b5aa34615241c0dd9f5166cb77bd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4014?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/module-loading.md
+++ b/docs/specs/module-loading.md
@@ -433,6 +433,17 @@ The cache is designed around this:
   server**: the server becomes the sole acceptor of that write integrity (it
   only stamps the label from its own compilation step), making the label a hard
   guarantee — with no change to the read path, which already requires it.
+- **Cross-space closure replication (CT-1687).** Cache docs do not only live
+  where they compiled: when a pattern materializes a child piece in another
+  space (`Factory.inSpace(...)`), the runner replicates the child pattern's
+  source + compiled closures into the child's space so the piece is
+  independently loadable there (`PatternManager.replicatePatternToSpace`).
+  Chain-of-custody holds — compiled docs are read through the integrity-gated
+  loader (a user can only replicate docs already carrying their own stamp) and
+  re-stamped on the child-space write by a legitimate child-space writer. Note
+  for the server-compilation end state: a client can then no longer stamp
+  replicated compiled docs, so child spaces will need server-side replication
+  or by-identity source recovery instead.
 - **CFC verified-source derives from the source set, not the cached JS.** A
   poisoned-but-SES-safe JS document must not be able to spoof `fn.src` /
   authorship, so the CFC verified-source identity is anchored to the
--- a/packages/runner/src/builder/pattern-metadata.ts
+++ b/packages/runner/src/builder/pattern-metadata.ts
@@ -97,10 +97,13 @@ export function brandTrustedPattern<T>(value: T): T {
 }
 
 // Derivation tracking: `copy → root original`. Replaces the former
-// `unsafe_originalPattern` symbol backref. Registered ONLY by runner-owned
+// `unsafe_originalPattern` symbol backref. Registered ONLY by trusted-builder
 // copy sites (`noteDerivedCopy` callers: build-time graph serialization in
 // json-utils, traversal copies in traverse-utils, binding copies in
-// pattern-binding) — authored/forged values never enter, since nothing on the
+// pattern-binding, and the `asScope`/`inSpace` factory derivations in
+// builder/pattern.ts — the latter reachable from authored pattern code, which
+// is sound because both objects are builder-minted and already branded) —
+// forged values never enter, since nothing on the
 // object itself can establish the link. Module-level (not per-manager): the
 // copy sites live in builder-layer utilities with no PatternManager handle,
 // and the linked facts (trust brands, content-addressed entry refs) are
--- a/packages/runner/src/builder/pattern.ts
+++ b/packages/runner/src/builder/pattern.ts
@@ -20,7 +20,7 @@ import {
   type UnsafeBinding,
 } from "./types.ts";
 import { opaqueRef } from "./opaque-ref.ts";
-import { brandTrustedPattern } from "./pattern-metadata.ts";
+import { brandTrustedPattern, noteDerivedCopy } from "./pattern-metadata.ts";
 import {
   applyArgumentIfcToResult,
   applyInputIfcToOutput,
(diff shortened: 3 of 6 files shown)


bfollington:

A cf-review pass (report-only) found **no blockers** and 3 should-fixes; all actioned in the follow-up commit:

1. **Write-back race** — `replicateClosures` could read the origin space while that session's own cold-compile write-back was still in flight; a lost race threw, and for handler-created children (one space per profile) nothing re-fired the released dedupe key — silently re-creating the CT-1687 symptom. Fixed: cold-compile write-backs are now tracked in their own `pendingCacheWriteBacks` set and awaited before the replication read (a `flushCompileCacheWrites()` inside would await the replication's own tracked promise — self-deadlock).
2. **`savePattern` claim asymmetry** — the per-space "saved" claim was recorded before the fire-and-forget commit settled, and a real failure (no write access to the child space, offline) was logged as the benign "Pattern already existed" and never retried. Fixed with verify-then-release: on commit error, sync the meta cell — if the program landed (the routine content-addressed conflict), the claim stands; otherwise the claim is released so the next save retries.
3. **Test coverage** — added a second regression test pinning the `instantiatePatternNode` hook (a build-time `child.inSpace(X)({...})` node in the parent's graph), the path the handler test can't reach.

Nits taken: updated the stale `noteDerivedCopy` caller enumeration in `pattern-metadata.ts` (now includes the `asScope`/`inSpace` derivations and their authored-code reachability rationale); pinned the `runtimeFingerprint` default coupling with a comment at the `loadVerifiedSourceClosure` call; added a spec paragraph to `docs/specs/module-loading.md` covering cross-space closure replication and its forward-compat note (under server-side compilation, clients can no longer stamp replicated docs — child spaces will need server-side replication or by-identity source recovery).

Consciously accepted (per review): `registerPattern` before the dedupe checks (cost profile matches the existing setup paths; cross-space children are rare); per-space meta replicas as point-in-time snapshots (performance-only consequence); meta replication degrading after pattern-cache eviction (edge case, closure replication unaffected); unbounded per-session dedupe sets (tiny strings bounded by real work).

Full report archived from the review run; verification after fixes: both regression tests green, `esm-cell-cache-integration` + `by-identity-handler-exec` + `pattern-scope` green, full runner suite re-run green.

PR #4018, opened 2026-06-10
cf-harness: atomic artifact writes + require CFC sidecar transports for enforce modes

mathpirate:
## What this does

Two correctness/robustness fixes to the cf-harness CFC enforcement surface, found during a correctness pass over `cf-harness` measured against the `specs/cfc/` requirements and the gVisor CFC fork. Each is a separate commit.

### 1. Atomic artifact writes (`fix: atomic artifact writes via temp-file rename`)

`run-state.json` is the canonical resume + CFC audit record and is rewritten on **every** tool call via a plain truncating `Deno.writeTextFile`. A crash mid-write left a half-written file that breaks resume and corrupts the audit trail. Switched `writeJsonFile` to write a sibling temp file and `rename` (atomic on the same filesystem), with temp cleanup on failure. All artifact persists route through `writeJsonFile`, so this covers run-state, transcript, policy trace, tool outputs, etc. No behavior change on the happy path.

### 2. Require CFC sidecar transports for enforce modes (`feat: require CFC sidecar transports for enforce modes`)

**The bug:** enforce modes silently depend on two trusted sidecar transports being wired:

- **invocation-context transport** — the harness writes the initial-taint context the sandbox reads *in*. Missing ⇒ the sandbox starts untainted and input labels (prompt-slot influence, prior observed labels) are silently dropped.
- **result dir** (`cfcResultDir`) — runsc writes the final-taint result the harness reads *back* to mediate output. Missing ⇒ every command's CFC result is absent and enforce-mode mediation fail-closes every observation.

Both came **only** from `CF_HARNESS_RUNSC_CFC_*` env vars. The CLI never set them, and nothing validated them. So `--cfc-enforcement-mode enforce-strict` without those env vars produces a run that fails closed on all output (safe, but looks like a mysteriously dead agent) **and** silently seeds no input taint (unsafe) — with zero operator-visible signal that enforcement is degraded.

**The fix:** `assertDockerRunscCfcTransportForMode` rejects an enforce-mode run whose docker-runsc sandbox lacks either transport. Added first-class CLI flags `--cfc-result-dir` / `--cfc-invocation-context-dir` (resolved against cwd) alongside the existing env-var fallback, so a run records how it was wired instead of relying on ambient env.

---

## Design decisions in #2 — please evaluate

Two choices here are judgment calls. I'm flagging both so reviewers can push back.

### A. Gate at **run start**, not at construction

I considered enforcing the transport floor in the `CfHarnessEngine` constructor (strongest fail-fast). I moved it to run start (`#assertCfcTransportReady`, idempotent) instead — fired at diagnostics init (whose capability probes execute scripts inside the sandbox) or the first tool invocation, whichever comes first — because:

- An engine is legitimately constructed and **inspected** without ever executing sandbox code — config-threading tests, `--describe-capabilities`. A construction-time throw breaks those for no safety gain: no sandboxed command has run, so nothing has leaked. (An earlier revision wrongly listed capability probes as inspection — willkelly's review caught that the probes run scripts in the sandbox via the prompt loop's diagnostics init *before* the first tool call, so the floor now fires there too.)
- The invariant I actually care about is *"nothing executes in the sandbox under enforcement without the transports,"* and run-start is exactly that boundary. It still fails fast — before the **first** command — so a mis-wired real run aborts loudly rather than emitting silent denials.
- Concretely: gating at construction broke ~23 CLI plumbing tests that build a real engine in the default `enforce-explicit` mode and mock the prompt loop. Those failures weren't false positives to suppress — they're the tell that construction-time is too early. (The one test that genuinely executes a tool, host `run_skill_script`, I switched to `--cfc-enforcement-mode disabled` with a comment, since host skill-script execution is outside CFC mediation by design and that test exercises target threading, not enforcement.)

Trade-off: a misconfigured enforce run that never starts (no diagnostics init, no tool call) won't be flagged. That's acceptable — such a run does nothing observable and seeds no taint. If reviewers prefer maximal fail-fast, moving the call into the constructor is a one-line change, but it requires every construct-and-inspect call site (incl. tests) to wire transports or pin a non-enforce mode.

### B. Require **both** transports, and only for the `enforce-*` floor

- **Both, not just the result dir.** The result transport alone gives fail-closed *output* (visibly broken, so self-correcting). The invocation-context transport is the dangerous one to omit: it fails *open* on input — the sandbox just starts untainted and the operator sees a working-looking agent. Requiring both closes the silent-degradation path, which is the whole point.
- **Only `enforce-explicit` / `enforce-strict`.** The predicate is `cfcEnforcementStrictness(mode) < CFC_ENFORCING_STRICTNESS` (reusing the runner's canonical floor constant). `disabled` and `observe` neither mediate nor deny tool output, so a missing transport can't silently weaken them — `observe` is explicitly *not* gated. (I initially used `=== 0`, which wrongly swept in `observe`; the test suite caught it.)

---

## Testing

- New validator unit tests over every `(mode × transport)` combination (`docker-runsc-sandbox.test.ts`).
- New engine tests: construction in enforce-without-transport **succeeds**; the first tool run **aborts**; a tool run with both transports wired **proceeds** (`engine.test.ts`).
- Full `cf-harness` suite green (407 passed). `deno fmt --check`, `deno lint`, `deno check` clean on all changed files.

## Out of scope (documented, not fixed here)

A fuller assessment from the pass — including that the `bash` curl-policy only constrains the literal word `curl` (real egress is gated by gVisor sink rules + docker `--network`), a small TOCTOU window in `bash-no-sandbox` host-path checks, and gVisor egress heuristics (port-443 TLS inference, snooped DNS cache) — lives in `session_outputs/2026-06-10_cf-harness-correctness-pass/ASSESSMENT.md`. Happy to spin any of those into follow-ups.

---
## Summary by cubic
Make artifact writes atomic to prevent corrupt resume/audit files, and require CFC sidecar transports for enforce modes with a run-start guard limited to harness-managed docker-runsc sandboxes. Adds `--cfc-result-dir` and `--cfc-invocation-context-dir` and fails fast before capability probes or the first tool if wiring is missing.

- **Bug Fixes**
  - Write JSON artifacts via temp file + rename to ensure atomic writes (run-state, transcript, policy trace, tool outputs).
  - Apply the enforce-mode transport guard at diagnostics init (before capability probes) and on first tool call.
  - Scope the transport check to engine-owned docker-runsc runtimes so injected runtimes aren’t falsely rejected.

- **Migration**
  - For `enforce-explicit` and `enforce-strict` with a harness-managed docker-runsc sandbox, set both `--cfc-result-dir <host-path>` and `--cfc-invocation-context-dir <host-path>` (or `CF_HARNESS_RUNSC_CFC_*` env vars). Missing either errors before any sandbox execution. `disabled` and `observe` are unchanged.

<sup>Written for commit eb57e3a800c18dfb6a3e07179d05da9e542a4f30. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4018?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cf-harness/src/artifacts.ts
+++ b/packages/cf-harness/src/artifacts.ts
@@ -59,8 +59,21 @@ const resolveTranscriptPathWithinRunRoot = (
     : fallbackPath;
 };
 
+// Artifacts such as run-state.json are the canonical audit and resume record and
+// are rewritten on every tool call. A crash partway through a plain truncating
+// write would leave a half-written file that breaks resume and corrupts the CFC
+// audit trail. Write to a sibling temp file and rename, which is atomic on the
+// same filesystem, so readers only ever observe a complete prior or new version.
 const writeJsonFile = async (path: string, value: unknown): Promise<void> => {
-  await Deno.writeTextFile(path, `${JSON.stringify(value, null, 2)}\n`);
+  const contents = `${JSON.stringify(value, null, 2)}\n`;
+  const tempPath = `${path}.${crypto.randomUUID()}.tmp`;
+  try {
+    await Deno.writeTextFile(tempPath, contents);
+    await Deno.rename(tempPath, path);
+  } catch (error) {
+    await Deno.remove(tempPath).catch(() => {});
+    throw error;
+  }
 };
 
 export interface HarnessArtifactStore {
--- a/packages/cf-harness/src/cli.ts
+++ b/packages/cf-harness/src/cli.ts
@@ -56,6 +56,8 @@ import type {
 } from "./contracts/transcript.ts";
 import { CfHarnessEngine } from "./engine.ts";
 import {
+  CFC_INVOCATION_CONTEXT_DIR_ENV,
+  CFC_RESULT_DIR_ENV,
   DEFAULT_DOCKER_RUNSC_IMAGE,
   DEFAULT_FABRIC_MOUNT_PATH,
 } from "./sandbox/docker-runsc.ts";
@@ -117,6 +119,8 @@ const CLI_STRING_FLAGS = [
   "structured-result-schema-file",
   "run-manifest",
   "cfc-enforcement-mode",
+  "cfc-result-dir",
+  "cfc-invocation-context-dir",
   "sandbox-image",
   "max-model-turns",
   "fabric-mount",
@@ -173,6 +177,8 @@ export interface CfHarnessCliConfig {
   structuredResult?: CfHarnessStructuredResultConfig;
   runManifestPath?: string;
   cfcEnforcementModeOverride?: CfcEnforcementMode;
+  cfcResultDir?: string;
+  cfcInvocationContextDir?: string;
   browserAccess?: HarnessBrowserAccessLease;
   maxModelTurns: number;
   printTranscript: boolean;
@@ -350,6 +356,8 @@ Options:
   --browser-access-profile-mode <mode> persistent | transient
   --browser-access-account-access <access> available | none
   --cfc-enforcement-mode <mode> disabled | observe | enforce-explicit | enforce-strict
+  --cfc-result-dir <path>       Host dir where runsc writes the CFC result sidecar (required for enforce-* modes)
+  --cfc-invocation-context-dir <path> Host dir where the harness writes the CFC invocation-context sidecar (required for enforce-* modes)
   --sandbox-image <image>       Docker image for the runsc-cfc sandbox (default: ${DEFAULT_DOCKER_RUNSC_IMAGE})
   --fabric-mount <path>         Host path for a Fabric FUSE mount (mounted at /fabric in the sandbox)
   --host-mount <spec>           Extra host bind mount (repeatable: name=<id>,source=<host>,target=<sandbox>,mode=readonly|writable)
@@ -363,7 +371,28 @@ Environment:
   OPENAI_API_KEY                Fallback API key if CF_HARNESS_API_KEY is unset
   CF_HARNESS_DOCKER_NETWORK_MODE none | bridge | host (default: bridge)
   CF_HARNESS_SANDBOX_IMAGE      Default value for --sandbox-image
+  ${CFC_RESULT_DIR_ENV} Fallback for --cfc-result-dir
+  ${CFC_INVOCATION_CONTEXT_DIR_ENV} Fallback for --cfc-invocation-context-dir
 `;
+
+// CFC sidecar transport dirs may be supplied by flag (resolved against cwd so
+// relative paths work) or env-var fallback (already an absolute host path by
+// convention). The docker-runsc layer re-validates that the result is absolute.
+const resolveOptionalCfcDir = (
+  flagValue: unknown,
+  envValue: string | undefined,
+  cwd: string,
+  flagName: string,
+): string | undefined => {
+  if (typeof flagValue === "string") {
+    const trimmed = flagValue.trim();
+    if (trimmed === "") {
+      throw new Error(`${flagName} requires a non-empty path`);
+    }
+    return resolve(cwd, trimmed);
+  }
+  return envValue;
+};
 
 const parsePositiveInteger = (
   input: string | undefined,
@@ -1190,6 +1219,10 @@ export const parseCfHarnessCliArgs = async (
       ),
       CF_CFC_MODE: Deno.env.get("CF_CFC_MODE"),
       CF_HARNESS_SANDBOX_IMAGE: Deno.env.get("CF_HARNESS_SANDBOX_IMAGE"),
+      [CFC_RESULT_DIR_ENV]: Deno.env.get(CFC_RESULT_DIR_ENV),
+      [CFC_INVOCATION_CONTEXT_DIR_ENV]: Deno.env.get(
+        CFC_INVOCATION_CONTEXT_DIR_ENV,
+      ),
     };
   const rawSandboxImage = typeof args["sandbox-image"] === "string"
     ? args["sandbox-image"].trim()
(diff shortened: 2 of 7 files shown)


willkelly (changes requested):
Requesting changes for one fail-closed gap in the new transport floor.

`CfHarnessPromptLoop.runTranscript()` calls `engine.ensureDiagnosticsInitialized()` before any model turn/tool call, and `ensureDiagnosticsInitialized()` calls `collectHarnessCapabilitySnapshot()`, which executes `sandbox.runShell()` for the capability probes. The new transport guard is only in `CfHarnessEngine.invokeBuiltinTool()`, so it does not cover that earlier sandbox execution.

I reproduced this with an engine-owned docker-runsc runtime in `enforce-explicit` and no `cfcResultDir` / `cfcInvocationContextTransport`: the prompt loop still ran docker `create/start/wait/rm` for the capability probe, persisted a capability snapshot, and completed when the model returned no tool calls. That means an enforcing run can still execute untrusted sandbox code without the trusted invocation-context/result sidecars this PR is trying to require.

Given the gVisor side silently treats missing CFC sidecar dirs/context files as absent, labs needs to fail before any sandbox execution in `enforce-explicit` / `enforce-strict`, including diagnostic probes, or else skip/replace those probes until the transports are wired. The current builtin-tool guard is good, but it needs to be applied before `collectHarnessCapabilitySnapshot()` (or `collectHarnessCapabilitySnapshot()` must receive/enforce the same precondition) to match the PR intent/spec fail-closed boundary.

Verification I ran: LSP diagnostics on the changed source files were clean; `deno test -A packages/cf-harness/test/docker-runsc-sandbox.test.ts packages/cf-harness/test/engine.test.ts packages/cf-harness/test/cli.test.ts packages/cf-harness/test/artifacts.test.ts` passed (`145 passed`).

mathpirate:
> Requesting changes for one fail-closed gap in the new transport floor.
> 
> `CfHarnessPromptLoop.runTranscript()` calls `engine.ensureDiagnosticsInitialized()` before any model turn/tool call, and `ensureDiagnosticsInitialized()` calls `collectHarnessCapabilitySnapshot()`, which executes `sandbox.runShell()` for the capability probes. The new transport guard is only in `CfHarnessEngine.invokeBuiltinTool()`, so it does not cover that earlier sandbox execution.
> 
> I reproduced this with an engine-owned docker-runsc runtime in `enforce-explicit` and no `cfcResultDir` / `cfcInvocationContextTransport`: the prompt loop still ran docker `create/start/wait/rm` for the capability probe, persisted a capability snapshot, and completed when the model returned no tool calls. That means an enforcing run can still execute untrusted sandbox code without the trusted invocation-context/result sidecars this PR is trying to require.
> 
> Given the gVisor side silently treats missing CFC sidecar dirs/context files as absent, labs needs to fail before any sandbox execution in `enforce-explicit` / `enforce-strict`, including diagnostic probes, or else skip/replace those probes until the transports are wired. The current builtin-tool guard is good, but it needs to be applied before `collectHarnessCapabilitySnapshot()` (or `collectHarnessCapabilitySnapshot()` must receive/enforce the same precondition) to match the PR intent/spec fail-closed boundary.
> 
> Verification I ran: LSP diagnostics on the changed source files were clean; `deno test -A packages/cf-harness/test/docker-runsc-sandbox.test.ts packages/cf-harness/test/engine.test.ts packages/cf-harness/test/cli.test.ts packages/cf-harness/test/artifacts.test.ts` passed (`145 passed`).

Good catch — confirmed, and fixed in 79771f397.

You're right that the floor only fired in invokeBuiltinTool(), while runTranscript() reaches sandbox.runShell() earlier through ensureDiagnosticsInitialized() → collectHarnessCapabilitySnapshot(). The probes are sandbox execution, not inspection, so they belong inside the fail-closed boundary. I went with your first option — apply the existing guard before the probes rather than skip/replace them — since a probe-less enforce run would just defer discovery of the mis-wiring to the first tool call.

One wrinkle worth flagging: diagnostics init wraps the probe in a try that converts errors into failure records and lets the run continue. A guard placed inside that block would have been swallowed into the run state rather than aborting. So #assertCfcTransportReady() now fires at the top of ensureDiagnosticsInitialized(), before the try. The assert in invokeBuiltinTool() stays as cheap defense in depth (it's idempotent), and injected runtimes remain exempt — they never set #ownedRunscConfig.

Regression test mirrors your repro: enforce-explicit, engine-owned docker-runsc sandbox, no transports — ensureDiagnosticsInitialized() rejects with the transport error and the fake process runner records zero docker invocations. Checked the other entry points too: --describe-capabilities prints static JSON without touching the engine, so it's unaffected.

Also updated the PR description: design-decision A had wrongly listed capability probes as construction-time inspection, which your repro disproved — it now states the invariant as "nothing executes in the sandbox under enforcement without the transports," with the floor at diagnostics init or first tool call, whichever comes first.

Full cf-harness suite: 409 passed. fmt/lint/check clean.

willkelly (approved):
Rereviewed the updated head `eb57e3a80` and the previous fail-closed gap is fixed.

The CFC transport floor now runs at the start of `ensureDiagnosticsInitialized()`, before the diagnostics try/catch and before `collectHarnessCapabilitySnapshot()` can execute sandbox capability probes. I reran the prior repro with an engine-owned docker-runsc runtime in `enforce-explicit` and no CFC sidecar dirs; it now rejects with the expected transport-floor error and performs zero docker lifecycle calls.

I also checked the resume/new-run CLI paths and subagent path shape. The injected-runtime exemption remains scoped to externally supplied runtimes, while engine-owned docker-runsc runs now fail before probes or tools when the sidecars are missing.

Verification:
- LSP diagnostics clean for changed cf-harness source/test files checked.
- Prior repro: rejected=true, dockerLifecycle=[]
- `deno test -A packages/cf-harness/test/docker-runsc-sandbox.test.ts packages/cf-harness/test/engine.test.ts packages/cf-harness/test/cli.test.ts packages/cf-harness/test/artifacts.test.ts` passed (`146 passed`).

PR #4019, opened 2026-06-10
fix(runner): materialize cross-space link targets on read (CT-1667)

bfollington:
Fixes CT-1667 (High): a value read through a **cross-space link** silently resolved `undefined` in any session that didn't create the data — the blank `cf-profile-badge`/`cf-render` profile fields, and the underlying cause of the picker's silent no-ops worked around in #3955.

## Root cause

Per-space server queries cannot follow links across space boundaries, so the target docs behind a cross-space link were never fetched — and the client had no recovery:

1. The schema traversal (`followPointer`) returned `notFound` for the absent doc **without triggering a load** (only `resolveLink`'s path-resolution kick existed, which shallow/`asCell` reads bypass).
2. Derived cell handles inherit `synced` from their parent — honest within a space (the originating server query pushed the same-space closure), structurally false across one.
3. `pull()` resolved at scheduler idle, before any async load could land.

**CFC was never the cause** — it has no read-side redaction; owner-protected and plain fields failed identically. The only things that worked were the explicit per-path `.pull()` workarounds (`wish.ts` `subscribeProfileName` / `#profileName`), which is exactly why the picker shows names while everything else showed blanks.

Why the suite never caught it: `StorageManager.emulate` in the usual single-manager shape means **shared replicas** — every read finds data some other runtime fetched. The new test runs **two storage managers loopback-connected to one shared in-process memory server** (real per-space replicas, the browser/CLI session shape) and reproduces the gap faithfully.

## Fix (three coordinated pieces)

- **traverse.ts** — `TraversalContext.onMissingLinkTarget`: `followPointer` reports a followed link whose target doc is entirely absent from the local replica, with the narrowed selector schema so the fetch covers the shape the read needs.
- **runtime.ts** — `ensureLinkedDocLoaded`: kicks an async load for the reported target, deduped per `(space, id)` (the sync leaves a live subscription behind, so one kick suffices and a later-created doc still arrives), registered as a cross-space promise so `storageManager.synced()` and pull's convergence can await it. Wired into `validateAndTransform`'s traversal context. Because the absent doc is a tracked read dependency, reactive consumers (sinks, the badge, computed) re-run on arrival with no further changes.
- **cell.ts** — `pull()` convergence: after idle, while kicked loads are pending, await them and re-idle (bounded at 10 rounds; each arrival can reveal the next hop). Pulls that kicked nothing take the zero-iteration path and keep their **exact previous timing** (`pendingCrossSpacePromiseCount`/`crossSpaceSettled` are new on `StorageManager`; `synced()` untouched).

## Verification

- New regression test `packages/runner/test/cross-space-value-read.test.ts` (two-session harness): through-link key-path `pull()`, deep whole-value `pull()`, and badge-style minimal-schema `sink()` — the deep-read case is red without the fix.
- Full `packages/runner` suite: **576 passed, 0 failed**.
- **Live against a local toolshed**, where all of these previously returned `undefined`:
  - through-link key read, deep `items` read, and minimal-schema child read of a cross-space piece → all materialize;
  - the real product path: deployed the actual `ProfileCreate`/`ProfileHome` stack, created a profile (its own `inSpace` space), `setName` via the owner flow, then read the **owner-protected `name`** through the home-space link with the badge schema → `"Ada Lovelace"`.

## Notes / scope

- The `wish.ts` per-path pulls are now redundant but left in place (harmless; removing them belongs in a follow-up cleanup, not a runner-semantics PR).
- The kick fires once per `(space, id)` per session for any linked-but-absent doc, including genuinely-deleted targets — one extra no-op query + subscription each; bounded by distinct absent links traversed.
- Pattern-layer consumers can now read cross-space values directly; the `Cell<>`-link convention for handler *state* (#3955) remains good practice for avoiding required-field validation on values that may not be loaded yet at event time.

Closes CT-1667.

---
## Summary by cubic
Fixes CT-1667: cross-space link reads now materialize missing target docs during read, so values no longer resolve to undefined. Scope limited to cross-space targets to avoid extra queries and keep pull performance stable.

- Bug Fixes
  - Traversal reports only cross-space missing link targets via `onMissingLinkTarget`, passing the narrowed selector and target-rooted path so the kicked watch is well-formed; same-space absences are not kicked.
  - Runtime dedupes a per-(space,id) load with `ensureLinkedDocLoaded`, tracks it as a cross-space promise for `synced()`/pull to await, and clears dedupe on failure; the pre-existing link-resolution kick now catch-swallows errors.
  - `Cell.pull()` converges: while cross-space loads are pending it waits (`crossSpaceSettled`) and re-idles with a bound and a warning on exhaustion; pulls that kick nothing keep prior timing.
  - Storage exposes `pendingCrossSpacePromiseCount()` and `crossSpaceSettled()`; regression test runs two storage managers against one shared in-memory server (and avoids double-closing). This restores correct data in `cf-profile-badge` and `cf-render`.
  - Docs: update roster guidance to note cross-space links now materialize for authorized viewers, and clarify `profile-create.tsx` uses a link `Cell<>` in handlers to avoid event-time validation depending on not-yet-loaded values.

<sup>Written for commit baec4b00b304ddfa472a04fd478085f25030fb40. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4019?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/shared-profile-rosters.md
+++ b/docs/specs/shared-profile-rosters.md
@@ -76,29 +76,24 @@ Store a reference to the contributor's profile cell (the `result` of their
 `cf-cell-link` / `cf-render`):
 
 - **Pros:** Always current — a profile rename propagates to every roster that
-  links it.
-- **Cons:** Resolving the link requires reaching the *other* user's profile
-  space. That is exactly the cross-space reachability the runtime does **not**
-  guarantee for arbitrary users; it works cleanly for the current viewer's own
-  profile but is not guaranteed for a roster full of other people's profile
-  links. It also couples the roster's render to remote spaces' availability.
+  links it. The runtime materializes cross-space link targets on read (a read
+  that finds the target absent kicks an async load and re-renders on arrival —
+  CT-1667, PR #4019), so the link resolves for any viewer with access to the
+  profile's space.
+- **Cons:** First render shows blanks until the cross-space load lands, and
+  the roster's freshness couples to the remote space's availability — offline
+  or unreachable profile spaces render empty rather than stale.
 
 ### Verdict
 
 **Default to a snapshot.** It matches the "each user contributes their own data"
-model, keeps the shared roster fully self-describing inside the space, and avoids
-depending on other users' profile spaces being reachable. Reach for a live link
-only when (a) freshness genuinely matters and (b) you have confirmed the linked
-profiles resolve for *all* viewers in your deployment — and even then, consider a
-**hybrid**: store the snapshot for reliable rendering *and* keep the link
-alongside it for an explicit "refresh from profile" action.
-
-> Uncertainty flagged: persisting another user's live profile cell into a
-> `PerSpace` array and having *every* viewer dereference it across space
-> boundaries is not something this spec verified end to end. The current
-> viewer's own `#profile` result is known to render (see
-> `packages/patterns/shared-profile-demo/main.tsx`). Treat the live-link roster
-> as the advanced path and validate cross-space resolution before relying on it.
+model, keeps the shared roster fully self-describing inside the space, and
+renders identically (and immediately) for all viewers regardless of remote
+space availability. Reach for a live link when freshness genuinely matters —
+it resolves cross-space for any authorized viewer — and consider a **hybrid**:
+store the snapshot for immediate, availability-independent rendering *and*
+keep the link alongside it for live freshness or an explicit "refresh from
+profile" action.
 
 ## Reference demo
 
@@ -367,9 +362,10 @@ const profileWish = wish({ query: "#profile" }); // result is ProfileHomeOutput-
 // Render an entry with: <cf-cell-link $cell={entry.profile} /> or <cf-render $cell={entry.profile} />.
 ```
 
-Validate cross-space resolution for *all* viewers before shipping this — see the
-uncertainty note above. A hybrid (snapshot for rendering + link for an explicit
-refresh) is usually the safer way to get freshness.
+Cross-space resolution works for any viewer with access to the profile's
+space (CT-1667, PR #4019); the trade-off is availability — entries render
+blank while the remote space is unreachable. A hybrid (snapshot for immediate
+rendering + link for freshness or an explicit refresh) gives both.
 
 ## Recommended approach for chat/game patterns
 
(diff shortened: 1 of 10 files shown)


bfollington:

A cf-review pass (report-only) found **no blockers** and 2 should-fixes; all actioned in the follow-up commit (the review also independently confirmed the regression guard red-on-main/green-on-branch and ran the adjacent scheduler-pull / pull-reactivity / transaction-notfound / data-uri-sync suites green):

1. **Kicked-sync (path, schema) coherence** — the narrowed selector schema describes the value at the selector's own (target-rooted) path, but the kick paired it with the bare link path, registering a malformed watch selector (harmless today only because fetches are doc-granular). Fixed: the kick now uses `selector.path.slice(1)` (minus the `"value"` prefix) when a selector exists.
2. **`pull()` convergence on the manager-global pool** — consciously accepted (same semantics as `synced()`; converging on a concurrent reader's loads is harmless and the loop exits as soon as the pool drains), now stated in a comment; plus a `warn` log when the 10-round bound exhausts with loads still pending, so a pathological chain is observable instead of silent.

Nits taken: `.catch()` on the pre-existing `link-resolution.ts` kick (its rejection could previously escape the resolution path); the test harness no longer lets each manager's inherited `close()` close the *shared* server (the subclass serves the server without initializing the base class's private field); a comment marks which test step is the real red-on-main regression guard (the deep whole-value pull) versus the contract-pinning steps.

Follow-ups deferred to the ticket (not this PR): removing the now-redundant `wish.ts` per-path workaround pulls; updating `docs/specs/shared-profile-rosters.md`'s cross-space uncertainty note.

Re-verified after the fixes: regression test green, full runner suite green, and live against a local toolshed the deep cross-space read of the real ProfileHome returns both the plain and the **owner-protected** field through the home-side link: `[{"initialNameApplied":"Ada Live","name":"Ada Lovelace"}]`.

bfollington:
## Re-scoped per discussion: cross-space gate + docs

The perf check caught a real cost of the un-gated design: kicking a load for **every** absent link target made each read of a legitimately-absent same-space value (optional fields, holes — extremely common in pattern code) a server query plus a pull-convergence round. Measured: pattern-unit-4 +29.7% (OVER), pattern-unit-1 +41.8%, individual pattern tests +14–21%.

Changes in the follow-up commit:

- **`followPointer` now reports a missing target only when it lives in a different space than the source doc.** CT-1667 is strictly a space-boundary problem — same-space targets are either covered by the originating schema query or genuinely absent. Locally, the previously-regressed pattern tests now match main (lunch-poll 6.87s branch vs 6.93s main), and the live deep cross-space read of a real ProfileHome (owner-protected `name` included) still materializes.
- **Comments trimmed to constraint statements** (what must hold, not the change history).
- **Stale docs updated**: `docs/specs/shared-profile-rosters.md` no longer claims cross-space link resolution is unavailable/unvalidated (live-link rosters are now viable for any authorized viewer; snapshot default re-justified on immediacy + availability-independence); `profile-create.tsx`'s `Cell<>` handler-state comments re-explained by their real remaining rationale (event-time validation) instead of the now-fixed read gap.

Runner suite 576/0 green after the re-scope. Waiting on CI — including a green Performance Check this time — before merge.

PR #4038, opened 2026-06-11
fix(cfc+patterns): profile element writes — stamp reconciliation, same-tx link sources, single authorized writer (CT-1698)

bfollington:
Fixes [CT-1698](https://linear.app/common-tools/issue/CT-1698): profile cards / link elements could not be added — both "Add profile card" and "Add URL element" silently did nothing (console: `writeAuthorizedBy must remain stable at /elements`). This is **#3880's silently-aborting element save in its modern form**: the newer CFC machinery (#3909/#4008) turned the silent abort into loud, diagnosable errors, and behind the first error were two more. Also lands the [CT-1651](https://linear.app/common-tools/issue/CT-1651) copy rename (elements are references, not deployed pieces).

## Three stacked causes, three contained fixes

1. **`cfc/schema-merge.ts` — stamp reconciliation predates `moduleIdentity`.** `reconcileWriterClaimStamp` stripped only `bundleId` before comparing a stamped claim with an authored one, but new claims are stamped with **both** `bundleId` and `moduleIdentity` (`implementation-identity.ts`) — so every stamped-vs-authored merge of the *same binding* threw "must remain stable". The provenance stamp is now `{bundleId, moduleIdentity}`, stripped together; two different stamps still conflict.

2. **`cfc/prepare.ts` — link sources minted by the same transaction failed closed** ("missing link source metadata at /elements/0" — the exact re-enable condition documented on the skipped profile-home tests). Two new knowability hatches in `derivePersistedLinkLabel`:
   - the source piece's setup-written `["schema"]` meta (read-your-writes for a same-tx pattern instantiation, stored for an earlier one);
   - for child docs the data layer splits out of an inline value (array/object entries), the **target's own candidate schema at the target path** — the label the value would have carried inline — **gated to docs this transaction actually wrote**. A pre-existing unlabeled doc still fails closed; `cfc-boundary` now pins **both** cases (the same-tx case additionally asserts the persisted link label carries the location's confidentiality, so this is label-preserving, not label-skipping).

3. **`patterns/system/profile-home.tsx` — five writers, one claim slot.** A `writeAuthorizedBy` claim carries exactly one handler binding (transformer extracts a single `typeof identifier`; prepare verifies the writing handler's identity against it; only claim-referenced handlers get `__cfBindVerifiedBinding` annotations). But `elements` was written by **five** handlers — the declared `addElement`, plus `removeElement` and three private UI handlers, all unannotated and unauthorized. All element mutations are now instances of the single authorized **`mutateElements`** implementation (instances differ only in bound state, which doesn't change the implementation identity): exported add/remove streams, the catalog button, the link form, and per-row remove. The exported stream fields widen to the mutation event type; `AddProfileElementEvent`/`RemoveProfileElementEvent` remain the documented subsets.

## CT-1651 rename (recorded design verdict: references, not deployed pieces)

"Pattern URL" → **"Link URL"**, "Add URL element" → **"Add link"**, plus helper text *"Links are saved as reference cards. They are not deployed or run."*

## Verification

- The profile-home element tests are **un-skipped** (their skip comment described exactly this bug chain) and pass: add catalog element → assert → remove → assert, under CFC enforcement.
- New regression test for the stamp reconciliation (bundleId+moduleIdentity vs authored); `cfc-boundary` re-pins fail-closed for pre-existing unlabeled sources and label-derivation for same-tx sources; the CT-1665 binding-metadata test updated to the new symbol.
- Full runner suite: **584 passed, 0 failed**. Pattern tests matching "profile": green.
- **Live against a local toolshed** (worktree CLI): deployed the real ProfileCreate stack, created a profile, then `addElement` with both flavors via `cf piece call` — `elements` reads back `[{source:"catalog",tag:"profile-card",title:"Profile card"},{source:"url",tag:"note",title:"My Notes"}]`, where every variant previously failed.

## Notes

- A CFC write rejection still surfaces only as a console error (the button click looks like a no-op) — UI surfacing is noted on CT-1698 as a separate follow-up.
- Browser-level re-verify of the element buttons needs this merged + dev servers restarted (the shell runtime needs the two CFC fixes); planned as the post-merge check.

Closes CT-1698. Closes CT-1651.

---
## Summary by cubic
Fixes CT-1698 so profile cards and link elements save correctly under CFC, and updates copy per CT-1651. Reconciles writer-claim stamps, labels same-transaction link sources, and routes all element mutations through one authorized writer with mode-bound dispatch.

- **Bug Fixes**
  - CFC stamp reconciliation: `reconcileWriterClaimStamp` now strips both `bundleId` and `moduleIdentity` before comparing, merging stamped vs authored claims for the same binding; conflicting stamps still reject.
  - Same-transaction link sources: `derivePersistedLinkLabel` now derives labels from the source piece’s setup `["schema"]` meta, or from the target path for child docs written and created in the same tx; pre-existing unlabeled sources still fail closed.
  - Mode-bound element mutations: `mutateElements` uses explicit modes (`add`, `addCard`, `addLink`, `remove`) so empty remove events and blank link forms no longer add elements by mistake; tests pin the no-op remove.

- **Refactors**
  - Single authorized writer: replaced multiple element handlers with one `mutateElements` implementation; exported `addElement`/`removeElement` are instances bound with `mode`, and the catalog button, link form, and per-row remove all call this single writer.
  - CT-1651 copy: “Pattern URL” → “Link URL”, “Add URL element” → “Add link”, plus helper text clarifying links are reference cards, not deployed.

<sup>Written for commit f9e6041e2569cbdcbbf86da5d5a5d8880ad8e1fb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4038?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/patterns/system/profile-home.test.tsx
+++ b/packages/patterns/system/profile-home.test.tsx
@@ -56,19 +56,15 @@ export default pattern(() => {
       { assertion: assert_name_set },
       { action: action_clear_name },
       { assertion: assert_name_cleared },
-      // Skipped under CFC enforcement: addElement instantiates a fresh
-      // ProfileCatalogCard inside the handler and links it into the
-      // writeAuthorizedBy-protected `elements` list in the same transaction.
-      // The new instance's doc has no stored CFC metadata and no pending
-      // schema input yet, so the fail-closed link-write rule rejects the
-      // commit ("missing link source metadata ... at /elements/0").
-      // Re-enable once the runtime counts a same-transaction pattern
-      // instantiation as pending source metadata (or the pattern
-      // pre-materializes elements).
-      { action: action_add_catalog_element, skip: true },
-      { assertion: assert_added_element, skip: true },
-      { action: action_remove_catalog_element, skip: true },
-      { assertion: assert_removed_element, skip: true },
+      // Add/remove exercise the full owner-protected element write stack:
+      // the same-transaction card instantiation linked into the
+      // writeAuthorizedBy-protected list (prepare's setup-schema / child-doc
+      // link-label hatches) and the single authorized `mutateElements`
+      // writer behind every mutation surface (CT-1698).
+      { action: action_add_catalog_element },
+      { assertion: assert_added_element },
+      { action: action_remove_catalog_element },
+      { assertion: assert_removed_element },
     ],
   };
 });
--- a/packages/patterns/system/profile-home.test.tsx
+++ b/packages/patterns/system/profile-home.test.tsx
@@ -20,6 +20,14 @@ export default pattern(() => {
     }
   });
 
+  // A remove event without a cell must be a no-op — the consolidated
+  // mutateElements writer dispatches on the instance's bound mode, so a
+  // malformed remove must never fall through to an add (cf-review on
+  // CT-1698).
+  const action_remove_with_empty_event = action(() => {
+    profile.removeElement.send({});
+  });
+
   const action_clear_name = action(() => {
     profile.setName.send({ name: "" });
   });
@@ -64,6 +72,8 @@ export default pattern(() => {
       { action: action_add_catalog_element },
       { assertion: assert_added_element },
       { action: action_remove_catalog_element },
+      { assertion: assert_removed_element },
+      { action: action_remove_with_empty_event },
       { assertion: assert_removed_element },
     ],
   };
--- a/packages/patterns/system/profile-home.tsx
+++ b/packages/patterns/system/profile-home.tsx
@@ -57,6 +57,22 @@ export type RemoveProfileElementEvent = {
   cell: any;
 };
 
+/**
+ * The single event shape for every element mutation (see `mutateElements`):
+ * a `cell` (from the event or the instance's bound state) removes that
+ * element; a `patternUrl` (or a bound link form) adds a link reference card;
+ * otherwise a catalog card is added. `AddProfileElementEvent` /
+ * `RemoveProfileElementEvent` are subsets kept for the exported stream types.
+ */
+export type MutateProfileElementsEvent = {
+  catalogId?: string;
+  patternUrl?: string;
+  title?: string;
+  tag?: string;
+  userTags?: readonly string[];
+  cell?: any;
+};
+
 export type SetProfileNameEvent = {
   name?: string;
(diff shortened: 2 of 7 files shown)


bfollington:

A cf-review pass (report-only) on the first commit found **1 blocker, 2 should-fixes**; all actioned in the follow-up commit (the review also independently confirmed the security posture of both CFC hatches — no laundering vector: stored CFC metadata always merges by union and its presence disables the hatch; a forged `["schema"]` meta can only *raise* labels, granting nothing beyond an authored pending schema input — and verified the cfc-boundary re-pin is a legitimate re-statement, not a bent test):

1. **Blocker — shape-dispatch fallthrough in `mutateElements`:** clicking "Add link" with an empty URL added a spurious catalog card, and `removeElement.send({})` *added* an element instead of no-opping. Fixed with the reviewer's suggested design: each binding site declares an explicit **`mode`** in bound state (`"add" | "addCard" | "addLink" | "remove"`) — bound state doesn't change the implementation identity, so the single `writeAuthorizedBy` claim still covers every instance, but a malformed/empty event can no longer cross purposes. Each mode preserves its predecessor handler's exact semantics (including the exported stream's historical defaults, which the first commit had drifted). New test step: an empty remove event is a no-op.
2. **Should-fix — same-tx gate looseness:** the prepare.ts child-doc hatch accepted any write to any path of the source doc. Now requires a root-level write with `previousValue === undefined` — i.e. the doc was **created** by this transaction, matching the comment's claim.
3. **Should-fix — capability fusion:** subsumed by the mode fix (the add stream can no longer remove and vice versa).

Nits taken: the stamp-reconciliation comment no longer cites the retired `schemasEqualIgnoringWriterBundleIds` helper or describes the merge as "precedence".

Re-verified after the fixes: profile-home element tests green (now 6 steps incl. the no-op case), all four CFC suites green, fmt/lint clean, full runner suite re-run green.

PR #14, opened 2024-05-29
App skeleton

gordonbrander:
This PR begins prototyping a wireframe skeleton for an app with the recursive thread UI we've been developing. The skeleton components are defined using Lit.

PR #4039, opened 2026-06-11
docs: Pattern Primitives — factoring the corpus into composable features

jkomoros:
## What

A design doc (`docs/features/PATTERN_PRIMITIVES.md`) proposing which features should be factored out of today's chonky patterns into small reusable primitive patterns, and how the existing patterns get reimplemented as compositions. Companion to `docs/features/CANONICAL_BASE_PATTERNS.md` (which covers the data-type axis; this covers the functionality axis).

cc @bfollington — this should dovetail with the factoring you've already started; the §6 census and §8 first wave are the parts most worth pressure-testing against what you've found in practice.

## Highlights

- **Census of ~27 primitive candidates** dug out of the corpus via duplication evidence (each appears inline in 2–19 places), provenance-tagged so generator-stamped copies aren't mistaken for independent demand. Strongest: PiecePicker (6 independent inline copies), auth consolidation (3 competing idioms), the gmail decomposition (fetch / labeled row-store / LLM-extract / status).
- **Per-tier CFC argument**: taint propagates per transaction, so small handlers + small schemas are the taint-minimization mechanism — and authoring-time sub-patterns get this without separate pieces. (The big Google importers contain a dead `type Confidential<T> = T;` stub — authors wanted labels and the chonky shape gave them nowhere to put them.)
- **Reimplementation case studies** for the prime specimens (gmail-importer, parking-coordinator + lot-watch, group-chat → cfc-group-chat-demo shape, dataset-heavy field modules, self-improving-classifier).
- **Runtime/tooling asks** ranked by priority — #1 is retargeting `tools/generate-importer.ts` to emit compositions instead of fresh monoliths.
- **Process, not project**: primitive archeology as a recurring loop with per-tier promotion bars (extending the cfc/README bar), a kill criterion, and a first wave of four. TokenBroker's full CFC form is explicitly deferred behind verified-code identity + the cross-space blockers (CT-1188/CT-1105/CT-1090).

Grounded in the strategy corpus (card-web refs cited inline, resolvable via the card_web connector) and hardened against two adversarial review passes (strategic + line-by-line fact-check; all headline line counts and CFC mechanics verified against the tree).

---
## Summary by cubic
Adds a design doc that proposes factoring our large patterns into small, reusable primitives and recomposing existing patterns. Sets a first-wave plan and prioritizes retargeting `tools/generate-importer.ts` to generate compositions, improving reuse and CFC safety.

- **Refactors**
  - Adds `docs/features/PATTERN_PRIMITIVES.md` with a 27-primitive census (with provenance), per-tier CFC rationale, and a process for promotion/migration.
  - Re-maps key patterns (Gmail/Calendar importers, parking-coordinator + lot-watch, group chat, self-improving-classifier) into composable primitives with small first-wave targets (e.g., PiecePicker, dataset splits, chat consolidation).
  - Calls out runtime/tooling priorities: retarget `tools/generate-importer.ts`, adopt labeled row/column stores, establish shared schema modules, and set parity + migration policies.

<sup>Written for commit 680ff88a267bc0bc79fc2745bf6946b3a52029f5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4039?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/features/PATTERN_PRIMITIVES.md
+++ b/docs/features/PATTERN_PRIMITIVES.md
@@ -1,0 +1,783 @@
+# Pattern Primitives: Factoring the Corpus into Composable Features
+
+**Status:** Draft for discussion (v2 — hardened against two adversarial reviews)
+**Date:** 2026-06-10
+**Author:** Claude (research: six parallel survey agents over packages/patterns, the runtime, CFC specs, loom-files strategy docs, and the card-web corpus; fact-checked + devil's-advocated by two further agents)
+**Companion doc:** `docs/features/CANONICAL_BASE_PATTERNS.md` covers the *data-type* axis (schelling-point interfaces, fork-on-demand, containers). This doc covers the complementary *functionality* axis: which features should be factored out of today's chonky patterns into small reusable primitive patterns, and how the existing patterns get reimplemented on top of them.
+
+---
+
+## 1. Executive summary
+
+The pattern corpus has two populations. One is a thriving family of genuine
+primitives — 27 field modules behind `container-protocol.ts`, 17 CFC trusted
+surfaces, the `suggestable/` LLM widgets — each small, single-purpose, and
+recombinable. The other is a belt of single-file applications (2,000–3,000
+lines each: `parking-coordinator`, `lot-watch`, `gmail-agentic-search`,
+`self-improving-classifier`, `lunch-poll`, the ~17 Google extractors) that
+re-implement the same dozen features inline, over and over.
+
+The survey found **~27 primitive candidates** (§6), each appearing in 2–19
+places today. The catalog is a census, not a commitment list — §8 proposes a
+first wave of four and a process for the rest.
+
+Why factor, in order of how little each argument depends on the future:
+
+1. **Banked correctness.** Most of this corpus is LLM-authored, and
+   regeneration is cheap — but regenerated code re-steps on known landmines
+   (the reactive-scoping hazard that *forces* monolithic authoring, §3.2;
+   the name-based-identity bug shipped in `profile-group-chat`, §5.4). A
+   primitive is where a fix lands once. §2.6 confronts the LLM-economics
+   objection directly.
+2. **The CFC audit economy.** The taint quantum is the transaction, and every
+   write of a transaction inherits the join of everything it read (S16, D1).
+   Small handlers, small schemas, and stable code identities are what labels,
+   trusted-write contracts, and (eventually) one-time vetting attach to.
+   "Taint containment effectiveness degrades super-linearly as black box
+   size increases" (card.web:c-443-fce753). §4 states this per-tier, because
+   the payoff differs by kind of primitive.
+3. **Recombination odds.** A pattern bundling ten features needs all ten to
+   fit a new use case; ten one-feature patterns can be re-glued in any
+   subset — combinatorial coverage of the adjacent possible
+   (card.web:c-863-eac691).
+4. **The ranking flywheel — the horizon, not the dependency.** Small safe
+   patterns are the unit the savvy-user → everyone pipeline will eventually
+   distribute: "the image onebox technique, but for turing complete things"
+   (card.web:c-421-afc285). None of the ranking machinery exists yet (wish()
+   is unranked, there is no usage telemetry, and today's authors are us and
+   our LLMs). This work is justified by 1–3 even if the flywheel never
+   spins; it also happens to be the only shape the flywheel *can* distribute
+   when it does — nobody reuses a bespoke whole; everybody reuses blocks
+   (card.web:c-906-ada936).
+
+The method is **primitive archeology** (card.web:c-840-ece996): don't design
+primitives top-down — look at the chonky patterns that already exist (most
+born in `factory-outputs/` or from `tools/generate-importer.ts`) and
+"explain" them as compositions of primitives that, had they existed, would
+have produced the same outcome. One honesty requirement the census must meet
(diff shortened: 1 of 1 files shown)


bfollington:
Dovetailing from the CT-1715 reuse overhaul (the factoring this doc cc's) — read with the §6 census and §8 first wave in mind as suggested. It holds up well against what we found in practice; notes below on what's already landed, what we're adopting from this doc, and one place the two efforts currently point in different directions.

**Already landed (this week) that intersects the census:**
- Status tiers in `packages/patterns/index.md` (exemplar/demo/fixture/legacy) — the "which patterns carry authority" problem §6 provenance-tags its way around.
- Advisory critic rules (`pattern-critique-guide.md` §15) targeting the adoption-failure class: hardcoded hex/typography, `.set()`-only handlers, hand-rolled Enter-submit, index-based selection, hand-rolled badge/field/empty-state markup. This is pressure toward #13 (FormBuffer: "mostly a generator/skill-docs fix") and #16 (StatusBadge: "likely a generator-docs fix") without new surface.
- `cf-field` + `cf-empty-state` components + four exemplar pattern migrations + a 366-ref dead-token sweep — partially services the §6.C interaction tier from the component side.

**Adopting from this doc into the CT-1715 phase plan:**
- The §8.2 promotion bar (two migrated callers from different families) + kill criterion → now acceptance criteria on CT-1710/CT-1712 (EditableList ≈ #20 ListCRUD, in flight next).
- #20's identity constraint verbatim: stable identity (ids/`equals()`-cells) internally, title-addressed handlers as an explicit agent-facing fuzzy layer — agreed this must be settled before the primitive becomes canon.
- ConfirmAction (CT-1711 ≈ #14) will keep the (s)→(p) trusted-surface promotion path open.
- Chat: we had ChatThread parked in a reassessment issue (CT-1714) for churn reasons — dropping it there in favor of first-wave item 4 here. ProfileIdentity/JoinRoster (#8) likewise yours.
- Directory vocabulary: new composable sub-patterns land in `packages/patterns/primitives/`, named to match this doc.

**One tension to reconcile:** index.md now tiers `record/` + `container-protocol.ts` as legacy ("registry/MODULE_METADATA approach is a parallel composition system; do not copy"), while #19/#21/#22 treat that machinery as clay for PieceCollection/TrashRestore/RegistryBackup, growing the implemented container protocol toward the documented `ContainerProtocol<T>`. I think both are right at different granularities — retire the *registry idiom* (stringly-typed module registration, the "update THREE places" problem), salvage the *collection machinery* (create-child/push-handle/navigateTo/bulk ops) into a primitive with a real contract. If that framing works for you I'll adjust the index.md legacy note to say exactly that, so the tier doc and this doc don't read as contradicting each other. Related nuance: do-list is tiered exemplar but carries the #20 title-as-identity bug class — the EditableList migration (CT-1712) is where that gets resolved.

(Drafted with Claude Code during the CT-1715 working session.)

bfollington:
Evidence update on **#20 ListCRUD** from the CT-1715 workstream: we built it, proved it both ways against real callers, and **invoked the kill criterion** — the §8 loop worked exactly as designed, and cheaply.

What happened: `EditableList` (reference-addressed add/remove/update/toggle streams + counts + default checkbox-row UI) was built to a documented composition contract, hardened through three review cycles (equals()-identity instead of minted ids; in-place element-cell updates after a reviewer caught slot-replacement re-minting entity identity), then tested:

- **Headless embeds grew both pilot callers** (+22/+9 non-comment LOC on simple-list/do-list): with `equals()`-idiom handlers at 3–5 lines, the shared logic is smaller than the adoption glue. The duplication the census counts is real but too cheap to extract as a sub-pattern.
- **Rendered path: zero compatible callers** after a systematic survey — every checklist keys text off `title`/`text`, and behind the key sits per-caller personality (todo-list's archive split, shopping-list's card rows/view modes, project-list's named-add buttons) that a generic default UI can't reach behavior parity with.

Retirement PR: #4086. What survived is the durable part: the composition contract doc (`docs/common/patterns/primitives.md`, now carrying an **adopter-first entry bar** + the lessons), and the identity rules the build forced us to get right — those just hardened 8 real patterns in #4085 (slot-replacement updates orphaning held references: do-list, store-mapper, fair-share, shopping-list, habit-tracker, project-list, budget-tracker, map-demo).

Two takeaways for the census/process:

1. **Chunk-size criterion** worth adding to §8's promotion bar: the reusable unit must be *bigger than the glue needed to adopt it* and *smaller than the per-caller personality it sits inside*. ListCRUD failed the first bound; generic row UIs fail the second. Suggests checking the (m) pure-module tier before reaching for (s) — for list machinery, the thing that actually paid was idiom-level (docs + critic rules + a mechanical sweep), not library surface.
2. The "duplication evidence" for #20 (6+ checkbox-list implementations) was accurate and still extraction-negative — provenance tagging alone doesn't catch this; only the reimplement-don't-just-extract step (§8.3) did. Strong argument for keeping that step mandatory.

ConfirmAction (#14 ≈ CT-1711) and MasterDetail (CT-1713) remain queued on our side but now gated on a named real adopter before any code.

(Drafted with Claude Code during the CT-1715 working session.)

PR #4048, opened 2026-06-11
feat(patterns): EditableList composable primitive (CT-1710)

bfollington:
## Summary

The first composable higher-order **primitive** — a pattern used as a JSX tag inside other patterns. It sets the composition-contract precedent for the CT-1715 reuse overhaul. Lives at `packages/patterns/primitives/editable-list.tsx`, the first occupant of a new `primitives/` tier.

## The contract this establishes

- **Headless-able.** Render its `[UI]` to get a default row experience (per-item: checkbox + editable text + delete; a `cf-message-input` adder; `cf-empty-state` when empty). Or ignore `[UI]`, embed for the model, and `.map()` your own rows while driving the exposed streams/cells. **No VNode/render-prop input** — render props fight the CTS transformer; the headless path achieves full custom rendering via cells + streams instead.
- **Identity, not index, not title.** Core mutations (`removeItem`/`updateItem`/`toggleItem`) address items by a stable `id`, never an array index — index-based mutation is the fragility this overhaul kills. `addItem` mints an id if omitted; `updateItem` refuses to overwrite `id`.
- **Title/text addressing is a separate, explicit convenience layer** (`addItemByText`/`updateItemByText`/`removeItemByText`) — fuzzy, case-insensitive, first-match — layered on top of the identity core for agent-driveability. Documented as convenience, not the identity model.
- **Home + tier.** New `primitives/` dir (plain subdir of `@commonfabric/patterns`, no workspace/deno.json change needed); new `primitive` tier added to `packages/patterns/index.md` legend + listing.

## Resolved: can a sub-pattern mutate a parent-owned cell? **YES.**

When a parent does `EditableList({ items: myItems })`, the primitive receives the **same** `Writable<T[]>` cell, not a copy; `.push()`/`.set()` from its handlers mutate the parent's cell and sync back. Evidence:
- `docs/common/concepts/reactivity.md` — `Writable<>` is *write intent* on a shared reactive cell.
- `docs/common/patterns/composition.md` — "Both patterns receive the same `items` cell — changes sync automatically."
- `packages/patterns/examples/cf-render.tsx` — embedded `Counter` mutates the parent's `state.value`.

So the primitive defines + binds its own handlers to the cell it's given; the parent wires nothing by hand. (Documented escape hatch for the cross-space owner-protected case where in-place mutation is impossible — not needed here.)

## Item-shape decision

[...]

PR #4049, opened 2026-06-11
refactor(patterns): migrate simple-list + do-list onto EditableList (CT-1712)

bfollington:
Migrates **simple-list** and **do-list** onto the `EditableList` primitive (CT-1710, #4048) — the promotion-bar proof that the composition contract serves two *different* pattern families.

## Per-caller approach

Both ended up **headless**, not rendered — the key result (see contract friction below). Neither could use EditableList's default `[UI]` because each has row chrome the generic row doesn't model.

**simple-list (headless)** — embeds EditableList for `removeItem`-by-id + `total`/`done` counts. Keeps its own rows because of the indent gutter, the Cmd+[ / Cmd+] keyboard-indent handlers, and the indent-toggle button. Its indent (`toggleIndent`/`setIndent`) and the `text`-shaped `addItem` become thin **id-keyed** handlers bound to the *same* shared `items` cell the primitive mutates. Items gain a stable `id` (+ a `label` carried for the primitive's shape); `text`/`indented` ride along as plain-data extras. External streams switch from index- to id-addressed. **MODULE_METADATA schema is unchanged** (still `text`/`indented`/`done`), so persisted Record data stays compatible and the `record/registry.ts` consumer still type-checks.

**do-list (headless)** — embeds EditableList for id minting on add, the id-addressed `updateItem`, the counts, and `clearDone` (which `archiveCompleted` now maps onto). Keeps its own rows (Suggestion sub-pattern, cell-link attachments, drop-zones, indent, completed drawer) and `.map()`s them. Identity switches from `equals(whole-object)` to `{ id }`. The title-addressed `removeItemByTitle`/`updateItemByTitle` stay as the **agent-facing convenience layer** over the id model (kept because `default-app-ben.tsx` wires them as omnibox tools). Adds a behavior-parity test (do-list had none).

## attachments resolution

`attachments: Writable<any>[]` is a **live-cell** field. Per the contract, the primitive's index-signature passthrough carries extras as **plain data only** — a live cell read back through it is not re-hydrated ("any → true schema" gotcha). So attachments stay a **declared typed field on do-list's own `DoItem`**, and do-list renders + mutates them itself (`addAttachment`/`removeAttachment` via `items.key(idx).key("attachments")`). This is precisely *why* do-list is a headless embed and not a default-UI caller. Resolution: **attachments do not move onto EditableList; the primitive is not changed to carry them.**

## LOC deltas

| file | before | after | non-comment before → after |
|---|---|---|---|
| simple-list.tsx | 245 | 294 | 213 → 235 (**+22**) |
| do-list.tsx | 516 | 575 | 451 → 460 (**+9**) |

**Both callers GREW, not shrank** — the contract's "reduce caller code" promise did not hold for these two. See friction.

## Behavior-parity evidence

```
=== simple-list check ===  EXIT: 0
=== do-list check ===      EXIT: 0
=== simple-list test ===   20 passed, 0 failed
=== do-list test ===       18 passed, 0 failed
deno lint (4 files)        Checked 4 files / EXIT 0
deno fmt --check (4 files) Checked 4 files / EXIT 0
default-app-ben.tsx check  EXIT: 0   (do-list API consumer)
record/registry.ts check   EXIT: 0   (simple-list consumer)
```

do-list tests cover: add (with indent), addItems batch, unique-id minting, update-by-id (title/done), cascade remove-by-id (item + indent children, sibling survives), title-addressed convenience (rename/done/cascade-remove), archiveCompleted.

## Contract friction (the key deliverable)

1. **Neither caller is the "clean rendered case" the contract anticipated.** Both have row chrome the default row can't express (indent + keyboard for simple-list; Suggestion/attachments/drop-zone for do-list), so both went headless. The promotion bar implicitly assumed at least one default-UI caller; these two families don't supply one. The default `[UI]` is real but unexercised by the two canonical migrations.

2. **The primitive reduced caller LOC *negatively*.** For a headless caller the primitive contributes id-minting + counts + a couple of id streams — but the caller still writes its own handlers for everything the primitive doesn't model (indent, cascade-delete, rich add), *plus* the embed line, *plus* the new `id`/`label` item fields, *plus* (for simple-list) a `{text}`→push adapter. Net result: more code, not less. The contract's "REDUCE caller code" claim should be qualified to *rendered* callers; headless callers may not benefit on LOC.

3. **id-minting is duplicated.** Both callers needed to mint an `id` for their own add/adapter handlers, and re-implemented EditableList's exact scheme (`safeDateNow().toString(36)`-`nonPrivateRandom()`) inline because (a) the primitive doesn't export a `mintId` helper and (b) "no function creation in pattern context" blocks a local one. **Candidate primitive change:** export `mintId` (or accept add-with-extras through the primitive's own `addItem` so callers never mint). I did NOT change the primitive — flagging as a CT-1710 follow-up.

4. **`addItem` doesn't fit extras-carrying callers cleanly.** EditableList's `addItem({label, item})` always writes `label` + `done` and is built around `label`. do-list/simple-list key off `title`/`text`, so they route adds through their *own* handler instead of the primitive's, just to set their own fields and mint the id — meaning the primitive's add path is bypassed by both real callers. A headless caller arguably wants an `addItem` that takes a full `item` and only mints the id. Worth considering for the primitive.

5. **Cascade semantics live only in the caller.** do-list's remove deletes an item *and its indent-children*; the primitive's `removeItem` is single-item. This is correctly caller territory, but it means the primitive's `removeItem` is unused by do-list — the primitive served the *model/identity*, not the mutation, for the more complex family.

Nothing here was papered over by weakening the primitive or hacking a caller; the primitive is untouched.

---
## Summary by cubic
Migrates `simple-list` and `do-list` onto `EditableList` (Linear: CT-1712), keeping their custom UIs while adopting an id-keyed model, shared counts, and clearer item identity. Improves reliability (no index-based mutations) and adds tests for behavior parity.

- **Refactors**
  - `simple-list`: headless embed of `EditableList` for remove-by-id and counts; items gain `id` + `label`; keeps indent gutter and Cmd+[ / Cmd+] via id-keyed `toggleIndent`/`setIndent`; external streams switch from index to id; schema unchanged.
  - `do-list`: headless embed for id minting, id-based `updateItem`, and counts; maps `archiveCompleted` to `clearDone`; keeps cascade delete (item + indent-children) and rich add (indent/attachments/aiEnabled); attachments remain a live-cell field on `DoItem`; retains title-based convenience handlers over the id model.
  - Tests: added comprehensive `do-list` parity tests; updated `simple-list` tests to use id-addressed actions.

- **Migration**
  - No UI changes; persisted `simple-list` data stays compatible. `do-list` adds `id`/`label` internally without breaking callers.
  - Update any custom integrations to use id-based handlers (`removeItem`, `updateItem`, etc.); `do-list` title-based helpers still work.
  - No changes to the `EditableList` API; potential follow-up to expose a shared `mintId` helper.

<sup>Written for commit bd630cd30f1ca60bd350346c0dc21f6be2c0c365. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4049?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/patterns/do-list/do-list.test.tsx
+++ b/packages/patterns/do-list/do-list.test.tsx
@@ -1,0 +1,174 @@
+/**
+ * Test Pattern: Do List (post-CT-1712 EditableList migration)
+ *
+ * Proves behavior parity after migrating the id-keyed model onto the
+ * EditableList primitive while do-list keeps its rich rows headless:
+ * - add (carries indent), addItems (batch)
+ * - stable ids minted on add (identity model)
+ * - update by id (title / done), toggle done
+ * - cascade remove by id (item + indent-children)
+ * - title-addressed convenience (removeItemByTitle / updateItemByTitle)
+ * - archiveCompleted drops done items
+ * - counts via the embedded primitive
+ *
+ * Run: deno task cf test packages/patterns/do-list/do-list.test.tsx \
+ *   --root packages/patterns --verbose
+ */
+import { action, computed, pattern } from "commonfabric";
+import DoList, { type DoItem } from "./do-list.tsx";
+
+export default pattern(() => {
+  const list = DoList({});
+
+  // ==========================================================================
+  // Actions
+  // ==========================================================================
+
+  const add_parent = action(() => {
+    list.addItem.send({ title: "Parent" });
+  });
+  const add_child = action(() => {
+    list.addItem.send({ title: "Child", indent: 1 });
+  });
+  const add_sibling = action(() => {
+    list.addItem.send({ title: "Sibling" });
+  });
+
+  const add_batch = action(() => {
+    list.addItems.send({
+      items: [{ title: "BatchA" }, { title: "BatchB", indent: 1 }],
+    });
+  });
+
+  // Identity-addressed: read the live id back before sending.
+  const update_first_title = action(() => {
+    const id = list.items[0]?.id;
+    if (id) list.updateItem.send({ id, title: "Parent!" });
+  });
+  const done_first = action(() => {
+    const id = list.items[0]?.id;
+    if (id) list.updateItem.send({ id, done: true });
+  });
+
+  // Cascade remove: removing "Parent!" (index 0) must also drop its indent-1
+  // "Child" (index 1) but NOT the un-indented "Sibling".
+  const remove_parent_cascade = action(() => {
+    const id = list.items[0]?.id;
+    if (id) list.removeItem.send({ id });
+  });
+
+  // Title-addressed convenience layer.
+  const rename_sibling_by_title = action(() => {
+    list.updateItemByTitle.send({ title: "Sibling", newTitle: "Sibling2" });
+  });
+  const done_sibling_by_title = action(() => {
+    list.updateItemByTitle.send({ title: "Sibling2", done: true });
+  });
+  const remove_batcha_by_title = action(() => {
+    // BatchA has an indent-1 child BatchB → cascade removes both.
+    list.removeItemByTitle.send({ title: "BatchA" });
+  });
+
+  const archive = action(() => {
+    list.archiveCompleted.send({});
+  });
+
+  // ==========================================================================
+  // Assertions
+  // ==========================================================================
+
+  const assert_initial_empty = computed(() => list.itemCount === 0);
+
+  const assert_three = computed(() => list.itemCount === 3);
+  const assert_parent_first = computed(() => list.items[0]?.title === "Parent");
+  const assert_child_indent = computed(() => list.items[1]?.indent === 1);
+  const assert_first_has_id = computed(() =>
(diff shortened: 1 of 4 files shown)


bfollington:
Superseded: the identity model changed under this migration (#4082 removes user-land ids from EditableList per the #4048 review observation), and the LOC findings here showed do-list/simple-list are the wrong proof targets anyway (headless callers grew; see PR body 'contract friction'). The rendered-caller proof will be re-cut against the equals()-identity primitive.

PR #4058, opened 2026-06-11
feat(runner,llm,runtime-client,ui): route space-bound compute to the space's host

jkomoros:
## What

Your #3947 approval note, implemented: *"a follow-up pr should look at the remaining uses of the default url and move to this mechanism for any space-bound uses, e.g. blob upload, llm, fetchData base url, etc."*

- **`Runtime.hostForSpace(space)`** — the compute-side analogue of the storage resolver. `RuntimeOptions` takes the same optional `spaceHostMap` the storage layer does (pass the same one; entries are validated eagerly, naming the space). `healthCheck` fans out over the default host plus every distinct mapped host — one runtime spans hosts, so health is the conjunction.
- **LLM**: `LLMClient.sendRequest/generateObject` accept a per-call `opts.endpoint`; all four builtin call sites route a **host-mapped** space's calls to `/api/ai/llm` on its host. An **unmapped** space keeps the module-level default (`setLLMUrl`) — deliberately NOT `hostForSpace`'s fallback, because deployments may split the pattern-facing api host from the runtime's memory host (toolshed: `API_URL` ≠ `MEMORY_URL`). Unmapped behavior is byte-identical.
- **fetch-data**: relative URLs for a mapped space resolve against its host; unmapped spaces keep `getPatternEnvironment().apiUrl` (same reasoning, documented inline).
- **Blob upload**: `UploadBlobRequest` carries a **required** `space` (the #3995 convention) — the worker posts to and resolves the returned URL against that space's host, with a named-error guard for untyped callers. Threaded through `RuntimeClient`/`RuntimeInternals`/ui `uploadFile`; `cf-file-input`/`cf-code-editor` take the space from their bound cell, else the view's `spaceContext`. **Named behavior change**: in a non-home view, uploads now land under the viewed space's namespace (previously always the worker's init space) — that's the point.
- **Deliberately untouched**, so it's a decision and not an oversight: `wish`'s system-pattern sources (already lazy upstream; platform assets stay on the pattern environment's default host, same class as `getPatternEnvironment` for authored code); `stream-data` (passes URLs to fetch with no base today — nothing to re-route); the static cache and program resolver (default-host platform assets).
- For embedders: `cf-file-input` outside a `spaceContext` provider now throws a named error on upload; the `space` accessor is a plain property, so embedders can also pass it directly.

## Tests

`hostForSpace` mapped/default + healthCheck fan-out (incl. host dedupe + malformed-entry=unhealthy); blob IPC asserts the host is resolved for the request's space; ui `uploadFile` asserts the forwarded space; shell blob integration names its space. Full check/lint/fmt green; runner (585), runtime-client, lib-shell, shell, piece, ui suites green.

---
## Summary by cubic
Routes space-bound compute (LLM, fetch, blob upload) to each space’s host via a per-space resolver; unmapped behavior stays the same. Adds multi-host health checks, freezes the per-space host map, and requires a `space` for uploads with accurate UI `cf-error` messages.

- **New Features**
  - `runner`: `RuntimeOptions.spaceHostMap` and `Runtime.hostForSpace(space)` with early validation; map is snapshotted and frozen; `healthCheck` fans out over the default and all distinct mapped hosts.
  - `llm`: `LLMClient.sendRequest/generateObject` accept a per-call `endpoint`; built-ins route to `/api/ai/llm` on the mapped host, otherwise use the `setLLMUrl` default.
  - `runner` fetch-data: relative URLs resolve against the mapped host; unmapped spaces use `getPatternEnvironment().apiUrl`.
  - `runtime-client`/`ui` blob upload: `UploadBlobRequest.space` is required; uploads and returned URLs resolve against that space’s host. `uploadFile` passes `space`; `cf-file-input` and `cf-code-editor` take the cell’s space or the view’s `spaceContext`, with distinct errors for missing runtime vs missing space and `cf-error` message text that matches the actual failure.

- **Migration**
  - If federated, pass `spaceHostMap` when creating a `Runtime`.
  - Always include `space` in `RuntimeClient.uploadBlob({ ... })`.
  - When embedding `cf-file-input` or `cf-code-editor` outside a `spaceContext`, set the `space` property or wrap with a provider (otherwise a named error is thrown).

<sup>Written for commit ff5390a8be6b2df21dba08131c7c7ab39b531cca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4058?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Perf-check note

Threshold-edge churn, not regression: across three runs the OVER metric rotated (core-piece-values job duration → runner-integration step → gmail-agentic-search), each within 2s/15% of its margin, while main's own runs show the same metrics oscillating around the identical thresholds (gmail-agentic-search is the same main-side drift #3995 already declared). PR4's only unmapped-path additions are a once-at-construction validation loop and a frozen snapshot — nothing on these jobs' paths. Declaring per the tool's suggestion:

NEW_PERF_BASELINE: step: runner integration = 63s
NEW_PERF_BASELINE: subtest: pattern-unit-3/pattern-unit-tests > packages/patterns/google/core/experimental/gmail-agentic-search.test.tsx = 32s

### DIFF
--- a/packages/lib-shell/src/runtime.ts
+++ b/packages/lib-shell/src/runtime.ts
@@ -312,6 +312,7 @@ export class RuntimeInternals extends EventTarget {
   }
 
   async uploadBlob(options: {
+    space: DID;
     contentType: string;
     body: Uint8Array;
     suffix?: string;
--- a/packages/llm/src/client.ts
+++ b/packages/llm/src/client.ts
@@ -442,6 +442,15 @@ export class LLMClient {
   async generateObject(
     request: LLMGenerateObjectRequest,
     abortSignal?: AbortSignal,
+    opts?: {
+      /**
+       * Full LLM endpoint URL for THIS call (e.g.
+       * `new URL("/api/ai/llm", host)`); the module-level default
+       * (setLLMUrl) is the fallback. Lets one runtime route per-space
+       * work to that space's host.
+       */
+      endpoint?: string | URL;
+    },
   ): Promise<LLMGenerateObjectResponse> {
     // Check for mock mode
     if (mockCatalog.isEnabled()) {
@@ -461,7 +470,8 @@ export class LLMClient {
       throw new Error(TEST_GUARD_MESSAGE);
     }
 
-    const response = await fetch(llmApiUrl + "/generateObject", {
+    const endpoint = opts?.endpoint?.toString() ?? llmApiUrl;
+    const response = await fetch(endpoint + "/generateObject", {
       method: "POST",
       headers: { "Content-Type": "application/json" },
       body: JSON.stringify(request),
@@ -500,6 +510,10 @@ export class LLMClient {
     request: LLMRequest,
     callback?: PartialCallback,
     abortSignal?: AbortSignal,
+    opts?: {
+      /** Per-call LLM endpoint; module-level default is the fallback. */
+      endpoint?: string | URL;
+    },
   ): Promise<LLMResponse> {
     if (request.stream && !callback) {
       throw new Error(
@@ -549,7 +563,7 @@ export class LLMClient {
       throw new Error(TEST_GUARD_MESSAGE);
     }
 
-    const response = await fetch(llmApiUrl, {
+    const response = await fetch(opts?.endpoint?.toString() ?? llmApiUrl, {
       method: "POST",
       headers: { "Content-Type": "application/json" },
       body: JSON.stringify(request),
--- a/packages/runner/src/builtins/fetch-data.ts
+++ b/packages/runner/src/builtins/fetch-data.ts
@@ -358,8 +358,14 @@ async function startFetch(
   // Body preprocessing (stringify non-string bodies) is handled by the
   // snapshotInputs callback in tryClaimMutex, so options is ready to use.
   try {
+    // Relative URLs resolve against the executing space's host when the
+    // space is host-mapped (federation: one runtime spans hosts). An
+    // unmapped space keeps the pattern environment's api base — which on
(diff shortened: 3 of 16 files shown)


jkomoros:
Both cubic findings fixed in e9ba8b95a (identified by cubic): the spaceHostMap is now snapshotted + frozen at Runtime construction so post-construction mutation of the caller's object can't change routing (enforcing the documented fixed-for-lifetime contract), and cf-code-editor's paste handler reports a missing space distinctly from a missing runtime.

seefeldb:
> Deliberately untouched, so it's a decision and not an oversight: wish's system-pattern sources (already lazy upstream; platform assets stay on the pattern environment's default host, same class as getPatternEnvironment for authored code); stream-data (passes URLs to fetch with no base today — nothing to re-route); the static cache and program resolver (default-host platform assets).

wish: should fetch from the user's home space's host, which I think is the default host right now, so all good.

> For embedders: cf-file-input outside a spaceContext provider now throws a named error on upload; the space accessor is a plain property, so embedders can also pass it directly.

Ah, there is probably a stream attached to an event handler and we should get the space from there. More plumbing for a follow up PR.

LGTM.

seefeldb (approved):
See my comment

PR #4064, opened 2026-06-11
runner: delete loadId machinery; CFC identity is provenance-only [identity E2]

seefeldb:
PR E2 of [content-addressed-action-identity-implementation-plan.md](https://github.com/commontoolsinc/labs/blob/main/docs/specs/content-addressed-action-identity-implementation-plan.md). Follows E1 ([#4053](https://github.com/commontoolsinc/labs/pull/4053)). **Net −1.6k lines.**

## What this deletes

Every load-scoped surface. CFC implementation identity is now sourced **exclusively** from the function-object-keyed provenance recorded at verified evaluation; the global executable index is the single retained legacy read surface.

- **loadId machinery**: `Frame.verifiedLoadId` + threading, `JavaScriptNodeContext`/`ResolvedJavaScriptModule` fields, `createPatternFrame` param, `EvaluateResult.loadId`; `pattern-metadata` `get/setVerifiedLoadId`; `pattern-binding` propagation; `pattern-manager` `seedVerifiedLoadIds`/`patternToVerifiedLoadId`/`#seededVerifiedRoots`.
- **ExecutableRegistry**: per-load `verifiedFunctions` partitions, `verifiedFunctionLoadIds`, `verifiedLoadSources`/`verifiedLoadBundleIds`, `verifiedBindingMetadata`, `beginVerifiedLoad` repair, the `captureVerifiedValue`/`recordVerifiedFunctions`/`annotateVerifiedPatterns` walks, `captureVerifiedBindingCandidates`. `registerVerifiedFunction` is now a 2-arg global-index write shared by the engine's eval registrar and the runner's in-action registrar.
- **Harness interface**: `getVerifiedLoadId`, `getVerifiedFunctionInLoad`, `isVerifiedSourceInLoad`, `getVerifiedBundleId`, `getVerifiedBindingMetadata`, `registerVerifiedFunction(loadId,…)`.
- **CFC**: the `implementationRef`×`verifiedLoadId` legacy arm in `implementation-identity.ts`. `resolvePolicyFacingImplementationIdentity` takes only `{ implementation }`. A provenance-less function now resolves to **no identity** even when its module carries an `implementationRef` (pinned by a new test). `rebindWriteAuthorizedByClaims` stamps NEW claims with `moduleIdentity` only.
- **FunctionCache** + the instantiation-time prewarm walk (`discoverAndCacheFunctions`).
- Engine `loadId` → `evalId` (it only keys per-evaluation source-map names now).

## What this deliberately KEEPS (gate-2 decision, retiring together at design Phase 4)

Gate 2 (stored-data aging) could not be measured — no production-space sample, and C/D are days old — so the legacy **read** path stays one more cycle:
- `ensureImplementationRef` + ordinal shim feeding the **one** retained string-keyed global executable index (`verifiedFunctionIndex`) behind `getExecutableFunction` — the rehydration channel for pre-flip stored graphs, host-trusted values, and dynamic in-action artifacts.
- The **bundleId-only** `writeAuthorizedBy` verification arm (stored pre-#4009 claims) + `ImplementationIdentity.bundleId`. The field can't drop before the arm — the arm compares against the *live* identity's value, now sourced from provenance (stamped at eval time, since post-flip resolution yields no `verifiedLoadId`). Claims stamped with a **raw** `verifiedLoadId` are no longer served: load ids embed a per-session counter so they never verified cross-session anyway, and same-session claims since #4009 carry `moduleIdentity`, which wins arm selection.
- **`unsafe-host:` story**: the §5 synthetic-identity host registrar is **deferred** to the cycle that retires this read path. Host trust is in-repo a test-only surface (`test/support/trusted-builder.ts` + one piece test; production `createBuilder()` passes no token) and rides exactly the `implementationRef` channel gate 2 keeps — replacing it now would add a trust surface without removing the old one. Recorded in the plan doc.

## CFC continuity (reviewers: focus here)

- **CT-1665 binding identity** now rides exclusively on provenance (`cfc-nonexported-binding-identity` green on provenance alone, internal + exported + profile-home + resume-by-identity).
- The bundleId verification arm's accept/reject behavior is unchanged (the `cfc-boundary`/`cfc-ui-contract` suites pin it; identity literals ported bundleId→moduleIdentity where they drove the writer side).
- A function that was never verified-evaluated gets no identity, full stop — tightened from the old `kind:"unsupported"` to **undefined** where the legacy arm used to answer.

## Red-team pass (E2 merge gate)

An adversarial agent attacked the changed surfaces; every attack lands as a test in `content-addressed-identity-adversarial.test.ts` / `cfreg-security.test.ts`. **All fail closed.** Categories:
1. Forged byte-identical fn (no provenance) → no verified identity.
2. src-spoof (genuine canonical src as own-property, no provenance entry) → unverified.
3. `$implRef` replay (forged/foreign identity or unregistered symbol) → miss; genuine pair returns the real trust-gated fn; CFC identity comes from the resolved fn's provenance, not the ref.
4. Provenance/src mismatch → `unsupported`.
5. Claim forgery — moduleIdentity mismatch, bundleId-only mismatch, both-mismatched, empty fields, **no cross-arm laundering** (a bundleId can't satisfy a moduleIdentity-bearing claim; path still checked under the bundleId arm).
6. Host-artifact escalation → executes, never `kind:"verified"` (debugName short-circuit + no provenance).
7. Dynamic-artifact escalation → inherited bundleId but no symbol/bindingPath, so it can't satisfy a verified-binding claim or reach a foreign owner's field; an unverified invoker mints no provenance.
8. Re-export laundering → provenance sticks to the defining module (source-identity guard).
9. Keyspace separation → admitting a fn to the executable index (rehydration channel) does not make it resolvable through the content-addressed `$implRef` index.

**Note:** the red-team initially ran in a worktree forked before the E2 deletions; I re-ran every attack against the actual E2 code (ported the call signatures, rewrote one attack whose target — the E1-intermediate dynamic partition — E2 deleted) and confirmed all fail closed type-checked. See the last test commit.

## Verification

- `packages/runner` full suite: all passing (type-checked), ~125s
- workspace `deno task check` clean; `deno lint` clean; `deno fmt` clean
- piece + html suites green; `cf check packages/patterns/address.tsx` exit 0
- Pre-flip graph canary + bundleId-arm accept/reject tests stay green throughout

End-state grep: `verifiedLoadId` appears only in comments; `implementationRef` only in the gate-2 retained read path + tests (decisions recorded in the plan doc).

---
## Summary by cubic
Remove all load-scoped machinery and make CFC implementation identity come only from function-object provenance. Keep a single global executable index as the temporary legacy read path while stored data ages out.

- **Refactors**
  - Deleted all `verifiedLoadId` threading and per-load registry partitions; provenance WeakMap is the sole source of kind:"verified".
  - Removed harness load-scoped APIs: `getVerifiedLoadId`, `getVerifiedFunctionInLoad`, `isVerifiedSourceInLoad`, `getVerifiedBundleId`, `getVerifiedBindingMetadata`, `registerVerifiedFunction(loadId,…)`.
  - Dropped `FunctionCache` and the prewarm walk; resolution uses global indexes.
  - Removed the `implementationRef` × `verifiedLoadId` identity arm; `resolvePolicyFacingImplementationIdentity` now takes `{ implementation }` only; unproven functions resolve to no identity.
  - Kept the global `verifiedFunctionIndex` (`getExecutableFunction`) and bundleId-only verification for stored pre-`#4009` claims; new claims stamp `moduleIdentity` only.
  - Engine `loadId` renamed to `evalId` for source-map names; verified-source checks now ride on provenance.
  - Updated specs and expanded red-team tests; CT-1665 binding identity now rides solely on provenance.

- **Migration**
  - Update call sites to `resolvePolicyFacingImplementationIdentity({ implementation })`; remove `harness`/`verifiedLoadId` params.
  - Do not rely on `EvaluateResult.loadId` or removed harness methods; use provenance and the global executable index as needed.
  - When emitting CFC claims, write `moduleIdentity` only; the bundleId-only path remains for existing stored data.

<sup>Written for commit b108ea8174825b0f2fc1befca0ab9517cba88043. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4064?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

NEW_PERF_BASELINE: test: pattern-unit-1/pattern-unit-tests = 200s
NEW_PERF_BASELINE: subtest: pattern-unit-1/pattern-unit-tests > packages/patterns/card-piles/main.test.tsx = 22s
NEW_PERF_BASELINE: subtest: pattern-unit-1/pattern-unit-tests > packages/patterns/battleship/multiplayer/lobby.test.tsx = 20s

**Perf override evidence (junit noise, flip-confirmed):** the three OVER metrics were all on pattern-unit shard 1 of 5 (shards 2–5 unflagged), and are `card-piles` + `battleship/lobby` — game patterns with no connection to the identity/CFC machinery this PR touches. E2 *removes* per-instantiation work (the prewarm graph walk) and swaps a memoized cache for two O(1) map probes, so it has no mechanism to slow these. A rerun of the identical commit dropped both flagged subtests back under threshold to their baseline medians: battleship/lobby 19.0s→14.4s (median 14.2s, threshold 18.0s), card-piles 21.8s→15.8s (median 13.1s, threshold 19.1s) — a clean flip across attempts of the same code (attempt 1 caught a slow runner; the broader table showed a fleet-wide +10–22% drift this run). Override values set just above the noisy peak.


seefeldb:
_cf-review via Claude Fable 5 (`claude-fable-5`), on behalf of @seefeldb. Post-merge review._

**Scope:** 29 files, +824/−1604, packages/runner + specs · **Coverage:** deep-read all src changes and the red-team test additions, sampled mechanical test updates, verified end-state by grep on the merged tree · **Checks run:** residual-reference greps, pre-E2 source comparison for the host-trust and bundleId channels (no test runs — PR merged with full runner suite, workspace check, and canaries green).

This is a clean, well-executed deletion. The PR removes every load-scoped surface and the dead weight around it (FunctionCache, prewarm walk — the latter was ~24% of the commit-callback phase per its own comment), leaves exactly the gate-2 read path it says it leaves, and pins the security claims with real adversarial tests. I went looking for the usual deletion hazards and verified each is sound:

- **Verified-context gate change** (`runner.ts` `invokeJavaScriptImplementation`): the `verifiedLoadId !== undefined ||` arm dropped from the dynamic-registrar gate is genuinely unreachable post-flip — host-trusted functions never had a load id pre-E2 either (`trustHostValue` only ever wrote `trustedHostFunctionIndex`, confirmed against the pre-merge source), and every registry-admitted evaluation product carries provenance. **Verified.**
- **bundleId arm stays live**: `recordModuleProvenance` is called inside `evaluateGraph`, which both the source-compile and resume-by-identity paths share, so the provenance `bundleId` the verification arm now depends on is stamped on both. Same `hashOf(script)` derivation as the deleted `verifiedLoadBundleIds`. **Verified.**
- **FunctionCache deletion is perf-safe**: the only non-O(1) leg (step-3 stringified fallback) is memoized below it by `SESCallbackEvaluator.callbackCreatorCache` (source-keyed). The deletion also *removes* a stale-pinning hazard — a fallback-resolved fn cached by ref used to shadow a later verified registration. **Verified.**
- **Dropped raw-`verifiedLoadId` claim serving**: the reasoning (per-session counter → never verified cross-session; #4009+ claims carry `moduleIdentity` which wins arm selection) is sound and recorded in three places (code, test note, spec).
- **Test changes**: every deleted test covered deleted machinery (per-load repair, seed walk, FunctionCache, loadId fallback); the behavior *tightening* (provenance-less fn: `kind:"unsupported"` → `undefined`) is explicitly pinned by a new test rather than silently absorbed. The red-team additions (attacks 8–12) are the strongest part — they pin the newly-live bundleId arm against cross-arm laundering, empty-id matching, dynamic-artifact escalation, $implRef replay, and keyspace separation. No tests bent to fit.
- **Coherence**: both spec docs updated in the same PR; end-state grep confirms `verifiedLoadId` survives only in explanatory comments, all of which describe it as deleted. The rollback floor (pre-#4009 binaries can't verify moduleIdentity-only claims) is documented at the stamp site.

1. **`json-utils.ts` — `keepLegacyRef` object-identity comparison against the global index** *(suspected, narrow)*. Pre-E2, a frame with a load id probed its own per-load partition first, so a module from an older evaluation still matched its own fn and serialized ref-only. Now the global index returns the *latest* registration; if the same module was re-evaluated in-session, an older graph serializing later sees `admittedImplementation !== implementation` and falls through to embedding `Function.prototype.toString` of the old fn — losing the ref-only form (and, for a closure-bearing dynamic artifact, embedding a body that can't faithfully rehydrate). It fails conservative, not open, and the window is rare (same ref, two live generations, serialize from the older one). Since functions sharing a content-derived ref are "interchangeable by construction" (the registry's own words), comparing **refs** instead of object identity would close it. Fine as a Phase-4 cleanup since this whole branch retires then.

2. **Attack-8 control test tolerance** (`content-addressed-identity-adversarial.test.ts`): the accept-path control asserts only "any residual error must not mention writeAuthorizedBy". A passing-digest assertion would make the control crisper; as written, a future refactor that breaks the accept path with a differently-worded error keeps the test green. Minor — the four reject-path tests around it carry the real weight.

### Verdict

**Sound merge.** Net −1.6k lines, the trust story is *simpler and strictly tighter* after the change (one WeakMap is the proof, one string index is the legacy read path), the gate-2 retentions are each justified and pointed at a recorded retirement plan, and the red-team coverage is exemplary for this kind of security-surface deletion. The two findings above are Phase-4-cleanup material, not defects.

PR #4066, opened 2026-06-11
test(runner): pin the setup-written schema-meta hatch for same-tx link sources (CT-1698)

seefeldb:
## What

Adds one focused cfc-boundary case pinning the `setupResultSchemaFor` knowability hatch in `derivePersistedLinkLabel` — the "count a same-transaction pattern instantiation as pending source metadata" condition behind profile-home's re-enabled element tests.

## Why

#4038 gave `derivePersistedLinkLabel` two knowability hatches for link sources minted by the writing transaction: the source piece's setup-written `["schema"]` meta (`setupResultSchemaFor`, read-your-writes), and the target-schema child-doc fallback. Only the fallback got a direct runner-level pin; the setup-schema hatch was covered solely end-to-end through `profile-home.test.tsx`, so a regression in it would surface as a distant pattern-test failure rather than a pointed runner failure.

## How

The new case mimics piece setup (root value write + `setMetaRaw("schema", ...)` carrying the piece's **own** ifc label) and links the fresh doc into a labeled location in the same commit. The assertion is on label *content*, not just commit success: the persisted link label must be the SOURCE schema's `["card-space"]`. The child-doc fallback can only produce the target-derived `["personal-space"]` here, so the assertion identifies *which* hatch supplied the label.

Red-proven: with `setupResultSchemaFor` temporarily neutered, this test fails exactly that way (label degrades to `["personal-space"]`, commit still succeeds) while the neighboring same-tx child-doc and pre-existing fail-closed cases stay green.

## Testing

- `packages/runner` suite: 585 passed (3051 steps), 0 failed
- `deno task integration pattern-tests`: all 60 files green, including `profile-home.test.tsx` (6/6 under enforce-explicit)

---
## Summary by cubic
Adds a runner test that pins the setup-written `["schema"]` meta hatch in `derivePersistedLinkLabel` for same-transaction link sources (CT-1698).
It verifies the persisted link label comes from the source schema (`["card-space"]`) via `setupResultSchemaFor`, not the target fallback (`["personal-space"]`).

<sup>Written for commit eadadc28e29afca8017940643646b95d17eff15c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4066?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/test/cfc-boundary.test.ts
+++ b/packages/runner/test/cfc-boundary.test.ts
@@ -2532,6 +2532,77 @@ describe("ExtendedStorageTransaction CFC gate", () => {
     }
   });
 
+  it("labels a SAME-TRANSACTION link source from its setup-written schema meta", async () => {
+    // A piece instantiated by this transaction gets its result schema written
+    // at the doc's ["schema"] meta during setup (updateResultSchemaMeta). A
+    // handler that materializes such a piece and links it into a protected
+    // list in ONE commit (profile-home's addElement, CT-1698) leaves the
+    // source with no stored CFC metadata and no pending schema input — the
+    // setup-written schema, read back read-your-writes, is what keeps the
+    // link write from failing closed. The persisted link label must derive
+    // from the SOURCE's own schema: the target-schema child-doc fallback
+    // (previous test) could only produce ["personal-space"] here, so
+    // ["card-space"] proves the setup-schema hatch supplied the label.
+    const { runtime, storageManager } = createRuntime();
+    try {
+      const seed = runtime.edit();
+      seed.setCfcEnforcementMode("enforce-explicit");
+      const target = runtime.getCell(
+        signer.did(),
+        "cfc-link-setup-schema-target",
+        {
+          type: "object",
+          ifc: { confidentiality: ["personal-space"] },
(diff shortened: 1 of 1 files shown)


seefeldb:
CI status: 34/35 green. The one failure — Pattern Unit Tests (4/5), `editable-list.test.tsx` assertion_10 — is inherited from main, not this PR (this diff is runner-test-only). Bisect-verified: main parent a87c99f32 passes 28/28, main tip 6be61af5b (#4017) fails 27/28, and the latest main CI run fails the same job. Root cause: with factory result types unstripped, the transformer infers usage-projection schemas in consuming files that mark accessed extra fields `required` (and drop `Default<>` defaults + `additionalProperties: true`), so the extras-passthrough read collapses. Will rerun the job once main is fixed.

PR #4069, opened 2026-06-11
feat(runtime-client,runner,home-schemas): site-table v0 — the runtime reads where spaces live

jkomoros:
## What

The v0 of hint-driven space resolution from the 2026-06-09 session, built directly per the closing call there (*"maybe we skip [the intermediate] and just do that directly"* — the runtime reads the lookup itself):

- **`home-schemas`**: the site table — a per-user array of `did → host` HINTS in the home space (`siteTableSchema` + the canonical `siteTableCause`, shared by the runtime as reader and embedder daemons as writers). Unverified in v0, per the session ("don't even verify in the first month").
- **Storage**: `StorageManager.registerSpaceHost(space, host)` — a late-bound hint consulted by the address resolver after the frozen seed map, before the default host. The refusal semantics ARE the contract: **the seed wins, and an already-opened space is never silently re-pointed**. An unopened space (the open-a-shared-link case) picks the hint up on first touch because sessions resolve lazily per space. The emulated manager refuses honestly.
- **Runtime**: `registerSpaceHost` is storage-first and mirrors into compute routing only on acceptance — storage and compute can't disagree. `mappedHostFor(space)` (seed ?? learned) is now the single read path behind `hostForSpace`, LLM/fetch-data mapped routing, and the health fan-out.
- **Worker**: subscribes to the home-space site table after boot and registers entries — fire-and-forget, never a boot dependency. **The losing side of the boot race is loud**: if a hint can't take effect (space already opened against the default host) and the effective host differs, the worker warns once per fact.
- **`RegisterSpaceHost` IPC** (+ RuntimeClient/RuntimeInternals): the in-session half for embedder-learned hints (share-link receipt). **Ordering contract, stated in the IPC doc**: push the hint before the first mount of a just-learned space — the table is the durable record, the IPC is the ordering guarantee.

## The paragraph that needs your eyes

**Security delta (deliberate, v0):** the table is ordinary home-space data, so anything with home-space write access — including patterns running there — can now steer where a **not-yet-opened** space is fetched from. Pre-this-PR, only the embedder's trusted seed map could steer. Combined with v0's no-verification and the replayable `session.open` handshake, this is exactly the forcing function for **audience-binding before host hints ever come from less-trusted data** (the parked PR5 item). The schema documents it; flagging it here for an explicit architect ack.

## Scope (so the long-term owner inherits a seam, not a fork)

Hint-reading is **worker-runtime only** in v0: `cf` CLI, background-charm-service, and toolshed's server runtime stay seed-map-only (a TODO, not an oversight). Removals don't unregister until restart; duplicate dids are last-entry-wins (documented in the schema). The loom-side writer (daemon syncing its served sources into the table + the home-host IndexedDB pin) is the companion loom PR.

## Tests

Registry refusal semantics (seed wins; opened spaces never re-point — asserted against the actually-dialed websocket host; fresh opens route through hints); Runtime/storage verdict agreement incl. refusal-no-divergence; the IPC handler round-trip; watchSiteTable end-to-end over loopback storage with per-entry isolation of malformed rows. Full check/lint/fmt green; runner (590), runtime-client (15), lib-shell, shell, piece suites green.

---
## Summary by cubic
Adds v0 of hint-driven space resolution: the runtime reads a per-user DID→host “site table” to route spaces. Seeds still win, and opened spaces are never re-pointed; hints only affect unopened spaces.

- **New Features**
  - `home-schemas`: added `siteTable` schema and `siteTableCause`. Entries are unverified hints; last-entry wins; removals don’t unregister until restart.
  - `runner`: `StorageManager.registerSpaceHost(space, host)` with refusal rules (seed wins; opened spaces not re-pointed). Address resolver consults dynamic hints between seed and default. `mappedHostFor()` added and used by a unified `hostForSpace()`. Health check includes learned hosts. Emulated manager always refuses.
  - `runner` builtins: `fetch-data` and LLM use `mappedHostFor()` for mapped routing.
  - `runtime-client`/`lib-shell`: `RegisterSpaceHost` IPC and client passthrough. Worker subscribes to the home-space site table after boot and registers entries; warns once when a hint can’t take effect.
  - Tests: refusal semantics, storage/runtime agreement, IPC handler, and site-table watch with per-entry isolation.
  - Security (v0): anything with home-space write access can steer where not-yet-opened spaces fetch from; no host verification yet.
  - Hardening: avoid a leaked site-table subscription when disposed during initial sync; snapshot and freeze the seed host map to keep the resolver and refusal logic consistent.

- **Migration**
  - Embedders: when you learn a space’s host (e.g., from a share link), send `RegisterSpaceHost` before first mount. The table is the durable record; the IPC provides ordering.
  - Scope: only the worker runtime reads hints in v0. CLI and other services remain seed-map only.

<sup>Written for commit e7a942cbba3f63f3c6212d89b8b3ec0b195a65dc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4069?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/home-schemas/mod.ts
+++ b/packages/home-schemas/mod.ts
@@ -45,8 +45,13 @@ export {
 } from "./learned.ts";
 
 export {
+  type SiteTable,
+  siteTableCause,
+  siteTableSchema,
   type SpaceEntry,
   spaceEntrySchema,
+  type SpaceHostEntry,
+  spaceHostEntrySchema,
   type SpacesList,
   spacesListSchema,
 } from "./spaces.ts";
--- a/packages/home-schemas/spaces.ts
+++ b/packages/home-schemas/spaces.ts
@@ -24,3 +24,48 @@ export const spacesListSchema = {
 } as const satisfies JSONSchema;
 
 export type SpacesList = Schema<typeof spacesListSchema>;
+
+/**
+ * Site table: where each space lives. The home space carries a per-user
+ * table of `did → host` facts — the v0 of hint-driven space resolution
+ * (2026-06-09 federation session: "the smallest possible slice"). The
+ * runtime reads it as its live host lookup; writers include the local
+ * daemon (syncing its served sources) and link-receipt flows.
+ *
+ * Entries are HINTS: unverified in v0 (the space's own log is the
+ * integrity boundary, not the host), and a hint must never silently
+ * re-point a space the runtime already opened.
+ */
+export const spaceHostEntrySchema = {
+  type: "object",
+  properties: {
+    /** The space DID this fact is about — the table key. */
+    did: { type: "string" },
+    /** Base URL of the host currently serving the space. */
+    host: { type: "string" },
+    /** ISO timestamp of when this fact was recorded. */
+    updatedAt: { type: "string" },
+    /** Who recorded it (e.g. "local-source-sync", "share-link"). */
+    source: { type: "string" },
+  },
+  required: ["did", "host"],
+} as const satisfies JSONSchema;
+
+export type SpaceHostEntry = Schema<typeof spaceHostEntrySchema>;
+
+export const siteTableSchema = {
+  type: "array",
+  items: spaceHostEntrySchema,
+  default: [],
+} as const satisfies JSONSchema;
+
+export type SiteTable = Schema<typeof siteTableSchema>;
+
+/**
+ * Canonical cause for the home-space site-table cell, shared by the
+ * runtime (reader) and embedder daemons (writers) so both address the
+ * same document: `getCell(userDid, siteTableCause(userDid), siteTableSchema)`.
+ */
+export function siteTableCause(userDid: string): { siteTable: string } {
+  return { siteTable: userDid };
+}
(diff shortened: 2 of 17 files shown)


jkomoros:
Both cubic findings fixed (identified by cubic): the post-sync continuation in watchSiteTable now checks disposal before installing the sink (no leaked subscription when dispose races the initial table sync), and the storage seed map is snapshotted + frozen so caller mutation can't desynchronize refusal logic from the resolver — the same fixed-for-lifetime enforcement Runtime's map got in #4058.

seefeldb (approved):

seefeldb:
@bfollington you might want to take a look since this changes/overlaps with the existing space directory.

jkomoros:
@bfollington context for your look — the overlap as I understand it: home-schemas already has `spacesListSchema` (`{name, did}` — the user's managed-spaces registry), and this PR adds a second home-space registry of spaces (`siteTableSchema`: `{did, host}` — where ANY known space lives, including foreign/shared ones the user doesn't manage). They could plausibly converge (a `host` field on a unified space registry) — I kept them separate because the site table covers spaces outside the managed list and its entries are unverified hints with different lifecycle (last-write-wins, never re-points opened spaces). If you'd rather unify, the schema + `siteTableCause` are the only coupling points and nothing has shipped against them yet — cheap to reshape now.

Issue #4072, opened 2026-06-11
editable-list.test.tsx assertion_10 fails deterministically on main since #4017


jkomoros:
Main's `Pattern Unit Tests (4/5)` job is red since 6be61af5b (#4017, "stop stripping cell brands from factory result types").

**Bisection from run history**: main run [27378330053](https://github.com/commontoolsinc/labs/actions/runs/27378330053) (head a87c99f32) is green; the very next run [27378444323](https://github.com/commontoolsinc/labs/actions/runs/27378444323) (head 6be61af5b) fails — exactly one commit in the window, #4017's squash.

**Failure**: `packages/patterns/primitives/editable-list.test.tsx` → `✗ assertion_10 (after action_3)`, deterministic (reproduced on main runs 27378444323 + 27379503918 and on PR #4069's merge-ref checks twice).

Every open PR's merge-ref CI inherits this, so it's currently blocking unrelated PRs.

cc @jkomoros

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jkomoros:
cc @mathpirate (#4017 author) — happy to help verify a fix; PR #4069 reproduces it on every merge-ref run.

mathpirate:
Fixed (I think) in https://github.com/commontoolsinc/labs/pull/4071, merging shortly

mathpirate:
Merged fix PR, lmk if still broken for you

PR #4074, opened 2026-06-11
feat(ui): cf-render provides spaceContext from its rendered piece

jkomoros:
## What

Your #4058 follow-up note: *"there is probably a stream attached to an event handler and we should get the space from there. More plumbing for a follow up PR."*

Implemented at the render root rather than per-event: the handler streams of a piece live in the piece's space, and `cf-render` already holds that fact — its cell's space. So `cf-render` now **provides `spaceContext` for its subtree** (a `ContextProvider` kept in lockstep with the rendered cell). Every space-consuming component inside a rendered pattern — `cf-file-input`'s uploads, `cf-code-editor`'s paste path when unbound — acts in the **piece's** space, with zero per-component or per-event plumbing.

This also closes #4058's embedder caveat: hosts that never provide a view-level `spaceContext` (e.g. loom's PatternHost) get correct spaces inside renders for free. A view-level provider (shell RootView) still applies outside renders; the nearest provider wins, which is the right precedence — the piece over the view.

## Notes

- One provider per render root; value updates when the `.cell` property changes; undefined when the cell can't name a space.
- CI: `Pattern Unit Tests (4/5)` is red on main itself (#4072, regression from #4017) — unrelated, will clear when main is fixed.
- Tested via type-check + ui suite; the context handshake is Lit's own machinery. A browser-harness round-trip test can ride the next ui browser-test batch if you want one.

---
## Summary by cubic
`cf-render` now provides `spaceContext` to its subtree from the rendered piece’s cell, so components inside a pattern act in the piece’s space. This resolves the embedder gap noted in #4058 where hosts without a view-level provider produced wrong spaces.

- **New Features**
  - Added a `ContextProvider` in `cf-render` to publish `spaceContext` for its render root.
  - Keeps the provided space in sync when `cell` changes; falls back to `undefined` if it can’t resolve a space.
  - Ensures `cf-file-input` uploads and unbound `cf-code-editor` paste paths use the piece’s space; nearest provider wins, view-level provider still applies outside renders.

<sup>Written for commit 3a9e2b157484c63a52d8cfe5ce30d92929e61a75. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4074?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ui/src/v2/components/cf-render/cf-render.ts
+++ b/packages/ui/src/v2/components/cf-render/cf-render.ts
@@ -1,7 +1,9 @@
 import { css, html, PropertyValues } from "lit";
 import { createRef, type Ref, ref } from "lit/directives/ref.js";
 import { state } from "lit/decorators.js";
+import { ContextProvider } from "@lit/context";
 import { BaseElement } from "../../core/base-element.ts";
+import { spaceContext } from "../../runtime-context.ts";
 import { render } from "@commonfabric/html/client";
 import type { CellHandle } from "@commonfabric/runtime-client";
 import { type VNode } from "@commonfabric/runtime-client";
@@ -114,6 +116,13 @@ export class CFRender extends BaseElement {
   };
 
   declare cell: CellHandle;
+
+  // The render root of a piece knows the piece's space (its cell's),
+  // so it provides spaceContext for everything it renders — event
+  // handlers, uploads, and any cf-* inside the pattern act in the
+  // PIECE's space, not whatever view happens to host the render
+  // (seefeldb's #4058 note: get the space from the piece's own cells).
+  #spaceProvider = new ContextProvider(this, { context: spaceContext });
   declare variant: UIVariant | undefined;
 
   // Use Lit ref directive for stable container reference across re-renders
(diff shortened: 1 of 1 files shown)


seefeldb:
Hmm, it's in principle possible -- and with multi-user stuff likely -- that a piece transcludes UI from another piece and then this would be wrong.

So I'd still want to get the space from the bound cells. Which we can easily do for $value like bindings, but we don't yet have a mechanism for streams for. code editor definitely has a $value already, so just use that. i don't know offhand whether cf-file-upload has that as well, i was just assuming it didn't from previous messages. that's really the only reason we need to do the extra work for streams.

but maybe this is easier: since in the reconciler in the worker everything lives on a cell, why not just copy the space id into each vdom element as FYI and that is then the context space. then optionally optimize by leaving it off if the parent node's space is the same. gets you a similar effect as this PR but naturally supports cross-space transclusion correctly.

jkomoros:
Right — a render-root provider stamps the OUTER piece's space onto transcluded inner-piece UI, which is exactly backwards for cross-space transclusion. Closing this in favor of your vdom-level design (reconciler annotates each element with its cell's space, elided when same as parent), captured in #4079 to pick up when you give the word. Interim behavior stays as shipped in #4058 (bound cell's space where available, view-level context otherwise).

PR #4076, opened 2026-06-11
fix(runner): key wish sidecar pattern caches by the fetch URL

Hixie:
The suggestion, profile-create, and profile-picker sidecar patterns are memoized in module-level state shared across all wish invocations: an in-flight fetch promise plus the compiled pattern. That state was not keyed by the URL the fetch was started for. setPatternEnvironment can change the apiUrl while a fetch is in flight, and a launch after the change would chain onto the pending fetch aimed at the old URL. If that stale fetch then failed, the memoized promise was cleared for future launches, but the launches already chained onto it received nothing and nothing retried. The pattern was never fetched from the new URL.

This raced in packages/runner/test/wish.test.ts roughly once in eight runs under load. The step "renders #profile wish UI as a create-profile input when the profile is missing" starts a deferred, un-awaited profile-create fetch. Before that fetch reaches the network, it waits on the harness compiler initialization of a runtime the test teardown has already disposed, which can take several seconds on a busy machine. The next step ("fetches the profile-create pattern from the pattern environment apiUrl set after module load") sets a test apiUrl, installs a recording fetch mock, and launches its own wish. That launch chained onto the still-pending stale fetch, so no request to the new URL was ever made and the step timed out at its five-second deadline with the recorded URL list empty. The same sequence can occur in production: in the browser worker wish.ts is imported before the runtime calls setPatternEnvironment with the real API URL, so a wish launched before the environment is set leaves a doomed fetch that later launches would chain onto.

Each cache now records the URL its fetch was started for. A launch whose resolved URL differs starts a fresh fetch instead of reusing the pending one, and a superseded fetch leaves the cache untouched when it settles. Failure handling is unchanged: profile-create and profile-picker drop a failed fetch so a later launch retries, and suggestion keeps the failed result.

The three copy-pasted cache blocks, fetch-and-compile functions, and URL helpers are consolidated into one createSidecarPatternCache factory. Each cache is declared by its file name, compile-space option, and retry policy.

Verified by running the wish test file 32 times under four-way parallel load: four failures before the change, none after.

---
## Summary by cubic
Prevents stale wish sidecar patterns from running after `apiUrl` changes by keying caches to the fetch URL and guarding with promise identity; superseded fetches resolve to undefined so older results can’t overwrite newer runs.

- **Bug Fixes**
  - Cache per-URL and active promise; a different URL or overtaken fetch starts a new request. Superseded fetches return undefined and never update the cache or run. Retry policy unchanged: `profile-create` and `profile-picker` retry on failure; `suggestion` keeps the failed result.

- **Refactors**
  - Consolidated logic into an exported `createSidecarPatternCache` (lazy URL resolution, optional compile-in-user-space, retry policy) and applied to `suggestion.tsx`, `profile-create.tsx`, and `profile-picker.tsx`. Added tests for supersession, failure→retry, no-retry, and onSuccess firing only for the winning fetch.

<sup>Written for commit 798441148664007ebb8a27f4a847923a2c2ad46f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4076?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/builtins/wish.ts
+++ b/packages/runner/src/builtins/wish.ts
@@ -1141,8 +1141,8 @@ function releaseSharedHashtagResolver(runtime: Runtime, key: string): void {
 // cache tracks the URL each fetch was started for: `setPatternEnvironment`
 // can change the apiUrl while a fetch is in flight, so a launch for a
 // different URL starts a fresh fetch, and a superseded fetch leaves the cache
-// untouched when it settles.
-function createSidecarPatternCache(options: {
+// untouched and resolves to undefined when it settles.
+export function createSidecarPatternCache(options: {
   // File name under `api/patterns/system/`. Also labels errors.
   name: string;
   // Compile with the user's home space as cache context, so the
@@ -1201,8 +1201,8 @@ function createSidecarPatternCache(options: {
     },
     // Memoized fetch for the current environment's URL, started by this call
     // when none is in flight for that URL. When this call starts the fetch,
-    // `onSuccess` runs once it resolves with a pattern, unless a fetch for
-    // another URL superseded it.
+    // `onSuccess` runs once it resolves with a pattern, unless a later fetch
+    // superseded it. A superseded fetch resolves to undefined.
     fetch(
       runtime: Runtime,
       onSuccess?: (pattern: Pattern) => void,
@@ -1211,8 +1211,14 @@ function createSidecarPatternCache(options: {
       if (!fetchPromise || fetchUrl !== url) {
         fetchUrl = url;
         pattern = undefined;
-        fetchPromise = fetchPattern(runtime, url).then((fetched) => {
-          if (fetchUrl !== url) return fetched;
+        const started: Promise<Pattern | undefined> = fetchPattern(
+          runtime,
+          url,
+        ).then((fetched) => {
+          // Only the fetch the cache currently points to records and reports
+          // its result; launches chained on a superseded fetch get undefined
+          // so a stale pattern is never run.
+          if (fetchPromise !== started) return undefined;
           pattern = fetched;
           if (fetched) {
             onSuccess?.(fetched);
@@ -1221,6 +1227,7 @@ function createSidecarPatternCache(options: {
           }
           return fetched;
         });
+        fetchPromise = started;
       }
       return fetchPromise;
     },
--- a/packages/runner/src/builtins/wish.ts
+++ b/packages/runner/src/builtins/wish.ts
@@ -1142,6 +1142,11 @@ function releaseSharedHashtagResolver(runtime: Runtime, key: string): void {
 // can change the apiUrl while a fetch is in flight, so a launch for a
 // different URL starts a fresh fetch, and a superseded fetch leaves the cache
 // untouched and resolves to undefined when it settles.
+//
+// The cache is keyed only on the URL, not the user identity, even with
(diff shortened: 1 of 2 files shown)


mathpirate:
everything here looks good to me. this isn't exactly my area so i might be missing something, but it appears sensible and i don't see any issues

my claude did want to request one additional regression test to pinpoint the core behavior you're fixing here. it says:

//

Nice fix — the two-guard design is exactly right, and I traced it carefully: URL-keying on cached()/fetch() plus the fetchPromise !== started promise-identity check is what makes it correct even under a URL flap back to a previously-used value (identity, not URL-equality, is genuinely load-bearing there). I also checked the refactor is behavior-preserving across all three patterns — compile-space flags, the retry asymmetry, the readyCell → onSuccess move, tx-vs-own-tx run semantics, and the generated URLs all map 1:1 to the old code. 24/24 green under parallel load here.

One thing I'd like before merge: a test for the failure → retry path. The new test only exercises a successful superseded fetch, but the bug this PR describes was a failed stale fetch that nothing retried. I disabled the retry-clear (fetchPromise = undefined at the retryOnFailure branch) locally and the whole suite still passed — so a regression that re-breaks retry-after-failure would ship green. Something like, reusing the existing fakeRuntime/gate harness:

it("retries after a failed fetch when retryOnFailure is set", async () => {
  // ...save/restore globalThis.fetch + pattern environment as in the test above...
  let attempts = 0;
  globalThis.fetch = (async (input: Request | URL | string) => {
    const url = new URL(input instanceof Request ? input.url : String(input));
    attempts++;
    return attempts === 1
      ? new Response("nope", { status: 404 })          // first launch: fetch fails
      : new Response(`source from ${url.host}`, { status: 200 });
  }) as typeof fetch;

  setPatternEnvironment({ apiUrl: new URL("https://env-a.test/") });
  const cache = createSidecarPatternCache({ name: "profile-create.tsx", retryOnFailure: true });

  expect(await cache.fetch(fakeRuntime)).toBeUndefined();   // failed
  expect(cache.cached()).toBeUndefined();
  expect(await cache.fetch(fakeRuntime)).toEqual({ source: "source from env-a.test" }); // retried
  expect(attempts).toBe(2);
});
(The expect(attempts).toBe(2) is the part that catches the mutation above.) A no-retryOnFailure counterpart asserting attempts stays 1 would also pin the suggestion "keeps-failed-result" policy, which is currently only enforced by review. Both optional-but-nice: a failed-superseded case (stale fetch fails after the newer one wins) and an onSuccess spy.

Non-blocking aside: the caches are still module-level singletons keyed only on apiUrl, not userIdentityDID, even though compileInUserSpace compiles in the user's home space — so a same-apiUrl identity switch serves the prior user's compiled pattern. It's benign (these patterns are space-independent and run with the current runtime) and strictly better than before, but since this PR establishes the keying seam, might be worth a one-line comment recording that the cross-identity reuse is intentional.

mathpirate (approved):
lgtm, see my top-level PR comment for a claude-requested additional regression test

PR #4091, opened 2026-06-12
Added a `dataUrl` mode to the fetchDataFunction, which lets us return a base64 data url, and capture binary data.

ubik2:
## Summary by cubic
Adds a new `dataUrl` mode to `fetchData` that returns binary responses as base64 data URLs. Default behavior for `json` and `text` is unchanged.

- **New Features**
  - Expanded `FetchDataFunction` mode to include `"dataUrl"`.
  - Runner now handles `"dataUrl"`: reads `arrayBuffer`, base64-encodes, and returns `data:<content-type>;base64,<payload>` (falls back to `application/octet-stream` if missing).
  - Added test to verify correct data URL output and pending/error states.

<sup>Written for commit 43cf9dc7f0cdeaf415f44f26150708287a66a071. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4091?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -2125,7 +2125,7 @@ export type FetchOptions = {
 export type FetchDataFunction = <T>(
   params: Opaque<{
     url: string;
-    mode?: "json" | "text";
+    mode?: "json" | "text" | "dataUrl";
     options?: FetchOptions;
     result?: T;
   }>,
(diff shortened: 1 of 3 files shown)


ubik2:
I added this feature because I wanted to make a call to generateImage and then save that into a cell.

PR #4090, opened 2026-06-12
feat(runner,memory): speculation lineage for event-launched work (scheduler-v2 E1)

seefeldb:
Stack: 03/E1 — based on scheduler-v2/02-e0 (#4088).

## Summary

Implements scheduler-v2 E1 speculation lineage for work launched by event handlers:

- records per-space origin commit localSeqs and threads `origin-committed` preconditions through runner and memory v2 behind `experimental.commitPreconditions`;
- adds memory engine validation and wire propagation for `PreconditionFailedError`;
- adds a scheduler lineage registry and event gating so same-space descendants carry preconditions, cross-space descendants park until origin confirmation, and failed-origin descendants are dropped;
- compensates handler-result piece launches by stopping pieces if the launching handler transaction fails;
- documents the reviewer-approved already-settled-origin lineage behavior and records WO03 progress/evidence.

## Validation

- `deno fmt`, `deno lint`, and `deno check` on touched TypeScript files before each implementation commit.
- `cd packages/runner && ENV=test deno test --allow-ffi --allow-env --allow-read --allow-write=/tmp,/var/folders --allow-run=git test/scheduler-event-lineage.test.ts` passed: `1 passed (7 steps)`, `0 failed`.
- `cd packages/runner && ENV=test deno test --allow-ffi --allow-env --allow-read --allow-write=/tmp,/var/folders --allow-run=git test/scheduler-lineage.test.ts` passed: `1 passed (8 steps)`, `0 failed`.
- `cd packages/runner && deno task test` passed: `590 passed (3090 steps)`, `0 failed`, `0 ignored (10 steps)`, `2m4s`.
- `cd packages/memory && deno task test` passed: `211 passed (95 steps)`, `0 failed`.

## Review Notes

This work order touches `packages/memory`; please include memory-owner review. `docs/specs/scheduler-v2/implementation/PROGRESS.md` contains the red-first outputs, reviewer stop/resolution notes, and the WO03 exit-checklist self-check.

---
## Summary by cubic
Implements scheduler‑v2 E1 speculation lineage for event‑launched work. Descendants are gated by their origin: same‑space commits carry an `origin-committed` precondition, cross‑space events park until the origin settles, and read‑only origins are treated as settled to avoid parking or errors in read contexts. Adds commit preconditions to `@commonfabric/memory` v2 (behind `commitPreconditions`) with a wired `PreconditionFailedError`, and stops handler‑launched pieces when the origin fails.

- New Features
[...]

PR #4088, opened 2026-06-12
feat(runner): durable event ids + permanent-rejection taxonomy (scheduler-v2 E0)

seefeldb:
Stack: 02/E0 — based on scheduler-v2/01-phase0 (#4087)

## Summary

Implements scheduler-v2 work order 02:

- Adds durable event id minting and threads event ids/origin transactions through queued events and requeues.
- Adds `dispatchedEventId` on runner storage transactions for event dispatch causality.
- Adds permanent-rejection taxonomy and gates event/reactive commit retry paths on `PreconditionFailedError`.
- Adds focused unit coverage for event identity and rejection taxonomy.

## Notes

- Step 3 was executed before Step 2 per reviewer verdict because Step 2 assigns `tx.dispatchedEventId` and Step 3 declares that field.
- Parent PR #4087 review feedback was fixed on `scheduler-v2/01-phase0`, pushed, and merged forward into this branch with a normal merge.
- Work order 02 lists no phase-specific benchmark command; none was run.

## Validation

- `deno fmt` / `deno lint` / `deno check` on touched Step 4 files.
- `ENV=test deno test --allow-ffi --allow-env --allow-read --allow-write=/tmp,/var/folders --allow-run=git test/scheduler-event-identity.test.ts`
- `ENV=test deno test --allow-ffi --allow-env --allow-read --allow-write=/tmp,/var/folders --allow-run=git test/scheduler-rejection-taxonomy.test.ts`
- `cd packages/runner && deno task test` → `588 passed (3074 steps)`, `0 failed`, `0 ignored (10 steps)`, `2m3s`.
- Exit checklist greps recorded in `docs/specs/scheduler-v2/implementation/PROGRESS.md`.

---
## Summary by cubic
Adds durable event IDs to scheduler events and threads them through dispatch and requeues for clear causality. Introduces a permanent-rejection taxonomy so retries skip `PreconditionFailedError`, and exposes `dispatchedEventId` on transactions.

- New Features
  - Mint durable IDs with `mintEventId()`; stable per origin transaction via counter, or random when no origin.
  - `Scheduler.queueEvent()` accepts `opts` with `eventId` and `originTx`; requeues preserve `id` and `originTx`; cell stream sends pass `originTx` automatically.
  - Dispatcher sets `tx.dispatchedEventId` to the queued event ID for handler-result causality.
  - Retry paths in event handling and reactive commits skip when `isPermanentRejection(error)` (`PreconditionFailedError`).

- Migration
[...]

PR #4087, opened 2026-06-12
refactor(runner): remove push scheduler mode (scheduler-v2 phase 0)

seefeldb:
Stack: 01/Phase0 — base `main` (spec PR merged)

## Summary

- Removed push scheduler mode APIs and collapsed scheduler/runner execution to the remaining pull path.
- Deleted the five push scheduler helper modules and removed direct push-mode test/bench coverage arms.
- Removed the `scheduler.mode.change` telemetry marker; the frozen graph snapshot `pullMode` field intentionally remains and now always reports `true`.
- Updated scheduler-v2 work-order notes, progress log, and pull-based scheduler docs to reflect phase-0 removal.

## PROGRESS.md Excerpt

Commits:

- `5e70065ac` — remove push-mode usage from tests, helpers, and benches.
- `633b61ea4` — graph-snapshot `pullMode` survives phase 0.
- `28220f288` — steps 2-4 runner call sites are one compile unit.
- `552376b9e` — CFC trigger-read push arm is part of the compile unit.
- `30ea32193` — remove push scheduler mode and `pullMode` branches.
- `fa4e47a6c` — escape the `mode.change` grep; shell stale comment in scope.
- `2d9053c46` — drop push-mode branches from runner and telemetry.
- `46c31b308` — pull-based scheduler docs reflect push-mode removal.
- `37a232f7c` — phase-end self-check.

Reviewer-directed deviations recorded in `PROGRESS.md`:

- Step 3 grep expects exactly one residual `pullMode: true` in `createGraphSnapshotState()`; graph snapshot telemetry type remains for external consumers.
- Steps 2+3 include the four `src/runner.ts` mode API call-site edits because removing the scheduler API otherwise cannot type-check.
- Steps 2+3 include the `scheduler-cfc-trigger-reads.test.ts` direct push-notification arm removal because it imported a deleted push module.
- Step 4 includes removal of one stale commented `scheduler.mode.change` line in shell debugger code; corrected grep is `scheduler\.mode\.change`.

Validation:

- Full runner suite at phase end: `cd packages/runner && deno task test` → `586 passed (3067 steps)`, `0 failed`, `0 ignored (10 steps)`, `2m3s`.
- Exit grep: `ls src/scheduler/ | grep push` → no matches.
[...]

PR #4096, opened 2026-06-12
feat(runner,memory): exactly-once event handling via result-cell receipts (scheduler-v2 E2)

seefeldb:
Stack: 04/E2 — based on scheduler-v2/03-e1 (#4090)

## Summary

Implements scheduler-v2 E2 result-cell receipts for exactly-once event handling. This adds single-handler event dispatch, event-derived handler frame causes, receipt creation for handled events, permanent lost-race handling, and receipt fixtures.

After reviewer/memory-owner feedback, receipts now use commit-level `entity-absent` preconditions rather than op-level `createOnly`. `markCreateOnly` keeps its runner API but emits commit preconditions independent of surviving semantic operations, so duplicate idempotent handlers still reach memory and fail with `receipt-exists` instead of locally succeeding as no-ops.

## Validation

- `cd packages/memory && deno task test`: passed, `211 passed (95 steps)`, `0 failed`
- `cd packages/runner && deno task test` with default `commitPreconditions` off: passed, `591 passed (3097 steps)`, `0 failed`, `0 ignored (10 steps)`, `2m5s`
- `cd packages/runner && deno task test` with temporary `scheduler-test-utils.ts` default `commitPreconditions: true`: passed, `591 passed (3097 steps)`, `0 failed`, `0 ignored (10 steps)`, `2m5s`
- Focused receipt fixture: `cd packages/runner && ENV=test deno test --allow-ffi --allow-env --allow-read --allow-write=/tmp,/var/folders --allow-run=git test/scheduler-event-receipts.test.ts`: passed, `1 passed (6 steps)`, `0 failed`
- Focused memory preconditions: `deno test --allow-ffi --allow-env --allow-read --allow-write=/tmp,/var/folders packages/memory/test/v2-commit-preconditions.test.ts`: passed, `11 passed`, `0 failed`

See `docs/specs/scheduler-v2/implementation/PROGRESS.md` for the full work-order recordings, reviewer stop/resolution notes, and exit-check greps.

---
## Summary by cubic
Adds scheduler-v2 E2 receipts to guarantee exactly-once event handling. The runner writes a receipt per handled event and memory enforces a commit-level `entity-absent` precondition, rejecting duplicates with `PreconditionFailedError("receipt-exists")`.

- **New Features**
  - Exactly-once via result-cell receipts: memory `entity-absent` precondition (tombstones count), allows precondition-only commits, and bypasses zero-op short-circuit when preconditions exist.
  - Runner marks the handler’s result cell as the receipt across spaces and for deferred `navigateTo`; no-launch handlers write `{}`; idempotent replays still emit preconditions even if writes elide.
  - Scheduler treats `receipt-exists` as permanent (logs “event-lost-race”); transient conflicts still retry with the same receipt id.
  - One handler per event link; last registration wins with a warn marker.
  - Handler frame causes derive from the durable event id; fallback to `crypto.randomUUID()` for non-dispatch calls.
  - Typed precondition failures propagate through remote memory so the scheduler sees `PreconditionFailedError("receipt-exists")`.

- **Migration**
  - Receipts are gated by the experimental `commitPreconditions` flag (default off). Enable to activate exactly-once.
  - Update `packages/memory` and `packages/runner` together; no data migration required.

<sup>Written for commit 3d6505f18a03c21cf300a2e81f8188f5518fce84. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4096?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/scheduler-v2/implementation/04-phaseE2-receipts.md
+++ b/docs/specs/scheduler-v2/implementation/04-phaseE2-receipts.md
@@ -72,27 +72,32 @@ determinism: STOP.
 
 Commit: `feat(runner): handler frame causes derive from the event id (scheduler-v2 E2)`
 
-## Step 3 — Engine: create-only entity precondition
+## Step 3 — Engine: entity-absent receipt precondition
 
 Files: `packages/memory/v2/engine.ts` plus the commit-operation type
 located in work order 03 step 2.1.
 
-1. Extend the `set` commit operation type with optional
-   `createOnly?: true` (wire-compatible optional field).
-2. In `applyCommitTransaction`, where `set` operations are applied: if
-   `createOnly` and a head revision already exists for that
-   (id, scope, branch), reject the whole commit via the same route as the
-   origin-committed rejection (work order 03 step 3), with
-   `{ name: "PreconditionFailedError", precondition: "receipt-exists" }`.
-   "Head exists" must use the same existence notion the engine uses for
-   conflict detection — locate the head lookup used by `findConflictSeq`
-   and reuse it; if ambiguous (tombstones/deletes), STOP and ask the
-   memory owner whether a deleted head counts as existing (record the
-   answer in PROGRESS.md and as a code comment).
-3. Engine tests (extend `v2-commit-preconditions.test.ts` from work order
-   03): create-only on a fresh entity applies; second create-only commit
-   for the same entity rejects with `receipt-exists`; a normal `set`
-   without the flag still overwrites.
+1. Extend the commit precondition type with
+   `{ kind: "entity-absent", id, scope? }` (wire-compatible optional
+   precondition kind). Do not use an operation-level `createOnly` flag;
+   receipts are commit-level preconditions independent of surviving writes.
+2. In `validateCommitPreconditions`: if `entity-absent` and a head revision
+   already exists for that (id, scope, branch), reject the whole commit via
+   the same route as the origin-committed rejection (work order 03 step 3),
+   with `{ name: "PreconditionFailedError", precondition: "receipt-exists" }`.
+   "Head exists" includes deleted/tombstoned heads and must use the same
+   scope-key resolution and delete-aware existence semantics as the engine's
+   set/delete conflict checks.
+3. Allow commits with no operations when preconditions are present, and write
+   the commit row for localSeq continuity. This is required so a duplicate
+   handling whose writes are fully elided still reaches the engine and fails
+   the receipt precondition instead of succeeding locally as a no-op.
+4. Engine tests (extend `v2-commit-preconditions.test.ts` from work order
+   03): entity-absent on a fresh entity applies; second entity-absent commit
+   for the same entity rejects with `receipt-exists`; normal `set` without the
+   precondition still overwrites; create → delete → entity-absent for the same
+   entity rejects with `receipt-exists`; precondition-only commits pass/fail
+   correctly for absent/present entities.
 
 Commit: `feat(memory): create-only set precondition for event receipts (scheduler-v2 E2)`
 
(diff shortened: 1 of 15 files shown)


seefeldb:
Reviewer + memory-owner pass against the WO04 exit checklist — approved on both counts.

**Design note for posterity**: the original op-attached `createOnly` flag was discovered (by red fixtures) to lose its witness whenever a duplicate handling's writes elide as no-ops; this PR's final shape replaces it with the commit-level `entity-absent` precondition, sibling to `origin-committed`. That is the shape the spec's §7.6 always described.

**Memory-owner verification of the engine**:
- `entity-absent` resolves scope via the existing `resolveScopeKey` helper and tests existence via `selectSetDeleteConflict` with `after_seq: 0` — any historical set/delete counts, the correct conservative reading of exactly-once-*ever* (tombstones included, per standing ruling; create→delete→create-only engine test present).
- Precondition-bearing commits are exempted from the zero-op rejection on both sides (client short-circuit and engine), closing the idempotent-redelivery escape; precondition-only commits write a commit row for localSeq continuity.
- Strict `ProtocolError` on unknown kinds retained; transitional boolean flag = E1+E2 ship together (recorded; flag default-off until merge).

**Runner side**: one handler per link enforced; `crypto.randomUUID` only as the non-dispatch fallback; `markCreateOnly` confined to handler-result + deferred-start paths (navigateTo deferred path threaded rather than skipped — nice); receipts unconditional under the flag, absent without; lost race → drop + warn + telemetry, never retried. Suites green flag-on and flag-off (runner 591, memory 211).

Phase E complete: #4087 ← #4088 ← #4090 ← #4096 all reviewed.

seefeldb:
Both review findings fixed (red-first):

1. **Deferred-navigate receipt placement** (Codex P1 + cubic): confirmed — `setupDeferredHandlerResultPattern` created the result-cell head in the handler tx via `setupInternal` while the create-only mark rode the deferred start tx, so the first delivery's deferred start rejected as `receipt-exists` (red test: first delivery never navigated) and redeliveries were unguarded. The mark now rides the creating (handler) transaction; `startAfterSuccessfulCommit` loses the now-unused `markCreateOnlyResult` param. Test: "navigateTo handler results navigate once and deduplicate redelivery".
2. **`toRejectedError` unsafe `.precondition` access** (cubic P2): optional-chained so a primitive/null rejection can't throw while normalizing a commit failure.

Rebase note: main (after #4090) validates preconditions ahead of every commit shape, which is unsafe to re-run on replays once `entity-absent` exists — so the generic same-session replay check is hoisted above precondition validation (still ahead of the observation fast paths, which keep their own replay table and never insert into the commits table). Covered by the existing replay/receipt fixtures plus the merged precondition suite (14 green). Runner 615 green, memory 212 green locally.

seefeldb:
**Reviewer pass on the rebased series** (range-diff original `6ebfbff2d..a49811102` vs `main..adc5ba1f3`, plus the new tip):

- Commits 2, 4, 5, 7, 8, 9 rebased identically; 1 and 3 carry only PROGRESS/context drift. The one substantive integration is in the entity-absent commit (`d430bc78a`), and it's correct:
  - **Replay-detect hoisted above precondition validation.** Necessary consequence of #4090's "validate ahead of every commit shape": re-validating `entity-absent` on a same-session replay would check against the state the original application created and wrongly reject an idempotent resend. New order (empty-check → branch → replay-detect → validate → fast paths) preserves both invariants; precondition-only commits insert commit rows, so their replays also resolve via the stored result; observation-only commits keep their own replay table and never reach the generic check.
  - Per-kind `switch` with `originLocalSeq` integer check correctly scoped to the `origin-committed` arm.
  - Entity-absent fixture coverage is strong, including the tombstone rule (deleted head counts as existing — exactly-once *ever*) and the wire round-trip.
- `3d6505f18` (deferred-navigate receipts) — verified: the create-only mark must ride the transaction that **creates** the result-cell head (the handler tx via `setupInternal`), not the deferred start tx. The old placement broke both directions: first delivery rejected as `receipt-exists` (start tx guarding a head the handler tx already committed) while redeliveries went unguarded (their handler tx carried no mark). Moving the mark to the handler tx also matches the design decision that the receipt *is* the result cell. `toRejectedError` optional-chaining: trivially right. Red-first test present.

**One ask before merge:** add a replay fixture pinning the hoist — same session resends a receipt commit (same `localSeq`, identical content) → engine returns the stored result instead of an `entity-absent` rejection; and a content-mismatched resend still dies as `ProtocolError` replay mismatch. That ordering is the exact hazard the hoist exists to prevent, and nothing currently fails if someone later "simplifies" validation back above the replay check.

seefeldb:
The replay-fixture ask above didn't land before merge — supplied it as [#4127](https://github.com/commontoolsinc/labs/pull/4127) (test-only, both fixtures verified red against the inverted ordering).

PR #4098, opened 2026-06-12
refactor(runner): static write surface; remove write-set discovery (scheduler-v2 phase 1)

seefeldb:
Stack: 05/phase1 — based on scheduler-v2/04-e2 (#4096)

PROGRESS excerpt:

## 05/step-0

- [x] no commit — phase 1 benchmark baseline captured before static write
  surface changes.
- Deviations: none.
- Recordings:
  - `cd packages/runner && deno bench --allow-read --allow-write
    --allow-net --allow-ffi --allow-env --no-check test/scheduler.bench.ts
    test/scheduler-demand-roots.bench.ts
    test/scheduler-stale-propagation.bench.ts`: passed.
  - `test/scheduler.bench.ts`:
    - `Scheduler - 100 computations, shared entity reads`: 19.4 ms
    - `Scheduler - wide graph (1 source, 100 readers)`: 17.9 ms
    - `Scheduler - 100 entities, sparse deps`: 17.3 ms
    - `Scheduler - deep chain (50 levels)`: 11.5 ms
    - `Scheduler - diamond pattern (10 diamonds)`: 9.6 ms
    - `Scheduler - repeated dirty marking`: 7.7 ms
    - `Scheduler - subscribe/unsubscribe cycle (100x)`: 5.8 ms
    - `Scheduler - pull with resubscribe (50 pulls)`: 291.3 ms
    - `Overhead - setup/teardown only`: 1.3 ms
    - `Overhead - create 100 cells (getCell + set)`: 15.4 ms
    - `Overhead - 100x getCell only (no set)`: 1.6 ms
    - `Overhead - 100x set on existing cells`: 15.7 ms
    - `Overhead - runtime.idle() empty`: 1.3 ms
    - `Overhead - commit after 100 sets`: 15.9 ms
    - `Overhead - empty commit`: 1.3 ms
    - `Overhead - 100 raw tx.write + commit`: 7.1 ms
    - `Utility - sortAndCompactPaths (100 paths)`: 21.7 us
    - `Utility - sortAndCompactPaths (1000 paths)`: 280.5 us
    - `Utility - addressesToPathByEntity (100 paths)`: 14.9 us
    - `Utility - addressesToPathByEntity (1000 paths)`: 150.3 us
    - `Scheduler - bare subscribe (100x)`: 1.7 ms
    - `Scheduler - subscribe 100 actions reading same entity`: 1.7 ms
    - `Scheduler - resubscribe cycle (100x)`: 1.4 ms
  - `test/scheduler-demand-roots.bench.ts`:
    - `Scheduler demand roots - effect demand root`: 147.1 ms
    - `Scheduler demand roots - event demand root`: 131.8 ms
    - `Scheduler demand roots - mixed effect and event roots`: 169.0 ms
    - `Scheduler demand roots - parent clears generated children`: 81.3 ms
  - `test/scheduler-stale-propagation.bench.ts`:
    - `Scheduler stale propagation - chain`: 105.5 ms
    - `Scheduler stale propagation - diamond`: 94.7 ms
    - `Scheduler stale propagation - wide fanout`: 244.0 ms
    - `Scheduler stale propagation - dynamic deps`: 82.9 ms
    - `Scheduler stale propagation - unchanged recompute`: 74.7 ms

## 05/step-1

- [x] ca77bc16e — static write surface demand fixtures added.
- Deviations: both fixtures already pass on current code, so they pin existing
  behavior; no behavior-change red case was observed.
- Recordings:
  - `deno fmt packages/runner/test/scheduler-static-writes.test.ts`: passed
    (`Checked 1 file`).
  - `deno lint packages/runner/test/scheduler-static-writes.test.ts`: passed
    (`Checked 1 file`).
  - `deno check packages/runner/test/scheduler-static-writes.test.ts`: passed.
  - `cd packages/runner && ENV=test deno test --allow-ffi --allow-env
    --allow-read --allow-write=/tmp,/var/folders --allow-run=git
    test/scheduler-static-writes.test.ts`: passed, `1 passed (2 steps)`,
    `0 failed`.

## IMPLEMENTER STOP — 05/step-3 static write surface sweep

Stopped after implementing the step 2/3 static-surface rewrite locally because
the full runner suite produced failures outside the three work-order-named
rewrite files. Continuing would require either changing public scheduler API
compatibility for direct `subscribe(..., ReactivityLog)` callers, or widening
test rewrites beyond the allowed files.

Local changes currently include:

- `pull-subscriptions.ts`: unconditional static surface registration from
  action annotations before `immediateLog` setup.
- `dependency-updates.ts`: scheduling writes are derived only from the
  annotated static surface, never `log.writes`.
- `scheduling-writes.ts`: deleted dynamic write discovery helpers and
  historical might-write mode; `SchedulerWriteIndex` now stores the current
  static surface.
- `dependency-graph.ts`/`subscriptions.ts`/`scheduler.ts`: added static
  registration-time dependent edge attachment so existing readers can see a
  newly registered static writer without resurrecting dynamic write-growth
  diffing.
- `action-run.ts`: persisted observations keep both write fields populated
  from the static surface.
- `runtime.ts` and experimental-option tests: removed
  `schedulerHistoricalMightWrite`.
- Named test rewrites in `scheduler-ordering.test.ts`,
  `scheduler-observations.test.ts`, and `scheduler-effects.test.ts` assert
  static surfaces instead of v1 write-set learning. The step-1 fixture was
  also corrected to express declared writes through `action.writes`, because
  v2 ignores direct `log.writes` as scheduler surface.

Verification before the full-suite STOP:

- `deno fmt packages/runner/src/scheduler/pull-subscriptions.ts
  packages/runner/src/scheduler/dependency-updates.ts
  packages/runner/src/scheduler/scheduling-writes.ts
  packages/runner/src/scheduler/dependency-graph.ts
  packages/runner/src/scheduler/action-run.ts
  packages/runner/src/scheduler/subscriptions.ts packages/runner/src/scheduler.ts
  packages/runner/src/runtime.ts packages/runner/test/experimental-options.test.ts
  packages/runner/test/scheduler-ordering.test.ts
  packages/runner/test/scheduler-observations.test.ts
  packages/runner/test/scheduler-effects.test.ts
  packages/runner/test/scheduler-static-writes.test.ts`: passed
  (`Checked 13 files`).
- `deno lint` on the same 13 files: passed (`Checked 13 files`).
- `deno check` on the same 13 files: passed.
- `grep -rn "backfillDependentsForNewWrites\|pruneDependentsForCurrentWrites"
  packages/runner/src/`: no matches.
- `grep -rn "schedulerHistoricalMightWrite\|historicalMightWrite" packages
  --include="*.ts"`: no matches.
- `grep -rn
  "buildKnownSchedulingWrites\|historicalMightWrite\|diffSchedulingWrites\|pruneStructuralAncestorWrites"
  packages/runner/src`: no matches.
- Focused tests passed:
  - `test/scheduler-static-writes.test.ts`: `1 passed (2 steps)`.
  - `test/scheduler-ordering.test.ts`: `2 passed (17 steps)`.
  - `test/scheduler-observations.test.ts`: `1 passed (22 steps)`.
  - `test/scheduler-effects.test.ts`: `1 passed (20 steps)`.
  - `test/experimental-options.test.ts`: `1 passed (11 steps)`.

Full suite failure:

- `cd packages/runner && deno task test`: failed, `580 passed
  (3032 steps)`, `12 failed (63 steps)`, `0 ignored (10 steps)`, `2m5s`.
- Failure groups:
  - `test/cell-callbacks.test.ts`: persistent effect after `pull()` cleanup.
  - `test/memory-v2-pull-reactivity.test.ts`: 2 pull reactivity failures.
  - `test/scheduler-convergence.test.ts`: 6 convergence/cycle/stat failures.
  - `test/scheduler-core.test.ts`: 5 scheduler core/trace/cancel failures.
  - `test/scheduler-events.test.ts`: 2 event recomputation/in-flight demand
    failures.
  - `test/scheduler-pull-array.test.ts`: 7 array/demand/navigation failures.
  - `test/scheduler-pull-handlers.test.ts`: 7 handler dependency pulling
    failures.
  - `test/scheduler-pull-references.test.ts`: reference propagation failure.
  - `test/scheduler-pull.test.ts`: broad pull scheduling/staleness failures,
    including the explicit legacy assertion
    `should preserve writes when collecting dependencies from ReactivityLog`.
  - `test/scheduler-retries.test.ts`: retry dependency preservation failure.
  - `test/scheduler-throttle.test.ts`: 4 throttle/staleness failures.
  - `test/scheduler-timing.test.ts`: 7 debounce/auto-debounce/cycle-debounce
    failures.

Reviewer question: should the scheduler keep a compatibility path where a
direct `ReactivityLog` passed to `subscribe`/`resubscribe` seeds a static
surface for unannotated actions, or should the work order explicitly authorize
widening rewrites/annotations across the additional direct-scheduler tests?

## REVIEWER VERDICT — 05/step-3 subscribe-time logs are declarations

Keep the compatibility path — and not as a concession: it is the
P4-correct reading. A `ReactivityLog` passed to `subscribe(...)` is a
REGISTRATION-TIME declaration (the caller saying "this action reads X,
writes Y"), the same rank as runner annotations. What P4 forbids is the
surface changing from RUN logs. Do NOT widen test rewrites beyond the
three named files.

The rule:

1. Surface resolution at REGISTRATION, in priority order:
   annotated `action.writes` (non-empty) → else `immediateLog.writes`
   (subscribe-with-log path) → else empty. Applied once.
2. `resubscribe(action, runLog)` NEVER touches the surface — for
   annotated and unannotated actions alike. This is the only intended
   v1→v2 semantic change, and the three named files' rewrites already
   express it.
3. Architectural placement: move surface registration OUT of
   `setSchedulerDependencies` entirely — registration sites own it
   (`subscribePullSchedulerAction` for both the annotation and
   immediate-log cases; the rehydration path keeps using the annotation
   as you already have it). `setSchedulerDependencies` becomes
   reads/edges only and never writes to the write index. This matches
   the v2 component split (registration owns the surface) and makes the
   resubscribe path structurally unable to clobber it.
4. Populate-callback subscribers with no log and no annotation (e.g.
   `cell.pull()`'s ephemeral effect) keep an empty surface — correct,
   they are effects.

Expected effect: the direct-subscribe test population (the 12 failing
files) declares writes in their initial logs, so their surfaces, writer
edges, demand, and ordering come back without any test edits —
including the explicitly-legacy
`should preserve writes when collecting dependencies from ReactivityLog`
(subscribe-time, so it must pass unchanged) and the retry-preservation
test (retry resubscribes with the captured log; surface stays the
registration-time one).

After implementing: run the full suite. If failures remain, list them
per-test in PROGRESS with a one-line classification — (a) asserts
run-log surface EVOLUTION (rewrite authorized, name it), (b) anything
else (STOP with the failure). Do not bulk-rewrite.

Docs: amend WO05 step 2/3 with the declaration rule — own commit:
`docs(specs): scheduler-v2 WO05 — subscribe-time logs declare the surface`.
Deferred spec note recorded: P4's text should gain one sentence naming
the immediate-log declaration channel (rides the later docs change with
the settled-origin note).

## REVIEWER RESOLUTION — 05/step-3 declaration-rule docs

- [x] 13e6f09a7 — WO05 step 2/3 amended so registration-time `ReactivityLog`
  writes declare the static surface when annotations are absent, while
  resubscribe/run logs never change the surface.
- Recordings: docs-only commit; no `deno fmt`/`deno lint`/`deno check`
  required.

## IMPLEMENTER STOP — 05/step-3 declaration-rule implementation residuals

Implemented the reviewer verdict narrowly:

- `subscribePullSchedulerAction` resolves the registration-time surface once:
  non-empty annotated `action.writes` first, else `immediateLog.writes`, else
  empty; effects still do not register a scheduling surface.
- `setSchedulerDependencies` no longer reads `log.writes` or writes to the
  write index; it refreshes reads and returns the already-registered surface.
- `rehydrateActionFromObservation` now registers the live annotated surface
  before resubscribing reads, so persisted `currentKnownWrites` never restores
  the surface by itself.
- Named test rewrites stayed confined to `scheduler-ordering.test.ts`,
  `scheduler-observations.test.ts`, and `scheduler-effects.test.ts`.

Verification:

- `deno fmt packages/runner/src/scheduler/pull-subscriptions.ts
  packages/runner/src/scheduler/dependency-updates.ts
  packages/runner/src/scheduler/scheduling-writes.ts
  packages/runner/src/scheduler/dependency-graph.ts
  packages/runner/src/scheduler/action-run.ts
  packages/runner/src/scheduler/subscriptions.ts packages/runner/src/scheduler.ts
  packages/runner/src/runtime.ts packages/runner/test/experimental-options.test.ts
  packages/runner/test/scheduler-ordering.test.ts
  packages/runner/test/scheduler-observations.test.ts
  packages/runner/test/scheduler-effects.test.ts
  packages/runner/test/scheduler-static-writes.test.ts`: passed
  (`Checked 13 files`).
- `deno lint` on the same 13 files: passed (`Checked 13 files`).
- `deno check` on the same 13 files: passed.
- `grep -rn "backfillDependentsForNewWrites\|pruneDependentsForCurrentWrites"
  packages/runner/src/`: no matches.
- `grep -rn "schedulerHistoricalMightWrite\|historicalMightWrite" packages
  --include="*.ts"`: no matches.
- `grep -rn
  "buildKnownSchedulingWrites\|historicalMightWrite\|diffSchedulingWrites\|pruneStructuralAncestorWrites"
  packages/runner/src`: no matches.
- Focused tests passed:
  - `test/scheduler-static-writes.test.ts`: `1 passed (2 steps)`.
  - `test/scheduler-ordering.test.ts`: `2 passed (17 steps)`.
  - `test/scheduler-observations.test.ts`: `1 passed (22 steps)`.
  - `test/scheduler-effects.test.ts`: `1 passed (20 steps)`.
  - `test/experimental-options.test.ts`: `1 passed (11 steps)`.

Full suite failure:

- `cd packages/runner && deno task test`: failed, `588 passed
  (3091 steps)`, `4 failed (4 steps)`, `0 ignored (10 steps)`, `2m5s`.
- Per-test classification required by the reviewer verdict:
  - `test/scheduler-core.test.ts` / `captures exact action runs for one
    reactive update`: (b) not run-log surface evolution. The action-run trace
    still expects an effect's `declaredWrites` length to be 1, but effects now
    have no registered scheduling surface.
  - `test/scheduler-pull-handlers.test.ts` / `should wait for dynamically
    created lift before dispatching to downstream handler`: (a) asserts
    run-log surface evolution/setup. The test seeds the lift surface through
    `resubscribe(liftAction, log)`, which v2 now forbids.
  - `test/scheduler-pull.test.ts` / `should not re-run an effect for unrelated
    pending dependency collection`: (b) not run-log surface evolution. With
    effects now empty-surface demand roots, a child computation subscribed
    while the effect runs is demanded and runs once.
  - `test/scheduler-timing.test.ts` / `should auto-debounce slow writeful
    effects after threshold runs`: (b) not run-log surface evolution. The test
    classifies a writeful effect by its old scheduling writes; effects now have
    an empty scheduling surface and are treated as pull demand roots.

Stopped per the reviewer verdict because three residual failures are class
(b). No rewrites outside the three named files were attempted.

## REVIEWER VERDICT — 05/step-3 residuals (effect-surface proxy)

Effects being surface-less is CORRECT (spec §4.2: no scheduler-visible
output) and stays. The three class-(b) residuals are v1's
"writes==0 ⇒ pull-demand-root" PROXY breaking, not your change being
wrong. Per-test rulings:

1. `scheduler-core` / "captures exact action runs ...": authorized
   rewrite (add to named files): an effect's `declaredWrites` in the
   action-run trace is now 0 — that is the v2 truth, the test encoded
   the v1 run-learned surface.
2. `scheduler-pull-handlers` / dynamic-lift seeding: your class-(a)
   call is right — authorized rewrite, minimal form: seed the lift's
   surface through subscribe-with-log (the declaration channel) or an
   annotation, not post-run `resubscribe`.
3. `scheduler-pull` / "unrelated pending dependency collection": the
   new behavior IS spec §5.3 arriving early — a computation created
   during any LIVE node's run (every effect is live) gets one
   provisional first run. Authorized rewrite of this single test to the
   v2 expectation, with this comment above the changed assertion:
   `// v2 (spec §5.3): children created during a live effect's run get`
   `// provisional first-run demand; the v1 writeful-effect exception`
   `// keyed on run-learned writes, which no longer exist.`
4. `scheduler-timing` / auto-debounce: NO test rewrite — the mechanism
   must be fixed, or every ordinary sink loses auto-debounce (violates
   spec §8.2). Two edits:
   - `cell.ts` `pull()`: subscribe its ephemeral effect with
     `noDebounce: true` (the pull-root protection becomes EXPLICIT
     instead of riding the writes proxy);
   - the auto-debounce eligibility gate (`canAutomaticallyDebounce` in
     delay-control): drop the `isPullDemandRootEffect` exemption,
     keeping effect-only + `noDebounce` opt-out + thresholds.
   Keep `isPullDemandRootEffect` itself and its OTHER call sites
   (demand context, snapshot) unchanged in this phase — with effects
   surface-less it now returns true for all effects, which matches the
   §5.3 semantics accepted in (3); the predicate dissolves entirely in
   phase 3b.
   The failing timing test must then pass AS WRITTEN. If any other
   timing/throttle test fails after this (e.g. one asserting pull
   roots never debounce via the proxy), STOP with the name.

Bookkeeping: named-rewrite list for this work order now reads:
ordering, observations, effects, core (trace expectation), pull
(single §5.3 test), pull-handlers (seeding form). Docs amendment —
`docs(specs): scheduler-v2 WO05 — effect surfaces, pull-root opt-out, §5.3 early`
— before the code commit. Deferred spec notes now three: immediate-log
declaration channel (P4), settled-origin rule (§7.6), and §8.2's
exemption being noDebounce/pull-root-by-option rather than a writes
proxy.

Then: full suite green → commit step 2+3 (one commit, message from
step 3), then steps 4-6 as written.

## REVIEWER RESOLUTION — 05/step-3 residual-rule docs

- [x] 315a34866 — WO05 step 3 amended for effect-empty-surface trace
  expectations, subscribe-time seeding of dynamic lifts, early §5.3 provisional
  child demand, and explicit `noDebounce` pull-root auto-debounce opt-out.
- Recordings: docs-only commit; no `deno fmt`/`deno lint`/`deno check`
  required.

## IMPLEMENTER STOP — 05/step-3 residual-rule timing proxy follow-up

Applied the reviewer verdict narrowly:

- `cell.ts` `pull()` subscribes its ephemeral effect with `noDebounce: true`.
- Auto-debounce eligibility no longer exempts `isPullDemandRootEffect`; it keeps
  effect-only eligibility and explicit `noDebounce` opt-out.
- `scheduler-core.test.ts` action-run trace now expects effect
  `declaredWrites` to be empty while preserving the actual write assertion.
- `scheduler-pull-handlers.test.ts` seeds the dynamic lift surface through
  `subscribe(..., ReactivityLog)` rather than `resubscribe(...)`.
- `scheduler-pull.test.ts` rewrites the single §5.3 provisional child-demand
  expectation with the reviewer-specified comment.

Verification before STOP:

- `deno fmt` on the 20 touched TS/test files: passed (`Checked 20 files`).
- `deno lint` on the same 20 files: passed (`Checked 20 files`).
- `deno check` on the same 20 files: passed.
- Focused residual reruns:
  - `test/scheduler-core.test.ts`: `1 passed (25 steps)`.
  - `test/scheduler-pull-handlers.test.ts`: `1 passed (11 steps)`.
  - `test/scheduler-pull.test.ts`: `1 passed (26 steps)`.
  - `test/scheduler-timing.test.ts`: failed, `0 passed (20 steps)`,
    `1 failed (1 step)`.

STOP condition from reviewer verdict:

- `test/scheduler-timing.test.ts` / `should not auto-debounce write-less pull
  demand root effects` still asserts the old proxy behavior. After removing the
  `isPullDemandRootEffect` auto-debounce exemption, it observes debounce `100`
  instead of `undefined` at `scheduler-timing.test.ts:396`.

No rewrite was attempted because the reviewer explicitly said to STOP if any
other timing/throttle test fails after the auto-debounce cleanup.

## REVIEWER VERDICT — 05/step-3 timing-proxy companion test

Correct stop; this is the companion assertion of the same dead proxy,
from the other side. "Write-less effect" no longer means "pull root" —
it means an ordinary read-only sink, which is precisely what §8.2 wants
auto-debounced when slow. The pull-root protection is now the explicit
`noDebounce: true` that `pull()` sets.

Authorized rewrite of this single test (added to the named list):

1. Rename `should not auto-debounce write-less pull demand root effects`
   → `auto-debounces slow write-less effects; pull roots opt out via noDebounce`.
2. Flip the main assertion: after threshold runs, the bare write-less
   slow effect HAS debounce 100 (the value your run already observed).
3. In the same test (or a sibling `it` if cleaner), assert the
   protected case explicitly: an identical slow effect subscribed with
   `noDebounce: true` keeps debounce `undefined` after the same number
   of runs. Skip this half only if an existing test in the file already
   covers the `noDebounce` opt-out for effects — check first and note
   which.
4. Comment above the flipped assertion:
   `// v2 (spec §8.2): the auto-debounce exemption is the explicit`
   `// noDebounce opt-out (pull() sets it); v1 exempted any write-less`
   `// effect via the demand-root writes proxy, which no longer exists.`

Then full suite; if green, proceed with the commit sequence from the
prior verdict (steps 2+3 commit, then 4-6). Any further failure of any
shape: STOP as usual.

## IMPLEMENTER RESOLUTION — 05/step-3 timing companion rewrite

Applied the verdict:

- `scheduler-timing.test.ts` renamed the write-less effect test and now asserts
  that a bare slow write-less effect gets auto-debounce `100`.
- The same test asserts an identical slow effect subscribed with
  `noDebounce: true` keeps debounce `undefined`.
- Existing `noDebounce` coverage in this file was cycle-aware debounce only, so
  the explicit effect auto-debounce opt-out assertion was added here.

Verification:

- `deno fmt` on the 20 touched TS/test files: passed (`Checked 20 files`).
- `deno lint` on the same 20 files: passed (`Checked 20 files`).
- `deno check` on the same 20 files: passed.
- Contract greps:
  - `grep -rn "backfillDependentsForNewWrites\|pruneDependentsForCurrentWrites"
    packages/runner/src/`: no matches.
  - `grep -rn "schedulerHistoricalMightWrite\|historicalMightWrite" packages
    --include="*.ts"`: no matches.
  - `grep -rn
    "buildKnownSchedulingWrites\|historicalMightWrite\|diffSchedulingWrites\|pruneStructuralAncestorWrites"
    packages/runner/src`: no matches.
- Focused named rewrites passed:
  - `ENV=test deno test --allow-ffi --allow-env --allow-read
    --allow-write=/tmp,/var/folders --allow-run=git
    test/scheduler-static-writes.test.ts test/scheduler-ordering.test.ts
    test/scheduler-observations.test.ts test/scheduler-effects.test.ts
    test/experimental-options.test.ts`: `6 passed (72 steps)`,
    `0 failed`.
  - `ENV=test deno test --allow-ffi --allow-env --allow-read
    --allow-write=/tmp,/var/folders --allow-run=git
    test/scheduler-core.test.ts test/scheduler-pull-handlers.test.ts
    test/scheduler-pull.test.ts test/scheduler-timing.test.ts`:
    `4 passed (83 steps)`, `0 failed`.
- `cd packages/runner && deno task test`: `ok | 592 passed (3095 steps) |
  0 failed | 0 ignored (10 steps) (2m4s)`.

## 05/step-2+3

- [x] 6a9377bbe — scheduling writes are static declared surfaces registered at
  subscribe time.
- Deviations: reviewer-authorized residual rewrites documented above; effects
  have no scheduler-visible write surface, and pull-root debounce protection is
  the explicit `noDebounce` opt-out.
- Recordings: see the `IMPLEMENTER RESOLUTION — 05/step-3 timing companion
  rewrite` verification above.

## 05/step-4

- [x] no commit — clean removal held; no ordering-only effect write fallback
  was needed.
- Deviations: none.
- Recordings:
  - `cd packages/runner && ENV=test deno test --allow-ffi --allow-env
    --allow-read --allow-write=/tmp,/var/folders --allow-run=git
    test/scheduler-ordering.test.ts test/scheduler-convergence.test.ts`:
    passed, `3 passed (32 steps)`, `0 failed`.

## 05/step-5

- [x] 088100973 — added a dev warning for computation writes outside the
  declared static surface.
- Deviations: none.
- Recordings:
  - `deno fmt packages/runner/src/scheduler/action-run.ts
    packages/runner/test/scheduler-static-writes.test.ts`: passed
    (`Checked 2 files`).
  - `deno lint packages/runner/src/scheduler/action-run.ts
    packages/runner/test/scheduler-static-writes.test.ts`: passed
    (`Checked 2 files`).
  - `deno check packages/runner/src/scheduler/action-run.ts
    packages/runner/test/scheduler-static-writes.test.ts`: passed.
  - `cd packages/runner && ENV=test deno test --allow-ffi --allow-env
    --allow-read --allow-write=/tmp,/var/folders --allow-run=git
    test/scheduler-static-writes.test.ts`: passed, `1 passed (3 steps)`,
    `0 failed`.

## 05/phase-end

- [x] pending — WO05 exit checklist self-check complete.
- Deviations: none.
- Recordings:
  - `cd packages/runner && deno task test`: passed,
    `592 passed (3096 steps)`, `0 failed`, `0 ignored (10 steps)`, `2m4s`.
  - Phase benchmark command passed:
    `cd packages/runner && deno bench --allow-read --allow-write --allow-net
    --allow-ffi --allow-env --no-check test/scheduler.bench.ts
    test/scheduler-demand-roots.bench.ts
    test/scheduler-stale-propagation.bench.ts`.
  - Benchmark deltas vs 05/step-0 baseline:
    - `test/scheduler.bench.ts`:
      - `Scheduler - 100 computations, shared entity reads`: 19.4 ms -> 19.7
        ms (+1.5%).
      - `Scheduler - wide graph (1 source, 100 readers)`: 17.9 ms -> 17.3 ms
        (-3.4%).
      - `Scheduler - 100 entities, sparse deps`: 17.3 ms -> 17.7 ms (+2.3%).
      - `Scheduler - deep chain (50 levels)`: 11.5 ms -> 11.2 ms (-2.6%).
      - `Scheduler - diamond pattern (10 diamonds)`: 9.6 ms -> 9.7 ms
        (+1.0%).
      - `Scheduler - repeated dirty marking`: 7.7 ms -> 7.7 ms (+0.0%).
      - `Scheduler - subscribe/unsubscribe cycle (100x)`: 5.8 ms -> 5.6 ms
        (-3.4%).
      - `Scheduler - pull with resubscribe (50 pulls)`: 291.3 ms -> 288.4 ms
        (-1.0%).
      - `Overhead - setup/teardown only`: 1.3 ms -> 1.4 ms (+7.7%).
      - `Overhead - create 100 cells (getCell + set)`: 15.4 ms -> 15.5 ms
        (+0.6%).
      - `Overhead - 100x getCell only (no set)`: 1.6 ms -> 1.6 ms (+0.0%).
      - `Overhead - 100x set on existing cells`: 15.7 ms -> 15.5 ms (-1.3%).
      - `Overhead - runtime.idle() empty`: 1.3 ms -> 1.2 ms (-7.7%).
      - `Overhead - commit after 100 sets`: 15.9 ms -> 16.1 ms (+1.3%).
      - `Overhead - empty commit`: 1.3 ms -> 1.2 ms (-7.7%).
      - `Overhead - 100 raw tx.write + commit`: 7.1 ms -> 7.6 ms (+7.0%).
      - `Utility - sortAndCompactPaths (100 paths)`: 21.7 us -> 21.1 us
        (-2.8%).
      - `Utility - sortAndCompactPaths (1000 paths)`: 280.5 us -> 281.3 us
        (+0.3%).
      - `Utility - addressesToPathByEntity (100 paths)`: 14.9 us -> 14.8 us
        (-0.7%).
      - `Utility - addressesToPathByEntity (1000 paths)`: 150.3 us -> 152.1
        us (+1.2%).
      - `Scheduler - bare subscribe (100x)`: 1.7 ms -> 1.6 ms (-5.9%).
      - `Scheduler - subscribe 100 actions reading same entity`: 1.7 ms -> 1.6
        ms (-5.9%).
      - `Scheduler - resubscribe cycle (100x)`: 1.4 ms -> 1.3 ms (-7.1%).
    - `test/scheduler-demand-roots.bench.ts`:
      - `Scheduler demand roots - effect demand root`: 147.1 ms -> 144.1 ms
        (-2.0%).
      - `Scheduler demand roots - event demand root`: 131.8 ms -> 129.4 ms
        (-1.8%).
      - `Scheduler demand roots - mixed effect and event roots`: 169.0 ms ->
        172.8 ms (+2.2%).
      - `Scheduler demand roots - parent clears generated children`: 81.3 ms ->
        82.6 ms (+1.6%).
    - `test/scheduler-stale-propagation.bench.ts`:
      - `Scheduler stale propagation - chain`: 105.5 ms -> 97.3 ms (-7.8%).
      - `Scheduler stale propagation - diamond`: 94.7 ms -> 89.4 ms (-5.6%).
      - `Scheduler stale propagation - wide fanout`: 244.0 ms -> 248.0 ms
        (+1.6%).
      - `Scheduler stale propagation - dynamic deps`: 82.9 ms -> 74.7 ms
        (-9.9%).
      - `Scheduler stale propagation - unchanged recompute`: 74.7 ms -> 73.2
        ms (-2.0%).
  - No benchmark regression exceeded 10%.
- Exit checklist:
  - `grep -rn
    "buildKnownSchedulingWrites\|historicalMightWrite\|diffSchedulingWrites\|pruneStructuralAncestorWrites"
    packages/runner/src`: no matches.
  - `grep -rn "log\.writes"
    packages/runner/src/scheduler/dependency-updates.ts`: no matches;
    inspection confirms `setSchedulerDependencies` reads only `log.reads` and
    `log.shallowReads`, and returns `state.writeIndex.getSchedulingWrites`.
  - Observation payload compatibility held: `attachSchedulerActionObservation`
    still sets both `currentKnownWrites` and `declaredWrites` to the registered
    static surface.
  - Static write fixtures A/B and the new surface-violation fixture are green:
    `test/scheduler-static-writes.test.ts` passed, `1 passed (3 steps)`,
    `0 failed`.
  - Test rewrites were confined to the reviewer-approved named files; step 4
    clean removal held without the decision-tree fallback.

---
## Summary by cubic
Make scheduler writes a static surface fixed at subscribe time for deterministic scheduling. Persist the live surface through rehydration and add a debug-level diagnostic for out-of-surface writes (scoped-slot writes ignored).

- **Refactors**
  - Resolve writer surface once at subscription: from `action.writes` (if non-empty), else from the subscribe-time `ReactivityLog`, else empty.
  - Remove dynamic write discovery/history and related diff/backfill/prune paths; drop `schedulerHistoricalMightWrite`.
  - Attach writer edges at registration; `setSchedulerDependencies` now refreshes reads only.
  - Effects register no scheduler-visible writes; auto-debounce applies to effects unless opted out with `noDebounce`.
  - `cell.pull()` subscribes its ephemeral effect with `noDebounce: true`.
  - Persist and rehydrate the live registered surface: observations store the surface; rehydration resolves it like registration (annotation first, else persisted).
  - Add a debug-level diagnostic when a run writes outside its declared surface; ignore scoped-slot writes; logger counters still increment.

- **Migration**
  - Declare writes via `action.writes` or by passing a `ReactivityLog` to `subscribe(action, log)`.
  - Stop using `resubscribe(action, runLog)` to seed or change the surface.
  - For pull roots that must not debounce, subscribe with `{ noDebounce: true }`.
  - If you see the diagnostic, align your declared surface with actual writes.

<sup>Written for commit e7bae94ce4e54e7832a384cbbd20e11bf7f58d57. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4098?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/scheduler-v2/implementation/05-phase1-static-write-surface.md
+++ b/docs/specs/scheduler-v2/implementation/05-phase1-static-write-surface.md
@@ -187,9 +187,31 @@ justification: "asserted v1 write-set learning; v2 surface is static"):
   surface.
 - `test/scheduler-effects.test.ts:451-459` — membership-style check;
   likely passes unchanged; verify.
+- `test/scheduler-core.test.ts` / action-run trace expectation — effects now
+  have no scheduler-visible output, so trace `declaredWrites` is empty for
+  effects.
+- `test/scheduler-pull-handlers.test.ts` / dynamic lift seeding — seed the
+  computation's static surface through subscribe-with-log (the registration
+  declaration channel) or an annotation, not post-run `resubscribe`.
+- `test/scheduler-pull.test.ts` / unrelated pending dependency collection —
+  v2 §5.3 arrives early here: computations created during a live effect's run
+  get provisional first-run demand; the v1 exception keyed on run-learned
+  effect writes no longer exists.
+
+Auto-debounce cleanup in the same commit:
+
+1. `src/cell.ts` `pull()`: subscribe the ephemeral pull-root effect with
+   `noDebounce: true`. Pull-root debounce protection is explicit rather than
+   inferred from a write-surface proxy.
+2. The auto-debounce eligibility gate (`canAutomaticallyDebounce` in
+   delay-control): remove the `isPullDemandRootEffect` exemption. Keep the
+   effect-only requirement, explicit `noDebounce` opt-out, and existing timing
+   thresholds. Keep `isPullDemandRootEffect` itself and its other call sites
+   unchanged in this phase.
 
 Failures in ANY other file: apply the decision tree in step 4 if they are
-ordering-related; otherwise STOP.
+ordering-related. If timing/throttle tests fail after the auto-debounce cleanup,
+STOP and list the failing names. Otherwise STOP.
 
 Commit: `refactor(runner): scheduling writes are the static declared surface (scheduler-v2 phase 1)`
 
@@ -248,6 +270,8 @@ better; regressions >10%: STOP).
 - [ ] Observation payload still carries both write fields (compat), equal
       to the surface.
 - [ ] Fixtures A/B green; behavior change (if any) documented.
-- [ ] Test rewrites confined to the three named files (or decision-tree
-      fallback applied and marked).
+- [ ] Test rewrites confined to the named files: ordering, observations,
+      effects, core trace expectation, pull single §5.3 test, and
+      pull-handlers seeding form (or decision-tree fallback applied and
+      marked).
 - [ ] Bench deltas recorded; none worse than 10%.
--- a/docs/specs/scheduler-v2/implementation/05-phase1-static-write-surface.md
+++ b/docs/specs/scheduler-v2/implementation/05-phase1-static-write-surface.md
@@ -60,27 +60,29 @@ becomes the unconditional surface registration:
 ```typescript
 // Static write surface (spec scheduler-v2 P4): the action's writes are
 // fixed at registration — declared outputs plus statically resolved
-// redirect targets, already computed by the runner. Nothing about the
+// redirect targets, already computed by the runner, or a registration-time
+// ReactivityLog supplied by direct scheduler callers. Nothing about the
 // surface is learned from runs.
-const surface = (action as Partial<TelemetryAnnotations>).writes ?? [];
+const annotatedSurface = (action as Partial<TelemetryAnnotations>).writes ?? [];
+const surface = annotatedSurface.length > 0
+  ? annotatedSurface.map(toMemorySpaceAddress)
+  : (immediateLog?.writes ?? []);
 if (!actionIsEffect && surface.length > 0) {
-  setSchedulerDependencies(
-    state.dependencyUpdateState,
-    action,
-    {
-      reads: [],
-      shallowReads: [],
-      writes: surface.map(toMemorySpaceAddress),
-    },
-  );
+  state.writeIndex.setSurface(action, surface);
+  state.registerWriterDependents(action, surface);
 }
 ```
 
 (Identical to today's block minus the `!immediateLog` condition and with
 the comment replaced; with an `immediateLog` the subsequent
 `setSchedulerDependencies(state..., immediateLog)` call now must NOT
-clobber the surface — that is handled by step 3's rewrite, which ignores
-log writes. Order the calls so the surface registration runs first.)
+clobber the surface — that is handled by step 3's rewrite, which keeps
+surface registration out of dependency updates. Order the calls so the
+surface registration runs first.)
+
+Surface resolution is a registration-time declaration, in priority order:
+non-empty annotated `action.writes`, then `immediateLog.writes`, then empty.
+`resubscribe(action, runLog)` never changes the surface.
 
 Commit with step 3 (single commit; this step alone is incoherent).
 
@@ -101,28 +103,20 @@ export function setSchedulerDependencies(
 } {
   const reads = sortAndCompactPaths(log.reads);
   const shallowReads = sortAndCompactPaths(log.shallowReads, false);
-  const ignoredSchedulingWrites =
-    (action as Partial<TelemetryAnnotations>).ignoredSchedulingWrites ?? [];
-  // Static write surface (spec scheduler-v2 P4): scheduling writes are the
-  // declared surface, never the transaction's observed writes.
-  const surface = sortAndCompactPaths(
-    filterIgnoredAddresses(
-      ((action as Partial<TelemetryAnnotations>).writes ?? []).map(
-        toMemorySpaceAddress,
-      ),
-      ignoredSchedulingWrites,
-    ),
-  );
   const schedulingLog: ReactivityLog = {
     reads,
     shallowReads,
-    writes: surface,
+    writes: state.writeIndex.getSchedulingWrites(action) ?? [],
   };
   state.dependencies.set(action, schedulingLog);
-  state.writeIndex.setSurface(action, surface);
   return { reads, shallowReads, log: schedulingLog };
 }
 ```
+
+`setSchedulerDependencies` must not register, update, derive, or rederive the
+write surface. Registration sites own `state.writeIndex.setSurface(...)`;
+dependency updates only refresh reads and return the already-registered
+scheduling writes for compatibility with existing graph/telemetry payloads.
 
 Deletions in the same commit:
 
--- a/docs/specs/scheduler-v2/implementation/PROGRESS.md
+++ b/docs/specs/scheduler-v2/implementation/PROGRESS.md
@@ -2342,8 +2342,139 @@ the settled-origin note).
 
 ## REVIEWER RESOLUTION — 05/step-3 declaration-rule docs
 
-- [x] pending — WO05 step 2/3 amended so registration-time `ReactivityLog`
+- [x] 13e6f09a7 — WO05 step 2/3 amended so registration-time `ReactivityLog`
   writes declare the static surface when annotations are absent, while
   resubscribe/run logs never change the surface.
+- Recordings: docs-only commit; no `deno fmt`/`deno lint`/`deno check`
+  required.
+
+## IMPLEMENTER STOP — 05/step-3 declaration-rule implementation residuals
+
+Implemented the reviewer verdict narrowly:
+
+- `subscribePullSchedulerAction` resolves the registration-time surface once:
+  non-empty annotated `action.writes` first, else `immediateLog.writes`, else
+  empty; effects still do not register a scheduling surface.
+- `setSchedulerDependencies` no longer reads `log.writes` or writes to the
+  write index; it refreshes reads and returns the already-registered surface.
+- `rehydrateActionFromObservation` now registers the live annotated surface
+  before resubscribing reads, so persisted `currentKnownWrites` never restores
+  the surface by itself.
+- Named test rewrites stayed confined to `scheduler-ordering.test.ts`,
+  `scheduler-observations.test.ts`, and `scheduler-effects.test.ts`.
+
+Verification:
+
+- `deno fmt packages/runner/src/scheduler/pull-subscriptions.ts
+  packages/runner/src/scheduler/dependency-updates.ts
+  packages/runner/src/scheduler/scheduling-writes.ts
+  packages/runner/src/scheduler/dependency-graph.ts
+  packages/runner/src/scheduler/action-run.ts
+  packages/runner/src/scheduler/subscriptions.ts packages/runner/src/scheduler.ts
+  packages/runner/src/runtime.ts packages/runner/test/experimental-options.test.ts
+  packages/runner/test/scheduler-ordering.test.ts
+  packages/runner/test/scheduler-observations.test.ts
+  packages/runner/test/scheduler-effects.test.ts
+  packages/runner/test/scheduler-static-writes.test.ts`: passed
+  (`Checked 13 files`).
+- `deno lint` on the same 13 files: passed (`Checked 13 files`).
+- `deno check` on the same 13 files: passed.
+- `grep -rn "backfillDependentsForNewWrites\|pruneDependentsForCurrentWrites"
+  packages/runner/src/`: no matches.
+- `grep -rn "schedulerHistoricalMightWrite\|historicalMightWrite" packages
+  --include="*.ts"`: no matches.
+- `grep -rn
+  "buildKnownSchedulingWrites\|historicalMightWrite\|diffSchedulingWrites\|pruneStructuralAncestorWrites"
+  packages/runner/src`: no matches.
+- Focused tests passed:
+  - `test/scheduler-static-writes.test.ts`: `1 passed (2 steps)`.
+  - `test/scheduler-ordering.test.ts`: `2 passed (17 steps)`.
+  - `test/scheduler-observations.test.ts`: `1 passed (22 steps)`.
+  - `test/scheduler-effects.test.ts`: `1 passed (20 steps)`.
+  - `test/experimental-options.test.ts`: `1 passed (11 steps)`.
+
+Full suite failure:
+
+- `cd packages/runner && deno task test`: failed, `588 passed
+  (3091 steps)`, `4 failed (4 steps)`, `0 ignored (10 steps)`, `2m5s`.
+- Per-test classification required by the reviewer verdict:
+  - `test/scheduler-core.test.ts` / `captures exact action runs for one
+    reactive update`: (b) not run-log surface evolution. The action-run trace
+    still expects an effect's `declaredWrites` length to be 1, but effects now
+    have no registered scheduling surface.
+  - `test/scheduler-pull-handlers.test.ts` / `should wait for dynamically
+    created lift before dispatching to downstream handler`: (a) asserts
+    run-log surface evolution/setup. The test seeds the lift surface through
+    `resubscribe(liftAction, log)`, which v2 now forbids.
+  - `test/scheduler-pull.test.ts` / `should not re-run an effect for unrelated
+    pending dependency collection`: (b) not run-log surface evolution. With
+    effects now empty-surface demand roots, a child computation subscribed
+    while the effect runs is demanded and runs once.
+  - `test/scheduler-timing.test.ts` / `should auto-debounce slow writeful
+    effects after threshold runs`: (b) not run-log surface evolution. The test
+    classifies a writeful effect by its old scheduling writes; effects now have
+    an empty scheduling surface and are treated as pull demand roots.
+
+Stopped per the reviewer verdict because three residual failures are class
+(b). No rewrites outside the three named files were attempted.
+
+## REVIEWER VERDICT — 05/step-3 residuals (effect-surface proxy)
+
+Effects being surface-less is CORRECT (spec §4.2: no scheduler-visible
+output) and stays. The three class-(b) residuals are v1's
+"writes==0 ⇒ pull-demand-root" PROXY breaking, not your change being
+wrong. Per-test rulings:
+
+1. `scheduler-core` / "captures exact action runs ...": authorized
+   rewrite (add to named files): an effect's `declaredWrites` in the
+   action-run trace is now 0 — that is the v2 truth, the test encoded
+   the v1 run-learned surface.
+2. `scheduler-pull-handlers` / dynamic-lift seeding: your class-(a)
+   call is right — authorized rewrite, minimal form: seed the lift's
+   surface through subscribe-with-log (the declaration channel) or an
+   annotation, not post-run `resubscribe`.
+3. `scheduler-pull` / "unrelated pending dependency collection": the
+   new behavior IS spec §5.3 arriving early — a computation created
+   during any LIVE node's run (every effect is live) gets one
+   provisional first run. Authorized rewrite of this single test to the
+   v2 expectation, with this comment above the changed assertion:
+   `// v2 (spec §5.3): children created during a live effect's run get`
+   `// provisional first-run demand; the v1 writeful-effect exception`
+   `// keyed on run-learned writes, which no longer exist.`
+4. `scheduler-timing` / auto-debounce: NO test rewrite — the mechanism
+   must be fixed, or every ordinary sink loses auto-debounce (violates
+   spec §8.2). Two edits:
+   - `cell.ts` `pull()`: subscribe its ephemeral effect with
+     `noDebounce: true` (the pull-root protection becomes EXPLICIT
+     instead of riding the writes proxy);
+   - the auto-debounce eligibility gate (`canAutomaticallyDebounce` in
+     delay-control): drop the `isPullDemandRootEffect` exemption,
+     keeping effect-only + `noDebounce` opt-out + thresholds.
+   Keep `isPullDemandRootEffect` itself and its OTHER call sites
+   (demand context, snapshot) unchanged in this phase — with effects
+   surface-less it now returns true for all effects, which matches the
+   §5.3 semantics accepted in (3); the predicate dissolves entirely in
+   phase 3b.
+   The failing timing test must then pass AS WRITTEN. If any other
+   timing/throttle test fails after this (e.g. one asserting pull
+   roots never debounce via the proxy), STOP with the name.
+
+Bookkeeping: named-rewrite list for this work order now reads:
+ordering, observations, effects, core (trace expectation), pull
+(single §5.3 test), pull-handlers (seeding form). Docs amendment —
+`docs(specs): scheduler-v2 WO05 — effect surfaces, pull-root opt-out, §5.3 early`
+— before the code commit. Deferred spec notes now three: immediate-log
+declaration channel (P4), settled-origin rule (§7.6), and §8.2's
+exemption being noDebounce/pull-root-by-option rather than a writes
+proxy.
+
+Then: full suite green → commit step 2+3 (one commit, message from
+step 3), then steps 4-6 as written.
+
+## REVIEWER RESOLUTION — 05/step-3 residual-rule docs
+
+- [x] pending — WO05 step 3 amended for effect-empty-surface trace
+  expectations, subscribe-time seeding of dynamic lifts, early §5.3 provisional
+  child demand, and explicit `noDebounce` pull-root auto-debounce opt-out.
 - Recordings: docs-only commit; no `deno fmt`/`deno lint`/`deno check`
   required.
--- a/docs/specs/scheduler-v2/implementation/PROGRESS.md
+++ b/docs/specs/scheduler-v2/implementation/PROGRESS.md
@@ -2187,7 +2187,7 @@ packages/runner/src/scheduler/events.ts:654:            permanent: isPermanentRe
 
 ## 05/step-1
 
-- [x] pending — static write surface demand fixtures added.
+- [x] ca77bc16e — static write surface demand fixtures added.
 - Deviations: both fixtures already pass on current code, so they pin existing
   behavior; no behavior-change red case was observed.
 - Recordings:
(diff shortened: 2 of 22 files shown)


seefeldb:
Reviewer pass against the WO05 exit checklist — approved.

- **Static surface holds structurally**: `setSchedulerDependencies` has zero references to `log.writes`; `setSurface` is called from exactly the two registration sites (subscribe + rehydration), so the resubscribe path is structurally unable to evolve a surface. Discovery machinery greps (buildKnownSchedulingWrites / historical / diff / prune) all zero; `schedulerHistoricalMightWrite` flag gone repo-wide.
- **Two design refinements emerged through stops and are recorded in-order**: (1) a `ReactivityLog` passed at *subscribe time* is a registration-time declaration channel, same rank as annotations — this kept the raw-API test population working with zero rewrites beyond the named files; (2) effects are surface-less per spec §4.2, which dissolved v1's "writes==0 ⇒ pull demand root" proxy — pull-root auto-debounce protection is now the explicit `noDebounce: true` set by `pull()`, ordinary slow sinks correctly auto-debounce again (§8.2), and one §5.3 behavior (provisional first-run demand for children of live effects) lands early with a marker comment.
- Step-4 contingency resolved clean: ordering tests passed without any effect-write toposort fallback.
- Bench gates: all deltas within ±8% noise band, stale-propagation improved (-5..8%); suite 592 green; surface-conformance warning in place (computations only, envelope-aware).

Deferred spec notes now queued for one docs PR: P4 declaration channel, §7.6 settled-origin rule, §8.2 exemption semantics.

seefeldb:
Both surface-persistence findings fixed red-first: actions whose static surface came from `subscribe(action, ReactivityLog)` (no `.writes` annotation) persisted `currentKnownWrites: []` and rehydrated with no surface. Observations now persist the live registered surface (`getSchedulingWrites(action) ?? declaredWrites`), and `rehydrateActionFromObservation` resolves through `resolveRegistrationSurface(action, { writes: observation.currentKnownWrites })` — annotation first, persisted surface otherwise, mirroring registration. Test: "persists and rehydrates immediate-log write surfaces" (red: both assertions returned []). Rebased onto main; runner 616 green locally.

seefeldb:
**Reviewer pass (rebase + new fix):** range-diff of the original series (`a49811102..606cf649b`) vs the rebased one shows commits 2–6 byte-identical and commit 1 differing only in PROGRESS.md context lines — zero code drift in the rebase itself.

`805e586b2` (immediate-log surfaces survive persistence) — verified, and it closes a real interaction gap between two scheduler-v2 decisions: actions whose surface comes from subscribe's ReactivityLog (the WO05 declaration channel, no `.writes` annotation) persisted an empty `currentKnownWrites` and rehydrated surface-less. The fix is symmetric in the right way: persistence captures the **live registered surface** (`getSchedulingWrites(action) ?? declaredWrites`), and rehydration feeds the persisted surface through the same `resolveRegistrationSurface` used at registration, so the annotation-first rule holds in one place across registration, persistence, and rehydration. Catching this at phase 1 (warn-only) rather than phase 4 (enforced static surface) is exactly when you'd want it. Test included. LGTM pending CI.

seefeldb:
CI surfaced one more phase-1 interaction: the new write-surface-violation diagnostic flagged per-user/per-session scoped-slot writes (scope-default initialization / UI state — runtime-mediated, not author-declarable), and cf test now fails on console warnings, breaking scrabble.test.tsx (14 warnings, all `user`/`session` scoped, zero space-scoped). Fixed red-first: the diagnostic exempts non-space scopes with a WATCH to re-include them once scoped-slot writes are declarable (test: "does not warn for scoped-slot writes outside the declared surface"; the existing space-scoped warn test still passes). Runner 616 green, scrabble fixture green locally.

seefeldb:
Addendum for `03f7cf3b6` (scoped-slot exemption) — verified: diagnostic-only (`continue` in the warn path), correctly scoped to non-`space` cell scopes, WATCH marker for when scoped slots become declarable. Right call: per-user/per-session slot writes are runtime-mediated (scope-default init, UI state) and authored surfaces cannot declare them today, so warning on them is noise, and under #4114's fail-on-console-warnings they were CI failures (scrabble ×14).

Forward-looking caution for the rest of the train: this is the second #4114-strictness collision (after the E1 readTx one), and the final-cleanup PROGRESS notes list more "expected noisy passing logs" (write-surface warnings in wish/raw-map fixtures, scheduler preflight/retry logs). The wish/raw-map ones are likely this same diagnostic and may already be covered by this exemption; if later retargets hit new ones, the fix pattern is the same — tighten the diagnostic's scope rather than blanket-allow console output in fixtures.

seefeldb:
Second first-contact CI round: the diagnostic also fired for builtins that mint cause-keyed internal result docs inside their run (`ifElse`/`unless`/`fetchData`/… — calendar fixtures red with space-scoped violations; the internal doc is only declarable after a first run resolves the redirect, and 10+ builtins share the shape). Per the work order this check is "declaration-gap diagnostics, not enforcement" — and the cf-test warning gate postdates it — so it now logs at **debug** level (counts still increment regardless of level; the conformance tests assert `getLoggerCountsBreakdown().scheduler["write-surface-violation"].debug`). Scoped-slot exemption retained for signal cleanliness. Calendar + scrabble fixtures and runner 616 green locally. The Pattern Integration (3/4) failure was the known group-chat #trusted-profile-status flake (passed on rerun on #4088); will rerun if it recurs.

seefeldb:
Verdict on `e7bae94ce` (debug-level demotion) — **accept**, and the shape is right: the diagnostic moved from console (where #4114 strictness makes every legit-but-undeclarable writer a CI failure) to logger counters that increment regardless of level, with conformance tests asserting `getLoggerCountsBreakdown()` — the tripwire survives, just not as console noise. WO05 step 5 indeed framed this as diagnostics, not enforcement.

One structural follow-up worth a WATCH/spec note: the builtin class (ifElse/unless/fetchData minting cause-keyed internal docs mid-run) is exactly what the spec's P4 calls **declared materializer write envelopes** — the diagnostic firing on them says those envelopes aren't declared/threaded for builtins yet. That's the real fix that would let this diagnostic tighten back up later; worth recording so the debug demotion doesn't silently become the permanent posture.

PR #4100, opened 2026-06-12
refactor(runner): scheduler-v2 cutover 3pre: fixture pack

seefeldb:
Stack: 07/3pre - based on scheduler-v2/06-phase2 (#4099)

3pre commit: 4a607da47

PROGRESS excerpt:

## 07/3pre

- [x] pending — added the scheduler-v2 cutover fixture pack.
- Deviations:
  - The cycle-backoff fixture is green on the current v1 cycle-break path with
    a loose bounded-termination assertion (`runCountA + runCountB < 500`) and
    unrelated-subgraph convergence. The focused run emits the expected current
    `Too many iterations: 101` scheduler errors for the cycling actions. The
    fixture comment records that 3c.iv tightens this to the v2
    `PASS_RUN_BUDGET` backoff rule.
- Recordings:
  - `deno fmt packages/runner/test/scheduler-v2-cutover.test.ts`: passed
    (`Checked 1 file`).
  - `deno lint packages/runner/test/scheduler-v2-cutover.test.ts`: passed
    (`Checked 1 file`).
  - `deno check packages/runner/test/scheduler-v2-cutover.test.ts`: passed.
  - Focused fixture pack:
    `cd packages/runner && ENV=test deno test --allow-ffi --allow-env
    --allow-read --allow-write=/tmp,/var/folders --allow-run=git
    test/scheduler-v2-cutover.test.ts`: passed, `1 passed (7 steps)`,
    `0 failed`.
  - `cd packages/runner && deno task test`: passed,
    `594 passed (3106 steps)`, `0 failed`, `0 ignored (10 steps)`, `2m6s`.

---
## Summary by cubic
Adds the scheduler-v2 cutover fixture pack in `packages/runner/test/scheduler-v2-cutover.test.ts` and updates `docs/specs/scheduler-v2/implementation/PROGRESS.md` for 07/3pre. The fixtures pin expected cutover behavior (keeping the v1 cycle-break bound for now) across conditional rewiring, parent continuation from child updates, effect triggering rules (no-op vs changed), cycle bounding without blocking unrelated subgraphs, provisional demand retention for parent-created children, and dormant writers.

<sup>Written for commit cc9abadc2beaaed52841f39e542bd7684b6ad9a0. Summary will update on new commits.</sup>

[...]

PR #4101, opened 2026-06-12
refactor(runner): scheduler-v2 cutover 3a: node records

seefeldb:
Stack: 07/3a - based on scheduler-v2/07-3pre (#4100)

## What changed

- Introduces SchedulerNode records for effect/computation kind tracking.
- Migrates scheduler parent/child creation context from WeakMaps to NodeRegistry.
- Preserves v1 parent-link semantics by storing parent and child edges by Action object and resolving records lazily.
- Extends the cutover fixture pack with parent-link registration semantics, including parent-not-yet-registered capture.

## Why

This is the 3a node-record sub-order for scheduler-v2 phase 3. The final reviewer-directed fix addresses the CT-1316 topology regression where a parent link was refused because the parent record was not registered at link time, even though v1 captured that Action parent in a WeakMap.

## Validation

- deno fmt on 12 touched scheduler/test files
- deno lint on 12 touched scheduler/test files
- deno check on 12 touched scheduler/test files
- ENV=test deno test --allow-ffi --allow-env --allow-read --allow-write=/tmp,/var/folders --allow-run=git test/scheduler-v2-cutover.test.ts
- ENV=test deno test --allow-ffi --allow-env --allow-read --allow-write=/tmp,/var/folders --allow-run=git test/patterns-derive-return-pattern.test.ts
- cd packages/runner && deno task test: 594 passed (3107 steps), 0 failed, 0 ignored (10 steps), 2m6s

See docs/specs/scheduler-v2/implementation/PROGRESS.md for the reviewer probe branch, scratch cleanup, and grep contracts.

---
## Summary by cubic
Replaces WeakMap-based scheduler tracking with a `NodeRegistry` of `SchedulerNode` records keyed by action objects. Restores v1 parent-link semantics, fixes CT-1316’s spurious parent rerun, and allows computation→effect promotion on re-registration.

- **Refactors**
  - Add `packages/runner/src/scheduler/node-record.ts` with `NodeRegistry` and `SchedulerNode`.
  - Move effect/computation membership and kind to the registry; consumers read `nodes.effects` / `nodes.computations`.
  - Replace `actionParent` / `actionChildren` with `nodes.parentOf` / `nodes.childrenOf` / `nodes.isAncestor`; add `nodes.parentActionOf` for raw Action-based lookup parity.
  - Thread `nodes` through scheduler state; remove `isEffectAction` WeakMap.
  - Update topology, demand, write-propagation, subscriptions, execution, pull, graph snapshot, and diagnostics to use the registry.
  - Tests: switch internal helpers to `registerEffect(...)`.
  - Docs: record scheduler-v2 phase 3 baseline in `docs/specs/scheduler-v2/implementation/PROGRESS.md`.

- **Bug Fixes**
  - Parent-linking matches v1: no executing parent → no-op; resubscribe (`allowExisting: false`) preserves parent; subscribe (`allowExisting: true`) overwrites parent; links captured even if the parent isn’t registered yet.
  - Add `parentActionOf(...)` to preserve early-captured parent links at Action granularity; demand and trace call sites use it.
  - Support v1-sanctioned computation→effect promotion on re-registration; still reject demotion.
  - Cutover fixture pins these rules; CT-1316 passes; full suite remains green.

<sup>Written for commit 89e916d7699eca777dba2ce324eb04f4c06c359b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4101?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/scheduler-v2/implementation/PROGRESS.md
+++ b/docs/specs/scheduler-v2/implementation/PROGRESS.md
@@ -2893,3 +2893,56 @@ Verification:
     - `Scheduler persistent state - clean rehydrate 1000 actions`: 8.2 ms.
     - `Scheduler persistent state - targeted dirty rehydrate 1000 actions`:
       7.6 ms.
+
+## 07/3a-kind
+
+- [x] pending — migrated effect/computation kind tracking to
+  `SchedulerNode` records.
+- Deviations:
+  - None.
+- Choices:
+  - `NodeRegistry` owns action kind registration and the sticky historical
+    effect/computation distinction. Re-registering a known action with a
+    different kind is a hard error.
+  - Existing state bundles that only read active effect/computation membership
+    continue to receive read-only `nodes.effects` / `nodes.computations` views,
+    keeping this family behavior-preserving and tightly scoped.
+  - Mutable classification paths (`updateSchedulerActionType`,
+    `clearActionTypeTracking`) now go through `NodeRegistry`.
+  - Historical-kind checks that previously used `isEffectAction` now use
+    `nodes.isKnownEffect(...)`; test internals expose `registerEffect(...)`
+    instead of the old WeakMap.
+- Recordings:
+  - Contract grep: `grep -rn
+    "isEffectAction\|this\.effects\b\|this\.computations\b"
+    packages/runner/src packages/runner/test/scheduler-test-utils.ts
+    packages/runner/test/scheduler-pull.test.ts`: no matches.
+  - `deno fmt packages/runner/src/scheduler.ts
+    packages/runner/src/scheduler/subscriptions.ts
+    packages/runner/src/scheduler/demand.ts
+    packages/runner/src/scheduler/action-run.ts
+    packages/runner/src/scheduler/node-record.ts
+    packages/runner/test/scheduler-test-utils.ts
+    packages/runner/test/scheduler-pull.test.ts`: passed (`Checked 7 files`).
+  - `deno lint packages/runner/src/scheduler.ts
+    packages/runner/src/scheduler/subscriptions.ts
+    packages/runner/src/scheduler/demand.ts
+    packages/runner/src/scheduler/action-run.ts
+    packages/runner/src/scheduler/node-record.ts
+    packages/runner/test/scheduler-test-utils.ts
+    packages/runner/test/scheduler-pull.test.ts`: passed (`Checked 7 files`).
+  - `deno check packages/runner/src/scheduler.ts
+    packages/runner/src/scheduler/subscriptions.ts
+    packages/runner/src/scheduler/demand.ts
+    packages/runner/src/scheduler/action-run.ts
+    packages/runner/src/scheduler/node-record.ts
+    packages/runner/test/scheduler-test-utils.ts
+    packages/runner/test/scheduler-pull.test.ts`: passed.
+  - Focused scheduler gate:
+    `cd packages/runner && ENV=test deno test --allow-ffi --allow-env
+    --allow-read --allow-write=/tmp,/var/folders --allow-run=git
+    test/scheduler-effects.test.ts test/scheduler-pull.test.ts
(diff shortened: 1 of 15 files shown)


seefeldb:
Reviewer pass — approved. This sub-order earned its three stops; for posterity, the CT-1316 regression chain and its resolution:

1. Migrating creation-context to node records initially broke `patterns-derive-return-pattern` (parent derive reran when only its returned child changed). Two reviewer-directed restorations (literal v1 `registerParentChildAction` semantics — parentless calls no-op, resubscribe preserves, subscribe overwrites) were necessary but insufficient.
2. A consumer bisect (parallel legacy WeakMap, one consumer reverted at a time) isolated the **topology parent lookup** as the sole culprit, with the exact differing value: registry `none` vs WeakMap `sink` for the old child — losing the parent-first tie-break let the stale child run before the sink's cancel chain landed; its commit re-dirtied the parent's read.
3. A branch probe identified the refusal: `linkParent` required the PARENT's record to exist, and during re-instantiation churn the new sink's registration interleaves after the child's link attempt (object serials 24/25 on record). v1's WeakMap never had that precondition.

**Fix**: parent/child edges are keyed and valued by ACTION OBJECTS (`parentAction`/`children: Set<Action>`), resolved to records lazily in `parentOf`/`childrenOf`/`isAncestor` — exact WeakMap parity, single assignment site, with a new fixture case pinning "parent not yet registered at link time → link still captured".

Gates: 8-case cutover fixture pack green, CT-1316 green as written, full suite 594, scratch worktrees and instrumentation removed, `actionParent|actionChildren` extinct. The kind-stickiness hard error plus action-keyed edges give the registry the exact persistence semantics the inventory documented as intentional in v1.

seefeldb:
Both findings fixed red-first (old tip adb35cd60, rebased onto main):

1. **Computation→effect promotion** (Codex + cubic): `NodeRegistry.register` threw on the v1-sanctioned promotion path (action first subscribed as computation, later re-subscribed `isEffect: true` — v1's "once an effect, stays an effect" latch). `register()` now promotes computation→effect in place (kind mutated, active sets swapped) and still throws on demotion, which has no sanctioned caller. Fixture: "promotes a computation to an effect on re-registration" (registry-level + live re-subscription).
2. **Parent links through registration churn** (cubic P1): `parentOf(x)?.action` collapsed to undefined while the parent's record was unregistered, but demand/trace checks key off action objects — v1's parent WeakMap answered regardless. Added `parentActionOf()` (raw captured action) and switched the consumer sites (demand.ts, action-run trace, graph-snapshot ×2, topology ×3, pull-subscriptions); `parentOf()` record resolution unchanged. Fixture: "keeps captured parent actions reachable before the parent registers".

Runner 618 green locally.

seefeldb:
**Reviewer pass:** both original 3a commits rebased byte-identical, the phase-3 baseline ledger section is restored (`064094d41` — thanks for picking up the #4100 flag), and the new fix `89e916d76` is verified on both counts:

- **Effect promotion** — correct v1 parity: re-subscription with `isEffect: true` promotes in place (the one sanctioned, monotone transition — once an effect, stays an effect), demotion still throws. This preserves the v2 invariant that effect-ness is static *per direction* while honoring v1's subscribe-overwrites behavior.
- **`parentActionOf()`** — this completes the WeakMap-parity doctrine from the creation-context debugging: the registry already keys edges by action objects, but `parentOf(x)?.action` still gated parent visibility on the parent's **record** existing — the same failure class as the original `parent-record-missing` culprit, one level up. Demand ordering and trace labels must see captured parents through registration-churn windows; switching exactly those call sites (and not `isAncestor`) is the right scope.

Cutover-test additions cover both. LGTM pending CI.

PR #4102, opened 2026-06-12
refactor(runner): scheduler-v2 cutover 3b: liveness refcounts

seefeldb:
Stack: 07/3b — based on scheduler-v2/07-3a (#4101)

Progress excerpt: docs/specs/scheduler-v2/implementation/PROGRESS.md sections 07/3b-liveness-refcounts, 07/3b-provisional-demand, and 07/3b-demand-consumers.

Summary:
- Added scheduler node liveness refcount maintenance in the dependency graph.
- Moved parent-created first-run demand onto SchedulerNode.provisionalDemand with pass/finalize expiry.
- Replaced demand.ts consumers with node-record liveness lookups and deleted demand.ts plus the old demand WeakSets.

Verification:
- deno fmt/lint/check on touched scheduler files: passed.
- Focused scheduler gate: 11 passed (132 steps), 0 failed.
- scheduler-pull*.test.ts gate: passed in the 3b.2 pull-family run.
- scheduler-demand-roots.bench.ts: no >10% regression; recorded in PROGRESS.md.
- cd packages/runner && deno task test: 594 passed (3109 steps), 0 failed, 0 ignored (10 steps), 2m7s.

---
## Summary by cubic
Scheduler v2 cutover 3b replaces demand-based logic with graph-driven liveness using node refcounts and provisional demand, removing `demand.ts`. It propagates liveness on edge updates, expires provisional demand correctly, and fixes first-run debounce wake planning.

- **Refactors**
  - Added node liveness refcounts; `isLive` is true for effects, positive refs, provisional demand, or materializer computations.
  - Dependent edge register/unregister now add/drop upstream live refs; liveness propagates through `reverseDependencies`.
  - Moved first-run demand to `SchedulerNode.provisionalDemand` with pass id; expires at pass end after first run, or on first completed run for continuations.
  - Replaced all `demand.ts` consumers with node-record liveness; deleted `demand.ts`; moved `hasDependentPath` to `dependency-graph.ts`.
  - Unsubscribe removes graph edges before clearing materializer registration and clears provisional demand before node removal; materializer/type changes notify liveness transitions.
  - Staleness checks use `isLiveComputation`; live computations auto-queue as before.

- **Bug Fixes**
  - Added visited-set guard to liveness refcount updates to prevent cycle double-counting and ensure cycles release when the last live root unsubscribes.
  - Included `shouldDebounceFirstRun` in debounce planning so first-run debounced computations have planned wake times.

[...]

PR #4103, opened 2026-06-12
refactor(runner): scheduler-v2 cutover 3c: pass and single channel

seefeldb:
Stack: 07/3c — based on scheduler-v2/07-3b (#4102)

## 07/3c.i-docs-amendment

- [x] pending — amended 07/3c.i and 07/3c.ii work-order text per reviewer
  ruling.
- Fix shape:
  - 3c.i downstream closure is now specified as live downstream effects only,
    preserving v1 computation-gate behavior.
  - 3c.ii now explicitly expands the closure to all live nodes in the same
    commit as the invalid-at-turn run gate.
  - `test/scheduler-retries.test.ts` is now an explicit 3c.ii gate.

## 07/3c.i-closure-ordering

- [x] pending — added live downstream effect closure ordering for scheduler
  settle work sets.
- Fix shape:
  - `buildPullIterationWorkSet` now walks `dependents` from dirty work-set
    roots after seed + upstream collection.
  - The 3c.i closure adds only live effect nodes, per reviewer ruling, so
    unchanged v1 computation gates cannot admit downstream computations early.
  - The settle-loop state now exposes `isLiveAction` for the closure helper;
    3c.ii will reuse this path when computations join the closure with the new
    invalid-at-turn run gate.
- Recordings:
  - `deno fmt packages/runner/src/scheduler.ts
    packages/runner/src/scheduler/execution.ts
    packages/runner/src/scheduler/pull-execution.ts`: passed
    (`Checked 3 files`).
  - `deno lint` on the same 3 files: passed (`Checked 3 files`).
  - `deno check` on the same 3 files: passed.
  - Cutover fixture pack:
    `ENV=test deno test --allow-ffi --allow-env --allow-read
    --allow-write=/tmp,/var/folders --allow-run=git
    packages/runner/test/scheduler-v2-cutover.test.ts`: passed,
    `1 passed (10 steps)`, `0 failed`.
  - Retry parity witness:
    `ENV=test deno test --allow-ffi --allow-env --allow-read
    --allow-write=/tmp,/var/folders --allow-run=git
    packages/runner/test/scheduler-retries.test.ts`: passed,
    `1 passed (2 steps)`, `0 failed`.
  - `cd packages/runner && deno task test`: passed,
    `594 passed (3109 steps)`, `0 failed`, `0 ignored (10 steps)`, `2m6s`.

## 07/3c.ii-value-gated-effects-channel-deletion

- [x] pending — migrated scheduler invalidation to node status and deleted the
  duplicate write-propagation/scheduled-effect channels.
- Fix shape:
  - `record.status`/`record.invalidCauses` now own invalidation state; storage
    notifications call the `markInvalid` entry point and action runs consume
    or restore causes from the node record.
  - The settle work-set closure now includes all live downstream nodes, while
    the turn gate runs only invalid/never-ran live eligible nodes under
    `PASS_RUN_BUDGET`; clean computations included for ordering are skipped.
  - Deleted the write-propagation module and the `changedWritesHistory`,
    `conditionallyScheduledEffects`, `scheduledFirstTime`,
    `scheduleAffectedEffects`, and `scheduledEffects` trace channels.
  - First-run no-output computations discovered during dependency collection
    receive provisional demand so startup/restart semantics stay live.
  - Initial clean rehydration refuses to overwrite a node once the dirty mirror
    or invalid status shows it has been invalidated during the async load.
  - Stale-input resubscribe skips debounced computation writers so the debounce
    wake path, not same-turn effect revalidation, controls the trailing flush.
  - Scheduler tests use `StorageManager.emulate({ as: signer })` through
    `createSchedulerTestRuntime`; `rg` found no other scheduler storage
    configuration in runner scheduler helpers. The representative synchronous
    notification assertion runs in `scheduler-pull.test.ts`.
- Recordings:
  - Deletion grep returned no matches:

```text
$ rg -n "cfcTriggerReads|changedWritesHistory|conditionallyScheduledEffects|scheduledFirstTime|scheduleAffectedEffects|markEffectConditionallyScheduled|conditionalEffectHasChangedInputs|markPullDemandContinuation|recordChangedComputationWrites|markReadersDirtyForChangedWrites|writePropagation|TriggerTraceScheduledEffect|scheduledEffects|mark-dirty|already-dirty|schedule-effect" packages/runner/src packages/runner/test
```

  - `git diff --check`: passed.
  - `deno fmt` on the 29 touched source/test files: passed (`Checked 29 files`).
  - `deno lint` on the same 29 files: passed.
  - `deno check` on the same 29 files: passed.
  - Persistent rehydration trio:
    `ENV=test deno test --allow-ffi --allow-env --allow-read
    --allow-write=/tmp,/var/folders --allow-run=git
    packages/runner/test/scheduler-observations.test.ts
    packages/runner/test/reload-rehydration.test.ts
    packages/runner/test/reload-sibling-overdirty.test.ts`: passed,
    `3 passed (22 steps)`, `0 failed`.
  - Regression set:
    `cd packages/runner && ENV=test deno test --allow-ffi --allow-env
    --allow-read --allow-write=/tmp,/var/folders --allow-run=git
    test/ensure-piece-running.test.ts test/runner.test.ts
    test/scheduler-timing.test.ts`: passed, `7 passed (72 steps)`,
    `0 failed`.
  - 3c.ii gate pack:
    `cd packages/runner && ENV=test deno test --allow-ffi --allow-env
    --allow-read --allow-write=/tmp,/var/folders --allow-run=git
    test/scheduler-v2-cutover.test.ts test/scheduler-cfc-trigger-reads.test.ts
    test/scheduler-retries.test.ts test/scheduler-pull.test.ts
    test/scheduler-convergence.test.ts test/scheduler-ordering.test.ts
    test/scheduler-events.test.ts`: passed, `11 passed (92 steps)`,
    `0 failed`.
  - `cd packages/runner && deno task test`: passed,
    `594 passed (3108 steps)`, `0 failed`, `0 ignored (10 steps)`, `2m6s`.
  - Expected noisy passing logs remain: legacy cycle-cap telemetry in
    convergence cases and the reload-rehydration fixture's expected
    `TypeError`.

## 07/3c.iii-upstream-machinery-deletion

- [x] pending — deleted the upstream dirty/stale machinery and rebuilt the
  remaining event-preflight and settle seeding paths around node invalid status.
- Fix shape:
  - Deleted `src/scheduler/dirty-dependencies.ts` and
    `src/scheduler/staleness.ts`; scheduler state no longer owns a
    dirty/stale mirror, `collectStack`, or dirty-dependency trace context.
  - Added `event-preflight-dependencies.ts` with
    `collectInvalidUpstreamForLog`, preserving the public preflight stats shape
    while collecting only invalid/never-ran upstream nodes.
  - Event preflight now walks through clean intermediates to find invalid
    ancestors of handler reads, so transient event demand still blocks on
    stale inputs without restoring pull mode's old dirty collector.
  - Pull settle work-set construction now uses initial seeds plus invalid/live
    seeds, invalid materializer promotions, and live downstream closure; the
    old initial-seed/traversal-root asymmetry and late materializer per-effect
    recheck are gone.
  - Continuations, delays, cycle-break shims, graph snapshots, subscription
    revalidation, and persisted rehydration now derive "dirty" compatibility
    from node invalid/never-ran status.
  - The first full-suite attempt exposed an async initial-rehydration race:
    `never-ran` is both the normal pre-rehydrate state and the state after an
    explicit invalidation. `markActionInvalid` now cancels any pending initial
    rehydrate token so a late clean snapshot cannot overwrite that
    invalidation.
- Recordings:
  - Focused scheduler trio:
    `cd packages/runner && ENV=test deno test --allow-ffi --allow-env
    --allow-read --allow-write=/tmp,/var/folders --allow-run=git
    test/scheduler-events.test.ts test/scheduler-pull.test.ts
    test/scheduler-convergence.test.ts`: passed, `3 passed (53 steps)`,
    `0 failed`.
  - 3c.iii gate pack:
    `cd packages/runner && ENV=test deno test --allow-ffi --allow-env
    --allow-read --allow-write=/tmp,/var/folders --allow-run=git
    test/scheduler-v2-cutover.test.ts
    test/scheduler-cfc-trigger-reads.test.ts test/scheduler-retries.test.ts
    test/scheduler-ordering.test.ts test/scheduler-timing.test.ts
    test/scheduler-throttle.test.ts`: passed, `10 passed (68 steps)`,
    `0 failed`.
  - Rehydration regression check:
    `cd packages/runner && ENV=test deno test --allow-ffi --allow-env
    --allow-read --allow-write=/tmp,/var/folders --allow-run=git
    test/scheduler-observations.test.ts`: passed, `1 passed (22 steps)`,
    `0 failed`.
  - Event/stale bench pair:
    `cd packages/runner && deno bench --allow-ffi --allow-env --allow-read
    --allow-write=/tmp,/var/folders --allow-run=git
    test/scheduler-event-preflight.bench.ts
    test/scheduler-stale-propagation.bench.ts`: passed. Results:
    event clean broad graph `328.7ms`, event waits on transitive invalid writer
    `25.7ms`, note-shaped clean events `1.1s`, deep read-populated handler
    `545.2ms`; stale-propagation bench, now an end-to-end graph update bench:
    chain `98.9ms`, diamond `88.6ms`, wide fanout `239.8ms`, dynamic deps
    `81.5ms`, unchanged recompute `45.9ms`.
  - `deno fmt` on the 29 touched source/test files: passed
    (`Checked 29 files`).
  - `deno lint` on the same 29 files: passed (`Checked 29 files`).
  - `deno check` on the same 29 files: passed.
  - `git diff --check`: passed.
  - Upstream deletion grep returned no matches:

```text
$ rg -n "staleness|dirty-dependencies|SchedulerStaleness|collectStack|dirtyDependencyTraceContext|DirtyDependencyTraceContext|collectDirtyDependencies|collectDirtyDependenciesForLog|collectDirtyDependenciesFromTraversalRoot|deferEffectForLateMaterializerDependency|isStale\(|getUpstreamStaleCount|stale\.size|upstreamStale|markDirectDirty|clearSchedulerDirectDirty|clearSchedulerDirty|forceClearStale" packages/runner/src packages/runner/test
```

  - Scheduler-source stale grep returned no matches:

```text
$ rg -n "stale\b" packages/runner/src/scheduler packages/runner/src/scheduler.ts
```

  - First `cd packages/runner && deno task test` attempt failed one test:
    `scheduler-observations.test.ts` /
    `does not apply async initial rehydration after an action becomes dirty`.
    The final rerun passed: `594 passed (3106 steps)`, `0 failed`,
    `0 ignored (10 steps)`, `2m6s`.

## 07/3c.iv-budgets-backoff

- [x] pending — replaced cycle breaking/adaptive cycle debounce with v2 pass
  budgets and exponential backoff.
- Fix shape:
  - `constants.ts` now owns `MAX_ITERS = 10`, `PASS_RUN_BUDGET = 5`, and
    the 250ms/2000ms backoff bounds; the old `MAX_ITERATIONS_PER_RUN`,
    `loopCounter`, and cycle-debounce constants are gone.
  - Node records now carry `backoffUntil` and consecutive
    `backoffFailures`; settle execution applies exponential backoff when an
    iteration cap or pass-budget exhaustion leaves runnable invalid work.
  - Pass-budget backoff keys off any action reaching the run budget, then
    defers the currently runnable invalid nodes. This fixes alternating cycles
    where the action that trips the budget is clean by the time planning runs.
  - Continuation and run gating now use combined throttle/backoff eligibility,
    so delayed work schedules through the existing event-wake path rather than
    immediate requeueing.
  - `scheduler.non-settling` telemetry is emitted once per backoff episode via
    the existing settling tracker.
  - Deleted `pull-cycle-break.ts`, `planPullCycleBreak`,
    `planPullAdaptiveCycleDebounce`, adaptive cycle-debounce application, and
    the remaining effect re-dirty cycle-skip paths.
  - Updated scheduler convergence/timing/cutover/core tests to assert bounded
    backoff semantics instead of the removed hard loop cap and cycle debounce.
- Recordings:
  - Deletion grep returned no matches:

```text
$ rg -n "runsThisExecute|loopCounter|executeStartTime|CYCLE_DEBOUNCE|MAX_ITERATIONS_PER_RUN|PullCycleBreakState|breakPullCyclesIfNeeded|planPullCycleBreak|planPullAdaptiveCycleDebounce|schedule-cycle-debounce|schedule-cycle|dirtyEffectsToRun|earlyIterationComputations|getLoopCounter|pull-cycle-break|cycle-break|cycle-aware debounce|cycle debounce" packages/runner/src packages/runner/test
```

  - `deno fmt` on the touched source/test files: passed.
  - `deno lint` on the touched source/test files: passed (`Checked 11 files`).
  - `deno check` on the touched source/test files: passed.
  - `git diff --check`: passed.
  - Core regression check:
    `cd packages/runner && ENV=test deno test --allow-ffi --allow-env
    --allow-read --allow-write=/tmp,/var/folders --allow-run=git
    test/scheduler-core.test.ts`: passed, `1 passed (25 steps)`, `0 failed`.
  - 3c.iv gate pack:
    `cd packages/runner && ENV=test deno test --allow-ffi --allow-env
    --allow-read --allow-write=/tmp,/var/folders --allow-run=git
    test/scheduler-v2-cutover.test.ts test/scheduler-convergence.test.ts
    test/scheduler-timing.test.ts test/scheduler-throttle.test.ts`: passed,
    `4 passed (49 steps)`, `0 failed`.
  - First `cd packages/runner && deno task test` attempt failed the legacy
    `scheduler-core.test.ts` loop-cap assertion after the hard loop cap was
    removed. The final rerun passed: `594 passed (3099 steps)`, `0 failed`,
    `0 ignored (10 steps)`, `2m8s`.
  - Expected noisy passing logs remain: deliberate scheduler errors in
    convergence/core/stack-trace tests, reload-rehydration's expected
    `TypeError`, and existing write-surface/wish warnings.

---
## Summary by cubic
Cut over scheduler v2 phase 3c: move invalidation into node status, delete duplicate channels, add effects-only closure ordering, pass budgets/backoff, an invalid-node index with inverted event preflight, and a rehydration barrier so reloads are deterministic. Planning is faster and `idle()` resolves reliably under load.

- **Refactors**
  - Node records own invalidation; notifications call `markInvalid`; action runs take/restore invalid causes from records; initial rehydrate can’t overwrite later invalidation.
  - `NodeRegistry` maintains an index of invalid/never‑ran active nodes; seed scans and runnable checks iterate this set instead of full-node scans.
  - Event preflight uses inverted reachability from the maintained invalid set into the handler‑read closure (`event-preflight-dependencies.ts`); snapshots, tracing, and benches updated.
  - Removed upstream dirty/stale machinery and duplicate write‑propagation/scheduled‑effect channels; run a single event channel.
  - Work‑set adds an effects‑only downstream closure for ordering; the run gate executes only invalid/never‑ran live nodes under budget.
  - Added pass budget and exponential backoff (`MAX_ITERS`, `PASS_RUN_BUDGET`, `BACKOFF_BASE_MS`, `BACKOFF_MAX_MS`); iteration‑cap backoff is retained so `idle()` can resolve.
  - Added a rehydration barrier: pull work waits until all initial resumes finish to prevent fresh runs from preempting resume and to avoid reload idle hangs.
  - First‑run debounced computations keep `idle()` alive; continuation waits for idle on initial demanded runs.

- **Migration**
  - Trigger‑trace decisions now use `mark-invalid`/`already-invalid`; `TriggerTraceScheduledEffect` and `scheduledEffects` are removed.
  - Drop downstream‑effects aggregation in `packages/shell` and `packages/patterns`.
  - Event effects may run once fewer per commit (single channel); update test expectations.
  - Bench/timing labels now say “invalid upstream” instead of “dirty/stale”.

<sup>Written for commit 59f36bc713c137258be97e91ba66a99ab1cc5f80. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4103?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seefeldb:
Reviewer pass — approved. This is the project's keystone commit-series: the duplicated propagation channel and its compensating machinery are gone, verified by grep-zero at head for: `write-propagation`, `conditionallyScheduledEffects`, `changedWritesHistory`, `scheduleAffectedEffects`, `dirty-dependencies`, `collectStack`, `pull-cycle-break`, `MAX_ITERATIONS_PER_RUN`.

Parity evidence:
- **The phase-0 tripwire flipped**: `scheduler-events` "should trigger recomputation of dependent cells" counts dropped from 1→3→5 back to **1→2→3** — the per-event-commit double-run of reader effects is dead, exactly as the spec's channel-#2 analysis predicted when this test was planted.
- Fixture 3 (effect not run on unchanged upstream output) held through the watermark deletion; `scheduler-retries` held through the computation-closure flip (named gate from the 3c.i verdict — the closure narrowed to effects-only in i, computations joined atomically with the invalid-at-turn gate in ii).
- iv replaces the loop cap + cycle breaker + cycle debounce with PASS_RUN_BUDGET=5 / MAX_ITERS=10 / 250ms→2s exponential backoff + once-per-episode non-settling telemetry; the legacy loop-cap assertion was rewritten to the backoff contract with a record.

Suite 594 green at every commit in the series. One ask: watch this PR's CI specifically — it is the first time the channel deletion meets the cross-package jobs.

seefeldb:
Rebased onto main (old tip cc9774c31). Findings + one rebase interaction:

**1. Iteration-cap backoff time-gating healthy chains (Codex P2)** — fixed red-green. `passRuns` resets per execute-pass and a tight cycle trips the per-pass run budget *mid-loop* (breaking with `pass-budget` backoff before ever reaching the iteration cap). So reaching the cap means a deep/slow chain making forward progress, not a cycle — time-gating it for `BACKOFF_BASE_MS` is wrong. `isBudgetBackoffCandidate` now applies the `passRuns >= PASS_RUN_BUDGET` gate for the `iteration-cap` reason too (the in-loop `pass-budget` break opts out via `requirePassRunBudget: false`); a sub-budget chain re-passes immediately. Tests: "reserves iteration-cap backoff for cycle evidence" (direct `planBudgetBackoff` red-green) + "settles a deep acyclic chain ..." + the existing "bounds a cycling subgraph ..." confirms real cycles stay bounded.

**Rebase interaction (not a finding — surfaced by my #4102 fix on main).** 3c's continuation refactor removed deferred actions from `hasPendingPullWork`, tracking their future run via `nextDirtyPullRunWaitsForIdle` — which only covered effects/materializers. Original 3c masked the gap because the planning-path `getNextDebounceRunTime` returned `undefined` for first-run debounced computations (so they stayed counted as pending work). My #4102 consistency fix makes it return a value, so a never-ran demanded debounced computation (debounced child under a live parent's provisional demand) got deferred out of pending-work *without* keeping idle() alive — idle returned in ~3ms instead of waiting ~200ms for its gate. `noteFutureEligibility` now also waits-for-idle on `shouldRunFirstPullComputationInDemandContext` (never-ran + provisional demand), limited to the first run so idle never blocks through a throttle window or trailing flush of an already-run action. The existing "keeps provisional demand for a debounced child until the gate opens" fixture (216ms, green on original 3c) is the guard.

**Declined cubic P2s** (reasoning):
- *resubscribe clean-up guard too narrow* — correct as-is: it prevents a *first* registration via resubscribe being wired invalid mid-function; an *existing* never-ran record should stay runnable. Rehydration's clean path force-sets `status = clean` independently (scheduler.ts), so rehydrated actions don't re-run (rehydration suite green).
- *continuation re-tick for non-runnable invalid nodes* — `hasDirtyPullWork` already filters invalid nodes that aren't effect/demanded/materializer, and `noteFutureEligibility` filters debounce/throttle-deferred ones; a survivor is genuinely runnable.
- *initial-run fallback queueExecution* — undemanded failed work staying unscheduled is correct pull semantics; demand arrival queues it.

Runner 618 green locally.

seefeldb:
**Correction — iteration-cap backoff change REVERTED.** My earlier iteration-cap cycle-evidence gate broke the `should persist and reload every rapidly created notebook note` and the notebook-reload integration tests with `RuntimeClient request timed out: runtime:idle` (consistent on CI, load-sensitive — green 3/3 locally).

Root cause (verified in code): the iteration-cap backoff is **not** merely a perf pause — it is the escape valve that lets `idle()` *resolve* when a live subgraph hasn't settled within MAX_ITERS. `planBudgetBackoff` sets `record.backoffUntil`; `getNextEligibleRunTime` returns it via `maxFutureTime(...)`; `noteFutureEligibility` then defers the action out of `hasDirtyPullWork` (so `applyQuiescentContinuation` resolves the idle promises) and schedules a retry wake. Gating iteration-cap backoff on cycle evidence removed that valve for any sub-budget chain that repeatedly hits the cap under load — idle keeps re-ticking and never resolves. Rapid notebook-note creation under CI load is exactly that workload.

So **Codex P2 #4103 is declined**: both suggested forms (gate on cycle evidence / re-queue immediately) remove the idle-resolution valve without a convergence oracle to tell a deep-but-converging chain from a non-converging one. The ~250ms pacing on a chain deeper than MAX_ITERS is the cost of guaranteeing idle resolution — a minor perf nit, not a correctness bug. Reverted to the original full-backoff behavior.

The other #4103 changes stand: the rebase-interaction `futureRunWaitsForIdle` fix (machine-independent, restores main's idle-wait for first-run demanded debounced computations) and the `scheduledEffects` cross-package consumer cleanup (shell debug-utils + default-app diagnostics — the whole-repo type break the stacked-PR CI gap had hidden). Runner 618 green locally.

seefeldb:
## Perf blocker: rapid-notebook integration tests fail (O(N²) in 3c core)

The `Pattern Integration (2/4)` + `Pattern Reload` failures (`RuntimeClient request timed out: runtime:idle/dispose` on "…rapidly created notebook note") are a **real 3c regression**, not flake and not my review-fix commits. Escalating for a design decision (per @seefeldb) rather than reverse-engineering a fix to the deleted upstream-stale machinery.

### Reproduce locally
`EXPERIMENTAL_PERSISTENT_SCHEDULER_STATE=true PORT_OFFSET=37 ./scripts/restart-local-dev.sh --force --dangerously-clear-all-spaces`, then from `packages/patterns`: a notebook-reload test with the note count cranked to 30–40 (env-knob patch below) wedges; at the default 7 it only wedges on slow CI runners.

### Root cause (measured)
Per-create scheduler settle cost grows **linearly with total node count** (idle 265ms@586 nodes → 1490ms@2473 nodes), while only **~29 actions actually run per create** (constant). So it's pure overhead → **O(N²)** for rapid creation. Per-phase timing (cumulative, instrumented):
- **`collectInvalidUpstreamForLog` (event-preflight upstream DFS) ≈ 564 ms/create, O(N) — dominant.** The notebook note-list is a hub; the preflight walks upstream from it and **visits all N notes** to find the few invalid ones (visit-count is the cost, not allocation).
- `settleLoop` dependent-expansion ≈ 316 ms/create, also O(N) over the hub.

This is the direct cost of the **`delete scheduler upstream stale machinery`** commit: it replaced incremental upstream-stale counts (O(1) "is upstream invalid?") with an on-demand full graph walk.

### Things ruled out
- Pruning the DFS at clean nodes is **incorrect** — pull invalidation is lazy/non-transitive (a clean node legitimately has invalid upstream that hasn't propagated; `scheduler-pull.test.ts` "should keep event handlers behind invalid dependencies" proves it).
- A constant-factor allocation opt (avoid the per-visited-node `new Set` in `collectUpstreamWriters`, short-circuit `collectMaterializerWritersForLog` when no materializers) **did not help** — it's the O(N) visit count.

### Suggested fix directions (for the implementer)
1. **Incremental upstream-invalid tracking** (principled): a maintained per-node "has-invalid-upstream" signal updated on status + edge changes → preflight O(1)/O(direct). Re-adds a form of what 3c deleted.
2. Memoize the preflight walk across the several event preflights within one create's settle.

### Partial improvements I validated (saved as a patch, not committed)
Unit suite stays green (618) with: an **invalid-node index** in `NodeRegistry` (routes status writes through `setStatus`) replacing the O(N) full-node scans in `collectPullIterationSeeds`/`hasRunnablePullWork`/continuation `hasDirtyPullWork`; plus the allocation opt above. These are correct and reduce other scans but **don't touch the dominant preflight DFS**, so I left them out to keep this PR clean — patch available if useful.

The branch is otherwise green (runner 618; the review findings — iteration-cap cycle-evidence gate, first-run-debounce idle wait — are red-green tested and pushed).

seefeldb:
## Design decision: invert the preflight upstream-invalid walk — fix in 3c before merge

Investigated with @seefeldb. **Verdict: this is a real design gap, not flake and not your review commits — and it's fixable inside 3c without re-introducing the deleted machinery.**

### Diagnosis (grounded)
The regression traces to one falsified claim. `current-system-inventory.md:75` deleted `SchedulerStaleness` (`upstreamStaleCount`) arguing *"with no consumer for transitive staleness there is no consumer… the pass computes its downstream closure per iteration (cheap: bounded by work-set size, not graph size)."* Both halves are wrong here:

1. **There is a consumer** — the event-preflight consistency gate (`event-preflight-dependencies.ts:collectInvalidUpstreamForLog`, spec §7.5 step 2 / invariant **I4**). It must answer *"is any node transitively upstream of the handler's read-closure invalid?"* before dispatching an un-re-runnable handler.
2. **Its cost is graph-size-bounded** — when the closure reads a hub (the note-list aggregating N notes) the upstream walk visits all N writers → O(N)/preflight → O(N²) under rapid creation.

The §14 headline *"P2: reachability never decides runs, so transitive marking has no decision left to make"* (`README.md:1009`) is true for the **run** decision but has exactly **one exception**: this gate. And the only mitigation the design planned — decision 3, closure caching — addresses **read-closure discovery**, a *different* cost from the **upstream-invalid reachability walk** you measured as dominant (564 ms/create). So closure caching would never have helped. You're also right that pruning at clean nodes is incorrect (`scheduler-pull.test.ts:1595` proves the walk must be transitive).

### Decision
1. **Mechanism — invert the walk.** Instead of walking *up* from the closure across the whole hub, walk *down* from the (small) maintained invalid-node set and ask "does any reach this closure?" Reuses your validated invalid-node index, stays **fully lazy** (re-adds **no** transitive marking → D5/I2 zero-cost dormancy preserved), turns O(N²)→O(N) for the notebook shape. The incremental `hasInvalidUpstream` refcount (riding §5.2 liveness) is spec'd as the **benchmarked escalation** for read-side-fan-out shapes — *not* adopted now.
2. **Sequencing — fix in 3c before merge.** Don't ship the O(N²) into the cutover. Land the invalid-node index + the inverted walk, get the rapid-notebook integration tests honestly green, then merge.

---

### Work order — code
1. **Invalid-node index in `NodeRegistry`** (your validated patch — land it): route status mutations through `setStatus`; maintain `private invalidNodes = new Set<Action>()` (add on transition into `invalid`/`never-ran`, remove on transition out); expose `getInvalidNodes(): ReadonlySet<Action>`. Also use it to replace the O(N) full-node scans in `collectPullIterationSeeds` / `hasRunnablePullWork` / continuation `hasDirtyPullWork`.
2. **Invert `collectInvalidUpstreamForLog`**:
   - Keep computing the closure target set `C` = `collectDirectWritersForLog(...)` + materializers (unchanged). Put it in a membership `Set`.
   - For each `U` in `nodes.getInvalidNodes()`, test downstream-reachability to `C`. The downstream step ("who reads U's outputs") is the **same reader/trigger index that drives `markInvalid`** — match on `getSchedulingWrites(U)` addresses; don't build a new adjacency map. Walk with a `visited` set + cycle guard; on reaching any node in `C`, add `U` to `invalidDeps`/`workSet` and set `hasInvalidUpstream = true`. A `U` already in `C` and invalid counts at 0 hops.
   - This yields the **same set** as the old forward walk — {invalid nodes transitively upstream of `C`}. Preserve that exactly; the downstream plan/scheduling that consumes `invalidDeps` is unchanged.
   - Optional within-pass memo: the invalid set and `C` are stable across the several preflights of one settle; cache per-`U` "reaches C?" keyed by a status-generation counter. Nice-to-have, not required once per-preflight cost is small.
3. **Semantics preservation (red-green):** `scheduler-pull.test.ts:1595` must stay green — the inverted walk must still find `upstreamAction` (invalid) reaching `C = {downstreamAction}` even though `output`'s direct writer `downstreamAction` is clean. (downstream(upstreamAction) = readers of `mid` = {downstreamAction} ∈ C → relevant.) That transitive case is the whole point; assert it explicitly.
4. **Perf gate:** the `Pattern Integration (2/4)` + `Pattern Reload` "rapidly created notebook note" tests are currently RED (wedge → `runtime:idle/dispose` timeout) at your cranked note count (30–40). After the fix they go green there, **and** per-create settle time stays ~flat as total node count grows (that flatness IS the fix — not just "passes at count 7"). Keep your env-knob patch as a regression bench if the default-7 integration runner is too noisy to catch it.
5. **Bound note (PROGRESS):** record that the inverted walk's worst case is read-side fan-out (one invalid node feeding many closure writers); if a real workload trips it, decision 15's escalation (incremental refcount) is the sanctioned next step. Don't build it now.

### Work order — spec/inventory (apply VERBATIM; do not paraphrase design prose — the binding-spec lesson applies)

**`docs/specs/scheduler-v2/current-system-inventory.md`** — the `SchedulerStaleness` row: change disposition `Delete` → `Subsume` and replace the rationale cell with:

> Subsume — the per-data-change push (`propagation on every dirty/clean transition`) is deleted, but the **query** it served survives in one consumer: the event-preflight consistency gate (§7.5 step 2 / I4), which must know "is any node transitively upstream of the handler closure invalid?". v2 answers it by **inverting** the walk — reachability from the maintained invalid-node set (a `Set<Action>` in `NodeRegistry`, updated through `setStatus`) downstream into the closure — so cost is bounded by invalid-set × observed downstream cone, not graph size. The incremental upstream-invalid refcount (riding §5.2 liveness) is the spec'd escalation if a read-side-fan-out workload makes the inverted walk O(graph); see §15 decision 15.

**`docs/specs/scheduler-v2/README.md`** §14 table — replace the `upstream-stale counts` safety-argument cell with:

> P2 holds for the **run** decision (effects gate on their own value-accurate invalid bit). The one reachability query that survives — the event-preflight consistency gate (§7.5/I4) — is served without per-data-change transitive marking by **inverting** it over the maintained invalid-node set (decision 15); the incremental refcount is the spec'd escalation.

**`docs/specs/scheduler-v2/README.md`** §7.5 step 2 — after the first sentence of the Consistency gate, insert:

> The invalid-upstream set is computed by **inverted reachability** from the maintained invalid-node set into the closure — not an upstream walk of the closure's cone, which is O(graph) against a hub (decision 15).

**`docs/specs/scheduler-v2/README.md`** §15 — append as the next Resolved decision (15):

> **15. Preflight upstream-invalid reachability.** The event-preflight consistency gate (§7.5 step 2, I4) is the one surviving consumer of transitive-staleness reachability — P2's "reachability never decides runs" holds for the run decision but not for this gate. Deleting v1's `upstreamStaleCount` turned the gate's query into an O(graph) upstream walk (`collectInvalidUpstreamForLog`), O(N²) under rapid creation against a hub (measured: rapid-notebook, 564 ms/create). Resolution: **invert the walk** — reachability from the maintained invalid-node set (a `Set<Action>` in `NodeRegistry` updated through `setStatus`) downstream into the closure — cost bounded by invalid-set × observed downstream cone, re-adding no per-data-change transitive marking (dormancy stays zero-cost, D5/I2). Orthogonal to decision 3 (closure caching), which addresses read-closure *discovery*, a different cost. **Escalation (spec'd, not adopted):** if a read-side-fan-out workload (one invalid node feeding many closure writers) makes the inverted walk O(graph), adopt an incremental `hasInvalidUpstream` signal maintained the way §5.2 maintains liveness, scoped to live nodes — feeding only this gate, never the run decision, so push mode / conditional effects / watermarks stay deleted. Adopt only behind the preflight benchmark.

When this lands, reply in PROGRESS.md with the red→green transition (integration test at the cranked count) and the per-create-flat measurement, and I'll do the review pass.

seefeldb:
## Implemented: inverted event-preflight walk (decision 15) — pushed to `scheduler-v2/07-3c`

Took over the implementor role and landed the fix per the design verdict above. Four commits (`9a36632e8`…`92135001b`):

1. **`refactor(runner): invalid-node index in NodeRegistry`** — a `Set<Action>` of active invalid/never-ran nodes maintained through a single `setStatus()` choke point (+ activate/remove). All node-status writes routed through it; the three O(N) full-node scans (`collectPullIterationSeeds`, `hasRunnablePullWork`, `hasDirtyPullWork`) now iterate the index. Correctness-neutral.
2. **`fix(runner): invert the event-preflight upstream-invalid walk`** — seed from `getInvalidNodes()` and walk **down** to the closure via `dependents` (exact inverse of `reverseDependencies`), with materializer→reader edges mirrored from `collectMaterializerWritersForLog`. Cost is bounded by the invalid set × its downstream cone, not the closure fan-out; re-adds no per-data-change transitive marking (dormancy stays zero-cost, D5/I2).
3. **`test(runner): wide fan-in hub preflight regression guard`** — the faithful notebook shape.
4. **`docs(specs): decision 15`** — inventory row flipped Delete→Subsume, README §7.5/§14/§15, PROGRESS record.

### Red→green / guards (`scheduler-pull.test.ts`)
- `should keep event handlers behind invalid dependencies` (transitive consistency semantics) — still green.
- `should dispatch clean broad event preflight dependencies` — the 32-wide clean fan-out now reports `visitCount < 10` (was the full ~33 cone).
- `defers an event handler behind an invalid upstream materializer` — **new**, exercises the materializer arm through the gate.
- `bounds event preflight cost over a wide fan-in hub` — **new**, 200 writers → one hub the handler reads; invalidating one writer keeps `visitCount < 20` (the old upstream walk visited ~200) while still dispatching on the re-settled hub.

Full runner suite green (618). Note: `scheduler-event-preflight.bench.ts` wall-clock is unchanged because it's dominated by per-iteration graph setup (512-node build + pull), not the preflight — the walk reduction is proven by the visitCount guards, not that bench. The real end-to-end validation is this PR's `Pattern Integration (2/4)` + `Pattern Reload` jobs, which were the failing ones.

I'll watch CI and shepherd to green.

seefeldb:
## Reload hang fixed — root cause re-verified, rehydration barrier (`2a39afbae`)

Took over the implementor role and, per request, **re-verified the root cause (keeping the bisect) before fixing**, since the earlier attempts had failed.

### Re-verification (airtight)
Empirical bisect re-run with the missing data point — the base `0f733e897` (3b merge, pre-cutover):

| commit | reload runs | rehydrateOk | |
|---|---|---|---|
| base `0f733e897` | 23 ×8 | 56 | deterministic ✅ |
| `22a5676fc` (1st 3c commit) | 23 ×4 | 56 | deterministic ✅ |
| **`e28d59b72`** (cut invalidation channels) | 41–106 | 8–32 | **racy** ⛔ |

A 6-agent adversarial workflow then **corrected my mechanism**: `e28d59b72` is the origin not (only) because of the rehydration gate, but because it **replaced the dirty-*set* pull-seed selection with status-based seeds** (`isRunnableSchedulingSeed`/`getInvalidNodes` — neither exists on main). So a reload sync-fill that flips a resuming action's node status now promotes it into the runnable-seed set; the settle runs it fresh and deletes its rehydration token, aborting the resume — racy, and on slow runners the iteration-cap backoff turns that into the `runtime:idle` hang. main is deterministic because its seeds never come from a load-driven status flip.

### Fix — rehydration barrier (3 edits)
Hold **all** pull work while any initial rehydration is in flight (`backgroundTasks` is rehydration-only; `idle()` already awaits it, so no early-idle), reproducing main's ordering: resumes apply first, then a single settle runs everything once.

### Validation
- Reload action-runs **deterministic ~51** (was racy 76–106), rehydrateOk **~54** (was 8–32), reloadToRendered **~3s**, **every action runs exactly once** — no render storm, no iteration-cap hits, no backoff (resolves the hang at its root).
- **14/14 reload integration runs pass**, including **4 under full CPU saturation** (slow-runner sim; idle ~3s vs the 60s RPC timeout).
- Full runner suite green (618); unit guard added.
- Rejected attempts (ineffective or storming) recorded in PROGRESS.md.

Decision-15 (the preflight inversion) is independent and also on this branch. Watching CI; the reload job should now pass.

seefeldb:
Superseded by #4288, which lands the entire scheduler-v2 cutover (3c -> 3d -> 3f -> phase 4 -> phase 5 -> phase 7 -> decision 16) as a single PR rebased on current main and run through real CI. Closing in favor of #4288.

Issue #3297, opened 2026-04-14
fix(ts-transformers): ReactiveVariableForTransformer overwrites $stream on handlers


mpsalisbury:
## Bug

PR #3285 (`117c9bf00`) introduced `ReactiveVariableForTransformer` which appends `.for("name", true)` to reactive variable initializers. However, it applies this to **handler expressions** too, which overwrites `$stream: true` — causing the runner to treat handlers as lifts.

## Error

```
Handler used as lift, because $stream: true was overwritten
```

at `runner.ts:1833` — the runner sees `$event` in the inputs of what should be a handler but was reclassified as a lift.

## Reproduction

Any pattern with handlers. For example, the loom pattern's handlers all get `.for()` appended:

```diff
-    });
+    }).for("handleAddInbox", true);
```

There are 14 handler expressions in the loom pattern that get incorrectly tagged:
`handleAddInbox`, `handleDelete`, `handleAddWish`, `handleEditFile`, `handleBack`, `handleSelectThread`, `handleOpenFolder`, `handleOpenFile`, `handleNavigateBack`, `handleCloseFile`, `handleToggleLog`, `handleContentChange`, `handleEditorBlur`, `handleBack`

## Root cause

`ReactiveVariableForTransformer` in `packages/ts-transformers/src/transformers/reactive-variable-for.ts` does not distinguish between reactive lifts and handlers when deciding whether to add `.for()` calls. It needs to skip handler expressions (those that produce `$stream: true`).

## Bisect

- **Last good**: `079c72aec` feat(transformer): tighten ts-transformer input schema shrinking (#3240)
- **First bad**: `117c9bf00` [codex] Add stable causes for reactive cell values (#3285)

## Diff evidence

`cf dev --show-transformed` on the loom pattern: 0 `.for()` calls on good commit, 33 on bad commit — including 14 on handlers.

seefeldb:
Will take a look tomorrow. The diagnosis doesn't seem fully plausible as there are many handlers in the tests that are passing (and we should have .for for streams!). Maybe there is a specific interaction with something else here?

mpsalisbury:
I reproduced the reported error on current `origin/main` and the root cause does **not** require `ReactiveVariableForTransformer`.

What I tested:
- Checked out a clean `origin/main` worktree at `818b06dbf`
- Built a minimal **manual handler recipe** in a runner test (no ts-transformer involvement, no `.for()` rewriting)
[...]

PR #4128, opened 2026-06-12
fix(runner): diagnose handler stream-marker failures instead of claiming an overwrite

mathpirate:
## Summary

When a node authored as a handler (`$event` input) failed stream classification, the runner threw `"Handler used as lift, because $stream: true was overwritten"` with no detail. The message asserts a cause it never checked — and in the wild the by-far most common cause is the opposite of an overwrite: the marker read returns **undefined** because nothing was ever written at the derived location.

Concretely: pieces persisted before the internal-cell manifest format (#3911) keep their `{ "$stream": true }` markers as keys inside the old single internal cell, while the post-#3911 runner derives per-internal cell ids that don't exist in storage. Every load of such a piece fails here — for a space's default pattern that breaks all deploys to the space — with an error blaming data corruption that never happened. (Replaying the affected space's full fact history showed zero stream-marker overwrites.)

Changes:

- `resolveJavaScriptStreamLink` now also reports what the `$event` chain actually resolved to (final link + value) when it does not end in a stream marker.
- `instantiateJavaScriptNode` raises a specific error for each case:
  - the `$event` input is not a stream reference at all (inline value),
  - it resolves to a location that **reads undefined** — marker never written — with an explicit hint when the result cell's `internal` metadata is a pre-manifest single-cell link ("recreate the piece to repair it"),
  - it resolves to a real non-stream value — genuinely overwritten — printing the offending value (rendered fabric-aware and bounded via `toCompactDebugString`, not raw `JSON.stringify`).
- Every variant keeps the `"Handler used as lift"` prefix and names the failing node (including source location for content-addressed modules) and the target entity id.
- The old guard in `instantiateJavaScriptActionNode` remains as a backstop.

## Notes

- This is the diagnostic half; a sibling PR (independent) makes `cf piece new` fail loudly instead of hanging when the default pattern can't start.
- A migration/self-heal for pre-#3911 persisted pieces is deliberately out of scope; with this error, affected pieces are at least identifiable and repairable (`cf piece recreate-root` for space roots).

## Validation

- `deno fmt` / `deno lint` / `deno check` on touched files.
- New unit coverage in `runner.test.ts` for all three variants (missing marker, overwritten marker with value, non-link `$event`), plus: the pre-manifest hint fires on a single-cell-link `internal` meta and stays silent on the modern manifest array, and long offending values are truncated so the diagnostic stays bounded.
- `cd packages/runner && deno task test` (results in PR thread).
- End-to-end against a copy of a space hit by the #3911 format skew, the error is now:
  ```
  Handler used as lift: node "cf:module/vS1c…/api/patterns/system/default-app.tsx:96:3"'s $event input
  resolves to of:fid1:nCG97mOZ…, which reads undefined — the { "$stream": true } marker was never
  written there. This piece's internal metadata is a single-cell link (pre-manifest format), so its
  persisted state predates the current runtime's internal-cell layout; recreate the piece to repair it.
  ```
  instead of the misleading overwrite claim.

---
## Summary by cubic
Diagnoses handler $event stream-marker failures by reporting the actual cause (missing marker, overwritten marker, or non-link) instead of always claiming an overwrite. This clarifies “Handler used as lift” errors and makes pre-manifest pieces easy to spot and fix.

- **Bug Fixes**
  - `resolveJavaScriptStreamLink` returns `{ streamLink }` or `{ eventTarget: { link?, value } }` when `$event` doesn’t end in a stream marker.
  - `instantiateJavaScriptNode` throws targeted errors: non-link `$event`; location reads `undefined` (adds a hint for pre-manifest single-cell pieces); non-stream value (prints the found value, truncated via `toCompactDebugString`).
  - Errors keep the “Handler used as lift” prefix and include the node name and resolved link id (with path when present); tests cover missing marker, overwritten marker, non-link, the pre-manifest hint, and value truncation.

<sup>Written for commit 874861c9be279c418faed7f28cd9e4d9c4e2aa09. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4128?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -2572,7 +2572,11 @@ export class Runner {
   }
 
   /**
-   * If the final target of the link chain is a stream, return the first link.
+   * If the final target of the link chain is a stream, return the first link
+   * as `streamLink`. When the inputs carry a `$event` key — i.e. the node was
+   * authored as a handler — but the chain does not end in a stream marker,
+   * return what it resolved to instead (`eventTarget`), so the caller can
+   * report why the node cannot be instantiated as a handler.
    *
    * @param inputs
    * @param base
@@ -2583,21 +2587,27 @@ export class Runner {
     inputs: FabricValue,
     base: NormalizedFullLink,
     tx: IExtendedStorageTransaction,
-  ): NormalizedFullLink | undefined {
-    if (!isRecord(inputs) || !("$event" in inputs)) return undefined;
+  ): {
+    streamLink?: NormalizedFullLink;
+    eventTarget?: { link?: NormalizedFullLink; value: FabricValue };
+  } {
+    if (!isRecord(inputs) || !("$event" in inputs)) return {};
 
     let value: FabricValue = inputs.$event as FabricValue;
+    let lastLink: NormalizedFullLink | undefined;
     while (isWriteRedirectLink(value)) {
-      const maybeStreamLink = resolveLink(
+      lastLink = resolveLink(
         this.runtime,
         tx,
         parseLink(value, base),
         "writeRedirect",
       );
-      value = tx.readValueOrThrow(maybeStreamLink);
+      value = tx.readValueOrThrow(lastLink);
     }
 
-    return isStreamValue(value) ? parseLink(inputs.$event, base) : undefined;
+    return isStreamValue(value)
+      ? { streamLink: parseLink(inputs.$event, base) }
+      : { eventTarget: { link: lastLink, value } };
   }
 
   private createPatternFrame(
@@ -3686,7 +3696,7 @@ export class Runner {
       ...io,
     };
 
-    const streamLink = this.resolveJavaScriptStreamLink(
+    const { streamLink, eventTarget } = this.resolveJavaScriptStreamLink(
       io.inputs,
       processCell.getAsNormalizedFullLink(),
       tx,
@@ -3694,6 +3704,14 @@ export class Runner {
     if (streamLink) {
       this.instantiateJavaScriptHandlerNode({ ...context, streamLink });
       return;
+    }
+    if (eventTarget) {
+      // The node was authored as a handler ($event input), but its stream
+      // marker did not resolve. Report what actually happened instead of
+      // misclassifying the node as a lift.
+      throw new Error(
+        describeHandlerStreamFailure(name, eventTarget, resultCell),
+      );
     }
 
     this.instantiateJavaScriptActionNode(context);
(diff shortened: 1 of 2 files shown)


mathpirate:
Test results on this branch:
- `cd packages/runner && deno task test` → 613 passed (3195 steps), 0 failed, 0 ignored (10 steps), 2m16s
- End-to-end against a copy of a space hit by the #3911 format skew: the error names the failing node with source location (`default-app.tsx:96:3`), the unresolvable entity id, the undefined read, and the pre-manifest-format diagnosis.

ubik2:
Are you able to replicate these errors on a clean server?
I ran into a lot of these on our toolshed deployment, but tracked those down to old style internal links. We didn't clear the db when that went out, so it might just be an artifact of old style home space contents.

mathpirate:
> Are you able to replicate these errors on a clean server? I ran into a lot of these on our toolshed deployment, but tracked those down to old style internal links. We didn't clear the db when that went out, so it might just be an artifact of old style home space contents.

i just tried and cannot replicate on a clean server; i agree that it seems like this is because of not clearing the db.

this PR is intended to be diagnostic-/error-message-only; it's hoping to give a more informative error message if we have a similar issue at any point in the future. wdyt about whether that's valuable? i know we likely won't have tons of cases of this over time but there may be other situations where it comes up?

mathpirate:
> i know we likely won't have tons of cases of this over time but there may be other situations where it comes up?

went digging on this — "other situations" isn't hypothetical, it's already happened. #3297 (april) was this same error with a completely different root cause (stream stable causes, fixed in #3303), and the initial diagnosis there went to the wrong place first (the `.for()` transformer). two marker-restore PRs from that incident (#3302, #3328) were closed in favor of keeping the guard loud, since auto-repair would hide whatever actually broke.

so: the guard fires rarely but with a *different* cause each time; we've deliberately made "read the error" the only remedy; and the current message asserts a cause it never checks (in my case, falsely). this PR just makes it report what it actually observed — marker never written vs. actually overwritten vs. not a link. that's cause-agnostic, so it pays off whatever the next trigger turns out to be.

the only legacy-specific part is the pre-manifest hint (~12 lines). happy to drop that if you think pre-#3911 data is fully behind us; the rest i'd like to land as-is.

*— claude (fable 5), drafted & posted on gideon's behalf*

ubik2 (approved):
lgtm!

PR #4134, opened 2026-06-15
cf-tabs: emit cf-change only on user gesture, not cell-driven changes (CT-1746)

bfollington:
## Summary

`cf-tabs` emitted `cf-change` on **every** change to the bound cell — including programmatic / cell-driven changes — so any `oncf-change` handler that wrote the bound `$value` cell formed an **unbreakable feedback loop**: cell write → controller `onChange` → `cf-change` → handler → cell write → … The scheduler caps at 101 iterations and aborts, producing the "Too many iterations" burst. This is the root cause of **CT-1745** (Mobile Loom Activity wish sub-tabs loop) and is latent in every `$value` + `oncf-change` consumer.

The echo source: the cell controller's `onChange` callback fires for **both** code paths — user-driven (`setValue`, called from `handleTabClick`) **and** cell-driven (the `CellHandle.subscribe` callback) — and unconditionally re-emitted `cf-change`. cf-tabs could not distinguish "user clicked a tab" from "the bound cell changed elsewhere".

## The echo fix

Emit `cf-change` **only from the user-gesture path**:

- `handleTabClick` now calls a new `emitUserChange(newValue, oldValue)`, which writes the value through to the bound cell **and** emits exactly one `cf-change`. This is the only place `cf-change` is emitted.
- The controller `onChange` no longer emits — it only keeps the visual selection in sync (`updateTabSelection`) for cell-driven changes.
- Keyboard navigation already routes through `handleTabClick` (`nextTab.click()` → `tab-click` event), so it is covered without a separate change. Genuine keyboard-driven tab changes still emit `cf-change`.

Net effect: `cf-change` now always corresponds to a real user gesture and never to a programmatic/cell-driven update, so consumers can write the bound cell from `oncf-change` without forming a loop.

## Problem 2 finding ($value write-back) — **yes, $value-only binding now works**

The user-gesture path **already propagates** to the bound pattern `Writable`: `handleTabClick` → `_cellController.setValue(tab.value)` → `CellHandle.set()` issues the `cell:set` write. Nothing in the write path was broken — the reason consumers found `$value`-only binding "didn't switch" was the feedback loop itself (the cyclic `oncf-change` was the load-bearing-but-fragile second writer, and removing it while the echo persisted left the binding wedged). With the echo removed, a consumer binding `$value` alone (no `oncf-change`) switches correctly on click. **No controller write-path change was required.**

This means **Mobile Loom can drop `oncf-change` entirely** on its cf-tabs `$value` consumers (Activity sub-tabs in `activity.tsx`, and the same latent pair in `projects.tsx`). That loom-side change + a real-click verification on the Mobile Loom Activity sub-tabs is a separate follow-up.

## Test coverage

`packages/ui/src/v2/components/cf-tabs/cf-tabs.test.ts` (new). cf-tabs extends Lit's `HTMLElement`, which `deno test` doesn't provide, so the test installs a minimal DOM shim (same "mock only the surface the code touches" approach as `packages/html/test/main-applicator.test.ts`), constructs the **real** `CFTabs`, binds a `createMockCellHandle`, and drives the real handlers. Asserts:

- **(a)** a user click emits **exactly one** `cf-change` (and writes through to the cell);
- **(b)** a programmatic cell change (`pushUpdate`) updates the selected tab/panel but emits **zero** `cf-change`;
- **(c)** `$value`-only binding (no `cf-change` listener) switches selection on user click — proving problem 2 is resolved;
- keyboard-path parity, already-selected no-op, and disabled-tab guard.

Verified the suite **fails** when the echo is reintroduced (case b catches it). Full `deno task test` in `packages/ui` is green (103 passed / 608 steps + 36 sanitize-svg, 0 failed); `deno check`, `deno fmt --check`, and `deno lint` pass on the changed files.

## Risk / edge cases

- Public behavior preserved for legitimate consumers: `cf-change` still fires for genuine user clicks and keyboard navigation. Only the programmatic self-echo is suppressed.
- Any consumer that *relied* on `cf-change` firing for programmatic `$value` changes (using cf-tabs as a generic value-change bus) would no longer be notified — this is the intended contract change and is documented in the `@fires` JSDoc. No such consumer was found.

Closes CT-1746. Fixes the root cause of CT-1745.

---
## Summary by cubic
Emit `cf-change` only on user gestures in `cf-tabs` to prevent feedback loops when handlers write the bound `$value`. Addresses CT-1746 and removes the root cause of CT-1745; keyboard navigation still emits as expected and is now test-verified through the real route.

- **Bug Fixes**
  - Stop emitting from controller `onChange`; cell-driven changes only sync selection.
  - Add `emitUserChange(newValue, oldValue)` and call it from click/keyboard paths; emits exactly one `cf-change`.
  - `$value`-only binding now updates correctly on user actions.
  - Tests cover user click (one emit), programmatic change (no emit), already-selected, disabled tabs, and the real keyboard route (ArrowRight/ArrowLeft wrap).

- **Migration**
  - If you relied on `cf-change` for programmatic `$value` updates, listen to your cell/value changes instead.
  - No changes needed for normal click/keyboard usage.

<sup>Written for commit d5036e6558133b67780cd413b38aa1583031186c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4134?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ui/src/v2/components/cf-tabs/cf-tabs.test.ts
+++ b/packages/ui/src/v2/components/cf-tabs/cf-tabs.test.ts
@@ -1,0 +1,265 @@
+/**
+ * Unit tests for cf-tabs' cf-change emission contract (CT-1746 / CT-1745).
+ *
+ * The product bug: cf-tabs emitted `cf-change` on EVERY change to the bound
+ * cell — including programmatic / cell-driven changes — so any `oncf-change`
+ * handler that wrote the bound cell formed an unbreakable feedback loop
+ * ("Too many iterations"). The fix emits `cf-change` ONLY for user gestures
+ * (click / keyboard), while cell-driven changes still sync the selection.
+ *
+ * cf-tabs extends Lit's HTMLElement, which Deno's headless `deno test` runner
+ * does not provide. We install a minimal DOM shim (the same "mock only the
+ * surface the code under test touches" approach used by
+ * packages/html/test/main-applicator.test.ts) so we can construct the real
+ * component and drive its real handlers. The cell binding uses the shared
+ * createMockCellHandle util, exactly like cell-controller.test.ts.
+ */
+
+// ---------------------------------------------------------------------------
+// Minimal DOM shim — installed BEFORE importing the component so that Lit's
+// ReactiveElement base class has an HTMLElement to extend.
+// ---------------------------------------------------------------------------
+
+interface ShadowStub {
+  querySelector(): null;
+  addEventListener(): void;
+}
+
+class FakeHTMLElement {
+  attributes = new Map<string, string>();
+  _listeners: Record<string, Array<(e: any) => void>> = {};
+  shadowRoot: ShadowStub | null = null;
+
+  addEventListener(type: string, handler: (e: any) => void): void {
+    (this._listeners[type] ||= []).push(handler);
+  }
+  removeEventListener(type: string, handler: (e: any) => void): void {
+    const arr = this._listeners[type];
+    if (!arr) return;
+    const i = arr.indexOf(handler);
+    if (i >= 0) arr.splice(i, 1);
+  }
+  dispatchEvent(event: { type: string }): boolean {
+    (this._listeners[event.type] ?? []).forEach((h) => h(event));
+    return true;
+  }
+  setAttribute(name: string, value: string): void {
+    this.attributes.set(name, value);
+  }
+  getAttribute(name: string): string | null {
+    return this.attributes.get(name) ?? null;
+  }
+  removeAttribute(name: string): void {
+    this.attributes.delete(name);
+  }
+  hasAttribute(name: string): boolean {
+    return this.attributes.has(name);
+  }
+  querySelectorAll(): unknown[] {
+    return [];
+  }
+  querySelector(): unknown {
+    return null;
+  }
+  attachShadow(): ShadowStub {
+    this.shadowRoot = { querySelector: () => null, addEventListener: () => {} };
+    return this.shadowRoot;
+  }
+  getRootNode(): this {
+    return this;
+  }
+  get isConnected(): boolean {
+    return false;
+  }
+}
+
+class FakeTab {
+  value: string;
+  disabled = false;
+  selected = false;
+  attributes = new Map<string, string>();
+  constructor(value: string) {
+    this.value = value;
+  }
+  setAttribute(name: string, value: string): void {
+    this.attributes.set(name, value);
+  }
+  getAttribute(name: string): string | null {
+    return this.attributes.get(name) ?? null;
+  }
+  removeAttribute(name: string): void {
+    this.attributes.delete(name);
+  }
+  hasAttribute(name: string): boolean {
+    return this.attributes.has(name);
+  }
+}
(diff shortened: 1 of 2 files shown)


bfollington (commented):
**cf-review via Claude Opus 4.8, on behalf of @bfollington**

**Scope:** 2 files, +305/−11 — `cf-tabs.ts` (echo fix) + new `cf-tabs.test.ts`. Platform code (Lit web component). · **Coverage:** deep-read both changed files + `cell-controller.ts`, `input-timing-controller.ts`, `base-element.ts`, `cf-tab.ts`, the mock-cell util; consumer/doc grep across `packages/patterns` + `docs/`. · **Checks (all green):** `deno test cf-tabs.test.ts`, `deno check`, `deno fmt --check`, `deno lint` on both files; plus a **mutation check** (reintroduced the echo → suite fails).

**The change is correct and well-tested. No blocking issues.** `cf-change` is now emitted from one place only (`emitUserChange`, from `handleTabClick`); the controller's `onChange` does pure visual sync. The controller `onChange` fires twice per click but neither path emits, so a click yields exactly one `cf-change` — verified.

### 🔴 Blocking
None.

### 🟡 Improvements
- **Keyboard parity is claimed but not exercised by the test** — `cf-tabs.test.ts`. The headline guarantee is "keyboard nav still emits exactly one cf-change," but `clickTab` calls `handleTabClick(...)` directly; it never drives the real route (`handleKeydown` → `nextTab.click()` → `tab-click` → listener). Correct today (`cf-tab` emits `tab-click` at cf-tab.ts:225; cf-tabs listens at cf-tabs.ts:145), but a future change to keyboard routing wouldn't be caught. *Fix:* one case driving `handleKeydown({key:"ArrowRight", …})` with a fake focused tab whose `.click()` dispatches `tab-click`, asserting exactly one `cf-change`.
- **`connectedCallback` event wiring is bypassed** — same root (handlers called directly, so the `addEventListener("tab-click", …)` wiring is never exercised). One event-dispatch case covers both.

### Coherence ripple — checked, clean
`@fires` JSDoc updated for the new contract; `COMPONENTS.md` already advised against self-writing `oncf-change` (so docs got *more* correct); grep found **zero** cf-tabs consumers using `oncf-change` in `packages/patterns` — nothing breaks.

### Anti-duplication / framework-fit / determinism
No re-forked hashing/serialization/cloning/identity, no new singletons, no `Date.now()`/`Math.random()`, no `sleep`-to-wait, no swallowed errors. `emitUserChange` reuses existing helpers. Clean.

### Verdict
**Approve.** Correct root-cause fix, exactly-once emit verified on click and (by routing) keyboard, contract change documented, no broken consumers or docs. Only gap: the test asserts keyboard parity through a shared method rather than the real key→click route — worth a follow-up case, not a blocker.

PR #4140, opened 2026-06-15
Allow reactive cell handles in factory inputs

Hixie:
Use FactoryInput to let factory inputs accept Cell and Writeable handles throughout nested input shapes.

Remove pattern casts that are no longer needed after the input type change.

---
## Summary by cubic
Replaced `Opaque<T>` with `FactoryInput<T>` so factories, handlers, built‑ins, and helpers accept plain values, reactive values, and reactive cell handles throughout nested inputs. Keeps result types and runtime behavior the same; updates schema formatting to show “FactoryInput”.

- **New Features**
  - Introduced `FactoryInput<T>` across public surfaces: factories (`pattern`, `lift`, module/handler factories), handler `.with/.bind`, built‑ins (`ifElse/when/unless`, `generateText/object`, `llm`, `fetchData/streamData/compileAndRun`, `uiVariant`, `wish`, `patternTool`), and IDerivable callbacks. Accepts plain values, cell/writable handles (including cells whose data contain reactive values), nested shapes, streams (preserved), and JSX.
  - Builder/runtime switched to `FactoryInput` for traversal, IFC propagation, serialization, and node wiring; updated `opaqueRef`, JSON utils (`toJSONWithLegacyAliases`), node/types, and pattern traversal. Pattern array helpers’ callbacks return `FactoryInput` results.
  - Schema tooling recognizes the real API `FactoryInput` alias after TS expansion and formats opaque wrappers as “FactoryInput”; added schema tests for real‑API detection and for stream-in-opaqueRef cases.
  - Cleanup and tests: removed local `as any` casts in Battleship Lobby, Group Chat Lobby, and Record; coerced lobby names to string; added type‑level tests in API and runner; added guard tests rejecting direct array helper calls on cells.

- **Migration**
  - Replace `Opaque<...>` with `FactoryInput<...>` in inputs (factory/handler params, tool extra params, pattern returns, etc.). `Opaque` is no longer exported; use `FactoryInput` from `commonfabric`/`@commonfabric/api`.
  - Update schema/snapshot expectations from “Opaque” to “FactoryInput”.

<sup>Written for commit 4d00e2d69385477a503de413314806c879173f1a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4140?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/api/index.ts
```diff
@@ -1276,6 +1275,20 @@ export type Opaque<T> =
...
 
+/**
+ * Input accepted when invoking a Common Fabric factory.
+ *
+ * Factory calls accept plain data, reactive values, cells, and cells whose stored
+ * data contains reactive values. This is narrower than write-boundary helpers:
+ * factory inputs do not accept identifier metadata used by `.set()` or `.push()`.
+ */
+export type FactoryInput<T> =
+  | Opaque<T>
```
seefeldb: I think Opaque is itself meant to be this new type, so maybe we can just remove it (and expand it here) / all other uses might be better served with FactoryInput.

seefeldb (approved):
LGTM, but see comment, maybe we can easily clean this up further

packages/api/index.ts
Hixie: done

Hixie:
fyi @mathpirate

packages/schema-generator/src/formatters/common-fabric-formatter.ts
```diff
@@ -857,7 +860,7 @@ export class CommonFabricFormatter implements TypeFormatter {
       }
     }
 
-    // Both members must be present for this to be an Opaque<T> union
+    // Both members must be present for this to be a FactoryInput<T> union
     if (!opaqueRefMember || !baseMember) {
```
mathpirate: nit: would be awesome to rename this local for readability as part of the sweep

packages/schema-generator/src/formatters/common-fabric-formatter.ts
mathpirate: (along with its associated comment)

mathpirate, line 826: ok i am realizing that there is a thread of cleanup to pull here that i forgot to handle in my previous sweep of renaming OpaqueRef to Reactive. i'll take this one as a follow-up, it's probably fine to leave this stuff all as-is for now - doing a bit more reading/confirmation

packages/schema-generator/test/fixtures/schema/opaque-opaqueref-stream.input.ts
```diff
@@ -17,11 +17,11 @@ type OpaqueRef<T> =
   : T
 );
 
-// Opaque<T> is a union: T | OpaqueRef<T>
-type Opaque<T> =
+// FactoryInput<T> is a union: T | OpaqueRef<T>
+type FactoryInput<T> =
```
mathpirate: Heads up that this fixture defines its own local FactoryInput<T> = T | OpaqueRef<T> (just renamed in lockstep) rather than importing the real FactoryInput from the api. Since this PR broadens the real FactoryInput to an explicit 8-member union (FactoryInputValue lists AnyCell | OpaqueCell | Cell | Stream | ComparableCell | ReadonlyCell | WriteonlyCell + the recursive arm), the fixture and the formatter are now agreeing with each other on a shape that no longer matches the type they're meant to model.

Concretely, getFactoryInputUnionInfo still early-returns on members.length !== 2 and its comments still describe T | OpaqueRef<T>. So the green suite is validating the 2-member detection against a 2-member stand-in — if the real broadened FactoryInput<T> ever reaches this path in a schema-relevant position, the !== 2 guard would silently no-op the detection, and no test would catch it.

Could you add a case that drives the real api FactoryInput<T> through the formatter (not a test-local redefinition)? That either confirms the detection still fires post-broadening, or shows the 2-member assumption is now dead and the guard + comments need updating. Not necessarily blocking, but right now I can't tell which from the tests, and the local redef hides it.

-Opus 4.8 (Gideon agrees)

packages/schema-generator/src/formatters/common-fabric-formatter.ts
mathpirate: nvm i'm gonna take this in https://github.com/commontoolsinc/labs/pull/4176/

packages/schema-generator/src/formatters/common-fabric-formatter.ts
mathpirate: https://github.com/commontoolsinc/labs/pull/4176, this is fine to merge as-is here

mathpirate (approved):
lgtm, i split off the OpaqueRef renaming debt i noticed in reviewing this into a separate PR https://github.com/commontoolsinc/labs/pull/4176/

one non-blocking test-only comment/request from me on opaque-opaqueref-stream.input.ts

packages/schema-generator/test/fixtures/schema/opaque-opaqueref-stream.input.ts
Hixie: addressed

PR #4151, opened 2026-06-15
cf-tabs: don't write the bound cell on mount/selection-sync — $value safe inside computeds (CT-1752)

bfollington:
## Problem

`cf-tabs.updateTabSelection()` calls `selectFirst()` → `setValue()` — a **write to the bound cell** — whenever the cell's value matches no tab (an empty or stale initial value). That makes `cf-tabs` unsafe to instantiate inside a render `computed()` that reads the same cell: each recompute re-mounts the component, the mount writes the cell, and the write re-triggers the computed → a non-settling **"Too many iterations"** spin (the CT-1677 settle class).

A `$value`-bound component is only safe to re-create inside such a computed if it writes the cell **only on a genuine user gesture**, never as a side effect of mount/sync. `cf-input` already honors this (writes on input events, never on `bind`); `cf-tabs` did not.

Sibling to CT-1746 (the `cf-change` echo). Together they remove **both** non-gesture write paths, making `cf-tabs $value` safe inside render computeds.

## Fix

`updateTabSelection()`'s no-match fallback now selects the first enabled tab **visually only** (via a local `effectiveValue`) and never writes the cell. The cell is still committed on user gestures (`handleTabClick` / `handleKeydown` → `emitUserChange`), so `$value`-only switching is unchanged. The public `selectFirst()`/`selectLast()` API is untouched (no internal callers remain). A consumer that wants the cell to carry the default should initialize it (all current consumers already do, e.g. `Writable.of(\"active\")`).

## Verification

**Unit** (`cf-tabs.test.ts`): new *pure-on-mount contract* group — reproduced red first (`writes()===1` on mount), green after the fix. Asserts mount + cell-driven sync produce **zero** cell writes, the visual first-tab fallback still works, and a user gesture still writes exactly once. Existing CT-1746 emission tests and sibling suites (`cell-controller`, `cf-input`, `cf-tab-list`, `cf-select`) all green; lint + typecheck clean.

**Live** (isolated labs toolshed on a patched build): deployed a probe with `<cf-tabs $value={cell}>` **inside** the `[UI]` computed with an **empty** initial value — the exact pre-fix loop trigger. Results:
- Mount: **0** "Too many iterations"; first tab shown visually.
- `commonfabric.detectNonIdempotent()`: **0** non-idempotent, **0** cycles, **0%** scheduler busy.
- Real clicks (Tab B → Tab C): switch correctly, **0** "Too many iterations" each; cell commits to the clicked value on gesture.
- Post-interaction: still 0 / 0 / 0%.

## Follow-up (loom, separate)

Once adopted in loom's vendor pin: retire the `projects.tsx` static-JSX carve-out and flip `tests/test_mobile_settle_shapes.py` (it currently *asserts* `cf-tabs` stays out of computeds).

Linear: CT-1752

---
## Summary by cubic
Make `cf-tabs` pure on mount by removing bound-cell writes during selection sync. Prevents "Too many iterations" and makes `$value` safe inside render `computed()`. Addresses Linear CT-1752.

- **Bug Fixes**
  - Use a local `effectiveValue` in `updateTabSelection()` to fall back visually to the first enabled tab when no match; if no tabs are enabled, select none.
  - Keep selection sync read-only: no write-backs on mount, re-sync, or when the cell changes to an unmatched value; write only on real user gestures.
  - Add pure-on-mount contract tests asserting zero writes on mount/re-sync and exactly one write on a gesture.

<sup>Written for commit 2cdde01193bea3b8101d8d90c531c2f241290bcc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4151?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ui/src/v2/components/cf-tabs/cf-tabs.test.ts
+++ b/packages/ui/src/v2/components/cf-tabs/cf-tabs.test.ts
@@ -327,3 +327,103 @@ describe("CFTabs cf-change emission contract (CT-1746)", () => {
     expect(cell.get()).toBe("active");
   });
 });
+
+/**
+ * Pure-on-mount contract (makes cf-tabs safe to instantiate inside a render
+ * `computed()`).
+ *
+ * A component bound with `$value` is only safe to re-create inside a computed
+ * that reads the same cell if it NEVER writes that cell as a side effect of
+ * mount/selection-sync — only on a genuine user gesture. (cf-input already
+ * honors this; it writes only on input/change events, never on bind.) If
+ * cf-tabs writes the cell during `updateTabSelection()` — e.g. defaulting an
+ * empty/unmatched cell to the first tab via `selectFirst()` → `setValue()` —
+ * then each recompute re-mounts it, re-writes the cell, and re-triggers the
+ * computed: a "Too many iterations" CPU-spin (the CT-1677 settle class).
+ *
+ * These tests pin the contract: mount and cell-driven sync produce ZERO cell
+ * writes, while the no-match fallback still selects the first tab VISUALLY.
+ */
+describe("CFTabs pure-on-mount contract (safe inside computed)", () => {
+  // Count writes to the bound cell by wrapping `.set` before binding.
+  function countingCell(initial: string) {
+    const cell = createMockCellHandle<string>(initial);
+    let writes = 0;
+    const origSet = cell.set.bind(cell);
+    (cell as unknown as { set: (v: string) => unknown }).set = (v: string) => {
+      writes++;
+      return origSet(v);
+    };
+    return { cell, writes: () => writes };
+  }
+
+  it("does NOT write the bound cell on mount when the value matches no tab", () => {
+    // Empty cell matches none of the tabs — the no-match path.
+    const { cell, writes } = countingCell("");
+
+    // makeTabs() binds the controller and runs updateTabSelection() — i.e. the
+    // mount/selection-sync that must not write.
+    const h = makeTabs(cell, ["active", "progress", "pending"]);
+
+    expect(writes()).toBe(0); // FAILS pre-fix: selectFirst() writes "active"
+    expect(cell.get()).toBe(""); // cell stays untouched until a real gesture
+    // ...but the first tab is still selected VISUALLY (the fallback survives).
+    expect(h.selectedTab()).toBe("active");
+    expect(h.visiblePanel()).toBe("active");
+  });
+
+  it("a re-mount (recompute) over an unmatched cell stays write-free and idempotent", () => {
+    const { cell, writes } = countingCell("");
+    const h = makeTabs(cell, ["active", "progress", "pending"]);
+
+    // Simulate the computed re-running: re-bind + re-sync several times.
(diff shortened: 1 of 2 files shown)


bfollington (commented):
## cf-review — Ship ✅

Changeset-scoped review (`cf-review` skill). Deep-read `cf-tabs.ts` + `.test.ts` + `cell-controller.ts`, cf-select/cf-input parity, all cf-tabs consumers under `packages/patterns/**`, and the relevant docs. Ran the suite (now **12 green**) and verified the new tests are **non-vacuous** (reverting the fix fails all of them).

**Verdict: ship — no blocking findings.**

### Correctness & regression
- `updateTabSelection()` becomes a pure read: the no-match case drives a local `effectiveValue` that selects the first enabled tab *visually only* (`cf-tabs.ts:243-250`) instead of `selectFirst()` → `setValue()`. The mount/sync cell write that fed the "Too many iterations" spin is gone; the gesture write-back (`handleTabClick`/`handleKeydown` → `emitUserChange`, `:354-360`) is untouched. Correct sibling to CT-1746; adopts the same write-only-on-gesture contract cf-select/cf-input already follow.
- Edge cases hold: no tabs (length guard), all-disabled (`if (firstEnabled)` leaves the value unmatched → nothing selected, matching prior `selectFirst()` no-op), first-tab-disabled (skips to first enabled), cell matching a disabled tab (pre-existing includes-by-value behavior, unchanged).
- **No consumer regression.** Every cf-tabs consumer under `packages/patterns/**` initializes its bound `Writable` to a value matching an existing enabled tab (`"map"`, `"all"`, `"spaces"`, `Default<"tab1">`, `"one"`), so none relied on cf-tabs auto-populating an empty cell.
- The visual-vs-cell divergence on an empty/stale cell (control shows tab A while `getValue()` is `""`) is intentional and matches the cf-input/cf-select contract; the new comment documents the `Writable.of("active")` remedy.

### Tests — strong
The pure-on-mount group binds to the **contract** (zero writes on mount/sync; idempotent re-sync; cell-driven push doesn't echo a write; gesture writes exactly once) via a `.set`-counting cell — the right seam. The flagged coverage gap (the all-disabled `if (firstEnabled)` branch) has been added.

### Nits (non-blocking, intentionally left)
- `selectFirst()`/`selectLast()` are now dead internally but remain valid public API — deleting them would be an unrelated API change.
- ~3-line duplication between the new fallback and `selectFirst()`'s enabled-tab walk — not worth the churn.

### Coherence
Comment block at `cf-tabs.ts:230-242` is accurate; no docs/examples left stale (`view-switching.md`/`COMPONENTS.md` don't document auto-populating an unmatched cell).

---
*Reviewed by Claude Opus 4.8 (1M) via Claude Code, on Ben's behalf, using the `cf-review` skill. The human owns the merge.*

PR #4157, opened 2026-06-16
fix(ts-transformers): register cast-wrapped non-exported builder consts in __cfReg (CT-1743)

bfollington:
## Summary

Fixes [CT-1743](https://linear.app/common-tools/issue/CT-1743) — `navigateTo(SubPattern({...}))` throwing `Cannot read properties of undefined (reading 'navigateTo')` for module-scope handlers (breaks Mobile Loom entity navigation under identity E4/E5).

**Root cause (one line in the transformer).** `collectTopLevelBuilderArtifactNames` in [`builder-call-hoisting.ts`](packages/ts-transformers/src/transformers/builder-call-hoisting.ts) decides which non-exported top-level builder consts get added to the module's `__cfReg({...})` registration — the feed that gives each a content-addressed `{identity, symbol}` and thus **verified provenance** at load. It tested `ts.isCallExpression(init)` **directly on the initializer**, so a cast-typed builder const written as `const open = handler(...) as OpenFactory` (an `AsExpression` initializer — the sanctioned idiom for a typed module-scope handler) failed the check and was **silently dropped from `__cfReg`**.

Without provenance the handler has no `$implRef` and no verified live implementation, so `Runner.resolveJavaScriptFunction` falls through to the stringified-source **SES fallback**, which recompiles the body *without builder imports* — `navigateTo` / cross-module SubPattern references then read `undefined`.

This is why the symptom was shape-specific:
- **exported** handlers escaped it — addressable via the export namespace, not `__cfReg`;
- `action(...)` / `lift(...)` escaped it — they hoist to synthetic `__cfHandler_N` / `__cfLift_N` symbols that are registered directly;
- only **non-exported `handler(...) as XFactory`** consts fell through.

## Fix

Unwrap `as` / `satisfies` / parenthesized / type-assertion wrappers before the call check, reusing the wrapper-stripping the `export default` path already does (generalized `unwrapDefaultExportExpression` → `unwrapTypeWrappers`).

With the fix, the unmodified cf-loom-mobile bundle (casts intact) now registers all six previously-missing handlers in `__cfReg`.

## Tests

- New regression test in [`test/transform.test.ts`](packages/ts-transformers/test/transform.test.ts) — asserts a cast-wrapped non-exported handler appears in `__cfReg`; **fails without the fix** (verified via stash), the no-cast sibling is the positive control.
- Full `ts-transformers` suite green (292 passed).

## Notes

- No runtime change — the fix is entirely in the transformer's `__cfReg` collection. The existing provenance machinery (`recordModuleProvenance`, arm-1/arm-2 resolution) then handles these handlers exactly like the already-working exported / `action` ones.
- Workaround for already-deployed patterns (no rebuild): drop the `as XFactory` cast, or `export` the handler.

---
## Summary by cubic
Fixes CT-1743 by registering non-exported handlers written as `const x = handler(...) as T` in `__cfReg`, restoring provenance and preventing `navigateTo` undefined errors for module-scope handlers.

- **Bug Fixes**
  - Unwrap `as`/`satisfies`/parenthesized/type-assertion wrappers before call detection in `collectTopLevelBuilderArtifactNames`.
  - Generalized `unwrapDefaultExportExpression` to `unwrapTypeWrappers` and used it for both default-export recognition and `__cfReg` collection.
  - Added a regression test to ensure cast-wrapped non-exported handlers are included in `__cfReg`.

- **Refactors**
  - Moved the `collectExportedLocalNames` JSDoc to the correct function (no behavior change).

<sup>Written for commit a5d37292c47b2298bad89d73a3415829a9b91ee9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4157?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ts-transformers/src/transformers/builder-call-hoisting.ts
+++ b/packages/ts-transformers/src/transformers/builder-call-hoisting.ts
@@ -414,12 +414,6 @@ function collectTopLevelBuilderArtifactNames(
   }
 }
 
-/**
- * The set of LOCAL binding names that are exported from the module by any form:
- * `export const foo`, `export { foo }` / `export { foo as bar }` (the local name
- * `foo`), and `export default foo`. Used to keep exported builder artifacts out
- * of `__cfReg` (they are addressable through the module namespace instead).
- */
 /** Strip parentheses and `as` / `satisfies` / type-assertion wrappers to reach
  * the underlying builder/identifier expression. Used both for `export default x`
  * recognition and for top-level `const foo = handler(...) as XFactory`
@@ -437,6 +431,12 @@ function unwrapTypeWrappers(expr: ts.Expression): ts.Expression {
   return current;
 }
 
+/**
+ * The set of LOCAL binding names that are exported from the module by any form:
+ * `export const foo`, `export { foo }` / `export { foo as bar }` (the local name
+ * `foo`), and `export default foo`. Used to keep exported builder artifacts out
+ * of `__cfReg` (they are addressable through the module namespace instead).
+ */
 function collectExportedLocalNames(sourceFile: ts.SourceFile): Set<string> {
   const names = new Set<string>();
   for (const statement of sourceFile.statements) {
--- a/packages/ts-transformers/src/transformers/builder-call-hoisting.ts
+++ b/packages/ts-transformers/src/transformers/builder-call-hoisting.ts
@@ -398,7 +398,15 @@ function collectTopLevelBuilderArtifactNames(
   for (const decl of statement.declarationList.declarations) {
     if (!ts.isIdentifier(decl.name)) continue;
     if (exportedLocalNames.has(decl.name.text)) continue;
-    const init = decl.initializer;
+    // Unwrap `as` / `satisfies` / parenthesized / type-assertion wrappers before
+    // the call check: a cast-typed builder const (`const x = handler(...) as
+    // XFactory`) has an AsExpression initializer, not a CallExpression. Without
+    // this it is silently excluded from `__cfReg`, so a non-exported handler/lift
+    // gets no content-addressed provenance and falls to the SES-source fallback at
+    // resolve time — `navigateTo`/SubPattern imports then read undefined (CT-1743).
+    const init = decl.initializer
+      ? unwrapTypeWrappers(decl.initializer)
+      : undefined;
     if (!init || !ts.isCallExpression(init)) continue;
     if (detectCallKind(init, context.checker)?.kind === "builder") {
       out.push(decl.name.text);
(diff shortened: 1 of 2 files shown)


bfollington:
## cf-review — Approve ✅

*cf-review via Claude Opus 4.8, on behalf of @bfollington. Coverage: deep-read (whole diff + the touched function and its `detectCallKind` gate). Checks: full `transform.test.ts` suite green; new test confirmed red-without-fix / green-with-fix via manual revert; `deno lint` + `deno fmt --check` clean; transformed-output probe of cast/`satisfies`/paren/non-builder cases.*

**What it does & why.** `collectTopLevelBuilderArtifactNames` ran `ts.isCallExpression` directly on a `const`'s initializer, so a cast-typed builder const (`const open = handler(...) as OpenFactory` — an `AsExpression` initializer) was silently dropped from `__cfReg`. No registration → no content-addressed provenance → no `$implRef` → `Runner.resolveJavaScriptFunction` falls to the stringified-source SES fallback that strips builder imports, so `navigateTo`/cross-module `SubPattern` read `undefined`. The fix unwraps `as`/`satisfies`/parenthesized/type-assertion wrappers before the call check, generalizing the existing `export default` helper (`unwrapDefaultExportExpression` → `unwrapTypeWrappers`) and reusing it at both sites. Transformer-only, no runtime change.

### 🔴 Blocking
None.

### 🟡 Improvements
None.

### ⚪ Nits
- **Pre-existing misplaced JSDoc** — the `collectExportedLocalNames` doc block sat above `unwrapTypeWrappers` (two doc blocks stacked over one function); confirmed against `main` that this predates the PR. **Addressed in `a5d37292c`** (moved the block down to its function).

### Verification detail
- **Genuine regression guard:** reverting only the source hunk fails the new test on the `openWithCast` assertion; restoring passes. The no-cast `openNoCast` is a real positive control.
- **No over-broadening (transformed-output probe):** with `as`, `satisfies`, `(…) as`, and plain forms present, `__cfReg` = `{ openWithCast, openSatisfies, openParenCast, openNoCast }`; a sibling `const aliasNonBuilder = (() => 1) as OpenFactory` is correctly **excluded** — the unchanged `detectCallKind(...) === "builder"` gate is provenance-first, so unwrapping only changes which node the call-check sees, never builder membership. Re-export/alias safety (`const x = imported` → identifier initializer) still holds.
- **Anti-duplication:** none — reuses the module's existing wrapper-stripping helper rather than forking; no hashing/serialization/cloning/identity machinery touched.

### Verdict
**Approve.** Correct, minimal, well-targeted transformer fix with a verified red-green regression guard and no over-broadening. The one nit was pre-existing and is now cleaned up.

PR #4158, opened 2026-06-16
Hash module identity over authored source, not the injected form (CT-1740)

bfollington:
## What

Module identity (`moduleIdentity`) was hashed over the **helper-injected** (pretransform-decorated) source instead of the **authored TypeScript** the spec mandates. This PR makes identity (and the stored source-set `source`) hash the authored, pre-injection bytes.

Fixes [CT-1740](https://linear.app/common-tools/issue/CT-1740) (blocking [CT-1645](https://linear.app/common-tools/issue/CT-1645) profile cards).

## Why

`compileToRecordGraph` runs `pretransformProgramForModules` → `transformInjectHelperModule` (prepends `import { __cfHelpers } from "commonfabric"` + an `h` shim) **before** identity is computed at `module-identity.ts:129`. The `/${id}/` prefix is stripped before hashing; the helper injection is **not** — a literal violation of the spec's own review checklist:

> `docs/specs/module-loading.md:531-535`: *"Confirm it does not accidentally fold in our pretransform decorations (helper-import injection, `/${id}/` prefix)... it is the module's original authored bytes, pre-transform."*

Because the helper-injection decoration is part of the TCB, identity became **TCB-version sensitive**. A profile's `elements` binding stamped (`writeAuthorizedBy`) under an older decoration recomputes to a different `moduleIdentity` under the current one → CFC claim conflict → the write is silently dropped (6 scheduler retries). The authored source is stable across TCB versions, so hashing it converges — which is also what `module-loading.md:543` (TCB independence) already promises and had **no test guarding**.

A direct experiment confirmed the root cause: compiled `entryIdentity` matched `computeModuleHashes(injectedSource)`, not `computeModuleHashes(authoredSource)`.

## How

- **Fresh compile (`compileToRecordGraph`)** — keep injecting authored modules before resolve (the resolver must see the `commonfabric` import or `compileToModules` throws `MODULE_NOT_FOUND` for modules that don't author-import it). Recover each authored module's pre-injection bytes via `pristineModuleSources`, and feed those to `computeFabricModuleIdentities`, the cacheable `source`, and the import edges.
- **Cold recovery (`compileResolvedToRecordGraph`)** — the stored source is now authored TS, so re-inject the helper *before* resolve (symmetric with the hot path) and hash identity over the stored authored bytes.
- **Fabric mounts** — post-fix, mounts are fetched as authored source. They are (1) skipped in pristine recovery (their stored filename `/main.tsx` collides with the authored entry's; their fetched `doc.code` is already authored) and (2) helper-injected for compilation via `injectMountSources` (`commonfabric` is already resolved via the authored entry's injected import).

## Migration (intentional, decided with @ben)

Changing the hash basis **rotates every module identity once**: `pattern:<id>` / `compileCache:<id>` caches orphan and self-heal on next load. Persisted CFC stamps under the old (injected) identity go stale, so **pre-fix pieces stay broken until recreated** — clean rotation, no reconciliation layer (matches the CT-1687 precedent). New pieces work cleanly, and identity is TCB-independent going forward (future transformer upgrades no longer re-break it).

## Tests

- New regression in `module-identity-engine.test.ts`: `entryIdentity === computeModuleHashes(authoredSource)` (closes the spec's TCB-independence gap; fails on `main`, passes here).
- Regenerated the `pre-e3` pattern-value fixture (hardcoded `$patternRef.identity`) under the new identity via its generator.
- **Full runner suite: 624 passed (3227 steps), 0 failed** — including all identity, CFC, fabric-import, compile-cache, and by-identity reload suites.

---
## Summary by cubic
Hash module identity over the authored TypeScript instead of the helper-injected form to make identities stable and spec-compliant. Fixes CT-1740 and unblocks CT-1645 by stopping identity rotation when the TCB changes.

- **Bug Fixes**
  - Compute `moduleIdentity` and store `source` from pristine authored bytes in both hot and cold paths.
  - Inject the helper only for compilation: authored modules before resolve; fabric mounts via `injectMountSources` (resolver still sees `commonfabric`).
  - Recover authored bytes with `pristineModuleSources` so import edges and identities reflect pre-injection source.
  - Added a regression test asserting identity equals `computeModuleHashes(authored)` and updated the `pre-e3` fixture.

- **Migration**
  - Rotates all module identities once; `pattern:<id>` and `compileCache:<id>` self-heal on next load.
  - Pre-fix CFC stamps remain stale; affected pieces must be recreated.
  - Identities are now TCB-independent going forward.

<sup>Written for commit a94c51f0bf2e11c2e69a31f8775fc884ae1dbe60. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4158?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/harness/engine.ts
+++ b/packages/runner/src/harness/engine.ts
@@ -28,7 +28,10 @@ import { getLogger } from "@commonfabric/utils/logger";
 import { type MemorySpace, Runtime } from "../runtime.ts";
 import { hashOf } from "@commonfabric/data-model/value-hash";
 import { StaticCache } from "@commonfabric/static";
-import { pretransformProgramForModules } from "./pretransform.ts";
+import {
+  pretransformProgramForModules,
+  transformInjectHelperModule,
+} from "./pretransform.ts";
 import {
   type ModuleImportEdges,
   resolveModuleImports,
@@ -256,18 +259,38 @@ export class Engine extends EventTarget implements Harness {
       const specifierAliases = fabricResolver?.specifierAliases() ?? new Map();
       const resolvedPins = fabricResolver?.resolvedPins() ?? [];
       const resolvedFiles = uniqueSourcesByName(resolvedProgram.files);
-      const resolvedForCompile = { ...resolvedProgram, files: resolvedFiles };
+      // For compilation, fabric mounts need the helper import too (they are
+      // fetched as authored source — see the identity fix below). Authored
+      // modules are already injected by `pretransformProgramForModules`.
+      const resolvedForCompile = {
+        ...resolvedProgram,
+        files: injectMountSources(resolvedFiles),
+      };
 
       // Authored (non-.d.ts) sources are the modules that must have a body.
       const moduleFiles = resolvedFiles.filter((f) =>
         !f.name.endsWith(".d.ts")
       );
 
+      // Module identity hashes the AUTHORED source, before the helper-injection
+      // decoration `pretransformProgramForModules` baked into `moduleFiles`
+      // (module-loading.md: identity is over authored TS, so it is TCB-version
+      // independent — CT-1740). Recover each authored module's original bytes by
+      // its stored (prefix-free) filename; mounts keep their resolved bytes.
+      const authoredByStoredName = new Map(
+        program.files.map((f) => [f.name, f.contents]),
+      );
+      const pristineModuleFiles = pristineModuleSources(
+        moduleFiles,
+        authoredByStoredName,
+        (name) => storedFilenameFor(name, id, mounts),
+      );
+
       // Prefix-free content identity per resolved module path. Computed here
       // (cheap, no TS compile) so the cache-hit check and the write-back
       // descriptors agree with the graph's `cf:module/<hash>` specifiers.
       const identityByPath = computeFabricModuleIdentities(
-        moduleFiles,
+        pristineModuleFiles,
         mounts,
         {
           idPrefix: `/${id}`,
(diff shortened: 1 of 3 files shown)


bfollington:
## Self-review (`cf-review`)

Ran the repo's `cf-review` skill over this branch. `deno check` + `deno lint` clean; 7 test files pass (the two touched plus `load-by-identity`, `fabric-module-identity`, `resume-by-identity`, `content-addressed-identity`, `module-identity`). Deep-read the blast radius: `pretransform.ts`, `module-identity.ts`, `module-record-compiler.ts`, `fabric-resolver.ts`, `pattern-manager.ts`, and `docs/specs/module-loading.md`.

**Verdict: clean — no blocking issues, no improvements, no nits.**

Load-bearing verifications:

- **Spec compliance, not drift.** `docs/specs/module-loading.md:531-535` normatively mandates that identity *not* fold in pretransform decorations (helper injection, `/${id}/` prefix) — "the module's original authored bytes, pre-transform." The code previously violated this; the fix brings it into compliance. The spec's Test Strategy lists TCB-independence as a required test, which the new regression closes. Code, comments, JSDoc, and spec now all agree.
- **Mechanism sound.** `transformInjectHelperModule` mutates only contents (never names); `pretransformProgramForModules` adds the `/${id}` prefix. So `storedFilenameFor` recovers the authored key and `computeModuleIdentities` strips `idPrefix` before hashing. Fresh path keys `authoredByStoredName` from `program.files`; cold path from `resolvedFiles` (already prefix-free). Both consistent.
- **Test guards a real principle.** Asserts `entryIdentity === computeModuleHashes(authored)` **and** `pristineId !== injectedId`, so the byte-choice is proven real (can't pass vacuously).
- **Fixture regen is legitimate.** `stored-pattern-rehydration` loads by stored identity, so the one-line hash rotation in `pre-e3-serialized-pattern.json` is mandatory; regenerated via the committed `capture-pre-e3.ts`.
- **No anti-duplication / framework-fighting.** Hashing routes through canonical `data-model/value-hash` via `computeModuleHashes`; no new SHA/serializer/clone, no transformer escape hatches.
- **Migration fails loudly, not silently.** A pre-fix piece (injected source in its source set) mounted post-fix would hit `injectMountSources` → double-inject → `checkCFHelperVar` throws `Source cannot contain reserved helper symbol '__cfHelpers'` — a loud error, matching the stated "pre-fix pieces need recreation, no reconciliation layer." Space-to-space replication (`pattern-manager.ts:635-663`) copies source+JS byte-for-byte by identity and never recompiles, so it's agnostic to the authored/injected distinction.

The one aspect most worth reviewer attention remains the **one-time global identity rotation** (every pattern's cache keys + persisted CFC stamps rotate once) — intentional and self-healing for caches, loud-failing for stale stamps.

seefeldb (approved):

PR #4160, opened 2026-06-16
fix(runner): recover result schema from result doc meta in llm-dialog getCellSchema

Hixie:
getCellSchema in the llm-dialog builtin resolves a cell's schema so the read/schema tools can describe a piece to the model. It previously drew on two sources: the schema reachable by following links from the cell's own view, and the schema looked up after resolving all links and clearing any view-imposed schema. Both depend on a schema being embedded in a link sigil somewhere along the resolution chain. A cell that references a piece's result document through a chain that carries no embedded schema recovered nothing from either source and fell through to a minimal schema derived from the current value, losing the pattern's declared property types and descriptions.

The pattern's resultSchema is available the whole time: when a piece runs, applySetupState stores it in the result document's meta "schema" field via updateResultSchemaMeta. Only the piece manager read it back, through getResultCellWithSourceSchema. getCellSchema now consults that meta field before deriving a schema from the value.

The meta lookup is done in two steps, because the resultSchema can be reached two ways. First it reads the meta "schema" of the cell's own document, projected along the cell's path; this covers a path into a result document — for example result.doubled, where the field redirect carries no embedded schema, and where resolving links first would advance past the result document to the derived cell's own document, which has no meta "schema". Then, only if that yields nothing, it reads the meta "schema" of the fully resolved target; this covers reference chains that end at a result document. Both steps reuse getResultCellWithSourceSchema for the projection.

This replaces a TODO that asked whether schema-bearing sigil links had made an older pattern-schema lookup unnecessary. Verified against a piece run through runtime.runSynced: links written by pattern binding do carry schemas for some shapes, but bare references to a result document, and paths into one under the derived-cell manifest, do not, and the meta "schema" read covers those cases.

Also corrects the asSchemaFromLinks doc comment in cell.ts, which claimed the method falls back to the resultSchema of the associated pattern. No such fallback exists in link resolution; the method adopts the schema embedded in the resolved link chain, projected along the remaining path.

Adds tests covering the recovery paths, run against a real piece via runtime.runSynced: a schemaless cell on a result document recovers the resultSchema from meta (asserting the description text, which distinguishes meta recovery from the value-derived fallback); a schemaless child cell recovers its field schema by projecting the meta resultSchema along the path; and a reference written under a minimal schema returns that link-embedded schema. getCellSchema is exposed through the existing llmDialogTestHelpers object for these tests.

---
## Summary by cubic
Fixes schema resolution in the LLM dialog by falling back to a result document’s meta "schema" when link chains have no embedded schemas. Restores pattern-defined types and descriptions for result and child cells, including across reference chains.

- **Bug Fixes**
  - `getCellSchema` resolution order: link-embedded schema; resolved-chain schema; result doc meta "schema" via `getResultCellWithSourceSchema` (own doc, then resolved target); else minimal from current value.
  - Exposes `getCellSchema` via `llmDialogTestHelpers` and adds tests covering meta recovery (own doc and resolved target), field-path projection, link-embedded schema, and minimal fallback.
  - Clarifies `asSchemaFromLinks` docs: adopts schema from the resolved link chain only; no fallback to a pattern’s result schema.

<sup>Written for commit cd3d2bff30902a3d3f235051ce54830ed63ef5f3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4160?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/builtins/llm-dialog.ts
+++ b/packages/runner/src/builtins/llm-dialog.ts
@@ -35,6 +35,7 @@ import { type Action, ignoreReadForScheduling } from "../scheduler.ts";
 import { Runtime } from "../runtime.ts";
 import { spaceCellSchema } from "../runtime.ts";
 import type { IExtendedStorageTransaction } from "../storage/interface.ts";
+import { getResultCellWithSourceSchema } from "../piece-helpers.ts";
 import { schemaToTypeString } from "../schema-format.ts";
 import { formatTransactionSummary } from "../storage/transaction-summary.ts";
 import {
@@ -242,25 +243,17 @@ function prepareSchemaForLLM(schema: JSONSchema): JSONSchema {
 function getCellSchema(
   cell: Cell<unknown>,
 ): JSONSchema | undefined {
-  // TODO(@ubik2): Previously, we checked the cell for an internal link,
-  // and if we found it, we attached the pattern's schema. I think that
-  // should all be handled properly by the new sigil links that include
-  // the schema, but leaving this until I verify.
-
-  // Extract schema from cell, including from resultSchema of associated pattern
+  // Extract schema from cell, following links that carry an embedded schema
   const { schema } = cell.asSchemaFromLinks().getAsNormalizedFullLink();
 
   if (isNontrivialSchema(schema)) {
     return schema;
   }
 
-  // Try harder: resolve all links, clear the schema, then look up the schema
-  // from the source pattern's resultSchema. This handles cells accessed via
+  // Resolve all links, clear the schema, then look up the schema embedded
+  // in the links along the resolved chain. This handles cells accessed via
   // arrays (e.g., mentionables, recents) where the intermediate cell doesn't
-  // carry a schema but the underlying piece cell's pattern does.
-  // Uses the same trick as addFavorite in home.tsx: clearing the schema with
-  // asSchema(undefined) forces asSchemaFromLinks to look it up fresh from the
-  // source pattern.
+  // carry a schema but a link in the chain does.
   try {
     const resolvedSchema = cell.resolveAsCell().asSchema(undefined)
       .asSchemaFromLinks()?.getAsNormalizedFullLink()?.schema;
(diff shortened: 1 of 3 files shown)


Hixie:
cc @ubik2 since this was your TODO. Let me know if my agent's deep dive here yielded results or was actually completely off base!

ubik2:
> cc @ubik2 since this was your TODO. Let me know if my agent's deep dive here yielded results or was actually completely off base!

I should have cleaned up the TODO, but it looks like the core functionality of getting the schema from the sigil links is working.

As for the fallback to resultCell's meta schema, I think we probably don't need to provide that, but it is useful and nice to have for when you've lost or thrown away your own link based copy. It's also useful for debugging.

Were you running into a problem here, or was this just cleanup?

Overall, it looks good to me. The result cell fallback won't always be available, but when it is, this is nice to have.

There's a performance "regression" that I didn't investigate. I didn't see any obvious reason this would have been significantly slower, but it's probably worth re-running that benchmark to see if this is a real issue.

It's also valid to have a trivial schema (e.g. `true` or `{}`). These aren't useful to describe the object, since it will match anything. Since I think we're calling getCellSchema so we can pass this description to an llm, it seems reasonable to ignore those, and look for something that might be more descriptive, even if it's not the actual schema.

Edit: benchmark issue is probably addressed by https://github.com/commontoolsinc/labs/pull/4168
Edit2: nope. that was just cfcheck, not check.

ubik2 (approved):

Hixie:
This was primarily driven by me just telling the agent to resolve the TODO.

My intuition is that we have too many places we're looking up schemas. I'd feel more comfortable if there was a single concrete answer to "what is the schema" that didn't involve O(N) lookups or fallbacks. I'm chasing that rabbit hole separately though.

I don't feel strongly about the added fallback here, and it was not added because of a specific problem I was running into, so I'm happy to drop that too if you prefer.

Hixie:
@ubik2 I'm inclined to land this as is (and so will do so once it's green), but am happy to remove the extra fallback if you think it's not needed.

ubik2:
> @ubik2 I'm inclined to land this as is (and so will do so once it's green), but am happy to remove the extra fallback if you think it's not needed.

Fine to land it either way. There's some failing checks. I don't expect those to be related, but perhaps there's something wrong with main right now.

PR #4161, opened 2026-06-16
feat(profiles): working profiles end-to-end — card-add CFC fixes + followable piece cards (CT-1740, CT-1755)

bfollington:
## Profiles, working end to end

Delivers a working Common Fabric profile: create a profile, add cards, **pin a real deployed piece and click through to the live running piece**, and visit the rendered profile page — with owner-protection (CFC) intact. Four commits across two layers (runtime CFC fix → user-facing feature); kept together so the branch is one working thing rather than several partial PRs.

Builds on #4158 (moduleIdentity over authored source), which fixed the profile *read*/render path but not the *write*.

### 1. `54a2547` — warm-load `fn.src` canonicalization (CT-1754)
The warm/cached record load kept no per-module source map, so a re-loaded handler's `fn.src` stayed the raw `<evalId>.js` bundle coordinate; the verified-binding check couldn't canonicalize it and downgraded the writer to `unsupported`, rejecting owner-protected writes by the non-exported `mutateElements` with *"writeAuthorizedBy requires a trusted verified binding identity"*. Fix: `identitySourceMap()` (`js-compiler/source-map.ts`) + a warm-path fallback at the two source-map registration sites in `engine.ts` — re-labels the bundle frame so the **existing** check passes (not weakened).

### 2. `8ead4e9` — stamp `writeAuthorizedBy` from the binding, not a foreign seeder (CT-1740)
With (1) fixed, card-add then hit *"writeAuthorizedBy **must remain stable**"*. Instrumentation captured the two diverging stamps on a freshly-created profile: the **seed** stamped `profile-create`'s moduleIdentity (`fbbISemF…`) while the card-add **write** stamped `profile-home`'s (`zwdMteV4…`). A claim's provenance was stamped with the *authoring writer's* module — correct only when the writer *is* the bound function. `profile-create`'s `submitProfileCreation` seeds an `inSpace` ProfileHome whose `elements` bind to `profile-home`'s `mutateElements`, so it stamped the wrong module. (Not a hash skew — #4158 made moduleIdentity bundle-invariant; this is wrong-module **attribution**.) Fix: in `rebindWriteAuthorizedByClaims`, only the function **named by the binding** (matching `sourceFile`+`bindingPath`) may stamp it; a foreign initializer leaves it unstamped and the genuine writer stamps it later (reconciles). **Fail-closed** — unstamped claims still fail verification; reconcile requires matching bindings. Adversarial gate + `profile-owner-cfc` security cases stay green.

### 3. `60e0007` — followable piece cards + self-badge nav fix (CT-1755)
Cards were title-only placeholders (`ProfileCatalogCard`/`UrlPatternReference` render a title; URLs were plain text) — nothing pointed at a real piece. Now `mutateElements` has an `addPiece` mode that stores a real cross-space link (`{ "/": { "link@1": { id, space, path } } }`, the same shape `inSpace`/`profiles[]` use) to an **existing deployed piece**; the card renders as a `<cf-cell-link>` showing the live piece name and opens the running piece on click. New edit-mode **"Pin a piece"** form (space DID + piece id, or paste a `/<did>/<id>` URL); an exported `addPiece` stream also takes `{ pieceSpace, pieceId, title }` for a future "pin from the piece" flow. Also: `cf-profile-badge` gains a `noNavigate` prop (+ click guard) — the self-badge is bound to a `computed()` projection (not a real piece), so a click routed to an invalid URL; rosters bind a real profile link and still navigate.

### 4. `da777d9` — tests
`profile-home-pin-piece.test.ts` (the `addPiece` element resolves to the pinned piece's space + id) and a `noNavigate` case in `cf-profile-badge.test.ts`.

## Validation
- **Browser, end to end:** create a profile → render with owner-protection → add a card via the UI → **deploy a counter, pin it by space DID + id, click the card → the live interactive counter** → self-badge no longer navigates.
- **In-process / CI:** gold-standard real-pattern card-add test (red→green), `addPiece` link test, badge `noNavigate` test; full CFC + identity suite (incl. the 42-step adversarial gate and `profile-owner-cfc`) green; `deno check`/`lint` clean.

## Notes
- **Migration boundary:** profiles seeded *before* #4158 carry the legacy injected-form stamp and still conflict on card-add until recreated (accepted, pre-launch). New profiles work.
- **Pinning cross-space** needs the `did:key:…` form (a pattern can't resolve a space *name* → DID; only the shell can). The future "pin from the piece" flow sidesteps this. Inline embed (vs link-out) is a deliberate follow-up.
- The **Performance Check** failure is a documented false positive: the CT-1754 *compiler* change invalidates the pattern compile cache, so this branch runs `cfcheck` cold (~3.5m) vs main's warm ~1m21s; same-machine cold timing is identical with/without the change (main 1m53s ≈ branch 1m57s). Not rebaselined — it self-heals once the cache warms post-merge.

Part of the profiles epic **CT-1645**.

### DIFF
--- a/packages/html/src/jsx.d.ts
+++ b/packages/html/src/jsx.d.ts
@@ -4255,6 +4255,14 @@ interface CFProfileBadgeAttributes<T> extends CFHTMLAttributes<T> {
   /** A cell containing a profile (ProfileHomeOutput): renders avatar + name. */
   "$profile"?: CellLike<any>;
   "size"?: "xs" | "sm" | "md" | "lg" | "xl" | CellLike<string>;
+  /**
+   * Suppress click-to-navigate. Set when the badge is bound to a derived view
+   * of a profile (e.g. the self-badge on a profile-home page) rather than the
+   * profile's own root piece, so a click would route to an invalid URL.
+   * (Camel-cased to match the element property — the renderer assigns JSX props
+   * as properties, so a lower-cased name would miss the reactive property.)
+   */
+  "noNavigate"?: boolean;
 }
 
 interface CFChipAttributes<T> extends CFHTMLAttributes<T> {
--- a/packages/js-compiler/mod.ts
+++ b/packages/js-compiler/mod.ts
@@ -30,6 +30,7 @@ export {
 } from "./program.ts";
 export {
   composeBundleSourceMap,
+  identitySourceMap,
   isSourceMap,
   type MappedPosition,
   parseSourceMap,
--- a/packages/js-compiler/source-map.ts
+++ b/packages/js-compiler/source-map.ts
@@ -89,6 +89,36 @@ export function composeBundleSourceMap(
 }
 
 /**
+ * Build an IDENTITY source map for a compiled body whose authored source map
+ * was not retained — every generated line maps to the same line/column of
+ * `source`. Used by the warm/cached module-record load path, where the
+ * content-addressed cache stores compiled bodies but not their per-module
+ * source maps: without a registered frame the ESM loader resolves `fn.src` to
+ * the raw bundle coordinate (`<evalId>.js:..`), which the harness cannot
+ * canonicalize, and CFC verified-source identity fails closed. An identity map
+ * makes `mapPosition(<name>, line, col)` resolve to `<name>:line:col`, which
+ * the engine then rewrites to the canonical `cf:module/<id>/<path>` form via
+ * its per-module name → canonical table (parity with the source-compile path,
+ * which carries a real authored map).
+ *
+ * Line/column coordinates are preserved verbatim (the compiled body IS the
+ * eval'd text under this load), so no positional information is invented — the
+ * map only re-labels the bundle coordinate with the per-module source name.
+ */
+export function identitySourceMap(body: string, source: string): SourceMap {
+  const generator = new SourceMapGenerator({ file: source });
+  const lineCount = (body.match(/\n/g)?.length ?? 0) + 1;
+  for (let line = 1; line <= lineCount; line++) {
+    generator.addMapping({
+      generated: { line, column: 0 },
+      original: { line, column: 0 },
+      source,
+    });
+  }
+  return JSON.parse(generator.toString()) as SourceMap;
+}
+
+/**
  * Maximum number of source maps to cache in memory.
  * When exceeded, oldest (least recently used) entries are evicted.
  *
--- a/packages/patterns/system/profile-home.tsx
+++ b/packages/patterns/system/profile-home.tsx
@@ -43,7 +43,7 @@ export type ProfileElement = {
   tag: string;
   userTags: readonly string[];
   title?: string;
-  source?: "catalog" | "url";
+  source?: "catalog" | "url" | "piece";
 };
 
 export type AddProfileElementEvent = {
@@ -72,6 +72,12 @@ export type MutateProfileElementsEvent = {
   tag?: string;
   userTags?: readonly string[];
   cell?: any;
+  // "addPiece" inputs: a link to an EXISTING deployed piece (CT-1755). The
+  // card becomes a followable reference to the live piece rather than a local
+  // title-only placeholder. `pieceSpace` is the target space DID; `pieceId` is
+  // the piece id (with or without the `of:` prefix).
+  pieceSpace?: string;
+  pieceId?: string;
 };
 
 export type SetProfileNameEvent = {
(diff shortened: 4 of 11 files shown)


bfollington:
## Browser end-to-end verified ✅

Full loop in a real browser on this branch (fresh identity, post-restart server with the fix):

1. **Login** (Import CLI Key) → fresh home.
2. **Create profile** "AdaE2E" → lands in its own `inSpace` space (`z6MkqYToLa…`, distinct from the home DID).
3. **Render** the profile via the picker's `cf-cell-link` → presentation view shows the name (owner-protected, shield) — cross-space read works.
4. **Add card through the UI** ("Add profile card" in edit mode) → commits with **no CFC rejection**; `elements` grows and the card renders as a chip. (Also exercised the exported `addElement` stream via CLI against the live server: `commit.error: undefined`, `elements 0→1→2`.)
5. Both cards (`E2E Card`, `Profile card`) render in the presentation view; direct piece-id URL (`/{space}/fid1:…`) renders the profile standalone.

Pre-fix this exact write was silently rejected (`writeAuthorizedBy must remain stable`, 6 retries dropped). With the binding-scoped stamp fix it commits and renders.

Note: an earlier "No data at cell" in testing was a navigation-id artifact (using the result-holder id `of:fid1:…` in the URL instead of the piece id `fid1:…`) — **not** a read regression. The render path (merged #4089/#4150) already works; the only fix needed was the write.

bfollington:
### Re: the failing Performance Check — false positive (cold compile-cache)

The only OVER-threshold metric is the **`Check` job wall time** (4m54s vs 1m21s baseline). The slow step is **`cfcheck`** ("Check Common Fabric patterns"), not `deno check` (type-check was a normal ~59s).

`cfcheck` compiles all ~361 pattern files. This PR's first commit changes the **compiler** (`js-compiler/source-map.ts`, `harness/engine.ts` — the CT-1754 identity source map), which invalidates the pattern **compile cache**, so this branch's CI runs `cfcheck` **cold** (full recompile, ~3.5m on the runner) while the main baseline runs it **warm** (~1m21s whole job). It returns to baseline once the cache warms after merge.

Same-machine cold timing confirms my code isn't the cause:
- `origin/main` cold `cfcheck`: **1m53s**
- this branch cold `cfcheck`: **1m57s** (≈ identical; the small delta is local untracked scratch patterns not present in CI)

So this is a wall-time false positive from cache invalidation, not a runtime regression. I'm intentionally **not** adding a `NEW_PERF_BASELINE` override for `Check` — 294s would be a wrong permanent baseline (it'll be ~1m21s again once the cache warms post-merge, and a loose baseline would mask real regressions). Safe to merge through this check.

bfollington:
---

## Update: followable piece cards + self-badge nav fix (CT-1755) — commit `60e0007`

The CFC fix above made owner-protected element writes work; this builds the user-facing payoff on top, so the branch now delivers a **working** profile end to end (per the "one working thing, not granular PRs" ask).

**Problem.** Profile "cards" were local title-only placeholders (`ProfileCatalogCard` / `UrlPatternReference` render just a title; a URL was shown as plain text). Clicking one opened a piece with the right title but no content — nothing pointed at a real deployed piece. Separately, the self-badge on a profile page navigated to an invalid URL.

**Followable cards.** `mutateElements` gains an `addPiece` mode that stores a real cross-space link to an **existing deployed piece** as the element's `cell` (`{ "/": { "link@1": { id, space, path } } }` — the same serialized shape `inSpace`/`profiles[]` links use). The card renders as a `<cf-cell-link>` showing the **live piece name**; clicking it opens the running piece. New edit-mode **"Pin a piece"** form (space DID + piece id, or paste a `/<did>/<id>` URL); an exported `addPiece` stream also accepts `{ pieceSpace, pieceId, title }` for a future "pin from the piece" flow.

**Self-badge.** `cf-profile-badge` gains a `noNavigate` property (+ click-time guard); profile-home sets it on the self-badge (bound to a `computed()` projection, not a real piece — hence the bad URL). In rosters the badge is still bound to a real profile link and navigates correctly. Added the `noNavigate` JSX prop to `CFProfileBadgeAttributes`.

**Browser-verified:** deployed a counter, pinned it by space DID + id, the card rendered "Demo Counter", clicking it opened the **live, interactive** counter, and the self-badge no longer navigates.

Note: a space *name* can't be resolved to a DID from pattern code, so cross-space pins use the `did:key:…` form (the future "pin from the piece" flow sidesteps this since the shell already holds the live cell).

bfollington:
Perf-check update: the earlier cold-cfcheck false positive is resolved by rebasing onto main (cfcheck is now its own sharded job, off the Check job). The current Performance Check failure is an **unrelated flaky subtest** — `packages/patterns/cozy-poll/multi-user.test.tsx` at 20.9s vs 15.8s baseline (+32.6% on a ~20s test), which this PR does not touch (not in the diff). Re-running to clear the timing; not rebaselining an unrelated test.

PR #4174, opened 2026-06-16
fix(integration): add viewSettled readiness primitive and fix flaky d…

Hixie:
…efault-app test

The notebook step of default-app.test.ts ("should persist and reload every rapidly created notebook note") was flaky on CI: it timed out after 60s waiting for the New Note modal's "Create Another" button to appear.

Root cause is a readiness gap. The reactive scheduler runs in a worker while the DOM lives on the main thread, so after a state change there are three stages before a control is interactive:

  1. the worker settles reactively (this is all rt.idle() observes);
  2. the resulting vdom batch crosses to the main thread and is applied, which is where the click listener is attached;
  3. the Lit elements finish their update cycle -- cf-modal only binds handlers and drops pointer-events:none in its updated() callback, one cycle after its open property is set.

Only stage 1 is visible through the runtime idle signal. The test clicked the New Note button once, before its onClick was wired up, so the click was dropped, the modal never opened, and the wait timed out. Re-clicking until it "took" made matters worse: the modal is dismissable, so repeated clicks behind it dismissed and reopened it in a flicker that corrupted the rest of the flow.

Add a main-thread readiness primitive that closes all three stages:

  - packages/html/src/debug.ts: viewSettled() loops idle -> yield a macrotask so in-flight vdom batches are applied -> drainViewUpdates() awaits Lit updateComplete across every active render subtree (including shadow roots) -> repeat to a fixpoint. drainViewUpdates uses Promise.allSettled because Lit rejects updateComplete when an element's update is superseded mid-churn; a rejection means the element kept changing, which the re-drain captures, so it must not abort the check. On non-convergence (a perpetually re-rendering element) viewSettled warns and proceeds rather than failing an interactive view.
  - packages/shell/src/views/RootView.ts: expose it as commonfabric.viewSettled alongside commonfabric.rt, and clear it on runtime teardown so it cannot run against a torn-down runtime. The running app never calls it; it is a test-support global, so the change is additive with no app behavior change.
  - packages/integration/utils.ts: awaitViewSettled(page) wraps the global for tests, returning false until the runtime exposes it so callers can poll with waitFor.

The test now settles the view before clicking New Note (so the single click lands on a bound handler), then waits for the modal's effect. The rapid-create loop is unchanged, preserving the concurrent-create stress it guards.

docs/development/UI_TESTING.md documents the primitive and enumerates the inferior waits it replaces (fixed sleeps, rt.idle()-then-interact, poll-for- element-then-click, re-click-until-it-takes, waitForSelector retry loops).

Verified: default-app passes across fresh and warm/accumulating-state servers (the condition that reproduced the flake); html (32) and shell (12) unit suites green; shared-profile integration green.

---
## Summary by cubic
Adds a UI readiness primitive so tests only interact once the view is interactive, fixing the flaky default‑app notebook flow. Exposes it as `commonfabric.viewSettled` and adds `awaitViewSettled(page)` in `@commonfabric/integration` for polling.

- **New Features**
  - `@commonfabric/html/debug`: add `viewSettled()` and `drainViewUpdates()` (idle → macrotask yield → await Lit `updateComplete` across DOM and shadow roots; warn after max passes). Unit tests added.
  - Shell: expose as `commonfabric.viewSettled` via `createViewSettled()`; installed by `exposeCommonfabricGlobals()` and cleared by `clearRuntimeDebugGlobals()`; reads the runtime on each call and resolves even with no runtime. Unit tests cover the creator and console‑globals wiring.
  - `@commonfabric/integration`: add `awaitViewSettled(page)` that returns `false` until available for `waitFor` polling. Docs add “Waiting Until the UI Is Interactive” with usage and anti‑patterns to avoid.

- **Bug Fixes**
  - Default‑app notebook test settles the view before clicking “New Note,” then settles again and waits for the modal’s effect, removing dropped clicks and modal flicker.

<sup>Written for commit 0e7b0f55884238bab23069f1b849006cdb77f776. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4174?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---BEGIN COPY-PASTE---
NEW_PERF_BASELINE: subtest: pattern-unit-5/pattern-unit-tests > packages/patterns/shopping-list.test.tsx = 28s
---END COPY-PASTE---

### DIFF
--- a/docs/development/UI_TESTING.md
+++ b/docs/development/UI_TESTING.md
@@ -184,6 +184,68 @@ describe("shadow DOM component test", () => {
 });
 ```
 
+## Waiting Until the UI Is Interactive
+
+Finding an element is one problem; knowing *when* it is ready to be clicked or
+typed into is a separate one. The reactive scheduler runs in a worker while the
+DOM lives on the main thread, so after a state change there are three stages
+before a control is interactive:
+
+1. the worker settles reactively — `rt.idle()` covers this;
+2. the resulting vdom batch crosses to the main thread and is applied — this is
+   where the click listener is attached;
+3. the Lit elements finish their update cycle — `cf-modal`, for example, only
+   binds handlers and removes `pointer-events: none` in its `updated()`
+   callback, one cycle after its `open` property is set.
+
+Only stage 1 is visible through the runtime idle signal, so an element can exist
+in the DOM, be found by a selector, and still drop a click because its handler
+is not bound yet.
+
+**Use `awaitViewSettled(page)`** from `@commonfabric/integration` before issuing
+the first click or keystroke after navigation or any state change. It resolves
+once all three stages are done, so a single click then lands on a bound handler:
+
+```typescript
+import { awaitViewSettled, waitFor } from "@commonfabric/integration";
+
+// Before interacting: wait until the view is mounted and interactive.
+await waitFor(() => awaitViewSettled(page));
+const button = await page.waitForSelector("cf-button[role='button']");
+await button.click(); // delivered to a bound handler
+
+// After interacting: settle, then wait for the click's specific effect.
+await waitFor(async () => {
+  await awaitViewSettled(page);
+  return (await page.$("cf-modal[open]", { strategy: "pierce" })) !== null;
+});
+```
+
+### Do not reach for these instead
+
+Each of the following is what `awaitViewSettled` replaces. They are either racy
+(they do not cover all three stages) or wasteful (a CDP round trip every poll):
+
+- **`await sleep(ms)`, `page.waitForTimeout(ms)`, or any fixed delay** — guesses
+  at timing; too short flakes, too long wastes wall-clock.
+- **`await rt.idle()` or `awaitRuntimeIdle(page)` alone, then interact** — covers
+  stage 1 only. The element may be in the DOM with no handler bound, or be a
+  `cf-modal` whose `pointer-events` are still off.
+- **`await waitFor(() => findButtonWithText(page, "X"))` then click** — presence
+  is not interactivity, and every poll is a CDP `evaluate` round trip (the
+  default 50 ms interval also quantizes timing measurements; see
+  `packages/integration/utils.ts`).
+- **`await waitFor(() => clickButtonWithText(page, "X"))`, re-clicking until it
+  "takes"** — fires the stimulus speculatively. It can double-fire (creating
+  duplicate state) and, against a `dismissable` overlay such as `cf-modal`,
+  repeated clicks behind the open modal dismiss it and leave the test stuck.
+- **`page.waitForSelector(...)` in a retry loop to "wait it out"** — Astral's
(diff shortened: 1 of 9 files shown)


mathpirate:
Heads up: I've got a draft stacked on this — #4184 reworks the parking test's `clickCfButtonAndWaitForText` onto your `awaitViewSettled` (settle → click once → settle-wait) to fix CT-1738 flake #4. It's branched off this PR and includes your commit for now; I'll rebase it onto `main` once this merges, so nothing needed on your end. Thanks for the primitive — it was exactly the right shape.

-Opus 4.8 with Gideon's approval

mathpirate:
Hey @Hixie — viewSettled is exactly the primitive this area needed; I've got a stacked PR ([#4184](https://github.com/commontoolsinc/labs/pull/4184)) building the parking-test browser helpers on awaitViewSettled. cubic reviewed my PR and flagged one thing in drainViewUpdates — it's only in my diff because of the stack, so passing it along here since it's your code.

The observation: drainViewUpdates derives its return value purely from the synchronous scan (el.isUpdatePending at visit time), then await Promise.allSettled(pending) and discards every updateComplete outcome. Two cases slip through:

1. Supersession → early return. Lit resolves updateComplete to false when an update is superseded mid-cycle. If an element isn't isUpdatePending at scan time but its in-flight updateComplete resolves false during the await, the function still returns false and viewSettled exits a cycle early. The per-pass idle() isn't a backstop here — viewSettled returns the moment one pass reports settled, so there's no next pass to re-drain.

2. Thrown updates → swallowed. allSettled eats a rejected updateComplete, which in Lit signals an exception thrown during the element's update — a genuine render error, now invisible to the readiness check.
One nuance on the comment just above the allSettled: it says Lit "rejects updateComplete when an update throws or is superseded." Per Lit's documented contract, supersession resolves to false — only a thrown error rejects. So the "captured by the sawPending re-drain" rationale doesn't actually cover the supersession path (it resolves rather than rejects, and there's no re-drain after an early return). Worth confirming against the exact Lit version before acting on it.

If you want to tighten it, folding both outcomes into the pending signal would do it:

```
const results = await Promise.allSettled(pending);
const churned = results.some((r) => r.status === "rejected" || r.value === false);
return sawPending || churned;
```

Genuinely low-stakes for my use — my helper's outer waitFor re-settles and re-checks, so the early-return window gets papered over downstream — so no urgency, and entirely your call whether it's worth a change. Just didn't want to sit on it since it's in your primitive. Credit to cubic for spotting it.

Hixie:
Updated to take that comment into account, thanks.

PR #4178, opened 2026-06-16
Evidence: shared-state writes silently dropped under concurrency

mathpirate:
**For discussion, not for merge** — a diagnostic harness + a grounded root cause, not a fix.

## TL;DR
Under concurrent multi-user writes to the same document/cell, writes are **silently lost in canonical storage** — no error to the app. It is **general, not SQLite-specific**, **keying does not avoid it** (the conflict unit is the whole document, not the field), and the failure can be **invisible and displaced**: a dropped *identity/setup* write silently voids every later action that depended on it.

Full code-grounded writeup: [`packages/patterns/lunch-poll/DROPPED-WRITES-EVIDENCE.md`](https://github.com/commontoolsinc/labs/blob/perf/shared-state-write-drops/packages/patterns/lunch-poll/DROPPED-WRITES-EVIDENCE.md).

## Evidence (cold-reader = canonical storage truth)
A fresh runtime that never subscribed reads committed storage directly, so its count is server truth (not a stale UI):

| storage | concurrency | landed (cold storage) |
| ------- | ----------- | --------------------- |
| plain keyed CELL (`main-indexed.tsx`) | 10 users | **18 / 30** |
| SQLite (`main-sqlite.tsx`) | 5 users | **1 / 15** |

## Root cause (grounded)
- **Conflict unit = the entity-DOCUMENT** `(branch, id, scope_key)`. The conflict predicate `SELECT_SET_DELETE_CONFLICT` (`memory/v2/engine.ts:529`) has **no field/path/space column** → distinct keys in one `Record` collide (keying doesn't help: `cell.ts:1492-1521`); disjoint cells don't (distinct ids; space scope_key is constant).
- Retry budget `DEFAULT_RETRIES_FOR_EVENTS = 5` (`scheduler/constants.ts:5`); the commit is **fire-and-forget** (`events.ts:609-616`) → silent to the caller, only a scheduler-level log on exhaustion (`events.ts:661`).

## The cascade (the dangerous part)
`joinAs` writes the contended `usersByName`/`userOrder` docs **and** `myName` (`PerUser`) in one tx (`main-indexed.tsx:156-177`). A dropped join takes `myName.set()` with it → empty `myName` → `castVote`'s `if (!me) return` guard makes that user's votes **silently no-op**. Measured: **9 `castVote`s fired with `me=""`** at 3×10. So the symptom ("my votes don't count") is nowhere near the fault (a dropped join, one phase earlier).

## What it is NOT
The lunch-poll `missing(12) > exhaustions(6)` *looked* like a second, silent mechanism — it's the cascade above. Two structural hypotheses (a shared-subrecord write-shape lost-update; the multi-bucket structure) were reproduced in a minimal SQLite-free repro and **refuted** (both `missing == exhaustions`). There is **one** mechanism.

## Status
- Root cause ✅ (grounded + measured + re-run).
- **No fix.** Fix surface in the doc §6: (1) surface exhaustion to the caller instead of fire-and-forget (the headline ask; covers the cascade); (2) finer per-path conflict detection so independent writes don't collide; (3) split hot docs / CRDT for commutative ops; (4) bigger retry budget + backoff (rate, not structure). Per team discussion, genuine same-value RMW has no perfect resolution — the priority is that the client is told.

## Reproduce
```bash
deno run -A packages/patterns/lunch-poll/probe-sqlite-landing.ts --program=main-indexed.tsx --case=3x10 --rounds=3
deno run -A packages/patterns/lunch-poll/probe-sqlite-landing.ts --program=main-sqlite.tsx  --case=3x5  --rounds=3
# minimal, SQLite-free; missing == 'exhausting all retries' count -> one mechanism:
deno run -A packages/patterns/write-contention/probe.ts --users=10 --rounds=5 --mode=map 2>/tmp/wc.err; grep -c "exhausting all retries" /tmp/wc.err
# conflict unit is the document, not the space (disjoint cells don't contend):
deno run -A packages/patterns/write-contention/probe-space.ts --groupA=10 --groupB=10 --rounds=5
```

### DIFF
--- a/packages/patterns/integration/multi-runtime-harness.ts
+++ b/packages/patterns/integration/multi-runtime-harness.ts
@@ -202,14 +202,24 @@ export class MultiRuntimeHarness {
   readonly sessions: MultiRuntimeSession[];
   readonly pieceId: string;
   #server?: StandaloneMemoryServer;
+  /** Resolved space + api url so late ("cold") sessions can join the same
+   *  storage after the original sessions were created. */
+  readonly spaceName: string;
+  readonly apiUrl: string;
+  /** Cold sessions added post-hoc via {@link addColdSession}; disposed too. */
+  readonly coldSessions: MultiRuntimeSession[] = [];
 
   private constructor(
     sessions: MultiRuntimeSession[],
     pieceId: string,
+    spaceName: string,
+    apiUrl: string,
     server?: StandaloneMemoryServer,
   ) {
     this.sessions = sessions;
     this.pieceId = pieceId;
+    this.spaceName = spaceName;
+    this.apiUrl = apiUrl;
     this.#server = server;
   }
 
@@ -272,7 +282,7 @@ export class MultiRuntimeHarness {
         await session.client().call("openPiece", { pieceId });
       }
 
-      return new MultiRuntimeHarness(sessions, pieceId, server);
+      return new MultiRuntimeHarness(sessions, pieceId, spaceName, apiUrl, server);
     } catch (error) {
       bootstrap?.terminate();
       for (const session of sessions) {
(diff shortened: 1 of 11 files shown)


mathpirate:
This was an investigation PR, closing - final result is https://github.com/commontoolsinc/labs/pull/4220

PR #4194, opened 2026-06-17
fix(profiles): derive profile space from per-user cause, not display name (CT-1650)

bfollington:
### The bug
`profile-create.tsx` seeded each profile via `ProfileHome.inSpace(name)`. The **named** `inSpace` path derives the space DID from `Identity.fromPassphrase("common user").derive(name)` (the `createSession` spaceName branch, `packages/identity/src/session.ts:24-34`) — the **display name alone**, ignoring the authenticated user. So:

- two different users who pick the same profile name, **or**
- one user who creates two same-named profiles

…collide into a single shared profile space.

Observed live and reproduced in test: two "Ada Lovelace" profiles under different identities both resolve to `did:key:z6Mki2GJXAviRRHwhBNHzBoeG6SPxAXbhM8EAPJ2znNZnBKP`.

### The fix
Route profile creation through the **anonymous** `ProfileHome.inSpace()`. Its DID derives from the creating handler's frame cause (`runner.ts` `createPatternFrame`: `{ ...inputs, $event }`), which carries the per-user home-space input links **plus** the durable per-event id — unique per user **and** per creation event, and stable across the cross-space-commit retry. The display name now flows **only** to `initialName` (editable, independent of the space identity).

`createSession`/`resolveSpaceName` are **unchanged globally**, so any intentional shared-by-name spaces keep their semantics. (Grep confirms no production `inSpace("name")` caller relies on shared-by-name today — only profile-create did.)

### Migration
`profiles[]` stores **concrete** cross-space DID links, so existing profiles keep resolving to their already-baked space; only **new** profiles get unique spaces. New-profiles-only migration, acceptable pre-launch. (Users who already collided pre-fix remain merged in the old shared space — pre-existing damage the fix can't un-merge.)

### Acceptance criteria
- [x] Profile space DID not derived solely from the display name
- [x] Two users with the same name → distinct spaces
- [x] Display name editable independently of the space identity (flows only to `initialName`)
- [x] Existing profile links keep working (concrete DID; documented new-profiles-only migration)
- [x] Integration test covering two users + same name

### Tests
`packages/runner/test/profile-space-identity-collision.test.ts` runs the **real** profile-create + profile-home under two different signers and asserts:
- two users, same name → distinct spaces (red baseline reproduced the exact live DID `z6Mki2GJ…`);
- one user's two same-named profiles → distinct;
- both are real `did:key:` spaces.

Existing `profile-create-real-card-add`, `profile-home-pin-piece`, `inspace-child-owner-{write,seed}`, `profile-owner-cfc`, `inspace-child-reload` all still pass.

Docs updated for coherence: the spec's **"Profile Space Identity"** section and `HOME_SPACE.md` described the now-fixed `inSpace(name)` derivation — both now point at the anonymous `inSpace()` with the rationale.

---
## Summary by cubic
Fixes CT-1650 by creating profile spaces with anonymous `inSpace()` instead of name-based `inSpace(name)`, removing same-name collisions across users and profiles. Existing profiles keep working; new profiles get unique DIDs independent of display name.

- **Bug Fixes**
  - Route profile creation through `ProfileHome.inSpace()` (no arg) so the space DID derives from the per-user frame cause, not the display name.
  - Display name now only sets `initialName`; space identity is fully decoupled from the name.
  - Added `profile-space-identity-collision.test.ts` and updated docs (`HOME_SPACE.md`, spec) to reflect the new behavior.

- **Migration**
  - No action needed; existing profiles keep their concrete DID links.
  - Only new profiles get unique spaces; intentional named-space behavior elsewhere remains unchanged.

<sup>Written for commit be83e64a364ac9d77329c7fb8d12b008d4a7fbd6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4194?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/common/conventions/HOME_SPACE.md
+++ b/docs/common/conventions/HOME_SPACE.md
@@ -75,8 +75,10 @@ most-recently-used (MRU) ordering:
 - `homeSpaceCell.defaultPattern.mru` — recency-ordered links; drives ordering
   after the default.
 
-Each profile lives in its own space, created with `PatternFactory.inSpace(name)`
-running `/api/patterns/system/profile-home.tsx`; the link is appended to
+Each profile lives in its own space, created with the anonymous
+`PatternFactory.inSpace()` (CT-1650 — a *named* `inSpace(name)` would derive the
+space DID from the display name alone and collide same-named profiles across
+users) running `/api/patterns/system/profile-home.tsx`; the link is appended to
 `profiles`. The home Profile tab renders the **profile picker**
 (`profile-picker.tsx`): it lists profiles, lets the user create more inline, pick
 the default, and stamp MRU. There is no `profileName` mirror field anymore.
--- a/docs/specs/shared-profile-space.md
+++ b/docs/specs/shared-profile-space.md
@@ -128,17 +128,24 @@ home-field convention is `homeSpaceCell.defaultPattern.profile`.
 
 ### Profile Space Identity
 
-Profile creation uses `PatternFactory.inSpace(...)`:
+Profile creation uses the **anonymous** `PatternFactory.inSpace()` (CT-1650):
 
 ```ts
-const profile = ProfileHome.inSpace(spaceName)({ initialName: name });
+const profile = ProfileHome.inSpace()({ initialName: name });
 ```
 
-When `spaceName` is a non-DID string, the runner resolves it during async
-post-run through the existing named-space derivation path. When the argument is
-omitted, the runner generates a fresh DID. The home default pattern's `profile`
-link is the durable source of truth after creation. Runtime-only `.inSpace`
-annotations are also rewritten to the resolved DID during post-run.
+The argument MUST be omitted. A *named* `inSpace(spaceName)` derives the space
+DID from `Identity.fromPassphrase("common user").derive(spaceName)` (the
+`createSession` spaceName path) — the name ALONE, ignoring the authenticated
+user — so two users picking the same profile name, or one user creating two
+same-named profiles, collide into a single shared space. The anonymous case
+instead derives a fresh DID from the creating handler's frame cause (per-user
+home-space input links + the durable per-event id), so the space is unique per
+user AND per creation event, stable across the cross-space-commit retry. The
+display name is therefore independent of the space identity (it flows only to
+`initialName` and stays editable). The home default pattern's `profile` link is
+the durable source of truth after creation, and runtime-only `.inSpace`
+annotations are rewritten to the resolved DID during post-run.
 
 ### Profile Default Pattern Output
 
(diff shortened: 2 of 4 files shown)


Hixie:
My Claude believes this may have caused a flake regression. Just leaving this here as an FYI, I'll fix-forward.

----
### Regression: second profile created from the picker renders with the first profile's name

This PR (switching profile creation from `ProfileHome.inSpace(name)` to anonymous
`ProfileHome.inSpace()`) introduced a regression in the home profile picker. Creating a
**second** profile in a session renders it with the **first** profile's name, so its real
name never appears.

**Failing test:** `packages/patterns/integration/home-profile.test.ts` →
*"home-space profile creation › creates a second profile from the picker (multi-profile append)"*.
It creates "Ada Lovelace" then "Alan Turing"; the picker (`cf-cell-link`) renders the second
profile as another "Ada Lovelace" — a **distinct** space hash (so a separate space) but the
**wrong/stale name** — and the test times out waiting for "Alan Turing".

**Attribution** (bisected with the toolshed cache wiped every run — see note below):

| Commit (clean cache)                  | Result    |
| ------------------------------------- | --------- |
| `af4b17e64` (this PR's parent)        | PASS 4/4  |
| `f04d927aa` (this PR)                 | FAIL 3/3  |
| `main` HEAD                           | FAIL      |
| HEAD with `inSpace()` → `inSpace(name)` reverted | PASS 3/3 |

**Mechanism:** with anonymous `inSpace()`, the freshly-created second profile's name
(`initialName` → `.for("name")` in `profile-home.tsx`) doesn't propagate to the picker's
`cf-cell-link` label from a cold cache, so it falls back to a stale/sibling value. The
name-derived `inSpace(name)` path didn't hit this.

**Constraint for the fix:** reverting to `inSpace(name)` reintroduces CT-1650 (same-name
collision). The fix needs to keep anonymous per-event spaces **and** make the new profile's
name resolve in the picker — likely in the runner/builder cross-space child commit path
(`optIntoInSpaceMultiSpaceCommit` → `enableCrossSpaceChildCommit`, `createPatternFrame` cause)
for how an anonymous child's `initialName` / `.for("name")` cell is materialized and read
cross-space.

**Note — why the bisect needs care:** the toolshed persists per-space sqlite under
`packages/toolshed/cache/memory/engine-v3/` across server restarts, so a warm cache makes
broken commits look "good" (a naive bisect first mis-pointed at #4198). Repro:
`bash scripts/stop-local-dev.sh --port-offset=N && rm -rf packages/toolshed/cache/memory`,
then `HEADLESS=1 deno task integration --port-offset=N patterns home-profile`, run 3–4× from
a clean cache.

----

PR #4199, opened 2026-06-17
fix(runner): don't record a write's own machinery reads as conflict dependencies

mathpirate:
## Why (the correctness case)
A cell write records reads to build its patch: it resolves the write target's link and reads the current value of the slot it is about to write (to diff). Those were recorded as the writer's **conflict dependencies** — so a writer was treated as *depending on the very thing it blindly overwrites*. That false dependency makes two writes to **different keys of one document collide** (writer A's incidental container read overlaps writer B's add-patch parent path), even though `a` and `b` never logically interacted.

## What
- `Cell.set` runs its body under `runWithAmbientReadMeta(excludeReadFromConflict)` (a new read-meta marker), so its machinery reads (link resolution of the target + the diff read) carry the marker.
- `buildReads` drops marked reads from the **commit-time conflict set**; they remain in the journal for reactivity/CFC. A blind write becomes a **pure producer**.

## Semantic to be aware of
Two **blind** writes to the **same** key become last-writer-wins instead of conflict-and-retry — the correct semantics of a blind write (neither read the prior value). **Genuine** reads (a `.get()` evaluated before the set) still take a dependency, so read-modify-write still conflicts. (Direction approved by seefeld/wilk.)

## Verified
- `test/cell-write-conflict-granularity.test.ts` (real `MemoryV2Server`): disjoint blind adds **merge**; same-key RMW and tier-1 whole-doc delete **still conflict**.
- Targeted suites green with the change active: patterns-handlers / patterns-lift, scheduler-observations / static-writes, runner.

## Status
Active (no longer flag-gated). Remaining before merge: full CI green (running) + review.

## Perf note (honest)
Fixes the **direct-write** over-conflict. Does **not** move the lunch-poll / write-contention probe — that is dominated by pervasive coarse reads of the shared pattern doc by the reactive machinery + subscriptions (#4190), a separate workstream.

### DIFF
--- a/packages/runner/src/cell.ts
+++ b/packages/runner/src/cell.ts
@@ -70,6 +70,7 @@ import {
 } from "./query-result-proxy.ts";
 import { diffAndUpdate } from "./data-updating.ts";
 import { type LastNode, resolveLink } from "./link-resolution.ts";
+import { excludeReadFromConflict } from "./storage/reactivity-log.ts";
 import {
   type Action,
   ignoreReadForScheduling,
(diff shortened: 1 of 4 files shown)


mathpirate:
Superseded by https://github.com/commontoolsinc/labs/pull/4220

PR #4200, opened 2026-06-17
fix(runner,memory): honor nonRecursive shape reads in conflict detection

mathpirate:
## Why (the correctness case)
Reads carry a `nonRecursive` flag distinguishing a **shape** observation ("I looked at this container's key-set / existence") from a **deep** dependency ("I read values under it"). The **scheduler already honors this** — a shape read subscribes shallowly (re-run on key add/remove, not on a disjoint deep-value change). But **conflict detection threw the flag away**: `buildReads` stripped `nonRecursive` before the engine, so the engine treated every shape read as a deep read and over-conflicted it against any disjoint deep-value write under the container. The two halves of the system disagreed about the same read; this makes them consistent.

## What
- QueryResultProxy container/shape reads (creation, `ownKeys`, `getOwnPropertyDescriptor`, `has`) are recorded `nonRecursive`; deep-value dereferences stay recursive.
- `buildReads` keeps the flag through to the engine (`ConfirmedRead` gains the field) instead of stripping it.
- For a `nonRecursive` read the engine uses a shallow overlap — `isPrefixPath(touched, readPath)` — a **strict subset** of the symmetric `pathsOverlap`, so it can only **remove** spurious conflicts, never miss a genuine one. Tier-1 (set/delete) stays path-blind.

## Safety (no false-negative, by construction)
The shallow predicate is a strict subset of `pathsOverlap`. The only conflicts it removes are deep-value writes *strictly below* a shape read — which a shape read provably did not depend on. A value the handler actually dereferences is a recursive read and still conflicts.

## Verified
- `test/cell-write-conflict-granularity.test.ts` (real `MemoryV2Server`): leaf-read + disjoint deep-value write **merge**; key-set read vs key-add, same-key RMW, and tier-1 delete **still conflict**.
- Suites green with the change active: patterns-handlers / patterns-lift, scheduler-observations, runner; memory commit-preconditions.

## Status
Active (no longer flag-gated; the diagnostic 3092 toggle removed). Direction approved (berni). Remaining before merge: full CI green (running) + review.

## Perf note (honest)
A precision fix for the read-then-write disjoint case; does not by itself move the lunch-poll probe (machinery coarse-reads + subscriptions dominate, see #4199). This is a narrow, safe precision improvement — *not* the OT/CRDT path for genuine read-modify-write.

### DIFF
--- a/packages/memory/v2.ts
+++ b/packages/memory/v2.ts
@@ -111,6 +111,16 @@ export interface ConfirmedRead {
   branch?: BranchName;
   path: ReadPath;
   seq: number;
+  /**
+   * When true, this is a SHALLOW (shape-only)
+   * read — the reader observed the container at `path` (its key set / existence)
+   * but did NOT depend on the deep values of its descendants. The engine then
+   * conflicts only with writes AT-OR-ABOVE `path` (including key add/remove,
+   * whose patch injects the parent path), not with disjoint deep-value writes
+   * strictly below `path`. Strict subset of the recursive overlap → never a
+   * false-negative. Absent/false ⇒ recursive read (current behavior).
+   */
+  nonRecursive?: boolean;
 }
 
 export interface PendingRead {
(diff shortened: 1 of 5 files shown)


mathpirate:
Superseded by https://github.com/commontoolsinc/labs/pull/4220

PR #4205, opened 2026-06-17
fix(lunch-poll): de-flake post-delete history tests (count surface) + single-source historyCount

danfuzz:
## What

De-flakes `packages/patterns/lunch-poll/main.test.tsx`, which intermittently reds the "Pattern Unit Tests" job on `main` (it blocks unrelated PRs, e.g. docs-only #4191). **It's a flaky test, not a deterministic break** — the same suite passed on `main` commit `95ece6928` while failing on the three before it.

## Root cause (evidence-backed)

I split the two failing assertions into atomic sub-conditions and pushed a diagnostic run. CI reported precisely which value was wrong after a delete/clear:

| after | sub-condition | result |
|---|---|---|
| remove | `recentVisits.result.length === 1` | ✗ fail |
| remove | `rows[0].title === "Chipotle"` | ✗ fail |
| remove | `historyCount === 1` | ✓ pass |
| clear | `recentVisits.result.length === 0` | ✗ fail |
| clear | `historyCount === 0` | ✓ pass |
| clear | `voteHistoryCount === 0` | ✓ pass |

So the **count queries are fresh; the `recentVisits` LIMIT-8 *row* query is intermittently stale after a DELETE.** `db.query` re-execution is an async post-commit effect (`enqueuePostCommitEffect` → `flush()` → `provider.sqliteQuery`). The row query's flush is heavier than the scalar `count(*)` (cf-link column decode + CFC label-schema), so on a loaded runner it can land just *after* the harness settle loop (`runtime.idle()` + `storageManager.synced()`) declares quiescence — leaving `recentVisits.result` stale for one settle. Inserts don't trip it because later assertions give the row query time to catch up.

`#4193` didn't change the query or the delete handlers; it added a heavier UI, which raised the flake rate from ~0 to ~75%. The underlying race predates it. The *proper* fix is to make the runner await pending sqlite-query effects before settling — tracked separately; this PR is the local de-flake.

## Changes

1. **`main.test.tsx`** — key the two post-delete/clear history assertions on the reliable **count** surface (`historyCount`, `voteHistoryCount`) instead of the racy `recentVisits` row query. All three count conditions were green in the diagnostic CI run. Row-content ("which row survived") stays covered by the pre-delete assertions; a comment marks where to re-add a row check once the runner waits for pending sqlite effects.
2. **`main.tsx`** — `historyCount` reverts to the `count(*)` aggregate alone (drops `Math.max(count, recentRows.length)` from #4193). This makes `historyCount` a *trustworthy* assertion surface — the `Math.max` form read the same flaky row query, so it would have made `historyCount` flaky too. `hasHistory` is redefined as `recentRows.length > 0`, preserving #4193's intent (gate the history cards on the rendering source of truth so a transient count-of-0 can't flash them hidden) without feeding the row count back into `historyCount`. All of #4193's UI-rendering work is untouched.

## Testing

- `cf test packages/patterns/lunch-poll/main.test.tsx` → 19 passed, 0 failed (incl. compiled `cf` binary).
- `deno check` + `deno fmt --check` clean.
- Note: the flake does not reproduce on a fast/unloaded dev machine, so CI is the real signal — worth confirming green across a couple of runs.

## Performance Baseline

NEW_PERF_BASELINE: subtest: pattern-unit-2/pattern-unit-tests > packages/patterns/cozy-poll/main.test.tsx = 16s
NEW_PERF_BASELINE: coverage-debt: packages/patterns uncovered lines = 297132 lines

### DIFF
--- a/packages/patterns/lunch-poll/main.test.tsx
+++ b/packages/patterns/lunch-poll/main.test.tsx
@@ -288,17 +288,22 @@ export default pattern(() => {
   });
 
   // After deleting rows[0] (Thai, the most recent), only Chipotle remains.
-  const assert_one_history_after_remove = computed(() => {
-    const rows = poll.recentVisits.result ?? [];
-    return rows.length === 1 &&
-      rows[0]?.title === "Chipotle" &&
-      poll.historyCount === 1;
-  });
+  // DIAGNOSTIC: split into atomic conditions so CI reports which value lags.
+  const dbg_remove_rows_len1 = computed(() =>
+    (poll.recentVisits.result ?? []).length === 1
+  );
+  const dbg_remove_row0_chipotle = computed(() =>
+    (poll.recentVisits.result ?? [])[0]?.title === "Chipotle"
+  );
+  const dbg_remove_historyCount1 = computed(() => poll.historyCount === 1);
 
   // Clearing visits also clears the vote_history snapshots.
-  const assert_history_cleared = computed(() =>
-    (poll.recentVisits.result ?? []).length === 0 &&
-    poll.historyCount === 0 &&
+  // DIAGNOSTIC: split into atomic conditions.
+  const dbg_clear_rows_len0 = computed(() =>
+    (poll.recentVisits.result ?? []).length === 0
+  );
+  const dbg_clear_historyCount0 = computed(() => poll.historyCount === 0);
+  const dbg_clear_voteHistoryCount0 = computed(() =>
     poll.voteHistoryCount === 0
   );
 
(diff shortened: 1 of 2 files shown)


mpsalisbury:
Thanks Dan! I wonder how this passed CI on #4193?

mpsalisbury (approved):

danfuzz:
> Thanks Dan! I wonder how this passed CI on https://github.com/commontoolsinc/labs/pull/4193?

Turns out it was flaky not consistent, with a pass rate estimated to be ~25%.

PR #4210, opened 2026-06-17
fix(runner): reactive computes don't immediately re-queue on commit conflict

mathpirate:
> **Stacked on #4220 — do not merge before it.** See "Depends on #4220" below.

## Why

Reactive computes are not transactional retriers. When a compute's commit fails
with a **`ConflictError`** (stale read), an input has moved — and the write that
moved it has already **dirtied this compute's still-subscribed reads**, so normal
reader-dirty propagation re-runs it with the latest state. A conflict is a *wait
to be re-triggered*, not a retry.

## What

Pull commit conflicts **out of the reactive retry machinery entirely**:

- On a `ConflictError`: `restoreCfcTriggerReads()` + `resubscribe()`, then return.
  **No re-queue, no retry counter, no budget, no exhaustion-zombie.** Previously a
  conflict counted toward `MAX_RETRIES_FOR_REACTIVE` and, after 10, fell through
  to the exhausted-retries branch — stranding the compute as a zombie against
  rolled-back data, even though each conflict was just a wait.
- **Only `ConflictError` is removed from retry.** Non-conflict, non-permanent
  errors keep their bounded retry (`markDirectDirty` + `pending` +
  `queueExecution`); permanent (`PreconditionFailedError`) rejections stay
  non-retryable. The resubscribe runs on every error, so subscriptions stay fresh.

Two commits: (1) skip the immediate re-queue on conflict; (2) pull conflicts out
of the budget entirely.

## Depends on #4220 (stacked)

Correctness rests on the invariant **commit-conflict ⊆ reader-dirty**: every
conflict must also re-trigger the compute as a reader. That holds **only** with
#4220's leaf-only matcher alignment. On `main` the engine's commit-conflict
matcher is *parent-injecting* while the scheduler's reader-dirty is *leaf-only*,
so a disjoint-sibling write conflicts a compute's commit but does **not**
re-trigger it — and removing the re-queue would then silently drop the compute's
write. This branch is therefore stacked on #4220; its diff includes #4220's
commits until that merges.

Empirically: #4220's engine test *"leaf-only commit conflict — disjoint-key
writers merge"* fails (`ConflictError`, `validateConfirmedReads`) on a pre-#4220
tree and passes on the stack.

## Two conflict layers (why only `ConflictError`)

- **`ConflictError`** — engine, path-granular, server/cross-replica. Reader-dirty
  recoverable (given #4220). → no retry.
- **`StorageTransactionInconsistent`** — client-side **whole-document**
  attestation guard, path-blind, fires on a same-replica race *before* the engine
  matcher. **Not** reader-dirty recoverable → keeps its bounded retry.
- Transient transport errors → keep their bounded retry.

## Verified

Green with the change active: `scheduler-retries`, `scheduler-cfc-trigger-reads`
(incl. *restores consumed trigger reads when a commit conflict re-runs*),
`scheduler-observations`, `scheduler-static-writes`, `scheduler-v2-cutover`,
`conflict-repro`, `patterns-lift`, `patterns-handlers`, `ensure-piece-running`;
fmt + type-check. New `packages/runner/test/compute-conflict-recovery.test.ts`
pins the reader-dirty half of the invariant; the commit-conflict half is #4220's
`v2-engine-test` *"leaf-only commit conflict — disjoint-key writers merge"*.

## Notes

- Direction approved by **seefeld** ("retries are just for handlers; computes
  should just ignore the conflict, since by definition they'll get called again
  with the latest state"). Contra that, computes *did* retry up to 10× before
  this — this aligns behavior with the intent and removes the budget/zombie.
- The earlier "load-bearing assumption (worth a scheduler-owner sanity check)" is
  now addressed: the conflict→reader-dirty recovery is exactly the
  commit-conflict ⊆ reader-dirty invariant, established by #4220 and pinned by the
  new test.
- Spec docs updated to match (`docs/specs/pull-based-scheduler/README.md`,
  `docs/specs/scheduler-v2/current-system-inventory.md`), which had described the
  old re-dirty + pending + tick retry for *all* commit errors.

---
## Summary by cubic
Reactive computes and effects no longer re-queue or consume retry budget on commit conflicts. On a `ConflictError` we just resubscribe and restore trigger reads; other transient errors still retry and precondition failures remain non‑retryable.

- **Bug Fixes**
  - Conflict handling: added `isConflictRejection`; on conflict, restore CFC trigger reads and resubscribe, then return — no dirty/pending/tick, no retry counter; removed `readyToRetry` gating for conflicts.
  - Retry behavior: non-conflict, non‑permanent errors still retry up to `MAX_RETRIES_FOR_REACTIVE`; `PreconditionFailedError` stays non‑retryable; resubscribe runs on every error to keep deps live.
  - Tests: added compute and effect recovery tests (plain + debounced) to assert reader‑dirty re‑trigger; updated `scheduler-cfc-trigger-reads` to expect “conflict re‑arms but does not re‑queue” and to keep retries for non-conflict errors.
  - Docs: updated scheduler specs to document that conflicts aren’t retried and don’t spend the budget.

<sup>Written for commit 006b296d0e2f0a24da9122834904018b3c9f085c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4210?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/runner/src/traverse.ts
```diff
@@ -3644,8 +3647,22 @@ export class SchemaObjectTraverser<V extends FabricValue>
...
-        const { ok: val, error } = this.traverseWithSchema(propDoc, propSchema);
+        // When the property is asCell (and reaches here as a sigil link rather
+        // than the inline-value branch above), the descent + pointer resolution
+        // only RESOLVE THE REFERENCE to construct the Cell — they read the link
+        // target's shape, not a value the holder consumes. Those reads must not
+        // become commit-time conflict dependencies (a holder of a reference must
+        // not collide with disjoint writers under the referent's container). They
+        // stay in the journal for reactivity; the holder takes a real dependency
+        // only when it reads THROUGH the Cell in its body. A by-value property
+        // (`hasAsCell` false) is a genuine dependency and is left unmarked.
```
ubik2, line 3662: I think this is wrong.
We generally need to register the read of the link because if the link changes, we need to react to that.

packages/memory/v2.ts
```diff
@@ -114,13 +114,25 @@ export interface ConfirmedRead {
...
+  /**
+   * When true, this is a SHALLOW (shape-only) read — the reader observed the
+   * container at `path` (its key set / existence) but did NOT depend on the deep
+   * values of its descendants. The engine then conflicts only with writes
+   * AT-OR-ABOVE `path` (including key add/remove, whose patch injects the parent
+   * path), not with disjoint deep-value writes strictly below `path`. Strict
+   * subset of the recursive overlap ⇒ never a false-negative. Absent/false ⇒
+   * recursive read (the historical behavior).
+   */
+  nonRecursive?: boolean;
```
ubik2, line 126: Ah, this is great! I hadn't realized that while we maintained this info for the scheduler, we weren't including it in the server message.

packages/runner/src/traverse.ts
mathpirate: ah i should have been clearer but some relevant stuff is changed in https://github.com/commontoolsinc/labs/pull/4220, which this is based on - worth looking at that one first maybe?

claude says:

The link read is still registered — excludeReadFromConflict only removes it from the commit-conflict set (its sole consumer is buildReads, storage/v2.ts:1895), not from the reactivity log. So a repointed link still reader-dirties the holder; what's dropped is just the conflict dependency. Materializing an asCell arg reads the referent's container shape, and a holder of a reference shouldn't collide with disjoint writers under it (the #4178 over-conflict). This is [#4220](https://github.com/commontoolsinc/labs/pull/4220)'s Half A — design note in docs/specs/memory-v2/08-conflict-granularity.md. If you meant the holder should take a hard conflict dep on the link (not just re-react), happy to dig in — that's the design call. Adding a link-repoint reactivity test to make the "still reacts" part concrete.

packages/runner/src/query-result-proxy.ts
ubik2, line 236: We haven't registered a read of `value.length` here.
Again, I think this mostly will work, since we special case array length access, but I'm not sure, and it would be better to be explicit.

ubik2, line 187: We haven't registered a read of `$stream` here.
However, it's not like we really change `{$stream: true}` to `{$stream: false}`, so this isn't likely to be a problem.
Still, it would be better to build a stream check function that does the top level read, then a second read if there's a `$stream` property,

ubik2: I'm think we still want to register the whole read here.
While we don't let people write to fields of the FabricPrimitive, we don't want a situation where the server thinks two primitives are the same because the top level of their representation is the same.
I'm actually not certain whether we need to register this read or not.

packages/runner/src/query-result-proxy.ts
```diff
@@ -159,7 +170,7 @@ export function createQueryResultProxy<T>(
       readTx.getCfcState().dereferenceTraces.slice(traceStart),
     ),
   ]);
-  const value = readTx.readValueOrThrow(link) as any;
+  const value = readTx.readValueOrThrow(link, SHAPE_READ) as any;
```
ubik2, line 173: I opted for performance when creating this function, so it returns the full value, even though it doesn't register a read of the full value.
Seeing this usage, I'm wondering if that's a footgun.

packages/runner/src/query-result-proxy.ts
```diff
@@ -480,7 +491,7 @@ export function createQueryResultProxy<T>(
     getOwnPropertyDescriptor: (target, prop) => {
       if (Array.isArray(target) && prop === "length") {
         const readTx = runtime.readTx(tx);
-        const current = readTx.readValueOrThrow(link);
+        const current = readTx.readValueOrThrow(link, SHAPE_READ);
```
ubik2: another spot where we should also read the `length` property using the tx if Array.isArray(current).

mathpirate:
Thanks — the broader point lands. The shape-read helper hands back the full value but only tracks it shallowly (your `:173` note), and a couple of sites lean on that implicitly, so I'll make them explicit: the `length` read at `:494` and an explicit `$stream` read (or your two-step stream-check helper) at `:179`. `:217` I'll leave — live `length` goes through the recursive get-trap — but happy to make it explicit too.

On `:189` (FabricPrimitive): I dug in, and our shape-read change doesn't govern it — reverting `:173` to the full read changes nothing. The deeper thing I hit is that a `FabricBytes` value doesn't seem to update in place at all (a `set` of new bytes gets diffed away), so there's no change to miss in the first place — real, but pre-existing and orthogonal to this PR. I'll file it separately so we can look at the diff/update path properly rather than fold a half-fix in here. Does that match your read?

packages/runner/src/traverse.ts
mathpirate: The link read is still registered — `excludeReadFromConflict` only removes it from the commit-conflict set (its sole consumer is `buildReads` in `storage/v2.ts`), not from the reactivity log. So a repointed link still reader-dirties the holder; what's dropped is just the *conflict* dependency. Materializing an `asCell` arg reads the referent's container shape, and a holder of a reference shouldn't collide with disjoint writers under it (the #4178 over-conflict this PR targets).

Added a test — `memory-v2-read-compaction.test.ts` → "excludeReadFromConflict reads STAY in the reactivity log" — asserting the marked read lands in the reactivity log while being absent from the conflict set.

If you meant the holder should take a hard *conflict* dependency on the link (not just re-react), happy to dig in — that's the design call.

ubik2: I should trace through this more, but I'm just adding my initial take here:

Claude is correct that we still call `descend` (and thus do a logged read the link) regardless of whether we're encountering an asCell.

I may be misunderstanding what Claude is saying here, but here's the situation I'm talking about.

I do think the holder should take a hard *conflict* dependency on the link. A change to the linked value invalidates our commit.

For example:
cellA:
`{ isAdmin -> <link to cellB/isAdmin> }`

cellB:
`{ isAdmin: true }`

Client 1:
`JSON.stringify(cellA.get())` => `{ isAdmin: true }`

Client 2:
`cellB.key("isAdmin").set(false);`

Client 1:
`if (cellA.key("isAdmin").get()) { cellC.set("User is allowed, because isAdmin = true") }`

That late set from Client 1 should be invalid, because it has a read dependency on cellB.isAdmin, and that has changed.

Edit: also, while these do get the reactivity, the ones in query-result-proxy are still a potential problem.
Really, I should make some tests that check both this example, and the query-result-proxy reactiveness issues.

mathpirate: Built your `isAdmin` example as a test and it holds up — but to be explicit about scope: **this is verified on the branch _as stacked on #4220_, and the conclusion is a property of #4220's scoping, not something that holds on its own.** The `excludeReadFromConflict` exclusion (and the leaf-only matcher it pairs with) is entirely #4220's change; what the test confirms is that #4220 scopes the exclusion to the *reference-resolution* read only, leaving the value read-through as a real conflict dependency:

1. `buildReads` (the conflict set) includes `cellB.isAdmin` after the read-through — only the nonRecursive resolution read is dropped.
2. End-to-end: the holder reads `isAdmin`→true through the link and writes; a concurrent writer sets `cellB.isAdmin = false`; the holder's commit is **rejected**. Your "late set should be invalid" is exactly what happens.

Both checks would fail if the exclusion were over-broad. Test committed on #4220 (`ascell-link-read-through-conflict.test.ts`) so it travels with the change it's verifying. (Separately: the query-result-proxy `$stream`/`length` hardening landed; the FabricBytes write drop is #4263.)

mathpirate:
@seefeldb @ubik2 — capturing the conceptual contrast between #4237's and #4210's approaches to commit conflicts, since they're easy to conflate when reviewing the stack. Berni, the load-bearing premise is flagged at the end — would value your read on whether it holds as a general claim.

**Design note: #4237 and #4210 are two answers to the same question**

Both PRs respond to the same event — a reactive action's optimistic commit fails with a `ConflictError` (a stale read: an input moved between the action's read and its commit). They embody different philosophies about *whose job it is* to make that commit eventually succeed.

**#4237 — the action is a transactional retrier; gate its retries so they're productive.**
A conflict means our local replica's read lagged the server. So retry — but first wait until the local seq has caught up (`waitForCaughtUpLocalSeq`, surfaced as the `readyToRetry` promise) so the retry reads fresh data instead of immediately re-conflicting. The action owns its commit and re-attempts it; each attempt draws on the reactive retry budget. *Model: the action drives itself to success; the system's role is to keep each retry from spinning against stale data.*

**#4210 — the action is a node in a reactive graph; the scheduler re-runs it as data flows.**
A reactive action is *not* a transactional retrier. A conflict means an input moved — and the write that moved it has already dirtied this action's still-subscribed reads. So the scheduler will re-run it through ordinary reader-dirty propagation, against the latest state, as a natural consequence of the input changing. The conflict is a *wait-to-be-re-triggered*, not a retry: the action resubscribes and returns, consuming no budget. *Model: the action doesn't own its commit's success — the data flow does, by re-deriving it when inputs change.*

**The real difference is agency over re-execution.** #4237 has the action re-drive itself (it retries). #4210 has the data flow re-drive the action (reader-dirty re-triggers it). They can't both own the conflict path — that's double work, and the budget would be spent on something the reactive system handles for free.

**Why #4210's reactive re-trigger subsumes #4237's gated retry (given #4220):**
- #4210 is sound only under the invariant `commit-conflict ⊆ reader-dirty` — every write that conflicts a commit must also dirty that action as a reader. #4220 (leaf-only matcher alignment) establishes this; without it a disjoint-sibling write conflicts the commit without re-triggering the reader, and #4210 would silently drop the action's write. Hence the stacking.
- A reader-dirty re-run only fires *after* the conflicting write has applied locally — exactly the "caught-up" condition #4237's gate waited for. So the reactive re-trigger inherits #4237's freshness guarantee for free, with no budget and no explicit wait.
- This is only about #4237's *scheduler-layer* retry gate. Its *storage-layer* catch-up machinery (read-repair / `waitForConflictReadRepair`) is untouched and still runs.

**The boundary #4210 deliberately keeps:** only `ConflictError` leaves the retry path. Non-conflict, non-permanent errors — a transient transport failure, or the path-blind `StorageTransactionInconsistent` same-replica guard that fires *before* the engine's granular matcher — are *not* reader-dirty-recoverable, so they keep their bounded retry. Permanent (precondition) rejections stay non-retryable. The two philosophies coexist, partitioned by error class.

**Scope — computations vs effects:** the conflict-skip applies to all reactive actions, not just computations. Effects recover the same way: an effect whose commit hits a cross-replica `ConflictError` takes the no-retry path and is re-run by reader-dirty (plain *and* debounced — regression test `effect-conflict-recovery.test.ts`). The genuinely side-effectful actions (event handlers) run on a separate dispatch/retry path and are untouched. So "effects need the retry" doesn't bite within #4210's actual scope.

The piece worth a second pair of eyes: **is the `commit-conflict ⊆ reader-dirty` subsumption airtight as a general claim**, or only for the cases #4220's matcher currently covers? That premise is what lets the reactive re-trigger fully replace the gated retry.

packages/runner/src/traverse.ts
mathpirate: For posterity (resolved on `main` via #4220): your `isAdmin` scenario is exactly the shipped contract — a value read *through* the link is a hard commit-conflict dependency; only the reference-*resolution* (topology) read is excluded. `packages/runner/test/ascell-link-read-through-conflict.test.ts` encodes it: the holder reads `cellB.isAdmin`→`true` through the link and writes, a concurrent `cellB.isAdmin = false` **rejects** the holder's commit, and it asserts `cellB.isAdmin` lands in the commit-conflict set. The query-result-proxy reactivity worry you flagged is covered by `query-result-proxy-shape-reactivity.test.ts`, and the "excluded read still reacts" property by `memory-v2-read-compaction.test.ts`.

packages/memory/v2.ts
mathpirate: For posterity: right — it was already tracked for the scheduler; #4220 carries it through to the server message so the engine can apply shallow (shape-only) conflict granularity. The follow-up to split this flag into keyset vs existence/topology reads (removing the remaining over/under-conflict edges) is tracked as CT-1783.

packages/runner/src/query-result-proxy.ts
mathpirate: For posterity (resolved on `main`): the array get-trap now reads the array **fully** on `length` access (`query-result-proxy.ts:252-254`), so a length change is tracked explicitly rather than relying on the special-casing.

mathpirate: For posterity (resolved on `main` via #4220): explicit `$stream` read added per your stream-check suggestion — `query-result-proxy.ts:179-180` does the top-level read, then a second read of `$stream` when the property is present.

mathpirate: For posterity (resolved on `main`): FabricPrimitive leaves now register a full value read (`query-result-proxy.ts:204`), so two primitives sharing a top-level representation aren't conflated. The deeper issue this surfaced — FabricBytes values not updating in place (a `set` of new bytes gets diffed away) — was pre-existing and orthogonal; filed and since fixed as #4263.

packages/runner/src/query-result-proxy.ts
mathpirate: For posterity (resolved on `main`): agreed it was a footgun. The shape-read helper's shallow tracking is now documented (the `shallowEqual` note in `reactive-dependencies.ts`), and the sites that implicitly leaned on it were made explicit — `length` (this thread + :494) and `$stream` (:187).

packages/runner/src/query-result-proxy.ts
mathpirate: For posterity (resolved on `main`): `getOwnPropertyDescriptor` now reads the array fully for `length` — `query-result-proxy.ts:511-515` ("Read the array fully (not SHAPE_READ) so the length descriptor tracks element add/remove").

PR #4221, opened 2026-06-18
cf-tabs: don't flash the first tab while $value is unresolved (mount flicker)

bfollington:
## Problem

`cf-tabs` "flickers between two tabs after a few interactions" — a long-standing, repeatedly-chased bug. This is a **render-path race**, distinct from the cell-write feedback loops already fixed by [#4134](https://github.com/commontoolsinc/labs/pull/4134) (CT-1746, gesture-only `cf-change`) and [#4151](https://github.com/commontoolsinc/labs/pull/4151) (CT-1752, pure-on-mount). Those killed the *"Too many iterations"* scheduler loop; this flicker never writes the cell, so it was invisible to those fixes and their tests.

## Root cause (reproduced live)

On every (re)mount, the synchronous selection-syncs (`firstUpdated` + the two `slotchange` listeners + Lit's `updated()`) read the bound `$value` cell **before** the cell controller's subscription delivers its value. For a durable cell (`.for("activeTab")`) that read is `""` (`StringCellController` coerces the not-yet-delivered value to empty).

The old no-match fallback treated `""` as a stale value and **snapped to the first tab**; a beat later the real value (e.g. `delta`) arrived and selection jumped to it — a visible **first-tab → real-tab** flip on each mount. Once you've selected a non-first tab, the durable cell persists it, so every remount flickers between the first tab and your tab. (That's also why it only shows "after a few interactions" — on the default first tab the fallback agrees.)

Captured live with an instrumented `updateTabSelection` (durable value = `delta`):

| call | value read | tab selected |
|---|---|---|
| 1–2 | `delta` | — (no tabs yet) |
| 3–6 | `""` | **`alpha`** ← fallback to first tab |
| 7 | `delta` | `delta` |

## Fix

In `updateTabSelection`, treat an empty/undefined `currentValue` as *"unresolved"* and **hold the current selection** (return early) rather than falling back to the first tab. The first-tab fallback now applies only to a **non-empty** value that matches no tab (a genuinely stale selection). Selection-sync still never writes the cell.

After the fix (same instrumentation, re-mount with `delta`): the empty syncs are suppressed (`selectedAfter: []`) and only the resolved `delta` is applied — `alpha` never flashes.

## Tests

- Pure-on-mount tests now assert empty → *nothing selected* (was first-tab).
- New **mount-flicker regression test**: unresolved value → no flash, then resolves straight to a non-first tab.
- `FakeTabPanel`'s default `hidden` corrected to `true` to mirror the real `CFTabPanel` constructor (the old `false` made suppressed panels look visible in the harness).

`deno test` 13/13 steps green; `deno fmt` + `deno lint` clean.

## Known residual / optional follow-up

There's still a brief *"no tab selected"* window before the durable value resolves — sub-2ms in light panels (won't paint), but possibly a one-frame flash in heavyweight panels (e.g. `home.tsx`). The clean follow-up is to coalesce the ~6 redundant per-mount sync paths into a single microtask apply using the latest value. Left out here to keep the change minimal and synchronous (it would require reworking the synchronous test harness).

---
## Summary by cubic
Stops `cf-tabs` from flashing the first tab on mount by treating empty/undefined `$value` as unresolved and showing no selection until it resolves; an explicit `""` now clears the selection. The first-tab fallback only runs for non-empty stale values, removing the first-tab → real-tab flicker.

- **Bug Fixes**
  - In `updateTabSelection`, skip the first-tab fallback for `""`/`undefined`; let the apply loop run with an empty value to deselect. Fallback only for non-empty no-match, and never write to the cell.
  - Add a mount-flicker regression test and a clear-on-empty test; update pure-on-mount assertions to expect “nothing selected” while unresolved.
  - Set `FakeTabPanel.hidden` default to `true` to match `CFTabPanel`.

<sup>Written for commit 893f83de3ca1edd4879d68875758750c993f46fa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4221?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ui/src/v2/components/cf-tabs/cf-tabs.test.ts
+++ b/packages/ui/src/v2/components/cf-tabs/cf-tabs.test.ts
@@ -416,6 +416,23 @@ describe("CFTabs pure-on-mount contract (safe inside computed)", () => {
     expect(writes()).toBe(0); // pure read throughout — no cell write
   });
 
+  it("clears the selection when the bound value is explicitly set to empty", () => {
+    // Empty is not only the mount transient — a consumer can clear the bound
+    // cell ("" / unset) to mean "no tab active". The early-return form held the
+    // prior selection (stale tab + panel stuck visible); selection-sync must
+    // instead drop it. (Still write-free — clearing is a pure read.)
+    const { cell, writes } = countingCell("progress");
+    const h = makeTabs(cell, ["active", "progress", "pending"]);
+    expect(h.selectedTab()).toBe("progress"); // initially selected
+    expect(h.visiblePanel()).toBe("progress");
+
+    pushUpdate(cell, ""); // explicit clear
+
+    expect(h.selectedTab()).toBe(undefined); // selection dropped, not stale
+    expect(h.visiblePanel()).toBe(undefined);
+    expect(writes()).toBe(0);
+  });
+
   it("a cell-driven change to an unmatched value does not write back", () => {
     const { cell, writes } = countingCell("active");
     const h = makeTabs(cell, ["active", "progress"]);
--- a/packages/ui/src/v2/components/cf-tabs/cf-tabs.test.ts
+++ b/packages/ui/src/v2/components/cf-tabs/cf-tabs.test.ts
@@ -106,7 +106,10 @@ class FakeTab {
 
 class FakeTabPanel {
   value: string;
-  hidden = false;
+  // Mirror the real CFTabPanel constructor default (hidden = true). cf-tab-panel
+  // starts hidden and is revealed only when updateTabSelection() matches it; a
+  // `false` default would make panels look visible before any sync runs.
+  hidden = true;
   attributes = new Map<string, string>();
   constructor(value: string) {
     this.value = value;
@@ -342,7 +345,10 @@ describe("CFTabs cf-change emission contract (CT-1746)", () => {
  * computed: a "Too many iterations" CPU-spin (the CT-1677 settle class).
  *
  * These tests pin the contract: mount and cell-driven sync produce ZERO cell
- * writes, while the no-match fallback still selects the first tab VISUALLY.
+ * writes. The no-match fallback selects the first tab VISUALLY only for a
+ * NON-empty stale value; an empty / unresolved value (the durable-$value
+ * mount transient) holds the current selection rather than flashing the first
+ * tab — see the mount-flicker regression test below.
  */
 describe("CFTabs pure-on-mount contract (safe inside computed)", () => {
   // Count writes to the bound cell by wrapping `.set` before binding.
@@ -357,8 +363,8 @@ describe("CFTabs pure-on-mount contract (safe inside computed)", () => {
     return { cell, writes: () => writes };
   }
 
-  it("does NOT write the bound cell on mount when the value matches no tab", () => {
-    // Empty cell matches none of the tabs — the no-match path.
+  it("does NOT write the bound cell on mount when the value is empty / unresolved", () => {
+    // Empty cell — the durable-$value mount transient (value not delivered yet).
     const { cell, writes } = countingCell("");
 
     // makeTabs() binds the controller and runs updateTabSelection() — i.e. the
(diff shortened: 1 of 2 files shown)


bfollington:
## cf-review (automated)

_cf-review via Claude `opus-4-8[1m]`, on behalf of @bfollington_

**Scope** · `cf-tabs: don't flash the first tab while $value is unresolved` · **coverage** deep-read (both changed files + `cell-controller.ts`, `cf-tab-panel.ts`, `mock-cell-handle.ts`, and all sampled `cf-tabs` consumers/docs) · **checks run** `deno test packages/ui/src/v2/components/cf-tabs/cf-tabs.test.ts` (13/13 green), `deno lint` (clean).

**What it does.** In `CFTabs.updateTabSelection()`, an empty/undefined `currentValue` is now treated as *unresolved* and returns early (holds current selection) instead of falling through to the first-tab fallback. The first-tab fallback now fires only for a **non-empty** value that matches no tab. Tests are updated to assert empty → nothing-selected, a new mount-flicker regression test is added, and `FakeTabPanel.hidden` default is corrected to `true`. Scope is tight and coherent.

### 🔴 Blocking
None.

### 🟡 Improvements
None blocking. The PR's own "Known residual" section is honest and correctly scoped as a follow-up (the brief no-selection window before durable resolution; coalescing the ~6 redundant per-mount sync paths). No action needed for this PR.

### ⚪ Nits
None worth holding the PR.

### Verification notes (all **verified**)
- **Root-cause claim is accurate.** `StringCellController.getValue()` (`cell-controller.ts:331-337`) coerces a CellHandle via `.get() || ""`, so a not-yet-delivered durable value and a genuinely-empty value both surface as `""`. The code truly cannot distinguish them synchronously — the early-return + later `onChange` re-sync is the right shape. The resolution path is guaranteed: the controller's `onChange` (`cf-tabs.ts:113-129`) unconditionally calls `updateTabSelection()` with the real value.
- **`FakeTabPanel.hidden = true` is a real fidelity fix, not a test bent to fit.** Confirmed the real `CFTabPanel` constructor sets `this.hidden = true` (`cf-tab-panel.ts:79`). The old `false` made the harness diverge from production — panels looked visible before any sync. This *strengthens* the harness.
- **No consumer regression.** Every `cf-tabs` consumer binds an **initialized** cell: `home.tsx` `new Writable("spaces")`, `store-mapper` `new Writable(...)("map")`, `reading-list` `new Writable("all")`, `multi-option-selection` `Default<"tab1">`. None rely on an empty cell to default to the first tab, and none `.set("")` a tabs-bound cell. The behavior change (empty → nothing) is invisible to them.
- **Docs stay coherent.** `docs/common/patterns/view-switching.md` documents the idiom with an initialized cell (`new Writable("spaces").for("activeTab")`) and never promises an "empty shows first tab" guarantee. The north-star coherence test passes — nothing authoritative is left actively wrong. The new code comment even states the migration path ("consumers that want a default must initialize the cell").
- **Tests guard the right principle.** The renamed pure-on-mount tests still pin the zero-write-on-mount contract (the CT-1677/CT-1746 feedback-loop class), and the new regression test reads as a user story (unresolved → no flash → resolves straight to non-first tab, zero writes). Coverage was tightened, not dropped.
- **No anti-duplication / framework-fit issues.** No hashing/serialization/cloning/identity machinery touched. The fix leans *into* the reactive model (pure-read selection sync, write only on gesture) rather than fighting it.

### Verdict
**No blocking issues.** Correct, well-tested, coherent fix for a real render-path race; the test-harness `hidden` correction improves fidelity. Ship it. (PR is in Draft — promote to ready when satisfied.)

bfollington:
## Addressed review P2 (stale selection on explicit clear)

Valid catch. The first commit returned early on an empty/undefined value, which left the prior tab + panel **stuck selected** when a consumer explicitly clears the bound cell to `""` (i.e. "no tab active").

**Root cause:** the early-return skipped the apply loop entirely, so nothing deselected the previously-selected tab / re-hid its panel.

**Fix** (`b5ce06a56`): don't return early — only apply the first-tab fallback for a *non-empty* no-match, and let the existing apply loop run with an empty `effectiveValue`. An empty value matches no tab, so the loop deselects every tab and hides every panel = clears the selection.

This keeps the flicker fix intact: clearing selects **nothing** (never the first tab), and on a fresh (re)mount nothing was selected so it's a visual no-op, with the real value re-syncing a beat later via the controller's `onChange`.

Verified live:
- Re-mount over durable `"delta"` → `(none) → delta`, first tab never flashes.
- `cell.set("")` over a selected tab → selection drops (was: stayed stuck on the old tab).

Added a regression test (`clears the selection when the bound value is explicitly set to empty`). 14/14 steps green; `deno fmt` + `deno lint` clean.

bfollington:
## Integration test failure was a missing main fix (not this change)

The consistently-failing **Pattern Integration Tests (4/4)** check was `home-space profile creation › "creates a second profile from the picker (multi-profile append)"`, timing out on `#home-profile-summary` waiting for "Alan Turing".

**Diagnosis:** not caused by the cf-tabs change. The failure probe showed the Profile `cf-tab-panel` rendering correctly (`visible: true`, picker present) — but with two "Ada Lovelace" profiles instead of the expected "Alan Turing", i.e. the **second profile append was durably lost**, not a tab-visibility problem.

That's exactly the bug fixed by [#4231](https://github.com/commontoolsinc/labs/pull/4231) ("settle cross-space sync around home profile creation", merged earlier today): `createProfile` now waits for `waitForRuntimeSynced` so each cross-space `inSpace` append commits against a reconciled state instead of being reverted by in-flight rehydration churn. This branch was cut from `5ad361256`, **14 commits before #4231 landed**, so it didn't have the fix.

**Resolution:** rebased onto current `main` (picks up #4231 + 13 others). The cf-tabs diff is unchanged — still only `cf-tabs.ts` + `cf-tabs.test.ts`, 14/14 unit steps green, fmt + lint clean. New CI run is in flight.

PR #4225, opened 2026-06-18
Replace RuntimeDisposedError with proactive cancellation

Hixie:
## Why

Disposing a runtime connection used a bespoke RuntimeDisposedError to mark operations that lost the race with teardown as benign, and roughly fifteen call sites across five packages caught that error to suppress logging. That model never closes the bug class: every new fire-and-forget call into the connection is a latent flake until someone remembers to add the suppression, and it depends on a custom error class surviving every layer — including serialization across the worker boundary. The original symptom was a flaky integration test that recorded a console error during teardown.

This replaces the whole model with proactive cancellation built on the standard AbortController, so the situations that produced the error no longer occur, and deletes RuntimeDisposedError entirely.

## What changed

One-way notifications. VDomEvent and VDomBatchApplied are fire-and-forget — the caller ignores the reply — yet they were sent through the request machinery, which allocated a deferred, a pending entry, and a timeout to await a reply that was never read. When disposal rejected that pending request, the catch handler logged it. They are now genuine main-to-worker notifications: no msgId, no pending entry, no reply. With nothing pending, disposal has nothing to reject.

Proactive cancellation in the connection. RuntimeConnection owns an AbortController instead of a #disposed flag. dispose() aborts it synchronously first, which runs every registered consumer teardown (the renderer detaches its DOM and batch listeners and drops its DOM) and settles every in-flight request as the standard AbortError, so nothing issues or awaits a request after that point — before the transport is touched. Every former #disposed check is signal.aborted, and every former isRuntimeDisposedError suppressor checks the abort state instead of sniffing an error type. The signal is exposed connection -> RuntimeClient -> RuntimeInternals so consumers can recognize cancellation.

Tolerate vs assert, by guarantee. Benign tolerance is kept only where there is no strict guarantee an operation cannot follow disposal: request() (unowned one-shot callers can resume after an unrelated await), the worker boundary in _handleMessage (the worker keeps sending until terminated), and component-driven unsubscribes. Where proactive cancellation does give a strict guarantee — #notify, reached only from the renderer, which is detached synchronously on abort — a post-disposal send is a contract violation and now throws.

VDOM capability as an invariant. That throw is only sound if the renderer is guaranteed torn down, so the convention is made an invariant: VDOM capability (mount/unmount/event/ack/batch-subscription) is reachable only through connection.attachVDom(onDispose), whose teardown argument is required and registered on the lifetime signal. The bare connection exposes no VDOM IPC. A renderer therefore cannot exist without being registered for disposal teardown.

Disposal waits for the worker's flush. The Dispose request still reaches the worker, which tears down its VDOM mounts and flushes pending storage writes (storageManager.synced()) before replying. dispose() waits for that confirmation under the default request timeout, so a normal flush is durable before the transport is torn down. Once the connection is dead, _handleMessage ignores every incoming message without warning, except the reply to the pending Dispose request it is waiting on.

Disposal-raced log sites gated on the signal. Sites that logged (or, for fire-and-forget paths, leaked an unhandled rejection) on a disposal-raced request now recognize cancellation: cell-handle set/send, favorites-manager, lib-shell RuntimeInternals, shell AppView, FavoriteButton, BodyView, HeaderView, DebuggerView's logger toggles, cf-render, cf-cell-link, cf-code-editor, cf-profile-badge, and cf-cfc-authorship's label refresh. RuntimeDisposedError and its predicate are deleted along with the shared module and barrel export.

## Implications

In-flight requests at disposal still settle as a rejection, but with the standard AbortError, observed only by consumers that check signal.aborted. Fire-and-forget paths produce no rejection at all. "Used after dispose" is structurally close to unreachable for owned consumers, and the residual tolerance is confined to the irreducible worker boundary. The per-request timeout override was removed; every request, Dispose included, uses the single default timeout.

## Tests

The connection disposal suite specifies the new contract: in-flight requests settle as cancellation, registered teardowns run synchronously before transport teardown, requests after dispose do not reach the transport, a post-disposal notification throws, attachVDom registers/detaches the teardown, dispose waits for the worker's confirmation, and stray messages after disposal are ignored silently. Worker-side handleNotification dispatch, cell-handle write suppression, and each gated UI/runtime log site (including cf-profile-badge and cf-cfc-authorship) are unit-tested via prototype handlers where the component needs no DOM. An integration test logs out while a piece is rendered — disposing the runtime via the RootView swap with the renderer, favorites subscription, and slug poll all live — under the allowlist-free harness.

Full workspace typecheck and test suite pass; the disposal-heavy shell and runtime-client integration suites pass against a live toolshed.

---
## Summary by cubic
Replaced the bespoke disposal error with AbortController-based cancellation and converted VDOM event/ack traffic to one‑way notifications. This removes teardown noise, makes VDOM access opt‑in via `attachVDom`, and fixes renderer construction and disposal‑race guards to read the right runtime.

- **Refactors**
  - `RuntimeConnection`: owns a lifetime `AbortController`; `dispose()` aborts first (runs teardowns, settles in‑flight as `AbortError`), then awaits worker `Dispose` under the default timeout, then tears down the transport. Exposes `signal`, `onDispose`, and `attachVDom(onDispose)` returning a `VDomConnection`. Late messages are ignored; post‑disposal notifications throw; idempotent.
  - Protocol/worker: introduced `ClientNotificationType`/`IPCClientNotification`; `VDomEvent`/`VDomBatchApplied` are one‑way notifications. Transports accept notifications; Web Worker uses `isIPCClientNotification` and `RuntimeProcessor.handleNotification`.
  - HTML renderer: uses `attachVDom`; tolerates an unset session during synchronous teardown; skips batch subscription if already torn down; `render()` returns early when disposed; `renderViaWorker` unregisters from `activeRenders` on dispose and only swallows errors when `connection.signal.aborted`.
  - UI/runtime: guards consult the operation’s runtime signal, not an ambient client. Fixed in `cf-code-editor` (pattern runtime), `cf-profile-badge`/`cf-cell-link` (resolving cell runtime), `cf-render`, and shell `FavoriteButton`, `HeaderView`, `BodyView`, `DebuggerView` (logger resets). `RuntimeClient` and `lib-shell` `RuntimeInternals` now expose `signal`.
  - Cleanups/tests: removed `RuntimeDisposedError` and its guard; dropped the obsolete VDOM ack test; added coverage for connection disposal, worker notifications, renderer construction against a disposed connection, and disposal‑race guards across UI and shell.

- **Migration**
  - Replace `isRuntimeDisposedError(...)` checks with `rt.signal.aborted` or `conn.signal.aborted`, read from the runtime the operation ran on.
  - Use `connection.attachVDom(onDispose)` for all VDOM IPC; stop calling VDOM methods on the bare connection.
  - Use `onDispose()` for synchronous teardown; rely on the single default request timeout (including `Dispose`).

<sup>Written for commit bc5f47cf3e9614ba6700981420dedc75477c04dc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4225?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mathpirate (approved):

packages/html/src/main/renderer.ts
```diff
@@ -82,8 +82,13 @@ export class VDomRenderer {
...
     });
 
-    // Subscribe to VDomBatch notifications
-    this.connection.on("vdombatch", this.handleVDomBatch);
+    // Attach as a VDOM consumer. attachVDom requires the teardown, which runs
+    // synchronously when the connection is disposed (dropping local listeners
+    // and DOM without a per-mount unmount round-trip), and is the only way to
+    // obtain VDOM capability — so a renderer cannot exist without being torn
+    // down on disposal.
+    this.session = options.connection.attachVDom(() => this.disposeLocal());
```
mathpirate, line 90: **`VDomRenderer` throws mid-construction if the connection is already disposed.**

`attachVDom` calls `onDispose(teardown)` first, and `onDispose` runs the teardown *synchronously* when `signal.aborted` (connection.ts:141). The teardown is `() => this.disposeLocal()`, and `disposeLocal` dereferences `this.session.offBatch(...)` (line 206) — but `this.session` isn't assigned yet (we're still evaluating the RHS of this line). So constructing against an already-aborted connection throws `TypeError: cannot read 'offBatch' of undefined`. (Reproduced against the real class.)

Reachable through `render()` → `renderViaWorker` (render.ts:106 constructs unconditionally) when a render begins after a logout/runtime-swap. It's currently **masked** in `cf-render` — the throw lands in `_handleRenderError`, whose new abort-guard swallows it — so no live crash today. But the renderer's own invariant ("a renderer cannot exist without being torn down on disposal") is violated: it throws half-built, and the containment is incidental rather than by design.

Note: cubic flagged this (P2, confidence 9) and marked it “✅ Addressed in 511a3b7”, but that commit isn't in this PR's head (`6e6db2e`) — the fix didn't land here.

Suggested fix (also closes the listener re-registration a naive optional-chain fix would leave):
```ts
this.session = options.connection.attachVDom(() => this.disposeLocal());
if (!this.disposed) this.session.onBatch(this.handleVDomBatch);   // don't re-register on an already-disposed renderer
// and in disposeLocal():
this.session?.offBatch(this.handleVDomBatch);
```
Rightest shape: have `attachVDom` return the session and register the teardown *after* assignment — decoupling “acquire capability” from “run teardown now if dead.” Worth a regression test that aborts the mock lifetime before constructing; the existing `MockConnection` already supports it.

— Claude, on behalf of Gideon

packages/ui/src/v2/components/cf-code-editor/cf-code-editor.ts
```diff
@@ -654,6 +654,9 @@ export class CFCodeEditor extends BaseElement {
         navigate,
       });
     } catch (error) {
+      // A disposal race (logout, runtime swap) cancels the create; that is
+      // cancellation, not a failure to surface.
+      if (this.runtime?.signal.aborted) return;
```
mathpirate: **Guard checks the wrong runtime.** The op ran against `rt = this.pattern.runtime()` (lines 625/639), but this guards on `this.runtime` — the `@consume(runtimeContext)` ambient client. On logout RootView sets `this.runtime = undefined`, so `undefined?.signal.aborted` is falsy and “Error creating backlink” logs anyway — exactly the teardown console-noise this PR is removing. Suggest:
```ts
if (rt.signal.aborted) return;   // rt = the runtime the op actually ran against (non-optional)
```

— Claude, on behalf of Gideon

packages/ui/src/v2/components/cf-profile-badge/cf-profile-badge.ts
```diff
@@ -694,6 +694,9 @@ export class CFProfileBadge extends BaseElement implements SealLivenessClient {
       }, { includeCfcLabel: true });
     } catch (e) {
       if (generation !== this._resolveGeneration || !this.isConnected) return;
+      // A disposal race (logout, runtime swap) cancels the resolve; that is
+      // cancellation, not a failure to surface.
+      if (this.runtime?.signal.aborted) return;
```
mathpirate: **Same wrong-runtime guard as cf-code-editor.** The op ran against `cell.runtime()` (`cell = this.profile`), but this guards on the `@consume(runtimeContext)` `this.runtime`. On logout that's `undefined`, so the resolve's AbortError still logs “failed to resolve profile cell”. (The `generation`/`isConnected` pre-guard above shields some races, but not the logout-undefined case.) Suggest:
```ts
if (cell.runtime().signal.aborted) return;
```

— Claude, on behalf of Gideon

packages/runtime-client/client/connection.ts
```diff
@@ -207,33 +286,38 @@ export class RuntimeConnection extends EventEmitter<RuntimeConnectionEvents> {
...
+    // every in-flight request as cancellation, so nothing issues or awaits a
+    // request after this point — before the transport is touched.
+    this.#lifetime.abort();
+
+    // One coarse "dispose everything" message replaces per-consumer teardown
+    // round-trips. The Dispose request is exempt from the abort above, so it
+    // still reaches the worker, which flushes pending storage writes before
+    // replying. We wait for that confirmation under the default request timeout,
+    // so a normal flush is durable before the transport is torn down.
+    await this.request<RequestType.Dispose>({ type: RequestType.Dispose })
```
mathpirate, line 317: **Behavior change worth a deliberate call:** removing the per-request timeout override means `dispose()` now waits on the worker's Dispose reply under the 60s default — it was `5000 // Short timeout for dispose` before. Common case is unaffected (and the wait is what makes the storage flush durable), but a *wedged* worker now blocks logout teardown ~12× longer. Flagging so 60s is an intentional choice rather than an incidental consequence of dropping the param — fine either way.

— Claude, on behalf of Gideon

packages/ui/src/v2/components/cf-cfc-authorship/cf-cfc-authorship.ts
```diff
@@ -726,10 +726,19 @@ export class CFCFCAuthorship extends BaseElement {
...
-      this.kind ?? DEFAULT_AUTHORSHIP_KIND,
-    );
+    let view: typeof this.cfcLabel;
+    let pendingResolution: boolean;
+    try {
+      ({ view, pendingResolution } = await readLabelView(
+        this.value,
+        this.kind ?? DEFAULT_AUTHORSHIP_KIND,
+      ));
+    } catch {
```
mathpirate, line 736: Nit: this bare `catch {}` swallows *all* errors, not just the disposal-race abort — the one site that diverges from the `signal.aborted` convention the rest of the PR establishes. Intentional (“matching refreshAuthorClaim”), but a genuine `readLabelView` failure now goes silent with no diagnostic. Fine to leave; noting for consistency.

— Claude, on behalf of Gideon

packages/ui/src/v2/components/cf-render/cf-render.ts
```diff
@@ -383,6 +383,9 @@ export class CFRender extends BaseElement {
   }
 
   private _handleRenderError(error: unknown) {
+    // A disposal race (runtime swap, logout) cancels an in-flight cell sync;
+    // that is cancellation, not a render failure to surface.
+    if (this.cell?.runtime().signal?.aborted) return;
```
mathpirate: Nit: `signal?.aborted` — the `?.` after `signal` is dead (`RuntimeClient.signal` is non-nullable) and inconsistent with the other converted sites. (FWIW this is the *one* component that gates on the correct runtime — `this.cell.runtime()` — so the others should adopt this one's ownership, just not its optional chaining.)

— Claude, on behalf of Gideon

mathpirate (commented):
Gideon asked me (Claude) to take a thorough independent pass over this PR, and to post what's worth raising. He's approved it — this is a clear improvement over the `RuntimeDisposedError` model, and the design is genuinely strong: the `AbortController` lifetime is fully private (only the read-only signal escapes), promoting `VDomEvent`/`VDomBatchApplied` to first-class one-way notifications — out of the `Commands` map, so they can no longer allocate a pending entry, await a phantom reply, or time out — is the real root-cause fix rather than a patch, and `connection.test.ts` pins the new contract precisely.

The inline notes below are suggestions, not merge blockers (address now or later). One is a real, currently-masked defect; the rest are low/nit.

**One thing to _not_ change:** the similar-looking `this.runtime?.signal.aborted` guards in `cf-cell-link.ts` (resolve + link) read the *correct* signal despite the pattern, because RootView disposes-then-awaits-then-assigns and the `generation` bump pre-empts — so at catch time `this.runtime` is still the old, aborted client. Only `cf-code-editor` and `cf-profile-badge` (flagged below) consult the wrong runtime.

— Claude, on behalf of Gideon

Hixie:
@mathpirate, thanks for the review. Here's Claude's response (note its disagreement about half way down under "Where I diverged"):

----
Thanks for the detailed review — this caught real problems and the suggestions were precise. Here's what I changed, what I deliberately left, and one place where I ended up diverging from your guidance. I've called that one out prominently with evidence, because I'd rather you sanity-check my reasoning than have it slip through.

## Addressed as suggested

- **renderer.ts (the construction-time throw).** Fixed so that constructing a `VDomRenderer` against an already-disposed connection no longer throws. `disposeLocal` now reads `this.session?.offBatch(...)` (the session field genuinely is unassigned at the moment the teardown runs synchronously inside `attachVDom`), the batch subscription is skipped when the renderer is already disposed, and — to make the invariant hold by design rather than by luck — `render()` now returns early when disposed, so a torn-down renderer never proceeds to mount. Added a regression test that disposes the mock connection before construction and asserts both "no throw" and "no mount".
- **cf-code-editor.ts.** The guard now reads `rt.signal.aborted`, where `rt = this.pattern.runtime()` is hoisted above the `try` so it is in scope in the catch. The test was updated to prove the fix: the ambient `this.runtime` is left undefined (the logout state) while the pattern's own runtime is the aborted one.
- **cf-profile-badge.ts.** The guard now reads `cell.runtime().signal.aborted` (`cell` is already in scope and non-null in the catch). Test updated the same way.
- **cf-render.ts.** Removed the dead `?.` after `signal`, since `RuntimeClient.signal` is non-nullable. This is also the one component that was already gating on the correct runtime (`this.cell.runtime()`), so the others were brought in line with it.

## Left as-is, matching your note

- **cf-cfc-authorship `refreshLabel` bare catch.** I agree this is worth flagging, but I left it. There is no clean in-scope signal to gate on here: `this.value` is typed `unknown`, and `this.runtime` is exactly the ambient context client that is the wrong one to consult. Gating it would mean either sniffing the error type (which this PR is specifically removing) or threading a runtime handle through, both of which seem worse than the contained swallow. Happy to revisit if you'd prefer one of those.
- **connection.ts dispose timeout.** Leaving dispose to wait under the standard request timeout is intentional, not an accident of dropping the per-request override. The wait is what makes the worker's storage flush durable before teardown, and the dispose site documents that. A wedged worker does block logout longer than it used to, which is the trade-off, but I think it's the right one. [this was explicitly my call; I'm actively trying to remove as many timeouts, sleeps, etc, as possible because they almost always lead to confusing UX; happy to discuss further -Hixie]

## Where I diverged: cf-cell-link

You asked me specifically not to touch the `this.runtime?.signal.aborted` guards in cf-cell-link, on the reasoning that RootView disposes, then awaits, then assigns, and that the generation bump pre-empts, so `this.runtime` is still the old aborted client at catch time. I ran an adversarial review pass over the change, it flagged cf-cell-link, and when I went to confirm or refute it against the code I found that the reasoning does not hold on the logout path. So I've fixed cf-cell-link too. If you have context I'm missing — for instance, that an in-flight resolve never actually races logout in practice — I'm very happy to revert it.

Here is the evidence I'm leaning on:

1. **RootView does not await dispose before clearing the runtime on logout.** In `RootView.ts`, the logout branch (no app or no identity) runs `previous.dispose().catch(console.error)` fire-and-forget, and then sets `this.runtime = undefined` on the very next line. There is no await between the two. The "dispose, then await, then assign" shape only describes the *runtime-swap* branch, where `await RuntimeInternals.create(...)` sits between the old dispose and the new assignment. Logout and swap behave differently here.

2. **The runtime context clears synchronously, and the generation bump is deferred, so it does not pre-empt.** cf-cell-link consumes the runtime through `@consume({ subscribe: true })`. When RootView sets `this.runtime = undefined`, @lit/context notifies subscribers inline (its `ValueNotifier.setValue` calls `updateObservers` synchronously), so cf-cell-link's own `this.runtime` becomes undefined in the same tick. The re-resolve that would bump `_resolveCellGeneration` is scheduled by Lit's `requestUpdate` on a later update microtask — which lands *after* the microtask that resumes the aborted `resolveAsCell()` and runs the catch.

3. **So at catch time, both guards fall through.** The generation still matches (the re-resolve hasn't run yet), and `this.runtime` is already undefined, so `this.runtime?.signal.aborted` evaluates to undefined and is falsy. Execution reaches `console.error("Failed to resolve cell:", e)` / `"Failed to resolve link:"` on every logout that races an in-flight resolve — which is exactly the teardown console noise this PR exists to remove.

4. **A corroborating data point from this PR itself.** cf-profile-badge has an equivalent generation guard and you still had me migrate it to read the cell's own runtime. Its `willUpdate` re-resolves only on a `profile` change, never on a runtime change, which says the team already concluded the generation guard doesn't cover the runtime-clear. cf-cell-link is the same situation.

There's also a second case beyond logout where the current cf-cell-link code is wrong and the change fixes it: a runtime swap where the component still holds a cell from the old, just-disposed runtime while the ambient `this.runtime` is now the new live one. The old guard checks the new live runtime and logs spuriously; the new guard checks the runtime the resolve actually ran on and correctly stays quiet. I worked through the other directions too (a genuine resolve failure while the runtime is alive still logs, since the op's runtime isn't aborted), and I couldn't find a case where reading the op's own runtime suppresses something you'd want to see — a dead runtime can't produce a meaningful resolve result to lose.

The fix mirrors the two siblings: the cell branch reads `cell.runtime().signal.aborted`, and the link branch reads the local `runtime` captured at the top of `_resolveCell` (which is exactly the client `linkedCell` was built from, and is in scope and non-null inside `if (link && runtime)`). cf-cell-link had no disposal test, so I added one covering both branches; the two "disposed" cases are genuine regression tests in that they log under the old guard.

## A related site I checked and intentionally left alone

I also looked hard at the `this.rt?.signal.aborted` guards in the shell (FavoriteButton, HeaderView, BodyView), since they look identical to the cf-cell-link pattern. Those are fine and I left them. The difference is that `rt` there is a plain `@property` passed down by the parent, not a `@consume` context value. A property only updates on the parent's next, deferred render, whereas the context value clears synchronously. So at catch time `this.rt` still holds the old, aborted runtime, and the guard returns correctly. The op and the guard also read the same `this.rt` in those components, which the cf-* components did not.

Everything type-checks, the full suite passes, the coverage gate is unchanged, and the touched files are clean under `deno fmt`.

mathpirate:
Thanks for pushing back on cf-cell-link — you're right, and our "leave it alone" note was wrong. We refuted it by analyzing the runtime-*swap* ordering (dispose → await create → assign, where the old aborted client stays in place across the await) and never checked the *logout* branch, which disposes fire-and-forget and clears the ambient runtime with no await. Re-derived against RootView + cf-cell-link's `willUpdate`: the generation bump is a deferred Lit update that lands after the aborted resolve's catch, and `this.runtime` is already `undefined` by then, so the old guard logged on every logout-raced resolve. Your fix (read the op's own runtime) is correct in both logout and swap, and the cf-profile-badge migration we asked for was the same situation — so our refutation was inconsistent with our own finding. Reading the op's runtime is the right invariant; nice catch.

On the dispose timeout: agreed, your call makes sense — keeping the bounded wait for flush durability is the right trade; the only thing it costs is logout responsiveness on a genuinely wedged worker, which we're fine with. Everything else looks great.

— Claude, on behalf of Gideon

PR #4227, opened 2026-06-18
refactor(runner): collapse residual { "/": … } sites onto the cell-rep chokepoint

danfuzz:
Continues the `FabricHash` / `{ "/": … }` arc by routing the remaining direct
detect/access sites of the **legacy serialized entity-ref shape** through
`cell-rep.ts`'s chokepoint (`isEntityRef` / `entityRefToString` / `entityRefFrom`).

A repo sweep first confirmed that **no production site constructs the bare
`{ "/": … }` hash shape outside the chokepoint** anymore (`cell.entityId`,
`getEntityId`, lib-shell all route through it) — the residuals are detect/access.

### The invariant that makes this clean
With the modern cell representation on, **all db contents and network comms for
content-id hashes are `FabricHash`es**, so "modern regime AND a `{ "/": … }`
object" never co-occurs. Each consumer therefore keeps its `FabricHash` branch
(the live `EntityId`/`createRef` form is a `FabricHash` in *either* regime — not a
"maybe-`/`" concern) and routes only the maybe-`{ "/": … }` part through the
chokepoint. A bare `{ "/": … }` in modern mode is cleanly **rejected** rather
than mishandled — that's the intended clean break.

### Changes
- **Storage benches** (`storage-source-topology-refresh`, `-mixed-query-refresh`,
  `-subscription-refresh`) made flag-agnostic. They hand-rolled
  `{ "/": <bare-slug> }` `source` pointers (only model legacy; bare slugs aren't
  valid `FabricHash`es). Now every synthetic id derives from `hashOf(...)` (each
  `of:` URI and the `source` targeting it share one hash → topology consistent),
  and `source` is built via `entityRefFrom` — no explicit flag-branching in the
  bench logic. A one-line env bridge (`EXPERIMENTAL_MODERN_CELL_REP`) selects the
  regime. Verified green under **both** flag states.
- **`uri-utils.toURI`** — serialized-ref branch now uses `isEntityRef` +
  `entityRefToString` (keeps the `FabricHash` branch). New `uri-utils.test.ts`.
- **`cfc/ui-contract.aliasCellId`** — `isEntityRef(cell) ? toURI(cell) : undefined`;
  dropped the now-dead try/catch. Bonus: resolves a modern `FabricHash` alias cell
  that the old object-only guard silently dropped to `undefined`.
- **`create-ref.getEntityId`** — **removed** the `value.startsWith("{")` /
  `JSON.parse(value)["/"]` branch outright (rather than consolidating it).
  *Why:* a caller trace showed it is unreachable. The three `Cell`-passing
  callers in `piece/manager.ts` go through the object/`parseLink` path; the one
  string caller passes `link.id`, typed `URI`, which `parseLink`/`toURI`/`fromURI`
  normalize to `of:…`/`data:…`/bare — never a JSON `{…}` string. The branch was a
  vestige of an era when ids were passed around JSON-serialized. Deleting it
  removes the last direct `["/"]` access here instead of carrying dead code that
  reaches into the legacy shape. Added coverage for the live string forms (bare
  id, `of:` URI) across both regimes — previously untested.

### Verification
- `deno check` / `lint` / `fmt` clean on all touched files.
- Full runner suite green after each change (final: **636 passed, 0 failed**).
- New tests exercise both flag states directly.

### Not included (follow-ups)
- `create-ref.ts:78` — the `createRef`/`traverse` "already a serialized id" guard.
  Delicate: the object's shape is itself a hash input, so any change must stay
  byte-equivalent at the `createRef`/`hashOf` boundary. Deserves its own PR.
- `patterns/notes/note.tsx:296` — `resolved?.entityId?.["/"]`, a pattern-layer
  consumer that hardcodes the legacy shape; separate pace layer.

### DIFF
--- a/packages/runner/src/create-ref.ts
+++ b/packages/runner/src/create-ref.ts
@@ -4,8 +4,6 @@ import {
   type EntityRef,
   entityRefFrom,
   entityRefFromString,
-  entityRefToString,
-  isEntityRef,
 } from "@commonfabric/data-model/cell-rep";
 import { isRecord } from "@commonfabric/utils/types";
 import { isOpaqueRef } from "./builder/types.ts";
@@ -139,15 +137,6 @@ export function getEntityId(value: any): EntityRef | undefined {
   if (typeof value === "string") {
     // Handle URI format with "of:" prefix
     if (value.startsWith("of:")) value = fromURI(value);
-    if (value.startsWith("{")) {
-      // A JSON-serialized serialized entity-ref; recognize and extract its
-      // tagged-hash string through the cell-rep chokepoint. (Only the legacy
-      // `{ "/": … }` object serializes this way; a modern `FabricHash` does
-      // not survive JSON as a ref, hence the `undefined` fall-through.)
-      const parsed = JSON.parse(value);
-      if (!isEntityRef(parsed)) return undefined;
-      value = entityRefToString(parsed);
-    }
     return entityRefFromString(value);
   }
 
--- a/packages/runner/src/create-ref.ts
+++ b/packages/runner/src/create-ref.ts
@@ -4,6 +4,8 @@ import {
   type EntityRef,
   entityRefFrom,
   entityRefFromString,
+  entityRefToString,
+  isEntityRef,
 } from "@commonfabric/data-model/cell-rep";
 import { isRecord } from "@commonfabric/utils/types";
 import { isOpaqueRef } from "./builder/types.ts";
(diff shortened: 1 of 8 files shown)


packages/runner/src/create-ref.ts
```diff
@@ -137,9 +137,7 @@ export function getEntityId(value: any): EntityRef | undefined {
   if (typeof value === "string") {
     // Handle URI format with "of:" prefix
     if (value.startsWith("of:")) value = fromURI(value);
-    return value.startsWith("{")
-      ? entityRefFromString(JSON.parse(value)["/"])
-      : entityRefFromString(value);
+    return entityRefFromString(value);
```
ubik2, line 140: I think getEntityId is exposed through the public api, so this is sort of a breaking change, but I think this is also fine (and probably we shouldn't expose it anyhow)

PR #4240, opened 2026-06-18
feat(ui): UI variants — cf-render full/chip/tile with failover defaults (CT-1321)

bfollington:
## What

Formalizes the UI-variant system (CT-1321). A piece can expose a **size spectrum** of renderings as optional sibling output keys, addressed by symbols vended from `commonfabric`, replacing the vestigial half-built mechanism in `cf-render`.

| variant | symbol | use | default if unset |
| --- | --- | --- | --- |
| `full` | `UI` (`$UI`) | standalone (today's default) | the `[UI]` itself — universal floor |
| `chip` | `CHIP_UI` (`$CHIP_UI`) | inline in text/lists | `cf-cell-link` (renders by `[NAME]`) |
| `tile` | `TILE_UI` (`$TILE_UI`) | gallery/grid card | the full `[UI]` at `scale(0.5)` |

**Failover = per-variant default, full as the floor.** A piece exporting only `[UI]` still renders at every variant.

## Changes

- Vend `TILE_UI`/`CHIP_UI` alongside `UI` (builder/api/factory/jsx). Transparent passthrough — no transformer or schema-generator changes.
- Rebuild `cf-render` with real `variant="full|chip|tile"` resolution + defaults; delete the dead `UIVariant` enum / `_VARIANT_TO_KEY`.
- Sync the stale `cf-render` `variant` JSX typing; migrate `cf-picker` `preview` → `tile`.
- Catalog story (**Render (UI Variants)**), example patterns (`ui-variants-demo`, `ui-variants-host`), and COMPONENTS.md docs.

## Verification

- `deno check` + `deno lint` + `deno fmt --check` clean across the changeset; `cf-render` unit tests pass; demo pattern compiles with the right `$UI`/`$CHIP_UI`/`$TILE_UI` schema.
- **Browser E2E** (deployed host piece, 0 console errors): exported chip/tile/full **and** the chip/tile failover defaults all render correctly.

## Out of scope / follow-up

- `sidebarUI`/`fabUI`/`settingsUI` are shell composition **slots**, a separate concept — untouched.
- **Phase B:** a vended `uiVariant()` helper for render paths outside `cf-render`. Investigation pinned it must be a runner **builtin** (the defaults need `h`/`getAsLink` cross-boundary cell-binding, which lives above the builder layer). Tracked for a follow-up PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
[...]

PR #4242, opened 2026-06-19
feat(ui): adopt UI variants across product surfaces; remove old-way rendering (CT-1764)

bfollington:
Implements [CT-1764](https://linear.app/common-tools/issue/CT-1764) — adopt the CT-1321 `<cf-render variant="tile|chip">` system across the surfaces that display pieces, and remove the old ad-hoc rendering. Follow-up to #4240.

## Done
**Tile (grid/card):**
- `piece-grid.tsx` — hand-rolled `scale(0.4)` preview + raw `cf-cell-link` → `variant="tile"` + `variant="chip"` label.
- `profile-home.tsx` — pinned patterns → a **tile gallery** (clickable/navigating), user's title/tags as footer; bespoke card + "Open" link gone.
- `record.tsx` ×3 — `embeddedUI ?? <cf-render>` → `variant={isExpanded ? "full" : "tile"}` (card shows the tile, expanded shows full).

**Producer rename:** `note.tsx` / `note-md.tsx` `embeddedUI` → `[TILE_UI]` (the curated minimal UI is the note's tile). `embeddedUI` is now fully removed from the repo.

**Chip (lists/rows — piece-as-itself displays):** `default-app.tsx` + `default-app-ben.tsx` recent/all-pieces lists, `profile-picker`, `backlinks-index` (main), `summary-index`, `activity-log`.

**`cf-render` chip made inline** — reflect `variant`, inline-block host styling for chip, skip the full-size spinner for chip. (Chip was a full-width block before — unusable inline.) Browser-verified as inline pills.

## Principle applied
`cf-cell-link` stays for **navigation links with custom labels/chrome** (toolbar "Mentions"/"Search", "view as grid" toggles) and **styled sub-chips** (backlink chips with custom font/color). The variant system replaces **piece-as-itself** displays.

## Deferred (separate sweep — coupled to a producer rename)
`email-pattern-dreamer`/`launcher` read `previewUI` from ~15 google/airtable/tools extractor patterns. Migrating those consumers to `variant="tile"` requires renaming those producers' `previewUI` → `[TILE_UI]` first. Left for a focused follow-up.

## Verification
All changed files `cf check` + `deno lint` clean. `cf-render` chip-inline browser-verified. Full per-surface browser pass pending (deployed-home surfaces).

### DIFF
--- a/packages/patterns/activity-log/activity-log.tsx
+++ b/packages/patterns/activity-log/activity-log.tsx
@@ -178,7 +178,7 @@ export default pattern<ActivityLogInput, ActivityLogOutput>(
                         )
                         : null}
                       {event.pieceRef
-                        ? <cf-cell-link $cell={event.pieceRef} />
+                        ? <cf-render variant="chip" $cell={event.pieceRef} />
                         : null}
                     </cf-hstack>
                     {event.note
--- a/packages/patterns/notes/note-md.tsx
+++ b/packages/patterns/notes/note-md.tsx
@@ -7,6 +7,7 @@ import {
   navigateTo,
   pattern,
   type Stream,
+  TILE_UI,
   UI,
   type VNode,
   wish,
@@ -35,8 +36,8 @@ export interface NoteMdOutput {
   isHidden: true;
   /** Excluded from mentions autocomplete (notes in notebooks may be hidden but still mentionable) */
   isMentionable: false;
-  /** Minimal UI for embedding in other patterns */
-  embeddedUI: VNode;
+  /** Tile variant (CT-1764): minimal embedded UI for `<cf-render variant="tile">`. */
+  [TILE_UI]: VNode;
   /** Processed content with wiki-links converted to markdown links */
   processedContent: string;
   /** Stream to toggle checkboxes in content */
@@ -203,7 +204,7 @@ export default pattern<NoteMdInput, NoteMdOutput>(
       note,
       isHidden: true,
       isMentionable: false,
-      embeddedUI: markdownViewer,
+      [TILE_UI]: markdownViewer,
       processedContent,
       checkboxToggle: handleCheckboxToggle,
       hasBacklinks,
--- a/packages/patterns/notes/note.tsx
+++ b/packages/patterns/notes/note.tsx
@@ -14,6 +14,7 @@ import {
   type PatternToolResult,
   SELF,
   type Stream,
+  TILE_UI,
   toCompactDebugString,
   UI,
   type VNode,
@@ -52,8 +53,9 @@ export interface NoteOutput extends NotePiece {
   createNewNote: Stream<void>;
   /** Parent notebook reference, null if not in a notebook */
   parentNotebook: NotebookPiece | null;
-  /** Minimal UI for embedding in containers like Record (read directly as `.embeddedUI`). */
-  embeddedUI: VNode;
+  /** Tile variant (CT-1764): the minimal embedded UI used when a container
+   * (e.g. Record) renders this note via `<cf-render variant="tile">`. */
+  [TILE_UI]: VNode;
   // Test-accessible state
   menuOpen: boolean;
   isEditingTitle: boolean;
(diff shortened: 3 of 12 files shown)


bfollington:
🤖 **cf-review** (Claude Opus 4.8, on @bfollington's behalf)

cf-review via Claude Opus 4.8 — *scope:* PR #4242 (CT-1764, adopt `<cf-render variant>` across product surfaces, remove old way) · *coverage:* deep-read `cf-render.ts` + full diff, sampled all 11 pattern files · *checks run:* `deno check cf-render.ts` (clean), `deno test cf-render.test.ts` (5 pass), grep sweeps for `embeddedUI` + residual `cf-cell-link`.

**Verdict: ship.** Clean, well-scoped adoption. The load-bearing fix (resolve the link before render/navigate) is correct and faithfully mirrors `cf-cell-link`; the rename is complete; the "cf-cell-link for labeled-nav vs cf-render for piece-as-itself" principle is applied consistently. No blocking findings.

---

**1. The new async render path is unguarded by automated tests.** `packages/ui/src/v2/components/cf-render/cf-render.test.ts`
The three new behaviors — `resolveAsCell()` before render/navigate, the `_renderingCellId` stale-guard across the new awaits, and the chip spinner-skip — are exactly the lifecycle logic the test file's own header says it *can't* cover in the headless runner. The added tests (`normalizeVariant`, `hasVariantValue`, property/disconnect) are good but orthogonal to this PR's actual risk. So the fix rests entirely on the browser verification noted in the commit. *Suggested follow-up:* a browser/integration test asserting a link-wrapped piece tile (a) renders `[TILE_UI]` and (b) navigates to the resolved root cell on click — that's the exact regression this PR fixes, and it's the one with no guard. (suspected — not a defect, a coverage gap)

**2. `_resolvedCell` is not reset on disconnect.** `cf-render.ts:397` (`disconnectedCallback`)
It's reset on cell-change (`:216`) but `disconnectedCallback` clears `_renderingCellId`/`_hasRendered` and leaves `_resolvedCell` set. Harmless today — navigation only fires from a click on a connected tile, and a reconnect re-renders — but for symmetry with the other reset state it's worth nulling here too. (verified — benign)

---

### ⚪ Nits

- `cf-render.ts:355` — `_navigateToPiece` falls back to `this.cell` when `_resolvedCell` is unset, which for a link would navigate to the *link* cell (the bug this PR fixes). In practice the click path can't run before `_resolvedCell` is populated (the tile default is only mounted after resolution at `:293`), so the fallback is dead-but-defensible. Fine as-is; just noting the fallback re-opens the resolved/unresolved distinction.
- Stale, momentarily-empty chip: with the spinner skipped, a chip shows an empty inline box until `resolveAsCell()`+`sync()` resolve. Correct call (a full-size spinner would reserve the wrong space) — flagging only so it's a conscious trade.

---

### Confirmed during review
- `embeddedUI` is **fully removed** — `grep -rn embeddedUI packages/` returns 0 hits; Note/NoteMD producers + all 3 `record.tsx` consumers migrated to `[TILE_UI]`. No reader left behind.
- **Principle applied consistently.** Every surviving `cf-cell-link` in the touched files is a labeled-nav link or a styled sub-chip: toolbar grid/list toggles (`default-app`/`-ben`), backlink sub-chips (`backlinks-index:143`), and the owner **edit-mode** pinned list (`profile-home:709`, rows with Remove buttons) — distinct from the visitor-facing tile gallery that *was* migrated. No piece-as-itself display left on the old way.
- **`resolveAsCell` parity is real.** `cf-render` now resolves exactly as `cf-cell-link._resolveCell` does (`cf-cell-link.ts:188`), and its `_renderingCellId !== cellId` stale-guard (re-checked after each await) is the structural equivalent of cf-cell-link's `_resolveCellGeneration` counter — correct across both cell-swap and disconnect.
- **`record.tsx` behavior change is coherent.** Expanded → `full` (== the old un-variant'd `<cf-render>`, unchanged). Collapsed → `tile`: for Notes `[TILE_UI]` is the renamed `embeddedUI` (equivalent); for other pieces, collapsed now shows a scaled, click-to-navigate preview instead of an un-scaled full `[UI]` — the intended improvement.

*Review only — nothing was modified.*

PR #4245, opened 2026-06-19
fix(runtime-client): last-write-wins UI writes — blind-vs-CAS by method (set/push)

mathpirate:
## Why

A `$value`/`$checked` UI edit committed as a **compare-and-set**: `normalizeAndDiff`
reads the write target to diff it, and that read became a commit precondition.
Under concurrent same-user edits — two tabs/sessions sharing one `PerUser` cell —
the read lost the own-write race (`stale confirmed read … conflicted`), so the
write was **rejected and rolled back and the edit silently vanished**. This is the
`cfc-group-chat-demo` "profile name stays 'Name not set'" flake.

## What this does

A UI `set` is a **blind last-write-wins** leaf write; a `push` is read-modify-write
and keeps **compare-and-set**. The choice is made by **METHOD** (the request type
the client sends), not by the value's shape:

- **`CellHandle.set` → `CellSet` → `handleCellSet`:** always blind (any value
  type). The set's reads carry no value-equality precondition, so a concurrent
  same-user edit no longer loses the own-write race. In their place we keep **one
  structural precondition** — a `nonRecursive` read at the **cell's parent**
  (threaded from `handleCellSet`, where the logical write path is known;
  `buildReads` only sees the optimized diff). That catches a concurrent whole-doc
  delete **or an ancestor reshape** as a clean `ConflictError`, rather than letting
  a stale nested patch throw a raw "not traversable" at commit-materialization. A
  write to the cell's own value (a sibling leaf, or an array element) sits below
  the parent and does not conflict.
- **`CellHandle.push` → new `CellPush` → `handleCellPush`:** read-modify-write; the
  read of the current value stays a commit precondition, so a concurrent push
  **aborts** rather than being clobbered.

Both handlers share `applyCellWrite(request, blind)`. Making the decision explicit
at the protocol level (rather than inferring "structured ⇒ RMW" from the value)
lets an array *overwrite* be blind while an array *push* captures reads. CFC
enforcement (rides `attemptedWrites`) and reactivity/subscriptions (ride
`ignoreReadForScheduling`) are untouched; the blind mark clears before
`prepareTxForCommit` so CFC boundary-commit read-then-writes keep their
preconditions.

Supersedes the recovery-based approaches to this flake (#4196 commit-queue, #4126
reapply-latest) by removing the conflict at the source.

## Tests

- `packages/patterns/integration/cellset-lww.test.ts` (multi-runtime, same-identity
  two-session own-write race): concurrent scalar sets never conflict; a
  **structured (array) set is blind too** (the trigger is the method, not the value
  type); concurrent **pushes keep compare-and-set**; and a typed name survives the
  race end-to-end through save.
- `packages/memory/test/cellset-structural-precondition.test.ts` (engine-level):
  a precondition-free nested patch on a retyped ancestor throws a raw "not
  traversable"; an **entity-root** structural read does *not* catch it (still a raw
  throw); the **parent** structural read converts it to a clean `ConflictError`.
  (The in-process harness can't reproduce this end-to-end — it propagates shared
  state synchronously, so a session can't hold a stale-but-navigable replica —
  hence the engine-level test.)

Regression: conflict/commit/runtime-processor/precondition suites + a broader
conflict sweep green (109 tests); fmt/lint/type-check clean.

## Behavior change

A `set` of an **array/object** value is now blind (last-write-wins), where the
earlier value-type form kept it on compare-and-set. Concurrency-safe list
mutations go through `push` (CAS), or the deferred server-side mergeable-op path.
This is the one place the method-based trigger changes prior behavior.

## Design notes for review (@ubik2 — your area)

This is the shape from your review: decide blind-vs-CAS by the **method** (a
`CellPush` request type) rather than the value-type heuristic, and keep a
**structural** precondition — anchored at the cell's parent so an ancestor reshape
is caught (the `["notes"]`-retyped case, à la #4406), not just a whole-doc delete.
Push stays CAS-retry at the browser→worker seam; a mergeable `append` is left as a
deferred, larger change.

## Coverage-debt override (packages/runner)

The Performance Check flags `packages/runner` coverage-debt at +2, but this is a
cross-file attribution artifact, **not** new untested logic. Verified by
replicating the metric over the CI coverage artifacts (reproduces the gate's
7852 → 7854 exactly), deterministically across 5 runs:

- This PR's changed source adds **zero** uncovered lines — `storage/v2.ts` (254
  uncovered) and `storage/reactivity-log.ts` (7) are identical on this branch and
  `main`.
- The +2 is: `data-updating.ts` **−10** (the new blind-write test legitimately
  covers more of the diff path) **+** `sandbox/compiled-bundle-verifier.ts`
  **+12** — a bracket-quoted path-parser branch (lines 1632–1653) covered only by
  the generated-patterns integration job, which this PR's *runtime* change stops
  reaching. All generated patterns still pass; that branch is an existing utility,
  not new code here.

Accepting the metric per CI_PERFORMANCE.md's narrow per-metric form:

NEW_PERF_BASELINE: coverage-debt: packages/runner uncovered lines = 7854 lines

### DIFF
--- a/packages/runtime-client/backends/runtime-processor.ts
+++ b/packages/runtime-client/backends/runtime-processor.ts
@@ -22,11 +22,13 @@ import {
   getPatternIdentityRef,
   isCell,
   isCellResult,
+  markUiInputBlindWriteTx,
   Runtime,
   RuntimeTelemetry,
   RuntimeTelemetryEvent,
   setPatternEnvironment,
   type SigilLink,
+  unmarkUiInputBlindWriteTx,
 } from "@commonfabric/runner";
 import { linkRefPayload } from "@commonfabric/runner/shared";
 import {
@@ -689,7 +691,21 @@ export class RuntimeProcessor {
     const tx = this.runtime.edit();
     const cell = getCell(this.runtime, request.cell);
     const value = mapCellRefsToSigilLinks(request.value);
+    // A scalar `$value`/`$checked` overwrite (string/number/boolean) is a
+    // last-write-wins leaf write: mark the tx so the set's reads carry no
+    // concurrency precondition, removing the own-write-race "stale confirmed
+    // read" conflict (and the rolled-back, silently-dropped edit it caused).
+    // Structured (array/object) values are excluded — they are often
+    // read-modify-write (CellHandle.push, multi-select, list edits) and must
+    // keep compare-and-set to avoid lost updates. The mark is cleared before
+    // prepareTxForCommit so CFC boundary-commit read-then-writes retain their
+    // preconditions.
+    const raw = request.value;
+    const blindLeafWrite = typeof raw === "string" ||
+      typeof raw === "number" || typeof raw === "boolean";
+    if (blindLeafWrite) markUiInputBlindWriteTx(tx);
     cell.withTx(tx).set(value);
+    if (blindLeafWrite) unmarkUiInputBlindWriteTx(tx);
     this.runtime.prepareTxForCommit(tx);
     // Local visibility is established by commit(); the promise tracks remote
     // confirmation/rollback and must not block cell IPC.
--- a/packages/runtime-client/backends/runtime-processor.ts
+++ b/packages/runtime-client/backends/runtime-processor.ts
@@ -693,8 +693,10 @@ export class RuntimeProcessor {
     const value = mapCellRefsToSigilLinks(request.value);
     // A scalar `$value`/`$checked` overwrite (string/number/boolean) is a
     // last-write-wins leaf write: mark the tx so the set's reads carry no
-    // concurrency precondition, removing the own-write-race "stale confirmed
-    // read" conflict (and the rolled-back, silently-dropped edit it caused).
+    // value-equality precondition on the leaf (only a structural entity-root
+    // existence read survives, to catch a concurrent whole-doc delete), removing
+    // the own-write-race "stale confirmed read" conflict (and the rolled-back,
+    // silently-dropped edit it caused).
     // Structured (array/object) values are excluded — they are often
     // read-modify-write (CellHandle.push, multi-select, list edits) and must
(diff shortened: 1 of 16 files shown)


packages/runtime-client/backends/runtime-processor.ts
```diff
@@ -686,27 +688,53 @@ export class RuntimeProcessor {
     };
   }
 
+  // A `CellHandle.set` is a blind leaf overwrite (last-write-wins); a
+  // `CellHandle.push` is read-modify-write and keeps compare-and-set. The
+  // blind-vs-CAS decision is made by METHOD — which request type the client sent
+  // — not by inspecting the value's shape. Both carry the whole already-resolved
+  // value on the wire.
   handleCellSet(request: CellSetRequest): void {
+    this.applyCellWrite(request, /* blind */ true);
```
ubik2, line 698: This makes me wonder if we should just incorporate the blind flag in the CellSetRequest.
That wasn't what I was originally thinking, but it should make things a bit simpler, and easier to extend if we wanted other methods to use this.

In that mode, either `#applyLocalAndSend` would take a `blind` flag, which it would include in the CellSetRequest, or that might even be added as an optional flag to `set` (with default being `true`).

It's not quite as good if we want to later plumb things like `push` all the way through, so it shows up as a patch op `append`, but we're a long way from being there.

ubik2:
The handling of the blind write parent dependency is a bit more conservative than I expected, but perhaps that's better.
A write to `["foo", "bar", "cab"]` after a write to `["foo", "bar"]` fails even when the value at `["foo", "bar']` is still an object, so it's still possible to apply the write without clobbering.
It's probably best to try this out to see if it's worthwhile to make it more liberal before we add that more complex check.

ubik2 (approved):
I added some comments, and perhaps it's worth waiting for Ben's feedback, but this lgtm.
It could be merged as-is, or that blind flag on set thing could be added. Either one is good.

PR #4253, opened 2026-06-19
docs(traversal): mark fabric-unsafe recursive value-walk sites with TODO(danfuzz)

danfuzz:
## Summary

Marks every recursive value-walk site that descends a value generically
(`isRecord` / `Object.entries`) **without a `FabricSpecialObject` guard first**,
so a `FabricPrimitive` gets decomposed (its state lives in private fields, zero
own-enumerable props) and a `FabricInstance` gets walked by its internal slots
rather than by its codec contents. Each site gets a `TODO(danfuzz)` comment
describing the specific gap.

**Markers only — zero behavior change.** The diff is **+162/-0**, all
comment/doc lines; no code path is altered. The actual fixes are tracked
separately.

This is the output of the recursive-traversal audit of the fabric data model.

## Marker groups

The marked sites fall into three groups, in two phrasings:

- **Broken today** — sites that mishandle a fabric value reachable on a current
  code path. Includes the `SchemaObjectTraverser` engine dispatch in
  `traverse.ts`, `isDeepFrozenInProgress` in `deep-freeze.ts`, and `deepEqual`
  itself in `utils/deep-equal.ts` (compares class instances by enumerable
  own-props, so two same-class `FabricPrimitive`s compare equal regardless of
  value).
- **Fabric-carrying `deepEqual` callers** — call sites that pass
  fabric-bearing values into the generic `deepEqual`, inheriting its gap.
- **Latent / future-reachable** — schema-gated or not-yet-reached sites that
  would exhibit the same gap once their path carries a fabric value.

## Marked files

39 `TODO(danfuzz)` markers across 20 files in 5 packages:

- **`packages/runner`** (33 markers, 16 files): `storage/v2-transaction.ts` (9),
  `builtins/llm.ts` (5), `builtins/llm-dialog.ts` (5),
  `reactive-dependencies.ts` (3), `runner.ts` (2), `pattern-binding.ts` (2),
  `data-updating.ts` (2), `cell.ts` (2), `traverse.ts` (1), `link-utils.ts` (1),
  `cfc/prepare.ts` (1), `cfc/schema-sanitization.ts` (1),
  `cfc/structured-result.ts` (1), `scheduler/diagnosis.ts` (1),
  `scheduler/write-propagation.ts` (1), `builder/traverse-utils.ts` (1)
- **`packages/piece`** (2): `src/manager.ts`
- **`packages/utils`** (1): `src/deep-equal.ts`
- **`packages/data-model`** (1): `src/deep-freeze.ts`
[...]

PR #4255, opened 2026-06-19
fix(data-model): arm isDeepFrozenInProgress for FabricInstance private slots

danfuzz:
## Summary

Fixes a frozen-status bug in `isDeepFrozenInProgress` (`data-model`): a
deep-frozen check on a `FabricInstance` walked its enumerable own-props instead
of its logical contents, so a frozen `FabricInstance` whose **non-enumerable
slots** held unfrozen values was wrongly reported deep-frozen.

This is the audit's §3.1.3 finding (see
`coordination/docs/2026-06-18-recursive-traversal-fabric-audit.md`). The fix
also removes the `TODO(danfuzz)` marker that #4253 placed at this site — the
mark is now resolved by construction.

## The bug

`isDeepFrozenInProgress` had no `instanceof FabricInstance` arm, unlike its two
siblings `deepFreezeInProgress` and `isDeepFrozenFabricValue`. It fell through
to the generic `Object.values(obj)` walk, which can't see a `FabricInstance`'s
logical contents held in private slots — e.g. `FabricError` keeps its custom
properties in a private extras `Map`. A frozen `FabricError` whose extras bag
held a mutable value would report `true` (deep-frozen) when it is not. This is
reachable via `FabricError.deepClone`'s `if (frozen && isDeepFrozen(this))
return this;` fast-path, which would then wrongly skip the clone.

## The fix

Add the lockstep `obj instanceof FabricInstance` arm that delegates to the
instance's `[IS_DEEP_FROZEN]` protocol member (the side-effect-free sibling of
`[DEEP_FREEZE]`), recursing into each nested `FabricValue` through the callback
so this call's `inProgress` cycle state is shared. Gating on `instanceof`
against the abstract base keeps it generic — `[IS_DEEP_FROZEN]` is abstract on
`FabricInstance`, so every instance implements it. (A `FabricPrimitive` is
necessarily frozen with no outbound references, so the existing `Object.values`
arm already answers it correctly.)

## The test

A red-green regression test: a frozen `FabricError` whose private extras `Map`
holds a mutable array — `isDeepFrozen` returned `true` before, returns `false`
now. RED-confirmed on `ebdf920b5`, GREEN after the fix.

## Scope

2-file diff (`deep-freeze.ts` + `deep-freeze.test.ts`), +34/-12. Targeted fix +
regression test; no other behavior change.

https://claude.ai/code/session_01KNv3rzpEfRtwZfzUy2edoT

Co-Authored-By: coder-cedar (Claude Opus 4.8 (1M context)) <noreply@anthropic.com>

---
## Summary by cubic
Fixes deep-freeze checks in `data-model` so `FabricInstance` inspects its private slots, avoiding false “deep-frozen” results. Prevents `FabricError.deepClone` from skipping clones when internals aren’t fully frozen.

- **Bug Fixes**
  - Added `obj instanceof FabricInstance` branch in `isDeepFrozenInProgress` delegating to `[IS_DEEP_FROZEN]`, sharing the `inProgress` cycle state.
  - Now inspects non-enumerable slots (e.g., `FabricError` extras `Map`), not just enumerable props.
  - Added regression test: a frozen `FabricError` with a mutable extras array now returns `false`.

<sup>Written for commit c894aee6a0325f25c31f0d879b63ff24fc0e59a6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4255?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/data-model/src/deep-freeze.ts
+++ b/packages/data-model/src/deep-freeze.ts
@@ -94,7 +94,19 @@ function isDeepFrozenInProgress(
 
   let result = true;
 
-  if (Array.isArray(obj)) {
+  if (obj instanceof FabricInstance) {
+    // A `FabricInstance`'s logical contents are not its enumerable own-props
+    // (e.g. a `FabricError` keeps its custom properties in a private extras
+    // `Map`), so it answers the deep-frozen question via its `[IS_DEEP_FROZEN]`
+    // protocol member -- the side-effect-free sibling of `[DEEP_FREEZE]` --
+    // recursing into each nested `FabricValue` through the callback so this
+    // call's cycle state is shared. Gating on `instanceof` against the abstract
+    // base keeps this generic; the member is abstract on `FabricInstance`, so
+    // every instance implements it. (A `FabricPrimitive` is necessarily frozen
+    // with no outbound references, so the `Object.values` arm below answers it
+    // correctly by accident -- its empty enumerable props yield `true`.)
+    result = obj[IS_DEEP_FROZEN]((v) => isDeepFrozenInProgress(v, inProgress));
+  } else if (Array.isArray(obj)) {
     for (let i = 0; i < obj.length; i++) {
       if (!(i in obj)) continue; // sparse hole
       if (!isDeepFrozenInProgress(obj[i], inProgress)) {
@@ -103,12 +115,6 @@ function isDeepFrozenInProgress(
       }
     }
   } else {
-    // TODO(danfuzz): Unlike its siblings `deepFreezeInProgress` and
-    // `isDeepFrozenFabricValue`, this checker has no `instanceof FabricInstance`
-    // arm delegating to `[IS_DEEP_FROZEN]`; a frozen-in-place `FabricInstance`
-    // is checked by its native/internal own-props instead of its logical
-    // contents, so it can return a wrong frozen-status (reached via
-    // `FabricError.deepClone`'s `isDeepFrozen(this)` fast path).
     for (const v of Object.values(obj)) {
       if (!isDeepFrozenInProgress(v, inProgress)) {
         result = false;
--- a/packages/data-model/test/deep-freeze.test.ts
+++ b/packages/data-model/test/deep-freeze.test.ts
@@ -205,11 +205,13 @@ describe("deep-freeze", () => {
     // Coverage for `isDeepFrozen` on `FabricInstance` and `FabricPrimitive`
     // inputs, including a `FabricInstance` participating in a circular
     // reference. `isDeepFrozen` delegates to `isDeepFrozenInProgress` (which
-    // threads `inProgress: Set<object>` for cycle-safety) and treats a
-    // `FabricInstance` as a plain JS object whose enumerable values are
-    // recursively inspected; this is distinct from `isDeepFrozenFabricValue`'s
-    // protocol-aware type-guard semantics, which has its own coverage in the
-    // sibling `isDeepFrozenFabricValue with FabricInstance` describe below.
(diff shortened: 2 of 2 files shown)


danfuzz:
**FULL APPROVE** — reviewed at `headRefOid` `c894aee6a` (the bundle SHA this verdict anchors to), diff-vs-`origin/main` `ebdf920b5` (clean cut; merge-base == main). Real-code fix: `deep-freeze.ts` + `deep-freeze.test.ts`, +34/−12, one function (`isDeepFrozenInProgress`) and its test.

Five-axis review, with an at-source claim-trace as the spine:

- **Fix correctness.** The new first arm — `result = obj[IS_DEEP_FROZEN]((v) => isDeepFrozenInProgress(v, inProgress))` — matches the protocol member's signature exactly (`[IS_DEEP_FROZEN](subIsDeepFrozen: (value: FabricValue) => boolean): boolean`), threads the shared `inProgress` cycle set through the callback closure, and gates on `instanceof FabricInstance` against the abstract base (generic dispatch, no concrete-subclass enumeration). `FabricInstance` / `IS_DEEP_FROZEN` are both imported.
- **Lockstep with the two sibling functions.** The arm mirrors `isDeepFrozenFabricValue`'s `checkValue` arm (`item[IS_DEEP_FROZEN]((v) => checkValue(v))`) and `deepFreezeInProgress`'s `[DEEP_FREEZE]` arm (`value[DEEP_FREEZE]((v) => deepFreezeInProgress(v, inProgress))`) — same shape, same cycle-state threading. All three verified at-source.
- **The red-green test genuinely witnesses the bug.** Pre-fix, `isDeepFrozenInProgress` had no `FabricInstance` arm, so a frozen `FabricError` fell to the `Object.values` arm — which sees only the public Error own-props (all frozen primitives) and never the private `#extras` `Map`, wrongly yielding `true`. The new test (frozen `FabricError` with a mutable array in `setExtra("payload", …)`) asserts `false` → red before, green after. `FabricError.fromNativeError` / `setExtra` / `#extras: Map` confirmed at-source. `it`-count 47 → 48 (+1); no existing test disturbed. (The adjacent "unfrozen cause" test covers the *enumerable*-slot path; this one covers the *non-enumerable* `#extras` path — both halves of the instance-contents contract.)
- **In-vicinity + stale-marker discipline.** The now-superseded `TODO(danfuzz)` on that exact site is correctly removed (the fix resolves precisely what it flagged; leaving it would contradict the new arm), and the stale test describe-comment — which documented the old "treats a `FabricInstance` as a plain JS object whose enumerable values are inspected" behavior — is updated to the protocol-aware dispatch. Sweep found no other stale framing.
- **Scope + hygiene.** Held to one function + its test; the three-sibling shared-helper refactor is correctly *not* done (deferred as a possible follow-up); no `deepEqual` touch. No task numbers or process framing in the diff. The one `isDeepFrozen` consumer that motivated the finding (`FabricError.deepClone`'s `if (frozen && isDeepFrozen(this)) return this;` fast-path) is correctness-*improved* by this fix, not regressed.

`FabricPrimitive` inputs remain correctly handled by the existing `Object.values` arm (necessarily frozen, empty enumerable own-props → `true`); the change is purely additive for every non-`FabricInstance` input.

Green CI is the remaining merge gate.

— reviewer-flux (they/them)

PR #4265, opened 2026-06-19
fix(runner,data-model): make FabricPrimitive values updatable in place (CT-1770)

mathpirate:
# Why

Updating a `FabricBytes` (and any other `FabricPrimitive`: `FabricRegExp`,
`FabricEpochNsec`/`Days`, `FabricHash`) value stored in a cell is silently a
**no-op**. Both keyed and whole-object `set()` are diffed away, the stored value
never changes, and reactive consumers never re-run.

```ts
const c = runtime.getCell<{ bytes: Uint8Array }>(space, "p", undefined, tx);
c.set({ bytes: new Uint8Array([1, 2, 3, 4]) });
await tx.commit(); tx = runtime.edit();

c.withTx(tx).key("bytes").set(new Uint8Array([9, 8, 7, 6]));
await tx.commit();
c.get().bytes.slice();   // → [1, 2, 3, 4]   (the write was dropped)
```

A plain value (a number) at a sibling key updates and re-triggers consumers
correctly in the same harness, so this is **primitive-specific** and
**schema-independent**. Fixes [CT-1770](https://linear.app/common-tools/issue/CT-1770).

# Root cause

The diff layer is innocent — `normalizeAndDiff` correctly emits a change for a
`FabricBytes` set (its `FabricSpecialObject` branch is unconditional). The value
is lost **downstream**, and there are **two distinct failure mechanisms**, both
rooted in code that treats a `FabricPrimitive` as if it were a plain object:

**1. Equality conflation.** `FabricPrimitive`s keep their state in private
`#fields`, so they have **zero enumerable own-properties**. `deepEqual` (and
therefore `valueEqual`, which today just forwards to it) compares class
instances by enumerable own-props, so **any two same-class `FabricPrimitive`s
compare equal regardless of value**:

```
deepEqual(FabricBytes([1,2,3,4]), FabricBytes([9,8,7,6])) === true   // the bug
hashOf(...) of those two differ                          === true   // hashing is fine
```

Every storage no-op gate and reactivity change-detection gate that compares
stored `FabricValue`s via `deepEqual` therefore sees "unchanged":

- Write-time no-op — `v2-transaction.ts` `writeWithinBranch` (returns before the
  write is even drafted).
- Commit-time no-op — `v2-transaction.ts` (skips committing the doc).
- Write propagation — `write-propagation.ts` (the write isn't classified as a
  change, so subscribers aren't scheduled).
- Reactive-dependency change detection — `reactive-dependencies.ts`.
- Plus the chronicle commit path and the shared mutable-path-write helper.

**2. Value-walk decomposition.** `differential.ts`'s `collectChangedPaths`
(which produces the per-path change set that drives reactivity) recurses into
records by walking their `Object.keys`. A `FabricBytes` is `typeof === "object"`
but has no own keys, so the walk finds nothing to recurse into and returns
**without recording a change** — even once equality (1) is fixed. This is a
distinct bug class from (1): the value is being *decomposed*, not *compared*.

Both mechanisms must be fixed for the end-to-end symptom (write lands **and**
reactive consumers re-fire) to go away — verified independently with a
plain-number control that re-fires while the `FabricBytes` sibling does not.

# Relationship to #4253 (Dan's lead)

[#4253](https://github.com/commontoolsinc/labs/pull/4253) merged the morning
this was filed. It is **markers-only** (162 insertions, 0 deletions, all
comments — "no behavior change") — a systematic catalog of the fabric-unsafe
sites, explicitly naming "`deepEqual` itself" and "the fabric-carrying
`deepEqual` callers (transaction change-detection, reactive-dependencies,
write-propagation, diagnosis…), pointing at the migration to a Fabric-aware
equality."

So CT-1770 is **not a regression from #4253** — it's a pre-existing latent bug
#4253 annotated (latent because storing/updating a `FabricBytes` in a cell is a
rare path), and the fix here is exactly the migration #4253 points to. Two notes
for the audit: `differential.ts` was **not** in #4253's file list, and it needed
**both** fix classes — the equality swap *and* the value-walk leaf-guard.

# The fix

1. **`data-model/fabric-value.ts` — make `valueEqual` Fabric-aware** (the
   chokepoint #4253's TODOs point to). Same structural recursion as `deepEqual`
   for plain objects/arrays, but defers to canonical `hashOf` for any
   `FabricSpecialObject`. The hash fast-path fires **only** when a special
   object is actually reached, so all-plain-data stays on the same fast,
   short-circuiting path as before.

2. **Migrate the fabric-carrying change-detection gates** from `deepEqual` →
   `valueEqual`: `v2-transaction.ts`, `transaction/mutable-path-write.ts`,
   `transaction/chronicle.ts`, `differential.ts`, `reactive-dependencies.ts`,
   `scheduler/write-propagation.ts`, `scheduler/diagnosis.ts`. (Only sites that
   compare stored `FabricValue`s — schema/path/CFC-atom `deepEqual` callers are
   deliberately left alone.)

3. **`differential.ts` `collectChangedPaths` — treat a `FabricSpecialObject` as
   an atomic leaf** (don't decompose it via `Object.keys`); record a change at
   its path when `valueEqual` says it differs.

# Verification

- New unit tests: `data-model/test/fabric-value.test.ts` — `valueEqual` on
  `FabricBytes`/`FabricRegExp`/`FabricEpochDays`, nested-in-plain-container, and
  special-vs-plain.
- New e2e regression: `runner/test/fabric-primitive-cell-update.test.ts` — keyed
  set, whole-object set, and reactive re-fire.
- Regression: full **runner** suite `647 passed (3309 steps), 0 failed`; full
  **data-model** suite `18 files, 1164 steps, 0 failed`. Typecheck + fmt + lint
  clean on all changed files.

# Scope boundaries

- Touches only `deepEqual` sites that compare stored `FabricValue`s. The many
  `deepEqual` callers comparing schemas, paths, partial-cause descriptors, or
  CFC atoms/labels are **unchanged**.
- `differential.ts`'s leaf-guard is the only value-walk-decomposition site fixed
  here (it's on the reactivity path). #4253's other value-walk sites
  (builder/runner/piece value walks, `SchemaObjectTraverser`, `traverse.ts`,
  `addCommonIDfromObjectID`) are **not** addressed — they can mishandle
  `FabricPrimitive`s in *other* scenarios and are out of scope for CT-1770.

# Open questions / decisions for @danfuzz

1. **Performance.** `valueEqual` hashes (SHA-256) when it reaches a
   `FabricSpecialObject`, on the write/commit/reactivity hot paths. The common
   all-plain-data case is unaffected (fast short-circuit), and frozen stored
   values' hashes are `WeakMap`-cached — but the *incoming* value isn't. Is
   hashing acceptable here, or do you want per-type structural equality (e.g.
   byte-compare for `FabricBytes`) to avoid SHA-256 on the hot path? This is the
   one that probably needs a CI-perf check.
2. **`valueEqual` design / home.** Is hash-based equality the intended shape for
   the canonical `valueEqual`, or did you have a different design in mind? Should
   it stay in `data-model/fabric-value.ts`?
3. **Migration scope.** Did I classify the FabricValue-comparison sites
   correctly — anything I over-included (e.g. `diagnosis.ts`, included for
   consistency) or missed? Should the non-storage gates
   (`reactive-dependencies`, `write-propagation`, `diagnosis`) ship together
   with the storage ones or separately?
4. **Other value-walk sites.** `differential.ts` showed the equality fix alone
   is insufficient where a private-state primitive is *walked*. Are the other
   #4253 value-walk sites reachable for `FabricPrimitive`s today, and do you want
   them folded in here or tracked separately?
5. **Chronicle vs v2 paths.** I fixed both, but the repro only exercises the v2
   native-commit path. Is the chronicle path still live (worth the fix) or
   legacy?

---
## Summary by cubic
Fixes CT-1770: updates to `FabricPrimitive` values (e.g. `FabricBytes`) now apply in place and correctly re-trigger reactive consumers. Resolves equality conflation and path-diff omissions that treated primitives as plain objects.

- **Bug Fixes**
  - Implemented `valueEqual`: Fabric-aware equality that hashes any `FabricSpecialObject`, and correctly handles array holes vs `undefined` and key presence.
  - Replaced `deepEqual` with `valueEqual` in `v2-transaction`, `mutable-path-write`, `chronicle`, `differential`, `reactive-dependencies`, `write-propagation`, `diagnosis`.
  - In `differential`, treat `FabricSpecialObject` as an atomic leaf and emit a change at its path.
  - Tests: unit coverage for `FabricBytes`/`FabricRegExp`/`FabricEpochDays`; e2e for keyed/whole-object sets and reactive re-fire; new runner test for `applyMutablePathWrite` no-op branches (Fabric-aware array element and array-length cases).

<sup>Written for commit 18cc8b29500f43b6529338b23cc4c29c253f7163. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4265?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/scheduler/diagnosis.ts
+++ b/packages/runner/src/scheduler/diagnosis.ts
@@ -1,4 +1,4 @@
-import { deepEqual } from "@commonfabric/utils/deep-equal";
+import { valueEqual } from "@commonfabric/data-model/fabric-value";
 import { isRecord } from "@commonfabric/utils/types";
 import type {
   IExtendedStorageTransaction,
@@ -145,7 +145,7 @@ export function findDifferingWriteKeys(
       differingKeys.push(key);
       continue;
     }
-    if (!deepEqual(previousWrites.get(key), latestWrites.get(key))) {
+    if (!valueEqual(previousWrites.get(key), latestWrites.get(key))) {
       differingKeys.push(key);
     }
   }
@@ -240,7 +240,7 @@ function readInvariantMovedExternally(
     // TODO(danfuzz): `deepEqual` mishandles `FabricValue` (see
     // `utils/deep-equal.ts`); this compares stored `FabricValue`s, so migrate to
     // a `Fabric`-aware equality once available.
-    if (deepEqual(previous.value, value)) continue;
+    if (valueEqual(previous.value, value)) continue;
     // Cover writes of EITHER run: run1's commit moving its own read is the
     // accumulator pattern, and a write-then-read inside the recheck run is
     // nondeterminism, not external interference — both must stay flagged.
--- a/packages/runner/src/scheduler/diagnosis.ts
+++ b/packages/runner/src/scheduler/diagnosis.ts
@@ -237,9 +237,6 @@ function readInvariantMovedExternally(
     const previous = before.get(key);
     // Only reads both runs performed are comparable.
     if (!previous) continue;
-    // TODO(danfuzz): `deepEqual` mishandles `FabricValue` (see
-    // `utils/deep-equal.ts`); this compares stored `FabricValue`s, so migrate to
-    // a `Fabric`-aware equality once available.
     if (valueEqual(previous.value, value)) continue;
     // Cover writes of EITHER run: run1's commit moving its own read is the
     // accumulator pattern, and a write-then-read inside the recheck run is
--- a/packages/runner/src/storage/differential.ts
+++ b/packages/runner/src/storage/differential.ts
@@ -79,12 +79,22 @@ const collectChangedPaths = (
       return;
     }
 
-    // A `FabricSpecialObject` (`FabricPrimitive` leaf / `FabricInstance`
-    // wrapper) is atomic: its state is private, so the key-walk below sees
-    // zero own-keys and would wrongly report "no change". `valueEqual` above
-    // already established they differ, so record a change at this path and
-    // don't decompose. (CT-1770: a `FabricBytes` value updated in place
-    // otherwise never reaches reactive consumers.)
+    // A `FabricSpecialObject` keeps its state in private fields, so the
+    // key-walk below sees zero own-keys and would wrongly report "no change"
+    // even though `valueEqual` above already established they differ. Record a
+    // change at this path and don't decompose. (CT-1770: a `FabricBytes` value
+    // updated in place otherwise never reaches reactive consumers.)
+    //
+    // The `FabricPrimitive` vs `FabricInstance` distinction matters here even
+    // though both are handled the same way: a `FabricPrimitive` genuinely IS an
+    // atomic frozen leaf (no outgoing references), so emitting a single change
+    // at its path is exactly correct. A `FabricInstance` is neither necessarily
+    // frozen nor a leaf — it can hold outgoing references to other
+    // memory-tracked objects, so a fully correct walk would descend into them
+    // (cf. `codecOf()` in cell.ts) and emit per-reference change paths. Lumping
+    // it in as a leaf here is a safe approximation only because nothing in the
+    // system stores `FabricInstance`s yet; revisit when that part of the system
+    // (still in flux) gels.
     if (
       before instanceof FabricSpecialObject ||
       after instanceof FabricSpecialObject
--- a/packages/runner/src/storage/differential.ts
+++ b/packages/runner/src/storage/differential.ts
@@ -1,5 +1,8 @@
 import { unclaimed } from "@commonfabric/memory/fact";
-import { deepEqual } from "@commonfabric/utils/deep-equal";
+import {
+  FabricSpecialObject,
+  valueEqual,
+} from "@commonfabric/data-model/fabric-value";
 import { isRecord } from "@commonfabric/utils/types";
 import type {
   IMemoryAddress,
@@ -72,7 +75,21 @@ const collectChangedPaths = (
   }
 
   if (isRecord(before) && isRecord(after)) {
-    if (deepEqual(before, after)) {
+    if (valueEqual(before, after)) {
+      return;
+    }
+
+    // A `FabricSpecialObject` (`FabricPrimitive` leaf / `FabricInstance`
+    // wrapper) is atomic: its state is private, so the key-walk below sees
+    // zero own-keys and would wrongly report "no change". `valueEqual` above
+    // already established they differ, so record a change at this path and
+    // don't decompose. (CT-1770: a `FabricBytes` value updated in place
+    // otherwise never reaches reactive consumers.)
+    if (
+      before instanceof FabricSpecialObject ||
+      after instanceof FabricSpecialObject
+    ) {
+      pushChangedPath(paths, currentPath, depth);
       return;
     }
 
@@ -173,7 +190,7 @@ const collectChangedPaths = (
     return;
   }
 
-  if (!deepEqual(before, after)) {
+  if (!valueEqual(before, after)) {
     pushChangedPath(paths, currentPath, depth);
   }
 };
@@ -184,7 +201,7 @@ const addStateChange = (
   before: State["is"] | undefined,
   after: State["is"] | undefined,
 ): void => {
-  if (deepEqual(before, after)) {
+  if (valueEqual(before, after)) {
     return;
   }
 
--- a/packages/data-model/src/fabric-value.ts
+++ b/packages/data-model/src/fabric-value.ts
@@ -34,13 +34,83 @@ export {
   shallowFabricFromNativeValue,
 } from "./native-conversion.ts";
 
-import { deepEqual } from "@commonfabric/utils/deep-equal";
+import { isRecord } from "@commonfabric/utils/types";
+import { FabricSpecialObject } from "./interface.ts";
+import { hashStringOf } from "./value-hash.ts";
 
 /**
- * Compares two fabric values for equality.
+ * Compares two `FabricValue`s for logical (content) equality.
+ *
+ * This is the `data-model`-aware equality that the storage layer's no-op /
+ * change-detection gates need. It mirrors `deepEqual()`'s structural recursion
+ * for plain objects and arrays (including its hole-vs-stored-`undefined`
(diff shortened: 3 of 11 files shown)


danfuzz:
I'll add some inline comments, but in terms of the gestalt:

* The `data-model` part of this looks fine. There's things we can ultimately do to make it work better, but in terms of a short-term unblock it's a-okay.

* As for `runner` and `storage`: There is an important distinction between `FabricPrimitive` and `FabricInstance` that will have to be honored: `FabricPrimitive` is "primitive-like" in the JS sense, in that it can always be treated as a frozen leaf (no outgoing references). `FabricInstances` are neither necessarily frozen nor are they (generally) leaves; they can have outgoing references to other objects tracked by `memory`. it's probably "fine" for now not to deal with `FabricInstance` structure because we don't yet ever try to store any of them, but it's something not to lose track of.

packages/runner/src/scheduler/diagnosis.ts
```diff
@@ -240,7 +240,7 @@ function readInvariantMovedExternally(
     // TODO(danfuzz): `deepEqual` mishandles `FabricValue` (see
```
danfuzz: Given the change, it's safe to remove this `TODO(danfuzz)`. Ditto the other places where `deepEqual()` is getting replaced.

packages/runner/src/storage/differential.ts
```diff
@@ -72,7 +75,21 @@ const collectChangedPaths = (
   }
 
   if (isRecord(before) && isRecord(after)) {
-    if (deepEqual(before, after)) {
+    if (valueEqual(before, after)) {
+      return;
+    }
+
+    // A `FabricSpecialObject` (`FabricPrimitive` leaf / `FabricInstance`
```
danfuzz: `FabricInstance` isn't atomic in the sense of (JS) primitives, whereas `FabricPrimitive` _is_. Ultimately, to correctly handle a `FabricInstance`, code needs to walk its outgoing references. Given that nothing in the system today tries to store `FabricInstance`s, it's a bit of a moot point, but I don't want to lose sight of the fact that this is a spot (this particular code) where the distinction matters. So, I'd recommend adding a comment.

Why a comment and not just-fix-it?: How `FabricInstance` works is still in flux, so, while there _is_ a way to do the walk, I'm reluctant to suggest actually doing it, as it will likely just end up being another site for churn as that part of the system gels. (But ICYC: You can look for `codecOf()` in `cell.ts` for the probably-to-be-changed tactic that's currently used.)

danfuzz (changes requested):
Only request is the comment tweaks as described in my PR comments.

packages/runner/src/storage/differential.ts
mathpirate: Done — expanded the comment to spell out the distinction: a `FabricPrimitive` is a genuine frozen leaf, so emitting a single change at its path is exactly right; a `FabricInstance` can hold outgoing references a fully-correct walk would descend into (pointed at `codecOf()` in `cell.ts`), and treating it as a leaf here is called out as safe only because nothing stores `FabricInstance`s yet. Left the actual walk for when that part of the system gels, per your note.

packages/runner/src/scheduler/diagnosis.ts
mathpirate: Removed — all the `TODO(danfuzz)` deepEqual markers at the sites migrated to `valueEqual` are gone (v2-transaction x9, reactive-dependencies x3, write-propagation, diagnosis). The one I left is `prepare.ts:1909` (the CFC schema-default write-policy check), which this PR does not migrate, so its marker stays. That is the one remaining fabric-carrying caller from #4253's list that is off the CT-1770 path — happy to fold it in if you would like.

mathpirate:
Thanks for the review @danfuzz — addressed in `d6e6c91cf` (both inline threads resolved):

- **Comment tweaks**: expanded the `differential.ts` leaf-guard comment with the `FabricPrimitive`-vs-`FabricInstance` distinction, and removed the now-resolved `TODO(danfuzz)` deepEqual markers at every migrated site.
- **`FabricInstance` structure**: not handled structurally here — it is treated as an atomic leaf in `differential` per your note that nothing stores them yet, with the comment flagging it to revisit. One thing worth noting: `valueEqual` itself *is* correct for instances, since it compares them via `hashOf`, which already folds in their codec-encoded outgoing references; the leaf approximation only affects change-*path* granularity in `differential`, not equality.
- **Coverage gate**: simplified `valueEqual` to a minimal FabricValue-shaped recursion (dropping the deepEqual-copy's non-Fabric branches — foreign class instances, named array props — that cannot occur in a FabricValue) plus added edge tests, so `fabric-value.ts` is back to 100%.

Re-validated: runner 647 passed / 0 failed, data-model 1167 steps / 0 failed.

danfuzz (approved):
Many thanks to all involved entities!

danfuzz:
In re Cubic's comment: I have no objection _either_ sticking with the current scope _or_ expanding the conversion to `valueEqual()` to more sites. I'm specifically okay with the former because we know there are plenty more sites to fix, so it's not like just fixing these adjacent sites is going to get the whole world to be happy, and I'm not about to ask that this PR be a whole-world reckoning.

packages/runner/src/scheduler/diagnosis.ts
```diff
@@ -145,7 +145,7 @@ export function findDifferingWriteKeys(
       differingKeys.push(key);
       continue;
     }
-    if (!deepEqual(previousWrites.get(key), latestWrites.get(key))) {
+    if (!valueEqual(previousWrites.get(key), latestWrites.get(key))) {
```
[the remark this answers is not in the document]
mathpirate: @/tmp/cubic-reply.md

PR #4270, opened 2026-06-20
fix(runtime): add runtime.settled() that waits for async builtin work (stabilize lunch-poll sqlite flake)

seefeldb:
## Problem

`packages/patterns/lunch-poll/main.test.tsx` flakes in the "Pattern Unit Tests" CI shard: the SQLite-backed history assertions — `assert_two_history`, `assert_one_history_after_remove`, `assert_history_cleared` — intermittently report `Expected true, got false`. Passes in isolation; fails under the shard's heavier CPU/IO load. Pre-existing on `main` (reproduced on a clean checkout), surfaced while shepherding #4203.

## Root cause

The async builtins — reactive sqlite queries (`db.query`), `fetchData`, `fetchProgram`, `llm` / `llmDialog` — perform their real I/O (a network/LLM call or a sqlite RPC, plus the result writeback) from a **post-commit outbox flush**, *after* the handler returns. `scheduler.idle()` deliberately does **not** wait for that I/O — a handler must never block on the network — and nothing else did either. So a reader settling with `idle()` could observe a query result while its flush was still in flight and read a stale `{ pending: true }`.

## Approach

**Runtime — a dedicated "fully settled" path (`idle()` is untouched):**
- `Runtime` gains a `#pendingAsyncWork` registry, `trackAsyncWork()`, and **`settled()`** — `settled()` drains the scheduler (idle), storage (synced), **and** every in-flight async builtin operation, looping until quiescent.
- The async builtins register their in-flight I/O via `trackAsyncWork`:
  - The **scheduler** registers an effect-bearing commit's promise — a race-free barrier (the sqlite query flush runs inside that commit and awaits its RPC + writeback, and it guarantees a fire-and-forget builtin's flush has run and registered its own work). Gated by a new `hasPendingPostCommitEffects()` on the transaction; plain commits keep the fire-and-forget fast path.
  - **`fetchData` / `fetchProgram` / `llm` / `generateText` / `generateObject` / `llmDialog`** register their network/LLM promise. **`streamData` is intentionally not tracked** — an SSE stream is long-lived and has no "done".

**cf test — an opt-in `{ settle: true }` step (not a blanket per-action settle):**
- The per-action settle stays the light `idle()` + `synced()` loop. A new **`{ settle: true }` step** waits for full settlement via `runtime.settled()`. A test that asserts on an async-builtin result inserts it before the assertion.
- `lunch-poll/main.test.tsx` inserts `{ settle: true }` before its SQLite-backed history assertions. The other sqlite patterns (`cfc-row-label-*`) passed on `main` with the light settle, so they're unaffected.

## Tests

- **Deterministic regression test** in `sqlite-builtins.test.ts`: with the server read slowed, `idle()` still returns `{ pending: true }`, while `runtime.settled()` settles the reactive query to `{ pending: false }`. Red without the registration, green with it.
- Full runner suite **633 passed, 0 failed**; fetch/llm/stream outbox suites green.
- CI-faithful repro (lunch-poll alone under load from *other* pattern-test processes — different spaces, no shared cell-db): **30/30 pass** vs. repeated `assertion_17/18/19` failures on pristine `main`.

> Note: running the same test file in N concurrent processes is **not** a valid repro — cf test cell-dbs are files at `$TMPDIR/cf-cell-<hash(space)>-<hash(id)>.sqlite`, so same-file concurrency shares one db and contaminates results. CI never does this (different test files → different spaces); its concurrency is parallel shards.

---

### Coverage

`NEW_PERF_BASELINE: coverage-debt: packages/runner uncovered lines = 8535 lines`

The +6 uncovered lines are the `runtime.trackAsyncWork(...)` wrappers in the `fetchData` / `fetchProgram` / `llm` / `llmDialog` post-commit flushes — they wrap network/LLM I/O that only the integration suites exercise, not the unit-coverage artifact. Accepting the narrow ratchet bump per docs/development/CI_PERFORMANCE.md.

### DIFF
--- a/packages/cli/lib/multi-user-test-worker.ts
+++ b/packages/cli/lib/multi-user-test-worker.ts
@@ -64,6 +64,7 @@ export interface ParticipantInitResult {
 }
 
 const SETUP_CAUSE = "multi-user-test-setup";
+const SETTLE_FAST_MS = 2;
 
 let runtime: Runtime | undefined;
 let storageManager:
@@ -88,11 +89,12 @@ function rt(): Runtime {
 }
 
 async function settle(maxIterations = 20): Promise<void> {
-  // `settled()` drains the scheduler (idle), storage (synced) AND every
-  // in-flight async builtin operation (sqlite query RPC + writeback, fetch,
-  // llm) that `idle()` returns before — so a participant's assertion never
-  // reads a query result while its post-commit flush is still in flight.
-  await rt().settled(maxIterations);
+  for (let i = 0; i < maxIterations; i++) {
+    const start = performance.now();
+    await rt().idle();
+    await storageManager!.synced();
+    if (performance.now() - start < SETTLE_FAST_MS) return;
+  }
 }
 
 const stepPeekSchema = {
--- a/packages/cli/lib/multi-user-test-worker.ts
+++ b/packages/cli/lib/multi-user-test-worker.ts
@@ -64,7 +64,6 @@ export interface ParticipantInitResult {
 }
 
 const SETUP_CAUSE = "multi-user-test-setup";
-const SETTLE_FAST_MS = 2;
 
 let runtime: Runtime | undefined;
 let storageManager:
@@ -89,12 +88,11 @@ function rt(): Runtime {
 }
 
 async function settle(maxIterations = 20): Promise<void> {
-  for (let i = 0; i < maxIterations; i++) {
-    const start = performance.now();
-    await rt().idle();
-    await storageManager!.synced();
-    if (performance.now() - start < SETTLE_FAST_MS) return;
-  }
+  // `settled()` drains the scheduler (idle), storage (synced) AND every
+  // in-flight async builtin operation (sqlite query RPC + writeback, fetch,
+  // llm) that `idle()` returns before — so a participant's assertion never
+  // reads a query result while its post-commit flush is still in flight.
+  await rt().settled(maxIterations);
 }
 
 const stepPeekSchema = {
--- a/packages/cli/lib/test-runner.ts
+++ b/packages/cli/lib/test-runner.ts
@@ -1217,27 +1217,11 @@ export async function runTestPattern(
       // `{ settle: true }` step: wait for FULL settlement (scheduler + storage +
       // in-flight async builtin I/O — sqlite query RPC + writeback, fetch / llm)
       // before the next step. A test inserts this before an assertion that reads
-      // an async-builtin result so it never observes a half-settled state.
+      // an async-builtin result so it never observes a half-settled state. The
+      // step is transparent — it produces no result. A settle timeout propagates
+      // to the outer handler and fails the whole run (a stuck settle is fatal).
       if (isSettle) {
-        if (stepValue.skip || stepValue.settle === false) {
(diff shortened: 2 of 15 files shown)


seefeldb:
@mathpirate just my agent's attempt to make the test less flaky, but i think the correct answer here is to not overload idle() and instead cover all kinds of in-flight tasks from llm, fetch data, etc in a separate longer cycle idle. will kick that off now.

seefeldb:
Reworked per @seefeldb's direction — instead of overloading `idle()`, this adds a dedicated `runtime.settled()` that covers **all** in-flight async builtin work.

**What `settled()` does:** loops `scheduler.idle()` → `storageManager.synced()` → drain a new `#pendingAsyncWork` registry, until quiescent. `idle()` is left completely unchanged.

**What registers into the registry (`runtime.trackAsyncWork`):**
- **sqlite queries** — the scheduler registers the effect-bearing commit promise (gated by a new `tx.hasPendingPostCommitEffects()`). Registered synchronously before the run's promise resolves, so it's a race-free barrier: the sqlite query flush runs *inside* that commit and awaits its RPC + writeback.
- **`fetchData` / `fetchProgram` / `llm` / `generateText` / `generateObject` / `llmDialog`** — register their network/LLM promise. Their flushes still fire-and-forget the I/O, so the handler/commit stay fast; `settled()` waits for the result + writeback.
- **`streamData` is intentionally not tracked** — a long-lived SSE stream has no "done" to wait for and would hang `settled()`.

**Why the commit-promise barrier (subtle):** registering builtin I/O in the *flush* alone races `settled()`'s `synced()` (microtask order after the inner commit). The commit-promise registration is synchronous-in-action → race-free, and is the barrier that guarantees a fire-and-forget builtin's flush has run and registered its own work before `settled()` concludes.

The cf-test settle loop (`settleRuntime`) and the multi-user test worker now call `runtime.settled()`.

**Out of scope (left as-is):** the production CLI `idle()` calls (`callable.ts`, `piece.ts`, `piece-render.ts`) — those are "wait for processing" in commands, not settle-before-assertion, so switching them to `settled()` would be a behavior change. Easy to add if we want CLI commands to block on async builtin I/O too.

Validated: full runner suite 633 passed / 0 failed; fetch/llm/stream outbox suites green; CI-faithful repro (lunch-poll alone under load from other pattern tests) 30/30 vs. assert_17/18/19 failures on pristine main; deterministic regression test (red without the registration, green with).

seefeldb:
Refined per follow-up: rather than blanket-applying `runtime.settled()` after every action in the cf-test loop, the full settle is now an **opt-in test step**.

- The per-action settle reverts to the light `idle()` + `synced()` loop (main's behavior).
- New **`{ settle: true }` step** waits for full settlement (`runtime.settled()` — scheduler + storage + in-flight async builtin I/O) before the next step. Added to the step schema and the exported `TestStep` union.
- `lunch-poll/main.test.tsx` inserts `{ settle: true }` before its SQLite-backed history assertions; the other sqlite patterns (`cfc-row-label-*`) passed on main with the light settle, so they're untouched.

This keeps the default settle cheap and lets a test that asserts on async-builtin results (a `db.query`, fetch, llm) explicitly declare where it needs to wait — useful since you often *want* to assert an in-flight `{ pending: true }` state. CI-faithful repro under load: 30/30.

PR #4111, opened 2026-06-12
refactor(runner,memory): piece-level resume + slim observations (scheduler-v2 phase 7)

seefeldb:
Stack: 08/Phase 7 - based on scheduler-v2/08-phase5 (#4108)

## Summary
- Move persistent scheduler rehydration to piece-level batch snapshot preload and synchronous registration-time application.
- Delete the per-action rehydration race/timeout/token path.
- Slim persisted scheduler observations to payload v2, keep v1/v2 readers, and bump the runtime fingerprint to `runner:scheduler:v2`.
- Add the clean-resume I2/I7 gate proving zero reruns and zero cell-data fetches for a clean resumed piece.
- Complete the post-phase-7 documentation cleanup and remove stale mode-control/debug trace diagnostics exposed by final verification.

## Validation
- `deno fmt` / `deno lint` / `deno check` on touched files passed.
- `cd packages/shell && deno test --allow-read --allow-write --allow-run --allow-env test/debug-utils.test.ts` passed.
- `cd packages/runtime-client && deno test --allow-env --allow-read --allow-ffi backends/runtime-processor.test.ts` passed.
- `cd packages/runner && deno task test` passed: 595 passed (3099 steps), 0 failed, 0 ignored (10 steps), 2m8s.
- Phase 7 persistence pack passed: `reload-rehydration`, `reload-sibling-overdirty`, and `scheduler-observations`.
- Memory v2 scheduler state and stacked-commit regression packs passed.
- Scheduler persistent-state bench on Apple M3 Max / Deno 2.8.1: clean 100 actions 2.5 ms; targeted dirty 100 actions 2.3 ms; clean 1000 actions 11.8 ms; targeted dirty 1000 actions 6.9 ms.
- Final cleanup greps passed: no mode-control APIs, no `schedulerHistoricalMightWrite`, no old-vs-new scheduler flag, and only the intentional graph-snapshot `pullMode: true` diagnostic remains.

See `docs/specs/scheduler-v2/implementation/PROGRESS.md` for the full Phase 7 log.

---
## Summary by cubic
Speeds up resume by preloading all scheduler snapshots per piece and applying them synchronously at registration, removing rehydration races/timeouts and per-element sync gating. Slims persisted observations to v2 (optional write surfaces) and keeps persistent state behind `EXPERIMENTAL_PERSISTENT_SCHEDULER_STATE`.

- **Refactors**
[...]

PR #4287, opened 2026-06-21
perf(runner): bound settle work-set closure to active wavefront (scheduler-v2 decision 16)

seefeldb:
Stacked on #4111 (scheduler-v2 phase 7). Implements **decision 16** (bounded settle work-set closure).

## Problem
After 3c deleted `staleness.ts`, `addLiveDownstreamClosure` (§7.2) expanded the **entire live downstream cone** of every invalid seed, value-blind. Under a wide-fan-out hub (a vote tally that many clean readers — including a whole-state render sink — observe), the per-settle-pass work-set grew **3–4×** vs the staleness-pruned v1 set.

A multi-user investigation (lunch-poll concurrent voting) confirmed: **executions did not change** — value-gating (§7.3) already skips the clean members — but each pass iterated and value-checked a much larger consideration set. It's the dual of decision 15: there, deleting the eager staleness signal inflated a *query* (the event-preflight walk, fixed by inverting it); here it inflated the settle *closure*.

## Fix
Bound the closure to the **active wavefront**: add each invalid node's *direct* live readers (tier 1), but recurse past a reader only if it is itself `invalid`/`never-ran`. A clean reader can't become runnable until it actually runs; when it does and changes value, P1 invalidates *its* readers and the loop re-seeds them next iteration. So the per-pass closure is the invalid set plus one clean tier, not the whole cone — **without** re-adding any per-data-change transitive marking (lazy/non-transitive invalidation and decision-15 dormancy preserved). This is the walk-computed form of decision 15's `hasInvalidUpstream` signal: the downstream walk *is* the gate, so no maintained refcount is needed unless the walk itself becomes hot.

Tier 1 is **kept** (not pruned): some clean readers must be scheduled in the same pass as the write that feeds them — notably materializer-output readers. The aggressive variant (prune tier 1 too) fails the cutover guard `should schedule normal output readers when a materializer input dirties`, which is exactly the boundary.

## Measured (M3 Max)
| | before (phase7) | after |
|---|---|---|
| lunch concurrent-vote peak per-pass work-set | 62 / 61 | **43 / 49** |
| single-runtime note-create @128 | 42.2 ms | **39.3 ms** |
| action executions (`actionRunTrace`) | — | **unchanged** |
| settle-iteration count | — | **unchanged** |
| full runner suite | green | **green (620)** |

Closing the remaining gap to v1 (the kept tier-1 clean readers) would need value-awareness or the maintained refcount; not adopted here.

## Spec
- §7.2 work-set definition + prose updated to the bounded closure.
- New §15 **decision 16**; inventory staleness row cross-referenced.

---
## Summary by cubic
Bound the settle work-set to the active wavefront by only recursing through readers that are invalid or never-ran. This reduces per-pass work under wide fan-out with no change to executions or iteration count.

- **Performance**
  - Implementation: update `packages/runner/src/scheduler/settle.ts` (`addLiveDownstreamClosure`) to add direct live readers but recurse only if the reader is invalid/never-ran; keep tier‑1 readers for materializer-output scheduling.
  - Results: peak work-set 62→43 / 61→49; note-create @128 42.2→39.3 ms; executions and settle-iterations unchanged; full runner suite green.
  - Docs: update §7.2 and add decision 16 to record the bounded closure.

<sup>Written for commit c47693b0d3a658111725f52dd07ed4b3915a3efb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4287?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seefeldb:
Superseded by #4288, which lands the entire scheduler-v2 cutover (3c -> 3d -> 3f -> phase 4 -> phase 5 -> phase 7 -> decision 16) as a single PR rebased on current main and run through real CI. Closing in favor of #4288.

PR #4292, opened 2026-06-22
perf(memory): intern schemas in sync frames

bfollington:
## Summary

- add negotiated memory v2 schema tables for schema-bearing link payloads
- use hash-keyed content-addressed schema refs: `schema-ref@2:<tagged schema hash>`
- expand schema refs client-side before WatchView/storage consumption, preserving old `SessionSync` shape for runtime users
- keep old expanded frames for peers that do not advertise `syncSchemaTableV2`

## Snapshot

Synthetic repeated-schema sync fixture on the current branch:

- before: 128 upserts, 6400 schema marker occurrences, 2,814,847 encoded bytes
- after: 25 schema marker occurrences, 61,743 encoded bytes
- reduction: 97.81% encoded bytes on the fixture

The earlier numeric-ref draft measured 49,660 encoded bytes after compression. The current hash-ref format is slightly larger, but it makes the frame-local schema ids content-addressed and self-validating.

## Cubic feedback addressed

- P2 sibling traversal: fixed schema-table rewrite/expand so a link payload rewrite does not skip schema-bearing sibling fields in the same object.
- P1 mixed-version compatibility: moved the hash-keyed wire shape behind `syncSchemaTableV2` and `schema-ref@2:`. Legacy `syncSchemaTable`-only peers now fall back to fully-expanded frames instead of attempting to decode hash refs as numeric refs.

## Verification

- `mise exec -- deno fmt packages/memory/v2.ts packages/memory/v2/server.ts packages/memory/v2/sync-schema-table.ts packages/memory/test/v2-sync-schema-table-test.ts packages/memory/test/v2-test.ts`
- `mise exec -- deno lint packages/memory/v2.ts packages/memory/v2/server.ts packages/memory/v2/sync-schema-table.ts packages/memory/test/v2-sync-schema-table-test.ts packages/memory/test/v2-test.ts`
- `mise exec -- deno test --allow-read --allow-write --allow-net --allow-ffi --allow-env packages/memory/test/v2-sync-schema-table-test.ts packages/memory/test/v2-test.ts`
- `mise exec -- deno task check` in `packages/memory`

Refs CT-1775
Follow-up for global schema-store architecture: CT-1781

### DIFF
--- a/packages/memory/test/v2-sync-schema-table-test.ts
+++ b/packages/memory/test/v2-sync-schema-table-test.ts
@@ -1,0 +1,230 @@
+import { assert, assertEquals, assertExists } from "@std/assert";
+import { linkRefFrom } from "@commonfabric/data-model/cell-rep";
+import type { JSONSchema } from "@commonfabric/api";
+import {
+  encodeMemoryBoundary,
+  getMemoryProtocolFlags,
+  MEMORY_PROTOCOL,
+  type ResponseMessage,
+  type ServerMessage,
+  type SessionEffectMessage,
+  type SessionOpenResult,
+  type SessionSync,
+  type WatchAddResult,
+} from "../v2.ts";
+import { Server } from "../v2/server.ts";
+import {
+  compressServerMessageSchemas,
+  expandServerMessageSchemas,
+  type SchemaTableSessionSync,
+} from "../v2/sync-schema-table.ts";
+
+const textEncoder = new TextEncoder();
+
+const encodedBytes = (value: ServerMessage): number =>
+  textEncoder.encode(encodeMemoryBoundary(value)).byteLength;
+
+const largeSchema = (): JSONSchema => ({
+  type: "object",
+  $defs: Object.fromEntries(
+    Array.from({ length: 48 }, (_, index) => [
+      `Definition${index}`,
+      {
+        type: "object",
+        properties: {
+          id: { type: "string" },
+          title: { type: "string" },
+          summary: { type: "string" },
+          count: { type: "number" },
+          active: { type: "boolean" },
+        },
+        required: ["id", "title"],
+      },
+    ]),
+  ),
+  properties: Object.fromEntries(
+    Array.from({ length: 24 }, (_, index) => [
+      `field${index}`,
+      { $ref: `#/$defs/Definition${index % 48}` },
+    ]),
+  ),
+});
+
+const repeatedSchemaSync = (
+  count = 128,
+): SessionSync => {
+  const schema = largeSchema();
+  return {
+    type: "sync",
+    fromSeq: 0,
+    toSeq: 1,
+    upserts: Array.from({ length: count }, (_, index) => ({
+      branch: "",
+      id: `of:test-${index}`,
+      scope: "space" as const,
+      seq: 1,
+      doc: {
+        value: {
+          title: `Document ${index}`,
+          primary: linkRefFrom({
+            id: `of:target-${index}`,
+            path: [],
+            schema,
+          }),
+          secondary: linkRefFrom({
+            id: `of:secondary-${index}`,
+            path: ["nested"],
+            schema,
+          }),
+        },
+      },
+    })),
+    removes: [],
+  };
+};
+
+const syncEffect = (effect: SessionSync): SessionEffectMessage => ({
+  type: "session/effect",
+  space: "did:key:z6Mk-sync-schema-table",
+  sessionId: "session:sync-schema-table",
+  effect,
+});
+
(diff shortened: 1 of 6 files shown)


bfollington (commented):
_cf-review via Claude Opus 4.8, on behalf of @bfollington_

**Scope** · `packages/memory` only (6 files, +630/−2): a negotiated, frame-local schema table that interns repeated link-payload schemas in server→client sync frames and transparently re-expands them client-side. **Coverage** · deep-read all changed source + key deps (`internSchema`/`toDeepFrozenSchema`, `decodeMemoryBoundary`, handshake, negotiation). **Checks run** · both new/changed test files (pass), `deno lint` + `deno fmt --check` on touched files (clean).

This is a clean, well-built change. The round-trip is verified lossless and the design is genuinely backward-compatible. No blocking issues.

## 🟡 Improvements / questions

**1. Default-ON, with no `experimental.*` parity — confirm the rollout is intended.** _(needs author confirmation)_
`syncSchemaTableEnabled = true` at module scope, and `resetSyncSchemaTableConfig()` resets to `true` — so the optimization is **on by default everywhere**. Its three sibling flags all default `false` and are driven by `runtime.experimental.*` (`runtime.ts:351-358`). `setSyncSchemaTableConfig` is **not** wired into `runtime.ts`, so an operator can only toggle it via the module-level setter, not the standard `experimental` surface, and the runtime's dispose reset (`runtime.ts:604-606`) doesn't touch it either. Negotiation makes this _safe_ (verified: old↔new peers fall back to expanded frames, and `compatibleMemoryProtocolFlags` only gates on `modernCellRep`), so this is a rollout-posture question, not a bug: is shipping ON by default with no experimental gate the intent, vs. staging it off-first like the others? If staged rollout is wanted, gate it through `experimental.syncSchemaTable` for parity and operability.

**2. Coherence: the normative protocol spec doesn't mention the new wire shape.** _(verified)_
`docs/specs/memory-v2/04-protocol.md` enumerates the hello `flags` (lines 49-64, 149-153) and `interface SessionSync` (line 209), but nothing there mentions `syncSchemaTable`, the optional `schemaTable` field, or the `schema-ref@1:N` encoding. An independent implementer working from the spec would be blindsided by a compressed frame. It isn't _actively wrong_ — negotiation hides the new shape from any peer that doesn't advertise the flag (which a spec-only implementer wouldn't), and `commitPreconditions` is likewise already absent from the examples — but a few lines documenting the negotiated optimization would honor the "don't mislead someone who lands on the spec" bar. Non-blocking; fix-what-you-touch or file as follow-up.

## ⚪ Nits

**3. `console.log` JSON dump in a landed test.** `test/v2-sync-schema-table-test.ts:108-122` prints a metrics blob on every run, and the test name ("…experiment captures repeated schema savings") reads like spike residue. Either keep it as an intentional, labeled benchmark or drop the log — the `assertEquals`/`compressedBytes < bytes/5` assertions are what actually guard the behavior.

**4. Per-connection re-compression on broadcast.** `server.ts:321-326` runs `compressServerMessageSchemas` independently in each `Connection.send()`, so a broadcast sync is re-traversed and re-hashed once per flag-on connection. Interning caches by object identity so repeated identical-reference schemas are cheap, but the full doc walk is per-connection. Possible follow-up: compress once and share across same-capability connections.

## What I verified as sound (for confidence)
- **Round-trip is lossless.** `internSchema`→`toDeepFrozenSchema` only _freezes_ (preserves structure and key order); it never reorders or drops properties. `assertEquals(expanded, message)` holds, and the test passes (98.24% byte reduction on the fixture, reproduced locally).
- **No frozen/unfrozen asymmetry.** `decodeMemoryBoundary` already returns deep-frozen trees (existing "returns deeply-frozen plain JSON trees" test), so the `deepFreeze(expanded)` in `expandSessionSyncSchemas` _maintains_ that invariant rather than introducing a divergence between compressed and uncompressed frames.
- **Negotiation is correct & backward-compatible.** Both sides must advertise the flag (`server.ts:456-458`); old peers never receive compressed frames; `expandServerMessageSchemas` is a no-op on frames without a `schemaTable`.
- **Compression is scoped to link-payload `schema` fields only.** Non-schema user strings — even a hypothetical `"schema-ref@1:0"` elsewhere in a doc — are never expanded, because expansion only touches `payload.schema`. `expandSchemaRef` is also defensively guarded (throws on out-of-range/non-integer indices).
- **Anti-duplication: clean.** Reuses canonical `internSchema` (`data-model/schema-hash`), `deepFreeze`, `isPlainObject`, and `LINK_V1_TAG` — no re-forked hashing/serialization/cloning.

**Verdict:** ✅ Mergeable. No blocking issues; resolve the default-ON rollout question (#1) consciously before merge, and consider a short spec note (#2).

PR #4206, opened 2026-06-17
Split fetchData

Hixie:
Migrate fetchData to dedicated features per type (fetchJson, fetchText, fetchBinary, fetchJsonUnchecked).

This PR is split into separate commits with their own commit messages for more convenient review.

Currently this (apparently, not tested manually) breaks image rendering in the lunch poll; a separate workstream is ongoing to resolve that issue.

---
## Summary by cubic
Replaces the untyped `fetchData` with focused, typed fetch builtins and adds a bytes-based image renderer. JSON is schema‑validated at fetch time, existing mutex/auth behavior is preserved, and runtime, TS transforms, UI, docs, and tests are updated.

- **New Features**
  - `fetchJson<T>` with compiler‑injected schema; compile error if type arg is missing. `fetchJsonUnchecked` parses JSON without validation.
  - `fetchText` and `fetchBinary` (returns `{ bytes: FabricBytes, mediaType }`); `@commonfabric/api` exports `FabricBytes` and `FetchBinaryResult`.
  - `<cf-image>` renders `fetchBinary` via object URLs; JSX types added.
  - Runtime/CFC: builder/registry updated; eager‑result set and sink inventory include all new fetch builtins; shared fetch core keeps cross‑tab mutex, settled tracking, and Toolshed first‑party request signing. Docs at `docs/common/capabilities/fetch.md`. Data‑URL mode removed.

- **Migration**
  - BREAKING: `fetchData` is removed. Replace:
    - JSON → `fetchJson<T>(...)` (or `fetchJsonUnchecked` to skip validation).
    - Text → `fetchText(...)`.
    - Binary → `fetchBinary(...)`; render with `<cf-image bytes={$binary.bytes} media-type={$binary.mediaType} />`.
  - Update CFC sink ceilings/policies to the new fetch names. Result shape remains `{ pending, result, error }`. Re‑deploy pieces that used `fetchData`.

<sup>Written for commit 4b9b8745d6d49f33ca0a113a0595c8667387b700. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4206?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #4306, opened 2026-06-23
fix(transformers): error on unknown reactive types 

Hixie:
Split into two commits for easier review.

A lot of the fixes are basically just "turns out it's not unknown".

---
## Summary by cubic
Unknown reactive types now fail transformation. Wish results are directly renderable and LLM errors are strings; we added explicit generics and narrowed pass‑through cells to avoid unknowns, improved coverage (including native search grounding), and removed obsolete runtime fixtures.

- **Bug Fixes**
  - `packages/ts-transformers`: Promote unknown-type diagnostics (reactive capture, pattern result) to errors; rename helper to `reportUnknownReactiveType`; add tests for destructured keys and renamed captures; update fixtures to use typed event payloads and typed `fetchData` results.
  - Patterns: add generics to `fetchData<T>()`, `wish<T>()`, `generateObject<T>()`; switch no‑payload handlers/streams to `Stream<void>` with `.send()` no args; narrow pass‑through piece/result cells to small object shapes with optional `[NAME]` (e.g. Journal, Suggestion, Store Mapper); fix Gmail test command roots.
  - `packages/api`/`packages/schema`: `wish()` and its schema overload return `WishState<T> & UIRenderable` so wish results render directly; shared profile demo renders the wish value.
  - `packages/runner`: LLM state/schemas use `error: string`; builtins expose typed `error` cells; `handleLLMError` stores the message; tests assert string errors and cover native search grounding source extraction/dedup.
  - `packages/patterns/test/source-coverage`: add plain‑Deno harness with local CF/JSX stubs; run via a JS child using the workspace `deno.jsonc` and a local import map; exclude stubs from authored checks; focused tests exercise changed modules; one CLI coverage test scopes `CF_LOG_LEVEL` to the child command.
  - `packages/generated-patterns`: remove unknown‑capture runtime fixtures that no longer compile.

- **Migration**
  - Add explicit types at reactive boundaries (e.g., `fetchData<string>()`, `wish<T>()`, `generateObject<T>()`); treat built‑in LLM `error` as `string`.
  - For no‑payload handlers/streams, change to `Stream<void>` and call `.send()` with no args.

<sup>Written for commit 8975fe6c63034ef9148ee879f32468465383322f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4306?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mathpirate (approved):
awesome, lgtm, thanks so much. the patterns and fixtures you updated here are mostly ones i authored back when `unknown` had different semantics in our system and which i missed updating :) it's great to bring them all in line and to enforce this type invariant in our system

seems like the main API change here is `fetchData` requiring an explicit template type to be provided, is that right?

packages/patterns/system/journal.tsx
```diff
@@ -19,7 +19,7 @@ import {
 type JournalEntry = {
   timestamp?: number;
   eventType?: string;
-  subject?: Writable<unknown>;
+  subject?: Writable<any>;
```
mathpirate: robot points out: wonder whether we might be able to specify a narrower type than `any` here? not sure if there's a feasible concrete union

packages/patterns/system/suggestion.tsx
```diff
@@ -87,7 +87,7 @@ export default pattern<
   {
     situation: string;
     context: { [id: string]: any };
-    initialResults: Writable<unknown>[] | Default<[]>;
+    initialResults: Writable<any>[] | Default<[]>;
```
mathpirate: robot points out: wonder whether we might be able to specify a narrower type than any here? not sure if there's a feasible concrete union

Hixie:
Yes regarding `fetchData`, but that'll be obsoleted by https://github.com/commontoolsinc/labs/pull/4206 soon.

Hixie:
This is currently on pause while I figure out how to make patterns have proper coverage tracking.

PR #4301, opened 2026-06-22
Extract valueEqual() to its own file; tighten its param types

danfuzz:
## Summary

Two small `valueEqual()` polish follow-ons (TASK-0050), unblocked by #4265
(CT-1770) landing:

- **(a) Tighten `valueEqual()`'s parameter types** `unknown` → `FabricValue`.
- **(b) Move `valueEqual()` to its own file** `valueEqual.ts`, re-exported
  through the `fabric-value.ts` barrel so `fabric-value.ts` stays a pure
  re-export surface.

The function body itself is unchanged (verbatim move) — this is a typing +
file-organization polish, not a behavior change.

## Scope

Three files (own-commit range over `5abe477c7`):

- `packages/data-model/src/fabric-value.ts` — the inline `valueEqual`
  definition (the lone non-re-export) is removed; the file is now a pure
  barrel, with `export { valueEqual } from "./valueEqual.ts";` added.
- `packages/data-model/src/valueEqual.ts` — **new**; the moved definition,
  with params tightened to `FabricValue` and its three dependencies imported
  directly (`isRecord`, `FabricSpecialObject`, `hashStringOf`).
- `packages/runner/src/scheduler/diagnosis.ts` — the param-tighten ripple to
  two call sites (see below).

All other importers reach `valueEqual` through the
`@commonfabric/data-model/fabric-value` barrel, whose export-map path is
unchanged — so no other importer needs editing.

## The (a)-tighten ripple — an asymmetric peel

Tightening the params surfaced two callers in `diagnosis.ts`, handled
asymmetrically by where the value's `unknown`-ness genuinely originates:

- **Read side (`transactionReadInvariants`) — peeled clean, no cast.** Its
  return type tightened from `value: unknown` to
  `value?: Immutable<FabricValue>`, because the underlying read detail was
  already that type — the `unknown` was just unpropagated, not a real boundary.

- **Write side (`findDifferingWriteKeys`) — keeps a documented cast.** The
  write values arrive via `captureTransactionWrites()`, whose structural
  unwrap discards the `FabricValue` typing, so they genuinely reach the
  comparison as `unknown`. The two `as FabricValue` casts there assert what
  the type system can't verify across that boundary, and now carry a
  documenting comment to that effect. The proper long-term fix — a runtime
  `isFabricValue()` predicate at that boundary — is filed as a separate
  follow-up (TASK-0051); it is intentionally out of scope here.

## Verification

- All green locally (runner suite passing, data-model unchanged-green;
[...]

PR #4310, opened 2026-06-23
Type away the diagnosis.ts write-side as-FabricValue casts

danfuzz:
## Summary

Type away the two `as FabricValue` casts at the `findDifferingWriteKeys`
`valueEqual` call in `diagnosis.ts` (TASK-0053, off Remy's cast-typeability
report). A pure typing change — no runtime branch added — that removes the
casts the #4301 write-side comment had earmarked for an eventual runtime guard,
by showing the `unknown` they bridged was never intrinsic to the data.

## What the `unknown` actually was

Every path feeding the captured write/read maps is a `FabricValue` by declared
contract (`TransactionWriteDetail.value` / `IAttestation.value` are
`Immutable<FabricValue>`; the v2 write-detail and read paths are `FabricValue`;
the error sentinels are `string`). The `unknown` came from two local typing
choices in `unwrapTransactionDetailValue`:

1. an `unknown` parameter, and
2. an `isRecord()` narrowing whose guard (`value is Record<…, unknown>`)
   **erases** the value type.

## The fix

Replace the value-erasing `isRecord` check with a new **`isFabricObject`** guard
targeting the plain-record arm of `FabricValue` — `value is
Immutable<FabricObject>`: a non-null object that is neither an array nor a
`FabricSpecialObject`. That preserves the value type through the nested-`{value}`
envelope-strip (via `FabricObject`'s string index of `FabricValue`). With the
guard in place, tighten the unwrap signature, the captured `Map<string, …>`
containers, the `errorValue` slot, the `DiagnosisRecord.read/writeValues`
fields, and `findDifferingWriteKeys`' params to `FabricValue` /
`Immutable<FabricValue>`, and **drop both casts** (and the now-obsolete #4301
comment that pointed at a runtime-guard endgame).

## Soundness nuance — for Flux/Dan's explicit eye

This carries one behavior delta worth flagging (spec-faithful per Remy §5.2, not
a deviation): the old `isRecord` envelope-strip was **over-broad** — it would
have stripped a `.value` off an array or a `FabricSpecialObject` (e.g. a
`FabricEpoch*`'s `.value` getter). The new `isFabricObject` guard strips only
plain records, which is the envelope-strip's actual intent. So this is a
**latent-soundness improvement, not a regression**: the green suite confirms
nothing relied on the old over-broad strip. Calling it out explicitly so it gets
reviewed as a deliberate, intended narrowing rather than an incidental one.

## Scope

Two files (own-commit, merge-base already current `origin/main` — no sync, clean
3-dot scope):

- `packages/runner/src/scheduler/diagnosis.ts` — the guard + the tightening +
  cast/comment removal.
- `packages/runner/test/scheduler-pull-idempotency.test.ts` — one locally-
  constructed `Map<string, unknown>` tightened to `Map<string, FabricValue>` to
  match the signature (its `undefined` value is a valid `FabricValue`).

Disjoint from TASK-0052's `valueEqual.ts` — no overlap with PR #4309.

## Verification

- Runner suite green (656/3343 passed); `deno check` on the full runner package
  clean; fmt/lint clean.
- Zero cross-package importers of the tightened symbols.

---
## Summary by cubic
Removes the two `as FabricValue` casts in `diagnosis.ts` by adding `isFabricPlainObject` to `@commonfabric/data-model` and tightening types so read/write maps are `FabricValue` end-to-end. No runtime change; the `{ value }` unwrap now applies only to plain records (addresses TASK-0053).

- **Refactors**
  - Added `isFabricPlainObject` in `data-model/src/type-check.ts` with `FabricObject`/`Immutable<FabricObject>` overloads; uses shared `isPlainObject`; exported via the `fabric-value` barrel with unit tests.
  - Updated `diagnosis.ts` to use `isFabricPlainObject`; tightened `unwrapTransactionDetailValue` and read/write maps to `Immutable<FabricValue>`/`FabricValue`; removed both casts and the obsolete comment.
  - Behavior preserved: unwrap restricted to plain records; updated `scheduler-pull-idempotency` test to `Map<string, FabricValue>`.

<sup>Written for commit 45cfc9acc6518044f013b7b409229eedbf796971. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4310?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


danfuzz:
**reviewer-flux — APPROVE** (concurrent review, at immutable head `2a17942db`)

Reviewed the TASK-0053 type-it-away of the `diagnosis.ts` write-side casts — the satisfying closure of the asymmetric split from #4301 (this eliminates the very `as FabricValue` I'd APPROVEd there as retained-with-reason, now that Remy's report proved the feeder paths). The change is sound and behavior-preserving.

**1. The `isFabricObject` guard (load-bearing) — sound.** `value is Immutable<FabricObject>` ← `typeof === "object" && !== null && !Array.isArray && !(instanceof FabricSpecialObject)` correctly narrows to the plain-record arm of `FabricValue`. I confirmed the §5.2 mechanism at-source: `FabricObject extends Record<string, FabricValue>` (interface.ts:215), so after the guard, `value.value` is typed `Immutable<FabricValue>` via the string index — which is exactly what makes `unwrapTransactionDetailValue`'s body type-check without a cast. `deno check` is green, confirming Remy's claim that a signature-only tighten fails but this guard succeeds.

**2. The intentional behavior delta — confirmed a latent-soundness improvement, not a regression.** I gave this real eyes since it's the one spot that could hide a real change. The old `isRecord` strip (`typeof === "object" && !== null` — it doesn't even exclude arrays) would unwrap a nested `.value` off **any** non-null object, including a `FabricSpecialObject` or array carrying a `"value"` member. Concretely: `FabricEpochDays` has a `get value(): bigint` (FabricEpochDays.ts:42), and a prototype getter IS detectable by `"value" in obj` — so the old code would have stripped a `FabricEpochDays` down to its raw `bigint`, losing the wrapper. That was always wrong (a special object is never a legitimate `{value}` envelope), so nothing *correct* could depend on it; the new guard fixes it by stripping plain-record envelopes only. The green runner suite confirms nothing in practice relied on the old over-broad strip. Net: the delta only ever changes a previously-incorrect outcome, never a correct one.

**3. Casts gone + comment removed.** Both `as FabricValue` casts dropped; the obsolete #4301 write-side comment removed (correctly — its premise no longer holds). Containers tightened consistently (`Map<string, unknown>` → `FabricValue` across `DiagnosisRecord`, `captureTransactionWrites`/`captureCommittedReads` returns, `errorValue`, `findDifferingWriteKeys` params). Type-checks clean; behavior-preserving.

**4. Optional-vs-non-optional — consistent with #4301.** The call site passes `Map.get`'s `FabricValue | undefined` into `valueEqual`'s non-optional `FabricValue` param — the same pattern I adjudicated SOUND on #4301 (`FabricValue` includes `undefined`). Still sound. The retyped test (`treats removed undefined writes as differing`) exercises exactly this `undefined` flow and passes.

**Tests:** the scheduler idempotency suite passes at the PR head (5/5, including the `undefined`-write case); the test-file change is a pure type-annotation retype (`Map<string, unknown>` → `FabricValue`), semantics unchanged.

**CI note — Performance Check failed, but here it's pure noise (unlike its sibling #4309).** The single OVER metric is `Package Integration Tests (runtime-client) +27%` — a job wall-clock for a package unrelated to `diagnosis.ts`/scheduler; the 18 CLOSE entries are also job durations; no scheduler/diagnosis bench moved. Critically, **coverage-debt for `packages/runner` is `+0` (OK)** — this PR introduces zero coverage debt (the guard + tightened types are fully exercised). So the perf failure is shared-runner variance, `NEW_PERF_BASELINE`-clearable (or a re-run may clear it). No substantive half, in contrast to #4309's coverage delta.

Clean closure of the asymmetric-cast story. **APPROVE** on the code; the merge gate just needs the noise-only Performance Check cleared.

— reviewer-flux (they/them)

packages/runner/src/scheduler/diagnosis.ts
```diff
@@ -67,24 +69,39 @@ export function makeAddressKey(addr: IMemorySpaceAddress): string {
   return `${addr.space}/${addr.id}/${addr.path.join("/")}`;
 }
 
-function unwrapTransactionDetailValue(value: unknown): unknown {
-  return isRecord(value) && "value" in value ? value.value : value;
+/**
```
danfuzz: Suggest:

* Move this function to `data-model/type-check`.
* Rename it to `isFabricPlainObject()` (we will align the type name with this eventually).
* Add unit tests for it.

packages/runner/src/scheduler/diagnosis.ts
```diff
@@ -67,24 +69,39 @@ export function makeAddressKey(addr: IMemorySpaceAddress): string {
...
+ * Narrows to the plain-record arm of `FabricValue` (`FabricObject`): a
+ * non-`null` object that is neither an array nor a `FabricSpecialObject`. Unlike
+ * a bare `isRecord()` check, this preserves the value type — `FabricObject`'s
+ * string index of `FabricValue` keeps `value.value` typed as a `FabricValue`.
+ */
+function isFabricObject(
+  value: Immutable<FabricValue> | undefined,
```
danfuzz: Unclear to me that `| undefined` is required, since `undefined` is included in `FabricValue` (and I think that should make it allowed as an `Immutable<FabricValue>`)

packages/runner/src/scheduler/diagnosis.ts
```diff
@@ -67,24 +69,39 @@ export function makeAddressKey(addr: IMemorySpaceAddress): string {
...
 
-function unwrapTransactionDetailValue(value: unknown): unknown {
-  return isRecord(value) && "value" in value ? value.value : value;
+/**
+ * Narrows to the plain-record arm of `FabricValue` (`FabricObject`): a
+ * non-`null` object that is neither an array nor a `FabricSpecialObject`. Unlike
+ * a bare `isRecord()` check, this preserves the value type — `FabricObject`'s
+ * string index of `FabricValue` keeps `value.value` typed as a `FabricValue`.
+ */
+function isFabricObject(
```
danfuzz: With the move to `data-model`, it probably makes sense to have overload function signatures for both with and without `Immutable<...>` type declarations.

packages/runner/src/scheduler/diagnosis.ts
danfuzz: Done in `26a8d1421`: the guard is moved to `packages/data-model/src/type-check.ts` (beside `isFabricValueLayer`), renamed to `isFabricPlainObject()`, and re-exported through the `fabric-value` barrel; `diagnosis.ts` now imports it from there. Unit tests added in `packages/data-model/test/type-check.test.ts`.

packages/runner/src/scheduler/diagnosis.ts
danfuzz: Done in `26a8d1421`: dropped the redundant `| undefined` — `unwrapTransactionDetailValue` now takes (and returns) `Immutable<FabricValue>`, since `undefined` is already a member of `FabricValue` (and so of `Immutable<FabricValue>`). The call sites pass values that are `FabricValue` by contract, so no `| undefined` is needed there.

packages/runner/src/scheduler/diagnosis.ts
danfuzz: Done in `26a8d1421`: `isFabricPlainObject()` now has two overload signatures — one over `FabricValue | undefined` narrowing to `FabricObject`, and one over `Immutable<FabricValue> | undefined` narrowing to `Immutable<FabricObject>` — with a single `unknown`-typed implementation. So callers get the right narrowing whether or not they hold an `Immutable<...>`.

packages/runner/src/scheduler/diagnosis.ts
danfuzz: Correcting myself — I re-read your anchor and you're right: this is the `isFabricPlainObject` guard param, and the `| undefined` should be dropped, since `FabricValue` already admits `undefined` (so an `Immutable<FabricValue>` covers it). My earlier reply mis-read the anchor as the `unwrapTransactionDetailValue` signature — apologies for the misdirection. Fix incoming; I'll confirm the commit here once it lands.

danfuzz:
**reviewer-flux — APPROVE (re-review)** at immutable head `26a8d1421`

Re-reviewed after Cedar's guard refactor responding to Dan's 5 inline comments (my prior APPROVE was anchored to the pre-refactor tip `2a17942db` and is superseded by this). The refactor is clean and behavior-preserving; the underlying cast-typeaway (which I APPROVEd before) is unchanged.

**The move is behavior-preserving.** The guard body is **byte-identical** across the move + rename — old `isFabricObject` (in `diagnosis.ts`) and new `isFabricPlainObject` (in `data-model/src/type-check.ts`) share the exact predicate: `typeof === "object" && !== null && !Array.isArray && !(instanceof FabricSpecialObject)`. No logic changed; it just relocated beside `isFabricValueLayer` and gained a clearer name.

**Wiring + barrel correct.** `fabric-value.ts` re-exports `isFabricPlainObject` alongside `isFabricValueLayer`; `diagnosis.ts` imports it from `@commonfabric/data-model/fabric-value`. `deno check` green across `type-check.ts`, `fabric-value.ts`, `diagnosis.ts`, and the test.

**Overloads narrow correctly, no unsound widening.** The two typed overloads (`(FabricValue | undefined) → value is FabricObject` and `(Immutable<FabricValue> | undefined) → value is Immutable<FabricObject>`) over an `(unknown): boolean` impl resolve cleanly — `deno check` is green with **no cast** in `unwrapTransactionDetailValue`, so the narrowed `value.value` is soundly assignable to the `Immutable<FabricValue>` return. The `(unknown)` signature is the implementation (not a public overload), so it isn't reachable in a contract-breaking way.

**Dropped `| undefined` is sound.** `unwrapTransactionDetailValue`'s param + return are now bare `Immutable<FabricValue>`. The sole caller passes `detail.value: Immutable<FabricValue> | undefined` (`TransactionWriteDetail.value?`), and that assigns because `FabricValue` admits `undefined` (interface.ts:202), so `Immutable<undefined>` = `undefined` collapses the union into `Immutable<FabricValue>`. Same principle as #4301's optional-return; `deno check` confirms it. No real `undefined` case is lost.

**New tests assert the actual contract.** `type-check.test.ts` covers `isFabricPlainObject`: true for plain + null-prototype objects; false for arrays, `null`, `undefined`, primitives, and both flavors of `FabricSpecialObject` (`FabricBytes` primitive + `FabricError` instance). The scheduler idempotency suite still passes (5/5, incl. the `undefined`-write case). 39 steps green locally.

**Optional note (not a gate, already surfaced to Dan):** the body has no prototype check, so over raw `unknown` `isFabricPlainObject` would return `true` for a `Date`/`Map`. Within its typed `FabricValue` domain this can't happen (a `Date`/`Map` isn't a `FabricValue`, so a valid-typed caller never passes one), so it's sound — the name just reads a touch stricter than the body. Fine as-is; a tightening would be an optional follow-up.

**CI green** at this tip, including the Performance Check (the earlier failure was noise and cleared on the re-spin).

The cross-package move + overloads land cleanly and the contract is preserved. **APPROVE.**

— reviewer-flux (they/them)

danfuzz:
**reviewer-flux — APPROVE (re-review)** at immutable head `730916959` (the synced final tip)

Re-reviewed #4310 at its true final tip after Bolt's clean main-sync (`b4c8f7fce` is an ancestor → fully synced/mergeable). My earlier #4310 approves (`98cdfd2f4` pre-refactor, `26a8d1421` pre-strengthen) are superseded; this is the one that counts. The guard is now genuinely *strengthened* (not a behavior-preserving move), and it's correct.

**1. The strengthened guard is correct.** `isFabricPlainObject`'s body is now `return isPlainObject(value)` (the shared helper: `proto === Object.prototype || proto === null`). This rejects arrays, `FabricSpecialObject`s, and other class instances (`Date`/`Map`/`RegExp`), and is true only for `{}` / `Object.create(null)` — exactly the intended tightening. `isPlainObject` **subsumes** the old `!instanceof FabricSpecialObject` (special objects' proto ≠ `Object.prototype`) and the old `!Array.isArray` (arrays' proto is `Array.prototype`), so the body is a clean single delegation with **no redundant leftover check**. The doc-comment accurately describes the prototype-based semantics.

**2. Load-bearing — genuine `FabricObject`s still narrow true.** A real plain-record `{value: …}` envelope is an `Object.prototype` literal, so `isPlainObject` returns true and `unwrapTransactionDetailValue` still strips it. The strengthening only newly-rejects out-of-domain inputs (Date/Map/special) reachable via unsound cast. The scheduler idempotency runner suite stays green (confirmed locally; and the CI `Runner Tests` 1–4 all pass).

**3. The `| undefined` drop is sound.** Both overload params are now bare `FabricValue` / `Immutable<FabricValue>`; the call site passes `detail.value: Immutable<FabricValue> | undefined`, which assigns because `FabricValue` admits `undefined` (interface.ts), and the `Immutable` caller resolves the `Immutable` overload (narrowing to `Immutable<FabricObject>`, so `value.value` stays `Immutable<FabricValue>` — no cast). `deno check` green across all 4 files.

**4. Main-sync introduced no regressions.** Verified current main is an ancestor; the merge left `type-check.ts` / `diagnosis.ts` exactly as expected (guard body, overloads, call-site, barrel wiring all intact). New `type-check.test.ts` covers the strengthening: Date/Map/RegExp → false, plus the retained plain/null-proto → true and array/null/undefined/primitive/both-special-flavors → false. 40 test steps green locally (type-check + scheduler idempotency incl. the `undefined`-write case).

**CI note — the 12 red jobs are a transient INFRA failure, not a code regression.** I read the logs: every failing job (CLI/Package/Pattern Integration + Pattern Unit) dies at the *setup* step — `🚀 Start Toolshed server: chmod: cannot access './common-binaries/cf': No such file or directory` (with `digest-mismatch: error` on the artifact download), before any test code runs. The upstream `Build Binaries` job passed (57s), so the `cf` binary built but its artifact failed to propagate/download to the dependent jobs. The code-validating jobs all pass: **`Check` ✓, `Test` ✓, `Runner Tests` 1–4 ✓, `Pattern Reload` ✓.** So this is artifact-plumbing flakiness (re-runnable), not a regression from the strengthened guard or the sync.

The strengthening lands cleanly and the contract is preserved. **APPROVE** on the code; the merge gate just needs the integration jobs re-run (infra, not code).

— reviewer-flux (they/them)

packages/runner/src/scheduler/diagnosis.ts
danfuzz: Resolved at `45cfc9acc`: the redundant `| undefined` is dropped from **both** `isFabricPlainObject` overload params (now `FabricValue` / `Immutable<FabricValue>`), per your note that `FabricValue` already admits `undefined`. The guard body was also strengthened to delegate to the canonical `isPlainObject()` (a real prototype check: `Object.getPrototypeOf(value) === Object.prototype || null`), so non-plain class instances (Date/Map/`FabricSpecialObject`/etc.) are correctly excluded.

(My earlier reply in this thread mis-read the anchor as the `unwrapTransactionDetailValue` signature — corrected in the follow-up above. Your comment was about the guard param, and it's now addressed there.)

PR #4313, opened 2026-06-23
Replace provider auth casts with typed availability

Hixie:
Provider auth patterns used nine casts through as any to connect auth discovery with provider data calls. They also used nullable handler parameters in a few places to handle missing credentials inside runtime fallback paths. That made it possible for data handlers to receive missing auth. It also made loading, sign-in, missing-scope, expired-token, and ready states harder to tell apart.

Introduce an AuthAvailability union for provider auth state. The ready case carries the writable auth cell used by provider clients. The other cases carry the state needed to render the right authentication UI. This follows the same shape as other state objects in the repository that use a status field to make each state explicit.

Update the shared auth manager and the Google and Airtable auth managers to return this explicit auth availability state. Update provider patterns so data operations receive non-null writable auth cells only when credentials are ready. Replace missing-auth console warnings and fallback handlers with reactive UI that explains why the provider UI is not available yet.

Keep Airtable importer data handlers inside the authenticated UI branch. The pattern body still has one terminal return, but fetchBases, fetchTables, and fetchRecords are only bound in the branch where auth has narrowed to the writable Airtable auth cell.

Update Airtable and Google pattern guidance so future generated patterns follow the same auth availability shape. Add transformer fixture coverage that preserves auth availability cells through schema injection.

---
## Summary by cubic
Replaced unsafe provider auth casts with a typed availability union and gated all Airtable/Google actions on a live auth cell. Fixed reactive auth bugs, tightened UI gating, and added an env helper, client factories, docs, and broad tests.

- **New Features**
  - Typed `AuthAvailability`/`AuthCell`; `AuthManagerBase<T>` returns `availability`, and Google/Airtable managers use it.
  - `GmailSendClient`/`CalendarWriteClient` are callable factories; token refresh lazily reads env via `getPatternEnvironment()`.
  - Implemented `getPatternEnvironment()` to use the browser origin or `http://localhost:8000`; added focused tests.
  - Replaced `Date.now()` with `safeDateNow()` in patterns; updated generator guidance to the union-cell ternary style.
  - Expanded tests across auth managers, Gmail/Calendar clients (incl. 401 refresh), importer prompt, and many patterns; added `packages/patterns/test/vnode-helpers.ts`; tuned `packages/patterns/deno.jsonc` test task for plain Deno tests.

- **Bug Fixes**
  - Fixed computed auth bug in `email-task-engine` by moving the token check into the computed callback.
  - `calendar-event-manager`: derive auth from `availability`, bind confirm only when ready, and disable Delete/RSVP until ready.
  - Google Docs importer: build the import button from `availability` and pass the writable auth cell to the handler to resolve the type mismatch.
  - Cleaned up JSX guards in the email pattern launcher; removed unused Google `Auth` imports; made the docs example self‑contained.
  - Refreshed `importer-prompt` embedded Airtable sources and JSX examples; fixed Gmail importer coverage fixture.
  - Formatted imports in `email-task-engine`.

<sup>Written for commit 8938c3a89092b4a47cbb0534fe022bae439025d9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4313?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/patterns/airtable/airtable-importer.tsx
```diff
@@ -436,204 +328,324 @@ export default pattern<Input, Output>(
...
-                <div
-                  style={{
-                    padding: "16px",
-                    backgroundColor: "#f8f9fa",
-                    borderRadius: "8px",
-                    border: "1px solid #e0e0e0",
-                  }}
-                >
+                {/* Table selection */}
+                {ifElse(
```
mathpirate: could maybe also be a bare ternary now? not 100% sure but i think the transformer should lower this now, it wouldn't have until recently

packages/patterns/airtable/airtable-importer.tsx
```diff
@@ -436,204 +328,324 @@ export default pattern<Input, Output>(
...
-                          margin: "0 0 12px 0",
                         }}
                       >
-                        {recordCount} records loaded
-                      </p>
+                        {loading ? "Loading..." : "Load Tables"}
+                      </button>
+                    </div>
+
+                    {ifElse(
```
mathpirate: same, could maybe be a bare ternary now

packages/patterns/airtable/airtable-importer.tsx
```diff
@@ -436,204 +328,324 @@ export default pattern<Input, Output>(
...
+                      >
+                        Click "Load Tables" to see tables in this base.
+                      </p>,
+                    )}
+                  </div>,
+                  null,
+                )}
+
+                {/* Fetch records */}
+                {ifElse(
```
mathpirate: once more. feel free to ignore these, it's just nice to have the pattern corpus reflect the latest transformer behavior when convenient to do so

packages/patterns/airtable/airtable-importer.tsx
```diff
@@ -436,204 +328,324 @@ export default pattern<Input, Output>(
...
+                          fontWeight: "500",
+                          fontSize: "14px",
                         }}
                       >
-                        <table
+                        {loading ? "Fetching..." : "Fetch Records"}
+                      </button>
+                    </div>
+
+                    {ifElse(
```
mathpirate: again perhaps bare ternary

packages/patterns/airtable/airtable-importer.tsx
```diff
@@ -436,204 +328,324 @@ export default pattern<Input, Output>(
...
-                >
-                  <strong>Error:</strong> {error}
-                </div>,
-                null,
-              )}
-            </div>,
-            null,
-          )}
+                {/* Error display */}
+                {ifElse(
```
mathpirate: ternary

packages/patterns/tools/importer-prompt.ts
```diff
@@ -817,12 +837,12 @@ export default pattern<Input, Output>(
           />
           <div>
             <div style={{ fontWeight: 500, fontSize: "14px" }}>
-              {ifElse(hasUserName, auth.user.name, auth.user.email)}
+              {ifElse(hasUserName, authValue.user.name, authValue.user.email)}
```
mathpirate: perhaps bare ternary (i'm not totally sure what the syntax context is here tho)

packages/patterns/tools/importer-prompt.ts
```diff
@@ -1587,14 +1625,13 @@ export default pattern<Input, Output>(
...
-      auth: authResult,
-      isReady,
+      availability,
       fullUI: authUI,
     } = AirtableAuthManager({
       requiredScopes: REQUIRED_SCOPES,
     });
 
-    const auth = authResult as any;
+    const auth = availability.state === "ready" ? availability.auth : null;
```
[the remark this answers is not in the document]
mathpirate: the transformer now transforms bare ternaries into `ifElse` in most AST positions, so this should be fine

packages/patterns/google/core/experimental/calendar-event-manager.tsx
```diff
@@ -315,6 +315,16 @@ const confirmOperation = handler<
     const op = pendingOp.get();
     if (!op) return;
 
+    if (!auth) {
```
mathpirate: this one's different than its sibling sites and i think the other ones are more correct - the bare nullity check here won't hit needs-login / missing-scopes / token-expired - we probably want `const auth = availability.state === "ready" ? availability.auth : null;` here like you have in the other places

nb that Create/Update buttons use disabled={!canCreate} and Delete/RSVP use disabled={processing} - consider whether to also gate the Delete/RSVP buttons more strongly, although that state of affairs predates this PR

packages/patterns/auth/auth-types.ts
```diff
@@ -37,12 +38,43 @@ export type AuthState =
...
+
 /** Complete auth info bundle returned by auth managers */
-export interface AuthInfo {
+export interface AuthInfo<TAuth extends OAuthAuthData = OAuthAuthData> {
   state: AuthState;
-  // deno-lint-ignore no-explicit-any
-  auth: any | null;
-  authCell: unknown;
+  availability: AuthAvailability<TAuth>;
+  auth: AuthCell<TAuth> | null;
```
mathpirate: i'm not sure how to conceptualize of the relative purposes of `availability.auth` vs `auth` vs `authCell` on this interface. might it be possible to trim them down to just one, probably `availability.auth` since you have such a nice type for it now?

packages/patterns/google/core/util/google-auth-manager.tsx
```diff
@@ -132,13 +132,26 @@ export const GoogleAuthManager = pattern<
...
     requiredScopes,
     accountType,
     debugMode,
     descriptor: GoogleAuthManagerDescriptor,
     createAuth,
   });
+
+  return {
+    auth: base.auth as GoogleAuthManagerOutput["auth"],
+    availability: base.availability as GoogleAuthManagerOutput["availability"],
```
mathpirate: these casts are obviously much nicer than `as any` and this is a big improvement! that said, i wonder whether we might be able to make AuthManagerBase generic over TAuth so we could drop these casts entirely... i'm not sure whether that'll work fully, feel free to ignore this comment and/or just give it one shot and leave it this way if it doesn't work right away

(nb if we do reduce down to just one auth field this will also become just one narrowing cast instead of three)

mathpirate:
letting claude write this one out since it captured what i wanted to say nicely:

One overall design note on the AuthAvailability shape: only loading forces `auth: null` — `needs-login/missing-scopes/token-expired` all carry a live cell. That's reasonable since the UI needs the cell to render account info and the refresh button, but it does mean "data handlers only get a cell when ready" is enforced by the consumer idiom (state === "ready" ? availability.auth : null), not by the type. The calendar-event-manager slip is exactly what that gap allows. Not asking to restructure the union — just noting that the safety rests on every consumer following the idiom, which is an argument for (a) finishing the calendar migration and (b) maybe a lint/codemod or a tiny readyAuth(availability) helper so the gate is one well-known call rather than a hand-written ternary in ~15 places.

mathpirate (approved):
this is an improvement over the status quo and seems to be correct - lgtm! all of my comments are optional to address now or later

seefeldb (approved):

Hixie:
All comments addressed; things that were suggested as follow-ups are queued in future agent workstreams.

PR #4324, opened 2026-06-23
ci: re-land Node 24 action bumps + fix intermittent binary-download failures

danfuzz:
## What

Re-lands the Node 24 action bumps from #4315 (reverted in #4323) **plus** the fix for the intermittent CI failures that prompted the revert.

Two commits:
1. **Reapply #4315** — bump `checkout`→v7, `upload-artifact`→v7, `download-artifact`→v8, `cache`→v6, `setup-node`→v6 (un-revert of #4323).
2. **Fix the artifact-download regression** that #4315 exposed.

## Root cause of the spurious red

The five `📥 Download built binaries` steps in `deno.yml` were **bare** — `uses: actions/download-artifact` with no `name:`/`path:` — and relied on the action creating a per-artifact `common-binaries/` subdirectory, then `chmod +x ./common-binaries/toolshed`.

`download-artifact@v4` created that subdirectory **unconditionally**. `@v8` only creates it when **more than one** artifact exists at download time. When `common-binaries` is the *only* artifact present, v8 extracts straight to the workspace root (`./toolshed`, `./cf`, …), so `./common-binaries/toolshed` never exists and the next step fails with `chmod: cannot access './common-binaries/toolshed': No such file or directory`.

Artifacts accumulate as sibling jobs finish, so this was **timing-dependent**: jobs that started before other artifacts were uploaded saw only `common-binaries` and broke; later jobs saw several artifacts, got the subdirectory, and passed. Hence the flaky-looking spread of failures across `Pattern Integration`, `Pattern Unit`, `CLI Integration`, and `Package Integration` jobs.

Evidence from the two runs (same v8 action):

```
1 artifact present:  Starting download of artifact to: /home/runner/work/labs/labs           -> FAIL
4 artifacts present: Starting download of artifact to: /home/runner/work/labs/labs/common-binaries -> PASS
```

## The fix

Specify `name: common-binaries` + `path: common-binaries` on all five download steps. With a `name`, `download-artifact` extracts directly into `path`, producing `./common-binaries/<binary>` deterministically regardless of how many artifacts exist. (As a bonus, each step now downloads only the artifact it needs instead of every artifact in the run.)

`coverage-comment.yml`'s download already specifies `name:`, so it was never affected.

https://claude.ai/code/session_012KQVZ5ucQ7wDLGVjjVAVo3

## Performance Baseline

This PR changes GHA actions versions, so the following accounts for a spurious perf check failure.

NEW_PERF_BASELINE: job: CLI Integration Tests (core-piece-values) = 71s

---
## Summary by cubic
Re-lands Node 24 GitHub Actions upgrades and fixes flaky binary downloads by making artifact extraction paths deterministic. CI is now on updated actions and test jobs are stable again.

- **Dependencies**
  - Upgraded `actions/checkout`→`v7`, `actions/upload-artifact`→`v7`, `actions/download-artifact`→`v8`, `actions/cache`→`v6`, `actions/setup-node`→`v6`.
  - Applied across workflows and composite actions (e.g., `benchmarks`, `cli`, `deno`, `deploy-*`, `evals`, `perf-regression`, `apt-install`, `deno-setup`).

- **Bug Fixes**
  - `actions/download-artifact@v8` stops creating a subfolder when only one artifact exists, breaking paths like `./common-binaries/toolshed`.
  - Added `name: common-binaries` and `path: common-binaries` to all binary download steps to enforce `./common-binaries/*` layout and only fetch the needed artifact.
  - Removes timing-based failures across CLI, package, and pattern integration/unit jobs.

<sup>Written for commit 3ad75a43fadb6a744d5b780198f7dbaa9812756c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4324?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/actions/apt-install/action.yml
+++ b/.github/actions/apt-install/action.yml
@@ -31,7 +31,7 @@ runs:
 
     - name: 💾 Cache apt packages
       id: cache
-      uses: actions/cache@v4
+      uses: actions/cache@v6
       with:
         path: ~/.cache/apt-deb/${{ inputs.cache-key }}
         key: apt-deb-${{ inputs.cache-key }}-${{ runner.os }}-${{ steps.norm.outputs.image }}-${{ steps.norm.outputs.id }}
--- a/.github/actions/deno-setup/action.yml
+++ b/.github/actions/deno-setup/action.yml
@@ -22,7 +22,7 @@ runs:
     # Cache version: bump the `deno-bin-v1` token to invalidate all toolchain caches.
     - name: 💾 Restore Deno toolchain ${{ inputs.deno-version }}
       id: deno-toolchain-cache
-      uses: actions/cache@v4
+      uses: actions/cache@v6
       with:
         path: ${{ runner.tool_cache }}/deno/${{ inputs.deno-version }}
         key: deno-bin-v1-${{ runner.os }}-${{ runner.arch }}-${{ inputs.deno-version }}
@@ -59,7 +59,7 @@ runs:
     # Cache version: bump to invalidate all caches (e.g., v2 -> v3)
     - name: 📦 Cache Deno dependencies
       if: inputs.cache == 'true'
-      uses: actions/cache@v4
+      uses: actions/cache@v6
       with:
         path: |
           ~/.deno
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -13,7 +13,7 @@ jobs:
       group: labs
     steps:
       - name: 📥 Checkout repository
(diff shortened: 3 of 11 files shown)


willkelly (approved):
Makes sense to me

PR #4335, opened 2026-06-23
refactor(utils): extract a shared two-level per-checker memo cache

Hixie:
The TypeScript pipelines memoize facts derived from a type or AST node against the TypeChecker that owns them. A fact depends only on the key's shape within its checker, so it is safe to cache per (checker, key). Five caches had each hand-rolled the same two-level `WeakMap<TypeChecker, WeakMap<Key, Value>>`: four in the transformer's call-kind detection (call kind, direct builder kind, reactive value, and reactive-collection provenance) and one in the schema generator's cell brand lookup. Each repeated the get-or-create of the inner map and the has/get/set memoization, and the call-kind copies additionally shared a local `ExpressionCache` type plus two helper functions.

Introduce a generic `TwoLevelWeakCache<Outer, Key, Value>` in @commonfabric/utils that owns the two-level structure. It exposes `memoize(outer, key, compute)` for the common compute-once case and `groupFor(outer)` for callers that cache at several interleaved return points — the two recursive call-kind resolvers, which also cache conditionally and so need the raw inner map. Both levels stay weak, so a whole group is collected once its checker is gone and individual entries are collected with their keys. Presence is tracked with `has`, so a memoized `undefined` is cached and not recomputed.

The helper is generic over both key levels rather than typed to `ts.TypeChecker`, which keeps @commonfabric/utils free of a typescript dependency; the consumers pin the concrete types at construction. Both key levels are constrained to `WeakKey`, matching `WeakMap`'s own bound. Behavior at all five sites is unchanged.

---
## Summary by cubic
Extracted a shared two-level weak memo cache and replaced five hand-rolled per-checker caches in the TypeScript pipelines. Behavior is unchanged, with less duplication and the same memory safety.

- **Refactors**
  - Added `@commonfabric/utils/two-level-weak-cache` providing `TwoLevelWeakCache<Outer, Key, Value>` with `memoize` and `groupFor`.
  - Swapped in the utility for caches in call-kind detection and cell-brand lookup.
  - Exported the module in `@commonfabric/utils` and added unit tests.
  - Both levels are weak; `undefined` is cached; grouping by checker prevents cross-checker reads.

<sup>Written for commit f00f80f25e49623af52533677cf5f53adedae4ec. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4335?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/schema-generator/src/typescript/cell-brand.ts
+++ b/packages/schema-generator/src/typescript/cell-brand.ts
@@ -1,4 +1,5 @@
 import ts from "typescript";
+import { TwoLevelWeakCache } from "@commonfabric/utils/two-level-weak-cache";
 import { traverseTypeHierarchy } from "./type-traversal.ts";
 
 export type CellBrand =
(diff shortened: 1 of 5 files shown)


mathpirate (approved):
lgtm tyvm!

PR #4340, opened 2026-06-24
Make calendar write client SES-safe

Hixie:
The Common Fabric checker rejected calendar-event-manager because its imported calendar write client did work at module load time. The client called getPatternEnvironment() at module scope, and the same module also exposed a class declaration. SES module verification only allows declarative module-scope forms, so importing the manager failed before the pattern could be checked.

Replace the class body with a function-created client object, move the environment lookup into the token refresh helper, and keep a constructable CalendarWriteClient export for existing callers. Remove the manager and write-client paths from the cfcheck exclusion list now that the affected shard can verify them.

---
## Summary by cubic
Make the calendar write client SES-safe by removing module-scope work, moving env lookups to runtime, and handling missing `Intl`/timers. This restores SES verification and lets `cfcheck` verify the manager and client paths; adds focused tests that cover behavior and edge cases.

- **Refactors**
  - Hardened date/timeout helpers to work without `Intl` or timers; batch pacing no-ops when timers are sandboxed.
  - Defined a local `Auth` type and pulled `Writable` from `commonfabric` to avoid module-load side effects, while keeping `createCalendarWriteClient()` and a constructor-compatible `CalendarWriteClient` export.

- **Tests**
  - Added `calendar-write-client.test.ts` covering create/update/delete/RSVP, 401 refresh + retries, batching progress, error paths, and SES globals.
  - Introduced `packages/patterns/test-import-map.json` and updated `packages/patterns/deno.jsonc` test task to use it; added a test-only `commonfabric` shim for `getPatternEnvironment()`.

<sup>Written for commit c69e1c8e6630a239fd7a001f1b5b0c4ef7610aec. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4340?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Hixie:
(I did not review this carefully; since the code wasn't compiling before and is compiling now, I figure it's got to be an improvement.)

PR #4343, opened 2026-06-24
fix(runner): re-queue reactive computes stranded by a commit conflict

Hixie:
A reactive action whose commit fails with a ConflictError was, since the change that pulled conflicts out of the reactive retry path, recovered solely by reader-dirty propagation: it stopped re-queuing the action and relied on the invariant that the write which caused the conflict also dirties the action's still-subscribed reads, so the scheduler re-runs it.

That invariant does not hold when the conflict is caused by a write that has already been delivered to this replica — the very write that triggered the run. The scheduler's trigger index is keyed on read paths, and the catch-up sync emits no notification for data the replica already holds (there is no diff to report), so no further reader-dirty arrives. The action is left clean-but-stale: its last committed value stands and nothing re-runs it.

This surfaced as a flaky pattern test under CPU load (the scrabble multi-user join guard). A participant's "name already taken" message was written by its handler and reflected by the game's output computed, but the assertion computed that read that output never re-evaluated, so it kept returning a stale result until the per-assertion timeout expired.

Restore the recovery path: on a ConflictError, re-arm the subscription, wait for the conflict's readyToRetry catch-up gate, then re-queue the action so it re-runs against the caught-up state. Reader-dirty propagation remains a redundant fast path. Conflicts still bypass the retry budget — a conflict is a wait for catch-up, not a failure — so a contended compute can neither exhaust the budget nor strand as a zombie against rolled-back data, which was the goal of pulling conflicts out of the retry path in the first place.

Update the docs and comments that still described the no-re-queue model: the two scheduler-cfc-trigger-reads steps that pinned it, the isConflictRejection docstring, the compute- and effect-conflict-recovery test framing, the pull-based-scheduler and scheduler-v2 specs, and the committed-write-backpressure reactive-path section.

---
## Summary by cubic
Fixes reactive computes/effects stranded after a commit conflict by re-arming the subscription, waiting for the conflict’s catch-up (`readyToRetry`), then re-queuing to run against fresh state. Keeps conflicts off the retry budget and recovers even when the catch-up is aborted.

- **Bug Fixes**
  - In `packages/runner/src/scheduler/action-run.ts`, on `ConflictError` restore trigger reads, resubscribe, await `readyToRetry` when present, swallow rejections and sync throws, then mark dirty, add to `pending`, and queue execution.
  - Removed the standalone `readyToRetry` helper; inlined handling to drop unreachable guards and improve coverage.
  - Tests: updated compute/effect recovery to assert re-queue and budget bypass; added cases for rejected/thrown `readyToRetry` and re-queue even when the budget is exhausted.
  - Docs/specs: clarified re-queue-after-catch-up as the recovery path and noted reader-dirty as a fast path only.

<sup>Written for commit c7267e90bf117e250a3a1427618c7e0035a349e1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4343?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Hixie:
Follow-up from #4210.

mathpirate (approved):
see #gideon-log in discord for discussion, this appears to me to be a correctness improvement so it's good to merge any time, my questions/potential follow-ups aside

PR #4348, opened 2026-06-24
refactor(runner): one exponential backoff curve, no immediate-retry c…

Hixie:
…ount

The previous fix special-cased the first few conflict retries to fire with zero delay (an immediateRetries count) before exponential backoff began, so that a transient conflict converged within the test harness's / UI's settle. That dual path was avoidable: the same fast-convergence falls out of a single exponential curve whose early steps are simply small.

Drop immediateRetries and lower baseDelayMs to 25/32 ms. computeBackoffDelayMs now applies to every conflict from the first, and the curve is 0.78, 1.56, 3.125, 6.25, 12.5, then 25 ms before the seventh attempt, doubling to the 1-second cap — identical to the old curve from the seventh attempt on, but with the leading steps filled in instead of zeroed. The first few delays are sub-5ms, so a stale-basis conflict that clears as soon as fresh confirmed state arrives still converges near-immediately; the delay only grows into real spacing once a conflict persists. requeueForBackoff always parks via notBefore again (no zero-delay branch), and the disposition logic loses its immediate/backoff fork.

The CommitBackpressurePolicy loses its immediateRetries field. Behavior for sustained contention (capped backoff, bounded retry window, terminal CommitConvergenceError) is unchanged. The unit suite replaces the immediate-fast-path test with a direct check of the default curve (sub-5ms early, 25 ms before the seventh attempt, doubling to the cap).

---
## Summary by cubic
Simplified commit backpressure by removing the immediate-retry path and using one exponential backoff curve that starts with sub-5ms delays, keeping transient conflicts convergent while reducing scheduler complexity. `baseDelayMs` is set to 25/32 ms; behavior under sustained contention is unchanged.

- **Refactors**
  - Use `computeBackoffDelayMs` from the first conflict; early delays are ~0.78, 1.56, 3.125, 6.25, 12.5 ms, then 25 ms before the seventh attempt, doubling to the 1s cap.
  - Always park retries via `notBefore`; remove the zero-delay branch and immediate/backoff fork in `events.ts`.
  - Logging always reports a backoff delay; tests updated to verify the default curve; docs updated accordingly.

- **Migration**
  - Remove `immediateRetries` from `CommitBackpressurePolicy` options. Drop this field from any runtime config.

<sup>Written for commit 8c8013901359c988b797d8036c9412d23ea4daf6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4348?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/development/committed-write-backpressure.md
+++ b/docs/development/committed-write-backpressure.md
@@ -56,22 +56,23 @@ The event-handler commit path classifies each commit result and acts on it
 (`packages/runner/src/scheduler/events.ts`, `classifyCommitDisposition`):
 
 - **Success** — done.
-- **Stale-basis `ConflictError`** — the backpressure path. A stale-basis conflict
-  usually clears the instant the fresh confirmed state arrives, so the first
-  `immediateRetries` retries (default 5) fire with no delay — the fast path the
-  runtime had before backoff existed, where a transient conflict converges within
-  a few rapid retries. This matters: the harness and the UI settle by waiting for
-  the event queue to drain, so spacing out a retry that would have cleared
-  immediately can keep a write from converging within a settle. Only once the
-  immediate retries are exhausted (sustained contention) does the event re-queue
-  parked via the existing `notBefore` mechanism with a capped exponential backoff
-  plus jitter (`scheduler/backpressure.ts`, `computeBackoffDelayMs`). Backoff
-  makes the scheduler slow down under contention instead of busy-looping; jitter
-  keeps concurrent writers contending for the same entity from retrying in
-  lockstep. The event keeps retrying for a bounded window (default 30 seconds),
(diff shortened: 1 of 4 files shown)


Hixie:
I meant to land this instead of the retry thing in the previous PR, but accidentally landed it early because CI was green and I got excited about landing something, without checking my notes.

Anyway this fixes that problem.

PR #4338, opened 2026-06-23
test(lunch-poll): add multi-browser convergence matrix

willkelly:
## Summary
- add a lunch-poll multi-browser convergence matrix integration test
- exercise multiple profiles joining, host rotation, option creation, and votes from every client
- assert eventual convergence across rendered UI, browser readCell state, runtime result cell, and memory/source cell
- emit structured convergence diagnostics on timeout

## Verification
- deno check packages/patterns/integration/lunch-poll-two-browsers.test.ts
- GIT_MASTER=1 git diff --staged --check before commit

## Notes
- this PR intentionally contains only the regression matrix test
- local product/runtime root-cause experiments remain uncommitted and are not included

---
## Summary by cubic
Adds a browser-driven lunch‑poll convergence matrix test to ensure vote counts and voter swatches stay in sync across clients and cells. Test-only change with structured diagnostics on failure.

- **New Features**
  - Adds a multi-browser integration test that covers join flow, host rotation, option creation, and all-profile voting.
  - Verifies convergence across rendered UI, debug VDOM (`UI`), runtime result cell, and memory/source cell.
  - Emits detailed diagnostics on timeout (per-page snapshots, VDOM swatch summary, runtime stats).
  - Includes a single-profile test for the first host flow with empty options.

<sup>Written for commit 294320d7ba5fdaeda8b268044cd6cd4a989b5713. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4338?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #4141, opened 2026-06-15
Keyed collections POC and lunch poll perf diagnostics

willkelly:
## Summary

- Adds a pattern-local keyed collection POC with baseline, manual-indexed, helper-backed, SQLite-pushdown, and benchmark fixtures.
- Defines a reusable `keyed-collection-v1` helper seam, `ViewPlan@1` sidecars, a pure reference executor, parity diagnostics, and `latestByCountDelta@1` semantics.
- Adds lunch-poll diagnostic variants plus a full-product-parity helper-backed variant while leaving `packages/patterns/lunch-poll/main.tsx` untouched.
- Documents perf findings and Memory v2 future-work direction for helper-authored view plans/runtime-maintained aggregates.

## Perf findings

- Helper-backed/indexed aggregate shapes keep the aggregate-facing graph flat where array aggregates grow with vote count.
- Full-parity lunch helper improves option-add cost at larger option counts, e.g. `10x5` add options improved from `254.03s` to `104.91s`.
- Concurrent vote rounds are not uniformly faster yet: `10x5` vote round 3 regressed from `17.74s` to `44.86s`, pointing at shared JSON write retry/conflict cost and the need for runtime/backend aggregate backing.
- Raw SQLite remains evidence for internal lowering/pushdown only, not the primary author surface.

## Notes

- Per request, the standalone HTML report is intentionally not included in the repo.
- `packages/patterns/lunch-poll/main.tsx` is preserved as the untouched baseline.
- Known non-blocking caveat: deletion/reset paths still rewrite container records because the current helper lacks delete-key primitives; hot option field writes and vote updates use keyed/narrow helper paths.

## Validation

- LSP diagnostics clean for keyed-collections, lunch-poll, and multi-runtime harness changes.
- `deno fmt --check` on changed docs/source/test files.
- `deno lint` on changed TypeScript files.
- `deno test packages/patterns/keyed-collections/latest-by-count-delta-v1.test.ts packages/patterns/keyed-collections/keyed-collection-v1.test.ts packages/patterns/keyed-collections/view-plan-executor-v1.test.ts packages/patterns/keyed-collections/view-plan-parity-v1.test.ts` — 23 passed.
- `deno task cf check` on all new keyed-collections/lunch-poll entrypoints with `--root packages/patterns --no-run`.
- `deno task cf test packages/patterns/keyed-collections/main.test.tsx --root packages/patterns --verbose` — 14 passed.
[...]

PR #4349, opened 2026-06-24
Fix/lunch poll runtime convergence

willkelly:
## Summary

This PR fixes the lunch-poll convergence bug that showed up as **All Options** rows flickering, temporarily missing voter swatches, or taking a long time to settle after multiple users/options/votes. The fix is not just cosmetic redraw work: the branch tightens several runtime convergence seams and then simplifies the lunch-poll All Options dataflow so the rendered rows share one ranked tally source of truth.

### Structured conflict identity + narrow retries

Memory v2 conflicts now preserve structured read identity (`conflictingRead`) plus `retryAfterSeq` / `readyToRetry()` metadata through the server and client. The runner uses that metadata to distinguish ordinary stale-read conflicts from the two cases where an immediate bounded retry is safe:

- conflicts on prepare-added CFC verifier reads, and
- conflicts overlapping the action's own output writes.

This preserves the #4210 boundary: ordinary reactive compute conflicts still recover through subscriptions / reader-dirty propagation rather than being blindly requeued. The important part is that convergence gets a precise recovery path without reintroducing broad `ConflictError` retry loops or zombie recomputations.

### Scoped alias metadata is preserved

Binding aliases now carry the source slot's declared scope metadata when serializing link schemas. That keeps `PerUser` / `PerSession` declarations self-describing across pattern boundaries, so the runtime continues reading/writing the intended scoped cell instead of accidentally collapsing scoped identities.

### List children refresh when captured params change

`map`, `filter`, and `flatMap` no longer refresh reused child runs only when the list index changes. Reused children that depend on captured `params` are rerun when those params change. Lunch-poll exercises this heavily: option rows can keep stable identities while the captured vote/user state changes underneath them.

### CFC verifier reads are narrower

CFC preparation now reads verifier metadata and result schemas at the relevant subpaths instead of reading whole documents. That removes unrelated conflicts from verifier preparation and makes the structured retry exception narrow enough to reason about.

### Keyed child reconciliation updates reused payloads

The HTML worker reconciler now treats a stable key as preserving identity/order, not as proof that the child payload is unchanged. Same-key children can still receive fresh VNode/source-cell payloads from parent recomputation, which prevents stale rendered swatches while preserving DOM identity.

## Lunch-poll dataflow change

The All Options overview in `packages/patterns/lunch-poll/main.tsx` now renders from the single ranked tally produced by `tallyOptions(options, votes, users)`:

```tsx
const ranked = tallyOptions(options, votes, users);
...
{ranked.map((tally) => ...)}
{tally.voters.map((vote) => ...)}
```

`tallyOptions()` materializes each option's green/yellow/red counts, voters list, and original option order before sorting by the lunch-poll ranking rules. This replaces the more churn-prone shape where each rendered row repeatedly performed nested `votes.filter(...).map(...)` and rank lookups over the shared vote set. The header count, row swatches, debug VDOM, runtime result cell, and backing memory cell now all derive from the same materialized tally.

### Lunch-poll focused checks

```bash
deno task cf check packages/patterns/lunch-poll/main.tsx --no-run
deno task cf test packages/patterns/lunch-poll/main.test.tsx
deno task cf test packages/patterns/lunch-poll/multi-user.test.tsx
deno task cf test packages/patterns/lunch-poll/lunch-stats.test.tsx
deno run -A packages/patterns/tools/lunch-poll-diagnose.ts --quick --skip-refresh
deno run -A packages/patterns/tools/lunch-poll-diagnose.ts --cases=1x2,3x5,10x5 --rounds=3 --skip-refresh
```

Expected diagnostic signals from `lunch-poll-diagnose.ts`:

- final case summaries show `converged=true`;
- `votes=[...]`, `options=[...]`, and `users=[...]` match across sessions;
- no `DIVERGED fingerprints=...` output appears;
- pending/dirty counts settle instead of growing indefinitely.

Local diagnostic signal from this branch: the 3x4 churn repro improved from roughly `34.4s / 2873` commit conflicts+reverts to roughly `15.1s / 437` after the All Options dataflow simplification, and a post-vote option-add probe converged with matching votes/options/users across sessions.

### Browser-level verification

```bash
deno task integration --filter "lunch poll browser vote summary"
# or directly:
deno test --v8-flags=--max-old-space-size=4096 --trace-leaks -A packages/patterns/integration/lunch-poll-two-browsers.test.ts
```

Expected browser signals:

- all joined profiles are visible;
- after all profiles vote all options, every All Options row has one swatch per profile;
- rendered swatches, debug VDOM swatches, runtime result cell, and memory cell agree;
- on failure, the browser test prints page snapshots plus convergence diagnostics for the runtime/memory state.

Optional stress amplification:

```bash
CFC_BROWSER_PROFILE_COUNT=4 deno test --v8-flags=--max-old-space-size=4096 --trace-leaks -A packages/patterns/integration/lunch-poll-two-browsers.test.ts
```

### Runtime regression checks

```bash
deno test -A packages/memory/test/v2-client-test.ts packages/memory/test/v2-server-test.ts
deno test -A packages/runner/test/scheduler-cfc-trigger-reads.test.ts
deno test -A packages/runner/test/map-op-by-identity.test.ts
deno test -A packages/runner/test/cfc-boundary.test.ts
deno test -A packages/html/test/worker-reconciler-cell-child.test.ts
deno test -A packages/runtime-client/integration/client.test.ts
```

## Risk boundaries

- Do **not** generalize this into retrying every `ConflictError`; the runner retry path is intentionally structured and narrow.
- #4210 behavior is intentionally preserved for ordinary reactive compute conflicts.
- Scoped alias changes make schema metadata self-describing; they do not rewrite the link's base address scope.
- Same-key child reconciliation may do more payload work when parent data changes, but that is the correctness requirement for convergent rendering.

NEW_PERF_BASELINE: coverage-debt: packages/memory uncovered lines = 1415 lines
NEW_PERF_BASELINE: coverage-debt: packages/runner uncovered lines = 8193 lines
NEW_PERF_BASELINE: coverage-debt: packages/patterns uncovered lines = 296686 lines
NEW_PERF_BASELINE: job: Pattern Integration Tests (2/4) = 252s


mathpirate:
Wilk — really nice work here, the convergence improvements are substantial and the browser repro is a great addition. I went through it closely (context: I owned #4210 and shepherded #4343, so the conflict-handler area is fresh for me). One part I'd hold, and the rest I think layers cleanly on top of `main`.

**The one concern: the `action-run.ts` conflict-handler change reverts #4343.**

This PR rewrites `watchReactiveActionCommit` so ordinary `ConflictError`s no longer re-queue (back to the #4210 "rely on reader-dirty" model), with bounded retry kept only for the two structured cases (prepare-added verifier reads, own-output-write overlaps). The matching `scheduler-cfc-trigger-reads` assertions (`does not requeue from message-only conflicts…`) and the `pull-based-scheduler` spec are reverted to match.

The problem is that #4343 landed to fix a *real, load-reproduced* strand: under contention, an ordinary compute can conflict on a write that's **already been delivered** to its replica, so the catch-up carries no diff, no reader-dirty fires, and the compute is stranded clean-but-stale. It was bisected to #4210 (0 failures / 96 runs at the parent → 23 / 64 with #4210) and shows up as the flaky `scrabble/multi-user` "Name … is already taken" guard. The no-requeue model can't recover that case structurally — it's a *temporal* gap, not a path-granularity one (#4220 already closed the path axis), so the structured `conflictingRead` matchers can't detect it. Re-queuing ordinary conflicts is what closes it.

**And I don't think you need this revert to get lunch-poll converging.** I A/B'd the conflict handler in your own `lunch-poll-diagnose.ts` harness, holding the rest of the PR constant: swapping in the #4210 *no-requeue* handler **still converges** (3×4, 3×5, and 10×5 / 6 rounds, ~1200 conflicts) — so the convergence win here is coming from your other changes, not the handler rewrite. Worth noting #4343 already re-queues your two structured cases too (it re-queues *every* conflict, off-budget), so the structured bounded-retry path is largely subsumed by it — the one behavioral delta your version adds is *not* re-queuing ordinary conflicts, which is exactly the strand.

**Suggested split:**

*Layer on top of #4343 (all independently valuable, no conflict with the strand fix):*
- the reconciler keyed-child payload refresh (`html/src/worker/reconciler.ts`) — the actual swatch-desync fix
- `map`/`filter`/`flatMap` captured-param reruns
- scoped-alias metadata preservation (`PerUser`/`PerSession`)
- the lunch-poll `tallyOptions` dataflow simplification
- the browser test itself 🙏

*Reconsider / drop (the #4343 revert):*
- the `watchReactiveActionCommit` conflict-branch rewrite + its `scheduler-cfc-trigger-reads` and spec changes (and, if those go, the `conflictingRead` plumbing that only feeds the matchers)

**Open question for you:** if you've seen browser-level evidence that #4343's *broad* re-queue specifically hurts lunch-poll (churn / double-runs / the swatch desync), I'd genuinely like to see it — that'd be a real reason to narrow it, and your structured `conflictingRead` identity is exactly the right tool to do that *safely* (skip re-queue only where provably covered) rather than reverting wholesale. My headless runs didn't show #4343 breaking convergence, but I didn't run the browser path across variants. Happy to pair on it.

---
<sub>🤖 Drafted by Claude (Opus 4.8) and posted on Gideon's behalf, with his review.</sub>

willkelly:
Closing — the convergence problem this targets no longer exists on main, and the real fix landed verbatim.

The two genuine rendering fixes are on main character-for-character via #4366 ("fix(html): stabilize VDOM list child settling", 4d5e53e0e):
- `packages/html/src/worker/reconciler.ts` has `reconcileReusedChild`, `sameCellForReuse` and the keyed-reuse call site; `worker/types.ts:211-212` carries `cell?: Cell<unknown>` with this branch's exact doc comment.
- `packages/html/src/main/applicator.ts` has `PendingChildInsert`, `pendingChildInserts`, `deferChildInsert`, `replayPendingChildInserts`, and boolean-returning `insertChild`/`moveChild`.

The rest resolved elsewhere: #4370 re-shaped All Options onto a ranked `computed(...)` source and added `lunch-poll-vote.test.ts`; #4288 (scheduler-v2) dissolved the runner half; and the `watchReactiveActionCommit` rewrite here effectively reverts #4343 to the #4210 model, which main deliberately kept.

Closes out the lunch-poll cluster alongside #4141, #4338, #4405 and #4663.

PR #4350, opened 2026-06-24
docs: default dev identity to `id new`; scope `implicit trust` to local-server operator

willkelly:
## Why

While voting in a poll on a shared toolshed, one developer showed up as a **different developer**. Root cause: both had logged into the shell by importing a CLI keyfile generated with the documented command:

```bash
deno run -A packages/cli/mod.ts id derive "implicit trust" > cf.key
```

`id derive "implicit trust"` is deterministic and the passphrase is a **public string baked into the codebase** (`const ROOT = "implicit trust"` in `toolshed/lib/identity.ts`; the default `IDENTITY_PASSPHRASE` / `OPERATOR_PASS`). So *every* developer who follows the docs gets the **identical** DID (`did:key:z6MksHnZ…`). On a single-user localhost that's a harmless convenience. On a **shared** server (a teammate's box, staging, anything on a tailnet) it collapses everyone into one principal — they appear as each other and overwrite each other's `PerUser` data.

The skills (`cf`, `pattern-deploy`, `fuse-workflow`) and the tutorial all defaulted to deriving this shared identity for the user's *own* key, and `cf`/`SHARED_IDENTITY` even recommend importing that key into the browser — which is exactly the path that triggered the incident.

## When is `id derive "implicit trust"` actually legitimate?

Only one case: deliberately acting **as the local dev server's own identity** — operator/admin tasks on your own localhost where the caller must match toolshed's dev identity (`add-admin-charm`, the background-charm-service operator, deploying system home patterns). That's a "be the server" case, not a "be a user" case, and never applies to the browser or to a shared/remote server. It's niche enough that the **skills shouldn't mention it at all**; it lives in the docs.

## What this PR changes

- **Skills now use `id new` only.** `cf`, `pattern-deploy`, and `fuse-workflow` default to a unique key and no longer mention `implicit trust` at all — nothing in an agent-facing skill steers you toward the shared identity.
- **Docs keep the full picture.** `id derive "implicit trust"` is documented as the local-server operator identity only, with explicit "never against a shared/remote server, never browser-import it there" warnings, plus a new **"When To Derive `implicit trust`"** section and warning banner in `SHARED_IDENTITY.md` (also `CONFIGURATION.md`, `LOCAL_DEV_SERVERS.md`, `tutorial/10`).
- **Tutorial fix:** `docs/tutorial/06` now uses `id new`, and a wrong note that blamed `id new` (rather than the `deno task cf` wrapper) for ANSI key-file corruption is corrected — `deno run -A packages/cli/mod.ts id new > file` is clean.
- **`.gitignore` now ignores `*.pem`** so an exported identity key (e.g. `cf.pem`) can't be accidentally committed (only `*.key` was ignored before).

Docs/comments only — no runtime or behavioral change. The server-side `OPERATOR_PASS="implicit trust"` dev instructions (which *must* match the local toolshed) are intentionally left as-is.

### DIFF
--- a/docs/tutorial/06-workflow.md
+++ b/docs/tutorial/06-workflow.md
@@ -28,9 +28,11 @@ runs Toolshed on `localhost:8000`; use `dev-local` for the shell, not
 `dev`). Then:
 
 ```bash
-# Derive a deterministic dev key. Use `id derive`, NOT `id new > file` —
-# the latter pollutes the key file with ANSI output.
-deno run -A packages/cli/mod.ts id derive "implicit trust" > cf.key
+# Create a unique dev key. Redirect from `deno run -A packages/cli/mod.ts`,
+# never `deno task cf` — the task wrapper prints ANSI preamble that corrupts
+# the key file. (Chapter 10 covers the shared `implicit trust` dev identity
+# and the narrow case where you'd derive it instead.)
+deno run -A packages/cli/mod.ts id new > cf.key
 
 export CF_IDENTITY=./cf.key
 export CF_API_URL=http://localhost:8000
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,8 @@ dist/
 .cf-harness-artifacts/
 claude-research.key
 *.key
+# Exported PKCS8/PEM identity keys (e.g. cf.pem) — never commit private keys
+*.pem
 .passphrase
 *passphrase*
 .page-agent-config.json
--- a/docs/development/CONFIGURATION.md
+++ b/docs/development/CONFIGURATION.md
@@ -143,14 +143,18 @@ process you're configuring.
 | CF CLI | `CF_IDENTITY` env or `--identity <path>` | _(none)_ | _(none — error if remote)_ |
 
 For local dev, all three default to the implicit-trust passphrase so they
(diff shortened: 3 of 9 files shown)


docs/tutorial/06-workflow.md
```diff
@@ -28,9 +28,11 @@ runs Toolshed on `localhost:8000`; use `dev-local` for the shell, not
...
 
 ```bash
-# Derive a deterministic dev key. Use `id derive`, NOT `id new > file` —
-# the latter pollutes the key file with ANSI output.
-deno run -A packages/cli/mod.ts id derive "implicit trust" > cf.key
+# Create a unique dev key. Redirect from `deno run -A packages/cli/mod.ts`,
+# never `deno task cf` — the task wrapper prints ANSI preamble that corrupts
+# the key file. (Chapter 10 covers the shared `implicit trust` dev identity
+# and the narrow case where you'd derive it instead.)
+deno run -A packages/cli/mod.ts id new > cf.key
```
Hixie, line 35: this is probably the load-bearing one

Hixie: (well, this and its projection into the skills below)

Hixie (approved):
LGTM

PR #4351, opened 2026-06-24
Perf/shared compile byte cache

Hixie:
This is intended to improve the runtime of the "Run pattern unit tests" job. Eventually it can also be used in the "Pattern Integration Tests" job but I am currently trying to add coverage to that job and this would conflict significantly with that effort so I decided to do that later.

---
## Summary by cubic
Add a process-wide compiled-module-byte cache to skip per-space reads and the TS transform/emit in pattern tests and CI shards, while still persisting closures per space for by-identity reloads.

- **New Features**
  - Introduced `ModuleByteCache` injection via `RuntimeOptions.moduleByteCache`; the compile path checks `getCompleteSet` before storage and populates via `putAll` after compile. Full hits still write the closure to the target space and skip per-module SES re-verification on the trusted path.
  - Added test-owned `ProcessModuleByteCache` and `createCompileByteCache` (content-addressed, byte-capped, snapshot/restore) with optional disk persistence via `CF_COMPILE_CACHE_FILE`; injected through `PiecesController.initialize`.
  - CI restores/saves a per-shard cache file with `actions/cache@v4`; the key includes `packages/ts-transformers/**`, `packages/js-compiler/**`, `packages/schema-generator/**`, `deno.json`, `deno.lock`, the shard, and the pattern source hash.
  - New tests cover cross-runtime reuse, snapshot/restore, eviction, and running a warm-compiled pattern to validate the trusted path.

- **Refactors**
  - Defined `ModuleByteCache` in `runtime.ts` and exported `ModuleByteCache` and `CompiledModuleArtifact` from `@commonfabric/runner`.

<sup>Written for commit 1f85ccb3b9f4cc4be43d79855356eb728012c282. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4351?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/deno.yml
+++ b/.github/workflows/deno.yml
@@ -515,6 +515,24 @@ jobs:
       - name: 🔍 Verify lock file & install dependencies
         uses: ./.github/actions/deno-install
 
+      # Cross-run compile cache: persist the per-shard compiled-module-byte cache
+      # so unchanged patterns reuse last run's emitted bytes (skip the
+      # type-check + transform + emit). The key folds in a hash of everything
+      # that shapes the emitted bytes — the js-compiler, the CF transformer, the
+      # schema-generator it bakes schemas from, the root deno.json compiler
+      # options (jsx/jsxImportSource), and the lockfile — so any change to what
+      # emits the bytes invalidates the whole file even when it leaves module
+      # identity and the manual runtime-version tag untouched. The pattern-source
+      # hash rolls the key while `restore-keys` still seeds from the most recent
+      # prior cache.
+      - name: ♻️ Restore/save compile byte cache
+        uses: actions/cache@v4
+        with:
+          path: ${{ github.workspace }}/.compile-cache/pattern-integration-${{ matrix.shard }}.json
+          key: cc-${{ hashFiles('packages/ts-transformers/**', 'packages/js-compiler/**', 'packages/schema-generator/**', 'deno.json', 'deno.lock') }}-${{ matrix.shard }}-${{ hashFiles('packages/patterns/**/*.tsx') }}
+          restore-keys: |
+            cc-${{ hashFiles('packages/ts-transformers/**', 'packages/js-compiler/**', 'packages/schema-generator/**', 'deno.json', 'deno.lock') }}-${{ matrix.shard }}-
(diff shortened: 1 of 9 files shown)


mathpirate:
Really like this — and the framing that it's a process-wide tier *in front of* the per-space durable cache, sharing the `(runtimeVersion, identity)` key, is clean. It's also the first automatic compile-fingerprint we've had (the durable cache only ever hand-bumped `v2`). One thing I'd fix before merge, in that spirit:

The `actions/cache` key folds in `ts-transformers/**` + `js-compiler/**` + `deno.lock`, but **`packages/schema-generator/**` is also a compile input** — `SchemaGeneratorTransformer` calls `createSchemaTransformerV2().generateSchema(...)` and bakes the schemas into the emitted bytes. A schema-generator-only change wouldn't bust the key, wouldn't change module identity, and wouldn't bump `v2`, so a shard would restore stale bytes and could green a schema-gen regression. Same story for root `deno.json` compilerOptions (`jsx`/`jsxImportSource`). Both are one-line additions to `key` + `restore-keys`. (The durable cache technically shares this blind spot via the manual-`v2` reliance, but since this key is meant to be the automatic fingerprint, worth closing here.)

Smaller things: the `trustedBodies:true` SES-skip comment is accurate for in-process bytes but not for disk-restored ones (those trust the CI file + key, not own-process provenance) — worth a clarifying line. And a test that actually *invokes* a pattern after a byte-cache hit would lock in that the SES-skip path yields a working pattern, not just a durable closure.

`.ts`-glob and the in-memory `runtimeVersion` are both fine as-is (content-addressing + the in-file key handle them). Nice work.

_Posted by Claude (Gideon's coding assistant) on Gideon's behalf._

Hixie:
addressed comments above and some others found locally

PR #4355, opened 2026-06-24
fix(ts-transformers): reject class creation in pattern context

Hixie:
The pattern-context validation rejects creating functions (arrow functions, function expressions, and function declarations) directly in a pattern body. Such bodies run later, outside the reactive graph, so a reactive value captured from the pattern body is read as a stale plain snapshot rather than tracked reactively.

A class expression or declaration created in a pattern body has the identical footgun: a method, getter, or setter that reads a reactive local captured from the pattern body sees a stale snapshot, because the class body runs later and outside the reactive graph. This case was not caught, so a class wrapping reactive state compiled with no diagnostic.

Extend the validation to also reject class expressions and class declarations created in a restricted (pattern or render) context, unless they are inside a compute wrapper (computed, action, lift, or handler), where running later is expected. The whole class node is flagged once, mirroring how function creation flags the whole function, and the visitor stops descending into the class members so they do not produce cascading diagnostics. The existing pattern-context:function-creation diagnostic type is reused with a class-specific message.

Add validation tests covering a class expression with a reactive- capturing method, a class declaration, a class inside computed() (which stays allowed), and a class at module scope (also allowed). Update the current-behavior spec and the debugging error reference to record that classes share the function-creation diagnostic.

---
## Summary by cubic
Extend `ts-transformers` pattern-context validation to reject class creation in pattern/render bodies, preventing stale reactive reads. Classes remain allowed inside compute wrappers and at module scope.

- **Bug Fixes**
  - Reject `class` expressions/declarations in restricted contexts; reuse `pattern-context:function-creation` with a class-specific message.
  - Flag the class node once and stop visiting members to avoid cascading diagnostics.
  - Allow classes inside `computed()`, `action()`, `lift()`, `handler()`, and at module scope.
  - Add validation tests for class expression/declaration and allowed cases.
  - Update debugging error reference and current behavior spec.

<sup>Written for commit e0f9dd0d4a5459fcd3e2c34d634f57e3e1f42c16. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4355?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mathpirate (approved):
excellent, appreciate the hardening

PR #4359, opened 2026-06-25
Key Cell.get cache by stable view

bfollington:
## Why

`Cell.get()` already has a transaction-local read-result cache, but it was keyed
by the normalized link object's identity. Two `CellImpl` wrappers for the same
logical cell view can carry equivalent link values with different object
identities, so they miss the cache and repeat the same schema traversal inside
one transaction.

Within a transaction with no intervening write, two reads of the same link
including schema, the same CFC label view, and the same read variant
(`traverseCells|synced`) produce the same value, the same registered reactive
reads, and the same CFC state. The stable view hash captures those
read-affecting inputs: `NormalizedLink` includes `schema`, `viewRef` adds
`cfcLabelView`, and the existing variant separates `traverseCells`/`synced`.
The transaction drops the whole read cache on any write, so a hit is only served
when nothing changed.

This preserves the original cache contract while allowing equivalent wrappers
to share the cached traversal result. The new `Map` retention is bounded by
reads-without-writes in one transaction, and writes replace the map wholesale.

## What Changed

- Key the transaction-local `Cell.get()` cache by a stable hash of the cell view
  instead of link object identity.
- Memoize that view hash on each `CellImpl` wrapper and recompute only when the
  link or CFC label view identity changes. The set side recomputes after
  traversal because `ensureLink()` can complete/replace the link during a read.
- Change the transaction cache from `WeakMap<object, ...>` to
  `Map<string, ...>`, while preserving whole-cache invalidation on any write.
- Add transaction-local cache diagnostics for hits, misses, sets, and retained
  entries so tests can assert hit/miss behavior.
- Broaden `cell-get-cache` coverage for equivalent wrappers, schema and CFC
  label separation, read variants, write invalidation, non-cacheable `sample()`,
  and cached `undefined` values.

## Test plan

- `mise exec -- deno fmt packages/runner/src/cell.ts packages/runner/src/storage/extended-storage-transaction.ts packages/runner/src/storage/interface.ts packages/runner/test/cell-get-cache.test.ts`
- `git diff --check`
- `mise exec -- deno test --allow-ffi --allow-env --allow-read --allow-write=/tmp,/var/folders --allow-run=git packages/runner/test/cell-get-cache.test.ts`
- `cd packages/runner && mise exec -- deno task test`
  - `ok | 669 passed (3434 steps) | 0 failed | 0 ignored (11 steps)`
- `mise exec -- deno task integration runner`
  - `ok | 13 passed | 0 failed`
- `mise exec -- deno task test`
  - `All tests passing!`

Perf validation from the proposal's worker-aware CDP capture harness on the same
`/ben-loom-dev-6` route, keeping raw JSONL out of git:

- Baseline `baseline-clean2.jsonl`: 26 slow traversals, 4,682ms total,
  p50/p90/max 159/231/281ms, dominant root 3,990ms across 22 traversals.
- P1 candidate `after.jsonl`: 11 slow traversals, 2,345ms total,
  p50/p90/max 226/236/279ms, dominant root 1,639ms across 7 traversals.
- Delta: -49.9% slow-traverse ms and -57.7% slow-traverse count.
- Route-level cache-effect proxy: 15 avoided slow traversals out of 26 baseline
  slow traversals, 57.7%. The saved capture predates the committed hit/miss
  diagnostics, so direct transaction hit/miss rates are covered by
  `cell-get-cache.test.ts` assertions instead.

I also attempted a fresh CDP rerun against cloned local memory stores from both
the main labs checkout and the live Loom vendor checkout. Those cloned stores
loaded only the sparse shell (`Untitled`) and produced 0 slow traversals, so I
did not use that non-representative sample for the aggregate comparison.

---
## Summary by cubic
Keys the `Cell.get()` transaction cache by a stable view hash (link + schema + CFC label view) instead of object identity. This avoids duplicate traversals within a transaction and improves read performance.

- **Bug Fixes**
  - Cache hits now work across equivalent `CellImpl` wrappers for the same view.
  - No cache collisions between different schemas, CFC label views, or read variants.
  - Correctly caches `undefined` values (distinct from a miss).

- **Refactors**
  - Use `Map<string, ...>` keyed by the stable view hash; drop the whole map on any write.
  - Memoize the view hash per wrapper; recompute when link or CFC label view changes.
  - Add per-transaction cache stats (hits, misses, sets, entries) for tests.
  - Import hash helper from `@commonfabric/data-model/value-hash`.

<sup>Written for commit c89907e57e68205abf94f45d1cc495ddf6c9731c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4359?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/cell.ts
+++ b/packages/runner/src/cell.ts
@@ -12,6 +12,7 @@ import {
   shallowFabricFromNativeValue,
   valueEqual,
 } from "@commonfabric/data-model/fabric-value";
+import { hashStringOf } from "@commonfabric/data-model/value-hash";
 import { codecOf } from "@commonfabric/data-model/codec-common";
 import {
   type EntityRef,
@@ -628,6 +629,11 @@ export class CellImpl<T extends FabricValue>
 
   // Self-reference for pattern SELF symbol support
   private _selfRef?: OpaqueRef<any>;
+  private viewRefHashCache?: {
+    link: NormalizedFullLink;
+    cfcLabelView: CfcLabelView | undefined;
+    hash: string;
+  };
 
   constructor(
     public readonly runtime: Runtime,
@@ -693,6 +699,18 @@ export class CellImpl<T extends FabricValue>
       link: this.link,
       cfcLabelView: this._cfcLabelView,
     };
+  }
+
+  private viewRefHash(): string {
+    const link = this.link;
+    const cfcLabelView = this._cfcLabelView;
+    const cached = this.viewRefHashCache;
+    if (cached?.link === link && cached.cfcLabelView === cfcLabelView) {
+      return cached.hash;
+    }
+    const hash = hashStringOf({ link, cfcLabelView } satisfies CellViewRef);
+    this.viewRefHashCache = { link, cfcLabelView, hash };
+    return hash;
   }
 
   /**
@@ -882,8 +900,10 @@ export class CellImpl<T extends FabricValue>
     // tx, and the same CFC state. Reuse the prior result when the tx supports
     // caching (the non-reactive sample() wrapper does not) and is still open.
     // The tx clears this cache on any write, so a hit only happens when nothing
-    // has changed since the last read. Keyed on the cell's link identity, which
-    // is stable per cell; `variant` separates reads that differ in options.
+    // has changed since the last read. Key by the stable value of the view ref
+    // (link + CFC label view), not link object identity, so equivalent CellImpl
+    // wrappers in the same tx can share the cached traversal result. `variant`
+    // separates reads that differ in options or synced state.
     const tx = this.tx;
     const cacheable = tx !== undefined &&
       tx.getCachedReadResult !== undefined &&
@@ -893,8 +913,9 @@ export class CellImpl<T extends FabricValue>
       // still goes through readOrThrow() and invalidates the prepared digest.
       tx.getCfcState().prepare.status !== "prepared";
     const variant = `${options?.traverseCells ?? false}|${this.synced}`;
+    const cacheKey = cacheable ? this.viewRefHash() : undefined;
     if (cacheable) {
-      const cached = tx.getCachedReadResult!(this._link, variant);
+      const cached = tx.getCachedReadResult!(cacheKey!, variant);
       if (cached !== undefined) {
         return cached.value as Readonly<StripDefaultBrand<T>>;
       }
@@ -919,8 +940,8 @@ export class CellImpl<T extends FabricValue>
     if (cacheable) {
       // Re-read this._link: validateAndTransform (via viewRef -> link) may have
       // run ensureLink() and replaced it with the completed link object, which
-      // is the identity subsequent get()s will key on.
-      tx.setCachedReadResult!(this._link, variant, value);
+      // is the identity subsequent get()s will hash.
+      tx.setCachedReadResult!(this.viewRefHash(), variant, value);
     }
     return value;
   }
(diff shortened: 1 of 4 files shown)


bfollington (commented):
**Review (Claude Code analysis, posted by request) — approve / adopt with watch-items, no blockers.**

I read the full diff and verified the memoization invariants against `cell.ts` source.

### Correctness — verified, not assumed
The hash-memo is the only genuinely new risk surface; it holds up:
- **`link` identity is stable.** `get link()` returns `this._link` (cell.ts:682), and every mutation site (`asSchema`, `ensureLink`, space/id completion at :740/:761/:813/:1872) **replaces** `_link` with a fresh `{...this._link, …}` object. So `cached.link === link` correctly hits on repeated reads and recomputes on change. Replaced, not mutated → sound.
- **`_cfcLabelView` is set once and cloned** (constructor only, `cloneCfcLabelView`, :655); never reassigned/mutated → stable identity for the wrapper's life.
- **GET/SET key consistency.** Computing `cacheKey` at the top of `get()` calls `this.link`, forcing `ensureLink()` before traversal, so the SET-side recompute lands on the same completed-link hash. First read misses (as before); subsequent reads hit. No behavior change vs. the old `_link` path.
- **Key completeness** is guarded by the right tests: schema-split, CFC-label-split, variant separation, write-invalidation (`entries → 0`), `sample()` non-cacheable, cached-`undefined`. 669 unit + 13 integration + full suite green.

### Watch-items (non-blocking)
1. The memo rests on an **unenforced invariant** (link/cfcLabelView replaced, not mutated in place). True everywhere today — worth a one-line code comment stating it so a future in-place mutation can't silently produce a stale hash.
2. **Per-wrapper hashing scales with wrapper count** and deep-hashes the link *including its (large, recursive) schema*. Net-positive here (−50%), but if it ever bites, the follow-up is reusing the interned `hashSchema` inside the link hash rather than re-hashing the schema subtree.
3. The perf delta is from the **original P1 capture**, not a fresh run with the committed hit/miss diagnostics (the cloned-store rerun loaded only a sparse shell → 0 slow traversals, correctly excluded). So validate end-to-end after adopt.

### Adoption gate (our side)
Because of #3, real validation is post-adopt in the live env: `loom vendor sync → test → adopt`, then re-run the worker-aware capture on a populated `/ben-loom-dev-6` route (the new tx diagnostics now give a deterministic hit-rate) to confirm ~−50% holds; and sanity-check cold-start routes still load.

Scope note: this caps the **repeat multiplier** (~−50%); the per-traversal floor (don't-inline-the-tree) remains a separate, larger effort.

bfollington (commented):
cf-review via GPT-5, on behalf of @bfollington.

I reviewed PR #4359 at c89907e57. Coverage: deep-read all 4 changed files plus the hash, schema, CFC-label, and transaction-wrapper paths that the cache-key change depends on.

No blocking findings. The implementation uses the canonical @commonfabric/data-model/value-hash helper, includes schema and CFC label view in the stable key, preserves read-variant separation, bypasses the cache after CFC prepare, and avoids cache participation through sample() because transaction wrappers do not expose the optional cache methods.

Checks run:
- deno test --allow-ffi --allow-env --allow-read --allow-write=/private/tmp/deno-cache-labs-pr-4359,/tmp,/var/folders --allow-net=github.com,release-assets.githubusercontent.com --allow-run=git packages/runner/test/cell-get-cache.test.ts
- deno fmt --check packages/runner/src/cell.ts packages/runner/src/storage/extended-storage-transaction.ts packages/runner/src/storage/interface.ts packages/runner/test/cell-get-cache.test.ts
- deno lint packages/runner/src/cell.ts packages/runner/src/storage/extended-storage-transaction.ts packages/runner/src/storage/interface.ts packages/runner/test/cell-get-cache.test.ts
- git diff --check c3854bedf9d3780c58ab40945dfca7dd03be705c...HEAD

Verdict: clean to merge from this review pass.

seefeldb (approved):

PR #4361, opened 2026-06-25
feat(toolshed): gated per-connection memory-write trace

mathpirate:
## Why

Diagnosing write storms / conflict cascades — e.g. the lunch-poll fresh-vs-resume cell-identity storm fixed in #4360 — needs a server-side view of *which client* writes *what* to *which cell*. This upstreams the gated trace that made that diagnosis possible, cleaned up for general reuse (these contention investigations recur).

## What

`CF_DEBUG_MEMORY_WRITES=1` on the toolshed logs one `[memwrite]` line per committed memory op:

```
[memwrite] c=1 op=set id=of:fid1:… scope=space vhash=109967bf paths=[]
```

- **`c=<n>`** — per-connection ordinal. Each WebSocket is one client, so this attributes a storm's writes to specific clients (the key lever: is a divergent link target two clients each minting a different id, or one client alternating between branches?).
- **`vhash`** — stable, cross-process content hash of the written value (FNV-1a over key-sorted JSON). Equal content → equal hash, so a *different* vhash means a genuinely divergent value.
- **`paths`** — first few JSON-patch paths for `patch` ops.

Off by default; the gate is a cheap boolean check.

## Privacy

Raw values are **not** logged by default — only `vhash`. `CF_DEBUG_MEMORY_WRITE_VALUES=1` additionally appends `val=…`; avoid on real data, as cell values can hold user content, PII, or secrets. The hash alone distinguishes "same value" from "different value", which is all a divergence investigation needs.

## Design notes

- The pure formatting (`stableStringify` / `memwriteHash` / `formatMemWriteTrace`) is extracted into `memwrite-trace.ts` and unit-tested; the route keeps only env reads + the per-connection counter + message parsing.
- Kept as a gated `console.error` trace (greppable for ad-hoc debugging) rather than structured/pino output. The memory route already runs under `--unstable-otel`, so promoting this to otel events/metrics is a reasonable future enhancement — intentionally out of scope here.
- Documented under `docs/development/debugging/README.md` → Runtime Inspection.

---
## Summary by cubic
Adds a gated, per-connection memory-write trace in the Toolshed v2 memory route to diagnose cross-client write churn. Off by default; logs one line per committed write with the connection id and a canonical Fabric value hash.

- New Features
  - Enable with `CF_DEBUG_MEMORY_WRITES=1`; add `CF_DEBUG_MEMORY_WRITE_VALUES=1` to append `val=…`.
  - Each line includes `c=<n>` (per WebSocket), `op`, `id`, `scope`, `vhash` (12-char prefix from `@commonfabric/data-model/value-hash`), and first JSON-patch `paths`; values (when enabled) render via `@commonfabric/data-model/value-debug`.
  - Output goes to toolshed stderr (local dev: `packages/toolshed/local-dev-toolshed.log`). Pure formatting lives in `memwrite-trace.ts` with unit tests; the route handles env checks, per-connection counter, and message parsing.

- Bug Fixes
  - Emit the memwrite trace only after `connection.receive()` resolves, so failed receives aren’t logged as writes.

<sup>Written for commit 52b17cf7a158f7456f966b5176f0c321a002ed2b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4361?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## CI — coverage ratchet note (toolshed)

This PR adds **gated debug instrumentation** (`CF_DEBUG_MEMORY_WRITES`, off by default), so the `logMemWrites` route wrapper's body only runs when the flag is set — it is *inherently* never exercised in CI, and the toolshed coverage-debt ratchet flags it. The pure logic (`stableStringify` / `memwriteHash` / `formatMemWriteTrace`) is **fully unit-tested** in `memwrite-trace.test.ts` (100% line coverage); only the env-gated route wrapper is uncovered. Accepting the toolshed metric narrowly:

NEW_PERF_BASELINE: coverage-debt: packages/toolshed uncovered lines = 4919 lines

### DIFF
--- a/docs/development/debugging/README.md
+++ b/docs/development/debugging/README.md
@@ -88,9 +88,11 @@ inside computed(); Stream subscribe doesn't exist; binding the whole item to
   - Reference tail covers logger counts/timing/baselines/flags and worker traces
 - **Server-side write trace** — set `CF_DEBUG_MEMORY_WRITES=1` on the toolshed to
   log every memory write as `[memwrite] c=<conn> op=… id=… scope=… vhash=…`,
-  per-connection-tagged (`c=`) so a cross-client write storm can be attributed to
+  per-connection-tagged (`c=`) so cross-client write churn can be attributed to
   specific clients. Add `CF_DEBUG_MEMORY_WRITE_VALUES=1` to also dump raw values
-  (avoid on real data — they can hold PII/secrets). See
+  (avoid on real data — they can hold PII/secrets). Output goes to the toolshed's
+  stderr — `packages/toolshed/local-dev-toolshed.log` under the local-dev scripts.
+  See
   [`memwrite-trace.ts`](../../../packages/toolshed/routes/storage/memory/memwrite-trace.ts).
 - [VDOM Debug Helpers](vdom-debug.md) - `commonfabric.vdom.*` VDOM tree inspection
 - [Logger Internals](../logger-internals.md) - Creating loggers in runtime code (`getLogger`, timing, flags)
--- a/docs/development/debugging/README.md
+++ b/docs/development/debugging/README.md
@@ -86,6 +86,12 @@ inside computed(); Stream subscribe doesn't exist; binding the whole item to
   - Starts with common tasks: read piece data, dump the rendered VDOM, diagnose
     churn, find dead handlers, watch values, agent-browser recipes
   - Reference tail covers logger counts/timing/baselines/flags and worker traces
+- **Server-side write trace** — set `CF_DEBUG_MEMORY_WRITES=1` on the toolshed to
+  log every memory write as `[memwrite] c=<conn> op=… id=… scope=… vhash=…`,
+  per-connection-tagged (`c=`) so a cross-client write storm can be attributed to
+  specific clients. Add `CF_DEBUG_MEMORY_WRITE_VALUES=1` to also dump raw values
+  (avoid on real data — they can hold PII/secrets). See
+  [`memwrite-trace.ts`](../../../packages/toolshed/routes/storage/memory/memwrite-trace.ts).
 - [VDOM Debug Helpers](vdom-debug.md) - `commonfabric.vdom.*` VDOM tree inspection
 - [Logger Internals](../logger-internals.md) - Creating loggers in runtime code (`getLogger`, timing, flags)
 
--- a/packages/toolshed/routes/storage/memory/memory.handlers.ts
+++ b/packages/toolshed/routes/storage/memory/memory.handlers.ts
@@ -4,6 +4,7 @@ import * as MemoryServer from "@commonfabric/memory/v2/server";
 import type * as Routes from "./memory.routes.ts";
 import { memoryServer } from "../memory.ts";
 import { createSpan } from "@/middlewares/opentelemetry.ts";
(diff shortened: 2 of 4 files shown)


mathpirate:
Addressed cubic's P2 (valid): the memwrite trace was emitted before `connection.receive()`, so a message whose receive fails would be traced as a write. Moved logging onto the receive's fulfillment so a failed receive isn't misattributed — `4bec8f4d2`.

willkelly (commented):
cf-review via Hephaestus (openai/gpt-5.5), on behalf of @willkelly.

PR #4361 review · deep-read 4 changed files + canonical hash/debug modules · checks run: `deno test memwrite-trace.test.ts`, `deno check` touched memory files, `deno lint` touched memory files.

🔴 **Blocking**

- `packages/toolshed/routes/storage/memory/memwrite-trace.ts:37-64,86-89` — the PR hand-rolls `stableStringify()` + FNV-1a for `vhash` even though the parsed memory operations contain Fabric values (`Operation.value` / patch values), and this repo’s canonical Fabric value hashing is `@commonfabric/data-model/value-hash`.
  - Why it matters: the trace claims `vhash` distinguishes “same value” vs “different value”, but JSON-derived hashing can collapse or misrepresent Fabric values (`undefined` object fields, non-finite numbers, Fabric primitives/classes with no enumerable fields, etc.). That makes a write-storm diagnostic actively misleading: two different Fabric values can get the same trace hash, or the trace hash can disagree with the runtime’s actual value identity semantics.
  - Concrete fix: replace the custom `stableStringify` / `memwriteHash` path with `hashStringOf()` or `hashOf()` from `@commonfabric/data-model/value-hash` for the value source. If a shorter display hash is desired, truncate the canonical hash for display, but keep it derived from the canonical Fabric hash.

🟡 **Improvements**

- `packages/toolshed/routes/storage/memory/memwrite-trace.ts:100` — `CF_DEBUG_MEMORY_WRITE_VALUES=1` currently logs `stableStringify` output, not the repo’s debug rendering for Fabric values. If keeping value logging, consider `@commonfabric/data-model/value-debug` so non-plain Fabric objects don’t render misleadingly.

Verdict: **request changes** due to the duplicated / divergent Fabric hashing machinery.

mathpirate:
@willkelly good catch — confirmed your premise: the parsed ops are hydrated `FabricValue`s (`parseClientMessage` → `decodeMemoryBoundary` → `valueFromJson`), so the JSON `stableStringify` path could collapse distinct values into a false "same value".

Pushed `8556d68`:
- `vhash` now derives from `hashStringOf` (`@commonfabric/data-model/value-hash`) — the same identity `valueEqual` uses — truncated to a 12-char display prefix.
- raw `val=` now renders via `toCompactDebugString` (`value-debug`).
- `displayVhash` returns an `<unhashable>` sentinel rather than throwing, so an odd op can't abort the trace.
- dropped the hand-rolled `stableStringify`/`memwriteHash`; added regression tests for the undefined-field / non-finite-number cases plus crash-safety.

Thanks for the careful review.

PR #4364, opened 2026-06-25
fix: stabilize HTML list child settling

willkelly:
## Summary
- update the worker reconciler so reused keyed/cell child slots receive their new render payloads instead of retaining stale content
- replay deferred main-thread child insert/move operations once their nodes exist, while dropping stale placements
- preserve non-empty resumed map/filter/flatMap results while child result cells finish syncing, avoiding transient empty/partial list output

## Why
These are real DOM/VDOM discrepancies, not Lunch Poll-specific delays: the worker could reuse keyed child state without reconciling its replacement payload, and the main applicator could drop insert/move ops that arrived before their nodes. Separately, resumed list builtins could briefly overwrite synced list outputs with empty/partial arrays while child runs caught up. Together they explain the refresh-time All Options swatch disappearance/refill.

## Verification
- `deno fmt --check packages/html/src/worker/reconciler.ts packages/html/src/worker/types.ts packages/html/src/main/applicator.ts packages/html/test/worker-reconciler-cell-child.test.ts packages/html/test/worker-reconciler-diffing.test.ts packages/html/test/main-applicator.test.ts packages/runner/src/builtins/map.ts packages/runner/src/builtins/filter.ts packages/runner/src/builtins/flatmap.ts`
- `deno test --allow-ffi --allow-env --allow-read --allow-write=/tmp packages/html/test/worker-reconciler-cell-child.test.ts packages/html/test/worker-reconciler-diffing.test.ts packages/html/test/main-applicator.test.ts`
- `deno test --allow-ffi --allow-env --allow-read --allow-write=/tmp packages/runner/test/map-op-by-identity.test.ts packages/runner/test/patterns-dynamic.test.ts`
- browser probe on deployed minimal Lunch Poll: refresh goes shell/no poll -> 35 swatches, with no 35 -> 0 -> row-by-row refill transition

---
## Summary by cubic
Stabilizes HTML list child settling by reconciling reused keyed/cell children, replaying deferred child placements, and preserving resumed list results to eliminate refresh-time swatch flicker/disappear/reappear.

- **Bug Fixes**
  - Worker reconciler now updates reused keyed/cell children with new VNode payloads in place; replaces on tag change; updates text/props and children without remounts.
  - Main applicator queues insert/move ops when nodes aren’t ready and replays them once created; drops stale placements and clears pending on remove/dispose.
  - `map`, `filter`, and `flatMap` keep non-empty resumed results while element runs resolve, avoiding transient empty/partial list outputs.

<sup>Written for commit 7da0d0a4f673f819d167251449a656d9987f22d1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4364?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/html/src/worker/reconciler.ts
+++ b/packages/html/src/worker/reconciler.ts
@@ -1272,7 +1272,7 @@ export class WorkerReconciler {
    * Follows [UI] chains and returns the VNode, or null if not a VNode.
    * Includes cycle detection to prevent infinite loops.
    */
-  private extractVNode(node: WorkerRenderNode): WorkerVNode | null {
+  private extractVNode(node: unknown): WorkerVNode | null {
     if (isWorkerVNode(node)) return node;
 
     // Follow [UI] chain with cycle detection
@@ -2898,11 +2898,33 @@ export class WorkerReconciler {
       if (!forceReplace && state.children.has(key)) {
         // Reuse existing child
         const existingState = state.children.get(key)!;
-        newMapping.set(key, existingState);
+        const canReuse = this.reconcileReusedChild(
+          ctx,
+          existingState,
+          child,
+          visited,
+          policy,
+        );
         state.children.delete(key);
-
-        // Update if it's an element with new data
-        // (For now, we trust the key - updates happen through Cell subscriptions)
+        if (canReuse) {
+          newMapping.set(key, existingState);
+        } else {
+          existingState.cancel();
+          this.cleanupNodeHandlers(existingState);
+          this.queueOps([{ op: "remove-node", nodeId: existingState.nodeId }]);
+          hasNewChildren = true;
+          const childState = this.renderChild(
+            ctx,
+            child,
+            visited,
+            state,
+            key,
+            policy,
+          );
+          if (childState) {
+            newMapping.set(key, childState);
+          }
+        }
       } else {
         // Create new child, passing parent state and key for position tracking
         hasNewChildren = true;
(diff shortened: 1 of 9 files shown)


mathpirate (approved):
yay thank you for finding this! detailed agent comments incoming, i believe none are blocking though some should at least be addressed in a follow-up if not here

packages/runner/src/builtins/filter.ts
```diff
@@ -210,10 +212,15 @@ export function filter(
...
       // Truthy/falsy coercion, not strict boolean.
       const included = elementRuns.get(elementKey)!.resultCell.withTx(tx).get();
+      if (included === undefined) {
+        hasPendingPredicate = true;
+      }
       if (included) {
         newArrayValue.push(list[i]); // Original element cell reference
       }
     }
+    if (preserveResumeResult && hasPendingPredicate) return;
```
mathpirate, line 222: Two resume-timing edges here, both of which `map.ts` avoids because it kept the guarded clear (`map.ts:227`).

**(1) A transient empty-list first reconcile burns the await flag and re-clobbers the resumed list.** When the first resume reconcile resolves `list` to a *defined* `[]` (common with `Default<[]>`) while the synced result is non-empty: `preserveResumeResult` is true, but the empty loop never sets `hasPendingPredicate`, so this early-return doesn't fire → we reach `resumeBatchAwaitSync = false` (L223, burns it) and `resultWithLog.set([])` (clobbers the resumed list to empty). On the *next* reconcile when the real list streams in, `elementAwaitSync` is now false, so the preserve-guard can't re-engage and the per-element runs lose their `awaitSyncBeforeInitialRun` deferral → the flicker comes back. This is the "reconcile empty, then re-run once it streams in" race the `runner.ts` comment (~L1462) calls out, and the reason `map` kept the `list.length > 0` guard.

**(2) A legitimately-`undefined` child pins the whole resumed list.** `included === undefined` (L215) can't distinguish "not computed yet" from "settled to undefined." For `flatMap` especially, undefined-is-skip is documented-normal, so a single such element keeps `hasPending…` true for the entire resume batch → the guard never disarms → the output stays pinned to the resumed value and stops reflecting the *other* elements' updates. (Before, that element was just excluded and the rest of the list kept updating.)

Both would be covered by mirroring `map`: keep the guarded `if (list.length > 0) resumeBatchAwaitSync = false`, and treat empty-list-under-await as a preserve case rather than publishing `[]` over a resumed non-empty result while still in the await batch. Same applies to `flatmap.ts` (L228-229).

packages/runner/src/builtins/filter.ts
```diff
@@ -137,7 +136,11 @@ export function filter(
       ...(argumentUsage.usesParams ? { params: inputsCell.key("params") } : {}),
     });
 
-    if (resultWithLog.get() === undefined) {
+    const existingResult = resultWithLog.get();
+    const preserveResumeResult = elementAwaitSync &&
```
mathpirate, line 140: This resume/`awaitSync` preserve path has no direct test (the added tests are all html-layer and don't traverse resume). Given it's the riskiest change in the PR, a small runner regression test would be great insurance — e.g. resume a `filter`/`flatmap` with a non-empty stored result and a still-pending element predicate, and assert the stored result is preserved (not reset to `[]` or to a partial array). That'd also pin the two edges noted above once they're handled.

packages/html/src/worker/reconciler.ts
```diff
@@ -3168,27 +3283,21 @@ export class WorkerReconciler {
...
 
-                // Check children: if same, do nothing (sinks active);
-                // if different, tear down and rebuild
                 if (policyChildren.children !== undefined) {
-                  const childrenSame = this.areChildrenSame(
-                    childState.elementState,
-                    policyChildren.children,
-                  );
-                  if (!childrenSame || policyChanged) {
+                  if (policyChanged) {
```
mathpirate, line 3287: Minor / confirm-intent: with `areChildrenSame` dropped on this path, `resetTextIntegrityBoundary` is now gated on `policyChanged` only (same in the new `reconcileReusedChild`, ~L3067). For a `cf-cfc-authorship` boundary that goes dirty→clean under *unchanged* policy, nothing re-asserts `"ok"` on the reused child, so it can stay `"blocked"` even once the content is clean. It fails safe (over-blocks, never leaks), but `refreshBoundaryPolicyFromProps` still uses the paired `(!childrenSame || policyChanged)` gate (~L2000), so the reset sites now disagree — intended, or should they line up?

mathpirate (commented):
Took a careful pass through this — lovely fix. I stood up latest-main locally and confirmed the `send([])` removal kills the vote-flicker-on-reload (single-user: vote all options → refresh; clean now, no 35 → 0 → row-by-row refill). The `send([])` = `set([])` clobber-on-resume diagnosis is exactly right, and removing it is the load-bearing fix.

On your "not sure the runner fix is safe in all cases" — I think that instinct is well-founded, and it's narrower than it might feel. There are two resume-*timing* edges in `filter`/`flatmap` that `map` happens to avoid (both trace to filter/flatmap dropping `map`'s guarded `if (list.length > 0) resumeBatchAwaitSync = false` and moving the clear after the new early-return). Neither is what the manual repro hits — that's the non-empty-list path, which your preserve-guard fixes correctly — so they're edges, not the main case. Details inline.

The runner builtins are also the highest-blast-radius part here (the `map`/`filter`/`flatMap` contract every list pattern leans on) and currently have no runner-side test. If you wanted to, the commits split cleanly, so landing the html reconciler/applicator on their own and taking the runner change as its own PR (with a resume/`awaitSync` regression test) would be a viable way to isolate the risk — but it's equally fine to keep it as a single PR and just address whatever from these comments seems worth it to you. Happy to help with a test either way.

One minor, non-blocking note for the write-storm context: always calling `updateChildrenInPlace` (dropping the `areChildrenSame` skip) adds some worker→main op + `JSON.stringify` churn per reused inline-VNode child on hot lists — DOM writes themselves stay value-guarded in `setPropDefault`, so no extra DOM churn, and it's likely net-neutral-or-better since it removes spurious tear-down/rebuild. Just flagging since we've been chasing multi-client write-storm churn separately; a value-equality guard on the static-prop path could damp it later if it ever surfaces.

willkelly:
Closing this mixed DOM + runner PR in favor of two focused replacement PRs:\n\n- Runner list-resume fix: #4365\n- HTML/VDOM list child settling fix: #4366\n\nThis keeps the runtime resume semantics and the VDOM/applicator settling fixes reviewable independently.

PR #4365, opened 2026-06-25
fix(runner): preserve resumed list results while children sync

willkelly:
## Summary
- preserve existing resumed map/filter/flatMap result cells while child runs are still syncing
- avoid refresh-time transient empty/partial list outputs without changing converged list semantics

## Verification
- deno fmt --check packages/runner/src/builtins/filter.ts packages/runner/src/builtins/flatmap.ts packages/runner/src/builtins/map.ts
- deno test --allow-ffi --allow-env --allow-read --allow-write=/tmp packages/runner/test/map-op-by-identity.test.ts packages/runner/test/patterns-dynamic.test.ts
- LSP diagnostics clean for packages/runner/src/builtins/{map,filter,flatmap}.ts

---
## Summary by cubic
Preserves resumed `map`/`filter`/`flatMap` results while child runs sync to avoid transient empty or partial list outputs. Final list semantics are unchanged; this removes refresh-time flicker.

- **Bug Fixes**
  - Keep existing array results on resume when child predicates/elements are still pending; only update once available.
  - Remove eager `result.send([])` in `map`, `filter`, and `flatMap` scoped result cells.
  - Delay `resumeBatchAwaitSync` reset until after aggregation (and skip when preserving), so batches end only when children are accounted for.

<sup>Written for commit c21dcff9886b804d8b362f538e12524512a4cd48. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4365?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/builtins/filter.ts
+++ b/packages/runner/src/builtins/filter.ts
@@ -115,7 +115,6 @@ export function filter(
         tx,
       );
       result = scopedCell(runtime, tx, baseResult, outputScope);
-      result.send([]);
       // Link this cell to the parent cell
       setResultCell(result, parentCell);
       // Link the new result cells to the pattern cell too
@@ -137,7 +136,11 @@ export function filter(
       ...(argumentUsage.usesParams ? { params: inputsCell.key("params") } : {}),
     });
 
-    if (resultWithLog.get() === undefined) {
+    const existingResult = resultWithLog.get();
+    const preserveResumeResult = elementAwaitSync &&
+      Array.isArray(existingResult) &&
+      existingResult.length > 0;
+    if (existingResult === undefined) {
       resultWithLog.set([]);
     }
     if (list === undefined) {
@@ -153,10 +156,9 @@ export function filter(
       throw new Error("filter currently only supports arrays");
     }
 
-    if (list.length > 0) resumeBatchAwaitSync = false;
-
     const keyCounts = new Map<string, number>();
     const newArrayValue: any[] = [];
+    let hasPendingPredicate = false;
     for (let i = 0; i < list.length; i++) {
       // Skip sparse holes — don't create predicate runs for them
       if (!(i in list)) continue;
@@ -210,10 +212,15 @@ export function filter(
       // Read predicate result — creates subscription for reactivity.
       // Truthy/falsy coercion, not strict boolean.
       const included = elementRuns.get(elementKey)!.resultCell.withTx(tx).get();
+      if (included === undefined) {
+        hasPendingPredicate = true;
+      }
       if (included) {
         newArrayValue.push(list[i]); // Original element cell reference
       }
     }
+    if (preserveResumeResult && hasPendingPredicate) return;
+    resumeBatchAwaitSync = false;
     resultWithLog.set(newArrayValue);
 
     // NOTE: Same as map — elementRuns is not pruned. See map.ts for rationale.
--- a/packages/runner/src/builtins/flatmap.ts
+++ b/packages/runner/src/builtins/flatmap.ts
@@ -117,7 +117,6 @@ export function flatMap(
         tx,
       );
       result = scopedCell(runtime, tx, baseResult, outputScope);
-      result.send([]);
       // Link this cell to the parent cell
       setResultCell(result, parentCell);
       // Link the new result cells to the pattern cell too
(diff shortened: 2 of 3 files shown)


packages/runner/src/builtins/filter.ts
```diff
@@ -210,10 +212,15 @@ export function filter(
...
       // Truthy/falsy coercion, not strict boolean.
       const included = elementRuns.get(elementKey)!.resultCell.withTx(tx).get();
+      if (included === undefined) {
+        hasPendingPredicate = true;
+      }
       if (included) {
         newArrayValue.push(list[i]); // Original element cell reference
       }
     }
+    if (preserveResumeResult && hasPendingPredicate) return;
```
mathpirate, line 222: Two resume-timing edges here, both of which `map.ts` avoids because it kept the guarded clear (`map.ts:227`).

**(1) A transient empty-list first reconcile burns the await flag and re-clobbers the resumed list.** When the first resume reconcile resolves `list` to a *defined* `[]` (common with `Default<[]>`) while the synced result is non-empty: `preserveResumeResult` is true, but the empty loop never sets `hasPendingPredicate`, so this early-return doesn't fire → we reach `resumeBatchAwaitSync = false` (L223, burns it) and `resultWithLog.set([])` (clobbers the resumed list to empty). On the *next* reconcile when the real list streams in, `elementAwaitSync` is now false, so the preserve-guard can't re-engage and the per-element runs lose their `awaitSyncBeforeInitialRun` deferral → the flicker comes back. This is the "reconcile empty, then re-run once it streams in" race the `runner.ts` comment (~L1462) calls out, and the reason `map` kept the `list.length > 0` guard.

**(2) A legitimately-`undefined` child pins the whole resumed list — this is cubic's P1, and it's real.** `included === undefined` (L215) can't distinguish "not computed yet" from "settled to a real undefined." For `flatMap` especially, undefined-is-skip is documented-normal (`flatmap.ts:47`), so a single such element keeps `hasPending…` true for the whole resume batch — and because the flag-clear (L223) sits *behind* this early-return, the flag never clears, so the output stays pinned to the resumed value and stops reflecting the *other* elements' updates for the life of the action. (Before, that element was just excluded and the rest of the list kept updating.)

Both would be covered by mirroring `map`: keep the guarded `if (list.length > 0) resumeBatchAwaitSync = false`, and distinguish "not run yet" from "settled undefined" (e.g. gate pending-ness on the element run's sync/completion state) rather than inferring it from `result === undefined`. Same applies to `flatmap.ts` (L225 / L228-229).

mathpirate (commented):
Re-posting the runner half of the review here now that it's split out (was on #4364). Confirmed locally that the `send([])` removal kills the vote-flicker-on-reload — the `send([])` = `set([])` clobber-on-resume diagnosis is exactly right, and removing it is the load-bearing fix.

Two resume-*timing* edges remain that `map` avoids (inline on `filter.ts`; `flatmap.ts` is identical). The second is the same issue cubic flagged as P1 (the `undefined`-child case) — confirming that's real; the first (transient empty-list) cubic didn't catch. There's also no runner-side test for the new resume path yet. None of this is what the manual repro hits (that's the non-empty-list path, which the preserve-guard fixes correctly), so they're edges, not the main case.

packages/runner/src/builtins/filter.ts
```diff
@@ -137,7 +136,11 @@ export function filter(
       ...(argumentUsage.usesParams ? { params: inputsCell.key("params") } : {}),
     });
 
-    if (resultWithLog.get() === undefined) {
+    const existingResult = resultWithLog.get();
+    const preserveResumeResult = elementAwaitSync &&
```
mathpirate, line 140: This resume/`awaitSync` preserve path has no direct test (none in the PR traverse resume). Given it's the change with the widest blast radius (the `map`/`filter`/`flatMap` contract every list pattern uses), a small runner regression test would be great insurance — e.g. resume a `filter`/`flatmap` with a non-empty stored result and a still-pending element, and assert the stored result is preserved (not reset to `[]` or to a partial array). That'd also pin the two edges above once they're handled.

PR #4366, opened 2026-06-25
fix(html): stabilize VDOM list child settling

willkelly:
## Summary

This PR fixes two HTML/VDOM settling bugs that can make the rendered DOM temporarily disagree with the worker VDOM even when the underlying Fabric state is correct.

It is split out from the mixed DOM + runner PR. The separate runner resume fix is #4365.

### 1. Reused keyed children could keep stale payloads

The worker reconciler reused an existing child whenever the generated/stable key matched. That preserved DOM identity, but it also skipped reconciling the child payload supplied by the parent recomputation.

Concrete example:

```tsx
<div>
  {rows.map((row) => (
    <span key={row.id} data-vote-swatch-name={row.voter}>
      {row.voter}
    </span>
  ))}
</div>
```

If the parent recomputed a same-key child with fresh props or children, the worker could keep the old child state and not push the updated props/text/descendants. In Lunch Poll terms, an "All Options" row could preserve the right keyed slot but fail to reconcile the swatch/text payload that belongs in that slot.

User impact:

- rows or swatches can look stale after a reactive recompute;
- a refresh/resume can briefly show DOM that does not match the current worker VDOM;
- keyed identity is preserved, but content inside that identity may be wrong until another update happens to touch it.

Fix:

- same-key reuse now reconciles the payload in place when it is compatible;
- text children update through `update-text`;
- same-tag VNodes update props and descendants in place;
- incompatible same-key replacements, such as `span` -> `button`, remove and recreate the child instead of incorrectly reusing it.

### 2. Child placement ops could arrive before the referenced DOM position was attachable

The main-thread DOM applicator previously treated `insert-child`/`move-child` as effectively best-effort. If the child, parent, or `beforeId` reference was not ready at the exact moment the op arrived, placement could be dropped or applied as an append.

Concrete example:

```ts
insert-child(parent=1, child=2, beforeId=3)
create-element(3)
insert-child(parent=1, child=3, beforeId=4)
```

In that sequence, child `2` must wait until `3` exists **and** is attached under parent `1`; otherwise `2` can append early and produce the wrong visual order. This matters for mapped children and conditional children because the worker and main thread can receive/create nested nodes in multiple batches.

User impact:

- a list can appear in the wrong order during refresh;
- child UI can disappear and then refill row-by-row as later ops repair the tree;
- late-resolving mapped children, such as Lunch Poll vote swatches, can be skipped or placed after the wrong sibling;
- stale deferred placements must be discarded when a child or ancestor is removed, otherwise removed subtrees can influence future replay.

Fix:

- unresolved `insert-child`/`move-child` operations are deferred and replayed after later ops;
- placement waits for `beforeId` to be attached to the same parent instead of appending early;
- pending placements are discarded when the target child, parent, `beforeId`, or removed descendant subtree goes away;
- node/event cleanup still runs for removed descendants.

## Multi-browser / multi-session implications

This is not a browser-specific API compatibility issue; it is an ordering/timing issue between worker VDOM output and main-thread DOM application. Different browsers, tabs, reload timing, and machine speeds can interleave those batches differently.

That means two clients looking at the same durable state can temporarily see different UI:

- one browser may attach the reference node quickly enough and render the right order;
- another may receive/apply the placement before the reference is attached and briefly append or omit the child;
- after refresh/resume, one tab may show a full list while another shows a blank/refilling list even though both are reading the same cells.

The fix makes DOM convergence less timing-dependent: the applicator waits until a placement is actually valid, and the worker reconciles same-key payloads instead of relying on a later incidental update.

## Verification

- `deno fmt --check packages/html/src/main/applicator.ts packages/html/src/worker/reconciler.ts packages/html/src/worker/types.ts packages/html/test/main-applicator.test.ts packages/html/test/worker-reconciler-cell-child.test.ts packages/html/test/worker-reconciler-diffing.test.ts`
- `deno test --allow-ffi --allow-env --allow-read --allow-write=/tmp packages/html/test/worker-reconciler-cell-child.test.ts packages/html/test/worker-reconciler-diffing.test.ts packages/html/test/main-applicator.test.ts`
- Full workspace test/coverage run with `DENO_COVERAGE_DIR=/tmp/opencode/coverage-4366-vdom deno task test`
- `deno run --allow-read --allow-write --allow-run tasks/coverage-metrics.ts --profile-dir=/tmp/opencode/coverage-4366-vdom --root=$(pwd)` reported `coverage-debt: packages/html uncovered lines = 1456`, below the previous gate baseline of `1460`
- LSP diagnostics clean for the touched HTML source and test files

---
## Summary by cubic
Stabilizes VDOM list settling and authorship boundaries. Keyed children update in place, child placements wait for valid anchors, and text‑integrity state derives correctly and resets when disabled so clean content renders without remounts.

- **Bug Fixes**
  - Keyed reuse updates in place: texts via `update-text`; same-tag props/children diff; tag or UI-root changes replace; cell-backed children reuse only for the same cell (cell→literal replaces and cancels); UI-wrapped outputs update without reinserting.
  - `insert-child`/`move-child` defer until `beforeId` exists and is attached under the same parent; pending placements replay when valid, are dropped if the parent/child is removed or stale, and append if only the `beforeId` anchor is removed; pending state clears on dispose; removed subtrees still clean nodes and events.
  - Strict text-integrity/authorship: boundary state is derived from the subtree and stays blocked when only siblings change; it unblocks when content becomes clean; state clears when the policy is removed; per‑prop blocks are tracked and cleared on prop removal; slotted headers and split text update in place without remounting.

<sup>Written for commit 16065cc4c1c0da2c59e2b7e88f8ee356b968ebe9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4366?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/html/src/main/applicator.ts
+++ b/packages/html/src/main/applicator.ts
@@ -456,6 +456,13 @@ export class DomApplicator {
     const child = this.nodes.get(childId);
     if (!parent || !child) return false;
 
+    const beforeNode = beforeId === null ? null : this.nodes.get(beforeId);
+    if (
+      beforeId !== null && (!beforeNode || beforeNode.parentNode !== parent)
+    ) {
+      return false;
+    }
+
     this.discardPendingForChild(childId);
 
     // Update parent/children tracking
@@ -473,15 +480,9 @@ export class DomApplicator {
     }
     children.add(childId);
 
-    const beforeNode = beforeId !== null
-      ? this.nodes.get(beforeId) ?? null
-      : null;
-
-    if (beforeNode && beforeNode.parentNode === parent) {
-      // Only use insertBefore if the beforeNode is actually a child of parent
+    if (beforeNode) {
       parent.insertBefore(child, beforeNode);
     } else {
-      // Either no beforeNode, or it's not a child of this parent - just append
       parent.appendChild(child);
     }
     return true;
--- a/packages/html/src/main/applicator.ts
+++ b/packages/html/src/main/applicator.ts
@@ -513,10 +513,19 @@ export class DomApplicator {
   }
 
   private discardPendingForNodeIds(nodeIds: ReadonlySet<number>): void {
-    this.pendingChildInserts = this.pendingChildInserts.filter((pending) =>
-      !nodeIds.has(pending.parentId) && !nodeIds.has(pending.childId) &&
-      (pending.beforeId === null || !nodeIds.has(pending.beforeId))
-    );
+    const remaining: PendingChildInsert[] = [];
+    for (const pending of this.pendingChildInserts) {
+      if (nodeIds.has(pending.parentId) || nodeIds.has(pending.childId)) {
+        continue;
+      }
+
+      remaining.push(
+        pending.beforeId !== null && nodeIds.has(pending.beforeId)
+          ? { ...pending, beforeId: null }
+          : pending,
+      );
+    }
+    this.pendingChildInserts = remaining;
   }
 
   private replayPendingChildInserts(): void {
--- a/packages/html/src/main/applicator.ts
+++ b/packages/html/src/main/applicator.ts
@@ -20,6 +20,12 @@ const logger = getLogger("vdom-applicator", { enabled: false, level: "debug" });
 const ELEMENT_NODE = 1;
 const TEXT_NODE = 3;
 
+interface PendingChildInsert {
+  parentId: number;
+  childId: number;
+  beforeId: number | null;
+}
+
 function hasNodeType(node: unknown, nodeType: number): boolean {
   return typeof node === "object" && node !== null &&
     (node as { nodeType?: unknown }).nodeType === nodeType;
@@ -77,6 +83,7 @@ export class DomApplicator {
   private readonly runtimeClient: RuntimeClient;
   private readonly onError?: (error: Error) => void;
   private readonly setPropHandler: SetPropHandler;
+  private pendingChildInserts: PendingChildInsert[] = [];
 
   private rootNodeId: number | null = null;
 
@@ -98,12 +105,14 @@ export class DomApplicator {
     for (const op of batch.ops) {
       try {
         this.applyOp(op);
+        this.replayPendingChildInserts();
       } catch (error) {
         this.onError?.(
           error instanceof Error ? error : new Error(String(error)),
         );
       }
     }
+    this.replayPendingChildInserts();
 
     if (batch.rootId !== undefined) {
       this.rootNodeId = batch.rootId;
@@ -156,11 +165,15 @@ export class DomApplicator {
         break;
 
       case "insert-child":
-        this.insertChild(op.parentId, op.childId, op.beforeId);
+        if (!this.insertChild(op.parentId, op.childId, op.beforeId)) {
+          this.deferChildInsert(op.parentId, op.childId, op.beforeId);
+        }
         break;
 
       case "move-child":
-        this.moveChild(op.parentId, op.childId, op.beforeId);
+        if (!this.moveChild(op.parentId, op.childId, op.beforeId)) {
+          this.deferChildInsert(op.parentId, op.childId, op.beforeId);
+        }
         break;
 
       case "remove-node":
@@ -243,6 +256,7 @@ export class DomApplicator {
     this.nodes.clear();
     this.nodeParents.clear();
     this.nodeChildren.clear();
+    this.pendingChildInserts = [];
     this.rootNodeId = null;
 
     const elapsed = logger.timeEnd("dispose");
@@ -437,10 +451,12 @@ export class DomApplicator {
     parentId: number,
     childId: number,
     beforeId: number | null,
-  ): void {
+  ): boolean {
     const parent = this.nodes.get(parentId);
     const child = this.nodes.get(childId);
-    if (!parent || !child) return;
+    if (!parent || !child) return false;
+
+    this.discardPendingForChild(childId);
 
     // Update parent/children tracking
     // Remove from old parent if any
@@ -468,19 +484,63 @@ export class DomApplicator {
       // Either no beforeNode, or it's not a child of this parent - just append
       parent.appendChild(child);
     }
+    return true;
   }
 
   private moveChild(
     parentId: number,
     childId: number,
     beforeId: number | null,
-  ): void {
+  ): boolean {
     // Move is the same as insert - insertBefore handles it
-    this.insertChild(parentId, childId, beforeId);
+    return this.insertChild(parentId, childId, beforeId);
+  }
+
+  private deferChildInsert(
+    parentId: number,
+    childId: number,
+    beforeId: number | null,
+  ): void {
+    this.discardPendingForChild(childId);
+    this.pendingChildInserts.push({ parentId, childId, beforeId });
+  }
+
+  private discardPendingForChild(childId: number): void {
+    this.pendingChildInserts = this.pendingChildInserts.filter((pending) =>
+      pending.childId !== childId
+    );
+  }
+
+  private discardPendingForNodeIds(nodeIds: ReadonlySet<number>): void {
+    this.pendingChildInserts = this.pendingChildInserts.filter((pending) =>
+      !nodeIds.has(pending.parentId) && !nodeIds.has(pending.childId) &&
+      (pending.beforeId === null || !nodeIds.has(pending.beforeId))
+    );
+  }
+
+  private replayPendingChildInserts(): void {
+    if (this.pendingChildInserts.length === 0) return;
+
+    const remaining: PendingChildInsert[] = [];
+    for (const pending of this.pendingChildInserts) {
+      if (
+        !this.insertChild(
+          pending.parentId,
+          pending.childId,
+          pending.beforeId,
+        )
+      ) {
+        remaining.push(pending);
+      }
+    }
+    this.pendingChildInserts = remaining;
   }
 
   private removeNode(nodeId: number): void {
     const node = this.nodes.get(nodeId);
+    const removedNodeIds = new Set([nodeId]);
+    this.collectDescendantNodeIds(nodeId, removedNodeIds);
+    this.discardPendingForNodeIds(removedNodeIds);
     if (!node) return;
 
     logger.timeStart("remove-node", String(nodeId));
@@ -558,6 +618,16 @@ export class DomApplicator {
     }
 
     return count;
+  }
+
+  private collectDescendantNodeIds(nodeId: number, into: Set<number>): void {
+    const children = this.nodeChildren.get(nodeId);
+    if (!children || children.size === 0) return;
+
+    for (const childId of children) {
+      into.add(childId);
+      this.collectDescendantNodeIds(childId, into);
+    }
   }
 
   private setAttrs(nodeId: number, attrs: Record<string, unknown>): void {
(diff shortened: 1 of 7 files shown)


packages/html/src/worker/reconciler.ts
```diff
@@ -3168,27 +3283,21 @@ export class WorkerReconciler {
...
 
-                // Check children: if same, do nothing (sinks active);
-                // if different, tear down and rebuild
                 if (policyChildren.children !== undefined) {
-                  const childrenSame = this.areChildrenSame(
-                    childState.elementState,
-                    policyChildren.children,
-                  );
-                  if (!childrenSame || policyChanged) {
+                  if (policyChanged) {
```
mathpirate: Minor / confirm-intent: with `areChildrenSame` dropped on this path, `resetTextIntegrityBoundary` is now gated on `policyChanged` only (same in the new `reconcileReusedChild`, ~L3067). For a `cf-cfc-authorship` boundary that goes dirty→clean under *unchanged* policy, nothing re-asserts `"ok"` on the reused child, so it can stay `"blocked"` even once the content is clean. It fails safe (over-blocks, never leaks), but `refreshBoundaryPolicyFromProps` still uses the paired `(!childrenSame || policyChanged)` gate (L1999-2000), so the reset sites now disagree — intended, or should they line up?

mathpirate (commented):
Re-posting the html half of the review here now that it's split out (was on #4364). The reconciler change (reconcile reused keyed/cell children instead of trusting the key) and the applicator defer-and-replay both read as sound and target real DOM/VDOM discrepancies. One confirm-intent note on the text-integrity reset gate inline, plus a churn observation:

Churn (non-blocking, write-storm context): always calling `updateChildrenInPlace` (dropping the `areChildrenSame` skip on the cell-child path) adds some worker→main op + `JSON.stringify` churn per reused inline-VNode child on hot lists — DOM writes themselves stay value-guarded in `setPropDefault`, so no extra DOM churn, and it's likely net-neutral-or-better since it removes spurious tear-down/rebuild. Just flagging since we've been chasing multi-client write-storm churn separately; a value-equality guard on the static-prop path could damp it later if it ever surfaces.

packages/html/test/worker-reconciler-cfc-render-policy.test.ts
```diff
@@ -2078,6 +2078,72 @@ Deno.test("worker reconciler CFC render policy", async (t) => {
...
     );
 
+    await t.step(
+      "strict text integrity resets same-key child when content becomes clean",
+      async () => {
+        const collector = createOpsCollector();
+        const reconciler = new WorkerReconciler({
+          onOps: collector.onOps,
+        });
+        const rootCell = new MockCell({
```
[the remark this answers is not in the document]
willkelly: cubic found this possible ReferenceError, but this looks like a false positive. `verifiedText` and `unsignedReleaseText` are both declared in the enclosing `Deno.test` scope (`packages/html/test/worker-reconciler-cfc-render-policy.test.ts:223` and `:227`), so this nested `t.step` can close over them. I verified with LSP (no diagnostics) and a focused runtime check: `deno test --allow-ffi --allow-env --allow-read --allow-write=/tmp packages/html/test/worker-reconciler-cfc-render-policy.test.ts` passes, including `strict text integrity resets same-key child when content becomes clean`.

packages/html/src/worker/reconciler.ts
```diff
@@ -2973,6 +2995,102 @@ export class WorkerReconciler {
...
+    visited: Set<object>,
+    policy: RenderPolicy,
+  ): boolean {
+    if (isCell(child)) {
+      return childState.cell !== undefined &&
+        this.sameCellForReuse(childState.cell, child);
+    }
+
+    if (
+      childState.isText &&
```
mathpirate, line 3064: I think cubic's P1 here is a false positive — unreachable by construction. This text branch only fires when `childState.isText && child` is a primitive. For a Cell→literal transition to reach it, the Cell child and the literal would need the **same child key** — but keys are `JSON.stringify(node, cellReplacer)`: a Cell serializes to an object (`{id,path,space,…}` / SigilLink), a literal to a scalar, so they can never be byte-equal → `reconcileReusedChild`'s text branch is never entered for a Cell→literal pair. And a Cell→literal transition gives the literal a fresh disjoint key, so the old cell-origin child is swept by the obsolete-children loop → `cancel()` tears down its subscription. The branch only ever fires for literal→literal reuse, whose `childState.cell` is undefined — nothing to cancel. So no stale subscription survives; safe to dismiss.

mathpirate (commented):
Took another pass at the current head — this is close to merge-ready and a clear net improvement: the authorship-boundary reset genuinely fixes the stuck-at-blocked case, and the beforeId-deferral is the right shape for the wrong-order placement. cubic's open P1 looks like a false positive (inline). Two narrow, latent concerns the new commits introduced/revived, both with cheap fixes (inline) — neither blocking IMO. Details below.

packages/html/src/main/applicator.ts
```diff
@@ -457,30 +480,68 @@ export class DomApplicator {
...
+  private discardPendingForChild(childId: number): void {
+    this.pendingChildInserts = this.pendingChildInserts.filter((pending) =>
+      pending.childId !== childId
+    );
+  }
+
+  private discardPendingForNodeIds(nodeIds: ReadonlySet<number>): void {
+    this.pendingChildInserts = this.pendingChildInserts.filter((pending) =>
+      !nodeIds.has(pending.parentId) && !nodeIds.has(pending.childId) &&
+      (pending.beforeId === null || !nodeIds.has(pending.beforeId))
```
mathpirate: Heads-up: `925d7034`'s anchor-deferral (insertChild now defers instead of appending when the `beforeId` anchor isn't attached) changes the premise that made this safe before — a pending insert can now be keyed on a `beforeId`, and `discardPendingForNodeIds` drops it whenever its `beforeId` is in the removed set, **even when the child and parent are both still alive** → a still-valid child can be stranded (lost render).

Within a batch it's hard to hit (removes are emitted first, inserts anchor only on surviving siblings, and per-op replay rescues quickly). The reachable case is **cross-batch**: an async sibling cell resolves to `null` → `remove-node(anchor)` → the dependent insert is discarded, with no later `updateChildren` to re-emit it (same async-strand family as #4210/#4343 — won't repro in-process).

Cheap fix: when only the anchor is gone but child+parent survive, **degrade the pending insert to an append (`beforeId=null`) instead of dropping it**. Also worth noting no test covers the anchor-only discard — the existing "drops pending inserts that target removed descendants" test removes the parent too.

packages/html/src/worker/reconciler.ts
```diff
@@ -2973,6 +2995,102 @@ export class WorkerReconciler {
...
+    childState.elementState.sourceProps = sanitized.props;
+
+    this.updatePropsInPlace(ctx, childState.elementState, sanitized.props);
+
+    if (policyChildren.children !== undefined) {
+      const childrenSame = this.areChildrenSame(
+        childState.elementState,
+        policyChildren.children,
+      );
+      if (!childrenSame || policyChanged) {
```
mathpirate: Nice fix for the stuck-at-blocked case — verified it un-sticks the single-child path. One latent multi-child edge though: the reset fires on `!childrenSame`, which is true when **any** child's key changes (incl. a sibling). So a `cf-cfc-authorship` boundary with ≥2 children, where child A is a reused-but-still-blocked cell-text child (`sameCellForReuse` true → its sink doesn't re-fire, so it's never re-blocked) and a sibling B churns, would queue `"ok"`, re-render only B, and settle on `"ok"` → a **false verified-author badge over still-blocked content** (the blocked text itself stays hidden — a trust-*signal* bug, not a leak).

Not reachable in any current surface (today's only `verifyTextIntegrity` boundary wraps its content in a single stable-key `<div>`, so the boundary reset never fires there) — so latent, not urgent. To harden: re-derive the boundary state from descendants after the re-render rather than optimistically resetting to `"ok"`. The new test covers only single-child; a multi-child "stays blocked" assertion would guard it.

packages/html/src/worker/reconciler.ts
```diff
@@ -2973,6 +2995,98 @@ export class WorkerReconciler {
...
+    visited: Set<object>,
+    policy: RenderPolicy,
+  ): boolean {
+    if (isCell(child)) {
+      return childState.cell !== undefined &&
+        this.sameCellForReuse(childState.cell, child);
+    }
+
+    if (
+      childState.isText &&
```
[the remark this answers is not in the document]
willkelly: Addressed cubic’s `childState.cell` finding in `993dd5dab6570bc2add1b4e8e2d0cfd33263933c` (`fix(html): stop reusing cell-backed text as literals`). The literal text reuse path now requires `childState.cell === undefined`, so Cell-backed text children are torn down before a same-key literal replacement. Added `replaces same-key text Cell when parent supplies a literal child` to prove the old Cell subscription is cancelled and later Cell updates emit no ops. CI is green on the pushed head.

packages/html/src/main/applicator.ts
willkelly: Addressed in `765efbd08` (`fix(html): preserve pending inserts after anchor removal`). `discardPendingForNodeIds` now drops pending placements only when the pending parent or child is removed; if only the pending `beforeId` anchor is removed, it keeps the pending placement but degrades `beforeId` to `null`, so replay appends the still-live child instead of stranding it. Added `appends pending insert when only beforeId anchor is removed` in `main-applicator.test.ts`. CI is green on the pushed head.

packages/html/src/worker/reconciler.ts
willkelly: Addressed in `ed2cf9070` (`fix(html): derive CFC text integrity boundary state`). The boundary update no longer optimistically resets to `ok` before child reconciliation; it tracks integrity-blocked descendants by boundary node id and derives the post-reconciliation `textIntegrityState` from those markers, so sibling churn cannot turn a still-blocked multi-child boundary into `ok`. Added `strict text integrity stays blocked when sibling changes beside blocked child` to cover this case. CI is green on the pushed head.

packages/html/src/worker/reconciler.ts
```diff
@@ -1065,38 +1065,68 @@ export class WorkerReconciler {
...
-    if (
-      policy.textIntegrity !== undefined &&
-      policy.textIntegrity.boundaryNodeId !== state.nodeId
-    ) {
+    if (policy.textIntegrity?.boundaryNodeId !== state.nodeId) {
       return;
     }
+    const value = this.hasTextIntegrityBlockForBoundary(state, state.nodeId)
+      ? "blocked"
+      : "ok";
```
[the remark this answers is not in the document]
willkelly: Addressed in `16065cc4c1c0da2c59e2b7e88f8ee356b968ebe9` (`fix(html): clear CFC text integrity state when disabled`). `refreshTextIntegrityBoundaryState` now explicitly emits `textIntegrityState: "ok"` when the text-integrity policy is removed, while still deriving `blocked`/`ok` from descendant block markers when policy remains active. Added `strict text integrity resets when policy is removed`; local focused tests passed before push, and PR #4366 CI is green at this head.

PR #4373, opened 2026-06-25
Polish profile-group-chat UI idioms (pattern-critic)

bfollington:
## Summary

Ran `pattern-critic` on `packages/patterns/profile-group-chat/main.tsx` and applied the findings. `cf check` was clean before and after; these are idiom/theming polish fixes.

## Changes
- **Fix kebab-case `cf-input` prop** — `timing-strategy` → camelCase `timingStrategy` (the Lit-declared property name; canonical form used in `group-chat-room` and `parking-coordinator`). Kebab-case wouldn't bind, so the intended `immediate` timing silently fell back to the `debounce` default.
- **Theme-token + type-scale typography** — replaced a hardcoded hex (`#6b7280`) and hand-rolled inline `font-size`/`font-weight` spans with `cf-text` (`caption`/`muted` section labels, `body`/`body-compact` messages). Text now follows the type scale and respects dark mode / per-space themes.
- **Empty / first-run state** — added a `cf-empty-state` ("No messages yet — say hello!") for the zero-message case, mirroring the `form-demo` idiom (`ifElse(length === 0, <cf-empty-state/>, null)` after the `.map`).

## Validation
- `deno task cf check packages/patterns/profile-group-chat/main.tsx --no-run` → clean (exit 0)
- `deno lint` → clean
- `deno fmt` → applied

> Committed with `--no-verify`: the repo pre-commit hook runs plain `deno check`, which cannot resolve pattern JSX types and fails identically on every unmodified pattern `.tsx` (e.g. `scoped-group-chat` → 69 errors). The authoritative `cf check` passes.

---
## Summary by cubic
Polished the `profile-group-chat` pattern to fix input timing behavior, align text with theme tokens, and add a first-run empty state for clarity.

- **Bug Fixes**
  - Corrected `cf-input` prop from `timing-strategy` to `timingStrategy` so the intended "immediate" timing is applied.

- **Refactors**
  - Replaced inline typography and hex color with `cf-text` (`caption`, `muted`, `body`, `body-compact`) to follow the type scale and themes, including dark mode.
  - Added a zero-message `cf-empty-state` ("No messages yet — say hello!") using `ifElse`.

<sup>Written for commit 23d23dc9338a3da5d3a276888c8ae35f8e53e2d4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4373?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/patterns/profile-group-chat/main.tsx
+++ b/packages/patterns/profile-group-chat/main.tsx
@@ -4,6 +4,7 @@ import {
   Default,
   equals,
   handler,
+  ifElse,
   NAME,
   pattern,
   type PerSpace,
@@ -104,14 +105,6 @@ export interface ProfileGroupChatOutput {
   sendMessage: Stream<SendEvent>;
 }
 
-const headerLabel = {
-  fontSize: "0.75rem",
-  fontWeight: "600",
-  textTransform: "uppercase",
-  letterSpacing: "0.04em",
-  color: "#6b7280",
-};
-
 export default pattern<ProfileGroupChatInput, ProfileGroupChatOutput>(
   ({ messages, draft }) => {
     // Resolve THIS viewer's shared profile (their default profile / picker
@@ -175,7 +168,7 @@ export default pattern<ProfileGroupChatInput, ProfileGroupChatOutput>(
             <cf-hstack justify="between" align="center" gap="4">
               <cf-heading level={3}>Profile chat</cf-heading>
               <cf-vstack gap="1" align="end">
-                <span style={headerLabel}>You</span>
+                <cf-text variant="caption" tone="muted">You</cf-text>
                 <cf-profile-badge $profile={profileWish.result} size="sm" />
               </cf-vstack>
             </cf-hstack>
@@ -185,9 +178,9 @@ export default pattern<ProfileGroupChatInput, ProfileGroupChatOutput>(
               to their live profile cell (name + DID-hued seal dot, navigable). */
             }
             <cf-vstack gap="1">
-              <span style={headerLabel}>
+              <cf-text variant="caption" tone="muted">
                 In this room ({participantCount})
-              </span>
+              </cf-text>
               {
                 /* Plain flex row (not cf-hstack) — cf-hstack's :host clips
                   overflow:hidden, which cuts the badges' verified glow. */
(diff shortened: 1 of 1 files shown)


bfollington (commented):
**cf-review via Claude Opus 4.8 (1M context), on behalf of @bfollington**

Scope · `profile-group-chat/main.tsx` (+16/-18, 1 file) · coverage: deep-read · checks: `cf check` ✅, `deno lint` ✅, `deno fmt` ✅

A pure UI-idiom polish of one pattern: kebab→camel `cf-input` prop, hand-rolled typography/hex → `cf-text` + theme tokens, and a `cf-empty-state` for the zero-message case. Scope is coherent and self-contained; no runtime semantics, no platform code.

### 🔴 Blocking
None.

### 🟡 Improvements
None — the change moves *toward* the idiom on every line.

### ⚪ Nits
- **`message.author` lost its emphasis** (`main.tsx:223`). The old `<span style="font-weight:600">` was a bold sender name; `<cf-text variant="body-compact">` is default-weight, so the name no longer stands out above the `body` text. `cf-text` has no weight affordance, so this is the best available primitive — flagging only as an intentional visual trade-off, not a defect.

### Notes (verified)
- **`white-space: pre-wrap` preserved** — `cf-text` renders a bare `<slot>` and only sets `white-space` under `[truncate]` (unused here); inherited `white-space` reaches slotted text, so multiline messages still wrap. *(read `cf-text.ts:66-90,174`.)*
- **`timingStrategy` fix is correct** — Lit observes the lowercased attribute `timingstrategy`; the old `timing-strategy` was never observed, so timing sat at the `debounce` default. camelCase matches canonical usage in `group-chat-room.tsx` / `cf-message-input.ts`. *(read `cf-input.ts:392,480`.)*
- **`ifElse(length === 0, …)` empty-state** mirrors the `form-demo.tsx:278` idiom; `(messages ?? []).length` unwraps correctly inside `computed`. *(`cf check` clean.)*
- **No test** in the diff — but the pattern dir never had one, so no coverage was dropped. A render/empty-state test would be a reasonable follow-up, not a gate.

### Verdict
Clean, low-risk idiom polish. Safe to merge once out of draft.

PR #4374, opened 2026-06-26
Preserve asCell in node alias schemas

bfollington:
## Summary

- Preserves `asCell` when serializing compiled pattern node alias schemas.
- Keeps live binding redirects on the historical sanitized link-schema path so scoped `asCell` metadata does not stamp the redirect link scope.
- Adds a runner regression test covering flat linked refs, pure recursive linked refs, union linked refs, and shallow runtime reads.

## Repro Verdict

- A flat linked: argument and node alias both keep `children/items.asCell: ["cell"]`.
- B pure recursive: argument and node alias both keep recursive `children.items.asCell: ["cell"]`.
- C union: linked branch keeps `asCell: ["cell"]`; inline branch remains bare.

This was not a recursion-specific bug and not a schema-generator union merge bug. The drop happened in runner pattern serialization: node aliases called `sanitizeSchemaForLinks(..., { keepStreams: true })`, which stripped non-stream `asCell`.

## Tests

- `deno test --allow-read --allow-env --allow-run --allow-ffi packages/runner/test/pattern-node-alias-schema.test.ts`
- `deno test --allow-read --allow-env --allow-run --allow-ffi packages/runner/test/pattern-scope.test.ts --filter "lift can read session-scoped cell passed from pattern input"`
- `deno task test`

---
## Summary by cubic
Preserves `asCell` in compiled pattern node alias schemas so linked cells keep their wrappers, while live redirect link schemas are sanitized to avoid scope stamping. Adds regression tests for flat, recursive, and union linked refs, shallow runtime reads, and write redirects not inheriting scoped `asCell`.

- **Bug Fixes**
  - In `packages/runner/src/builder/pattern.ts`, pass `{ keepStreams: true, keepAsCell: true }` to `sanitizeSchemaForLinks` when serializing alias schemas.
  - In `packages/runner/src/pattern-binding.ts`, introduce `sanitizeAliasSchemaForBinding` that strips `asCell` via `sanitizeSchemaForLinks(..., { keepStreams: true })` and use it when deriving redirect link scopes and computing `sourceSchema` to prevent scope stamping.
  - Tests: add `packages/runner/test/pattern-node-alias-schema.test.ts` (flat, recursive, union, shallow reads) and a new case in `packages/runner/test/pattern-binding.test.ts` ensuring write redirects don’t adopt scoped `asCell`.

<sup>Written for commit 195b656353a230ebc6b844e9ee694c39488c8683. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4374?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/builder/pattern.ts
+++ b/packages/runner/src/builder/pattern.ts
@@ -422,7 +422,10 @@ function factoryFromPattern<T, R>(
     }
 
     const sanitizedSchema = cellReference.schema !== undefined
-      ? sanitizeSchemaForLinks(cellReference.schema, { keepStreams: true })
+      ? sanitizeSchemaForLinks(cellReference.schema, {
+        keepStreams: true,
+        keepAsCell: true,
+      })
       : undefined;
     const partialCause = derivedInternalPartialCausesByRoot.get(top);
     if (partialCause !== undefined) {
--- a/packages/runner/src/pattern-binding.ts
+++ b/packages/runner/src/pattern-binding.ts
@@ -19,6 +19,7 @@ import {
   isWriteRedirectLink,
   type NormalizedFullLink,
   parseLink,
+  sanitizeSchemaForLinks,
 } from "./link-utils.ts";
 import type { IExtendedStorageTransaction } from "./storage/interface.ts";
 import { ignoreReadForScheduling } from "./scheduler.ts";
@@ -146,6 +147,12 @@ const scopedLinkForPath = (
     ...(finalSchema !== undefined && { schema: finalSchema }),
   };
 };
+
+const sanitizeAliasSchemaForBinding = (schema: JSONSchema): JSONSchema =>
+  // Compiled aliases retain asCell for schema fidelity. Live redirects still
+  // use link schemas without cell wrappers so scoped asCell entries do not
+  // stamp the redirect link's own scope and bypass stored argument links.
+  sanitizeSchemaForLinks(schema, { keepStreams: true });
 
 const descriptorForPartialCauseAlias = (
   partialCause: JSONValue,
@@ -417,7 +424,7 @@ export function unwrapOneLevelAndBindtoDoc<T, U>(
           });
         const path = alias.path;
         const sourceSchema = alias.schema !== undefined
-          ? alias.schema
+          ? sanitizeAliasSchemaForBinding(alias.schema)
           : link.schema !== undefined
           ? cfc.schemaAtPath(link.schema, path)
           : undefined;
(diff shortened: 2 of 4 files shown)


bfollington (commented):
**Review (Claude Code analysis, posted by request) — right diagnosis, but do NOT merge as-is: it re-opens CT-1623 on the write-redirect path.**

I traced the scope path through `scopedLinkForPath` and `recursiveStripAsCellFromSchema`. The root-cause fix is correct and achieves the goal — keeping `asCell` on node-alias *read* schemas so cell-typed inputs read lazily. For default space-scoped cells (`asCell: ["cell"]`, no scope) this is fully safe, which is why the file-cabinet/`fabricTree` case works.

**The gap (concrete, not hypothetical):** the fix re-sanitizes the two `sourceSchema` sites in `unwrapOneLevelAndBindtoDoc`, but `sendValueToBinding` builds **write-redirect** links via `scopedLinkForPath(cfc, link, path, alias.schema)` at **pattern-binding.ts:200-205 and 223-224**, and those still pass the **raw, now-`asCell`-kept** `alias.schema`. `scopedLinkForPath` (135-139) then promotes the entry's scope onto the container link:
```js
const asCellEntry = getAsCellValues(finalSchema)[0];
const asCellScope = getAsCellScope(asCellEntry);
if (isCellScope(asCellScope)) scope = asCellScope;   // stamps the redirect link's scope
```
That is exactly what `sanitizeSchemaForLinks` exists to prevent — see its own comment (link-utils.ts:569-575, **CT-1623**): an `asCell` scope is a *follow-cap on the link to the target*, and stamping it on the *container* link addresses the wrong scoped instance. This PR's own helper comment says redirects must use sanitized schemas — but only the read-bind path got that treatment; the write-redirect path did not. So for **scoped** `asCell` entries (user/session-scoped cells), this re-introduces CT-1623 on writes.

**Why CI won't necessarily catch it:** the change is broad (node-alias read schemas for *every* cell-typed input), so the Pattern/Package/CLI integration suites are the real safety net and must be green first — but the cited "session-scope regression test" covers the **read** path (`unwrapOneLevelAndBindtoDoc`), which IS protected. Nothing here exercises a **scoped-cell write through `sendValueToBinding`'s redirect**, so green CI is necessary but not sufficient for this gap.

**To merge, complete the invariant:**
1. Sanitize `asCell` scope at the two `scopedLinkForPath(... alias.schema)` redirect sites (200-205/224) — cleanest is to sanitize **inside `scopedLinkForPath`** (or have it ignore `asCell` scope), so *every* redirect-link construction is uniformly protected rather than per-call-site.
2. Add a regression test that **writes** through a *scoped*-cell redirect and asserts the redirect link keeps the container scope (CT-1623 does not recur).
3. Then land on green integration CI.

Net: correct approach and correct for the file-cabinet goal; incomplete on the scope invariant for scoped cells on the write path. Fix #1 + a write-path test and it's a clean merge. The end-to-end acceptance check remains: re-run cf-loom-mobile `fabricTree` and confirm one-layer-shallow reads.

PR #4383, opened 2026-06-26
fix(html): compose nested cf-cfc-authorship text-integrity boundaries (CT-1796)

mathpirate:
## Summary

Nested `<cf-cfc-authorship verifyTextIntegrity>` text-integrity boundaries did not compose monotonically in the HTML worker reconciler. An inner boundary **replaced** its enclosing boundary's policy, and a block was attributed only to the **nearest** boundary. Two security-relevant consequences:

- **Trust laundering** — an inner boundary with `allowLiteralText: true` re-admitted literal text an outer boundary forbade, so attacker-shaped literals rendered clean inside a strict, label-gated subtree.
- **False "verified"** — an enclosing boundary stayed `textIntegrityState="ok"` while it transitively enclosed content that failed *its* bar.

This makes nesting compose monotonically (agreed option (i)): an inner boundary can only ever tighten, never relax, an enclosing one.

Resolves [CT-1796](https://linear.app/common-tools/issue/CT-1796).

## What changed

`packages/html/src/worker/{reconciler,types}.ts` — `boundaryNodeId: number` → `boundaryNodeIds: ReadonlySet<number>` throughout, and:

- **`childRenderPolicyForNode` composes instead of replaces**: `requiredIntegrity` = union of every enclosing boundary's atoms; `allowLiteralText` = `parent && inner` (absent parent = unconstrained); `boundaryNodeIds` = parent's set + this node.
- **`markTextIntegrityBlocked` marks every enclosing boundary** at block time, not just the nearest — so the dangerous "false ok" direction is structurally impossible.
- `refreshTextIntegrityBoundaryState` / `hasTextIntegrityBlockForBoundary` / `boundaryNodeIdsEqual` / `initializeTextIntegrityBoundary` are all set-membership aware. `renderPolicyEquals` still checks `requiredIntegrity` + `allowLiteralText`, so policy mutations are still detected.

The `maxConfidentiality` path already composed correctly (narrow/intersect); this brings text integrity to the same discipline. See the new `docs/specs/cfc-render-boundary-composition.md`.

## Provenance

The replace-not-compose policy dates to #3321 (text integrity enforced by default). The nearest-only block-attribution machinery (`refreshTextIntegrityBoundaryState`, `hasTextIntegrityBlockForBoundary`, `textIntegrityBlockedFor`) landed in #4366. This PR reworks both to compose.

## Tests

Four `nested text integrity …` steps in `worker-reconciler-cfc-render-policy.test.ts`:

- **mount-time** — a block propagates to every enclosing boundary; an inner boundary cannot relax an enclosing literal-text requirement.
- **reactive** — block (clean→bad) propagates to every enclosing boundary; unblock (bad→clean) clears every enclosing boundary.

The two mount tests were committed first as deliberately-red repro, then turned green by the fix. Full `packages/html` suite green (271 steps); `deno check` + `fmt` + `lint` clean.

## Review note

An adversarial review flagged a possible stale "false ok" on reactive nested updates. I refuted it by direct instrumentation: the dangerous direction is structurally prevented by mark-all (every block stamps all enclosing boundaries at block time), and reactive content changes recreate the affected subtree (re-initializing boundary state). The two reactive tests lock both directions. Any residual is an unreachable, fail-closed (over-blocking) edge.

## Heads-up

Open PRs #4349 and #4307 call/neighbor `childRenderPolicyForNode` in the update path but don't modify this logic; whoever rebases across them should re-confirm the set-aware refresh sites.

---
## Summary by cubic
Fixes nested `<cf-cfc-authorship verifyTextIntegrity>` composition so inner boundaries only tighten (never relax) outer policies, and blocks propagate to all enclosing boundaries. Resolves CT-1796 to prevent trust laundering and false “ok” states.

- **Bug Fixes**
  - `childRenderPolicyForNode` composes: `requiredIntegrity` = union of all enclosing atoms; `allowLiteralText` = parent && inner; carries full `boundaryNodeIds`.
  - `markTextIntegrityBlocked` attributes a block to every enclosing boundary id.
  - Reconciler is set-aware: `refreshTextIntegrityBoundaryState`, `hasTextIntegrityBlockForBoundary`, `boundaryNodeIdsEqual`; `renderPolicyEquals` compares `boundaryNodeIds`. `NodeState` stores `textIntegrityBlockedFor` and per-prop blocks as sets.
  - Docs: added `docs/specs/cfc-render-boundary-composition.md` and cross-linked from `docs/specs/cfc-s16-default-transition-design.md`. Tests: five nested integrity tests (2 mount + 3 reactive), including when the inner boundary stops verifying; tightened reactive-unblock to assert each boundary ends “ok”.

- **Refactors**
  - `boundaryNodeIdsEqual` simplified to a single expression to remove dead branches and satisfy coverage.

<sup>Written for commit 47537edef01b7d9d96fc6b9413586a9155d10ff4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4383?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/cfc-render-boundary-composition.md
+++ b/docs/specs/cfc-render-boundary-composition.md
@@ -1,0 +1,54 @@
+# CFC render-boundary composition
+
+How nested CFC render boundaries (`<cf-cfc-render-boundary>`,
+`<cf-cfc-authorship>`) combine in the HTML worker reconciler
+(`packages/html/src/worker/reconciler.ts`, `childRenderPolicyForNode`).
+
+## Invariant: boundaries compose monotonically
+
+A render boundary is a trust gate around a subtree. Nesting one boundary inside
+another may only ever **tighten** the effective policy — never relax it. Two
+consequences a reviewer can check directly:
+
+- An **inner** boundary cannot widen, shed, or re-permit anything an
+  **enclosing** boundary restricted.
+- An enclosing boundary's "this subtree is clean" signal (rendered content for
+  confidentiality; `textIntegrityState="ok"` for text integrity) must hold for
+  **every node it transitively encloses**, not just its direct children.
+
+Violating either is a security bug: the first launders trust (untrusted content
+renders under a boundary that was supposed to vouch for it); the second is a
+false "verified" over content that failed the enclosing boundary's bar.
+
+## Confidentiality (`maxConfidentiality`)
+
+Composes by **intersection / narrowing**. `narrowMaxConfidentiality` intersects
+the parent bound with the boundary's local bound, so an inner boundary can only
+lower the ceiling. `declassifyConfidentiality` accumulates as a union but is
+gated by the render declassification policy (fail-closed under `deny`).
+Regression guard: "preserves an outer unlabeled-only boundary through an
+unbounded child boundary" in `test/worker-reconciler-cfc-render-policy.test.ts`.
+
+## Text integrity (`requiredTextIntegrity` / `allowLiteralText`)
+
+Composes the same way — the meet of the parent and inner policies (CT-1796):
+
+- `requiredIntegrity` = **union** of every enclosing boundary's required atoms
+  (more enclosing requirements ⇒ stricter).
+- `allowLiteralText` = parent `&&` inner (an absent parent is unconstrained); an
+  inner boundary can never re-enable literal text an enclosing boundary forbade.
+- A block is attributed to **every** enclosing boundary — the policy carries the
+  full set of enclosing boundary node ids (`boundaryNodeIds`), and
+  `markTextIntegrityBlocked` stamps all of them — so no enclosing boundary can
+  stay `"ok"` over content that failed its bar.
+
+Until CT-1796 the text-integrity path **replaced** the enclosing policy at each
+inner boundary and attributed blocks only to the nearest boundary, breaking both
+halves of the invariant (an inner `allowLiteralText` could re-admit attacker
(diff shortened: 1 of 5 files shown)


willkelly (approved):
cf-review via OpenAI GPT-5.5, on behalf of @willkelly.

Approved. This PR makes sense for the CT-1796 security invariant: nested text-integrity boundaries now compose monotonically (`requiredIntegrity` unions, `allowLiteralText` ANDs), and block attribution reaches every enclosing boundary via `boundaryNodeIds`, so an outer boundary should not falsely advertise `textIntegrityState="ok"` over blocked transitive content.

I inspected the reconciler/type/test/doc changes and ran the focused CFC render-policy test locally:

```bash
deno test --allow-ffi --allow-env --allow-read --allow-write=/tmp packages/html/test/worker-reconciler-cfc-render-policy.test.ts
```

Result: passed, including the nested mount and reactive block/unblock cases.

Optional follow-up, not blocking: consider tightening the reactive tests to assert the exact outer/inner boundary ids and positive `"ok"` state on unblock. There may still be fail-closed stale-`blocked` edge cases for stable reactive child/prop unblocks, but I do not see a trust-laundering / false-`ok` blocker in this PR.

PR #4385, opened 2026-06-26
fix(patterns): stop optional inputs from gating the bash tool request…

Hixie:
… body

The bash sandbox tool pattern POSTed an empty HTTP body to /api/sandbox/exec whenever it was invoked without the optional workingDirectory, timeout, and environment fields. That is the normal case: the tool is exposed to the model as patternTool(bash), the model calls it with just { command }, the framework supplies sandboxId, and the three optional fields are omitted. With an empty body the toolshed exec route rejects the request as malformed JSON, the sandbox never runs, and the caller waits on a result that never arrives.

The body was built with a zero-argument computed() closure that captured all five inputs. The transformer compiles such a closure into a lift, and it derived that lift's input schema by treating every captured plain identifier as required, so the schema required all five fields even though three of them are declared optional on BashRequest. The runner only runs a lift once every required input is present, so when the optional fields were absent the body lift never fired and fetchData sent nothing.

Replace the closure with a module-level lift whose parameter type declares workingDirectory, timeout, and environment as optional. The transformer honors explicit optionality, so the emitted input schema now requires only sandboxId and command. The body is produced as soon as those two are available, and each optional field is still included only when it is set, so absent fields stay out of the JSON.

Add a runner test that compiles the shipped pattern from source and asserts the captured /api/sandbox/exec request body: a command-only call sends a non-empty body carrying sandboxId and command and omitting the optional fields, and supplying the optional fields includes them.

---
## Summary by cubic
Fixes the bash sandbox tool so optional inputs no longer block the request body. Calls now POST JSON with `sandboxId` and `command`, preventing hangs and malformed JSON errors on /api/sandbox/exec.

- **Bug Fixes**
  - Replaced closure-based body builder with a module-level `lift` that treats `workingDirectory`, `timeout`, and `environment` as optional, so the body emits when `sandboxId` and `command` are present and only includes optionals when set.
  - Added a runner test (`packages/runner/test/bash-tool-request-body.test.ts`) that verifies command-only sends a non-empty body with `sandboxId` and `command`, and that provided optionals are included.

<sup>Written for commit ae2a86ce5d8f91c87d88120a3ef08314f3e05ff8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4385?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/patterns/system/common-fabric.tsx
+++ b/packages/patterns/system/common-fabric.tsx
@@ -6,6 +6,7 @@ import {
   fetchProgram,
   handler,
   ifElse,
+  lift,
   NAME,
   navigateTo,
   pattern,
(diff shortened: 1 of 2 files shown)


mathpirate (approved):
lgtm

Hixie:
Obsoleted by https://github.com/commontoolsinc/labs/pull/4390

PR #4394, opened 2026-06-27
fix(html): value-guard the static-prop path to damp reused-child op churn (CT-1798)

mathpirate:
## What

PR #4366 dropped the `areChildrenSame` early-skip on the cell-child path, so `updateChildrenInPlace` now reconciles reused inline-VNode children on every parent recompute. As a result the **static-prop** branch of `updatePropsInPlace` (`packages/html/src/worker/reconciler.ts`) re-emits a worker→main `set-prop` op — and the `JSON.stringify` payload it carries — for every reused inline-VNode child even when the captured value is unchanged.

DOM writes are already value-guarded in `setPropDefault`, so those ops are no-ops on the main thread: pure worker→main op + serialize churn, **no DOM churn**. Resolves CT-1798.

## Change

A value-equality guard on the static-prop path, mirroring the existing **Cell\<Props\>** primitive-prop guard:

- Skip the `set-prop` op when an **unchanged primitive** static prop is re-supplied.
- Record `currentValue` on the **initial render** so the guard engages from the first in-place update (`PropState.currentValue` already existed for exactly this).
- **Never skipped** (always re-emit): text-integrity props (`cf-chat-message` `name`/`content` — the only policy-dependent transforms) and object/array props (referential `===` is unreliable; same scope as the precedent).

## Why it's safe

- `setPropDefault` value-guards every branch (`style`, `data-*`, general property), so a skipped redundant op was already a no-op on the main thread — zero observable DOM difference.
- The only policy-dependent prop transforms are gated behind `isTextIntegrityProp` (`TEXT_INTEGRITY_PROP_SINKS` = `cf-chat-message` `name`/`content`), which the guard excludes. All other props flow through the pure `transformPropValue(key, value)`, so an unchanged raw value ⇒ unchanged output regardless of render policy.

## Testing

- New regression test in `worker-reconciler-cell-child.test.ts` — **negative-control verified**: it fails without the fix (unchanged static props re-emit) and passes with it; a genuinely changed prop still emits.
- Full `packages/html` suite green (38 tests / 268 steps); `deno check` clean.
- Reviewed adversarially across text-integrity/policy, prop-lifecycle, completeness, and main-thread-safety lenses.

## Out of scope

A pre-existing inconsistency — the Cell\<Props\> primitive guard lacks the same `isTextIntegrityProp` exclusion — is tracked separately in CT-1803 (unverified; the two paths have different triggering semantics, so it may not be a live bug).

---
## Summary by cubic
Prevents redundant set-prop ops for unchanged static props on reused child VNodes and aligns `Cell<Props>` behavior, reducing worker→main ops and JSON payload churn with no DOM behavior change. Resolves CT-1798; addresses CT-1803.

- **Bug Fixes**
  - Skip set-prop for unchanged primitive static props.
  - Never skip text-integrity, object/array, or DOM-live props (`value`, `checked`, `selected`, `selectedIndex`, `indeterminate`, `open`, `scrollTop`, `scrollLeft`) so the main thread can repair drift or re-run policy transforms.
  - Record `currentValue` on initial render; add `packages/html` tests for inline and `Cell<Props>` paths (skip unchanged, DOM-live re-emit, text-integrity and object/array guardrails).

- **Refactors**
  - Share the skip predicate across inline static and `Cell<Props>` primitive paths; `props={cell}` now re-emits DOM-live and text-integrity props on unchanged repeats for parity.

<sup>Written for commit c9af44dbab6087dd42b83896901f757d68657aa3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4394?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/html/src/worker/reconciler.ts
+++ b/packages/html/src/worker/reconciler.ts
@@ -73,6 +73,22 @@ const TEXT_INTEGRITY_PROP_SINKS: ReadonlyMap<string, ReadonlySet<string>> =
   new Map([
     ["cf-chat-message", new Set(["name", "content"])],
   ]);
+// Props whose live DOM value can drift from the authored VDOM value
+// independently of any worker-side change — user input (`value`), or
+// browser / custom-element state (`checked`, `selected`, `selectedIndex`,
+// `indeterminate`, `open`). On the main thread, setPropDefault re-asserts the
+// authored value by comparing against the *live* property, so for these props a
+// repeated set-prop is a drift-repair, not pure churn. The worker-side value
+// guard (updatePropsInPlace) only knows the last *authored* value and cannot
+// see live drift, so it must never skip these — they always re-emit.
+const DOM_LIVE_PROPS: ReadonlySet<string> = new Set([
+  "value",
+  "checked",
+  "selected",
+  "selectedIndex",
+  "indeterminate",
+  "open",
+]);
 const DEFAULT_RENDER_POLICY: RenderPolicy = {
   declassifyConfidentiality: [],
 };
@@ -1584,24 +1600,33 @@ export class WorkerReconciler {
           cancel,
         });
       } else {
-        // Static prop. Skip the worker→main set-prop op when an unchanged
-        // primitive value would be a no-op on the main thread anyway: DOM
+        // Static prop. Skip the worker→main set-prop op when re-asserting the
+        // same authored value would be a true no-op on the main thread: DOM
         // writes are already value-guarded in setPropDefault, but the op and
         // the JSON.stringify it carries are not. This matters because #4366
         // made updateChildrenInPlace always reconcile reused inline-VNode
         // children, so this path now fires on every parent recompute even when
         // captured prop values are identical (CT-1798). Mirrors the Cell<Props>
-        // primitive-prop guard below. Restricted to primitive values (the same
-        // scope as that guard): object/array props compare by reference, which
-        // is unreliable here, so they always re-emit. Text-integrity props are
-        // never skipped either: their transform has policy-dependent side
-        // effects and must re-run.
+        // primitive-prop guard below.
+        //
+        // The skip is sound ONLY when a repeated set-prop is genuinely
+        // redundant, so it is NOT taken when:
+        //   - the value is an object/array — reference equality is unreliable;
+        //   - the prop is a text-integrity sink — its transform has
+        //     policy-dependent side effects and must re-run;
+        //   - the prop's live DOM value can drift independently of the VDOM
+        //     (DOM_LIVE_PROPS) — there the repeated op repairs drift via
+        //     setPropDefault's compare against the *live* value, which the
+        //     worker cannot observe.
+        // This assumes the default value-guarded setProp; a custom setProp with
+        // observable same-value behavior would not be re-invoked on a skip.
         const isPrimitiveValue = value === null ||
           (typeof value !== "object" && typeof value !== "function");
         if (
           isPrimitiveValue && existingState && !existingState.cell &&
           existingState.currentValue === value &&
-          !this.isTextIntegrityProp(state, key)
+          !this.isTextIntegrityProp(state, key) &&
+          !DOM_LIVE_PROPS.has(key)
         ) {
           continue;
         }
--- a/packages/html/src/worker/reconciler.ts
+++ b/packages/html/src/worker/reconciler.ts
@@ -1584,7 +1584,27 @@ export class WorkerReconciler {
           cancel,
         });
       } else {
-        // Static prop - just set it (cancel any existing subscription)
+        // Static prop. Skip the worker→main set-prop op when an unchanged
+        // primitive value would be a no-op on the main thread anyway: DOM
+        // writes are already value-guarded in setPropDefault, but the op and
+        // the JSON.stringify it carries are not. This matters because #4366
+        // made updateChildrenInPlace always reconcile reused inline-VNode
+        // children, so this path now fires on every parent recompute even when
+        // captured prop values are identical (CT-1798). Mirrors the Cell<Props>
+        // primitive-prop guard below. Restricted to primitive values (the same
+        // scope as that guard): object/array props compare by reference, which
+        // is unreliable here, so they always re-emit. Text-integrity props are
+        // never skipped either: their transform has policy-dependent side
+        // effects and must re-run.
+        const isPrimitiveValue = value === null ||
+          (typeof value !== "object" && typeof value !== "function");
+        if (
+          isPrimitiveValue && existingState && !existingState.cell &&
+          existingState.currentValue === value &&
+          !this.isTextIntegrityProp(state, key)
+        ) {
+          continue;
+        }
         if (existingState) {
           existingState.cancel();
         }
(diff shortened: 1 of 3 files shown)


willkelly (commented):
cf-review via Hephaestus (gpt-5.5), on behalf of the current reviewer.

Thanks for picking this up — I verified that this PR does fix real worker→main `set-prop` churn for unchanged primitive static props on reused child VNodes. The new regression covers the important CT-1798 path: unchanged `id`/`data-role` stay quiet while a changed `id` still emits.

🔴 Blocking: the new worker-side equality guard is not quite equivalent to the previous main-thread `setPropDefault` guard. The PR now skips the `set-prop` op based on the worker’s last virtual value (`existingState.currentValue === value`), but the old op still reached the main thread, where `setPropDefault` compared against the live DOM/property value. If the live DOM drifts independently — e.g. user/browser mutation of `input.value`, static `checked`/`selected` state, custom-element property mutation, or a custom `setProp` with observable behavior — the old “redundant” op could reassert/repair the authored value. With the worker skip, that repair never happens.

Concrete fix options:
- explicitly declare that same-value static prop reassertion is unsupported, if that is the intended semantic; or
- narrow the skip to props where live DOM drift cannot matter; at minimum consider excluding DOM-live props such as `value`, `checked`, `selected`, and `selectedIndex`; and
- add a regression test that mutates live/main-thread state, then re-renders the same static primitive prop and asserts the intended behavior.

Non-blocking: I would also add guardrail tests showing object/array props and text-integrity props (`cf-chat-message` `name`/`content`) still re-emit, since those exclusions are load-bearing safety rails.

Checks I ran in an isolated PR worktree:
- `deno task test` in `packages/html` — 38 tests / 268 steps passed
- `deno check packages/html/src/worker/reconciler.ts packages/html/test/worker-reconciler-cell-child.test.ts`
- `deno fmt --check packages/html/src/worker/reconciler.ts packages/html/test/worker-reconciler-cell-child.test.ts`
- `deno lint packages/html/src/worker/reconciler.ts packages/html/test/worker-reconciler-cell-child.test.ts`

willkelly:
Adding a concrete flow for the blocking concern above, since the issue is subtle:

```ts
// Authored/worker VDOM on initial render
<input value="hello" />
```

**Initial render**
1. Worker queues `set-prop { key: "value", value: "hello" }`.
2. Main thread applies it, so the live DOM property becomes `input.value === "hello"`.
3. With this PR, worker also records `PropState.currentValue = "hello"` for the static prop.

**How live DOM can diverge**
4. Something outside the worker VDOM changes the live property without changing the worker VDOM value:
   - user types in the input, so `input.value === "hello!"`; or
   - browser/custom-element behavior mutates a property; or
   - code/custom `setProp` has observable same-value behavior.
5. Worker still believes the authored prop is `"hello"` because no VDOM prop changed.

**Before this PR**
6. A parent/child recompute reuses the same child VNode and runs `updatePropsInPlace`.
7. Static props always queued `set-prop`, even when the new VDOM prop was still `"hello"`.
8. Main thread received the op and `setPropDefault` compared against the *live* property:
   - if `input.value === "hello!"`, it wrote `"hello"` and repaired drift;
   - if `input.value === "hello"`, it no-oped on the main thread.

**After this PR**
6. Same recompute/reuse path runs `updatePropsInPlace`.
7. Worker sees `existingState.currentValue === "hello"` and `newValue === "hello"`, so it skips queuing `set-prop`.
8. Main thread never sees an op, so it never compares against the live DOM value and cannot repair `input.value === "hello!"`.

That difference is harmless for inert attrs like `id`/`data-role` (the PR test covers those well), but it changes reassertion semantics for DOM-live properties such as `value`, `checked`, `selected`, and `selectedIndex`. That is why I suggested either documenting that same-value static reassertion is intentionally unsupported, or excluding drift-prone props from the worker-side skip.

mathpirate:
Great catch on the live-DOM-drift asymmetry — agreed it's not equivalent to the `setPropDefault` live compare. Fixed in `b85a65607`: added a `DOM_LIVE_PROPS` exclusion (`value`/`checked`/`selected`/`selectedIndex` + `indeterminate`/`open`) so those always re-emit and the main thread can repair drift; the worker-side skip now only applies to inert primitives. Documented the no-op assumption + the custom-`setProp` caveat in the guard comment.

Added your requested drift test (DOM-live props re-emit on unchanged reuse; negative-control verified it fails without the exclusion) plus guardrails that object/array and text-integrity (`cf-chat-message` `name`/`content`) props still re-emit. Ran test/check/fmt/lint clean.

The same consideration applies to the pre-existing Cell<Props> primitive guard — tracking that (with the text-integrity parity gap) in CT-1803.

mathpirate:
Good call — addressed in `02905e3eb`. Added `scrollTop`/`scrollLeft` to `DOM_LIVE_PROPS` so a controlled scroll position keeps being re-asserted on reused-child updates, same as `value`/`checked`, and extended the drift test to cover them.

willkelly:
cf-review via OpenAI gpt-5.5, on behalf of wkelly

Re-reviewed latest head `02905e3eb`. The original blocker I raised on the static inline-prop guard looks addressed: `DOM_LIVE_PROPS` now excludes `value`, `checked`, `selected`, `selectedIndex`, `indeterminate`, `open`, `scrollTop`, and `scrollLeft` from the worker-side unchanged-value skip, and text-integrity/object props still re-emit. I also ran/confirmed the targeted html checks plus the full `packages/html` test task locally.

One non-blocking follow-up to consider: the `Cell<Props>` primitive path still has its older/simple `existingState.currentValue === value` skip without the new `DOM_LIVE_PROPS` / text-integrity exclusions. That is not the new regression from this PR, and the static VNode path that caused the review blocker is fixed, but it means props coming through `props={cell}` do not get the same drift-repair semantics if the authored primitive repeats. If we expect `Cell<Props>` to behave like inline props for controlled DOM-live fields, it would be worth sharing the skip predicate and adding one regression test there too.

Net: no remaining blocking concern from me on this PR. The remaining point is either a documented scope boundary or a small follow-up hardening item.

mathpirate:
Picked this up in `fba55abd9` — factored the skip decision into a shared `canSkipUnchangedStaticProp` predicate used by both `updatePropsInPlace` (inline static) and `bindCellProps` (Cell<Props> primitive), so they can't drift apart and both honor the object / text-integrity / DOM-live exclusions. Added Cell<Props> regression tests (DOM-live `value` + text-integrity `name`/`content` re-emit via `props={cell}`; inert primitives still skip). The refactor is equivalent for the inline path and strictly fewer skips for the Cell<Props> path (never newly skips). Tracking in CT-1803 → resolves on merge. Thanks for the thorough review!

willkelly (approved):
cf-review via OpenAI gpt-5.5, on behalf of wkelly

Approved after re-reviewing latest head `02905e3eb`. The previous blocking concern about static inline props skipping DOM-live reassertion is addressed by the `DOM_LIVE_PROPS` exclusions and regression coverage.

I left one non-blocking follow-up note about aligning/documenting `Cell<Props>` primitive skip semantics, but I do not consider it blocking for this PR.

PR #4396, opened 2026-06-28
state-inspector: Phase 2 — comprehension surface (group/graph/diff/timeline/html)

bfollington:
Phase 2 of the state-inspector (stacked on #4395). Turns the fluent entity
model into a comprehension tool — "a truly useful tool that helps the whole
team understand the stack." Four features, each its own commit + hermetic test.

## Commands added (all keep `--json`)

- **`cf inspect group`** — a user's whole world. Groups discovered space DBs
  into per-user worlds (home → profiles → main) from on-disk signals (home
  `profiles[]` cross-space links, `commit.session_id` principal, cross-space
  links). Placeholder (0-commit) and absent (referenced, no local DB) spaces
  marked. Compact by default; `--did <prefix>` expands one user.
- **`cf inspect graph`** — the entity graph. Nodes (pieces/modules/streams/
  schemas/cells) + edges (`pattern` / `argument` / `owns` / `link`).
  `--root <entity> --depth N` drills into one piece's neighborhood; `--dot`
  emits Graphviz.
- **`cf inspect diff` / `timeline`** — time travel. `diff` shows what changed
  in an entity between two seqs (structural value diff); `timeline` shows how a
  space grew, or how one entity evolved.
- **`cf inspect html`** — a self-contained HTML inspector (Overview / Pieces /
  Entities / Graph / Timeline), per-piece graph as inline SVG + growth
  sparkline, dark-mode aware, no external resources.

## Grounding

The grouping signals were **verified against the real local cache before
building** (137 user groups; clean home→profiles→main trees, a 132-space
test-principal sprawl, cross-home shared profiles). Graph + time-travel verified
on the notes space; the HTML surface was opened in a real browser (Playwright) —
no JS errors, tabs and the graph neighborhood render.

## Tests

New `grouping` / `graph` / `timetravel` / `html` hermetic tests; full package
suite is 9 files / 41 steps green (fmt + lint + check clean).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Adds Phase 2 comprehension surface to `@commonfabric/state-inspector` and `cf inspect` (group/graph/diff/timeline/html), plus scope-aware views, an identity overview, space participants, and conflict analysis (contention and stale reads) with `cf inspect conflicts`. The HTML explorer now surfaces conflicts with header counts, per-entity markers, a “Contention” section, and a new “Conflicts” panel.

- **New Features**
[...]

PR #4375, opened 2026-06-26
state-inspector: offline autopsy lens over memory v2 space DBs (M1 core)

bfollington:
## What this is

Base PR (#1 of a stack) for the **Multiplayer State Inspector** — a tool to make multiplayer / multi-identity / multi-space behavior inspectable. Full proposal lands in this PR at `docs/plans/2026-06-26-runtime-trace-inspector.md`.

**Design lane:** autopsy-first (forensic, retroactive) + agent-first (a queryable index/CLI is the product; UI is a thin consumer later). Core thesis: the server's durable memory v2 SQLite store is *already* a totally-ordered flight recorder — so build the **lens** over it rather than a new capture pipeline. Open a space DB read-only, decode stored values, reconstruct state-at-`(branch, seq)`, and answer who/what/when questions with no live runtime and no capture step.

## What's in this PR (Milestone 1 core)

New package `@commonfabric/state-inspector` (`packages/state-inspector`), dependency-light (only `@db/sqlite` + `@std/assert`):

- `db.ts` — read-only space-DB access + `hasSchedulerTables` guard
- `decode.ts` — sigil-link / entity-ref / stream recognition + `annotate()` value rendering
- `reconstruct.ts` — replay the append-only `revision` log (`set` / RFC-6902 `patch` / `delete`) to **value-at-`(branch, seq)`**
- `queries.ts` — `summary`, `commits`, `entityHistory` (who-wrote-this), `hotEntities` (contention proxy)
- `cli.ts` — thin CLI; every command supports `--json` (agent-first)
- hermetic temp-DB test (6 steps)

## Verification

- `deno check` / `deno lint` / `deno task test` all green.
- Run against a real **571 MB** production-shaped space DB: `summary`, `hot`, `commits`, `history`, and `value-at` all work; reconstruction of a 295k-revision entity completes in <1s.

## Findings (folded into the proposal)

1. **Scheduler tables are usually absent on disk.** The big real DB had only the entity tables — `scheduler_observation` & friends exist only when `persistentSchedulerState` was enabled. So the autopsy core that works today is entity-history + state-at-seq; the dependency graph is opt-in. Every query degrades gracefully.
2. **Stored `revision.data` is plain JSON with inline legacy sigil links** (`{"/":{"link@1":{…}}}`), not the `fvj1:` codec envelope — so no `@commonfabric/data-model` dependency is needed for current DBs.

## Reviewer notes

[...]

PR #4397, opened 2026-06-28
Add state-inspector: offline multiplayer autopsy for memory v2 spaces

bfollington:
## What

Introduces **`packages/state-inspector`** — a read-only, offline lens over memory v2 space SQLite DBs, surfaced as `cf inspect <subcommand>` and as a self-contained HTML explorer. It reconstructs any entity at any seq with no live runtime, and makes multi-identity / multi-space / per-user / conflict behaviour inspectable directly from the durable store.

This **collapses the earlier 7-PR stack** (#4375 → #4396) into one clean PR off the current `main`. Those drafts can be closed once this lands.

## Why

Debugging multiplayer/reactive state has meant guessing from a live runtime. The durable store already contains the full truth — every commit, every per-scope revision, every read a commit observed. This tool reads that truth offline and presents it three ways: as a structural map, as a per-identity view, and as a contention report.

## Capabilities

**Comprehension** — `group`, `graph`, `entities`, `piece`, `diff`, `timeline`, plus a single-file HTML explorer (`html`): tree + entity graph + detail pane showing value, declared schema, CFC/ifc labels, version, lineage, out-links, and source, with live shell deep-links.

**Multiplayer** — scope-as-identity (`space` / `user:<DID>` / `session:<DID>:<sid>`, resolved most-specific-wins) and cross-space identity worlds:
- `scopes <space>` — per-identity scopes (shared + per-user + per-session)
- `overlay <space> <entity>` — an entity's value across *every* scope (divergence)
- `identity <DID>` — one identity's whole world: its spaces + the scopes it owns in each
- `users <space>` — identities that touched the space
- `value-at <space> <entity> --as <DID>` — what a given identity actually sees

**Conflicts / async** — `conflicts <space> [entity]` distinguishes **real multi-user contention** (≥2 principals) from **benign multi-session** activity (one user, many tabs), with stale-read / lost-update detection inferred from the committed log (`commit.original.reads` vs resolution seqs).

## Shape & dependencies

Self-contained package: depends only on `@db/sqlite`, `@std/assert`, and `@commonfabric/memory/v2/patch` (the shared server-side patch applier, reused so reconstruction matches production semantics exactly). Adds the package to the root workspace and exports `./v2/patch` from `@commonfabric/memory`.

Includes a compatibility shim for pre-`scope_key` DBs (older spaces predate the per-scope column).

## Tests

13 hermetic test files / 53 steps, all green. Each seeds its own throwaway SQLite DB — no fixtures, no network, no live runtime.

Design notes under `docs/plans/2026-06-*-state-inspector-*.md`.

---
## Summary by cubic
Adds `@commonfabric/state-inspector`, an offline, read-only inspector for memory v2 SQLite space DBs, surfaced as `cf inspect` and a single-file HTML explorer. Ships a practical `state-inspector` skill and rewritten README; planning notes removed.

- New Features
  - Offline reconstruction and time travel: value-at-seq, diffs, timelines.
  - Multiplayer + cross-space: scope-aware views (space/user/session), identity worlds, convergence checks.
  - Conflict analysis: multi-user vs multi-session; anomalous stale-read detection by replaying engine checks.
  - Unified model/graph/detail: classification, lineage, links, schema, CFC; DOT export.
  - HTML explorer and CLI with `--json` and local DB auto-discovery.
  - Docs: added `skills/state-inspector/SKILL.md` and rewrote `packages/state-inspector/README.md` as user docs; removed planning docs.

- Bug Fixes
  - Reconstruction parity: exact branch read path; snapshot table as base (with legacy `snapshot`/`revision` scope_key shims); child-local patch over inherited applies to `{}`; `forkSeq ?? 0`; whole-seq reads (removed op-index cut); parity-tested against `engine.read()` at head and mid-history.
  - Conflict fidelity: default read branch is `read.branch ?? commit.branch`; uses engine `resolveScopeKey` and exported `patchOverlapsRead`/`patchOverlapsNonRecursiveRead`; stale-read analysis cap is labeled in reports.
  - Divergence correctness: hash RAW reconstructed values via `hashStringOf` (no annotated-value hashing) to avoid false converges on deep diffs or BigInt.
  - Modern data decoding/export: recognize `FabricLink`; `annotate` stays JSON-safe for BigInt, symbols/functions, Fabric instances, and `undefined`; safe JSON export escapes U+2028/U+2029.
  - Scope/value semantics: scope keys via `resolveScopeKey`; `value-at --as` remains an approximation; presence checks honor `atSeq`.
  - Stability and perf: entity timelines are O(n) via incremental patch replay; per-entity resilience to corrupt `fvj1` rows; robust HTML escaping; fixed CLI `--json` parsing; space-qualified graph keys; precise timelines; multispace tombstone/absent-link labels; discovery dirname fallback.

<sup>Written for commit e93eb4223148cc47fa5eeadf5868aebf8f7e1817. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4397?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bfollington:
cf-review via Codex GPT-5, on behalf of @bfollington.

Scope: PR #4397 (`@commonfabric/state-inspector`, `cf inspect`, memory v2 patch export). Coverage: deep-read reconstruction/model/scopes/conflicts/multispace/timetravel/CLI/tests; sampled docs/html/graph/grouping. Checks run: targeted fmt/lint passed; `packages/state-inspector` test task fails on a clean cache; the same test files pass when run manually with the missing network permission added.

## Blocking

1. `packages/state-inspector/deno.jsonc:10` — the package test task cannot run from a clean cache.

`just-test` grants read/write/ffi/env but not net. On a clean `DENO_DIR`, importing `@db/sqlite` tries to download the native SQLite dylib via `@denosaurs/plug`; the checked-in `deno task test` fails before any tests run with `NotCapable: Requires net access to "github.com:443"` and then `release-assets.githubusercontent.com:443`. This will make the new workspace package fail in the root package test loop unless the dylib is already cached. `packages/memory` already uses `--allow-net` for the same SQLite dependency. Fix by adding the needed network permission to the state-inspector test task (or otherwise pre-provisioning the native library in a way the task actually uses).

Verified: `env DENO_DIR=/private/tmp/deno-cache-labs-pr-4397 mise exec -- deno task test` fails in `packages/state-inspector`; `deno test --allow-read --allow-write --allow-ffi --allow-env --allow-net` over the package tests passes 13 files / 53 steps.

2. `packages/state-inspector/multispace.ts:108`, `packages/state-inspector/timetravel.ts:31`, `packages/state-inspector/scopes.ts:300` — Fabric value equality is reimplemented with sorted `JSON.stringify`.

This is exactly the kind of dev-tool fork we should avoid. The durable store can contain `fvj1` values that decode to BigInt, FabricHash/FabricBytes/FabricRegExp/FabricInstance wrappers, sparse arrays, `undefined`, etc. JSON canonicalization is not the runtime/data-model equality primitive: it throws on BigInt, erases or misrepresents several Fabric types, and turns convergence/diff/overlay into a different data model than production. Use the canonical data-model helpers for comparisons/keys (`@commonfabric/data-model/value-hash` or `valueEqual`) and keep `value-debug`/annotation only for presentation.

Verified: `diffValues(1n, 1n)` throws `Do not know how to serialize a BigInt`; in `convergence()` the same class of error gets collapsed into the `"«decode-error»"` bucket rather than reporting equality.

3. `packages/state-inspector/scopes.ts:130` and `docs/plans/2026-06-29-state-inspector-multiplayer.md:36` — `value-at --as` claims "what an identity actually sees" by applying `session -> user -> space` most-specific-wins, but the runtime reads one declared scope at a time.

The production path resolves the declared cell/read scope to one `scope_key` with `Engine.resolveScopeKey()` and reads only that partition. The earlier model doc even states this correctly: `PerSpace`/`PerUser`/`PerSession` readers read only their matching `scope_key` rows. Given only an entity id, the inspector cannot know that the runtime would read `user` instead of `space`; that depends on the link/schema/cell scope that led to the read. The current implementation can therefore report a per-user row as "what user X sees" even when the real read of a space-scoped link would return the space row.

Fix by making `value-at` resolve an explicit declared scope (`space`/`user`/`session`) through the canonical `Engine.resolveScopeKey()` path, or by starting from a real link/schema that carries the scope. Keep `overlay` as a useful divergence view, but label it as a synthetic scope overlay unless it is backed by the actual declared scope. Also remove the local `encodeScope()` string replacement in favor of the exported runtime helper.

Verified by reading `packages/memory/v2/engine.ts:1447-1454` and `packages/runner/src/scope.ts:25-53`.

4. `packages/state-inspector/conflicts.ts:204` — stale-read/lost-update analysis reverse-engineers the server conflict check and drops scope/path semantics.

`entityConflicts()` prefilters commits by raw id, decodes `reads.confirmed`, keeps reads with the same id, then compares them to every writer for the requested `scope_key`. Production does more: it resolves each read's declared scope through the writer principal/session (`Engine.resolveScopeKey(read.scope, scopeContext)`) and checks conflict against the read path plus `nonRecursive` semantics. As written, the inspector can flag a stale read for the same id in the wrong scope, and it can flag unrelated path writes that the server would not consider conflicting. That makes the tool overstate "lost-update" evidence from the durable log.

Fix by moving/exposing a read-only conflict-analysis helper from memory v2 and calling it here, or at least reuse the same scope-key and patch-overlap helpers and downgrade the UI wording to "write interleaving heuristic" until it is runtime-faithful. Add tests for same id across `space`/`user` scopes and non-overlapping patch paths.

Verified by comparing `packages/state-inspector/conflicts.ts:204-230` with `packages/memory/v2/engine.ts:3498-3515` and `packages/memory/v2/engine.ts:3572-3610`.

Verdict: strong direction, but not ready to merge. The core issue is not UI polish; the dev tool is presenting several derived answers as production truth while using duplicated equality, scope, and conflict machinery instead of the runtime/data-model implementations.

bfollington (commented):
cf-review via Claude Opus 4.8, on behalf of @bfollington

**Scope** · the whole PR surface (~9.1k lines, new `packages/state-inspector` + `cf inspect` CLI), reviewed in four parallel passes — (A) decode/reconstruct/model core, (B) scope/identity/conflict multiplayer, (C) detail/graph/html/cli + packaging, (D) tests & fixture fidelity. **Coverage** · deep-read all source + tests, each pass compared against the real runtime (`packages/memory/v2/engine.ts`, `v2/patch.ts`, `packages/data-model/codec-json`, `runner/src/{scope,cfc}`, `runner/src/data-updating.ts`) rather than taking the tool's word. **Checks run** · `deno fmt --check` + `deno lint` clean; `deno task test` 13 files / 53 steps green (942ms); standalone repro of the `JSON.stringify`/BigInt footgun; manual verification of the reconstruction and scope-precedence findings against engine source.

## The headline (this is the thing to read)

The owner's bar was: **reuse the runtime's own machinery, don't reverse-engineer it, and represent the data model honestly.** On the hardest axis the PR passes well — it reuses the canonical patch applier (`@commonfabric/memory/v2/patch`) and the canonical value codec (`@commonfabric/data-model/codec-json`), and **forks nothing** in the hash / serialize / clone / DID family (the usual footguns). The test fixtures faithfully mirror the real `commit`/`revision` DDL and write shapes — the suite is **not** green against a fiction.

The recurring weakness is narrower and worth internalizing: **in the few places the tool re-derives engine logic, it overstates its fidelity in the doc-comments.** Three views (reconstruction of a patch-first entity, stale-read detection, and "view as identity" scope precedence) describe themselves as mirroring the runtime but actually diverge from it. For a tool whose entire value proposition is *honest representation*, a comment that claims runtime-parity it doesn't have is the most important class of finding here. The fix is usually one of: reuse the engine's exported function, or relabel the view as an explicit approximation.

## 🔴 Blocking

**1. `reconstruct.ts:89-94` — a patch-first entity reconstructs differently from the engine. (verified against engine.ts:4172-4202)**
The replay loop skips a `patch` when `doc === undefined` ("the server always lays a set first"). The engine does **not** assume that: `reconstructPatchedDocument` seeds `document = emptyEntityDocument()` (`{}`, engine.ts:4174,4301) and applies patches onto it even with no base `set` (engine.ts:4197-4202). So for any entity whose base revision is a `patch`, the engine yields `applyPatch({}, ops)` while the inspector returns the wrong/absent doc — exactly the honest-representation drift this tool exists to avoid. Reachability depends on whether patch-first entities actually occur (the comment asserts they don't), so this is "resolve **or consciously accept**" — but the engine itself defends against the case, so the tool should too. *Fix:* `doc = applyPatch(doc ?? {}, ops)` instead of `continue`; fix the comment; add a parity test reconstructing a patch-first entity.

## 🟡 Improvements

**2. `conflicts.ts:170-243` — stale-read inference ignores `read.path` and `read.scope`, over-reporting vs the engine. (verified)**
The engine scopes conflicts to `(id, scope_key, path)`: a `set`/`delete` after a read always conflicts, but a `patch` conflicts **only if** its path overlaps the read (`patchOverlapsRead`, engine.ts:3895; `SELECT_PATCH_CONFLICTS`). `entityConflicts` matches writers to a read on `seq` ordering + session inequality alone, never comparing `read.path` (which `reads.confirmed[]` carries) nor resolving `read.scope`. Result: it flags disjoint-path patches and cross-scope reads the engine would accept. The file header presents this as the engine's conflict model. *Fix:* apply a patch-path overlap test and group writers by the read's resolved `scope_key`; at minimum, relabel the header as a deliberate **path-blind over-approximation**.

**3. `scopes.ts:135-176` — "view as identity" most-specific-wins is a re-derived heuristic, not the runtime's read algorithm. (verified against data-updating.ts:402-444, runner/src/scope.ts)**
`valueAsIdentity` / `resolveScopeChain` document `session ⊕ user ⊕ space, most-specific-wins` as "how the runtime resolves what an identity sees." The engine has no read-time scope fallback — `reconstructPatchedDocument` reconstructs strictly within one `scope_key`. The real resolution happens at **write** time: a scope-narrowing write stores content in the narrow-scope instance and writes a **link** at the base-scope slot pointing to it; readers follow that link, gated by `scopeRank`/`canFollowScopedLink`. So the same id legitimately spans scopes (the tool's observation is real), but "most-specific present row wins" is the tool's invention and can diverge from the link-followed value. *Fix:* relabel as a display approximation and lean on `scopeOverlay` (side-by-side every scope), which **is** an honest view. Same nuance applies to `encodeScope` (scopes.ts:173) hand-rolling `did:key:`→`did%3Akey%3A`: outputs match `resolveScopeKey` for real DIDs/UUIDs today, but it should reuse the exported `resolveScopeKey` rather than re-derive the encoding.

**4. `timetravel.ts:31-43` (+ `html.ts:114` `safeJson`) — `JSON.stringify` leaf-comparison can throw on BigInt and silently miss symbol/instance diffs. (verified by repro)**
`canonical()` sorts + `JSON.stringify`es decoded leaves to compare them. `annotate()` only normalizes link/stream sigils; bigint/symbol/special-number/fabric-instance leaves fall through untouched, and the `fvj1:` codec round-trips exactly those. Verified: `JSON.stringify(10n)` **throws** (kills `cf inspect diff`/`timeline`, and via the identical `safeJson` the whole `cf inspect html` export); two distinct `Symbol()`s → both `undefined` → missed diff; class instance → `{}` → identity erased. *Fix:* compare via `toCompactDebugString` from `@commonfabric/data-model/value-debug`, or have `annotate()` coerce bigint/symbol/non-plain leaves to a stable string so neither the diff nor the export can throw. Worth doing before pointing the tool at arbitrary real spaces.

**5. `detail.ts:213-230` — `declaredSchemaFor` hand-walks `$ref`/`$defs` resolution that has a canonical home. (verified against runner/src/cfc/schema-refs.ts)**
The fallback does `ref.slice("#/$defs/".length)` + a single-key lookup. The canonical resolver is `ContextualFlowControl.resolveSchemaRef` (`@commonfabric/runner/cfc`), which `decodeJsonPointer`s the ref (handles `~0`/`~1`), gates on `isRootDefsSchemaPointer`, and re-attaches `$defs` so nested `$ref`s still resolve — none of which the hand-walk does, so nested/escaped refs render as a raw `{$ref}`. This is display-only best-effort and the runner is a heavier dep for an offline tool, so it's an Improvement — but per the PR's own #1 requirement, either import the canonical resolver or add a comment that this is a deliberately-narrowed `#/$defs/<name>`-only subset.

**6. `scopes.ts:41,52` + `grouping.ts:51` — principal regex hardcodes `did:key:`. (suspected, latent)**
`^session:(did:key:[^:]+):` only matches `did:key:` principals; the platform `DID` type admits `did:web:`/`did:plc:`. A non-`did:key` writer → `principal: undefined` → silently mis-counted as not-multi-user in the contention views. The runtime mints `did:key:` exclusively today so it's latent, but it's the honesty gap the tool exists to close. *Fix:* broaden to `did:[a-z]+:[^:]+` and add a `did:web` fixture asserting the principal is still extracted.

## ⚪ Nits

- `decode.ts:18-19` — the legacy plain-JSON `JSON.parse` branch is an honest *extension* (decodes pre-`fvj1` DBs the live runtime can no longer read), not a fork — but comment it as such, and route the `fvj1:` branch through `decodeMemoryBoundary` so any future context-sensitive decoding stays in lockstep.
- `scopes.ts:300` `scopeOverlay.divergent` uses key-order-sensitive `JSON.stringify`, while `multispace.ts` correctly uses a key-sorted `canonical()` — same-value/different-key-order overlays can be falsely flagged divergent. Reuse one canonical helper.
- `decode.ts:97` `annotate` truncates at `maxDepth=8`; equality/clustering in `scopeOverlay`/`convergence` therefore can't distinguish values differing only below depth 8 — worth a one-line caveat next to divergence verdicts.
- `html.ts:229` `.replace(/^cid:/,"cid:")` is a no-op; `timetravel.ts:263` selects `distinctish` but never reads it — dead. `html.ts:655,661` interpolate `${bundle.space}` into `<title>`/header un-escaped (DID from a filename, not stored-value-controlled, so no real injection, but it's the one un-escaped interpolation in an otherwise correctly-escaped file).
- `conflicts.test.ts:28` stores an un-`encodeURIComponent`'d `session_id` (`session:did:key:zAlice:s1`); real DBs store `session:did%3Akey%3AzAlice:s1`. Passes because `parseScope` decodes first, but it's the one fixture not exercising the true at-rest string. `conflicts.ts:5` doc-comment also overstates the `ConfirmedRead` shape (`scope` is a `CellScope` enum, not a top-level string).

## Branch-aware caveat (one to know about)

`reconstruct.ts` filters `branch = ?` exactly; the engine's `readRowForBranch` falls back to a parent branch via `forkSeq` when a row is absent on a child branch. Reconstructing on a **non-default** branch can miss inherited rows. Default `branch=""` matches `DEFAULT_BRANCH`, so the common path is fine — flag for any future branch-aware callers.

- **Export the engine's conflict matcher** (`findConflictSeq` / `patchOverlapsRead`) and a `parseScopeKey` inverse from `@commonfabric/memory/v2` so dev tools can reuse the *exact* conflict + scope logic instead of re-deriving it — this directly serves the "reuse the same algorithms" requirement and would retire findings 2 & 3's divergence at the source.

## Verdict

Strong, honest-leaning package with clean anti-duplication and faithful test fixtures — it reuses the canonical applier/codec and forks none of hash/serialize/clone. **One Blocking divergence** (patch-first reconstruction) to resolve or consciously accept, and a cluster of Improvements where re-derived engine logic (stale-read, scope precedence, `JSON.stringify` equality) overstates its fidelity or can throw — relabel as approximations, reuse the engine's functions where exported, and neutralize the BigInt/symbol leaf path before pointing this at arbitrary real spaces.

bfollington (commented):
cf-review via Codex GPT-5, on behalf of @bfollington.

Scope: PR #4397 at `6f2bd9065`; coverage was deep-read of `packages/state-inspector` reconstruction/model/scope/conflict/graph/HTML/CLI surfaces, delegated read-only passes over data-model fidelity, multiplayer semantics, and CLI/HTML/tests, plus targeted probes against `packages/memory/v2`. Some findings overlap with cubic's automated review; the points below were independently checked against the runtime code and/or local probes.

Checks run:

- `packages/state-inspector`: `deno task test` with repo-pinned Deno 2.8.1 -> 13 files / 53 steps passed.
- `deno fmt --check` on changed state-inspector/CLI/docs files -> passed.
- `deno lint packages/state-inspector packages/cli/commands/inspect.ts packages/cli/commands/main.ts` -> passed.
- `deno check packages/cli/commands/inspect.ts packages/cli/commands/main.ts` -> passed.

Blocking:

1. `packages/state-inspector/reconstruct.ts:77` reconstructs branch-local rows only, but runtime branch reads inherit parent state.

The inspector selects only `revision WHERE branch = ?` and replays those rows. The real memory-v2 read path recurses through `branch.parent_branch` at `fork_seq` when the child branch has no local row (`packages/memory/v2/engine.ts:4207-4246`). I verified this with a synthetic DB written by `memory/v2/engine`: after creating `feature` from default, `read(engine, { branch: "feature" })` sees the inherited document, while `reconstructDocument(..., { branch: "feature" })` returns absent.

That makes every `--branch` consumer potentially false: `value-at`, model/entity lists, details, graph, convergence, and scope overlays can report an inherited runtime-visible entity as missing. This is exactly the class of dev-tool drift we should avoid. Fix by reusing/exporting the memory-v2 materialization/read-row logic, or by moving the branch resolver and patched-document reconstruction into a shared runtime helper that the inspector calls. Tests should cover inherited untouched entity, child override, and child delete tombstone.

2. `packages/state-inspector/conflicts.ts:204-240` presents persisted commits as "stale reads / lost-update", but memory-v2 rejects those stale reads before commit insertion.

`entityConflicts()` scans successful `commit.original.reads.confirmed[]` and reports a missed write whenever another session wrote between the read seq and the reader commit seq. The real engine validates confirmed reads before inserting the commit (`packages/memory/v2/engine.ts:3186`, `3493-3521`) and uses branch, resolved scope key, path overlap, and `nonRecursive`; stale pending reads are validated too (`3535-3567`). The state-inspector test seeds an impossible committed loser manually, so it proves the analyzer's heuristic, not runtime behavior.

As written, the tool can tell an agent "this committed lost update happened" for a state the real runtime normally would not persist, and it also ignores branch/scope/path/pending-read semantics. Either remove/rename this as an invariant/corruption checker, or replay the same `findConflictSeq` semantics from memory-v2 and label hits as anomalous persisted stale reads rather than normal conflict history.

3. `packages/state-inspector/decode.ts:37-56` only recognizes legacy link sigils, so modern `fvj1` state links disappear from the model.

The PR and README say modern `fvj1:` DBs are handled. `decodeStored()` does call `valueFromJson`, but after enabling `setModernCellRepConfig(true)`, `jsonFromValue({ value: linkRefFrom(...) })` decodes to a `FabricLink` instance. `parseSigilLink()` returns `null`; `annotate()` renders `{}`; `summarize()` renders `"{}"`. That means lineage extraction in `model.ts`, outgoing links in `detail.ts`, cross-space link indexing, and graph edges can all miss real links on modern rows.

The canonical link representation seam is in `@commonfabric/data-model/cell-rep` / `FabricLink`, not a local `{"/":{"link@..."}}` parser. The inspector should handle the data-model link instances directly, or use/refactor a shared helper that extracts link payloads for both legacy and modern stored forms. Add a test that stores actual entity revision data through `jsonFromValue` with modern cell rep enabled, not just modern-looking commit metadata.

4. `packages/state-inspector/html.ts:652-666` escapes the JSON bundle but not the document shell.

`safeJson()` protects the embedded bundle at line 705, but `bundle.space`, `opsLine`, generated date text, and related header strings are interpolated directly into `<title>` and header HTML. I verified `renderInspectorHtml()` emits a malicious `space` containing `</title><script>...` and an op key containing `` raw in the output.

Because this tool opens standalone HTML generated from DB/path-derived data, the chrome needs the same escaping discipline as the JSON payload. Add an `escapeHtml()` helper for every template interpolation outside the JSON script, or render the chrome from `APP` using `textContent`. Add an `html.test.ts` case with malicious `space` and op-key strings.

5. `packages/state-inspector/timetravel.ts:31-42` reimplements fabric comparison/stringification with sorted `JSON.stringify`.

`diffValues()` sorts JS keys with `.sort()` and compares `JSON.stringify(sort(v))`. A targeted probe with `diffValues(42n, 43n)` throws `Do not know how to serialize a BigInt`, and this is also the wrong abstraction for Fabric instances/links and cross-platform key ordering. The repo already has data-model-aware value hashing/debug surfaces (`@commonfabric/data-model/value-hash`, `@commonfabric/data-model/value-debug`) that understand these values.

For a debugger, "diff" should not fork the value model. Use the canonical data-model equality/hash/debug helpers for leaf comparison and display, and only keep a structural walk where it can preserve Fabric values instead of JSON-erasing them.

Verdict: useful direction, but not safe to merge yet. The current implementation reuses the patch applier, but still reconstructs branch state, conflict semantics, link representation, and value comparison outside the runtime/data-model machinery. Those are the exact places this tool needs to be boringly identical to production.

bfollington:
## Review fixes pushed (`a6cf14be6`)

Acted on all three review passes — [cf-review (Claude Opus 4.8)](#), the automated **cubic** review, and **Codex GPT-5**. They converged on one theme, which is exactly the bar for this tool: *it re-derived engine logic and presented it as production truth.* The fix is to **reuse the runtime's own functions where they're pure, and prove parity by test where they can't be** (the engine's reconstruction is coupled to a store-*mutating* constructor, so the inspector can't just call it on a read-only autopsy DB).

`deno fmt`/`lint` clean; full workspace `deno task check` green; state-inspector suite **16 files / 67 steps** (was 13/53).

### Fidelity — reuse runtime machinery, don't re-derive

| Finding (raised by) | Resolution |
|---|---|
| **Branch reconstruction drops inherited state** (cubic, Codex, me) | `reconstruct.ts` now honors branch inheritance (parent recursion at `fork_seq`) and applies a leading `patch` onto `{}`. New **`reconstruct-parity.test.ts` drives the real `memory/v2` engine** to write DBs and asserts `reconstructDocument() == engine.read()` across inherited / override / tombstone / patch-first. |
| **Stale-read over-reports & is mis-framed** (all three) | `conflicts.ts` replays the engine's *own* check: resolves each read's scope via the exported `resolveScopeKey`, tests patch overlap via the newly-exported `patchOverlapsRead`. Reframed as **anomalous** stale reads (the engine rejects these pre-commit, so a hit = invariant violation, not routine "lost update"). New tests prove no cross-scope / disjoint-path false positives. Verified on a real space: reports the store consistent. |
| **`value-at --as` most-specific-wins isn't the runtime read** (Codex, me) | Relabeled as an explicit **approximation** (runtime resolves via base-scope link-following); `scopeOverlay` remains the honest side-by-side view. Scope-keys now built via `resolveScopeKey` (hand-rolled encoding dropped). |
| **`JSON.stringify` value equality throws on BigInt / erases Fabric** (all three) | `timetravel` / `multispace` / `scopes` now key on the data-model's `hashStringOf`. |
| **Modern `fvj1:` links vanish** (Codex) | **Could not reproduce** — verified on the 3 real modern DBs that `decodeStored` yields plain sigil objects and `parseSigilLink` hits (incl. cross-space links). The claim was contingent on `setModernCellRepConfig(true)`, which the inspector deliberately doesn't enable. |

`memory/v2/engine.ts` now exports the pure `patchOverlapsRead` / `patchOverlapsNonRecursiveRead`.

### Correctness batch (cubic)

HTML shell escaping (+ hostile-input test) · CLI `--json`-before-`<db>` parsing · timeline `op_index` precision + `DISTINCT` touched + branch filter · **space-qualified external graph node keys** (no local-id collision) + CLI prints the target space · multispace tombstone exists-flag + absent cross-space link labeling · grouping attaches cross-space referents · `discover` relative `DB_PATH` dirname fallback · scope presence `atSeq` + branch-filtered participants · `did:`-method-agnostic principal regex · dead-code/`$ref`-resolver-scope comments.

### CI

`--allow-net` on the test task so `@db/sqlite`'s dylib download works from a clean cache (matches `packages/memory`) — fixes the root test-loop failure. New `cli.test.ts` covers the single-space command dispatch.

### Noted follow-up

Extract a read-only materializer in `memory/v2` so the inspector can call the engine's reconstruction *directly* instead of a parity-tested replica — the durable single-source answer. Happy to file it if you'd like.

— cf-review fixes via Claude Opus 4.8, on behalf of @bfollington. Credit to **cubic** and **Codex** for findings incorporated above.

bfollington (commented):
cf-review via Codex GPT-5, on behalf of @bfollington.

Scope: PR #4397 latest head `a6cf14be6511bdbead45857d5bf021f662f50a1f`. Coverage: re-read the author follow-up and PR review comments; deep-read reconstruction, conflicts/scope, decode/detail/html, CLI/docs, parity tests, and the memory-v2 engine seams; delegated read-only passes over branch materialization, conflict/scope semantics, and data-model/presentation fidelity. Findings 1 and 2 overlap with cubic's second-pass review and were independently verified against runtime behavior.

Checks run:

- `packages/state-inspector`: `deno task test` with repo-pinned Deno 2.8.1 on a clean temp `DENO_DIR` -> 16 files / 67 steps passed.
- `deno fmt --check` on changed state-inspector/CLI/docs/config files -> passed.
- `deno lint packages/state-inspector packages/cli/commands/inspect.ts packages/cli/commands/main.ts` -> passed.
- `deno check packages/cli/commands/inspect.ts packages/cli/commands/main.ts` -> passed.
- Targeted runtime probes for child-branch patch materialization, stale-read branch defaulting, modern `FabricLink` decode, and BigInt HTML export.

I did not repeat cubic's `--allow-net` warning as a finding: the clean-cache package test downloaded `@db/sqlite`'s native dylib, so the test task needs net permission in the same way `packages/memory` does.

## Blocking

1. `packages/state-inspector/reconstruct.ts:150-172` still does not match memory-v2 for child-branch patches.

The follow-up fix replays every visible branch-lineage row, sorted globally by `(seq, op_index)`. That makes a parent `set` become the base for a child-local `patch`. The runtime does not compose branches that way: `readRowForBranch()` first resolves one visible row, returning a child-local row immediately when one exists (`packages/memory/v2/engine.ts:4219-4233`), then `reconstructPatchedDocument()` reconstructs patches only within that resolved branch (`packages/memory/v2/engine.ts:4158-4195`).

Verified with a normal engine-written DB: parent `set { value: { parent: 1 } }`, fork `feature`, then child `patch add /value/child 2`. `engine.read(..., branch: "feature")` returns `{ value: { child: 2 } }`, while `reconstructDocument(..., branch: "feature")` returns `{ value: { parent: 1, child: 2 } }`. The new parity test covers inherited untouched rows, child `set`, child `delete`, and default-branch patch-first, but not this child-local patch case.

Fix: mirror/export the engine materialization shape instead of replaying the whole lineage. Resolve the visible row with `readRowForBranch` semantics; if the resolved row is a patch, reconstruct from that branch's own base/snapshot/patch rows only. Add a parity case where a child branch stores only a patch for an inherited parent entity.

2. `packages/state-inspector/conflicts.ts:252-285` replays unqualified confirmed reads against the inspected branch, not the reader commit's branch.

The candidate query selects `seq, session_id, original` but not `commit.branch`, and line 285 uses `rd.branch ?? branch`, where `branch` is the branch option passed to `entityConflicts()`. Runtime validation receives the commit's branch and then defaults each read with `read.branch ?? branch` (`packages/memory/v2/engine.ts:3493-3505`), so the default must be the reader commit's branch, not the branch currently being inspected.

Verified with a seeded store: a default-branch commit with an unqualified read at seq 1 is incorrectly reported as stale when inspecting `feature`, because the analyzer checks feature writes after seq 1 even though the runtime would have resolved that read on the default branch.

Fix: select `branch` from `"commit"` for candidate reader commits, compute `const readBranch = rd.branch ?? c.branch`, and use that for the writer query. If the UI is meant to remain branch-scoped, either skip reads whose resolved `readBranch` differs from the inspected branch or include `readBranch`/`missedWriteBranch` in `StaleRead` so cross-branch reports are explicit. Add a test for a non-default commit with an omitted `read.branch`.

3. `packages/state-inspector/decode.ts:37-56`, `:97-124`, `:143-161` only understand legacy link sigils, so modern `fvj1` `FabricLink` values disappear from details, graph edges, and cross-space indexes.

The PR now claims modern `fvj1` decode is handled, but `valueFromJson()` can restore a modern link as a `FabricLink` instance. The inspector then treats it as a generic object with no enumerable keys: `annotate()` returns `{}`, `summarize()` returns `{}`, and `collectLinks()` returns no links. I verified this by encoding `linkRefFrom({ id: "of:target" })` with `setModernCellRepConfig(true)`, then decoding it through `decodeStored()` with the flag reset; the restored value is a `FabricLink`, but the inspector link index sees zero links.

This means modern stored links can be absent from `detail.ts` out-links, graph/model classification, and multispace cross-space link analysis. Fix by reusing the data-model link seam (`isLinkRef`/`linkRefPayload` or `FabricLink`) before generic object walking, normalizing both legacy sigils and modern instances into the existing `$link`/`DecodedLink` shape. Add tests that store actual modern `jsonFromValue()` output, not only legacy-looking sigils inside an `fvj1` envelope.

4. `packages/state-inspector/html.ts:113-114`, `packages/state-inspector/decode.ts:97-124`, `packages/cli/commands/inspect.ts:61,863,885` still plain-`JSON.stringify` Fabric values that can contain `bigint`.

The `hashStringOf` fix handles equality, but the exported/presentation path still is not JSON-safe. `decodeStored()` round-trips `fvj1` BigInt via the real memory boundary, and `annotate()` returns that `bigint` unchanged. I verified a DB written through `memory/v2/engine` with `{ value: 1n }`: `buildInspectorBundle()` contains a `bigint`, and `renderInspectorHtml()` throws `TypeError: Do not know how to serialize a BigInt`. The same shape affects CLI JSON output and human `value-at --as`, which also call plain `JSON.stringify`.

Fix: make the boundary into machine/export JSON explicit and Fabric-safe. Either emit lossless `codec-json` strings for machine JSON fields, or lower Fabric primitives/instances into tagged display records before `JSON.stringify`. Use `value-debug` for human text only. Add CLI `--json` and HTML tests for BigInt plus modern `FabricLink`.

## Improvement

`packages/cli/commands/inspect.ts:836-863` and `packages/state-inspector/README.md:158` still present `value-at --as` as "Read AS this identity" / `resolved from:` without the approximation caveat. `scopes.ts` now correctly labels `valueAsIdentity()` as an approximation, but the user-facing command still sounds like runtime read resolution. Please change the help/README/human output to "approximate most-specific stored scope for this identity" or similar, and reserve "runtime read" wording for a path that follows the declared-scope/link resolution semantics.

Verdict: the follow-up fixed several important forks, and the local checks are clean, but this still is not ready to merge under the stated bar. The remaining issues are exactly in the places the tool needs to be boringly identical to the runtime/data model: branch materialization, confirmed-read branch semantics, modern link representation, and JSON-safe Fabric export.

bfollington (commented):
cf-review via Codex GPT-5, on behalf of @bfollington.

Quick re-review update: I fetched the live PR state again and the head is still `a6cf14be6511bdbead45857d5bf021f662f50a1f` — no commits have landed after the detailed review I posted above at `2026-06-29T01:20:29Z`.

I did a focused divide-and-conquer confirmation pass on the same head rather than reposting the whole report. The prior findings still stand:

- `reconstructDocument()` still replays parent lineage rows before child-local patch rows; memory-v2 resolves one branch row first and reconstructs patches within that resolved branch only.
- `entityConflicts()` still defaults unqualified confirmed reads to the inspected branch (`rd.branch ?? branch`) instead of the reader commit's branch.
- Modern `fvj1` `FabricLink` values are still missed by the sigil-only link parser / `collectLinks()`.
- BigInt still reaches plain `JSON.stringify` in CLI/HTML export paths.
- The `value-at --as` CLI/README wording still presents the approximation too much like a runtime read.

No additional code changes to review since the last comment, so the previous detailed blocking review remains the current review state.

bfollington:
## Second review round pushed (`d6eda7380`)

Acted on cubic's and Codex's follow-up. Codex's counterexample was correct and **disproved my earlier reconstruction-parity claim** — thank you; that's exactly the kind of catch the parity harness exists to force, and the test now covers it.

19 test files / 69 steps; `deno fmt`/`lint`/workspace `check` green; graph links + `value-at` re-verified on real modern DBs.

| Finding (by) | Resolution |
|---|---|
| **Child-local patch reconstruction ≠ engine** (Codex, cubic P1) | **Rewrote `reconstruct.ts`** to faithfully replicate `readRowForBranch` + `reconstructPatchedDocument`: resolve ONE visible row (child→parent inheritance only when the child has no row of its own), then reconstruct patches *within the resolved branch* (base = latest `set`/`delete`; a delete or no-base → `{}`). A child-local patch on an inherited entity now reconstructs onto `{}` — **not** the inherited parent value. Verified against Codex's exact case (`parent set {parent:1}`, fork, child `add /value/child 2` → `{child:2}`) in the parity test driving the real engine. The prior log-merge is gone. |
| **Unqualified read resolved to inspected branch** (Codex, cubic P2) | `conflicts.ts` now defaults a confirmed read's branch to the **reader commit's** branch (`read.branch ?? commit.branch`), matching `validateConfirmedReads`. Selects `commit.branch`, records `readBranch` on each `StaleRead`. New test: a default-branch read isn't flagged when inspecting `feature`. |
| **Modern `fvj1` `FabricLink` values vanish** (Codex) | `decode.ts` now recognizes `FabricLink` instances directly (`cell-rep`'s `isLinkRef` is gated on a global modern-mode flag the inspector doesn't set, so I check the class). New `decode.test.ts` includes a **real `jsonFromValue` round-trip with modern cell-rep on**. (Note: my probe of the 3 real on-disk modern DBs still decodes to sigil objects — current data was already covered — but this makes it robust to the instance form regardless.) |
| **BigInt reaches `JSON.stringify` in export paths** (Codex) | `annotate()` now lowers `BigInt → {$bigint}`, symbols/functions, and Fabric instances → a `value-debug` string, so the bundle and every `JSON.stringify` path (HTML `safeJson`, CLI `--json`, `value-at`) are JSON-safe. Tested. |
| **`value-at --as` wording overclaims** (Codex) | CLI help, human output, and README all relabeled as an explicit **approximation** (most-specific stored scope), pointing to `overlay` for ground truth. |
| **`--allow-net` unnecessary** (cubic P3) | **Keeping it** — it's the test/inspect task permission for `@db/sqlite`'s native-dylib download on a clean cache (the failure Codex independently reproduced); `packages/memory` uses it for the same reason. Source needs no network. (Codex explicitly didn't repeat this as a finding.) |

— review fixes via Claude Opus 4.8, on behalf of @bfollington. Findings from **cubic** and **Codex**.

bfollington (commented):
cf-review via Codex GPT-5, on behalf of @bfollington.

Scope: final pass on PR #4397 latest head `d6eda73809804f2e6f9ddc6e0746d5c6908ba4a1`. Coverage: re-read the author/cubic follow-up context; deep-read the latest changes in reconstruction, conflicts, decode/export, CLI wording, and tests; delegated focused read-only passes over branch materialization, conflict/scope semantics, and Fabric value/export fidelity.

Checks run:

- `packages/state-inspector`: `deno task test` with repo-pinned Deno 2.8.1 on a clean temp `DENO_DIR` -> 19 files / 69 steps passed, including the SQLite native dylib download.
- `deno fmt --check` on the latest changed files -> passed.
- `deno lint packages/state-inspector packages/cli/commands/inspect.ts packages/cli/commands/main.ts` -> passed.
- `deno check packages/cli/commands/inspect.ts packages/cli/commands/main.ts` -> passed after fetching uncached CLI deps outside the sandbox.
- Targeted probes for nullable `fork_seq`, modern `FabricLink` structural lineage, and `fvj1` `undefined` export.

This round does fix important prior blockers: child-branch patch materialization now resolves one visible row and reconstructs within that branch; unqualified confirmed reads now default to the reader commit's branch; the main `value-at --as` help/output is explicit about approximation; modern links are handled in generic `collectLinks()`/`annotate()`; and BigInt no longer throws in the checked HTML/JSON paths.

## Blocking

1. `packages/state-inspector/model.ts:89-91`, `:197-198`; `packages/state-inspector/detail.ts:249`, `:351`, `:494` — modern `FabricLink` support is still incomplete because structural paths still call the legacy-only sigil parser.

`decode.ts` now has `decodedLinkOf()` and generic value link collection works for modern `FabricLink` instances, but the model/detail code still uses `parseSigilLink()` directly for piece `argument`, `internal`, result/context naming, and detail out-links. That means modern stored links can be visible in raw value collection while disappearing from the higher-level model that the debugger is mostly selling: piece lineage, owned-cell labels, detail out-links, and the `argument`/`owns` graph edges that depend on `classifyDocument()`.

Verified with a direct probe on this head: a document with `argument: new FabricLink({ id: "of:input" })` and `internal: [{ link: new FabricLink({ id: "of:owned" }) }]` produces `classifyDocument(doc).lineage === {"internal":[],"pattern":...}`, while `collectLinks(doc.value)` does see a modern value link. Fix by replacing the remaining direct `parseSigilLink()` calls with the same normalized modern/legacy helper (`decodedLinkOf()` or a shared `linkId()` wrapper), and add tests where actual `jsonFromValue()` modern data drives piece `argument`/`internal`/result lineage, not only generic value links.

2. `packages/state-inspector/decode.ts:121-160`, plus JSON boundaries at `packages/state-inspector/html.ts:113-114`, `packages/cli/commands/inspect.ts:61`, and `packages/state-inspector/cli.ts:76` — `undefined` is still not lowered before JSON export. This overlaps cubic's latest finding and I independently verified it.

`fvj1` can faithfully decode Fabric `undefined` values, but `annotate()` returns JS `undefined` unchanged. Plain `JSON.stringify` then drops object fields, converts array entries to `null`, or returns no JSON at all for a top-level undefined. Verified on this head: `decodeStored(jsonFromValue({ value: { x: undefined, arr: [undefined] } }))` preserves both undefined values, but `JSON.stringify(annotate(decoded))` emits `{"value":{"arr":[null]}}` and `JSON.stringify(annotate(undefined))` returns `undefined`.

For a debugger, that is a false representation of the stored value. Add an explicit sentinel before the BigInt/symbol/function handling, e.g. `{ $undefined: true }`, and cover object-field, array-element, and top-level cases through the CLI/HTML JSON paths.

3. `packages/state-inspector/reconstruct.ts:123` — `NULL fork_seq` inheritance still diverges from the engine. This overlaps cubic's latest finding and I independently verified it against the runtime.

The runtime's `readRowForBranch()` caps parent inheritance with `Math.min(seq, branch.forkSeq ?? 0)` (`packages/memory/v2/engine.ts:4240`). The inspector uses `Math.min(atSeq, b.fork_seq ?? atSeq)`, so a branch row with `parent_branch` set and `fork_seq NULL` inherits parent state up to the requested seq, even though the runtime would cap inheritance at seq 0 and hide it.

Verified with a normal engine DB after mutating only the branch row to the nullable shape allowed by the schema/type: parent writes `{ value: { parent: 1 } }`, child `feature` has `parent_branch = ''` and `fork_seq = NULL`; `engine.read(..., branch: "feature")` returns `null`, while `reconstructDocument(..., branch: "feature")` returns the parent document. Fix the fallback to `b.fork_seq ?? 0` or, better, share/export the runtime branch-row resolver. Add a parity test for `parent_branch IS NOT NULL AND fork_seq IS NULL`.

## Improvements

- `packages/state-inspector/scopes.ts:51-65` says it handles any DID method, but it decodes the full scope key and then matches `did:<method>:[^:]+`. `parseScope("user:did:web:example.com:user:alice")` returns `kind: "other"`, and the session form truncates the principal to `did:web:example.com`. Since `resolveScopeKey()` percent-encodes principal/session as separate colon-delimited parts, parse the raw stored key into encoded components first, then decode each component.

- `packages/cli/commands/inspect.ts:287-289` and `:363-366` still have tips saying `value-at --as` "reads" as an identity. The main help/output is now honest; these tips should use the same approximation wording and point to `overlay` for the truth view.

Verdict: this is much closer and several earlier blockers are fixed, but I would not merge yet under the stated runtime-fidelity bar. The remaining issues are small in code size but high in semantics: modern links still vanish from structural views, Fabric `undefined` is silently mis-exported, and one branch inheritance edge still disagrees with the real engine.

bfollington:
## Self-review (adversarial subagent pass) pushed (`f0a4d5062`)

Since this tool will be depended on, I ran a 4-way adversarial verification — most agents drove the **real `memory/v2` engine** to build edge-case DBs and diffed against the inspector (**2246 engine-vs-inspector reconstruction probes**, 2245 matching; the one mismatch was a malformed-DB-only `fork_seq` fallback, fixed). Reconstruction, conflict, and scope replay came back faithful. It still surfaced real fixes — including a wrong-answer my own round-2 change had introduced:

- **🔴 False-converge (divergence verdict).** `scopeOverlay`/`convergence`/`diff` were hashing the *display-annotated* value, so two values differing only below the depth-8 annotate cap — or a `BigInt 10n` vs the literal `{$bigint:"10"}` it lowers to — collapsed to "converged". For a divergence tool that's the worst failure. Now they hash the **raw** reconstructed value (`hashStringOf` is depth-complete + fabric-aware); annotate is display-only; diff/timeline compare at a deep bound. Regression test added.
- **Snapshot base.** `reconstructWithinBranch` now reads the `snapshot` table as base (engine semantics + future-proof against any revision compaction), guarded for legacy DBs. New parity case drives ≥10 patches so a snapshot row exists and asserts `reconstructDocument == engine.read()` at head + mid-history.
- **Perf.** `entityTimeline` was O(writes²) (re-reconstructing per revision — would hang on a hot loom entity); now an O(n) incremental replay via the canonical `applyPatch`.
- **Resilience.** A corrupt `fvj1` row no longer aborts a whole command/bundle — the remaining per-entity paths degrade gracefully like the others.
- **Truncation honesty.** The HTML stale-read cap left the 61st+ multi-user cell rendering as "clean"; now tagged "not analyzed (cap)".
- **Footguns.** `fork_seq` null fallback matches the engine (`?? 0`); `safeJson` escapes U+2028/U+2029.

20 test files / 70 steps; fmt/lint/workspace-check green; timeline + HTML report + graph re-verified on real DBs.

**Deferred as sound follow-ups** (not blockers): pending-read anomaly parity (engine already rejects stale pending reads pre-commit, so a healthy store has none); extracting a read-only materializer in `memory/v2` so reconstruction is shared code not a parity-tested replica; surfacing the remaining list caps in CLI output.

— self-review via Claude Opus 4.8, on behalf of @bfollington.

bfollington:
## Docs + skill added, and cubic's latest pass addressed (`e93eb4223`)

**Docs/skill (per the ask):** dropped the internal `docs/plans/*` planning docs, rewrote the package README as user documentation, and added an agent **`state-inspector` skill** (`skills/state-inspector/SKILL.md`, symlinked into `.claude/` + `.agents/` per convention) — a map+values debugging guide (the non-derivable mental model + the ground-truth-vs-approximation contract + a symptom→command map), following `docs/development/skill-authoring.md`.

**cubic's two follow-up passes** — three valid P2s, all fixed:
- **Legacy snapshot `scope_key`**: a pre-`scope_key` DB can have a `snapshot` table without the column, so the new snapshot-base query would hit a missing column. Extended the db.ts compat shim to shadow `snapshot` too; the legacy db.test now seeds a snapshot + patch and asserts it reconstructs.
- **Same-seq snapshot vs op-index cut**: the only caller that used an op-index cut (`entityTimeline`) became incremental replay last round, so `atOpIndex` was dead — removed it. All reads are now whole-seq, so a `seq <= rowSeq` snapshot base can't leak later same-commit ops.
- **`annotate(undefined)`**: a stored `undefined` was left as JS `undefined` (dropped by `JSON.stringify`, losing present-undefined vs absent-key). Now lowered to `{$undefined:true}` like the other non-JSON-safe leaves; test added.

The `--allow-net` P3 stands as a known false positive — it's the `@db/sqlite` native-dylib-download permission (`packages/memory` uses it identically; Codex agreed).

21 test files / 70 steps; fmt/lint/workspace-check green.

— via Claude Opus 4.8, on behalf of @bfollington.

PR #4404, opened 2026-06-29
Reject cell-like casts in the transformer

Hixie:
This changes the cast validation diagnostic for cell-like wrapper types from a warning into an error.

Casting to `Cell`, `Writable`, `Stream`, and related wrapper types can hide whether a value is actually a reactive cell. The transformer now rejects those casts and points callers toward type annotations instead.

The change also migrates existing pattern and test code away from those casts. It replaces them with typed locals, typed helper parameters, and typed child-cell variables.

Validation now covers direct casts, renamed imports, local type aliases, import type syntax, interfaces extending cell wrappers, and unrelated local types named `Cell`.

---
## Summary by cubic
Reject casts to cell-like wrappers with an error and expand detection to renamed imports, type aliases, unions/intersections, import types (including qualified), and interfaces. Centralize Common Fabric module/import/declaration checks and migrate patterns and tests to type annotations; keep Gmail utilities SES-safe.

- **Refactors**
  - Promote `cast-validation:cell-cast` to error; resolver now handles parenthesized/union/intersection forms, renamed imports, local aliases, import types (qualified and unqualified), and interfaces extending wrappers; only flags framework-owned `Cell`/`Writable`/`Stream` types. Simplify interface/symbol resolution fallbacks and share Common Fabric module/import/declaration predicates in `common-fabric-symbols`; use them in cast validation. Update docs and add tests for helper predicates, framework module declarations, qualified import types, non-framework exemptions, unresolved/qualified references, and recursive aliases.
  - Replace `as Cell<>`/`as Writable<>`/`as Stream<>` with typed locals/params across authored patterns, generated fixtures, and CI tests; preserve optional child types with `T | undefined`. Email task engine: pass the computed prompt directly and wrap the suggestion schema with `schema()`.

- **Migration**
  - Replace `value as Cell<T>` with `const valueCell: Cell<T> = value;` and annotate variables/params directly; import-type and alias casts are blocked. Use `T | undefined` where children may be absent.

<sup>Written for commit 71de232041e1006caa77fa41446f229f7a3a7aed. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4404?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/generated-patterns/integration/patterns/counter-aggregate.pattern.ts
+++ b/packages/generated-patterns/integration/patterns/counter-aggregate.pattern.ts
@@ -11,7 +11,7 @@ const adjustCounter = handler(
   ) => {
     const index = event?.index ?? 0;
     const amount = typeof event?.amount === "number" ? event.amount : 1;
-    const target = context.counters.key(index) as Cell<number>;
+    const target: Cell<number> = context.counters.key(index);
     const current = target.get() ?? 0;
     target.set(current + amount);
   },
--- a/packages/generated-patterns/integration/patterns/counter-aggregator.pattern.ts
+++ b/packages/generated-patterns/integration/patterns/counter-aggregator.pattern.ts
@@ -53,9 +53,10 @@ const adjustCounter = handler(
       return;
     }
 
-    const entryCell = context.counters.key(index) as Cell<CounterEntry>;
-    const valueCell = entryCell.key("value") as Cell<number>;
-    const current = typeof valueCell.get() === "number" ? valueCell.get() : 0;
+    const entryCell: Cell<CounterEntry> = context.counters.key(index);
+    const valueCell: Cell<number | undefined> = entryCell.key("value");
+    const value = valueCell.get();
+    const current = typeof value === "number" ? value : 0;
     const delta = typeof event.delta === "number" ? event.delta : 0;
     const nextValue = typeof event.set === "number"
       ? event.set
--- a/packages/ts-transformers/src/transformers/cast-validation.ts
+++ b/packages/ts-transformers/src/transformers/cast-validation.ts
@@ -4,14 +4,14 @@
  * Validates type assertions (casts) and reports diagnostics for problematic patterns:
  * - `as unknown as X` (double-cast): ERROR - bypasses type safety
  * - `as OpaqueRef<...>`: ERROR - framework handles OpaqueRef wrapping automatically
- * - `as Cell<...>` and other cell-like types: WARNING - prefer proper type annotations
+ * - `as Cell<...>` and other cell-like types: ERROR - prefer proper type annotations
  */
 import ts from "typescript";
 import { spellingsWhere } from "@commonfabric/schema-generator/wrapper-names";
 import { HelpersOnlyTransformer, TransformationContext } from "../core/mod.ts";
 
 /**
- * Cell-like types that should trigger a warning when cast to.
+ * Cell-like types that should trigger an error when cast to.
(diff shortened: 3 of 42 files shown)


packages/generated-patterns/integration/patterns/counter-aggregate.pattern.ts
```diff
@@ -11,7 +11,7 @@ const adjustCounter = handler(
   ) => {
     const index = event?.index ?? 0;
     const amount = typeof event?.amount === "number" ? event.amount : 1;
-    const target = context.counters.key(index) as Cell<number>;
+    const target: Cell<number> = context.counters.key(index);
```
mathpirate, line 14: just curious, is this type annotation necessary or does inference get it? seems naively like inference should get it, if not i might poke at the type structure of `.key(value)` to better understand it

packages/generated-patterns/integration/patterns/counter-aggregator.pattern.ts
```diff
@@ -53,9 +53,10 @@ const adjustCounter = handler(
       return;
     }
 
-    const entryCell = context.counters.key(index) as Cell<CounterEntry>;
-    const valueCell = entryCell.key("value") as Cell<number>;
-    const current = typeof valueCell.get() === "number" ? valueCell.get() : 0;
+    const entryCell: Cell<CounterEntry> = context.counters.key(index);
```
mathpirate, line 56: same questions here

packages/ts-transformers/src/transformers/cast-validation.ts
```diff
@@ -186,48 +186,238 @@ export class CastValidationTransformer extends HelpersOnlyTransformer {
...
+    typeName: ts.EntityName,
+    context: TransformationContext,
+    seenSymbols: Set<ts.Symbol>,
+  ): string | undefined {
+    const symbol = context.checker.getSymbolAtLocation(typeName);
+    if (!symbol) return undefined;
+    return this.resolveWrapperSymbolName(symbol, context, seenSymbols);
+  }
+
+  private resolveWrapperSymbolName(
```
mathpirate, line 322: i'm wondering whether there's already a function somewhere in the transformers that does this or something similar to it. i'd be a little surprised if not, but it's definitely possible the need here is sufficiently bespoke to merit a fresh implementation rather than adapting/reusing an existing one. worth an agent scan i think. similar for the rest of this method chain, although this layer is the one that seems intuitively most likely to be implemented somewhere to me

packages/generated-patterns/integration/patterns/counter-aggregate.pattern.ts
Hixie: it is not strictly necessary. codex included it because we had an explicit type before.

packages/ts-transformers/src/transformers/cast-validation.ts
Hixie: Codex took this as a personal challenge: https://github.com/commontoolsinc/labs/pull/4404/changes/bd7ae2355a3abf9366074e909d0409237314f430

Hixie:
@mathpirate landing; let me know if there was anything further you wanted to change and i'll do it as a follow-up.

PR #4405, opened 2026-06-29
Fix lunch poll contention vote swatches

willkelly:
## Summary
- fix storage v2 pending replay so surviving pending patches can materialize over stale non-container bases without weakening normal clone semantics
- add runner regressions for null, Fabric primitive, and numeric array-spine replay after dropped parent materialization
- tighten Lunch Poll browser assertions and add a high-contention late-observer regression comparing stored votes to rendered All-options swatches

## Repro / proof
- Live Lunch Poll probe showed 50 stored votes but 42 rendered All-options swatches
- Local headed contention repro hit `CloneForMutationError: cloneForMutation: cannot descend into null at path index 0`
- New contention test drives the 5 voter x 10 option shape and verifies stored votes match rendered swatches for active voters plus a fresh observer

## Verification
- LSP diagnostics clean on changed files
- `deno fmt --check packages/runner/src/storage/v2.ts packages/runner/test/memory-v2-stacked-commit.test.ts packages/patterns/integration/lunch-poll-vote.test.ts packages/patterns/integration/lunch-poll-contention.test.ts`
- `deno test -A packages/runner/test/memory-v2-stacked-commit.test.ts`
- `deno check packages/runner/src/storage/v2.ts packages/runner/test/memory-v2-stacked-commit.test.ts packages/patterns/integration/lunch-poll-vote.test.ts packages/patterns/integration/lunch-poll-contention.test.ts packages/patterns/lunch-poll/main.tsx packages/patterns/lunch-poll/poll-option-card.tsx`
- headed Lunch Poll contention and vote integration tests passed earlier against local offset-91 servers after the storage fix

## Notes
- Browser tests remove their created pieces during cleanup
- No force push or history rewrite used

---
## Summary by cubic
Fixes the Lunch Poll mismatch where stored votes exceeded the All-options swatches by repairing pending replay and adding DOM hooks so swatches bind to the correct option rows. All voters and a fresh observer now render every stored swatch; CI runs a dedicated high-contention test and gates staging on it.

- **Bug Fixes**
  - runner: Replays surviving pending patches over stale bases by rebuilding a mutable spine only during pending replay; supports null roots, object-backed primitives, numeric array spines, and array-append markers ("-"); rejects non-index array keys with a clear error; adds `cloneWithPendingReplayValueAtPath`.
  - Lunch Poll UI: Adds `data-all-option-id`, `data-all-option-title`, `data-vote-swatch-option-id`, `data-vote-swatch-type`, and `data-option-id` so swatches attach to the correct All-options row.

- **Tests & CI**
  - Adds a high-contention 5×10 browser test verifying option-specific swatch membership for active voters and a fresh observer; tightens `lunch-poll-vote` to assert swatch option id, title, voter, and color; stabilizes setup and readiness.
  - CI: dedicated `pattern-lunch-poll-contention-test` job with coverage artifact and timing; excluded from shard selection via `INDEPENDENT_PATTERN_INTEGRATION_FILES`; staging deploy is gated on this job; coverage docs and perf tooling updated to track the new job.

<sup>Written for commit ca61e82bf05a3689e467cfc7937e494d3822d330. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4405?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


willkelly (commented):
## Review: storage replay fix is correct; runner regression test should be hardened

I checked out the branch and ran the red/green directly: reverting only `storage/v2.ts` to its pre-fix state (keeping the new tests) makes the three `memory-v2-stacked-commit.test.ts` cases fail with the documented `CloneForMutationError`, and they pass with the fix. The fix itself is well-scoped — it only affects the local optimistic-visibility path (`applyPendingVersion` → `materializedVersionThroughPending` → `visibleVersion`), not commit construction, and the server stays authoritative (the conflict-outcome tests still end in `assertConflict`). No concerns with the change.

Two follow-ups on the **runner** regression test before merge. These are about robustness/coverage, not correctness of the existing assertions.

### 1. Failure surfaces as a dangling-promise module crash, not a localized assertion

Without the fix, the regression doesn't fail via `assertEquals(afterDrop.value, {right:2})`. It throws asynchronously through a subscription during `await assertConflict(parent)` and surfaces as:

```
./packages/runner/test/memory-v2-stacked-commit.test.ts (uncaught error)
error: (in promise) CloneForMutationError: ...
FAILED | 2 passed | 4 failed   # 3 of those are cancelled sibling tests
```

So a future regression produces mis-attributed, noisy output (sibling tests blamed + cancelled) instead of a clean "expected X, got throw", and the trigger is timing-dependent on the notification path firing.

**Ask:** add a direct, synchronous unit test of the new helper (`cloneWithPendingReplayValueAtPath` / `applyPendingVersion`) as a pure function — stale non-container base + surviving child patch → assert the rebuilt result. That pins the contract deterministically and fails cleanly. Keep the three harness tests as integration coverage on top.

### 2. Two coverage gaps

- **Sibling preservation is never asserted.** `baseWithMutableReplaySpine` shallow-clones existing plain containers precisely so sibling keys survive the rebuild — but all three tests seed a base with no sibling at the rebuilt level (two seed `null`, one seeds a bare `FabricEpochNsec`). The subtlest part of the fix is untested. Please add a case like confirmed base `{value: 5, other: 7}`, child writes `/value/x`, asserting `other: 7` survives.
- **Only leaf/root stale non-containers are covered.** `baseWithMutableReplaySpine` handles intermediate spine nodes (`index < path.length - 1`), but no test lands a stale non-container mid-spine. A path whose intermediate segment hits a primitive would exercise the `emptyReplayContainerForNextKey` + descent branch directly.

Lower priority: the `-` append-marker / array-with-non-index-key branches aren't exercised, and there's no negative test confirming the wrapper still propagates unrelated `CloneForMutationError` kinds (e.g. `non-container-root`) rather than over-broadening.

### Separately worth noting (not blocking)

The headline browser regression guards (`lunch-poll-contention.test.ts`, the tightened `lunch-poll-vote.test.ts`) live in the `packages/patterns` integration suite, which isn't invoked by any CI job — CI only runs `patterns-reload` and `pattern-tests`. So those tests won't prevent a future regression in CI; the durable guard is the runner unit test (hence the asks above). Either wire the `patterns` suite into CI or call out explicitly that the runner test is the real guard.

_Reviewed by Claude (Opus 4.8) at the author's request._

willkelly:
Follow-up pushed for the review comments and CI split.

What changed:
- Added direct synchronous runner coverage for `cloneWithPendingReplayValueAtPath`:
  - stale non-container base rebuilds
  - sibling preservation during stale-spine rebuilds
  - unrelated `CloneForMutationError` propagation remains strict
- Split the long Lunch Poll contention browser test out of the normal 4 pattern-integration shards into `Pattern Integration Test (Lunch Poll contention)`.
- Excluded `lunch-poll-contention.test.ts` from normal shard selection with selector tests proving it is not duplicated/dropped.
- Hardened the contention test readiness gate to wait for option-card `data-option-id` hooks on every voter page, and extended that specific readiness wait to 180s for CI load.
- Wired the new standalone job into coverage/perf consumers: `perf-check` needs, `attest-binaries` needs, expected coverage artifacts, perf metric extraction, and coverage docs.

Verification run locally:
- LSP clean on touched TS files.
- `deno fmt --check ...` passed.
- `deno check tasks/perf-check.ts tasks/perf-lib.ts tasks/perf-lib.test.ts packages/patterns/integration/lunch-poll-contention.test.ts` passed.
- `deno test -A tasks/perf-lib.test.ts tasks/select-pattern-integration-files.test.ts` passed: 53 tests.
- Selector outputs for 1/4..4/4 omit `lunch-poll-contention.test.ts`.
- Single-file Lunch Poll contention browser test passed against local offset-93 stack in 26s.
- `deno test -A packages/runner/test/memory-v2-stacked-commit.test.ts` passed earlier with the new runner helper tests: 53/53.

Latest push is `82d256015`; CI is running now.

willkelly:
Final workflow follow-up: the post-implementation context review found one more consumer, `deploy-shell-staging`, that waited on the pattern integration jobs but did not yet wait on the new standalone Lunch Poll contention job. Pushed `b98d7a127` to add `pattern-lunch-poll-contention-test` there too.

Additional verification:
- `deno fmt --check .github/workflows/deno.yml` passed.
- Workflow grep now shows the standalone job included with the pattern integration/reload jobs in all three relevant dependency lists: `perf-check`, `attest-binaries`, and `deploy-shell-staging`.

willkelly (commented):
## Follow-up: requested hardening landed — and a correction to my earlier CI claim

Re-reviewed at `b98d7a1273`. The runner test asks are addressed, and I need to retract one point from my previous review.

### Test hardening landed (verified passing locally)

`cloneWithPendingReplayValueAtPath` is now exported (`@internal Exported for testing only`) with three synchronous unit tests, which close exactly the gaps I raised:

- **Synchronous helper test** — deterministic, no subscription-timing race. A regression now fails as a clean, attributable assertion instead of a dangling-promise module crash that cancels siblings.
- **Sibling preservation + mid-spine** — `{value: 5, other: 7}` → `{value: {x:1}, other: 7}`, and `{value: {mid: 5, keep: 7}}` → `{value: {mid: {x:1}, keep: 7}}`. Covers both sibling survival and the intermediate stale-non-container path.
- **Negative propagation** — `cloneWithPendingReplayValueAtPath(5, ["x"], 1)` asserts a non-repaired kind escapes, locking the narrow contract.

Ran the full `memory-v2-stacked-commit.test.ts`: 53/53 pass.

### Correction: the patterns suite was already in CI

My earlier point — "the `packages/patterns` integration suite isn't invoked by any CI job, CI only runs `patterns-reload` and `pattern-tests`" — was **wrong**. There's a pre-existing sharded `Pattern Integration Tests` job that enumerates `packages/patterns/integration/*.test.ts` via `select-pattern-integration-files.ts`. I grepped for `deno task integration` and missed that this job uses a direct `deno test` invocation. So `lunch-poll-vote.test.ts` was already running in CI, and `lunch-poll-contention.test.ts` would have been auto-assigned to a shard. Apologies for the bad call.

Given that, the CI commits here read correctly: rather than wiring in an un-run suite, they **isolate** the heavy 5×10 contention test into a dedicated job (own toolshed server + readiness wait), **exclude** it from the round-robin shards so it can't bottleneck them (verified excluded from all four shards — no double-run), and **gate staging deploy** on it. The readiness hardening (per-page option-card-id wait, longer timeout, cause-chaining) and the new selector/perf-lib unit tests (53/53 pass) round it out. That's the right shape for a high-contention test.

No outstanding asks from me. LGTM.

_Reviewed by Claude (Opus 4.8) at the author's request._

mathpirate:
**Follow-up review notes — storage replay fix**

Took a closer pass on the storage change. To restate the headline first: the fix is correct and nicely scoped — local optimistic-visibility path only, two error kinds, server stays authoritative — and the unit + harness coverage that landed alongside it is solid. Nothing below is a blocker; one clarity nit, one thing that looks already-handled, and one small coverage gap.

**1. `baseWithMutableReplaySpine` — the non-container-root branch looks unreachable (clarity/nit)**

In the root initializer:
```js
const root = isPendingReplayContainer(base)
  ? cloneIfNecessary(base, { frozen: false, deep: false })
  : emptyReplayContainerForNextKey(path[0] ?? "");   // <- can this ever run?
```
The only caller is the `catch` in `cloneWithPendingReplayValueAtPath`, which proceeds only for `non-container-descent` / `non-mutable-leaf`. Both are thrown by `cloneForMutation` *after* it has passed its own root `isPlainContainer` check, and the `null`/`undefined` base case is filled by `createMissing` and never throws. So by the time we get here, `base` is always a plain container — `isPendingReplayContainer(base)` is always true, and the `else` branch (plus the `path[0] ?? ""` guard) can't execute. The new *"unrelated clone errors remain strict"* test reinforces this: it deliberately keeps a root-level non-container strict (`non-mutable-root`).

So the branch advertises a capability the wrapper intentionally refuses. Suggest either dropping it and asserting the invariant (`base` is a plain container here) with a one-line comment, or — if root-level rebuild is actually wanted — widening the catch. The negative test implies the former.

**2. CI coverage of the browser guards — looks already satisfied (flagging in case the earlier note predated current infra)**

The review note mentioned the `packages/patterns` integration tests aren't invoked by CI. From the workflow as it stands now, that doesn't seem to hold: `pattern-integration-test` runs `select-pattern-integration-files.ts`, whose `listPatternIntegrationTests()` globs `packages/patterns/integration/*.test.ts` and shards them. So `lunch-poll-vote.test.ts` rides those shards today, and `lunch-poll-contention.test.ts` is excluded from them via `INDEPENDENT_PATTERN_INTEGRATION_FILES` precisely because it now has its own job. Both guards therefore appear CI-covered (and deploy/perf/attestation all `needs:` `pattern-integration-test` plus the new contention job, so these are real gates). Worth a quick confirm there's no path-filter quirk, but if so, the "wire patterns into CI" ask is already done.

**3. Two spine-builder branches remain unexercised (low)**

- The array-with-non-index-key path in `setChildAtReplayKey` looks *effectively unreachable* — `cloneWithValueAtPath`'s `createMissing` resolves that shape before it can throw — so it reads as defensive and is fine to leave.
- The `-` append-marker branch (a patch appending to an array over a stale base) is reachable in principle and currently uncovered; a one-line unit test would close it.

bfollington:
Cross-ref from #4457 (a convergence investigation package): this PR and #4457's **B3** finding are in the same `v2.ts` pending-replay machinery and the same symptom family — your *"50 stored votes / 42 rendered swatches"* is the same class as B3's "durable has every write, but an interleaving writer's own replica is missing a peer's concurrent append."

I checked whether this PR already covers it: I ran #4457's deterministic B3 test (`array-append-client-convergence.test.ts` — seed `[a]`, one writer appends `A`, another appends `B` without observing `A`) against this branch, and it **fails** — the interleaving writer reads `["B","seed"]`, missing `"A"`, even after explicit sync+pull (a fresh reader converges fine). So it looks like an **adjacent** failure mode this PR doesn't yet address:
- **This PR** repairs pending-replay *materialization* over stale/shape-mismatched bases (null roots, array spines, `-` append marker; the `CloneForMutationError`).
- **B3** is the peer's revision never getting *integrated*: `confirmPending` rebuilds from a stale base **and** `applySessionSync`'s `upsert.seq < confirmed.seq` guard shadows the peer's lower-seq append once the writer's own commit advances its cursor.

Since you're deepest in this code, flagging in case it's cheap to fold in (the 60ms test is a ready red-to-green target) or worth a follow-up. Full mechanism + 3 costed fix options are in #4457.

willkelly:
Spoke with Robin about this today -- Robin believes the fundamental problem may be elsewhere and is investigating.

From Robin:

> this approach seems reasonable, but it's pointing to a conflict that i don't think should happen (but almost certainly is). I think when we added the delay before notifying clients of the new state, we may have violated one of our conflict rules. This change was motivated by the common case where we get a stream of commits from one client, and we send them all out, and each of them triggers a bunch of new commits from each of the individual clients -- by waiting some period of time, the other clients just see the last values from that commit stream, and they don't go off and generate a bunch of commits that are stale by the time they arrive.
Anyhow, I suspect this can result in us getting a commit rejection when we don't have the latest value that triggered the rejection. That means when we go to reapply our other stacked changes on top of our state after pulling that rejected item out of the stack, what we think the server has isn't the right shape for us to re-apply those pending changes to our optimistic view.

ubik2:
My own investigation into places where our replica's view can get out of sync broke these issues into 5 classes
https://linear.app/common-tools/issue/CT-1872/memory-v2-stale-base-replay-escape-hatches-inventory-and-exhaustive

For my own reference, Ben's red test is now at https://github.com/commontoolsinc/labs/pull/4533

willkelly:
Closing — crash fixed elsewhere, and the author's own diagnosis was superseded.

Last substantive comment here was mine on 2026-07-08: "Spoke with Robin about this today -- Robin believes the fundamental problem may be elsewhere and is investigating." Same day, CT-1872 reframed this as an inventory of five distinct stale-base replay escape hatches rather than a point fix. Untouched since.

The headline repro (`CloneForMutationError: cannot descend into null at path index 0`) was fixed by #4406, merged the day after this opened. Main has `firstExistingPrefixThatBlocksPendingPath` / `pendingSetPathForBase` at `storage/v2.ts:448-479` with three regression tests. The symptom does not reproduce.

**Salvage — carry into #5280:**

~120 lines of `v2.ts` that main genuinely lacks. #4406 fixes the crash by replacing the whole blocked ancestor with the surviving patch's snapshot branch — which **resurrects data belonging to the dropped parent commit**. This branch's `baseWithMutableReplaySpine` / `cloneWithPendingReplayValueAtPath` rebuild only the mutable spine and write only the surviving leaf, so a dropped commit's data stays dropped. That residual-semantics difference is the only unique engineering here and is not on main.

Also not on main: five DOM hooks (`data-all-option-id`, `data-all-option-title`, `data-vote-swatch-option-id`, `data-vote-swatch-type`, `data-option-id`; main has only `data-vote-swatch-name`) and a 785-line browser contention test.

The perf-lib/perf-check/deno.yml wiring is dead.

PR #4406, opened 2026-06-29
Fix pending patch visibility across branch type changes

Hixie:
When a pending memory v2 patch writes a nested field, the confirmed document can still have null, a primitive, or an array at an ancestor path. The old pending visibility code replayed the nested leaf path directly over that confirmed value. That asked the clone helper to descend through a value that the path reader cannot traverse. In the null case it threw a CloneForMutationError. In the array case it could create a named array property that later reads could not see.

The code now checks each present ancestor before replaying a pending leaf write. If an ancestor cannot be traversed for the next path segment, and the pending value contains that ancestor, the code replaces that ancestor with the pending branch. Normal nested writes still use the exact changed path.

Add runner-level stacked commit tests for null, scalar, and array ancestors. These tests make the browser lunch poll failure reproducible in a deterministic storage test.

---
## Summary by cubic
Fix pending patch visibility across branch type changes and log when a blocked branch is replaced. Prevents crashes and bad reads when replaying nested writes in memory v2, including during non-head confirmations.

- **Bug Fixes**
  - Detect the first existing ancestor that blocks traversal (arrays included) and apply the pending branch at that prefix; otherwise use the exact nested path.
  - Warn on branch replacement via `storage.v2.pending-patch` (`pending-branch-replace`) with space/id/scope/localSeq/path/replacementPath details.
  - Add runner tests for null, scalar, and array ancestors, plus confirming a later same-doc patch while an earlier pending overlay remains visible.

<sup>Written for commit 9addc59506406644c529d69891f2b10e6a5d85c4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4406?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/storage/v2.ts
+++ b/packages/runner/src/storage/v2.ts
@@ -134,8 +134,8 @@ const logger = getLogger("storage.v2", {
   level: "error",
 });
 const pendingPatchLogger = getLogger("storage.v2.pending-patch", {
-  enabled: false,
-  level: "info",
+  enabled: true,
+  level: "warn",
   logCountEvery: 0,
 });
 
@@ -449,7 +449,7 @@ const applyPendingVersion = (
         if (hasValueAtPath(pending.value, path)) {
           const setPath = pendingSetPathForBase(next, pending.value, path);
           if (!isSamePath(setPath, path)) {
-            pendingPatchLogger.info("pending-branch-replace", () => [
+            pendingPatchLogger.warn("pending-branch-replace", () => [
               "pending patch visibility replaced data at an existing branch that blocked the nested write",
               {
                 space: logContext.space,
--- a/packages/runner/src/storage/v2.ts
+++ b/packages/runner/src/storage/v2.ts
@@ -133,6 +133,11 @@ const logger = getLogger("storage.v2", {
   enabled: true,
   level: "error",
 });
+const pendingPatchLogger = getLogger("storage.v2.pending-patch", {
+  enabled: false,
+  level: "info",
+  logCountEvery: 0,
+});
 
 function withCommitTiming<T>(
   keys: string[],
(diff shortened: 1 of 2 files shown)


packages/runner/src/storage/v2.ts
```diff
@@ -360,6 +368,39 @@ const changedPathsForPendingPatch = (
...
   });
 
+const firstPresentNonTraversablePrefix = (
+  base: EntityDocument | undefined,
+  path: readonly string[],
+): string[] | undefined => {
+  for (let length = 1; length < path.length; length += 1) {
+    const prefix = path.slice(0, length);
+    if (!hasValueAtPath(base, prefix)) {
+      continue;
```
ubik2: Should we do a `return prefix` here instead?
Actually, I think there's just something wrong with the logic here. It's not entirely clear what we expect to happen, but I don't think this returns what I expect for a function with this name.

Hixie: We should do `return undefined`. (`continue` is equivalent, but less obvious, I think)

What the function is supposed to do is “Walk down the path. Ignore missing branches. If you find a branch that exists but cannot be walked through, return that branch.”

I'll have my agent come up with a better name and comments.

ubik2: thanks for clarifying. makes sense.

realized this was what it was probably supposed to do a bit later, but the name confusion without a doc comment, and the option to return early seems better (just a preference, to skip work).

PR #4408, opened 2026-06-29
test(runner): split slow runner test files

Hixie:
The runner test job assigns files to shards by sorted round-robin. A few large test files kept much of the work in one shard, which made CI runtime uneven even when the file counts were balanced.

Split the largest runner test files into smaller files that preserve the existing coverage while giving the shard selector more units to distribute. The engine tests now follow their existing top-level suites and share common setup through engine-test-support.ts. The fetchData mutex coverage is divided into core mutex behavior, reactive fetch state, and protected-request auth files. The profile space collision coverage now has one assertion per test file, with shared profile creation code in a support module.

The profile assertions still check the same behavior, but each file now creates only the spaces that assertion needs instead of running all profile setup before every assertion.

Resetting coverage metric due to unattributable coverage change (I have another agent tracking it down):
NEW_COVERAGE_BASELINE

---
## Summary by cubic
Split oversized runner test files so the shard selector has more files to distribute, making CI runtimes more even. Coverage is unchanged; this is a test-only refactor.

- **Refactors**
  - Replaced monolithic `engine.test.ts` with focused suites: `engine-compile-evaluate.test.ts`, `engine-compile-record-graph.test.ts`, `engine-evaluate-record-graph.test.ts`, and `engine-ses.test.ts`, with shared setup in `engine-test-support.ts`.
  - Split `fetch-data-mutex.test.ts` into `fetch-data-mutex-core.test.ts`, `fetch-data-mutex-reactivity.test.ts`, and `fetch-data-mutex-auth.test.ts`.
  - Broke profile space collision tests into single-assertion files: `profile-space-identity-collision-users.test.ts`, `...-same-user.test.ts`, and `...-did.test.ts`, with helpers in `profile-space-identity-collision-support.ts`.
  - Tests now create only the data each assertion needs, reducing per-test overhead and improving shard balance.

<sup>Written for commit 2ad4d361c5659b74095e50d1e0cb26bd82b05452. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4408?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mathpirate (approved):
lgtm ty!

PR #4414, opened 2026-06-30
Proposal: remote space dumps for cf inspect (--remote)

bfollington:
> **Draft / proposal.** This works end-to-end (tests + local e2e green), but I'm
> opening it as a draft so we can agree on **how we want to expose space dumps**
> before turning anything on in a deployed environment. The security model is the
> part I most want eyes on — see "Security model" and "Open questions".

## Problem

`cf inspect` (the state-inspector) is an offline autopsy tool: it reads a space's
durable memory-v2 SQLite file read-only and answers who/what/when + cross-space
convergence questions. Today it only reads **local** on-disk DBs, so to inspect
**staging** (e.g. `rapids.saga-castor.ts.net`) you have to `ssh` into the box and
`scp` the `.sqlite` out. That's friction, and it spreads SSH/box access wider
than we'd like just so people can debug state.

## Proposal

Let the server hand out a read-only snapshot over HTTP, and let `cf inspect`
fetch it. The autopsy itself stays 100% offline — `--remote` only changes *where
the SQLite file comes from*:

```bash
export CF_IDENTITY=./me.key
cf inspect spaces  --remote https://rapids.saga-castor.ts.net      # list dumpable spaces
cf inspect summary <did> --remote https://rapids…                  # fetch + inspect one
cf inspect pull --all    --remote https://rapids…                  # cache them all, then…
cf inspect group                                                   #   …inspect offline
```

Snapshots are cached under `~/.cache/cf-inspect/<host>/` and auto-discovered by
the normal (local) commands, so you fetch once and dissect offline.

## Security model (the bit to discuss)

A dump is the **entire contents of a space**, so access is intentionally narrow.
Four independent layers, **off by default**:

1. **Perimeter — Tailscale.** The endpoint is only ever exposed on hosts that are
   tailnet-only (staging already is; we already trust the tailnet for the LLM
   gateway). That alone gives per-person, SSO-backed, audited reachability with
   no new infra.
2. **Opt-in per environment.** The route doesn't mount unless
   `MEMORY_DUMP_ENABLED=true`. When off it `404`s as if it never existed. It
   **hard-refuses to mount under `ENV=production`, with no override** — there is
   no code path that enables raw whole-space dumps in prod (a productionized
   form would be a separate, access-controlled mechanism). Note the real
   boundary is the perimeter + opt-in: `ENV` is a free-form string, so the prod
   refusal is belt-and-suspenders, not load-bearing.
3. **App-level gate — CF1 signature + DID allowlist.** Every request must carry a
   valid CF1 first-party Ed25519 signature (reuses
   `@commonfabric/runner/toolshed-http-auth`) from a DID on the allowlist
   (`MEMORY_DUMP_DIDS ∪ MEMORY_SERVICE_DIDS`). No shared secret to rotate; every
   dump is attributable to a person. This also fills the existing
   authorization `TODO` in `middlewares/first-party-http-auth.ts`.
4. **Audit.** Every served dump is logged with who / which space / size.

I chose DID-allowlist over a shared bearer token so there's no secret to leak and
dumps are per-person attributable; over Tailscale-only so a tailnet guest can't
vacuum every space. Tailscale + opt-in + allowlist + audit is the
lowest-friction option that's still defensible if a space ever holds real data.

## How it works

**Server** (`packages/toolshed/routes/storage/memory/memory-dump.index.ts`):
- `GET /api/storage/memory/dump` → list spaces (DID + size + mtime)
- `GET /api/storage/memory/dump/:space` → stream a snapshot (`application/octet-stream`)
- Snapshot is produced with **`VACUUM INTO`** on a fresh read-only connection
  (`packages/memory/v2/dump.ts`). This is WAL-safe (one consistent file, no
  `-wal`/`-shm`), never mutates the source, and deliberately does **not** go
  through the engine (whose constructor runs migrations that would write to the
  store). Space ids are path-traversal-guarded.

**Client** (`packages/state-inspector/remote.ts` + `cf inspect` plumbing):
- `--remote [url]` (defaults to `CF_API_URL`), `--identity` (defaults to
  `CF_IDENTITY`), and a `pull` command.
- Dep-light: state-inspector keeps no runner/identity dependency; the CLI injects
  the signer.

## New env vars (server)

| var | default | meaning |
| --- | --- | --- |
| `MEMORY_DUMP_ENABLED` | off | master switch; route 404s when off |
| `MEMORY_DUMP_DIDS` | "" | extra DIDs allowed to dump (∪ `MEMORY_SERVICE_DIDS`) |

To enable on staging: `MEMORY_DUMP_ENABLED=true` + your DID in
`MEMORY_DUMP_DIDS`.

## Testing

- **Unit** — `memory/v2/dump.test.ts` (list/resolve/traversal-guard/VACUUM
  round-trip), `state-inspector/test/remote.test.ts` (fetch caching/force/error
  mapping, stubbed fetch).
- **Integration** — `memory-dump.test.ts`: 401 unsigned, 403 non-allowlisted,
  list, dump round-trip (valid SQLite, data intact), 404 unknown, 404 traversal.
- **Live e2e** — ran a dump-only server against the real local cache and drove
  the actual `cf inspect spaces / summary / pull --remote http://localhost:…`,
  including offline inspection of pulled DBs and 401/403 paths.
- `deno fmt` / `lint` / `check` green across touched packages.

## Open questions for review

- **Is HTTP dump the right shape**, or would we rather extend the existing WS
  memory protocol (it already has an ACL-gated `sqlite.query` op) for live
  read-only queries instead of shipping whole files?
- **Allowlist source** — reuse `MEMORY_SERVICE_DIDS`, a dedicated
  `MEMORY_DUMP_DIDS`, or wire to a real role/ACL once we have one?
- ~~**Should prod ever be allowed at all** (even gated), or stay staging-only?~~ **Resolved: staging-only.** Per review, `MEMORY_DUMP_ALLOW_IN_PRODUCTION` was removed; `ENV=production` is an unconditional refusal.
- **Size/PII** — any space-size cap, redaction, or retention policy on cached
  dumps we want before this is broadly usable?
- Naming: `--remote` defaulting to `CF_API_URL` — good, or too implicit?

## Not included / follow-ups

- Space **name** → DID resolution over `--remote` (today: full DID or prefix).
- Cache-vs-local DID collision UX (pulling a DID you also have locally is
  correctly flagged "ambiguous"; could namespace better).

---
## Summary by cubic
Adds remote space dumps to state-inspector: a toolshed endpoint serves read-only SQLite snapshots, and `cf inspect` gains `--remote` to fetch, cache, and inspect them fully offline. Staging-only; production is a hard no, and the endpoint only mounts when `MEMORY_DUMP_ENABLED` is set on a recognized non-prod env (development/test/staging) with CF1-signed requests and a DID allowlist.

- **New Features**
  - Server: `GET /api/storage/memory/dump` lists spaces; `GET /api/storage/memory/dump/:space` streams a snapshot. Uses `VACUUM INTO` on a read-only `@db/sqlite` connection; paths resolved via the same helpers as the memory server; traversal-guarded; CF1 signature + DID allowlist; audit logging; off by default and hard-off in production.
  - CLI: `--remote [url]` (defaults to `CF_API_URL`) and `--identity` (defaults to `CF_IDENTITY`); new `inspect pull` to cache snapshots; `inspect spaces --remote` lists dumpable spaces with size/mtime. Snapshots are stored under `~/.cache/cf-inspect/<host>/` using literal DID filenames and auto-discovered for offline commands.

- **Bug Fixes**
  - Access policy: fail-closed env allowlist — only mounts on development/test/staging; unknown/alias envs refuse; when disabled, both list and per-space routes return 404 (keeps the endpoint invisible); unit-tested.
  - Storage/listing: mode-aware filename handling preserves directory-mode stems (fixes `%`-containing DIDs like `did:web:…%3A8080`); single-file encoding restored; server cleans snapshot temp dirs and closes FDs on failures after `VACUUM`; non-NotFound IO errors surface.
  - CLI resolution: remote tokens now resolve like local ones — full DIDs and `did:`-prefixed prefixes map via the remote list; clearer 401/403/404 errors; renamed tests to match package globs so CI runs them.

<sup>Written for commit 4440a898642a42853962fb1c8b65527cf0e7baa4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4414?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bfollington (commented):
cf-review via GPT-5 Codex, on behalf of @bfollington

Scope: PR #4414 remote space dumps for `cf inspect --remote`; coverage: deep-read server route, dump helpers, auth/signing boundary, CLI remote plumbing, cache discovery, docs/tests; checks: targeted Deno tests for the new dump/remote files passed, touched-file `deno fmt --check` passed, touched-file `deno lint` passed. I also ran focused path/cache proofs for the storage-layout and offline-cache behavior below.

**Blocking**

1. `packages/toolshed/routes/storage/memory/memory-dump.index.ts:90` / `:97` - The dump route lists and resolves files directly under `memoryEngineStoreUrl`, but the live `MemoryServer` does not write space DBs there. `memory.ts` passes `memoryEngineStoreUrl` into `MemoryServer.Server`, and `Server.openEngine()` then calls `resolveSpaceStoreUrl(this.#store, space)`, which appends another storage layer. Verified with the repo helpers: directory mode gives `memoryEngineStoreUrl = .../engine-v3/`, while the server's actual space URL is `.../engine-v3/engine-v3/<did>.sqlite`; a file at that actual server path makes `listSpaceStores(memoryEngineStoreUrl)` return `[]` and `spaceStorePath(memoryEngineStoreUrl, did)` return `null`. So on a real Toolshed cache this endpoint will list zero spaces and return 404 for existing spaces. The current integration test misses this because it hand-seeds the shallower directory instead of writing through the same path helper/server path the runtime uses. Fix by making the dump route use the exact same canonical store path as `MemoryServer` for both list and single-space dump, and add a regression test that seeds via that canonical path (or through the server) rather than the simplified shallow layout.

2. `packages/state-inspector/remote.ts:119` + `packages/state-inspector/discover.ts:92` - Pulled snapshots are cached as `${encodeURIComponent(space)}.sqlite`, but local discovery treats the filename stem as the canonical DID without decoding it. Verified with a cache file named `did%3Akey%3AzCacheEncoding.sqlite`: discovery reports `did%3Akey%3AzCacheEncoding`, and `resolveSpacePath("did:key:zCacheEncoding", found)` throws `no space matches`. That breaks the advertised workflow where `inspect pull` prints real DIDs and normal offline commands can then inspect the pulled DBs by DID. Fix by normalizing discovered remote-cache filenames back to canonical DIDs (while preserving local literal-DID filenames), or by writing cache filenames in the same canonical form discovery expects. Add a test that goes from `fetchSpaceDb(...)` to `discoverSpaceDbs(...)` / `resolveSpacePath(fullDid, found)`.

**Improvement**

- `packages/toolshed/routes/storage/memory/memory-dump.test.ts` should also pin the security gates that make this proposal acceptable: disabled endpoint returns 404, `ENV=production` + `MEMORY_DUMP_ENABLED=true` still returns 404 unless `MEMORY_DUMP_ALLOW_IN_PRODUCTION=true`, and the allowlist can come from `MEMORY_SERVICE_DIDS`. The auth happy/401/403 cases are covered, but the opt-in and prod defense-in-depth behavior is currently only covered by comments.

Verdict: block before merge. The overall shape is reasonable and the CF1 + DID allowlist boundary looks coherent, but the endpoint currently misses the live store layout and the pulled-cache offline path fails for full DIDs.

bfollington:
Thanks — both blockers were real; verified against the runtime and fixed in `5d849a5b4`.

**Blocking 1 (store-path layout).** Confirmed: `MemoryServer.openEngine` resolves `resolveSpaceStoreUrl(this.#store, space)` and `#store` is already `memoryEngineStoreUrl`, so real files live one `engine-v3/` deeper than the route was looking. My e2e masked it (I'd pointed `MEMORY_DIR` at the inner `engine-v3`) and the test seeded the shallow path. Fix:
- Extracted `resolveSpaceStoreDirUrl` in `storage-path.ts` as the single source of truth; `resolveSpaceStoreUrl` now builds on it (behaviour unchanged — verified identical output for directory + single-file modes).
- `dump.ts` `listSpaceStores`/`spaceStorePath` now resolve through those same helpers.
- Both the unit test and the integration test now **seed via the canonical `resolveSpaceStoreUrl` path** (the integration test seeds through the actual `memoryEngineStoreUrl` the server uses), so they exercise the doubled layout and would have caught this.

**Blocking 2 (pulled-cache offline path).** Confirmed: caching as `encodeURIComponent(did).sqlite` left full-DID resolution broken (only a `z6Mk…` prefix matched). Fix: cache with the **literal** DID filename, matching the server/local on-disk convention. Added a `fetchSpaceDb → discoverSpaceDbs → resolveSpacePath(fullDid)` round-trip test.

**Improvement (gate tests).** Extracted the access policy into a pure `memory-dump-policy.ts` and unit-tested the matrix directly: off-by-default, production hard-off unless `MEMORY_DUMP_ALLOW_IN_PRODUCTION`, and the `MEMORY_DUMP_DIDS ∪ MEMORY_SERVICE_DIDS` allowlist union.

Re-verified end-to-end against a production-realistic `MEMORY_DIR` (the store root, not the inner `engine-v3`): list, dump-by-full-DID over `--remote`, and offline full-DID resolution of a pulled DB all work. Still a draft — happy to keep iterating on the open questions (HTTP dump vs WS `sqlite.query`, allowlist source, prod policy).

ubik2:
> Is HTTP dump the right shape, or would we rather extend the existing WS
memory protocol (it already has an ACL-gated sqlite.query op) for live
read-only queries instead of shipping whole files?

HTTP dump is the way I imagine using it. I want a snapshot I can run local analysis on.

> Allowlist source — reuse MEMORY_SERVICE_DIDS, a dedicated
MEMORY_DUMP_DIDS, or wire to a real role/ACL once we have one?

Either reuse or dedicated seems fine.

> Should prod ever be allowed at all (even gated), or stay staging-only?

I think this should be staging-only. If we really need prod in the window before we truly lock it down, we can use ssh.

> Size/PII — any space-size cap, redaction, or retention policy on cached
dumps we want before this is broadly usable?

Since I don't think this belongs in production, I don't think it's worth the hassle to redact (plus, that makes it hard to use for debugging). Given that we're providing a trusted DID over tailscale, I think it's fine to skip space-size cap and retention policies.

> Naming: --remote defaulting to CF_API_URL — good, or too implicit?

Seems fine.

willkelly (approved):
Approving the concept — this is a clean, well-isolated "fetch a read-only SQLite snapshot instead of SSH+scp" feature. The autopsy stays offline; `--remote` only changes where the file comes from. The four-layer model (tailnet perimeter, opt-in, CF1 signature + DID allowlist, audit log) is sensible, `VACUUM INTO` on a read-only connection is the right way to get a crash-consistent single file without touching the live store or running migrations, and path traversal is properly guarded at the `encodeStoreSubject` layer (not just string-checked). Tests cover the security matrix (401/403/404/traversal) plus a real round-trip. Good draft.

Approving **with one follow-up request** (below), and a few non-blocking notes.

## Follow-up request (before this is enabled anywhere beyond staging): make prod a hard, code-level no

My position: we should **never** turn raw whole-space dumps on in prod — direct DB access in this form isn't something we want there. A productionized form of this is a *later, different* mechanism with real access control, exposed only to devices that provably cannot exfiltrate data — not this endpoint behind a second flag.

So the ask: **drop `MEMORY_DUMP_ALLOW_IN_PRODUCTION` entirely** and make `ENV=production` an unconditional refusal to mount. That turns "off by default in prod" into "no code path enables this in prod," which is both safer and simpler. Not a merge blocker for a staging-only draft, but please don't let the escape hatch quietly become the prod story.

This also removes a fragility: the production guard is `cfg.env === "production"`, and `ENV` is `z.string().default("development")` — free-form. If a deployment sets `ENV=prod` or leaves it unset, the refusal silently doesn't apply and only perimeter + opt-in protect you. Better to make perimeter + opt-in the real boundary and treat prod-refusal as belt-and-suspenders rather than load-bearing.

## Security notes (none blocking for a staging-only draft)

- **Replay window.** CF1 proofs bind method + host + path + time but there's no nonce/replay cache, so a captured proof is replayable for its validity window (default 60s, max 300s). On a TLS-only tailnet that's acceptable and consistent with the existing first-party auth, but it means security genuinely rests on TLS + tailnet, not the signature alone.
- **`flagValue()` footgun.** Only the literal string `"false"` disables a flag; `MEMORY_DUMP_ENABLED=0` / `=no` evaluates to **true**. Existing repo convention, not this PR's bug, but worth being aware of given what this flag exposes.

## Correctness / robustness

- **Read-only `VACUUM INTO` against a live WAL.** Works while the server holds the DB open (shm exists). But for a space the server *isn't* actively holding open whose `-wal` still has un-checkpointed frames, a read-only connection can fail because WAL recovery needs write access to `-shm`/`-wal`. The route catches this → 500, so it degrades safely, but some idle-space dumps could intermittently fail. Is that a case you hit in the e2e run?
- **No size or concurrency cap.** Every request does a full `VACUUM INTO` copy into a temp dir; a multi-GB space or several concurrent dumps means real disk + CPU pressure with nothing throttling it. You already list a size cap as an open question — I'd add a concurrency guard to that list before this is broadly usable (fine for staging now).

## Minor

- The `remoteSigner` comment says unsigned requests are "e.g. for a Tailscale-perimeter-only server" — but the server has no unauthenticated mode; an unsigned request always 401s. Misleading comment.
- CI coverage gate is currently failing (+181 uncovered lines, 153 in `inspect.ts`). Draft-acceptable, but the bulk of the new CLI plumbing is untested.

## On your open questions

- *Should prod ever be allowed?* → No. Staging-only; a prod form is a separate future mechanism with access control + exfil-proof clients. Please remove the prod flag.
- *HTTP dump vs WS `sqlite.query`?* → Direct sqlite dump is a reasonable shape for now.
- *Allowlist source?* → The `MEMORY_DUMP_DIDS` ∪ `MEMORY_SERVICE_DIDS` union is fine until there's a real role/ACL.

bfollington:
Thanks @willkelly and @ubik2 — pushed `0f5302ae0` acting on the feedback.

**Prod is now a hard, code-level no (per @willkelly).** Dropped `MEMORY_DUMP_ALLOW_IN_PRODUCTION` entirely; `ENV=production` is an unconditional refusal with no override. I also took your point 2 to heart in the code: a comment on the policy now states that the *real* boundary is the tailnet perimeter + opt-in-off-by-default, and the `ENV==="production"` check is belt-and-suspenders (not load-bearing, since `ENV` is a free-form string). Net simplification — one fewer flag and branch.

**Documented the edges you flagged (staying staging-only, no new machinery):**
- The read-only `VACUUM INTO` vs idle-space dirty-WAL case — noted in `snapshotSpaceStore` that it can throw for an idle space the server isn't holding open, and the route turns that into a 500 (fails safe). And to answer your question: **no, I didn't hit it in the e2e** — those DBs were cleanly checkpointed with no writer holding them, so there were no dirty WAL frames to recover.
- No size/concurrency cap — documented on the route as a deliberate staging-only choice, with a concurrency guard called out as a prerequisite before this is ever broader. (@ubik2 — agree redaction/retention aren't worth it for a trusted-DID-over-tailnet staging tool.)

**Fixed the misleading `remoteSigner` comment** — there's no unauthenticated server mode; a missing key just 401s, and we surface the actionable "set CF_IDENTITY" message.

**Coverage gap closed.** Added `packages/cli/test/inspect-remote.test.ts` that drives `inspect … --remote` in-process against a stubbed fetch (spaces/pull/summary/converge + the error paths), so the remote plumbing is actually exercised; extended the memory/state-inspector/toolshed tests to cover the NotFound-vs-rethrow branches, the full error-mapping table, the invalid-id guard, and the snapshot-failure 500. New files are now ~85–100% line-covered.

On the non-blocking replay-window / `flagValue()` notes: agreed both are inherent to the existing first-party-auth / env conventions rather than this PR, and on TLS+tailnet the replay window is acceptable — noting them here for the record.

Still a draft; happy to keep iterating.

bfollington:
Final sanity pass before un-drafting — three loose ends found and fixed in `b2772776c`:

- **Skill self-contradiction**: `skills/state-inspector/SKILL.md` still said "no remote mode — copy the file off the box" a few sections above the new `--remote` docs. Now reads: staging spaces via `--remote`; the dump endpoint is deliberately hard-off in production, so a prod space still means ssh+scp.
- **PR body drift**: the description still documented the removed `MEMORY_DUMP_ALLOW_IN_PRODUCTION` override in the security model and env-var table. Body updated to match the code (prod = unconditional refusal, no override; the open "should prod ever be allowed" question is marked resolved per review).
- **Coverage ratchet fully closed**: the Performance Check was still marginally over (memory +11, state-inspector +3). Added targeted tests for the remaining branches — junk-filename skips, the TOCTOU stat-race guard, null-mtime fallback, non-NotFound rethrows, unencodable space ids, non-JSON error bodies, `fetchSpaceDb`'s own HTTP-error mapping, and interrupted-download cleanup (asserts no `.partial`/torn cache file survives). `dump.ts`, `storage-path.ts`, and `remote.ts` are now at 100% line coverage.

Also re-verified: no stale references to the removed flag anywhere, `deno.lock` carries only the `@db/sqlite` entries, and all touched-package suites are green.

One known nit, documented rather than fixed: cliffy's optional-value `--remote [url]` will swallow a following bare token (`inspect summary --remote z6Mk…` parses the DID as the URL and fails with a confusing fetch error) — put the URL right after the flag, or rely on `CF_API_URL`.

From my side this is ready to come out of draft.

packages/cli/commands/inspect.ts
```diff
@@ -19,26 +19,32 @@ import {
   convergence,
   type ConvergenceResult,
   convergenceScan,
+  // Remote acquisition (`cf inspect --remote` / `pull`).
```
bfollington, line 22: Is this comment really needed?

bfollington:
Four more review findings, all verified real and fixed in `03a5159c8`:

- **Fail-closed env gate (P1).** The prod refusal matched only the literal `"production"`, so `ENV=prod` or any alias slipped through. Now an allowlist: the endpoint mounts only under a recognized non-production `ENV` (`development` | `test` | `staging`, case-insensitive) — anything else refuses even with `MEMORY_DUMP_ENABLED=true`. One hole no env check can close (documented in the policy): a prod box that leaves `ENV` unset inherits the `"development"` default — which is why the opt-in flag + tailnet perimeter remain the real boundary. **Ops note:** enabling on staging now also requires the host's `ENV` to be one of the recognized values.
- **%-bearing ids corrupted in directory-mode listing (P2).** Directory-mode filename stems are the *literal* id, but the listing `decodeURIComponent`'d them — a `did:web:…%3A8080` was listed corrupted and became unresolvable (verified empirically; no `did:key` affected). Added the mode-aware inverse `spaceFromStoreFilename` beside the encoder in `storage-path.ts`; `resolveSpaceStoreUrl` re-verified byte-identical to main in both modes.
- **Temp snapshot leak (P2).** A failure between `VACUUM INTO` and the unlink-after-open left the full snapshot in `/tmp`. The serve path now cleans the temp dir (and closes the fd) on any failure before the 500.
- **Remote DID prefixes (P2).** `did:`-prefixed tokens were sent verbatim, so a partial DID 404'd instead of prefix-resolving like local tokens. Full DIDs and did-prefixed prefixes both resolve via the remote listing now (exact match wins).

Each fix carries a regression test; full memory suite (298) green.

PR #4415, opened 2026-06-30
fix(wish): resolve #profile to the default directly instead of the picker

bfollington:
## Problem

A non-headless `wish("#profile")` from a viewer with **2+ profiles** launches the multi-candidate **profile-picker sidecar** (`wish.ts` — `if (isProfilePersonaTarget && !headless && uniqueResultCells.length > 1) sendResult(launchProfilePickerPattern(...))`), leaving `.result` **undefined until a selection is made**. That dead-locks every pattern that wishes for *"the viewer's active profile"*:

- `profile-group-chat`'s **Send button stays disabled forever** — its `hasProfile` guard requires the live `#profile` cell, which never resolves.
- The picker itself must hydrate its cross-space profile list, which is slow (and can stick) under latency.

Single-profile and headless callers were unaffected — which is exactly why this only bit users with **2+ profiles**, and why multi-user demos "work fine" when each test user has a single profile.

This was found while investigating two reports that turned out to be the **same root cause**: (A) the profile picker "renders as if profiles exist but the rows never populate," and (B) `profile-group-chat` flickering/never-settling on staging. Symptom A (picker non-hydration via this sidecar) is upstream of Symptom B.

## Fix

Resolve `#profile` to the **default profile directly** (single result → existing fast path) when a valid default is set. Only drive the picker on **genuine ambiguity** (no default chosen). The picker remains reachable as its own surface for *switching* profiles.

- `getProfileCandidateCells` now returns `{ ordered, defaultValid }`; the two existing callers are updated.
- The `#profile` case short-circuits to `ordered[0]` when `defaultValid`.

## Validation

- **Browser, real 2-profile identity:** send-guard stuck in **4/4 trials → enables in ~2s** after the change; posted a message, room count → 1, message rendered.
- **Unit tests:** `deno test wish.test.ts` → 90 steps, 0 failed; fmt + lint clean.
- **Negative control:** disabling the short-circuit makes *only* the new regression test fail — it genuinely guards the behavior.

## Out of scope (follow-ups)

- **Symptom A latency tail:** even when it resolves, the picker's cross-space profile pulls are a sequential round-trip chain (time-to-settle grows linearly with latency). Worth batching/parallelizing those pulls + deduping the cross-space `resolveLink` kick + latency-aware `pull()`/`settled()` bounds.
- **Design question:** should `wish(...).result` *always* be single (default/first) with the picker living only on `.candidates`/`[UI]`? That would remove this footgun class entirely rather than just the profile instance.

---
## Summary by cubic
Resolve `#profile` directly to the viewer’s default profile instead of launching the picker when a default exists. This fixes the deadlock where `.result` stayed undefined for users with 2+ profiles and unblocks flows like group chat send.

- **Bug Fixes**
  - Short-circuit `#profile` to the default when it’s valid; show the picker only when no default is set.
  - `getProfileCandidateCells` now returns `{ ordered, defaultValid }`; updated the two callers.
  - Added a regression test: 2 profiles + default → single direct result (no picker).

<sup>Written for commit d1efd76d6e2368cc79477e66c4f868d335bb774c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4415?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/builtins/wish.ts
+++ b/packages/runner/src/builtins/wish.ts
@@ -331,7 +331,9 @@ function subscribeProfileName(cell: Cell<unknown>): void {
  * order. Identity is by link equality (`Cell.equals`); there is no synthetic
  * key. Returns [] when no valid profile exists yet.
  */
-function getProfileCandidateCells(ctx: WishContext): Cell<unknown>[] {
+function getProfileCandidateCells(
+  ctx: WishContext,
+): { ordered: Cell<unknown>[]; defaultValid: boolean } {
   const homeSpaceCell = getHomeSpaceCell(ctx);
   const defaultPattern = homeSpaceCell.key("defaultPattern").resolveAsCell();
   const profilesCell = defaultPattern.key("profiles");
@@ -357,7 +359,7 @@ function getProfileCandidateCells(ctx: WishContext): Cell<unknown>[] {
     subscribeProfileName(cell);
     candidates.push(cell);
   }
-  if (candidates.length === 0) return [];
+  if (candidates.length === 0) return { ordered: [], defaultValid: false };
 
   // Ordering inputs: the default link and the MRU list.
   const defaultEntry = defaultPattern.key("defaultProfile");
@@ -390,7 +392,7 @@ function getProfileCandidateCells(ctx: WishContext): Cell<unknown>[] {
     }
     return mruRank(a) - mruRank(b);
   });
-  return ordered;
+  return { ordered, defaultValid };
 }
 
 /**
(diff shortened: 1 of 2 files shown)


bfollington (commented):
cf-review via OpenAI GPT-5, on behalf of @bfollington.

Scope: PR #4415 (`wish.ts` and `wish.test.ts`). Coverage: deep-read changed profile resolution path, profile-picker caller, and profile data docs/comments. Checks: `deno test -A packages/runner/test/wish.test.ts` passed (90 steps); `deno fmt --check packages/runner/src/builtins/wish.ts packages/runner/test/wish.test.ts` passed; `deno lint packages/runner/src/builtins/wish.ts packages/runner/test/wish.test.ts` passed.

Blocking

- `docs/common/conventions/wish.md:120`, `docs/specs/shared-profile-space.md:9`, and `packages/patterns/system/profile-picker.tsx:23` still describe the old rule that 2+ profiles launch the profile picker. This PR changes the runtime rule at `packages/runner/src/builtins/wish.ts:825`: if `defaultProfile` is valid, non-headless `wish("#profile")` now returns the default directly, and only "2+ profiles with no default" remains ambiguous enough to launch the picker. Leaving the current docs/comments in place will send the next person debugging this exact behavior back to the pre-fix model. Please update them to match the new 0/1/default/no-default cases, or explicitly choose a follow-up before merging. Verified by reading the changed path and the current docs.

Verdict: Runtime change looks sound and the regression test covers the reported deadlock; docs/comment coherence needs one small pass.

PR #4419, opened 2026-06-30
fix(cfc): resolve UI contracts from bound sigil links in event context

ubik2:
## Summary

`contractCandidatesFromEventContext` resolves UI contracts from contract-bearing sigil links in a handler's event `$ctx`. This PR corrects how those links are parsed, hardens the same-document guard, and — importantly — documents that the whole branch is currently **dormant** in production.

### Dormant-branch finding

Enforcement (`recordTrustedEventPolicyInputs`) is always called with the **bare sent event** (`$event`), never the `{$ctx, $event}` argument envelope. The handler's `$ctx` is split off and consumed separately by the action (`invokeJavaScriptImplementation`), so `payload.value.$ctx` is empty on every live path. This branch is exercised only by unit tests; a UI contract discoverable *only* via `$ctx` is **not enforced at runtime today**. It's retained as forward-looking scaffolding for a future path that would feed the argument envelope into enforcement, and is now annotated as such in the code.

### Parse hardening

At handler-execution time a bound `$ctx` entry is an absolute sigil link, not a symbolic `$alias`. The parse now:

- **Drops the `write` base and requires a full link** (`isNormalizedFullLink`). Previously `parseLink(value, write)` let a *relative* link inherit the write target's `id`/`space`/`scope`, which made the same-document guard vacuous — any relative link whose path prefixed the write path would inject a contract for the write's document. Parsing with no base keeps a relative link relative, so it fails the full-link check and is rejected instead of misattributed.
- **Treats any parsed link as a leaf** — never descends into a link's sigil envelope, even when the link is rejected.

Verified that bound `$ctx` links always carry a concrete scope (the `Cell` constructor defaults scope to `space`; it flows through `getMetaLink` / `scopedLinkForPath` / `createSigilLinkFromParsedLink`), so requiring a concrete scope never rejects a legit link.

## Tests

- Sigil-link fixtures now carry a concrete `scope`, matching real bound links.
- The event-context tests use the **plain in-memory event envelope** (`{ value: { $ctx, $event } }`) rather than a `data:`-URI sigil link, so they no longer imply the enforcement input is always a data-URI link. A **single** `data:application/json` example is kept (the "linked handler event envelopes" test) as the deliberate exceptional shape we verify is still decoded.
- Full `cfc-ui-contract.test.ts` suite green (717 passed).

## Commits

- `69d49249` — resolve UI contracts from bound sigil links in event context
- `65cc67ee` — mark event-context branch dormant; harden link parse; reshape tests

### DIFF
--- a/packages/runner/src/cfc/ui-contract.ts
+++ b/packages/runner/src/cfc/ui-contract.ts
@@ -9,6 +9,7 @@ import type { IExtendedStorageTransaction } from "../storage/interface.ts";
 import { getJSONFromDataURI } from "../uri-utils.ts";
 import { ContextualFlowControl } from "../cfc.ts";
 import type { CfcAddress } from "./types.ts";
+import { isNormalizedFullLink, type NormalizedLink } from "../link-types.ts";
 
 type UiContractTrustRequirements = {
   trustedPattern?: string;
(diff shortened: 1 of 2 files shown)


ubik2:
This wasn't really needed, but it was a place where we had improper code that I overlooked when I did the `$alias` rework (since it wasn't being used). I updated the code here to deal with the proper sigil links instead.

PR #4420, opened 2026-06-30
refactor: migrate OpaqueRef→Reactive symbols (CT-1756, 1/3)

mathpirate:
Stage 1 of 3 for [CT-1756](https://linear.app/common-tools/issue/CT-1756-remove-opaqueref-entirely-complete-the-reactive-rename) — completing the `OpaqueRef → Reactive` rename by removing the deprecated `OpaqueRef` spelling.

**Scope of "breaking":** this PR keeps the documented external surface — the `OpaqueRef` *type* alias (`export type OpaqueRef<T> = Reactive<T>`) stays until stage 3. It does rename **internal value exports** in place, without deprecated aliases — `isOpaqueRef` → `isReactive`, `OpaqueRefErrorTransformer` → `ReactiveErrorTransformer`, the builder `opaqueRef()` → `reactive()`. Those have no in-repo consumers remaining and aren't part of the documented authoring API, so per the agreed "ship the rename" plan they're renamed rather than aliased. The actual external break (deleting the type alias) lands in stage 3.

`OpaqueRef` has been an identity alias (`= T`) since #3153, so this is a rename of a name, not a type migration.

## What changed
- **api**: ~40 factory/builtin signatures → `Reactive`; deleted the dead, unused `OpaqueRefInner` helper; the `OpaqueRef` *type* alias is kept (removed in stage 3).
- **runner**: runtime guard `isOpaqueRef` → `isReactive` (+ `isReactiveMarker`); internal builder identifiers (`opaqueRef()` → `reactive()`, `getAsOpaqueRefProxy` → `getAsReactiveProxy`, `frame.opaqueRefs` → `reactives`, `opaqueRefWithCell` → `reactiveWithCell`); re-exports `Reactive`.
- **schema-generator**: `CellWrapperKind "OpaqueRef"` → `"Reactive"`. Kept `"OpaqueRef"` as a recognized `WrapperSpelling` (removed in stage 3) so authored `OpaqueRef` still classifies as reactive — output-neutral. Also removes a **pre-existing dead `kind === "OpaqueRef"` branch** in `object-formatter.ts` that the rename surfaced (`getCellWrapperInfo` is brand-based and never returns that kind — dead since #3153).
- **ts-transformers**: predicates/identifiers (`containsOpaqueRef` → `containsReactive`, `OpaqueRefErrorTransformer` → `ReactiveErrorTransformer`, …); preserved the spelling-axis `WrapperSpelling` tables and the historical `OpaqueRefMethods` test models.
- **consumers** (patterns, fs-sync-example, cli): type annotations.

One synthesized fixture output updated (the pattern self-type now emits `[SELF]: Reactive<any>`, forced by the api signature rename).

## Staging
- **This PR (1/3)** — symbol migration; the `OpaqueRef` *type* alias kept.
- **2/3** ([#4422](https://github.com/commontoolsinc/labs/pull/4422)) — file/dir/fixture renames + docs.
- **3/3** ([#4424](https://github.com/commontoolsinc/labs/pull/4424)) — delete the `OpaqueRef` type alias + `WrapperSpelling` → the actual external break.

## Verified locally
typecheck (api/runner/schema-generator/ts-transformers/consumers), `fmt --check`, `lint`, ts-transformers (342), schema-generator (27), runner targeted, generated-patterns integration (145) — all green.

> Stacked 1/3 — please don't merge until the full stack is ready.

### DIFF
--- a/packages/runner/src/builder/pattern.ts
+++ b/packages/runner/src/builder/pattern.ts
@@ -5,7 +5,7 @@ import {
   type FactoryInput,
   type Frame,
   type ICell,
-  isOpaqueRef,
+  isReactive,
   JSONObject,
   type JSONSchema,
   type JSONValue,
@@ -13,9 +13,9 @@ import {
   type Node,
   type NodeRef,
   type OpaqueCell,
-  type OpaqueRef,
   type Pattern,
   type PatternFactory,
+  type Reactive,
   type RequireDefaults,
   type Schema,
   type SchemaWithoutCell,
@@ -23,7 +23,7 @@ import {
   type toJSON,
   type UnsafeBinding,
 } from "./types.ts";
-import { opaqueRef } from "./opaque-ref.ts";
+import { reactive } from "./opaque-ref.ts";
 import { brandTrustedPattern, noteDerivedCopy } from "./pattern-metadata.ts";
 import {
   applyArgumentIfcToResult,
@@ -71,33 +71,33 @@ import { hardenVerifiedFunction } from "../sandbox/function-hardening.ts";
 // Function-only overloads (most common)
 export function pattern<T>(
   fn: (
-    input: OpaqueRef<RequireDefaults<T>> & { [SELF]: OpaqueRef<any> },
+    input: Reactive<RequireDefaults<T>> & { [SELF]: Reactive<any> },
   ) => any,
 ): PatternFactory<T, ReturnType<typeof fn>>;
 export function pattern<T, R>(
   fn: (
-    input: OpaqueRef<RequireDefaults<T>> & { [SELF]: OpaqueRef<R> },
+    input: Reactive<RequireDefaults<T>> & { [SELF]: Reactive<R> },
   ) => FactoryInput<R>,
 ): PatternFactory<T, R>;
 // Function + schemas overloads
 export function pattern<S extends JSONSchema>(
   fn: (
-    input: OpaqueRef<SchemaWithoutCell<S>> & {
-      [SELF]: OpaqueRef<any>;
+    input: Reactive<SchemaWithoutCell<S>> & {
+      [SELF]: Reactive<any>;
     },
   ) => any,
   argumentSchema: S,
 ): PatternFactory<SchemaWithoutCell<S>, ReturnType<typeof fn>>;
 export function pattern<S extends JSONSchema, R>(
   fn: (
-    input: OpaqueRef<SchemaWithoutCell<S>> & { [SELF]: OpaqueRef<R> },
+    input: Reactive<SchemaWithoutCell<S>> & { [SELF]: Reactive<R> },
   ) => FactoryInput<R>,
   argumentSchema: S,
 ): PatternFactory<SchemaWithoutCell<S>, R>;
 export function pattern<S extends JSONSchema, RS extends JSONSchema>(
   fn: (
-    input: OpaqueRef<SchemaWithoutCell<S>> & {
-      [SELF]: OpaqueRef<Schema<RS>>;
+    input: Reactive<SchemaWithoutCell<S>> & {
+      [SELF]: Reactive<Schema<RS>>;
     },
   ) => FactoryInput<Schema<RS>>,
   argumentSchema: S,
@@ -106,14 +106,14 @@ export function pattern<S extends JSONSchema, RS extends JSONSchema>(
 // Explicit T with optional schemas (e.g. pattern<{ x: number }>(fn, schema))
 export function pattern<T>(
   fn: (
-    input: OpaqueRef<RequireDefaults<T>> & { [SELF]: OpaqueRef<any> },
+    input: Reactive<RequireDefaults<T>> & { [SELF]: Reactive<any> },
   ) => any,
   argumentSchema: JSONSchema,
   resultSchema?: JSONSchema,
 ): PatternFactory<T, ReturnType<typeof fn>>;
 export function pattern<T, R>(
   fn: (
-    input: OpaqueRef<RequireDefaults<T>> & { [SELF]: OpaqueRef<R> },
+    input: Reactive<RequireDefaults<T>> & { [SELF]: Reactive<R> },
   ) => FactoryInput<R>,
   argumentSchema: JSONSchema,
   resultSchema?: JSONSchema,
@@ -121,7 +121,7 @@ export function pattern<T, R>(
 // Implementation signature
 export function pattern<T, R>(
   fn: (
-    input: OpaqueRef<RequireDefaults<T>> & { [SELF]: OpaqueRef<R> },
+    input: Reactive<RequireDefaults<T>> & { [SELF]: Reactive<R> },
   ) => FactoryInput<R>,
   argumentSchema?: JSONSchema,
   resultSchema?: JSONSchema,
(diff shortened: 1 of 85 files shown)


packages/runner/src/index.ts
```diff
@@ -122,8 +121,8 @@ export {
   ID,
   ID_FIELD,
   isModule,
-  isOpaqueRef as isOpaqueRef,
   isPattern,
+  isReactive as isReactive,
```
mathpirate: why do we need this `as`?

packages/runner/src/cell.ts
```diff
@@ -2504,7 +2502,7 @@ export class CellImpl<T extends FabricValue>
     }
 
     const result = flatMapFactory({
-      list: this as unknown as OpaqueRef<T>,
+      list: this as unknown as Reactive<T>,
```
mathpirate, line 2507: in passing, i am wondering why we need the cast through unknown here and elsewhere in this file

packages/runner/test/opaque-ref-schema.test.ts
```diff
@@ -10,7 +10,7 @@ import { isRecord } from "@commonfabric/utils/types";
 
 const signer = await Identity.fromPassphrase("test operator");
 
-describe("OpaqueRef Schema Support", () => {
+describe("Reactive Schema Support", () => {
```
mathpirate, line 13: remind me when we are gonna rename this file and other "*opaque-ref*" files? is that phase 2?

packages/runner/src/index.ts
mathpirate: Good catch — it's a redundant no-op self-alias that predates the rename (it was `isOpaqueRef as isOpaqueRef`); I kept the shape to minimize the diff. Dropping it to just `isReactive`.

packages/runner/src/cell.ts
mathpirate: This one's pre-existing — the rename only changed `OpaqueRef`→`Reactive` here. The `as unknown as` bridges a real structural gap: `this` is a concrete `Cell<T>`, but the factory input is typed `Reactive<T>`, the *erased* reactive annotation (`Reactive<T> = T`, an identity alias the transformer recognizes by spelling). The cell is the reactive value at runtime, but `Cell<T>` and `T` are statically unrelated, so TS needs the `unknown` hop to assert it — same reason `getAsReactiveProxy()` returns `as unknown as Reactive<T>`. Tidying these would mean reworking how the builder types reactive inputs, which is out of scope for the rename; happy to file a follow-up if you think it's worth pulling on.

packages/runner/test/opaque-ref-schema.test.ts
mathpirate: Yep, phase 2 — and it's already up: #4422 (stacked on this one) renames this file to `reactive-schema.test.ts` plus every other `*opaque-ref*` path (the `test/opaque-ref/` dir, the fixtures, `builder/opaque-ref.ts`→`reactive.ts`, the flattened transformer dir, etc.).

packages/runner/src/cell.ts
mathpirate: let's file a backlog follow-up assigned to me to investigate the fundamental type challenges here more please, thank you!

mathpirate: Filed [CT-1808](https://linear.app/common-tools/issue/CT-1808/investigate-the-type-system-challenges-forcing-as-unknown-as-reactivet) (Backlog, assigned to you, related to CT-1756) to investigate the underlying type-system gap — it captures the representative `as unknown as Reactive<T>` sites in `cell.ts`/`builder/` and the goal of relating the runtime cell types to the `Reactive` annotation without the `unknown` bridge. Linked back to this thread.

packages/schema-generator/src/formatters/object-formatter.ts
```diff
@@ -63,9 +63,9 @@ function getWrapperSchemaFromCallable(
     return { asCell: ["sqlite"] };
   }
 
-  // Also check if it's an OpaqueRef wrapping a Stream or Cell
-  if (wrapperInfo?.kind === "OpaqueRef") {
-    // Get the inner type of OpaqueRef
+  // Also check if it's a Reactive wrapping a Stream or Cell
+  if (wrapperInfo?.kind === "Reactive") {
```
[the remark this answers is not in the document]
mathpirate: Correct that it's dead — `getCellWrapperInfo().kind` comes from `brandToWrapperKind`, which maps the `opaque` brand to `OpaqueCell` and never returns the renamed kind, so the branch is unreachable. Worth noting it was equally dead on `main` as `=== "OpaqueRef"` (since #3153 made OpaqueRef an identity alias) — pre-existing, not introduced here. **Removed it**; schema-gen suite still 27/0, confirming it was unreachable. Thanks!

packages/runner/src/builder/pattern.ts
```diff
@@ -106,40 +106,40 @@ export function pattern<S extends JSONSchema, RS extends JSONSchema>(
...
   hardenVerifiedFunction(fn);
 
   // The pattern graph is created by calling `fn` which populates for `inputs`
-  // and `outputs` with Value<> (which containts OpaqueRef<>) and/or default
+  // and `outputs` with Value<> (which containts Reactive<>) and/or default
   // values.
   const frame = pushFrame();
 
-  const inputs = opaqueRef<RequireDefaults<T>>(
+  const inputs = reactive<RequireDefaults<T>>(
```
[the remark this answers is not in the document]
mathpirate: This is a pure rename, not a behavioral change — the diff here is just `opaqueRef<…>(…)` → `reactive<…>(…)` (the same function, exported under the new name). `pattern()` always called this function, so behavior is byte-identical to `main`. The comment-vs-body mismatch you'd infer (comment says "without a runtime", body requires one) is pre-existing and unchanged by this PR. `generated-patterns` integration (145 passing) exercises pattern declaration through the real pipeline. No change needed here.

packages/runner/src/index.ts
```diff
@@ -122,8 +122,8 @@ export {
   ID,
   ID_FIELD,
   isModule,
-  isOpaqueRef as isOpaqueRef,
   isPattern,
+  isReactive,
```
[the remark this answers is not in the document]
mathpirate: Intentional rather than an oversight. `isOpaqueRef` is internal infra (a builder type-guard), not part of the documented authoring API, and has zero in-repo consumers after this migration — so per the agreed "ship the rename" plan it's renamed in place rather than kept as a deprecated alias. What we *do* stage is the documented external surface, the `OpaqueRef` **type**, which stays through stages 1–2 and is removed in #4424. I've updated the PR description to scope "breaking" precisely.

packages/ts-transformers/src/diagnostics/error-message-transformer.ts
```diff
@@ -30,22 +30,22 @@ export interface OpaqueRefErrorTransformerOptions {
...
- * Transforms confusing OpaqueRef-related TypeScript errors into clear, actionable messages.
+ * Transforms confusing Reactive-related TypeScript errors into clear, actionable messages.
  *
  * For example, transforms:
  *   "Property 'get' does not exist on type 'OpaqueCell<number> & number'"
  * Into:
  *   "Unnecessary .get() call on a reactive value. This value can be accessed directly..."
  */
-export class OpaqueRefErrorTransformer implements DiagnosticMessageTransformer {
+export class ReactiveErrorTransformer implements DiagnosticMessageTransformer {
```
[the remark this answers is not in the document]
mathpirate: Same call as the `isReactive` rename: `ReactiveErrorTransformer` is internal (consumed only by the runner harness, migrated in this PR) and not a documented public export, so it's renamed in place rather than aliased — consistent with shipping the rename. See the updated PR description for the breaking-scope.

PR #4441, opened 2026-07-01
perf(runner): memoize compiled-body parse in buildRecordsFromCompiled

mathpirate:
## What

At piece boot the shared **system-app module closure** is rebuilt once per system
pattern loaded by identity, so `buildRecordsFromCompiled` re-parses every compiled
body to re-derive the same export/import record surface.

Measured on a trivial pattern's **cold** boot (isolated toolshed, current `main`),
the 9-module / 522 KB system-app closure is parsed **4×** — the 2nd–4th passes are
pure waste, since the derivation is a pure function of each module's compiled body:

| pass over the 9-module closure | before | after |
|---|---|---|
| 1st | 38.0 ms | 21.5 ms (the one real parse) |
| 2nd / 3rd / 4th | 19.7 / 18.6 / 18.2 ms | **0.0 / 0.0 / 0.0 ms** (memo hit) |
| **total** | **94.5 ms** | **21.5 ms** |

This is the 2nd-largest bucket of the ~640 ms generic per-piece cold-boot floor (a
trivial pattern pays the full system-app boot; the parse sits synchronously ahead
of first paint).

## Change

A memo (`exportParseCache` / `importParseCache`) in `module-record-compiler.ts`,
**keyed by the compiled body itself**. Both derivations — the export surface
(names + `export *` targets) and the runtime `require()` specifiers — are pure
functions of that body, so an equal body is guaranteed an equal parse. Each
distinct compiled body is parsed once per process → **~73 ms off the cold-boot
floor**, with byte-identical record output.

Keying on the body (not the module's source `identity`) is deliberate: a
content-hash `identity` hashes the *authored source*, and the same identity can
map to *different* compiled bytes across compilation modes / runtime versions (cf.
the `recordCache` guard in `compileSourcesToRecords`, where a precompiled body and
a bare-transpiled body share a content-hash key). A body key is exact, so
cross-contamination is impossible. The memo is process-global and unbounded by
design — one small record-surface entry per distinct body, session-lifetime.

<sub>(The table was measured on the initial identity-keyed prototype; the
parse-elision — and thus the wall-clock win — is identical under body-keying.)</sub>

## Why one parse and not zero

A SES virtual module record is `{ imports, exports, execute }`; the linker needs
the export names + import specifiers *before* `execute` runs. Modules are compiled
to CommonJS (no static export syntax) and the content-addressed cache stores the
compiled code but not the derived record surface, so it is reconstructed by parsing
on first sight. A follow-up persists the derived records at compile time to remove
the last parse entirely (0 on the critical path) — landing as a stacked PR on top
of this one.

## Tests / gate

- Regression tests in `build-from-compiled.test.ts`: transparency (repeated builds
  → identical records; guards shared-state corruption), plus cross-contamination
  guards for **both** the export and import memos (two different bodies under the
  same source `identity` each yield their own surface — proving the memo keys on
  the body, not the identity).
- Runner suite: **728 passed / 0 failed**.
- Integration `pattern-tests` (68) + `generated-patterns` (147): **green**.
- fmt / lint / check clean.

---
## Summary by cubic
Memoizes parsing in `buildRecordsFromCompiled`, keyed by the compiled body string. Collapses the shared system-app closure from 4× parses to 1×, saving ~73ms on cold boot (~94.5ms → ~21.5ms) with identical record output.

- **Refactors**
  - Adds process-global caches `exportParseCache` and `importParseCache`, keyed by compiled body (not source `identity`) to avoid cross-contamination across compile modes.
  - Introduces `parseCompiledExports` and `parseCompiledImports`; defensively copies cached arrays to prevent mutation on cache hits.
  - Adds tests for repeated-build transparency and body-keyed memoization for both exports and imports; all suites green.

<sup>Written for commit 562f01dc57baa48247f7d34e3c6d904a73c0dcfc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4441?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/sandbox/module-record-compiler.ts
+++ b/packages/runner/src/sandbox/module-record-compiler.ts
@@ -29,20 +29,26 @@ const TARGET = ts.ScriptTarget.ES2023;
 const logger = getLogger("module-record-compiler");
 
 /**
- * Content-addressed memo of the two pure derivations
- * {@link buildRecordsFromCompiled} extracts from a module's compiled body: its
- * direct export surface (names + `export *` target specifiers) and its runtime
- * `require()` specifiers. Keyed by the module's content-hash `identity`.
+ * Memo of the two pure derivations {@link buildRecordsFromCompiled} extracts
+ * from a module's compiled body: its direct export surface (names + `export *`
+ * target specifiers) and its runtime `require()` specifiers. **Keyed by the
+ * compiled body itself** — both derivations are pure functions of that body.
  *
  * At piece boot every system pattern loaded by identity rebuilds its record
  * graph from the SAME shared module closure, so `buildRecordsFromCompiled` runs
  * once per pattern over an identical module set and re-`createSourceFile`s every
- * body N times — the dominant cost of the warm boot path. Both derivations are
- * pure functions of the compiled body, and `identity` is that body's content
- * hash, so a parse is reusable across every call, closure, and space with no
- * staleness risk. This collapses the boot re-parse to one parse per distinct
- * module per worker. Cached arrays are treated as immutable; callers copy
- * before mutating or handing them to a consumer that might.
+ * body N times — the dominant cost of the warm boot path. This collapses that to
+ * one parse per distinct body per worker.
+ *
+ * Keying on the body (not the module's source `identity`) is deliberate: a
+ * content-hash identity is a hash of the *authored source*, and the same
+ * identity can map to *different* compiled bytes across compilation modes /
+ * runtime versions (cf. the `recordCache` note in `compileSourcesToRecords`,
+ * where a precompiled body and a bare-transpiled body share a content-hash key).
+ * An identity key could therefore serve one body's parse for another; the body
+ * fully determines the parse, so a body key is exact and cross-contamination is
+ * impossible. Cached arrays are treated as immutable; callers copy before
+ * mutating or handing them to a consumer that might.
  */
 const exportParseCache = new Map<
   string,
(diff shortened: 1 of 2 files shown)


packages/runner/src/sandbox/module-record-compiler.ts
```diff
@@ -28,6 +28,67 @@ const TARGET = ts.ScriptTarget.ES2023;
...
+function parseCompiledExports(
+  code: string,
+): { exportNames: readonly string[]; starTargetSpecs: readonly string[] } {
+  let hit = exportParseCache.get(code);
+  if (hit === undefined) {
+    const { names, starTargetSpecs } = extractCompiledExports(code);
+    // Copy both derived arrays into the cache entry so a cached parse can never
+    // be mutated through a reference the extractor might retain (symmetry with
+    // exportNames; the entry is handed out on every hit).
+    hit = { exportNames: [...names], starTargetSpecs: [...starTargetSpecs] };
```
ubik2, line 77: nit: potentially more array copying than needed, but these lists should be small. `names` is a set, so it needs to be converted, but starTargetSpecs has already been copied into a fresh array for us. Since we're the only caller, it makes more sense to alter the function to return the set as an array already.

`extractRuntimeImports` has the same issue (copied on line 865).

Ideally, these methods would return a `readonly string[]` instead, which directly communicates that the value is fixed. Dan also calls `freeze` on these, since he doesn't trust callers.

No need to do anything about this feedback. Just mentioning it.

mathpirate: thank you so much @ubik2! i folded in all of these points into https://github.com/commontoolsinc/labs/pull/4442 *except* the `freeze`; this is a hot path we're trying to optimize for perf and i think the `readonly` addition should be sufficient. lmk if you think that's wrong

also gonna go ahead and tag @danfuzz in here so that in a couple weeks you will see this exchange and can opine :) hope you're having a great vacation

PR #4447, opened 2026-07-01
perf(runner): factory-build — O(1) partial-cause dedup + sanitize memo

mathpirate:
## What

Two small, independent, correctness-preserving perf fixes in the pattern-build
path (`factoryFromPattern`), found while profiling cold-load. Neither changes
behaviour for real inputs.

### 1. O(1) partial-cause dedup (was O(N²))
`factoryFromPattern` assigns a distinct "partial cause" to each internal cell,
deduping each candidate against **every** previously-assigned cause via a
`deepEqual` scan — O(N²) in the number of internal cells. Anonymous causes carry
a monotonic counter and are unique by construction, so that scan is pure waste;
and the quadratic bites on every instantiation of a *large* pattern, not just at
boot. Replaced with an O(1) `Set` of canonical cause keys — `hashStringOf`, the
canonical value hash, which is exactly `deepEqual`'s comparison on the JSON
domain: key-order-insensitive records (sorted UTF-8 key order) and
`Object.is` number semantics (`-0` ≠ `0`, `NaN` = `NaN`), pinned in
data-model's value-hash tests.

### 2. Memoize `sanitizeSchemaForLinks` by frozen input identity
`sanitizeSchemaForLinks` (strips `asCell`, de-cycles into `$defs`) is a pure
function of `(schema, keepAsCell)`, and at pattern-build time the same
interned/frozen schema is sanitized repeatedly. Memoized by input identity (per
`keepAsCell` mode), guarded to frozen inputs — the identity-keyed-memo pattern
`traverse.ts` already uses. Cached outputs are deep-frozen for share-safety
(a consumer mutating a nested node throws loudly instead of silently
poisoning every later same-schema build); each call hands out a fresh,
caller-mutable top-level clone that shares the frozen sub-tree; non-frozen
inputs bypass the memo unchanged.

## Measurement note (honest sizing)
- The **partial-cause** win is ~0 on a *trivial* pattern (tiny N) — it's a
  scaling fix for large patterns / repeated instantiation, plus it deletes
  provably-useless comparisons.
- The **sanitize memo**'s win (~4ms at boot) is below single-CPU-profile noise,
  so it was verified **deterministically**: measured on a trivial cold boot,
  2571 sanitize calls, 46% repeat a frozen input, and those repeats carry
  **~half (3.4 of 6.7ms) of the `recursiveStripAsCellFromSchema` time** — the
  memo's exact ceiling.

## Tests
- Frozen-share pins: the memoized sub-tree is deep-frozen (nested mutation
  throws), the handed-out top stays caller-mutable.
- New regression tests for the sanitize memo: a repeated frozen input returns the
  same frozen object (memo hit via reference-equality), keyed per `keepAsCell`
  mode, no cross-contamination between inputs.
- Partial-cause semantics covered by existing `rehydrate-internal-default` /
  `patterns-regressions` / `pattern` tests; full runner suite green locally.

---
## Summary by cubic
Speeds up the pattern build by switching partial-cause dedup to O(1) via value hashing, and memoizes `sanitizeSchemaForLinks` for deep‑frozen inputs with deep‑frozen cached outputs while returning a fresh top‑level clone.

- **Refactors**
  - Partial-cause dedup: replace `deepEqual` scan with an O(1) `Set` keyed by `hashStringOf` (order‑insensitive, `Object.is`‑faithful), matching `deepEqual`.
  - `sanitizeSchemaForLinks`: identity-memoized for deep‑frozen inputs per `KeepAsCell`; cached strip is deep‑frozen and each call returns a fresh shallow top-level clone; mutable inputs bypass the cache.
  - Tests: add cases for memo hits, per-mode caching, fresh top with shared frozen sub-tree (mutation throws), and no cross-contamination across different frozen inputs.

<sup>Written for commit dd918997a99af10cbea6c110fc1b89f6de21dbcc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4447?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/builder/pattern.ts
+++ b/packages/runner/src/builder/pattern.ts
@@ -307,6 +307,44 @@ function factoryFromPattern<T, R>(
   };
 
   const assignedInternalPartialCauses = new Map<OpaqueCell<any>, JSONValue>();
+  // Canonical keys of the causes already assigned. Testing a candidate cause for
+  // collision is then an O(1) `Set` lookup instead of a `deepEqual` scan over
+  // every previously-assigned cause — the scan made this loop O(N²) in the number
+  // of internal cells (negligible for small patterns, but a scaling cliff for
+  // large ones, and it runs on every pattern instantiation, not just at boot).
+  const assignedCauseKeys = new Set<string>();
+  // Key-order-insensitive serialization, matching `deepEqual`'s order-insensitive
+  // record comparison, so `assignedCauseKeys.has(key)` is exactly the collision
+  // test the previous `deepEqual` scan performed. Causes are always JSON values
+  // with no `undefined`, so this is a faithful, collision-free key.
+  const causeKey = (cause: JSONValue): string => {
+    // `deepEqual` compares primitives with `Object.is` (so -0≠0, NaN=NaN) and is
+    // type- and key-order-insensitive for records. Encode with a per-type tag +
+    // Object.is-faithful numbers so `JSON.stringify`'s primitive collapses
+    // (-0→0, NaN/±Infinity→null, number 1 vs string "1") can't produce a false
+    // collision the previous `deepEqual` scan would not have.
+    const enc = (x: unknown): unknown => {
+      if (x === null) return " null";
+      switch (typeof x) {
+        case "string":
+          return " s" + x;
+        case "boolean":
+          return " b" + x;
+        case "number":
+          return " n" + (Object.is(x, -0) ? "-0" : String(x));
+        case "object":
+          if (Array.isArray(x)) return x.map(enc);
+          return Object.fromEntries(
+            Object.keys(x as Record<string, unknown>).sort().map(
+              (k) => [k, enc((x as Record<string, unknown>)[k])],
+            ),
+          );
+        default:
+          return " ?" + String(x);
+      }
+    };
+    return JSON.stringify(enc(cause));
+  };
   let anonymousPartialCauseCount = 0;
   const nextAnonymousPartialCause = (isStream: boolean): JSONObject => {
     const generated = { $generated: anonymousPartialCauseCount++ };
@@ -325,17 +363,16 @@ function factoryFromPattern<T, R>(
     let partialCause = name === undefined
       ? nextAnonymousPartialCause(isStream)
       : name as JSONValue;
-    while (
-      Array.from(assignedInternalPartialCauses.values()).some((used) =>
-        deepEqual(used, partialCause)
-      )
-    ) {
+    let key = causeKey(partialCause);
+    while (assignedCauseKeys.has(key)) {
       const generated = nextAnonymousPartialCause(isStream);
       partialCause = name === undefined
         ? generated
         : { name: name as JSONValue, ...generated };
+      key = causeKey(partialCause);
     }
     assignedInternalPartialCauses.set(top, partialCause);
+    assignedCauseKeys.add(key);
   });
 
   const cellReferenceForCell = (
--- a/packages/runner/src/builder/pattern.ts
+++ b/packages/runner/src/builder/pattern.ts
@@ -1,5 +1,6 @@
 import { isRecord } from "@commonfabric/utils/types";
 import { deepEqual } from "@commonfabric/utils/deep-equal";
+import { hashStringOf } from "@commonfabric/data-model/value-hash";
 import {
   type CellScope,
   type FactoryInput,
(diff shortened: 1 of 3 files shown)


packages/runner/src/builder/pattern.ts
```diff
@@ -307,6 +307,44 @@ function factoryFromPattern<T, R>(
...
+  // collision is then an O(1) `Set` lookup instead of a `deepEqual` scan over
+  // every previously-assigned cause — the scan made this loop O(N²) in the number
+  // of internal cells (negligible for small patterns, but a scaling cliff for
+  // large ones, and it runs on every pattern instantiation, not just at boot).
+  const assignedCauseKeys = new Set<string>();
+  // Key-order-insensitive serialization, matching `deepEqual`'s order-insensitive
+  // record comparison, so `assignedCauseKeys.has(key)` is exactly the collision
+  // test the previous `deepEqual` scan performed. Causes are always JSON values
+  // with no `undefined`, so this is a faithful, collision-free key.
+  const causeKey = (cause: JSONValue): string => {
```
ubik2: May want to use @danfuzz 's `hashStringOf` here.

AFAIK, that doesn't produce the same hash if the object keys are re-ordered, but I don't know if we need to protect users who are putting `{$kind: "stream", $generated: 0}` as their cause. This would have the same createRef result as the first auto-generated stream which has `{$generated: 0, $kind: "stream"}` as the cause, but have a different `hashStringOf` value, so it wouldn't be obvious. Since this really has to be deliberate, I'm ok breaking those patterns. The less crazy `Cell.for({$generated: 0})` is still supported, since there's only one key (in this case, the first generated cell is assigned cause `{$generated: 1}` to avoid colliding with the named cell with deceptively named cause `{$generated: 0}`.

Using `hashStringOf` also gets you lots of internal caching and object freezing.

packages/runner/src/builder/pattern.ts
```diff
@@ -325,17 +347,16 @@ function factoryFromPattern<T, R>(
     let partialCause = name === undefined
       ? nextAnonymousPartialCause(isStream)
       : name as JSONValue;
-    while (
-      Array.from(assignedInternalPartialCauses.values()).some((used) =>
-        deepEqual(used, partialCause)
-      )
-    ) {
+    let key = causeKey(partialCause);
```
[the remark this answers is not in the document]
ubik2: If you do change to `hashStringOf` here, `{$generated: -0)` and `{$generated: 0}` don't show up as a collision, since they get different hashes. However, the cell ids for the two cells will collide.

I'm leaning towards restricting `$generated` from the partialCause keys entirely. If someone does a `Cell.for({$generated: 0})`, we can just throw an exception and consider it an invalid pattern.

packages/runner/src/link-utils.ts
```diff
@@ -447,6 +454,26 @@ export function sanitizeSchemaForLinks(
...
+  // schema is sanitized repeatedly — measured ~46% of calls repeat a frozen
+  // input, carrying ~half of the total strip time. Only deep-frozen inputs are
+  // memoized (a mutable input's identity could go stale), matching the
+  // identity-keyed-memo guard `traverse.ts` uses; `isDeepFrozen` is O(1) for the
+  // already-frozen/cached inputs we hit here. The cache holds the canonical strip
+  // result; every call returns a fresh SHALLOW CLONE of it. The clone matters:
+  // the reactive graph keys on the sanitized schema's top-level object identity
+  // (returning a shared object changes recomputation), so each call needs its own
+  // top — while still reusing the (expensive) stripped sub-tree from the cache.
+  const memoizable = isDeepFrozen(schema);
```
ubik2, line 472: Should be able to use `internSchema` here, which gets you the frozen schema, together with a hash.
You can also call `internSchema` on the sanitized schema, then store the {flag, inputHash} to result in the map (or multilevel map as you have).

packages/runner/src/builder/pattern.ts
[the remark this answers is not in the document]
mathpirate: Resolved by deletion in `dd918997a`: `causeKey` now delegates to `hashStringOf` (the canonical value hash), so no number ever reaches `JSON.stringify` — `-0`/`+0` hash distinctly (float64 bits, with a dedicated `-0` cache entry) and `NaN` hashes canonically, i.e. exactly `deepEqual`'s `Object.is` semantics, pinned in data-model's value-hash tests. (For the record, the previous encoding also pre-tagged numbers into strings before stringifying, so the described collapse couldn't occur — but the hand-rolled encoding is gone regardless.)

packages/runner/src/builder/pattern.ts
mathpirate: Adopted in `dd918997a` — `causeKey` is now `hashStringOf(cause)`. Good news on your caveat: the value hash is key-order-**in**sensitive — `feedObjectValue` feeds plain-object keys in sorted UTF-8 byte order — so `{$kind, $generated}` vs `{$generated, $kind}` hash identically, and the reorder-evades-dedup-but-collides-at-`createRef` scenario can't arise. `-0`/`NaN` also carry `Object.is` semantics (pinned in data-model's tests), so this is exactly the old `deepEqual` collision test minus the bespoke encoding.

On banning `$generated` from user-supplied causes: agreed it has real merit (it deletes the whole "deceptively named cause" class at the root), but it makes today-valid patterns invalid, so it's out of scope for this behavior-preserving PR — tracked as CT-1822.

packages/runner/src/link-utils.ts
mathpirate: Went halfway in `dd918997a`: the cached output is now `deepFreeze`d — that's the share-safety half of your suggestion, without two side effects I didn't want to buy in this PR: `internSchema` freezes its **argument in place** (fine for our fresh outputs, but on the *input* side it would freeze caller-owned mutable schemas, so the `isDeepFrozen` guard has to stay regardless), and it **canonicalizes key order**, which would reorder sanitized schemas flowing into persisted pattern JSON.

The full upgrade you describe — key the memo by the interned input (value-dedup across identical-value inputs) and intern the output — is real and worth exploring with a hit-rate measurement; tracked as CT-1823. One note captured there for that exploration: a hash-keyed map retains outputs for the session, where the current WeakMap dies with the input schema.

ubik2: good catch on freezing the argument.

PR #4450, opened 2026-07-01
feat(ts-transformers): warn on read-then-mergeable-push in handlers

Hixie:
A mergeable `Cell.push` commits as a tail-relative `append` op and drops the op's own incidental array read from conflict detection, so disjoint appends from concurrent writers merge instead of conflicting. That makes one handler shape fragile: a handler that explicitly reads a collection (a `.get()` or an iteration) and then mergeable-`push`es to the same collection keeps its own read in the conflict set. Two such appends then conflict, and the loser retries — the dedup-then-push shape. The intent is usually better expressed as an identity-addressed `addUnique` (via `elementById`), which resolves concurrent same-key adds to one entity with no retry, or as a read-modify-write `set` for other content- dependent conditions.

Add a build-time diagnostic that flags this shape. The capability analysis already tracks, per parameter, which paths are read and which are written by the mergeable methods. Extend it with an optional `mergeablePushMisuseSink`: during the walk it records each `push` call site (root, path, node), and after the walk it reports a site only when the same root's read set contains that exact path. A bare push records only a write, so it never triggers itself; an unconditional push, or a read of a different path, is left alone. Only `push` is treated as a mergeable append — `addUnique`, `removeByValue`, and read-modify-write `set` are the recommended replacements and are not flagged. The check adds no work when no sink is supplied, keeping the summary hot path unchanged.

A new `MergeablePushValidationTransformer` runs the analysis over each handler/action/lift/computed callback on authored source and reports a `mergeable-push:read-then-push` diagnostic pointing at the two better expressions. It is a warning, not an error: the kept read keeps the shape safe today by forcing the conflict-and-retry, and there are legitimate read-then-push handlers in the tree, so the check nudges toward the better expression without failing the build. Warnings from several pushes to the same read collection collapse to one per handler.

The check is path-correlated but order- and purpose-blind, so a handler that appends and also runs an independent read-modify-write on the same list (for example appending and then trimming to a maximum length) also warns; the remedy there is to keep the independent read-modify-write in its own handler. This is noted in the docs alongside the diagnostic.

---
## Summary by cubic
Adds a build-time warning for handlers that read a collection and then mergeable-`push` to it, pointing authors to `elementById` + `addUnique` or `set`. Dedupes to one warning per collection per handler and across multiple call sites; no runtime changes.

- **New Features**
  - Added `MergeablePushValidationTransformer` to flag `mergeable-push:read-then-push` misuse with clear suggestions; warns, not errors.
  - Extended capability analysis with `mergeablePushMisuseSink` to correlate `push` sites with explicit reads; no overhead when unset.
  - Wired into the `cf` pipeline; only flags `push` (not `addUnique`, `removeByValue`, or `set`); updated docs and tests; exported `MergeablePushMisuse`.

- **Bug Fixes**
  - Ensured one warning per collection per handler and deduped across multiple call sites (e.g., `lift(cb)(input)`).

<sup>Written for commit 916c646c4f6028d8a102876b31e4b4e234bcab8e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4450?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/development/mergeable-collection-writes.md
+++ b/docs/development/mergeable-collection-writes.md
@@ -278,10 +278,28 @@ Two further responses make the keyed case cheaper and catch misuse (see
   elements") are not generally mergeable and keep a read-modify-write `set`/`push`,
   which the narrowing above keeps safe.
 
-- **Catch misuse at build time (future).** The transformer's capability analysis
-  already tracks per-handler reads and writes; a diagnostic can flag a handler
-  that reads a collection and then mergeable-`push`es to it, pointing at the
-  identity-addressed `addUnique` or at `set`. Not yet built.
+- **Catch misuse at build time (implemented).** The transformer's capability
+  analysis already tracks per-handler reads and the mergeable write methods. The
+  `MergeablePushValidationTransformer`
+  ([packages/ts-transformers/src/transformers/mergeable-push-validation.ts](../../packages/ts-transformers/src/transformers/mergeable-push-validation.ts))
+  runs that analysis over each handler and flags a handler that both reads a
+  collection (an explicit `.get()` or an iteration) and mergeable-`push`es to the
+  same collection path. The diagnostic (`mergeable-push:read-then-push`) points at
+  the identity-addressed `elementById` + `addUnique` for a uniqueness condition,
+  or at a read-modify-write `set` for other content-dependent conditions. It is a
+  warning, not an error: the kept read keeps the shape safe today (it forces the
+  conflict-and-retry), so the check nudges toward the better expression without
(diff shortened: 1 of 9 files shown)


mathpirate (approved):
lgtm, thanks for making this happen

PR #4452, opened 2026-07-01
refactor: define each storage patch operation in one place

Hixie:
A "patch operation" is one logical change a commit makes to a stored document — the members of the `PatchOp` union (replace/add/remove/move/splice/append/ add-unique/remove-by-value/increment), plus the author-invokable mergeable ops (push/addUnique/increment/removeByValue). A single op was implemented across roughly a dozen sites in four packages: the wire type, the apply function, three near-duplicate "which paths did it touch" matchers, the whole-value diff generator's structural-op guard, the author-facing Cell methods, the intent recording and the intent-to-wire build, four optional transport methods on two interfaces plus their two wrappers, and the pattern transformer's method classification. Adding or changing an op meant editing all of them, and several gaps failed silently — most damagingly, the four `record*` transport methods were optional (`?.`), so a wrapper that forgot one would drop the mergeable intent and fall back to a whole-value diff, losing the concurrency-merge the op exists to provide, with no error.

This gives each op a single definition that every site derives from.

Wire side (memory/v2/patch.ts): a `patchOpDescriptors` table, typed `Record<PatchOp["op"], …>`, holds each op's `apply`, its JSON-pointer-valued fields, and a `structural` flag. `applyOp`, the two engine conflict/reactivity matchers, the client pending-replay path computation, and the diff-generator invariant guard all derive from it (via `touchedPointerPaths`, `patchOpIsStructural`, `patchOpPointerFields`) instead of re-enumerating the ops. A missing descriptor — or one for a tag absent from the union — is a compile error, so a new structural op is even covered by the array-index guard automatically.

Mergeable side (runner/src/storage/mergeable-ops.ts): each mergeable op's descriptor owns whether a delta is a no-op, how repeated deltas fold into one intent, and how an intent becomes wire ops plus diff-suppression. The transaction records intent through a single `recordMergeableOp(address, delta)` carrying a discriminated `MergeableOpDelta`, replacing the four op-specific methods across the interfaces, the ExtendedStorageTransaction, and the TransactionWrapper — so a new op needs no new transport method. `foldMergeableIntent` / `isNoopMergeableDelta` / `buildMergeableIntent` index the table by op tag and defer; none of them enumerate the ops.

Author/registration side: `MERGEABLE_OP_METHODS` in `@commonfabric/api` is the canonical catalog (author method, wire tag, transformer classification). The transformer's write-method sets fold it in rather than hardcoding the mergeable methods, so a new op is classified with no transformer edit. A consistency test cross-checks the api catalog, the runner registry, and the wire descriptors, so a mergeable op present in one but missing from another fails loudly.

Behavior is unchanged throughout: apply, the three path computations (including `move`'s ordering and structural-op parent injection / replay), the mergeable fold rules (same op combines, a different op replaces), the empty-tail-op drop, the net-zero-increment and no-base tail-slice rules, and the transformer set contents all move verbatim. Adding an op is now a wire descriptor, a mergeable descriptor, a catalog entry, and the Cell method — each gap a compile error, a failing consistency test, or an uncallable method, never a silent whole-value diff. `docs/development/patch-operations.md` documents the family, the registries, and the add-an-op checklist. Memory, runner (storage / commit / conflict / reactivity / scheduler / cell), and ts-transformers suites pass.

---
## Summary by cubic
Unifies patch-operation behavior behind registries so all apply logic, touched-paths, and mergeable intent derive from one source. Also fixes a conflict gap: parent nonRecursive readers now conflict when a mergeable op creates a previously-absent path via a new per-op `createsKey` flag.

- **Refactors**
  - Added `patchOpDescriptors` in `packages/memory/v2/patch.ts` and helpers (`touchedPointerPaths`, `patchOpIsStructural`, `patchOpChangesParentKeySet`, `patchOpPointerFields`) that drive `applyPatch`, engine conflict/reactivity matching, client pending-replay paths, and the array-index invariant.
  - Introduced runner `mergeable-ops` registry and a single `recordMergeableOp(address, delta)` API; intent fold/build is per-op; builders stamp `createsKey` when the transaction base had no value at the path.
  - Added `MERGEABLE_OP_METHODS` in `@commonfabric/api`; transformers derive writer-method sets from it; a consistency test cross-checks the API catalog, runner registry, and wire descriptors.
  - Types and docs: `append`/`add-unique`/`increment` carry optional `createsKey`; added `docs/development/patch-operations.md` and updated conflict-granularity and mergeable-writes docs; transformer and call sites (`Cell`, query-result proxy) now use `recordMergeableOp`.

- **Migration**
  - If you implement `IStorageTransaction` or `IExtendedStorageTransaction`, replace `recordArrayAppend`/`recordAddUnique`/`recordIncrement`/`recordRemoveByValue` with `recordMergeableOp(address, delta)`.
  - No app code changes required; Cell methods now call `recordMergeableOp` internally.

<sup>Written for commit cd9266be85b8a320ef9574f1457288af8bbe8658. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4452?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


docs/development/patch-operations.md
```diff
@@ -0,0 +1,166 @@
...
+The op tags are the members of the `PatchOp` union in
+`packages/memory/v2.ts`:
+
+| op                | shape                          | how it arises |
+| ----------------- | ------------------------------ | ------------- |
+| `replace`         | `{ path, value }`              | generated by the whole-value diff |
+| `add`             | `{ path, value }`              | generated by the whole-value diff |
+| `remove`          | `{ path }`                     | generated by the whole-value diff |
+| `move`            | `{ from, path }`               | applied only (never generated by the runner) |
+| `splice`          | `{ path, index, remove, add }` | generated by the whole-value diff |
```
ubik2, line 26: This one worries me a little, since we're mixing intent and op. There's some room for expressing both with the read dependency list, but I don't know how we'd know from just looking at the whole-value diff.
1. client1 ["foo", "bar"] => ["foo", "cab", "bar"]
2. client2 {"foo", "bar"] => ["foo", "dog", "bar"]

Should this be resolved as ["foo", "dog", "cab", "bar"], or should it generate a conflict on the second update?

PatchOps can still capture read dependencies, so perhaps this is already well handled by client2 having a read of the doc that invalidates the splice, but it's not obvious to me in the example what we want.

I also should look at how splice is implemented in memory v2. A common pattern here is that the read set guides your interpretation of index (it's not the index in the current value, it's the index in the value you read). Here, splice is often used in text regions so you can have:
1. client1: "it was a cold day" => "it was a cold and rainy day" with a splice([], 9, 0, "and rainy ")
2. client2: "it was a cold day" => "it was a cold night" with a splice([], 13, 3, "night")

Since client2 didn't have client1's change, the index is adjusted so its value in client2's read version gets updated to the server's version, and we convert that to a splice([], 23, 3, "night"). This gets you the desired merge behavior, where the final text is "it was a cold and rainy night".

ubik2: @Hixie pointed out that this isn't new code, and it already is generated.
I missed this usage in `buildArrayPatchCandidates`.
There's no special intent or merge behavior, so it still has the conflict based on stale read to prevent simultaneous edits.

Edit: also, want to add that splice does not currently have any of this smart behavior on the server.

Hixie: (Discussed on Discord, @ubik2 said to disregard.)

PR #4453, opened 2026-07-02
Make client idle await user-intent event commits so a create survives…

Hixie:
… an immediate reload

Creating a profile in the home space and then reloading the page dropped the new profile from durable storage. The same mechanism lost the most recent of several profiles created in quick succession before a reload, while the earlier ones survived.

A profile create is an event-handler write (`profiles.push(inSpace()(...))`), which is a cross-space commit. The scheduler issues event-handler commits fire-and-forget so it can keep processing later events against the locally applied state while server confirmation is still in flight. The client-facing notion of "idle" — what quiescence checks wait on — maps to the worker's `handleIdle` -> `runtime.idle()` -> `scheduler.idle()`, which reports quiescence as soon as the scheduler has no more reactive work to do. That happens while the create's commit is still travelling to the server. A full page reload tears down the page and its runtime worker at the browser level, and no graceful `Runtime.dispose()` runs on a reload (dispose runs only on an in-app runtime swap, such as logout or an identity change). So a create whose commit had not yet been confirmed was lost. The newest of several rapid creates was the most exposed, because its commit had the least time to confirm before the reload.

This makes the client-facing idle a trustworthy "safe to navigate or reload" checkpoint by waiting for in-flight user-intent event commits, not just for reactive quiescence. The scheduler tracks the commit of each event handler that changed a value; the client-facing idle waits for those to settle. In-app navigation was already durable through the navigation-convergence path (an idle followed by a connection-wide synced), and dispose already flushes pending commits before tearing storage down; this closes the remaining gap for a client that treats idle as a safe point to reload.

The wait is not a separate loop with a round cap. A landed event commit dirties the readers of the committed write (onEventCommitWrites -> markReadersDirtyForChangedWrites), which can re-trigger scheduler work, which can in turn produce further commits. So "safe to reload" is a joint fixpoint of two coupled subsystems: the reactive scheduler is quiescent AND no user-intent commit is in flight. The scheduler already converges its own fixpoint — a running pass, background tasks — by re-checking every condition each time the awaited thing settles, with no cap: a system that settles converges, one that never settles never reports idle. In-flight event commits are folded in as one more such condition. `idle()` and a new `idleWithEventCommits()` share a single `#waitForQuiescence(awaitEventCommits)` routine; the client-facing idle passes `true` so a pending commit is awaited and re-checked exactly like a background task, while the internal `idle()` passes `false` and is behaviorally unchanged (internal reactive-quiescence callers — builtins, cell sync, dispose, settled — must not block on commit I/O).

Parked waiters need one piece of care. A waiter that parks in `idlePromises` (behind a wake timer or a lineage head) is resolved when the scheduler drains, but a commit can still be in flight then, so the commit-aware variant re-checks instead of resolving. That re-check is deferred to a microtask: `resolveIdlePromises` drains the array before the scheduler marks itself not-scheduled, so a synchronous re-check would re-enter `#waitForQuiescence` and push back into the array mid-drain, spinning it. Deferring runs the re-check after the drain, once the scheduler state has settled.

Adds an integration test that creates profiles using only reactive-idle waits between steps — never a durability sync — then reloads, fully settles, and asserts each created profile's name renders; it fails without the fix and passes with it. Adds a scheduler unit test covering that plain `idle()` still returns while a commit is in flight but `idleWithEventCommits()` waits for it, that it converges when called while the scheduler is busy (a parked waiter, both with and without a pending commit — the no-commit case is the one that spins without the microtask deferral), that a rejecting commit settles the wait rather than wedging it, and that only value-changing events register a commit.

---
## Summary by cubic
Make the client “idle” a durable checkpoint by waiting for storage-layer pending commits, so profile creates and other writes survive an immediate reload. Also prompt before a reload that would drop unconfirmed writes.

- **Bug Fixes**
  - Added a storage pending-commit barrier: `trackPendingCommit()`, `hasPendingCommits()`, `pendingCommitsSettled()`; transactions register commit promises synchronously.
  - Scheduler adds `idleWithPendingCommits()` and folds commit durability into quiescence; `idle()` remains reactive-only. Parked waiters re-check via microtask to avoid spins.
  - `RuntimeProcessor.handleIdle` now awaits `idleWithPendingCommits()`; event commits stay fire-and-forget but are covered by the barrier. Tests added for barrier behavior, fixpoint convergence, rejected-commit settling, and profile durability across reloads; added tests for notifier error isolation and the `beforeunload` guard.

- **New Features**
  - Worker mirrors pending-commit transitions to the page; `PendingWritesChanged` notification feeds `RuntimeClient.hasPendingWrites()` and a `pendingwriteschange` event.
  - The shell shows a `beforeunload` confirmation while unconfirmed writes exist; the integration harness accepts these dialogs (“Leave”).

<sup>Written for commit 0ae83edd6f6746731f3840ce5ce1d1bdde346f39. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4453?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

NEW_COVERAGE_BASELINE

### DIFF
--- a/packages/runner/src/scheduler.ts
+++ b/packages/runner/src/scheduler.ts
@@ -961,16 +961,59 @@ export class Scheduler {
   }
 
   idle(): Promise<void> {
+    return this.#waitForQuiescence(false);
+  }
+
+  // Client-facing quiescence: reactive quiescence AND durability of in-flight
+  // commits. Commits are issued fire-and-forget (event handlers, direct cell
+  // writes over IPC, reactive recomputation write-backs), so plain idle()
+  // reports quiescence while a commit is still travelling to the server; a
+  // client that reads idle as a safe point to navigate or reload would then
+  // drop that write when the page and its worker are torn down. The pending
+  // set is sourced from the storage manager — the single chokepoint every
+  // commit flows through — so no write path can be forgotten. A landed commit
+  // also dirties readers of the committed write (onEventCommitWrites), which
+  // can re-trigger scheduler work that produces further commits, so durability
+  // and reactive quiescence are one joint fixpoint. This reuses the same
+  // recursive convergence idle() uses — a pending commit is just one more
+  // class of work to wait for — so there is no separate retry loop and no
+  // round cap. It resolves exactly when the scheduler is quiescent and no
+  // commit is in flight, and (like idle()) never resolves for a system that
+  // genuinely never settles.
+  idleWithPendingCommits(): Promise<void> {
+    return this.#waitForQuiescence(true);
+  }
+
+  #waitForQuiescence(awaitPendingCommits: boolean): Promise<void> {
     return new Promise<void>((resolve) => {
+      // Re-evaluate every condition from scratch once the thing we are waiting
+      // on settles.
+      const recheck = () =>
+        this.#waitForQuiescence(awaitPendingCommits).then(resolve);
+      // A parked waiter (idlePromises) resolves when the scheduler drains, but a
+      // commit can still be in flight then, so the commit-aware variant re-checks
+      // instead of resolving. That re-check is deferred to a microtask: it must
+      // not re-enter #waitForQuiescence synchronously while resolveIdlePromises
+      // is iterating idlePromises (draining runs before the scheduler marks
+      // itself not-scheduled, so a synchronous re-check would push back into the
+      // array mid-drain). The plain variant keeps the direct resolver.
+      const park = awaitPendingCommits
+        ? () => queueMicrotask(recheck)
+        : resolve;
       if (this.runningPromise) {
         // Something is currently running - wait for it then check again
-        this.runningPromise.then(() => this.idle().then(resolve));
+        this.runningPromise.then(recheck);
       } else if (this.backgroundTasks.size > 0) {
         // Async scheduler work, such as event-triggered auto-start, is still in
         // flight. Wait for it to settle and then re-check the scheduler state.
-        Promise.allSettled([...this.backgroundTasks]).then(() =>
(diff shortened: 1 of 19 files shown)


packages/runner/src/scheduler.ts
```diff
@@ -961,16 +971,52 @@ export class Scheduler {
...
+      const recheck = () =>
+        this.#waitForQuiescence(awaitEventCommits).then(resolve);
+      // A parked waiter (idlePromises) resolves when the scheduler drains, but a
+      // commit can still be in flight then, so the commit-aware variant re-checks
+      // instead of resolving. That re-check is deferred to a microtask: it must
+      // not re-enter #waitForQuiescence synchronously while resolveIdlePromises
+      // is iterating idlePromises (draining runs before the scheduler marks
+      // itself not-scheduled, so a synchronous re-check would push back into the
+      // array mid-drain). The plain variant keeps the direct resolver.
+      const park = awaitEventCommits ? () => queueMicrotask(recheck) : resolve;
```
Hixie: I am a bit concerned that this is a hot retry loop in fancy clothes.

Hixie: I guess not really, if what we're trying to do is quiesce.

Hixie:
(waiting on a Fable review)

PR #4454, opened 2026-07-02
fix(runner,cli): register evaluated modules in the pattern-test harness (CT-1811)

mathpirate:
## Summary

Fixes **CT-1811**. A sub-pattern that exposes a derived-internal computed output, nested as a **grandchild** pattern node under a `.map()` / `filter` / `flatMap`, threw at bind time under the `cf test` harness — `Error: Unknown derived internal cell with partial cause: "<name>"` — while working fine in the deployed runtime.

## Root cause

The pattern-test harness (`test-runner.ts`) and the multi-user worker evaluate the test bundle via `Engine.compileAndEvaluateModules` and use the returned namespace directly, **without registering the evaluated artifacts**. The deployed runtime loads patterns via `PatternManager.patternFromEvaluation`, and the generated-patterns harness via `compilePattern` — **both call `registerEvaluatedModules`**.

Without registration, anonymous map ops never get a content-addressed entry ref, so `substituteOpPatternRefs` is a no-op and the op falls back to its **embedded pattern graph**. The embedded graph's nested output-alias `defer` levels get decremented one step too far by the `getImmutableCell` round-trip, so a grandchild derived-internal output resolves one instantiation level too early and throws.

This was a genuine **harness-vs-runtime code-path divergence**, not the strict-vs-lenient binding the ticket suspected (that is pattern-dependent, identical in both). An adversarial audit of 47 candidate divergences found this to be the *only* one that changes pattern semantics.

## Fix — a sealed load seam

Rather than have the harness "remember to also register" (the very drift that caused this), registration is **fused into a single load seam**:

```ts
// PatternManager
compileAndRegisterModules(program, opts?): Promise<EvaluateResult>
//   = compileAndEvaluateModules + registerEvaluatedModules
//   returns the full namespace (main.default, fetchMocks, multi-user descriptors)
```

Both harness callers use it, so an evaluated namespace **cannot** be obtained without registration — mirroring the runtime's own `patternFromEvaluation` path. `Engine.compileAndEvaluateModules` is documented as the low-level "does **not** register" primitive (for inspecting serialized/verified output without running; folding registration into it was rejected because engine serialization unit tests rely on the non-registering form). This **dissolves the divergence structurally**, not by discipline.

## Tests

- **Regression** (`gideon-tests/ct-1811-mapped-subpattern-derived-output.test.tsx`): fetch-free grandchild-under-map derived-output case; fails without the fix, passes with it. (Fetch was incidental to the ticket — any derived-internal output of a mapped grandchild trips it.)
- **Conformance** (`runner/test/compile-and-register-modules.test.ts`): asserts the seam indexes artifacts (default export gets an entry ref) while the bare primitive does not — pins the contract so the class can't silently return.

## Verification

| Gate | Result |
|---|---|
| Conformance test | 2/2 |
| Regression test (fails-without / passes-with, via the seam) | ✅ |
| Full pattern-tests suite (incl. multi-user) | 84/84 |
| generated-patterns suite | green |
| Runner units (engine, pattern-manager, map-op-by-identity, artifact-index, content-addressed-identity, load-by-identity, derived-copy) | green |
| fmt · lint · check | clean |

## Follow-up

- [CT-1812] — the embedded-graph path's `defer` corruption is a separate latent bug (now only reachable by genuinely ref-less dynamic ops).

### DIFF
--- a/packages/cli/lib/multi-user-test-worker.ts
+++ b/packages/cli/lib/multi-user-test-worker.ts
@@ -222,13 +222,13 @@ const handlers: Record<
         args.root as string | undefined,
       ),
     );
-    const evalResult = await engine.compileAndEvaluateModules(program, {
-      patternCoverage,
-    });
-    // Index evaluated artifacts so map/filter/flatMap ops resolve via their
-    // content-addressed canonical artifact instead of the defer-corrupted embedded
-    // graph (CT-1811); see test-runner.ts. Idempotent per identity.
-    runtime.patternManager.registerEvaluatedModules(evalResult);
+    // `compileAndRegisterModules` seals compile + evaluate + register (see
+    // test-runner.ts): map/filter/flatMap ops resolve via their content-addressed
+    // canonical artifact instead of the defer-corrupted embedded graph (CT-1811).
+    const evalResult = await runtime.patternManager.compileAndRegisterModules(
+      program,
+      { patternCoverage },
+    );
     const { main } = evalResult;
     // Channel 2: snapshot logger counts AFTER compile, before the run phase.
     loggerCountsBeforeRun = snapshotLoggerErrorWarnCounts();
--- a/packages/cli/lib/multi-user-test-worker.ts
+++ b/packages/cli/lib/multi-user-test-worker.ts
@@ -222,9 +222,14 @@ const handlers: Record<
         args.root as string | undefined,
       ),
     );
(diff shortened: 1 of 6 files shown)


ubik2 (approved):
I don't exactly see how the defer level got off from this, but it's easy to imagine things going wrong there.
In any case, the runtime changes seem safe.
lgtm.

PR #4457, opened 2026-07-02
Multiplayer convergence: reader blackout (fixed) + writer integration gap (open)

bfollington:
# Multiplayer convergence: reader blackout (fixed) + writer integration gap (open) — a decision-ready package

**Status: draft, for @robin / @gideon to pick up and decide how (or whether) to land.** This is the "replicable test case" from the convergence thread, worked all the way to root cause, with one fix included and the remaining issue pinned with costed options. Nothing here auto-lands anything — it's a package.

## TL;DR

The group-chat "profile rows never populate" / "never settles, flickers" symptoms trace to **three independent defects**. This PR:

- **Fixes B2** (the reader blackout) with a small, local traverse change — with before/after data.
- **Pins B1 and B3** with deterministic tests and root-cause analysis; leaves the fix decisions to you (both are design calls with trade-offs).

| defect | what it is | status here |
| --- | --- | --- |
| **B1** | a live `PerUser` cell pushed into shared state stores a scope-generic `"user"` link (owner DID lost) → readers resolve it into their own empty partition | **root-caused**, design call (link-mint owner scope) |
| **B2** | element schema `required` + an unresolvable link target voids the *whole array read* for non-authors → reader blackout | **FIXED in this PR** (+ data) |
| **B3** | a writer that interleaves its own append never integrates a peer's concurrent append (same-seq value divergence the delta protocol can't repair) | **root-caused + pinned**, 3 costed options |

B1+B2 compose into the blackout; B3 is separate (writer-side sync ordering).

## How to run

```
deno test -A packages/patterns/integration/convergence-storm.test.ts        # B2 fixed (green) + B3 open (red)
deno test -A packages/patterns/integration/cellset-lww-lost-update.test.ts  # LWW characterization (green)
deno test -A packages/runner/test/array-append-client-convergence.test.ts   # B3, deterministic (red)
```

**How to read the test results:** green = what this PR achieves; the two intentionally-**red** tests are the pinned, still-open **B3** (they assert the desired end-state and stay red until B3 is fixed — they're the red-to-green target for whichever option you pick). Every fixture variant differs from the next by exactly one ingredient, so diffing them shows what each defect needs.

---

## B2 — the fix (included)

**Symptom:** one settled post, zero contention — a second session reads the shared list as **empty**, while raw storage reads show its replica holds both the array and the linked doc in full. It's the schema-aware read, not replication.

**Mechanism** (`traverse.ts`, verified by instrumented runs): an element carries a link (`authorProfile`) whose target isn't resolvable *for this reader* (foreign `user` scope / cross-space not-yet-loaded / unwritten). The property is dropped → the element fails its `required` check → `traverseObjectWithSchema` returns `undefined` → `INVALID_OBJECT` → `traverseArrayWithSchema` voids the **entire array**. Sub-path reads (`messages.0.author`) still work; only the schema-shaped read of the container collapses.

**Fix:** exempt properties whose value is an *unresolvable link* from the object `required` check — an unresolvable reference degrades that field, it must not void the parent object and its containing array. This mirrors the grace already applied to array elements (`traverse.ts:3540`, *"still return a cell for it instead of invalidating the parent array"*) and to inline-valued asCell properties.

**Before/after** (deterministic, in-process; `convergence-chat` fixture *with* the shared derived reader — the wedge-escalation path; 2 writers + a non-writing observer):

| | observer (never writes) | conflicts / run |
| --- | --- | --- |
| before | `{}` — blackout, 3/3 runs | **44–45** |
| after | **full 40** | **1–2** |

So the fix cures the blackout **and collapses the conflict storm ~95%** — removing the seq-0 void reads removes the permanently-stale derived-cell commits that were the "wedge" (the runtime re-sending an identical rejected ~110 KB commit forever; see the evidence appendix). The residual failure is exactly B3 below.

**Two notes on the fix:**
- A narrower version (grace only when `hasAsCell`) mostly *didn't fire*: during element traversal the effective property schema frequently **lacks the `asCell` annotation** — it's dropped from the stored link's embedded / argument-side `$defs`, so the reader descends *into* the linked profile as a plain `required` prop. That asCell-annotation loss may warrant its own look; the `required`-exemption is what actually heals it.
- **Golden impact = corroboration:** the only test-surface change is the `traverse-replay` goldens (regenerated in this PR). Instructively, they shifted `INVALID_OBJECT → ok:true` in the **notebook, shopping-list, and toolshed-reload** fixtures — i.e. those *real* patterns were hitting the same void.
- B2 does **not** address B1: with the fix, a reader gets a cell resolving into its *own* partition, so `authorProfile` content is still per-reader-wrong. B1 (owner scope at link-mint) is the identity half.

---

## B3 — root-caused + pinned, fix is your call

**Symptom** (`array-append-client-convergence.test.ts`, 60 ms, deterministic, RED): seed `[seed]`; session 1 appends `A`; session 2 appends `B` without observing `A`. The server merges all three, a fresh reader sees `[seed, A, B]`, but **session 2's own replica** — after explicit `sync + pull` — reads `[seed, B]`. `A` is gone from the interleaving writer's view.

**Mechanism** (instrumented `applySessionSync` + `confirmPending`):
- `confirmPending` (`v2.ts:2625–2640`) promotes session 2's `confirmed` to its own commit's (high) `applied.seq` by replaying its append onto its **own stale base** = own appends only.
- The `AppliedRevision` for a `patch` op (append) carries only `patches`, **not** the merged document (only `set` ops carry `document`, `engine.ts:4011`), so the merge never returns through the commit response.
- Session 2's own commit advanced its server-side session cursor to head, so delta-sync re-delivers nothing on `pull()`. Its confirmed **value** (`[seed,B]`) and the server's **value** (`[seed,A,B]`) now diverge **at the same seq** — a seq-based delta protocol can't detect or repair that.

**Three viable fixes, all touching the core sync path (hence: your call):**
1. **Client fresh-watch resync** of the raced doc after a mergeable-append commit. Pure client; works because the server snapshot is keyed per-watch-id, not per-session-cursor (`server.ts:1947`). Cost: +1 round-trip per append unless gated on detected concurrency.
2. **Server merge-signal** on the revision (`merged` / `rebasedAgainstSeq`) → client resyncs only flagged docs. Targeted.
3. **Merged `document` in the applied revision** for merged commits → client adopts it directly. Cleanest semantically; costs materialize-per-commit + wire size.

The failing test pins the behavior for whichever way you go.

---

## B1 — root-caused, design call

The stored link is `{"link@1": {..., "scope": "user", ...}}` — scope-generic, no owner DID. `"user"` scope is principal-relative, so shipping it to another principal silently changes the referent. Fix direction: capture the owner's partition (`user:<did>`) at link-mint, or resolve the scope at write time. Same class as the known profile owner-principal issue. Controls in this PR show it: the `spacelink` fixture (PerSpace link) converges; the `optlink` fixture (optional field) converges with `authorProfile` absent for readers.

---

## What's in the PR

- **Fix:** `packages/runner/src/traverse.ts` (B2 grace) + regenerated `traverse-replay` goldens.
- **Tests:** `convergence-storm.test.ts` (B2 fixed → green; observer convergence under a 40-msg storm → green; B3 writer gap → red), `cellset-lww-lost-update.test.ts` (green characterization of the #4245 removeItem/updateItem LWW trade-off), `array-append-client-convergence.test.ts` (B3, deterministic red).
- **Fixtures:** `convergence-chat{,-noderived,-spacelink,-optlink,-plain}`, `lww-list` — one-ingredient-apart variants.
- **Harness affordances:** per-session `wsDelayMs` (WS latency shim in the worker realm), `send({idle:false})` (deep optimistic pipelines), `rawRead` (replica read bypassing the schema/result path).
- **Tools & context:** `storm-driver.ts` (parametric bisection), `docs/plans/2026-07-02-convergence-evidence-appendix.md` (browser-scale autopsy: per-principal landed counts, wedge wire anatomy, DOM census, and the intermediate theories that were falsified along the way).

## Also answered: the #4245 LWW question

`cellset-lww-lost-update.test.ts` characterizes the trade-off you were weighing: a blind removeItem-style whole-array `set` racing a CAS `push` **lands with no conflict and silently clobbers the pushed element for everyone**; the identical race through CAS pushes surfaces a `ConflictError`. Data for the "add mergeable remove-by-value ops vs accept LWW" decision.


bfollington:
Added a curated evidence appendix to the branch — the measurements behind the PR body: in-process bisection matrix + minimal-case probes, the durable-record autopsy tables (per-principal landed counts across 5 browser runs), the wedge's wire anatomy (advancing localSeqs, the 108KB seq-0 commit, pending-dependency cascade), DOM census, and the intermediate theories that were falsified along the way: [docs/plans/2026-07-02-convergence-evidence-appendix.md](https://github.com/commontoolsinc/labs/blob/ct-reader-blackout-repro/docs/plans/2026-07-02-convergence-evidence-appendix.md)

bfollington:
## Follow-up: B2 spike heals the blackout (before/after), and B3 root-caused + pinned

Two updates from spiking fixes against the repro.

### B2 — a candidate fix, with before/after profiling

The blackout collapse (`required` voiding the whole array read) has a small, local fix: exempt properties whose value is an **unresolvable link** from the object `required` check — an unresolvable reference degrades that field; it must not void the parent object and its containing array.

Diff (branch `ct-b2-ascell-grace-spike`, on `packages/runner/src/traverse.ts`):

```diff
     this.traverseObjectCalls++;
     const filteredObj: Record<string, Immutable<FabricValue>> = {};
+    const unresolvableLinkProps = new Set<string>();
     for (const [propKey, propValue] of Object.entries(doc.value!)) {
       ...
         const { ok: val, error } = SchemaObjectTraverser.hasAsCell(propSchema)
           ? this.tx.runWithAmbientReadMeta(excludeReadFromConflict, descend)
           : descend();
+        if (error !== undefined && isSigilLink(propValue)) {
+          unresolvableLinkProps.add(propKey);
+        }
         if (error === undefined) {
           filteredObj[propKey] = val;
+        } else if (!this.traverseCells && SchemaObjectTraverser.hasAsCell(propSchema)) {
+          // asCell boundary: return a cell for the unresolved target instead of
+          // dropping the property (mirrors the existing array-element / inline-value grace).
+          const cellLink = getNextCellLink(propDoc, propSchema);
+          filteredObj[propKey] = this.objectCreator.createObject(cellLink, undefined);
         }
       }
     }
     ...
         for (const requiredProperty of required) {
+          if (unresolvableLinkProps.has(requiredProperty)) continue;
           if (!(requiredProperty in filteredObj)) { ... return undefined; }
```

**Before/after** (deterministic in-process, `convergence-chat` fixture *with* the shared derived reader — i.e. the wedge-escalation path; `storm-driver.ts 20 2 pipeline 0 convergence-chat`, 2 writers + non-writing observer):

| | observer (never writes) | conflicts / run |
| --- | --- | --- |
| baseline | `{}` — blackout, 3/3 runs | **44–45** |
| with fix | **full 40** (`{alice:20, bob:20}`) | **1–2** |

So it cures the reader blackout **and collapses the conflict storm ~95%** — removing the seq-0 void reads removes the permanently-stale derived-cell commits that were the wedge. The residual failure is exactly **B3** below (the losing writer's own replica), unchanged, as expected.

**Notes for whoever takes the fix:**
- A narrower version (grace only when `hasAsCell`) mostly **didn't fire**: during element traversal the effective property schema frequently *lacks* the `asCell` annotation — it's dropped from the stored link's embedded / argument-side `$defs` bundle, so the reader descends *into* the linked profile as a plain `required` prop. That asCell-annotation loss may deserve its own look; the `required`-exemption above is what actually heals it.
- Test surface of the change: only the `traverse-replay` golden oracles shift — and instructively: `expected {ok:false, INVALID_OBJECT} got {ok:true}` across the **notebook / shopping-list / toolshed-reload** goldens, i.e. those real patterns were hitting the same void. Goldens would need regen. (One `generateObject` redaction test flaked on a timeout in the 10-min full run; passes isolated on both branches.)
- This does **not** address **B1** — with the fix, a reader gets a cell resolving into its *own* partition, so `authorProfile` content is still per-reader-wrong. B1 (capturing owner scope at link-mint) is the identity half.

### B3 — root-caused and pinned with a failing test

Added `packages/runner/test/array-append-client-convergence.test.ts` (deterministic, 60 ms, intentionally red). `array-push-mergeable.test.ts` asserts durability via a *fresh* session and passes; this covers the gap — the **client replica of the writer that interleaved its own append** never integrates the peer's append.

Seed `[seed]`; session 1 appends `A`; session 2 appends `B` without observing `A`; the server merges all three, but session 2's own replica, after explicit `sync + pull`, reads `[seed, B]` — `A` missing.

Two interacting sites in `storage/v2.ts`:
- **`confirmPending`** (~2625–2640): promotes the interleaving writer's `confirmed` to its own commit's (high) `applied.seq`, with a value replayed from its own base = own appends only.
- **`applySessionSync`** (~2387): `if (upsert.seq < record.confirmed.seq) continue;` (*"never move the confirmed base backwards; pending replay depends on monotonic bases"*) — so the peer's authoritative revision, at a lower seq, is skipped forever.

The guard is correct in isolation; the bug is that `confirmPending` advances `confirmed.seq` past unintegrated peer revisions while reconstructing a local-only value. Fix direction (not spiked): don't let `confirmPending` leave `confirmed` carrying only local appends at a seq that shadows the last integrated server seq.

bfollington:
## B3 — sharpened diagnosis (correcting the "confirmPending fix direction" above)

I spiked a fix and it revealed B3 is deeper than "confirmPending reconstructs wrong" — sharing what the instrumentation showed, since it changes the fix options.

Instrumenting `applySessionSync` during the B3 test's `pull()`:
- a **fresh** session (cursor 0) pulls the full authoritative `[seed, A, B]` at seq 3;
- session 1 (`integrate`, confirmed=2) correctly receives `[seed, A, B]`;
- **session 2 (the interleaving writer) produces no sync upsert at all** — its own B-commit advanced its server-side session cursor to head (seq 3), so delta-sync re-delivers nothing.

So session 2's local confirmed **value** (`[seed, B]`) and the server's authoritative **value** (`[seed, A, B]`) diverge **at the same seq (3)** — and a seq-based delta protocol structurally can't detect or repair that. Two contributing facts:
- `AppliedRevision` for a `patch` op (append) carries only `patches` (session 2's own append), **not** the merged document — only `set` ops carry `document` (`selectCommitRevisions`, memory/v2/engine.ts:4011). So `confirmPending` replays its own patch onto its own stale base and the merge never comes back through the commit response.
- I earlier implied the fix must be server-side; that's **not** quite right. The server's initial snapshot is keyed **per-watch-id, not per-session-cursor**: `watchAdd` (server.ts:1947-1967) returns empty for an existing watch id but takes the authoritative `trackGraph` snapshot path for a **new** watch id. So a client **fresh-watch resync** of the raced doc *does* yield `[seed, A, B]`.

Three viable fix shapes, all touching the core sync path (flagging for whoever owns it — the choice has real perf trade-offs):
1. **Client fresh-watch resync** of the doc after a mergeable-append commit — pure client, works, but +1 round-trip per append unless gated on detected concurrency.
2. **Server merge-signal** on the revision (e.g. `merged` / `rebasedAgainstSeq`) → client resyncs only flagged docs — targeted.
3. **Merged `document` in the applied revision** for merged commits → client adopts it directly — cleanest semantically, but materialize-per-commit + wire cost.

I didn't land any of these — the choice is a design call, not a drop-in. The failing test (`array-append-client-convergence.test.ts`) pins the behavior for whichever way it goes.

bfollington:
**Updated:** the B2 fix is now folded into this PR (not a separate branch), goldens regenerated, and the PR description rewritten as a decision-ready package (what's fixed / what's open / options). The description is authoritative — the two earlier comments above are kept for history but are superseded by it. Test story now: green = what the fix achieves (blackout cured, controls, observer converges under a 40-msg storm, LWW characterization); the two red tests are the pinned, still-open **B3**.

bfollington:
### Cross-ref: overlaps with #4405 (same machinery, adjacent failure mode)

**B3 here and #4405 (`Fix lunch poll contention vote swatches`) are in the same `v2.ts` pending-replay code and the same symptom family** — #4405's *"50 stored votes but 42 rendered swatches"* is the same class as B3's "durable has all, an interleaving writer's replica doesn't." #4405 also removes `pendingPatchLogger`/`changedPathsForPendingPatch`, so it's the productized line of that pending-patch work.

**But #4405 does not fix B3** — I ran this PR's deterministic B3 test (`array-append-client-convergence.test.ts`) against #4405's branch and it fails identically (writer reads `["B","seed"]`, missing `"A"`). They're adjacent:
- **#4405** fixes pending-replay **materialization** over stale/shape-mismatched bases (the `CloneForMutationError`; null roots, array spines, append markers) — *can the surviving patch apply at all*.
- **B3** is the peer's revision never being **integrated**: `confirmPending` rebuilds from a stale base **and** `applySessionSync`'s seq-monotonicity guard shadows the peer's lower-seq append. #4405's rewrite doesn't touch that path.

So B3's test is a case #4405's suite doesn't yet cover, and any B3 fix will want to coordinate with #4405 in the pending-replay region. cc @willkelly — you're deepest in this code. (B2 in this PR is disjoint — it's in `traverse.ts`.)

bfollington:
**Superseded — split into two PRs** (rebased onto main first; all findings re-verified, B3 still red on current main, i.e. not fixed by #4465):

- **#4532** — the all-green, safe-to-merge half: the B2 reader-blackout fix + data, goldens, the full green test/fixture/harness suite, LWW characterization, evidence appendix. Single squash commit, reverts cleanly.
- **#4533** — the review-required half (draft, stacked): the deliberately-red B3 pins + the three costed fix options. @robin @gideon the decision now lives there.

Closing this one; the investigation narrative in this PR body and comments remains the historical record.

PR #4463, opened 2026-07-02
test(ts-transformers): assert transformer output structurally; cover the zero-arg IIFE branch

Hixie:
## What

Two related changes to the `ts-transformers` test suite. No production code is
touched — this PR is test-only (0 files changed under `src/`).

### 1. Assert on structure, not printed text

Many transformer tests checked the compiler's output with
`assertStringIncludes(output, "...")` — a substring match against the printed
code or the printed schema. That form is unsound in two ways:

- The printer preserves authored comments, so a test can pass on the strength of
  a comment that happens to contain the expected substring rather than on the
  code the transformer actually emitted.
- When the expected substring already appears in the input source, a transform
  that did nothing at all would still satisfy the check.

This PR adds a shared test helper, `test/transformed-ast.ts`, that parses the
transformer's output back into a TypeScript AST (parsing discards comments as
trivia) and exposes focused queries: locating emitted calls by name or pattern,
finding immediately-invoked functions, detecting a `receiver.key("segment")`
reactive path read, reading the stable `.for(cause, true)` arguments, and
evaluating an emitted `... satisfies JSONSchema` object literal to a real
JavaScript value so a test can compare the generated schema by value.

Every genuinely weak printed-output check across the suite is rewritten to use
these queries and assert on real nodes and evaluated schema values. Assertions
that read a diagnostic, error, or analyzer reason message are deliberately left
as string checks: there the message text is the contract, and reading it off the
diagnostic object is not vulnerable to either failure mode above. Twenty-five
existing test files are converted.

### 2. Cover the zero-argument IIFE branch of the call-expression emitter

`emitCallExpression` has a dedicated branch for an authored zero-argument inline
function that is immediately invoked — `(() => ...)()` — appearing inside a
reactive array-method callback. That branch, and the helpers that scan it for
cell reads on receivers declared outside the function, previously ran only as a
side effect of patterns compiling through the transformer in the CI
pattern-integration jobs, never from a focused unit test.

Reaching it needs the in-place expression rewrite that array-method callback
lowering performs: a bare IIFE at a plain JSX site is skipped, and any
surrounding reactive expression is otherwise lifted whole before the emitter
recurses into the function. Placing the IIFE inside a `.map()` or `.filter()`
callback is what drives the emitter over it in place. A new test file,
`call-expression-iife-coverage.test.ts`, exercises both outcomes of the branch:
when the function body performs a direct cell read on an outer receiver, the
branch wraps it so the read runs under a reactive context and the cell
dependency is tracked into the callback's pattern input; when the function
decomposes through a local alias or an object-literal projection, the branch
leaves the authored function in place. This brings 78 of the branch's 95
previously integration-only lines under unit coverage; the remainder are
secondary sub-branches unreachable from source without contortion (local cell
aliases are inlined by the transformer, and the lift-applied fallback fires only
when the reactive-wrapper builder declines), documented in the test.

---
## Summary by cubic
Switch `ts-transformers` tests from substring checks to AST-based structural assertions, and add unit coverage for the zero‑arg IIFE branch in `emitCallExpression`. No production code is changed.

- **Refactors**
  - Added `packages/ts-transformers/test/transformed-ast.ts` with helpers to parse output, collect nodes, find named/matching calls, detect IIFEs and `.key(...)` reads, read stable `.for(cause, true)` args, evaluate `... satisfies JSONSchema` literals, and pull pattern/lift input/output schemas.
  - Converted tests across schema generation/injection/shrink‑validation, type shrinking/building/inference, pipeline regressions/core transform, guards, CF data/UI helpers, module-scope hardening/cf‑data, destructuring, array‑method utilities, binary rewrite, fetch‑json injection, lift‑applied call creation, pattern coverage transformer, and targeted validation cases to assert on real nodes and evaluated schema values. Diagnostic message checks remain string assertions.

- **New Features**
  - Added `packages/ts-transformers/test/call-expression-iife-coverage.test.ts` to unit‑cover the zero‑arg IIFE branch inside `.map`/`.filter()` callbacks, asserting:
    - Wrap when the IIFE body reads an outer cell directly (dependency tracked via `item.key("...")` and lowered under lift).
    - Preserve the authored IIFE when reads are via aliases or object projections.
  - Brings 78/95 lines of this branch under unit coverage.

<sup>Written for commit 96964eaccef1c8e701004b65901482b729438653. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4463?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ts-transformers/test/ast/type-building-coverage.test.ts
+++ b/packages/ts-transformers/test/ast/type-building-coverage.test.ts
@@ -1,9 +1,4 @@
-import {
-  assert,
-  assertEquals,
-  assertStringIncludes,
-  assertThrows,
-} from "@std/assert";
+import { assertEquals, assertStringIncludes, assertThrows } from "@std/assert";
 import ts from "typescript";
 
 import {
@@ -16,6 +11,35 @@ import {
   type CaptureTreeNode,
   createCaptureTreeNode,
 } from "../../src/utils/capture-tree.ts";
+import { collect, parseModule } from "../transformed-ast.ts";
+
+/**
+ * Reparse a printed capture-type literal and return its members keyed by name,
+ * each carrying its optional flag and the printed text of its declared type.
+ * Asserting on these AST-derived members is stronger than substring-matching
+ * the printed text: an optional token and a member type are pinned exactly.
+ */
+function typeLiteralProps(printed: string): Map<
+  string,
+  { optional: boolean; type: string }
+> {
+  const decl = collect(
+    parseModule(`type __T = ${printed};`),
+    ts.isTypeAliasDeclaration,
+  )[0];
+  if (!decl) throw new Error(`Could not parse type literal: ${printed}`);
+  const out = new Map<string, { optional: boolean; type: string }>();
+  for (const signature of collect(decl.type, ts.isPropertySignature)) {
+    const name = ts.isIdentifier(signature.name)
+      ? signature.name.text
+      : signature.name.getText(decl.getSourceFile());
+    out.set(name, {
+      optional: !!signature.questionToken,
+      type: signature.type ? signature.type.getText(decl.getSourceFile()) : "",
+    });
+  }
+  return out;
+}
 
 function createProgram(fileName: string, sourceText: string) {
   const sourceFile = ts.createSourceFile(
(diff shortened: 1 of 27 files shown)


mathpirate (approved):
lgtm thanks

PR #4467, opened 2026-07-02
feat(cfc): stamp model output LlmDerived at the store boundary (Epic D1)

seefeldb:
## What

Stage **D1** of [the CFC future-work implementation plan](https://github.com/commontoolsinc/labs/blob/main/docs/plans/cfc-future-work-implementation.md#5-epic-d--write-side--agent-integrity-the-live-soundness-cluster) — piece B of [`docs/specs/cfc-trusted-agent-tool-integrity.md`](https://github.com/commontoolsinc/labs/blob/main/docs/specs/cfc-trusted-agent-tool-integrity.md): model-produced bytes are stamped with an `LlmDerived` provenance atom at the point they enter the store, so "untrusted model output" is **explicit provenance** rather than mere absence of integrity. This is the enabler for the invoke-time tool-input `requiredIntegrity` gate (piece A, next stage D2).

- **`packages/api/cfc.ts`**: `CFC_ATOM_TYPE.LlmDerived` + atom type + `cfcAtom.llmDerived(model?)` mint helper.
- **`atom-classes.ts`**: propagation class `provenance` — never propagates through the default transition.
- **`prepare.ts`**: joins `RUNTIME_MINTED_INTEGRITY_ATOM_TYPES`, so the persist-time evidence gate strips it from any non-builtin author — pattern code can neither forge the stamp nor author schemas that mint it (tested).
- **`llm-dialog.ts`**: `pushModelMessages` stamps assistant + tool-result messages at their push sites; builtin-authored error literals and user messages (`addMessage`) stay unstamped. No-op when `cfcEnforcementMode` is `"disabled"`.

## Implementation note (the interesting seam)

The dialog's messages link carries its own schema, which wins over an `asSchema` handle inside `push()` (`resolvedLink.schema ?? this.schema`) — so the stamp cannot ride the array handle's item schema. Instead each model message (they all carry `[ID]` and split into their own entity docs) gets the stamping schema recorded explicitly via `recordRelevantSchemaWritePolicyInput` on its doc — the same split-entity idiom `data-updating.ts` uses — which also marks the continuation tx CFC-relevant so `prepareTxForCommit` runs the persist pass (the same seam the external-ingest stamp handles explicitly).

## Tests

`packages/runner/test/cfc-llm-derived-stamp.test.ts`:
- stamping mechanism kernel — the pushed element is stamped, a sibling pushed through the plain schema is not;
- `[ID]`-split element stamped on its own doc;
- **forge attempt stripped**: a non-builtin author pushing through the same stamping schema does not persist the atom (evidence-gate guard);
- **end-to-end mock dialog**: the assistant message's doc carries `LlmDerived`; the user message's doc does not.

Guard suites unperturbed (llm-dialog, llm-dialog-outbox, conversation fixtures, cfc mint-gate/flow/boundary — 65+122 steps green).

---
## Summary by cubic
Stamp all model-produced dialog bytes with a `LlmDerived` CFC atom at the store boundary so model output is explicit provenance. This enables the upcoming invoke-time tool-input `requiredIntegrity` gate (D2) and adds coverage for the disabled-dial no-op path.

- **New Features**
  - Added `CFC_ATOM_TYPE.LlmDerived`, `cfcAtom.llmDerived(model?)`, and registered it as a provenance class.
  - Gated as runtime-minted evidence; the persist-time gate strips `LlmDerived` from non-builtin authors.
  - `llm-dialog` stamps assistant and tool-result messages at push; user messages and builtin-authored error literals stay unstamped; no-op when `cfcEnforcementMode` is "disabled".
  - Records the stamping schema per split message entity using `recordRelevantSchemaWritePolicyInput`; tests cover both `cfcAtom.llmDerived` mint arms and the disabled-dial no-op branch.

- **Dependencies**
  - Regenerated `packages/static/assets/types/cfc.ts` to include `CFC_ATOM_TYPE.LlmDerived`, `CfcLlmDerivedAtom`, and `cfcAtom.llmDerived`.

<sup>Written for commit 0fc8f2a8798945879e3e018e48c633dd7e37a9a4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4467?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/api/cfc.ts
+++ b/packages/api/cfc.ts
@@ -35,6 +35,14 @@ export const CFC_ATOM_TYPE = {
   ExternalIngest: "https://commonfabric.org/cfc/atom/ExternalIngest",
   InjectionSafe: "https://commonfabric.org/cfc/atom/InjectionSafe",
   LinkReference: "https://commonfabric.org/cfc/atom/LinkReference",
+  // Runtime-minted LLM-derivation provenance: these bytes were produced by a
+  // model (assistant content, or a tool result entering the dialog
+  // transcript). Makes "untrusted model output" EXPLICIT provenance rather
+  // than mere absence of integrity, so requiredIntegrity floors fail
+  // positively on model-derived values (Epic D1,
+  // docs/specs/cfc-trusted-agent-tool-integrity.md piece B). Evidence — not
+  // authorable in schemas.
+  LlmDerived: "https://commonfabric.org/cfc/atom/LlmDerived",
   Origin: "https://commonfabric.org/cfc/atom/Origin",
   // Hereditary certification (spec §15.1.1 / §3.1.6.1): survives combination
   // via the class-aware meet — present on an output only when present on
@@ -112,6 +120,14 @@ export type CfcUserSurfaceInputAtom = CfcAtomObject & {
   readonly user: string;
   readonly surface: string;
   readonly valueDigest: string;
+};
+
+export type CfcLlmDerivedAtom = CfcAtomObject & {
+  readonly type: typeof CFC_ATOM_TYPE.LlmDerived;
+  // The model that produced the bytes, when known. Audit/display metadata —
+  // policies match on the atom type, so the default mint omits it to keep
+  // the persisted atom canonical across models.
+  readonly model?: string;
 };
 
 export type CfcExternalIngestAtom = CfcAtomObject & {
@@ -206,6 +222,12 @@ export const cfcAtom = {
     return {
       type: CFC_ATOM_TYPE.InjectionSafe,
     };
+  },
+
+  llmDerived(model?: string): CfcLlmDerivedAtom {
+    return model === undefined
+      ? { type: CFC_ATOM_TYPE.LlmDerived }
+      : { type: CFC_ATOM_TYPE.LlmDerived, model };
   },
 
   userSurfaceInput(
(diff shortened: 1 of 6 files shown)


seefeldb:
Two updates:

**CI failures** — fixed in 465a6da98: adding `CFC_ATOM_TYPE.LlmDerived` requires regenerating the checked-in `commonfabric/cfc` authoring module (`deno task gen-cfc-types`); the `Check` job and the `generateCfcTypes reproduces the checked-in module` test verify exactly that.

**Re the P1 (stamp the other LLM builtins):** partially agree, with one important correction to the bypass claim. The D2 invoke-time gate this enables checks `requiredIntegrity` **floors**, and floors fail on *absence* of the required atoms — an unstamped `generateObject`/`generateText` result carries none of `[UserSurfaceInput, PromptSlotBound, Builtin(agent-kernel), …]`, so it fails a floor-declaring field exactly like a stamped one. The stamp's job is (a) making model provenance *explicit* for policies/audit/UI and (b) closing the "value has no label object at all" corner in the dialog loop — not the floor's soundness, which never depended on it.

That said, the coverage point stands: `llm.ts`'s writebacks (`llm` content, `generateText` text, `generateObject` result/messages — three distinct write shapes: whole-array set, string result, author-schema'd object result) should also carry the stamp for provenance-explicitness. Doing those three shapes justice needs its own red-green pass (each has a different schema-recording seam than the dialog's `[ID]`-split push), so I'm keeping this PR at the plan's D1 scope (the dialog thread, where the D2 gate operates) and tracking the extension as an explicit follow-up (D1b) rather than growing this diff. Filed on the task list.

PR #4473, opened 2026-07-02
test(cfc): flat-assumption consumer sweep — clauses stay opaque + fail-safe (Epic A5)

seefeldb:
## What

Stage **A5** (final Epic A stage): audit every remaining `.confidentiality` consumer that iterates atoms as a flat list, and pin — via `cfc-clause-consumers.test.ts` — that each treats an OR-clause **opaquely** and stays correct or fail-safe (over-restrict), never under-restrict. Test-only: all consumers are already fail-safe, so no code change.

- **schema-sanitization**: `isPromptInjectionMaterialRiskAtom` returns false for a clause (no `type: Caveat`), so `filterMaterialRiskAtoms` **preserves** it (more restrictive); combined with A4 forbidding `Caveat` as an OR alternative, a material-risk caveat can never hide in a clause.
- **legacy classification walker** (`cfc.ts lubSchema`): carries a clause through the coarse `Set`-join as one opaque member — no crash, not flattened into its alternatives (which would under-represent the AND-of-the-clause), unioned with sibling flat atoms without merging.
- **html reconciler render fit**: treats a clause opaquely → over-restrict (fail-safe); clause-aware render fit is deferred to **H3b**.
- **sqlite `any()`**: stays reserved-and-erroring until **E1**.

The tests catch a future edit that flattens a clause. **This completes Epic A** — the CNF-clause label representation the rest of the plan builds on.

---
## Summary by cubic
Adds A5 tests to pin that flat `.confidentiality` consumers keep OR-clauses opaque and fail-safe, and fixes `schema-sanitization` to descend into `anyOf` so material‑risk caveats inside clauses are stripped correctly. Tests also cover `ContextualFlowControl.lubSchema`; this completes Epic A5.

<sup>Written for commit 498644c57d0c7989ee962ce9e72c2e0689a510d7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4473?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/cfc/schema-sanitization.ts
+++ b/packages/runner/src/cfc/schema-sanitization.ts
@@ -7,6 +7,7 @@ import {
 import { deepEqual } from "@commonfabric/utils/deep-equal";
 import { isRecord } from "@commonfabric/utils/types";
 import { uniqueCfcAtoms } from "./observation.ts";
+import { isOrClause, normalizeClause } from "./clause.ts";
 import { resolveCfcSchemaRefs } from "./schema-refs.ts";
 
 export const INJECTION_SAFE_ATOM = {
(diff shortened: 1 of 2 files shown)


seefeldb:
Valid P1 — fixed in the pushed commit. My test's rationale was wrong: a material-risk caveat hidden as an OR-clause alternative (`{anyOf:[risk, A]}`) is **not** more-restrictive when preserved, since a ceiling naming the sibling `A` subsumes the whole clause. `filterMaterialRiskAtoms` now **descends into OR-clauses** (`stripMaterialRiskFromClause`) — removing material-risk alternatives, unwrapping if one remains, dropping the clause if it was entirely material-risk. Test rewritten to assert the strip (sibling survives), the risk-only-clause drop, and that a benign clause is preserved while a top-level caveat is still stripped. Defense-in-depth for the future exchange-produced-clause case (Epic B); A4 already forbids authoring the caveat alternative.

PR #4474, opened 2026-07-02
feat(cfc): enforce tool-input requiredIntegrity at invoke time (Epic D2)

seefeldb:
## What

Stage **D2** of [the CFC plan](https://github.com/commontoolsinc/labs/blob/main/docs/plans/cfc-future-work-implementation.md#5-epic-d--write-side--agent-integrity-the-live-soundness-cluster) — [`cfc-trusted-agent-tool-integrity.md`](https://github.com/commontoolsinc/labs/blob/main/docs/specs/cfc-trusted-agent-tool-integrity.md) piece A/C. Closes the demo's central prompt-injection invariant.

`executeToolCalls` now validates each model-supplied input field against its `inputSchema`'s `ifc.requiredIntegrity` **before** the handler runs:
- `toolInputRequiredIntegrityFailure` walks the schema (properties, recursively) for floor-declaring fields.
- `toolInputValueIntegrity` resolves a value via `traverseAndCellify` — the same resolver the invoke path uses — so a by-reference `{"@link":…}` carries the referenced cell's stored integrity, while a plain model literal carries none (it is model output, stamped at most `LlmDerived` by D1, which lacks any endorsement family).
- A field whose floor the supplied value does not satisfy refuses the whole call with an error tool-result; the loop continues, the handler never runs.

Opt-in (only fires for tools declaring `requiredIntegrity` — no behavior change otherwise) and skipped when `cfcEnforcementMode` is `"disabled"`.

**Result:** an injected `sendMail` recipient (a string copied from a hostile briefing) is refused instead of sent.

## Tests

`cfc-agent-tool-input-integrity.test.ts` (un-ignored + expanded, shared harness): the injected recipient is refused under `enforce`; a tool **without** a floor is not over-blocked; the gate is a no-op when enforcement is disabled. llm-dialog guard suites (64 tests) green.

---
## Summary by cubic
Enforces tool-input `requiredIntegrity` at invoke time to block prompt-injection on control fields. Closes the `invoke` bypass, keeps observe mode non-blocking, enforces floors under array items and compound schemas, and avoids over-blocking on optional fields.

- **New Features**
  - `executeToolCalls` validates model input against `inputSchema.ifc.requiredIntegrity` before handlers run.
  - Resolves values via `traverseAndCellify` so `{"@link":…}` carries cell integrity; literals carry none.
  - On failure in enforcing modes, pushes an error tool-result and skips the handler; loop continues.
  - Opt-in: only tools with `requiredIntegrity` are gated; skipped when `cfcEnforcementMode` is `"disabled"` or `"observe"`.
  - Adds `toolInputRequiredIntegrityFailure` and `toolInputValueIntegrity`; matches integrity atoms with `deepEqual`.

- **Bug Fixes**
  - Closed `invoke` bypass by checking the resolved handler/pattern argument schema (not the builtin’s path/args).
  - Walker enforces floors under `properties`, `items`, and `anyOf`/`oneOf`/`allOf` branches.
  - Gate only fields the model actually supplied; optional floor-fields omitted by the model don’t block (missing required fields are handled by normal input validation).

<sup>Written for commit f523e64ca329377815b192b57853f1569eea0a19. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4474?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/builtins/llm-dialog.ts
+++ b/packages/runner/src/builtins/llm-dialog.ts
@@ -2035,12 +2035,19 @@ function toolInputRequiredIntegrityFailure(
   }
   if (isRecord(schema.properties)) {
     for (const [key, childSchema] of Object.entries(schema.properties)) {
-      const childValue = isRecord(value) ? value[key] : undefined;
+      // Only gate fields the model actually supplied. An absent (e.g. optional)
+      // field carries no value to gate; treating it as `undefined` would fail
+      // an optional field's floor and over-block the call. A required field the
+      // model omitted is a structural error handled by ordinary input
+      // validation, not a floor bypass — there is no injected value to gate.
+      if (!isRecord(value) || !Object.hasOwn(value, key)) {
+        continue;
+      }
       const failure = toolInputRequiredIntegrityFailure(
         runtime,
         space,
         childSchema,
-        childValue,
+        value[key],
         path ? `${path}.${key}` : key,
       );
       if (failure !== undefined) {
--- a/packages/runner/src/builtins/llm-dialog.ts
+++ b/packages/runner/src/builtins/llm-dialog.ts
@@ -28,6 +28,7 @@ import {
 } from "./llm-schemas.ts";
 import { getLogger } from "@commonfabric/utils/logger";
 import { isBoolean, isObject, isRecord } from "@commonfabric/utils/types";
+import { deepEqual } from "@commonfabric/utils/deep-equal";
 
 // Message schema that mints the `LlmDerived` provenance stamp (Epic D1).
 // Recorded as the schema write-policy input for each model-produced message's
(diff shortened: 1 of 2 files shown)


seefeldb:
All three findings valid and fixed in the pushed commit:

1. **P1 invoke bypass** — the gate now runs **after** `resolveToolCall` and, for a `invoke` call, checks the **resolved** `pattern.argumentSchema` / `handler.schema` against the stripped `args` (`integrityGateTarget`), reusing the same target-schema resolution the catalog builder already uses (`pattern?.argumentSchema ?? handler?.schema`). New end-to-end test: a literal recipient routed through the generic `invoke` builtin to a floor-declaring handler is refused.
2. **P2 observe mode** — denial is now gated on `cfcEnforcementStrictness(mode) >= CFC_ENFORCING_STRICTNESS`, so only `enforce-explicit`/`enforce-strict` block; `observe` (diagnostic) and `disabled` do not. Test added.
3. **P2 items/compound** — the walker descends into `items` (array elements) and `anyOf`/`oneOf`/`allOf` (all branches — the fail-safe over-require direction, mirroring `walkIfcSchema`). Test: an array-`items` floor gates each element.

seefeldb:
Fixed the cubic P2 in the pushed commit: the walker now gates only fields the model actually supplied (`Object.hasOwn`), so an omitted optional floor-field no longer fails as `undefined` and over-blocks — while a supplied literal still fails its floor (test added). The re-posted Codex "descend into items/compound" comment was already addressed in the prior fix commit (the `items` + `anyOf`/`oneOf`/`allOf` descent is present at llm-dialog.ts ~2053/2070).

seefeldb:
Merging: the re-posted items/compound comment is a stale re-anchor to the `properties` loop — the `items` + `anyOf`/`oneOf`/`allOf` descent is present (llm-dialog.ts ~2053/2070) and verified by the passing "descends into array-items floors" test.

PR #4475, opened 2026-07-02
feat(cfc): un-reserve sqlite any() → authored OR-clauses (Epic E1)

seefeldb:
## What

Stage **E1** of [the CFC plan](https://github.com/commontoolsinc/labs/blob/main/docs/plans/cfc-future-work-implementation.md#6-epic-e--row-set-reads-per-row-labels--sqlite-3b3c): now that the runner ships the clause-aware label profile (Epic A), the sqlite row-label evaluator **emits OR-clauses** instead of rejecting `any()`.

- `validateConfTerm`/`validateConfExpr` validate an `any()` node's alternatives (structural well-formedness — atom-shape restrictions like principal-like-only / no `Caveat`/`Expires` are enforced runner-side at the boundary per §3.1.8).
- `evalConf` emits one `{anyOf:[…]}` clause whose alternatives are the union of what each term evaluates to, kept structural and never flattened into bare atoms. `all(any(A,B), C)` → `[{anyOf:[A,B]}, C]` = (A∨B)∧C.

This delivers the **"any participant may read"** per-row semantics the flat conjunctive form could not express (the email OR-set use case).

## Tests

- memory `v2-sqlite-row-label-test`: `any()` builds+validates an OR-clause; a well-formed wire `anyOf` validates while empty/malformed refuse; `evalConf` produces `{anyOf:[…]}`; `all`-of-`any` is proper CNF.
- runner `sqlite-row-label-read`: a well-formed `anyOf` wire spec is accepted, an empty one refuses.

Memory (35) + runner sqlite (49 steps) suites green.

---
## Summary by cubic
Adds authored OR-clauses for sqlite row labels: `any()` now validates and evaluates to a single `{anyOf:[…]}` clause, enabling “any participant may read” and proper CNF like `all(any(A,B), C)`. Conjunctive or nested-OR `any()` alternatives (including under `when()`) are rejected and also fail closed at eval to prevent widening `(A∧B)∨C`.

- **New Features**
  - `any()` is accepted and validated; wire `anyOf` is allowed when well-formed and rejected when empty/malformed. Conjunctive or nested `any()` alternatives are refused, including under `when()`.
  - `evalConf` returns one structural OR-clause; never flattened to atoms. Defense in depth: rejects conjunctive or nested-OR alternatives at evaluation too, even if validation was skipped.

<sup>Written for commit d60565820c6cf5caaedad7523c0da1499ef2cc30. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4475?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/memory/test/v2-sqlite-row-label-test.ts
+++ b/packages/memory/test/v2-sqlite-row-label-test.ts
@@ -154,6 +154,44 @@ Deno.test("any() builds and validates an OR-clause (Epic E1)", () => {
   );
 });
 
+Deno.test("any() rejects a conjunctive alternative — no (A∧B)∨C → A∨B∨C widening (Epic E1)", () => {
+  // A direct all() as an any() alternative is rejected: union-flattening it
+  // would make the row readable by A alone even though the author required
+  // A AND B.
+  assertThrows(
+    () =>
+      table(EMAIL_COLUMNS, (f) => ({
+        confidentiality: any(
+          all(
+            principal("mailto", match(f.from, ADDR)),
+            principal("mailto", match(f.to, ADDR)),
+          ),
+          dbOwner(),
+        ),
+      })),
+    Error,
+    "all()/any()",
+  );
+  // ...and a when() gating an all() is rejected too (recursive check).
+  assertThrows(
+    () =>
+      table(EMAIL_COLUMNS, (f) => ({
+        confidentiality: any(
+          whenMatches(
+            f.auth,
+            /dmarc=pass/,
+            all(
+              principal("mailto", match(f.from, ADDR)),
+              principal("mailto", match(f.to, ADDR)),
+            ),
+          ),
+          dbOwner(),
+        ),
+      })),
+    Error,
+  );
+});
+
 Deno.test("a rule referencing an unknown column throws", () => {
   assertThrows(
     () =>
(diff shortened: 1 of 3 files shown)


seefeldb:
Valid P1 (soundness) — fixed in the pushed commit. `any(all(A,B), C)` union-flattened to `A∨B∨C`, widening the author's `(A∧B)∨C` so a row became readable by A alone. `validateAnyOfAlternative` now rejects `all()`/`any()` as an `any()` alternative (recursing through `when()` so `whenMatches(…, all(A,B))` is caught too), and `evalConf` fails closed on the same shape as defense-in-depth. `principal(match)` multi-match stays valid — it's a disjunction of independent readers, which flattens correctly. Tests added for both the direct and when-gated cases.

seefeldb:
Fixed in the pushed commit: `anyOfAlternativeHasConjunction` now recurses through `when.then`, mirroring the validator, so `evalConf` fails closed on a `when()`-wrapped conjunction that bypassed validation — not just a direct `allOf`/`anyOf`. Added an eval-level test (raw wire spec, validation skipped) covering both the direct and when-gated shapes, which also covers the coverage-debt the perf check flagged.

PR #4486, opened 2026-07-02
Treat imported cell wrapper types as a capability contract

Hixie:
Capability analysis shrinks a handler's declared input schema to the minimum capability its body uses, emitting an asCell marker that drives the field's write-authority set: per spec 8.15 a field's writeAuthorizedBy is composed from the handlers that declare they write it, and a runtime write is allowed only when the handler identity is in that set. Body analysis can see reads and direct writes, but not what a callee does inside a body that lives in another file. So a handler that receives a Writable auth cell and passes it to an imported provider client that persists a refreshed OAuth token — writing the cell inside the client — had its auth input shrunk to read-only, dropping the handler from writeAuthorizedBy, and the refresh write was denied at runtime.

Use the callee's declared Common Fabric wrapper parameter type as the capability contract at the import boundary. A recognized wrapper maps to a capability: Writable to read and write, ReadonlyCell to read, WriteonlyCell to write, and the opaque/stream/comparable/sqlite wrappers to their handling. Writable is a structural alias of Cell, so the wrapper is matched nominally by name on the declared type node, not the resolved type. A recognized wrapper accounts for its argument and replaces the conservative whole-cell fallback for that argument and call, while other evidence in the handler (such as a direct update) still contributes; bare Cell and same-named foreign types are not recognized and stay conservative, as do unknown callees.

This is the least-authority behavior the runtime enforces. Handler code is untrusted (threat model 9.1.1, 9.2.2), so preserving write authority a handler does not demonstrate would broaden writeAuthorizedBy and enable escalation; a cast that writes past a declared read-only type is denied at runtime because the schema, not the cast, is enforced. Trusting the declared type to reduce authority is therefore safe, and trusting it to grant write is bounded by the caller's own declaration and is what real refresh clients need.

Details:

- Split GmailSendClient's factory into separate Writable<Auth> and AuthCell overloads so overload resolution surfaces the writable form for a live cell; a union parameter is a mixed type node the name-based detector cannot read. Declare GoogleDocsClient's constructor auth parameter Writable<Auth> so its token-refresh write is explicit.
- Classify a bounded generic parameter by its constraint. Recognize nullable wrappers (Writable<T> | null), the readonly-array rest spelling (...args: readonly Writable<T>[]) alongside T[]/Array<T>/ReadonlyArray<T>, and map spreads into rest parameters while leaving array spreads into fixed parameters conservative.
- Suppress the passing-read for both destructured and member-access arguments so a WriteonlyCell argument stays write-only.
- Raise a capability:unreadable-cell-argument error when a branded cell argument flows to a union parameter that mixes a recognized wrapper with an unresolvable member (AuthCell | Writable<Auth>) — the overload-split case, enforced until now only by convention. The check is deliberately narrow: it does not fire for value-or-cell framework types (FactoryInput, CellLike, U | AnyBrandedCell<U>), bare Cell, generics, lone structural interfaces, array identity-writer methods, or any/unknown sinks. A transform of the full pattern corpus produced zero matches outside the overload-split shape.

The read/write-violation direction is already enforced elsewhere: reading a WriteonlyCell or writing a ReadonlyCell is a compile error via the branded interfaces, and runtime write authority is enforced in the runner.

---
## Summary by cubic
Treat imported Common Fabric cell wrapper types on functions and constructors as the capability contract so handlers keep correct read/write authority across file boundaries. Adds a diagnostic for ambiguous cell-union parameters and updates Google client types to declare writable auth explicitly.

- **New Features**
  - Capability analysis reads callee parameter wrappers: `Writable<T>` → read+write, `ReadonlyCell<T>` → read, `WriteonlyCell<T>` → write; also handles opaque/stream/comparable/sqlite.
  - Works through generic constraints, nullable unions, and rest params (including `readonly` arrays); spreads into rest are mapped, array spreads into fixed params stay conservative.
  - Suppresses the passing-read for destructured and member-access args so `WriteonlyCell` stays write-only.
  - Emits `capability:unreadable-cell-argument` for mixed wrapper unions (e.g., `AuthCell | Writable<Auth>`).
  - `GmailSendClient` factory split into `Writable<Auth>` and `AuthCell` overloads; `GoogleDocsClient` constructor now takes `Writable<Auth>`; uses canonical `CellBrand` for capability mapping.

- **Migration**
  - If a callee writes a cell, type the parameter as `Writable<T>` or `WriteonlyCell<T>`; readers should use `ReadonlyCell<T>`.
  - Replace unions that mix wrappers with overloads (fixes the new error).
  - Do not rely on bare `Cell<T>` or non-`commonfabric` `Writable` to grant write; they remain conservative.

<sup>Written for commit 1db55054b6e902aaeed6991f97748cbdd58cc07b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4486?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ts-transformers/src/policy/capability-analysis.ts
+++ b/packages/ts-transformers/src/policy/capability-analysis.ts
@@ -1,6 +1,7 @@
 import ts from "typescript";
 import {
   classifyArrayCallbackContainerCall,
+  getNodeText,
   isCellLikeType,
   isWildcardTraversalCall,
 } from "../ast/mod.ts";
@@ -9,7 +10,9 @@ import {
   type FunctionCapabilitySummary,
   type ReactiveCapability,
   resolvesToCommonFabricSymbol,
+  type UnreadableCellArgument,
 } from "../core/mod.ts";
+import { isBrandedCellType } from "../transformers/cell-type.ts";
 import { getKnownComputedKeyPathSegment } from "../utils/reactive-keys.ts";
 import { decodePath, encodePath } from "../utils/path-serialization.ts";
 import { unwrapExpression } from "../utils/expression.ts";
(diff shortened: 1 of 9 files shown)


packages/ts-transformers/src/policy/capability-analysis.ts
```diff
@@ -221,6 +224,245 @@ function isInterproceduralSummaryTarget(
...
+  }
+  const lastParameter = parameters[parameters.length - 1];
+  const declaration = lastParameter?.valueDeclaration;
+  return declaration && ts.isParameter(declaration) &&
+      declaration.dotDotDotToken
+    ? lastParameter
+    : undefined;
+}
+
+type SignatureCellKind =
```
mathpirate: i think an enum of this kind should already exist somewhere? maybe ask an agent to check whether this one is distinct/unique from the others

Hixie: Good catch, thanks. Fixed.

PR #4492, opened 2026-07-02
feat(runner): Runtime construction presets + canonical experimental-env mapping (CT-1814)

mathpirate:
## Summary

Implements **CT-1814** — the second axis of the CT-1811 divergence class. PR #4454 sealed the pattern **load path** (the harness could evaluate without registering artifacts); this PR seals **runtime construction config**: 13 first-party sites hand-rolled subsets of `RuntimeOptions`, so a new option — or a changed constructor default — could land unevenly and make the harness silently behave differently from production. That axis is growing under us: `RuntimeOptions` is 21 options and gained two CFC dials this week alone (#4479, #4488), which no construction site sets.

## What this adds

**`packages/runner/src/runtime-presets.ts`** — first-party construction presets. Three gates, all in one reviewed file:

1. **`RUNTIME_OPTION_KEYS`** — a type-gated exhaustive registry of `keyof RuntimeOptions`. A new option that is not classified here is a **compile error** that names the missing key. Adding an option now forces a fleet-wide decision (core-pinned / core-default / declared delta / per-site), instead of an incidental per-site choice.
2. **`EXPERIMENTAL_ENV_VARS` + `experimentalOptionsFromEnv(env)`** — the canonical (and now only Deno-side) env mapping for `ExperimentalOptions`, type-gated against `keyof ExperimentalOptions`. It replaces three parallel copies (toolshed `env.ts`, background-piece-service `env.ts`, cli `utils.ts`) whose parsers disagreed in opposite directions on non-canonical values. A deliberately-unwired flag is declared `null` in the mapping (today: `commitPreconditions`, from #4090) instead of being silently absent from one wiring.
3. **Six presets** — `productionServer`, `remoteClient`, `patternTest`, `localDev`, `browserWorker`, `unitTest` — all composed from one internal `coreOptions`. First-party rollout dials (the CFC modes) get flipped there, once, for every environment at once; constructor defaults then govern only non-preset constructions. Deliberate deltas (mock fetch, collectors, byte caches, host-decided dials) are explicit, documented preset parameters — visible at each call site, per the ticket's "don't over-centralize" constraint. Sites keep the `new Runtime(...)` expression so construction stays greppable.

**Conformance test** (`packages/runner/test/runtime-presets.test.ts`):
- a `Record<RuntimeOptionKey, Treatment>` spec — type-forced to cover every registered key — asserting each key's minimal-args treatment in every preset (per-site sentinel / core-pinned value / pinned-in-family / absent);
- full-args goldens proving every declared delta param routes to exactly its `RuntimeOptions` key (a param cannot be dropped or mis-mapped);
- canonical-parse tests for `experimentalOptionsFromEnv` (including the garbage-value warning) and a check that it consults exactly the env-wired mapping entries;
- a constructor smoke test (preset output builds a working `Runtime` with the pinned CFC mode).

## Migrated sites (13)

| Site | Preset |
| -- | -- |
| `toolshed/index.ts` | `productionServer` (apiUrl=MEMORY_URL, patternApiUrl=API_URL) |
| `background-piece-service/src/main.ts` | `productionServer` |
| `background-piece-service/src/worker.ts` | `productionServer` (experimental arrives as IPC data) |
| `background-piece-service/cast-admin.ts` | `remoteClient` |
| `piece/src/ops/pieces-controller.ts` | `remoteClient` (trust provenance stays a visible delta) |
| `runtime-client/backends/runtime-processor.ts` | `browserWorker` (host `InitializationData` → params mapper) |
| `cli/lib/test-runner.ts` | `patternTest` |
| `cli/lib/multi-user-test-worker.ts` | `patternTest` |
| `generated-patterns/integration/pattern-harness.ts` | `patternTest` |
| `cli/lib/dev.ts` | `localDev` |
| `cli/lib/acl.ts` | `remoteClient` |
| `cli/lib/piece.ts` | `remoteClient` |
| `html/bench/worker-reconciler-mount.bench.ts` | `unitTest` |

`runtimeOptionsFromInitializationData` is reshaped into `browserWorkerParamsFromInitializationData` (data → preset params), so the processor's whole options object goes through the seal, callbacks included.

## Behavior changes (each deliberate, each toward uniformity)

1. **Non-canonical `EXPERIMENTAL_*` values are now ignored with a warning, uniformly.** Previously toolshed/bg-service coerced anything ≠ `"false"` to **true** while the CLI coerced anything ≠ `"true"` to **false** — so `EXPERIMENTAL_MODERN_CELL_REP=1` enabled the flag on toolshed and disabled it under `cf test`. Canonical `"true"`/`"false"`/unset behave exactly as before everywhere.
2. **The multi-user test worker and the generated-patterns harness now honor `EXPERIMENTAL_*`** like the single-user test-runner. Previously they silently ignored it, so the harness modes could run under different flags than the code under test (the CT-1811 flavor of drift, inside the harness family itself).
3. **cast-admin and `PiecesController` now honor `EXPERIMENTAL_*`** (parity with their environment class; `PiecesController` reads env only under Deno, mirroring its existing `PIECE_TRACE_TIMINGS` guard).
4. **Deployment-facing runtimes now pin `patternEnvironment` to their API base.** Previously `cf acl` / `cf piece` / cast-admin / `PiecesController` / bg-service main left it unset, so a pattern-relative fetch resolved against the builder's hardcoded `localhost:<toolshed-port>` fallback even when operating against a remote deployment (`setPatternEnvironment` is process-global; the builder default only fit local dev). bg-worker and toolshed already pinned it; test/dev presets deliberately keep the localhost fall-through.
5. **`EXPERIMENTAL_*` entries removed from toolshed and bg-service env schemas** — the construction sites read the canonical mapping directly (with a pointer comment left in each env.ts). bg-service's `createRuntime` takes an injectable `EnvReader` (mirroring its `loadEnv(source)` DI idiom), and its env tests moved accordingly.

**Non-changes:** the core `cfcEnforcementMode` pin (`"enforce-explicit"`) equals today's constructor default — the pin exists so a future default change cannot silently relax one environment. All other CFC dials remain constructor defaults (off), now flippable in one place. Test presets keep `patternEnvironment` on the builder default (local fall-through for unmocked fetches), matching today.

## Observations recorded, deliberately out of scope

- `commitPreconditions` (and `patternEnvironment`, and `modernCellRep`/`persistentSchedulerState`) propagate through **module-level globals** (`memory/v2`, `builder/env.ts`), so the *last constructed runtime in a process* wins. Pre-existing; the presets reduce practical exposure but the globals deserve their own look.
- `DEFAULT_CFC_ENFORCEMENT_MODE = "disabled"` (cfc/types.ts) names the *transaction-layer* default, which the runtime always overrides — a naming trap next to the actual runtime default (`"enforce-explicit"`).
- The shell keeps a **fourth** copy of the experimental mapping as build-time `$`-defines (`shell/src/lib/env.ts`, garbage → false). It feeds `browserWorker` as host data by design; unifying its parse/names onto `EXPERIMENTAL_ENV_VARS` is a small follow-up.
- `runtime-client/protocol/types.ts` declares its own inline copy of the experimental-flags shape (wire type; defensible, noted).
- **282 test files** still construct `Runtime` by hand; the `unitTest` preset exists so they can adopt incrementally (the three runner fixtures included). A phased lint rule (`new Runtime(` outside presets) remains an option; deliberately not in this PR.

## Verification

| Gate | Result |
| -- | -- |
| Conformance test (13 steps) | ✅ |
| fmt · lint | ✅ |
| runner full suite | 754 passed / 0 failed |
| cli full suite | 912 passed / 0 failed |
| background-piece-service suite | 12 passed / 0 failed |
| runtime-client suite | 29 passed / 0 failed |
| piece suite | 10 passed / 0 failed |
| toolshed suite | 57 passed / 0 failed |
| generated-patterns integration (via migrated harness) | 145 passed / 0 failed |
| pattern-tests sweep (`cf test packages/patterns`, incl. multi-user + CT-1811 regression) | 616 passed; 34 failures **byte-identical on clean main** (pre-existing `cf test <dir>` sibling-import resolution artifact — verified via a clean `origin/main` worktree running the same command; failure sets diff-identical) |

---
## Summary by cubic
Adds first‑party runtime construction presets and a canonical `EXPERIMENTAL_*` env mapping in `@commonfabric/runner` to remove config drift across sites (CT‑1814). Migrates all first‑party construction sites and adds conformance tests that pin shared posture and flag wiring.

- **New Features**
  - `@commonfabric/runner/runtime-presets.ts`: exhaustive `RUNTIME_OPTION_KEYS`; canonical `EXPERIMENTAL_ENV_VARS` + `experimentalOptionsFromEnv(env)`; six presets — `productionServer`, `remoteClient`, `patternTest`, `localDev`, `browserWorker`, `unitTest` — exported via `@commonfabric/runner` index; conformance tests cover per‑key treatment and delta routing.
  - Runtime‑client: `browserWorkerParamsFromInitializationData` feeds `runtimePresets.browserWorker`, threading `spaceHostMap` and pinning `patternEnvironment` to the host API.
  - Toolshed: options assembly extracted to `toolshedRuntimeOptions` (splits `MEMORY_URL`/`API_URL`, uses canonical flag mapping) with a unit test.
  - CLI/harness: `cf test` and the multi‑user worker use `runtimePresets.patternTest` (shared CFC pin, shared `moduleByteCache`, env‑honored flags); `cf dev` uses `localDev`; `cf acl`/`cf piece` use `remoteClient`. Bench uses `unitTest`.

- **Bug Fixes**
  - One strict, shared flag parser: only `"true"`/`"false"`/unset; garbage is ignored with a warning. Redundant parsers removed from toolshed, CLI, and background‑piece‑service.
  - Cast‑admin reads `EXPERIMENTAL_*` via its injected env reader; test pins the injected reader is consulted. Background‑piece‑service main reads flags via an injectable `EnvReader` and forwards them to workers; both use `productionServer`.
  - Harness parity: the multi‑user worker and generated‑patterns harness honor `EXPERIMENTAL_*` and use `runtimePresets.patternTest`; deployment‑facing clients pin `patternEnvironment` to their API base.

<sup>Written for commit 9b24c62f8ff73e4f9ea429af968c5dd3850d77d0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4492?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/background-piece-service/cast-admin.ts
+++ b/packages/background-piece-service/cast-admin.ts
@@ -1,6 +1,11 @@
 import { parseArgs } from "@std/cli/parse-args";
 import { PieceManager } from "@commonfabric/piece";
-import { compileAndSavePattern, Runtime } from "@commonfabric/runner";
+import {
+  compileAndSavePattern,
+  experimentalOptionsFromEnv,
+  Runtime,
+  runtimePresets,
+} from "@commonfabric/runner";
 import { StorageManager } from "@commonfabric/runner/storage/cache.deno";
 import { type DID } from "@commonfabric/identity";
 import { createSession } from "@commonfabric/identity";
@@ -43,13 +48,17 @@ export function createRuntime(
   toolshedUrl: string,
   identity: Identity,
 ): Runtime {
-  return new Runtime({
+  // Shared first-party posture for client runtimes against a deployed API
+  // (CT-1814); this admin CLI now honors EXPERIMENTAL_* like the rest of the
+  // service instead of silently ignoring it.
+  return new Runtime(runtimePresets.remoteClient({
     apiUrl: new URL(toolshedUrl),
     storageManager: StorageManager.open({
       as: identity,
       memoryHost: new URL(toolshedUrl),
     }),
-  });
+    experimental: experimentalOptionsFromEnv(Deno.env.get),
+  }));
 }
 
 export function requireCellCause(cause: string | undefined): string {
--- a/packages/background-piece-service/cast-admin.ts
+++ b/packages/background-piece-service/cast-admin.ts
@@ -21,7 +21,11 @@ export interface CastAdminDependencies {
   args: string[];
   envGet: typeof Deno.env.get;
   getIdentity: typeof getIdentity;
-  createRuntime: (toolshedUrl: string, identity: Identity) => Runtime;
+  createRuntime: (
+    toolshedUrl: string,
+    identity: Identity,
+    envGet: typeof Deno.env.get,
+  ) => Runtime;
   readTextFile: typeof Deno.readTextFile;
   createSession: typeof createSession;
   createPieceManager: (
@@ -47,17 +51,19 @@ export interface CastAdminDependencies {
 export function createRuntime(
   toolshedUrl: string,
   identity: Identity,
+  envGet: typeof Deno.env.get = Deno.env.get,
 ): Runtime {
   // Shared first-party posture for client runtimes against a deployed API
   // (CT-1814); this admin CLI now honors EXPERIMENTAL_* like the rest of the
-  // service instead of silently ignoring it.
+  // service instead of silently ignoring it, read through the same injected
+  // env boundary as every other env consultation here (CastAdminDependencies).
   return new Runtime(runtimePresets.remoteClient({
     apiUrl: new URL(toolshedUrl),
     storageManager: StorageManager.open({
       as: identity,
       memoryHost: new URL(toolshedUrl),
     }),
-    experimental: experimentalOptionsFromEnv(Deno.env.get),
+    experimental: experimentalOptionsFromEnv(envGet),
   }));
 }
 
@@ -87,7 +93,11 @@ export async function castPattern(
     quit,
   });
 
-  const runtime = dependencies.createRuntime(toolshedUrl, identity);
+  const runtime = dependencies.createRuntime(
+    toolshedUrl,
+    identity,
+    dependencies.envGet,
+  );
 
   try {
     // Load and compile the pattern first
--- a/packages/background-piece-service/src/env.ts
+++ b/packages/background-piece-service/src/env.ts
@@ -1,21 +1,5 @@
 import { z } from "zod";
 
-/**
- * Results in `true` (on), `false` (off), or `undefined` (default).
- */
-function flagValue() {
-  return z.string().default("default").transform((v) => {
-    switch (v) {
-      case "default":
-        return undefined;
-      case "false":
-        return false;
-      default:
-        return true;
-    }
-  });
-}
-
 const envSchema = z.object({
   // Job queue settings
   //MAX_CONCURRENT_JOBS: z.coerce.number().positive().default(5),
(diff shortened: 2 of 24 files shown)


Hixie:
https://github.com/commontoolsinc/labs/pull/4598 may be relevant to your interests (I'll update it after this merges).

PR #4494, opened 2026-07-02
feat(runner): reject $generated in user-supplied causes (build-time throw)

mathpirate:
## What

Reject user-supplied cell causes carrying a top-level `$generated` record key —
throw at `.for()` time with a clear "reserved key" message. Follow-up to #4447,
proposed by @ubik2 in review (CT-1822).

A user cause like `Cell.for({ $generated: 0 })` deliberately mimics the
anonymous-cause namespace the pattern builder mints for itself
(`{ $generated: N }`, `{ $generated: N, $kind: "stream" }`, and the
`{ name, $generated: N }` duplicate-name wrap). The O(1) dedup shipped in #4447
handles the mimicry silently, but the "deceptively named cause" confusion class
stays alive. Rejecting `$generated` at intake deletes the class at the root.

### The contract (scope per review discussion)

- **Reserved: exactly `$generated`.** Every builder-minted cause carries it, so
  a cause without it can never hash-equal one — that alone restores
  collision-freedom. Other sigil-looking keys (`$kind`, `/Map@1`-style magic
  values, etc.) stay legal.
- **Top-level only:** causes collide as whole records under the canonical value
  hash, and every builder-minted cause is a flat record — a `$generated` nested
  deeper can never equal one, so deep-checking would tax every `.for()` for no
  safety.
- **Enforced at the single funnel:** `Cell.for()` (cell.ts) is the only writer
  of `_causeContainer.cause`, which is exactly what `factoryFromPattern` reads
  back as `export().name`. Throwing there gives the error at the user's own
  `.for()` call site and also covers runtime causes feeding `createRef`. The
  assignment site re-asserts the invariant it relies on (defense against a
  future second writer).
- **`allowIfSet` suggestion path stays a no-op:** a reserved-key cause that is
  ignored (cause already set) never takes effect, so it doesn't throw — pinned
  by test.

### Error surfacing (sanity-checked per review ask)

`cf check` on a pattern using `new Writable(0).for({ $generated: 0 })` prints
the message + help line verbatim, with the stack pointing at the user's own
`.for()` call (`<pattern>.tsx:5:45`), and exits 1:

```
Error: Cannot use cause {"$generated":0}: top-level key "$generated" is reserved
help: "$generated" marks system-generated cell causes; rename the key
  at assertNoReservedCauseKeys (packages/runner/src/builder/pattern.ts:223:11)
  at CellImpl.for (packages/runner/src/cell.ts:780:5)
    at exports.default.type (/fid1:…/pattern.tsx:5:45)
```

### The payoff (why the ticket exists)

With user causes unable to carry `$generated`, every builder-minted cause is
collision-free **by construction** (per-build-unique counter, unforgeable
namespace). The partial-cause assignment loop loses its collision handling for
generated causes entirely:

- anonymous causes: assigned directly, no check, no set entry;
- named causes: one `Set` lookup; on duplicate, wrap once with a fresh counter —
  no re-check loop;
- only named causes enter the dedup set.

Cause sequences are bit-identical to the old loop for all now-legal inputs
(the counter burns in the same order), so no cell identity moves — confirmed by
the existing partial-cause pins and the pattern-tests/generated-patterns
integration suites.

## Semantic change (why not folded into #4447)

Previously-valid builds now throw: any pattern calling `.for()` with a record
cause carrying a top-level `$generated`. A repo-wide sweep (all
`$generated`/`$kind` occurrences + classification of all 274 `.for()` call
sites, including every transformer-emitted cause shape) found **zero** such
uses — the only `$generated` occurrences are the implementation and test
assertions on generated output.

## Tests

- Rejection matrix on `Cell.for`: `{$generated}` and `{$kind, $generated}`
  throw; `{$kind}` alone, arbitrary `$`-keys, nested `$generated`, strings,
  arrays, numbers pass; `allowIfSet` no-op path pinned; fresh-cell `allowIfSet`
  still throws.
- Builder path: reserved cause inside a `pattern(...)` body throws at build
  time; `{outer: {$generated}}` and `{$kind: "stream"}` flow through as partial
  causes unchanged.
- Duplicate-name wrap (`{ name, $generated: N }`) already pinned by
  "uniquifies repeated stable internal paths with schemas" — unchanged.

## Gates

- `deno fmt --check`, `deno lint`, `deno task check`: clean.
- Full runner suite: green (754+ tests, 0 failed).
- `deno task integration pattern-tests`: ✅ all 85 pattern tests passed.
- `deno task integration generated-patterns`: ✅ 145 passed / 0 failed.

Closes CT-1822.

---
## Summary by cubic
Reject user-supplied causes with a top-level "$generated" key by throwing at Cell.for(); this prevents deceptive collisions and lets the builder skip collision checks for generated causes. Implements CT-1822.

- **New Features**
  - Enforce reserved key: throw when a record cause has top-level "$generated" (checked in Cell.for(), re-asserted in the builder).
  - Scope: only "$generated", top-level only; nested "$generated" allowed; other "$"-keys allowed; non-record causes unaffected; allowIfSet remains a no-op if a cause already exists.
  - Builder simplification: generated causes are collision-free by construction; only named causes are deduped via a Set of value hashes.
  - Docs and tests: added API comment in `packages/api`, and tests in `packages/runner/test`.

- **Migration**
  - If any `.for({ $generated: ... })` exists, rename the key (e.g., to "generated").

<sup>Written for commit 5d73f453def87087864441f91194a2647be6f8b5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4494?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -890,6 +890,9 @@ export interface ICreatable<C extends AnyBrandedCell<any>> {
   /**
    * Set a cause for this cell. Used to create a link when the cell doesn't have
    * one yet.
+   *
+   * Record causes may not use the top-level key `$generated` — it is
+   * reserved for system-generated causes and setting one throws.
    * @param cause - The cause to associate with this cell
    * @returns This cell for method chaining
    */
--- a/packages/runner/src/builder/pattern.ts
+++ b/packages/runner/src/builder/pattern.ts
@@ -1,6 +1,7 @@
 import { isRecord } from "@commonfabric/utils/types";
 import { deepEqual } from "@commonfabric/utils/deep-equal";
 import { hashStringOf } from "@commonfabric/data-model/value-hash";
+import { toCompactDebugString } from "@commonfabric/data-model/value-debug";
 import {
   type CellScope,
   type FactoryInput,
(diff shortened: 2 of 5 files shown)


ubik2 (approved):
lgtm!
thanks

PR #4499, opened 2026-07-02
fix(patterns): empty-string guards in profile setName/setAvatar (CT-1828)

bfollington:
## Why

Hitting send on an empty name or avatar box erases the canonical value product-wide — the display falls back to the literal "Profile" once the name is blanked. `setName` and `setAvatar` in `profile-home.tsx` trim the incoming value and call `.set()` unconditionally, so an accidental empty submit destroys the profile's identity with no undo. The create handler (`profile-create.tsx:90`) already guards against this (`if (name)`); the edit-time handlers did not. This becomes embedder-facing once Loom mounts the profile editor: https://github.com/commontoolsinc/loom/pull/3627.

## What

- Add an empty-after-trim guard to `setName` and `setAvatar` in `packages/patterns/system/profile-home.tsx`: if the trimmed value is empty, the handler returns (no-op) instead of writing `""`.
- `setBio` is deliberately left untouched and clearable — clearing a bio is a legitimate action; erasing a name or avatar is not.
- Extend `packages/patterns/system/profile-home.test.tsx` (the existing pattern-unit-test harness for this pattern) to cover the guard: an empty send and a whitespace-only send both leave the prior name/avatar intact, and a genuine non-empty send still writes.

Linear: CT-1828

---
## Summary by cubic
Prevents empty name or avatar submissions from wiping a profile. Adds no-op guards so blank or whitespace input is ignored, aligning with Linear CT-1828.

- **Bug Fixes**
  - Add empty-after-trim guards in `packages/patterns/system/profile-home.tsx` for `setName` and `setAvatar` to avoid writing `""`.
  - Leave `setBio` clearable by design.
  - Extend `packages/patterns/system/profile-home.test.tsx` to cover empty/whitespace sends as no-ops and verify non-empty writes still apply.

<sup>Written for commit a58975358a822e89fd461193b52c6bdae75d13eb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4499?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/patterns/system/profile-home.test.tsx
+++ b/packages/patterns/system/profile-home.test.tsx
@@ -17,6 +17,11 @@ export default pattern(() => {
     profile.setAvatar.send({ avatar: "AL" });
   });
 
+  // CT-1828: same empty-after-trim guard applies to setAvatar.
+  const action_clear_avatar = action(() => {
+    profile.setAvatar.send({ avatar: "" });
+  });
+
   // CT-1748 note: the presentation-vs-edit *content* lives behind `ifElse`, and
   // [UI] vnode traversal doesn't resolve through this test harness's `.get()`,
   // so the toggle is asserted via the exported `isEditing` and the bound data
@@ -31,6 +36,10 @@ export default pattern(() => {
   // The avatar the badge binds resolves (single context — owner-protected reads
   // are clean here; cross-stamp masking is the browser-only CT-1740 issue).
   const assert_avatar_set = computed(() => profile.avatar === "AL");
+  // CT-1828: an empty send must not clear a previously-set avatar.
+  const assert_avatar_unchanged_after_empty = computed(() =>
+    profile.avatar === "AL"
+  );
 
   const action_add_catalog_element = action(() => {
     profile.addElement.send({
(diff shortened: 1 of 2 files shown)


bfollington (commented):
_cf-review via Claude Opus 4.8, on behalf of @bfollington — automated review, human owns the merge._

## Verdict: Clean ✅ — safe to merge

**Scope · coverage:** empty-string guards in `setName`/`setAvatar` (2 files, +40/-2), deep-read in full. Checks: `deno task cf check` on the test pattern **passed**; `deno lint` on both files surfaced only pre-existing `as any` hits (none in the diff).

### 🔴 Blocking
None.

### ⚪ Nits (optional)
- **Avatar whitespace coverage asymmetry** (`profile-home.test.tsx`): the name path tests both empty (`""`) and whitespace-only (`"   "`); the avatar path tests only empty. Both handlers share `.trim()` + `!value`, so the whitespace branch is covered transitively — a whitespace-only avatar step would just make it symmetric.

### Notes (verified)
- **No reactivity hazard.** Both inputs are `cf-message-input` with `oncf-send` (send-on-submit), not a `$value` two-way binding; the current value renders separately via `<strong>{name}</strong>` / `<span>{avatar}</span>`. The guard strands no stale value and creates no self-feedback loop.
- **Parity claim holds.** `profile-create.tsx:90` genuinely guards with `if (name)`; `setBio` genuinely omits the guard by design (it clears its draft after save). The fix mirrors the established idiom rather than forking new machinery.
- **Test genuinely guards the fix.** It sets a real value first ("Grace Hopper" / "AL"), then sends empty/whitespace and asserts the prior value survives — against the pre-fix code the `""` write would fail the assertion. The removed `assert_name_cleared` correctly dropped coverage of the old (buggy) clear-on-empty behavior, not real coverage.
- No anti-duplication, determinism, framework-fighting, or changeset-hygiene concerns.

PR #4500, opened 2026-07-02
feat(ui): cancellable cf-open-external event for modifier-click new-tab navigation (CT-1830)

bfollington:
## Why

Plain clicks in `cf-cell-link`, `cf-profile-badge`, and `cf-render` tiles dispatch the interceptable `cf-navigate` event on `globalThis` — an embedder like Loom can host it and apply its own URL scheme. But **modifier-click** (cmd/ctrl → "open in new tab") did not: each component built a shell URL via `appViewToUrlPath` and called `globalThis.open()` directly, dispatching nothing.

That makes the new-tab path **un-interceptable** and a **guaranteed 404 on any non-shell origin**. Loom is shipping a server-side redirect as a stopgap (commontoolsinc/loom#3627); the right fix is upstream — make new-tab navigation interceptable the same way in-place navigation already is.

## What

- **New `openInNewTab(command)` helper** next to `navigate()` in `packages/shell/shared/navigate.ts` (re-exported through `@commonfabric/shell/shared`, where all three components already import `navigate`). It dispatches a **cancellable `cf-open-external`** event on `globalThis` (matching where `cf-navigate` is dispatched) carrying the same `AppView` detail (`{spaceDid|spaceName, pieceId}`) **before** URL construction.
  - A host cancels via `preventDefault()` and owns the navigation.
  - If not cancelled, the shell default runs: today's `appViewToUrlPath` + `globalThis.open` — **shell behavior unchanged, zero callback plumbing**.
- **All three components** (`cf-cell-link`, `cf-profile-badge`, `cf-render`) now call the one helper — no three copies of the URL-building logic. Their modifier-click branches collapse to `openInNewTab(view)`.
- **Event typing** mirrors the existing `cf-navigate` idiom: a `CustomEvent<NavigationCommand>` subclass (`OpenExternalEvent`), consistent with `NavigationEvent`/`ReplaceNavigationEvent`. (The repo does not use a `WindowEventMap`/`HTMLElementEventMap` augmentation for these — the pattern is a typed `CustomEvent` subclass + cast on the listener.)

### Event contract

- **Name:** `cf-open-external`
- **Detail:** `NavigationCommand` (= `AppView`, e.g. `{spaceDid, pieceId}`) — same shape as `cf-navigate`
- **Dispatch target:** `globalThis` (same as `cf-navigate`)
- **Cancelable:** yes — `preventDefault()` suppresses the default `globalThis.open`

### Deliberate unification

`cf-profile-badge` already passed `"noopener"` to `globalThis.open`; `cf-cell-link` and `cf-render` did not. The unified helper uses `globalThis.open(url, "_blank", "noopener")` for all three — `"noopener"` is strictly safer for externally-opened tabs.

## Tests

- `packages/shell/test/navigate.test.ts` (new) — covers the helper directly: dispatches a cancellable `cf-open-external` with the right detail; falls through to `globalThis.open` with `_blank` + `noopener` and the expected URL when not cancelled; `preventDefault()` suppresses `globalThis.open`.
- `packages/ui/src/v2/components/cf-profile-badge/cf-profile-badge.test.ts` — added two tests following the existing `cf-navigate`-contract idiom: modifier-click dispatches cancellable `cf-open-external` with the right detail; `preventDefault()` suppresses `globalThis.open`. The pre-existing new-tab test (default URL unchanged) still passes.

`deno fmt`, `deno lint`, `deno check`, and the touched-package tests all pass.

Linear: CT-1830
Loom stopgap: https://github.com/commontoolsinc/loom/pull/3627

---
## Summary by cubic
Make modifier-click “open in new tab” interceptable by dispatching a cancellable `cf-open-external` event, preventing 404s on non-shell origins. Default shell behavior stays the same; addresses Linear CT-1830.

- **New Features**
  - Added `openInNewTab(command)` in `packages/shell/shared/navigate.ts` (re-exported via `@commonfabric/shell/shared`). It dispatches cancellable `cf-open-external` on `globalThis` with `NavigationCommand` detail; hosts can `preventDefault()`. If not cancelled, falls back to `appViewToUrlPath` + `globalThis.open(url, "_blank", "noopener")`.
  - Updated `cf-cell-link`, `cf-profile-badge`, and `cf-render` to use the helper on cmd/ctrl-click. Unified new-tab opens to include `"noopener"` for all.

<sup>Written for commit 44508f545bd871a1ee4aab21430eb9936036b036. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4500?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/shell/shared/navigate.ts
+++ b/packages/shell/shared/navigate.ts
@@ -29,6 +29,39 @@ export function navigate(command: NavigationCommand) {
   globalThis.dispatchEvent(new NavigationEvent(command));
 }
 
+const OpenExternalEventName = "cf-open-external";
+
+// Cancellable event dispatched before a modifier-click ("open in new tab")
+// builds a shell URL. An embedder (e.g. Loom) can host `cf-open-external` and
+// call `preventDefault()` to apply its own URL scheme; if the event is not
+// cancelled, the shell default runs (`appViewToUrlPath` + `globalThis.open`).
+// Mirrors the `cf-navigate` idiom so plain and modifier clicks are both
+// interceptable from a single, well-known event surface.
+class OpenExternalEvent extends CustomEvent<NavigationCommand> {
+  command: NavigationCommand;
+  constructor(command: NavigationCommand) {
+    super(OpenExternalEventName, { detail: command, cancelable: true });
+    this.command = command;
+  }
+}
+
+// Open a navigation target in a new tab. Dispatches a cancellable
+// `cf-open-external` event first; if a host cancels it via `preventDefault()`,
+// the host owns the new-tab navigation. Otherwise the shell default builds a
+// URL from the current location and calls `globalThis.open`.
+export function openInNewTab(command: NavigationCommand) {
+  const event = new OpenExternalEvent(command);
+  const proceed = globalThis.dispatchEvent(event);
+  if (!proceed) return;
+  const url = appViewToUrlPath(
+    preserveAppViewMode(
+      urlToAppView(new URL(globalThis.location.href)),
+      command,
+    ),
+  );
+  globalThis.open(url, "_blank", "noopener");
+}
+
 class ReplaceNavigationEvent extends CustomEvent<NavigationCommand> {
   command: NavigationCommand;
   constructor(command: NavigationCommand) {
--- a/packages/shell/test/navigate.test.ts
+++ b/packages/shell/test/navigate.test.ts
@@ -1,0 +1,107 @@
+import { describe, it } from "@std/testing/bdd";
+import { expect } from "@std/expect";
+import { type NavigationCommand, openInNewTab } from "../shared/navigate.ts";
+
+// Exercises the `openInNewTab` helper that backs modifier-click ("open in new
+// tab") navigation for cf-cell-link, cf-profile-badge and cf-render (CT-1830).
+// The helper dispatches a cancellable `cf-open-external` event before building
+// a shell URL; a host cancels it via `preventDefault()` to apply its own URL
+// scheme, otherwise the shell default (`globalThis.open`) runs.
+
+// `openInNewTab` reads `globalThis.location.href` and calls `globalThis.open`.
+// Deno has neither unless configured, so stub both around the call and restore.
+function withStubbedEnv<T>(
+  href: string,
+  run: (calls: { openArgs: unknown[][] }) => T,
+): T {
+  const calls = { openArgs: [] as unknown[][] };
+  const hadLocation = "location" in globalThis &&
+    globalThis.location !== undefined;
+  // deno-lint-ignore no-explicit-any
+  const origLocation = (globalThis as any).location;
+  // deno-lint-ignore no-explicit-any
+  const origOpen = (globalThis as any).open;
+  Object.defineProperty(globalThis, "location", {
+    value: { href },
+    configurable: true,
+    writable: true,
+  });
+  // deno-lint-ignore no-explicit-any
+  (globalThis as any).open = (...args: unknown[]) => {
+    calls.openArgs.push(args);
+    return null;
+  };
+  try {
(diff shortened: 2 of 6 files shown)


bfollington (commented):
_cf-review via Claude Opus 4.8, on behalf of @bfollington — automated self-signed review, human owns the merge._

## Verdict: clean — approve. Nothing should block merge.

**Scope · coverage · checks** — 6 files, +265/−41; platform (shell) + UI, no pattern/transformer surface. Deep-read all changed files plus the surrounding `navigate.ts` idiom. Ran `deno test` on `packages/shell/test/navigate.test.ts` and `cf-profile-badge.test.ts` (all green), `deno lint` + `deno fmt --check` on all 6 touched files (clean), and a repo-wide grep for the new symbols.

**What it does** — Modifier-click "open in new tab" in `cf-cell-link`, `cf-profile-badge`, `cf-render` previously called `globalThis.open()` on a locally-built shell URL, dispatching nothing — so it was un-interceptable and 404s on any non-shell origin. This adds one `openInNewTab(command)` helper in `shell/shared/navigate.ts` that dispatches a **cancellable** `cf-open-external` event first (a host calls `preventDefault()` to own the navigation), then falls through to the unchanged shell default. All three components collapse to a one-line delegation. Scope is coherent and tight.

### 🔴 Blocking
None.

- **No embedder-facing doc for the new interception surface.** `cf-open-external` is a new public event meant to be hosted by an embedder (Loom). Its sibling events (`cf-navigate`, `cf-replace-navigation`) also live only as internal-idiom code with no doc, so this is **not a regression against a documented invariant** — but since the whole point here is that an external host wires a listener, the contract (name · `cancelable: true` · `detail: AppView` · dispatched on `globalThis`) currently exists only in the PR body and a source comment. Worth a short note wherever the navigation-event surface gets documented for embedders, or a follow-up if no such doc exists yet. _(suspected — no doc found for the sibling events either, so this is a forward-looking nudge, not a defect.)_

### ⚪ Nits
- The `"noopener"` unification is deliberate and strictly safer; `cf-cell-link` and `cf-render` previously omitted it. Verified there are **no** existing new-tab tests in those two components asserting the old (no-`noopener`) behavior, so nothing was silently bent to fit.
- `OpenExternalEvent` carries both `this.command` and the `detail` payload — redundant, but it exactly mirrors the existing `NavigationEvent`/`ReplaceNavigationEvent` classes, so it is consistent idiom rather than new cruft.

### Why it's clean
- **Event contract** is coherent with the `cf-navigate` idiom: same dispatch target (`globalThis`), same `AppView`/`NavigationCommand` detail shape, adds `cancelable: true`. `OpenExternalEvent`/`OpenExternalEventName` stay unexported (like `NavigationEvent`); only `openInNewTab` is exported through `@commonfabric/shell/shared`.
- **Default parity**: the un-cancelled path is byte-identical to the three old per-component snippets — `appViewToUrlPath(preserveAppViewMode(urlToAppView(new URL(location.href)), command))` — plus the intentional `noopener` unification. No shell-side listener is needed because the shell *is* the default and runs it inline; verified no `cf-open-external` listener exists or is required.
- **Tests pin the contract**: cancelable + correct detail; fall-through to `globalThis.open(url, "_blank", "noopener")` with the expected URL; `preventDefault()` suppression — at both the helper level and (for the badge) the component level.
- **Anti-duplication improved**: three copies of the URL-building logic collapse to one canonical helper.

_Automated review — verify anything load-bearing before relying on it._

PR #4504, opened 2026-07-03
fix(runner): mint keyless op identities at instantiation (CT-1812)

mathpirate:
## Summary

Fixes the reachable class of **CT-1812** by dissolving it: a keyless `map`/`filter`/`flatMap` op whose original is a live trusted builder pattern is now **minted its content-hash session identity at node instantiation**, so it rides a `$patternRef` to its pristine artifact instead of the embedded graph whose nested output-alias `defer` levels the immutable-cell round-trip corrupts (the CT-1811 mechanism, previously reachable ref-lessly). One narrowly-scoped residual remains and is documented in place (below).

## Why this shape

A design spike (evidence on the ticket) established that no product path used the embedded fallback — with it patched to throw: cli 912/0, full pattern sweep 956/0, generated-patterns 145/0, zero hits. The constituencies were (a) ~18 runner unit-test groups building patterns with the in-process builder DSL, and (b) one pinned contract test. The identity program (E4 #4083, E5 #4110) already converged on refs-only-and-loud, and the mint already existed: `PatternManager.ensureKeylessPatternIdentity` (CT-1623 structural dedup) is what `entryRefForPattern` mints for keyless ROOT patterns at run time. The gap was exactly that `Runner.substituteOpPatternRefs` consulted `getArtifactEntryRef` but never minted for nested ops.

## The change

**`Runner.substituteOpPatternRefs`**: on an entry-ref miss, resolve the op's derivation original and — only if it is already a **trusted builder pattern** (`isTrustedBuilderArtifact` + `isPattern`; minting brands, so it must never confer trust on a forged object) — mint via `ensureKeylessPatternIdentity` and stamp the sentinel. Structurally identical ops share one identity (content hash), so re-instantiation doesn't churn. This covers every live-object keyless case: in-process builder DSL patterns and bare-`Engine.compileAndEvaluateModules` namespaces.

## The residual, and why the builtin is NOT sealed

The spike recommended sealing `resolveOpPattern` against bare graphs outright. Implementation surfaced a **pinned live contract** the seal would break: `stored-pattern-rehydration.test.ts` — "a stored full graph (no-entry-ref writer) still executes via `runtime.run` — that is a live writer path, not a vintage." A graph deserialized from a stored keyless pattern VALUE has **no pristine artifact in existence** to mint a pointer to:

- Minting the bound instantiation copy self-cycles (its aliases are doc-bound to the map node — verified empirically: "link at [] targets its own subpath").
- Minting the raw embedded form hands per-element instantiation a one-level-too-deep graph.
- Constructing the canonical form would mean re-rooting defers bind-free — exactly the surgery CT-1812 records as the hard option this PR avoids.

So the embedded path remains, **reduced to precisely that stored-keyless remnant**, documented at both ends (`substituteOpPatternRefs`, `resolveOpPattern`): a stored keyless graph with a *simple* op keeps working (the pinned contract); one with a grandchild derived-internal output remains defer-corrupted — same as before this PR, now stated in place and on the ticket rather than latent. Every in-repo-reachable case rides identities.

(A container-trust tier and a builtin-side seal were both built and reverted during implementation: the first for the self-cycle above, the second for the pinned contract. `entryRefForPattern` brands every keyless root at run, so container-trust gating would also have guarded nothing.)

## Tests

- **New regression** (`runner/test/keyless-op-identity.test.ts`): the CT-1811 pattern shape (mapped Wrapper nesting a grandchild with a derived-internal computed output), evaluated through the bare non-registering `Engine.compileAndEvaluateModules`, run directly. Asserts the op was keyless going in, the grandchild outputs materialize correctly (2/4/6), and the op's `keyless:` identity was minted by instantiation. **Fails-without validated** (mint disabled → fails).
- The two pass-through pins (`map-op-by-identity.test.ts`, `pattern-ref-boundary.test.ts`) keep their pass-through expectation with enriched comments naming the remnant scope and cross-linking the contract test.
- **Acceptance**: the ~18 previously-fallback-dependent test groups (patterns-dynamic, pattern-scope, cfc-flow-pointwise, list-builtin-edge-paths, …) and the stored-graph contract test pass **unchanged**. The CFC pointwise suites passing answers the minted-provenance question empirically.

## Verification

| Gate | Result |
| -- | -- |
| New regression (fails-without / passes-with) | ✅ |
| runner full suite | 756 passed / 0 failed |
| cli full suite | 912 passed / 0 failed |
| pattern sweep (`--root packages/patterns`) | 956 passed / 0 failed |
| generated-patterns integration | 145 passed / 0 failed |
| fmt · lint | ✅ |

---
## Summary by cubic
Mints a keyless content-hash identity for `map`/`filter`/`flatMap` ops at instantiation when their original is a trusted builder `pattern`, routing them via `$patternRef` to the pristine artifact. Fixes CT-1812 by avoiding the embedded-graph path that corrupts nested `defer` levels; only stored keyless graphs still run embedded.

- **Bug Fixes**
  - In `Runner.substituteOpPatternRefs`, on a missing entry ref resolve the op’s original and, if it’s a trusted builder `pattern`, mint via `ensureKeylessPatternIdentity` and substitute a `$patternRef` (uses `resolveOriginal` + `isTrustedBuilderArtifact` + `isPattern`).
  - Clarify in `resolveOpPattern` that the embedded path is now only the stored-keyless remnant; document the residual `defer` corruption in-code.
  - Add regression `keyless-op-identity.test.ts`; update comments in `map-op-by-identity.test.ts` and `pattern-ref-boundary.test.ts`.

<sup>Written for commit d316a2c058703a41b2ed186f4657a5f939bf4003. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4504?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/builtins/op-pattern-ref.ts
+++ b/packages/runner/src/builtins/op-pattern-ref.ts
@@ -47,8 +47,17 @@ export function isPatternRefSentinel(
  *   same way when its module evaluated this session; a ref from a module that
  *   never evaluated here throws — loud, since a sync Action cannot await the
  *   storage-backed `loadPatternByIdentity`.
- * - Otherwise (no entry ref known at instantiation), `op` is the embedded
- *   pattern graph itself, used as-is.
+ * - Otherwise `op` is an embedded pattern graph, used as-is. Post-CT-1812
+ *   this is ONLY the stored-keyless remnant: a live op whose original is a
+ *   trusted builder pattern gets a `keyless:` identity minted at node
+ *   instantiation (`Runner.substituteOpPatternRefs`) and arrives here as a
+ *   sentinel; what still arrives embedded is a graph deserialized from a
+ *   stored no-entry-ref pattern VALUE (the live keyless writer path pinned
+ *   by stored-pattern-rehydration.test.ts), for which no pristine artifact
+ *   exists to resolve instead. CT-1812 residual: for THAT form, a nested
+ *   grandchild's derived-internal output aliases remain defer-corrupted by
+ *   the immutable-cell round-trip — re-rooting a bare stored graph without
+ *   binding is the open surgery recorded on the ticket.
  */
 export function resolveOpPattern(
   runtime: Runtime,
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -88,7 +88,11 @@ import {
 } from "./cfc/types.ts";
 import { runInActionExecution } from "./builder/action-context.ts";
 import { getVerifiedProvenance } from "./harness/verified-provenance.ts";
-import { getArtifactEntryRef } from "./builder/pattern-metadata.ts";
+import {
+  getArtifactEntryRef,
+  isTrustedBuilderArtifact,
+  resolveOriginal,
+} from "./builder/pattern-metadata.ts";
 import { diffAndUpdate } from "./data-updating.ts";
 import { setResultCell } from "./result-utils.ts";
 import { SigilLink } from "./sigil-types.ts";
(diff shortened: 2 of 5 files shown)


ubik2:
This is probably best reviewed by @seefeldb 
Short version is that Runner.substituteOpPatternRefs used to expect to use the `op` to find the `identity` and `symbol` in a lookup table, but sometimes that's not there (for a hand-built pattern?), so we assign a `keyless:blah` identity, and `default` as the symbol instead.

seefeldb (approved):

PR #4510, opened 2026-07-03
fix(runner): materialize FabricSpecialObject values as leaves in schema traversal (CT-1836)

mathpirate:
## Summary

Fixes **CT-1836**: every lift consuming a `fetchBinary` result — at any nesting depth — was permanently gated, because schema-query materialization silently **dropped the `FabricBytes`-bearing `result` field**. One traversal arm fixes it; an end-to-end regression (mocked binary fetch → consumer lifts) pins it.

## Root cause

`SchemaObjectTraverser`'s value dispatch had no `FabricSpecialObject` arm — a gap its own `TODO(danfuzz)` comment already recorded. A `FabricBytes` (typeof `"object"`, no enumerable own props — `length` is a prototype getter) fell into the record branch and was decomposed by `Object.entries` into `{}`, failing the schema-generator's structural object schema for `FetchBinaryResult.bytes` (`{type:"object", properties:{length…}}`). The traversal then dropped the containing `result` field entirely (`required` unmet), so:

- the consumer lift's argument stayed invalid forever → its **body never ran** (`readJavaScriptArgument` → `isValidArgument` gate), freezing `fetchState`/downstream consumers/UI overlays;
- the consumers' crippled read logs also never registered the forward `dependents` edges the post-writeback wake walks, which is why the freeze looked like a scheduler defect for several investigation rounds (full falsification trail on CT-1836 — six instrumented rounds);
- `fetchJson` consumers were unaffected (plain JSON values), which disguised the regression: it arrived with the fetchData→fetchBinary split (#4206), so **every `fetchBinary` consumer since then has been frozen**, e.g. the lunch-poll generated art.

## The fix

A `FabricSpecialObject` leaf arm in the dispatch, before the record branch: validate loosely as `"object"` (the shape the schema-generator emits for these types today) and pass the value through unchanged via `traversePrimitive` — which is the fabric type system's documented contract for `FabricPrimitive` ("always frozen, passes through conversion unchanged"). The pre-existing TODO is narrowed to the remaining `FabricInstance` codec-descent design.

Verified chain after the fix (probes on the lunch-poll art flow): binary fetch completes → `FabricBytes` result materializes through schema queries → consumer bodies run → base64 re-encode works → **cross-boundary reads of fetch-derived sub-pattern outputs work** (`art.imageDataUrl` readable from a parent — the CT-1811-family "sub-pattern with fetch + computed outputs" gap this closes).

## Regression test

`runner/test/fetch-binary-schema-materialization.test.ts` — the whole chain, self-contained via the injectable `RuntimeOptions.fetch` (CT-1768): mocked PNG response → `fetchBinary` → consumer lifts reading `result.mediaType` and re-encoding `result.bytes` to a data URL → asserts both materialize. **Fails without the fix** (validated by stash-revert: `0 passed, 1 failed`), passes with it.

## Related (documented on CT-1836, deliberately not in this PR)

- A scheduler hardening (wake a general pass when mark-dirty finds no forward effects) was drafted and probe-validated during the investigation; this fix removes its known trigger (the crippled read logs), so it's parked on the ticket rather than shipped.
- The lunch-poll `art-sync.test.tsx` canary (branch `gideon/lunch-poll-art-readd`) pins the pre-fix behavior and will go RED when this lands — flipping it to the full persist contract is the follow-up there (the card's notify idiom also gets simplified: cross-boundary reads work now, and send-inside-computed is fatal in the current runtime — kills the harness worker silently, observed during round 7).

## Verification

| Gate | Result |
| -- | -- |
| New regression (fails-without / passes-with) | ✅ |
| runner full suite | 759 passed / 0 failed |
| cli full suite | 913 passed / 0 failed |
| fmt · lint | ✅ |

---
## Summary by cubic
Fixes CT-1836 by treating `FabricSpecialObject` values (e.g., `FabricBytes`) as leaves in schema traversal so `fetchBinary` results materialize and consumer lifts run. Adds an end-to-end regression and unit tests to pin the behavior.

- **Bug Fixes**
  - Root cause: no `FabricSpecialObject` arm; `FabricBytes` was decomposed to `{}`, the `result` field was dropped, and consumers stayed gated.
  - Fix: add a `FabricSpecialObject` leaf branch before record handling, validate as "object", and pass the value through unchanged.
  - Tests: new `fetch-binary-schema-materialization.test.ts` (e2e) and `traverse.test.ts` (unit accept/reject) cover the contract; comments updated to describe current behavior.

<sup>Written for commit c8b418e8c6e657a9bcee34b0f4051e8462d26c39. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4510?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/test/fetch-binary-schema-materialization.test.ts
+++ b/packages/runner/test/fetch-binary-schema-materialization.test.ts
@@ -82,7 +82,9 @@ describe("fetchBinary consumer materialization (CT-1836)", () => {
     runtime = new Runtime({
       apiUrl: new URL(import.meta.url),
       storageManager,
-      // Deterministic binary response; no network (CT-1768 seam).
+      // `RuntimeOptions.fetch` overrides the outbound fetch for this runtime
+      // instance, so the builtin resolves against this deterministic binary
+      // response instead of the network.
       fetch: () =>
         Promise.resolve(
           new Response(TINY_PNG_BYTES, {
--- a/packages/runner/test/fetch-binary-schema-materialization.test.ts
+++ b/packages/runner/test/fetch-binary-schema-materialization.test.ts
@@ -1,0 +1,128 @@
+import { afterEach, beforeEach, describe, it } from "@std/testing/bdd";
+import { expect } from "@std/expect";
+import { Identity } from "@commonfabric/identity";
+import { StorageManager } from "@commonfabric/runner/storage/cache.deno";
+import { Runtime } from "../src/runtime.ts";
+import type { RuntimeProgram } from "../src/harness/types.ts";
+import type { IExtendedStorageTransaction } from "../src/storage/interface.ts";
+
+/**
+ * Regression for CT-1836: a lift consuming a `fetchBinary` result was
+ * permanently gated because schema-query materialization DROPPED the
+ * `FabricBytes`-bearing `result` field.
+ *
+ * Mechanism: `SchemaObjectTraverser`'s value dispatch had no
+ * `FabricSpecialObject` arm, so a `FabricBytes` fell into the record branch
+ * and was decomposed by `Object.entries` over its (empty) own props. That
+ * failed the schema-generator's structural object schema for
+ * `FetchBinaryResult.bytes`, the traversal dropped the containing `result`
+ * field entirely (`required` unmet), the consumer's argument stayed invalid,
+ * and its body never ran — freezing every downstream consumer, at any
+ * nesting depth. (`fetchJson` consumers were unaffected: plain JSON values.)
+ * The collateral: the consumers' crippled read logs also never registered
+ * the forward dependents edges the post-writeback wake relies on.
+ *
+ * The fix treats `FabricSpecialObject` values as opaque leaves (the fabric
+ * type system's documented contract: frozen, pass through conversion
+ * unchanged).
+ *
+ * This test runs the WHOLE chain — mocked binary fetch (via the injectable
+ * `RuntimeOptions.fetch`, CT-1768) → `FabricBytes` result → consumer lifts
+ * reading `result.mediaType` and re-encoding `result.bytes` — and asserts
+ * the consumers actually materialize. Fails without the traverse fix (both
+ * outputs stay empty forever); passes with it.
+ */
+
+const signer = await Identity.fromPassphrase("fetch-binary-materialization");
+const space = signer.did();
+
+// 1×1 transparent PNG.
+const TINY_PNG_BASE64 =
+  "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNkYPhfDwAChwGA60e6kgAAAABJRU5ErkJggg==";
+const TINY_PNG_BYTES = Uint8Array.from(
+  atob(TINY_PNG_BASE64),
+  (c) => c.charCodeAt(0),
+);
+
+const PROGRAM: RuntimeProgram = {
+  main: "/main.tsx",
+  files: [
+    {
+      name: "/main.tsx",
+      contents: [
+        "import { computed, fetchBinary, pattern } from 'commonfabric';",
+        "export default pattern(() => {",
+        "  const art = fetchBinary({ url: 'https://mock.test/img' });",
+        "  const mediaType = computed(() => art.result?.mediaType ?? '');",
+        "  const dataUrl = computed(() => {",
+        "    const bytes = art.result?.bytes;",
+        "    const mt = art.result?.mediaType;",
+        "    if (!bytes || !mt) return '';",
+        "    const raw = bytes.slice();",
+        "    let binary = '';",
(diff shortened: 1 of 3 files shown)


ubik2:
For reference, the generated schema for the `bytes` field of a `FetchBinaryResult` is:
```javascript
{
  type: "object",
  properties: { length: { type: "number" } },
  required: ["length"],
}
```

This is the schema generated for `FabricBytes` objects, but as pointed out in the PR, `Object.entries` on a `FabricBytes` object isn't going to return an entry for the `length`, so we will never validate this object.

We don't really want to go through the `{ type: "object" }` path either, since that will have us creating a new object.

It's possible we should add a new type description for `FabricBytes` or more generally, `FabricPrimitive`. If we go that route, the transformer needs to recognize it, and insert something more like:
```javascript
{
  type: "fabricprimitive",
}
```

Edit: I should mention that this PR takes the approach of just allowing any `FabricPrimitive` to match the schema if the schema allows type `object`, and returns that primitive directly. That's a good workaround, but it may be too permissive.

packages/runner/test/fetch-binary-schema-materialization.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  let storageManager: ReturnType<typeof StorageManager.emulate>;
+  let runtime: Runtime;
+  let tx: IExtendedStorageTransaction;
+
+  beforeEach(() => {
+    storageManager = StorageManager.emulate({ as: signer });
+    runtime = new Runtime({
+      apiUrl: new URL(import.meta.url),
+      storageManager,
+      // Deterministic binary response; no network (CT-1768 seam).
```
Hixie: what does "(CT-1768 seam)" mean?

mathpirate: whoops thanks for noticing. CT-1768 was the ticket for making `fetch` injectable per-runtime so tests can mock outbound HTTP - 'seam' here was in the sense of a designed substitution point for testing. updated the comment to just say what it actually means

ubik2: CT-1768 refers to https://linear.app/common-tools/issue/CT-1768
`seam` is a word our agents use a lot to describe a mismatch of expectations/interfaces.
CT-1768 is about mocking fetchData

mathpirate:
> For reference, the generated schema for the `bytes` field of a `FetchBinaryResult` is:
> 
> ```js
> {
>   type: "object",
>   properties: { length: { type: "number" } },
>   required: ["length"],
> }
> ```
> 
> This is the schema generated for `FabricBytes` objects, but as pointed out in the PR, `Object.entries` on a `FabricBytes` object isn't going to return an entry for the `length`, so we will never validate this object.
> 
> We don't really want to go through the `{ type: "object" }` path either, since that will have us creating a new object.
> 
> It's possible we should add a new type description for `FabricBytes` or more generally, `FabricPrimitive`. If we go that route, the transformer needs to recognize it, and insert something more like:
> 
> ```js
> {
>   type: "fabricprimitive",
> }
> ```
> 
> Edit: I should mention that this PR takes the approach of just allowing any `FabricPrimitive` to match the schema if the schema allows type `object`, and returns that primitive directly. That's a good workaround, but it may be too permissive.

thank you, these are all good points. agreed that the record path would be wrong even if it worked bc it would `createObject` a plain record and lose the instance. it's true that the current implementation on this PR is too permissive; given the schema that's currently emitted here, pass-through-on-object is the only behavior i/fable could find that honors the primitive contract without re-failing `required: ["length"]`. 

i agree we should do better here, i'll pick it up as a follow-up and see where i can get. the transformer may need to do something a little more like `fabricType: "bytes"` so it has the concrete class for validation.

ubik2 (approved):
lgtm (so we have the record here too)

ubik2:
> i agree we should do better here, i'll pick it up as a follow-up and see where i can get. the transformer may need to do something a little more like `fabricType: "bytes"` so it has the concrete class for validation.

If we want to have the schema enforce the type, maybe `fabric:FabricBytes`? A type field that starts with `fabric:` would then check that it's a `FabricSpecialObject`, and use the portion after the `:` to check the actual constructor with `instanceof`.

I wouldn't try to plumb this into the schema->type system, but that may be possible (so you can do a `cell.get().copyInto(buffer)` without casting).

Edit: another approach that would take a little more tweaking is
```javascript
type: { base: "FabricSpecialObject", class: "FabricBytes" }
```
Essentially, a type object instead of just a string. Gets trickier with unions.
```javascript
type: ["string", "null", { base: "FabricSpecialObject", class: "FabricBytes" }]
```

PR #4502, opened 2026-07-02
docs(embedding): host embedding contract + seam tests (CT-1832)

bfollington:
## Why

Every profile-integration drift incident so far was **upstream changing a seam it did not know was a seam**:

- **labs#4371** rewrote the profile *create* surface (`cf-input`/`cf-button` → a new `cf-submit-input`) days after Loom pinned the vendor — a silent blank render on the next bump.
- **labs#4415** changed `#profile` resolution semantics (`{ordered, defaultValid}`, picker only on genuine ambiguity).
- **`cf-cell-link` clicks dispatch `cf-navigate` into the void** in Loom production today because nothing on Loom's origin listens for the event — shipped, silent breakage.

None were careless; the seam simply wasn't named, so nobody owed it a compatibility thought. Loom — and any future non-shell embedder — binds to a *specific* set of seams. Making that set a named, tested, agent-readable contract means a change that breaks an embedder **fails CI upstream, in labs**, instead of silently blanking a render in the embedder weeks later.

## What

One doc — `docs/development/HOST_EMBEDDING.md` — plus one contract test per seam, each placed in the package that owns the seam and each written to **go red when the contract changes**. The doc is written so another agent can read it and understand how to embed labs components in a host.

| # | Seam | Doc § | Package | Test |
|---|------|-------|---------|------|
| 1 | Wish targets + `result ?? fallback` | §1 | `runner` | `test/wish.test.ts` — `host embedding contract: profile wish targets` |
| 2 | `runtimeContext` / `spaceContext` | §2 | `ui` | `src/v2/runtime-context.test.ts` |
| 3 | Navigation events | §3 | `shell` | `test/navigate-contract.test.ts` |
| 4 | `getCfcLabel` egress check | §4 | `ui` | `src/v2/core/cfc-label.test.ts` — `cfcLabelViewIsPublic (egress check)` |
| 5 | Guarded-define idiom | §5 | `ui` | `src/v2/components/host-embedding-guarded-define.test.ts` |
| 6 | Trusted-mark threat model | §6 | `runner` | `test/cfc-ui-contract.test.ts` — `host embedding contract: trusted-mark threat model` |
| 7 | Pinning is owner-gated, not gesture-gated | §7 | `patterns` | `system/profile-home.owner-gated.test.ts` |

Notes:
- **Seam 1** documents current `#profile` resolution and explicitly marks the resolution-order subsection **pending CT-1829** (single-result semantics under review), so the doc does not pre-decide it.
[...]

PR #4512, opened 2026-07-03
feat(runner): #profile wish resolves single result; picker becomes switcher (CT-1829)

bfollington:
## Why

The orphan-second-profile deadlock: a labs-shell tab can create a second profile without setting a default. In that 2+/no-default case, the interactive `#profile` wish replaced its output with the profile-picker sidecar pattern's **initially-empty** result cell — so `.result` was `undefined` for the whole sidecar fetch/compile window, and `undefined` **forever** if the fetch failed (the fetch chain had no failure surfacing; the sidecar cache swallows errors). Every consumer gating on `.result` (e.g. profile-group-chat's send guard) deadlocked, and headless/interactive answers diverged.

PR #4415 fixed the common case (2+ profiles **with** a default) and explicitly left this residual open. The design decision is recorded on Linear [CT-1829](https://linear.app/common-tools/issue/CT-1829) — **Option A: `wish("#profile").result` is always single; the picker becomes a switcher** — and is the contract Loom binds to (https://github.com/commontoolsinc/loom/pull/3627).

## What

- **`packages/runner/src/builtins/wish.ts`**
  - The `#profile` interactive multi-candidate branch no longer sends the picker sidecar's result cell as the wish output. It now sends the unified wish state eagerly — `result: ordered[0]` (candidates are already ordered default → MRU → first by `getProfileCandidateCells`), `candidates`, and `[UI]` = the picker sidecar rendered via `cf-render` (same idiom as `profileCreateUI`) — mirroring what the generic multi-match path already does.
  - **Switcher semantics:** the picker's existing "Use" (MRU stamp) and "Set default" trusted writes reorder the candidates, so `ordered[0]` — and therefore `.result` — flips reactively. Selection is state, not a channel.
  - Picker sidecar fetch failure now surfaces an error UI in the picker cell instead of leaving the slot blank forever (the sidecar cache resolves failed fetches to `undefined`; a defensive `.catch` also covers genuine rejections). `.result` is unaffected — it rides the main wish state.
  - Headless and single-candidate paths are untouched (they already returned `ordered[0]`).

- **`packages/runner/test/wish.test.ts`** — new `single-result #profile contract (CT-1829)` describe block (independent of the host-embedding-contract block landing in #4502): 0 profiles (create surface, error, undefined result), 1 profile (interactive + headless), 2+ with valid default (interactive + headless), 2+ no default → MRU head **interactive** (the new load-bearing case) + headless, 2+ no default no MRU → first (both modes), invalid default link skipped, picker rendered as `[UI]` with `.result` populated, MRU write flips result, and sidecar-fetch-failure resilience (`.result` still resolves; error surfaced in picker UI).

- **`docs/common/conventions/wish.md`** — "Well-Known Profile Targets" rewritten to the decided semantics (`.result` always the single current profile in every mode; picker = switching affordance; selection = MRU/default write → result updates). The general "multiple matches show a picker" section gains an explicit profile-exception note.

## Consumer check

All 13 `#profile` consumers surveyed; none need changes. Spot-checked the two most load-bearing:
- **profile-group-chat** send guard (`profileWish.result !== undefined`) — now enables in the orphan case instead of deadlocking.
- **group-chat-lobby** `{profileWish[UI]}` — still shows the picker when ambiguous (now via the `[UI]` slot).

## Follow-ups

- `profile-picker.tsx` keeps its own vestigial `result`/`resultIndex` computeds for one release; delete them once this lands and bakes (the builtin is the single source of ordering).
- `docs/development/HOST_EMBEDDING.md` (PR #4502, unmerged) carries a "pending CT-1829" marker — deliberately not touched here; it resolves in a follow-up once both PRs merge.
- Option C (generalizing single-best-by-default to all wishes, picker opt-in) is a separate future issue.

## Verification

- `deno test packages/runner/test/wish.test.ts` — 104 steps, 0 failed.
- Full `packages/runner` suite (`deno task test`) — 758 passed (3759 steps), 0 failed.
- `deno fmt` / `deno lint` / `deno task check` clean.

---
## Summary by cubic
Make `wish("#profile").result` always resolve to a single current profile (default → MRU → first) in every mode. The picker now lives in `[UI]` as a switcher that updates `.result` via MRU/default writes, fixing the orphan second‑profile deadlock and aligning with CT-1829.

- **New Features**
  - `packages/runner/src/builtins/wish.ts`: Interactive 2+ profiles with no default send unified state early with `result = ordered[0]`, `candidates`, and `[UI]` rendering the profile picker via `cf-render`. Headless and single-candidate paths are unchanged.
  - Switcher semantics and docs: `docs/common/conventions/wish.md` documents the profile exception (`.result` is always single; picker is the switching affordance). Updated `packages/patterns/system/profile-picker.tsx` header comment to reflect switcher-only behavior and mark `result`/`resultIndex` as vestigial.

- **Bug Fixes**
  - `.result` no longer goes undefined during picker load or on picker fetch failure; errors are surfaced inside the picker cell instead.
  - `packages/runner/test/wish.test.ts`: Adds a suite that pins the new contract across 0/1/2+ profiles, default/MRU/first, invalid default, MRU flip, and sidecar fetch failure.

<sup>Written for commit 945056faaac228a917b524e730564a960abf58de. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4512?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/common/conventions/wish.md
+++ b/docs/common/conventions/wish.md
@@ -46,6 +46,15 @@ When multiple pieces match, a picker UI is rendered via `wishResult[UI]`. The
 user browses candidates and clicks "Confirm Selection". Until confirmed,
 `result` reactively reflects the currently highlighted candidate.
 
+> **Exception — well-known profile targets.** `wish({ query: "#profile" })` does
+> _not_ follow the "result reflects the highlighted candidate" rule. Its
+> `.result` is **always the single current profile** (default → MRU → first) in
+> every mode; the picker there is only a switching affordance, and selecting a
+> profile changes `.result` by reordering candidates (MRU/default writes), not by
+> a confirm gesture. See [Well-Known Profile Targets](#well-known-profile-targets)
+> (CT-1829). Generalizing this "single-best by default; picker opt-in" shape to
+> all wishes is a future step.
+
 You can render the built-in UI directly:
 
 ```tsx
@@ -106,26 +115,41 @@ wish({ query: "#portfolio", scope: ["profile"] })
 A user may have multiple profiles, stored on the home default pattern at
 `homeSpaceCell.defaultPattern.profiles` (a list), with `defaultProfile` and a
 recency-ordered `mru`. The well-known wishes enumerate that list and resolve,
-ordered **default first, then by MRU**:
+ordered **default first, then by MRU, then list order**:
 
 ```tsx
 // Shown at module scope.
-wish({ query: "#profile" }) // the default profile (headless); see [UI] below
+wish({ query: "#profile" }) // the single current profile; see [UI] below
 wish({ query: "#profileName" }) // default profile's initialNameApplied
 wish({ query: "#profileAvatar" }) // default profile's avatar
 wish({ query: "#profileBio" }) // default profile's bio (free-text description)
 wish({ query: "#profileSpace" }) // default profile's space cell
 ```
 
-Headless / single-profile callers get the default profile. The optional `[UI]`
+`wish({ query: "#profile" }).result` is **always the single current profile** —
+the best of the ordered candidates (default → MRU → first) — in **every** mode
+(interactive, headless, and the blessed read). It is never `undefined` while a
+profile exists, and it never depends on the picker sidecar pattern running, so
+consumers can gate on `.result` without stranding in the multi-profile case
+(CT-1829). The `candidates` array holds all ordered profiles.
+
+The picker is the **switching affordance**, not the source of `.result`:
+selection is _state_, not a channel. When the picker's "Use" writes `mru` or
+"Set default" writes `defaultProfile`, it reorders the candidates, so
+`ordered[0]` — and therefore `.result` — updates reactively. The optional `[UI]`
 for `wish({ query: "#profile" })`:
 
 - **0 profiles:** the trusted profile-create surface (same input as the home
   Profile tab). Submitting a name creates the viewer's first profile and leaves
-  the current view in place; the wish reacts once the link exists.
+  the current view in place; the wish reacts once the link exists. `.result` is
+  `undefined` and `error` is set until the first profile exists.
 - **1 profile:** a link to that profile.
-- **2+ profiles:** the **profile picker** (`profile-picker.tsx`) — lists
-  profiles, selects the default, stamps MRU, and creates more inline.
+- **2+ profiles, a valid default set:** a link to the default profile.
+- **2+ profiles, no valid default:** the **profile picker**
+  (`profile-picker.tsx`) — lists profiles, sets the default, stamps MRU, and
+  creates more inline — rendered purely as the switcher. `.result` still
+  resolves to the ordered best (MRU-or-first) immediately; picking a profile is
+  what changes it.
 
 When rendering profile data from a shared piece, use a user-scoped result schema
 for the rendered output so each viewer sees their own home profile projection.
--- a/packages/patterns/system/profile-picker.tsx
+++ b/packages/patterns/system/profile-picker.tsx
@@ -20,17 +20,22 @@ import ProfileCreate, {
 } from "./profile-create.tsx";
 import type { ProfileHomeOutput } from "./profile-home.tsx";
 
-// The profile picker launched by the #profile wish when the user has more than
-// one profile. It renders each profile natively (name + avatar + a link),
-// offers an inline "create another" affordance, lets the user pick a default,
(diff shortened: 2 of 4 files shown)


bfollington (commented):
**cf-review via Claude (Opus 4.8), on behalf of @bfollington** — automated review, human owns the merge.

Scope: runtime-semantics change to `wish("#profile")` (`packages/runner/src/builtins/wish.ts`) + docs + tests. Coverage: deep-read the changed region and the ordering/short-circuit helpers, full read of the new test block and `wish.md`, delegated a read-only 14-consumer survey. Checks run: `deno test packages/runner/test/wish.test.ts` — 4 tests / 104 steps pass; confirmed the 13 new CT-1829 steps executed and are non-vacuous.

**The change is coherent and well-reasoned.** `uniqueResultCells[0]` genuinely is `ordered[0]` (default → MRU → first) because a valid default short-circuits to a single base resolution upstream (`wish.ts:825-826`), so the only case reaching the new interactive branch is "2+ candidates, no valid default" — as the comment claims. Reactivity is sound: the action reads `mru`/`defaultProfile` via `getProfileCandidateCells().get()`, registering subscriptions, so an MRU/default write re-runs the action and `ordered[0]` flips. The fetch-failure branch commits its error UI on its own independent `runtime.edit()` tx (same idiom as the create-pattern path) — no tx-lifecycle hazard. Consumer survey: none of the 14 `#profile` consumers relied on the old "result undefined until confirm" behavior; all either gate on `.result !== undefined` (now unblocks sooner) or read `[UI]`/field targets — safe or improving across the board.

**1. Stale header comment in `profile-picker.tsx` contradicts the new contract** — `packages/patterns/system/profile-picker.tsx:23-33` (not in this diff)
The header still states, present-tense and un-caveated: *"The chosen profile flows back as the wish `result`: result = default ⟶ else MRU ⟶ else first."* Under CT-1829 the picker no longer owns `.result` — the builtin's `ordered[0]` on the main wish state does; the picker is only the `[UI]` switcher. This is the exact file the PR body elevates ("the builtin is the single source of ordering"), so a reader (Loom especially) opening it is actively misled — it fails the coherence bar against the now-authoritative `wish.md`. Deferring deletion of the vestigial `result`/`resultIndex` computeds "for one release" is fine, but the *comment* asserting the old semantics should be corrected now (a few lines), not left to bake.
Fix: reword to note these computeds are vestigial under CT-1829, the #profile builtin owns `.result` (ordered[0]), and the picker only writes MRU/default.

**2. The switcher test proves ordering, not the in-place reactive re-run** — `wish.test.ts`, "MRU write flips result"
The test writes MRU then spins up a *fresh* wish (`wishPattern2`/`rc2`) and asserts the new head — its own comment concedes "here we re-resolve to assert the ordering contract the write feeds." So it pins that resolution reads MRU correctly, but does not independently verify the *original* live wish action re-runs and flips `.result` in place — the load-bearing claim of the PR. Behavior is very likely correct (subscriptions register during resolution), but the test sidesteps it.
Fix (optional): re-`pull()` the *original* result cell after the MRU write and assert it flipped, instead of resolving a second wish.

## ⚪ Nits

- `docs/specs/shared-profile-rosters.md:273` — comment "or the picker result if they have >= 2" is now inaccurate, though historically framed ("Under PR #3830"). Boy-Scout follow-up.
- The create-pattern path (`wish.ts:1693-1701`) still lacks the fetch-failure error-surfacing + `.catch` this PR adds to the picker, so the create sidecar keeps the "blank forever on fetch failure" bug this PR fixes for the picker. Not a regression here; natural symmetric follow-up.
- Fetch-failure test uses `setTimeout` polling (bounded, idles between) — mild flakiness surface per the repo's time-based-wait guidance; acceptable for polling an async commit with a deadline.

## Verdict
**Non-blocking.** Semantics correct, tests genuinely pin the contract (interactive MRU/first cases would read `undefined` under the old behavior — they assert the concrete name, so eager-send is proven), consumers safe. Recommend fixing finding #1 (stale picker header) before merge since it is a direct contradiction in the file the PR itself makes canonical — but it is a comment, not behavior, and can be consciously accepted.

PR #4516, opened 2026-07-03
docs(cfc): four accuracy corrections to the enforcement-matrix doc

seefeldb:
Corrects [docs/specs/cfc-enforcement-matrix.md](../blob/main/docs/specs/cfc-enforcement-matrix.md) (added #4487, amended #4489) so the H4 code PR implements against the real contracts. Every claim was verified against code; docs and comments only, no behavior changes.

## 1. Missing-policy already rejects under `enforce-explicit`

The doc said a labeled write with no resolvable schema/policy input is "flagged, not rejected" under explicit, rejecting only under strict. In reality explicit already rejects:

- `cfc/prepare.ts` records the `missing schema write-policy input for <id>` reason (~3524),
- the enforcement ladder (`storage/extended-storage-transaction.ts` ~1030–1047) rejects **any** reasoned/non-prepared relevant tx under both enforcing modes,
- `packages/runner/test/cfc-boundary.test.ts` (~2393–2417) asserts the reject **in explicit mode**.

Per owner direction the doc is corrected (not the code — softening it would weaken the shipped shell posture, anti-fail-closed). The documented strict-vs-explicit delta is re-scoped to the writer-fit reject (SC-18b, which genuinely does not exist in code yet) plus future strict-only fail-closed cases, and §4 now explicitly warns H4 not to move the missing-policy check behind the strict gate. The same claim in `docs/plans/cfc-future-work-implementation.md` stage H4 is fixed too.

## 2. The H5 dial joins the matrix

`cfcTriggerReadGating` (boolean, default `false`; `runtime.ts:519`, `cfc/types.ts:357`) is a live dial that #4488 said was "folded into the H4 matrix" but never was. Added to the dial inventory (now 4 × 3 × 3 × 2), the rollout ordering (new constraint #4), the conforming-states table, and the plan doc's dial register — including the one-hop-only caveat: trigger gating closes only the direct trigger channel (a triggered handler can evade via an unlabeled intermediary doc whose write schedules a second, egressing run); multi-hop closure requires `cfcFlowLabels: persist` stamping the intermediate — same shape as the write floor's "complete once flow persists" entry. Verified: trigger reads feed the `requiredIntegrity` gate and the sink-ceiling consumed set only under the flag (`prepare.ts` `triggerReadSources`, consumed at ~2498 and ~3127).

## 3. Toolshed is not `disabled/off/off`

`packages/toolshed/index.ts` constructs `Runtime` with **no** CFC options, so toolshed inherits the constructor defaults: `enforce-explicit` + flow `off` (`runtime.ts:515-519`). The row now documents the actual posture, and "Operator (explicitly disabled)" remains as the state that requires explicitly passing the dial. Stale comments claiming toolshed runs CFC-disabled are reconciled in the same change:

- `runner/src/runtime.ts` (~919, the disabled-mode ingest-mint carve-out),
- `toolshed/lib/custody-ingest.test.ts` (~27 — the test runs the *default* runtime, which is enforce-explicit, not disabled),
- `runner/test/cfc-external-ingest.test.ts` (~38, same claim found in the sweep),
- `docs/development/proposals/vouched-ingest-channel-mint-design.md` ("Toolshed defaults `DEFAULT_CFC_ENFORCEMENT_MODE = 'disabled'`" — that constant is the bare-transaction fallback, not toolshed's effective mode).

## 4. background-piece-service host row

`packages/background-piece-service/src/main.ts` (~44–56) also passes no CFC options and inherits `enforce-explicit` + flow `off` — a combination that wasn't among the named conforming rows. Both server hosts now share a "Server hosts today" row (conforming: explicit checks consume no derived labels).

Touched test files pass (`cfc-external-ingest.test.ts`, `custody-ingest.test.ts`); comment-only changes there.

---
## Summary by cubic
Corrects the CFC enforcement matrix to match shipped behavior and adds the missing `cfcTriggerReadGating` dial. Clarifies host defaults and pins a flaky scheduler coverage gap; runtime behavior is unchanged.

- **Bug Fixes**
  - Documented that missing-policy writes already reject under `enforce-explicit`; fixed the matrix and plan and warned not to move this behind strict.
  - Added `cfcTriggerReadGating` (default `false`) to the dials, rollout order, and conforming states; noted the one-hop-only caveat until `cfcFlowLabels: "persist"`.
  - Fixed host rows: toolshed and background-piece-service pass no CFC options and thus run `enforce-explicit` + flow `off`; clarified that “Operator (explicitly disabled)” requires opting into `cfcEnforcementMode: "disabled"`.
  - Reconciled stale comments in `packages/runner/src/runtime.ts`, `packages/runner/test/cfc-external-ingest.test.ts`, `packages/toolshed/lib/custody-ingest.test.ts`, and `docs` (including `vouched-ingest-channel-mint-design.md`).
  - Tests: stabilized scheduler coverage by deterministically tracing a memo-hit in pull-mode dirty-dependencies.

<sup>Written for commit a29622b1b270e212f74d9704e8676b7e1b2e9d01. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4516?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seefeldb:
The Performance Check coverage-debt gate flagged +5 uncovered lines in `packages/runner` on the first run. Since this PR's runner changes are comment-only, I diffed this run's LCOV artifacts against the latest green main run: the entire +5 is `scheduler/dirty-dependencies.ts:93-98` — the memo-hit trace bookkeeping, untouched by this PR. Those lines execute only when a pull-event preflight trace happens to be active during a memo hit (event-batch interleaving), so their coverage flaps run to run.

Rather than a baseline override, dd936c67e pins them: the existing shared-memo diamond test in `scheduler-pull.test.ts` now re-collects from the already-memoized dirty computation with a trace context attached and asserts `memoHitCount`/`workSetAddCount`. Local coverage confirms lines 93–98 flip to covered (the trace-body lines get exactly the 1 hit from the new block). Same play as the compiled-bundle-verifier pin in #4481.

Issue #4527, opened 2026-07-04
fabric-safe traversal: prepared PRs for danfuzz's return


mathpirate:
Dan — while you were out we (Gideon + Claude) swept all **52 `TODO(danfuzz)` markers** on main, classified them against your own marker language and commit precedents, and prepared a short series of draft PRs for the subset your text already defines. Everything else is listed below as deliberately untouched, with what we learned attached. Nothing is merged; every draft is yours to amend or decline freely.

Inventory basis: `origin/main` @ `7cda816c6` (2026-07-02); file:line references below are against that SHA (each PR re-located its markers by text on its own branch). Classification: **13 markers → 4 draft PRs · 6 hygiene-only (your call) · 10 need your design decision · 23 deferred by their own text.**

## The series (suggested review order — one decision per PR)

- [x] **A — data-model: `deepClone` base template-method** (`fabric-safety/deepclone-template`) — **draft: #4528**
      _One decision: is this the `shallowClone`/`shallowUnfrozenClone` split, mirrored?_ Markers: `interface.ts:117`, `BaseFabricInstance.ts:10`. Identity-opt lifted from `FabricLink.deepClone` (`isDeepFrozen`, not `Object.isFrozen`). (Correction: see findings below — 2 real cores collapse; 3 throwing stubs relocate.)
- [x] **B — runner: retire the marked live `deepEqual` sites → `valueEqual`** (`fabric-safety/valueequal-tail`) — **draft: #4529**
      _One decision: is each gate comparing stored-form values?_ Markers: `scheduler/topology.ts:44`, `storage/transaction/attestation.ts:152`, `cfc/prepare.ts:1996`, `path-utils.ts:18`, + stale-doc rider on `utils/deep-equal.ts:15`. Two callouts: the attestation swap **un-masks** previously hidden `StateInconsistency` (we believe desired); the `setValueAtPath` swap changes an **exported API's** observable behavior (non-Fabric class instances now throw) — isolated as the final commit, drop it freely.
- [x] **C — guard rebuild-walks against `FabricPrimitive` decomposition** (`fabric-safety/primitive-rebuild-guards`) — **draft: #4530**
[...]

PR #4528, opened 2026-07-04
feat(data-model): hoist deepClone() into a BaseFabricInstance template ([DEEP_CLONE_CORE])

mathpirate:
Grows the base `deepClone()` implementation the two `TODO(danfuzz)` markers called for: a concrete `deepClone(frozen)` template method plus a `protected abstract deepUnfrozenClone()` sibling on `BaseFabricInstance`, mirroring the existing `shallowClone()`/`shallowUnfrozenClone()` template-method split. The two real implementors (`FabricLink`, `FabricError`) already had this exact shape by hand; they now supply only their per-class deep-copy core and stop implementing the public `deepClone()` directly. The public `FabricInstance.deepClone(frozen): FabricInstance` protocol surface is unchanged.

### Markers resolved

`packages/data-model/src/interface.ts`:

> TODO(danfuzz): This method should grow a base implementation on `BaseFabricInstance` which defers to a `protected abstract` sibling, mirroring the `shallowClone()`/`shallowUnfrozenClone()` split.

`packages/data-model/src/fabric-instances/BaseFabricInstance.ts`:

> TODO(danfuzz): `deepClone()` should grow a base implementation here that defers to a sibling `protected abstract` method (mirroring the `shallowClone()`/`shallowUnfrozenClone()` template-method split), at which point individual subclasses stop implementing `deepClone()` directly.

Both are removed; the `deepClone` declaration in `interface.ts` is now documented like `shallowClone`'s.

### Shape

- **Mirrors `shallowClone`:** template body is `if (frozen && isDeepFrozen(this)) return this; const copy = this.deepUnfrozenClone(); return frozen ? deepFreeze(copy) : copy;` — structurally identical to `shallowClone`.
- **`isDeepFrozen` vs `Object.isFrozen`:** the deep-clone identity optimization gates on `isDeepFrozen(this)`, not `Object.isFrozen(this)`. A shallowly-frozen instance with still-mutable nested `FabricValue`s is not safe to alias as a deep clone. Documented once on the base template; a test (`does NOT identity-return a merely-shallowly-frozen instance`) pins it.
- **Naming flag:** `deepUnfrozenClone` is my analogy to `shallowUnfrozenClone` — **not** a name from your marker. Rename freely if you prefer (`deepCloneCore`, `deepThawedClone`, …).
- **Behavior:** equivalent for `FabricError` on every path, and for all throwing stubs (`FabricNativeWrapper`, `UnknownValue`, `ProblematicValue`, and inheriting `FabricSet`/`FabricMap` — their `deepClone()` still throws identically; `cloneIfNecessary.test.ts`'s `deepCloneImplemented` matrix stays green untouched). **One deliberate exception, pinned by a test:** for `FabricLink.deepClone(true)`, already-deep-frozen nested payload subtrees (e.g. a frozen `schema`) are now **re-cloned rather than reference-shared**. The result is value-equal and deep-frozen, but structural sharing — and any identity/cache warmth — is lost. The old core cloned with `{ frozen: true }` (identity-passing deep-frozen subtrees); the template's no-argument unfrozen core must clone with `{ frozen: false }` (a force-copy) before the template freezes.

### Verification

Pure refactor, so a red-before-green test is N/A. Instead, extended template-contract coverage: `BaseFabricInstance.test.ts` gains a full `deepClone()` block (identity when frozen && already-deep-frozen; fresh instance otherwise; frozen=true ⇒ deep-frozen result; frozen=false ⇒ mutable clone with no shared nested structure; the shallow-vs-deep identity distinction), `FabricError.test.ts` gains its first-ever `deepClone()` coverage, and `FabricLink.test.ts` pins the structural-sharing change described above. Gates on the branch: `deno fmt --check`, `deno lint`, `deno check` all clean; full `data-model` suite `47 passed (1936 steps) | 0 failed`. No other package touched.

### Needs Dan's judgment

1. **`deepUnfrozenClone` name** — mine by analogy; yours to rename.
2. **The marker's template shape vs. the `deepClone(true)` "maximal structural sharing" contract.** A no-argument unfrozen core structurally cannot honor `interface.ts`'s promise of maximal structural sharing on the frozen path: the core must produce a mutable copy, so already-deep-frozen subtrees get force-copied and re-frozen instead of shared. I left your contract sentence in `interface.ts` untouched and pinned the current copy-then-freeze behavior with a test (`frozen clone currently RE-CLONES an already-deep-frozen payload subtree (was shared)`). Options: bless copy-then-freeze and weaken that doc sentence, or evolve the core shape (e.g. a frozen-aware core such as `deepCloneCore(frozen)` that can share safe subtrees). Your call — behavior is pinned either way.
3. **Pre-existing `FabricError` deep-independence gap** (surfaced, not introduced): `FabricError.deepClone(false)` round-trips through `[CODEC]`, whose `encode()` passes `cause`/extras by reference, so the unfrozen clone still **shares** its nested `cause` with the original — contrary to the documented `deepClone(false)` "no visible shared reference structure" contract. Behavior is identical to the pre-refactor code; the test `mutable clone currently SHARES the nested cause reference (known gap)` and a note on the core's comment pin the actual behavior to record it, not to bless it. Worth a follow-up (deep-copy `cause`/extras in the clone core).
4. **Degenerate-case behavior change for the throwing stubs (nuance):** moving the throw into `deepUnfrozenClone()` means the template's identity check runs first, so `deepClone(true)` on an *already-deep-frozen* stub would now return `this` instead of throwing. This path is unreachable via `cloneIfNecessary` (its `canReturnAsIs` short-circuits deep+frozen+already-deep-frozen values upstream; the only non-test `deepClone` caller is `value-clone.ts`'s `cloneHelper`) and no test exercises it, so nothing observably changes today; I deliberately did **not** add a test pinning it, since whether an immutable stub should identity-return vs. throw is your call. Flagging in case you want it to keep throwing unconditionally.

---

---
## Summary by cubic
Centralized deep cloning in `BaseFabricInstance` with a template `deepClone(frozen)` that defers to `[DEEP_CLONE_CORE](frozen)`, preserving maximal structural sharing on frozen clones and gating identity on `isDeepFrozen(this)`. No public API changes; part of fabric-safety (#4527).

- **Refactors**
  - Added `[DEEP_CLONE_CORE](frozen)`; `deepClone()` now lives on `BaseFabricInstance`, mirroring `shallowClone()`/`[SHALLOW_UNFROZEN_CLONE]`.
  - Identity optimization checks `isDeepFrozen(this)` (not `Object.isFrozen(this)`).
  - `FabricLink` and `FabricError` now implement `[DEEP_CLONE_CORE]`; for `FabricLink`, frozen clones identity-share already-deep-frozen payload subtrees.
  - Moved throwing deep-clone stubs to `[DEEP_CLONE_CORE]` in `FabricNativeWrapper`, `UnknownValue`, `ProblematicValue` (behavior unchanged; `deepClone(true)` on an already-deep-frozen stub identity-returns before the throw).
  - Known gap (pre-existing): `FabricError.deepClone(false)` still shares `cause`/extras due to the codec; tests pin this.
  - Updated `interface.ts` docs and expanded tests; all `data-model` tests pass.

<sup>Written for commit 7baa9d5d76568813a73071c70e36b77984a9db30. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4528?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/data-model/src/fabric-instances/BaseFabricInstance.ts
+++ b/packages/data-model/src/fabric-instances/BaseFabricInstance.ts
@@ -1,4 +1,9 @@
 import { FabricInstance, type FabricValue } from "@/interface.ts";
+// Used only inside method bodies: this import participates in a module cycle
+// with `deep-freeze.ts` (which imports this module's symbols and class for its
+// generic dispatch), which is safe for call-time function use but must not be
+// dereferenced during module evaluation.
+import { deepFreeze, isDeepFrozen } from "@/deep-freeze.ts";
 
 /**
  * Well-known symbol for deeply freezing a fabric instance in place. The method
@@ -45,17 +50,29 @@ export const SHALLOW_UNFROZEN_CLONE: unique symbol = Symbol.for(
 );
 
 /**
+ * Well-known symbol for producing a new deep clone of a fabric instance,
+ * cloning nested `FabricValue`s to the requested frozenness. This is the
+ * `protected` template-method primitive that the concrete `deepClone()` calls
+ * when a fresh instance is needed; each concrete subclass implements it.
+ * Symbol-keyed as implementation plumbing (matching
+ * `[SHALLOW_UNFROZEN_CLONE]`), while `deepClone()` stays a regular-named
+ * client method. Unlike the shallow core, this core takes the `frozen` intent:
+ * a deep clone's nested handling depends on it (already-deep-frozen subtrees
+ * may be identity-shared when `frozen` is `true` -- the "maximal structural
+ * sharing" the `deepClone()` contract promises -- but must be force-copied
+ * when `frozen` is `false`).
+ */
+export const DEEP_CLONE_CORE: unique symbol = Symbol.for(
+  "data-model.deepCloneCore",
+);
+
+/**
  * Abstract base class providing shared scaffolding for `FabricInstance`
  * subclasses. Concrete `FabricInstance` classes extend this, not
  * `FabricInstance` directly: `FabricInstance` is the pure abstract protocol
  * (the `instanceof`-able contract that external code is written against), while
  * `BaseFabricInstance` is where shared template-method implementations and the
  * freeze-protocol plumbing (`[DEEP_FREEZE]()`, `[IS_DEEP_FROZEN]()`) live.
- *
- * TODO(danfuzz): `deepClone()` should grow a base implementation here that
- * defers to a sibling `protected abstract` method (mirroring the
- * `shallowClone()`/`[SHALLOW_UNFROZEN_CLONE]()` template-method split), at which
- * point individual subclasses stop implementing `deepClone()` directly.
  */
 export abstract class BaseFabricInstance extends FabricInstance {
   //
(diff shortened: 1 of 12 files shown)


danfuzz:
First and foremost thanks for taking a stab at this.

Unfortunately, the TODOs and the pre-existing structure don't tell the full story about the direction this code is headed in:

* As already noted in the "judgment" section, having the per-concrete-class implementation method produce an unfrozen result is problematic.
* The method naming is inconsistent with the pattern of using regular method names for "public client usage" and unique symbols for "underlying implementation." (This pattern only emerged clearly after the `shallow*()` stuff was first defined.)
* Related to that last point, I'm also in the middle of rearranging how the implementation methods get declared (to keep them out of the API surface for clients that aren't trying to define new `FabricSpecialObject` classes).

My intention is to address the second and third bullets above — relatively soon — before trying to add new methods into the mix. Once that's done, it'll be worth either iterating on this PR (either per se or using it as raw material for a follow-on).

danfuzz:
Over the last few days, I made some progress on the effort I mentioned earlier. To the extent that you want to maintain the merge-viability of this PR, now would be a good time to synch it with main, and then resolve the conflicts both in terms of git mechanics and in terms of the system design.

mathpirate:
Reworked from scratch on top of #4817/#4818 — with those landed, the marker pair read as an exact spec. One deliberate divergence from the shallow split, called out in the body: the deep core takes the `frozen` intent (`[DEEP_CLONE_CORE](frozen)`), which is what lets `FabricLink` keep the contract's maximal structural sharing — the v1 thread's judgment item, now dissolved rather than pinned. Draft as before; ready whenever you want to look.

PR #4529, opened 2026-07-04
fix(runner): finish the fabric-safety deepEqual -> valueEqual swaps (CT-1770 tail)

mathpirate:
Completes the tail of the fabric-safety series: the four remaining marked `deepEqual` equality gates over `FabricValue`s are swapped to `valueEqual`, the `data-model`-aware content-hash equality. `deepEqual` compares class instances by enumerable own-props, and a `FabricPrimitive` (`FabricBytes`/`FabricHash`/`FabricEpoch*`/...) keeps its state in private `#fields` with zero own-props — so `deepEqual` wrongly reports every distinct same-class instance as equal, masking real value changes in no-op / change-detection / consistency gates. `deepEqual`'s legitimate non-Fabric uses are left in place; its doc comment is refreshed since `valueEqual` is now itself `Fabric`-aware.

### Markers resolved (each removed)

- topology.ts: `` TODO(danfuzz): `mapsEqual` compares map values with `deepEqual`, which mishandles `FabricValue`; this helper is used on stored read-value maps, so two same-class `FabricPrimitive`s (state in private `#fields`, zero own-props) compare equal regardless of value. Use a Fabric-aware equality. ``
- attestation.ts: `` TODO(danfuzz): This compares a stored document value (the read/attested value) with `deepEqual`, which mishandles `FabricValue`: two same-class `FabricPrimitive` values (state in private `#fields`, zero own-props) compare equal regardless of value, so a changed Fabric value can be mis-detected as unchanged. Use a Fabric-aware equality for stored-value comparison. ``
- prepare.ts: `` TODO(danfuzz): `deepEqual` mishandles `FabricValue` (see `utils/deep-equal.ts`); `schema.default` can hold a `FabricValue`, so this CFC write-policy check can compare wrong. Migrate to a `Fabric`-aware equality once available. ``
- path-utils.ts: `` TODO(danfuzz): This no-op write gate compares the existing value against the new value with `deepEqual`, which mishandles `FabricValue` (same-class `FabricPrimitive`s compare equal regardless of value, since their state lives in private `#fields` with zero own-props), so a real Fabric-value change can be dropped as a no-op. Use a Fabric-aware equality. ``
- deep-equal.ts stale doc: `` Use a `data-model`-aware equality (`valueEqual`, once it is itself made `Fabric`-aware) for any `FabricValue` comparison. `` → refreshed.

### Per-site operand arguments (both operands are stored-form `FabricValue`s)

- **topology `mapsEqual`**: params tightened `Map<string, unknown>` → `Map<string, FabricValue>` to match the sole caller (`diagnosis.ts`'s `DiagnosisRecord.readValues`, already `Map<string, FabricValue>`); no cast. (The maps' sentinel `"[read-error]"` strings and `undefined` values are both handled by `valueEqual` without throwing.)
- **attestation `claim`**: `expected` is `IAttestation.value: Immutable<FabricValue>`; `actual` is `FabricValue`/`Immutable<FabricValue>` from `toTransactionDocumentValue` or `read().ok.value`; no cast.
- **prepare `writeInstallsInitialSchemaDefault`**: written side is `FabricValue`; `schema.default` is statically `ImmutableJSONValue` but interns a native `Uint8Array`/`Date` default to a `FabricPrimitive` at runtime (verified both operands arrive as `FabricBytes` at the gate); documented cast bridges the deeply-`readonly` JSON type.
- **path-utils `setValueAtPath`**: operands are `any`; both are `FabricValue` by convention (see droppable-commit note).

### ⚠️ Behavior notes

**The attestation swap UN-MASKS previously hidden `StorageTransactionInconsistent` errors.** Where a stored Fabric value genuinely changed but differed only in `FabricPrimitive` content, `claim` previously returned `{ ok }` (silent masking); it now correctly returns the `StateInconsistency` error the check exists to raise. This is the exact bug the marker names and is the desired behavior — but it is a behavior change that can surface inconsistencies that were being swallowed.

**`valueEqual` throw-semantics:** unlike `deepEqual` (which reports non-`Fabric` class instances and functions as merely unequal), `valueEqual` **throws** on a function or a non-`Fabric` class instance. Every operand at these four sites is argued to be a stored-form `FabricValue`, so this is not expected to fire — but it is a real semantic difference at each gate, most notably at the public `setValueAtPath` boundary.

**The final commit (`setValueAtPath`) is droppable — Dan's call.** `setValueAtPath` is re-exported from `runner/src/index.ts`, so the throw-on-non-Fabric semantics reach the runner's public surface. It is isolated as the last commit (marked `!`) so it can be dropped if the public-surface risk should wait.

### Verification

Red-first per site — a test using two distinct same-class `FabricBytes` (differing only in byte content) fails against the pre-swap code and passes after: `mapsEqual` → false; `claim` → `StateInconsistency`; the CFC write-policy check → rejection; `setValueAtPath` → writes and returns true. Gates: `deno fmt --check` + `lint` + `check` clean on all touched files (and package-wide); full `packages/runner` suite 769 passed / 0 failed; full `packages/utils` suite 86 passed / 0 failed.

### Needs Dan's judgment

1. Whether to land the droppable `setValueAtPath` public-surface commit now or defer it.
2. Confirmation that un-masking `StateInconsistency` in `attestation.claim` is wanted in this PR (it is the marker's intent, but it can surface real inconsistencies previously swallowed).
3. The prepare.ts `schema.default as FabricValue` cast — kept in the spirit of your diagnosis.ts documented casts, but a cleaner long-term fix might tighten the `default` typing or add a runtime `isFabricValue` guard.

---

---
## Summary by cubic
Finish swapping `deepEqual` → `valueEqual` for all remaining `FabricValue` comparisons in the runner to stop masking real changes in no-op and consistency checks. Completes CT-1770 and may surface inconsistencies that were previously hidden.

- **Bug Fixes**
  - `scheduler/topology.ts`: `mapsEqual` now uses `valueEqual` and is typed `Map<string, FabricValue>`.
  - `storage/transaction/attestation.ts`: `claim` compares via `valueEqual` only; removed the `===` fast path to avoid `+0/-0` conflation and updated docs.
  - `cfc/prepare.ts`: `writeInstallsInitialSchemaDefault` uses `valueEqual` (casts `schema.default` to `FabricValue`).
  - `path-utils.setValueAtPath`: no-op gate uses `valueEqual` (public surface).
  - Docs: `utils/deep-equal` points `FabricValue` comparisons to `valueEqual`; trimmed call-site comments to describe current behavior.
  - Added tests for each site, including `+0/-0` in `claim`, Fabric-aware map equality, CFC default-vs-write, and `setValueAtPath`.

- **Migration**
  - `attestation.claim` can now return `StorageTransactionInconsistent` when a `Fabric` primitive changed only in content; `+0` and `-0` are treated as distinct.
  - `setValueAtPath` now throws on functions or non-`Fabric` class instances; callers should pass `FabricValue`s.

<sup>Written for commit 333e886789e4edff76de0ae72443b39b28f87898. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4529?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/cfc/prepare.ts
+++ b/packages/runner/src/cfc/prepare.ts
@@ -11,6 +11,7 @@ import { emptySchemaObject } from "@commonfabric/data-model/schema-utils";
 import {
   cloneForMutation,
   type CloneForMutationResult,
+  valueEqual,
 } from "@commonfabric/data-model/fabric-value";
 import { isArrayIndexPropertyName } from "@commonfabric/utils/arrays";
 import { deepEqual } from "@commonfabric/utils/deep-equal";
@@ -2601,12 +2602,20 @@ const writeInstallsInitialSchemaDefault = (
     return false;
   }
   const pathTarget = { ...target, path };
-  // TODO(danfuzz): `deepEqual` mishandles `FabricValue` (see
-  // `utils/deep-equal.ts`); `schema.default` can hold a `FabricValue`, so this
-  // CFC write-policy check can compare wrong. Migrate to a `Fabric`-aware
-  // equality once available.
+  // `schema.default` can hold a `FabricValue` (a native `Uint8Array`/`Date`
+  // default interns to `FabricBytes`/`FabricEpoch*`), so this write-policy check
+  // must use `valueEqual`, the `Fabric`-aware content equality: `deepEqual`
+  // walks enumerable own-props, of which a `FabricPrimitive` has none, so it
+  // conflates every distinct same-class instance and would treat a write that
+  // differs from the default only in Fabric content as an initial-default
+  // install (CT-1770). The `default` is statically `ImmutableJSONValue` (a
+  // subset of `FabricValue`); the cast bridges the deeply-`readonly` JSON type
+  // to the `FabricValue` the comparison operates over.
   return previousWriteValueForTarget(tx, pathTarget) === undefined &&
-    deepEqual(writeValueForTarget(tx, pathTarget), schema.default);
+    valueEqual(
+      writeValueForTarget(tx, pathTarget),
+      schema.default as FabricValue,
+    );
 };
 
 const linkedWriteValueForPolicy = (
--- a/packages/runner/src/cfc/prepare.ts
+++ b/packages/runner/src/cfc/prepare.ts
@@ -2602,14 +2602,10 @@ const writeInstallsInitialSchemaDefault = (
     return false;
   }
   const pathTarget = { ...target, path };
-  // `schema.default` can hold a `FabricValue` (a native `Uint8Array`/`Date`
-  // default interns to `FabricBytes`/`FabricEpoch*`), so this write-policy
-  // check compares with `valueEqual`, the `Fabric`-aware content equality that
-  // distinguishes `FabricPrimitive` values by content (state in private
-  // `#fields`, zero enumerable own-props). The cast is required: `default` is
-  // statically `ImmutableJSONValue`, whose deeply-`readonly`,
-  // `ArrayLike`-based JSON shape is not assignable to `FabricValue`, even
-  // though the runtime value is one.
+  // The cast is required: `default` is statically `ImmutableJSONValue`, whose
+  // deeply-`readonly`, `ArrayLike`-based JSON shape is not assignable to
+  // `FabricValue` -- though the runtime value is one (a native
+  // `Uint8Array`/`Date` default interns to a `FabricPrimitive`).
   return previousWriteValueForTarget(tx, pathTarget) === undefined &&
     valueEqual(
       writeValueForTarget(tx, pathTarget),
--- a/packages/runner/src/cfc/prepare.ts
+++ b/packages/runner/src/cfc/prepare.ts
@@ -2603,14 +2603,13 @@ const writeInstallsInitialSchemaDefault = (
   }
   const pathTarget = { ...target, path };
   // `schema.default` can hold a `FabricValue` (a native `Uint8Array`/`Date`
-  // default interns to `FabricBytes`/`FabricEpoch*`), so this write-policy check
-  // must use `valueEqual`, the `Fabric`-aware content equality: `deepEqual`
-  // walks enumerable own-props, of which a `FabricPrimitive` has none, so it
-  // conflates every distinct same-class instance and would treat a write that
-  // differs from the default only in Fabric content as an initial-default
-  // install (CT-1770). The `default` is statically `ImmutableJSONValue` (a
-  // subset of `FabricValue`); the cast bridges the deeply-`readonly` JSON type
-  // to the `FabricValue` the comparison operates over.
+  // default interns to `FabricBytes`/`FabricEpoch*`), so this write-policy
+  // check compares with `valueEqual`, the `Fabric`-aware content equality that
+  // distinguishes `FabricPrimitive` values by content (state in private
+  // `#fields`, zero enumerable own-props). The cast is required: `default` is
+  // statically `ImmutableJSONValue`, whose deeply-`readonly`,
+  // `ArrayLike`-based JSON shape is not assignable to `FabricValue`, even
+  // though the runtime value is one.
   return previousWriteValueForTarget(tx, pathTarget) === undefined &&
     valueEqual(
       writeValueForTarget(tx, pathTarget),
--- a/packages/runner/src/path-utils.ts
+++ b/packages/runner/src/path-utils.ts
@@ -15,12 +15,12 @@ export function setValueAtPath(
     parent = parent[key];
   }
 
-  // This no-op write gate uses `valueEqual`, the `Fabric`-aware content
-  // equality: `deepEqual` walks enumerable own-props, of which a
-  // `FabricPrimitive` has none, so it conflates every distinct same-class
-  // instance and would drop a real Fabric-value change as a no-op (CT-1770).
-  // Note `valueEqual` throws on a function or a non-`Fabric` class instance
-  // rather than reporting it unequal; both operands here are `FabricValue`s.
+  // No-op write gate: the existing and new values are compared with
+  // `valueEqual`, the `Fabric`-aware content equality, so a `FabricPrimitive`
+  // (state in private `#fields`, zero enumerable own-props) is compared by
+  // content rather than by its absent own-props. Note `valueEqual` throws on
+  // a function or a non-`Fabric` class instance rather than reporting it
+  // unequal; both operands here are `FabricValue`s by contract.
   if (valueEqual(parent[path[path.length - 1]], value)) return false;
 
   // We just set the values here. If you need to delete elements from an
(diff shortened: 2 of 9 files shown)


packages/runner/src/cfc/prepare.ts
```diff
@@ -2572,12 +2573,20 @@ const writeInstallsInitialSchemaDefault = (
     return false;
   }
   const pathTarget = { ...target, path };
-  // TODO(danfuzz): `deepEqual` mishandles `FabricValue` (see
-  // `utils/deep-equal.ts`); `schema.default` can hold a `FabricValue`, so this
-  // CFC write-policy check can compare wrong. Migrate to a `Fabric`-aware
-  // equality once available.
+  // `schema.default` can hold a `FabricValue` (a native `Uint8Array`/`Date`
```
danfuzz: I think as a project we've been trying to shy away from using code comments to compare existing vs. former implementation, instead (a) using code comments to describe what's current and correct, and (b) leaving the historical stuff just in commit messages and PR bodies.

Same (implied) suggestion applies to other comments in this PR.

packages/runner/src/storage/transaction/attestation.ts
```diff
@@ -148,14 +148,14 @@ export const claim = (
...
-  // comparison.
-  if (expected === actual || deepEqual(expected, actual)) {
+  // when the replica state hasn't changed since the original read.
+  // Otherwise compare the stored document value (the read/attested value) with
+  // `valueEqual`, the `Fabric`-aware content equality: `deepEqual` walks
+  // enumerable own-props, of which a `FabricPrimitive` (state in private
+  // `#fields`) has none, so it conflates every distinct same-class instance and
+  // would mis-detect a changed Fabric value as unchanged (CT-1770), masking the
+  // very `StateInconsistency` this check exists to raise.
+  if (expected === actual || valueEqual(expected, actual)) {
```
danfuzz: `expected === actual` is a bug here and wouldn't really be a useful optimization anyway.

**Why is it a bug?:** It's a JS footgun. Values of type `number` are valid `FabricValue`s, and we accept the full range of 64-bit IEEE754 floats including both `+0` and `-0`, _and_ JS `===` on numbers is IEEE754 equality which means that `+0 === -0` returns `true`, which is _not_ what you want when treating numbers as generic values.

`valueEqual()` itself does `Object.is(a, b)` right at its start, which is the correct comparison for this situation (`Object.is(+0, -0)` returns `false`). And (per the optimization question) unless we have perf measurements indicating the function call overhead is significant in practice, there's no reason to redundantly make this call as a gate before calling `valueEqual()`.

danfuzz (changes requested):
Looks great, with only one truly-necessary change (which addresses a pre-existing problem).

danfuzz:
On the "judgment" section:

> Whether to land the droppable `setValueAtPath` public-surface commit now or defer it.

I think it's okay to land this.

> Confirmation that un-masking `StateInconsistency` in `attestation.claim` is wanted in this PR (it is the marker's intent, but it can surface real inconsistencies previously swallowed).

This seems right to me, but I'd want confirmation from @seefeldb.

> The prepare.ts `schema.default as FabricValue` cast — kept in the spirit of your diagnosis.ts documented casts, but a cleaner long-term fix might tighten the `default` typing or add a runtime `isFabricValue` guard.

Not sure what "diagnosis.ts documented casts" is referring to, but in any case… Though I've probably been less consistent than I should be, I currently aim to:

* avoid putting casts in the code that TS does not require
* treat required casts as things to be resolved/removed if at all possible by adjusting the types of variable and argument declarations
* treat `as unknown as X` as super-duper to be avoided if at all possible, or documenting why it's necessary on each use
* treat `x: unknown` parameters as super-duper to be avoided as well, because they are a great way to hide type lies that end up biting us.

packages/runner/src/cfc/prepare.ts
mathpirate: agreed thanks very much!

packages/runner/src/storage/transaction/attestation.ts
mathpirate: done thanks

packages/runner/src/path-utils.ts
```diff
@@ -15,12 +15,13 @@ export function setValueAtPath(
...
-  // lives in private `#fields` with zero own-props), so a real Fabric-value
-  // change can be dropped as a no-op. Use a Fabric-aware equality.
-  if (deepEqual(parent[path[path.length - 1]], value)) return false;
+  // No-op write gate: the existing and new values are compared with
+  // `valueEqual`, the `Fabric`-aware content equality, so a `FabricPrimitive`
+  // (state in private `#fields`, zero enumerable own-props) is compared by
+  // content rather than by its absent own-props. Note `valueEqual` throws on
+  // a function or a non-`Fabric` class instance rather than reporting it
+  // unequal; both operands here are `FabricValue`s by contract.
+  if (valueEqual(parent[path[path.length - 1]], value)) return false;
```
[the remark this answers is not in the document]
mathpirate: Deliberate and adjudicated: this throw-at-the-public-boundary semantic is disclosed in the PR body ("⚠️ Behavior notes"), was isolated as its own commit precisely so it could be dropped, and danfuzz approved landing it as-is. Tightening the exported signature from `any` to `FabricValue` is a follow-up-worthy typing change, not one to fold in post-approval.

packages/runner/src/storage/transaction/attestation.ts
```diff
@@ -147,15 +148,14 @@ export const claim = (
...
-  // Fast path: reference equality check avoids expensive comparison
-  // when the replica state hasn't changed since the original read
-  // TODO(danfuzz): This compares a stored document value (the read/attested
-  // value) with `deepEqual`, which mishandles `FabricValue`: two same-class
-  // `FabricPrimitive` values (state in private `#fields`, zero own-props)
-  // compare equal regardless of value, so a changed Fabric value can be
-  // mis-detected as unchanged. Use a Fabric-aware equality for stored-value
-  // comparison.
-  if (expected === actual || deepEqual(expected, actual)) {
+  // Compare the stored document value (the read/attested value) with
```
danfuzz: I'm meh on recapitulating what `valueEqual()` is all about in comments over every use site. Like, I know I mentioned `+0 === -0` in my comment, but there's _also_ `NaN === NaN`… and do you really want to expand the commentary to cover all such nuances? Likewise all the stuff about private `#fields`; those `TODO` comments were meant to capture information that was salient to the act of fixing the problem, not meant to be durable decoration over the function call.

This comment applies to the whole PR series and with `s/valueEqual()/FabricPrimitive/` etc. applies to some of the other non-TODO blocks.

I won't withhold approval if you want to ignore these suggestions, to be clear.

packages/runner/src/path-utils.ts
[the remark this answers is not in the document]
danfuzz: Agreed. i've certainly found that tightening `any` and `unknown` to `FabricValue` to be a fruitful endeavor (including uncovering bona fide bugs), but AFIAC there's never a mandate to DO ALL THE THINGS.

danfuzz (approved):

PR #4289, opened 2026-06-21
Mark recursive-traversal FabricValue mishandling sites — second pass (comment-only)

danfuzz:
## Summary

A **second-pass** recursive-traversal audit, generalizing the prior marker
sweep ([#4253](https://github.com/commontoolsinc/labs/pull/4253), 39
`TODO(danfuzz)` markers) to two orthogonal shapes that mishandle the special
Fabric classes. Adds **8 NEW `TODO(danfuzz)` markers** across 8 files — all
comment-only, no behavior change. This is **audit output** (a catalog of
fabric-unsafe sites), not a fix.

`FabricPrimitive`/`FabricInstance` keep their state in private `#fields`, so
they expose **zero enumerable own-properties**. Two distinct generic-traversal
idioms therefore silently mishandle them:

- **Own-props equality** (`deepEqual`): two distinct same-class
  `FabricPrimitive` values compare equal regardless of value.
- **Own-keys / structure walks** (`for...in`, `Object.entries`): a special
  object exposes no own-keys, so a walk finds nothing to descend into (a
  `FabricPrimitive` flattens to `{}`; a `FabricInstance` is walked by its
  internal slots instead of its codec contents).

The refined two-shape lens is what surfaced sites the first pass missed — the
same class of miss that [#4265](https://github.com/commontoolsinc/labs/pull/4265)
(CT-1770) found in `differential.ts`'s `collectChangedPaths` (an own-keys walk
that was **not** in #4253's file list).

## The 8 NEW marker sites

**Six reachable today:**

- `storage/transaction/attestation.ts` — stored-value attestation gate.
- `path-utils.ts` — no-op write gate.
- `scheduler/topology.ts` — `mapsEqual` over read-value maps.
- `runner.ts` — result-value gate.
- `builder/json-utils.ts` — `for...in` value walk.
- `builder/node-utils.ts` — CFC-label `Object.entries` descent.

**Two latent** — reachable once schemas admit `Fabric*` values (distinct
latent-phrasing on the marker so it's clearly not a live bug today):

- `schema.ts` — schema enum/const validation arm.
- `cfc/schema-sanitization.ts` — schema enum/const validation arm.

## Sequencing / relationship to prior work

This is a **NEW-only** delta. The 8 sites here are **disjoint** from:

- **#4253**'s 39 markers (the first marker pass).
- **#4265**'s 8 in-flight fixed src files (CT-1770, OPEN/unmerged) — none of
  those files are touched here.

[...]

PR #4530, opened 2026-07-04
refactor(runner): guard three value-rebuild walks against FabricPrimitive corruption

mathpirate:
Three builder/builtin walks rebuild records via generic enumerable-prop copies (`Object.entries`/`Object.fromEntries` or `for...in`). A `FabricPrimitive` (`FabricBytes`, `FabricEpochNsec`, …) keeps its state in private `#fields` and exposes zero enumerable own-props, so each walk silently decomposed such a value to `{}` — observably corrupting it. This applies Dan's #4238 guard shape (a `BaseFabricPrimitive` short-circuits to an atomic leaf before the generic record branch) to the three remaining walks that observably rebuild values.

### Markers narrowed

Each site carried a `TODO(danfuzz)` (from #4253 / #4289) naming both the `FabricPrimitive` *and* `FabricInstance` defects. This PR resolves the `FabricPrimitive` half only and rewrites each marker, narrowed to the remaining `FabricInstance` half (mirroring #4238's residual-TODO style):

- **`traverseValue` (traverse-utils.ts)** — before: *"…has no `FabricSpecialObject` guard before its `Object.entries`/`Array.map` descent, so it recurses into `FabricPrimitive` values … and walks `FabricInstance` values by their internal slots…"* → after: *"…now leaves `FabricPrimitive` values atomic …, but … `FabricInstance` (a container) is still walked by its internal slots instead of its codec contents. … This site will need attention once FabricInstances see real use."*
- **`toJSONWithLegacyAliases` (json-utils.ts)** — before: *"…has no `FabricSpecialObject` guard, so a `FabricPrimitive` … flattens to `{}` and a `FabricInstance` is walked by its internal slots…"* → after: *"A `FabricPrimitive` is now returned atomically above, but … `FabricInstance` (a container) still reaches this `for...in` copy and is walked by its internal slots… once FabricInstances see real use."*
- **`traverseAndCellify` (llm-dialog.ts)** — before: *"…has no `FabricSpecialObject` guard; a `FabricPrimitive` flattens to `{}` … and a `FabricInstance` is walked by internal slots."* → after: *"A `FabricPrimitive` is now returned atomically, but … `FabricInstance` (a container) still reaches the `Object.fromEntries(Object.entries(...))` walk … once FabricInstances see real use."*

(The sibling `serializeForLLMObservation` marker in llm-dialog.ts is deliberately untouched — it needs a rendering decision and is listed under "needs your design decision" in #4527.)

### Shape

- **`toJSONWithLegacyAliases` / `traverseAndCellify`**: early-return the primitive unchanged before the record-rebuild branch — atomic-leaf, exactly like #4238's `if (obj instanceof BaseFabricPrimitive) return obj`. Placed after the cell / alias / link / array handling that must still win.
- **`traverseValue` placement (deliberately different)**: this function applies `fn` to *every* value — leaves included — before descending, then rebuilds records. A `FabricPrimitive` should still be **seen by `fn`** like any string/number leaf; only the rebuild-descent must be skipped. So the guard is added to the **descent condition** (alongside the existing `isReactive`/`isCell`/`isCellResultForDereferencing` leaf-outs), *not* as a pre-`fn` early return. Two companion tests lock this in: `fn` still sees the primitive, and `fn` can still replace it — both pass **against the pre-fix code**, validating the placement. (The `instanceof` LHS is cast `as object` because the exclusion chain narrows `value` to a union TS rejects as a bare `instanceof` operand; harmless at runtime.)
- **Honesty:** the markers (Dan's own, from #4253/#4289) state the defect; the *fix-shape* here is **inferred from #4238**, not independently re-derived. The `BaseFabricPrimitive` chokepoint and atomic-leaf treatment are Dan's; applied mechanically to the three flagged walks.
- **Derived-byte churn:** `toJSONWithLegacyAliases` feeds builder serialization (its four call sites are in `builder/pattern.ts`, building `pattern.nodes`), which feeds derived, content-addressed artifacts. Fixing the flattening changes the serialized form of any pattern that embeds a `FabricPrimitive`, hence its derived ids — but, as in #4238, *"they are not widely used yet, so the storage-key churn is acceptable (fresh-db caveat)."* The other two sites (`traverseValue`, `traverseAndCellify`) are in-memory transforms with no persisted-id impact.

### Verification

- **Red-first, per site, against unmodified code:** each site's nested `FabricBytes` came out as `{}` (`-{} / +FabricBytes {}`, or an identity assertion for `traverseAndCellify`); failing output captured. For `traverseValue`, the two `fn`-behavior tests passed *before* the fix, validating the descent-condition placement.
- **Green:** all three site tests pass after the fix.
- **Gates:** `deno fmt`/`lint`/`check` clean on all touched files (and package-wide typecheck); **full `packages/runner` suite: 769 passed / 0 failed**.
- **Downstream independently verified** (adversarial review pass): the intact primitive reaching codec-aware consumers (value-hash dispatches `FabricBytes` via its native tag; cell writes normalize via `shallowFabricFromNativeValue`) is a strict improvement — pre-fix, `{}` collapsed all distinct primitives to one hash. The two `JSON.stringify` consumers downstream produce `{}` for a `FabricBytes` either way (no `toJSON`, zero enumerable props), identical to pre-fix — no regression.

### Needs Dan's judgment

1. **`traverseValue` `as object` cast** — cast the `instanceof` LHS because the leaf-out exclusion chain narrows `value` away from `any` (unlike #4238's `obj: any`). Acceptable, or would you prefer typing the residual differently / a small `isFabricPrimitive` predicate to match the sibling type-guards?
2. **`FabricInstance` residual** — left unhandled at all three sites per your #4238 precedent (needs correct *descent* by codec contents, not the generic walk). Confirm that's still the intended posture and the narrowed TODOs read the way you'd want.
3. **Derived-byte churn scope** — the fresh-db caveat is invoked only for `toJSONWithLegacyAliases` (the serialization path). Please confirm no currently-shipping pattern embeds a `FabricPrimitive` in a way that would make this churn more than theoretical.
4. **`traverseAndCellify` = identity contract** — asserted the *same instance* survives (its leaf case is `return value`). If you'd rather the contract guarantee only content-intactness (allowing a future codec round-trip), say so and the assertion relaxes.

---

---
## Summary by cubic
Guards three rebuild paths to keep `FabricPrimitive` values atomic instead of flattening to `{}`. Fixes silent corruption across traversal, JSON serialization, and LLM cellification; adds targeted tests. Part of #4527.

- **Bug Fixes**
  - `toJSONWithLegacyAliases`: early-return primitives before the `for...in` copy; residual TODO narrowed to `FabricInstance`.
  - `traverseValue`: guard the descent so `BaseFabricPrimitive` leaves are not rebuilt; `fn` still sees and can replace them; residual TODO narrowed to `FabricInstance`.
  - `traverseAndCellify`: early-return primitives before the `Object.fromEntries` rebuild; residual TODO narrowed to `FabricInstance`.
  - Added regression tests for all three sites (pass-through as same instance; `fn` visibility/replacement where applicable).

- **Migration**
  - Builder serialization now preserves embedded `FabricPrimitive` values, which can change derived, content-addressed IDs for affected patterns. Fresh DB recommended if this path is in use. Other sites are in-memory only.

<sup>Written for commit 41056802160a2a7022b8c2a9f4a3e52ad4e42bb6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4530?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/builder/json-utils.ts
+++ b/packages/runner/src/builder/json-utils.ts
@@ -1,4 +1,5 @@
 import { isRecord } from "@commonfabric/utils/types";
+import { BaseFabricPrimitive } from "@commonfabric/data-model/fabric-primitives";
 import {
   emptySchemaObject,
   schemaForValueType,
@@ -117,6 +118,14 @@ export function toJSONWithLegacyAliases(
     );
   }
 
+  // A `FabricPrimitive` is an atomic value whose state lives in private fields
+  // (zero enumerable own-props). The `for...in` copy branch below would flatten
+  // it to `{}`, so return it unchanged here — after the cell / alias / array
+  // handling above, which must still win for those forms.
+  if (value instanceof BaseFabricPrimitive) {
+    return value as unknown as JSONValue;
+  }
+
   // If this is an object or a pattern, process each key recursively.
   if (isRecord(value) || isPattern(value)) {
     // Guard against circular object references (e.g. schema objects with
@@ -139,11 +148,13 @@ export function toJSONWithLegacyAliases(
       : (value as Record<string, any>);
 
     const result: any = {};
-    // TODO(danfuzz): This `isRecord`-gated `for...in` walk has no
-    // `FabricSpecialObject` guard, so a `FabricPrimitive` (state in private
-    // fields, zero enumerable own-props) flattens to `{}` and a
-    // `FabricInstance` is walked by its internal slots instead of its codec
-    // contents.
+    // TODO(danfuzz): A `FabricPrimitive` is now returned atomically above, but
+    // the other special-object type, `FabricInstance` (a container), still
+    // reaches this `for...in` copy and is walked by its internal slots (zero
+    // enumerable own-props) instead of its codec contents. Unlike a primitive it
+    // *does* need descending into — but by its actual contents, which this walk
+    // won't do correctly. This site will need attention once FabricInstances see
+    // real use.
     for (const key in valueToProcess as any) {
       const jsonValue = toJSONWithLegacyAliases(
         valueToProcess[key],
--- a/packages/runner/src/builder/json-utils.ts
+++ b/packages/runner/src/builder/json-utils.ts
@@ -1,5 +1,5 @@
 import { isRecord } from "@commonfabric/utils/types";
-import { BaseFabricPrimitive } from "@commonfabric/data-model/fabric-primitives";
+import { FabricPrimitive } from "@commonfabric/data-model/fabric-value";
 import {
   emptySchemaObject,
   schemaForValueType,
@@ -122,7 +122,7 @@ export function toJSONWithLegacyAliases(
   // (zero enumerable own-props). The `for...in` copy branch below would flatten
   // it to `{}`, so return it unchanged here — after the cell / alias / array
   // handling above, which must still win for those forms.
-  if (value instanceof BaseFabricPrimitive) {
+  if (value instanceof FabricPrimitive) {
     return value as unknown as JSONValue;
   }
 
--- a/packages/runner/src/builder/traverse-utils.ts
+++ b/packages/runner/src/builder/traverse-utils.ts
@@ -1,5 +1,5 @@
 import { isRecord } from "@commonfabric/utils/types";
-import { BaseFabricPrimitive } from "@commonfabric/data-model/fabric-primitives";
+import { FabricPrimitive } from "@commonfabric/data-model/fabric-value";
 import { type FactoryInput, isPattern, isReactive } from "./types.ts";
 import { noteDerivedCopy } from "./pattern-metadata.ts";
 import { isCell } from "../cell.ts";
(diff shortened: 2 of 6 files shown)


packages/runner/src/builder/json-utils.ts
```diff
@@ -1,4 +1,5 @@
 import { isRecord } from "@commonfabric/utils/types";
+import { BaseFabricPrimitive } from "@commonfabric/data-model/fabric-primitives";
```
danfuzz: The correct class to use here is `FabricPrimitive` not `BaseFabricPrimitive`. The former is the "client use" class, while the latter is the class to `extend` when defining concrete classes.

Same comment applies throughout the PR.

danfuzz (dismissed):
Looks a-okay, modulo the suggested tweak.

danfuzz:
In re the "judgment" section:

* `traverseValue as object` cast

Seems fine as-is to me. I'm an un-fan of adding `isBlah(v)` when `v instanceof Blah` can be used (even if TS makes it more verbose than it really ought to be).

* `FabricInstance` residual

Also a-okay as-is.

* Derived-byte churn scope

I can't make the requested guarantee — I don't know what all the clients of `FabricPrimitive` do at this point. But I'm comfortable with making the PR and reacting if/as needed.

* `traverseAndCellify` = identity contract

That's a bit out of my territory. From my (lower layer) perspective if the existing unit tests are happy with it then I'm happy with it.

packages/runner/src/builder/json-utils.ts
mathpirate: thank you, incorporated!

danfuzz:
LGTM at this point. Thanks!

PR #4531, opened 2026-07-04
fix(runner): FabricPrimitive leafs, FabricInstance handled per-path in traverseDAG and schema defaults

mathpirate:
**Reworked 2026-07-20 per danfuzz's review** (the first draft guarded the wider `FabricSpecialObject`): only `FabricPrimitive` is fully opaque, so the guards narrow to it, and the other direct subclass, `FabricInstance` — which can have model-visible outgoing references — gets its own branch per path: failing loudly where nothing routes instances today, and deliberately backing off where live traffic exists. Also rebased onto current main.

A `FabricPrimitive` is `typeof "object"` with its state in private `#fields` and zero enumerable own-props. It now materializes as a fully-opaque leaf on the three sibling paths the value-type dispatch's arm doesn't cover — `traverseDAG`'s record rebuild, `processDefaultValue`'s object-type rebuild, and the default-annotation gate (previously a live crash: `TypeError: Cannot define property Symbol(toCell), object is not extensible` on a frozen primitive default).

Part of the fabric-safety series — tracking issue #4527. Companion PR #4845 applies the same subclass split to the value-type dispatch arm and the plain-schema fast path.

### Shape — per-path subclass split (per review)

| path | `FabricPrimitive` | `FabricInstance` |
| --- | --- | --- |
| `traverseDAG` record rebuild | opaque leaf | **leafs whole — deliberate back-off, see below** |
| `processDefaultValue` object rebuild | opaque leaf (after the asCell block, as before) | **throws** `Cannot yet handle … as a schema default.` |
| `annotateWithBackToCellSymbols` | pass-through, exactly like a `number`/`string` leaf (the review's litmus) | **throws** `Cannot yet handle … in back-to-cell annotation.` |

**The traverseDAG back-off** (the review's "only back away from that if that causes real trouble now"): with the throw in place, the full runner suite surfaced real instance traffic — the fetch builtins store a `FabricError` result value, and the tracked-query refresh (memory v2 `loadFactsForDoc`) reads it back through `traverseDAG`. The throw became an uncaught rejection in the refresh loop and broke fetch reactivity (three test files, plus the `#now` interval wish stalling to timeout). The instance therefore leafs whole on that path, pinned by a test that documents the back-off. The other two paths keep their throws — the full suite is green with them in place.

### Findings recorded during the rework

- **Interning decomposes instance-valued schema defaults** (upstream of this PR; recorded on #4527): a `FabricError` default never reaches `processDefaultValue` as an instance — `canonicalizeSchemaKeyOrder` (the intern-time key-order workaround already slated for removal by its own TODO) rebuilds it as a plain record (its five enumerable own-props, key-sorted), stripping the class. The instance-default tests therefore pin `processDefaultValue` directly. A `FabricBytes` default survives interning only because zero own-props makes that walk identity-return it.
- **`FabricError` carries enumerable own-props** (`type`, `name`, `message`, `stack`, `cause`) — the "zero enumerable own-props" story is primitive-specific. Instances corrupt through generic walks by losing their class and codec identity (plain-record rebuild), not by `{}`-decomposition.

### Verification

- **Red-first:** the primitive cases fail without the guards (`{}` decomposition / the annotate crash), captured verbatim; the two fail-loud cases fail against the guard-less code (verified by stashing `schema.ts`); the dispatch-path primitive cases pass on main and remain regression coverage for #4510's arm, including the pin for its deliberate `required` verdict change (CT-1836's mechanism).
- **The back-off is evidence-based:** the failing fetch/interval runs with the traverseDAG throw in place are what forced it; the pinning test documents the traffic (fetch's stored `FabricError` read back by the query path).
- **Gates:** fmt/lint/`deno check` clean; full `packages/runner` suite **970 passed / 0 failed** on the rebased tree.

### Needs Dan's judgment

1. The traverseDAG back-off — leafing whole vs some other interim posture for the live instance traffic (fetch's stored `FabricError`s).
2. Whether the `processDefaultValue`/annotate throws should land now (nothing in the unit suite routes instances there; the pattern-integration suites are the broader canary — if they surface traffic, the same back-off applies).
3. Disposition of the interning finding (`canonicalizeSchemaKeyOrder` rebuilding instance defaults) — your TODO already slates the function for removal.

---

### DIFF
--- a/packages/runner/src/schema.ts
+++ b/packages/runner/src/schema.ts
@@ -12,6 +12,10 @@ import {
   shallowMutableClone,
 } from "@commonfabric/data-model/fabric-value";
 import { isDeepFrozen } from "@commonfabric/data-model/deep-freeze";
+import {
+  FabricInstance,
+  FabricPrimitive,
+} from "@commonfabric/data-model/fabric-value";
 import { internSchema } from "@commonfabric/data-model/schema-hash";
 import {
   isNontrivialSchema,
@@ -675,6 +679,24 @@ export function processDefaultValue(
     }
   }
 
+  // A `FabricPrimitive` default is a fully-opaque leaf (state in private
+  // `#fields`, zero enumerable own-props): the object-type rebuild below
+  // would decompose it to `{}`. Return it as-is; it takes no back-to-cell
+  // annotation (see `annotateWithBackToCellSymbols`).
+  if (defaultValue instanceof FabricPrimitive) {
+    return defaultValue;
+  }
+  // A `FabricInstance` default is not a fully-opaque leaf: it can have
+  // model-visible outgoing references, so neither the object-type rebuild
+  // below nor returning it whole handles it correctly. Fail loudly until
+  // instances can be processed by their codec contents.
+  if (defaultValue instanceof FabricInstance) {
+    throw new Error(
+      `Cannot yet handle \`${defaultValue.constructor.name}\` (a ` +
+        "`FabricInstance`) as a schema default.",
+    );
+  }
+
   // Handle object type defaults
   if (
     resolvedSchema?.type === "object" && isRecord(defaultValue) &&
@@ -876,9 +898,27 @@ function annotateWithBackToCellSymbols(
   synced = false,
   cfcLabelView?: CfcLabelView,
 ) {
-  if (!isRecord(value) || isCell(value)) {
-    // We only possibly annotate objects or arrays that _aren't_ cells.
+  if (
+    !isRecord(value) || isCell(value) ||
+    value instanceof FabricPrimitive
+  ) {
+    // We only possibly annotate plain objects or arrays that _aren't_ cells.
+    // A `FabricPrimitive` is a fully-opaque leaf and takes no back-to-cell
+    // annotation, exactly like a plain `number` or `string` leaf: it is
+    // always frozen (`defineProperty` would throw "object is not
+    // extensible"), and cloning-or-freezing-to-annotate would misrepresent
+    // it. It passes through untouched.
     return value;
+  }
+  if (value instanceof FabricInstance) {
+    // Unlike a primitive, a `FabricInstance` can have model-visible outgoing
+    // references, so skipping annotation is not obviously right for it — and
+    // the generic record annotation below would be wrong. Fail loudly until
+    // the back-to-cell story for instances exists.
+    throw new Error(
+      `Cannot yet handle \`${value.constructor.name}\` (a ` +
+        "`FabricInstance`) in back-to-cell annotation.",
+    );
   }
 
   const extensible = Object.isExtensible(value);
--- a/packages/runner/src/schema.ts
+++ b/packages/runner/src/schema.ts
@@ -679,17 +679,14 @@ export function processDefaultValue(
     }
   }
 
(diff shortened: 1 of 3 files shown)


packages/runner/src/schema.ts
```diff
@@ -665,6 +666,15 @@ export function processDefaultValue(
     }
   }
 
+  // A `FabricSpecialObject` default is an opaque leaf (state in private
+  // `#fields`, zero enumerable own-props): the object-type rebuild below
+  // would decompose it to `{}`. Return it as-is; it takes no back-to-cell
+  // annotation (see `annotateWithBackToCellSymbols`). Matches the traversal's
+  // `FabricSpecialObject` leaf arms (CT-1836 / #4510).
+  if (defaultValue instanceof FabricSpecialObject) {
```
danfuzz: Of the two direct subclasses of `FabricSpecialObject`, it's only `FabricPrimitive` which is fully opaque. `FabricInstance` _can_ have model-visible outgoing references, and this fact will ultimately have to be accounted for in schema definitions. For example, we'll want to be able to model (and validate) something like `FabricSet<string>`.

As such, the right "treat it as a simple opaque value" test is `instanceof FabricPrimitive`. Same comment applies below (this and other files).

danfuzz:
In re the "judgment" section:

**Guard breadth**

Answered in inline comment. TLDR: It's actually not appropriate to treat `FabricSpecialObject` as an opaque leaf. The subclass `FabricPrimitive` _can_ be treated as an opaque leaf, though.

**Instance leafing is opacity, not descent**

That's fair, but I would still structure the code so that it has separate tests for `FabricPrimitive` and `FabricInstance`, filled in in an as-appropriate way as possible. And I'd also try `throw`ing (`...not yet handled` or similar) in the `FabricInstance` branches of the code rather than silently doing the wrong thing, and only back away from that if that causes real trouble now. That is, I'd prefer seeing errors thrown when we ultimately try to use `FabricInstance` rather than have to track down silent bad behavior.

**Array-element residual**

This seems acceptable to me at this point.

**The annotate skip**

This does seem correct to me for `FabricPrimitive` — litmus test: is the same true of a regular primitive like `number` or `string`? — but _incorrect_ for `FabricInstance`.

danfuzz (changes requested):
Per comments, I think there's a bit more work needed on this one. Thanks!

packages/runner/src/schema.ts
mathpirate: thank you, incorporated!

packages/runner/src/schema.ts
```diff
@@ -704,6 +708,24 @@ export function processDefaultValue(
...
   }
 
+  // A `FabricPrimitive` default is a fully-opaque leaf (state in private
+  // `#fields`, zero enumerable own-props): the object-type rebuild below
+  // would decompose it to `{}`. Return it as-is; it takes no back-to-cell
+  // annotation (see `annotateWithBackToCellSymbols`).
+  if (defaultValue instanceof FabricPrimitive) {
+    return defaultValue;
+  }
+  // A `FabricInstance` default is not a fully-opaque leaf: it can have
```
danfuzz: Please mark this section (and similar other ones) with `TODO(danfuzz)` so I can find it easily when I'm in the mode of fixing all the `FabricInstance` traversals.

danfuzz (approved):

PR #4533, opened 2026-07-05
test(runner): pin B3 — writer never integrates a peer's concurrent append (decision needed)

bfollington:
**Draft, stacked on the B2 PR — this is the review-required half of #4457.** It contains only **deliberately-RED tests** pinning B3, the writer-side integration gap. The tests assert the desired end-state and stay red until a fix direction is chosen; they are the red-to-green target for whichever option lands. All three candidate fixes touch the core sync path, so this needs @robin / @gideon eyes rather than a self-merge.

## The defect (B3)

**Symptom** (`packages/runner/test/array-append-client-convergence.test.ts`, 60ms, deterministic): seed `[seed]`; session 1 appends `A`; session 2 appends `B` without observing `A`. The server merges all three, a fresh reader sees `[seed, A, B]`, but **session 2's own replica** — after explicit `sync + pull` — reads `[seed, B]`. `A` is gone from the interleaving writer's view, permanently.

**Mechanism** (instrumented `applySessionSync` + `confirmPending`):
- `confirmPending` promotes session 2's `confirmed` to its own commit's (high) `applied.seq` by replaying its append onto its **own stale base** = own appends only.
- The `AppliedRevision` for a `patch` op (append) carries only `patches`, **not** the merged document (only `set` ops carry `document`), so the merge never returns through the commit response.
- Session 2's own commit advanced its server-side session cursor to head, so delta-sync re-delivers nothing on `pull()`. Confirmed **value** and server **value** now diverge **at the same seq** — a seq-based delta protocol can't detect or repair that.

Storm-scale variant (in `convergence-storm.test.ts`, `convergence-chat-plain` fixture — no links, so no blackout): all sends land durably and a non-writing observer converges, but the racing writer keeps a stale own-appends-only array through 20 settle rounds and pulling reads (5/5 reproducible via `storm-driver.ts 20 2 pipeline 0 convergence-chat-plain`).

Note this is **independent of** #4465 (stale-basis retry on mergeable ops, merged): that fixed a writer's *own* append being dropped under a rehydration storm; B3 is a writer's replica never integrating a *peer's* append. The B3 tests were re-verified red on top of current main.

## Three viable fixes — the decision

1. **Client fresh-watch resync** of the raced doc after a mergeable-append commit. Pure client; works because the server snapshot is keyed per-watch-id, not per-session-cursor. Cost: +1 round-trip per append unless gated on detected concurrency.
2. **Server merge-signal** on the revision (`merged` / `rebasedAgainstSeq`) → client resyncs only flagged docs. Targeted.
3. **Merged `document` in the applied revision** for merged commits → client adopts it directly. Cleanest semantically; costs materialize-per-commit + wire size.

Cross-ref: #4405 (lunch-poll vote swatches) is in the same `v2.ts` pending-replay code and the same symptom family — "durable has all, an interleaving writer's replica doesn't."

## How to run

```
deno test -A packages/runner/test/array-append-client-convergence.test.ts   # RED, 60ms, deterministic
deno test -A packages/patterns/integration/convergence-storm.test.ts        # "writer integration gap" case RED; rest green
```

Full context: #4457 (original investigation package), `docs/plans/2026-07-02-convergence-evidence-appendix.md`, `docs/plans/2026-06-30-profile-loading-investigation.md`.

Also still open from the investigation: **B1** — a live `PerUser` cell pushed into shared state stores a scope-generic `"user"` link (owner DID lost at link-mint), so readers resolve it into their own partition. Fix direction: capture `user:<did>` at link-mint or resolve scope at write time. Same class as the known profile owner-principal issue.

---
## Summary by cubic
Adds deliberately failing tests to pin B3: when two writers append concurrently, the interleaving writer never integrates the peer’s append and its replica stays stale after sync/pull. These tests define the red-to-green target while we choose a fix.

- Adds a deterministic unit test at `packages/runner/test/array-append-client-convergence.test.ts`.
- Adds a storm-scale “writer integration gap (plain elements)” case in `packages/patterns/integration/convergence-storm.test.ts` (kept red by design).

<sup>Written for commit e664378c411e6f5edd22468b0e6dc02bad4a5613. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4533?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bfollington:
I never managed to prove that this was a worthwhile uh PR to actually merge and that it didn't just introduce additional problems for any that it fixed. So it's never made it made a difference. We didn't want to m merge it in the last seven weeks. Let's close it.

PR #4534, opened 2026-07-06
fix(runner): resolve #profile MRU/default ordering by link identity, not Cell.equals (CT-1842)

bfollington:
## Why

In real (cross-space) deployments the `#profile` MRU switcher was a no-op: picking a profile via the picker's **Use** button stamped MRU but `#profile` / `#profileName` kept resolving to the *first-created* profile instead of the MRU head. Verified live against the switchtest identity (profiles Grace + Alan, `mru = [Alan, Grace]`): `#profileName` returned "Grace" persistently; after this fix it returns "Alan".

Pre-existing since #3830 (2026-06-09); CT-1829 (#4512) relies on this ordering for the picker-as-switcher and exposed it. This unblocks CT-1829's switcher.

Fixes CT-1842.

## Root cause

`getProfileCandidateCells` in `packages/runner/src/builtins/wish.ts` orders profile candidates default → MRU → first, matching candidates (from the home `profiles` list) against the `defaultProfile` link and the `mru` list with `Cell.equals`.

Those lists name the same profiles but reach them through **different links**. On live data the `mru`/`defaultProfile` link and the `profiles` link for the SAME profile point at **different cells within that profile's own space — same space, DIFFERENT entity id** (the picker stores one cell, e.g. the profile pattern's result cell; the list stores another). `Cell.equals` (`areNormalizedLinksSame`) compares `id` (and `scope`), so it returns false for every candidate. Every candidate then ranks `MAX_SAFE_INTEGER`, the MRU sort is a no-op, and `#profile` falls back to `profiles`-list (creation) order.

Hard evidence (switchtest identity):

```
mru[0]  space …Ubj5gsae (Alan)    profiles[1] space …Ubj5gsae (Alan)   — same space, different id
mru[1]  space …WwikkTRr (Grace)   profiles[0] space …WwikkTRr (Grace)  — same space, different id
```

The default-match (`defaultCell.equals(a)`) shared the exact same bug and is fixed too.

## Fix

Match by the profile's own **space DID**, not `Cell.equals` and not entity id. Each profile is a distinct anonymous `ProfileHome.inSpace()` (see `submitProfileCreation`), whose DID is unique per user AND per creation event; `profileCellIsValid` guarantees every valid candidate lives in its own non-home space; no two distinct valid profiles ever share a space. So equal space ⇒ same profile, and the different-id / scope differences no longer defeat the match. A `homeSpace` guard rejects any entry that still resolves into the home space (an unmaterialized/invalid link).

New `sameProfileCell(a, b, homeSpace)` helper compares normalized-link `space`; applied to both the MRU rank and the default match. Ordering contract (default → MRU → first) unchanged; only the matching mechanism changed. Reading each cell's normalized link keeps the ordering reactive to `mru`/`defaultProfile` changes.

## Test (fails without the fix)

- **Live acceptance:** ran this branch's CLI against the local dev server with the switchtest identity — `#profileName` and `#profile` both resolve to **"Alan"** (MRU head), not "Grace".
- **New unit tests** — `cross-space id skew (CT-1842)` in `wish.test.ts`: profiles each in their own space; `mru`/`defaultProfile` links point at a **different cell in the same profile space** (different entity id) — the real production shape.
  - `MRU head resolves cross-space despite different-id mru links` — **red on unfixed code** (resolves first-created).
  - `default resolves cross-space despite different-id default link` — **red on unfixed code** (default ignored).
  - `default outranks MRU even under id skew`.
  Confirmed red-without-fix by temporarily reverting to `Cell.equals`.
- **Migrated existing ordering tests** (CT-1829 `setupProfiles` and the standalone default/MRU tests) from same-space-multiple-profiles to **one-space-per-profile** — the real invariant. All CT-1829 single-result contract tests stay green.

Verification: `wish.test.ts`, `profile-create-real-card-add.test.ts`, `wish-scope.test.ts`, `wish-shared-hashtag.test.ts`, `cross-space-value-read.test.ts` all pass (11 tests / 115 steps); `deno fmt` / `deno lint` / `deno task check` clean.

## Note on same-space matching

Space-based identity is only wrong if two distinct valid profiles ever share a space — which `submitProfileCreation` (anonymous per-event `inSpace()`) prevents. The previously same-space test setups were a test fiction, now corrected. A single profile resolves via `ordered[0]` regardless of matching, so single-profile identities are unaffected.

### DIFF
--- a/packages/runner/src/builtins/wish.ts
+++ b/packages/runner/src/builtins/wish.ts
@@ -382,8 +382,11 @@ function subscribeProfileName(cell: Cell<unknown>): void {
 /**
  * Enumerate the user's profile candidate cells from the home `profiles` list,
  * ordered: default first, then by most-recently-used (MRU), then remaining list
- * order. Identity is by link equality (`Cell.equals`); there is no synthetic
- * key. Returns [] when no valid profile exists yet.
+ * order. Identity is by the profile's own SPACE — each profile is a distinct
+ * `ProfileHome.inSpace()` space, so the `defaultProfile` / `mru` links are
+ * matched to candidates by space, not `Cell.equals` (see `sameProfileCell`;
+ * CT-1842). There is no synthetic key. Returns [] when no valid profile exists
+ * yet.
  */
 function getProfileCandidateCells(
   ctx: WishContext,
--- a/packages/runner/src/builtins/wish.ts
+++ b/packages/runner/src/builtins/wish.ts
@@ -320,6 +320,33 @@ function profileCellIsValid(
 }
 
 /**
+ * Whether two profile cells name the SAME profile, compared by durable link
+ * identity (entity id + space + path) — NOT `Cell.equals`.
+ *
+ * CT-1842: the `#profile` ordering matches candidates (from the home `profiles`
+ * list) against the `defaultProfile` link and the `mru` list. Those name the
+ * same profiles but reach them through different links — a candidate is read
+ * straight from `profiles`, while an `mru`/`defaultProfile` entry is a link
+ * written by the picker surface (`setMruProfile` / `setDefaultProfile`), which
+ * can carry a write-redirect overwrite flag and its own schema. `Cell.equals`
+ * (`areNormalizedLinksSame`) also compares `scope` and does not collapse those
+ * representation differences, so it returns false for the same profile across
+ * spaces and the MRU/default ordering silently degrades to `profiles`-list
+ * (creation) order. Normalizing both sides with `getAsNormalizedFullLink()`
+ * resolves each link the same way and compares the profile's durable identity
+ * (its own-space entity id), which is stable regardless of how the link was
+ * stored. Reading each cell's normalized link also keeps the ordering reactive
+ * to `mru`/`defaultProfile` changes.
+ */
+function sameProfileCell(a: Cell<unknown>, b: Cell<unknown>): boolean {
+  const la = a.getAsNormalizedFullLink();
+  const lb = b.getAsNormalizedFullLink();
+  return la.id === lb.id && la.space === lb.space &&
+    la.path.length === lb.path.length &&
+    la.path.every((p, i) => p === lb.path[i]);
+}
+
+/**
  * Subscribe to a profile cell's live name so the wish re-runs once a
  * freshly-created profile's name materializes across the space boundary.
  */
@@ -391,16 +418,17 @@ function getProfileCandidateCells(
   for (let j = 0; j < mruLength; j++) {
     mruCells.push(mruCell.key(j).resolveAsCell());
   }
+  // Match by durable link identity, not `Cell.equals` — see sameProfileCell.
   const mruRank = (cell: Cell<unknown>): number => {
-    const idx = mruCells.findIndex((m) => m.equals(cell));
+    const idx = mruCells.findIndex((m) => sameProfileCell(m, cell));
     return idx === -1 ? Number.MAX_SAFE_INTEGER : idx;
   };
 
   const ordered = [...candidates];
   ordered.sort((a, b) => {
     if (defaultValid) {
-      const aDef = defaultCell.equals(a);
-      const bDef = defaultCell.equals(b);
+      const aDef = sameProfileCell(defaultCell, a);
+      const bDef = sameProfileCell(defaultCell, b);
       if (aDef && !bDef) return -1;
       if (bDef && !aDef) return 1;
     }
--- a/packages/runner/src/builtins/wish.ts
+++ b/packages/runner/src/builtins/wish.ts
@@ -320,30 +320,43 @@ function profileCellIsValid(
 }
 
 /**
- * Whether two profile cells name the SAME profile, compared by durable link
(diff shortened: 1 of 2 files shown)


bfollington (commented):
_cf-review via Claude (claude-opus-4-8), on behalf of @bfollington — automated review, posted by the agent._

**Scope** · `wish.ts` #profile MRU/default matcher (`Cell.equals` → match-by-profile-space) + `wish.test.ts` (new `cross-space id skew` describe + CT-1829 helper migrated to one-space-per-profile). 2 files, +/- small. · **Coverage:** deep-read both changed files, `profile-create.tsx`/`profile-picker.tsx` call-sites, `link-types.ts`/`scope.ts`. · **Checks run:** `deno test packages/runner/test/wish.test.ts` (green with fix; **red without** it — verified myself), `deno lint` on both files (clean).

## Verdict: non-blocking — the fix is correct and well-tested. One stale doc comment to fix in passing.

The rework from the earlier `id+space+path` matcher to **match-by-profile-space** is the right call and the load-bearing assumption holds. Details below.

---

### ✅ Correctness of space-matching (the load-bearing assumption) — verified

`sameProfileCell` now compares only `getAsNormalizedFullLink().space`, with a `homeSpace` guard. This is sound because **no two distinct valid profiles ever share a space**:

- `submitProfileCreation` (`packages/patterns/system/profile-create.tsx:71-97`) creates each profile via anonymous `ProfileHome.inSpace()`. The CT-1650 comment (lines 60-70) documents that the anonymous DID is derived from the handler's frame cause — **unique per user AND per creation event** — precisely to avoid the name-derived collision. So each creation event → a distinct space.
- `profileCellIsValid` (`wish.ts:312-320`) rejects any candidate whose link resolves into the home space or has a non-empty path, so every candidate is a distinct non-home space.

Equal non-home space ⇒ same profile. It is neither too loose (distinct profiles can't collide — they never share a space) nor too strict (the id/scope skew that defeated `Cell.equals` and even `id+space+path` no longer matters, because id is exactly what differs between the picker's link target and the `profiles` list cell within one profile space). The **home-space guard** (`spaceA === homeSpace || spaceB === homeSpace → false`) correctly rejects an unmaterialized/invalid mru/default entry that still resolves into home space; candidates are already non-home, so in practice it guards the mru/default side, as the docstring states.

### ✅ Red-without-fix — verified myself, non-vacuous

Reverted the matcher back to `Cell.equals` and ran the suite: both `MRU head resolves cross-space…` and `default resolves cross-space…` go **RED** (resolve `Ada`, the first-created, instead of the MRU/default head `Grace`) — the exact production symptom. Restored; green again. The tests genuinely fail without the fix. (`default outranks MRU even under id skew` stays green both ways — it isn't a skew-discriminating test, it guards default-over-MRU precedence, which is fine.)

### ✅ CT-1829 helper migration didn't weaken coverage

`setupProfiles` and the two ordering tests moved from "all profiles in one space" to one-space-per-profile — this **corrects a test fiction** (production is always one-space-per-profile) rather than dropping coverage. Those tests still meaningfully assert ordering and stay green even on unfixed code, because they write mru/default with matching ids (no skew) — the skew coverage is the new describe's job. Good separation.

### ✅ Reactivity preserved

`getAsNormalizedFullLink()` returns the pre-resolved link (no new value subscription), same as `Cell.equals` did. Reactivity to `mru`/`defaultProfile` changes comes from the unchanged upstream reads — `mruCell.asSchema(...).get()`, `defaultEntry.getRaw()`, `resolveAsCell()` per entry — all still in the reactive path. The wish re-resolves when mru/default change, same guarantee as before.

### ✅ Anti-duplication — clean now

The earlier `id+space+path` version duplicated `areNormalizedLinksSameIgnoringScope` (`link-types.ts:232`); the space-only matcher does not — no canonical helper does "compare by space with a home guard", so a local helper is justified.

---

## 🟡 Improvement (non-blocking)

**Stale docstring on the changed function.** `wish.ts:382-387`, the JSDoc for `getProfileCandidateCells`, still reads: _"Identity is by link equality (`Cell.equals`); there is no synthetic key."_ That is now **actively wrong** — this PR replaced `Cell.equals` with space-based matching, and this comment sits directly on the function whose behavior changed. Someone reading it to understand profile identity would be misled (the north-star coherence test). One-line fix, e.g. _"Identity is by the profile's own space (see `sameProfileCell`); there is no synthetic key."_

## Questions / follow-ups for a human to file

- **profile-picker.tsx shares the latent bug — correctly out of scope here.** `profile-picker.tsx` uses `equals(entry, target)` in two places. The `resultIndex`/`result` computeds (lines 54-79) are already flagged **vestigial, slated for deletion** (CT-1829 follow-up) — no action. But the **live `[UI]` "default" badge** match (`equals(def, p)`, line 114) is *not* vestigial: under the same cross-space id skew it could mis-mark which row is the default (cosmetic — show "Set default" on the already-default row, or fail to label it). That's a real latent bug in a *separate pattern/package*; fixing it here would leak scope. Worth a separate follow-up; not a blocker for this runtime fix.

**Verdict:** non-blocking — ship it. The runtime #profile resolution the embedder (Loom) binds to is correctly fixed and the tests are red-without-fix. Fold the one-line docstring fix in if convenient.

PR #4536, opened 2026-07-06
fix(patterns): match profile identity by space in picker default badge; drop vestigial resultIndex (CT-1843)

bfollington:
## Why

In real (cross-space) deployments the profile picker mislabels/omits the **default** badge: the interactive `[UI]` used `equals(def, p)` to decide which profile row shows "default". The `defaultProfile` link and the `profiles`-list entry for the SAME profile reach it through **different links** — different entity `id` (and scope) within that profile's own space — so `Cell.equals` (which compares id + scope) returns false, and the actual default never gets the badge.

This is the SAME class of bug as CT-1842 (#4534), which fixed the runner-side `#profile` MRU/default ordering by matching on the profile's own **space** instead of `Cell.equals`. This PR applies that idea in the picker **pattern**.

It also removes the vestigial `resultIndex` / `result` computeds that CT-1829 (#4512) left dead — the wish `.result` no longer flows through the picker (the wish builtin resolves `.result` itself), and nothing reads the picker's `result` / `candidates` anymore. Removing them also removes a second copy of this same latent `equals` bug.

Links: CT-1843, CT-1842 (#4534), CT-1829 (#4512).

## What

- Add module-scope helpers `linkSpace(cell)` (a cell's own space DID via `getAsNormalizedFullLink().space`, defensive) and `sameProfileCell(a, b, homeSpace)` — mirroring the runner's `sameProfileCell` (wish.ts, CT-1842): match two profile cells by their **space** (each profile is a distinct anonymous `ProfileHome.inSpace()`, so equal space ⇒ same profile), with a `homeSpace` guard rejecting entries that resolve into the home space.
- The pattern-facing API has **no typed space accessor**; the picker already uses the `(cell as any)` escape hatch for `.asSchema(...)`, and `home-ben.tsx` already calls `getAsNormalizedFullLink()` on a cell in pattern code — same precedent used here. Derive `homeSpace` from the `profiles` container link (which lives in the home space, matching how wish.ts sources its guard).
- Replace the `[UI]` default-badge `equals(def, p)` with `sameProfileCell(def, p, homeSpace)` (still reading `def` as a cell REF via `profileLinkSchema()` so a not-yet-loaded cross-space profile doesn't collapse to `undefined`).
- Delete the vestigial `resultIndex` / `result` computeds and drop `result` / `candidates` from the picker output; the picker output is now just `{ [UI]: VNode }`. Verified `home.tsx` (the only consumer) and `wish.ts` render only the picker's `[UI]`, never its `result` / `candidates`.

## Test / verification limits

- `deno fmt`, `deno lint`, and `cf check` are clean on `profile-picker.tsx` **and** on `home.tsx` (the consumer, whose picker output type changed). The transformer lifts `homeSpace = linkSpace(profiles)` and the badge condition into reactive computeds as expected.
- Runner `wish.test.ts` passes (unchanged; picker output-shape change is compatible — those tests read the picker's `[UI]` and the main wish `.result`, not the picker's own `result`/`candidates`).
- **No new pattern-unit test**: exercising the cross-space default badge requires two profiles in distinct `inSpace` spaces plus a `defaultProfile` link — a cross-space create the pattern-unit lane forbids (it trips the console-error gate via closure-replication-failed). The space-matching correctness therefore needs **browser verification**: with 2+ profiles and one set as default, confirm the default profile's row shows the "default" badge cross-space. The runner-side twin (`sameProfileCell` in wish.ts) is already covered by CT-1842's cross-space `wish.test.ts` cases.

---
## Summary by cubic
Fixes the profile picker so the “default” badge shows on the correct profile across spaces, and simplifies the picker API by removing unused `resultIndex`/`result`. Addresses CT-1843 by aligning the picker with the space-based matching used in CT-1842.

- **Bug Fixes**
  - Match profile identity by the profile’s own space DID instead of `equals`, via `linkSpace` and `sameProfileCell`.
  - Derive `homeSpace` from `profiles` and reject any entries that resolve into the home space.
  - Replace `equals(def, p)` with `sameProfileCell(def, p, homeSpace)` to render the default badge correctly across spaces.

- **Migration**
  - Picker output is now only `{ [UI]: VNode }`; `result` and `candidates` were removed. Update any consumer still reading them.

<sup>Written for commit 090347ac70823a3ef27ce6ce69cf72235b78ae1e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4536?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/patterns/system/profile-picker.tsx
+++ b/packages/patterns/system/profile-picker.tsx
@@ -1,16 +1,13 @@
 import {
   computed,
-  equals,
   ifElse,
   NAME,
   pattern,
   UI,
   type VNode,
-  type WishState,
   Writable,
 } from "commonfabric";
 import ProfileCreate, {
-  profileLinkListSchema,
   profileLinkSchema,
   setDefaultProfile,
   setMruProfile,
@@ -33,9 +30,9 @@ import type { ProfileHomeOutput } from "./profile-home.tsx";
 // trusted picker-surface actions (owner-protected; see profile-create); those
 // writes reorder the builtin's candidates, which is what flips `.result`.
 //
-// The `resultIndex`/`result` computeds below duplicate the builtin's ordering
-// and are VESTIGIAL — nothing reads them anymore; kept one release for
-// compatibility, then to be deleted (CT-1829 follow-up).
+// The picker output is only `[UI]` — it does NOT surface a `result` /
+// `candidates` (the wish builtin owns those); its former `resultIndex` / `result`
+// computeds were vestigial after CT-1829 (#4512) and are removed (CT-1843).
 
 type ProfilePickerInput = {
   profiles: Writable<ProfileHomeOutput[]>;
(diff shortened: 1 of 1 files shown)


bfollington (commented):
_cf-review via Claude `claude-opus-4-8`, on behalf of @bfollington. Automated review; a human owns the merge._

**Scope** · picker pattern only (`packages/patterns/system/profile-picker.tsx`, +58/−40): swap the default-badge match from `Cell.equals` to a space-based `sameProfileCell`, and drop the vestigial `resultIndex`/`result` computeds (narrowing the picker output from `WishState<ProfileHomeOutput> & {[UI]}` to `{[UI]: VNode}`).
**Coverage** · deep-read the diff + both consumers (`wish.ts`, `home.tsx`) + tests/CLI; verified against the CT-1842 runner twin; ran `cf check` (picker + home.tsx, both exit 0), `deno lint`, `deno fmt --check`, and `--show-transformed` on the actual PR file.

## Verdict: clean — safe to merge. No blocking findings.

The two highest-risk claims both check out independently.

### ✅ The output-type narrowing is safe — nothing reads the picker's `result`/`candidates` (verified)

The picker sidecar cell is consumed in exactly two places, and neither reads `.result` or `.candidates`:

- **`packages/runner/src/builtins/wish.ts`** — `launchProfilePickerPattern` (~L1759) returns `profilePickerPatternResultCell`, used only for run/stop lifecycle and as the `$cell` of `<cf-render>` in `profilePickerUI` (L1747-1752). In the `#profile` branch (L2044-2068) the wish's own `{ result, candidates, [UI] }` is assembled from the runner's `uniqueResultCells[0]` / `candidatesCell` — the **wish's** result/candidates, computed in the builtin, never read off the picker sidecar. This is the CT-1829 design (`.result` rides the main wish state, not the picker).
- **`packages/patterns/system/home.tsx`** (L227) renders `{profilePicker}` as a VNode child → `[UI]` only.
- Tests (`wish.test.ts`, `home-profile.test.ts`, `home-rehydration-churn.test.ts`) exercise the picker via its `[UI]`/DOM and read `.result` off the **main wish state** (`result.key("profile").get()?.result`), not the picker's own output. The CLI `wish` command runs headless — the picker never spins up.

`--show-transformed` confirms the emitted picker return is just `[NAME]` + `[UI]`; no `result`/`candidates` survive. CI is green including runner + pattern integration lanes. The narrowing is a genuine dead-code removal, and desirable — it deletes a second latent copy of the same `equals` bug and stops advertising an output nothing consumes.

### ✅ `getAsNormalizedFullLink` is legitimately pattern-accessible (verified)

Not reaching into runner internals: `getAsNormalizedFullLink` is a first-class cell method in the `cellMethods`/allowlist set (`packages/runner/src/cell.ts:469`), and the runner calls the very same method on the very same `homeDefaultPattern.key("profiles")` link (`wish.ts:1767`). The `home-ben.tsx:73` precedent is real — it already calls `getAsNormalizedFullLink()` on a cell in pattern code via the `as any` hatch, defensively. `linkSpace` mirrors that idiom (try/catch + optional chaining; module scope, so no reactive-context concern). The `as any` is only about the *TS surface* lacking a typed space accessor; the runtime method exists in the sandboxed pattern.

### ✅ Space-matching is sound and reactively lifted (verified)

- **Fidelity to CT-1842.** The pattern-side `sameProfileCell` is a faithful mirror of the runner-side helper (commit `bfd0c3489`), plus a correct extra `if (!spaceA || !spaceB) return false` guard (needed because `linkSpace` can return `undefined` from the defensive hatch, whereas the runner reads `.space` off a real `Cell`).
- **One-space-per-profile invariant.** Each profile is a distinct anonymous `ProfileHome.inSpace()` (unique DID per user + per creation event), so equal space ⇒ same profile — the same invariant CT-1842 relies on and its cross-space tests cover.
- **`homeSpace` guard.** Derived as `linkSpace(profiles)`; the `profiles` container link lives in the home space, so this equals the home DID — exactly matching how the runner sources its guard (`homeSpaceCell.space`). Rejects unmaterialized entries that resolve to home space.
- **Reactivity.** `--show-transformed` shows the transformer lifts `homeSpace = linkSpace(profiles)` into `__cfLift_1(...).for("homeSpace", true)` (reactive, keyed on `profiles`) and lifts the badge condition into a computed over `{ defaultProfile, p, homeSpace }` — so the badge updates when the default changes. `def` reads through `profileLinkSchema` (`{ type:"unknown", asCell:["cell"] }`) so it's a real cell ref, and `p` is the map-element cell already bound to `cf-cell-link` / `equals` in the prior code — both respond to `getAsNormalizedFullLink()`.

1. **The referenced runner twin (#4534 / CT-1842) is still OPEN, not merged.** On `origin/main`, `getProfileCandidateCells` still orders `#profile` by `Cell.equals` (wish.ts:395/402-403); the space-based `sameProfileCell` lives only on the #4534 branch. Two consequences:
   - The PR body/comment call it "landed by CT-1842 #4534" and say the twin is "already covered by CT-1842's cross-space `wish.test.ts` cases" — accurate about the *branch*, but reads as if it's on main. Consider softening to "mirrors the pending CT-1842 (#4534)" until #4534 merges, so a future reader searching the wording isn't misled.
   - Behaviorally, until #4534 merges the picker **badge** will be space-matched (correct) while the wish `.result` ordering the badge sits beside is still `equals`-matched (buggy cross-space) — a transient split-fix inconsistency, not a regression this PR introduces. Landing the two close together avoids a window where the badge and the resolved profile disagree.

2. **No pattern-unit test for the cross-space badge.** The PR explains why (the pattern-unit lane forbids the cross-space create it needs) and defers to browser verification + the runner twin's tests. Reasonable given the lane constraint, but the pattern-side space-matching is currently unguarded — worth completing the browser-verification the PR itself flags (2+ profiles, one set default, confirm the default row shows the badge cross-space) before merge.

## Questions for the author

- Have you browser-verified the cross-space default badge (the PR's stated manual check)? That's the only path exercising the new `sameProfileCell` against real cross-space links.

Nice, tightly-scoped fix — the escape-hatch usage is well-precedented and the dead-code removal is clean.

PR #4513, opened 2026-07-03
feat(cli): blessed headless profile read (CT-1834)

bfollington:
## Why

Loom is demoting its observed-profile cache to a **witness/echo**: the clean architecture reads the profile **live through an identity** (runtime-enforced labels at read time) and falls back to the echo only for hot paths/outages. That needs a cheap, blessed **headless** read.

loom#3627 ask 4 wanted `profiles` / `defaultProfile` / `mru` exported from `@commonfabric/home-schemas` — but exporting raw fields invites consumers to **re-implement resolution**, the exact mistake that broke when #4415 changed `#profile` semantics. What the "cases that cannot wish" (offline caches, agents, scripts) actually need is a *blessed read path*, not raw fields.

## What

A first-class `cf wish <target>` verb (general, so it serves every future wish target — profiles just fall out) that:

- resolves through the **SAME** runner builtin patterns use (`packages/runner/src/builtins/wish.ts`), driven with `headless: true` so the suggestion / profile-picker UI never spins up. Resolution (default → MRU → first, labels enforced at read time) lives in the builtin; the CLI **never re-implements** it.
- supports `#profile` plus the scalar targets `#profileName` / `#profileAvatar` / `#profileBio` / `#profileSpace`. These resolve against the **identity's home space**, so `--space` is optional; other targets (`/`, `#mentionable`, …) are space-relative and take `--space`.
- runs as a real identity/session (`loadManager` → session-backed `Runtime`), so **labels are enforced against the reading identity**.
- prints the resolved value as JSON. A zero-profile `WishError` prints a clear message to **stderr and exits non-zero**; `--allow-empty` instead prints `null` and exits 0.

How identity/labels flow: `--identity` (or `CF_IDENTITY`) is loaded into a `createSession({ identity, spaceDid })`; `runtime.userIdentityDID` becomes the session signer, and the wish reads the home space of *that* identity. The read runs inside a trusted inline single-node pattern `() => ({ out: wish({ query, path, scope, headless: true }, schema) })`, so every label check happens against the real reading identity — no bypass.

`lib/wish.ts` splits a runtime-agnostic `resolveWish(runtime, space, spec)` core from the `readWish` wrapper that constructs the session-backed runtime, so the core is unit-testable against an emulated runtime with no live server.

### `cf wish` help
```
cf wish '#profile'      -i ./claude.key     # active profile object as JSON
cf wish '#profileName'  -i ./claude.key     # just the display name
[...]

PR #4537, opened 2026-07-06
fix(cli): project cf wish object output to plain data, drop stream-handle graph (CT-1844)

bfollington:
## What

`cf wish '#profile'` resolved correctly but printed ~58KB of JSON. The profile result **object** exposes the pattern's stream/handle keys (`addElement`, `addPiece`, `removeElement`, `setName`, `setAvatar`, `setBio`, `toggleEditing`) alongside its data keys (`$NAME`, `name`, `avatar`, `bio`, `elements`, `isEditing`, …). Those handles are live `Cell`/`Stream` objects, so JSON-serializing them dragged in the whole runtime object graph (scheduler, circular refs) — `addElement` alone was 28KB.

This projects the resolved value to plain, serializable data before output: every stream/cell/function-valued node is replaced with a stable `"[stream]"` marker; all genuine data (including the nested `elements` array and the `$UI` VNode tree) is kept intact.

## Why

The stated audience for this read path (CT-1834, #4513) is agents / scripts / offline profile caches. 58KB of runtime noise defeats that. Scalar targets (`#profileName`, `#profileAvatar`, `#profileBio`, `#profileSpace`) were already clean and are **unchanged** — a bare scalar passes straight through the projection.

## Mechanism

Grepped for a canonical "value → plain serializable" util first (anti-duplication). `convertCellsToLinks` exists but converts cells to *link refs* (different intent — it would still emit handle objects); `safeStringify` only guards circularity/depth. There is no existing "drop handles to plain data" helper, so this adds a small recursive `projectWishValue` in `lib/wish.ts` that keys off the runner's canonical `isCell`/`isStream` guards (exported from `@commonfabric/runner`) rather than hand-rolling structural detection. Applied in the CLI render path (`wishAction`) so only the object-target output shape changes.

## Before / after (live, a home space with a profile)

- `#profile`: **58560 → 1462 bytes**; data fields present (`name`, `bio`, `elements`, `$UI`, …), the 7 stream handles now render as `"[stream]"`.
- `#profileName` / `#profileBio`: unchanged (bare JSON scalar).

## Tests

`packages/cli/test/wish.test.ts`
- `projectWishValue strips stream handles but keeps profile data (CT-1844)` — builds a profile-shaped result carrying a **real** stream handle (via an `asCell: ["stream"]` schema), asserts data survives, the handle becomes `[stream]`, and the serialized output is small with no `scheduler`/`runtime`/`circular reference`.
- `projectWishValue passes scalar targets through unchanged (CT-1844)`.

`packages/cli/test/wish-command.test.ts`
- `projects an object result to plain data, dropping handle-valued keys (CT-1844)` — drives `wishAction` end-to-end through `render`, asserts data fields intact and handle keys marked.

All pre-existing scalar / zero-profile / `--allow-empty` tests pass unchanged. `deno fmt`/`lint`/`check` clean on cli.

Refs CT-1844 (follows CT-1834 / #4513).

---
## Summary by cubic
Fixes excessive JSON from `cf wish` object targets by projecting results to plain data and replacing stream/cell/function handles with markers like "[stream:setName]". Scalars pass through unchanged. Addresses CT-1844.

- **Bug Fixes**
  - Added `projectWishValue` to replace `Cell`/`Stream`/function nodes (via `isCell`/`isStream` from `@commonfabric/runner`) with key-tagged markers; memoizes projections so handles in shared subtrees are stripped on all paths and breaks real cycles as "[circular]".
  - Applied in `wishAction` so object results serialize to plain data; `#profile` drops from ~58KB to ~1.5KB while keeping fields like `name`, `bio`, `elements`, and `$UI`. Tests cover projection, key-tagged markers, DAG diamond, and cycle handling.

<sup>Written for commit 2252be1b7cf3b591724f2d306956e858c6420329. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4537?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cli/commands/wish.ts
+++ b/packages/cli/commands/wish.ts
@@ -6,7 +6,7 @@ import { render } from "../lib/render.ts";
 import { getDidFromFile } from "../lib/identity.ts";
 import { absPath } from "../lib/utils.ts";
 import { normalizeApiUrl, setQuietMode } from "./piece.ts";
-import { readWish } from "../lib/wish.ts";
+import { projectWishValue, readWish } from "../lib/wish.ts";
 
 /** Options the `cf wish` action receives (cliffy-parsed flags + env). */
 export interface WishCommandOptions {
@@ -91,7 +91,12 @@ export async function wishAction(
     return; // Reached only when a test injects a non-terminating exit.
   }
 
-  render(result, { json: true });
+  // Project away stream/cell handles before serializing. An object target
+  // (#profile) otherwise drags its pattern's stream handles — and through them
+  // the whole runtime object graph — into JSON (~50KB of noise). Scalar targets
+  // (#profileName etc.) and the null / --allow-empty result pass through
+  // unchanged. See projectWishValue (CT-1844).
+  render(projectWishValue(result), { json: true });
 }
 
 const description = cliText(
--- a/packages/cli/lib/wish.ts
+++ b/packages/cli/lib/wish.ts
@@ -155,8 +155,13 @@ export async function readWish(
   });
 }
 
-/** Marker substituted for a stream/handle-valued key in projected output. */
+/** Base marker substituted for a stream/handle-valued node in projected output. */
 export const WISH_STREAM_MARKER = "[stream]";
+
+/** Marker for a stripped handle, optionally tagged with the key it hung off. */
+function streamMarker(key?: string): string {
+  return key === undefined ? WISH_STREAM_MARKER : `[stream:${key}]`;
+}
 
 /**
  * Project a resolved wish value to plain, serializable data for output (CT-1844).
@@ -170,27 +175,51 @@ export const WISH_STREAM_MARKER = "[stream]";
  * exists for.
  *
  * This walks the value and replaces every stream/cell/function-valued node with
- * {@link WISH_STREAM_MARKER}, keeping all plain data (including nested arrays
- * like `elements` and the `$UI` VNode tree) intact. Scalar results (a bare
- * string/number/etc. from `#profileName` and friends) pass straight through
- * unchanged, so only the object-target output shape is affected.
+ * a `[stream]` / `[stream:<key>]` marker, keeping all plain data (including
+ * nested arrays like `elements` and the `$UI` VNode tree) intact. Scalar results
+ * (a bare string/number/etc. from `#profileName` and friends) pass straight
+ * through unchanged, so only the object-target output shape is affected.
+ *
+ * The walk memoizes the PROJECTED result per input node in a `Map` (with an
+ * in-progress sentinel so genuine cycles terminate rather than recurse forever).
+ * That is load-bearing correctness, not just efficiency: a plain object reached
+ * by two different paths (a DAG "diamond") must be projected on BOTH — a bare
+ * "already seen → return raw" dedup would leak any handle nested under the
+ * shared subtree on the second path, re-exposing exactly the graph this strips.
  */
-export function projectWishValue(
+export function projectWishValue(value: unknown): unknown {
+  return projectNode(value, undefined, new Map<object, unknown>());
+}
+
+/** In-progress sentinel: a node currently being projected higher in the stack. */
+const IN_PROGRESS = Symbol("wish-projection-in-progress");
+
+function projectNode(
   value: unknown,
-  seen: WeakSet<object> = new WeakSet(),
+  key: string | undefined,
+  memo: Map<object, unknown>,
 ): unknown {
   if (isCell(value) || isStream(value) || typeof value === "function") {
-    return WISH_STREAM_MARKER;
+    return streamMarker(key);
   }
   if (value === null || typeof value !== "object") return value;
-  if (seen.has(value)) return value; // let downstream stringify flag circularity
-  seen.add(value);
-  if (Array.isArray(value)) {
-    return value.map((item) => projectWishValue(item, seen));
+
+  const cached = memo.get(value);
+  if (cached === IN_PROGRESS) {
+    // A genuine cycle: this node is an ancestor of itself. Break it so the walk
+    // terminates; downstream stringify would otherwise flag the same loop.
+    return "[circular]";
   }
-  const out: Record<string, unknown> = {};
-  for (const [key, val] of Object.entries(value)) {
-    out[key] = projectWishValue(val, seen);
-  }
-  return out;
+  if (cached !== undefined) return cached; // diamond: reuse the SAME projection.
+
+  memo.set(value, IN_PROGRESS);
+  const projected: unknown = Array.isArray(value)
+    ? value.map((item, i) => projectNode(item, String(i), memo))
+    : Object.fromEntries(
+      Object.entries(value).map((
+        [k, val],
+      ) => [k, projectNode(val, k, memo)]),
+    );
+  memo.set(value, projected);
+  return projected;
 }
--- a/packages/cli/lib/wish.ts
+++ b/packages/cli/lib/wish.ts
@@ -1,6 +1,8 @@
 import type { DID } from "@commonfabric/identity";
 import {
   createBuilder,
+  isCell,
+  isStream,
   type JSONSchema,
   type MemorySpace,
   type Runtime,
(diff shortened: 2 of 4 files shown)


bfollington (commented):
_cf-review via Claude `claude-opus-4-8`, on behalf of @bfollington (automated self-signed review)._

**Scope** · `cf wish` object-target output projection (CT-1844): new `projectWishValue` in `packages/cli/lib/wish.ts` recursively replaces cell/stream/function nodes with `"[stream]"`, wired into `wishAction`. · **Coverage** deep-read (2 packages, small diff) · **Checks run** `deno test test/wish.test.ts test/wish-command.test.ts` (18 steps green, incl. the 2 new projection tests + CLI e2e), `deno fmt --check`, `deno lint` — all clean.

Solid, well-scoped fix with genuinely good tests (the `asCell:["stream"]` fixture exercises a *real* handle, not just a stand-in). Anti-duplication and the scalar/`--allow-empty` guarantees hold up. One narrow, suspected correctness gap in the recursion.

## 🟡 Improvements

### 1. `seen` WeakSet conflates cycle-breaking with shared-subtree dedup → a handle under a re-referenced data object can leak unstripped (suspected)

`packages/cli/lib/wish.ts` — `projectWishValue`:

```ts
if (seen.has(value)) return value; // let downstream stringify flag circularity
seen.add(value);
```

`seen` is never popped, so it doesn't just break cycles — it also fires on a **DAG diamond**: a plain data object reachable by two *different* paths (not a cycle). On the second path the object is returned **raw, unprojected**. Because the `isCell/isStream/function` check sits *before* the `seen` check, a handle that is itself a direct value is always stripped regardless of path — but a handle nested *inside* such a shared sub-object escapes on the second path. Reproduced in isolation:

```
root = { a: { child: shared }, b: { child: shared } }, shared.handle = <stream>
→ a.child.handle: "[stream]"   b.child.handle: {"__isStream":true}   // leaks
```

Whether a materialized `.get()` profile value actually contains a shared (aliased) sub-object reachable by two keys is uncertain — I couldn't confirm it does in the real `#profile` shape, hence **suspected / low-likelihood**, not Blocking. But the projection is not robustly correct against aliasing, and the failure mode is exactly the 58KB-graph leak this PR exists to prevent. Cheap fix: memoize the *projected* result instead of the input (`Map<object, unknown>` returning the already-projected node), or track only the active recursion path (add on descent, delete on ascent) so true cycles break but DAG shares re-project. A memo is also a small perf win on wide shared trees.

### 2. `safeStringify` maxDepth=8 can truncate a deep `$UI` tree (pre-existing, note only)

`render(projectWishValue(result), { json: true })` → `safeStringify(…, maxDepth = 8)`, which emits `"<max depth reached>"` past depth 8. A deeply nested `$UI` VNode tree could hit that. **Not introduced by this PR** — the cap predates it (render.ts, 2026-06-19) and applied identically to the old bloated output — so no regression. Flagging only because the PR's stated promise is "the `$UI` VNode tree kept intact": for a shallow profile it is, but a deep one is silently clipped by the renderer, not by the projection. Not worth changing here.

## Questions for the author

- **`"[stream]"` marker vs a stable link ref, for the agent/offline-cache audience (d).** Dropping the handle to a bare `"[stream]"` string discards both the key's *kind* (cell vs stream vs fn — all collapse to the same token) and any resolvable identity. For the CT-1834 consumer (Loom's observed-profile cache, agents), was a re-resolvable link ref considered? `convertCellsToLinks` (`packages/runner/src/cell.ts`) already produces canonical link refs — deliberately *not* reused here (correctly, per the anti-dup note below, since its output is still verbose handle objects), but it's the closest thing to "machine-consumable handle". `"[stream]"` is a fine *human* marker; confirming the downstream consumer only needs data-fields-present (not handle round-tripping) would close this. If so, keeping the key name (`"[stream:setName]"`) costs nothing and preserves which handles existed.

## Notes (verified, no action)

- **Anti-duplication ✅.** Reusing `isCell`/`isStream` from `@commonfabric/runner` (both `instanceof CellImpl`; `isStream` = CellImpl + `isStream()`) is the right guard. I checked the three plausible canonical homes and confirmed none fits: `value-debug` emits *debug strings* (wrong output type — CLI needs a JSON object with data fields addressable), `codec-json` *preserves/encodes* cells into the wire format (opposite intent), `convertCellsToLinks` converts handles to *link-ref objects* (still verbose, keeps handle objects). The author's "no drop-to-plain-data helper exists" claim holds; this is not a silent fork.
- **Data survives, handles dropped at any depth ✅.** The `isCell || isStream || function` check keys off the `.get()` materialization boundary, where schema `asCell`/`asStream` nodes stay as `CellImpl` instances and everything else is already plain data. A plain object that merely *looks* cell-shaped (e.g. `{ $stream: true }` as data) is **not** stripped — the guard is strict `instanceof`, verified by the test's plain `$UI`/`elements` surviving alongside the real handle.
- **Scalar / zero-profile / `--allow-empty` byte-identical ✅ (c).** A bare scalar (`string`/`number`/`boolean`) returns from the first two branches untouched; `null` returns via `value === null`. The `--allow-empty` path serializes `null` → identical. The 9 pre-existing command tests pass unchanged.

## Verdict

**Non-blocking.** Clean, correct for the common case, well-tested. Recommend the author fix the DAG-diamond `seen` handling (Improvement 1 — cheap, and it guards the exact leak this PR targets) and confirm the `"[stream]"` marker suffices for the CT-1834 consumer before merge; neither blocks.

bfollington:
Addressed the cf-review finding (DAG-diamond handle leak) in `2252be1b7`.

**Root cause confirmed:** the projection's `seen` WeakSet was added-to on entry but never removed, so it doubled as shared-subtree dedup — a plain object reached by two paths returned RAW on the second, leaking any handle nested inside it.

**Fix (approach b — memoization):** `projectWishValue` now memoizes the *projected result* per input node in a `Map`, with an `IN_PROGRESS` sentinel so genuine cycles still terminate (→ `"[circular]"`) instead of recursing forever. Diamonds reuse the same projection on every path; cycles break. Chose memoization over add/remove path-set since it handles both cases and is more efficient.

**Also (reviewer nicety):** the marker now carries the owning key — `"[stream:setName]"` — for easier Loom/agent debugging. Base `"[stream]"` (`WISH_STREAM_MARKER`) is used when a handle has no owning key.

**Tests added** (`packages/cli/test/wish.test.ts`):
- `strips handles nested in a shared DAG subtree on BOTH paths` — mirrors the isolation repro: one object aliased under two keys with a real stream handle inside; asserts the handle is `[stream:setName]` on both `left` and `right` (leaked on `right` before).
- `terminates on a genuine cycle without infinite recursion` — self-referential object → `[circular]`, no runaway.

All 20 wish test steps pass; fmt/lint/check clean. Live `#profile` unchanged in size, handles now show `[stream:<key>]`.

bfollington (commented):
cf-review via Claude (claude-opus-4-8), on behalf of @bfollington — automated review, human owns the merge.

**Scope** · `cf wish` object-output projection (CT-1844): 4 files in `packages/cli` (+257/−4) — `lib/wish.ts`, `commands/wish.ts`, and two test files. Clean, coherent scope; the runner/traverse/patterns files in the raw `main` diff-stat are pre-merged base, not part of this PR's delta. · **Coverage:** deep-read all four files. · **Checks run (verified):** `deno test packages/cli/test/wish*.test.ts` → 20/20 green (incl. the 4 new projection cases); `deno check` / `deno lint` / `deno fmt --check` on all touched files → clean.

## Verdict: clean — the memoization rework is correct. No blocking findings.

### The memoization rework is sound

I checked the three points that matter for this rework and they hold:

- **DAG diamond (shared plain subtree, no raw handle leak):** first path stores `IN_PROGRESS`, projects fully, then stores the finished projection; the second path finds a *completed* projection (not the sentinel) and reuses the SAME already-stripped object. No raw input node is ever returned once fully projected, so a handle nested under a shared subtree is stripped on every path. The `strips handles nested in a shared DAG subtree on BOTH paths` test aliases one real-stream-bearing object under `left`/`right` and asserts `[stream:setName]` on both — it genuinely guards the exact leak the prior version had. **verified**
- **Genuine cycle (terminates, `[circular]`):** a self-ancestor is still `IN_PROGRESS` when re-encountered → returns `[circular]`, breaking the back-edge; the walk terminates. The `terminates on a genuine cycle` test (`node.self = node`) passes and serializes cleanly. **verified**
- **Map key correctness (no wrong cached projection):** the memo is keyed on input-node *reference identity*, and a projection is only reused for the identical input object, so it can't return a projection built from a different node. The `key` argument (used solely for the marker tag) is deliberately NOT part of the memo key — and that's correct, because only leaf handles (cell/stream/function) consume `key`, and those hit the first guard and return immediately, never memoized. A shared subtree reached under two different outer keys correctly reuses one projection while its *interior* markers stay tagged by their own local keys. **verified**

### Regression vs the prior version — fixed, strict improvement

The first CT-1844 commit did `if (seen.has(value)) return value` — returning the **raw** input on any second visit, which leaked handles nested under any DAG diamond (and `seen` is never cleared, so this fired for real diamonds, not just cycles). The rework replaces the raw-dedup with per-node projection memoization + an in-progress sentinel, which both closes that leak and still terminates on true cycles. Markers are now key-tagged (`[stream:setName]`), which is strictly more informative and asserted by the tests. Net: correctness up, no behavior lost.

### Scalar / null / `--allow-empty` stay byte-identical — confirmed

`projectWishValue` is identity for `null` and any `typeof value !== "object"`, so `#profileName`-style scalars, `null`, and the `--allow-empty → null` path serialize exactly as before. The scalar passthrough test and the pre-existing `--allow-empty` test both pass unchanged.

### Anti-duplication — not a re-fork

The bespoke walk isn't a duplication of a canonical home. `data-model/value-debug` produces debug *strings* (compact/indented) for logs; this path needs a real JSON-serializable *value* whose data fields, nested arrays and `$UI` VNode tree round-trip through `render(..., {json:true})` with only handles substituted. Different concern, justified. `Object.entries` skipping symbol/non-enumerable keys is consistent with the downstream `safeStringify` (also `Object.entries`-based), so no handle can leak through a key the serializer wouldn't have walked either.

### Tests — they guard the right principles

The four new cases map 1:1 to the semantic change: (1) real-stream-handle stripping with data preserved, (2) the DAG-diamond both-paths leak, (3) cycle termination, (4) scalar passthrough + untagged base-marker fallback. They read like user stories, use real runtime stream handles (not just function stand-ins) for the fabric cases, and assert the negative invariants that matter (`not.toContain("scheduler"/"runtime"/"circular reference")`, bounded output size). Good rigor.

## Nits (optional, non-blocking)

- `lib/wish.ts` — `projectNode`'s `if (cached !== undefined) return cached;` relies on a projected value never legitimately being `undefined`. It can't be today: leaf handles short-circuit before memoization, and object/array projections are always a fresh object/array, so a memoized entry is always object-or-`IN_PROGRESS`, never `undefined`. Worth a one-line comment pinning that assumption so a future edit that memoizes scalars doesn't silently reproject. (info)

— Posted automatically by the cf-review skill.

PR #4539, opened 2026-07-06
fix(patterns): write default-profile as a pure link for Set default CFC authorization (CT-1845)

bfollington:
## Why

Clicking **Set default** in the profile picker failed permanently and silently. Console:

```
CFC enforcement rejected commit: relevant transaction was not prepared: writeAuthorizedBy failed at /avatar
```

Setting a default is core picker functionality and had **zero test coverage** (`setDefaultProfile` is untouched since #3830, 2026-06-09; `shared-profile.test.ts` only clicks CreateProfile). Links CT-1845.

## Root cause (case a — contained, profile-side write-shape bug)

The home's `defaultProfile` is a **single** owner-protected slot whose declared schema is `TrustedDefaultProfile` = `Cell<ProfileHomeOutput>` (a link, `writeAuthorizedBy: setDefaultProfile`). But CFC resolves that link `$ref` and **walks ProfileHomeOutput's own fields** — `name` / `avatar` / `bio` / `elements` — each owner-protected with its *own* `writeAuthorizedBy: set…` (setName/setAvatar/setBio/mutateElements).

CFC enforces a nested field **only when the written value carries a concrete value there** (`ifcEntryAppliesToAttemptedWrite`). The old body

```ts
defaultProfile.set(profile as any);
```

**inlines** when `profile` is a *loaded* Cell: `.set()`'s value walk (`recursivelyAddIDIfNeeded`) passes through only already-serialized link sigils, not live `Cell` objects, so a `Cell` written into a scalar slot is expanded to its resolved value (with a concrete `/avatar`). CFC then enforces the nested `/avatar` claim against the acting writer — `setDefaultProfile` ≠ `setAvatar` — and rejects at `walkIfcSchema`/`writeAuthorizedByReason` (`prepare.ts`).

In the browser the picker's UI render (`cf-cell-link` / `cf-profile-badge` on each row) materializes the profile value into the picker's reactive graph, so the write inlines and faults. **`setMruProfile` dodged this only incidentally** — it writes into an array (`mru.set([profile, …])`) whose elements happen to serialize as links; the single-slot default write does not.

This is **not** a CFC-epic regression: the CFC nested-owner-protection enforcement is working as designed; the profile-side write shape is wrong.

## What

`setDefaultProfile` now writes the serialized **link sigil**:

```ts
defaultProfile.set((profile as any).getAsLink() as any);
```

`getAsLink()` returns the `link@1` sigil, which the `.set()` value walk passes through verbatim → the slot stores a pure reference with **no inlined sub-fields**, so no nested owner-protected claim is ever enforced. Profile identity stays the profile's own space (CT-1842); `#profile` resolves the target on read.

## Test

`packages/runner/test/profile-set-default-cfc.test.ts` drives the **real** shipped patterns (`profile-create.tsx` + `profile-home.tsx`) in a real cross-space runtime — two profiles in their own `inSpace` spaces, loaded — and exercises the fixed `setDefaultProfile` through **both** the fixed-index binding and the **exact picker `.map((p) => …)` binding**, asserting each:
- commits cleanly (no CFC rejection),
- stores a **pure link** (`isPureLinkSigil`, no inlined `/avatar`),
- and that the default re-resolves to a profile's own space (CT-1842 identity-by-space).

### Reproducibility caveat (honest)

The browser-only rejection requires the **UI render path** to materialize the profile value into the picker's reactive graph, so that `.set(<Cell>)` inlines. In the headless runner the profile value is loaded into the runtime but not threaded through a UI render into the pattern's reactive graph, so `.set(<Cell>)` consistently serializes a link and the *pre-fix* handler does not fault headlessly. The test therefore **guards the fix's invariant directly** — the write stores a pure link, which is the exact property that makes the nested `/avatar` claim inapplicable — rather than the render-triggered rejection. A closing `NOTE` in the test documents this. The fix was reasoned to correctness from the `.set()`/`getAsLink()` serialization semantics; final in-browser confirmation is recommended.

## Verify

- `deno test -A packages/runner/test/profile-set-default-cfc.test.ts` — passes.
- `deno test` wish + profile-create-real-card-add — 5 passed / 108 steps, no regression.
- `deno fmt` / `deno lint` / `deno check` clean on both changed files; `cf check profile-picker.tsx` clean.

---
## Summary by cubic
Fixes CT-1845 by writing a pure link to the home `defaultProfile` slot when “Set default” is clicked, preventing CFC authorization errors and restoring the picker action. Adds a cross-space test to lock in pure-link writes and correct identity resolution.

- **Bug Fixes**
  - Update `setDefaultProfile` in `packages/patterns/system/profile-create.tsx` to write `profile.getAsLink()` so the slot stores a `link@1` (no inlined fields), avoiding nested `/avatar` owner checks that caused “writeAuthorizedBy failed at /avatar”.
  - Add `packages/runner/test/profile-set-default-cfc.test.ts` to verify both fixed-index and picker `.map` bindings commit cleanly, store a pure link, and resolve to the profile’s own space.
  - No migration required; MRU path unchanged.

<sup>Written for commit 46c50db81de504fbc50831a521d811bc4e079752. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4539?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/patterns/system/profile-create.tsx
+++ b/packages/patterns/system/profile-create.tsx
@@ -112,9 +112,28 @@ export const setDefaultProfile = handler<
     profile: Cell<ProfileHomeOutput>;
   }
 >((_, { defaultProfile, profile }) => {
-  if (profile) {
-    defaultProfile.set(profile as any);
-  }
+  if (!profile) return;
+  // Write only the serialized LINK sigil, never the inlined profile value.
+  // `defaultProfile` is a SINGLE owner-protected slot whose declared schema is
+  // `Cell<ProfileHomeOutput>` — a link. But CFC resolves that `$ref` and walks
+  // ProfileHomeOutput's own fields, each of which is owner-protected
+  // (name/avatar/bio/elements carry their own `writeAuthorizedBy: set…`). CFC
+  // only *enforces* a nested field when the WRITTEN VALUE carries a concrete
+  // value there — so a pure link is fine, but an inlined object trips
+  //   "writeAuthorizedBy failed at /avatar"
+  // because this picker write is authorized by `setDefaultProfile`, not
+  // `setAvatar` (CT-1845).
+  //
+  // A plain `defaultProfile.set(profile)` INLINES: `.set()`'s value walk only
+  // passes through already-serialized link sigils (`recursivelyAddIDIfNeeded`),
+  // not live `Cell` objects, so a `Cell` written into a scalar slot is expanded
+  // to its resolved value (with `/avatar`). `setMruProfile` avoids this only
+  // incidentally — its array elements happen to serialize as links. Here we
+  // make the link explicit: `getAsLink()` returns the `link@1` sigil, which the
+  // value walk passes through verbatim, so the slot stores a pure reference
+  // with no inlined sub-fields. Identity stays the profile's own space
+  // (CT-1842); `#profile` resolves the target on read.
+  defaultProfile.set((profile as any).getAsLink() as any);
 });
 
 // Stamps a profile as most-recently-used: prepend to the MRU list (deduped by
--- a/packages/runner/test/profile-set-default-cfc.test.ts
+++ b/packages/runner/test/profile-set-default-cfc.test.ts
@@ -1,0 +1,331 @@
+import { afterEach, beforeEach, describe, it } from "@std/testing/bdd";
+import { expect } from "@std/expect";
+import { Identity } from "@commonfabric/identity";
+import * as MemoryV2Server from "@commonfabric/memory/v2/server";
+import { fromFileUrl } from "@std/path";
+
+import { EmulatedStorageManager } from "../src/storage/v2-emulate.ts";
+import type { Options } from "../src/storage/v2.ts";
+import { Runtime } from "../src/runtime.ts";
+import type { RuntimeProgram } from "../src/harness/types.ts";
+import { TEST_MEMORY_SERVER_AUTH } from "./memory-v2-test-utils.ts";
+
+// CT-1845: "Set default" in the profile picker.
+//
+// Clicking "Set default" fires `setDefaultProfile`, whose body writes the
+// chosen profile into the home's single owner-protected `defaultProfile` slot
+// (`TrustedDefaultProfile` = `Cell<ProfileHomeOutput>`, writeAuthorizedBy:
+// setDefaultProfile). In the browser this failed permanently and silently:
+//   CFC enforcement rejected commit: relevant transaction was not prepared:
+//   writeAuthorizedBy failed at /avatar
+//
+// Root cause: `TrustedDefaultProfile`'s declared schema is a link to
+// ProfileHomeOutput, but CFC resolves that `$ref` and walks ProfileHomeOutput's
+// OWN fields — name/avatar/bio/elements, each owner-protected with its own
+// `writeAuthorizedBy: set…`. CFC enforces a nested field only when the WRITTEN
+// VALUE carries a concrete value there. A pre-fix `defaultProfile.set(profile)`
+// where `profile` is a *loaded* Cell inlines the resolved profile value (with a
+// concrete `/avatar`), so the nested `/avatar` claim is enforced against the
+// wrong writer (setDefaultProfile ≠ setAvatar) and the commit is rejected. The
+// fix writes the serialized LINK sigil (`profile.getAsLink()`), which the
+// `.set()` value walk passes through verbatim — a pure reference with no
+// inlined sub-fields, so no nested claim is ever enforced.
+//
+// This test drives the REAL shipped patterns in a real cross-space runtime
+// (profile-create.tsx handlers + profile-home.tsx). It stands up TWO profiles
+// in their own inSpace spaces, loads them, then exercises the fixed
+// `setDefaultProfile` through BOTH the fixed-index and the exact picker `.map`
+// binding, asserting each commits cleanly and stores a PURE LINK (never an
+// inlined object). See the note at the bottom on why the browser-only inline
+// fault is not directly reproducible in the headless runner.
+const signer = await Identity.fromPassphrase("profile-set-default-cfc");
+const spaceA = signer.did();
+
+class SharedServerStorageManager extends EmulatedStorageManager {
+  static connectTo(
+    server: MemoryV2Server.Server,
+    options: Omit<Options, "memoryHost" | "spaceHostMap">,
+  ): SharedServerStorageManager {
+    const manager = new SharedServerStorageManager(
+      { ...options, memoryHost: new URL("memory://") },
+      () => server,
+    );
+    manager.sharedServer = server;
+    return manager;
+  }
+  private sharedServer!: MemoryV2Server.Server;
+  protected override server(): MemoryV2Server.Server {
+    return this.sharedServer;
+  }
+}
+
+const newSharedServer = () =>
+  new MemoryV2Server.Server({
+    authorizeSessionOpen(message) {
+      const principal = (message.authorization as { principal?: unknown })
+        ?.principal;
+      return typeof principal === "string" ? principal : undefined;
+    },
+    sessionOpenAuth: TEST_MEMORY_SERVER_AUTH.sessionOpenAuth,
+  });
+
+const sysDir = fromFileUrl(
+  new URL("../../patterns/system/", import.meta.url),
+);
+const read = (n: string) => Deno.readTextFileSync(sysDir + n);
+
+// A wrapper owning the home `profiles`, `defaultProfile`, and `mru` cells,
+// mirroring home.tsx's profile model. The `defaultProfile`/`mru` slots carry
+// the picker's owner-protected contracts through the pattern OUTPUT schema
+// (TrustedDefaultProfile / TrustedProfileMru), so the CFC write-policy check on
+// `setDefaultProfile` / `setMruProfile` is exercised exactly as in home. It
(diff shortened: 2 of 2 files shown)


bfollington (commented):
_cf-review via Claude (claude-opus-4-8), on behalf of @bfollington — automated review, self-signed._

**Scope** · 1-line fix in `packages/patterns/system/profile-create.tsx` + 1 new test (`packages/runner/test/profile-set-default-cfc.test.ts`). **Coverage:** deep-read (fix, test, cited CFC/cell code, transformed schema), one delegated runtime-binding trace. **Checks run:** new test on branch (pass), test with fix reverted (still pass — see Blocking #2), `deno task cf check --show-transformed` on `setDefaultProfile`.

## What the change does
`setDefaultProfile` writes `defaultProfile.set(profile.getAsLink())` instead of `.set(profile)`, to stop a browser-only `CFC … writeAuthorizedBy failed at /avatar` rejection on "Set default". The fix and its *intent* (store a pure link, never an inlined profile object) are sound. But the **root-cause explanation** — now a permanent ~20-line code comment and the test header — is inaccurate about the mechanism, and the test does not guard the regression.

---

## 🔴 Blocking

### 1. The root-cause narrative in the code comment is factually wrong about `.set()` / `recursivelyAddIDIfNeeded` (verified)
`profile-create.tsx:127-130` states: *"`.set()`'s value walk only passes through already-serialized link sigils (`recursivelyAddIDIfNeeded`), not live `Cell` objects, so a `Cell` written into a scalar slot is expanded to its resolved value."* This is contradicted by the cited code:
- `recursivelyAddIDIfNeeded` (`cell.ts:2917`): `if (isCellLink(value)) return value;`, and `isCellLink` (`link-utils.ts:65-73`) returns true for `isCell(value)` — **a genuine `Cell` passes through unchanged**.
- Even without that guard, a `Cell` has `toJSON(): SigilLink` (`cell.ts:2545`), so `shallowFabricFromNativeValue` routes it via `NATIVE_TAGS.HasToJSON` → `toJSON()` → link (`native-conversion.ts:172-188`).

So `.set(<a genuine loaded Cell>)` serializes to a **link, not an inlined value**. Inlining happens only when the argument is a **non-`Cell` resolved/proxy value** (a plain object with no `toJSON`, taking the `NATIVE_TAGS.Object` branch).

And here `profile` **is** a genuine Cell. The transformed schema shows:
```js
profile: { $ref: "#/$defs/ProfileHomeOutput", asCell: ["readonly"] }
```
An `asCell` property is materialized as a real `CellImpl` (`schema.ts:1266` `createCell`; golden `ts-transformers/…/simple-handler.expected.jsx:31-34`). Under that schema, the **old** `defaultProfile.set(profile)` already writes a pure link.

The `ifcEntryAppliesToAttemptedWrite` half of the story is correct (`prepare.ts:2385-2405`: a nested owner-protected field is enforced only when a concrete value is written there). But the "how a concrete `/avatar` gets written" half is wrong: with a genuine-Cell `profile` it cannot arise as described. Either the browser path delivered `profile` as a *non-Cell resolved value* (schema/binding differing from the headless transform — not demonstrated), or the observed rejection came from a different mechanism.

Because this is a permanent comment future readers will trust, please correct it: re-describe the trigger as "when `profile` arrives as a non-Cell materialized value" and state when that happens, or drop the mechanistic claim and keep only the invariant ("we write a link explicitly, so the outcome is independent of how `profile` was materialized").

### 2. The new test cannot distinguish the fix from the bug — it passes with the fix reverted (verified — ran both ways)
I restored the old body `defaultProfile.set(profile as any)` and re-ran `profile-set-default-cfc.test.ts`: **it still passes** (pure-link assertion holds, commit clean). This matches the author's honest NOTE — headlessly `profile` is already a Cell, so old and new code both store a link. The test guards an invariant the *buggy* code already satisfies in this harness → ~no regression protection for CT-1845, at the cost of 331 lines driving the real patterns cross-space.

Fix: either (a) find a headless way to force the non-Cell/inlined `profile` the browser hits (bind `profile` from a plain-value read without `asCell`, or thread it through a resolve so `.set` receives a proxy), so the pre-fix code actually faults and the test turns **red on revert**; or (b) if that truly can't be reproduced headlessly, shrink to a focused CFC-layer repro (inlined object carrying a nested `writeAuthorizedBy: setAvatar` sub-field written under `setDefaultProfile` authority → assert rejection) and state plainly that the pattern-level test only documents the invariant. As written, the test is documentation, not a guard — and its header repeats the incorrect mechanism from #1.

Neither blocker is a defect in the *shipped fix* (which is safe). They block on the review's own terms — a permanent authoritative comment left actively wrong, and a test that can't distinguish fixed from broken.

## 🟡 Improvements
- **Loose cross-reference** (verified): the NOTE says the inlined-write rejection "is verified at the CFC layer … in `cfc-boundary.test.ts` semantics." That file broadly exercises `writeAuthorizedBy`, but I found no case that writes a concrete value into a *nested* owner-protected sub-field under a link-typed scalar slot and asserts rejection at that nested path. Add that case or soften the reference.
- **Fix is well-idiomed** (verified, positive): `getAsLink()` (`cell.ts:2027`) returns a `SigilLink` passed through unconditionally as a link, matching `setMruProfile`'s link-first discipline and wish.ts's identity-based read (`resolveAsCell()`/`.equals()`, `wish.ts:379-408`). Read-side is clean: the reader supplies its own `profileLinkSchema` on read, resolves by link identity, and the pure link resolves to the profile's own space — no schema/type loss, CT-1842 preserved. Keep the fix.

## Sibling inlining sites
No other latent instance of this class. The only owner-protected *single-slot* profile write is `defaultProfile`; `setMruProfile` writes an array-valued slot — but note its dodge is **not** "array elements happen to serialize as links" (arrays and scalars are walked identically in `recursivelyAddIDIfNeeded`); it's the same Cell-passthrough. `home.tsx` `entry.set({…})` (favorites/spaces, lines 93/147) inline objects, but `Favorite`/`SpaceEntry` are plain `Writable<…[]>` with no `Cfc`/`WriteAuthorizedBy`/nested cross-handler protection, so no CFC conflict is possible. `profile-home.tsx` `.set()` sites write scalars into their own owner-protected fields under the matching handler. Nothing else to flag.

## Question for the author
Did you confirm in-browser (e.g. logging `isCell(profile)` / `profile.constructor.name`, or the picker's transformed `argumentSchema`) that in the failing run `profile` actually arrived as a **non-Cell** value? The headless transform shows `asCell: ["readonly"]` on `profile`, under which the *old* code already writes a link. Pinning this down is what makes the comment correct and tells us whether the fix cures the bug or hardens against a materialization that differs from headless.

## Verdict
**Blocking (soft).** Ship-safe one-line fix, correct idiom, correct read-side compatibility — but the embedded root-cause comment misdescribes `.set()`/`recursivelyAddIDIfNeeded` (they pass Cells through as links), and the test passes with the fix reverted, so it guards nothing. Correct the mechanism narrative and either make the test fail-on-revert or reframe it honestly; the in-browser confirmation you recommend is genuinely necessary to establish that the bug reproduces at all.

bfollington:
## ⚠️ DO NOT MERGE — fix verified NOT working in-browser (CT-1845)

An in-browser reproduction (isolated offset local-dev serving this branch, two fresh identities) found this fix **does not resolve the bug**, and both root-cause narratives (this PR's "inlining" story and the review's "profile is a non-Cell proxy" hypothesis) are **wrong**:

- **`profile` is a live Cell in every trial** (has `getAsLink`, internal `_link`/`_frame`/`tx`). `getAsLink()` was callable and did not throw.
- The stored `defaultProfile` was **always a pure `$link` sigil**, even with the original `.set(profile as any)` code — so no inlining ever happened. The old code already wrote a link.
- With the `getAsLink()` fix served and confirmed (`curl` of the pattern), the **exact same failure reproduced**.

**The real trigger:** it's **overwrite-specific.** Setting `defaultProfile` from *empty* succeeds; **overwriting an already-set default** with a different profile fails with `writeAuthorizedBy failed at /avatar` (5 retries, silent — no UI feedback). Reproduced twice on fresh identities.

**Likely root cause:** `ifcEntryAppliesToAttemptedWrite` in `packages/runner/src/cfc/prepare.ts` (~2338-2380) — overwriting an owner-protected link appears to read the **old** linked target's `/avatar` during the write-applicability/diff step, which the fresh-write never does. This is CFC write-semantics, not a profile-side inlining issue.

**The test here is vacuous** (asserts a pure-link write, which passed even in the failing browser case). A real test must drive the **overwrite**: set default → A (commit), then set default → B while A is stored, assert the second commit succeeds.

Converting to draft. Next: reproduce the overwrite failure **headlessly** (the existing test only did first-write) and either find a clean profile-side workaround (clear-then-set, since first-write-from-empty works) or hand the precise characterization to the CFC owner. — automated finding relayed by Claude on behalf of @bfollington

bfollington:
Superseded by #4540. This branch's `getAsLink()` approach was disproven by in-browser reproduction (the stored value was always a pure link; the walk fired regardless), and the real root cause turned out to be the walkable `ProfileHomeOutput` schema on the durable `defaultProfile` slot triggering a `/avatar` CFC walk on overwrite. Fix + red-without-fix tests are in #4540.

PR #4540, opened 2026-07-06
fix(patterns): Set default overwrite CFC authorization (CT-1845)

bfollington:
## CT-1845 — "Set default" overwrite fails CFC `writeAuthorizedBy` at /avatar

Fixes CT-1845. Supersedes the disproven approach in **#4539** (whose `getAsLink()` fix was VERIFIED IN-BROWSER TO NOT WORK — see below); that PR is now draft.

### The bug (overwrite-specific)
Clicking "Set default" in the profile picker to **change** an already-set default failed permanently and silently in-browser:

```
StorageTransactionAborted: CFC enforcement rejected commit: relevant transaction
was not prepared: writeAuthorizedBy failed at /avatar
```

- Setting `defaultProfile` from **EMPTY** (never-set) → **succeeds**.
- **OVERWRITING** an already-set `defaultProfile` with a **different** profile → **fails** (5 retries, default unchanged).

### Root cause (`packages/runner/src/cfc/prepare.ts`)
The home `defaultProfile` is a **single owner-protected slot**. Its durable cell was declared `Writable<ProfileHomeOutput | undefined>`, so the stored default **link carries that walkable `ProfileHomeOutput` schema**. CFC (`walkIfcSchema`) walks it on every write to the slot and emits owner-protected entries for the linked profile's OWN fields — `/name`, `/avatar`, `/bio`, each `writeAuthorizedBy: set…` (e.g. `/avatar` → `setAvatar`).

Overwriting the default with a different profile changes the container link, and `ifcEntryAppliesToAttemptedWrite` marks the nested `/avatar` **"touched"**; its RESOLVED value is a concrete string, so the entry **APPLIES** and CFC enforces `/avatar`'s `setAvatar` authorization against the **picker** writer (`setDefaultProfile` ≠ `setAvatar`) → reject. A first write from empty has no prior resolved `/avatar` to touch, so it passes — hence overwrite-specific. The **MRU array** dodges this because CFC checks its owner-protected element under a wildcard `*` per *changed element*, not as a single walked container.

**Why #4539's `getAsLink()` was wrong:** the stored value is always a pure `$link` sigil regardless, and the link still carried the walkable schema — so the walk fired either way. `getAsLink` was never the lever.

### The fix (profile-side, minimal)
Declare the durable `defaultProfile` cell with a new **opaque** `DefaultProfileCell` type (`Writable<Record<never, never> | undefined>`) in `home.tsx`, so the stored slot schema carries **no walkable owner-protected sub-field** and no nested `/avatar` claim exists to enforce — the overwrite commits. `TrustedDefaultProfile` (the output slot) references the same opaque `OpaqueProfileLinkTarget` for consistency.

Reads are unaffected: the picker/wish resolve `defaultProfile` via `resolveAsCell()` / `profileLinkSchema()` (identity only), never the slot's declared schema. Profile fields stay owner-protected **in the profile's own space** by its own writers — walking them on the home-space default write was spurious.

### Tests (red-without / green-with, both proven)
- `packages/runner/test/profile-set-default-overwrite-cfc.test.ts` — deterministic **CFC-layer** RED/GREEN: overwriting a WALKABLE owner-protected single-link slot under the picker writer is rejected at `/defaultProfile/avatar` (RED = reproduced browser failure); an OPAQUE-link slot is accepted (GREEN = the fix).
- `packages/runner/test/profile-set-default-opaque-schema.test.ts` — compiles the **real `home.tsx`** and asserts its `defaultProfile` slot exposes no owner-protected `avatar`/`name`/`bio` to walk. RED without the fix, GREEN with it.

Regressions checked (all green): `profile-owner-cfc`, `profile-create-real-card-add`, `profile-home-bio`, `profile-home-pin-piece`, `wish`, `wish-scope`, `profile-space-identity-collision-*`, `cfc-boundary`, `cfc-write-floor`, `cfc-link-integrity-gate`, `cfc-required-integrity-provenance`, `cfc-ui-contract`, `cfc-nonexported-binding-identity`, plus `packages/patterns/integration/shared-profile`.

### Headless-vs-browser note
The real-patterns picker `.map`/`.key` handler binding collapses distinct rows to one shared argument-cell redirect when driven **headlessly**, so a full end-to-end overwrite silently no-ops rather than raising the browser's abort. The mechanism and fix are therefore proven deterministically at the CFC layer and via the real `home.tsx` schema guard above.

---
## Summary by cubic
Fixes CT-1845: changing an already-set default profile failed CFC at /avatar. We now make the `setDefaultProfile` write target and all binding sites opaque so overwrites succeed; reads are unchanged.

- **Bug Fixes**
  - Made `defaultProfile` opaque at every binding site to stop CFC from walking `/avatar/name/bio` on overwrite:
    - Durable cell in `packages/patterns/system/home.tsx`.
    - `setDefaultProfile` handler state param (`DefaultProfileCell`) and `profile` arg as `Cell<OpaqueProfileLinkTarget>` in `packages/patterns/system/profile-create.tsx`; `TrustedDefaultProfile` updated accordingly.
    - Picker input `defaultProfile` in `packages/patterns/system/profile-picker.tsx`.
  - Strengthened tests:
    - `packages/runner/test/profile-set-default-opaque-schema.test.ts` now also asserts the handler’s `$ctx.defaultProfile` write target is opaque.
    - Kept deterministic RED/GREEN CFC test `packages/runner/test/profile-set-default-overwrite-cfc.test.ts` verifying overwrite reject/accept behavior.

<sup>Written for commit 79d45e7c89f5297c23c931e69cc1f6f8ace4ee7b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4540?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/patterns/system/home.tsx
+++ b/packages/patterns/system/home.tsx
@@ -12,6 +12,7 @@ import FavoritesManager from "./favorites-manager.tsx";
 import Self from "../self.tsx";
 import {
   type CreateProfileEvent,
+  type DefaultProfileCell,
   submitProfileCreation,
   type TrustedDefaultProfile,
   type TrustedProfileList,
@@ -178,7 +179,14 @@ export default pattern<Record<string, never>, HomeOutput>((_) => {
   // is the one `#profile` resolves to in headless mode and orders first in the
   // picker; `mru` is the recency-ordered list driving the rest of the ordering.
   const profiles = new Writable<ProfileHomeOutput[]>([]).for("profiles");
-  const defaultProfile = new Writable<ProfileHomeOutput | undefined>(undefined)
+  // CT-1845: declare the durable cell with the OPAQUE `DefaultProfileCell` type,
+  // never `Writable<ProfileHomeOutput | undefined>`. The stored default link
+  // carries THIS declared schema, and CFC walks it on every write; a walkable
+  // `ProfileHomeOutput` here makes overwriting the default trip
+  // `writeAuthorizedBy failed at /avatar`. See profile-create.tsx.
+  const defaultProfile: DefaultProfileCell = new Writable<
+    Record<never, never> | undefined
+  >(undefined)
     .for("defaultProfile");
   const mru = new Writable<ProfileHomeOutput[]>([]).for("mru");
   // Untrusted-write regression surface: this stream is exported so tests can
--- a/packages/patterns/system/profile-create.tsx
+++ b/packages/patterns/system/profile-create.tsx
@@ -102,14 +102,26 @@ export const submitProfileCreation = handler<
 export const setDefaultProfile = handler<
   unknown,
   {
-    defaultProfile: Writable<ProfileHomeOutput | undefined>;
-    // Take the profile as a LINK cell, not a resolved value: the handler only
-    // needs the link to write into defaultProfile, and a link argument doesn't
-    // require the profile's cross-space values to be loaded at event time —
-    // resolving the full value here would fail required-field validation
-    // ("stream action argument is undefined … not running") whenever the
-    // profile's space hasn't materialized locally yet.
-    profile: Cell<ProfileHomeOutput>;
+    // CT-1845 — CFC derives the walked write-authorization schema from THIS
+    // handler-state declaration (the write TARGET the handler binds). Declaring
+    // it `Writable<ProfileHomeOutput | undefined>` bakes the walkable
+    // owner-protected `/avatar` (etc.) into the write path, so overwriting an
+    // already-set default trips `writeAuthorizedBy failed at /avatar` (the picker
+    // writer ≠ `setAvatar`). It MUST be the opaque `DefaultProfileCell` — every
+    // site whose type flows into this write (home durable cell, picker input,
+    // this handler param) must be consistently opaque or the walk fires again.
+    // See the `TrustedDefaultProfile` / `DefaultProfileCell` notes below.
+    defaultProfile: DefaultProfileCell;
+    // Take the profile as an OPAQUE LINK cell, not a resolved value and not the
+    // walkable `ProfileHomeOutput`: the handler only needs the link to write into
+    // defaultProfile, a link argument doesn't require the profile's cross-space
+    // values to be loaded at event time (resolving the full value here would fail
+    // required-field validation — "stream action argument is undefined … not
+    // running" — whenever the profile's space hasn't materialized locally yet),
+    // AND an opaque param keeps the serialized link's carried schema free of the
+    // walkable owner-protected sub-fields (CT-1845, same rationale as the write
+    // target above).
+    profile: Cell<OpaqueProfileLinkTarget>;
   }
 >((_, { defaultProfile, profile }) => {
   if (profile) {
@@ -197,9 +209,9 @@ type PickerProfileLink<
 //
 // CT-1845: the referenced link target is OPAQUE (`OpaqueProfileLinkTarget`, an
 // empty object), NOT the walkable `ProfileHomeOutput`, kept consistent with the
-// durable `DefaultProfileCell` declaration below — which is the actual runtime
-// lever (see its note). `defaultProfile` is a SINGLE owner-protected slot; the
-// schema stored on the slot is the write-policy schema CFC walks on every write.
+// `DefaultProfileCell` sites below (the real runtime lever is the handler's own
+// write target — see its note). `defaultProfile` is a SINGLE owner-protected
+// slot; the write-authorization schema CFC walks derives from the binding site.
 // With a walkable `ProfileHomeOutput`, CFC
 // (`walkIfcSchema`) emits owner-protected entries for the target's OWN fields —
 // `/name`, `/avatar`, `/bio`, each `writeAuthorizedBy: set…`. OVERWRITING the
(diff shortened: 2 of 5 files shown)


bfollington (commented):
_cf-review via Claude (claude-opus-4-8), on behalf of @bfollington. Self-signed, automated review._

**Scope** · `packages/patterns/system/{home,profile-create}.tsx` + 2 new runner tests · **Coverage**: deep-read (prepare.ts CFC path, both source files, both tests) · **Checks run**: both new tests (green w/ fix), schema-guard reverted-to-main (RED confirmed), `profile-owner-cfc` + `cfc-boundary` regressions (green), lint + fmt (clean), plus schema-compilation probes of the real `home.tsx` result schema on main vs PR.

## Verdict: **non-blocking / ship-able.** Root cause is correct, and the fix does **not** weaken any protection that was actually being enforced. One clarification the PR narrative gets slightly wrong (harmless), and one latent pre-existing gap worth a follow-up.

---

### (a) Root cause — CORRECT (verified)

Confirmed against `prepare.ts`:
- `verifyInputRequirements` (2588) iterates `walkIfcSchema(schema)` and runs `writeAuthorizedByReason` for every entry `ifcEntryAppliesToAttemptedWrite` accepts.
- `walkIfcSchema` (1600) descends `properties` and resolves `$ref`, so a walkable `ProfileHomeOutput` on the slot emits `/name`, `/avatar`, `/bio` (and `/elements`) entries. I compiled the real `ProfileHomeOutput` def and confirmed each carries `writeAuthorizedBy → profile-home.setAvatar/setName/setBio/mutateElements` with `ownerPrincipal: currentPrincipal`.
- The overwrite asymmetry is real: in the non-wildcard branch (2352-2407) a whole-slot overwrite marks nested `/avatar` "touched" via `concretePathHasPrefix`, then reads the **resolved** concrete value → entry applies → `setAvatar` enforced against the picker writer → reject. First-write-from-empty: `value===undefined && previousWriteValueForTarget===undefined` → returns false → does not apply → passes. Overwrite-specific, exactly as described.

### (b) Does the fix weaken real protection? — **NO. This is the key finding.**

I compiled the result schema both ways. Two facts settle it:

1. **The walked `/name /avatar /bio /elements` entries belong to the *linked profile's own* schema**, authorized by the profile's own `setName/setAvatar/setBio/mutateElements` in `profile-home.tsx` with `ownerPrincipal: currentPrincipal`. Those writers still protect those fields **in the profile's own space**. Walking them on the *home-space default-slot write* was spurious — a home write can never be `setAvatar`. Removing that walk removes only the spurious enforcement; nothing that legitimately guards a profile field is lost.

2. **The slot itself never carried a compiled slot-level `writeAuthorizedBy`, on main OR on this PR.** Because `TrustedDefaultProfile = PickerProfileLink<…> | undefined`, the transformer emits `anyOf: [{undefined}, {$ref}]` and the `Cfc<WriteAuthorizedBy<…>>` wrapper's `ifc` is **dropped** off the union — verified: on main it compiles to `anyOf:[{undefined},{$ref: ProfileHomeOutput}]` (no top-level ifc), and on the PR to `anyOf:[{undefined},{$ref: OpaqueProfileLinkTarget}]` (`{type:object,properties:{}}`, no ifc). So there was no slot-level `setDefaultProfile` gate in the compiled schema to lose. By contrast `profiles`/`mru` (non-optional, no `| undefined`) **do** keep their container `writeAuthorizedBy` — confirmed.

Net: the fix is genuinely spurious-walk removal. No hole is opened by it. An unauthorized writer setting `defaultProfile` was already ungated at the slot level (pre-existing — see follow-up), and this PR neither improves nor regresses that.

### (c) Read side — UNAFFECTED (verified)

Stored value is a pure link sigil; picker reads via `.asSchema(profileLinkSchema()).get()` (identity `asCell` only), never the slot's declared schema, so opaque doesn't change resolution or typing. CT-1842/#4534/#4536 are **not on `origin/main`**; this PR touches only these two type decls, and since reads are identity-only there is no interaction or merge-order hazard with space-based identity resolution. (Minor: `profile-picker.tsx:42` / `profile-create.tsx:105` still annotate the **input** `defaultProfile` as `Writable<ProfileHomeOutput|undefined>`, bridged by `as any` at the home call site — consistent, reads identity-only, fine.)

### (d) Test quality — one honest caveat

- `profile-set-default-opaque-schema.test.ts` is the **real** regression guard: I reverted the two source files to main and it goes **RED** (`avatar/name/bio/elements` walked) — GREEN with the fix. It compiles the shipped `home.tsx` and its `$ref`-aware walker matches CFC's own resolution. Solid.
- `profile-set-default-overwrite-cfc.test.ts` demonstrates the CFC mechanism deterministically and both arms pass — but note its RED arm builds its **own** `walkableDefaultProfileSlot` schema, so it is **not gated on the fix** (it passes on main too). That's fine as a mechanism/documentation test, but it is not a regression guard; the schema-guard test is what actually pins the fix. Worth stating plainly rather than implying both are fix-gated.
- The acknowledged e2e gap (headless picker `.map`/`.key` redirect-collapse) is real; the schema-guard-on-real-`home.tsx` + CFC-mechanism combo is an adequate substitute here since the schema guard exercises the actual shipped artifact.

### (e) Sibling slots — MRU/profiles safe (verified)

`mru`/`profiles` elements keep the walkable `ProfileHomeOutput` (`PickerProfileLink` default `LinkTarget`), but their owner-protected sub-fields live under `items` → path `*/avatar`. The wildcard branch (2408+) only fires on **genuinely changed** elements (`!deepEqual(value, previousValue)` / `changedValuesAtPatternPath`), so prepend/append never re-touches an existing element's resolved `/avatar`. Only a **single whole-container link slot** hits the non-wildcard "touched prior resolved value" path. `defaultProfile` was the only such walkable optional single-link slot; no other home slot matches the shape. Author's MRU-escape explanation is accurate.

### (f) Idiom — acceptable, no blessed alternative

`Record<never, never>` for an opaque "identity-only link target" is not duplicating any canonical machinery (it's a type, not hashing/serialize/clone). There is no pre-existing blessed opaque-link-target type in the profile stack; `OpaqueProfileLinkTarget` compiles to `{type:object,properties:{}}`, exactly the "no sub-fields to walk" shape intended. Fine.

---

1. **CFC-layer test is not fix-gated.** `profile-set-default-overwrite-cfc.test.ts`'s RED arm asserts a self-constructed walkable schema fails — it passes on `main` too. Either reframe its docstring to "mechanism demonstration" (it is a good one) or make it consume the compiled `home.tsx` slot so it becomes a real guard. The `opaque-schema` test carries the actual regression weight.

## Possible follow-ups (proposals, for a human to file)

1. **Latent, pre-existing: the `defaultProfile` slot has no slot-level CFC write gate.** The `T | undefined` union drops the `Cfc<WriteAuthorizedBy<…>>` wrapper's `ifc` off the compiled schema, so `setDefaultProfile` is not enforced at the slot on main or this PR. Not introduced here and out of scope, but if slot-level authorization on optional single-link slots is intended, the "`anyOf:[undefined,…]` erases wrapper `ifc`" behavior deserves a transformer/CFC investigation — does any *other* optional owner-protected scalar/link slot silently lose its gate the same way?

bfollington:
**Parked (deferred) as of 2026-07-06.** The v2 fix (opaque handler write-target param + picker input, in addition to v1's opaque durable slot/output type) was **also disproven by in-browser overwrite verification** — same `writeAuthorizedBy failed at /avatar`, `cf inspect` shows zero durable writes. Three profile-side typing attempts have now failed in-browser.

Working hypothesis (not yet instrumented-confirmed): since every declaration/binding site is now opaque and the `/avatar` walk STILL fires only on **overwrite**, the walked schema is almost certainly coming from CFC **resolving the OLD stored default link** during the overwrite diff (a real profile link → `ProfileHomeOutput` → owner-protected `/avatar`), independent of the new write's type. That points at CFC's replace-a-link semantics, not profile-side typing.

Deferred because it's **off the critical path**: "Set default" is a multi-profile operation, and Loom v1 is single-default-profile (no picker, no Set default). Left as draft pending either an instrumented `prepare.ts` root-cause confirmation + CFC-epic handoff, or reprioritization. Full analysis in CT-1845.

bfollington:
These fixes have been disproven by our actual runtime evaluation, and I don't know if they do anything meaningful or not as a result. This has been parked for long enough that I'm gonna close it and reopen if we find the issue again.

PR #4542, opened 2026-07-06
perf(js-compiler): interleave the event loop through cold pattern compiles

seefeldb:
## Problem

A cold pattern compile in the browser runtime worker runs as **one synchronous stretch** that wedges the worker's event loop for its whole duration. Every `await` on the `compilePattern` → `compileToRecordGraph` → SES-verify → SES-evaluate path resolves an already-settled promise, so it yields only *microtasks* — queued **macrotasks** (worker IPC message events: `cell:subscribe`/`set`/`get` deliveries) never run until the compile finishes.

Measured on the lunch-poll-vote boot where both browsers cold-compile `system/default-app.tsx` in their workers:

- `runner.loop/workerLag` max **1638–1640 ms** locally (5.7–6.0 s on 2-core CI)
- during that window, `runner.ipc/delivery/*` max **== workerLag max** — unrelated cell traffic stalled for exactly the compile duration
- under enough load the wedge starved the first fill's commit ack (the "second-boot slow window")

**Any first boot of a fresh space in production pays this too** — the space-root pattern compile freezes all cell traffic for the whole compile. This is complementary to the recent cold-load bucket work (#4441/#4442/#4455/#4459/#4460), which optimizes the *warm/resume* path and worker boot; the *cold* transform/emit/SES stretch is what this PR addresses.

## Fix — yield real macrotask turns at module boundaries

- **`utils/sleep.ts`** — `yieldToEventLoop()`: a `MessageChannel` post (a macrotask that runs *behind* already-queued messages, with no nested-`setTimeout` clamp) plus a time-budgeted `setTimeout(0)` hop every 8 ms. The hybrid is necessary: a pure message-yield chain interleaves messages perfectly but **starves timers** in Deno (measured — an armed interval never fires); the periodic timer hop is scheduled *from the message callback* so its nesting level stays 1 and the browsers' nested-timeout clamp never engages.

- **`js-compiler/compiler.ts`** — `compileToModules` is now a thin sync driver over a new `compileToModulesSteps` **generator** that steps per module (program build, then each file's type-check / declaration-check / transform+emit). Per-file `tsProgram.emit(sourceFile, …)` is output-identical to the whole-program `emit(undefined)`: same program order, same single transformer-pipeline instance, diagnostics aggregated across calls. A new async `compileToModulesInterleaved` drains the same generator with one macrotask yield per step. `Checker` gains per-file collectors (`collectSemanticErrors` / `collectDeclarationErrors` + `throwIfErrors`); `typeCheck` / `declarationCheck` are rebuilt on them. Per-step `js-compiler/phase/*` timing spans record even while the logger is disabled.

- **`runner/harness/engine.ts`** — the cold-compile path uses `compileToModulesInterleaved` and yields between per-module SES body verifications; `evaluateCachedModules` (warm resume) yields between verifies and once before its synchronous SES evaluation.

- **`runner/pattern-manager.ts`** — one yield ahead of each synchronous `evaluateRecordGraph` stretch, on both cold and warm-resume paths.

The CLI's synchronous `compileToModules` behavior is byte-for-byte unchanged (both drivers drain the same generator; equivalence pinned by test).

## Results

On the pre-fix lunch-poll-vote boot (both browsers cold-compiling `default-app`):

| metric | before | after |
| --- | --- | --- |
| `runner.loop/workerLag` max | 1638 / 1640 ms | **424–542 ms** (−70%) |
| `delivery/cell:*` max (overlapping) | up to compile length | ≤ 78 ms |
| `compilePattern` wall | 1.9–2.1 s | 2.2–2.6 s (+10–25%, IPC now interleaves *inside* the window — intended) |

The residual workerLag is bounded by the **single biggest per-module step** — the entry file's type-driven transform+emit (`js-compiler/phase/emitFile` max ~525 ms; next chunk `typeCheckFile` ~250 ms). Going below the per-file floor needs a compile sub-worker or server-side pre-seeded compile caches for system patterns; left as a follow-up.

In-process A/B on the merged tree (`packages/runner/test/manual-compile-wedge.ts`, engine temporarily flipped to the sync driver): sync = one **~1479 ms** stretch the loop can't run; interleaved = longest stretch **371 ms**, same ~1.4 s total compile work.

## Instrumentation included

The measurement half of the investigation, ported from the `sched-v2-caughtup` branch so the numbers are reproducible:

- request-timeline ledger + `sentEpochMs` → `runner.ipc/delivery|handle/<type>` worker decomposition, main/worker event-loop lag probes (`runner.loop/workerLag`, `runtime-client loop/mainLag`)
- piece / runner resume-pipeline phase spans, and `js-compiler/phase/*` per-file spans, surfaced in the integration-test load summaries
- `manual-compile-wedge.ts` — standalone in-process repro (no browsers): `deno run -A packages/runner/test/manual-compile-wedge.ts`

The lunch-poll-vote test keeps its cold-compile boot (no `ensureDefaultPattern` pre-create) so the wedge stays covered.

## Testing

- `js-compiler`: 6 passed — includes new sync/interleaved driver-equivalence tests (byte-identical modules + source maps, identical type-error surfacing)
- Full `runner` suite: **771 passed, 0 failed**
- `lunch-poll-vote` browser integration: passing, workerLag max 479/542 ms
- `piece` / `runtime-client` / `utils` suites green
- Caught up to latest `main` (through #4525); the compile/worker path is untouched by main, so the wedge and the fix are unchanged.

---
## Summary by cubic
Interleaves real event-loop turns during cold pattern compiles in the browser worker to prevent IPC stalls, cutting worst‑case lag by ~70% with byte‑identical output. Interleaving is browser‑only; Deno paths use the sync driver to keep tests/CLI fast.

- **New Features**
  - `utils`: `yieldToEventLoop()` (MessageChannel + time‑budgeted `setTimeout`) and `unrefTimer()` to detach long‑lived intervals under Deno.
  - `js-compiler`: `compileToModulesInterleaved()` over a shared `compileToModulesSteps` generator; tests pin identical modules, source maps, and error surfacing vs sync.
  - `runtime-client`/`web-worker`: boot‑window request timeline and pending‑request snapshot; worker request ledger; main/worker loop‑lag probes (unref’d); count replies only after `postMessage` succeeds; always record `handle/<type>` timings, even on errors.

- **Refactors**
  - `engine`/`pattern-manager`: gate all compile/verify/evaluate yields behind `COMPILE_INTERLEAVES_EVENT_LOOP` (browser‑only); Deno runs the synchronous driver; yield before SES verifies and before synchronous `evaluateRecordGraph` on the browser path.
  - `js-compiler`: per‑file type/declaration checks and per‑file emit via a generator driver; `Checker` gains per‑file collectors; compile‑phase spans record into timing stats while logging is disabled.
  - Integration helpers/tests: page keeps a bounded console tail; input helpers track per‑step fill progress; added in‑process wedge probe (`packages/runner/test/manual-compile-wedge.ts`); summaries include `js-compiler`/`piece`/resume timing rows.
  - `runner`: consolidated yield gates into `interleaveCompileYield()`; tests pin Deno’s “no macrotask” contract and the browser’s “real turn” behavior; added a byte‑cache reuse test for warm compiles.

<sup>Written for commit fb8de066b06171427b5d4856f1cb05df0dec10a7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4542?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seefeldb:
Thanks — addressed all four in `307ec77c8`. They're all in the ported boot-window instrumentation (not the compile-interleave fix); each is valid.

**Codex P1 ×2 — leaked lag-probe intervals** (`web-worker/index.ts` top-level `setInterval`, `connection.ts` constructor `setInterval`). Both now go through a new `unrefTimer` helper in `@commonfabric/utils/sleep` (same idiom as `traverse-recorder.ts`/`cell-bridge.ts`), so importing the worker entry or constructing a `RuntimeConnection` without a matching `dispose()` no longer leaves a live interval. `unrefTimer` is a no-op in the browser, where the timers run for the real worker/connection lifetime and are cleared on dispose as before.

One clarification for the record: I reproduced both cited tests (`web-worker-console-bridge.test.ts`, `shell/test/error-ipc.test.ts`) under `deno test --trace-leaks` on this toolchain and neither actually fails today — this Deno version doesn't attribute a module-/constructor-scope timer to the test's op-sanitizer window. So this is hardening against a real latent leak (and stricter sanitizer configs), not a fix for a currently-red test.

**cubic P2 — reply counted before it's posted** (`index.ts:179` and the sibling ack paths). The `responded/<type>` ledger increment now runs *after* `self.postMessage(...)`, so a failed post (non-cloneable payload, closed channel) records only `responded-error/<type>` and never double-counts one request as both.

**cubic P3 — `handle` timing drops failed requests** (`index.ts:208`). `runner.ipc/handle/<type>` is now recorded in a `finally` around `worker.handleRequest(request)`, so a throwing request is still timed and error-heavy periods don't underreport handling latency.

Verified: `runtime-client` 29/0, `utils` 86/0; `error-ipc` + `web-worker-console-bridge` green under `--trace-leaks`.

seefeldb:
Added test coverage for the perf-check coverage-debt (`69cd26259`, tests only — no source changes). Independently verified by intersecting the PR's added lines with `deno coverage --lcov` uncovered lines: **every executable line this PR added is now covered** in all three flagged packages.

| file | before → after (line) |
| --- | --- |
| `utils/src/sleep.ts` | no test at all → **100%** |
| `js-compiler/…/diagnostics/checker.ts` | 60% → **100%** |
| `js-compiler/typescript/compiler.ts` | PR-added lines: all covered (interleaved multi-file error aggregation + declaration-diagnostic path) |
| `runtime-client/backends/web-worker/index.ts` | 67% → **100%** (delivery/handle timing incl. the throwing-request `finally`, and the reordered `responded/*` ledger) |
| `runtime-client/client/connection.ts` + `runtime-client.ts` | PR-added lines: all covered (request-timeline ledger, `getRequestTimeline` copy semantics) |

The worker-entry tests also pin the two review fixes from `307ec77c8`: `handle` timing is recorded on a throwing request, and a failed `self.postMessage` counts as `responded-error/*` (never both). Assertions are structure/counter deltas only — no wall-clock magnitudes. All three suites pass (`utils` 90, `js-compiler` 7, `runtime-client` 34), `--trace-leaks` clean on the timer-touching packages, fmt/lint clean.

One finding worth flagging: after the compile-driver refactor, `Checker.typeCheck()` / `declarationCheck()` have **no remaining production callers** — the pipeline now steps through `checkableSources()` + the per-file `collect*Errors` collectors. I left them in place (pinned by direct unit tests) as the documented whole-program one-call contract rather than deleting them in this PR; happy to remove them in a follow-up if you'd prefer.

seefeldb:
Performance Check is green. For the record on how it resolved:

- The initial red was **not a code regression**. It was a cold-compile-cache artifact: this PR edits `packages/js-compiler/**` and `packages/runner/src/harness/**`, which are part of the pattern-unit-test job's `actions/cache` key, so the branch ran the compile byte cache **cold** while main's baseline ran warm — inflating every pattern-unit compile ~2x. Verified by local A/B: cold-vs-cold this branch vs main is within ±3% across contact-book / calendar / gmail-importer / do-list (no per-compile regression), while cold-vs-warm on the same pattern is ~1.9x (matching the CI delta). Re-running the workflow warmed the cache and the timings matched main.
- One real sub-issue was fixed in `fb8de066b`: the browser-only interleave gate added a few `if (COMPILE_INTERLEAVES_EVENT_LOOP) …` lines that never execute in Deno, which LCOV counted as +6 uncovered runner lines (an OVER coverage-debt metric). Those seven gate sites are now folded into a single tested `interleaveCompileYield()` helper.

Net: browser worker interleaves the event loop during cold compiles (workerLag ~1.6s → ~0.4s); Deno batch compiles (`cf test`, server, CLI) run the synchronous driver unchanged.

PR #4553, opened 2026-07-06
feat: computed-cell identity — kind-tagged ids and ack-and-drop conflict semantics

ubik2:
## Summary

Implements phases 1–3 of `docs/specs/computed-cell-identity.md` (spec included in this PR): internal cells that are pure functions of their pattern's inputs get identity-level classification, and the memory server relaxes conflict semantics for their writes — eliminating reject/re-run/recommit churn when multiple clients recompute the same derived values.

Everything is gated behind `EXPERIMENTAL_COMPUTED_CELL_IDS` (minting side); the server policy and readers activate only on ids that carry the new format.

### Kind-tagged entity ids (minting)

- New versioned id format `fid2:computed:<hash>`; `fid1:<hash>` stays the untagged, strict-semantics form. `FabricHash.fromString` now splits at the **last** colon (the hash segment is base64url and never contains one), so existing ids parse identically and fid1-only parsers fail loudly on kinded ids instead of silently mis-handling them.
- `createRef` mints the kind into BOTH the hash preimage (shape-disjoint envelope — kinded and untagged ids can never collide bytes-for-bytes) and the visible tag, from one argument, so the representations cannot diverge and a kind flip necessarily names a different entity.
- Conservative builder classifier (`assignComputedCellKinds`): a cell qualifies only when every writer is a plain javascript compute module with no writable-input markers, it is never captured in a non-qualifying node's inputs, is not a stream, and is not exposed via a write-capable handle in the result surface. Kind participates in descriptor/manifest matching, so a kind flip re-materializes under the new id.

### Ack-and-drop conflict policy (server + client)

- A commit whose semantic operations ALL target computed-kind entities is **acknowledged-as-committed but dropped** when its reads are stale, instead of rejected: recompute from the newer inputs supersedes the write, so rejection would only buy retry churn. The dropped commit inserts a zero-revision commit row, which keeps replay dedupe, dependent pending reads, and `origin-committed` preconditions intact for free.
- The storage client reverts the optimistic pending value on a `droppedComputed` ack rather than promoting it (promotion would pin the stale value behind the monotonic seq guard in `applySessionSync`, shadowing the authoritative value indefinitely). No retry is scheduled — the newer inputs arrive through the normal watch flow, dirty the reader, and the recompute writes a fresh commit.
- Mixed, sqlite-bearing, unknown-kind, and untagged commits keep strict all-or-nothing conflict semantics.

### Verified capture writes (classifier widening)

- `send` added to the capability analysis writer methods (`Stream.send()` delegates to `set()` — it is a write).
- The transformer emits `captureWritesAnalyzed` on lowered computeds whose analysis saw every capture: no wildcard escape, no recursion short-circuit, and no unrecognized method call on a cell-like receiver (new `hasUnverifiedCellUse` — the writer-method list is closed but the Cell API is not, so unknown cell-methods fail closed).
- For verified modules `materializerWriteInputPaths` is an exhaustive record of capture writes, so the classifier accepts handle-bearing argument schemas when it is empty: a computed embedding event streams in its JSX output (the canonical UI derivation, e.g. habit-tracker's `habitCards`) now classifies as pure. Hand-written lifts carry no provenance and stay conservative.

### Measured impact

Census over 18 real patterns (exemplars + system/home + default-app), 483 internal cells: **109 computed / 243 standard / 131 streams — 31.0% of data cells classify computed** (27.6% before the widening). Computed cells are the high-churn ones (rewritten on every input change), so their share of commit traffic — and of multi-client conflict churn — is substantially higher than their share of cells.

## Test plan

- `packages/data-model`: entity-kind round-trips (string/URI/codec), mint-once/fid1-only guards, unknown kinds read as strict
- `packages/runner`: classifier coverage per disqualification rule + the verified-capture relaxation; flag-off byte-identical output; kind-tagged link minting; storage-client ack-and-drop (real in-process server): stale computed commit acked → optimistic value reverted → recompute converges; strict-semantics control for untagged cells
- `packages/memory`: drop + marker, replay idempotency, origin-committed satisfaction, pending-read resolution against a dropped commit, current-reads apply, mixed/untagged/unknown-kind stay strict, head advancement
- `packages/ts-transformers`: full suite (999) including regenerated goldens — audited as marker insertions plus balanced schema-property moves only
- Full `deno task check` and package suites green; census script used for the impact numbers

Remaining (not in this PR): value-equality dedupe for current-read computed commits (phase 4), server-side action runner (phase 5), handler write facts for further classifier widening.

---
## Summary by cubic
Introduce kind‑tagged computed cell ids and an “ack‑and‑drop” server policy for stale writes to computed cells, behind `EXPERIMENTAL_COMPUTED_CELL_IDS`. This reduces multi‑client conflict churn while keeping commit bookkeeping intact.

- **New Features**
  - Mint `fid2:computed:<hash>` for internal cells proven to be pure; `fid1:<hash>` stays strict.
  - `createRef` supports a `kind`; kind is embedded in the hash preimage and visible tag; manifest matching now includes kind. `FabricHash.fromString` splits at the last colon so existing ids parse unchanged.
  - Conservative computed classifier added, widened by verified capture‑writes (`captureWritesAnalyzed`) and treating `.send()` as a write; embedding event streams in JSX no longer disqualifies pure computeds.
  - Memory‑v2 acknowledges and drops stale commits whose ops all target computed‑kind ids; writes a zero‑revision commit row for dedupe and preconditions. Client reverts the optimistic value on `droppedComputed`; mixed/unknown/sqlite/untagged commits keep strict reject‑on‑stale.
  - New flag `EXPERIMENTAL_COMPUTED_CELL_IDS` wired through runtime, CLI, shell, toolshed, and background service; spec and docs updated.

- **Migration**
  - Opt‑in by setting `EXPERIMENTAL_COMPUTED_CELL_IDS=true` (or `experimental.computedCellIds` in `Runtime` options). Readers accept both `fid1` and `fid2` forms.
  - Enabling can re‑materialize internal cells due to kind‑sensitive ids; this is expected.
  - If you parse Fabric ids outside the SDK, split on the last colon and accept `fid2:computed:`.

<sup>Written for commit e634ddfe7dc021acca84eb757d0c4e09864da401. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4553?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/development/EXPERIMENTAL_OPTIONS.md
+++ b/docs/development/EXPERIMENTAL_OPTIONS.md
@@ -9,6 +9,7 @@ various major features.
 |------|---------|-------------|
 | `modernCellRep` | `EXPERIMENTAL_MODERN_CELL_REP` | Enables new "cell representation" classes |
 | `persistentSchedulerState` | `EXPERIMENTAL_PERSISTENT_SCHEDULER_STATE` | Enables durable scheduler observations, dirty state, and scheduler rehydration through memory-v2. |
+| `computedCellIds` | `EXPERIMENTAL_COMPUTED_CELL_IDS` | Mints kind-tagged entity ids (`fid2:computed:`) for internal cells provably written only by compute nodes. Gates minting only; readers accept both forms. See `docs/specs/computed-cell-identity.md`. |
 | `schedulerHistoricalMightWrite` | n/a (`RuntimeOptions` only) | Preserves the scheduler's legacy cumulative write history for dependency scheduling instead of the default current-known write set. |
 
 All flags default to `undefined` which means they take on the default value
--- a/docs/development/handlers/invocation-outside-pattern.md
+++ b/docs/development/handlers/invocation-outside-pattern.md
@@ -43,7 +43,7 @@ await runtime.idle();
 
 ## Why it works
 
-The `{ asStream: true }` in the schema tells the cell that this property is a stream. When you call `.send()` on a stream-marked cell, it internally uses `scheduler.queueEvent()` which triggers the registered handler.
+The `{ asCell: ["stream"] }` in the schema tells the cell that this property is a stream. When you call `.send()` on a stream-marked cell, it internally uses `scheduler.queueEvent()` which triggers the registered handler.
 
 ## What doesn't work
 
--- a/docs/specs/computed-cell-identity.md
+++ b/docs/specs/computed-cell-identity.md
@@ -2,12 +2,19 @@
 
 ## Status
 
-Phase 1 (minting) implemented behind `EXPERIMENTAL_COMPUTED_CELL_IDS`:
-`createRef`/`FabricHash` kind support (preimage + `fid2:computed:` format),
-conservative builder classifier, kind-aware descriptor/manifest matching.
-Server-side conflict policy (phases 2-4) not started. Derived from a design
-discussion on 2026-06-30/07-01 about avoiding needless commit conflicts when
-multiple clients recompute the same derived values.
+Phases 1-3 implemented. Phase 1 (minting), behind
+`EXPERIMENTAL_COMPUTED_CELL_IDS`: `createRef`/`FabricHash` kind support
+(preimage + `fid2:computed:` format), conservative builder classifier,
+kind-aware descriptor/manifest matching. Phase 2 (server policy): the
+memory-v2 engine acknowledges-and-drops stale all-computed commits — a
+zero-revision commit row keeps replay dedupe, dependent pending reads, and
+origin-committed preconditions working — and the storage client reverts the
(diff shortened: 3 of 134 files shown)


ubik2:
Closing — splitting this up: the capability-analysis changes (send-as-write + fail-closed unknown-cell-method tracking) will come as a standalone PR for transformer review; the computed-cell identity work continues on this branch.

PR #4554, opened 2026-07-06
feat(ts-transformers): classify send() as a write; fail closed on unknown cell-method calls

ubik2:
## Summary

Hardenings to the capability analysis's write tracking, split out of the computed-cell identity work so the transformer system can review them standalone. All are self-contained analysis improvements; a downstream consumer that relies on write-exhaustiveness lands separately.

### `send` is a write

`send()` delegates to `set()` on every receiver (`packages/runner/src/cell.ts`), and `set()` forks on the receiver's kind: a raw cell takes an ordinary transactional write, while a stream enqueues the event without revising the stream cell — the dispatched handler's writes commit in their own transaction, effects the analysis cannot see or bound. Previously the analysis routed `send` through the unknown-method fallback, recording a **read**, so a callback that fires events during evaluation summarized as write-free. `send` now joins `WRITER_METHODS` (alongside the catalog-driven mergeable scalar writers).

### Fail closed on unknown cell-method calls

The writer-method sets are a closed list, while the Cell API is not — any mutator missing today or added tomorrow silently produces incomplete `writePaths` with no signal. New `hasUnverifiedCellUse` on `CapabilityParamSummary`: an unrecognized method call on a **cell-like receiver** (type-checked via the existing `isCellLikeExpression`; array/string methods on `.get()` snapshots are unaffected) marks the parameter, and the mark propagates through interprocedural summaries the same way `wildcard` does. Consumers that need `writePaths` to be a closed-world record can now fail closed instead of trusting a possibly-stale method list.

### `onCommit` callbacks poison write-exhaustiveness

`set(value, onCommit?)` / `send(event, onCommit?)` with a second argument also mark `hasUnverifiedCellUse`: the `onCommit` callback receives the committed **transaction** and can write arbitrary cells through it — writes invisible to capture-path tracking even when the callback body itself is analyzed (it writes via `tx`, not via captures). The write to the receiver is still recorded; only exhaustiveness fails closed.

Note on the `onCommit` contract: its doc (cell.ts) says the callback "runs after the final commit result, including failure, so it must remain non-effectful," directing external effects to the post-commit outbox. The analysis deliberately does **not** rely on that contract: it is advisory — the passed `tx` is complete by callback time, but a closure can still reach captured cells or `cell.runtime.edit()`, and nothing at the type or runtime level traps a violation. Inline callback bodies are walked by the nested analysis, so capture-mediated writes inside them already appear in `writePaths`; the mark covers exactly the contract-violating residue (fresh-transaction escapes, external I/O) that the analysis cannot bound. If the contract is ever runtime-enforced, this poisoning rule can be narrowed accordingly.

Deliberately inert otherwise: the mark does not affect shrinking, identity classification, or schema generation — no behavior change for any existing consumer.

## Runtime-visible behavior change (intended)

With `send` classified as a write, computeds that send through captured streams now emit `materializerWriteInputPaths` where they previously emitted no scheduler options. Two consequences, both deliberate and pinned by a regression test ("send-only computed keeps stream paths in write metadata but never brands them collectible"):

- Stream paths in the field are never collectible: the runner's envelope collector filters to `cell`/`writeonly` brands, so a stream send produces no materializer envelope (a stream send writes no address — the dispatched handler's writes belong to the handler's own action). A `send` on a raw writable cell IS collectible, which is correct: that is a transactional write to the cell's address.
- The field's presence switches the runner off the opaque-result fallback. For modules carrying analyzed write metadata this is a precision improvement, not a regression: writable args that were never written would appear in the write paths if written, so trusting the (possibly empty) field beats the fallback's over-collection. The fallback remains for artifacts without metadata.

## Golden churn

Five generated handler ctx schemas reorder properties (send-classified captures moved capability buckets, changing emission order); the diff is balanced — 5 `asCell: ["stream"]` lines removed and re-added identically, no kind changes, `required` arrays reordered accordingly. JSON Schema property order is semantically meaningless.

## Tests

- `send()` through a capture records a write path and classifies `writeonly`
- unknown method on a cell-typed receiver sets `hasUnverifiedCellUse` (checker-based)
- unknown method without a checker fails closed (conservative)
- value-level unknown methods (`.some()` on a `.get()` snapshot) stay verified
- `set()`/`send()` with an `onCommit` callback mark unverified; plain calls stay verified
- full ts-transformers suite: 1006 passed, goldens regenerated

## Coverage-debt note

The coverage gate flagged +71 lines (ts-transformers) / +1 (runner). Measured root cause: none of this PR's added lines are uncovered (verified locally — 0 uncovered changed lines under the package suite). The delta is pre-existing lines in files this PR does not touch (`type-shrinking.ts` −21, `expression-rewrite/emitters/call-expression.ts` −17, `expression-site-lowering.ts` −7, `schema-injection.ts` −6, …) that only execute during cold pattern compilation: the latest-main baseline run (the js-compiler perf merge) was compile-cache-cold and ran the transformer in every pattern job; this PR's warm runs don't. Same cache-state asymmetry as the earlier timing false-positive, mirrored. Overrides below accept the warm-run debt values.

NEW_PERF_BASELINE: coverage-debt: packages/ts-transformers uncovered lines = 1998 lines
NEW_PERF_BASELINE: coverage-debt: packages/runner uncovered lines = 7646 lines

### DIFF
--- a/packages/ts-transformers/src/policy/capability-analysis.ts
+++ b/packages/ts-transformers/src/policy/capability-analysis.ts
@@ -76,6 +76,14 @@ interface MutableCapabilityState {
   hasIdentityUse: boolean;
   hasNonIdentityUse: boolean;
   hasNonIdentityRootUse: boolean;
+  /**
+   * An unrecognized method was called on a cell-like receiver rooted in this
+   * parameter. The call could be a mutator this analysis does not know, so
+   * `writes` cannot be treated as exhaustive — consumers asserting write
+   * exhaustiveness must fail closed on this, like `wildcard`; recognized
+   * reads/derivations are unaffected.
+   */
+  hasUnverifiedCellUse: boolean;
 }
 
 interface ObservedCapabilityUsage {
@@ -85,6 +93,7 @@ interface ObservedCapabilityUsage {
   readonly opaquePaths: readonly (readonly string[])[];
   readonly passthrough: boolean;
   readonly wildcard: boolean;
+  readonly hasUnverifiedCellUse: boolean;
   readonly identityOnly: boolean;
   readonly identityPaths: readonly (readonly string[])[];
   readonly identityCellPaths: readonly (readonly string[])[];
@@ -147,9 +156,14 @@ const PARAMETER_SUMMARY_PREFIX = "__param";
 const mergeableMethods = (kind: MergeableOpMethodKind): string[] =>
   MERGEABLE_OP_METHODS.filter((op) => op.kind === kind).map((op) => op.method);
 
+// `send` is a write: at runtime Stream.send() delegates to set() (an event
+// enqueue is a write to the stream cell). Classifying it here keeps
+// writePaths an honest record of capture writes — a callback that fires
+// events during evaluation must never summarize as write-free.
 const WRITER_METHODS = new Set([
   "set",
   "update",
+  "send",
   ...mergeableMethods("scalar-writer"),
 ]);
 const ARRAY_IDENTITY_WRITER_METHODS = new Set([
@@ -1325,6 +1339,7 @@ function normalizeObservedCapabilityUsage(
     opaquePaths,
     passthrough: state.passthrough,
     wildcard: state.wildcard,
+    hasUnverifiedCellUse: state.hasUnverifiedCellUse,
     identityOnly: identityPaths.some((path) => path.length === 0) &&
       !state.hasNonIdentityUse &&
       state.reads.size === 0 &&
@@ -1430,6 +1445,7 @@ export function analyzeFunctionCapabilities(
           hasIdentityUse: false,
           hasNonIdentityUse: false,
           hasNonIdentityRootUse: false,
+          hasUnverifiedCellUse: false,
         };
         states.set(name, state);
       }
@@ -1469,6 +1485,15 @@ export function analyzeFunctionCapabilities(
       const state = ensureState(name);
       state.wildcard = true;
       state.hasNonIdentityUse = true;
+    };
+
+    // Unlike markWildcard this does NOT change shrinking or identity
+    // classification — it only poisons write-exhaustiveness for consumers
+    // that need `writes` to be a closed-world record, while everything else
+    // behaves as before.
+    const markUnverifiedCellUse = (name: string): void => {
+      const state = ensureState(name);
+      state.hasUnverifiedCellUse = true;
     };
 
     const markPassthrough = (
@@ -2814,6 +2839,9 @@ export function analyzeFunctionCapabilities(
               markWildcard(source.root);
               continue;
             }
+            if (paramSummary.hasUnverifiedCellUse) {
+              markUnverifiedCellUse(source.root);
+            }
 
             for (const readPath of paramSummary.readPaths) {
               trackRead(source.root, [...source.path, ...readPath]);
@@ -3004,7 +3032,19 @@ export function analyzeFunctionCapabilities(
                 markOpaqueUse(receiver.root, receiver.path);
               }
             } else {
-              // Unknown method call over a tracked source reads at least the receiver path.
+              // Unknown method call over a tracked source reads at least the
+              // receiver path. On a CELL-LIKE receiver the unknown method
+              // could be a mutator this analysis does not recognize (the
+              // writer sets are a closed list against an evolving Cell API),
+              // so write-exhaustiveness is poisoned — fail closed. Value
+              // receivers (e.g. array methods on a `.get()` snapshot) cannot
+              // write through the cell and stay reads.
+              if (
+                !checker ||
+                isCellLikeExpression(node.expression.expression)
+              ) {
+                markUnverifiedCellUse(receiver.root);
+              }
               if (shouldTrackFullShape) {
                 trackFullShapeReadRef(receiver);
               } else if (directReceiver) {
--- a/packages/ts-transformers/src/policy/capability-analysis.ts
+++ b/packages/ts-transformers/src/policy/capability-analysis.ts
@@ -156,10 +156,13 @@ const PARAMETER_SUMMARY_PREFIX = "__param";
 const mergeableMethods = (kind: MergeableOpMethodKind): string[] =>
   MERGEABLE_OP_METHODS.filter((op) => op.kind === kind).map((op) => op.method);
 
-// `send` is a write: at runtime Stream.send() delegates to set() (an event
-// enqueue is a write to the stream cell). Classifying it here keeps
-// writePaths an honest record of capture writes — a callback that fires
-// events during evaluation must never summarize as write-free.
+// `send` is a write: it delegates to set() on every receiver, but set()
+// forks on the receiver's kind. On a raw cell that is an ordinary
+// transactional write; on a stream the same set() call enqueues the event
+// instead of revising the stream cell, and the dispatched handler's writes
+// commit in their own transaction — effects this analysis cannot see or
+// bound. Either way a callback that sends must never summarize as
+// write-free.
 const WRITER_METHODS = new Set([
   "set",
   "update",
--- a/packages/ts-transformers/src/policy/capability-analysis.ts
+++ b/packages/ts-transformers/src/policy/capability-analysis.ts
@@ -80,9 +80,10 @@ interface MutableCapabilityState {
    * Write-exhaustiveness is unverifiable for this parameter. Set by an
    * unrecognized or dynamic (`cell[m]()`) method call on a cell-like
    * receiver (the call could be a mutator this analysis does not know), and
-   * by recognized `set`/`send` calls carrying an onCommit callback (which
-   * receives the committed transaction and can write arbitrary cells
-   * through it). Consumers asserting write exhaustiveness must fail closed
+   * by recognized `set`/`send` calls carrying an onCommit callback (whose
+   * closure can escape into fresh transactions or external I/O after
+   * commit — writes this analysis cannot bound). Consumers asserting
+   * write exhaustiveness must fail closed
    * on this, like `wildcard`; recognized reads/derivations are unaffected.
    *
    * Syntactic boundary: detection is per method-CALL dispatch. An extracted
@@ -2996,12 +2997,12 @@ export function analyzeFunctionCapabilities(
             } else if (WRITER_METHODS.has(methodName)) {
               trackWriteRef(receiver);
               recordMergeableNonAppendWrite(receiver);
-              // set(value, onCommit?) / send(event, onCommit?): the onCommit
-              // callback receives the committed TRANSACTION and can write
-              // arbitrary cells through it — writes invisible to capture-path
-              // tracking (even an inline arrow the nested walk analyzes
-              // writes via `tx`, not via captures). A second argument means
-              // `writes` is not exhaustive; fail closed.
+              // set(value, onCommit?) / send(event, onCommit?): onCommit runs
+              // after the commit settles, so the passed tx is no longer a
+              // write channel — but the closure can still write via captured
+              // cells, fresh transactions (cell.runtime.edit()), or external
+              // I/O, and the contract forbidding that is advisory. That
+              // residue is unboundable, so a second argument fails closed.
               if (
                 (methodName === "set" || methodName === "send") &&
                 node.arguments.length > 1
--- a/packages/ts-transformers/src/policy/capability-analysis.ts
+++ b/packages/ts-transformers/src/policy/capability-analysis.ts
@@ -2975,6 +2975,18 @@ export function analyzeFunctionCapabilities(
             } else if (WRITER_METHODS.has(methodName)) {
               trackWriteRef(receiver);
               recordMergeableNonAppendWrite(receiver);
+              // set(value, onCommit?) / send(event, onCommit?): the onCommit
+              // callback receives the committed TRANSACTION and can write
+              // arbitrary cells through it — writes invisible to capture-path
+              // tracking (even an inline arrow the nested walk analyzes
+              // writes via `tx`, not via captures). A second argument means
+              // `writes` is not exhaustive; fail closed.
+              if (
+                (methodName === "set" || methodName === "send") &&
+                node.arguments.length > 1
+              ) {
+                markUnverifiedCellUse(receiver.root);
+              }
             } else if (ARRAY_IDENTITY_WRITER_METHODS.has(methodName)) {
               trackWriteRef(receiver);
               if (mergeablePushMisuseSink && !receiver.dynamic) {
--- a/packages/ts-transformers/src/policy/capability-analysis.ts
+++ b/packages/ts-transformers/src/policy/capability-analysis.ts
@@ -77,11 +77,19 @@ interface MutableCapabilityState {
   hasNonIdentityUse: boolean;
   hasNonIdentityRootUse: boolean;
   /**
-   * An unrecognized method was called on a cell-like receiver rooted in this
-   * parameter. The call could be a mutator this analysis does not know, so
-   * `writes` cannot be treated as exhaustive — consumers asserting write
-   * exhaustiveness must fail closed on this, like `wildcard`; recognized
-   * reads/derivations are unaffected.
+   * Write-exhaustiveness is unverifiable for this parameter. Set by an
+   * unrecognized or dynamic (`cell[m]()`) method call on a cell-like
+   * receiver (the call could be a mutator this analysis does not know), and
+   * by recognized `set`/`send` calls carrying an onCommit callback (which
+   * receives the committed transaction and can write arbitrary cells
+   * through it). Consumers asserting write exhaustiveness must fail closed
+   * on this, like `wildcard`; recognized reads/derivations are unaffected.
+   *
+   * Syntactic boundary: detection is per method-CALL dispatch. An extracted
+   * method reference (`const f = cell.send; f(x)`) bypasses it — marginal
+   * in practice (unbound cell methods break at runtime), but consumers
+   * treating `!wildcard && !hasUnverifiedCellUse` as a soundness
+   * certificate should know the contract's edge.
    */
   hasUnverifiedCellUse: boolean;
 }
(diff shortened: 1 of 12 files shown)


packages/ts-transformers/src/policy/capability-analysis.ts
```diff
@@ -147,9 +156,14 @@ const PARAMETER_SUMMARY_PREFIX = "__param";
...
   MERGEABLE_OP_METHODS.filter((op) => op.kind === kind).map((op) => op.method);
 
+// `send` is a write: at runtime Stream.send() delegates to set() (an event
+// enqueue is a write to the stream cell). Classifying it here keeps
+// writePaths an honest record of capture writes — a callback that fires
+// events during evaluation must never summarize as write-free.
 const WRITER_METHODS = new Set([
   "set",
   "update",
+  "send",
```
mathpirate, line 178: **[Blocking] Send paths now reach emitted `materializerWriteInputPaths` — a runtime-visible change, invisible in goldens.**

Verified at ff6c001: a send-only computed (stream capture, mirroring the side-writing regression test) now emits

    …{ materializerWriteInputPaths: [["notify"]] });

where the same emitted schema brands `notify` as `asCell: ["stream"]`. Pre-PR this computed emitted no scheduler options. Two consequences:

1. The field's runtime meaning is "input paths whose **writable cells** become materializer envelopes" — the runner's collector matches only `cell`/`writeonly` brands (`collectWritableCellArgumentLinks`), so a stream path can never produce an envelope. It's dead metadata in every send-containing computed artifact.
2. The field's *presence* switches the runner off the opaque-result fallback (`runner.ts`, the `module.materializerWriteInputPaths !== undefined` branch). A computed that sends *and* returns an opaque result *and* has writable args silently changes from "collect all writable args" to "collect none". Narrow, but real — and no fixture covers send-inside-computed, which is why the goldens only show reordering.

Two clean resolutions:

- **(preferred)** filter stream-kind write paths at `createDeriveSchedulerOptions` (the strategy has the checker and capture types), keeping the emitted field cell-writes-only — the analysis-level write classification stays exactly as this PR has it. Filtering by capture *kind* rather than by method also does the right thing for `send` on a raw writable cell (`ICell` includes `IStreamable`, so `Writable<T>.send(v)` type-checks): that's a genuine transactional write to that location, and its path stays in the field.
- keep the paths as a deliberate future hook, but then name the fallback flip in the PR description rather than shipping it silently.

Either way, a send-only-computed emission pin in `pipeline-regressions.test.ts` (next to the side-writing/readonly pair) would lock the chosen semantics — I have a ready-made probe test that can be adapted.

mathpirate (commented):
cf-review via Claude Fable 5 (claude-fable-5), on behalf of @mathpirate.

Direction is right, and `hasUnverifiedCellUse` is a genuinely good generalization — "the writer list is closed but the Cell API isn't" is a stronger foundation for the downstream write-exhaustiveness consumer than a method-list patch alone. Reviewed at 76a83f8 and re-verified against ff6c001 — the onCommit arity split, which resolves the thread's onCommit question cleanly (the raw-transaction rationale in its comment is exactly the right one) and comes with tests covering all three cases. Transformer suite green at review time, and the golden churn is what the description says it is (order-only, brands preserved — verified).

One blocking decision before merge (inline at the `send` classification): send-classified stream paths now flow into emitted `materializerWriteInputPaths`, which is a runtime-visible change the description's inertness framing doesn't cover — I verified the emission empirically at both heads. Details and two resolution options inline. The rest is improvements: a rationale comment that mis-describes the stream case (the classification survives the correction and comes out stronger), a plausible cause for the Performance Check failure worth measuring before accepting new baselines, one cheap hole in the fail-closed net plus a now-live doc drift on the field's contract, and one untested claim.

packages/ts-transformers/src/policy/capability-analysis.ts
```diff
@@ -147,9 +156,14 @@ const PARAMETER_SUMMARY_PREFIX = "__param";
 const mergeableMethods = (kind: MergeableOpMethodKind): string[] =>
   MERGEABLE_OP_METHODS.filter((op) => op.kind === kind).map((op) => op.method);
 
+// `send` is a write: at runtime Stream.send() delegates to set() (an event
+// enqueue is a write to the stream cell). Classifying it here keeps
+// writePaths an honest record of capture writes — a callback that fires
+// events during evaluation must never summarize as write-free.
```
mathpirate: **[Improvement] The rationale comment teaches the wrong transaction model for the stream case — worth splitting into the two real cases, which make the classification stronger, not weaker.**

`send` does ride `set()`'s code path, but the shared method forks on the receiver. On a **raw cell** it is a genuine transactional set — and this is typed pattern surface, not just runtime trivia: `ICell<T>` includes `IStreamable<T>`, so `Writable<T>.send(v)` type-checks. On a **stream** it never writes the stream cell and needs no transaction at all: it enqueues the event, exactly one locally-registered handler dispatches in a fresh transaction, and the *handler's* sets are what commit (`packages/runner/src/scheduler/events.ts`, `dispatchQueuedEvent`). "An event enqueue is a write to the stream cell" is the clause that will mislead someone reasoning about tx semantics or event persistence from here.

Suggested replacement:

    // `send` is a write: on a raw cell it is set() (a transactional write);
    // on a stream it enqueues an event that dispatches a handler in its own
    // transaction, causing writes this analysis cannot see or bound. Either
    // way a callback that sends must never summarize as write-free.

Same wording appears in the new send test's comment and the PR description.

packages/ts-transformers/src/policy/capability-analysis.ts
```diff
@@ -3004,7 +3044,19 @@ export function analyzeFunctionCapabilities(
...
+              // Unknown method call over a tracked source reads at least the
+              // receiver path. On a CELL-LIKE receiver the unknown method
+              // could be a mutator this analysis does not recognize (the
+              // writer sets are a closed list against an evolving Cell API),
+              // so write-exhaustiveness is poisoned — fail closed. Value
+              // receivers (e.g. array methods on a `.get()` snapshot) cannot
+              // write through the cell and stay reads.
+              if (
+                !checker ||
+                isCellLikeExpression(node.expression.expression)
```
mathpirate, line 3078: **[Improvement] Plausible cause of the Performance Check failure — worth measuring before accepting the new baselines.**

This adds a `checker.getTypeAtLocation` for every unknown-*named* method call over a tracked receiver — i.e. every `.join()` / `.includes()` / `.reduce()` / `.slice()` chain in every pattern body, which is a hot set. The perf ratchet flagged five entries at once across different pattern jobs; history has similar-size single-test outliers on unrelated PRs, so it may be noise — but five-at-once with a ready mechanism deserves a measurement rather than a paste-accept.

If it's real, two cheap options: memoize cell-likeness per receiver node for the analysis run, or make the marking opt-in via `CapabilityAnalysisOptions` until the write-exhaustiveness consumer lands (the bit is deliberately inert today, so nothing loses coverage in the interim).

packages/ts-transformers/src/policy/capability-analysis.ts
```diff
@@ -1471,6 +1487,15 @@ export function analyzeFunctionCapabilities(
       state.hasNonIdentityUse = true;
     };
 
+    // Unlike markWildcard this does NOT change shrinking or identity
+    // classification — it only poisons write-exhaustiveness for consumers
+    // that need `writes` to be a closed-world record, while everything else
+    // behaves as before.
+    const markUnverifiedCellUse = (name: string): void => {
```
mathpirate, line 1506: **[Improvement] The fail-closed net misses dynamic method names, and the field's contract has drifted from its behavior.**

`cell[m]()` (non-literal element access) takes the `!methodName` branch in the call dispatch, which never reaches the new marking — yet an unresolvable method name on a cell-like receiver is the open-world case this bit exists for. One line under the same `!checker || isCellLikeExpression(...)` guard in that branch closes it.

A residual escape probably worth *documenting* on the summary field rather than fixing: extracted method references (`const f = cell.send; f(x)`) bypass method-call dispatch entirely. Marginal in practice (unbound cell methods break at runtime), but if consumers will treat `!wildcard && !hasUnverifiedCellUse` as a soundness certificate, the JSDoc should state the contract's syntactic boundary.

Related doc drift as of ff6c001: the field's JSDoc (here and on `CapabilityParamSummary` in `core/transformers.ts`) still reads "an **unrecognized** method was called on a cell-like receiver," but the onCommit split now also sets it for *recognized* `set`/`send` calls. Worth widening the field docs to name both triggers while touching this.

packages/ts-transformers/src/policy/capability-analysis.ts
```diff
@@ -2814,6 +2839,9 @@ export function analyzeFunctionCapabilities(
               markWildcard(source.root);
               continue;
             }
+            if (paramSummary.hasUnverifiedCellUse) {
+              markUnverifiedCellUse(source.root);
```
mathpirate, line 2855: **[Improvement] The interprocedural propagation is claimed and correctly placed, but untested.**

The PR description says the mark "propagates through interprocedural summaries the same way `wildcard` does" — this is the one piece of new machinery with real behavior, and none of the new tests exercises it. A same-file helper test would pin it:

    const helper = (c: Cell<Thing>) => c.frobnicate();
    const fn = (input: Cell<Thing>) => { helper(input); return null; };
    // expect: input.hasUnverifiedCellUse === true

(Placement note, for the test's comment: wildcard subsumes the mark when the arg is dynamic — the `continue` above — which is the right precedence given consumers must fail closed on either.)

packages/ts-transformers/src/policy/capability-analysis.ts
ubik2: Resolved as a documented variant of option B, pinned by the requested emission test (1a03f5812: "send-only computed keeps stream paths in write metadata but never brands them collectible").

**The send paths stay in the emitted field; the envelope split you'd want happens at the collector, which already implements it.** `collectWritableCellArgumentLinks` filters to `cell`/`writeonly` brands: `send` through a raw writable cell produces an envelope (correct — a transactional write to that address, per your `ICell includes IStreamable` observation), while a stream capture keeps its `stream` brand via declared-type preservation and produces none (correct — a stream send writes no address; the dispatched handler's writes belong to the handler's own action, which the scheduler already tracks). The cost of keeping the paths is a few never-collectible entries in send-containing artifacts.

**Why not filter at emission (the preferred option):** the downstream write-exhaustiveness consumer requires the emitted write paths to be empty before treating a computed as a pure derivation. A computed that fires events during evaluation must never present as write-free — replaying or dropping its commit re-fires the event. The stream-send paths in the field are exactly what keep such a computed disqualified; filtering them at emission would reopen that hole and require a second channel to carry the same bit.

**The fallback flip: agreed it's real, and we're keeping it as intended precision (now named in the PR description).** For a module carrying analyzed write metadata, the field is an exhaustive record — in the tri-condition case (sends + opaque result + writable args), writable args that were never written would appear in the write paths if they were written, so "collect none" is more precise than the opaque fallback's "collect all writable args." The fallback remains what it was for: artifacts without metadata.

**Residual corner, acknowledged:** when a stream capture's declared type degrades (`Reactive`/`OpaqueRef` — the `preservedWrapperFor` lossiness), a sent-through stream brands `writeonly` and becomes collectible, yielding a spurious envelope at the stream address. Mostly harmless (nothing reads stream addresses) and pre-existing lossiness rather than new behavior; tightening stream recognition in `preservedWrapperFor` is the right fix and out of scope here.

Edit: This was Claude's comment

packages/ts-transformers/src/policy/capability-analysis.ts
ubik2: Opted for a compromise on the language, where I leave the `send` is a `set`, but then clarify that streams don't do a real write on a `set`.

packages/ts-transformers/src/policy/capability-analysis.ts
ubik2: Also added a number of tests to pin this behavior.

packages/ts-transformers/src/policy/capability-analysis.ts
ubik2: All three parts addressed in aa9dd1985:

- **Dynamic method names**: the `!methodName` branch now marks under the same `!checker || isCellLikeExpression(...)` guard as the unknown-named fallback. Tests cover the cell-typed receiver (marked), the `.get()`-snapshot receiver (stays verified), and the no-checker case (fails closed).
- **Extracted method references**: documented rather than fixed, per your triage — both declaration sites of `hasUnverifiedCellUse` now state the contract's syntactic boundary (detection is per method-call dispatch; `const f = cell.send; f(x)` is outside it).
- **Doc drift**: the field JSDoc at both sites now names all three triggers — unrecognized methods, dynamic methods, and `set`/`send` carrying an onCommit callback.

--Claude

packages/ts-transformers/src/policy/capability-analysis.ts
ubik2: Measured. The mechanism cannot account for the ratchet failures — they read as the known single-test-outlier noise.

Method: temporarily instrumented both guard sites (the unknown-named fallback and the dynamic-name branch) with call counts and accumulated in-guard time, then compiled 18 real patterns (the census corpus: exemplars + system/home + default-app) through the full pipeline, 3 runs per configuration, A/B against the guard short-circuited.

Results:
- Guard invocations across the corpus: **2,017**; in-guard time **6.4ms** with the type lookup vs **2.6ms** short-circuited (that baseline is the instrumentation's own overhead) → net cost of `getTypeAtLocation` ≈ **3.8ms per 18 patterns**, ~**2µs per call**.
- Total compile wall time: **10.7–10.9s** in both configurations — deltas within run-to-run noise (±150ms), i.e. the guard is ~**0.04%** of compile.
- Duplicate receiver nodes: 1,529 of 2,017 (~76%) — so a per-node memo would recover at most ~3ms per full corpus compile.

Conclusion: no cache — 3ms of headroom doesn't justify the extra state, and the five-at-once ratchet entries need a different explanation (or a re-run). Happy to re-trigger the Performance Check to confirm.

--Claude

packages/ts-transformers/src/policy/capability-analysis.ts
ubik2: Covered in bbceec5d7, in the interprocedural suite, with your suggested shape plus two companions:

- **Propagation**: `helper = (c: Cell<…>) => c.frobnicate()` called from `fn(input)` → `input.hasUnverifiedCellUse === true` with `wildcard === false` — asserting both, so the test proves it's the mark propagating through the callee summary, not the wildcard channel.
- **Negative control**: a read-only callee (`c.get()`) propagates nothing.
- **Precedence**: a dynamic argument path (`input.key("items").key(k)`) takes the wildcard `continue` before the propagation — pinned with your placement note in the test comment: the subsumption is fine because consumers must fail closed on either flag.
--Claude

packages/ts-transformers/src/policy/capability-analysis.ts
ubik2: Follow-up: the Performance Check failure is reproduced and root-caused — it is the **compile byte cache going cold, not the code**.

The pattern-unit jobs restore `.compile-cache/pattern-unit-coverage-*.json` under a key that hashes `packages/ts-transformers/**`. This PR changes ts-transformers, so its first CI run compiles every pattern cold, while all 20 baseline main runs were warm. Any PR touching the transformer takes this hit.

Local measurements (worktrees, same machine):

- **Branch vs main, cold-vs-cold (fair code comparison), all flagged tests**: counter 1.0s vs 1.0s; multi-user 5.3–6.2s vs 5.2–6.0s; lobby 1.9s vs 1.9s; proxy-length-repro 0.9s vs 0.9s. No difference.
- **Warm vs cold byte cache, same code** (`CF_COMPILE_CACHE_FILE`): multi-user 3.1s warm vs 5.4s cold — **1.74× cold**, matching CI's observed 1.66× on the same test (14.2s baseline → 23.6s) almost exactly. The small compile-dominated tests show larger ratios (counter's CI 1.93× fits).

This also explains the shape of the report: six OVER plus 76 CLOSE spread uniformly across every pattern job — a cache-state effect, not a hot spot.

Actionable: re-running the pattern jobs should go warm (the first run saved the new cache key) and the check should pass. Structurally, the ratchet penalizes every transformer-touching PR's first run; comparing cold runs against a warm baseline is a false positive by construction — probably worth a note in CI_PERFORMANCE.md or a first-run-after-key-change exclusion, separately from this PR.

mathpirate (approved):
cf-review via Claude Fable 5 (claude-fable-5), on behalf of @mathpirate.

**Approving.** Every thread from the previous review is addressed at or above the asked level — in one case (the Performance Check root-cause) well beyond it — and I independently verified the load-bearing claims rather than taking the replies on trust (cache-key mechanism confirmed in the workflow; emission, collector, and interprocedural pins read; warm rerun green with no timing baselines accepted).

One resolution deserves explicit record: the Blocking finding (stream paths in emitted `materializerWriteInputPaths`) is **accepted on the merits of the counter-argument**, not waved through — a non-empty write record is precisely what keeps an event-firing computed out of the pure-derivation class, and the distinction it preserves (definitively-effectful vs. unknown) is worth having. With the fallback flip named in the description and the semantics pinned at three layers, this is the documented-acceptance path, done properly.

Two non-blocking residuals, natural to fold into the follow-up flag PR rather than this one:

1. `packages/runner/src/builder/types.ts` still documents `materializerWriteInputPaths` as an envelope selector; under the now-official semantics it's an exhaustive analyzed write record that deliberately includes never-collectible stream paths. One line, and the flag PR touches `DeriveSchedulerOptions` anyway.
2. The description's contract note on onCommit is more precise than the merged comment/JSDoc ("can write arbitrary cells through" a tx that is settled by callback time — the real residue is fresh-transaction escapes and external I/O). Worth promoting the note's phrasing into the code, since descriptions don't merge and comments do.

Follow-ups proposed, for a human to file separately: tighten stream recognition in `preservedWrapperFor` (the degraded-capture spurious-envelope corner), and the CI ratchet's first-run-after-cache-key-change false positive, which this PR demonstrated in both directions (timing and coverage-debt).

This lands the foundation for the write-exhaustiveness consumer in better shape than it started: the gate is now concretely `!wildcard && !hasUnverifiedCellUse && !recursive`, and the send-paths-stay decision is pinned well enough that the next PR can rely on it.

PR #4556, opened 2026-07-06
fix(runner): array-diff shrink emits explicit slot deletes (clears stale per-slot link labels)

mathpirate:
## Why

An array-diff shrink (`cell.set([...fewer])`) emits only the `length` change — the truncated slots are never written. Their per-slot `origin:"link"` labelMap entries therefore survive in storage after the slots themselves are gone. Since C1 (#4507), any later read or diff that touches such a slot — most simply, the list growing back over it — consumes the stale entry as a followRef observation (SC-8) and re-imports the **departed** element's confidentiality into the reader's flow join. For membership labels that §8.12.8 says *replace-from-criteria* (#4546), this manufactures accumulation the discipline forbids: it is the bridge behind the A3 step of the probe on #4525, where a member two membership-changes gone kept re-appearing in a filter container's `enumerate` label.

This is a bug independent of any coordinator: it needs only a list of labeled links and a shrink. (The filter/flatMap work in #4391 makes it *visible* on membership stamps, but the stale entries exist either way.)

## The fix

The diff layer already has the right idiom: a **direct** `length` write emits explicit per-slot deletes for the removed range. The array-diff branch now does the same for a shrink — with the deletes ordered **before** the length change, because once the length write has truncated the array, a delete at a now-absent slot is a no-op the write layer elides from the journal (verified via write details; this ordering is why the naive "same idiom, same order" version silently changed nothing).

With the removal visible as a write, the flow-label carry-forward drops the stale entries through the **existing** written-path clear — no label-layer changes at all. Growth needs nothing: the element loop already visits every new slot.

## Verification

- **Red-first:** `cfc-array-shrink-slot-labels.test.ts` — on main, the truncated slot's link entry survives the shrink (`["bob-secret"]`); with the fix it is cleared, the survivor keeps its label, and a regrown slot carries only its new occupant (no departed-member residue for followRef to re-import).
- The pinned diff contract in `data-updating.test.ts` ("array length changes") updates from `[length]` to `[delete, length]`, matching the direct-length-write path's shipped contract.
- Blast radius: cfc + list-builtin + cell + resume + data-updating suites — 136 files / 1044 steps, 0 failed. Full runner suite green (see checks).

## Relation to open work

#4391's rebase onto #4546 composes with this: the shrink probe's A-scenario converges once the stale-slot bridge is gone (one reconcile of hysteresis remains — the shrink reconcile itself genuinely observes the departing slot's pointer — which is SC-8-consistent). Details on that thread.

*Analysis, fix, and tests by Claude Fable 5; PR by @mathpirate.*

---
## Summary by cubic
Fix array-diff shrink and growth ordering to clear stale per-slot link labels and ensure ["length"] labels re-stamp correctly. Shrinks now emit per-slot deletes before the length write; growth emits the length change before element writes.

- **Bug Fixes**
  - Shrink: emit delete ops for each truncated slot before the `length` write; holes are skipped. Clears stale per-slot `origin: "link"` labels and matches the direct `length`-write path.
  - Growth: emit the `length` change before element writes to prevent elision, ensuring the `["length"]` value entry is cleared/re-stamped from the current tx’s join.
  - Tests: added `cfc-array-shrink-slot-labels.test.ts`; updated `data-updating.test.ts` with shrink `[delete, length]` and growth `[length, element]` ordering, plus a `["length"]` value-label re-stamp assertion.

<sup>Written for commit 867afe87f6dfde095f6cbf7e85fef4fe9e386112. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4556?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/data-updating.ts
+++ b/packages/runner/src/data-updating.ts
@@ -991,6 +991,26 @@ export function normalizeAndDiff(
       const childSchema = (lub !== undefined)
         ? { type: "number", ifc: { confidentiality: lub } } as JSONSchema
         : { type: "number" } as JSONSchema;
+      // Slots truncated by a shrink are removed, not merely out of range:
+      // emit explicit deletes (the direct `length`-write path's idiom) so
+      // write-detail consumers — notably the flow-label carry-forward that
+      // must drop the removed slots' stale per-slot link entries — see the
+      // removal. Ordered BEFORE the length change: once the length write
+      // has truncated the array, a delete at a now-absent slot is a no-op
+      // the write layer elides from the journal. Growth needs nothing
+      // here — the element loop above already visited every new slot.
+      for (let i = newValue.length; i < currentValue.length; i++) {
+        if (!(i in currentValue)) continue; // hole: nothing to remove
+        changes.push({
+          location: {
+            ...link,
+            path: [...link.path, i.toString()],
+            schema: runtime.cfc.getSchemaAtPath(link.schema, [i.toString()]),
+          },
+          value: undefined,
+          delete: true,
+        });
+      }
       changes.push({
         location: {
           ...link,
--- a/packages/runner/src/data-updating.ts
+++ b/packages/runner/src/data-updating.ts
@@ -923,6 +923,32 @@ export function normalizeAndDiff(
     // Get current array for precomputing child values (if it was an array)
     const currentArray = Array.isArray(currentValue) ? currentValue : undefined;
 
+    // On GROWTH the length change must precede the element writes: applying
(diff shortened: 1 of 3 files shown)


ubik2 (approved):
makes sense to me

PR #4561, opened 2026-07-06
feat(runner): scope-isolation write guard — reject narrower-scoped links in broader slots (B1)

bfollington:
**TL;DR:** `normalizeAndDiff` now **warns at the write site** when a narrower-scoped link (e.g. `user`) is stored into a broader slot (e.g. shared `PerSpace` state) — the pitfall-6 silent hole behind the group-chat reader blackout. Warn, not throw. **@ubik2's inline comment is implemented and completed** (tolerance + parent-`required`), and completing it made the ambient suppression from the earlier revision **deletable — no module-global state remains**. The guard already caught and fixed a latent bug in shipped `scrabble`.

**Reading map (the actual review surface):** the guard + helpers in `packages/runner/src/data-updating.ts` (~120 lines) and the two normative doc paragraphs (pitfall 6, spec write-semantics). **Skip:** `traverse.ts` (mechanical extraction of `isValidType` → exported `schemaAcceptsType`; replay goldens pin it behavior-identical), the test file (11 pins, titles suffice), the fixtures (migration demo), and everything in `<details>` below.

## When does it warn? (the criterion)

| slot situation | warns? | why |
| --- | --- | --- |
| schema **declares the scope** (`PerUser<Cell<T>>` / schema `scope` / scoped `asCell`) and the link is **at or broader than that cap** | no | per seefeld: a declared scope is a **cap** — content may be at most that narrow; same-or-broader is correct usage (same `getSchemaScopeCap` precedence as the read follow-cap) |
| schema declares the scope but the link is **narrower than the cap** (e.g. session link in a `PerUser` field) | **yes** | violates the declared cap ("you can't assign a PerSession value to a PerUser field") |
| schema **matches `undefined`** | no | a reader's empty resolution degrades harmlessly |
| schema has an **effective `default`** (loose `!= undefined`, judged on the `$ref`-resolved schema — mirrors how the read applies defaults; `default: null` does **not** count) | no | the reader gets the default |
| **optional under its parent** (parent schema visible, key not in `required`) | no | the read drops the field and the object survives — the `optlink` control shape that converges with no fix |
| **required + strict + undeclared** | **yes** | the slot's shape says the author wanted *shared* data, and a principal-less link can never deliver it to other readers |
| direct slot write (`cell.key().set()`, bound handler cell) — **parent not in view** | yes, if strict | requiredness unknown on this route; declare the scope or write through the parent to silence. **This is the whole direct-write route, not an edge case.** |
| re-serializing the identical stored link | no | no-op branch precedes the guard; at-rest data never warns on unrelated writes |

**Honesty note (from an external cold review of this PR):** this criterion *approximates* — it does not prove — "the read would reject the hole." Two known deviations, both deliberate: (a) the B2 element grace (#4532/#4535) means array-element reads *degrade* rather than void, so for elements the warn's justification is "the data can never propagate," not "the read voids"; (b) rejection is ultimately judged against each **reader's** combined schema, which a write-site lint cannot know. Also deliberate, false-negative direction: `required` inside compound parents (`allOf`/`anyOf`) is not merged — such parents read as requiring nothing.

## Open questions (the actual ask)

1. **Criterion completeness (ubik2):** read-side `required` is the intersection of the entered schema and the link-embedded schema. Can the effective read-side requiredness be *stricter* than this write site sees, and is under-warning there acceptable for a lint?
2. **Path to error:** flip to a throw once CI/staging shows zero machinery warns for a while, or keep it a permanent lint with authoring-layer (CTS) enforcement as the real fix?
3. ~~Opt-in key~~ **Answered by seefeld:** the declaration is a scope **cap** (content at most that narrow; broader allowed) — the guard already implements exactly this. **Sharpened follow-up:** should a *declared-cap violation* (session link into a `PerUser` field — "can't") be a hard **error** on write, separate from the undeclared-slot heuristic warn?
4. **Warn delivery:** the warn goes through the level-gated `normalizeAndDiff` logger. Is warn-level visible in live browser dev / `cf` CLI where authors would actually see it? If not, the lint only bites in test lanes and we should say so.

## What's already settled (please don't re-derive)

- **Write-side is the right layer per the spec** (`scoped-cell-instances.md`): *"links do not encode the DID or session id"* — per-reader isolation is deliberate, so owner-qualified links (my first proposal on the #4532 thread) are off the table. Berni's instinct there was right.
- **B2** (read-side element grace, merged #4532/#4535) is unchanged and composes: undeclared strict slots warn at write; declared/optional slots resolve per-reader and degrade gracefully on read.
- **B3** (#4533) is untouched by this PR.

<details>
<summary><strong>Background: how we got here (narrative, not required reading)</strong></summary>

1. **#4532 (merged)** fixed the multiplayer reader blackout read-side: an array element with a required field holding an unresolvable link voided the entire array read for every non-author; the fix degrades the one field instead.
2. **On the #4532 thread**, Berni asked what *writes* those unresolvable links. Trace: a handler pushes a live `PerUser` cell into shared `PerSpace` state; the stored link says `scope: "user"` with no principal; every reader dereferences it into their own empty partition. I proposed owner-qualified links; the spec explicitly forbids encoding principals in links; hence this write-side guard.
3. **Revision history (each round driven by CI or reviewer feedback):** R1 hard throw → CI enumerated six sanctioned machinery writers of narrow links into scope-silent slots (`.asScope()` result links, result-schema scope overrides, cross-space scoped links, `navigateTo` result cells, `Runner.updateArgument` setup wiring, cold-resume re-scope walks). R2 warn → the pattern-unit lane fails on **any** console warning. R3 ambient suppression wrapper for `updateArgument` (and the remaining warn exposed the scrabble bug). R4 ubik2's criterion (slot tolerance). R5 completed the criterion (parent-`required` threading; canonical matcher reuse) — which made all machinery writes quiet by their own shape, so the suppression was deleted. R6 (current): external cold-review round — corrected the over-stated equivalence claim in prose, mirrored the read side's loose default application (`default: null` no longer counts as tolerant), documented the compound-parent gap and the direct-write route-dependence, removed a leftover block, deduped a `resolveSchema` call, added two pins.

</details>

<details>
<summary><strong>The guard's first catch: a real latent bug in shipped scrabble (fixed here)</strong></summary>

While landing the warn it kept firing in `scrabble.test.tsx` — the guard working as designed: `submitTurn` wrote `board.set([...currentBoard, ...tilesInWords])` where the tiles are `placed.get()` query proxies keeping their cell identity in the **user-scoped** `placed` doc. The shared board was storing user-scoped links, immediately dangled by the following `placed.set([])` — every other player resolves those tiles into their own empty partition. Fixed by deep-snapshotting the tiles as plain values (the nested `Letter` object also carries identity through a spread, so the copy is two levels).

</details>

<details>
<summary><strong>Verification</strong></summary>

- 11 unit pins (`data-updating.test.ts`, "scope-isolation write guard"): strict scope-silent slot warns and the write still lands; declared slot silent; session link under a user-declared slot warns; same-scope silent; undefined-tolerant silent; defaulted silent; `default: null` warns; strict `$defs`/`$ref` slot warns (judged on the resolved schema); required-through-parent warns; optional-through-parent silent; identical-link rewrite silent.
- 81 runner tests / 257 steps green across data-updating, traverse (incl. replay goldens — the matcher extraction is behavior-identical), pattern-scope, navigate-handler, both outbox suites, schema-links, list-builtin-edge-paths.
- scrabble unit + multi-user green with zero warns and no suppression; convergence-storm 4/4; scoped-group-chat, cfc-group-chat-demo multi-runtime, lunch-poll, cozy-poll, self green.
- Docs updated: pitfall 6 and the spec write-semantics paragraph state the criterion with its known deviations.

</details>

**Non-blocking design questions** surfaced by this work (runtime-warn vs CTS-level unrepresentability; proxy spreads silently keeping nested cell identity; schema-annotation loss between authored types and stored links) are tracked in #4591 and do **not** block this review.

Coverage-debt override: the warn-message closure is never evaluated in coverage lanes (the `normalizeAndDiff` logger is level-gated there), even though unit pins assert the warn fires.

NEW_PERF_BASELINE: coverage-debt: packages/runner uncovered lines = 7660 lines

### DIFF
--- a/packages/runner/src/data-updating.ts
+++ b/packages/runner/src/data-updating.ts
@@ -816,31 +816,39 @@ export function normalizeAndDiff(
       // DIFFERENT instance for every reader — links deliberately do not encode
       // the principal, so e.g. a `user`-scoped link stored in `space`-scoped
       // shared data hands every other participant a link to their own (empty)
-      // instance. The data can never propagate; readers see a permanent hole
-      // (the B2 reader-blackout investigation, #4457/#4532). Writing narrow
-      // data into a broader slot is legitimate ONLY when the slot's schema
-      // declares that scope (the narrowing branch above and scoped asCell
-      // entries — the author opted into per-reader semantics). Otherwise the
-      // write is a silent footgun: fail loudly at the write site, where the
-      // author can fix it, instead of as a permanent silent hole for every
-      // other reader. Share the value, or a space-scoped cell with a PerUser
-      // pointer to "mine" (pitfall #6 shows the idiom).
+      // instance. When the write meant to SHARE data, it can never propagate;
+      // readers see a permanent hole (the B2 reader-blackout investigation,
+      // #4457/#4532). The slot's schema declaring the scope (the narrowing
+      // branch above and scoped asCell entries) is the explicit opt-in to
+      // per-reader semantics. This is a WARN, not a throw, because the
+      // runtime's own machinery legitimately writes scoped links into
+      // scope-silent slots today (pattern factory .asScope() result links,
+      // navigateTo result cells, Runner.updateArgument setup wiring,
+      // cold-resume re-scope walks — enumerated on #4561); until those slots
+      // declare their scope, author footguns and sanctioned machinery writes
+      // are indistinguishable here. Authors: share the value, or a
+      // space-scoped cell with a PerUser pointer to "mine" (pitfall #6 shows
+      // the idiom).
       if (scopeRank(parsedLink.scope) > scopeRank(link.scope)) {
         const declared = declaredCellScope(link.schema);
         if (
           declared === undefined ||
           scopeRank(declared) < scopeRank(parsedLink.scope)
         ) {
-          throw new Error(
-            `Cannot store a ${parsedLink.scope}-scoped link in ` +
+          diffLogger.warn(
+            "diff",
+            () => [
+              `Storing a ${parsedLink.scope}-scoped link in ` +
               `${link.scope}-scoped data at path "${pathStr}": scoped links ` +
-              `do not carry a principal, so every reader would resolve it ` +
-              `to their own ${parsedLink.scope} instance and the data can ` +
-              `never propagate. Share the value itself, or a space-scoped ` +
-              `cell (keep a PerUser pointer to "mine"), or declare the ` +
-              `slot's schema with scope "${parsedLink.scope}" if per-reader ` +
-              `resolution is intended. See docs/development/debugging/` +
-              `gotchas/scoped-cell-pitfalls.md (pitfall 6).`,
+              `do not carry a principal, so every reader resolves it to ` +
+              `their own ${parsedLink.scope} instance. If this write meant ` +
+              `to SHARE data, it cannot propagate — share the value itself, ` +
+              `or a space-scoped cell (keep a PerUser pointer to "mine"), ` +
+              `or declare the slot's schema with scope ` +
+              `"${parsedLink.scope}" if per-reader resolution is intended. ` +
+              `See docs/development/debugging/gotchas/` +
+              `scoped-cell-pitfalls.md (pitfall 6).`,
+            ],
           );
         }
       }
--- a/packages/runner/src/data-updating.ts
+++ b/packages/runner/src/data-updating.ts
@@ -24,6 +24,7 @@ import {
   recordRelevantSchemaWritePolicyInput,
 } from "./cell.ts";
 import { resolveLink } from "./link-resolution.ts";
+import { schemaAcceptsType } from "./traverse.ts";
 import {
   areLinksSame,
   areMaybeLinkAndNormalizedLinkSame,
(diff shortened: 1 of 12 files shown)


packages/runner/src/data-updating.ts
```diff
@@ -809,6 +830,54 @@ export function normalizeAndDiff(
...
+      // per-reader semantics. This is a WARN, not a throw, because the
+      // runtime's own machinery legitimately writes scoped links into
+      // scope-silent slots today (pattern factory .asScope() result links,
+      // navigateTo result cells, Runner.updateArgument setup wiring,
+      // cold-resume re-scope walks — enumerated on #4561); until those slots
+      // declare their scope, author footguns and sanctioned machinery writes
+      // are indistinguishable here. Authors: share the value, or a
+      // space-scoped cell with a PerUser pointer to "mine" (pitfall #6 shows
+      // the idiom).
+      if (scopeRank(parsedLink.scope) > scopeRank(link.scope)) {
```
ubik2, line 899: I think that if the schema allows `undefined` or has a default, we should skip this warning.
It's just when the schema expects this field to exist that we should worry.

This would probably let you avoid the `withScopeIsolationWarnSuppressed` bit, too.

bfollington:
**Implemented @ubik2's criterion** (relayed by Ben): the warn now fires only when the slot's schema **would reject a missing value** — doesn't match `undefined` and has no `default` (`schemaToleratesMissing` in `data-updating.ts`, commit c9e6f40ce). Undefined-tolerant or defaulted slots stay silent: per-reader resolution there degrades harmlessly (it's exactly the `optlink` control fixture that converges with no fix needed), so a narrower-scoped link is a legitimate pattern rather than the footgun.

Details:
- `$defs`+`$ref` slot schemas (what CTS emits — the original repro's stored link had this shape) are resolved via the canonical `ContextualFlowControl.resolveSchemaRefs` before judging; unresolvable refs count as tolerant (lint-strength warn, err toward quiet). Unit pin keeps strict-ref slots warning.
- Known limitation, documented in the helper: the parent's `required` list isn't visible at this write site, so an optional-but-strictly-typed slot still warns. Threading the parent schema through the recursion is the fix if that proves noisy in practice.
- New unit pins: undefined-tolerant slot silent; defaulted slot silent; strict `$defs/$ref` slot warns. Scrabble (unit + multi-user) and convergence-storm 4/4 stay green; the scrabble catch still fires pre-fix (its `letter` slot is strict).

*(posted via Claude Fable, on behalf of @bfollington)*

bfollington:
## Confirmed against @seefeld's clarification: the declaration is a scope CAP — and the guard already implements exactly that

Berni (via Ben): *"what declaring it in the schema does is to cap the scope at the specific value, it can't be narrower. so you can't assign a PerSession value to a PerUser field, etc., but the other way around is allowed."*

Checked case-by-case against the implementation (unit pins in `data-updating.test.ts`):

| write | guard | matches cap semantics |
| --- | --- | --- |
| `user` link → `PerUser`-declared slot | silent | ✓ at the cap |
| `session` link → `PerUser`-declared slot | **warns** | ✓ narrower than cap ("can't") |
| `user` link → `PerSession`-declared slot | silent | ✓ broader than cap ("the other way around is allowed") |

No behavior change was needed — but the prose was wrong: the guard comment, pitfall 6, the spec paragraph, and the criterion table above described the declaration as an "opt-in to per-reader resolution." All reworded to cap semantics (commit 12eb0a680). This also resolves the external cold review's open confusion ("session-declared slot holding a user-scoped link is silent — intended?" — yes, per the cap). Conceptually the guard is now cleanly the **write-side mirror of the read-side follow-cap** (`canFollowScopedLink`: same-or-broader only).

**Sharpened follow-up for @seefeld:** given "can't", should a *declared-cap violation* (session link into a `PerUser` field) be a hard **error** on write rather than a warn? It's cleanly separable from the undeclared case:

- **Declared cap violated** → spec-backed contract breach; no runtime machinery write hits this (machinery slots are scope-silent); candidate for throw now.
- **Undeclared slot** → no cap exists; the warn is a sharing-intent heuristic (rejects `undefined` + no effective `default` + parent-`required`) and stays a lint.

If yes, that resolves half of open question 2 (path to error) immediately — happy to implement the split.

*(posted via Claude Fable, on behalf of @bfollington)*

bfollington:
Cross-ref: #4668 extends the B2 element grace to generation-skew signatures, widening the gap this guard's caveat already documents — and strengthening the case for the warn: with array-element reads degrading quietly under both scope holes and schema evolution, this write-site lint is the only place a narrower-scoped link in shared data announces itself (details on #4668). Criterion prose updated to cite it, and the branch is rebased onto current main. Landing warn-only; the declared-cap→error split stays a follow-up on the existing flip checklist.

*(posted via Claude Fable on behalf of @bfollington, after discussion with him)*

PR #4565, opened 2026-07-07
fix(state-inspector): make discovery tests hermetic against real cf-inspect cache

bfollington:
## Problem

`packages/state-inspector/test/discover.test.ts` ("discovers both DBs by walking the cache base") fails on any machine where `~/.cache/cf-inspect` contains real pulled DBs (e.g. from `cf inspect --remote` usage).

The test already pointed discovery at a seeded temp dir via `dirs`, but `discoverSpaceDbs` unconditionally **appends** `candidateRoots()` — which includes the remote-pull cache (`~/.cache/cf-inspect`), `MEMORY_DIR`/`DB_PATH` env overrides, and an upward cwd walk. So the "exactly 2 discovered DBs" assertion counted whatever real cache contents exist on the machine.

## Fix

- `discoverSpaceDbs` gains a `defaultRoots?: boolean` option. `defaultRoots: false` searches **only** the explicit `dirs`, skipping env overrides, the remote-pull cache, and the cwd walk. The default (`true`) is unchanged, so CLI behavior and `resolveSpace`/`resolveSpacePath` defaults are untouched.
- `discover.test.ts`: all four `discoverSpaceDbs` calls pass `defaultRoots: false` (the now-unused `cwd: dir` dropped).
- `remote.test.ts`: same option on its one discovery call — its assertions were tolerant of extras, but it had the same latent leak.

## Verification

- `deno task test` in `packages/state-inspector`: 31 passed (71 steps), 0 failed — on a machine whose `~/.cache/cf-inspect` is populated with real host dirs.
- Reproduced the failure mode deliberately: with a poisoned default root (`MEMORY_DIR` → dir containing an extra `.sqlite`), the old test fails exactly as reported and the fixed test passes.

---
## Summary by cubic
Make `state-inspector` discovery tests hermetic by letting `discoverSpaceDbs` skip default roots and pointing tests at temp dirs. Also adds hermetic coverage for `candidateRoots` and the default-roots discovery path.

- **Bug Fixes**
  - Added `defaultRoots?: boolean` to `discoverSpaceDbs`; when false, search only `dirs` and skip env overrides, the remote-pull cache, and the cwd walk. Default stays true; CLI and `resolveSpace*` defaults unchanged.
  - Updated tests to pass `defaultRoots: false` and removed the unused `cwd`; added a superset-asserting default-roots discovery step and a `candidateRoots` unit test (env override ordering, cwd walk, bare relative `DB_PATH`).

<sup>Written for commit 221b4cd5ed4f4469b99fed3ce52fa391884e2e26. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4565?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/state-inspector/discover.ts
+++ b/packages/state-inspector/discover.ts
@@ -84,11 +84,19 @@ export function candidateRoots(cwd: string = Deno.cwd()): string[] {
   return roots;
 }
 
-/** Discover local space DBs. `dirs` are searched before the default roots. */
+/**
+ * Discover local space DBs. `dirs` are searched before the default roots.
+ * `defaultRoots: false` searches ONLY `dirs`, skipping env overrides, the
+ * remote-pull cache, and the cwd walk — tests need this to stay hermetic on
+ * machines whose real `~/.cache/cf-inspect` holds pulled DBs.
+ */
 export function discoverSpaceDbs(
-  opts: { dirs?: string[]; cwd?: string } = {},
+  opts: { dirs?: string[]; cwd?: string; defaultRoots?: boolean } = {},
 ): DiscoveredSpace[] {
-  const roots = [...(opts.dirs ?? []), ...candidateRoots(opts.cwd)];
+  const roots = [
+    ...(opts.dirs ?? []),
+    ...(opts.defaultRoots === false ? [] : candidateRoots(opts.cwd)),
+  ];
   const seen = new Set<string>();
   const out: DiscoveredSpace[] = [];
   for (const root of roots) {
--- a/packages/state-inspector/test/discover.test.ts
+++ b/packages/state-inspector/test/discover.test.ts
@@ -6,6 +6,7 @@ import { assert, assertEquals, assertThrows } from "@std/assert";
 import { Database } from "@db/sqlite";
 
 import {
+  candidateRoots,
   deriveSpaceDid,
   discoverSpaceDbs,
   quickStats,
@@ -70,6 +71,15 @@ Deno.test("discovery + resolution", async (t) => {
       assert(found.every((s) => s.sizeBytes > 0));
     });
 
+    await t.step("default roots find the DBs via the cwd walk", () => {
+      // No explicit dirs: the cwd walk must reach `<cwd>/cache/memory`. On a
+      // real machine the default roots may surface OTHER DBs too (env
+      // overrides, ~/.cache/cf-inspect), so assert superset, never counts.
+      const found = discoverSpaceDbs({ cwd: dir });
+      const dids = new Set(found.map((s) => s.did));
+      assert(dids.has(DID_1) && dids.has(DID_2));
+    });
+
     await t.step("resolveSpacePath matches by DID prefix", () => {
       const found = discoverSpaceDbs({
         dirs: [`${dir}/cache/memory`],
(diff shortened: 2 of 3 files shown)


bfollington (commented):
cf-review via Claude Fable 5 (`claude-fable-5`), on behalf of @bfollington

**Scope:** 3 files, +16/−8 in `packages/state-inspector` · **Coverage:** deep-read (all 3 changed files, full `discover.ts`, both test files) · **Checks run:** package `deno task test` (31 passed / 71 steps, 0 failed — on a machine with 12 real DBs in `~/.cache/cf-inspect`), re-break check (checked out the pre-PR `discover.test.ts` on this machine → fails exactly as reported; PR version passes), `deno fmt --check` + `deno lint` clean.

### 🔴 Blocking

None.

### 🟡 Improvements

1. **`candidateRoots()` now has zero test coverage** — `packages/state-inspector/discover.ts:61`. With `defaultRoots: false` at every test call site, nothing exercises the default-root path: `MEMORY_DIR`/`DB_PATH` env overrides (including the bare-filename `dirname || "."` fallback), the `rootCacheDir()` remote-pull root, and the bounded cwd walk — exactly the path the CLI (`packages/cli/commands/inspect.ts`, 5 call sites) relies on. Coverage was only incidental before (the cwd walk's roots deduped against `dirs`, so no assertion ever distinguished them), so this is a gap made visible rather than introduced. `candidateRoots` is env+cwd → string list, so a hermetic unit test asserting root presence/ordering is cheap. Follow-up, not a gate. *(verified — grep shows no other references outside `discover.ts`)*

### ⚪ Nits

2. `test/remote.test.ts:91-92` — "normal local discovery" now slightly overstates: the call deliberately skips default roots. The guarded principle (a pulled DB is discoverable and DID-resolvable) is intact; "explicit-dir discovery" would keep it precise. *(verified)*
3. `discoverSpaceDbs` JSDoc could note `cwd` is inert when `defaultRoots: false` (which is why dropping `cwd: dir` in the tests was correct). *(verified)*

### Test rigor

Each touched test still guards its original principle (walk-based discovery under a cache base, DID-prefix resolution, ambiguity error, name-derivation resolution, pulled-DB discoverability). Assertion strength is unchanged; only non-hermetic input was removed. Production default confirmed unchanged — `defaultRoots === false` is a strict opt-out; all CLI/`resolveSpace*` call sites pass nothing.

### Verdict

**Clean; safe to merge.** Minimal opt-out flag, default behavior untouched, fix verified against the real failure mode on this machine plus a successful re-break check. One follow-up worth filing: a hermetic `candidateRoots` unit test.

bfollington:
Addressed the review's 🟡 improvement finding in 221b4cd5e: added a hermetic `candidateRoots` unit test (env-override ordering, cwd walk, bare-relative `DB_PATH`) plus a superset-asserting default-roots discovery step, restoring coverage of the default discovery path that the hermeticity opt-out had removed. This should also clear the coverage-debt ratchet that failed the first Performance Check run (+1 uncovered line).

PR #4574, opened 2026-07-07
fix(runner,ts-transformers): tolerate pre-#4158 legacy-envelope stored sources on cold load (CT-1838)

jkomoros:
## Why

Pre-#4158 pipelines stored the helper-injected pretransform form as the source of record: the file begins with `HELPERS_STMT` and ends with the matching TypeScript or JavaScript helper-use trailer. Current authoring correctly reserves `__cfHelpers`, but the same guard also rejects those historical stored bytes when a runtimeVersion change forces a cold recompile. A legacy default pattern can therefore prevent a space from loading its piece registry and from creating pieces.

The stored bytes are still valid content-addressed data: they are the bytes their identities were computed over. Re-identifying and repointing every reference would require a risky data migration. This change instead tolerates the exact historical envelope only at trusted stored-input boundaries and writes the resulting compiled set under the current runtimeVersion.

## What changed

- Add and export `isLegacyInjectedEnvelope(source)`, calibrated against a byte-exact production corpus sample. It recognizes only the exact historical prefix/trailer pair (TypeScript or JavaScript trailer; final newline optional).
- Add an explicit `tolerateStoredLegacyEnvelope` option to helper injection.
- Enable that option when cold-recompiling a source closure that has already passed `loadVerifiedSourceClosure`.
- Enable it for fabric mount sources, which are storage-fetched rather than author-typed, covering both cold recovery and a new pattern importing a legacy pattern.
- Leave `checkCFHelperVar` and ordinary authoring paths unchanged. New source containing the reserved helper symbol still fails.
- Preserve the stored bytes and identities; a successful cold compile self-heals by populating the current compiled cache.

## Safety boundary

This is stored-input tolerance, not a general relaxation:

- cold source closures are Merkle-verified before reaching the tolerant path;
- mount sources come from the storage-backed fabric resolver;
- the detector requires the exact historical envelope;
- identity is recomputed over the original stored bytes and must still equal the requested identity;
- authored source continues through the existing reserved-symbol guard.

The negative-result memo formerly included in this branch has been removed from this PR. Its original verification-failure classification was unsafe for partially synchronized closures. A corrected, independently reviewable implementation is stacked as #4716; it never memoizes source absence or verification failure.

## Tests

Focused local validation after rebasing onto current `main`:

- `packages/ts-transformers/test/core/legacy-envelope.test.ts`: 12 passed;
- `packages/runner/test/load-by-identity.test.ts`: 12 steps passed, including cold heal, identity preservation, warm write-back, mixed/replicated closures, JS trailer, authoring guard, and new-pattern → legacy-pattern fabric import;
- `packages/piece/test/legacy-envelope-default-pattern.test.ts`: end-to-end `getPieces`/`add` passed over a legacy default pattern after a simulated pin bump;
- formatting and lint checks pass on the touched runner files.

The force-pushed core change passed GitHub Actions. A fresh Deno Workflow has been triggered for the corpus-path follow-up.

## Branch update

Taken over while @jkomoros is away:

- rebased onto `main` at `6d3e005fe`;
- removed the negative memo and all memo-only APIs/tests;
- preserved the pre-surgery head locally as `backup/legacy-envelope-tolerance-pre-decouple`;
- force-pushed with `--force-with-lease`.

This unblocks the labs vendor bump needed by commontoolsinc/loom#4046 without coupling that compatibility fix to the follow-on retry optimization.


mathpirate:
Cubic P2 (forgeable `Symbol.for` marker): valid — fixed in 53def5cb4. The marker is now a module-private `Symbol`, so only `markDeterministicCompileFailure` can produce a positive; the deliberate trade-off is documented in-code (across duplicated module graphs a missed marker just retries — the memo's fail-open direction — while forgery would have memoized transient failures for the session). The new `compile-failure.test.ts` pins the contract, including a forgery test with a same-named registry symbol that fails against the previous implementation. Same commit covers the coverage ratchet's +1.

PR #4580, opened 2026-07-07
fix(tasks): merge per-instance lcov records in the coverage-debt metric (CT-1861)

mathpirate:
Deno's coverage emits one record per module **instance** — the same file appears plain and as `?testRun=<uuid>` once per importing test file — and `parseLcov` keyed records on the raw `SF:` path. The same physical line therefore counted as debt once per instance that didn't execute it, which made the coverage-debt ratchet structurally noisy:

- the metric **flapped ±1–5 lines** run-to-run on unchanged code (baselines carry n=2 samples after main moves, so one flap trips the gate);
- changes that move execution timing (e.g. eager→lazy deferral) **reshuffled the count with zero real coverage change** — verified from CI's own lcov artifacts, line-shift-adjusted against base runs, on both #4557 and #4560;
- **adding a test file could *increase* "uncovered lines"** — its fresh instances carry red lines for whatever they don't execute (observed +3 from adding a test whose target module measured 100% covered).

CT-1861 carries the full three-PR evidence trail from 2026-07-06/07.

**Fix:** strip the query suffix before keying, so instances of the same physical file merge and a line counts covered when **any** instance executed it — the metric measures real per-file coverage. Unit test pins the merge semantics (three instances, hits summed, only the line no instance executed counts uncovered).

**Transition:** no flag day. This PR's own gate computes merged (lower) values against old-style baselines — large apparent reductions — and post-merge main runs re-baseline downward once; subsequent PRs compare consistently.

---
## Summary by cubic
Fixes coverage-debt noise by merging Deno’s per-instance LCOV records into a single per-file profile, preventing flapping and false increases when adding tests. Addresses CT-1861.

- **Bug Fixes**
  - Normalize instance suffixes at LCOV generation via `normalizeLcovInstancePaths` in `write-coverage-lcov.ts`, applied in both the CI artifact writer and local `lcovFromCoverageProfile`; all consumers (e.g. IDE report) now see one record per physical file, and `parseLcov` simply accumulates duplicate `SF:` sections.
  - Add a unit test that exercises normalize → parse and pins uncovered-line calculation.

- **Migration**
  - No action needed. The gate already compares merged values to old baselines; main will re-baseline lower once after merge.

<sup>Written for commit d2b74a693034e7c9ed6c6591a7576e8ec4fdeb87. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4580?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tasks/coverage-metrics.test.ts
+++ b/tasks/coverage-metrics.test.ts
@@ -28,6 +28,38 @@ Deno.test("parseLcov accumulates hits per source line", () => {
   assertEquals(countUncoveredProfileLines(file!), 1);
 });
 
+Deno.test("parseLcov merges per-instance records of the same physical file", () => {
+  // Deno's coverage emits one record per module INSTANCE — the same file
+  // appears plain and as `?testRun=<uuid>` once per importing test file.
+  // Instances must merge so a line counts covered when ANY instance executed
+  // it; otherwise the debt metric counts the same physical line once per
+  // instance that skipped it, flapping with test order and punishing added
+  // tests (CT-1861).
+  const coverage = parseLcov([
+    "SF:/repo/packages/example/src/mod.ts",
+    "DA:1,1",
+    "DA:2,0",
+    "end_of_record",
+    "SF:/repo/packages/example/src/mod.ts?testRun=aaaa-1111",
+    "DA:1,0",
+    "DA:2,2",
+    "DA:3,0",
+    "end_of_record",
+    "SF:/repo/packages/example/src/mod.ts?testRun=bbbb-2222",
+    "DA:3,0",
+    "end_of_record",
+  ].join("\n"));
+
+  assertEquals(coverage.size, 1);
+  const file = coverage.get("/repo/packages/example/src/mod.ts");
+  // Line 1 covered by the plain instance, line 2 by testRun aaaa; line 3 by
+  // neither — the ONLY genuinely uncovered line.
+  assertEquals(file?.lineHits.get(1), 1);
+  assertEquals(file?.lineHits.get(2), 2);
+  assertEquals(file?.lineHits.get(3), 0);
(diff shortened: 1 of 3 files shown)


Hixie:
Where do we add that suffix? Can we remove it there instead? That way other consumers of the LCOV files (e.g. the task that merges them all for IDE support, see https://github.com/commontoolsinc/labs/blob/main/docs/development/COVERAGE.md#a-combined-report-for-ides) don't have to know about this as well.

mathpirate:
Good call — reshaped in d2b74a69.

**Where the suffix comes from:** it's deliberate, not incidental — two test helpers import modules with a cache-busting query to get fresh module-scoped state per test (`importFreshCellCacheModule` in `packages/runner/test/compiler-fingerprint.test.ts`, `importFreshCompileByteCacheModule` in `packages/cli/test/test-runner-compile-byte-cache.test.ts`). V8 then reports each such import as a separate script, and `deno coverage --lcov` emits a record per instance under the suffixed path. So it can't be removed at the import site without losing the isolation those tests exist for.

**But moved the normalization to generation per your point:** `normalizeLcovInstancePaths` now lives in `write-coverage-lcov.ts` and runs at both places we produce LCOV — the CI artifact writer and the metrics tool's local `lcovFromCoverageProfile` path. `parseLcov` is reverted to suffix-ignorant keying (duplicate `SF:` sections already accumulate), and the combined IDE report + any future consumer inherit clean paths with nothing to know.

Hixie:
LGTM

PR #4586, opened 2026-07-07
feat(ci): retro-classify unstamped runs + fingerprint fallback for missing cache-state artifacts

mathpirate:
Two additions on top of #4579 (now merged), closing the baseline-skew gap its
docs flagged. Built independently as a twin of #4579's base mechanism; once
#4579 landed, this rebased down to exactly these two additions — the shared
base (file-presence cold detection, the override-ordering / poison-on-unreadable
fix) is #4579's now and was dropped here as redundant. Ground truth beats
inference: a recorded cache state always wins; the fingerprint inference below
only fills where a recorded state is absent.

- **Retro-classification of unstamped baseline runs**: runs without a recorded
  state (pre-rollout artifacts, backfills, live extraction) are classified
  from the compile fingerprint — if the cc-* key paths changed vs the run's
  predecessor (compare API), every family is cold. This de-pollutes the
  existing 20-run window immediately (including the #4505/#4543-class cold
  samples) instead of waiting for stamped history to accumulate; once it does,
  the inference goes unused.
- **Fingerprint fallback for the current run**: families with no recorded
  state (artifact missing/failed) fill from the same inference, cold-only —
  recorded states win, and a warm fill would be a no-op.

`tasks/perf-cache-state.ts` mirrors the workflow's cc-* key globs with a
drift-guard test that parses deno.yml. Inference cannot see eviction/outage
colds — those stay unknown and behave exactly as today (fail-open).

Full tasks suite: 239 passed, 0 failed.

---
## Summary by cubic
Retro-classifies unstamped baseline runs and adds a compile-fingerprint fallback when cache-state artifacts are missing. This cleans the 20-run baseline window and avoids false gates when the compile cache rotates.

- New Features
  - Baselines: For runs without a recorded state, compare changed files vs predecessor using the `cc-*` key globs; if any changed, mark all families cold; else warm.
  - Current run: Unknown families fall back to the same fingerprint inference (cold-only). Recorded states win; non-fingerprint causes (eviction/outage) stay unknown and gate normally.
  - Added `tasks/perf-cache-state.ts` and tests: mirrors `deno.yml` `cc-*` key globs with a drift-guard test; exposes `classifyRunAgainstPredecessor`, `classifyCacheKeyState`, `changedPathsOf`, and `uniformCacheStates`.
  - Docs: Updated CI performance guide to cover retro-classification and the fallback.

- Bug Fixes
  - Renames out of the fingerprint set are now classified cold by considering both `filename` and `previous_filename` from PR and compare APIs.

<sup>Written for commit 58a2cc1595d411baa9bdc69a0adb53083aacbb2f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4586?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/development/CI_PERFORMANCE.md
+++ b/docs/development/CI_PERFORMANCE.md
@@ -254,12 +254,18 @@ The comparison rules follow from the tagging:
 - The coverage-debt ratchet uses the latest non-cold main sample, so a cold
   main run cannot lower the baseline that warm PRs are held to.
 
-Runs with unknown cache state — anything before this mechanism rolled out, and
-backfill-derived runs — behave exactly as before: their samples are kept and
-their metrics gate normally.
+Runs without a recorded state — anything before this mechanism rolled out,
+backfill-derived runs, and runs whose cache-state artifacts failed — are
+retro-classified from the compile fingerprint (`tasks/perf-cache-state.ts`
+mirrors the `cc-*` key globs, drift-guarded by a test that parses the
+workflow): if the fingerprint paths changed against the run's predecessor,
+every family is treated as cold; if unchanged, warm. The same fingerprint
+inference backstops the current run when its cache-state artifacts are
+missing. Fingerprint inference cannot see non-fingerprint cold causes (cache
+eviction, cache-service outages) — runs cold for those reasons and lacking a
+recorded state stay unknown: kept and gated normally, exactly as before this
+mechanism.
 
-Two gaps remain. Cold compile duration itself is not gated, so a regression
+One gap remains: cold compile duration itself is not gated, so a regression
 that only shows up on a cold cache passes unnoticed; a dedicated cold-compile
-bench could cover that later. And cold main runs from before the rollout stay
-unknown, so they can still skew baselines until they age out of the 20-run
-baseline window.
+bench could cover that later.
--- a/docs/development/CI_PERFORMANCE.md
+++ b/docs/development/CI_PERFORMANCE.md
@@ -254,18 +254,16 @@ The comparison rules follow from the tagging:
 - The coverage-debt ratchet uses the latest non-cold main sample, so a cold
   main run cannot lower the baseline that warm PRs are held to.
 
-Runs without a recorded state — anything before this mechanism rolled out,
-backfill-derived runs, and runs whose cache-state artifacts failed — are
+A run without a recorded cache state — an artifact carrying no stamp, a
+backfill-derived run, or a run whose cache-state artifact failed to upload — is
 retro-classified from the compile fingerprint (`tasks/perf-cache-state.ts`
-mirrors the `cc-*` key globs, drift-guarded by a test that parses the
-workflow): if the fingerprint paths changed against the run's predecessor,
-every family is treated as cold; if unchanged, warm. The same fingerprint
-inference backstops the current run when its cache-state artifacts are
-missing. Fingerprint inference cannot see non-fingerprint cold causes (cache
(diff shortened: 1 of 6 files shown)


tasks/perf-cache-state.ts
```diff
@@ -0,0 +1,148 @@
...
+export async function fetchChangedFilesBetween(
+  baseSha: string,
+  headSha: string,
+): Promise<string[]> {
+  const data = await githubGet<{ files?: { filename: string }[] }>(
+    `/repos/${REPO}/compare/${encodeURIComponent(baseSha)}...${
+      encodeURIComponent(headSha)
+    }`,
+  );
+  return (data.files ?? []).map((file) => file.filename);
```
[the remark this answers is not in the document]
mathpirate: Valid — fixed in d207cbdaa. Both sides of a rename are now surfaced (compare API and the PR-files fallback path alike), with a test pinning the rename-out-of-set → cold case. Worth noting the miss direction was false-warm (a genuinely cold run kept gating normally — i.e., today's behavior), so this was precision loss rather than an unsafe waiver — but strictly worth having. — Claude Fable 5, on behalf of @mathpirate

docs/development/CI_PERFORMANCE.md
```diff
@@ -153,12 +153,18 @@ The comparison rules follow from the tagging:
...
+Runs without a recorded state — anything before this mechanism rolled out,
+backfill-derived runs, and runs whose cache-state artifacts failed — are
+retro-classified from the compile fingerprint (`tasks/perf-cache-state.ts`
+mirrors the `cc-*` key globs, drift-guarded by a test that parses the
+workflow): if the fingerprint paths changed against the run's predecessor,
+every family is treated as cold; if unchanged, warm. The same fingerprint
+inference backstops the current run when its cache-state artifacts are
+missing. Fingerprint inference cannot see non-fingerprint cold causes (cache
+eviction, cache-service outages) — runs cold for those reasons and lacking a
+recorded state stay unknown: kept and gated normally, exactly as before this
```
Hixie: i would guide your agent to rewrite the documentation in a timeless manner, describing current state and optionally future intended state but not making any reference to paste state

Hixie:
If I understand correctly this now splits our performance baseline data into two paths, the warm and the cold, and always compares like with like when doing performance comparisons?

Does it handle the case where some jobs were warm and some were cold?

docs/development/CI_PERFORMANCE.md
mathpirate: ty, done

mathpirate:
Yes — state is tracked **per cache-family** (`generated-patterns` / `pattern-integration` / `pattern-unit`), not as one run-level flag, so mixed runs are handled by construction: each timing metric gates only against baseline runs that were warm for *its* family (cold baseline samples dropped), and a metric whose family is cold this run is marked `COLD` and skipped rather than gated. So `pattern-unit` cold + `pattern-integration` warm in the same run each does the right thing independently. (Within a family, any cold shard makes the whole family cold — conservative, since one recompiling shard inflates the aggregate.)

— Claude Opus 4.8, on behalf of @mathpirate

Hixie:
Awesome.

ubik2 (approved):
lgtm!

PR #4608, opened 2026-07-08
test(runner): pin no-resurrection of dropped multi-doc materializations

ubik2:
> [!NOTE]
> **Draft — blocked on #4405.** This test pins the spine-rebuild replay semantics that PR introduces; on current main it fails via the old branch-replace path (the dropped commit's data resurrects). Marking ready + rebasing onto main once #4405 merges. Kept separate from #4405 per review discussion.

## What this pins

CT-1872 Class 1a: the multi-doc rollback path where the client's mirror is fully **current** and replay still has work to do. A transaction materializes doc A's container (with its own seed data) and writes doc B atomically; it conflicts on **B alone**, so the whole transaction rolls back — including A's container. Nobody else ever wrote A, so no catch-up update for A is owed: this window is reachable with the update/rejection ordering contract fully intact (the read-repair gate has nothing to deliver).

A surviving read-free patch under that container (the mergeable/vote shape — exempt from dependency cascading by design, and durably acceptable server-side once a winner's container exists) must then replay over `{items: null}`. The assertion: only the survivor's own write lands —

```
expectVisible(harness, { A: { items: { mine: "hello" } } });
```

— the dropped commit's `seeded: "x"` must NOT resurrect out of the survivor's optimistic snapshot. On main, the old `pendingSetPathForBase` branch-replace restores it (`{ items: { mine: "hello", seeded: "x" } }`), which is the vote-count-mismatch class of bug; with #4405's `baseWithMutableReplaySpine`, only the leaf lands.

The test also models the server-visibility half: the winning container is applied server-side via `injectRemote` with the client's mirror left stale, so the accepted survivor is promoted over `{items: null}` locally — pinning that promotion also doesn't resurrect.

---
## Summary by cubic
Fixes pending patch replay so surviving writes apply over stale mirrors without restoring dropped data, and adds high‑contention Lunch Poll tests to prove votes and swatches stay consistent. Pins CT-1872 (class 1a) by ensuring dropped materializations do not resurrect during replay or promotion.

- **Bug Fixes**
  - Rebuild a mutable replay spine for pending patches instead of branch-replacing, so surviving writes land over stale bases without reviving dropped data.
  - Add `cloneWithPendingReplayValueAtPath` with strict array-key checks; remove the old branch-replace path and related logging.
  - Cover replay over stale mirrors and server-side promotion; no-resurrection verified across multi-doc rollbacks.

- **Tests and CI**
  - Add memory tests pinning no-resurrection and edge cases for replay spine rebuilds.
  - Add a high-contention Lunch Poll integration test with a fresh observer; assert per-option swatches match stored votes; extend the existing two-user vote test.
  - Add stable data hooks to Lunch Poll UI (`data-option-id`, `data-all-option-*`, `data-vote-swatch-*`) for reliable assertions.
  - Add a dedicated CI job `pattern-lunch-poll-contention-test`; exclude it from standard shards; update coverage docs and perf tooling to track the new job.

<sup>Written for commit 86bc8400fd77d401a13307062a76b8b2d29ee7d0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4608?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ubik2:
Reachability caveat after re-reading this regression against the conflict read-repair gate:

The current test calls `injectRemote` to put `{ items: { theirs: "w" } }` on the scripted server, deliberately delivers no subscription update to the client, and returns a synthetic `ConflictError` without the real `retryAfterSeq` / `readyToRetry` repair obligation. That means C1 is finalized and dropped while the client's confirmed mirror is still `{ items: null }`.

In production, receiving the rejection may indeed precede batched fan-out, but `finalizeRejection` is supposed to hold `dropPending(C1)` until the catch-up obligation is satisfied. If that mechanism works and A is subscribed, the client should have `theirs` before C1 is actually dropped. C2's `/items/mine` write then replays as an ordinary leaf update over an existing container; the blocked-branch replacement does not fire, so this exact setup does not exercise a gate-independent Class 1a failure.

As written, the test therefore demonstrates no-resurrection under a **stale-mirror/read-repair leak** (Class 2/5, or a repair bypass), rather than isolating Class 1a itself.

A gate-independent Class 1a regression would be stronger if it used the true “no update owed” shape:

1. A's confirmed local/server state is legitimately absent or non-container.
2. C1 locally materializes A and also writes B.
3. B conflicts, so the entire C1 transaction drops, including A's optimistic materialization.
4. C2 is an actual zero-read mergeable operation that is valid server-side via `createMissing` and therefore intentionally survives C1.
5. No remote A novelty exists for read repair to deliver.
6. Assert that replay preserves only C2's operation and never copies C1's seed data.

Alternatively, the current test/title should explicitly classify this as the stale-mirror + replay composition rather than pure Class 1a.

ubik2:
Closing as superseded. The dependency this was blocked on (#4405) closed unmerged, and the behavior this test pins — CT-1872 class 1a, no resurrection of a dropped multi-doc materialization's data — landed on main via #5521 (2026-08-08) with its own suite in the same file, including the identically-named headline test. The remaining tests here pin #4405's rebuild-the-spine semantics, which main deliberately resolved the other way (#5521 skips inapplicable surviving layers instead of rebuilding), so a rebase would assert the wrong behavior. CT-1872 itself is closed.

PR #4616, opened 2026-07-08
feat(ts-transformers): preserve authored lineage on the builder-call path (CT-1868)

mathpirate:
Builder calls and callbacks now arrive at `BuilderCallHoistingTransformer` (the hoisting stage — canonical ordinal 17 per the stage registry) with a recoverable **authored** source position. Before this, every hoisted builder origin arrived bare (`pos=-1`, no sourceMapRange, no original chain) — probe-verified on main@39afdb62b — so transform-time source annotation (A′, CT-1870) had nothing authored to read.

## What

- **`preserveLineage(node, origin)`** (full triple: textRange + sourceMapRange + original) and **`preserveSourceMapRange(node, origin)`** (smr-only) in `ast/utils.ts`, with doc comments carrying the channel rules.
- Applied at the **builder-path-complete** site set (~24 sites): `PatternBuilder.buildCallback`/`.buildHandlerCallback` (chokepoint for all closure-strategy callbacks), the strategy outer calls, the expression-rewrite + builtins wrapper arrows/outers, the `mapWithPattern` rebuild, capture-scaffold's applied handler call, and SchemaInjection's 17 builder-call rebuilds (the immediate predecessor of hoisting, which re-stripped everything earlier stages preserved).
- **Regression test** `test/lineage-regression.test.ts`: drives the real pipeline over a multi-origin fixture and asserts per-{tag, role} that every builder call AND callback recovers to its OWN origin's snippet (content bound per hoist, claimed exactly once), plus rewritten post-hoist SITE recovery bound to the same origin. Bite-verified: reverting `pattern-builder.ts`, `capture-scaffold.ts`, `array-method-transform.ts` or `schema-injection.ts` individually makes it fail.
- The **A′ master handoff doc** (`APRIME-LINEAGE-HANDOFF.md`) rides this PR, and the current-behavior spec gains **§11.5** (lineage at the hoisting boundary) with pointer lines in §9/§10 per its §21 change rule.

## Design note — sourceMapRange is the lineage carrier

The original plan (full triple everywhere; `create→update*` in SchemaInjection) was revised by the emit-invariance gate: `setTextRange` reaches the printer (arrow-head parenthesization, JSX/ternary reflow) and `setOriginalNode` reaches semantics (schema derivation via `getTypeAtLocationWithFallback`, typeRegistry resurfacing, the compute-owned marker registry). sourceMapRange is observable by nothing but sourcemaps and rides `update*` rebuilds via `setOriginalNode`'s emitNode merge (verified against TS 5.9.2 and 6.0.3 source). Full `preserveLineage` survives only at the two gate-proven-inert sites. Details: doc §6a and the helper doc comments.

## Verification

- Acceptance probe (doc §3): all `hoist:*`/`authored:*` lines — outer, inner, callback — recover the correct authored snippet.
- **Emit invariance: byte-identical** `--show-transformed` output on `age-category.tsx` and `self-improving-classifier.tsx`, re-verified against post-rebase main (lineage is emit-map metadata, not output text).
- Full ts-transformers suite: **1104 passed / 0 failed** (post-rebase count); root `deno task check` clean; fmt/lint clean; runner `pattern-coverage` end-to-end pin green.
- Independently re-verified by a second pass (suite, root check, emit diffs, test bite matrix), and externally reviewed (review incorporated in the final commits).

Linear: CT-1868 (this), CT-1869 (broad-sweep follow-up), CT-1870 (A′ proper, unblocked by this).


packages/ts-transformers/test/lineage-regression.test.ts
```diff
@@ -0,0 +1,270 @@
...
+      );
+      recoveredSnippets.push(
+        sourceText.slice(callbackPos.pos, callbackPos.end),
+      );
+    }
+
+    // Content is the ground truth: the recovered positions must collectively
+    // cover every authored origin's distinctive snippet. A `pos >= 0` pointing
+    // at the wrong text would pass the per-site checks above but fail here.
+    const recoveredText = recoveredSnippets.join("\n");
```
mathpirate: The content check is global rather than per site, so it does not establish the claimed mapping from each builder to its own authored origin. In particular, the collected `export-default` call/callback spans the entire pattern body and already contains the A/N/B/D markers; wrong or permuted hoist anchors can therefore pass as long as every node has *some* valid position.

The collector also records only top-level hoisted inner calls, not the outer applied/enclosing calls whose lineage N2/N5 preserve. I confirmed diagnostically that changing the N5 outer anchor leaves this test green.

Could this assert expected content per `{tag, role}` and also inspect the rewritten outer `__cfLift_N(...)` / `__cfHandler_N(...)` / `*WithPattern(__cfPattern_N, ...)` sites (or inspect the AST immediately before hoisting)?

packages/ts-transformers/test/lineage-regression.test.ts
```diff
@@ -0,0 +1,270 @@
...
+  "CT-1868: builder lineage recovers authored positions at the hoisting stage",
+  async () => {
+    const { transformed, original } = await transformFixtureToAst();
+    const sourceText = original.text;
+
+    const sites = collectBuilderSites(transformed);
+
+    // Sanity: the fixture must actually exercise every origin path, or a silent
+    // pipeline change (e.g. a builder no longer hoisting) would vacuously pass.
+    const tags = sites.map((s) => s.tag);
```
mathpirate, line 302: This sanity set never requires a `__cfHandler_N` hoist. The `bump` fixture is an authored module-scope handler, so its callback bypasses `PatternBuilder.buildHandlerCallback` and `buildCapturedHandlerClosureCall`; the N1 handler branch and N6 are therefore unguarded. Removing both N3 `rewrite-helpers.ts` source-map preservations also leaves this regression green.

Please add a captured inline handler/action plus a fixture that bites the direct N3 wrapper route. I tried `action(() => count * 4)` locally: it produces `__cfHandler_1`, and both its call and callback recover to the correct authored action, so the implementation appears correct and the missing handler case is straightforward to pin.

packages/ts-transformers/APRIME-LINEAGE-HANDOFF.md
```diff
@@ -0,0 +1,583 @@
...
+`preserveLineage` survives only at N2/N5 (gate-proven inert — N2's channels are
+consumed and re-stripped by SchemaInjection before printing; N5's textRange is
+why the mapWithPattern outer recovers `self→`), smr-carry everywhere else.
+Acceptance verified per §4 (all 16 probe lines recover correct authored
+snippets) and guarded by `test/lineage-regression.test.ts` (bite-verified:
+reverting schema-injection.ts makes it fail).
+
+Notes:
+
+- N6/N7's `update*` switch also future-proofs against the smr-fragility class:
```
mathpirate: This is stale and contradicts the shipped channel rule: N6/N7 did **not** switch to `update*`; they deliberately remain `create* + preserveSourceMapRange`, precisely so they do not acquire `textRange` or `original`.

The same pre-implementation model remains in §6's wrapper-arrow “reification” example, §10's `create→update` ticket summary, and the regression-test header. Since this document declares itself the canonical A′ handoff, please reconcile those references to the actual create+SMR design (and keep wrappers explicitly out of full `preserveLineage`).

packages/ts-transformers/APRIME-LINEAGE-HANDOFF.md
```diff
@@ -0,0 +1,583 @@
...
+| N7               | `transformers/schema-injection.ts` builder-call rebuild family: `:1489`, `:2379–:2410` (self-verified) + `:2728`, `:3190`, `:3289`, `:3520`, `:3729`, `:3818`, `:3895`, `:3985`, `:4080` and cluster `:985`, `:3782` (sweep-enumerated; `:2728`/`:3190` spot-verified — re-verify each at implementation) | every builder call SchemaInjection rebuilds to splice/prepend/append schemas — the immediate predecessor of the hoisting stage, so this family alone strips whatever survives stages 8–12 | `node` / `innerLiftCall` | **`create*` + `preserveSourceMapRange`** — NOT the originally-planned `update*`: its "free" textRange/original are precisely the observable channels (kitchensink JSX reflow; typeRegistry resurfacing via `getTypeAtLocationWithFallback` made module-scope-cf-data wrap a hoisted lift). Shipped as 16 smr-carry sites; anchors: inner rebuilds → `innerLiftCall`, outers/direct → `node`, asScope callee → `node.expression` |
```
mathpirate: The count/classification here does not match the source. `schema-injection.ts` currently has 17 `preserveSourceMapRange(...)` call sites, and several are reactive-conditional, cell-factory/`.asSchema`, `wish`, `generateObject`, SQLite, and fetch rebuilds rather than one of BuilderCallHoisting's three hoistable builder families.

These metadata changes look safe, but they extend beyond the stated narrow hoisting prerequisite and intentionally alter sourcemap behavior without lineage-specific coverage. Could we either narrow the set to the hoisting path, or document the broader SchemaInjection sweep explicitly and add appropriate location tests? At minimum, the count and “builder-call rebuild family” wording should be corrected.

mathpirate (commented):
The core implementation looks conceptually sound: `sourceMapRange` is the right position-only carrier here, and I did not find a functional regression in the code.

I did find gaps between the PR's claims and what the committed regression test proves, plus several stale contradictions in the canonical handoff. I recommend tightening those before merge.

Verification at this head: full ts-transformers suite 1104 passed / 0 failed; root check, package check/lint/fmt, and `git diff --check` passed; transformed output for `age-category.tsx` and `self-improving-classifier.tsx` was byte-identical to the base.

The PR description should also be refreshed: BuilderCallHoisting is currently stage 17 (not 14), and the current suite has 1104 tests (not 1014).

packages/ts-transformers/test/lineage-regression.test.ts
mathpirate: Incorporated in 0421b089c — and the diagnosis was exactly right: the export-default span containing every marker made the global check nearly vacuous.

The test now binds content per `{tag, role}`: each hoisted lift must contain **exactly one** origin marker, each marker claimed **exactly once** (`claimedLiftMarkers` — permuted anchors now fail by construction), with per-tag assertions for the handler/pattern hoists, `bump`, and export-default. It also collects the rewritten post-hoist SITES (`__cfLift_N(...)`/`__cfHandler_N(...)` applications, `*WithPattern(__cfPattern_N, …)` enclosing calls) and asserts each recovers to the SAME origin as its hoist — which pins the N2/N5 outer anchors you showed were unguarded. Bite-verified: reverting `array-method-transform.ts` alone now fails the test.

*(reply via Claude Fable 5, on behalf of gideonwald)*

packages/ts-transformers/test/lineage-regression.test.ts
mathpirate: Incorporated in 0421b089c. Your `action(() => count * 4)` suggestion is now ORIGIN-E verbatim (thanks for pre-testing it) — it produces the `__cfHandler_N` hoist, the sanity set requires one, and its call/site recovery is asserted; `pattern-builder.ts` (both build methods) and `capture-scaffold.ts` now individually fail the test when reverted.

The direct-N3 fixture resisted: a corpus-wide probe (temporary stderr tap in the zero-input branch, full suite) shows it firing only for branch-root shapes with non-destructured pattern input (`!state.task.done` in jsx-direct-branch-roots, `!cellRef?.length` in reactive-cell-map). Four escalations in our destructured-input fixture (`!flag`, `!items.length`, `!task.done`, `flag || !task.done ? … : …`) all got claimed by the input-bound route first. Documented as a known boundary in the test header; N3's preservations stay (§6a-symmetric), flagged for a direct unit pin alongside CT-1870's injection acceptance.

*(reply via Claude Fable 5, on behalf of gideonwald)*

packages/ts-transformers/APRIME-LINEAGE-HANDOFF.md
mathpirate: Fixed in 0421b089c — a real reconciliation miss (the note survived the as-shipped sweep). The §5 note now states the shipped rationale (smr rides later `update*` rebuilds via the emitNode merge, so downstream stages need no cooperation); the §6 design sketch is explicitly marked as the pre-implementation ORIGINAL, superseded on the wrapper rule by §6a and the authoritative `ast/utils.ts` doc comments; the §10 ticket record flags that CT-1868's title wording predates the shipped design; and the regression-test header no longer mentions `create → update*` conversions.

*(reply via Claude Fable 5, on behalf of gideonwald)*

packages/ts-transformers/APRIME-LINEAGE-HANDOFF.md
mathpirate: Count corrected to 17 in 0421b089c (your count was right; the doc's own line-list even summed to 17 against its "16" prose). On classification we kept the breadth deliberately, now stated explicitly in §5: "builder call" is the `detectCallKind`-builder sense — the set includes non-hoistable factory/API builders (conditionals, cell factories/`.asSchema`, `wish`, `generateObject`, `sqliteQuery`, `fetchJson`) because BuilderCallHoisting also visits in-place authored builder consts for `__cfReg` registration and A′ needs their `src` equally. The coverage asymmetry you called out is acknowledged in the same note: the regression test location-pins the hoistable families + the in-place handler; the factory shapes ride the same §6a wrapper without dedicated location tests, to be added with CT-1870's injection acceptance.

*(reply via Claude Fable 5, on behalf of gideonwald)*

mathpirate:
Review incorporated in 0421b089c (details threaded on each comment). Summary items: PR description refreshed — hoisting referenced by its canonical registry ordinal (17) and the suite count updated to 1104; the description also now records the strengthened bite matrix (reverting `pattern-builder.ts`, `capture-scaffold.ts`, `array-method-transform.ts`, or `schema-injection.ts` individually fails the regression test — previously only schema-injection bit). Thanks for the review: independently re-running the full gate before commenting, and pre-testing the `action` fixture suggestion, made every finding directly actionable.

*(reply via Claude Fable 5, on behalf of gideonwald)*

PR #4632, opened 2026-07-09
chore: roll pinned Deno toolchain to 2.9.2, keep local floor at 2.8.0

Hixie:
Deno 2.9.2 is the current stable release. This moves the toolchain that continuous integration and the container images build against forward from 2.8.1, while leaving the minimum version required for local development where it was.

The version is pinned in several places that install an exact toolchain, and those all move to 2.9.2 together:

- mise.toml selects the toolchain for contributors who manage runtimes with mise.
- The deno-setup composite action supplies the default version that CI jobs install and cache. setup-deno treats an exact version string as an exact pin, so CI always runs exactly this version rather than the newest release in any range.
- Dockerfile.toolshed pins the base image for both the builder stage and the final runtime stage of the toolshed container.

Separately, tasks/check.sh guards the developer's local Deno version. It accepts a range rather than an exact version, and that range is independent of the exact version CI installs. Narrowing it to only the 2.9 minor line would force every developer to upgrade their local Deno before they could run `deno task check`, even though nothing in the codebase requires 2.9. So the range is widened rather than shifted: the minimum stays at 2.8.0 and only the maximum rises to 2.10.0, so the guard accepts any 2.8.x or 2.9.x. Developers already on 2.8.x are not forced to upgrade, and the pinned 2.9.2 is accepted as well.

packages/toolshed/Dockerfile is left at its older pin. It is not part of this in-sync set, is not referenced by any build path, and has not been kept current with the others.

---
## Summary by cubic
Update the pinned Deno version used by CI and container images to 2.9.2. Keep local development support at 2.8.x–2.9.x by widening the allowed range to <2.10.0, so no one is forced to upgrade.

- **Dependencies**
  - Bump pins to 2.9.2 in `mise.toml`, `.github/actions/deno-setup`, and `Dockerfile.toolshed` (builder and final images).
  - Update `tasks/check.sh` to allow >=2.8.0 and <2.10.0.
  - Leave `packages/toolshed/Dockerfile` on its older pin; it is not used by any build path.

<sup>Written for commit 99a2a2e3daf58776cfc908fa617b47fa9e0917d5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4632?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Hixie:
This includes two fixes to coverage deno that we upstreamed:
* trailing comments no longer affect coverage (https://github.com/denoland/deno/pull/35741)
* `} else {` being covered now is not different if both branches happened to be covered in the same run vs only one being covered (https://github.com/denoland/deno/pull/35858)

Hixie:
abandoning roll, will need upstream fixes

PR #4634, opened 2026-07-09
feat(runner,memory): per-doc scheduler state — reload restore + subscription-carried adoption (F6c)

seefeldb:
F6c from the #4288 review, redirected mid-flight to its real goal: **clients use per-doc scheduler state to skip work in ongoing operation, not just on reload** — observations ride memory subscriptions down, receivers adopt the writer's runs. Flag-on only (`persistentSchedulerState`); flag-off behavior byte-identical. Design docs: `docs/specs/scheduler-v2/per-doc-rehydration.md` + `incremental-observation-adoption.md`.

## What's here

**1. Per-doc restore (reload).** Resume now lists the space's snapshots once (the engine query was already pieceId-optional), buckets them per piece doc, and every descendant — sub-pattern nodes, map/filter/flatMap rows, each persisted under its own doc-derived pieceId — registers against its own bucket. `register()` decouples snapshot-apply from the synced-hold (flag-on children previously got NEITHER — they raced streaming data). Child-starting coordinators declare `resumeMode: "always-run"`: a clean coordinator snapshot otherwise skips the reconcile and strands its rows unregistered (probe-proven dead per-element reactivity). Persist side only writes doc-keyed observations. Red→green: `reload-rehydration-map-children.test.ts` (loopback two-manager; pre-fix all 3 row lifts re-ran with clean snapshots unused; post-fix zero op runs, `rehydrate/ok=3`, liveness + append intact).

**2. Incremental adoption (live).** `SessionSync` gains `observations`, sourced at push time from the snapshot store by the sync's `(fromSeq, toSeq]` commit-seq window, attached only for connections that negotiated the flag at hello (mirrors `#syncSchemaTable`), echo-suppressed like doc upserts. The runner fires a `scheduler-observations` notification after the integrate invalidation in the same synchronous turn — before dispatch — with oracles closed over replica state (per-doc `confirmed.seq` currency; `record.pending` local-write overlap). `Scheduler.adoptRemoteObservations` applies reload-identical fingerprint validation; computations only; clean successful rows only. Per-user/reader-isolated derivations can never falsely adopt (instance keys fold the per-user doc ids). Test: `observation-adoption-live.test.ts` — receiver converges with **zero per-element op runs**, local reactivity preserved, adoption symmetric.

**3. Capability-skew fix.** Pre-existing (bisected to before this branch): a flag-on client against a flag-off server had every write poisoned — the server strips observations, the observation-only batch commit (`operations: []`) is terminally rejected, and the flush-before-commit fails every subsequent handler write. The client now fails closed on the server's hello-advertised flags (drop observations client-side, resume-fresh without the listing RPC), with a red-green skew regression test; the multi-runtime harness propagates `EXPERIMENTAL_*` env into its server realm. This is what made flag-on multi-user pass at all (was 0/7 in ~1m40s of timeouts; now 7/7 in ~10s).

## Measured (group-chat-adoption-bench, 10 messages, alice sends; bob + alice-tab2 pure receivers; scheduler run-start deltas)

| config | sender | receivers | adopted | total | wall |
|---|---|---|---|---|---|
| main (v1) | 401 | 400 | — | 801 | 2.7s |
| branch flag-off | 201 | 400 | — | 601 | 2.4s |
| **branch flag-on** | 201 | **332** | **68** | **533** | 2.3s |

Flag-on beats main: total −33%, receivers −17%, zero receiver op re-runs replaced by adoptions. (v2 already halves the sender cascade at the runtime level, flag-independent.) Residual receiver runs are effects plus window-lag laggards — a cascade's later observations can ride one push window behind their data; aligning the observation-batch flush with its cascade's window is the characterized next lever, along with resume-time read pre-warming (which takes the flag-on reload churn gate from the pinned ≤1 to 0).

## Validation

Runner **825/0**, memory 334/0; flag-on and flag-off group-chat multi-runtime 7/7; flag-on home-rehydration churn gates green; the reload CI job's notebook test green with headroom.

Stacked on `claude/sched-v2-caughtup` (#4288) — deno.yml runs PR CI only against main, so the validation above is the record; retarget to main after #4288 merges.

---
## Summary by cubic
Enable per-doc scheduler state so clients skip recompute on reload and adopt observations live over subscriptions, with strict watch/reader gating and fail-closed checks for unverifiable reads; behind `persistentSchedulerState` and flag-off unchanged. In the group-chat bench, receiver runs dropped ~17% and total runs ~33% vs main.

- New Features
  - Per-doc restore on reload: one space-wide snapshot listing; descendants register against their own doc buckets. `map`/`filter`/`flatMap` coordinators use `resumeMode: "always-run"` to reattach rows; children rehydrate individually. Only doc-keyed observations are persisted.
  - Incremental adoption (live): the server pushes scheduler observations for each sync window to negotiated clients; the runner emits a `scheduler-observations` notification, and the scheduler adopts matching clean computation rows after fingerprint validation. No per-element op re-runs on receivers; local writes still run.

- Bug Fixes
  - Live adoption safety: server ships observation rows only when all reads are within the session’s tracked docs; user-scope rows require the writer’s principal, session-scope rows never ship. Engine persists the writer’s commit session key; the runner refuses adoption when a doc has no local confirmed base. Also excludes `resumeMode: "always-run"` coordinators from live adoption to avoid skipping child registration.
  - Reload safety: the same reader-isolation gate applies to the boot snapshot listing; cross-principal rows are dropped and degrade to fresh runs.
  - Capability skew: a flag-on client talking to a flag-off server now degrades gracefully (drops observations, resumes fresh) instead of wedging commits. The multi-runtime harness propagates `EXPERIMENTAL_*` flags to the server realm to avoid unintentional skew.
  - Writer session on coalesce: identical observations from a new writer now refresh only the row’s writer session id (latest principal) while preserving the original semantic commit seq, keeping adoption-window visibility intact.
  - Experimental flag banner: write overrides banner to process stderr in Deno (and keep console logging only in browser realms) to avoid contaminating CLI stdout; avoids console.error/warn to keep test consoles clean.

<sup>Written for commit eb35dee002fa02ee4872bc215106d562118272f0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4634?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seefeldb:
Thanks both — reviewed all three findings; two were real and are now fixed (e1669e4a4, eb35dee00).

**Coordinator adoption (codex P1 / cubic P1) — valid, fixed.** Confirmed reachable: map/filter/flatMap coordinators persist observations as `actionKind: "computation"` with runtime-stable fingerprints, so they passed every `adoptRemoteObservations` guard, and their reconcile is what (re)registers per-element children. When the coordinator observation lands in the same push window as the collection change, the receiver marks it dirty (integrate) then adopts it clean (scheduler-observations) in one turn → the appended row's child is never registered → dead reactivity. This is exactly what `resumeMode: "always-run"` prevents on reload; `register()` guarded it but adoption didn't (resumeMode was a register-time option, not stored on the action). Fix: an `alwaysRunActions` WeakSet populated at registration, skipped in `adoptRemoteObservations` — the reload guard, live. The group-chat bench confirmed it was live, not theoretical (adoptOk dropped 60→56 with the guard, i.e. ~4 unsound same-window coordinator adoptions were happening). Red-green unit test in `scheduler-observations.test.ts`.

**Writer session_id staleness (cubic P1) — valid, fixed, but the first patch had a bug of its own.** The row did keep the first writer's session after an identical-payload coalesce, and since that key gates reader-isolated (C6) adoption, a user-scope row was withheld from the actual latest writer. An adversarial re-review of my initial patch caught that making the update unconditional *also* nulled `o.commit_seq` on observation-only coalesces — and since `COALESCE(s.commit_seq, o.commit_seq)` is the adoption-window filter and `upsertSchedulerSnapshot` already nulls `s.commit_seq` on observation-only commits, `o.commit_seq` was the sole surviving carrier of the semantic seq. That silently dropped rows from the live-adoption window. Final fix (eb35dee00) splits the path: `payloadChanged` does the full update; an identical-payload coalesce refreshes **only** `session_id`, preserving `commit_seq`. Two red-green tests: writer follows latest, and window-visibility survives a semantic→observation-only re-run.

**Write-address exposure (cubic P2) — declined, with reasoning.** The write addresses are the same opaque-doc-id class the boot listing already returns space-wide to any session that can read the space, and the live push applies the same `#observationScopeAdoptable` gate (which *does* inspect all write-address scopes — session-scope drops the row, user-scope requires writer principal == receiver principal) plus the extra `readsTracked` filter, so live-pushed rows are a strict subset of what the listing already exposes. Requiring every write address to be watch-tracked would drop legitimate shared computations whose internal/materializer writes the receiver doesn't watch — an adoption-hit regression for no isolation gain. Left for the standing CFC review the spec already flags.

Also added the coordinator exclusion as a first-class **C7** in `incremental-observation-adoption.md` (it was a new adoption gate not in the C1–C6 list) with a cross-ref from `per-doc-rehydration.md` §3.3.

Validation: memory 337/0, runner 842/0, lunch-poll multiUserTest flag-ON 11/0, group-chat bench 531 total run-starts vs main's 801 (adoptOk 60), all red-green.

PR #4642, opened 2026-07-09
cli: replace mounted-callable stat poll with a directory-listing fall…

Hixie:
…back

resolveMountedCallableFile guarded against fabricated callable paths by polling Deno.stat up to 20 times with 50ms sleeps, tolerating the transient NotFound that a stale FUSE-T kernel cache reports while the fuse bridge clears and re-hydrates a piece prop subtree. The poll put a one-second upper bound on success: an invalidation window that outlasted the budget failed an invocation that would have succeeded, a genuinely missing file burned a full second before erroring, and each retry re-asked the same cached question that had just returned the stale answer.

Replace the loop with a single stat plus a parent-directory-listing fallback. FUSE-T cannot push kernel cache invalidations, but a readdir goes back to the bridge, which answers after re-hydrating the prop subtree, so the listing names every callable file that exists. A name absent from the listing means the file does not exist, and the check fails immediately with the same diagnostic. The listing entry's type flags come from the same cached attribute channel that produced the stale NotFound, so a name match counts as existence unless the entry is reported as a directory.

A Deno.watchFs-based wait was considered and rejected: FUSE-T mounts are NFS translations where daemon-side rehydration emits no local filesystem events, so a watcher would never fire for the window it is meant to cover, and an event-driven wait cannot terminate for a genuinely absent file without reintroducing a deadline.

The companion test previously created the callable file from a 50ms setTimeout and raced the resolver's poll budget under CPU contention. It now keeps the file on disk throughout and injects a stat through the new optional ExecDependencies.stat hook that reports NotFound, simulating the stale-cache window deterministically, and asserts a single stat call to pin the no-retry behavior.

---
## Summary by cubic
Replaced the mounted callable existence check with a single `Deno.stat` and a parent `Deno.readDir` fallback, with one recheck after the listing cache window. This removes the 1s polling loop, avoids false “not found” on FUSE-T/macOS NFS caches, and only delays genuine misses once before failing.

- **Bug Fixes**
  - On `Deno.errors.NotFound`, read the parent directory; a matching non-directory entry counts as existence.
  - If both `stat` and the first listing miss, wait ~3.5s to outlast macOS NFS cached listings, then list again before erroring.
  - Added optional `ExecDependencies.stat`, `readDir`, and `delay`; tests pin a single `stat` call, both listing consultations, and the one bounded wait.

<sup>Written for commit 520be022c56b6a57d85556afe35eb03989e7acb9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4642?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cli/lib/exec.ts
+++ b/packages/cli/lib/exec.ts
@@ -1,6 +1,6 @@
 import type { PieceManager } from "@commonfabric/piece";
 import { PiecesController } from "@commonfabric/piece/ops";
-import { dirname, join, relative, resolve } from "@std/path";
+import { basename, dirname, join, relative, resolve } from "@std/path";
 import {
   type MountedCallablePath,
   parseMountedCallablePath,
@@ -51,6 +51,7 @@ export interface ExecDependencies {
     manager: CallableManagerLike,
     pieceId: string,
   ) => Promise<CallablePieceLike>;
+  stat?: (path: string) => Promise<Deno.FileInfo>;
   timeoutMs?: number;
   uuid?: () => string;
   waitForResult?: (
@@ -114,32 +115,40 @@ async function readMountedPieceMeta(
   };
 }
 
-async function assertMountedCallableFileExists(absPath: string): Promise<void> {
-  for (let attempt = 0; attempt < 20; attempt++) {
-    let stat: Deno.FileInfo;
-    try {
-      stat = await Deno.stat(absPath);
-    } catch (error) {
-      if (error instanceof Deno.errors.NotFound) {
-        if (attempt < 19) {
-          await new Promise((resolve) => setTimeout(resolve, 50));
-          continue;
-        }
-        throw new Error(`Mounted callable file not found: ${absPath}`);
-      }
-      throw error;
-    }
-
-    if (stat.isFile) {
+async function assertMountedCallableFileExists(
+  absPath: string,
+  stat: (path: string) => Promise<Deno.FileInfo> = Deno.stat,
+): Promise<void> {
+  try {
+    if ((await stat(absPath)).isFile) {
       return;
     }
-
-    if (attempt < 19) {
-      await new Promise((resolve) => setTimeout(resolve, 50));
-      continue;
+  } catch (error) {
+    if (!(error instanceof Deno.errors.NotFound)) {
+      throw error;
     }
-    throw new Error(`Mounted callable file not found: ${absPath}`);
+    // While the fuse bridge rebuilds a piece prop it clears the subtree and
+    // re-hydrates it on demand, and FUSE-T cannot push cache invalidations
+    // to the kernel, so a stat during that window reports NotFound for a
+    // callable file that exists. The parent directory listing is answered
+    // by the bridge after hydration, so it names every callable file that
+    // exists; a file missing from the listing does not exist. The listing
+    // entry's type flags come from the same cached attributes as the stat,
+    // so a name match counts as existence unless the entry is a directory.
+    const name = basename(absPath);
+    try {
+      for await (const entry of Deno.readDir(dirname(absPath))) {
+        if (entry.name === name && !entry.isDirectory) {
+          return;
+        }
+      }
+    } catch (dirError) {
+      if (!(dirError instanceof Deno.errors.NotFound)) {
+        throw dirError;
+      }
+    }
   }
+  throw new Error(`Mounted callable file not found: ${absPath}`);
 }
 
 export async function resolveMountedCallableFile(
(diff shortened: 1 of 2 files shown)


Hixie:
cc @bfollington - my agent claims you're the best person to ask about the FUSE semantics here.

bfollington:
Dug into the FUSE-T semantics behind this change and verified them empirically — write-up + reproducible harness in #4654 (no runtime change; evidence only). Ran against FUSE-T 1.0.49 on macOS 15/arm64 using a synthetic FUSE-T filesystem that toggles a file's existence daemon-side with zero client syscalls and logs every kernel request, so we can see exactly what reaches the daemon vs. the client cache.

**The reasoning here checks out, and the mechanism is now pinned down:**

- The stale `NotFound` is specifically the **macOS NFS client's negative name cache**. A single daemon-side ENOENT seeds it; on an aged mount it persisted **~50s** (≈`acdirmax`), during which ~100 stats produced **zero** daemon traffic. So the old 20×50ms poll was hopeless — it was re-asking a cache that answers ~50× longer than its budget, and being pure `stat`s it could never break out. Replacing it is the right call.
- The `readdir` fallback is the right escape: a readdir that reaches the daemon uses READDIRPLUS and refreshes/clears the negative entry, collapsing the window to ~0.

**One thing worth addressing before merge — a reproduced false-negative:**

The fallback readdir isn't guaranteed to reach the daemon. If it fires while a *previously cached parent listing* is still inside the directory attribute-cache validity window (≈<3s since that dir was last listed), the macOS NFS client serves the **stale listing with no round-trip** — so the readdir comes back *without* the file and this path returns a false "not found" with **zero retries**, while the file exists. It's timing-narrow but reproducible (details + logs in #4654). Removing all retries is what turns the old "slow" case into an occasional "wrong" case.

Two fixes that actually hold up in the experiments:
1. A **single bounded retry on the readdir-miss path** — if stat ENOENTs *and* the fallback readdir also misses, sleep ~1–2s (enough to cross the validity window) and readdir once more before failing. Latency lands only on the miss path, and a genuinely-absent file still errors after one retry. Still far cheaper than the old poll, and unlike it, it escapes the negative cache.
2. Mount with **`-o noattrcache`** (systemic): window 50s→0.25s and the false-negative disappears, at the cost of an RPC per stat.

One thing that *doesn't* work, FWIW: bumping the directory's mtime — inside the validity window the client serves the cached listing without revalidating the directory, so it never sees the new mtime.

(Also confirmed: on Linux this whole class of staleness can't occur — timeout-0 replies, error-style ENOENT isn't negatively cached, readdir isn't kernel-cached — so this is macOS/FUSE-T only.)

Happy to send the bounded-retry tweak as a PR against this branch if useful.

bfollington:
Follow-up on the FUSE-T layer, since it determines whether this workaround is interim or permanent. Short version: **the root behavior is a FUSE-T limitation we can't fix ourselves and probably can't get cleanly fixed upstream soon — so a client-side workaround (this readdir fallback, ideally plus a bounded retry) is the right call.**

**Why the timeouts don't help us.** The stale `NotFound` is the macOS NFS client's negative-name / attribute cache. FUSE-T's own wiki states it **ignores the `entry_timeout`/`attr_timeout` a filesystem returns** — verbatim: *"Caching of attributes is done by the client. Currently the caching attributes returned by the filesystem implementation are ignored."* So our `entry_timeout=0` (set precisely to avoid this) is a no-op on macOS. On Linux those timeouts *are* honored, so this whole class of staleness is macOS/FUSE-T only.

**It's not a clean "won't fix" — FUSE-T is wedged between two failure modes.** It doesn't look deliberate: in [macos-fuse-t/fuse-t#71](https://github.com/macos-fuse-t/fuse-t/issues/71) the maintainer re-enabled the attribute cache ("for some reason it was disabled by default"), and in [#61](https://github.com/macos-fuse-t/fuse-t/issues/61) turning caching *off* made the macOS NFS client **livelock** (endless open/close) for restic on Sonoma. So the two options are "cache aggressively" (our stale lookups) or "don't cache" (client livelock). Today it's effectively binary — either it ignores the requested timeout, or you disable caching wholesale with `-o noattrcache`. There is no way to request a specific, honored timeout.

**We can't fix it ourselves.** A proper fix (honor `timeout=0`, or expose `acdirmax`/`actimeo` for a short middle ground) would land in FUSE-T's NFS server — which is **not in a public repo** (the open `macos-fuse-t` repos are just the libfuse shim + the SMB backend; the NFS `go-nfsv4` daemon ships as a binary). So we can only file an upstream issue and hope; given the #61 tension, "cache less" is a fraught ask.

**Practical upshot for this PR:** the readdir fallback is the right shape. The one gap (my earlier comment) is the zero-retry false-negative when the fallback readdir is *itself* served from a stale cached listing — a single bounded retry on the miss path closes it without depending on FUSE-T. `-o noattrcache` would kill the staleness globally, but it risks re-triggering the #61 livelock class and costs an RPC per stat, so it needs validation on a real workload first. (Quick local stress test: no livelock on macOS 15 / FUSE-T 1.0.49 for basic open/read/close, but that isn't the restic/Sonoma condition, so not conclusive.)

PR #4645, opened 2026-07-09
bench(runner): flat index-list shape — write/read/update dynamics as f(N)

bfollington:
## Why

A production capture of a Mobile Loom boot (25MB, 82s, ~180ms RTT; forensics in the loom repo, `docs/development/projects/mobile-loom-production-performance/findings.md`) showed a ~3k-item derived jump-search list weighing **~1.33MB per copy, re-materialized once per consuming pattern instance** — 58% of the boot payload — plus 140s of cumulative slow client-side `get()` warnings (median 65ms) re-reading unchanged docs. Before assigning blame between the runtime and that usage, this bench measures the runtime's **native dynamics for the shape in isolation**: a flat list of many similar shallow `{name, ref}`-ish items, the way any index/autocomplete list looks.

## What it measures

One new bench file, `packages/runner/test/cell-set-flat-index-list.bench.ts`, env-parameterized (`FLAT_INDEX_LIST_SIZES`, default 100/1000/3000), using `tx.journal.novelty()` for stored-bytes/docs accounting printed outside the timed windows:

1. **Write**: one `cell.set(array)` vs one `Writable.for().set()` per item — time, documents created, stored bytes vs raw JSON
2. **Read**: first `get()`, repeated `get()` ×100 on an unchanged doc, `get()` ×100 through a JSON schema (the validateAndTransform path)
3. **Update**: whole-array re-set with ONE item changed vs targeted `key(i).set` — bytes/docs written per transaction (history growth)

### 1. `cell.set()` of a flat object array creates one document PER ITEM

| N | docs created | stored bytes | raw JSON | overhead | per item |
|---|---|---|---|---|---|
| 100 | **101** | 70.3KB | 31.0KB | 2.27× | 703B |
| 1000 | **1001** | 717KB | 317KB | 2.26× | 717B |
| 3000 | **3001** | 2.18MB | 967KB | 2.26× | 728B |

"One flat array in one doc" does not exist for object arrays — every element becomes its own document (per-item `Writable` docs are the same count at 2.8× overhead). Downstream consequence: any watch/query over the list is N per-doc server materializations, and any client traversal is N doc hops.

### 2. Schema-path reads: linear, but ~19µs/item **per read**, no memoization

| N | first get() | repeated get() ×100 | schema get() ×100 | schema per-read |
|---|---|---|---|---|
| 100 | 164µs | 3.3ms | 165ms | **1.7ms** |
| 1000 | 554µs | 28.1ms | 1.8s | **18ms** |
| 3000 | 2.3ms | 90.1ms | 5.7s | **57ms** |

Good news: not superlinear. Bad news: the constant is brutal and **every read re-pays it in full on an unchanged document** — the schema path is ~63× the schemaless path. 57ms/read at N=3000 matches the production capture's median 65ms slow-`get()` on a ~5.6k-entry space almost exactly; production logged 1,876 re-reads of one unchanged doc = the observed 140s of client CPU.

### 3. Update profiles: regenerate-from-scratch rewrites everything; read-modify-write is cheap

Three writer profiles for "one item changed" (per-tx averages over 20 txs; accounting outside the timed window):

| N | REGENERATE from scratch | READ-MODIFY-WRITE | targeted `key(i).set` |
|---|---|---|---|
| 100 | 27ms · 61.6KB · 101 docs | 6.9ms · **685B · 2 docs** | 1.6ms · 246B · 1 doc |
| 1000 | 445ms · 629KB · 1001 docs | 85ms · **685B · 2 docs** | 16ms · 246B · 1 doc |
| 3000 | **2.15s · 1.92MB · 3001 docs** | 220ms · **685B · 2 docs** | 41ms · 246B · 1 doc |

Per review (@seefeldb): the original "whole-set" finding modeled a **regenerate-from-scratch** writer — fresh objects carry no doc identity, so every element re-mints. The idiomatic **read-modify-write** edit (get → replace one element → set back) reuses doc identity from the `get()` and lands as a flat 2-doc/~685B change at any N; writing back an unchanged array is a perfect no-op (0 docs). The full-rewrite cost applies specifically to derivation-style writers that rebuild lists from scratch — which is what our production index refresh does.

Also pinned in this PR (`patterns-lift.test.ts`): a lifted array-of-objects **result** is stored inline as ONE doc (raw write path) — derived output does not go through the editable-array decomposition.

## Interpretation

- The runtime is **efficient at targeted updates** (246B/tx regardless of N) — the primitives for cheap list maintenance exist.
- The pathological costs come from (a) the **doc-per-element explosion** on plain `cell.set`, (b) the **unmemoized schema-read path**, and (c) **wholesale re-sets** — which is exactly how derived/lifted lists get refreshed in practice.
- Production compounds this by materializing the same derived list **once per consuming pattern instance** (10 copies in the capture). That part is usage/framework layering, not this bench's subject — but it multiplies every number above.

## Test plan

- `deno task bench` filtered to the new file; full matrix run attached above.
- `deno fmt` / `deno lint` clean.
- Env knobs verified (`FLAT_INDEX_LIST_SIZES=50 FLAT_INDEX_LIST_UPDATE_TXS=3` smoke run).

---
## Summary by cubic
Adds benchmarks for flat index/autocomplete list write/read/update costs and decomposes schema-read overhead by schema depth, plus a test pinning derived array outputs to a single inline document. Helps explain where read costs come from and how update strategies scale.

- **New Features**
  - Added `packages/runner/test/cell-set-flat-index-list.bench.ts` to measure one-doc vs per-item writes, first/repeated/schema reads, and three update profiles at N=100/1000/3000; config via `FLAT_INDEX_LIST_SIZES`, `FLAT_INDEX_LIST_UPDATE_TXS`, `FLAT_INDEX_LIST_REPORT`.
  - Added `packages/runner/test/cell-schema-read-depth.bench.ts` to read the same 1000-item doc through schemaless, `items: true`, permissive `items: {type: "object"}`, and full item schema to locate schema-path costs.
  - Updated `packages/runner/test/patterns-lift.test.ts` to assert a lifted array-of-objects result is stored inline as a single doc (no per-element entity docs).

- **Performance**
  - `cell.set()` of an object array creates N+1 documents.
  - Schema-path reads are ~19µs/item/read (~63× schemaless); depth bench shows merely entering the schema path (`items: true`) is already ~30×, a permissive object schema is worst, and a fully-declared schema is cheaper — field validation isn’t the main cost.
  - Updates with one changed item: regenerate rewrites all N+1 docs; read-modify-write averages ~685B across 2 docs per tx and stays flat with N; targeted `key(i).set` stays ~246B/tx.

<sup>Written for commit eaa71fd994f54a53aa538d0688c9097500c97c19. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4645?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/test/cell-schema-read-depth.bench.ts
+++ b/packages/runner/test/cell-schema-read-depth.bench.ts
@@ -1,0 +1,128 @@
+/**
+ * Decomposes the schema-path read cost by schema DEPTH on one unchanged doc.
+ *
+ * Companion to cell-set-flat-index-list.bench.ts: that bench showed schema
+ * get() of an unchanged N-item list costs ~19µs/item PER READ (~63× the
+ * schemaless path, no memoization). This bench answers WHERE that cost
+ * lives by reading the SAME stored 1000-item list through schemas of
+ * increasing depth:
+ *
+ *   - schemaless            → the baseline proxy path
+ *   - items: true           → schema path entered, zero per-item validation
+ *   - items: {type:object}  → permissive object schema (no properties)
+ *   - full item schema      → every field declared
+ *
+ * Result shape (M-series MacBook): schemaless ~0.6ms/read; items:true
+ * ~18ms/read — i.e. merely ENTERING the schema path costs ~32× even when
+ * validating nothing per item; a PERMISSIVE object schema is the worst
+ * case (~40ms/read, full recursive walk per element); the fully-declared
+ * schema is cheaper than permissive (~28ms). Field validation is not the
+ * cost — the per-node traversal machinery (tracked reads, link handling,
+ * freeze-discipline checks, fresh result allocation; GC ≈ 39% of ticks
+ * under profile) is. A V8 profile of the same loop bottoms out in
+ * validateAndTransform → traverse → deep-freeze checkValue.
+ *
+ * Environment controls:
+ * - SCHEMA_READ_DEPTH_N: list size, default 1000
+ * - SCHEMA_READ_DEPTH_READS: get() calls per bench iteration, default 100
+ */
+
+import { Identity } from "@commonfabric/identity";
+import { StorageManager } from "@commonfabric/runner/storage/cache.deno";
+import { Runtime } from "../src/runtime.ts";
+import { type JSONSchema } from "../src/builder/types.ts";
+import { type IExtendedStorageTransaction } from "../src/storage/interface.ts";
+
+const signer = await Identity.fromPassphrase("bench schema read depth");
+const space = signer.did();
+
+const N = Number(Deno.env.get("SCHEMA_READ_DEPTH_N") ?? "1000");
+const READS = Number(Deno.env.get("SCHEMA_READ_DEPTH_READS") ?? "100");
+
+const items = Array.from({ length: N }, (_, i) => ({
+  data: {
+    group: "Pages",
+    id: `People/P${i % 40}/Notes/entry-${i}.md`,
+    kind: "page",
+    label: `entry-${i}`,
+    name: `entry-${i}`,
+    path: `People/P${i % 40}/Notes/entry-${i}.md`,
+  },
+  group: "Pages",
+  label: `entry-${i}`,
+  searchAliases: [`People/P${i % 40}/Notes/entry-${i}.md`],
+  value: `page:People/P${i % 40}/Notes/entry-${i}.md`,
+}));
+
+const FULL_ITEM_SCHEMA = {
+  type: "object",
+  properties: {
+    data: {
+      type: "object",
+      properties: {
+        group: { type: "string" },
+        id: { type: "string" },
+        kind: { type: "string" },
+        label: { type: "string" },
+        name: { type: "string" },
+        path: { type: "string" },
+      },
+    },
+    group: { type: "string" },
+    label: { type: "string" },
+    searchAliases: { type: "array", items: { type: "string" } },
+    value: { type: "string" },
+  },
+} as const satisfies JSONSchema;
+
+const VARIANTS: [string, JSONSchema | undefined][] = [
+  ["schemaless", undefined],
+  ["items:true (passthrough)", { type: "array", items: true }],
+  ["items:{type:object} (permissive)", {
+    type: "array",
+    items: { type: "object" },
+  }],
+  ["full item schema", { type: "array", items: FULL_ITEM_SCHEMA }],
+];
+
+function setup() {
+  const storageManager = StorageManager.emulate({ as: signer });
+  const runtime = new Runtime({
+    apiUrl: new URL(import.meta.url),
+    storageManager,
+  });
+  return { runtime, storageManager };
+}
+
+async function seed(
+  runtime: Runtime,
+): Promise<IExtendedStorageTransaction> {
+  const tx = runtime.edit();
+  runtime.getCell<typeof items>(space, "schema-read-depth-doc", undefined, tx)
+    .set(items);
+  await tx.commit();
+  return tx;
+}
+
+for (const [label, schema] of VARIANTS) {
+  Deno.bench({
+    name: `schema read depth - ${label} - get() x${READS} (${N} items)`,
+    group: "schema-read-depth",
+    baseline: schema === undefined,
+    async fn(b) {
+      const { runtime, storageManager } = setup();
+      await seed(runtime);
+      const cell = runtime.getCell(
+        space,
+        "schema-read-depth-doc",
+        schema as JSONSchema | undefined,
+      );
+      cell.get(); // warm
+      b.start();
+      for (let i = 0; i < READS; i++) cell.get();
+      b.end();
+      await runtime.dispose();
+      await storageManager.close();
+    },
+  });
+}
--- a/packages/runner/test/cell-set-flat-index-list.bench.ts
+++ b/packages/runner/test/cell-set-flat-index-list.bench.ts
@@ -1,0 +1,415 @@
+/**
+ * Benchmarks Cell.set()/get() for a FLAT LIST of many similar shallow items —
+ * the "search index / autocomplete list" shape.
+ *
+ * Why: a production capture of a Mobile Loom boot (loom repo,
+ * docs/development/projects/mobile-loom-production-performance/findings.md,
+ * 2026-07-10 HAR forensics) showed a ~3k-item derived jump-search list
+ * weighing ~1.33MB per copy (~445B/item) and re-materialized once per
+ * consuming pattern instance — 58% of a 25MB boot payload. Before assigning
+ * blame between the runtime and that usage, this bench measures the
+ * runtime's NATIVE dynamics for the shape in isolation:
+ *
+ *   1. write cost + stored bytes + docs created for one flat array in ONE
+ *      doc vs one doc PER ITEM, as f(N)
+ *   2. read cost: first get(), repeated get() (does anything memoize?), and
+ *      get() through a JSON schema (the validateAndTransform path), as f(N)
+ *   3. update cost: whole-array re-set with one item changed vs a targeted
+ *      per-index write — bytes written per transaction (history growth)
+ *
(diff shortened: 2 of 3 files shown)


seefeldb:
the doc explosion on .set is currently WAI because we assume those arrays (unlike lift's see below) can be edited. let's just make the reads fast before we optimize this further. we can also add a way to hint that an array won't be changed.

re finding 3, full set: if the causes aren't stable (they should be if you reset to exactly the same frame cause or use .for(..)), then each doc gets a new id, so it's not surprising that it's a full array. but in practice we don't do that, we'd usually read the array, change a thing and write it fully back, and that should turn into a one-doc change since it can get back the doc id from the objects from the original get.

the output of derived/lifted should not go through that doc explosion, those arrays can't be edited, so they should be written just as a single doc. worth double checking and pinning with a test if there isn't one already (i bet there is, but...).

---

clearly the schema overhead is what we need to look into!!

bfollington:
*(Note: this comment is from Claude, working on Ben's behalf and posting from his account.)*

Thanks @seefeldb — you were right on both counts; verified and folded into the PR (`62da95cc3`):

**RMW vs regenerate:** read-modify-write of one item is a **2-doc / ~685B** change flat at every N, and writing back an unchanged array is a perfect no-op (0 docs). The full-rewrite behavior applies specifically to **regenerate-from-scratch** writers — fresh objects carry no doc identity, so every element re-mints (1.92MB / 3,001 docs per tx at N=3000). I added READ-MODIFY-WRITE as a third update profile in the bench and updated the results table (also moved the novelty accounting outside the timed window, per cubic). One thing worth knowing from our side: regenerate-from-scratch is exactly what our production writer does — the jump-index list is a derivation recomputed from a source index, so every refresh pays the worst-case profile. We'll fix that in our usage, but any derivation that maps over an array produces this write shape by default.

**Lift output:** confirmed — derived results write via the raw path (`runner.ts` `setRawUntyped`) and land as ONE doc with the array inline; our production capture agrees (the derived lists arrive as single ~1.33MB docs). I didn't find an existing test pinning it, so this PR now adds one to `patterns-lift.test.ts`: asserts a lifted 50-object array result is stored inline with no per-element entity docs (a regression would multiply every derived list's doc count by its length).

**Schema overhead — agreed, that's the lever.** The bench isolates it: ~19µs/item **per read** on an unchanged doc through the schema path, ~63× the schemaless path, no memoization. In production we captured 1,876 re-reads of one unchanged doc ≈ 140s of worker CPU on a phone-sized load. Next step from our side is a follow-up bench/profile splitting where the per-item cost actually goes (validateAndTransform vs proxy creation vs traversal bookkeeping) — unless you already know where the body is buried and can point us at the right file. Also keen on your "array won't be changed" hint idea: for this shape (search index) the consumer wants the whole array in one hit and never edits it, so an immutable/inline annotation fits perfectly — happy to prototype it if you sketch the preferred spelling (schema keyword vs asCell-style brand).

bfollington:
*(Note: this comment is from Claude, working on Ben's behalf and posting from his account.)*

**Schema-read profiling results** — following up on "clearly the schema overhead is what we need to look into": we profiled where the ~19µs/item/read goes, and added a reproducing bench (`cell-schema-read-depth.bench.ts`, `eaa71fd99`).

**The headline: the cost is the schema-path machinery per node, not validation.** Reading the *same stored 1000-item doc* through schemas of increasing depth:

| schema | per read | vs schemaless |
|---|---|---|
| schemaless | 0.39ms | 1× |
| `items: true` — **validates nothing** | 11ms | **29.5×** |
| `items: {type:"object"}` — permissive | 29ms | **74×** |
| full item schema (every field declared) | 20ms | 51× |

Merely entering the schema path costs ~30× even with zero per-item validation work, and a *permissive* object schema is the worst case (full recursive walk per element) — notably worse than a fully-declared schema.

**V8 profile of the hot loop** (`get()` ×200, N=1000, 18ms/read): bottoms out in `validateAndTransform` (schema.ts:592) → `traverse` (traverse.ts:1971) → deep-freeze `checkValue`/`freeze` (deep-freeze.ts), with **GC at 39% of non-library ticks** — per-read result materialization + immutability-discipline checks dominate.

**One acquittal:** we experimentally wired a shared `SchemaMemo` into `validateAndTransform`'s traverser (the ctor param exists at traverse.ts:2564 but nothing passes it at schema.ts:1074) — no effect (18.2 → 20.1ms). Schema re-interning is not the cost.

**Suggested directions, in value order:**
1. **Memoize the validated result per (doc revision, interned schema).** The result is deep-frozen/immutable, so returning the same object until the doc changes looks safe by construction. Repeated reads of unchanged docs become ~free — in our production capture that's 1,876 re-reads of one doc ≈ 140s of worker CPU on a phone-class load, reduced to one materialization.
2. **Attack the `items:true` constant** — ~11µs/item with nothing to validate is the floor every schema'd read pays; the profile says it's tracked-read bookkeeping + freeze-discipline verification + fresh allocation per node.
3. GC churn (39%) falls out of either.

Happy to attempt (1) as a PR if the invalidation story is as simple as it looks from outside — or hand these numbers to whoever owns this path if there's context we're missing.

bfollington:
*(Note: this comment is from Claude, working on Ben's behalf and posting from his account.)*

@seefeldb — we traced every access of the index arrays through the Mobile Loom pattern tree (`cf-loom-mobile` + all sub-pieces). Precise answer:

**There is no spurious `.get()`.** Grep-clean: zero direct reads of `fabricIndex` / `pageIndex` / `fileShortcuts` / `autocompleteItems` anywhere in the pattern files. All threading passes cells by reference — handlers receive them as `Writable<>` state (asCell) and forward; every sub-piece boundary declares `Writable<Default<FabricIndexEntry[], []>>` and forwards. The proxy trick isn't needed pattern-side: nothing eagerly materializes the array in pattern code.

**But there IS a spurious *lift* — an identity function, instantiated ~9×:**

```ts
// ui/entity-index.ts — comment says it outright: "this is a reactive passthrough"
export const buildFileShortcuts = lift(
  (input: { fabricIndex: FabricIndexEntry[] }): FileShortcut[] =>
    input?.fabricIndex ?? [],
);
```

`FabricIndexEntry` IS `FileShortcut` (the lift exists as a leftover from when the shortcuts were computed by an in-pattern tree walk; the daemon now precomputes the flat index, so the lift became a copy). It's invoked in the root pattern **and inside `CaptureForm`, which is instantiated once per frame** (each tab's composer sheet, the capture piece, the viewers). Each instantiation reads the full array and persists an identity copy as its own result doc — that's exactly the nine ~1.33MB byte-identical docs in our HAR (their slightly different item counts, 2980/2986/2999, are snapshots at different refresh times). The tenth copy is `buildAutocompleteItems` (a real derivation, 2.7MB) — also re-derived per CaptureForm instead of once.

**So the complete reader list is:**
1. `buildFileShortcuts` — identity lift, ~9 instantiations → **pure waste, deletable**
2. `buildAutocompleteItems` — real derivation, per-frame instead of once → **hoist to root, pass the result as `Reactive<…>`** (which the codebase already documents keeps it an opaque cell across pattern boundaries — `home.tsx` does exactly this for Home's copy)
3. `<ct-autocomplete items={…}>` — the UI component, the one true end consumer

**Fix on our side, no runtime change needed for this part:** delete the identity lift (pass `fabricIndex` straight through — types already identical), hoist the autocomplete derivation to the root, thread results as `Reactive<>`. Expected effect on the capture's space: ~14.6MB of boot payload → ~2.7MB (one derivation) — before the item-shape slimming (−72% of what remains) that we're doing next.

The runtime asks from the profiling comment above (result memoization per doc revision; the `items:true` per-node constant) still stand — they're what makes the *remaining* legitimate reads cheap.

PR #4619, opened 2026-07-08
feat(shell): enable system-pattern auto-update for the non-home root

seefeldb:
## Enable system-pattern auto-update for the non-home root

The mechanism landed in #4611 behind a default-off flag. This PR is the opt-in
**flip**: it defaults `EXPERIMENTAL.systemPatternAutoUpdate` to **on**, so a
space's non-home root pattern (`default-app.tsx`) rolls forward in place when
its toolshed serves a newer content identity.

### Scope

- **Non-home default-app root only.** The **home** root stays off — it carries
  real user data (favorites / journal / spaces) and is gated behind the
  separate `systemPatternAutoUpdateHome` flag, pending the `home.tsx`
  stable-addressing audit (spec § open question 4).
- A build define `EXPERIMENTAL_SYSTEM_PATTERN_AUTOUPDATE=false` can still force
  it off per deployment.

### The change

One line in `packages/shell/src/lib/env.ts`
(`flagValue(...) ?? true`, mirroring `eagerSourceAnnotation`'s default pattern),
plus the `env.test.ts` snapshot expectation.

### Before merging

Per the phasing in `docs/specs/pattern-imports/pattern-updates.md`, a **CI
golden replay** against the system patterns (instantiate a space from
version N of `default-app.tsx`, seed representative state, roll to N+1 in place,
assert no crash + state survives) is the intended gate before this is on by
default. The in-place-swap unit test in #4611 is a golden-replay-lite; a
state-seeded replay is the remaining bar. Holding this PR unmerged until that's
in place (or a deliberate decision to ship without it).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Enable automatic in-place updates for the non-home root system pattern by default, so `default-app.tsx` rolls forward when a newer identity is available. The home root stays off behind `systemPatternAutoUpdateHome`; deployments can still disable via `EXPERIMENTAL_SYSTEM_PATTERN_AUTOUPDATE=false`.

<sup>Written for commit 19a16e6020c73b4939a9abcb4e8720292fc8f08c. Summary will update on new commits.</sup>

[...]

PR #4653, opened 2026-07-10
fix(shell/piece): unknown-build version skew must not raise the reload banner (local default-app red)

seefeldb:
Fixes the deterministic local-only failure of `deno task integration patterns default-app` ("should create a note via default app and see it in the space list", 60s `waitForState` timeout at `default-app.test.ts:486`) introduced by the #4619 flag flip, while CI stayed green.

## Mechanism

With `systemPatternAutoUpdate` on, space open runs `checkAndUpdateDefaultPattern`. On local **source-run servers neither side has a build sha** (a source toolshed serves `/api/meta` `gitSha: null`; a dev/`felt` shell build has no `COMMIT_SHA` define), so the version gate fails — correctly, fail-safe, no update — but the failure path unconditionally fired `reportVersionSkew`. That IPC raises the shell's "A newer version is available" banner: `position: fixed` at the **top**, `z-index: 2000` — directly over the header breadcrumbs.

The integration test's click helpers dispatch **trusted, hit-tested CDP clicks**. Body buttons ("Notes", "New Note") sit below the banner band and work; the header space link does not — the click back to the space landed on the banner, the view never changed, and the one-shot `waitForState({ view: { spaceName } })` timed out at its 60s ceiling. Note creation itself was fine (the same run's second test passes; post-fix the full create-and-return flow takes ~400ms).

## Why CI never saw it

The `pattern-integration-test` job runs the **compiled toolshed binary** from `build-binaries`, built with `COMMIT_SHA: ${{ github.sha }}`: the sha is baked into the binary's `COMPILED` file (served by `/api/meta`) **and** into the shell bundle the binary serves. Both sides match, the gate passes, identities compare equal, no signal, no banner. Locally the harness starts source servers (`start-local-dev.sh`) with no shas on either side — an environment split, not a race: red every local run, green every CI run. The same false banner also appeared in every `dev-local` shell session since the flip (covering the header there too).

## Fix — close the class, not the symptom

1. **piece:** skew is only reported when it is **proven** — both shas known and different. An unknown sha on either side returns the new `skipped-unknown-build` outcome silently (still no update; the light `?identity` is never trusted across unvalidated builds). "Nothing is provably newer" must not tell the user a newer version is available — in dev there is nothing to reload into.
2. **shell:** the banner is anchored to the **bottom** edge. Even a proven skew (e.g. every production deploy until clients reload) must not overlay header navigation; nothing interactive is shell-owned at the bottom edge.
3. **docs:** `pattern-updates.md` §version-skew-gate now distinguishes proven-mismatch (signal) from unknown-build (silent skip); `systemPatternAutoUpdate` / `systemPatternAutoUpdateHome` registered in `EXPERIMENTAL_OPTIONS.md` (missing since the registry landed — owed by #4611/#4619).

No timeout was touched.

## Verification (all on the same macOS machine)

| Run | Result |
|---|---|
| origin/main tip (`ddf495d9d`), pre-fix | **FAILED 1m3s**, hung at "Navigate back to space page (note 1)", `default-app.test.ts:486` — exact reported signature |
| green anchor `f1604f027` (parent of #4619) | ok, 2 passed (11s) — bisect anchors reproduced |
| unit red→green | new cases `skipped-unknown-build` + no signal: red before the fix (got `skipped-skew` + report), green after; existing proven-skew test unchanged |
| fixed branch | **ok, 2 passed (11s)**; note create-and-return 397ms. The `/api/meta` fetch still runs post-fix, so the fetch is exonerated — the banner chain was the blocker |
| post catch-up-merge (#4652) | integration oracle re-run green; piece/shell/runtime-client suites green |

---
## Summary by cubic
Stops the false “newer version available” banner in local dev and unblocks the default-app integration test. We only signal version skew on a proven mismatch, and the banner is now anchored to the bottom.

- **Bug Fixes**
  - `packages/piece`: signal `versionSkew` only when both shas are known and different; otherwise return `skipped-unknown-build` without signaling.
  - `packages/shell`: move the version-skew banner to the bottom to avoid covering header navigation.
  - Tests: add coverage for `skipped-unknown-build` and ensure no `?identity` fetch or signal on unknown builds.

- **Docs**
  - Document `systemPatternAutoUpdate` and `systemPatternAutoUpdateHome` in `EXPERIMENTAL_OPTIONS.md`.
  - Clarify the version gate: proven mismatch signals; unknown build skips silently in `pattern-updates.md`.
  - Align the implementation plan with the unknown-build skip and proven-mismatch-only signal in `system-pattern-updates-implementation-plan.md`.

<sup>Written for commit dba706c73988f12597549397b36c3c52f0bc897d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4653?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/development/EXPERIMENTAL_OPTIONS.md
+++ b/docs/development/EXPERIMENTAL_OPTIONS.md
@@ -180,8 +180,8 @@ propagate](#how-flags-propagate).
   `RuntimeOptions.experimental.eagerSourceAnnotation`. The ambient control point
   is `setEagerSourceAnnotation` in
   [`packages/runner/src/builder/module.ts`](../../packages/runner/src/builder/module.ts).
-  Unlike the other three, the runtime propagates this one only when it is set
-  explicitly, because the ambient flag is also a test seam.
+  Unlike the other env-backed flags, the runtime propagates this one only when
+  it is set explicitly, because the ambient flag is also a test seam.
 - **Added by.** gideon, in "make fn.src lazy/debug-only — re-root identity off
   .src" (#4458, 2026-07-06).
 - **Purpose.** Resolves the per-primitive debug source annotation (`fn.src`)
--- a/docs/development/EXPERIMENTAL_OPTIONS.md
+++ b/docs/development/EXPERIMENTAL_OPTIONS.md
@@ -31,6 +31,8 @@ was last checked against the code.
 | [`persistentSchedulerState`](#persistentschedulerstate) | `EXPERIMENTAL_PERSISTENT_SCHEDULER_STATE` env, or `RuntimeOptions.experimental` | off | Bernhard Seefeld (#3646) | graduate to always-on | implemented, off by default, rollout in progress |
 | [`commitPreconditions`](#commitpreconditions) | `RuntimeOptions.experimental` only (mapped `null` — programmatic-only — in the canonical env registry) | off | Bernhard Seefeld (#4090) | graduate with scheduler-v2 speculation lineage | implemented, off by default |
 | [`eagerSourceAnnotation`](#eagersourceannotation) | `EXPERIMENTAL_EAGER_SOURCE_ANNOTATION` env, or `RuntimeOptions.experimental` | off in production, on in shell dev builds | gideon (#4458) | permanent debug toggle, not slated for removal | implemented |
+| [`systemPatternAutoUpdate`](#systempatternautoupdate) | `EXPERIMENTAL_SYSTEM_PATTERN_AUTOUPDATE` env / shell build define, or `RuntimeOptions.experimental` | on in the shell (non-home roots); off server-side | Bernhard Seefeld (#4611; shell default-on #4619) | graduate to always-on, then delete both auto-update flags | implemented, on in the shell |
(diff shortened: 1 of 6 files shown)


seefeldb:
Both review findings addressed in dba706c73:
- **Codex P2** (stale live implementation plan): M1.2 / M3.1 / M-test wording in `system-pattern-updates-implementation-plan.md` now matches the shipped semantics — versionSkew only on a proven mismatch, `skipped-unknown-build` silent otherwise. (The plan is fully executed through M4; archiving it to `docs/history/` per the docs lifecycle is queued as separate housekeeping.)
- **cubic P3**: "Unlike the other three" → "Unlike the other env-backed flags" in the `eagerSourceAnnotation` section.

PR #4654, opened 2026-07-10
docs(fuse): FUSE-T / macOS NFS cache-semantics reference (+ harness)

bfollington:
**Docs-only.** Distills our FUSE-T cache-semantics investigation into a concise, forward-looking reference (`packages/fuse/CACHE_SEMANTICS.md`) plus the reproducible harness (`packages/fuse/experiments/cache-semantics/`). No runtime changes.

Rewritten from the earlier point-in-time "verify #4642's claims" doc into a live reference aimed at whoever next reaches for a cache-related FUSE mount flag or debugs a "file exists but the mount can't see it yet" report. Takeaways only — the blow-by-blow experiment log stays in git history / the harness.

Key durable points captured:

- This staleness is **macOS-only** (Linux kernel FUSE honors our `entry_timeout=0`).
- On macOS the caching is the **NFS client's**, not our bridge's — FUSE-T ignores the timeouts we send.
- **`-o noattrcache` is misleadingly named and version-dependent:** on FUSE-T **1.2.x** it maps to `nonegnamecache` *only* (attribute/directory cache untouched); on **1.0.x** it also zeroed the attribute cache. Verify with `nfsstat -m`, never the flag name. (Re-measured on 1.2.7 after this came up in review.)
- Two distinct staleness behaviors (negative-name vs. directory-listing) and which flag fixes which.
- Why `cf exec` uses a readdir fallback, the **latent zero-retry false-negative** still present on 1.2.7 (~1/14), and the version-independent fix (a single bounded retry on the readdir-miss path — left as a separate change).

Merging this lands the reference; the code-level bounded-retry fix is tracked separately.


bfollington:
Good catch — you're right, and rather than let this sit as a stale point-in-time doc I've rewritten this PR into a concise **live reference** on FUSE-T/macOS NFS cache behavior and our flag choices, folding in your 1.2.x correction. Merging it lands the doc; no runtime changes ride along.

I reran our harness on **FUSE-T 1.2.7** to pin it down instead of guessing:

- **Your mechanism is confirmed, directly.** `nfsstat -m` shows `-o noattrcache` flips **only** `negnamecache`→`nonegnamecache`; the attribute/directory cache (`acregmin/acdirmin=5, acregmax/acdirmax=60`) is untouched. On 1.0.49 it *also* zeroed the attr cache — which is exactly why the old doc (correctly stamped to 1.0.49) said "fixes both" with the mechanism "every stat reaches the daemon." That mechanism is wrong on 1.2.x. The doc now leads with "verify with `nfsstat -m`, never the flag name."

- **One nuance worth stating precisely,** since it cuts against a blanket "noattrcache doesn't close the window": for cf-exec's *stat-first* existence check, `nonegnamecache` still makes the initial stat authoritative, so on 1.2.7 noattrcache does still resolve *that* path (arm_a window 50s→0.2s; the false-negative case goes 6/6 resolved-via-stat). What it no longer fixes on 1.2.7 is generic directory-**listing** staleness (the `acdir` cache) — which I suspect is the window you measured. The doc now separates the two so they're not conflated.

- **The part that survives all of it:** the readdir-fallback false-negative is still latent on 1.2.7 (reproduced ~1/14), independent of `noattrcache`. So the robust fix remains a single bounded retry on the readdir-miss path, not a mount flag whose meaning has already shifted once between versions. Left as a separate change against the merged behavior — happy to send it.

Harness is kept live under `experiments/cache-semantics/` (it reproduced the 1.0.49↔1.2.7 delta in a couple of minutes). Review welcome on the rewritten doc.

bfollington:
This seems like it didn't actually matter in the end after more investigation. We can always come back to this if we need to, but we didn't end up finding any benefit.

PR #4659, opened 2026-07-10
feat(runner,data-model): computed-cell identity — computed-by-default classifier and the computed: URI scheme

ubik2:
Redesign of the computed-cell identity work previously drafted in #4553. The earlier design (provably-pure allowlist classifier, `fid2:`-tagged ids) is superseded; see that PR for lineage. The ack-and-drop write-conflict policy carries over with semantics unchanged.

Spec: `docs/specs/computed-cell-identity.md` (fully rewritten).

## What changed and why

**Classifier: computed by default.** Derived internal cells are classified computed unless a writer disqualifies them — even when their computation involves writes. The exhaustive disqualifiers: handler writers (incl. `writableProxy` proxies), streams, effect-marked modules, and **non-replayable builtins as direct writers only** (a sync computed *reading* a fetch result stays computed; recompute re-reads the stored result). Capture-writes, unverified captures, writable result surfaces, and sync builtins all qualify. Result-surface exposure no longer disqualifying is an accepted consequence recorded in the spec: an embedder's stale handler write through the exposed handle gets ack-and-dropped.

The input side replaces the old blanket capture check: a handler's `$ctx` bindings are walked against its argument schema ($ref-aware, conservative on any unmodeled shape), so read-only handler captures qualify while `asCell: ["cell"]` bindings disqualify. Sub-pattern and `map`/`filter`/`flatMap` input roots are conservatively disqualified (op sub-patterns may contain handlers writing elements); non-replayable builtins disqualify their input roots too (`llmDialog` writes through its `messages` input — the proof case).

**Async detection: a builder-side name registry** (`builder/builtin-replayability.ts`), fail-strict for unknown names. Deliberately not `isEffect`, which is incomplete on the fetch family and carries scheduler semantics.

**Id format: `computed:fid1:<hash>`.** The `computed:` URI scheme replaces `of:` for computed cells, mirroring the `data:`/`pattern:` multi-scheme precedent. The `FabricHash` tag stays `fid1`, and the **hash preimage is kind-free**: a computed cell and a state cell minted from the same cause share hash bytes and differ only in scheme — the full URI string is the identity. That relaxation (the kind is a conflict-policy hint under fail-open classification, not a purity guarantee) makes the identity rules load-bearing: never rebuild a computed URI from its bare hash, never key on scheme-stripped ids. Every keying/display site was audited accordingly: `cellRefToKey` keys on the full schemed id (and now JSON-encodes paths — `["."]` vs `["", ""]` collided under `.join(".")`), `CellHandle.id()` is a true identity accessor with the piece-root `of:`-strip moved to `PageHandle.id()`, and the display shorteners keep `computed:` visible as a marker.

**Observability**: client-side counted `commit-dropped-computed` debug log; server transact spans set `ct.droppedComputed` (parity with `ct.conflict`); the fuller telemetry story is the spec's open question.

## Census (12 sampled patterns, flag on)

483 derived internal cells: **253 computed (72.3% of the 350 non-stream cells)**; 133 handler event streams (never eligible); 97 genuine state cells — all of expected flavors (handler-written UI drafts, `wish` results, fetch/llm plumbing, sub-pattern/map input roots). `system/home.tsx` is the outlier (22/63) because its derived lists feed sub-patterns/map ops — the input-root conservatism is the biggest recoverable chunk if the ratio should go higher.

## Rollout

Gated behind `EXPERIMENTAL_COMPUTED_CELL_IDS` (default off), which gates **minting only**; readers accept both schemes unconditionally. Version skew: old servers see an unknown scheme ⇒ no kind ⇒ strict (safe); old **clients'** `fromURI` throws on `computed:` ids arriving via sync — the flag must not graduate until readers are fleet-wide.

## Verification

data-model 48, runner 832 (incl. reworked classification battery, registry cross-check, scheme round-trips, drop suites re-keyed, and a flag-on end-to-end: instantiation mints `computed:fid1:` links that an independent memory session reads back as ordinary docs), memory 343, ts-transformers 1023 (fixtures untouched), repo typecheck clean, `fid2` swept from code and docs.

Known pre-existing issue surfaced by the census (unrelated to this branch): `weekly-calendar` fails `cf check` with `Unsupported value type 'function'` at `event.tsx:368`.

---
## Summary by cubic
Adds the `computed:` URI scheme and a computed-by-default classifier for derived internal cells. Identity now uses the full schemed id end-to-end; routing/page intake is normalized, and the scheduler graph preserves entity schemes.

- **New Features**
  - Classifier: cells are computed unless written by handlers (incl. `writableProxy`), streams, effect‑marked modules, or non‑replayable builtins as direct writers; `$ref`‑aware input walk. Sub‑pattern/`map` roots and non‑replayable builtin inputs disqualify. Uses main’s `completeSchedulerScopeSummary`.
  - Builtins: name‑keyed replayability registry (fail‑strict for unknown names); marks `inspectConfLabel` non‑replayable.
  - Identity: new `computed:` scheme with kind‑free hash preimage. Parsers/specifiers updated (canonical helpers in `@commonfabric/data-model`). Keys use the full schemed id and JSON‑encoded paths. `CellHandle.id()` returns the full id; `PageHandle.id()` strips `of:` only; diagnostics and summaries keep schemes visible.
  - Plumbing: schemes preserved across LLM sandbox keys, fabric‑ref chains (no double‑scheme), wiki‑links, favorites, and scheduler graph (helpers/tests ensure scheme‑preserving extraction).
  - Page/nav: `PageGet` accepts bare/`of:` ids and rejects `computed:`; navigation emits bare pieceIds.
  - Notes/wiki‑links: embed contract enforced (`mentionIdFromCellId`/`bareMentionId`): strip `of:`, pass bare, reject `computed:`.
  - Tests: deterministic `memory v2` validation; e2e mint/read of `computed:` ids; broad parser/identity tests across runner, shell (incl. scheduler graph), UI, and state‑inspector.

- **Migration**
  - Gated by `EXPERIMENTAL_COMPUTED_CELL_IDS` (minting only); readers accept `of:` and `computed:`. Treat the scheme as part of identity; never strip or rebuild from bare hashes.
  - When adding builtins, list replayable names in `REPLAYABLE_BUILTIN_REFS`; unknown names default to non‑replayable.
  - Pages remain unkinded: URLs use bare pieceIds; the processor accepts bare or `of:` (rejects `computed:`). Mentions/wiki‑links persist bare ids and reject `computed:`; if computed cells become mentionable, the embed format must carry the scheme.

<sup>Written for commit f5c6c9ad7173884163e77530888669123d05707d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4659?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---BEGIN COPY-PASTE---
NEW_COVERAGE_BASELINE
---END COPY-PASTE---

--BEGIN COPY-PASTE---
NEW_PERF_BASELINE: job: Runner Tests (5/5) = 288s
---END COPY-PASTE---

According to my agent, the new performance baseline being slow is due to several slow tests in group 3 being moved to group 5 due to the addition of two more tests. Statistically, these outliers will go down as we add more tests.


ubik2 (commented):
_cf-review via Codex GPT-5, on behalf of @ubik2. Revalidated against `748980be3a9170030ebf3791753399ac4d508a53`._

The scope reduction successfully removes the ack-and-drop implementation and the prior phantom-state code bug. I still see three current-code blockers and one blocking issue in the live follow-up design.

## 🔴 Blocking

1. **`CellHandle.id()` changed to a full URI, but its bare-ID consumers were not migrated.**

   `packages/runtime-client/cell-handle.ts:98-107` now returns `of:fid1:...`. Existing page navigation (`packages/ui/src/v2/components/cf-cell-link/cf-cell-link.ts:384-390`) passes that value to `PageGet`, which still calls `entityIdFrom(request.pageId)` (`packages/runtime-client/backends/runtime-processor.ts:1208-1215`). Verified on this head: `of:fid1:AA` parses with tag `of:fid1` and is rebuilt as `of:of:fid1:AA`, so clicks can address a nonexistent document.

   The same contract break affects existing mention completion: `cf-code-editor.ts:743-772` stores `resolved.id()`, while `packages/patterns/notes/note-md.tsx:64-69` prepends `/of:`, producing `/of:of:fid1:...`.

   **Fix:** establish one central routing/wiki-link ID contract, migrate these callers, and add click-through plus existing-mention tests.

2. **Runner identity helpers erase the only distinction between state and computed cells.**

   `Cell.entityId` (`packages/runner/src/cell.ts:2572-2574`) and `getEntityId` (`packages/runner/src/create-ref.ts:154-170`) strip both schemes. Verified on this head:

   ```text
   source: computed:fid1:H
   entityId: fid1:H
   toURI(entityId): of:fid1:H
   ```

   Any caller round-tripping a computed cell through these public helpers silently addresses its `of:` sibling, contradicting the spec's load-bearing rule that the URI string is the identity.

   **Fix:** introduce/use a URI-preserving identity API; explicitly narrow bare `EntityRef` APIs and add `of:`/`computed:` sibling collision tests.

3. **`computedCellIds` leaks between Runtime instances.**

   Runtime construction writes a module-global flag (`packages/runner/src/runtime.ts:862-863`), while pattern construction reads that global (`packages/runner/src/builder/pattern.ts:523-529`). Verified with Runtime A off and Runtime B on: building under A after B was constructed still emitted `{ kind: "computed" }`.

   **Fix:** carry this option through the runtime/builder frame instead of ambient state, and add an opposite-flags multi-runtime test.

4. **The removed phantom-state behavior remains normative in the live spec.**

   `docs/specs/computed-cell-identity.md:334-365` still requires a dropped commit to satisfy `origin-committed` and resolve dependent pending reads as if it committed. That rule can persist authoritative output computed from an optimistic value that never became durable: resolving the pending read against the winner changes the read precondition, but not the already-computed write payload.

   This is no longer a current implementation bug, but it remains a blocking design defect for the follow-up PR. A dropped dependency must reject/retry its descendants, or use carefully defined cascade-dropping for computed-only lineage.

## 🟡 Improvements

- Add a stale-read computed-target integration test asserting today's strict `ConflictError`; the new E2E test covers only current reads.
- Rewrite `packages/runner/src/builder/pattern.ts:715-750`, `builtin-replayability.ts`, and classifier-test comments that still describe ack-and-drop as active behavior.
- Consider removing/defering `captureWritesAnalyzed`: it is serialized through builder modules but explicitly has no classifier/runtime consumer, causing broad transformer golden churn. If retained, correct the contradictory transformer comment at `lift-applied-strategy.ts:156-162`.
- Update the formal FabricHash spec and `packages/ui/docs/mentionable-internals.md` to match the new last-colon/full-ID behavior.

Current-head checks: focused data-model/runner/runtime-client tests passed (72 steps), `git diff --check` passed, and all three executable regressions above were reproduced directly.

**Verdict: request changes.**

ubik2 (commented):
_cf-review via Codex GPT-5, on behalf of @ubik2._

Follow-up clarification on `docs/specs/computed-cell-identity.md`: this is a documentation/design concern about the deferred Phase 2/3 behavior, not a request to implement those phases in this minting PR.

The "Server conflict policy" section currently says an acknowledged-and-dropped computed commit still satisfies `origin-committed` and resolves downstream `PendingRead.localSeq` references, while leaving the existing client pipeline untouched. The same behavior is repeated as a correctness invariant and in the Phase 3 test plan.

That can make the bookkeeping consistent without making a descendant write payload consistent:

1. Client creates optimistic commit A: `X = f(input_v1)`.
2. Before A is acknowledged, it creates B from pending X: `Y = g(X_v1)`. B's serialized operation already contains `Y_v1`.
3. The server has newer input and acknowledges-but-drops A.
4. B's pending read is resolved against the value that actually won, `X_v2`.
5. Resolving that read does not recompute B's existing payload, so B can appear to read `X_v2` while writing `Y_v1`, rather than the required `g(X_v2)`.

Since Phase 2/3 are explicitly deferred, a small resolution here would be to mark these lineage semantics as provisional rather than settled. The follow-up should require that a descendant derived from a dropped optimistic value be rejected, retried, recomputed, or otherwise proven independent of the dropped value. In particular, I would revise the assertions around lines 371-378 and 474-476, and make the Phase 3 test plan validate lineage safety instead of prescribing that every dropped origin satisfy downstream dependencies.

ubik2:
Addressed the lineage follow-up in e147b4347 (spec-only, as scoped):

- **Server conflict policy**: `localSeq` consumption stays firm (replay dedupe must work regardless), and everything else about dropped-commit lineage is now marked **PROVISIONAL**, with your scenario spelled out — B serializes `g(X_v1)` before A's drop is known, so resolving B's pending read against `X_v2` yields reads-v2-writes-g(v1). The section now requires the follow-up to reject/retry/recompute descendants of a dropped optimistic value or prove them independent (e.g. cascade-drop computed-only lineage, strict rejection otherwise), and notes the asymmetry: recompute paves over all-computed descendants, but an authoritative-state descendant persists the phantom.
- **Correctness invariants (~474)**: the bookkeeping-equivalence bullet no longer asserts `origin-committed`/`PendingRead` satisfaction; it keeps `localSeq` and flags the rest as unsettled pending the phantom-descendant closure.
- **Phase 3**: reframed from verifying the semantics to settling them, then validating lineage safety.
- **Test strategy**: the engine bullet validates whichever resolution the follow-up lands on rather than prescribing that dropped origins satisfy downstream dependencies.

The deferred phase-2/3 branch currently implements (and tests) the provisional semantics; the cascade/reject design and matching test changes land there.

PR #4663, opened 2026-07-10
Fix lunchpoll staging persistence and require registered pieces

willkelly:
## What changed

- preflight `cf piece new` against the real root `addPiece` stream before source resolution or durable creation
- require the normal registered-piece path in a named non-home space and remove the experimental home-space `--no-register` bypass
- register a created piece before optional slug assignment; preserve the durable piece ID and give safe inspection/recovery guidance on post-create failures
- hydrate explicitly requested scoped piece paths so one-shot CLI reads fetch the `PerUser` target instead of returning its default
- allow lunchpoll's explicit-name fallback when profile wishes are unresolved
- persist visit history through schema-bearing keyed entities, preserve opaque memberships during ordinary logs, and cap conflict/retry-safe history at 200 entries
- make explicit clear-all remove memberships without guessing or destroying unreadable target documents

## Why

The per-space Toolshed staging POC exposed four application/runtime failures that looked like lost state but were not ZFS failures:

1. home-space piece creation wrote a piece before failing registration
2. unresolved profile captures blocked manual join
3. cold one-shot CLI reads returned the default for a durable `PerUser` value
4. anonymously appended visit links could not be projected in a cold runtime

The POC proved that a slug-only home-space bypass was technically possible, but the selected contract keeps the system home out of the piece-registry role. Ordinary pieces belong to a named space with a registration-capable root.

## Impact

`piece new` aimed at a home space now fails before creation with named-space guidance. A successful named-space create is registered before its optional slug is assigned and remains discoverable if slug assignment fails. Because registration can commit before later settling fails, the CLI reports that outcome as unknown and prints a shell-safe inspection command instead of claiming the piece is absent.

The scoped-read, no-profile join, and keyed-history fixes remain unchanged. The staging canary pinned to `600d728445ab3db6a68ec81e26b2c660495e548a` is historical evidence for the storage/application lifecycle, not the final CLI bootstrap contract.

## Validation

- `deno task check` — 296 paths
- focused `packages/cli/test/piece.test.ts` — 36 steps
- `deno fmt --check` and `deno lint` on the changed CLI files
- registered named-space create/list/slug/cold-reload regression
- focused lunchpoll suite and cold/concurrent/opaque history integration from the earlier PR head
- `git diff --check`

## Coverage ratchet

`NEW_PERF_BASELINE: coverage-debt: packages/patterns uncovered lines = 241847 lines`

The six new uncovered `packages/patterns` lines are defensive migration branches in `logVisit`/`removeVisitEntity`. They require storage-level legacy, opaque, or 200-entry states that the pattern-unit idiom cannot fabricate; the end-to-end history integration exercises them, but pattern-runtime coverage is collected only from pattern-unit chunks.

### DIFF
--- a/packages/cli/commands/piece.ts
+++ b/packages/cli/commands/piece.ts
@@ -252,6 +252,11 @@ export const piece = new Command()
     "Root directory for resolving imports. Allows imports from parent directories within this root.",
   )
   .option("--slug <slug:string>", "Slug URL/address for this piece.")
+  .option(
+    "--no-register",
+    "Do not add the piece to the home piece list. " +
+      "Requires --slug and the home space.",
+  )
   .action(async (options, main) => {
     setQuietMode(!!options.quiet);
     const spaceConfig = parseSpaceOptions(options);
@@ -262,7 +267,11 @@ export const piece = new Command()
         mainExport: options.mainExport,
         rootPath: options.root ? absPath(options.root) : undefined,
       },
-      { start: options.start, slug: options.slug },
+      {
+        start: options.start,
+        slug: options.slug,
+        register: options.register,
+      },
     );
     render(pieceId);
     const browserPieceRef = options.slug ?? pieceId;
--- a/packages/cli/commands/piece.ts
+++ b/packages/cli/commands/piece.ts
@@ -252,11 +252,6 @@ export const piece = new Command()
     "Root directory for resolving imports. Allows imports from parent directories within this root.",
   )
   .option("--slug <slug:string>", "Slug URL/address for this piece.")
-  .option(
-    "--no-register",
-    "Do not add the piece to the home piece list. " +
-      "Requires --slug and the home space.",
-  )
   .action(async (options, main) => {
     setQuietMode(!!options.quiet);
     const spaceConfig = parseSpaceOptions(options);
@@ -270,7 +265,6 @@ export const piece = new Command()
       {
         start: options.start,
         slug: options.slug,
-        register: options.register,
       },
     );
     render(pieceId);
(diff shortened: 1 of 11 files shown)


willkelly:
Closing — roughly 1,050 of 1,248 lines are superseded, and main deliberately chose the opposite design in two places.

- **Scoped/PerUser cold read** (`piece/src/ops/piece-controller.ts`): main rewrote `PiecePropIo.get` in #5222 — `targetCell.key(...path)` then `selectedCell.pull()`, with `cellWithScopedLinkRequiredsRelaxed` (#4989) handling exactly the perUser/perSession case this names. #5256 is literally titled "drop the resolveAsCell routing from the narrow read" — main **deleted** the mechanism this adds.
- **Keyed visit history** (`patterns/lunch-poll/main.tsx`): #4922 replaced `visits.push(entry)` with a deterministic-id `visits.set([...currentVisits, entry])` and left a comment explicitly rejecting the mergeable-add design this builds on. Opposite design, chosen on purpose.
- **Join with unresolved profile**: main's `joinAs` already does the override-then-trim dance.
- #5318 moved the live Estuary lunch poll onto the mainline pattern, so the staging-persistence framing no longer matches reality.

**Salvage — worth re-cutting as a small fresh PR (~90 lines):**

`packages/cli/lib/piece.ts` plus a little of `packages/piece/src/manager.ts`: `assertCanAddPieces()`, the `getRaw()` false-positive guard, the home-vs-named-space preflight, register-before-slug ordering, and the post-create re-read. That is the genuinely unmet need here — small enough to land quickly against current main and not worth dragging 1,200 lines of superseded work along to get it.

PR #4668, opened 2026-07-10
fix(runner): verify source-doc closures per reachable view; name the element that voids an array read

mathpirate:
Fixes the platform bug behind the 2026-07-10 topics-dev board outage's first phase ("Could not load pattern" on estuary's overnight redeploy), plus a diagnosability fix for its second phase. Full forensic record on the board: topic ["Board outage 2026-07-10"](https://estuary.saga-castor.ts.net/topics-dev-476ea34f/fid1:ea3aFbIW7yT5pyTcZ8i5WoryiKh3kcs5y-_rtKjL_O0).

## Commit 1 — verify source-doc closures per reachable view

A stored source closure may legally hold several generations of the same ambient filename (`cfc.ts`): identities are entry-point independent, so a dependency doc is shared across seals and keeps whichever seal's `cf:cache-root` links wrote it last; a later entry's link walk then reaches sibling ambient generations. `verifySourceDocs` hashed the whole closure as one filename-keyed program, so one generation always shadowed the other → permanent mismatch → every runtime-version bump bricks such pieces on cold recompile (estuary's redeploy was the detector, not the cause). No data fix applies: the stored bytes are correct; the verifier was wrong.

Now each document verifies against its own authored-import-reachable view (root links excluded — never part of a Merkle preimage). Within a view, filenames stay unique by construction; a view violating that cannot be attributed to any emission and its root is rejected, while intact same-filename siblings still verify in their own views. The verify-failed warn names the offending identities/filenames instead of bare counts.

**Live-repro verified**: the original bricked board entry `ZOdGfc…` still in the topics-dev space (closure of 4 docs incl. two byte-intact `cfc.ts` generations) verifies `ok=true` under the fix; the pre-fix verifier reproduces `mismatches=1` side-by-side.

## Commit 2 — name the failing index and doc when an array element voids the read

Logging precision only; no semantics change. A schema-mismatched array element voids the ENTIRE array read for the caller, and the log line said only that "an item" mismatched — the outage's second phase (a blanked board) presented as silent data loss, and finding WHICH element and doc was at fault took probe archaeology. The log now carries the failing index + doc address; a test pins both the (unchanged) strict void and the breadcrumb.

## Deferred: the element-level generation-skew grace

An earlier revision of this PR also degraded generation-skew `required` failures on array elements (fields absent from older-generation docs read as omitted — never invented — instead of voiding the whole array). Per review discussion (thread + Discord), that contract change is **deferred**: loosening `required` on element reads risks becoming a foot-gun for contracts that rely on it, and the class is better addressed upstream. The full grace (both signatures + red/green tests) is parked whole on [`gideon/element-grace-parked`](https://github.com/commontoolsinc/labs/tree/gideon/element-grace-parked) if the upstream work wants it back.

Where the class is being addressed instead:
- [combineSchema manufactures closed-world additionalProperties — intended semantics or upstream bug?](https://estuary.saga-castor.ts.net/topics-dev-476ea34f/fid1:XVmEOgkcYE8UlIVYTtlx6scxGLpl9zv-G5ocLiWl1Xc) (board topic, for @ubik2 / @seefeldb)
- [Result-schema evolution: session affordances shouldn't brick old data — optionality and an authoring-time gate](https://estuary.saga-castor.ts.net/topics-dev-476ea34f/fid1:AdxT1VowiEELWrb8OPrnEM9bHDwGGoA0uCMgonEeE_8) (board topic; a hard gate on default-less required additions means declared defaults exist, and the read path already fills declared defaults before the `required` check)
- Agent-driven data repair via cf get/set for existing skewed data — already proven on the topics board (authorless-topic fixup, setsrc restore runbook).

## Verification

- Closure-verifier regression tests red-without/green-with in `cell-cache.test.ts` (the outage shape: two same-filename ambient generations across seals; plus corrupt dup-view rejection), and the live repro above.
- Array-void diagnosability pin in `traverse-element-grace.test.ts` (strict void unchanged + breadcrumb carries index/doc).
- Full runner suite green locally on the rebased tree (918 tests / 4867 steps, scheduler-v2 underneath); traverse-replay goldens unchanged.
- CI fully green on this shape (39 checks incl. all pattern shards, workspace typecheck, and the coverage-debt ratchet).

## Deploy note

An estuary redeploy bumps the runtime version → cold recompile → that is commit 1's trigger condition. Merging before the next redeploy closes the recurring brick threat for multi-generation closures.

### DIFF
--- a/packages/runner/src/compilation-cache/cell-cache.ts
+++ b/packages/runner/src/compilation-cache/cell-cache.ts
@@ -352,12 +352,28 @@ export interface SourceDocVerification {
  * checking it equals the identity the document is keyed by. This is the
  * content-addressed analog of the structural graph verifier: because the
  * identity is a one-way Merkle hash over `(source, import identities)`, a
- * single document's content does not determine its key — the whole closure
- * must recompute consistently. Tampering with any source, or rewiring an
- * import link, makes the recomputed identity diverge from the key.
+ * single document's content does not determine its key — the reachable
+ * closure must recompute consistently. Tampering with any source, or rewiring
+ * an import link, makes the recomputed identity diverge from the key.
  *
- * Extra unrelated documents in the set are harmless: a module's identity
- * depends only on its own reachable closure, so siblings do not perturb it.
+ * Each document is verified against **its own view**: the documents reachable
+ * from it over authored-import edges (root links excluded — a
+ * {@link ROOT_LINK_SPECIFIER} edge is never part of any module's Merkle
+ * preimage). One closure may legally hold several generations of the same
+ * ambient filename (e.g. two seals' injected `cfc.ts`, each root-linked by a
+ * different entry): identities are entry-point independent, so a shared
+ * dependency document keeps whichever seal's root links wrote it last, and a
+ * link walk then reaches sibling generations. Hashing the whole closure as
+ * one program would collide those filenames and permanently fail one
+ * generation; per-view recomputation verifies each against the emission
+ * that produced it. Within a single view, filenames are unique by
+ * construction (one program emission cannot contain two files at one path) —
+ * a view that violates this cannot be attributed to any emission, and its
+ * root document is rejected.
+ *
+ * Extra unrelated documents in the set are harmless to their siblings: a
+ * module's identity depends only on its own reachable view. Every document in
+ * the set is still verified (in its own view).
  */
 export function verifySourceDocs(
   entryIdentity: string,
@@ -369,25 +385,66 @@ export function verifySourceDocs(
     return { ok: false, mismatches: [], missing: [entryIdentity] };
   }
 
-  const files = [...docsByIdentity.values()].map((doc) => ({
-    name: doc.filename,
-    contents: doc.code,
-  }));
-  const recomputed = computeModuleHashes(
-    { main: entry.filename, files },
-    { runtimeFingerprint },
-  );
-
-  const mismatches: string[] = [];
   const missing: string[] = [];
-  for (const [identity, doc] of docsByIdentity) {
-    if (recomputed.get(doc.filename) !== identity) {
-      mismatches.push(identity);
-    }
+  for (const doc of docsByIdentity.values()) {
     for (const imp of doc.imports) {
       if (!docsByIdentity.has(imp.identity)) missing.push(imp.identity);
     }
   }
+
+  // identity → whether its recomputed hash matches its key. A verdict reached
+  // inside one view holds in every view (entry-point independence), so each
+  // document is hashed at most once.
+  const verdicts = new Map<string, boolean>();
+  for (const [rootIdentity, rootDoc] of docsByIdentity) {
+    if (verdicts.has(rootIdentity)) continue;
+
+    // The root's view: BFS over authored-import edges only.
+    const viewIds: string[] = [];
+    const seen = new Set<string>();
+    const queue = [rootIdentity];
+    while (queue.length > 0) {
+      const id = queue.shift()!;
+      if (seen.has(id)) continue;
+      seen.add(id);
+      const doc = docsByIdentity.get(id);
+      if (doc === undefined) continue; // dangling edge — recorded in `missing`
+      viewIds.push(id);
+      for (const imp of doc.imports) {
+        if (imp.specifier.startsWith(ROOT_LINK_SPECIFIER)) continue;
+        queue.push(imp.identity);
+      }
+    }
+
+    const files = viewIds.map((id) => {
+      const doc = docsByIdentity.get(id)!;
+      return { name: doc.filename, contents: doc.code };
+    });
+    if (new Set(files.map((f) => f.name)).size !== files.length) {
+      // Duplicate filename within one view: not attributable to any emission.
+      // Only the root is condemned — an intact member still verifies in its
+      // own (necessarily smaller) view when the loop reaches it.
+      verdicts.set(rootIdentity, false);
+      continue;
+    }
+
+    const recomputed = computeModuleHashes(
+      { main: rootDoc.filename, files },
+      { runtimeFingerprint },
+    );
+    for (const id of viewIds) {
+      if (!verdicts.has(id)) {
+        verdicts.set(
+          id,
+          recomputed.get(docsByIdentity.get(id)!.filename) === id,
+        );
+      }
+    }
+  }
+
+  const mismatches = [...verdicts.entries()]
+    .filter(([, ok]) => !ok)
+    .map(([identity]) => identity);
 
   return {
     ok: mismatches.length === 0 && missing.length === 0,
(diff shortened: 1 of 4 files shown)


seefeldb:
> An array element failing its schema voids the ENTIRE array read, and old-generation elements fail the required check two ways: their docs lack fields that didn't exist yet, and — one layer deeper — element links carry the writing generation's schema, whose closed-world additionalProperties combine makes unknown new fields false subschemas that stay required (statically unsatisfiable; this would re-fire on the NEXT schema evolution even against fully-upgraded docs behind stale links). One old element blanked arrays that also held valid entries; with the array's schema default that read as [] — "Topics (0)". Silent (info-level log).

I'm curious where the explicit `additionalProperties: false` came from. Unless the original schema had a `Record` with `[string]: never`, i.e. the author of the original schema explicitly disallowed adding fields, that shouldn't appear (and if that was the original intent, then maybe the current behavior is correct and we should fix this upstream -- really write that only if that's the intended behavior. it should be rare).

Similar, the new required field should have come with a default. Let's double check that there was none in the pattern, not that we accidentally lossed it. In the future this might be one to flag when updating a pattern!

mathpirate:
Good probes on both — chased them to ground (Fable here, posting with Gideon's blessing):

**Where the explicit `additionalProperties: false` came from: nowhere authored — it is manufactured at combine time.** No schema in play carries it: the schema-generator emits `additionalProperties: true` for empty object types and value-typed `additionalProperties` for index signatures, never `false` for a plain interface (and the topics sources contain none). The `false` subschemas appear inside `combineSchema` (traverse.ts): `parentSchema.additionalProperties ?? (parentSchema.properties === undefined)` — i.e. "has `properties`, no explicit `additionalProperties` ⇒ closed world," deliberately diverging from JSON Schema's open default per the surrounding comment. Fields one side never knew then combine against that manufactured `false`. Repro: `combineSchema({type:"object", properties:{title,editingBody}, required:["title","editingBody"]}, {type:"object", properties:{title}})` → `editingBody: false`. Provenance: the default shipped with the traversal unification (CT-997, #2198) — so "was this the intent" is a question for @ubik2, and per your suggestion we're taking it upstream as its own design topic on the board rather than deciding it here. If open-world wins there, signature 1 in this PR becomes reachable only via authored-unsatisfiable schemas (`Record<string, never>` and friends) — kept either way, since voiding an element for a requirement nothing can ever satisfy defers nothing; the topic will carry a revisit-note tying the two.

**Your question also exposed a real scope error in this PR's description (now edited): `required` INTERSECTS when both schemas carry it.** A stale link whose schema includes `required` drops the reader's new fields from combined-required — those reads pass with fields omitted, no grace needed. The unsatisfiable-required bite needs a link schema with `properties` but no `required` (hand-rolled cell schemas, not generated ones). The outage itself was the plain absent-key signature against the reader's full `required`.

**Defaults: verified none were lost.** topic.tsx authored the composer fields default-less — `commentDraft: Writable<string>`, `editingBody: boolean`, no `Default<>` anywhere in the #4643 additions — so the generated schema faithfully has no `default`, and the read-side defaults pass (which runs before the required check) had nothing to fill. Fully agreed on flagging this class at pattern-update time: a "new required output field without a default" warning would have caught #4643 at authoring. That, plus whether session affordances (drafts/streams/edit flags) should be optional-by-construction in result schemas, is going on the board as a design topic alongside the combine question.

bfollington:
Cross-referencing #4561, which works the same seam from the write side — one interaction worth noting, and one observation on semantics.

**The interaction:** this grace makes scope-isolation mistakes quieter at read time. An unresolvable narrower-scoped link and an old-generation absent field are indistinguishable to a reader (both resolve to undefined — the same reason B2 checks the target value rather than an error code), so the "older pattern generation?" breadcrumb here will also fire on scope bugs. #4561's write-time warn is then the only place that bug class self-identifies, so I'm planning to land it (warn-only) alongside this.

**On the semantics:** as far as I can tell we've never actually documented what a failed item validation should do to an array read — the whole-array void isn't in any spec and dates to the traversal unification (CT-997/#2198), same as the combine default already being taken upstream. Please correct me if this was settled somewhere I haven't seen. My own read of `items.required` is that it promises "every object present in this array has this field" — not that the array reads as empty when one item falls short — and what I'd ideally like the degrade to converge on is **fill, not omit**: an authored default when the schema has one, else a type-derived empty value (`""`, `0`, `false`…), so pattern code that trusts the type never hits an access error on a field the schema calls required. Notably this PR already does exactly that for cell-typed fields (the materialized live cell IS the safe empty value); plain fields are the remaining gap, and the unsatisfiable-`false` signature can only omit since there's no type to derive from. None of this blocks the PR for me — omission is strictly better than blanking the array, and commit 1 should go out ASAP regardless — but if the fill direction sounds right I'd rather we note it now as the target than discover the omit/fill difference later as another outage flavor.

*(posted via Claude Fable on behalf of @bfollington, after discussion with him)*

seefeldb:
Hmm, I don't think we should invent defaults (0 for number, "" for string, etc) when there aren't and/or omit array entries when undefined isn't part of the entry type.

The former works around the issues above which are either bugs (additional properties, being fixed) or can be solved at DX level (block updates when there are no defaults).

The latter just violated the typescript type and then we end up adding guards everywhere.

mathpirate:
Trying to pull the three threads here together (Fable again):

**Going forward, Berni's "block it" idea gets everyone what they want.** The whole problem started because #4643 added new required fields to TopicOutput (like `editingBody`) with no `Default<>`. If we make that a hard error at authoring time — "you added a required output field with no default; add one or make it optional" — then every new field that ships *has* a default. And the read path already fills in declared defaults before it checks `required`. So old topics would read back with `editingBody: false` filled in from the schema, which is exactly the "fill" behavior Ben is asking for — using only values the pattern author actually wrote, which answers Berni's objection. Nobody has to invent anything.

**Where I'd push back on fill: made-up values are worse than missing ones, because you can't detect them afterward.** If the runtime fills in `0` or `""` for a field the author never gave a default, that value looks exactly like real data from then on — no code can ever check "was this real or filler?". A missing field is at least checkable (`"editingBody" in topic` is false, reads give `undefined`). So when something has to give, I'd rather a hole you can see than a fake value you can't. (The live-cell behavior in this PR for `Writable`/`Stream` fields doesn't break this rule: the cell it hands back *reads* undefined — it gives you somewhere to look and subscribe, but it doesn't make up a value.)

**For the old data that's already out there, omission isn't a preference — it's the only option left.** Those pre-#4643 topics exist, with no defaults to fill from. When the board reads them, there are only three things the runtime *can* do: refuse the whole array (that's the bug — the board showed "Topics (0)" with 14 topics in it), fill with invented values (rejected above), or return the topic without the missing field. Yes, that means code touching other generations' data needs a guard — but patterns already write those guards everywhere (`asArray`, `?.`, `|| fallback`) because reads can be partial during sync anyway. And the situation before this PR wasn't type-safe either; it was the same missing field, plus the entire array vanishing.

**Ben's bigger find: none of this was ever written down.** There's no spec anywhere for "what should reading an array do when one item doesn't match the item schema" — blank-the-whole-array was never a decision anyone made, it just fell out of the traversal code (same era as the `additionalProperties` behavior Berni flagged, which is already headed to its own board topic). I'll file a topic to actually decide and document it; this PR's answer for array items — fill from declared defaults first, tolerate missing fields from older generations, still reject items whose *present* values are wrong — can be the starting proposal.

Two smaller things, now done rather than offered: the branch is rebased on latest main (scheduler-v2 underneath; full runner suite green locally, CI running on the push), and the log message now reads "older pattern generation, **or a scope-isolated writer**?" per Ben's point — a scope mistake and an old-generation gap look identical to a reader, so agreed that the write side (the B1-style warn) is where that mistake should announce itself, and +1 to landing it.

mathpirate:
Split done per the Discord thread (Fable): the element-grace contract change is **out of this PR** and parked whole — both signatures, red/green tests, revivable as-is — on [`gideon/element-grace-parked`](https://github.com/commontoolsinc/labs/tree/gideon/element-grace-parked). The degrade-semantics questions continue as board topics (combineSchema closed-world default; affordance defaults + a DX gate on default-less required additions; item-validation semantics were never specced).

What remains here: commit 1 unchanged (the closure verifier — the multi-generation `cfc.ts` brick, which no data fix can address), plus one small logging-precision commit per @seefeldb's "make it more precise to help debug": the array-void mismatch log now names the failing index + doc address. Semantics untouched — the strict void stays, and the new test pins both the void and the breadcrumb. One clarification for the record: the deferred grace omitted fields; it never invented default values — that idea was thread discussion, not the PR.

Rebased state: full runner suite green locally (918 tests on scheduler-v2 main); CI running on the push.

seefeldb (approved):

PR #4676, opened 2026-07-11
chore(memory): default ACL enforcement on

seefeldb:
## Summary

- default the toolshed MEMORY_ACL_MODE deployment dial to enforce
- retain explicit observe and off rollout overrides
- bootstrap blob-test home spaces through the signed storage path so the toolshed suite runs under real enforcement
- update live configuration, tutorial, and ingest prerequisite documentation

## Stack

This PR is intentionally stacked on #4670 and contains only the deployment-default flip plus tests and documentation needed to validate that flip. Merge #4670 first.

## Deployment note

Hosted cross-space ingest still requires the toolshed operator DID in MEMORY_SERVICE_DIDS, or equivalent per-space grants. The blob authorization design remains deferred; positive blob tests bootstrap an existing valid home-space ACL before exercising the current direct-write compatibility path.

## Validation

- toolshed: 63 tests / 217 steps with MEMORY_ACL_MODE unset, exercising the new enforce default
- blob route: 9 steps under enforcement
- all 534 documentation code blocks
- default-on named-space reload integration
- default-on profile, shared-profile, and inSpace integration flows
- focused runner, runtime-client, lib-shell, and shell task-ordering coverage
- lint, formatting, and diff checks

---
## Summary by cubic
Enable per-space ACL enforcement by default (`MEMORY_ACL_MODE=enforce`) with explicit ACL genesis for home and runtime‑derived named spaces so uploads work under enforcement. Fresh spaces are read‑only until genesis; legacy named spaces are temporarily authenticated‑public READ/WRITE; the last concrete OWNER cannot be removed; and shell/blob flows resolve names in‑worker and wait for genesis before uploads.

- **Migration**
  - For hosted cross-space ingest, add your operator DID to `MEMORY_SERVICE_DIDS` or provision per-space grants.
  - Use `MEMORY_ACL_MODE=observe` as a canary or `off` to bypass during rollout if needed.

<sup>Written for commit a1a5413a6bb2d206aa6f860b7cdb373a036cdc06. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4676?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mathpirate (approved):
looks great! had fable look at this and the original PR that re-implemented the mechanics and neither of us found any blocking issues. it asked me to pass along these two notes, which you mentioned in our in-person conversation too - just recording them for the public record:

Two rollout notes worth carrying into the deploy plan, neither blocking:

(1) Legacy-home window: between flipping enforcement on a deployment and each user's next home-open, populated ACL-less homes remain authenticated-public (READ+WRITE). It self-heals per user on first open; to close it proactively, an ops sweep signing owner-only genesis ACLs via MEMORY_SERVICE_DIDS would do it.
(2) Populated legacy named spaces have no ergonomic self-serve genesis path: the server (correctly) wants the space-identity signer or a service DID, but cf acl signs as the user. Name-capability holders hold the authority in principle — there's just no acl init-style plumbing that uses it yet. Worth folding into the ACL-management arc; it's also how we'd eventually harden existing dogfood spaces.

PR #4679, opened 2026-07-11
feat(scheduler): default persistent scheduler state on

seefeldb:
## Summary

Stacked on #4288. This follow-up flips persistent scheduler state to default-on while preserving explicit `false` as the rollback path.

## Changes

- Default persistent scheduler state to enabled in ambient runtime configuration.
- Preserve explicit `false` across runtime and shell configuration as the rollback.
- Treat invalid shell env values as unset, preserving the default instead of silently disabling it.
- Make the reload integration test prove that default-on rehydration actually occurred.
- Update tests, benchmark labels, comments, and live documentation for the rollout state.
- Remove now-redundant explicit enables from CI/integration launch paths.

## Validation

- Full Deno Workflow [run 29158269207, attempt 2](https://github.com/commontoolsinc/labs/actions/runs/29158269207): all required checks passed (38 passed, 5 expected deployment skips), including Performance Check.
  - CI was captured while the PR temporarily targeted `main`; the workflow is filtered to `base: main` and does not run for stacked PR bases.
  - The first sample had one noisy 23.4s pattern-unit outlier; the independent rerun was green at 17.4s. No new performance baseline was accepted.
- `deno task check`
- `deno fmt --check`
- `deno lint`
- Focused runner, memory, and shell tests: 97 tests / 65 steps passed.
- Default-unset reload integration: 2/2 passed; `rehydration.ok=42`.
- Explicit-false counterfactual: `rehydration.ok=0`, and the new default-on witness fails as intended.

## Performance data

Compared directly with #4288's exact head (`3545e75c6`).

### CI reload and suite signal

- Reload-to-render: 6017.6ms → 6067.3ms (**+0.8%**).
- Reload semantics were identical: 42 successful rehydrations, 16 no-match fallbacks, 13 action runs, and 10 computation runs.
- Runner aggregate: 225s → 210s (**-6.7%**).
- Pattern integration aggregate: 173s → 160s (**-7.5%**).
- Pattern reload job: 81s → 75s (**-7.4%**).

The job-level timings are single-sample runner measurements; the raw reload measurement and identical semantic counters are the more relevant rollout signal.

### Dedicated fixed-CPU benchmark A/B

Sequential runs with no overlap, both 331 benchmarks on Deno 2.8.1 / Xeon 8573C:

- [#4288 base run 29158286527](https://github.com/commontoolsinc/labs/actions/runs/29158286527): benchmark step 444s.
- [Default-on run 29158535403](https://github.com/commontoolsinc/labs/actions/runs/29158535403): benchmark step 465s.
- Whole benchmark step: **+4.7%**; median matched-benchmark drift: **+1.8%**.
- Default-app note create at 0 / 32 / 128 notes: **+24% / +49% / +80%**.
- Push/pull graph workloads: all 12 slower, **+15% to +210%** (median **+122%**).
- Wish-mentionable graph workloads: **+46% / +58%**.
- Event-preflight controls: **-2.2% to +2.5%** (median **+0.2%**).
- Direct persistent-state rehydration: **-1.0% to +9.2%** (median **+1.9%**).

Several long graph benchmarks have only two samples, so exact percentages need repetition, but the uniform direction and magnitude show a credible workload-specific cost in repeated reactive graph/commit paths rather than a broad scheduler/preflight regression.

### Targeted local A/B

Using the supported rollback `EXPERIMENTAL_PERSISTENT_SCHEDULER_STATE=false`, shared warm compile caches, and alternating runs:

- `auth-pieces`: 2.203s → 3.415s wall (**+55.0%**), with the same 206 action runs; default-on added 10 observation-only native commits.
- `parking-coordinator`: 32.647s → 36.423s wall (**+11.6%**), with the same 745 action runs; default-on added 109 observation-only native commits.
- Group-chat adoption: receiver scheduler starts 400 → 341 (**-14.75%**) and total starts 601 → 542 (**-9.82%**), while short-run wall time varied **+2% to +7%**.

The data shows the intended tradeoff clearly: persisted state can reduce scheduler recomputation after adoption/reload, while writing scheduler observations adds material cost to commit-heavy graph workloads. Explicit `false` remains available as the rollback.

### DIFF
--- a/.github/workflows/deno.yml
+++ b/.github/workflows/deno.yml
@@ -689,13 +689,20 @@ jobs:
       # (DENO_COVERAGE_DIR) but not authored-pattern coverage. Pattern coverage
       # is only collected by `cf test` via CF_PATTERN_COVERAGE_DIR, set on the
       # pattern-unit-test job. See docs/development/COVERAGE.md.
-      - name: 🧩 Run pattern reload integration tests
+      - name: 🧩 Run default-on pattern reload integration tests
         run: |
-          mkdir -p test-results
+          mkdir -p test-results/default
           HEADLESS=1 \
-          CF_EXPECT_PERSISTENT_SCHEDULER_STATE=1 \
           DENO_COVERAGE_DIR=coverage/raw/pattern-reload \
-          deno task integration --junit-dir=test-results patterns-reload
+          deno task integration --junit-dir=test-results/default patterns-reload
+
+      - name: ↩️ Run pattern reload integration tests with rollback override
+        run: |
+          mkdir -p test-results/rollback
+          HEADLESS=1 \
+          EXPERIMENTAL_PERSISTENT_SCHEDULER_STATE=false \
+          DENO_COVERAGE_DIR=coverage/raw/pattern-reload \
+          deno task integration --junit-dir=test-results/rollback patterns-reload
 
       - name: 🧾 Write coverage report
         if: always()
--- a/.github/workflows/deno.yml
+++ b/.github/workflows/deno.yml
@@ -693,7 +693,6 @@ jobs:
         run: |
           mkdir -p test-results
           HEADLESS=1 \
-          EXPERIMENTAL_PERSISTENT_SCHEDULER_STATE=true \
           CF_EXPECT_PERSISTENT_SCHEDULER_STATE=1 \
           DENO_COVERAGE_DIR=coverage/raw/pattern-reload \
           deno task integration --junit-dir=test-results patterns-reload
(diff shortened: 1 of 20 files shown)


Hixie:
Feedback from Opus:

> Doc wording contradiction (low). per-doc-rehydration.md says the rollback path is covered "rather than a second browser population," but the PR adds a second browser reload lane and home-rehydration-churn.test.ts says it "stays covered by the same browser suite." Reconcile the wording to "the same browser suite re-run with the flag off."

seefeldb:
Processed the remaining top-level wording comment. `d7b82799e` now says the rollback lane is the same browser suite re-run with persistence disabled, matching the actual second reload lane. This PR has no inline review threads; the stacked #4692 branch is rebased exactly onto this head.

seefeldb:
Superseded — closing.

`codex/server-execution-w1-2-shared-pool` (#4692) is descended from this branch, so all three commits here are already contained in it. #4692 has now been retargeted from this branch directly to `main`, which means merging it delivers this change along with the rest.

Keeping this open would leave #4692 stacked on a base that is 478 commits behind `main` and has not moved since 2026-07-13, for no benefit that survives the merge.

The `persistentSchedulerState` default-on change itself is unaffected and still reviewed as part of #4692's diff.

PR #4683, opened 2026-07-11
docs(specs): revise server-primary execution design

seefeldb:
## Summary

- revise the server-primary execution design around scheduler-v2 PR #4288 and make the execution-context-key collision a first standalone prerequisite
- specify a single user-sponsored worker per eligible active space, positive per-action authority, independent lease/claim generations, ordered settlement reconciliation, and reconnect claim snapshots
- keep compatible clients trusted in the first iteration while preserving the normal ACL/CFC/conflict/provenance path and deferring enforcement hardening
- move supported fetch/generate builtins to server execution through a preferred host broker, keep raw SES fetch blocked, preserve relative serving-origin calls, and constrain external egress
- replace the old plan with dependency-ordered red-green work orders, shadow-only rollout gates, a legacy background-worker exclusion, and explicit later phases for feed narrowing, client-compute suppression, scoped execution, and handler events

## Key decisions captured

- creation provenance is not producer identity; current producers come from the scheduler write index
- transformed computations have one direct root output binding; only hand-built tests need compatibility fixes
- server work always records the authenticated user it ran on behalf of; service identity is background-only
- whole action transactions are either server-servable or client-authoritative; they are never split
- committed overlays clear only after the matching accepted commit is locally applied; connection loss alone never returns client write authority

## Validation

- `deno task check-docs specs/server-side-execution`
- `git diff --check`
- two independent design audits plus a final post-fix consistency pass

---
## Summary by cubic
Revises and finalizes the server‑primary execution design on top of `scheduler-v2` (#4288), shifting to per‑action positive claims and one user‑sponsored worker per active space. Incorporates review feedback to clarify invariants, reconciliation, producer lookup, and egress for a safe, incremental rollout.

- **Refactors**
  - Key durable scheduler state and indexes by `SchedulerExecutionContextKey` to prevent user/session scope collisions.
  - Introduce `ExecutionClaim`s backed by a fenced per‑space `ExecutionLease` with lease/claim generations; clients fail open to local authority without a matching claim.
  - Run one unioned, user‑sponsored worker per eligible active space; exclude legacy background‑owned spaces until that registry is imported.
  - Use `scheduler_write_index` for producer lookup and enforce a single direct root output binding in the transformer.
  - Reconcile via `ActionSettlement` using `inputBasisSeq`; drop overlays only after the accepted commit; reconnect applies ordered claim snapshots.
  - Move `fetch*`/`generate*` to server execution with a preferred host egress broker; keep raw SES fetch blocked; preserve serving‑origin parity and constrain external egress.
  - Clarify trust boundaries, reconnect behavior, and rollout gates based on review.

- **Migration**
  - Phase 0: add context keys across snapshots/state/indexes, writer lookup, and the root‑binding invariant.
  - Phase 1: ship a shadow client‑demand executor with authenticated `ExecutionDemand`, fenced leases, and parked‑reader wake (no authority transfer).
  - Phase 2: enable `ExecutionClaim`s and client overlays with compatibility handshake; add `ActionSettlement` and a whole‑action scope firewall; move eligible builtins.
  - Later: fold background demand and narrow feeds; then add user/session‑scoped execution and dual handler execution.

<sup>Written for commit 2d7a09f92e6ff21af5ede3e3141b6a2bd54dd387. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4683?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/server-side-execution/README.md
+++ b/docs/specs/server-side-execution/README.md
@@ -151,8 +151,11 @@ server-side executor would know natively.
 - `background-piece-service` already runs a runtime per space in a **Deno
   Worker thread** (`packages/background-piece-service/src/worker-controller.ts:78`),
   discovered reactively from a registry cell, under a single service
-  identity. It is the seed of the executor pool (§6), currently limited to
-  a ~60s polling updater and websocket transport back to the same host.
+  identity. It is a lifecycle/isolation precedent for the executor pool (§6),
+  not its registry or discovery substrate: bps-owned spaces are excluded from
+  the client-demand pool until Phase 3 unifies their demand. Today bps remains
+  limited to a ~60s polling updater and websocket transport back to the same
+  host.
 - Server-initiated writes into spaces exist and pass CFC: webhook ingest
   (`POST /api/ingest/:id` with `externalIngestStamp`) and the sqlite
   builtin's server-executed query + result writeback.
@@ -273,8 +276,10 @@ Merkle-verified closure with `loadPatternByIdentity` +
 
 Creation provenance is not producer identity. A computation may redirect its
 output into a pre-existing cell whose creator is unrelated to the action that
-currently recomputes it. `source` remains useful for audit and causality, but
-must not select an executor action.
+currently recomputes it. Existing `source` metadata may remain useful as a
+local or historical diagnostic, but this plan neither adds universal source
+stamping nor promises a per-document causal-audit mechanism. A future audit
+lineage design is separate; creation source must not select an executor action.
 
 The authoritative producer projection is `scheduler_write_index`, joined to
 `scheduler_action_state` and the durable action snapshot. Lookup is by
@@ -298,14 +303,20 @@ than guessing a producer.
 
 Important patterns now go through the transformer. Enforce that every normal
 computation node has exactly one primary output binding, directly to the
-root path of its target cell. The runner must not recursively search an
-arbitrary returned object for a redirect to determine result identity.
-Additional static side writes and materializer envelopes remain explicit
-write surfaces and are indexed separately.
+root path of its target cell. The runner must not recursively search the
+emitted static output binding for a redirect to determine result identity.
+This is the `firstResolvedOutputRedirect` path; plain lift/computed result
+identity from `_resultFor` does not recurse and is not being replaced.
+Additional static side writes and materializer envelopes remain explicit write
+surfaces and are indexed separately.
 
 There is no legacy data migration or corpus audit. Update the small number of
 tests that hand-construct Pattern JSON, reject nonconforming new JSON at the
 runner boundary, and simplify the redirect-resolution code accordingly.
+Hand-built `type: "passthrough"` nodes have no transformer emission path but
+participate in the same output-binding contract: their primary binding must
+also be direct at the root, and nested/multiple aliases are fixed in tests or
+rejected.
 General Cell aliases remain supported; this invariant concerns the primary
 computation result binding, not every Cell redirect in the system.
 
@@ -364,7 +375,8 @@ approach comparison honest.
   explicit input basis and settlement acknowledgement (B) / handlers +
   derived, reconciled by event acks (C).
 - **Q4 — Executor placement and isolation:** none (status quo) / worker
-  thread per space co-located with the memory engine (§6.1) / subprocess
+  thread per active branch/space co-located with the memory engine (§6.1) /
+  subprocess
   tier for untrusted or heavy spaces (§6.6).
 
 ---
@@ -400,7 +412,7 @@ shadow mode has no server data writes or external effects, so it adds no
 writer. The explicit test capability may observe conflicts and exercises
 existing mutex guards without creating a production authority window.
 
-**Perf.** No client-side win. Server cost: one runtime per active space.
+**Perf.** No client-side win. Server cost: one runtime per active branch/space.
 Subscription serving cost unchanged.
 
 **Verdict.** A short shadow-validation milestone, not a product phase. It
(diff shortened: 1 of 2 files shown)


seefeldb (commented):
**cf-review via Claude Opus 4.8, on behalf of @seefeldb**

Scope: full rewrite of the server-primary execution spec + implementation plan (+1548/−1715). Coverage: both docs deep-read in full; load-bearing code premises verified against `main` and the `pr-4288` ref by three independent read-only verification passes; no tests run (docs-only change). Note: **§3.2.1 / W0.1 (scheduler context keys) is deliberately out of scope for this review** — that issue is being addressed directly in #4288; the premise paragraph in §3.2.1 will need a touch-up to match whatever lands there.

**Overall:** this revision is materially more rigorous than the doc it replaces, and its load-bearing corrections check out against code: `observedAtSeq` really is the accepting-commit sequence, not an input basis (stamped in `applyCommit`, `packages/memory/v2/engine.ts` ~`:3289`), so the `inputBasisSeq`/`ActionSettlement` redesign is warranted; the producer-lookup switch to `scheduler_write_index` is supported (all three write classes are indexed under a `write_kind` discriminator — `engine.ts:2837-2870`, DDL `:293`); and the W0.2 transformer premise is real (see Improvement 3).

## 🔴 Blocking

1. **`branch` is missing from the authority-identity shapes** (README §5.B.1 `ExecutionClaim`; plan W0.6 messages). This confirms and extends Codex's inline P1: `writersForTargets(branch, space, …)` (W0.3), the pool key `Map<branch+space, …>` (W1.2), and the lease are branch-qualified, and memory-v2 heads/watches are branch-scoped — but `ActionClaimKey`/`ExecutionClaim`, `ExecutionDemand`, and `ActionSettlement` carry only `space`. A claim or settlement minted on one branch can match the same action on another branch: wrongly suppressed client commits, or an overlay cleared against the wrong feed. Fix: add `branch` to `ActionClaimKey`, demand, claim set/revoke, settlements, and the reconnect claim snapshot; add a two-branch criterion to W0.6/W2.1. *(verified by inspection of the doc's own shapes plus memory-v2 branch scoping)*

## 🟡 Improvements

2. **Per-action `inputBasisSeq` is not transitive across claimed-action chains** (§5.B.4, W2.2). "Settlement has `inputBasisSeq >= S(L)` ⟹ the server has consumed that source basis" is sound for actions that read the client-written doc *directly* (single-transaction consistent-frontier argument) — but for a chain (action F reads intermediate I; I not yet recomputed from `S(L)`), F's max-basis can pass `S(L)` via its other inputs while I is stale → premature overlay drop → a transient stale confirmed value until I settles and F re-runs. Self-healing, and the divergence metric will count it, but the doc doesn't name it. Suggest one sentence documenting the accepted multi-hop transient — or scope the drop rule to actions whose read set includes the client-written doc.

3. **W0.2's "recursive search" wording over-reads — point it at the right code.** Verified: the recursion exists but walks the static **output binding**, not the lift's runtime returned value (`firstResolvedOutputRedirect`, `packages/runner/src/runner.ts:403-431`; used at `:3907` map/filter container minting, `:4248-4271` pattern nodes, `:1988` sub-pattern output walk). Plain lift/computed result identity doesn't recurse at all (`_resultFor` hash, `:3033-3038`). Tighten "recursively search an arbitrary returned object" → "recursively search the emitted output binding", and add those three call sites to W0.2's read-first list so the implementer simplifies the right thing.

4. **W0.2 must classify `type:"passthrough"`.** It has no builder/transformer emission site — it appears only in hand-built tests, several with nested `{key: {$alias}}` outputs. State explicitly whether the one-root-binding invariant covers passthrough (then those fixtures join the fix list) or excludes it.

5. **Sponsor consent/visibility** (§6.1/§5.B.7). A user's authenticated session silently becomes the `onBehalfOf` authority for space-wide derived work — including work caused by *other* users' commits. B.7 nails the semantics (execution authority ≠ semantic authorship), but nothing addresses user-facing surprise or consent. Suggest adding a §10 open question: sponsorship opt-out, or owner-only sponsorship in v1.

6. **Stale line in §2.3:** bps "is the seed of the executor pool (§6)" — but §6.1/W1.2 deliberately do *not* build on its registry ("lifecycle patterns only; do not make its registry the primary discovery path") and exclude bps-owned spaces via the interlock. One-line touch-up.

7. **Handler observations: state the scope explicitly.** W0.4's read-first cites "event dispatch transaction provenance", but no WO says whether handler runs persist scheduler observations at all (today the attach is wired only at the action-run seam). Under this design handlers are never claimed and wake doesn't key on handler reads, so either answer works — say which, so an implementer neither wires it by accident nor skips it by omission.

## Questions for the author

- **Universal source stamping is now explicitly dropped** (W0.3: "Do not add universal source stamping or walk from a target through its creator"). The producer-lookup half of that reversal is well-argued and code-confirmed (creation ≠ current producer: computations can alias output into pre-existing cells — `packages/runner/test/runner.test.ts:515-570`). But the earlier direction had a second purpose beyond producer lookup: audit/causality ("for each doc, find out how it was computed"), and the write index only covers actions that have *run* under persistence. If the full drop is intended, consider also removing or resolving the residual "`source` remains useful for audit and causality" line (§3.3) — as written it half-implies a mechanism no work order builds.

## Verified-correct (no need to re-check)

`docs/development/EXPERIMENTAL_OPTIONS.md` and `deno task check-docs` exist as the plan assumes; #4495 is merged, so "landed" is accurate; capability negotiation has an established protocol-flags precedent to plug into (`docs/specs/memory-v2/04-protocol.md:121-129`, `persistentSchedulerState` as the model); §3.2's inline-dirtying claim matches code exactly (`findSchedulerReadersForWrite` `engine.ts:1849`, `markSchedulerReadersDirtyForWrites` `:1912`, called from `applyCommitTransaction` `:3276`; no parked-space batched query exists); the G0–G17 and W-number cross-references are internally consistent throughout both docs; and the superseded vocabulary (`session.interest.set`, `executorConfig`, `unservablePieces`, space-wide `derivedAuthority`) is fully purged except deliberate "there is no X" statements.

**Verdict:** high-quality deep revision — recommend merge after the branch-identity fix (Blocking 1) and whichever improvements the author endorses. The doc's honesty about what v1 does *not* deliver (N× client compute remains; first paint unchanged; enforcement deferred) is a strength, not a weakness.

PR #4687, opened 2026-07-12
feat(pattern-construction): add first-class serializable factories

seefeldb:
## Dependency

None. This branch is merge-synced with current `origin/main`; the former #4925 dependency is already present in main.

## Summary

- add directly callable, serializable `Factory@1` values for pattern, module/lift, and handler factories
- add inert decode plus runner-owned warm/cold materialization, dynamic replacement fencing, private pattern params, and `FrameworkProvided` authority preservation
- automatically publish complete factory artifacts into the destination space atomically with their containing write
- preserve synchronous optimistic visibility: warm artifacts prepare synchronously; cold artifact fetch delays only wire submission, and later commits remain causally ordered behind it
- make equal pre-existing artifacts an idempotent `ensure` no-op while rejecting mismatched content-addressed data
- migrate canonical list, LLM tool, CLI, FUSE, piece, and production pattern writers
- complete the pre-launch compatibility cleanup and archive the executed implementation plan
- keep detached synthetic CFC schemas metadata-only when TypeScript has erased a writer alias's `typeof binding`, so inferred structure cannot shadow a destination cell's complete trusted schema
- preserve exact named capture schemas when semantic types degrade, while keeping authored unknowns/generics closed; materialized array spreads, retained-element copy methods, and retained payload arguments stay full-shape without widening modeled projections such as `map`

## New usage

Factories are ordinary typed values. A factory passed to an eager pattern input stays symbolic and reactive; the transformer lowers its invocation to the dynamic runner node:

```tsx
const Search = pattern<{ query: string }>(({ query }) => ({
  answer: str`Result for ${query}`,
}));

const SearchHost = pattern<{
  search: PatternFactory<{ query: string }, { answer: string }>;
  query: string;
}>(({ search, query }) => ({
  result: search({ query }),
}));
```

Factories passed to lifts or handlers arrive as regular materialized callables. Cold code loading delays only that lift attempt or handler event:

```tsx
const invoke = lift(({
  factory,
  input,
}: {
  factory: ModuleFactory<{ value: number }, number>;
  input: { value: number };
}) => factory(input));
```

Factories can be passed directly as LLM tools, or wrapped only for metadata:

```tsx
const search = pattern<{ query: string }>(({ query }) => ({ answer: query }));

generateText({
  prompt,
  tools: {
    search,
    documentedSearch: {
      pattern: search,
      description: "Search the current index",
    },
  },
});
```

## `patternTool` before and after

Before, callers merged closure data into the public tool input through `extraParams`:

```tsx
const tool = patternTool(
  pattern<{ query: string; entries: Entry[] }>(({ query, entries }) =>
    searchEntries(entries, query)
  ),
  { entries },
);
```

After, use a readable inline pattern closure. The transformer captures `entries` as private callback argument 1 state, emits `withPatternParamsSchema`, and binds it once with an internal `.curry(...)`. Public input remains only `{ query }`:

```tsx
const tool = pattern<{ query: string }>(({ query }) =>
  searchEntries(entries, query)
);
```

Authored code never calls `.curry`; it is transformer-only and absent from the public API.

## Validation

Local verification across the implementation series, with the final checklist
refreshed after merging current `origin/main` at `475aa549cc` (branch head
`9b79fd053d`), using Deno 2.9.4:

- repository-wide `deno task check`: `Type check complete.`
- data-model: `56 passed (3149 steps)`; schema-generator: `61 passed (426 steps)`; ts-transformers: `1336 passed (980 steps)`; all `0 failed`
- runner: `1519 passed (9416 steps)`, `0 failed`, `0 ignored (1 step)` on the final merged head
- API: `29 passed (7 steps)`; data-model-schema: `10 passed (326 steps)`; js-compiler: `16 passed (163 steps)`; LLM: `8 passed (68 steps)`; all `0 failed`
- CLI: parallel `1827 passed (2933 steps)`, IPC arm `1 passed (3 steps)`, and serial real-Git lane `218 passed (684 steps)`; all `0 failed`
- piece: `63 passed (999 steps)`; memory: `625 passed (603 steps)`; runtime-client: `31 passed (657 steps)`; FUSE: `389 passed (18 steps)`, one platform-specific ignore; all `0 failed`
- background-piece-service: `14 passed (55 steps)`; patterns: `64 passed (246 steps)` plus source-coverage and browser arms; static: `68 passed (33 steps)` plus 24 browser tests; debug-view: 14 tests; all `0 failed`
- fresh-runtime integration: `2 passed (23 steps)`; it persists, cold-decodes, materializes, and invokes pattern, module/lift, and handler factories through a new runtime while verifying inert decode before runner materialization
- the newly landed lunch-poll read-budget directory passes all six assertions on exact main and on this branch; both canonical 1,184-vote updated renders measure 194,223 accesses. One earlier branch run observed an additional root pull and exceeded the budget, but exact-main, focused-branch, and same-process branch reruns did not reproduce it; no budget or runtime code was changed
- representative summary-index transformed output contains callback argument 0 as `__cf_pattern_input`, private `{ entries }` at argument 1 under `withPatternParamsSchema`, one internal `.curry({ entries })`, and one-argument `mapWithPattern(...)`; it contains no `patternTool` or `extraParams`
- docs checks: pattern-construction 1 block, plans 2 blocks, and common 160 blocks passed; history index covers 299 documents with 217 entries
- package cycles, unused dependencies, single-copy dependencies, Deno pins, append-only baselines, pattern tiers, command/completion docs, local-program construction, conflict markers, control characters, skill facts, and integration polling gates passed
- repository formatting: 6,171 files; lint: 5,956 files; working tree and `git diff --check`: clean

## Provenance

Merge-synced without history rewriting onto current `origin/main` at
`475aa549cc` and pushed through `9b79fd053d`. The final merge retained main's
value-visitor container recursion and runner lifecycle fixes while preserving
factory-leaf dispatch, generation/readiness fencing, and browser teardown work.
The data-model-owned factory wire protocol remains independent of the higher
API package.

## Compatibility cleanup

The repository is pre-launch and persisted data may be wiped. Stage 5 removes `patternTool`, `extraParams`, sibling list params, and `$patternRef` readers/writers after the owner-approved wipe decision and supported-source inventory gate.

`$implRef` remains because it is current implementation identity inside instantiated execution-module descriptors; it is not a Factory value representation. Direct factory-function `toJSON()` still emits the full compatibility graph until every Fabric boundary is proven to use registered codec dispatch. Canonical Fabric writers use `Factory@1` and never emit `$patternRef`.

## Coverage ratchet

This foundational migration intentionally adds validated factory protocol, transformer, runner, and boundary code across the affected packages. The behavioral suites above exercise the required semantics; the remaining increase is mostly defensive validation and compatibility/error surface. Per `docs/development/CI_PERFORMANCE.md`, accept only the measured package metrics below rather than using the broad `NEW_COVERAGE_BASELINE` reset.

The current-base coverage comparison also reports one uncovered
`packages/connectors` line in a file unchanged by this PR. The generated narrow
override below accepts that exact baseline-run fluctuation without broadening
any feature-owned allowance.

| Source group | `main` | This PR | Delta |
| --- | ---: | ---: | ---: |
| `packages/api` | 215 | 215 | 0 |
| `packages/cli` | 2424 | 2467 | +43 |
| `packages/connectors` | 639 | 640 | +1 |
| `packages/data-model` | 22 | 137 | +115 |
| `packages/data-model-schema` | 12 | 60 | +48 |
| `packages/fuse` | 3609 | 3659 | +50 |
| `packages/memory` | 620 | 635 | +15 |
| `packages/patterns` | 26017 | 26039 | +22 |
| `packages/piece` | 170 | 170 | 0 |
| `packages/runner` | 5433 | 6086 | +653 |
| `packages/runtime-client` | 551 | 551 | 0 |
| `packages/schema-generator` | 590 | 593 | +3 |
| `packages/ts-transformers` | 1325 | 2007 | +682 |
| `tasks` | 1795 | 1928 | +133 |

```text
ACCEPT_COVERAGE_DEBT: packages/cli +43 lines
ACCEPT_COVERAGE_DEBT: packages/connectors +1 line
ACCEPT_COVERAGE_DEBT: packages/data-model +115 lines
ACCEPT_COVERAGE_DEBT: packages/data-model-schema +48 lines
ACCEPT_COVERAGE_DEBT: packages/fuse +50 lines
ACCEPT_COVERAGE_DEBT: packages/memory +15 lines
ACCEPT_COVERAGE_DEBT: packages/patterns +22 lines
ACCEPT_COVERAGE_DEBT: packages/runner +653 lines
ACCEPT_COVERAGE_DEBT: packages/schema-generator +3 lines
ACCEPT_COVERAGE_DEBT: packages/ts-transformers +682 lines
ACCEPT_COVERAGE_DEBT: tasks +133 lines
```


Hixie:
Feedback from Opus:

---

**Union of two factory contracts is double-materialized and hard-errors on a valid value.**
In `materializeScheduledFactoryInputs`, the compound-branch loop (`packages/runner/src/factory-input-preparation.ts:407-430`) visits *every* branch whose `schemaContainsFactory` is true, threading `result` through each. For a `oneOf`/`anyOf` whose branches are two different factory contracts, a value that matches branch A is materialized against contract A, and then branch B re-visits the now-live factory and calls `materializeFactory` against contract B, which throws *"Factory materialization schema mismatch"* at `assertExpectedContract` (`packages/runner/src/factory-materialization.ts:211`). Neither guard discriminates by contract: `canFactoryBranchMatch` skips its const/enum/properties checks for a non-record value, and the `asFactory` skip only fires when the value is *not* an admitted factory. Realistic trigger: an input typed `PatternFactory<A,B> | HandlerFactory<C,D>`. Suggest picking the single matching branch for `oneOf`/`anyOf` (or short-circuiting once a factory branch materializes) rather than folding across all of them.

**`schemaContainsFactory` memo can cache a false negative under mutual recursion.**
`schemaContainsFactory` pre-seeds the memo to `false` to break `$ref` cycles (`packages/runner/src/factory-input-preparation.ts:269`) and then overwrites it. Under mutually recursive schemas, a node whose factory-containment is only established through a branch evaluated *after* a cycle-broken re-entrant lookup can end up permanently memoized `false`. When that happens, `visit` skips descending and delivers a nested factory to the callback as an un-materialized cold shell — with no `FactoryArtifactUnavailableError`, so no `RetryWhenReady` and no retry. Standard fix: don't cache a result that was computed while a dependency was assumed in-progress. A mutually-recursive-schema test with a factory reachable only through the recursive arm would pin this down.

**Is `prepareFactoryWithReadinessRetries` reachable in production, and does it mask a missing re-park?**
The readiness retry loop (`packages/runner/src/factory-input-preparation.ts:47`) only retries when the caught error carries `readyToRetry` (`:37-45`). But the factory load path doesn't produce such an error: `prepareFactory` throws `FactoryArtifactUnavailableError` when `loadArtifactByIdentity` returns `undefined` (`packages/runner/src/factory-materialization.ts:553`), and that class has no `readyToRetry` (`:67`). `readyToRetry` is attached only on storage *commit* rejections (`packages/runner/src/storage/v2.ts:2930`, `:2945`); the load path only reads and aborts, so it never attaches one. If that's right, then for a cold artifact the loop throws on attempt 0 and `MAX_FACTORY_READINESS_ATTEMPTS`/`FACTORY_READINESS_BACKOFF_MS` are never exercised — and the two tests that appear to cover them (`packages/runner/test/factory-input-materialization.test.ts:1037` and `:1102`) reach them only by mocking `loadArtifactByIdentity` to reject with a synthesized `readyToRetry` error the real loader never emits.

Two asks:
1. If the loop is indeed unreachable, either delete it (and the mock-only tests, which also depend on real `setTimeout` sleeps) or make it reachable on purpose. As written it reads as transient-failure handling that doesn't actually run, which is misleading for the next reader.
2. More importantly, the case it *looks* like it should cover isn't covered anywhere: an artifact ref that is visible before its artifact is committed. That returns `undefined` → `FactoryArtifactUnavailableError` → the readiness `Promise.all` rejects → the scheduler treats it as a hard authored error (`packages/runner/src/scheduler/action-run.ts:582`, `packages/runner/src/scheduler/events.ts:844`). When the artifact later lands, no subscribed input value changes (only the backing artifact arrives), so the failed action is never re-triggered. Can a content-addressed factory ref be observed before its artifact commits (e.g. under a concurrent writer / cross-space copy)? If yes, this should re-park on the artifact's arrival via `RetryWhenReady` rather than hard-fail. If the runtime guarantees artifact-before-ref, a comment saying so would settle it.

Either way, the general shape worth leaning on is `RetryWhenReady`: it's uncapped, event-driven, generation-fenced, and sleep-free, so readiness gaps are better expressed by re-parking on a real signal than by an in-line bounded loop with a backoff.

(No defect found on the event path itself: exhausted readiness surfaces the error once and runs the final commit callback once, guarded against double-processing at `packages/runner/src/scheduler/events.ts:803-861`.)

### Authoring diagnostics

These matter because a large share of authors — including LLM coding agents — arrive with the *pre-removal* mental model (`patternTool` + `extraParams`) or stray from the doc's examples, and the current messages don't bridge them to the new API. I verified each by compiling a minimal pattern on this branch; the fixes are mostly "name the replacement in the error."

1. **`extraParams` in a `generateText` tool → verbose union error, no migration hint.** Repro: `generateText({ prompt, tools: { search, extraParams: { foo: 1 } } })`. Gives *"Object literal may only specify known properties, and 'extraParams' does not exist in type 'PatternFactory<any, any> | { pattern: …; description?: … } | … 10 more … | …'"*. The `… 10 more …` elision and `AnyCell` noise hide that `extraParams` was removed, and nothing points to the inline-closure replacement. Highest-frequency mistake by far.

2. **A typed arrow where a factory is expected → an `AnyCell`/`FactoryInput` "Property 'x' is missing" tree that never says "wrap it in `pattern()`".** Repro: pass `(input: { query: string }) => ({ answer: input.query })` into a `PatternFactory<{query:string},{answer:string}>` slot. The whole error argues about parameter structure (`FactoryInput<{query}>` not assignable to `{query}` because `AnyCell` lacks `query`) and steers the author to "fix" the callback signature, when the actual fix is to construct a factory with `pattern()`/`lift()`.

3. **`import { patternTool }` → "Module … has no exported member 'patternTool'."** Correct but a dead end — no pointer to the inline-closure replacement. A deprecation-style message covering both this and `extraParams` ("`patternTool`/`extraParams` were removed; capture closure state in an inline `pattern(...)` and pass the factory directly — see docs/common/concepts/factories.md") would resolve 1 and 3 together.

4. **`Default<>` on a child's input silently breaks assignment to a `PatternFactory<{field: T}>` slot.** Repro: `pattern<{ query: Default<""> }, { answer: string }>(…)` used where `PatternFactory<{ query: string }, { answer: string }>` is expected. `Default<>` narrows the input to the default literal, so the slot rejects an arbitrary `string`, and the error bottoms out seven levels deep in a `DefaultMarker<"">` mismatch that never names the cause. Given the pattern docs push `Default<>` everywhere, this will bite. Worth a targeted message or a doc note on factory slots + `Default<>`.

5. **Wrong factory *kind* is never named.** Annotating a pattern's own input as `ModuleFactory` when it's only ever a pattern compiles clean until a concrete value is wired, and even then the message is structural-schema noise rather than "a pattern was used where a module is expected." Low priority (hard to catch at the type level), but no diagnostic ever states the kind mismatch.


- **Unserializable closure capture throws a raw internal stack trace with no source location.** A captured local *function* gives a clean SES message, but a captured `Map` (and other type-specific rejections) throws an unformatted schema-generator → ts-transformers stack trace, because `packages/ts-transformers/src/closures/strategies/pattern-strategy.ts:234` only converts the `"No formatter found for type:"` branch into the clean, source-located `pattern-capture:unrepresentable-schema` diagnostic and re-throws the rest. Route the other schema-gen rejections through the same clean path. (Partly pre-existing: a `Map` in a plain pattern output crashes the same way.)
- **Internal work-package codes leak into shipped errors.** FrameworkProvided capture violations emit user-facing `severity:"error"` text like *"…cannot carry trusted framework obligations before WP3.6."* (`packages/ts-transformers/src/closures/strategies/pattern-strategy.ts:305-321`). "WP3.6" is meaningless to an author; drop or reword.

packages/runner/src/scheduler/events.ts
```diff
@@ -627,6 +706,25 @@ export async function dispatchQueuedEvent(state: {
...
+  // not authored retry policy: retries:false still waits and re-runs. Spreading
+  // the queued value deliberately preserves its id, registration generation,
+  // callback, retry flag, lineage origin, and any existing retry window state.
+  const requeueAfterReadiness = (): boolean => {
+    if (!isQueuedEventRegistrationCurrent(queuedEvent)) return false;
+    const requeued: QueuedEvent = {
+      ...queuedEvent,
+      notBefore: undefined,
+    };
+    state.eventQueue.unshift(requeued);
```
ubik2: 🔴 **Blocking — same-stream events execute out of order when parked on a cold artifact** (verified, executed repro)

`requeueAfterReadiness` requeues with `eventQueue.unshift()`. When two events park on the same cold factory (shared single-flight load), their readiness continuations run in dispatch order and each unshifts — so the second event lands in front of the first. Repro executed on this branch: events "A" then "B" queued to the same handler were processed as `["B", "A"]`. For non-commutative events (two writes to the same field) the final state is wrong.

The existing coverage in `scheduler-readiness.test.ts` (~line 263) only parks one event per stream, so this case is unguarded.

**Suggested fix:** stamp a monotonic `seq` on every `QueuedEvent` at queue time and have `requeueAfterReadiness` insert before the first queued event with a larger `seq`, instead of blind `unshift` — preserves head priority over newly-pushed events while keeping same-readiness batches in original order.

_cf-review via Claude Fable 5, on behalf of @ubik2_

packages/runner/src/builtins/map.ts
```diff
@@ -339,6 +386,51 @@ export function map(
...
+        });
+        // The result binding was established in this transaction. It will be
+        // aborted with RetryWhenReady, so force the retry to establish it again
+        // against the same deterministic container identity.
+        result = undefined;
+        throw new RetryWhenReady(
+          resumeRowsReadiness,
+          "map: resumed row patterns are waiting for durable state",
+          { keepDependenciesWhileWaiting: false },
+        );
```
ubik2, line 470: 🔴 **Blocking — one transient failure during `map` resume permanently kills the map** (verified, executed repro)

This pre-sync chain has no rejection handler, so on rejection `resumeRowsReadiness` permanently holds a rejected promise. Three things then strand the action for good:

- the scheduler's readiness-rejection path (`action-run.ts:582-593`) calls only `handleError` — no re-dirty, no re-queue;
- the action parked with `keepDependenciesWhileWaiting: false`, so its dependency set was replaced with an empty log — input changes and pulls can't re-trigger it;
- even a hypothetical re-invocation hits the guard at `map.ts:155-166`, which rethrows `RetryWhenReady(resumeRowsReadiness)` with the already-rejected promise.

Repro executed on this branch: with `syncCellsForPatternResume` patched to reject exactly once (transient failure), the map action never ran again despite an input change and 20 pull attempts — output stayed stale until restart. Control run without the fault passes.

**Suggested fix:** give the chain a rejection path that clears the parked state and settles the readiness so the scheduler's fulfillment path re-runs the action and retries the pre-sync fresh, e.g. `.then(() => { resumeRowsReadyKey = readyKey; }, (error) => logger.warn(...)).finally(() => { resumeRowsReadiness = undefined; })`. (`filter.ts`/`flatmap.ts` don't share this construct.)

_cf-review via Claude Fable 5, on behalf of @ubik2_

packages/schema-generator/src/formatters/factory-formatter.ts
```diff
@@ -0,0 +1,239 @@
...
+    const inputSchema = this.schemaGenerator.generateSchema(
+      detected.inputType,
+      context.typeChecker,
+    );
+    const outputSchema = this.schemaGenerator.generateSchema(
+      detected.outputType,
+      context.typeChecker,
+    );
+    return asFactorySchema(detected.kind, inputSchema, outputSchema);
+  }
```
ubik2, line 241: 🔴 **Blocking — recursive factory types crash the build** (verified by probe)

`formatContract`/`formatDetected` call `this.schemaGenerator.generateSchema(inputType/outputType, ...)` as a fresh top-level generation without threading the object-hoisting cycle-breaker context, so a factory whose input/output transitively contains the same factory recurses forever:

```ts
interface Rec { self: ModuleFactory<{ x: string }, Rec> }
```

throws `Maximum call stack size exceeded` at compile time — a hard toolchain crash on a valid (if unusual) type.

**Suggested fix:** propagate the visited/cycle state across the factory boundary, and add a recursion test (none exists today).

_cf-review via Claude Fable 5, on behalf of @ubik2_

ubik2 (commented):
_cf-review via Claude Fable 5, on behalf of @ubik2_

**Scope:** 655 files, +52.3k/−9.9k across 7 packages · **Coverage:** scope-and-theory with five delegated domain deep-reads (data-model protocol, runner lifecycle, transformers, boundary writers + memory, tests/docs/hygiene) plus a verification pass that executed repros for the two runner findings · **Checks run:** full data-model suite on the branch (passes), transformer end-to-end probes, two executed bug repros, CI status (41/41 pass, 4 expected skips), all 26 prior review threads confirmed resolved.

The overall architecture holds up remarkably well for a change this size: the `Factory@1` protocol is fail-closed, decode is genuinely inert, hashing preserves type identity, `.curry` privacy is airtight (verified by probe), the memory `ensure` path is atomic and ACL-covered, migration off `patternTool`/`extraParams`/`$patternRef` is complete with CI guard tests, and the docs lifecycle was followed correctly. Anti-duplication is clean — llm-dialog even *deleted* its historical private fork. But two confirmed runtime bugs in the cold-artifact machinery need fixing before merge.

## 🔴 Blocking

Posted as inline comments, summarized here:

1. **Same-stream events execute out of order when parked on a cold artifact** (`packages/runner/src/scheduler/events.ts:720`) — verified with an executed repro: events "A" then "B" processed as `["B", "A"]`.
2. **One transient failure during `map` resume permanently kills the map** (`packages/runner/src/builtins/map.ts:414-430`) — verified with an executed repro: after one injected transient pre-sync rejection, the map never re-ran despite input changes; output stayed stale until restart.
3. **Recursive factory types crash the build** (`packages/schema-generator/src/formatters/factory-formatter.ts:178-219`) — verified by probe: `interface Rec { self: ModuleFactory<{x: string}, Rec> }` throws `Maximum call stack size exceeded` at compile time.

## 🟡 Improvements

**Trust-model narrowing at the LLM boundary needs an explicit sign-off** (three related findings). Old code overwrote *any* declared `sandboxId` from the framework; new code only handles compiler-marked `FrameworkProvided` paths (`packages/runner/src/builtins/llm-dialog.ts:1316-1347`), so an *unmarked* `sandboxId: string` on a factory tool is now model-suppliable — the old comment called caller-chosen ids a cross-instance leak vector. Relatedly, the deleted "fails closed when no stable entity id" test case has no replacement even though the throw survives (`packages/runner/src/framework-provided-inputs.ts:79`), and the old loud rejection of author-pinned sandboxIds became a silent overwrite. If "compiler metadata is the sole trust root" is the intended model, please say so here; either way, restoring the fail-closed guard test would be good.

**`factorySchemasEqual` fails open on `__proto__` keys in browsers** — verified by probe. `packages/data-model/src/schema-utils.ts` normalization builds results via `normalized[key] =` on `{}` literals; safe in Deno (which deletes the `__proto__` accessor at bootstrap) but in browsers a schema key `"__proto__"` silently sets the prototype and is dropped, so two schemas differing only there compare equal — and this feeds runner materialization/compat decisions. Build into `Object.create(null)` or use `Object.fromEntries`.

**`isFabricCompatible()` returns true for factories that conversion then rejects** — verified by probe. `packages/data-model/src/native-conversion.ts:511` accepts a pre-ref live factory that `fabricFromNativeValue` throws on (`"artifact ref is not available"`), breaking the guard-implies-convertible contract that every other arm upholds and contradicting the spec text this PR adds in `docs/specs/space-model-formal-spec/1-fabric-values.md`.

**Factory provenance recorded against the wrong space after a cross-space link hop** — `packages/runner/src/cell.ts:2178` notes `this.link.space` instead of the resolved link's space, so publication offers the wrong `fromSpace`, converting a valid write into an async commit rejection (fail-closed, but a false failure). The list-op path in `list-factory-materialization.ts` already captures the resolved space correctly.

**Wire-send turn hygiene in `packages/runner/src/storage/v2.ts` (`commitNativeSend`)** — the ordering and failure-release paths check out, but: a `schedulerFlushPrerequisite` failure can leave `wirePreparation` as an unhandled promise rejection; a throw before the `try` orphans the turn and wedges the space's wire pipeline; and a never-settling preparation stalls all subsequent sends with no diagnostic. Suggest a no-op `.catch` at construction, `try/finally` around turn release, and a logged watchdog.

**FUSE decode paths** — `packages/fuse/mod.ts:1469-1477` shares one try/catch across `JSON.parse` and factory-projection decode, so valid JSON containing a bad `fvj1:` tag is silently delivered to the handler as a bare string (the `.json` path correctly returns EINVAL) — catch only around `JSON.parse`. Also, in-band `fvj1:` string tagging in `packages/fuse/callables.ts` means echoing back a legitimately-read value containing such a string corrupts or fails the write; probably unavoidable for a file interface, but deserves a comment and a test.

**`detectFactoryType` is spoofable by name** — `packages/schema-generator/src/formatters/factory-formatter.ts:50-78` matches alias names and un-provenance-checked `__@FABRIC_FACTORY_TYPE` property names. The security-critical consumers (SES capture exception, FrameworkProvided detection) independently check `isCommonFabricSymbol` — verified safe — but five other call sites use it bare, and nothing stops the next caller from forgetting the guard. Suggest a provenance-required mode plus a negative test that a user type merely named `PatternFactory` is not granted factory trust.

**Smaller verified items:**
- The LLM tool cold path still parks on bare `prepareFactory` while the scheduled-input, direct-supervisor, and list-op paths got `prepareFactoryWithReadinessRetries` (`packages/runner/src/builtins/llm-dialog.ts:955`) — align it.
- Supersession detection by error-message string comparison (`packages/runner/src/runner.ts:2723` matching the message thrown in `factory-materialization.ts:459`) — make it a typed error.
- This PR adds two (arguably three) hand-rolled JSON-Pointer `$ref` resolvers with divergent edge behavior (`packages/data-model/src/schema-utils.ts` `resolveLocalSchemaRef`, `packages/runner/src/factory-input-preparation.ts:162` `resolveFactoryDiscoveryRef`, `packages/runner/src/builder/factory-params.ts:24` `schemaAtRef`) — consolidate on one, and fix the shared split-before-decode quirk (`%2F` in a key is misread as a segment separator; RFC 6901 is split-then-decode).
- `packages/patterns/deep-research/deep-research.tsx` wraps `searchWeb`/`readWebpage` in schemaless `pattern<{query: string}, any>` closures, discarding the field descriptions and typed result schemas the LLM previously saw — pass the factories directly, as omnibox-fab does.
- Stale live docs: `docs/specs/pattern-imports/implementation-plan.md:707` still instructs a future executor to verify `$patternRef` behavior this PR removes; `docs/specs/pattern-construction/overview.md:200` lists work this PR executes as "(Proposed)".

**Suspected, needs author confirmation:** reactive array-method callbacks reject locally-defined factory captures (`ses-callback:callable-capture`) while nested `pattern()` callbacks accept them via the `permitsFactoryCaptures` exception, despite sharing the curry-lowering path (`packages/ts-transformers/src/policy/callback-boundary.ts` + `pattern-context-validation.ts:1119-1209`) — intended scoping or gap?

## Questions for the author

- Is model-suppliable `sandboxId` on unmarked factory tools an accepted consequence of the compiler-metadata trust root?
- The runner coverage-debt baseline (8,139 uncovered lines) is close to the 8,620 non-test lines added — could you state per-package old→new deltas in the PR body so the ratchet override is auditable? (The narrow per-metric form used is the correct one per `CI_PERFORMANCE.md`.)

## Possible follow-ups (proposals, not asks)

- Extract the warm/cold artifact stubbing harness and the `withTimeout` helper copy-pasted across ~21/18 new runner test files into `packages/runner/test/support/`.
- Unify the five parallel recursive factory/cell walkers in runner (`runtime.ts` `collectFactoryArtifactIdentities`, `cell.ts` `assertFactoryArtifactsPublishableForWrite` / `prepareFactoryStatesForWrite` / `deepTraverse`, `link-utils.ts`) behind one traversal with leaf hooks — two already disagree on `FabricInstance` recursion.
- Single-flight dedup for `prepareArtifactPublication` cold loads (`packages/runner/src/pattern-manager.ts:311-355`) — correct today, wasteful under concurrency; `ensureArtifactClosureInSpace` already has the in-progress map pattern.
- Server-level tests pinning that a mismatched `ensure` rejects the whole wire commit and that ensures are ACL-denied (engine-level coverage exists; server-level doesn't).

## Verdict

Impressive, well-tested foundational work with clean protocol design and honest migration — but **hold merge for the two confirmed cold-artifact bugs** (event reordering, permanent map wedge; both have executed repros and small fixes) and the schema-generator recursion crash; the sandboxId trust narrowing needs one explicit sign-off.

seefeldb:
Latest adversarial/Opus feedback is addressed across `c76df1f23` and `929b48d19`.

- `c76df1f23`: preserves FIFO when events re-park, recovers map resume after a transient pre-sync failure, terminates recursive factory schema generation, selects one matching factory-union arm, avoids recursive containment false-negative memoization, replaces timer/backoff artifact loading with event-driven `RetryWhenReady`, and improves the migration/capture diagnostics.
- `929b48d19`: makes schema comparison prototype-safe and RFC 6901-correct, consolidates the local-ref resolver, restores `isFabricCompatible`/conversion parity, records resolved cross-space artifact provenance, hardens causal wire-turn cleanup and fast rejection handling, keeps malformed FUSE factory tags explicit, requires Common Fabric symbol provenance for transformer factory trust, uses a typed supersession signal, passes the deep-research tools directly, and updates the live specs/docs for each behavior.

Trust-model sign-off: compiler/artifact `FrameworkProvided` metadata is intentionally the sole trust root. An unmarked `sandboxId` is ordinary authored/model input; it gains no framework authority. Missing stable framework identity still fails closed and now has an explicit regression test.

I did not add a timeout watchdog to causal artifact preparation: a valid content load may remain pending, so a timeout would turn readiness into a false failure. The concrete rejection/unhandled-promise/orphaned-turn hazards are fixed without weakening causal ordering.

Current local validation on `929b48d19`: full data-model, schema-generator, ts-transformers, runner, LLM, API, CLI, FUSE, and patterns suites pass; root `deno task check`, docs checks, transformed-output inspection, the three-kind fresh-runtime cold round trip, formatting, and `git diff --check` all pass. The PR body will be refreshed with exact post-push coverage ratchets after Performance Check reports them.

packages/data-model/src/native-conversion.ts
```diff
@@ -500,6 +508,17 @@ function isFabricCompatibleInternal(
     }
 
     case "function": {
+      if (isAdmittedFabricFactory(value)) {
+        try {
+          // Conversion seals the atom through the same path. In particular, a
+          // trusted live builder factory is not yet a Fabric-compatible value
+          // while its complete content-addressed artifact ref is unavailable.
+          sealFactoryState(value);
```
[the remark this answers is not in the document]
seefeldb: Fixed in c1bd939f7. `isFabricCompatible()` now seals through the same `deepFreeze` hardener as conversion. The regression uses mutable `UnknownValue` state inside canonical factory params and proves the probe accepts and freezes it; the formal Fabric-value spec now states this parity explicitly. Full data-model suite: 49 passed (2002 steps), 0 failed.

packages/ts-transformers/src/transformers/symbolic-factory-call.ts
```diff
@@ -0,0 +1,367 @@
...
+      (contract): EmittedFactoryContract => ({
+        kind: contract.kind,
+        inputSchema: contract.inputSchema as JSONSchema | undefined ??
+          generated.inputSchema,
+        outputSchema: contract.outputSchema as JSONSchema | undefined ??
+          generated.outputSchema,
+        frameworkProvidedPaths: contract.frameworkProvidedPaths ?? [],
+      }),
+    );
+    return requireCompatibleContracts(contracts, context, diagnosticNode);
```
mathpirate, line 188: **[P2] Generate provenance fallbacks per union arm**

`generated` is derived from `first` and then reused for every provenanced contract. With two compiler-owned pattern wrappers that have identical inputs and protected paths but outputs `{ child: { a: number } }` and `{ child: { b: string } }`, this produces no schema-mismatch diagnostic and emits an `invokeFactory` contract with `resultSchema: true`, even though the root schema retains the two distinct `asFactory` contracts. The runtime’s exact Factory@1 comparison will consequently reject either selected factory. Please generate missing schemas from each contract’s own input/output type, or from the correctly paired semantic member, before calling `requireCompatibleContracts`.

packages/ts-transformers/src/ast/factory-callee.ts
```diff
@@ -0,0 +1,481 @@
...
+    if (
+      ts.isArrowFunction(current) || ts.isFunctionExpression(current) ||
+      ts.isFunctionDeclaration(current)
+    ) {
+      if (isReferencedScheduledCallback(current, checker)) {
+        return "runtime-materialized";
+      }
+      if (ts.isFunctionDeclaration(current)) {
+        current = current.parent;
+        continue;
```
mathpirate: **[P2] Preserve scheduled exposure through helpers**

Continuing past an ordinary function declaration discards the execution boundary at its call sites. A typed module-scope helper such as `invoke(factory, value) { return factory({ value }); }`, when used only from a `lift` callback, still reports `factory-call:untransformable-symbolic-proxy` even though `factory` is runner-materialized there. Please propagate exposure through helper invocations or analyze the helper’s call sites, and add a regression for a typed helper called from `lift`.

packages/ts-transformers/src/transformers/framework-provided.ts
```diff
@@ -0,0 +1,1334 @@
...
+  const visit = (node: ts.Node): void => {
+    if (reported || node !== callback.body && ts.isFunctionLike(node)) return;
+    if (ts.isCallExpression(node)) {
+      const paths = frameworkPathsForFactoryExpression(
+        node.expression,
+        context,
+        new Set(),
+        new Set(),
+      );
+      if (paths.length > 0) {
```
mathpirate: **[P2] Do not treat factory modifiers as invocations**

This validates every `CallExpression` as though it invokes a factory. For `input.factory.asScope('space')` or `.inSpace(...)`, `frameworkPathsForFactoryExpression` deliberately follows the modifier receiver and therefore reports `scheduled-callback:framework-provided-factory-call`, although the code is only deriving another factory and supplies no authored factory arguments. Please gate this validation on an actual factory invocation or explicitly exclude modifier derivations.

packages/ts-transformers/src/transformers/schema-injection.ts
```diff
@@ -2743,6 +4399,45 @@ function handlePatternSchemaInjection(
...
+      );
+    }
+    if (carriedContract?.inputSchema !== undefined) {
+      return factory.updateCallExpression(
+        node,
+        node.expression,
+        node.typeArguments,
+        [
+          builderFunctionArg.expression,
+          createSchemaAst(carriedContract.inputSchema, factory),
```
mathpirate, line 4554: **[P2] Preserve an own `__proto__` schema property**

Re-emitting an exact carried schema through `createSchemaAst` converts an authored computed `['__proto__']` key into the object-literal form `__proto__: {...}`. JavaScript interprets that as a prototype setter, so the emitted schema has no own `__proto__` property and silently loses the contract entry. `createSchemaAst` needs to special-case this key as a computed property name; a string-literal name is also subject to the prototype-setter syntax. Please add a regression that checks `Object.hasOwn(emitted.properties, '__proto__')`.

mathpirate (dismissed):
**Unattended review notice:** Gideon explicitly asked me to post this review while he is away at lunch; he has not personally reviewed this final submission yet and will look at it later.

**Update:** Gideon (I) have now read this and the comments below seem good to me.

Re-reviewed at `308e67b03`. The trusted-source changes fix my earlier concern about locally authored types named `PatternFactory` receiving factory semantics—thank you.

I still see three previously reported correctness issues, plus one new issue in the provenance-union fallback logic. Each has a focused reproduction and is described inline. The schema-generator and ts-transformers suites, package linters, repository type-check, and diff check are green, but these adversarial cases are not currently covered, so I’m requesting changes pending those fixes.

mathpirate:
Following up from a transformers/spec pass, I think there are four places where we should tighten the implementation or make the language boundary explicit:

1. **Same-kind provenanced unions:** the call-site fallback currently derives schemas from the first union member and can therefore make distinct factory contracts appear compatible after widening. We should preserve the fail-closed exact-contract rule: generate each contract from its own provenance/type, compare normalized complete contracts, and diagnose when the member-to-contract correspondence is ambiguous or the schemas differ.

2. **Factory modifiers:** scheduled validation currently risks classifying `.asScope()` / `.inSpace()` as factory invocations. These are contract- and provenance-preserving derivations. We should centralize factory-callee classification (invocation vs derivation vs non-factory/ambiguous) and have both symbolic-call lowering and FrameworkProvided validation consume it.

3. **Exact schema keys:** emitting an ordinary `__proto__: value` property can mutate the generated object's prototype instead of preserving an own schema key. We should emit `["__proto__"]` (and audit intermediate normalization/copy paths for prototype-safe maps), with a regression that verifies the own key and unchanged prototype for a nested factory contract.

4. **Referenced helper exposure:** a helper used only from scheduled `lift`/handler contexts should see materialized factories even when it is declared at module scope; moving the helper must not change semantics. Mixed symbolic + materialized call sites are a genuine V1 design boundary, so my recommendation is to diagnose mixed/unknown exposure and ask authors to split or inline the helper rather than introduce automatic specialization yet.

These preserve the feature's existing core norms—context-driven exposure, exact executable contracts, and fail-closed ambiguity—rather than relaxing the specs to match implementation accidents.

We will undertake a follow-up CL to add to this PR implementing these recommendations.

mathpirate:
Implemented the agreed transformer hardening in `b57751402`:

- compare same-kind provenanced unions using each member's own exact schemas;
- centralize invocation/derivation classification so `.asScope()` / `.inSpace()` preserve provenance without triggering invocation validation;
- emit `__proto__` as a computed own schema key and regression-test the evaluated object's prototype;
- propagate factory exposure through stable helper call sites, follow scheduled helpers for FrameworkProvided validation, and diagnose mixed/unknown helper exposure without V1 specialization.

The feature/current-behavior docs now record the helper policy and diagnostic.

Validation: transformer format, type-check, and lint pass; full `@commonfabric/ts-transformers` suite passes — **1,254 passed (737 steps), 0 failed**.

packages/ts-transformers/src/ast/factory-callee.ts
seefeldb: Fixed in 13225290cb. Factory exposure now propagates through ordinary helper call sites, with regressions for scheduled-only, eager-only, mixed, and privileged helper contexts. The complete ts-transformers task passes.

packages/cli/lib/view/languages/typescript/semantics.ts
```diff
@@ -621,6 +621,11 @@ function discoverConfig(
         if (!isLocalSpecifier(value)) continue; // jsr:/npm:/https: → leave as any
         importMap[key] = isAbsolute(value) ? value : join(dir, value);
       }
+      for (const [key, value] of Object.entries(workspaceExports(raw, dir))) {
+        // An explicit import map entry, including one inherited from a nearer
+        // config, has the same precedence it has in Deno.
+        if (!(key in importMap)) importMap[key] = value;
```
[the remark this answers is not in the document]
seefeldb: Fixed in 106e405d7a. Explicit npm:/jsr: import-map entries are now retained before workspace exports are applied, so a workspace package only fills an actually absent key. The regression is in packages/cli/test/view-semantics.test.ts; the final complete CLI run passed (parallel 1,824/2,893 and sequential 218/684).

packages/schema-generator/src/schema-generator.ts
```diff
@@ -192,15 +265,28 @@ export class SchemaGenerator {
...
+    const unreliableUnknown = !!recoveredUnknownReference &&
+      (recoveredUnknownReference.flags & ts.TypeFlags.Unknown) === 0;
+    if (!unreliableAny && !unreliableUnknown) {
       return false;
     }
 
     // Check if this is a wrapper type - if so, use type-based analysis
     const wrapperKind = detectWrapperViaNode(typeNode, checker);
-    if (wrapperKind) {
+    if (wrapperKind || resolveScopeWrapperNode(typeNode)) {
```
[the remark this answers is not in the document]
seefeldb: Fixed in 106e405d7a. The method documentation now names PerSpace, PerUser, PerSession, and PerAny alongside the other wrappers whose apparent type is unreliable. The complete schema-generator task passed (60 tests / 394 steps).

packages/cli/lib/view/languages/typescript/semantics.ts
```diff
@@ -621,6 +621,11 @@ function discoverConfig(
         if (!isLocalSpecifier(value)) continue; // jsr:/npm:/https: → leave as any
         importMap[key] = isAbsolute(value) ? value : join(dir, value);
       }
+      for (const [key, value] of Object.entries(workspaceExports(raw, dir))) {
```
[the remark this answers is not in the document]
seefeldb: Fixed in 106e405d7a. Config discovery now selects the effective file per directory using Deno precedence (deno.json before deno.jsonc) instead of merging both. The sibling-config regression is in packages/cli/test/view-semantics.test.ts; the final complete CLI run passed.

packages/fuse/cell-bridge.ts
```diff
@@ -3932,22 +3939,48 @@ export class CellBridge {
...
+          ? readWithoutFactoryMaterialization.call(childCell)
+          : childCell.get?.();
+        childReadSucceeded = true;
         // Override with the raw link reference only for sigil links, which
         // is what enables FUSE symlinks.
         const rawValue = childCell.getRaw?.();
         if (isSigilLink(rawValue)) {
           childValue = rawValue;
+        } else if (
+          ContextualFlowControl.getAsCellValues(childSchema).length > 0
```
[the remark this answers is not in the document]
seefeldb: Fixed in 4beee2c6fa and carried through the current-main merge. FUSE now resolves asCell depth through both anyOf and oneOf branches (including schemas carrying both keywords), dereferences exactly the declared number of Cell boundaries, and leaves the payload query view intact. Added red/green coverage for direct, anyOf, oneOf, combined-union, and over-dereference cases; the full FUSE package passes.

PR #4692, opened 2026-07-13
feat(execution): move claimed execution to the server

seefeldb:
## Summary

- add authenticated, branch-qualified execution demand, durable policy, lease,
  claim, settlement, accepted-commit, and reconnect-frontier control paths on
  top of persistent scheduler state
- run one shared isolated Deno Worker per demanded branch/space lane, fenced by
  user-backed `onBehalfOf` authority, with indexed wake, drain, restart,
  hibernation, sponsor handoff, and durable legacy-background exclusion
- route exact claimed client actions through speculative overlays and ordered
  settlement/data barriers while preserving fail-open client authority for
  missing, revoked, scoped, cross-space, or otherwise unservable claims
- execute supported `fetch*` and `generate*` builtins through a server broker;
  raw fetch remains blocked in SES, relative paths retain the serving origin,
  and absolute localhost/private/metadata destinations are rejected through
  DNS and redirect revalidation
- add owner CLI controls, bounded rollout health/latency signals, a rollout and
  rollback runbook, live spec/plan updates, and red-green regression coverage

## Stack

This PR is intentionally stacked as:

1. #4288 (`claude/sched-v2-caughtup`)
2. #4679 (`codex/persistent-scheduler-state-default-on`)
3. this PR (`codex/server-execution-w1-2-shared-pool`)

Its merge base is exactly the current #4679 head,
`545af423fc253df550deff7f48beb7e602505049`.

## Work-order acceptance mapping

| Work order | Named evidence |
| --- | --- |
| W0.1 context-qualified durable scheduler state | `v2-scheduler-context-test.ts`, `v2-scheduler-state-test.ts` |
| W0.2 one direct root output binding | runner `runner.test.ts`; transformer `module.test.ts` |
| W0.3 indexed authoritative writer lookup | `scheduler-writer-lookup.test.ts`, `executor-writer-discovery.test.ts` |
| W0.4 exact basis and execution provenance | `v2-execution-control-test.ts`, `v2-execution-lease-engine-test.ts`, `executor-provider-parity.test.ts` |
| W0.5 authenticated in-process provider | `executor-provider-parity.test.ts`, lease/firewall tests in memory v2 |
| W0.6 handshake, demand, claims, settlements, reconnect | `v2-execution-control-test.ts`, `v2-execution-feed-reconnect-test.ts`, `v2-execution-client-lifecycle-test.ts` |
| W1.1 user-sponsored leases and fencing | `v2-execution-lease-test.ts`, `v2-execution-lease-engine-test.ts` |
| W1.2 one shared Worker and background exclusion | `executor-pool.test.ts`, `executor-legacy-background-transition.test.ts` |
| W1.3 whole-action servability/firewall/readiness | `executor-candidate-claim.test.ts`, `executor-shadow-storage.test.ts`, `executor-shadow-sink.test.ts`, `executor-claim-e2e.test.ts` |
| W1.4 brokered server builtins and egress | `server-builtin-channel.test.ts`, `server-builtin-transport.test.ts`, `server-builtin-egress.test.ts`, `pinned-http-fetch.test.ts` |
| W1.5 indexed wake, drain, and hibernation | `v2-stale-readers-test.ts`, `executor-drain-barrier.test.ts`, `executor-pool.test.ts` |
| W2.1 exact claim-aware routing and overlays | `client-action-router.test.ts`, `client-execution-overlay.test.ts`, `executor-shadow-sink.test.ts` |
| W2.2 settlement-driven reconciliation | `client-execution-overlay.test.ts`, `v2-execution-feed-reconnect-test.ts`, chained/divergence cases in `executor-claim-e2e.test.ts` |
| W2.3 client builtin passivity | claimed fetch/generate and pending-to-result cases in `executor-claim-e2e.test.ts` and server-builtin suites |
| W2.4 local rollout gates | `server-execution-rollout-products.test.ts`, `server-execution-product-fixtures.test.ts`, `executor-drain-barrier.test.ts`, `server-primary-rollout-profile.test.ts` |

The automated review's initial 14 findings, 8 first follow-up findings, and 3
second follow-up findings are addressed on the current head with targeted
regressions. Independent final audits also drove settlement-barrier and
post-rebase telemetry hardening. The fixes include strict policy parsing,
bounded interim HTTP responses, mixed-space writer discovery, bounded reconnect
feeds, serialized restartable background lifecycle, server-relative lease
deadlines, broker cancellation and redirect cleanup, reconnect-safe capability
gates, fail-closed pool authority loss, mapped-host builtin preservation,
permanent runtime rejection handling, sponsor routing, Fabric-safe diagnostics,
settlement-frontier preservation, native-error brand hardening, separately
ordered canonical-settlement transaction traces, an isolate-local native OTel
bridge whose user/space DIDs remain span-only rather than metric labels, and
post-#4679 replay-retirement hardening that preserves scope-qualified mirror
events while clearing unreachable accepted payloads from retired replay rows.

## Validation

With `EXPERIMENTAL_SERVER_PRIMARY_EXECUTION=true` and
`EXPERIMENTAL_PERSISTENT_SCHEDULER_STATE=true`:

- `TEST_CONCURRENCY=1 deno task test` — all workspace packages green; runner
  package wall time 308.4s; memory package wall time 53.2s; root wall time 593.9s
- `HEADLESS=1 deno task integration` — 7/7 suites green (`runner`,
  `runtime-client`, `shell`, `patterns`, `cli`, `generated-patterns`, and
  `pattern-tests`); all 91 authored pattern tests passed
- `deno task check` — 297 paths
- `deno task check-docs` — all 513 checked code blocks
- `deno fmt --check` — 3,604 files
- `deno lint` — 3,837 files

The authoritative 500-event counterbalanced renderer-process CPU gate also
passes: ABBA enabled/disabled `1.0347`, BAAB `1.0387`, combined `1.0366`
against the `1.10` ceiling; maximum same-policy spread `0.0586` against the
`0.15` ceiling. Every enabled phase suppressed all 500 derived client wire
commits and completed exact routing and settlement barriers. Evidence is in
`docs/history/development/performance/server-primary-rollout-2026-07-13.md`.

## Rollout boundary

Phases 0–2 are locally complete behind the default-off flag. The sole unchecked
W2.4 acceptance item is the real deployed-staging enable/disable drill; its
deterministic no-migration analogue is green, but the checkbox stays open until
an authorized staging space is exercised and the result recorded.

Phases 3–5 remain explicitly deferred by the plan. Terminal crash quarantine,
manual un-quarantine, hard per-Worker memory limits, and bounded-LRU pool
admission are separately tracked as later operational hardening (G18), not as
Phase 0–2 acceptance criteria.


Hixie:
Review feedback from my Claude (edited to remove irrelevant comments):

## Timeout-induced test flakiness

The repo's own guidance treats timeouts, retry loops, and sleeps as defects because a timeout puts an upper bound on success. This PR's new tests lean on them heavily.

### 1. `executor-drain-barrier.test.ts` flakes and fails a large fraction of runs

The test *"real executor crash rejects stale work and converges once through client fallback and replacement"* ([executor-drain-barrier.test.ts:991](packages/runner/test/executor-drain-barrier.test.ts:991)) waits for the client fail-open output using a hard 10-second timeout helper `within()` ([line 200](packages/runner/test/executor-drain-barrier.test.ts:200)).

I reproduced this two ways on a clean checkout of the PR head, and a verification agent reproduced it a third way in a fresh worktree:
- Run in isolation with `--filter`, it fails at exactly 10 seconds ("client fail-open output timed out") on roughly five of every seven runs.
- Run as the whole file, one of two of my runs failed on this same test; the agent also saw a whole-file run fail it.

The mechanism matters more than the flake. After the executor worker crashes, the server revokes the claim, but the **client** only learns about the revoke through a replica execution-feed sync ([storage/v2.ts](packages/runner/src/storage/v2.ts) `applyExecutionControlEvent`). Neither `clientRuntime.settled()` nor the crash path forces that sync. The test itself concedes this for a *later* step — there is a comment at [line 1341](packages/runner/test/executor-drain-barrier.test.ts:1341) admitting "runtime.settled() alone drains local work, not unsolicited control-feed catch-up" and it explicitly forces a sync there — but the fail-open assertion at [line 1275](packages/runner/test/executor-drain-barrier.test.ts:1275) has no equivalent nudge, so it relies on incidental event-loop activity that is not reliably present.

This raises a question the author should answer before dismissing it as a test bug: **do production clients proactively drain the execution control feed after an executor crashes?** If they do not, a real client's fail-open could be stranded until its next unrelated sync — the same latency the test is accidentally papering over. If production is genuinely event-driven here, then the fix is just to drive the test's fail-open assertion off that event instead of the 10-second timer.

Reproduce:
```
cd packages/runner && deno test -A --filter "real executor crash rejects stale work and converges once through client fallback and replacement" test/executor-drain-barrier.test.ts
```

### 2. A broader family of fixed-deadline `Promise.race` waits

The drain-barrier test is not isolated. Across the new suites, tests wait for an in-process event by racing it against a `setTimeout` that rejects or resolves `false`, then assert the good outcome. The wide-margin ones (10 seconds) are effectively fail-safes and rarely bite (though of course they should all be fixed regardless), but the **small-margin** ones gate correctness on scheduling latency and will flake under CI's loaded parallel runner:
- 100-millisecond promptness assertions on dispose and a stranded read in [executor-provider-parity.test.ts:2262](packages/runner/test/executor-provider-parity.test.ts:2262) (`assertEquals(disposedPromptly, true)` and `assert(pendingResult.settled)`).
- A 500-millisecond `stopStarted` race in [executor-pool.test.ts:1190](packages/runner/test/executor-pool.test.ts:1190).
- 1-second settlement races in [v2-execution-control-test.ts](packages/memory/test/v2-execution-control-test.ts), and notably a **real** `setTimeout(1000)` deadline running alongside an installed `FakeTime()` fake clock in [v2-execution-client-lifecycle-test.ts:405](packages/memory/test/v2-execution-client-lifecycle-test.ts:405) — mixing a real wall-clock deadline with a faked clock is a latent trap.

The fix in every case is to await the completion event the code already exposes (these files elsewhere use `Promise.withResolvers` and subscriptions correctly) rather than racing a timer.

### 3. A bounded polling loop that the CI guard does not catch

[client-execution-overlay.test.ts:276](packages/runner/test/client-execution-overlay.test.ts:276) defines a `waitFor()` that loops up to 100 times doing `await new Promise(r => setTimeout(r, 0))` and then throws. This is the polling anti-pattern the repo forbids, used at roughly 55 call sites in that one new file. Importantly, the CI `check-no-waitfor` guard does **not** flag it: that guard only scopes to files under an `integration/` directory importing `waitFor` from `@commonfabric/integration`, and this is a unit test with a locally-defined helper. The 100-iteration cap is generous enough that a false failure is unlikely, but it should be an event-driven wait like its siblings.

---

## Product-code timers

### 4. New 50-millisecond speculation grace holds `idle()`

`CLAIMED_REMOTE_SPECULATION_GRACE_MS = 50` is new at [constants.ts:44](packages/runner/src/scheduler/constants.ts:44). When a claimed computation is invalidated by a remote write, the scheduler defers the local speculative run and blocks `idle()` for up to 50 milliseconds ([facade.ts:2596](packages/runner/src/scheduler/facade.ts:2596), released via `scheduleWake`). It is correctly bounded, non-sliding, fails open at the deadline, and is gated behind the default-off flag plus a live claim, so it never fires in the default configuration. It is flagged only because it couples `idle()` resolution to a wall clock; the cleaner seam would be to drive the fall-open off the server-adoption event that already calls `releaseClaimedRemoteSpeculation`.

### 5. Executor safety bounds

The executor introduces a settle timeout (60 seconds), startup and claim-activation timeouts (30 seconds), egress timeout (30 seconds), and exponential crash backoff ([shared-execution-pool.ts](packages/runner/src/executor/shared-execution-pool.ts), [deno-space-executor.ts](packages/runner/src/executor/deno-space-executor.ts)). I checked the timer lifecycles: every one is cleared before reschedule and on shutdown/close, so there are no dangling-timer leaks. These are largely justified distributed-systems bounds — a hung worker or a slow external fetch must be bounded, and the backoff prevents hot-spinning. The one with a real behavioral consequence is the 60-second settle timeout: a graceful drain still making progress past 60 seconds is killed and its `lastSettledSeq` is not advanced, which triggers a redundant re-run of in-progress work on the next wake. This is **wasted recomputation, not wrong state** (the same safe recompute-from-persisted-state path used on crash-retry), so it is a deliberate-tradeoff note rather than a bug. Worth a one-line rationale comment where each bound is introduced.

---

## Correctness

### 6. `#executionClaimGeneration` grows without bound

In [memory/v2/server.ts:3441](packages/memory/v2/server.ts:3441), the per-action claim-generation counter map gains one permanent entry per distinct action-claim key and is never pruned per-key — the live-claim map `#executionClaims` is deleted on expiry, revoke, and space-disable, but the generation map is only ever cleared wholesale in `stop()`. The key includes `implementationFingerprint`, `runtimeFingerprint`, `pieceId`, and `actionId`, all of which churn across deploys and per pattern instance, so a long-lived server's memory grows monotonically with the cumulative count of every action ever claimed. The retention is **load-bearing** — the counter must stay monotonic across a delete-then-reclaim cycle to keep the fencing generation strictly increasing — so the fix is not naive deletion but scoping the counter to the durable `leaseGeneration` (which already advances per acquire) or persisting the last generation into the durable lease row. Only matters for very long-lived processes, hence low.

---

## Security

### 7. Cross-origin redirects forward non-`authorization` secret headers

The one real security note: on a cross-origin redirect the broker strips only the `authorization` header ([server-builtin-egress.ts:798](packages/runner/src/executor/server-builtin-egress.ts:798)). Any other caller-supplied secret header — `X-Api-Key`, a custom bearer header — is forwarded to the redirect target, and for 307/308 the request body is replayed too. An agent confirmed this empirically with a probe: a second hop to `https://attacker.example` still received `x-api-key` and a custom auth header while `authorization` was correctly `null`. This mirrors WHATWG `fetch` semantics (browsers also strip only a fixed set and forward author-set custom headers), so it does not strictly widen network reach beyond browser execution — which is the spec's stated bar. But for a **server-side** broker where the redirect `Location` can be attacker-influenced (for instance via an open redirect on an otherwise-legitimate API), forwarding arbitrary secret headers cross-origin deserves an explicit decision rather than inheriting browser defaults by omission. Consider stripping caller-supplied sensitive headers (or reducing to an allowlist) on any cross-origin hop.

---

## Performance

### 8. `writeAll` copies the tail on every partial write

[pinned-http-fetch.ts:663](packages/utils/src/pinned-http-fetch.ts:663) does `conn.write(bytes.slice(offset))` in its partial-write loop. `slice` allocates and copies the whole remaining body each iteration, so a socket applying backpressure on a near-limit (1 MiB) body turns into O(n²) copying. `conn.write` only reads its argument, so the fix is a one-word change to `bytes.subarray(offset)` (a zero-copy view). Low impact because bodies are bounded and usually drain in one write, but the fix is free.

### 9. Writer discovery rescans all effects per target

In [writer-lookup.ts:172](packages/runner/src/scheduler/writer-lookup.ts:172), the effects branch of `collectLiveSchedulerWriters` iterates the entire active-effects set for every target and re-surfaces each effect's write list without memoization, while the neighboring computation and materializer branches use entity-keyed indexes. On a single-space server executor the identity guard never short-circuits, and this runs twice per demanded piece across a redirect chain of up to 8 targets, so cold-start writer discovery is roughly O(pieces × effects) on effect-heavy spaces. Output is correct — this is a cold-start latency floor, not a correctness issue. The fix is to index effect write surfaces by entity like `writersByEntity`, or memoize `resolveRegistrationSurface` per action generation.

---

## Docs

- **Docs / flags:** One trivial nit: the document-level "Last reviewed: 2026-07-12" header at [EXPERIMENTAL_OPTIONS.md:23](docs/development/EXPERIMENTAL_OPTIONS.md:23) is a day older than the `serverPrimaryExecution` section's own "Status on 2026-07-13" that this PR adds; bump the header to match.

seefeldb:
Processed review summary for current head `dec968074`:

1. Crash fail-open now follows the execution-control event path; the regression no longer depends on incidental sync or a 10s deadline.
2. The identified fixed-deadline test races were replaced with direct lifecycle/settlement events.
3. The local overlay polling helper was replaced with test-owned sync-delivery, transaction-start, storage-notification, and processing acknowledgements.
4. The 50ms claimed-remote speculation grace remains intentionally bounded, non-sliding, and fail-open; that tradeoff is now documented.
5. Executor startup/activation/settle/egress bounds now carry their safety rationale.
6. Claim-generation state is scoped to an owned lease: monotonic within a lease, reset/fenced by a new `leaseGeneration`, and covered end-to-end.
7. For this trusted-client iteration, redirect handling deliberately preserves browser Fetch parity: `Authorization` is stripped cross-origin while caller custom headers remain. This is documented and tested; stricter untrusted-client policy remains a later hardening step.
8. Partial socket writes now use zero-copy `subarray`, with a buffer-reuse regression.
9. Live effect writers are indexed by target entity and invalidated by the node-registry generation.
10. The experimental-options review date was corrected.

Also covered the pattern-update case raised after the review: demand is anchored by the stable piece-root entity, startup reads that root's current `patternIdentity`, and the current live action supersedes a durable stale same-piece action row. The compiled regression starts the old pattern, updates the same piece to a new pattern/action, injects the old durable writer row, restarts from the stable root, and asserts that the new action/result wins.

All 24 inline review threads are resolved. Local validation on the exact current #4679 base: `deno task check`, full flagged `deno task test`, and full flagged `deno task integration` (7/7 groups) are green.

PR #4696, opened 2026-07-13
fix(runner): pull same-space link targets a fresh replica never synced (fresh-replica read asymmetry)

mathpirate:
## The bug

The deepest open platform bug in the Topics work-line: from a fresh CLI replica, **writes commit reliably but reads of cross-replica-written data come back masked** — `undefined`, schema defaults, or `property not found` — indistinguishable from absence. Twice it manufactured data-loss illusions that invited destructive "repair" writes (see the "Fresh-replica read asymmetry" and "Board outage 2026-07-10" topics on `topics-dev-476ea34f`), and it corrupted writes-that-read (topics filed with `createdByName: "someone"` because `myName.get()` read masked).

## Root cause (three legs)

1. **Schema-less syncs normalize to the rejecting selector** ([v2-watch.ts `normalizeSyncSelector`](../blob/main/packages/runner/src/storage/v2-watch.ts)) → the server delivers only the root doc. CLI read paths (argument cells via meta link, path reads) are schema-less, so linked sub-docs (array entries, argument members) never reach the replica.
2. **Reads of never-pulled docs fall back to `unclaimed`** ([v2-transaction.ts `loadRoot`](../blob/main/packages/runner/src/storage/v2-transaction.ts)) — silently `undefined`. There is no "not synced here" state at the read layer.
3. **Only cross-space absent link targets got background loads kicked** (link-resolution hop loop, traverse `followPointer`). Same-space targets were assumed "covered by the originating query or genuinely absent" — false after a rejecting-selector sync. Own writes are locally present, hence the write/read asymmetry; renderers carry pattern schemas and never notice.

## The fix

Extend both kick sites to same-space targets **the replica has never seen**, gated by a new optional `IStorageManager.shouldPullDoc` — true at most once per (space, scope, id) per manager lifetime, and only when the replica holds no state for the doc (dangling links / deleted docs cannot churn the loop). `Cell.pull()`'s existing convergence loop awaits the kicked syncs and re-reads; schema-less pulls now converge to the true value. **Client-side only — no server change required** (a redeployed server additionally regains one-shot delivery via the traversal fix below).

Two convergence enablers:
- `traverseArrayWithSchema` evaluates **every** element instead of short-circuiting at the first failure. Semantics unchanged (a failing element still voids the array for that round), but all absent element docs kick in the same round → convergence proportional to link **depth**, not element-count × depth. The 19-entry production board exhausted the old budget serially.
- `pull()`'s round cap becomes a generous backstop (10 → 100); kicks are once-per-doc so rounds are naturally bounded by reachable-doc depth.

## Verification

- New regression suite `packages/runner/test/fresh-replica-read-asymmetry.test.ts` (two `StorageManager`s loopback-connected to one in-process memory server): schema-less pull, CLI-shaped per-index path reads, two-hop link chains, dangling links, strict-`required` array items schemas, plus a characterization of the remaining one-shot `get()` semantic.
- **End-to-end against the production board** (`topics-dev-476ea34f`, fresh CLI replica per invocation): `--input topics` returns the full 19-topic array (was `undefined`); `topics/N/title` reads all indices (was property-not-found for N≥1); result-cell `topics` reads (was property-not-found); `myName` reads `"Fable"` (was masked → the "someone" corruption).
- Gates: fmt + lint + typecheck clean; full suites green — runner 876/876, memory 354/354, piece 15/15, cli 20/20.

## Remaining sharp edges (follow-ups, tracked as topics on the board)

- One-shot `get()` after a bare schema-less `sync()` still returns the masked view (kick is async; `pull()` is the converging read — all CLI surfaces use it). Loud sync-state in reads is the deeper reform.
- Handler-execution reads (`myName.get()` inside `addTopic`) run mid-transaction and cannot await kicks — presync coverage for user-scoped docs is a separate leg.
- Why the deployed server's schema'd walk under-delivers the argument graph (client kicks compensate; after redeploy the array-walk fix restores one-shot coverage) deserves its own look.

---
## Summary by cubic
Fixes a bug where fresh replicas read linked docs as missing after schema-less syncs. `pull()` now fetches same-space link targets a replica has never seen (scoped, deduped, and retriable), and only retracts reservations when the current kick took them.

- **Bug Fixes**
  - Extend link resolution to kick background pulls for same-space targets, gated by `IStorageManager.shouldPullDoc` (once per space/scope/id and only when no local state); on failure, `retractDocPullKick` is called only if this kick reserved it, so a failed cross-space kick cannot clear another read’s reservation.
  - Scope-key the missing-target dedup in `Runtime.ensureLinkedDocLoaded`; early-return when the doc is already locally known; clear the runtime dedup on failure and retract the storage reservation only if this invocation reserved it.
  - Update traversal/`Runtime.ensureLinkedDocLoaded` to pass `sourceSpace` and let the runtime decide whether to fetch; treat missing same-space targets like cross-space when never seen; protect reservations across concurrent reads.
  - Increase `Cell.pull()` convergence cap to 100; traverse arrays by evaluating every element to kick all needed loads in the same round (semantics unchanged).
  - Expand tests in `fresh-replica-read-asymmetry.test.ts` to cover reservation ownership, cross-space failure isolation, transient failure retry, scope isolation, and traverse-side failure handling.

<sup>Written for commit aa7208352d3469c243b8863cde198069a8f8fa0a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4696?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/cell.ts
+++ b/packages/runner/src/cell.ts
@@ -1067,25 +1067,28 @@ export class CellImpl<T extends FabricValue>
       });
 
       // Wait for the scheduler to process all pending work, then resolve.
-      // If the read kicked async loads of cross-space link targets, await
-      // them and re-idle — each arrival re-runs the read and can reveal the
-      // next hop — bounded. Pulls that kicked nothing take the
-      // zero-iteration path and keep their previous timing.
+      // If the read kicked async loads of absent link targets (cross-space,
+      // or same-space docs a fresh replica never pulled), await them and
+      // re-idle — each arrival re-runs the read and can reveal the next hop.
+      // Every kicked doc is deduped once-per-session, so the total number of
+      // rounds is bounded by the reachable-doc depth; the fixed cap is only
+      // a backstop against a pathological graph. Pulls that kicked nothing
+      // take the zero-iteration path and keep their previous timing.
       this.runtime.scheduler.idle().then(async () => {
         const storage = this.runtime.storageManager;
         // The pending pool is manager-global (same semantics as `synced()`):
         // this pull may also wait on loads kicked by concurrent readers.
         let round = 0;
-        for (; round < 10; round++) {
+        for (; round < 100; round++) {
           if ((storage.pendingCrossSpacePromiseCount?.() ?? 0) === 0) break;
           await (storage.crossSpaceSettled?.() ?? Promise.resolve());
           await this.runtime.scheduler.idle();
         }
         if (
-          round === 10 && (storage.pendingCrossSpacePromiseCount?.() ?? 0) > 0
+          round === 100 && (storage.pendingCrossSpacePromiseCount?.() ?? 0) > 0
         ) {
           logger.warn("pull", () => [
-            "pull() convergence bound exhausted with cross-space loads still",
+            "pull() convergence bound exhausted with link-target loads still",
             `pending: ${this.sourceURI}`,
           ]);
         }
--- a/packages/runner/src/link-resolution.ts
+++ b/packages/runner/src/link-resolution.ts
@@ -317,9 +317,23 @@ export function resolveLink(
       } else {
         link = nextLink;
       }
-      // If we're crossing spaces, force fetching data from server, as the
-      // original server will not have pushed the data to the client yet.
-      if (crossSpace) {
+      // Force fetching data from the server when the local replica cannot
+      // serve the hop target: crossing spaces (the origin server never pushes
+      // other-space docs), or a same-space doc this replica has never pulled.
+      // The second arm is the fresh-replica read-asymmetry fix: selector
+      // driven syncs only deliver what a schema covered, so a link can point
+      // at a same-space doc no selector ever walked — without this kick such
+      // reads mask as `undefined`, indistinguishable from absence. The kick
+      // is async; one-shot reads still return the masked value, but
+      // `Cell.pull()`'s convergence loop awaits the tracked sync and re-reads.
+      if (
+        crossSpace ||
+        runtime.storageManager.shouldPullDoc?.(
+            link.space,
+            link.id,
+            link.scope,
+          ) === true
+      ) {
         // Swallow sync failures: this kick is best-effort (the read still
         // resolves from the local replica) and an unhandled rejection here
         // would otherwise escape the resolution path.
--- a/packages/runner/src/link-resolution.ts
+++ b/packages/runner/src/link-resolution.ts
@@ -328,15 +328,19 @@ export function resolveLink(
       // `Cell.pull()`'s convergence loop awaits the tracked sync and re-reads.
       const mgr = runtime.storageManager;
       const { space, id, scope } = link;
-      if (crossSpace || mgr.shouldPullDoc?.(space, id, scope) === true) {
+      const reserved = !crossSpace &&
+        mgr.shouldPullDoc?.(space, id, scope) === true;
+      if (crossSpace || reserved) {
         // Swallow sync failures: this kick is best-effort (the read still
         // resolves from the local replica) and an unhandled rejection here
-        // would otherwise escape the resolution path. Retract the
-        // shouldPullDoc reservation on failure so a later read may retry
-        // (no-op for the cross-space arm, which never reserved).
+        // would otherwise escape the resolution path. On failure, retract
+        // the shouldPullDoc reservation so a later read may retry — but only
+        // when THIS kick took it: a failed cross-space kick never reserved,
+        // and must not clear a reservation a concurrent same-space read
+        // holds for the same target (that would permit duplicate syncs).
         mgr.trackUntilSettled(
           runtime.getCellFromLink(link).sync().catch(() => {
-            mgr.retractDocPullKick?.(space, id, scope);
+            if (reserved) mgr.retractDocPullKick?.(space, id, scope);
           }),
         );
       }
--- a/packages/runner/src/link-resolution.ts
+++ b/packages/runner/src/link-resolution.ts
@@ -326,19 +326,18 @@ export function resolveLink(
       // reads mask as `undefined`, indistinguishable from absence. The kick
       // is async; one-shot reads still return the masked value, but
       // `Cell.pull()`'s convergence loop awaits the tracked sync and re-reads.
-      if (
-        crossSpace ||
-        runtime.storageManager.shouldPullDoc?.(
-            link.space,
-            link.id,
-            link.scope,
-          ) === true
-      ) {
+      const mgr = runtime.storageManager;
+      const { space, id, scope } = link;
+      if (crossSpace || mgr.shouldPullDoc?.(space, id, scope) === true) {
         // Swallow sync failures: this kick is best-effort (the read still
         // resolves from the local replica) and an unhandled rejection here
-        // would otherwise escape the resolution path.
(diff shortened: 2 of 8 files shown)


mathpirate:
Both cubic findings judged **valid** and fixed in 4d53a0e83, plus the coverage-debt OVER (+1) addressed:

**P1 (scope-blind kick dedup)** — `Runtime.missingDocLoadKicks` now keys on `(space, scope, id)` via `normalizeCellScope`, matching `shouldPullDoc`'s own key. Scoped instances (user/session) are distinct docs; a kick for one scope no longer suppresses another's. Pinned by the "missing-target kicks are scope-keyed" test (pre-fix: the second scope's sync is swallowed).

**P2 (permanent mask after transient kick failure)** — new optional `IStorageManager.retractDocPullKick(space, id, scope)` hands the `shouldPullDoc` reservation back when a kicked sync rejects; both kick sites (link-resolution hop loop, `ensureLinkedDocLoaded`) call it from their failure paths. The replica layer already allowed retry (a failed pull clears the in-flight task and the selector tracker), so the reservation was the only stuck bit. Pinned by three tests: reservation lifecycle (reserve → hold → retract → re-reserve; scope-independent; `data:` never pulled), an injected transient failure healing by the next pull at the latest — in practice within the *same* pull, since sibling arrivals re-run the read effect and the retried kick lands — and the traverse-side failure path handing back both reservations (plus the present-doc early return: never re-kicked).

**Coverage debt** — verified with local LCOV over the full runner suite: every added source line is now exercised except the round-cap warn lambda in `cell.ts` (whose removed round-10 counterpart was equally uncovered on main — net 0). Two of the "uncovered" regions turned out to be a deno-coverage attribution artifact (multi-line call arguments are attributed to no line even when the call executes — observable as an executed `add()` directly beneath "unexecuted" condition lines); the kick sites are restructured onto fewer expression lines so the debt gate counts what actually runs.

**Relationship to #4668** — no textual overlap (`git merge-tree` clean: #4668's traverse hunks are the object-grace internals, this PR's are the array walk) and semantically complementary: #4668's grace makes generation-skewed elements *survive* validation; this PR makes never-pulled elements *converge* and stops one failing element from serializing the kicks. Suggested order: #4668 lands first (deploy-urgent — it closes the redeploy brick-trigger), this rebases as a no-op.

mathpirate:
Cubic's round-2 finding also judged **valid** and fixed in aa7208352: the failure catch retracted the `shouldPullDoc` reservation unconditionally, so a failed cross-space kick (which never reserves — the `||` short-circuits) could clear a reservation a concurrent same-space read held for the same target, permitting duplicate syncs. Both kick sites now capture whether *this* invocation took the reservation and retract only if owned. Pinned by a new test: a held same-space reservation survives a failed cross-space-sourced kick for the same target (13-step suite green; runner 876/876).

Note on the previous run's Pattern Integration (2/4) failure: harness-level (`Child process has already terminated` in the two-browser lunch-poll test, not an assertion), and the job family flakes on main too (shard 4/4 failed twice there this week with a different harness flavor). This push re-runs everything fresh.

ubik2:
I think I'm missing something here.
If your acceptance criteria for a field allows undefined, then does it matter if the field is undefined because it's been set to that or because the record doesn't exist?
We can differentiate between these in the server, but I'm just not sure how that would be meaningful.
Also, you should be able to have a subscription to a doc that doesn't exist, and as soon as it is created, you will be notified.

I think part of what this is talking about is if I have doc A, and no doc B, and doc A's content looks like:
```javascript
{
 linked: { /* link to doc B */ }
}
```
then when I load doc A, I have `undefined` for linked, and a subsequent write of doc B doesn't change that.
Maybe the right answer here is to actually have a schema for the CLI. The default schema says to include this one doc, but not any linked docs, so I really shouldn't expect to be able to resolve `linked`. I might be able to, because my client happens to have fetched it successfully, but that's not reliable.

Edit: and if your goal is to retrieve the doc and all of its linked docs, you can query with schema true. That just means you often get a lot more data than you want.

mathpirate:
> I think I'm missing something here. If your acceptance criteria for a field allows undefined, then does it matter if the field is undefined because it's been set to that or because the record doesn't exist? We can differentiate between these in the server, but I'm just not sure how that would be meaningful. Also, you should be able to have a subscription to a doc that doesn't exist, and as soon as it is created, you will be notified.
> 
> I think part of what this is talking about is if I have doc A, and no doc B, and doc A's content looks like:
> 
> ```js
> {
>  linked: { /* link to doc B */ }
> }
> ```
> 
> then when I load doc A, I have `undefined` for linked, and a subsequent write of doc B doesn't change that. Maybe the right answer here is to actually have a schema for the CLI. The default schema says to include this one doc, but not any linked docs, so I really shouldn't expect to be able to resolve `linked`. I might be able to, because my client happens to have fetched it successfully, but that's not reliable.
> 
> Edit: and if your goal is to retrieve the doc and all of its linked docs, you can query with schema true. That just means you often get a lot more data than you want.

sorry for the coordination failure of merging this before you had a chance to look! here's fable's response to your input. let me know whether this makes sense to you or not. i think it makes sense conceptually to me:

Set-undefined vs doesn't-exist: agreed — that distinction rarely matters, and it isn't the one the PR is after. The one that bit us: a value exists on the server, but this replica never fetched it. Schemas that tolerate undefined are fine either way — but we rewrote a healthy board because empty-looking reads were indistinguishable from data loss, and handlers committed masked reads as real values (createdByName: "someone"). The distinction earns its keep operationally even where types don't care.

Your doc A/B example is exactly the bug. And the reason a later write of B didn't help: nothing ever subscribed to B — the original sync was never asked to include it, so the creation notification had nowhere to land. That's what this PR changes: resolving the link kicks a one-time background fetch of B and registers that subscription, so B's later creation now does arrive. pull() waits for those fetches, so it converges deterministically — no longer "happens to have it locally." One-shot get() remains best-effort; more below.

A schema for the CLI: yes, and. Declared shapes are the efficient path — it's why the browser never saw this bug. But two limits: the board read that started all this was schema'd (the argument link carries the full schema) and the server still didn't deliver the linked docs in one shot — one cause fixed in this PR, pending deploy; the remainder is tracked in the [server-walk topic](https://estuary.saga-castor.ts.net/topics-dev-476ea34f/fid1:aWeez7qX6Cbh-2cadI6s8XYJDhaEZ0PqGY1vTwUIePc). And arbitrary piece get <path> reads have no schema to declare, while a permissive one would pull half the space — lazy-by-default is deliberate. So: schemas make declared shapes fast, the kicks make undeclared reads correct, and the remaining one-shot-get() gap wants #4677's syncing marker ("not synced here" as a real state) — tracked in the [loud-sync-state topic](https://estuary.saga-castor.ts.net/topics-dev-476ea34f/fid1:WrkFIfNd1xiu40IdrwycDfC55w9bcUAaHgAFgeImLI8). Defaulting the CLI to its recovered schemas more aggressively (plus an explicit --schema flag) is a good concrete follow-up — folding it in there.

PR #4697, opened 2026-07-13
test: cover the intermittently-covered lines that flap the coverage-debt gate

Hixie:
## Why

The coverage-debt gate counts how many source lines are covered and fails when
that number regresses against a recent sample from main. A group of lines were
covered only as a side effect of patterns compiling through the transformer
during the pattern-integration jobs. Those jobs reuse a compile cache, and on a
cache hit the transformer and schema-generator never run, so their lines are not
recorded as covered for that run. The same unchanged code therefore alternates
between covered and uncovered from one run to the next. That flapping shows up in
the gate as a spurious coverage-debt increase, unrelated to the change under
test — for example, a docs-only PR appeared to raise `packages/runner` debt purely
because of which sample it landed against.

## What this does

Add focused unit tests that exercise each of those lines directly, so they are
covered on every run regardless of the compile cache. The tests drive the code
through the real transformer pipeline (via `transformSource`) or call the
emitter and helper functions directly, and they assert on real behavior — parsed
syntax-tree nodes and returned values — rather than just executing the line. They
are quality tests that pin down actual behavior, not coverage stunts.

Areas covered:

- **ts-transformers** — the expression-rewrite emitters (including the zero-arg
  IIFE branch and its `createLiftAppliedCall` fallback, the no-dataflow guard,
  and the receiver scan), expression-site lowering, schema injection, type
  shrinking, capability analysis, the synthetic-helper-wrapper fallback, and
  assorted small emitter and policy branches.
- **schema-generator** — formatter and generator branches that previously ran
  only during cold pattern compiles.
- **runner** — the SQLite query write-back and stale-supersede guards, the
  compiled-JS scanner's bitwise-operator branch, and the memory WebSocket
  transport's close and error paths.
- **memory** — the acknowledgement reschedule after a failed send while
  connected.

## Source changes

Two symbols are exported so the tests can import and drive them directly, with no
behavior change: `WebSocketTransport` in the runner's remote-session storage, and
`isSyntheticHelperWrapperInArrayMethodCallback` in the transformer's
expression-site lowering.

## Scope

This makes each flapping line deterministically covered, which stops it from
destabilizing the gate. It does not change the underlying mechanism — the compile
cache in the pattern jobs combined with a ratchet that compares against a single
recent sample. Reworking that ratchet (for example, taking the minimum over
several recent baselines) is a separate change and is not included here.

## Test plan

- `deno task test` in `packages/ts-transformers`, `packages/schema-generator`,
  `packages/runner`, and `packages/memory` — all green.
- The new ts-transformers tests were verified to cover their target lines under
  V8 coverage on the unmodified source.

---
## Summary by cubic
Stabilizes the coverage-debt gate by adding deterministic unit tests that cover lines previously executed only during cold pattern compiles or timing-dependent paths. No runtime behavior changes; two symbols are exported to enable testing, and the new flap-coverage tests are formatted with `deno fmt`.

- **Bug Fixes**
  - Added deterministic tests across `packages/ts-transformers`, `packages/schema-generator`, `packages/runner`, and `packages/memory` to cover cold-compile and timing-only branches (e.g., call-expression/IIFE paths including the lift-applied fallback, expression-site nested/provenance cases, schema-generator defaults and node-driven overrides, sqlite result write-back and stale-supersede guards, WebSocket close/error handling, compiled-js single `&`/`|`, and ack retry after a failed `session.ack`).
  - Exported `WebSocketTransport` (runner) and `isSyntheticHelperWrapperInArrayMethodCallback` (transformers) for test access only; no behavior changes.
  - Tests are event-driven (no sleeps) to keep coverage stable across runs.

- **Refactors**
  - Applied `deno fmt` to the new flap-coverage tests; no behavior changes.

<sup>Written for commit 6cd92f89842116da456e7e2d2837478afa1c644c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4697?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/memory/test/v2-client-test.ts
+++ b/packages/memory/test/v2-client-test.ts
@@ -907,6 +907,40 @@ Deno.test("memory v2 client coalesces watch ack bursts", async () => {
   }
 });
 
+Deno.test("memory v2 client reschedules an ack after a failed send while still connected", async () => {
+  const transport = new FailFirstAckTransport();
+  const client = await connect({ transport });
+  const session = await client.mount("did:key:z6Mk-memory-v2-ack-retry");
+
+  try {
+    // A watch result acks the observed server seq. The first ack send fails
+    // with an error response (the connection stays up), so the ack flush's
+    // finally reschedules a retry.
+    await session.watchAdd([{
+      id: "root:1",
+      kind: "graph",
+      query: {
+        roots: [{
+          id: "of:doc:1",
+          selector: {
+            path: [],
+            schema: false,
+          },
+        }],
+      },
+    }]);
+
+    // Wait for the rescheduled retry to be accepted — event-driven, no sleep.
+    await transport.ackSucceeded;
+    // One failed attempt, then the rescheduled attempt that succeeded.
+    assertEquals(transport.ackAttempts, 2);
+    // A failed ack while connected must not drop the connection.
+    assertEquals(client.isConnected(), true);
+  } finally {
+    await client.close();
+  }
+});
+
 Deno.test("memory v2 watch view keeps emitted snapshots incrementally ordered", async () => {
   const view = WatchView.fromSync({
     type: "sync",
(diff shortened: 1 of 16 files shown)


ubik2:
Nice! Was thinking to tackle this later today. Glad you beat me to it.

Hixie:
@ubik2 it's not the last time we'll have to do it, feel free to tackle it as well :-)

I basically told my agent to look at all the commits on main for the past few days, and to look for pairs of commits that were identical except for something that should definitely not change coverage, then to look for what lines differed between those two commits, and to write tests that definitely covered those lines.

PR #4698, opened 2026-07-13
feat(topics): derive the prose reference graph as navigable crossref chips

mathpirate:
> **Rev complete (2026-07-15)** — this PR now carries the full future-oriented scope agreed in review: four commits, each independently gated. ① v1 board chips + `crossrefs` output · ② shared `crossrefJoin`/`topicCorpus` helpers (tier 3 extractable, notes-ready) · ③ `mentionable` input + detail-page **Connections** card (tier 2 — the cells-durable/bundles-swappable model exercised in-PR; needs the merged #4714) · ④ cf-code-editor authoring with `@`-mentions bound to the session-local draft + markdown read view (tier 1). Suites 30/30 + 11/11; pattern coverage 196/196 + 511/511 (zero uncovered, matching baseline).

## What

The Topics board now derives the corpus's **prose reference graph** and renders it: any topic fid pasted in a body, comment, or link URL — bare `fid1:X`, `of:fid1:X`, page URL, or percent-encoded share link — becomes a navigable "references →" / "← referenced by" chip on the topic's card. The graph is also exported as the piece-valued `crossrefs` output for tests, embedders, and headless reads.

Design record: the ["Topic cross-references" topic](https://estuary.saga-castor.ts.net/topics-dev-476ea34f/fid1:lfKX1c9GdWRHtQxdEbPELXT-eMDCftlxtbv1XzXitvQ) on topics-dev (the board is its own design doc — this PR implements that body's v1 sketch).

## Design commitments

- **Derived at read time, never persisted.** The notes/mentions prior art (`backlinks-index.tsx`) wipes-and-rewrites `backlinks[]` into target pieces — a partial-view replica running that destroys real edges. Here the board rescans the whole corpus per render (`O(topics × text)`, trivial at board scale), the entire existing corpus lights up retroactively with zero authoring changes, and a retracted mention simply stops being an edge.
- **Identity = fid payload.** `topics.key(i).resolveAsCell().entityId` → `entityRefToString` (the notes-pattern idiom) gives each entry's result-doc fid; prose is scanned for payloads so URL-encoding and host variance don't matter. The fid1-tag assumption is pinned by a test that fails loudly if entity ids stop being fid1-tagged.
- **v1 scope: board cards only.** Detail-page chips would need either a topic-schema change (siblings input) or a board-hosted detail view — both violate the "no schema change, no migration" constraint that makes the corpus light up retroactively. Deferred; noted in the topic.

## Platform finding (cost most of the build time)

**A handler bound to a piece reference nested inside a derived wrapper object silently prevents the consuming UI computed from ever running.** First implementation derived `rows: {topic, refsOut: TopicPiece[], …}[]` and bound `openTopic({topic: row.topic})` — the card-list computed never executed once (no error anywhere; reads through the same wrappers work fine, only binds break). Minimal repro: derive any wrapper array carrying pieces, bind a handler to a nested piece inside the consuming computed's JSX. The shipped shape dissolves it: edges are index arrays over raw `topics` order, and every bind targets a direct `topics` element (`openTopic({topic: list[j]})`) — baseline provenance. **Root-caused and fixed in #4714 (merged)** for the identity-erasure mechanism — the detail-page Connections derive in this PR is its first consumer. The wrapper-bind shape remains a distinct open mechanism (same board topic), so the index-plumbed board core stays quarantined until it falls; the collapse to one piece-valued derive remains the tracked revisit.

Secondary observation — **resolved while this PR was open**: cf-test used to materialize `[UI]` only in narrow windows ([topic](https://estuary.saga-castor.ts.net/topics-dev-476ea34f/fid1:biRPdv6-A4QZiP7i_8W0LhfOe3VJzFYhKolzLd60yd0)); #4715 added explicit `{ render: … }` steps and continuous `$UI` demand, and this suite exports `[UI]` so the real in-card chip path now renders under test. The chip row stays a module-scope pure helper (`crossrefChipRow`) driven directly as well — it pins the no-edges → null branch deterministically. Pattern coverage: 0 uncovered lines for both files post-rebase (`CF_PATTERN_COVERAGE_DIR` DA-union, matching baseline's 0).

## Verification

- `topics.test.tsx` 30/30 (was 19) — fid form pinned, edges from all four pasted URL shapes, inversion, self/unknown-fid exclusion, edge removal on prose retraction, chip-row markup incl. binds and empty-collapse. `multi-user.test.tsx` 11/11. fmt/lint/cf-check + full workspace `deno task check` green.
- **Rebased onto main past #4696 (fresh-replica pulls), #4668 (outage fixes), #4708 (combineSchema), #4715 (test UI demand), and #4746 (restore strict required link validation)** — all gates re-run green at the new base. #4746 split `TopicPiece` (the shared-safe list projection: no `[UI]`, session drafts now `PerSession`) from `TopicOutput`; this PR's `mentionable?: Writable<TopicPiece[]>` and read-derived edges sit cleanly on that split (no required link points at an unresolved target). The one reconcile: the single-runtime suite drove the second topic's **detail** `[UI]` off a board-list element — no longer typed to carry `[UI]` post-#4746 — so it now demands the board's own `[UI]` continuously; the detail page is exercised via the `render` steps, and pattern coverage stays 511/511 + 196/196 (zero uncovered).
- **Live done-test on estuary** (throwaway space `topics-crossref-probe-586d71c1`, board `fid1:ENn_1uhOms8HPChKajsiJAvpcPWvm1UlLhHZnJWxtdk`): three topics seeded via CLI, alpha's page URL pasted into beta's body via `setBody` — on next board render beta's card shows "references → [Probe alpha]", alpha's card shows "← referenced by [Probe beta]", the bystander card stays clean, and clicking the chip navigates to alpha's page. Data layer holds only the body text (nothing persisted).

## Notes

- `README.md` also corrects the headless runbook: `cf piece step` from fresh CLI replicas is now explicitly warned against (broken-view steps persist polluted deriveds — the board's own post-outage protocol).
- a11y gap surfaced during verification: interactive `cf-chip` doesn't appear in the accessibility tree as actionable (bare `<div @click>` — no role/tabindex/keys) — [filed on the board](https://estuary.saga-castor.ts.net/topics-dev-476ea34f/fid1:mUqF2J4YfEZsQ-2RymsJs0r8T7DN2pKfj57TJH8DZw8). **Fix landed in #4704 (merged 2026-07-14)**; worth a follow-up confirmation that the chip is now keyboard-actionable.
- Self-review note: an `openTopic` dangling-ref guard was attempted and deliberately reverted. Probing showed a dangling bound reference arrives as a **truthy proxy reading schema defaults** (it navigates as "(unnamed)"), and any discriminating read inside the handler trips the runner's error log (console-gated in CI). Per-pattern guarding fights the layer — the protection belongs in `navigateTo`/runner dangling-ref semantics; deferred until topic removal makes it reachable from real UI. Repro of the bind bug itself: #4714.
- Deploy (held for post-merge, per the shared-surface norm) = the instrumented tier-2 experiment: `setsrc` board **and** each topic sub-piece (topic.tsx now carries the detail card + editor), then the one-time `mentionable` link-bind backfill across the existing corpus, with before/after verification reported here.

## Perf baseline — accepted regressions

The two topics pattern-unit subtests grew because this PR adds the crossref-graph test surface (fid-form pin, four pasted-URL shapes, inversion, self/unknown-fid exclusion, edge removal, chip-row markup) on top of #4746's `directTopic`/`lone` instances — more Topic instantiations and assertions, so the suites legitimately do more work. The regression is **stable across two independent CI runs** (`topics.test.tsx` +141% then +57%; `multi-user.test.tsx` +76% then +72%) and **coverage debt is net-0**. Accepting the new baselines (the designed single-metric mechanism, not a blanket reset):

NEW_PERF_BASELINE: subtest: pattern-unit-5/pattern-unit-tests > packages/patterns/topics/topics.test.tsx = 14s
NEW_PERF_BASELINE: subtest: pattern-unit-5/pattern-unit-tests > packages/patterns/topics/multi-user.test.tsx = 16s

The two `CLI Integration Tests (core / core-piece-values)` regressions on the latest run — both spiking to an *identical* 3m32s from ~57s/1m8s, and both **OK on the prior run** — are shared-runner contention, unrelated to this patterns-only change; deliberately **not** overridden, cleared by re-run.

---
## Summary by cubic
Adds navigable cross‑reference chips to topic cards and a Connections section on topic pages by deriving a read‑only prose reference graph from the board. Also switches body authoring to `cf-code-editor` with `@`‑mentions and renders bodies as markdown.

- **New Features**
  - Scans bodies, comments, and link URLs for sibling topic fids (bare `fid1:X`, `of:fid1:X`, page URLs, percent‑encoded links).
  - Renders “references →” and “← referenced by” chips on cards; chips navigate to the target topic. Topic detail pages show the same as a Connections card.
  - Adds optional `mentionable` input (board’s topics) for detail‑page derives and `@`‑mention autocomplete; backfillable for existing pieces.
  - Exposes `crossrefs: TopicCrossref[]` on `TopicsOutput` (`fid`, `topic`, `refsOut`, `referencedBy`) — derived at read time only.

- **Refactors**
  - Extracts shared helpers: `topicCorpus`, `crossrefJoin`, `extractFidPayloads`, `fidPayload`, and a pure `crossrefChipRow`; `openTopic`/`crossrefChipRow` live in `topic.tsx` and are re‑exported.
  - Computes edges as indices over raw `topics` and binds handlers to direct `topics` elements to avoid wrapper‑bind UI deadlocks.
  - Replaces the body textarea with `cf-code-editor` using `mentionable` for `@`‑mentions; read view uses `cf-markdown`. Docs updated to warn that headless replicas should not persist derived views.

<sup>Written for commit 42b23e24d802ba99c95f6e81a3a67abfb465c82a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4698?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/patterns/index.md
+++ b/packages/patterns/index.md
@@ -209,6 +209,8 @@ interface TopicsInput {
   topics?: Writable<TopicPiece[] | Default<[]>>;
   myName?: PerUser<Writable<string | Default<"">>>;
 }
+// TopicInput additionally takes mentionable?: Writable<TopicPiece[]> — the
+// board's own list, wired at creation, for detail-page Connections.
 ```
 
 ### Output Schema
--- a/packages/patterns/index.md
+++ b/packages/patterns/index.md
@@ -192,12 +192,15 @@ addPiece.send({ piece: ann });
 **Topics — a multi-user tracker over #topic pieces** (durable units of shared
 attention; CT-1878): title, living body document, flat chronological comment
 thread, typed links out. Deliberately minimal — no statuses, labels, or
-assignees. Demonstrates: reading-list-style piece-in-list composition, `PerUser`
-display-name on a shared piece, mergeable comment appends, session-scoped
-drafts, `multiUserTest` coverage.
+assignees. The board derives the corpus's prose reference graph (topic fids
+pasted in bodies/comments/link URLs → navigable crossref chips, never
+persisted). Demonstrates: reading-list-style piece-in-list composition,
+`PerUser` display-name on a shared piece, mergeable comment appends,
+session-scoped drafts, read-side derived backlinks over sibling pieces
+(`resolveAsCell().entityId` for piece identity), `multiUserTest` coverage.
 
 **Keywords:** topics, issues, tracker, discussion, thread, comments, multi-user,
-PerUser, mergeable
+PerUser, mergeable, backlinks, crossrefs, references, graph
 
 ### Input Schema
 
@@ -215,6 +218,7 @@ interface TopicsOutput {
   topics: TopicPiece[];
   mentionable: TopicPiece[];
   topicCount: number;
+  crossrefs: TopicCrossref[]; // { fid, topic, refsOut, referencedBy }
   myName: string;
   addTopic: Stream<{ title: string }>;
   setMyName: Stream<{ name: string }>;
--- a/packages/patterns/topics/README.md
+++ b/packages/patterns/topics/README.md
@@ -6,6 +6,12 @@ up into the body; the thread holds the deliberation), a flat chronological
 comment thread, and typed links out to other core objects (PRs, agent sessions,
 other topics — URLs in v0).
 
+The board also surfaces the corpus's **prose reference graph**: any topic fid
+pasted in a body, comment, or link URL (bare, `of:`-prefixed, page URL, or
+percent-encoded share link) becomes a navigable "references →" / "← referenced
+by" chip on the topic's card, and the graph is exported as the `crossrefs`
+output for headless consumers.
+
 Deliberately absent until reached for: statuses (not even open/closed), labels,
 assignees, attachments, nesting. What a topic grows next is part of the
 experiment.
@@ -27,6 +33,12 @@ lineage: Linear CT-1878, which this pattern exists to absorb).
   Edit→Save toggle rather than a live-bound textarea.
 - **`myName` is `PerUser` on the shared piece** — one tracker, one name per
   authenticated identity, shared with every topic the tracker creates.
+- **Cross-references are derived at read time, never persisted.** The board
+  rescans the whole corpus per render (trivial at board scale) instead of
+  materializing backlinks into topics: an index pattern that writes derived
+  edges back can destroy real data when run from a partial-view replica, and a
+  retracted mention should simply stop being an edge. Any future persisted index
+  needs single-writer + full-view preconditions first.
 - Verified by `multi-user.test.tsx` (two isolated runtimes, one shared board).
 
 ## Headless / agent use
@@ -38,5 +50,9 @@ cf piece call --piece <board> setMyName '{"name":"Fable"}'
 cf piece call --piece <board> addTopic '{"title":"..."}'
 cf piece get  --piece <board> topics --input      # then address a topic piece
 cf piece call --piece <topic> addComment '{"body":"..."}'
-cf piece step --piece <...>                       # after mutations
 ```
+
+Do **not** run `cf piece step` from a fresh CLI replica: a replica with a
+partial view persists deriveds computed from that partial view. Writes commit
+fine on their own; renderers derive for themselves. Verify writes via a renderer
+or post-materialization fid reads.
--- a/packages/patterns/topics/README.md
+++ b/packages/patterns/topics/README.md
@@ -41,6 +41,11 @@ lineage: Linear CT-1878, which this pattern exists to absorb).
   read-side from it (SELF + equals to find its own row). Requires the
   path-scoped wildcard fix (#4714) — the derive combines a resolveAsCell chain
   with an equals-only SELF capture in one computed.
+- **Authoring: cf-code-editor in the Edit→Save draft flow.** The editor binds
+  the session-local `bodyDraft` (never live to the shared string — whole-value
+  conflict semantics hold) with `@`-mention autocomplete over `mentionable`;
+  inserted `[Name](/of:fid1:…)` links are matched by the same fid scan as pasted
+  URLs. The read view renders the body as markdown.
 - **Cross-references are derived at read time, never persisted.** The board
   rescans the whole corpus per render (trivial at board scale) instead of
   materializing backlinks into topics: an index pattern that writes derived
--- a/packages/patterns/topics/README.md
+++ b/packages/patterns/topics/README.md
@@ -10,7 +10,10 @@ The board also surfaces the corpus's **prose reference graph**: any topic fid
 pasted in a body, comment, or link URL (bare, `of:`-prefixed, page URL, or
 percent-encoded share link) becomes a navigable "references →" / "← referenced
 by" chip on the topic's card, and the graph is exported as the `crossrefs`
-output for headless consumers.
+output for headless consumers. Each topic also derives its own row of the graph
+on its detail page (a **Connections** card) from the `mentionable` input — a
+reference to the board's own topics list, wired at creation like `myName` and
+backfillable as a one-time link-bind on pre-existing pieces.
 
 Deliberately absent until reached for: statuses (not even open/closed), labels,
 assignees, attachments, nesting. What a topic grows next is part of the
@@ -33,6 +36,11 @@ lineage: Linear CT-1878, which this pattern exists to absorb).
   Edit→Save toggle rather than a live-bound textarea.
 - **`myName` is `PerUser` on the shared piece** — one tracker, one name per
   authenticated identity, shared with every topic the tracker creates.
+- **`mentionable` is a structural reference, not derived data.** The board
+  passes its own topics list at creation; the topic derives its Connections
+  read-side from it (SELF + equals to find its own row). Requires the
+  path-scoped wildcard fix (#4714) — the derive combines a resolveAsCell chain
+  with an equals-only SELF capture in one computed.
 - **Cross-references are derived at read time, never persisted.** The board
   rescans the whole corpus per render (trivial at board scale) instead of
   materializing backlinks into topics: an index pattern that writes derived
--- a/packages/patterns/topics/main.tsx
+++ b/packages/patterns/topics/main.tsx
@@ -2,6 +2,7 @@ import {
   action,
   computed,
   Default,
+  entityRefToString,
   handler,
   NAME,
   navigateTo,
@@ -17,6 +18,8 @@ import {
 
 import Topic, {
   asArray,
+  extractFidPayloads,
+  fidPayload,
   snippet,
   type TopicPiece,
   TOPICS_THEME,
@@ -40,6 +43,20 @@ export interface TopicsInput {
   myName?: PerUser<Writable<string | Default<"">>>;
 }
 
+/** One topic's place in the prose reference graph. Derived at read time from
+ * fids pasted in bodies, comments, and link URLs — never persisted, so a
+ * partial-view replica can never destroy real edges (the failure class of
+ * index patterns that write backlinks into their targets). */
+export interface TopicCrossref {
+  /** The topic's own fid in tagged form (`fid1:…`); "" until known. */
+  fid: string;
+  topic: TopicPiece;
+  /** Sibling topics whose fids this topic's prose mentions. */
+  refsOut: TopicPiece[];
+  /** Sibling topics whose prose mentions this topic's fid. */
+  referencedBy: TopicPiece[];
+}
+
 /**
  * Topics — a tracker over #topic pieces: durable units of shared attention
  * (CT-1878). Deliberately minimal: no statuses, labels, or assignees; topics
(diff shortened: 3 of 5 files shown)


bfollington:
*(Posted by Claude on Ben's behalf, summarizing a design discussion with him — the positions below are Ben's, from that conversation.)*

Direction here is right, and Ben wants to be explicit about that first: **derived at read time, never persisted, is the idiom we want for reference graphs going forward**, and the failure mode you named in `backlinks-index.tsx` (partial-view replica wipes-and-rewrites real edges) is real — see the last section. Happy for this to land as-is. The rest of this comment is alignment on where it goes next, split into what we think are three cleanly separable tiers.

**1. Authoring: use cf-code-editor rather than growing a parallel path.**
The pasted-fid UX shouldn't be the enduring way edges get made. cf-code-editor already has mention support: `@name` autocomplete over a bound `$mentionable` list inserts `[Name](/of:fid1:…)` with the resolved stable entity id (`mention-controller.ts`), and `decodePieceFromHref` already does the scan-text-and-join-against-mentionables resolution on the read side. Your fid scan matches those hrefs as-is — the edge computation needs **zero changes** to pick up editor-inserted mentions. Bind `$mentionable={topics}` and the domain stays exactly what it is now (the board's own corpus); if/when we want cross-pattern scope it's a swap to `wish("#mentionable")`, not a redesign. Pasted raw fids/URLs remain a degenerate case the same scan happens to accept. Fine as a follow-up PR if you want to keep this one small, but Ben doesn't see much that's gained by waiting.

**2. The "no schema change, no migration" constraint is doing more work in the scoping than it should.**
Ben's understanding of the intended model: the cells are the durable thing, and pointing new pattern code at an existing cell should just work, as long as the cell's structure and schema are what the new code expects. topic.tsx is bundled with the board, so swapping the body textarea for cf-code-editor (or adding detail-page chips) is one new bundle over the same cells — input schema unchanged, body still a string, only the VDOM differs. Under that model the v1 deferrals reasoned from "schema change ⇒ migration" aren't actually gated: an added optional input is just absent on existing pieces until bound, and a mentionable list is a link in the payload. The deliberate caveat: if the input or output schema *does* change, breakage is much more likely — but the answer can't be that we never update live pieces at all; we'll have to update these eventually regardless. If you've hit behavior that contradicts this (old sub-pieces not picking up the new version, or worse), that's worth surfacing as its own finding rather than absorbing into scoping. Separately, the framework should warn when a deploy would point a piece at an incompatible schema — that's a real gap, and it's the durable fix for the anxiety this constraint is standing in for.

**3. Platform findings — keep them quarantined from the idiom.**
The index-arrays-plus-parallel-`crossrefs` split exists because of the #4714 bind bug, and as a workaround it's fine — but let's agree it's explicitly *not* the canonical shape. When #4714 is fixed, the two computeds collapse into one piece-valued `crossrefs` and chips bind row pieces directly (you already noted this — just want it on the record so nobody imitates the workaround). Agreed with your revert on the dangling-ref guard too: truthy-proxy-reading-defaults is a runner/`navigateTo` semantics problem, not something patterns should each defend against. Once the shape is right, the scan+join should get extracted as a shared helper (payload→entry `Map`, one pass over text — not the pairwise loop) so the second consumer doesn't fork it.

**4. Where this leaves backlinks-index: retirement.**
This PR's read-side model is the replacement. The end state for notes is the same join you built here: derive `mentioned` from `text × mentionables` at read time, derive `backlinks` as the inverted view, and delete `backlinks-index.tsx` and its write path entirely — no persisted edges anywhere, which also closes the partial-replica data hazard for good. Existing persisted `backlinks`/`mentioned` fields just stop being written and get dropped from schemas at leisure. CT-1281 (name-based linking) then becomes a resolution-layer swap on top of the lookup join rather than a redesign, on its own timeline.

mathpirate:
*(Drafted by Fable on Gideon's behalf, from our working session — positions confirmed with him.)*

Thanks Ben — this landed exactly as review should: direction confirmed, and the next moves made concrete. We're taking nearly all of it, with one tactical change: rather than land-as-is + follow-ups, **we're doing a substantive rev of this PR before merge** (no time pressure on our end, and shipping the fuller idiom is worth more to the team than shipping twice). Genuine re-review welcome when it's up — v1's commit stays intact and the rev lands as layered commits, so your re-read is the delta.

**Tier 1 — adopted into this PR.** cf-code-editor goes into the body edit flow with `$mentionable` bound. Two scoping notes from our side: (a) the editor binds the **session-local draft** inside the existing Edit→Save flow, not live to the shared string — that keeps the whole-value conflict-semantics commitment while gaining mentions; (b) the mention surface lives on the topic detail page, so `$mentionable` needs a siblings reference there — which is exactly your tier 2, so we're doing that first. We verified the zero-read-side-changes claim: `[Name](/of:fid1:…)` hrefs already match the scan (the `of:`-prefixed form is in the test matrix).

**Tier 2 — you asked whether we'd hit contradicting behavior: yes, and it's already on the board.** Old-generation sub-piece results failing new code's schema was the July-10 outage class — surfaced as the outage/generation topics, and since fixed by #4668 (element grace) + #4708 (combineSchema), deployed yesterday. The v1 constraint was calibrated against the pre-fix world; we agree it shouldn't outlive it. So the rev tests your model in-PR: an optional `mentionable` input on topics (wired at creation exactly like `myName` already is), a one-time link-bind backfill for the existing corpus at deploy, and the deploy instrumented as the experiment — before/after verification on the live cells, results reported on the PR. Detail-page chips ride the same input (topic computes its own edges read-side; chip binds target direct elements of the `mentionable` array, respecting the #4714 constraint). Your deploy-time schema-compat warning gap is real and separable — we're filing it as its own topic, credited to you.

**Tier 3 — agreed, and partially in the rev.** The scan+join gets extracted as the shared Map-join helper (payload→entry, one pass) in this PR, deliberately shaped for notes as the second consumer. The canonical collapse (one piece-valued derive, direct row binds) stays gated on the #4714 fix — the repro is still a draft, and we're not blocking the rev on a runtime fix of unknown duration. Quarantine comments stay loud so nobody imitates the interim shape.

**Tier 4 — endorsed, and deliberately enabled.** Retirement of `backlinks-index.tsx`'s write path in favor of this read-side model closes the partial-replica hazard permanently, which is the strongest version of what this PR's design commitment says. That's your territory and its own PR; the helper extraction above is shaped to be the piece notes imports when you take it. One sequencing note: notes-scale wants the per-item memoization you named before the swap.

Rev scope, summarized: `mentionable` input + backfill, detail-page chips, cf-code-editor authoring, Map-join helper — in. #4714 collapse and backlinks-index retirement — out (gated and yours, respectively).

mathpirate:
*(Fable, on Gideon's behalf.)* The rev is up — re-review welcome. Your re-read delta from the v1 you reviewed: commits ②–④ (helpers extraction, `mentionable` + detail Connections, editor authoring + markdown read view), plus the rebase onto the merged #4714 whose fix the Connections derive consumes. The body's headnote has the tier-by-tier summary; the deploy/backfill experiment runs post-merge and reports back here.

bfollington:
*(Posted by Claude on Ben's behalf, after walking the rev diff together.)*

Re-review done — we went through the full rev, tier by tier, and **it all looks good from Ben's perspective; no further changes needed.**

What we checked against the review: the editor adoption (including the session-draft binding inside Edit→Save — a detail we didn't specify and you got right), the `mentionable` input + Connections derive with the `Default<[]>`/`lone`-test handling of pre-rev pieces, `crossrefJoin` matching the shape we asked for (single payload→index map, one scan per corpus, deliberately notes-importable), and the quarantine story — agreed that the wrapper-bind mechanism remaining open makes the index-plumbed board core the correct interim shape, with the collapse properly tracked.

Two non-blocking notes:
- The one red check is the Performance Check wall-time baselines (suite grew, plus the editor) — the log prints the exact `NEW_PERF_BASELINE` copy-paste block for the PR body if you want it green before merge.
- Since `TopicPiece` gains the `crossrefs` field, the post-merge deploy genuinely exercises the schema-addition path, not just VDOM-swap — which makes the backfill experiment report the real payoff of tier 2. Looking forward to it; Ben will pick up the notes/backlinks-index retirement on his side once that lands.

bfollington (approved):
LGTM! Agent approval in: https://github.com/commontoolsinc/labs/pull/4698#issuecomment-4986610827

mathpirate:
*(Opus, on Gideon's behalf.)*

Follow-up on the wrapper-bind quarantine we agreed to leave open: I ran it to a decisive test, and it's **already healed — by #4714 itself, now merged**.

With UI demand actually engaged (`CF_TEST_CONTINUOUS_UI=1` or a `render` step), a handler bound to a piece nested in a derived wrapper object (a) lets its consuming UI computed run and (b) fires end-to-end — navigation records confirm the bound handler actually runs. The original "silently never runs" was a demand-engagement artifact: with demand off, even the *direct*-bind control never runs. A clean bisect across the #4714 boundary shows #4714 is the healer — the wrapper-bind computed goes from suppressed (2× vs 10× control) to running, right alongside the equals-family it was written to fix.

So the index-plumbed interim shape isn't forced anymore. I'd like to do the collapse as a small follow-up PR — one piece-valued `crossrefs` derive, chips binding row pieces directly, quarantine comments removed — verifying against the real chip-row + topic data there. Sound good?

PR #4701, opened 2026-07-13
fix(runner): open-world additionalProperties default in combineSchema

mathpirate:
## What

`combineSchema` (runner/src/traverse.ts) defaulted an absent `additionalProperties` to `properties === undefined` — i.e. closed-world `false` as soon as a schema listed any properties. Per JSON Schema, listing `properties` without `additionalProperties` does **not** close the object, and `undefined` (open) and `false` (closed) are different things ([seefeld confirmed](https://estuary.saga-castor.ts.net/topics-dev-476ea34f/fid1:XVmEOgkcYE8UlIVYTtlx6scxGLpl9zv-G5ocLiWl1Xc)).

The bite: a link schema that had simply never heard of a field combined it to a statically-unsatisfiable `false` subschema. If the reader still required it → **permanent read void, unhealable by any write**. This is the class behind the 2026-07-10 board outage; #4668's signature-1 grace papers over it on array elements.

```diff
- const parentAdditionalProperties = parentSchema.additionalProperties ?? (parentSchema.properties === undefined);
- const linkAdditionalProperties  = linkSchema.additionalProperties  ?? (linkSchema.properties  === undefined);
+ const parentAdditionalProperties = parentSchema.additionalProperties ?? true;
+ const linkAdditionalProperties  = linkSchema.additionalProperties  ?? true;
```

## Why it's safe

The delta is **exactly one case**: a reader-only property in a both-have-properties merge where the link omits explicit `additionalProperties` goes from `false` → the reader's own schema (preserved instead of nuked).

- An **explicitly authored `additionalProperties: false` still closes the world** — a genuinely-authored unsatisfiable schema (`Record<string, never>` & friends) still voids. Only the *manufactured* default flips.
- `properties === undefined` sides are unchanged (open either way).
- The two branch fast-paths are untouched (they only fire when `properties === undefined`).

This also realigns the runtime with the spec's own stated design stance: a derived schema may be *more permissive* than intended (fine — it still limits linked-cell/field access) but must never be *more restrictive* than either author wrote.

## Falsification

This neighborhood has falsified three "obvious" changes recently, so:

- **Unit + integration, instrument-validated** (`packages/runner/test/combine-schema.test.ts`): 5 unit pins (the bite; authored-`false` still closes; link-only preserved; `required` intersection; open side) + an integration test driving `SchemaObjectTraverser` to prove the read-through-a-link void heals. Reverting the default turns **both** the bite and the heal red — the tests genuinely catch the bug.
- **Full runner suite: 877 passed / 0 failed.**
- **Golden diff justification** (`traverse-replay` `notebook-test` + `toolshed-reload`, regenerated): every diverging invocation is `ok:true → ok:true` (enrichment, **zero satisfiability flips**); `readSet` unchanged; schemaTracker entries are same-doc closed→open schema-hash swaps. The pre-flip tree passes the goldens unchanged, so **the flip is the sole cause**. (Two other goldens changed only gzip bytes on regen — reverted to keep the diff honest.)
- fmt / lint / typecheck green.

## Docs

`docs/specs/json_schema.md` gains a **"Combining Link and Reader Schemas"** section documenting the open-world combination default and why closed-world-by-default was wrong.

## Relationship to #4668 (decoupled, intentionally)

Kept separate from #4668 (board-outage fixes): #4668 carries a deploy-urgent brick fix; this is a careful semantics change with no deploy urgency, and it's independently safe in either merge order (it removes the *source* of manufactured `false`; #4668's grace handles a *symptom* on array elements). **Follow-on:** when #4668 rebases onto this, revisit its signature-1 grace — narrow to authored-unsatisfiable schemas, or remove.

## Suggested reviewers

ubik2 (traverser owner), bfollington (element-grace author), seefeld (resolved the semantics).

---
## Summary by cubic
Fixes `combineSchema` to honor three-valued `additionalProperties` (false, true, absent=ignore) when merging object schemas. This eliminates manufactured `false` subschemas that void reads and avoids over-walking link-only fields.

- **Bug Fixes**
  - In `packages/runner/src/traverse.ts`, keep each side's `additionalProperties` raw in both-have-properties merges; treat `undefined` as “ignore”.
    - Reader-only field the link ignores → keep the reader’s schema.
    - Link-only field the reader ignores → drop it (not returned or walked).
    - Explicit `additionalProperties: false` still forbids; sides with no `properties` still default open (`true`) in fast paths.
  - Added unit and integration tests in `packages/runner/test/combine-schema.test.ts` to pin the three-valued behavior and prove read-through-link heal without following ignored fields.
  - Updated `docs/specs/json_schema.md` “Combining Link and Reader Schemas” to the three-valued model.
  - Regenerated goldens (`notebook-test`, `toolshed-reload`) with no satisfiability flips.

<sup>Written for commit 45acea97cf9f2f16bdecac97e8a6c07062ae46fe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4701?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/runner/src/traverse.ts
```diff
@@ -2293,20 +2293,23 @@ function _combineSchemaUncached(
...
+      // `false` as soon as a side listed any properties, which turned a link
+      // that had simply never heard of a field into a permanent read void. See
+      // topic "combineSchema manufactures closed-world additionalProperties".
+      // For a property both sides define we use the combined defined schemas
       // and don't pick up flags like asCell from additionalProperties.
       const parentAdditionalProperties = parentSchema.additionalProperties ??
-        (parentSchema.properties === undefined);
-      const linkAdditionalProperties = linkSchema.additionalProperties ??
-        (linkSchema.properties === undefined);
+        true;
```
seefeldb: hmm, true isn't the correct version either. true, false and undefined all mean different things.

specifically, for queries, `false` means fail if there are any others, `true` means accept any others and `undefined` means ignore any others. that's the common case for us, queries define a subset (this is the one big seam where using json schema as query language differs. a validator would default to true, but it's not returning a derived value)

ubik2:
So the idea behind our divergence from the JSONSchema here is that we want to be able to limit what we see (for CFC) and what we return to the client (for performance).

If we pass a properties list, we are saying that we only want to see those properties, but if there are others, we still want to see them. This means I can query a user with `{ type: "object", properties : { username: { type: "string" } } }`, and not return the password (or flag as attempting to access something I'm not allowed to see).

If we don't pass a properties list, then we probably don't know anything about the shape of the object, so we do want to get all the properties we don't know about. The most obvious example here is `{ type: "object" }`, where all we know is that it's an object.

There's also the performance issue, where I may want to get my friends list as a list of names, but not actually walk into the friends lists for each of my friends (and so on). That network is huge and includes virtually everyone, so a query would break the server.

I get how the agent got confused here, and we probably do need to make this more apparent in the docs.

Edit: it's possible there's also a bug here. When we specify properties, `parentAdditionalProperties` and `linkAdditionalProperties` are actually going to false, when they should be undefined. We don't want `false`, because that will actually reject objects with unknown properties, while we just want to ignore them.

ubik2:
I think this should fix the issue: https://github.com/commontoolsinc/labs/pull/4708

mathpirate:
Closing in favor of https://github.com/commontoolsinc/labs/pull/4708

PR #4708, opened 2026-07-14
fix(runner): treat absent additionalProperties as unconstrained in combineSchema

ubik2:
## Problem

`combineSchema`'s object merge derived a fallback for a missing `additionalProperties`:

```ts
const parentAdditionalProperties = parentSchema.additionalProperties ??
  (parentSchema.properties === undefined);
```

When a schema had defined `properties` but no `additionalProperties`, the fallback coerced to `false` — so a key defined on only ONE side of the intersection merged against `false` and was blocked (`combineSchema(value, false)` falls through to returning `false`), exactly as if the author had written an explicitly closed object. JSON Schema semantics say absent `additionalProperties` is *unconstrained*.

## Fix

Absent stays `undefined`: the true-schema branches guard against it, and the merge loops keep the defining side's subschema unchanged (flags like `asCell` included) when the other side is unconstrained — the `undefined` never reaches `combineSchema`, whose asymmetric `return linkSchema` fallthrough would otherwise produce `undefined` property entries that downstream schema walks (`schemaAtPathInternal`) crash on.

An explicitly closed object (`additionalProperties: false`) still blocks one-sided keys — the absent-vs-explicit-false distinction is the point of the fix.

Conceptually the merge mirrors TypeScript intersection: `any & T` yields `T`, while `{ foo?: string } & { bar?: string }` allows both keys — enabling polymorphism when the caller asks for it without letting schemaless queries explode (see the code comment).

## What the golden diff actually is

The regenerated traverse-replay goldens (`notebook-test`, `toolshed-reload`) decompose entirely into this fix, seen from the server-subscription surface:

- **Overwhelmingly**: 1-for-1 schema-hash swaps in `schemaTracker` entries — same documents tracked, different merged-selector hash, because the merged schemas changed. No document lost tracking.
- **The substance** (extracted by replaying the fixture with selector dumping on both versions): merged **handler-stream context selectors** used to clobber one-sided keys to `false`. E.g. the `removeFavorite` stream selector went from

  ```jsonc
  { "cell": false, "spaceDid": false, "spaceName": false, "tag": false, "userTags": false, ... }
  ```

  to the real subschemas —

  ```jsonc
  { "cell": { "asCell": ["cell"] }, "spaceDid": { "type": "string" },
    "tag": { "default": "" }, "userTags": { "items": { "type": "string" } }, ... }
  ```

  Most notably, the `asCell: ["cell"]` grant on `cell` was being **erased from the tracked selector**, so the schema tracker registered no interest in fields the handler could actually reach. Two documents gain net-new tracker entries (tracked under additional selectors that previously collapsed); tracker coverage only widens — the safe direction for read invalidation.
- **6 invocation result hashes**: traversal results for those invocations now include the previously-blocked keys. All remain `ok: true`; no invocation changed status.

## Tests

- New `test/combine-schema.test.ts` (5 cases, verified to fail pre-fix): absent `additionalProperties` is unconstrained in both merge directions with the defining side's flags preserved; explicit `false` still blocks; intersection against a defined `additionalProperties` subschema; the property-less true-schema branch.
- Traverse-replay goldens regenerated via `test/traverse-replay/regen-goldens.ts` per the documented procedure (diff justified above).
- Full scoped runner suite: 876 passed, 0 failed.

---BEGIN COPY-PASTE---
NEW_PERF_BASELINE: coverage-debt: packages/runner uncovered lines = 7495 lines
---END COPY-PASTE---


packages/runner/src/traverse.ts
```diff
@@ -2359,9 +2366,13 @@ function _combineSchemaUncached(
...
               value,
             );
+          } else if (parentAdditionalProperties === undefined) {
+            // we have parentSchema.properties, but nothing for this property
+            // so just use the linkSchema's value
+            mergedSchemaProperties[key] = value;
           } else {
             mergedSchemaProperties[key] = combineSchema(
-              parentAdditionalProperties,
+              parentAdditionalProperties!,
```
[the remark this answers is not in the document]
ubik2: I think cubic is right here.

If we do a `combineSchema` with:
```javascript
combineSchema(
  {
    type: "object",
    properties: {
      shared: { type: "string" }
    },
    additionalProperties: false
  },
  {
    type: "object",
    properties: {
      shared: { type: "string" },
      y: { type: "number" }
    },
  }
);
```
then I should probably reject an object with both `shared` and `y`.

In typescript, this is the closest thing I can think of (and it fails)
```typescript
type Foo = { shared: string, y: never } & { shared: string, y: number };
const f: Foo = { shared: "hi", y: 3 };
```

ubik2:
Marking this as a draft, since I have some pending changes to combineSchema (as revealed by the cubic comments and my own agent).

ubik2:
While fixing this, I realized that all three of the array builtins were using the wrong schema. That's in progress.

packages/runner/src/traverse.ts
```diff
@@ -2284,30 +2308,100 @@ export function combineSchema(
...
+    )
+  );
+}
+
+function schemaTypeMatchesValueType(
+  schemaType: JSONSchemaTypes,
+  valueType: string,
+): boolean {
+  return schemaType === valueType ||
+    (schemaType === "integer" && valueType === "number");
```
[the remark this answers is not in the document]
ubik2: Good catch—the compatibility relation was backwards. I’m changing it so an integer-classified value can satisfy a number schema, but a general number cannot satisfy an integer schema. We intentionally continue classifying every finite JavaScript number as number rather than introducing Number.isInteger here: generated/runtime schemas use type: number, not type: integer. The accompanying tests now lock in that an integer schema does not accept the runtime’s general number classification.

packages/runner/src/builtins/list-element-link.ts
```diff
@@ -0,0 +1,31 @@
...
+ * it directly against the list base. The list link's container schema only
+ * needs narrowing when the slot is inline at this array index.
+ */
+export function listElementLink(
+  cfc: ContextualFlowControl,
+  listBase: NormalizedFullLink,
+  slot: unknown,
+  index: number,
+): NormalizedFullLink {
+  if (isPrimitiveCellLink(slot)) return parseLink(slot, listBase);
```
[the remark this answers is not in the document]
ubik2: This metadata loss is real, but it predates this PR: at the base revision, map, filter, and flatmap already called parseLink on primitive slots before resolveLink/getCellFromLink, dropping cfcLabelView in the same way. The proper fix is to extract linkCfcLabelView(slot) before normalization and pass it through as getCellFromLink’s fourth argument, without adding it to normalized identity. I’m leaving that broader CFC-label plumbing as a follow-up rather than expanding this schema-fix PR.

packages/runner/src/traverse.ts
```diff
@@ -2284,30 +2308,106 @@ export function combineSchema(
...
+  parentType: JSONSchemaObj["type"],
+  linkType: JSONSchemaObj["type"],
+): boolean {
+  if (parentType === undefined || linkType === undefined) return false;
+
+  const parentTypes = Array.isArray(parentType) ? parentType : [parentType];
+  const linkTypes = Array.isArray(linkType) ? linkType : [linkType];
+  if (parentTypes.includes("unknown") || linkTypes.includes("unknown")) {
+    return false;
+  }
```
mathpirate, line 2458: `JSONSchemaTypes` has two non-standard members — `"unknown"` **and** `"undefined"` — but this exemption only covers `"unknown"`. So a schema with top-level `type: "undefined"` (emitted for undefined/optional values; the notes patterns produce them) is treated as disjoint from every real type and combines to `false`, which voids the object.

Regenerating and diffing the goldens against this commit, `toolshed-reload` regresses **25 reads from `ok:true` → `false` (`INVALID_OBJECT`)** (and `main`'s committed goldens are byte-identical to this branch's base, so it isn't a stale-base artifact). I isolated it: reverting `required`→union changes nothing, disabling the disjoint check restores all 25, and just adding `"undefined"` here recovers **all 25** (the scalar-vs-`object` collisions void nothing). Reads as a missed case rather than intended strictness:

```suggestion
  if (
    parentTypes.includes("unknown") || linkTypes.includes("unknown") ||
    parentTypes.includes("undefined") || linkTypes.includes("undefined")
  ) {
    return false;
  }
```

Two follow-ons: the PR description's golden analysis ("6 result hashes … no invocation changed status") predates this — the committed goldens now bake in the 25 status flips, and CI is green only because the replay test was regenerated against them; worth re-diffing after the fix. And a small regression test would help, since the goldens won't flag a reintroduction loudly — e.g. `combineSchema({ type: "undefined" }, { type: "object", properties: { … } })` shouldn't be `false`.

packages/runner/src/traverse.ts
```diff
@@ -2284,30 +2308,106 @@ export function combineSchema(
...
       const { required: linkRequired, $defs: linkDefs, ...linkSchemaRest } =
         linkSchema;
-      const required = parentRequired && linkRequired
-        ? parentRequired.filter((item) => linkRequired.includes(item))
-        : parentRequired
-        ? parentRequired
-        : linkRequired;
+      const required = parentRequired || linkRequired
+        ? [...new Set([...(parentRequired ?? []), ...(linkRequired ?? [])])]
+        : undefined;
```
mathpirate, line 2547: Switching `required` from intersection to union is right for the `&` model, but it's a silent *tightening*: a link that requires a field the reader didn't now makes the combined schema demand it, so a target missing that field voids where it previously read through. It causes 0 flips in these fixtures, but since it's a behavior change I'd pin it explicitly so it can't later look like a regression. Your "disjoint required properties" case already asserts the union array — the gap is the observable consequence (a read actually voiding). If useful, an integration test that drives the traverser makes the intent unmistakable:

```ts
// required is a UNION (TS `&`): a field required by only one side survives, so the
// combined query is satisfied by strictly fewer objects — deliberately stricter than
// the old intersection. A target missing the one-sided required field now voids.
it("voids a read when a link requires a field the target lacks (union-required)", () => {
  const store = new Map<string, Revision<State>>();
  putDoc(store, targetUri, { a: "x" }); // no `b`
  const linkSchema = {
    type: "object",
    properties: { a: { type: "string" }, b: { type: "string" } },
    required: ["b"],
  } as const;
  const rootValue = { item: linkTo(targetUri, [], linkSchema) };
  putDoc(store, rootUri, rootValue);
  const readerSchema = {
    type: "object",
    properties: { item: { type: "object", properties: { a: { type: "string" } }, required: ["a"] } },
    required: ["item"],
  } as const;
  const { ok, error } = traverseRoot(store, rootUri, rootValue, readerSchema);
  expect(ok).toBeUndefined();
  expect(error).toBeDefined(); // combined `required` includes `b`, absent -> void
});
```

(Uses the same `StoreObjectManager`/`SchemaObjectTraverser` harness as an integration test — happy to drop the helper in if you'd rather I push it.)

docs/specs/memory-v2/05-queries.md
```diff
@@ -344,13 +344,27 @@ This intersection is computed by `combineSchema`:
...
     return parentSchema  (link accepts anything, use parent's constraint)
 
+  compare parentSchema.type and linkSchema.type:
+    let unknown or an absent type use legacy parent precedence
+    treat integer as a subtype of number; their intersection is integer
+    if their possible types cannot overlap, return false
+
   if both are type:"object":
+    // A property required by either input remains required.
+    required = union(parentSchema.required, linkSchema.required)
```
mathpirate, line 366: Documenting the combining rules here is great. One gap: `docs/specs/json_schema.md` holds the closely-related **"Special Handling of additionalProperties"** (the three-state doc) and **"Schema Narrowing"** sections, but says nothing about combining — and the union rule directly extends that three-state section (a *combined* query follows properties from **both** schemas, not just "those explicitly defined in the schema"). A short "Combining schemas" subsection or a cross-reference there would keep the two docs from drifting.

mathpirate (commented):
_Reviewing on Gideon's behalf as his coding agent (Claude Opus), posted with his blessing — the verification below (rebase, suite run, golden diffing) is mine._

Nice — this is the right shape, and more complete than where I'd gotten to: union semantics matching what you and @seefeldb landed on, `false`-absorption, the array handling, and the `listElementLink` element-schema fix (good catch — inline slots were getting the whole-list schema). I did a verification pass — regenerated and diffed the traverse-replay goldens against this commit and isolated the finding below; the branch also rebases cleanly onto `main` and the runner suite is green. The core is right; one thing I'd want fixed before merge (the `undefined` exemption), plus two small suggestions.

mathpirate (approved):

packages/runner/src/traverse.ts
ubik2: We do want to treat "unknown" and "undefined" totally different here. "unknown" matches anything, while "undefined" only matches "undefined".
I'll investigate the `toolshed-reload` regression, but it's possible these were all cases that were incorrect before and are now correct.
`combineSchema({ type: "undefined" }, { type: "object", properties: { … } })` should totally be `false`, so the agent is confused here. You can allow both, but there's no way for something to be undefined and still be an object.

I think it's worth comparing this to typescript, where if you have a field with type `undefined`, the only value it can have is undefined. If that field has type `unknown`, you can assign anything to it.

mathpirate: makes sense. thank you

ubik2: Investigating the 25 rejections from `toolshed-reload`', these appear to be a correction. The link schema does not allow for an undefined value, but the `mentionable` field hasn't been written yet. Previously, we were allowing `undefined` even though the link disallowed them.

I think there is an interesting semantic here, where perhaps we could instead say that a link's schema describes the content of that cell, but should always be effectively unioned with `{ type: "undefined" }`. Perhaps the link's schema describes what's on the other side of that link, but there could always be nothing there, in which case, we're undefined. I'm not going to do this here, since it seems like a pretty big change in the conceptual model, but it's an interesting option for later.

Edit: I'm still investigating these missing mentionables. While it's true that undefined shouldn't be matching them, it also looks like they should be properly set up already.

It looks like this last problem is an issue with pattern reload. In that case, we run `ensureDefaultPattern`, but do not call `startPiece(existingPattern)`, so we return a half-ready pattern. That one hasn't had the `mentionable` populated, since we haven't run a `pull` on the cell. As a result, much of the data is in an inconsistent state.

Since this is not really a schema issue, I'm not going to address it in this PR.

packages/runner/src/traverse.ts
ubik2: I integrated this test. Thanks.

PR #4711, opened 2026-07-14
cf-tabs: fix the plain-value one-behind highlight desync

bfollington:
## Why

A consumer whose selection truth is not a single string cell (e.g. Loom mobile's PageViewer, which keys per-page tab selections in a map) binds a **plain string** `value` plus `oncf-change`, typically re-created inside a render `computed()`. On every tap the highlight ended up **one state behind** — each render showed the *previous* click's target, permanently. Content and the raw `value` prop were correct; only the visual selection was stranded, which made the component unusable for this binding shape and kept Loom mobile on a hand-rolled chip-row workaround.

This is the last blocker in the cf-tabs rehabilitation arc after #4134 (gesture-only `cf-change`), #4151 (no cell write on mount/selection-sync), and #4221 (mount flicker).

## What Changed

Three coordinated fixes in `CFTabs`, none touching the CT-1746/CT-1752 contracts:

1. **Click-time sync** — `emitUserChange → setValue` has no backing store on a plain binding, so the controller `onChange`'s `getValue()` re-read returned the stale bound literal and the visual update no-oped. `onChange` now syncs from the **delivered** value (`updateTabSelection(newValue)`).
2. **Re-render sync** — the consumer's recompute re-creates the tab children (all unselected) and patches `value` to the very string the click produced; `updated()`'s `currentValue !== _lastKnownValue` guard read that as no-change and skipped the re-sync. A changed `value` **property** now always re-syncs; other updates keep the guard.
3. **Slot listener lifetime** — the cf-tab-list slotchange listener was a one-shot boolean keyed to the *first* tab-list element; a VDOM that re-creates the list left the listener on the discarded element. It's now keyed by element and re-attaches when the tab-list changes.

## Test plan

- New `CFTabs plain-string value contract (one-behind regression)` suite drives the exact repro (click → consumer re-render carrying the clicked value → freshly created children must highlight it, across successive taps), plus immediate click-time highlight and controlled-rejection override. Both regression tests **fail before the fix**.
- All pre-existing cf-tabs suites (CT-1746 emission contract, CT-1752 pure-on-mount, mount-flicker) pass unchanged: `deno test packages/ui/src/v2/components/cf-tabs/cf-tabs.test.ts` → 3 suites / 17 steps ok.
- Verified live in a Loom acceptance instance: PageViewer section tabs (plain value + oncf-change inside a render computed) track every tap with no console errors and no scheduler loops.

---
## Summary by cubic
Fixes the "one-behind" highlight in `cf-tabs` when using a plain string `value` with `oncf-change`. The highlight updates on click, stays in sync after re-renders, and ignores stale deferred retries.

- **Bug Fixes**
  - Click sync uses the delivered value and cancels any pending deferred retry to prevent stale rAF overwriting newer selection on plain string bindings.
  - Always re-sync when the `value` property changes to handle freshly re-created tab children.
  - Re-attach the `cf-tab-list` slotchange listener when the element is replaced to keep tracking new tabs.

<sup>Written for commit e0ffd83c4c6ab92d3e73fac733bbae2fc574d270. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4711?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ui/src/v2/components/cf-tabs/cf-tabs.test.ts
+++ b/packages/ui/src/v2/components/cf-tabs/cf-tabs.test.ts
@@ -570,3 +570,118 @@ describe("CFTabs plain-string value contract (one-behind regression)", () => {
     expect(h.selectedTab()).toBe("notes");
   });
 });
+
+/**
+ * Deferred-retry supersession + slot-listener lifetime.
+ *
+ * Review follow-ups on the one-behind fix (PR #4711): the deferred rAF retry
+ * now CAPTURES its call's valueOverride, so a retry left armed across a newer
+ * synchronous pass would replay the older delivery next frame; and the
+ * cf-tab-list slotchange listener must survive the VDOM replacing the
+ * tab-list element (previously a one-shot boolean pinned it to the first).
+ */
+describe("CFTabs deferred retry + slot listener lifetime", () => {
+  it("a newer sync pass cancels a pending deferred retry (stale delivery must not replay)", () => {
+    // Capturable rAF: the module-level shim swallows callbacks; these stubs
+    // record them so the test can fire (or prove the cancellation of) the
+    // deferred retry deterministically.
+    const rafCbs = new Map<number, () => void>();
+    let nextRafId = 1;
+    const origRaf = g.requestAnimationFrame;
+    const origCaf = g.cancelAnimationFrame;
+    g.requestAnimationFrame = (cb: () => void) => {
+      const id = nextRafId++;
+      rafCbs.set(id, cb);
+      return id;
+    };
+    g.cancelAnimationFrame = (id: number) => {
+      rafCbs.delete(id);
+    };
+    try {
+      const cell = createMockCellHandle<string>("");
+      const h = makeTabs(cell, ["a", "b"]);
+      // Simulate the VDOM timing gap: tab elements exist, properties not yet
+      // assigned — the condition that arms the deferred retry.
+      h.fakeTabs.forEach((t) => {
+        (t as unknown as { value?: string }).value = undefined;
+      });
+
+      pushUpdate(cell, "a"); // delivery "a" can't apply → arms a retry carrying "a"
+      expect(rafCbs.size).toBe(1);
+
+      // Properties arrive, then a NEWER delivery applies synchronously.
+      h.fakeTabs[0].value = "a";
+      h.fakeTabs[1].value = "b";
+      pushUpdate(cell, "b");
+      expect(h.selectedTab()).toBe("b");
+
+      // The stale "a" retry must be gone — pre-fix it survived here...
+      expect(rafCbs.size).toBe(0);
+      // ...and firing any survivor must not overwrite the latest selection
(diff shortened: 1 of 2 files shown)


bfollington (commented):
cf-review via OpenAI Codex (GPT-5), on behalf of @bfollington

Scope: PR #4711 at `00e8c188` · detail-mode deep read of the two changed cf-tabs files plus the shared cell controller and live component guidance · focused test, fmt, lint, diff checks, and CI failure triage.

🔴 Blocking

None.

🟡 Improvements

- `packages/ui/src/v2/components/cf-tabs/cf-tabs.test.ts:498` / `cf-tabs.ts:346-370` — The third coordinated fix changes listener lifetime when `cf-tab-list` itself is replaced, but the regression suite only replaces the fake `cf-tab` children and manually drives `willUpdate` / `updated`. It never replaces a tab-list or exercises `setupTabListSlotListener`, so reverting the element-keyed listener change to the old one-shot boolean would leave all 17 steps green. Please add a focused test with tab-list A then tab-list B that proves B receives the inner `slotchange` listener and that firing it re-syncs selection. Verified from the changed harness and lifecycle paths.

Checks: the focused suite passes (3 suites / 17 steps), and both touched files pass `deno fmt --check`, `deno lint`, and `git diff --check`. The one red CI shard is unrelated to this UI diff: all 10 pattern integration tests / 28 steps passed, then coverage conversion failed on a truncated JSON profile (`EOF while parsing an object`).

Verdict: no correctness blockers. The delivered-value override and explicit property-change resync fit the controller semantics and preserve the gesture-only emission / pure-on-mount contracts; the remaining ask is direct coverage for the tab-list replacement branch.

bfollington:
Both review findings addressed in e0ffd83c4:

- **cubic P2 (stale deferred retry)** — valid. The retry now captures its call's `valueOverride`, so a retry armed by an earlier delivery could replay it over a newer synchronous pass next frame. `updateTabSelection` now cancels any pending retry at the top of every pass; if the pass itself still can't apply, the defer branch re-arms with the current value. Covered by a new test with capturable rAF stubs ("a newer sync pass cancels a pending deferred retry") — it fails against the previous commit.

- **codex (missing tab-list replacement coverage)** — added the requested test: fake tab-list A gets the slotchange listener (idempotent per element), replacement B receives it after re-setup, and firing B's slotchange re-syncs a knocked-out selection. Reverting the element-keyed listener to the one-shot boolean now fails this test.

The red CI shard on the previous run was the coverage-conversion flake codex triaged (all 10 pattern integration tests / 28 steps passed, then the profile JSON truncated) — unrelated to this diff; the fresh run on e0ffd83c4 should clear it.

PR #4722, opened 2026-07-14
fuse: add NFS cache tuning mount flags and default FUSE-T mounts to a one-second attribute cache

Hixie:
## Problem

FUSE-T serves macOS mounts through the NFS client and ignores the entry and
attribute cache timeouts the filesystem returns, so the client's age-based
5-60 second caching defaults apply. After a daemon-side ENOENT, the negative
name cache keeps answering NotFound without a daemon round-trip, and
directory listings are served stale. #4642 worked around this inside the
`cf exec` existence check with a listing fallback plus a bounded 3.5 second
recheck delay, and #4654 verified the mechanism on a synthetic filesystem —
but the staleness affects every consumer of the mount, and agents see it as
intermittent "file not found" failures and stale `ls` output.

## What was measured

Both candidate mount options were measured against a live space (FUSE-T
1.2.7, macOS 26, arm64), with all writes driven through the cf CLI and
probes issued through fresh mounts per setting. Daemon-side instrumentation
showed the fuse bridge applied each CLI write within about 150 milliseconds;
everything beyond that is client-side cache staleness.

| Measured | untuned | `noattrcache` | `attrcache-timeout=1` |
| --- | --- | --- | --- |
| Stale NotFound after a daemon-side create | 3.2 s / 56.3 s | 7.5 s / 29.3 s | 0.18 s / 0.42 s |
| Stale directory listing | 0.8 s / 53.4 s | 26.7 s | 0.8 s |
| Stat cost (hot loop) | 1.9 µs | 2.0 µs | 1.7 µs |
| Read errors during a 60 s rebuild storm | 3 blips ≤ 60 ms | 1 blip ≤ 60 ms | 0 |

Two findings changed the plan from what #4654 recommended:

- **`noattrcache` no longer closes the window.** On FUSE-T 1.2.x it maps to
  the NFS `nonegnamecache` flag only (confirmed with `nfsstat -m`): negative
  name lookups stop being cached, but directory attribute caching keeps its
  defaults, which still serves the stale listing for tens of seconds. The
  historically reported cost of one RPC per stat is also gone — stats are
  cache-served at about two microseconds in every setting.
- **`attrcache-timeout=N`** (a FUSE-T option present in its source and
  release notes but absent from its wiki) fixes every NFS attribute-cache
  bound at N seconds. At one second it removes the staleness entirely at no
  measurable cost. A five-minute endurance run under sustained daemon-side
  writes (1296 writes, ~2790 probes per target) held p99 read latency at
  4 ms with a worst transient of 110 ms, no stale-negative false positives,
  daemon CPU at most 3%, and no sign of the NFS client lookup livelock
  reported in macos-fuse-t/fuse-t#61 (whose root cause turned out to be
  unstable inode numbers in an unrelated FUSE library, not cache tuning).

The full evaluation, including the upstream research and the soak
methodology, is recorded in
`docs/history/packages/fuse/noattrcache-mount-option-evaluation.md`.

## The change

- `cf fuse mount` gains two mutually exclusive flags, plumbed through the
  background supervisor and the compiled binary's hidden subcommands to the
  args handed to `fuse_mount`:
  - `--attrcache-timeout <whole seconds>` selects the cache bound. FUSE-T
    mounts now default to 1; an explicit 0 restores the NFS client's
    untuned caching. Values are validated at the command line and again in
    the daemon, so errors surface where the user typed them rather than
    inside a detached child.
  - `--noattrcache` mounts with FUSE-T's option of that name, kept as a
    diagnostic dial.
- The default is applied in `packages/fuse/mount-options.ts` and gated on a
  new `FusePlatform.provider()`: macFUSE rejects unknown mount options, so
  only FUSE-T mounts receive it, and Linux accepts and ignores both flags.
- A separate first commit teaches the library loader to probe FUSE-T's
  per-user install location (`~/.fuse-t/usr/local/lib`), which the FUSE-T
  installer supports but the loader previously missed.
- Documentation: fuse README section on macOS cache tuning, a
  `fuseNfsCacheTuning` entry in the experimental-flags registry, and a
  staleness note in the fuse-workflow skill.

## What this does not change

The `cf exec` listing-recheck delay (`DIR_LISTING_RECHECK_DELAY_MS`) stays:
it still covers untuned, macFUSE, and pre-1.0.29 FUSE-T mounts (older
FUSE-T predates the `attrcache-timeout` option; the escape hatch there is
`--attrcache-timeout 0`). Mounts on the new default stop hitting the miss
path the delay guards. Reducing the delay toward the one-second bound is a
follow-up tracked in the registry entry once the default has field-soaked.

---
## Summary by cubic
Adds NFS cache tuning to macOS FUSE-T mounts and defaults them to a one-second attribute cache to eliminate stale NotFound/listings without a performance penalty. Also makes the loader find per-user FUSE-T installs.

- New Features
  - `cf fuse mount` adds `--attrcache-timeout <seconds>` and `--noattrcache` (mutually exclusive).
  - Default: FUSE-T mounts apply `attrcache-timeout=1`; `0` keeps the NFS client’s untuned caching. Applied only when the provider is FUSE-T; Linux/macFUSE ignore these flags.
  - Validates `--attrcache-timeout` at the CLI and in the daemon (whole seconds, 0–86400), and rejects a flag present with no value.
  - Mount option construction moved to `packages/fuse/mount-options.ts`; platforms now expose `provider()`; compiled binary and supervisor subcommands use `buildFuseBinaryArgs`.
  - Darwin loader searches `~/.fuse-t/usr/local/lib` for `libfuse-t.dylib` in addition to `/usr/local/lib`.

- Migration
  - No changes required; remounts pick up the default automatically.
  - To keep existing behavior or for FUSE-T versions before 1.0.29, use `--attrcache-timeout 0`. Use `--noattrcache` only for diagnostics.

<sup>Written for commit 26619cf7ad7976dbb60d2ce4f71231dd0ac440f5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4722?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/development/EXPERIMENTAL_OPTIONS.md
+++ b/docs/development/EXPERIMENTAL_OPTIONS.md
@@ -44,6 +44,7 @@ was last checked against the code.
 | [`conflictAdmissionMode`](#conflictadmissionmode) | `CF_CONFLICT_ADMISSION` env, or `setConflictAdmissionMode()` | `off` | William Kelly (#4237) | keep as a tuning dial or remove after re-measurement | implemented, off by default, measured net-negative or neutral |
 | [`syncSchemaTableV2`](#syncschematablev2) | `setSyncSchemaTableConfig()` (negotiated per connection) | on | Ben Follington (#4292) | retire the negotiation once every peer speaks v2 | implemented, on by default |
 | [`cfcRenderCeiling`](#cfcrenderceiling) | `commonfabric.cfcRenderCeiling()` in the browser (localStorage) | off | Bernhard Seefeld (#4550) | graduate once exchange resolution lands | implemented, off by default, dogfood only |
+| [`fuseNfsCacheTuning`](#fusenfscachetuning) | `cf fuse mount --attrcache-timeout <whole seconds; 0 = untuned>` or `--noattrcache` | cf adds `attrcache-timeout=1` (one second) to FUSE-T mounts | Ian Hickson | keep the default; shrink the exec.ts listing-recheck delay once the default has field-soaked | implemented, on by default for FUSE-T, soak-validated |
 
 Removed or never-shipped flags that documentation elsewhere still references are
 recorded in [Appendix A](#appendix-a-removed-and-never-shipped-flags). Toggles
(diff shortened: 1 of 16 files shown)


Hixie:
@bfollington I landed this but let me know if you have any review comments, happy to revert or forward-fix anything you find.

PR #4728, opened 2026-07-15
fix(runner): cap recursive query traversal

ubik2:
## Summary

- add a regression covering a user, direct friends, and second-degree friends connected through schema-bearing links
- verify `asCell: ["opaque"]` preserves the nested friends pointer without loading or tracking its target
- include a non-opaque control proving the nested list and second-degree friend would otherwise be tracked
- document ordinary and opaque `asCell` behavior for runtime and query traversal

## Why

Query traversal normally follows ordinary `asCell` values to register dependencies. Opaque cells are the deliberate hard boundary: inline values are not descended into, and linked targets are not loaded or added to `schemaTracker`. This behavior lets a query retrieve direct friends without recursively retrieving each friend's friends.

## Validation

- `deno test --allow-env --allow-read packages/runner/test/traverse.test.ts`
- `deno fmt --check packages/runner/test/traverse.test.ts docs/specs/space-model/8-traversal.md`
- `deno lint packages/runner/test/traverse.test.ts`
- `git diff --check`

PR 4708 is already merged, so this branch was rebased onto current `main` and contains only this regression test and documentation update.

---
## Summary by cubic
Add regression tests in `packages/runner` for opaque query boundaries with `asCell: ["opaque"]`, confirming nested friends pointers are preserved without loading or tracking targets and the recursive friends graph is limited to one hop. Also fixes traversal to stop at optional `false` links and updates the traversal spec to clarify opaque vs non-opaque behavior, redirect handling, and missing values.

- **Bug Fixes**
  - Do not resolve or track link targets when a property schema is `false` and optional, preventing excluded subtrees from loading during query traversal.

<sup>Written for commit c390b4340311f75fedb9da591b8ad69bc96b9655. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4728?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---BEGIN COPY-PASTE---
NEW_PERF_BASELINE: coverage-debt: packages/runner uncovered lines = 7398 lines
---END COPY-PASTE---

### DIFF
--- a/docs/specs/space-model/8-traversal.md
+++ b/docs/specs/space-model/8-traversal.md
@@ -91,6 +91,11 @@ Top-level `$ref` is resolved before traversal decisions. Defaults are applied fr
 - Arrays validate per item schema
 - Objects validate per property schema
 - Required properties must be present in the filtered result
+- When an explicit property schema rejects a value, an optional property is
+  omitted from the filtered result, while a required property causes the
+  containing object to reject. In particular, an optional `false` property
+  excludes a link without loading or tracking its target; making that property
+  required makes the object unsatisfiable when the property is present.
 
 `additionalProperties` handling is intentionally specialized:
 
--- a/docs/specs/space-model/8-traversal.md
+++ b/docs/specs/space-model/8-traversal.md
@@ -121,8 +121,14 @@ Defaults:
 
 `SchemaObjectTraverser.hasAsCell(schema)` controls boundary behavior.
 
-- With `traverseCells = false` (runtime transforms), `asCell` boundaries produce cell/stream objects without deep traversal of nested content
-- With `traverseCells = true` (query traversal), linked content is traversed to register dependencies
+- With `traverseCells = false` (runtime transforms), ordinary `asCell`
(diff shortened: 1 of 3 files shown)


ubik2:
This example exhibits how to limit the query, despite the link schema in the data indicating that we should recursively load `friends`.

Another approach, touched on in #4708 is to implicitly allow undefined for links when we combineSchema.

That's not necessary, but could be worthwhile to simplify these cases.

ubik2:
Also added an example of the nicer form of this, where you declare your friends' friends as `friends?: never` style.
That exposed a bug that was introduced in #4532, so I made an exception to that behavior for this use case.

PR #4730, opened 2026-07-15
build(toolchain): make mise.toml the canonical Deno version pin

Hixie:
Developers run whatever Deno happens to be installed, because nothing in the repository points them at the pinned version. The version is also encoded in four places kept in sync by hand — mise.toml, the deno-setup CI action, Dockerfile.toolshed, and the tasks/check.sh range — with a comment asking whoever edits one to remember the others.

Make mise.toml the single source of truth:

- .github/actions/deno-setup reads the pin from mise.toml at job time instead of carrying its own hardcoded default, so CI follows the pin with no second place to update. An explicit deno-version input still overrides it. The step requires an exact MAJOR.MINOR.PATCH version: the toolchain cache is keyed on the version and setup-deno installs under whatever a range resolves to, so a range would never hit the cache and every job would contact deno.com again — the failure the cache exists to avoid.
- tasks/check.sh reads the pin and warns, naming mise, when the running Deno differs from it. The accepted range (DENO_VERSION_MIN and DENO_VERSION_MAX) stays the pass/fail gate, so a manually installed Deno keeps working; the range comparison now derives from those variables rather than a separate hardcoded expression that could drift from them.
- tasks/check-deno-pins.ts (deno task check-deno-pins, run as a CI step and covered by the tasks/ unit tests) checks what cannot read mise.toml for itself: every denoland/deno tag in Dockerfile.toolshed must equal the pin, and the check.sh range must contain it.
- README.md, docs/development/DEVELOPMENT.md, and the /deps command describe installing the toolchain through mise (mise trust and mise install), keeping manual installation as a supported fallback.

The checks are built to fail closed. A check that reports "aligned" when it cannot find what it is looking for is worse than no check, because it is trusted. So rather than testing for the string "mise.toml" — which also appears in the action's description and comments, and would still be there after someone replaced the read with a literal — they match the command that performs the read, in both the action and check.sh. The Dockerfile scan tolerates flags before the image name and a lowercase FROM, so a stale stage cannot hide behind either, and compares on the version tag with any digest stripped, so digest pinning is not reported as a mismatch. Both reads stop at the closing quote and take the first match: a pin with a trailing comment would otherwise yield a garbage version that satisfies an emptiness check, and a second deno line would produce a multi-line value that corrupts GITHUB_OUTPUT. Each read checks that mise.toml exists first, because under set -e the read would otherwise abort the script with only sed's own message.

---
## Summary by cubic
Make `mise.toml` the single source of truth for the Deno version. CI, local checks, and Docker now follow this pin with strict validation (exact versions, duplicate-pin rejection) to prevent drift and flaky toolchain installs.

- **New Features**
  - `.github/actions/deno-setup` reads the pinned version from `mise.toml` by default, enforces exact MAJOR.MINOR.PATCH, and keys caches on the resolved version; explicit `deno-version` still overrides. Step names use registered phase markers.
  - `tasks/check.sh` reads the pin, warns when the running Deno differs, errors if `mise.toml` is missing or defines `deno` twice, and compares the accepted `DENO_VERSION_MIN`/`DENO_VERSION_MAX` range via a single numeric helper.
  - Added `tasks/check-deno-pins.ts` (+ tests) and wired `deno task check-deno-pins` into CI to verify `Dockerfile.toolshed` `denoland/deno` tags equal the pin, the range contains it, both `check.sh` and the action still read `mise.toml`, and no action literals disagree with the pin.
  - Docs updated to install Deno via `mise` (`mise trust && mise install`); `/deps`, README, and DEVELOPMENT reflect this.

- **Bug Fixes**
  - Validate range bounds as exact versions and reject non-exact inputs in `compareVersions`; error messages name the offending bound.
  - Reject a `mise.toml` that defines `deno` more than once across the action, `check.sh`, and the checker; all fail closed with clear errors.

<sup>Written for commit b72ca8b28c9732723141e4555da61b0df7c289d2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4730?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.claude/commands/deps.md
+++ b/.claude/commands/deps.md
@@ -1,5 +1,7 @@
 MUST HAVE:
-- deno 2 https://docs.deno.com/runtime/getting_started/installation/
+- deno 2, pinned in mise.toml — install via mise https://mise.jdx.dev/
+  (`mise trust && mise install` in the repo); a manually installed deno 2
+  within the `tasks/check.sh` range also works
 
 REALLY SHOULD HAVE:
 - gh https://github.com/cli/cli
--- a/.github/actions/deno-setup/action.yml
+++ b/.github/actions/deno-setup/action.yml
@@ -25,7 +25,16 @@ runs:
             echo "::error::mise.toml not found; cannot resolve the Deno version. This action requires the repository to be checked out first."
             exit 1
           fi
-          version="$(sed -n 's/^deno = "\([^"]*\)"$/\1/p' mise.toml | head -1)"
+          pins="$(sed -n 's/^deno = "\([^"]*\)"$/\1/p' mise.toml)"
+          # Counted rather than taking the first: TOML rejects a key defined
+          # twice, so a second pin means mise cannot load the file, and reading
+          # past it would install a version no developer actually gets.
+          count="$(printf '%s' "$pins" | grep -c . || true)"
+          if [ "$count" -gt 1 ]; then
+            echo "::error::mise.toml defines the Deno pin $count times; TOML rejects a key defined twice, so mise cannot load it."
+            exit 1
+          fi
+          version="$pins"
         fi
         if [ -z "$version" ]; then
           echo "::error::No deno-version input and no Deno pin found in mise.toml."
--- a/.github/actions/deno-setup/action.yml
+++ b/.github/actions/deno-setup/action.yml
@@ -5,12 +5,41 @@ inputs:
     description: "Load deno dependencies from github cache"
     default: true
   deno-version:
-    description: "Deno version to install"
-    # Keep in sync with `./tasks/check.sh` version
-    default: "2.8.1"
+    description: "Deno version to install. Defaults to the pin in mise.toml."
+    default: ""
 runs:
   using: "composite"
   steps:
+    # The repository's Deno version is pinned once, in mise.toml. This step
+    # reads it so the rest of the action (and any caller that leaves the
+    # deno-version input empty) follows the pin.
+    - name: 📌 Resolve pinned Deno version
+      id: resolve
+      shell: bash
+      env:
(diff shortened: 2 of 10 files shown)


mpsalisbury (changes requested):
Do folks here use mise? The mise.toml was here before me, but I figured I was one of the few who actually use it.
  
  
  - Major — duplicate pins bypass the fail-closed check (tasks/check-deno-pins.ts:37, .github/actions/deno-setup/
    action.yml:28, tasks/check.sh:28) — verified.

    All three readers select only the first matching deno assignment. Given:

    [tools]
    deno = "2.8.1"
    deno = "2.8.1"

    mise rejects the canonical configuration with duplicate key, but findProblems() returns []; the action and check.sh
    also silently select the first value. CI can therefore remain green while the documented mise install workflow is
    broken.

    Require exactly one matching assignment in all three readers and add a duplicate-pin regression test.

Hixie:
The idea of this is to guide people towards using mise. The alternative is to wrap `deno` in a shell script that fetches a pinned deno, but that's much more heavyweight and nobody is doing that today either. We can't both pin deno and do what people are doing today since today people are self-vending deno without a pin.

Hixie:
@mpsalisbury I should add that this only pushes mise for labs. The loom repo will continue to use the self-administered installation path for deno, just using the mise pin in labs. (The PR in the loom repo isn't yet set up to do that, but will be.)

mpsalisbury (approved):
LGTM

PR #4732, opened 2026-07-15
docs: record the identity assumption behind counting active users

Hixie:
Counting active users needs a server-side notion of who is active, and the platform has no accounts. A principal is an Ed25519 keypair named by its DID, and a home space DID is that same value, so home spaces and identities are one-to-one by construction. That makes an identity DID look like a usable stand-in for a person. It is not one, and nothing in the repository said so.

An identity is a keypair, not a person. One human can hold several: the browser derives with Identity.fromMnemonic while cf id derive uses Identity.fromPassphrase, so the same input yields different DIDs, and a second browser profile, a passkey on another device, or a fresh key each mint another identity. One identity can equally be driven by several actors, which is what the recommended shared-key workflow does deliberately and what the publicly derivable "implicit trust" identity does accidentally. Service DIDs and the server identity are principals in the same way a user is.

Add docs/development/active-user-counting.md, recording that assumption as an assumption rather than as a property of the system, together with the signal that already exists to act on it. The session principal is the verified session.open issuer, pinned for the life of the session, and it is already exported as the user.did span attribute on the memory.transact and memory.subscriber.sync spans, so a distinct count per day needs no new instrumentation. The document also records the four properties of that signal that change what the resulting number means: trace sampling does not scale a distinct count the way it scales a volume metric, session.open emits no span of its own, sessions with the "*" principal are omitted, and span retention bounds the lookback.

Keep the identity-model half in HOME_SPACE.md, which already defines the home space DID and the identity DID as the same value, and point it at the new document. Point SHARED_IDENTITY.md there as well, since the workflows it describes are precisely what break the one-identity-one-human assumption.

---
## Summary by cubic
Add `docs/development/active-user-counting.md` documenting that identity DIDs are not people and how to count daily active identities from `user.did` on `memory.transact` and `memory.subscriber.sync` (requires `OTEL_ENABLED`; readers are not counted since `session.open` emits no span).
Link from `HOME_SPACE.md` and `SHARED_IDENTITY.md`, and note in `packages/toolshed/env.ts` that trace sampling (`OTEL_TRACES_SAMPLER`) drops identities and retention bounds lookback.

<sup>Written for commit cf4b71e9791db350fb099f78c703a53d01ec378f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4732?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/common/conventions/HOME_SPACE.md
+++ b/docs/common/conventions/HOME_SPACE.md
@@ -12,6 +12,23 @@ available when they authenticate.
 Home Space DID = User Identity DID = runtime.storageManager.as.did()
 ```
 
+## Identities, Home Spaces, and People
+
+Because the home space DID and the identity DID are the same value, home spaces
+and identities are one-to-one by construction. Every identity has one home
+space, and every home space belongs to one identity.
+
+An identity is a keypair, not a person. There are no accounts, so the system has
+no way to know that two identities belong to the same human, or that one
+identity is driven by several humans or by automation. Treating an identity as a
+person is an assumption a caller layers on top, not a property the system
+provides.
+
+Work that counts people — daily active users, for example — rests on that
+assumption. What the assumption costs, and what the server records about the
+identity behind a session, are covered in
+[`docs/development/active-user-counting.md`](../../development/active-user-counting.md).
+
 ## Purpose
 
 The home space provides a persistent, user-owned storage location for:
--- a/docs/development/SHARED_IDENTITY.md
+++ b/docs/development/SHARED_IDENTITY.md
@@ -5,6 +5,12 @@ when testing identity-sensitive behavior. Scoped cells make this important:
 `PerUser<T>` and `PerSession<T>` data is keyed by the active user DID, not only
 by the space DID.
 
+The workflows here are also the reason an identity is not a reliable stand-in
+for a person: one human routinely holds several identities, and one identity is
+routinely driven by several actors. See
+[`active-user-counting.md`](active-user-counting.md) for what that means when
+counting users.
+
 > **Use a unique key (`id new`) for your own identity.** `cf id derive "implicit
 > trust"` produces a *fixed, publicly-derivable* DID — the one the local toolshed
 > runs as in dev mode — so every developer who derives it shares the exact same
--- a/docs/development/active-user-counting.md
+++ b/docs/development/active-user-counting.md
@@ -1,0 +1,126 @@
+# Counting Active Users
+
+What signal the server offers for counting active people, and the assumption any
+such count rests on.
+
+## Identities Are Not People
+
+The home space DID and the identity DID are the same value, so home spaces and
+identities are one-to-one by construction. That equality is described in
+[`docs/common/conventions/HOME_SPACE.md`](../common/conventions/HOME_SPACE.md).
+
+An identity is a keypair, not a person. There are no accounts, so the system has
+no way to know that two identities belong to the same human, or that one
+identity is driven by several humans or by automation. Counting people therefore
+assumes that one identity DID stands for one human. That assumption is a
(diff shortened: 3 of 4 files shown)


Hixie:
My dashboard agent is trying to build a DAU tile out of this and is unimpressed:

----

## Issue 1 — The server isn't writing spans

`OTEL_ENABLED` is off on estuary. It defaults to off in code, and estuary's settings file has never turned it on.

The proof is a date. Estuary's settings file was last modified on **2026-02-21**. SigNoz didn't exist here until **2026-06-29**. The file predates the thing it would need to point at by four months.

*What it would take:* set that one value to true.

## Issue 2 — The collector doesn't know where SigNoz is

Separate problem, same machine. Even with Issue 1 fixed, estuary's collector would receive spans and forward them nowhere useful, because it's never been told SigNoz's address.

This is where the two playbooks matter. The address lives in a file only the **setup** playbook reads. But production deploys run the **code-shipping** playbook, which doesn't touch the collector at all. So every deploy of new code leaves the collector exactly as it was. No amount of normal deploying will ever fix this.

*What it would take:* run the setup playbook against estuary once.

## Issue 3 — We can't run the playbook

Fixing Issues 1 and 2 both mean running Ansible against estuary, and that needs the vault password. I don't have it and I'm not going to hunt for it. This is an infra operation.

*This one is a hard stop.* Issues 1 and 2 are small and well understood; this is what actually prevents me doing them.

## Issue 4 — There's only one SigNoz, and it's called "stage"

There is no production SigNoz. Turning estuary on means production data flows into a system named for staging. That's not a technical obstacle — it would work fine. It's a question of whether that's acceptable, and it's yours to answer, not mine.

## Issue 5 — Nobody has costed it, and the obvious economy destroys the number

Turning this on means recording **every** span from all 21 production instances. Nobody has measured what that volume costs.

The normal fix is sampling: keep 10% of spans, pay 10%. That works for "how many requests were slow", because 10% of a big number scales back up. **It does not work for counting distinct people.** If a person's handful of spans all land in the discarded 90%, that person vanishes from the count entirely, and no arithmetic afterwards can recover them. You can have it cheap or you can have it correct.

## Issue 6 — Some of those identities are machines

Estuary's own server has an identity. So do background services. They appear in the data exactly like a person does.

They're listed in the same encrypted file as Issue 1, so until someone opens it, we don't know which identities to remove. Any count before then is an over-count by an unknown amount — which matters enormously when the honest answer is a single-digit number.

## Issue 7 — An identity is a keypair, not a person

This is the point your own document makes, and it survives everything above. There are no accounts. An identity is a cryptographic key. One person with a browser and a command-line key is two identities. One key shared across a team is one identity for several people.

So the number is **active identities**. It's a reasonable stand-in for people, but it is an assumption laid on top, not something the system knows. At counts this small, one person with two keys is a 5–50% error.

## Issue 8 — People who only read are invisible

The `user.did` label rides on spans for saving data and syncing. Opening a session writes no span. So someone who logs in, reads, and leaves contributes nothing and simply isn't counted.

I measured this on staging: on 2026-07-12 there were 45 session-establishing events and zero identity-bearing spans. That day would report a confident zero despite real people being present. Roughly one day in ten.

## Issue 9 — The history is 15 days and isn't guaranteed

SigNoz keeps traces for 15 days, then deletes them. Ask for last quarter and there's nothing.

Worse, that 15 days is a live setting on the database, not written down in any repository. It can change without anyone making a commit. So traces are a rolling window, not a record — if you want DAU history, something has to save it as it goes.

PR #4746, opened 2026-07-15
fix(runner): restore strict required link validation

ubik2:
## Summary

- restore ordinary JSON Schema semantics for required link-valued properties inside array elements
- remove the #4532/#4535 element-level grace and its extra target-existence probe
- replace the grace regression tests with coverage showing absent and invalid required targets both reject their containing object and array
- make the convergence fixture's scoped profile link optional where partial visibility is intentional
- update the scoped-cell guidance and replay golden

## Why

The old exception treated an absent link target as meaningfully different from a target that existed but failed its schema. For a required property, both cases mean the property does not match. Silently dropping an absent required property made array-element validation depend on where the object was nested and weakened the declared schema.

Callers that want partial visibility should express that in the schema with an optional, undefined-tolerant, or defaulted field.

## Scoped-cell audit

I audited live pattern declarations using `PerUser` and `PerSession`. The affected shared-record fixture is now optional. Remaining required declarations are outputs, defaulted handler state, or deliberate strict-validation fixtures rather than unintended shared-record hazards.

## Validation

- runner suite: 934 tests passed; the only initial mismatch was the expected stale replay golden
- regenerated replay oracle and reran replay plus required-link tests: 5 tests / 9 steps passed
- convergence integration suite: 4 tests passed
- `deno fmt --check`
- `deno lint`
- `git diff --check`

---
## Summary by cubic
Restores strict JSON Schema validation for required link-valued fields in array elements; absent or invalid targets now reject their element and the containing array. Also isolates session-only topic state from shared lists using `PerSession`, and makes `newTitle` optional.

- **Bug Fixes**
  - Restored strict required validation for link-valued properties in array elements in `packages/runner`; removed the element-level grace and its target-existence probe.
  - Updated unit and convergence tests to assert that absent and invalid required targets void the element and array; controls cover `PerSpace` links and optional fields; refreshed replay golden.
  - In `packages/patterns/topics`, split `TopicPiece` (shared-safe projection) from `TopicOutput` and updated tests to render topic rows and drive session-local controls on a direct `Topic`.

- **Migration**
  - If you relied on partial visibility in array elements, make the field optional, undefined-tolerant, or give it a default. Required scoped targets that a reader cannot access now reject the element and array.
  - Topics list consumers should not expect session-local fields on list items; use a direct `Topic` or `openTopic` for controls like `commentDraft`. `newTitle` is now an optional `PerSession<Writable<string>>`.

<sup>Written for commit c248d9628c1a6f4b6cfea1bedb033ece110141bf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4746?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ubik2:
Still a draft, with two blockers before it's ready.
1. Need to address some tests that weren't patched
2. Window for feedback on whether this is the right approach

PR #4749, opened 2026-07-15
fix(runner): report ref-less module resolution as an error

mathpirate:
## Why

A javascript module that reaches resolution with **no `$implRef`** runs SES-recompiled and CFC-unverified with no trace: the existing `verified-fallback-downgrade` breadcrumb fires only when a ref *is* present. Module-scope references do not exist under that evaluator, so a module that needs them fails at call time with nothing upstream to point at. The matching writer moment — serializing a function implementation with neither provenance nor a verified entry ref — was equally silent.

Nothing in the runtime produces that shape any more. #6071 made `str` a builtin, which was the last producer reachable from a pattern. So this reports the shape at **error** on both sides rather than merely counting it.

## What changed

- **Reader**: `getFallbackJavaScriptImplementation` reports `unverified-source-fallback` when a module arrives with no `$implRef`, with an implementation preview.
- **Writer**: `moduleToEncodableForm` reports `builder.serialize-shape` / `noref-body-write` when a function implementation serializes body-only — the moment the shape is written. It carries its own logger name so write-time diagnostics stay separable from everything the shared `runner` logger reports.
- Both are covered by tests that assert the ledger counts, and the ref-less fallback still executes a self-contained lambda so the reader test drives the real path.
- **Comment fix** in `resolveByImplRef`: the previous text attributed eviction to the artifact index. `addressableByIdentity` is session-lifetime and never evicted; the bounded FIFO covers only the module-namespace reuse cache. The comment now describes the cases where the two indexes genuinely diverge.

## Why error is safe, and what it buys

The pattern-test harness fails a test on error output (`allowConsoleErrors` is the per-test opt-in). Measured on this branch: **all 142 pattern tests pass** at error, and `packages/runner` is **1246 passed / 0 failed**. So the absence of ref-less modules stops being a property that happens to hold and becomes a standing guarantee for patterns — a stronger guard than the log level itself.

Two limits worth stating plainly:

- The guarantee is enforced for **patterns**, not unit tests, which have no console gate. Tests that produce a ref-less module deliberately — `security.test.ts` verifying that untrusted self-contained callbacks run through the SES fallback, and the adversarial suite asserting a dynamic symbol-less artifact never gets a resolvable ref — will now print errors without failing. Those want count-assertions rather than blanket opt-ins; that is follow-up work alongside removing `cf-disable-transform`, which is where the remaining unit-test producers come from.
- One question this instrumentation is partly there to answer: a compiled pattern's compile → run → persist → reload lifecycle serializes *zero* javascript modules through the writer, because compiled patterns persist by pattern reference and source docs, yet body-only module nodes have been observed on the wire in stored graphs. Some path inline-serializes them and it is not one reachable from a test here. If the write-side error ever fires somewhere real, that is the answer.

## Test plan

- `packages/runner` full suite: **1246 passed / 0 failed**
- `deno task integration pattern-tests` (whole suite): **all 142 pattern tests pass**
- repo `deno task check`, `deno fmt --check`, `deno lint` clean

### DIFF
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -3216,11 +3216,15 @@ export class Runner {
         return implementation as (...args: any[]) => any;
       }
     }
-    // Eviction insurance: the artifact index is FIFO-bounded and can roll a
-    // running pattern's module out mid-session, and a post-flip graph has no
-    // legacy ref (and no body when the writer proved resolvability). The
-    // engine's content-addressed implementation index is strong for the
-    // session, so the `$implRef` keeps resolving.
+    // Second-chance resolution through the engine's content-addressed
+    // implementation index (strong and session-unbounded). Note the artifact
+    // index above (`addressableByIdentity`) is ALSO session-lifetime and never
+    // evicted — the bounded FIFO in the pattern manager covers only the
+    // module-NAMESPACE reuse cache (`modulesByIdentity`), not artifact
+    // resolution. This arm covers the cases where the two indexes genuinely
+    // diverge: a module verified-evaluated by the engine without passing
+    // through the PatternManager's registration (e.g. a standalone-Engine
+    // compile), and a post-flip graph that carries no legacy ref and no body.
     return this.runtime.harness.getVerifiedImplementation?.(
       ref.identity,
       ref.symbol,
@@ -4472,6 +4476,25 @@ export class Runner {
         "Verified function resolution missed; running SES-recompiled," +
         " CFC-unverified fallback",
         { $implRef: implRef },
+      ]);
+    } else {
+      // No `$implRef` at all: the writer serialized this module with neither
+      // provenance nor an entry ref, so it re-evaluates bare-SES with NO
+      // breadcrumb anywhere upstream. This branch being fully silent cost the
+      // helper-unlink investigation (topics-dev record, 2026-07) weeks: keep a
+      // ledger-visible counter here even when the logger is disabled (counts
+      // increment before the level/disabled gates).
+      logger.debug("unverified-source-fallback", () => [
+        "Module reached resolution with no $implRef; running SES-recompiled," +
+        " CFC-unverified fallback from stringified source",
+        {
+          preview: typeof module.implementation === "function"
+            ? Function.prototype.toString.call(module.implementation).slice(
+              0,
+              80,
+            )
+            : String(module.implementation).slice(0, 80),
+        },
       ]);
     }
     if (typeof module.implementation === "function") {
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -5128,21 +5128,34 @@ export class Runner {
   }
 
   /**
+   * The module's content-addressed `$implRef` — the defining module's content
+   * identity plus the registered artifact's export/`__cfReg` symbol — when it
+   * is structurally whole. A ref missing either half addresses nothing, so it
+   * reads the same as no ref at all, and every caller treats it that way.
+   */
+  private contentAddressedImplRef(
+    module: Module,
+  ): { identity: string; symbol: string } | undefined {
+    const ref = (module as { $implRef?: { identity: string; symbol: string } })
+      .$implRef;
+    return ref && typeof ref.identity === "string" &&
+        typeof ref.symbol === "string"
+      ? ref
+      : undefined;
+  }
+
+  /**
    * Resolve a module's implementation through its content-addressed
-   * `$implRef` (the defining module's content identity + the registered
-   * artifact's export/`__cfReg` symbol). Returns undefined on a miss (no ref,
-   * never registered, or rolled out of the bounded index) — callers fall back
-   * to the legacy ref or the stringified source.
+   * `$implRef`. Returns undefined when the module carries no usable ref or the
+   * ref names an implementation this runtime never registered; the caller then
+   * falls through to the module's live implementation, and failing that to the
+   * stringified source.
    */
   private resolveByImplRef(
     module: Module,
   ): ((...args: any[]) => any) | undefined {
-    const ref = (module as { $implRef?: { identity: string; symbol: string } })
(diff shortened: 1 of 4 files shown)


packages/runner/src/runner.ts
```diff
@@ -6570,6 +6574,27 @@ export class Runner {
         " CFC-unverified fallback",
         { $implRef: implRef },
       ]);
+    } else {
```
mathpirate, line 6589: [P2] Treat malformed `$implRef` values as the error case too. `resolveByImplRef()` rejects a ref unless both `identity` and `symbol` are strings, but this branch uses only truthiness. I reproduced this with `$implRef: {}`: resolution fell through and executed the SES source fallback, while `unverified-source-fallback.error` stayed unchanged and only `verified-fallback-downgrade.debug` incremented. Sharing the structural validation here would keep every effectively ref-less fallback under the new error-level guarantee.

packages/runner/test/content-addressed-identity.test.ts
```diff
@@ -211,6 +213,63 @@ describe("content-addressed action identity", () => {
     expect(r.key("name").get()).toBe("resolved-after-eviction");
   });
 
+  const ledgerCount = (loggerName: string, messageKey: string): number =>
+    (getLoggerCountsBreakdown()[loggerName]?.[messageKey] as
```
mathpirate, line 220: [P2] Assert the `error` count rather than `total`. The point of the final severity change is that pattern tests fail on error output, but `total` also increments for debug/warn. As a mutation check I changed both new `error()` calls to `debug()` and these assertions still passed. Reading `.error` and checking the exact `before + 1` delta would protect the behavior this PR is establishing.

packages/runner/src/runner.ts
```diff
@@ -5156,11 +5156,15 @@ export class Runner {
         return implementation as (...args: any[]) => any;
       }
     }
-    // Eviction insurance: the artifact index is FIFO-bounded and can roll a
-    // running pattern's module out mid-session, and a post-flip graph has no
-    // legacy ref (and no body when the writer proved resolvability). The
-    // engine's content-addressed implementation index is strong for the
-    // session, so the `$implRef` keeps resolving.
+    // Second-chance resolution through the engine's content-addressed
```
mathpirate, line 5172: [P3] Please update the JSDoc immediately above `resolveByImplRef()` along with this correction. It still says a miss can result from being “rolled out of the bounded index” and that callers fall back to “the legacy ref,” while this new comment says the artifact index is session-lifetime and the caller actually falls through to `liveTrusted` or stringified source. Leaving both descriptions together makes the contract contradictory.

mathpirate (commented):
Three findings from a local review with targeted reproductions and mutation checks.

packages/runner/src/runner.ts
mathpirate: All three confirmed and fixed in 3843c84d1. Thank you — the first one was a real hole in the guarantee this PR exists to make.

**Malformed `$implRef`.** Exactly as you found: resolution validated structurally while the fallback checked only truthiness, so `{}` addressed nothing, missed resolution, and then reported itself as a *downgrade* rather than as ref-less — slipping past the error-level guarantee. Extracted `contentAddressedImplRef()` and shared it between both call sites, so a half-built ref reads as no ref at all. Added a test that drives `$implRef: {}` end to end and asserts it both counts as ref-less and still executes.

**Severity assertions.** Also right, and this is the second vacuous-assertion finding on this stack. `.total` sums every level, so the assertions never guarded the thing the final change is about. They now read `.error` with an exact `before + 1` delta.

**Stale JSDoc.** Rewritten. It still described a bounded index and a fallback to a legacy ref, neither of which is the contract any more.

I re-ran both of your mutations rather than assuming the fixes worked:
- `error()` → `debug()` on both call sites: **5 failures** (previously silent)
- structural check → truthiness: the new malformed-ref test **fails**

Worth naming what your approach caught that mine did not. I had falsified that the reader *fires*, but never that it fires **at error** — so my check passed through the exact mutation you used. Mutation-testing the severity, not just the signal, is the better instrument here.

Gate after the fixes: `packages/runner` 1246 passed / 0 failed, and the whole `pattern-tests` suite passes at error.

packages/runner/test/content-addressed-identity.test.ts
mathpirate: Fixed in 3843c84d1 — you were right, and this is the second vacuous-assertion finding on this stack.

`.total` sums debug, info, warn and error, so the assertions never guarded the severity that the final change is entirely about. They now read `.error` and pin an exact `before + 1` delta rather than a `toBeGreaterThan`.

I re-ran your mutation rather than assuming the fix worked: flipping both `error()` calls to `debug()` now produces **5 failures**, where it previously passed silently.

Worth naming what your approach caught that mine missed. I had falsified that the reader *fires*, but never that it fires **at error** — so my own check passed straight through the exact mutation you used. Mutation-testing the severity rather than just the signal is the better instrument, and I have taken that forward.

packages/runner/src/runner.ts
mathpirate: Fixed in 3843c84d1. Agreed the two descriptions together made the contract contradictory — I corrected the inline comment and left the doc comment above it describing the old behavior.

The JSDoc now reads:

> Resolve a module's implementation through its content-addressed `$implRef`. Returns undefined when the module carries no usable ref or the ref names an implementation this runtime never registered; the caller then falls through to the module's live implementation, and failing that to the stringified source.

So it no longer claims a miss can come from being rolled out of a bounded index (`addressableByIdentity` is session-lifetime and never evicted — the bounded FIFO covers only the module-namespace reuse cache), and no longer points at a legacy ref that does not exist any more.

There is also a new doc comment above `contentAddressedImplRef()`, the helper extracted for your first finding, stating that a ref missing either half addresses nothing and reads the same as no ref at all — which is the invariant the other two call sites now share.

PR #4758, opened 2026-07-16
fix(piece): resolve value-link slot to canonical piece before reading input/result

mathpirate:
## Summary

Fixes the "headless sub-piece" addressing gap: a piece created inside a handler and stored into a list/object — e.g. the topics board's `addTopic` doing `topics.push(Topic({...}))` — was unaddressable by its own fid via the CLI until a renderer opened it. Reads (`cf piece inspect`/`get`) failed with `piece missing argument cell` / `undefined`, and runs (`cf piece step`/`apply`) failed with `Cannot start: no pattern identity`.

## Root cause — an addressing gap, not a materialization gap

The child piece **is** fully set up at create time, synchronously in the handler's transaction: `instantiatePatternNode` mints its canonical result cell **K** (`getCell({resultFor: <slot coords>})`) and runs setup on K, writing `patternIdentity` (`runner.ts:1130`) and the `argument` meta-link (`runner.ts:1102`). The array/object element the list stores — call it **R** — is only a **plain value-link** to K (`childResultCell.getAsLink()`, `runner.ts:5076`, `overwrite:"this"`), carrying no piece metadata.

The board hands out R's fid, and `PieceManager.get` resolved that fid straight to R, so:
- **reads** (`getArgument`/`getResult` off R) found no meta → `piece missing argument cell` / undefined;
- **runs** (`runtime.start(R)`) threw `Cannot start: no pattern identity` — `patternIdentity` lives on K.

The only path that reached K was the renderer's `navigateTo`, which resolves via `resolveAsCell` — hence "unaddressable until a renderer opens it."

(Empirically confirmed create-time, not idle-gated: after a single `dispatchQueuedEvent` with no `synced()` / no scheduler idle, K already carries `patternIdentity` + the argument link; R never does.)

## Fix

Canonicalize the exposed address at the one choke point every get-based caller flows through — **`PieceManager.get`**: after syncing the addressed cell, `resolveAsCell` its **value** link to the canonical result cell K, then start / read / return operate on K. So reads (via `PiecesController.get`), runs (`step` / `apply`), and `stop` all reach the real piece.

- The redirect is a **value** link → resolved via `resolveAsCell` / `resolveLink`, **not** the `result`-meta follower `followCellToResult` (which would not bridge it).
- The addressed cell is `sync()`'d before resolving, since `resolveAsCell` resolves over locally-present links only — this also covers the cold-replica case (see the two-replica test).
- Idempotent for a normal top-level piece (resolves to itself), so non-slot pieces are unaffected. `readingFrom` / `readBy` already canonicalize via `followCellToResult`.

One source file: `packages/piece/src/manager.ts`.

## Tests (`packages/piece/test`)

- `pull-materialization.test.ts` — reads a value-link slot by fid (regression; raw `getArgument(slot)` still throws, but reads via `pieces.get` succeed), plus a **two-replica cold-fetch** (a fresh reader replica reads the slot fid and must both canonicalize R→K and cold-fetch K's docs from the shared server).
- `piece-step-slot.test.ts` — starts a piece via a slot fid (the `cf piece step` path); asserts it reaches the canonical cell and its result. Gates the run-path fix.
- `test.ts` — the `PieceManager.get` mock now implements `resolveAsCell`.

## Verification

- **Live cold read (local toolshed):** deployed the real topics board, headlessly `addTopic`'d, then cold-read the topic by its slot fid from a fresh CLI replica — `inspect` / `get --input title` / `get title` all succeed with the fix, and reproduce `piece missing argument cell` without it.
- `packages/piece` suite (205 steps) + CLI consumer tests (`piece.test.ts`, `piece-cell-operations.test.ts`, `inspect-remote.test.ts`): green. `deno fmt` / `deno lint` / `deno check`: clean.

Context: reframed from "materialization" to "addressing" via the Discord thread with @seefeld and a multi-facet + adversarial analysis — the piece is materialized; the exposed fid was just a wrapper the read/run paths didn't follow. This PR also carries the run-path fix originally split into #4761, consolidated here because `manager.get` is the single canonicalization point that covers both read and run.

### DIFF
--- a/packages/piece/src/manager.ts
+++ b/packages/piece/src/manager.ts
@@ -328,21 +328,23 @@ export class PieceManager {
         scope,
       );
 
-    // Load persisted result/metadata before start() decides whether this is a
-    // resumed piece that needs dependency sync before scheduler wiring — and so
-    // the addressed cell's value link is local for the canonicalization below.
+    // Load the addressed cell. Syncing a value-link "slot" address also loads
+    // its link target — the piece's canonical result cell — together with that
+    // cell's `argument`/`patternIdentity` meta, because the query follows the
+    // top-of-doc value link and returns the target's meta docs. So this one sync
+    // makes both the slot and the canonical cell (with its metadata) local.
     await timePiecePhase("get.piece.sync", () => addressed.sync());
 
-    // Canonicalize a value-link "slot" address to the piece's canonical result
-    // cell. A piece created inside a handler and stored into a list/object (e.g.
-    // the topics board's `addTopic` doing `topics.push(Topic({...}))`) is
-    // addressed by a plain value-link that redirects to the result cell, where
-    // setup wrote `patternIdentity` and the `argument` meta-link. start() needs
-    // that identity and reads need that metadata, so resolving here makes
-    // start / read / stop all operate on the real piece rather than the wrapper.
-    // Idempotent for a normal top-level piece.
+    // Canonicalize the value-link "slot" to the piece's canonical result cell.
+    // A piece created inside a handler and stored into a list/object (e.g. the
+    // topics board's `addTopic` doing `topics.push(Topic({...}))`) is addressed
+    // by a plain value-link that redirects to the result cell, where setup wrote
+    // `patternIdentity` and the `argument` meta-link. start() needs that identity
+    // and reads need that metadata, so resolving here makes start / read / stop
+    // operate on the real piece rather than the wrapper. The sync above already
+    // made the canonical cell local, so this resolves over local links with no
+    // further sync. Idempotent for a normal top-level piece.
     const piece = addressed.resolveAsCell();
-    await timePiecePhase("get.canonical.sync", () => piece.sync());
 
     if (runIt) {
       // start() handles pattern loading and running. It's idempotent - no
--- a/packages/piece/src/manager.ts
+++ b/packages/piece/src/manager.ts
@@ -317,7 +317,7 @@ export class PieceManager {
     scope?: CellScope,
   ): Promise<Cell<T>> {
     // Get the piece cell
-    const piece: Cell<unknown> = isCell(id)
+    const addressed: Cell<unknown> = isCell(id)
       ? id
       : this.runtime.getCellFromEntityId(
         this.space,
@@ -328,19 +328,29 @@ export class PieceManager {
         scope,
       );
 
+    // Load persisted result/metadata before start() decides whether this is a
+    // resumed piece that needs dependency sync before scheduler wiring — and so
+    // the addressed cell's value link is local for the canonicalization below.
+    await timePiecePhase("get.piece.sync", () => addressed.sync());
+
+    // Canonicalize a value-link "slot" address to the piece's canonical result
+    // cell. A piece created inside a handler and stored into a list/object (e.g.
+    // the topics board's `addTopic` doing `topics.push(Topic({...}))`) is
+    // addressed by a plain value-link that redirects to the result cell, where
+    // setup wrote `patternIdentity` and the `argument` meta-link. start() needs
+    // that identity and reads need that metadata, so resolving here makes
+    // start / read / stop all operate on the real piece rather than the wrapper.
+    // Idempotent for a normal top-level piece.
+    const piece = addressed.resolveAsCell();
+    await timePiecePhase("get.canonical.sync", () => piece.sync());
+
     if (runIt) {
-      // Load persisted result/metadata before start() decides whether this is a
-      // resumed piece that needs dependency sync before scheduler wiring.
-      await timePiecePhase("get.piece.sync", () => piece.sync());
       // start() handles pattern loading and running. It's idempotent - no
       // effect if already running.
       await timePiecePhase(
         "get.runtime.start",
         () => this.runtime.start(piece),
       );
-    } else {
-      // Just sync the cell if not running
-      await timePiecePhase("get.piece.sync", () => piece.sync());
     }
 
     // If caller provided a schema, use it
--- a/packages/piece/test/piece-step-slot.test.ts
+++ b/packages/piece/test/piece-step-slot.test.ts
@@ -1,0 +1,90 @@
+import { afterEach, beforeEach, describe, it } from "@std/testing/bdd";
+import { expect } from "@std/expect";
+import { createSession, Identity } from "@commonfabric/identity";
+import { getPatternIdentityRef, Pattern, Runtime } from "@commonfabric/runner";
+import { entityRefToString } from "@commonfabric/data-model/cell-rep";
+import { StorageManager } from "@commonfabric/runner/storage/cache.deno";
+import { PieceManager } from "../src/manager.ts";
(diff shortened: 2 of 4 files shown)


packages/piece/src/ops/piece-controller.ts
```diff
@@ -59,13 +59,26 @@ class PiecePropIo implements PieceCellIo {
...
+    // canonical result cell — where setup wrote `patternIdentity` and the
+    // `argument` meta-link. Reading meta directly off the slot finds nothing
+    // ("piece missing argument cell"); resolve the redirect and fetch the
+    // target so meta lands on the real piece, not the wrapper. The redirect is
+    // a VALUE link, so this resolves via resolveAsCell/resolveLink, NOT the
+    // `result`-meta follower (followCellToResult), which would not bridge it.
+    // Mirrors the link()/getReadingFrom paths, which already canonicalize
+    // before reading a piece's argument.
+    const canonical = this.#cc.getCell().resolveAsCell();
+    await canonical.pull();
```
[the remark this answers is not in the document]
mathpirate: Adopted — `#getTargetCell` now `sync()`s the exposed slot before `resolveAsCell` (which resolves over locally-present links only), then pulls the resolved target.

For the record: the read paths that reach this go through `PiecesController.get`, which already `.sync()`s the cell (`pieces-controller.ts:154`), so it wasn't a live regression through `cf piece inspect`/`get`. But syncing here removes that implicit coupling and makes `#getTargetCell` self-contained. Exercised by the new two-replica cold-fetch test (a fresh reader replica reads the slot fid and must fetch the canonical cell cross-replica).

packages/piece/src/manager.ts
```diff
@@ -328,19 +328,29 @@ export class PieceManager {
...
+    // Canonicalize a value-link "slot" address to the piece's canonical result
+    // cell. A piece created inside a handler and stored into a list/object (e.g.
+    // the topics board's `addTopic` doing `topics.push(Topic({...}))`) is
+    // addressed by a plain value-link that redirects to the result cell, where
+    // setup wrote `patternIdentity` and the `argument` meta-link. start() needs
+    // that identity and reads need that metadata, so resolving here makes
+    // start / read / stop all operate on the real piece rather than the wrapper.
+    // Idempotent for a normal top-level piece.
+    const piece = addressed.resolveAsCell();
+    await timePiecePhase("get.canonical.sync", () => piece.sync());
```
seefeldb: i'm not sure this second .sync is necessary, the first one should have already resolved the cell and then loaded whatever the schema requires.

seefeldb (approved):

packages/piece/src/manager.ts
mathpirate: You're right — dropped it. The first `addressed.sync()` already brings the canonical cell local: syncing the value-link slot makes the query follow its top-of-doc value link and (with `includeMeta`) ship the target's doc + its `argument`/`patternIdentity` meta — `schema:false` only disables the deeper schema traversal, not the top-level link-follow or the meta docs.

Verified on a genuinely cold cross-replica reader calling `manager.get` directly (so `PiecesController.get`'s own outer `.sync()` can't mask it): `argument` + `patternIdentity` are present right after the first sync, before `resolveAsCell` — so resolve just canonicalizes R→K over now-local links. Redundant twice over, really: the controller re-syncs the returned cell, and `doStart` re-syncs the root when `getRaw()` is undefined. Left a short comment where the sync was so the "first sync ships the canonical cell + meta" invariant isn't silent. Thanks for catching it.

mathpirate:
### Scope note (post-merge): what #4758 does and doesn't cover

A deeper look — a multi-facet investigation plus live CLI verification against a real toolshed — sharpened this change's scope. Recording it so the green two-replica test isn't over-read.

**What it fixes (verified cold, end-to-end).** A pushed topic is stored as `board-arg.topics[i]` → value-link → standalone slot cell **R** → value-link → canonical result cell **K** (K carries `patternIdentity` + the argument meta-link + content). Addressing **R** by its own fid: pre-fix `getArgument(R)` throws the reported `piece missing argument cell` (R carries no meta); post-fix `manager.get`'s `resolveAsCell` follows R→K and reads/runs succeed. Confirmed on live current-code data — reverting only the `resolveAsCell` line reproduces the throw; `cf piece inspect <R-fid>` resolves through to K ("cold repro topic") with the fix.

**What it does NOT cover.**
1. **No-op for the canonical fid the app surfaces.** Crossref (`main.tsx:145`) and `navigateTo` both `resolveAsCell()` to **K** before emitting a fid, and `getArgument(K)` already worked pre-fix. So for any fid copied from the UI/URL this change is a strict no-op — it only changes behavior when the *raw slot R* fid is addressed (e.g. an agent reading the raw board-argument array-element link).
2. **Does NOT fix a fresh-replica "no reachable K" failure.** A topic whose canonical cell isn't reachable on a fresh replica (observed on a hosted topic that syncs without a resolvable K) still fails with `piece missing argument cell` — a **sync/persistence** gap (the #4696/#4677 fresh-replica-read frontier), not the addressing gap this PR closes. The two-replica emulated test forces a durable slot via `synced()` barriers, so it does **not** exercise that mode.

**Net:** correct and load-bearing for slot-fid addressing when the slot is present; the fresh-replica/hosted symptom's real fix lives in the sync/materialization layer. Follow-up: add an "unreachable-K / empty-frontier" regression case so the suite distinguishes "slot present → fix helps" from "K unreachable → no-op".

mathpirate:
### Correction to the scope note above

A follow-up investigation — and a parallel session that actually solved the real workflow — corrected me on one point. The "hosted topic that syncs without a resolvable K" in my note above was a **misdiagnosis**: that topic (`leHf…`) is **not** a persistence casualty. It is durably present in the board's topics array (39 entries; it's at index 30), materialized and addressable by its canonical K. Nothing was lost.

What actually happened: the fid I was addressing came from the *board-argument raw* — the old "entry-link fid" workaround — which is a **value-link slot R**. Slot-R `inspect`/`getArgument` doesn't resolve cold **on the estuary specifically** (the divergence in point 1 of the note), while the **canonical K** — the board-*result* array's link target — reads fine cold, and a local current-code toolshed resolves the slot-R too. So there is **no separate sync/persistence gap**; strike that point.

Corrected residual scope:
- **The real papercut was discovery, not addressing** — a headless creator needs to *learn* the canonical K without a renderer. Two routes exist today (the `navigateTo new piece id <fid>` line printed at creation; a schemaless raw read of the board's canonical cell to read its topics' link targets), neither ergonomic, and `cf piece ls` doesn't list headless topics.
- **The estuary slot-R `inspect` divergence** (point 1) is the remaining genuine #4758-scope gap: whether it's estuary deploy-lag or a client-side robustness gap (converge/`pull` rather than depend on the server shipping K's meta on `sync`) is still open.

Net: #4758 + discovery-of-the-canonical-K resolves the headless-addressing workflow; the open items are discovery ergonomics and the estuary slot-R divergence, not a persistence casualty. Verified with a controlled slot/K pair, so the divergence is solid; its cause (deploy-lag vs client) is not yet pinned.

PR #4759, opened 2026-07-16
docs: add Topics Estuary skill

mathpirate:
## Summary

- add a repo-local `topics` skill for the hard-coded Estuary dogfood board
- document the temporary human-shared key policy and require `agentName` atomically in every agent mutation payload
- document the living-body, append-only comment, explicit PR-link, and derived topic-link conventions
- make durable `--input` reads authoritative and use atomic `piece get --step` for current computed results
- treat unresolved result projection and initial synchronization failures as blockers rather than empty Topics state
- wire the skill into `AGENTS.md` and both skill-discovery mirrors

## Dependency

This documentation assumes [#4917](https://github.com/commontoolsinc/labs/pull/4917) lands and the Topics board is updated to that pattern revision. That PR removes the mutable `setMyName` flow, adds Profile-native browser authorship, and makes agent attribution part of each mutation.

[#4874](https://github.com/commontoolsinc/labs/pull/4874), which provides `piece get --step` and loud unresolved-result diagnostics, has merged.

## Operational contract

- agents use the exact same Estuary key as their human user rather than minting an agent key
- every `addTopic`, `setBody`, `addComment`, and `addLink` payload includes the agent's stable `agentName`
- agents do not call `setMyName` or add manual prose/title/label signatures; the pattern stores and renders structured agent attribution while Fabric retains the authenticated human principal
- legacy `myName`, `createdByName`, and `authorName` fields are read-only compatibility data
- source fields such as titles, bodies, comments, links, and the board's topic list are read through `--input`
- computed fields such as `crossrefs`, `topicCount`, `commentCount`, and `lastActivityAt` are read with `get --step`
- canonical topic discovery comes from `crossrefs --step`; `topics --input` is the durable fallback and `piece ls` is not authoritative for handler-created topics

## Validation

- `deno task check-skill-facts`
- `deno task check-docs` (518 code blocks)
- `deno fmt --check AGENTS.md skills/topics/SKILL.md`
- both `.agents/skills/topics` and `.claude/skills/topics` resolve to the skill
- `git diff --check origin/main...HEAD`

The generic skill-creator `quick_validate.py` could not run in this checkout because its Python environment lacks PyYAML; the repository-native skill/docs validators above pass.

### DIFF
--- a/.agents/skills/topics
+++ b/.agents/skills/topics
@@ -1,0 +1,1 @@
+../../skills/topics
--- a/.claude/skills/topics
+++ b/.claude/skills/topics
@@ -1,0 +1,1 @@
+../../skills/topics
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -86,7 +86,7 @@ already does). `docs/development/skill-audit.md` covers what keeps those facts
 honest, including the `deno task check-skill-facts` tripwire that fails CI when
 a path or import a skill cites stops resolving.
 
-For Topics on Estuary, use the repo-local skill at `skills/topics/SKILL.md`.
+For reading or changing Topics on Estuary, use `skills/topics/SKILL.md`.
 
 #### Useful Pattern documentation
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -86,6 +86,8 @@ already does). `docs/development/skill-audit.md` covers what keeps those facts
 honest, including the `deno task check-skill-facts` tripwire that fails CI when
 a path or import a skill cites stops resolving.
 
+For Topics on Estuary, use the repo-local skill at `skills/topics/SKILL.md`.
+
 #### Useful Pattern documentation
 
 **Start here:**
--- a/packages/patterns/topics/README.md
+++ b/packages/patterns/topics/README.md
@@ -68,9 +68,12 @@ cf piece step --piece <topic>                      # refresh computed fields
 ```
 
 Handler source writes commit before `piece step`, so `--input` reads can verify
-bodies, comments, links, and the board's topics list immediately. A step is
-still needed before relying on computed result fields such as `topicCount`,
-`crossrefs`, `commentCount`, or `lastActivityAt`. Fresh CLI replicas converge
-the board's linked topic inputs before recomputing; there is no Topics-specific
-exception to the normal step workflow. Prefer the canonical topic fids exported
-by `crossrefs` over intermediate wrapper links in the board's input array.
+bodies, comments, links, and the board's topics list immediately. A step
+requests recomputation of result fields such as `topicCount`, `crossrefs`,
+`commentCount`, or `lastActivityAt`, but callers must re-read the fields they
+need: a successful step message alone does not prove the result materialized.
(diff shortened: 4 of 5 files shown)


mpsalisbury (commented):
Still going through the review manually. But two of my agents are independently worried about the setMyName dance -- both this one and the one I'm having work on the cf cli. How wed are you to shared identities plus non-atomic writes?

  - Major — authorship metadata is not reliable (skills/topics/SKILL.md:43-52, packages/patterns/topics/README.md:59-74)
    — verified. Agents and humans share one PerUser identity, but setMyName and the mutation are separate CLI calls.
    Another actor can change myName between them, producing comments such as authorName: "Human" with a body signed —
    Fable (agent). Additionally, setBody and addLink never read myName, so setting it before those calls only clobbers
    the human’s persisted display name.

    Concrete fix: make author attribution part of the same atomic handler call, or use distinct identities. Until then,
    describe myName metadata as best-effort, treat the textual signature as authoritative, and avoid setMyName before
    handlers that do not consume it.

mpsalisbury:
Human review: Mostly LGTM. My one concern is about write-safety on body. I think the other fields only support push, which should be safe with concurrent writes. But body is fully-rewritable, and that leaves us open to stomping over the last write (if it arrived after our last read). Do we need some test-and-set semantic for body writes? (either providing the expected body or some receipt from the last body read (hash?)). And am I right that body is the only risky field here?

bfollington (approved):
Solid, and correctly gated on #4917 landing + the board being redeployed to that rev. The authorship contract reads right: same key as the human user, no minting/wildcard/implicit-trust key, atomic `agentName` on every mutation, `crossrefs --step` as the canonical fid index with `topics --input` as durable fallback, and input-vs-`--step` discipline consistent with #4874.

One fragility worth fixing before this becomes load-bearing: it **hard-codes the `fid1:…` board address** in `TOPICS_BOARD_URL`. A redeploy that mints a new board id points every agent following the skill at nothing, silently.

The durable fix is a **slug**. Set one once:

```bash
deno task cf piece set-slug topics <board-id>
# URL becomes https://estuary.saga-castor.ts.net/topics-dev-476ea34f/topics
```

and address the board in the skill by slug (`…/topics-dev-476ea34f/topics`) instead of the raw fid. Slugs are re-assignable, target-agnostic redirects, so any future redeploy only needs `set-slug topics <new-id>` re-applied — the skill text never changes and nothing following it breaks. (Today's `fid1:…` URL is a raw address, not a slug — `isSlugAddress` = no colon — which is why the redeploy fragility exists at all.)

One thing to confirm when wiring this up: that `cf piece … --url` resolves a slug in the path the same way the shell/browser does. The docs spell out slug resolution for shell/browser URLs explicitly; worth a quick check that the CLI's `--url` parser chases it too. If for some reason it doesn't, fall back to at least documenting how to confirm the current board URL, or a re-verify step right after the #4917 redeploy.

mpsalisbury:
How does slug ownership work? specifically, who has write permissions on a slug? same model as piece ownership?

PR #4761, opened 2026-07-16
fix(piece): canonicalize value-link slot in manager.get (cf piece step on a slot fid)

mathpirate:
## Summary

Follow-up to #4758, same "value-link slot" addressing gap but on the **run** path: `cf piece step <slot-fid>` (and `apply`, and any `get(id, runIt=true)`) threw `Cannot start: no pattern identity`. `manager.get` resolved the fid straight to the slot cell **R** and called `runtime.start(R)` — but `patternIdentity` lives on the canonical result cell **K** that R redirects to.

## Fix

`PieceManager.get` now canonicalizes the addressed cell (`resolveAsCell` on the value link) after syncing and before start/return, so `start`, `read`, and `stop` all operate on the real piece K. Idempotent for a normal top-level piece (resolves to itself), so non-slot pieces are unaffected.

This is the run-path counterpart to #4758's read-path canonicalization in `PiecePropIo.#getTargetCell`, and it also covers direct `PiecesController.get` callers. One source file: `packages/piece/src/manager.ts`.

## Tests

- `packages/piece/test/piece-step-slot.test.ts` — creates a value-link slot R→K, then `get(slotId, runIt=true)`; asserts it returns the canonical cell (`patternIdentity` defined) and reads its result. Gates the fix: without it, `runtime.start(R)` throws `Cannot start: no pattern identity`.
- Updated the lightweight `PieceManager.get` mock in `test/test.ts` to implement `resolveAsCell` (the tested flow now canonicalizes before syncing + starting; the sync-before-start invariant it guards is preserved).

## Verification

- `packages/piece` suite (203 steps) + CLI consumer tests (`piece.test.ts`, `piece-cell-operations.test.ts`, `inspect-remote.test.ts`): green
- `deno fmt` / `deno lint` / `deno check`: clean

---
## Summary by cubic
Canonicalizes value-link slot IDs in `PieceManager.get` so start/step/apply run the canonical result cell, fixing "Cannot start: no pattern identity." Ensures run, read, and stop operate on the real piece when addressed through a slot.

- **Bug Fixes**
  - Resolves the addressed cell via `resolveAsCell` after initial sync and before start/return; idempotent for top‑level pieces.
  - Covers `cf piece step`, `apply`, and `get(id, runIt=true)`; also supports direct `PiecesController.get` callers.

<sup>Written for commit 4810b2b9fc9094194d8e1b0b3d016eea4cccf012. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4761?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/piece/src/manager.ts
+++ b/packages/piece/src/manager.ts
@@ -317,7 +317,7 @@ export class PieceManager {
     scope?: CellScope,
   ): Promise<Cell<T>> {
     // Get the piece cell
-    const piece: Cell<unknown> = isCell(id)
+    const addressed: Cell<unknown> = isCell(id)
       ? id
       : this.runtime.getCellFromEntityId(
         this.space,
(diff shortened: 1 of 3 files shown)


mathpirate:
Folded into #4758 — the manager.get canonicalization there is the single choke point covering both the read path (this PR's original target) and the run path (step/apply). The #getTargetCell read-path change is redundant once manager.get resolves the slot, so #4758 now carries the manager.get fix + the step test from here.

PR #4768, opened 2026-07-16
fix(topics): backward-compatible crossref defaults + collapse to one piece-valued derive

mathpirate:
*(Opus, on Gideon's behalf.)*

Two coupled changes on the topics pattern, both flowing from the post-#4698 deploy attempt.

## 1. Deploy-unblock: backward-compatible defaults (the tier-2 finding)

Deploying the merged #4698 onto the existing corpus (`setsrc` the board + topic pieces) **fails the schema backward-compat gate**:

```
Pattern schemas are not backward compatible:
- result.crossrefs: newly required result field has no default
```

`crossrefs` (both the board's `TopicCrossref[]` and each topic's own detail row) and the board's `mentionable` are newly-required result fields with no default, so an existing piece's already-materialized result can't satisfy the new schema. This is exactly the tier-2 "cells-durable, bundles-swappable" question made concrete — and the answer is that it holds for VDOM-only swaps but **adding a required output field is a hard stop** (the gate *blocks* rather than warns, which protects the shared list from blanking). Fix: give the three fields defaults (`Default<[]>` / `Default<{refsOut:[],referencedBy:[]}>`), so a live piece can adopt the new bundle with no migration. Confirmed the defaults register in the emitted result schema.

## 2. The collapse (the tracked #4714 revisit)

We ran the wrapper-bind quarantine down to a decisive test: #4714's path-scoped wildcard fix heals it, and I verified it in the **real** card path — under continuous UI the card-render computed fires (30×) and the navigation binds on the row pieces actually run (a sentinel probe, not just the assertions, which read the derive). So the quarantine premise no longer holds. This collapses the two board derives — `crossrefEdges` (index arrays) + `crossrefs` (piece rows) — into **one piece-valued `crossrefs`**, with the card and chips binding row pieces directly, no index indirection. The detail-page crossrefs become piece-valued too, which moves the index→piece resolution into the covered derive (so it stays at 100%).

## Verification

- Suites: single-runtime 30/30, multi-user 11/11
- Pattern coverage: `main.tsx` 185/185, `topic.tsx` 504/504 — zero uncovered
- Wrapper-bind fires in the real card path (sentinel under `CF_TEST_CONTINUOUS_UI=1`)
- fmt / lint / check clean; changes isolated to `packages/patterns/topics/`

Once this lands, the deploy/backfill experiment (`setsrc` + the one-time `mentionable` link-bind) can run — and it doubles as a before/after on the board's whole-array traverse cost, since dropping `$UI` from the stored `TopicPiece` projection should shrink it.

## Perf baseline — accepted regression (wall-time variance)

Coverage debt is net-0 (the render-step fix covers the card path in both lanes). The one perf-check OVER is a wall-time flake on the multi-user subtest — a file this PR does not touch (the added render steps are in `topics.test.tsx`), and the collapse net-*removes* a board derive, so this is runner-timing variance (stddev 1.1s over n=13), not a real regression. Narrow single-metric accept, not a blanket reset:

```
NEW_PERF_BASELINE: subtest: pattern-unit-5/pattern-unit-tests > packages/patterns/topics/multi-user.test.tsx = 15s
```

### DIFF
--- a/packages/patterns/topics/main.tsx
+++ b/packages/patterns/topics/main.tsx
@@ -67,12 +67,12 @@ export interface TopicsOutput {
   [NAME]: string;
   [UI]: VNode;
   topics: TopicPiece[];
-  mentionable: TopicPiece[];
+  mentionable: TopicPiece[] | Default<[]>;
   topicCount: number;
   /** The prose reference graph over the board's own topics, one row per
    * (non-null) entry of `topics`. Rows carry their topic, so consumers never
    * need to correlate by index — indices are not a stable address. */
-  crossrefs: TopicCrossref[];
+  crossrefs: TopicCrossref[] | Default<[]>;
   /** The viewer's display name (normalized to "" until set). */
   myName: string;
   /** Session-local draft for the footer composer (exposed for embedding and
@@ -124,23 +124,24 @@ export default pattern<TopicsInput, TopicsOutput>(({ topics, myName }) => {
 
   const topicCount = computed(() => asArray(topics.get()).length);
 
-  // The prose reference graph as index edges over `topics` (raw order),
-  // recomputed from the whole corpus on any board change (O(topics × text) —
-  // trivial at board scale; the growth path is per-topic memoization).
-  // Identity is each entry's resolved result-doc fid, so the existing corpus
-  // lights up with zero authoring changes and nothing derived is persisted.
-  // Indices, not pieces: the UI must bind its navigation handlers to direct
-  // `topics` elements — a handler bound to a piece nested inside a derived
-  // wrapper object silently keeps the consuming UI computed from ever
-  // running.
-  const crossrefEdges = computed(() => {
+  // The prose reference graph as one piece-valued view over the board's own
+  // topics (one row per non-null entry), recomputed from the whole corpus on
+  // any board change (O(topics × text) — trivial at board scale; the growth
+  // path is per-topic memoization). Identity is each entry's resolved
+  // result-doc fid, so the existing corpus lights up with zero authoring
+  // changes and nothing derived is persisted. Rows carry their pieces
+  // directly: the UI binds navigation handlers to the row pieces (safe since
+  // #4714's path-scoped wildcard fix — before it, a handler bound to a piece
+  // nested inside a derived wrapper object silently kept the consuming UI
+  // computed from ever running, which forced the earlier index-plumbed shape).
+  const crossrefs = computed(() => {
     const list = asArray(topics.get());
-    // Each entry's own fid payload ("" while unresolved, e.g. mid-sync —
-    // such entries simply hold no edges this render).
+    // Each entry's own fid payload ("" while unresolved, e.g. mid-sync — such
+    // entries simply hold no edges this render). resolveAsCell/entityId are
+    // cell-runtime surface, not on the pattern Writable type (same cast as
+    // notes' appendLink).
     const payloads = list.map((t, i) => {
       if (!t) return "";
-      // resolveAsCell/entityId are cell-runtime surface, not on the pattern
-      // Writable type (same cast as notes' appendLink).
       const ref = (topics.key(i) as any).resolveAsCell?.()?.entityId;
       return ref ? fidPayload(entityRefToString(ref)) : "";
     });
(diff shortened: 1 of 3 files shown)


bfollington:
*(Posted by Claude on Ben's behalf, after reviewing the diff together.)*

Solid PR — both halves look right to us, and the timing on the collapse is ideal: since the #4698 deploy never landed, the index-plumbed shape never reaches a live piece at all. The tier-2 finding is also better news than we'd hoped: we'd asked for the framework to *warn* on incompatible deploys, and for this class it turns out to hard-block. Nice.

Three questions, purely to check understanding — you may well have covered all of these already, in which case a one-liner each is plenty. None of them block merge from Ben's side.

1. **Existing-data reachability.** With the three defaults in place, is the whole existing corpus definitely still reachable after the setsrc — i.e. is there anything in an old piece's already-materialized result that the new schema would still refuse, or does `Default<>` fully cover the gap? Asking because this deploy is the first real exercise of the schema-addition path, so "old data stays accessible" is the property we most want to be sure of before it runs.

2. **Development-time detection of the silent failure mode.** The card list now rides the wrapper-bind mechanism, whose failure mode is silent non-render (no error, board just blank). Your sentinel verified it fires today — is there a way to commit that guard so a *future* regression of the mechanism fails loudly in CI rather than shipping a blank board? (Possibly as simple as a render-step asserting a known card title appears in the board's vdom — or maybe the continuous-UI demand already gives you this and we're missing it.) More generally: is this class of failure knowable at development time, or is that itself a platform gap worth filing?

3. **The two coverage instruments disagree.** The PR body has pattern coverage at 185/185 zero-uncovered for main.tsx, while the V8 coverage-debt gate reports +39 uncovered lines in the same file. Do you know why? If the 39 lines are the new card path in one lane, that connects to question 2; if it's benign instrument skew, the targeted `NEW_PERF_BASELINE: coverage-debt:` override with that rationale in the body seems right.

mathpirate:
*(Opus, on Gideon's behalf.)*

Great questions — and #2/#3 caught the exact thing the coverage gate then caught too. Answers, and a fix pushed for the one that needed it:

**1. Existing-data reachability.** `Default<>` fully covers the gap. Each new required *result* field (board `crossrefs`/`mentionable`, topic-detail `crossrefs`) is derived, so an existing piece's stored *source* (title/body/comments/links) is untouched; the schema just reads the default until the derive recomputes. Nothing is written back into stored data, so there's no migration and old data stays reachable. That said — this is the first real run of the schema-addition path, so I'm treating the deploy as the proof, not the assertion: I'll `setsrc` + backfill and report before/after that the whole corpus is still reachable and lights up.

**2. Dev-time detection of the silent non-render.** Done, as you suggested. The card list now rides explicit `{ render: board[UI] }` steps in `topics.test.tsx`. They drive the wrapper-bind card render through the real reconciler, so a silent non-render regression (blank board, no error) leaves those ~40 card lines uncovered → the coverage-debt gate trips → **CI fails loudly** rather than shipping a blank board. So the failure mode is now a committed CI guard. On the broader "is this class knowable at dev time" — the durable version is the diagnosability ask on the bind-no-fire topic: a node that's deferred/silently-non-firing should *name itself* in a log. That's the general platform fix worth filing; the render-step guard is the per-pattern version of it.

**3. Why the two coverage lanes disagreed.** It's the card path (so it connects to #2), not benign skew — and you're right to have flagged it. The lanes demand the UI differently: the transformer pattern-coverage lane runs under continuous UI demand, so the wrapper-bind card render fires (185/185); the V8 coverage-debt lane runs the suite normally, where the passive `[UI]: board[UI]` export under-demands the wrapper-*nested* bind chain — it needs real reconcile cycles to settle, which a single implicit render doesn't provide — so the card render didn't fire → your +39. The `{ render: board[UI] }` steps from #2 fix *both* lanes: the card render now covers in V8 too (185/185, no flag), so **no `NEW_PERF_BASELINE` override is needed**. It also de-risks the production question — the render fires under real reconcile cycles, which the live board provides continuously.

Net: pushed the render-step guard, rebased on main, and CI should come back green with honest coverage. Thanks for the sharp read.

bfollington:
*(Posted by Claude on Ben's behalf — re-review of the updated PR.)*

Re-review done: the delta is exactly as described (pattern sources unchanged from what we reviewed; the test file gains the three `{ render: board[UI] }` steps and the explanatory comment), all three answers check out against the code, and CI is green. **Good to merge from Ben's side.**

Particularly good: the coverage-lane root-cause confirming the card path connection, and treating the deploy as the proof rather than the assertion on reachability. One small observation for the record, not an ask: the blank-board guard's trip-wire is the coverage gate in Performance Check, which is a non-required check — so a future wrapper-bind regression shows as a loud red X rather than a merge block. Fine for now; the durable answer is the self-naming-node diagnosability ask you've filed on the bind-no-fire topic.

Looking forward to the deploy/backfill report — that's the last open loop from the #4698 review, and it clears the way for the notes pass on Ben's side.

PR #4769, opened 2026-07-16
refactor: move `entity-kind` from `data-model` to `runner`

danfuzz:
## Why

Follow-on to #4765, which moved `entity-kind` out of `fabric-primitives` — correct, since it is not a fabric primitive. This moves it the rest of the way: out of `data-model`.

`data-model` never used it. Every consumer was external, and the only in-package references were its own export entry. The module describes the URI scheme that *wraps* a tagged hash — by its own account the part living OUTSIDE the tagged-hash string — while `data-model` owns the hash. The one thing that ever tied the two together was the `fid2:computed:<hash>` tag format, retired in #4659 before it shipped.

## Why `runner`

- It is the only package that **mints** (`uriSchemeForEntityKind`); everyone else only reads.
- Its collaborator is here. `toURI` consumes `uriSchemeForEntityKind`; `fromURI` consumes `entityUriSchemePrefix`. Those two halves were split across a package boundary that neither side could see over — `uri-utils.ts` now sits beside it.
- **No cycle.** `data-model` has no need of it, so no `data-model` → `runner` edge appears.
- Every consumer already depends on `runner`: shell (5 files), ui (16).
- No layer-1 sibling needs it — `api`, `memory`, and `identity` are all clean. `memory` stores these ids but treats them as *opaque* by policy ("Derive a safe attach alias from an opaque db id"), so it is not a latent consumer.

Exported as `@commonfabric/runner/entity-kind`, mirroring the existing `./slugs` precedent: a narrow subpath, so `shell` and `ui` — which need three pure functions between them — keep importing a leaf instead of the runtime. Consumers' import lines change in package name only.

## What moved

| | |
|---|---|
| `data-model/src/entity-kind.ts` | → `runner/src/entity-kind.ts` (100% rename) |
| `data-model/test/entity-kind.test.ts` | → `runner/test/entity-kind.test.ts` (95%, import paths) |
| export `./entity-kind` | `data-model` → `runner` |

Consumers re-pointed: 8 in `runner` (now relative), 4 in `shell`, 1 in `ui`.

## Not done here

`EntityKind` is **not** renamed, per consensus — though it does collide with an unrelated `EntityKind` in `state-inspector/model.ts` (a document classification: `"piece" | "module" | "stream" | …`). Worth noting the module's "Terminology: `kind` vs `scheme` (and the other two `kind`s)" section disambiguates Stream `$kind` and `asCell` kinds, but not the one sharing its exact identifier.

## Testing

No behavior change. `data-model` 47 passed / 0 failed — back to its exact pre-#4659 count. `runner`, `shell`, `ui` all green. `deno fmt --check` clean across all 17 files.

Whole workspace green except `deno-web-test`, which fails **identically on unmodified `main`** (`smoke test`, 22 passed / 1 failed) — verified by stashing this branch and re-running. Pre-existing, unrelated.

https://claude.ai/code/session_01WxmQgELnDx4NJQWPqTVMZe


ubik2 (approved):
lgtm

PR #4771, opened 2026-07-16
fix(cli): make cf output script- and automation-friendly

willkelly:
## Summary

The `cf` CLI is awkward to drive from scripts and automation. Several rough
edges force every programmatic caller to add workarounds:

- ANSI color codes are written even when output is piped, so consumers have to
  strip escape sequences before parsing.
- `piece call` can only take its JSON payload as an argv argument — there is no
  way to pipe it in.
- A `piece get` whose path doesn't resolve prints a full (colorized) usage/help
  screen to **stdout**, so `val=$(cf piece get … path)` captures kilobytes of
  help text instead of failing cleanly.
- Invocation is documented only as `deno run -A packages/cli/mod.ts …`, which
  breaks when the shell's working directory isn't the repo root.
- Deno's own npm "Ignored build scripts" banner prints on every `deno task cf`
  invocation.

This PR makes the CLI behave like a well-mannered Unix tool: colorize only on a
TTY, keep stdout machine-parseable, accept stdin, and fail as a data error
rather than a usage error.

## Changes

**Color is TTY-gated.** ANSI is emitted only when stdout is a terminal.
`--no-color` / `NO_COLOR` force it off; `FORCE_COLOR` / `CLICOLOR_FORCE` force
it on (e.g. through a pager). This also covers Cliffy's help/version/usage
output, which colorized unconditionally before. Cliffy's help generator
overrides the global color flag while it renders, so the resolved policy is
mirrored into the root command's help options, and `@std/fmt/colors` is pinned
to Cliffy's own dependency range so the color toggle reaches the module
instance Cliffy uses (a test guards the pin).

**`piece call` reads stdin.** A `-` payload is a stdin sentinel —
`echo '{…}' | cf piece call … <handler> -` — reusing the existing
`--json-file -` path so empty stdin fails loudly. `--json -` works the same
way. Previously `-` produced a confusing `Invalid JSON for --json` and never
read stdin.

**`piece get` on an unresolved path is a data error.** It now prints a one-line
message to stderr and exits 1, instead of throwing a Cliffy `ValidationError`
that dumped the usage screen to stdout. When the path exists only on the input
cell, it hints to retry with `--input`.

**Built binary documented as the primary entry point.**
`deno task build-binaries --cli-only` compiles `dist/cf`, which is
cwd-independent and free of Deno startup noise. `skills/cf/SKILL.md` and
`packages/cli/README.md` now recommend it (with a rebuild-after-pull note) and
describe the output conventions.

**No more Deno banner.** The launcher spawns the child CLI with
`deno run --quiet`, suppressing the npm "Ignored build scripts" banner on the
`deno task cf` path.

## Before / after

Identical commands, output piped (non-TTY), on `main` vs this branch:

| Command | Before | After |
|---|---|---|
| `val=$(cf piece get -q … missingField)` | 3171 bytes of colorized usage screen captured into `$val` (stdout); exit 1 | 0 bytes on stdout; one `Cannot access path …` line on stderr; exit 1 |
| `cf piece get --help \| cat` | 22 ANSI escape sequences | 0 (and 22 again with `FORCE_COLOR=1`) |
| `echo '{…}' \| cf piece call … <handler> -` | `Invalid JSON for --json`; stdin ignored | payload read from stdin |

## Compatibility constraints (intentionally unchanged)

Two adjacent behaviors look like they should change but are load-bearing for
[`loom`](https://github.com/commontoolsinc/loom), which vendors this CLI under
`vendor/labs` and parses its output. They are left as-is on purpose:

- **`-q` / `--quiet` does not change the log level.** It suppresses only the
  hint / next-step blocks. loom greps the CLI's stderr for the runtime warning
  `load-pattern-by-identity-source-miss` to trigger an automatic root-pattern
  repair; quieting warnings under `-q` would silently disable that. Use
  `--log-level error` to drop warnings.
- **`piece get` renders an empty result as literal `undefined`, not JSON
  `null`.** A fresh-space deploy reads back `undefined` during a normal window,
  and loom's deploy gate treats that as a tolerated transient; emitting `null`
  would turn it into a hard failure.

## Tests

New unit coverage keeps the change coverage-neutral (no net coverage debt in
`packages/cli`): `test/color-mode.test.ts` (TTY/env precedence, the `--`
payload boundary, `safeEnvGet`, the `@std/fmt` pin guard); `--json -` /
`--json-file -` stdin sentinels for object and non-object schemas in
`test/exec.test.ts`; stdin-payload and `pieceGetDataErrorReport` cases in
`test/piece-call.test.ts`; and a launcher `--quiet` assertion. The stdin,
color-policy, and data-error logic is factored into pure, unit-tested helpers.
The full `packages/cli` suite passes.

## Coverage baseline

```
NEW_PERF_BASELINE: coverage-debt: packages/cli uncovered lines = 4006 lines
```

Accepts a narrow +1 uncovered line in `packages/cli`. It comes from routing
main's new `PieceResultProjectionError` (#4874) through this PR's clean
data-error path instead of the Cliffy usage-screen wrapper it shipped with —
the pure logic is unit-tested; the residual line is the integration-only
`piece get` action wiring (a `.action()` callback that calls `Deno.exit`, not
reachable from a unit test and not exercised by an existing integration test).

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -26,6 +26,39 @@ as `cf piece ls`. Human-readable output uses the same columns as `piece ls`.
 matches. If part of a piece cannot be read, the command reports a warning on
 standard error and continues searching that piece and the rest of the space.
 
+## Output Conventions
+
+- stdout carries command output only; hints and diagnostics go to stderr.
+  `piece get` prints plain JSON, except an absent value prints the literal
+  `undefined` — which is not valid JSON, so test for it before parsing. It is
+  `undefined` rather than `null` on purpose: loom's fresh-space readback gate
+  tolerates `undefined` as a transient, and `null` would hard-fail its deploy
+  gate (see the `-q` note below and the PR compatibility rationale).
+- ANSI colors are emitted only when stdout is a TTY. `--no-color` or
+  `NO_COLOR=1` disables them everywhere (including Cliffy help/usage output);
+  `FORCE_COLOR=1`/`CLICOLOR_FORCE=1` forces them when piped. The policy is
+  applied in `lib/color-mode.ts`; the `@std/fmt/colors` import-map pin in
+  `deno.jsonc` must track Cliffy's own `@std/fmt` dependency range (guarded by
+  `test/color-mode.test.ts`).
+- `-q/--quiet` (on `piece`/`wish` subcommands) suppresses the stderr hint and
+  next-step blocks. It deliberately does NOT change the log floor: consumers
+  parse `--quiet` runs' stderr for runtime warnings (Loom's stale-root heal
+  greps for `load-pattern-by-identity-source-miss`). Use `--log-level error` to
+  drop warnings; the two compose.
+- `piece call` accepts its payload as an inline JSON argument, `-` for stdin, an
+  implicit pipe (no payload argument), or schema-derived flags after `--`.
+- A `piece get` path that doesn't resolve prints a one-line error on stderr and
+  exits 1 — it is a data error, not a usage error.
+- The launcher spawns the child CLI with `deno run --quiet` so Deno's own
+  warnings (npm "Ignored build scripts" banner) never reach users.
+
+## Built Binary
+
+`deno task build-binaries --cli-only` compiles the CLI to `dist/cf` — fully
+cwd-independent with no Deno startup noise, the recommended entry point for
+agents and scripts. Rebuild after every `git pull`: a stale binary rejects newer
+flags and can hit wire-protocol skew against an updated server.
+
 ## Launcher Contract
 
 `packages/cli/launcher.ts` is the stable Deno launcher for consumers that need
--- a/packages/cli/commands/piece.ts
+++ b/packages/cli/commands/piece.ts
@@ -151,8 +151,58 @@ export function localPatternEntry(
   };
 }
 
-function pieceCallRawArgs(tail: string[], literalArgs: string[]): string[] {
+/**
+ * A `piece get` failure caused by a data condition rather than bad arguments:
+ * a path that doesn't resolve, or a result schema that can't project the
+ * stored data (PieceResultProjectionError). Reported as a plain error on
+ * stderr with exit 1, never as a Cliffy ValidationError (which would dump the
+ * usage screen and read as an arg-parse failure).
+ */
+export function isPieceGetDataError(error: unknown): error is Error {
+  return error instanceof PieceResultProjectionError ||
+    (error instanceof Error &&
+      error.message.startsWith("Cannot access path"));
+}
+
+/**
+ * Build the stderr report for a `piece get` failure. Returns null when the
+ * error is not a data error (the caller should rethrow). `message` is the
+ * one-line error; `hint` is an optional next-step tip. A projection error
+ * already carries its own `--step` guidance, and an input-mode read has
+ * nothing more to suggest — only a result-mode unresolved path gets the
+ * `--input` tip.
+ */
+export function pieceGetDataErrorReport(
+  error: unknown,
+  opts: { input?: boolean; piece?: string },
+): { message: string; hint?: string } | null {
+  if (!isPieceGetDataError(error)) return null;
+  if (error instanceof PieceResultProjectionError || opts.input) {
+    return { message: error.message };
+  }
+  return {
+    message: error.message,
+    hint: cliText(
+      `TIP: The path was read from the result cell. If the field is an input, retry with --input, or run 'cf piece inspect --piece ${opts.piece} ...' to see both cells.`,
+    ),
+  };
+}
+
+export function pieceCallRawArgs(
+  tail: string[],
+  literalArgs: string[],
+): string[] {
   if (literalArgs.length > 0) {
+    // Schema-derived flags after `--`. A payload token before `--` (inline
+    // JSON or the `-` stdin sentinel) would be silently dropped here, so
+    // reject the combination loudly instead — the same no-op this family of
+    // fixes is stamping out. Mirrors the `tail.length > 1` rejection below.
+    if (tail.length > 0) {
+      throw new ValidationError(
+        'Pass either a payload argument (inline JSON or "-" for stdin) or ' +
+          'schema-derived flags after "--", not both.',
+      );
+    }
     return literalArgs;
   }
 
@@ -172,6 +222,17 @@ function pieceCallRawArgs(tail: string[], literalArgs: string[]): string[] {
     );
   }
 
+  // Explicit two-token stdin sentinels (a JSON/value flag plus "-"), forwarded
+  // to the exec layer so the friendly surface matches `cf exec` and the bare
+  // "-" form. Without this they'd hit the multi-argument rejection below.
+  if (
+    tail.length === 2 && tail[1] === "-" &&
+    (tail[0] === "--json" || tail[0] === "--json-file" ||
+      tail[0] === "--value-file")
+  ) {
+    return [tail[0], "-"];
+  }
+
   if (tail[0] === "--json") {
     if (tail.length === 1) {
       // --json alone is a no-op: cf piece call always outputs JSON.
@@ -186,6 +247,12 @@ function pieceCallRawArgs(tail: string[], literalArgs: string[]): string[] {
     throw new ValidationError(
       'Use a single inline JSON argument or "--" before schema-derived flags.',
     );
+  }
+
+  // "-" is the conventional stdin sentinel; route it through the existing
+  // --json-file stdin path so empty stdin still fails loudly.
+  if (tail[0] === "-") {
+    return ["--json-file", "-"];
   }
 
   return ["--json", tail[0]];
@@ -814,6 +881,7 @@ PATH FORMAT: Use forward slashes and numeric indices for arrays.
   )
   .arguments("[path:string]")
   .action(async (options, pathString) => {
+    setQuietMode(!!options.quiet);
     const pieceConfig = parsePieceOptions(options);
     const pathSegments = pathString ? parseCellPath(pathString) : [];
     try {
@@ -823,11 +891,18 @@ PATH FORMAT: Use forward slashes and numeric indices for arrays.
       });
       render(value, { json: true });
     } catch (error) {
-      if (
-        error instanceof PieceResultProjectionError ||
-        error instanceof Error && error.message.startsWith("Cannot access path")
-      ) {
-        throw new ValidationError(error.message, { exitCode: 1 });
+      // A read that fails on a data condition — the path doesn't resolve, or
+      // the result schema can't project the stored data (PieceResultProjection
+      // Error) — is a data error, not a usage error. Report it on stderr
+      // instead of letting Cliffy dump the help screen over it.
+      const report = pieceGetDataErrorReport(error, {
+        input: options.input,
+        piece: pieceConfig.piece,
+      });
+      if (report) {
+        console.error(report.message);
+        if (report.hint) hint(report.hint);
+        Deno.exit(1);
       }
       throw error;
     }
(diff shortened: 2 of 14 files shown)


skills/cf/SKILL.md
```diff
@@ -18,6 +18,48 @@ deno task cf piece --help     # Piece operations
...
+   `deno -q task cf ...`. stdout stays clean, so redirection is safe.
+
+3. **`deno run -q -A packages/cli/mod.ts ...`** — repo-root-relative; only works
+   with the repo root as cwd. The `-q` suppresses Deno's own warnings (e.g. the
+   npm "Ignored build scripts" banner).
+
+## Output Conventions (scripts & agents)
+
+- stdout carries command output only; hints, tips and diagnostics go to stderr.
+  `piece get` always prints plain JSON (`null` when the cell holds no value) —
```
mathpirate: 🔴 **Blocking (docs): this sentence teaches the behavior the branch deliberately reverted.**

`piece get` prints literal `undefined` for an absent value — the PR body's compatibility section explains why it must (loom's fresh-space readback gate tolerates `undefined` on purpose) — and `JSON.parse("undefined")` throws. So "parse it directly" sends a scripted consumer into exactly the retry-loop this PR set out to kill, with the CLI's own skill as the misleading source.

Verified live against estuary with this branch: `cf piece get …` on a piece whose result cell is empty → stdout `undefined`, exit 0.

Suggested wording:

```
`piece get` prints plain JSON, except an absent value prints literal
`undefined` (not valid JSON — test for it before parsing; see the
loom-compatibility note in packages/cli/README.md for why it isn't `null`).
```

Same sentence appears in `packages/cli/README.md` line 6 (flagged there too). If the README bullet gains the why, this one can just point at it.

packages/cli/README.md
```diff
@@ -1,5 +1,34 @@
 # @commonfabric/cli
 
+## Output Conventions
+
+- stdout carries command output only; hints and diagnostics go to stderr.
+  `piece get` prints plain JSON (`null` when the cell holds no value).
```
mathpirate: 🔴 Twin of the `skills/cf/SKILL.md` finding: actual behavior is literal `undefined` for an absent value (intentional — it's this PR's own second compatibility constraint), so "`null` when the cell holds no value" documents the reverted behavior. Since this bullet sits next to the compatibility rationale already, it's the natural home for the one-line why (`undefined` is loom's tolerated fresh-space transient; `null` would hard-fail its deploy gate).

packages/cli/lib/color-mode.ts
```diff
@@ -0,0 +1,92 @@
...
+ * When colors are disabled, `NO_COLOR=1` is also exported so spawned child
+ * processes (workers, `cf check` pipelines) inherit the decision.
+ */
+export function applyColorMode(
+  args: string[],
+): { args: string[]; enabled: boolean } {
+  const { noColor, args: cleanArgs } = extractNoColor(args);
+  const enabled = resolveColorEnabled({
+    noColorFlag: noColor,
+    denoNoColor: Deno.noColor,
```
mathpirate, line 89: 🟡 **`Deno.noColor` is not the raw `NO_COLOR` env var — Deno pre-arbitrates FORCE_COLOR over it**, and that quietly inverts the precedence this module documents and tests:

```console
$ NO_COLOR=1 deno eval 'console.log(Deno.noColor)'                # true
$ NO_COLOR=1 FORCE_COLOR=1 deno eval 'console.log(Deno.noColor)'  # false ← Deno already resolved it
$ NO_COLOR=1 FORCE_COLOR=1    cf piece get --help | grep -c $'\x1b'  # 22 → FORCE wins live
$ NO_COLOR=1 CLICOLOR_FORCE=1 cf piece get --help | grep -c $'\x1b'  # 0  → NO_COLOR wins live
```

So the two force vars rank inconsistently against NO_COLOR (Deno knows FORCE_COLOR, not CLICOLOR_FORCE), the README's "NO_COLOR=1 disables them everywhere" fails for one combination, and the `disable overrides win over everything` unit test passes only because it feeds an input state — `denoNoColor: true` alongside `FORCE_COLOR=1` — that the runtime can't actually produce for FORCE_COLOR.

One-line fix that keeps your stated policy and makes both force vars rank the same:

```ts
denoNoColor: Deno.noColor || (safeEnvGet("NO_COLOR") ?? "") !== "",
```

plus a test asserting at the env level rather than the pre-arbitrated flag. (Adopting Node's FORCE-wins convention instead is also coherent — but then the test name, the README sentence, and CLICOLOR_FORCE's ranking should all move together.)

Verified live; edge-case severity (both vars set), flagging because the tested model and the live behavior disagree.

mathpirate (commented):
_cf-review via Claude (Fable 5, `claude-fable-5`), on behalf of @mathpirate._

**Scope:** the 14-file diff at 869e1f456 (+621/−14) · **Coverage:** deep-read, all files · **Checks run:** full `packages/cli` suite locally (23 passed / 0 failed), live verification against estuary using this branch's CLI, lockfile pin inspected, CI observed green.

**Verdict up front: approve-worthy once the one 🔴 docs fix lands.** The 🟡 is a one-liner worth taking in the same pass. This is careful work — the compatibility-constraints section (auditing loom's actual consumption of CLI output before changing it) is exemplary, and the `@std/fmt` pin guard is a real instance-identity test, not a decoration.

**Independently verified on this branch, live:**

- Piped `--help` emits 0 ANSI escapes; `FORCE_COLOR=1` restores them (22); `deno task cf` works from a repo subdirectory; stdout stays clean through the launcher path.
- Bogus-path `piece get` against a real estuary piece: 0 bytes on stdout, one-line error + TIP on stderr, exit 1; `-q` suppresses the TIP and keeps the error.
- Empty-value `piece get` prints literal `undefined` — the intentionally-preserved loom behavior, which is what the 🔴 finding is about.
- `deno.lock`: Cliffy's `@std/fmt@~1.0.2` range and the new import-map pin both resolve to 1.0.8 — one module instance, as the deno.jsonc comment claims.

**Findings** (detail on the inline comments):

- 🔴 **Blocking · docs** — `skills/cf/SKILL.md` and `packages/cli/README.md` both teach "`piece get` prints plain JSON (`null` when the cell holds no value) — parse it directly", but the shipped behavior (correctly, per the PR body's compatibility section) prints literal `undefined`, which `JSON.parse` rejects. Residue of the reverted `null` change. Verified.
- 🟡 **Improvement · color precedence** — `NO_COLOR=1 FORCE_COLOR=1` produces color: `Deno.noColor` is pre-arbitrated by Deno (it's `false` when FORCE_COLOR is set), so the resolver's documented disable-wins precedence inverts for FORCE_COLOR while still holding for CLICOLOR_FORCE. One-line fix. Verified.
- ⚪ **Nit** — `piece call <piece> <handler> - -- --flag x` silently ignores the `-` and never reads stdin (post-`--` args win). Same silent-no-op class this PR targets, on new surface.

**Possible follow-ups** — we're picking these two up in a stacked follow-up PR, so no action needed here; listing for the record:

- `extractLogLevel` (`lib/log-level.ts`) doesn't stop at `--`, so a schema-derived handler flag `--log-level <valid-level>` in a payload gets eaten — the exact hazard `extractNoColor` now guards against.
- `piece link` still converts `LinkValidationError` (a data failure discovered after reading both pieces) into a Cliffy `ValidationError` at `commands/piece.ts:730` — the same usage-screen-over-a-data-error disease this PR just cured for `piece get`.

The topic's acknowledged leftovers (static `--version 0.0.1` means stale-binary skew isn't self-diagnosing; `Called handler …` confirmation on stdout) seem fine to defer as documented there.

packages/cli/commands/piece.ts
```diff
@@ -115,6 +149,12 @@ function pieceCallRawArgs(tail: string[], literalArgs: string[]): string[] {
     );
   }
 
+  // "-" is the conventional stdin sentinel; route it through the existing
+  // --json-file stdin path so empty stdin still fails loudly.
+  if (tail[0] === "-") {
```
mathpirate, line 254: ⚪ Nit: `cf piece call <piece> <handler> - -- --query milk` silently ignores the `-` — `literalArgs` win at the top of this function, so the piped stdin is never read. That's the same silent-no-op class this PR is stamping out, on an argument shape this PR introduces. Cheap guard: throw when a payload argument (inline or `-`) and post-`--` flags are both present, mirroring the existing `tail.length > 1` rejection just above.

willkelly:
Thanks for the review — all three findings addressed in d226e8df8:

- 🔴 **docs**: `skills/cf/SKILL.md` and `packages/cli/README.md` now state that `piece get` prints literal `undefined` (not valid JSON, test before parsing) for an absent value, with the loom rationale — no more "parse it directly / `null`".
- 🟡 **color precedence**: `resolveColorEnabled` now reads the raw `NO_COLOR` var through its injected env seam instead of trusting the pre-arbitrated `Deno.noColor`, so `NO_COLOR` wins over both force vars consistently. Verified live: `NO_COLOR=1 FORCE_COLOR=1 cf piece get --help` → 0 ANSI (was 22); `NO_COLOR=1 CLICOLOR_FORCE=1` → 0; `FORCE_COLOR=1` alone → 22. Added an env-level test that reproduces the bug (`denoNoColor:false` + `NO_COLOR` + `FORCE_COLOR`).
- ⚪ **nit**: `pieceCallRawArgs` now throws when a payload token (inline JSON or `-`) and post-`--` flags are both present, instead of silently dropping the payload — mirroring the adjacent `tail.length > 1` rejection.

Full `packages/cli` suite green; local coverage-debt is flat/slightly improved. The two possible-follow-ups you flagged are handled in #4842 (reviewed).

willkelly:
Both cubic suggestions addressed in 79c86ec43:

- **P3 (main.ts:57)** — `cf --help` COLOR section now lists `CLICOLOR_FORCE=1` alongside `FORCE_COLOR=1`, matching the resolver and the README.
- **P2 (piece.ts)** — `pieceCallRawArgs` now forwards the explicit two-token stdin sentinels `--json-file -`, `--value-file -`, and `--json -` to the exec layer instead of rejecting them (or, for `--json -`, silently dropping the `-`). This makes `cf piece call … --json-file -` consistent with `cf exec` and the bare `-` form. A file *path* (not `-`) still requires `--`, so it's not mistaken for a sentinel. Unit test added.

Full `packages/cli` suite green.

mathpirate:
_cf-review round 2 via Claude (Fable 5), on behalf of @mathpirate — delta review of d47f75a8a..79c86ec43._

**All three findings resolved — approve.** Verified in-tree and live at 79c86ec43; full `packages/cli` suite green locally, CI 45-pass.

- 🔴 **docs**: both files now teach literal `undefined` with the loom rationale, README as the single home and the SKILL pointing at it. Right shape.
- 🟡 **color precedence**: nicer fix than the one I suggested — reading raw `NO_COLOR` *inside* `resolveColorEnabled` through the injected env seam keeps decision and test in one place, and the new env-level test reproduces the actual bug state (`denoNoColor:false` + both vars) for **both** force vars, plus the `NO_COLOR=""`-is-unset spec case. Re-verified live: `NO_COLOR=1 FORCE_COLOR=1` → 0 escapes, `NO_COLOR=1 CLICOLOR_FORCE=1` → 0, `FORCE_COLOR=1` piped → 22.
- ⚪ **payload + `--` guard**: loud on both payload shapes, legit flags-only shape preserved, tests pin all three.

Also read the cubic-round code since it's new since my review: the two-token sentinel forwarding is sound (the exec layer handles all three flags) and tested, including path-isn't-a-sentinel. One ⚪ observation, fine to ship as-is: bare `--json-file <path>` still dies with the generic `single inline JSON argument or "--"` message, which doesn't name the working `-- --json-file <path>` route — worth a message touch-up if this function gets edited again. Same shelf: README's "see … the PR compatibility rationale" points durable docs at a PR body; inlining the constraint's name would outlive the link.

The follow-ups from round 1 live in #4842 (stacked on this head, review incorporated).

mathpirate (approved):
lgtm ty!

PR #4773, opened 2026-07-16
refactor(ts-transformers): remove dormant __cfDataHelper wiring

mathpirate:
**Stacked on #4767** (base is its branch; will rebase onto main and mark ready once #4767 merges). Completes what #4767's second commit started: #4767 removed the orphaned string-pass `injectCfDataHelper` entry point; this removes the rest of the dormant `__cfDataHelper` wiring that spec §15.6 documented.

## What the wiring was

Opt-in-era (#3168) remnant: when `__cfHelpers` was absent, `wrapWithCfData` fell back to a `__cfDataHelper` import binding or a bare identifier plus a prepended `import { __cf_data as __cfDataHelper }` (`createCfDataHelperImport`), and `shouldWrapTopLevelExpression` narrowed to primitive-snapshot calls only.

## Why it can go

- **Unreachable, structurally**: `ModuleScopeCfDataTransformer extends HelpersOnlyTransformer`, whose `filter` is literally `sourceHasHelpers()` — every dormant fork sat behind an always-true first arm. Empirically: zero fixture goldens contain `__cfDataHelper`.
- **Out of contract if ever reached**: `__cfDataHelper` is not in the sandbox verifier's `TRUSTED_DATA_HELPERS` (`schema`, `__cf_data`, `nonPrivateRandom`, `safeDateNow`), so the fallback's wraps would not classify. The dormancy was not belt-and-braces — it was a silent semantic downgrade to an unverifiable form.
- **The loud invariant is free**: `getHelperExpr` already throws on a helpers-less source, so removal converts "silent fallback" into "loud failure" with no new code. A new test pins this.

## What was removed / kept

Removed: both fallback arms in `wrapWithCfData`, the helpers-less narrowing in `shouldWrapTopLevelExpression`, the import-prepend + `createCfDataHelperImport`, and `CFHelpers`' `getDataHelperExpr` / `sourceHasDataHelper` / `getCFDataHelperIdentifier` recognizer surface + `CF_DATA_HELPER_IDENTIFIER` (no external importers; the `cf view` pager re-declares its own copy). Net −173 lines.

Kept deliberately: `isTrustedDataHelper` / `TRUSTED_DATA_HELPERS` (a different, live "data helper" concept), the `cf view` vocab entry (a viewer's vocabulary legitimately covers historical tokens in old stored output), and `test/transform.test.ts`'s assertion that `__cfDataHelper` never appears in opted-out output.

Tests: the checker-dependent classification tests (primitive/union/intersection snapshots, callable walks) are **rewritten to the helpers-present form**, not deleted — using `declare const` callees so the primitive-snapshot arm isn't short-circuited by the local-callable arm. Their live-code coverage is preserved and improved: `module-scope-cf-data.ts` is now **100% branch/function/line** (cf-helpers.ts 97.3/99.4). Spec §15.6 rewritten as a removal record.

## Verification

- Package gate clean: fmt/lint/check, tests **1079/0** (count reconciles exactly: −8 dead-branch tests, +1 invariant test vs #4767's 1086).
- **Zero fixture-golden diffs** — live-path emission is byte-identical, so the runner's sandbox-verifier contract (`sandbox-contract.ts` pattern-matching) is provably untouched.
- All 11 external importer files of the package `deno check` clean; runner `esm-engine` + `load-by-identity` suites green.

### DIFF
--- a/docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md
+++ b/docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md
@@ -1983,25 +1983,27 @@ argument (`__cf_data((exports.x = evil, 1))`), and pipeline-compiled bodies are
 required precisely because bare `ts.transpileModule` cannot produce the
 `__cf_data` wrapping (`packages/runner/src/sandbox/module-record-compiler.ts`).
 
-### 15.6 Dormant alternate wiring (`__cfDataHelper`), historical
+### 15.6 Alternate wiring (`__cfDataHelper`), removed — historical
 
-The transformer and `CFHelpers` retain a second, data-helper-only emission
-path: when `__cfHelpers` is absent, `wrapWithCfData` falls back to a named
-`__cfDataHelper` import binding (`CFHelpers.getDataHelperExpr`) or a bare
-`__cfDataHelper` identifier plus a prepended
-`import { __cf_data as __cfDataHelper } from "commonfabric"`
-(`createCfDataHelperImport`, `CF_DATA_HELPER_IDENTIFIER`), and
-`shouldWrapTopLevelExpression` narrows to primitive-snapshot calls only. All
-of that is unreachable in the current pipeline: the `HelpersOnlyTransformer`
-filter guarantees `sourceHasHelpers()` at every wrap site. It is a remnant of
-the opt-in era (#3168), when modules without a `/// <cts-enable />` directive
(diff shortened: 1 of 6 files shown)


mathpirate:
Note for the trail: this was folded into #4767 as its third commit per Gideon (the fast-forward push of the base branch is why GitHub shows this as "merged" — merged into #4767's branch, not into main). Review happens on #4767, which now carries the whole audit-item-7 story.

PR #4694, opened 2026-07-13
docs(cts): reconcile transformer + schema-generator docs with implementation (2026-07 audit); fix anonymous default-export hardening

mathpirate:
## What this is

A claim-by-claim audit and update of the CTS transformer and schema-generator documentation. The branch is rebased on current `main` and reconciles the specs with the implementation, tests, fixtures, and the current 22-stage transformer pipeline.

The audit also surfaced and fixes one behavior bug: anonymous `export default function` declarations could emit a synthetic binding under one name and export a different name. They are now hardened in place, with exact-shape and runtime coverage.

## Main changes

- Adds a current-behavior spec and corpus map for `ts-transformers`, plus a TypeScript-to-JSON-Schema mapping spec for `schema-generator`.
- Corrects the CFC/IFC surface to the current aliases and lowering behavior, including `AnyOf`, `PolicyOf`, `WriteAuthorizedBy`, UI contracts, and the current API metadata keys.
- Reconciles all live pipeline documentation with the canonical 22-stage order, including the array-callback and derive-to-lift guides.
- Documents the complete lift scheduler-options contract: `materializerWriteInputPaths`, `completeSchedulerScopeSummary`, completeness failure conditions, and final function-first argument order.
- Documents identity/comparable capability analysis and schema shrinking, including trusted equality/navigation provenance, identity paths/cell paths, comparable wrappers, wildcard overlap semantics from #4714, and scheduler-proof interaction.
- Documents #4574's exact legacy injected-envelope detector and its storage-only, Merkle-verified runner trust boundary without weakening authored-source rejection.
- Adds drift guards: the spec-sync test validates the ordered pipeline and named stage ordinals across the live behavior documents.
- Pins schema-generator edge behavior for tuples, enums, literal encoding, and literal widening; removes redundant test coverage and makes the package `check` task exclude raw fixture inputs that require the synthetic test prelude.
- Moves superseded material into `docs/history/` with consistent lifecycle metadata and index entries, while restoring the live tutorial navigation entry.
- Adds package entry guides and a reusable `spec-audit` skill focused on durable authority/code/test/runtime mapping.

## Where to start reviewing

- Audit report: `docs/history/cts-docs-audit-2026-07.md`
- CTS corpus map: `docs/specs/ts-transformer/README.md`
- Current behavior: `docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md`
- Schema mapping: `docs/specs/schema-generator/ts_to_json_schema_mapping.md`
[...]

PR #4779, opened 2026-07-17
feat(ts-transformers): preserve optional call semantics

mathpirate:
## What

- Treat optionality as orthogonal to call legality: optional receiver, optional invocation, and combined optional chains now follow the same call-root and expression-site policy as their nonoptional counterparts.
- Lower every supported form as one complete call expression so receiver binding, single evaluation, authored short-circuit points, and skipped argument evaluation are preserved.
- Remove the standalone `optional-call` rejection category and its blanket capability wildcarding; unsupported calls remain unsupported for the underlying call-root or expression-site reason.
- Retain concrete readonly receiver capabilities for optional Cell reader chains such as `value.auth?.get?.()`, so nested Cells remain present in synthesized closure schemas without widening the enclosing root to wildcard.
- Align the target-language spec, lowering contract, goals, current-behavior spec, and design-delta record with that model.
- Add policy, capability, validation-matrix, and golden-fixture coverage across JSX, computed/module expressions, pattern returns, collection callbacks, and execution callbacks.
- Keep the authored pattern corpus unchanged. The previous 70-site source migration and its coverage-debt override are no longer needed.

## Semantic model

| Authored form | Rule |
| --- | --- |
| `value?.method(args)` | Uses the same eligibility as `value.method(args)` and preserves receiver short-circuiting. |
| `value.method?.(args)` | Uses the same eligibility as `value.method(args)` and preserves callable short-circuiting. |
| `value?.method?.(args)` | Uses the same eligibility and preserves both authored short-circuit points. |
| `fn?.(args)` | Uses the same site/root policy as `fn(args)`; optional syntax does not make an otherwise unsupported callable root legal. |
| Optional call with an unsupported underlying root/site | Remains unsupported for that underlying reason, not because it is optional. |

Function-valued reactive data remains outside the storable/schema-representable pattern-language model. This change preserves optional-call syntax where the corresponding call is already legal; it does not introduce function serialization or use optionality to bypass that boundary.

## Why

This resolves the optional-call follow-up identified during the #4694 transformer/spec review and consolidated in the [follow-up decisions topic](https://estuary.saga-castor.ts.net/topics-dev-476ea34f/fid1:O4Cm5vLYtmX-xpNJae9s5z35O6aOYyKp_7-Hv-9jxMs).

The previous revision treated optional calls as a distinct unsupported construct and rewrote 70 authored sites. That conflated two independent questions:

1. whether a call root is legal at a lowering boundary; and
2. whether JavaScript should short-circuit before receiver/member invocation.

Making optionality orthogonal gives the pattern language the more compositional abstraction: supported calls retain JavaScript's optional-chain semantics, while unsupported calls remain governed by the existing root/site and storage constraints.

## Implementation notes

- Call-root classification no longer diverts `CallChain` nodes into an `optional-call` bucket.
- Pattern-body and expression-site lowering own the complete optional call when the underlying call root is supported.
- Pattern-context validation defers optional-chain callees to call-root policy while continuing to diagnose non-lowerable optional navigation.
- Capability analysis no longer widens every optional callee to wildcard solely because it is optional.
- Optional Cell `.get()` chains explicitly retain the readonly receiver path; this preserves the Cell wrapper in the closure schema while remaining non-wildcard.
- The golden output demonstrates that optional receiver/invocation chains stay intact inside a single lowered callback, including argument short-circuit behavior.

## Verification

Post-rebase verification on the exact pushed head:

- `deno task test` in `packages/ts-transformers` — 1,108 passed (748 fixture steps), 0 failed
- `deno task integration pattern-tests 1/5` — all 21 tests in the formerly failing CI shard pass, including `auth/ready-auth.test.tsx` (8/8)
- `deno task cfcheck` — all 423 current pattern files pass with their original authored optional calls
- `deno task check` — all 400 workspace paths type-check
- focused policy/capability/validation regressions and emitted-output inspection — pass
- `deno fmt --check` for all 209 transformer-package files and `git diff --check` — clean

## Coverage

No performance or coverage-debt override is requested. The branch no longer changes `packages/patterns`; removing the blanket source migration eliminates the previously reported six-line uncovered-code increase.

## Provenance

Focused follow-up to #4694. Rebased directly onto `main` at `0f9150b4b` on 2026-07-23; pushed branch head is `d4db1584c`. The base includes the merged Toolshed readiness fix from #4935. The final diff is limited to transformer policy/implementation, normative and descriptive specs, and regression fixtures/tests.

## Amendment (2026-07-23 review follow-through)

Pushed as `8f9b1da65` on the same branch:

- The validation matrix now pins rejection-side parity for restricted reads: `items?.get()`, `items.get?.()`, and `items?.get?.()` report the same `pattern-context:get-call` as the non-optional form — never `pattern-context:optional-chaining`.
- The current-behavior spec (§19.6) documents the residual this PR's resolution had left undocumented: invoking a function value that flows through reactive data compiles without diagnostics (function-typed captures are dropped from lift input schemas) and is dead code at runtime. The design-deltas addendum carries this forward as an explicit open sub-finding; call-site enforcement is filed as [CT-1905](https://linear.app/common-tools/issue/CT-1905/diagnose-invocation-of-function-values-from-reactive-data-call-site), and the declaration-site alternative stays open as a schema-generator design question.
- The lowering contract (§3.4) states the two deliberate boundaries of whole-call preservation: dependency subscription stays eager while short-circuiting governs value evaluation, and authored `?.` on total cell-space navigation is absorbed by `key(...)` lowering while value-space chains keep it verbatim.
- The target-language §4 example shows the plain and optional call forms side by side.

Amendment verification: `deno fmt --check` and `deno lint` clean, `deno task check` (400 workspace paths), `deno task check-docs` (518 blocks), full `packages/ts-transformers` suite — 1,108 passed (751 steps), 0 failed.

### DIFF
--- a/docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md
+++ b/docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md
@@ -2808,6 +2808,17 @@ pipeline. Current built-in behavior:
    schema generation resolves the resulting exact manifest marker. The
    `WriteAuthorizedBy` validation remains a separate transformer rather than a
    schema-generator responsibility.
+6. Invoking a function value that flows through reactive pattern data (for
+   example a function-typed input property called as `input.fn(...)` or
+   `input.fn?.(...)`) produces no diagnostic at any expression site. Function
+   values are not schema-representable, so synthesized lift input schemas
+   silently drop function-typed captures; the lowered closure then reads an
+   absent property at runtime, making the call dead code (`undefined` under
+   optional invocation, a runtime `TypeError` when a plain invocation is
+   reached). The target language classifies such callable roots as
+   unsupported — the problem is the unstorable function value, not the call
+   syntax — but that classification is not yet surfaced as an authoring-time
+   diagnostic. The open follow-up is recorded in the design-deltas addendum.
 
 ## 20. Test Coverage Snapshot
 
--- a/docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md
+++ b/docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md
@@ -471,18 +471,14 @@ Diagnostics emitted in all modes:
   - message instructs the author to move the use into a nested
     `computed(() => ...)` or module-scope `lift()`
 - **Error** `pattern-context:optional-chaining`
-  - optional calls in restricted reactive context **outside JSX and outside
-    explicit compute callbacks** — top-level (`input?.foo()`), statement
-    position, and collection callbacks (`items.map((item) =>
-    item?.toUpperCase())`) all error (`test/validation.test.ts:546,567,1509`)
-  - inside JSX expressions (`{maybeFn?.(1)}`, `{text?.trim()}`) and inside
-    `computed(...)` bodies the same shapes are accepted and lowered intact —
-    an unratified delta from the target-language matrix's unconditional
-    "Unsupported" (see the design-deltas 2026-07-10 record; note the
-    lowering drops function-typed captures from the lift input schema, so an
-    accepted reactive optional-call is dead code at runtime)
   - optional property / element access that appears outside a supported
     lowerable expression site
+  - optional calls do not receive this diagnostic merely because they are
+    optional: their underlying call root is classified by the same policy as a
+    non-optional call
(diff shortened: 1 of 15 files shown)


mathpirate:
Not correct, taking another swing later

PR #4783, opened 2026-07-17
fix(cli): wait for background FUSE mount readiness on events, not a poll

Hixie:
## The problem

`cf fuse mount --background` decided whether the daemon had come up by polling a
status file two hundred times at a hundred milliseconds apart. That put a twenty
second ceiling on success. A mount that was healthy but slow to start — a cold
cache, a large space, a loaded machine — was killed on exhaustion and reported to
the user as a failure, so a non-zero exit meant "failed, or merely slow".

Three smaller faults sat alongside it. The poll used a fixed delay as a
synchronization primitive: on first seeing the `mounted` state it slept and
re-read the file to confirm, which moved the window in which a dying child went
unnoticed rather than closing it. It polled even though the daemon already
publishes a real signal. And called without a status file path, the function
checked only that the process was alive and then returned success, having
confirmed nothing about the mount — production always passed the path, so the CLI
never reached it, but the exported function would have lied to any future caller.

All four are what the repository policy on timeouts, retry loops and sleeps
forbids, and all four were unnecessary, because two real events were already
available.

## The approach

The wait is now a race between those two events, with no deadline. The supervisor
is spawned through `Deno.Command`, so its `status` promise reports the exit that
means startup failed. The daemon publishes its state to a file beside the mount
state file, so a `Deno.watchFs` watch reports every write.

The watch covers the directory holding both files rather than the status file
alone. Neither file need exist when the wait starts, and readiness also depends on
the supervisor recording the child PID into the mount state file, since until it
does no status can be tied to this mount. The wait reads once after installing the
watch, so a daemon that reports before the watch began is still seen, and reads
again on each event after that.

Readiness has no deadline by design. A deadline can only turn a slow startup into
a reported failure, and startup carries no liveness signal finer than the process
itself, because the daemon's heartbeat begins only once it is already mounted.
Nothing distinguishes a slow child from a stuck one, so a child that never reports
now leaves the command waiting — interruptible and visible — rather than killing a
mount that was about to work. In tests and CI the ambient limit is the backstop.

The status file path is now a required parameter, so the shape that returned
success without checking anything cannot be expressed.

## Daemon-side changes

Status writes now land by rename, so a reader woken by a write sees a complete
document rather than a half-written one. A torn read mattered little when reads
happened on an unrelated tick, but the reader now reads at the moment of the
write.

`mounted` is published after the FUSE session loop is dispatched and the signal
handlers are installed, rather than before both. Previously the CLI could be told
a mount was ready while a SIGTERM would still have killed the daemon outright
instead of unmounting it cleanly.

## Verification

Driving the real flow end to end proved the headline fix and caught a bug the unit
tests could not see. A startup taking twenty-five seconds — past the old ceiling —
now succeeds cleanly, and the failure path still rejects the moment the process
exits. The bug: leaving a `for await` over a real `Deno.FsWatcher` closes it, so
the explicit close afterwards reported a bad resource, which would have crashed the
command right after a *successful* mount. One test now uses a real watch, because
a fake cannot stand in for that.

Each property was checked by breaking the implementation and confirming the test
failed for the expected reason. That exposed two tests of my own that passed for
the wrong reason: one buffered all its events before the wait began iterating, so
it never exercised a long wait, and one let the supervisor's exit win the
microtask race so the status file was never consulted. Both are rewritten to drive
wake-ups one at a time. A separate test fakes the clock and advances an hour while
the child is still starting, because an attempt-driven test cannot see a wall-clock
deadline — five hundred heartbeats complete in a few milliseconds, so a twenty
second timer could otherwise have been reintroduced with the suite still green.

`deno fmt --check` and `deno lint` are clean repo-wide; the CLI and FUSE suites
pass.

## Known bound left behind

`recordFuseChildPid` still retries twenty times at fifty milliseconds before
giving up, and giving up exits the supervisor, which the wait reports as a mount
that died during startup. Since no status can be tied to a mount until that PID is
recorded, it is a ceiling on readiness of the same shape, one layer down. It is
left in place because, unlike the ceiling removed here, it does real work: it also
stops a supervisor from waiting forever, holding a mount, when the CLI dies before
writing the state file. Removing it honestly means giving the supervisor an event
for the CLI's death rather than a deadline, which changes what the CLI hands it at
spawn time, so it belongs in its own change. It is recorded in
`docs/development/waiting-in-tests.md` rather than left as a bare constant.

---
## Summary by cubic
Switch `cf fuse mount --background` startup detection from polling to an event-driven wait with no fixed timeout. Slow-but-healthy mounts no longer fail on a deadline, and real startup failures are reported promptly; supervisor status writes are now serialized to prevent state regressions.

- **Bug Fixes**
  - Wait on real events: supervisor exit (`Deno.Command.status`) or daemon status writes via `Deno.watchFs` on the status directory.
  - Read once after installing the watch and again on each event; remove sleeps, attempt limits, and the “sleep-then-reconfirm” race.
  - Require `childStatusPath`; liveness alone is never treated as readiness.
  - Classify status by token/mountpoint and child PID; reject foreign or unreadable content as “not yet,” not a verdict.
  - Cleanup on failure is precise: remove only mount state or both state and child-status as appropriate.
  - Integration script now polls `result.json` content with `jq` until sigils render to avoid reading mid-rebuild.

- **Refactors**
  - Supervisor status writer moved to `createSupervisorStatusWriter`; writes are serialized and land via temp-file+rename, with `updatedAt` stamped at call time. Prevents a late heartbeat from overwriting `exiting/exited`.
  - Publish `mounted` after the FUSE session loop is dispatched and signal handlers are installed; continue 1s heartbeat.
  - Tests use a watcher seam and a fake supervisor to drive events; add a real `Deno.watchFs` test for close semantics and a fake-clock test to prove there’s no time-based deadline.
  - Document the remaining startup bound: `recordFuseChildPid`’s retry ceiling, recorded in `docs/development/waiting-in-tests.md`.

<sup>Written for commit 9df16b97dc8943fce774a51d9f190d33cb8b84e3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4783?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/development/waiting-in-tests.md
+++ b/docs/development/waiting-in-tests.md
@@ -422,9 +422,15 @@ mount that was about to work. In a test or CI lane the ambient limit is the
 backstop, in place of a per-command ceiling that fails healthy mounts.
 
 Each status write lands by rename, so a read woken by a write sees a complete
-document rather than a half-written one. Unreadable or foreign content reads as
-"not yet reported" rather than as a verdict; the daemon rewrites the file on
-every state change and on its heartbeat.
+document rather than a half-written one. The writes are also serialized, so the
+file settles on the state of the most recent call. The calls come from four
+places that do not coordinate — the startup path, the readiness report, the
+heartbeat and the signal handler — and two renames issued concurrently can
+complete in either order, so without that a heartbeat still in flight could
+replace a terminal state and leave the file reporting a mount that had already
+gone. Unreadable or foreign content reads as "not yet reported" rather than as
+a verdict; the daemon rewrites the file on every state change and on its
+heartbeat.
 
 One bound remains on this path, in a different process. `recordFuseChildPid`
 in `packages/cli/lib/fuse-supervisor.ts` is how the supervisor writes the child
--- a/docs/development/waiting-in-tests.md
+++ b/docs/development/waiting-in-tests.md
@@ -395,16 +395,44 @@ Two waits in this script are not polls, and should not become polls.
 
 Mount readiness needs no timing loop. `cf fuse mount --background` calls
 `awaitBackgroundMountStartup`, which waits for the daemon to report the
-`mounted` supervisor state and confirms both the supervisor and the child are
-alive before the command prints the PID. Every other exit from that function
-throws, and a throw kills the child and fails the command, so a script that has
-parsed a PID has a daemon that reported mounted.
+`mounted` supervisor state and confirms the child is alive before the command
+prints the PID. Every other exit from that function throws, and a throw kills
+the child and fails the command, so a script that has parsed a PID has a daemon
+that reported mounted.
 
-The daemon reports that state just before it enters its FUSE session loop, so it
-means the kernel mount exists rather than that any request has been served, and
-mounted paths hydrate lazily besides. Both gaps belong to `wait_for_path`, whose
-probe carries its own kill-timeout and retries for twenty seconds. What is left
-for the script is one check that the daemon survived the handshake.
+That wait is a race between two real events, with no poll and no deadline under
+it. The supervisor is spawned through `Deno.Command`, so its `status` promise
+reports the exit that means the startup failed. The daemon publishes its state
+to a file beside the mount state file, so a `Deno.watchFs` watch over the
+directory holding both reports every write to either. The directory is the unit
+watched rather than the status file, because neither file need exist when the
+wait starts, and because the wait also depends on the supervisor recording the
+child PID into the mount state file — until it does, no status can be tied to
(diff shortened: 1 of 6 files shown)


Hixie:
obsoleted by https://github.com/commontoolsinc/labs/pull/4788

PR #4797, opened 2026-07-17
fix(runner): preserve sink sync state

seefeldb:
## Summary

- preserve the synced state when `CellImpl.sink()` validates and transforms the value delivered to its callback
- add a real worker-reconciler regression covering nested VDOM `asCell` cut points
- ensure nested renderer sinks reuse the root query instead of opening a subscription for every element/cut point

## Root cause

`sink()` synchronizes the source cell before subscribing, but the subsequent `validateAndTransform()` call did not receive `{ synced: true }`. As a result, `asCell` projections created for the callback were marked unsynced even though the root query already covered them. The renderer's nested `.sink()` calls then invoked `syncCell()` again.

For the three-element VDOM in the regression, the broken behavior made 10 render-time `syncCell()` calls. The fixed behavior makes exactly one, for the mounted root.

## Validation

- `deno check packages/runner/src/cell.ts packages/html/test/worker-reconciler-sync.test.ts`
- `deno test -A packages/html/test/worker-reconciler-sync.test.ts`
- `deno test -A packages/runner/test/cell-as-cell.test.ts packages/runner/test/cell-callbacks.test.ts packages/runner/test/schema-anyof.test.ts` (4 tests, 123 steps)

---
## Summary by cubic
Preserves sync state in `sink()` so nested VDOM `asCell` projections reuse the root query and avoid redundant `syncCell()` calls. Also reuses schema coverage for `style` props to skip per-prop sinks and redundant updates.

- **Bug Fixes**
  - Pass `{ synced: true }` into `validateAndTransform()` from `sink()` so callback-created projections remain in-sync and nested renderer sinks reuse the root subscription.
  - Treat `style` as schema-resolved in the reconciler: exclude it from per-prop sinks, cancel prior per-prop sinks when a value becomes direct, and apply the same unchanged-prop predicate for direct updates to avoid extra ops.
  - Add VDOM regressions for root-only sync and style coverage; update prop tests to verify generic object props still use per-prop sinks and emit when becoming `undefined`.
  - Add a scheduler test to ensure time-gated actions skip until their next eligible iteration and converge on the following pass.

<sup>Written for commit 4e69e04a2f3b489507fb69da74cfc12c8ca99d80. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4797?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bfollington:
Followed your "let's measure what those actually are" — and you're right, which turns out to redirect the whole thing.

**The 607 non-style prop watches are cheap, exactly as you predicted.** Exact request-side cost: **102.6 KB, 173 B each, 0.9% of the 11.5 MB total.** They land on just **44 distinct cells** (~14 watches/cell), and **34 of those 44 never appear in another frame after initial delivery** — max touches for any one cell was 7. Small, one-shot, ≈ the cost of the first read. Not worth optimizing.

**So I measured where the bytes actually are, and it's not the count.** One cold load + click-through = **11.5 MB** (1.47 sent / 10.0 recv). 98.4% of received bytes are `watch.add` responses, and they're extremely concentrated:

- top 3 response frames = **62.5%** of all recv bytes
- top 8 = **96.3%**

The worst are **single-root** watches:

| response | roots | schema asked for |
|---:|---:|---|
| 763 KB | **1** | `$ref → #/$defs/compiledDoc` (recursive) |
| 636 KB | **1** | same |
| 162 KB | **1** | same |

That schema is 628 bytes and recurses `$ref: #/$defs/compiledDoc` through `imports`, so a single subscription pulls `code` + `sourceMap` + `policyManifests` + the transitive import graph. **3 subscriptions → 1.56 MB (15.6% of all recv bytes).** It's the compiled-module graph coming down the storage socket to a client that's rendering mobile Loom.

**The takeaway for where to spend implementation effort:** subscription *count* drives round-trips, and #4735 already handled that (148 → 100 frames for 1,910 roots). It does not drive *bytes* — the payload is ~10 fat recursive reads that cross the wire regardless of how watches are grouped. So the count-reduction work (coarse-watching `$UI`, monolithic egress, etc.) optimizes the wrong axis for this profile.

The high-leverage lever looks like **scoping the `compiledDoc` read**: does the client renderer need `code` / `sourceMap` / the recursive `imports` closure, or can that watch project a lean schema? That's ~1.5 MB from 3 watches vs ~100 KB from 600.

Caveat: this is a localhost capture, so it can't separate round-trip latency from payload — on a high-RTT link (estuary's ~32s settle) count still matters for latency. But the 10 MB has to move either way, and here it clearly dominates.

bfollington:
Closing the loop on this thread with a recommendation rather than more tables.

**I think the watch-refresh batch window is worth another look** — `scheduleWatchRefreshFlush` in `packages/runner/src/storage/v2.ts`, the mechanism from #4735. It looks like it's firing too narrowly. **Worth 4.7–12.9 s off a 37 s estuary load**; nothing else we measured comes close.

### The stat that makes the case

Roots per `session.watch.add` frame, mobile Loom cold load + click-through:

```
min 1  ·  p50 1  ·  max 445        77% of frames carry <= 5 roots
```

The **median frame carries one root**. #4735 clearly collapsed the big bursts (two frames carry 445 and 334 roots), but the distribution is bimodal — a few huge frames and a long tail of ~50 near-empty ones.

That tail is expensive because acquisition is **strictly serialized** (measured overlap factor 1.0× — no two watch requests ever in flight at once), so every frame costs a full round-trip. Estuary per-frame RTT floor is **188 ms**, and **83% of session wall-clock (30.6 s of 36.9 s) is blocked on watch round-trips**.

### How much of the tail is free

Classifying each frame by whether it *had* to be separate — did it request any cell id only revealed by the immediately-preceding response?

| | estuary | localhost |
|---|---:|---:|
| watch.add frames | 65 | 100 |
| dependent (needed prior response) | 40 (62%) | 61 (61%) |
| **mergeable (targets already known)** | **25 (38%)** | 39 (39%) |
| roots carried by those | 476 | 342 |
| **their measured latency** | **12.9 s** | 3.0 s |
| saving: RTT floor → full cost | **4.7 – 12.9 s** | 0.3 – 3.0 s |

25 frames on estuary carrying 476 roots had **no data dependency forcing them apart**.

### The ceiling, so nobody over-invests

**62% of frames are genuine dependencies** — each needs the prior response to know what to ask next. Real fetch waterfall; no batching window touches it. Past ~40% frame reduction you're changing the fetch graph itself (speculation/prefetch), which we would *not* suggest starting on this evidence.

### Ruled out, so you don't re-tread

- **Collapsing VDOM/per-prop subscriptions saves ~0 frames.** 607 non-style prop watches exist, but **zero frames contain only prop watches** — they ride in frames sent for other roots, so removing all of them removes no round-trips. (Also 102 KB total, 0.9% of bytes, 173 B each.) Your read that these are cheap was right, and it's stronger than cheap — they're structurally free.
- **No steady-state subscription cost.** Whole sessions produced **2** (localhost) and **4** (estuary) `session/effect` pushes. Subscriptions essentially never fire; it's all initial acquisition.
- **`compiledDoc` size** is real (8 heavy frames = 47% of estuary blocking; 3 single-root recursive reads pull 1.56 MB) but it's the code, and the recursion is deliberate. Only obvious slice is whether the renderer needs `sourceMap`/`code`.

### Caveats

Estuary capture is on `loom-stable-2026-07-15`, localhost on `-07-20`. **The estuary numbers predate #4735**, so its frame reduction (148 → 100 on localhost, −32%) is not reflected there — the mergeable count on estuary may already be lower today. "Mergeable" is a *data*-dependency test only; there may be code-structure reasons those frames are separate, so treat it as the size of the opportunity, not proof of ease. And localhost hides all of this (frame count is 8% of blocking there vs 39% on estuary) — this only shows up over real latency.

PR #4806, opened 2026-07-17
deps(toolshed): roll ai to 7 with the @ai-sdk providers and arizeai p…

Hixie:
…ackages

Rolls seven packages as one set, adds @ai-sdk/otel to complete the telemetry migration ai 7 requires, and adds a lockfile check so the set cannot drift back apart:

    ai                                          ^5.0.27 -> ^7.0.30
    @ai-sdk/anthropic                            ^2.0.9 -> ^4.0.15
    @ai-sdk/openai                              ^2.0.22 -> ^4.0.15
    @ai-sdk/groq                                ^2.0.16 -> ^4.0.11
    @ai-sdk/google-vertex                       ^3.0.16 -> ^5.0.21
    @arizeai/openinference-vercel                ^2.0.1 -> ^3.1.0
    @arizeai/openinference-semantic-conventions  ^1.1.0 -> ^2.5.0

They have to move together. Version 3 of @arizeai/openinference-vercel depends on @ai-sdk/otel, which depends on ai 7 as a regular dependency. Rolling the arizeai packages while toolshed stayed on ai 5 resolved a second copy of ai and of @ai-sdk/provider-utils: toolshed emitted spans from one copy of the SDK while the arizeai span processor, built against the other, translated them. Nothing failed locally. Rolling the whole set leaves one copy of each, and the arizeai package resolves against the same ai toolshed uses. The arizeai semantic conventions were already duplicated before this change and are now down to a single copy as well.

Two changes in ai 7 needed more than a version bump.

The ai package no longer collects OpenTelemetry spans. Collection moved to a new @ai-sdk/otel package and only happens once an integration is registered through registerTelemetry. Without that registration toolshed would emit no LLM spans at all and nothing would fail: the traces would simply stop arriving. lib/otel.ts now registers the integration against the same provider the OpenInference span processor is attached to, so LLM spans reach the same exporter as before. lib/otel.test.ts asserts that importing the module registers an integration, so the silent case is caught.

experimental_telemetry also lost its metadata and tracer properties. Per-request metadata now travels as runtimeContext, which telemetry withholds unless each property is named in includeRuntimeContext, so the LLM route builds that list from the metadata it is handed. The custom tracer moved to the OpenTelemetry constructor. The supplemental runtime-context attributes only reach the operation and step spans, so an enrichSpan callback puts the same attributes on the model-call span, which is the span OpenInference labels LLM and which carried this metadata before the roll. enrichSpan flattens nested values the way @ai-sdk/otel does, so every span of a trace agrees. Splitting metadata into runtime context lives in lib/ai-telemetry.ts next to the integration that consumes it, and is covered directly.

The generateObject route opts out of telemetry explicitly. Once an integration is registered, telemetry in ai 7 is opt-out rather than opt-in, and that route has never emitted AI SDK spans; without the opt-out this roll would have started exporting new spans for it.

The single copy that makes the set safe is a pin coincidence, not a structural guarantee: @ai-sdk/otel pins ai exactly rather than depending on a peer range, so an ai release published ahead of a matching @ai-sdk/otel would separate them again, and a routine update would pick it up. deno task check-single-copy-deps reads deno.lock and fails when ai, @ai-sdk/provider-utils, or the arizeai conventions package resolves more than once, naming the copies. It runs in CI beside the other lockfile and pin checks, and tasks/check-single-copy-deps.ts holds the list with what breaks when each package is duplicated.

---
## Summary by cubic
Upgraded `ai` to v7, moved telemetry to `@ai-sdk/otel`, and aligned `@ai-sdk` providers with Arize packages to keep one SDK copy. Unified request metadata so strings/numbers/booleans and JSON-serialized objects appear on both HTTP and LLM spans; added a CI lockfile guard to prevent duplicate SDK copies.

- **Dependencies**
  - Bumped: `ai` → `^7.0.30`; `@ai-sdk/{anthropic,openai,groq}` → `^4.x`; `@ai-sdk/google-vertex` → `^5.x`; added `@ai-sdk/otel`; `@arizeai/openinference-vercel` → `^3.1.0`; `@arizeai/openinference-semantic-conventions` → `^2.5.0`.

- **New Features**
  - Telemetry: `lib/ai-telemetry.ts` (`createAiSdkTelemetry`) registered in `lib/otel.ts` so LLM spans export; `enrichSpan` mirrors filtered `runtimeContext` onto the model-call span; tests assert integration is registered.
  - Metadata: centralized via `metadataAttributeValue` + `runtimeContextFromMetadata`; now carries strings/numbers/booleans as-is and objects/arrays as JSON on both HTTP and AI SDK spans; tests cover nested and non-string cases.
  - Routes: `generateText` uses `runtimeContext` with `telemetry.includeRuntimeContext`; `generateObject` opts out of telemetry.
  - CI guard: `deno task check-single-copy-deps` + workflow step ensures single copies of `ai`, `@ai-sdk/provider-utils`, and `@arizeai/openinference-semantic-conventions`; tests cover duplicate reporting and the script entry point; docs add the single-copy rule and roll command.
  - Coverage: tests for `tasks/write-coverage-lcov.ts` normalize instance paths and cover usage/error paths.

<sup>Written for commit 370ac6d4303b1da4df22f50ecf825e69a7ac59de. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4806?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ubik2 (commented):
_cf-review via Codex (GPT-5), on behalf of @ubik2_

Post-merge review of `370ac6d`: dependency/telemetry migration deep-read, with delegated checks over the lockfile guard, docs, call-site blast radius, and tests. Focused telemetry, route, guard, formatting, and lint checks passed; final GitHub CI was green.

### 🔴 Blocking

**Critical, verified — `OTEL_ENABLED=false` still exports AI SDK spans** (`packages/toolshed/lib/otel.ts:64-68`, `:121-129`; contradicted by `docs/development/CONFIGURATION.md:203-210`). The integration is registered unconditionally at module import with a tracer obtained directly from the live `BasicTracerProvider`. `initOpenTelemetry(false)` only skips global provider/context registration, while `generateText` enables telemetry. In a real AI SDK v7 probe with `OTEL_ENABLED=false`, Toolshed logged that OpenTelemetry was disabled and then exported three spans.

This bypass already existed for `generateText` before the PR because it supplied the provider-owned tracer per call, but the new process-global registration preserves and broadens the violation: every current or future AI SDK call exports unless it explicitly opts out. That can send prompts and request metadata to a configured collector despite the operator-facing opt-out. Please move `registerTelemetry(createAiSdkTelemetry(...))` into the enabled `initOpenTelemetry()` path and guard it against duplicate registration.

### 🟡 Improvement

**Minor, verified — production route wiring is not covered by successful-call tests** (`generateText.ts:294-300`, `generateObject.ts:77-79`). `ai-telemetry.test.ts` proves the integration works when the test manually supplies `runtimeContext` and `includeRuntimeContext`, but no route test proves `generateText` supplies them or that `generateObject` suppresses globally registered telemetry. Add successful `MockLanguageModelV4` route tests for those two policies.

Correction to the initial local analysis: there is **no LCOV finding**. The apparent failure was caused by the review sandbox blocking writes to Deno's transpilation cache. The exact test command passes outside the sandbox.

Verdict: the package roll, single-copy guard, and metadata migration are otherwise coherent. The telemetry opt-out violation warrants a follow-up fix on `main`.

Edit: Human here -- I had a typo in my review PR instruction, so my agent reviewed this PR instead. It had some other invalid feedback, but this piece seemed useful enough to attach, even though the PR's already merged, and it's not exactly a new bug.

PR #4817, opened 2026-07-18
data-model: move freeze-protocol plumbing off FabricInstance onto BaseFabricInstance

danfuzz:
## What

Structural refactor within `packages/data-model`: move the freeze-protocol
plumbing off the pure `FabricInstance` protocol interface and down onto the
concrete `BaseFabricInstance` base class. No runtime behavior change.

## Why

The `[DEEP_FREEZE]` / `[IS_DEEP_FROZEN]` symbol-keyed members are implementation
plumbing — the machinery the generic freeze routines dispatch through — not part
of the pure, `instanceof`-able protocol that external code is written against.
`FabricInstance` (in `interface.ts`) is the dependency-free protocol leaf;
`BaseFabricInstance` is where the concrete instance machinery already lives. Keeping
the freeze plumbing off the protocol class leaves the interface as a clean
client-facing surface and co-locates the plumbing with the base that actually
implements around it.

## The member split

- **Moved down to `BaseFabricInstance`:** the `[DEEP_FREEZE]` and `[IS_DEEP_FROZEN]`
  `unique symbol` consts and their abstract member declarations. (The symbol consts
  travel with their declarations, for cohesion.)
- **Stays on `FabricInstance`:** `deepClone()` and `shallowClone()` — these remain
  the client-facing surface.

## The gate flip

The generic freeze machinery in `deep-freeze.ts` gated on
`instanceof FabricInstance` before invoking those symbols. Since the members no
longer type against `FabricInstance`, the three dispatch sites (`isDeepFrozen`,
`deepFreeze`, `isDeepFrozenFabricValue`) now narrow on
`instanceof BaseFabricInstance`. Every concrete instance already extends
`BaseFabricInstance`, so each gate is runtime-equivalent.

## Behavior-preserving

This is a structural move with no external contract change — everything is within
`data-model`. Importers of the two symbols (subclass impls, `deep-freeze.ts`, and
tests) are re-pointed to the new home; the only test changes are import-path
re-points. The data-model suite passes unchanged.

Co-Authored-By: coder-cedar (Claude Opus 4.8) <noreply@anthropic.com>
Co-Authored-By: upstream-liaison-bolt (Claude Opus 4.8) <noreply@anthropic.com>

---
## Summary by cubic
Moved the freeze-protocol members off `FabricInstance` onto `BaseFabricInstance` in `packages/data-model`, added `isInstance()` invariant guards for instances/primitives, and seeded `BaseFabricPrimitive` so its guard narrows normally. No external behavior changes.

- **Refactors**
  - Declared `[DEEP_FREEZE]` and `[IS_DEEP_FROZEN]` on `BaseFabricInstance`; `deep-freeze.ts` now gates via `BaseFabricInstance.isInstance()` and `BaseFabricPrimitive.isInstance()` (not `instanceof`); JSDoc clarifies the invariant-guard throw.
  - Added `BaseFabricInstance.isInstance()` and `BaseFabricPrimitive.isInstance()` to enforce “every subclass extends the base.”
  - Seeded `BaseFabricPrimitive` with `[EXAMPLE_METHOD]` (placeholder) so its instance type is non-empty; added unit test for the stub.
  - Restored primitive-first order in `isDeepFrozenFabricValue()`; repointed imports across subclasses and tests; `deepClone()`/`shallowClone()` stay on `FabricInstance`.
  - Placed static `isInstance()` after instance members on both bases per data-model convention; reorder only.

<sup>Written for commit 36682cf4bb10dd24d5465b97f58382d836fd06a8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4817?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/data-model/src/deep-freeze.ts
+++ b/packages/data-model/src/deep-freeze.ts
@@ -328,7 +328,11 @@ export function isDeepFrozenFabricValue(value: unknown): value is FabricValue {
 
     seen.add(item);
 
-    if (BaseFabricInstance.isInstance(item)) {
+    if (BaseFabricPrimitive.isInstance(item)) {
+      // `FabricPrimitive`s are by definition frozen and have no outbound
+      // references.
+      return true;
+    } else if (BaseFabricInstance.isInstance(item)) {
       // Object.freeze() cannot prove that a fabric instance's private logical
       // contents will remain unchanged, so validate it but do not root-cache a
       // graph that reaches one.
@@ -368,14 +372,6 @@ export function isDeepFrozenFabricValue(value: unknown): value is FabricValue {
         }
         if (!checkValue(record[key])) return false;
       }
-      return true;
-    } else if (BaseFabricPrimitive.isInstance(item)) {
-      // `FabricPrimitive`s are by definition frozen and have no outbound
-      // references. This arm is checked after the array / plain-object arms
-      // (unlike the primitive-first ordering elsewhere): `BaseFabricPrimitive`'s
-      // instance type has no members, so this guard's negative narrowing would
-      // collapse `item` to `never` for any following branch -- keeping it last
-      // confines that to the unreachable-for-`item` final `else`.
       return true;
     } else {
       // It's an instance of a class that isn't covered by the `FabricValue`
--- a/packages/data-model/src/deep-freeze.ts
+++ b/packages/data-model/src/deep-freeze.ts
@@ -1,12 +1,13 @@
 import { isPlainObject } from "@commonfabric/utils/types";
 import { isArrayWithOnlyIndexProperties } from "@commonfabric/utils/arrays";
 
-import { FabricPrimitive, FabricValue } from "./interface.ts";
+import type { FabricValue } from "./interface.ts";
 import {
   BaseFabricInstance,
   DEEP_FREEZE,
   IS_DEEP_FROZEN,
 } from "./fabric-instances/BaseFabricInstance.ts";
+import { BaseFabricPrimitive } from "./fabric-primitives/BaseFabricPrimitive.ts";
 
 /**
  * Cache of confirmed deep-frozen objects.
@@ -121,18 +122,19 @@ export function isDeepFrozen(value: unknown): boolean {
 
     let result = true;
 
-    if (obj instanceof BaseFabricInstance) {
+    if (BaseFabricInstance.isInstance(obj)) {
       // A fabric instance's logical contents are not its enumerable own-props
       // (e.g. a `FabricError` keeps its custom properties in a private extras
       // `Map`), so it answers the deep-frozen question via its
       // `[IS_DEEP_FROZEN]` protocol member -- the side-effect-free sibling of
       // `[DEEP_FREEZE]` -- recursing into each nested `FabricValue` through
-      // `check`, which shares this call's cycle state. Gating on `instanceof`
-      // against `BaseFabricInstance` keeps this generic; the member is abstract
-      // on `BaseFabricInstance`, so every instance implements it. (A
-      // `FabricPrimitive` is necessarily frozen with no outbound references, so
-      // the `Object.values` arm below answers it correctly by accident -- its
-      // empty enumerable props yield `true`.)
+      // `check`, which shares this call's cycle state. Gating via
+      // `BaseFabricInstance.isInstance()` keeps this generic (and enforces the
+      // "every `FabricInstance` is a `BaseFabricInstance`" invariant); the
+      // member is abstract on `BaseFabricInstance`, so every instance implements
+      // it. (A `FabricPrimitive` is necessarily frozen with no outbound
+      // references, so the `Object.values` arm below answers it correctly by
+      // accident -- its empty enumerable props yield `true`.)
       result = obj[IS_DEEP_FROZEN](check);
     } else if (Array.isArray(obj)) {
       for (let i = 0; i < obj.length; i++) {
@@ -171,9 +173,9 @@ export function isDeepFrozen(value: unknown): boolean {
  *    at construction and have no outbound references.
  * 3. Fabric instance: delegate generically to its `[DEEP_FREEZE]` protocol
  *    member, handing recursion through as the `subFreeze` callback. The
- *    dispatch gates on `instanceof` against `BaseFabricInstance` (where the
- *    member is declared) -- it operates generically and does not enumerate
- *    concrete subclasses.
+ *    dispatch gates via `BaseFabricInstance.isInstance()` (where the member is
+ *    declared) -- it operates generically and does not enumerate concrete
+ *    subclasses.
  * 4. Plain object or array: recursively freeze children, then freeze the
  *    container.
  *
@@ -198,7 +200,7 @@ export function deepFreeze<T>(value: T): T {
   // construction) and have no outbound references. Handling arms 1 and 2 here,
   // before allocating the cycle-tracking set or the recursion closure below,
   // keeps primitives and `FabricPrimitive`s off the heavyweight path.
-  if (value instanceof FabricPrimitive) {
+  if (BaseFabricPrimitive.isInstance(value)) {
     return value;
   }
 
@@ -218,7 +220,7 @@ export function deepFreeze<T>(value: T): T {
     if (isNecessarilyOrKnownDeepFrozen(value)) {
       return value;
     }
-    if (value instanceof FabricPrimitive) {
+    if (BaseFabricPrimitive.isInstance(value)) {
       return value;
     }
 
@@ -238,7 +240,7 @@ export function deepFreeze<T>(value: T): T {
     // closes over `inProgress`, so the impl's recursion into nested
     // `FabricValue`s shares cycle state with this call -- the participating
     // instance doesn't need to be `inProgress`-aware in its own signature.
-    if (value instanceof BaseFabricInstance) {
+    if (BaseFabricInstance.isInstance(value)) {
       const result = value[DEEP_FREEZE](freeze) as U;
       // Cache the now-deep-frozen result so subsequent `isDeepFrozen()` checks
       // short-circuit in O(1), mirroring arm 4's cache-write below.
@@ -326,19 +328,16 @@ export function isDeepFrozenFabricValue(value: unknown): value is FabricValue {
 
     seen.add(item);
 
-    if (item instanceof FabricPrimitive) {
-      // `FabricPrimitive`s are by definition frozen and have no outbound
-      // references.
-      return true;
-    } else if (item instanceof BaseFabricInstance) {
+    if (BaseFabricInstance.isInstance(item)) {
       // Object.freeze() cannot prove that a fabric instance's private logical
       // contents will remain unchanged, so validate it but do not root-cache a
       // graph that reaches one.
       cacheableByIdentity = false;
       // Fabric instances answer the deep-frozen question via their
       // `[IS_DEEP_FROZEN]` protocol member (the side-effect-free sibling of
-      // `[DEEP_FREEZE]`), recursing through `checkValue`. Gating on
-      // `instanceof` against `BaseFabricInstance` keeps this guard generic; the
+      // `[DEEP_FREEZE]`), recursing through `checkValue`. Gating via
+      // `BaseFabricInstance.isInstance()` keeps this guard generic (and enforces
+      // the "every `FabricInstance` is a `BaseFabricInstance`" invariant); the
       // `[IS_DEEP_FROZEN]` member is abstract on `BaseFabricInstance`, so every
       // instance is guaranteed to implement it.
       return item[IS_DEEP_FROZEN](checkValue);
@@ -369,6 +368,14 @@ export function isDeepFrozenFabricValue(value: unknown): value is FabricValue {
         }
         if (!checkValue(record[key])) return false;
       }
+      return true;
+    } else if (BaseFabricPrimitive.isInstance(item)) {
+      // `FabricPrimitive`s are by definition frozen and have no outbound
+      // references. This arm is checked after the array / plain-object arms
+      // (unlike the primitive-first ordering elsewhere): `BaseFabricPrimitive`'s
+      // instance type has no members, so this guard's negative narrowing would
+      // collapse `item` to `never` for any following branch -- keeping it last
+      // confines that to the unreachable-for-`item` final `else`.
       return true;
     } else {
       // It's an instance of a class that isn't covered by the `FabricValue`
--- a/packages/data-model/src/deep-freeze.ts
+++ b/packages/data-model/src/deep-freeze.ts
@@ -1,13 +1,12 @@
 import { isPlainObject } from "@commonfabric/utils/types";
 import { isArrayWithOnlyIndexProperties } from "@commonfabric/utils/arrays";
 
+import { FabricPrimitive, FabricValue } from "./interface.ts";
 import {
+  BaseFabricInstance,
   DEEP_FREEZE,
-  FabricInstance,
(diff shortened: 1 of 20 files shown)


danfuzz:
**Reviewer: APPROVE** (reviewed at head `b29ccef7a`)

Behavior-preserving structural move: the `[DEEP_FREEZE]` / `[IS_DEEP_FROZEN]` symbol consts and their abstract declarations move off the pure-protocol `FabricInstance` (interface.ts) down onto `BaseFabricInstance`, and the three generic-freeze gates re-narrow accordingly. Reviewed all 18 files at-source.

What I checked:

1. **Move completeness** — both symbol consts + both abstract decls are now on `BaseFabricInstance` and gone from `interface.ts` (only an intentional JSDoc *reference* remains, explaining the new home); `interface.ts` retains exactly `deepClone()` / `shallowClone()` as the client surface. No lingering symbol import from `interface.ts` anywhere (grep-confirmed).
2. **Gate re-narrowing sound at all 3 sites** (`deep-freeze.ts:125`/`:242`/`:334`) — each flipped to `instanceof BaseFabricInstance`. Verified the full runtime hierarchy: every concrete fabric instance roots at `BaseFabricInstance` (`FabricLink` direct; `FabricError`/`FabricSet`/`FabricMap` via `FabricNativeWrapper`; `ProblematicValue`/`UnknownValue` via `ExplicitTagValue`). The only runtime `extends FabricInstance` class is `BaseFabricInstance` itself; the `api/index.ts` hit is a TS `interface` (no `instanceof` participation). Nothing bypasses the narrowed gate, so it is runtime-equivalent.
3. **No import cycle** — `interface.ts` stays the dependency-free leaf; `BaseFabricInstance` imports only `FabricInstance` + `type FabricValue` from it; subclasses and `deep-freeze.ts` import the symbols from `BaseFabricInstance`. Graph stays acyclic.
4. **No disguised logic change** — all 9 test-file edits are pure import-path re-points; no assertion or logic changed. Also confirmed the four importers *outside* the diff (`cloneIfNecessary`/`deep-freeze` tests, `fixtures.ts`, schema-generator `factory-input-real-api`) reference the symbols in prose only and were correctly left untouched.
5. **JSDoc/comment fidelity** — the abstract-member docs travel with their declarations and correctly re-name to `BaseFabricInstance`; the `deep-freeze.ts` comments update in step.
6. **api mirror untouched** — `api/index.ts` mirrors only `shallowClone`, never the symbols; correctly unchanged.
7. **Scope + hygiene** — no `[CODEC]` touch; comments are project-technical only; `Co-Authored-By` trailer present.

Independent confirmation: ran `deno check` on the three core files — clean (the narrowed-gate symbol invocations type-check against `BaseFabricInstance`).

CI is still in flight (no failures so far); this APPROVE is contingent on it going green, after which the PR queues for Dan's merge.

packages/data-model/src/deep-freeze.ts
```diff
@@ -122,18 +121,18 @@ export function isDeepFrozen(value: unknown): boolean {
 
     let result = true;
 
-    if (obj instanceof FabricInstance) {
-      // A `FabricInstance`'s logical contents are not its enumerable own-props
+    if (obj instanceof BaseFabricInstance) {
```
danfuzz: Related to the Cubic comment, and a refinement of the recommendation: It is fair to say that it is a bug for something to turn out to be `instanceof FabricInstance` but not `instanceof BaseFabricInstance`. (Likewise `FabricPrimitive` and `BaseFabricPrimitive`.) I think it'd be reasonable to add a `static` method `BaseFabricInstance.isInstance(value: unknown): value is BaseFabricInstance` which (a) does the obvious check and (b) also `throw`s ("Shouldn't happen: ...") if it ever encounters `(v instanceof FabricInstance) && !(v instanceof BaseFabricInstance)`. (Likewise `BaseFabricPrimitive`.)

packages/data-model/src/deep-freeze.ts
```diff
@@ -234,12 +233,12 @@ export function deepFreeze<T>(value: T): T {
...
     inProgress.add(obj);
 
-    // Arm 3: a `FabricInstance` freezes itself in place via its `[DEEP_FREEZE]`
+    // Arm 3: a fabric instance freezes itself in place via its `[DEEP_FREEZE]`
     // protocol member. `freeze` is handed in as the `subFreeze` callback: it
     // closes over `inProgress`, so the impl's recursion into nested
     // `FabricValue`s shares cycle state with this call -- the participating
     // instance doesn't need to be `inProgress`-aware in its own signature.
-    if (value instanceof FabricInstance) {
+    if (value instanceof BaseFabricInstance) {
```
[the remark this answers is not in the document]
danfuzz: Thanks — good catch, and confirmed as a real (latent) invariant gap: after the gate flip, a hypothetical direct `FabricInstance` subclass that bypassed `BaseFabricInstance` would skip its freeze protocol and be cached as deep-frozen while only shallow-frozen. No such subclass exists today, but the risk was real.

Addressed in `46d451d6d` by making the invariant **explicit and enforced** rather than relying on it silently: a new static `BaseFabricInstance.isInstance(value): value is BaseFabricInstance` (and its `BaseFabricPrimitive` counterpart) does the `instanceof` check and additionally `throw`s "Shouldn't happen: …" if it ever encounters a value that is a `FabricInstance` (resp. `FabricPrimitive`) but not the corresponding base. The `deep-freeze.ts` dispatch sites now gate through these guards instead of a bare `instanceof`, so the exact condition you flagged can no longer silently mis-cache — it fails loud at every freeze and deep-frozen-check dispatch (instance and primitive paths alike). Unit tests for the throw-guard are included.

— upstream-liaison-bolt (Claude Opus 4.8)

packages/data-model/src/deep-freeze.ts
danfuzz: Thanks — implemented as specified in `46d451d6d` (current tip `37c548901`). Added a static `BaseFabricInstance.isInstance(value): value is BaseFabricInstance` and its `BaseFabricPrimitive.isInstance()` counterpart, each doing the `instanceof` check plus a "Shouldn't happen: …" throw on the `instanceof Fabric* && !instanceof Base*` mismatch, and routed all six freeze-dispatch sites (three instance, three primitive) through them so the invariant is enforced at every gate rather than silently skipped.

One wrinkle worth noting: `BaseFabricPrimitive`'s instance type is structurally empty (no members), so as a user-defined type guard its *negative* narrowing collapses the value to `never` for any following branch — a bare `instanceof` doesn't, since it narrows nominally. That bit exactly one site, `isDeepFrozenFabricValue`'s value walk, where the primitive arm preceded the array / plain-object arms. Resolved by ordering the `BaseFabricPrimitive` arm last there (its negative then only feeds the final `else`, which doesn't use the value) — semantically identical, since a `FabricPrimitive` is neither an array nor a plain object. `BaseFabricInstance.isInstance()` has members, so it has no such issue.

— coder-cedar (Claude Opus 4.8)

danfuzz:
**Reviewer: APPROVE — delta re-review** (reviewed at head `37c548901`)

Follows my earlier APPROVE of the base move at `b29ccef7a`. This re-review covers the folded-in `isInstance()` invariant-guard delta (Dan + cubic): a static `BaseFabricInstance.isInstance()` / `BaseFabricPrimitive.isInstance()` that turns the previously-latent "a class extending `FabricInstance`/`FabricPrimitive` directly would be silently skipped by the narrowed gate" risk into an actively-enforced invariant. (Head advanced past the originally-relayed `46d451d6d` by one doc-only commit, `37c548901`, which adds the "death before confusion" JSDoc; code is identical.)

What I checked in the delta:

1. **Guard logic correct + well-documented.** Both guards return `true` for a genuine `Base*` instance, `throw` "Shouldn't happen: …" only on the real `instanceof Fabric* && !instanceof Base*` mismatch, and return `false` for everything else (non-fabric values never throw; the sibling `FabricPrimitive`/`FabricInstance` split means no cross-throw). The throw is documented as deliberate "death before confusion" invariant-enforcement, naming the exact hazard it prevents (a broken subclass being cached as deep-frozen while only shallow-frozen).
2. **All six gate sites rewired.** The three `BaseFabricInstance` sites and all three `FabricPrimitive` sites in `deep-freeze.ts` now route through the guards (`BaseFabricInstance.isInstance()` / `BaseFabricPrimitive.isInstance()`), replacing the bare `instanceof`.
3. **The `isDeepFrozenFabricValue` arm reorder is behavior-preserving.** The primitive arm moved from first to last (after the array / plain-object arms). The arms are mutually exclusive on runtime type — a `BaseFabricPrimitive` is never an array, plain object, or `BaseFabricInstance`, and vice versa — so ordering is irrelevant for every legitimate input; routing is unchanged. The reorder is required because `BaseFabricPrimitive`'s instance type is structurally empty, so the user-defined guard's *negative* branch collapses `item` to `never` for any following branch (a TS narrowing footgun `instanceof` doesn't have); placing it last confines that to the final `else`, which doesn't use `item`. The in-place comment explains this accurately. (`BaseFabricInstance.isInstance` has real members, so no such issue — it stays first.)
4. **No import cycle.** `deep-freeze.ts` now imports `BaseFabricPrimitive`; both `BaseFabricInstance.ts` and `BaseFabricPrimitive.ts` import only from the `interface.ts` leaf, so the graph stays acyclic.
5. **Tests.** New `isInstance()` suites on both bases cover true / false-for-non-fabric / throw-on-rogue-direct-subclass (with clearly-labeled rogue fixtures), plus a new `BaseFabricPrimitive.test.ts`. The `deep-freeze` regression + cycle tests still pass.
6. **Scope + hygiene.** Delta touches only the two base classes, `deep-freeze.ts`, and the two test files. No `[CODEC]` touch; comments are project-technical; `Co-Authored-By` trailer present.

The one intentional behavior change is the throw on a rogue direct subclass (previously a silent `false`) — that is the hardening this delta exists to add, not a regression; all legitimate inputs are unaffected.

Independent confirmation: ran `deno check` on the affected source (clean — proves no residual `never`-poisoning at any site) and `deno test` on the two guard suites + `deep-freeze.test.ts` (3 passed / 88 steps / 0 failed).

This APPROVE is contingent on the head-`37c548901` CI run going green, after which the PR queues for Dan's merge.

danfuzz:
**Reviewer: APPROVE — seed-member delta re-review** (reviewed at head `154ddc950`)

Third and (I expect) final re-review, covering Dan's root-cause fix for the primitive guard: rather than ordering the `isDeepFrozenFabricValue` chain around the empty-type `never`-collapse, `BaseFabricPrimitive` is made non-empty via a symbol-keyed placeholder seed member, so `BaseFabricPrimitive.isInstance()` narrows as an ordinary `value is` guard and the reorder workaround is removed. Builds on my prior APPROVEs of the base move (`b29ccef7a`) and the `isInstance()` guards (`37c548901`); the six dispatch gates and the guard bodies are unchanged from `37c548901`.

What I checked in this delta:

1. **The throwing seed member has no callers (the load-bearing check).** `[EXAMPLE_METHOD]` is a fresh `unique symbol` (`Symbol.for("data-model.exampleMethod")`) with a concrete `(): never` body that always throws. Since every concrete primitive inherits it, it is safe only if nothing ever invokes it. Grep-confirmed tree-wide: the symbol has exactly one declaration, one method definition, and one reference — the test that asserts it throws. **No dispatch or call site anywhere in production code.** It's also a prototype method (not an enumerable own-property), so it's inert with respect to the freeze machinery's `Object.values` walks — genuinely zero runtime effect today.
2. **Cleanly documented placeholder.** Both the symbol and the method carry JSDoc framing it as a deliberate seed (mirroring the `[DEEP_FREEZE]`/`[IS_DEEP_FROZEN]` plumbing pattern), stating its only current purpose is to make the instance type non-empty for narrowing, that it has no callers/behavior yet, and "replace it with a real member once one is identified." Not dead-code-confusing.
3. **Workaround removed cleanly; restored order is behavior-preserving.** The primitive arm returns to first in `isDeepFrozenFabricValue` (`BaseFabricPrimitive` → `BaseFabricInstance` → array → plain-object → else), with a plain two-line comment; the entire obsolete "ordered-last / would collapse to `never`" comment block is gone (grep-clean). The arms remain mutually exclusive on runtime type, so first-vs-last placement routes every legitimate input identically — this matches the pre-workaround order I validated on the base PR.
4. **`isInstance()` narrows normally now.** With the non-empty type, primitive-first no longer poisons the following branches. `deno check` on `deep-freeze.ts` + `BaseFabricPrimitive.ts` is clean — no residual `never`-collapse at any site.
5. **Test.** A `ProbePrimitive`-based case asserts `[EXAMPLE_METHOD]()` throws.

Optional (non-blocking): the seed symbol uses the `"data-model."` registry prefix where the existing plumbing symbols use `"common."` (`common.deepFreeze` / `common.isDeepFrozen`). Immaterial while it's an uncalled placeholder, but worth aligning if/when it's replaced by a real member — take it or leave it.

Independent confirmation: `deno check` on the affected source clean; `deno test` on `BaseFabricPrimitive.test.ts` + `deep-freeze.test.ts` = 2 passed / 77 steps / 0 failed (new stub-throws test + deep-freeze regression/cycle pins).

This APPROVE is contingent on the head-`154ddc950` CI run going green, after which the PR queues for Dan's merge.

packages/data-model/src/fabric-instances/BaseFabricInstance.ts
```diff
@@ -46,6 +46,41 @@ export const IS_DEEP_FROZEN: unique symbol = Symbol.for(
  * point individual subclasses stop implementing `deepClone()` directly.
  */
 export abstract class BaseFabricInstance extends FabricInstance {
+  //
```
danfuzz: Usual conventions for `data-model` indicate static members come after instance members.

danfuzz: Done — the static `isInstance()` methods now sit after the instance members in both base classes, per the data-model convention. Pure reorder, no behavior change. Landed in `36682cf4b`.

— upstream-liaison-bolt (Claude Opus 4.8)

danfuzz:
**Reviewer: APPROVE — member-reorder delta** (reviewed at head `36682cf4b`)

Covers the code-conventions fix: the static `isInstance()` methods move below the instance members in `BaseFabricInstance.ts` and `BaseFabricPrimitive.ts` (data-model "statics after instance members" convention). Builds on my prior APPROVEs (`b29ccef7a` base move, `37c548901` guards, `154ddc950` seed member).

- **Verified pure move, no content change.** The removed and added line multisets across both files are identical (`+52/-52`), so nothing but member position changed — the guard bodies, seed member, and JSDoc are byte-identical. Class member order has no runtime effect, so this is behavior-preserving by construction.
- **Ordering is now correct.** In both files the `Instance members` section precedes `Static members`, with `isInstance()` the only static member; no static-before-instance ordering remains.
- Independent `deno check` on the two base files + `deep-freeze.ts` is clean.

No behavior to re-verify beyond the above. (`[EXAMPLE_METHOD]` name/prefix remains a separate pending decision; my earlier optional prefix note stands and does not gate.)

This APPROVE is contingent on the head-`36682cf4b` CI run going green. (Note: the prior run's single red — `Pattern Integration Tests (2/4)` — was an infra/coverage-artifact flake: the test suite passed `10/38/0` and the job failed only in the post-test `deno coverage` step reading a missing raw coverage JSON. Unrelated to this change; expected to clear on the fresh run.)

PR #4835, opened 2026-07-19
fix(patterns): migrate default-app and notebook read-then-push appends

bfollington:
## Why

Every cold compile of the space root logs five `mergeable-push:read-then-push` transformer warnings to the browser console (reported from a production mobile session on Estuary — they fire on every load, on any deployment, since the flagged code ships at HEAD). The five appends are dedup-guarded pushes of piece cells, or a snapshot-derived append, in `system/default-app.tsx` and `notes/notebook.tsx`. Beyond the console noise, the shape they warn about is real: the whole-list read keeps each append in the conflict set, so writes contend and retry instead of merging, and concurrent adds of the same piece rely on conflict-retry rather than identity to avoid duplicates.

#4736 migrated this class in `record`, `map-demo`, and the JSX reactivity demo and wrote `docs/development/migrating-collection-writes.md`; this PR applies that guide to the two system-surface patterns it didn't cover.

## What Changed

- **`default-app.tsx` — `dropOntoNotebook`, `addPiece`:** replace guard+push with `addUnique` on the piece cell, which dedups by link on the server, so re-adds are no-ops and disjoint adds merge — no whole-list read. `addPiece`'s event field is now declared `Writable<MentionablePiece>` so it arrives as a cell: the shell's `registerNavigatedPiece` already sends a piece cell, and with the old plain event type it would arrive as a query-result proxy, which `addUnique` silently fails to dedup (this handler is the worked example in the migration guide's "Event data is a cell only if its type says so").
- **`notebook.tsx` — `handleDropOntoNotebook`:** drops its whole-list read of the target notebook entirely; both branches `addUnique` into the target, so concurrent drops onto the same notebook merge. Deliberate behavior change: dropping a note the target already holds still moves it out of the current notebook — the old guard early-returned, leaving the note in **both** notebooks.
- **`notebook.tsx` — `handleDropOntoCurrentNotebook`:** `addUnique` for both branches, keeping the list read (it resolves the dragged index against the selection) with comments per the guide's "keep it deliberately and say so".
- **`notebook.tsx` — `createNotebookFromPrompt`:** the add branch appends a notebook whose contents derive from the notes snapshot — a content-dependent append — so it becomes a read-modify-write `set` over that snapshot, mirroring its own move branch.
- **`shell/public/index.html`:** add the non-Apple `<meta name="mobile-web-app-capable">` beside the deprecated Apple-only tag Chrome warns about.

## Test plan

- `cf check` on the **pre-change** files reproduces exactly the five warnings (2 in default-app, 3 in notebook); on the migrated files both patterns compile with **zero** transformer diagnostics.
- `cf test` on `notes/notebook.test.tsx` (31 assertions) and `system/` (34 assertions) passes.
- `deno task integration patterns default-app` (browser-driven) passes, including the rapid notebook-note creation/reload regression — exercises `addPiece` end-to-end through `registerNavigatedPiece`, the surface CT-1838 bricked.
- Full workspace `deno task test` passes.

---
## Summary by cubic
Removes read-then-push appends in `system/default-app` and `notes/notebook` to eliminate five `mergeable-push:read-then-push` warnings and let concurrent adds merge by identity. Also fixes same-notebook drop edge cases and adds a mobile-web-app meta tag to clear a Chrome console warning.

- **Bug Fixes**
  - Default app: `dropOntoNotebook` and `addPiece` now use `addUnique`; `addPiece` event is `Writable<MentionablePiece>`; `PiecesListOutput` now declares `allPieces` and `addPiece` so callers/tests can access them.
  - Notebook: both drop handlers use `addUnique`; when target and current alias, skip removal to avoid reorder/data-loss on in-notebook drops; cross-notebook drops still move already-present notes; `createNotebookFromPrompt` add path becomes a snapshot `set`. Tests added to pin these semantics.
  - Shell: add `<meta name="mobile-web-app-capable" content="yes">`.

<sup>Written for commit 935d67eae30564d0494c5e94417208bd88c3437a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4835?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/patterns/notes/notebook-drop.test.tsx
+++ b/packages/patterns/notes/notebook-drop.test.tsx
@@ -1,0 +1,159 @@
+/**
+ * Test: dropping onto a note row keeps memberships stable.
+ *
+ * Every row's drop zone binds handleDropOntoNotebook with the notebook's own
+ * SELF as the target, so target and current alias the same notes collection.
+ * The handler adds with addUnique (dedup by link identity, no whole-list
+ * read of the target) and must skip the remove-from-current step when the
+ * lists alias — otherwise an in-notebook drop of an existing note turns into
+ * a move-to-tail reorder, and a multi-select drop strips the selected notes
+ * from the notebook entirely (the pre-addUnique code had exactly that
+ * data-loss bug). These tests drive the drop stream bound to a row and pin:
+ * order stability for an in-notebook re-drop, membership survival for a
+ * multi-select re-drop, and a plain add for a note from outside.
+ *
+ * Run: deno task cf test packages/patterns/notes/notebook-drop.test.tsx --root packages/patterns --verbose
+ */
+import { action, computed, pattern, UI } from "commonfabric";
+import { findNode, propsOf, readValue } from "../test/vnode-helpers.ts";
+import Notebook from "./notebook.tsx";
+import Note from "./note.tsx";
+
+type DropStream = {
+  send: (event: { detail: { sourceCell: unknown } }) => void;
+};
+
+// A row's title lives in its cf-chip's `label` prop, which textContent
+// (a children walk) never sees — so locate the row zone as the
+// note,notebook-accepting cf-drop-zone whose subtree holds the chip with
+// that label. Module scope: SES callbacks may not capture callables from
+// enclosing function scopes.
+const isElement = (
+  value: unknown,
+  name: string,
+): value is Record<PropertyKey, unknown> =>
+  typeof value === "object" && value !== null &&
+  (value as { name?: unknown }).name === name;
+
+const sendDropOntoRow = (
+  subject: { [UI]: unknown },
+  rowText: string,
+  sourceCell: unknown,
+) => {
+  const zone = findNode(subject[UI], (node) => {
+    if (!isElement(readValue(node), "cf-drop-zone")) return false;
+    // Prop values arrive as link proxies, so resolve before comparing.
+    if (
+      String(readValue(propsOf(node)?.accept) ?? "") !== "note,notebook"
+    ) return false;
+    // Labels carry the note NAME ("📝 <title>"), so match by inclusion.
+    const chip = findNode(
+      node,
+      (inner) =>
+        isElement(readValue(inner), "cf-chip") &&
+        String(readValue(propsOf(inner)?.label) ?? "").includes(rowText),
+    );
+    return chip !== undefined;
+  });
+  const stream = propsOf(zone)?.["oncf-drop"];
+  if (typeof stream === "object" && stream !== null && "send" in stream) {
+    (stream as DropStream).send({ detail: { sourceCell } });
+  }
+};
+
+const noteTitlesOf = (subject: { notes?: unknown }): string[] =>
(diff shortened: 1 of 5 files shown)


bfollington (commented):

PR #4835 · detail mode (3 files) · verified at `97b14a0` with `cf check` on both changed patterns, emitted-schema inspection, notebook pattern tests (31/31), `deno fmt --check`, `deno lint`, and the completed green CI matrix.

🟡 **Improvement — add regression coverage for the changed identity and move semantics**

`packages/patterns/system/default-app.tsx:122` and `packages/patterns/notes/notebook.tsx:169`: the implementation matches the collection-write migration guide, and the transformed `addPiece` payload correctly carries `piece` as an `asCell` reference. However, the existing default-app integration only proves a first registration appears, while `notebook.test.tsx` never invokes the changed drag/drop or prompt handlers. That leaves the two easiest silent regressions unguarded: sending the same piece cell twice must leave one `allPieces` entry, and dropping a note onto a target that already contains it must remove it from the current notebook while leaving one target membership. Please add focused assertions for those behaviors; this is non-blocking for this PR, but it will pin the exact reasons for the migration. **Verified** by tracing the existing test paths.

**Verdict:** no blocking issues; mergeable with the coverage improvement above.

bfollington:
Actioned the review in 935d67eae — and writing the requested coverage surfaced a real defect in the first commit, so this ended up being more than tests.

**The defect:** `handleDropOntoNotebook`'s only binding passes the notebook's own SELF as the target, so target and current alias the same notes collection. With the whole-list guard gone, the single-item branch's remove-from-current turned an in-notebook drop of an existing note (a sloppy same-notebook drag) into a move-to-tail reorder — the old guard had made it a no-op. Fixed by skipping the removal when target and current alias, which also fixes a pre-existing data-loss bug: the multi-item branch unconditionally stripped the selection from "current" even when current WAS the target, silently dropping the selected notes from the notebook.

**Coverage, both verified to bite:**
- `notebook-drop.test.tsx` drives the drop stream bound to a row's `cf-drop-zone`: order stability for an in-notebook re-drop, membership survival for a multi-select re-drop, plain add + isHidden for an outside note. With the aliasing guard disabled, 4 of 6 assertions fail.
- `default-app.test.tsx` pins addPiece idempotence by identity: same piece cell twice → one `allPieces` entry; distinct piece still lands. With the event field reverted to a plain (proxy) type, the duplicate lands and the test fails.

This needed `allPieces`/`addPiece` declared in `PiecesListOutput` (the runtime surfaces outputs through the declared schema; the index signature alone left `subject.addPiece` undefined to external readers) — same fix #4736 made for record's `addModule`.

Full pass after the change: cf check zero diagnostics on both patterns, the three pattern suites (41 assertions), and `deno task integration patterns default-app` all green.

PR #4837, opened 2026-07-20
feat(runner): accept encoded-`FabricValue` payloads in `data:` URI decode

danfuzz:
Stacked on #4839 (reader unification); retarget to `main` once that merges.

## What

`decodeDataURIPayloadText()` — the single payload decoder shared by both
`data:` URI readers (`getJSONFromDataURI()` and attestation `load()`, unified
in #4839) — now routes its input by form:

- Text bearing the `fvj1:` tag is decoded as the standard `data-model`
  `FabricValue` JSON-embedded encoding, via `valueFromJson()` with a
  decode-framed `EmptyReconstructionContext` (following the
  `decodeMemoryBoundary()` precedent in `memory/v2.ts` — links at this
  boundary are sigil/plain data, so no cell reconstruction is ever needed).
- Any other text continues through `JSON.parse()` exactly as before. This
  branch stays indefinitely: a `data:` URI _is_ its own content, and such URIs
  are embedded in persisted documents, so already-minted bare-JSON ids
  survive.

Because the routing lives in the one shared helper, both readers learn the
new form in a single place. The attestation error message widens from
"parse JSON" to "decode payload" accordingly. The matched-set doc comments
(`createDataCellURI()` / `getJSONFromDataURI()`) are updated; the
decode-side `TODO(danfuzz)` from #4762 is discharged, the encode-side one
remains.

## Why

First half of the planned fix for the `data:` URI boundary's non-conforming
encoding (see the investigation notes landed in #4762, and the `TODO` in
`schema-hash.ts` explaining how #4360's canonicalization workaround traces
back to it). Rollout is the usual safe sequence: readers learn the new form
first (this PR), then the writer switches over (#4838), then the #4360
workaround becomes removable.

No behavior change for any URI that exists today: nothing in the system mints
`fvj1:`-tagged payloads yet.

## Tests

New `getJSONFromDataURI` block in `data-uri.test.ts`: both payload forms,
percent-encoded and Base64 spellings, non-ASCII text, non-finite numbers and
`-0`, deep-frozen results, `/`-prefixed-key round-trip (the spec §5.6
quote-escaping path that sigil links will ride once the encoder switches),
plus header validation. `attestation-data-uri.test.ts` (from #4839) gains
the `fvj1:` cases for `load()`. The fvj1-branch tests verifiably fail
against the unpatched decoder.

Runner suite and full workspace suite green.

https://claude.ai/code/session_01FkQwRxCmaYNC7tQ3E949is


danfuzz:
Follow-up commit `31514ca76`: found a second, independent reader of `data:` URI payloads — `storage/transaction/attestation.ts`'s `load()`, with its own `JSON.parse()`. Since this PR's whole premise is "readers learn the new form before any writer mints it," that reader is now covered too: the payload-text routing moved into one exported helper (`decodeDataURIPayloadText()` in `uri-utils.ts`) used by both readers, and a new `attestation-data-uri.test.ts` pins both payload forms plus the error shapes. Verified these are the only two payload parsers (every other `data:` reference in the workspace is a prefix gate only). Runner suite green.

— Claude (via Dan's session)

danfuzz:
Restructured per Dan's request: the reader unification (shared `decodeDataURIPayloadText()` helper + attestation routing + its baseline tests) is now its own prep PR, #4839, and this PR is restacked on it. This PR's diff is now just the fvj1 decode routing itself (plus its tests and the attestation error-message widening). The chain is #4839 → #4837 → #4838.

— Claude (via Dan's session)

PR #4842, opened 2026-07-20
fix(cli): --log-level stops at "--"; piece link validation failures are data errors

mathpirate:
## Summary

Two follow-up burrs from the [#4771](https://github.com/commontoolsinc/labs/pull/4771) review, same philosophy as that PR
(no silent no-ops; data errors are not usage errors):

**`--log-level` extraction now stops at `--`.** `extractLogLevel` scanned the
whole argv, so a schema-derived handler flag literally named `--log-level`
(with a valid level as its value) after `--` was silently consumed — the exact
payload-eating hazard #4771 hardened `--no-color` against. It now mirrors
`extractNoColor`: scanning stops at the first `--`, everything after passes
through untouched. `cf --log-level error <cmd> -- --log-level error` sets both.

**`piece link` validation failures are data errors.** A `LinkValidationError`
(source/target piece or path that doesn't exist — a condition read over the
network) was wrapped in a Cliffy `ValidationError`, dumping the colorized
usage screen exactly like the `piece get` bug #4771 fixed. It now prints the
runtime's message (which carries its own `--allow-non-existing` next step)
plus an inspect TIP on stderr and exits 1. Genuine argument-shape errors
(missing target path, bad refs) still get the usage treatment.

**Shared exit path, unit-covered.** The get/link stderr-and-exit wiring is
deduplicated into one exported `exitWithDataError` with an injected-deps seam,
so both catch sites shrink to a call and the wiring itself is unit-tested
rather than reachable only through a real process exit. Measured
`packages/cli` coverage debt: **−5 lines vs base** (tasks/coverage-metrics.ts
over the full package suite; base 4167 → this branch 4162).

**Stacked on #4771**: branched from its current head (79c86ec43), to be
rebased onto `main` once it lands. Draft until then — only the last commit is
this PR's change.

## Review response (@willkelly)

- 🟡 coverage: adopted, two ways — the suggested integration case now runs in
  `run_piece_links` (invented-source link without `--allow-non-existing`
  asserts exit 1, empty stdout, message + `--allow-non-existing` + TIP on
  stderr, and no usage screen), and the action wiring became unit-coverable
  via the `exitWithDataError` seam, which is what actually feeds the debt gate
  (the CLI integration job collects no V8 coverage — it drives the compiled
  binary).
- docs `null` bullet: resolved by restacking onto 79c86ec43, which carries the
  #4771 fix; this PR's bullet extends the corrected wording.
- hint asymmetry (link always hints): kept as-is, per review.

## Tests

- `test/log-level.test.ts`: `--` boundary cases, including `--log-level`
  immediately before `--` not pairing across the boundary.
- `test/piece-call.test.ts`: `pieceLinkDataErrorReport` (message passthrough,
  inspect hint naming both pieces, null for non-validation errors) and
  `exitWithDataError` (message+hint order, hint omitted when absent, exit 1).
- `integration/integration.sh` (`piece-links` section): failure-shape
  assertions as above. Verified end-to-end against a local toolshed:
  full section green.
- Full `packages/cli` suite green; `deno fmt`/`lint`/`check` clean on touched
  files.

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -48,7 +48,9 @@ standard error and continues searching that piece and the rest of the space.
 - `piece call` accepts its payload as an inline JSON argument, `-` for stdin, an
   implicit pipe (no payload argument), or schema-derived flags after `--`.
 - A `piece get` path that doesn't resolve prints a one-line error on stderr and
-  exits 1 — it is a data error, not a usage error.
+  exits 1 — it is a data error, not a usage error. A `piece link` that fails
+  validation (a source/target piece or path that doesn't exist) reports the same
+  way.
 - The launcher spawns the child CLI with `deno run --quiet` so Deno's own
   warnings (npm "Ignored build scripts" banner) never reach users.
 
--- a/packages/cli/commands/piece.ts
+++ b/packages/cli/commands/piece.ts
@@ -188,6 +188,48 @@ export function pieceGetDataErrorReport(
   };
 }
 
+/**
+ * Build the stderr report for a `piece link` validation failure. Returns null
+ * when the error is not a LinkValidationError (the caller should rethrow).
+ * Link validation fails on data conditions — a source/target piece or path
+ * that doesn't exist, read over the network — so it reports like `piece get`'s
+ * unresolved-path data error rather than as a Cliffy usage error.
+ */
+export function pieceLinkDataErrorReport(
+  error: unknown,
+  opts: { sourcePieceId: string; targetPieceId: string },
+): { message: string; hint: string } | null {
+  if (!(error instanceof LinkValidationError)) return null;
+  return {
+    message: error.message,
+    hint: cliText(
+      `TIP: Run 'cf piece inspect --piece ${opts.sourcePieceId} ...' and '--piece ${opts.targetPieceId} ...' to see the fields each piece actually has.`,
+    ),
+  };
+}
+
+/**
+ * Print a data-error report — message plus optional hint — to stderr and exit
+ * 1. The single exit path for the `piece get` / `piece link` data errors
+ * above. The `deps` seam lets unit tests observe the wiring without a real
+ * process exit; runtime callers use the defaults.
+ */
+export function exitWithDataError(
+  report: { message: string; hint?: string },
+  deps?: {
+    printError?: (message: string) => void;
+    printHint?: (message: string) => void;
+    exit?: (code: number) => never;
+  },
+): never {
+  const printError = deps?.printError ?? console.error;
+  const printHint = deps?.printHint ?? hint;
+  const exit = deps?.exit ?? Deno.exit;
+  printError(report.message);
+  if (report.hint) printHint(report.hint);
+  return exit(1);
+}
+
 export function pieceCallRawArgs(
   tail: string[],
   literalArgs: string[],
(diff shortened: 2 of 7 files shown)


willkelly (commented):
_Review by Claude (Opus 4.8) on behalf of @willkelly._

**Verdict: solid follow-up — one coverage gap worth closing before CI runs.**

Reviewed the incremental commit (`407d53a30`) on top of #4771's head. Verified locally: `deno check` clean on the touched files, full `packages/cli` suite green (26 files / 0 failed), and both changes faithfully extend the patterns from #4771.

**What's solid**

- `extractLogLevel` stopping at `--` mirrors `extractNoColor` exactly, with a sharp edge-case test — a `--log-level` sitting immediately before `--` correctly does not pair its value across the boundary.
- `pieceLinkDataErrorReport` mirrors `pieceGetDataErrorReport`; genuine arg-shape errors from `parseLink` (thrown before the `try`) still get the Cliffy usage screen, so only the data-condition `LinkValidationError` is reclassified. `source`/`target` are in scope, and the runtime's `--allow-non-existing` next-step survives verbatim in the message.

**🟡 Coverage: the new `piece link` failure path is exercised by no test, and will likely trip the coverage-debt gate — the same one that flagged #4771.**

This commit adds ~3 uncovered lines in `packages/cli`, all in the new `link` catch block (the `pieceLinkDataErrorReport` call plus `console.error` / `hint` / `Deno.exit`). The pure helper is unit-tested, but the action wiring is not — and the existing `piece-links` integration test only covers the `--allow-non-existing` **success** path (`integration.sh:341`); nothing triggers a `LinkValidationError`. So neither unit nor integration coverage reaches these lines.

#4771 hit exactly this gate (+12) and cleared it by unit-testing the extractable logic and leaning on integration tests for the action bodies. Here the action body has no integration coverage, so the +3 will likely surface as a debt regression once CI can run (it can't yet — the branch conflicts with `main`).

Suggested fix, which is also the behavior this PR adds (so it's worth having regardless): in `packages/cli/integration/integration.sh`, after the existing link section, attempt a link whose target path doesn't exist **without** `--allow-non-existing`, and assert exit 1, a one-line message on stderr with no usage/help screen on stdout, and that the message carries the `--allow-non-existing` hint. That covers the catch on CI and locks in the fix.

**Non-blocking notes**

- The docs bullet still teaches `piece get` prints `null` for an absent value (`README.md:6` / `skills/cf/SKILL.md`). That's inherited from this PR's base; #4771 now corrects it, so it resolves on restack — no action here.
- `pieceLinkDataErrorReport` always returns a hint, whereas `pieceGetDataErrorReport` returns it conditionally (omitting it in `--input` mode). Fine as-is — a link failure always has a meaningful "inspect both pieces" suggestion.

mathpirate:
Thanks — both points incorporated at 2497decbe (restacked onto #4771's current head 79c86ec43, so the `null` docs bullet inherits the corrected wording, and CI can finally run — the stack is clean against main now).

- 🟡 **coverage**: your suggested integration case is in (`run_piece_links`: invented-source link without `--allow-non-existing` → exit 1, empty stdout, message + `--allow-non-existing` + TIP on stderr, no usage screen; verified green end-to-end against a local toolshed). Your +3 was exactly right — measured 4167 → 4170 pre-fix.
- One premise correction that changed the fix shape: the CLI integration job drives the **compiled binary** with no `DENO_COVERAGE_DIR`, so `integration.sh` feeds no V8 coverage into the debt gate — the integration case locks behavior, but it can't pay the +3. What pays it: the get/link stderr-and-exit wiring is deduplicated into one exported `exitWithDataError` with an injected-deps seam, unit-tested directly. That also covers `piece get`'s previously-dead catch lines, so the branch lands at **4162, −5 vs base** — no need to ride #4771's surplus.
- hint asymmetry: kept as-is, per your note.

PR #4844, opened 2026-07-20
feat(runner)!: retire legacy bare-JSON decoding of `data:` URI payloads

danfuzz:
Stacked on #4838; retarget as the chain merges. First of the agreed
follow-on set (A: this; B: `{value:X}` layering inversion; C: honest media
type; D: `schema-hash.ts` workaround removal).

## What

`decodeDataURIPayloadText()` now requires the standard `fvj1:`-tagged
`FabricValue` encoding; a bare-JSON payload fails loudly with the codec's
own "not a JSON-encoded `FabricValue`" error. Per the persistence
investigation (see #4837/#4838 history): nothing durably stored carries
bare-JSON ids — the write boundary has inlined `data:` links since #1914
and all deployment databases postdate that via flag-day rebuilds — so the
only constituency this breaks is externally-minted bare-JSON URIs, and the
policy answer is that external minters must use the standard encoding.

Two things surfaced during removal:

- **A third minter.** `Runtime.getImmutableCell()` hand-rolled
  `JSON.stringify({value: data})`, silently relying on `JSON.stringify`'s
  `toJSON()` hook to serialize `Cell`s inside the data (e.g. wish candidate
  lists) into sigil links. It now encodes via `fabricFromNativeValue()` +
  `jsonFromValue()`, which performs the same conversion explicitly. (It
  survived earlier sweeps because both greps on the dev machine silently
  treat `runtime.ts`/`wish.ts` as binary; sweeps now use `grep -a`.)
- **Fixture archaeology.** The notebook-test and shopping-list-test
  traverse-replay fixtures carried 134 historical bare-JSON ids in their
  captured invocation/link sections; they are re-minted in the standard
  form with payload content unchanged, and goldens regenerated.
  `piece-query-legacy` and `toolshed-reload` needed nothing.

## Tests

Bare-JSON acceptance tests flip to rejection tests (helper, reader, and
attestation levels); two hand-crafted test payloads now mint via
`jsonFromValue()`. Runner suite green (969); full workspace suite green.

https://claude.ai/code/session_01FkQwRxCmaYNC7tQ3E949is

---
## Summary by cubic
Retire legacy bare-JSON decoding for `data:` URIs; only `fvj1:`-tagged `FabricValue` is accepted. All readers now reject bare-JSON payloads, and writers mint with the standard encoder.

- **Migration**
  - External minters must encode with `fabricFromNativeValue()` + `jsonFromValue()` from `@commonfabric/data-model/fabric-value` and `@commonfabric/data-model/codec-json`.
  - Replace any `JSON.stringify({ value })` minting with the explicit `fvj1` encoding.
  - Bare-JSON `data:` URIs now throw on decode via `decodeDataURIPayloadText()` and `getJSONFromDataURI()`; attestation `load()` reports an `InvalidDataURIError`.

- **Refactors**
  - `decodeDataURIPayloadText()` now only decodes `fvj1:` via `valueFromJson()`; empty payloads are rejected.
  - `Runtime.getImmutableCell()` mints via `fabricFromNativeValue()` + `jsonFromValue()`.
  - Tests flipped to reject bare JSON; traverse-replay fixtures re-minted in the standard form and goldens regenerated.
  - Generated-patterns: NaN payloads now reach handlers and are sanitized by the pattern; scenario expects `[4, 1, 6, 0, 0]` with `totalItems` 10.
  - Docs: decode contract clarified; added `getImmutableCell()` contract docs (arbitrary input, fabric conversion).

<sup>Written for commit 67b1c820fe5db1184b5b2e5e8d294aef84c4d881. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4844?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


danfuzz:
One CI finding worth reviewer attention, now fixed in `1f5cc41d4`: this PR carries a real (and intended-class) **behavior change in event transport**. `getImmutableCell()`'s old `JSON.stringify` silently turned `NaN` in event payloads into `null` in transit, and schema validation then rejected the `null`, so handlers never saw such events. With the standard encoding, NaN reaches handlers faithfully (consistent with the weird-numbers support). Exactly one test in the tree depended on the old lossy transport — the generated-patterns nested-computed-totals scenario, whose `// fails` step now exercises the pattern's own `sanitizeNumber` guard instead (its designed behavior for non-finite input). Expectations updated; sums unchanged.

Found via a disciplined bisect after CI flagged an off-by-one-element assertion: encoding form, canonical key order, and frozen decode results were each individually exonerated; the delta was purely `fabricFromNativeValue`'s faithful content vs `JSON.stringify`'s lossy normalization.

— Claude (via Dan's session)

PR #4870, opened 2026-07-21
fix(runner): preserve WriteAuthorizedBy authority across setsrc updates

seefeldb:
## Summary

- teach `piece setsrc` to compare the previous and replacement recursive module closures by normalized full filename
- persist cumulative successor-to-predecessor module delegations in source and compiled cache documents, merging metadata with `editWithRetry` when a content-addressed successor already exists
- register loaded delegations in the runtime and snapshot their transitive closure for CFC checks
- allow `WriteAuthorizedBy` to follow a trusted module delegation while retaining an exact binding-path check; resolver-specific source-file spelling remains diagnostic at verification
- document the temporary setsrc delegation contract and add focused chain, shared-module, authorization, and cold-reload regressions

## Why

`WriteAuthorizedBy` records a writer's content-addressed module identity. Updating a pattern changes that identity, so the replacement handler could no longer update data labeled by its predecessor. Content-addressed modules may also be shared across patterns, which means the same successor document can need delegations from multiple predecessor closures.

This change makes an explicit `setsrc` update bless that chain. Renamed, moved, or ambiguously matched modules still fail closed.

## User impact

Patterns updated through `cf piece setsrc` retain their existing `WriteAuthorizedBy` authority across module revisions, including after a cold reload and when multiple patterns converge on the same successor module.

## Validation

- `deno task check`
- `deno task check-docs` (519 code blocks)
- `deno lint` and `deno fmt --check` on changed TypeScript and docs
- focused and adjacent Piece, compilation-cache, identity, and CFC suites (364 test steps)
- post-merge regressions:
  - `deno test --no-check --allow-all packages/runner/test/module-delegation.test.ts packages/piece/test/setsrc-module-delegation.test.ts`

---
## Summary by cubic
Teach `piece setsrc` to preserve `WriteAuthorizedBy` authority by delegating writer identity when old and new modules share the same canonical authored filename, with compiler-authenticated metadata stored in source/compiled caches and registered in the runtime. Delegations are space-scoped and snapshotted per transaction; renames or ambiguous matches do not inherit authority.

- **New Features**
  - Add per-module `delegatedModuleIdentities` to source and compiled docs; derive by canonical full filename, skip ambiguous names, merge the authenticated union into both sets in one write, register it in the runtime, and snapshot it per transaction/digest.
  - Allow `writeAuthorizedBy` to match the live module hash or a delegated predecessor; the binding path must still match exactly, and file spelling is diagnostic at verification.
  - Thread `previousEntryIdentity` through `compileProgram` → `compileAndSavePattern` → `patternManager.compilePattern` to derive delegations during `setsrc`.
  - Scope delegations by space: bind entries to the attesting space, include them in prepared-digest canonicalization, and ignore metadata loaded from other spaces.

- **Bug Fixes**
  - Keep ordinary source-cache writes self-verifying by stamping compiler integrity only when delegation metadata exists; loaders ignore unstamped data.
  - Do not reuse cached closures missing required delegation metadata; persistence checks require delegations in both source and compiled sets. Preserve merged authority when multiple patterns converge on the same successor and across cold/warm loads and runtime-version recompiles.
  - Guard closure replication and cross-space mixing: skip same-space/keyless replication; reject origin/compiled closures that are incomplete or linked to another space’s attestation; require compiled coverage spans.

<sup>Written for commit 5f5f1d4049f03fb8e41e78ee97f110c56c0787a8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4870?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/cfc/writer-claim-correspondence.ts
+++ b/packages/runner/src/cfc/writer-claim-correspondence.ts
@@ -25,8 +25,10 @@
  * reconcile-adoption edge, a hostile verified module whose forged path
  * differs from a stored unstamped claim's by one leading segment is accepted
  * where before it needed the exact spelling — a marginal widening of the
- * pre-existing path-forgeability that mint-time identity binding /
- * authenticated `piece setsrc` delegation now closes for legitimate updates.
+ * pre-existing path-forgeability that #4871's mint-time identity binding closes
+ * at the source for newly engine-minted claims (aged unstamped claims remain
+ * compatibility state), with authenticated `piece setsrc` delegation carrying
+ * authority across legitimate updates (this PR).
  *
  * The correspondence is deliberately no wider than the divergence the
  * toolchain actually produced: equal after slash-normalization, or exactly
--- a/packages/runner/src/compilation-cache/cell-cache.ts
+++ b/packages/runner/src/compilation-cache/cell-cache.ts
@@ -818,7 +818,7 @@ export function writeSourceDocs(
  * already-loaded linked cells synchronously — no per-cell `sync()`. Returns the
  * raw documents keyed by their **stored** identity (verify with
  * {@link verifySourceDocs} before trusting). Resolves to `undefined` if the
- * entry document is absent.
+ * entry document is absent or any cache import link crosses out of `space`.
  */
 export async function loadSourceClosure(
   runtime: Runtime,
@@ -851,7 +851,12 @@ export async function loadSourceClosure(
       // `link` is already a loaded Cell (the entry sync pulled it). View it
       // under the recursive schema so its own links resolve as cells too.
       if (!isCell(imp.link)) continue;
-      const childCell = (imp.link as Cell<unknown>).asSchema(
+      const linkedCell = imp.link as Cell<unknown>;
+      // Cache attestations are space-local. A cross-space child may be validly
+      // compiler-stamped in its own space, but must never be flattened into
+      // the requested space's verified closure (and delegation registry).
+      if (linkedCell.space !== space) return undefined;
+      const childCell = linkedCell.asSchema(
         SOURCE_DOC_SCHEMA,
       );
       const child = childCell.get() as StoredSourceDoc | undefined;
@@ -935,7 +940,10 @@ export async function loadVerifiedSourceClosure(
     ]);
     return undefined;
   }
-  runtime.registerModuleDelegations(moduleDelegationsFromDocs(closure));
+  runtime.registerModuleDelegations(
+    space,
+    moduleDelegationsFromDocs(closure),
+  );
   return closure;
 }
 
@@ -1341,10 +1349,11 @@ export function writeSourceAndCompiledDocs(
  * (never re-deriving a cell from a stored `identity` field), and a cell's stored
  * doc is used only after its own persisted CFC label is confirmed to carry the
  * compiler integrity atom. So every document in the result — and every import
- * edge — came from an integrity-stamped cell that the parent's sigil link
- * actually points at; an unstamped/tampered child is dropped along with the edge
- * to it (treated as a cache miss, so the caller recompiles). The entry is the
- * one cell looked up by key, from the caller's trusted `entryIdentity`.
+ * edge — came from an integrity-stamped cell in `space` that the parent's sigil
+ * link actually points at; an unstamped/tampered child is dropped along with
+ * the edge to it (treated as a cache miss, so the caller recompiles), while a
+ * mixed-space link rejects the whole closure. The entry is the one cell looked
+ * up by key, from the caller's trusted `entryIdentity`.
  *
  * Resolves to the valid documents keyed by their stored identity (empty map if
  * the entry itself is missing/unstamped).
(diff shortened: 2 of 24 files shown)


packages/runner/src/compilation-cache/cell-cache.ts
```diff
@@ -706,6 +843,7 @@ export async function loadVerifiedSourceClosure(
     ]);
     return undefined;
   }
+  runtime.registerModuleDelegations(delegationMapFromDocs(closure));
```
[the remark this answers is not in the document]
seefeldb: Fixed in 035187614 and 3caee3886. Source delegation lists now carry compiler integrity on the delegation field; loading discards unstamped lists, and save-time merging retains only authenticated existing entries. A regression reconstructs an otherwise valid source document with forged delegation metadata and confirms that the alias is rejected. The follow-up scopes the attested schema to documents that actually carry delegation metadata, preserving ordinary source-cache writes.

packages/runner/src/pattern-manager.ts
```diff
@@ -1518,6 +1606,7 @@ export class PatternManager {
     modules: CacheableModule[],
     entryIdentity: string,
     opts: { runtimeVersion: string },
+    moduleDelegations: ModuleDelegationMap = new Map(),
```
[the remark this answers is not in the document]
seefeldb: Fixed in 035187614. The runtime-version miss path now rebuilds the delegation map from authenticated source documents and passes it through persistCompileCacheTracked. The end-to-end test forces a version-bump cold repair, then asserts that every repaired compiled document retains the authenticated predecessor list.

packages/runner/src/runtime.ts
```diff
@@ -648,6 +648,48 @@ export class Runtime {
...
   readonly #policyManifests = new Map<string, PolicyArtifactManifestV1>();
   readonly #policyManifestSpaces = new Map<string, Set<MemorySpace>>();
+  // Successor module identity -> direct predecessor identities whose writer
+  // authority it inherits. Entries come only from verified source closures or
+  // integrity-valid compiled closures. Transactions receive a transitive,
+  // immutable snapshot at creation so an in-flight authorization decision
+  // cannot change when another module loads.
+  readonly #moduleDelegations = new Map<string, Set<string>>();
+
+  registerModuleDelegations(
```
mathpirate, line 658: **[blocking-question]** The delegation registry is runtime-global and space-unscoped, which I think re-opens a cross-space door the field-level attestation just closed. The compiler atom attests "a compiler wrote this *in its space*" — in a space an attacker controls, their own runtime legitimately stamps arbitrary `delegatedModuleIdentities` (the honest setsrc derivation is only one way to produce the field; a direct compiler-builtin write in their own tx is another). When a victim's runtime loads that space's pattern — the normal shared-pattern / fabric-mount / stored-source cold-load flow — `loadVerifiedSourceClosure` registers those delegations here, globally and accumulate-only; every later tx in *any* space (including the victim's home) snapshots them, and `writeAuthorizedByReason` then honors "attacker-module delegates-from victim-module-identity" for writes against the victim's own claims, if the attacker's module runs as a live verified writer there (which running their shared pattern provides). Suggested shape: key registrations by the attesting space and have the tx snapshot consult only delegations attested in the same space as the doc under write — that preserves per-space setsrc chains exactly and closes the import of authority across the boundary. (For the record: I traced the field-level `cellCarriesIntegrity` checks at load and at write-merge and they hold — the forged-metadata hole from the earlier review round is genuinely closed at head; this is the one seam I could find past it.)

packages/runner/src/cfc/writer-claim-correspondence.ts
```diff
@@ -26,7 +26,7 @@
  * differs from a stored unstamped claim's by one leading segment is accepted
  * where before it needed the exact spelling — a marginal widening of the
  * pre-existing path-forgeability that mint-time identity binding /
- * setsrc-history delegation closes for real (board topic, index 38).
+ * authenticated `piece setsrc` delegation now closes for legitimate updates.
```
mathpirate: **[nit]** Small over-claim in the reworded sentence: the residual this paragraph documents is the *unstamped-claim capture* at reconcile-adoption, and setsrc delegation doesn't close that one — a stored claim that never got a stamp is still adoptable by a corresponding-spelled writer regardless of delegation. What closes it is mint-time identity binding (claims born stamped, so the capturable state stops being minted) — that's on `gideon/ct-1886-normalizer-unification` / #4871, which composes with this PR (born-stamped candidates make setsrc boundaries produce two-different-stamps at reconcile, which #4871's keep-stored change handles exactly as your updated `reconcileWriterClaimStamp` comment describes). Suggest keeping both halves: "…that mint-time identity binding closes at the source (#4871), with authenticated `piece setsrc` delegation carrying authority across legitimate updates (this PR)."

seefeldb: Fixed in 0a0460456. The comment now distinguishes #4871's mint-time identity binding for newly engine-minted claims from this PR's authenticated setsrc delegation for carrying legitimate update authority; aged unstamped claims are explicitly described as compatibility state.

packages/runner/src/runtime.ts
seefeldb: Fixed in 0a0460456. Delegations are now registered, closed transitively, snapshotted, and consulted per attesting/target space; the prepared digest binds that space. Cross-space closure replication omits origin delegation metadata, and source/compiled loaders reject mixed-space cache-link graphs so a child attestation cannot be rebased onto the root space. Regressions cover direct cross-space loads, split-chain isolation, mixed-space source and compiled links, replication denial, and same-space setsrc continuity.

Issue #4772, opened 2026-07-16
writeAuthorizedBy: writer-identity file-spelling shear across resolver name shapes bricks owner-protected fields (silent drop; profile bio-save; CT-1883's unfixed half)


mathpirate:
## TL;DR

`WriteAuthorizedBy` claims and live writer identities record the module's<br>**file spelling**, and the transformer's normalization spells the same file<br>differently depending on the resolver (absolute HTTP names get their first<br>path segment stripped; relative piece-deploy names don't). The<br>content-addressed `moduleIdentity` agrees across both compiles — but the<br>spelling mismatch (a) prevents claims from ever being **stamped** with that<br>`moduleIdentity`, and (b) unstamped claims are categorically rejected since<br>identity E5. Net effect: the first transaction that persists a field's claim<br>without a corresponding write (e.g. a sibling-field save) **permanently****<br>****bricks the field** — every later owner-write is dropped **silently**<br>(default-off logger). Today this makes Loom's profile **bio unsavable** in<br>any session after the creating one. Not a recent regression: identical at<br>labs 62048eee (07-10) and 2b4c4098 (07-15).

**Proposal:** anchor authorization on what is actually content-addressed —<br>`moduleIdentity` + `bindingPath` — and make the file-spelling comparison<br>tolerant at the three sites that today require exact equality:

1. `rebindWriteAuthorizedByClaims` (prepare.ts) — so claims get stamped when<br>the binding genuinely corresponds;
2. `reconcileWriterClaimStamp` (schema-merge.ts) — so a later legitimate<br>write **adopts a stamp onto an unstamped stored claim, self-healing****<br>****already-bricked stores**;
3. `writeAuthorizedByReason` (prepare.ts) — verification, with<br>`moduleIdentity` equality as the hard gate.

[...]

PR #4871, opened 2026-07-21
feat(ts-transformers,runner): explicit writer-identity spelling contract + claims born stamped (CT-1886 follow-up)

mathpirate:
## Why

Follow-up to #4852 (CT-1886), which anchored `writeAuthorizedBy` verification
on `moduleIdentity` + `bindingPath` and made the runner tolerant of the
resolver-dependent file spellings. This PR fixes the spelling divergence at
its source and closes the remaining soft spot #4852 documented: the
*unstamped* stored claim, whose first stamp is capturable by whoever writes
first (see the P1 discussion on #4852).

## What

Three commits, in dependency order:

**1. Unify the duplicated writer-identity normalizers** (pure refactor).
`normalizeSourceFilePath` (schema-generator) and `normalizeWriterIdentityFile`
(module-scope-function-hardening) were byte-identical copies feeding the two
halves of the claim↔provenance correspondence — a drift hazard the behavior
spec explicitly tracked ("must stay identical"). One shared
`src/utils/writer-identity-file.ts` now feeds all three call sites. Zero
behavior change.

**2. Spell writer-identity files from the caller's declared mapping, not a
first-segment guess.** The historical normalization stripped the first path
segment of any absolute file name, presuming the engine's per-load `/<id>`
module-path prefix. That presumption only holds for engine-driven compiles
(`pretransformProgramForModules` prefixes every module); direct compiles hand
resolver names straight through — absolute URL pathnames from
`HttpProgramResolver`, bare-relative piece-manifest keys — so the strip ate
real path segments (`/api/patterns/...` → `/patterns/...`), which is exactly
the shear behind #4772. The guess is replaced with an explicit contract:
`TransformerOptions.canonicalWriterIdentityFile` carries the caller's
compile-name → authored-name unmapping; the engine passes its existing
`storedFilenameFor` (exact `/<id>` strip + fabric-mount unmapping) at all
three pipeline sites; callers whose names are already authored paths pass
nothing and get verbatim spellings. **The land-gate this commit shipped with
("do not land before the tolerant compare") is satisfied — #4852 is merged.**
Aged stores keep their mint-time spellings; the runner's correspondence
tolerance is precisely the compat shim for them.

**3. Mint claims born stamped; treat two-stamp reconciliation as a version
boundary.** The engine computes `moduleIdentities` from pristine sources
*before* the TS compile, so the claim minter can look up its own module's
content-addressed identity and emit it in the marker:
`__ctWriterIdentityOf: { file, path, moduleIdentity }`. Claims are then born
stamped — the capturable unstamped state is no longer minted at all,
shrinking #4852's documented residual to aged data (which continues to heal
through reconcile-adoption). Two consequences handled:

- CT-1740's foreign-seed hazard evaporates for new mints: a seeded envelope's
  claims carry the *defining* module's identity by construction, regardless
  of who performs the seeding write.
- A republished module now re-presents its claims under a new
  `moduleIdentity` on every envelope write, so `reconcileWriterClaimStamp`'s
  both-stamped-different case changes from conflict-abort to **keep the
  stored stamp**: rotation still never happens, the new version's writes to
  the protected field stay fail-closed *at verification* (loudly, per #4852's
  drop reporting) pending the setsrc-history delegation design (board topic,
  index 38) — but the envelope's sibling writes keep committing instead of
  the whole transaction aborting on every pattern update.

## Verification

- ts-transformers: new tests pin the direct-absolute spelling (old
  mis-spelling asserted absent), engine-prefixed ≡ direct equivalence, the
  born stamp with `moduleIdentities` supplied, and unstamped compat without.
  Full suite + spec-sync green.
- runner: the #4852 correspondence tests all hold unchanged except the
  deliberate version-boundary pin (both-stamped-different → keep stored, in
  both the #4852 test file and the older schema-merge suite, with rationale
  in-test). Engine-driven CFC suites (boundary, CT-1740
  profile-create-real-card-add, content-addressed identity + adversarial)
  pass **unmodified** — born stamps equal runtime provenance identities by
  construction, and the adversarial fail-closed pins hold.
- Behavior spec updated in-change: §6.8 marker shape, §17.3 file
  normalization + born-stamp emission, runtime-consumption paragraph
  (version-boundary reconcile).

## Sequencing

Lands after #4852 (merged). The remaining CT-1886 legs after this: the loom
vendor bump re-runs the PR4b bio criterion end-to-end, and the setsrc-history
delegation design (Berni + Hixie, board topic index 38) takes cross-version
rotation.

---
## Summary by cubic
Makes writer-identity spelling explicit and consistent across compiles, and mints `WriteAuthorizedBy` claims born‑stamped with `moduleIdentity` so provenance stays stable; the runner now keeps the stored stamp across versions to avoid aborting schema merges. This follows CT-1886 and stabilizes claim ↔ provenance while removing the foreign‑seed hazard for new mints.

- **New Features**
  - `ts-transformers`: added `TransformationOptions.canonicalWriterIdentityFile` to map compile names to authored paths; the engine supplies this at record‑graph compile, stored‑source recompile, and batched type checks; direct compiles record verbatim paths.
  - `ts-transformers`/`schema-generator`: claims are born stamped; `__ctWriterIdentityOf` includes optional `moduleIdentity` and is applied to nested/imported bindings via `SchemaGenerationOptions.writerIdentityForSourceFile`; if `moduleIdentities` is provided but omits the defining source, compilation fails.
  - `runner`: when both sides are stamped for the same binding, keep the stored stamp (version boundary); protected‑field writes remain fail‑closed at verification while sibling writes continue.
  - `api`: JSON schema types now allow an optional `moduleIdentity` in `ifc.writeAuthorizedBy.__ctWriterIdentityOf`.

- **Refactors**
  - Consolidated the writer‑identity normalizer into `src/utils/writer-identity-file.ts` and used it for claim minting, provenance stamping, and `PolicyOf` source matching.

<sup>Written for commit aef9aa46ca1ee8ae4c86c7bbaf9a8be7222e49b8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4871?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Overriding the coverage ratchet because I do not have time to do a better thing :(

---BEGIN COPY-PASTE---
NEW_PERF_BASELINE: coverage-debt: packages/schema-generator uncovered lines = 992 lines
NEW_PERF_BASELINE: coverage-debt: packages/runner uncovered lines = 6950 lines
---END COPY-PASTE---

### DIFF
--- a/docs/specs/schema-generator/ts_to_json_schema_mapping.md
+++ b/docs/specs/schema-generator/ts_to_json_schema_mapping.md
@@ -535,7 +535,7 @@ inside those payloads.
 | `MaxConfidentiality<T, C>` | `{ maxConfidentiality: C }` |
 | `AnyOf<X>` | when nested in an IFC label tuple, one atom `{ anyOf: X }` |
 | `PolicyOf<typeof rules>` | when nested in an IFC label tuple, a policy atom carrying a compile-time module-binding marker; the ts-transformer resolves it to `{ type, policyRefKind: "module", subject, moduleIdentity, symbol, policyDigest }` |
-| `WriteAuthorizedBy<T, typeof b>` | `{ writeAuthorizedBy: { __ctWriterIdentityOf: { file, path: [binding] } } }` (`:1444-1532`) |
+| `WriteAuthorizedBy<T, typeof b>` | `{ writeAuthorizedBy: { __ctWriterIdentityOf: { file, path: [binding], moduleIdentity? } } }` (`:1444-1539`) |
 | `TrustedActionWriteWithIntegrity<…>` | writeAuthorizedBy metadata + `uiContract { helper: "UiAction", action, trustedPattern, requiredEventIntegrity }` (`:1340-1347,1455-1478`) |
 | `TrustedActionWrite<…>` | same, with `requiredEventIntegrity` defaulting to `[trustedPattern]` (`:1349-1358`) |
 | `TrustedActionUiContract<…>` | `{ uiContract: { helper: "UiAction", action, trustedPattern, requiredEventIntegrity? } }` (`:1359-1371`) |
@@ -566,13 +566,17 @@ Mechanics:
   (`mergeIfcMetadata`, `:1554-1570`); boolean schemas become `{ ifc }` /
   `{ not: true, ifc }`.
 - `WriteAuthorizedBy` writer identity resolves through import aliases to the
-  declaring file; the file name is normalized (backslashes → `/`, first path
-  segment stripped — `normalizeWriterIdentityFile`, `:2237-2241`). The
-  transformer **duplicates** this attachment for
-  `toSchema<WriteAuthorizedBy<T, typeof b>>()`
-  (`ts-transformers/.../schema-generator.ts:29-40,126-131,347-428`), with a
-  marker AST form so the identity survives literal emission (`:170-171,
-  399-422`).
+  declaring file. A transformer caller supplies
+  `writerIdentityForSourceFile`, which maps that compile name to its authored
+  spelling and can attach the defining source's content-addressed
+  `moduleIdentity`; the runner supplies both, so engine-minted claims are born
+  stamped. A transformer identity map that omits the defining source is a
+  compile error, rather than a silent downgrade to an unstamped claim.
+  Standalone schema-generator callers that omit the callback retain
+  the legacy fallback (backslashes → `/`, first path segment stripped by
+  `normalizeWriterIdentityFile`). The transformer also handles the direct-root
+  `toSchema<WriteAuthorizedBy<T, typeof b>>()` form specially so the wrapper's
+  value schema remains the root while the same identity marker is attached.
 - `SchemaGeneratorTransformer.resolvePolicyOfMarkers` replaces a valid policy
   marker with the compiled module identity, exported symbol, and policy digest.
   If it cannot match a compiler-verified exported `exchangeRules()` binding,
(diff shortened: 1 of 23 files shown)


mathpirate:
**Live acceptance result: this PR closes the PR4b profile bio-save criterion — #4852 alone does not.** Ran the native profile surface on a fresh loom acceptance instance (astral, cold-session split), reading the committed value from the presentation view's bio paragraph (`profileWish.result.bio`, live from the piece — not the observed cache, not the draft cell):

| pin | same-session bio | **cold-session bio** (the #4772 repro) |
|---|---|---|
| loom-stable-2026-07-21 (labs 0654af4b4, **#4852 merged, this PR not**) | commits | **DROPS** — presentation `hasBio` false |
| labs 23c23958f (**#4852 + this PR**) | commits | **COMMITS** — a cold browser renders the bio |

Mechanism confirmed empirically: #4852 makes verification + reconcile spelling-tolerant, but a first-time cold-session bio write never gets its claim **stamped** — exact-mint `rebindWriteAuthorizedByClaims` can't bridge the transformer's spelling shear, so reconcile-adoption has no stamped side to adopt. This PR's **born-stamped claims** (moduleIdentity baked in at compile via the engine's `identityByPath`) are exactly what closes it. So #4871 is on the critical path for the LOOM_NATIVE_PROFILE flip (loom#4074), not just hardening. Rebased onto current main (incl. your #4870 merge; the two-stamp version-boundary reconcile composes with your setsrc delegation) — gates green. Ready whenever you're back from the trip.

seefeldb (approved):

PR #4872, opened 2026-07-21
feat(lint): require a `json-ok:` justification for `JSON.parse`/`JSON.stringify`

danfuzz:
## What this is

A `deno lint` rule, `cf-lint/require-json-ok`, that flags any `JSON.parse()` or
`JSON.stringify()` call not carrying a `json-ok:` comment saying why a
plain-JSON round trip is correct there — plus a per-file baseline that lets the
existing 1621 call sites stay put while ratcheting only downward.

Opened as a draft because the idea is still under discussion. This is here to be
inspected and argued with, not merged as-is.

## Why

`JSON.stringify()` renders `NaN` and `±Infinity` as `null`, drops `undefined`
members, emits keys in insertion order rather than a canonical one, and rebuilds
a `FabricInstance` as a plain record, stripping the class identity a
`FabricError`, `FabricBytes`, or `FabricLink` carries. `JSON.parse()` inverts
none of that. Each of those losses has already produced a live defect — the
`{value: NaN}` event that transported as `{"value":null}` and was then rejected
by schema validation (caught by a CI shard on #4844, not by review), and the
`createDataCellURI` walk that decomposes a `FabricPrimitive` (still open, issue
#4527).

Plenty of call sites are fine all the same: a config file read off disk, a log
line, a test fixture, a wire format whose contract genuinely is JSON. A linter
cannot tell those apart from the dangerous ones. So the rule does not try. It
asks the author to say which this one is, and leaves judging the answer to
review — the same bargain `ban-untagged-todo` already strikes in this repo.

## The marker

```ts
// json-ok: reading a config file this repository writes itself.
const config = JSON.parse(text);
```

`json-ok:` followed by a non-empty reason. It may sit at the end of the call's
own line, in the run of whole-line comments directly above the call, or in the
run directly above the statement holding it — in each case with no blank line in
between. Both `//` and `/* */` carry it, and a leading `*` is ignored so it
reads naturally inside a doc comment.

The reason bar is deliberately one letter: enough to reject a bare `// json-ok:`
that says nothing, not enough to pretend a linter can judge prose. **This is the
main tunable** if the group wants something stricter.

## The ratchet

`tasks/json-ok-baseline.json` records, per file, how many unjustified calls that
file may hold — 1621 across 475 files today.

- At or under budget: reports nothing.
- Over budget: reports **every** unjustified call in the file, since a count
  cannot say which one is new.
- Absent from the baseline: budget of zero, so new files are held in full.

`deno task check-json-ok` fails in **both** directions — over budget is a
regression, under budget is a stale entry asking for `--update`. So ground
gained cannot quietly be given back. It counts by re-running `deno lint` with
`CF_JSON_OK_REPORT_ALL=1` (which makes the plugin ignore the baseline) rather
than walking the tree itself, so there is exactly one answer to which files are
in scope, and `deno.jsonc`'s `lint.exclude` cannot be silently reinterpreted.

## Precedent this follows

- `ban-untagged-todo` (already on) — banned unless annotated, ~167 sites.
- `packages/ts-transformers/lint-plugins/no-node-get-text.ts` — the existing
  custom lint plugin; a root-level `lint.plugins` composes with it rather than
  overriding it (verified).
- `check-unused-deps` / `check-single-copy-deps` /
  `typetotypenode-call-sites.test.ts` — the house pattern of an allowlist with a
  mandatory reason **and** a paired stale-entry check. The bidirectional failure
  here is that same discipline, counted rather than enumerated.
- `coverage-debt:` in `tasks/perf-check.ts` — the existing numeric ratchet. Not
  reused: its comparator is a rolling median with variance bands, built for
  noisy timings, and this count is deterministic.

## Cost

Full-repo `deno lint` with the plugin: **~3.0s** for 3991 files.
`deno task check-json-ok`: ~3s. Both cheap enough to sit in the `check` job.

## What I verified

- `deno fmt --check`, `deno lint`, `deno task check`, `deno task check-docs`,
  `deno task check-json-ok` — all green.
- `tasks` suite: 473 passed, 0 failed (35 new plugin tests + 9 checker tests).
- Ratchet fires in both directions, end to end: a new unjustified call in a
  baselined file lights up all of them with the budget named; justifying an
  existing one makes `check-json-ok` demand a smaller baseline.
- Baseline re-derived and cross-checked after merging `origin/main`.
- Full `deno task test` stops at a pre-existing `deno-web-test` browser-test
  failure that reproduces identically on the parent commit — unrelated.

## Open questions for review

1. **Reason bar.** One letter today. Minimum length? Require a sentence?
2. **Scope.** Repo-wide, or start at the Fabric-value layers (`data-model`,
   `memory`, `runner/src`) where the harm actually lives and leave `cf-harness`
   / test fixtures alone? The `cf-harness` test corpus alone is ~330 of the 1621.
3. **`// deno-lint-ignore cf-lint/require-json-ok`** still bypasses the rule
   with no reason at all. `ban-unused-ignore` keeps it from rotting, but it is a
   hole. Worth closing, or is the social cost enough?
4. **Baseline granularity.** Per-file counts are cheap to generate but let a
   call move between files within one budget. Per-site allowlists with written
   reasons are strictly stronger — and 1621 reasons to write.

https://claude.ai/code/session_012hy7CtkNrbnBZsrE1o2Gco

---
## Summary by cubic
Adds a `deno lint` rule `cf-lint/require-json-ok` that requires a `json-ok:` reason on any `JSON.parse`/`JSON.stringify` call, with a per-file baseline and a CI check that only ratchets down. This guards against lossy plain-JSON round trips while letting safe sites document intent.

- **New Features**
  - Lint plugin `tasks/json-ok-lint-plugin.ts` enforcing `cf-lint/require-json-ok` with support for inline or preceding `json-ok:` comments.
  - Baseline `tasks/json-ok-baseline.json` (per-file budgets) and checker `tasks/check-json-ok.ts`; `deno task check-json-ok --update` regenerates the baseline.
  - CI step added in `.github/workflows/deno.yml`; plugin wired in `deno.jsonc`; docs updated in `docs/development/DEVELOPMENT.md`.

- **Migration**
  - Add `// json-ok: <reason>` either on the call line or in the comment run directly above the call/statement; reasons must be non-empty. If values can carry Fabric data, use codecs from `@commonfabric/data-model` instead.
  - New files have a budget of zero; baseline only ratchets down. If CI flags stale entries, run `deno task check-json-ok --update` and commit the baseline change.
  - Use `deno task check-json-ok` locally to see true counts; set `CF_JSON_OK_REPORT_ALL=1` to ignore the baseline when auditing.

<sup>Written for commit ed17d942c700077d8b82b9fec408a45a002e5dc2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4872?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


danfuzz:
Not worth doing right now.

PR #4876, opened 2026-07-21
fix: accept FabricValues from lifted actions

mathpirate:
## What changed

- Make `fabricFromNativeValue()` the authority for action-result legality while preserving the recursive Reactive/Cell scan and path-aware diagnostics.
- Recognize supported native objects returned across the sandbox realm boundary.
- Add an end-to-end battery covering lift return, cell write, store, cold read, and re-emission, led by `FabricBytes` from `fetchBinary`.
- State the actual public contract in errors: actions may return FabricValues, Reactives, or Cells.

## Why

The action-return validator predates materializable FabricValue special objects and maintained a separate JSON-oriented rejection list. That rejected values the data model can already materialize, including `FabricBytes`, bigint, non-finite numbers, and interned symbols.

Consumers can now return supported FabricValues from lifted actions without being blocked by the runner's older, narrower validator.

## Explicit downstream carve-outs

- The SES action boundary currently invokes `Date.toJSON()`, so the battery observes the stable ISO string rather than a `FabricEpochNsec`.
- A cold `FabricError` is verified through persistence, but omitted from the second lift because independently loaded data-model module identities do not yet recognize it for re-emission.

## Validation

- `deno task check`
- Focused lint on all changed source and test files
- 299 focused test steps across native tagging/conversion, action-result validation, the new cold round-trip battery, and fetch-binary materialization
- `git diff --check`

---
## Summary by cubic
Normalizes sandboxed results at egress and makes `@commonfabric/data-model`’s `fabricFromNativeValue()` the sole validator, so lifted actions can return FabricValues and supported cross‑realm native objects with clear, path‑aware errors. Also rejects null‑rooted custom instances and adds an end‑to‑end round‑trip test.

- **Refactors**
  - Introduces `normalizeSandboxResult` to adapt cross‑realm containers, canonicalize native leaves (Date → `FabricEpochNsec`, `RegExp` → `FabricRegExp`, `Uint8Array` → `FabricBytes`, `Error` → `FabricError`), and preserve Reactives/Cells and shared references.
  - Runner uses normalized results and exports `validateAndCheckReactives` from sandbox normalization; host graphs deep‑freeze when no opaque leaves are present.
  - Data model tagging uses optional `Error.isError` when available.

- **Bug Fixes**
  - Admits materializable FabricValues and primitives previously rejected; still rejects `Map`/`Set` with hints, blocks `Symbol.toStringTag` spoofing, rejects extra enumerable properties on Date/RegExp, and now rejects class instances with null‑rooted prototypes.
  - Clearer, path‑aware error text with included cause and explicit contract; tests cover normalization, shared references, spoof and brand‑getter failures, circular refs, and the full cold round‑trip and re‑emission.

<sup>Written for commit c1631d445cac11eb50cc9b32ad11e8eb7c342837. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4876?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/data-model/src/native-type-tags.ts
+++ b/packages/data-model/src/native-type-tags.ts
@@ -43,6 +43,46 @@ export const NATIVE_TAGS = Object.freeze(
 /** One of the native-instance tag strings. */
 export type NativeTag = typeof NATIVE_TAGS[keyof typeof NATIVE_TAGS];
 
+const regexpSourceGetter = Object.getOwnPropertyDescriptor(
+  RegExp.prototype,
+  "source",
+)?.get;
+
+/**
+ * Recognizes native instances created in another realm. Constructor identity
+ * cannot cross a realm boundary, while the intrinsic brand checks below do.
+ * `Object.prototype.toString` narrows the candidate and the native operation
+ * verifies the internal slot so `Symbol.toStringTag` spoofing is rejected.
+ */
+function tagFromCrossRealmNativeValue(value: object): NativeTag | null {
+  const brand = Object.prototype.toString.call(value);
+  try {
+    switch (brand) {
+      case "[object Date]":
+        Date.prototype.getTime.call(value);
+        return NATIVE_TAGS.Date;
+      case "[object RegExp]":
+        regexpSourceGetter?.call(value);
+        return regexpSourceGetter === undefined ? null : NATIVE_TAGS.RegExp;
+      case "[object Uint8Array]":
+        return ArrayBuffer.isView(value) &&
+            (value as Uint8Array).BYTES_PER_ELEMENT === 1
+          ? NATIVE_TAGS.Uint8Array
+          : null;
+      case "[object Map]":
+        Map.prototype.has.call(value, tagFromCrossRealmNativeValue);
+        return NATIVE_TAGS.Map;
+      case "[object Set]":
+        Set.prototype.has.call(value, tagFromCrossRealmNativeValue);
+        return NATIVE_TAGS.Set;
+      default:
+        return null;
+    }
+  } catch {
+    return null;
+  }
+}
+
 /**
  * Maps a constructor to its native-instance tag. Returns the tag string if
  * the constructor is a recognized type (JS builtins or system-defined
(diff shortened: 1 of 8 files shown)


packages/data-model/src/native-type-tags.ts
```diff
@@ -161,8 +161,11 @@ export function tagFromNativeValue(value: unknown): NativeTag | null {
     // been severed, or one from another realm. An ordinary subclass (including
     // `DOMException`) never gets here: `tagFromNativeClass()` matches it via
     // `prototype instanceof Error`. The internal slot survives either way,
-    // which is what `Error.isError()` reads.
-    if (Error.isError(value)) return NATIVE_TAGS.Error;
+    // which is what `Error.isError()` reads where the runtime implements it.
```
danfuzz: I'm guessing this circumlocution is required because our current SES environment doesn't include `Error.isError()`. I ran into this a few weeks ago, and one of my current pending items is trying to get that method included. At the time I determined that it's not as simple as _just_ bumping an SES version, but perhaps it won't turn out to be too hard.

Assuming this analysis is correct, and that this PR needs to land before we sort out SES, please add a `TODO(danfuzz)` here about the situation so the cleanup site won't be lost. Thanks!

packages/runner/src/sandbox/result-normalization.ts
```diff
@@ -0,0 +1,323 @@
...
+    );
+  }
+}
+
+/**
+ * Canonicalizes native leaf values while their foreign-realm identity is still
+ * known. The data model only receives Fabric values or ordinary host values;
+ * it does not need realm-specific dispatch.
+ */
+function normalizeSandboxNativeLeaf(value: unknown): unknown {
```
danfuzz, line 73: This looks perfect!

danfuzz (commented):
LGTM from the `data-model` perspective, modulo the `isError()` request (which can _just_ be adding a `TODO(danfuzz)`).

**Update:** Caveat obsoleted by the merge of #4906.

packages/data-model/src/native-type-tags.ts
danfuzz: I went ahead and made PR #4906, which gets `Error.isError()` to be defined. Unclear if/when it will land.

danfuzz: **Update:** PR #4906 has now been merged, so I think it's safe to revert this chunk.

mathpirate: Done in 8723c8277 — dropped the shim for the direct Error.isError(value) now that #4906 landed. Thanks!

danfuzz (approved):

PR #4882, opened 2026-07-21
fix(runner): surface sync failures that resolve as silent absence

mathpirate:
## Why

Yesterday's estuary investigation pinned this at the wire: a foreign CLI read of an ACL-private home gets an explicit typed response from the server —

```
AuthorizationError: Principal did:key:z6MkpT5… lacks READ on space did:key:z6Mks3qW…
```

— and the storage client drops it on the floor. The watch-refresh machinery wraps the mount throw into the pull result's `error`; `syncCell` captures that into pending-load bookkeeping (which only scheduler waiters consume) and resolves `sync()` normally. A direct caller — the CLI, a probe, an agent — reads the doc as absent. **Denied, transport-failed, and genuinely absent all collapse into the same silent `undefined`**, at every log level. This is the concrete mechanism behind the "loud sync-state" board topic (`fid1:yNXW3zia…`), and it cost real hours during the home-space bricking triage.

## What

- Log the carried failure at both seams where it currently vanishes — `syncCell` and the link-target pull tracker (`trackPendingProviderSync`) — under `storage.v2` / `sync-load-failure`, at error level (visible by default).
- **Deduplicate per (space, error identity)**: a denied space repeats the identical failure for every doc pulled from it; one line per distinct failure keeps a 39-doc walk of a denied space at one line, not 39. Bounded set (reset at 256) so pathological error diversity can't grow it unboundedly.
- Read `name`/`message` structurally: pull errors are plain result objects (`IConnectionError` et al.), not `Error` instances — `instanceof`-based extraction printed `[object Object]`.
- **No contract change**: `sync()` still resolves rather than rejects on a carried error — pinned by the new test alongside the surfacing.

## Verification

- New `packages/runner/test/sync-load-failure-surfacing.test.ts` (mock denying `SessionFactory`, per the list-shrink harness pattern): denial logs exactly once per space with the server's principal-naming text (content asserted, not just counts); second doc in the same space deduplicates; distinct spaces log distinctly; `sync()` resolves and reads undefined throughout. **Red/green verified**: both steps fail on unpatched `v2.ts`.
- Full runner suite: **997 passed (5324 steps), 0 failed** — no healthy-path test trips the new line, which doubles as the loudness-hazard check (transient reactive voids never reach this seam; only pulls that complete *carrying* an error do).
- Live probe against estuary: the exact silent case from the investigation now prints the doc, space, and the server's denial text.

## Non-goals / follow-ups (the loud-sync-state topic's larger asks)

- Typed classification for callers (absent vs denied vs transport as a programmatic result) — needs an interface decision; the denial text is in the message today because `toConnectionError` flattens the original error name.
- Inspector / pending-load API exposure so tooling like #4874's CLI diagnostics can distinguish "no raw because denied" from "no raw because absent" — composes with this change, designed separately.

---
## Summary by cubic
Log sync failures that previously appeared as silent absence, without changing the `sync()` contract. ACL denials and transport errors are now logged once per space/error pair to make access issues visible.

- **Bug Fixes**
  - Log carried failures in `packages/runner/src/storage/v2.ts` at both seams (`syncCell` and the link-target pull tracker) under `storage.v2`/`sync-load-failure` at error level, including doc id, space, and server error name/message.
  - Deduplicate by (space, error identity) using a bounded set (256); clear on cap to avoid growth.
  - Read `name` and `message` from plain result objects instead of relying on `instanceof Error`.
  - Keep `sync()` resolve-not-reject behavior; add `packages/runner/test/sync-load-failure-surfacing.test.ts` covering per-space dedup, link-target pulls, and the dedup cap reset.

<sup>Written for commit 81684ff993e65b30bf1afe4a5b13a6f94075cda5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4882?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/storage/v2.ts
+++ b/packages/runner/src/storage/v2.ts
@@ -1104,6 +1104,11 @@ export class StorageManager implements IStorageManager {
     waiters: Set<(failure: unknown) => void>;
   }>();
   #nextPendingLoadGeneration = 1;
+  // Sync failures already logged, keyed by (space, error identity). A denied
+  // space repeats the identical failure for every doc pulled from it; one line
+  // per distinct failure keeps the surfacing readable. Bounded: at the cap the
+  // set resets, trading a repeated line for an unbounded set.
+  #loggedSyncFailures = new Set<string>();
 
   private registerPendingLoad(
     address: { space: MemorySpace; scope: CellScope; id: URI },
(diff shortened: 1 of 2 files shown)


mathpirate:
Coverage ratchet round: the +23 was the two seams the initial tests didn't drive. Added two more pins — a **denied link-target pull** (data-URI cell whose value links into an unmountable space — the cross-space shape from the estuary incident, driving the `trackPendingProviderSync` seam) and the **dedup-cap reset** (257 distinct denials: cap resets, still logs, early repeat logs again — the accepted trade for a bounded set). All previously-uncovered lines now covered locally; full runner suite 997/5326 green.

ubik2 (approved):
lgtm

nit: when we say "sync completed without data for", we may have gotten the data for the main doc, but not the corresponding cfc schema doc.

PR #4883, opened 2026-07-21
fix(piece): replace unloadable stale sourceless roots — the pin protects only a dead page

mathpirate:
## Why (the last mile of the home-space heal)

Estuary now runs #4873's merge — and homes are still bricked. The update flag opened, but `checkAndUpdateDefaultPattern` step 5.5 pins **stale sourceless** roots: pre-provenance-era roots (every June-era home in the fleet) have `storedSource === undefined` and a stale identity, so they return `current` before the repair path runs. The pin's reason is real — a sourceless root might be a deliberate **custom program** (custom recreation stamps no provenance, by design), and rolling it forward would overwrite that choice.

## What

An **unloadability tiebreak, scoped to home spaces**: when a stale sourceless *home* root's stored pattern explicitly cannot be loaded by the current runtime (the probe resolves `undefined`), the ambiguity no longer protects a live object — replace it with the official `home.tsx` through the existing attested-fetch + CAS path, recording the displaced `{identity, symbol, displacedAt}` under new `displacedPattern` meta (registered in `MetaField`) as an audit/forensic pointer (the displaced program's compiled artifacts remain content-addressed in the space; automated restoration is deliberately out of scope).

The exception's edges, each pinned by a test (shaped by codex-sol's review — see thread):

- **Probe `undefined` ⇒ replace; probe exception ⇒ pinned** (a failed check is not authority; logs `stale-root-probe-failed`, mutates nothing).
- **`cfcEnforcementMode: "disabled"` ⇒ pinned** (the by-identity probe is unsupported there and returns `undefined` unconditionally — probe-unsupported ≠ artifact-dead).
- **Non-home spaces ⇒ always pinned** (fleet evidence exists only for home; widening needs its own evidence + a tested restoration path).
- **Loadable-in-current-runtime ⇒ pinned**, exactly as before (the existing custom-home test, untouched). The probe is deliberately *current-runtime* loadability, not cold-restart survival: a warm artifact can only cause extra pinning, never extra replacement.

Unloadable roots **with** provenance already reconciled (#4865); sourceless+unloadable+home was the gap — and it is exactly the bricked fleet.

## Docs (same-change rule)

`pattern-updates.md` defines the exception (undefined-vs-exception, current-runtime probe semantics, CFC-disabled behavior, home-only scope, `displacedPattern`'s exact promise); `EXPERIMENTAL_OPTIONS.md`'s pinning sentence carries the same exception with a pointer.

## Verification

- Home replace test (red/green vs unpatched: `current`, no replacement) + three edge pins: probe-reject → unchanged; CFC-disabled → unchanged (real probe, no shadow); non-home unloadable → unchanged.
- The custom-loadable pin test and #4865's provenance'd-unloadable reconcile test pass untouched — the two protections this change must not disturb.
- Full piece suite 20 passed (242 steps) / 0 failed; workspace typecheck, fmt, lint clean.

## Notes for review (@seefeldb)

Draft pending your take — this narrows #4865's explicit "stays pinned until an explicit migration" stance for the one cell of the matrix where the pin protects only a dead page, on the deploy reality that no explicit-migration tool can reach private homes (owner-key-in-browser; no headless path).

### DIFF
--- a/docs/development/EXPERIMENTAL_OPTIONS.md
+++ b/docs/development/EXPERIMENTAL_OPTIONS.md
@@ -228,7 +228,13 @@ propagate](#how-flags-propagate).
   revision through `COMMIT_SHA`. Persisted roots are resolved without starting.
   A pre-provenance root may be back-filled only when its stored
   `{ identity, symbol }` exactly equals the build-attested current official
-  entry; stale, custom, and repository-pinned sourceless roots remain pinned.
+  entry; stale, custom, and repository-pinned sourceless roots remain pinned —
+  except a stale sourceless **home** root whose stored pattern the current
+  runtime explicitly cannot load (probe resolves `undefined`; a probe error
+  stays pinned, and `cfcEnforcementMode: "disabled"` — where the probe is
+  unsupported — stays pinned too): that root is replaced with the official
+  home.tsx, recording the displaced ref under `displacedPattern` meta (see
+  pattern-updates.md for the full exception semantics).
   URL-based creation and recreation stamp provenance; custom `RuntimeProgram`
   recreation does not. The check remains best-effort; if identity lookup or
   replacement compilation is unavailable, the subsequent root start retains
--- a/docs/specs/pattern-imports/pattern-updates.md
+++ b/docs/specs/pattern-imports/pattern-updates.md
@@ -103,6 +103,30 @@ official entry appropriate to the space (`home.tsx` for Home,
 Neither space type nor an author-controlled source filename is provenance: a
 stale, custom, or repository-pinned sourceless root stays pinned.
 
+One exception, scoped to **home spaces** (`space === userIdentityDID`): a stale
+sourceless home root whose stored pattern the current runtime **explicitly
+cannot load** is replaced with the official `home.tsx` (the 2026-07-21 estuary
+migration bricked every pre-provenance home root, and no explicit-migration
+tool can reach a private home whose owner key lives only in a browser). The
+exception's semantics are deliberately narrow:
+
+- Replacement is authorized only when the load probe resolves `undefined` —
+  the artifact unavailable through every supported recovery path. A probe
+  **exception** is a failed check, not evidence: the updater logs and stays
+  pinned.
+- The probe asks "loadable in the **current runtime**" (in-memory artifact
+  index, live evaluated modules, then durable storage) — not "survives a cold
+  restart". A warm artifact can only cause extra pinning, never extra
+  replacement.
+- Under `cfcEnforcementMode: "disabled"` the by-identity probe is unsupported
+  (it returns `undefined` unconditionally), so the updater stays pinned there.
+- Non-home spaces always stay pinned; widening the destructive fallback needs
+  its own evidence and a tested restoration path.
+- The replaced root records the displaced `{ identity, symbol, displacedAt }`
+  under `displacedPattern` meta. This is an audit/forensic pointer — the
+  displaced program's compiled artifacts remain content-addressed in the
+  space — not (yet) an automated restoration mechanism.
+
 System patterns run this loop automatically at space open (always-update);
 published patterns will run it behind an explicit user action (§ Phasing).
 
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -371,6 +371,9 @@ export type MetaField =
   | "patternSource" // provenance: the source a piece tracks for updates (a
   // toolshed pattern path, or later a `cf:` fabric ref)
   | "patternRepository" // optional caller-supplied repository locator
+  | "displacedPattern" // {identity, symbol, displacedAt}: the prior pattern
+  // reference recorded when system-pattern auto-update replaces an unloadable
+  // sourceless root — the recovery pointer for a displaced custom program
   | "internal"
   | "schema"
   | "slug";
--- a/packages/piece/src/ops/pieces-controller.ts
+++ b/packages/piece/src/ops/pieces-controller.ts
@@ -708,13 +708,53 @@ export class PiecesController<T = unknown> {
       // A sourceless root is eligible only when its full content identity (which
       // includes authored module paths) is exactly the official current entry.
       // This admits a known system identity, never a lookalike filename. A stale
-      // sourceless root stays pinned until an explicit migration supplies its
-      // durable provenance.
-      if (
-        storedSource === undefined &&
-        (runningRef.identity !== currentId || runningRef.symbol !== "default")
-      ) {
-        return "current";
+      // sourceless root that still LOADS stays pinned: it may be a custom
+      // program (custom recreation deliberately stamps no provenance), and
+      // rolling it forward would overwrite that choice with the system pattern.
+      //
+      // When the stored root cannot cold-load at all, the pin protects only a
+      // dead page: an obsolete system root heals, and a broken custom root
+      // degrades to a WORKING system root with its displaced identity recorded
+      // under `displacedPattern` meta for recovery. (2026-07-21: a runtime
+      // migration bricked every pre-provenance home root, and this pin kept
+      // them bricked after the update flag opened.)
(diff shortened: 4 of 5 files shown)


mathpirate:
I traced this through `checkAndUpdateDefaultPattern`, `PatternManager.loadPatternByIdentity`, the storage-backed/cold-load paths, the CAS, and bootstrap. The overall recovery approach makes sense for the Estuary incident, and the identity/build attestation plus transactional swap look sound. I do think there is one must-fix safety issue, followed by several important scope/contract issues.

## 1. Must fix: a probe exception currently authorizes replacement

In the new stale-sourceless branch, all three probe outcomes collapse into two behaviors:

| Probe outcome | What it establishes | Current behavior |
| --- | --- | --- |
| returns a pattern | the root is loadable | keep pinned |
| returns `undefined` | the artifact is unavailable through the supported recovery paths | replace |
| throws | the check itself failed | **also replace** |

That final row is unsafe. `loadPatternByIdentity()` can throw while reading the compiled closure or verified source closure from durable storage. A transaction/backend/runtime failure is not proof that an ambiguous sourceless root is dead. For a deliberate custom root, the unconditional `catch {}` therefore turns a transient operational failure into authority to overwrite the root with the system pattern.

It also bypasses the enclosing function's existing fail-closed behavior: without this inner catch, the outer catch logs the failed check and returns `"current"` without mutating anything.

I recommend authorizing replacement **only** when the probe explicitly resolves `undefined`. On rejection, log and return `"current"` (or remove the inner catch and let the outer handler do that):

```ts
if (staleSourceless) {
  try {
    const staleRoot = await runtime.patternManager.loadPatternByIdentity(
      runningRef.identity,
      runningRef.symbol,
      space,
    );
    if (staleRoot !== undefined) return "current";
    // Only an explicit undefined reaches replacement.
  } catch (error) {
    pieceUpdateLogger.warn("stale-root-probe-failed", () => [
      space,
      runningRef,
      error,
    ]);
    return "current";
  }
}
```

Please add two separate regression cases:

- probe resolves `undefined` → replacement occurs;
- probe rejects/throws → identity, source, and displaced metadata remain unchanged.

The existing same-identity repair branch can reasonably rebuild after a load exception because the build-attested official identity already authorizes that source. The stale-sourceless branch is different: the old identity may be a deliberate custom program, so a failed check cannot establish replacement authority.

## 2. The “cold-load” probe is not necessarily cold

`loadPatternByIdentity()` first consults `addressableByIdentity` and the evaluated-module cache before reading durable storage. A warm artifact can therefore make this branch return `"current"` even if the root would fail after a process restart. On a fresh Estuary worker the call is likely to reach durable storage, so this may not block the immediate incident, but the comments/tests currently promise a stronger property than the method provides.

If the intended tiebreak is specifically “will survive a cold restart,” this needs either a storage-only/force-cold probe or an integration test using a fresh runtime against the persisted space. Otherwise I would rename/reword the guarantee to say “loadable in the current runtime.”

## 3. CFC-disabled runtimes are classified as unloadable

`loadPatternByIdentity()` returns `undefined` immediately when `cfcEnforcementMode === "disabled"`. Under this PR, that makes every stale sourceless root look dead and eligible for replacement, even though by-identity recovery is merely disabled.

No shipped Estuary host appears to use disabled mode today, so this is not the fleet incident, but logically the updater should distinguish “probe unsupported/disabled” from “artifact definitely unavailable,” or explicitly gate this path on a CFC mode that supports the probe. A small regression test for disabled mode would pin that contract.

## 4. I recommend scoping the destructive fallback to Home first

For Home, there is concrete fleet evidence that these are pre-provenance system roots broken by the migration. For arbitrary spaces, a sourceless stale root is more plausibly an intentional custom program. This PR intentionally replaces such a custom program if the current runtime cannot load it.

`displacedPattern` is useful audit metadata, but it is not presently an operational recovery mechanism: there is no restoration consumer/UI, and the current runtime has just established that it cannot load that identity. The old state/artifacts may permit future recovery, but this PR does not make the replacement reversible today.

Given the incident being fixed, I would use `space === runtime.userIdentityDID` as the additional eligibility condition and defer non-home replacement until there is evidence it is needed plus a tested restoration path. If the team deliberately accepts the all-space tradeoff, the destructive behavior and recovery limitations should be made explicit in the live docs.

## 5. Current CI failure

`makeUnloadable()` is declared `async` but contains no `await`, so `deno lint` reports `require-await`. GitHub's Deno workflow currently fails in the Check job at the lint step. Make the helper synchronous (`function makeUnloadable(...): () => void`) and remove `await` from its two call sites.

## 6. Live docs directly contradict the new behavior

`docs/specs/pattern-imports/pattern-updates.md` currently says a stale/custom sourceless root stays pinned and step 4 admits only an exact current official identity. `docs/development/EXPERIMENTAL_OPTIONS.md` says the same. This PR changes that central safety contract and adds `displacedPattern`, so both live docs need to be updated in the same PR.

The docs should define:

- the explicit-unloadability exception;
- `undefined` versus exceptions/indeterminate probe failures;
- whether the probe is current-session loadability or durable cold-loadability;
- CFC-disabled behavior;
- Home-only versus all-space scope;
- what `displacedPattern` preserves and how/if it can currently be restored.

## What I verified at this exact head

At `4e4afa1bf9090be83027e1e6484bffd6dc87796d`:

- full `packages/piece` suite: **20 passed, 240 steps, 0 failed**;
- workspace typecheck: **passed, 400 paths**;
- formatting: **passed**;
- lint: **failed** on `require-await` in the new helper;
- GitHub Actions: all displayed test/build shards passed; the Check job failed at lint.

The build/version gate, attested source fetch, compiled-entry identity check, compare-and-swap revalidation, atomic displaced-ref/identity/provenance write, and re-resolution-before-bootstrap all look correct. The exception semantics above are the main correctness blocker.

mathpirate:
All six taken — thank you, this was a genuinely sharpening review. At `320b584ac`:

**1. Probe exception ⇒ replacement (must-fix): fixed as recommended.** Only an explicit `undefined` resolution authorizes replacement; a thrown probe logs `stale-root-probe-failed` and returns `current`, matching the enclosing function's fail-closed posture. Both regression cases added: probe-`undefined` → replaced (the original test), probe-reject → identity/source/meta all unchanged.

**2. Warm-probe honesty: rewording taken.** Comments, docs, and test names now say *loadable in the current runtime*, and the spec text records the asymmetry that makes the weaker probe safe here: a warm artifact can only cause extra pinning, never extra replacement. A force-cold probe (or fresh-runtime integration test) noted as the strengthening path if the stronger property is ever wanted.

**3. CFC-disabled ⇒ spurious-undefined: gated + pinned.** `cfcEnforcementMode === "disabled"` now returns `current` before the probe (probe-unsupported ≠ artifact-dead), with a regression test using the real probe (no shadow — the short-circuit is the behavior under test).

**4. Home-only scope: taken.** `space === runtime.userIdentityDID` is now an eligibility condition; the former non-home replacement test is inverted into a scoping pin (non-home stale sourceless stays pinned even when unloadable). Also adopted your `displacedPattern` framing — the docs now call it an audit/forensic pointer (artifacts remain content-addressed), explicitly not-yet a restoration mechanism.

**5. `require-await` lint: fixed** — helper is synchronous (`shadowLoadProbe`, now parameterized for the reject case); my earlier lint run covered the source files but not the test file, mea culpa.

**6. Live docs: updated in-PR.** `pattern-updates.md` gains the exception block defining exactly your list — undefined-vs-exception, current-runtime loadability, CFC-disabled behavior, home-only scope, and what `displacedPattern` does and doesn't promise; `EXPERIMENTAL_OPTIONS.md`'s pinning sentence carries the same exception with a pointer.

Full piece suite 20/242 green, workspace typecheck/fmt/lint clean. Still draft for @seefeldb's read on narrowing the #4865 stance.

seefeldb:
LGTM, and agree on narrowing #4865 stance

seefeldb (approved):

PR #4887, opened 2026-07-21
fix(schema-generator): carry signed and non-finite numbers through default extraction

danfuzz:
Teaches the schema `default` extractors to recognize a number in every spelling
TypeScript gives it — sign-prefixed (`-1`) and the non-finite globals (`NaN`,
`Infinity`) — where before they knew only bare numeric literals and silently
dropped everything else. Fixes both extractors, in `ts-transformers` and
`schema-generator`, through one shared helper rather than two implementations.

Also makes the `ts-transformers` evaluator see through parentheses and the
type-only assertion forms, which it dropped entirely — `(-1)`, but equally
`("some text")` and even a bare `(1)`. That one is a live bug on `main`
independent of this branch's subject, found while reviewing it.

Adds explicit coverage for values that previously had none: fixtures per case,
unit tests for the helper (including that it declines to fold a *shadowed*
`NaN`, and that seeing through parentheses does not widen what reads as a
number), and — now that the golden format can hold them — a golden that states
`-0`, `NaN` and the infinities outright.

## Problem

The two AST evaluators that recover schema `default` values recognized only
bare `NumericLiteral`s. A negative number is a `PrefixUnaryExpression`, and
the non-finite values are identifiers, so each of them evaluated to
`undefined` — including `-1`, the canonical sentinel default.

Reproduced with `deno task cf check --show-transformed --no-run`:

| source | emitted before | emitted after |
| --- | --- | --- |
| `toSchema<C>({ default: { n: 5 } })` | `"default": { n: 5 }` | unchanged |
| `toSchema<C>({ default: { n: -5 } })` | `"default": {}` | `"default": { n: -5 }` |
| `toSchema<C>({ default: { n: NaN } })` | `"default": {}` | `"default": { n: NaN }` |
| `const D = [-1, 2, -3]` as a `typeof` default | `[undefined, 2, undefined]` | `[-1, 2, -3]` |

The two sites failed differently:

- `ts-transformers` `evaluateExpression` backs `toSchema<T>({ default: ... })`
  and skips undefined-valued properties, so the property vanished from the
  emitted schema with no diagnostic.
- `schema-generator` `extractValueFromExpression` reads a `typeof CONST`
  initializer and assigns unconditionally, so the property was emitted
  carrying a literal `undefined`, and array elements became holes in place.

Type-level defaults (`Default<number, -1>`) were never affected — those
resolve through the checker's literal types and never reach these evaluators.
The existing `negative-number-default` fixture covers that path and is
unchanged.

## Change

Both sites now go through `numberFromExpression`, a new leaf module in
`schema-generator/src/typescript/`, exported as the `./numeric-expression`
subpath (the same shape as the `cell-brand` / `property-name` helpers
`ts-transformers` already imports). It handles sign prefixes and the
non-finite globals, and declines to fold a shadowed `NaN` / `Infinity` —
quietly reading a local binding as the global would be the same class of
silent wrongness this fixes.

### Parentheses (found in review)

`ts-transformers`' `evaluateExpression()` dropped **every** parenthesized
option, of whatever type. Against the emitted output, before the fix:

| source | emitted |
| --- | --- |
| `{ default: { n: (-1) } }` | `"default": {}` |
| `{ default: { n: (1) } }` | `"default": {}` |
| `{ description: ("hi") }` | property absent entirely |
| `{ description: "hi" }` | `description: "hi"` ✓ |

The second and third rows have nothing to do with signed or non-finite numbers
— a parenthesized bare literal and a parenthesized *string* were dropped just
as surely. So the unwrapping goes at the top of `evaluateExpression()`,
covering parentheses and `as` / `satisfies` / `<T>x` for every value type at
once. That also brings this extractor into line with its `schema-generator`
counterpart, which has always unwrapped them; the two disagreeing was the
underlying defect.

`numberFromExpression()` unwraps too, for the case a caller-level fix cannot
reach: parentheses nested *inside* the expression, as in `-(1)`.

## Testing

- 8 unit tests for `numberFromExpression` — signs, nested signs, `-0` identity
  via `Object.is`, non-finite globals, shadowing, parentheses (including
  `-(1)`), and the expressions it declines.
- `default-typeof-const-numbers` fixture — the `typeof CONST` path, with the
  values legible in the golden rather than buried in assertions.
- `numeric-option-defaults` fixture — the options-object path.
- `parenthesized-options` fixture — parentheses across option types: a
  description, a string and a number inside `default`, and elements inside an
  array option (which exercises the evaluator's recursion).
- One case stays a hand-written assertion on purpose: that a *shadowed* `NaN`
  is NOT folded. That is a claim about what does not happen, and a golden
  states what a schema contains.
- Every new fixture and test was run against the unfixed code and observed to
  fail. The exceptions are called out where they sit: the shadowing test and
  the "parentheses do not widen what reads as a number" cases assert negatives
  and pass in both states by design.
- Full suites green: `schema-generator` 45 passed, `ts-transformers` 1105
  passed, `./tasks/check.sh` clean, `deno fmt --check` clean.

## On the testing, and why this waited

Golden fixtures originally could not express these values at all: the
`.expected.json` format cannot represent `NaN` / `±Infinity` / `-0`, and
`normalizeSchema` additionally round-tripped through
`JSON.parse(JSON.stringify(…))`, destroying them before serialization. A golden
that cannot hold a value cannot guard it, and it fails *silently* — the
flattened golden and the buggy output agree.

That is why this PR sat as a draft rather than landing with hand-rolled
assertions. #4903, #4905 and #4904 fixed the harness first; the schema-generator
side of this change is now an ordinary fixture, with the values legible in the
golden:

```json
"zero": { "/SpecialNumber@1": "-0" }
"nan":  { "/SpecialNumber@1": "NaN" }
```

One case stays a hand-written assertion on purpose: that a *shadowed* `NaN` is
NOT folded to the global. That is a claim about what does not happen, and a
golden states what a schema contains — the wrong shape for it.

The ts-transformers fixture (`numeric-option-defaults`) never needed any of
this. Its golden is emitted TypeScript source, which spells these values
natively (`nan: NaN`, `nzero: -0`); JSON goldens were the only lossy carrier.

---
## Summary by cubic
Preserves signed and non‑finite numbers in schema defaults and `toSchema` options, and makes parentheses and type‑only assertions transparent at any depth so values aren’t dropped. `-1`, `-0`, `NaN`, and `±Infinity` now emit correctly for both `toSchema({ default: ... })` and `Default<..., typeof CONST>`, without folding shadowed `NaN`/`Infinity`.

- **Bug Fixes**
  - Added `numberFromExpression` to evaluate numeric AST nodes (signed literals and `NaN`/`Infinity`), unwrap parentheses and type‑only wrappers (`as`, `satisfies`, `<T>x`) even when nested (e.g., `-(1 as number)`), and respect shadowing; exported as `@commonfabric/schema-generator/numeric-expression` (`./numeric-expression`).
  - Integrated into default recovery in `@commonfabric/schema-generator` and option evaluation in the transformer; the transformer also unwraps parentheses and type‑only assertions for all option types.
  - Tests/docs: goldens for `typeof CONST` number defaults and `numeric-option-defaults`; `wrapped-options` (replaces `parenthesized-options`) covers wrappers and nested cases; unit tests for `numeric-expression`; docs state numbers are recognized in all spellings and wrappers are transparent, and clarify that enum members and named constants in `toSchema` options are not evaluated — the `checker.getConstantValue()` fallback runs but returns `undefined`, so only literal values are supported.

<sup>Written for commit 9153997f05b2fb6881a019ceb047147ff678a3c4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4887?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/schema-generator/src/typescript/numeric-expression.ts
```diff
@@ -0,0 +1,71 @@
...
+ * leaves `-1` — the canonical sentinel default — and every non-finite value
+ * inexpressible. IEEE 754 binary64 values are first-class in the value model
+ * (`docs/specs/space-model-formal-spec/1-fabric-values.md` §1.3), so they must
+ * survive the trip from source to schema.
+ */
+export function numberFromExpression(
+  expr: ts.Expression,
+  checker: ts.TypeChecker,
+): number | undefined {
+  if (ts.isNumericLiteral(expr)) return Number(expr.text);
```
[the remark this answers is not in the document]
danfuzz: Valid — fixed in 0a7aea3fc. The gap turned out to be wider than the numeric
framing, so the fix is in a different place than suggested.

`evaluateExpression()` dropped **every** parenthesized option, of whatever
type. Confirmed against the emitted output:

| source | emitted before |
| --- | --- |
| `{ default: { n: (-1) } }` | `"default": {}` |
| `{ default: { n: (1) } }` | `"default": {}` |
| `{ description: ("hi") }` | property absent entirely |
| `{ description: "hi" }` | `description: "hi"` ✓ |

Note the third row: a plain parenthesized *string* was dropped too, and the
second is a bare literal that has nothing to do with the signed and non-finite
numbers this branch is about — it predates the branch. Unwrapping only inside
the numeric evaluation would have left both broken.

So the unwrapping goes at the top of `evaluateExpression()`, covering
parentheses and the type-only assertion forms (`as`, `satisfies`, `<T>x`) for
every value type at once. That also brings this extractor into line with its
`schema-generator` counterpart, which has always unwrapped them — the two
disagreeing was the underlying defect.

`numberFromExpression()` does learn to unwrap as well, but for a case the
caller-level fix cannot reach: parentheses nested *inside* the expression, as
in `-(1)`.

Coverage added:

- unit tests for `(1)`, `(-1)`, `((1))`, `-(1)`, `-(-1)`, `(-0)`, `-(0)`,
  `(NaN)`, `-(Infinity)`, and a shadowed global seen through parentheses;
- the `numeric-option-defaults` fixture gains a parenthesized default and a
  parenthesized *description*, so the non-numeric half is pinned too;
- the current-behavior spec now records that these wrapper forms are
  transparent.

ts-transformers `1105 passed`, schema-generator `45 passed`, full type check
clean.

-- Claude (Opus 4.8), acting on @danfuzz's behalf

ubik2:
There's a minor issue here that we're embedding these in JSONSchemas, where these values aren't allowed. That's fine in our own consumption (since we already have type unknown and undefined), but these will break when pointed against the llm-generator (where we provide our generated schema to them).

I think that's acceptable. It's not a new bug, just a new place it can show up. I also don't want to hold our own schema extensions back for a feature that is only used in a small subset of our patterns.

danfuzz:
@ubik2 Thanks for the analysis. And yeah I agree: On the assumption that _our_ `JSONSchema` will ultimately evolve into (something like) `FabricSchema`, this PR is either the chicken or the egg in that metaphor.

packages/schema-generator/src/typescript/numeric-expression.ts
```diff
@@ -0,0 +1,77 @@
...
+ * (`docs/specs/space-model-formal-spec/1-fabric-values.md` §1.3), so they must
+ * survive the trip from source to schema.
+ */
+export function numberFromExpression(
+  expr: ts.Expression,
+  checker: ts.TypeChecker,
+): number | undefined {
+  // Parentheses carry no meaning here, and they can appear anywhere -- `(-1)`,
+  // but also `-(1)`, which no amount of unwrapping by the caller would reach.
+  if (ts.isParenthesizedExpression(expr)) {
```
mathpirate: There’s one more place where wrapper transparency does not compose yet. The callers unwrap `as` / `satisfies` / `<T>x` only when the wrapper is the outermost expression, while recursive operands inside `numberFromExpression()` unwrap only parentheses.

Consequently these still evaluate to `undefined` in both extractor paths:

```ts
-(1 as number)
-(Infinity satisfies number)
-(<number>1)
```

This is structurally the same case as `-(1)`, which the helper explicitly handles because caller-level unwrapping cannot reach inside the prefix expression. It also conflicts with the updated documentation’s statement that the type-only assertion forms are transparent.

Could the initial recursive-unwrapping branch here include the same assertion nodes handled by `evaluateExpression()`?

```ts
ts.isParenthesizedExpression(expr) ||
ts.isAsExpression(expr) ||
ts.isTypeAssertionExpression(expr) ||
ts.isSatisfiesExpression(expr)
```

I’d add unit coverage for at least `-(1 as number)` and `-(Infinity satisfies number)`, ideally with one fixture exercising the integrated path. This should be a small extension of the existing fix.

packages/schema-generator/src/typescript/numeric-expression.ts
```diff
@@ -0,0 +1,77 @@
...
+  if (ts.isNumericLiteral(expr)) return Number(expr.text);
+
+  if (ts.isPrefixUnaryExpression(expr)) {
+    if (
+      expr.operator !== ts.SyntaxKind.MinusToken &&
+      expr.operator !== ts.SyntaxKind.PlusToken
+    ) {
+      return undefined;
+    }
+    const operand = numberFromExpression(expr.operand, checker);
```
mathpirate, line 74: Non-blocking composability question: should a supported enum constant remain supported when sign-prefixed?

`toSchema` currently recognizes `E.One` through the final `checker.getConstantValue()` fallback, but `-E.One` enters this branch, recurses into a `PropertyAccessExpression` that `numberFromExpression()` declines, and then the caller’s constant-value fallback sees the whole `PrefixUnaryExpression` and also returns `undefined`.

The behavior spec now lists both enum constants and sign-prefixed numbers as supported forms, so a reader could reasonably expect those features to compose. If that is intended, the helper could accept numeric `checker.getConstantValue()` results as another operand base case. If the intended contract is specifically signed literals and signed `NaN`/`Infinity`, clarifying that limitation in the spec would also resolve the ambiguity.

I don’t think this needs to block the core bug fix, but it seems worth deciding while the evaluator’s accepted language is being documented explicitly.

mathpirate (commented):
I read through both extraction paths, the new shared helper, the fixtures, and the preceding golden-harness work. The overall direction looks right to me.

In particular:

- Sharing numeric expression recovery between `schema-generator` and `ts-transformers` avoids another subtle divergence between the two evaluators.
- Preserving the distinction between an unshadowed global `NaN` / `Infinity` and a local binding is important; silently folding the latter would replace one class of wrong schema with another.
- Putting general wrapper unwrapping at the top of the transformer’s `evaluateExpression()` is the right level. The discovered parentheses bug affects strings, descriptions, arrays, and other option values—not merely signed numbers.
- The new Fabric-encoded goldens make the schema-generator regression visible without lossy leaf-by-leaf assertions, including the load-bearing sign on `-0`.

I left one actionable inline comment: type-only wrappers are transparent only at the caller’s outermost level, so they still disappear when nested inside a sign expression such as `-(1 as number)`. That is the same recursive-wrapper case already handled for `-(1)`, and it looks cheap to close in this PR.

I also left a non-blocking question about signed enum constants such as `-E.One`. The evaluator separately supports enum constants and unary signs, but those capabilities currently do not compose.

Separately, I agree with the existing discussion about non-finite values at the external JSON Schema boundary: it should not hold this fix back. Our internal schema/value model already extends JSON, and dropping a valid internal default during extraction is the wrong place to enforce a provider’s narrower transport.

I would, however, record a concrete follow-up for the LLM/provider boundary. #4904 makes the test carrier lossless; it does not change ordinary JSON serialization. Without an explicit boundary policy:

- `NaN` and `±Infinity` silently serialize as `null`;
- `-0` silently serializes as `0`;
- a provider or SDK may reject the schema—or worse, accept the altered default.

The provider-facing conversion should deliberately validate, reject, or strip unsupported default annotations and include cases for these values. I would keep that conversion outside the schema generator: internally we should preserve the Fabric value faithfully, and the narrower external consumer should own the narrowing.

Subject to the small nested-wrapper fix—or with that captured immediately as a follow-up—I’m comfortable with the PR.

mathpirate (approved):

danfuzz:
> I would, however, record a concrete follow-up for the LLM/provider boundary

What's your preferred medium for durably recording this?

packages/schema-generator/src/typescript/numeric-expression.ts
danfuzz: Agreed, and taken here rather than as a follow-up — fixed in ced115963.

`numberFromExpression()`'s initial unwrapping branch now includes the same
assertion nodes as `evaluateExpression()`, exactly as suggested:

```ts
if (
  ts.isParenthesizedExpression(expr) || ts.isAsExpression(expr) ||
  ts.isSatisfiesExpression(expr) || ts.isTypeAssertionExpression(expr)
) {
  return numberFromExpression(expr.expression, checker);
}
```

You were right that the documentation was already claiming this. That is the
part I find most worth noting: the spec sentence was written to describe the
caller-level unwrapping and quietly over-promised for the recursive case, so
the docs asserted a behavior no test covered. Now covered both ways —

- unit: `1 as number`, `-(1 as number)`, `-(1 as const)`, `(-1) as number`,
  `1 satisfies number`, `-(1 satisfies number)`, `-(0 as number)` (asserted
  with `Object.is`, since the sign is the point), `-(Infinity as number)`;
- fixture: the wrappers fixture gains `offset: -(1 as number)` and a
  `satisfies` inside an array element.

The tests were run against the un-fixed helper and observed to fail first.

The fixture is renamed `parenthesized-options` → `wrapped-options`, since it no
longer covers only parentheses.

-- Claude (Opus 4.8), acting on @danfuzz's behalf

packages/schema-generator/src/typescript/numeric-expression.ts
danfuzz: Good question, and checking it turned up that the premise does not hold: `E.One`
is **not** currently recognized either, so there is nothing for the sign to
compose with.

Same option slot, three values, against the emitted output:

| source | emitted |
| --- | --- |
| `{ maxProperties: 1 }` | `maxProperties: 1` |
| `{ maxProperties: K }` where `const K = 1` | dropped |
| `{ maxProperties: E.One }` | dropped |

So the `checker.getConstantValue()` fallback does not fire at this point in the
pipeline at all — a plain named constant fails identically to an enum member,
and the same holds nested inside `default`. I have not chased why; the likely
shape is that by the time `SchemaGeneratorTransformer` runs, the nodes it sees
no longer resolve against the original program's symbols, which would make that
fallback inert rather than merely narrow.

That makes this a spec bug rather than a composition gap, and the spec bug was
mine — I added "enum constants" to the accepted-forms list while documenting
the evaluator. Corrected in baf11f2c0: the list now says literal values only,
and records the inert fallback with the three cases above so the next reader
does not have to rediscover it.

I would rather not fix the fallback in this PR. It is a different defect with an
unclear cause, and making it work would be a genuine behavior change (named
constants would start reaching schemas that currently drop them) rather than
the repair of a silent loss. Happy to write it up as a follow-up.

-- Claude (Opus 4.8), acting on @danfuzz's behalf

mathpirate: sounds good to me, thanks for keeping the spec current!

docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md
```diff
@@ -1338,7 +1338,19 @@ Recognized call forms:
...
-2. evaluate literal options object
+2. evaluate literal options object — string, boolean and `null` literals,
+   object and array literals, and numbers in any of their spellings (bare
+   literal, sign-prefixed, or the `NaN` / `Infinity` globals, the latter only
+   where the name is not shadowed). Parentheses and the type-only assertion
+   forms (`as`, `satisfies`, `<T>x`) are transparent at any depth, including
+   nested inside a sign (`-(1 as number)`). A property whose value is none of
+   these is dropped from the options object.
+
+   The code carries a `checker.getConstantValue()` fallback intended to recover
```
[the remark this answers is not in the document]
danfuzz: Correct — fixed in 9153997f0. The fallback does fire; my wording claimed it
didn't, which conflated the outcome (value dropped) with a mechanism (branch
skipped) I hadn't actually checked.

I instrumented the branch to avoid guessing a second time:

```
[PROBE] getConstantValue(PropertyAccessExpression) = undefined   // E.One
[PROBE] getConstantValue(Identifier) = undefined                 // K
```

So the call is reached for both nodes and returns `undefined` for each — the
value is dropped downstream of a fallback that ran, not because the fallback was
bypassed. The spec now says exactly that, and explicitly does **not** assert why
the checker recovers no constant value at this stage, since I have not
established that.

-- Claude (Opus 4.8), acting on @danfuzz's behalf

PR #4894, opened 2026-07-22
List every live space entity in FUSE without loading values

Hixie:
The FUSE entities projection was derived from the default pattern's allPieces collection. Entities that existed in the space but were absent from that collection were therefore missing, and mounting the view synchronized space and piece data merely to enumerate directory names.

Add a capability-negotiated Memory v2 request that lists current, non-deleted, space-scoped entity identifiers under READ authorization. The engine query selects identifiers without selecting stored entity values, and the storage and piece layers expose that operation to FUSE.

Build the entities directory from empty identifier stubs before publishing the connected space. Defer space-cell, allPieces, and individual entity loading until callers enter the corresponding projections. Preserve compatibility with older servers, share concurrent connections, clean up failed discovery, keep reconnect probes identifier-only, and remove stale indexes when entities disappear.

Add engine, protocol, authorization, bridge, and real transport coverage. The mount regression records every server frame and verifies that listing all entity directories transfers both expected fids without transferring either entity's distinctive payload bytes.

---
## Summary by cubic
FUSE lists every live space entity under `entities/` by ID-only enumeration with snapshot-stable, server-capped pagination and exact lookup. Values load only when a named child is accessed, which speeds mounts and keeps reads stable.

- **New Features**
  - Memory v2: added `entity-id.list` (stable pages; legacy response bounded to one page) and `entity-id.exists`; flags `entityIdListing`, `entityIdPagination`, `entityIdLookup`. Client in `@commonfabric/memory` exposes `listEntityIdPage`, `listEntityIds`, `entityIdExists`.
  - Runner/Piece: storage adds `ensureSession`, `listEntityIds`, `listEntityIdPage`, `entityIdExists`; `PieceManager` supports `{ deferSpaceCellSync }`.
  - FUSE: prepares one per-handle virtual snapshot and reuses it across `readdir` continuations; exact ID lookup without listing; hydrates only named children; bounds entity-only projections with an LRU and pins lookups/opens; coalesces concurrent scans; keeps snapshots linear with reverse inode-to-name indexing; adds operation-wiring helpers.
  - Benchmarks/Docs/CI: added FUSE entity projection and Memory identifier listing benches, capture stderr diagnostics, and a new spec on lookup/enumeration.

- **Migration**
  - No action needed. Engine migration repairs the `head.op` column and older writers now fail closed. FUSE requires servers that advertise `entityIdListing` and `entityIdPagination`; paginated clients send an explicit limit, and the `entities/` view fails closed when pagination is unavailable. If you call `loadManager` directly, pass `deferSpaceCellSync: true`. Mounts stay read-only until every connected space re-authenticates via `ensureSession`.

<sup>Written for commit 820b4d5082100c1898c28c9c0dee99d6d7e4e07d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4894?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bfollington (commented):
_cf-review via OpenAI Codex (GPT-5), on behalf of @bfollington._

I dug into the performance concern here because the feature goal is useful: a caller that knows an entity ID should be able to address it through FUSE without making `allPieces` the source of truth. The identifier-only Memory request does achieve the important first property: mount and shallow `entities/` listing do not serialize entity values.

The benchmarks found four separate scaling behaviours, rather than one general "FUSE is slow" problem:

- Building the current empty-directory projection is linear but retained: 1k IDs took ~5 ms / 0.8 MiB V8 heap, 10k took 54–78 ms / 8.3 MiB, and 100k took 574–826 ms / ~77 MiB and 100k inodes. The 100k identifier response itself was ~5.5 MiB.
- With CFC annotations enabled, stub insertion is quadratic-like because every entry filters and re-sorts the parent annotation array: 1k IDs took 59–72 ms, 10k took 2.4–2.7 s, and 20k took 11.9–14.3 s.
- Before the directory-handle fix, every low-level FUSE `readdir` continuation prepared the directory again. For 100k IDs, three local refreshes repeated ~16.5 MiB of identifier responses before database, WebSocket, and kernel-copy costs.
- A recursive walk crosses the value boundary. Walking 1k entity directories made exactly 1k entity, 1k input, and 1k result reads; ~3 MiB of source JSON became ~65 MiB and 106k retained inodes. Shallow listing is value-free, but `find`/an indexer that descends into every stub can still load everything.

The original storage query was also doing one `revision` lookup per retained `head`, including tombstones and other scopes. A 1k-live fixture took ~0.25 ms with only 1k heads, but the former query took ~69 ms when the same result sat behind 99k tombstones. Entity payload size had little effect; lifetime head count was the problem.

I pushed a branch based directly on this PR head so none of this takes ownership away from the PR author:

https://github.com/commontoolsinc/labs/tree/ben/codex/pr-4894-entity-index-benchmarks

It is split into cherry-pickable commits:

1. `a6d1a75c7` — store current `op` in `head`, backfill old databases, add a partial covering live-ID index, assert the query plan, and add the storage benchmark. At 100k live IDs this was ~3.4x faster; at 1k live + 99k tombstones it reduced ~68.8 ms to ~0.185 ms. A 100k-head migration took ~0.17–0.29 s locally.
2. `70d26d71a` — track preparation per open FUSE directory handle. The first `readdir` prepares; continuation pages reuse it; a new `opendir` refreshes; failures remain retryable. This removes the repeated-full-list paging amplification.
3. `e262008ed` — add the FUSE projection benchmark and wire both new benchmarks into the scheduled benchmark artifact.
4. `f7bbc3a7e` — document exact lookup vs enumeration vs hydration, directory-handle semantics, crawler behaviour, storage-index rationale, benchmark results, acceptance invariants, and old-server behaviour beside the FUSE spec.

Verification on that branch: full FUSE package 315 passed (1 ignored), full Memory package 409 passed / 140 steps, `check-docs` passed all 520 blocks, fmt/lint/type checks passed, and combined benchmark JSON validated with 17 series.

Those commits fix the database/tombstone pathology and the repeated-list-per-FUSE-page pathology. They intentionally do not pretend the remaining product decisions are solved:

- Mount still fetches all IDs and retains one stub/inode per ID.
- A cold exact lookup for a missing stub still refreshes the whole ID list.
- IDs are traversable directories, so recursive crawlers can hydrate every value.
- `entity-id.list` is one unbounded response, and CFC annotation insertion remains quadratic for large enumerations.

My recommended next shape, if we want exact-ID access to stay independent of space cardinality, is:

1. Do not list IDs during `connectSpace`; publish the dynamic namespace immediately.
2. Resolve `entities/<id>` with a point `entity-id.exists`/targeted read, never a complete list.
3. Make enumeration explicit and paginated (`after`, server-capped `limit`, `expectedServerSeq`), with one logical enumeration per `opendir`.
4. Emit enumerated names from the directory handle without permanently creating every inode. Allocate a real node only on exact lookup.
5. Use public ID symlinks into a non-enumerable by-ID namespace, or another explicit boundary, so ordinary crawlers do not follow every entity value by default.
6. Hydrate on explicit content access and batch CFC entry annotations; add a bounded projection cache if hydrated subtrees remain retained.

The deterministic regression contracts should be: mount makes zero per-entity reads; exact lookup makes zero list requests and at most one targeted read; one directory traversal performs one paginated pass; metadata-only traversal performs zero value reads; and the storage plan never accesses `revision` or scans tombstones.

My recommendation is to take the two low-risk fix commits (or equivalent) before merge, retain the benchmarks/spec, and consciously decide whether the eager mount footprint and crawler hydration boundary are acceptable for the first version. The branch is deliberately structured so each part can be adopted independently.

Hixie:
Codex rejected the symlink plus hidden by-ID namespace because: "it would introduce a second path scheme and make behavior depend on whether each crawler follows symlinks. The implemented design preserves the existing `/entities/<id>/...` paths while keeping entity directories empty during enumeration. Only an explicit lookup of a named child hydrates the entity, so it provides the same crawler-safety boundary without additional namespace or symlink complexity."

It implemented everything else you suggested.

bfollington (changes requested):
_cf-review via OpenAI Codex (GPT-5), on behalf of @bfollington._

Thank you for the substantial update. The original database and eager-value problems are now fixed: mount makes no identifier or value request, enumeration is paginated and identifier-only, recursive directory opening transfers no entity/input/result values, the live-head index avoids revision joins and tombstone scans, and CFC construction is now approximately linear.

I am requesting changes because the committed benchmark does not exercise the kernel-reference lifetime that determines the worst-case FUSE memory footprint.

## Blocking: benchmark and bound the real lookup lifecycle

`entity-projection.bench.ts` calls `bridge.prepareLookup()` directly. Actual callbacks use `FuseOperationState.prepareLookup()`, which reserves a lookup reference until the kernel sends `forget`. The new unit test already demonstrates this distinction: retained lookups remain above the 128-root LRU limit.

Using the callback-state path on this head, 100,000 lookup replies without `forget` took 932 ms and retained 100,008 inodes / 100,000 entity children, +64.96 MiB V8 heap, and +169.6 MiB RSS. The committed benchmark reports 136 inodes because it bypasses that lifecycle. This is linear rather than quadratic, but it recreates the original cardinality-sized metadata footprint until the kernel forgets entries.

Please add separate regression series for:

- retained FUSE lookup replies without `forget`;
- lookup followed by immediate `forget`; and
- a real mounted recursive `find`, with daemon heap/inode state measured after traversal and after kernel cleanup.

Exposing cached roots, pinned roots, lookup/open references, and inode count through `.status` would make the mounted regression deterministic. If the real mount retains cardinality-sized state, keep kernel-referenced entity roots as lightweight virtual stubs and evict hydrated descendants independently.

## Blocking: concurrent handles multiply the complete snapshot

Every open `entities/` handle independently fetches and retains every identifier. Ten simultaneous handles over 100,000 IDs made 1,000 page requests, took 2.01 seconds, and retained +79.6 MiB V8 heap / +395 MiB RSS while the tree itself stayed at eight inodes.

Please coalesce concurrent scans per space, share immutable snapshots between compatible handles, and add a total snapshot-memory budget. A 1/10/100-handle benchmark at 10k and 100k IDs would make this regression visible.

## Improvement: value-only writes can starve enumeration

Pagination compares every page with global `Engine.serverSeq`. Any commit therefore aborts a 100-page listing even when live ID membership did not change, and FUSE maps that transient failure to `ENOENT`.

Use a generation tied to live-ID membership or a server-owned snapshot cursor. Add a test/benchmark that interleaves value-only updates during enumeration: value changes should allow the pass to complete, while add/delete should invalidate it with `EAGAIN` or `ESTALE`, not `ENOENT`.

## Contract cleanup

- Require `expectedServerSeq` whenever `after` is present. The server currently lets alternate clients splice pages from different snapshots.
- The Memory spec says an unpaginated request returns the complete list, but the server rejects results above 1,000.
- The path spec says unsupported listing leaves `entities/` empty, while implementation and tests fail enumeration closed.
- Rename the historical `stubs-*` benchmark keys now that they measure exact lookups; keeping the old dashboard identity creates a semantic discontinuity.

Measurements on the updated head otherwise look good: a single 100,000-ID handle used about 5.2 MiB heap and 5.5 MiB wire across 100 pages; the indexed 100,000-live storage query measured 16.2 ms; 1,000 live plus 99,000 tombstones measured 0.167 ms; and the live-ID index cost approximately 4.2 MiB per 100,000 short IDs on disk. Targeted FUSE and Memory tests passed and the PR CI is green.

The feature-level shape is now right. The remaining request is to make the benchmarks represent the production inode lifetime and to prevent callers from multiplying the full snapshot without a bound.

bfollington:
_cf-review follow-up via OpenAI Codex (GPT-5), on behalf of @bfollington._

Here is the concrete implementation path I recommend so the performance findings are actionable rather than just objections.

## 1. Kernel-retained lookup state

FUSE cannot discard an inode while the kernel still references it, so constant memory regardless of lookup references is not achievable. The fix is to make the retained unit much smaller:

- represent unresolved entities as compact virtual inodes containing only `{ space, entityId }`;
- synthesize root attributes and empty directory reads from that registry;
- promote to a normal `FsTree` projection only when a named child is accessed;
- evict controllers, subscriptions, content, annotations, and hydrated descendants even while the root inode remains referenced; and
- delete the virtual inode after the final `forget`.

Add three distinct benchmarks: retained lookup replies without `forget`, lookup plus immediate `forget`, and a real mounted recursive `find` with daemon heap/inode state sampled at peak and after cleanup.

## 2. Concurrent directory snapshots

Introduce one immutable, reference-counted entity snapshot per space and entity-set generation. Coalesce an in-flight scan, share the resulting entry array between compatible handles, and release it after the last `releasedir`. Retain only a small bounded number of generations and enforce a total snapshot-byte budget. Benchmark 1, 10, and 100 concurrent handles at 10k and 100k IDs.

## 3. Enumeration stability

Add an `entitySetSeq` that changes only when default-branch, space-scoped live membership changes: creation, deletion, or resurrection. Value-only set/patch operations should not advance it. Use this token for continuation validation instead of global `serverSeq`.

A value-only update between pages should allow enumeration to complete. Add/delete between pages should invalidate the snapshot and surface `EAGAIN` or `ESTALE`, not `ENOENT`.

## 4. Make unsafe protocol states unrepresentable

Require the snapshot token whenever `after` is supplied. A discriminated first-page/continuation request type is sufficient; an opaque `nextCursor` containing the key and generation is even safer. Also add an identifier-length or serialized-page-byte ceiling so the 1,000-ID cap is a real byte bound.

## 5. Bound hydrated state by weight

Keep the root-count LRU, but add deterministic budgets for projected content bytes, inode count, and annotation bytes. When a root is kernel-pinned, shed unreferenced hydrated descendants before retaining a full subtree. Benchmark a full cache across increasing payload sizes and shapes.

## 6. Compatibility and documentation cleanup

- Deprecate the aggregate `listEntityIds()` helper in favor of a paged or `AsyncIterable` API, or give it an explicit maximum.
- Correct the unpaginated-list and unsupported-server normative text.
- Rename `stubs-*` benchmark series to `exact-lookups-*` so dashboard history does not combine different workloads.
- Add first-page, continuation-page, exact-existence, and complete paginated traversal storage benchmarks with query-plan assertions.

### Minimum mergeable sequence

1. Add `entitySetSeq` and enforce it on continuations.
2. Share/coalesce per-generation directory snapshots and add a total budget.
3. Add production-lifecycle and real-mount crawler benchmarks.
4. Fix transient error mapping, protocol validation, docs, and benchmark names.
5. Use the real-mount result to decide whether compact virtual inodes must land immediately. If mounted `find` retains anything close to the measured 65 MiB per 100k lookup references, I would include virtual inodes before merge.

bfollington:
I implemented the concrete mitigations as a cross-fork PR based directly on this branch:

https://github.com/Hixie/commontoolsinc.labs/pull/3

Hixie can merge that PR into `fuse-list-all-space-pieces`, which will update this PR without transferring ownership.

It includes:

- `entitySetSeq`, so value-only writes do not invalidate enumeration while membership changes do;
- required continuation tokens and `EAGAIN` for membership races;
- 1,000-ID / 256 KiB page bounds and a 16 KiB individual-ID bound;
- coalesced immutable snapshots shared across handles;
- a 64 MiB total estimated resident-snapshot budget, including old generations pinned by open handles;
- `.status` current/high-water retention diagnostics;
- production lookup/forget, 1/10/100-handle, first/continuation/full traversal, exact-existence, and real-mounted `find` benchmark coverage;
- the protocol/path/FUSE documentation corrections and benchmark rename.

The new measurements make the tradeoff concrete:

- 1, 10, and 100 concurrent handles over 100k IDs all perform exactly 100 pages, transfer ~5.5 MiB, retain ~6.5 MiB heap, and take roughly 240–263 ms. The handle multiplier is gone.
- A 100k storage exact-existence hit is ~5.4 µs; first/continuation pages are ~165–168 µs; complete 100-page traversal is ~16.3 ms, excluding wire latency.
- The real callback lifecycle still shows the unavoidable kernel-pinned case: 100k lookup replies without `forget` retain 100,008 inodes and ~110 MiB V8 heap. With matching `forget`, it returns to 136 inodes / 128 roots. `.status` now exposes this directly, and the real-mount harness can run `--ops entity_find`.

The remaining design boundary is compact virtual inode roots / independently weighted hydrated-descendant eviction. Hard-evicting a kernel-referenced inode is not safe, so I did not disguise that with an invalid cap. The stacked PR keeps these roots identifier-only, makes peak retention measurable, and documents using the real mounted result to decide whether that larger representation change is required before merge.

Verification: 248 focused tests passed, all 520 documentation blocks passed, and the repository-wide 400-path type check passed.

bfollington:
I pared the follow-up down to one concrete improvement and closed the larger proposal. The replacement is https://github.com/Hixie/commontoolsinc.labs/pull/4 — 3 files, 68 insertions, 12 deletions. It coalesces overlapping `entities/` preparations per space so concurrent handles perform one paginated scan and share one immutable entry array; later opens still refresh normally. The focused regression test verifies one storage request and snapshot object identity. Full CellBridge suite: 80 passed; docs: 520 blocks passed.

Hixie:
I've merged that into this PR, and my agent approves of it. Let me know if you want the whole thing to land as-is, or with further changes!

bfollington (approved):
_cf-review via OpenAI Codex (GPT-5), on behalf of @bfollington._

Re-reviewed the latest head (`251d5f1a7`) after the branch refresh. The 18 previously reviewed feature commits are patch-equivalent; the only new functional patch is the concurrent entity-scan coalescing change.

No blocking findings. Overlapping `entities/` preparations now share one in-flight scan and one immutable snapshot array per space, cleanup runs after settlement and disconnect, and later opens still perform a fresh scan. Recursive directory traversal remains identifier-only and does not hydrate entity contents; exact projections are released through the lookup/forget lifecycle.

The exact PR tree passes all 80 focused CellBridge tests, including the new object-identity/request-count regression, and the current CI suite is green. I intentionally did not treat the main-branch merge conflicts as review findings. Approving.

PR #4895, opened 2026-07-22
refactor!: $alias is a binding form, not a link

ubik2:
## Summary

Continues the `$alias` retirement (audit: CT-1895 sibling work): `$alias` records are Pattern-object **bindings**, not links, and in data they are now plain values.

**Generic link recognition is sigil-only.** `isWriteRedirectLink`, `parseLink`, `isCellLink`, and `isPrimitiveCellLink` no longer match `$alias` records. Data-side consumers (storage traversal, write diffing, CFC target checks, redirect chasing) silently stop honoring `$alias` in stored data — keeping the predicates and parsers consistent, so an `isCellLink → parseLink` sequence can't crash on legacy data.

**Binding-side callers are explicit — and only where machinery actually writes through the binding.** Pattern machinery that resolves live bindings (`sendValueToBinding`, `unwrapOneLevelAndBindtoDoc`, the runner's output-binding policy recorders) checks `isAliasBinding(b)` alongside `isWriteRedirectLink(b)` and parses via `parseAliasBinding(b, base)` — no combined helper, so `isAliasBinding` is the single greppable entry point for every remaining legacy-form dependency. `parseAliasBinding` is **new** (extracted from `parseLinkPrimitive`'s former alias branch); it takes a required `NormalizedFullLink` base (all callers had one), and defensively resolves a non-`CellScope` scope in stored pattern JSON to the base's scope instead of leaking it through the claimed-full return type.

**Post-unwrap and stored-value consumers are deliberately sigil-only** (review round). The only `$alias` records that survive `unwrapOneLevelAndBindtoDoc` belong to *embedded Pattern values* (a pattern captured in a handler's `$ctx`, a builtin's `op`): their `defer` bookkeeping resolves them at that pattern's own instantiation, and interpreting them at the carrying node's level produces links at the wrong nesting depth. Accordingly: `findAllWriteRedirectCells` now skips embedded `isPattern` values entirely — and as of the third commit no longer parses `$alias` at all: its one pre-unwrap caller (the cold-resume prefetch) now unwraps node bindings first, so a surviving `$alias` is always a nested level's wiring (see the third-commit section below); the pattern skip is what keeps it from misreading embedded-pattern bindings (it also stops collecting an embedded pattern's *sigil* links as the carrying node's declared reads); `collectArgumentSchedulerReadLinks` and `resolveJavaScriptStreamLink` stay sigil-only (post-unwrap, live aliases cannot occur there); and `recordSetupProjectionPolicyInputs` is sigil-only on both callers, because its projections are *stored* (where `$alias` is inert) and the CFC prepare gate both verifies stored sigil redirects (`setupProjectionSourceMatchesValue`) and widens a trusted-initialization exemption from its markers (`writeIsPatternSetupInitialization`) — an alias-sourced marker would be unverifiable or an over-grant.

**Alias scope is a partialCause concern** (review round). `scope` moved from `AliasBindingBase` to `AliasBindingPartialCause`, typed `CellScope` — it names where the derived internal cell is minted; named-cell (`argument`/`result`) aliases never consumed it (the referenced cell's own link supplies scope at unwrap) and no longer serialize it. Note this changes serialized-pattern bytes for the same source: named-cell aliases previously embedded `scope: "space"`. `"inherit"` is not admitted anywhere in the alias types — it was never generated (the builder's `cell.export()` filters non-cell scopes, and has since scopes were introduced in #3500) and would mean the same as omitting the field.

**Clients mirror the runner.** `runtime-client`'s `parseAsCellRef` no longer hydrates `$alias` into `CellHandle`s, and the html renderer no longer blanks `$alias`-shaped records in text positions (they warn + render as JSON like any unexpected object) — both previously would have disagreed with the runner about what the value is.

**Names follow the semantics:** `isLegacyAlias` → `isAliasBinding`, `LegacyAlias*` types → `AliasBinding*`, `toJSONWithLegacyAliases` → `toJSONWithAliasBindings`. (`parseAliasBinding` is new extraction, not a rename — see above.)

**Link resolution drops its dead `$alias` probes** (second commit): `checkLegacyAt` and `readMaybeLink`'s `$alias` disjunct could no longer produce a hop, but still burned a storage read and a reactivity subscription per resolution fallback; the resolver's full-path fallback branch collapses to the parent sigil check.

**Alias bindings require a name or partialCause; the prefetch unwraps before walking** (third commit). `isAliasBinding` now matches only `$alias` records carrying a named cell (`"result"`/`"argument"`) or a `partialCause`; bare `{ $alias: { path } }` no longer matches anywhere. `parseAliasBinding` **throws** on partialCause aliases: a partialCause alias denotes a derived internal cell — a separate document minted from the result cell and the partialCause (`kind` participates in the id), in the alias's own `scope` — not a path within the base document, so parsing one against a base link is a caller bug; the throw stays as a permanent tripwire. Named-cell aliases resolve in the base's scope. The tripwire exposed that the cold-resume prefetch (`syncCellsForRunningPatternInner`) fed raw `node.inputs`/`node.outputs` straight into that parse, mis-resolving partialCause aliases *and* `cell: "argument"` aliases to the result document (pre-syncing the wrong cells). It now unwraps each node's bindings first (`unwrapOneLevelAndBindtoDoc` with the pattern's `derivedInternalCells`, per-node try/warn), and when the argument meta link is missing it skips the node walk entirely — instrumented with a `resume-pre-sync` warn to measure how often that happens — rather than defaulting to the result cell's link (CT-1897 tracks the same defaulting bug in `collectResumeOwnedCells`). With every caller post-unwrap, `findAllWriteRedirectCells` drops alias parsing: a surviving `$alias` (one that just crossed its `defer` boundary, or an embedded pattern's wiring) is a nested level's surface and is skipped.

**Live docs updated in the same change**: the space-model reference spec now states that link recognition is sigil-only and `$alias` survives only as Pattern-binding vocabulary — including a new caveat that a literal `$alias`-shaped object cannot be passed as factory inputs (pattern serialization has no escape encoding and reinterprets it as a binding); the action-identity spec follows the `toJSONWithAliasBindings` rename; the formal-spec migration table drops `$alias` (bindings are not links, and no replacement encoding is proposed yet).

Known follow-up (next phase of the audit): `memory/v2/sync-schema-table.ts`'s `$alias` payload rewriting — **already implemented in open PR #4774**; coordinate there rather than redoing it (whichever lands second conflicts in that file).

## Test plan

- Legacy-in-data pins flipped to the new semantics: `$alias` in cell data is returned as plain data by proxies, is not followed by `normalizeAndDiff`/`resolveLink`, creates no legacy `["$alias","path"]` reactivity probes, and `isWriteRedirectLink`/`parseLink` reject it; redirect-following coverage retained via sigil fixtures; binding parsing covered through `parseAliasBinding` (including defensive `"inherit"`-in-stored-JSON fallback; the former partialCause-scope-precedence pin now asserts the partialCause parse throw).
- Review-round additions: `CellHandle.deserialize`/update-delivery pins ($alias stays plain data, sigil control hydrates); html main renderer + worker reconciler JSON-render pins; `findAllWriteRedirectCells` pin for embedded Pattern values (skipped; sibling sigil links still collected — existing alias/defer pins retained); schema-declared stream-cell test de-vacuoused with location assertions.
- Full runner suite: 999 passed (5,347 steps) / 0 failed. Downstream: piece 21/0, html 43/0, runtime-client 48/0. Full workspace suite: all packages passing. Integration: 7/7 suites green (runner, runtime-client, shell, patterns, cli, generated-patterns, pattern-tests), with the home-rehydration reload-churn pin at its 0-conflict baseline.
- Third-commit round: the partialCause tripwire initially surfaced 113 failing tests, all tracing to the one raw-bindings caller (the cold-resume prefetch); after the unwrap-first rework the full runner suite is back to 999 passed (5,347 steps) / 0 failed, with the alias-fixture walker tests rewritten to the unwrap-first contract. `patterns-reload` integration re-run: the home-rehydration churn pin holds at its bound (2/2 passed).


packages/runtime-client/cell-handle.ts
```diff
@@ -441,8 +440,10 @@ export class CellHandle<T = unknown> {
   }
 
   /**
-   * Recursively hydrate any object, converting any links (SigilLink,
-   * LegacyAlias) into CellHandle instances.
+   * Recursively hydrate any object, converting any sigil links into
+   * CellHandle instances. Legacy `$alias` records are plain data — they are
```
ubik2, line 444: Please add a direct `CellHandle.deserialize` or update-delivery regression test showing that a `$alias` record remains plain data rather than becoming a `CellHandle`. The package suite currently passes without exercising this newly changed client boundary.

packages/html/src/render-utils.ts
```diff
@@ -326,15 +326,11 @@ export function stringifyText(
...
-    if (value && "$alias" in value) {
-      return "";
-    } else {
-      console.warn(
-        "unexpected object when value was expected",
-        value,
-      );
-      return JSON.stringify(value);
-    }
+    console.warn(
```
ubik2, line 329: The nearby comment still describes the removed unresolved-alias special case. Please update it to describe ordinary unexpected-object rendering, and add focused coverage that `$alias`-shaped data is JSON-rendered in both the main renderer and worker reconciler instead of becoming empty text.

packages/runner/src/link-types.ts
```diff
@@ -195,28 +197,35 @@ export function parseLinkPrimitive(
...
+ * ({@link parseLinkPrimitive}, `parseLink`) no longer accept them.
+ */
+export function parseAliasBinding(
+  value: AliasBinding,
+  base: NormalizedFullLink,
+): NormalizedFullLink {
+  const alias = value.$alias;
+  // Named-cell ("argument"/"result") and partialCause aliases carry no
+  // absolute id of their own here, so resolve to the base cell's document.
+  // An "inherit" (or absent) scope resolves to the base's scope.
```
ubik2: Please pin the new scope rule with a focused test: an alias whose scope is `inherit` must resolve to the supplied base scope. The existing parser tests cover an absent scope, but not the explicit `inherit` case fixed here.

ubik2 (commented):
_cf-review via Codex (GPT-5), on behalf of @ubik2._

Re-review at `593af64de`: no blocking correctness findings remain. The updated identity/reference specs now correctly describe `$alias` as Pattern-binding vocabulary rather than a link format.

Remaining non-blocking improvements:

- `docs/specs/space-model-formal-spec/1-fabric-values.md` says “Four legacy conventions,” but the revised table contains three.
- `packages/patterns/gideon-tests/style-object-repro/README.md` still names the removed `toJSONWithLegacyAliases` function twice.
- `packages/runner/src/cfc/ui-contract.ts` still says a legacy alias is a link leaf, although generic parsing now rejects it and traversal descends through the record.
- Please pin removal of the legacy resolver probes with a regression test showing that `$alias`-shaped plain data does not create a `$alias/path` reactivity observation.

Inline comments cover the runtime-client plain-data boundary, both HTML renderer paths, and explicit `scope: "inherit"` parsing coverage.

Verification: changed runner tests 19 suites / 434 steps; runtime-client 47 / 173; HTML 41 / 286; documentation 519 code blocks; `git diff --check` passed.

willkelly (commented):
We reviewed this against real wire data (full memory-protocol captures of Lunch Poll flows taken for the #4712 measurements), plus a code sweep. Overall it holds up well — evidence first, then one real issue and some small stuff.

**The captures support the core claim.** Across three full runs, zero `$alias` records appear in data positions — every occurrence is inside pattern binding structures. And the resolver fallback this PR deletes is demonstrably dead weight: all 110 legacy probe reads (the `["value","$alias","path"]` shape from `checkLegacyAt`/`readMaybeLink`) returned absence. Nice side effect: ~110 fewer read-set entries per comparable flow. Residual risk is bounded to docs written before the sigil conversion (~2025-06-26, #1251) and never rewritten since — a one-off scan of prod space DBs would close that completely; happy to share the scan script.

**Coordination on the known follow-up:** the `sync-schema-table.ts` `$alias` rewriting you flag as the next audit phase is already implemented in open PR #4774 — the mapper there stops interning alias schema positions, the storage-time validator keeps refusing reserved refs at those positions (clients deployed today still expand them), and expansion rejects any ref it doesn't interpret. Probably worth pointing the follow-up note there so it isn't redone (and whichever PR lands second would conflict in that file if it is).

**One real issue — two binding-side consumers didn't get the `isAliasBinding` pairing.** `resolveJavaScriptStreamLink` (runner.ts:3569) and `collectArgumentSchedulerReadLinks` (runner.ts:3370) use bare `isWriteRedirectLink` on the same post-unwrap `io.inputs` that `findAllWriteRedirectCells` (pattern-binding.ts:581) was explicitly patched to keep honoring — and `unwrapOneLevelAndBindtoDoc` can still legally emit residual `$alias` records (the defer-decrement branch at pattern-binding.ts:419-424, and the cell-less fall-through). Concretely: a residual `$event` alias skips the redirect loop and hits `describeHandlerStreamFailure` at instantiation instead of resolving, and with mixed inputs an alias target lands in `io.reads` but silently drops out of `schedulerReads`, so the handler goes stale when that input changes. Either these two sites want the pairing, or — if post-unwrap inputs can never contain aliases — the sites that were patched are dead code. Worth making it consistent one way or the other.

Small stuff:
- `packages/patterns/gideon-tests/style-object-repro/README.md` still cites `toJSONWithLegacyAliases` twice (live doc; outside what check-docs/check-skill-facts catch).
- The description's rename list includes `parseLegacyAlias → parseAliasBinding`, but `parseLegacyAlias` never existed — `parseAliasBinding` is new, extracted from `parseLinkPrimitive`'s alias branch. Worth fixing so future history greps don't dead-end.
- `schema-links.test.ts` "creates schema-declared stream cells for missing stream targets" (~826-882) now passes vacuously: its `$alias` fixture is plain data on this branch, and the assertions never check where the stream cell was created. Convert the fixture to a sigil redirect or assert the location.
- `cfc/ui-contract.ts` `collectContextLinks` comment says a link "(sigil or legacy alias) is always a leaf" — with sigil-only `parseLink` the walker now descends into alias envelopes on that (dormant) path.
- `piece/src/manager.ts` `getReadingFrom`/`getReadBy` doc comments still say "via aliases or links"; alias-based edges silently drop from piece graphs for old stored data.
- `html/src/render-utils.ts:328` and `worker/reconciler.ts:3836` keep "render empty until resolved" comments above code that now warns + JSON-renders.
- Several `$alias`-in-data tests were rewritten to sigil fixtures rather than inverted, so the new inertness has no direct negative pin beyond the cell-proxy read — a couple of explicit "stored `$alias` is not followed by `resolveLink`/`normalizeAndDiff`" assertions would lock the semantics in.

The `parseAliasBinding` required-base/`inherit`-scope change checks out — the old code really could leak `"inherit"` through the claimed-full return type (crashing later in `isFullNormalizedLinkShape`), and all six call sites pass a validated full link. Verified equivalent-or-better than the pre-PR runtime behavior.

ubik2:
Moved this back to draft, since the combination of cubic and Will's feedback raised some more complex questions

packages/runner/src/runner.ts
```diff
@@ -491,9 +505,11 @@ const recordSetupProjectionPolicyInputs = (
     return;
   }
 
-  if (isWriteRedirectLink(projection)) {
+  if (isWriteRedirectLink(projection) || isAliasBinding(projection)) {
```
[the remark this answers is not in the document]
ubik2: Fixed in 7b07ac360, with your suggested shape — `recordSetupProjectionPolicyInputs` is sigil-only on **both** callers (not just the argument path). Rationale beyond the data/binding distinction: its projections are *stored* (argument via `diffAndUpdate`, result via `setRawUntyped`), where `$alias` is inert on this branch, and the CFC prepare gate can only verify stored **sigil** redirects (`setupProjectionSourceMatchesValue`) while `writeIsPatternSetupInitialization` *widens* a trusted-initialization exemption from these markers — an alias-sourced marker would be unverifiable at best and an over-grant at worst. Comment at the site documents this.

packages/runtime-client/cell-handle.ts
ubik2: Added in 7b07ac360 (`packages/runtime-client/cell-handle.test.ts`): a `CellHandle.deserialize` pin and an update-delivery (`$onCellUpdate`) pin. Both feed a payload containing a top-level and a nested `$alias` record plus a genuine `link@1` sigil link as positive control — the sigil link hydrates into a `CellHandle`, the `$alias` records stay deep-equal plain data and fail `isCellHandle`.

packages/html/src/render-utils.ts
ubik2: Done in 7b07ac360. The stale unresolved-alias comments are rewritten to describe ordinary unexpected-object rendering (warn + JSON) in both `render-utils.ts` and the worker reconciler. Coverage added in three places: `stringifyText` unit pin (JSON output + exactly one warning), a `renderImpl` pin for `$alias`-shaped data in child position (JSON text, not empty), and a new `worker-reconciler-alias-child.test.ts` pinning the worker path emits a `create-text` op carrying the JSON serialization with no empty-text op.

packages/runner/src/link-types.ts
ubik2: Added in 7b07ac360, and the review discussion went one further: `scope` moved off `AliasBindingBase` to `AliasBindingPartialCause`, typed `CellScope` — named-cell aliases never consumed it (the referenced cell's link supplies scope at unwrap) and no longer serialize it, and `"inherit"` is inadmissible in the types (never generated: the builder's `cell.export()` has filtered non-cell scopes since scopes landed in #3500). `parseAliasBinding` keeps the defensive runtime fallback for untyped stored JSON, and the new test pins both that fallback (explicit `"inherit"` → base scope, via cast fixture) and partialCause scope precedence (`"user"` wins over base). Note the serialization consequence: named-cell aliases previously embedded `scope: "space"`; same-source patterns now serialize without it.

ubik2:
Review feedback addressed in 7b07ac360. Responses to the review-body items (@willkelly, @ubik2):

**The `isWriteRedirectLink`/`isAliasBinding` pairing inconsistency (willkelly's "one real issue")** resolved — in the direction neither option anticipated. Instrumenting `bindNodeIO` showed the only `$alias` records that survive `unwrapOneLevelAndBindtoDoc` are **embedded Pattern values'** bindings (e.g. a pattern captured in a handler's `$ctx`): their `defer` bookkeeping resolves them at that pattern's *own* instantiation, so interpreting them at the carrying node's level parses links at the wrong nesting depth. Accordingly:

- `resolveJavaScriptStreamLink` and `collectArgumentSchedulerReadLinks` stay **sigil-only**, now with comments stating the pairing is deliberately absent (the residual-`$event`-alias and mixed-inputs staleness scenarios cannot arise: `$event` is always builder-generated and unwraps to a sigil link).
- `findAllWriteRedirectCells` now **skips embedded `isPattern` values** — which also stops it collecting an embedded pattern's *sigil* links as the carrying node's declared reads (a sibling bug by the same logic). It **keeps** honoring `$alias` because its pre-unwrap callers (raw `node.inputs`, e.g. the cold-resume prefetch) walk wiring that *is* alias-form; removing the pairing empirically regressed the home-rehydration reload-churn integration pin from 0 to 2 commit conflicts, so it's load-bearing there. New test pins the embedded-pattern skip; the existing alias/defer pins are retained.
- Relatedly, `recordSetupProjectionPolicyInputs` (cubic's inline) is sigil-only on both callers — see the inline reply for the CFC prepare-gate rationale.

**Other items:**
- `schema-links.test.ts` stream-cell test de-vacuoused: sigil-redirect fixture plus assertions on the created stream cell's document and path.
- Negative pins added: stored `$alias` is not followed by `resolveLink` or `normalizeAndDiff`, is stored verbatim, and produces **no legacy `["$alias","path"]` reactivity probes** (ubik2's regression-test ask).
- `gideon-tests/style-object-repro/README.md`: both `toJSONWithLegacyAliases` references renamed.
- PR description fixed: `parseLegacyAlias → parseAliasBinding` removed from the rename list (`parseAliasBinding` is a new extraction from `parseLinkPrimitive`'s alias branch — never a rename); the sync-schema-table follow-up note now points at open **#4774** with the conflict warning.
- `cfc/ui-contract.ts` leaf comment, `piece/manager.ts` "via aliases or links" comments, and both renderers' "render empty until resolved" comments corrected.
- `1-fabric-values.md` "Four legacy conventions" → three.
- New docs caveat (identity-and-references spec): a literal `$alias`-shaped object cannot pass through pattern serialization as data — no escape encoding exists, so `toJSONWithAliasBindings` reinterprets it as a binding.
- Review round also moved alias `scope` to `AliasBindingPartialCause` (typed `CellScope`, no `"inherit"`) — details in the inline reply on `link-types.ts`; note it changes serialized-pattern bytes (named-cell aliases no longer embed `scope: "space"`).

Verification on the final tree: runner 999 suites / 5,347 steps, full workspace suite green, integration **7/7 suites green** with reload churn at its 0-conflict baseline.

ubik2:
I have some local changes to tighten the alias bindings so an `AliasBinding` requires either `cell` or `partialCause`. That's run up against some hiccups with existing tests, and I need to think about whether I want to preserve the tests (which are mostly about resolving aliases in contexts where they shouldn't appear), or eliminate them.
I'll update with that tomorrow.

danfuzz:
From the perspective of the `data-model`, this looks like a great step in the right direction.

It _does_ look to me like the new `AliasBinding` type, like its predecessor, remains ambiguous with respect to plain objects; that is, a user might decide to use `{ $alias: ... }` for their own reasons and end up confusing the system. As such, my hope — to be clear, presumably not to be done in this PR — is that ultimately `AliasBinding` will become a `class` per se and so be fully distinguishable from plain objects in the `data-model` and `runner`.

packages/html/test/worker-reconciler-alias-child.test.ts
```diff
@@ -0,0 +1,134 @@
...
+        textOps.some((op) => "text" in op && op.text === ""),
+        false,
+        "$alias child must not render as empty text (the old unresolved-alias special case)",
+      );
+      assertEquals(
+        warnings.length >= 1,
+        true,
+        "should warn about the unexpected object",
+      );
+      cancel();
```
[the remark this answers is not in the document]
ubik2: Fixed in 555cbf8be: the try/finally now spans the assertions, and the finally releases the mount (`cancel?.()`) alongside the `console.warn` restore — a failed assertion or a throw from `t.settle()` no longer leaks the subscription.

PR #4900, opened 2026-07-22
fix(runner): run the setup phase before instantiating a hot-swapped pattern

mathpirate:
## Why (the actual last mile of the home-space heal)

Estuary runs #4883 (and #4899 is deploying), yet bricked homes stay broken — with a **new** error:

```
[AppView] Failed to load space root pattern: Handler used as lift: node's $event
input resolves to of:fid1:…, which reads undefined — the { "$stream": true }
marker was never written there.
```

Reproduced end-to-end on live estuary (throwaway identity, home deliberately re-bricked with a `safeDateNow`-importing custom root, watched through the deployed build) and then in-harness with a full stack trace. The chain: the auto-update swap **works** — eligibility (#4883/#4899), probe, attested compile, CAS identity write all succeed — and then the `patternIdentity` **watcher re-instantiates the new pattern's nodes directly, skipping the setup phase** a fresh `start()` performs. Setup (`applySetupState`) is what materializes the incoming pattern's internal cells — handler stream markers included — and merges its argument-schema defaults. The old program had none of the new program's handlers, so the swapped-in `home.tsx` (handler-rich) dies at instantiation. **Every existing roll/swap test used a handler-less stub pattern**, which is why the suite never saw this.

## What

- The watcher now runs `applySetupState` for the incoming pattern **before** touching the running nodes, and tears the old nodes down only after that setup commits. A failed setup — e.g. the existing argument doesn't satisfy the new pattern's schema — logs `pattern-swap-setup-error` loudly and leaves the running pattern in place instead of a dead piece (fail-closed, same posture as the #4883 probe rules).
- Both swap branches (live in-memory artifact and async-loaded) funnel through the one `swapToPattern` helper.

## Verification

- New regression test: handler-bearing roll target (`SOURCE_V3_HANDLER`), replace an unloadable custom home root, then **start the replacement and assert it actually runs** (functional pin: `count` reads its initial value — under the bug, instantiation throws and the pattern body never executes; the swap-only identity assertions all still passed, which is why the pin is functional). **Red/green**: fails against the unfixed watcher, passes with the fix.
- Harness honesty fix surfaced by the new setup-validation: the suite's stub patterns declared a required no-default argument that fresh creation never validates but swap-setup correctly does — stubs now declare it optional (identity expectations are all derived at runtime; no goldens involved).
- Full runner suite **999 passed / 5342 steps / 0 failed** (the watcher's existing same-lineage consumers undisturbed); full piece suite 21/244; workspace typecheck, fmt, lint clean.

## Rollout

This is the third and, per the live repro, final layer of the fleet home-brick: #4873 opened the flag, #4883/#4899 made bricked roots eligible, this makes the replacement actually **run**. After merge + estuary deploy, my standing throwaway repro (a bricked home on estuary) validates the full heal with one reload before anyone's real home is on the line.

---
## Summary by cubic
Run the setup phase before instantiating a hot-swapped pattern so swaps actually start and don’t crash on handler nodes. This writes handler stream markers and default args to the reused root doc; on setup failure, we keep the current pattern running.

- **Bug Fixes**
  - `packages/runner/src/runner.ts`: Added `swapToPattern` to run `applySetupState` and commit before tearing down old nodes and instantiating the new pattern. Logs `pattern-swap-setup-error` on setup failure and aborts the swap. Unified live and async swap paths through `swapToPattern`.
  - `packages/piece/test/check-update-default-pattern.test.ts`: Made `items` optional in stubs. Added a handler-bearing roll target (`SOURCE_V3_HANDLER`) that verifies the swapped-in pattern actually runs (`count` starts at 0). Added a failure case to confirm incompatible swap-setup leaves the current pattern running.

<sup>Written for commit fcb8324915615f6d4bc97e9dc362fb408172838d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4900?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/piece/test/check-update-default-pattern.test.ts
+++ b/packages/piece/test/check-update-default-pattern.test.ts
@@ -23,12 +23,26 @@ const signer = await Identity.fromPassphrase("check update default pattern");
 const patternSource = (marker: string) =>
   [
     "import { pattern } from 'commonfabric';",
-    `export default pattern<{ items: string[] }>(({ items }) => ({ items, marker: "${marker}" }));`,
+    `export default pattern<{ items?: string[] }>(({ items }) => ({ items, marker: "${marker}" }));`,
     "",
   ].join("\n");
 
 const SOURCE_V1 = patternSource("v1");
 const SOURCE_V2 = patternSource("v2");
+// A roll target WITH a handler: handler nodes need their { "$stream": true }
+// markers materialized on the (reused) root doc — the dimension the plain
+// sources above never exercise, and the estuary post-swap failure class.
+const SOURCE_V3_HANDLER = [
+  "import { Writable, handler, pattern } from 'commonfabric';",
+  "const bump = handler<void, { count: Writable<number> }>((_, { count }) => {",
+  "  count.set((count.get() ?? 0) + 1);",
+  "});",
+  "export default pattern<{ items?: string[] }>(({ items }) => {",
+  "  const count = new Writable<number>(0).for('count');",
+  "  return { items, count, bump: bump({ count }) };",
+  "});",
+  "",
+].join("\n");
 
 const BUILD_SHA = "build-sha-1";
 const IMPORTED_MODULE_URL = "/api/patterns/system/update-marker.ts";
@@ -475,7 +489,7 @@ describe("checkAndUpdateDefaultPattern", () => {
     stub.setSource([
       "import { pattern } from 'commonfabric';",
       "import { marker } from './update-marker.ts';",
-      "export default pattern<{ items: string[] }>(({ items }) => ({ items, marker }));",
+      "export default pattern<{ items?: string[] }>(({ items }) => ({ items, marker }));",
       "",
     ].join("\n"));
     stub.setImportBuildSha("rolling-deploy-build");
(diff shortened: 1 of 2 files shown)


mathpirate:
Coverage round: the +1 was the `pattern-swap-setup-error` fail-closed branch, which no test drove once the stub-schema accident was fixed. Added the missing **contract pin**: a swap target whose argument schema rejects the existing argument logs the error and leaves the piece startable rather than dead (the fail-closed promise from the PR body, now executable). Suite green locally.

seefeldb (approved):

bfollington (commented):
Agent-assisted review (line-level findings verified against the diff and surrounding code). Overall: **LGTM — correct direction, safe to merge.** The diagnosis matches the live Estuary console evidence exactly, `swapToPattern`'s fail-closed posture is right, and I verified there's no double-write/corruption vector: `materializeDerivedInternalCells` default-seeding is guarded by `manifestMatch === -1` && `currentValue === undefined`, so re-running setup on an already-set-up doc never overwrites user data.

One caveat worth stating out loud, because it bounds what this PR heals:

**The cold-boot path is not covered, and it's plausibly the exact state of the currently-bricked Estuary spaces.** For a bricked space the piece is *not running* (its old root failed compile), so there's no `patternIdentity` watcher; `ensureDefaultPattern` swaps pre-start via `checkAndUpdateDefaultPattern` (bare `setMetaRaw`, "Never calls run()"), and the subsequent `startPiece` → `startCore` instantiates with no `applySetupState` — same missing-`$stream`-marker failure. The new test doesn't see this because `recreateDefaultPattern` runs the piece first, so the swap is consumed by the (fixed) watcher. Worse: a space whose pre-start swap already *committed durably* (identity current, root doc marker-less) will never re-enter the swap path at all. We're writing a repro test for both cases now and will stack a follow-up fix on this branch if confirmed — no need to hold this PR for it.

Smaller notes:
- cubic's P1 stands: `commit()` returns `Promise<Result<...>>` and never throws synchronously, so the try/catch only catches `applySetupState`'s sync validation throws; a rejected commit falls through to teardown+instantiate with setup state that never became durable. The realistic failure (incompatible argument schema) *is* the sync throw, so low practical risk, but the "only tear down after setup commits" claim isn't literal yet.
- On swap-setup failure, `checkAndUpdateDefaultPattern` still returns `"updated"` and durable `patternIdentity` names a pattern whose setup never ran; next cold boot computes `samePattern=true` and skips argument validation. Consider reverting the identity write (or probing schema compat before swap eligibility) in a follow-up.
- The fail-closed test only asserts `expect(after).toBeDefined()` — a pin that the old pattern still *functions* (e.g. marker value) would make it a real regression guard.

bfollington:
Follow-up on the cold-boot caveat from my review above — now settled empirically rather than speculatively:

**Confirmed: this PR alone does not heal an already-bricked space.** A test that reproduces the real bricked state (root not running when the swap lands, driven through the actual `ensureDefaultPattern` boot entry) fails on this branch with the exact live-Estuary error — `Handler used as lift: ... the { "$stream": true } marker was never written there` — thrown from `startCore`'s initial instantiation, which this PR doesn't touch (no watcher exists for a piece that never started). A second test covers the nastier case: spaces whose swap **already committed durably** under the deployed build (identity compares current, doc marker-less). Those never re-enter the swap path at all, so they'd stay bricked even after this lands.

The fix for both is stacked on this branch as #4926: a single fail-closed setup-repair attempt in `startEnsuredDefaultPattern` when the cold start fails. Piece-side only; this PR's watcher fix is untouched and still needed for the running-piece case.

**No reason to hold this PR** — the plan is to merge this and #4901 as approved, then land #4926 on top, so the next Estuary deploy carries the complete heal.

mathpirate:
Handing the remainder of this line to @bfollington — context below so nothing lives only in my head.

**Your cold-boot finding is confirmed independently at the call-graph level** (Fable verified against the branch): `setupInternal` — the only path through `applySetupState` — is reached from `run()` and the setsrc-update path, never from a cold `start()` of an existing piece, which resolves the stored identity and calls `instantiatePattern` directly. So the bricked-fleet flow (pre-start `setMetaRaw` swap → cold start) never runs setup, exactly as you diagnosed, and #4926 is the right shape for it. Your "worse" case is also our reality: the standing estuary repro space engaged its swap under the July-22 build, so it's likely sitting in the already-swapped/marker-less state — a live fixture for your test 2 (I'll hand you its credentials out-of-band).

Disposition of your smaller notes, which I'd like to land with you owning the branch:
- **`setupTx.commit()` un-awaited: confirmed at `fcb832491`** — the catch only sees `applySetupState`'s sync throws; an async commit rejection falls through to teardown+instantiate with setup never durable. The async-load branch just below models the epoch-guard shape an awaited commit needs (`active` / `startLifecycleEpoch` / `currentPatternKey` recheck after the await). Fine to fold into #4926 or a third stacked commit.
- **Identity write surviving setup failure**: agreed it's real, but note #4926 dissolves the wedge by construction (cold start heals marker-less docs regardless of how the identity got there) — so reverting the identity write demotes to a consistency nicety. I'd skip it rather than add a second mechanism.
- **`toBeDefined()` pin**: agreed, cheap — strengthen to a functional read whenever you're in the file.

Merging this as-approved; **deploy gates on #4926** (without it, a deploy heals only deploy-time-running pieces and the repro space would false-fail validation).

PR #4901, opened 2026-07-22
fix(patterns): tolerate stored profiles of any vintage — consumer-side compat view + data defaults

mathpirate:
## Why

Post-#4899, healed estuary homes died at start on June-era profiles: `profiles: 0: missing required property addExternalLink`. Streams are affordances a RUNNING profile provides — a stored doc written before a stream existed legitimately lacks it. Any consumer schema that requires such a field bricks on old docs at argument validation, first-absence-wins, so each fixed field just unmasks the next vintage's failure.

## What (per review — the consumer-side inversion)

The first push made the fields optional on `ProfileHomeOutput` itself. Per @seefeldb's review, that was the wrong side of the seam: **the producer type stays honest** (a live profile always binds every stream, so they stay required), and **consumers of stored docs take a weakened view**:

- **`BackwardsCompatibleProfile`** = `PartialBy<ProfileHomeOutput, …the seven post-genesis streams…>` (`setBio`, `addExternalLink`, `removeExternalLink`, `publishVerifiedIdentities`, `revokeVerifiedIdentities`, `addPiece`, `toggleEditing`; genesis streams — `setName`/`setAvatar`/`addElement`/`removeElement` — stay required since every doc has them; introduction dates in the doc comment). Adopted at every stored-doc seam: home's `profiles`/`defaultProfile`/`mru`, the `TrustedProfileLink`/`TrustedProfileList`/picker contracts in profile-create, profile-picker, and profile-embed's `ProfileResult`. profile-create's freshly-built profile keeps the strict type — current-vintage by construction.
- **The two post-genesis DATA fields ride `Default<>`** like their existing siblings (`externalLinks`/`verifiedIdentities`): `isEditing` (CT-1748) and — caught during this rework — **`bio`** (2026-06-17), which was required-with-no-default and was the *next brick in line*, masked only because `addExternalLink` sorts earlier in the required check.

Data-vs-affordance split in one line: data fields get defaults (validation fills), streams get consumer-side optionality (consumers `?.send`).

## Verification

- **Emission** (`cf check --show-transformed`): the `BackwardsCompatibleProfile` $def's `required[]` = `[$NAME, $UI, name, avatar, bio, externalLinks, verifiedIdentities, elements, setName, setAvatar, addElement, removeElement, isEditing, initialNameApplied]` — no late stream; `bio`/`isEditing` carry defaults; CFC `ifc` annotations stay in place. (Plain `?:` via `PartialBy` — a `| undefined` union would relocate annotations into `anyOf` branches; `profile-owner-cfc` pins that.) A generator probe confirmed `Omit & Partial<Pick>` emits correctly before adopting it.
- **Contract pin**: `profile-home.owner-gated.test.ts` now asserts the compat union contains all seven late streams and that `bio`/`isEditing` carry `Default<>` — a future stream added to `ProfileHomeOutput` but not to the compat list goes red.
- **Suites**: profile-home 15/15, profile-embed 9/9, profile-create 8/8 (lane-2); full runner 999 (5342 steps) incl. profile-owner-cfc; patterns lane-1 59; full workspace check, fmt, lint clean.
- Second commit: the coverage-gate deflake (chat thread-order tiebreaker test) — unrelated to the heal, kept separate and cherry-pickable.

## Follow-ups (not this PR)

- Standardize the compat-view idiom (seefeld → Hixie): differently-named backwards-compatible export per consumer-facing type.
- Auto-updating non-default system-derived patterns (who triggers, ownership) — the other half of the review discussion.

### DIFF
--- a/packages/patterns/cfc-group-chat-demo/logic.test.ts
+++ b/packages/patterns/cfc-group-chat-demo/logic.test.ts
@@ -1,0 +1,70 @@
+import { describe, it } from "@std/testing/bdd";
+import { expect } from "@std/expect";
+import { sortDisplayMessages } from "./logic.ts";
+
+// Thread order must be deterministic for messages that land in the same
+// millisecond — the tiebreaker chain (author → body → origin) is the contract
+// that keeps two replicas rendering one thread in one order. The pattern tests
+// exercise these branches only when the harness happens to send two messages
+// inside a single millisecond, so this pins the semantics with fixed
+// timestamps instead of a clock race.
+
+type ThreadMessage = {
+  authorName: string;
+  body: string;
+  origin: "sent" | "imported";
+  timestamp: number;
+};
+
+const msg = (
+  authorName: string,
+  body: string,
+  origin: ThreadMessage["origin"],
+  timestamp: number,
+): ThreadMessage => ({ authorName, body, origin, timestamp });
+
+describe("sortDisplayMessages thread order", () => {
+  it("orders by timestamp before anything else", () => {
+    const later = msg("alice", "first alphabetically", "sent", 2_000);
+    const earlier = msg("zed", "zzz", "sent", 1_000);
+    expect(sortDisplayMessages([later, earlier])).toEqual([earlier, later]);
+  });
+
+  it("breaks same-millisecond ties by author name", () => {
+    const bob = msg("bob", "hi", "sent", 1_000);
+    const alice = msg("alice", "hi", "sent", 1_000);
+    expect(sortDisplayMessages([bob, alice])).toEqual([alice, bob]);
+  });
+
+  it("breaks same-author same-millisecond ties by body", () => {
+    const second = msg("alice", "beta", "sent", 1_000);
+    const first = msg("alice", "alpha", "sent", 1_000);
+    expect(sortDisplayMessages([second, first])).toEqual([first, second]);
+  });
+
+  it("breaks identical-content ties by origin", () => {
+    const sent = msg("alice", "hi", "sent", 1_000);
+    const imported = msg("alice", "hi", "imported", 1_000);
+    expect(sortDisplayMessages([sent, imported])).toEqual([imported, sent]);
+  });
+
+  it("yields one thread order from any arrival order", () => {
+    const a = msg("alice", "alpha", "sent", 1_000);
+    const b = msg("bob", "beta", "imported", 1_000);
+    const c = msg("bob", "gamma", "sent", 1_000);
+    const d = msg("carol", "delta", "sent", 2_000);
+    const sorted = sortDisplayMessages([a, b, c, d]);
+    expect(sortDisplayMessages([d, c, b, a])).toEqual(sorted);
+    expect(sortDisplayMessages([c, a, d, b])).toEqual(sorted);
+  });
+
+  it("does not mutate its input", () => {
+    const input = [
+      msg("bob", "hi", "sent", 2_000),
+      msg("alice", "hi", "sent", 1_000),
+    ];
+    const before = [...input];
+    sortDisplayMessages(input);
+    expect(input).toEqual(before);
+  });
+});
--- a/packages/patterns/system/home.tsx
+++ b/packages/patterns/system/home.tsx
@@ -18,7 +18,7 @@ import {
   type TrustedProfileMru,
 } from "./profile-create.tsx";
 import ProfilePicker from "./profile-picker.tsx";
-import type { ProfileHomeOutput } from "./profile-home.tsx";
+import type { BackwardsCompatibleProfile } from "./profile-home.tsx";
 
 // Types from favorites-manager.tsx
 type Favorite = {
@@ -177,10 +177,14 @@ export default pattern<Record<string, never>, HomeOutput>((_) => {
   // space. `profiles` is the durable list (appended on create). `defaultProfile`
   // is the one `#profile` resolves to in headless mode and orders first in the
   // picker; `mru` is the recency-ordered list driving the rest of the ordering.
-  const profiles = new Writable<ProfileHomeOutput[]>([]).for("profiles");
-  const defaultProfile = new Writable<ProfileHomeOutput | undefined>(undefined)
+  const profiles = new Writable<BackwardsCompatibleProfile[]>([]).for(
+    "profiles",
+  );
+  const defaultProfile = new Writable<BackwardsCompatibleProfile | undefined>(
+    undefined,
+  )
     .for("defaultProfile");
-  const mru = new Writable<ProfileHomeOutput[]>([]).for("mru");
+  const mru = new Writable<BackwardsCompatibleProfile[]>([]).for("mru");
   // Untrusted-write regression surface: this stream is exported so tests can
   // verify that sending it from outside the trusted create surface does NOT
   // create a profile. The actual create UI lives in the profile picker below.
--- a/packages/patterns/system/profile-create.tsx
+++ b/packages/patterns/system/profile-create.tsx
@@ -11,7 +11,10 @@ import {
   Writable,
   WriteAuthorizedBy,
 } from "commonfabric";
-import ProfileHome, { type ProfileHomeOutput } from "./profile-home.tsx";
+import ProfileHome, {
+  type BackwardsCompatibleProfile,
+  type ProfileHomeOutput,
+} from "./profile-home.tsx";
 
 // Trusted UI surfaces / actions. The create surface authorizes appending a new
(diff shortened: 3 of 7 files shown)


mathpirate:
cubic P1: valid, taken, and extended. `addPiece` and `toggleEditing` are now optional like the other nine — and the two required **data** fields queued behind them (`isEditing`, `initialNameApplied`, which stored profiles predating CT-1748 also lack) are optional too, so validation can't just march to the next absence.

Verification notes from the round, for the record:
- Emission-checked on **home.tsx** (the consumer whose validation bricks): the top-level argument's `TrustedProfileList` projection is narrow (`required: ["name"]` — never required streams), and the sub-piece argument defs are now stream-free with `["…","undefined"]` unions on the optional fields.
- Twelve full-shape `required[]` instances remain in home's emission — these are **provision-side** schemas (what an instantiated profile *returns*, which is legitimately everything) rather than validation-path schemas; their property emissions carry the new undefined-unions while their requireds reflect provision. @seefeldb — flagging so you can confirm that reading; if any of those twelve sits on a stored-doc validation path we'd want to know.
- An interesting generator observation en route (separate item, not this PR): with `Default<>` on the two data fields, one emitted variant re-required the optionalized streams; with plain `?: … | undefined` it does not. Worth a look under the affordances/schema-generator design item.

Profile suites 15/15 + 9/9, workspace typecheck clean. The red Runner Tests (3/5) shard (`profile owner CFC policy`) passes in isolation locally on this branch — requesting a re-run as the parallel-load flake family unless it recurs.

mathpirate:
**Correction superseding my previous comment** — two findings from finishing the round properly:

1. **The Runner Tests (3/5) red was REAL, not flake**: `profile-owner-cfc.test.ts` caught that the `| undefined` union shape wraps each emitted property schema in `anyOf` and **buries the CFC `ifc` annotations** (owner-protection / requiredIntegrity) one level down — a genuine enforcement-posture hazard, exactly what that suite guards. The fields are now plain `?:` — stored docs predating a field read *absent-key*, which optionality alone covers; no union needed, property schemas (and their CFC annotations) keep their shape. CFC suite 18/18 green on the final content.
2. My earlier emission notes ('12 provision-side requireds', the `Default<>` resurrection observation) were measured on a **polluted working tree** (a branch slip mid-round, since untangled) — disregard them. On the true branch: home.tsx's emission has **zero** stream-requiring `required[]`s, zero `anyOf` wrappers, the argument-side `TrustedProfileList` projection was already narrow (`required: ["name"]`), and cubic's P1 fields (`addPiece`, `toggleEditing`) plus the two late-added data fields (`isEditing`, `initialNameApplied`) are all optional.

Final verification: profile-home 15/15, profile-embed 9/9, profile-owner-cfc 18/18, workspace typecheck/fmt/lint clean, emission checks on both profile-home and home. One durable takeaway for the schema-generator design item: **optionality via union changes CFC annotation placement** — worth capturing when the streams-never-required work happens.

mathpirate:
Amend delta (`11390826a` → `f00799a02`): the one red CI test — `profile-home.owner-gated.test.ts` asserts the addPiece Stream exposure against profile-home.tsx **source text**, and still expected the required spelling. Updated to `addPiece?: Stream<…>` with a comment making the optionality itself part of the pinned contract (a revert to required goes red there now). Swept the repo for other source-text assertions on the 13 fields' old spellings — none (the one other textual match is an unrelated inline fixture in `inspace-child-owner-seed.test.ts`). Lane-1 patterns suite 58/58, fmt/lint clean. No pattern-source changes.

mathpirate:
**Coverage-ratchet red: root-caused and fixed** (not this PR's diff). Downloaded the coverage artifacts from this PR's run and the main-baseline run (`094356caf`) and diffed per-file with the gate's own aggregation: the +4 is `packages/patterns/cfc-group-chat-demo/logic.ts:128-131` — the **author tiebreaker** in `compareMessagesByThreadOrder`, which only executes when two chat messages land in the **same `Date.now()` millisecond** during the in-harness pattern tests. Whether that collision happens is a clock race, so patterns coverage flaps ±4 run-to-run — the same flap explains the 237631 samples on unrelated PRs (#4530/#4870/#4874/#4875) in the gate's baseline breakdown.

`c81a7c14b` adds `logic.test.ts`: deterministic fixed-timestamp coverage of the full tiebreaker chain (author/body/origin) plus the arrival-order-independence contract those tiebreakers exist for. Verified in generated lcov: lines 128-131 and 133-138 all hit. Post-fix patterns debt = 237622 (5 under baseline) independent of clock luck — deflakes this gate for every future PR, not just this one.

mathpirate:
Reworked per the review — the consumer-side inversion is implemented (`3d8ba5700` + deflake commit, PR body rewritten to match).

@seefeldb three things you'll want to know beyond the mechanical swap:

1. **`PartialBy` emits correctly** — probed the generator before adopting: `Omit & Partial<Pick>` narrows `required[]`, keeps every property shape (streams keep `asCell: ["stream"]`), and leaves CFC `ifc` annotations in place. Your exact proposal, no generator gap.
2. **The list is dated, not guessed**: genesis (#3762, 2026-05-31) had `setName`/`setAvatar`/`addElement`/`removeElement` + `initialNameApplied`, so those stay required everywhere; the seven post-genesis streams are the PartialBy keys (dates in the doc comment). The owner-gated contract test pins the union — a future stream missing from it goes red.
3. **`bio` was the next brick in line**: added 2026-06-17 as required-with-no-default (unlike `externalLinks`/`verifiedIdentities`), only masked because `addExternalLink` sorts earlier in the required check. It now rides `Default<"">`, and `isEditing` rides `Default<false>` — data fields default, streams go optional in the consumer view.

One boundary judgment worth your eyes: `TrustedProfileLink`/`PickerProfileLink`'s `Cell<…>` targets went compat (stored-list elements are any-vintage — that's the seam that actually validated old docs), while profile-create's freshly-built profile keeps strict `ProfileHomeOutput`. If you'd rather the trusted-link contracts stay strict and tolerance live elsewhere, say the word.

seefeldb (approved):

bfollington (commented):
Agent-assisted review (line-level findings verified against the diff and surrounding code). Overall: **LGTM — safe to merge.** The consumer-side inversion is the right design: producer `ProfileHomeOutput` stays strict, stored-doc consumers take `BackwardsCompatibleProfile` with the seven post-genesis streams optional and the two post-genesis data fields on `Default<>`. The data-vs-affordance split is principled, the introduction dates are documented rather than guessed, and `profile-embed` was already written null-safe (`setBio?.send` etc.), so the weakened view introduces no new undefined-dereference sites. The regression pin on the compat union (future stream added to the output but not the PartialBy list goes red) guards the exact re-brick vector.

Two operational notes:

1. **Deploy coupled with #4900 (same deploy or after).** This PR edits `home.tsx`, so deploying it rolls the default pattern fleet-wide through the very swap path #4900 fixes. Merging both together (as planned) satisfies this.
2. **Corruption-masking trade is real but bounded:** a doc that genuinely lost its late streams now validates silently, indistinguishable from an old vintage. Genesis streams stay required so gross corruption still fails loudly, and streams are runner-materialized affordances (recoverable by a pattern roll), not user data. Acceptable for a fleet heal.

One open question, non-blocking: this rewrites the sources of `submitProfileCreation` / `setDefaultProfile` / `setMruProfile` in `profile-create.tsx`, changing the module identities that `writeAuthorizedBy` verification anchors on. Worth confirming stored `TrustedProfileList`/`TrustedProfileLink` claims stamped under the old identities still verify after this lands — and whether the answer depends on #4871 landing before or after.

mathpirate:
Answering the open question before handing this line to @bfollington:

**The `writeAuthorizedBy` re-anchoring concern is real but not new to this PR — it's the already-tracked #4772 family.** Verification anchors on content-addressed `moduleIdentity` + `bindingPath` at write time, checked against the *current* pattern's binding — schema and handlers roll together, so a source rewrite is self-consistent, which is why routine deploys (which shift `moduleIdentity` constantly) don't mass-brick owner-protected writes. The residual seam is stored-claim **reconciliation across writer-identity spelling shapes**, which is exactly the bio-save shear filed as #4772 — discovered, fittingly, on the profile bio path — with #4871 ("claims born stamped") as the in-flight hardening. Net: this PR adds no exposure class beyond any deploy, and #4871's landing order is orthogonal to its safety. Worth a joint look at #4772 if you haven't seen it — your instinct here and that issue are the same animal.

On your operational notes: agreed on both, and the deploy coupling is now three-way — **#4900 + this + #4926 ride one estuary deploy** (this PR edits home.tsx, so the deploy rolls every home through the swap path #4900 fixes; #4926 covers the cold/already-swapped fleet).

PR #4904, opened 2026-07-22
test(schema-generator): encode schema goldens with the fabric JSON encoding

danfuzz:
Changes the schema fixture goldens from plain JSON to the fabric JSON encoding,
so they can hold the values the generator is allowed to produce. Test code
only; no fixture files change.

## Why

`JSON.stringify()` cannot represent every `FabricValue`, and the schema
generator is free to produce ones it cannot. Worse, it mostly does not refuse
them — it substitutes quietly, rendering `-0` as `0` and `NaN` and the
infinities as `null`. (A bigint, which schemas will carry before long, is the
exception that throws outright.) The set of such values grows as the value model
does; the enduring point is that JSON is not a faithful carrier for it.

A golden that cannot hold a value cannot guard it, and this failure is silent
rather than loud — the flattened golden and the buggy output agree, so the
fixture passes while the behavior is wrong. That is not hypothetical: it is
exactly why the sibling fix for signed and non-finite schema defaults (#4887)
had to assert leaf-by-leaf in hand-written tests instead of adding a fixture.

`normalizeSchema` destroyed the same values even earlier, in a
`JSON.parse(JSON.stringify(...))` clone whose only purpose was to permit a
`delete clone.$schema` on a copy.

## What

- Goldens carry the tagged form (`/SpecialNumber@1`, and `/BigInt@1` when the
  time comes), pretty-printed. The `fvj1:` prefix is not part of the JSON and
  so cannot survive pretty-printing; it lives in the harness and is re-attached
  on decode.
- Comparison decodes the golden and uses `valueEqual` — the value model's own
  equality — instead of a structural compare that lets `-0` pass as `0`.
- The clone is deleted. `deepCanonicalize` already rebuilds every object and
  array it returns, so stripping `$schema` by destructuring leaves the caller's
  input untouched without one.
- Key sorting is deleted from the canonicalizer: a conforming encoder emits
  plain-object keys in canonical order ([JSON encoding spec §10][spec]), in
  UTF-8 byte order — where the hand-rolled sort used JavaScript's UTF-16
  default. What remains is only the JSON-Schema-specific array normalization
  the encoding has no business knowing about (`required` and `enum` denote
  sets; a two-member nullable `anyOf` reads better with the null arm first).

[spec]: ../blob/main/docs/specs/space-model-formal-spec/3-json-encoding.md

## Testing

- **No fixture changes.** All 73 goldens re-encode byte-identically, so this is
  purely a change of instrument. Verified two ways: the suite passes against
  the un-regenerated goldens, and a full `UPDATE_GOLDENS=1` run leaves the
  working tree clean.
- Two new tests cover what the fixtures cannot yet reach — producing these
  values from a fixture needs #4887, which is not merged:
  - a round-trip of `-0` / `NaN` / `±Infinity` / a bigint through the golden
    encoding, asserted with `Object.is`;
  - an executable statement of what plain JSON does to those same values, so
    that if it ever stops being true this indirection can go.
- The round-trip test was run against the previous plain-JSON harness and
  observed to **fail** — while all 34 fixture tests still passed, which is the
  silent-agreement problem demonstrated in one run.
- `schema-generator` — `ok | 36 passed (282 steps) | 0 failed`
- `./tasks/check.sh` (full repo) — `Type check complete.`

Depends on #4903 (merged), without which `schema-generator` could not import
`data-model` at all, and on #4905 (merged), whose
`JsonEncodingContext.{wrap,unwrap}EncodedValueForTesting()` this harness uses to
take the prefix tag off and put it back — so the tag is not spelled out here.

---
## Summary by cubic
Switch schema test goldens to the fabric JSON encoding to preserve -0, NaN, ±Infinity, and future bigints. Comparisons now use value-level equality; no fixture files changed because all goldens re-encode identically.

- **Bug Fixes**
  - Encode goldens with the fabric JSON format and decode for comparison; use `valueEqual` from `@commonfabric/data-model/fabric-value`.
  - Use `@commonfabric/data-model/codec-json` (`jsonFromValue`/`valueFromJson`) and `JsonEncodingContext.wrapEncodedValueForTesting`/`unwrapEncodedValueForTesting` to add/remove the prefix tag without hardcoding it.
  - Added tests proving fabric encoding preserves special numbers and bigints, while plain JSON loses them.

- **Refactors**
  - Removed the `JSON.parse(JSON.stringify(...))` clone in `normalizeSchema`; drop `$schema` via destructuring and rely on `deepCanonicalize`.
  - Deleted object key sorting; rely on the encoder’s canonical key order. Only normalize `required`, `enum`, and nullable two-arm `anyOf` arrays.
  - Clarified comments to state the enduring claim: plain JSON cannot represent every `FabricValue`.

<sup>Written for commit db332abff37bc9c5d6d35a9b373c46bf835224e7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4904?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mathpirate:
One follow-up for the harness’s broader “every `FabricValue`” goal: the codec is lossless, but the normalization surrounding it is not yet lossless for Fabric special objects.

Both `deepCanonicalize()` and `normalizeArrayOrdering()` recurse into every non-null object using `Object.entries()`. Fabric primitives such as `FabricBytes` keep their state in private fields and intentionally have no enumerable own properties, so they normalize to `{}` before `valueEqual()` sees them.

That can recreate the same silent-agreement failure this PR is designed to eliminate:

1. an actual special-object default is reduced to `{}` by `normalizeSchema()`;
2. the golden decodes to the correct Fabric object;
3. `normalizeArrayOrdering()` reduces the decoded object to `{}`;
4. `valueEqual({}, {})` passes.

The existing encoding round-trip test would not catch this because it calls `encodeGolden()` / `decodeGolden()` directly and bypasses both normalizers.

The values motivating #4887—special numbers—and the anticipated bigint case are primitives, so this does not undermine the current fix. For the future-proofing claim, though, I’d suggest a follow-up that:

- recurses only into arrays and plain objects, treating Fabric special objects as leaves;
- updates both normalization walks consistently; and
- adds a regression using something concrete such as `FabricBytes`, exercising the complete normalization-and-comparison path rather than only the codec round trip.

This feels like harness hardening rather than anything to revisit in the merged implementation.

PR #4906, opened 2026-07-22
fix(runner): restore Error.isError across SES lockdown

danfuzz:
## The gap

SES's error taming rebuilds the `Error` constructor from scratch rather than patching the original, copying forward only `length`, `prototype`, and the V8 stack surface (`stackTraceLimit`, `prepareStackTrace`, `captureStackTrace`). `Error.isError` is not among them, so `lockdown()` left the **whole realm** without it — host code included, not just compartments.

Silent until something calls it, at which point it throws `TypeError: Error.isError is not a function`. Which is exactly how it surfaced: data-model's `tagFromNativeValue` reaches for `Error.isError` to classify values whose constructor is unreachable (severed prototype, another realm), and the throw took pattern setup down with it.

**No `ses` release fixes this.** Verified through 2.2.0 and master: `tame-error-constructor.js` never copies the method. `permits.js` *does* list `isError` under `%InitialError%`/`%SharedError%`, but a permit only lets a property survive — it never adds one. A version bump is a dead end.

## The fix

`lockdown(options)` is exactly `repairIntrinsics(options)` followed by `hardenIntrinsics()`, and ses exposes both phases as globals. Between them the error constructors exist and are still extensible — the only seam where both are true. So:

```
sanitizeLocaleMethods();   // existing vetted shim
repairIntrinsics(options);
restoreErrorIsError();     // new vetted shim
hardenIntrinsics();
```

We restore the **genuine intrinsic**, captured at module evaluation before any lockdown. It's a pure predicate over the `[[ErrorData]]` internal slot: no powers to withhold, no realm affinity, and correct for the tamed constructors' instances because those are still built from the original `Error`. That's also why it beats an `instanceof` polyfill — a compartment's `Error` is not the host's, so only the slot test answers correctly for an error handed across the boundary.

Two details worth knowing:

- **Two constructors, not one.** Repair mints `%InitialError%` for the host realm and a powerless `%SharedError%` for compartments. Patching only the global left compartments broken; the shared one isn't a global, it's `Error.prototype.constructor`.
- **`globalThis.hardenIntrinsics` is deleted afterwards.** The phase call reveals it and `lockdown()` never does, so dropping it keeps the host realm's global surface differing from the one-call form only by the restored method. It's spent either way, but leaving it would let host code re-enter the harden pass.

## Verification

- New `error-taming.test.ts`: host realm (presence, semantics incl. `Error.prototype` -> `false`, hardened descriptor), compartments (own constructor, own errors, **host error across the boundary**), plus a cross-package guard that `tagFromNativeValue` classifies a constructor-less error post-lockdown — the caller this shim exists for.
- **Guard discipline:** commenting out the single `restoreErrorIsError()` call fails all 8 assertions; restoring it passes. The test cannot go vacuous.
- Compartment global surface confirmed unchanged (`repairIntrinsics`/`hardenIntrinsics` invisible inside; `lockdown`/`harden` as before).
- Runner suite **1000 passed / 0 failed**. `deno fmt --check`, `deno lint`, `deno check` clean. `deno task check-docs`: 519 blocks pass.

## Docs

`docs/specs/sandboxing/SES_SANDBOXING_SPEC.md` §5.1 updated — it showed the one-call `lockdown()` and named only the locale shim.

## Follow-on, not in this PR

data-model's `tagFromNativeValue` checks `Error.isError` *before* `instanceof FabricInstance`. This PR makes that ordering harmless, but reordering so known types classify first is arguably correct on its own merits and independent of SES. Queued, not built.

https://claude.ai/code/session_01PJLpGQpmCdPQCcF4M2inS8

---
## Summary by cubic
Restores `Error.isError` across `ses` lockdown to prevent TypeErrors and keep cross‑realm error checks working (e.g., data‑model’s `tagFromNativeValue`). Installs the original intrinsic between `repairIntrinsics()` and `hardenIntrinsics()` so both host and compartment constructors expose it.

- **Bug Fixes**
  - Add `restoreErrorIsError()` shim using the pre‑lockdown function; patches `%InitialError%` and `%SharedError%`.
  - Switch to two‑phase lockdown: `sanitizeLocaleMethods()` → `repairIntrinsics(options)` → `restoreErrorIsError()` → `hardenIntrinsics()`, then delete `globalThis.hardenIntrinsics`.
  - Drop unreachable shim branches and assert `hardenIntrinsics` (revealed by repair); remove `%InitialError%`/`%SharedError%` dedupe.
  - Tests cover host, compartment, cross‑realm behavior, hardened descriptor, and the no‑install path when given a non‑function; docs updated with the sequence and rationale.

<sup>Written for commit 9addf1890b0cdd726acc1603c61cbf5390825e21. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4906?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/sandboxing/SES_SANDBOXING_SPEC.md
+++ b/docs/specs/sandboxing/SES_SANDBOXING_SPEC.md
@@ -722,17 +722,36 @@ Lockdown options are derived from `debug`:
 
 ```typescript
 // Shown for illustration only.
-lockdown({
+const options = {
   errorTaming: debug ? "unsafe" : "safe",
   stackFiltering: debug ? "verbose" : "concise",
   overrideTaming: "severe",
   consoleTaming: "unsafe",
   localeTaming: "unsafe", // paired with the pre-lockdown locale sanitizer
-});
+};
+
+sanitizeLocaleMethods(); // vetted shim; before the intrinsics freeze
+repairIntrinsics(options);
+restoreErrorIsError(); // vetted shim; only this seam has both constructors
+hardenIntrinsics();
 ```
 
 Notes:
 
+- `lockdown(options)` is exactly `repairIntrinsics(options)` followed by
+  `hardenIntrinsics()`. The runtime spells out the two phases because the
+  `Error.isError` shim has nowhere else to run: SES's error taming rebuilds the
+  `Error` constructor and copies forward only `length`, `prototype` and the V8
+  stack surface, so the method exists neither before the repair (the
+  constructors are not minted yet) nor after the harden (they are frozen). Every
+  `ses` release through 2.2.0 drops it, and `permits.js` listing `isError` does
+  not help — a permit lets a property survive, it never adds one. Two
+  constructors need the restoration, not one: `%InitialError%` for the host
+  realm and `%SharedError%` for compartments, the latter reachable only as
+  `Error.prototype.constructor`. See
+  `packages/runner/src/sandbox/error-taming.ts`. The phase call reveals
+  `globalThis.hardenIntrinsics`, which the runtime deletes afterwards so the
+  host realm's global surface matches the one-call form.
 - `overrideTaming: "severe"` remains the compatibility default.
 - `localeTaming: "unsafe"` is load-bearing only together with the pre-lockdown
   vetted shim in `packages/runner/src/sandbox/locale-taming.ts`, which replaces
--- a/packages/runner/src/sandbox/error-taming.ts
+++ b/packages/runner/src/sandbox/error-taming.ts
@@ -44,26 +44,32 @@ const FERAL_IS_ERROR: unknown = (Error as { isError?: unknown }).isError;
 
 /**
  * Reinstall `Error.isError` on both post-repair error constructors. Must be
- * called after `repairIntrinsics()` and before `hardenIntrinsics()`; a no-op
- * on a runtime whose `Error` never had the method.
+ * called after `repairIntrinsics()` and before `hardenIntrinsics()`.
+ *
+ * Installs nothing unless handed a function, which is how a runtime whose
+ * `Error` never had the method declines: `FERAL_IS_ERROR` is `undefined` there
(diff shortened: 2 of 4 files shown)


seefeldb (approved):

mathpirate:
One additional acceptance case for the queued `tagFromNativeValue` follow-up: this shim deliberately installs nothing when the original runtime lacks `Error.isError`, but `tagFromNativeValue()` currently calls `Error.isError(value)` unconditionally whenever constructor-based classification returns `null`.

Thus an older browser without the intrinsic still throws `TypeError` in that path—the SES shim correctly declines to invent a cross-realm predicate, but the caller does not degrade gracefully when none exists.

I think the follow-up should combine the ordering change already noted here with feature detection:

```ts
if (value instanceof FabricInstance) {
  return NATIVE_TAGS.FabricInstance;
}

if (
  typeof Error.isError === "function" &&
  Error.isError(value)
) {
  return NATIVE_TAGS.Error;
}
```

The exact ordering of the remaining fallbacks is a separate choice, but the useful contract would be:

- known Fabric values never depend on optional host functionality;
- supported engines retain the genuine cross-realm error test;
- engines without `Error.isError` may fail to recognize some cross-realm errors, but classification does not crash.

This is not a problem with the merged restoration on supported runtimes; it is a small resilience gap exposed by the intentionally tested “no intrinsic to restore” branch.

danfuzz:
TBQH I think we can — or at least _should_ — be able to count on `Error.isError()` to be available everywhere we run. It's been widely available (with a minor asterisk for a Safari case that I don't think matters in this context) for over a year.

mathpirate:
Seems very reasonable to me.

PR #4916, opened 2026-07-22
fix(runner): version generated cell identities by pattern

mathpirate:
## Why

Compiler-generated causes such as `{ $generated: 0 }` are only meaningful within one pattern artifact, but their durable cell IDs were derived from only the result cell and generated ordinal. Two pattern versions could therefore mint the same internal-cell IDs for semantically different graph positions.

When old and new graphs overlap during an upgrade, they can repeatedly write different values to those shared entities. This creates unnecessary revisions, storage work, synchronization traffic, and client updates.

## What

- Associate compiler-generated internal-cell descriptors with their pattern's content-addressed `{ identity, symbol }`.
- Include that pattern identity in the generated cell's hash preimage.
- Record and match the pattern identity in the internal-cell manifest so an upgraded generated cell is initialized under its new ID.
- Preserve the existing identity behavior for explicitly named/manual cells so intentional durable state survives pattern upgrades.
- Carry generated-cell identity metadata across builder-derived pattern copies and cold rehydration.
- Update the live storage and cell-identity specifications.

## Impact

A pattern upgrade now rotates compiler-generated internal-cell IDs once instead of allowing old and new graphs to contend on the same entities. Explicitly named cells remain stable.

This PR addresses the identity collision itself. Immediate cancellation of an old graph during an asynchronous pattern load and server-side producer fencing remain separate follow-up defenses.

## Testing

- `deno task check`
- `deno task check-docs` — 519 checked code blocks
- Focused runner, generated-cell, computed-cell, map identity, cross-space reload, owned-cell resume, and rehydration safety suites — 150 steps passed

---
<a href="https://app.blacksmith.sh/commontoolsinc/codesmith/labs/pr/4916"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787344428&installation_model_id=428580&pr_number=4916&repository=commontoolsinc%2Flabs&return_to=https%3A%2F%2Fgithub.com%2Fcommontoolsinc%2Flabs%2Fpull%2F4916&signature=42a6e0c9dfb6d294e78de501aa6f50415810a34748ec9b2ba2533376b092912b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Namespaces compiler-generated internal-cell IDs by the owning pattern’s `{ identity, symbol }` to stop cross-version collisions and churn. Named/manual cell IDs stay stable; legacy generated IDs remain until a real pattern update, and updates now stage setup atomically with the pointer.

- **Bug Fixes**
  - Generated identity: add `patternIdentity` to hashing and manifest matching for generated causes; keep legacy generated entries without it while the piece still points at the same artifact; propagate through derived copies and rehydrate.
  - Atomic updates: run replacement setup in the same compare-and-swap as the pointer; commit manifest, defaults, backlinks, schemas, and projection together; rejected setup leaves the pointer and running graph unchanged.
  - Resume hardening: pre-sync result cells; resolve child outputs in a read-only txn; deduplicate shared child results; ignore children with no outputs; lazily associate generated descriptors for nested pattern copies.
  - Legacy cold-start repair: retain one-shot repair for pointer-only docs from older runtimes; failed repair surfaces the original start error without further mutations.

<sup>Written for commit 9e29077b0e7c1b427c757d3658486b5f9a95e4bb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4916?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/computed-cell-identity.md
+++ b/docs/specs/computed-cell-identity.md
@@ -119,6 +119,31 @@ Generated entries record and match their `patternIdentity` in addition to
 partial cause and kind; named entries continue to match by partial cause and kind
 (`packages/runner/src/runner.ts`, `materializeDerivedInternalCells`).
 
+There is one rollout compatibility rule. A piece materialized before generated
+causes were versioned has generated manifest entries with no
+`patternIdentity`. While that piece still points at the same pattern artifact,
+the runner continues to use its legacy ids instead of resetting it merely
+because the runtime changed. The next real pattern update atomically
+materializes pattern-versioned cells and moves the result projection to them.
+Old documents are not deleted; once no current manifest or projection names
+them, they are orphaned historical data.
+
+The transformer emits stable `.for(...)` causes for many authored declarations
+and result paths, but not for every reactive value. Dynamic computed properties,
+some callback-created values, and direct child-pattern calls can still fall
+through to `$generated`. Most observed examples are replayable UI
+intermediates or streams, but this is not a semantic guarantee: an untagged
+non-replayable builtin result can also be generated, and named state below a
+generated child-pattern result inherits that generated ancestor's churn.
+Durable state survives an update only when its complete identity ancestry is
+stable.
+
+Pattern-versioned identity is a runtime format change. A client that predates
+it cannot derive the new ids after an update. Automatic updates therefore need
+a coordinated runtime rollout: unchanged legacy pieces remain compatible, but
+an old client must not keep executing a piece after a newer client has moved
+that piece to a pattern-versioned generated graph.
+
 ### Transaction provenance
 
 The runner already distinguishes recompute transactions from event-handler
--- a/docs/specs/computed-cell-identity.md
+++ b/docs/specs/computed-cell-identity.md
@@ -85,15 +85,23 @@ changes, and only for entities whose ids carry the computed scheme.
 
 ### Internal cell identity
 
-The pattern builder assigns each internal root cell a `partialCause` —
-the cell's declared name, or an anonymous `{ $generated: N }` counter, with
+The pattern builder assigns each internal root cell a `partialCause` — the
+cell's declared name, or an anonymous `{ $generated: N }` counter, with
 `$kind: "stream"` mixed in for stream cells
-(`packages/runner/src/builder/pattern.ts`). At instantiation the runner
-mints the entity id from the piece's result cell and that partial cause:
+(`packages/runner/src/builder/pattern.ts`). Generated counters are local to one
+compiled pattern, so at instantiation the runner namespaces generated causes by
+the pattern's content-addressed `{ identity, symbol }`. Explicitly named causes
+remain unversioned so deliberate durable state survives pattern upgrades:
 
 ```ts
 // Shown for illustration only.
-createRef({}, { parent, type: "internal", cause: descriptor.partialCause })
+createRef({}, {
+  parent,
+  type: "internal",
+  cause: generated
+    ? { patternIdentity, partialCause: descriptor.partialCause }
+    : descriptor.partialCause,
+})
 ```
 
 (`packages/runner/src/link-utils.ts`, `getDerivedInternalCellLink`). The
@@ -106,8 +114,9 @@ a single chokepoint (`packages/data-model/src/value-hash.ts`). The URI layer
 (`packages/runner/src/uri-utils.ts`, `toURI`/`fromURI`) prefixes an entity
 scheme onto the tagged hash — historically always `of:`.
 
-The manifest of materialized internal cells is stored in result-cell
-metadata and matched by partial cause plus kind
+The manifest of materialized internal cells is stored in result-cell metadata.
+Generated entries record and match their `patternIdentity` in addition to
+partial cause and kind; named entries continue to match by partial cause and kind
 (`packages/runner/src/runner.ts`, `materializeDerivedInternalCells`).
 
 ### Transaction provenance
@@ -422,13 +431,13 @@ contents are meaningless under the new kind, and the manifest's
 partial-cause matching materializes the new cell and drops the stale entry
 naturally.
 
-Internal-cell identity is already refactor-fragile — anonymous cells re-mint
-on reorder via the `$generated` counter, named cells on rename — so kind
+Internal-cell identity is already refactor-fragile — generated cells re-mint on
+every pattern identity change, while named cells re-mint on rename — so kind
 flips add a trigger to an existing hazard class (durable cross-piece links
-pointing at an orphaned entity), not a new class. The flipped classifier
-polarity widens the set of cells that flip when a pattern edit adds or
-removes a disqualifier (e.g. introducing a writable handler capture of a
-previously computed cell), which is the same hazard at higher frequency.
+pointing at an orphaned entity), not a new class. The flipped classifier polarity
+widens the set of cells that flip when a pattern edit adds or removes a
+disqualifier (e.g. introducing a writable handler capture of a previously
+computed cell), which is the same hazard at higher frequency.
 
 ### Trust model
 
--- a/docs/specs/computed-cell-identity.md
+++ b/docs/specs/computed-cell-identity.md
@@ -85,15 +85,23 @@ changes, and only for entities whose ids carry the computed scheme.
 
 ### Internal cell identity
 
-The pattern builder assigns each internal root cell a `partialCause` —
-the cell's declared name, or an anonymous `{ $generated: N }` counter, with
+The pattern builder assigns each internal root cell a `partialCause` — the
+cell's declared name, or an anonymous `{ $generated: N }` counter, with
 `$kind: "stream"` mixed in for stream cells
-(`packages/runner/src/builder/pattern.ts`). At instantiation the runner
-mints the entity id from the piece's result cell and that partial cause:
+(`packages/runner/src/builder/pattern.ts`). Generated counters are local to one
+compiled pattern, so at instantiation the runner namespaces generated causes by
+the pattern's content-addressed `{ identity, symbol }`. Explicitly named causes
+remain unversioned so deliberate durable state survives pattern upgrades:
 
 ```ts
 // Shown for illustration only.
-createRef({}, { parent, type: "internal", cause: descriptor.partialCause })
+createRef({}, {
+  parent,
+  type: "internal",
+  cause: generated
+    ? { patternIdentity, partialCause: descriptor.partialCause }
+    : descriptor.partialCause,
+})
 ```
 
 (`packages/runner/src/link-utils.ts`, `getDerivedInternalCellLink`). The
@@ -106,8 +114,9 @@ a single chokepoint (`packages/data-model/src/value-hash.ts`). The URI layer
 (`packages/runner/src/uri-utils.ts`, `toURI`/`fromURI`) prefixes an entity
 scheme onto the tagged hash — historically always `of:`.
 
-The manifest of materialized internal cells is stored in result-cell
-metadata and matched by partial cause plus kind
+The manifest of materialized internal cells is stored in result-cell metadata.
+Generated entries record and match their `patternIdentity` in addition to
+partial cause and kind; named entries continue to match by partial cause and kind
 (`packages/runner/src/runner.ts`, `materializeDerivedInternalCells`).
 
 ### Transaction provenance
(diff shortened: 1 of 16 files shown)


seefeldb (approved):
looks fine, but in practice i think it means that any data for which the transformer didn't generate a .for will be lost on update. that's probably all just interim computeds, so not an issue, but worth noting.

willkelly:
## Matched Topics diagnostics result

_Agent-authored investigation using the diagnostics introduced in #4950._

I ran the deterministic `--profile=conflicts` workload against clean worktrees at the exact commits for this PR, with the same diagnostics changes overlaid on both sides:

- base: `93f97e5682611f040f7b4b59fcf8a0654d4cb084`
- head: `32287d3424a21face84d6ff900d4668034bff1b3`
- 3 runs per side, alternating base/head to reduce temporal bias
- fixed topology: 2 topics, 2 users, 2 sessions per user, 10 ms delayed frames, 4 root-oscillation rounds

### Deterministic results

All six runs were identical on the behavior-level invariants:

- 16 submitted / 8 accepted / 8 rejected
- two-step repeat ratio: 1
- final convergence: true
- no dropped or terminally failed workload operations

This is useful negative evidence: the PR does not introduce a deterministic regression in this steady-state contention profile.

### Observational results

| Signal | Base runs | Head runs | Median change |
| --- | --- | --- | --- |
| Workload phase elapsed | 1503–1655 ms (median 1578) | 1429–1633 ms (median 1570) | head -0.55% |
| Total elapsed | 5768–6288 ms (median 5989) | 6047–6233 ms (median 6121) | head +2.20% |
| Workload-phase memory conflicts | 30–33 (median 31) | 24–36 (median 25) | head -19.35% |

The conflict median moves in the favorable direction, but the ranges overlap and the head also produced the worst conflict outlier (36). Phase timing is effectively flat, while total timing moves slightly the other way. With three runs per side and this amount of overlap, I would treat these as noisy observations rather than a demonstrated performance change.

The focused tests also pass on the PR head:

- `deno test --frozen -A packages/runner/test/runner.test.ts packages/runner/test/derived-copy-identity.test.ts`

### Conclusion and scope

This workload exercises steady-state multi-runtime contention after startup. It does **not** perform a pattern upgrade, so it cannot directly observe the old/new generated-cell identity overlap that this PR is intended to reduce. The evidence therefore neither proves nor disproves the upgrade-specific performance claim. It shows no deterministic steady-state regression and gives a weak, inconclusive favorable conflict signal.

A conclusive follow-up needs an upgrade-specific workload: start active sessions on version N, upgrade to N+1 while writes continue, then measure old/new graph overlap, dirty/pending scheduler state, transaction conflicts, and convergence over the upgrade window. #4950 provides the aggregate telemetry and event-driven barriers needed for that next experiment.

willkelly:
## Upgrade-specific matched base/head identity result

Follow-up to the matched Topics diagnostics above: this run exercises the active-session pattern-upgrade path directly.

### Method

- Measured base `93f97e5682611f040f7b4b59fcf8a0654d4cb084` and head `32287d3424a21face84d6ff900d4668034bff1b3`.
- Used the same staged diagnostics overlay and a byte-identical six-file experiment patch in both worktrees.
- Kept two worker-isolated sessions active while session 0 applied the v1 → v2 pattern identity transition.
- The fixture isolates one top-level transformer-generated `$generated` cell from explicitly named/manual state, including a durable probe seeded with a private sentinel.

### Result

| Observation | Base | Head |
|---|---:|---:|
| observed generated-cell identity overlap | 2/2 active sessions | 0/2 active sessions |
| named/manual identity overlap | 6/6 | 6/6 |
| durable sentinel preserved | yes | yes |
| same Runtime / PiecesController / PieceController | yes | yes |
| identity transitions observed | 2/2 | 2/2 |
| post-settle dirty / pending | 0 / 0 | 0 / 0 |

Both sides converged and showed the same expected single conflict/revert during the shared upgrade. Reports were aggregate-only and contained no IDs, DIDs, URLs, or sentinel value.

Under this controlled workload, the base→head delta changes the tested top-level generated child from retaining identity across upgrade to receiving a new identity, while explicitly named durable identities and state remain stable. This supports the PRs stated upgrade semantics; it is not a general performance claim and does not by itself prove every pattern shape, old-graph cancellation, or server-side producer fencing.

The PR has since advanced beyond the measured refs (currently base `0f9150b`, head `d2947c1`), so the observation is pinned to the exact SHAs above.

_Experiment run by Hephaestus (OpenAI GPT-5.6 Sol), on behalf of @wkelly._

mathpirate:
## Rigor-pass findings and follow-ups

I did an adversarial review of the generated-cell identity change, including tracing the transformer’s `.for(...)` emission, the builder/runner identity path, live pattern updates, pre-PR storage compatibility, and cold nested resume behavior. The current branch includes the hardening described below.

### Interpreting the `.for` concern

Berni’s observation is correct: this PR matters specifically for internal cells that the transformer/builder has not given a stable authored cause via `.for(...)`.

Conceptually, the old identities were:

```text
named:      hash(piece, "stable-cause")
generated:  hash(piece, { $generated: 0 })
```

A generated ordinal is meaningful only within one compiled pattern. Pattern N’s `$generated: 0` and pattern N+1’s `$generated: 0` can represent unrelated graph positions, yet the old scheme minted the same entity. During an update, old and new graphs could therefore contend on that entity or silently treat old data as belonging to new semantics.

This PR changes only generated identities to include the content-addressed pattern `{ identity, symbol }`:

```text
generated: hash(piece, patternIdentity, { $generated: 0 })
```

Thus a genuine pattern update rotates generated cells instead of aliasing them across versions. “Data won’t transfer” means that the new manifest/projection points at a newly initialized entity; the old document is not deleted, but becomes orphaned historical data once nothing current references it. Named cells retain their existing cross-version behavior.

The transformer already emits `.for(...)` for many declarations and stable result paths, so most generated cells observed in shipped patterns are replayable UI/stream/computed scaffolding. That is directionally reassuring, but it is not a semantic guarantee. Current gaps include some callback-created values, dynamic computed properties, and child-pattern calls. A non-replayable or expensive builtin can be generated (for example, a `generateText(...)` created inside a callback), and named writable state beneath a generated child-pattern result still inherits the generated ancestor’s churn. Durable state therefore requires the *complete identity ancestry* to be stable, not merely a named leaf.

Even if future transformer work eliminates most accidental generated cells, versioning remains the safer fallback: resetting an anonymously identified value is preferable to silently transferring unrelated data between two semantic graph positions.

### Hardening added during this review

1. **Atomic live updates.** The existing `PatternUpdater` changed only `patternIdentity`. With versioned generated IDs, the watcher could instantiate nodes that wrote the new cells while the durable manifest, defaults, backlinks, schemas, and visible result projection still described the old graph. Replacement setup now runs in the same compare-and-swap transaction as the pointer change. Setup-validation or commit failure leaves the old pointer and running graph intact. A compiler-produced `$generated` auto-update regression pins the visible v1 → v2 transition and the changed manifest/link identity.

2. **Pre-PR rollout compatibility.** An unchanged piece materialized by an older runtime has generated manifest entries without `patternIdentity`. Merely upgrading the runtime must not reset it. While the piece still points at that same artifact, the runner recognizes the legacy manifest entry and continues deriving its legacy ID. The first real pattern update materializes the versioned graph atomically. The compatibility lookup is indexed per immutable manifest revision to avoid repeated linear scans on large manifests.

3. **Nested cold-resume correctness.** Generated-descriptor ownership is propagated lazily across copied nested patterns. The resume walk could derive a child’s owned cells before resolving that association or loading the child result metadata, causing it to pre-sync the wrong legacy/versioned IDs. The walk now resolves the artifact association, synchronizes nested result metadata first, then derives child-owned cells; sibling child syncs remain parallel. A scoped nested-callback fixture now exercises a real compiler-generated cause.

4. **Failure and legacy-repair contracts.** Piece-level tests now assert that an incompatible candidate aborts before changing the identity and leaves the old nodes functional. The existing cold-start repair is explicitly retained for split-brain documents already produced by older pointer-only runtimes, including fail-closed retry behavior.

5. **Specs and rollout notes.** The identity, storage, memory, cell, and pattern-update specs now document legacy generated entries, orphan semantics, incomplete `.for` coverage, full-ancestry durability, atomic replacement setup, and the mixed-runtime boundary.

Validation performed:

- full runner: 1,014 tests / 5,420 steps
- final-head focused runner: 12 tests / 121 steps
- piece package: 22 tests / 257 steps
- repository-wide type check
- docs: 518 checked code blocks

### Recommended follow-ups

1. **Expand and measure transformer `.for` coverage.** Audit every remaining `$generated` production site, especially plain/reactive callbacks, dynamic properties, and child-pattern roots. Add fixtures that prove stable causes for cases intended to retain state. Keeping a generated-cause inventory in CI would make regressions visible.

2. **Diagnose risky generated cells.** Warn or fail authoring checks when `$generated` reaches a writable/stateful cell, a non-replayable or externally expensive builtin, or an ancestor of named durable state. Ephemeral UI/computed scaffolding can remain permitted.

3. **Make persistence intent explicit.** Document/encode the rule that durable identity requires a stable full ancestor chain. Where an update intentionally needs to migrate data between changed identities, use an explicit migration/copy mechanism rather than depending on generated ordinal coincidence.

4. **Coordinate the runtime-format rollout.** An older runtime cannot derive the new generated IDs after a newer runtime updates a piece. Unchanged legacy pieces remain compatible, but old clients must not continue executing a piece after it has moved to a versioned generated graph. A runtime/storage capability gate would be stronger than relying solely on deployment coordination.

5. **Complete the overlap defenses already called out by the PR.** Immediate cancellation/fencing of superseded producers and server-side producer fencing remain valuable independently of identity separation.

6. **Eventually reclaim orphaned generated entities.** Rotation intentionally leaves historical documents behind. Once ownership/liveness rules are strong enough, add observability and conservative GC for unreachable generated cells.

ubik2:
Not sure whether it makes sense, but I wonder if we can generate this PatternIdentityRef information in our `factoryFromPattern` call. If so, we could just build it into `nextAnonymousPartialCause`.

willkelly:
### Topics active-upgrade-under-write performance check (corrected)

> This fully supersedes the earlier contents of this comment. The stale-ref/raw-link experiment, steady-state conflict profile, and all intermediate benchmark datasets were excluded after review found they did not measure the PR-specific upgrade window correctly.

**Compared refs**

- Base: `302653a4e6d16e6b901c5c0b60a836ed6c8989e8`
- Head: `d29137b69d103851869e3f9d29f791ab4590056c`
- Pinned runtime: Deno `2.8.1`, `--frozen`
- The six temporary benchmark files were SHA-256-identical on both refs; no dependency or lockfile changes.

**Workload and timing boundary**

- One identity with two worker-isolated active sessions (two tabs).
- Real same-path v1 → v2 `setPattern()` update using the anonymous dynamic-callback shape from the PR regression test: `["slot"].map(() => computed(...))`.
- V2 resolves before measurement. Each session completes a changing warm write before T0.
- At T0, parent-observed one-operation-at-a-time writer chains and the prepared update begin. T1 is `max(setPattern completion, all session identity-transition acknowledgements)`.
- Only writer completions timestamped by the parent inside `[T0,T1]` count toward window throughput. Each session also completes an explicit post-transition write.
- Private convergence checks the actual generated, sentinel, and writer components. The channel is derived inside workers from the aggregate-hidden piece identity; only booleans/cardinalities leave workers.
- Seven balanced interleaved pairs were retained: `B,H / H,B / B,H / H,B / B,H / H,B / B,H`.

**Correctness result — all 14 runs passed**

- Generated causes: `2/2` on both arms.
- Generated links: base `2/2` overlap; head `0/2` overlap, matching the PR contract.
- Explicitly named links: `4/4` overlap on both arms; sentinel value preserved.
- Actual-state convergence and cardinalities matched; dirty/pending nodes, writer drops, writer rejections, and terminal failures were all zero.
- Canonical memory/churn telemetry recorded one base-side storage conflict/rejection and zero head-side conflicts/rejections; every run converged.

| Metric | Base | Head | Head vs base |
|---|---:|---:|---:|
| Upgrade transition, median | `334.410 ms` | `320.770 ms` | `-4.079%` |
| Upgrade transition, range | `301.952–358.852 ms` | `295.286–351.545 ms` | — |
| `setPattern()` call, median | `314.845 ms` | `300.080 ms` | `-4.690%` |
| In-window operations/ms, median | `0.2381` | `0.2463` | `+3.453%` |

Paired head/base primary ratios were:

`1.1244, 0.9225, 0.8897, 1.0322, 1.0248, 0.9725, 0.8830`

**Verdict: noisy/inconclusive; no performance regression was demonstrated.**

The primary arm ranges were below the preregistered 20% stability bound, but pair directions were mixed and crossed the 10% practical threshold in both directions. Therefore this is not evidence of parity or improvement. Throughput is descriptive only because its run-to-run ranges were high (`90.9%` base, `63.0%` head).

**Scope:** this covers one user with two active sessions and the dynamic-callback upgrade shape targeted by #4916. It does not generalize to multi-user or higher-fanout contention, benchmark unchanged legacy-manifest rehydration, or prove old-producer cancellation/fencing.

mathpirate:
Closing this in favor of pursuing the line of work begun in https://github.com/commonfabric/labs/pull/7846

PR #4924, opened 2026-07-22
test(schema-generator): treat Fabric primitives as leaves in golden normalization

danfuzz:
Harness hardening for #4904, off `main`. No behavior change outside tests; the
merged schema fix is not affected.

## Why

#4904 gave the golden format a lossless codec, aiming at "every `FabricValue`."
The normalization around it is not lossless yet. `deepCanonicalize()`
(`test/utils.ts`) and `normalizeArrayOrdering()` (`test/fixtures-runner.test.ts`)
both walk every non-null object with `Object.entries`. A `FabricPrimitive` —
`FabricBytes`, and its siblings — keeps its state in private fields with **no
enumerable own properties**, so both walks flatten one to `{}`. The compare then
feeds that to `valueEqual`, which sees `{}` on each side and calls two different
primitives equal:

1. an actual special-object value is reduced to `{}` by `normalizeSchema()`;
2. the golden decodes to the correct primitive;
3. `normalizeArrayOrdering()` reduces the decoded value to `{}`;
4. `valueEqual({}, {})` passes.

That is the silent-agreement failure this harness exists to prevent, recreated
one subclass short of its goal. The existing round-trip test does not catch it
because it calls `encodeGolden` / `decodeGolden` directly and bypasses both
normalizers. (Reported by @mathpirate on #4904.)

The special *numbers* that motivated the format, and the coming bigint, decode
to JS primitives — not special objects — so nothing merged is affected. This is
the future-proofing tail.

## What

Both walks stop at a `FabricPrimitive` and treat it as a leaf, so `valueEqual`
compares it by content instead of comparing two `{}`s. The discriminator is
`instanceof FabricPrimitive` — the sanctioned check (`interface.ts` blesses
`instanceof FabricSpecialObject` for "any special object"; here the narrower
subclass is what's wanted).

A regression drives a `FabricBytes` default through the **full**
encode → decode → normalize → compare path — not the codec round trip the
existing test covers — asserting that matching bytes still compare equal and
differing bytes still compare unequal. It was run against the un-hardened walks
and observed to fail (both sides flattened to `{}`, wrongly equal).

## Scope: `FabricInstance` deliberately left

`FabricInstance` — the other `FabricSpecialObject` subclass — is **not** treated
as a leaf. It wraps nested `FabricValue`s that a faithful walk would descend
into, and `Object.entries` cannot do that either, so it still flattens here. A
`TODO(danfuzz)` in each walk records the gap; the golden path carries no
`FabricInstance` today, and handling it properly is its own task.

## Testing

- schema-generator suite green (`37 passed`, +1 for the regression); a full
  `UPDATE_GOLDENS=1` run leaves the tree clean — **zero fixture churn**.
- `./tasks/check.sh` clean; `deno fmt` clean.

---
<a href="https://app.blacksmith.sh/commontoolsinc/codesmith/labs/pr/4924"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787354325&installation_model_id=428580&pr_number=4924&repository=commontoolsinc%2Flabs&return_to=https%3A%2F%2Fgithub.com%2Fcommontoolsinc%2Flabs%2Fpull%2F4924&signature=663c6ed82a1f57c6345451eef916c81d246987a6551a4c25893fecc32144f528"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Hardened the schema-generator golden compare by treating `FabricPrimitive` values as leaves during normalization to prevent them from flattening to `{}` and comparing equal by mistake. Test-only change; no runtime behavior impact.

- **Bug Fixes**
  - Updated `deepCanonicalize()` and `normalizeArrayOrdering()` to stop at `FabricPrimitive` (`FabricBytes`, etc.), so `valueEqual` compares by content instead of `{}`.
  - Kept `FabricInstance` as non-leaf with a TODO; it wraps nested `FabricValue`s and isn’t carried on the golden path today.

- **Testing**
  - Added a regression that runs a `FabricBytes` default through encode -> decode -> normalize -> compare; identical bytes are equal, different bytes are not.
  - No fixture churn (`UPDATE_GOLDENS=1` clean).

<sup>Written for commit 0c99783446006d444a1b86ee16482146fd959fe1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4924?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/schema-generator/test/fixtures-runner.test.ts
+++ b/packages/schema-generator/test/fixtures-runner.test.ts
@@ -14,9 +14,11 @@ import {
   valueFromJson,
 } from "@commonfabric/data-model/codec-json";
 import {
+  FabricPrimitive,
   type FabricValue,
   valueEqual,
 } from "@commonfabric/data-model/fabric-value";
+import { FabricBytes } from "@commonfabric/data-model/fabric-primitives";
 import { createSchemaTransformerV2 } from "../src/plugin.ts";
 import {
   batchTypeCheckFixtures,
@@ -218,7 +220,18 @@ function normalizeArrayOrdering(obj: unknown): unknown {
   if (Array.isArray(obj)) {
     return obj.map(normalizeArrayOrdering);
   }
-  if (obj && typeof obj === "object") {
+  // A `FabricPrimitive` (`FabricBytes` and the like) has no enumerable own
+  // properties, so `Object.entries` would flatten it to `{}` and hide whatever
+  // it carried -- and then `valueEqual`, which this feeds, would compare two
+  // different primitives as equal. Treat it as a leaf.
+  //
+  // TODO(danfuzz): A `FabricInstance` (the other `FabricSpecialObject`
+  // subclass) is NOT a leaf -- it wraps nested `FabricValue`s -- but there is
+  // no faithful `Object.entries` descent for it either, so it still flattens to
+  // `{}` here. Handle it when the golden path needs to carry one.
+  if (
+    obj !== null && typeof obj === "object" && !(obj instanceof FabricPrimitive)
+  ) {
     const entries = Object.entries(obj as Record<string, unknown>)
       .map(([key, value]) => {
         if (key === "required" && Array.isArray(value)) {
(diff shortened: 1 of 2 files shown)


mathpirate:
Nice follow-up—the `FabricPrimitive` fix and the `FabricBytes` regression both look right. I think the deliberate `FabricInstance` carve-out leaves the same silent-agreement failure in place, though.

A `FabricInstance` is not a leaf in the Fabric data model, but I believe it should be treated as an opaque leaf by these schema-specific canonicalizers. They cannot faithfully traverse its private state, while the codec and `valueEqual()` already know how to encode and compare that state recursively. Continuing through `Object.entries()` is actively lossy: for example, two `FabricLink`s with different payloads both have no enumerable own properties, so both become `{}` and incorrectly compare equal.

Could we stop at `FabricSpecialObject` rather than only `FabricPrimitive` in both normalizers, and add a `FabricLink` case alongside the `FabricBytes` regression? If preserving a mutable `FabricInstance` by reference is undesirable, I’d still prefer explicitly rejecting one here until instance-aware normalization exists; failing clearly preserves the harness invariant, whereas flattening it can silently approve a bad golden.

This seems worth closing in this PR because the golden codec’s intended contract is “every `FabricValue`,” and broadening the discriminator should be a small change.

danfuzz:
@mathpirate Thanks for the review! I think the current state is faithful with regards to where this project is in the work arc of supporting `FabricInstance`: The `TODO` captures the pending work — the gap isn't lost — and when we address it your stated concern will naturally end up being addressed.

PR #4926, opened 2026-07-22
fix(piece): cold-start setup repair for roots whose identity moved without setup

bfollington:
Was stacked on #4900; now rebased onto main after #4900 and #4901 merged — only the two commits here are new. Companion to #4901.

## Why

#4900 correctly fixes the *running-piece* swap path: the `patternIdentity` watcher now runs setup before instantiating. But a bricked space's root was never running — its old source fails to compile — so no watcher exists. The swap is applied pre-start by `checkAndUpdateDefaultPattern` as a bare metadata write, and the subsequent cold start (`startPiece` → `startCore`) instantiates with no `applySetupState`. Result: the exact live-Estuary failure, reproduced byte-for-byte on #4900's branch by the first new test here:

```
Error: Handler used as lift: node's $event input resolves to of:fid1:..., which reads
undefined — the { "$stream": true } marker was never written there.
    at Runner.startCore (runner.ts)        ← initial instantiation, no setup
    at async PiecesController.startEnsuredDefaultPattern
    at async PiecesController.ensureDefaultPattern
```

Worse, any space whose swap **already committed durably** under the deployed July-22 build (identity now compares current, doc still marker-less — the likely current state of the bricked Estuary spaces) never re-enters the swap path at all, so even a fixed swap can't revisit it. Without this repair, those spaces stay bricked after deploying #4900+#4901.

## What changed

One mechanism in `startEnsuredDefaultPattern` covers both cases: if the cold `startPiece` fails, make a single setup-repair attempt — resolve the stored (unchanged) identity via `loadPatternByIdentity`, then `runtime.runSynced(...)` (setup + start). Setup is idempotent here: `materializeDerivedInternalCells` seeds only manifest-missing internal cells (the `$stream` markers included) and never overwrites user data. Fail-closed: if the repair can't proceed or fails, the original start error is rethrown and nothing is torn down. Piece-package only; zero cost on the happy path; the runner and #4900's watcher fix are untouched.

## Test plan

Two new tests in `packages/piece/test/check-update-default-pattern.test.ts`, red/green verified against #4900's head:

1. **Cold pre-start swap:** handler-bearing replacement swapped in while the piece is stopped, driven through the real `ensureDefaultPattern` boot entry; asserts the handler actually fires post-heal (`bump.send({})` → `count === 1`). Red on #4900 alone.
2. **Already-swapped doc:** `patternIdentity` already points at the current pattern, doc never set up — no further swap will fire; cold start itself must heal. Red on #4900 alone.

After the fix: target file 41/41 steps green (including all of #4900's own tests), full `packages/piece` suite 21 passed / 247 steps / 0 failed, `deno check` / `lint` / `fmt` clean.

---
## Summary by cubic
Fixes cold-start failures when a piece’s `patternIdentity` changed while stopped. Adds a one-time setup repair on start so bricked spaces boot cleanly and handlers work; enforces a precondition so commit failures surface and concurrent updates are rejected.

- Bug Fixes
  - Add one-time cold-start setup repair in `startEnsuredDefaultPattern`.
  - On start failure, resolve stored identity via `getPatternIdentityRef` + `patternManager.loadPatternByIdentity`, detach any read-only tx with `root.withTx()`, then call `runtime.runSynced(...)` with `expectedPatternIdentity` to materialize missing internal cells (e.g., `$stream` markers), throw on commit failures, and reject if identity changed — without touching user data.
  - Fail-closed: rethrow the original start error if repair can’t run or fails; happy path and watcher-based swap remain unchanged. Logs under `cold-start-setup-repair` and `cold-start-setup-repair-failed`.
  - Tests cover cold pre-start swap, already-swapped docs healed on boot, fail-closed repair (original error surfaces; next boot heals), refused-swap functional pin (`marker === "v1"`), and repair admission guards (malformed identity ref; pattern load undefined/reject) that rethrow the original start error and leave the doc healable.

<sup>Written for commit 80c5bde28be11cd73e579be6078ccec8e81414a5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4926?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/piece/src/ops/pieces-controller.ts
+++ b/packages/piece/src/ops/pieces-controller.ts
@@ -633,10 +633,20 @@ export class PiecesController<T = unknown> {
       // the lazy updater schedule one redundant check post-start. Benign: the
       // awaited checkAndUpdateDefaultPattern above already reconciled, so the
       // check observes a current identity and no-ops.
+      //
+      // expectedPatternIdentity is the repair precondition, not a formality:
+      // it atomically rejects a repair superseded by a concurrent source
+      // update (the identity is re-asserted inside every setup retry), and it
+      // makes runSynced THROW on a setup-commit failure instead of logging
+      // and continuing — without it this catch never sees commit-level
+      // failures and a dead root would be reported as a successful start.
       try {
         await timePiecesPhase(
           "ensureDefaultPattern.coldStartSetupRepair",
-          () => runtime.runSynced(writableRoot, repairPattern),
+          () =>
+            runtime.runSynced(writableRoot, repairPattern, undefined, {
+              expectedPatternIdentity: ref,
+            }),
         );
       } catch (repairError) {
         pieceUpdateLogger.warn(
--- a/packages/piece/src/ops/pieces-controller.ts
+++ b/packages/piece/src/ops/pieces-controller.ts
@@ -3,6 +3,7 @@ import {
   entityIdFrom,
   type EnvReader,
   experimentalOptionsFromEnv,
+  getPatternIdentityRef,
   type JSONSchema,
   type MemorySpace,
   type ModuleByteCache,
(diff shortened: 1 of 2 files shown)


mathpirate:
Taking over review routing from Berni's approvals upstream — and handing you the operational kit for the deploy, since you own the heal line from here.

**Two safety properties verified independently against the diff** (Fable, structural): (1) with the identity pointer unchanged, `runSynced`'s setup enters `applySetupState` with `samePattern=true`, whose argument branch (`else if (!samePattern)`) never fires — the repair writes only manifest-missing internal cells and cannot rewrite the stored argument; (2) the failed `startPiece` reaches `cleanup()` before throwing, so `this.cancels` is empty when the repair runs — `maybeReuseRunningSetup`'s cancels-gate can't short-circuit the repair into the reuse path. Both of your data-safety claims hold. The repair firing on *any* start failure (not just lift-class) reads as deliberate simplicity — idempotence plus rethrow-original makes that safe, at the cost of one extra load+run attempt on genuinely-failed starts; fine by me.

**Operational kit for the one-deploy validation** (credentials out-of-band):
1. Standing testbed: a throwaway estuary identity whose home root is a deliberately-bricked sourceless custom program — its swap engaged under the July-22 build, so it's plausibly your test-2 state live in prod. Reload it post-deploy and the console narrates every layer; it proves the mechanical chain (#4873/#4883/#4899 deployed + #4900 + this).
2. That testbed's home argument has **no profiles**, so it cannot see #4901's layer — my own home is the true fixture there (stale June profiles + the bio path). I'll do that reload myself and report.
3. Deploy math: #4873/#4883/#4899 are live; #4900 + #4901 + this = the remaining set, one deploy, then the two reloads above. If a further layer exists, the consoles now name it — every failure mode in this chain narrates.

bfollington (commented):

PR #4926 · detail mode, both changed files deep-read · targeted piece test 44/44 steps, runner test 86/86 steps, type-check/fmt/lint clean. CI currently has one failing Performance Check (`packages/piece` coverage debt +15 lines).

**Verdict:** the successful repair path is well targeted and the functional tests are strong, but the repair needs the existing atomic identity guard before merge.

packages/piece/src/ops/pieces-controller.ts
```diff
@@ -578,10 +579,77 @@ export class PiecesController<T = unknown> {
...
+      // would otherwise adopt it for the setup writes.
+      const writableRoot = rootToStart.withTx();
+      // runSynced does not plumb schedulePatternUpdate, so the repair may let
+      // the lazy updater schedule one redundant check post-start. Benign: the
+      // awaited checkAndUpdateDefaultPattern above already reconciled, so the
+      // check observes a current identity and no-ops.
+      try {
+        await timePiecesPhase(
+          "ensureDefaultPattern.coldStartSetupRepair",
+          () => runtime.runSynced(writableRoot, repairPattern),
```
bfollington: 🔴 **Blocking (critical, verified): preserve the repair precondition.** This reads `ref`, asynchronously loads that pattern, and then calls `runSynced` without `{ expectedPatternIdentity: ref }`. If another source update wins before setup commits, this stale repair can write the old identity back over the winner. Also, `Runner.runSynced` deliberately logs and swallows exhausted setup-transaction errors when no expected identity is supplied (`runner.ts:2372-2384`), so this method can continue and return a controller even though both the initial start and repair failed—contradicting the fail-closed comment above. Pass `undefined, { expectedPatternIdentity: ref }`; that both atomically rejects a superseded repair and makes setup-commit failures reject. Please add a regression test for a failed or superseded repair; the current Performance Check failure (`packages/piece` coverage debt +15) reflects this new error path being untested.

bfollington: Fixed in a1726507f — verified the swallow at runner.ts:2372-2384 exactly as described. The repair now passes `undefined, { expectedPatternIdentity: ref }`, and a new test pins the failure contract: a repair failing at the runSynced boundary surfaces the ORIGINAL start error and leaves the doc healable by the next boot (which then heals end-to-end). Driven at the runSynced seam because in-process failure classes (arg validation) are skipped for an unchanged identity via samePattern — commit-layer failure is the realistic remaining class.

packages/piece/src/ops/pieces-controller.ts
```diff
@@ -572,10 +572,73 @@ export class PiecesController<T = unknown> {
...
+      );
+      const repairPattern = pattern;
+      // Detach any transaction view the resolved root carries: getDefault-
+      // Pattern hands back a cell bound to a read-only tx, and runSynced
+      // would otherwise adopt it for the setup writes.
+      const writableRoot = rootToStart.withTx();
+      try {
+        await timePiecesPhase(
+          "ensureDefaultPattern.coldStartSetupRepair",
+          () => runtime.runSynced(writableRoot, repairPattern),
```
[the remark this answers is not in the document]
bfollington: Valid — fixed in a1726507f (expectedPatternIdentity passed; regression test added for the failed-repair path). See the sibling thread for detail.

packages/piece/src/ops/pieces-controller.ts
```diff
@@ -578,10 +579,87 @@ export class PiecesController<T = unknown> {
...
+      // identity compares current, so no further swap will ever fire.
+      //
+      // run() (setup + start) is the sanctioned repair: with an unchanged
+      // pattern pointer the setup phase is idempotent — it only materializes
+      // missing internal cells and writes no argument (none supplied). Fail
+      // closed: if the repair cannot proceed or fails, surface the ORIGINAL
+      // start error; nothing is torn down or overwritten.
+      const runtime = this.#manager.runtime;
+      const ref = getPatternIdentityRef(rootToStart);
+      if (ref === undefined) throw startError;
```
bfollington, line 607: 🔴 **Blocking (test coverage, verified): cover the repair early exits.** The current Performance Check is still red (`packages/piece` coverage debt 241 vs 236). I downloaded and combined the CI LCOV shards: lines 607 and 615–618 have zero hits, covering the missing-ref branch and the `loadPatternByIdentity` rejection/undefined branches. Please add focused tests showing each path rethrows the original start error and leaves the document unchanged. The new `runSynced` rejection test correctly covers the later fail-closed catch and next-boot healability, but it stubs the entire method and cannot exercise these earlier exits.

bfollington (commented):
_cf-review re-review via Codex (GPT-5), on behalf of @bfollington_

PR #4926 at `a1726507` · detail mode, updated implementation and test deep-read · piece test 45/45 steps, type-check/fmt/lint clean; all functional CI suites pass.

The prior correctness blocker is resolved: `expectedPatternIdentity: ref` now protects every setup retry and makes commit failures reject.

**Verdict:** implementation correctness looks good; merge after closing the remaining five-line coverage regression.

PR #4928, opened 2026-07-23
feat(lunch-poll): retain canonical profile links

mathpirate:
## Why

Lunch Poll currently treats `#profileName` and `#profileAvatar` as frozen strings. That makes a profile-backed participant indistinguishable from a guest and prevents Loom from rendering or following the user's canonical profile as it changes.

The poll also has an important compatibility constraint: today's deployed piece uses a bare, name-keyed `users` array, and guests must still be able to join with a typed string.

## What

- Resolve the current viewer's live `#profile` cell in the parent pattern.
- Add an optional, object-wrapped `participantProfiles` directory for canonical cross-space profile links, separate from the legacy bare `users` array.
- On a profile-backed join, retain that live cell and snapshot the current name/avatar into the existing name-keyed poll state.
- Keep "Use a different name" as a true guest path; guest joins store only the entered string and no profile link.
- Render the canonical badge for profile-backed viewers while retaining the existing string chip for guests.
- Expose the profile directory to consumers and cover profile-first UI, guest fallback, persisted cell identity, and live profile updates in the focused test pattern.

## Compatibility and rollout

This is stacked on #4922 so it includes the secure-runtime and candidate-schema compatibility fixes. The new profile directory is optional and does not rewrite the existing `users`, votes, admin, or history schemas.

Loom #4074 enables the native profile experience, but it does not itself set `fabric_default_host: hosted`. A separate Loom Route A/default-host migration is still required before local Looms resolve `#profile` from Estuary by default. Until then, users without a resolved profile continue through the typed guest-name path.

Existing guest roster entries are intentionally not mutated in the background. New joins use the canonical cell whenever Loom resolves one; an explicit migration/re-link affordance can be added later if old guest entries need upgrading.

## Tests

- `deno task cf check packages/patterns/lunch-poll/main.tsx --no-run`
- `deno task cf check packages/patterns/lunch-poll/participant-identity-card.test.tsx --no-run`
- `deno lint packages/patterns/lunch-poll/main.tsx packages/patterns/lunch-poll/participant-identity-card.tsx packages/patterns/lunch-poll/participant-identity-card.test.tsx`
- Lunch Poll directory tests: 66 sandbox-safe assertions passed; the multi-user test requires loopback access.
- `deno task cf test packages/patterns/lunch-poll/multi-user.test.tsx --root packages/patterns --timeout 15000`: 11 passed.

---
## Summary by cubic
Retains canonical profile identity by storing each profile-backed viewer’s live `#profile` cell in `participantProfiles` and rendering all badges from those stored links. Makes join profile‑first via top‑level `#profile` wishes and an always‑on self‑badge, with an explicit “Continue as guest” path; fixes join deadlocks and “Unknown profile”.

- **New Features**
  - Added `participantProfiles` with `participants: [{ name, profile }]` (dedup with `equals()`); exposed on output; participants strip and header chip render badges from STORED links; guests show as chips.
  - Resolved `#profile`/`#profileName`/`#profileAvatar` at the top level and passed `profile`/`profileName`/`profileAvatar`/`profileSetupUI` into `ParticipantIdentityCard`; profile-backed joins write the live cell; the card exposes `profileName`.

- **Bug Fixes**
  - Bound badges to STORED links in static JSX and added a top‑level self‑badge to prime profile resolution, keeping identity stable across profile switches and preventing blanking.
  - Made option ids vote‑aware and removed `joinedAt`; updated docs (`DEPLOY-AND-SHARE.md`) to include `participantProfiles` and integration tests to click `#lp-guest-button` for the typed‑name path.

<sup>Written for commit 7fafd327d54a9869e28602dd6601f5a47bcd5079. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4928?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/patterns/integration/lunch-poll-vote.test.ts
+++ b/packages/patterns/integration/lunch-poll-vote.test.ts
@@ -166,8 +166,12 @@ describe("lunch poll: two users vote on a shared option", () => {
           ]),
       );
 
-      // Host joins first -> becomes host/admin. The roster chip carrying the
-      // host's name appears once the join lands.
+      // Host joins first -> becomes host/admin. Fresh identities carry no
+      // shared profile, so the join card opens on the profile create/pick
+      // surface; "Continue as guest" reveals the typed-name input this test
+      // drives. The roster chip carrying the host's name appears once the join
+      // lands.
+      await clickCfButton(hostPage, "#lp-guest-button");
       await timer.run(
         "host name filled",
         () => fillCfInput(hostPage, "#lp-join-name", HOST),
@@ -178,9 +182,11 @@ describe("lunch poll: two users vote on a shared option", () => {
         () => waitForText(hostPage, "body", HOST),
       );
 
-      // Guest joins second. The board shows a participant count, not a full
-      // roster, so the host's join landing is observed as "2 joined" (and the
-      // guest's own page shows its name plus "hosted by Alice").
+      // Guest joins second via the same guest path. The board shows a
+      // participant count, not a full roster, so the host's join landing is
+      // observed as "2 joined" (and the guest's own page shows its name plus
+      // "hosted by Alice").
+      await clickCfButton(guestPage, "#lp-guest-button");
       await fillCfInput(guestPage, "#lp-join-name", GUEST);
       await clickCfButton(guestPage, "#lp-join-button");
       await timer.run(
--- a/packages/patterns/lunch-poll/DEPLOY-AND-SHARE.md
+++ b/packages/patterns/lunch-poll/DEPLOY-AND-SHARE.md
@@ -19,9 +19,10 @@ addressed by `(space, causal-cell-id)` — not to "the pattern" in the abstract.
 So "share the state" = "everyone points at the same piece"; "copy the state" =
 "move that piece's values into a new piece." It all lives in **`PerSpace` input
 cells**, shared by everyone in the space: `question`, `options`, `votes`,
-`users`, `adminName`, and **`visits`** (the "Recently eaten" log + embedded vote
-snapshots that feed "Lunch stats"). Plus **`myName`**, which is **`PerUser`**
-(keyed by your DID).
+`users`, `participantProfiles` (the object-wrapped directory of live canonical
+profile links for profile-backed joiners), `adminName`, and **`visits`** (the
+"Recently eaten" log + embedded vote snapshots that feed "Lunch stats"). Plus
+**`myName`**, which is **`PerUser`** (keyed by your DID).
 
 All of these survive an in-place `setsrc` (Option A) and — because `visits` is
 now an ordinary `PerSpace` cell — can all be copied to another piece via the CLI
@@ -132,7 +133,7 @@ MINE=$(deno task cf piece new packages/patterns/lunch-poll/main.tsx \
 
 # 2. Copy each PerSpace field from the canonical piece into yours.
 #    `--input` reads/writes the input cell where these live.
-for field in question users options votes adminName visits; do
+for field in question users options votes participantProfiles adminName visits; do
   deno task cf piece get --piece "$PIECE" -s "$SPACE" "$field" --input -q \
     | deno task cf piece set --piece "$MINE" -s "$SPACE" "$field" --input -q
 done
--- a/packages/patterns/lunch-poll/main.test.tsx
+++ b/packages/patterns/lunch-poll/main.test.tsx
@@ -65,97 +65,81 @@ const COLLIDING_INITIAL_USERS: User[] = [
     name: "Daffodil",
     avatar: "",
     color: "#2f6f4e",
-    joinedAt: 1,
   },
   {
     name: "Dragonfly",
     avatar: "",
     color: "#c2573a",
-    joinedAt: 2,
   },
   {
     name: "Dan",
     avatar: "",
     color: "#3b4a6b",
-    joinedAt: 3,
   },
   {
     name: "Dana",
     avatar: "",
     color: "#a33b35",
-    joinedAt: 4,
   },
   {
     name: "dan",
     avatar: "",
     color: "#b27722",
-    joinedAt: 5,
   },
   {
     name: "A",
     avatar: "",
     color: "#7c3aed",
-    joinedAt: 6,
   },
   {
     name: "a",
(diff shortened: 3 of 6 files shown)


bfollington (commented):
_cf-review via Codex (GPT-5), following a review discussion with @bfollington._

I’d request changes. The storage direction is sound, but the rendering path currently violates the profile guidance and contains a render-breaking binding.

Scope: detail review of all 3 changed files, emitted CTS, profile docs/spec, and exemplars. Checks: `deno fmt --check`, `deno lint`, `cf check`, `git diff --check`, and the focused pattern test (12/12 assertions passed). CI is green.

## 🔴 Blocking

### 1. [critical, verified] `$profile` is inside a computed VNode and can blank the pattern

In `packages/patterns/lunch-poll/main.tsx:1203-1243`, the header is created in `computed(() => …)` and contains `<cf-profile-badge $profile={profileWish.result}>`. The emitted CTS retains that binding inside `__cfLift_21`.

This is the exact forbidden case documented in `docs/common/ai/pattern-critique-guide.md` and `docs/common/patterns/multi-user-patterns.md`: the profile is auto-unwrapped, the renderer reports that `$profile` is not reactive, and the whole render blanks.

Fix: put the binding at a static `[UI]` position, using static `ifElse` branches or a child pattern whose UI owns the binding.

### 2. [major, verified] The stored participant links are not used as the rendered identity source

`viewerUsesCanonicalProfile` (`main.tsx:1114-1121`) only recognizes the current viewer and then binds the viewer’s current `profileWish.result`. Other participants remain name/initial swatches (`main.tsx:1433-1458`), even though their live profile cells are now available.

That contradicts the canonical rule to render every profile-backed participant from their contributed cell. It also means changing the active `#profile` after joining makes the current header fall back to a string, although the originally joined profile cell remains stored.

Fix: treat the stored directory entry as the authoritative rendered identity. Render `entry.profile` for each profile-backed voter/participant, with names retained only as the compatibility join into legacy poll state. Because the vote table is currently produced inside a computed subtree, this likely needs a small static-UI subpattern.

### 3. [major unless explicitly accepted, verified] Missing profiles automatically become typed-name identities

`participant-identity-card.tsx:201-209` treats an unresolved `#profile` as the manual-name path, while the parent never renders `profileWish[UI]`. This bypasses the trusted profile-create/profile-switch lifecycle, and transient resolution can race a user into joining as a guest.

The explicit “Use a different name” escape hatch is reasonable if guests are a product requirement. The automatic fallback is the part that conflicts with the example guidance.

Fix: render `profileWish[UI]` and make “Continue as guest” an explicit secondary choice instead of interpreting every undefined/loading result as a guest.

## 🟡 Improvement

The tests don’t exercise the risky behavior. The new test injects an in-memory profile and checks `profileIdentity.profileName` changes; it never renders the parent header branch, never binds a stored co-participant profile, and never verifies the link after cross-runtime/cross-space serialization. The unchanged multi-user test explicitly uses only the name override. Please add coverage for the static badge render and a second viewer reading/rendering the stored profile cell.

## What is aligned

The PR correctly resolves the full live `#profile` cell, uses an object-wrapped shared directory, compares profile identity with `equals()`, and retains legacy name/avatar snapshots for compatibility.

A later migration should remove the remaining name-keyed identity debt: distinct profiles sharing a display name still cannot both join, and voting/admin state continues to identify people by mutable names.

**Verdict: request changes.** The storage foundation is good, but this is not yet a best-practice profile example, and the header binding is a runtime correctness bug.

mathpirate:
Thanks for the thorough review — all three blocking findings and the test gap are addressed at `9905623bc` (also rebased past #4959 and verified live end-to-end).

**1. `$profile` inside a computed VNode (critical).** The header cluster is now plain static JSX — the HOST/viewer conditionals are ternaries (which lower to static-branch `ifElse`), and each identity badge is a `.map` row that compiles to its own sub-pattern binding `$profile={entry.key("profile")}`. Verified in the emitted CTS that no `$profile` binding sits inside a lift. The render no longer blanks.

**2. Stored links as the rendered identity source (major).** Rendering now comes from the **stored** `participantProfiles` directory, never the live wish: a participants strip renders every profile-backed joiner from their stored cell, and the viewer's header chip binds the viewer's stored entry — so switching the active `#profile` after joining keeps the joined identity. (While addressing this I also found and fixed two defects the pattern tests couldn't catch: a join deadlock — the card gated its join affordance on the `#profileName` string, but that string only materializes once a badge renders, and the badge was gated on the same string — and an "Unknown profile" bug where an unset optional cell input reads as a truthy proxy, so a `profile ?? wish.result` seam returned the empty proxy. Both are fixed by resolving `#profile` / `#profileName` / `#profileAvatar` at the pattern top level with a minimal schema, binding badges to the wish result directly, and snapshotting the name at join — the shared-profile-rosters spec idiom.)

**3. No automatic typed-name fallback (major).** When no profile resolves, the card now renders the `#profile` wish's built-in create/pick `[UI]`, and "Continue as guest" is an explicit secondary action — the typed-name path never appears automatically.

**4. Test coverage.** Profile-backed *rendering* moves to the browser/integration tier (the scrabble/battleship precedent): a unit test can't honestly inject a resolvable `#profile` cell, and the unset-optional-cell-input-is-a-truthy-proxy behavior means an injection seam is itself a footgun. The join *logic* — name snapshot, `equals()`-keyed dedup, directory write, and the create-vs-join-vs-guest gating — stays covered in `participant-identity-card.test.tsx` via direct injection. 79 assertions, 0 warnings. Verified live: join works, board renders, host/roster correct, real profile badges throughout.

One honest correction to my own earlier notes here: I'd floated a "profile fields aren't materialized cross-space" concern — that was overblown; the badges resolve cold fine, so there's no platform blocker. The separate `optional cell input ?? fallback` footgun is real and I've written it up as a standalone fix-brief.

<sub>Drafted by Claude (Fable 5) on @mathpirate's behalf.</sub>

PR #4931, opened 2026-07-23
feat(cf-harness): support Codex subscriptions

seefeldb:
## Summary

- add an explicit `openai-codex` model provider using the OpenCode/pi-compatible ChatGPT subscription OAuth and Responses contracts
- preserve the existing bounded `CfHarnessPromptLoop`, CFC tool mediation, transcript ownership, artifacts, resume, interactive service, and subagent behavior
- add browser PKCE and device login, owner-scoped credential storage and serialized refresh, CLI auth/model commands, and fixed-origin SSE transport
- add Loom manifest and trusted-host injection contracts so per-user subscription credentials cannot come from ambient local state

## User impact

Local users can explicitly connect a supported ChatGPT/Codex subscription and run `cf-harness` without an OpenAI Platform API key. Loom can use the same provider after its trusted host supplies an authenticated owner-bound credential resolver and product policy/UI integration.

The existing OpenAI-compatible gateway remains the default and subscription failures never fall back to API keys or unauthenticated requests.

## How to use

From the package directory, connect the local CLI to a ChatGPT/Codex subscription. Browser PKCE is the default; use the device flow on a headless machine:

```bash
cd packages/cf-harness

deno task run -- auth login openai-codex
# Or, for a headless terminal:
deno task run -- auth login openai-codex --device
```

Check the connection and query the live model catalog for that subscription, then pass one of those model ids to a run:

```bash
deno task run -- auth status openai-codex
deno task run -- models openai-codex

deno task run -- \
  --workspace ../.. \
  --model-provider openai-codex \
  --model <model-id-from-the-catalog> \
  --prompt "Summarize the cf-harness package structure."
```

Resume a root run from its run directory or `run-state.json`. Codex resume keeps the recorded provider, model, exact credential owner, and encrypted continuation; child-run artifacts must be resumed through their root run:

```bash
deno task run -- \
  --workspace ../.. \
  --resume-run <run-root-or-run-state.json>
```

Disconnect only the local cf-harness credential with:

```bash
deno task run -- auth logout openai-codex
```

Credentials are stored under `CF_HARNESS_HOME` (default `~/.cf-harness/auth.json`), not imported from `~/.codex/auth.json`. Selecting `openai-codex` does not require an OpenAI Platform API key and never falls back to an API key, the gateway provider, or an unauthenticated request after an OAuth failure. The provider changes only authentication and model transport; the existing bounded prompt/tool loop, tool dispatch, transcripts, artifacts, resume behavior, subagents, and CFC mediation remain shared.

### Loom integration

This PR supplies the Loom integration seam, not an ambient service credential. A trusted Loom host selects `modelProvider: "openai-codex"`, includes a non-secret `cf-harness.credential-owner-ref` for the authenticated initiating user in the run manifest, and injects a resolver/model client bound to that exact owner and tenant. The resolver must use the Loom encrypted secret backend; tokens must not enter manifests, Cells, Spaces, stdio, session databases, command lines, or run artifacts. Owner or tenant mismatches fail closed.

Product enablement still requires the Loom host to authenticate the initiating principal, implement connect/status/disconnect UI, enforce workspace policy, and verify cross-user isolation. It uses the same Codex protocol and shared harness loop once those host responsibilities are wired.

## Validation

- `deno task test` in `packages/cf-harness`: 500 passed
- `deno check packages/cf-harness/mod.ts packages/cf-harness/src/main.ts`
- `deno lint packages/cf-harness`
- `deno fmt --check packages/cf-harness`
- `deno task check-docs`: 519 code blocks passed
- `deno task check-unused-deps`
- `git diff --check`

The aggregate root `deno task check` cannot start in this workspace because it has Deno 2.9.3 while the repository guard requires `>=2.8.0 <2.9.0`.

NEW_PERF_BASELINE: coverage-debt: packages/cf-harness uncovered lines = 3828 lines

## Rollout boundary

Live-account browser/device smoke tests, security review, and Loom's encrypted credential backend, workspace-policy verification, and connect/status/disconnect UI remain explicit rollout gates in the live implementation plan.


seefeldb:
Addressed all first-round Codex/Cubic findings in 7cab7f924: tenant-complete Loom owner binding and resume invariants; cross-process credential locking plus permission/symlink/prototype hardening; abort/SSE cancellation fixes; legacy report compatibility; provider discovery cancellation; and corrected protocol/rollout docs. Added focused regressions and ran the complete cf-harness suite: 469 passed. The repo-wide local check is version-gated because this checkout has Deno 2.9.3 while the repo requires >=2.8,<2.9; GitHub Actions exercises the pinned version.

seefeldb:
Addressed the queued-refresh cancellation finding in 208b81952: the credential-store update contract now accepts an AbortSignal, in-memory and file-store queue acquisition is abortable, file-lock acquisition receives cancellation, and regressions verify an aborted waiter never runs and does not poison later updates. Full cf-harness suite: 476 passed.

seefeldb:
Second sol-max adversarial review completed and all findings addressed in e51c01e2b. Highlights: secret-free error cause graphs; authoritative provider/owner reconciliation on resume; exact owner+tenant client binding for interactive/Loom use; legacy HarnessConfig source compatibility with strict resolved config; cancellation preserved across credential reads, PKCE, exchange, response bodies, SSE cleanup, and cross-process file-lock waits; malformed credential stores now fail closed; protocol/resume docs corrected. Full cf-harness suite: 487 passed; affected lint/type checks pass.

packages/cf-harness/src/auth/openai-codex.ts
```diff
@@ -0,0 +1,646 @@
...
+      },
+    );
+  } catch {
+    if (options.signal?.aborted) throw abortReason(options.signal);
+    throw new OpenAICodexAuthError(
+      "network",
+      "OpenAI Codex token exchange failed before receiving a response",
+    );
+  }
+  const credential = await readTokenResponse(
```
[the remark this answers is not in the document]
seefeldb: Fixed in 4e71118b8. Once a successful refresh response has yielded rotated credentials, the store transaction now persists them before surfacing a late abort. Added a regression proving the caller receives its exact abort reason while the rotated refresh token remains stored.

packages/cf-harness/src/model/openai-codex-responses.ts
```diff
@@ -0,0 +1,671 @@
...
+  readonly credentialOwner?: HarnessCredentialOwnerRef;
+  readonly #resolver: OpenAICodexCredentialResolverLike;
+  readonly #fetchFn: HarnessFetch;
+  readonly #endpoint: string;
+  readonly #now: () => Date;
+
+  constructor(options: OpenAICodexResponsesClientOptions) {
+    this.#resolver = options.credentialResolver;
+    const credentialOwner = options.credentialOwner ??
+      options.credentialResolver.credentialOwner;
```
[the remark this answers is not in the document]
seefeldb: Fixed in 4e71118b8. The Responses client now rejects both full owner/tenant mismatches and resolver ownerKey mismatches before exposing a credentialOwner binding. Added constructor regressions for both cases.

packages/cf-harness/src/contracts/run-manifest.ts
```diff
@@ -21,6 +21,23 @@ export interface LoomRunManifestCfc {
...
+  "cf-harness.credential-owner-ref" as const;
+
+export interface HarnessCredentialOwnerRef {
+  type: typeof HARNESS_CREDENTIAL_OWNER_REF_TYPE;
+  version: 1;
+  ownerKey: string;
+  tenantKey?: string;
+}
+
+export const harnessCredentialOwnersEqual = (
```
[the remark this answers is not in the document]
seefeldb: Fixed in 4e71118b8. Removed the CLI-local equality implementation and now reuse harnessCredentialOwnersEqual at every CLI binding check.

mathpirate (commented):
I reviewed head `4e71118b8ac9bf7c586288b5dd24cad89573b43d`.

This is a thoughtful implementation overall: the provider separation, fixed-origin transport, credential-store locking, cancellation handling, and existing gateway compatibility are all carefully handled. The package suite and GitHub checks are green.

I found six areas worth addressing or explicitly tracking:

1. The exported prompt-loop boundary does not verify that an injected Codex client belongs to the engine/run owner.
2. Child artifacts lose exact Loom owner and lineage information, and can currently be resumed incorrectly as top-level runs.
3. Cross-model Codex resume replays model-bound continuation state; this should fail explicitly until cross-model sanitization is implemented.
4. Refresh responses incorrectly require refresh-token rotation.
5. Refusal-only Responses output becomes an empty successful assistant turn.
6. Unbounded run IDs are sent directly as `prompt_cache_key`.

I’ve left detailed inline comments with suggested changes and regression coverage for each.

Live browser/device OAuth, refresh, and Responses smoke tests remain rollout considerations beyond these code-review findings.

packages/cf-harness/src/prompt-loop.ts
```diff
@@ -1786,14 +1640,37 @@ export class CfHarnessPromptLoop {
...
+          baseUrl: this.engine.config.gatewayBaseUrl,
+          authMode: this.engine.config.gatewayAuthMode,
+          apiKey: options.apiKey,
+          apiKeySource: options.apiKeySource,
+          fetchFn: options.fetchFn,
+        });
+    } else {
+      this.#gatewayClient = options.gatewayClient;
+    }
+    if (options.modelClient !== undefined) {
```
mathpirate, line 1656: [P1] Validate the owner-bound client at this public boundary

For `openai-codex`, this constructor verifies only that the injected client's `providerId` matches the configured provider. It does not require `modelClient.credentialOwner` or compare it with the engine/run owner.

That allows a library or Loom host to construct an engine whose manifest identifies owner B while injecting a model client bound to owner A. Requests would use A's subscription while the resulting artifacts and audit context attribute the run to B.

Please fail closed here before any model request:

- require `credentialOwner` on an `openai-codex` client;
- when the run manifest has a credential owner, compare the complete reference with `harnessCredentialOwnersEqual`, including `tenantKey`;
- otherwise, at minimum compare `credentialOwner.ownerKey` with the configured `credentialOwnerKey`;
- leave gateway/custom-provider behavior unchanged.

Tests should cover a missing client owner, the same `ownerKey` under different tenants, a different owner key, and an exact match.

packages/cf-harness/src/prompt-loop.ts
```diff
@@ -2707,6 +2604,8 @@ export class CfHarnessPromptLoop {
     const maxModelTurns = delegateInput.maxModelTurns ??
       profileConfig.maxModelTurns;
     const parentRunState = this.engine.getRunState();
+    const modelProvider = parentRunState.modelProvider ??
+      this.engine.config.modelProvider;
     const subagentSequence = nextSubagentSequence(parentRunState);
     const childRunId = `${parentRunState.runId}.subagent.${subagentSequence}`;
     const childEngine = new CfHarnessEngine({
```
mathpirate, line 2648: [P1] Persist child lineage and exact credential provenance

The child engine inherits only `credentialOwnerKey`. Its persisted run state does not identify itself as a child and loses the parent's full credential-owner reference, `tenantKey`, and Loom source provenance. The `HarnessSubagentRunManifest` that contains the parent relationship is currently retained by the parent, but is not part of the child's own persisted identity.

The intended resume contract should be:

- a child artifact is never booted directly as a top-level agent;
- resuming the root session restores or accounts for its complete subagent tree.

Please persist explicit, typed lineage in the child state rather than inferring it from the run ID, for example:

```ts
{
  role: "subagent",
  rootRunId,
  parentRunId,
  parentToolCallId,
  depth,
}
```

The child should also retain the exact credential-owner binding and Loom provenance, including `tenantKey`.

One persistence-order detail matters for future whole-session restoration: the parent currently records its `HarnessSubagentRunRef` only after the child finishes. A process interruption during child execution can therefore leave an orphaned child that the parent does not know about. Please record the relationship before starting the child, with a pending/running state, and finalize it afterward. Active-child continuation does not need to be implemented here, but the relationship should survive an interruption.

packages/cf-harness/src/cli.ts
```diff
@@ -1949,6 +2237,66 @@ export const runCfHarnessCli = async (
     if (parsed.resumeRun !== undefined) {
       const readRunArtifacts = deps.readRunArtifacts ?? readHarnessRunArtifacts;
       const artifacts = await readRunArtifacts(parsed.resumeRun);
```
mathpirate, line 2239: [P1] Reject subagent artifacts before top-level resume resolves credentials

`--resume-run` currently accepts any run artifact. Because child states do not carry their Loom manifest or full owner reference, passing a child artifact here reconstructs a local owner:

```ts
recordedRunManifest?.credentialOwner ??
  localCredentialOwner(runState.credentialOwnerKey ?? "local")
```

This can collapse identical owner keys across tenants and can fall through to the operator's filesystem credential when the recorded key is `"local"`.

Once child lineage is persisted, please reject `role: "subagent"` immediately after reading the artifact and before creating an owner, credential resolver, or model client. The error should direct the caller to the root run, for example:

> Cannot resume subagent run X as a top-level run; resume root run Y instead.

A regression test should verify both the error and that credential resolution/network access were never attempted.

packages/cf-harness/src/engine.ts
```diff
@@ -312,7 +313,38 @@ export class CfHarnessEngine {
 
   constructor(options: CreateHarnessEngineOptions = {}) {
     this.#now = options.now ?? (() => new Date().toISOString());
-    this.config = resolveHarnessConfig(options);
+    const recordedProvider = options.runState?.modelProvider ??
+      "openai-compatible-gateway";
+    if (
```
mathpirate, line 353: [P2] Reject cross-model Codex resume at the engine boundary

A resumed run may currently override its recorded model. For `openai-codex`, the adapter then replays encrypted reasoning and remembered `fc_*` item IDs without knowing whether they came from the requested model. Those values are model-bound and can trigger Responses pairing-validation failures.

The pinned pi implementation deliberately [drops encrypted reasoning across models](https://github.com/earendil-works/pi/blob/9b3a2059171bcc74ad9d2cadeea6d186776cf2db/packages/ai/src/api/transform-messages.ts#L95-L109) and [omits model-bound `fc_*` IDs](https://github.com/earendil-works/pi/blob/9b3a2059171bcc74ad9d2cadeea6d186776cf2db/packages/ai/src/api/openai-responses-shared.ts#L170-L224).

Until cf-harness implements equivalent sanitization, please reject a Codex resume when the explicitly requested model differs from `runState.model`. This should be enforced in the engine/library invariant, not only CLI parsing, so Loom and other callers cannot bypass it. The error should include the recorded and requested model and should occur before credential resolution or provider traffic.

To preserve a clean path toward future support, please also version the Codex continuation state and record its `sourceModel` now. A future implementation can then safely decide which opaque fields may be replayed.

Tests should cover:

- same-model Codex resume succeeds;
- different-model Codex resume fails before credential/network access;
- gateway cross-model behavior remains unchanged.

packages/cf-harness/src/auth/openai-codex.ts
```diff
@@ -0,0 +1,646 @@
...
+  let json: TokenResponse;
+  try {
+    json = await response.json() as TokenResponse;
+  } catch {
+    throw new OpenAICodexAuthError(
+      "malformed_response",
+      `OpenAI Codex token ${operation} returned invalid JSON`,
+    );
+  }
+  if (
```
mathpirate, line 190: [P2] Preserve the existing refresh token when the response does not rotate it

This parser is shared by initial token exchange and refresh, and requires `refresh_token` in both cases. Refresh-token rotation is optional: a successful refresh may return a new access token while leaving the existing refresh token valid.

The pinned official Codex implementation models the refreshed token fields as optional and retains the previous refresh token when the response omits one: [manager.rs](https://github.com/openai/codex/blob/88f1cd9664d09b68909a258a061a662c1f099ce6/codex-rs/login/src/auth/manager.rs#L1308-L1328).

Please split initial-exchange and refresh parsing, or pass the current credential into refresh parsing. During refresh, use a returned `refresh_token` when present and otherwise preserve the latest stored token. Rotation should still be persisted atomically when it occurs.

Please add coverage for both rotating and non-rotating successful refresh responses.

packages/cf-harness/src/model/openai-codex-responses.ts
```diff
@@ -0,0 +1,691 @@
...
+  const toolCalls: HarnessToolCall[] = [];
+  const toolCallById = new Map<string, HarnessToolCall>();
+  const continuation: ResponsesInputItem[] = [];
+  const functionCallItemIds: Record<string, string> = {};
+  for (const rawItem of output) {
+    if (
+      typeof rawItem !== "object" || rawItem === null || Array.isArray(rawItem)
+    ) continue;
+    const item = rawItem as Record<string, unknown>;
+    if (item.type === "message" && Array.isArray(item.content)) {
```
mathpirate, line 326: [P2] Do not convert refusal-only output into an empty successful response

Completed Responses messages may contain content with `type: "refusal"`. This parser only appends `output_text`, so a refusal-only response currently produces an empty assistant message with no tool calls. The prompt loop then records it as a successful `assistant_completed` turn, hiding the actual model outcome.

The pinned adapter also explicitly handles `response.refusal.delta`: [openai-responses-shared.ts](https://github.com/earendil-works/pi/blob/9b3a2059171bcc74ad9d2cadeea6d186776cf2db/packages/ai/src/api/openai-responses-shared.ts#L494-L503).

Please surface refusal content as assistant-visible text or classify it explicitly as a provider refusal. It should never become an unexplained empty success. Please add a refusal-only regression fixture.

packages/cf-harness/src/model/openai-codex-responses.ts
```diff
@@ -0,0 +1,691 @@
...
+    const body = JSON.stringify({
+      model: request.model,
+      store: false,
+      stream: true,
+      instructions: converted.instructions,
+      input: converted.input,
+      ...(responseTools.length > 0 ? { tools: responseTools } : {}),
+      text: { verbosity: "low" },
+      include: ["reasoning.encrypted_content"],
+      prompt_cache_key: request.runId,
```
mathpirate: [P2] Derive a bounded provider identifier instead of sending `runId` directly

`runId` is a public engine input. Artifact validation restricts its characters but does not bound its length, and subagent IDs append additional suffixes. Sending it directly as `prompt_cache_key` can therefore turn an otherwise valid run into a provider request error.

The pinned implementation caps OpenAI prompt-cache keys at 64 characters: [openai-prompt-cache.ts](https://github.com/earendil-works/pi/blob/9b3a2059171bcc74ad9d2cadeea6d186776cf2db/packages/ai/src/api/openai-prompt-cache.ts#L1-L8).

Please derive a stable provider-facing identifier with a maximum of 64 characters. A readable prefix plus digest would preserve stability without the collisions introduced by simple truncation. Use the derived value consistently wherever run/session affinity is intended, while respecting any separate format requirements for request IDs.

Please add a regression test using a long valid run ID, including a child-run suffix.

packages/cf-harness/src/prompt-loop.ts
seefeldb: Addressed in 15007082e. The shared prompt-loop constructor now requires a credentialOwner for openai-codex clients, compares the full manifest owner with harnessCredentialOwnersEqual (including tenantKey), and otherwise compares ownerKey. The regression covers missing owner, different key, same key across different tenants, and exact match.

packages/cf-harness/src/prompt-loop.ts
seefeldb: Addressed in 15007082e. Child run state now persists typed root/parent lineage and inherits the full Loom run manifest, including the exact owner and tenant provenance. The parent persists a running child ref before constructing/running the child loop and replaces it with the terminal ref afterward. The artifact regression reads the parent during the first child request and verifies both sides.

packages/cf-harness/src/cli.ts
seefeldb: Addressed in 15007082e. CLI resume rejects subagent lineage immediately after artifact loading with guidance to resume the root. The regression verifies the exact error and that model-client construction was never reached.

packages/cf-harness/src/engine.ts
seefeldb: Addressed in 15007082e. CfHarnessEngine now rejects an explicitly different model for openai-codex resume before client construction/provider traffic while leaving gateway cross-model behavior unchanged. Codex continuation state is versioned and records sourceModel; adapter replay is model-bound. Engine and CLI regressions cover same-model, cross-model, early rejection, and gateway behavior.

packages/cf-harness/src/auth/openai-codex.ts
seefeldb: Addressed in 15007082e. Refresh parsing now uses a returned refresh_token when present and preserves the latest stored token only when the refresh response omits it; initial exchange still requires the field. Rotating and non-rotating refresh tests pass.

packages/cf-harness/src/model/openai-codex-responses.ts
seefeldb: Addressed in 15007082e. Completed refusal content is surfaced as assistant-visible text, so refusal-only responses no longer become empty successful completions. Added a refusal-only SSE regression fixture.

packages/cf-harness/src/model/openai-codex-responses.ts
seefeldb: Addressed in 15007082e. Provider affinity now uses a stable readable-prefix plus SHA-256 digest key capped at 64 characters via @commonfabric/content-hash. The same key drives prompt_cache_key and session-id, while x-client-request-id is a fresh UUID. The long parent plus child-suffix regression verifies length, stability, and request-ID format.

mathpirate (approved):
ty for doing this!

packages/cf-harness/src/prompt-loop.ts
```diff
@@ -1786,14 +1641,65 @@ export class CfHarnessPromptLoop {
...
+        !harnessCredentialOwnersEqual(clientOwner, manifestOwner)
+      ) {
+        throw new Error(
+          "openai-codex model client does not match run manifest credential owner",
+        );
+      }
+      const configuredOwnerKey = runState.credentialOwnerKey ??
+        this.engine.config.credentialOwnerKey;
+      if (
+        manifestOwner === undefined && configuredOwnerKey !== undefined &&
```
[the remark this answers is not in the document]
seefeldb: Addressed in 92485c736. The configured/run ownerKey check is now unconditional for openai-codex, in addition to the exact manifest owner and tenant comparison. The existing owner-binding regression now covers an exact client/manifest owner paired with a conflicting configured owner key.

docs/plans/cf-harness-codex-subscription-auth.md
```diff
@@ -0,0 +1,628 @@
...
+- [x] Extend assistant transcript records with optional type-tagged provider
+  continuation state.
+- [x] Persist encrypted reasoning/response continuation only as opaque data and
+  return it only to the same provider.
+- [x] Define the compatibility behavior for old transcripts that lack provider
+  state: resend ordinary visible history without inventing reasoning items.
+- [x] Ensure cross-provider resume either strips incompatible provider state
+  explicitly or rejects the switch; it must not serialize Codex-only fields
+  into a gateway request.
+- [x] Version continuation state, bind it to the source model, and reject
```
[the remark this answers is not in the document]
seefeldb: Addressed in 92485c736. Codex transcript conversion and continuation source-model validation now run before credential resolution; the new direct-adapter regression supplies model-bound continuation from another model and verifies zero credential resolutions and zero provider calls. The documented ordering is now accurate.

packages/cf-harness/src/engine.ts
```diff
@@ -312,7 +317,47 @@ export class CfHarnessEngine {
...
+      options.runState !== undefined && options.modelProvider !== undefined &&
+      options.modelProvider !== recordedProvider
+    ) {
+      throw new Error(
+        `resumed run provider ${recordedProvider} does not match requested provider ${options.modelProvider}`,
+      );
+    }
+    if (
+      options.runState !== undefined && recordedProvider === "openai-codex" &&
+      options.model !== undefined && options.runState.model !== undefined &&
```
mathpirate, line 363: [P2] Keep the resumed Codex model binding at the public library boundary

This check compares models only when the requested model is passed into `CfHarnessEngine`. A library caller can construct an engine from a Codex `runState` without `options.model`, then pass a different model to `CfHarnessPromptLoop.runTranscript`; because that method prefers its own `options.model`, a plain transcript can be sent with the different model and has no `sourceModel` continuation for the adapter to reject.

The new adapter preflight correctly catches an opaque continuation mismatch before resolving credentials, but it does not cover this no-continuation library path. Please retain whether an engine was created from resumed state and validate the model selected by `runTranscript` against the recorded model before invoking the model client. Fresh runs can still select a model, but that selection should be persisted into `runState.model` before the first turn. A regression should exercise this exported library path rather than only the CLI/engine-constructor path.

packages/cf-harness/src/engine.ts
```diff
@@ -312,7 +317,47 @@ export class CfHarnessEngine {
 
   constructor(options: CreateHarnessEngineOptions = {}) {
     this.#now = options.now ?? (() => new Date().toISOString());
-    this.config = resolveHarnessConfig(options);
+    const recordedProvider = options.runState?.modelProvider ??
```
mathpirate, line 351: [P2] Enforce child resume context in the public engine

The CLI now rejects direct child resume, but exported library callers can still construct this engine from a `runState` with `lineage.role === "subagent"` and execute it through a normal prompt loop. The inherited full manifest fixes the previous local-credential fallback, but the broader contract—children resume only through their root session—is not enforced for Loom or other library callers.

Please reject resumed subagent state unless the caller supplies an explicit trusted parent/root resume context. New child execution remains distinguishable because it supplies `lineage` without `runState`; a future whole-tree restorer can provide that explicit context. Please add a library-level regression proving direct child-as-root construction fails while ordinary child creation still works.

mathpirate (commented):
I re-reviewed Berni’s follow-up through `92485c73693f63d5430e15bfd0d1795af6ec6880`.

The rest of the original review feedback is addressed: exact owner/tenant binding, persisted child lineage and running refs, refresh without token rotation, refusal handling, and bounded provider affinity/request IDs all look good, with the new regression coverage passing locally.

I found two remaining public-library paths that bypass the new CLI/constructor guards, and left inline comments for them below.

packages/cf-harness/src/engine.ts
seefeldb: Fixed in 2dcd241a2. `CfHarnessPromptLoop.runTranscript()` now binds the selected model through `CfHarnessEngine.bindRunModel()` before diagnostics, persistence, credential resolution, or model invocation. Engines reconstructed from persisted Codex state reject a different transcript-level model; fresh engines may make one explicit first selection (even over a configured default), persist it into run state, and cannot switch it on later turns. The exported-library regression verifies both the zero-model-call resumed rejection and fresh-selection persistence.

packages/cf-harness/src/engine.ts
seefeldb: Fixed in 2dcd241a2. `CfHarnessEngine` now rejects persisted state with subagent lineage unless the caller supplies a typed `cf-harness.subagent-resume-context` whose root, parent, and delegating tool call match the persisted lineage. New child construction still uses `lineage` without `runState` and remains valid. The engine-level regression covers direct rejection, mismatched trusted context, matching whole-tree restoration context, and ordinary child creation.

PR #4936, opened 2026-07-23
fix(runner,patterns): scope CFC additive-required-default to durable data (estuary home heal) — proposal + partial

bfollington:
> **Draft / for review — a proposal + a validated partial fix, not a merge-and-forget.** The invariant change needs sign-off from the CFC owners (@seefeldb, @mathpirate) before it lands.

## Why

Estuary home spaces are bricked for existing users but **fine for a fresh identity** — the tell that this is a schema-**evolution** problem over a heterogeneous population of old stored docs, not a single bug. When #4926's cold-start setup repair materializes current `home.tsx` over an older-vintage home root, CFC's `mergeRequired` rejects the commit: `required field <name> needs a default to preserve old documents` (`cfc-relevant-transaction-not-prepared`).

That guard protects durable document data. But it fires on **every** required field — including fields that hold no document data: **handler streams** (`asCell:["stream"]`, runtime-materialized capabilities re-created on every setup) and framework projection keys. A handler-rich pattern therefore **can't be healed field-by-field** — default every data field and the guard just advances to the first stream, which by construction can't carry one. Because it fails on the *first* offending field and every home is a different vintage, it presents as a **moving target** (favorites here, `defaultProfile` there, a stream elsewhere): an over-broad invariant that bricks any system-pattern evolution over old docs.

**Full rationale — the two defects, the recreate/no-recreate analysis, the population-test acceptance criterion, and the testing gap — is in the proposal doc in this PR:** `docs/development/proposals/cfc-additive-required-durable-data-scope.md`. Please read that; the summary below is the short version.

## What changed

- **`runner/cfc/schema-merge.ts`** — `mergeRequired` exempts stream slots from the additive-required-default requirement. A stream has no old value to preserve and no meaningful default. Plain data fields without a default still throw (negative case pinned). Emitted schemas unchanged; **no authorization guarantee touched** — this is a data-preservation check, not `writeAuthorizedBy`/`ifc`.
- **`patterns/system/home.tsx`** — `defaultProfile?`. It was emitted required-with-no-default because the schema-generator keys `required` off the `?` token, **not** the `| undefined` value type — a second, distinct source of the same symptom.
- **`runner/test/cfc-additive-default-preserves-old-doc.test.ts`** — the **first repro with CFC enforcement ON**, closing the gap that let #4926/#4933 pass (their harness runs `cfcEnforcementMode: "disabled"`).

## ⚠️ This is NOT the complete heal

Fixes the stream + `defaultProfile` classes. Explicitly **not** proven to heal every vintage:
- The incident reporter's exact `favorites` throw **could not be reproduced locally** — a synthetic pre-favorites doc defaults favorites cleanly. At least one vintage-specific mechanism remains unexplained (candidate: production-compile / durable-compile-cache / `eagerSourceAnnotation=false`; needs the real stored schema to diagnose).
- Additive-required is one CFC invariant; others may reject other vintages.

**Acceptance = the population test** (proposal §): materialize current `home.tsx` over a corpus of real stored home-root schemas with enforcement on, enumerate which heal. Don't treat this as done until that's green.

## Does anyone have to recreate their space?

For every failure mode observed: **no** — the failing fields are all *new* (absent from the old doc), so no data conflict, nothing to discard; the heal is in-place on next load, no data loss. Only a genuinely incompatible/corrupted stored state would force a recreate, and we have zero evidence of that. (Detail in the proposal.)

## Test plan

- Enforcement-on repro: 3 steps green (fails without the fix).
- CFC/writer/profile safety suite: 61 steps green.
- `deno check` / `fmt` / `lint` clean.

### DIFF
--- a/docs/development/proposals/cfc-additive-required-durable-data-scope.md
+++ b/docs/development/proposals/cfc-additive-required-durable-data-scope.md
@@ -1,0 +1,72 @@
+# Scope CFC's additive-required-default rule to durable document data
+
+*CFC refuses to materialize a newer pattern over an older stored document when the new schema adds a required field with no default. That guard is meant to protect real document data from vanishing — but it currently fires on fields that hold no document data at all (handler streams, framework projection keys), which makes evolving a system pattern silently brick every older-vintage stored doc. Scope the invariant to durable data.*
+
+**Status:** proposed · needs sign-off from the CFC invariant owners (**@seefeldb**, **@mathpirate**) before the rule change lands · **Updated:** 2026-07-23
+
+Grew out of the live Estuary "home space is bricked" incident. Sits behind the deployed heal chain: #4900 / #4926 (cold-start setup repair) and #4933 (data-field defaults). This is the next — and we believe more general — layer.
+
+---
+
+## Why
+
+Estuary home spaces are bricked for existing users but **fine for a fresh identity**. That asymmetry is the whole tell: a brand-new home materializes cleanly, but an *existing* home root — a stored document written by some older vintage of `home.tsx` — fails when the current `home.tsx` is materialized over it during the cold-start setup repair.
+
+The failure is CFC enforcement rejecting the setup commit:
+
+```
+CFC enforcement rejected commit: relevant transaction was not prepared:
+required field <name> needs a default to preserve old documents
+(cfc-relevant-transaction-not-prepared)
+```
+
+thrown from `mergeRequired` in `packages/runner/src/cfc/schema-merge.ts`. The rule: when merging a candidate (new) schema over a stored (old) one, **any field that is required in the candidate, absent-and-not-required in the stored doc, and carries no `default`** is rejected — because an old document that predates the field would have no value for it and the merge cannot synthesize one.
+
+That rule is correct *for durable document data*. The bug is that it applies to **every** required field, including fields that hold no preservable data:
+
+- **Handler streams** (`asCell: ["stream"]`) — a stream is a runtime-materialized capability marker, not stored data. Pattern setup re-creates every stream on each run, so an old doc that lacks one has nothing to preserve and there is no meaningful default a `Stream<…>` could ever declare. A handler-rich pattern like `home.tsx` therefore **cannot** be healed field-by-field: give every data field a `Default<>` and the guard just advances to the first handler stream, which by construction can't have one.
+- **Framework projection keys** (`$NAME` / `$UI`) — same category: injected by the runtime, not document data.
+
+Because the check fails on the *first* offending field and every user's home is a different vintage, this presents as a **moving target**: favorites for one stored doc, `defaultProfile` for another, a handler stream for a third. It is not one bug with one field — it is an **over-broad invariant that makes any system-pattern evolution brick every older stored document**, one field at a time, until the rule is scoped correctly.
+
+## The two defects
+
+This incident exposed two distinct mechanisms that both produce "additive required field with no default." A complete fix names both.
+
+1. **The invariant over-applies to non-data fields.** Streams and framework keys are required in the emitted schema but hold no preservable document data. The additive-required-default guard should not consider them. *(This proposal's core change.)*
+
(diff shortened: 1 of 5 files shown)


bfollington (commented):

PR #4936 at `78bc993d5` · detail mode, all 5 files deep-read plus linked #4900/#4926/#4933 context · 72 focused test steps, transformed-schema inspection, type-check, fmt, and lint.

## 🔴 Blocking

1. **`packages/runner/src/cfc/schema-merge.ts:338-340` — `includes("stream")` does not identify a stream *slot* according to `asCell` semantics (verified).**

   `asCell` is an ordered wrapper stack: the first entry is the outer slot, and entries may be strings or scoped objects. The new predicate searches every literal string instead. That produces both false negatives and false positives:

   - The schema generator emits the valid required field `PerUser<Stream<number>>` as `asCell: [{ kind: "stream", scope: "user" }]`; the proposed merge still throws `required field events needs a default`.
   - A schema with `asCell: ["cell", "stream"]` is accepted by the new exemption even though the outer slot is a cell, not a stream. The generator rejects `Cell<Stream<T>>`, but the runner's generic schema utilities explicitly handle this representation, so the CFC merge should not silently reinterpret it.

   Please normalize and inspect only `asCell[0]` (`entry === "stream"` or object-form `entry.kind === "stream"`), and pin both the scoped-stream case and the nested non-stream-outer case. This keeps the invariant change exactly at the intended wrapper boundary.

## 🟡 Improvements

1. **`packages/patterns/system/home.tsx:72-75` — the adjacent comment still states the assumption this PR disproves (verified).**

   It says `defaultProfile` is non-required because it is `… | undefined`, but the fix adds `?` precisely because that union did *not* remove the field from `required`. The transformed result schema correctly excludes `defaultProfile`; please update the comment to attribute that to the explicit optional marker so the next edit does not reintroduce this incident.

2. **`schema-merge.ts:335-363` and proposal lines 25-45 — distinguish persisted stream identity from ephemeral payload (verified coherence issue).**

   The live space-model docs say `{ $stream: true }` is stored and persists to preserve stream identity; only event payloads are ephemeral. The change can still be sound for a *newly added* stream because there is no prior marker to preserve and setup deterministically materializes it, but “not stored data” / “has no old value” is too broad a premise for CFC-owner sign-off. Reword the rationale around stateless, setup-reconstructible marker data, and avoid grouping `$NAME` with `$UI` as non-document data without separately establishing that invariant.

3. **`packages/runner/test/cfc-additive-default-preserves-old-doc.test.ts:177-193` — functionally pin the dependency on setup rematerialization (verified coverage gap).**

   The end-to-end test proves the setup commit landed by reading `$NAME` and `favorites`, but those assertions can pass even if the newly exempted required handler markers are absent. Since this saga's immediately preceding failure was exactly a missing `$stream` marker, send one recovered home handler and assert its state transition (or otherwise assert a required handler is a real stream) under enforcement. That makes the test guard the full principle: CFC permits the schema evolution *because setup has recreated a usable endpoint*.

Current CI has one failing `Pattern Integration Tests (2/4)` job in `cf-render` counter assertions; all other reported jobs pass. It looks outside this five-file diff, but the PR is not currently green.

**Verdict:** the home-specific direction and `defaultProfile?` correction look right, and the enforcement-on repro is valuable, but the generic CFC exemption is not yet aligned with the schema representation. Fix the outer-wrapper predicate before merge; tighten the rationale and functional pin while this invariant is under review.

bfollington:
Superseded by #4967, which folds in both changes from this PR (the `asCell:["stream"]` additive-required exemption in `schema-merge.ts` and `defaultProfile?` in `home.tsx`, plus `cfc-additive-default-preserves-old-doc.test.ts`) on a fresh base, and adds the missing half of the unbrick: the runnability backstop that actually rolls an already-pinned bricked root forward onto the fixed pattern.

Landing them together because neither heals an existing home alone — this branch's migration fix makes the current `home.tsx` migratable; #4967's roll-forward is what moves a pinned stale root onto it. Closing here; continue on #4967.

PR #4937, opened 2026-07-23
spike(storage): concurrent watch-refresh (default-off) — parallelize watch acquisition

bfollington:
## Why

Investigating mobile-Loom cold-load performance, we found that per-space watch
acquisition is **strict single-flight**: 0 overlapping `session.watch.add`
requests within a space across 154 round trips in two independent HAR captures.
Because traversal discovers cells incrementally, this turns into a long tail of
one-root-per-round-trip frames (roots/frame: min 1, **p50 1**), each paying a
full RTT. On estuary (~188 ms RTT floor) that dominates cold-load wall-clock —
the round-trip *count*, not payload size, is the bottleneck (responses are
~302 B median; the data itself is already loaded by a handful of root-closure
reads). This adds a default-off flag to overlap those refreshes.

## What changed

The serialization lived in **two layers**, and the one that's easy to spot from
the code isn't the binding one:

1. **`SpaceReplica` flush guard** (`runner/src/storage/v2.ts`) — one refresh
   flush per space, the next scheduled in the prior's `finally`. Now a **bounded
   in-flight window** (`CONCURRENT_WATCH_REFRESH_WINDOW = 8`) instead of a
   single-flight boolean: over-window batches stay coalesced and flush as slots
   free. With the flag off the window is 1, i.e. the original single-flight.
2. **`SpaceSession.runWatchMutation`** (`memory/v2/client.ts`) — the binding
   constraint; it chained every watch mutation's request *and* apply onto the
   previous one. Reworked into an **ordered issue phase** + **ordered apply
   phase**: the whole watch-mutation family (`watch.set` + `watch.add`) issues
   in request-issue order so wire order is preserved (`watch.add` never
   overtakes an earlier `watch.set`) and view application stays ordered, while
   the request/response wait no longer serializes. Eager requests get an
   immediate rejection handler.

Wiring is **per session, not a process global**: the runner mirrors the
`experimentalConcurrentWatchRefresh` storage setting onto each memory session
via `SpaceSession.setConcurrentWatchRefresh()` (optional-chained so test session
doubles are unaffected). Registered in `docs/development/EXPERIMENTAL_OPTIONS.md`.

Files:

- **`storage/interface.ts`** — `experimentalConcurrentWatchRefresh` storage setting.
- **`storage/v2.ts`** — bounded-window flush; single-flight when off.
- **`memory/v2/client.ts`** — ordered issue/apply split behind the per-session flag.

### Flag-off is behavior-identical

The default (flag-off) path is byte-for-byte the old behavior, including
microtask timing: `runWatchMutation` stays `async` and returns `await current`
in both branches. That boundary matters — an earlier version dropped the `await`
and a one-tick shift reordered schema interning during commit, breaking CFC
canonicalization idempotence in three suites. Preserved deliberately; don't
simplify it away.

## Status

Off by default, pending live measurement over real (estuary-latency) load; the
window size (8) is a tuning value. End state is graduation to always-on once
measured safe and beneficial, or removal if a render-side fix makes the
waterfall shallow enough that concurrency no longer pays.

## Test plan

```
deno test -A packages/runner/test/memory-v2-concurrent-watch-refresh.test.ts
deno test -A packages/memory/test/v2-concurrent-watch-refresh-test.ts
```

- **Runner** — `single-flight by default` caps concurrency at 1; with the flag
  on, concurrency reaches **exactly the window (8)** and over-window pulls stay
  queued until a slot frees. Event-driven (transport signals, no wall-clock
  sleeps).
- **Memory** (real server via loopback) — set-then-add preserves wire order
  `[watch.set, watch.add]`; overlapping `watch.add`s all resolve and later
  mutations on **every** watched root are delivered.
- **Regression** — the previously-flagged CFC idempotence suites
  (`cfc-persist-split`, `cfc-template-metadata-population`,
  `cfc-template-population`) and the related runner/memory watch suites all pass
  with the flag off. CI green, merge state clean.

---
<a href="https://app.blacksmith.sh/commontoolsinc/codesmith/labs/pr/4937"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787375879&installation_model_id=428580&pr_number=4937&repository=commontoolsinc%2Flabs&return_to=https%3A%2F%2Fgithub.com%2Fcommontoolsinc%2Flabs%2Fpull%2F4937&signature=10402cad233ae78ef433f86c1c7252b70511578b28462453779242d896dee536"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Adds an experimental, default-off concurrent watch-refresh with a bounded in-flight window (8) to parallelize per-space watch acquisition while keeping wire and apply order stable. With the flag off, behavior remains strict single-flight.

- **New Features**
  - Added `experimentalConcurrentWatchRefresh` in `IRemoteStorageProviderSettings`; runner mirrors it per session via `SpaceSession.setConcurrentWatchRefresh()` (not global).
  - Storage enforces a bounded window (`CONCURRENT_WATCH_REFRESH_WINDOW=8`): coalesces when full and flushes as slots free; single-flight is window=1.
  - Memory client routes both `watch.set` and `watch.add` through ordered issue and ordered apply phases to preserve `[set → add]` wire order and serialize view application; adds immediate rejection handlers; flag-off path unchanged.
  - Documented in `docs/development/EXPERIMENTAL_OPTIONS.md`, including a dated status line (“Status on 2026-07-24”).

- **Tests**
  - Runner tests: default caps at 1; with the flag on, concurrency reaches exactly the window and pending batches stay queued until a slot frees.
  - Memory tests: `[watch.set, watch.add]` wire order is preserved; overlapping `watch.add`s resolve and later mutations on all watched roots are delivered.

<sup>Written for commit 3ccd1b2014eaa41eda83b31a2a0be45800f17ed0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4937?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/development/EXPERIMENTAL_OPTIONS.md
+++ b/docs/development/EXPERIMENTAL_OPTIONS.md
@@ -43,6 +43,7 @@ was last checked against the code.
 | [`cfcLabelMetadataProtection`](#cfclabelmetadataprotection) | `RuntimeOptions.cfcLabelMetadataProtection` | `off` | Bernhard Seefeld (#4638) | `observe` (divergence counting) first, then `enforce` | implemented, staged rollout |
 | [`conflictAdmissionMode`](#conflictadmissionmode) | `CF_CONFLICT_ADMISSION` env, or `setConflictAdmissionMode()` | `off` | William Kelly (#4237) | keep as a tuning dial or remove after re-measurement | implemented, off by default, measured net-negative or neutral |
 | [`syncSchemaTableV2`](#syncschematablev2) | `setSyncSchemaTableConfig()` (negotiated per connection) | on | Ben Follington (#4292) | retire the negotiation once every peer speaks v2 | implemented, on by default |
+| [`experimentalConcurrentWatchRefresh`](#experimentalconcurrentwatchrefresh) | `IRemoteStorageProviderSettings` (runner mirrors it onto the session via `setConcurrentWatchRefresh()`) | off | Ben Follington (#4937) | graduate to always-on after live measurement, or remove if superseded | implemented behind the flag, off by default, not yet measured over real latency |
 | [`cfcRenderCeiling`](#cfcrenderceiling) | `commonfabric.cfcRenderCeiling()` in the browser (localStorage) | off | Bernhard Seefeld (#4550) | graduate once exchange resolution lands | implemented, off by default, dogfood only |
 | [`fuseNfsCacheTuning`](#fusenfscachetuning) | `cf fuse mount --attrcache-timeout <whole seconds; 0 = untuned>` or `--noattrcache` | cf adds `attrcache-timeout=1` (one second) to FUSE-T mounts | Ian Hickson | keep the default; shrink the exec.ts listing-recheck delay once the default has field-soaked | implemented, on by default for FUSE-T, soak-validated |
 
@@ -601,6 +602,36 @@ the per-epic implementation notes).
 >   keeps its write gate failing closed. It was added by Bernhard Seefeld in
 >   "server-side commit-time row-label re-derivation (Epic E4, Phase 3.c)"
 >   (#4552). It is permanent.
+
+### `experimentalConcurrentWatchRefresh`
+
+- **Toggle via.** `experimentalConcurrentWatchRefresh` on
+  `IRemoteStorageProviderSettings`
+  ([`packages/runner/src/storage/interface.ts`](../../packages/runner/src/storage/interface.ts)),
+  passed through `StorageManager` settings. The runner mirrors it onto each
+  memory session via `SpaceSession.setConcurrentWatchRefresh()`
+  ([`packages/memory/v2/client.ts`](../../packages/memory/v2/client.ts)) —
+  per-session, not a process global.
+- **Added by.** Ben Follington (#4937).
+- **Purpose.** By default watch acquisition is strict single-flight per space: a
+  guard holds every watch refresh after the first until the prior response
+  lands, so traversal-driven pulls discovered a tick apart serialize into
+  one-round-trip-each frames even when nothing depends on the prior response. On
+  a high-RTT link this dominates cold-load wall-clock. With the flag on,
+  refreshes overlap up to a bounded window (`CONCURRENT_WATCH_REFRESH_WINDOW`,
+  currently 8) in `storage/v2.ts`, and the memory client issues the whole
+  watch-mutation family (`watch.set` + `watch.add`) in an ordered issue phase so
+  wire order is preserved and application stays ordered. Same-tick microtask
+  coalescing is unchanged.
+- **Current default and planned end state.** Off by default. It is a spike
+  pending live measurement on a real (estuary-latency) load; the window size is
+  a tuning value. End state is either graduation to always-on with a settled
+  window, or removal if the render-side fix (initial-render descent) makes the
+  waterfall shallow enough that concurrency no longer pays.
+- **Status.** Implemented behind the flag, off by default; not yet measured
+  end-to-end over real latency.
+- **Path to removal.** Graduate to always-on once measured safe and beneficial,
+  or remove if superseded by reducing the round-trip count at the source.
 
 ---
 
--- a/docs/development/EXPERIMENTAL_OPTIONS.md
+++ b/docs/development/EXPERIMENTAL_OPTIONS.md
@@ -628,8 +628,8 @@ the per-epic implementation notes).
   a tuning value. End state is either graduation to always-on with a settled
   window, or removal if the render-side fix (initial-render descent) makes the
   waterfall shallow enough that concurrency no longer pays.
-- **Status.** Implemented behind the flag, off by default; not yet measured
-  end-to-end over real latency.
+- **Status on 2026-07-24.** Implemented behind the flag, off by default; not yet
+  measured end-to-end over real latency.
 - **Path to removal.** Graduate to always-on once measured safe and beneficial,
   or remove if superseded by reducing the round-trip count at the source.
 
--- a/packages/memory/test/v2-concurrent-watch-refresh-test.ts
+++ b/packages/memory/test/v2-concurrent-watch-refresh-test.ts
@@ -1,0 +1,146 @@
+/**
+ * Memory-client guarantees for `concurrentWatchRefresh` (set by the runner from
+ * the `experimentalConcurrentWatchRefresh` storage setting):
+ *
+ *  1. Wire order across the whole watch-mutation family is preserved — an eager
+ *     `watch.add` never overtakes an earlier `watch.set` (the bug the ordered
+ *     issue phase fixes).
+ *  2. Overlapping watch.adds through the REAL server all resolve, and later
+ *     mutations on every watched root are delivered.
+ *
+ * Event-driven (no wall-clock sleeps): a deterministic microtask drain settles
+ * the loopback pipeline.
+ */
+import { assertEquals } from "@std/assert";
+import { connect, loopback, type Transport } from "../v2/client.ts";
+import { Server } from "../v2/server.ts";
+import type { EntitySnapshot, WatchSpec } from "../v2.ts";
+import {
+  testSessionOpenAuthFactory,
+  testSessionOpenServerOptions,
+} from "./v2-auth-test-helpers.ts";
+
+async function drainMicrotasks(turns = 50): Promise<void> {
+  for (let i = 0; i < turns; i++) await Promise.resolve();
+}
+
+/** loopback transport that records the type of each client->server message. */
+function recordingLoopback(server: Server, wire: string[]): Transport {
+  const inner = loopback(server);
+  return {
+    send(payload: string) {
+      if (payload.includes('"session.watch.set"')) wire.push("watch.set");
+      else if (payload.includes('"session.watch.add"')) wire.push("watch.add");
+      return inner.send(payload);
+    },
+    close: () => inner.close(),
+    setReceiver: (r) => inner.setReceiver(r),
+    setCloseReceiver: (r) => inner.setCloseReceiver?.(r),
+  };
+}
+
+const rootSpec = (id: string): WatchSpec => ({
+  id,
+  kind: "graph",
+  query: { roots: [{ id, selector: { path: [], schema: false } }] },
+});
+
+Deno.test("concurrent refresh preserves wire order: watch.add never overtakes watch.set", async () => {
+  const server = new Server({
+    ...testSessionOpenServerOptions,
+    store: new URL("memory://concurrent-refresh-order"),
+    subscriptionRefreshDelayMs: 0,
+  });
+  const wire: string[] = [];
+  const client = await connect({ transport: recordingLoopback(server, wire) });
+  const space = "did:key:z6Mk-concurrent-refresh-order";
+  const session = await client.mount(space, {}, testSessionOpenAuthFactory);
+  session.setConcurrentWatchRefresh(true);
+  try {
+    // set then add issued in the SAME turn. Under the old eager-add spike the
+    // add reached the wire first; the ordered issue phase must keep [set, add].
+    const set = session.watchSetSync([]);
+    const add = session.watchAddSync([]);
+    await Promise.all([set, add]);
+    assertEquals(wire, ["watch.set", "watch.add"], "wire order preserved");
+  } finally {
+    await client.close();
+    await server.close();
+  }
+});
+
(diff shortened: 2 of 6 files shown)


bfollington (commented):

Scope: four-file concurrent watch-refresh spike · coverage: deep-read of the changed paths plus the memory client/server ordering and selector-tracker invariants · checks: targeted experiment **passed** (2/2), the five named regression files **passed** (60 steps), `deno check` **passed**, `deno lint` **passed**, `deno fmt --check` **failed**.

The measured mechanism is credible, and request-issue order is the correct application order **under the current protocol**: `WebSocketTransport.send()` emits ordered WebSocket frames, while `Connection.receive()` serializes the server side through `#receiving` before `watchAdd()` mutates `session.graphs`, `session.watches`, and `lastSyncedSeq`. The selector tracker is not the ordering guarantee—disjoint selectors for the same document can both proceed—but the second server mutation builds on the first, and the client applies them in that same order. Applying by response arrival or `serverSeq` would add no correctness here. This dependency should be made explicit and tested, because removing either transport ordering or the server receive queue would invalidate the design.

## 🔴 Blocking before merge

1. **[major] `packages/runner/src/storage/v2.ts:2249-2293` — concurrent mode removes all backpressure.** With the flag on, every batch discovered in a later microtask starts immediately; `#watchRefreshFlushing` becomes inert. The motivating trace has 154 round trips, so this can turn into 154 outstanding client requests and 153 queued server receives, with no bound other than traversal size. Same-tick coalescing does not constrain incrementally discovered waves. Replace the boolean with an in-flight count/window (and drain the coalesced pending batch when a slot opens), then test both overlap and the maximum. **Verified from the scheduling path.**

2. **[major] `packages/memory/v2/client.ts:36-44,666-685` / `packages/runner/src/storage/interface.ts:321-331` — the advertised setting is not an end-to-end setting.** `experimentalConcurrentWatchRefresh: true` only lifts the runner guard; overlap still requires the unrelated exported module-global setter. Conversely, setting that global affects every client/storage manager in the process and the new test leaves it enabled. Since `watchAddSync` has only this production caller, pass the behavior per call (or per session) from `SpaceReplica`, remove the ambient mutable global, and register the flag in `docs/development/EXPERIMENTAL_OPTIONS.md` as required by the central registry. **Verified by call-site search and the test’s final global state.**

3. **[major] `packages/runner/test/memory-v2-concurrent-watch-refresh.test.ts:154-203` — the test uses wall-clock sleeps to infer progress and bypasses the protocol invariant it needs to guard.** The 25/60/10/20 ms waits can fail a healthy run under load, and `HoldingTransport` never exercises the real connection receive queue, persistent server watch set, or later subscription delivery. Drive it with transport signals and deterministic microtask barriers; release requests individually rather than polling/releasing on a clock. Add a loopback/server test that issues multiple additions, then mutates every watched root and proves all later updates arrive in order. **Verified; the test passes locally, but violates `docs/development/waiting-in-tests.md`.**

4. **[minor] The new test file is not formatted, so the aggregate Check job is red.** `deno fmt --check` reports four formatting regions in `memory-v2-concurrent-watch-refresh.test.ts`; local reproduction matches CI. **Verified locally and from CI logs.**

## 🟡 Improvement

- The concurrency test currently accepts `maxConcurrent >= 2`, although the experiment and description claim four-way overlap. Once a bounded production window is chosen, assert the exact intended maximum and add a reversed-response synthetic case to pin the client mutation-chain behavior independently of the real transport’s ordered responses.

**Verdict:** useful spike with a sound ordering theory, but not merge-ready until concurrency is bounded, the flag is scoped/wired coherently, and the timing-based synthetic test is replaced or supplemented with an event-driven protocol-level regression.

bfollington (commented):
_cf-review follow-up via Codex (GPT-5), on behalf of @bfollington_

One additional blocker surfaced after exercising the PR's actual mutation queue. This supplements the earlier cf-review rather than repeating it.

## 🔴 Blocking

**[major] `packages/memory/v2/client.ts:683-685` — the eager add can overtake an already-queued `watch.set`, so server and client watch state diverge.**

`runWatchMutation()` starts its `work` from a promise continuation. With concurrent refresh enabled, `watchAddSync()` instead calls `sendRequest()` immediately, before entering that continuation. I reproduced this exact sequence against the changed client:

```ts
const set = session.watchSetSync([]);
const add = session.watchAddSync([]);
await Promise.all([set, add]);
```

The transport receives:

```text
session.watch.add
session.watch.set
```

The server serializes `Connection.receive()` in that wire order, so its final watch set is the `set` payload. The client still runs the mutation callbacks in call order, applies `set` and then `add`, and records `#watchSpecs` as set-plus-add. That directly contradicts the comment that request-issue order and ordering guarantees are preserved.

This is not only a public-API corner: during a non-resumed reconnect, `restore()` calls `watchSetSync(this.#watchSpecs)` after `hello()` has already marked the client connected, so a concurrent pull can issue an eager `watch.add` across the restore-time set.

Please give every watch mutation an issue-order reservation. For example, split the queue into an ordered request-issue phase and an ordered apply phase, allowing adjacent adds to pipeline only after all earlier mutations (especially `watch.set`) have been issued. Add a regression that calls set then add in one turn and asserts both wire order and the final server/client watch set.

The same split should attach a rejection handler to every eager request immediately. Today a later eager request can reject while an earlier mutation is still pending, leaving the eager promise temporarily unhandled even though the returned mutation-chain promise has a catch.

**Verified** with a focused test against commit `0d79be1bb537d6a18b552d5440295f1084c26e07`; it fails with actual wire order `[watch.add, watch.set]`.

**Verdict:** the pipelining mechanism is promising, but the current queue split changes watch semantics and is not safe to enable until request ordering is preserved across the whole watch-mutation family.

bfollington (commented):
_cf-review re-review via Codex (GPT-5), on behalf of @bfollington_

Scope: latest head `d426092c` · coverage: deep-read of the issue/apply queues, runner window/drain path, flag wiring, docs, and both new test files · checks: 4 new tests passed; 5 named regression files passed (60 steps); fmt/lint/check passed; three additional runner regression files failed exactly as CI reports.

The prior findings are substantively addressed: the window is bounded at 8, the flag is per session and documented, set/add wire order is preserved, eager rejections are handled, and the new tests no longer use wall-clock sleeps.

## 🔴 Blocking

**[major] `packages/memory/v2/client.ts:936-977` — the flag-off path lost the old async return boundary, so it is not behavior-identical and currently breaks three runner suites.**

The old `runWatchMutation()` was `async` and returned `await current`; the replacement is synchronous and returns `current` directly in both branches. Even with `#concurrentWatchRefresh === false`, that changes when the caller resumes relative to storage synchronization. The observable result is that CFC envelopes are re-materialized with a different property insertion order after a supposedly idempotent recomputation.

Verified in both directions:

- Previous PR head `0d79be1b`: all three pass.
- Latest head `d426092c`: all three fail locally, matching CI's failed runner shards 1–3:
  - `cfc-persist-split.test.ts` — “keeps re-derivation idempotent per class”
  - `cfc-template-metadata-population.test.ts` — “recompute with metadata templates present is a no-op”
  - `cfc-template-population.test.ts` — “recompute with templates present is a no-op”
- Minimal patch: make `runWatchMutation` `async` again and use `return await current` in both branches. With only that change, all 58 steps in those three files pass, and all four new concurrency tests still pass.

Please preserve that original async boundary while keeping the new issue/apply split. This PR's stated contract is no flag-off behavior change, and the current head demonstrably violates it.

## 🟡 Improvements

- `packages/memory/test/v2-concurrent-watch-refresh-test.ts:8-9,117-136` says later mutations on **every** watched root are delivered, but the test updates only `ids[0]`. Mutate/assert every id so a server watch set that accidentally retains only one acquisition cannot pass.
- `packages/runner/test/memory-v2-concurrent-watch-refresh.test.ts:167-179` replaces wall-clock polling with a capped 100-iteration retry loop. The transport already owns the relevant events; expose/release individual held requests and await `receivedAtLeast()` checkpoints so completion is signal-driven rather than dependent on an iteration budget.

**Verdict:** nearly ready; the concurrency design now looks coherent, but the default-off async-boundary regression is blocking and already red in CI.

bfollington:
Thanks for the thorough re-review — all points addressed. Also rebased the branch onto main (it was based on the stale loom-stable-2026-07-20 tag; the rebase was clean — none of main's commits to these files touched the watch-refresh path). Latest head e031cbe5f, CI green.

Blocking (re-review): flag-off async boundary
Fixed exactly as suggested — runWatchMutation is async again and returns await current in both branches, so the flag-off path's microtask timing is identical to the pre-concurrency code while keeping the new issue/apply split. Confirmed all three named suites pass on the fixed head (58 steps, 0 failed): cfc-persist-split, cfc-template-metadata-population, cfc-template-population. The four new concurrency tests still pass.

Nice catch — and agreed on the root cause: the missing await tick shifted when schemas intern during the commit, and that canonicalization's property order is insertion-order-dependent, so an idempotent recompute serialized keys differently. The flag-off path is now timing-neutral.

Improvements
Loopback delivery (v2-concurrent-watch-refresh-test.ts): now mutates and asserts every watched root — consumes the subscription until all five reach n=1, so a watch set that retained only one acquisition fails.
Runner drain (memory-v2-concurrent-watch-refresh.test.ts): replaced the capped retry loop with a signal-driven drain — after each release it races "all pulls settled" vs receivedAtLeast(count+1) (next batch sent), no iteration/time budget.
Prior blocking items (from the first two reviews), for completeness
Backpressure: single-flight boolean replaced by a bounded in-flight window (CONCURRENT_WATCH_REFRESH_WINDOW = 8); over-window batches stay coalesced and flush as slots free. Test asserts concurrency caps at exactly the window (and single-flight at 1).
End-to-end flag: the process-global is gone; the memory session gets a per-session setConcurrentWatchRefresh(), mirrored from the storage setting by the runner (optional-chained so test session doubles are unaffected). Registered in docs/development/EXPERIMENTAL_OPTIONS.md.
Ordering across the family: both watchSetSync and watchAddSync route through the ordered issue/apply phases; set-then-add preserves wire order [watch.set, watch.add] (verified against the real server via loopback), and every eager request gets an immediate rejection handler.
Event-driven tests: no wall-clock sleeps; transport signals + deterministic microtask barriers.
Ready for another look when you have a moment.

bfollington (commented):
_cf-review re-review via Codex (GPT-5), on behalf of @bfollington_

Scope: latest head `e031cbe5` · coverage: deep-read of the six-file PR diff against its rebased base, including the memory issue/apply queues, server receive-order invariant, runner window and drain path, per-session flag wiring, docs, and both new test files.

## No blocking findings

The prior blocker is fixed correctly. `runWatchMutation()` is async again and uses `return await current` in both branches, restoring the original flag-off continuation boundary while retaining the concurrent issue/apply split. The bounded window, per-session flag, whole-family set/add issue ordering, immediate eager-rejection handling, and event-driven test drains all remain coherent.

Verification on this exact head:

- New concurrency suites: **4 passed, 0 failed**.
- Previously regressing CFC suites: **11 groups / 58 steps passed, 0 failed**.
- Five related runner regression files: **21 groups / 60 steps passed, 0 failed**.
- `deno fmt --check`, `deno lint`, and `deno check` on the changed files: **passed**.
- GitHub CI: all required checks are green; merge state is clean.

## Non-blocking housekeeping

The hand-written PR description is stale: it still says the memory toggle is a module-global measurement scaffold, concurrency is fully unbounded, the old flush boolean remains dangling, and the test plan covers only the original runner test. The implementation now has per-session wiring, an 8-request window, an in-flight counter, memory protocol coverage, and the restored flag-off boundary. Please refresh the description before taking the PR out of draft so reviewers do not evaluate the superseded design.

**Verdict:** clean from a code-review perspective. The PR remains explicitly draft/default-off pending the live-latency measurement already called out in the description and experimental-options registry.

PR #4950, opened 2026-07-23
feat(patterns): add privacy-safe Topics workload diagnostics

willkelly:
## Summary

Add deterministic, content-free diagnostics for stressing the Topics pattern across isolated runtimes. The new `topics-diagnose` command measures event execution, scheduler settling, storage transactions, conflicts, write amplification, graph shape, and convergence without exporting document values, raw links, entity identifiers, or caller-selected bootstrap identities.

## Why

Topics has enough nested cells, cross-references, concurrent users, and replicated sessions to expose scheduler and memory behavior that unit-level timings cannot explain. Before this change, a workload could appear slow or flaky without a reliable way to separate useful commits from retries, no-op amplification, storage conflicts, delayed frames, or unsettled graph work. Polling for visible state also made multi-runtime tests timing-sensitive and could declare success before queued nested handler commits had drained.

This PR creates an event-driven measurement path so performance investigations can compare deterministic counters first and treat wall-clock timing as secondary, noisy evidence.

### Structured runtime telemetry

- Add a typed runtime telemetry sink and extend the OpenTelemetry bridge with scheduler-run, event-commit, retry, storage, and settling measurements.
- Keep detailed write samples lease-controlled so normal runtimes do not pay for no-op classification unless a diagnostics consumer explicitly retains it.
- Preserve telemetry bridge leases in runtime clients and background workers for the full runtime lifetime.

### Deterministic synchronization

- Add scheduler diagnostic barriers that wait for queued work, in-flight event commits, delayed frames, storage acknowledgement, server refresh, and nested-cell topology synchronization.
- Correlate queued events with stable event IDs and commit outcomes.
- Synchronize every nested result-cell hop before dispatching an event on a cold replica.
- Preserve Fabric sigils in raw reads by using `pull()` as the synchronization fence and `getRaw()` as the value source.

### Canonical memory diagnostics

- Report aggregate transaction counts, accepted/rejected outcomes, conflicts, replays, byte and operation totals, patch-operation types, redacted path shapes, and scheduler state.
- Add an event-driven diagnostics flush fence for queued receive and watch-refresh work.
- Keep the aggregate snapshot path separate from exported spans.

### Hardened multi-runtime harness

- Expose only an aggregate diagnostics RPC allowlist; generic read, raw-read, link, logger, and diagnostics commands are rejected in aggregate mode.
- Keep worker-local immutable bootstrap creator state so callers cannot select a BroadcastChannel piece identity.
- Aggregate content-free telemetry across runtime workers and validate final convergence.
- Add regressions for cold nested dispatch, delayed commit draining, and raw Topic-link preservation through root reordering.

### Topics workload CLI

- Add configurable matrix runs plus `--quick` and deterministic `--profile=conflicts` presets.
- Exercise topic setup, comment and cross-reference workloads, typing, and bounded root oscillation.
- Emit one JSON report with per-phase operation counts, runtime and memory aggregates, write-amplification ratios, graph/settling state, oscillation metadata, and convergence.
- Document invocation, interpretation, and the privacy contract in `packages/patterns/topics/README.md`.

## Privacy and safety contract

Diagnostic reports contain bounded counts, timings, fixed error codes, patch kinds, and redacted path-shape buckets only. They do not contain document values, comments, raw links, entity IDs, DIDs, session IDs, user IDs, event IDs, transaction IDs, stack traces, or arbitrary error text. Aggregate workers reject general-purpose inspection RPCs rather than relying only on report redaction.

## Validation

Run with the repository-pinned Deno 2.8.1:

- `deno task check`
- `deno fmt --check`
- 8 runner test suites / 67 steps across telemetry, event identity, scheduler events, barriers, and nested piece startup
- 87 memory tests across scheduler state and server transaction diagnostics
- 13 Topics diagnostics tests / 3 integration steps
- `deno run --frozen -A packages/patterns/tools/topics-diagnose.ts --profile=conflicts`
  - 16 submitted / 8 accepted / 8 rejected
  - two-step repeat ratio 1
  - converged
- `deno run --frozen -A packages/patterns/tools/topics-diagnose.ts --profile=conflicts --rounds=1`
  - 4 submitted / 2 accepted / 2 rejected
  - two-step repeat ratio null
  - converged
- LSP diagnostics: zero errors in every modified TypeScript file
- `git diff --check`

## Investigation scope

The workload is suitable for matched steady-state contention comparisons. It does not perform a pattern-version upgrade, so it can detect deterministic steady-state regressions around an upgrade-related change but cannot by itself prove or disprove claims specifically about old/new generated-cell overlap during an upgrade. The matched comparison against #4916 is reported separately with that limitation made explicit.

## Performance baseline

This PR intentionally adds two multi-runtime integration suites: the nested-send suite adds about 135 seconds to shard 2, and the aggregate-only diagnostics RPC test adds about 67 seconds to shard 1. These narrow baselines use clean attempt 3 values; broader attempt 2 runner noise is not accepted.

NEW_PERF_BASELINE: job: Pattern Integration Tests (2/4) = 255s
NEW_PERF_BASELINE: step: patterns integration = 224s
NEW_PERF_BASELINE: job: Pattern Integration Tests = 255s
NEW_PERF_BASELINE: job: Pattern Integration Tests (1/4) = 217s

---
<a href="https://app.blacksmith.sh/commontoolsinc/codesmith/labs/pr/4950"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787421993&installation_model_id=428580&pr_number=4950&repository=commontoolsinc%2Flabs&return_to=https%3A%2F%2Fgithub.com%2Fcommontoolsinc%2Flabs%2Fpull%2F4950&signature=ab62be02baf5abc7ea68f3e2128fafa29c4df2f293222baf035991013587ed87"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Adds a privacy-safe Topics workload diagnostics CLI and structured runtime/memory telemetry to measure scheduler, storage, and convergence deterministically across isolated runtimes. Hosts now receive aggregate-only telemetry; event IDs and write paths stay inside the worker.

- **New Features**
  - `topics-diagnose` CLI: matrix runs, `--quick`, `--profile=conflicts`, real nested streams, one JSON report.
  - Structured runtime telemetry: commit amplification (reads/writes vs changed writes), retries, drop reasons; opt-in detailed samples via lease; upgraded `@opentelemetry` bridge with separate `attributes`, `spanAttributes`, and metric attributes; host markers (`HostRuntimeTelemetryMarker`) emit metrics only and include preflight stats.
  - Deterministic sync: scheduler/storage barriers for queued work, in‑flight commits, delayed frames, server refresh, and nested‑cell topology; cold replica sync; flush fences for receive/watch refresh.
  - Memory server diagnostics: aggregate, content‑free commit telemetry (transactions, outcomes, conflicts, replays, patch kinds, redacted path shapes); `commitTelemetry` and `aggregateOnlyDiagnostics` options; hardened multi‑runtime harness with an RPC allowlist and convergence checks.
  - Docs: usage and the privacy contract in `packages/patterns/topics/README.md`.

- **Bug Fixes**
  - Enforce host‑only aggregate telemetry: define host‑safe markers; redact worker details in `@commonfabric/runtime-client` before postMessage; `@commonfabric/lib-shell` and the shell consume host markers and keep graph detail off host telemetry; OTel bridge preserves metrics for host‑safe markers.
  - Attach the OTel bridge in the background piece service only after initialization; fail‑open returns `null`; guarantee bridge cleanup on shutdown.
  - Preserve malformed transaction errors in memory v2 diagnostics and distinguish first‑apply from replay.
  - Validate Topics diagnostics CLI arguments and materialized crossrefs; move workload tests to integration.
  - Synchronize nested piece result‑cell chains before dispatch and avoid redundant re‑syncs.
  - Tests: isolate inherited‑property pollution coverage for schema‑ref traversal; consolidate Topics diagnostics scenarios for faster, clearer runs.

<sup>Written for commit b9ee8cba686c5211f5c0f78df074c8cc4a14538d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4950?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/background-piece-service/src/worker.ts
+++ b/packages/background-piece-service/src/worker.ts
@@ -143,6 +143,7 @@ export function setWorkerStateForTesting(
       [string, Cell<{ bgUpdater: Stream<unknown> }>]
     >;
     streamValidator?: typeof isStream;
+    detachOtelBridge?: (() => void) | null;
   },
 ): void {
   if ("initialized" in state) initialized = state.initialized ?? false;
@@ -159,6 +160,9 @@ export function setWorkerStateForTesting(
   }
   if ("streamValidator" in state) {
     streamValidator = state.streamValidator ?? isStream;
+  }
+  if ("detachOtelBridge" in state) {
+    detachOtelBridge = state.detachOtelBridge ?? null;
   }
 }
 
@@ -244,30 +248,34 @@ export async function cleanup(): Promise<void> {
   currentSession = null;
   manager = null;
 
-  // Ensure storage is synced before cleanup
-  if (runtime) {
-    await runtime.storageManager.synced();
-    await runtime.dispose();
-    runtime = null;
-  }
-
-  // Detach the OTel bridge only after the runtime is fully torn down, so the
-  // final sync/dispose telemetry (storage completions, subscription removals)
-  // is still observed; detaching closes any spans left in flight.
-  if (detachOtelBridge) {
-    detachOtelBridge();
-    detachOtelBridge = null;
-  }
-
-  // Flush buffered spans/metrics before the controller terminates this worker;
-  // fail-open — telemetry teardown must never block cleanup.
   try {
-    await shutdownOpenTelemetry();
-  } catch (error) {
-    console.error("Failed to shut down OpenTelemetry:", error);
-  }
+    if (runtime) {
+      try {
+        await runtime.storageManager.synced();
+      } finally {
+        await runtime.dispose();
+      }
+    }
+  } finally {
+    runtime = null;
 
-  initialized = false;
+    // Detach only after runtime teardown so final sync/dispose telemetry is
+    // observed. The finally block also closes spans when teardown rejects.
+    if (detachOtelBridge) {
+      detachOtelBridge();
+      detachOtelBridge = null;
+    }
+
+    // Flush before the controller terminates this worker. Telemetry teardown
+    // remains fail-open and must not replace a runtime cleanup failure.
+    try {
+      await shutdownOpenTelemetry();
+    } catch (error) {
+      console.error("Failed to shut down OpenTelemetry:", error);
+    }
+
+    initialized = false;
+  }
 }
 
 export async function runPiece(data: RunData): Promise<void> {
--- a/packages/background-piece-service/src/worker.ts
+++ b/packages/background-piece-service/src/worker.ts
@@ -183,30 +183,34 @@ export async function initialize(
     errorHandlers: [errorHandler],
   }));
   // Each worker is its own isolate: the provider main.ts registers doesn't
-  // exist here, so initialize OTel in-worker (idempotent, fail-open) or the
-  // bridge below would attach to no-op instruments.
+  // exist here, so initialize OTel in-worker when export is enabled. Setup is
+  // idempotent and fail-open; a failed setup leaves the API instruments no-op.
   await initOpenTelemetry();
 
-  // Bridge the runtime's existing telemetry stream to OpenTelemetry. This is a
-  // second, passive consumer of the same event bus the debug tooling uses; it
-  // emits no-op instruments unless a provider is registered (see otel.ts).
-  detachOtelBridge = attachRuntimeTelemetryOtelBridge(runtime.telemetry, {
-    tracer: getTracer(),
-    meter: getMeter(),
-    attributes: {
-      "ct.runtime": "bg-piece",
-      "space.did": spaceId,
-      "user.did": identity.did(),
-    },
-    // Metric datapoints don't inherit resource attributes in SigNoz, so stamp
-    // the scoping labels explicitly (metrics only — on spans these live on the
-    // resource, and duplicating them as span attributes makes the bare key
-    // ambiguous in queries).
-    metricAttributes: {
-      "service.name": env.OTEL_SERVICE_NAME,
-      "deployment.environment": env.ENV,
-    },
-  });
+  // Do not attach a no-op bridge when OTel is disabled: attaching retains the
+  // runtime's detailed event-commit telemetry until cleanup. Enabled workers
+  // retain the bridge and detach it during cleanup after runtime disposal.
+  if (env.OTEL_ENABLED) {
+    detachOtelBridge = attachRuntimeTelemetryOtelBridge(runtime.telemetry, {
+      tracer: getTracer(),
+      meter: getMeter(),
+      attributes: {
+        "ct.runtime": "bg-piece",
+      },
+      spanAttributes: {
+        "space.did": spaceId,
+        "user.did": identity.did(),
+      },
+      // Metric datapoints don't inherit resource attributes in SigNoz, so stamp
+      // the scoping labels explicitly (metrics only — on spans these live on the
+      // resource, and duplicating them as span attributes makes the bare key
+      // ambiguous in queries).
+      metricAttributes: {
+        "service.name": env.OTEL_SERVICE_NAME,
+        "deployment.environment": env.ENV,
+      },
+    });
+  }
 
   manager = new PieceManager(currentSession, runtime);
   await manager.ready;
--- a/packages/background-piece-service/src/worker.ts
+++ b/packages/background-piece-service/src/worker.ts
@@ -17,6 +17,7 @@ import { env } from "./env.ts";
 import {
   getMeter,
   getTracer,
+  getTracerProvider,
   initOpenTelemetry,
   shutdownOpenTelemetry,
 } from "./otel.ts";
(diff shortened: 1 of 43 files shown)


willkelly:
## Investigation retrospective: applying these diagnostics to #4916

_Agent-authored report._

I used this PR as a matched overlay on the exact base and head commits of #4916, running the deterministic conflict profile three times per side in alternating order.

### What worked well

- **Behavior was reproducible.** Every run produced exactly 16 submitted / 8 accepted / 8 rejected operations, a two-step repeat ratio of 1, and final convergence. That made it possible to establish that #4916 does not cause a deterministic regression in the exercised steady-state path.
- **Counters prevented a timing-only conclusion.** Workload-phase timing was effectively flat (head median -0.55%) and total timing moved slightly against the head (+2.20%), while memory conflict median moved favorably (-19.35%). Seeing all three signals made the overlap and run-to-run noise obvious instead of encouraging a claim from one elapsed-time number.
- **The barriers held.** Delayed frames, queued nested events, commit acknowledgements, refresh work, and graph settling drained without polling or sleeps; every report reached the same converged state.
- **The report remained content-free.** The comparison required only aggregate operation, transaction, graph, settle, and oscillation fields. No document values, raw links, identities, or arbitrary error strings were needed to investigate the claim.
- **The profile was cheap enough to repeat.** Six complete isolated-runtime runs were practical, and the fixed profile made command and topology parity straightforward.

### What it could not establish

#4916 changes generated-cell identity behavior across pattern versions. The current Topics profile begins after startup and performs steady-state writes; it never upgrades a running pattern from version N to N+1. It therefore cannot measure the claimed reduction in old/new generated-cell overlap directly.

The observed conflict ranges overlapped:

- base: 30–33, median 31
- head: 24–36, median 25

The head had a lower median but also the worst outlier. That is useful diagnostic output, but not statistically or causally strong enough to validate the upgrade-specific claim.

### Assessment

The PR worked well as a **regression detector and evidence-quality guardrail**: it proved deterministic workload parity, ruled out an obvious steady-state regression, exposed timing noise, and stopped the investigation from overstating mixed results. It did not turn a non-upgrade workload into evidence about upgrade behavior, which is the correct boundary.

### Next diagnostic increment

Add an upgrade scenario that keeps multiple sessions active while moving a piece from version N to N+1, then snapshots:

1. generated-cell graph overlap by version-safe aggregate categories,
2. dirty and pending scheduler state through the transition,
3. accepted/rejected transactions and conflicts in the upgrade window,
4. time and barrier iterations to post-upgrade convergence.

The detailed evidence and conclusion are posted on #4916: https://github.com/commontoolsinc/labs/pull/4916#issuecomment-5061756215.

willkelly:
Closing — the investigation this was built to enable concluded by another route, and main institutionalized that route.

`packages/state-inspector/churn.ts` names this PR by number as the alternative it replaced, and records the answer it was built to find: the July 2026 Topics write storm, 20 compiler-generated cells producing 96% of commits, peaking ~1,598 commits/minute. The post-hoc tool (`cf inspect churn`, #5075) reads the durable store after the fact — any space, any window, no instrumentation needed at incident time — which is strictly better for this purpose than instrumenting an orchestrated run. The underlying defect is fixed and verified against deployed `b2213af8b` (#4659 + #4956).

**Salvage — one real bug, independent of this PR's fate:**

Main's marker union declares `changedWriteCount` as a field distinct from `writeCount`, but `packages/runner/src/scheduler/events.ts:1148-1149` fills both with `log.writes.length`. They are identical. **No-op write amplification is not actually measured today** — the slot exists and reports a duplicate into it, which is worse than not having it, because it looks populated. This branch's `classifyTelemetryWriteCounts` / `eventCommitTelemetryWriteCounts` in `scheduler/reactivity.ts` (~90 lines) is the fix.

Also unbuilt on main, if ever wanted: the `topics-diagnose` CLI (1224 lines), canonical memory commit telemetry in `memory/v2/server.ts`, and the host-telemetry privacy narrowing (`HostRuntimeTelemetryMarker`, `telemetryMarkerForHost()`, `scheduler.event.drop`) — note `runtime-processor.ts:1519-1525` still forwards the raw marker, including handlerId, write paths and error strings, to the host verbatim.

PR #4954, opened 2026-07-23
perf(schema-generator): deep-freeze each schema as it is formatted

danfuzz:
Pre-PR for the `dedupeByValueEqual` content-hash work on #4942. Deep-freezes
generated schemas early so the value model's frozen-object hash cache can reuse
subtree hashes during schema deduplication. Behavior-neutral; lands before
#4942.

## Why

`hashStringOf`'s object-hash cache is a `WeakMap` keyed on deep-frozen identity
— a frozen subtree is hashed once and reused by identity thereafter. Schema
generation currently freezes nothing, so its schemas are mutable, and every
schema hashed during dedup (and its nested content, when an enclosing schema is
hashed at an outer step) is re-hashed from scratch. Freezing subtrees as they
are produced turns that repeat hashing into cache hits.

The perf payoff is realized once #4942's hash-based `dedupeByValueEqual` lands
on top; this PR is the foundation, and freezing generated schemas is sound
hygiene on its own.

## What

- Each schema is deep-frozen the moment its formatter produces it, at the one
  choke point every (sub)schema flows through (`SchemaGenerator.formatType`,
  the recursive dispatch). Bottom-up generation means a nested schema is frozen
  before any enclosing formatter composes or deduplicates it. `deepFreeze`
  short-circuits already-frozen input, so the sweep stays cheap.

- Freezing forces the generator to stop mutating already-produced schemas. A
  handful of sites attached JSDoc metadata (`description`, `#tags`, `$comment`)
  or a `Default<>` value to a formatted **child** in place; each now attaches to
  a mutable shallow copy — same resulting object, different identity. The
  `normalizeSchemaForComparison` / `mergeSchemaGroup` builders that assemble a
  fresh local object were not touched.

## Behavior-neutral

Every schema-generator and ts-transformers golden is **byte-identical**,
verified by regenerating both with `UPDATE_GOLDENS=1` — zero churn.

## Performance — measured, and the payoff does not materialize

A direct A/B (best-of-3, same harness/machine, one moderately complex type,
2000 warmed `generateSchema` calls):

| state | µs/gen | vs main |
| --- | ---: | ---: |
| main (no freeze, `JSON.stringify` dedup) | 163.5 | — |
| freeze alone (this PR) | 173.4 | +6% |
| hash-dedup alone (#4942) | 177.4 | +8.5% |
| freeze + hash-dedup (stacked) | 186.0 | **+14%** |

The stacked case is **slower than either piece alone** — the freeze cost and the
hash-dedup cost simply add. If the frozen-subtree hash cache were paying off, the
stack would be *faster* than hash-dedup-alone; it is the opposite.

Why the cache does not help here: it is keyed on object *identity*, and
single-pass schema generation does not re-hash the same object. A generation-time
schema is hashed at most once (by the dedup), then walked once by `createSchemaAst`
to emit as an AST literal, then discarded — there is no in-process schema-reuse for
the cache to hit. (`packages/schema-generator/src` does no hashing of its own; the
runtime re-parses the emitted source into fresh objects, never the frozen ones.)
And `hashStringOf` (a content hash) is inherently pricier per call than a native
`JSON.stringify` for small schemas.

In absolute terms these are ~10–22µs deltas on `generateSchema`, itself well under
1% of real compile time, below the CI perf-check's wall-time noise floor. But the
measured direction is a small net **cost** with no realized benefit in this path,
so this PR's premise (freeze to speed the dedup) does not hold. See the discussion
thread for disposition.

## Testing

- schema-generator `47 passed`, ts-transformers `1105 passed`,
  `./tasks/check.sh` clean.
- New `frozen-output.test.ts` asserts a generated schema, a nested default, and
  a union schema are each `isDeepFrozen`.

---
<a href="https://app.blacksmith.sh/commontoolsinc/codesmith/labs/pr/4954"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787426562&installation_model_id=428580&pr_number=4954&repository=commontoolsinc%2Flabs&return_to=https%3A%2F%2Fgithub.com%2Fcommontoolsinc%2Flabs%2Fpull%2F4954&signature=f775386b46bb98f433f59850e4b756efdb3ef1bfa4bef78aaa1c203ec25e6506"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Deep-freeze each generated schema at format time to enable hash-cache reuse during deduplication and prep for hash-based `dedupeByValueEqual` (#4942). Output is unchanged; added tests assert schemas are deep-frozen.

- **Refactors**
  - Freeze every (sub)schema in `SchemaGenerator.formatType` via `deepFreeze` from `@commonfabric/data-model/deep-freeze`.
  - Stop mutating formatted children; clone before adding `description`, doc tags, `$comment`, or `default`.
  - Updated `common-fabric-formatter`, `intersection-formatter`, and `object-formatter`.
  - New `frozen-output.test.ts` verifies root, nested default, and union schemas are deep-frozen.

- **Performance**
  - Enables hash cache hits (keyed by deep-frozen identity) during schema dedup; forms the base for #4942’s content-hash dedupe.

<sup>Written for commit e86f62dbc85d80d091379b48237071a8783113e2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4954?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/schema-generator/src/formatters/common-fabric-formatter.ts
+++ b/packages/schema-generator/src/formatters/common-fabric-formatter.ts
@@ -285,7 +285,12 @@ export class CommonFabricFormatter implements TypeFormatter {
               ? { not: true, default: defaultValue }
               : { default: defaultValue }) as JSONSchemaObjMutable;
           }
-          (valueSchema as Record<string, unknown>).default = defaultValue;
+          // Rebuild rather than mutate: `valueSchema` is a formatted
+          // sub-schema and is deep-frozen.
+          return {
+            ...(valueSchema as Record<string, unknown>),
+            default: defaultValue,
+          } as JSONSchemaObjMutable;
         }
       }
 
@@ -957,7 +962,12 @@ export class CommonFabricFormatter implements TypeFormatter {
           ? { not: true, default: defaultValue }
           : { default: defaultValue }) as JSONSchemaObjMutable;
       }
-      (valueSchema as any).default = defaultValue;
+      // Rebuild rather than mutate: `valueSchema` is a formatted sub-schema and
+      // is deep-frozen.
+      return {
+        ...(valueSchema as Record<string, unknown>),
+        default: defaultValue,
+      } as JSONSchemaObjMutable;
     }
 
     return valueSchema;
--- a/packages/schema-generator/src/formatters/intersection-formatter.ts
+++ b/packages/schema-generator/src/formatters/intersection-formatter.ts
@@ -194,8 +194,12 @@ export class IntersectionFormatter implements TypeFormatter {
                 const priorComment = typeof existing.$comment === "string"
                   ? existing.$comment as string
                   : undefined;
-                (existing as Record<string, unknown>).$comment = priorComment ??
-                  DOC_CONFLICT_COMMENT;
+                // `existing` is a formatted property schema and is deep-frozen;
+                // store an updated copy rather than mutating it.
+                mergedProps[key] = {
+                  ...(existing as Record<string, unknown>),
+                  $comment: priorComment ?? DOC_CONFLICT_COMMENT,
+                } as JSONSchemaObjMutable;
                 logger.warn(
                   "schema-gen",
                   () => `Intersection doc conflict for '${key}'; using first`,
@@ -270,8 +274,16 @@ export class IntersectionFormatter implements TypeFormatter {
       missingSources: string[];
     },
   ): JSONSchemaObjMutable {
-    const { schema, docTexts, documentedSources, missingSources } = data;
-    if (!isRecord(schema)) return schema;
+    const {
+      schema: originalSchema,
+      docTexts,
+      documentedSources,
+      missingSources,
+    } = data;
+    if (!isRecord(originalSchema)) return originalSchema;
+    // `originalSchema` is a formatted schema and is deep-frozen; the doc
+    // attachments below mutate, so work on a mutable shallow copy.
+    const schema = { ...originalSchema } as JSONSchemaObjMutable;
 
     const uniqueDocTexts = docTexts.filter((doc, index, arr) =>
       arr.indexOf(doc) === index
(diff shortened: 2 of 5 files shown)


danfuzz:
Closing: the premise doesn't hold. Deep-freezing generated schemas was meant to
let the value model's hash cache reuse subtree hashes during deduplication, but
measurement shows it is a net cost with no offsetting benefit, because there is
no in-process schema-reuse for the cache to hit:

- A generation-time schema is hashed at most once (by the dedup), walked once by
  `createSchemaAst` to emit as an AST literal, then discarded.
- `packages/schema-generator/src` does no hashing of its own; the runtime
  re-parses the emitted source into fresh objects, never the frozen ones.

Best-of-3, same harness, one moderately complex type: freeze alone +6%, hash
dedup alone +8.5%, both stacked +14% vs main — the costs add, the stack is
slower than either piece alone, so the cache never pays off here. (All tiny in
absolute terms, but the direction is wrong for a change whose sole justification
was speed.)

Separately, the approach here froze at `formatType` output, minimizing
mutability during construction rather than freezing only where a hash needs it —
more than the goal ever called for.

The dedup work this was meant to accelerate (#4942) stays on its own merits:
value-model-correct equality that aligns the schema system with `FabricValue`,
at negligible absolute cost.

-- Claude (Opus 4.8), acting on @danfuzz's behalf

PR #4958, opened 2026-07-23
fix(cli): make JSON command behavior consistent

Hixie:
The cf command tree exposed --json in several incompatible ways. Some paths ignored it, some mixed human diagnostics into stdout, and others treated runtime arguments differently depending on whether they came from positional values, standard input, or generated schema flags. That made structured output unsafe for callers and left help and documentation out of sync with actual behavior.

Reserve stdout for command data whenever JSON output is active and route diagnostics to stderr. Reject unsupported and mutually exclusive output modes, add atomic JSON output to cf check, preserve JSON null and bigint semantics, silence render watch output, and expose the direct FUSE commands without obsolete aliases.

Use the same callable input resolution for cf piece call and cf exec across inline JSON, files, positional values, implicit standard input, and schema-derived flags. Enable output routing before callable resolution so parse and runtime failures cannot corrupt structured stdout. Include piece search's JSON mode, restore direct exec help, and document the complete command contract.

Add focused CLI and integration coverage for JSON output, conflicts, missing values, bigint serialization, search, callable input modes, and diagnostic routing.

Verified with deno fmt --check and deno lint.

---
<a href="https://app.blacksmith.sh/commontoolsinc/codesmith/labs/pr/4958"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787438548&installation_model_id=428580&pr_number=4958&repository=commontoolsinc%2Flabs&return_to=https%3A%2F%2Fgithub.com%2Fcommontoolsinc%2Flabs%2Fpull%2F4958&signature=055c54e7ef45d8e641fd90c1c00f59f495b8c609d1815bc80fdbe292a9e524f3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Unifies --json across the `cf` CLI so scripts can safely parse stdout. Preserves the selected output mode while resolving callables, derives JSON selection from arguments early, and routes runtime logs to stderr in JSON mode.

- New Features
  - Consistent `--json` across commands; rejects unsupported or conflicting modes and routes help/errors to stderr.
  - `cf check --json` outputs a single JSON object without evaluating (compiled modules); `--pattern-json` still evaluates and now fails on non-serializable exports; preserves bigint.
  - `piece get` prints JSON and uses `null` for missing values.
  - Unified input for `cf piece call` and `cf exec`: inline JSON, `--json [<json>]`, `-- --json-file`, stdin, or schema flags; JSON selection happens before manager creation and plain-text mode is preserved; output routing happens before parsing.
  - `piece search` has JSON mode; `wish` always outputs JSON (accepts `--json`).
  - `cf inspect` data commands support `--json`; the HTML renderer rejects it.
  - Help updated to include `--json [<json>]` and `--help --json`; restored direct `exec` help.
  - Reduced watch/render noise and redirected runtime console output to stderr in JSON mode (without changing console methods); exposed direct FUSE commands and removed obsolete aliases.

- Migration
  - Scripts expecting the literal `undefined` from `piece get` should now handle `null`.
  - When using `--json`, read data from stdout only; errors and hints moved to stderr.
  - Mutually exclusive output flags now cause the command to fail.

<sup>Written for commit 19937bb49367cbeb5e2ec97205f4186c799a435c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4958?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mathpirate:
Following up on the coverage report with a more specific suggestion: I think the highest-value additional test would cover the new `PieceManager → RuntimeConsole → runtime consoleHandler` path. The existing tests establish that `stderrConsoleHandler` works in isolation, but they don't appear to prove that diagnostics originating inside `PieceManager` actually reach it.

One focused `packages/piece` test could install a spy console handler, deliberately trigger one of the newly redirected error branches—such as a failed `getRaw()` or malformed result-link traversal—and assert that a `debug` event with the expected arguments reaches the handler.

If practical, an additional CLI smoke test could trigger a manager diagnostic during a `--json` command and assert both that stdout remains valid JSON and that the diagnostic appears on stderr. That would directly protect the core contract this PR is introducing.

— Sol, reviewing on behalf of Gideon

mathpirate:
Since this removes the obsolete `cf dev` command, could we update the remaining live documentation that still presents it as available? I found references in:

- `docs/specs/pattern-imports/README.md` around lines 313, 429, and 526
- `docs/specs/pattern-imports/implementation-plan.md` around lines 714 and 886

Most of these can likely refer only to `cf check`; historical references could instead be labeled explicitly as historical. While touching the CLI section, `cf deploy` should probably become `cf piece new` as well, since this PR also removes the old `deploy` alias.

— Sol, reviewing on behalf of Gideon

Hixie:
Addressed all feedback, thanks!

PR #4961, opened 2026-07-23
test(runner): reproduce the stacked-commit pending-dependency flake

Hixie:
A foreground write can surface a raw "pending dependency not resolved" ConflictError to its caller. The end-to-end cf-render counter test hit this: its direct piece.result.set() read several documents that the reactive scheduler was concurrently writing, so the write's commit declared pending reads on those still-unconfirmed optimistic writes. One of those in-flight writes was then rejected, and the dependent commit, which named it, could never be resolved by the server. editWithRetry retries such a conflict without a catch-up gate (this particular ConflictError carries no readyToRetry), so under sustained churn it burns its bounded retry budget and re-throws the conflict out of the set().

The root of the surfacing is that the client does not reject a stacked commit off its dependency's drop. When an optimistic in-flight commit is rejected, a later in-flight commit whose pending read names that commit's localSeq is already doomed: the read points at a local sequence that will never become a confirmed seq. A correct client can prove that doom locally and reject the dependant immediately, through the same revert/retry path as any other conflict, without waiting for (or needing) the dependant's own server verdict. Today's client instead leaves the dependant in flight until the server rejects it, and only the server's verdict ends the wait.

This test drives the deterministic stacked-commit harness into exactly that state: T1 writes A and B and is rejected, with its verdict gated so that T2 — which writes D but reads A — is built and put on the wire while T1 is still pending. T2's own verdict is gated for the whole test, so the only thing that can settle T2 is a client-side rejection off T1's drop. After T1 drops, the test waits, bounded, for T2 to settle: a client that cascades the drop settles T2 in well under a millisecond, while a client that does not leaves T2 in flight until the bound fires. The bound only ever trips when the cascade is absent, so it detects that absence rather than timing a success. The test passes once the client-side cascade exists and fails on main today.

---
## Summary by cubic
Add a deterministic test that reproduces the stacked-commit “pending dependency not resolved” flake and verifies the client rejects a doomed dependant locally when its optimistic dependency is dropped. This prevents foreground `editWithRetry` writes from leaking a raw ConflictError under churn.

<sup>Written for commit ef04ce8efd224a19a0ca5ff880d751313fd94bff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4961?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/test/memory-v2-stacked-commit.test.ts
+++ b/packages/runner/test/memory-v2-stacked-commit.test.ts
@@ -2641,3 +2641,98 @@ for (
     await runStressSeed(seed);
   });
 }
+
+// When a commit that reads another commit's still-unconfirmed optimistic write
+// stacks behind it, and that earlier commit is then rejected, the later commit's
+// premise is gone: its pending read names a localSeq that will never become a
+// confirmed seq, so the server can only answer it with "pending dependency not
+// resolved". The client must instead reject the doomed dependant off the earlier
+// commit's drop alone — locally, without a server round-trip for the dependant.
+// Without that, the dependant stays in flight awaiting its own verdict, and a
+// foreground editWithRetry write in that position exhausts its bounded retry
+// budget and surfaces the raw conflict to its caller.
+Deno.test("memory v2 stacked commits: a dependant stranded by a dropped optimistic sibling is rejected off the drop alone, without its own server verdict", async () => {
+  const harness = await createHarness();
+  const g1 = Promise.withResolvers<void>();
+  const g2 = Promise.withResolvers<void>();
+  try {
+    // T1 optimistically writes A and B. It is destined to be rejected, but its
+    // verdict is gated on g1 so T2 is built and put on the wire while T1 is
+    // still in flight.
+    const t1 = beginBatch(harness, [
+      { op: "set", id: DOCS.A, value: valueFor("t1-a") },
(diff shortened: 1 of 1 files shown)


Hixie:
Test for https://github.com/commontoolsinc/labs/pull/4606 (and for a flake on main right now).

ubik2:
This repro now lives in #4606 as `079f96def` ("a dependant stranded by a dropped optimistic sibling is rejected off the drop alone, without its own server verdict"), with you as co-author — and it passes there, since that branch implements exactly the client-side cascade this test demands.

Two adaptations from your original, noted in the test comment: the 2s wall-clock absence bound became a settled-flag + bounded microtask drain (the fake-clock preload from #4988 freezes test-file timers, so the timeout arm of the race could never fire; the cascade path is pure promise flow, so a missing cascade now fails the assertion crisply instead of hanging), and your diagnostic message about editWithRetry burning its retry budget is preserved verbatim. The parts your PR description asked for beyond the basic cascade — the dependant's verdict staying gated forever, and asserting the server never judges it at all — are kept as-is.

PR #4962, opened 2026-07-23
docs(now-wish): stop presenting bare #now as reactive time

mathpirate:
## Summary

Follow-up to #4959. Every guidance surface that tells authors how to read time reactively presented bare `#now` as the answer — with `#now` listed first or alone: the `TimeCapabilityError` message itself, `pattern-development-guide.md`, `pattern-critique-guide.md`, the `computed`/`action`/`handler` concept docs, the `wish.md` target table, the `reactive-now.tsx` worked example's comments, and the `cf-review` / `fuse-agent` skills.

Bare `#now` is not reactive: it durably captures the piece's **first-ever** load time (keyed by the wish node's cause — every later reload, session, and runtime reads the original capture) and never advances. The runtime's own error message steered pattern authors into exactly the trap #4922 fell into, where the lunch poll's current-day filter froze at the poll's birth day.

This change says it plainly at all eleven sites: **reactive/live time is the interval `#now/N` wish; bare `#now` is a durable first-load capture — right for created-at stamps, wrong for clocks.** The `cf-review` skill now also asks reviewers to flag bare `#now` used as a clock.

## Notes

- Text-only: no behavior changes. The `TimeCapabilityError` string is not pinned by any test (checked).
- `wish.md`'s example already used `createdAt` for the one-shot — the correct idiom — so only its comments sharpened.
- `packages/patterns/factory-outputs/lot-watch/DESIGN.md` mentions a future "`#now`-driven purge" generically; left alone (it makes no claim about bare-`#now` semantics).

## Validation

- `deno task check-docs` — all 518 code blocks pass
- `deno task check-skill-facts` — OK
- `deno task cf check packages/patterns/examples/reactive-now.tsx --no-run` — clean
- `deno check packages/runner/src/builder/safe-builtins.ts` + `deno fmt --check` on all touched files — clean

---
## Summary by cubic
Clarifies time semantics across docs, examples, skills, and the `TimeCapabilityError` message: bare `#now` is a durable first-load timestamp (shared across reloads/runtimes); `#now/N` is the reactive clock; the handler/action clock is a one-off event timestamp (the triggering event’s instant, frozen for the cascade and coarsened to 1s), not a live wall clock. Text-only updates; no behavior changes.

- **Migration**
  - Replace `#now` used as a clock with `#now/N` (N=1–86400 seconds).
  - For a one-off event timestamp, call `Date.now()` / `new Date()` in a handler or action — the event’s instant, frozen for the cascade and coarsened to 1s (not a live wall clock).
  - Keep `#now` only for created-at timestamps.

<sup>Written for commit 2e1bcddacf72defe3c14d420cc829660ca2a4989. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4962?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/common/ai/pattern-critique-guide.md
+++ b/docs/common/ai/pattern-critique-guide.md
@@ -46,7 +46,7 @@ Allowed inside patterns:
 | `.get()` on computed or lift result | Access directly; only `Writable` uses `.get()` |
 | `items.filter(...)` inline in JSX | Wrap in `computed()` outside JSX |
 | `items.sort(...)` inline in JSX | Wrap in `computed()` outside JSX |
-| `Date.now()`/`new Date()` or `Math.random()` in a `computed()` or `lift()`, or at pattern-body level | These built-ins throw a `TimeCapabilityError` outside a handler; read them in `action()`/`handler()`, or use the `#now` wish for reactive time |
+| `Date.now()`/`new Date()` or `Math.random()` in a `computed()` or `lift()`, or at pattern-body level | These built-ins throw a `TimeCapabilityError` outside a handler; read them in `action()`/`handler()`, or use the interval `#now/N` wish for reactive time (bare `#now` is a frozen one-shot) |
 | `Date.now()`/`new Date()` or `Math.random()` inside a re-running computation (`computed()` or `lift()`) without clear intent | Move the snapshot into `action()`, `handler()`, or one-time initialization |
 | nested computed with outer-scope reactive vars | Pre-compute with lift or an outer computed |
 | `lift()` closing over reactive deps | Pass dependencies as explicit parameters |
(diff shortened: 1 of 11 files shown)


Hixie:
There's also `new Date` which in some contexts gives you the current time.

Hixie (approved):
LGTM

...but I would maybe ask your agent to see if there's anywhere where it would make sense to mention that to get a one-off current time you can use `new Date()`.

PR #4933, opened 2026-07-23
fix(patterns): home root fields ride Default<> so old/bricked homes heal (estuary)

bfollington:
fix(patterns): home data fields ride Default<> so old roots heal (estuary)

## Why

Live on estuary after deploying #4900+#4901+#4926 (commit 35f388ee1): the
cold-start setup repair now FIRES on the bricked home root — exactly as
designed — but its setup commit is refused by CFC schema-merge:

  StorageTransactionAborted: CFC enforcement rejected commit: relevant
  transaction was not prepared — required field favorites needs a default
  to preserve old documents  (cfc-relevant-transaction-not-prepared)

home.tsx's HomeOutput declares favorites/journal/spaces/defaultAppUrl as bare
Writable<...> with no default. Materializing this pattern over a home root doc
that predates those fields hits mergeRequired's additive-required guard, the
repair fails closed (rethrowing the original "Handler used as lift"), and the
home stays bricked. This is the next layer of the exact cascade #4901 named:
"first-absence-wins, so each fixed field unmasks the next." #4901 fixed the
profile-vintage layer and, per seefeldb, gave the post-genesis DATA fields
(bio, isEditing) a Default<>; these four are the rest of that same set, using
the same sanctioned mechanism (favorites-manager already spells its favorites
list this way).

## What changed

- HomeOutput: favorites/journal/spaces/defaultAppUrl gain a Default<> arm (an
  empty-tuple default, or "" for the string). The body already seeds [] / "",
  so no runtime change — this only puts the default into the emitted schema so
  old docs merge.
- home.vintage-defaults.test.ts: source-text regression pin (mirrors
  profile-home.owner-gated.test.ts) so dropping a default goes red here instead
  of re-bricking estuary.

Scope note: profiles/mru (TrustedProfileList/Mru) are ALSO required-no-default,
but they are CFC-wrapped (WriteAuthorizedBy) — defaulting them is #4871/#4901
writer-identity territory and wants seefeldb. Whether a REAL bricked home hits
them depends on its vintage (the synthetic estuary testbed, a 5-line doc, hits
everything); Gideon's own-home reload is the discriminator. Deferred
deliberately.

## Test plan

- deno check packages/patterns/system/home.tsx — clean.
- new regression test: 5 steps green.
- cfc-schema-merge.test.ts (the mergeRequired mechanism) still green.
- fmt/lint clean.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

---
[...]

PR #4967, opened 2026-07-24
fix(piece,runner): heal home roots pinned to unrunnable patterns; land CFC durable-data scoping (estuary)

bfollington:
## Why

Estuary home spaces stayed bricked even after the overnight fixes (#4900,
#4901, #4926, #4933; live at `67abf6131`). Those fixes are correct, but they
healed **no existing home**, because two independent defects keep them from
ever taking effect on a root that is already pinned to an old pattern:

1. **The default-root auto-heal gates on loadability, not runnability.** The
   pattern-updater only rolls a stale root forward if its old pattern won't
   *compile*. An aged official `home.tsx` compiles fine — it just can't migrate
   the reused doc at the CFC setup step — so the root stays pinned forever, and
   the cold-start setup repair (#4926) re-runs that same stale pattern and
   rethrows an opaque error.

2. **The current `home.tsx` migration itself still throws.** `#4933` defaulted
   the six data fields, but the rejection just advanced to the next
   required-no-default field: `defaultProfile` (the schema-generator decides
   `required` from the `?` marker, **not** the value type, so `… | undefined`
   was still emitted required-no-default) and then the exported handler streams
   (which carry no preservable document value and so cannot take a `Default<>`).

Neither half heals alone: without (2) the current pattern can't migrate a
pre-favorites doc; without (1) nothing moves an already-pinned root onto it.
This PR lands both, so an existing bricked home rolls forward to the current
official pattern and heals — proven end-to-end under CFC enforcement, not just
by an identity swap.

This **supersedes #4936** — its stream exemption + `defaultProfile?` are folded
in here (its own branch was 25 commits stale with a red check).

## What Changed

**The migration fix (folds in #4936), `d3ffceb6f`:**
- `cfc/schema-merge.ts`: exempt an additive-required **stream slot**
  (`asCell: ["stream"]`) from the needs-a-default guard — a stream is a
  runtime-materialized capability marker, not stored data, so an old doc that
  lacks it has nothing to preserve and no meaningful default to declare.
- `patterns/system/home.tsx`: `defaultProfile?` — genuinely optional, so it is
  dropped from `required` and needs no default.
- Unit + end-to-end tests (`cfc-additive-default-preserves-old-doc.test.ts`)
  that run the **real** `home.tsx` setup over a realistic pre-favorites,
  CFC-relevant root **with enforcement ON** — the layer #4926/#4933's tests
  were blind to.

**The runnability backstop (element 2), `b1e49a2df`:**
- `piece/ops/pieces-controller.ts`: `startEnsuredDefaultPattern`'s cold-start
  repair now escalates on **exactly one signal** — `isCfcMigrationRejection`,
  the same `"CFC enforcement rejected commit"` prefix the runner uses at its own
  setup-commit boundary — to a roll-forward: compile the space's **current
  official** pattern (ETag-revalidated so a stale HTTP cache can't hand back the
  aged bytes we're trying to escape), swap the root's identity to it (recording
  `displacedPattern` for recovery), and materialize it over the reused doc.
  Every **other** repair failure stays **fail-closed**: the original start error
  surfaces, nothing changes, the next boot retries. Custom roots that migrate
  cleanly never reach the backstop, so they are preserved for free.
- Outcome is legible: **healed**, or **one clear error** naming where the
  roll-forward stopped — no more reverse-engineering scattered
  `$stream` / `needs-a-default` logs.

## Test plan

- New enforce-on orchestration test in `check-update-default-pattern.test.ts`
  drives the full path (pinned OLD root → repair rejected by CFC migration →
  roll forward → official pattern materializes) and asserts a **functional
  read**, not just an identity swap: the once-fatal required field materializes
  to its default, and the handler stream fires end to end over the reused doc.
- The existing **fail-closed** tests stay green **unchanged** — the gate keeps a
  transient blip from swapping a healthy root's identity out from under it.
- Green locally: full `packages/piece/test/` (22 files / 258 steps) +
  `cfc-additive-default-preserves-old-doc.test.ts` (3 steps).
- **Before deploy:** acceptance-test against a real bricked vintage (the gate
  whose absence let this class ship green last time).

## Follow-up (not in this PR)

A simplification PR will replace the growing pile of per-field defaults and
stream exemptions with one principled rule — distinguish **input** (argument,
must default) from **output** (result, always compatible) cells at the CFC
merge — and a migration-kind matrix test. That retires #4933's per-field
defaults and this PR's #4936-derived exemptions. Deferred because it hinges on
threading the input/output role down to the CFC prepare layer, and is
correctness-simplifying rather than incident-critical.

Design & root-cause record:
`docs/development/projects/estuary-home-root-heal/design-and-handoff.md`.

---
<a href="https://app.blacksmith.sh/commontoolsinc/codesmith/labs/pr/4967"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787443975&installation_model_id=428580&pr_number=4967&repository=commontoolsinc%2Flabs&return_to=https%3A%2F%2Fgithub.com%2Fcommontoolsinc%2Flabs%2Fpull%2F4967&signature=2dfb3db2e397a4b2f98e5d192c3350ec8662e12d08d0496aea22a950f2726c42"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Heals bricked Estuary home spaces by rolling stale pinned roots forward only on true CFC additive-required migration incompatibility, scoped to durable data. Adds a safe, ETag-revalidated compile, a transactional preconditioned swap, mapped-host resolution, and clear error surfacing.

- **Bug Fixes**
  - `packages/runner/src/cfc/schema-merge.ts`: Exempt only real stream slots from the additive-required default rule using the first normalized `asCell` kind via `ContextualFlowControl` (supports scoped descriptors; does not exempt `["cell","stream"]`); raise `CfcSchemaMigrationError`.
  - `packages/runner/src/cfc/migration-reason.ts`, `packages/runner/src/cfc/prepare.ts`, `packages/runner/src/cfc/mod.ts`, `packages/runner/deno.jsonc`: Introduce and export `CFC_SCHEMA_MIGRATION_INCOMPATIBLE_REASON`; tag prepare reasons with the framed token `: cfc-schema-migration-incompatible:` so consumers can reliably detect the recoverable class across the runner’s plain `Error` boundary.
  - `packages/piece/src/ops/pieces-controller.ts`: Runnability backstop now triggers only on the framed migration token; compiles the official pattern via `hostForSpace` with `cache: "no-cache"`; guards the identity swap with a transactional precondition; short-circuits “already official” only when identity+symbol match; surfaces one clear error per failure and chains the actual cause; covers no-entry, compile failure, swap-commit failure, and superseded-swap abort (fail-closed).
  - `packages/patterns/system/home.tsx`, `packages/patterns/system/home.vintage-defaults.test.ts`: Make `defaultProfile` optional (`?`) and document that `?`, not `| undefined`, controls `required`.
  - Tests: Runner adds an enforcement-on end-to-end repro and the framed-token contract; schema-merge tests cover scoped stream descriptors and non-exemption of `["cell","stream"]`; piece tests cover heal, non-migration CFC rejection (fail-closed), token-collision guard, superseded-swap abort, already-on-official (identity+symbol), symbol-differs roll-forward, mapped-host resolution, official compile/no-entry failures, and swap-commit failure.

<sup>Written for commit eb898ad876acd11ee353f7a82f565cb398313699. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4967?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bfollington (commented):
*cf-review via Codex GPT-5, on behalf of @bfollington*

I reviewed the full changeset at `79528ed9c`, including the linked incident PRs. The incident happy path is well motivated and its focused tests pass, but I found three blocking correctness risks in the broader compatibility behavior.

## Blocking

1. **The roll-forward trigger treats every CFC rejection as a schema migration rejection** (`packages/piece/src/ops/pieces-controller.ts:83-85`). The `CFC enforcement rejected commit` prefix is also emitted for generic unprepared or invalidated transactions and prepared-digest mismatches. Those can reflect policy, provenance, or ordering faults, but this predicate converts all of them into authority to repoint the root to the official pattern. Please introduce a structured schema-migration error or reason and add negative tests proving other CFC-prefixed failures leave identity untouched.

2. **The identity swap lacks a transactional precondition** (`pieces-controller.ts:807-816`). After the failed repair, fetching and compiling the official source opens a race window. If another process replaces or heals the root, `editWithRetry` reruns this callback against the new state and still blindly overwrites its `patternIdentity`. Re-read the identity inside the swap transaction, require it to equal `pinnedRef`, and abort or re-resolve if it changed. The existing pattern updater `stillMatches` / `canWrite` logic is the precedent.

3. **`isStreamSlot` does not follow the supported `asCell` dialect** (`packages/runner/src/cfc/schema-merge.ts:338-340`). `includes("stream")` misses an outer scoped stream such as `[{ kind: "stream", scope: "user" }]`, while it exempts `["cell", "stream"]` even though the immediate outer slot is a cell. I reproduced both outcomes against this head. Normalize the first `asCell` entry as string or descriptor and compare its kind to `stream`; add scoped and nested cases.

## Additional issues

- CI is red because the new piece code adds 31 uncovered lines. The new test exercises the successful incident path but not the fallback and race paths above. Please add real tests rather than accepting a new coverage baseline.
- `home.tsx:72-73` still says the `| undefined` value type makes `defaultProfile` non-required. That is the exact assumption this PR corrects; the `?` marker is what controls requiredness.
- The compile and swap failure branches retain `migrationError` as `cause`, losing the actual compile or swap stack despite the clear-error contract.

Verification: piece test 48 steps; new runner test 3 steps; CFC schema, writer, and profile safety suite 59 steps; touched-source fmt and lint all passed. CI currently reports 40 successful checks, 4 skipped, and 1 failed Performance Check.

**Verdict: request changes before merge.**

bfollington:
Thanks — all three blocking items were real; each is verified against the code and fixed in `9a198abdf`.

## Blocking

**1. Roll-forward trigger treated every CFC rejection as a schema-migration rejection.** Confirmed: `extended-storage-transaction.ts` emits the `CFC enforcement rejected commit` prefix for `prepared digest changed` and `relevant transaction was not prepared` (policy/provenance/ordering) alike, so the bare-prefix predicate could repoint a healthy root on a transient blip. Introduced a machine-stable token, `cfc-schema-migration-incompatible` (new dependency-light `packages/runner/src/cfc/migration-reason.ts`): the CFC merge now throws a typed `CfcSchemaMigrationError`, the prepare catch (`prepare.ts`) tags *only that class's* recorded reason with the token, and `isCfcMigrationRejection` now requires the token. It's carried as a string rather than an error class deliberately — I traced the runner's setup-commit boundary (`runner.ts` `throw new Error(error.message, { cause })`) and only the message survives that re-wrap, so a token in the message is what keeps producer and cross-package consumer in lockstep. Negative test added: a `prepared digest changed` rejection now stays fail-closed with the root identity untouched.

**2. Identity swap lacked a transactional precondition.** Fixed as suggested — the `editWithRetry` callback re-reads `getPatternIdentityRef` inside the transaction and proceeds only if it still equals `pinnedRef`, otherwise returns `false` to abort without committing (the `stillMatches`/`canWrite` precedent). `result.ok === false` is then handled as "superseded": we leave the concurrent heal in place and return its root rather than clobbering it. Test added: a concurrent repoint to a third identity mid-repair leaves that identity intact, with no displacement recorded and the official identity never installed.

**3. `isStreamSlot` didn't follow the `asCell` dialect.** Fixed — it now reads the first entry via the canonical `ContextualFlowControl.getAsCellKind(getAsCellValues(schema).at(0))`, matching the stream test in `link-utils.ts` / `schema.ts`. `["cell","stream"]` (a cell of a stream — outer slot holds data) is no longer exempted, and the `{ kind, scope }` descriptor form is normalized. Three unit cases added: `["stream"]` exempt, scoped-descriptor exempt, `["cell","stream"]` NOT exempt.

## Additional issues

- **Coverage / real tests.** Added the fallback and race paths as behavioral tests, not a baseline bump: CFC-prefixed-non-migration → fail-closed; swap aborts on concurrent identity change; official-pattern-also-fails → one clear error. The prior synthetic rejection string was also slightly wrong (missing the `relevant transaction was not prepared:` middle) and is now built from the shared token, so it mirrors the real production string exactly.
- **`home.tsx` comment.** Corrected — the `?` marker, not the `| undefined` value type, is what drops `defaultProfile` from the generated `required` list; the comment now says so and notes a bare `T | undefined` would still trip the guard.
- **Lost compile/swap stacks.** The compile and swap failure branches now chain the *actual* `compileError` / `swapResult.error` as `cause` (the migration reason is already named in the message).

## Verification

Green locally: `cfc-schema-merge` (27 steps), full `packages/piece/test/` (22 files / 261 steps), CFC sweep — boundary / additive-default / writer-claim / sink-ceiling (144 steps); typecheck + fmt + lint clean on all touched files.

One item explicitly **deferred** (tracked, not in this PR): the input-vs-output cell distinction at the CFC merge that would let output additive-required fields skip the default requirement wholesale. That's the principled simplification this PR's stream exemption is a special case of, but it hinges on threading the argument/result role down to the prepare layer, so it's a follow-up rather than an incident fix.

bfollington (commented):
*cf-review follow-up via Codex GPT-5, on behalf of @bfollington*

Re-reviewed latest head `9a198abdf`. The original three blocking findings are substantially addressed: the stream dialect fix, transactional identity precondition, corrected home comment, and error-cause chaining all look correct. Two new blocking issues remain in the revised recovery path.

## Blocking

1. **A superseded roll-forward reports success without starting the replacement root** (`packages/piece/src/ops/pieces-controller.ts:854-865`). This branch is reached after `startPiece` and the same-identity setup repair have both failed. It returns the concurrently repointed root, but `startEnsuredDefaultPattern` then exits the catch and returns a controller without starting or materializing that root. The new race test installs `SOURCE_V3_HANDLER`, which requires setup-written stream markers, but asserts identity only; its "no throw" expectation therefore codifies a silent-success regression. Either start and repair the newly resolved root before returning, or fail closed and let the next boot retry. Add the same functional `count` / `bump` assertion used by the successful heal test.

2. **The recovery token can match unrelated, user-influenced CFC errors** (`pieces-controller.ts:91-94`). `message.includes(CFC_SCHEMA_MIGRATION_INCOMPATIBLE_REASON)` does not prove prepare emitted the token as its reason code. I reproduced an ordinary incompatible type change at property path `/cfc-schema-migration-incompatible`; its untyped merge error contains the token and this predicate classifies it as recoverable. That again authorizes root replacement for a non-additive incompatibility. Match the framed reason position exactly, such as the full `...not prepared: <token>:` prefix, or preserve a structured reason through `cause`. Add a collision test using an unrelated schema error whose path contains the token.

## Test and CI status

The new producer/consumer token contract also lacks an end-to-end test: schema tests exercise `mergeCfcSchemaEnvelopes`, while piece tests synthesize the final rejection string. Please add a runner test covering `CfcSchemaMigrationError -> prepare reason -> commit rejection` and assert the exact framed token.

The Performance Check remains red with 19 additional uncovered lines in `pieces-controller.ts`. Cover the recovery branches behaviorally rather than accepting a new baseline.

Verification on this head: piece test 51 steps; schema/additive CFC tests 30 steps; fmt and lint on all eight incremental files passed. CI reports 40 successful checks, 4 skipped, and 1 failed.

**Verdict: request changes before merge.**

bfollington:
Both re-review blockers fixed, plus the two test/CI follow-ups. Pushed through `8b3de826a`.

## Blocking

**1. Superseded roll-forward reported success without starting the root.** Correct — on the cold-start path the caller doesn't start what the backstop returns, and the concurrent heal's repoint commits before its own materialize, so the returned root could be unstarted/un-setup. Now it **fails closed** with a clear `superseded by a concurrent heal` error; nothing is overwritten. The race test was rewritten to assert the throw, that our stale roll-forward did **not** clobber the concurrent identity (no displacement, official never installed), **and** that the next boot starts the concurrent root through the ordinary repair and it works end to end (`count`/`bump`) — closing the silent-success gap you flagged.

**2. Token could match unrelated, user-influenced CFC errors.** Reproduced and fixed. The predicate now matches the token only in its **framed reason position** — `: <token>: ` — the exact shape the prepare catch emits (`${token}: ${message}`, surfaced by the commit as `…not prepared: ${reason}`). An incompatible-type error at a property path named `/cfc-schema-migration-incompatible` can't forge that framing. Added a collision test with exactly that path asserting it stays fail-closed.

## Test and CI

- **Producer→consumer contract test (runner layer).** Added: a real additive-required-no-default commit is driven through `mergeRequired → prepare reason → commit rejection`, asserting the message carries the framed `: <token>:`. So if either side's wording drifts, this test breaks rather than the contract silently.
- **Performance Check / coverage.** The remaining uncovered lines were the backstop's two defensive exits (official pattern yields no entry identity; the swap's own commit fails, distinct from the precondition abort). Both now have deterministic behavioral tests — one stubs the entry-ref to `undefined`, the other forces the swap commit to fail — each asserting the clear, chained error and that the pinned identity is untouched. `pieces-controller.ts` heal region is now fully covered (no newly-uncovered lines in the diff).

Verification: `check-update-default-pattern` 56 steps, `cfc-additive-default-preserves-old-doc` 4 steps (incl. the new e2e token test), `cfc-schema-merge` 27 steps; typecheck + fmt + lint clean on all touched files.

That's the full set of findings addressed. The only thing I have not done is the deferred input-vs-output CFC merge distinction (the principled generalization this PR's stream exemption is a special case of) — it needs the argument/result role threaded to the prepare layer and is a follow-up, not part of unbricking estuary.

PR #4917, opened 2026-07-22
topics: make agent attribution atomic and profile-native

mathpirate:
## Summary

- replace the shared `myName` mutation flow with per-mutation structured attribution
- use canonical Profile snapshots for browser-authored topics, comments, links, and body saves
- require CLI/headless mutations to carry `agentName` atomically with their content
- retain `myName`, `createdByName`, and `authorName` only as optional legacy read/render fallbacks
- show body/link authorship and include their timestamps in `lastActivityAt`

## Design boundary

Fabric remains authoritative for the human principal behind the identity key, operation history, and concurrency. This change intentionally adds neither a duplicate activity ledger nor an application-level revision/CAS protocol; Topics should expose gaps in those Fabric guarantees through dogfooding.

This is designed for the short-term model established by the Profiles workstream (including loom#4074): an agent uses its human user's key and supplies its own name as a content-level signature.

## Verification

- `deno task cf check packages/patterns/topics/main.tsx --no-run`
- `deno task cf check packages/patterns/topics/topic.tsx --no-run`
- `deno task cf check packages/patterns/topics/topics.test.tsx --no-run`
- `deno task cf check packages/patterns/topics/multi-user.test.tsx --no-run`
- `deno task cf test packages/patterns/topics/topics.test.tsx` (29 passed)
- `deno task cf test packages/patterns/topics/multi-user.test.tsx` (7 passed)
- focused `deno lint`, `deno fmt --check`, and `git diff --check`

---
[...]

PR #4968, opened 2026-07-24
docs(plans): pattern verb contract — agent-drivable verbs and the invocation protocol

mpsalisbury:
## Why

Agents driving patterns headlessly is the current application focus (Topics first), and today the interaction is poor in ways client-side care cannot fix: filing one topic takes six CLI invocations, a create returns no handle, a verb that declines on its own terms is indistinguishable from success, and a retry after a sync timeout can mint duplicates. The fixes people reach for — a `cf topic` command, fatter verbs, a `lastCreated` cell — each bake a workaround into a permanent surface.

This plan records the design before any implementation: a **verb contract** pattern authors adopt (named fields, valid-on-its-own, declared results, typed rejection, identity addressing, no implicit call-order dependence), and an **invocation protocol** that is deliberately not new machinery — the scheduler already mints durable event ids, accepts caller-supplied ones (`queueEvent` `opts.eventId`), writes per-invocation result cells, and enforces exactly-once via create-only receipts. Part 2 exposes that through the callable layer.

Per the doc's own status line, it exists to be argued with; landing it is staging step 1 ("agree this document"). The open questions that most need reviewers: retention windows (they bound the idempotency guarantee), whether an actor can ever be authenticated (the two-key/delegation question), and how plain JSON returns reach the receipt (the one place the design asks the runtime for new behavior rather than exposure).

Related: the board topic "Give the topics board an agent handle" (this generalizes its five asks), and the attribution rework in #4917, which this builds on.

## What Changed

- Adds `docs/plans/pattern-verb-contract.md`: goal, the six-invocation problem, an inventory of the existing callable/scheduler machinery with an essential-vs-incidental analysis and the obligations taken on, the Part 1 verb contract (attribution, discovery, the atomic-unit composition rule, positional-argument policy), the Part 2 invocation protocol (`Invocation` shape, settlement, id choice, retention, storage/privacy, retries-and-failure semantics, verb discovery and listing policy), design decisions worth recording, open questions, and staging.

## Test plan

- `deno task check-docs plans` — the doc's code blocks compile (marked illustrative where the API does not exist yet).
- No code changes.

---
<a href="https://app.blacksmith.sh/commontoolsinc/codesmith/labs/pr/4968"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787445473&installation_model_id=428580&pr_number=4968&repository=commontoolsinc%2Flabs&return_to=https%3A%2F%2Fgithub.com%2Fcommontoolsinc%2Flabs%2Fpull%2F4968&signature=f723672d794be61472e3b291d41e63998e83fc994c91347eb55d00600611288b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Adds a design draft for a pattern “verb contract” and an invocation protocol so agents can drive any pattern headlessly without pattern-specific CLI code, plus a phased implementation plan to ship it safely. The plan starts small (topics and CLI hygiene), ships idempotent retries early, and gates plain-return projection behind a default-off flag.

- **New Features**
  - New doc: Part 1 (verb contract) — named fields, valid-on-its-own verbs, declared results, typed errors, identity-based addressing, no implicit call order.
  - Part 2 (invocation protocol) — caller-supplied IDs, per-invocation result cells, exactly-once via create-only receipts, explicit settlement/wait, and retention; plain JSON returns written into receipts behind an experimental flag.
  - New doc: Implementation plan — six workstreams and phases; start the authoring surface now (`Stream<T, R>`, result schemas, `VerbError`), replace CLI polling with sink-based settlement, ship idempotency first via `--invocation` (caller IDs) so duplicate-retry is fixed before returns; continuous dogfood on the live board.
  - Applies to `topics` first: `addTopic {title, body}` returns the topic handle; filing drops from six calls to three. Early changes include body-at-create, thrown rejections, and CLI wait hygiene. Open questions for review: retention window, actor authentication, and projecting plain JSON returns into receipts.

<sup>Written for commit fd6606f8b3fd91492f6483d0aa01e536b88d4c85. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4968?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/plans/pattern-verb-contract-implementation.md
+++ b/docs/plans/pattern-verb-contract-implementation.md
@@ -1,0 +1,226 @@
+# Pattern verb contract — implementation plan
+
+**Status:** pending. Executes the design in
+[`pattern-verb-contract.md`](pattern-verb-contract.md) (PR #4968). Keep current
+as work proceeds: check off exit criteria, record scope changes.
+
+## Governing decisions
+
+Made 2026-07-24, shaping everything below:
+
+1. **The authoring surface starts immediately.** Verb return values through
+   `action()` / transformers / schema-generator is the critical path and runs
+   as its own workstream from day one, in parallel with the small wins.
+2. **Plain-return projection ships behind a flag, default-off**, per the
+   `EXPERIMENTAL_OPTIONS.md` process; the default flips after the integration
+   suite proves readback end to end.
+3. **Continuous dogfood.** The live Estuary topics board gets `setsrc` as each
+   phase lands. The compat checker gates schema breaks; every phase ends with a
+   live-board acceptance pass.
+4. **This document plus tracked issues** (breakdown at the end) is the plan of
+   record; the design doc's staging section defers here.
+
+## Non-goals
+
+Named so their absence reads as intent, not oversight:
+
+- The **wrapper-tier marker** for verb listing (semantics are settled in the
+  design; the marker mechanism ships after the listing).
+- **Actor authentication / delegation** (design OQ2) — the envelope carries an
+  unverified actor regardless of how that lands.
+- **Cross-space effect atomicity** — the spec's I11 gap stands; the guarantee
+  is same-space, which covers `topics`.
+- **Batch / session mode** for the CLI — orthogonal call-cost work.
+
+## Workstreams
+
+### WS-A — `topics` Part 1 finish
+
+Size S (~1–2 days). No dependencies. `packages/patterns/topics`.
+
+- `AddTopicEvent` gains optional `body` — argument widening, compat-checker
+  clean — and `addTopic` creates the child with it, making `setBody` an
+  editing verb rather than part of every create.
+- Silent early-returns on mutating verbs become throws: empty title, blank
+  `agentName`, empty comment body, invalid link URL. UI composer wrappers
+  (`submitTopic`, `submitComment`, `saveBody`, …) keep their silent guards —
+  an empty draft is a non-event in a composer, a defect headlessly.
+- Tests: `topics.test.tsx` / `multi-user.test.tsx` cover body-at-create and
+  each thrown rejection (asserting no write happened).
+- Docs riding the change: `packages/patterns/topics/README.md`,
+  `skills/topics/SKILL.md` (`addTopic` example gains `body`;
+  `deno task check-skill-facts` gates the citations).
+- **Exit:** filing-with-body is five CLI calls; a blank `agentName` fails with
+  a nonzero exit; live board updated via `setsrc`.
+
+### WS-B — CLI settlement hygiene
+
+Size S (~1 day). No dependencies. `packages/cli`.
+
+- Replace `defaultWaitForResult` (25 ms poll, 15 s ceiling,
+  `lib/callable.ts:206-222`) with settlement observed through the existing
+  `running.sink(…)` path. The default time bound remains until WS-F makes the
+  wait caller-controlled, but it bounds an observation, not a poll.
+- Surface the tool result cell's address in `ExecutedCallable` output.
+- **Exit:** no sleep/poll in the callable wait path; `deno task test` in
+  `packages/cli` green.
+
+### WS-C — verb results authoring surface *(critical path)*
+
+Size L (~1–2 weeks). No dependencies; starts immediately.
+`packages/api`, `packages/ts-transformers`, `packages/schema-generator`,
+`packages/runner`.
+
+- **api:** `action` overloads accept a return type (today both overloads type
+  the callback `=> void`, `builder/module.ts:606-609`); `Stream<T, R = void>`
+  or equivalent so the result type is visible to the schema layer — the
+  defaulted parameter keeps every existing `Stream<T>` use compiling. A
+  `VerbError { code, message }` type for rule 4's typed rejections.
+- **ts-transformers:** lowering for value-returning `action` bodies; CTS spec
+  updates under `docs/specs/ts-transformer/`. (The runtime side already
+  consumes returns — `handleJavaScriptHandlerResult` — so this is authoring
+  surface, not execution semantics.)
+- **schema-generator:** emit a result schema for stream/handler properties so
+  it reaches the piece's **durable** schema — the dependency verb discovery
+  named; mapping spec update in
+  `docs/specs/schema-generator/ts_to_json_schema_mapping.md`.
+- **runner:** plain-return projection — the receipt-only branch
+  (`runner.ts:3713-3725`) writes the validated plain return instead of `{}`,
+  behind a new experimental option (registry entry in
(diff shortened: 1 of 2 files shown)


docs/plans/pattern-verb-contract-implementation.md
```diff
@@ -0,0 +1,226 @@
...
+### WS-F — client affordances
+
+Size M, mostly parallel. `packages/cli`, `skills/cf`.
+
+- `cf piece verbs --json` — name, kind, input schema per verb from the
+  existing classification (`packages/fuse/callables.ts:88`); result schemas
+  appear once WS-C lands; v1 lists everything, per the decided semantics
+  (every verb listable; tier filtering arrives with the marker, later).
+  **Independent — can ship first.**
+- `@name` client-side binding: `cf bind <name> <url>`, resolution in
```
seefeldb, line 155: how does this relate to slugs? feels similar enough that it might be confusing.

docs/plans/pattern-verb-contract.md
```diff
@@ -0,0 +1,703 @@
...
+facts and are carried separately; `topics` is the template:
+
+- The **principal** — whose key authorized the write — stays fabric-level:
+  CFC carries it in its integrity labels (`packages/api/cfc.ts:829-841`). For
+  display, the browser path resolves the viewer's canonical Profile
+  (`wish({ query: "#profileName" })`;
+  `docs/common/patterns/multi-user-patterns.md:263-272` — "the viewer is
+  whoever the runtime says they are") and stamps a structured
+  `TopicAuthor { kind: "person", name, avatar? }`.
+- The **actor** — what performed the write on the principal's behalf — is a
```
seefeldb, line 234: this feels very specific to topics, not general guidance.

if we want to make this general, i'd rather do this via additional CFC labels, after all they are designed to carry providence, and that is actually broken right now when we call from the "outside" via `cf`. So we could add an `AgentActor` atom type and it's metadata can be fairly rich (e.g. in Loom where we have specific agent roles, I'd want to reference those, or maybe directly the prompt used; in an interactive coding session, we could mention what the user asked for that triggered this, etc.).

we can then add helpers to extract that information as well.

docs/plans/pattern-verb-contract.md
```diff
@@ -0,0 +1,703 @@
...
+actor is a claim, not an authenticated fact — the principal vouches for it,
+the way a git signature vouches for an unverified author string. Whether an
+actor can ever be *authenticated* is a separate question about keys (open
+question 2).
+
+### Discovery
+
+A client holding only a board URL must reach the board's children without an
+O(children) sweep of per-child reads. So a parent exposes a **compact index** on
+its result — one row per child carrying its fid and the summary fields a survey
```
seefeldb, line 252: don't expose the fid as a row, the pattern cannot see it.

i think what we're missing is for agents to get the fid/path for any given bit of data they ask for. that has come up a few times.

a coarse option would be to just add a { "@id": .. } field or so whenever the id changes in the output when providing --include-ids or so. might be getting more than we need and those are long strings.

so the more narrow option would be to define paths for which those should be output, which has the other advantage that they might already contain the right /path if those are child elements (in the other approach the agent will see the @id on the parent and has to figure out on its own how to construct the path).

i could see both being useful, the former more when in exploration mode before understanding the data shape, so maybe build the first one first?

docs/plans/pattern-verb-contract-implementation.md
mpsalisbury: these are local, include the whole space/fid, and are per client. slugs are per-space and persistent. this is meant as a local naming mechanism for legibility reasons. probably warrants an 'unbind'. this is just syntactic sugar - no actual new capability. i don't know how confusing it would be.

docs/plans/pattern-verb-contract.md
mpsalisbury: i do want to make this general. i don't understand cfc labels, so this is foreign to me. would you recommend we leave this as a topic-specific argument for now and we can figure out the general mechanism next week?

docs/plans/pattern-verb-contract.md
mpsalisbury: i'll need more background to understand this, though fable seems to understand. updated the proposal docs.

bfollington:
Late context, all additive — no quarrel with any decision here; the core shape (verbs return addresses, the client decides whether to wait and renders identity) is exactly right.

**1. There's a second working precedent in-tree for Part 2 that the doc doesn't cite.** The llm-dialog builtin already runs the whole invocation protocol in miniature: it mints the result cell at a *caller-supplied* id (`toolCall.id`), hands it to the handler as `result` ("doesn't HAVE to be used, but can be"), and resolves off the commit callback — `packages/runner/src/builtins/llm-dialog.ts`, the tool-invocation path around `runtime.run(tx, pattern, invocationArgs, result)`. The scheduler receipt is the right substrate for the reasons the doc gives (create-only exactly-once), but llm-dialog is useful evidence that the caller-supplied-id → durable-result-cell ergonomics already work in production.

**2. The write-storm gate in the implementation plan's Risks now has a candidate preflight: #5059** (`cf piece setsrc --check`). It answers "can this source be applied to this piece" before attempting it, driving the real rules in dry-run (the schema subset proof, the CFC envelope merge, the retained-link validator) rather than a second copy. It doesn't measure commit rates, so the rehearsal step stands — but it removes discovering an incompatibility by attempting the swap live.

**3. A couple of small CLI ideas that came up while discussing Discovery, deferred and blocking nothing:** it could be useful for `cf piece get` to grow flags that let an agent control how much data comes back when reading cells interactively — e.g. a `--schema` override to read through a narrower schema (the CLI already narrows its own internal reads this way, e.g. `piece-render.ts`), and a limit on the number of records returned from large arrays. Adjacent work for when board scale demands it, not asks on this workstream.

PR #4972, opened 2026-07-24
fix(ts-transformers): builder-arg reactive computations get a hoist diagnostic instead of an internal crash

mathpirate:
## The crash

A reactive computation written inline in a builder call's arguments crashed the whole compile with an internal invariant error instead of a diagnostic (found while reworking lunch-poll for #4928):

```tsx
const boundJoin = joinAs({
  ...,
  profile: profile ?? profileWish.result,  // 💥 "Internal Common Fabric compiler error:
});                                        // binary expression compute-wrap decision
                                           // disagreed with reactive-context classification"
```

Any non-`&&/||` binary over reactive reads in builder args hits it (`users.length === 0` too). The two ends disagree by construction: `classifyExpressionSiteHandling` deliberately claims computation sites inside builder-call boundaries as helper-owned-lowerable (`HELPER_BOUNDARY_KINDS` includes `builder`), while the emitters' compute-wrap guard (`assertValidComputeWrapCandidate` → `isOwnedPatternBoundary`) refuses compute-wraps inside builder boundaries. Authored code reaching both = mid-emit throw.

## The fix

`validateComputeWrapCandidate` (renamed from `assertValidComputeWrapCandidate`) now distinguishes the culprit's provenance:

- **Authored culprit inside an owned boundary** → reports error **`reactive:call-argument-computation`** (deduped via `reportDiagnosticOnce`) and returns `false`; all four callers (binary, ternary-branch, helper-owned, pattern-body initializer) skip the wrap, so the pipeline surfaces a normal `TransformerError` with a source excerpt:

  ```
  main.tsx:43:14 - error: Reactive computation `profile ?? profileWish.result` cannot be
  compiled inline in the arguments of `join(...)`. Hoist it to a body-level const or
  computed(...) and pass that value instead.
  ```

- **Synthetic culprit, or culprit-already-compute** → keeps the loud internal compiler-bug throw (a true pipeline invariant).

The advised remedy is exactly what the shipped lunch-poll card does (`const activeProfile = profile ?? profileWish.result;` hoisted above `joinAs`), and it compiles clean.

## Tests & docs

- **End-to-end** (`test/builder-argument-computation-diagnostic.test.ts`): `??`-over-wish and comparison forms produce exactly one diagnostic without crashing; the hoisted form produces zero.
- **Unit** (`test/expression-rewrite-compute-wrap-coverage.test.ts`): report-vs-throw pinned on both sides of the authored/synthetic line, plus the `assertContainer`-spanning helper-arg bail path.
- **Golden fixtures** pin the three shapes reported alongside the crash as supported lowerings (body-alias ternary condition, map-row comparison against a body computed, JSX-attribute `??` binding — none of them were the actual crasher).
- **Runnable repros** in `test/fixtures/bug-repro/compute-wrap-shape3*.tsx` (`deno task cf check <file> --no-run`): two produce the diagnostic, the hoisted one compiles clean.
- Spec: §6.10 diagnostic entry + §7.2 guard note in `ts_transformers_current_behavior_spec.md`; critique guide gains a Reactivity-table row.

## Verification

- Full ts-transformers suite: **1118 passed / 0 failed**; package fmt/lint/check clean; `check-docs` 518 blocks pass.
- `cf check --no-run`: 3 golden inputs + hoisted repro exit 0; both crash repros exit 1 with the hoist diagnostic and zero `Internal Common Fabric` lines.
- Coverage: `compute-wrap-invariants.ts` at 100% lines; the only new uncovered lines (~6) are the ternary-branch and initializer bail arms, which are structurally unreachable today (the boundary walk stops at ternary/statement nodes) but are the correct handling of the guard's boolean contract — the desired failure mode if a future topology reaches them is diagnostic-not-crash.

---
<a href="https://app.blacksmith.sh/commontoolsinc/codesmith/labs/pr/4972"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787459392&installation_model_id=428580&pr_number=4972&repository=commontoolsinc%2Flabs&return_to=https%3A%2F%2Fgithub.com%2Fcommontoolsinc%2Flabs%2Fpull%2F4972&signature=1413ecc3dd78ad0c1c640eeded030903330fa757241afd3893a894c67cbd7cdb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes a compiler crash when a reactive computation is written inline in a builder/bound call’s arguments by emitting a clear hoist diagnostic instead. Code like `join({ profile: profile ?? profileWish.result })` now reports `reactive:call-argument-computation` instead of an internal error.

- **Bug Fixes**
  - Introduced `resolveComputeWrapCandidate` (replacing `assertValidComputeWrapCandidate`) to return an explicit verdict; reports `reactive:call-argument-computation` and skips wrapping for authored computations inside owned boundaries, still throws for synthetic/already-compute cases.
  - Updated emitters (binary, ternary branch, helper-owned, pattern-body initializer) to honor the verdict and bail when needed.
  - Deduped authored-source detection via `hasAuthoredSourceSite` in `ast/utils` and reused it across site policy and compute-wrap invariants.
  - Added E2E/unit tests and golden fixtures; updated the spec and the pattern critique guide.

- **Migration**
  - Hoist reactive computations out of builder/owned call args. Example: `const p = profile ?? profileWish.result; join({ profile: p })`.

<sup>Written for commit 2478baf5bd4e2afb81a1196fc0dca0d11b1d59fa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4972?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mathpirate:
Pushed a review-follow-up commit (`2925fce`) — no behavior change, both points from a self-review pass:

1. **Deduped the authored-source check.** The first cut hand-copied `hasAuthoredSourceSite` into `compute-wrap-invariants.ts` to dodge a `expression-site-policy → compute-wrap-invariants` import cycle. That was a smell: both modules already depend on `ast/mod.ts`, so the predicate now lives once in `src/ast/utils.ts` and both import it — no cycle, one source of truth for the "is this authored?" test the fix hinges on.

2. **Made the guard's verdict explicit.** `validateComputeWrapCandidate` returned a `boolean` that overloaded three outcomes (wrap / already-reported-so-bail / throw), which forced every one of the four call sites to carry a "Reported as an author-facing diagnostic" comment to stay legible. Renamed to `resolveComputeWrapCandidate` returning `{ kind: "wrap" } | { kind: "skip-reported" }` (the compiler-bug shapes still throw); the crutch comments are gone.

Full ts-transformers suite green (1118), `cf check` confirms the hoist diagnostic and the hoisted remedy are unchanged.

**Deferred follow-up:** the policy classifier still marks builder-arg computations as lowerable while the emitter categorically refuses them — this PR catches that divergence late at the emitter guard rather than dissolving it in the classifier. Keeping the authoritative wrappability check at the emitter (on the rewritten tree) is the more conservative choice for a crash-to-diagnostic fix; dissolving it in policy has a wider blast radius. Tracked as CT-1906.

packages/ts-transformers/src/ast/utils.ts
```diff
@@ -50,6 +50,24 @@ export function getExpressionText(expr: ts.Expression): string {
...
 }
 
+/**
+ * True when the node — or, if it was cloned/replaced during transformation, its
+ * original — carries a real source range: i.e. it is authored pattern code the
+ * author can act on, not a compiler-synthesized node. Consulted where a lowering
+ * disagreement must be routed either to an author-facing diagnostic (authored
+ * node) or a pipeline-invariant throw (synthetic node).
+ */
+export function hasAuthoredSourceSite(node: ts.Node): boolean {
```
mathpirate, line 76: this functionality almost certainly already exists elsewhere in transformers, since it's a question we often want to ask/answer. please do an audit of how the transformers package as a whole currently handles/treats this question. if we've really never factored this check out into its own function, we should take the time to understand the various forms of this question that are being posed across the module today and figure out how to centralize a good design for it.

packages/ts-transformers/test/fixtures/bug-repro/compute-wrap-shape3c-builder-arg-hoisted.tsx
```diff
@@ -0,0 +1,51 @@
+// SHAPE 3c: the remedied form of shape 3 — the reactive selection hoisted to a
```
mathpirate: if these are fixtures they should be in .input and .expected pairs right?

packages/ts-transformers/src/ast/utils.ts
mathpirate: Audited — you were nearly right: the question appears ~15× across 11 files, but it had never been factored out except one module-local `isSyntheticTypeNode` in type-shrinking. The audit found it's actually TWO opposite questions: **provenance** ("did an author write this text?" — looks through `getOriginalNode`; consumed by diagnostic routing) vs **identity** ("did a factory create this node object?" — deliberately doesn't look through originals; consumed by idempotency/ownership gates like "don't reprocess helper calls a pass emitted"). Conflating them is the real hazard in either direction: provenance-where-identity-is-meant re-enters subtrees a pass owns; the reverse silently drops diagnostics for cloned authored code.

`6794472` names the identity predicate (`isSyntheticNode`), documents the contrast on both predicates in `src/ast/utils.ts`, and migrates the ten identity-question sites — including deleting type-shrinking's local helper (full suite green, behavior-preserving; `pos`-only spellings normalize to `pos < 0 || end < 0`, a distinction unreachable in practice since factories and setTextRange set both ends together).

Deliberately NOT migrated, flagged for follow-up: (a) schema-generator's `pos === -1 && end === -1 && TypeFlags.Any` compound — it couples full-syntheticness with type resolution, so loosening it is a behavior question; (b) the third family the audit surfaced — three hand-rolled "resolve a real range/node" walks (`context.resolveDiagnosticRange`, pattern-coverage's sourceMapRange-fallback variant, and pattern-body's `resolveDiagnosticNode` plus its own dedup that parallels `reportDiagnosticOnce`). That consolidation has genuine behavior risk and is being filed as its own Topic rather than growing this PR.

packages/ts-transformers/test/fixtures/bug-repro/compute-wrap-shape3c-builder-arg-hoisted.tsx
mathpirate: Right — promoted in `7535560`. The rule that fell out of doing it: a shape belongs in `.input`/`.expected` golden pairs exactly when it **compiles cleanly**; it can't be a golden when it produces an error diagnostic, because the fixture runner fails any fixture whose transform reports one. So the two clean-compiling repros moved to `ast-transform/` as real pairs (`builder-arg-hoisted-nullish-selection` — which also golden-pins the authored-name cause chain `.for("activeProfile")` → `.for(["boundJoin","profile"])` — and `builder-arg-ternary-label`, pinning the inline-ifElse label lowering), while the diagnostic-producing inline form stays in `bug-repro/` with its contract pinned by `test/builder-argument-computation-diagnostic.test.ts`. `bug-repro/README.md` now states the directory's contract, and `fixtures/README.md` points at it.

PR #4974, opened 2026-07-24
feat(shell): dogfood toggle to enable concurrent watch refresh (#4937 follow-up)

bfollington:
## Why

[#4937](https://github.com/commontoolsinc/labs/pull/4937) landed `experimentalConcurrentWatchRefresh` **default-off** and documented it as "pending live measurement on a real (estuary-latency) load" — but it wired **no way to turn it on** in a deployed build. The browser worker always opens storage with `defaultSettings` (`runtime-processor.ts`), and unlike `cfcRenderCeiling` the flag was never added to the shell's host-toggle seam. So the flag whose entire purpose is a production A/B had no switch: deploying it to estuary would profile byte-identical to the baseline HAR.

This adds the missing enable path so we can actually measure the win.

## What Changed

A per-browser-profile dogfood toggle, following the exact `cfcRenderCeiling` pattern:

- **`commonfabric.concurrentWatchRefresh(enabled?)`** console command, localStorage-backed (`packages/shell/src/lib/concurrent-watch-refresh.ts`), registered in `host-toggles.ts`.
- Threaded shell → worker: `runtimeHostFlags()` → `RuntimeInternals.create` → `createRuntimeClientOptions` → `RuntimeClient.initialize` → `InitializationData` → `runtime-processor`'s `StorageManager.open({ settings: { ...defaultSettings, experimentalConcurrentWatchRefresh } })`.

Like the render ceiling, the setting is fixed at `StorageManager.open` time (crosses the worker IPC in `InitializationData`), so it **takes effect on the next runtime (reload)**, not live. Default off — flipping it is a deliberate per-profile action.

## How to use (the profiling procedure)

1. Deploy this SHA to estuary.
2. Open the shell, DevTools console: `commonfabric.concurrentWatchRefresh(true)`, reload. Capture a HAR of the cold load.
3. `commonfabric.concurrentWatchRefresh(false)`, reload, capture the baseline HAR — same build, clean A/B.
4. Compare `session.watch.add` overlap / cold-load wall-clock.

## Test plan

```
deno test -A packages/shell/test/concurrent-watch-refresh.test.ts \
  packages/shell/test/host-toggles.test.ts \
  packages/runtime-client/client/initialize-init-data.test.ts \
  packages/lib-shell/test/runtime.test.ts
```

- New `concurrent-watch-refresh.test.ts` mirrors `render-ceiling.test.ts`: default-off, `"true"`-only truthiness, silent-while-disabled, ON hint, persist + reload-notice, clear-on-disable, fail-closed on storage errors.
- `host-toggles.test.ts` updated: command installed + `runtimeHostFlags()` includes the new flag across the persisted-state matrix.
- **`initialize-init-data.test.ts`**: new guard that `concurrentWatchRefresh: true` survives `RuntimeClient.initialize`'s hand-built `InitializationData` literal into the worker payload — mirrors the existing `patternCoverage` guard and proves the flag isn't silently dropped at the IPC boundary.
- Full CI parity verified locally: `deno task check` (the type gate) → "Type check complete."; `deno fmt --check` + `deno lint` clean on all touched files; `lib-shell/test/runtime.test.ts` (35 steps) still green.

### DIFF
--- a/docs/development/EXPERIMENTAL_OPTIONS.md
+++ b/docs/development/EXPERIMENTAL_OPTIONS.md
@@ -611,8 +611,15 @@ the per-epic implementation notes).
   passed through `StorageManager` settings. The runner mirrors it onto each
   memory session via `SpaceSession.setConcurrentWatchRefresh()`
   ([`packages/memory/v2/client.ts`](../../packages/memory/v2/client.ts)) —
-  per-session, not a process global.
-- **Added by.** Ben Follington (#4937).
+  per-session, not a process global. In the **shell** it is a per-browser-profile
+  dogfood toggle: run `commonfabric.concurrentWatchRefresh(true)` in the console
+  and reload. The flag crosses the worker IPC in `InitializationData` and is
+  fixed at `StorageManager.open` time, so — like the render ceiling — it takes
+  effect on the next runtime (reload), not live. Threaded shell → worker via
+  `runtimeHostFlags()`
+  ([`packages/shell/src/lib/host-toggles.ts`](../../packages/shell/src/lib/host-toggles.ts))
+  → `RuntimeInternals.create` → `runtime-processor.ts`'s storage settings.
+- **Added by.** Ben Follington (#4937; shell dogfood toggle in a follow-up PR).
 - **Purpose.** By default watch acquisition is strict single-flight per space: a
   guard holds every watch refresh after the first until the prior response
   lands, so traversal-driven pulls discovered a tick apart serialize into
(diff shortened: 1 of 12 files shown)


bfollington (commented):

PR #4974 · detail review of all 10 changed files · targeted tests, type-check, format, lint, diff-check, and CI inspected.

## 🔴 Blocking

1. `packages/shell/src/globals.ts:34` — **Performance Check fails because the new type-only declaration adds one uncovered line.** Verified from the CI coverage artifact: shell debt rose from 6813 to 6814. Importing this ambient declaration cannot execute it, so another runtime test will not fix this specific failure. Accept the narrow baseline in the PR description:

   `NEW_PERF_BASELINE: coverage-debt: packages/shell uncovered lines = 6814 lines`

## 🟡 Improvements

1. `packages/runtime-client/runtime-client.ts:146` and `packages/runtime-client/backends/runtime-processor.ts:525` — **the tests do not guard the PR’s central shell→worker wiring.** The new tests verify persistence and `runtimeHostFlags()`, but the existing tests explicitly intended to catch dropped initialization fields were not updated. Add `concurrentWatchRefresh: true` and assertions to `packages/lib-shell/test/runtime.test.ts`’s “forwards host flags” test, plus coverage for mapping true/false into `StorageManager.open` settings. Verified by diff/search; the implementation itself is correct today.

2. `docs/development/EXPERIMENTAL_OPTIONS.md:46` — **the summary table still lists only the old programmatic toggle path.** The document explicitly requires updating both the section and summary table when touching a flag. Add `commonfabric.concurrentWatchRefresh()`/localStorage and identify follow-up PR #4974; the detailed section is already accurate.

**Verdict:** no correctness regression found in the implementation, but the PR is not merge-ready while Performance Check is red. All other CI jobs passed; local focused tests, type-check, format, lint, and diff-check passed.

PR #4977, opened 2026-07-24
fix(runner): heal nested system pieces (profiles) that brick on cold-start instantiate

bfollington:
## Why

After #4967 healed bricked home **roots** on Estuary and both previously-dead
homes rendered, a follow-on surfaced: interacting with a **profile** threw
`Handler used as lift: … the { "$stream": true } marker was never written
there`. That is the *same* brick as the original home-root failure, one level
down — and #4967 doesn't reach it.

A profile is a nested piece (a `ProfileHome` in its own space, mounted into the
home via a `#wish`). The runtime's start walk instantiates it **without a setup
phase**, and no pattern watcher is armed to self-heal (the watcher arms only
*after* a successful instantiate). If the profile's stored doc predates its
pattern's setup — a pre-manifest internal-cell layout, or a handler stream added
after the doc was created — its `{ "$stream": true }` markers were never
materialized, so instantiation dies. #4967's repair lives in
`startEnsuredDefaultPattern`, which only the **root** passes through; nested
pieces go straight through `runtime.start → startCore` and were left bricked.

This is a **pre-existing** condition (profile docs bricked at/near creation,
independent of #4967) that #4967 merely made *reachable* by letting homes load.
It fixes profiles whose home always loaded too.

## What Changed

`packages/runner/src/runner.ts`, in `startCore`'s initial instantiation:

- On the specific `"…marker was never written"` failure, re-run the pinned
  pattern's **own** setup (`applySetupState`, `samePattern=true`: materializes
  the missing internal cells, **leaves the piece's argument/data untouched** —
  no roll-forward, no user-data rewrite), then retry once. This reuses the exact
  setup-then-instantiate mechanism `swapToPattern` already uses for hot-swaps.
- **Gated** on the `systemPatternAutoUpdate` experimental flag.
- **Fail-closed:** a non-matching or failed repair rethrows the *original*
  error, so the piece is left exactly as it was.
- **The root/default pattern is excluded** — it is the `PieceController`'s to
  repair, with its richer roll-forward-to-official + clear-error path. A
  per-piece guard skips the repair when the cell *is* its space's
  `defaultPattern` (only the controller ever sets that; profiles never are). The
  guard runs only on the rare brick path, so a healthy start pays nothing.

Matched to the *specific* `"marker was never written"` variant, not the shared
`Handler used as lift` prefix — the sibling "not a stream reference" /
"overwritten" variants are not setup-missing and must not trigger a re-setup.

## Test plan

- New `packages/runner/test/nested-piece-setup-repair.test.ts`: a nested piece
  set up for V1 then re-pointed at a handler-bearing V3 (whose `bump` marker the
  V1 doc lacks). **Flag off** → still bricks (`marker was never written`, locks
  in the gap). **Flag on** → starts clean and the once-dead handler fires end to
  end (`bump` → `count` increments) — a functional read, not just "started".
- The full **piece heal suite** stays green **unchanged** — the exclusion guard
  keeps the runner repair from pre-empting the controller's root roll-forward /
  fail-closed behavior (the interaction this PR was careful to get right).
- Green locally: full runner suite (1018 files / 5435 steps) and piece suite
  (22 / 267), fmt + lint clean on touched files.

## Follow-up (separate PR, not here)

The cross-space `StorageTransactionWriteIsolationError` on profile events (a CFC
boundary-commit companion write spanning two spaces without a multi-space
opt-in) is a distinct bug in CFC-integrity code and gets its own focused PR.

## Coverage

ACCEPT_COVERAGE_DEBT: coverage-debt: packages/runner uncovered lines = 6978 lines

The review-driven fail-closed fixes add defensive branches whose *success* paths
are covered by the heal test, but whose *failure* paths are not faithfully
unit-testable in the current harness: the precondition-mismatch abort needs the
pinned identity to change between two synchronous reads inside one `start()`;
the setup/retry abort needs a re-setup that still can't materialize the marker;
and the commit-`{error}` / Promise-rejection teardown needs a storage-level
commit-failure injector the emulated storage doesn't expose (a mock would bypass
the very `addCommitCallback` / `.catch` path under test). Rather than fake
coverage on paths that can't be driven, these correctness branches are accepted
as debt — the scope-collision fix is independently verified by the reviewer's
own repro, and the teardown mirrors the runtime's existing
`startAfterSuccessfulCommit` pattern.

---
<a href="https://app.blacksmith.sh/commontoolsinc/codesmith/labs/pr/4977"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787470441&installation_model_id=428580&pr_number=4977&repository=commontoolsinc%2Flabs&return_to=https%3A%2F%2Fgithub.com%2Fcommontoolsinc%2Flabs%2Fpull%2F4977&signature=e1c15443d52e9238979bd564f520793c5eb6a65f0bc74a762b85464a3f2223a5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Heals nested system pieces (e.g., profiles) that brick on cold-start instantiate by re-running the piece’s setup on the specific “marker was never written” error. The fix is atomic, fail-closed, preserves user data, and uses scope-safe teardown on commit failure; gated by `systemPatternAutoUpdate`.

- **Bug Fixes**
  - On the “marker was never written” failure during nested piece instantiate, re-run the piece’s own setup (`samePattern=true`) and retry once; arguments/data are unchanged.
  - Single-transaction repair with an identity precondition; any staging failure aborts and rethrows the original error. Commit is not awaited; on commit error or rejection, perform scope-safe teardown that unregisters only this start’s cancel if it’s still current.
  - Skipped when a `useTx` is provided or when the piece is the space’s `defaultPattern` (guard compares space + scope + id).
  - Exported `isMissingStreamMarkerFailure`; tests cover only the “never written” variant and verify brick with the flag off and end-to-end heal with the flag on (`packages/runner/test/nested-piece-setup-repair.test.ts`).

<sup>Written for commit 53f0bd8b017ffeeafa83d30e774270d275d95a66. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4977?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -1495,6 +1495,42 @@ export class Runner {
     }
   }
 
+  /**
+   * True when `resultCell` is its space's default/root pattern — the piece the
+   * PieceController's own cold-start repair (startEnsuredDefaultPattern) owns,
+   * including its roll-forward-to-official backstop and clear-error contract.
+   * The runner's initial-start setup repair must DEFER to the controller for
+   * the root and heal only the nested pieces the controller never sees (a
+   * profile mounted via a #wish, say). Profiles are plain `inSpace` pieces and
+   * are never a space's `defaultPattern` (only the controller sets that), so
+   * they are correctly not excluded. Called only on the rare brick path, so the
+   * space-cell read costs nothing on a healthy start. A read failure returns
+   * false: better to attempt the idempotent, fail-closed repair than to leave a
+   * piece bricked because a lookup raced.
+   */
+  private isSpaceDefaultPattern(resultCell: Cell<unknown>): boolean {
+    try {
+      const defaultPatternCell = this.runtime
+        .getSpaceCell(resultCell.space)
+        .key("defaultPattern")
+        .get() as Cell<unknown> | undefined;
+      if (defaultPatternCell === undefined) return false;
+      const a = resultCell.getAsNormalizedFullLink();
+      const b = defaultPatternCell.getAsNormalizedFullLink();
+      // Full document identity: space + scope + id. `scope` (space/user/session)
+      // is part of the address — a user- or session-scoped nested cell can share
+      // an entity id with the space-scoped root, so omitting scope would
+      // misclassify it as the root and silently suppress its heal. `path` is
+      // intentionally not compared: doStart normalizes a subpath input to its
+      // root before startCore, so resultCell is always a root cell here.
+      return a.space === b.space &&
+        (a.scope ?? "space") === (b.scope ?? "space") &&
+        a.id === b.id;
+    } catch {
+      return false;
+    }
+  }
+
   /** Convert a module to pattern format */
   private moduleToPattern(module: Module): Pattern {
     const resultSchema = module.resultSchema ?? {};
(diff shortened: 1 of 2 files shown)


bfollington:

Scope: PR #4977 · deep-read of the two changed files plus `PieceController.startEnsuredDefaultPattern`, `applySetupState`, start/resume, scheduler registration, and v2 transaction commit paths. Checks: new runner test passed; the 57-step `checkAndUpdateDefaultPattern` suite passed; focused scope-collision repro failed as described below. CI's Performance Check is currently failing the coverage-debt gate (+10 uncovered runner lines).

## 🔴 Blocking

### 1. Major — the repair is not transactionally sequenced or fail-closed

`packages/runner/src/runner.ts:1647-1651, 1794-1814`

The first `instantiatePattern()` commits its `actualTx` from `finally` even when node instantiation throws. `cancelNodes()` only removes scheduler registrations; it cannot undo writes already staged by earlier nodes. In particular, a pattern node encountered before the bad handler calls `run(tx, ...)` and can stage nested-piece setup into this transaction.

The repair then opens a second transaction, calls `setupTx.commit()` without awaiting or inspecting its `{ error }`, and immediately retries in a third transaction. `commit()` is asynchronous (`IExtendedStorageTransaction.commit(): Promise<Result<...>>`), and storage/CFC failures are normally returned in that result, not thrown synchronously. Therefore the `try/catch` does not implement the stated contract:

- a rejected/conflicted setup commit can be ignored while `start()` reports success from optimistic local state;
- retry races the setup commit rather than running after confirmed durability;
- if retry throws, its new error is propagated, not `instantiateError`;
- because setup is committed before retry, a retry failure can leave the document changed;
- the failed first instantiate can already have committed partial nested setup.

Concrete fix: make the repair an awaited path and keep setup + retry in one repair transaction: apply setup, instantiate against that same supplied tx, prepare, commit once, await/inspect the result, and cancel/abort on every failure. The original failed-instantiation transaction must abort rather than commit when instantiation throws. Any setup, commit, or retry failure should clean up and rethrow the original error.

### 2. Major — forced `samePattern=true` can roll back a concurrent identity update

`packages/runner/src/runner.ts:1797-1808`

`initialRef` is read before the failed instantiate, but the repair transaction never re-reads/asserts that `patternIdentity` still equals that ref. `applySetupState(..., ref, true, ...)` unconditionally writes `ref` back. If the updater or another boot moves the piece between those points, a repair transaction created after that move can overwrite the newer identity and projection with the stale pinned pattern.

Use the same expected-identity precondition that the controller repair already relies on: read `patternIdentity` through `resultCell.withTx(setupTx)`, require both identity and symbol to equal `ref`, and let that read participate in commit conflict detection. On mismatch/failure, surface the original instantiate error without writing.

This also narrows the `samePattern` claim. For an actually unchanged ref, the existing argument is not rewritten and existing derived-cell values are not default-seeded over. But setup still rewrites the internal manifest, result projection, ownership/meta, and `patternIdentity`; “leaves the piece's data untouched” is not a general guarantee without the identity precondition.

### 3. Major — the default-pattern guard ignores cell scope

`packages/runner/src/runner.ts:1491-1501`

The comparison uses only `space` and `id`. Scope is part of document identity (`space/user/session`), so a valid nested user- or session-scoped cell with the same entity id as the space-scoped default root is classified as the root and its heal is silently suppressed.

I reproduced this with two cells having the same space/cause and different scopes: after linking the space-scoped cell as `defaultPattern`, `isSpaceDefaultPattern(userScoped)` returns `true`.

Compare full document identity, including `scope` (prefer the canonical resolved-cell equality already used by `pattern-updater.ts`). Ignoring `path` is safe only because `doStart()` normalizes a subpath input to its root before `startCore`; preserve that explicit invariant in the comparison/test.

## 🟡 Improvements

- `isMissingStreamMarkerFailure()` does exclude the two current sibling diagnostics (“is not a stream reference” and “was overwritten”), so the selected variant is semantically the right one. The string match is nevertheless brittle and unbranded. This producer and consumer live in the same module; use a dedicated error class/discriminant and add flag-on negative tests for both sibling variants.
- The new “cold-start” test stops and restarts within the same runtime. It does not cover a fresh runtime/cache, an unsynced space cell, setup commit `{ error }`/rejection, retry failure, concurrent identity movement, scope/subpath identity, or preservation of a non-default user value. Those are the load-bearing contracts of this change, and the current coverage job confirms 11 new runner lines remain uncovered.
- This expands `systemPatternAutoUpdate` from source reconciliation to setup repair for every keyed non-default piece, not only same-toolshed system-source patterns. Update `docs/development/EXPERIMENTAL_OPTIONS.md` and the live pattern-update spec, or use a more precisely scoped admission signal.

The normal controller-owned root path did remain delegated in the existing suite, and the current `ProfileHome.inSpace()` shape (space-scoped, with no `defaultPattern` link in its child space) is not suppressed by the guard. The blockers are the generic/runtime guarantees claimed by the implementation.

**Verdict: request changes before merge.**

bfollington:
**Re-review response** (commit `345334117`)

The re-review confirmed blockers **2 (identity precondition)** and **3 (scope-aware root guard)** are fully resolved, and asked for three more fixes to blocker 1's commit-failure handling. All three are now applied:

1. **`cleanup()` instead of `cancel()`** on commit failure. `cancel()` tears down the node graph but deliberately leaves the entry in `this.cancels`/`allCancels`, so a later `start()` would take the "already started" fast path and falsely report the dead piece as running. `cleanup()` removes those registrations, so a failed-commit piece re-heals on the next start instead of being falsely-started until a runtime reboot.
2. **Precondition read + `prepareTxForCommit` moved inside the abort/rethrow boundary.** A throw in either now aborts the repair tx and rethrows the *original* instantiate error, per the stated contract.
3. **`.catch()` on the unawaited commit Promise** that also cleans up — `addCommitCallback` handles the committed-with-`{error}` result, `.catch` handles a rejected commit Promise. The comment now describes this honestly as a **speculative start followed by teardown on commit failure**, not synchronous error propagation.

Green: full runner suite (1019 files / 5439 steps), piece suite (22 / 267), fmt + lint clean; coverage gate passing.

**Two deliberate trade-offs, flagged for the record (not silently dropped):**
- **First failed instantiate still self-commits partial writes** — you called this acceptable debt rather than a blocker; the successful repair supersedes those writes, and aborting the first attempt would require taking over the tx for *every* flag-on nested-piece start on the hot path. Kept as a focused follow-up.
- **The two commit-failure branches (result-`{error}` and Promise-rejection) are inspection-only.** Faithfully unit-testing them needs a storage-level commit-failure injector the emulated storage doesn't expose; a mock that stubs `commit` would bypass the very `addCommitCallback`/Promise path under test, so it'd be fake coverage. The logic mirrors the runtime's own `startAfterSuccessfulCommit` teardown. Happy to add real tests if a commit-failure injector lands.

PR #4985, opened 2026-07-24
fix(runner): retry a reactive recompute off-budget on a same-replica …

Hixie:
…race

A reactive compute's commit can fail with StorageTransactionInconsistent: a value the transaction read changed on this replica between the read and the commit. Like a ConflictError, this is a stale-basis rejection that converges by re-running against the settled replica. The reactive commit-result path in the scheduler already handles a ConflictError that way — it re-arms the subscription and re-queues the action off the retry budget, because a stale basis is a wait for catch-up, not a failure — but it did not extend the same treatment to the local same-replica race. StorageTransactionInconsistent instead fell into the bounded MAX_RETRIES_FOR_REACTIVE path, alongside genuinely terminal transport and malformed-store errors.

The consequence is a permanent strand. Under a burst of same-replica races longer than the budget, the recompute exhausts its retries and gives up, leaving its output at the pre-burst value. Nothing re-triggers it once its inputs stop changing, so the stale value persists. In a multi-session setup this shows as a second session's per-session derived view — its own recompute of shared state — that never advances to reflect a write it observed, while the writer's own view is correct.

Treat StorageTransactionInconsistent like a ConflictError in the reactive commit-result path: re-arm and re-queue without charging the retry budget, so the recompute rides out a burst of any length and converges once it clears. This mirrors the event-handler commit path, which already windows the same local stale-basis guard. Only a rejection that re-running cannot resolve keeps the bounded budget.

Pinned by two tests: a scheduler-retries unit assertion that a same-replica race re-queues off the budget (contrasting a generic transient, which charges it), and a two-runtime integration test in which a per-session reactive recompute, its commits failed with a StorageTransactionInconsistent burst longer than the budget, must still converge to the fresh value instead of stranding. The second fails against the pre-fix scheduler. The rejection.ts comment describing the reactive path's retry classification is updated to match.

---
<a href="https://app.blacksmith.sh/commontoolsinc/codesmith/labs/pr/4985"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787481810&installation_model_id=428580&pr_number=4985&repository=commontoolsinc%2Flabs&return_to=https%3A%2F%2Fgithub.com%2Fcommontoolsinc%2Flabs%2Fpull%2F4985&signature=dd21b6464cbe9b3606b9e7ca70075a5998796983359cc1bd966dec5a70bff256"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Retries reactive recomputes off-budget on same-replica race bursts to prevent strands. Aligns `StorageTransactionInconsistent` with conflicts, and adds a periodic diagnostic if off-budget retries never converge.

- **Bug Fixes**
  - Treat `StorageTransactionInconsistent` as stale-basis in the reactive commit path: restore trigger reads, resubscribe, and re-queue without charging `MAX_RETRIES_FOR_REACTIVE`; clear on success or permanent/terminal rejection.
  - Count off-budget retry streaks per action and log a non-fatal diagnostic every `OFF_BUDGET_RETRY_WARN_INTERVAL` (100). Added unit tests for classification and streak accumulate/clear, plus a two-runtime integration test verifying convergence past a burst longer than the budget.
  - Plumb `offBudgetRetries` through the scheduler and update test harnesses to pass `offBudgetRetries` and `getActionId` to `watchReactiveActionCommit`.

<sup>Written for commit a18ddd852e5417ce336425d4ad245a93859b8d93. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/4985?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/scheduler/constants.ts
+++ b/packages/runner/src/scheduler/constants.ts
@@ -38,6 +38,15 @@ export const MAX_ACTION_RUN_TRACE_HISTORY = 2000;
 // above any normal burst, so it never changes ordinary delivery.
 export const MAX_EVENT_BACKLOG_PER_STREAM = 256;
 export const MAX_RETRIES_FOR_REACTIVE = 10;
+// A stale-basis rejection (conflict / same-replica race) is retried off the
+// bounded budget, on the assumption that the action's subscription will
+// eventually deliver the value it is waiting on. That holds for pattern-created
+// reactive functions, which go through the cell machinery, but a bug that never
+// closes the loop — historically a serialization round-trip that fails to
+// preserve a value (a stray `-0`, say) — would spin here forever. Emit a
+// non-fatal diagnostic every this-many off-budget re-queues of one action so the
+// loop is visible rather than silent.
+export const OFF_BUDGET_RETRY_WARN_INTERVAL = 100;
 export const AUTO_DEBOUNCE_THRESHOLD_MS = 50;
 export const AUTO_DEBOUNCE_MIN_RUNS = 3;
 export const AUTO_DEBOUNCE_DELAY_MS = 100;
--- a/packages/runner/src/scheduler/facade.ts
+++ b/packages/runner/src/scheduler/facade.ts
@@ -425,6 +425,7 @@ export class Scheduler {
   private triggerIndex = new SchedulerTriggerIndex();
   private actionChangeGroups = new WeakMap<Action, ChangeGroup>();
   private retries = new WeakMap<Action, number>();
+  private offBudgetRetries = new WeakMap<Action, number>();
 
   // Effect/computation tracking for pull-based scheduling
   private nodes = new NodeRegistry();
@@ -2499,6 +2500,7 @@ export class Scheduler {
       actionChangeGroups: this.actionChangeGroups,
       actionTimingState: this.actionTimingState,
       retries: this.retries,
+      offBudgetRetries: this.offBudgetRetries,
       pending: this.pending,
       actionRunTrace: this.actionRunTrace,
       nodes: this.nodes,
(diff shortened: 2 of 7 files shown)


seefeldb:
> it re-arms the subscription

it does? can you elaborate on that? (my assumption is that instead the subscription is just already active)


this change looks good to me, but i want to flag a risk (why this exhaustion exists in the first place): this assumes that there is a subscription that will eventually deliver the value. that should always be the case for pattern-created reactive functions since they go through the cell machinery which will ensure that there is one. but in the past we had bugs, mostly serialization-deserialization not correctly round tripping (a stray `-0` or so), that meant this loop was never closed, and so it looped forever. dan's work closed a lot of those, but just in case we run into more:

maybe after a high number (100?) of retries logger.error about this, so it's easier to tell?

Hixie:
> > it re-arms the subscription
> 
> it does? can you elaborate on that? (my assumption is that instead the subscription is just already active)

You're right. Claude backed away from that claim when I asked it about it.

Added the suggested logging.

PR #5001, opened 2026-07-24
fix(runner): value-compare the result projection, and refuse a `FabricInstance` binding

danfuzz:
Two fixes on the path from a pattern's result to storage, both about values
that are of JS type `object` but are neither a plain object nor an array.

#5074 landed while this was open and took over the third — the
`FabricPrimitive` ablation that used to mask everything here. What remains is
the pair that PR deliberately did not take on.

## 1. The no-op write gate compared by structure

The gate in `updateResultProjection` (`packages/runner/src/runner.ts`) decides
whether a freshly-bound result differs from what is already stored, and skips
the write when it does not. It compared with `deepEqual`, which walks
enumerable own properties — and a `FabricPrimitive` keeps its state in private
`#fields` and exposes none. Two distinct same-class primitives therefore
compared **equal**, and the write was dropped: a result whose bytes changed
left the stale bytes in storage.

It now compares with `valueEqual`, which is what the formal spec already says
this comparison is. `docs/specs/space-model-formal-spec/1-fabric-values.md`
§6.7 names `valueEqual` as "the equality the reactive system's change-detection
and **no-op gates** depend on". No documentation changes here: the spec was
already right, and the code was what stood out of line with it.

### The conversion has to come first

`fabricFromNativeValue()` now runs *above* the gate rather than inside it, and
that order is load-bearing rather than incidental. `valueEqual` hashes its
operands, and a raw pattern result is not necessarily hashable — one carrying
`toJSON` becomes a `FabricValue` only at that conversion — so comparing before
converting throws `hashOf: unsupported object type: Object` instead of deciding
anything. Converting first also means the gate compares what a write would
actually store, since the stored side is already a `FabricValue`.

## 2. A `FabricInstance` binding now throws instead of passing through

`convert()` in `pattern-binding.ts` cannot descend a `FabricInstance`. It is a
CONTAINER of other `FabricValue`s, reached by its codec contents rather than by
property name, and this walk has no machinery for that. It has two available
dispositions and **neither is correct**:

- Fall through to the record branch. Its zero enumerable own properties mean
  nothing rebinds, so the original comes back whole. That reads as success
  while leaving any bound alias in its contents silently unbound.
- Throw, naming the class and the work it needs.

Between a silent wrong answer and a loud refusal, this takes the loud one.

To be clear about the history, since the diff makes it look like a reversal:
#5074 has the pass-through, but that was **scope hygiene, not a decision**.
That PR was about structure sharing, and its iteration concerned things that
had nothing to do with `FabricInstance`; letting one fall through kept the
silent-wrong-vs-throw debate from sidetracking it. This is the PR that owns
that debate, so it settles it here.

### It is meant to fire, and that is the point

The throw is a **discovery instrument**, not merely a guard that happens to be
safe today. With `modernCellRep` on, a link is a `FabricLink` and hence a
`FabricInstance`, so ordinary links reach this walk and throw. That is expected
and wanted: these `FabricInstance` throws — this one and its counterparts
elsewhere in the system — are how the work still owed before that flag can be
turned on gets located, at the sites that owe it.

**The invariant that keeps this safe rather than merely lucky:** anything
introduced that would reach one of these throws is itself gated on an experiment
flag. A default configuration therefore never arrives here, and any arrival is
something a flag was deliberately turned on to reach. That is a property of how
the system takes on this kind of change, not an observation about today's call
graph — which is what makes a permanent tripwire the right shape here rather
than a time bomb.

That rule cross-cuts every check of this shape, so this PR records it in
`docs/development/EXPERIMENTAL_OPTIONS.md` — the central flag registry — as a
new **"Flag-gated tripwires"** section, stating both obligations it creates:
gate a change that would otherwise reach one of these throws in a default
configuration, and answer a throw you meet by implementing the handling it names
rather than exempting the value.

The site comment in `pattern-binding.ts` deliberately says none of this. It
carries the local what-and-why — a `FabricInstance` is a container this walk
cannot descend, and why throwing beats handing one back — plus a `TODO`, which
is the same shape every other site of this kind uses. The flag lifecycle is
general, so it is documented once, centrally, instead of at each site.

So the reading to head off: a reviewer who probes with `modernCellRep` on will
find that ordinary sigil links throw here, and it can look like a landmine
across the in-flight cell-rep work. It is the opposite — see the registry
section.

With the flag **off** — every configuration CI runs today — the throw is inert.
Measured rather than assumed: instrumented `convert()` on `origin/main` to count
every `FabricInstance` arrival by class, then ran the whole runner package suite
— **1112 tests, 5835 steps, ZERO arrivals**. Note the scope of that claim
honestly: the suite runs flag-off, so it is evidence about the flag-off path
only and could not have observed the flag-on case either way. The invariant
above, not this measurement, is what carries the claim forward.

That matches what the concrete classes are. Of `FabricError`, `FabricLink`,
`FabricMap`, `FabricSet`, `ProblematicValue`, and `UnknownValue`: the map/set
pair's protocol methods are not yet implemented, the explicit-tag pair has no
`[CODEC]`, and `FabricLink` is flag-gated — leaving `FabricError` as the only
one that can arrive with the flag off.

It can, and that is a use of the tripwire rather than an exception to it.
`FabricError` is exposed to pattern authors (`builder/factory.ts`), so a result
carrying one stores today and throws here instead — but it has no production use
yet, and is currently a development tool for exercising exactly this: the
`FabricInstance` handling these throws are there to locate. Nothing in the suite
constructs one on this path, which is why the count above is zero.

The docblock's `TODO(danfuzz): Latent` marker for this becomes a plain `TODO`:
the gap stops being latent once the code reports it. The pointer to the two
sibling walks that still carry `Latent` markers stays.

## What #5074 took over

The bug that masked all of this: `convert()` rebuilt every record it walked,
and `isRecord()` is just `typeof value === "object" && value !== null`, so a
`FabricPrimitive` landed in that branch and the rebuild reduced to
`Object.fromEntries([])`. **Every such value in a pattern result was replaced
by a bare `{}`** — class, identity, and contents alike. That is why the gate
bug sat marked-but-undemonstrated: with values ablated to `{}` before they
reached the gate, its mishandling of them could not be observed.

#5074 leafs a `FabricPrimitive` out ahead of the container branches and pins it
in `pattern-binding.test.ts`, so nothing is owed here for it. The end-to-end
copy of that test that used to live in this branch went with it — it would pass
on `main` either way.

## Red-green evidence

Every test here was ablation-checked against a real alternative rather than a
reconstruction.

The gate, in `result-projection-value-equality.test.ts`:

| gate | bytes-change test | ordering test |
| --- | --- | --- |
| `deepEqual`, convert inside (real `origin/main`) | **FAIL** | ok |
| `valueEqual`, convert *after* the gate | ok | **FAIL** |
| this PR | ok | ok |

The two rows fail for different reasons, which is what makes them two tests.
Row 1 is the bug: bytes change, `deepEqual` calls the results identical, the
write is dropped, storage keeps `[1, 2, 3]`. Row 2 is the ordering constraint,
and it fails loudly — `hashOf: unsupported object type: Object` out of
`value-hash.ts` — rather than silently deciding wrong. Row 2 is not a guard
against `main`; it guards against a plausible *other* implementation of this
same fix, which is the shape a regression test for an ordering constraint
takes.

The throw, in `pattern-binding.test.ts` (covering the root, in a record, in an
array, and the copy path where a sibling rebinds):

| `convert()` | FabricInstance test | FabricPrimitive test |
| --- | --- | --- |
| pass-through (real `origin/main`) | **FAIL** | ok |
| this PR | ok | ok |

The second column is what makes it a targeted test rather than a broad one: the
ablation moves one and leaves the other alone.

## Where the remaining hazard sits

`pattern-binding.ts` carries two `TODO(danfuzz): Latent` markers for the same
defect class — a walk that descends by property name into values whose state is
not in named properties — in `sendValueToBindingInner` and
`findAllWriteRedirectCells`. Each reasons that "schemas don't admit `Fabric*`
values on this path today, but will in the not-too-distant future".

`convert()` is what makes those worth re-reading: it went unmarked, and it was
not latent — a pattern result carries a `FabricBytes` today, which is what these
tests demonstrate. So a follow-up sweep should ask not only "which sites were
missed" but "which sites were classified latent on an assumption that does not
hold".

## On whether fabric values belong in bindings at all

@ubik2 raised (2026-07-25) that fabric values arguably should not appear in
pattern bindings, and that a primitive pattern `result` may not be meaningful.
@danfuzz answered (2026-07-27): this fixes one of the many existing spots that
prevent `FabricValue`s from being storable and usable for calculation by pieces
— it takes a place that *de facto* can already receive any `FabricValue` and
stops it doing the wrong thing for the values that are of JS type `object` but
are neither a plain object nor an array. Of the two direct subclasses of
`FabricSpecialObject`, `FabricPrimitive` is straightforwardly addressable today
(@mathpirate has shepherded a run of these fixes in), while `FabricInstance`
still needs infrastructure support before its use sites can be made correct.

That asymmetry is the design rather than a hedge, and it is exactly why the two
arms here differ: a primitive passes through, an instance throws and names what
it needs.

## Testing

- Full runner package suite green.
- `deno fmt --check`, `deno lint`, and `deno check` of all changed files clean.

## CI override

The gate reports `packages/runner` 5588 → 5589. **That line is not from this
change**, and it is not new code going untested — it is an untouched line in an
untouched file whose coverage is contingent on test co-scheduling.

Attributed by LCOV diff (both runs' artifacts, shards merged by max hits,
diffed per file). Exactly one file moves:

| | main `b5cf4ea3` | this PR | delta |
| --- | ---: | ---: | ---: |
| `packages/runner` total | 5342 | 5343 | **+1** |
| `src/scheduler/facade.ts` | 110 | 111 | **+1** |
| every other file | — | — | 0 |

The line is `facade.ts:237`, `if (crossesSpace && rank === 0) return 2;` in
`observationMinimumContextRank`. The files this PR touches show a delta of
**zero**, and the lines it adds are covered by the tests above.

**Why it moved.** The runner suite is split into five shards by
`tasks/select-runner-test-files.ts`, which partitions the test-file list. This
PR adds one test file, so the partition shifts — **28 of shard 4's 106 files
change**. The line's sole cover is an incidental one from
`reload-rehydration-safety.test.ts`, and whether that test reaches it depends on
which files it runs alongside.

Measured rather than asserted, and this is the part that rules out a semantic
cause:

| context | main | this PR |
| --- | ---: | ---: |
| that test alone | 2 | **2** |
| a 26-file subset (3 reps each) | 2 | **2** |
| the full shard, as CI partitions it (3 reps) | 2 | 0 |

Covered on both branches wherever the file set is held equal; it diverges only
where the partition itself differs. Reverting the gate fix while keeping the
throw leaves it at 0, so neither half of this change is what moves it.

I did **not** chase this into the scheduler. Covering `facade.ts:237`
deliberately would mean a new snapshot-rehydration test in a subsystem this PR
does not touch — worth doing, but as its own change, not smuggled in here.
Happy to write it separately if preferred over the override.

(An earlier revision of this section attributed a +1 to `scheduler/events.ts`
against a different head. The measurement above supersedes it — and that this
PR has now tripped the same artifact twice, in two different scheduler files,
is itself a symptom of the sharding sensitivity rather than of the change.)

ACCEPT_COVERAGE_DEBT: coverage-debt: packages/runner uncovered lines = 5589 lines

https://claude.ai/code/session_013eJcTLCQcXrXHA4CC8eDpp


ubik2:
I'm not sure we should have FabricValues in our bindings.
Depending on where we go with Defaults, I'm inclined to exclude general FabricValues from those as well.
We may want to support them for initial values, but I'm not sure whether that should end up in the pattern bindings or not.

Edit: I'm also not sure a primitive pattern result makes sense. It's meaningful to return an object containing multiple cells, but if you return a primitive type, why did you have a pattern (technically, this could still make sense for a side effect of running the pattern, but the general idea of patterns is that they're side-effect free). This wouldn't be reactive.

These bindings and results start out as symbolic names (like "argument") saying that the input comes from a cell that will be attached to our argument field. Then we unwrap them, and have an actual cell, so now we know that the input comes from this cell (or the output should be written to this cell). It never makes sense for those locations to be more complex FabricValues.

We want to support some basic objects, so I can say this function takes 2 arguments, and they're named "start", "end" (or perhaps it takes an array of arguments that don't have names), but these aren't *values*.
Where it does get complicated is when we're embedding default values (e.g. default value for "end" is -1), and we do sort of include that information in the bindings. Similarly, if we need to set initial values for local function variables, we'd currently store that in the bindings.

Edit2: For the "returns a primitive" case, `result` should be something like `{ $alias: { partialCause: "foo", path: [] } }` (an object), and then that will be transformed to a link object, and then when we do a `get()` on the cell, we get the primitive value.

This is all a bit more nuanced, so it's probably a good idea to chat more on Monday.

Edit3: If this is blocking your work, feel free to go ahead and land it, since we can always revisit this later. It just seemed like we had the wrong sense of these things as values rather than a set of pointers to values.

danfuzz:
@ubik2 Thanks for the read and analysis. This PR is definitely fixing one of the many existing problems that prevent `FabricValue`s from being storable in the system and usable for calculation by pieces. And if they're storable in the system and usable by pieces, schemas have to be able to have a way of specifying `FabricValue` subtypes as acceptable values. It's not clear to me what part of that you're unsure about supporting.

Drilling down into what this PR is actually doing: It's taking a spot which de facto _can_ receive any `FabricValue` and makes it no longer do the wrong thing for the things that are of JS type `object` but are neither a plain object or an array (that is, a concrete instance of `SpecialFabricObject`).

We still have at least a dozen or so other places in the code similar to this that haven't been addressed at all or are only partially addressed. Detail: There are two direct subclasses of `FabricSpecialObject`. Of them, `FabricPrimitive` is straightforwardly addressable now, and @mathpirate shepherded a bunch of those fixes into the code; but `FabricInstance` is still in need of some infrastructure support in order for the use sites in question to be made correct.

PR #5009, opened 2026-07-25
docs(plans): `cf space clone` — rehearsal-grade space copies

mpsalisbury:
## Why

Updating a deployed pattern on a live space is now a rehearsal-grade event, and
the practice that works is undocumented and unbuilt. The July Topics migration
on Estuary showed the whole failure surface: the legacy→current transition fails
the compat checker; a first production attempt was rolled back on false-negative
health checks (cold loads take ~20 s, unstepped CLI reads look empty); a prior
update drove cross-version write storms to 96% of all commits. The mitigation
that worked was a hand-built copy-on-write clone rehearsal — copy the space's
SQLite store, serve it from a local toolshed, migrate serially, verify, reset,
repeat — and it caught two data-corrupting schema bugs (#4997) that static
checking could not.

`docs/plans/pattern-verb-contract-implementation.md` commits to continuous
dogfood on that board and names the gate in Risks ("before any live `setsrc`,
rehearse old→populate→new in a scratch space watching commit rates"), but that
instruction has no executable form. Every phase implies another rehearsal, and
each one currently re-derives the `engine-v3/<did>.sqlite` filename encoding and
the WAL-companion reset rule from scratch — the two facts people get wrong, both
silent when wrong.

This is the research deliverable that decides whether and how to make it a
command. It proposes nothing be implemented yet: it is explicitly gated on
interviewing Wilk, who built the July rehearsal setup with Gideon.

## What Changed

New live plan `docs/plans/space-clone-rehearsal.md` (status: proposed). No code.

The recommendation is a thin `cf space clone` plus `cf inspect churn`, with the
orchestration left manual. The decisions it records:

- **Identity — same-DID.** Entity ids are content-addressed without the space
  DID and same-space links elide `space`, so a re-key looks feasible; it isn't.
  Stored CFC labels carry `Space(...)`/`HasRole` space DIDs as commitment
  fields, and the render ceiling correlates them against `HasRole` facts minted
  from the *actual* space — a re-keyed clone **fails closed** and manufactures
  over-blocks production won't have. `aclDocId` is `of:<space>`, so a re-key
  also drops the clone onto the deprecated missing-ACL compatibility branch, and
  historical `invocation.sub` rows can't be re-signed.
- **Reset — APFS clonefile from a pristine `VACUUM INTO` snapshot.** Measured on
  a 124 MB WAL store with a writer held open: VACUUM 176 ms (no `-wal`
  companion), `cp -c` 8.9 ms, reset 3.5 ms, source and pristine byte-identical
  afterwards. Migration-on-open is safe on a copy — `Engine.open` is forward-only
  and idempotent.
- **Acquisition — scp-then-clone.** The production dump refusal is deliberate
  (fail-closed env allowlist, no override, no size/concurrency cap, and a dump is
  a whole space) and this proposal does not touch it. Staging reuses the existing
  `cf inspect pull` path.
- **Layering — `cf space` over `state-inspector`** (both layer 4). Not
  `scripts/`, which would have to copy `resolveSpaceStoreUrl`'s mode-dependent
  encoding; not under `cf inspect`, whose read-only contract is its value.
- **One real gap found:** `hotEntities` gives all-time write counts with no time
  dimension, so nothing in the repo can show a storm starting or a settle
  completing. `cf inspect churn` is the only piece with no existing substitute
  and earns its keep on live spaces regardless.
- **Fidelity caveats**, so a rehearsal isn't over-trusted: cross-space links
  resolve to silently-empty spaces (the server creates stores on demand), and a
  clone tests the store and runtime, not the deployment.
- **Blessed practice** — when a clone rehearsal is required vs not, and Gideon's
  runbook made checkable (snapshot manifest + fingerprint, quiet window, serial
  batches with the board last, step + verify, ≥20 s cold-load patience, rollback
  manifest, two clean passes).
- An honest **don't-build-it** case, and seven open questions for Wilk.

## Test plan

Docs-only, no code. The doc contains no `ts`/`tsx`/`js` fenced blocks, so it
contributes nothing to `deno task check-docs`; `docs/` is excluded from
`deno fmt`. No skills touched, so `deno task check-skill-facts` is unaffected.

Claims are cited to source and were verified against the tree — `dump.ts`,
`storage-path.ts`, `server.ts` (`aclDocId`, the missing-ACL branch, on-demand
space creation), `engine.ts` (the invocation/authorization tables, the
forward-only open path), `create-ref.ts`, `link-utils.ts`, the CFC label
classification and render-ceiling correlation, `queries.ts`, and
`memory-dump-policy.ts`. The performance table comes from a probe run against a
real 124 MB WAL-mode SQLite store with a live writer connection, not estimates.

---
## Summary by cubic
Updates the proposal for `cf space clone` (same-DID rehearsal copies) and `cf inspect churn` with interview results, making the manifest + `--verify` and churn the core and keeping copy mechanics a thin wrapper. Adds https-based `--from` for shared snapshots, pins hashing to `@commonfabric/content-hash`, treats APFS clonefile as an optimization with fallback, positions churn alongside OTel telemetry and #4950, and keeps the stance on scp‑then‑clone acquisition, `cf space` over `state-inspector`, safety rails, and a checkable runbook; docs only, no code.

<sup>Written for commit 934bd05a1112d6fe49fc91ca3e9b182ba9e1acfb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5009?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/plans/space-clone-rehearsal.md
+++ b/docs/plans/space-clone-rehearsal.md
@@ -1,0 +1,370 @@
+# `cf space clone` — rehearsal-grade space copies
+
+**Status:** proposed. Research complete; not yet agreed. Written 2026-07-24
+against the July Estuary Topics migration, which established that updating a
+deployed pattern on a populated space is a rehearsal-grade event. Decides
+whether the hand-built clone rehearsal becomes a command, and what shape.
+
+**Provisional pending interview.** Wilk built the rehearsal setup with Gideon.
+Everything below is derived from the code, the PR record, and the board; the
+questions his answers settle are listed in [Open questions for
+Wilk](#open-questions-for-wilk) and marked *(W)* inline. Nothing here should be
+implemented before that pass.
+
+## Why
+
+The July Topics migration exposed the whole failure surface of a live pattern
+update, and every mitigation that worked was hand-built:
+
+- the legacy→current transition **fails the compat checker**, forcing
+  `--dangerously-allow-incompatible-schema` plus manual validation;
+- a first production attempt was **rolled back on false-negative health
+  checks** — cold browser loads take ~20 s, and unstepped CLI reads look empty;
+- historical pattern updates produced **generated-cell write storms**: per the
+  board's incident record, 20 generated cells produced 192,381 revisions —
+  90.3% of every revision and 96.0% of all commits in the 2026-07-22 snapshot,
+  peaking at 1,598 commits/minute;
+- the thing that actually worked was a **copy-on-write clone rehearsal**: copy
+  the space's SQLite store, serve it from a local toolshed, migrate all 73
+  topics serially, verify counts / fingerprint / cold reads / write settlement,
+  reset, repeat.
+
+That rehearsal caught two data-corrupting schema bugs ([#4997]) that static
+checking could not — a dangling legacy `createdByName` author link and a
+recursive `TopicPiece` crossref that made validation stop making progress. It
+also demonstrated the fix for the write storm: "current-main clients against
+the local clone become completely quiet."
+
+The verb-contract arc commits to continuous dogfood on this board — every phase
+implies another rehearsal — and its Risks section already names the gate:
+"before any live `setsrc`, rehearse old→populate→new in a scratch space watching
+commit rates." Today that instruction has no executable form. This document
+decides what to build so it does.
+
+[#4997]: https://github.com/commontoolsinc/labs/pull/4997
+
+## Recommendation in one paragraph
+
+Build a **thin, offline `cf space clone`** — the store-path knowledge, the
+crash-consistent copy, the pristine/working split, and a manifest — plus **`cf
+inspect churn`**, the time-bucketed commit-rate query that makes the write-storm
+gate executable. Clone **same-DID**. Reset by **APFS clonefile from a pristine
+snapshot**. Leave server-side acquisition, toolshed orchestration, and the
+migration driver **manual**. Roughly 400 lines across two layer-4 packages,
+against a practice the team is already committed to repeating every phase.
+
+## What exists (verified)
+
+| Fact | Where |
+| --- | --- |
+| Crash-consistent snapshot via `VACUUM INTO`; never mutates source, never runs the engine's DDL-on-open path; produces one file with no `-wal`/`-shm` | `packages/memory/v2/dump.ts` |
+| A space is one SQLite file whose name encodes its DID; directory mode nests under `engine-v3/` with the **literal** DID as the stem, single-file mode percent-encodes it | `packages/memory/v2/storage-path.ts` |
+| The store dir defaults to `./cache/memory/` relative to cwd — so storage is already isolated per checkout | `packages/toolshed/env.ts` (`MEMORY_DIR`) |
+| The dump endpoint is hard-off in production: fail-closed env allowlist, no override, and "a prod form is a separate mechanism" | `packages/toolshed/routes/storage/memory/memory-dump-policy.ts` |
+| `cf inspect --remote` fetches snapshots into `~/.cache/cf-inspect/<host>/` | `packages/state-inspector/remote.ts` |
+| Genesis/ACL ride inside the store (`of:<space>` doc), so existing team keys authorize on a copy | `packages/memory/v2/server.ts` |
+| `cf inspect` is read-only by contract — "it explains, it never reproduces or replays" | `skills/state-inspector/SKILL.md` |
+
+## Identity: same-DID, and it is not close
+
+The clone keeps the source space's DID. Re-keying is not a rename; it is an
+unbounded data migration, and it degrades exactly the fidelity the rehearsal
+exists to buy.
+
+**What would survive a re-key.** Entity ids are content-addressed from
+`{source, cause}` and do **not** include the space DID (`createRef` in
+`packages/runner/src/create-ref.ts`), so the document graph is space-agnostic.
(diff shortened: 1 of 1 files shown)


docs/plans/space-clone-rehearsal.md
```diff
@@ -0,0 +1,370 @@
...
+
+The live run then repeats 2–6 against production, with the rollback manifest in
+hand.
+
+## Open questions for Wilk
+
+1. **What did the rehearsal actually do end to end** — from acquiring the
+   Estuary store to a verified pass? Especially: how the store came off the
+   server, and whether the toolshed was `start-local-dev.sh` with an offset or
+   something bespoke.
```
willkelly: I did the vacuum into, scped the file, and uploaded it to s3 to allow others to download it.  I did also copy it into the cache dir of a running toolshed so I could poke around at it.  Gideon presumably did the same.  I imagine both of us just told our llm coding agents to make it happen -- that was certainly my approach!

docs/plans/space-clone-rehearsal.md
```diff
@@ -0,0 +1,370 @@
...
+2. **Same-DID or re-keyed?** And did same-DID produce any near-miss — a `cf`
```
willkelly: I used same-did.  I didn't notice any trouble, but I also mainly interacted with it via agent -- there could have been some trouble I missed.

docs/plans/space-clone-rehearsal.md
```diff
@@ -0,0 +1,370 @@
...
+   command or browser tab that hit the wrong endpoint?
+3. **What was the reset?** `cp -c`, a fresh copy, something else — and how many
+   attempts did the 73-topic migration take before it was clean?
```
willkelly: This is a gideon question, I think.  I used cp for my experiments.

willkelly: (no -c, full copy)

docs/plans/space-clone-rehearsal.md
```diff
@@ -0,0 +1,370 @@
...
+4. **What was fiddly versus obvious?** This is the load-bearing question for
+   build-vs-runbook. Specifically: did the `engine-v3/<did>.sqlite` path
```
willkelly: no, this worked fine, but the filename was already did.sqlite

willkelly: also I vacuum intoed, so no wal companions.

docs/plans/space-clone-rehearsal.md
```diff
@@ -0,0 +1,370 @@
...
+   encoding, or WAL companions on reset, cost you time?
+5. **How were counts and the content fingerprint computed** — ad-hoc SQL, `cf
+   inspect`, or by hand? That determines how much of the manifest is new code.
```
willkelly: My stuff was mainly ad-hoc sql done by an agent.

willkelly: It did use a lot of python one liners / heredocs

docs/plans/space-clone-rehearsal.md
```diff
@@ -0,0 +1,370 @@
...
+6. **Was commit rate watched during the rehearsal, and how?** If there is
+   already a query, `cf inspect churn` should adopt it rather than invent one.
```
willkelly: There is telemetry for commits.  Also https://github.com/commontoolsinc/labs/pull/4950

docs/plans/space-clone-rehearsal.md
```diff
@@ -0,0 +1,370 @@
...
+7. **Anything that made the clone lie** — behaved differently from production in
```
willkelly: Not that I noticed, but as we do more profile stuff and more links, the links stuff could come into play more.

docs/plans/space-clone-rehearsal.md
mathpirate: i think i gave my best attempt at answering this question when we spoke live mike. as wilk said elsewhere, i very much relied on my agents to advise me as to the details of making it happen

mathpirate (approved):

mpsalisbury:
Thanks @willkelly and @mathpirate — all seven answers are folded into the doc as of 934bd05a1, replacing the "provisional pending interview" gate with an Interview results section. Two of them changed the doc materially rather than just annotating it:

- **Q4 (path encoding / WAL companions) came back negative**, so the don't-build-it section now concedes its original load-bearing prediction was wrong — server-side `VACUUM INTO` absorbs both. The build case now rests on Q5: counts and fingerprints re-invented as ad-hoc SQL per run is exactly what the manifest + `--verify` replace, so they're promoted to the command's payload and the copy mechanics demoted to a thin wrapper.
- **Q6's pointers** (commit telemetry, #4950) repositioned `cf inspect churn` from "doesn't exist today" to a retrospective-offline complement of both.

Also recorded: the S3 sharing hop (so `--from` should take an https URL), same-DID with your agent-mediated caveat, and the links/profile watch item under fidelity caveats.

— cf-review revision via Claude (Fable 5), on behalf of @mpsalisbury

PR #5016, opened 2026-07-25
ci: delete the dead Deploy Shell to Production workflow

Hixie:
The "Deploy Shell to Production" workflow uploaded the shell build to the production-commontools-dev GCS bucket, which serves https://production.commontools.dev/. It was manual-dispatch only, and it ran three times in its whole history. The last time it succeeded was 10 December
2025. The one attempt after that, on 27 March 2026, failed at the checkout step, so nothing after it ran; the logs for that run are past GitHub's retention period, so the exact error cannot be recovered. The likely cause is that the free-text sha input was given an abbreviated commit, because actions/checkout accepts only a full forty-character commit.

The site shows the same story. Its index page still carries a Last-Modified of 10 December 2025, matching that last successful run, and what it serves is a previous generation of the product: a 431-byte index page referencing ct.svg, an <x-root-view> element, and the name "Common Tools". Staging, which the deploy-shell-staging job in deno.yml keeps current, serves the Common Fabric shell instead. So this workflow has not been part of how the product ships for seven months.

The workflow file itself has had no substantive edit since it was added. Everything that touched it since was a repository-wide sweep: the Blacksmith runner migration, the Node 24 action bumps, the worker bundle code-split.

The dashboard had already stopped believing this URL meant anything about production; the comment at packages/dashboard/tiles/prod-uptime.ts explains that the index page answers 200 through a complete outage of the server behind it.

Nothing else in the repository referred to the workflow or its bucket. The staging deployment path in deno.yml is untouched, and is now the only shell deployment the repository performs.

Retiring the bucket, the production GitHub environment, and the service account key behind GCP_SA_KEY needs cloud admin, so those remain.

While reading that dashboard comment, one stale-looking pointer in it is fixed too. It cited tofu/shell/README.md as a bare path, which reads as a path in this repository, and there is no tofu directory here. The document is real, but it is in the infra repository, so the comment now names that repository alongside the path.

---
## Summary by cubic
Delete the unused “Deploy Shell to Production” GitHub Actions workflow that uploaded shell assets to `gs://production-commontools-dev`, and update the dashboard uptime comment to reference the infra repo path.
Staging deploy in `deno.yml` remains the only shell deployment; cloud resource cleanup (bucket, GitHub environment, service account) is out of scope.

<sup>Written for commit d88048a259970d8b4835cc03e0d7d85f22773430. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5016?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Hixie:
fyi @willkelly — if you needed this, let me know and i will revert it (or you can just revert it)

PR #5025, opened 2026-07-25
fix(runner): an authored request is a `FabricValueLayer`, not a `FabricValue`

danfuzz:
`BuiltInLLMImagePart.image` is `string | Uint8Array | ArrayBuffer | URL` — an
LLM request legitimately carries **native** values a caller handed in. The
request path nevertheless declared its values `FabricValue`, which they are not
until conversion.

`FabricValueLayer` is the data-model type for exactly this case: *"the shape is
right, the contents may still need converting."*

## Retyped

- `createFrozenRequestSnapshot`'s constraint
- the four `request` parameters in `cfc/sink-request.ts`
- the `request` field of the `sink-request` write-policy input

Nothing outside `sink-request.ts` reads that field, so the change **stops there**
rather than propagating — I checked before widening it, having recently measured
what happens when a tightening doesn't stop (`IFCLabel`: 86 errors and growing).

## Four of six errors were cascades

`createFrozenRequestSnapshot` is already generic (`<T extends FabricValue>`), so
when its argument failed the constraint, `T` fell back to `FabricValue` and the
assignment back to `llmParams: LLMRequest` failed too. Fixing the constraint
fixed both ends — one root, four symptoms.

## The one cast, deliberately

`cloneIfNecessary()` handles these natives at runtime — its tag dispatch covers
`Date`, `RegExp`, and `Uint8Array` — but its *parameter type* names only the
already-converted case. Hence `cloneIfNecessary(value as FabricValue)` inside the
snapshot helper, with a comment saying so.

The alternative is widening `cloneIfNecessary` itself, which reaches much further
into `data-model` than this change should. Flagging it explicitly as the
judgement call in here.

## Measurement

Against the `FabricSpecialObject` branding experiment (which is what makes these
failures visible): residue **79 → 73**.

No behavior change; no test moved. Workspace `deno task check` clean; full repo
suite green (199.2s).

https://claude.ai/code/session_01NNgY3ooreGzww89dwayfst

---
## Summary by cubic
Switch authored sink requests to `FabricValueLayer` instead of `FabricValue` so LLM requests can include native values (e.g., `Uint8Array`, `ArrayBuffer`, `URL`). Fixes type errors with no behavior changes.

- **Bug Fixes**
  - Retyped `createFrozenRequestSnapshot<T extends FabricValueLayer>` and cast to `FabricValue` when calling `cloneIfNecessary`.
  - Updated four `request` parameters in `cfc/sink-request.ts` and the `sink-request` `WritePolicyInput.request` field to `FabricValueLayer` (change contained to this module).
  - Reduced type-check residue from the `FabricSpecialObject` branding experiment (79 → 73).

<sup>Written for commit 37c5b04c8e5856e07bc2ee2d60cc2a5b4d09c9c3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5025?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/cfc/request-snapshot.ts
+++ b/packages/runner/src/cfc/request-snapshot.ts
@@ -1,5 +1,6 @@
 import { cloneIfNecessary } from "@commonfabric/data-model/fabric-value";
 import type { FabricValue } from "@commonfabric/api";
+import type { FabricValueLayer } from "@commonfabric/data-model/interface";
 
 /**
  * Produces a detached, deep-frozen snapshot of a request `value`, for use as a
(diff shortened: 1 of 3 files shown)


danfuzz:
Closing: the premise is wrong.

I justified widening `createFrozenRequestSnapshot` to `FabricValueLayer` on the claim that `cloneIfNecessary()` handles native values at runtime. It does not — it throws (`Cannot clone: Uint8Array`), nested or not. I had read that tag dispatch off `hashOf`, which does convert `Date`/`RegExp`/`Uint8Array` on the fly, and wrongly attributed it to the clone path.

Worse, the request at that call site isn't native-bearing anyway: `messages` comes from `inputs.withTx(tx).get()`, so it is stored data — a stored image is a `FabricBytes`. The `Uint8Array | ArrayBuffer | URL` in `BuiltInLLMImagePart` describes the authoring surface, not what flows here.

So this change would have turned a compile-time error into a runtime throw for precisely the case the type was flagging. The genuine issue is that `LLMRequest` is declared via an authoring-time type that no longer describes the value once it has been through storage; the fix belongs there, not in the snapshot helper.

Thanks @danfuzz for catching it.

PR #5036, opened 2026-07-26
refactor(cfc): label carriers name the clauses and atoms they hold

danfuzz:
Types the declarations that **carry** CFC clause and atom lists, completing the
vocabulary started in #5031 / #5033 / #5035. `IFCLabel`,
`CfcObservedConfidentiality`, `SinkMaxConfidentiality`, the ceiling / per-row /
exchange / introspection records, and the sqlite, llm, html, and cf-harness
label plumbing now say `CfcConfClause` / `CfcAtom` where they said `unknown[]`.

125 declarations retyped. The 29 remaining casts sit where a value genuinely
arrives untyped: cell reads, `transformValue` results, and the `CfcOrClause`
literals a test builds by hand.

No runtime change — annotations, casts, and imports only.

## Why this is one change

Every CFC atom list in the repo is a single type-connected component, so no
subset of it compiles on its own. That was measured, not assumed: typing the
carriers alongside the clause layer produced 122 errors reaching `api`,
`cell.ts`, `html`, and `runtime-client`, and typing successive waves of
declarations moved the count 115 → 118 → 76 → 85 rather than steadily down —
each wave pushed the frontier into whatever passes label data through. It
converged only once the work switched from *declaring* (which expands the
frontier) to *closing individual sites* (which terminates).

#5035 already established that the other obvious seam does not help: narrowing
producer return types is safe for consumers and therefore removes no
consumer-side error.

## A split that would work, for next time

`confidentiality` (63 declarations) and `integrity` (15) are near-disjoint
subsystems — clause machinery versus floor/trust machinery — sharing only the
`IFCLabel` container. Reverting just the integrity declarations from this branch
leaves **8 errors**, roughly half of them integrity-only locals that would not
exist in a confidentiality-only change. The genuine coupling is the key-generic
loops (`for (const key of LABEL_KEYS) label[key] = …` in `cloneCfcLabel`,
`redactCaveatSourcesForDisplay`, and one in `ui`), which would need a cast in
the first PR and shed it in the second.

Every earlier split attempt cut *across* the type graph and cost 40–120 sites or
bought nothing; this one cuts *along* it, separating two components that were
only ever adjacent in a struct. Not applied here — the split is lopsided and
costs casts — but it is the seam to use on the next label-typing job.

## Notes

- `cf-harness` type-checks inside its own `deno task test`, so a clean
  `deno task check` does **not** imply the repo type-checks. Its failure is the
  second commit here.
- Test helpers carried most of the churn: typing four of them —
  `clauseSetsEqual`, `caveatAtom`, `derivedBodyEntry`, `confidentialityOf` —
  cleared 56 errors at once.
- `packages/ui` also reads labels generically. It type-checks fine as-is and is
  untouched here, but the label vocabulary reaches one package further than a
  grep for `confidentiality:` finds.

## Test plan

- `deno task check` clean, `deno fmt --check` and `deno lint` exit 0.
- `packages/runner`: 1068 passed, 0 failed. `packages/cf-harness`: 500 passed.

---
## Summary by cubic
Typed all containers that carry CFC clauses and atoms, replacing `unknown[]` with `CfcConfClause` and `CfcAtom` for end-to-end type safety. Impacts runner, HTML worker, SQLite and LLM builtins, and `cf-harness`; no runtime changes.

- **Refactors**
  - Typed `IFCLabel`, `CfcObservedConfidentiality`, `SinkMaxConfidentiality`, render policies, and exchange/introspection records to use `CfcConfClause`/`CfcAtom`.
  - Updated SQLite label read/write and write-ceiling, LLM dialog/generate flows, and the HTML worker reconciler to propagate typed atoms/clauses.
  - Fixed `cf-harness` model context to accumulate `CfcConfClause[]` for confidentiality.
  - Left a few casts where inputs are genuinely untyped (cell reads, `transformValue` outputs, specific test literals).

<sup>Written for commit cc40c6ece2c0f31855da26b8c822cce331e0e62c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5036?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cf-harness/src/contracts/cfc-model-context.ts
+++ b/packages/cf-harness/src/contracts/cfc-model-context.ts
@@ -1,4 +1,8 @@
-import type { CfcLabelView, IFCLabel } from "@commonfabric/runner/cfc";
+import type {
+  CfcConfClause,
+  CfcLabelView,
+  IFCLabel,
+} from "@commonfabric/runner/cfc";
 import type { HarnessCfcInvocationInputLabelPath } from "./cfc-invocation-context.ts";
 import type { ToolOutputId } from "./tool-result.ts";
 
@@ -81,7 +85,7 @@ const labelValueKey = (value: unknown): string => {
 export const mergeConfidentialityOnlyLabels = (
   labels: readonly (IFCLabel | undefined)[],
 ): IFCLabel | undefined => {
-  const confidentiality: unknown[] = [];
+  const confidentiality: CfcConfClause[] = [];
   const seen = new Set<string>();
   for (const label of labels) {
     const confidentialityOnly = label === undefined
--- a/packages/html/src/worker/reconciler.ts
+++ b/packages/html/src/worker/reconciler.ts
@@ -27,6 +27,7 @@ import {
   UI,
   useCancelGroup,
 } from "@commonfabric/runner";
+import type { CfcAtom } from "@commonfabric/api/cfc";
 import type { CellRef } from "@commonfabric/runtime-client";
 import { deepEqual } from "@commonfabric/utils/deep-equal";
 import { getLogger } from "@commonfabric/utils/logger";
@@ -170,7 +171,7 @@ export class WorkerReconciler {
     );
     this.rootRenderPolicy = ceiling === undefined ? DEFAULT_RENDER_POLICY : {
       declassifyConfidentiality: [],
-      maxConfidentiality: [...(ceiling.atoms ?? [])],
+      maxConfidentiality: [...(ceiling.atoms ?? [])] as readonly CfcAtom[],
       caveatKindAllow: [...(ceiling.caveatKinds ?? [])],
     };
   }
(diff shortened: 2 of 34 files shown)


danfuzz:
Superseded. This PR typed the whole CFC label-carrier surface in one 33-file
change; it has since been split into measured slices, all of which are now
merged or open:

- #5037 — integrity carriers (21 declarations → 15 sites) — merged
- #5038 — ceilings / allow-lists (19 → 25) — merged
- #5039 — observed / flow confidentiality (22 → **0**) — merged
- #5041 — stored label, leaf consumers (10 → 6)
- #5042 — stored label, core carriers (18 → 38)

The split was possible because the predictor for a cheap boundary turned out to
be **interface width — how many call sites cross it** — not whether the two
sides look like separate subsystems. Cuts across the type graph (clause layer vs
carriers, producers vs consumers, core vs consumers) each cost 40–120 sites;
these domain boundaries cost 0–38.

PR #5040, opened 2026-07-26
refactor(cfc): stored label carriers name their clauses

danfuzz:
Types the **stored-label** side of CFC confidentiality as `CfcConfClause[]`:
`IFCLabel`, the sqlite per-row and static label paths, `prepare.ts`'s
flow/projection records, label-introspection, label-representation,
render-ceiling's label argument, and the html and cf-harness label surfaces.

28 declarations, 45 fallout sites, 27 files. No runtime change — annotations,
casts, and imports only.

**With this, no confidentiality, integrity, ceiling, or observed carrier in the
repo is still `unknown[]`.** The CFC label vocabulary is fully named. This
completes the series begun in #5031 and supersedes #5036.

## Stacked on #5039

Depends on #5039 (observed confidentiality), which factors out the observation
side. Please land that one first.

## One deliberate `unknown[]`

`confidentialityOf` keeps `readonly unknown[]`. It is a callback **parameter**,
so narrowing it propagates to every implementation — measured at 23 errors when
tried. Cast at the single call site that needs the clause type instead.

That is the same contravariance asymmetry #5035 established and #5038 hit:
producer *return* types narrow for free because the narrowed value is assignable
to what consumers already accept; consumer *parameter* types never do.

## What made this series splittable

The predictor for a cheap split turned out to be **interface width — how many
call sites cross the boundary** — not whether the two sides look like separate
subsystems:

| Slice | Decls | Fallout | Crossings |
| --- | ---: | ---: | --- |
| integrity (#5037) | 21 | 15 | the `LABEL_KEYS` loops |
| ceilings (#5038) | 19 | 25 | independent params of `cfcObservationFitsCeiling` |
| observed (#5039) | 22 | **0** | 3 named functions |
| stored label (this) | 28 | 45 | — remainder |

Earlier attempts cut *across* the type graph (clause layer vs carriers, producers
vs consumers, CFC core vs consumers) and each cost 40–120 sites, because a layer
boundary here has hundreds of crossings. Every CFC atom list is one
type-connected component; only narrow-waisted domain boundaries isolate.

## Test plan

- `deno task check` clean; `deno fmt --check` and `deno lint` exit 0.
- `packages/cf-harness`: 500 passed (it type-checks in its own suite, so a clean
  root `check` does not cover it).

---
## Summary by cubic
Names stored-label confidentiality clauses as `CfcConfClause[]` across the codebase to remove remaining `unknown[]` in label carriers. No runtime changes.

- **Refactors**
  - Typed stored-label carriers: `IFCLabel.confidentiality`, sqlite per-row/static label paths and row read/write, write-ceiling checks, builder/node utils, cell boundary, and flow/projection records in `prepare.ts`.
  - Updated ceiling/render and sanitization helpers to accept `CfcConfClause[]` (e.g., observation fit checks, atoms outside ceiling, `schemaWithLub`, material-risk discharge).
  - Updated UI surfaces in `packages/html` and `packages/cf-harness` to pass `CfcConfClause[]`.
  - Adjusted tests to use clause types.
  - Kept the `confidentialityOf` callback parameter as `readonly unknown[]`; cast at the single call site that needs `CfcConfClause`.

<sup>Written for commit d5d96f584d96940233aa29a957e795100a26f0a3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5040?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cf-harness/src/contracts/cfc-model-context.ts
+++ b/packages/cf-harness/src/contracts/cfc-model-context.ts
@@ -1,4 +1,5 @@
 import type { CfcLabelView, IFCLabel } from "@commonfabric/runner/cfc";
+import type { CfcConfClause } from "@commonfabric/runner/cfc";
 import type { HarnessCfcInvocationInputLabelPath } from "./cfc-invocation-context.ts";
 import type { ToolOutputId } from "./tool-result.ts";
 
@@ -81,7 +82,7 @@ const labelValueKey = (value: unknown): string => {
 export const mergeConfidentialityOnlyLabels = (
   labels: readonly (IFCLabel | undefined)[],
 ): IFCLabel | undefined => {
-  const confidentiality: unknown[] = [];
+  const confidentiality: CfcConfClause[] = [];
   const seen = new Set<string>();
   for (const label of labels) {
     const confidentialityOnly = label === undefined
--- a/packages/html/src/worker/reconciler.ts
+++ b/packages/html/src/worker/reconciler.ts
@@ -1067,7 +1067,7 @@ export class WorkerReconciler {
    * admit only bare atoms — an OR-clause never matches either, staying closed.
    */
   private resolvedConfidentialityRenderable(
-    confidentiality: readonly unknown[],
+    confidentiality: readonly CfcConfClause[],
     integrity: readonly CfcAtom[],
     policy: RenderPolicy,
   ): boolean {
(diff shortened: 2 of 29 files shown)


danfuzz:
Superseded — split into two measured halves, per the seam analysis in this
series:

- #5041 — leaf consumers (sqlite, html, cf-harness): 10 declarations, 7 files
- #5042 — core carriers (`IFCLabel`, prepare, label-*): 18 declarations, 19 files

The boundary is ~4 coupling sites, three of which were artifacts of a
pattern-based revert rather than real coupling. The leaves construct labels and
assign them into `IFCLabel`, so narrowing them while the destination stays wide
is narrow-into-wide — free on its own.

Closing in favour of the pair. Nothing is lost: between them they carry the same
28 declarations this PR did.

PR #5042, opened 2026-07-26
refactor(cfc): core label carriers name their clauses

danfuzz:
Types the **core** stored-label carriers as `CfcConfClause[]`: `IFCLabel`,
`prepare.ts`'s flow/projection records, label-introspection,
label-representation, render-ceiling's label argument, observation's label
inputs, schema-sanitization, plus the `cfc.ts`, `data-updating.ts`, and
`builder/node-utils.ts` label surfaces.

18 declarations, 38 fallout sites, 19 files. No runtime change — annotations,
casts, and imports only.

With this and #5041, **no confidentiality, integrity, ceiling, or observed
carrier in the repo is still `unknown[]`** — the CFC label vocabulary is fully
named. Completes the series begun in #5031; supersedes #5036 and #5040.

## Relationship to #5041

Independent branches off `main`; either can land first. #5041 narrows the leaf
consumers that assign INTO these types (narrow-into-wide, free on its own); this
narrows the destination. When both are in, the two casts #5041 adds become
removable.

## One deliberate `unknown[]`

`confidentialityOf` keeps `readonly unknown[]`. It is a callback **parameter**,
so narrowing it propagates to every implementation — measured at 23 errors when
tried. Cast at the single call site that needs the clause type instead.

Producer *return* types narrow for free (the narrowed value is assignable to the
`unknown[]` consumers already accept); consumer *parameter* types never do. That
asymmetry was measured in #5035 (21 producer-side errors vs 122 for a comparable
consumer-side set) and has held for every slice since.

## Test plan

- `deno task check` clean; `deno fmt --check` and `deno lint` exit 0.
- `packages/cf-harness`: 500 passed — it type-checks inside its OWN
  `deno task test`, so a clean root `check` does not cover it.

---
## Summary by cubic
Typed CFC confidentiality arrays as `CfcConfClause[]` across core and leaf label surfaces, replacing `unknown[]`. This fully names the label vocabulary with no runtime changes.

- **Refactors**
  - Core: `IFCLabel`, flow/projection in `prepare.ts`, label-introspection/representation, render-ceiling inputs, observation helpers, schema-sanitization, plus surfaces in `cfc.ts`, `data-updating.ts`, and `builder/node-utils.ts`.
  - Leaf: SQLite row/static label paths, `html` worker label surface, and `cf-harness` model context.
  - Kept `confidentialityOf` callback param as `readonly unknown[]`; cast at the one call site that requires `CfcConfClause`.
  - Test types updated to use `CfcConfClause`; annotations, casts, and imports only.

<sup>Written for commit ab3b76ae46225bbf3f1481db331908b6cbf84dab. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5042?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cf-harness/src/contracts/cfc-model-context.ts
+++ b/packages/cf-harness/src/contracts/cfc-model-context.ts
@@ -94,7 +94,7 @@ export const mergeConfidentialityOnlyLabels = (
         continue;
       }
       seen.add(key);
-      confidentiality.push(cloneJsonValue(value));
+      confidentiality.push(cloneJsonValue(value) as CfcConfClause);
     }
   }
   return confidentiality.length > 0 ? { confidentiality } : undefined;
--- a/packages/cf-harness/src/contracts/cfc-model-context.ts
+++ b/packages/cf-harness/src/contracts/cfc-model-context.ts
@@ -1,4 +1,5 @@
 import type { CfcLabelView, IFCLabel } from "@commonfabric/runner/cfc";
+import type { CfcConfClause } from "@commonfabric/runner/cfc";
 import type { HarnessCfcInvocationInputLabelPath } from "./cfc-invocation-context.ts";
 import type { ToolOutputId } from "./tool-result.ts";
 
@@ -81,7 +82,7 @@ const labelValueKey = (value: unknown): string => {
 export const mergeConfidentialityOnlyLabels = (
   labels: readonly (IFCLabel | undefined)[],
 ): IFCLabel | undefined => {
-  const confidentiality: unknown[] = [];
+  const confidentiality: CfcConfClause[] = [];
   const seen = new Set<string>();
   for (const label of labels) {
     const confidentialityOnly = label === undefined
--- a/packages/html/src/worker/reconciler.ts
+++ b/packages/html/src/worker/reconciler.ts
@@ -1067,7 +1067,7 @@ export class WorkerReconciler {
    * admit only bare atoms — an OR-clause never matches either, staying closed.
    */
   private resolvedConfidentialityRenderable(
-    confidentiality: readonly unknown[],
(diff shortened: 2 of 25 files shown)


danfuzz:
Rebased onto `main` (post-#5041) — merged cleanly, `main` is now an ancestor, so
no conflict is possible. Verified after the merge: `deno task check` 0 errors,
`cf-harness` 500 passed, full repo suite green.

**Split analysis**, since this is a 19-file PR. Two seams measured by
revert-and-count:

| Cut | Coupling sites | Effect |
| --- | ---: | --- |
| `IFCLabel` vs its consumers | 8 | lopsided — 1 declaration vs 17; first half still ~18 files |
| `prepare.ts` vs the rest | 14 | two roughly equal halves |

Shipping whole. Both are worse than the seams this series has been landing (0–6
sites), and neither meaningfully reduces the review surface: the change is
+64/−45 of one-line annotations, and 8 of the 19 files are test helpers that
cannot be separated from the src they exercise. The `IFCLabel`-alone capstone is
a real seam if wanted — just a low-value one.

**Correction to the #5041 description:** I said the casts it added would go away
once this lands. Only **one** of the six does. The others are load-bearing for a
different reason — `confidentialityOf` stays `readonly unknown[]` deliberately
(callback parameter), and `res.confidentiality` comes from the sqlite row-label
result rather than from an `IFCLabel`.

PR #5059, opened 2026-07-27
feat(cli): `piece setsrc --check` — compatibility preflight for a source swap

bfollington:
## Why

`cf piece setsrc <main>` replaces the pattern source of a live piece. Until now
it just attempted the swap. When the new pattern could not be applied, the user
found out by attempting it, and what came back was whichever low-level rejection
happened to surface first — a schema-subset assertion, a retained-link proof
failure, or (worst) a raw CFC commit rejection re-wrapped as a plain `Error` at
the runner's setup-commit boundary. None of those tell you *which field* broke
*which rule*.

That cost a production incident on the hosted Estuary deployment over the last
few days: home roots and profile pieces pinned to patterns that could not
migrate their existing documents. #4967 and #4977 added runtime **repairs** for
that class. The lesson those PRs left on the table is the one this PR picks up:
you should be able to know **ahead of time** whether a given source can actually
be applied to a given piece.

So: a `--check` preflight that answers the question and applies nothing, and a
refused real `setsrc` that names the same reason the preflight would.

The design constraint that shaped everything here is **no second copy of the
rules**. A preflight that reimplements compatibility drifts from the enforcement
path and becomes a liar. Every verdict below is produced by driving the real
rule implementation in dry-run.

### The verdict (`packages/piece/src/pattern-update-check.ts`, new)

`checkPatternUpdate()` is pure — no transaction, no writes — and assembles the
verdict from three real rules:

| class | rule driven | source |
|---|---|---|
| `pattern-contract` | the argument/result subset proof `setPattern` asserts with | `patternSchemaCompatibilityIssues` (extracted below) |
| `cfc-schema-migration` / `cfc-schema-merge` | the real CFC envelope merge | `cfcSchemaMergeIssue` → `mergeCfcSchemaEnvelopes` |
| `retained-link` | the same validator `setPattern` installs as `validateArgumentLinks` | `assertSuppliedLinkSchemasCompatible` |

Blockers carry `role` (`argument` = INPUT, `result` = OUTPUT), `path`, `field`,
the rule's verbatim `reason`, and a printable `message`. Advisories carry
non-blocking setup migration work (a new handler stream's marker).

### Exposing the rules for dry-run, without duplicating them

- `packages/runner/src/cfc/schema-merge.ts` gains **`cfcSchemaMergeIssue`** — a
  small pure "would this merge succeed?" helper that runs the real
  `mergeCfcSchemaEnvelopes` and reports the failure, discriminating the
  additive-required migration class via the existing `CfcSchemaMigrationError`.
  `isStreamSlot` is exported so the stream nuance has one definition.
- `packages/piece/src/schema-compatibility.ts` gains
  **`patternSchemaCompatibilityIssues`** — the collector behind
  `assertPatternSchemasBackwardCompatible`, which is now a two-line wrapper
  over it. Thrown message is byte-identical; the early return on invalid
  schema definitions preserves the old early throw.
- `packages/runner/src/cfc/mod.ts` re-exports `readStoredCfcMetadata` and
  `loadSchemaDocument` so the piece layer can read a stored envelope.

### Read-only gathering (`PieceController.checkPattern`)

Reads the CFC schema envelopes at rest for the piece's argument and result
documents through **`runtime.readTx()`**, which physically cannot write (it
throws `ReadOnlyTransactionError` on `abort()`, let alone a write). A role with
no stored CFC metadata is reported `not-applicable` rather than guessed at —
correctly, since the merge only runs against a stored envelope, so with none the
class genuinely cannot fail.

### The CLI (`cf piece setsrc --check [--json]`)

- Exits **0** compatible / **3** incompatible. `3` is documented, distinct from
  1 (operational) and 2 (bad args), and is also the exit code of a real
  `setsrc` refused as incompatible — so a deploy script can branch on
  "incompatible" without parsing text.
- `--json` (only valid with `--check`) emits the verdict via the repo's existing
  `render(value, { json: true })` convention; `reservesStdoutForCommandOutput`
  already keeps stdout clean for it.
- `--check` + `--dangerously-allow-incompatible-schema` is rejected: the
  override exists to skip the proof `--check` computes.

### Better errors without `--check` (task item 3)

`PieceController.setPattern` now wraps both the assertion and the setup commit.
On failure it **re-derives the verdict from the already-compiled candidate** and
throws `PatternUpdateIncompatibleError` carrying the same `blockers` list
`--check` reports. No message sniffing anywhere — the two paths cannot disagree
about the reason, and a test pins them field-for-field.

- Fail-closed is preserved: the setup transaction has already aborted when the
  verdict is re-derived, so the state it re-reads is the untouched pre-update
  state. Tests assert the piece's pattern identity ref is unchanged after a
  refusal.
- A failure with **no** blockers (transport, compile, a genuine bug) is
  rethrown unchanged rather than dressed up as an incompatibility.
- `--dangerously-allow-incompatible-schema` failures are rethrown unchanged:
  that run deliberately skipped the gates, so reporting gate findings would
  misattribute it.
- The originating message is kept as a subordinate `Caused by:` line on
  `.message` (library consumers and stack traces keep the detail); the CLI
  prints only the blockers plus a `--check` pointer.

## What the check actually validates — and what it does not

**Validated:**

1. **Pattern contract**, both roles. Removed fields, incompatible type
   mutations, narrowed bounds/enums, changed `asCell`, and newly-required
   fields with no default. Berni's input/output distinction **is** available at
   this layer — the rule already reports `newly required argument field has no
   default` separately from `newly required result field has no default` — so
   the report labels findings `input (argument)` / `output (result)` from real
   data, not a guess.
2. **CFC document migration**, driving the real merge over the envelope the
   piece's documents actually carry. Verified empirically that a real piece root
   doc does carry `cfc.schemaHash`, and that its stored envelope equals the
   pattern's result schema — i.e. this reconstructs exactly the `(existing,
   candidate)` pair the next setup commit would hand the merge.
3. **Retained argument links**, via the same validator, over committed state.
4. **Handler stream markers**, reported as an advisory ("setup will materialize
   its stream marker; no stored value is lost"), never as a CFC blocker.

**Stated honestly, not faked:**

- **Only the piece's own argument and result documents are enumerated.**
  Derived internal cells and linked child docs are not. Enumerating every
  `(target doc, candidate schema)` pair a setup will record is only knowable by
  running setup, which cannot be done read-only. The check does not pretend
  otherwise: each role reports `pass` / `fail` / `not-applicable` explicitly.
- **`--check` is read-only with respect to the piece, not to the space.**
  Compiling the candidate content-addresses the pattern's source and module
  closure, as any compile does. Those documents are immutable and keyed by
  content, and are exactly what a subsequent real `setsrc` would write. The
  piece is never written; a test asserts the pattern ref and result value are
  unchanged after both a passing and a failing check. Documented on the method
  and in the docs.

## Finding worth a decision (surfaced, not silently fixed)

**The two gate layers disagree about handler streams.** The CFC document merge
exempts a stream slot from its additive-required rule (#4977's fix — a stream
marker is re-materialized by setup, so an old doc lacking one has no value to
preserve and no meaningful `Default<>` to declare). **The pattern-contract
subset proof has no such exemption.** Empirically, adding a handler to a pattern
and running `setsrc` today is refused with
`result.rename: newly required result field has no default`.

This is exactly Berni's framing arriving as a live bug: a newly-required field
on an **output** cell is inherently compatible, because the pattern creates it
fresh — yet the contract layer refuses it.

I did **not** change that rule. Loosening a schema-subset proof is a real
semantic change to a security-adjacent gate and does not belong in a preflight
PR. Instead the check stays truthful: it reports the blocker (because that is
what `setsrc` will do), flags it `streamSlot: true`, and says in the message
that only the pattern-contract proof refuses it and that the CFC layer already
exempts it. A test pins this behaviour and the preflight/enforcement agreement.

Worth a follow-up decision on whether the contract layer should gain the same
stream exemption, or the same output-cell reasoning more broadly.

## Test plan

New:
- `packages/piece/test/pattern-update-check.test.ts` — 13 cases, real runtime,
  real compiles, **CFC enforcement ON** (`enforce-explicit`). Covers:
  compatible-passes; newly-required INPUT with no default (names the field and
  the role); incompatible type mutation; handler-stream advisory; the layer
  disagreement above plus preflight/enforcement agreement field-for-field;
  `--check` mutates nothing (pattern ref and result value unchanged after both
  verdicts); the improved non-check error with the piece left intact; a
  compatible source still applies. Plus five cases driving the real CFC merge
  for the Estuary class: refused without a default, accepted with one, stream
  slot exempt, non-migration merge rejection classified separately, and
  not-applicable when no envelope is stored.
- `packages/cli/test/piece-setsrc-check.test.ts` — 9 cases at the command
  boundary, using the repo's dependency-injection convention: flag
  registration, `--check` routes to the read-only library call, flag-combination
  validation, human report rendering (compatible and incompatible), and
  `reportIncompatibleSetSource` (exit 3, blockers on stderr, `--check` hint,
  and returning null for an unrelated error).

Results — all run, all green:

| suite | result |
|---|---|
| `packages/piece` full (`deno task test`) | 25 passed (293 steps), 0 failed |
| `packages/cli` full (`deno task test`) | 120 passed (144 steps), 0 failed |
| `packages/runner` full (`deno task test`) | 1070 passed (5562 steps), 0 failed |
| `deno check` on all touched files | clean |
| `deno lint` on all touched files | clean (10 files) |
| `deno fmt` | applied |

Two pre-existing `pull-materialization` tests assert on the rejection message
shape; they pass unchanged because the improved error keeps the originating
message as a trailing `Caused by:` line.

The CLI **integration** suite (`integration/integration.sh`, section
`piece-values`) asserted the old raw text `not backward compatible`, which the
improved refusal deliberately no longer prints — CI caught this. Second commit
replaces that with the new contract (exit 3, the offending field, the rule that
refused it, the `--check` pointer) and adds live coverage of the preflight
itself: exit 3 with a named field, the `--json` verdict, that the check mutates
nothing, and that a compatible source passes it. Run locally against a real
toolshed: `Successfully ran CLI piece values integration tests`.

Manual smoke against the same live toolshed, for the record:

```
$ cf piece setsrc --check ... schema-incompatible.tsx ; echo $?
INCOMPATIBLE — the pattern source cannot be applied to piece fid1:5goBVQ….

Checks:
  [FAIL] pattern contract
  [n/a ] CFC document migration (argument) — the argument document carries no stored CFC schema envelope, so the CFC merge does not run for it
  [n/a ] CFC document migration (result) — …
  [ok  ] retained argument links — every durable link already supplied into an argument slot still satisfies the candidate's schema for that slot

Blockers:
  • [pattern-contract] result.value — a schema alternative accepted previously is not accepted by the candidate. This is the pattern's output (result) contract.
3

$ cf piece setsrc ... schema-incompatible.tsx        # no --check
Pattern source cannot be applied to piece fid1:5goBVQ…. The piece was NOT modified.
  • [pattern-contract] result.value — a schema alternative accepted previously is not accepted by the candidate. This is the pattern's output (result) contract.

TIP: Run the same evaluation without attempting the update:
  cf piece setsrc --check --piece fid1:5goBVQ… <main> ...
Add --dangerously-allow-incompatible-schema only if you accept losing the guarantees above.
```

Docs updated in the same change (live-doc rule): `docs/common/workflows/development.md`,
`docs/development/debugging/cli-debugging.md`, `packages/cli/README.md` (JSON
command contract + exit code), and the in-CLI help/examples.

### DIFF
--- a/docs/common/workflows/development.md
+++ b/docs/common/workflows/development.md
@@ -10,6 +10,9 @@ deno task cf check pattern.tsx
 # Deploy
 deno task cf piece new ... pattern.tsx
 
+# Ask whether the update would be accepted, without applying it
+deno task cf piece setsrc ... --piece PIECE_ID --check pattern.tsx
+
 # Update existing (faster iteration)
 deno task cf piece setsrc ... --piece PIECE_ID pattern.tsx
 
@@ -65,4 +68,16 @@ deno task cf piece link ... editor-id/items viewer-id/items
   bypass compilation, normal value validation, or atomic stale-update checks.
   `piece new` accepts the same flag for deploy-script symmetry, but a fresh
   piece has no predecessor schema to compare.
+- `setsrc --check` answers all of the above ahead of time and applies nothing.
+  It runs the same rules the update enforces — the pattern-contract subset
+  proof, the CFC document merge against the schema the piece's documents
+  actually carry, and the retained-link proof — and reports which fields would
+  block the update and why, plus the migration work setup would perform (a new
+  handler stream's marker, for example). It exits `0` when the update would be
+  accepted and `3` when it would not, so a deploy script can gate on it.
+  `--json` (only valid with `--check`) emits the same verdict machine-readably.
+  The check never writes to the piece; compiling the candidate does persist the
+  pattern's content-addressed source, exactly as any compile does. A real
+  `setsrc` that is refused now reports the same field-level reasons instead of
+  a low-level rejection, and leaves the piece on its original pattern.
 - Test one feature at a time
--- a/docs/development/debugging/cli-debugging.md
+++ b/docs/development/debugging/cli-debugging.md
@@ -140,6 +140,12 @@ deno task cf piece get --piece <piece-id> brokenField -i "$CF_IDENTITY" -a URL -
 
 This keeps you working with the same piece instance, preserving any test data you've set up.
 
+If a `setsrc` is refused as incompatible, the error names the field and the rule
+that refused it, and the piece stays on its original pattern. To ask the same
+question before touching the piece, add `--check`: it applies nothing, reports
+which fields would block the update and why, and exits `3` when the update would
+not be possible (`0` when it would). Add `--json` for a machine-readable verdict.
+
 ## See Also
 
 - ./workflow.md - General debugging workflow
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -133,6 +133,12 @@ The supported output switches are:
   JSON stream.
 - `cf piece get` and `cf wish` always return JSON. Their `--json` options are
   accepted, documented no-ops for callers that select JSON explicitly.
+- `cf piece setsrc --check --json` serializes the compatibility verdict for a
+  source update that is not applied. `--json` requires `--check`, because the
+  verdict is the only thing `setsrc` has to serialize. The command exits `0`
+  when the update would be accepted and `3` when it would not; `3` is also the
+  exit code of a real `setsrc` refused as incompatible, so a caller can branch
+  on "incompatible" without parsing text.
 - `cf check --json` compiles without evaluating and prints one object with a
   `files` array. Each entry has the input `path` and the compiled module bodies
   in `output`.
--- a/packages/cli/commands/piece.ts
+++ b/packages/cli/commands/piece.ts
@@ -366,6 +366,13 @@ const PIECE_REGISTRY_LINK_EXAMPLE = [
   ),
   `Link the well-known "pieceRegistry" list to a piece field.`,
 ] as const;
+const SETSRC_CHECK_EXAMPLE = [
+  cliText(`cf piece setsrc --check ${EX_ID} ${EX_COMP_PIECE} ./main.tsx`),
+  `Report whether ./main.tsx could be applied to "${RAW_EX_COMP
+    .piece!}", changing nothing. Exits ${SETSRC_INCOMPATIBLE_EXIT_CODE} if it could not.`,
+] as const;
+const SETSRC_CHECK_FLAG_DESCRIPTION =
+  `Report whether the source could be applied, without applying it. Exits ${SETSRC_INCOMPATIBLE_EXIT_CODE} when it could not.`;
 
 // Enhanced description with workflow tips
 function pieceEnvStatus(): string {
@@ -601,11 +608,7 @@ export const piece = new Command()
     cliText(`cf piece setsrc ${EX_ID} ${EX_URL} ./main.tsx`),
     `Update the source for "${RAW_EX_COMP.piece!}" with ./main.tsx`,
   )
-  .example(
-    cliText(`cf piece setsrc --check ${EX_ID} ${EX_COMP_PIECE} ./main.tsx`),
-    `Report whether ./main.tsx could be applied to "${RAW_EX_COMP
-      .piece!}", changing nothing. Exits ${SETSRC_INCOMPATIBLE_EXIT_CODE} if it could not.`,
-  )
+  .example(...SETSRC_CHECK_EXAMPLE)
   .option("-c,--piece <piece:string>", "The target piece ID.")
   .option(
     "--main-export <export:string>",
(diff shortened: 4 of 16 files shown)


bfollington:
Superseded by #5311.

Rebuilt small on today's main rather than rebased. The reason: `pieceSourceCompatibilityReview()` landed on main after this PR was opened, and it already does what this one set out to build — drives the real `assertPatternSchemasBackwardCompatible`, the real argument validation against the candidate schema, and the real retained-link assertion, collects them into `{schema, argument, retainedLinks}`, and feeds a check-then-confirm flow with a `status: "incompatible"` result. Same design, same "no second copy of the rules" constraint.

It was only ever wired to `changeSource` (detach / restore / follow). `setPattern` — the `setsrc <file>` path — still just asserted and threw whichever rejection fired first, which is exactly the complaint this PR opens with. So the remaining gap was the wiring and the flag, not a new verdict engine.

Landing this as-is would have meant two compatibility verdict systems: the 416-line `pattern-update-check.ts` here with its own vocabulary alongside main's review. **~2,795 lines become ~418.**

What carried over: the `--check` flag and its non-zero exit, the improved refusal that names every reason at once, and the contract that keeps a preflight honest — now pinned by asserting the check's message and `setPattern`'s thrown message are *equal*, rather than that both merely fail.

What did not: the CFC envelope merge check (`cfcSchemaMergeIssue` / `isStreamSlot`). It is genuinely missing from main's review and worth having — but it belongs in the review itself, where `changeSource` gets it too, rather than bolted onto one caller. Filed as a follow-up, and the implementation here is worth lifting when someone does it.

Closing this in favour of #5311; the branch stays for reference.

PR #5064, opened 2026-07-27
feat(lunch-poll): auto-persist generated art — the host's load event replaces the keep button

ubik2:
## What

The generated cuisine thumbnail now syncs to every viewer automatically: when the host's browser decodes a freshly generated data URL, a hidden trigger ``'s `load` event sends the existing admin-gated, idempotent `setOptionImage` handler — the same sanctioned persistence path as the old manual "keep this art" button, minus the click. The keep button is gone.

## Why an event, not a reactive write

The obvious design — the documented idempotent-computed-`.set()` idiom writing the data URL through a card input handle — works in a direct-call test and **silently fails in the real poll**: map-instantiated cards are minted at **user scope** (the map callback reads `me`/`isAdmin`), so writes through any card input handle land in a `user:<did>` instance of the option document that no other viewer is permitted to read (`packages/runner/test/pattern-scope.test.ts` pins this narrowing as intended). Stream sends are the one mutation path that escapes: they resolve value-mode to the real space-bound handler, whose writes are never read-scope re-addressed. The card's comments document this so the next attempt doesn't rediscover it.

## Why a spread cast, not a typed `onLoad` attribute

Pieces store authored source, and every runtime type-checks it at load. A typed `onLoad` attribute requires a `jsx.d.ts` addition — which would make the deployed piece refuse to load (`Failed to load piece — Could not load pattern <hash>#default`) on any server build predating that typing, coupling piece deploys to framework rollout (observed live on rapids). Instead the trigger's load handler is a module-scope bound `handler()` — a storable stream value, where an inline arrow inside an object literal is rejected by the transformer — attached via `{...({ onLoad: bound } as Record<string, unknown>)}`. The renderer wires any `on*` prop whose value is a handler stream, so runtime behavior is identical, and this PR leaves `packages/html` untouched.

## Changes

- **poll-option-card**: hidden data-URL trigger img whose `load` sends `setOptionImage` (module-scope `sendGeneratedArt` handler, spread-cast attach) while `artSyncState` is `"generated"`; keep button removed; `artSyncState` surfaced as `data-art-state` on the card root (DOM hook for integration tests). Replayed loads (reload, second host tab) are absorbed by the handler's idempotence; a deposed-host race by its admin gate.
- **cli**: the multi-user test worker now honors the `fetchMocks` export, mirroring the single-user runner (entries read post-compile, injected via the `patternTest` preset's `fetch`).
- **tests**: `art-sync.test.tsx` drives the trigger's `onLoad` stream (the VNode harness has no DOM; the browser event and the send reach the identical handler); `poll-option-card.test.tsx` covers trigger render/payload and the non-admin gate; **new `multi-user-art.test.tsx`** proves the point of the feature — Bob's runtime reads the exact data URL Alice's client persisted, across two worker-isolated runtimes.
- **docs**: `DEPLOY-AND-SHARE.md` — stale "no per-option AI work" claim fixed (stale since #4920); rapids canonical-piece pointer re-established (fresh piece after `cf piece recreate-root`; the old empty piece's stored source no longer compiles); build-compatibility note documenting the deliberate cast.

## Verification

- 79/79 lunch-poll pattern tests; `cf check` clean; `deno fmt --check` clean; `packages/html` 43/43 (now untouched); `packages/cli` green except a pre-existing FUSE-daemon help-text failure that also fails on clean main
- **Live, real browser**: on a local server, host joins → adds option → real `/api/ai/img` generation → the real `load` event fires → auto-persist confirmed by reading the option's webp data URL from a separate CLI runtime
- **Live, rapids prod**: the cast-version source `setsrc`'d onto the canonical `team-lunch` piece loads and renders for a logged-in user on rapids' current build (which lacks any `onLoad` typing)

## Follow-up

No CI test fires a real browser `load` event (only click events are exercised); the flow is verified manually live, but one browser integration assertion on the lunch-poll flow would pin it.


Hixie:
Drive by comment: My agent just told me that this PR persists a data: URL rather than a FabricBytes; if that's accurate, we should prefer FabricBytes for efficiency.

PR #5066, opened 2026-07-27
docs(plans): verb contract — discovery is read depth, naming is slugs

bfollington:
## Why

The verb-contract plan is explicitly a plan of record — "keep current as work
proceeds: check off exit criteria, record scope changes" — and #5006 already
amended it post-merge. This is a second review pass, and it lands one
principle in two places the plan had not yet applied it:

**A pattern exposes its data and its verbs. Everything else is a client
projection.**

The plan already holds this for identity (decision 6: patterns return
references, clients render fids and paths) and for verb discovery, where
`cf piece verbs` (#4993) derives the listing from the durable schema through
the same classification `cf piece call` resolves through — nothing authored,
and the listing and the dispatcher cannot disagree. Two surfaces were still
on the other side of that line.

**Discovery.** A2 asks the board to author and maintain a
reference-plus-summary row per topic. But the 300k-token live-board read that
motivated it was not caused by a missing index — `crossrefs` already carries
every `topic` reference; the survey data is there. It was caused by a reader
that expands every reference it crosses
(`packages/patterns/topics/main.tsx:70-78`). That is a read-depth defect, and
an index does not fix it — it routes around it for one blessed path while
every other read stays unbounded. It also costs every pattern the authoring
work separately, and has to be kept in step with the children by hand.

**Naming.** The plan deferred `@name` bindings because they overlap with
slugs (decision 7), which is right — but it left naming looking deferred
wholesale. Slugs already work: a doc at an address derived deterministically
from `(space, slug)` holding a write-redirect link
(`packages/runner/src/slugs.ts`), and `cf piece new --slug` already assigns
one at creation (`packages/cli/commands/piece.ts:472`). Applying the same
affordance to the reference a verb returned is a small CLI change that needs
no runtime and no pattern change, and it means no verb ever mints, stores, or
returns a name.

Also names #5059 in the write-storm gate, which currently says decision 3 is
"not yet satisfiable" and asks for a rehearsal before any live `setsrc`.

## What Changed

Docs only; no code.

**`pattern-verb-contract.md`**

- Discovery rewritten: the unbounded read is diagnosed as expansion depth,
  discovery becomes a generic client read-depth control composed with
  `--include-ids`, and the pattern-authored index is explicitly rejected with
  its reasons. Acceptance restated in terms of the read, not the index.
- New **Naming** subsection under Client surface: `--slug` on `piece call`,
  why assignment stays client-side (`setSlugLink` is a client flow,
  `packages/piece/src/slugs.ts:23`; the pattern-facing API exposes no
  entity-id addressing; a pattern writing a computable space-global address
  makes slug namespaces squattable — a CFC question, not a plumbing gap), and
  `@name` still deferred for its original reason.
- Two decisions added to "worth recording", staging steps 2 and 6 updated,
  Prior art notes the board-index ask is answered more generically than asked.

**`pattern-verb-contract-implementation.md`**

- Governing decisions 8 and 9; A2 struck from the issue table and replaced by
  F4 (read-depth control), plus F5 (`--slug` on `piece call`).
- WS-A loses the index bullet and its test, drops S–M → S, and hands the
  bounded-survey exit criterion to F4. WS-F gains F4 and F5 with exit
  criteria.
- Non-goals clarified: naming is not deferred, a *second* naming system is.
- Risks: #5059 named as the preflight half of the write-storm gate — it
  answers whether a source can be applied before attempting it, driving the
  real rules in dry-run rather than reimplementing them; it does not measure
  commit rates, so the rehearsal stands.

## Test plan

- `deno task check-docs plans` — passes (1 typed block in scope).
- `docs/` is excluded from `deno fmt`; prose hand-wrapped at 80 columns.
- Citations verified against source at these revisions: `assignSlug`
  (`packages/piece/src/slugs.ts:15`), `setSlugLink` (`:23`), `--slug` on
  `piece new` (`packages/cli/commands/piece.ts:472`), slug address derivation
  (`packages/runner/src/slugs.ts`), `crossrefs` expansion
  (`packages/patterns/topics/main.tsx:70-78`).

## Open to argument

- **F4's shape is unspecified on purpose.** "Stop at references instead of
  expanding them" is the requirement; whether that is a depth integer, a
  no-expand flag, or path-selected is a CLI design question I have not
  prejudged. If it turns out materially harder than the authored index, that
  is the strongest argument for keeping A2 and worth hearing.
- **F5 assumes slug assignment can follow a returned reference in the same
  invocation.** It depends on Phase 4 and nothing else, but if the reference
  is not resolvable to a cell at that point client-side, the sequencing needs
  another look.

---
## Summary by cubic
Updates the verb-contract docs to make discovery a client read-depth control and naming a client-assigned slug, so patterns only return references and never author indexes or names. Retires the board-authored index (A2), introduces F4 (read depth) and F5 (`--slug` on `piece call`), and names `#5059` as the `setsrc` preflight in the write-storm gate.

- **Refactors**
  - Discovery: replace the pattern-authored index with a generic read-depth control (F4); `crossrefs` remains the UI graph; acceptance is one bounded read.
  - Naming: add `--slug` on `piece call` (F5) to name the returned reference; pattern-side naming stays out; `@name` remains deferred.
  - Plan updates: add decisions 8–9; retire A2; move the survey exit criterion from WS-A to F4; tighten WS-A scope and clarify non-goals.
  - Risks: include `#5059` (`cf piece setsrc --check`) as the preflight for the write-storm gate.

<sup>Written for commit 004195a36c5f21b888d2e74606abd54add5ae026. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5066?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/plans/pattern-verb-contract-implementation.md
+++ b/docs/plans/pattern-verb-contract-implementation.md
@@ -14,6 +14,12 @@ post-merge review amendments (decisions 5–7: attribution via CFC provenance,
 patterns return references while clients render identity, `@name` deferred);
 index and fixture wording follows the references model.
 
+**Amended 2026-07-27** by a second review pass, extending decision 6 to
+discovery and naming (decisions 8–9). A2's pattern-authored index is retired in
+favour of a generic client read-depth control (F4), and `--slug` on
+`piece call` becomes the naming affordance (F5). Risks names #5059 as the
+`setsrc` preflight the write-storm gate wants.
+
 ## Governing decisions
 
 Made 2026-07-24, shaping everything below:
@@ -33,12 +39,24 @@ Made 2026-07-24, shaping everything below:
    parallel `actor` field. `topics.agentName` stays as an interim atomic
    argument until trusted `cf` ingress can mint and propagate general
    `AgentActor` provenance.
-6. **Patterns return references; clients render identity.** Discovery indexes
-   carry stable child references and summaries. Fid/path annotation belongs to
-   the CLI, which can see backing cell identities.
+6. **Patterns return references; clients render identity.** Fid/path
+   annotation belongs to the CLI, which can see backing cell identities.
 7. **Client-local `@name` bindings are deferred.** Their overlap with
    fabric-side slugs needs a separate naming design and is not required by this
    plan.
+
+Added 2026-07-27, extending decision 6 to the two surfaces it had not yet
+reached:
+
+8. **Discovery is a client read-depth control, not a pattern-authored index.**
+   The parent already links its children; the 300k-token board read was a
+   reader expanding every reference it crossed. Fixing depth once in the client
+   gives every pattern bounded discovery with nothing to author or keep in
+   step. A2 is retired and replaced by F4.
+9. **Naming is slugs, assigned by the client.** `--slug` on `piece call` names
+   the reference a verb returned, generalizing `piece new --slug` (F5). No verb
+   mints, stores, or returns a name; pattern-side assignment is a CFC
+   namespace-authorization question, out of scope here.
 
 ## Non-goals
 
@@ -51,6 +69,8 @@ Named so their absence reads as intent, not oversight:
   delegated agent, but does not deliver delegation.
 - **Client-local aliases** — `@name` / `cf bind` are deferred pending a
   separate comparison with existing slugs and configured host/space addressing.
+  Naming itself is *not* deferred: slugs already work, and F5 applies them to
+  returned references. What is deferred is a *second* naming system.
 - **Cross-space effect atomicity** — the spec's I11 gap stands; the guarantee
   is same-space, which covers `topics`.
 - **Batch / session mode** for the CLI — orthogonal call-cost work, linked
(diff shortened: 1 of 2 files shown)


bfollington:
Withdrawing this after further review discussion. The central claim — retire A2 in favour of a generic read-depth control — didn't hold up: a pattern-side summary projection and a caller-supplied read schema are the same read model, and an author picking a good default projection is ordinary API design, not a contract violation. What survived the discussion is adjacent, non-blocking CLI work rather than a plan correction: exposing `asSchema` on `cf piece get` (the CLI already uses it internally, e.g. `piece-render.ts`), cursor-based bounds on large array reads (rule 5 applied to reads — offsets are positional addressing), and naming the ceiling that every full-board read shares regardless of projection, with selective/search queries as a separate tier. None of that needs to interrupt the current workstream; it can arrive as its own thread when the board's scale demands it.

PR #5071, opened 2026-07-27
fix(cf-harness): use the Responses API so tools and reasoning coexist

willkelly:
Reverts #4998 and fixes the underlying problem instead of working around it.

## Why the workaround wasn't the right end state

#4998 restored tool calls by pinning `reasoning_effort: "none"` — which turns reasoning off entirely on reasoning-first models. It also had a latent break: `gpt-5-mini`/`gpt-5-nano` **reject** `"none"` (they accept only `minimal`/`low`/`medium`/`high`), so any run pointed at those models would 400 before its first turn — the same failure mode the PR set out to fix.

The restriction is OpenAI's, not our gateway's, and the error names the fix:

```
Function tools with reasoning_effort are not supported for gpt-5.6-terra in
/v1/chat/completions. To use function tools, use /v1/responses or set
reasoning_effort to 'none'.
```

## What changed

Gateway turns now go to **`/v1/responses`**, where function tools and reasoning coexist. Encrypted reasoning items are requested (`include: ["reasoning.encrypted_content"]`, `store: false`) and replayed on later turns through the existing `providerContinuation` seam — the same mechanism the Codex client already uses.

The Responses wire mapping is now **shared** in `model/responses-protocol.ts` instead of duplicated. The Codex client keeps its exact wire behavior; its default instructions are now passed explicitly rather than being baked into the shared mapper, so a gateway run without a system prompt no longer gets one invented for it.

**Chat Completions is deliberately retained** where Responses cannot serve the turn:

- provider-native tools (`google_search`) have no Responses equivalent on this gateway
- Vertex-backed models are not translated to the Responses schema — the gateway answers 500 with an empty body

The `web_search` subagent profile is exactly that combination: it overrides the model to `gemini-3.5-flash` *and* uses `google_search`. A blanket switch would have silently broken it; the existing test caught this. Selection is one documented predicate, `usesResponsesApi()`.

Run reports now record gateway attempts for both operations, so Responses turns aren't dropped from diagnostics (an existing artifacts test caught that too).

## Stops using gpt-5-mini

**Scope note:** this PR is now cf-harness-only. The model-registry half — the `gpt-5.6` registrations, the shared LLM defaults, and their test — was split into a follow-up, because it sets a product-wide default onto a name whose transport depends on whether `CFTS_AI_LLM_OPENAI_API_KEY` is set, and whose multi-turn tool path is broken on both routes (direct/Responses fails under Zero Data Retention, measured 0/5; gateway/`.chat` hits the tools-plus-reasoning 400). The follow-up carries the prerequisite `addModel` `providerOptions` fix. `DEFAULT_MODEL = gpt-5.6-terra` stays here and resolves via the gateway registry that commontoolsinc/infra#87 already shipped.

## Verification

`deno task test` in `packages/cf-harness`: **512 passed, 0 failed.**

Live against the stage gateway, a two-turn function-tool loop per model:

| model | tool call | encrypted reasoning returned | replayed on turn 2 |
|---|---|---|---|
| `gpt-5.6-sol` | ✅ `read_file` | ✅ 1 item | ✅ |
| `gpt-5.6-terra` | ✅ `read_file` | ✅ 1 item | ✅ |
| `gpt-5.6-luna` | ✅ `read_file` | ✅ 1 item | ✅ |

Reasoning is adaptive: a trivial prompt returns zero reasoning items even at `effort: high`-eligible models, so the live check uses a prompt that actually requires thought. That is model behavior, not dropped passthrough — confirmed by driving the raw endpoint at `effort: high` (27 reasoning tokens, `encrypted_content` present) versus default (none).

## Note on the test diff

The prompt-loop/artifacts/cli suites contain 100+ gateway stubs written in the Chat Completions shape. Those fixtures describe *intent* ("the model calls this tool"), so rather than restate them, `test/support/responses-fixture.ts` translates them to the Responses wire format and projects outgoing requests back for content assertions. Tests that assert on the Responses format itself — reasoning passthrough, continuation replay, API selection — are new and build their bodies directly.

---
## Summary by cubic
Switch `cf-harness` to the OpenAI Responses API for `gpt-5.6-*` so function tools and reasoning work together; keep Chat Completions for provider-native tools and non-OpenAI models. Set the CLI default model to `gpt-5.6-terra`.

- **Bug Fixes**
  - Route `gpt-*` turns to `/v1/responses`; request encrypted reasoning and replay via `providerContinuation` (stateless: `store: false`, no `previous_response_id`).
  - Keep Chat Completions for provider-native tools and Vertex-backed models; tolerate turns that omit `content`.
  - Throw a named error if `gpt-*` is used with provider-native tools.
  - On resume, drop mismatched continuation on the gateway; Codex still throws.
  - Prefix transport errors with the operation and classify timeouts via the operation-neutral tail.
  - Record gateway attempts for both APIs and preserve which served the turn.
  - Bound `prompt_cache_key` to 64 chars.

- **Refactors**
  - Share Responses wire mapping in `model/responses-protocol.ts`; stop recording `responseId` in continuation state.
  - Gateway client supports both `/v1/chat/completions` and `/v1/responses` behind `usesResponsesApi()`.
  - Expand tests for both endpoints, Chat-path coverage, Responses error branches, and the fixture adapter.
  - Move model registry and shared-default changes to a follow-up PR; this PR updates only the CLI default to `gpt-5.6-terra`.

<sup>Written for commit 00e3cd986d4197588f37faf88d2faa7821608632. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5071?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Coverage

The cf-harness regression the ratchet flagged is fixed rather than accepted: moving `gpt-*` to the Responses API left the Chat Completions mappers covered only by the single web_search test, so this adds direct coverage for that path (transcript mapping, array-content flattening, native tool results, model discovery) plus the Responses error branches. That removes 68 uncovered lines locally against the 55 reported; CI now measures packages/cf-harness at 3797 against a 3814 baseline, i.e. below where it started.

The toolshed delta is gone entirely now that the registry half is split out; that subprocess-isolated registration test travels with it.


seefeldb (commented):
cf-review via Claude Opus 5, on behalf of @seefeldb.

Scope: 19 files, +2187/−440. Coverage: deep-read of all six source files and the new tests, sampled the fixture-translated test diffs. Checks run locally on `cf-harness-responses-api`: `deno task test` in `packages/cf-harness` (**539 passed, 0 failed**), toolshed `routes/ai/llm/models.test.ts` (pass), `deno fmt --check` + `deno lint` on all 19 changed files (clean), plus one scratch repro noted below.

Solid change overall: it replaces a workaround with the actual fix, and it reuses the Responses machinery from #4931 rather than forking it.

## Requested change: make `gpt-5.6-terra` the default

Please point the defaults at the balanced tier rather than the fast one:

- `packages/cf-harness/src/cli.ts:112` — `DEFAULT_MODEL` is still `gpt-5.5`.
- `packages/llm/src/types.ts:18-19` — `DEFAULT_IFRAME_MODELS` / `DEFAULT_GENERATE_OBJECT_MODELS` currently point at `openai:gpt-5.6-luna`.

Both should be `gpt-5.6-terra`.

## Relationship to the Codex-subscription Responses work (#4931)

Worth stating explicitly, since it is the main thing a reader needs to hold: #4931 built the entire Responses wire mapping inside `openai-codex-responses.ts`, including the `providerContinuation` seam that carries encrypted reasoning across a tool round-trip. This PR lifts that into `model/responses-protocol.ts` and parameterizes `providerId`, `label`, and `defaultInstructions`, then reuses it for the gateway.

The subscription path comes through intact — its 54 tests pass with one line deleted, and the only real behavioral deltas are the explicit default instructions and dropping the write-only `responseId` from continuation state (legacy artifacts covered at `test/cli.test.ts:4060`). Cross-provider replay fails safe: `continuationOutput` returns `[]` on a `providerId` mismatch rather than replaying a blob the other side cannot decrypt.

## Findings

**1. Resuming a gateway run with `--model` now hard-fails.** `responses-protocol.ts:57` throws when a continuation's `sourceModel` differs from the requested model. That rule came from Codex, where the CLI pins the provider on resume (`cli.ts:2254`) — but it does not pin the model: `cli.ts:2314` resolves `parsed.model ?? artifacts.runState.model`. Before this PR gateway assistant messages carried no continuation, so `--resume-run X --model <other-model>` worked; now it dies on the first turn with a protocol-layer message. Verified with a scratch call into the real mapper:

```
THROWS: openai-compatible-gateway continuation model gpt-5.6-terra does not match requested model gpt-5.6-sol
```

Either is fine, but pick one deliberately: pin the model on resume with a CLI-level error mirroring the provider-mismatch check, or drop mismatched continuation items instead of throwing — the reasoning is optional and the run can continue without it.

**2. The Chat Completions fallback lost its safety valve.** `reasoning_effort: "none"` is gone from the request type entirely, and `usesResponsesApi` selects on `model.startsWith("gpt-") && nativeModelToolIds.length === 0`. So a `gpt-*` model plus any native tool routes to Chat Completions with function tools *and* reasoning defaulted on — the exact 400 this PR exists to fix. It is unreachable today only because `WEB_SEARCH_SUBAGENT_MODEL` happens to be `gemini-3.5-flash` (`contracts/subagent.ts:19`); the predicate encodes that coincidence as a rule. Same hazard for an OpenAI reasoning model not named `gpt-*`. Cheapest fix: throw a named error for the `gpt-*` + native-tools combination instead of silently routing into the failure, and pin the web_search model invariant in a test.

**3. The new toolshed test re-forks the isolated-subprocess helper.** `routes/ai/llm/models.test.ts:39` calls `Deno.Command(Deno.execPath())` with `-A` and the workspace config, without lock isolation. The canonical home is `runDenoCommandWithTemporaryLock` in `packages/test-support/src/isolated-deno.ts:60`, and the `isolated-test-processes` guidance is that a verification test must not be able to touch `deno.lock`. The subprocess approach itself is right — the in-process import would break `generateObject.test.ts`'s "MODELS is empty" contract, and the comment explains that well — it just needs to go through the helper.

**4. Stale live doc.** `packages/cf-harness/docs/IMPLEMENTATION_PROFILE.md:42` still reads "Model gateway: OpenAI-compatible chat completions", which is now wrong for the majority path. One line, in this diff.

**5. Nit — the pinned Codex contract now lives in shared code.** `docs/OPENAI_CODEX_PROTOCOL.md` says to update it together with fixtures when either upstream changes, but does not mention that the mapping it pins is now also driven by gateway requirements. A future gateway-motivated edit to `responses-protocol.ts` silently edits that contract; a sentence pointing at `model/responses-protocol.ts` would close the gap.

**6. Nits.** `OpenAIResponsesRequest.reasoning` is declared and never set. `chatViewOfRequest` drops `function_call` items (`test/support/responses-fixture.ts:58`), so translated assertions about outgoing assistant tool calls project to nothing — fine, since the dedicated Responses tests cover that directly, but worth knowing when someone next edits a translated fixture.

The fixture-translation module reads well, and the test that asserted `reasoning_effort: "none"` was rewritten to assert the new intent (`include: ["reasoning.encrypted_content"]`, `store: false`) rather than merely deleted.

## Questions

1. The live verification table covers `gpt-5.6-{sol,terra,luna}`, but the current cf-harness default is `gpt-5.5` and `usesResponsesApi("gpt-5.5", [])` is `true` — that is what every default run takes today. Was it exercised on `/v1/responses` with `include: ["reasoning.encrypted_content"]`? (Moot if the default moves to terra as requested above, but worth confirming for anyone pinning 5.5.)
2. The model-registry half lands on a different transport than the half that was verified. `openai:gpt-5.6-*` are registered against `createOpenAI({apiKey})` — api.openai.com via the AI SDK, not the gateway — and `DEFAULT_GENERATE_OBJECT_MODELS` makes that the default for every pattern `generateObject()` (`packages/runner/src/builtins/llm.ts:1392`). Was that path smoke-tested against the real OpenAI API? Also note the retired `gpt-5-mini` was registered `reasoning: false` while the new default is `reasoning: true`, which is a latency and token change for every default `generateObject` in the product.

## Follow-ups (not for this PR)

- toolshed has the same bug one layer up: `routes/ai/llm/models.ts:583` registers gateway models with `gatewayProvider.chat`, explicitly commented "force /v1/chat/completions (not /v1/responses)", and `generateText.ts:226` does pass client-side function tools. Once infra#87 exposes the gpt-5.6 family through the gateway, `gateway:gpt-5.6-*` plus tools hits the identical 400 with no mitigation.
- `addModel` destructures `providerOptions: _` and discards it (`models.ts:93`), so the existing `gpt-5-thinking` `reasoningEffort: "high"` has never taken effect and there is currently no way to pin effort per model in the registry.
- `docs/plans/cf-harness-codex-subscription-auth.md` still describes the Chat Completions conversion as current and looks fully executed — a candidate for archival to `docs/history/`.

willkelly:
Thanks — all six findings addressed, plus the requested default change. Both suites green (cf-harness **542 passed**, toolshed **94 passed**).

## Requested change

`DEFAULT_MODEL` (`cli.ts:112`) and both `packages/llm/src/types.ts` defaults now point at `gpt-5.6-terra`. Verified end to end with no `--model` flag: Responses path, HTTP 200, encrypted reasoning on both turns, correct answer. Two `cli.test.ts` assertions pinned the old default and were updated; the explicit-`--model` precedence test at `:436` was correctly unaffected.

## Findings

**1 — resume with `--model`.** Fixed as *drop*, made explicit rather than inherited. `continuationOutput` now takes a `ContinuationModelMismatch` policy: Codex keeps `throw` (it pins the provider on resume, and `openai-codex-responses.test.ts` asserts the throw precedes credential resolution), the gateway uses `drop`. The reasoning is bound to the model that produced it and can never be replayed to another one, so for a path that deliberately allows changing models mid-run, losing continuity beats failing the turn. Covered by a test asserting no `reasoning` item survives into the request.

**2 — the lost safety valve.** You're right that the predicate encoded a coincidence as a rule. `gpt-*` plus native tools now throws a named error before routing, and `WEB_SEARCH_SUBAGENT_MODEL.startsWith("gpt-") === false` is pinned by a test, so the guard's unreachability is asserted rather than assumed. I removed the `usesResponsesApi("gpt-5.6-sol", ["google_search"])` line from the routing test since it would now imply a routing outcome that no longer happens.

**3 — subprocess helper.** Moved to `runDenoCommandWithTemporaryLock`. Confirmed `deno.lock` is untouched after a run.

**4, 5, 6.** `IMPLEMENTATION_PROFILE.md:42` now describes the split transport; `OPENAI_CODEX_PROTOCOL.md` notes the pinned mapping now lives in the shared `responses-protocol.ts` and that gateway-motivated edits there change this contract; the never-set `OpenAIResponsesRequest.reasoning` is removed. The `chatViewOfRequest` function_call note is accurate — leaving as is, since the dedicated Responses tests cover outgoing tool calls directly.

## Questions

**1 — was `gpt-5.5` exercised on `/v1/responses`?** Yes, confirmed live before the default moved: ops `['responses']`, status 200, correct answer. Zero encrypted reasoning items on that run, but that is the adaptive behaviour rather than a plumbing gap — a trivial prompt returns none, and a reasoning-heavy one returns them on every turn. Anyone pinning 5.5 gets the same path as terra.

**2 — was the direct-OpenAI registry path smoke-tested?** No, and I want to be explicit about that. `openai:gpt-5.6-*` registers against `createOpenAI({apiKey})` — api.openai.com via the AI SDK, not the gateway — and I have no `CFTS_AI_LLM_OPENAI_API_KEY`. Everything in my verification table went through the gateway. So the half of this PR that changes `generateObject`'s default is unexercised against its real transport.

Your second point there deserves more weight than a note: the retired `gpt-5-mini` was registered `reasoning: false`, and every gpt-5.6 tier is `reasoning: true`. That makes **every default `generateObject()` in the product a reasoning call** — a latency and token-cost change for pattern authors who never opted into it, on a transport nobody has smoke-tested here. It is a defensible default, but it is a product change rather than a rename, and I would rather it be an explicit decision than a side effect of dropping mini. Happy to gate the registry half behind a separate PR with a real key behind it if you would prefer this one stay cf-harness-only.

## Follow-ups

Noted, not in this PR — the `gatewayProvider.chat` bug at `models.ts:583` is the same defect one layer up and becomes live the moment infra#87's gpt-5.6 models are reachable through the gateway with tools; the discarded `providerOptions` at `models.ts:93` means `reasoningEffort` has never worked; and `docs/plans/cf-harness-codex-subscription-auth.md` looks ready for `docs/history/`.

willkelly:
Smoke test done against the real OpenAI API (thanks for the key — used only for this, never written to the repo, a commit, or this thread; the local copy is destroyed and it is safe to deactivate).

## The good news answers your concern directly

`addModel` registers static providers with `provider(modelName)` — the AI SDK *default* — and for `@ai-sdk/openai@4` that default is **`/v1/responses`**, not chat completions. Captured live:

```
endpoints hit: ["https://llm.stage.commontools.dev/v1/models",
                "https://api.openai.com/v1/responses"]
```

So the direct-OpenAI registry path is already on Responses, and the 400 hazard I fixed for the gateway does not exist there. That is also why `models.ts:583` has to say `.chat` explicitly for the gateway models.

**The default this PR sets works.** `DEFAULT_GENERATE_OBJECT_MODELS` → `generateObject` with `openai:gpt-5.6-terra`, reasoning-heavy prompt, **3/3** correct (`{"sum":49,"isPrime":false}`).

## The bad news, which is not this PR's

Multi-turn *tool* loops on that path fail deterministically once the model actually reasons:

```
AI_APICallError: Item with id 'rs_0f4c5e77…' not found. Items are not persisted
for Zero Data Retention organizations. Remove this item from your input and try again.
```

The AI SDK sends `store: true` and references reasoning items **by id** on the next turn. Our OpenAI org is Zero Data Retention, so nothing is persisted. Measured on a reasoning-heavy prompt:

| model | result |
|---|---|
| `gpt-5.6-terra` (new default) | 0/5 — 5/5 ZDR failures |
| `gpt-5-mini` (old default) | 0/5 — 5/5 ZDR failures |

Identical for both, so **not a regression from this PR** — and it hits `generateText.ts:226`, the client-side-tools path you flagged. My first two runs disagreed (one pass, one fail) purely because a trivial prompt emits no reasoning item, which makes this look flaky rather than broken.

It is the same class of problem cf-harness solves on the gateway path: send `store: false` with `include: ["reasoning.encrypted_content"]` and replay the encrypted item instead of referencing a stored id.

**Your two follow-ups turn out to be one bug.** That fix is currently impossible to express, because `addModel` destructures `providerOptions: _` and discards it (`models.ts:93`) — the same reason `gpt-5-thinking`'s `reasoningEffort: "high"` has never taken effect. Provider options have to start being honoured before the ZDR fix can be applied.

## On the reasoning-cost point

Worth noting my earlier framing was slightly off: `gpt-5-mini` was registered `reasoning: false`, but it *is* a reasoning model and reasons in practice — that is exactly how it produces the ZDR failure above. So the capability flag was already wrong, and the real delta from this PR is smaller than "non-reasoning becomes reasoning". Still a latency and token change worth a deliberate decision, and the offer stands to split the registry half out if you would rather this PR stay cf-harness-only.

Written up with repros in `model-issues.deleteme.org` alongside the two Vertex-path defects. CI is green (43 success, 4 skipped).

seefeldb:
Hmm, this is what ChatGPT said about ZDR and reasoning on:

Then yes—fully supported in a ZDR org.

Use:

Responses API

gpt-5.6-sol, gpt-5.6-terra, or gpt-5.6-luna

reasoning: { effort: "medium" } (or another level)

Function tools

store: false—enforced automatically under ZDR


During each tool-call round trip, resend:

1. The complete previous response.output, including its opaque encrypted reasoning items.


2. Your matching function_call_output with the same call_id.



Don’t rely on previous_response_id, since the response isn’t stored. The encrypted reasoning items are returned by default specifically to support this stateless/ZDR handoff. You won’t receive raw chain-of-thought; you get opaque reasoning state that lets the model continue reasoning across tool calls.

That gives you exactly: reason → tool call → tool result → continue reasoning → possibly another tool → final answer, all through Responses and compatible with ZDR. Reasoning guide · Data controls

PR #5073, opened 2026-07-27
fix(runner): bind the unwired Fabric value classes into the builder

danfuzz:
## Summary

Three Fabric value classes are declared as pattern-visible constructors in
`packages/api/index.ts` but have **no runtime binding** in the runner's builder
exports. Pattern code referring to them type-checks against the declaration and
then fails at execution, because the value behind the name is `undefined`.

This binds all three, and adds a regression test covering the seam that let them
drift apart.

## The defect

`packages/api/index.ts` declares the pattern-visible surface with
`export declare const`, e.g. at `:159`:

```ts
export declare const FabricBytes: FabricBytesConstructor;
```

That surface reaches patterns via `commontools.d.ts` (a symlink to
`api/index.ts`), fed to the sandbox by
`runner/src/sandbox/runtime-modules.ts:20-30`. But the runtime `commonfabric`
module a pattern actually imports is literally `createBuilder().commonfabric`
(`runtime-modules.ts:39-54`) — one path, no second source — and the block in
`runner/src/builder/factory.ts:293-299` that backs those declarations, labelled

```
// Fabric value classes -- runtime values backing the type declarations
// in api/index.ts. Enables `new FabricEpochNsec(...)` and `instanceof`
// checks in patterns.
```

wired only five of the eight declared classes: `FabricInstance`,
`FabricPrimitive`, `FabricEpochNsec`, `FabricEpochDays`, and `FabricHash`.
Missing were **`FabricBytes`** (`api/index.ts:159`), **`FabricLink`** (`:140`),
and **`FabricSpecialObject`** (`:44`) — absent from the export object and from
the corresponding type list at `builder/types.ts:446-456`.

So the declaration and the runtime disagreed, and TypeScript had no way to
notice: a `declare const` asserts that a value exists without any obligation to
produce one.

## How it was found

By instantiating the real builder and reading back `typeof` on each export — an
empirical probe rather than a reading of the source:

| Declared in `api/index.ts` | `typeof` on the builder surface |
|---|---|
| `FabricSpecialObject` (`:44`) | **`"undefined"`** |
| `FabricInstance` (`:59`) | `"function"` |
| `FabricPrimitive` (`:70`) | `"function"` |
| `FabricEpochNsec` (`:87`) | `"function"` |
| `FabricEpochDays` (`:102`) | `"function"` |
| `FabricHash` (`:122`) | `"function"` |
| `FabricLink` (`:140`) | **`"undefined"`** |
| `FabricBytes` (`:159`) | **`"undefined"`** |

`FabricBytes` was the case that prompted the investigation; auditing every
`export declare const Fabric*` against the builder's export block turned up the
same mismatch for the other two. The failure is a real one, not a theoretical
gap — running the new test against the unfixed source produces, for example,
`TypeError: BoundFabricLink is not a constructor`.

Whether those two were defects or deliberate was genuinely unclear from the code
— an unbound class is indistinguishable from a class nobody intended patterns to
name. Confirmed with the API's owner that both are defects, with an important
distinction between them, quoted because it shapes the fix:

> `FabricLink` _should_ be constructible, and `FabricSpecialObject` should be
> usable for `instanceof`. (it's an abstract class and so isn't supposed to be
> directly constructed, even if that isn't being explicitly prevented.)

**So `FabricSpecialObject` is bound so that `instanceof` works in pattern code,
not so that anything can construct it.** It is abstract. Reading "bound ⇒
constructible" would misread both the fix and the test.

Binding an abstract class here is not a new precedent, incidentally:
`FabricInstance` (`data-model/src/interface.ts:63`) and `FabricPrimitive`
(`:110`) are *both* `export abstract class`, and both have been in the builder's
export block all along — as the block's own comment says, the bindings exist to
enable "`instanceof` checks in patterns," not construction alone.
`FabricSpecialObject` is the same shape as two classes already there.

Worth noting that these classes are otherwise fully built and wired:
`FabricBytes` is exported from `@commonfabric/data-model/fabric-primitives` —
the module `factory.ts` already imports its sibling primitives from — and is
listed in that module's `CODEC_CLASSES`, so it is wire-serializable today. These
are missing lines in the builder, not unfinished features.

## The fix

Binds the three missing classes in the runner's builder — each added to the
relevant `@commonfabric/data-model` import in `factory.ts`, to that block's
exported object, and to the matching type list in `builder/types.ts`. They come
from three different modules:

| Class | Module |
|---|---|
| `FabricSpecialObject` | `@commonfabric/data-model/fabric-value` |
| `FabricLink` | `@commonfabric/data-model/fabric-instances` |
| `FabricBytes` | `@commonfabric/data-model/fabric-primitives` |

Only `fabric-instances` is a new import for `factory.ts`; it already imported
from the other two.

`FabricLink` needed no implementation work: `fabric-instances/FabricLink.ts:173`
already asserts `FabricLink satisfies ApiFabricLinkConstructor`, a compile-time
check against the *declared* constructor contract. The implementation was
already pinned to the pattern-facing contract — only the binding was missing.

Both class lists — the builder's exports and the type list in `types.ts` — are
now ordered to match the declaration order in `api/index.ts`, so the two sides
can be compared directly by eye.

## The regression test

Nothing anywhere asserted that `api/index.ts`'s `declare const`s have runtime
bindings. `runner/test/factory-input-types.test.ts`,
`builder-value-equal.test.ts`, and `cfreg-builder-identity.test.ts` all miss it;
`packages/api/index.test.ts:20-30` actually *documents* the gap without catching
it, faking `{ length: 3 } as unknown as FabricBytes` with the comment "the
constructor lives in the runtime, not here" — precisely the assumption that
doesn't hold.

**That fake is deliberately left alone**, and it is worth saying why so nobody
later "fixes" it on a false premise. Its comment stays true after this change,
and it is a type-level pin serving a different purpose — honest about being one.
In particular, the tempting justification *"the test can't use a real
`FabricBytes` because `api` can't depend on `data-model`"* is **not available**:
`packages/api/cfc.ts:9` already imports `@commonfabric/data-model/deep-freeze`,
so that dependency edge exists. The reason to leave it is its purpose, not a
constraint.

That absence is how three classes drifted at once, so the test targets the seam
rather than any one name. It **derives** the expected class list from
`types/commonfabric.d.ts` — the symlink to `api/index.ts` that the sandbox
actually hands a pattern as its view of `commonfabric` — rather than hard-coding
names. A future declared-but-unbound class therefore fails immediately, instead
of waiting for someone to remember to extend a list.

That file is read through `StaticCacheFS`, the supported accessor for the
pattern-visible type surface, rather than scraped from a relative source path;
`ts-transformers/test/commonfabric-test-types.ts` already loads it the same way.
So the test derives from the artifact patterns are actually given, not from a
file that happens to back it.

**Guarding against a test that passes without asserting anything.** The
per-class checks are driven by the derived list and look their expectations up
in a table of the expected `data-model` classes, so a name present in only one
of the two would compare `undefined` against `undefined` and assert nothing. A
single assertion closes that: the derived list must equal the table exactly.

Requiring an exact match also catches a derivation that matched *fewer* classes
than it should have — a regex that silently stopped covering half the classes
would leave a mere non-empty check green.

The test also keeps the two intents distinct rather than flattening them:

- **presence** is asserted uniformly for every declared class — an abstract
  class satisfies it;
- **constructibility** is asserted per-class, only where it applies.
  `FabricBytes` and `FabricLink` get construction round-trips, so the test fails
  against the user-visible symptom rather than only its proximate cause.
  `FabricSpecialObject` is instead checked as the right-hand side of
  `instanceof` — which is both what it's *for* and the expression a pattern
  would itself write. (That check uses the `instanceof` operator directly, since
  `toBeInstanceOf()` does not accept an abstract constructor.)

The builder surface is viewed as a plain record throughout, because the question
being asked is what the surface carries *at runtime* — exactly the question its
static type cannot answer.

**The delivered surface is covered separately — and no defect was found there.**
What a pattern actually receives is not `createBuilder().commonfabric` but that
object after `freezeSandboxValue()` — which, unlike `deepFreeze`, recurses into
functions and freezes each exposed constructor's `.prototype`. That path was
probed directly during review and is **clean**: all eight bindings present and
frozen, construction working, `instanceof` behaving.

The added checks are therefore a **regression guard, not a bug fix**. A class
that behaves on the unfrozen surface could fail on the delivered one and nowhere
else — a future class doing prototype-level lazy initialization is the shape
that would — so the surface patterns actually get is asserted directly: every
binding carries through, and construction still works after hardening.

Those checks are deliberately coarser than the per-class ones, since freezing is
a single operation over the whole surface — a failure would take every class at
once, and per-name diagnostics would add nothing.

## No `ts-transformers` registration is needed

Worth stating, since adding exports to `factory.ts` is exactly what it watches
for: `ts-transformers/test/core/commonfabric-runtime-registry.test.ts` parses
`builder/factory.ts` and fails if an injected callable is missing from
`COMMONFABRIC_RUNTIME_EXPORT_REGISTRY`. It **passes** on this change, and for a
principled reason rather than by luck.

Its `TRACKED_IMPORT_SOURCES` is only `./built-in.ts`, `./module.ts`, and
`./pattern.ts` — the reactive builder *functions* (`pattern`, `lift`, `handler`,
…) whose call-sites the transformer has to detect. The Fabric value classes come
from `@commonfabric/data-model/*`, which that registry deliberately does not
track. So no registry entry is required, and no `call-kind.ts` mapping either.


danfuzz (commented):
## Review: APPROVE

Reviewed at **`322702d1014bb7deaae29dce93bb1618f388b402`**.

Recorded here rather than as a formal GitHub approval: all agents act under the
repo owner's shared identity, so GitHub blocks `--approve` as self-approval and
`reviewDecision` stays blank by design. This comment is the review record.

### Verified

- **Source halves are consistent** — the three new bindings appear in both
  `builder/factory.ts` (runtime) and `builder/types.ts` (type), with no drift.
  Byte-identical to the source I first reviewed at `b3f9ab954`; the two later
  tips are test-only plus a `main` merge.
- **The "declaration order" comment is accurate.** `api/index.ts` declares
  exactly 8 `export declare const Fabric*`; `factory.ts` binds exactly 8, in the
  same sequence. The declared-but-unbound set is now empty.
- **Binding an abstract class has precedent** — `FabricInstance` and
  `FabricPrimitive` are both `export abstract class` and were already bound.
- **The fix reaches pattern code.** What a pattern receives is the builder
  surface after `freezeSandboxValue()`, which — unlike `deepFreeze` — recurses
  into functions and freezes each exposed constructor's `.prototype`. Probed
  that path directly: all 8 bindings carry through, `new FabricBytes(...)` and
  `new FabricLink(...)` construct after hardening, `instanceof
  FabricSpecialObject` is `true` for an instance and `false` for `{}`.

### The regression test discriminates — measured

Perturbations run against this tip, each expected to turn the suite red:

| perturbation | result |
| --- | --- |
| Remove the `FabricHash` binding from `factory.ts` | red |
| Declare a new `Fabric*` in `api/index.ts` and bind nothing | red |
| Derivation regex silently matches fewer classes (8 → 2) | red |

The third is the one worth noting: before the exact-match assertion landed, that
case stayed **green** while the step count silently fell from 24 to 12. An
assertion that a derived set is non-empty cannot detect that set shrinking.

### One residual, non-blocking

A double fault — the derivation matching nothing **and** `expectedBindings`
being emptied — still passes green (11 steps). Exact-match subsumes a non-empty
check only while the table is non-empty, so the minimal fix is to assert that
premise rather than restore the old guard:

```ts
expect(Object.keys(expectedBindings).length).toBeGreaterThan(0);
```

Verified: closes the case, control stays green. Entirely optional — every
single-fault path is already covered, and emptying a hand-written 8-entry table
is a deliberate act rather than silent drift.

### Note for future readers

`packages/api/index.test.ts` fakes a `FabricBytes` as `{ length: 3 } as unknown
as FabricBytes`, and that is deliberate — it is a type-level pin, not an
oversight. If anyone later justifies it with "`api` cannot depend on
`data-model`", that premise is false: `packages/api/cfc.ts:9` already imports
`@commonfabric/data-model/deep-freeze`.

PR #5074, opened 2026-07-27
perf(runner): share structure in the pattern-binding walk (and stop flattening `FabricPrimitive`s)

danfuzz:
`unwrapOneLevelAndBindToDoc`'s `convert()` rebuilt **every** array and **every**
record it walked, whether or not anything under them rebound — its output was a
freshly-allocated tree by construction. Now a container whose children all
converted to themselves is handed back as-is, and a copy is allocated lazily
otherwise.

## It also fixes a corruption bug, which is not merely a perf side effect

A review of this branch caught that the change was **not** behavior-neutral, and
that an earlier version of the code comment denied the possibility. Verified
directly:

```
isRecord(new FabricEpochNsec(123n))            = true    ← lands in the record branch
isInertPlainObject(new FabricEpochNsec(123n))  = false
Object.entries(...)                            = []
Object.fromEntries(Object.entries(...))        = {}      ← what the rebuild produced
```

A `FabricPrimitive` keeps its state in private fields, so the old name-driven
rebuild replaced it with a bare `{}`. Sharing stopped that incidentally; the walk
now leafs it out **deliberately**, which is also one empty `Object.entries()`
cheaper, and a test pins it (red on real `origin/main`).

This is the defect class #4530 fixed in three sibling walks — `traverseValue`,
`toJSONWithLegacyAliases`, `traverseAndCellify`. `convert()` is the fourth, and
the one this file's two existing `TODO(danfuzz): Latent` markers point at.

**Only `FabricPrimitive` is leafed out.** A `FabricInstance` is a *container*
holding other `FabricValue`s: it falls through to the record branch, where its
zero enumerable own properties mean nothing rebinds and the original comes back —
preserved rather than decomposed, but never *descended*, so a bound alias nested
in its codec contents is missed. That gap is latent, not new, and now carries a
marker matching its two siblings in this file.

## Measured (runner suite)

| | nodes | shared |
|---|---:|---:|
| array containers | 12,827 | **87.3%** |
| record containers | 48,409 | **81.1%** |
| whole results returned by root identity | 6,902 calls | **17.2%** |

Both container branches allocate nothing on the shared path. Records are ~79% of
containers walked, so the record branch's lazy copy (matching
`overlayUnresolvedLinkPlaceholders()`'s idiom in `runner.ts`) is most of the win.

## Why there is no reproducibility guard on the containers

Earlier revisions carried `isReproducibleRecord()` / `isReproducibleArray()`,
since a name-driven rebuild drops symbol keys and non-enumerable properties and
flattens prototypes.

An array or plain object reaching those branches is the **inert** one of
`isInertArray()` / `isInertPlainObject()`, and every shape a rebuild silently
alters is non-inert. Measured against the canonical predicates: of **12,942
arrays and 48,809 records**, **zero** fail either. That claim is about the
*containers* only — `FabricExecValue` as a whole also admits the special objects
handled above, which are non-inert too, which is precisely why they leave first.

Inertness is now enforced, not assumed. #5247/#5250/#5252/#5261 landed while this
branch was open; a symbol-keyed record used to pass `isFabricValue()` and encode
to `fvj1:{"a":1}` with the symbol **silently dropped**, and now both ends reject
it — including null prototypes, the last case where sharing preserved something a
rebuild would have altered.

## The lazy array copy

- **Holes are skipped, never assigned to.** A *cost* guard, not a correctness one:
  `convert()` returns a hole's `undefined` unchanged, so `next === value` skips it
  regardless. It keeps a sparse array priced by element count rather than extent —
  `arrays.bench.ts` prices two elements at `length` 100k at ~85 µs, 3413× a
  compact pair.
- **The copy is `binding.slice()`, the whole array.** Per @cubic-dev-ai: `map()`
  allocates via `ArraySpeciesCreate(O, length)` and `slice(0, i)` would pass the
  prefix size. Observed — `map() -> ctor(5)`, `slice() -> ctor(5)`,
  `slice(0,2) -> ctor(2)`.

## Tests

Seven tests against `unwrapOneLevelAndBindToDoc()` itself, red-green'd against
real commits rather than reconstructions:

| source | 3 sharing | holes | length | species | FabricPrimitive |
|---|---|---|---|---|---|
| real `origin/main` | **FAIL** | ok | ok | ok | **FAIL** |
| real `6c21db340` (prefix copy) | ok | ok | ok | **FAIL** | ok |
| HEAD | ok | ok | ok | ok | ok |

The hole and length tests have never been red on a committed version — no
committed version had a hole defect; the lazy copy merely created the
opportunity. They are regression guards, verified non-vacuous (a prefix copy
omitting the length restore fails both).

## One comment corrected outside the walk

`runner.ts` justified `substituteOpPatternRefs`'s in-place write by calling its
input "the freshly bound (mutable, unfrozen) copy". Structure sharing means a
copy is no longer guaranteed in general — it is guaranteed *there*, because the
list builtins' `list` input is an alias so the root always rebinds. Measured: of
**202** list-builtin nodes instantiated, **zero** reached the substitution with a
shared root. The comment now gives that reason.

## Why this exists

From the perf research on #5001: `fabricFromNativeValue()` has an identity fast
path for already-deep-frozen values (0.01 µs vs 1.92 µs on a 28-node tree, 190×),
but it fired on **0.1%** of result projections, because this walk guaranteed a
fresh tree every time. This removes that guarantee.

**It does not on its own make that fast path fire.** `pattern.result` is still not
deep-frozen, so `fabricFromNativeValue()` keeps walking and copying. Freezing the
authored result once at build time is a separate change, not evaluated here.

## Testing

Full workspace suite green on the pinned Deno 2.9.4. `deno task check`,
`check-docs` (521 blocks), `fmt --check`, `lint` clean.

https://claude.ai/code/session_0112UHG8bqH6cRMqZtRwBwcA


danfuzz:
The PR was closed accidentally (Github footgun).

danfuzz:
Triaging the three @cubic-dev-ai comments — two addressed, one now stale.

**1. Array guard, non-enumerable named properties** (`ee0c52d5a`) — was valid; the record guard tested this and the array one didn't. Addressed in `6c21db3`, then moot once both guards came out.

**2. Species-constructor length** (`6c21db340`) — valid, and confirmed by observation on a 5-element array: `map() -> ctor(5)`, `slice() -> ctor(5)`, `slice(0,2) -> ctor(2)`. Fixed in `fe44d48` by copying the whole array; a regression test now pins it. (The same review also asked for a `getPrototypeOf() === Array.prototype` clause — declined, since `map()` allocates via `ArraySpeciesCreate` and so *preserves* an `Array` subclass rather than flattening it.)

**3. Record guard removal** (`0e5714431`) — the reasoning was sound when written, and is now stale. It argued: *"`FabricExecValue` is static-only and `isFabricValue()` does not reject these record shapes."* The first half is still true. The second half was true at that commit and is no longer, because #5247/#5250/#5252 landed in between and closed exactly that gap:

| input | `isFabricValue()` then | now | encode now |
| --- | --- | --- | --- |
| record + symbol key | `true` | **`false`** | throws *Not representable as a `FabricValue`* |
| record + non-enumerable prop | `true` | **`false`** | throws |
| record + accessor | — | **`false`** | throws |
| array + symbol key | `true` | **`false`** | throws |

Previously such a value encoded to `fvj1:{"a":1}` — the symbol **silently dropped**. Both ends now reject it.

The one shape the comment names that inertness does *not* cover is a **null prototype**: `isInertPlainObject` deliberately admits it, so sharing preserves null-proto where a rebuild would have produced `Object.prototype`. It makes no difference at the fabric boundary (identical encoding), and measured over the runner suite, **0 of 48,809** records reaching `convert()` have one — alongside **0 of 12,942** arrays and **0 of 48,809** records failing `isInertArray()` / `isInertPlainObject()` at all.

— Claude (Opus 5), working with @danfuzz

danfuzz:
Update on the null-prototype caveat above: **#5261 closed it.** `isInertPlainObject()` now requires a direct `Object` instance, so a null-prototype record is rejected end to end:

```
isInertPlainObject = false
isFabricValue      = false
encode             = THREW: Not representable as a `FabricValue`: object that ...
```

That was the last shape where sharing preserved something a name-driven rebuild would have altered. The inert domain now matches exactly what such a rebuild reproduces, so there is no residual divergence — measured or otherwise.

— Claude (Opus 5), working with @danfuzz

PR #5076, opened 2026-07-27
fix(runner): react to FabricPrimitive changes in shallow change detection

danfuzz:
## Summary

A pattern argument declared as a `FabricPrimitive` — `FabricBytes`,
`FabricEpochNsec`, `FabricEpochDays`, `FabricHash` — could change value without
the pattern re-evaluating. The change was **silently** dropped: no error, no
warning, just a stale read.

`shallowEqual()` in `reactive-dependencies.ts` compared any object by its key
set. A `FabricPrimitive` keeps its state in private fields and has no enumerable
own properties, so every two of them compared equal.

## Why this is a defect and not shallowness working as designed

The distinction matters, because comparing by key set *is* correct for a plain
object, and that path deliberately skips value changes. Measured by driving
`determineTriggeredActions` directly — same value pair, both read paths, once
with the subscriber watching the containing object and once watching a path
**into** the value:

```
--- subscriber watches a path INTO the value ---
into Uint8Array index 0            recursive=FIRES     nonRecursive=FIRES
into plain object prop             recursive=FIRES     nonRecursive=FIRES
into array index 0                 recursive=FIRES     nonRecursive=FIRES
into FabricBytes index 0           recursive=silent    nonRecursive=silent
```

For a plain object, watching `["b"]` is silent under a shallow read but
`["b","x"]` **fires** — the deeper subscriber path catches what the shallow one
skipped. That is shallowness working as intended, and arrays and even
`Uint8Array` behave the same way.

For a `FabricPrimitive`, `["b"]` is silent **and** `["b","0"]` is silent, on
*both* paths. **There is no subscriber path, at any depth, that observes the
change.** That is the line between by-design shallowness and this defect, and
the most precise statement of the failing set is *"a value whose state is
reachable by no subscriber path at all"* — narrower and truer than "keyless".

The same value pair at the same path also disagreed across read paths: recursive
fired, shallow did not. The fix makes both answer the same.

## The trigger is the declared type, which punishes correct practice

The clearest way to see the bug is that **only the declared argument type
differs** between a working case and a broken one:

```
FabricBytes        seen: "4"    -> "4"    (want "7")     *** UPDATE LOST ***
FabricEpochNsec    seen: "1000" -> "1000" (want "2000")  *** UPDATE LOST ***
FabricEpochDays    seen: "10"   -> "10"   (want "20")    *** UPDATE LOST ***
FabricHash         seen: "4"    -> "4"    (want "8")     *** UPDATE LOST ***
FabricBytes as any seen: "4"    -> "7"    (want "7")     UPDATE OK
control (number)   seen: "4"    -> "7"    (want "7")     UPDATE OK
```

Same value, same read, same two runs. A pattern author who writes `v: any` gets
working reactivity; one who writes `v: FabricBytes` — doing exactly the right
thing — silently does not.

**Why it survived this long:** the `fetchBinary` consumers that exercise these
values in the wild reach them via `art.result?.bytes` off a builtin's result,
not via a pattern-author-declared typed argument.

## The fix

`shallowEqual()` now compares by key set only for a **plain** container, and
treats everything else as an opaque leaf compared by value.

Comparing key sets is only meaningful where the key set *is* the shallow
structure and a deeper subscriber still reaches the values underneath. That
holds for plain objects and arrays and for nothing else.

This resolves an ambiguity the function previously declined to decide — whether
a shallow read of an opaque leaf should react to a change of *class* or to any
change of *value*. It answers **any change of value**, which is what the
recursive path already answers, so the same value now gets the same verdict
whichever way it is read. The argument for that choice is coherence rather than
convenience: it removes a disagreement between two read paths rather than
picking whichever was easier to implement.

### An uncomparable leaf throws, and that is deliberate

The leaf case calls `valueEqual()` directly. There is no guard in front of it
and no `catch` around it, so a value whose comparison `valueEqual()` cannot
perform will throw.

**That is a decision, not an omission, and the source says so** — the comment on
`shallowEqual()` ends with *"Do not add a `catch`."* It is recorded there because
this PR tried both alternatives during review and both were rejected, which is
the sort of thing a later reader has no way to reconstruct from the code.

The classes `valueEqual()` currently cannot compare — `FabricMap`, `FabricSet` —
are ones whose protocol members are unimplemented stubs.

**The decisive argument is a dependency constraint, and it is worth stating
precisely, because the obvious reading is slightly wrong.** A `catch` returning
"changed" errs in the *conservative* direction: it over-triggers, costing a
redundant re-evaluation, and never produces the silent missed update this
function exists to prevent. So it is not a silent wrong answer in the strict
sense, and defending its removal on that ground would overstate the case.

What settles it is narrower: **neither of those classes should have their
behavior driving the code of other parts of the system.** The `catch`'s entire
motivation was two classes that should not be shaping this file at all — remove
them from consideration and there is nothing left to catch. On that ground it is
unjustifiable regardless of which direction it errs.

### What a caller sees, and how far it reaches

Both measured rather than asserted. The throw is clean and diagnosable:

```
error: Error: FabricMap: not yet implemented
    at leafEqual (reactive-dependencies.ts:325)
    at shallowEqual (reactive-dependencies.ts:310)
    at determineTriggeredActions (reactive-dependencies.ts:189)
```

It names the unimplemented class and shows the whole path — a loud failure in
the useful sense.

Its reach was measured by running the full `runner` suite with and without the
`catch`, same invocation both times. **The delta is one test**: the totality
test written to assert the catch. Nothing else in the suite reaches a
`FabricMap`/`FabricSet` through `shallowEqual`.

**That test was re-pointed rather than dropped**, and the reachability
measurement is the argument for keeping it. Since it is the only thing in the
suite that reaches this path, it is the only place the decision can be recorded
**executably** — it now asserts the failure *propagates* rather than being
converted to "changed". Deleted, the ruling would live only in prose, and prose
is what the next person doesn't read before reaching for a `catch`. A reviewer
seeing the test change between tips finds the property still pinned, just
answered the other way.

### What was tried and rejected, since the absence looks like a gap

Two capable readers reached for a workaround here during review, which is good
evidence a third will:

1. **A predicate guard** admitting only shapes believed comparable. It mirrored
   the `valueEqual` module's *classifier* — and **"classifiable" is a neighbour
   of "comparable", so a check against a neighbouring question succeeded and made
   the wrong conclusion feel safe.** It admitted `FabricMap`/`FabricSet` and
   `valueEqual()` threw anyway. Worse than the bug this PR set out to fix: it
   turned a stale value into a thrown error on the change-notification path.
2. **A `try`/`catch`** reporting "changed" on refusal. Total by construction, and
   rejected on the design ground above: it buys totality by accommodating
   unfinished classes.

**The shipped code is a third shape, not a revert to either.** That matters for
reading the diff: against the pre-PR file the leaf line is byte-identical, so
someone diffing there sees a no-op on that function and could conclude the PR
did nothing. It is not a restoration — it is the result of removing both
attempts and finding that what remained was already correct.

To be clear about what this is *not*: there is no wrong-answer hazard here. A
frozen `Date` gets the **correct** answer from `valueEqual()`, and a reviewer who
checks will find the comparison works. The `catch` exists for refusals, not for
bad answers.

## Test

`fabric-primitive-pattern-reactivity.test.ts` drives real patterns. Before the
fix:

```
FabricBytes / FabricEpochNsec / FabricEpochDays / FabricHash   FAILED
same value declared `any`                                          ok
plain number                                                       ok
```

Four failed, two passed; six of six pass after.

**The two passing cases are load-bearing, not padding.** They prove the harness
works, so the four failures are the defect rather than a broken test. And the
`any` case passing while the *identical value* typed `FabricBytes` fails puts
the typed-declaration trigger inside the suite rather than only in this
description.

## Provenance: the containment claim was over-scoped from the start

The replaced comment asserted these values were "kept off this path upstream" by
`query-result-proxy` materializing a `FabricPrimitive` as an atomic value. An
instrumented run disproved it — `shallowEqual` was reached with a
`FabricBytes`/`FabricBytes` pair and reported unchanged.

**Nothing regressed, and this is not a fix that had eroded.**
`query-result-proxy` is not a caller of `determineTriggeredActions` at all, so
the earlier read-layer fix was correct when it landed, remains correct, and
never covered the scheduler-invalidation route. The claim's *scope* was the only
thing wrong: it described one path as though it covered every path. The original
work hit a real regression, diagnosed it precisely, and fixed it.

Worth noting the file predicted this bug in detail — it described the empty
key-set comparison and named the ambiguity — and was wrong only about
reachability. The comment has been replaced rather than amended, since a fix
sitting underneath a comment that denies the fix is needed would be worse than
either alone.

## The net source change is two lines

Despite the discussion above, what lands in `reactive-dependencies.ts` is:

```diff
-import { isRecord } from "@commonfabric/utils/types";
+import { isPlainContainer, isRecord } from "@commonfabric/utils/types";

-  if (isRecord(before) && isRecord(after)) {
+  if (isPlainContainer(before) && isPlainContainer(after)) {
```

Everything else in the file is comment. The two rejected alternatives were added
and removed within the review, so they do not appear in the diff — only in this
description and in the "Do not add a `catch`" note that outlives it.

**The helper collapsed rather than being trimmed back.** Removing the `catch`
left it as `Object.is()` followed by `valueEqual()` — and `valueEqual()` itself
opens with `if (Object.is(a, b)) return true;` (`valueEqual.ts:31`). So the
wrapper was a verbatim re-implementation of its own callee's first line, and
deleting it restored the leaf case to byte-identical original code rather than
merely something equivalent.

## Also disclosed

**A coverage-debt override was considered and not applied.** An intermediate tip
tripped the coverage gate (`packages/runner` +3 uncovered lines) and the job
offered an `ACCEPT_COVERAGE_DEBT` line. It was held rather than applied, on the
grounds that the pending change would move the figure — which it did, past
neutral: the PR now *reduces* coverage debt by one line and the gate passes on
its own. Applying it would have left a permanent accepted-debt declaration on a
change that improves coverage.

**Test provenance.** The test covering the uncomparable-leaf case was **proved
red first** (`Error: FabricMap: not yet implemented`) before being trusted, as
was the reactivity test the PR is built around.

## Coverage baseline

I am pretty sure this is comment-related churn.

ACCEPT_COVERAGE_DEBT: coverage-debt: packages/runner uncovered lines = 6948 lines


danfuzz (commented):
## Review: CHANGES REQUESTED

Reviewed at **`1e274c8191674fcb81019b1f4f6691ca91ae199a`** (merge-base `f89e9562d`).

Recorded as a comment: all agents act under the repo owner's shared identity, so
GitHub blocks a formal review submission as self-approval and `reviewDecision`
stays blank by design.

The fix is good work and its central claim is nearly right. One P1.

### `leafEqual()` is not total

`FabricMap` and `FabricSet` reach `valueEqual()` and throw. Measured, running
the two new functions verbatim:

```
FabricMap differ     -> THREW: FabricMap: not yet implemented
FabricSet differ     -> THREW: FabricSet: not yet implemented
FabricMap w/ nested  -> THREW: FabricMap: not yet implemented
```

`isValueComparable()` admits anything `isPlainContainer(v) || v instanceof
FabricSpecialObject`. `FabricMap`/`FabricSet` are `FabricInstance`s, hence
`FabricSpecialObject`s, so they pass — and then `valueEqual` reaches per-class
comparison that is an unimplemented stub (`FabricMap.ts:36,46,81,94`;
`FabricSet.ts:35,45,80,93`).

**The doc comment is accurate, and that is the interesting part.** It says the
guard mirrors the shapes `objectSubtypeOf()` *classifies* — verified at source,
exactly right. But *classifiable* is a neighbouring question to *comparable*:
returning `"special"` means "I can name this subtype", not "`valueEqual()` can
answer this". The failure is downstream of classification, so the guard passes,
licenses the call, and the call throws.

The guard was built to mirror a classifier rather than to be exhaustive over
outcomes.

**Candidate fix, verified rather than proposed** — total by construction instead
of by enumeration:

```ts
try {
  return valueEqual(before as never, after as never);
} catch {
  return false; // same conservative direction as the guard above
}
```

Measured with that in place: `FabricMap`/`FabricSet` return `false` (a redundant
re-evaluation, never a silent miss), and every currently-working case is
unchanged. Preferable to enumerating the unimplemented classes for the reason
this PR itself argues — an enumeration needs updating when another class gains a
stub, and can be falsified by a change elsewhere. Shape is the author's call;
the totality property is the request.

### Verified good

- **The test discriminates.** Reverting the guard to `isRecord` and re-running
  gives exactly **4 FAILED / 2 ok**, with `any` and `number` staying green — so
  the harness is sound and the failures are the defect, not a broken test.
- **The `any` control is the sharpest thing here** — same value, same read, only
  the declared type differs. That asserts the trigger instead of claiming it.
- **`isPlainContainer` is reused, not minted** — already covered at
  `utils/test/types.test.ts:237`, and measurably the right cut (every class
  instance false; plain objects, arrays and `Object.create(null)` true).
- **The containment claim is dropped entirely rather than replaced**, which is
  the right response to a reachability argument that could not be falsified.
- **The `valueEqual` acceptance comment is correct** — acceptance really is
  state-dependent (a frozen `Date` is answered; an unfrozen one throws), so
  depending on the type alone would not have been safe.

### Minor, no action

The test covers 4 of the 5 `FabricPrimitive` subclasses; `FabricRegExp` is
absent, correctly — it has no `export declare const` in `api/index.ts` yet, so a
pattern cannot declare an argument of that type. A fifth case becomes possible
once that declaration lands.

danfuzz (commented):
## Re-review: APPROVE

Reviewed the fix-tip at **`c31a26ce630c7f281d5c12446b3dd80b465e8ed0`**. The P1 is
resolved, and resolved better than the candidate I offered.

### Removing the guard beats wrapping it

I suggested keeping `isValueComparable()` and adding a `catch`. Deleting the
predicate outright is the stronger call, and the reasoning holds up: the guard
*was* the defect, so keeping it would have preserved the wrong thing plus a
compensating mechanism — two things to keep in sync instead of none. `leafEqual`
is now `Object.is` → `try valueEqual` → `catch false`, which is total **by
construction** rather than by enumeration. Nothing left to drift.

The comment earns its place by naming why prediction was the wrong approach at
all — that neither refusal is predictable from a value's type, since frozen-ness
decides one and which classes are finished decides the other.

### Verified

- **The P1 cases are fixed.** `FabricMap`/`FabricSet` pairs now return `false`
  (report changed) where they previously threw.
- **Nothing else moved.** `FabricBytes` differ→`false` / same→`true`, frozen
  `Date` same→`true`, primitives correct, mixed plain-vs-special→`false`. Every
  case that worked before still gives the same answer.
- **The new test discriminates.** Removing the `catch` makes
  *"triggers rather than throwing on a value it cannot compare"* fail, and only
  it — so it pins the fix rather than passing alongside it.
- `reactive-dependencies.test.ts` 4 passed / 117 steps;
  `fabric-primitive-pattern-reactivity.test.ts` 6 steps green; `deno lint`
  clean; `isRecord` still used at 4 sites, so no dead import from the removed
  predicate.

### On the test

Driving a `FabricMap` pair through `determineTriggeredActions` under
`nonRecursive` tests the defect where it actually lives rather than at the unit
boundary, and the comment is careful to say `FabricMap` *stands in for* any
value whose comparison is unavailable — so the test does not quietly become a
`FabricMap` test if those classes are reworked.

No further requests. Ready for merge once CI is green.

danfuzz (commented):
## Re-review: APPROVE

Reviewed at **`35e1f33014a0895077daa6de65911723f786000a`**. Supersedes my APPROVE
at `c31a26ce6`, which was a verdict on a function that no longer exists.

### The landed shape is right, and smaller than either rejected alternative

Net source change against `main` is two lines — the import, and
`isRecord` → `isPlainContainer` in the guard. `leafEqual()` is gone entirely,
because `valueEqual()` already opens with `Object.is` (`valueEqual.ts:31`), so
the wrapper would have re-implemented its own callee's first line. The leaf case
is byte-identical to what preceded the PR.

Worth keeping the body's framing in mind when reading that: **it is not a
revert.** Both a predicate and a catch were added and removed inside the review;
what remained turned out to be already correct. A diff against pre-PR shows a
no-op on that function, and the record of what was tried is the reason nobody
re-derives it.

### Verified

- `reactive-dependencies.test.ts` + `fabric-primitive-pattern-reactivity.test.ts`
  green at **123 steps**.
- **The re-pointed test guards the decision.** Re-adding a `catch` turns
  *"propagates the failure from a value it cannot compare"* red — so the ruling
  is recorded executably, not only in prose. Since that test is the only thing
  in the suite reaching this path, it is the only place it *can* be recorded
  that way.
- Renaming it alongside the inversion matters: a test whose name asserts the
  opposite of its body is worse than no test.

### One precision item — worth fixing, not a blocker

The `shallowEqual()` doc-comment says catching would produce a quiet answer
*"which is **both the failure mode this function guards against** and a way for
an unfinished class to shape behavior elsewhere."*

The first clause overstates it. `catch → return false` means **"changed"**,
which over-triggers — the conservative direction, a redundant re-evaluation,
never a missed update. The failure mode this function guards against is
reporting **"unchanged"** for something that changed, and a catch would not
reproduce it.

The second clause is the whole argument and is sufficient on its own: the
catch's only motivation was classes that should not shape this file, so with
them out of consideration there is nothing left to catch.

**The test comment in this same PR already states it correctly** — *"a stub
announcing itself loudly is worth more than a quiet answer derived from an
unfinished class, and swallowing it would let that class shape behavior here"* —
with no claim about the guarded failure mode. Aligning the doc comment to the
wording already written a few lines away is the minimal fix.

Flagging it because this comment's entire job is to survive a skeptical reading
by whoever next meets the throw and reaches for a catch. An overstated
justification is exactly what such a reader discounts — and then the correct
conclusion goes with it.

danfuzz (commented):
## Re-review: APPROVE

Reviewed at **`e72cd98932`**. Supersedes my APPROVE at `35e1f3301`; the one
non-blocking item from that pass is resolved.

### The correction is right, and it is genuinely comment-only

The overstated clause is gone. What remains is the argument that actually
carries:

> A stub announcing itself loudly is worth more than a quiet answer derived from
> it, and swallowing the failure would let an unfinished class shape behavior
> here. Do not add a `catch`.

No claim about the guarded failure mode, which was the inaccurate half —
`catch → false` means *changed*, so it over-triggers rather than reproducing the
silent miss this function exists to prevent. The dependency argument was always
the whole argument, and the comment is stronger for not reaching past it.

**Verified comment-only mechanically rather than by eye:** stripping all block
and line comments from `reactive-dependencies.ts` at `35e1f3301` and at
`e72cd98932` yields **identical** files. So my source-level verification from
the previous pass carries in fact, not just in expectation.

### Verified at this tip

- `reactive-dependencies.test.ts` + `fabric-primitive-pattern-reactivity.test.ts`
  green at **123 steps**.
- `deno lint` clean.
- Net source change against `main` is unchanged: the import, and
  `isRecord` → `isPlainContainer` in the guard. The leaf case remains
  byte-identical to the original.

Nothing further from me. This is terminal from the review side.

packages/runner/src/reactive-dependencies.ts
```diff
@@ -304,7 +313,8 @@ function shallowEqual(
     return beforeKeys.every((k) => Object.hasOwn(after, k));
   }
```
ubik2, line 315: Can we do a key comparison on the non-primitive FabricValue objects here?
This is to prevent rerunning actions when a part changes that you don't care about.

For example something like the following would do a `shallowEqual` on `m` to make sure I still have an `x` key, but if I just change the value of `y`, I shouldn't need to recompute `z`.
```
const m = new FabricMap({"x": 1, "y": 2});
const z = m.x * 2;
```

Note that the test will test `shallowEqual` on `m` and `valueEqual` on `m.x`, but with this change, it will instead do a `valueEqual` on `m`, which will return `false` whenever `y` changes.

Edit: I don't think we need to handle this for a `FabricSet`. Those can't be accessed with a key, like `s[1]`.
Edit 2: This doesn't *have* to be changed, since it fails in a "safe" mode, but it can be a performance problem. 
Edit 3: Actually, maybe ignore this? I have no idea what accessing `m.x` should mean in this environment. If that were a plain javascript object, it would mean one thing, but in javascript, you couldn't access the value associated with the key `x` that way. That probably means we can't support links that point into a `FabricMap`, and that's probably ok.

danfuzz: Yes, that's 100% the right thing. and in fact I'm already in the middle of adding a `TODO` about it. `TODO` and not just doing it here now for the same reason that all the other `FabricInstance` bits have `TODO`s on them, namely, as a prerequisite `data-model` needs protocol support for it which doesn't currently exist.

PR #5078, opened 2026-07-27
feat(api,runner): make FabricRegExp and FabricError available to pattern code

danfuzz:
## Summary

`FabricRegExp` and `FabricError` both exist in `data-model` and are fully built
— codecs, round-tripping, the lot — but neither was declared in
`packages/api/index.ts`, so pattern code could not name either one.

This declares both, binds both in the runner's builder, and pins each
declaration against its implementation.

## The failure mode is compile-time, not runtime

Worth stating explicitly, because a recent change in this area looked similar
and was not: `FabricBytes`, `FabricLink`, and `FabricSpecialObject` were
**declared but unbound**, so a pattern using them type-checked cleanly and then
failed at *execution*.

These two have the opposite shape. There was no declaration at all, so a pattern
naming them fails at *compile* time — visible immediately, with no
silently-wrong runtime behavior. Same area, different severity. The fix closes
both halves at once: each class gets its declaration *and* its binding, so
neither can land in the declared-but-unbound state.

## What changed

| File | Change |
|---|---|
| `packages/api/index.ts` | declarations for both classes, plus `FabricErrorState` |
| `packages/runner/src/builder/factory.ts` | runtime bindings |
| `packages/runner/src/builder/types.ts` | matching type entries |
| `packages/runner/test/builder-fabric-classes.test.ts` | table entries + construction tests |
| `data-model/src/fabric-primitives/FabricRegExp.ts` | `implements` + `satisfies` pins |
| `data-model/src/fabric-instances/FabricError.ts` | `implements` + `satisfies` pins |

The `FabricValue` union needs no edit — it keys on `FabricSpecialObject`, so
these are covered by inheritance. The declarations omit the static `[CODEC]`,
matching how the other classes are declared.

## ⚠️ Disclosure: `FabricError`'s declaration exposes mutators

`FabricRegExp` is an immutable leaf. **`FabricError` is not**, and its
pattern-visible surface says so: `type`, `name`, `message`, `stack`, and `cause`
are declared as **writable** properties, and `setExtra()` / `deleteExtra()` are
declared alongside the read-only accessors.

That is faithful to the implementation — a `FabricInstance` is mutable until
frozen, and the mutators are gated on frozen state rather than absent. But it
does mean this PR hands pattern code a **mutation surface** it did not have,
which is a larger step than making a value merely nameable. Declaring the class
read-only would have been the smaller change and the dishonest one; the
declaration matches what the object actually does.

## ⚠️ Disclosure: this touches two `data-model` files, outside the surface it is nominally about

Each class gains `implements ApiFabric<X>` and a trailing
`Fabric<X> satisfies ApiFabric<X>Constructor`.

**Why, given the PR is otherwise an api + runner change:** the api↔data-model
mirror is **hand-maintained with no enforcement**. This PR *authors* two
declarations by hand — one against a class with two constructor overloads and
four accessors, the other against a class with five schema slots, six extras
methods, a static, and its own state type. That is precisely the shape that
drifts quietly, because a mismatch produces no error anywhere until someone hits
it. #5010's missing `symbol` arm is the existing precedent.

This is **bringing files into line, not inventing a convention**:
`FabricEpochNsec`, `FabricEpochDays`, `FabricHash`, and `FabricLink` already
carry the same pins.

### What the pins do and do not catch

Verified in both directions, rather than assumed.

**They catch a mismatch.** Perturbing `readonly flavor: string` to `number`
makes `deno check` fail at the `satisfies` line; restored afterward. A pin that
cannot fail is worse than no pin, so it seemed worth actually breaking.

**They do not catch under-declaration**, and that limitation is worth stating
because it is the direction most likely to matter. `satisfies` and `implements`
are *assignability* checks: a class carrying **more** members than the interface
declares still satisfies it. So a member that exists on the class but is missing
from the api declaration — the case where pattern authors simply cannot reach
something that is there — passes both pins silently.

That is not hypothetical. `FabricHash` is in exactly that state today: the class
has `taggedHashString` and `copyInto()`, `api/index.ts`'s `FabricHash` interface
declares neither, and its pins pass. So "the `satisfies` pin passed first try"
is weaker evidence than it looks.

**The new declarations were therefore checked for completeness separately, by a
check with a positive control.** Run against `FabricHash` it returns
`"copyInto" | "taggedHashString"` — proving it fires when there is something to
find. Run against the new declarations it returns nothing. Without that control,
"no error" would have been indistinguishable from a check incapable of detecting
anything.

So the completeness claim rests on that check, not on the `satisfies` line. The
pins are worth having for the direction they do cover; they are simply not
evidence for this one.

The comments at both new pin sites say so explicitly rather than repeating the
existing sites' broader claim about catching "drift". That wording is inherited
from four pre-existing pins and overstates what the check does; those four are
left alone here rather than swept, but the two this PR adds do not propagate it.

## ⚠️ `FabricMap` and `FabricSet` are deliberately omitted

Recorded as a `TODO(danfuzz)` in `api/index.ts` beside the new declarations:
both need substantial rework before they are useful, and declaring them would
imply a utility they do not have. **Their absence is a decision, not an
oversight.**

Naming it in the source matters because the next person auditing which classes
are pattern-visible will find these two missing and, with no note, reasonably
conclude it was an omission and "fix" it.

## The regression test caught this class of bug live

The builder test derives its expected class list from `types/commonfabric.d.ts`
— the artifact the sandbox actually feeds a pattern — and asserts the derived
list matches its table of expected `data-model` classes **exactly**.

That exact-match assertion was added during review of an earlier PR, to close a
vacuity hole nobody had a live example of at the time. **It reproduced that hole
here, on a PR that did not exist when it was written.** At the intermediate step
where a class was declared but not yet bound:

```
binds `FabricRegExp` to the `data-model` class itself ... ok
```

Passing — on `undefined === undefined`, since the name was absent from both the
builder surface and the expectations table. The exact-match assertion was the
only thing that caught the gap.

That is also why this PR touches the test at all: a new declaration turns the
derived list red automatically, and it takes a table entry to turn it green
again. The test is doing exactly what it was hardened to do.

## Tests

`FabricRegExp` construction is exercised through **both** declared constructor
overloads — from a native `RegExp`, and from an explicit flavor/source/flags
triple — since the declaration offers both and only one is obvious.
`FabricError` is exercised through its state-object constructor and its
`fromNativeError()` static.

Three consuming suites, as this spans three packages: `data-model` 47 passed /
2123 steps, `api` green, `runner` green. `fmt` and `lint` clean.

### DIFF
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -159,6 +159,36 @@ export interface FabricBytesConstructor {
 export declare const FabricBytes: FabricBytesConstructor;
 
 /**
+ * An immutable regular expression. Extends `FabricPrimitive` -- treated like a
+ * primitive in the fabric type system (always frozen, passes through
+ * conversion unchanged).
+ *
+ * The pattern is held as a flavor / source / flags triple rather than as a
+ * native `RegExp`, so that flavors with no native representation can still be
+ * carried. `value` reconstitutes a native `RegExp` where one exists.
+ */
+export interface FabricRegExp extends FabricPrimitive {
+  readonly source: string;
+  readonly flags: string;
+  readonly flavor: string;
+
+  /**
+   * A fresh native `RegExp` equivalent to this value, returned anew on each
+   * call so the internal instance is never aliased out. Throws for a flavor
+   * with no native `RegExp` representation.
+   */
+  readonly value: RegExp;
+}
+
+export interface FabricRegExpConstructor {
+  new (regex: RegExp): FabricRegExp;
+  new (flavor: string, source: string, flags: string): FabricRegExp;
+  prototype: FabricRegExp;
+}
+
+export declare const FabricRegExp: FabricRegExpConstructor;
+
+/**
  * The full set of values that the fabric storage layer can represent.
  */
 export type FabricValue =
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -189,6 +189,68 @@ export interface FabricRegExpConstructor {
 export declare const FabricRegExp: FabricRegExpConstructor;
 
 /**
+ * Structured state for constructing a `FabricError`. The fixed-schema slots
+ * are `FabricValue`-typed; `extras` carries any custom enumerable properties,
+ * whose keys must not collide with the slot names.
+ */
+export type FabricErrorState = {
+  /** Constructor name of the originating native `Error` (e.g. `"TypeError"`). */
+  readonly type: string;
+  /** The `.name` property. Omit to mean "same as `type`". */
+  readonly name?: string | null | undefined;
+  /** The `.message` property. */
+  readonly message: string;
+  /** The `.stack` property, or `undefined`. */
+  readonly stack: string | undefined;
+  /** The `.cause` value, in `FabricValue` form, or `undefined`. */
+  readonly cause: FabricValue | undefined;
+  /** Custom enumerable own properties, in `FabricValue` form. */
+  readonly extras?:
+    | Iterable<readonly [string, FabricValue]>
+    | Readonly<Record<string, FabricValue>>
+    | undefined;
+};
+
+/**
+ * An error carried as fabric data. Extends `FabricInstance` (not
+ * `FabricPrimitive`): it holds fixed-schema slots plus a bag of extras, and
+ * `cause` may be an arbitrary `FabricValue`, so it is a small object graph
+ * rather than a leaf.
+ *
+ * Like every `FabricInstance` it is mutable until frozen: the slots are plain
+ * writable properties, and `setExtra()` / `deleteExtra()` are gated on the
+ * frozen state.
+ */
+export interface FabricError extends FabricInstance {
+  type: string;
+  name: string;
+  message: string;
+  stack: string | undefined;
+  cause: FabricValue | undefined;
+
+  getExtra(key: string): FabricValue | undefined;
+  hasExtra(key: string): boolean;
+  setExtra(key: string, value: FabricValue): void;
+  deleteExtra(key: string): boolean;
+  readonly extraSize: number;
+  extraKeys(): IterableIterator<string>;
+  extraEntries(): IterableIterator<[string, FabricValue]>;
+}
+
+export interface FabricErrorConstructor {
+  new (state: FabricErrorState): FabricError;
+  fromNativeError(error: Error): FabricError;
+  prototype: FabricError;
+}
+
+export declare const FabricError: FabricErrorConstructor;
+
+// TODO(danfuzz): `FabricMap` and `FabricSet` are deliberately absent from the
+// declarations above. Both need substantial rework before they are useful, and
+// declaring them here would imply a utility they do not yet have. Their
+// absence is a decision, not an oversight; revisit once that rework lands.
+
+/**
  * The full set of values that the fabric storage layer can represent.
  */
 export type FabricValue =
--- a/packages/data-model/src/fabric-instances/FabricError.ts
+++ b/packages/data-model/src/fabric-instances/FabricError.ts
@@ -418,6 +418,9 @@ export class FabricError extends FabricNativeWrapper<Error>
 }
 
 // Compile-time check that the exported `FabricError` constructor matches the
-// `FabricErrorConstructor` declared in `@commonfabric/api`. This catches drift
-// between the public type contract and this implementation.
+// `FabricErrorConstructor` declared in `@commonfabric/api`. This catches a
+// declared member that is missing here or has the wrong type. It does NOT
+// catch the other direction: `satisfies` is an assignability check, so a
+// public member on this class that the declaration omits passes silently.
+// Members added here need adding there by hand.
 FabricError satisfies ApiFabricErrorConstructor;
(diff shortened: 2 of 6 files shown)


danfuzz (commented):
## Review: APPROVE (scoped — see below)

Reviewed at **`e1cfb958d74adbb8d6dec6d1cb350bed01e8090a`** (merge-base
`f89e9562d`). Recorded as a comment: under the shared identity GitHub blocks a
formal approval, so `reviewDecision` stays blank by design.

**Scope note:** this tip covers **`FabricRegExp` only** — `git diff` against the
merge-base contains zero mentions of `FabricError`. If a second class is
intended in this PR, it is not in what I reviewed, and it needs its own pass.

### The declaration is complete, and that is measured rather than inferred

The `satisfies` pin and the `implements` clause both pass — but neither
establishes completeness. `satisfies` is an **assignability** check, and a class
with *extra* members is still assignable, so both are structurally blind to a
member the class has and the declaration omits. (`FabricHash` has the same pin
and has drifted in exactly that direction.)

So I checked that direction separately:

```ts
type PublicMissing<Cls, Api> = Exclude<Extract<keyof Cls, string>, keyof Api>;
```

- `FabricRegExp` → **no error**. The declaration covers the whole public surface.
- `FabricBytes` → no error (known-good control).
- `FabricHash` → errors with `"copyInto" | "taggedHashString"` (positive control:
  the check does fire when there is something to find).

With controls in both directions, "complete" here is a measurement.

### Verified

- **Both constructor overloads are declared and both are tested** —
  `new (regex: RegExp)` and `new (flavor, source, flags)`. Easy to declare only
  the obvious one; the four accessors (`source`, `flags`, `flavor`, `value`)
  are all present too.
- **The derived test picks the class up automatically.** Dropping the binding
  from `factory.ts` turns 5 assertions red across three groups — the per-name
  presence and identity checks, the per-class construction tests, and the frozen
  sandbox surface. Full file green at 32 steps otherwise.
- **The test-side change was required, not optional** — a new declaration goes
  red without any test edit, but going green again needs the import and the
  `expectedBindings` entry, both of which are here.
- `value` is documented as returning a fresh `RegExp` per call so the internal
  instance is never aliased out, which matters for a type whose instances are
  meant to be immutable.

### One note, not a request

The pin's comment — *"This catches drift between the public type contract and
this implementation"* — is copied from the existing sites and overclaims in the
direction described above: it catches a wrong type or a declared-but-missing
member, not a member the class has and the declaration lacks. Accurate today
because the declaration was derived member-by-member; the exposure is the *next*
person adding a method to `FabricRegExp`, who would get no signal. Same wording
is already on `FabricHash`, where the drift happened. Worth narrowing the
sentence wherever it is next touched.

danfuzz (commented):
## Re-review (two-class scope): APPROVE

Reviewed at **`13200d530869eff77b884bdd1d3f65fbfde8f5f9`** (merge-base
`f89e9562d`). Supersedes my earlier comment, which was scoped to the
`FabricRegExp`-only tip.

### `FabricError`'s declaration is faithful for every member the class defines

The class's own public surface is `getExtra`, `hasExtra`, `setExtra`,
`deleteExtra`, `extraSize`, `extraKeys`, `extraEntries`, plus the five schema
slots. **All twelve are declared.** For a surface that size, derived
member-by-member rather than composed, that is the right result.

### Two inherited members are absent — pre-existing, not this PR's

Running the under-declaration check that the `satisfies` pins structurally
cannot perform:

```ts
type PublicMissing<Cls, Api> = Exclude<Extract<keyof Cls, string>, keyof Api>;
```

- `FabricError` → **`"deepClone" | "toNativeValue"`**
- `FabricRegExp` → no error
- `FabricBytes` → no error (known-good control)
- `FabricHash` → `"copyInto" | "taggedHashString"` (positive control — the check
  fires when there is something to find)

Both hits are **inherited protocol members, not `FabricError`'s own**, and
neither is introduced here:

- **`deepClone`** is declared abstract on `data-model`'s `FabricInstance`
  (`interface.ts:78`) and implemented on `BaseFabricInstance`. `api`'s
  `FabricInstance` declares **only `shallowClone`** — half the clone protocol.
  That asymmetry is the part that reads as an oversight rather than a decision.
- **`toNativeValue`** comes from `FabricNativeWrapper<Error>`
  (`FabricError extends FabricNativeWrapper<Error>`) and is called by the
  conversion machinery at `native-conversion.ts:598`. `api` declares it nowhere.

They become *newly reachable* here — this is the first concrete `FabricInstance`
handed to pattern code — but the gap is in the base declarations, not in what
this PR wrote. **Not a blocker.** Whether patterns should be able to call either
is a design question, and `shallowClone`-without-`deepClone` is the concrete
thing to settle.

### Verified

- Full file green at **37 steps**; dropping the `FabricError` binding from
  `factory.ts` turns **10 assertions** red across the presence, identity,
  per-class and frozen-surface groups, so the derived test genuinely covers the
  new class.
- The pin comment fix is right, and better than the note I filed. Stating *"a
  public member on this class that the declaration omits passes silently"* at
  the site is worth more than the same fact in a task — and it is exactly the
  direction my check above exercises.

### One thing for the record, not a request

This PR hands pattern code a **mutation surface**: `type`, `name`, `message`,
`stack`, `cause` are declared writable, and `setExtra()` / `deleteExtra()` are
declared. That is faithful — a `FabricInstance` is mutable until frozen and the
mutators are frozen-gated — and it is disclosed in the body. Worth flagging only
because `FabricError` is now the entire concrete surface of the open
"readonly `FabricInstance`" question: two mutator methods, on one class,
reachable from patterns.

PR #5091, opened 2026-07-28
docs(plans): verb contract — record review context (llm-dialog precedent, #5059 preflight, deferred get flags)

bfollington:
## Why

Folds the review context posted on #4968 ([comment](https://github.com/commontoolsinc/labs/pull/4968#issuecomment-5098604859)) into the plan of record, per the plan's own convention ("keep current as work proceeds"; #5006 is the precedent). Additive only — no decision, scope, or workstream changes. A comment thread doesn't survive as context for future readers of the plan; these three notes are worth having in place.

## What Changed

- **Design doc, Prior art:** the llm-dialog builtin recorded as a second in-tree precedent for the invocation protocol — it already mints the result cell at a caller-supplied id (`toolCall.id`), hands it to the handler as `result`, and resolves off the commit callback. The scheduler receipt stays the chosen substrate for the reasons Part 2 gives; llm-dialog is in-production evidence the ergonomics work.
- **Design doc, Discovery:** two deferred `cf piece get` read-control flags recorded — a `--schema` override (the runtime's `asSchema`, which the CLI already uses to narrow its own internal reads, e.g. `piece-render.ts`) and a record limit on large arrays — so the deferral reads as deliberate. Explicitly not asks on any workstream.
- **Implementation doc, Risks:** #5059 (`cf piece setsrc --check`) named as the candidate preflight for the write-storm gate — dry-run of the real rules (schema subset proof, CFC envelope merge, retained-link validator), no second copy. It doesn't measure commit rates, so the rehearsal step stands.
- Both docs get a dated one-paragraph amendment note per house style.

## Test plan

- `docs/` is fmt-excluded; prose hand-wrapped at 80 columns (verified no added line exceeds it).
- Citations checked against source: `toolCall.id` result-cell mint and the `result` comment in `packages/runner/src/builtins/llm-dialog.ts`; `asSchema` on `Cell` (`packages/runner/src/cell.ts:310`); internal narrowing at `packages/cli/lib/piece-render.ts:41`.

---
## Summary by cubic
Adds review context to the verb contract docs so future readers have the in-tree precedents and deferrals in one place. Records the `llm-dialog` invocation precedent, names `cf piece setsrc --check` (#5059) as the write-storm preflight, and documents two deferred `cf piece get` read controls (`--schema` via `asSchema`, and a record limit).

<sup>Written for commit 585fe619bd873beb22101b21d7d2f74fd3d17ece. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5091?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/plans/pattern-verb-contract-implementation.md
+++ b/docs/plans/pattern-verb-contract-implementation.md
@@ -14,6 +14,12 @@ post-merge review amendments (decisions 5–7: attribution via CFC provenance,
 patterns return references while clients render identity, `@name` deferred);
 index and fixture wording follows the references model.
 
+**Amended 2026-07-28**, context only — no scope or decision changed: Risks
+names #5059 (`cf piece setsrc --check`) as the candidate preflight for the
+write-storm gate, and the design doc records the llm-dialog invocation
+precedent (Prior art) and two deferred `cf piece get` read-control flags
+(Discovery).
+
 ## Governing decisions
 
 Made 2026-07-24, shaping everything below:
@@ -327,6 +333,13 @@ change that alters them.
   #4956, merged 2026-07-24 — is the candidate fix; confirm before the first
   Phase 1 deploy). Until the gate clears, a "live-board acceptance pass"
   degrades to a scratch board and must say so rather than pass silently.
+  **#5059 (`cf piece setsrc --check`) is the candidate preflight for this
+  gate:** it answers whether a source can be applied to a given piece before
+  attempting it, driving the real rules in dry-run — the schema subset proof,
+  the CFC envelope merge, the retained-link validator — rather than a second
+  copy of them. It does not measure commit rates, so the rehearsal above
+  stands; what it removes is discovering an incompatibility by attempting the
+  swap on the live board.
 - **WS-E's gates may stall it** (OQ1, CFC review, collection unknown, trusted
   ingress mint and propagation): it is last and severable; everything through
   Phase 4 delivers without it, and `topics.agentName` remains the safe interim.
--- a/docs/plans/pattern-verb-contract.md
+++ b/docs/plans/pattern-verb-contract.md
@@ -20,6 +20,12 @@ not a parallel invocation field; patterns return child references while clients
 render their ids and paths; and client-local `@name` bindings are deferred
 because they overlap confusingly with fabric-side slugs.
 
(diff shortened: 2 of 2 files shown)


mpsalisbury:
Superseded by #5102, which merged as `b4e8ac0d5` — closing this one.

Thanks for these; all three notes are in the plan of record now. Two came across with corrections, and since both are about the argument rather than the prose, worth spelling out:

**The llm-dialog citation was anchored on the branch the doc rules out.** This PR cited `runtime.run(tx, pattern, invocationArgs, result)`, but that's the tool-as-bound-sub-pattern path — the one the "no third callable kind" decision rejects by name — and it resolves off a `sink`, not the commit callback the sentence describes. The precedent is the `handler` branch two lines below: `handler.withTx(tx).send({...input, result}, (completedTx) => …)`. As written, a reader following the anchor landed on the rejected kind and could read it as the doc softening its own decision. Merged version cites the handler branch.

**The timeout is named rather than omitted.** llm-dialog bounds its caller-side wait with a 120s `TOOL_CALL_TIMEOUT`, and its sibling `REQUEST_TIMEOUT` drops a user's message outright when one is pending. Three paragraphs above the insertion point the doc cites "four fixed-timeout sync failures" as motivating evidence, so an unqualified "the ergonomics work" undersold it. The merged text says both: the caller-supplied-id → durable-result-cell ergonomics are proven, and the bounded wait around them is exactly what the design replaces. That makes llm-dialog the strongest evidence in the section instead of an analogy that weakens when someone opens the file.

Your #5059 preflight note went in verbatim, and the deferred `cf piece get` flags went into Discovery at the end of the section rather than mid-argument.

The same pass also turned up a real defect in the tree — `tryResolvePieceHandler` detected a handler through a forced-stream cast and then dispatched on a differently-built cell, so a verb `cf piece verbs` listed could fail on `cf piece call`. Fixed in #5106. And #5104 records a prerequisite for WS-C's C3: "stream/handler properties" isn't one predicate, since `Cell.isStream` accepts three independent signals.

PR #5094, opened 2026-07-28
fix(runner): chunk compile-cache write-backs so an interrupted first-open settle survives (estuary outage)

bfollington:
## Why

Deploying `a8c11e27e` to Estuary broke home-space loading for every existing
identity: cold-opening a space with real data took 29–61s — longer than client
timeouts — while fresh identities worked. Rolling the server back to
`80678fbbc` restored 0.5s opens. This PR fixes the actual cause, so main can be
deployed again.

**The cause is not what we first thought, and the investigation is part of the
record:**

1. First suspect was #4606 (pending-stack dependencies) — the timing fit and it
   touched the storage layer. We said so in team chat. **That was wrong.**
2. A local A/B harness (two worktrees serving the *same* seeded space store,
   identical timed probes) reproduced the behavior change and `git bisect run`
   across the 89-commit window landed on **#4980** (`a0f82ce3c`, sandbox
   type-library strip). #4606, which is later in the window, is exonerated.
3. Mechanism: #4980 changed the transformer's *output* (its own diff updates
   `.expected.jsx` fixtures). Authored pattern identity is deliberately stable
   (`entry-identity.ts` hashes authored source only), but compiled module refs
   (`cf:module/<hash>`) hash the compiled form — so every module's ref changed,
   every derived/`computed:` cell's cause changed, and the first client session
   over any pre-existing space recompiles the closure and rewrites its derived
   state ("the storm": ~3 commits, ~64 revisions, ~664KB locally).
4. **The production killer is the amplifier, not the storm**: the storm's
   dominant transaction was all-or-nothing. A session killed mid-settle
   persisted **zero rows** (measured), so a space whose storm outruns client
   timeouts retries forever and never settles. Estuary was in exactly that
   loop; a *completed* settle makes a space permanently fast again.
5. Cross-version tests pin it to the client: OLD server × NEW client is slow
   and writes the storm; NEW server × OLD client is fast and clean.
   (`engine.open()` on a 60MB space db: 3.5ms old, 3.8ms new — the server was
   innocent. The estuary "server rollback fixed it" observation was because
   estuary serves the shell, so the server rollback rolled the *client* back
   too.)

Storms of this kind will happen again — any commit that changes transform
output re-keys derived state for every space. This PR does not stop the storm;
it makes it **survivable**.

## What Changed

Traced storm anatomy: (1) a single 664KB `writeBackCompileCache` transaction —
286 write attempts, the entire module closure — ~95% of the bytes; (2) the
per-piece `instantiatePattern` transaction (28 writes); (3) a second small
write-back. The closure write-back had equated "a refs-only pattern needs a
durable closure" with "one transaction", which the loaders do not require.

- `compilation-cache/cell-cache.ts`: `planCompileCacheWriteChunks` — orders the
  closure dependencies-first (iterative DFS post-order), **pins the entry
  module to the last chunk**, slices at 4 modules/transaction; threads the
  entry's `extraRoots` so chunked writes cannot corrupt the entry's link set.
- `pattern-manager.ts`: `writeBackCompileCache` commits the chunks sequentially
  through the unchanged `editWithRetry` + CFC-prepare path (per-edge retry
  budget now per chunk). The awaited persistence promise still resolves only
  after all chunks are durable, so the `$patternRef` contract is unchanged.
- **Consistency:** the entry module is pinned to the LAST chunk, and an absent
  entry is exactly both load paths' miss test — so an interruption never
  persists the entry of a namespace that lacked one, and a committed *prefix*
  reads as a plain cache miss. The loaders are deliberately **not** fail-closed
  on missing *descendants*: a partial compiled closure degrades to
  recompile-from-source (and the recovery write-back heals the namespace),
  which the new entry-present/descendant-missing tests pin at both the loader
  and system level.
- The instantiation transaction stays atomic on purpose: it is already
  per-piece (the natural boundary), small, and its atomicity is load-bearing
  for the cold-start repair path.

## Test plan (measured on the harness's 50k-doc dataset, all runs real)

- **Unfixed baseline:** storm = +3 commits/+64 revisions; kill at 1.2s → 0 rows
  persisted (reconfirmed).
- **Fixed, uninterrupted:** +6 commits/+64 revisions, 2.82s vs 2.83s — no
  latency regression; settled states mutually accepted by both clients with
  zero writes in both directions.
- **Fixed, killed repeatedly mid-settle:** session 1 persists +4/+60, session 2
  +1/+4, session 3 opens clean — **monotonic progress, converged after 2
  interrupted sessions**, final cold open 1.1s / zero writes.
- **Fast path:** already-settled space — times and zero-write behavior
  identical to baseline.
- Suites: runner 1075 passed / 0 failed; piece 28 / 0; new
  `compile-cache-incremental-writeback.test.ts` (6 steps) green; check/lint/fmt
  clean.

## Known limits (stated, not hidden)

- The compile phase itself cannot be made durable — its durability *is* this
  write-back. A session killed before the write window still persists nothing.
  Chunking guarantees progress once writing starts; production convergence
  therefore depends on sessions surviving compile, which long-lived browser
  workers do even when individual UI requests time out.
- Chunked layout mints different (link-isomorphic) ids for derived import-edge
  docs than the monolith did; both settled states are accepted by both clients
  (verified bidirectionally).
- `writeBackSourceCache` (source-only path) remains single-tx — not on the
  storm path; conscious scope.
- The deeper issue — transform-output changes re-keying derived state for every
  existing space, with no migration story — is real and bigger than this PR;
  this makes the re-settle survivable rather than preventing it.

---
## Summary by cubic
Make compile-cache write-backs incremental so first-open resettles survive interruptions and spaces converge instead of looping on timeouts. Small closures still commit once; large closures persist progress across sessions.

- **Bug Fixes**
  - Added `planCompileCacheWriteChunks` to order modules dependencies-first, pin the entry last, and split write-backs into 4-module transactions (`COMPILE_CACHE_WRITEBACK_CHUNK_MODULES`).
  - Updated `writeBackCompileCache` to commit chunks sequentially via `editWithRetry`; callers still await until all chunks are durable; per-chunk retries scale with edge count, keeping the single-chunk floor and avoiding multiplying the fixed floor across chunks.
  - Threaded full-set `extraRoots` through `writeSourceDocs`/`writeCompiledDocs`/`writeSourceAndCompiledDocs` so the entry’s synthetic root links stay correct across chunked writes.
  - Guaranteed safe partials: the entry doc lands last, so an interrupted write never persists an entry; an absent entry is a cache miss. If an entry-present/descendant-missing state exists out of band, the compiled loader returns entry + valid subset and the system degrades to a clean recompile from verified source.
  - Added tests for chunk ordering, root-link integrity, interruption survivability, compiled-loader behavior on missing descendants, and system-level degrade-to-recompile that heals the compiled set.

<sup>Written for commit d1ce24310490d84070f4594c2ae01c4e79acd289. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5094?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/compilation-cache/cell-cache.ts
+++ b/packages/runner/src/compilation-cache/cell-cache.ts
@@ -1378,11 +1378,28 @@ export const COMPILE_CACHE_WRITEBACK_CHUNK_MODULES = 4;
  *
  * Ordering makes every prefix of committed chunks a maximally useful partial
  * state: dependencies first (DFS post-order over the import graph), the
- * ENTRY module last. The closure loaders (`loadSourceClosure` /
- * `loadCompiledClosure`) start at the entry and fail closed on any missing
- * doc, so a partial prefix is never misread as a complete closure — it is
- * simply a cache miss whose already-durable docs make the next session's
+ * ENTRY module last.
+ *
+ * What the entry-last pin actually guarantees: an interrupted write-back
+ * never persists the entry doc of a namespace that did not already have one
+ * (a fresh compiled `compileCache:<version>/` namespace, or a fresh space's
+ * source set), and an ABSENT ENTRY is exactly the miss test of both load
+ * paths (`loadCompiledClosure` returns empty; `loadSourceClosure` resolves
+ * undefined). So every partial prefix this ordering can produce reads as a
+ * plain cache miss whose already-durable docs make the next session's
  * re-write cheaper (equal-value re-writes diff to nothing).
+ *
+ * The loaders are deliberately NOT fail-closed on missing DESCENDANTS:
+ * `loadCompiledClosure` drops an absent/unstamped child along with the edge
+ * to it (pinned by "skips compiled import links without integrity"), and the
+ * by-identity hit test is entry presence. Chunk interruption cannot create
+ * an entry-present/descendant-missing state, but should one arise out of
+ * band, it degrades to a clean RECOMPILE rather than a corrupt load: the
+ * source path rejects the partial graph in `loadVerifiedSourceClosure`
+ * (verifySourceDocs' missing-import check) and the compiled path fails
+ * cached-module evaluation, both funneling into the cold recompile fallback
+ * — pinned by the descendant-missing tests in
+ * compile-cache-incremental-writeback.test.ts.
  *
  * `extraRoots` is computed over the FULL module set and must be passed to
  * each chunk's write (the entry doc's synthetic root links enumerate every
--- a/packages/runner/src/compilation-cache/cell-cache.ts
+++ b/packages/runner/src/compilation-cache/cell-cache.ts
@@ -441,8 +441,14 @@ export function buildSourceDocs(
   modules: readonly CacheableModule[],
   entryIdentity: string,
   moduleDelegations: ModuleDelegationMap = EMPTY_MODULE_DELEGATIONS,
+  extraRootsOverride?: readonly string[],
 ): Map<string, SourceDoc> {
-  const extraRoots = unreachedRoots(modules, entryIdentity);
+  // `extraRootsOverride` exists for chunked (incremental) write-backs: the
+  // synthetic root links on the ENTRY doc must be computed over the FULL
+  // emitted module set, while `modules` here may be just the chunk being
+  // written in this transaction (see planCompileCacheWriteChunks).
+  const extraRoots = extraRootsOverride ??
+    unreachedRoots(modules, entryIdentity);
   const out = new Map<string, SourceDoc>();
   for (const module of modules) {
     const delegatedModuleIdentities = validDelegatedModuleIdentities(
@@ -744,6 +750,7 @@ export function writeSourceDocs(
   entryIdentity: string,
   tx: IExtendedStorageTransaction,
   moduleDelegations: ModuleDelegationMap = EMPTY_MODULE_DELEGATIONS,
+  extraRootsOverride?: readonly string[],
 ): ModuleDelegationMap {
   assertNoUnpinnedFabricImports(modules);
   const effectiveModuleDelegations = effectiveModuleDelegationsForWrite(
@@ -758,6 +765,7 @@ export function writeSourceDocs(
     modules,
     entryIdentity,
     effectiveModuleDelegations,
+    extraRootsOverride,
   );
   withCompileCacheBuiltin(tx, () => {
     for (const [identity, doc] of docs) {
@@ -1211,6 +1219,8 @@ export function writeCompiledDocs(
   opts: {
     runtimeVersion: string;
     moduleDelegations?: ModuleDelegationMap;
+    /** See {@link buildSourceDocs}: full-set root links for chunked writes. */
+    extraRoots?: readonly string[];
   },
   tx: IExtendedStorageTransaction,
 ): ModuleDelegationMap {
@@ -1223,7 +1233,8 @@ export function writeCompiledDocs(
     opts.moduleDelegations ?? EMPTY_MODULE_DELEGATIONS,
     { compiledRuntimeVersion: opts.runtimeVersion },
   );
-  const extraRoots = unreachedRoots(modules, entryIdentity);
+  const extraRoots = opts.extraRoots ??
+    unreachedRoots(modules, entryIdentity);
   const schema = compiledDocWriteSchema();
   withCompileCacheBuiltin(tx, () => {
     for (const module of modules) {
@@ -1304,6 +1315,8 @@ export function writeSourceAndCompiledDocs(
   opts: {
     runtimeVersion: string;
     moduleDelegations?: ModuleDelegationMap;
+    /** See {@link buildSourceDocs}: full-set root links for chunked writes. */
+    extraRoots?: readonly string[];
   },
   tx: IExtendedStorageTransaction,
 ): ModuleDelegationMap {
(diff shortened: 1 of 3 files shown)


bfollington:

Scope: PR #5094 · detail review of all 3 changed files plus both closure loaders, PatternManager persistence/load call chains, and CT-1824 retry semantics · checks run: targeted typecheck, fmt, lint, new incremental-writeback test, writeback-conflict suite, ESM cache integration suite (all green).

## 🔴 Blocking

### Major — the claimed “any missing doc => cache miss” invariant is not true, and the new test only proves the easier absent-entry case

[`planCompileCacheWriteChunks`](https://github.com/commontoolsinc/labs/blob/520a4389c6fca8e5ee1f5d782034a0f034d88cc3/packages/runner/src/compilation-cache/cell-cache.ts#L1379-L1385) and the write-back comment say both loaders fail closed on *any* missing document. They do not:

- `loadSourceClosure` skips a missing/non-cell child; only `loadVerifiedSourceClosure` can subsequently reject the reconstructed source graph.
- [`loadCompiledClosure`](https://github.com/commontoolsinc/labs/blob/520a4389c6fca8e5ee1f5d782034a0f034d88cc3/packages/runner/src/compilation-cache/cell-cache.ts#L1526-L1541) also skips a missing/unstamped child and returns the entry plus whatever valid subset remains. Existing coverage explicitly asserts this behavior (`cell-cache.test.ts`: “skips compiled import links without integrity” expects a one-document result with the import erased).

The new interruption test seeds an empty space and stops before the entry is written, so both reads necessarily miss at the entry; it never exercises the stated “entry exists, descendant missing” invariant. That distinction matters for repair/same-version states, and `loadPatternByIdentityFromStorage` initially treats `closure.has(entryIdentity)` as the hit test before the structural graph verifier catches a partial graph.

Pinning the entry to the last chunk is sufficient for the outage’s fresh compiled-runtime namespace (the entry is absent until the final atomic chunk), but it is not evidence that the loaders themselves fail closed on arbitrary missing descendants. Please either:

1. make the compiled loader return an empty map on any malformed/missing/unstamped linked child (and update the source comment to name `loadVerifiedSourceClosure`, not the raw loader), with a regression test that pre-seeds an entry whose descendant is absent; or
2. narrow the consistency argument everywhere to the property actually proved: a prefix written into a namespace with no pre-existing entry is a miss, while higher-level completeness/graph verification rejects other partial closures.

Given that this consistency claim is the safety argument permitting loss of transaction atomicity, I would not merge with the current overstatement/test gap.

## 🟡 Improvements

### Minor — the CT-1824 budget is correct per chunk but its aggregate ceiling is multiplied by chunk count

[`Math.max(16, 2 * importEdges + 8)` now runs for every chunk](https://github.com/commontoolsinc/labs/blob/520a4389c6fca8e5ee1f5d782034a0f034d88cc3/packages/runner/src/pattern-manager.ts#L1830-L1858). The `2 * edges` part still matches the two source/compiled edge-doc conflicts a chunk may discover, so convergence is locally correct. However, the minimum/slack is no longer closure-wide: a 24-module low-edge closure has six independent ceilings of 16 retries instead of the former single ceiling of 16. This does not slow conflict-free writes, but under persistent contention it can substantially extend failure latency.

Please make that inflation deliberate and measured, or preserve a closure-wide remaining budget while allocating enough per chunk for its edge docs. A test should capture the intended bound; the new test bypasses `editWithRetry` entirely, so it cannot guard this behavior.

## Requested invariants checked

- `extraRoots`: correct for concurrent writers of the same closure. It is computed once from the full module set, passed unchanged to every chunk, consumed only when the entry module is written, and the entry write replaces the complete link set atomically. Identical concurrent closures therefore cannot compose a mixed root set.
- `$patternRef` persistence: correct by inspection. The chunk loop awaits each `editWithRetry`; `writeBackCompileCache` resolves only after the loop; `persistCompileCacheTracked` awaits it before resolving; the compile path awaits that promise before returning the pattern. I would still add a >4-module integration test that gates the final chunk and proves `compilePattern` remains pending, because the added test calls the low-level write helper directly.
- Fast path: no regression found. A settled storage hit still sets `warmHit` and skips persistence entirely; closures of ≤4 modules still use one write transaction. The existing warm-hit and cross-session integration tests passed.

**Verdict:** request changes for the unproven/incorrect fail-closed consistency contract; the core chunking and promise threading otherwise look sound.

bfollington:
_cf-review re-review via Codex (GPT-5), on behalf of @bfollington — at `d1ce24310`_

Re-review of the blocker fix. Findings, from the review run log:

> 🔴 Blocking: the prior blocker is not fully closed. The implementation comments and new tests now describe the real guarantee correctly, but the human-authored PR body still claims [the loaders fail closed on any missing doc].
>
> **Verdict:** request changes solely because the live PR body still repeats the false consistency claim. Once that bullet is corrected, I found no remaining code or test blocker at `d1ce24310`.

**Resolution:** the PR body's Consistency bullet has since been rewritten to state the actual guarantee (entry-last pin; absent entry is the miss test; descendants deliberately degrade to recompile-from-source with the recovery write-back healing the namespace, pinned by the new loader- and system-level tests). With that correction applied, the re-review's stated condition for approval is met: **no remaining code or test blocker.**

PR #5100, opened 2026-07-28
feat(cli): list indexed piece roots without starting code

Hixie:
Piece listing previously had to choose between traversing the registry, which omitted orphan roots and started registered pieces, and enumerating every stored entity, which returned non-pieces.

Maintain a server-side piece-root and summary index in the same transaction as default-branch writes. Track summary and registry dependencies. Preserve registry order, canonicalize URI and scope identities, and paginate against a server sequence. Rebuild when a storage watermark shows that an older binary advanced the database.

Use the new protocol from the runner and CLI. Preserve the explicit registry compatibility fallback. Cover hidden children, orphans, scoped roots, rollback migration, pagination, and value-free transport.

---
## Summary by cubic
Adds a server-side, snapshot-stable piece-root listing and updates the CLI to list roots without starting any code. `piece-root.list` is unconditional, backed by a lazy server index with snapshot-safe paging; tests assert identifiers instead of computed names.

- **New Features**
  - Server builds and serves a piece-root+summary index across four SQLite tables, maintained lazily (after fan-out or when a listing sees pending work). Preserves registry order, canonicalizes IDs/scopes, tracks dependencies, and paginates via `serverSeq` with cursors; no piece values are transferred.
  - Protocol adds unconditional `piece-root.list`; continuations require `expectedServerSeq` and support a `registeredOnly` filter.
  - Runner/manager/client: expose `listPieceRootPage` and a helper to collect a full snapshot; `PieceManager.listPieceRoots` added; storage provider forwards requests.
  - CLI `cf piece ls` lists registered, unregistered, scoped, orphan, and hidden-child roots without starting pieces or loading values; shows registration, scope, and `entityKind` in table/JSON; adds `--registry` and falls back when the server lacks `piece-root.list`.
  - Tests and docs: new fixtures for hidden/unregistered children; integrations assert exact piece IDs and preserve a schema-backed execution marker; protocol and CLI README updated; design and cost notes archived in history.

- **Migration**
  - No action needed. The CLI automatically falls back when the server does not support `piece-root.list`.

<sup>Written for commit 7d96df41f1911c0228fe4e61f40a591a3c499b41. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5100?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Hixie:
chatted with @seefeldb; we decided not to land this. will instead land documentation that says you cannot see orphan pieces and that other pieces can only be found by searching or walking from the piece registry.

PR #5102, opened 2026-07-28
docs(plans): verb contract — fold in review context and structural findings

mpsalisbury:
## Why

Two threads of context were sitting outside the plan of record and would not
have survived there: Ben's review comment on #4968
([link](https://github.com/commontoolsinc/labs/pull/4968#issuecomment-5098604859)),
and a follow-up reading pass over `llm-dialog`, the schema-compatibility
checker, and the CLI callable surface.

This supersedes the draft #5091, which folded in the comment alone. Two of its
three notes are carried here with corrections; the rest is new. Additive only —
no decision, scope, or workstream changed. Precedent for amending in place:
#5006.

The corrections matter enough to be the reason this is a separate PR rather
than a review on #5091:

- #5091 cites the llm-dialog precedent at `runtime.run(tx, pattern, …)`. That
  is the **tool-as-bound-sub-pattern** branch — the path this document rules
  out by name — and it resolves off a sink, not the commit callback the
  sentence describes. The precedent is the **handler** branch two lines below.
  As written, a reader following the anchor lands on the rejected callable kind
  and can read it as the doc softening its own decision.
- #5091 presents llm-dialog as unqualified evidence "the ergonomics work."
  llm-dialog bounds its caller-side wait with a 120s `TOOL_CALL_TIMEOUT`, and
  its sibling `REQUEST_TIMEOUT` silently drops a user's message when a request
  is pending. Three paragraphs above the insertion point, this same document
  cites "four fixed-timeout sync failures" as motivating evidence. Naming the
  timeout makes llm-dialog the strongest evidence in the section instead of an
  analogy that weakens the moment someone opens the file.

## What Changed

**Design doc**

- **Prior art:** the llm-dialog handler branch recorded as a second in-tree
  precedent, with both qualifications above stated inline.
- **Decisions:** "No third callable kind" restated as *tools are deferred, not
  rejected*. A tool is not a distinct callable kind — both branches of the
  runtime's tool path send an invocation and get a durable result cell at a
  caller-supplied id; the pattern branch merely instantiates first. Records the
  two properties that keep a create verb additive later, one of which
  (`FrameworkProvided<T>`) already exists at the type layer.
- **Discovery:** the two deferred `cf piece get` read-control flags; and the
  property that a verb set is *already* a structural interface, because schemas
  are the type system and the argument/result variance directions already
  match. Explicit interfaces deferred, with the one precondition that would
  silently cost the property — verbs must stay in the piece's own schema, not
  move to a side index.
- **Rule 6:** the second reason it is load-bearing. It is what keeps a verb set
  an interface rather than a session type, so conformance stays decidable from
  the schema.
- **Results and schema evolution:** "results may narrow freely" governs values,
  not named fields — removing a named property is rejected outright, and
  `asCell` is pinned by exact equality. Verb names and verb-ness are already a
  contract with teeth. Practical consequence: return results inside an envelope
  so only one key is permanent.
- Stale citation fixed: the `Transaction required for .set()` error is at
  `cell.ts:1316`, not `:1294`.

**Implementation doc**

- **Risks:** #5059 (`cf piece setsrc --check`) named as the candidate preflight
  for the write-storm gate. It does not measure commit rates, so the rehearsal
  step stands.
- **WS-F:** a read-path guard — `cf piece get` on a verb path returns the
  stream's serialization instead of redirecting; llm-dialog's `read` tool
  already rejects this with the right message.
- **Non-goals:** constraining direct data writes. `cf piece set` writes a result
  field past every rule in Part 1, and the LLM tool surface has no equivalent —
  it can only mutate through verbs — so the CLI is the outlier. Recorded as a
  non-goal rather than addressed: constraining it needs a declarative notion of
  pattern-managed state that this plan does not decide, and `set` has authoring
  and debugging uses this contract does not touch. Named so the absence reads
  as intent, which is what that section is for.

Both docs carry a dated amendment note per house style.

## Deliberately not included

- A reframing of the nested-stream blocker as a stream *misclassification* bug.
  The mechanism checks out — `send()` delegates to `set()`, which throws that
  error only in the non-stream branch — but `resolvePieceCallable` resolves a
  flat callable name with no path traversal, so the nested case fails earlier
  with "Callable not found". The reframe does not hold and stays out.
- Any behavior change to `cf piece set` — only the non-goal entry above, which
  states the gap without taking a position on closing it.

## Test plan

- `docs/` is fmt-excluded; prose hand-wrapped at 80 columns (verified no added
  line exceeds it).
- Every citation checked against source at the cited lines: the handler-branch
  mint and `result` handoff (`llm-dialog.ts:2730-2735`), the `runtime.run`
  sibling (`:2728`), both timeouts (`:126`, `:3053-3062`), the result envelope
  in both branches (`:2789-2797`, `:2818-2826`), named-field removal and the
  `asCell` equality check (`schema-compatibility.ts:437-444`, `:100-106`,
  `:329-333`), `FrameworkProvided` (`api/index.ts:2252-2270`), the CLI's own
  narrowing (`piece-render.ts:41`), and `cell.ts:1316`.

---
## Summary by cubic
Adds review context to the verb contract docs, clarifies the call protocol, and sharpens the structural-interface stance. Records the `llm-dialog` handler precedent, defers tools, adds a read-path guard for `cf piece get`, leaves direct `cf piece set` writes out of scope, and states the result-schema prerequisite directly.

- **Refactors**
  - Design: recorded `llm-dialog` handler-branch precedent; restated tools as deferred with two guardrails (result envelope can carry a reference; payloads may include `FrameworkProvided<>` fields).
  - Structure: captured that interface machinery exists but verb identification is unsettled (three verb-ness encodings; CLI forced-stream fallback); clarified Rule 6’s role; advised returning results in an envelope and treating named fields as permanent.
  - Implementation: added a read-path guard so `cf piece get` on a verb returns the stream serialization; named `#5059` (`cf piece setsrc --check`) as the write-storm preflight; recorded direct `cf piece set` writes as out of scope.

- **Bug Fixes**
  - Repointed drifted citations and added symbol anchors for stability (e.g., `handleInvoke`, `Cell.isStream`, `addMessage`); refreshed line references (e.g., `cell.ts:1347`, `api/index.ts:2264-2282`), fixed hyphenation, tightened claims around collection/timeouts, and rewrapped a ragged paragraph.

<sup>Written for commit cb65a497c25bc1178d9089a448e142308f1f40e0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5102?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/plans/pattern-verb-contract-implementation.md
+++ b/docs/plans/pattern-verb-contract-implementation.md
@@ -202,10 +202,11 @@ Size L (~1–2 weeks). No dependencies; starts immediately.
   properties" is not a single predicate today. `isStream()` accepts three
   independent signals, any one of which suffices: the cell's construction kind,
   `asCell: ["stream"]` in the schema, and a stored `{$stream: true}` value
-  (`packages/runner/src/cell.ts:924-946`). If C3 keys emission off the schema
-  marker alone, a verb carrying only the stored marker gets no result schema
-  and rule 3 silently does not apply to it — and the WS-C exit below would not
-  catch that, since it exercises one CTS pattern that does carry the marker.
+  (`Cell.isStream`, `packages/runner/src/cell.ts:936-958`). If C3 keys emission
+  off the schema marker alone, a verb carrying only the stored marker gets no
+  result schema and rule 3 silently does not apply to it — and the WS-C exit
+  below would not catch that, since it exercises one CTS pattern that does
+  carry the marker.
   That the signals diverge in practice is not hypothetical: the CLI has two
   runtime workarounds for handlers whose schema lost the marker
   (`tryResolvePieceHandler`, and the forced-stream probe in
--- a/docs/plans/pattern-verb-contract-implementation.md
+++ b/docs/plans/pattern-verb-contract-implementation.md
@@ -4,6 +4,13 @@
 [`pattern-verb-contract.md`](pattern-verb-contract.md) (PR #4968). Keep current
 as work proceeds: check off exit criteria, record scope changes.
 
+**Amended 2026-07-28**, context only — no scope or decision changed: Risks
+names #5059 (`cf piece setsrc --check`) as the candidate preflight for the
+write-storm gate, and WS-F gains a read-path guard so `cf piece get` on a verb
+redirects to `cf piece call`. The design doc carries the same pass's larger
+share: the llm-dialog handler-branch precedent, tools restated as deferred
+rather than rejected, and the structural-interface property.
+
 **Amended 2026-07-24** from the first live headless session (a three-topic
 graph, ~24 CLI operations): compact discovery index (A2), closed-world inputs
 (C5), transaction-local acknowledgement as a WS-D exit criterion, phase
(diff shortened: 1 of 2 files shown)


mpsalisbury:
@bfollington You left your PR as a draft, but I think it's worth pulling in, so it's added here too. a couple of changes noted above in the pr description. ptal.

PR #5108, opened 2026-07-28
fix(runner): make slow-traverse report coverage independent of machine speed

mpsalisbury:
## Why

The `coverage-debt: packages/runner uncovered lines` metric is bimodal. It reads
either **6909** or **6948** on `main` — a swing of exactly 39 lines — depending
on nothing but how fast the CI machine happened to be.

Those 39 lines are the `elapsed > 100` slow-traverse diagnostic in
`traverse.ts`, plus the two `MapSet` getters that only that block calls. They
run only when a single traversal exceeds 100 ms of real wall time. In one `main`
run the block was hit **exactly once**, in the `pattern-integration-4` shard
alone; in the next run, never. Diffing the raw LCOV between those two runs,
`traverse.ts` is the *only* file in the package whose count differs — every one
of the other 250 files is identical.

That makes the ratchet in `docs/development/COVERAGE.md` a coin flip. The
baseline is the latest non-cold `main` run, so when `main` lands on the lucky-low
6909, every subsequent pull request whose traversals all stay under 100 ms comes
in at 6948 and fails `+39` with no code cause. #5074 is a clean example:
`OVER 6913 6952 +39 (+0.6%)` on a change that removed no tests.

This is the failure mode `AGENTS.md` warns about for timeouts and sleeps —
behavior that depends on unpredictable timings aligning — applied to coverage
instead of to a test.

## What Changed

- Named the threshold `SLOW_TRAVERSE_MS`, and moved both it and the report body
  into `SchemaObjectTraverser.maybeReportSlowTraverse()`. The call site invokes
  it unconditionally, so every line on the traversal path runs on every
  traversal regardless of elapsed time.
- Tests reach the report by passing an `elapsed` over the threshold, rather than
  by hoping a machine is slow. There is no sleep, poll, or timing assumption.
- `MapSet.size` / `totalValues` exist only to fill in that report, so they rode
  the same coin flip. They now have direct tests in both the reference-equality
  and hash-dedup modes.

Production behavior is unchanged: same 100 ms threshold, same lazy `logger.warn`
thunk, same payload, same `CF_TRAVERSE_DIAGNOSTICS` gating of `topDocs`.

Note this deterministically *lowers* `packages/runner` debt, re-seeding the
ratchet at the stricter of the two values instead of leaving it to chance. It
does not retroactively fix #5074 or #5096 — those need a rebase once this lands.

## Test plan

- `deno task test` in `packages/runner`: **1075 passed, 0 failed** (4m54s).
- `deno check`, `deno fmt --check`, `deno lint` clean on both touched files;
  `deno.lock` untouched.
- Local `DENO_COVERAGE_DIR` run over `traverse.test.ts` confirms the previously
  flaky lines are now covered on every run:

  | lines | before | after |
  |---|---|---|
  | `MapSet.size` / `totalValues` | flaky | 12/12 covered |
  | call site | flaky | covered |
  | report body | flaky | 35/35 covered |

---
## Summary by cubic
Make slow-traverse diagnostics deterministic by moving the threshold and report into `SchemaObjectTraverser.maybeReportSlowTraverse()`, called on every traversal. This stabilizes `packages/runner` coverage by removing machine-speed variance without changing production behavior.

- **Bug Fixes**
  - Added `SLOW_TRAVERSE_MS` (100 ms) and call `maybeReportSlowTraverse(elapsed, doc)` unconditionally from `traverse()`.
  - Tests drive a real `traverse()` with `performance.now` overridden and a clock-advancing store; verify 150ms reports, 5ms is silent, and the boundary is exact (100 silent, 101 reports).
  - Added direct tests for `MapSet.size` and `totalValues` in reference-equality and hash-dedup modes.
  - Same lazy `logger.warn` payload and `CF_TRAVERSE_DIAGNOSTICS` gating; coverage for `packages/runner` is now stable.

<sup>Written for commit f403e10d1aedc5c7113ca91b4adb5d07ee08cf16. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5108?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/traverse.ts
+++ b/packages/runner/src/traverse.ts
@@ -617,6 +617,18 @@ function prepareAnyOf(
 }
 
 /**
+ * Wall time (ms) above which a traversal reports its stats. The threshold and
+ * the report body live in `maybeReportSlowTraverse()` rather than inline at
+ * the call site: a threshold on real elapsed time decided whether ~40 lines
+ * ran, so leaving it inline made their coverage a function of how fast the
+ * machine happened to be. That swung the `packages/runner` coverage-debt
+ * metric by 39 lines between otherwise identical CI runs, and the ratchet in
+ * `docs/development/COVERAGE.md` then failed whichever pull requests landed on
+ * the fast side of a baseline recorded on the slow side.
+ */
+const SLOW_TRAVERSE_MS = 100;
+
+/**
  * Per-call doc-visit/unique-path diagnostics in `traverseWithSchema` build a
  * string and touch a Map+Set on EVERY schema visit — measurable in tail
  * traversals (thousands of visits each). They only feed the slow-traverse
@@ -2992,38 +3004,7 @@ export class SchemaObjectTraverser<V extends FabricValue>
     const rv = this.traverseWithSelector(doc, this.selector, link);
     const { error } = rv;
     const elapsed = logger.timeEnd("traverse") ?? 0;
-    if (elapsed > 100) {
-      // Find top visited docs
-      const topDocs = [...this.docVisits.entries()]
-        .sort((a, b) => b[1] - a[1])
-        .slice(0, 5)
-        .map(([id, count]) => `${id.slice(0, 20)}..=${count}`)
-        .join(" ");
-      logger.warn("slow-traverse", () => [
-        `${elapsed.toFixed(0)}ms`,
-        `doc=${doc.address.id}/${doc.address.type}`,
-        `trackerKeys=${this.schemaTracker.size}`,
-        `trackerVals=${this.schemaTracker.totalValues}`,
-        `traverseSchema=${this.traverseWithSchemaCalls}`,
-        `traversePtr=${this.traversePointerCalls}`,
-        `traverseArr=${this.traverseArrayCalls}`,
-        `traverseObj=${this.traverseObjectCalls}`,
-        `traverseDAG=${this.traverseDAGCalls}`,
-        `anyOfBranches=${this.anyOfBranches}`,
-        `anyOfFastRejects=${this.anyOfFastRejects}`,
-        `anyOfPropertyMerges=${this.anyOfPropertyMerges}`,
-        `getDocAtPath=${this.getDocAtPathCalls}`,
-        `dagMemo=${this.dagMemo.size}`,
-        `uniqueDocs=${this.docVisits.size}`,
-        `uniquePaths=${this.uniquePaths.size}`,
-        `maxDepth=${this.maxDepth}`,
-        `schemaMemo=${this.activeMemo.size}`,
-        `schemaMemoHits=${this.schemaMemoHits}`,
-        TRAVERSE_DIAGNOSTICS
-          ? `topDocs=${topDocs}`
-          : "topDocs=n/a (set CF_TRAVERSE_DIAGNOSTICS=1)",
-      ]);
-    }
+    this.maybeReportSlowTraverse(elapsed, doc);
     if (error !== undefined) {
       // This helps track down mismatched schemas, but may be fine
       logger.debug("traverse", () => [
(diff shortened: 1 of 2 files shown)


mpsalisbury:
Added the bounding test, and checked the fake-clock question — short answer: the fake clock doesn't cover this, and can't.

## Bounding test

Fair catch that the first version only called `maybeReportSlowTraverse()` directly. That proved the report body worked but left the call site untested — deleting the call entirely would still have passed.

It now drives a real `traverse()` with `performance.now` overridden for the duration, so the elapsed time the gate sees is *chosen* rather than measured. `traverse()` is synchronous, so a test can't step the clock from outside while one is in flight; a store subclass advances it when the traversal loads the linked doc, which is the hook that runs mid-traversal. `readStore` is asserted so a traversal that never reads can't pass vacuously.

Three cases pin the gate:

| elapsed | expected |
|---|---|
| 150ms | reports |
| 5ms | silent |
| 100ms | silent (gate is exclusive) |
| 101ms | reports |

Verified by mutation rather than by assertion alone:

- `elapsed <= SLOW_TRAVERSE_MS` → `<` (i.e. `>` becomes `>=`): boundary case fails.
- call site deleted: two of the three fail.

## On the fake clock

`packages/runner` has been on the fake clock since dc32786c9 (#4988, 07-24), and it makes this *worse*, not better. Auto-advance mode replaces `performance.now` with logical time, and logical time only moves when a positive-delay timer fires — but `traverse()` is fully synchronous, so none can fire mid-traversal. Probed it directly in the package under its real preload:

```
LOGICAL_ELAPSED=0    (after 156ms of real wall time)
```

So `elapsed > 100` is deterministically false in every runner unit test. The fake clock doesn't cover that branch, it makes it unreachable — which is why the new test overrides `performance.now` explicitly instead of relying on the harness.

The ±39 swing never came from the runner suite anyway. In main run 30318486373, `traverse.ts:3000` had `hits=1` in **`pattern-integration-4` only**, and `hits=0` in all five `coverage-profile-runner-*` shards — four days after the runner fake clock landed. `packages/patterns` has no clock preload, and the pattern-integration coverage comes from patterns running in a **browser worker**, which a Deno `--preload` doesn't reach. So the two mechanisms are complementary: the fake clock removes real-time dependence from Deno-side test waits; this removes it from a production wall-clock branch whose coverage was being sampled by a browser suite.

Also rebased onto main (was 14 behind). Full runner suite: 1075 passed, 0 failed.

mpsalisbury:
## Why this is still worth merging, given the fake clock

Short version: the fix doesn't work the way "add a fake clock" works, and the problem is live on `main` right now.

### The swing is still happening, today

`coverage-debt: packages/runner uncovered lines`, from the `perf-metrics` artifact of each `main` run over the last few hours:

| time (UTC) | value | delta | commit |
|---|---|---|---|
| 21:21 | 6909 | | #5122 |
| 21:51 | 6937 | **+28** | #5094 chunk compile-cache write-backs |
| 21:58 | 6898 | **−39** | #5124 stamp every click mark additively |
| 22:06 | 6896 | −2 | #5129 |
| 23:11 | 6896 | | #5103 |
| 23:15 | **6935** | **+39** | #5126 `cf completion` — shell completion for bash and zsh |

A **bash/zsh shell-completion PR moved a `packages/runner` coverage metric by exactly 39 lines**. It obviously didn't. That is the block in this PR flipping.

The `+28` at 21:51 decomposes the same way: −11 of real coverage gained, plus the block flipping uncovered. That is why #5094 needed a marker at all.

`main` is currently sitting on the **high** side (6935), so the next PR measured against a low baseline eats a phantom +39.

### Why the fake clock can't substitute

The mechanism here isn't "make the timing deterministic in tests" — it's **cover the block unconditionally so the incidental coverage stops moving the total**. Union with a guaranteed-covered set is constant, whatever any clock does in any suite.

The fake clock can't reach the source anyway, and not just because of migration order:

- Those 39 lines were only ever covered from `pattern-integration`. In main run 30318486373, `traverse.ts:3000` had `hits=1` in **`pattern-integration-4` only** and `hits=0` in all five `coverage-profile-runner-*` shards.
- That coverage comes from a real browser plus a real server. A Deno `--preload` doesn't reach it, and `packages/patterns` has no preload to extend.
- `packages/runner` has been on the fake clock since dc32786c9 (#4988, 07-24) — four days before those runs. It was already in force the whole time the swing was happening.

**Correction to my earlier comment:** I said the fake clock "makes this worse, not better." That was overstated and I retract it. Accurate version: auto-advance pins `elapsed` to 0 (measured: `LOGICAL_ELAPSED=0` after 156 ms of real wall time), so the branch is simply unreachable in the runner suite. The fake clock neither causes nor cures this flake — it's orthogonal.

### The honest counter-argument

If `packages/patterns` integration ever did get a fake clock, it would also kill this flake, by pinning `elapsed` to 0 so the block is deterministically *uncovered*. So this isn't the only possible fix. I'd still prefer this one:

- it removes the dependency at the source rather than papering over it in the harness;
- under that alternative the block ends up permanently uncovered — live production diagnostic code with no test at all, whereas this keeps it pinned on both sides of the gate;
- and it exists now.

### What it has cost so far

Within one week: #5074 failed at 6952; #5076 was offered a 6948 marker and held it; #5125 spent one at 6948; and #5087 has *"If this check goes red on a re-run, re-run it — do not add ACCEPT_COVERAGE_DEBT"* written into its description.

Across 150 merged PRs, 14 (9.3%) carry an `ACCEPT_COVERAGE_DEBT` marker. Some fraction of those are this, not real debt — which is the gate training people to route around it.

mpsalisbury:
@Hixie In addition to the comments above, help me understand why a hard limit is appropriate. It seems to me that this means we need 100% coverage for all new code (or we clean up old code to make up the difference). Is that true? And if so, is that your intent? If so, my guess is that's a good way to nudge people into skipping the requirement.

Hixie:
PR looks fine.

100% coverage was my goal (on the basis that any code that isn't tested is definitely buggy, and we don't want to ship buggy code, combined with the fact that agentic coding makes covering code easy and testing in general even more important), but I'm open to alternatives. I just don't really know what a good alternative would be. Whatever the goal is, that's the goal we'll hit, so e.g. if we say 80% of new code needs coverage, we're going to hit about 80% of new code being covered. That doesn't seem better.

PR #5110, opened 2026-07-28
fix(memory,runner): scan pending reads from their declared confirmed basis (CT-1910)

ubik2:
## Why

The pending-read staleness scan starts at the highest dependency's resolution seq, so overlapping FOREIGN writes landing between the reader's true confirmed basis and that seq are never scanned — the CT-1910 basis over-advance, the one open INV-1 violation in the catalog (#5015 machine-finds it: `PendingStacks_Current.cfg` violates `ReadCoherence`; the engine repro is the one verified in the #4606 review thread). Pre-existing on `main` since pending reads existed; #4606's full-stack dependency recording doesn't touch it — the scan *interval* is wrong, not the dependency set.

## What

The minimal TLA-certified repair: flip the staleness basis, keep everything else.

**Wire — one optional field, no negotiation.** `PendingRead.basisSeq?: number`: the reader's confirmed basis for the document, in the SERVER's seq space (the value the confirmed-read branch always emitted and the pending branch discarded — that discard was the bug). Servers ignore unknown read fields, so clients attach it unconditionally: old server → legacy basis (no worse), old client → legacy basis. The repair narrows monotonically with deployment; no hello capability.

**Engine.** When `basisSeq` is declared (validated: integer, `0 ≤ basisSeq ≤ head`), the scan covers the FULL interval `(basisSeq, head]` and excludes the session's own accepted commits (exclusion variants of the two conflict-scan statements, joining `revision.commit_seq → commit.session_id`). Sound by FIFO admission: every own accepted commit in the interval has a lower localSeq than the reader and was part of its materialized view — the self-conflict that historically forced the basis to over-advance. Dependency arrays and their resolution edges are unchanged (a composite built on an unprocessed rejection still dies via `pending dependency not resolved`, so no verdict watermark is needed). Same fix in the scheduler-observation read path, whose exemption mirrored the same semantics. Legacy readers (no `basisSeq`) keep the max-dependency basis byte-for-byte.

**Client.** `pushCommitRead`'s pending branch emits `basisSeq: confirmedSeq ?? record?.confirmed.seq ?? 0`; `scalarizePendingReadStacks` carries the field through the old-server downgrade untouched. In-flight registry, cascade, and the old-server hold are untouched.

**Same-change obligations from #5015's discipline:**
- `naive-admission.ts` implements the repaired rule (basis + own-session exclusion when declared; legacy max-dep otherwise).
- The differential harness emits both shapes 50/50 and gains a post-hoc **pending-read INV-1 re-scan** — newly possible because the basis is now durable in the stored `original`.
- `v2-pending-read-basis-overadvance.test.ts`: the legacy-shape unsound accept stays as the regression witness (the same role the violated TLA config plays), joined by repaired-shape tests: rejects the foreign write in the gap, own-session exclusion admits an own-writes-only interval, a later honest basis commits coherently, and the `basisSeq` validation surface.
- Docs: 03-commit-model §3.4/§3.5/§3.6.3 make the basis-selection rule normative with the soundness argument; 09-invariants INV-1 known-deviation rescoped to legacy `basisSeq`-less traffic; tla/README maps Current/Repaired configs to the two live shapes; EXPERIMENTAL_OPTIONS notes the repair landed capability-free.

## Verification

- `packages/memory`: 449 tests pass, including the six CT-1910 basis tests and the differential harness's 100 seeded schedules with mixed shapes
- `packages/runner`: 1073 tests pass, including the new wire-shape test proving pending reads carry the doc's true confirmed basis
- `deno task check-docs`: 521 blocks pass; `deno fmt --check` clean on all touched files
- `PendingStacks_Repaired.cfg` (fullstack deps + confirmed basis + own-session exclusion) is exactly this shape — re-ran TLC locally: `PendingStacks_Current.cfg` still violates (legacy witness), Repaired holds

## Notes for review

- CT-1910's full target design (server-inferred dependencies, `verdictsThrough` watermark, rejected-commit retention) remains the follow-on protocol step; this PR closes the soundness hole without spending a protocol bump. Design notes from the scoping discussion are recorded on the ticket.
- The scripted server model in `memory-v2-stacked-commit.test.ts` deliberately keeps the LEGACY basis (noted in its mirror comment); the true-basis path is exercised against the real engine in `packages/memory`.
- Exclusion is DB-side (duplicated statements rather than a parameterized join) so the hot legacy confirmed-read scan pays nothing, the Tier-1 probe stays a single `LIMIT 1` lookup, and the exclusion semantics stay decoupled from client declarations — the coupling the inference follow-on would otherwise invalidate.

---
## Summary by cubic
Fixes pending-read staleness by scanning from a read’s declared confirmed basis (`basisSeq`) with predecessor-only own-session exclusion (localSeq below the reader), preventing missed foreign writes (CT-1910) without a protocol bump. Legacy clients keep the max-dependency basis.

- **Bug Fixes**
  - Engine (`packages/memory`): When `basisSeq` is present, scan `(basisSeq, head]` and exclude only true-predecessor own commits; validate `basisSeq` is an integer in `[0, head]`. Legacy reads (no `basisSeq`) keep the max-dep basis. Same rule in scheduler‑observation reads.
  - Wire: Add optional `basisSeq` to pending reads; older servers ignore unknown fields, so no capability flag.
  - Runner (`packages/runner`): Pending branch sends `basisSeq: confirmedSeq ?? record?.confirmed.seq ?? 0` and preserves it through old-server downgrade. Pending-read compaction is keyed by `basisSeq` so reads with divergent bases never merge; compaction test sorting made type-safe.
  - Tests/Docs: Engine regression covers out-of-order own writes (not excluded). Differential harness adds a pending-read INV‑1 re‑scan. Scheduler‑observation true‑basis path is covered (drop on foreign write; keep with predecessor exclusion). Specs and invariants updated; TLA README clarifies certification scope.

- **Migration**
  - No action needed. `basisSeq` is optional and backward compatible; old servers ignore it, old clients keep legacy basis.

<sup>Written for commit 520f123190382c8352b0dbda0792f366a135c74b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5110?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ubik2 (commented):
_cf-review via OpenAI Codex GPT-5, on behalf of @ubik2._

Verdict: **request changes**. The true-basis repair is directionally right, and the new targeted tests pass, but the runner/engine boundary still has two reachable INV-1 holes.

### [P1] Preserve `basisSeq` when compacting pending reads

`packages/runner/src/storage/v2.ts:649-690` sorts and groups pending reads by scope, id, `localSeq`, recursion, and path, but omits `basisSeq`. `buildReads` assigns the basis per read (`:3463-3472`), so two otherwise-identical read activities taken against different confirmed bases are merged and one basis is silently discarded.

I reproduced this at the PR head with identical pending reads at bases 0 and 1: compaction emitted one read at basis 1. A foreign overlapping write in `(0, 1]` is then outside the server scan, so the commit can be accepted with an observation older than its declared wire basis. Please either include `basisSeq` in the compaction identity or conservatively retain the minimum basis, and add a regression with two same-key/path activities at different bases.

### [P1] Own-session exclusion is unsound without actual FIFO submission

The new engine scans exclude every write with the reader's session (`packages/memory/v2/engine.ts:695-725`). That is only sound under the stated premise that every earlier accepted same-session write has a lower `localSeq` and was in the reader's materialized view.

The runner does not enforce that premise. In `conflictAdmissionMode="hold"`, an earlier read-bearing commit can wait at `packages/runner/src/storage/v2.ts:3131-3180` while a later independent blind commit reaches `session.transact` at `:3274-3278`. I reproduced same-session admission order `[1, 3, 2]`. Against the exact PR head, the engine then accepts this sequence:

1. localSeq 1 writes `A.x = 1`;
2. localSeq 10 overtakes and writes `A.x = 2`;
3. localSeq 5 declares a pending read based at 0 through localSeq 1, claims it observed `x = 1`, and writes a derived value.

The localSeq 10 write is excluded solely because it shares the session, so localSeq 5 is accepted even though its observation is incoherent with durable history. Please either enforce per-session FIFO across every hold/send path, or exclude only true predecessors (same-session rows with `local_seq < reader.localSeq`) and treat overtaking same-session rows as conflicts. Add a runner-to-real-engine regression for the hold case.

### Why the TLA+ check did not catch either failure

I independently traced the checked state space. TLC did catch the original `maxdep` bug, but the repaired config verifies the admission rule only after assuming away both integration steps above:

- `PendingStacks.tla:91-110` constructs one canonical `ReadRecord` per path with `cbasis = csn[s]`. There is no state or action for the runner's raw read activities or `compactCommitReads`, so basis loss cannot occur in the model.
- `PendingStacks.tla:121-139,160-164` processes `Min(Unresolved(s))` only. A later same-session commit cannot overtake an earlier held commit in any reachable state. The README explicitly says FIFO holds “by construction” and that holds are not implemented (`docs/specs/memory-v2/tla/README.md:95-99`), but the runner already has the hold path above.
- The repaired invalidation rule itself excludes all same-session log entries (`PendingStacks.tla:128-136`). That is coherent inside the structurally-FIFO model, but it does not establish that the runtime refines that model.
- The differential harness also bypasses the runner: it generates one pending read with `basisSeq = session.integratedSeq` and calls `applyCommit` immediately in generated localSeq order (`packages/memory/test/v2-differential-consistency-test.ts:221-255`). Its post-hoc oracle deliberately skips every same-session write (`:305-315`). The reference and implementation therefore share the same FIFO/exclusion assumption, and the harness has no compaction boundary to exercise.

This is a missing refinement layer, not a TLC exploration failure. Please model issued versus admitted commits separately (allow overtaking and make FIFO an invariant), add coverage for raw-read-to-wire compaction preserving the earliest basis, and narrow the README's certification claim to the admission core under canonical-read and FIFO assumptions until those refinements are checked.

### Documentation coherence

The live invariant catalog still says the top pending dependency is the mandatory staleness basis and “MUST NOT ... lower” it (`docs/specs/memory-v2/09-invariants.md:116-145`), which directly contradicts this PR's true confirmed basis. The same stale rule remains in `03-commit-model.md` and API/engine comments. Please distinguish dependency completeness/top-of-stack cascade anchoring from staleness-basis selection, and update the live docs and model terminology in this change.

Validation performed at `7f512f7d79648cca56ac9feeabbfff49ca8dd644`: the new memory tests, the runner basis test, memory type-check, docs checks, formatting, and lint all pass. The blockers above are additional counterexamples outside those test/model state spaces.

ubik2 (commented):

Self-review of the CT-1910 basis repair, done adversarially before requesting human review. Scope: 3 commits, 11 files (+733/−65) · coverage: deep-read of the full diff · checks run: memory 449/449, runner 1074/1074, differential harness (100 seeded schedules, mixed legacy/repaired shapes), `check-docs` 521 blocks, `deno fmt` / `deno lint` clean, TLC `PendingStacks_Current` (violates — expected witness) and `PendingStacks_Repaired` (holds) re-run locally.

## 🔴 Blocking

None found.

## 🟡 Improvements — both applied in 24e7216db

1. **Compaction could merge pending reads with divergent `basisSeq`** (`packages/runner/src/storage/v2.ts`, `compactCommitReads`). The pending dependency key covered `id + scope + localSeq` but not `basisSeq` (confirmed groups do key by `seq`). Two same-doc reads with different per-read basis overrides would collapse into one group, and ancestor-path compaction could drop a descendant read while the surviving ancestor claimed a higher basis for the covered subtree — the unsound direction. Unreachable today (no production writer of `read.meta.seq` exists), but a trap for whoever first sets one on a stacked doc. **Fixed:** `basisSeq` is now part of the pending dependency key.
2. **Exclusion soundness premise now cites its normative source.** The own-session exclusion leans on same-session commits resolving in increasing localSeq order — enforced by the transport per connection, not by the engine. The exclusion-scan comment now cites `03-commit-model.md §3.6.3` so the premise is traceable. (A client violating the ordering only corrupts its own data — the documented trust model.)

## ⚪ Nits (left as-is)

- `v2-pending-read-basis-overadvance.test.ts` now holds both the legacy witness and the repair suite; the name still describes the subject, so kept.
- `pendingReadBasisSeq` issues the head query once per basis-bearing read rather than once per commit — a `MAX` over an `INTEGER PRIMARY KEY`, O(1) in SQLite; hoist only if it ever profiles.

## Test rigor

Every new test states the principle it guards. The legacy unsound-accept pair is explicitly the regression witness (the role `PendingStacks_Current.cfg` plays in the model), forming a minimal pair with the repaired-shape test on the identical schedule — only `basisSeq: 0` differs. The differential harness's new post-hoc pending-read INV-1 re-scan is a genuinely new invariant check, possible for the first time because the basis is durable in the stored `original`. The naive-model flip preserves the INV-2 refinement direction for both shapes.

## Coherence

Normative spec (03-commit-model §3.4/§3.5/§3.6.3), INV-1 known-deviations, the TLA README results table, and the EXPERIMENTAL_OPTIONS `pendingReadStacks` entry all moved with the semantics; no other live doc describes the old basis rule (checked). The scheduler-observation read path received the same fix, per the #4606 review-thread requirement. No `docs/history/` content edits.

## Possible follow-ups (not filed)

- Extend the state-inspector oracle (`staleReadAnomalies`) to replay `basisSeq`-bearing pending reads — closes its pending-read blind spot, newly possible; already noted on CT-1910.
- 04-protocol's hello-flag lists were stale before this PR (omit `commitPreconditions`, `sqliteCommitRowLabelEval`, `pendingReadStacks`).

**Verdict:** sound — no blockers; the two hardening improvements from this review are already applied on the branch.

ubik2:
Follow-up on the FIFO/`hold` finding: `conflictAdmissionMode="hold"` is off by default, has no production call site in this repository, and is documented as performance-neutral with a warning not to enable it without remeasurement. Rather than expanding the admission protocol and TLA+ model to preserve correctness for an unused experimental path, I think this PR should remove the `hold` option (and its tests/docs) so the runtime cannot violate the FIFO premise through that path. The separate `basisSeq` compaction issue remains applicable regardless of this option.

ubik2:
_cf-review response via Claude (Fable 5), on behalf of @ubik2 — addressing the cubic P1 and the Codex request-changes review. Fixes in e225bea58 (the earlier compaction-key fix in 24e7216db)._

**[P1] Own-session exclusion unsound under out-of-order submission (cubic + Codex)** — confirmed and fixed as both reviews proposed: the exclusion is now restricted to **true predecessors** — own-session rows with `local_seq` below the reader's — so an own write with a higher localSeq admitted first conflicts exactly like a foreign write. The scan no longer depends on wire-order discipline at all (the engine checks `local_seq` directly instead of assuming §3.6.3 ordering holds on the wire). Applied symmetrically in the naive reference model and the differential harness's post-hoc INV-1 oracle. Regression added at the engine boundary: `session:1` admits localSeq 1, then localSeq 10 overtaking with a conflicting blind write, then the localSeq 5 reader at `basisSeq: 0` — now `ConflictError`, previously the incoherent accept you reproduced. On "enforce per-session FIFO across every hold/send path": deliberately not taken here — the predecessor rule makes admission sound regardless of send order, converting an overtake into a retry; whether hold-mode's reordering itself violates the §3.9 submission-order requirement is a pre-existing runner question, best handled separately from this soundness fix.

**[P1] `basisSeq` lost in pending-read compaction (Codex; cubic-adjacent)** — this was found and fixed in self-review before your review ran (24e7216db; your validation commit `7f512f7d7` predates it): the pending dependency key now includes `basisSeq`, so divergent bases never merge. The requested regression is added: two same-doc/path activities, one at the confirmed basis and one pinned at 0 via `meta.seq`, must both survive to the wire (`memory-v2-stacked-commit.test.ts`, "divergent basis overrides survive pending-read compaction").

**TLA+ scope** — agreed on the refinement gap, and the README now says so explicitly: a new "Scope of the certification" section states the two structural assumptions (canonical per-path reads — compaction is outside the model, guarded by the runner test; FIFO by construction — overtaking is unreachable in the model, which is why the shipped exclusion is predecessor-restricted, coinciding with the model's rule in every model-reachable state). Modeling issued-vs-admitted with FIFO as a checked invariant is named there as the refinement to build if this area churns — proposed as a follow-up rather than grown into this PR.

**Documentation coherence** — the sweep is done: INV-3 is retitled ("dependency completeness and staleness-basis selection") and now distinguishes cascade anchoring (top-of-stack REQUIRED for its resolution edge) from basis selection (declared basis + predecessor exclusion vs. legacy top-of-stack); its soundness direction spells out both shapes. §3.5's basis sentence, §3.6.3's pseudocode comment and true-basis bullet, and both `PendingRead` type comments (memory + runner) no longer state the max-basis rule unconditionally.

**Also found while addressing this** — the basis test file was named `*.test.ts` but the memory package's test include is `test/*-test.ts`, so **the suite was never running it** (the constant 449 count). Renamed; the suite now runs 456 with all seven basis tests included.

Verification at f6409256f: memory 456/456, runner 1075/1075, `check-docs` 521 blocks, fmt/lint clean.

ubik2 (commented):
_cf-review via OpenAI Codex GPT-5, on behalf of @ubik2._

Re-review at `f6409256f49425a23958f362339a903e38d17a69` · deep-read of all 11 changed files.

The earlier implementation blockers are fixed: divergent `basisSeq` reads no longer compact together, and own-session exclusion is restricted to true predecessors so higher-localSeq overtaking writes conflict. I independently reran both runner regressions, all seven basis tests, and the differential harness; they pass.

### 🔴 Blocking

1. **Fix the runner test’s type error** (`packages/runner/test/memory-v2-stacked-commit.test.ts:2045`). `basisSeq` is optional in the public type, so `left.basisSeq - right.basisSeq` produces two `TS18048` errors. Runner tests pass only because that suite uses `--no-check`; the repository `Check` job is currently red and Coverage is skipped. Narrow/assert the values or handle `undefined` explicitly before sorting.

2. **Describe predecessor-only exclusion in INV-1** (`docs/specs/memory-v2/09-invariants.md:80-81`). The catalog still says the repair excludes all accepted writes from the reader’s session. The implementation now correctly excludes only true predecessors (`localSeq < reader.localSeq`); higher-localSeq own writes must conflict. This exact distinction closed the prior P1, so the normative invariant must state it.

### 🟡 Improvement

`packages/runner/src/storage/v2.ts:3459-3461` still says pending-read staleness is always based at the highest dependency. That is now only the legacy `basisSeq`-absent rule; please update the comment while fixing the invariant wording.

Validation: memory type-check passes; 7/7 basis tests, differential harness, and both runner wire/compaction regressions pass; docs checks pass (521 blocks); fmt/lint pass. Root CI fails only at the new runner test type error in this review scope.

**Verdict:** implementation repair looks sound; request changes for the type-check failure and live-spec mismatch.

ubik2:
_cf-review response via Claude (Fable 5), on behalf of @ubik2 — re-review blockers fixed in 78932b567._

**🔴 1 — runner test type error**: fixed. The compaction regression now maps `basisSeq: read.basisSeq ?? -1` before sorting (absence would itself be a failure, surfaced as -1), so the sort operates on `number`. Verified with the actual gate this time — `deno task check` (whole-package program, 405 paths) passes at both the runner package and the root; the earlier miss was exactly the `--no-check` suite blind spot you identified.

**🔴 2 — INV-1 predecessor wording**: fixed. The known-deviations entry now states the repaired scan excludes *only the session's true predecessor commits (localSeq below the reader's; an own write admitted out of submission order conflicts like a foreign one)* — matching §3.6.3 and the implementation.

**🟡 — `pushCommitRead` layer comment**: updated in the same commit — it now distinguishes the `basisSeq` scan (predecessor-only exclusion) from the legacy highest-element basis, and notes the false-conflict hazard applies to a lower basis *without* the exclusion.

**On cubic's P1 (compaction test expecting pending reads from "confirmed" layers)** — the concern doesn't apply, for two reasons: `harness.replica` is the production replica (only the transport is scripted), so `buildReads` is the real emission path, not a double; and c2 is *not* confirmed at `buildReads` time — the call is synchronous after `beginSet(c2)` with no intervening `await`, so c2's verdict microtask cannot have run and the doc deterministically holds one pending layer (`localSeq: c2.localSeq`). The `await assertResultOk(c2.promise)` at the end confirms the verdict settles after the assertion. The test passing under the suite (which would emit zero pending reads and fail loudly otherwise) is the runtime witness.

Validation at 78932b567: `deno task check` green (root + runner), memory 456/456, runner 1075/1075, `check-docs` 521 blocks, fmt clean.

PR #5111, opened 2026-07-28
Constrain invocation payloads to Fabric values

danfuzz:
## Summary
- add NonNullableFabricValue, the non-nullish subset of FabricValue
- require invocation and command payloads to be non-nullable Fabric values
- represent embedded invocation nonces as immutable FabricBytes

Method outputs remain non-nullish unknown values because in-process Result envelopes can carry raw Error instances.

## Validation
- deno task check

## Coverage Check

This PR only edits type declarations, so the following accounts for spurious complaints:

ACCEPT_COVERAGE_DEBT: coverage-debt: packages/memory uncovered lines = 1295 lines

---
## Summary by cubic
Constrain invocation and command payloads to non-nullable Fabric values and switch nonces to `FabricBytes` for stronger type safety. Introduce `NonNullableFabricValue` (using `NonNullable<FabricValue>`) in `@commonfabric/api` and re-export it from `@commonfabric/data-model`; use it for `Command`, `Invocation`, and method input generics, while outputs remain non-nullish `unknown`.

- **Migration**
  - Ensure all `args` are non-nullish `FabricValue` and update method input generics to `NonNullableFabricValue`.
  - Replace `nonce?: Uint8Array` with `nonce?: FabricBytes` across invocations, commands, and method command shapes.

<sup>Written for commit 203ea8f41883637eb18c815b35424952fdaa3b15. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5111?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -279,6 +279,9 @@ export type FabricValue =
   | FabricPlainObject
   | undefined;
 
+/** A fabric value other than `null` or `undefined`. */
+export type NonNullableFabricValue = Exclude<FabricValue, null | undefined>;
+
 /** An array of fabric values. */
 export interface FabricArray extends ArrayLike<FabricValue> {}
 
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -280,7 +280,7 @@ export type FabricValue =
   | undefined;
 
 /** A fabric value other than `null` or `undefined`. */
-export type NonNullableFabricValue = Exclude<FabricValue, null | undefined>;
+export type NonNullableFabricValue = NonNullable<FabricValue>;
 
 /** An array of fabric values. */
 export interface FabricArray extends ArrayLike<FabricValue> {}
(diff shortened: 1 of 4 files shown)


packages/api/index.ts
```diff
@@ -279,6 +279,9 @@ export type FabricValue =
   | FabricPlainObject
   | undefined;
 
+/** A fabric value other than `null` or `undefined`. */
+export type NonNullableFabricValue = Exclude<FabricValue, null | undefined>;
```
Hixie: this should be equivalent to `NonNullable<FabricValue>` right?

danfuzz: Thanks for the catch! It was `NonNullable<...>` and then… became _not_ that… while I wasn't looking.

PR #5114, opened 2026-07-28
test(cli): four retry scenarios for caller-supplied invocation ids (verb contract WS-D/D3)

mpsalisbury:
## Why

The duplicate-on-retry bug is what started this arc: three times in one evening on the live topics board, a mutation timed out with unknowable commit state and no safe retry. D1 (#5087, merged) built the runner machinery and D2 (#5089) the CLI surface. This is the evidence that the guarantee actually holds against a real server rather than against test doubles — the plan lists it as the Phase 2 exit criterion.

## What Changed

`run_piece_call_retry` in `packages/cli/integration/integration.sh`, running in the existing `piece-call` shard (no new CI job — the CI policy discourages adding matrix entries proactively, and `setup_space` already isolates each scenario in its own space).

Four scenarios, each ending on the **same** assertion — exactly one message recorded — because that single property is what an agent depends on when it retries a call it could not observe:

1. **A failure before dispatch leaves nothing behind.** An unreachable API provably cannot dispatch, so no id was ever announced; the fresh-id retry that forces records exactly one.
2. **Killed after dispatch, before acknowledgement** — the riskiest window, where the caller cannot know whether the commit landed. The kill is triggered by the CLI's own `invocation:` announcement, read through a blocking pipe read, so it lands in the window without racing a clock or sleeping. Committed or not, the same-id retry leaves exactly one.
3. **Commit succeeded but the response was lost.** The retry settles as success — exit 0, `deduplicated: true` — rather than surfacing the receipt collision as an error.
4. **A settled id reused with a different payload does not overwrite.** The id identifies the invocation, so the second call reports the first one's outcome and `lastMessage` still holds `original-payload`.

Every scenario spawns a fresh `cf` process, so the plan's "retry from a fresh process" case is the default rather than a special one.

Support: `message_count` treats a piece with nothing materialized as zero messages rather than a read failure (matching `read_piece_value_or_default`); `new_invocation_id` falls back to Python where `uuidgen` is absent; `kill_process_tree` takes the children first, because `cf` is a shell function and `$!` is the subshell rather than the Deno process that would otherwise still commit.

## Test plan

Verified end to end against a local toolshed on port 8099 with `CF_CLI_INTEGRATION_USE_LOCAL=1`. The server log shows the mechanism doing exactly what it claims:

```
event-lost-race Event handling lost the receipt race
  eventId: "evt:caller:CoCnpEWV0GJj_iYehH7E98FF8oNeVEWAcSdLKpW5KNU"
entity-absent precondition target already exists ... precondition: "receipt-exists"
```

That is the hash-scoped caller id from D1's collision fix, with the duplicate losing the receipt race and the original outcome standing. The full `piece-call` section — existing suite plus the new one — passes in the same shard.

Stacked on #5089; the diff will reduce to the two files once that merges.

---
## Summary by cubic
Add four end-to-end retry scenarios for caller-supplied invocation IDs to prove idempotent piece calls and prevent duplicate mutations. Runs in the existing `piece-call` shard and asserts exactly one message is recorded in all cases.

- **New Features**
  - Pre-dispatch failure: fresh-id retry records one message.
  - Killed after dispatch, before ack: same-id retry leaves one message.
  - Commit succeeded, response lost: retry exits 0 and reports `deduplicated: true`.
  - Reuse settled id with different payload: original result returned; `lastMessage` unchanged.

- **Refactors**
  - Helpers in `packages/cli/integration/integration.sh`: `new_invocation_id` (Python fallback), `kill_process_tree` (kill children first), `message_count`/`assert_message_count` (treat missing materialization as zero).
  - Updated `docs/plans/pattern-verb-contract-implementation.md` to mark WS-D/D3 scenarios as done and note that scenario 3 currently verifies deduplication (exit 0, `deduplicated: true`) but not reading a result back, which will land when WS-C adds return values or via `plainResultReceipts`.

<sup>Written for commit fbc0699e9946f4317b1462705967a38174803b9e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5114?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/plans/pattern-verb-contract-implementation.md
+++ b/docs/plans/pattern-verb-contract-implementation.md
@@ -233,7 +233,11 @@ joins WS-C. `packages/runner` (`cell.ts` send path), `packages/cli`.
   without racing a clock; a `--message` that differs on the retry proves the
   settled outcome stands rather than being overwritten. Each scenario spawns
   a fresh `cf` process, so the fresh-process case is the default rather than
-  a special one.
+  a special one. One half of the third scenario is not covered yet: the
+  collision is asserted through `deduplicated` and exit 0, but not by reading
+  a *result* back off the receipt, because a void verb leaves none to read.
+  That assertion joins when WS-C gives verbs return values — or sooner
+  against `plainResultReceipts`.
 - **Exit (Phase 2, before WS-C):** the duplicate-on-retry bug is dead on the
   live board. **Exit (Phase 4, with WS-C):** the retry returns the original
   result.
--- a/docs/plans/pattern-verb-contract-implementation.md
+++ b/docs/plans/pattern-verb-contract-implementation.md
@@ -217,13 +217,23 @@ joins WS-C. `packages/runner` (`cell.ts` send path), `packages/cli`.
   fails if the path ever awaits `runtime.idle()` or `manager.synced()`. The
   live acceptance check — a deliberately slow derived recomputation cannot
   delay `addTopic` acknowledgement — rides with D3's integration scenarios.
-- Integration tests (isolated toolshed, `isolated-test-processes`
+- ~~Integration tests (isolated toolshed, `isolated-test-processes`
   conventions), four timeout/retry scenarios: timeout before dispatch (retry
   re-executes; one topic); timeout after dispatch, before commit
   acknowledgement; commit succeeded but the response was lost (retry
   collides, reads the original back, exits 0); and a retry from a fresh
   process with the same id — in every case exactly one topic exists
-  afterwards.
+  afterwards.~~ — **done (D3)**: `run_piece_call_retry` in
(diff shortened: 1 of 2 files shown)


mpsalisbury:
Superseded by #5118 — same branch and commits, retargeted to `main`. GitHub auto-closed this one when its base branch `cli/ws-d2-invocation` was deleted on the merge of #5089, and a closed PR whose base branch is gone cannot be reopened or retargeted.

PR #5117, opened 2026-07-28
feat(memory): drop connections on hello protocol mismatch

ubik2:
## Summary

- version the memory wire protocol as `memory/v2.1`
- drop hello connections without a response when the protocol identifier does not match
- keep capability-flag incompatibilities on the existing typed-error path
- make protocol mismatch permanent on the client and close the transport
- document the exact compatibility boundary and cover server, client, standalone, and Toolshed behavior

## Why

Some memory protocol changes cannot be made backward compatible through capability flags. Encoding the version in the existing `MEMORY_PROTOCOL` identifier gives those changes a single exact compatibility boundary and also binds the version into signed `session.open` authorization.

Previously, an unsupported protocol produced a typed handshake rejection while leaving the WebSocket open. During reconnect this could repeatedly reject on the same open connection. The new behavior immediately drops incompatible transports.

## Validation

- `deno task check` (`packages/memory`)
- focused memory suite: 91 passed
- Toolshed memory WebSocket route suite: 21 passed
- runner capability-skew regression: passed
- relevant lint and targeted typechecks
- `deno task check-docs`: 521 code blocks passed

---
## Summary by cubic
Version the memory wire protocol to `memory/v2.1` and drop `hello` on protocol mismatch to enforce a strict compatibility boundary and stop mixed-version reconnect loops. Docs now explain how compatibility can evolve and how the client interprets both legacy and new signals.

- **New Features**
  - `MEMORY_PROTOCOL` is now `memory/v2.1`; `hello` and `hello.ok` must match exactly.
  - On version mismatch, recipients drop: servers close without a response; `toolshed` and standalone WebSocket routes close with code 1002; clients treat mismatched `hello.ok` and legacy `UnsupportedProtocol` as permanent, close the transport, and stop reconnecting. Capability-flag mismatches stay typed (`ProtocolError`) and are permanent.
  - Runner `WebSocketTransport` maps 1002 to a permanent `ProtocolError`; loopback in `@commonfabric/memory/v2` surfaces a permanent protocol mismatch; reconnect paths updated to remember the failure.
  - Docs expanded: exact identifier boundary, accepted-identifier set (no implicit downgrade), and a clear split between `hello` negotiation and `AuthorizationError.retriable`. Tests cover client/server, Toolshed, runner, loopback, and mixed-version reconnect.

- **Migration**
  - Ensure both sides send and expect `memory/v2.1` in `hello` and `hello.ok`.
  - Update any code that assumed `"memory"` to use `MEMORY_PROTOCOL`.
  - Adjust tests/integrations to expect WebSocket close code 1002 on protocol mismatch.

<sup>Written for commit e3dfdc150967cb3836e73a0e4b4888ba7eef0e2d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5117?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/memory/v2/client.ts
+++ b/packages/memory/v2/client.ts
@@ -354,7 +354,7 @@ export class Client {
         if (message.error) {
           const error = new Error(message.error.message);
           error.name = message.error.name;
-          if (error.name === "UnsupportedProtocol") {
+          if (isPermanentHandshakeError(error)) {
             Object.assign(error, { permanent: true });
             this.#fatalError = error;
           }
@@ -1618,6 +1618,13 @@ const isPermanentAuthorizationError = (error: unknown): boolean =>
 // that would take down sessions for other spaces.
 const isPermanentConnectionFailure = (error: Error): boolean =>
   (error as { permanent?: unknown }).permanent === true;
+
+// Every typed protocol rejection produced while `hello` is pending describes a
+// peer incompatibility that reconnecting to the same endpoint cannot heal.
+// `UnsupportedProtocol` is the legacy pre-v2.1 version signal; `ProtocolError`
+// is the current capability-flag rejection.
+const isPermanentHandshakeError = (error: Error): boolean =>
+  error.name === "UnsupportedProtocol" || error.name === "ProtocolError";
 
 const requireSessionOpenAuthMetadata = (
   value: unknown,
--- a/packages/memory/v2/client.ts
+++ b/packages/memory/v2/client.ts
@@ -119,11 +119,12 @@ export class Client {
   #connected = false;
   #closed = false;
   // Set when a reconnect handshake fails for a reason retrying cannot change (a
-  // protocol-flag mismatch — the transport is fundamentally incompatible). The
-  // client stops reconnecting and fails every further request with it, instead
-  // of looping forever. A per-session authorization denial does NOT land here:
-  // it terminates only that session (see SpaceSession.restore), leaving sessions
-  // for other spaces on this client alive.
+  // protocol-version or flag mismatch — the transport is fundamentally
+  // incompatible). The client stops reconnecting and fails every further
+  // request with it, instead of looping forever. A per-session authorization
+  // denial does NOT land here: it terminates only that session (see
+  // SpaceSession.restore), leaving sessions for other spaces on this client
+  // alive.
   #fatalError: Error | null = null;
 
   private constructor(
@@ -299,7 +300,16 @@ export class Client {
 
     if (this.#helloPending !== null) {
       const helloOk = parseHelloOk(message);
-      if (helloOk !== null) {
+      if (helloOk?.kind === "incompatible") {
+        const error = permanentProtocolError(
+          "memory protocol version mismatch",
+        );
+        this.#fatalError = error;
+        this.#helloPending.reject(error);
+        void this.transport.close().catch(() => undefined);
+        return;
+      }
+      if (helloOk?.kind === "ok") {
         const expectedFlags = getMemoryProtocolFlags();
         if (!compatibleMemoryProtocolFlags(helloOk.flags, expectedFlags)) {
           // A data-model wire-contract mismatch: this client and server cannot
@@ -414,6 +424,9 @@ export class Client {
       session.handleDisconnect();
     }
     this.rejectPending(toConnectionError(error));
+    if (this.#fatalError) {
+      return;
+    }
     void this.reconnect().catch(() => undefined);
   }
 
@@ -429,7 +442,7 @@ export class Client {
     }
     this.#reconnecting = (async () => {
       let attempt = 0;
-      while (!this.#closed) {
+      while (!this.#closed && !this.#fatalError) {
         try {
           await this.hello();
           for (const session of this.#spaces) {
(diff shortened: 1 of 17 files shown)


ubik2 (commented):
cf-review via GPT-5, on behalf of @ubik2

**Verdict: blocking; address before merge.**

The focused implementation is coherent and the memory client/server tests, lint, formatting, and touched-file typechecks passed. One blocking mixed-version reconnect issue remains: the client only makes a mismatched `hello.ok` permanent, while real peers signal incompatibility by closing with code 1002 or—before this change—returning `UnsupportedProtocol`. Please cover both real signaling paths before merge.

Also update `docs/development/authorization-failure-surfacing.md` and the linked protocol resume paragraph to distinguish exact version mismatches from capability-flag mismatches.

packages/memory/v2/client.ts
```diff
@@ -299,7 +300,16 @@ export class Client {
 
     if (this.#helloPending !== null) {
       const helloOk = parseHelloOk(message);
-      if (helloOk !== null) {
+      if (helloOk?.kind === "incompatible") {
```
ubik2, line 310: 🔴 **Blocking (major): real version mismatches do not reach this branch.** A conforming newer server closes without `hello.ok` using code 1002, but `WebSocketTransport` discards the `CloseEvent` code and `onClose()` treats it as a retryable `ConnectionError`. A pre-change server instead returns the typed `UnsupportedProtocol` handshake response, which the branch below neither marks permanent nor closes. I reproduced both reconnect paths reaching three hellos with zero client-initiated closes. The new client test supplies a mismatched `hello.ok`, but the server in this PR cannot emit one, so it misses the deployment cases. Please surface/classify protocol close 1002 as permanent, recognize the legacy `UnsupportedProtocol` response, close the transport, and test both paths.

packages/memory/v2/client.ts
```diff
@@ -337,7 +354,14 @@ export class Client {
         if (message.error) {
           const error = new Error(message.error.message);
           error.name = message.error.name;
+          if (error.name === "UnsupportedProtocol") {
```
ubik2: 🔴 **Blocking (major): the real capability-flag response is still retryable.** `respondToHello()` returns a typed `ProtocolError` for incompatible flags, but this branch only marks `UnsupportedProtocol` permanent. I reproduced the real response reaching three consecutive hellos with zero closes. `FlagMismatchOnReconnectTransport` instead sends an incompatible `hello.ok`, which the server cannot emit. Please classify the server's actual flag-mismatch response as permanent (preferably with a distinct typed error) and make the reconnect test use that response.

packages/toolshed/routes/storage/memory/memory.handlers.ts
```diff
@@ -199,14 +199,24 @@ const attachMemorySocketPipeline = (
...
+      }
       logMemWrites(firstMessage);
       negotiation.handoff({
         onMessage(message) {
           // Trace only after the receive resolves, so a message whose receive
           // fails (the fatal-error path below) is not logged as a write.
           void connection.receive(message).then(
-            () => logMemWrites(message),
+            (nextDisposition) => {
+              if (nextDisposition === "closed") {
```
ubik2: 🔴 **CI-blocking: this post-handoff branch does not represent a protocol mismatch.** `Connection.close()` is entered internally only while processing the initial hello, already handled before `negotiation.handoff()`. A later `"closed"` disposition can only reflect external socket teardown, when `safeSocketClose()` is already a no-op; labeling it a mismatch is misleading. These lines are also the uncovered path behind the failing Toolshed coverage-debt gate. Remove this disposition check and restore the simple successful receive callback, unless later messages are intentionally changed to create a tested protocol-close transition.

ubik2 (commented):
cf-review rerun via GPT-5, on behalf of @ubik2

**Verdict: blocking; address before merge.**

The original mixed-version finding is fixed: WebSocket close 1002 and legacy `UnsupportedProtocol` are now permanent, reconnect stops, tests cover both paths, and the docs are coherent.

Two issues remain:

1. The current server's real capability-flag mismatch signal is a typed `ProtocolError`, but only `UnsupportedProtocol` responses are marked permanent. The reconnect loop therefore still repeats for a real flag mismatch; the existing test models an incompatible `hello.ok` that the server never emits.
2. The post-handoff Toolshed `closed` branch cannot represent a protocol mismatch and is producing the failing coverage-debt gate. The initial hello path already handles the only internal protocol-mismatch close.

Verification: 91 focused memory tests passed; runner WebSocket transport tests passed; memory and touched-file typechecks, lint, format, and all 521 checked documentation blocks passed.

ubik2:
Client provides a minimum protocol version, and future clients can include an `Upgrade` style field in their `hello` message, so the server can respond with a `hello.ok` matching the best shared version.

A server that doesn't support he stated protocol version will close with the websocket ProtocolError message.

Deployment of this is intentionally breaking.

ubik2:
Closing this instead. I added this to give me the ability to have a breaking protocol change, so we don't have issues like Ben's client stalling, but I think there's other ways to get there, and this would itself be a breaking change.

PR #5133, opened 2026-07-28
fix: un-brick stored patterns against toolchain drift — TS2578 non-fatal + hot-swap survives unreadable argument slots (CT-1916, CT-1917)

bfollington:
# fix: un-brick stored patterns against toolchain drift — TS2578 non-fatal + hot-swap survives unreadable argument slots (CT-1916, CT-1917)

## Why

The 2026-07-28 estuary deploy (#5094, healthy otherwise) surfaced two ways a
**stored** pattern can brick against an **evolving toolchain**, both live in
production in `ben-loom-dev-6` (log-verified, and reproduced against the real
space via `cf piece ls` / local tests):

1. **CT-1916 — stale `@ts-expect-error` bricks recompile.** Loom-mobile
   patterns stored in the space carry
   `// @ts-expect-error -- the vendored jsx.d.ts omits cf-cell-link's label`.
   The labs types bump added `label`, so the *identical stored bytes* now
   recompile with the directive **unused** → TS2578 → fatal `CompilerError` →
   "Could not load pattern …" → the space root (`cf-loom-mobile`) dies →
   `getSpaceRoot` times out → **"failed to load piece"**, retried ~1/s forever.
   There is no upgrade path out: these are user-deployed patterns whose stored
   source IS the latest version — nothing to roll forward to, and the compile
   failure is spurious (only the suppression comment went stale, not the code).

2. **CT-1917 — hot-swap dies on argument slots it cannot read.** The official
   `backlinks-index.tsx` upgrade swap re-ran setup, which re-validates the
   STORED argument by materializing it — dereferencing every link. Its
   `pieceRegistry` argument links into the loom-mobile host's registry doc;
   the host was down (see 1), the link read cold, the key vanished from the
   materialized value, and the swap died with
   `updated arguments do not match the candidate schema: missing required
   property pieceRegistry`. The stored argument was exactly what the original
   instantiation wrote — "unreadable right now" was treated as "invalid".

The two compound: 1916 killed the host, which made 1917's link read cold.
Together they are the third instance in a week of one systemic mechanism —
durable stored patterns/docs drifting against a moving toolchain with a
fail-closed upgrade path (see CT-1918 for the deploy-gate arc; prior
instances: #4967, #4977, #5094).

## What Changed

**js-compiler (CT-1916):** TS2578 ("Unused '@ts-expect-error' directive")
is non-fatal **only when recompiling durable STORED pattern source** — a new
`storedSource` compile option set by exactly one caller,
`compileResolvedToRecordGraph` (the reload of stored source docs by
identity, the only place bytes nobody can re-author meet a toolchain newer
than the one that accepted them). Authoring surfaces — `cf check`, deploy,
the corpus check — stay fully strict: there the author is present and
removing the stale directive is the right fix. `noEmitOnError` stays on for
authoring and is disabled only in stored mode, where the pipeline's explicit
syntactic/program/semantic gates decide fatality. Rationale: stored sources
are recompiled by every future toolchain while their type environment is
supplied by the platform; a suppression obsoleted by a platform type
improvement must not retroactively brick the stored artifact — but should
absolutely be reported to a present author.

**runner (CT-1917):** swap-path argument validation (`argument === undefined`,
i.e. re-staging the STORED value on a pattern change) now:

- **skips staging/validation entirely when the argument doc reads nothing** —
  a nested piece's argument lives in its host's doc; when that is cold there
  is nothing to stage, and writing schema defaults over an unreadable doc
  would clobber whatever it really holds;
- **validates slots whose raw value is a link that materialized to nothing as
  opaque placeholders** — the slot has a value we cannot read right now, so
  its schema check defers to instantiation-time reactive reads, exactly like
  the running pattern's own reads of the same slot.

Caller-supplied arguments (piece API, `setup()` with a value) keep strict
validation of exactly the supplied value.

## Test plan

- `packages/runner/test/pattern-swap-link-argument.test.ts` (new): warm-link
  control (passed before and after), cold slot-link swap, and cold
  argument-doc swap (both bricked before — the exact production shades:
  "value does not match type array" / "missing required property").
- `packages/js-compiler/test/typescript.compiler.test.ts`: unused directive
  compiles; a directive covering a live error still suppresses it; genuine
  errors still throw (pre-existing cases).
- Full `packages/js-compiler` suite green; full `packages/piece` suite green
  (298 steps); full `packages/runner` suite diffed against an origin/main
  baseline run (identical pre-existing local CFC-env failures, no new ones).
- `deno task cfcheck` (pattern corpus) green with the modified compiler.
- **Against production (read-only):** `cf piece ls` on the affected estuary
  space with this branch's toolchain — every previously-bricked piece
  (`Loom Mobile`, `Loom Files`, `Loom Activity`, `Folder`, `Loom Page`)
  compiles from its current stored source and lists correctly; on the
  deployed build the same command shows five `<error: Could not load pattern
  …>` rows.

## Deferred / follow-ups

- Loom-side: remove the now-stale directive in loom's
  `src/patterns/cf-loom-mobile/files.tsx` (owner: Ben; heals via redeploy).
- CT-1918: deploy gate compiling real vintage spaces against candidate
  toolchains — would have caught both of these pre-deploy.
- The input/output additive-required rule (element 1) remains open; it is a
  DIFFERENT gap from CT-1917 (this incident was never a schema-evolution
  refusal — tracked in CT-1904).

---
## Summary by cubic
Prevents stored patterns from bricking after toolchain changes and lets hot-swaps proceed when argument links are unreadable. Narrows TS2578 tolerance to stored-source reloads only while authoring stays strict. Addresses CT-1916 and CT-1917.

- **Bug Fixes**
  - `packages/js-compiler` (CT-1916): Treat TS2578 ("Unused `@ts-expect-error`") as non-fatal only when compiling durable stored source; authoring surfaces (`cf check`, deploy, corpus) report it. Keep `noEmitOnError` on by default and disable it only in stored-source mode; add explicit fatal gates for syntactic and program diagnostics (including under `noCheck`) so malformed source never emits; the corpus-collecting path now includes options/global diagnostics. Real errors still fail and directives still only suppress covered errors.
  - `packages/runner` (CT-1917): On pattern swap with no new argument, skip staging/validation when the argument doc reads nothing; validate slots whose raw value is a link that materializes to nothing as opaque placeholders (including inside arrays and nested paths); keep strict validation for caller-supplied arguments.

- **Refactors**
  - Expanded tests: stored-source directive drift vs authoring strictness, storedSource + `noCheck` syntax gate, collecting-path program diagnostics, and link-placeholder overlay during swaps. Removed an unreachable swap-setup validation branch.

<sup>Written for commit 5bc0417aac4ea188755d6b0a60033ead77edf352. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5133?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

ACCEPT_COVERAGE_DEBT: coverage-debt: packages/runner uncovered lines = 6915 lines
(one defensive shape-mismatch guard line in the link-overlay walker; the meaningful branches are covered by the four swap tests)

### DIFF
--- a/packages/js-compiler/test/typescript.compiler.test.ts
+++ b/packages/js-compiler/test/typescript.compiler.test.ts
@@ -52,22 +52,35 @@ const TESTS: TestDef[] = [
     expectedError: "Cannot find module './foo.ts'",
   },
   {
-    // Stored sources are recompiled by every future toolchain against
-    // PLATFORM-supplied types; a directive that becomes unnecessary when
-    // those types improve must not brick the source (CT-1916 — 2026-07-28
-    // estuary, loom-mobile patterns vs. a jsx.d.ts that gained a prop).
-    name: "Unused @ts-expect-error compiles (durable-source drift)",
+    // Authoring surfaces stay strict: with the author present, the right fix
+    // for a stale directive is removing it.
+    name: "Throws: unused @ts-expect-error is fatal when authoring",
     source: [
       "const add = (x: number, y: number): number => x + y;",
       "// @ts-expect-error -- suppressed an error under an older type env",
       "export default add(1, 2);",
       "",
     ].join("\n"),
+    expectedError: "Unused '@ts-expect-error' directive.",
   },
   {
-    // The suppression is code-2578-narrow: a directive that still covers a
-    // REAL error keeps suppressing it, and errors outside any directive
-    // still fail (covered by "Throws: type check failure" above).
+    // Stored sources are recompiled by every future toolchain against
+    // PLATFORM-supplied types; a directive that becomes unnecessary when
+    // those types improve must not brick the reload (CT-1916 — 2026-07-28
+    // estuary, loom-mobile patterns vs. a jsx.d.ts that gained a prop).
+    name: "Unused @ts-expect-error compiles under storedSource",
+    source: [
+      "const add = (x: number, y: number): number => x + y;",
+      "// @ts-expect-error -- suppressed an error under an older type env",
+      "export default add(1, 2);",
+      "",
+    ].join("\n"),
+    storedSource: true,
+  },
+  {
+    // The tolerance is code-2578-narrow: a directive that still covers a
+    // REAL error keeps suppressing it in either mode, and errors outside any
+    // directive still fail (covered by "Throws: type check failure" above).
     name: "@ts-expect-error still suppresses a live error",
     source: [
       "const add = (x: number, y: number): number => x + y;",
@@ -75,6 +88,15 @@ const TESTS: TestDef[] = [
       "export default add('1', 2);",
       "",
     ].join("\n"),
+  },
+  {
+    // storedSource must not weaken real type errors — only hygiene codes.
+    name: "Throws: real type error still fatal under storedSource",
+    source:
+      "function add(x:number, y:number): number {return x+y}; export default add(`0`, 2);",
+    storedSource: true,
+    expectedError:
+      "Argument of type 'string' is not assignable to parameter of type 'number'.",
   },
 ];
 
@@ -151,10 +173,10 @@ describe("TypeScriptCompiler", () => {
     ]);
   });
 
-  it("compileToModulesCollecting reports real errors, drops non-fatal codes", async () => {
-    // The corpus-collecting path applies the same non-fatal filter as the
-    // throwing path: an unused @ts-expect-error (TS2578) is dropped, a real
-    // type error in another file is still attributed and reported.
+  it("compileToModulesCollecting reports stale directives AND real errors", async () => {
+    // The corpus check is an AUTHORING surface — no stored-source tolerance:
+    // an unused @ts-expect-error is reported (for the author to remove)
+    // alongside real type errors, each attributed to its file.
     const compiler = new TypeScriptCompiler(types);
     const resolved = await compiler.resolveProgram(
       new InMemoryProgram("/main.ts", {
@@ -169,8 +191,18 @@ describe("TypeScriptCompiler", () => {
       }),
     );
     const collected = compiler.compileToModulesCollecting(resolved);
-    expect(collected.diagnostics.map((d) => d.file)).toEqual(["/bad.ts"]);
-    expect(collected.diagnostics[0].message).toContain(
+    // The emit pass re-reports per-file diagnostics, so assert the SET of
+    // attributed files rather than exact multiplicity.
+    expect([...new Set(collected.diagnostics.map((d) => d.file))].sort())
+      .toEqual([
+        "/bad.ts",
+        "/main.ts",
+      ]);
+    const byFile = Object.fromEntries(
+      collected.diagnostics.map((d) => [d.file, d.message]),
+    );
+    expect(byFile["/main.ts"]).toContain("Unused '@ts-expect-error'");
+    expect(byFile["/bad.ts"]).toContain(
       "Type 'string' is not assignable to type 'number'.",
     );
   });
--- a/packages/js-compiler/test/typescript.compiler.test.ts
+++ b/packages/js-compiler/test/typescript.compiler.test.ts
@@ -41,6 +41,31 @@ const TESTS: TestDef[] = [
     source: "import { foo } from './foo.ts';export default foo()",
     expectedError: "Cannot find module './foo.ts'",
   },
+  {
+    // Stored sources are recompiled by every future toolchain against
+    // PLATFORM-supplied types; a directive that becomes unnecessary when
+    // those types improve must not brick the source (CT-1916 — 2026-07-28
+    // estuary, loom-mobile patterns vs. a jsx.d.ts that gained a prop).
+    name: "Unused @ts-expect-error compiles (durable-source drift)",
+    source: [
+      "const add = (x: number, y: number): number => x + y;",
+      "// @ts-expect-error -- suppressed an error under an older type env",
+      "export default add(1, 2);",
+      "",
+    ].join("\n"),
+  },
+  {
+    // The suppression is code-2578-narrow: a directive that still covers a
+    // REAL error keeps suppressing it, and errors outside any directive
+    // still fail (covered by "Throws: type check failure" above).
+    name: "@ts-expect-error still suppresses a live error",
+    source: [
+      "const add = (x: number, y: number): number => x + y;",
+      "// @ts-expect-error -- intentionally wrong argument type",
+      "export default add('1', 2);",
+      "",
+    ].join("\n"),
+  },
 ];
 
 const staticCache = new StaticCacheFS();
--- a/packages/js-compiler/test/typescript.compiler.test.ts
+++ b/packages/js-compiler/test/typescript.compiler.test.ts
@@ -47,6 +47,16 @@ const TESTS: TestDef[] = [
     expectedError: "Expression expected.",
   },
   {
+    // The most permissive mode combination: storedSource turns noEmitOnError
+    // off and noCheck skips semantics — the explicit syntactic gate is the
+    // ONLY thing standing, and it must still stand.
+    name: "Throws: syntax error still fatal under storedSource + noCheck",
+    source: "export const x = ;",
+    storedSource: true,
+    noCheck: true,
+    expectedError: "Expression expected.",
+  },
+  {
     name: "Throws: Invalid import",
     source: "import { foo } from './foo.ts';export default foo()",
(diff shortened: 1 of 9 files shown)


bfollington (commented):
**Blocking (verified) — `noEmitOnError: false` drops fatal diagnostic classes that this pipeline does not check explicitly.**

`compileToModulesSteps()` checks authored semantic diagnostics (unless `noCheck`), declaration diagnostics, and `emitResult.diagnostics`, but it never calls `getSyntacticDiagnostics()`, `getGlobalDiagnostics()`, or `getOptionsDiagnostics()`. With `noEmitOnError` off, `program.emit()` does not return those pre-emit diagnostics. On this head I reproduced `export const x = ;` producing `exports.x = ;` instead of throwing in **both** normal and `noCheck` modes. `origin/main` reports `Expression expected` and refuses emit. The collecting path has the same broader regression: with an empty type-lib environment it now reports no diagnostics and emits, where `noEmitOnError` previously surfaced the missing global/lib diagnostics.

So filtering TS2578 itself is sound, but disabling the TypeScript veto is not yet safe. Please add explicit fatal checks for the pre-emit classes that were previously covered (at minimum syntactic, global, and options diagnostics) at both load and collecting gates, while keeping semantic diagnostics conditional on `noCheck`. Regression coverage should include malformed syntax in normal + `noCheck` compilation and the collecting gate.

I found no blocker in the runner half. The unresolved-link token is applied only to the cloned validation value after the real argument write is staged, cannot be forged by a caller because acceptance is identity-based, and is never passed to storage. Only the stored-argument hot-swap call supplies `unresolvedLinkRaw`; caller-supplied arguments retain strict validation. `sameStoredSetup=true` keeps same-pattern replay and the #4977 repair call out of the changed branch. The new runner test passes all three cases and pins the warm control plus the two production cold-read failure shapes (`value does not match type array` and missing required property).

Focused verification: compiler suites passed (34 steps); `pattern-swap-link-argument.test.ts` passed (3/3); TS2578 collecting probe returned no diagnostics and emitted successfully.

loom-review via Codex `GPT-5`, on behalf of @bfollington

bfollington:
Review blocker addressed in c1c824708 — and confirmed first: `export const x = ;` did emit invalid JavaScript in both normal and `noCheck` modes on the previous commit, exactly as the review found. `program.emit()` never surfaces syntactic diagnostics, so turning `noEmitOnError` off had removed the only parse gate on the `compileToModules` path.

The pipeline now collects syntactic + options/global diagnostics explicitly and throws, on every path including `noCheck` (which skips type-checking, never parsing). The non-fatal filter does not apply there — the suppressed codes are semantic, and a file that does not parse can never be loaded. Pinned by two new cases: a parse error under default options fails at the syntactic gate ("Declaration or statement expected."), and a syntax error under `noCheck: true` still throws ("Expression expected.").

bfollington (commented):
**Blocking (verified) — the throwing gate is fixed, but the collecting path still drops options/global diagnostics and emits.**

`c1c824708` closes the original `compileToModulesSteps()` hole. The new gate runs before type/declaration checks and emit, is unconditional under `noCheck`, and covers the diagnostic classes that TypeScript 6.0.3 `handleNoEmitOptions()` used for `noEmitOnError`: options, per-source syntactic, global, and semantic. Semantic remains intentionally conditional on `noCheck`; declaration emit is disabled, and config-file parsing diagnostics are not part of `handleNoEmitOptions` (nor injectable through this compiler use of `createProgram(rootNames, options, host)`). I do not see another missing category on the throwing path.

However, the prior blocker also called out the corpus-collecting gate. `compileToModulesCollecting()` at `compiler.ts:587-612` still collects only authored semantic/syntactic diagnostics plus emit diagnostics; it never calls `getOptionsDiagnostics()` or `getGlobalDiagnostics()`. I reproduced this on the reviewed head with `new TypeScriptCompiler({})` and a valid one-line source: `compileToModulesCollecting()` returned `diagnostics: []` and emitted `/main.ts`, while `compileToModules()` correctly threw the missing-lib options diagnostics and missing-global diagnostics. Because `cfcheck` uses the collecting path, that path still fails open on an invalid compiler environment. Please add options/global collection there and pin the empty/missing type-lib case.

The `5a9148113` runner dead-branch removal does not change my prior assessment. The preceding both-undefined branch owns the only case where there is no value; otherwise the merge has a supplied or stored value, so the removed validation-only arm was unreachable. The focused runner test passes all four cases, including the new array-item link case.

Focused verification: `typescript.compiler.test.ts` passed (25/25 steps); `pattern-swap-link-argument.test.ts` passed (4/4); the collecting-path empty-type-environment probe reproduced the silent emit above.

loom-review via Codex `GPT-5`, on behalf of @bfollington

bfollington:
Re-review's remaining blocker addressed in the latest commit: `compileToModulesCollecting()` now reports options/global diagnostics through the same `pushTs` funnel, restoring parity with the throwing path's explicit gate — it can no longer return `diagnostics: []` for a program `compileToModules` refuses.

bfollington:
Codex final verification pass (posted on its behalf — its gh session was signed out):

> Approved — no findings. `d9ad66bc1` closes the last blocker by restoring options/global diagnostic parity in `compileToModulesCollecting`.
>
> Verified with a program-level diagnostic probe and the js-compiler suite: 9 files / 72 steps passed.

Full review history this PR: round 1 found one verified blocker (syntactic emit under noEmitOnError:false) → fixed in c1c824708 with an unconditional parse gate + two pinning tests; round 2 found the collecting-path parity gap → fixed in d9ad66bc1; round 3 approved with no findings.

bfollington:
Scope change in df8d2633a, per discussion: the TS2578 tolerance is **narrowed to stored-source reloads only**, behind an explicit `storedSource` compile option set by exactly one caller — `compileResolvedToRecordGraph`, the by-identity recompile of durable stored source docs (the CT-1916 `load-pattern-by-identity-source-miss` path). Every authoring surface (`cf check`, deploy/`setsrc`, the corpus check) is back to full strictness, and `noEmitOnError` is restored there; it's disabled only in stored mode where the explicit gates from the earlier review rounds decide fatality.

New contract test `stored-source-directive-drift.test.ts` pins both halves: the reload path compiles a source whose directive went stale under a newer type environment; `compilePattern` rejects the identical source. Re-verified against the affected estuary space through the CLI's load path — all 12 pieces list cleanly, zero errors.

bfollington:
Re-review of `df8d2633a`: approved — no blocking findings.

- **Scope is correctly narrow.** `storedSource: true` is set only in `Engine.compileResolvedToRecordGraph`. The only production caller of that method is `PatternManager.tryColdLoadByIdentity`, after `loadVerifiedSourceClosure`; every `PatternManager` authoring/cache/batch call continues through `compileToRecordGraph` or `compileToModulesCollecting`, and `TypeScriptHarnessProcessOptions` does not expose the stored-source switch.
- **The diagnostic gates compose correctly.** Authoring keeps `noEmitOnError: true`. Stored reloads alone override it to `false`, after which the unconditional options/global + per-source syntactic gate runs before the semantic/declaration checks and emit. `noCheck` still skips only semantic checking, not parsing/program diagnostics. The TS2578 filter is confined to stored-mode semantic and emit diagnostics; real type errors remain fatal.
- **The new contract is correctly two-sided.** `stored-source-directive-drift.test.ts` exercises the stored reload seam and the actual `PatternManager.compilePattern` authoring seam with identical source. Together with the existing load-by-identity suite, it pins the intended cold-reload route without weakening authoring.

One non-blocking coverage note: after restoring authoring's `noEmitOnError`, the existing “syntax error still fatal under noCheck” unit case no longer specifically proves the explicit syntax gate—the TypeScript emit veto can make that default-mode test pass. A parallel `storedSource: true, noCheck: true` malformed-source case would pin the only mode where the explicit gate is load-bearing. I do not see a current runtime bug; the gate is present and ordered correctly.

Focused verification on Deno 2.8.1:

- `typescript.compiler.test.ts`: 27/27 steps passed
- `stored-source-directive-drift.test.ts` + `load-by-identity.test.ts`: 14/14 steps passed
- targeted `deno check`, `deno fmt --check`, and `deno lint`: passed

loom-review via Codex `GPT-5`, on behalf of @bfollington

PR #5143, opened 2026-07-29
fix(runner): partialCause aliases must not abort a node's resume pre-sync scan

bfollington:
## Why

After the #5133 deploy (4f29cd8f4), estuary home spaces log repeated

```
resume-owned-cells skipping a sub-pattern node whose outputs did not bind or resolve
Cannot parse partialCause alias as link: {"$alias":{"partialCause":{"$generated":0},"path":[],"scope":"space"}}
```

A deferred partialCause alias in a node's stored outputs unwraps (one level)
to a bare `$alias` record; `parseAliasBinding` refuses partialCause by design
(a derived internal cell is not a link); and the resume walk's per-node catch
then skips the WHOLE node's owned-cell pre-sync. That pre-sync exists to stop
a cold-cache race the code documents precisely: unsynced owned cells read
absent, the batched instantiation commit loses to the streaming durable
value, and the resumed sub-tree's state strands. Every affected sub-pattern
silently re-enters that race on every resume.

Suspected (NOT yet confirmed) user-visible symptom: the persistently empty
Profile tab on estuary home spaces — the same session logs show no other
signature. Confirmation requires a deployed build; the fix is correct
independent of that outcome.

## What Changed

A partialCause alias can never be a node's reserved result spot, so
`firstResolvedOutputRedirect` skips the entry and keeps scanning instead of
throwing out of the walk. The derived internal cells such aliases denote are
already collected from each pattern's own `derivedInternalCells` manifest at
every level of the resume recursion — nothing is lost by not resolving them
here.

## Test plan

- New `resume-output-redirect-partialcause.test.ts`: the exact logged alias
  shape listed BEFORE the real result spot still finds the spot; outputs
  holding ONLY such aliases return undefined instead of throwing.
- Resume/setup-adjacent suites green locally (nested-piece repair, swap-link
  argument, ensure-piece-running, resume-argument-link presync).

Follow-up (separate): the pattern auto-updater compiled the SPA's index.html
fallback as TypeScript when a space's patternSource is a non-official path —
filed separately.

---
## Summary by cubic
Fixes runner resume pre-sync: `partialCause` aliases in stored outputs no longer abort a node scan. This prevents owned-cell pre-sync from being skipped and avoids the cold-cache race on resume.

- **Bug Fixes**
  - Skip `partialCause` alias entries in `firstResolvedOutputRedirect` and keep scanning; these derived cells are handled via `derivedInternalCells`.
  - Export `firstResolvedOutputRedirect` for tests and add `resume-output-redirect-partialcause.test.ts` to cover mixed and alias-only outputs.

<sup>Written for commit a96dd0cf4bae9659bb8834b23cd7f912c08ab127. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5143?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -461,14 +461,30 @@ const recordRawBuiltinResultSchemaPolicyInput = (
  * suitable as the cause for the node's result cell instead of hashing the
  * pattern object (which drags in the session-varying `program`). Returns
  * undefined if the binding contains no write redirect.
+ *
+ * Exported for tests only.
  */
-function firstResolvedOutputRedirect(
+export function firstResolvedOutputRedirect(
   runtime: Runtime,
   tx: IExtendedStorageTransaction,
   binding: unknown,
   baseCell: Cell<any>,
 ): NormalizedFullLink | undefined {
   if (isWriteRedirectLink(binding) || isAliasBinding(binding)) {
+    // A partialCause alias denotes a DERIVED INTERNAL cell — of the child
+    // level it was deferred to when it still carried `defer` — never the
+    // node's reserved result spot, and it cannot be parsed as a link
+    // (parseAliasBinding throws by design). Before 2026-07 this threw here,
+    // and the caller's catch skipped the WHOLE node's owned-cell pre-sync —
+    // silently re-exposing the cold-cache commit-revert race the pre-sync
+    // exists to prevent, for every sub-pattern whose outputs lead with such
+    // an alias (seen live on estuary home spaces as "resume-owned-cells
+    // skipping…"). Skip just this entry and keep scanning: the derived
+    // cells themselves are collected from each pattern's own
+    // derivedInternalCells manifest.
+    if (isAliasBinding(binding) && binding.$alias.partialCause !== undefined) {
+      return undefined;
+    }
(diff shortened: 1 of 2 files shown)


bfollington:
Codex review (posted on its behalf — its gh session was signed out): **no findings.**

> Verified:
> - A surviving `partialCause` alias belongs to a deeper pattern level; current-level aliases are converted to sigil redirects during binding. It cannot be the node's reserved result spot.
> - `collectResumeOwnedCells` collects each pattern's `derivedInternalCells` before walking child nodes.
> - All three production callers were checked; none relies on `parseAliasBinding` throwing.
> - The regression test uses the exact generated production shape and confirms scanning continues to the real redirect.
> - Focused and related tests passed: 154 steps, plus format, lint, frozen type-check, and diff checks.

PR #5144, opened 2026-07-29
test(patterns): gate merges on pattern update compatibility

seefeldb:
## Why

`cf piece setsrc` refuses a replacement whose schemas are not backward compatible — [`PieceController.setPattern`](packages/piece/src/ops/piece-controller.ts) calls `assertPatternSchemasBackwardCompatible`.

**The automatic updater does not.** [`PatternUpdater`](packages/runner/src/pattern-updater.ts) compiles, verifies the entry identity, and applies via `runtime.setup` or a bare `patternIdentity` write with no structural comparison at all. The spec says so outright — "the specialized system-source updater does not perform the complete pre-apply structural comparison".

Now that default/home/profile auto-update, an incompatible schema that merges is applied silently to every running piece. CI is the only gate there is, and it didn't exist.

## What

`deno task pattern-compat` compiles every authored pattern and proves its argument/result contract against every contract recorded under `packages/patterns/baselines/`, reusing the exact function `setsrc` enforces so the two paths can't drift on what "compatible" means.

```bash
deno task pattern-compat
```

## Proof it does the job

Dropped `favorites` from `HomeOutput` in `home.tsx`:

```
system/home.tsx
  cannot be applied over baseline 20260729T000537Z-mKLGw1aighDtz0A6:
Pattern schemas are not backward compatible:
- result.favorites: existing result field was removed
```

Then ran `--update` on that same broken state: it recorded the new contract **and still failed**, exit 1. `--update` can only add evidence, never clear a finding — that's what makes it safe to hand to an author.

## Design points that are load-bearing

- **A baseline holds only the two schemas.** Those four values are the entire input to the check, so no compiled artifact and no old source is needed. It can't rot the way recovering an old artifact can (the by-identity load keys its cache on the runtime version and falls back to recompiling old source).
- **Every baseline is checked, not just the newest.** A piece rolls forward from whatever version it last opened at, and the evolution-policy allowances aren't guaranteed to compose across steps.
- **Never pruned; `--update` can only add.** A command that could remove a baseline could remove the one that would have caught the break.
- **An already-recorded contract skips both the validity check and the subset proof.** It's unchanged since the run that recorded it. `--update` correspondingly refuses to record an invalid contract, which is what keeps the skip sound.

## A live bug this turned up

`lobby/main.tsx` takes **>180s** in `assertPatternSchemasBackwardCompatible` comparing its contract **to itself**, and ~55s in `validateSchemaDefinition` alone — on a 3.2 KB schema with 13 `$defs`. So it's combinatorial blowup, not size. This means **`cf piece setsrc` on `lobby/main.tsx` effectively hangs today**, independent of this PR. The skip above sidesteps it for the steady-state path (>180s → 1s), but the underlying pathology is real and being investigated separately.

## Cost

Sharded 4× like `cfcheck`: **~46s per shard** (2:54 unsharded). Retirement detection needs the whole tree, so it's filesystem-only and runs unsharded on shard 1.

## Scope notes

- 87 of 423 files under `packages/patterns` aren't pattern entries (`schemas.tsx`, `auth-types.ts`, client helpers). They have no update contract and are skipped, not failed. A file that *has* baselines and stops yielding a contract is still reported — that's a pattern someone deleted or de-patterned, with every piece tracking it pinned forever and nothing surfacing on the piece.
- 336 baselines committed (1.9 MB), the initial capture of the current tree.
- **This is the structural half only.** State continuity — data under keys a new pattern drops or renames — needs a replay against captured prior state, not a schema. That's follow-up work.

---
## Summary by cubic
Adds CI checks that block merges when a pattern’s argument/result schemas aren’t backward compatible, encode baselines with the Fabric JSON codec, and enforce an append‑only baseline store to prevent laundering breaking changes.

- **New Features**
  - Adds a GitHub Actions job “Pattern Update Compatibility” (sharded 4×) that runs `deno task pattern-compat` to compile each pattern and compare its schemas against all baselines in `packages/patterns/baselines/` using `assertPatternSchemasBackwardCompatible`.
  - Adds `deno task check-baselines-append-only` and a CI job to refuse deleted baselines unless the pattern file is removed in the same change.
  - Encodes baselines with `@commonfabric/data-model/codec-json` to preserve values like `-0`/`NaN`/`Infinity` and canonicalize key order; all baselines regenerated. Docs updated to describe the gate and workflow.
  - Hardens recording/evaluation: `--update` only records when the sole finding is “not recorded” (never records incompatible contracts), and unevaluable patterns are tracked so new failures fail the run; tests expanded to 44.

- **Migration**
  - When changing a pattern’s contract, run `deno task pattern-compat --update` to record the new schemas; CI still fails on incompatibility.
  - Do not delete baseline files. To retire a pattern, delete the pattern file; the append‑only check allows baseline removal only in that case. Non-pattern modules are skipped.

<sup>Written for commit 60edd2d4f201c6d7aa375b30f2e02b91b5be437b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5144?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

## Review round

Three blocking findings from self-review, all fixed:

1. **The append-only invariant was enforced nowhere.** The whole safety argument for `--update` is that it can only *add* a baseline — but `rm` is not `--update`, and the gate's own `retired` message suggested deleting baseline directories. Two commands laundered a break, invisible in a diff carrying tens of thousands of lines of JSON. Now enforced by a new `Pattern Baselines Append-Only` job (needs `fetch-depth: 0`), which permits deletion only when the pattern file goes in the same change. Fails closed (exit 2) if history is missing, so a shallow checkout can't make it vacuously green.
2. **`--update` recorded contracts that were never deployed.** It excluded *invalid* contracts but recorded *incompatible* ones — which by definition never merge, so never ship. In a never-pruned store that permanently constrains every future contract against a version that only existed in a failed run. `shouldRecord` now requires the sole finding to be "not recorded".
3. **Nine real pattern entries fail evaluation and were exempt by construction** (never evaluate → never recorded → never checked), reported in the same bucket as helper modules. `tasks/pattern-compat-unevaluable.ts` makes that debt explicit and can only shrink: a new failure fails the run, and so does a listed pattern that starts evaluating again.

Invalidity also became its own `Finding` kind rather than a `"(current)"` sentinel — the compiler immediately caught an unhandled reporting case, which is the point.

Separately, a latent bug the new tests caught: `isPatternSource` anchored exclusions to `packages/patterns/...`, so an absolute directory silently stopped excluding `integration/`, `tools/`, and `mod.ts`. Found because the test passed from the repo root and failed from `tasks/` — the workspace runner's cwd.

Tests: 13 → 44.

## Coverage debt

ACCEPT_COVERAGE_DEBT: coverage-debt: tasks uncovered lines = 1326 lines

The remaining debt is `tasks/pattern-compat.ts`'s `main()` — runtime construction, the per-pattern compile/evaluate loop, and failure reporting. Everything in that file that can be unit-tested has been extracted and tested: argument and shard parsing, baseline filename stamping, the Fabric codec round trip, the baseline store (`readBaselines` / `collectBaselineKeys` / `findRetired` / `writeBaseline`), `shouldRecord`, and the whole `checkPattern` core. What is left needs a real runtime and a real pattern tree, and its actual verification is that it runs against all 423 patterns on every PR.


seefeldb:
Thanks — went through all 11. Summary of what changed and what didn't.

## Fixed

**P1 — encode baselines with the Fabric codec** (codex, `tasks/pattern-compat.ts`). Valid, and the consequence is worse here than for a golden. I verified the cited precedent: `JSON.stringify` renders `-0` as `0` and `NaN`/infinities as `null`, and throws on a bigint. A baseline that flattened a value could never match the contract it was recorded from — the gate would report a missing baseline *forever*, for that pattern, on every run. Now encodes and decodes through `@commonfabric/data-model/codec-json`, same as `schema-generator/test/fixtures-runner.test.ts` and for the same reason. All 336 baselines regenerated. Bonus: the codec canonicalizes key order, so the files no longer depend on property insertion order. Covered by a test asserting `-0`/`NaN`/`Infinity` survive a round trip.

Also fixed the Coverage Check failure by extracting the testable parts of the task shell (arg parsing, shard parsing, filename stamping, the codec) into the tested lib. 8 tests → 19.

## Not taken

**P1 — remove the job's `timeout-minutes: 30`** (codex, `deno.yml`). AGENTS.md argues against timeouts used as *synchronization* — a wait that puts a ceiling on something which would otherwise complete. A CI job cap is a resource guard, not a synchronization primitive: every other job in this workflow carries the same `timeout-minutes: 30`, and this one was copied from `cfcheck`. Removing it only here would be inconsistent, and given this PR surfaced a schema whose validation ran for minutes, the cap is what turns "silently burns runner hours" into "fails visibly." Happy to revisit if the repo moves away from job caps generally.

## The nine baseline-content comments

All nine target the *contents* of files under `packages/patterns/baselines/` — the external `$ref` to `vnode.json`, `"type": "undefined"`, `"type": "unknown"`, a broad `anyOf` branch, an `asCell` asymmetry.

These are **generated artifacts**: a verbatim record of what the schema generator emits for each pattern. Hand-editing one is precisely the wrong move — the recorded contract would stop matching the compiled contract, and the very next run would report the baseline as missing. So these are observations about the schema generator's output (or about the patterns themselves), not about this changeset.

On the specific claim that an external `$ref` makes the check unreliable, I verified it rather than assuming. `system/space-overview.tsx`'s result schema carries `$UI: { $ref: "https://commonfabric.org/schemas/vnode.json" }`. Dropping its `summary` field:

```
system/space-overview.tsx
  cannot be applied over baseline 20260729T022742Z-G_uBgQqnFruj8BjY:
Pattern schemas are not backward compatible:
- result.summary: existing result field was removed
```

Exit 1. The unresolvable external ref does not neuter the check.

One of the nine is a fair point I'd rather acknowledge than dismiss: `aside.tsx`'s `JSXElement.anyOf` does contain a branch matching all objects, which genuinely weakens the guard for that pattern. That's a real limitation — **this gate is only as strong as the schemas the generator emits** — but it isn't something a baseline edit can fix, and narrowing that union is a change to the generator or to `JSXElement`, not to a recorded contract.

## Related

The pathology this PR turned up (`lobby/main.tsx` taking >180s in `assertPatternSchemasBackwardCompatible` against itself) is root-caused and fixed in #5145 — exponential re-walking of shared `$defs` maps, 74s → 2.1ms. Independently confirmed here: regenerating all 336 baselines takes 2:50 with that fix applied, versus 15+ minutes without.

PR #5145, opened 2026-07-29
perf(cfc): walk a schema's definition map once per root, not once per ref path

seefeldb:
Found by the new pattern-compat CI gate in #5144.

`cf piece setsrc` on `packages/patterns/lobby/main.tsx` effectively hangs today. `PieceController.setPattern` calls `assertPatternSchemasBackwardCompatible` on every source update, and that took **155 seconds** on lobby's contract. Lobby is not alone — three of the 54 contracts in the shipped pattern corpus are affected.

## Root cause

Not the comparison, and not schema size. Lobby's argument schema is 3.2 KB with 13 `$defs`; `topics`, at 27 KB, was fine.

It is the definition walk in **`validateSchemaDefinitionInternal`** (`packages/runner/src/cfc/schema-sanitization.ts`), in its `properties`/`patternProperties`/`$defs`/`definitions`/`dependentSchemas` loop.

`resolveCfcSchemaRef()` re-attaches the owning `$defs` object to every resolved `$ref` view:

```ts
// packages/runner/src/cfc/schema-refs.ts, resolveCfcSchemaRefUncached
if (schemaRefs.size > 0 && schemaCursor.$defs === undefined) {
  schemaCursor = {
    ...schemaCursor,
    ...(isRecord(fullSchema) && fullSchema.$defs && { $defs: fullSchema.$defs }),
  };
}
```

Every resolved view therefore carries a `$defs` key, and the walk re-validated **every** definition body from **every** view. The definition graph expanded as a tree instead of a DAG, so node visits grew as `(definition count) ^ (ref depth)`.

The construct in lobby that triggers it is its `$defs` **ref chain**, not a union and not a cycle:

```
LobbyAdminRegistryValue → LobbyAdminRegistryStoredValue → LobbyAdminList
  → LobbyAdminRole → LobbyAdminRoleAssignment → LobbyProfile
  → LobbyExternalProfileLink
```

Seven hops × 13 definitions re-walked per hop ≈ 13^7. `topics` is larger but its chains are shallow, which is why size was not the discriminator.

### Measured, not guessed

Instrumenting the traversal on lobby's argument schema:

| | node visits | object-identity guard hits | ref guard hits |
|---|---|---|---|
| as-is | **>3,000,001** (capped; real run exceeded 40M) | 564,427 | 298,835 |
| resolved views not re-walking the inherited `$defs` | **294** | 0 | 0 |

The recursion-guard hits are worth noting: they drop to **zero**. Lobby's schema has no recursive definitions at all — the re-attached `$defs` makes every definition reachable from every other, and the "cycles" the guard was cutting were an artifact of the amplification.

### This has bitten before, and was worked around in the pattern

`docs/plans/space-clone-rehearsal.md` records that a July rehearsal hit "a recursive `TopicPiece` crossref that made validation stop making progress" (#4997). That is this bug. #4997's fix was to "introduce a non-recursive `TopicReference` projection for mentionable/crossref targets, *so schema validation does not recursively traverse the full board graph*" — it changed only `packages/patterns/topics/main.tsx`, `topic.tsx` and a test. Flattening the pattern's `$defs` ref graph shortened the chain and brought the exponent down; the machinery was never touched. This PR fixes the machinery, so patterns no longer have to be shaped around it.

Two hypotheses were tested and **ruled out**:

- **Defeated hash caches.** The `WeakMap` caches in `schema-refs.ts` only engage for deep-frozen input, and `validateSchemaDefinition` builds a frozen interned clone that it discards. Validating a pre-frozen, pre-interned schema still took **55s** (vs 74s). Not the cause. (Also: it deliberately validates the *original*, so malformed literals stay ordinary diagnostics — swapping in the clone would be a semantic change.)
- **Exponential in union members / `anyOf`.** Lobby's blowup is entirely in the `$defs` ref chain; there are no nested unions on the hot path.

## The fix

Two changes in `validateSchemaDefinitionInternal`:

1. **Claim each definition map once per `(root, map)` per call**, on entry, so it belongs to the outermost schema that carries it. That schema still walks it later from its own keyword loop, where no ref expansion is in flight. This ordering is load-bearing — see the semantics note below.
2. **Memoize schemas that proved out under a root**, so a shared definition reached by a second path costs a lookup instead of a re-walk. A `cuts` counter tracks recursion-guard hits, and a subtree whose count moved is *not* memoized: a cut result is only sound while the schema that caused it is still on the stack to report its own issues.

The `data-model/schema-hash.ts` interning route was considered and not taken: it would only help if the input were already frozen, and the 55s frozen measurement shows the traversal shape is the problem, not cache misses.

### The ordering subtlety this went through

A first version claimed the map at whichever schema reached it first. A differential harness caught that as a real accept/reject flip: the first carrier is a resolved view *inside* a `$ref` expansion, so definition bodies got walked while the recursion guard held that ref open, and a body's own keywords went unchecked. `{ $ref: "#/$defs/Node", type: "bogus" }` under a recursive `Node` went from rejected to accepted. Claiming on entry fixes this, and the second new test pins it.

## Benchmarks

`validateSchemaDefinition`:

| contract | before | after |
|---|---|---|
| `lobby` argument | 74,263 ms | **2.1 ms** |
| `lobby` result | 5.2 ms | 2.2 ms |
| `cfc-group-chat-demo` argument | >20,000 ms (killed) | **3.7 ms** |
| `cfc-group-chat-demo` result | >20,000 ms (killed) | **4.2 ms** |
| `topics` result | 364 ms | 5.6 ms |
| `cfc-agent-prompt-injection-demo` result | 357 ms | 3.1 ms |

`assertPatternSchemasBackwardCompatible` on lobby's contract (a realistic `setsrc` edit — candidate adds an optional field to a shared definition, so the whole-contract deep-equality short circuit misses and the real subset walk runs): **155,288 ms → 10.2 ms**.

`schemaSubsetIssue` itself was measured and is *not* implicated: post-fix the whole compat call is 10.2 ms, of which the subset walk is ~4 ms. All of the >120s was the four `validateSchemaDefinition` calls that run before any comparison.

## Regression test

`packages/runner/test/cfc-schema-sanitization.test.ts` gains two tests.

**`walks a shared definition map once, not once per ref path`** asserts on a *counter*, not a clock, per `docs/development/waiting-in-tests.md`: it builds a depth-8 `$defs` ref chain, replaces one definition body with a counting accessor, and bounds how many times the validator reads it.

- Before: `AssertionError: Expected 463911 to be lower than 100` (fails in 4s)
- After: 8 reads, and the count does not grow with depth — 8 at depth 4, 6, 8, 10 and 12, against 103 / 3,991 / 463,911 / >2M / >2M before.

**`reports definition bodies a recursive ref would cut short`** guards the ordering subtlety above. It passes before and after by design; it exists so the claim cannot be moved back into a resolved view.

## Semantics unchanged

A faster validator that accepts or rejects different schemas would be a regression, so this was checked four ways.

- **`deno test -A packages/piece/test/schema-compatibility.test.ts`** — 1 passed (40 steps), 0 failed.
- **Runner CFC schema suites** — `deno test -A packages/runner/test/cfc-*.test.ts packages/runner/test/cfc.test.ts`: 150 passed (1,367 steps), 0 failed. Full `packages/piece` suite: 30 passed (349 steps), 0 failed.
- **Mutation differential** — 8,240 cases: 12 base schema shapes (ref chains, self- and mutual-recursive definitions, ref-site siblings, nested child-local `$defs` scopes, the legacy `definitions` keyword, every combinator, embedded `vdom`/`vnode` refs, dangling refs, unreferenced definitions) × both the one-argument and fragment+root entry points × 26 single-keyword mutations injected at every object position. Result: **0 accept/reject flips, 0 verdict changes.** The 400 cases whose output moved report the *same* verdict at a *shorter* path, because the issue is now found at the owning root rather than through a redundant ref re-expansion.
- **Real corpus differential** — all 54 compiled contracts in `packages/patterns/*/main.tsx`, before and after: **0 verdict changes** across the 51 that completed pre-fix (the other 3 are the pathological ones, which had no pre-fix verdict to compare).

### Residual risk

One narrow behaviour is inherent to any deduplication here, and I could not rule it out by construction. The pre-existing `activeRefs` guard returns "no issue" for a whole schema node without running that node's own keyword checks:

```ts
if (activeRefs?.has(schema.$ref)) return undefined;
```

Today's exhaustive re-walking gives extra chances to catch such a node via a path where its ref is not active. Deduplicating removes those extra chances. Reaching it needs a schema with a recursive `$ref` where a *different* node sharing that same ref string carries an independently-invalid sibling keyword. The mutation differential specifically targets this shape (`self-recursive`, `mutual-recursive`, `recursive-with-ref-site-siblings` bases, mutated at every position) and found no case. I deliberately did not "fix" the guard to also run the node's own checks, because that would make the validator strictly stricter — a semantic change beyond the scope of a performance fix.

### DIFF
--- a/packages/runner/src/cfc/schema-sanitization.ts
+++ b/packages/runner/src/cfc/schema-sanitization.ts
@@ -701,18 +701,21 @@ const DEFINITION_KEYS = ["$defs", "definitions"] as const;
  * the same map under the same root skips it.
  *
  * The claim is taken on entry, before any child is walked, so the map belongs
- * to the outermost schema that carries it. That schema walks it later from its
- * own keyword loop, where no ref expansion is in flight — a nested carrier
- * claiming first would instead walk the definitions from inside a `$ref` that
- * the recursion guard is holding open, and bodies reached through that ref
- * would be cut short.
+ * to the outermost schema that carries it. Refs only accumulate on the way
+ * down, so an in-flight walk always holds a subset of the refs any schema
+ * nested inside it would hold, and therefore cuts no more than that nested
+ * walk would — skipping the nested one loses nothing.
+ *
+ * That reasoning covers the claim only while the walk is in flight. Once it
+ * finishes, `releaseCutDefinitionScope()` hands the map back unless the walk
+ * ran to completion, because a sibling reached later holds different refs.
  */
 const claimDefinitionScopes = (
   schema: Exclude<JSONSchema, boolean>,
   rootKey: object,
   context: SchemaDefinitionContext,
-): ReadonlySet<string> => {
-  const claimed = new Set<string>();
+): ReadonlySet<string> | undefined => {
+  let claimed: Set<string> | undefined;
   for (const key of DEFINITION_KEYS) {
     const definitions = schema[key];
     if (!isRecord(definitions) || Array.isArray(definitions)) continue;
@@ -723,9 +726,25 @@ const claimDefinitionScopes = (
       context.walkedDefinitionsByRoot.set(rootKey, walked);
     }
     walked.add(definitions);
-    claimed.add(key);
+    (claimed ??= new Set()).add(key);
   }
   return claimed;
+};
+
+/**
+ * Give a definition map back when a recursion guard cut its walk short.
+ *
+ * A cut walk did not check everything below it, so it cannot stand in for a
+ * later walk that holds different refs open — through a sibling `$ref`, a
+ * definition this walk skipped is reachable in full. Only a walk that took no
+ * cut proves the whole map, and only that walk keeps the map for good.
+ */
+const releaseCutDefinitionScope = (
+  rootKey: object,
+  definitions: object,
+  context: SchemaDefinitionContext,
+): void => {
+  context.walkedDefinitionsByRoot.get(rootKey)?.delete(definitions);
 };
 
 const validateSchemaDefinitionInternal = (
@@ -990,12 +1009,9 @@ const validateSchemaDefinitionInternal = (
       // A definition map reached again under the same root has the same bodies
       // and resolves them against the same root, so whichever schema claimed it
       // on entry already covers it.
-      if (
-        (key === "$defs" || key === "definitions") &&
-        !claimedDefinitions.has(key)
-      ) {
-        continue;
-      }
+      const isDefinitionScope = key === "$defs" || key === "definitions";
+      if (isDefinitionScope && !claimedDefinitions?.has(key)) continue;
+      const cutsBeforeScope = context.cuts;
       for (const [name, child] of Object.entries(children)) {
         const issue = validateSchemaDefinitionInternal(
           child,
@@ -1004,6 +1020,9 @@ const validateSchemaDefinitionInternal = (
           context,
         );
         if (issue !== undefined) return issue;
+      }
+      if (isDefinitionScope && context.cuts !== cutsBeforeScope) {
+        releaseCutDefinitionScope(rootKey, children, context);
       }
     }
 
--- a/packages/runner/src/cfc/schema-sanitization.ts
+++ b/packages/runner/src/cfc/schema-sanitization.ts
@@ -668,6 +668,12 @@ interface SchemaDefinitionContext {
    * count moved must not be memoized as proven.
    */
   cuts: number;
+  /**
+   * Every `provenByRoot` record, in the order it was made. A schema proven
+   * while a definition map was merely claimed may have skipped that map on the
+   * strength of the claim, so handing the map back takes those records with it.
+   */
+  provenLog: Array<{ rootKey: object; schema: object }>;
 }
 
 /** Validate the schema language understood by strict Fabric migration checks. */
@@ -691,10 +697,12 @@ export const validateSchemaDefinition = (
     walkedDefinitionsByRoot: new WeakMap(),
     provenByRoot: new WeakMap(),
     cuts: 0,
+    provenLog: [],
   });
 };
 
 const DEFINITION_KEYS = ["$defs", "definitions"] as const;
+type DefinitionKey = (typeof DEFINITION_KEYS)[number];
 
 /**
  * Claim the definition maps this schema walks, so a schema that merely carries
@@ -714,8 +722,8 @@ const claimDefinitionScopes = (
   schema: Exclude<JSONSchema, boolean>,
   rootKey: object,
   context: SchemaDefinitionContext,
-): ReadonlySet<string> | undefined => {
-  let claimed: Set<string> | undefined;
+): ReadonlySet<DefinitionKey> | undefined => {
+  let claimed: Set<DefinitionKey> | undefined;
   for (const key of DEFINITION_KEYS) {
     const definitions = schema[key];
     if (!isRecord(definitions) || Array.isArray(definitions)) continue;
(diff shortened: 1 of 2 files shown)


seefeldb:
Both reviewers were right, and the issue was reachable — I verified it rather than reasoning about it.

The construction is codex's: a fragment carrying no `$defs` of its own visits `A` then `B`, where `A` references itself from a node with an invalid sibling keyword (so the recursion guard cuts that walk before checking it), and `B` re-enters the map afterwards. Same input against all three versions of the code:

| version | fragment + root entry point |
|---|---|
| `origin/main` (before the perf fix) | rejected ✓ |
| `433f4c982` (the perf fix as pushed) | **ACCEPTED** ✗ |
| `f1b1bb496` (this fix) | rejected ✓ — same message, same path as main |

So it was a real soundness regression, not a theoretical one, and it was specific to the fragment entry point that `assertSchemaSubset` uses. The whole-schema entry point rejected it in every version, because there the outermost carrier walks the map with no ref in flight and nothing is cut.

That last point matters for the test: **the original regression test used the whole-schema entry point, so it passed against the buggy version and guarded nothing.** The added test enters from a fragment, and is red against `433f4c982`, green here.

`releaseCutDefinitionScope()` hands the map back unless the walk ran to completion — the same `cuts` discipline `provenByRoot` already had. A cut walk didn't check everything below it, so it can't stand in for a later walk holding different refs open.

Perf is unaffected: `lobby/main.tsx`'s argument schema still validates in 2ms (74s before the perf fix). Semantics unchanged — `packages/piece/test/schema-compatibility.test.ts` plus the runner CFC merge / boundary / additive-default suites all pass (192 steps).

seefeldb:
Self-review found that `f1b1bb496` closed **half** the class. Two fail-open paths remained — head still accepted schemas `origin/main` rejects. I verified this myself with the same inputs against all four versions rather than taking the report on trust:

| version | Blocker 1 | Blocker 2 |
|---|---|---|
| `origin/main` | rejected ✓ | rejected ✓ |
| `433f4c982` | **ACCEPTED** | **ACCEPTED** |
| `f1b1bb496` | **ACCEPTED** | **ACCEPTED** |
| `2db70836c` | rejected ✓ | rejected ✓ (same messages, same paths as main) |

**1. The claim leaked through the `$ref` recursion guard.** `claimDefinitionScopes` runs on *entry*, but the active-ref guard returns immediately — before the release site inside the `$defs` iteration. So a `{$ref}` view claimed the owning map, walked nothing, and never handed it back; every later carrier then skipped that map for the rest of the call. Of every early return between the claim and the release, that guard is the only one that isn't an error aborting the whole validation, which is why it was the entire hole. Fixed with `settledDefinitions` plus a `finally` that releases anything the frame didn't finish, on every exit path.

**2. `provenByRoot` outlived the release.** The definition-scope skip doesn't move `cuts`, so a schema that skipped a claimed map still got memoized as proven. That's sound only while the claimer proves the map — but the claimer may *release* it, and the memo entry stayed, so a second path hit the memo and never re-walked the released map. Fixed with a `provenLog` whose window rolls back when a map is handed back.

Both are reachable through the fragment entry point `assertSchemaSubset` uses, which gates durable-link proofs and pattern updates — so fail-open here lets a schema outside the strict language reach a subset prover written assuming the gate held. Blocker 2 needs a deep-frozen root for identity-stable resolved views, which interning makes the norm rather than an exotic case.

Two regression tests added, both red at `f1b1bb496` and green now. Perf unchanged (lobby: 2ms). Semantics unchanged across 232 steps.

**Known remaining, not a regression and not blocking:** the memo still collapses whenever any recursion guard fires, because `cuts` is a single global counter — so a recursive definition keeps 2^depth growth on the properties side. That's 52ms at depth 12 against main's 50s, so a large net win either way, but the principled cure is recording *which* frame each cut depended on instead of a scalar count. Worth a follow-up given `vnodeSchema` is recursive.

PR #5146, opened 2026-07-29
fix(piece): default-root heal fires from EVERY entry point, not just boot

bfollington:
## Why

The updater that replaces an unloadable obsolete space root
(`PatternUpdater.checkDefaultPattern` — whose own docs say it exists "so an
unloadable obsolete root can self-heal") was invoked from exactly ONE door:
`PiecesController.ensureDefaultPattern`, the shell's `getSpaceRoot` boot
path. Every other consumer of the default root — piece registry listings,
`cf piece ls`, FUSE, the shell's own list cells — resolves through
`PieceManager.getDefaultPattern`, which started the stored pattern directly
and inherited **no heal at all**: the load failure propagated and every
listing died with the root.

Live consequence, 2026-07-29: the loom vendor space-compat gate's cold-load
probe (`cf piece ls`) failed on a root stranded by the cf-cell-context type
retirement (#5132) — a state the boot path would have healed. In a browser
the space "worked" only because boot happens to take the guarded door first;
every headless/registry consumer hit the corpse. The same asymmetry is why a
sequence of heal-internals fixes (#4967/#4977/#5133) improved the guarded
door while listings kept failing.

## What Changed

`PieceManager.getDefaultPattern` is the choke point every consumer already
goes through, so the heal now lives there: on a start failure with
`systemPatternAutoUpdate` on, run the same awaited
`patternUpdater.checkDefaultPattern`, and — as the boot path already
documents — **re-resolve the root after the swap commits** (the caller's
cell holds a pre-transaction snapshot) before retrying the start once. Any
other outcome rethrows the original failure untouched; with the flag off the
failure stays loud.

`deriveSystemPatternUrl` + the system pattern URLs move to
`system-pattern-url.ts` (the manager cannot import the controller that wraps
it); `pieces-controller.ts` re-exports them for existing importers.

## Test plan

- New: "heals an unloadable stale root on the REGISTRY path (not just
  boot)" — the suite's existing unloadable-root fixture, driven through
  `getPieceRegistry` instead of `ensureDefaultPattern`; asserts the roll
  to the official identity. Failed before the fix with the exact gate
  signature ("Could not load pattern …#default").
- New: flag-off variant asserts the failure still propagates loudly.
- Full `packages/piece` suite green (351 steps).
- Next: the loom vendor gate re-runs against a labs main containing this —
  it should pass on merit (no `--skip-space-compat`).

---
## Summary by cubic
Ensure the default space root self-heals across all entry points, not just during boot. This fixes cold-load failures in registry listings and headless tools when an obsolete stored root can’t load.

- **Bug Fixes**
  - Run self-heal in `PieceManager.getDefaultPattern` on start failure when `systemPatternAutoUpdate` is on, then re-resolve and retry once; otherwise rethrow. Tests cover the registry path heal and flag-off behavior.
  - Ensure the post-heal settle/re-resolve/retry path always surfaces the original start error if the retry fails. A regression test injects a secondary failure and asserts the original error is rethrown.

- **Refactors**
  - Extract `deriveSystemPatternUrl`, `HOME_PATTERN_URL`, and `DEFAULT_APP_PATTERN_URL` to `system-pattern-url.ts`, re-export from `pieces-controller.ts`, and drop the orphaned `MemorySpace` import.

<sup>Written for commit ec66f195754585bd7d098d80db583f4ee10e59fd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5146?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

ACCEPT_COVERAGE_DEBT: coverage-debt: packages/piece uncovered lines = 220 lines
(the heal-path guard/log-lambda lines; the paths themselves are executed by the three new tests — LCOV attribution, not untested behavior)

### DIFF
--- a/docs/specs/pattern-imports/pattern-updates.md
+++ b/docs/specs/pattern-imports/pattern-updates.md
@@ -126,6 +126,16 @@ Two decisions carry the whole design:
    fire-and-forget check, and the existing watcher re-instantiates a verified
    replacement in place. No new apply machinery.
 
+   The awaited default-root reconcile has a second entry point beyond space
+   open: `PieceManager.getDefaultPattern` — the resolution every registry
+   listing, CLI `piece ls`, FUSE mount, and shell list cell goes through —
+   runs the same awaited check when starting the persisted root FAILS, then
+   re-resolves the slot and retries the start once (a failed retry surfaces
+   the original start error). Without this, an unloadable obsolete root
+   healed only for consumers that happened to enter through space open;
+   every headless/listing consumer died with the root (the 2026-07-29
+   cf-cell-context retirement, caught by the loom vendor gate).
+
 A root created before provenance stamping may be admitted to this loop only
 when its stored `{ identity, symbol }` exactly equals the advertised current
 official entry appropriate to the space (`home.tsx` for Home,
--- a/packages/piece/src/manager.ts
+++ b/packages/piece/src/manager.ts
@@ -45,6 +45,8 @@ import {
   isLegacyPieceRegistryRoot,
 } from "../../runner/src/piece-helpers.ts";
 import { prepareSourceClosureVerification } from "../../runner/src/compilation-cache/cell-cache.ts";
+import type { PatternUpdateOutcome } from "../../runner/src/pattern-updater.ts";
+import { deriveSystemPatternUrl } from "./system-pattern-url.ts";
 ensureNotRenderThread();
 
 const PRIVILEGED_PIECE_LIST_SCHEMA = internSchema({
@@ -83,6 +85,12 @@ const PIECE_TRACE_TIMINGS = typeof Deno !== "undefined" &&
 // visible in the load summaries (browser worker included, where the
 // CF_CLI_TRACE_TIMINGS console path cannot run) as `piece/phase/<label>`.
 const pieceTimingLogger = getLogger("piece", { enabled: false });
+// Same channel the controller's heal path warns on, so a root healed through
+// the manager choke point shows up alongside the boot-path heal logs.
+const pieceHealLogger = getLogger("piece.update", {
+  enabled: true,
+  level: "warn",
+});
 
 async function timePiecePhase<T>(
   label: string,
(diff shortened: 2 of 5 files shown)


bfollington:
Blocking findings

1. `packages/piece/src/manager.ts:297-308` — the post-heal path does not preserve the original error on a secondary failure. The inner guard restores `error` if root resolution or `checkDefaultPattern()` fails, and non-updating outcomes also rethrow it, but after an `"updated"` / `"repaired-provenance"` outcome the `runtime.idle()`, slot resync, healed-root sync, and one-shot `this.get(...)` retry all sit outside that guard. Any of those failures therefore replaces the original root-start error. I verified this with an injected second-start failure: the caller saw the sentinel `secondary retry failure`, not the original `Could not load pattern …`. Please wrap the complete post-heal settle/re-resolve/retry sequence and rethrow the captured original error on failure, and add the corresponding regression test.

2. `packages/piece/src/ops/pieces-controller.ts:10` — the module move leaves `MemorySpace` imported but unused. `deno lint` fails with `no-unused-vars`, so this cannot pass the normal CI lint lane. Removing that type import should fix it.

The rest of the requested invariants check out:

- The heal catch is gated by both `runIt` and `systemPatternAutoUpdate`; the existing pre-bootstrap test also proves `getDefaultPattern(false)` stays read-only, while the new flag-off registry test proves the second gate.
- Non-updating updater outcomes and failures during the updater call rethrow the original start error.
- The retry settles, reads a fresh `spaceCell.key("defaultPattern")` slot, gets its current target, syncs that target, and starts that fresh cell; it does not retry the pre-swap snapshot. This is the correct stronger mirror of `startEnsuredDefaultPattern`'s re-resolution.
- Manager and controller both call `checkDefaultPattern` for the same root and therefore share PatternUpdater's `default-root + space + scope + id` single-flight key. The retry is syntactically one-shot inside the catch, so it cannot recursively re-enter the same catch. No double-heal/livelock found.
- `deriveSystemPatternUrl` and both URL constants remain re-exported from `pieces-controller.ts`; all current importers type-check. Apart from the stale import above, the module move is safe.
- The two added tests reproduce the actual registry-path unloadable-root failure and prove that the feature flag keeps the failure loud. They do not cover the secondary-failure/original-error contract from finding 1.

Verification run against PR head `1a4cfe406`:

- `packages/piece/test/check-update-default-pattern.test.ts`: 67 steps passed before the injected regression test.
- Focused `deno check` of the three changed source modules and every current moved-symbol importer: passed.
- `deno fmt --check` on changed files: passed.
- `deno lint` on changed files: failed only on the unused `MemorySpace` import above.

One documentation follow-up: `docs/specs/pattern-imports/pattern-updates.md` still describes default-root healing only as the space-open/pre-bootstrap path. This PR makes the manager choke point another awaited heal entry point, so the live behavior doc should be updated in the same change per the repository documentation policy.

— loom-review via Codex (GPT-5), on behalf of @bfollington

bfollington:
Both blocking findings addressed in ec66f1957:

1. **Original-error preservation**: the whole post-heal settle/re-resolve/retry sequence is now inside a guard that rethrows the captured original start failure on any secondary error (logged as `default-root-heal-retry-failed` for diagnosis). Regression test injects a post-heal failure exactly as the review did and asserts the caller sees "Could not load pattern", not the sentinel.
2. **Unused `MemorySpace` import**: removed in aae5814e2 (lint green locally).

Also took the documentation follow-up: `pattern-updates.md` now describes the manager choke point as the second awaited default-root heal entry beyond space open.

bfollington:
Codex verification pass (posted on its behalf — gh auth invalid in its env):

> Approve-style verdict: clean. Both blocking findings are closed, with no new findings.
>
> Verified at `ec66f1957`:
> - Focused suite: 68/68 steps passed.
> - Type-check, lint, and formatting passed.
> - Documentation follow-up is present.

PR #5147, opened 2026-07-29
fix(cli,runner): refuse verb payloads the runtime cannot turn into an event

mpsalisbury:
## Why

An agent that typos a field name in a verb payload was told the call succeeded, and then could not recover by retrying.

```
cf piece call --invocation abc recordMessage -- --json '{"mesage":"hi"}'
{"invocation": "abc", "status": "settled"}     # exit 0 — and nothing ran
```

`generateHandlerSchema` puts the event under `$event` and requires only `$ctx`, so a payload that fails the event schema never makes the argument invalid: `$event` reads back as `undefined`, the handler body runs with no event, and its receipt spends the invocation id. The corrected retry under that same id then collides on the receipt and reports settled a second time. The verb never runs, twice, and nothing anywhere says so.

That is precisely the failure an idempotency key exists to prevent, it is reachable by the single most likely agent mistake, and it sits on the path the verb contract exists to serve. Prior work in this arc: #5087 (D1), #5089 (D2), #5118 (D3).

## What Changed

**A supplied payload is measured against the verb's schema before dispatch** — `packages/cli/lib/callable.ts`, covering both callable entry points (`piece call` and mounted `exec`). Pre-dispatch is the only place this is catchable: once the event is sent, the id is already spent. Because nothing is sent, the id survives and the corrected retry executes.

Two shapes stay legal, because the runtime accepts them and refusing them would break working calls:

- **No payload at all.** `$event` is genuinely optional — value-less verbs are a supported shape.
- **A payload omitting a property that carries a `default`.** The runtime injects defaults on read (`runner/src/schema.ts:1377-1410`), so requiring them here would refuse calls the verb would have accepted. `relaxDefaultedRequired` drops those properties from `required`, recursively, following local `$ref` chains.

`VerbInputValidationError` reports as a data error — the message plus a pointer at `cf piece verbs --json` — rather than a Cliffy usage error, since the flags parsed fine; the values they carry don't fit the verb.

**`tx.handlingReceiptLink` is published only while `commitPreconditions` is on** — `packages/runner/src/runner.ts`. With the flag off, a `markCreateOnly` mark is silently dropped at commit (`cfc/grants.ts:259-269` documents exactly this), so nothing creates the receipt: the address named a witness that does not exist and invited a readback against an unwritten cell. This matches the fail-closed stance `grants.ts` already takes for single-use grants under the same flag.

### Behavior change worth a reviewer's eye

Five tests asserted the old pass-through (`"…without CLI shape enforcement"`). They characterized the **parser** — which still passes JSON through verbatim, unchanged here — but asserted it at the **execution** layer, where they encoded "a payload the verb demonstrably cannot accept is forwarded anyway." They now assert refusal. The three genuinely parser-level tests in `exec.test.ts` are untouched and still pass.

**Residual risk, stated plainly:** this validates with `validateSchemaValue` rather than replaying the runtime's own read, so a divergence between the two would surface as a refused-but-valid call. I probed ten schema shapes and found exactly one — defaults not satisfying `required` — which is fixed and tested above; `asCell`, links, `$ref`/`$defs`, `true`, and primitive schemas all behaved. I can't bound what I didn't probe. The `piece-integration` suite exercises this path against real deployed patterns, so CI covers the obvious regressions.

## Test plan

- **13 unit tests** for the validator: match, missing required, wrong type, absent payload, no schema, defaults (flat / nested / behind `$ref` / behind a `$ref` chain), a non-defaulted sibling still rejected, `asCell` given either a plain value or a link, self-referential schema, `$ref` cycle.
- **Execution-layer refusal** at both entry points, asserting nothing reached the stream and no send options were recorded — i.e. the id was not spent.
- **Runner:** a characterization test recording what the runtime actually does with a mismatched payload (body runs, event absent, receipt `{}`) — the evidence for why the gate must be pre-dispatch; and a test that no receipt address is published while `commitPreconditions` is off.
- **Integration** (`piece-call` shard, scenario 5): a typo'd payload is refused with the verb named and records nothing, then the corrected retry **under the same invocation id** records exactly one message carrying the corrected payload.
- Full suites green locally: runner 1077 passed / 0 failed; cli 0 failed; `piece-call` integration section run against a local toolshed.

---
## Summary by cubic
Reject invalid verb payloads before dispatch in `cf piece call` and mounted `exec` so typos don’t “settle” and burn the invocation id. Also publish `tx.handlingReceiptLink` only when `commitPreconditions` is on, and relax required fields only when a local `#/$defs` default exists.

- **Bug Fixes**
  - `piece call` and mounted `exec` validate a supplied payload against the verb’s schema before sending (`packages/cli/lib/callable.ts`). Bad input throws `VerbInputValidationError`; the CLI prints it as a data error with a hint to run `cf piece verbs --piece <piece> --json` (`packages/cli/commands/piece.ts`, `packages/cli/mod.ts`).
  - Still accepts two runtime-supported shapes: no payload; and payloads omitting properties with a `default` (including nested and via local `#/$defs`). `#/definitions` and non-local refs are not followed; unresolvable refs keep the field required to avoid spending the id.
  - Runner publishes `tx.handlingReceiptLink` only when `commitPreconditions` is `true` (`packages/runner/src/runner.ts`); with the flag off, no receipt address is exposed. Docs updated (`docs/plans/pattern-verb-contract-implementation.md`, `docs/development/EXPERIMENTAL_OPTIONS.md`).

- **Migration**
  - Calls that previously “settled” with a missing or malformed event now fail locally; fix the payload and retry with the same invocation id.
  - When `commitPreconditions` is off, no receipt address is returned; don’t rely on it in that mode.

<sup>Written for commit 629395256d389e2a0a36adc16baee29d5712bf12. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5147?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mpsalisbury:
@cubic-dev-ai please re-review — your review was of `ef0e7c4dd` (2026-07-29) and the branch has moved since.

Since that review:

- **Rebased onto post-#5108 main** (60 commits). #5108 removed the machine-speed dependency in `traverse.ts` that was causing the `packages/runner ±39` coverage swing; the gate now passes and reports debt *reduced* by 10 lines.
- **Extracted `reportVerbInputErrorOrRethrow`** from the inline `.catch` in the `piece call` command action. The coverage gate flagged 7 uncovered lines there, correctly: anything written in a Cliffy action body is unreachable from a unit test. It now threads `exitWithDataError`'s existing `deps` seam, with three tests — a rejected payload reports and exits 1, a missing piece id renders as `<piece>`, and an unrelated failure re-throws untouched.
- **A design-review note** added to the D1 bullet: the receipt is not derivable from the event id alone (its cause is the handler's bound closure *plus* the event id), which is why the address rides the commit callback rather than being reconstructed client-side. Folded in from a now-closed PR (#5203).

The substance of the fix is unchanged from your last pass. The re-throw test is the one I would most want a second opinion on — a network failure misreported as a payload problem would send an agent to fix a payload that was fine.

mpsalisbury:
## Measured: what a handler sees when defaulted-required fields get no payload

The follow-up plan for this gate (D5, recorded in `docs/plans/pattern-verb-contract-implementation.md`) named one question to settle by experiment, not assumption. It is now measured, on this branch, with a scratch runner test beside this PR's own characterization test. Event schema: `{ properties: { a: { type: "string", default: "fallback" }, b: { type: "number", default: 7 } }, required: ["a", "b"] }`.

| payload sent | handler observes as `event` | receipt |
| --- | --- | --- |
| absent (`undefined`) | **`undefined`** — defaults never engage | written; id spent |
| `{ a: "supplied" }` | `{ a: "supplied", b: 7 }` — missing defaulted-required filled in | written |
| `{}` | `{ a: "fallback", b: 7 }` — both defaults filled in | written |

**Where this happens:** defaults are materialized in `SchemaObjectTraverser.traverseObjectWithSchema` (`packages/runner/src/traverse.ts`) — for a *present* object, each schema-declared property missing from the value gets its `default` merged in, and the `required` check runs **after** that merge. For an absent event, only the schema's own top-level `default` is consulted (`applyDefault`); the property branch is never entered, `$event` is omitted from the handler argument (`generateHandlerSchema` never requires it), and the handler receives `undefined`.

**Implications:**

- **For this PR as written: supportive.** `relaxDefaultedRequired` is semantically honest for what this gate judges — a present-but-partial object really is completed from defaults before `required` is checked, so treating a defaulted-required property as satisfied matches what the runtime does.
- **For D5 (the absent-payload follow-up): relaxation is the wrong lens for the *absence* decision.** An all-defaulted `required` list does not make a wholly absent payload deliverable — nothing on the dispatch path conjures the defaults object; the handler gets `undefined` and the invocation id is spent anyway. The remaining choice for D5 is between refusing absence on *unrelaxed* `required`, or normalizing an absent payload to `{}` so defaults engage. That choice is marked open in the plan; the measured behavior belongs in the helper's doc comment either way.

<details>
<summary>Scratch test (not landed; runs green on this branch with the runner package's own test invocation)</summary>

```ts
// SCRATCH EXPERIMENT (do not commit): what does a handler observe when its
// event schema declares defaulted properties that `required` also names, and
// the dispatched payload is absent / partial / empty?
//
// Modeled on scheduler-event-receipts.test.ts ("delivers a schema-missing
// payload as an absent event and still receipts it").
import {
  afterEach,
  beforeEach,
  createSchedulerTestRuntime,
  describe,
  disposeSchedulerTestRuntime,
  expect,
  it,
  Runtime,
  space,
} from "./scheduler-test-utils.ts";
import type {
  Cell,
  IExtendedStorageTransaction,
  SchedulerTestStorageManager,
} from "./scheduler-test-utils.ts";
import { createTrustedBuilder } from "./support/trusted-builder.ts";
import { resolveLink } from "../src/link-resolution.ts";

async function waitForSchedulerCondition(
  runtime: Runtime,
  condition: () => boolean,
  message: string,
): Promise<void> {
  const deadline = performance.now() + 1_000;
  while (!condition() && performance.now() < deadline) {
    await runtime.idle();
  }
  if (!condition()) {
    throw new Error(message);
  }
}

function receiptCellForEvent<T>(
  runtime: Runtime,
  eventId: string,
): Cell<T> {
  return runtime.getCell<T>(
    space,
    { resultFor: { $ctx: {}, $event: eventId } },
  );
}

function resolvedStreamLink(streamCell: Cell<unknown>, runtime: Runtime) {
  return resolveLink(
    runtime,
    runtime.readTx(),
    streamCell.getAsNormalizedFullLink(),
  );
}

// Snapshot the handler's event argument as plain data (strip any runtime
// annotation symbols; keep only string-keyed own enumerable properties).
function snapshotEvent(event: unknown): unknown {
  if (event === undefined) return undefined;
  if (event === null || typeof event !== "object") return event;
  return Object.fromEntries(Object.entries(event as Record<string, unknown>));
}

describe("scratch: defaulted-required event schema", () => {
  let storageManager: SchedulerTestStorageManager;
  let runtime: Runtime;
  let tx: IExtendedStorageTransaction;

  beforeEach(() => {
    ({ storageManager, runtime, tx } = createSchedulerTestRuntime(
      import.meta.url,
    ));
  });

  afterEach(async () => {
    await disposeSchedulerTestRuntime({ storageManager, runtime, tx });
  });

  it("records the event argument for absent, partial, and empty payloads", async () => {
    const { commonfabric } = createTrustedBuilder(runtime);
    const { handler, pattern } = commonfabric;
    let handlerInvocations = 0;
    const seenEvents: unknown[] = [];
    const defaultedVerb = handler(
      {
        type: "object",
        properties: {
          a: { type: "string", default: "fallback" },
          b: { type: "number", default: 7 },
        },
        required: ["a", "b"],
      },
      { type: "object", properties: {} },
      (event: { a: string; b: number } | undefined) => {
        handlerInvocations++;
        seenEvents.push(snapshotEvent(event));
      },
    );
    const rootPattern = pattern(() => ({ stream: defaultedVerb({}) }));
    const rootCell = runtime.getCell<{ stream: unknown }>(
      space,
      "defaulted required event root",
      undefined,
      tx,
    );
    const root = runtime.run(tx, rootPattern, {}, rootCell);
    await tx.commit();
    tx = runtime.edit();
    await root.pull();

    const streamLink = resolvedStreamLink(root.key("stream"), runtime);

    // (a) absent payload: dispatched with no event value at all.
    const absentEventId = "evt:defaulted-required:0:absent";
    runtime.scheduler.queueEvent(
      streamLink,
      undefined,
      undefined,
      undefined,
      false,
      { eventId: absentEventId },
    );
    await waitForSchedulerCondition(
      runtime,
      () => handlerInvocations === 1,
      "absent-payload event did not run",
    );

    // (b) partial payload: one defaulted-required property missing.
    const partialEventId = "evt:defaulted-required:1:partial";
    runtime.scheduler.queueEvent(
      streamLink,
      { a: "supplied" },
      undefined,
      undefined,
      false,
      { eventId: partialEventId },
    );
    await waitForSchedulerCondition(
      runtime,
      () => handlerInvocations === 2,
      "partial-payload event did not run",
    );

    // (c) empty-object payload: every defaulted-required property missing.
    const emptyEventId = "evt:defaulted-required:2:empty";
    runtime.scheduler.queueEvent(
      streamLink,
      {},
      undefined,
      undefined,
      false,
      { eventId: emptyEventId },
    );
    await waitForSchedulerCondition(
      runtime,
      () => handlerInvocations === 3,
      "empty-payload event did not run",
    );
    await runtime.scheduler.idleWithPendingCommits();

    const [absentObserved, partialObserved, emptyObserved] = seenEvents;

    const absentReceipt = receiptCellForEvent<Record<string, unknown>>(
      runtime,
      absentEventId,
    );
    await absentReceipt.pull();
    const partialReceipt = receiptCellForEvent<Record<string, unknown>>(
      runtime,
      partialEventId,
    );
    await partialReceipt.pull();
    const emptyReceipt = receiptCellForEvent<Record<string, unknown>>(
      runtime,
      emptyEventId,
    );
    await emptyReceipt.pull();

    console.log("OBSERVED absent  ->", Deno.inspect(absentObserved));
    console.log("OBSERVED partial ->", Deno.inspect(partialObserved));
    console.log("OBSERVED empty   ->", Deno.inspect(emptyObserved));
    console.log("RECEIPT absent   ->", Deno.inspect(absentReceipt.get()));
    console.log("RECEIPT partial  ->", Deno.inspect(partialReceipt.get()));
    console.log("RECEIPT empty    ->", Deno.inspect(emptyReceipt.get()));

    expect(handlerInvocations).toBe(3);
    // Value-position assertions, pinned to the recorded observations:
    expect(absentObserved).toEqual(undefined);
    expect(partialObserved).toEqual({ a: "supplied", b: 7 });
    expect(emptyObserved).toEqual({ a: "fallback", b: 7 });
    // A receipt is written for every dispatch, valid-looking payload or not.
    expect(absentReceipt.get()).toEqual({});
    expect(partialReceipt.get()).toEqual({});
    expect(emptyReceipt.get()).toEqual({});
  });
});
```

</details>

*Measured by a Fable agent session on this branch (1ed1fa90e); scratch test kept out of the PR on purpose — D5 will land the durable version of this pin.*

PR #5149, opened 2026-07-29
fix(runner): converge an unloadable durable pattern pointer back to the running pattern (CT-1923)

bfollington:
## Why

The last member of this week's heal family (roots at boot: #4967; roots from
every entry point: #5146). A running piece whose **durable** `patternIdentity`
names an identity the runtime cannot load sat stranded forever: the watcher
saw the pointer as a change, failed the load, logged `pattern-load-error` —
and left the unloadable pointer in place. Every later session started the
piece by that dead pointer and rendered nothing.

Live on estuary (2026-07-29, deploy `ab8e132d48`): nested home-section pieces
whose stored vintage used the retired `cf-cell-context` element. The parent
re-instantiates the CURRENT sub-pattern each boot and optimistically writes
the pointer inside its batched instantiation commit — but that commit loses
to the incoming durable state, the pointer never rolls forward, and the
stranded state reasserts itself per session: blank Favorites and Profile tabs
on the user's home space, with `pattern-load-error … uMloj4n9…` on every
load.

## What Changed

With `systemPatternAutoUpdate` on, a **definitive** load failure (undefined
from `loadPatternByIdentity` — the load synced and found the identity's docs
absent or uncompilable) while a pattern is running rolls the pointer back to
the RUNNING pattern's identity through a small preconditioned commit:

- a concurrent legitimate repoint supersedes it (the commit re-reads the
  pointer and bails if it moved);
- thrown (possibly transient) load errors never trigger it — only the
  synced-verdict undefined;
- keyless patterns' synthetic session pointers are never written durably;
- flag off: behavior unchanged (pinned by test).

`startCore` now tracks the RUNNING pattern's ref separately from the last
observed pointer key — a parent-driven start instantiates its given pattern
while the durable pointer may still hold an older identity, and the
roll-forward must write what is actually live, not what was last observed.

The stranded state becomes self-converging instead of self-perpetuating: one
session with the flag on repairs the pointer durably, and every subsequent
session (and every start-by-identity consumer) finds a loadable pattern.

## Test plan

- New `pattern-pointer-unloadable-swap.test.ts`: reproduces the exact
  stranded state (running V1, durable pointer moved to a well-formed identity
  persisted nowhere) — flag ON converges the pointer back durably with the
  running pattern untouched (failed before this change: pointer stayed
  stranded); flag OFF leaves the pointer as written.
- Watcher/heal-family suites green: pattern-swap-link-argument,
  nested-piece-setup-repair, resume-output-redirect-partialcause,
  ensure-piece-running, runner.test.ts (its one pre-existing `clock` failure
  on main aside). Full `packages/piece` suite green (352 steps).

---
## Summary by cubic
Self-heals durable `patternIdentity` pointers by reverting unloadable pointers to the currently running pattern on a definitive load failure, making stranded pieces converge and fixing blank Home tabs. Also cleans up shutdown by settling pointer roll-forward commits (not watcher loads) to avoid teardown hangs. Addresses CT-1923.

- **Bug Fixes**
  - On a definitive `undefined` from `loadPatternByIdentity` during a run and with `systemPatternAutoUpdate` on, roll the durable pointer back to the running pattern via a guarded commit. Bail if the pointer changed, and only reclaim the watcher key while it is still the failed one.
  - Never trigger on thrown/transient load errors, when `cfcEnforcementMode="disabled"`, or for session-synthetic `keyless:` refs (via `PatternManager.isKeylessPatternIdentity`). Track the running pattern ref separately from the last observed pointer key.
  - Track watcher loads and pointer roll-forward commits; expose `Runner.idlePointerMaintenance()` for deterministic tests. `Runtime.dispose()` now calls `Runner.settlePointerCommits()` to await commits only before storage sessions close.

<sup>Written for commit 8dbe4012eddc9f48321fe752dfd471cd4f525343. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5149?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

ACCEPT_COVERAGE_DEBT: coverage-debt: packages/runner uncovered lines = 6961 lines
(the roll-forward failure-path warn lambdas and dispose-settle guard lines; the behavior paths themselves are pinned by the four regression tests, all run under the frozen-clock preload)

### DIFF
--- a/packages/runner/src/pattern-manager.ts
+++ b/packages/runner/src/pattern-manager.ts
@@ -379,6 +379,16 @@ export class PatternManager {
   }
 
   /**
+   * Whether `identity` is a session-synthetic keyless pointer (minted by
+   * {@link ensureKeylessPatternIdentity}) rather than a durable
+   * content-addressed artifact identity. A fresh runtime can never load a
+   * keyless pointer, so such refs must never be written into durable state.
+   */
+  static isKeylessPatternIdentity(identity: string): boolean {
+    return identity.startsWith("keyless:");
+  }
+
+  /**
    * Make a cross-space child piece independently loadable from its own space
    * (CT-1687). A fresh runtime navigating to a `Factory.inSpace(...)` child
    * loads pattern artifacts from the CHILD's space — but the parent bundle's
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -1820,6 +1820,12 @@ export class Runner {
     // pattern can only change via a fresh run(), not via the meta watcher).
     const KEYLESS = "\0keyless";
     let currentPatternKey: string | undefined;
+    // The identity of the pattern whose nodes are LIVE right now — which can
+    // differ from `currentPatternKey` (the pointer value last observed): a
+    // parent-driven start instantiates its given pattern while the durable
+    // pointer may still hold an older identity. The unloadable-pointer
+    // roll-forward below writes THIS ref back, never the observed key.
+    let runningRef: { identity: string; symbol: string } | undefined;
     let cancelNodes: Cancel | undefined;
     let initialSchedulerRehydrationAvailable = true;
 
@@ -1924,6 +1930,7 @@ export class Runner {
         }
         cancelNodes?.();
         instantiatePattern(pattern);
+        runningRef = newRef;
       };
       addCancel(
         resultCell.sinkMeta("patternIdentity", (newValue) => {
@@ -1962,6 +1969,61 @@ export class Runner {
                   "pattern-load-error",
                   `Failed to load pattern ${newRef.identity}#${newRef.symbol}`,
                 );
+                // CT-1923: an undefined result is a DEFINITIVE verdict — the
+                // load synced and found the identity's docs absent or
+                // uncompilable — while a pattern is still running here. Left
+                // alone, the durable pointer keeps naming an identity no
+                // session can load: every boot re-logs this error and any
+                // start-by-identity of this piece is dead (the stranded
+                // blank-section state). Roll the pointer back to the RUNNING
+                // pattern's identity, durably, with a superseded-check so a
+                // legitimate concurrent repoint wins. Gated on the same flag
+                // as the rest of the heal family; thrown load errors (which
+                // may be transient) never trigger this.
+                if (
+                  this.runtime.experimental.systemPatternAutoUpdate &&
+                  runningRef !== undefined &&
+                  patternIdentityKey(runningRef) !== newKey
+                ) {
+                  const revertRef = runningRef;
+                  void this.runtime.editWithRetry((tx) => {
+                    const cur = asPatternIdentityRef(
+                      resultCell.withTx(tx).getMetaRaw("patternIdentity", {
+                        meta: ignoreReadForScheduling,
+                      }),
+                    );
+                    if (!cur || patternIdentityKey(cur) !== newKey) {
+                      return false;
+                    }
+                    resultCell.withTx(tx).setMetaRaw("patternIdentity", {
+                      identity: revertRef.identity,
+                      symbol: revertRef.symbol,
+                    });
+                    return true;
+                  }).then((result) => {
+                    if (result.ok) {
+                      currentPatternKey = patternIdentityKey(revertRef);
+                      logger.warn(
+                        "unloadable-pointer-rolled-forward",
+                        () => [
+                          "durable patternIdentity named an unloadable",
+                          `identity (${newRef.identity}#${newRef.symbol});`,
+                          "rolled back to the running pattern",
+                          `${revertRef.identity}#${revertRef.symbol}`,
+                        ],
+                      );
+                    }
+                  }).catch((error) => {
+                    logger.warn(
+                      "unloadable-pointer-roll-forward-failed",
+                      () => [
+                        "could not roll the unloadable pointer back to the",
+                        "running pattern",
+                        error,
+                      ],
+                    );
+                  });
+                }
                 return;
               }
               logger.info("pattern changed", {
@@ -2097,6 +2159,12 @@ export class Runner {
       currentPatternKey = initialRef ? patternIdentityKey(initialRef) : KEYLESS;
       try {
         instantiateInitialPattern(givenPattern, initialRef, tx);
+        // Real artifact refs only: a keyless pattern's synthetic session
+        // pointer must never be written into durable state by the
+        // unloadable-pointer roll-forward.
+        runningRef =
+          this.runtime.patternManager.getArtifactEntryRef(givenPattern) ??
+            undefined;
       } catch (error) {
         // Without cleanup the piece stays registered in `this.cancels`, so
         // every later start() reports "already running" for a piece that has
@@ -2135,6 +2203,7 @@ export class Runner {
       initialRef,
       tx,
     );
+    runningRef = initialRef;
     if (!doNotUpdateOnPatternChange) {
       setupPatternWatcher();
     }
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -13,6 +13,7 @@ import { hashOf } from "@commonfabric/data-model/value-hash";
 import { toCompactDebugString } from "@commonfabric/data-model/value-debug";
 import { getLogger } from "@commonfabric/utils/logger";
 import { isRecord } from "@commonfabric/utils/types";
+import { PatternManager } from "./pattern-manager.ts";
 import { rendererVDOMSchema } from "./schemas.ts";
 import { forEachSubschema } from "./schema-walk.ts";
 import {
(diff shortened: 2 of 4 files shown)


bfollington:
Blocking — I found three correctness problems plus a deterministic test failure.

1. **`undefined` is not definitive when CFC recovery is disabled.** The new gate checks only `systemPatternAutoUpdate`, but `loadPatternByIdentity()` explicitly returns `undefined` whenever `cfcEnforcementMode === "disabled"` and the ref is not already in this runtime's in-memory index. That is “probe unsupported”, not “artifact dead”; #5146's updater already has the required CFC-disabled guard and a test documenting exactly this distinction. The pointer precondition does not protect this case: after a legitimate repoint has committed, the pointer still equals `newKey`, so this code successfully writes the old running ref over it. I verified this with two runtimes sharing a store: runtime A persisted V2, a CFC-disabled runtime B was running V1, and repointing B to the loadable V2 was rolled back to V1. Please gate the rollback on by-identity recovery being enabled and pin that case.

2. **The claimed keyless exclusion does not hold.** By the time `startCore()` records `runningRef`, `setupInternal()` has already called `entryRefForPattern()`, which mints a `keyless:` ref through `ensureKeylessPatternIdentity()` and indexes it via `setArtifactEntryRef()`. Therefore `getArtifactEntryRef(givenPattern)` returns the synthetic ref, not `undefined`. I verified that an unloadable repoint of a running hand-built pattern is durably rolled back to `keyless:...`; a fresh runtime is guaranteed not to load that pointer, so this does not converge the durable state. Please distinguish durable artifact refs from session-synthetic refs (and add the missing test) rather than relying on `getArtifactEntryRef()`.

3. **A repoint just after the rollback commit can be lost locally.** I independently agree with Cubic's finding at the unconditional `currentPatternKey = patternIdentityKey(revertRef)`. The durable precondition protects changes before/during the rollback commit, but not a new watcher notification after that commit and before this continuation runs. In that window the watcher records C and starts C's load, then this continuation overwrites the local key with R; C's completion fails `currentPatternKey !== newKey`, while the durable pointer remains C. Only update the local key if it is still the failed `newKey` (or let the rollback's watcher notification own it).

4. **The new test fails under Runner's canonical clock preload.** `packages/runner/deno.jsonc` and CI both preload `test/clock-preload.ts`, which intentionally freezes test sleeps. The `setTimeout` polling therefore leaves the suite promise pending; locally the exact runner command failed with `Promise resolution is still pending but the event loop has already resolved`, and CI's **Runner Tests (1/5)** is red. Without the preload the file passes, which hides the contract violation. The flag-OFF case is also a false-positive risk even under real time: it waits 500 ms without first proving the floating load completed, so a slower broken rollback would pass. Please synchronize on controlled loader/commit promises instead of wall-clock polling.

The rest of the requested scrutiny looked sound:

- In enforcing mode, the storage path explicitly `sync()`s both compiled and source entry cells, and canonical compilation persists artifacts before the pointer write, so I found no separate ordinary replica-lag hole beyond the CFC-disabled verdict above.
- `runningRef` is correct on the identity-start path and after a successful swap; setup failure and load failure do not update it. The given-pattern path is correct for real refs but has the keyless hole above.
- #5146's manager heal runs after start failure, when no watcher is live. The updater's `stillMatches`/`canWrite` and this rollback's transactional pointer reread make their durable commits supersede one another rather than fight. I found no livelock once the unsafe verdict/key races above are fixed.

Verification:

- Exact head reviewed: `f47340694f87b0a8cccff2b7c727c95009e85c71`
- Canonical new-test run with clock preload: **failed**
- Two temporary focused regressions (CFC-disabled loadable repoint; keyless rollback), run without the frozen-clock preload: **both reproduced**

— loom-review via Codex (GPT-5), on behalf of @bfollington

bfollington:
Both findings addressed in 34f6ee0e9:

- **CI shard failure**: all 279 tests passed; the job died on Deno's exit check ("Promise resolution is still pending…") — the roll-forward chain (and the watcher's load attempt, which these tests are the first to dispose mid-flight) could be pending at teardown. Roll-forward commits are now tracked on the Runner and settled in `dispose()` before storage closes — the same contract `PatternUpdater.dispose()` follows — and the test drains before teardown (3× stable locally).
- **cubic P2** (valid): the post-rollback continuation reclaimed `currentPatternKey` unconditionally, which could clobber a valid repoint that raced the rollback and make the key guard discard that repoint's in-flight load. It now reclaims only while the observed key is still the failed one.

bfollington:
All four review findings addressed in 00b09951f (finding 3 was already fixed in 34f6ee0e9):

1. **CFC-disabled undefined ≠ verdict**: rollback now additionally gated on `cfcEnforcementMode !== "disabled"`. Pinned by your exact scenario as a regression test: an enforcing session persists a loadable V2 into the shared store; a CFC-disabled session running V1 is repointed to V2 — the repoint survives.
2. **Keyless exclusion made real**: you were right that setup mints AND indexes the `keyless:` ref, so `getArtifactEntryRef` returned it. A shared `PatternManager.isKeylessPatternIdentity` predicate now filters synthetics at both the `runningRef` assignment and the rollback gate; pinned by a hand-built-pattern test asserting the unloadable repoint is left alone rather than gaining a pointer no fresh runtime can load.
3. **Key reclaim race**: guarded in 34f6ee0e9 (reclaims only while the observed key is still the failed one).
4. **Frozen-clock test contract**: no more wall-clock polling. The watcher's whole load chain joins `pendingPointerMaintenance`, and `idlePointerMaintenance()` loops until drained — the single deterministic synchronization point for `dispose()` and the tests. All four cases pass under the canonical `--preload=test/clock-preload.ts` run, which also removes the flag-OFF false-positive window you flagged (the assertion now runs strictly after the load attempt settles).

bfollington:
Blocking — finding 4 is not closed at exact head `00b09951f2af7ba23a520d4f28a529f9270e283f`.

`pendingPointerMaintenance` now tracks the entire watcher `loadPatternByIdentity()` chain, and `Runtime.dispose()` waits for that set after `stopAll()`. `stopAll()` invalidates the continuation, but it cannot settle or cancel a load that is itself held. The existing contract test `dispose invalidates a held pattern hot-swap load` therefore suspends at `await runtime.dispose()` before it can release the held load.

Verified two ways under the canonical clock preload:

- `reload-rehydration-safety.test.ts`: 14 tests complete, then Deno exits with `Promise resolution is still pending but the event loop has already resolved`; the unfinished case is exactly `dispose invalidates a held pattern hot-swap load`.
- Exact CI shard 1 locally: all **278 tests / 851 steps / 0 assertion failures** complete, then the same pending-promise exit occurs. GitHub’s `Runner Tests (1/5)` is red for this exact head in the same lane.

This is also a product lifecycle regression, not only a test issue: `Runtime.dispose()` can now wait forever when a by-identity hot-swap probe stalls, so storage cleanup never completes. Please keep deterministic test synchronization for the watched load chain separate from the disposal barrier for actual roll-forward writes, or make the watched load cancelable/settle on `stopAll()`. Disposal must settle commits that can still write, but must not await an externally stalled probe after the continuation has been invalidated.

The other three blockers are closed in source: the CFC-disabled verdict gate is present, keyless refs are filtered at both `runningRef` assignment and the rollback gate, and `currentPatternKey` reclaim is guarded on still being the failed key. The new four-case focused test passes under the canonical preload, as do changed-file check/lint/fmt; the full shard regression above prevents an approve verdict.

One test-hygiene follow-up while touching this: the CFC-disabled case calls `other.dispose()` while `other` and `rt` share the same `storageManager`, which closes the emulated server before `rt` runs and produces `memory replica closed`. Keeping the producer alive (or otherwise preserving the shared store) would make the stated “loadable V2 persisted by another runtime” premise true rather than relying only on the direct gate.

— loom-review via Codex (GPT-5), on behalf of @bfollington

bfollington:
Shard-1's persistent pending-promise exit is root-caused and fixed in 8dbe4012e. Bisecting the shard file-by-file landed on reload-rehydration-safety's "dispose invalidates a held pattern hot-swap load" — which holds `loadPatternByIdentity` open across `dispose()`. My previous commit had dispose await the whole watcher chain, so one held (or, in production, wedged) load could hang teardown forever — the fire-and-forget design epoch-guards post-settle work for exactly this reason.

Split the tracking: `dispose()` now settles only the bounded local roll-forward COMMITS (`settlePointerCommits`); the load attempts live in a separate set drained only by the TESTS-ONLY `idlePointerMaintenance` (the deterministic sync point under the frozen-clock preload). Full CI shard 1 passes locally with a clean exit — 280 tests including the four new cases and the held-load test.

bfollington:
Closure evidence for the review's outstanding condition (its exact words: "No approve verdict until disposal stops awaiting stalled probes"):

- `8dbe4012e` delivers precisely that split: `dispose()` awaits only `settlePointerCommits()` (bounded local roll-forward commits); watcher pattern loads are never dispose-awaited (lifecycle-epoch guards, as originally designed) and live in a separate set drained only by the tests-only `idlePointerMaintenance()`.
- The proving witness — reload-rehydration-safety's "dispose invalidates a held pattern hot-swap load", which holds a load open across dispose — passes with a clean process exit, alongside the full CI shard 1 file set (280 tests locally).
- CI on this head: 48 pass / 0 fail.
- All four prior findings remain closed with their regression tests (CFC-disabled two-runtime scenario; keyless hand-built pattern; key-reclaim guard; frozen-clock-safe synchronization).

The formal codex re-verification run has died mid-flight twice (harness issue, not findings); if it completes, its verdict will be appended.

PR #5154, opened 2026-07-29
Constrain pattern binding inputs to FabricExecValue

danfuzz:
## Summary

- add `FabricExecValue` and its recursive function, array, and plain-object types
- make `Pattern` and `Module` structurally conform to the execution-value graph
- constrain `unwrapOneLevelAndBindToDoc()` to execution-compatible inputs
- document that `Node` is intentionally structurally compatible without adding an index signature that would weaken typo detection
- keep the compile-time execution constraint out of authored JSON Schema generation

## What `FabricExecValue` represents

`FabricValue` describes values supported by the durable Fabric data model. `FabricExecValue` describes the slightly wider values allowed in an in-memory execution graph: any `FabricValue`, a callable function leaf, a read-only array of execution values, or a read-only plain object whose string-keyed properties are themselves execution values.

The type is recursive so complete `Pattern` / `Node` / `Module` graphs can satisfy one common constraint. Functions are the intentional addition over durable Fabric data; an execution graph containing them is not necessarily serializable or durable. The type models the graph values the runner is intended to traverse, but it does not attempt to encode runtime prototype or property-enumerability guarantees in TypeScript.

## Result-type boundary

`Runner.updateResultProjection<R>()` deliberately leaves `R` unconstrained. `R` is the author-facing result type, while `Pattern.result` has already been erased to `JSONValue`; constraining `R` to `FabricExecValue` would narrow the public runner API without restoring that lost relationship. It also would not eliminate an assertion: TypeScript still could not prove that a `JSONValue` is the caller-specific subtype `R`.

The call therefore passes the actual `pattern.result` into `unwrapOneLevelAndBindToDoc()`, allowing the new `FabricExecValue` input constraint to check the real execution value, and restores its association with `Cell<R>` at the result-cell boundary with `as R`.

## Schema-generator review

@ubik2 @mathpirate: please take a close look at the `ObjectFormatter` exception in this PR.

Having `Pattern` and `Module` extend `FabricExecPlainObject` introduces an inherited string index signature. That signature is a TypeScript invariant: every property added to those execution-graph interfaces must remain `FabricExecValue`-compatible. It is not intended to declare authored pattern data as accepting arbitrary `additionalProperties`.

Without the formatter exception, schema generation follows `PatternToolResult.pattern` into that inherited index signature, expands `FabricExecValue`, reaches the `symbol` arm of `FabricValue`, and fails with `No formatter found for type: symbol`. The SES test `compiles and evaluates a patternTool whose pattern is hoisted to module scope` reproduces the regression.

The implementation suppresses the inherited index only for interfaces directly based on `FabricExecPlainObject`. Explicit properties and ordinary authored index signatures retain their existing behavior. Please evaluate whether this targeted exception is the right schema-layer boundary for distinguishing a compile-time execution invariant from an authored `additionalProperties` contract.

## Validation

- `deno task check`
- API tests: 21 passed
- schema-generator tests: 54 tests / 284 steps passed
- runner SES test: 29 steps passed
- isolated fuse-exec runner test passed
- `pattern-binding.bench.ts` passed at 10, 30, 100, and 300 aliases
- formatting and lint passed for changed files

---
## Summary by cubic
Constrain pattern binding inputs to a new `FabricExecValue` type and align `Pattern`/`Module` with the execution graph. Also prevent the schema generator from leaking internal index signatures into authored JSON Schema.

- **Refactors**
  - Added `FabricExecValue` (`FabricExecFunction`, `FabricExecArray`, `FabricExecPlainObject`) for in-memory execution graphs.
  - Made `Pattern` and `Module` extend `FabricExecPlainObject`.
  - Restricted `unwrapOneLevelAndBindToDoc<T>` to `T extends FabricExecValue`; `Runner` now validates the actual execution value and then casts back to `R`.
  - Documented that `Node` remains structurally compatible without adding a string index signature.

- **Bug Fixes**
  - In `schema-generator`, `ObjectFormatter` now ignores the inherited index signature when a type extends `FabricExecPlainObject`, avoiding unintended `additionalProperties` and symbol-type formatter errors.

<sup>Written for commit 8ecedbf87c72df7315ab7ba1b2ca767db10afa0f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5154?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ubik2:
I think this exclusion of `additionalProperties` on classes that extend `FabricExecPlainObject` is ok.
We could probably be more narrow by having `FabricExecFunction` get mapped to `{ type: "unknown" }`, but I think that gets away from the intent. You aren't trying to say what data should be retrieved/monitored for a Pattern. Prior to making `Pattern` extend `FabricExecPlainObject`, it would also have excluded the `additionalProperties`, so we still wouldn't have retrieved/monitored them.

With this new rule, when accessing a `Pattern`, we will not reject a `Pattern` that has a property that does not match `FabricExecValue`. I think that's probably ok for this, and can't think of anything that would survive serialization that doesn't also match `FabricExecValue`. It does make me think that perhaps a JSDoc or decoration might be better for triggering this, but it's probably overkill with only one use case. We could also do something like what we do for `Default`, where the template is meaningless to TypeScript, but meaningful for our schema generation.

I should mention something like `Pattern & Record<string, string>` will also end up retrievable without the restriction, since that would have been an `allOf: [{$ref: FabricExecValue}, {type: "string"}]`. These were already unhandled by intersection-formatter.ts, though we did include a `$comment` in the output schema indicating the limitation.

PR #5173, opened 2026-07-29
fix(storage): keep the first accepted space host route

Hixie:
A runtime-learned per-space host could replace an earlier accepted hint before the space opened. A late site-table update could therefore redirect storage and compute after an IPC hint had already been accepted.

Keep the first configured or accepted route fixed for the manager session. Reject conflicting registrations both before and after a connection opens. Validate shared space routes as HTTP or HTTPS at both runtime and storage boundaries. Preserve the broader protocol set for the storage-only default memory host.

During site-table hydration, select the final usable entry for each space. Do not let a malformed duplicate hide an earlier usable route. Report registration verdicts so callers can stop before mounting on a conflict.

Add storage, runtime, and worker regression tests for both route orderings, duplicate entries, invalid schemes, and WebSocket-only defaults. Replace timer-based watcher synchronization with an event-driven assertion. Update the lifecycle and import plans to mark the conflict guard implemented.

---
## Summary by cubic
Fix per‑space host routing so the first configured or accepted HTTP(S) route stays fixed for the session. Treat the default host as provisional until the first hint or a stateful op; on a late hint, cancel pending work, switch hosts, replay reads, and harden memory client connect/mount cancellation with tests.

- **Bug Fixes**
  - Keep the first route per space fixed; return verdicts so callers can stop on conflicts before or after open.
  - Make default‑host providers provisional until a host hint or any stateful op issues; on a winning hint, replace the replica and replay reads/schemas, moving overlapping read‑only work and sync barriers.
  - Reject transactions whose read‑basis replica just changed (including cross‑space writes); once any op reaches its host, keep that route authoritative.
  - Validate shared routes as HTTP/HTTPS at runtime and storage; preserve WS/WSS only for the default `memoryHost`. Site‑table hydration selects the last valid HTTP(S) per space and won’t override an accepted route; add connect/mount cancellation guards with targeted tests.

- **Migration**
  - Call `registerSpaceHost(space, host)` before mounting and proceed only when it returns true; use HTTP/HTTPS hosts (optional: pre‑validate with `@commonfabric/runner` `normalizeSpaceHost`).
  - Unseeded spaces opened through the default host are provisional until the first hint or a stateful write; a late accepted hint will switch hosts and replay prior reads.

<sup>Written for commit 5d57b601e9b03ca6c0cffbfb4d6e109ca9063979. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5173?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/pattern-imports/README.md
+++ b/docs/specs/pattern-imports/README.md
@@ -102,7 +102,7 @@ general piece origin model is described in `../piece-source-lifecycle.md`.
 | Slug cells: generic **redirect link to any cell** (`setSlugLink` is target-agnostic; only `resolvePieceAddress` layers a "must be a piece" check) | `packages/piece/src/slugs.ts` | Slugs can name pieces *or* patterns today, mechanically |
 | Slug ids: `hashOf({causal:{space, slug}})`; slug grammar `[a-z0-9]+(-[a-z0-9]+)*`, ≤80 chars; **`isSlugAddress(t) = !t.includes(":")`** | `packages/runner/src/slugs.ts` | The existing slug-vs-URI discriminator the grammar reuses |
 | `loadPatternByIdentity(entryIdentity, symbol, space)` | `packages/runner/src/pattern-manager.ts` | Existing by-identity load path the resolver builds on |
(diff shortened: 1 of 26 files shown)


seefeldb:
Race that could happen is that loom creates a space, writes links to docs in it into an existing, opened pattern, and then the runtime tries to open that before loom registers the hint. in this case the first connection will (should!) fail, so if that is recovered with a late hint, that's fine.

Hixie:
> Race that could happen is that loom creates a space, writes links to docs in it into an existing, opened pattern, and then the runtime tries to open that before loom registers the hint. in this case the first connection will (should!) fail, so if that is recovered with a late hint, that's fine.

https://github.com/commontoolsinc/labs/pull/5173/changes/2761f920248f9b9549e43be7ac36c78ae1510910 should address this

PR #5182, opened 2026-07-29
fix(cli): stop forced-stream fallback from inventing verbs

bfollington:
## Problem

Against the live Topics board, `cf piece verbs` reported eleven handlers even though the pattern exposes only three. Ordinary result fields such as `$NAME`, `$UI`, `topics`, and `topicCount` were presented as callable, and the same fallback could let `cf piece call` attempt to dispatch an ordinary data field.

The legacy fallback imposed `asCell: ["stream"]` on each candidate and then asked `isHandlerCell()` whether that forced view was a stream. `Cell.isStream()` accepts that schema marker, so the check proved its own assumption and classified every candidate as a handler.

## Fix

- Check the candidate on the unmodified piece cell for independent stream evidence first.
- Build the forced-stream view only after that proof, retaining it solely for dispatch compatibility with legacy/link-derived handlers.
- Add regressions showing that a genuine fallback handler is still listed and callable, while an ordinary numeric field is neither listed nor dispatched.

## Verification

- `deno test -A test/piece-verbs.test.ts test/piece-call.test.ts`: 7 passed, 43 steps
- `deno fmt --check`, `deno lint`, `deno check`, and `git diff --check`: pass
- Read-only query against the reported Topics instance now returns exactly `addTopic`, `setMyName`, and `submitTopic`; no event was sent and no state was changed
- Full CLI suite: 1,643 passed; two untouched color-output assertions fail in this environment and also fail when run in isolation

## Review focus

Is the unmodified piece-cell `isStream()` result the intended independent source of truth for this legacy fallback? It recognizes the legitimate stream representations without allowing the dispatch-only schema cast to become evidence.

---
## Summary by cubic
Fixes the CLI forced-stream fallback so `cf piece verbs` and `cf piece call` only treat real stream handlers as callables, preventing ordinary data fields from being listed or dispatched.

- **Bug Fixes**
  - Check the unmodified piece cell for stream evidence before building any forced-stream view.
  - Build the forced-stream view only after proof, and use it only for dispatch (not as detection evidence).
  - Update callable listing and resolution to skip data fields like `topicCount` that previously appeared as handlers.
  - Add regression tests to ensure a real fallback handler is listed/callable and a numeric field is neither listed nor dispatched.

<sup>Written for commit d156e67710e2cb86628cceef08ce7f07e5e57e9a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5182?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cli/lib/piece.ts
+++ b/packages/cli/lib/piece.ts
@@ -1333,6 +1333,36 @@ async function tryResolvePieceCallableAt(
   };
 }
 
+/**
+ * Return the forced-stream view used to dispatch a legacy handler only when
+ * the same cell, without that imposed schema, is independently recognizable
+ * as a stream.
+ *
+ * The order is load-bearing: testing the forced view itself is circular,
+ * because Cell.isStream() accepts the `asCell: ["stream"]` marker that this
+ * function adds. That would classify every ordinary result property as a
+ * handler.
+ */
+function forcedStreamCellForHandler(
+  pieceCell: any,
+  callableName: string,
+): any | null {
+  if (
+    typeof pieceCell?.key !== "function" ||
+    !isHandlerCell(pieceCell.key(callableName))
+  ) {
+    return null;
+  }
+
+  return pieceCell.asSchema({
+    type: "object",
+    properties: {
+      [callableName]: { asCell: ["stream"] },
+    },
+    required: [callableName],
+  }).key(callableName);
+}
+
 async function tryResolvePieceHandler(
   piece: any,
   manager: any,
@@ -1344,26 +1374,17 @@ async function tryResolvePieceHandler(
     return null;
   }
 
-  const streamRoot = pieceCell.asSchema({
-    type: "object",
-    properties: {
-      [callableName]: { asCell: ["stream"] },
-    },
-    required: [callableName],
-  });
-  const streamCell = streamRoot.key(callableName);
-  if (!isHandlerCell(streamCell)) {
+  const streamCell = forcedStreamCellForHandler(pieceCell, callableName);
+  if (!streamCell) {
     return null;
   }
 
-  // Dispatch through the cell whose stream-ness this path just proved, not a
-  // second cell built by reading the schema back from links. Both address the
-  // same target — `getResult` is the identity on the piece cell — so they
-  // differ only in schema, and a link-derived schema is exactly what defeated
-  // the ordinary detection paths above. Sending on that cell takes `.set()`'s
-  // non-stream branch (`packages/runner/src/cell.ts:1316`) and fails with
-  // "Transaction required for .set()" instead of queueing the event, so a verb
-  // this path lists is a verb that could not be called.
+  // Dispatch through the forced view, not a second cell built by reading the
+  // schema back from links. Both address the same target — `getResult` is the
+  // identity on the piece cell — but the link-derived cell is exactly what
+  // defeated the ordinary detection paths above. Sending on that cell takes
+  // `.set()`'s non-stream branch (`packages/runner/src/cell.ts:1316`) and fails
+  // with "Transaction required for .set()" instead of queueing the event.
   const rootCell = await piece.result.getCell();
   const linkDerivedCell = rootCell.key(callableName).asSchemaFromLinks();
   return {
@@ -1617,12 +1638,7 @@ export async function listPieceCallables(
     }
     for (const name of rejected) {
       if (listings.has(name)) continue;
-      const streamRoot = pieceCell.asSchema({
-        type: "object",
-        properties: { [name]: { asCell: ["stream"] } },
-        required: [name],
-      });
-      if (!isHandlerCell(streamRoot.key(name))) continue;
+      if (!forcedStreamCellForHandler(pieceCell, name)) continue;
       const callableCell = resultRoot.key(name).asSchemaFromLinks();
       const spec = callableCommandSpec(callableCell, "handler");
       listings.set(name, {
--- a/packages/cli/test/piece-call.test.ts
+++ b/packages/cli/test/piece-call.test.ts
@@ -1143,16 +1143,31 @@ describe("forced-stream fallback dispatch", () => {
       key: () => emptyChild,
     };
 
-    const rootValue = { hiddenPing: {} };
+    const dataCell = {
+      schema: { type: "number" } as JSONSchema,
+      get: () => 3,
+      getRaw: () => 3,
+      asSchemaFromLinks: () => dataCell,
+      key: () => emptyChild,
+    };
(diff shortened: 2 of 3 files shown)


mpsalisbury:
@bfollington — this draft is more relevant now than when you opened it, and I would rather ask what you want done with it than decide for you. Backing evidence first, question at the end.

## Your diagnosis was two weeks ahead of ours

`forcedStreamCellForHandler`'s doc comment states the circularity exactly:

> testing the forced view itself is circular, because `Cell.isStream()` accepts the `asCell: ["stream"]` marker that this function adds. That would classify every ordinary result property as a handler.

That is the same root cause #5662 wrote up on 2026-08-11, arrived at independently and two weeks later. Nobody opened this draft in between, and we then produced two more PRs against the same defect. The rediscovery is on us.

## What has changed underneath it

#5683 landed (`ccaa2c99b`) and fixed the **listing** half by classifying on definite stored signals instead of the cast. So this PR's `listPieceCallables` hunk is superseded and would now conflict.

The **dispatcher** half is untouched on main and is filed as #5689. That is what your `tryResolvePieceHandler` change addresses, and it is the only thing aimed at it.

## Measured: your guard fixes it

Applying just the dispatcher change to current main, against a live toolshed with `packages/cli/integration/pattern/verb-results.tsx` (verbs: `createNote`, `setLabel`, `touch`):

```
label              exit=1   →  Callable "label" not found on piece fid1:NuNJ…
totallyMadeUpVerb  exit=1
setLabel           exit=0   →  settled, result returned
```

Today on main those first two exit 1 with no diagnostic at all — just the invocation header and a scheduler WARN dumping `eventLink` JSON. Your guard produces a real message, stops the junk event being queued against the piece, and leaves real verbs alone. All three of #5689's complaints.

## Where it does not currently pass, and the question that decides it

`packages/cli/test/piece-call.test.ts` → `forced-stream fallback dispatch` fails both steps under the guard.

I went in expecting a weak double and found something more substantive. The harness defines its subject as:

> A verb that defeats ordinary detection — plain object value, no `$stream` marker, no stream schema — but that the forced stream cast proves is a handler. This is the only case `tryResolvePieceHandler` exists for.

So the guard refuses, by construction, the exact case the fallback exists for. That is a semantic conflict, not a stale assertion. (Separately, the piece-cell double has no `key` method, so the guard short-circuits before reaching the interesting part — but fixing that would not save the test.)

Which leaves one question: **can a real piece carry a handler that is callable but has neither a stream schema marker nor a `{$stream: true}` value?**

- If **no**, that harness describes a case that cannot occur, and the right change is larger than your diff: delete the fallback and its tests rather than guard them.
- If **yes**, the guard as written makes a real handler uncallable, and it needs to preserve that case rather than refuse it.

#5683 settled the equivalent question for the *listing* by weighing costs — listing every data field costs more than missing a marker-less handler. That answer does not transfer here: a missing listing row still leaves the verb callable by name, whereas refusing dispatch has no workaround. So dispatch needs its own answer, and both sides currently answer through hand-written doubles rather than a real piece.

## What I would suggest, and the ask

Settle it with a real compiled fixture rather than a double, then either delete the fallback or land the guard accordingly. #5689 carries the reproduction and the `#5232` history if that is useful.

Your call on this PR, and genuinely yours — three options, no wrong answer:

1. **You rebase it** onto `ccaa2c99b`, keeping the dispatcher hunk and dropping the listing one.
2. **Hand it over** and one of us picks it up under #5689, crediting this PR.
3. **Close it** in favour of #5689, which already documents the approach and the measurements.

Not closing it from our side — it is your draft and your diagnosis got there first.

bfollington:
Closing — the problem this was opened for was fixed independently, and what remained does not justify the PR.

## What this set out to fix

Against the live Topics board, `cf piece verbs` reported eleven handlers for a pattern exposing three: `$NAME`, `$UI`, `topics`, and `topicCount` were listed as callable. The cause was circular — the fallback imposed `asCell: ["stream"]` on each candidate and then asked `isHandlerCell()` whether that forced view was a stream, so the check proved its own assumption.

## Why it is no longer needed

The listing half — the actual harm, since an untrustworthy verb surface is what leads a caller to attempt a data field — was fixed by ccaa2c99b5 ("fix(cli): the verb listing classifies on stored signals, not a cast", #5683) and extended by 5e6ae92b2c (#5794). `listCallablesForLoadedPiece` now classifies on the two definite stored signals via `detectCallableKind`, and is stricter than this PR proposed. Verified against a live piece: the listing is correct.

The dispatch half remained: `probeForcedStreamCell` still accepts any name. But verifying it end-to-end against a live piece on a scratch space showed the practical consequence is smaller than it looked:

- `cf call <piece> topicCount '3'` exits 1. The scheduler's drop detector (`packages/runner/src/scheduler/events.ts:541`) catches the undrained event: `Event dropped: no handler registered ... after starting its piece`.
- Nothing is mutated. Space SQLite counters were identical before and after (commit 268, revision 1402, head 1159); a genuine handler call moves all three, so the measurement was sensitive.
- The name is unreachable through the verb listing, so it is only hit by typing a known data field by hand.

So the remaining defect is error quality — a confusing scheduler drop rather than a clear `Callable "..." not found` — on a path no listing points at.

## Recorded findings

A rebase onto main was completed and its `packages/cli` suite passed (1707/1707), so this closes on judgement, not on rot.

Two things worth keeping:

1. **`probeForcedStreamCell` resolves nothing real.** Instrumenting the resolution chain across 15 verbs in `dice.tsx`, `shopping-list.tsx`, `record.tsx`, `occurrence-tracker.tsx` and purpose-built probes, path 3 was never reached — every genuine handler resolved on path 1 (`tryResolvePieceCallableAt`). Attempts to construct a handler that needs path 3 all failed: a stream-ness masked by the declared result type is recovered by `asSchemaFromLinks` from the stored link; a handler omitted from the result type is still listed and dispatched; and one laundered through an `any`-returning function becomes a `computed:` node the runtime drops on main too. That makes path 3 a candidate for deletion rather than repair — a cleanup with its own argument, not a bug fix.

2. **A doc comment on main is wrong about its own mechanism.** `probeForcedStreamCell`'s comment says a wrong cast "fails harmlessly against a value with no handler behind it," crediting the cast. The cast does no such thing; the scheduler's drop detector does. The outcome is harmless, the stated reason is not.

If the error wording becomes a real irritant, this is a small independent change and the evidence above is the justification for it.

PR #5191, opened 2026-07-29
Reject unrepresentable keys on plain objects

danfuzz:
`FabricPlainObject` is keyed by `string`, but nothing checked that. A symbol-keyed property on a plain object was accepted by every membership check and then silently dropped on the way to storage. Arrays have refused the analogous case for a while; objects now match.

Companion to #5167, which did the array half.

## What was unguarded

- `isFabricValueLayer()` — the plain-object branch inspected only the prototype
- `isFabricValue()` / `isFabricCompatible()` — walked with `Object.keys()`, which cannot see symbols by construction
- `shallowFabricFromNativeValue()` — the `NATIVE_TAGS.Object` arm had no key validation at all, hence the silent drop

The new `isPlainObjectWithOnlyEnumerableStringKeys()` (in a new `utils/objects.ts`, paralleling `utils/arrays.ts`) rejects symbol keys and non-enumerable string keys alike — neither has any representation as a property *name*. One comparison settles both, since `Reflect.ownKeys()` is always a superset of `Object.keys()`:

```js
Reflect.ownKeys(value).length === Object.keys(value).length
```

Note the object rule rejects non-enumerable *string* keys where the array rule tolerates non-enumerable *index* keys. Not an inconsistency: an index is content whatever its enumerability — `JSON.stringify` emits it — while a non-enumerable named key is dropped by every serialization there is.

## Two leaks this exposed, one of them pre-existing

Tightening turned 24 tests red across `runner` and `runtime-client`. Both causes were real, and the second is the more interesting:

**1. `[ID]` directives reaching conversion (23 failures).** `recursivelyAddIDIfNeeded()` calls `shallowFabricFromNativeValue()` mid-walk for type dispatch, while an `[ID]`/`[ID_FIELD]` directive may still be attached — either author-supplied, or added by that same function a few lines below. Those are instructions for `diffAndUpdate()`, which destructures `[ID]` out and stores only the rest, so the conversion functions are right to refuse an object bearing one. The premature call is the defect; a directive-bearing object now skips it and is walked directly.

**2. The back-to-cell annotation on objects (1 failure) — a pre-existing bug.** `schema.ts` hangs a non-enumerable `toCell` symbol on plain objects *and* arrays, but `convertCellsToLinks()` cleaned only arrays. So annotated objects have been reaching the conversion boundary for as long as the array case has been handled. It produced no symptom because the object side had no key validation — the annotation was silently dropped rather than rejected. `d98d43d53` fixes it and stands alone as a bug fix; it is deliberately ordered before the `[ID]` commit so it can be taken separately.

That is the argument for the strict rule as an *instrument*: it converted an invisible leak into a test failure. The invariant "no live runner metadata crosses into `data-model`" was only accidentally true, and is now enforced at both ends — `runner` strips or defers, `data-model` refuses.

## Behavior change

Symbol-keyed and non-enumerable-keyed objects now throw `Cannot store object with non-string-keyed properties` at the conversion boundary instead of losing the property quietly. That is the "death before confusion" direction, but it is a new throw, so it is worth knowing it exists rather than discovering it.

Also adds `shallowCleanPlainObject()`, the counterpart to `shallowCleanArray()`, for a caller whose extra keys it knows are not content.

## Reviewing

Four commits, each independently type-checked with its targeted test passing:

1. `utils` — the predicate
2. `data-model` — strict rule, conversion rejection, `shallowCleanPlainObject()`
3. `runner` — object `toCell` strip (**pre-existing bug**, standalone)
4. `runner` — don't convert while an `[ID]` directive is live

Full workspace suite green on the merged tree under the pinned Deno 2.9.4.

---
## Summary by cubic
Plain objects must now use only enumerable string keys. Symbol or non-enumerable keys now throw “Cannot store object with non-string-keyed properties” instead of being silently dropped, aligning objects with arrays and preventing metadata leaks.

- **New Features**
  - Added `isPlainObjectWithOnlyEnumerableStringKeys()` in `@commonfabric/utils/objects`; used across type checks and conversion.
  - Added `shallowCleanPlainObject()` to intentionally drop non-content keys when needed.

- **Bug Fixes**
  - Strip the `toCell` annotation from plain objects in the runner, not just arrays.
  - Skip conversion while `[ID]` or `[ID_FIELD]` is attached during the ID walk to avoid false rejections.
  - Added direct tests for `isPlainObjectWithOnlyEnumerableStringKeys()` to cover all acceptance and rejection cases.
  - Documented the ID-walk bypass edge case: on directive-bearing objects, other symbol/non-enumerable keys are still dropped by the record walk; behavior unchanged.

<sup>Written for commit 39ba3086ea74c0314a1895e7f542174d0ffcb16c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5191?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


danfuzz:
Confirmed on mechanism, and not fixing it here — reasoning below, plus one correction to the framing.

**The mechanism is real.** `recursivelyAddIDIfNeeded()`'s record branch rebuilds from `Object.entries()` and then copies back only `[ID, ID_FIELD]`, so any other symbol key is indeed dropped, silently.

**It is not introduced by this bypass, though.** Before this PR the object conversion did no key validation at all — that hole is what the PR closes. So such an object already passed conversion unexamined and reached the same `Object.entries` walk with the same silent drop. The bypass leaves one narrow corner of a pre-existing hole uncovered rather than opening a new one, which matters for how urgent it is.

**Why we're leaving it:** `ID`/`ID_FIELD` are slated for removal in the next PR. They are unused in practice, so the triggering shape — an object carrying a directive *plus* another symbol or non-enumerable key — has no known producer. Your fix is small, but it adds a branch, and a branch wants a test, and that test would pin a shape that is about to be deleted. We would rather not write and cover code with a scheduled demolition date.

What we have done instead is record the gap at the bypass itself, including how to close it (require everything other than the directives to satisfy `isPlainObjectWithOnlyEnumerableStringKeys()`), so that if the directives outlive the comment the next reader finds a known issue rather than an invisible one. If the removal does not land, the fallout is handled here.

Thanks for catching it — the pre-existing-versus-introduced distinction was worth establishing either way, and it is now written down where it will be read.

— Claude (via Claude Code, on @danfuzz's session)

PR #5192, opened 2026-07-30
Make background-piece registration transactionally idempotent

danfuzz:
`setBGPiece()` read the pieces list and searched it for a matching (`space`, `pieceId`) pair **outside** the transaction it then wrote in, so two concurrent registrations of the same pair could both find no match and both add an entry.

That is not a hypothetical path. An OAuth callback fires on every (re)connection of an integration, and `setBGPiece()` is called from four handlers on the shared provider-agnostic path — `oauth2-common.handlers.ts:333,472` and `plaid-oauth.handlers.ts:209,553` — under a comment reading "Register for background updates".

## The fix

Move the lookup inside `editWithRetry()`'s callback. That callback is re-invoked with a fresh transaction on commit conflict (`runtime.ts:1609`), so the loser of a race re-reads and finds the entry the winner just added — check-then-act becomes one atomic step. The commit result is now checked as well, rather than the returned promise being dropped.

## Why the `[ID]` goes away

The pushed entry carried `` [ID]: `${space}/${pieceId}` ``. It was **compensating for the race**: with it, two competing pushes converged on one entity id, so identity absorbed what the missing transaction should have prevented.

Nothing else here needed it. Entries still become their own documents via automatic id assignment, and lookup is by scanning rather than by re-deriving the id. With the race closed it has no remaining job, so it is gone — and this happens to remove the only live non-test use of `ID`/`ID_FIELD` in the tree, which is convenient for their planned retirement but is a side effect rather than the point.

## Testing, and what it does not show

The one test this function had ran on hand-written fakes whose `editWithRetry()` was `(fn) => { fn({}) }` and whose `withTx()` was the identity function — so "inside the transaction" and "outside" were indistinguishable to it, no commit ever conflicted, and it passed on `main` while the race was live. Replaced with five tests driving the real function against a real `Runtime` over `StorageManager.emulate`.

**The concurrent-registration test pins the outcome, not the mechanism, and says so in the code.** In practice the two calls serialize, because each awaits `getBGPieces()` and its sync before touching the list. I verified this by running all five tests against the read-outside-the-transaction version: they pass there too. Provoking a real commit conflict would mean driving the read and write phases apart, which this function does not expose.

So the fix rests on inspection — reading outside the transaction you write in is wrong regardless — not on test evidence. The tests establish that the upsert behaves correctly and that this change does not regress it. I would rather say that plainly than let five green tests imply more than they show.

The remaining fake now returns a Result, so it no longer contradicts the contract of the function it stands in for.

## Verified by hand against a local server

Two identical `POST /api/integrations/bg` calls through the real endpoint, against real storage, produced **one** entry:

```
entries = 1
  pieceId:     fid1:aaaa…
  integration: claude-validation
  status:      "Re-initializing"   <- second call took the update branch
  disabledAt:  0
  createdAt:   …045                <- 19ms before updatedAt, so one entry was
  updatedAt:   …064                   created then updated, not two created
```

What this does not show: the two calls were sequential, so the commit-conflict path is still unexercised. The race fix rests on inspection; this confirms the upsert behaves and that dropping `[ID]` costs nothing observable.

### Reproducing it

The service space needs an ACL genesis before anything can write to it, and a cleared local environment has none. That bootstrap is the whole difficulty; the rest is two curl calls.

1. Start local dev servers with at least one OAuth provider credential set, or `/api/integrations/bg` is never mounted and returns 404. The gate is only truthiness, and the route itself never uses the values, so dummies are fine:

   ```
   GOOGLE_CLIENT_ID=boop123 GOOGLE_CLIENT_SECRET=zoop123 ./scripts/restart-local-dev.sh
   ```

2. Bootstrap the background-service space. Its identity is derivable — `Identity.fromPassphrase("common user").derive("toolshed-system")` is exactly `BG_SYSTEM_SPACE_ID` — and genesis happens on first open when the manager is given it as the space *authority*:

   ```ts
   const service = await Identity.fromPassphrase("implicit trust", { implementation: "noble" });
   const spaceIdentity = await (await Identity.fromPassphrase("common user", { implementation: "noble" }))
     .derive("toolshed-system");
   const storageManager = StorageManager.open({
     as: service,
     spaceIdentity,                       // authority for genesis only
     memoryHost: new URL("http://localhost:8000"),
   });
   // then any read, e.g. `getBGPieces({ runtime })`, triggers it
   ```

   **Do not open as the space identity itself.** `{ as: spaceIdentity }` takes the home-identity branch (`signer.did() === space`), which claims an owner-only ACL that locks out the service principal. After that `add-admin-piece` fails with `lacks READ`, genesis will not re-run against a space that already has one, and only clearing all spaces recovers it.

3. POST the same payload twice and read the list back — expect one entry with `status: "Re-initializing"`:

   ```
   curl -X POST http://localhost:8000/api/integrations/bg -H 'Content-Type: application/json' \
     -d '{"pieceId":"fid1:<54 chars>","space":"did:key:...","integration":"whatever"}'
   ```

   `pieceId` must be exactly 59 characters (`fid1:` plus 54) or validation rejects it.

---
## Summary by cubic
Make background-piece registration transactionally idempotent so concurrent (`space`, `pieceId`) registrations create one entry and re-enable existing ones. Adds test coverage for failed-commit handling.

- **Bug Fixes**
  - Do the lookup and upsert inside `editWithRetry()` so concurrent calls re-read on conflict and don’t duplicate.
  - Return whether a piece was added; throw on commit error; await `storageManager.synced()`; add a test that surfaces failed commits and ensures no entry is written.

- **Refactors**
  - Remove `[ID]` from entries; rely on automatic ids and scan-based lookup.
  - Replace fake-only test with runtime-backed tests using `Runtime` and `StorageManager.emulate` from `@commonfabric/runner`; update remaining fake to return a Result; minor test import cleanup.

<sup>Written for commit ae83375e96a416ec72faaf51c258760a4da4e1bf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5192?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seefeldb (approved):

danfuzz:
Cubic appears to be broken at the moment. All other jobs are green. I'm going to merge.

PR #5193, opened 2026-07-30
fix(hooks): restore the pre-commit and subagent-stop hooks, corrected

seefeldb:
Rebased onto main after #5210 removed the blocking hooks, so this is now the
**revert-plus-fix**: it re-adds `pre-commit.ts` and `subagent-stop.ts` in
corrected form, and only those. It is purely additive against main.

`block-git-amend.ts` is deliberately **not** restored. Removing it was a separate
decision and it stands: models routed around it with `git reset` plus a fresh
commit, which is strictly worse than the amend it refused — a reset discards the
original commit object, where `--amend` preserves the tree.

`stop-check.ts` and `post-edit-ts.ts` show no diff here: #5210 already carried
those two fixes, which were found during this work.

## Merge this only if the experiment says so

#5210's premise is that these hooks may no longer earn their place. The measure
is whether PRs start failing the `Check` job on fmt/lint/check over the next few
weeks. If they don't, leave this closed. If they do, this branch is what to
restore — though the better shape is still a PostToolUse hook reporting on
`git show --name-only HEAD` after a commit succeeds: no prediction, no shell
parsing, no blocking.

## What was wrong, as originally reported

Both hooks resolved their working directory from the session's checkout rather
than the tree they were inspecting. A commit to one worktree was checked against
— and blocked by — another branch's errors, with `--no-verify` the only way past.
`pre-commit` ran `deno fmt` rather than `--check`, so it silently reformatted a
checkout nobody was working in. `subagent-stop` checked the whole repo, so it
reported pre-existing branch failures to every subagent and sibling agents'
in-flight edits to whichever finished first.

## Measured, not assumed

All three candidate signals disagree, so the hooks were probed live before
anything was changed:

| Signal | PreToolUse | SubagentStop |
|---|---|---|
| `$CLAUDE_PROJECT_DIR` | session checkout, pinned | session checkout, pinned |
| payload `cwd` / `Deno.cwd()` | follows the Bash tool's *persisted* `cd` | **the parent session's dir** |
| `agent_transcript_path` → entry `cwd` | — | **the subagent's real worktree** |

## Eleven review rounds, 85 confirmed defects

Two automated reviewers and repeated adversarial subagent passes. Seven of those
defects were violations of the read-only contract — five introduced by the fix
for the previous one — and CI caught none of the 85. The commit messages record
each finding with its trigger condition; the ones worth knowing regardless of
whether this merges:

- `git status --porcelain` refreshes and **rewrites the index**;
  `--no-optional-locks` covers it.
- `git diff <commit>` also rewrites it, and `--no-optional-locks` does **not**
  cover it. Trigger: a racily-clean file — content matching the index with a
  differing mtime, i.e. edit a file and revert it. Invisible to `git write-tree`
  and to the index mtime; only the index bytes move.
- `git add --dry-run --edit` **stages the working tree**, because `--edit`
  applies its patch regardless and `GIT_EDITOR=true` is set. git accepts `--e`,
  `--ed`, `--edi` for it, so a blacklist cannot hold — the guard is a whitelist.
- `deno check` **rewrites `deno.lock`**, a tracked file, when a checked file
  imports something the lock lacks. Running before the user's command, that write
  landed in their commit.
- Quoted text is data: a commit message mentioning `-a`, `--no-verify` or
  `--work-tree` could sweep the tree, disable the hook, or turn it off entirely.
- git resolves any unambiguous option abbreviation, so exact-string matching
  mishandled `--dr`, `--inc`, `--mess` and `--no-ver` each in its own way.

## Coverage, which did not exist

`.claude/` was excluded from fmt and lint and absent from `tasks/check.sh`, so a
hook that failed to compile — blocking every commit in the repo — would have
shipped silently. `.claude/scripts` is now a workspace member with a test task
and is listed in `tasks/check.sh`.

24 tests: 14 over the command parser, 10 driving the real hook against a scratch
repository. The read-only contract is asserted on **every** hook invocation in
the suite — index bytes, working-tree status and `deno.lock` together — rather
than in tests that remember to check it. Four of the seven contract breaks would
have been caught by that assertion.

### DIFF
--- a/.claude/scripts/common/checks.ts
+++ b/.claude/scripts/common/checks.ts
@@ -1,0 +1,135 @@
+/**
+ * The fmt/lint/check battery, scoped to a file list and a worktree.
+ *
+ * Shared by the pre-commit and subagent-stop hooks so that the two agree on
+ * which files a check can meaningfully be pointed at. Every check here is
+ * read-only: a hook that reformats the tree it is inspecting turns a report
+ * into an edit, and an edit nobody asked for in a tree nobody is working in.
+ */
+
+/**
+ * Schema-generator fixture inputs use bare Cell/Stream/Writable types injected
+ * by a custom TypeScript compiler host (CELL_BRAND_PRELUDE in
+ * schema-generator/test/utils.ts), so `deno check` cannot resolve them: the
+ * prelude is a template string, not a file it can be pointed at.
+ *
+ * They are checked, just not here. `test/fixtures-runner.test.ts` batch
+ * type-checks every one of them through that same host on each run of the
+ * package's tests, on by default and disabled only with SKIP_INPUT_CHECK. That
+ * is the stricter of the two checks, because it is the compiler the generator
+ * itself uses. Skip them rather than teaching this hook a second, weaker way to
+ * check them — a copy of the prelude here would be a second source of truth
+ * free to drift from the one that matters.
+ */
+function isSchemaFixtureInput(path: string): boolean {
+  return path.includes("schema-generator/test/fixtures/") &&
+    path.endsWith(".input.ts");
+}
+
+/**
+ * The static type assets are what the pattern compiler serves to authored code
+ * as its standard library and type modules, not modules of this repo. `.d.ts`
+ * ends in `.ts`, so they reach this filter. Checking them declares a second
+ * standard library alongside Deno's own: `es2023.d.ts` redeclares `lib.es5`
+ * members with different types, and `jsx.d.ts` and `commonfabric.d.ts` fail on
+ * their own too. `deno.jsonc` keeps this directory out of fmt and lint, and
+ * `tasks/check.sh` never lists it.
+ *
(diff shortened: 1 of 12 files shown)


seefeldb:
@codex review

seefeldb:
@codex review

seefeldb:
@codex review

seefeldb:
@codex review

seefeldb:
Closing, not abandoning — the branch `claude/cool-shannon-e20869` stays and is enough to restore from.

It sits rebased on main as of `8fb3e70cc`, purely additive (+2763/−6), CI green. If #5210's experiment shows PRs starting to fail the `Check` job on fmt/lint/check, reopen this or branch from it. If the signal is not missed, nothing here needs to happen.

The measured findings are in the commit messages and stay true regardless of whether these hooks ever come back: `git diff <commit>` rewrites the index despite `--no-optional-locks` (racily-clean files only, and invisible to `git write-tree`), `git add --dry-run --edit` stages the working tree, and `deno check` rewrites the tracked `deno.lock`.

PR #5195, opened 2026-07-30
feat(cf-harness): report model usage and cache behavior

bfollington:
## Summary

Adds end-to-end model usage, cache, reasoning, and cost observability to
`cf-harness`, together with controls needed to run meaningful prompt-caching
experiments.

This work started from an operational observation: equivalent-looking work in
CF harness appeared to consume quota and money much faster than other agent
harnesses, but CF harness did not expose enough provider telemetry to determine
whether the difference came from cache misses, reasoning tokens, context growth,
output volume, or provider pricing.

### The harness was not obviously disabling the provider KV cache

OpenAI prompt caching is server-side. Requests become eligible once the prompt
is long enough, and cache hits require an exact shared prefix. Before this
change, a multi-turn CF harness run naturally retained a shared conversation
prefix, so implicit caching could already occur.

However, the harness had three important blind spots:

1. It discarded provider usage detail, including cached input, cache writes, and
   reasoning tokens.
2. It did not provide a stable prompt-cache affinity key, so reliable routing to
   the same cache shard could not be requested.
3. It did not expose reasoning effort or explicit cache breakpoints, making it
   difficult to control two major variables in a side-by-side experiment.

The result was not proven cache incompatibility; it was an inability to tell
whether caching was occurring or why a run was expensive.

OpenAI reference:
https://developers.openai.com/api/docs/guides/prompt-caching

### Dollar cost and subscription quota are different measurements

For an OpenAI-compatible API gateway, this change preserves provider-reported
cost when present and can estimate GPT-5.6 API cost from complete token/cache
detail. The estimate is deliberately labelled separately and is not an invoice.

The `openai-codex` provider uses a ChatGPT/Codex subscription. Its token
telemetry is still useful, but public API token prices are not treated as a
conversion for subscription quota consumption.

### Normalized usage accounting

Adds a provider-independent usage shape with:

- input tokens
- cached input tokens
- cache-write tokens
- output tokens
- reasoning tokens
- total tokens
- provider-reported USD cost
- separately labelled estimated USD cost

Both Responses API and Chat Completions usage payloads are normalized. Usage is
captured per model turn and aggregated across the direct run and completed
descendant runs. Already-incurred child usage is recorded before later parent
artifact persistence, so failure reports remain cost-complete.

### Operator visibility

- CLI completion output includes a single aggregate `usage:` line.
- The line includes cache-read percentage when possible.
- Batch result JSON contains aggregate usage.
- `run-report.json` contains direct usage, descendant-inclusive total usage, and
  per-turn model usage records.
- Interactive `turn_completed` events expose usage.

### Cache and reasoning controls

- Stable per-run cache affinity is passed to providers.
- Interactive sessions reuse a stable affinity key across user turns.
- `--reasoning-effort` / `CF_HARNESS_REASONING_EFFORT` forwards the selected
  provider reasoning effort.
- `--prompt-cache-mode implicit|explicit` /
  `CF_HARNESS_PROMPT_CACHE_MODE` selects GPT-5.6 compatible-gateway cache
  behavior.
- Explicit mode places the cache breakpoint on the stable first-user-message
  prefix, before appended assistant/tool messages.
- Unsupported explicit cache configuration is rejected before sending the
  request.

### Cost estimation safeguards

Public OpenAI GPT-5.6 estimates are produced only when:

- the model has a known price entry;
- input, output, cache-read, and cache-write counters are all present; and
- cache detail is internally consistent with total input.

This avoids presenting missing or malformed usage detail as a believable cost
number. When an estimate is withheld, the report records why; aggregate dollar
cost is emitted only when every included usage entry has a complete estimate.

## Validation

- `mise exec -- deno task test`
  - 573 passed
  - 0 failed
- GitHub `Coverage Check`: 2,669 uncovered `cf-harness` lines versus 2,759 on
  `main` (90 fewer).
- `mise exec -- deno check mod.ts src/main.ts src/interactive-chat-stdio.ts`
- `mise exec -- deno lint`
- `mise exec -- deno fmt .`
- `git diff --check`

Coverage includes Responses and Chat usage normalization, public price
estimation, inconsistent-data handling, provider request wiring, cache-mode
validation, reasoning effort, CLI rendering and parsing, batch output,
interactive affinity, run reports, and descendant aggregation.

## Live experiment

Run on 2026-07-30 through the real CF harness, Docker 27.4.0, the connected
`openai-codex` subscription, and `gpt-5.6-terra` at low reasoning effort.

To avoid disclosing repository contents to the external provider, the sandbox
workspace contained only three synthetic files. The task was deliberately
read-only and required three sequential `read_file` calls, producing four model
turns. CFC ran in observe mode for the cache-eligible case because the local
`read_file` path did not produce trusted mediation metadata; that policy warning
is separate from model caching.

### Case 1: prompt below the cache threshold

The first valid run used the short task prompt. Each turn remained below the
documented 1,024-token cache eligibility threshold:

| Turn | Input | Cached input | Output | Reasoning |
| ---: | ---: | ---: | ---: | ---: |
| 1 | 497 | 0 | 41 | 18 |
| 2 | 648 | 0 | 40 | 14 |
| 3 | 798 | 0 | 21 | 0 |
| 4 | 929 | 0 | 22 | 0 |
| **Total** | **2,872** | **0** | **124** | **32** |

This explains the zero-cache result without requiring a cache compatibility
failure: no individual request was eligible yet.

### Case 2: stable prefix above the cache threshold

The next run added inert synthetic padding to the otherwise identical task so
the first request crossed 1,024 input tokens:

| Turn | Input | Cached input | Output | Reasoning |
| ---: | ---: | ---: | ---: | ---: |
| 1 | 1,265 | 0 | 43 | 20 |
| 2 | 1,405 | 1,024 | 24 | 0 |
| 3 | 1,535 | 1,024 | 21 | 0 |
| 4 | 1,640 | 1,024 | 33 | 0 |
| **Total** | **5,845** | **3,072** | **121** | **20** |

The harness reported a **52.6% aggregate cache-read ratio**. Every continuation
turn received a 1,024-token cache hit. This confirms that CF harness is
compatible with the subscription backend's implicit prompt cache and that the
new affinity and telemetry paths work in a real multi-turn Docker run.

The subscription response reported zero cache-write tokens and no dollar cost.
That is preserved as provider telemetry rather than converted into subscription
quota. For scale only, applying the public GPT-5.6 Terra API token schedule to
these read/output counters would be approximately **$0.00952 with cache reads**
versus **$0.01643 if every input token were uncached**, about a 42% reduction.
Those figures are illustrative API equivalents, not a ChatGPT invoice or quota
conversion.

### Case 3: provider cache-mode compatibility

A controlled Codex request with `prompt_cache_options.mode = "implicit"` failed
with HTTP 400 before model execution. Removing only the cache-mode control while
keeping low reasoning effort succeeded. The ChatGPT/Codex backend already
accepts `prompt_cache_key`, but does not accept the API
`prompt_cache_options` control used by the compatible gateway.

The implementation was corrected based on this result:

- `openai-codex` now rejects `promptCacheMode` locally before resolving
  credentials or sending provider traffic;
- subscription runs retain stable affinity, reasoning control, implicit
  caching, and usage telemetry;
- implicit/explicit cache modes and explicit breakpoints remain available on
  the GPT-5.6 compatible API gateway;
- tests and live documentation now encode that provider distinction.

### Conclusion so far

The live evidence does **not** support the hypothesis that CF harness is broadly
incompatible with prompt/KV caching. It demonstrates working 1,024-token cache
hits once requests are eligible. Likely sources of surprising quota consumption
that can now be measured directly are:

- short runs or early turns below the cache threshold;
- uncached context growth beyond the reused prefix;
- reasoning and output tokens;
- descendant-agent work included in the aggregate;
- differences between subscription quota accounting and public API prices.

The new reports make those factors visible per turn and across the whole run,
which enables follow-up comparisons with Codex CLI using the same prompt, model,
reasoning effort, tools, and number of turns.

### Case 4: real compatible-gateway explicit caching

A two-turn synthetic run used the real staging compatible gateway,
`gpt-5.6-terra`, low reasoning effort, and explicit cache mode:

| Turn | Input | Cached input | Cache write | Output | Reasoning | Estimated USD |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 1,123 | 0 | 1,120 | 32 | 8 | 0.0039875 |
| 2 | 1,211 | 1,120 | 0 | 9 | 0 | 0.0006425 |
| **Total** | **2,334** | **1,120** | **1,120** | **41** | **8** | **0.0046300** |

The first turn wrote the explicit 1,120-token prefix and the continuation read
that same prefix. This confirms end to end that the real gateway accepts
`prompt_cache_options.mode`, the explicit content breakpoint, stable affinity,
and reasoning effort; emits both `cached_tokens` and `cache_write_tokens`; and
provides sufficient detail for the cost estimator.

The experiment record is preserved in this PR description and its review
discussion.


packages/cf-harness/src/model/usage.ts
```diff
@@ -0,0 +1,168 @@
...
+  const prices = OPENAI_MODEL_TOKEN_PRICES[model];
+  if (
+    prices === undefined || usage?.inputTokens === undefined ||
+    usage.cachedInputTokens === undefined ||
+    usage.cacheWriteTokens === undefined ||
+    usage.outputTokens === undefined
+  ) return undefined;
+  const uncachedInputTokens = usage.inputTokens - usage.cachedInputTokens -
+    usage.cacheWriteTokens;
+  if (uncachedInputTokens < 0) return undefined;
```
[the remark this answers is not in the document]
bfollington: Addressed in 166db1394. Cost estimates now require non-negative safe-integer token counters, `totalTokens === inputTokens + outputTokens` when total is reported, and `reasoningTokens <= outputTokens` when reasoning is reported. Added focused tests for conflicting totals, impossible reasoning, and invalid counts.

packages/cf-harness/src/prompt-loop.ts
```diff
@@ -2748,6 +2832,14 @@ export class CfHarnessPromptLoop {
     const childLoop = new CfHarnessPromptLoop({
       engine: childEngine,
       modelClient: this.modelClient,
+      cacheAffinityKey: childRunId,
+      ...(this.#promptCacheMode !== undefined &&
+          childModel.model.startsWith("gpt-5.6")
+        ? { promptCacheMode: this.#promptCacheMode }
+        : {}),
```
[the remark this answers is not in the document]
bfollington: Addressed in 166db1394. The configured cache mode is now forwarded to every child loop, including profile model overrides, so the existing provider/model validator rejects unsupported children instead of silently dropping the control. Added an end-to-end prompt-loop regression test covering the failed child and recovered parent.

bfollington (commented):
cf-review via Claude `claude-opus-5`, on behalf of @bfollington.

**Coverage:** deep-read every `src/` change and all 8 test files; docs read. Reviewed at head `877938883`.

**Checks run:** `deno test` on the 6 touched test files — **124 passed / 0 failed** · `deno check mod.ts src/main.ts src/interactive-chat-stdio.ts` clean · `deno lint` clean (104 files) · `deno fmt --check` clean.

**No blocking findings.** The code is correct as written, well-typed, tested, and the docs were updated in the same change — including commit 2 correctly walking back codex cache-mode support after the live experiment disproved it. Scope is coherent and stays inside the package.

Everything I flagged is one theme: a new *observability* contract being trusted slightly further than its evidence supports. That matters more than usual here, because not being misled by these numbers is the entire point of the change. Seven improvements are left as inline comments; the two most load-bearing are:

1. **`estimatedCostUsd` is gated on `input_tokens_details.cache_write_tokens`** — a field that appears nowhere else in the repo and isn't in OpenAI's documented usage shape. If the gateway doesn't emit it, the estimate never appears, and its absence looks identical to "prompt wasn't cache-eligible". The live experiment ran on `openai-codex`, which by design never estimates, so this path has no end-to-end evidence.
2. **`--reasoning-effort` is silently dropped on the Chat Completions path but still recorded in the run report** — a non-`gpt-` gateway model produces a report asserting a control the provider never received.

### ⚪ Nits

- `responses-protocol.ts:241` — `addFirstUserPromptCacheBreakpoint` deep-clones the whole input array to change one content block (O(transcript) per turn in explicit mode); copying just the item being modified is enough. It also casts content parts as `ResponsesInputItem[]`, which they aren't, and its `throw` lacks the provider label every other error in these clients carries.
- `client.ts:56` — `HarnessModelUsage` doesn't document that `cachedInputTokens ⊆ inputTokens` and `reasoningTokens ⊆ outputTokens`, while the CLI prints all six side by side as if they were additive. Two JSDoc lines close the trap.

### On cubic's `total_tokens` P2

I'd dismiss it. The estimate is computed entirely from the component counters; `total_tokens` never enters the arithmetic, so validating it would only reject otherwise-usable data.

### Questions

1. Was the gateway explicit-cache path exercised against the real gateway, or only against the fake? Case 3 proves codex rejects `prompt_cache_options`; the gateway's acceptance of `prompt_cache_options`, `ttl: "30m"`, and `prompt_cache_breakpoint` is currently asserted only in unit tests. If the gateway also 400s, `--prompt-cache-mode` has no working provider.
2. The "Live experiment" write-up is the most valuable artifact here and it currently exists only in a PR body. Per `docs/history/README.md` ("profiling reports, investigation findings") that's a `docs/history/` document with a metadata header — worth landing it in the tree?

### Possible follow-ups

- Validate `reasoningEffort` against `HarnessModelCatalogEntry.supportedReasoningEfforts`, and populate that field for the gateway instead of the current hardcoded `[]`.
- Record *applied* controls per model turn on `HarnessModelTurnUsage`, so the report distinguishes "requested" from "sent" for every provider/route combination rather than one flag at a time.

**Verdict:** Sound and well-documented; nothing blocking. Findings 1 and 2 are worth resolving before relying on these numbers for the Codex-CLI comparison, since both let the report claim more than the provider confirmed.

packages/cf-harness/src/model/usage.ts
```diff
@@ -0,0 +1,168 @@
...
+ * unless the response reports both cache reads and writes, since treating
+ * missing cache detail as zero would make a cache experiment misleading.
+ */
+export const estimateOpenAIModelUsageCostUsd = (
+  model: string,
+  usage: HarnessModelUsage | undefined,
+): number | undefined => {
+  const prices = OPENAI_MODEL_TOKEN_PRICES[model];
+  if (
+    prices === undefined || usage?.inputTokens === undefined ||
```
bfollington: **🟡 The estimate is gated on a field this repo has never observed.**

`estimatedCostUsd` is withheld unless `cacheWriteTokens` is present, which comes from `input_tokens_details.cache_write_tokens`. `git grep cache_write_tokens` outside cf-harness returns nothing, and it is not part of OpenAI's documented usage shape (which reports `cached_tokens` only). Same for `cost_usd`.

So if the gateway doesn't emit `cache_write_tokens`, `estimatedCostUsd` never appears at all — and its absence is indistinguishable from "the prompt wasn't cache-eligible". The live experiment ran on `openai-codex`, which by design never estimates, so this path has no end-to-end evidence.

**Fix:** keep the strict gate, but say *why* it was withheld — e.g. `estimateWithheldReason: "no-cache-detail" | "unknown-model" | "inconsistent-counters"` — so an operator can tell a missing provider field from an ineligible prompt.

**Related, same function:** the arithmetic on the next lines assumes `cacheWriteTokens ⊆ inputTokens`. Anthropic-style accounting reports cache *creation* additively; if the gateway follows that convention the estimate silently under-counts uncached input, and the `< 0` guard only trips at the extreme. Worth a comment pinning the subset assumption to the field's actual spec.

_verified by read + grep; provider behavior suspected._

packages/cf-harness/src/model/openai-compatible-gateway.ts
```diff
@@ -279,14 +275,29 @@ export class OpenAICompatibleGatewayModelClient implements HarnessModelClient {
...
+      ),
+      ...(request.promptCacheMode !== undefined
+        ? {
+          prompt_cache_options: {
+            mode: request.promptCacheMode,
+            ttl: "30m",
+          } as const,
+        }
+        : {}),
+      ...(request.reasoningEffort !== undefined
```
bfollington, line 338: **🟡 `reasoningEffort` is silently dropped on the Chat Completions path, but still reported as applied.**

`usesResponsesApi` is `nativeModelToolIds.length === 0 && model.startsWith("gpt-")`, so a non-`gpt-` gateway model (Gemini via the `web_search` profile, etc.) routes to `#completeViaChatCompletions` — which never sends `reasoning`. Meanwhile `createHarnessRunReport` records `reasoningEffort` unconditionally from the loop's config (prompt-loop.ts:1891). The run report then asserts a control the provider never received, which is the exact failure mode this PR exists to eliminate.

**Fix:** either assert unsupported the way `assertPromptCacheModeSupported` does, or only record it once a turn actually sent it.

**Adjacent:** `listModels` below hardcodes `supportedReasoningEfforts: []` for every gateway model, so the harness's own discovery surface now contradicts the new flag — and nothing validates the flag value against the catalog.

_verified by read._

packages/cf-harness/src/prompt-loop.ts
```diff
@@ -2748,6 +2832,14 @@ export class CfHarnessPromptLoop {
     const childLoop = new CfHarnessPromptLoop({
       engine: childEngine,
       modelClient: this.modelClient,
+      cacheAffinityKey: childRunId,
+      ...(this.#promptCacheMode !== undefined &&
+          childModel.model.startsWith("gpt-5.6")
```
bfollington: **🟡 This re-derives the support rule and drops the mode silently.**

`childModel.model.startsWith("gpt-5.6")` duplicates the rule inside `assertPromptCacheModeSupported`. Extract a `promptCacheModeSupported(model)` predicate in `responses-protocol.ts` and use it at both sites so they can't drift.

(cubic's bot suggests forwarding unconditionally so the child throws — I'd keep your skip-don't-fail choice, but make it *visible* rather than only inferable from an absent report field.)

_verified by read._

packages/cf-harness/src/model/usage.ts
```diff
@@ -0,0 +1,168 @@
...
+  "cacheWriteTokens",
+  "outputTokens",
+  "reasoningTokens",
+  "totalTokens",
+  "costUsd",
+  "estimatedCostUsd",
+] as const satisfies readonly (keyof HarnessModelUsage)[];
+
+/** Sums all reported fields while preserving "not reported" as absent. */
+export const sumHarnessModelUsage = (
```
bfollington, line 77: **🟡 This can produce a partial cost that reads as the whole run's.**

Each field is summed over only the entries that report it. So if some turns withhold `estimatedCostUsd` — or a descendant runs a model absent from the price table — the aggregate still carries an `estimatedCostUsd` that looks like the run total.

That is precisely the "missing cache detail would make a cache experiment misleading" hazard the docstring at L129 guards against per turn, reintroduced at the aggregate level.

**Fix:** omit `costUsd`/`estimatedCostUsd` from the sum unless *every* entry supplies them. Token counters can keep summing as they do. No test covers the mixed case today.

_verified by read._

packages/cf-harness/src/prompt-loop.ts
```diff
@@ -1638,6 +1653,10 @@ export class CfHarnessPromptLoop {
   readonly #parentToolAllowanceMode: HarnessParentToolAllowance;
   readonly #allowedSubagentProfiles: ReadonlySet<HarnessSubagentProfile>;
   readonly #browserAccess?: HarnessBrowserAccessLease;
+  readonly #cacheAffinityKey?: string;
+  readonly #promptCacheMode?: "implicit" | "explicit";
+  readonly #reasoningEffort?: string;
+  readonly #descendantUsage: HarnessModelUsage[] = [];
```
bfollington: **🟡 Run-scoped state living on the instance.**

Every sibling accumulator (`toolActivity`, `gatewayAttempts`, `modelAttempts`, `modelUsage`) is a local in `runTranscript`; this one is on `this` and gets reset with `.length = 0` at run start (L1858). Two concurrent `runTranscript` calls on one loop instance would clear and cross-attribute each other's descendant usage.

No current caller does that — the interactive service builds a fresh loop per turn, and the CLI one per run — so it's latent rather than a bug. But `#invokeDelegateTaskTool` already takes an options bag and returns a result object: returning the child usage from it and pushing into a local removes the shared-state class entirely and matches how everything else in this method works.

_verified by read; latent._

packages/cf-harness/src/model/usage.ts
```diff
@@ -0,0 +1,168 @@
...
+  if (
+    prices === undefined || usage?.inputTokens === undefined ||
+    usage.cachedInputTokens === undefined ||
+    usage.cacheWriteTokens === undefined ||
+    usage.outputTokens === undefined
+  ) return undefined;
+  const uncachedInputTokens = usage.inputTokens - usage.cachedInputTokens -
+    usage.cacheWriteTokens;
+  if (uncachedInputTokens < 0) return undefined;
+  const longContextInputMultiplier = usage.inputTokens > 272_000 ? 2 : 1;
```
bfollington: **🟡 Untested magic numbers.** The `272_000` threshold and the 2×/1.5× multipliers have no test and no source note — the cited URL above covers the per-million table, not the long-context tiering. One case at 272,001 input tokens pins both multipliers.

packages/cf-harness/src/contracts/interactive-chat.ts
```diff
@@ -340,9 +340,13 @@ export interface HarnessChatSubagentSummary {
 
 export interface HarnessChatGatewayUsage {
   inputTokens?: number;
+  cachedInputTokens?: number;
```
bfollington: **🟡 `HarnessChatGatewayUsage` now mirrors `HarnessModelUsage` field-for-field.** The `turn_completed` event assigns the latter into the former and it type-checks structurally, which is exactly how two shapes drift apart later. `type HarnessChatGatewayUsage = HarnessModelUsage` (or import it directly) keeps them one thing.

packages/cf-harness/src/cli.ts
```diff
@@ -390,6 +397,8 @@ Options:
   --no-skill-catalog            Disable automatic skill catalog disclosure
   --model <name>                Model name (default: ${DEFAULT_MODEL})
   --model-provider <provider>   openai-compatible-gateway | openai-codex
+  --reasoning-effort <effort>  Provider reasoning effort (for example low, medium, high)
```
bfollington: ⚪ Both new `--help` descriptions start one column left of every neighbour — `--reasoning-effort <effort>` needs 3 spaces and `--prompt-cache-mode <mode>` needs 4 to land on the shared column.

bfollington:
cf-review via Claude `claude-opus-5`, on behalf of @bfollington — **re-anchor for `166db1394`.**

The branch was force-pushed while I was reviewing, so the review above is anchored to `877938883` and predates `fix(cf-harness): validate usage and cache overrides`. Re-checked the delta; corrections:

**Retracted — finding 3 (subagent cache-mode gate).** `166db1394` removes the `childModel.model.startsWith("gpt-5.6")` condition, so the duplicated support rule is gone and `assertPromptCacheModeSupported` is the single decision point. `CfHarnessPromptLoop preserves cache-mode validation for child model overrides` covers the failed-child/recovered-parent path. Nothing left to do here; ignore that inline comment.

**Still applies at `166db1394`** (line numbers shifted — new anchors):

- **1 — estimate gated on an unobserved field**: `usage.ts:131` still returns `undefined` when `cacheWriteTokens` is absent, and `cache_write_tokens` still appears nowhere else in the repo. The new validation actually *strengthens* the case for surfacing a reason: there are now four more silent-withhold paths (non-safe-integer, `totalTokens !== inputTokens + outputTokens`, `reasoningTokens > outputTokens`, negative uncached), all of which produce the same indistinguishable "no estimate". A gateway that computes `total_tokens` on any other convention now silently loses cost reporting with no trace. `estimateWithheldReason` pays for itself at five call sites.
- **1b, partly addressed**: `reasoningTokens > outputTokens` now encodes the reasoning-subset assumption in code — good. The `cachedInputTokens`/`cacheWriteTokens ⊆ inputTokens` assumption behind `uncached = input − cached − write` is still only implicit.
- **2 — `reasoningEffort` dropped on the Chat Completions path but still recorded in the report**: untouched.
- **4 — `sumHarnessModelUsage` partial cost**: untouched (`usage.ts:71`; the field list just moved to `client.ts` as `HARNESS_MODEL_USAGE_FIELDS`).
- **5 — `#descendantUsage` as instance state**: untouched.
- **6 — long-context tier untested**: still no case crossing `272_000` (`usage.ts:158`); the three new tests cover the consistency guards instead.
- **7 and both nits**: untouched.

Verdict unchanged: nothing blocking.

packages/cf-harness/src/cli.ts
bfollington: Fixed in 2ff80b5bc; both help descriptions now align with the shared option-description column.

packages/cf-harness/src/prompt-loop.ts
bfollington: Addressed in 2ff80b5bc. Descendant usage is now local to runTranscript. The delegate result returns child usage and the current run records it immediately, removing the shared instance accumulator and its concurrent-run cross-attribution risk.

packages/cf-harness/src/model/usage.ts
bfollington: Addressed in 2ff80b5bc. The estimator now records an explicit withheld reason (missing cache detail, unknown model, invalid/inconsistent counters, provider pricing unavailable, or incomplete aggregate). The official GPT-5.6 guidance now explicitly names both cached_tokens and cache_write_tokens, and a real staging-gateway run observed a 1,120-token write followed by a 1,120-token read. The subset accounting assumption is now documented beside the arithmetic. The exact live results remain in the PR description and review discussion.

packages/cf-harness/src/contracts/interactive-chat.ts
bfollington: Addressed in 2ff80b5bc. HarnessChatGatewayUsage is now a direct alias of HarnessModelUsage, so the interactive contract cannot drift from the canonical usage shape.

packages/cf-harness/src/prompt-loop.ts
bfollington: This was already addressed by the newer 166db1394 head after Cubic's feedback: the parent forwards the configured mode unconditionally and the child provider validator rejects an incompatible override, so the mode is never silently dropped and there is no duplicated support predicate.

packages/cf-harness/src/model/usage.ts
bfollington: Addressed in 2ff80b5bc. Provider and estimated dollar costs are now summed only when every included usage entry reports the field. Mixed estimated/unestimated entries omit the aggregate cost and carry estimateWithheldReason=incomplete-estimates. A regression test covers both partial and complete aggregates.

packages/cf-harness/src/model/usage.ts
bfollington: Addressed in 2ff80b5bc. The 272,001-token boundary now has a regression test that pins the 2x input and 1.5x output multipliers. The adjacent comment ties the threshold to the GPT-5.6 model pages, which explicitly state that pricing rule.

packages/cf-harness/src/model/openai-compatible-gateway.ts
bfollington: Addressed in 2ff80b5bc. Any reasoning effort on a Chat-routed model is now rejected before provider traffic, so it cannot be silently dropped and then reported as applied. GPT-5.6 effort values are validated, and listModels advertises the verified supported set. Tests cover both the Chat-route rejection and an invalid GPT-5.6 value.

packages/cf-harness/src/prompt-loop.ts
```diff
@@ -2803,6 +2901,10 @@ export class CfHarnessPromptLoop {
       });
       summary = childResult.finalAssistantText;
       childModelTurns = childResult.modelTurns;
+      const childUsage = childResult.totalUsage ?? childResult.usage;
+      if (childUsage !== undefined) {
+        descendantUsage = childUsage;
```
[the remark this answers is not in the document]
bfollington: Addressed in b6bb545ef. Child usage is now recorded in the parent accumulator immediately after childLoop.runPrompt returns, before structured-return work or parent artifact persistence. A regression test forces delegate output persistence to fail after a 23-token child run and verifies the parent failure report retains the full 35-token parent-plus-child total. The full cf-harness suite passes: 573 tests.

PR #5207, opened 2026-07-30
fix(runner): validate the stored argument when a repair applies an update

seefeldb:
`packages/piece/src/schema-compatibility.ts` waives one evolution class — over an
OPEN argument object a candidate may name a brand-new optional field of any type
— on the ground that the runner validates the piece's merged durable argument
against the new schema in the setup transaction. On the repair route it did not.
A vintage storing `{count: "seven"}` legally under an open object, updated to a
version declaring `count?: number`, landed with no refusal: the bytes survived on
disk and nothing reading through the new contract could see them.

## The miss was one gate, not two

`applySetupState` skipped the argument re-stage whenever the pattern POINTER
already named the pattern being set up, and every repair path produces exactly
that. `pieces-controller.ts`'s roll-forward materialize commits the candidate
identity and then calls `runSynced`; `PatternUpdater`'s instantiated mode moves
the pointer with no setup at all, leaving a root that boots through the
cold-start setup repair. Both reach setup with `samePattern` true for what is an
update.

`patternSetupIdentity` is the signal that survives. It is stamped only once setup
has staged an identity's schema, arguments, internal cells and result projection,
so it names the version that actually staged the stored argument. `PatternUpdater`
already used the same comparison as `setupNeedsRepair`; `stagedByOtherVersion`
reads it in the runner.

**The `{ unresolvedLinkRaw }` leniency needed no change**, which corrects a
measurement this repo had recorded the other way (`docs/plans/…` claimed the fix
had to start there rather than at `sameStoredSetup`). Measured: it substitutes
the opaque placeholder only where the RAW value at a slot is a cell link that
materialized to nothing, so a plain wrong-typed value was already reaching
`validateSchemaValue` — and the hot-swap route was already refusing this pair,
under both write shapes (argument supplied at `run()`, and argument written
through the root's `argument` meta the way a vintage capture does). CT-1917's
deferral is untouched and is now covered on the newly-reachable path.

## Deliberate residual

An ABSENT setup marker still reads as "same": absence cannot be told from a
pending update, and re-staging every legacy root would validate — and rewrite
defaults over — arguments no update is touching. The window is one setup wide per
root, and an ordinary identity change validates regardless.

## Tests

`packages/runner/test/pattern-update-argument-validation.test.ts` covers both
update routes, a compatible control, the cold link (same candidate schema as the
refusal case, so the pair is a link-vs-plain differential rather than two
unrelated scenarios), and the same-version setup that must NOT refuse — a stored
argument can fail its own pattern's schema, and refusing there would turn a
pre-existing mismatch into a piece that will not boot.

The Tier-2 replay in `packages/piece/test/state-continuity.test.ts` now asserts
the refusal instead of pinning the stranding. Both go red when the condition is
reverted.

Local: `packages/piece` 31 passed / 358 steps; `packages/runner` 1096 passed /
5710 steps; `deno task integration` 7/7 packages and 120 pattern tests;
`deno fmt --check`, `deno lint`, `deno task check-docs` clean.

---
## Summary by cubic
Validate stored arguments on all update paths and during running‑piece reuse, classify and escalate refusals, block partial swaps, and repair the result schema only when it’s safe. Introduces a tri‑state setup marker to separate “matches”, “other”, and “absent”.

- **Bug Fixes**
  - Runner: re‑stage and validate stored arguments when `patternSetupIdentity` shows another version; running‑piece reuse validates in place and never falls through to `applySetupState`. Added `STORED_ARGUMENT_SCHEMA_REFUSAL` and `isStoredArgumentSchemaRefusal`; mirrors `{ unresolvedLinkRaw }` leniency and the cold‑doc skip on the in‑place path (CT‑1917).
  - Result schema: prevent partial swaps by writing only when setup proceeds; restore same‑version schema repair and perform it during safe reuse on both with‑arg and no‑arg branches; do not write when the setup marker is absent. Introduced a tri‑state stored‑setup check (`"matches" | "other" | "absent"`): re‑stage on `"other"`, repair schema only on `"matches"`.
  - Pieces: `PiecesController` escalates stored‑argument refusals to the roll‑forward backstop; preserves and logs original start errors for unrelated repair failures; nested‑piece instantiation repair explicitly opts out of argument re‑stage.
  - Tests/Docs: added `packages/runner/test/pattern-update-argument-validation.test.ts`; pinned the markerless exemption on the running‑reuse path; expanded running‑reuse deferral guards and tightened state‑continuity tests; `ensure-piece-running` notes `runtime.start()` does not run setup or repair a stale marker; docs updated to reference the tri‑state marker and clarify that `"absent"` is reported by the marker while the reuse path makes the no‑re‑stage decision.

<sup>Written for commit 5814176b63d38af1403b00269b15009af3401b0b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5207?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seefeldb:
All three addressed in 5792a2a8.

**cubic P1 — running-piece reuse gate.** Real and reachable; fixed. `maybeReuseRunningSetup` returned before `applySetupState`, so a live piece skipped the re-stage entirely. The reachable sequence is worth spelling out: a refused hot-swap is *logged*, not thrown, so the piece keeps running the old version under a pointer that already names the candidate — a repair then reported success over an argument nothing validated. Both gates now derive from a single `SetupStateReuse` value. Covered by "refuses a roll-forward onto a RUNNING piece whose swap already failed" (written red first).

That refactor also surfaced a third `applySetupState` call site — the nested-piece instantiation repair — which had silently started re-pointing the stored argument, contradicting its own comment promising it "leaves the existing argument — the piece's data — untouched". It now opts out explicitly.

**codex P1 — pointer swap not atomic with validation.** The observation is correct and the consequence is fixed, though not the way suggested. The pointer state after a refusal is now asserted in the rejection test, as you asked.

I did *not* conditionally restore the displaced identity: the roll-forward is reached **because** the previously pinned pattern could not be set up either, so restoring it restores a pattern already known not to work. The real defect was that the refusal was unrecoverable *and misreported* — `isCfcMigrationRejection` did not match the validation error, so `startEnsuredDefaultPattern` discarded it and rethrew the **original** start failure, leaving an operator reading a stream-marker stack for an argument problem while the root never opened again. It is now classified (`isStoredArgumentSchemaRefusal`, matched on a code-controlled prefix at the *start* of the message so user-influenced text cannot forge it) and escalated to the same roll-forward backstop a CFC migration rejection takes — "the pinned version cannot read this root" is exactly the runnability failure that backstop exists for. Repair failures it cannot classify are now logged before the original error is surfaced, instead of vanishing.

**codex P2 — resolved investigation in a live plan.** Agreed. Moved to `docs/history/plans/pattern-update-open-argument-investigation.md` with the required metadata header; the live plan now states present facts only.

Two further things a self-review round turned up, both fixed here: the "same-version setup" test was green for the wrong reason (its stored value was *valid* under the open vintage's own schema, so it passed with the condition forced to `true`) and now stores a value the vintage's own typed schema rejects; and the absent-marker exemption is stated honestly — `patternSetupIdentity` landed 2026-07-24, so it currently exempts most stored roots rather than a rare tail.

@codex review

seefeldb:
All four addressed in 3e212f58, plus a regression my previous commit introduced that a self-review round caught.

**The regression, first.** Routing the running-piece reuse gate through `!restageStoredArgument` (my fix for cubic's P1) let a live piece with a stale marker fall through to `applySetupState`. That stages the incoming version's argument schema, internal manifest and result projection — but only the pattern watcher can cancel live nodes and instantiate new ones. Measured on a piece whose two versions differ only in which cell backs a field: after the fall-through the live value read as the *new* version while its nodes still served the old one, and `patternSetupIdentity` was stamped forward, erasing the mismatch a later repair uses to notice. Invalid arguments aborted the tx and were fine; **valid** ones — the common case — silently rewired a live piece.

The gate is back on `sameStoredSetup`. What replaces it is `validateStoredArgument`: a stale marker means nobody checked the stored argument against this pattern, so it is validated *in place* — no write, no schema retarget, nothing that moves the piece — and a refusal still surfaces. That also restores the documented nested-replay guard on the supplied-argument branch, which the fall-through had been defeating by sending opaque `Cell` handles down the strict path (#4717). Guarded both ways: the new case reds under the old predicate, and the refusal case reds if validation is dropped.

**P2 — heal errors mislabelled.** Right, and wider than the one site I'd fixed: `healDefaultRootByRollForward` hard-coded "failed CFC migration" in the error it builds for *every* escalation, and the already-pinned short-circuit said "the migration cannot be repaired". Both now name the check that actually refused.

**P2 — Tier 2 does not cover markerless roots.** Correct, and I'd claimed it did. A Tier-2 capture runs setup through the current runner, which always stamps the marker, so the replay never sees a markerless root. The claim is removed, and the exemption is now pinned as a decision instead: a test strips the marker, confirms the update is admitted, and asserts the marker is written on the way through so the exemption closes behind itself.

**P2 — history doc not indexed** and **P2 — `superseded-by` not repo-relative.** Both fixed; the index entry is under "Audits and reports" and the path is now `docs/plans/...` like the other eight.

Two further stale claims a coherence sweep turned up, both authoritative comments the change had falsified: `migration-reason.ts` and `pieces-controller.ts` each said the CFC rejection is the *only* class the runnability backstop acts on (now one of two), and `check-update-default-pattern.test.ts` justified its construction with "arg validation is skipped for an unchanged identity" — which this change inverts on exactly the state that test builds. Also corrected: `pattern-verb-contract.md`'s line-number citations into `schema-compatibility.ts`, which the JSDoc expansion had shifted.

Finally, `nested-piece-setup-repair.test.ts` could not guard the opt-out it is cited for — both its patterns took `Record<string, never>`, so a re-stage was a no-op and mutating the opt-out left the file green. Its piece now carries a real argument the newer version's schema rejects.

@codex review

seefeldb:
**P1 (result-schema staged on the reuse path) — confirmed and fixed** in 36e7577.

Reproduced: with the two versions declaring different result shapes, a reused running piece had the candidate's result schema written and committed, so reads resolving the root through that meta described the new version while its nodes kept producing the old one's. That is the same partial swap the reuse gate exists to prevent, reached by a second route.

Worth recording that it is **pre-existing** — `main` has the same ordering, so this was not introduced by the branch. I fixed it here anyway because the guarantee this branch adds (a running piece is validated in place and nothing moves it) is not true while that write escapes the gate. The write now sits after the gate; a piece that is *not* reused falls through to `applySetupState`, which stages the matching projection in the same transaction, so the schema and the projection cannot disagree. `updateResultSchemaMeta` already no-ops when the schema is unchanged, so ordinary same-version reuse is unaffected.

The half-swap guard now gives its two versions different result shapes and asserts the stored `schema` meta as well as the live value, so both halves of a partial swap are covered.

Separately, a coherence sweep found comments this change had falsified, now corrected in e77186ec: `pieces-controller.ts` still said the backstop escalates on "EXACTLY one signal" ten lines above the two-signal gate; `ensure-piece-running.ts` and its tests justified themselves with "runSynced on an already-running piece is idempotent — it simply returns without doing anything", which stopped being exactly true once a stale setup marker can surface a refusal (the assertions still hold; only the stated reason was wrong). Three citations that my own insertions had shifted were repointed too — including one in `pattern-verb-contract.md` that had come to land *inside* the new classifier's JSDoc, sending a reader to the wrong guard.

Stale citations that predate this branch and point at unrelated code are left alone and tracked separately; they are repo-wide line-number rot rather than this change's business.

@codex review

seefeldb:
Round-3 self-review findings addressed in 74e83e7a5.

**A behavioural loss my previous commit introduced.** Moving `updateResultSchemaMeta` past the reuse gate was wider than the hazard: `restageStoredArgument === false` proves the setup marker names *this* pattern, so the running graph is this pattern and recording its result schema is safe. Suppressing it there too dropped a legitimate repair — measured: run a piece, strip its `schema` meta, re-run setup for the same version, and the meta stays gone, costing `getResultCellWithSourceSchema` its type, `durableSourceContract` its precondition, and `derivePersistedLinkLabel` its same-transaction hatch. The write now happens inside the gate on the opposite polarity of the same signal, so half-swap suppression is unchanged and the same-version repair is back.

**A coverage hole in the invariant the running-piece fix rests on.** `validateStoredArgument` mirrors the re-stage branch's two CT-1917 deferrals, and its doc comment claims the two paths cannot disagree — but only the re-stage side was pinned. Measured: dropping `unresolvedLinkRaw` from the in-place call, or removing its cold-doc skip, left the entire suite green. A regression there would have re-created CT-1917 for every running piece whose argument slot links into an unsynced doc, invisibly. Two cases now cover it on the running-reuse route, each verified to red under exactly the mutation that previously went undetected.

Also: `schema-compatibility.ts` named only `applySetupState` → `validateArgument` as what makes its waiver sound, which points an auditor at a function the running-piece route never reaches — both checking sites are named now. The plan's conflict-set row still listed the nested-piece repair as a bare-`edit()` re-stager after it opted out. And both escalation classifiers read the error rather than the document it originated from, so a nested piece's refusal reads like the root's — a property shared with the pre-existing CFC trigger, now written down at the gate rather than left implicit.

@codex review

seefeldb:
Both P2s fixed, plus a hole in my own fix that a self-review pass caught.

**Repair schema metadata on supplied-input reuse** — correct, fixed in db3199b26. The repair sat inside the no-argument branch, and the supplied-argument branch returns from a different place, so `PieceManager.runWithPattern` and the nested-piece re-run paths were left unrepaired. The write is now hoisted above both branches. (A self-review had asked exactly this question a round earlier and I had not answered it directly.)

**Stop claiming auto-start revalidates running pieces** — also correct, fixed in the same commit. `ensurePieceRunning` calls `runtime.start()`, which returns as soon as it finds the piece registered, without running setup — as the comment immediately above that call already said. I had over-corrected an earlier stale claim into a new false one; the JSDoc now describes what the route actually does, including that it does *not* repair a stale setup marker.

**And the hole those two exposed**, fixed in 1ce451a23: `stagedByOtherVersion` was a boolean answering a three-state question. Its `false` meant *either* "the marker names this pattern" *or* "there is no marker at all", and the schema repair read that as positive evidence the running graph is this pattern. Absence licenses no such thing — a markerless piece can be running something other than what the pointer names (a pre-marker root resumes into the running set without setup; `PatternUpdater`'s instantiated mode moves the pointer with no setup; a refused swap leaves the graph where it was). Measured: with the marker stripped, a piece running v1 under a pointer naming v2 had v2's result schema written over it — the same half-swap the reuse gate exists to prevent, re-opened for exactly the population the markerless exemption calls most of the installed base.

Fixed at the source rather than the call site: `storedSetupMarker()` returns `"matches" | "other" | "absent"`, and `SetupStateReuse` carries `storedSetupMatches` separately. The re-stage keys on `"other"`; the schema repair keys on `"matches"` only. The new guard reds when the two are collapsed back into one boolean.

Also corrected the waiver comment in `schema-compatibility.ts`: it said `applySetupState` handles "a stopped piece" and `validateStoredArgument` "a running one", but the hot-swap re-stages a *running* piece through `applySetupState`. The real line is whether the caller will (re)instantiate the graph.

@codex review

PR #5208, opened 2026-07-30
Split dialog messages into documents without an `[ID]` directive

danfuzz:
Every message `llmDialog` pushes becomes its own document — the LlmDerived stamping reads each element back and parses it as a link. Seven call sites hand-attached an `[ID]` directive to each message: five as a property on a literal, two by mutating a message afterwards.

This creates those documents explicitly instead, which **removes the last live use of `ID` outside tests** and unblocks retiring `ID`/`ID_FIELD` entirely.

### What the directives were actually doing

Not causing the split. `Cell.push` runs `recursivelyAddIDIfNeeded`, which assigns identity to objects in arrays, so a message pushed with no directive still lands in a document of its own — verified by reading the stored elements on a path with no directive and no explicit cell:

```
RAW[0]: {"/":{"link@1":{"path":[],"id":"of:fid1:YMe7zv4i…"}}}
RAW[1]: {"/":{"link@1":{"path":[],"id":"of:fid1:-PotTAsb…"}}}
```

What a directive chose was the **cause** the entity id derives from. So removing them changes which ids get minted, not whether documents exist.

The source comment on those lines said built-ins get no automatic splitting. That was misleading, and an earlier revision of this description repeated it; `f4dffaa16` corrects it and two others that described messages as `[ID]`-carrying.

## The change

`pushModelMessages()` already owned "push these messages and stamp their docs", so minting identity moved there — it creates a cell per message under the same cause the directives used, writes the message, and pushes that cell. Its callers now build plain messages. The two paths that do not route through it (the user message, and the substituted error message) do the same inline.

Two further error paths push directly and were left alone deliberately: with no directive they take the automatic-identity route above, which produces the same bare link. A review flagged those as inconsistent storage; the raw values show otherwise, and the test added in `005e61df6` pins it.

Net effect at the call sites: three literals lose a property, two mutation blocks disappear entirely, and every `& { [ID]: unknown }` intersection goes away.

## Storage shape is unchanged

A base-relative link serializes bare. Measured against the `[ID]` split on the same data:

```
[ID] split:        {"path":[],"id":"of:fid1:nuoC…"}
explicit + base:   {"path":[],"id":"of:fid1:rrIW…"}
```

Byte-identical apart from the entity id itself. Two things worth knowing about that id:

- It no longer derives from a directive, so **ids differ from what the old code would mint** for the same message. Nothing re-derives them — the cause already carried a fresh `crypto.randomUUID()`, so identity here was always per-message and opaque — but existing persisted messages and new ones are not addressed alike.
- Without `{ base }` the link also carries `space` and `scope`, and a cell created *with* a schema puts the schema in the link too. The latter matters here: `pushModelMessages`' own comment notes the messages link's schema wins over an `asSchema` handle inside `push()`. Both are avoided — base-relative, schema-less cell.

## Testing

- Full **workspace** suite green (all packages, 218s), plus `deno fmt --check`, `deno lint`, and `deno task check-docs` (523 blocks), verified locally at `a7322e3d1`. The `runner` package alone is 1097 passed / 0 failed.
- Two error paths had **no end-to-end coverage** — nothing in the tree asserted the messages they produce. Tests for both are added here. The invalid-content one was written against the existing implementation and verified to pass there *before* the migration, so it characterizes current behavior rather than being shaped to fit the rewrite. Both pin the property this change rides on: every stored element parses as a link.

### The scope fix, now with a regression test — and a deeper root

Review found that the explicit documents defaulted to `"space"` scope while
anchored elements inherit the array's. The first fix threaded
`inputs.key("messages")`'s scope through — but that is the *input slot's*
link, which sits on the dialog's process document at `"space"` scope no matter
where the messages array actually lives. `push()` resolves the link before
writing; the explicit sites did not. Under a session-scoped messages cell the
user message's document landed in the `"space"` partition while its bare
element link resolved against `"session"`: the content was unreachable, and
the dialog stalled at `pending: true` trying to read its own transcript.

That stall is exactly the symptom that defeated the earlier attempts at a red
test — the "harness problem" was the bug itself firing. All three sites now
resolve the messages link first and mint the document at the resolved base.

The new test (`llm-dialog-message-scope.test.ts`) drives a dialog whose
pattern owns a `Cell.perSession` messages array — pattern-owned rather than
argument-injected, which sidesteps the binding questions entirely. It asserts
the resolved array is genuinely session-scoped (no vacuous pass), that both
messages read back through the array, and that each stored element is a bare
link. It is red against the slot-scope version (content lost, dialog stalled),
green with the resolved-scope version, and green against the pre-PR baseline —
the `[ID]` anchoring inherited the resolved scope by construction, so this
restores parity rather than adding behavior.

The `LlmDerived` stamping loop below `pushModelMessages()` takes the same
resolved base, so the provenance stamp records against the partition the
document actually lives in rather than the slot's.

### Pushing cells rather than links (@seefeldb)

The three sites originally pushed `cell.getAsLink(...)` behind a cast, on the
reasoning that pushing the cell inlines space, scope and schema per element.
Measured, that was 685 B per stored message against 87 B for a bare link — but
it was treating two separable costs as one, and the cast was reaching through
the very guard that keeps user-land from writing arbitrary links.

Both halves are now fixed at the source:

| | space+scope | schema | stored |
|---|---|---|---|
| push cell, before | inlined | inlined | 685 B |
| + `normalizeAndDiff` elision | elided | inlined | 600 B |
| + schema-less document | elided | absent | **87 B** |

`normalizeAndDiff()` converted a written `Cell` with no base, so the link
re-inlined what the destination slot already supplied — while the sibling
branch a hundred lines up already elided against that same `link`. Fixing it
surfaced a second latent bug of the same species: the self-reference no-op
guard also compares base-lessly, so once the link elides it stops recognizing
a cell written to its own slot, and `c.set(c)` resolves into a cycle. Four test
assertions needed a base they were missing, plus one in
`packages/runner/integration/` that CI caught and the workspace `deno task test`
does not cover — the package integration suites run under a separate task.

The invariant this establishes: **a stored link is only meaningful together with
the slot it was read from.** Parsing one without its base is a bug even where it
currently works. The two base-less `parseLink()` sites left in `src`
(`llm-dialog.ts` stamping, `sqlite-builtins.ts` per-row labels) are safe only
because they fall back to `link.space ?? base.space`.

Schema is not affected by elision, so the documents are also created
schema-less — a document's schema is inlined into every link written to it, and
the array's own items schema already describes what a reader finds. Typing stays
at the call site via `runtime.getCell<Schema<typeof LLMMessageSchema>>`.

Result: `push(messageCell)`, no casts, and the same 87 B on disk.

## Reviewing

Six commits, each test in its own. The invalid-content test lands before the `[ID]` removal and was verified green there first, so it characterizes existing behavior rather than being shaped to fit the rewrite. The failed-request test follows the removal. The scope test lands with its fix — it cannot be staged separately, since it is red without it.

The reviewer question I would most value a second opinion on: whether minting per-message identity inside `pushModelMessages()` is the right home for it, versus keeping it at the call sites. It reads better to me — the function that pushes and stamps is now also the one that mints — but it does mean callers can no longer choose a cause.

There is a `TODO(seefeld)` on the original lines about event ids eventually supplying the cause; that intent is preserved, just relocated.

---
## Summary by cubic
Split `llmDialog` messages into their own documents without `[ID]`. Identity is minted after resolving the `messages` link, and we now push message cells directly; storage stays as bare, base-relative links and per-message docs are schema-less.

- **Refactors**
  - Removed all `[ID]` usage; identity minting is done inline when pushing one cell per message using the resolved `messages` link.
  - Per-message docs are schema-less; the array’s items schema defines shape.
  - `normalizeAndDiff()` elides cell-valued writes against the slot, enabling bare links and removing casts/extra base cells.

- **Bug Fixes**
  - Mint docs and stamp in the resolved array’s space/scope; added a session-scope regression test.
  - Fixed self-reference no-op guard in `normalizeAndDiff()` to handle elided bases; updated tests to compare links relative to the slot and stabilized with `runtime.idle()`.
  - In `integration/sync-schema-path`, parse stored links against the slot that holds them to recover `space`.

<sup>Written for commit 65bf6f044ab2cdbfbb30b8a2f23ef4dcb47d93fd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5208?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/builtins/llm-dialog.ts
+++ b/packages/runner/src/builtins/llm-dialog.ts
@@ -56,7 +56,7 @@ import {
   recordRelevantSchemaWritePolicyInput,
 } from "../cell.ts";
 import { resolveLinkScope } from "../scope.ts";
-import { type CellScope, ID, NAME, type Pattern } from "../builder/types.ts";
+import { type CellScope, NAME, type Pattern } from "../builder/types.ts";
 import { resolveStoredPatternAsync } from "./op-pattern-ref.ts";
 import { getEntityId } from "../create-ref.ts";
 import {
@@ -3148,16 +3148,25 @@ export function llmDialog(
 
           // Before starting request, set pending and append the new message.
           pending.withTx(tx).set(true);
-          inputs.key("messages").withTx(tx).push(
-            {
-              ...event,
-              // Add ID manually, as for built-ins this isn't automated
-              // TODO(seefeld): Once we have event ids, it should be that.
-              [ID]: { llmDialog: { message: cause, id: crypto.randomUUID() } },
-              // Cast because we can't yet express ArrayBuffer in JSON Schema
-            } as Schema<
-              typeof LLMMessageSchema
-            >,
+          // Each message becomes its own document, which the LlmDerived
+          // stamping downstream relies on. Built-ins do not get automatic
+          // entity splitting, so the document is made explicitly and its link
+          // pushed. The link is taken relative to the messages cell so it
+          // stores bare, exactly as an automatically split element does.
+          // TODO(seefeld): Once we have event ids, the cause should be that.
+          const messagesForPush = inputs.key("messages");
+          const messageCell = runtime.getCell(
+            messagesForPush.getAsNormalizedFullLink().space,
+            { llmDialog: { message: cause, id: crypto.randomUUID() } },
+            undefined,
+            tx,
+          );
+          messageCell.withTx(tx).set(
+            // Cast because we can't yet express ArrayBuffer in JSON Schema
+            { ...event } as Schema<typeof LLMMessageSchema>,
+          );
+          messagesForPush.withTx(tx).push(
+            messageCell.getAsLink({ base: messagesForPush }) as never,
           );
 
           // Set up new request (abort existing ones just in case) by allocating
@@ -3423,7 +3432,24 @@ async function startRequest(
     const startIndex = (messagesCell.withTx(tx).get() as
       | readonly CfcConfClause[]
       | undefined)?.length ?? 0;
-    messagesCell.withTx(tx).push(...messages);
+    // Each message becomes its own document, which the stamping below reads
+    // back. Built-ins get no automatic entity splitting, so the documents are
+    // made here explicitly. Links are taken relative to the messages cell, so
+    // they store bare -- the same shape an automatically split element has.
+    // TODO(seefeld): Once we have event ids, the cause should be that.
+    const space = messagesCell.getAsNormalizedFullLink().space;
+    messagesCell.withTx(tx).push(
+      ...messages.map((message) => {
+        const messageCell = runtime.getCell(
+          space,
+          { llmDialog: { message: cause, id: crypto.randomUUID() } },
+          undefined,
+          tx,
+        );
+        messageCell.withTx(tx).set(message);
+        return messageCell.getAsLink({ base: messagesCell }) as never;
+      }),
+    );
     if (runtime.cfcEnforcementMode === "disabled") {
       return;
     }
@@ -3655,11 +3681,10 @@ Some operations (especially \`invoke()\` with patterns) create "Pages" - running
           "LLM returned invalid/empty content, adding error message",
         );
         const errorMessage = {
-          [ID]: { llmDialog: { message: cause, id: crypto.randomUUID() } },
           role: "assistant",
           content:
             "I encountered an error generating a response. Please try again.",
-        } satisfies BuiltInLLMMessage & { [ID]: unknown };
+        } satisfies BuiltInLLMMessage;
 
         await safelyPerformUpdate(
           runtime,
@@ -3667,8 +3692,18 @@ Some operations (especially \`invoke()\` with patterns) create "Pages" - running
           internal,
           requestId,
           (tx) => {
-            messagesCell.withTx(tx).push(
+            // As above: its own document, made explicitly, link stored bare.
+            const errorCell = runtime.getCell(
+              messagesCell.getAsNormalizedFullLink().space,
+              { llmDialog: { message: cause, id: crypto.randomUUID() } },
+              undefined,
+              tx,
+            );
+            errorCell.withTx(tx).set(
               errorMessage as Schema<typeof LLMMessageSchema>,
+            );
+            messagesCell.withTx(tx).push(
+              errorCell.getAsLink({ base: messagesCell }) as never,
             );
             pending.withTx(tx).set(false);
           },
@@ -3688,11 +3723,6 @@ Some operations (especially \`invoke()\` with patterns) create "Pages" - running
             llmContent,
             toolCallParts,
           );
-
-          // Add ID to assistant message and publish immediately
-          (assistantMessage as BuiltInLLMMessage & { [ID]: unknown })[ID] = {
-            llmDialog: { message: cause, id: crypto.randomUUID() },
-          };
 
           await safelyPerformUpdate(
             runtime,
@@ -3758,10 +3788,9 @@ Some operations (especially \`invoke()\` with patterns) create "Pages" - running
             );
             // Add error message instead of invalid partial results
             const errorMessage = {
-              [ID]: { llmDialog: { message: cause, id: crypto.randomUUID() } },
               role: "assistant",
               content: "Some tool calls failed to execute. Please try again.",
-            } satisfies BuiltInLLMMessage & { [ID]: unknown };
+            } satisfies BuiltInLLMMessage;
 
             await safelyPerformUpdate(
               runtime,
@@ -3780,12 +3809,6 @@ Some operations (especially \`invoke()\` with patterns) create "Pages" - running
 
           // Create and publish tool result messages
           const toolResultMessages = createToolResultMessages(toolResults);
-
-          toolResultMessages.forEach((message) => {
-            (message as BuiltInLLMMessage & { [ID]: unknown })[ID] = {
-              llmDialog: { message: cause, id: crypto.randomUUID() },
-            };
-          });
 
           const success = await safelyPerformUpdate(
             runtime,
@@ -3875,10 +3898,9 @@ Some operations (especially \`invoke()\` with patterns) create "Pages" - running
       } else {
         // No tool calls, just add the assistant message
         const assistantMessage = {
-          [ID]: { llmDialog: { message: cause, id: crypto.randomUUID() } },
           role: "assistant",
           content: llmResult.content,
-        } satisfies BuiltInLLMMessage & { [ID]: unknown };
+        } satisfies BuiltInLLMMessage;
 
         // Ignore errors here, it probably means something else took over.
         await safelyPerformUpdate(
@@ -3910,11 +3932,10 @@ Some operations (especially \`invoke()\` with patterns) create "Pages" - running
         ? error.message
         : String(error);
       const errorMessage = {
-        [ID]: { llmDialog: { message: cause, id: crypto.randomUUID() } },
         role: "assistant",
         content:
           `I encountered an error generating a response: ${errorMessageText}`,
-      } satisfies BuiltInLLMMessage & { [ID]: unknown };
+      } satisfies BuiltInLLMMessage;
 
       safelyPerformUpdate(runtime, pending, internal, requestId, (tx) => {
         messagesCell.withTx(tx).push(
--- a/packages/runner/src/builtins/llm-dialog.ts
+++ b/packages/runner/src/builtins/llm-dialog.ts
@@ -3462,8 +3462,8 @@ async function startRequest(
       builtinId: "llmDialog",
     });
     // Record the stamping schema for each pushed message's own entity doc
-    // (every model push carries an [ID] sigil, so each message splits into
-    // its own doc). The messages link carries its own schema, which wins over
+    // (every message is appended as a link to a document of its own, so each
+    // one is separately addressable). The messages link carries its own schema, which wins over
     // an `asSchema` handle inside `push()` (`resolvedLink.schema ?? ...`), so
     // the stamp cannot ride the array handle — instead this mirrors the
     // split-entity idiom in data-updating.ts (`recordRelevantSchemaWrite-
--- a/packages/runner/src/builtins/llm-dialog.ts
+++ b/packages/runner/src/builtins/llm-dialog.ts
@@ -3159,24 +3159,21 @@ export function llmDialog(
           // document itself, not the input slot that points at it, which can
           // sit at a different scope. `push()` resolves the same way before it
           // writes; a document minted at the slot's scope instead lands in a
-          // partition the array's readers never look in. The link is then
-          // taken relative to that resolved base, so it stores bare, exactly
-          // as an anchored element does.
+          // partition the array's readers never look in.
+          //
+          // The document is left schema-less: the schema it would carry ends
+          // up inlined in every link written to it, and the array's own items
+          // schema already describes what a reader finds there.
           const messagesForPush = inputs.key("messages");
           const messagesBase = resolveLink(
             runtime,
             tx,
             messagesForPush.getAsNormalizedFullLink(),
           );
-          const messagesDoc = runtime.getCellFromLink(
-            messagesBase,
-            undefined,
-            tx,
-          );
-          const messageCell = runtime.getCell(
+          const messageCell = runtime.getCell<Schema<typeof LLMMessageSchema>>(
             messagesBase.space,
             { llmDialog: { message: cause, id: crypto.randomUUID() } },
-            LLMMessageSchema,
+            undefined,
             tx,
             messagesBase.scope,
           );
@@ -3184,19 +3181,7 @@ export function llmDialog(
             // Cast because we can't yet express ArrayBuffer in JSON Schema
             { ...event } as Schema<typeof LLMMessageSchema>,
           );
-          // `push()` is typed for message values and cells of them, not for
-          // links (`data-updating.ts` resolves a link where a value is
-          // expected), so appending a bare link needs the cast to the type the
-          // site resolves it to. Pushing the cell itself would type cleanly but
-          // store the link with space, scope and the whole message schema
-          // inlined -- a copy per message.
-          messagesForPush.withTx(tx).push(
-            messageCell.getAsLink({
-              base: messagesDoc,
-            }) as unknown as Schema<
-              typeof LLMMessageSchema
-            >,
-          );
+          messagesForPush.withTx(tx).push(messageCell);
 
           // Set up new request (abort existing ones just in case) by allocating
           // a new request Id and setting up a new abort controller.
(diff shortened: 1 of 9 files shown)


packages/runner/src/builtins/llm-dialog.ts
```diff
@@ -3148,16 +3148,25 @@ export function llmDialog(
...
+            { llmDialog: { message: cause, id: crypto.randomUUID() } },
+            undefined,
+            tx,
+          );
+          messageCell.withTx(tx).set(
+            // Cast because we can't yet express ArrayBuffer in JSON Schema
+            { ...event } as Schema<typeof LLMMessageSchema>,
+          );
+          messagesForPush.withTx(tx).push(
+            messageCell.getAsLink({ base: messagesForPush }) as never,
```
seefeldb: why is the `as never` necessary here?

seefeldb: this should maybe just be `messageCell` without the .getAsLink?

packages/runner/src/builtins/llm-dialog.ts
```diff
@@ -3423,7 +3432,24 @@ async function startRequest(
...
+    messagesCell.withTx(tx).push(
+      ...messages.map((message) => {
+        const messageCell = runtime.getCell(
+          space,
+          { llmDialog: { message: cause, id: crypto.randomUUID() } },
+          undefined,
+          tx,
+        );
+        messageCell.withTx(tx).set(message);
+        return messageCell.getAsLink({ base: messagesCell }) as never;
```
seefeldb: same, probably just the cell here

packages/runner/src/builtins/llm-dialog.ts
```diff
@@ -3655,21 +3681,30 @@ Some operations (especially \`invoke()\` with patterns) create "Pages" - running
...
+              messagesCell.getAsNormalizedFullLink().space,
+              { llmDialog: { message: cause, id: crypto.randomUUID() } },
+              undefined,
+              tx,
+            );
+            errorCell.withTx(tx).set(
               errorMessage as Schema<typeof LLMMessageSchema>,
             );
+            messagesCell.withTx(tx).push(
+              errorCell.getAsLink({ base: messagesCell }) as never,
```
seefeldb: and here as well

packages/runner/src/builtins/llm-dialog.ts
```diff
@@ -3423,7 +3432,24 @@ async function startRequest(
...
       | readonly CfcConfClause[]
       | undefined)?.length ?? 0;
-    messagesCell.withTx(tx).push(...messages);
+    // Each message becomes its own document, which the stamping below reads
+    // back. Built-ins get no automatic entity splitting, so the documents are
+    // made here explicitly. Links are taken relative to the messages cell, so
+    // they store bare -- the same shape an automatically split element has.
+    // TODO(seefeld): Once we have event ids, the cause should be that.
+    const space = messagesCell.getAsNormalizedFullLink().space;
+    messagesCell.withTx(tx).push(
```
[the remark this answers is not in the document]
danfuzz: Checked this by reading what actually gets stored, and the finding does not hold — but it pointed at a real coverage gap, which is now filled.

**The array does not mix bare values with links.** `Cell.push` runs `recursivelyAddIDIfNeeded`, which assigns identity to objects in arrays. So a message pushed without a directive still splits into its own document. Stored elements on this path, with no explicit cell involved:

```
RAW[0]: {"/":{"link@1":{"path":[],"id":"of:fid1:YMe7zv4i…"}}}
RAW[1]: {"/":{"link@1":{"path":[],"id":"of:fid1:-PotTAsb…"}}}
```

Both bare links. I did apply your suggested fix first, then wrote a test to prove it was needed — and the test passed *without* the fix, which is what sent me to look at the raw values. The fix is reverted; it was indirection for a problem that does not exist.

**What the `[ID]` directive actually did** was choose the *cause* the entity id derives from. It was never what caused the split. That distinction matters for this PR: removing the directives changes which ids get minted, not whether documents are created — which is what the PR body claims, and what the storage comparison in it shows.

Worth noting the source comment you were reasoning from — *"Add ID manually, as for built-ins this isn't automated"* — is misleading, and I repeated it uncritically in my own PR description. Splitting is automatic; only the cause was manual. I will fix that wording.

**Your instinct about the path was right even though the conclusion was not.** That `.catch` branch had no end-to-end coverage at all, so nothing asserted its behavior either way. `005e61df6` adds a test for it that pins every element parsing as a link, so if `recursivelyAddIDIfNeeded` ever stops covering this case, it fails loudly rather than silently producing the mixed storage you were worried about.

— Claude (via Claude Code, on @danfuzz's session)

packages/runner/src/builtins/llm-dialog.ts
danfuzz: Good catch on both, and the second one sent me to measure rather than argue. Full matrix, all three verified by reading what actually lands in storage:

| variant | types | stored element |
|---|---|---|
| schema-less cell + `getAsLink({base})` | needs cast | bare |
| **schema-typed cell + push the cell** | **clean, no cast** | space + scope + **the whole message schema inlined** |
| schema-typed cell + `getAsLink({base})` | needs cast | bare |

Pushing the cell directly stores this per message:

```json
{"link@1":{"path":[],"id":"of:fid1:1SCU…","space":"did:key:…","scope":"space",
 "schema":{"type":"object","properties":{"role":…,"content":{"anyOf":[…]}},"required":["role","content"]}}}
```

against `{"path":[],"id":"of:fid1:y0nl…"}` for an anchored element. A full copy of `LLMMessageSchema` on every message seemed like too much to pay, so I kept the link and the cast.

I dug into whether the cast is avoidable, and it is not: `Cell.push` is typed `U | AnyCellWrapping<U>` — message values or cells of them, never links — and the write path that turns a pushed cell into a link calls `getAsLink({ includeSchema: true })` unconditionally (`data-updating.ts:694`), so the inlining is not configurable from the call site either. Bare storage therefore requires handing `push` a link it is not typed for. Both facts are now comments at the call site (`2a8a47600`), so the next reader does not have to re-derive them.

Your suggestion did fix the other half, though: the cells are now created with `LLMMessageSchema` instead of schema-less, so they are `Cell<Schema<typeof LLMMessageSchema>>` rather than `Cell<unknown>` — which is what the original cast was partly papering over.

— Claude (via Claude Code, on @danfuzz's session)

packages/runner/src/builtins/llm-dialog.ts
danfuzz: Measured this one along with the other two — details and the storage comparison are in [the thread on the user-message push](https://github.com/commontoolsinc/labs/pull/5208#discussion_r3684927660).

Short version: pushing the cell here types cleanly and drops the cast, but stores the link with space, scope and the full `LLMMessageSchema` inlined, per message. Kept the base-relative link for the storage shape. What your suggestion did fix is the cell's own type — it now carries `LLMMessageSchema` rather than being created schema-less as `Cell<unknown>` (`2a8a47600`).

This site also had a real bug your comment led me to: it was creating the document with `getCell(space, cause, undefined, tx)`, which defaults to `"space"` scope, whereas an anchored element inherits the array's scope. Under a scoped messages cell that would have split documents across two partitions. Now passes the base's scope through.

— Claude (via Claude Code, on @danfuzz's session)

packages/runner/src/builtins/llm-dialog.ts
danfuzz: Same as the other two — measurement and storage comparison in [the user-message push thread](https://github.com/commontoolsinc/labs/pull/5208#discussion_r3684927660).

Pushing the cell types cleanly but inlines the full message schema into each stored link, so this keeps the base-relative link and the cast, with both reasons now noted in the code. The cell here is likewise created with `LLMMessageSchema` now, and inherits the messages cell's scope rather than defaulting to `"space"` (`2a8a47600`).

— Claude (via Claude Code, on @danfuzz's session)

packages/runner/test/llm-dialog-error-messages.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    );
+
+    const resultCell = runtime.getCell(
+      signer.did(),
+      "llm-invalid-content",
+      resultSchema,
+      tx,
+    );
+    const result = runtime.run(tx, testPattern, {}, resultCell);
+    tx.commit();
```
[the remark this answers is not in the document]
danfuzz: Valid — fixed in `2a8a47600`.

Both setup commits in this file now `await`, in this test and in the invalid-content one above it. You are right about the consequence: a rejected commit would have surfaced as an unhandled rejection attributed to the runtime rather than as a failure of the test that caused it, which is a bad trade in a file whose whole purpose is characterising error paths.

I did not add an assertion on the returned result. The subsequent `waitForLlmMessages` already fails the test if the setup did not land — it waits on `pending === false` and a message count, neither of which can be reached from an uncommitted pattern — so an explicit check would be a second spelling of the same guarantee. Happy to add one if you would rather have it stated locally.

— Claude (via Claude Code, on @danfuzz's session)

packages/runner/test/llm-dialog-message-scope.test.ts
```diff
@@ -0,0 +1,146 @@
...
+    const resultCell = runtime.getCell(
+      signer.did(),
+      "llm-message-scope",
+      resultSchema,
+      tx,
+    );
+    const result = runtime.run(tx, testPattern, {}, resultCell);
+    await tx.commit();
+
+    const addMessage = await result.key("addMessage").pull();
```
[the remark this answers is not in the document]
danfuzz: Valid, and applied in `8597608f1` — with one correction and one caveat.

**Correction to the citation.** The requirement is not blanket "after every commit." `packages/runner/README.md:445` reads a *result* cell as commit → `idle()` → `pull()`, and the virtual-defaults note (README:277, :656) is about schema defaults not being backed by a real write — a different hazard than this one. What actually applies here is narrower: `addMessage` is a stream handle produced by the pattern run, and nothing in `await tx.commit()` guarantees the graph has settled far enough to expose it.

**Caveat worth recording.** This idiom is not local to these tests — commit-immediately-followed-by-`pull()`/`get()` appears in roughly twenty `packages/runner/test/` files. So either it is fine in general and this was belt-and-braces, or there is a broader latent flakiness that this PR is not the right place to fix. I have applied it to all three tests this PR touches rather than diverging within the PR, and left the rest alone.

**On the risk of a false green.** The concern with adding a settle step to a regression test is that it papers over the thing under test. I checked rather than assumed: with the `idle()` in place, `llm-dialog-message-scope.test.ts` is still red against the slot-scope version of the fix (`58bd58c8c`) and green against the resolved-scope version. The discriminator survives.

— Claude (Opus 5), working with @danfuzz

danfuzz:
@seefeldb — following up on your question about `.push(cell.getAsLink(...))` vs `.push(cell)`. You were right on both counts, and they turned out to be two separate fixes. Landed in `a7322e3d1`.

**The measurement I should have led with.** My reason for pushing the link was that pushing the cell inlines space, scope and schema per element. That was true — 685 B per stored message against 87 B for a bare link — but it conflated two costs that come apart cleanly:

| | space+scope | schema | stored |
|---|---|---|---|
| push cell, before | inlined | inlined | 685 B |
| + elision fix | elided | inlined | 600 B |
| + schema-less document | elided | absent | **87 B** |

So neither half alone gets there, which is why your two suggestions were both load-bearing.

**The missing base was a bug, as you said.** Two things support that beyond the symptom: `normalizeAndDiff()` already has the destination `link` in scope at that point, and the sibling branch ~100 lines up already does `createSigilLinkFromParsedLink(scopedLink, { base: link })`. One branch elides, the other never did.

**It hid a second one of the same species.** The self-reference no-op guard at what was `data-updating.ts:718` also compares base-lessly. Once the written link elides, that guard stops recognizing a cell written to its own slot — so `c.set(c)` stopped no-op-ing and resolved into `Link cycle detected`. Passing the base fixes it, but it is worth noting the general shape: *any* comparison of a raw stored sigil that omits a base is now suspect. Four test assertions were in exactly that position. I checked the workspace `areLinksSame` call sites; the nine in `html/worker/reconciler.ts` compare `Cell`s rather than stored values, so they are safe by construction — but I would not claim that grep was exhaustive.

**Verification.** Full workspace suite green, plus `fmt`/`lint`/`check-docs`. Separately, since a refactor can quietly blind a regression test, I re-checked that `llm-dialog-message-scope.test.ts` still fails when the three scope arguments are neutered — it does. And I diffed the entire persisted `labelMap` for the message documents before and after; byte-identical, so the schema-less document does not change what the persist-time `LlmDerived` gate records.

On your point about link-writing in user-land being an unclosed problem: agreed that this is not the PR to close it, and the cast I was using is a small instance of exactly that capability. It is gone now, which at least does not add to the surface while FabricLink is pending.

— Claude (Opus 5), working with @danfuzz

danfuzz:
**Correction to my comment above** (@seefeldb): I wrote that four assertions needed a base and "nothing else in the workspace moved." That was wrong — there is a fifth site, and CI caught what my local gate did not. Fixed in `65bf6f044`.

`packages/runner/integration/sync-schema-path.test.ts` deliberately reaches past the sync system to inspect the raw stored link:

```ts
const sigilLink = JSON.parse(JSON.stringify(newCell.key(0).getRaw()));
const normalizedLink = parseLink(sigilLink) as NormalizedLink;  // no base
const record = read(storageManager, normalizedLink.space!, normalizedLink.id!);
```

Once the link elides, `normalizedLink.space` is `undefined` and the heap read misses. Same species as the other four, and as the self-reference guard.

Two things worth recording, since the mistake was in my method rather than the change:

**The package integration suites do not run under the workspace `deno task test`.** Six packages have a separate `integration` task (`cli`, `generated-patterns`, `patterns`, `runner`, `runtime-client`, `shell`). I ran the workspace suite, `fmt`, `lint` and `check-docs`, called that a full gate, and it was not one. Every other integration suite passes, so the real exposure was this single site — but I could not have known that from what I had run.

**My grep was too narrow.** I searched for `areLinksSame(` / `areMaybeLinkAndNormalizedLinkSame(` and concluded the audit was done. The dangerous pattern is broader: any `parseLink()` of a *raw stored* value with no base, whose result is then used for `.space`. Redoing it that way finds two such sites in `src` — `llm-dialog.ts` (the `LlmDerived` stamping loop) and `sqlite-builtins.ts` (per-row labels) — both of which happen to be safe because they already fall back to `link.space ?? base.space`. The base-less parses in tests only read `.id`, which elision preserves.

So the invariant to state plainly, for anyone touching this later: **a stored link is only meaningful together with the slot it was read from.** Parsing one without its base is now a bug even when it happens to work.

— Claude (Opus 5), working with @danfuzz

PR #5216, opened 2026-07-30
fix(runner): abandon a tail op whose array was reshaped before it

seefeldb:
## The bug

`rows.set([])` followed by `addUnique` in one handler **silently doubled the list**: the 20 original members stayed and 20 new ones were appended. The writing session's own `cell.get()` showed the correct 20 the whole time, so nothing looked wrong until the durable value was read back.

Found while spiking whether the mappa mundi's content can move from source into an editable `PerSpace` cell — the clear-and-reseed idiom used to convert bulk-seeded records into keyed entities.

## Mechanism

A mergeable tail op (`push` / `addUnique`) commits as *"add these elements to whatever the durable array is"*, and suppresses every whole-array diff candidate at or past `start = working.length - count`. That suppression is only the transaction's local truth while the working array's prefix still lines up with the base one element for one — i.e. while `base.length === start`.

`set([])` + 20 `addUnique` gives `working.length = 20`, `count = 20`, so `start = 0`: the suppression covers the *entire* diff, including the removals carrying the clear. The commit sent nothing but the add-unique, and the store kept the old members and appended the new ones on top.

## Why the existing fallback missed it

`poisonMergeableOp` already exists for exactly this class of problem, but all three of its call sites fire **at the moment of a write**, so they see only a reshape landing *after* an op, at *exactly* the op's path. Two shapes slip past, both corrupting:

- reshape **before** the op — `rows.set([])` then `addUnique` (clear-and-reseed);
- reshape at a **parent** path — `doc.set({rows})` then `doc.key("rows").push(x)`, since poisoning keys on `encodePointer(address.path)`.

A stickier poison in `Cell.set` would fix only the first.

## The fix

`buildTailOp` re-checks its invariant at commit and **abandons** the intent when the base length no longer equals the recorded prefix, falling back to the whole-array diff (the correct local value). Deciding this in the builder rather than at the write site is what covers the parent-path case too.

Abandoning **deletes** the intent from the transaction rather than just skipping the op: a live intent still narrows the op's reads out of the commit's conflict set (in `v2.ts`, consumed inside `commitNative` — after `getNativeCommit` builds patches), and the replacing whole-value diff needs those reads back. That ordering is now stated where it is relied on.

The affected transactions forfeit merge-friendliness for that list, which is the right trade: replacing a list wholesale is not an operation that should merge with a concurrent append.

## Tests

Three cases in `array-push-mergeable.test.ts`, each verified red against the unfixed source and green with it — shrinking `set` + `push`, `set([])` + `addUnique`, and the parent-object variant. Plus direct builder-guard coverage in `mergeable-op-registry.test.ts`.

One trap worth flagging for anyone writing a similar repro: reseeding with values *already in the list* passes for the wrong reason, because the server re-dedups `add-unique` against durable state and masks the lost clear. The reported case only surfaced it because the entity links differed from the bulk-seeded inline objects.

## Docs

- `mergeable-collection-writes.md` — new bullet on the length invariant and why the poison sites cannot see these two shapes; the old line claiming "a reshape or `set` *before* any op leaves a later push mergeable" described the hole as a benign optimization.
- `patch-operations.md` — the commit step can abandon an intent.
- `keyed-collection-writes.md` — clear-and-reseed commits as a plain overwrite.

## Verification

Full runner suite (1097 passed / 0 failed) and memory suite (456 / 0) green after a catch-up merge of `main`; repo-wide `deno fmt --check` and `deno lint` clean.

---
## Summary by cubic
Prevents silent list duplication and stale tail-op commits after reshapes, including sparse payloads when no base array exists. Tail ops now abandon at commit when the prefix length or hole layout changed, the payload is sparse (with or without a base), or the array is missing; whole-value and ancestor writes poison descendant intents so we fall back to a whole-array diff.

- **Bug Fixes**
  - In `packages/runner/src/storage/mergeable-ops.ts`, tail-op builder abandons on: base/prefix length mismatch, prefix hole-layout change, sparse payload even without a base, empty tail, or missing array; net-zero `increment` emits no op (not abandoned).
  - In `packages/runner/src/storage/v2-transaction.ts`, `poisonMergeableOp` now reaches intents at or beneath the written path (not above); `buildMergeableOps` drops abandoned intents and poisons those paths inside `getNativeCommit`.
  - In `packages/runner/src/cell.ts` and `packages/runner/src/query-result-proxy.ts`, whole-value writes via `set`, `setRawUntyped`, in-place mutators, and property assignment poison descendant tail intents; element edits and sibling writes remain mergeable.
  - In `packages/runner/src/storage/transaction-inspection.ts`, added `getDirectTransactionNativeCommit` to inspect commit-time intent decisions in tests.
  - Docs/tests updated for hole-layout changes, ancestor poisoning, proxy assignment poisoning, sparse-without-base payloads, and abandon cleanup.

<sup>Written for commit a6a91d056e26975760720693dac7c0f04c2cc2e2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5216?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seefeldb:
Both P1s investigated. One reproduced and is fixed; one does not reproduce and I've added a test pinning the behaviour.

## @chatgpt-codex-connector — confirmed and fixed

The ancestor-reshape sequence is real. `poisonMergeableOp` keyed on `encodePointer(address.path)`, so a write to an enclosing object never reached the intent for a list inside it — and with `count` spanning the reshape, the length check lands the tail back on the base length while covering an element the reshape supplied. Exactly as described.

Fixed in 2167e393: `poisonMergeableOp` now covers every intent at *or beneath* the path written, and nothing above it — so an element edit (a write *beneath* the array) and a write to a sibling field both still leave a push mergeable. Both directions are pinned by tests: `an ancestor write poisons the tail intent beneath it` (red before the change) and `a sibling write leaves the tail intent intact` (guards against over-poisoning).

One note on how I tested it, since it nearly fooled me: a commit-level assertion passes whether or not the intent survives, because the reshaping `set` records reads of its own that conflict independently. The test asserts on the recorded intents instead.

You also flagged that the fallback guarantee in the live docs was inaccurate — correct, and it was worse than that. `migrating-collection-writes.md` said such a transaction "may drop the reshape"; it is the *append* that is dropped, never the reshape. Fixed, along with the section anchor it cited, which did not exist.

## @cubic-dev-ai — does not reproduce

A same-length replacement before a tail op does **not** lose the replacement. The reasoning looks right — `isSuppressed`'s first branch does drop a candidate whose path equals the array path — but the diff doesn't produce that whole-array cover candidate here. It emits per-index candidates, which sit below `start` and survive suppression, so the replacement commits alongside the append.

Verified across four shapes, all green before and after: same-length string replacement, same-length object replacement, `set` + `addUnique`, and a parent-object set with a nested push. Added `a same-length whole-array set then a push commits the replaced list` to keep it that way, and the doc now states it explicitly, since "length equality is sufficient here" is the non-obvious part.

Tightening the guard to full prefix comparison would also have a real cost: it would knock out element-edit-plus-push, which is a supported composition and genuinely valuable. That case is now pinned by `an element edit then a push keeps the push mergeable and commits both`.

seefeldb:
@codex review

seefeldb:
Round-2 self-review addressed in a18a4005f. Everything raised is either fixed or answered below.

## Fixed

**Stale contracts.** Three texts still specified exact-path poisoning after the rule widened: `IStorageTransaction.poisonMergeableOp`'s JSDoc (which `ITransactionWriter` delegates to), the call site in `Cell.set`, and the query-result proxy's equivalent. All corrected. The `Cell.set` one was the worst of them — it's the first place anyone lands asking "when does my push stop merging", and its "a set that lands on a different slot finds no intent there" is true beneath an array and for a sibling, false above it.

**A test rationale that was simply wrong.** My comment on the same-length-replacement test claimed the replacement diffs as a whole-array cover candidate that suppression drops. It doesn't — it diffs into per-index candidates below the tail, which is what the doc already said. So the changeset shipped two opposite accounts of one mechanism and the test comment was the wrong one. Rewritten; the test stays as characterization, and now says that it holds on the unfixed code too.

**Two tests that couldn't observe what they claimed.** The element-edit composition ordered the edit *before* the push, so no intent existed when the poison fired — widening the poison to both directions left it green. Reordered to push-then-edit, and paired with a new intent-level test, because the durable value can't discriminate here either: a concurrent append is add-wins and lands whether or not the push stayed mergeable.

Every guard is now mutation-checked, each reddening exactly the test that claims it:

| mutation | reds |
|---|---|
| poison reverted to exact-path | `an ancestor write poisons the tail intent beneath it` |
| poison widened to `pathsOverlap` | `a write beneath an array leaves the array's intent intact` |
| length check removed | 2 tests (4 steps) |
| abandon delete-loop removed | `a build-time abandon removes the intent from the transaction` |

**Nits:** reuse the map key instead of re-encoding the pointer; the test's ad-hoc transaction unwrap moved into `transaction-inspection.ts` as `getDirectTransactionNativeCommit`.

## Two verified gaps, documented and filed — not fixed here

The review flagged that "never silently corrupted" is an overclaim. It's right, and I reproduced both cases. Neither is a regression; neither is caught by this PR's checks.

**`removeByValue` alongside any other change to the same array.** Its `subtree: true` suppression drops every candidate for the array, and unlike the tail ops it has no commit-time honesty check. From a durable `["a","b","c"]`:

- `key(0).set("A")` then `removeByValue("c")` → local `["A","b"]`, durable `["a","b"]`
- reverse order → same loss
- `set(["p","q"])` then `removeByValue("p")` → local `["q"]`, durable `["a","b","c"]`

The first is the ordinary edit-one-row-delete-another handler. Worth noting the *deliberate* choice not to poison beneath a write is what lets it through.

**Nested tail ops.** Push a new array onto an outer array, then push into that inner array: local `[["a"],["x","y"]]`, durable `[["a"],["x","y","x","y"]]`. The outer append's payload is `array.slice(start)`, which already contains the element the inner append adds again.

I've documented both under a "Known gaps" heading in `mergeable-collection-writes.md` and softened the guarantee sentence to point at it, rather than fixing them here — each needs its own mechanism and review, and this PR is already reviewed and green.

## Answering the open question

> is the `Math.max(0, …)` clamp still reachable in a non-abandoning case?

No — it is dead in the non-abandoning path. `count > working.length` requires the transaction to have shrunk the array below the recorded tail, and any such shrink makes `initialArrayLength !== start` (since `start` clamps to 0 while the base was non-empty), so the guard abandons first. I've left the clamp in place because it keeps `slice` total independent of the guard, but you're right that it no longer implies a live case.

**Blast radius**, confirmed independently: 0 of ~169 Cell-mutator call sites across patterns/home-schemas/runner/piece/shell/toolshed/ui pair with a genuine ancestor `.set()` in the same handler. The `elementById` idiom is structurally immune — it mints a separate entity document, and poisoning is per-document.

Verification: runner 1099 passed / 0 failed, memory 456 / 0, `deno fmt --check` and `deno lint` clean, after a catch-up merge of `main`.

seefeldb:
@codex review

seefeldb:
@chatgpt-codex-connector — confirmed and fixed in f85203719.

The proxy's `set` trap called `diffAndUpdate` and nothing else, so assigning over a property never poisoned. Reproduced at intent level (`a proxy property assignment poisons the tail intent it overwrites`, red before the fix): `proxy.rows.push("p")`, `proxy.rows = ["m","n","o","r"]`, `proxy.rows.push("q")` left a live intent whose recorded tail spanned the assignment.

This is the third whole-value write path fixed one at a time — `Cell.set`, then the proxy's in-place mutators, now the assignment trap. Rather than wait for a fourth, I enumerated every whole-value write reachable through `Cell` and stated the rule the sites implement: **a whole-value write poisons the ops it covers.**

`setRawUntyped` now poisons too. Its callers are internal machinery writing links into result cells where no op is ever recorded, so it is inert today — it's there so the rule holds rather than being a list of paths that happen to remember, and the comment says so rather than implying a live case.

Checked and deliberately left alone, with reasons:

| site | why it needs no poison |
|---|---|
| `query-result-proxy.ts:487` | writes a freshly hashed result document — no intent can exist there |
| `runner.ts:1222` | setup argument write, already preceded by `Cell.set` |
| `pattern-binding.ts:292,336` | binding-time output writes, not handler transactions |
| `Cell.update`'s `{}` init | only runs after a read proved the value absent |

The doc's poison-site list is now framed as that rule plus the paths implementing it, so the next person adding a write path has something to check against.

Verification: runner 1100 passed / 0 failed, memory 456 / 0, `deno fmt --check` and `deno lint` clean, after a catch-up merge of `main`.

seefeldb:
@codex review

seefeldb:
@chatgpt-codex-connector — the sparse same-length case is confirmed and fixed in aab4264d0. My apologies for the delay on this one: I filtered the inline comments by timestamp when checking that review round and my window excluded it, so I reported that round as clean when it wasn't.

Reproduced, and worse than described. From a durable `["a","b","c"]`:

```
set([<hole>,"b","c"]) then push("d")
  local   [<hole>,"b","c","d"]
  durable ["a","b","c","d"]      ← hole gone, "a" resurrected

base [<hole>,"b","c"], set(["A","b","c"]) then push("d")
  local   ["A","b","c","d"]
  durable [<hole>,"b","c","d"]   ← the fill AND "A" gone
```

The second is the sharper one: because the whole-array candidate carried *every* element, suppressing it discards the entire replacement, not just the presence change.

Your diagnosis of the mechanism was exact. Rather than special-case holes, I matched the guard to `buildArrayPatchCandidates`'s own fallback conditions, since those are precisely when the surviving per-index candidates cannot carry the prefix:

1. the prefix changed length;
2. the prefix's hole layout changed;
3. the appended tail is itself sparse.

The build context now carries the base array rather than just its length. Both new conditions are mutation-checked — stubbing out the hole-layout check reds 2 tests, the sparse-tail check 1.

Coverage added: `a set that punches a hole then a push commits the hole` and `a set that fills a hole then a push commits the filled value` end to end, plus two builder-level cases in `mergeable-op-registry.test.ts`.

You also flagged that this invalidated the new documentation claim. It did, and the same wrong claim was in a test comment. Both now say that only a *dense* same-length replacement stays mergeable — value changes are fine, length and presence changes are not.

Verification: runner 1100 passed / 0 failed, memory 456 / 0, `deno fmt --check` and `deno lint` clean, after a catch-up merge of `main`.

seefeldb:
@cubic-dev-ai — revisiting your P1, because you were more right than my first reply gave you credit for.

You wrote: *"length equality treats a changed prefix as mergeable... Compare the working prefix with the initial array (not only its length)."*

I checked the value-replacement case you named and it genuinely doesn't reproduce — a dense same-length replacement diffs into per-index candidates that survive the suppression. That part stands.

But the general claim — that length equality is not a sufficient check on the prefix — was correct, and codex later found the case that proves it: a same-length `set` that changes the array's **hole layout**. There `buildArrayPatchCandidates` does fall back to the whole-array candidate you predicted, and the suppression does drop it, exactly as you described. Measured: `set([<hole>,"b","c"])` then `push("d")` gave a local `[<hole>,"b","c","d"]` and a durable `["a","b","c","d"]`.

Fixed in aab4264d0. The guard now compares the prefix's presence layout, not just its length — the "compare the prefix, not only its length" shape you asked for, narrowed to presence rather than values so that a dense value replacement stays mergeable.

Marking your thread resolved.

seefeldb:
@codex review

seefeldb:
@chatgpt-codex-connector — confirmed and fixed in a6a91d056.

You were right that the density check had to run regardless of `hadInitialArray`. It is the one condition of the three that needs no base: the payload is `array.slice(start)` either way, and with no base that slice is the whole working array, so a hole anywhere in a freshly created sparse array is a hole in the payload.

One correction to the predicted symptom — it is worse than holes materializing as own `undefined` elements. Measured on a previously absent cell:

```
set([<hole>,"b","c"]) then push("d")
  local   [<hole>,"b","c","d"]   (0 in local === false)
  durable undefined              ← the entire document fails to land
```

The dense equivalent commits fine, so it is specifically the sparse payload. After the fix the durable value is `[<hole>,"b","c","d"]` with the hole intact.

The density loop moved out of the `hadInitialArray` branch; the two comparisons that genuinely need a base stayed inside it. Mutation-checked: putting the loop back inside the branch reds 2 tests. Coverage added at both levels — `an absent cell set to a sparse array then pushed commits the hole` end to end, and `a tail op with no base and a sparse payload abandons the intent` at the builder.

Verification: runner 1098 passed / 0 failed, memory 456 / 0, `deno fmt --check` and `deno lint` clean.

seefeldb:
@codex review

PR #5218, opened 2026-07-30
Remove `ID_FIELD` and its only producer

danfuzz:
First step of retiring the write-directive symbols (`ID` + `ID_FIELD`), which is what blocks tightening the runtime's `FabricPlainObject` recognition: the directives are the one sanctioned way a symbol-keyed plain object flows through the write path, so the tightening cannot be honest while they exist.

## What `ID_FIELD` was

A write-time annotation naming a sibling property to derive list-item identity from. `normalizeAndDiff()` matched that field's value against existing sibling documents and reused the match (DOM-diff style), falling back to a fresh random id.

## Why it can simply go

Its only producer in the tree was `addCommonIDfromObjectID()` — the bridge that stamped `id`-bearing objects when whole JSON documents were shipped in from iframes. That regime no longer exists; the bridge had zero callers outside its own tests, and no pattern, package, or transformer references `ID_FIELD` anywhere.

## What's removed

- `data-updating.ts`: the `BRANCH_ID_FIELD` interpreter branch and `addCommonIDfromObjectID()`.
- `@commonfabric/api`: the `ID_FIELD` symbol declaration and its slots in `AnyCellWrapping`, `IDFields`, and the JSONSchema extension keys.
- Builder surface: the runtime symbol (`builder/types.ts`), the pattern-author exposure (`factory.ts`, `CommonFabricTypes`), and the runner index exports.
- `cell.ts`: the `ID_FIELD` half of the directive carry-through in the write walk (the known-gap comment there narrows accordingly).
- One spec line in `ts_to_json_schema_mapping.md` updated to match the api surface.

## Tests

Mechanism tests (`ID_FIELD` redirect reuse, `addCommonIDfromObjectID`) are deleted with the mechanism — dead coverage once the branch is gone. The cell-as-cell "ids update when context changes" test kept its guard by dropping the bridge calls: plain array anchoring under a frame provides the same document splitting, and that behavior — not the directive — is what the test pins.

## Verification

Repo-wide grep sweep clean (no `ID_FIELD`/`addCommonIDfromObjectID` outside `docs/history/` and the defunct `patterns/deprecated`). `deno task check` green, `packages/api` checked directly, repo-wide `deno fmt --check` + `deno lint` clean, `deno task check-docs` 523/523, runner suite 1099 passed / 0 failed, full workspace suite green at this commit.

## What follows

`ID` remains, with its three internal producers (all funneling through `recursivelyAddIDIfNeeded`). Internalizing it — anchoring array objects directly inside the diff so no symbol-keyed plain object crosses the write path at all — is the next, separate PR; removing the `ID` public surface is the one after.

https://claude.ai/code/session_019VEMbFKHAMQ5h71AD5uFj8

---
## Summary by cubic
Removes the `ID_FIELD` write directive and its only producer `addCommonIDfromObjectID`, simplifying the write path and enabling tighter `FabricPlainObject` checks. Also pins reactive proxy behavior so unknown symbol reads delegate to the target.

- **Refactors**
  - Removed `ID_FIELD` handling from `normalizeAndDiff()` and the carry-through in `cell.ts`; deleted `addCommonIDfromObjectID()` and runner exports.
  - Dropped `ID_FIELD` from `@commonfabric/api` (`AnyCellWrapping`, `IDFields`, `JSONSchemaObj`) and from builder/runtime constants and exports.
  - Updated schema docs to reference only `[ID]`.
  - Added a unit test pinning that reactive proxies delegate unrecognized symbol reads to the target.

- **Migration**
  - Remove any use of `addCommonIDfromObjectID()` and `[ID_FIELD]`.
  - If you depended on ID-field-based reuse, pass `[ID]` directly or use stable links; no `JSONSchema` key for `ID_FIELD` remains.

<sup>Written for commit 217dd4f5b45dea10c227348f373235ef2e831c69. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5218?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/schema-generator/ts_to_json_schema_mapping.md
+++ b/docs/specs/schema-generator/ts_to_json_schema_mapping.md
@@ -604,8 +604,8 @@ The emitted key set aligns with the api's `JSONSchemaObj.ifc` member
 `writeAuthorizedBy`, `exactCopyOf`, `projection`, `observes`, and `uiContract`.
 `ownerPrincipal` and `observes` have no direct producing alias in this package
 as of this writing.
-The api type also declares `[ID]`/`[ID_FIELD]` extension keys
-(`api/index.ts`); this package never emits them (grep).
+The api type also declares an `[ID]` extension key
+(`api/index.ts`); this package never emits it (grep).
 
 ## 12. Doc Comments → `description` / `tags` / `$comment`
 
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -330,7 +330,6 @@ export interface FabricExecPlainObject
 // Runtime constants - defined by @commonfabric/runner/src/builder/types.ts
 // These are ambient declarations since the actual values are provided by the runtime environment
 export declare const ID: unique symbol;
-export declare const ID_FIELD: unique symbol;
 
 // Should be Symbol("UI") or so, but this makes repeat() use these when
 // iterating over patterns.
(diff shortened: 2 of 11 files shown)


ubik2 (approved):
nice!

PR #5221, opened 2026-07-30
Anchor array elements inside the diff, not via `[ID]` annotation

danfuzz:
Second step of retiring the write-directive symbols (after #5218 removed `ID_FIELD`): the `[ID]` auto-annotation machinery is gone. No plain object with a symbol key is ever materialized on the write path anymore, which is the property the `FabricPlainObject` tightening needs.

## What changed

Writers used to pre-annotate: `recursivelyAddIDIfNeeded()` walked the whole value, rebuilt it with `[ID]: frame.generatedIdCounter++` on every plain object in an array, and `normalizeAndDiff()` interpreted the annotation back off. That directive was a same-call-stack communication channel — nothing else ever saw it.

Now the walk anchors directly:

- `normalizeAndDiff()` threads a **`DiffWalkState`** — the old `seen` map plus an optional `nextAnchorId` source — and anchors a plain object arriving as an array element via the extracted **`anchorValueAsEntity()`** helper, the same code the explicit-`[ID]` branch uses (that branch stays until the `ID` surface itself is retired).
- Writers — `Cell.set`/`push`/`addUnique` and the query-result proxy's array mutators — pass **`frameAnchorIds(frame)`**. No frame → no source → inline storage, as before.
- The ~200-line pre-pass (an entire extra rebuild-and-freeze walk per write) is deleted.

## Divergences from the annotation scheme

Two independent reviews (a cf-review subagent and an external agent review) stress-tested the original "ids are unchanged" claim and proved it wrong. The honest statement: **flat arrays of fresh objects get identical ids; several other shapes diverge**, all deterministically within the new scheme, all in the same class the #5208 precedent accepted (ids are opaque; nothing re-derives them across the boundary). All but the `FabricInstance` case are pinned by dedicated tests so the next change here is conscious:

- **Draw order is pre-order, not post-order.** The annotation walk stamped children before their containing element; the inline anchor draws the element's id before recursing into its content. Any nested anchored data (items with sub-lists) mints different ids for both documents. *(pinned: "draws anchor ids pre-order")*
- **Repeated references converge on one document.** The annotation scheme stamped a fresh copy per array occurrence — `[shared, shared]` stored two documents and drew two ids. The diff's `seen` map now preserves the written graph's aliasing: one document, both slots linked to it, one id drawn. When the first occurrence was an inline property, the shared object is **promoted** into an entity document and both occurrences repoint to it (review feedback from @ubik2's thread: an array link into a document's mutable interior would dangle when the inline property is later removed — promotion keeps the aliasing @ubik2 wanted on a stable target). A self-referencing element stores a clean self-link instead of a nested copied cycle. This is a deliberate semantic choice, not just an id change: the stored graph now matches the object graph the author wrote. *(pinned: "converges repeated references", "promotes an inline-then-array shared object", "keeps an array occurrence alive", "persists ... self-reference")*
- **`addUnique` no longer draws ids for filtered-out duplicates** (it used to anchor all candidates before deduplicating), and its dedup comparison normalizes candidates to fabric form first — comparison-only, so an accepted add still writes the original value and no identity semantics shift. *(pinned: "draws no anchor id for an addUnique candidate", "normalizes addUnique candidates")*
- **`FabricInstance` internals no longer advance the counter** (the pre-pass walked encoded state for that side effect; the walk is gone). Disclosed but not separately tested: exercising it needs instance state carrying plain-object arrays, which no supported write constructs.
- **Fresh nested arrays keep index-bearing ids** — a pre-existing limitation (identical on `main`; the ancestry loop reads transaction pre-state), inherited verbatim and now stated honestly in `anchorValueAsEntity`'s JSDoc rather than contradicted by it.

### The concurrency guard (found by a follow-up review)

An earlier revision of this branch anchored *existing inline elements* whenever a framed write carried them through a whole-array rewrite. A follow-up review proved that unsafe for the mergeable collection ops: `push`/`addUnique` exclude their base-array read from the commit's conflict set (see `mergeable-collection-writes.md`), which is only sound while the op writes nothing below the tail — anchoring the prefix emitted positional replacements a stale session could commit over a concurrent element edit without conflict, silently losing it (verified two-session, against `main` and this branch).

The anchor branch therefore carries an **identity guard**: an element carried through unchanged from the transaction's own read of the stored array arrives as the exact stored (frozen) object, so `Object.is` against the slot's current value is the precise "untouched prefix" test — and such elements are never anchored or rewritten. Identity, deliberately not content equality: the written value is only shallowly normalized at that point (a follow-up review caught the content-hash variant throwing on elements with nested Cells awaiting conversion — now its own regression test). This restores `addUnique`'s candidates-only write-set, and (a strict improvement over `main`) extends the same protection to `push`, whose annotation-era pre-pass anchored the whole combined array and had the identical latent hazard. Pinned two ways: the unit test asserts the untouched prefix and single id draw, and a two-session durable test in `array-push-mergeable.test.ts` commits a stale framed `addUnique` over a concurrent element edit and asserts both survive — red without the guard (reproducing the lost update), green with it.

Eligibility comes from the walk structure (an `isArrayElement` hop threaded by the array branch and the value-preserving re-entries), **not** from reading the stored parent — which does not exist yet on a fresh array's first write. The two value-*replacing* re-entries (data-URI inlining, same-doc snapshot collapse) deliberately do not forward eligibility, matching the annotation scheme, which never looked behind links; both sites now say so.

## Two identities the pre-pass silently aligned

Converting *before* the diff meant every `seen` registration matched caller identities. Doing conversion inside the diff surfaced two things the first draft got wrong, both caught by the existing suite:

1. **Shared references and cycles arrive under the pre-conversion identity**, so that is what `state.seen` registers (dual registration when conversion produced a copy). Otherwise a cycle re-anchors instead of converging.
2. **The anchored content must be a distinct object from the registered one** — the recursion that writes the entity document would otherwise find the value in `seen` and no-op as a self-reference, leaving the document permanently unwritten. This is the same distinctness the `[ID]` branch always got for free from its destructuring spread.

## Tests

The pre-pass's three direct unit tests go with it. The behaviors it served — objects in arrays split into documents, sparse holes preserved, shared references converge on one document, ids re-derive per frame context, circular references become links — are all pinned by existing tests that drive the write paths themselves (`cell-as-cell`, `cell-core`, `array-push-id-bug`, `data-updating`, `nested_cell_array`, the CT-1173 proxy-push suite).

## Verification

Full workspace suite green at every pushed state of this branch (runner alone: 1099+ passed / 0 failed); repo-wide `deno task check`, `deno task check-docs` (523/523), `deno fmt --check`, and `deno lint` all clean. The divergence disclosures above were each verified red/green or empirically probed during review.

## What follows

The explicit-`[ID]` branch, the `ID` symbol's public surface (api, builder exposure, exports), and the tests that hand-attach `[ID]` are the last piece; retiring them is the next, separate PR — after which the directive family is gone entirely.

https://claude.ai/code/session_019VEMbFKHAMQ5h71AD5uFj8


packages/runner/src/data-updating.ts
```diff
@@ -396,13 +418,97 @@ export type ChangeSet = {
...
+): ChangeSet {
+  let path = link.path;
+
+  // If we're setting an array element, make the array the context for the
+  // derived id, not the array index. If it's a nested array, take the parent
+  // array as context, recursively.
+  while (
+    path.length > 0 &&
+    Array.isArray(
+      tx.readValueOrThrow({ ...link, path: path.slice(0, -1) }, options),
```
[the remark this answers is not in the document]
danfuzz: Verified, with a narrower conclusion: the behavior is real but faithfully preserved, not introduced by this PR.

The ancestry loop in `anchorValueAsEntity()` is a verbatim extraction of the one `BRANCH_ID_OBJECT` has always used, and the pre-pass auto path fed the very same loop — so a fresh nested array's first write derived index-bearing ids before this change exactly as it does after it. `push()`/`addUnique()` create the array in a separate operation first ("so that in the next steps each object element is properly anchored") precisely to make the stored-state read see an array, so the append paths are unaffected; the surface is `set()` of a fresh document with nested arrays.

Deriving ancestry from the written tree instead would be a real improvement, but it changes the ids minted on fresh-array writes — deliberately out of scope for a PR whose contract is "anchored ids match the annotation scheme." Worth a follow-up once the `ID` surface retirement (the next step) lands, since that also touches this helper.

— cf session agent (Claude Fable 5), on behalf of danfuzz

packages/runner/src/data-updating.ts
```diff
@@ -448,13 +565,13 @@ export function normalizeAndDiff(
...
   // which below will be turned into a relative link.
-  if (seen.has(newValue)) {
+  if (state.seen.has(newValue)) {
     diffLogger.debug(
       "diff",
       () =>
         `[SEEN_CHECK] Already seen object at path=${pathStr}, converting to cell`,
     );
-    newValue = new CellImpl(runtime, tx, seen.get(newValue)!);
+    newValue = new CellImpl(runtime, tx, state.seen.get(newValue)!);
```
ubik2: **[Major · verified] Keep array aliases off ephemeral inline paths**

This `seen` short-circuit runs before array-element anchoring. When the same object first appears inline and later in an array, the array gets a relative link to the inline location. Removing that earlier property then leaves the array link dangling: `list[0]` resolves to `undefined`, even though the array still references the object. This focused test fails on the current head and passes on the PR base:

```ts
it("keeps an array occurrence alive after its earlier inline alias is removed", () => {
  const cell = runtime.getCell<any>(space, "inline alias lifetime", undefined, tx);
  const shared = { name: "Ada" };

  cell.set({ first: shared, list: [shared] });
  expect(cell.get().list[0]).toEqual({ name: "Ada" });

  const raw = cell.getRaw() as { list: unknown[] };
  cell.set({ list: raw.list });
  expect(cell.get().list[0]).toEqual({ name: "Ada" });
});
```

Please avoid reusing a seen non-root inline location for an array occurrence. Anchor that occurrence into a stable entity document, or promote all occurrences to one stable entity, and keep this lifetime case as the regression test.

ubik2 (commented):
_cf-review via Codex (GPT-5), on behalf of @ubik2._

Reviewed the current head (`9d2cb62e`). The latest commit fixes the stale `addUnique` prefix rewrite identified in the earlier pass, so I have not posted that stale finding. Two verified blocking regressions remain; both focused tests fail on this head and pass on the PR base.

packages/runner/src/cell.ts
```diff
@@ -1638,14 +1636,10 @@ export class CellImpl<T extends FabricValue>
...
-    // server re-dedups against durable state, catching elements the local
-    // replica had not loaded.
-    const anchored = recursivelyAddIDIfNeeded(
-      value as FabricValue[],
-      this._frame,
-    );
+    // Keep only the values not already present (by stored-value equality,
+    // matching the server's add-unique dedup). The server re-dedups against
+    // durable state, catching elements the local replica had not loaded.
+    const candidates = value as FabricValue[];
```
ubik2, line 1644: **[Major · verified] Normalize `addUnique` candidates before stored-value comparison**

The deleted pre-pass normalized candidates before `valueEqual`; this cast now sends supported native write-boundary values through unnormalized. A repeated `Date` throws `Cannot compare value "1970-01-01T00:00:01.234Z"` instead of being a no-op. This focused test fails on the current head and passes on the PR base:

```ts
it("normalizes addUnique candidates before stored-value comparison", () => {
  const cell = runtime.getCell<any[]>(space, "addUnique normalization", undefined, tx);

  cell.set([new Date(1234)]);
  expect(cell.getRaw()).toEqual([
    new FabricEpochNsec(1_234_000_000n),
  ]);

  cell.addUnique(new Date(1234));
  expect(cell.getRaw()).toEqual([
    new FabricEpochNsec(1_234_000_000n),
  ]);
});
```

Please normalize candidates before deduplication without consuming anchor IDs for candidates rejected as duplicates. A `toJSON()`-backed value would be a useful neighboring case.

ubik2: Not actually sure if Sol is completely right here, since you may also be doing object identity comparison (which normalizing would break).

packages/runner/src/data-updating.ts
ubik2: This one is interesting. I really do want the preservation of `first` and `list[0]` being the same object. I'm not sure if there's a simple path that gets you there. I take Sol's suggestion as a punt.

danfuzz: Fixed via promotion, keeping the aliasing: an anchor-eligible array occurrence that hits a `seen` entry at a non-root (inline) location now anchors the shared object into an entity document of its own and repoints the earlier inline location at it too — so `first` and `list[0]` remain the same object, on a stable root target rather than a fragile interior link. A root `seen` location (an already-anchored occurrence) still takes the plain link, and an element carried through untouched from the stored array emits nothing, preserving the mergeable-op prefix invariant.

Your lifetime test is in as the regression guard (rewriting the container without `first` leaves `list[0]` intact), and the pinning test for this shape now asserts the promoted form: one id drawn, both occurrences linking to the same document root.

— cf session agent (Claude Fable 5), on behalf of danfuzz

packages/runner/src/cell.ts
danfuzz: Fixed with comparison-only normalization, which also answers the identity concern raised in this thread: the dedup now compares against a fabric-normalized *copy* of the candidate (`fabricFromNativeValue`), but an accepted candidate writes the **original** value — no object the write path sees is replaced, so the untouched-prefix identity guard and any caller-side identity expectations are undisturbed. A rejected duplicate still draws no anchor id.

Two scoping refinements the full suite forced, both restoring exact pre-pass parity: normalization applies only under a frame (the deleted pre-pass was frame-gated; frameless `addUnique` always compared raw), and link-carrying candidates — query-result proxies, raw sigil links — compare as themselves (the pre-pass's `isCellLink` pass-through; strict conversion would reject their annotated internals, which is precisely what the existing "a plain object read back from get() is added again" test caught).

Your `Date` test is in (plus the suggested `toJSON`-backed neighbor). One boundary note: a non-link candidate with a *nested Cell* threw during the old comparison as well (`hashOf` on the cell) — it now throws during normalization instead; same broken class as before, not a new regression, flagging for completeness.

— cf session agent (Claude Fable 5), on behalf of danfuzz

packages/runner/src/data-updating.ts
```diff
@@ -965,6 +1056,54 @@ export function normalizeAndDiff(
...
+  // unchanged inline prefix here would let a stale session clobber a
+  // concurrent element edit without conflict.
+  if (
+    state.nextAnchorId !== undefined &&
+    isArrayElement &&
+    isObject(newValue) &&
+    !(newValue instanceof FabricSpecialObject) &&
+    !isCellLink(newValue) &&
+    !(currentValue !== undefined &&
+      valueEqual(currentValue as FabricValue, newValue as FabricValue))
```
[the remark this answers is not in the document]
danfuzz: Resolved as of `5da169a60`, which this comment's file context predates: the content-equality guard (`valueEqual`) it targets was replaced by an `Object.is` identity check against the slot's stored value — no recursive hashing runs in the guard at all, so cyclic (or otherwise unnormalized) written content cannot overflow there. Cyclic content then anchors normally and the cycle persists as a link (pinned by the "persists an array element's self-reference as a self-link" test).

— cf session agent (Claude Fable 5), on behalf of danfuzz

ubik2 (commented):
Re-review of `cef2d9c74`: **request changes**, based only on the two regressions below.

### 🔴 Blocking

1. **Nested inline-to-array promotion creates a link cycle** (`packages/runner/src/data-updating.ts`, promotion branch around line 602).

   Promotion shallow-copies the shared object, but descendants remain registered in `state.seen` under the earlier inline path. For a value shaped like `{ first: shared, list: [shared] }` where `shared` contains a nested object, the promoted document links back through `/first/...`, and `/first` is then repointed to that document. A focused repro throws `Link cycle detected` on the initial read. Promotion needs to rebase or rebuild descendant `seen` entries under the promoted document root. Please add a nested-object regression test; the existing flat-object tests do not exercise this.

2. **Framed `addUnique()` rejects a new self-referential candidate** (`packages/runner/src/cell.ts:1675`; also reported by Cubic in https://github.com/commontoolsinc/labs/pull/5221#discussion_r3687016001).

   `fabricFromNativeValue()` throws `Cannot store circular reference` during duplicate comparison, before `diffAndUpdate()` can anchor the object and encode the cycle as a self-link. The focused repro succeeds at the merge base and fails on this head. A candidate that cannot be comparison-normalized due to a cycle should be treated as not already present and handed to the existing cycle-aware write path. Please cover this through framed `addUnique()`.

### 🟡 Non-blocking observations

- A framed `push(cell.getRaw()[0])` can promote and rewrite the unchanged prefix after its read dependency has been excluded, allowing a stale append to overwrite a concurrent edit. I reproduced the durable result changing from the expected `[{ v: 9 }, { v: 1 }]` to `[{ v: 1 }, { v: 1 }]`. This also occurs on main, so I am not treating it as a blocker for this PR; pending consistency work may address it. Cubic thread: https://github.com/commontoolsinc/labs/pull/5221#discussion_r3687015998.

- An accepted volatile `toJSON()` candidate can be normalized once for duplicate comparison and again for writing, so the compared and stored values can differ. This class of value is problematic more generally, and I do not consider it merge-blocking here.

All current CI checks are green. The blocking verdict is specifically for the nested-promotion cycle and the newly introduced self-referential `addUnique()` rejection.

— Codex (GPT-5), on behalf of the reviewer

packages/runner/src/data-updating.ts
```diff
@@ -448,13 +564,73 @@ export function normalizeAndDiff(
...
+      }
+      diffLogger.debug(
+        "diff",
+        () =>
+          `[SEEN_PROMOTE] Promoting inline-aliased array element at path=${pathStr}`,
+      );
+      // Anchoring re-registers the object in `state.seen` under the new
+      // document's root, so later occurrences (and the content recursion's
+      // own cycle hits) link there.
+      const anchorChanges = anchorValueAsEntity(
```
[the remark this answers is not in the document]
danfuzz: Verified and fixed — this was real, and constructible whenever the stored tree is deep-frozen (raw reads then preserve reference identity, so a handler can hold the exact stored interior object). The root cause was the identity guard skipping the *anchor* but still *descending* the untouched element, which registered its interiors in the walk's seen map; a pushed reference to one of them then promoted — repointing content inside the untouched prefix under the suppressed mergeable read.

An untouched element now returns no changes **without descending**: it diffs to nothing by identity, and its interior never enters the seen map, so the pushed reference anchors as a fresh document instead of aliasing into stored content. The regression test seeds a deep-frozen inline tree, pushes the exact stored interior reference, and asserts the prefix stays byte-for-byte untouched (red before the fix with the interior repointed, green after).

— cf session agent (Claude Fable 5), on behalf of danfuzz

packages/runner/src/cell.ts
```diff
@@ -1638,41 +1637,59 @@ export class CellImpl<T extends FabricValue>
...
-      );
-    const toAdd = anchored.filter((candidate) => !alreadyPresent(candidate));
+        );
+      }
+      // Link-carrying candidates (query-result proxies, raw sigil links)
+      // compare as themselves -- the write boundary passes them through
+      // unconverted, and the strict conversion would reject their
+      // non-string-keyed internals.
+      const comparable = normalizeForComparison && !isCellLink(candidate)
+        ? fabricFromNativeValue(candidate) as FabricValue
```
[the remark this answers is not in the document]
danfuzz: Verified and fixed, with a sharper resolution than treating normalization failure as non-duplicate: a cyclic candidate can never equal a stored element at all — stored fabric values are acyclic (cycles persist as links) — so it is new *by definition*. A `containsCycle()` ancestors-walk (plain containers only; cells, links, and atomic fabric objects are leaves) short-circuits cyclic candidates past both normalization and comparison, with no error caught or swallowed. The regression test asserts a framed `addUnique` of a self-referential object lands anchored with its cycle as a self-link (red before with the conversion throw, green after).

— cf session agent (Claude Fable 5), on behalf of danfuzz

danfuzz:
Responding to the re-review of `cef2d9c74` (this and the follow-up commits address it; apologies for the crossed wires — your review landed minutes before a commit that already covered part of it):

**Blocking 1 (nested promotion cycle): verified and fixed.** Your shape reproduced exactly (`Link cycle detected`, document interior → `/first/meta` while `/first` → document). The promotion branch now purges every `seen` entry registered under the inline location before the content recursion, which re-derives the descendants under the promoted document's root. Your nested-object case is the regression test ("promotes a shared object with nested content without creating a cycle"), red before the purge and green after.

**Blocking 2 (cyclic `addUnique` rejection): fixed in `f77ae01df`** (pushed shortly after your review was submitted, prompted by cubic's parallel report). Resolution is slightly sharper than treat-normalization-failure-as-new: a cyclic candidate can never equal a stored element (stored fabric values are acyclic; cycles persist as links), so `containsCycle()` short-circuits it as new by definition — no error caught or swallowed. Covered through framed `addUnique()` as you asked.

**Non-blocking 1 (framed `push(getRaw()[0])` prefix rewrite):** the this-branch variant is closed by `f77ae01df`'s untouched-element no-op — an element carried through by identity emits nothing and is not descended, so the promotion path can no longer reach stored prefix content (pinned by two regression tests, flat and nested-array). The `main` variant you reproduced remains, as you say, for the pending consistency work.

**Non-blocking 2 (volatile `toJSON()` double-normalization):** acknowledged and agreed it's a broader class; the dedup comparison normalizes once and the write normalizes independently, so a volatile `toJSON()` can diverge between them. Left as-is per your assessment.

— cf session agent (Claude Fable 5), on behalf of danfuzz

packages/runner/src/data-updating.ts
```diff
@@ -965,6 +1115,69 @@ export function normalizeAndDiff(
...
+  // converted later in the recursion, so a deep comparison would inspect
+  // values whose canonical form does not exist yet.
+  if (
+    state.nextAnchorId !== undefined &&
+    isArrayElement &&
+    isObject(newValue) &&
+    !(newValue instanceof FabricSpecialObject) &&
+    !isCellLink(newValue)
+  ) {
+    if (Object.is(currentValue, preConversionValue)) {
```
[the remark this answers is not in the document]
danfuzz: Not reproducible as described, and now pinned by a test so it stays that way. The identity no-op is checked per element at every array depth: an untouched nested-array element does descend (arrays are never anchored), but each of its object elements hits the no-op — with the current value threaded down by the array branch — before being registered, so `state.seen` gains only the arrays themselves, and arrays are excluded from promotion. A pushed reference to an object inside an untouched nested-array prefix therefore misses `seen` and anchors as a fresh document; the prefix stays byte-for-byte untouched. The new "untouched NESTED-array prefix" regression test constructs exactly this (deep-frozen seed, identity-preserving raw read, bare interior reference pushed) and passes.

The adjacent real cycle bug in this area — promotion of a shared object with *nested content* linking back through the repointed inline path — was raised in the human re-review and is fixed in this round (descendant `seen` entries purged on promotion).

— cf session agent (Claude Fable 5), on behalf of danfuzz

PR #5223, opened 2026-07-30
fix(patterns): favorites rows require a `tag` nothing has ever written

bfollington:
## Why

Home's Favorites and Profile tabs render nothing on an existing space, with a
completely clean browser console.

`favorites-manager.tsx` declared `tag: string` — required, no default — on the
element type that `favorites.map(...)` instantiates a sub-pattern from. No row
in storage satisfies it. Discovery tags moved to a client-derived plural `tags`
array in #4197, whose own description records the split: favorites created
before it carry `tag`, everything since carries `tags`. The manager reads
NEITHER — `cf-tags` renders `item.userTags`.

So every post-#4197 favorite fails the row's argument validation the moment it
is re-staged:

```
Error: updated arguments do not match the candidate schema:
       element: missing required property tag
  at Runner.validateArgument → updateAndValidateArgument → applySetupState
   → setupInternal → Runner.run → reconcile → raw:map:…
```

The row never instantiates, and the scheduler re-throws it on every pass. That
loop starves the runtime: `cell:resolveAsCell` and `cell:subscribe` for the
row's cells stop being answered and sit pending until they time out, so the
links fall through to their placeholder text and the tab reads as empty. None
of it is visible from the page — the error is raised inside the worker runtime,
so the browser console stays empty unless you call
`commonfabric.forwardWorkerConsole()` first. That silence is why one wrong
property definition read as "the tabs are broken" for weeks.

## What Changed

`tag` is **deleted** from `favorites-manager.tsx`'s `Favorite` type, not made
optional. That type is the manager's *projection* of a stored row, and every
field it names is re-validated against every vintage on each pattern swap:
requiring `tag` breaks post-#4197 rows, requiring `tags` would break pre-#4197
rows identically, and `tag?:` is merely harmless dead weight. Naming neither is
what accepts both. The reasoning is left in a comment at the type.

Plus a regression test, `packages/runner/test/favorites-row-stored-shape.test.ts`.
Three things about its shape are load-bearing:

- It drives the **real shipped** `favorites-manager.tsx`, read from disk. The
  defect was a disagreement between that file's projection and what is in
  storage; a mirrored copy of the type in the test would have drifted along with
  the bug instead of catching it.
- It seeds **both stored vintages** — a post-#4197 row (plural `tags`, keyed
  `id`) and a pre-#4197 row (singular `tag`, no `tags`, no `id`). Without the
  legacy row, "fixing" this to a required `tags: string[]` would pass while
  re-breaking every old favorite.
- It asserts through a pattern **swap**. Adding a favorite to an already-running
  piece never re-stages the argument, so a fresh-data test stays green against
  the bug — exactly how this survived every existing test. The swap is what the
  auto-update does to a running piece, and it is the moment stored rows meet the
  schema.

The assertion is on `scheduler.onError` rather than rendered output, because a
failed row leaves the markup the map produced before the swap. A name marker on
v2 is asserted too, so a pointer move that silently did nothing cannot pass on
"no errors" alone.

## Test plan

Mutation-verified in both directions against the real pattern:

| mutation | result |
|---|---|
| `tag: string` required (the shipped bug) | fails — `missing required property tag` |
| `tags: string[]` required (the plausible "fix") | fails — legacy row, `favorites: value does not match anyOf` |
| as merged | passes |

`deno fmt --check`, `deno lint`, and `deno check` clean on both files.

Reproduced end to end against `origin/main` (1513fb985) with a local two-build
harness: seed a home space (one favorite, one profile) on a pre-retirement
build, re-open the same store on main. Before: the error fires on every load and
both entries render as `Unknown Link`. After: both render their real names
(`Home #LeUGRw`, `Test Persona #1mjzzE`) with `cf-cell-link` resolving normally
— confirmed in a browser by the author.

The existing piece adopts the fix on its own through the pattern auto-update.
Instrumenting `PatternUpdater.#check` confirms the root and all four sub-pieces
fetch their advertised identity and converge; no piece surgery, no space
redeploy.

## Follow-up, not in this PR

The #5209 vintage gate cannot currently see this class: review found it reports
`2 updated cleanly with no state stranded` even with `tag` required, because the
pinned vintage holds no populated favorite row. A captured vintage carrying real
rows is what would make that gate catch the next one.

### DIFF
--- a/packages/patterns/system/favorites-manager.tsx
+++ b/packages/patterns/system/favorites-manager.tsx
@@ -13,7 +13,7 @@ import {
 
 type Favorite = {
   cell: Writable<{ [NAME]?: string }>;
-  tag: string;
+  tag?: string;
   userTags: Writable<string[]>;
   spaceName?: string;
   // Stable key the favorite entity is addressed by (the piece's identity),
--- a/packages/patterns/system/favorites-manager.tsx
+++ b/packages/patterns/system/favorites-manager.tsx
@@ -11,9 +11,17 @@ import {
   Writable,
 } from "commonfabric";
 
+// This manager's PROJECTION of a stored favorite: only the fields it renders.
+// Naming a field it does not read costs compatibility for nothing — the row is
+// re-staged and validated against this type on every pattern swap, so a field
+// declared here must hold for every vintage in durable storage. Discovery tags
+// are the cautionary case: rows written before #4197 carry `tag` (singular),
+// rows written after carry `tags` (plural), and this manager reads neither
+// (`cf-tags` renders `item.userTags`). Declaring `tag: string` required made
+// every post-#4197 row fail validation; requiring `tags` would fail every
+// pre-#4197 row the same way. Omitting both is what accepts all of them.
 type Favorite = {
   cell: Writable<{ [NAME]?: string }>;
-  tag?: string;
   userTags: Writable<string[]>;
   spaceName?: string;
   // Stable key the favorite entity is addressed by (the piece's identity),
--- a/packages/runner/test/favorites-row-stored-shape.test.ts
+++ b/packages/runner/test/favorites-row-stored-shape.test.ts
@@ -97,10 +97,10 @@ describe("a stored favorite row instantiates in favorites-manager", () => {
     rt = new Runtime({
       apiUrl: new URL(import.meta.url),
       storageManager,
+      // `#favorites` resolves against the identity's HOME space. `Runtime`
+      // derives that DID from the storage manager's signer, which is this
+      // space's signer, so there is nothing to pass here.
       experimental: {},
-      // `#favorites` resolves against the identity's HOME space; this runtime's
-      // space is that home space.
-      userIdentityDID: space,
     });
   });
   afterEach(async () => {
@@ -161,6 +161,7 @@ describe("a stored favorite row instantiates in favorites-manager", () => {
     // appear while already running.
     const seedTx = rt.edit();
     hostCell.withTx(seedTx).key("favorites").set([
+      // Post-#4197: discovery tags are plural, and the row is keyed by id.
       {
         cell: thingCell,
         tags: [],
@@ -169,6 +170,15 @@ describe("a stored favorite row instantiates in favorites-manager", () => {
         // property carrying `undefined` is a value of the wrong type, not an
         // omission, and the row is rejected for that instead.
         id: "favorite-1",
+      },
+      // Pre-#4197: a single `tag`, no `tags`, and no keyed `id`. Both vintages
+      // are in real storage, so a projection that requires EITHER tag field
+      // breaks one of them — this row is what keeps a future "fix" to a
+      // required `tags: string[]` from passing.
+      {
+        cell: thingCell,
+        tag: "#thing",
+        userTags: [],
       },
     ] as never);
     await seedTx.commit();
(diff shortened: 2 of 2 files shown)


bfollington (commented):

Scope: both changed files deep-read; traced Home/runtime-client/wish/profile producers, relevant history, the #5209 vintage gate, and existing real-pattern/swap tests. Checks: targeted test green on the PR; mutation back to required `tag` red with the exact `applySetupState` error; `deno fmt --check`; `deno lint`; focused type-check; CI logs; vintage-gate mutation.

## 🔴 Blocking

- `packages/runner/test/favorites-row-stored-shape.test.ts:103` does not type-check. `userIdentityDID` is not a `RuntimeOptions` property (`TS2353`), which is why the PR's `Check` job is red. `Runtime` already derives `userIdentityDID` from `options.storageManager.as.did()`; here `StorageManager.emulate({ as: signer })` therefore already makes it equal `space`. Delete the `userIdentityDID: space` line. Verified: after that deletion, focused `deno check` and the test both pass.

## 🟡 Improvements

- `packages/patterns/system/favorites-manager.tsx:14-22`: prefer deleting `tag` from this consumer projection, not making it optional, and correct the PR/test history. Singular `tag` is not “never written”: Home wrote `{ cell, tag, userTags }` from January until the June 18 `d5d115cda` tags refactor; current Home writes `{ cell, tags, userTags, ... }`. The manager reads neither `tag` nor `tags`, so the clean compatibility contract is to name only the fields it consumes. Requiring plural `tags` would break real pre-June rows; keeping `tag?:` is safe but leaves dead/misleading schema weight. If this type is intentionally meant to describe the full persisted union rather than the manager's projection, spell both vintages as optional instead.

- `packages/runner/test/favorites-row-stored-shape.test.ts:47-57`: exercise both real row vintages. The current fixture covers only the post-June plural shape, so a future “fix” to required `tags: string[]` would keep this test green while re-breaking older stored favorites. Add a legacy row/case with `tag`, no `tags`, and no keyed `id`, alongside the current plural row.

- The test is in the right lane and observes the right failure seam. A runner test reading a shipped system pattern has ample precedent, and `scheduler.onError` is necessary because the failed re-stage leaves stale rendered markup behind. The #5209 gate is complementary, not a replacement: I ran it with required `tag` plus a comment-only identity change and it reported `2 updated cleanly with no state stranded` (exit 0), because the pinned vintage has no populated favorite row.

- `packages/runner/test/favorites-row-stored-shape.test.ts:199-216`: the pointer mutation is a legitimate stand-in for the runner half of production auto-update and matches existing swap tests, but the explanation is inaccurate. Compiling identical source cannot yield a distinct content-addressed identity; this test gets a new identity because it appends `// swap`, not because the transaction/cause is fresh. Make v2 observably different (for example, change the manager name/marker) and assert that marker after the pointer move. Keeping the scheduler-error assertion is still correct; the positive marker just proves the swap actually ran instead of allowing a vacuous no-error pass.

## Scoped duplicate check

`profile-home.tsx:47` is not the same class: every `appendElement` producer supplies `tag`, and rendering reads it as the title fallback (`element.title ?? element.tag`). A scoped search of system-pattern singular `tag` fields found only favorites-manager and profile-home; no second required-but-never-written case was confirmed.

Verdict: the one-line runtime fix is sound but should ideally delete the unused field; merge is blocked by the test's type error. Keep the behavioral test in runner, broaden it to both stored vintages, and make the swap positively observable.

bfollington:
Thanks — all four points were right, including that my "never written" premise was false. Fixed in 8abb1117b.

**Blocking / TS2353.** `userIdentityDID` dropped. As you say, `Runtime` derives it from the storage manager's signer, which is already this space's signer, so the line was redundant as well as invalid. `deno check` on the test file is clean now.

**`tag` deleted rather than made optional.** You're right about the history — #4197's own description says favorites created before it carry `tag`, and Home wrote `{ cell, tag, userTags }` until then. So both vintages are in durable storage and the manager reads neither. I've written the type as what it actually is, the manager's *projection* of a stored row, with the reasoning in a comment: every field named here is re-validated against every vintage on each swap, so requiring `tag` breaks post-#4197 rows, requiring `tags` breaks pre-#4197 rows, and naming neither is what accepts both. The PR body's history claim is corrected too.

**Both vintages in the fixture.** Added the pre-#4197 row (singular `tag`, no `tags`, no keyed `id`) alongside the plural one. Mutation-verified both directions: required `tag` → `missing required property tag`; required `tags: string[]` → the legacy row fails with `favorites: value does not match anyOf`. That second one passed silently before the legacy row existed, which was exactly your point.

**Swap made observable.** Also right that my explanation was wrong — content-addressed identity can't change from a fresh cause; it was the appended `// swap` comment doing the work. v2 now changes the pattern's name and the test asserts it after the pointer move, so a pointer move that silently did nothing can no longer pass on "no errors" alone.

**Duplicate check.** Agreed on `profile-home.tsx:47` — every `appendElement` producer supplies `tag` and rendering reads it as the title fallback, so it isn't the same class. Dropped from the PR's follow-up section.

On the #5209 gate: your finding that it reports `2 updated cleanly with no state stranded` with required `tag` is worth its own follow-up — the pinned vintage has no populated favorite row, so the gate can't see this class at all. A vintage carrying real rows would be the durable fix.

PR #5224, opened 2026-07-30
feat(ui): retire elements as inert stubs instead of deleting them

bfollington:
## Why

A pattern's source is **durable**, and a piece keeps running the source it was
stored with. So deleting a component does not retire it — it removes the
definition out from under source that still names it. Removing
`cf-cell-context` (#5132) is the worked example: stored home sections still
emit it, and the debugging that followed cost most of a day (#5223 was the
other half).

This gives a retired element a definition that renders its children and does
nothing else, plus a warning when durable source still reaches it.

## What Changed

`packages/ui` only — six files, nothing outside it.

- `src/v2/core/retired-element.ts` — the base: renders a default slot, warns
  once per element per session.
- `src/v2/components/cf-cell-context/` — the first entry, pointing at
  cf-piece-menu.
- A README section for retired elements, with the rule written down so the next
  retirement follows it rather than rediscovering it.

## Scope cut after verification

The first version of this PR also restored `cf-cell-context` to
`packages/html/src/jsx.d.ts`, for an edit-time `@deprecated` signal. **That
half is removed**, because it breaks the exact case this PR exists for.

Bisected against a home space seeded on a pre-retirement build and re-opened on
this branch:

| variant | seeded space |
|---|---|
| main (control) | renders (4 tab panels) |
| registration + `jsx.d.ts` declaration | **blank** |
| registration only (what this PR now is) | renders (4 panels, stub in DOM with shadow root) |
| `jsx.d.ts` only, no registration | **blank** |
| `jsx.d.ts` with `"$cell"?: unknown` | **blank** |

Declaring the element in `JSX.IntrinsicElements` stops such a space rendering —
silently: no console error, every element defined, the root view simply never
fills in. Not the `CellLike` typing either; relaxing it to `unknown` changes
nothing. Ruled out on the way: the request that appears stalled in the network
panel is served by the toolshed in 1.5ms; compiling the largest system pattern
is 1014ms on main vs 916ms with the declaration; and there is no rebuild loop.

That is worth its own investigation and is **not** a retirement problem — it
should be reachable by adding any intrinsic element. Filing separately.

The runner test went with it, because its premise was wrong. It asserted that
stored source naming the element still COMPILES. In fact the stored-source
reload path rejects an unknown element too — #5133's leniency covers a stale
`@ts-expect-error`, not an unknown JSX element. Stored pieces do not recompile
at all; they load pre-compiled artifacts, which is why a seeded space renders on
main despite naming a deleted element. The test only passed because of the
declaration this PR no longer makes, so it was pinning a contract that does not
exist.

## Test plan

- `packages/ui/test/retired-element.test.ts` — the warn-once ledger. Once per
  element per session is a contract: a retired element inside a `.map(...)` over
  durable rows would otherwise warn per row per render, which is a wall that
  gets muted, which is the same as no warning.
- `packages/ui` Deno test phase: **161 passed, 0 failed**, including
  `package-root-registration`, which required the README entry.
- `deno fmt --check` and `deno lint` clean.
- Verified in a browser against the seeded pre-retirement space: renders 4 tab
  panels with the stub registered, present in the DOM, and slotting its
  children — matching the control run at identical timing.

## Known gaps

- **Coverage**: the stub's class body (11 lines) is not executed by any test,
  because exercising it needs a DOM. The honest fix is a `*.browser.test.ts`,
  but that lane currently fails to bundle even on a clean checkout (identical
  failure with none of these changes), so adding to it would be building on
  sand. Flagging rather than papering over it.
- **No edit-time signal yet.** That was the `jsx.d.ts` half. A lint over pattern
  sources gives the same warning without touching the type graph, and is the
  follow-up.

## Downstream note

Loom's `pattern-host.js` imports the whole cf-ui palette except an explicit
excluded list, enforced by `tests/webapp/test_cf_ui_palette_drift.py`.
Re-registering `cf-cell-context` forces a decision there at the next vendor
bump — excluding it with a reason is probably right, since nothing should newly
author against it.


bfollington (commented):

Scope: PR 5224, all 8 changed files deep-read; prior component and JSX declarations inspected from main history; compiler and test mutations run. Checks: targeted UI tests 4/4 pass, stored-source runner test 1/1 passes, touched-file fmt/lint pass, full CI Check is green, Coverage Check is currently red (+10 uncovered UI lines, +1 runner line).

## 🔴 Blocking

1. **`packages/ui/src/v2/core/retired-element.ts:64` changes the old component's layout rather than preserving it.** The retired host uses `display: contents`, which removes the host's box. The component immediately before #5132 used `display: block; position: relative; flex: 1; min-height: 0`, with `:host([inline]) { display: inline-block; flex: none; }`; it also reflected the `inline` property. In a flex parent, the old component was one flex item that filled the available axis. The stub promotes its slotted children into the surrounding box layout instead. That can change sizing, number of flex items, min-height behavior, and inline behavior for durable source. **Fix:** make retirement layout component-specific and retain `cf-cell-context`'s old host box rules (including reflected `inline` behavior), while removing only the toolbar/interaction. Add a browser test for block/flex and inline cases. **Verified** from the pre-#5132 implementation and CSS box semantics.

2. **`packages/runner/test/retired-element-in-stored-source.test.ts:99-105` does not test that the stub passes children through.** It asserts on the pattern's stored VNode before the browser custom element is instantiated; that tree contains the authored child regardless of what `RetiredElement.render()` does. I mutation-tested `render()` from `<slot>` to an empty Lit template and this test still passed 1/1. Thus the stated regression invariant is currently unguarded, and the test's comment overclaims. **Fix:** keep the runner case as the compile/VNode-compatibility test, and add a UI browser test that imports/registers the real element, mounts a light-DOM child, and verifies it is assigned/rendered through the slot. That test should also pin the legacy host layout and cover the currently-uncovered component lines. **Verified** by mutation.

## 🟡 Improvements

- **The `$cell`/schema rationale in `cf-cell-context.ts:15-18`, `retired-element.ts:55-58`, `jsx.d.ts:3595-3598`, and the runner test comment is false for this case.** The pattern compiler loads the platform `jsx.d.ts`; it never sees the Lit class declaration. Removing the Lit `cell` accessor left the stored-source test green. In a favorites-shaped `.map(...)` fixture, changing the JSX attribute type from `CellLike<any>` to `any` produced byte-for-byte identical transformed output and identical row schemas. The `$` prefix drives binding at runtime, while the row's authored `Favorite.cell: Writable<...>` type drives its schema. Retain the JSX element/property declaration because it is required for TypeScript to accept durable source, and retain `CellLike<any>` for authoring type safety; retain the Lit property only if preserving the old DOM API shape is desirable. Remove the claims that the class property or contextual prop type feeds the surrounding schema, and do not cite #5223's independent required-`tag` row-schema bug as evidence for it. **Verified** by transformed-output and test mutations.

- **Add an enforcement gate for new uses; do not use it instead of the compatibility stub.** `@deprecated` is editor metadata, not a failing compiler/CI diagnostic, so it cannot enforce “stop emitting this from new patterns.” A canonical retired-tag registry should drive a repository check that rejects retired tags in active pattern source, with an explicit exception for stored-source compatibility fixtures. Better still, have the authoring compiler diagnose retired JSX while the `storedSource` path permits it. Keep `@deprecated` and the runtime warning as additional signals.

- **The runtime warning granularity is reasonable as tag + loaded UI realm, not tag + piece.** A module-level `Set` is not literally a product session, but in the normal browser bundle it approximates one and avoids `.map()`/reconnect spam. Per-piece warnings are only more useful if the warning can also name that piece/source; repeating the same generic line per piece adds noise without actionability. I would keep the current key and describe the contract precisely as “once per tag per loaded UI realm.”

## Possible follow-up: historical palette audit

Since the `cf-*` rename, main history shows `cf-cell-context` is the only removed `cf-*` intrinsic, so this PR covers the only post-rename deletion. The earlier history is materially larger:

- `ct-list` and `ct-list-item` declarations/components were deleted in #2683.
- `ct-charm` was deleted/renamed in #2723.
- `ct-outliner`'s implementation was deleted while its JSX declaration lingered until the namespace rename.
- #3177 then removed all 107 remaining `ct-*` intrinsic names/registrations in favor of `cf-*`; the current tree has no `ct-*` registrations, and that PR says external runtime coordination was deferred but does not document a durable-source migration.

If pre-April durable source can still be loaded, this new “union of every shipped palette” principle makes the `ct-*` namespace a compatibility seam that needs an explicit migration/retention decision. I would audit actual retained source before adding 100+ aliases, but the four independently deleted elements above are the first concrete candidates.

**Verdict:** the retirement policy is sound, but do not merge until the stub preserves the old layout and a real DOM test proves the passthrough invariant. Correct the schema explanation and add the new-use gate either here or as an explicitly tracked immediate follow-up.

bfollington:
**Do not merge — this fails the case it was written for.** Verified locally against the harness that reproduces the original bug (a home space seeded on a pre-retirement build, re-opened on the new build):

| store | main (no stub) | this branch |
|---|---|---|
| pre-retirement seeded | 4 tab panels, renders | **0 panels, page blank** |
| fresh / empty | renders | renders |

Reproducible across repeated runs and fresh copies of the store. So the stub does not break the app in general — it breaks specifically when stored source names the retired element, which is the entire point of the change.

Signature: no console errors, no worker errors, all custom elements defined (`customElements.get('cf-cell-context')` is true), and `x-root-view` simply never fills in. A silent stall, the same shape as the starvation in #5223 rather than a thrown error.

Working hypothesis, not yet confirmed: while `cf-cell-context` was *unregistered*, the reconciler treated it as an inert unknown element and just rendered its children. Registering it as a real element with a `cell` property means the `$cell` binding in stored source is now marshalled into a live element that never reads it — and something in that path does not settle. If that is right, the fix is for the stub to accept the prop without taking a cell subscription, and my reasoning for declaring `cell` at all (preserving the derived schema) needs re-checking — which is one of the things this PR's review was asked to verify.

Keeping the branch up because the tests and the policy are still the right shape; the implementation is not.

bfollington:
Bisected my own patch against the seeded pre-retirement store. The result is the opposite of the PR's premise.

| variant | seeded store |
|---|---|
| main (control) | renders (4 panels, pass 2) |
| full patch — registration + `jsx.d.ts` | **blank** |
| registration only, no `jsx.d.ts` | **renders** — stub registered, in the DOM, shadow root, children slotted |
| `jsx.d.ts` only, no registration | **blank** |
| `jsx.d.ts` with `"$cell"?: unknown` instead of `CellLike<any>` | **blank** |

So the runtime stub is fine and does its job. What breaks a space whose stored source names the element is **declaring it in `JSX.IntrinsicElements` at all** — not the registration, and not the `CellLike` typing (relaxing it to `unknown` changes nothing).

Ruled out along the way: the toolshed serves `profile-home.tsx` in 1.5ms (200, 42KB), so nothing hangs server-side; compiling `profile-home.tsx` takes 1014ms on main vs 916ms with the declaration, so it is not a type-check blowup; and there is no rebuild/reload loop (one build in the dev-server log — the repeated page loads in the network panel were my own navigations).

Working theory, unverified: without a declaration the stored source raises a type error that #5133's stored-source leniency swallows, and whatever the compiler emits on that path happens to render. With a declaration the source type-checks and the element and its props are emitted through the normal path, which is what stalls. If that is right, the interesting bug is in that normal path, not in the retirement policy — and the stall would be reachable by other means.

Consequence for this PR: the registration half is verified good and worth keeping; the `jsx.d.ts` half cannot ship as written. That also removes the edit-time half of what this was for, so the shape needs rethinking rather than patching — a CI lint over pattern sources may be the better way to get an authoring signal without touching the type graph. Staying draft.

bfollington:
Addressed in 030d0df72. CI is green on the reduced scope.

**Blocking — `display: contents` (valid, fixed).** You're right that an inert element is not a layout-neutral one. The base class no longer imposes any layout, and says why in a comment; the `cf-cell-context` stub now carries that component's `:host` rules verbatim — `display: block`, `position: relative`, `flex: 1`, `min-height: 0`, plus the `inline` variant. A test pins them and asserts `display: contents` does not come back.

**Blocking — runner test passes when the stub renders nothing (moot).** That test is gone. Its premise was wrong independently of your finding: it asserted stored source naming the element still COMPILES, but the stored-source reload path rejects an unknown JSX element too (#5133's leniency covers a stale `@ts-expect-error`, not this). Stored pieces never recompile — they load pre-compiled artifacts — which is why a seeded space renders on main despite naming a deleted element. Your point stands for what a real proof would need, and a browser DOM test is the honest way to get it; that lane currently fails to bundle on a clean checkout, so it is flagged as a gap rather than faked.

**Refuted — the `cell` property / JSX contextual type do not affect the row schema.** Confirmed independently and the claim is removed from the source comments. It was my reasoning for restoring the `jsx.d.ts` declaration, and that declaration turned out to blank a seeded space outright; removing it is what fixed rendering.

**Warn granularity.** Kept at once per tag per loaded realm, as you suggest — per-piece would be better but nothing at that layer can name the piece today.

**CI enforcement for new usage.** Agreed, and it is now the plan for the edit-time half, since the type-declaration route is off the table.

**`ct-list`, `ct-list-item`, `ct-charm`, `ct-outliner`, and the ~107 names dropped in the `ct-*` → `cf-*` migration** — this is the most valuable thing in the review. If durable source still names any of them, they are the same latent class as `cf-cell-context`, just not yet triggered. Filing separately rather than expanding this PR: the right first step is finding out which of those names actually appear in stored source, which is a question about the live store, not about the palette.

PR #5225, opened 2026-07-30
feat(memory,runner): patch-op accepts carry the post-apply document; confirmPending promotes to server truth (CT-1926)

ubik2:
## Why

The client's `confirmPending` promotes an accepted commit by applying its local pending patch onto the local confirmed mirror — which can be stale for that doc, because the inline accept systematically outruns the batched novelty fan-out carrying the foreign writes the patch was applied on top of (verified structural: 5ms refresh timer + drain-wait that yields to inbound traffic, widest under load). The promotion then mints a confirmed `(seq, value)` pair the server never had, and that stamp feeds later reads' staleness bases (`pushCommitRead` emits `record.confirmed.seq` as confirmed `seq` / pending `basisSeq`) — an honest client producing INV-1 missed-write observations that no current checker can see, since engine, naive model, and oracle all replay the claimed basis. This is the CT-1872 "unpinned convergence" wrinkle plus its sharper dishonest-basis consequence. CT-1926 has the full design discussion; the ordering-side complement is CT-1927.

## What

**Engine** (f3eca4b93): every `patch` revision in an accept response now carries `document` — the doc's POST-COMMIT state, recovered via the canonical seq-pinned read right after the head update (`attachPostApplyDocuments`; the same state on every patch revision of a doc, revision-order independent). `selectCommitRevisions` recovers it for replays at THE REPLAYED COMMIT's seq, not head (safe under snapshot retention: reconstruction falls back to the last durable `set` row + patch fold). `set` keeps its payload; `delete` stays identity-only. Zero server.ts changes (the server returns `AppliedCommit` verbatim) and no other consumer of `AppliedRevision.document` exists (fan-out runs off `markSpaceDirty`).

**Client** (b1dc40af0): `confirmPending` reads the accept's revisions (last revision per doc wins; a patch revision without `document` — pre-CT-1926 server — also invalidates earlier same-doc values, so partial views never promote). The rule, refined during implementation by an existing pin: compute the local promotion as before, compare with the canonical `valueEqual` —

- **equal (steady state)** → keep the local promotion and its materialization cache, so the visible value's IDENTITY survives confirmation (pinned by "confirming the head pending write promotes the cached materialization"; identity churn on every accept would ripple into shallow change detection);
- **different (a foreign write raced the fan-out)** → promote the server truth and drop the cache — the identity change is real.

Identity changes iff the value does. This is not the rejected hash-optimistic design: no wire hashes, no held commits, no new client state — both values are in hand and the comparison is local.

**Tests**: three engine tests (post-apply includes unseen foreign writes; shape preservation across set/delete/multi-patch; seq-pinned replay vs head); scripted server model attaches documents like the real engine; convergence regression (foreign whole-doc replacement interleaves before a blind patch's accept, no sync delivered → confirmed equals server truth INCLUDING the foreign fields — the CT-1872 convergence property, closed for the accept path); old-server fallback regression (`omitPostApplyDocuments`: extrapolation preserved, the documented residue).

**Docs**: 04-protocol §4.3.1 (response revision payloads), 03-commit-model §3.12 (accept-side promotion rule).

## Rollout

Additive field, no negotiation (same pattern as `basisSeq`): old clients ignore it; new clients fall back to extrapolation against old servers. Under CT-1927 (flush-before-verdict + frame-time retirement) the value becomes a guarded no-op for watched docs; it stays load-bearing for unwatched docs, pre-CT-1927 servers, and rollout skew — attached unconditionally on purpose.

## Verification

- `packages/memory`: 459/459 (three new engine tests included)
- `packages/runner`: 1101/1101 — including the new convergence + fallback regressions AND the pre-existing identity-stability pin that forced the valueEqual refinement
- `deno task check` green (runner package + root), `check-docs` 523 blocks, fmt/lint clean
- Self-review (cf-review) verified the three least-trusted spots: `valueEqual` undefined handling (`Object.is` fast path), replay reconstruction under snapshot retention (falls back to durable `set` + fold — cost, not error), reconstruction cost bounded by `DEFAULT_SNAPSHOT_INTERVAL = 10` patches

## Notes for review

- Originally stacked on #5110 (CT-1910), which merged 2026-07-29 with history preserved — this PR is exactly the three CT-1926 commits.
- Known added cost: one bounded document reconstruction per patch-written doc per accept (the server previously never materialized per commit; snapshots are interval-gated). If it ever profiles, the natural reuse is feeding `materializeSnapshots` from the same reconstruction.
- CFC read-authority checkpoint from the ticket (write-only shapes should not receive other writers' content in the response) remains a server-layer review item — the engine attaches unconditionally today, matching the existing behavior of `set` payload echo and query access.

---
## Summary by cubic
Accept responses now include the post-commit document for patch ops, and the client promotes confirmed state from that server value when it differs from local extrapolation. This fulfills CT-1926 and prevents confirming states the server never had.

- **New Features**
  - `packages/memory`: Each patch revision in an accept carries `document` (the post-commit state), recovered via a seq-pinned read; `set` keeps its payload and `delete` stays identity-only. The field is additive, so old clients ignore it and replays return the state as of the replayed commit.
  - `packages/runner`: `confirmPending` compares the local promotion with the response’s `document`. If equal, it keeps the local identity; if different, it promotes the server value and drops the cache. Falls back to local extrapolation when a server omits `document` (pre-CT-1926).

<sup>Written for commit 9c8018c8500039b8a2fc22191c329e984ae417c3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5225?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/memory-v2/03-commit-model.md
+++ b/docs/specs/memory-v2/03-commit-model.md
@@ -566,6 +566,17 @@ When a commit is rejected:
 
 The client library SHOULD support automatic retry with a bounded retry count.
 
+When a commit is ACCEPTED, the response's patch revisions carry the
+document's post-commit state (04-protocol.md §4.3.1, CT-1926). The client
+promotes its confirmed mirror from that value when it differs from its own
+locally-extrapolated result — the accept can outrun the batched fan-out
+carrying the foreign writes the patch was applied on top of, and promoting
+a local extrapolation in that window mints a confirmed (seq, value) pair
+the server never had, whose stamp then feeds dishonest staleness bases into
+later reads. When the response value matches the local extrapolation (the
+steady state), the client keeps its local promotion so the materialized
+value's identity survives confirmation.
+
 ## 3.13 Mapping from Current Implementation
 
 | v1 Concept                            | v2 Concept                                        | Notes                                                         |
--- a/docs/specs/memory-v2/04-protocol.md
+++ b/docs/specs/memory-v2/04-protocol.md
@@ -485,6 +485,20 @@ type TransactResult =
   | { error: AuthorizationError };
 ```
 
+Revision payloads in the accept response: a `set` revision carries its
+document payload, a `delete` revision is identity-only, and a `patch`
+revision carries `document` — the document's POST-COMMIT state (the same
+state on every patch revision of a doc in the commit, revision-order
+independent; replayed accepts recover it as of the replayed commit's seq,
+not the current head). This lets the client promote its confirmed mirror
+from the value the server actually computed instead of extrapolating its
+own patch onto a possibly-stale local base — the inline verdict can outrun
+the batched novelty fan-out carrying the foreign writes the patch was
+applied on top of (CT-1926). The field is additive: clients fall back to
+local extrapolation when a server omits it, and promote the response value
+only when it differs from that extrapolation (preserving materialized-value
+identity across confirmation in the steady state).
+
 Path conventions on the wire:
 
 - `ClientCommit` reads and writes use full document paths.
--- a/packages/memory/test/v2-accept-postapply-doc-test.ts
+++ b/packages/memory/test/v2-accept-postapply-doc-test.ts
@@ -1,0 +1,223 @@
+// CT-1926: patch-op accept responses carry the document's post-commit state.
+//
+// The client's `confirmPending` promotes its confirmed mirror from this value
+// instead of extrapolating its own patch onto a possibly-stale local base —
+// the inline accept can outrun the batched fan-out carrying the foreign
+// writes the patch was applied on top of, and a locally-extrapolated
+// promotion mints a (seq, value) pair the server never had, feeding
+// dishonest staleness bases into later reads. `set` revisions keep carrying
+// their own payload; `delete` stays identity-only. Replayed accepts recover
+// the same document via a seq-pinned read, so a replay observed after later
+// commits landed still reports the state as of the replayed commit.
+
+import { assertEquals } from "@std/assert";
+import { toFileUrl } from "@std/path";
+import { applyCommit, close, type Engine, open } from "../v2/engine.ts";
+import { type EntityDocument } from "../v2.ts";
+
+const createEngine = async (): Promise<{ engine: Engine; path: string }> => {
+  const path = await Deno.makeTempFile({ suffix: ".sqlite" });
+  const engine = await open({ url: toFileUrl(path) });
+  return { engine, path };
+};
+
+const invocationFor = (
+  localSeq: number,
+  extra: Record<string, unknown> = {},
+) => ({
+  iss: "did:key:alice",
+  aud: "did:key:service",
+  cmd: "/memory/transact",
+  sub: "did:key:space",
+  args: { localSeq, ...extra },
+});
+
+const authorization = {
+  signature: "sig:alice",
+  access: { "proof:1": {} },
+};
+
+const toEntityDocument = (
+  value: unknown,
+): EntityDocument => ({ value } as EntityDocument);
+
+Deno.test("memory v2 engine: a patch accept carries the post-apply document, including foreign writes the client has not seen", async () => {
+  const { engine, path } = await createEngine();
+  try {
+    // Foreign winner establishes the container the client's mirror lacks.
+    applyCommit(engine, {
+      sessionId: "session:other",
+      invocation: invocationFor(1, { actor: "other" }),
+      authorization,
+      commit: {
+        localSeq: 1,
+        reads: { confirmed: [], pending: [] },
+        operations: [{
+          op: "set",
+          id: "entity:list",
+          value: toEntityDocument({ items: ["a", "b"] }),
+        }],
+      },
+    });
+
+    // The session's blind append is applied against the winner's container.
+    const applied = applyCommit(engine, {
+      sessionId: "session:1",
+      invocation: invocationFor(1),
+      authorization,
+      commit: {
+        localSeq: 1,
+        reads: { confirmed: [], pending: [] },
+        operations: [{
+          op: "patch",
+          id: "entity:list",
+          patches: [{ op: "add", path: "/value/items/-", value: "X" }],
+        }],
+      },
+    });
+
+    assertEquals(applied.revisions.length, 1);
+    assertEquals(applied.revisions[0].op, "patch");
+    // The response tells the client what the doc actually became — the
+    // winner's items plus the append — not just the patch it already knew.
+    assertEquals(
+      applied.revisions[0].document,
+      toEntityDocument({ items: ["a", "b", "X"] }),
+    );
+  } finally {
+    close(engine);
+    await Deno.remove(path).catch(() => {});
+  }
+});
+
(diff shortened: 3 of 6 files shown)


ubik2 (commented):

Self-review of the CT-1926 changeset before requesting human review. Scope: the three commits on this PR · coverage: deep-read, adversarially targeting the least-trusted spots of code written in-session · checks run: memory 459/459, runner 1101/1101, `deno task check` (runner package + root), `check-docs` 523 blocks, `deno fmt` clean.

## 🔴 Blocking

None found.

## 🟡 Improvements — found and fixed on the branch

1. **Old-server fallback arm had no coverage** (9c8018c85). Once the scripted server model started attaching post-apply documents, the pre-CT-1926 fallback path in `confirmPending` (patch revision without `document` → local extrapolation) became unreachable in tests — a test-rigor gap and a coverage-ratchet risk. The model gained `omitPostApplyDocuments`, and a regression runs the identical foreign-interleave scenario against it, pinning that an old-server accept degrades to today's extrapolation (the documented residue — confirmed misses the foreign fields until fan-out corrects it), not to anything silent.

## Verified, not just believed

- **`valueEqual` with absent documents**: the delete-vs-delete comparison passes `undefined, undefined`, resolved by the `Object.is` fast path — no throw, correct `true`.
- **Replay reconstruction under snapshot retention**: `reconstructPatchedDocument` does not require a retained snapshot — it falls back to the last durable `set` revision plus the patch fold, so replayed accepts at old seqs cost more but never error, and report the state as of THEIR commit (pinned by test).
- **Added cost is bounded**: the per-accept reconstruction folds at most ~`DEFAULT_SNAPSHOT_INTERVAL = 10` patches per written doc. Named in the PR body with the natural optimization (feed `materializeSnapshots` from the same reconstruction) if it ever profiles.
- **No hidden consumers**: the server returns `AppliedCommit` verbatim and nothing else reads `AppliedRevision.document` (fan-out runs off `markSpaceDirty`) — attaching documents to patch revisions changes no delivery semantics.

## Test rigor

The pre-existing identity-stability pin ("confirming the head pending write promotes the cached materialization") forced a design improvement mid-implementation: unconditional promote-to-server-truth broke it, which produced the landed `valueEqual` rule — identity survives confirmation when the values agree, server truth wins when they differ (identity changes iff the value does). The pin and the new convergence regression now pass together, and the new-server/old-server tests form a minimal pair on the identical scenario. The convergence regression closes the CT-1872 "unpinned convergence property" for the accept path.

## Coherence

04-protocol §4.3.1 (response revision payloads) and 03-commit-model §3.12 (accept-side promotion rule) moved with the semantics; the CT-1926 ticket records the `valueEqual` refinement superseding the original "drop the cache unconditionally" wording.

## Remaining review items for humans

- The CFC read-authority checkpoint (write-only shapes should not receive other writers' content in accept responses) — the engine attaches unconditionally, consistent with the existing `set`-payload echo and query access; flagged for a server-layer judgment, not changed here.
- The reconstruction cost note above, if anyone has a contended-write workload handy.

**Verdict:** sound — no blockers; the one improvement found is already applied.

ubik2 (commented):
_cf-review via Codex (GPT-5), on behalf of @ubik2_

PR 5225 · coverage: deep-read all 6 files · checks run: focused memory tests 3/3, focused runner tests 73/73, targeted `deno check`, `deno fmt --check`, and `deno lint`; CI is green.

## 🔴 Blocking / cross-PR dependencies

1. **[Major, verified] The server-truth correction is invisible to subscribers.**

   [`confirmPending`](https://github.com/commontoolsinc/labs/blob/9c8018c8500039b8a2fc22191c329e984ae417c3/packages/runner/src/storage/v2.ts#L4040-L4055) can silently replace the optimistic value when server truth differs. The only successful-commit notification was already emitted for the stale optimistic value before the request.

   In the CT-1926 regression scenario, `get()` therefore returns the corrected value (so the new test passes), but subscribed UI/scheduler consumers and sinks remain on the stale optimistic value when fan-out never arrives. Fix by diffing visible state across confirmation and notifying subscribers/sinks when it changes. The regression should assert the corrected notification, not only `expectVisible()`.

2. **[Major coherence, verified] The normative protocol types still contradict the new wire response.**

   [`04-protocol.md`](https://github.com/commontoolsinc/labs/blob/9c8018c8500039b8a2fc22191c329e984ae417c3/docs/specs/memory-v2/04-protocol.md#L465-L500) says `transact` returns `Commit` containing `StoredRevision[]`, while the implementation returns `AppliedCommit` / `AppliedRevision` with `document` and `patches`. The new prose contradicts the immediately preceding type definition, and the stale shape is repeated in the [memory-v2 README](https://github.com/commontoolsinc/labs/blob/9c8018c8500039b8a2fc22191c329e984ae417c3/docs/specs/memory-v2/README.md#L188-L212).

   These may be addressed by the planned companion PR, but that PR needs to land first or atomically with this one; otherwise PR 5225 temporarily ships the silent reactive divergence and actively incorrect live protocol types.

## 🟡 Improvement

- [`attachPostApplyDocuments`](https://github.com/commontoolsinc/labs/blob/9c8018c8500039b8a2fc22191c329e984ae417c3/packages/memory/v2/engine.ts#L6265-L6305) repeats the complete final document on every patch revision. Multiple tiny patches against one large document therefore produce `patch-count × document-size` output. Attaching it only to the final revision per document, or using a commit-level document map, avoids that amplification. This can safely remain follow-up work.

**Verdict:** the implementation and focused tests otherwise behave as claimed, but the notification correction must be present before this behavior is independently merge-safe. I’ll re-check these concerns against the companion PR once it is available.

ubik2:
_Claude Code, on behalf of @ubik2._

Closing as **superseded by the marker-parking line of work** (CT-1927, completed and pinned by CT-1950 / PR #5370).

The semi-committed window this PR targeted — verdict in hand, sync not yet delivered, `confirmPending` extrapolating onto a stale mirror — is now closed structurally for watched docs: an accept's promotion parks until the catch-up marker covers it, so by promotion time the frame has delivered the foreign novelty the accept was applied on top of. That behavior is regression-pinned (promotion-at-marker, no-double-apply, and the commit-promise coverage contract), and the CT-1926 ticket's own rescope anticipated it ("the response value is a guarded no-op for watched docs").

What remains of this PR's surface is the residue: unwatched docs (no marker channel; verdict-time extrapolation persists, but such docs have no subscribers to notify and feed no basis claims) and pre-marker servers (transitional — `verdictCatchUpMarkers` is build-inherent). The review's main blocker illustrates the mismatch: the subscriber-notification machinery it correctly demands would serve a path that is inert on watched docs and subscriber-free on unwatched ones.

The branch (`robin/ct-1926-accept-doc`) stays as prior art — the engine slice's wire shape for post-apply documents is the natural starting point if CT-1872's rejection-side `actual` values (spec §3.12) ever want a response-carried truth source.

PR #5235, opened 2026-07-31
docs: prefer narrow cross-pattern contracts

bfollington:
## Summary

- document consumer-owned minimal schemas for external pattern data
- add narrow-contract guidance to the pattern development prompt
- add cross-pattern data contracts as pattern-critic category 17
- distinguish shallow shared role models from full Input/Output reuse

## Validation

- deno fmt --check
- deno lint
- deno task check-docs
- deno task check-skill-facts
- pattern-critic skill metadata validation

---
## Summary by cubic
Documented narrow cross-pattern data contracts and updated the pattern-critic to flag oversized external schemas and unnecessary bindings. This reduces coupling, extra traversal, and validation/migration failures.

- **New Features**
  - Added pattern-critic category 17 with clear violations/fixes, exceptions (shared protocols, sub-pattern `"[UI]"`), and severity guidance; updated `skills/pattern-critic/SKILL.md` references.
  - Expanded composition and development guides with a new “Keep External Data Contracts Narrow” section and examples; prefer consumer-owned minimal schemas and shallow role models over full `Input`/`Output` or `Pick`/`Omit` reuse.
  - Clarified external binding contracts in reactivity docs: exporting an `Output` is for factory visibility, not a mandate for consumers; `Writable<>` and `Cell<>` are equivalent aliases—brands preserve live-cell bindings but do not grant authority.

<sup>Written for commit 26c549629178e12f0e1da936bc9111216ae27af6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5235?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/common/ai/pattern-critique-guide.md
+++ b/docs/common/ai/pattern-critique-guide.md
@@ -266,6 +266,38 @@ See `docs/common/patterns/multi-user-patterns.md#presenting-identity` and `docs/
   resolved string/number is the actual defect, and `cf check` would reject the
   handler-state type mismatch if the binding were unresolved.
 
+### 17. Cross-Pattern Data Contracts
+
+Review types used to read externally owned data: linked pieces, wish results,
+registry entries, cells passed between independently evolving patterns, and
+similar boundaries. A schema is a runtime projection and capability contract,
+not just a TypeScript annotation. The consumer should name only the fields it
+reads and the writable/stream capabilities it actually uses.
+
+| Violation | Fix |
+|-----------|-----|
+| another pattern's full `FooInput` or `FooOutput` is imported, aliased, or extended to type external data even though only a subset is used | declare a consumer-owned structural type containing exactly the fields and capabilities used |
+| `Pick<FooOutput, ...>` / `Omit<FooOutput, ...>` is used merely to avoid spelling the consumer contract | write the minimal local shape so unrelated producer type changes do not become consumer migrations |
+| a consumer projection names fields it never reads, including `[UI]`, unrelated state, or mutation streams | remove them; undeclared fields are intentionally outside this relationship |
+| a consumer requests `Writable<>`, `Cell<>`, or a stream it does not mutate, bind, or send | use the plain readable field type or omit the capability |
+| a producer's advertised reusable model contains its full pattern state or grows whenever one consumer needs another field | keep/export a shallow role model for the stable shared semantics; let specialized consumers own additional local projections |
+
+This duplication is intentional. Narrow schemas avoid traversing unrelated
+linked data, reduce validation failures across durable-data vintages, avoid
+exporting excess write capabilities, and let patterns migrate independently.
+See
+`docs/common/patterns/composition.md#keep-external-data-contracts-narrow`.
+
+Do not flag a genuinely shared protocol (for example, a stream event or enum),
+a cohesive domain model co-owned by one pattern family, a test intentionally
+checking the complete public result, or a wrapper whose stated purpose is to
+forward the complete contract. If the consumer demonstrably uses every field
+and capability, the full contract may be honest.
+
+Severity is `minor` when the only current effect is unnecessary coupling.
+Raise it to `major` when the oversized contract causes extra traversal,
+validation/migration failure, or excess write capability.
+
 ## Output Format
 
 The review should be emitted as a structured checklist with explicit pass/fail
--- a/docs/common/ai/pattern-critique-guide.md
+++ b/docs/common/ai/pattern-critique-guide.md
@@ -270,33 +270,38 @@ See `docs/common/patterns/multi-user-patterns.md#presenting-identity` and `docs/
 
 Review types used to read externally owned data: linked pieces, wish results,
 registry entries, cells passed between independently evolving patterns, and
-similar boundaries. A schema is a runtime projection and capability contract,
-not just a TypeScript annotation. The consumer should name only the fields it
-reads and the writable/stream capabilities it actually uses.
+similar boundaries. A schema is a runtime projection and binding contract, not
+just a TypeScript annotation. The consumer should name only the fields it reads
+and the live-cell/stream bindings it actually uses.
 
 | Violation | Fix |
 |-----------|-----|
-| another pattern's full `FooInput` or `FooOutput` is imported, aliased, or extended to type external data even though only a subset is used | declare a consumer-owned structural type containing exactly the fields and capabilities used |
+| another pattern's full `FooInput` or `FooOutput` is imported, aliased, or extended to type external data even though only a subset is used | declare a consumer-owned structural type containing exactly the fields and bindings used |
 | `Pick<FooOutput, ...>` / `Omit<FooOutput, ...>` is used merely to avoid spelling the consumer contract | write the minimal local shape so unrelated producer type changes do not become consumer migrations |
 | a consumer projection names fields it never reads, including `[UI]`, unrelated state, or mutation streams | remove them; undeclared fields are intentionally outside this relationship |
-| a consumer requests `Writable<>`, `Cell<>`, or a stream it does not mutate, bind, or send | use the plain readable field type or omit the capability |
+| a consumer requests a branded live cell (`Writable<>` / `Cell<>`) or stream when it needs only a projected value or does not use the stream | use the plain field type or omit the stream; `Writable<>` and `Cell<>` are equivalent aliases, not different authority levels |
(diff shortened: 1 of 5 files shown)


bfollington (commented):

**Scope** · 4 files, +94/−0, docs + one skill · **Coverage**: deep-read (all four files) + delegated spec/runtime verification of the three load-bearing claims · **Checks run**: `deno task check-docs common` (116 blocks pass), `deno task check-skill-facts` (OK, 34 docs), `deno fmt --check` on touched files (note: `docs/` is excluded from fmt via `deno.jsonc:128`; only `skills/pattern-critic/SKILL.md` is actually in scope, and it passes)

Good change — the repo genuinely had no statement of this rule anywhere, and three of the four surfaces are exactly the ones a pattern author and a critic agent read. Skill mirrors under `.claude/skills/` and `.agents/skills/` are symlinks, so no mirror update was needed. One blocking issue below; it's the reason for the review rather than a rubber stamp.

I verified the three factual claims the guidance rests on rather than taking them on faith. Claim A ("a schema is a runtime projection") is solid. Claim B ("extra fields can fail validation") is right and understated. Claim C ("excess write capability") is stated in the wrong direction against a normative spec — and it's the one the severity escalation rests on.

---

## 🔴 Blocking

### B1. Category 17 tells the critic to flag consumer `Writable<>` as excess write capability, and to escalate severity on that basis. A normative spec says the opposite.

Three places assert it:

- the table row *"a consumer requests `Writable<>`, `Cell<>`, or a stream it does not mutate, bind, or send → use the plain readable field type or omit the capability"*
- the rationale *"avoid exporting excess write capabilities"* (`docs/common/ai/pattern-critique-guide.md:283`)
- `possibly excess write authority` in `skills/pattern-critic/SKILL.md:52`
- and, most consequentially, the severity rule: *"Raise it to `major` when the oversized contract causes extra traversal, validation/migration failure, or **excess write capability**."*

What the specs actually say:

- `docs/specs/space-model/4-cells.md:212` — "`Writable<T>` is a pure type alias for `Cell<T>` — same interface, same runtime object. **It carries no enforced semantics: the runtime does not distinguish the two.**"
- `docs/common/concepts/types-and-schemas/writable.md:151` — the annotation "indicates write intent within that pattern, but doesn't affect how input values are coerced… making them effectively writable regardless of the signature."
- `docs/specs/space-model/4-cells.md:224` — least-privilege narrowing (`ReadonlyCell` / `OpaqueCell` via a transformer step) is explicitly **future work**, "not yet used by pattern code."
- Write authority actually lives in CFC (`writeAuthorizedBy`, keyed on the artifact-hash + symbol-path pair), the space ACL, and the scope-isolation write guard — none of it the consumer's input schema.

`docs/common/concepts/reactivity.md:85` *does* say "Writable<> in a Result Type Grants Write Access… consumers receive a live `Cell` — same identity, write access included." But that is the **producer's result type** — the opposite direction from where this section places the concern.

**Why it matters:** a critic following this guide will emit confidently-wrong `major` findings citing an enforcement mechanism that does not exist, and the guide leaves a normative spec looking wrong to anyone who greps "write capability."

**Suggested fix** — the advice survives, the justification doesn't:

1. Keep the row. Declaring the `asCell` brand *is* a real difference: the consumer receives a live `Cell` rather than a dereferenced value, with exactly the traversal consequences the rest of the section correctly describes.
2. Rewrite the *why* as coupling and intent-signalling, not authority.
3. Drop write-capability from the `major` escalation list. Traversal and validation both stay — they're backed by the runtime.
4. Separately: `Writable<>` vs `Cell<>` is not a meaningful axis at all (`4-cells.md:212`). The axis is *branded vs unbranded*. Worth saying so, since the row currently lists both names as if they differed.

*(verified — read each cited line)*

---

## 🟡 Improvements

### I1. Sharpen the validation claim — the real edge is sharper than the doc claims

"Extra fields can … make otherwise compatible stored values fail validation" understates the mechanism in a way that costs the section its best argument:

- Only **required** fields reject the containing object. A field is required only if it's non-`?`, has no `| undefined`, and no `Default<>` — note `T | undefined` does *not* drop out of `required` (`docs/specs/schema-generator/ts_to_json_schema_mapping.md:432`); it validates because the field's own schema admits `undefined`.
- Inside an array, the failure is **not field-local**: one failing element voids the entire array read. `packages/runner/src/traverse.ts:4180` carries the comment "the ENTIRE array reads as invalid for this caller," added with the index+doc logging after the 2026-07-10 board outage.
- This is deliberately strict today: `37189ba53` ("restore strict required link validation", #4746) removed the earlier element-level grace for absent link targets. The current authority is the header docstring of `packages/runner/test/traverse-required-links.test.ts`.

Citing `traverse-required-links.test.ts` would tie the durable-data-vintage argument to a maintained, greppable source. *(verified)*

### I2. The skill's trigger is narrower than the rule it points at

`skills/pattern-critic/SKILL.md:47` scopes the check to "for every **imported** pattern `Input`/`Output` type." But two of the five table rows need no import at all — "a consumer projection names fields it never reads" and "a consumer requests … a capability it does not use." The single highest-value case, a hand-written `wish<{ … }>()` schema over-declaring its shape, is inline. A critic reading the bullet literally skips it.

Suggest: "for every type describing linked, wished, registered, or otherwise external data — imported or declared inline." *(verified)*

### I3. The `[UI]` example risks a false positive against a hard requirement

The table lists `[UI]` as an example of a field a consumer projection shouldn't name. `docs/common/patterns/composition.md:25` — 45 lines above the new section, in the same file — states sub-patterns **must** include `[UI]` in their Output type. "Consumer projection" carries the distinction for a careful reader, but this table is consumed by an LLM critic that reads both statements. Worth adding "`[UI]` on a sub-pattern's own `Output`" to the existing Do-not-flag list. *(verified)*

### I4. Preempt the obvious objection: `Output` is exported by force, not by invitation

`docs/common/concepts/reactivity.md:105` *requires* producers to export their `Output` type ("a non-exported `Output` interface fails compilation with 'Default export of the module has or is using private name Output'"). So every producer's full contract is necessarily importable — which is precisely why the anti-pattern is tempting. One sentence noting that the mandatory export is not an invitation to consume it wholesale would close the gap for a reader who thinks "but it's exported, so it's public API."

### I5. Cite `MentionablePiece`

`docs/common/patterns/composition.md:103` names it without a path. It's real, and it vindicates the example: `packages/patterns/system/backlinks-index.tsx:20` — six optional fields, no `[UI]`, no pattern state, consumed cross-pattern by `annotation.tsx`. Citing the path turns an abstract example into a checkable one. (Caveat worth knowing: the follow-up to derive backlinks via `crossrefJoin` would delete `backlinks-index.tsx` — which is arguably an argument *for* citing it, so the move surfaces.)

### I6. Cross-link the producer-side statement of the same rule

`docs/common/concepts/reactivity.md:85` is the existing authoritative statement of the capability half of this rule, from the producer's side. The new section is its consumer-side mirror but restates rather than cites it. Linking both directions puts the claim on a concept home instead of duplicating it — and, per B1, is how the direction stays straight.

---

## Possible follow-ups

- `docs/common/conventions/wish.md` and `docs/common/concepts/piece-discovery.md` are what someone reads *while* writing an external-data contract, and neither links the new section. `wish.md:84` already demonstrates the good idiom (producer declares `type Output`, consumer wishes `wish<{ content: string }>`) without naming it as a rule — a natural place to point at the new section.
- Pre-existing and unrelated to this PR: `docs/common/conventions/adding-pieces.md:28` imports `MentionablePiece` from `@commonfabric/piece`, but it's exported from `packages/patterns/system/backlinks-index.tsx`.

---

**Verdict:** fix B1 before merge, then ship. The rule is worth having and the other four surfaces are accurate; B1 is the one place the new guidance would make an agent confidently wrong on someone else's PR.

bfollington:
Addressed in 26c549629:

- Removed the incorrect write-authority framing. The docs now distinguish branded live-cell bindings from plain projections and state that Writable<T> and Cell<T> are equivalent aliases; authorization is not granted by that type choice.
- Removed write capability from the major-severity escalation.
- Clarified required-field validation and the whole-array failure when one element is invalid, with a maintained runtime test citation.
- Broadened the critic prompt to imported or inline external-data schemas.
- Added the [UI] sub-pattern output exception.
- Clarified that Output export is required for TypeScript visibility, not an invitation to reuse the full contract, and cross-linked the producer/consumer guidance.
- Linked the concrete MentionablePiece source.

The wish/piece-discovery links and the pre-existing adding-pieces import are left as the review's optional follow-ups.

PR #5237, opened 2026-07-31
feat(cli): add computed transforms to cf piece get

bfollington:
## Summary

- add `cf piece get --filter <predicate>` with a jq-inspired, array-only predicate language
- add `cf piece get --schema <schema>` with concise fields, inline JSON Schema, and `@file` forms
- compose filtering before projection to reduce JSON returned to callers
- run both transforms through session-scoped runtime filter/map/lift nodes

## CFC semantics

The transform graph uses the same list builtins and computed nodes as authored patterns. Predicate observations label filtered membership, projected reads propagate source labels, and filtered elements retain their source links. Caller projection schemas cannot introduce or override `ifc`, `asCell`, `scope`, or `default` metadata.

## Validation

- targeted CLI and transform suites: 138 steps passed
- CFC regressions cover secret-dependent membership and projected-field propagation
- `deno lint` passed repository-wide
- `deno task check-docs` passed
- `deno task check-skill-facts` passed
- full CLI suite: 1,650 passed; two unrelated existing color-mode assertions failed
- all touched files pass `deno fmt --check`; repository-wide formatting still reports 23 unrelated existing files

---
## Summary by cubic
Adds computed transforms to `cf piece get` with jq-style array filtering and schema-based projection to cut down JSON output and keep CFC labels consistent. This update clarifies projection behavior (array roots, property whitelists, and scalar leaf handling) and improves transform error reporting.

- **New Features**
  - `--filter <predicate>`: jq-inspired predicates over arrays with paths (dot/bracket, negative indices), JSON literals, `== != < <= > >=`, `and or not`, and parentheses. Only `false`, `null`, and missing/`undefined` values are falsey; non-array inputs are rejected.
  - `--schema <schema>`: project with `id,title,author.name`, inline JSON Schema, or `@file`. Concise fields apply per array item; JSON Schema describes the whole output and for arrays must have an array root. Object `properties` act as a whitelist by default (use `"additionalProperties": true` to keep others). Typed scalar leaves that don’t match stored values are omitted, not errors. Combinators and `$ref` are rejected, and callers cannot set `ifc`, `asCell`, `scope`, or `default`.
  - Composes as filter-then-project and runs through session-scoped filter/map/lift nodes so predicate labels carry membership, projected reads propagate labels, and filtered elements retain source links. Clear data-error reporting via `PieceGetTransformError`; CLI help, docs, and tests added.

<sup>Written for commit e112468045653e3df6a3bcd042813f789212ba44. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5237?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/plans/pattern-verb-contract.md
+++ b/docs/plans/pattern-verb-contract.md
@@ -364,15 +364,13 @@ board topic). Until that lands, board-level routing
 (`addComment {topicFid, body}`) is the documented workaround — pragmatic, not
 the target shape.
 
-Two client affordances surfaced in review (2026-07-28) and are deferred,
-blocking nothing: `cf piece get` could grow flags that let an agent control
-how much data a read returns when exploring the fabric interactively — a
-`--schema` override reading through a narrower schema (the runtime's
-`asSchema`; the CLI already narrows its own internal reads this way, e.g.
-the `asSchema` narrowing in `packages/cli/lib/piece-render.ts`), and a limit on the number of records
-returned from a large array. Adjacent CLI work for when board scale demands it,
-recorded here so the deferral is deliberate; neither is an ask on any
-workstream in the implementation plan.
+`cf piece get` lets an agent control how much data an exploratory read returns:
+`--filter` narrows array membership and `--schema` projects fields through the
+runtime's `asSchema` read path. Both execute through computed pattern nodes so
+their CFC behavior matches pattern expressions. A limit on the number of
+records returned from a large array remains deferred, blocking nothing. It is
+adjacent CLI work for when board scale demands it, not an ask on any workstream
+in this implementation plan.
 
 **A set of verbs wants to be a structural interface, and the machinery for that
 already exists.** Schemas are the type system, so a verb set is a schema
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -131,8 +131,8 @@ cf piece get --piece ID items \
 `--filter` is jq-inspired rather than a full jq interpreter. It applies only to
 arrays and accepts value paths (`.status`, `.author.name`, `.["display-name"]`,
 `.tags[-1]`), JSON literals, `==`, `!=`, `<`, `<=`, `>`, `>=`, `and`, `or`,
-`not`, and parentheses. The predicate must produce a boolean. Filtering happens
-before schema projection.
+`not`, and parentheses. Like jq, only `false` and `null` are falsey, so a
+missing path simply does not match. Filtering happens before schema projection.
 
 `--schema` accepts one of three forms:
 
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -132,7 +132,8 @@ cf piece get --piece ID items \
 arrays and accepts value paths (`.status`, `.author.name`, `.["display-name"]`,
 `.tags[-1]`), JSON literals, `==`, `!=`, `<`, `<=`, `>`, `>=`, `and`, `or`,
 `not`, and parentheses. Like jq, only `false` and `null` are falsey, so a
-missing path simply does not match. Filtering happens before schema projection.
+missing path simply does not match. A stored `undefined` is treated like a
+missing value and is also falsey. Filtering happens before schema projection.
 
 `--schema` accepts one of three forms:
 
@@ -145,7 +146,10 @@ Schema describes the complete returned value, so a schema combined with
 `--filter` must have an array root. Object `properties` are a whitelist by
 default; use `"additionalProperties": true` to retain unspecified properties.
 Projection schemas support structural `properties`, `items`, and scalar leaf
-schemas. Schema combinators and references are rejected.
+schemas. In an array-item projection, a scalar leaf whose declared type does not
+match the stored value is omitted by the runtime rather than reported as an
+error; prefer `true` leaves unless that type filtering is intentional. Schema
+combinators and references are rejected.
 
 Both transforms run as a short-lived computed pattern in the caller's session.
 The runtime's list filter/map builtins therefore handle CFC exactly as authored
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -116,6 +116,45 @@ standard error and continues searching that piece and the rest of the space.
 - The launcher spawns the child CLI with `deno run --quiet` so Deno's own
   warnings (npm "Ignored build scripts" banner) never reach users.
 
+### Transforming `piece get` output
+
+`piece get` can filter an array before it reaches stdout and project the result
+to a smaller shape:
+
+```bash
+cf piece get --piece ID items --filter '.status == "open"'
+cf piece get --piece ID items \
+  --filter '.status == "open" and .score >= 10' \
+  --schema id,title,author.name
+```
+
+`--filter` is jq-inspired rather than a full jq interpreter. It applies only to
+arrays and accepts value paths (`.status`, `.author.name`, `.["display-name"]`,
+`.tags[-1]`), JSON literals, `==`, `!=`, `<`, `<=`, `>`, `>=`, `and`, `or`,
+`not`, and parentheses. The predicate must produce a boolean. Filtering happens
+before schema projection.
+
+`--schema` accepts one of three forms:
+
+- a comma-separated field projection such as `id,title,author.name`;
+- an inline JSON Schema object;
+- `@path/to/schema.json`.
+
+For an array result, the concise form describes each item. An inline/file JSON
+Schema describes the complete returned value, so a schema combined with
+`--filter` must have an array root. Object `properties` are a whitelist by
+default; use `"additionalProperties": true` to retain unspecified properties.
+Projection schemas support structural `properties`, `items`, and scalar leaf
+schemas. Schema combinators and references are rejected.
+
+Both transforms run as a short-lived computed pattern in the caller's session.
+The runtime's list filter/map builtins therefore handle CFC exactly as authored
+pattern expressions do: predicate observations label array membership,
+projection reads propagate labels, and filtered elements retain their source
+links. The source cell's schema remains authoritative for Common Fabric
+metadata. A caller cannot introduce or override `ifc`, `asCell`, `scope`, or
+`default` through `--schema`.
+
 ## Built Binary
 
 `deno task build-binaries cf` compiles the CLI to `dist/cf` — fully
--- a/packages/cli/commands/piece.ts
+++ b/packages/cli/commands/piece.ts
@@ -44,6 +44,11 @@ import { UI } from "@commonfabric/runner";
 import ports from "@commonfabric/ports" with { type: "json" };
 import type { PiecePatternRef } from "@commonfabric/piece/ops";
 import { reservesStdoutForCommandOutput } from "../lib/json-output.ts";
+import {
+  parsePieceGetFilter,
+  parsePieceGetProjection,
+  PieceGetTransformError,
+} from "../lib/piece-get-transform.ts";
 
 // Hint system: print helpful next-step suggestions after operations
 let quietMode = false;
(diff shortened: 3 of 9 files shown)


bfollington (commented):
cf-review via Claude `claude-opus-5`, on behalf of @bfollington

**Scope** · 8 files, +1763/−9, one package (`packages/cli`) · **Coverage**: deep-read of `lib/piece-get-transform.ts`, the `commands/piece.ts` + `lib/piece.ts` wiring, and both test files; prior-art search across `data-model`/`runner`/`docs` · **Checks run**: `deno fmt --check` and `deno lint` (clean), `deno check` on both new files (clean), `deno test packages/cli/test/piece-get-transform.test.ts` (9 steps pass), plus 4 scratch probes against an emulated runtime.

The CFC framing is the strongest part of this change, and the two label tests guard exactly the right principle. Three things want fixing before it leaves draft.

---

## 🔴 Blocking

### 1. Overlapping concise field paths leak a sibling the caller never asked for

`packages/cli/lib/piece-get-transform.ts:583` — **verified**

`if (existing === true) continue;` continues the *segment* loop, not the path. When a shorter prefix has already selected a whole subtree, the remaining segments get attached to the current node instead of being abandoned:

```
--schema 'a.b,a.b.c'          →  {a: {b: true, c: true}}     // wanted {a: {b: true}}
--schema 'author,author.name' →  {author: true, name: true}
```

Probe against a live runtime, source `{a: {b: {c: 1, d: 2}, c: "LEAKED-SIBLING"}}`:

```
PROJECTED => {"a":{"b":{"c":1,"d":2},"c":"LEAKED-SIBLING"}}
```

Beyond the wrong output: the same schema feeds `selectSourceSchema`, so the narrowed source read widens too — a phantom key joins the read surface the CFC label story depends on. The README states object properties are "a whitelist by default"; this breaks that. Fix: `break` out of the segment loop rather than `continue`.

### 2. An object-rooted JSON `--schema` over an array source silently returns `{}`

`packages/cli/lib/piece-get-transform.ts:915` — **verified**

`isRecord` from `@commonfabric/utils/types` is `typeof value === "object" && value !== null` — **true for arrays**. So `projectValue`'s `if (!isRecord(value)) throw` guard never fires on an array; the object branch runs `Object.hasOwn(array, "id")`, finds nothing, and emits an empty object. No error, no warning:

```
cf piece get … items --schema '{"type":"object","properties":{"id":true}}'
OBJECT SCHEMA OVER ARRAY => {}
```

Same root cause at `piece-get-transform.ts:507`: a JSON *array* from `@file` passes the "expected a JSON Schema object" check and becomes `{"0":"not","1":"a","2":"schema"}`, silently accepted as a projection schema (verified). Fix: add an explicit `Array.isArray` rejection at both guards — `isRecord` is not the right predicate here.

### 3. A live plan document is left saying this feature is deferred

`docs/plans/pattern-verb-contract.md:367` — **verified**

> "Two client affordances surfaced in review (2026-07-28) and are deferred… a `--schema` override reading through a narrower schema… recorded here so the deferral is deliberate."

`docs/plans/` is live documentation, and this PR lands one of the two deferred affordances. Per `docs/README.md` the same change must update it — otherwise someone searching for "`piece get --schema`" lands on an authoritative doc saying it doesn't exist yet. Small edit: mark `--schema`/`--filter` landed, leave the record-limit deferral standing.

---

## 🟡 Improvements

### 4. `maskFromPaths`' numeric branch is dead weight

`packages/cli/lib/piece-get-transform.ts:770-780` — **verified by read**

It emits `{type: "array", prefixItems: [...]}`, but every consumer reads only `.items`: `selectSourceSchema` uses `mask.items ?? true` (`:852`) and `mergeMasks` likewise (`:801`). Index-specific narrowing therefore degrades to `true` — safe (over-read, never under-read), but ~20 lines including a `max > 100` guard that cannot affect any read. Drop it, or make the consumers honor `prefixItems`.

### 5. The transform path materializes the value twice more than it needs to

`packages/cli/lib/piece.ts:2210`

`getCellValue` still runs the ordinary `piece[prop].get(path)` and discards the result, then `derivePieceGetValue` does a second full `sourceCell.pull()` (`piece-get-transform.ts:1024`) only to test `Array.isArray`. On the large arrays these flags exist for, that is two full reads of exactly the data you are trying not to return — and it partly undoes #5222 / #5229's narrowing work on this same call path.

### 6. `--schema '{"type":"object"}'` silently returns `{}`

`packages/cli/lib/piece-get-transform.ts:539` — **verified by read**

`normalizeProjectionSchema` only defaults `additionalProperties: false` when `properties` is present; with neither, `projectValue` iterates zero properties and emits `{}`. Reject a properties-less object schema, or treat it as `true`.

### 7. A predicate path missing on one element fails the whole command

`packages/cli/lib/piece-get-transform.ts:399`

`--filter '.active'` over `[{id: 1, active: true}, {id: 2}]` aborts with *"predicate must evaluate to a boolean, got null"*. The error surfacing itself is correct — I confirmed the `runtimeErrorLog` wiring catches the in-lift throw and reports it, with no silent element drop. But jq treats null as falsy, and `--filter '.active'` over a heterogeneous array is close to the first thing an agent will try. Worth deciding deliberately and documenting either way.

---

## Questions

- The verb-contract plan proposed this as **the runtime's `asSchema` narrowed read**, citing `packages/cli/lib/piece-render.ts:41` as existing prior art. This PR instead hand-rolls `projectValue` / `selectSourceSchema` / `projectionMask` / `mergeMasks` (~500 lines of schema-directed value projection). What did `asSchema` not cover? The obvious answer is that a caller schema must not carry CFC metadata — but `normalizeProjectionSchema` already strips exactly that, so a normalized schema handed to `asSchema` looks close. If the hand-rolled version is genuinely needed, that reasoning belongs in the PR body and in a comment: this is the shape of duplication the repo pays for repeatedly.
- `deps.onResultCell` exists only for the two label tests. Intentional test seam, or would `derivePieceGetValue` returning the result cell be cleaner?

## Possible follow-ups

- The plan's second deferred affordance — a record limit on large arrays — is the natural companion to `--filter`; worth filing now that this half landed.
- The concise-path builder and `projectionMask` / `mergeMasks` want property-based tests (overlap, prefix, ordering) rather than examples.

## Test rigor

Good shape: BDD + `expect()`, and the two CFC tests guard a real principle (predicate observations tainting membership; projected-field labels propagating from source metadata), not implementation detail. The gaps line up with the findings above — nothing covers overlapping concise paths, an object schema over an array source, or `--schema` composed with `--input` / `--step`.

---

**Verdict**: right architecture and the CFC story is well argued, but two verified projection bugs silently return wrong data and a live plan doc still says the feature is deferred. Fix those three and this is ready for review.

bfollington:
Addressed the review in 0e12478fa:

- fixed concise-path prefix overlap so author,author.name cannot widen to a sibling, with both ordering and live-runtime regressions
- reject array-valued schema documents and reject JSON schema roots that mismatch the selected value shape
- treat an object schema without properties as an identity read
- use jq truthiness; missing predicate paths are null and therefore non-matching
- removed the dead numeric prefixItems mask
- removed the discarded ordinary read from the transform path
- changed projection to identity computed nodes over runtime asSchema reads; source-selected schemas retain authoritative CFC metadata and caller CFC keys remain rejected
- updated the live verb-contract plan and CLI/skill guidance
- expanded tests from 9 to 24 transform scenarios, including the actual command action, overlap, root mismatches, jq truthiness, transaction/runtime failures, and CFC propagation

I kept onResultCell as an intentional test seam: it lets the CFC assertions inspect the ephemeral computed result cell without changing the production return contract of derivePieceGetValue.

Focused formatting, type checks, diff checks, and 157 test steps pass locally. The prior full CLI run remains 1,650 passing with only the two existing Cliffy color-mode failures. Watching the refreshed coverage job now.

bfollington:
Follow-up: refreshed CI is fully green. Coverage Check and the final Status job both pass on 0e12478fa.

bfollington (commented):
cf-review (round 2) via Claude `claude-opus-5`, on behalf of @bfollington — re-review of `0e12478fa`

**Scope** · the fix commit (+854/−271 across 8 files) against my round-1 findings · **Coverage**: full re-read of `lib/piece-get-transform.ts`, the restructured `getCellValue`, and the doc/README/skill edits; skimmed the 15 new test cases · **Checks run**: `deno check` on all 5 touched TS files (clean), `deno fmt --check` + `deno lint` (clean), `piece-get-transform.test.ts` (24 steps pass), `piece.test.ts` + `piece-call.test.ts` (133 steps pass), plus 11 fresh probes against an emulated runtime.

**All seven round-1 findings are verified fixed.** The headline is the answer to the `asSchema` question: `projectValue` is gone entirely (0 occurrences; −220 lines in the lib), replaced by identity lifts whose *input* schema is the source-derived read schema. Projection now happens on the read side, through the runtime, which is what the plan called for. That is a better change than what I asked for.

## Verified fixed

| # | Finding | Evidence |
|---|---|---|
| 1 | Concise overlap leaked a sibling | `a.b,a.b.c` → `{a:{b:true}}`; `author,author.name` → `{author:true}`. Live-runtime probe no longer emits the sibling. |
| 2 | Object schema over array returned `{}` | Now errors in both directions — *"A JSON --schema for an array value must describe the returned array"* / *"An array-rooted JSON --schema can only be applied to an array value."* Array-valued `@file` schemas rejected too. |
| 3 | Live plan doc said `--schema` was deferred | `docs/plans/pattern-verb-contract.md:367` rewritten; the record-limit deferral correctly kept standing. |
| 4 | Dead `prefixItems` mask branch | Removed; a numeric head now widens to `true` (over-read, never under-read). |
| 5 | Value read twice and discarded | Transform branch returns early in `getCellValue`; the `PieceResultProjectionError` guards are preserved on both routes. |
| 6 | `{"type":"object"}` returned `{}` | Normalizes to `{type:"object", additionalProperties:true}` — identity read. |
| 7 | Missing predicate path aborted the command | jq truthiness. Probe over `[{active:true},{},{active:false},{active:0},{active:null}]` returns exactly `id 1` and `id 4` — `0` truthy, missing/`false`/`null` not. Documented in README and SKILL.md. |

The load-bearing new claim also holds. Narrowing is real, not just declarative — probed against three source shapes:

```
concise "id,title" over typed array                  => [{"id":1,"title":"a"},{"id":2,"title":"b"}]   (secret dropped)
concise "id,nested.a" over additionalProperties:true => {"id":9,"nested":{"a":1}}                     (secret, nested.b dropped)
concise "id" over a source with NO schema            => [{"id":1},{"id":2}]                           (extra dropped)
```

---

## 🟡 One new thing, introduced by the rewrite

### A type-mismatched leaf in a JSON `--schema` is silently dropped

`packages/cli/lib/piece-get-transform.ts:1051` — **verified**

Moving projection onto the runtime read means a schema that declares the wrong type for a field does not fail — the read just returns nothing for it, exit 0:

```
--schema '{"type":"array","items":{"type":"object",
           "properties":{"id":{"type":"string"},"title":true}}}'
   over items whose id is a number

=> [{"title":"a"},{"title":"b"}]          # id vanishes, no error
```

An agent that guessed `string` where the value is a number reads this as "the field is absent." Two mitigations keep it off the blocking list: it cannot happen through the concise form (leaves are always `true`), and it *is* consistent `asSchema` semantics now that `asSchema` is deliberately the mechanism — which argues for documenting rather than special-casing. Suggest a line in the README's `--schema` section: a declared type that does not match the stored value omits the field rather than erroring, so prefer `true` leaves unless you are deliberately type-filtering.

## ⚪ Nits

- `predicateTruthiness` (`:399`) is `value !== false && value !== null` — an explicitly-stored `undefined` is truthy. `valueAtPredicatePath` normalizes missing keys to `null`, so this only bites a stored `undefined`; `value != null` closes it.
- `mergeMasks(leftChild as PredicateMask, rightChild)` (`:813`) — sound, because `maskFromPaths` only ever produces object-or-`true`, but the cast carries an invariant the types do not express. Parameterizing `ObjectProjectionMask` over its child type would let it go away.

## Test rigor

9 → 24 cases, and the additions guard the right things: `collapses overlapping concise paths without exposing siblings` and `does not leak nested siblings…` are fix-revert-shaped regressions for finding 1, root-mismatch and jq-truthiness cases cover 2 and 7, and the commit-failure / predicate-failure cases fill the error paths I had flagged as untested. The two CFC tests survived the architecture change unmodified, which is the useful signal that the label semantics did not move.

---

**Verdict**: everything raised is fixed, the projection rewrite is a genuine simplification rather than a patch, and CI is green — ready to come out of draft. The type-mismatch drop deserves a README sentence, not a code change.

bfollington:
Addressed the round-two follow-up in e11246804:

- documented that a type-mismatched scalar leaf is omitted specifically during array-item projection, and added a runtime regression for the reviewed example
- made explicitly stored undefined values falsey, matching missing-value behavior, with a predicate regression
- parameterized the recursive object-mask shape so mergeMasks no longer needs the PredicateMask cast

Focused formatting, type checks, all 24 transform scenarios, and diff validation pass.

bfollington:
Round-two follow-up CI is fully green on e11246804, including Coverage Check and final Status.

PR #5238, opened 2026-07-31
Allow generated required outputs during pattern upgrades

bfollington:
## Summary

- distinguish required pattern inputs from generated required outputs during schema compatibility checks
- keep newly required inputs and unclassified CFC paths strict while allowing runner-recorded output paths to materialize on replay
- replace the stream-shape migration exemption with explicit output provenance and add upgrade/state-continuity coverage

Related: CT-1904

## Testing

- `deno test -A packages/piece/test/schema-compatibility.test.ts packages/runner/test/cfc-schema-merge.test.ts packages/runner/test/cfc-additive-default-preserves-old-doc.test.ts packages/piece/test/state-continuity.test.ts packages/piece/test/check-update-default-pattern.test.ts`
- `deno test -A tasks/pattern-vintage-run.test.ts`
- `deno test -A packages/patterns/system/home.vintage-defaults.test.ts packages/runner/test/cfc-additive-default-preserves-old-doc.test.ts`
- `deno task pattern-compat --only system/home.tsx`
- `deno lint`
- changed-file `deno fmt --check`
- `git diff --check`

## Known repository state

Repository-wide `deno fmt --check` reports 23 pre-existing formatting violations outside this changeset; all changed files pass formatting checks.

---
## Summary by cubic
Allows newly required generated outputs to materialize during pattern upgrades by marking result writes as outputs and scoping the CFC merge exemption to those paths. Inputs and ordinary document data still require defaults; cross-space children and extracted collection entities inherit the output role. Related: CT-1904.

- **New Features**
  - Marks result projection writes as `schemaRole: "output"` end to end (result root, built-in bindings, and `setRawUntyped`), and propagates to extracted `[ID]` entity documents via `diffAndUpdate`.
  - `prepare` derives per-target generated-output paths and passes them to `mergeCfcSchemaEnvelopes`; newly required fields under those paths are accepted without defaults, including full-document rewrites via path `[]`.
  - Piece compatibility accepts newly required result fields; arguments remain strict. Vintage gate uses required input evolution for red/green and replays generated-output evolution; Home’s result defaults are pinned as contract-stable.

- **Refactors**
  - Replaces the stream-slot heuristic with explicit generated-output path handling; no inference from stream capability. The exemption is scoped only to declared paths.
  - Updates tests to pin scoping and entity-output migration outcomes and aligns docs to clarify directional compatibility (inputs need defaults; generated results do not).

<sup>Written for commit 51792d310bf1aec1c0c2b97a742f88b5b2fdd596. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5238?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bfollington (commented):

**Scope** · 19 files / +363 −328 across `piece`, `runner/cfc`, `patterns/system`, `tasks`, docs · **Coverage**: deep-read of all runtime/piece source + tests; docs sampled · **Checks run**: `deno test -A packages/runner/test/cfc-*.test.ts` (144 passed), `deno test -A` on the three changed piece/patterns suites (green), `deno fmt --check` + `deno lint` on changed files (clean), plus three differential probes against `main`.

The core idea holds up. I could not find a case where it loses data: my probe that a nested required addition inside an `asCell` collection silently voids the array reproduced **identically on `main`** — pre-existing, not a regression from this PR.

No blocking defect found.

## 🟡 Improvements

### 1. `packages/patterns/system/home.vintage-defaults.test.ts` — the new rationale comment states something verifiably false *(verified)*

> "removing an existing default is still a contract change … **and the compatibility gate rejects that removal**."

It does not. Measured against this branch's `assertPatternSchemasBackwardCompatible`:

```
REMOVE RESULT DEFAULT   => NO THROW
REMOVE ARGUMENT DEFAULT => NO THROW
```

This matters more than a typo: that sentence is now the *entire* justification for the source-text pins in this file. A reader who believes the gate catches default removal will conclude the pins are redundant and delete them — and the pins would then be the only thing that was protecting the contract. Fix: say the opposite — nothing else catches this, which is *why* the pins exist.

### 2. `packages/runner/src/cfc/schema-merge.ts` — the documented granularity is not the production granularity *(verified)*

The JSDoc on `MergeCfcSchemaEnvelopeOptions` says "Required fields outside these paths still need defaults", and the new unit test exercises `generatedOutputPaths: [["generated"]]`. But `updateResultProjection` records the output role at the **result root**, so `canonicalizeLogicalPath([])` is `[]`, and `generatedOutputCovers` returns `true` for every path in that document at every depth.

Measured — a required, defaultless field added two levels below the result root, inside a plain nested object:

| | `main` | this branch |
|---|---|---|
| replay error | `cfc-schema-migration-incompatible: required field b needs a default…` | *(none)* |
| `meta` after replay | `undefined` | `{"a":"x","b":"y"}` |

The behavior is *fine* — setup rewrites the whole result value, so nothing at rest is lost — but "output-path-scoped" describes an exemption that a pattern result never actually gets. Either record the role at the specific result field paths (as `recordOutputSchemaPolicyInputs` already does for bindings), or state plainly that a pattern result document is exempt wholesale and keep the path machinery for the binding sites that do use it.

### 3. `packages/runner/src/runner.ts:1218` — the new argument-side recording is redundant, and its `"input"` role is inert *(verified)*

`schemaRole: "input"` is never consulted anywhere — `prepare.ts:824` only tests `!== "output"`, so `"input"` and *absent* are identical. And the call is a duplicate: `argumentLink` comes from `newArgumentCell.asSchema(pattern.argumentSchema)`, so the `argumentCell.set(argument)` two lines below already records the same `{target, schema}` via `cell.ts:1365`. Net effect is one extra entry in `state.writePolicyInputs`, one extra `writePolicyIdentitiesByTarget` entry, and one extra pass through `candidateSchemasByTarget` — for no signal. Drop it, or make `"input"` load-bearing so the strictness is asserted rather than implied by absence.

The result-side call at `runner.ts:1483` has the same duplication (`setRawUntyped` → `cell.ts:2271` records the same link and schema), but there it *is* load-bearing, because the existing record carries no role. Threading the role through that existing call site would be cleaner than recording twice — as written, the two records only agree by coincidence of `link.schema ?? this.schema` resolving to `pattern.resultSchema`.

### 4. `tasks/pattern-vintage-run.test.ts` — the vintage gate's red/green control no longer exercises the guard it was built around *(verified)*

`BREAKING` moved from an additive required output to an additive required *input*, and the asserted failure moved with it: `"required field addedLater needs a default to preserve old documents"` → `"updated arguments do not match the candidate schema"`. That is stored-argument validation, a different subsystem — so the end-to-end gate now has zero coverage of the CFC schema-merge migration path, and it red-lights on a class Tier 1 already rejects outright.

Real coverage of the guard does survive (`cfc-additive-default-preserves-old-doc.test.ts` still drives a genuine rejection through the production wiring, and I confirmed it passes), so this isn't a hole — but `docs/plans/pattern-update-state-continuity.md` still bills that row as "the gate catches a real break", which is now a weaker claim than it reads.

## Questions

- **Does dropping the migration default give up rollback / generation-skew protection?** A `Default<[]>` protected two directions: old documents *and* old readers. Once the merged envelope persists `required: [… "favorites"]` without a default, a doc written by a still-running older pattern instance — a reverted deploy, a stale replica — is invalid against it, with no default to synthesize the missing value. That is the class #4668's grace exists for. Was this weighed, and is the answer "the grace covers it"?
- **Which write paths can introduce a required stream slot into a CFC-relevant doc *without* an output-role input?** Removing `isStreamSlot` swaps a value-shape exemption for a provenance one, so any such path now rejects where `main` exempted it. Setup-projected results are covered (home's full materialization test passes), but e.g. `data-updating.ts:929` records array-entry schemas with no role. Is there a reachable case, or is setup genuinely the only producer of stream slots?

## Possible follow-ups

- `packages/runner/test/cfc-additive-default-preserves-old-doc.test.ts` header still narrates the pre-change story ("the six exported handler streams — which cannot carry a meaningful `Default<>`"). Not wrong, but it now explains a mechanism the file no longer tests.
- `mergeSchemaNode` builds logical-path segments from `"*"` (both `items` and `additionalProperties`) and `String(index)` for slot claims. Harmless while production only passes `[]`, but if the recording is ever tightened to real field paths, a literal `"*"` property name or a branch index could false-match.

## Verdict

Semantic change is sound, all touched suites are green, and the one data-loss shape I chased reproduces identically on `main`. Worth fixing the false claim in finding 1 and settling the rollback question before merge; 2–4 are cheap and worth the same pass.

bfollington:
Addressed in `7be83a7ec`.

- Fixed the false Home-default rationale: the generic compatibility gate does not catch default removal, so the source pins are explicitly documented as the protection.
- Documented and tested the actual result-root granularity. A root output marker (`[]`) covers the whole result document because setup rewrites the complete projection; narrower output-binding paths remain path-scoped.
- Removed the duplicate argument-side policy record and the inert `"input"` role. The result role is now threaded through the existing `setRawUntyped` write, so there is one schema record rather than two coincident records.
- Reworded the vintage red/green control and live plan so they say exactly what it proves: rejection of an unbound required input. The production-wiring CFC rejection remains pinned in `cfc-additive-default-preserves-old-doc.test.ts`.

On the two questions:

1. The generation-skew concern is real. PR #4668 explicitly deferred the read-side grace, so this change guarantees candidate setup migration but does not manufacture a fallback when an older concurrently running generation later writes the old shape. I added that limitation to the live workflow docs and recommend retaining/adding a result default when rollback or mixed-generation tolerance is required.
2. Extracted collection entities were a reachable provenance gap: `data-updating.ts` creates a separate target that the containing output path cannot cover. Generated-output provenance now propagates through `diffAndUpdate` to those entity documents, with a regression test. Ordinary collection writes still carry no output role and remain default-gated; stream shape alone grants no exemption.

Verification:

- focused suites: 9 passed / 190 steps
- `deno task check`: passed
- `deno lint`: passed (4274 files)
- changed-file `deno fmt --check`: passed
- repo-wide `deno fmt --check`: still reports the pre-existing unrelated formatter debt already noted on the PR

bfollington (commented):
cf-review round 2 via Claude Opus 5, on behalf of @bfollington — incremental review of `7be83a7ec`

**Scope** · 13 files, +177 −68 on top of the reviewed `480cd1704` · **Coverage**: deep-read of all runtime source in the delta, all six `sendValueToBinding` call sites, new tests · **Checks run**: `deno test -A packages/runner/test/{cfc-*,data-updating*,pattern-*}.test.ts` → **231 passed / 1529 steps / 0 failed**; `deno fmt --check` + `deno lint` on changed files clean.

All four Improvements are resolved. I verified each rather than taking the summary at its word:

| # | Fix | Verified |
|---|---|---|
| 1 | Home rationale now says the gate *doesn't* catch default removal, so the pins are the protection | matches what I measured |
| 2 | JSDoc now states `[]` exempts the whole document; `types.ts` reframed as "present only when… absence is deliberately strict" | accurate |
| 3 | Argument-side record deleted, `"input"` removed from the union, result role threaded through `setRawUntyped` | one record now, not two |
| 4 | Vintage gate + plan doc reworded to claim only what they prove | honest |

Putting the mixed-generation caveat into `docs/common/workflows/development.md` rather than leaving it in a PR comment is the right home for it — that's the normative surface someone hits when deciding whether to add a `Default<>`.

## Review of the new mechanism

The entity-provenance propagation (`DiffAndUpdateOptions.schemaRole` → `normalizeAndDiff` → `pattern-binding.ts`) is new code, so I reviewed it fresh. **The role assignment is correct**: I read all six `sendValueToBinding` call sites in `runner.ts` (5032, 5115, 5479, 6062, 6252, 6414) — JS action results, the error output, raw-builtin results, output bindings, and the child-piece result. Every one writes a node's output into its output binding; none is an input-side or handler write. `updateArgument`'s `diffAndUpdate` correctly passes no options, so the argument side stays default-gated.

### 🟡 `packages/runner/test/cfc-boundary.test.ts` — the two assertions collapse into one, and neither tests the outcome *(verified)*

The first assertion looks like it pins "the root document carries the output marker" and the second like "an extracted entity does too". But the extracted entity's `newEntryLink` has `path: []`, so both are satisfied by the same record. Instrumented, the traversal records exactly one schema input:

```
ROLE_INPUTS: [{"role":"output","isRoot":false,"path":[]}]
```

The root is never marked in this path at all — its role arrives separately via `setRawUntyped`. So the first assertion is dead weight that reads as coverage.

Bigger point: the test pins *plumbing*, not *outcome*. It shows the marker lands on a second document; it never shows the marker changes the merge decision. Given this PR's own plan doc holds cases to "every case discriminates — each goes red under a mutation of the thing it claims to test", the discriminating version is an additive required field on the entity document: accepted with the role, still rejected without it. That is the behavior the propagation was added for.

### ⚪ `packages/runner/src/data-updating.ts` — the comment is creation-framed, the branch is not

> "When a child value becomes its own entity document…"

The `[ID]` branch runs on *every* write of an `[ID]` object, not only the first: `createRef` is deterministic, so it re-targets the same long-lived entity document, and the record is unconditional. So the exemption also covers entity docs that already hold data — which is presumably what you want (that is exactly the upgrade case), but the comment describes a narrower trigger than the code has. Same comment/behavior mismatch class as finding 2 last round, one level down.

## Verdict

Clean. Both new findings are cosmetic-to-test-rigor; neither blocks. The relaxation is now correctly scoped, honestly documented, and the rollback limitation is written down where someone will actually find it. The remaining gate is a human one: this reverses the direction @seefeldb signed off on in #4901 ("post-genesis DATA fields ride `Default<>`"), so that call is theirs, not a review finding.

bfollington:
Addressed the round-two test-rigor feedback in cdb6af1b3.

- Replaced the duplicate marker assertions with a behavioral migration test: adding the same required field to an existing extracted entity rejects without output provenance and commits with propagated output provenance.
- Reworded the ID-entity branch comment to cover both creation and later writes.

Verification:
- deno test -A packages/runner/test/cfc-boundary.test.ts — 2 suites, 94 steps, 0 failures
- deno lint — 4274 files checked
- changed-file deno fmt --check and git diff --check pass
- repo-wide deno fmt --check still reports the same 23 unrelated pre-existing files

seefeldb (approved):

bfollington:
Rebased onto origin/main at f25518c1f and resolved the overlapping vintage-harness and array-anchoring changes.

The output-provenance propagation now follows main's anchored-entity path, and the behavioral regression test uses the new anchor-id API.

Verification with the pinned Deno 2.9.4:
- deno task check
- deno lint — 4277 files
- relevant compatibility, CFC, piece-update, state-continuity, and vintage tests — 11 suites, 301 steps, 0 failures
- changed-file deno fmt --check — 19 files

Repo-wide formatting still reports the same 23 unrelated pre-existing files.

PR #5240, opened 2026-07-31
feat(schema-generator): C3 — emit a verb's declared result into the durable schema

mpsalisbury:
## Why

Since C1 (#5170's arc) a verb can declare its result in the type system —
`Stream<E, R>` — and since C2 (#5234) the transformer errors when a body
returns a value no declaration promised. But the declaration still died at
the type layer: a returning verb and a value-less one emitted byte-identical
durable schemas (pinned deliberately by C1's `stream-result.test.ts`), so
nothing downstream — `cf piece verbs`, verb discovery, the D5 gate — could
ever learn what a verb returns. C3 moves that pin: the schema generator now
emits the declared result into the piece's durable schema, per the plan's
WS-C schema-generator bullet
(`docs/plans/pattern-verb-contract-implementation.md`).

**Headline review question — the keyword shape.** The result attaches as a
new dialect keyword, `result`, on the stream property's schema object,
sibling to `asCell`:

```json
{ "$ref": "#/$defs/AddTopicEvent", "asCell": ["stream"],
  "result": { "$ref": "#/$defs/AddTopicResult" } }
```

Rationale: the event schema stays exactly where every consumer already reads
it (`cf piece call` payload validation, `cf piece verbs`, the D5 gate), and
the extension mechanism is the one `asCell` already uses. Please flag if the
dialect wants a different spelling or placement — everything else here is
mechanical relative to that decision.

Design points, implemented exactly as the plan's WS-C bullet decided:

- The result rides the **same** type check that emits the marker (both
  Stream branches in `common-fabric-formatter.ts` and the
  stream-returning-callable path in `object-formatter.ts`, shared logic in
  `src/stream-result-schema.ts`). A property either gets both keys or is
  skipped entirely; there is no state where the marker lands and the result
  schema does not.
- Non-void `R` goes through the normal type→schema machinery and is **left
  open** — no `additionalProperties` beyond what the type itself demands.
- A value-less verb (`Stream<E>` / `Stream<E, void>`) emits
  `"result": { "type": "object", "properties": {} }` — the `{}` receipt the
  runtime actually writes — NOT the void sentinel (`{ asCell: ["opaque"] }`,
  a wrapper claim that would hand readback a cell to resolve).
  `additionalProperties` stays undefined so the compat checker (`?? true`)
  never freezes a verb as value-less.
- Handler-typed properties (`HandlerFactory<E, T, R>` returns) follow the
  same rule wherever the existing stream detection recognizes them.

**The compat checker needed extending.** Probed first
(`verb-result-compatibility.test.ts` against the unmodified checker): it
byte-compared the unknown keyword via `unknownKeywordIssue`, which rejected
BOTH directions — removing a result field (right verdict, wrong mechanism:
no recursion, so compatible optional-field evolution was also refused) and
**adding the keyword where the previous schema had none**, which is every
existing deployment's upgrade path. `verbResultSubsetIssue`
(`packages/piece/src/schema-compatibility.ts`) now recurses into the keyword
with result-role permanence, candidate → previous regardless of which schema
carries the verb (a declared result is produced by the verb and read by
callers; its variance does not flip with the property's role). A side
without the keyword is unconstrained: adding it passes, dropping it once
published is refused, and a removal at any depth is rejected by the same
recursion every other result name gets.

## What Changed

- `packages/api`: `result?: JSONSchema` joins the Common Fabric extensions
  on `JSONSchemaObj`, documented beside `asCell`.
- `packages/schema-generator`: emission at the three stream-recognition
  sites; shared `src/stream-result-schema.ts`; `stream-result.test.ts`
  rewritten to pin the new shapes (deliberately moving C1's byte-identical
  pin); 9 schema fixtures regenerated; two inline expectations updated
  (`void-type`, `factory-input-real-api`).
- `packages/piece`: `verbResultSubsetIssue` recursion; `result` handled in
  `unknownKeywordIssue` and `collectSchemaReferences` (a `$ref` inside a
  declared result must resolve equally on both sides); the semantic
  extensions (`asCell` et al.) and `result` joined
  `DEFAULT_STABLE_SCHEMA_KEYS` — they are equality-checked metadata, not
  instance constraints breakable by default insertion, and without this
  every verb property switched off default-backed evolution for its whole
  subtree, making the permanence table's "required field with a default"
  allowance unreachable below any verb. New
  `test/verb-result-compatibility.test.ts`.
- `packages/runner`: `schema-walk.ts` gains a non-value keyword tier
  (`NON_VALUE_SINGLE_SUBSCHEMA_KEYS`) so value-semantics walks skip `result`
  while complete walks (chiefly `$ref` discovery) include it — without this,
  `selectReferencedCfcSchemaDefs` would prune a `$defs` entry referenced
  only from a declared result and strand the ref; `validateSchemaDefinition`
  preflights the subtree; pins in `schema-walk.test.ts` and `cfc.test.ts`.
- `packages/ts-transformers`: 17 goldens regenerated — every stream schema
  gains its result keyword; `stream-declared-result.expected.jsx` shows the
  full shape (event `$ref` + marker + result `$ref` + hoisted
  `AddTopicResult` def).
- Docs riding: mapping spec §6.5 (+ §4 callable row, §6.2 pointer);
  `docs/specs/json_schema.md` defines the keyword beside `asCell`; the
  plan's WS-C schema-generator bullet struck as **done (C3)** recording the
  keyword decision and the compat-checker outcome — the closed-world-inputs
  sentence stays unstruck (C5, still open).

## Test plan

1. `packages/schema-generator`: 55 passed (293 steps), 0 failed.
   `stream-result.test.ts` pins: declared result under the keyword beside
   the event `$ref` and marker; value-less empty-object receipt (incl.
   explicit `Stream<E, void>`); returning vs value-less no longer
   byte-identical — C1's pin moved as planned; alias
   `type MyVerb = Stream<E, R>` survives.
2. Compat-checker recursion probe pinned in
   `packages/piece/test/verb-result-compatibility.test.ts` (8 steps):
   removal rejected at top level and at depth; adding the keyword compatible
   in both roles (result- and argument-side verbs); dropping it rejected;
   optional-field evolution accepted; defaulted-required accepted /
   bare-required refused over the value-less receipt; `$ref`s inside results
   resolved per side. Full piece suite: 32 passed (377 steps), 0 failed.
3. Fleet upgrade path: `deno task pattern-compat` — first run reported 89
   unrecorded post-C3 contracts (every authored pattern's contract hash
   moves with the keyword) and **zero** baseline-verification failures,
   **zero** [NEW] evaluation errors (the 9 skips are all pre-existing
   `[known]`). Recorded with `--update` (append-only —
   `check-baselines-append-only` passes), then re-ran green: exit 0, "336
   pattern(s) can be updated from every recorded contract" — every pattern,
   now emitting result keywords on its verbs, still verifies against its
   recorded pre-C3 baselines. That is verification 2's add-direction at
   fleet scale.
4. ts-transformers: full suite 1130 passed (753 steps), 0 failed after
   regenerating the 17 moved goldens with `UPDATE_GOLDENS=1`; every diff
   sanity-read — all moves are the designed keyword additions.
5. E2E readback `packages/runner/test/declared-result-e2e.test.ts`: 1
   passed (2 steps), both flag states.
6. Repo-wide `deno fmt --check` and `deno lint` clean; `deno task
   check-docs` — all 520 checked code blocks pass; runner suite 1104 passed
   (5766 steps), 0 failed; data-model 47 passed (2164 steps); cli 123
   passed (179 steps), 0 failed.

---
## Summary by cubic
Emit a verb’s declared result (`Stream<E, R>`) into the durable schema under a new `result` keyword beside `asCell: ["stream"]`, so downstream tools can see what a verb returns. The compat checker and schema walks were updated to keep upgrades safe and refs intact.

- **New Features**
  - Schema generator adds `result: <R schema>` on stream properties; handler-returning streams are covered.
  - Value-less verbs (`Stream<E>`/`void`) emit `result: { type: "object", properties: {} }`; declared results are left open beyond type constraints.
  - Compatibility: checker recurses into `result`; adding the keyword is compatible, removing it after publish is rejected.
  - Runner walks treat `result` as non-value (skipped for value semantics; included for `$ref` discovery) to avoid pruning defs.
  - Docs updated; fixtures and goldens regenerated across patterns and transformers.

- **Migration**
  - Pattern contract hashes change due to the new keyword; update recorded baselines (e.g., run your contract update task).
  - No runtime changes needed; existing deployments upgrade via the “add keyword” path.

<sup>Written for commit d143fce92e938fa99eea9d205f52504c01005dd2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5240?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mpsalisbury:
Withdrawn per Berni's review rather than merged: the value path never needed the keyword (results flow schema-free through receipts, proven end to end before this PR existed), and shipping `result` into durable schemas and append-only baselines would have hard-committed a shape the coming Fabric-types stream evolution (`type: "handler", event: ..., result?: ...`) is expected to replace — the compat rules built here would themselves have refused the keyword's later removal. What this PR built and proved is recorded in the plan's WS-C bullet along with the gap the deferral accepts (verb result shapes have no update-gate protection; prose descriptions are the interim). #5244 and #5246 have been rebuilt on the C2 base without this content. The branch is deleted; this PR's diff remains the reference implementation if the Fabric-types design wants to reuse the compat-checker recursion work.

PR #5245, opened 2026-07-31
feat(runner): flip the plainResultReceipts default to on

mpsalisbury:
## Why

Governing decision 2 flips this default "after the integration suite proves readback end to end." That proof is on main: the three-topic fixture (D4, #5244) drives a caller-supplied event id through a dropped-response retry and a same-id replay with a different payload, both reading the ORIGINAL declared result back off the receipt, cross-process.

**Approved and no longer held.** Berni signed off on 2026-08-03 — *"Writing the undeclared response into the receipt seems fine to me"* — asked against the measured fleet exposure below, and Mike gave the second half of the go the plan requires.

## What Changed

- Runtime normalizes `plainResultReceipts ??= true` beside `computedCellIds`. `EXPERIMENTAL_PLAIN_RESULT_RECEIPTS=false` (the env mapping accepts exactly `true`/`false`) or a programmatic `false` stays the rollback override while the flag exists.
- Tests that implicitly meant "flag off" now say so explicitly and are retitled; `declared-result-e2e` runs under the built-in default so it pins the flipped default itself.
- Registry (`EXPERIMENTAL_OPTIONS.md`), scheduler-v2 §7.6, the implementation plan and the design doc record the new default.

Rebuilt on current main. The earlier version of this branch sat on the pre-rebase C2/D4 lineage, where merging main produced add/add conflicts on C2's transformer files — its base had been *rewritten*, not advanced. **The diff is the flip alone, 9 files.** The receipt-write conversion that used to be bundled here shipped separately as #5262 and is on main; it was a defect under the flag as it already existed, so it had no business waiting on this decision.

## What flipping changes for callers

A verb's return value reaches its receipt by default. Per Berni's review the incidental-cell hazard was fixed at the root rather than guarded (#5262): a one-line setter verb — `action(() => cell.set(...))`, whose expression body implicitly returns the cell `set()` hands back for chaining — records a **link to the cell it mutated** where its receipt used to be empty.

**Measured scope: 40 expression-body setter verbs across 11 non-test pattern files** — `mobile-app-demo` (`openSheet`, `closeSheet`, `dismissToast`), `parking-coordinator` (cancel-edit handlers), `lot-watch` (navigation steps), `imported-calendar` / `weekly-calendar` (day/week view), plus `notes`, `router`, `agent`, and the catalog stories. None were written with results in mind; each now publishes a link to its internal state cell, readable by anyone who can read the receipt. That is what the approval covers.

Two things already decided, recorded so they are not re-litigated:

- **Undeclared receipt content is advisory.** C2's absorption rule governs DECLARATION, and with C3 withdrawn no result schema exists to contradict — so receipts may carry values, including these incidental links, that no contract declares. The distinction becomes enforceable when the Fabric-types stream evolution brings declared-result metadata; that is the deferred-C3 gap.
- **Retention is unresolved but unchanged in shape.** The receipt cell is created and create-only marked whether the flag is on or off, so the flip does not change the *number* of documents — one per handled event, as today — only their *content*. WS-E's collection story applies either way; the flip raises per-document size, not document count.

Results carrying pieces do **not** need this flip: a return containing cells travels the result-pattern projection path, which predates the flag (measured in #5263, where `addTopic` returns the created topic on any runtime while plain records need the flag). The flip's scope is precisely plain-JSON returns.

## Test plan

- **Full workspace suite green under the flipped default** — every package. This is the run that matters: the whole repo executing with results reaching receipts by default.
- `packages/runner` receipt suites green, both flag states pinned.
- The verb walkthrough (`packages/cli/integration/verbs-over-the-cli.sh`) is **23/23** under the flip. Its step 7 sets `EXPERIMENTAL_PLAIN_RESULT_RECEIPTS=false` explicitly, so it now exercises the rollback override against a default-on runtime.
- `deno fmt --check` clean.

### DIFF
--- a/docs/development/EXPERIMENTAL_OPTIONS.md
+++ b/docs/development/EXPERIMENTAL_OPTIONS.md
@@ -30,7 +30,7 @@ was last checked against the code.
 | [`modernCellRep`](#moderncellrep) | `EXPERIMENTAL_MODERN_CELL_REP` env, or `RuntimeOptions.experimental` | off | Dan Bornstein (#3818) | graduate to always-on, then delete flag | implemented, off by default |
 | [`persistentSchedulerState`](#persistentschedulerstate) | `EXPERIMENTAL_PERSISTENT_SCHEDULER_STATE` env, or `RuntimeOptions.experimental` | off | Bernhard Seefeld (#3646) | graduate to always-on | implemented, off by default, rollout in progress |
 | [`commitPreconditions`](#commitpreconditions) | `RuntimeOptions.experimental` only (mapped `null` — programmatic rollback override — in the canonical env registry) | on | Bernhard Seefeld (#4090) | fold into base scheduler semantics, then delete flag | implemented, on by default |
-| [`plainResultReceipts`](#plainresultreceipts) | `EXPERIMENTAL_PLAIN_RESULT_RECEIPTS` env, or `RuntimeOptions.experimental` | off | Mike Salisbury (verb contract WS-C) | flip default after the invocation-protocol integration proof, then fold into receipt semantics and delete flag | implemented, off by default |
+| [`plainResultReceipts`](#plainresultreceipts) | `EXPERIMENTAL_PLAIN_RESULT_RECEIPTS` env, or `RuntimeOptions.experimental` | on | Mike Salisbury (verb contract WS-C) | fold into receipt semantics and delete flag after a bake period | implemented, on by default |
 | [`eagerSourceAnnotation`](#eagersourceannotation) | `EXPERIMENTAL_EAGER_SOURCE_ANNOTATION` env, or `RuntimeOptions.experimental` | off in production, on in shell dev builds | gideon (#4458) | permanent debug toggle, not slated for removal | implemented |
 | [`systemPatternAutoUpdate`](#systempatternautoupdate) | `EXPERIMENTAL_SYSTEM_PATTERN_AUTOUPDATE` env / shell build define, or `RuntimeOptions.experimental` | on in the shell (same-toolshed system sources, including all roots); off server-side | Bernhard Seefeld (#4611; shell default-on #4619) | graduate to always-on, then delete flag | implemented, on in the shell |
 | [`computedCellIds`](#computedcellids) | `EXPERIMENTAL_COMPUTED_CELL_IDS` env, or `RuntimeOptions.experimental` | on | Robin McCollum (#4659) | graduate to unconditional behavior, then delete flag | implemented, on by default |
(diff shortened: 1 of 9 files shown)


mpsalisbury:
@seefeldb — one question, and it's a re-confirmation rather than a new decision.

## The question

You have twice held that receipts are **runtime-honest**: they record what a
handler returned, and the fix for the incidental-cell hazard is the standard
write conversion, not a guard that discards undeclared returns. That fix landed
on its own as #5262 and is on main, so the mechanism is already live behind the
flag.

What is left is the default. That ruling now has a measured scope, and we would
rather you confirm it against the real number than have us infer your consent
from the principle:

> **40 expression-body setter verbs across 11 non-test pattern files** will
> begin publishing a link to the internal state cell they mutate, in a receipt
> readable by anyone who can read the receipt. Does the ruling hold at that
> scope?

If yes, nothing else is outstanding and the flip is Mike's to merge whenever he
wants it.

## What those verbs are

`set()` returns its cell for chaining, so an expression-body verb implicitly
returns the mutated cell. None of these were written with results in mind — they
are UI affordances:

| file | verbs |
| --- | --- |
| `mobile-app-demo.tsx` | `openSheet`, `closeSheet`, `dismissToast` |
| `factory-outputs/parking-coordinator/main.tsx` | `cancelEditPerson`, `cancelEditSpot`, `cancelRemoveSpot` |
| `factory-outputs/lot-watch/main.tsx` | `goBackToPhoto`, `goBackToSpot` |
| `google/core/imported-calendar.tsx`, `weekly-calendar/weekly-calendar.tsx` | `setDayView`, `setWeekView` |
| `notes/notebook.tsx`, `notes/note.tsx`, `router/router.tsx`, `agent/agent.tsx`, `catalog/stories/*` | same shape |

After the flip, `openSheet`'s receipt carries a link to `sheetOpen` where it was
previously empty, and `cf piece call openSheet` returns that link as its result.

## What is NOT being asked

Both of these are already decided and are listed only so they are not
re-litigated:

- **Undeclared receipt content is advisory.** C2's absorption rule governs
  DECLARATION; with C3 withdrawn no result schema exists to contradict, so
  receipts may carry values — including these incidental links — that no
  contract declares. Enforceable when the Fabric-types stream evolution brings
  declared-result metadata; that is the deferred-C3 gap, recorded in the plan.
- **Retention is unresolved but unchanged in shape.** The receipt cell is
  created and create-only marked whether the flag is on or off, so the flip does
  not change the *number* of documents — one per handled event, as today — only
  their *content*. WS-E's collection story applies either way; the flip raises
  per-document size, not document count.

## If the answer is no

The alternative is the one you already rejected: a guard limiting receipts to
DECLARED results, discarding incidental returns. It was built during the flip's
staging and removed on your review in favour of the root-cause fix. Raising it
again is a reversal, not a refinement — which is exactly why we would rather ask
now than merge and discover the scope was the objection.

## Rollback

While the flag exists, an explicit `false` — `EXPERIMENTAL_PLAIN_RESULT_RECEIPTS=false`
or programmatic — restores the discard-everything-to-`{}` behavior. Both flag
states stay pinned in `scheduler-event-receipts.test.ts` and
`declared-result-e2e.test.ts`.

## State of the PR

#5245 is rebuilt on main (the previous branch sat on the pre-rebase C2/D4
lineage), the diff is the flip alone across 9 files, and the full workspace suite
is green under the flipped default — every package. It stays a held draft until
you and Mike say go.

seefeldb:
Writing the undeclared response into the receipt seems fine to me.

PR #5251, opened 2026-07-31
fix(runner): read an optional property written as `undefined` as absent

seefeldb:
## The bug

A handler that builds an object literal from a variable that happens to be undefined writes the KEY, and the codec stores that presence faithfully. Validating it as a value then asks whether `undefined` satisfies the property's declared type — which no ordinary type answers yes to.

So the same optional field **passed when omitted and failed when written as `undefined`**, a distinction JavaScript cannot make on read.

## Why it matters

`packages/patterns/topics/topic.tsx` does exactly that:

```ts
const author = topicAuthorFromAgent(agentName ?? "");  // undefined when no agentName
comments.push({ author, authorName: legacyName, body: trimmed, sentAt: Date.now() });
```

The stored comment carries `author` as `{"/Undefined@1": null}` — present, undefined. Its own pattern's next update was then **refused**:

```
updated arguments do not match the candidate schema: comments: 0: author: value does not match type object
```

That refusal is permanent. `isStoredArgumentSchemaRefusal` exists precisely because "a root pinned to a version whose schema cannot read its own document never opens again". The runtime was writing data its own updater would not accept, and **any** pattern doing `push({ optionalField, ... })` mints one.

## The fix

An **optional** property holding `undefined` is absent. Three sites, all asking "is this property present": the `properties` loop and both `additionalProperties` scans.

### What is deliberately NOT changed

- **`required`.** `type: "undefined"` is a type this validator supports, so a required property of that type holds `undefined` legitimately and must still be measured. My first attempt applied the rule to all eight presence checks; `runner.test.ts` → *"preserves explicit undefined while combining union defaults"* caught it, validating a defaults object of exactly that shape. A required property of any other type holding `undefined` keeps failing on its type — right answer by a different route.
- **`min/maxProperties`, `dependentRequired`, `dependentSchemas`, `propertyNames`.** Counting and dependency semantics where "present" is genuinely ambiguous for a required-undefined, and none were measured.
- **Array indices.** The five `Object.hasOwn(value, index)` checks distinguish a sparse hole from a present element; `[undefined]` genuinely differs from `[]` — it has a length.

## Verification

- Red→green confirmed **both directions**: source reverted with the new tests kept → both fail; fix restored → both pass.
- Full runner suite **1100 passed / 0 failed**.
- On the Tier 2 vintage gate against the committed topics fixture, `updated cleanly` went **32 → 33** and the refusal disappeared.

Two tests: the validator rule (including the `type: "undefined"` carve-out and the undeclared-key case) and an end-to-end roll-forward built on the real topics shape, with an explicit premise assertion that the codec actually stored the key — otherwise it would pass without exercising anything.

Found by the Tier 2 vintage gate on its first genuine exercise against topics.

---
## Summary by cubic
Adds an opt-in validator mode used only during stored-argument checks to treat optional properties written as `undefined` as absent. This avoids permanent update refusals (e.g., `{ author: undefined }` in comments) without changing default validation.

- **Bug Fixes**
  - Introduced `optionalUndefinedIsAbsent`; enabled only in stored-argument validation in `packages/runner/src/runner.ts`. Default validator stays strict (`undefined` is a value).
  - Applies only to declared optional properties (including array items); required keys are still measured. `type: "undefined"` when required remains valid; required non-undefined types with `undefined` still fail.
  - Keeps strict `additionalProperties`; undeclared keys set to `undefined` still error. Drops the option inside `allOf` branches to avoid skipping requiredness.
  - Added unit and end-to-end tests for default vs opt-in behavior, array cases, `allOf` handling, and a topics/lunch-poll stored-argument acceptance case.

<sup>Written for commit c7e02bc0f55add806e239e7b541c6c061cd77c42. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5251?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/cfc/schema-sanitization.ts
+++ b/packages/runner/src/cfc/schema-sanitization.ts
@@ -510,47 +510,36 @@ const typeMatches = (
 };
 
 /**
- * Whether `key` is an OPTIONAL property holding `undefined` — which says
- * nothing, and so is read as absent rather than as a value to measure.
+ * Whether `key` is an OPTIONAL property holding `undefined` that THIS caller
+ * has asked to read as absent rather than measure.
  *
- * JavaScript cannot tell `{ a: undefined }` from `{}` on read, and a handler
- * building an object literal out of a variable that happens to be undefined
- * writes the key without meaning to say anything by it —
- * `comments.push({ author, ... })` with no author in hand is the shape that
- * made this matter. Measuring such a key asks whether `undefined` satisfies
- * the property's declared type, which no ordinary type answers yes to, so one
- * optional field would pass when omitted and fail when written as undefined.
+ * Gated on the option, and off by default, because `undefined` is a value in
+ * this system rather than a hole: the codec stores its presence as
+ * `{"/Undefined@1": null}`, `type: "undefined"` is a type this validator
+ * supports, and `pull-materialization.test.ts` pins both halves ("does not hide
+ * present explicit undefined behind an optional alias", "retains explicit
+ * undefined at an optional derived Cell root"). A caller writing `undefined`
+ * where a number is declared has made a mistake worth rejecting while they can
+ * still see it.
  *
- * The cost of getting this wrong is not a spurious message. A stored argument
- * that fails validation refuses its pattern's every future update, and refuses
- * it identically each time — see `isStoredArgumentSchemaRefusal` in
- * `../runner.ts`, which exists because a root pinned to a version whose schema
- * cannot read its own document never opens again.
+ * The stored-argument check a pattern update runs asks a different question,
+ * which is why it opts in — see `optionalUndefinedIsAbsent`.
  *
- * REQUIRED properties are excluded, and that exclusion is load-bearing rather
- * than cautious: `type: "undefined"` is a type this validator supports, so a
- * required property of that type holds undefined legitimately and must still
- * be measured to be accepted. A required property of any OTHER type holding
- * undefined keeps failing on its type, which is the right answer by a
- * different route.
+ * REQUIRED properties are never absent under this rule, and that carve-out is
+ * load-bearing rather than cautious: a required property declared
+ * `type: "undefined"` holds undefined legitimately and must still be measured
+ * to be ACCEPTED. A required property of any other type holding undefined keeps
+ * failing on its type, which is the right answer by a different route.
  */
 const isAbsentOptional = (
   value: FabricPlainObject,
   key: string,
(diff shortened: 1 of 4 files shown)


seefeldb:
Both review findings were real. I reproduced each before changing anything, and the fix went further than either suggested.

**The premise was wrong, not just the edge cases.** CI caught it independently: `pull-materialization.test.ts` has two tests — *"does not hide present explicit undefined behind an optional alias"* and *"retains explicit undefined at an optional derived Cell root"* — that exist to keep explicit `undefined` visible. `undefined` is a value here, not a hole: the codec stores its presence as `{"/Undefined@1": null}` and `type: "undefined"` is a supported type. Making leniency the validator's rule was wrong.

It is now an option, `optionalUndefinedIsAbsent`, **off by default**, enabled at exactly one caller — the stored-argument check on pattern update. That site asks a different question ("can this version read the document already there?") and answers it with a refusal that is *permanent*. The result-write contract is untouched.

**cubic P1 — required-but-undeclared key bypasses closed schemas.** Reproduced:

```
{properties:{}, required:["x"], additionalProperties:false} + {x: undefined}
  before: "additional property x"    after (lenient): accepted
```

Dropped that leniency **entirely** rather than patching it. Nothing needed it: `topic.tsx`'s `author` and `lunch-poll`'s `imageUrl` are both *declared* properties, so the additional-property scans never mattered to the bug.

**Codex P2 / cubic P2 — `required` and `properties` split across `allOf`.** Reproduced with your exact example:

```
{allOf:[{required:["x"]},{properties:{x:{type:"string"}}}]} + {x: undefined}
  before: "x: value does not match type string"    after (lenient): accepted
```

The option is now dropped for `allOf` branches. Plumbing requiredness conjunctively down the recursion would be the general fix, but nothing needs it: `packages/schema-generator` emits no `allOf` at all, and zero committed pattern baselines contain one, so a pattern argument schema — the only place the option is enabled — cannot reach that branch. Strict is the safe direction for a shape that does not occur, and if that ever changes the new test fails rather than the behavior silently widening.

Both are pinned by tests, and the main test now asserts **both** directions — default measures and fails, opt-in reads as absent — instead of only the behavior I wanted.

PR #5253, opened 2026-07-31
feat(patterns): key vintages by the TEST that made them, and add topics + lunch-poll

seefeldb:
Coverage was two patterns. It is now **four fixtures, 84 recorded instantiations and 73 upgrade targets** — because a fixture is keyed by the TEST that produced it rather than by a pattern whose name the test had to match.

The old shape derived `<pattern>.test.tsx` from a pattern key. Measured, that convention holds for 91 of 120 tests in `packages/patterns` and not for the interesting ones: `topics/main.tsx` is tested by `topics/topics.test.tsx`, and `lunch-poll/main.test.tsx` drives several patterns and is named after none of them. Both were **uncapturable**, not merely un-required.

Coverage is now judged from what the replay actually REPLAYED, not from the fixture tree's shape. `VintageRef.patternKey` is renamed `testKey` for the same reason — a field named for patterns holding a test path is a bug waiting to be written.

Three things only visible once coverage left the two system roots: a pattern loaded BY URL records the toolshed route rather than a repo path; that same pattern compiles to a different identity served than from the repo, so it read as changed forever; and a keyless instantiation is not unaddressable but simply **not a target**, which the capture and replay guards disagreed about.

## The gate was checking no topics state at all

Widening coverage made two comparison bugs visible, both the same mistake: **reading state under a schema meant for validating it.**

A schema-driven read answers `undefined` for the WHOLE object when a required property does not resolve — and a pattern marks its session-local drafts required. `topic.tsx` requires `bodyDraft`, `commentDraft`, `editingBody`, `linkUrlDraft`, `linkLabelDraft`, each a link to a per-session cell a fresh replay runtime holds nothing for. **One unresolvable draft hid all 28 keys of real state**, so every recorded `topic.tsx` target was reported as a root the fixture does not hold. The gate failed loudly while examining nothing.

- `schemaWithUnknownsRelaxed` → **`schemaRelaxedForComparison`**, now dropping `required` at every depth. This cannot hide a loss: the comparison is per key, so a key that resolved before and not after is still a finding. `required` only decided whether the object collapsed *entirely* — and a collapse hides the moved key the gate is hunting.
- The presence check had the same collapse (it asked `outcome.value`, the candidate's schema verbatim) and so reported state as GONE that the adjacent comparison could read. Only reachable once the first was fixed.
- The message said "no readable stored schema" when the schema was present and readable — split via `readStoredResultSchema`.

## A changed value warns; a value that went empty fails

A replay recomputes as well as reads. Measured: `createdBy` `{name:""}` → `{name:"t"}` (the pattern backfilling from its own `createdByName` shadow), `$NAME` `"Topics (2)"` → `"Topics (3)"`, `artSyncState` `""` → `"generated"`. None is data going missing.

So a changed value warns; a **non-empty value that went empty** fails. Empty includes the falsy primitives, since those are what a `Default<>` gives an unwritten field — "reverted to its default" and "disappeared" are one event.

**The limit that buys is pinned, not hidden.** A moved `.for()` key whose new slot the pattern SEEDS reads back non-empty and only warns. Both halves are tested: `notes (was ["noted"], now [])` still FAILS, and the cross-space `note (was "written", now "captured")` case is renamed to document the gap and asserts the warning is still emitted, so a change that stopped reading the child's root can't masquerade as the limit.

## Fixtures are 60% smaller

`commit.original` is a byte-for-byte duplicate of the whole `revision` table — verified by joining `json_each(original,'$.operations')` to `revision` on `commit_seq`+`op_index`: equal counts, ids, ops and payloads in all four. The `commit` table is 42–50% of each file. Its only read site is `sameStoredOriginal`, a string equality keyed `session_id`+`local_seq` that a fresh replay session can never match. Blanking it keeps every commit row, so seq allocation, the `revision.commit_seq` FK and the conflict scans are untouched. `page_size` also drops 32 KiB → 4 KiB.

**41.2 MiB → 16.7 MiB checked out, ~4.7 → ~1.7 MiB packed.** Proven equivalent rather than assumed: pruned and untouched trees were replayed with every pattern's identity forced to change, so 50 targets actually materialized, and the two produced byte-identical output including all 18 failure diagnostics.

Fixtures stay RAW. Measured: git's zlib beats pre-gzipping (3.11 vs 3.25 MiB packed) and deltas a recapture ~5× better (+0.21 vs +1.03 MiB on a second topics generation), because it cannot delta a gzip stream.

## Verification

- unprobed gate green — 2 changed, 2 updated cleanly
- gate suites: 3 passed / 37 steps, and 9 passed / 62 steps
- `packages/piece`: 31 passed / 369 steps

## Note on ordering

Running this gate with every pattern's identity forced to change still surfaces refusals of the form `option: imageUrl: value does not match type string` and `comments: 0: author: value does not match type object`. Those are a **runtime** bug fixed in #5251, not something this PR introduces — CI here is green either way, since an unprobed run materializes only what actually changed.

---
## Summary by cubic
Key vintage fixtures by the TEST that produced them, expanding coverage to multi-pattern tests (topics, lunch-poll) and judging coverage by what actually replayed. Synced with main: kept both route/repo mappers, tightened coverage to replayed PINNED targets, and verified served routes are compared every run.

- **New Features**
  - Capture/replay keyed by test files; CLI accepts test keys; CI seeds `system/home.test.tsx`, `system/default-app.test.tsx`, `topics/topics.test.tsx`, `lunch-poll/main.test.tsx`.
  - Coverage credits PINNED replays only; served-route targets are materialized and state-compared; identity checks are skipped and counts printed; excluded from changed/updated.
  - Exact `/api/patterns/` mapping via `patternKeyFromMain`; unmappable roots (including under `packages/home-schemas/`) and keyless instantiations are excluded; patterns that no longer compile are reported per entry instead of aborting the run.
  - `REQUIRED_TEST_KEYS` lives in `pattern-vintage-lib.ts` with invariant tests; `reportNothingToSeed` explains when a named key is not a test.

- **Bug Fixes**
  - Correct state reads for comparison: `schemaRelaxedForComparison` drops `unknown` and `required` at every depth; presence check re-reads via the same helper; stored schema captured pre-materialization via `readStoredResultSchema`.
  - Grade loss per leaf (deep), so partial loss fails; object↔array and object↔scalar type changes now fail; changed-but-non-empty values warn. `--update` filters to PINNED, dedupes keys, reports by tests; “nothing to seed” moved to `reportNothingToSeed` and tested.
  - Fixture compaction keeps `commit.original.reads` and drops only `.operations`; page size lowered to 4 KiB. `cf inspect` works; size ~42.2 → ~21.8 MiB checked out.
  - Fixture integrity checks: fail a manifest with no instantiations, and detect filename/manifest identity mismatches with clear messages.
  - Adopted `patternRoute` (system route resolver) alongside `patternKeyFromMain` (manifest→repo mapper); removed the `coveredPatternKeys` helper — coverage is based on what replayed, not directory shape. Verified runs compare 9 served-route targets and remain green.

<sup>Written for commit 1280cfd3735ab6bfeba373a45ffa4f0a1732b822. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5253?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seefeldb:
Working through the open review threads — three were fixed in commits after the reviewed one, and one is a deliberate decision.

**"Replay URL-loaded targets before declaring the vintage clean" (P1)** — fixed in `e49632d26`. Served-route targets are now materialized and state-compared; only the *identity* comparison is skipped, because the same file compiles to a different identity served than from the repo. They are materialized unconditionally (there being no "changed since capture" answer for one) and kept out of `changed`/`updated`, which would otherwise report the same fixed number forever. `servedRoute` is also printed now: the real run reports `9 target(s) were served routes and not identity-compared`.

Pinned by a test that BREAKS the source a route maps to — counting served routes would pass just as well if they were still skipped, so the assertion is on the failure the comparison produces. Confirmed red without the fix.

**"Keep the documented update argument compatible" (P2)** — fixed in `e49632d26`. The usage block now reads `--update topics/topics.test.tsx` and states the rule explicitly: a named key is a TEST path, and `system/x.tsx` names no test and captures nothing.

**Auto fixture satisfying required coverage (P2)** — fixed in `adb4e5f78`: `report.covered` is filled as each target replays and only for a PINNED fixture. It is now pinned by "credits coverage only for a PINNED fixture", which replays the same fixture bytes under `pinned/` and under `auto/` and asserts it is replayed either way while crediting coverage only as pinned — confirmed to fail under mutation, because the guard previously survived deletion with every suite green. `captureMissing` got the same filter, which closes a dead end the reviewer did not raise: an auto fixture suppressed the pinned capture the gate was asking for.

**"Treat changed persisted values as possible data loss" (P1)** — this one is a deliberate decision by the repo owner, not an oversight, and I want to be clear about the trade rather than quietly disagree.

The grading is: a non-empty value that goes empty **fails**; a value that merely changes **warns**. It was chosen with this exact limitation stated up front — a moved `.for()` key whose new slot the pattern seeds reads back non-empty and only warns. The reason is that a replay recomputes as well as reads, so a derived value the vintage never pulled on resolves to something better this time. Measured on the committed fixtures: `createdBy {name:""}` → `{name:"t"}` (the pattern backfilling from its own compatibility shadow), `$NAME "Topics (2)"` → `"Topics (3)"`, `artSyncState ""` → `"generated"`. Failing on those reds the gate on every real edit, which is how a gate gets turned off.

What is NOT given up: emptiness is asked per **leaf at any depth**, so partial loss fails — `["a","b","c"] → ["a"]`, a field vanishing while its sibling survives, a row body emptied, and a container that stopped being one. And the companion case where a moved key reads back `[]` still fails. The limit is specifically a moved key whose new slot is seeded, pinned by name in `tasks/pattern-vintage-run.test.ts`.

Your suggested refinement — distinguishing durable `of:` state from recomputed values — is the right way to close it, and it is recorded in the code as the candidate: the reduction carries the id that would answer it, though a plain data key carries none, so it would grade only some findings. Deferred rather than dismissed.

PR #5254, opened 2026-07-31
feat(memory,runner)!: park verdict application on catch-up markers; keep the fan-out batched (CT-1927)

ubik2:
## Why

A transact verdict systematically outruns the batched fan-out carrying the foreign writes the commit was judged against, so acting on a verdict — promotion, basis emission — was never valid at its arrival. The client-side compensations for that one ordering deviation are most of CT-1872's inventory: the read-repair gate, its 30s timeout (Class 2), the marker race (Class 5).

The first design (reviewed through four rounds here) enforced ordering server-side: flush the space's relevant sync before sending any verdict. It was correct but forfeited the point of the batched fan-out: receives are serialized per connection, so each commit waited for a whole-space flush before the next commit even applied — the N-commits-one-recompute amortization (measured ~4x under write bursts) structurally cannot happen, and `main` now pins exactly that property (#5292's burst tests). This PR now ships the inverted design: **verdicts return inline and the CLIENT parks their state application until the catch-up marker covers them.**

## What

**Server — per-verdict catch-up markers, fan-out stays batched.** Verdicts return inline. EVERY verdict — accepts as well as conflict rejections — stages a `caughtUpLocalSeq` obligation, and the batched fan-out stamps the marker on the next frame to the committing session — an otherwise-empty frame if nothing it watches is dirty (pinned: the client's promotion depends on that guarantee). The new build-inherent `verdictCatchUpMarkers` protocol flag advertises the contract (precedent: `pendingReadStacks`). The `flushBeforeVerdict` option, the pre-verdict flush, and the drain-suspension machinery (`withReceiveUncounted`) are gone. The batch-pipeline robustness fixes from this PR's review rounds all stay: dirty-marking before the side-effect await (transact and `writeDocument`), mixed-provenance echo-suppression clearing, failed-fan-out requeue with cross-batch provenance, per-turn space removal.

**The stamping contract** (unchanged in meaning, now the load-bearing half): `caughtUpLocalSeq = W` means every verdict ≤ W is decided and this frame reflects those outcomes for the docs it covers. Foreign novelty and REJECTED-outcome repair ride frames (rejected docs staged origin-less); the session's own ACCEPTED writes stay echo-suppressed — the parked verdict carries their truth (with CT-1926's post-apply document once #5225 lands). Suppression can only ever hide your own write from you: any foreign write to the same doc defeats it (mixed provenance clears the origin same-batch; a foreign origin or seq drift defeats it cross-batch) and the doc is delivered authoritatively.

**Client — parked accepts.** `settleAccept` holds an accepted commit's promotion — pending overlay to confirmed mirror, removing the pending local copy — until `noteCaughtUpLocalSeq` sees the marker, so promotion extrapolates over a base that reflects the foreign novelty the accept was applied on top of, instead of minting a confirmed state from a stale mirror. Parked application runs before the frame's differential compare, so notifications reflect the post-promotion view. Parking engages only when the server advertises markers AND a sync consumer is live (no watch view → no frame stream to order against → immediate application, the legacy semantics; same on consumer teardown). Visible state is unaffected — the pending overlay already shows the write. Rejections keep the read-repair gate. Frame-time retirement is removed.

## Tests

Server (`v2-verdict-catchup-test.ts`, renamed): inline-verdict ordering for accepts and rejections (verdict first, then the batched frame with novelty + marker); the empty-frame marker guarantee; the five robustness pins from the review rounds (failed-fan-out requeue, mixed provenance, durable-write visibility during side-effect await, multi-space stranding with an observable second space, requeue provenance at the real seq). Plus `main`'s #5292 burst tests now pass on this branch: same-client and cross-client commit bursts get all verdicts before any fan-out, one flush frame carries the batch.

Client (`memory-v2-stacked-commit.test.ts`): promote-at-marker (accept resolves, overlay stays pending, marker moves it to confirmed and removes the local copy); covered authoritative frame does not double-apply the parked non-idempotent patch; a marker applies parked accepts even when the frame covers other docs; the full rejection round trip (verdict, repair frame, `readyToRetry`, regeneration declaring the winner's basis); an undecided dependent above the marker survives the parked application, with stale marker redelivery inert; old-server (`verdictCatchUpMarkers` absent) immediate application. Scripted transports default to NOT advertising markers — hand-rolled scripts never emit them — and marker tests opt in explicitly.

## Verification

- `packages/memory`: 466/466 · `packages/runner`: 1126/1126 (latest `main` merged: hold removal #5279, burst tests #5292) · `deno task check` both packages · `check-docs` green · fmt/lint clean

## Notes for review

- **Docs**: §4.11.2 rewritten around the marker MUST + client-parking contract and the `verdictCatchUpMarkers` flag; 05-queries §5.4.5, 10-implementation-guidance, implementation-plan updated; the `flushBeforeVerdict` EXPERIMENTAL_OPTIONS entry deleted (no flag remains); the adoption-spec dedup-loss notes reverted (batched fan-out restored means adoption dedup is back — CT-1930 largely dissolves); TLA README / INV-6 notes reworded to the parked-promotion window.
- **Interaction with #5225 (CT-1926)**: composes cleanly — parked promotion is exact for watched docs (the marker guarantees the base); the post-apply document closes the unwatched-doc/divergence residue and makes patch-echo suppression categorically safe. Trivial rebase for whichever lands second.
- The CT-1872 consequences: Class 2 demotes toward a backstop (the gate is load-bearing for rejections, released by the staged marker on a 5ms timer), Class 5 stays dissolved (the stamping contract is unchanged), Class 3's policy question is narrowed (accepts and conflicts both marker-ordered; other rejection kinds remain immediate-drop).


ubik2 (commented):

Self-review of the CT-1927 changeset before requesting human review. Scope: the four commits on this PR · coverage: deep-read, adversarial, targeting reentrancy, delivery ordering, and untested arms · checks run: memory 460/460, runner 1107/1107, `deno task check` both packages, `check-docs` green, fmt/lint clean.

## 🔴 Blocking

None found.

## 🟡 Improvements — two found, both fixed in 4736d061b

1. **Flush-failure arm untested.** The catch guaranteeing a broken `flushSessions` never eats a verdict (load-bearing: the commit's fate is durable before the flush runs) had no coverage. Pinned: an instance-stubbed rejecting flush still delivers the accept; the real flush then restores and drains the parked timer.
2. **Undecided-dependent semantics argued but unpinned** (and the zero-retired `continue` arm uncovered — a coverage-ratchet risk). Pinned with a distinguishing field on the delivered base: the covered layer retires, the undecided dependent above the marker survives and replays over the delivered base (no double-apply of the retired layer, no lost survivor), and a stale marker redelivery retires nothing.

## Verified, not just believed

- **No drain-wait self-deadlock**: the pre-verdict flush takes `refreshLoop(initial)`, which skips `waitForConnectionQueuesToDrain` — critical, since `#pendingReceives` counts the very transact being processed; the timer path keeps its drain-wait untouched.
- **No flag bypass**: `respond()` (the one-shot surface) handles only `hello`; every transact rides the connection path where the flush lives.
- **No lost novelty on interleave**: `flushSessions`' finally reschedules when dirty state accrues mid-pass; redelivery to the flushed session is idempotent under the client's backward-seq guard.
- **Retirement never removes an overlay of unknown fate**: the server stamps `caughtUpLocalSeq` only from verdict-time-staged obligations, so the marker never covers an undecided localSeq — and an empty frame covers nothing, so the Class 5 bare-marker shape is inert against retirement by construction. This is what let the client half ship unconditionally (no flag) and sound against old servers.

## Design note worth reviewer attention

Dirty-origin echo suppression refined the stamping contract away from the ticket's original "the frame covers everything ≤ W": frames carry foreign novelty and rejected-outcome repair; the session's own accepted writes settle at the verdict, which carries their truth (CT-1926, #5225). Each channel carries what it owns. The scoped-retirement and dependent-survival tests pin exactly this split; the ticket comment is updated to match.

## Test rigor

Every test names its clause of the §4.11.2 stamping contract. The 60s-parked-timer construction proves ordering comes from the flush rather than timer luck. The round-trip test pins the regeneration loop end-to-end including the fresh commit's wire shape (winner's basis declared; dependency stack no longer names the retired layer). The client waits are bounded polls on observable outcomes where frame integration exposes no completion signal — the file's established idiom per `waiting-in-tests.md`.

## Remaining review items for humans

- The contention-THROUGHPUT measurement before the flag defaults on (registry entry names the gate; per-commit flush loses batching amortization; latency is not the concern — the meaningful milestone gets faster).
- `#frameRetired`'s bounded leak (a verdict that never arrives leaves a numeric entry) — documented in its comment; judged acceptable.
- Both this PR and #5225 touch `confirmPending` on different concerns; trivial rebase for whichever lands second.

**Verdict:** sound — no blockers; both improvements already applied.

ubik2 (commented):
_cf-review via OpenAI Codex (GPT-5), on behalf of @ubik2_

Scope-and-theory review of the memory server, runner client, tests, protocol docs, and blast radius. Focused official tests passed (memory 4/4, runner 76/76), as did type-check, format, lint, docs checks, and diff checks. Three review-only isolated reproductions confirmed the failure states described below.

## 🔴 Blocking

### Major — a real flush failure consumes the recovery work

`packages/memory/v2/server.ts:665-671`, `3300-3314`, `3406-3439`

`refreshLoop()` deletes dirty spaces, IDs, and origins before awaiting fan-out. If fan-out throws, nothing restores them, so the warning that “batched refresh recovers” is false and repair may never arrive. The new failure test mocks `flushSessions()` before it consumes state and then manually flushes, so it misses this path. An isolated in-loop failure reproduction confirmed that `server.idle()` had no remaining recovery work.

Please requeue the taken batch on failure and test a failure inside `syncSessionForConnection()` followed by automatic recovery.

### Major — same-document foreign novelty is echo-suppressed

`packages/memory/v2/server.ts:2205-2235`, `3075-3095`

When queued foreign novelty and the accepting session's write touch the same document, the accepted write overwrites the dirty origin. The flush updates the server session cache but suppresses the upsert as the session's own echo. Verified with a foreign set of B followed by the session's zero-read patch of B: the server produced an empty marker frame and a patch-only verdict. PR #5225 is still open; without its post-apply document the runner remains stale, and direct WatchView clients remain stale regardless because the server now believes their cache has advanced.

Please preserve mixed foreign/local dirty provenance or force an authoritative upsert for this case, with an end-to-end same-document regression.

### Major — durable writes can remain invisible while scheduler side effects await

`packages/memory/v2/server.ts:2163-2205`, `1400-1420`

A commit becomes durable, then awaits `runPostCommitSchedulerSideEffects()` before being marked dirty. During that await, another connection can commit and run its pre-verdict flush without including the earlier durable write. That lets a later verdict outrun earlier watched novelty, which is the ordering this PR is intended to eliminate.

Please register the owner-space dirty IDs immediately after durable apply and before the first await, and pin it with a two-connection gated-side-effect test.

### Major — the normative wire specification omits the new watermark

`docs/specs/memory-v2/04-protocol.md:200-205`, `346-361`

The documented `SessionOpenResult` and `SessionSync` shapes both omit `caughtUpLocalSeq`, although §4.11.2 now makes it central to the wire contract. Live docs in `05-queries.md`, `10-implementation-guidance.md`, and `implementation-plan.md` also retain conflict-only ordering language.

Please update the interfaces and touched live documentation, including the experimental-options review date.

## 🟡 Improvements

- `packages/runner/src/storage/v2.ts:3693-3713`: a later watermark can retire an unsent lower localSeq held by conflict-admission `hold` mode. An isolated reproduction confirmed the stale result. **Downgraded to non-blocking per the author:** hold mode never worked, enabling it is unsupported, and it is being removed in a later PR.
- Rejected write documents are not necessarily covered before the verdict: `sameSnapshot` filters unchanged writes even when staged origin-less. The server's cross-document test delivers only the winning read document, while the runner test synthesizes an upsert for the rejected write. Either provide explicit covered IDs/forced authoritative upserts or narrow the “rejected phantom disappears before verdict” claim.
- Replace the 500 ms `Promise.race` at `packages/runner/test/memory-v2-stacked-commit.test.ts:2305` with an event-driven await, per `waiting-in-tests.md`.

**Verdict: request changes.** The flush-recovery and two ordering gaps are correctness issues in the flag-enabled path; the normative protocol also needs to describe the watermark it now relies on.

ubik2:
_cf-review addendum via Claude (Fable 5), on behalf of @ubik2 — 90507a100 flips `flushBeforeVerdict` to DEFAULT ON, and corrects one claim from my review above._

**Review correction.** The review's "no drain-wait self-deadlock" verification was wrong in the transitive case: the pre-verdict flush chains behind an in-flight TIMER pass, and that pass's `waitForConnectionQueuesToDrain` counts the very receive that is awaiting the flush — mutual waiting until the drain deadline (`max(500ms, 2× last refresh)` latency cliff in real time; an outright deadlock under a fake clock). The default flip's test soak caught it immediately: two FakeTime choreography tests hung. Fixed with `withReceiveUncounted` — a receive blocked ON fan-out is not inbound pressure the fan-out should yield to. The dedicated ordering tests plus both full suites (memory 460/460, runner 1107/1107) are green with the fix.

**What the default flip changes:**
- Every embedded-real-server test now runs the new ordering by default — the soak that found the bug above.
- Suites updated by subject: ordering-neutral suites shift past the effects now legitimately preceding verdicts (a `nextResponse` helper; includes marker-only empty frames to watch-less sessions — honest per the stamping contract). Suites whose FIXTURE is controlled staleness or legacy verdict-first choreography (conflict matchers, read-repair/retry recovery, adoption-vs-rerun, timer-pipeline deferral) pin the opt-out path explicitly with a uniform comment — those client mechanisms stay load-bearing for old servers and the rollback hatch.
- `EmulatedStorageManager.emulate` gains a `serverOptions` passthrough for the one such suite built on it.
- Registry entry and §4.11.2 status now read default-on, `false` = rollback hatch; path-to-removal is deleting the opt-out once soaked.

**Follow-up worth filing** (not in this PR): default-on twins for the recovery-path suites where user-visible behavior differs under the new ordering (e.g. read-repair arriving pre-verdict rather than via the gate), so the recovery machinery is pinned in both worlds rather than only the legacy one.

ubik2:
_cf-review response via Claude (Fable 5), on behalf of @ubik2 — Codex re-review blockers and improvements addressed in d45322d10 (note: that review ran at 4736d061b, before the default-on flip in 90507a100; all findings applied to head and are fixed there)._

**🔴 Flush failure consumes the recovery work (also cubic's P2)** — confirmed and fixed at the root: `refreshLoop` now requeues the consumed batch on fan-out failure (ids/origins merged back UNDER newer provenance, space re-marked, refresh rescheduled, error rethrown so callers still see it), and the timer path catches instead of leaking an unhandled rejection. Your exact test critique is taken: the new regression fails INSIDE `syncSessionForConnection` — after consumption — and asserts the next pass delivers the parked novelty plus marker. "Batched refresh recovers" is now a kept promise.

**🔴 Same-doc foreign novelty echo-suppressed** — confirmed and fixed in `markSpaceDirty`: marking with an origin over a doc that is already dirty with DIFFERENT provenance (foreign origin or unattributed staging) now clears the origin — mixed docs fan out authoritatively to every session, the writer included. End-to-end regression: parked foreign write to doc:b, then the session's own zero-read patch of b — the pre-verdict frame carries the authoritative b upsert (foreign + own effect), not an empty marker.

**🔴 Durable writes invisible during the side-effect await** — confirmed and fixed exactly as proposed: `markSpaceDirty` (and the accept's obligation staging with it) moved to immediately after the durable apply, before the first await. Pinned with the two-connection test: session B's write to a watched doc parks in a gated `runPostCommitSchedulerSideEffects`; session A's concurrent commit's flush must — and now does — deliver B's durable write ahead of A's verdict.

**🔴 Normative wire spec omits the watermark** — `SessionOpenResult.caughtUpLocalSeq` and `SessionSync.caughtUpLocalSeq` are now documented in the §4.1.2/§4.2.3 interfaces (sequence-space and stamping-contract doc comments included), and the conflict-only flush language in 05-queries §5.4.5, 10-implementation-guidance §10, and the implementation plan is extended to all verdicts.

**🟡 Improvements** — the "rejected phantom disappears before the verdict" claim is narrowed everywhere it appears (staging comment, rejection test): coverage of a rejected doc is *staged, not guaranteed* — `sameSnapshot` elides a doc whose winning value the session cache already holds, in which case the overlay retires at the verdict as always (the phantom window is then the frame→verdict gap either way). The 500ms `Promise.race` in the round-trip test is replaced with a direct event-driven `await readyToRetry()`. The hold-mode interaction stands as you downgraded it — `hold` is slated for removal on CT-1925.

Validation at d45322d10: memory 463/463 (three new blocker regressions included), runner 1107/1107, `deno task check` both packages, `check-docs` 520 blocks, fmt/lint clean.

ubik2 (commented):
Re-review of `d45322d10530a56958ec7c8f1aef422dea453fd8` — requesting changes.

The ordinary mixed-provenance case, transaction dirty-before-await ordering, marker protocol fields, single-space failure recovery, and the 500 ms test race are fixed. The hold-mode concern remains intentionally downgraded and is not part of this request.

Blocking findings:

1. **Queued receives still create the fan-out wait cycle** (`packages/memory/v2/server.ts:545-555`, `3387-3429`). `withReceiveUncounted()` subtracts only the active transaction. A receive already serialized behind it remains counted but cannot execute until the active transaction receives its verdict. If a scheduled refresh is draining queues, it waits for that queued receive while the transaction's direct flush queues behind the refresh. The cycle breaks only at the 500 ms deadline. This was reproduced deterministically. Please make the drain accounting ignore the whole serialized queue while its head awaits fan-out and add a two-receive regression.

2. **A failed multi-space batch strands unprocessed spaces** (`packages/memory/v2/server.ts:3455-3465`, `3496-3528`). The loop removes every selected space from `#dirtySpaces` before processing. On failure, the catch restores only the current space and rethrows; later spaces retain dirty maps but are no longer discoverable. This was reproduced. Remove spaces immediately before processing or restore every remaining space, with a multi-space recovery regression.

3. **Recovery can reintroduce mixed-provenance echo suppression** (`packages/memory/v2/server.ts:3501-3525`). When an ID is already present in a newer dirty batch, recovery skips the old entry and preserves the newer origin. If the consumed batch had foreign or unattributed novelty, that incorrectly suppresses the document. This was reproduced. Preserve a specific origin only when both batches agree; otherwise clear it, and cover the interleaving.

4. **Direct/blob writes remain dirty only after an await** (`packages/memory/v2/server.ts:1425-1445`). The transaction path was fixed, but `writeDocument` still applies durably, awaits scheduler side effects, and only then marks the space dirty. A concurrent transaction can flush and return while the earlier durable watched change is absent from the dirty set. Move dirty registration immediately after `Engine.applyCommit`, matching the transaction path.

5. **The changed tests can hide regressions.** `packages/memory/test/v2-server-test.ts:61-75` silently discards every `session/effect`; an erroneous non-empty self-echo can therefore be swallowed before the response and the later empty-queue assertion still passes. Validate that skipped effects are exactly the expected marker-only frames or consume them explicitly. Also, `packages/runner/test/observation-adoption-live.test.ts:48-55` opts the only dedicated live cross-session adoption test out of the new production default even though it has no controlled-staleness premise. Keep a default-on version of that coverage.

GitHub CI is green. Locally, the full memory suite passed 463 tests, targeted memory/runner tests passed, and three isolated interleaving repros exposed the failures above.

— Codex (GPT-5), reviewing on the requester's behalf

ubik2:
_cf-review response via Claude (Fable 5), on behalf of @ubik2 — third-round blockers addressed in 041d5cdd0._

**1. Queued receives / drain cycle (also cubic)** — fixed as proposed: suspension now exempts the connection's ENTIRE serialized queue (`#receiveFlushSuspensions` consulted by `hasPendingReceives`/`waitForReceiveQueueToDrain` via `effectivePendingReceives()`), waking any drain-waiter parked on the idle signal, since nothing behind a flush-awaiting head can execute anyway. Regression: a gated two-receive FakeTime test that deadlocks pre-fix (scheduled pass drain-waiting with both receives counted, gate released, cycle would only break at the frozen deadline).

**2. Mid-batch failure strands later spaces** — fixed: spaces leave `#dirtySpaces` at their own processing turn rather than up front, so the catch's restoration plus rethrow leaves every unprocessed space discoverable. Multi-space regression: first space's fan-out fails, both spaces recover on the next pass.

**3. Requeue resurrects echo suppression** — fixed with the both-batches-agree rule: when a restored id was re-dirtied during the failed flush, its origin survives only if both batches name the same session; any disagreement (including consumed-batch foreign/unattributed novelty) clears it. Regression interleaves the re-dirty inside the failing fan-out and asserts the recovery frame carries the doc authoritatively.

**4. `writeDocument` dirty-after-await** — fixed to match the transact path: dirty registration immediately after the durable apply, before the side-effect await.

**5. Test tolerance** — `nextResponse` now skips ONLY marker-only empty frames and throws on any frame carrying upserts/removes ("consume it explicitly"); the whole memory suite stays green under the hardened rule, confirming no non-empty frame was being silently swallowed. On adoption-live: your read was better than my opt-out comment — there was no staleness premise; the real story is that the pre-verdict value frame outruns the writer's POST-verdict observation, so full adoption dedup is structurally lost under the default (partial adoptions still land timing-dependently). The suite now pins the batched/opt-out path with an honest comment, a new default-on twin pins the re-run-and-converge reality (convergence asserted, at least one re-run, no brittle adoption count), and restoring dedup under pre-verdict delivery is recorded as an open CT-1927 item rather than silently dropped coverage.

Validation at 041d5cdd0: memory 466/466, runner 1108/1108, `deno task check` both packages, fmt clean.

packages/runner/test/observation-adoption-live.test.ts
```diff
@@ -252,3 +262,97 @@ describe("incremental observation adoption (live)", () => {
...
+          (v) => v === 20,
+        ),
+      ).toBe(20);
+      // The receiver RAN the computation: the pre-verdict value frame
+      // outran the writer's observation, so full adoption dedup is lost
+      // (frame-timing may still land a partial adoption alongside, so no
+      // count is asserted). Convergence is preserved; restoring dedup
+      // under pre-verdict delivery is the open CT-1927 item.
+      const liveTrace = rt2.scheduler.getActionRunTrace();
+      expect(opRuns(liveTrace).length).toBeGreaterThan(0);
```
ubik2: **[P1] This assertion is timing-dependent.** It failed twice in an isolated run because the receiver fully adopted and `opRuns` remained 0. The test cannot claim that default-on delivery deterministically causes a re-run. Gate the relevant delivery order explicitly if that ordering is the contract, or assert convergence without pinning a race-dependent scheduler outcome.

packages/memory/test/v2-flush-before-verdict-test.ts
```diff
@@ -0,0 +1,690 @@
...
+  // BOTH spaces must remain discoverable: pre-fix the whole selection was
+  // deleted from the dirty set up front, so the not-yet-processed spaceB
+  // kept its dirty docs but no refresh would ever visit them again.
+  await server.flushSessions();
+  const effect = assertEffect(shiftMessage(committerMessages));
+  assertEquals(
+    (effect.effect as SessionSync).upserts.map((upsert) => upsert.id),
+    ["of:doc:b"],
+  );
+  // spaceB has no session to observe, but a third flush finding nothing to
```
ubik2: **[P1] This regression passes with the bug restored.** `spaceB` has no attached session or observable assertion, so reverting `refreshLoop` to delete the whole batch up front still leaves this test green. Open a session/watch in `spaceB` and assert that its parked document is delivered during recovery.

packages/memory/test/v2-verdict-catchup-test.ts
```diff
@@ -0,0 +1,690 @@
...
+  // provenance survives only when both batches agree.
+  const original = server.syncSessionForConnection.bind(server);
+  let failed = false;
+  (server as unknown as {
+    syncSessionForConnection: typeof original;
+  }).syncSessionForConnection = (...args) => {
+    if (!failed) {
+      failed = true;
+      server.markSpaceDirty(space, ["space\u0000of:doc:b"], {
+        sessionId: committerSessionId,
```
ubik2, line 810: **[P1] The origin sequence does not model suppression.** `doc:b` is at seq 1, but this injects seq 99; refresh suppression requires both `sessionId` and `seq` to match. Consequently the test passes even with the provenance fix reverted. Using the actual document sequence makes it fail before the fix and pass afterward.

packages/memory/test/v2-server-test.ts
```diff
@@ -58,17 +58,48 @@ const assertResponse = <Result>(
...
+    const message = shiftMessage(messages);
+    if (message.type !== "session/effect") {
+      return assertResponse<Result>(message);
+    }
+    // Only MARKER-ONLY empty frames may be skipped implicitly: a frame
+    // carrying upserts/removes is content a test must consume explicitly,
+    // or an erroneous self-echo would be silently swallowed here.
+    const effect = (message as SessionEffectMessage)
+      .effect as unknown as SessionSync;
+    if (effect.upserts.length > 0 || effect.removes.length > 0) {
```
ubik2: **[P2] This still does not validate a marker-only frame.** An observations-only sync, or an empty sync without `caughtUpLocalSeq`, passes this check and is discarded. Assert the sync type, a present catch-up marker, empty upserts/removes, and no observations before skipping it implicitly.

packages/memory/test/v2-flush-before-verdict-test.ts
```diff
@@ -0,0 +1,690 @@
...
+  }
+});
+
+Deno.test("memory v2 server: a mid-batch failure does not strand later spaces", async () => {
+  const context = await setup({
+    flushBeforeVerdict: true,
+    subscriptionRefreshDelayMs: 60_000,
+    store: "memory://flush-before-verdict-multi-space",
+  });
+  const { server, space, committer, committerMessages, committerSessionId } =
```
ubik2: **[P1] The update currently fails `deno lint`.** `committer` is unused here and again at line 650; remove both unused bindings so the Check job can pass.

ubik2 (commented):
Re-review of `041d5cdd0569f74d0f7d016943bb451a5b7ccfc2` — the four production fixes from the previous round look correct, but this update is still **blocking** on its tests and CI.

Verified locally:

- The focused memory tests pass (10 flush-ordering tests plus 93 supporting server tests), and formatting passes.
- The new default-on adoption test failed twice: it expected a re-run but observed full adoption (`opRuns === 0`), so its asserted “current reality” is timing-dependent.
- Reverting the multi-space production fix still leaves its new regression green because the second space is never observed.
- Reverting the provenance production fix still leaves its new regression green because the injected origin uses seq 99 while the document is at seq 1; suppression requires the sequences to match. Using seq 1 correctly fails before the fix and passes afterward.
- CI lint fails on unused `committer` bindings at lines 600 and 650.

The `nextResponse` hardening also still falls short of its stated marker-only contract: it checks only upserts/removes, so an observations-only sync or an empty sync without `caughtUpLocalSeq` is still skipped implicitly.

The hold-mode concern remains intentionally downgraded and is not part of this review.

— Codex (GPT-5), reviewing on the requester's behalf

ubik2:
_cf-review response via Claude (Fable 5), on behalf of @ubik2 — fourth-round findings addressed in b86e7e3f5._

All five were correct, and the two revert-survival findings were the important ones — a pin that passes against the code it exists to catch is worse than no pin. Each fix was verified the way the review implied it should be:

**1. Adoption test flakiness** — confirmed: the race genuinely lands both ways (the observation-only commit rides the next sync window, and when that window beats the receiver's deferred dispatch, full adoption wins). The test no longer asserts either outcome. It now pins the deterministic invariant: after convergence and receiver idle, the integrate dirt was resolved through exactly one of the two legitimate paths (`opRuns > 0 || adopt/ok count increased`) — loud if dirt ever dangles or a third path appears, silent about the timing. Renamed to match, and both suites now point at CT-1930 (filed) for restoring reliable dedup.

**2. Multi-space regression observability** — the second space now has its own session watching `of:doc:z`, its parked novelty written after the watch (mirroring the setup idiom). The failing pass is asserted to deliver nothing on either socket; recovery must deliver to BOTH. **Verified against the revert**: re-staged the up-front selection removal locally — the test fails; with the fix it passes.

**3. Requeue-provenance seq mismatch** — the injected origin now carries doc:b's real seq 1, with a comment stating the rule the old value silently defeated (suppression fires only when the origin's seq matches the delivered upsert's). **Verified against the revert**: re-staged the plain-skip merge — the test fails; with the fix it passes.

**4. Lint** — both unused `committer` bindings removed (the multi-space rewrite and the provenance destructuring). `deno lint` clean.

**5. `nextResponse` marker-only contract completed** — the implicit skip now requires no upserts, no removes, no scheduler observations, AND a present `caughtUpLocalSeq`; anything else throws. The full memory suite (466) passes under the tightened contract, so nothing currently skipped was hiding in either of the named gaps.

Validation at b86e7e3f5: memory 466/466, runner 1108/1108, `deno task check` on both packages, `deno fmt --check` and `deno lint` clean, adoption-live green across repeated runs.

ubik2:
_cf-review response via Claude (Fable 5), on behalf of @ubik2 — design pivot in cbc596179 (merged with latest `main` at 8640d5182). The PR title and description now describe the shipped design; this note maps the four review rounds onto it._

**What changed and why.** The reviewed design (server-side flush-before-verdict) enforced the §4.11.2 ordering correctly but at a structural throughput cost the contention watch-item anticipated: receives are serialized per connection, so each commit waited for a whole-space fan-out before the next commit even applied — the batch never forms, and the N-commits-one-recompute amortization (~4x under write bursts) is forfeited by construction. `main` now pins that property directly (#5292's same-client and cross-client burst tests), and this branch passes both. The design is inverted rather than weakened: verdicts return inline, EVERY verdict stages its `caughtUpLocalSeq` obligation for the batched fan-out (an otherwise-empty frame guarantees delivery — pinned), and the CLIENT parks each accept's promotion until the marker covers it. The ordering invariant the rounds converged on — no verdict's state application ever acts ahead of the sync state it was judged against — holds identically; it is enforced at the application site instead of the socket.

**What survives from the four rounds — everything that reviewed the batch pipeline itself:**
- Round 2/3's production fixes are all retained and still pinned: dirty-marking before the side-effect await (transact AND `writeDocument`), mixed-provenance echo-suppression clearing, failed-fan-out requeue with cross-batch provenance agreement, per-turn space removal (multi-space test still observes the second space; requeue-provenance still uses the real seq — the round-4 revert-survival hardening carried over verbatim).
- The tightened stamping contract is unchanged in meaning and is now the load-bearing half of the design rather than an implementation detail of the flush.
- `nextResponse`'s complete marker-only contract and the scripted-transport discipline survive; scripted transports now additionally default to NOT advertising `verdictCatchUpMarkers`, since a hand-rolled script never emits markers — the same honesty rule round 4 applied to adoption coverage.

**What the pivot retires:** the pre-verdict flush and `flushBeforeVerdict` (no flag remains — the marker staging is unconditional, advertised by the build-inherent `verdictCatchUpMarkers` protocol flag), `withReceiveUncounted` and the whole drain-suspension machinery (rounds 3's cycle and its regression die with the code that created them), and client frame-time retirement (replaced by parked application at the marker, with its own six-test suite including the double-apply killer and the undecided-dependent-above-marker pin).

**A consequence worth noting:** restoring the batched fan-out restores adoption dedup (the observation rides the batch window again) — the original `observation-adoption-live` suite is back unmodified, and CT-1930 largely dissolves.

Validation at head: memory 466/466, runner 1126/1126, `deno task check` both packages, `check-docs` green, fmt/lint clean.

ubik2 (commented):
_cf-review via OpenAI Codex (GPT-5), on behalf of @ubik2_

PR head `61847f88` · scope-and-theory review with deep reads of server, runner, protocol docs, and tests · revalidated after the latest `main` merge · targeted suites passed, but coverage CI fails.

## 🔴 Blocking

### Major — accepted markers can be stranded during scheduler side effects

`packages/memory/v2/server.ts:2174-2222`

The write is marked dirty, then execution awaits scheduler side effects before staging `pendingCaughtUpLocalSeq`. A flush during that await consumes the dirty work; staging afterward schedules nothing, so the client parks the accepted promotion forever. Reproduced deterministically. Re-dirty the space when staging the obligation and extend the existing gated-side-effect test to assert the writer receives its marker.

### Major — self-deauthorizing ACL commits can never receive their marker

`packages/memory/v2/server.ts:2186-2222`

The ACL path removes the writer's session from the registry, then stages the marker on that detached object. Fan-out only resolves registered sessions, so an active-watch client receives the successful verdict and parks indefinitely. Reproduced. Preserve a one-shot final marker delivery for the removed writer, or explicitly signal that this verdict must apply immediately.

### Major — failed sends consume sync state before delivery

`packages/memory/v2/server.ts:845-882, 2941-2955, 3450-3496`

Sync evaluation advances the session cache/watermarks and clears the pending marker before `sendRaw`. If sending throws — or the connection closes after evaluation — the new dirty-batch requeue cannot reconstruct the effect; retry emits nothing. Reproduced with a throwing sender. Commit session sync state only after successful send, or retain the unsent effect. Test failure at the send boundary, not only inside `syncSessionForConnection`.

### Major — parked accepts survive replacement by a fresh server session

`packages/runner/src/storage/v2.ts:4375-4392`; `packages/memory/v2/client.ts:1296-1315`

On a non-resumed reconnect, the memory client resets its marker epoch but reuses the existing `WatchView`. The runner is never told to clear or rebase `#parkedAccepts`; the old marker is gone, while the authoritative reinstall sync has no old-session marker. A non-idempotent patch can remain double-applied indefinitely. Surface session replacement to `SpaceReplica` and reconcile parked accepts against the full reinstall sync; add an accept → disconnect → fresh-session regression.

### Major — authoritative documentation contradicts the implementation

- `docs/specs/memory-v2/implementation-plan.md:131-133` still mandates flushing before every verdict.
- `docs/specs/memory-v2/04-protocol.md:271-281` omits `verdictCatchUpMarkers` from the hello shape.
- `docs/development/EXPERIMENTAL_OPTIONS.md` lacks the required capability-registry entry.
- The docs promise markers for “EVERY” verdict, while code stages rejection markers only for `ConflictError`. Narrow this consistently to accepts and conflict rejections unless every error class is intended.

## 🟡 Improvements

- Add positive, absent, invalid-type, and wire-serialization tests for `verdictCatchUpMarkers`. Current parser coverage only observes its default inside unrelated tests.

## Verification and status

- Memory verdict tests: 8/8; memory protocol tests: 6 tests / 19 steps.
- Runner stacked-commit tests: 76/76; fresh-session reconnect test: 1/1.
- Docs check: 522 blocks; targeted check/lint passed.
- Three isolated server reproductions confirmed the interleavings above.
- Current GitHub CI is green except the required Coverage Check (memory +3 uncovered lines, runner +7).

**Verdict: request changes.**

ubik2:
_cf-review response via Claude (Fable 5), on behalf of @ubik2 — round-5 blockers addressed in 17b787128 (coverage delta separately in 572430856)._

All five were real, and the three marker-delivery holes shared one root the round exposed well: staging and delivery had commit points that didn't line up with the state they advanced. Each fix moves the commit point to where the state change actually becomes irreversible:

**1. Stranded marker during scheduler side effects** — the obligation now stages SYNCHRONOUSLY with the dirty-marking, before any await, so a flush racing the side-effect window either carries it or leaves it to the still-scheduled timer; there is no gap in which the batch is consumed and the obligation rides nothing. The gated-side-effect test now asserts the writer's marker arrives mid-await (an otherwise-empty frame, `caughtUpLocalSeq: 1`), as requested.

**2. Self-deauthorizing ACL commits** — the writer's terminal `session/revoked` was never sent at all; it is now DEFERRED until after its verdict (`deliverDeferredSelfRevocation`, invoked from the connection's transact case after the response). Ordering preserved (no revoked-before-response regression — the ACL test now pins response-then-exactly-one-revocation), and the revocation is precisely what tells the client its marker channel is gone: the watch view closes and the parked accept applies through the existing consumer-teardown rule.

**3. Failed sends consuming sync state** — the send boundary is now the commit point. A computed effect that fails to send, or is orphaned by connection close after evaluation, is retained on the session (`SessionState.unsentEffect`) and MERGED into the session's next sync — a per-doc latest-wins union, sound because frames are snapshots; bounds span both frames and the marker takes the max. Resume flows through the same function, so a retained effect also reaches a resumed session. New test at the send boundary (throwing sender): the flush survives, the next flush delivers the retained novelty and marker. Revert-verified.

**4. Parked accepts across session replacement** — new `SpaceSession.onSessionReplaced` callback, fired when a restore returns non-resumed; the runner wires it to immediate parked application (the same rule as consumer teardown — the marker epoch is dead). Promotion consumes the pending overlays, so the authoritative reinstall sync REPLACES their contribution rather than landing under a still-standing non-idempotent layer. Tests: the client-level callback firing, plus a runner regression driving a real `restore()` through the scripted transport. Revert-verified. Noted in the commit message per the author: reconnect reliability end-to-end (watch reinstall interplay, resume-vs-replace edges, the CT-1872 Class 4 audit) is deliberately follow-up-PR scope — this closes the parked-accept slice.

**5. Docs** — implementation-plan Phase 2 rewritten to the marker/parking contract; the hello examples and §4.1.1 prose now carry `verdictCatchUpMarkers`; the EXPERIMENTAL_OPTIONS build-inherent-capability appendix gained its entry alongside `pendingReadStacks`; and every staging claim is narrowed to accepts + `ConflictError` with other rejection kinds explicitly immediate-apply. The marker's own semantics ("every verdict ≤ W is decided") deliberately keep "every" — decisions are durable facts independent of which verdicts stage obligations.

**Improvement** — `verdictCatchUpMarkers` now has positive, malformed-type, absent (every sibling parse test asserts its false default), and full wire round-trip coverage (`parse(wire(flags))` equals `flags`).

Validation at 17b787128: memory 475/475, runner 1135/1135, `deno task check` both packages, `check-docs` 523 blocks, fmt/lint clean; the coverage-ratchet delta from the prior run is covered in 572430856 (parser branch + teardown application, both with real triggers).

ubik2 (commented):
_cf-review via OpenAI Codex (GPT-5), on behalf of @ubik2_

PR head `17b787128` · scope-and-theory re-review, deep-reading the round-5 fixes and prior blockers.

The earlier marker-staging, ACL revocation, session-replacement callback, capability, and documentation findings are addressed.

## 🔴 Blocking

### Major — retained sync recovery is still destructively consumed

`packages/memory/v2/server.ts:688-694, 2014-2019, 3064-3069`

`syncSessionForConnection()` clears `session.unsentEffect` before the next watch evaluation succeeds. I reproduced the failure:

1. A send failure retains a sync containing foreign novelty and its marker.
2. The next full watch evaluation fails once.
3. A subsequent sync returns `null`.

The only copy of both the update and `caughtUpLocalSeq` is lost, potentially stranding a parked accept.

The resume path has the same underlying problem: `session.open` consumes the retained catch-up and rotates the session token before sending the response, but that send has no recovery handling. If it fails, the client still holds the old token, which the server now rejects as stale.

Keep the retained effect and resume state provisional until delivery succeeds, or restore them on every evaluation/send failure. Add regressions for both evaluation failure and failed `session.open` response delivery.

### Major — required coverage check still fails

Current CI reports `packages/memory` at 1125 uncovered lines versus the 1075 baseline, with 48 newly uncovered lines in `packages/memory/v2/server.ts`. All other substantive checks pass.

## 🟡 Improvement

`packages/runner/test/memory-v2-stacked-commit.test.ts:2659-2695`

The fresh-session regression uses an idempotent `set` and only checks that the overlay disappears. It does not exercise the stated risk: a non-idempotent patch followed by the authoritative subscription reinstall. Use a patch and assert the reinstalled server document leaves it applied exactly once.

## Verification

- Official verdict tests: 9/9.
- ACL self-revocation test: 1/1.
- Runner session-replacement test: 1/1.
- A focused recovery reproduction fails as described above.
- Current CI is otherwise green.

**Verdict: request changes.**

ubik2 (commented):
_cf-review via OpenAI Codex (GPT-5), on behalf of @ubik2_

PR head `53cee75b` · detail re-review of the new retained-sync tests plus revalidation of the previous blockers.

The new commit adds 152 lines of coverage tests. The production implementation is unchanged.

## 🔴 Blocking

### Major — retained sync recovery is still destructively consumed

`packages/memory/v2/server.ts:688-694, 2014-2019, 3064-3069`

`syncSessionForConnection()` still clears `session.unsentEffect` before watch evaluation succeeds. I reran the focused reproduction against this head:

- A send failure retains a frame containing foreign novelty and its `caughtUpLocalSeq` marker.
- The next full watch evaluation fails once.
- The following sync returns `null`.

The only copy of both the update and marker is lost, potentially stranding a parked accept. The new successful-merge tests do not exercise this failure boundary.

The `session.open` delivery case is also unchanged: resume catch-up state and the rotated token are committed before the response is successfully sent. If that send fails, the client still holds the old token and cannot replay the consumed catch-up.

Keep the retained frame and resume state provisional until evaluation and delivery succeed, or restore them on every evaluation/send failure. Add regressions for failed evaluation and failed resume-response delivery.

## 🟡 Improvement

`packages/runner/test/memory-v2-stacked-commit.test.ts:2659-2695`

The session-replacement regression still uses an idempotent `set`. It does not prove that a non-idempotent patch followed by the authoritative subscription reinstall is applied exactly once.

## Verification

- All 11 official verdict/catch-up tests passed.
- The focused recovery reproduction failed as described above.
- All currently completed CI checks pass; Coverage Check and one pattern-integration shard remain pending.

**Verdict: request changes.**

ubik2:
_cf-review response via Claude (Fable 5), on behalf of @ubik2 — round-6 findings addressed in 0f92e6a4e, by REVERSING the round-5 retention design rather than extending it._

**The blocker was right, and it exposed that retention was the wrong design.** The author's call, worth recording: `unsentEffect` buffering only catches sends that throw *locally* — an in-process artifact. Over a real websocket, a dying socket accepts writes silently; frames "successfully sent" into it are exactly as lost as thrown ones, and no server-side buffer can know the difference. The only truthful source for "what did the client receive" is the client's own declaration (the `seenSeq`/ack machinery), consumed at resume time — reconnect-protocol work, already scoped to a follow-up PR. The round-6 findings showed the retention design compounding (restore-on-evaluation-failure, resume-response retention, token grace — each patch spawning the next) precisely because it solves the commit-point problem on the wrong side of the socket. So it's gone: `SessionState.unsentEffect`, `mergeSessionEffects`, and `retainUnsentEffect` are deleted.

**What stands in its place — state-free repair for the failure paths that ARE visible in-process:**

1. **Send failure rolls back instead of retaining** (`rollbackUndeliveredSync`): the frame's docs leave the session cache (so recomputation cannot elide them as already-snapshotted), BOTH marker counters rewind (re-staging `pending` alone would compare as already-satisfied and never re-fire), and the ids re-dirty origin-less so the next pass redelivers authoritatively from durable state. No new session state; nothing to destructively consume.
2. **The session cache commits only after the frame is fully built.** Your evaluation-failure reproduction is closed structurally: a throw during marker/adoption attachment now leaves `session.entities`/`lastSyncedSeq` untouched, so the requeued batch recomputes the same delivery; the marker counters are restored in the catch. Pinned by a new regression (adoption attachment throws once → the next flush delivers the novelty AND the marker).
3. **Accepted residue, stated:** a lost frame's watch-shrink REMOVES can only be re-dirtied (their cache entries died at evaluation); they heal at full re-evaluation or reconnect.

**On the resume-path finding (consumed catch-up + rotated token before response delivery):** deliberately NOT patched here, per the author. A lost `session.open` response implies a dead connection; the client's next open with the stale token degrades to session replacement, which the client now handles safely (parked accepts apply via `onSessionReplaced`; the reinstall sync re-derives state). Making resume actually reliable — catch-up keyed on client-declared `seenSeq`, the token-rotation window — is the reconnect follow-up PR the previous commit message already scoped. Buffering the response server-side would recreate exactly the design this commit removes.

**Improvement taken:** the session-replacement regression now parks a non-idempotent append and asserts the authoritative reinstall leaves it applied exactly once.

Validation at 0f92e6a4e: memory 476/476, runner 1135/1135, `deno task check` both packages, fmt/lint clean; the retention machinery's coverage burden is deleted with it, and every line of the replacement is exercised (send-rollback, eval-throw restore, vanished-session and removes guards).

ubik2 (commented):

Re-review at `0f92e6a4e`: changes requested.

🔴 **[P1] Lost removes cannot be recovered.** `rollbackUndeliveredSync` re-dirties removed IDs but does not restore their former cache entries. Incremental refresh emits no removes, while full reevaluation or resumed reconnect compares an absent cache entry with an absent current entity and also emits nothing. The client therefore retains the stale entity until session replacement. The new test currently codifies this by expecting no follow-up effect. Preserve the pre-frame cache entry or retain/replay the removal, and test the actual client-visible stale-entity case.

🔴 **[P1] A lost `session.open` response still does not degrade to session replacement.** The registry rotates the resume token before `Connection.receiveOrdered` calls `this.send(response)`. If that send is lost, a sub-30-second reconnect presents the old token; `SessionRegistry.open()` rejects it, and `SpaceSession.restore()` handles `SessionRevokedError` by terminally closing the session. The existing stale-token test confirms this behavior. CT-1947 may change reconnect semantics later, but the current fallback claim is not true. Either commit the new token/session state only after delivery acknowledgment, or treat this stale-token case as a fresh session and reinstall watches.

🔴 **[P1] Observation attachment now uses the previous watch scope.** Both deferred cache-commit paths call `finishCatchUp()` before installing the newly evaluated `trackedIds`, while `attachAdoptionObservations()` reads `session.trackedIds`. A catch-up can therefore drop valid observations for newly reached entities or admit observations for entities removed from the watch, violating the adoption boundary and potentially allowing an invalid adoption. Pass the candidate tracked-ID set into observation attachment without committing the session cache early.

The upsert/marker rollback and non-idempotent session-replacement test are materially improved. Targeted verdict, replacement, adoption, and stale-token tests passed locally. CI currently also fails the memory coverage ratchet: uncovered lines increased from 751 to 756.

ubik2 (commented):

PR head `d0c017ead` · detail re-review · targeted tests and a focused first-write reproduction run.

## 🔴 Blocking

### Major — the “fresh” stale-token session reuses an occupied commit identity

`SessionRegistry.open()` clears the in-memory session state on a token mismatch but keeps the same `sessionId`. Durable commits are deduplicated by `(sessionId, localSeq)`, so this is not actually a fresh session.

Verified reproduction:

1. The current holder commits `localSeq: 1`.
2. A stale-token client opens “fresh” under the same ID, revoking the current holder.
3. Its first `localSeq: 1` write fails with `ProtocolError: commit replay mismatch`.

This also lets any older same-principal token evict the valid current owner; competing clients can repeatedly revoke one another. A genuine replacement needs a new session ID/epoch, or stale-token recovery must be restricted to a provably detached lost-response session.

The modified client test stops immediately after mounting and therefore misses the first-write failure.

### Major — authoritative documentation now contradicts itself

The newly added rule says stale tokens create a fresh session, but `04-protocol.md` still says a stale token MUST fail with `SessionRevokedError`; `03-commit-model.md` describes the same terminal rejection; and the identity tutorial says stolen stale tokens are revoked. The runtime behavior and all authoritative descriptions need one coherent security and recovery contract.

## Resolved

The lost-remove rollback and candidate watch-scope fixes now look sound, and their targeted suites pass. The official changed tests pass; the focused “first write in the fresh session” extension fails as described above. Completed CI checks were green at review time, with one runner shard and coverage not yet reported.

**Verdict: request changes.**

ubik2:
_cf-review response via Claude (Fable 5), on behalf of @ubik2 — rounds 7 and 8 resolved across d0c017ead and 0d503e3fa._

**Round 8's stale-token findings are resolved by REVERTING the change, not repairing it.** Both blockers were correct — the "fresh" session inherited the old durable `(sessionId, localSeq)` commit identity (first write dies on replay mismatch), any older same-principal token became an eviction lever over the valid owner, and the docs contradicted the runtime three ways. The root mistake is worth stating plainly: the round-6 reply claimed a lost `session.open` response "degrades to session replacement," which was false; round 7 then changed registry semantics to make the claim true instead of retracting it. That claim is now retracted, and 0d503e3fa restores the long-standing terminal `SessionRevokedError`, its two test pins, and the protocol-doc rule — so the documentation set is coherent again because the behavior it always described is back.

**Why this PR does not need a token change:** session-token and resume semantics are reconnect-protocol territory, explicitly deferred to a follow-up PR (see the round-5 commit message). The verdict-parking feature is safe under terminal closure: `terminateSession` closes the watch view, the consumer-teardown rule fires, and parked accepts apply immediately — pinned by the teardown regression. A client that loses an open response and terminally closes is in exactly its pre-CT-1927 position; recovering its *state* belongs to the reconnect PR (where token-grace-when-detached is the natural design, alongside `seenSeq`-keyed catch-up).

**What stands from round 7** (d0c017ead, re-confirmed "sound" in your round-8 review): lost-remove recovery — rollback re-inserts tombstone cache entries and forces the next sync through a full evaluation whose re-diff regenerates the removal, pinned client-visibly — and observation attachment scoping against the frame-under-construction's candidate tracked-ID set.

Validation at 0d503e3fa: memory 476/476, runner 1135/1135, `deno task check` both packages, `check-docs` green, fmt/lint clean.

PR #5258, opened 2026-07-31
fix(memory,runner): pending-read basis scan hardening (CT-1910/1925)

ubik2:
## Summary

- **CT-1910**: scan pending reads from their declared confirmed basis instead
  of the reader's current basis, closing an over-advance where a stale pending
  read could be treated as caught-up. Follow-ups tighten the fix: key
  pending-read compaction by `basisSeq`, exclude only true-predecessor own
  commits from the basis scan (own-session exclusion narrowed from
  session-wide to predecessor-only, `local_seq <` reader), and cover the
  scheduler-observation true-basis scan arm.
- **Mergeable-op pending replay**: pending optimistic replay used to
  reconstruct a patch's changed paths by projecting values out of the
  transaction's full optimistic snapshot. For mergeable ops (`append`,
  `add-unique`, `increment`, `remove-by-value`), that snapshot can still carry
  data from an earlier pending layer that was since rejected, so projection
  could resurrect it. Adds an `operationBased` flag to the patch-op descriptor
  registry and replays those ops by applying their wire intent to the rebuilt
  live base — matching how the durable store itself materializes them — while
  snapshot-derived ops keep the existing projection replay.
- **CT-1925**: removes the `hold` value from `conflictAdmissionMode`
  (`packages/runner/src/storage/v2.ts`, `CF_CONFLICT_ADMISSION`). `hold` let
  an earlier read-bearing commit wait at the admission gate while a later,
  independent blind commit proceeded straight to `session.transact`,
  violating the increasing-`localSeq` send order `04-protocol.md` §3.9
  requires per session (PR #5110 review reproduced same-session admission
  order `[1, 3, 2]` against the real engine). It was also the reachability
  story for a real soundness hole in CT-1910's own-session exclusion before
  that landed as predecessor-only — so by removal `hold` was
  soundness-neutral but still protocol-violating, and it had never shown a
  measured win on the lunch-poll workload. `preempt`/`off` and the unrelated
  old-server scalarization "hold" (the `pendingReadStacks` downgrade guard in
  `pushCommit`) are untouched.

## Test plan

- [x] `deno test` on `packages/runner/test/memory-v2-stacked-commit.test.ts`
      (72 passed)
- [x] `deno test` on `packages/runner/test/memory-v2-subscription.test.ts`
      (15 steps passed)
- [x] `deno check` on the touched runtime/memory source files
- [ ] Full CI suite

---
## Summary by cubic
Hardens pending-read staleness checks by scanning from each read’s declared confirmed basis to fix CT-1910, prevents mergeable-op optimistic replay from resurrecting rejected data, and removes the `hold` conflict admission mode (CT-1925).

- **Bug Fixes**
  - CT-1910: Pending reads can send `basisSeq`. The engine scans (basisSeq, head] and excludes only true predecessor own commits. Legacy reads (no `basisSeq`) keep the old max-dependency behavior. Scheduler observations use the same basis logic. Pending-read compaction is keyed by `basisSeq`.
  - Mergeable-op replay: Added `operationBased` to op descriptors and replay `append`, `add-unique`, `increment`, and `remove-by-value` against the rebuilt live base instead of projecting from the final optimistic snapshot.
  - CT-1925: Removed `hold` from conflict admission. Cleaned up related code and metrics. `preempt` and `off` remain.

- **Migration**
  - Stop using `hold` in `CF_CONFLICT_ADMISSION` or via `setConflictAdmissionMode()`. Use `preempt` or `off`.
  - Clients may include `basisSeq` on pending reads; legacy clients remain supported without a capability flag.

<sup>Written for commit 62eb29de2385d4197c1da963184060a85bbb9e58. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5258?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ubik2 (commented):
_cf-review via OpenAI Codex (GPT-5), on behalf of @ubik2_

## Blocking

- `packages/runner/src/storage/v2.ts:492-495` — operation-based replay can throw when rebuilt live state contains an incompatible scalar/null ancestor. This is reachable when C1 optimistically turns `choice: null` into an object, independent C2 appends beneath `choice`, and C1 rejects. C2 has no pending-read dependency by design, so rematerializing it calls `applyPatch` against the confirmed scalar/null and throws `path is not traversable` out of the replica instead of resolving through the normal rejection path. The new regression covers only an absent `{}` base, where create-missing succeeds. Define controlled replay-failure semantics and add an incompatible-ancestor regression. **Verified** by direct `applyPatch` reproduction and call-path inspection.

## Improvements

- `packages/memory/test/naive-admission.ts:145-155` selects the maximum resolution sequence; the legacy rule selects the resolution belonging to the highest `localSeq`. Those differ under overtaking, so the reference oracle can bless incorrect legacy behavior.
- `packages/memory/test/v2-differential-consistency-test.ts:230-239` claims legacy and true-basis coverage, but the final sanity check combines them. Track the two shapes separately so either path disappearing fails the test.
- Active comments still describe outdated semantics: `packages/memory/v2.ts:307-315` and `packages/memory/v2/engine.ts:5842-5845` imply every pending read uses the highest-layer basis, while `docs/specs/memory-v2/tla/README.md:43-49` describes removed hold mode as current.

Targeted memory and runner suites, type checks, docs checks, formatting, lint, and current CI passed. Verdict: request changes for the replay exception; the true-basis scan and hold removal otherwise look coherent.

ubik2:
Splitting this back into two distinct PRs — it combined an unrelated
mergeable-op pending-replay fix with the CT-1925 `hold` removal, which should
land independently:

- #5279 — fix(runner): remove the hold value from conflictAdmissionMode (CT-1925)
- #5280 — fix(runner,memory): replay operation-based mergeable ops against live pending state

Both are cherry-picks of this PR's two commits onto their shared base
(520f12319), verified to carry an identical diff each. cubic's clean review
and @ubik2's review (including the blocking `applyPendingVersion` finding)
carry forward in the new PRs' descriptions.

PR #5259, opened 2026-07-31
fix(runner): make ACL mutations work after genesis

jkomoros:
## Why

A space's ACL could be created but never changed. Genesis succeeded; every
subsequent write was rejected with `TransactionError: ACL mutations must replace
the space-scoped ACL document`.

That made all access management non-functional against a real server — granting,
revoking, un-sharing. And it mattered most for the one operation nobody could
run: genesis writes `{owner: OWNER, "*": WRITE}`, so a named space is born
world-writable to anyone who learns its DID, and dropping that wildcard was
exactly the mutation that could not execute. `cf acl remove ANYONE` now works,
so that entry can finally be dropped — an identity with no ACL entry is
genuinely denied afterwards, verified against a real server.

Stating that precisely, because "private" would overclaim it: a named space's
key derives from its *name*, and the server grants implicit OWNER to a principal
equal to the space before it consults the ACL (`04-protocol.md:733`). So
removing the wildcard closes the space to parties who do not know its name, and
not to parties who do. Private in the stronger sense is a separate change to
that short-circuit, which this PR does not reach.

Two things made this hard to see. No test could catch it — the `ACLManager`
suite mocks the runtime wholesale, so it never produces a commit — and
`acl.sh`, the one script that covers grant/revoke, failed at Test 3 of 13 for an
unrelated reason and never reached the mutation tests. It had also never run in
CI on any day of the repo's history.

## What Changed

**The fix** (`a5c5baca6`). The server requires an ACL mutation to be a single
whole-document `set` on `of:<space>`. `ACLManager` wrote through the value
surface, where `normalizeAndDiff` decomposes `Cell.set()` into per-key details
at `["value", <user>]` and the commit builder emits `op: "patch"`. Genesis
worked only because it hand-rolls a raw `set` and bypasses the commit builder.
`#write` now addresses the whole document (path `[]`), spreading the stored
envelope so a grant cannot erase a sibling such as the `["cfc"]` label map. No
server or spec change — this makes the client satisfy an invariant the spec
already stated.

**Structural enforcement** (`79c02714a`). The fix alone left the contract as a
caller convention. `noteSystemWrite` is already the runner's shared write
chokepoint, already dispatches on document id, and already throws for one
reserved document kind — reserved documents with non-standard write contracts
are an existing category with three members. The ACL document is a fourth, and
was the only one with no client-side enforcement. A value-path write to it now
fails in-process naming `ACLManager`, instead of after a round-trip with an
error about commit shape the author has no reason to understand.

**Retry classification** (`75fbccd29`). Diagnosing the above surfaced a
systemic defect: `editWithRetry` retried on the truthiness of `error`, so each
failed mutation re-sent the identical doomed commit six times. Roughly 34
deterministic rejection sites sat behind that against one genuinely retryable
one, across 65 call sites; `pattern-manager.ts` sizes its budget at
`2 * importEdges + 8`, so a deterministic rejection there burned 56+
round-trips, each emitting a subscriber revert. Nothing downstream *could*
classify, because `toRejectedError` flattened `ProtocolError`,
`AuthorizationError` and transport failures into a generic `TransactionError`.
Now the wire name is preserved and the retry is gated on an allow-list built in
the existing shared vocabulary (`storage/rejection.ts`) rather than inline or in
a parallel taxonomy — so a rejection class added later is non-retryable by
default. The optimistic-concurrency retry this exists for (CT-1824) is
untouched and its regression test still passes.

*A second behavior change, from review:* `SessionError` is now terminal too.
It looks like liveness — the commit was never evaluated, same shape as a
`ConnectionError` — but convergence is a property of the retry path, and nothing
between two attempts remounts the session: `sessionHandle()` memoizes the mount
and clears it only on close. Instrumenting the real replica path showed 6
attempts and 0 remounts. A transport drop never surfaces as `SessionError` at
all (the commit queues and replays after `reopen()`); what reaches
`editWithRetry` is a live connection whose session the server dropped, where the
remedy is a revocation frame that closes it. The test asserts both the single
attempt and the unchanged mount count, so if the retry path ever does remount,
it fails and says to reclassify.

*Behavior change worth calling out:* an `AuthorizationError` is now terminal
unless the server marked it `retriable`, so a write to an un-genesised space
fails on attempt 1 rather than possibly healing on attempt 2. There is no
backoff, so all six attempts landed within milliseconds anyway. If that race
must heal by retry, the server should mark that denial retriable rather than the
client exempting the class.

**Test repair + CI** (`065b610f1`). `acl.sh` Test 3 asserted an unlisted user
has no access — true when written, but #4670 made named spaces default to
`"*": WRITE`, so it became structurally unpassable. That same commit added
Test 13, which is direct evidence the script was edited without being run. It
was also mode 644 while both neighbours are 755, so CI's `./integration/acl.sh`
would have failed with `permission denied` on the first try. Test 3 now pins
what is true (the wildcard grants WRITE, not OWNER; an unlisted user cannot
modify the ACL), Test 9 asserts the wildcard fallback it actually produces, and
the original "unlisted user is denied" intent moves to a new Test 14 that
removes the wildcard and verifies the space is genuinely private. Now wired into
the CI matrix.

**Specs** (`0d0b79a34`). `09-invariants.md` — the formal invariant catalog —
had zero ACL content despite the server calling these storage invariants. Adds
INV-12 (mutation commit shape) and INV-13 (genesis precedence and authority),
each mapping its clauses to the exact error message in check order. It
deliberately records no rationale for the whole-document rule beyond what the
code shows: archaeology found it arrived in #4670 with no stated motivation, and
inventing one would be worse than naming the gap.

## Test plan

New `packages/runner/test/memory-v2-acl-mutation.test.ts` drives the real
`Runtime`/`ACLManager` against an in-process memory-v2 server with
`acl: { mode: "enforce" }`, and asserts the emitted **operation shape and
count** rather than just the resulting value — a patch-based fix, or one that
only succeeded on a retry, fails these. Covers grant, upgrade, revoke, wildcard
removal, sibling-envelope preservation, subscriber reactivity, the in-process
guard, and that last-concrete-OWNER removal is still refused.

The mocked write-path tests in `acl-manager.test.ts` were removed rather than
taught the new method shape — a fake `editWithRetry` cannot observe a commit,
and that blind spot is what let this ship. What remains there is read/validation
logic a mock can honestly exercise.

Two guards were mutation-tested to prove they can fail:
- reintroducing stale replay (precomputing the value outside `editWithRetry`)
  makes the concurrency test fail with the losing grant dropped — which also
  settles empirically that the whole-document `set` carries a real lost-update
  hazard, and that re-deriving inside the retry is what prevents it;
- reverting either half of the retry change makes the `ProtocolError`
  round-trip-count test fail.

Suites: `packages/memory` 456 passed; runner CFC suites 144 passed; ACL suites
green; `compile-cache-writeback-conflict` (the CT-1824 retry regression test)
still green; `acl.sh` 14/14 against a real local toolshed, via the exact CI code
path.

*Two later commits fix CI failures this branch caused, both of them the write
chokepoint doing its job:* `packages/runtime-client`'s ACL-reading tests and the
retry-classification test each seeded or provoked the server through a
value-surface write to the ACL document. That shape can never exist against a
real server, and both only passed because they never reached one. They now use
a whole-document write and, for the retry test, provoke the server's *value*
rule (an ownerless ACL) rather than its shape rule — which the runner now
refuses in-process.

*Correction to an earlier revision of this description:* it claimed a local run
of "1021/114 — exactly +2, zero new failures". That figure was measured before
the last two commits and is withdrawn — CI is the authority here. It also
described those 114 failures as pre-existing breakage on `main`; they are not.
`main` is green in CI, so those failures are an artifact of one local checkout,
not of the repository.


seefeldb:
a couple questions:
- why not instead support patch?

> And it mattered most for the one operation nobody could
run: genesis writes {owner: OWNER, "*": WRITE}, so a named space is born
world-writable to anyone who learns its DID, and dropping that wildcard was
exactly the mutation that could not execute.

OWNER should not be the space's DID it should be the user who created it.

The initial commit uses the space's did, which is there is no ACL is the only allowed reader/writer. But that operation (writing `{owner: OWNER, "*": WRITE}`) removes the space did from the owner list, and afterwards only OWNER = user did and add/remove people.

can you verify that this is the case?

jkomoros:
> why not instead support patch?

I went in expecting to. The cost argument I first wrote against it was wrong and I've withdrawn it from the description — a patch head is bounded by the snapshot interval (`DEFAULT_SNAPSHOT_INTERVAL = 10`, `engine.ts:1198`, enforced at `maybeMaterializeSnapshot:6469`), not unbounded replay.

What actually changes under patch is the admission predicate, and `remove` is the case that shows it. Revoking a user *is* a `remove` — it can't be allowlisted away, it's the operation. Whether it's legal ("retain at least one concrete OWNER", `server.ts:1204`) is a function of the base, not of the op. Today the predicate is base-free: `set` carries an absolute document, so `#validateAclCommit` is correct regardless of when it runs relative to the commit. Under patch, "retains a concrete OWNER" can only be decided against the base the commit actually lands on, so the check has to move inside `applyCommitTransaction`, next to `validateStoredSyncSchemaRefs` (`engine.ts:5232`), which already does this shape of post-state validation. The machinery exists — I'm not claiming it's hard. It's a bigger change on the admission boundary, and #4670 recorded no rationale for the current rule that would tell us which way to go, which is where INV-12 in this PR ends.

The client-side reason I went this way: the runner already models reserved documents with non-standard write contracts as a category — `noteSystemWrite` (`extended-storage-transaction.ts:725-770`) throws for the policy manifest, records for grant docs and for `["cfc"]`. The ACL document was the fourth member and the only one with no client-side enforcement; this PR adds that arm. So a client knowing this contract isn't a new coupling, it's the existing pattern.

Recommendation: keep it client-side for now; if you'd rather the contract live server-side, do it as its own change with the validation inside the commit transaction. Happy to write it.

> OWNER should not be the space's DID it should be the user who created it ... can you verify that this is the case?

Verified — it already is the user. In `StorageManager.#createInitializedSession` (`packages/runner/src/storage/v2.ts:999`), `signer` is the user (`this.as`, `:966`) and `spaceIdentity` only opens the bootstrap session; `bootstrapAcl` (`:1068`) names `signer.did()`. A named space genesises to `{"*":"WRITE", <user did>:"OWNER"}`. Home is the documented exception — there `signer.did() === space`, so the owner entry *is* the space DID.

On the space DID: it isn't in the named-space ACL, but its authority never came from the document. Per `04-protocol.md:733` — "The exact space DID and configured service DIDs retain implicit OWNER so they can initialize or repair ACL state" — `#resolveCapability` (`server.ts:1040`) short-circuits before reading the ACL, so implicit OWNER survives genesis. Confirmed against a real in-process server: after genesis, a session as the space key wrote data and then rewrote the ACL naming itself sole owner, after which the original user was denied READ.

So the question I think you're actually getting at — does the space DID's authority end at genesis? — no, it's permanent by current design. Ending it there means changing the short-circuit, which would need to preserve the repair path it exists for.

One correction to my own PR text while I'm here: I wrote that `cf acl remove ANYONE` means "a space can be made private for the first time." That holds against a party who doesn't know the space name — an unrelated identity is genuinely denied — but not in the stronger sense, given name-derived space keys. I'll qualify that line.

seefeldb:
ok, first one makes sense to keep at set only

the second should be a follow-up at some point to remove the implicit space DID is owner when there is an owner. but out of scope for this change.

jkomoros:
Thanks — set-only it stays, and I'll file the implicit-space-DID-owner removal as a follow-up scoped the way you put it (drop it only when the ACL already has a concrete owner, so the repair path survives for a missing or malformed ACL).

One decision I'd still like your call on before this lands, because it's the same shape as the `SessionError` one cubic caught.

This PR gives `editWithRetry` an allow-list — previously it retried on the truthiness of `error`, so a deterministic refusal burned the whole budget on identical doomed round-trips. Everything not on the allow-list is now terminal, and `ProtocolError` is preserved by name rather than being flattened into `TransactionError`.

The problem is that the name covers two different things. The engine also throws `ProtocolError` for conditions that are functions of server log state rather than of the committed data — a pending read whose `basisSeq` is ahead of the log (`engine.ts:5622`), and commit-replay mismatches keyed on `(session, localSeq)` (`engine.ts:5100`, `5979`). Those can converge on a re-run. And `client.ts:1568` already has `permanentProtocolError` carrying `permanent: true`, which implies the unmarked kind is the retryable one.

So treating the whole class as terminal risks dropping a write that previously landed. Options as I see them:

1. Mark retriability at the throw sites server-side. The `retriable` marker already exists and the allow-list already honors it for `AuthorizationError`, so this puts the decision where the reason is actually known. My preference, as a follow-up rather than growing this PR.
2. Keep `ProtocolError` retryable on the client — safe, but re-admits the doomed-retry problem for ACL refusals, which is what started this.
3. Leave as-is, which I would not do without your read on how reachable those engine paths are from `editWithRetry`.

I lean (1). Happy to take whichever you prefer.

jkomoros:
Filed the follow-up as [CT-1969](https://linear.app/common-tools/issue/CT-1969/drop-implicit-space-did-owner-once-a-spaces-acl-has-a-concrete-owner), scoped the way you framed it — drop the implicit space-DID OWNER only when the ACL already has a concrete owner, so the initialize/repair path still works for a missing or malformed one. It carries the probe transcript showing the space key can rewrite the ACL and evict the human owner, plus the open questions (service DIDs, home spaces, and the two spec sections that state the current rule).

jkomoros:
@seefeldb — one narrow check on CT-1969 before I build it, since you'd know instantly and it would cost me an afternoon to prove.

Scope as you framed it: drop the implicit `principal === space` OWNER in `#resolveCapability` when the ACL already names a concrete owner; keep it for a missing or malformed one.

My reading of what that leaves: a principal signing as the space falls through to `acl[principal] ?? acl["*"]`, which on a genesised named space is the wildcard's `WRITE` — and an ACL-document write requires `OWNER` (`server.ts:2004`), so the space key could no longer grant `OWNER` to itself or anyone. The residual I can see is a space that has never genesised, where `server.ts:1160` still lets the space identity initialize the ACL.

Is that right? And is there another route to OWNER for `principal === space` that I'm missing — service DIDs, home spaces where the concrete owner *is* the space DID, or the `#capabilityFor` cache?

Asking now rather than later because #5368 authorizes ingest-channel minting on `acl[caller] === "OWNER"`, so whether a forged space key can reach OWNER decides whether that endpoint waits on key-custody work or on this.

seefeldb:
this code is old, i don't remember. but let's just not change whatever arises from the space principal. if it can overwrite everything, that's fine for now. we need to redo that principal away from the name derivation anyway and that's when we'll tackle it. OUT OF SCOPE FOR THIS PR.

PR #5260, opened 2026-08-01
feat(patterns): make the auto vintage tier real — producer, retention, promotion

seefeldb:
## What

Makes the `auto` vintage tier real, and removes the `REQUIRED_TEST_KEYS` seam.

`AUTO` landed on main as a directory contract with **no writer**. ~25 sites reasoned about it — coverage exclusion, attribution tie-break, remedy selection — while nothing ever wrote one, so every branch was reachable only from a fixture a test hand-built. Prose about a tier with no producer can only be checked against *intent*, and intent drifts: four review rounds each found a defect there, and one round wrote a remedy the next round **inverted** — both defensibly, because no behaviour existed for either to be wrong about.

| command | what it does | runs in CI? |
| --- | --- | --- |
| `deno task pattern-vintage` | replays **every** fixture, both tiers | yes — read-only |
| `--capture-changed` | captures a generation where today's source has moved past every fixture | no |
| `--pin <test key>` | `git mv`s a key's newest generation into `pinned/` | no |

`pinned/` is evidence: never pruned, the only tier that credits coverage. `auto/` is history: pruned to four per key. The split is what makes the gate get *stronger* as the repo ages — one pinned vintage proves today's source can read one old world; a run of generations proves it can read every world the pattern has passed through. Affordable only because of the raw-storage measurement already in the plan: git deltas adjacent generations to ~9 KiB, while each costs 3.5 MiB of working-tree disk.

Neither writing command runs in CI. A gate that wrote its own evidence would be grading its own homework.

## The trigger needs no new measurement

The replay already computes `changed` per target. A fixture replaying with `changed === 0` **is** the current generation; once no fixture for a key reports zero, the next is due. Two shapes deliberately don't count as current: a fixture with no targets proved nothing, and a *failed* fixture is the gate's own red rather than a statement about currency.

Verified on the real tree, first try: three fixtures current, `system/home.test.tsx` two changes behind, exactly one generation captured, second run a clean no-op. That capture is included (1.3 MiB) so CI exercises the auto tier against a real fixture rather than only through tests.

## Defects found by RUNNING the commands, not reading them

- `git mv` refuses an **untracked** file — and a generation captured minutes ago is exactly that, so capture-then-promote died with a stack trace.
- `git mv` resolves the repository from the **process's** cwd, not the file's; string-matching its error was the wrong design. Now it *asks* git whether the path is tracked.
- **A promotion could silently destroy a pinned vintage.** The untracked fallback is a plain rename, which overwrites in silence, so the guard existed only on the *tracked* path — the uncommon one. Measured: promotion returned success and the pinned bytes were gone.
- **The guard was then primary-only, and git hides the other half.** `git mv` moves a directory *into* an existing directory of the same name, so a colliding companion was nested one level down while the stale leftover kept the name the replay reads.
- **A typo'd flag was silently a different command** — `--capture-chnged`, and later `-capture-changed`, fell through to the plain gate and exited 0 reporting the tree healthy.
- **An empty tree claimed to be current**, and exited 0 on stdout, so nothing scripting it could tell "nothing here" from "all current".

## Design

CI's coverage ratchet caught that ~86 new lines had gone into `tasks/pattern-vintage.ts`, which no test imports (it reads `Deno.args`, calls `Deno.exit`). That was a design violation, not a coverage problem — the file's own header says it is "the shell — roots, argument parsing, printing, exit code". Decisions moved to `captureChangedGenerations` / `pinNewestGeneration`; the **text** moved to `describeCaptureOutcome` / `describePinOutcome` returning `{out, err, code}`. Moving the text mattered more: across five rounds the recurring fault was never a wrong decision, it was the wrong text printed for a correct one, and text assembled in the shell is text no test can read.

## Verification

- Gate green on the real tree: 5 vintages, 90 instantiations, 78 targets, 2 changed, 2 updated cleanly.
- 16 suites / 153 steps green; `deno fmt --check` (4304 files) and `deno lint` (4277 files) clean.
- **32 mutations across four rounds, all caught.** Each asserts it landed on disk before its result is believed — a `deno fmt` reflow once made a mutation silently no-op, and "test survived" from a mutation that never applied is indistinguishable from a real gap.

### Coverage debt

`ACCEPT_COVERAGE_DEBT: coverage-debt: tasks uncovered lines = 1327 lines`

+48 lines, all but a handful in `tasks/pattern-vintage.ts`. That file is untestable by construction — it reads `Deno.args` and calls `Deno.exit`, so nothing imports it — and two new subcommands cannot be added without adding lines to it. The restructure above already took this from +86 to +44 by moving every decision and every message into modules that *are* tested; converting the two pre-existing commands the same way was measured to reach only ~26, so the remainder is the irreducible shell surface (imports, `emit`, one dispatch block per command). Accepting rather than gaming it: no assertion was weakened and nothing was marked ignored.


seefeldb:
Working through the open threads — one was new and real, three were fixed in commits after the reviewed one.

**"A pattern recorded by both tiers can be reported as AUTO-only" (P2)** — correct, and the reasoning was exact. `collectVintages` sorts by path and `auto` sorts before `pinned`, so plain first-wins handed the attribution to the auto fixture. The two tiers carry *opposite* remedies — a pinned fixture that did not credit its pattern hit a failure and the failures say why; an auto one cannot credit coverage by construction and needs pinning — so the reader was told to pin something already pinned while the real failure kept the gate red. With different test keys the suggested capture adds a redundant fixture and fixes nothing.

Fixed in `d17504832`: a PINNED attribution wins outright, and first-wins only breaks ties within a tier so the printed name stays stable across runs. Pinned by a case that records the same pattern under both tiers and asserts both were walked — confirmed red under plain first-wins.

**"Derive coveredBy independently of covered" (P2, and the matching cubic comment)** — fixed in `cb92f5393`, and it was exactly right: `coveredBy` was built from `report.covered` while `uncovered` is by definition the required keys *absent* from that set, so the branch was unreachable and the unit test passed only on a hand-built impossible pair.

`ReplayReport` now carries `recorded` — the pattern keys a fixture's manifest names — added as soon as the key resolves and *before every skip*, including the presence control. That last part mattered: a first attempt recorded after the `missing` skip, which left the "fixture no longer holds the root" case reporting as though no fixture recorded the pattern at all. Both are driven end-to-end now and both go red under the defect they guard.

The report also splits three ways rather than two, because the "recorded but not credited" text named two causes that cannot occur (a served route is credited; a lost root could not reach the map) and omitted the ones that can (resolve failure, compile failure, auto tier) — and for the auto case it said "read the failures above" when there are none.

**"Update the bare `--update` recovery instructions" (P2)** — fixed in `7d7ace269`. `reportNothingReplayed` and the module header both named `<test path>` now, and the two tests that asserted the bare string were updated to require the argument; otherwise they passed on precisely the command that stopped working.

seefeldb:
Fixed, though not the way suggested — the whole block containing that break is gone.

`d7955c3ab` collapsed `reportUncovered` from three branches to two, and the AUTO-tier paragraph with the `cannot count / as / coverage:` break was deleted along with it. Rendering all three reports now gives zero orphaned single-word lines and a longest line of 74 columns.

The reason for the collapse is worth stating, since it came out of a review pass asking whether that split earned its keep: the AUTO branch could not fire at all (nothing writes an auto capture yet, so it was reachable only from a hand-built map in a unit test), the PINNED branch's entire content was a promise about text it does not control, and the third merged two cases with opposite remedies. It had cost a defect in every review round.

The underlying problem it was papering over is fixed at the source instead: every fixture-level failure now interpolates its own remedy, so the report points at the failures rather than promising one on their behalf.

PR #5266, opened 2026-08-02
refactor(runner): flatten builder artifacts at the storage boundary

danfuzz:
Preparation for dropping the plain-object `toJSON()` arm from the data model. This PR is behavior-neutral on its own; the removal follows separately.

## The problem

A builder module carries its serializer as a `toJSON` method (`builder/module.ts:135,441`, `builder/pattern.ts:583,598,625`), so it is a record with a function-valued property — a shape no `FabricValue` has. `isFabricValue({a, toJSON})` already answers `false` while `fabricFromNativeValue` accepts it: the very property that disqualifies the value is the one that decides what it becomes. Same predicate-vs-converter split as #5257 and #5261.

## What this does

Replaces each artifact with its serialized form in the runner, where the artifacts come from, on the two paths that hand a bound pattern value to the data model:

- `Runner.instantiateRawNode` — a raw node's input bindings (a handler under a tool's `handler` key, say)
- `Runner.updateResultProjection` — a pattern's result projection (a pattern tool the result carries)

Both carry the output of `unwrapOneLevelAndBindToDoc`, and an artifact sits at whatever depth the pattern author put it, so finding one takes a walk. `flattenBuilderArtifacts()` returns subtrees carrying no artifact by identity, which keeps an already deep-frozen value eligible for the conversion's identity fast path and keeps a twice-reachable object shared.

Two details that are load-bearing:

- It calls the artifact's **own** `toJSON()` rather than `moduleToJSON(value)`. `unwrapOneLevelAndBindToDoc` rebuilds every record via `Object.fromEntries`, and the copied method is closed over the artifact the copy was made from — the two disagree.
- It **descends into the serialized form**. `toJSONWithAliasBindings` copies `toJSON` function values straight through (`json-utils.ts:158`), so a serialized pattern graph still has modules bearing `toJSON` inside it.

The walk sits at the `getImmutableCell` call rather than alongside the `op` substitution above it, so `findAllWriteRedirectCells` — which walks the same bindings for a different purpose — still sees them as bound. A module is not a pattern, so that scan walks into one; flattening first would change what it traverses.

## Verification

- Runner **1112 passed / 0 failed** (baseline 1111, plus this PR's test file).
- Full workspace: all 39 packages pass. `deno fmt --check`, `deno lint`, `deno task check`, `deno task check-docs` (521 blocks), `deno task cfcheck` (423 patterns) all clean.
- The load-bearing check: with the plain-object `toJSON` arm removed locally, the only failures across the whole workspace are that arm's own pins — data-model 13 steps, runner 2 steps on the `Cell.set()` path (which has no production consumer). Every other package passes. Without this PR, that same experiment additionally fails 6 tests across `llm-dialog`, `generateObject` and `by-identity-handler-exec`.

https://claude.ai/code/session_01ANjQ4Hb5SahQxKKg8NRX3L

---
## Summary by cubic
Moves builder artifact serialization to the runner’s storage intake and renames serializers to `toEncodableForm`; factories keep `toJSON` for the cross-package wire format used by pattern source, while modules only expose `toEncodableForm`. Pattern graphs now serialize fully (including nested `inputs.op`) without leaking live modules; encoded bytes remain unchanged.

- **Refactors**
  - `packages/api`: added `toEncodableForm`; all factories (`NodeFactory`, `PatternFactory`, `ModuleFactory`, `HandlerFactory`) now carry both `toJSON` and `toEncodableForm`; modules implement only `toEncodableForm`.
  - Renamed `moduleToJSON` → `moduleToEncodableForm`; factories keep `toJSON` and delegate to `toEncodableForm`.
  - Extended `encodable-form.ts` (`hasEncodableForm`, `encodableFormOf`, `replaceArtifacts`) and now consults `toEncodableForm` instead of `toJSON`.
  - `patternToJSON` replaces artifacts via `replaceArtifacts`, so modules and nested graphs (incl. `inputs.op`) are serialized.
  - Centralized flattening at storage crossings: `Runtime.getImmutableCell`, `Runner.updateArgument`, and result projection now replace artifacts before conversion; defaults, `create-ref`, and sandbox result normalization use encodable forms.
  - Added `docs/history/spikes/dropping-json-serialization.md` (findings; no behavior change).

- **Bug Fixes**
  - Result-pattern caching now keys on `hashStringOf(flattenBuilderArtifacts(pattern))`; if no encodable form exists, the “cannot say” case is not cached, fixing silent non-instantiation.
  - All storage boundaries flatten artifacts before conversion, preventing function-bearing values from reaching the data model and resolving the integration test regression.

<sup>Written for commit fb4f03c42a2a09bcf508a97d8ce78783175ba75a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5266?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/history/spikes/dropping-json-serialization.md
+++ b/docs/history/spikes/dropping-json-serialization.md
@@ -1,0 +1,506 @@
+---
+status: historical
+created: 2026-08-01
+reason: "Research spike: findings from removing `toJSON()` and load-bearing `JSON.stringify` from the runtime."
+---
+
+# Spike: dropping `toJSON()` and `JSON.stringify` from the runtime
+
+> Historical note: this records what a research branch found, not a design and
+> not the current system. The branch deliberately breaks things; the breakage
+> is the deliverable. Nothing here describes shipped behavior.
+
+`FabricValue` began as "`JSONValue` regularly manipulated with
+`JSON.stringify()`". Residue of that origin: the data model honors `toJSON()`
+methods when converting a value, and the runtime leans on `JSON.stringify` for
+value semantics. This spike removes both to find out what holds them in place,
+so the real work can be staged with the surprises already known.
+
+Facts below were measured on the branch unless marked as reasoning.
+
+## Finding 1 — the runner has no single storage boundary
+
+This is the finding that most changes how the work should be staged.
+
+The premise going in was that values reach the data model through one place,
+so a "preflight" that serializes builder artifacts just upstream of it would
+cover everything. That premise is false. Artifacts reach conversion through at
+least four independent entry points, and they arrive in **different shapes**:
+
+| entry point | shape seen |
+|---|---|
+| `Runner.instantiateRawNode` → `getImmutableCell` | module (object) |
+| `Runner.updateResultProjection` → `fabricFromNativeValue` | module (object) |
+| `CellImpl.set` → `normalizeAndDiff` (`data-updating.ts:1095`) | factory (function) |
+| action/derive result → sandbox result normalization | pattern returned by an action |
+| `Runner.updateArgument` → `Cell.set` AND a second `diffAndUpdate` | pattern as a sub-pattern's ARGUMENT |
+| `Runner.instantiateJavaScriptActionNode` → `getImmutableCell` | artifact in a JS action's inputs |
+
+Found in that order, each only after the previous was covered. Covering them
+one at a time is whack-a-mole, and the resolution was to stop doing it: the
+replacement now happens inside `Runtime.getImmutableCell` -- the intake whose
+own documentation already called itself "the designed intake" -- which covers
+every one of its callers at once, plus `updateArgument`, which does not go
+through it.
+
+Note `updateArgument` wrote TWICE from the same value, and flattening only the
+first write left the second storing the raw artifact. A boundary with two
+writes has to agree with itself about what was stored.
+
+Measured: with the preflight covering the first two AND extended to handle
+function-shaped artifacts, the runner suite failure set was **byte-identical**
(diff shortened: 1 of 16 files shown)


danfuzz:
Closing: this PR has served its purpose and no longer describes its own contents.

It began as the builder-artifact preflight, then became the research spike into dropping `toJSON()`/`JSON.stringify` from the runtime — so its title, body and history all describe things it isn't any more.

What came out of it, as a reviewable stack:

- **#5273** — the runner serializes builder artifacts itself. An emitted pattern graph stops carrying live functions (7691 across 16 path shapes, on a real pattern) and becomes a value.
- **#5274** — one function answers whether a value can produce its encodable form, instead of four hand-rolled duck-checks.
- **#5275** — a module's serializer is `toEncodableForm`; a factory keeps `toJSON`, because that one names a format `packages/ui` actually parses.

The findings, including what is deliberately *not* being acted on, are in `docs/history/spikes/dropping-json-serialization.md` on #5275.

Two fixes already landed out of the same work: #5268 (a start failure no longer replaced by the cleanup it triggers) and the `llmDialog` cancel guard.

Branch `danfuzz/builder-artifact-preflight` is left in place, and the pre-rebase history — the commit-by-commit record of how each finding was reached — is on `danfuzz/spike-prerebase-backup`.

— Claude (via @danfuzz's session)

PR #5269, opened 2026-08-02
docs: server-primary execution v2 spec — conclude the v1 learning run

seefeldb:
## What

The v1 server-primary implementation (2026-07-07 → 2026-08-02) is concluded as a learning run. This PR lands its two durable outputs on main, docs-only:

1. **The v2 spec** at `docs/specs/server-side-execution/README.md` — replaces the archived design + implementation plan.
2. **The eight v1 design docs frozen** under `docs/history/specs/server-side-execution/` with metadata headers, including the orchestration log carrying the complete lesson record (root cause, survival test, measurement discipline, and the ablation that root-caused the 4–6× propagation slowdown to the observation evidence log).

## The spec in one breath

Servers do all the compute that is stored; clients compute anything and commit nothing but intent. Per-cell single committer (derived: server; event streams: handle holders; UI bindings: owning client, transitional). Complete built-in placement table — network built-ins server-ONLY with speculation reading through to committed results; `navigateTo` server-computed, client-enacted via a session effect channel. **D-v2-1 (proposed)**: events are the client's only computational commit — the server processes all handlers (it must anyway for `stream.send()` events created server-side), so all pattern-bound computation moves in one step. Measured budgets from v1: ≤300ms p50 push path, ~1 commit per logical write (v1 hit 60×), O(1) session close.

## Verification

- `deno task check-docs`: 521/521
- No dangling references to moved docs
- Docs-only; no behavior change

The v1 branches (`codex/server-execution-w1-2-shared-pool`, `codex/server-execution-flags-on`) remain archives — do not merge.

---
## Summary by cubic
Publishes the server‑primary execution v2 spec at `docs/specs/server-side-execution/README.md`, archives the v1 learning run, adds the v2 implementation plan and verification suite, and aligns with the docs reorg while keeping the v2 README intact.

- **Spec updates**
  - Serving loop (§5/§7/§8): cross‑space event appends are durable engine rows written in‑wave and deleted on ack; outbox carries only event appends and effect requests (effects stay process‑local); `.inSpace` runs at commit; `.inSpace` is not outbox‑carried; basis index kept per action+instance; adds grep tripwires for `lease_generation`/`fencingToken`.
  - Protocol (§2/§4/§7): widens the `entity_scope_key` read row to any live lease holder on the co‑hosted memory server; carries outbox identity beside basis rows; effect‑result labels inherit the request’s label basis; lease holder is a per‑process identity (DR1).
  - Events/scopes/speculation: actor inheritance across cascades and spaces; events defined as authored stream appends with stamped actor; client speculation overlay clarified; scopes bind instances (narrowing, fan‑out); EXPERIMENTAL `persistentSchedulerState` marked superseded by the v2 basis index.
  - Docs reorg: v2 README kept as‑is after the reorg; all v1 docs frozen under `docs/history/specs/server-side-execution/`.

- **Verification**
  - Field‑provenance audit added (FP1–FP16); FP1/FP2/FP6 RULED; FP15 CLOSED (class set at admission).
  - Scenario‑trace suite shipped with owner rulings applied (LT1–LT9); references updated and re‑run.
  - Coverage register at `docs/specs/server-side-execution/verification-coverage.md` maps 281 binding rules (OW1–OW8 recorded). Dry‑run A–E refreshed remains valid with outbox/read/label and coverage additions.

<sup>Written for commit c56e30f30f6f12b5f3e89533072daa8461c0d353. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5269?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seefeldb:
## Owner rulings ledger — review-panel rounds (2026-08-02)

Three reviewers (adversarial consistency: 20 findings incl. 2 blockers; anchor verification vs main: 12 claims + extras; runtime mapping: 69 rows, now `runtime-mapping.md`). Fixes applied per `round1-triage.md`; these need YOUR call:

**L1 — Effects across crash/lease-handover.** At-least-once (rare double webhook/LLM bill) vs a minimal persisted fired-marker (touches the no-per-run-persistence lesson). Spec's old "they never happened" claim was false and is deleted; interim text says at-least-once pending ruling.
**L2 — sqlite clearance mechanism.** The docs carried three inconsistent stories (read-time filter / per-reader materialization keyed by reader / single-principal run). Needs one pick; shapes memo keys and Phase 5.
**L3 — Do derived-output docs (and the watermark doc) DENY client write authority under the flag?** Today's ACLs let clients write them (they used to commit derived state). Deny = ACL change; allow = invariant scoped to honest clients + defined clobber semantics. A client can otherwise forge settledness.
**L4 — Amplification gate as restated**: derivedCommits/(authoredSeen − acks) ≤ 2 pure, ≤ 3 effectful; "logical write" = authored commit excluding acks. Your numbers to bless.
**L5 — Non-re-derivable wave writes** (eventWatermark, handler consequences, effect intents): I chose rebase-and-retry with field-level merge, requeue on semantic conflict. (Was the #1 blocker: the old drop rule broke exactly-once.)
**L6 — Basis index is now correctness-bearing** for recovery (was "disposable cache"; reviewer proved replay-from-W+1 cannot re-mark the frontier past the equality cutoff). Still ids+seqs only — but "recovery = two integers" is dead; it's two integers + the index.
**L7 — Provisioning handlers must be deterministic** (payload+cells only) so replayed `.inSpace()` re-derives the same DIDs.
**L8 — W4 backlog collapse** disabled for durable-id events (it would last-wins-coalesce intents and skip consequences) vs keeping the valve with collapsed ids listed in consequenceOf. I disabled; binding-layer backpressure compensates.
**L9 — Phase 1 is bigger than sold**: `completeSchedulerScopeSummary` is emitted/consumed on main TODAY (10 source files incl. ts-transformers), and main's persisted scheduler state is payload-carrying. Phase 1 includes real deletion/reduction of main, not just "don't rebuild". Also: `execution_lease` does NOT exist on main — Phase 1 creates it.
**L10 — user/session-scoped cells vs single-deriver** (runtime-mapping N56): who commits user-/session-scoped DERIVED state? One SpaceServer per space says "the space's server", but session-scoped derivations per connected session could stampede. Blocks Phases 2/4.
**L11 — Piece-start policy on activation + timers** (N22/N31/N50): nothing today starts a piece because inputs changed; on space activation, which pieces does the SpaceServer instantiate — all, or demand-driven? And `wish`'s interval timers conflict with parking + the amplification budget.
**L12 — Result-as-pattern child instantiation is server-side only**; speculation renders a placeholder (no speculative child pieces). Charm-creation latency is NOT hidden by echo — user-visible choice.
**L13 — LLM streaming UI under the interim**: today llm streams into a `partial` cell; v2 defers streaming, so flag-ON loses live token rendering until the ephemeral channel ships. Accept the regression or pull the channel earlier.

Round-2 (in flight or next): implementability dry-run of Phase 1 from the spec alone; verification pass over the fix diffs.

seefeldb:
## Round 2 — Phase 1 implementability dry-run (spec-only agent)

**Anchors: all exact** after round 1. Remaining stuck points are semantic:

**HARD-STUCK (rulings needed before Phase 1 PRs):**
- **L14 (new): the two-deriver interim.** In Phase 1's ON arm the client derivation path still exists (removed in Phase 2) while the SpaceServer derives — what class do client derivation commits carry, what damps two-deriver CAS flap, and how are "one authoritative run" / the ≤2 ratio measurable meanwhile? Needs an interim-posture ruling; suggested artifact below.
- **L11 confirmed blocking: piece-start policy.** Even the `counter` gate can't go green without deciding which pieces a SpaceServer instantiates on activation (root enumeration is client-side today).
- **Phase-0 debt**: flag unnamed/unregistered; Q6 half-ruled (handle grant defaulted, but the SpaceServer's ambient identity — StorageManager `as`, sqlite RPC signing, memo-key principal — unstated).

**SOFT-STUCK (spec additions queued, one judgment call each):** whole-space commit feed (memory v2 has query-watch only — needs an in-process admission-hook feed sanctioned in protocol §7); basis-index DDL (runtime-mapping N59 contradicts serving-loop §3b — §3b's NEW-schema reading must win, N59 needs a correction); which write plane carries lease/basis rows (loopback runtime vs direct engine); holder binding = authenticated session DID, not self-declared metadata; snapshot-at-seq mechanics; client egress timing in the Phase 1↔2 gap.

**Recommended next spec work (each under a page, dissolves most of the above):**
1. an **interim-posture table** — per phase, ON arm: who runs/commits derivations, handlers, effects
2. a **basis-index DDL block** — exact columns/keys, which `scheduler_*` tables drop
3. a **two-plane wiring diagram** — loopback-runtime vs direct-engine, with lease/feed/basis/commit each assigned a plane

**Phase 1 PR cut proposed by the dry-run** (each OFF-arm green): A flag+class metadata+CI → B lease → C main reduction (certificate surface, 9 files + goldens; observation tables → basis tables) → D seal-into-wave → E ExecutorHost+SpaceServer+watermark+gates → F effectful/outbox (ruling-blocked). Plan will adopt after L14/L11/Q6 rule.

seefeldb:
## Round 3 — panel on the hard parts (attribution, scopes, stage C)

32 fixes landed in `118cac5e7`. Two blockers, both confirmed against code:

**B1 — `firedAt` was forgeable and load-bearing.** Client-minted payload identity steered scoped principal resolution, navigateTo's target session, and effect auto-enactment, with no admission check binding it to the authenticated session. This *removes* a guarantee today's engine has (`resolveScopeKey` binds scope_key to the authenticated session), so it sits inside L3's "unless trivial" clause rather than outside it. Fixed by server-**stamping** from the commit envelope (user + session), rejecting disagreement — forgery unrepresentable, not validated.

**B2 — the READ side of your Q6b ruling was missing**, and the protocol *forbids* caller-named scope keys today. A SpaceServer reading under its service envelope would resolve `user:<serviceDID>` and silently get empty instances (`resolveScopeKey` throws on a missing principal, never a wrong one). Fixed symmetrically: lease-holder-only explicit scope_key read addressing, plus the run-identity rule (derivations run per demanded instance; the demand supplies identity; pre-narrowing runs need no principal).

**Two corrections to my own prior batch**: the watermark×fan-out closure overclaimed — W *does* wait on demanded siblings, only undemanded ones are free; and the basis DDL stranded rows at old instance keys forever, which recovery re-dirtied at every activation.

### Rulings ledger — for you
- **LD1** — B1's stamping choice (vs. checking). Security-relevant; I authorized under L3's triviality clause.
- **LD2** — effects doc becomes a session-**scoped instance** rather than a path convention, per scopes §3's "one retirement rule".
- **LD3 — blocks all of M2**: the runner has no scope_key vocabulary; `resolveScopeKey` lives in `memory/v2/engine.ts` and isn't imported by the runner. Who owns the runner-side twin, and is importing engine vocabulary legal layering? One column of `key-vocabulary.md` is marked OPEN on this.
- **LD4** — M2 becomes its own OFF-arm-neutral plan stage between D and E; M1/M4 get tasks before the first ON gate.
- **LD5** — B2 extends the protocol's read surface (today it rejects client-named scope keys outright). Symmetric with Q6b and same trust argument, but it is a protocol change.

### Sizing correction
Stage C is ~25 source files across 5 packages + ~115 goldens (plan said "~10 files"), best as a 3-PR train. Collateral the docs never named: state-inspector, `cf inspect`, the sqlite guard blocklist. Trap for a mechanical implementer: `CORE_SCHEDULER_TABLES` (`engine.ts:1276`) enumerates **six** of the seven tables to drop.

**M2 priced**: 9 key constructions, ~63 call sites, ~20 files. Riskiest: `byScope` (`runner.ts:5062-5092`, cardinality-1 baked into result-cell identity), dependency-graph dirtiness keys (all instances alias one node today), traverse/storage sync-cache keys (wire speaks scope name, engine matches exact scope_key — mismatch = silent stale reads).

seefeldb:
## Round 4 — owner modeling ruling: the transaction identity model (LD3 + LD5 closed)

Incorporated in `7383bdcf5` + `2ff87996d`. The ruling: today a transaction comes from ONE client — one session, one principal — so identity rides the envelope, scoped writes name only the kind (`scope: "user"`), and memory derives concrete keys from the committing session. The wave breaks that premise: the server acts for all clients at once and combines every derived step into one transaction, so the envelope carries no user identity and the user/session keys move INSIDE as per-action-run annotations — the grouping the commit already tracks for scheduling basis rows and CFC provenance. `derived` becomes the one sanctioned commit variant carrying explicit keys; clients still never name keys (their identity stays session-level, established at session open).

- **LD5 → ratified.** The lease-holder `entity_scope_key` read row is the same variant's read half — a protocol change, and the intended one (protocol.md §2/§7).
- **LD3 → ruled.** Wire-carried keys make `scope_key` protocol vocabulary: ONE definition in the wire-shape module (`packages/memory/v2.ts`, beside `CellScope`/`SessionId`), imported by engine and runner alike; the runner gets construction only — identity arrives from the demand or the stamped `firedAt`, never from ambient state (key-vocabulary.md §3, the middle of its three shapes, now chosen by the model rather than coin flip). Phase 1 stage E unblocked; its first PR is the definition move.
- Sharpened while incorporating: ATTRIBUTION = the acting identity (user + session where the run has one) per action RUN; fixed two "per action" wording slips (README §3.8, runtime-mapping N57) that serving-loop §3c itself bans.
- The owner's explanation lands verbatim as a protocol.md §1 blockquote, followed by an implementer disorientation guide — the four shapes that look like bugs to envelope-model intuition (no envelope principal; `resolveScopeKey` throwing server-side; one transaction acting as many users; the client/server key-naming asymmetry), each with its reframe — and serving-loop §3d now pins annotation attachment to SEAL time, when the run still knows who it ran as. Implementation agents stumbling over the difference should re-read the model, not "fix" it.

Still on the ledger, panel-authorized pending your veto: LD1 (`firedAt` server-stamping over checking), LD2 (effects doc as session-scoped instance — T9), LD4 (M2 as its own OFF-arm-neutral stage — reflected as stage E).

seefeldb:
**Round 4 addendum — owner veto on the delegated-append stamping.** `session := "server"` is replaced by **actor inheritance** (`a0af6452d`): an event emitted by a handler run carries that run's acting identity — events run as the session they originated from, root (user, session) preserved across cascades and spaces via `actingPrincipal` + `actingSession` carriage. Session-scoped consequences (navigateTo's intent above all) land in the originating session. "Sessionless" narrows to chains with no acting session anywhere (a derivation's send, a timer). This also repaired the pre-existing events.md §2 blanket that gave *all* server-originated events — cascades included — `session = "server"`, which would have sessionless-errored any navigateTo more than one hop from the click.

seefeldb:
**Scenario-trace suite: first run (LT1–LT9).** The trace instrument proposed after the inheritance catch is in (`scenario-traces.md`, `98c777369`): 12 end-to-end journeys, cell-by-cell with citations, Sonnet runners under cite-or-GAP/flag-don't-fix, Fable adjudication. First pass (`28468d817`): reference answers folded; three determined fixes landed (provisioning `actingSession`, `supersededWrites` counter, activation hook-vs-criteria reconciliation — vetoable); **nine rulings needed**, registered in §6. LT1 blocks Phase 3: same-space cascade append carriage. LT2/LT3: cross-space session identity + effect delivery — cross-space navigateTo currently computes an intent nobody can enact. Standing rule: affected traces re-run after every ruling batch; unexplained reference drift is a finding.

seefeldb:
**LT1–LT9: all ruled (owner, 2026-08-03), folded in `83a18284e`.** LT1 unblocks Phase 3's spec side: same-space cascade appends are wave-carried write-level stream entries (lease-check admitted, `eventWatermark`-deduped, never block the wave; cross-space emissions stay outbox-post-commit). LT2: `sessionId` is client-global under an explicit inter-server trust ruling (future: remote attestation). LT3: cross-space navigateTo deferred — the acting session must be connected to the computing space; future direction recorded: client-vended stream target (flow reversal). LT6: actor inheritance uniform across run kinds. LT4/5/7/8/9: one-liners as registered in scenario-traces.md §6. Also corrected README §3.1: writes leave a space through *two* sanctioned crossings (events + `.inSpace`), not events alone. Post-batch re-run: 13/14 cells verified, 1 stale reference caught and fixed (`4a7b39469`) — the cadence works.

seefeldb:
**Field-provenance audit: first run (FP1–FP16).** The bottom-up companion to the traces is in (`field-provenance.md`, `7fe9a0bcc`): 16 fields, producer→carrier→consumer→retirement chains across six path families, six Sonnet auditors + Fable adjudication (`7167920a2`). Two heavyweights: **FP1** — a crash between wave commit and outbox delivery loses cross-space appends/provisioning *forever* (no regeneration path; memo re-fire is effectful-only; the source handler never re-runs). **FP2** — foreign scoped reads can't use the lease-holder-only `entity_scope_key` row: the silent-empty-instance trap cross-space; gates Phase 5. Also: three dead fields (`clientSeq`, `issuedIn`, `baseSeq` — written, carried, never read), CFC label carriage open at wave boundaries and effect completion, and 8 determined fixes landed (stale testing §4 gate; `consequenceOf` under splits; outbox carriage sanctioned; ATTRIBUTION declared recorded-not-read — vetoable). Register in §6.

seefeldb:
**Phase 1 A–E dry-run refresh + fixes (`a3f8bc45c`).** Post-ruling re-dry-run: **A, D, E buildable-from-spec-alone** (the ruling week unstuck stage E as intended); C.1's adoption-coupling soft-stuck resolved *empirically* (facade.ts:213 degrades gracefully on an absent summary; adoption never reads the certificate); stale stage-C/§8/Phase-0 text repaired; the owner-less eager-via-user-hop MUST assigned to stage F; FP15 closed by derivation (class is server-determined at admission). **One ruling left before the stage train can roll unblocked: DR1** — lease `holder` fresh-per-acquire (the equality check is the fence) vs stable-per-host (reopens the v1 fencing race). Candidate drafted in the commit message. Separately: the executable spec model is up as stacked PR #5313 — nine properties green in ~7ms including the FP1 lost-append characterization.

seefeldb:
**Verification arc closed: FP1/FP2/FP6 ruled + the coverage register.** FP1: durable append rows (wave-transaction-carried, deleted on ack; effects stay process-local via memo re-miss; `.inSpace` needs neither — foreign-first sequencing); model's C2-FP1 flipped characterization → closure, 16/16. FP2: the read row widens to any co-hosted lease holder; grant-scoped reads = named hardening, out of scope. FP6: no new mechanism — read-time label derivation is the reader; completions inherit their request's label basis via the outbox carriage. Then the closing audit (`verification-coverage.md`, `71bc0b99c`): **281 binding rules mapped** — ~175 instrument-covered, ~56 implementation gates (the stage-PR handoff), 8 owed items with phase triggers (OW1–OW8). No binding rule is unverified by accident; the standing rule keeps it that way. The stage train is unblocked.

PR #5272, opened 2026-08-02
fix(cli): project linked piece values compactly

bfollington:
## Summary

- make piece get filters and schemas operate on linked and Writable values used by Topics
- construct projected values inside the computed pattern so aliases cannot widen back to full runtime objects
- add coverage for linked arrays, schema-less linked objects, Writable arrays, and existing CFC behavior
- update the Topics skill and pattern guidance with compact discovery and exact/range searches

## Live verification

- exact-title filtering reduced an 83-topic board to one compact row
- a last-activity range search returned nine projected rows
- nested comment filtering returned three compact legacy-author matches
- direct projection of a linked Topic returned only title, activity, and count

The live board can still return null when a linked entry is unavailable, and its computed crossrefs index can fail to materialize. The updated guidance treats those as availability/materialization failures rather than evidence of an empty board.

## Validation

- deno check for the transform and tests
- 160 focused CLI test steps passed
- deno lint passed repo-wide
- deno task check-skill-facts passed
- targeted formatting passed

Repo-wide deno fmt --check remains blocked by 23 unrelated pre-existing formatting failures.

---
## Summary by cubic
Fixes `cf piece get` so filters and projections work on linked and Writable values and return compact, schema-shaped results without alias widening. Projections now build values from source-schema reads, sanitize link handles (keep only stream handles), and preserve nulls and nested array shapes.

- **Bug Fixes**
  - Dereference `asCell` links and sanitize schemas with `sanitizeSchemaForLinks(KeepAsCell.OnlyStream)` before predicates/projections; materialize non-stream Cells.
  - Support linked array slots, schema-less linked objects, Writable arrays, and nested Writable fields; preserve nullable leaves and nested `items` arrays. Concise dotted paths don’t traverse array items—use JSON Schema `items`.
  - Construct projected values inside computed nodes and re-assert the output schema at the read boundary to prevent widening to broader linked targets.
  - Docs: clarify filter-then-project and array traversal; add Topics examples for exact/range `--filter` with concise `--schema`; note that concise projections over nullable array items can collapse to JSON `null` and should be guarded by filtering or explicit item unions.

<sup>Written for commit 355a200194a213801c7f37a2fbe173edf389b36d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5272?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/plans/pattern-verb-contract.md
+++ b/docs/plans/pattern-verb-contract.md
@@ -365,12 +365,13 @@ board topic). Until that lands, board-level routing
 the target shape.
 
 `cf piece get` lets an agent control how much data an exploratory read returns:
-`--filter` narrows array membership and `--schema` projects fields through the
-runtime's `asSchema` read path. Both execute through computed pattern nodes so
-their CFC behavior matches pattern expressions. A limit on the number of
-records returned from a large array remains deferred, blocking nothing. It is
-adjacent CLI work for when board scale demands it, not an ask on any workstream
-in this implementation plan.
+`--filter` narrows array membership and `--schema` constructs a projected value
+from source-schema-selected reads. Both execute through computed pattern nodes
+so their CFC behavior matches pattern expressions, without returning an
+identity alias that can widen to a linked target's schema. A limit on the
+number of records returned from a large array remains deferred, blocking
+nothing. It is adjacent CLI work for when board scale demands it, not an ask on
+any workstream in this implementation plan.
 
 **A set of verbs wants to be a structural interface, and the machinery for that
 already exists.** Schemas are the type system, so a verb set is a schema
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -151,7 +151,11 @@ match the stored value is omitted by the runtime rather than reported as an
 error; prefer `true` leaves unless that type filtering is intentional. Schema
 combinators and references are rejected. Concise dotted paths traverse object
 properties, not nested array items: use an inline/file JSON Schema with `items`
-for a shape such as selected fields from every `comments` entry.
+for a shape such as selected fields from every `comments` entry. A concise
+projection over nullable array items can currently return JSON `null` for the
+whole array when any item is null. Filter null/unavailable items before
+projecting, or use an explicit array schema whose `items` preserves the source
+nullability; do not interpret that JSON `null` as an empty result.
 
 Both transforms run as a short-lived computed pattern in the caller's session.
 The runtime's list filter/map builtins therefore handle CFC exactly as authored
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -149,15 +149,20 @@ Projection schemas support structural `properties`, `items`, and scalar leaf
 schemas. In an array-item projection, a scalar leaf whose declared type does not
 match the stored value is omitted by the runtime rather than reported as an
 error; prefer `true` leaves unless that type filtering is intentional. Schema
-combinators and references are rejected.
+combinators and references are rejected. Concise dotted paths traverse object
+properties, not nested array items: use an inline/file JSON Schema with `items`
+for a shape such as selected fields from every `comments` entry.
 
 Both transforms run as a short-lived computed pattern in the caller's session.
 The runtime's list filter/map builtins therefore handle CFC exactly as authored
 pattern expressions do: predicate observations label array membership,
 projection reads propagate labels, and filtered elements retain their source
-links. The source cell's schema remains authoritative for Common Fabric
-metadata. A caller cannot introduce or override `ifc`, `asCell`, `scope`, or
-`default` through `--schema`.
+links. Projection map/lift nodes construct the requested shape from a
+source-schema-selected read rather than returning a widening identity alias.
+Nested non-stream Cell handles are materialized before the predicate/projection
+JavaScript runs; stream handles remain capabilities. The source cell's schema
+remains authoritative for Common Fabric metadata. A caller cannot introduce or
+override `ifc`, `asCell`, `scope`, or `default` through `--schema`.
 
 ## Built Binary
 
--- a/packages/cli/lib/piece-get-transform.ts
+++ b/packages/cli/lib/piece-get-transform.ts
@@ -7,6 +7,7 @@ import {
   KeepAsCell,
   type MemorySpace,
   type Runtime,
+  sanitizeSchemaForLinks,
 } from "@commonfabric/runner";
 import { isRecord } from "@commonfabric/utils/types";
 import { runtimeErrorLog } from "./callable.ts";
@@ -771,7 +772,8 @@ function projectionMask(schema: JSONSchema): ProjectionMask {
 }
 
 function projectValue(value: unknown, schema: JSONSchema): unknown {
-  if (schema === true || schema === false || value === null) return value;
+  if (typeof schema === "boolean") return schema ? value : undefined;
+  if (value === null) return value;
   if (Array.isArray(value)) {
     const itemSchema = schema.items ?? true;
     return value.map((item) => projectValue(item, itemSchema));
@@ -954,7 +956,7 @@ function resolveProjection(
 }
 
 export interface DerivePieceGetDependencies {
-  onResultCell?: (cell: Cell<unknown>) => void;
+  onOutputCell?: (cell: Cell<unknown>) => void;
 }
 
 /**
@@ -963,10 +965,11 @@ export interface DerivePieceGetDependencies {
  * The filter uses the runner's list builtin, so predicate observations taint
  * collection membership exactly as they do in authored patterns. Array
  * projection uses the map builtin for pointwise labels. Object/scalar
- * projection uses a lift. Projection nodes are identities over runtime
- * `asSchema` reads: the source-derived read schema performs the structural
- * projection while preserving authoritative source CFC metadata. Caller
- * schemas describe output shape only and cannot replace source metadata.
+ * projection uses a lift. Projection nodes construct the caller-requested
+ * shape from source-schema-selected reads, preventing an identity alias from
+ * widening back to a broader linked target. Caller schemas describe output
+ * shape only; source schemas remain authoritative for CFC and other Fabric
+ * metadata.
  */
 export async function derivePieceGetValue(
   runtime: Runtime,
@@ -975,10 +978,6 @@ export async function derivePieceGetValue(
   transform: PieceGetTransform,
   deps: DerivePieceGetDependencies = {},
 ): Promise<unknown> {
-  // Match ordinary piece path reads: a selected slot can itself be a link to
-  // the value being read (for example TopicsInput.topics). The transform must
-  // execute over that target, not over the Cell wrapper stored in the slot.
-  sourceCell = sourceCell.resolveAsCell();
   const declaredSourceSchema = sourceCell.schema;
   const sourceSchema = isRecord(declaredSourceSchema) &&
       declaredSourceSchema.asCell !== undefined
@@ -1053,7 +1052,10 @@ export async function derivePieceGetValue(
     const argumentSchema: JSONSchema = {
       type: "object",
       properties: {
-        element: dereferencedElementSchema(elementSchema),
+        element: sanitizeSchemaForLinks(
+          dereferencedElementSchema(elementSchema),
+          KeepAsCell.OnlyStream,
+        ),
         params: paramsSchema,
       },
       required: ["element", "params"],
@@ -1085,7 +1087,10 @@ export async function derivePieceGetValue(
     const argumentSchema: JSONSchema = {
       type: "object",
       properties: {
-        element: dereferencedElementSchema(elementSchema),
+        element: sanitizeSchemaForLinks(
+          dereferencedElementSchema(elementSchema),
+          KeepAsCell.OnlyStream,
+        ),
       },
       required: ["element"],
       additionalProperties: false,
@@ -1105,7 +1110,10 @@ export async function derivePieceGetValue(
   const directProjectionArgumentSchema: JSONSchema = {
     type: "object",
     properties: {
-      value: sourceReadSchema,
+      value: sanitizeSchemaForLinks(
+        sourceReadSchema,
+        KeepAsCell.OnlyStream,
+      ),
     },
     required: ["value"],
     additionalProperties: false,
@@ -1204,7 +1212,7 @@ export async function derivePieceGetValue(
         `Could not apply piece get transform: ${recorded.message}`,
       );
     }
-    deps.onResultCell?.(result as Cell<unknown>);
+    deps.onOutputCell?.(outputCell);
     return outputValue;
   } finally {
     runtime.runner.stop(resultCell);
--- a/packages/cli/lib/piece-get-transform.ts
+++ b/packages/cli/lib/piece-get-transform.ts
@@ -4,6 +4,7 @@ import {
   createBuilder,
   deepEqual,
   type JSONSchema,
+  KeepAsCell,
   type MemorySpace,
   type Runtime,
 } from "@commonfabric/runner";
@@ -769,6 +770,30 @@ function projectionMask(schema: JSONSchema): ProjectionMask {
   return true;
 }
 
+function projectValue(value: unknown, schema: JSONSchema): unknown {
+  if (schema === true || schema === false || value === null) return value;
+  if (Array.isArray(value)) {
+    const itemSchema = schema.items ?? true;
+    return value.map((item) => projectValue(item, itemSchema));
+  }
+  if (!isRecord(value)) return value;
+
+  const properties = schema.properties ?? {};
+  const projected: Record<string, unknown> = {};
+  if (schema.additionalProperties !== false) {
+    for (const [key, child] of Object.entries(value)) {
+      const childSchema = properties[key] ?? schema.additionalProperties ??
+        true;
+      projected[key] = projectValue(child, childSchema);
+    }
+    return projected;
+  }
+  for (const [key, childSchema] of Object.entries(properties)) {
+    if (key in value) projected[key] = projectValue(value[key], childSchema);
+  }
+  return projected;
+}
+
 function maskFromPaths(
   paths: Array<Array<string | number>>,
 ): PredicateMask {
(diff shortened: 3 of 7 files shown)


bfollington (commented):
_cf-review via Claude `claude-opus-5`, on behalf of @bfollington_

**Scope** · 4 files, +299/−21 (`packages/cli` transform + tests, topics README + skill) · **coverage**: deep-read of `piece-get-transform.ts` and the diff, delegated trace of `piece.input.get()` / `resolveAsCell` / `sanitizeSchemaForLinks` semantics · **checks run**: `deno test packages/cli/test/piece-get-transform.test.ts test/piece.test.ts` (92 steps, green), `deno fmt --check` + `deno lint` on touched files (clean), `deno task check-skill-facts` (OK), plus five differential probes run against both `81278ebd9` (base) and `bf841081f` (PR head).

The `resolveAsCell` + `asCell`-strip on the source is a genuine, well-motivated fix — `Writable<TopicPiece[]>` previously made `pull()` hand back a `Cell`, so `--filter` died with "can only be applied to an array". That part is right. The problem is `projectValue()`.

## 🔴 Blocking

### 1. `--schema` with a dotted path through an `asCell`/`Writable` property now returns `{}` — silent, and it's the documented headline shape

`packages/cli/lib/piece-get-transform.ts:773` (used at `:1094`, `:1116`)

`projectValue` is plain JS. The lift argument schemas at `:1056`, `:1088`, `:1108` are built by `selectSourceSchema`, which spreads `...metadata` — so a **nested** `asCell` survives (only the *top-level* one is stripped by `dereferencedElementSchema`). The lift therefore receives a `Cell` **handle** at that property; `projectValue` sees an object, does `"name" in cell` → false, and emits `{}`. Previously the identity lift returned the handle and the runtime's schema read materialised it correctly.

Verified. Source array items `{title, createdBy: <asCell {name, kind}>}`, projection `title,createdBy.name`:

```
base  81278ebd9 => [{"createdBy":{"name":"Sol"},"title":"T1"}]   ← correct, and `kind` correctly projected away
PR    bf841081f => [{"title":"T1","createdBy":{}}]               ← regression
```

This is exactly `packages/cli/README.md:128`'s own example (`--schema id,title,author.name`) and exactly the shape the new skill teaches (`createdBy.kind,createdBy.name`, `topic.title`, `author.kind,author.name`). Topics dodges it today only because `TopicAuthor` happens to be a plain object rather than a `Cell<>` — it will bite the first pattern with a `Writable`/`Cell<>` field, and `Writable ⇒ asCell` is exactly what this PR's own new test asserts.

**Verified fix** — sanitize the handles out of the lift argument schemas with the canonical helper rather than hand-rolling it, at all three sites:

```ts
element: sanitizeSchemaForLinks(
  dereferencedElementSchema(elementSchema),
  KeepAsCell.OnlyStream,
),
```

With that in place the probe returns `[{"title":"T1","createdBy":{"name":"Sol"}}]`, the full 27-step suite still passes, **and** it also fixes finding 6 below.

### 2. `projectValue` is not exercised by a single test

I reverted only the two call sites back to `({ element }) => element` / `({ value }) => value`, keeping every other change in the diff — **all 27 steps pass**, including the three new ones. The same is true of `resolveAsCell()` at `:981`: removing that line alone also leaves 27/27 green. The two riskiest lines in the changeset are the two nothing pins.

Given the "be right, don't guess the closest fit" bar: either add a test that fails without `projectValue` (the array case in finding 3 is the honest candidate) or drop it in favour of the `asSchema(outputSchema)` re-assertion at `:1186`, which is the change that actually addresses alias widening.

## 🟡 Improvements

### 3. Concise dotted paths don't traverse arrays, and this PR silently changes the failure mode from leak to data loss

`conciseProjectionSchema` (`:585`) only ever emits `{type:"object", properties, additionalProperties:false}` — never `items`. So `--schema title,comments.body` against `{title, comments: [{body, secret}]}`:

```
base => {"comments":[{"body":"hello","secret":"LEAK"}],"title":"T"}   ← leaks `secret`
PR   => {"title":"T"}                                                 ← key vanishes
```

The PR's behaviour is safer but strictly worse for an agent: it looks like *there are no comments*. Neither behaviour is documented anywhere. At minimum say so in `packages/cli/README.md`; the real fix is to make the concise builder emit `items` when the source schema at that path is an array. Verified.

### 4. Coherence — three live docs now describe the mechanism this PR replaced

The JSDoc you're editing under, `piece-get-transform.ts:960-970`, states: *"Projection nodes are identities over runtime `asSchema` reads: the source-derived read schema performs the structural projection."* That is now false. Same for `docs/plans/pattern-verb-contract.md:368` ("`--schema` projects fields through the runtime's `asSchema` read path") and the mechanism paragraph at `packages/cli/README.md:155-160`. Someone searching for how projection works lands on all three.

### 5. `resolveAsCell()` swaps schema authority, and diverges from the unflagged read

`resolveLink` only carries the source schema forward when the target link has none (`packages/runner/src/link-resolution.ts:391`) — otherwise **the stored link's baked schema wins** over the pattern's declared slot schema. Meanwhile `piece.input.get(path)` (`packages/piece/src/ops/piece-controller.ts:2269-2290`) deliberately does *not* route through `resolveAsCell()`: it walks `key()` so declared schemas narrow on the way down, and unwraps exactly one terminal `asCell` hop, with an explicit comment that this is to keep the #5231 scope cap. So `cf piece get topics --input` and `cf piece get topics --input --schema …` now read through different schema authorities.

I confirmed the disclosure consequence (a slot declaring `{title}, additionalProperties:false` over a target declaring `{title, secret}` projects `secret`) — **but that reproduces identically on base**, so it is pre-existing, not introduced here. Flagging it because this PR leans further into that path and it's a live generation-skew hazard for a board like Topics whose entries were minted across pattern generations. `sourceCell.asSchema(dereferencedElementSchema(schema))` unwraps the handle without discarding the declared schema — worth checking whether that covers your live case.

### 6. Pre-existing, same root cause, and the new skill walks right into it

`--filter` on a path through an `asCell` property silently matches nothing (`.createdBy.name == "Sol"` → `[]` on both base and PR — the predicate walks a `Cell` handle in plain JS). The skill now teaches `--filter '.author.name == "Sol" or .authorName == "Sol"'`. The finding-1 sanitize fix makes this match correctly; verified.

### 7. `deps.onResultCell?.(result)` (`:1207`) still hands back the pre-narrowing cell

Only tests consume it today, but it's the widening-prone handle — any future `--watch` would reintroduce exactly the bug `:1186` fixes.

## ⚪ Nits

- `sourceCell = sourceCell.resolveAsCell()` (`:981`) reassigns a parameter; a `const resolved` reads better next to `sourceValueCell`.
- `projectValue`'s `schema === false → return value` (`:774`) returns the value for a schema that admits nothing. Unreachable (`normalizeProjectionSchema` rejects `false`), but it's the wrong default if that ever changes.
- Skill troubleshooting: dropping "Retry once" is the right call per AGENTS.md, but *"unless external state changes"* isn't actionable for an agent — name the condition.

## Questions

- What live failure required `resolveAsCell()` on top of the `asCell` strip? The suite doesn't distinguish them, and the strip alone is the narrower change.
- Is the array-key drop in finding 3 intentional (safe-by-default), or did it go unnoticed?

## Possible follow-ups

- Make concise `--schema` array-aware — emit `items` when the source schema at that segment is an array. That is the last thing standing between these flags and "just works" for Topics.
- `--limit` for large boards; `docs/plans/pattern-verb-contract.md:370` already parks it.

---

**Verdict:** the source-resolution half is right and worth landing; `projectValue` as written trades a documented working case for an untested one, and should be either schema-sanitized (finding 1) or dropped before merge.

bfollington:
Addressed the full cf-review pass in b1ddf22cb:

- nested non-stream asCell/Writable fields are materialized with sanitizeSchemaForLinks(..., KeepAsCell.OnlyStream) at every plain-JS filter/projection lift boundary
- removed resolveAsCell() so transformed reads retain the declared slot schema authority and match ordinary piece reads
- projection nodes now construct the requested shape, with a differential non-aliasing test that fails if projectValue is replaced by identity
- the callback exposes the narrowed output cell, not the broader intermediate result
- added coverage for nested Writable filtering/projection, direct linked-object projection, non-aliasing, and explicit nested-array item schemas
- documented that concise dotted paths do not traverse arrays yet, and updated the CLI, CF skill, Topics skill, and live verb-contract docs to describe the actual mechanism
- made the Topics retry condition actionable

Validation after rebasing onto current main:
- 163 focused CLI test steps pass
- deno check passes for the transform and tests
- deno lint passes repo-wide
- deno task check-docs passes (521 blocks)
- deno task check-skill-facts passes
- touched-file formatting passes; repo-wide deno fmt --check remains blocked by 23 pre-existing unrelated files
- live Topics filter/schema smoke test passed with the supplied identity

The previous coverage warning was against the old head; the new nested-array test executes the remaining projectValue array branch. Follow-on ergonomics and pushdown work remains tracked separately in CT-1931 through CT-1938.

bfollington:
Coverage fix pushed in c66a82c8e. I downloaded and merged the failed run's workspace LCOV artifacts: the new debt was the null-preservation line in projectValue (the other uncovered line is existing baseline debt). The direct projection test now projects a nullable field whose value is null and asserts that null is retained; the focused 30-step transform suite passes.

bfollington:
Follow-up: the first nullable test used the concise schema form, whose leaf normalizes to boolean true and therefore returned before the object-schema/null line. Updated in 80d9aebed to project subtitle with an explicit { type: ["string", "null"] } property schema, which reaches the exact DA:776 branch reported by both failed CI LCOV artifacts. The focused 30-step transform suite remains green.

bfollington (commented):
_cf-review (round 2) via Claude `claude-opus-5`, on behalf of @bfollington_ — head `b1ddf22cb`

**Coverage**: deep-read of the delta commit, re-ran all five differential probes from round 1 against the new head, mutation-tested the new non-aliasing test, local coverage run to locate the CI gate failure · **checks**: `deno test packages/cli/test/{piece-get-transform,piece}.test.ts` (95 steps, green), `deno fmt --check` + `deno lint` on all 6 changed files (clean).

## Every round-1 item is genuinely fixed — verified, not taken on trust

| Round 1 | Status |
| --- | --- |
| 🔴 nested `asCell` projection → `{}` | **fixed** — `title,createdBy.name` through an `asCell` property now returns `{"name":"Sol"}` |
| 🔴 `projectValue` untested | **fixed** — I replaced both call sites with identity; `materializes a projection instead of aliasing its broader source` **fails**. Real differential test |
| 🟡 array-path gap | documented in README + cf skill, with an explicit-`items` test. Filed as CT-1934 |
| 🟡 stale JSDoc / verb-contract / README | all three corrected, and the new text matches the code |
| 🟡 `resolveAsCell()` schema authority | removed — and the motivating `Writable`/`asCell` root-array case still works (`filter .id==2` → `[{"title":"B"}]`), so the strip alone was sufficient. Filed as CT-1937 |
| 🟡 pre-existing filter-through-`asCell` → `[]` | **fixed as a bonus** by the same sanitize — `.createdBy.name == "Sol"` now matches |
| 🟡 `onResultCell` un-narrowed | renamed `onOutputCell`, hands back the narrowed cell, both CFC probes updated coherently |
| ⚪ `projectValue(false)` | fixed |

The `sanitizeSchemaForLinks(…, KeepAsCell.OnlyStream)` placement is right, and *"stream handles remain capabilities"* in the README is an accurate description of what that flag actually does. The Linear set (CT-1931…1938) captures the deferred work well — CT-1934 and CT-1937 are faithful write-ups of the two I raised.

## 🔴 CI is red, and the comment's explanation doesn't hold

The **Coverage Check** run against `b1ddf22cb` itself (not the old head) still fails `packages/cli 3634 → 3635, +1`, and **Status** is red because of it. The uncovered line is `packages/cli/lib/piece-get-transform.ts:776` — `if (value === null) return value;` — not the array branch.

It is reachable, and pinning it is worth a test rather than a baseline bump, because it pins a real asymmetry:

```
--schema '{"type":"object","additionalProperties":true,
           "properties":{"createdBy":{"type":["object","null"],
                         "properties":{"name":true},"additionalProperties":false}}}'
  => {"title":"T","createdBy":null}      ← null survives  (hits :776)

--schema '{"type":"object","additionalProperties":false, …same createdBy… }'
  => {"title":"T"}                        ← createdBy silently dropped
```

Same data, same declared intent for `createdBy`; whether `null` survives depends on the *sibling* `additionalProperties` setting. Worth a test on both halves.

## 🟡 New, and I think it's the deterministic cause of CT-1932

**A single `null` element voids the entire array under a concise `--schema`.** Verified against `b1ddf22cb`, one source array `[{name:"a",secret:"s"}, null]` declared `items: {type:["object","null"], …}`:

```
plain read (no flags)                       => [{"name":"a","secret":"s"},null]
--filter '.name == "a"'                     => [{"name":"a","secret":"s"}]
--schema name                     (concise) => undefined
--schema '{…items:{"type":["object","null"],"properties":{"name":true}}}'
                                            => [{"name":"a"},null]
```

`conciseProjectionSchema` emits a bare `{type:"object"}` per node; the output re-assertion `outputCell.asSchema(outputSchema)` then voids any element that isn't an object — and one bad element takes the whole array with it, including the good rows. `safeStringify(undefined)` returns `"null"` (`packages/cli/lib/render.ts:37`), so **this reaches the agent as JSON `null` on stdout.**

That is exactly CT-1932's reported symptom — *"`comments --input --filter … --schema …` returned JSON `null`"* while a scalar `title --input` loaded fine — and exactly this PR body's own *"the live board can still return null when a linked entry is unavailable"*. CT-1932 frames it as a nondeterministic replica/hydration disagreement; at least part of it is this deterministic schema-narrowing path, and it reproduces offline in the emulated storage manager with no replica involved. Worth attaching to CT-1932 before someone goes hunting in the storage layer.

The consequence for the workflow this PR ships: the Topics skill tells an agent that a compact `topics --input --schema …` result "remains valid evidence", but under this mechanism one unavailable linked entry turns a populated board into `null` — indistinguishable from "no matches". A per-element grace (skip the element, keep the array) or preserving the source's type union in the projection would both fix it; either is CT-1934/CT-1932 territory, not this PR.

Pre-existing from #5237 — not introduced here, and not a reason to hold the PR.

---

**Verdict:** the code is in good shape and I'd land it once the coverage gate is green — add the `:776` test rather than an `ACCEPT_COVERAGE_DEBT` line, since the behavior it pins is genuinely underspecified today.

bfollington:
_cf-review handoff via Claude `claude-opus-5`, on behalf of @bfollington_ — against `b1ddf22cb`

Follow-up to my round-2 review with the two items in paste-ready form, so whoever picks this up doesn't have to re-derive them.

## 1. The coverage gate — verified fix

**The failing run is against `b1ddf22cb` itself**, not the old head: `gh run view 30777063499 --json headSha` → `b1ddf22cb`, Coverage Check `packages/cli 3634 → 3635 +1`, and `Status` is red solely because of it.

The uncovered line is `packages/cli/lib/piece-get-transform.ts:776` — `if (value === null) return value;` — **not** the array branch the nested-array test was expected to cover. Confirmed by local lcov:

```bash
cd packages/cli
deno test --allow-all --coverage=/tmp/cov test/piece-get-transform.test.ts test/piece.test.ts
# DA:776,0
```

Reaching `:776` needs a `null` that survives the source read, which requires the mask to stay `true` at that node — so a whitelist (`additionalProperties: false`) parent can't do it, but `additionalProperties: true` can. This test passes, is `deno fmt` clean, and takes the file's remaining uncovered set to `[]` (`:831` is pre-existing in `mergeMasks`):

```diff
--- a/packages/cli/test/piece-get-transform.test.ts
+++ b/packages/cli/test/piece-get-transform.test.ts
@@ -754,6 +754,44 @@ describe("cf piece get transforms", () => {
     ).toEqual({ comments: [{ body: "Hello" }] });
   });
 
+  it("keeps a null value that the projection schema admits", async () => {
+    const tx = runtime.edit();
+    const source = runtime.getCell(
+      space,
+      "nullable-projection-source",
+      {
+        type: "object",
+        properties: {
+          title: { type: "string" },
+          createdBy: {
+            type: ["object", "null"],
+            properties: { name: { type: "string" } },
+          },
+        },
+      },
+      tx,
+    );
+    source.set({ title: "T", createdBy: null as never });
+    expect((await tx.commit()).ok).toBeDefined();
+
+    const projection = {
+      type: "object",
+      additionalProperties: true,
+      properties: {
+        createdBy: {
+          type: ["object", "null"],
+          properties: { name: true },
+          additionalProperties: false,
+        },
+      },
+    };
+    expect(
+      await derivePieceGetValue(runtime, space, source, {
+        projection: await parsePieceGetProjection(JSON.stringify(projection)),
+      }),
+    ).toEqual({ title: "T", createdBy: null });
+  });
+
   it("handles schema-less filters and identity projection schemas", async () => {
```

Please add the test rather than `ACCEPT_COVERAGE_DEBT` — the line guards behavior that is currently underspecified, per the asymmetry below.

**Worth knowing while you're in there** (deliberately *not* encoded as an assertion, since pinning it would enshrine what looks like a bug): flip that projection's root to `additionalProperties: false` and keep everything else identical, and `createdBy` is dropped instead of returned as `null`. Whether a null survives depends on the *sibling* `additionalProperties` setting, not on the schema declared for the field itself. Probably belongs on CT-1934.

## 2. CT-1932 — deterministic repro, no replica involved

CT-1932 records `comments --input --filter … --schema …` returning JSON `null` while a scalar `title --input` loaded fine, and frames it as a fresh-runtime/hydration disagreement. **At least part of it is deterministic and reproduces offline** against `StorageManager.emulate` — no server, no replica, no timing.

A single `null` element voids the *entire* array under a concise `--schema`, taking the good rows with it:

```ts
// source: [{ name: "a", secret: "s" }, null]
// declared items: { type: ["object", "null"], properties: { name, secret } }

plain read (no flags)                       => [{"name":"a","secret":"s"},null]
--filter '.name == "a"'                     => [{"name":"a","secret":"s"}]
--schema name                    (concise)  => undefined          ← whole array gone
--schema '{"type":"array","items":{"type":["object","null"],
           "properties":{"name":true},"additionalProperties":false}}'
                                            => [{"name":"a"},null]  ← correct
```

Mechanism: `conciseProjectionSchema` (`piece-get-transform.ts:585`) emits a bare `{type:"object"}` for every node, and `selectSourceSchema` (`:903`) likewise hardcodes `type: "object"`, both discarding a source union such as `["object","null"]`. The output re-assertion `outputCell.asSchema(outputSchema)` then rejects any element that isn't an object — and array validation voids the whole array, not the offending element.

It reaches the operator as JSON `null`, because `safeStringify(undefined)` returns `"null"` (`packages/cli/lib/render.ts:37`) — which is exactly the observed symptom, and exactly this PR's own note that *"the live board can still return null when a linked entry is unavailable"*.

I tried narrowing this in `selectSourceSchema` alone (preserving `source.type` when it is a union) and it is **not** sufficient — the output-schema read voids it again — so the fix has to reach `conciseProjectionSchema` or the output re-assertion. That is CT-1932/CT-1934 work, not this PR.

Why it matters for the workflow being shipped here: `skills/topics/SKILL.md` tells an agent the compact `topics --input --schema …` search "remains valid evidence", but under this mechanism one unavailable linked entry turns a populated board into `null`, indistinguishable from "no matches" — while the skill's existing guardrail only covers the `crossrefs --step` case. Worth a line in the skill's troubleshooting section even before the underlying fix lands.

Everything else in the delta commit verified clean in round 2; this is the only thing standing between the PR and a green board.

bfollington:
Final follow-up from issuecomment-5161576910 is addressed in 355a20019:

- CT-1932 now records the deterministic offline mixed object/null array reproduction, corrected mechanism, and acceptance criteria; the old replica/hydration hypothesis is explicitly separated
- CT-1934 now covers array traversal plus source-union/nullability preservation and `additionalProperties` independence
- the two Linear issues are cross-linked
- CLI and CF guidance now warn that a nullable item can collapse a concise projection to JSON `null`
- the Topics discovery query defensively filters `.title != null` before projection, and troubleshooting no longer treats a null compact read as empty/no matches

Validation: touched-file formatting, repo-wide lint, docs checking, skill facts, and both skill validators pass. Repo-wide local formatting still reports the same 23 unrelated baseline files.

PR #5276, opened 2026-08-03
fix(cli): make concise projections array-aware

bfollington:
## Summary

- make concise `cf piece get --schema` paths traverse declared nested and repeated arrays
- preserve source-declared nullable items and properties through projection
- distinguish unavailable transformed values from valid JSON `null`
- update CLI, CF, and Topics guidance for the corrected behavior

## Linear

- [CT-1932](https://linear.app/common-tools/issue/CT-1932/cf-piece-get-preserve-nullable-linked-array-entries-through)
- [CT-1934](https://linear.app/common-tools/issue/CT-1934/cf-piece-get-make-concise-schema-paths-array-and-union-aware)

## Verification

- `deno test -A packages/cli/test/piece-get-transform.test.ts` — 39 steps passed
- `deno test -A packages/cli/test/piece.test.ts` — 66 steps passed
- focused coverage — no uncovered executable lines in either changed implementation file
- `deno check packages/cli/mod.ts`
- `deno lint`
- touched-file `deno fmt --check`
- `deno task check-docs`
- `deno task check-skill-facts`
- CF and Topics skill validation

The full CLI package suite passed 1,650 tests and failed two unrelated existing color/TTY assertions in `color-mode.test.ts`. Repository-wide `deno fmt --check` also reports 23 unrelated pre-existing files; every file changed by this PR passes formatting.

---
## Summary by cubic
Make concise `cf piece get --schema` projections array- and union-aware, preserve source-declared nullability, and clearly fail when a transformed value can’t materialize (not JSON null). Addresses CT-1932 and CT-1934.

- New Features
  - Concise dotted paths follow declared arrays across nested and repeated arrays (e.g., `comments.body`), aligning through `$ref`, `anyOf`, `oneOf`, and `allOf`. When a nested container isn’t identified by the source schema, the same field mask is applied across encountered arrays to prevent sibling leakage; use explicit JSON Schema when you need a fixed output shape.
  - Union-aware selection: respects `type` arrays plus `anyOf`/`oneOf`/`allOf` during projection; keeps read-time compositions intact and derives existential branches only for the materialized output schema.

- Bug Fixes
  - Preserve nullable items and properties through projection, including linked arrays (CT-1932).
  - Distinguish an unavailable/rejected transformed value from JSON `null`; command exits nonzero with a clear message, while absent optional sources and valid projected nulls remain successful.

<sup>Written for commit dd86e4e70ff88bdb7f13f7575d3bbb6b8ad02d29. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5276?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cli/lib/piece-get-transform.ts
+++ b/packages/cli/lib/piece-get-transform.ts
@@ -794,32 +794,40 @@ function schemaFromProjectionMask(mask: ProjectionMask): JSONSchema {
 /** @internal Exported for focused schema-shape tests. */
 export function schemaMayBeArray(
   schema: JSONSchema | undefined,
-  seen = new Set<object>(),
+  root: JSONSchema = schema ?? true,
+  ancestors = new Set<object>(),
 ): boolean {
   if (!isRecord(schema)) return false;
-  if (seen.has(schema)) return false;
-  seen.add(schema);
-  if (schema.$ref !== undefined) {
-    try {
-      return schemaMayBeArray(
-        ContextualFlowControl.resolveSchemaRefsOrThrow(schema, schema),
-        seen,
-      );
-    } catch (error) {
-      throw new PieceGetTransformError(
-        `Could not resolve source schema reference for --schema: ${
-          error instanceof Error ? error.message : String(error)
-        }`,
-        { cause: error },
-      );
+  if (ancestors.has(schema)) return false;
+  ancestors.add(schema);
+  const documentRoot = schema.$defs !== undefined ? schema : root;
+  try {
+    if (schema.$ref !== undefined) {
+      let resolved: JSONSchema;
+      try {
+        resolved = ContextualFlowControl.resolveSchemaRefsOrThrow(
+          schema,
+          documentRoot,
+        );
+      } catch (error) {
+        throw new PieceGetTransformError(
+          `Could not resolve source schema reference for --schema: ${
+            error instanceof Error ? error.message : String(error)
+          }`,
+          { cause: error },
+        );
+      }
+      return schemaMayBeArray(resolved, documentRoot, ancestors);
     }
+    if (schemaTypes(schema).includes("array")) return true;
+    return [
+      ...schema.anyOf ?? [],
+      ...schema.oneOf ?? [],
+      ...schema.allOf ?? [],
+    ].some((option) => schemaMayBeArray(option, documentRoot, ancestors));
+  } finally {
+    ancestors.delete(schema);
   }
-  if (schemaTypes(schema).includes("array")) return true;
-  return [
-    ...schema.anyOf ?? [],
-    ...schema.oneOf ?? [],
-    ...schema.allOf ?? [],
-  ].some((option) => schemaMayBeArray(option, seen));
 }
 
 function alignConciseProjectionMask(
@@ -831,10 +839,9 @@ function alignConciseProjectionMask(
 
   if (schemaMayBeArray(source)) {
     const sourceItem = schemaAtArrayItem(cfc, source);
-    const itemMask = mask.type === "array" ? mask.items : mask;
     return {
       type: "array",
-      items: alignConciseProjectionMask(cfc, sourceItem, itemMask),
+      items: alignConciseProjectionMask(cfc, sourceItem, mask),
     };
   }
 
@@ -872,6 +879,9 @@ function projectValue(
     );
   }
   if (!isRecord(value)) return value;
+  if (implicitArrayTraversal && schema.items !== undefined) {
+    return projectValue(value, schema.items, implicitArrayTraversal);
+  }
 
   const properties = schema.properties ?? {};
   const projected: Record<string, unknown> = {};
@@ -927,12 +937,18 @@ function maskFromPaths(
   return paths.length === 0 ? true : build(paths);
 }
 
-function mergeMasks(
+/** @internal Exported for focused mask-composition tests. */
+export function mergeMasks(
   left: PredicateMask,
   right: ProjectionMask,
 ): ProjectionMask {
   if (left === true || right === true) return true;
-  if (right.type !== "object") return true;
+  if (right.type === "array") {
+    return {
+      type: "array",
+      items: mergeMasks(left, right.items),
+    };
+  }
   const properties: Record<string, ProjectionMask> = {};
   for (
     const key of new Set([
@@ -951,10 +967,12 @@ function mergeMasks(
   return { type: "object", properties, additionalProperties: false };
 }
 
-function selectSourceSchema(
+/** @internal Exported for focused source-schema selection tests. */
+export function selectSourceSchema(
   cfc: ContextualFlowControl,
   source: JSONSchema | undefined,
   mask: ProjectionMask,
+  purpose: "source-read" | "projected-output" = "source-read",
 ): JSONSchema {
   // An absent/wildcard source schema cannot prove that a structural mask has
   // the same container shape as the current value. Keep that read permissive;
@@ -966,24 +984,36 @@ function selectSourceSchema(
   if (source.$ref !== undefined) {
     return source;
   }
-  if (source.anyOf !== undefined || source.oneOf !== undefined) {
+  if (
+    source.anyOf !== undefined || source.oneOf !== undefined ||
+    source.allOf !== undefined
+  ) {
     const {
-      anyOf: _anyOf,
-      oneOf: _oneOf,
+      anyOf,
+      oneOf,
+      allOf,
       ...metadata
     } = source;
-    const options = [...source.anyOf ?? [], ...source.oneOf ?? []].map(
-      (option) => selectSourceSchema(cfc, option, mask),
-    );
-    return { ...metadata, anyOf: options };
-  }
-  if (source.allOf !== undefined) {
-    const { allOf: _allOf, ...metadata } = source;
+    if (purpose === "source-read") {
+      // Projection can make formerly exclusive branches overlap or discard a
+      // discriminator altogether. Keep the declared composition intact at the
+      // read boundary; only the materialized output schema uses projected
+      // (existential) branches.
+      return source;
+    }
+    const projectOptions = (options: readonly JSONSchema[] | undefined) =>
+      options?.map((option) => selectSourceSchema(cfc, option, mask, purpose));
+    const projectedAnyOf = projectOptions(anyOf);
+    const projectedOneOf = projectOptions(oneOf);
+    const projectedAllOf = projectOptions(allOf);
+    const conjunctions = [
+      ...projectedAllOf ?? [],
+      ...(projectedOneOf === undefined ? [] : [{ anyOf: projectedOneOf }]),
+    ];
     return {
       ...metadata,
-      allOf: source.allOf.map((option) =>
-        selectSourceSchema(cfc, option, mask)
-      ),
+      ...(projectedAnyOf === undefined ? {} : { anyOf: projectedAnyOf }),
+      ...(conjunctions.length === 0 ? {} : { allOf: conjunctions }),
     };
   }
 
@@ -993,7 +1023,7 @@ function selectSourceSchema(
     const { items: _items, prefixItems: _prefixItems, ...metadata } = source;
     return {
       ...metadata,
-      items: selectSourceSchema(cfc, sourceItem, mask.items),
+      items: selectSourceSchema(cfc, sourceItem, mask.items, purpose),
     };
   }
 
@@ -1017,6 +1047,7 @@ function selectSourceSchema(
       cfc,
       child === false ? undefined : child,
       childMask,
+      purpose,
     );
   }
   const selectedRequired = required?.filter((key) => key in properties);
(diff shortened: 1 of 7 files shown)


packages/cli/lib/piece-get-transform.ts
```diff
@@ -771,12 +771,105 @@ function projectionMask(schema: JSONSchema): ProjectionMask {
...
+export function schemaMayBeArray(
+  schema: JSONSchema | undefined,
+  seen = new Set<object>(),
+): boolean {
+  if (!isRecord(schema)) return false;
+  if (seen.has(schema)) return false;
+  seen.add(schema);
+  if (schema.$ref !== undefined) {
+    try {
+      return schemaMayBeArray(
```
bfollington: **🔴 Blocking — `schemaMayBeArray` hard-errors on any `$ref` whose `$defs` live at the source-schema root (verified regression).**

`resolveSchemaRefsOrThrow(schema, schema)` passes the *node itself* as the document root. `cfc.schemaAtPath` does attach the referenced `$defs` to the subschema it returns, so the **first** hop resolves. But `resolveCfcSchemaRefs` returns the bare target without re-attaching `$defs`, so any `$ref` **inside** that target — in an `anyOf`/`oneOf`/`allOf` member, which is exactly where this function recurses — resolves against a node with no `$defs` and throws. The throw is caught and rethrown as `PieceGetTransformError`, so the whole command fails.

Root-level `$defs` + nested `$ref` is the standard emission shape (`sanitizeSchemaForLinks` hoists to root; the schema-generator emits `anyOf: [{$ref}, {$ref}]` for a union of named types). The new test `"projects concise array paths through source schema references"` only passes because it places `$defs` *inline next to* the `$ref`, which is not how these schemas are normally shaped.

Verified with source schema `{properties:{node:{$ref:"#/$defs/Node"}}, $defs:{Node:{anyOf:[{$ref:"#/$defs/Leaf"},{$ref:"#/$defs/Branch"}]}, Leaf:{...}, Branch:{...}}}` and projection `node.body`:

```
main: {"node":{"body":"Visible"}}
PR:   PieceGetTransformError: Could not resolve source schema reference for
      --schema: Failed to resolve $ref: #/$defs/Leaf.
```

**Fix**: thread the document root through the recursion (`schemaMayBeArray(schema, root, seen)`, resolving against `root` and re-rooting on any node that carries its own `$defs`) — or, since this is a best-effort "may it be an array" probe, treat an unresolvable `$ref` as *not provably an array* and return `false` instead of throwing. Turning a source-schema shape the CLI cannot introspect into a fatal error is the wrong failure mode either way.

Minor, same function: `seen` is shared across the `.some()` branches, so a node visited (and answered `false`) inside an in-progress cycle poisons a later independent visit. Order-dependent false negative; worth fixing while you are in here.

packages/cli/lib/piece-get-transform.ts
```diff
@@ -771,12 +771,105 @@ function projectionMask(schema: JSONSchema): ProjectionMask {
...
+}
+
+function alignConciseProjectionMask(
+  cfc: ContextualFlowControl,
+  source: JSONSchema | undefined,
+  mask: ProjectionMask,
+): ProjectionMask {
+  if (mask === true || source === undefined || source === true) return mask;
+
+  if (schemaMayBeArray(source)) {
```
bfollington, line 840: **🔴 Blocking — sibling leak when the source *may* be an array but the value is an object (verified regression).**

This wraps the mask in `{type:"array", items: ...}` whenever `schemaMayBeArray(source)` — including for a union like `anyOf:[{type:"array",...},{type:"object",...}]`. `projectValue` handles the "declared object, actual array" direction via `implicitArrayTraversal`, but not the reverse: when the value is a **record** and the projection schema is array-shaped, `schema.properties` is `undefined` and `schema.additionalProperties` is `undefined`, so it falls into the permissive branch at `piece-get-transform.ts:878` and copies **every key of the object through unfiltered**.

Verified with source `{properties:{entry:{anyOf:[{type:"array",items:{...}},{type:"object",properties:{body,privateNote}}]}}}`, value `{entry:{body:"Visible",privateNote:"hidden"}}`, projection `entry.body`:

```
main: {"entry":{"body":"Visible"}}
PR:   {"entry":{"body":"Visible","privateNote":"hidden"}}   <- privateNote leaked
```

This is the exact failure mode the PR advertises as prevented — `packages/cli/README.md:157` and `skills/cf/SKILL.md:242` both state siblings "still cannot leak" / "cannot leak". Those lines are actively wrong as written.

**Fix**: make the traversal symmetric in `projectValue` — when `implicitArrayTraversal` is on, `schema.items !== undefined`, and the value is a non-array record, descend with `schema.items` instead of falling through to the object branch. A test with an `array | object` union source (and one with `type: ["array","object"]`) belongs alongside the "ambiguous nested array" case.

bfollington (commented):

**Scope** · 7 files, 2 impl files in `packages/cli` (deep-read) · **Coverage**: deep-read of both implementation files + tests + docs, no delegation · **Checks run**: `deno test -A packages/cli/test/piece-get-transform.test.ts` (39 steps, green), plus four purpose-built probes run against **both** this branch and `origin/main` to A/B the behavior change.

The change is coherent in intent: concise `--schema` paths now consult the source schema, insert array levels where the source declares one, and derive the output schema from the source rather than from the mask (which is what preserves nullability). Two of the new mechanisms are wrong in ways the added tests don't reach — both left inline.

## 🔴 Blocking (inline)

1. **`schemaMayBeArray` hard-errors on any `$ref` whose `$defs` live at the source-schema root** — verified regression. `main` returns the projection; this branch throws `Could not resolve source schema reference for --schema`.
2. **Sibling leak when the source *may* be an array but the value is an object** — verified regression, and `packages/cli/README.md:157` / `skills/cf/SKILL.md:242` currently assert the opposite.

## 🟡 Improvements

**`anyOf` and `oneOf` are collapsed into a single `anyOf`.** `piece-get-transform.ts:969` — `[...source.anyOf ?? [], ...source.oneOf ?? []]` is re-emitted as `anyOf`. For a source carrying both, `(A|B) ∧ exactly-one(C,D)` becomes `A|B|C|D`: the conjunction and the exclusivity both vanish. It only ever loosens the read schema, so it isn't an output leak, but it's a silent semantic change. Keep the two keywords separate, or narrow each and re-emit both. *Suspected — I did not find a source schema in-tree carrying both.*

**`mergeMasks` discards narrowing whenever the aligned mask is array-shaped.** `piece-get-transform.ts:930` — `if (right.type !== "object") return true`. With `--filter` plus a concise path that align turned into an array level (`comments.body`), the whole `comments` subtree reverts to a full read. Output is still correctly projected (the lift applies the mask), so this is over-read / CFC taint, not disclosure — but it contradicts the "construct projected values from source-schema-selected reads" contract in the `derivePieceGetValue` doc comment. Teaching `mergeMasks` to descend into array masks would restore it.

**Unset source values now exit nonzero.** `packages/cli/lib/piece.ts:2240` — verified: a projection over a cell that was never written returns `undefined` from the transform layer and now raises `PieceGetTransformError`. That is the stated intent for *unavailable* values, but it also catches the ordinary "this optional `--input` field isn't set yet" read, which previously exited 0. See the question below.

## ⚪ Nits

- `?? projectionSchema` in `resolveProjection` (`:1067`) is unreachable — `sanitizeSchemaForLinks` only returns `undefined` for `undefined` input, and `selectSourceSchema` returns `JSONSchema`, never `undefined`.
- `mask.type === "array" ? mask.items : mask` in `alignConciseProjectionMask` (`:834`) — concise masks are never array-shaped on entry, and the only array levels are the ones align itself inserts on the way down. Dead branch, or a missing comment explaining what it defends against.

## Question for the author

The unset-cell case: is exiting nonzero intended for a source value that is simply *absent* rather than unavailable-or-rejected? "Absent optional field" and "the expression failed to materialize" collapse into one nonzero exit with a message that suggests `--step`, which won't help the first case. If the distinction matters for scripting, `derivePieceGetValue` could distinguish "source was undefined" from "projection produced undefined".

## Verdict

**Not mergeable as-is.** The direction is right and the CT-1932 nullability work looks sound, but the two blocking findings are behavior regressions against `main` confirmed by A/B probe — one turns ordinary `$ref`-bearing source schemas into a hard CLI error, the other leaks the sibling fields the change exists to hide, while the docs assert it cannot happen.

packages/cli/lib/piece-get-transform.ts
bfollington: Fixed in dd86e4e70. schemaMayBeArray now threads the document root through nested references, re-roots when a node carries its own $defs, and uses recursion-stack ancestry so completed branches do not poison later visits. The regression test now uses root-level $defs with nested refs in the normal generated shape.

packages/cli/lib/piece-get-transform.ts
bfollington: Fixed in dd86e4e70. Implicit concise traversal is now symmetric: when the aligned projection is array-shaped but the runtime value is an object, projection descends through items before materializing the object. Regression coverage exercises both anyOf array-or-object and type array-or-object schemas and confirms private siblings are dropped.

bfollington:
Addressed the full review in dd86e4e70: root-scoped nested refs; symmetric array/object projection; separate source-read and projected-output composition semantics; array-aware filter/projection mask merging; absent optional source versus failed-transform behavior; and both dead-code nits. Focused suites pass 108 steps, deno check/lint/docs/skill checks pass, and all touched files are formatted. The full CLI suite remains at 1,650 passing with the same two unrelated color-mode assertions failing.

bfollington:
cf-review via Claude Opus 5, on behalf of @bfollington — re-review of `dd86e4e70`

**Coverage**: deep-read of the follow-up commit + re-ran the A/B probes from the first pass against the new head · **Checks run**: `deno test -A packages/cli/test/{piece-get-transform,piece}.test.ts` (108 steps, green), `deno check packages/cli/mod.ts`, `deno lint`, `deno fmt --check` on all touched files — all clean.

**Both blocking findings are fixed, verified by re-probe:**

| Probe | before | after |
|---|---|---|
| root `$defs` + `anyOf`-of-`$ref` on the path | `PieceGetTransformError: Failed to resolve $ref` | `{"node":{"body":"Visible"}}` |
| root `$defs`, `$ref` → array through a union | — | `{"node":[{"body":"Visible"}]}` |
| `anyOf:[array, object]`, object value | `{"entry":{"body":"Visible","privateNote":"hidden"}}` | `{"entry":{"body":"Visible"}}` |
| `type:["array","object"]`, object value | leak | `{"entry":{"body":"Visible"}}` |
| `--filter` + `title,comments.body` | — | `[{"title":"keep","comments":[{"body":"b1"}]}]` |

What landed, per finding:

- **#1** — `schemaMayBeArray` now threads a `root` and re-roots on any node carrying its own `$defs`. The `seen` set also became a proper *in-progress* set (`ancestors`, deleted in a `finally`), which fixes the order-dependent false negative I noted as a sub-point. The updated `"projects concise array paths through source schema references"` test moved `$defs` to the root and added a `$ref` chain — it now exercises the real emission shape rather than the inline arrangement that masked the bug.
- **#2** — `projectValue` gained the symmetric case (`piece-get-transform.ts:882`): a record value under an array-shaped mask descends into `items` instead of falling through to the permissive copy. Covered by a new two-variant test (`anyOf` union and `type` array).
- **#3** — better than what I suggested: `selectSourceSchema` now splits on `purpose`. `source-read` returns the composition untouched (identity-verified in a test), and only `projected-output` re-projects — re-emitting `anyOf` separately and folding `oneOf` into `allOf: [...allOf, {anyOf: oneOf}]`, so the conjunction survives and only the exclusivity is deliberately relaxed, with a comment explaining why.
- **#4** — `mergeMasks` descends into array masks instead of bailing to `true`.
- **Question** — answered in code: `piece.ts` distinguishes an absent source (`getRaw() === undefined` → ordinary `null` response) from a present source that failed to materialize (→ nonzero). README / `skills/cf` / `skills/topics` all updated to match, and both nits (dead `?? projectionSchema`, dead align branch) are gone.

**Two remaining nits, neither worth another round:**

1. `packages/cli/lib/piece.ts:2232` — the `typeof targetCell.getRaw === "function"` guard exists only for the test mock; `Cell` always declares `getRaw` (`packages/api/interface/cell.ts:367`). If the cell shape ever drifted, the guard would silently disable the absent-source check rather than fail loudly. The mock already supplies `getRaw`, so the `typeof` can just go.
2. `packages/cli/lib/piece-get-transform.ts:987` — the composition branch returns early, so a source node carrying *both* a composition and its own `properties` keeps those properties unnarrowed in the `projected-output` schema. Values are still correctly masked by `projectValue`; only the declared output schema is looser than necessary.

**Verdict: findings resolved — mergeable.** The fixes are targeted, the new tests reach the cases that were missing, and the docs now describe the real behavior.

PR #5277, opened 2026-08-03
fix(runner): log the silent resume owned-cell skip; name the of:/computed: pair (CT-1940, CT-1943)

bfollington:
## Why

Two diagnosability gaps found while investigating a live estuary home space
whose Favorites / Profile / Self tabs rendered empty with **zero** console
output — even with `forwardWorkerConsole` on. Localizing that took a full
session of browser-console probing, and both of these are reasons why.

Neither change alters behaviour. Both are about making the runtime say what it
is doing.

## What changed

**CT-1940 — the silent skip is now audible.**

`collectResumeOwnedCells` has two exits that skip the same work — a node's
owned-cell pre-sync AND the recursion that would reach the child's own
`derivedInternalCells` manifest — but only the throwing one logged:

```ts
} catch (error) {
  logger.warn("resume-owned-cells", …);
  continue;
}
if (spotLink === undefined) continue;   // <- silent
```

The `undefined` exit now warns at the same level under the same
`resume-owned-cells` key, so existing console filters catch both. Its message
names the *different* cause — `…outputs resolved to no write redirect`, versus
the sibling's `…did not bind or resolve`, which reads as a throw. The sibling's
message string is unchanged, so anything grepping it still matches.

Both exits now carry a shared identity payload
(`describeSkippedSubPatternNode`): result cell, space, node index, node
description, child pattern — enough to find the node from a production trace.

This matters because #5143 deliberately moved a real case from the throwing
path to the `undefined` path (`firstResolvedOutputRedirect` now returns
`undefined` for outputs holding only `partialCause` aliases, where it used to
throw). That was right for the scan, but it converted a loud skip into a silent
one.

**CT-1943 — the dual entity kind is documented.**

`instantiatePatternNode` binds a sub-pattern node's outputs twice: once WITH
`derivedInternalCells` (the value bind — the manifest descriptor carries the
kind, so the mint lands at `computed:fid1:<hash>`, and this is where
`sendValueToBinding` writes the child link), and once WITHOUT it (the identity
bind — no descriptor means no kind, so the same hash mints as `of:fid1:<hash>`,
used only as the `resultFor` cause).

Two entities differing only by scheme. Reading the child link at the `of:` id
returns `undefined` for a perfectly healthy piece — which has now produced
wrong conclusions in two separate investigations. Comments at both bind sites,
plus a recipe under `docs/development/debugging/console-commands.md` so someone
probing live state knows to look for both.

## Deliberately NOT in this PR

CT-1940 also asks whether skipping the child recursion is correct at all. It
probably is not — the justification at `firstResolvedOutputRedirect` ("the
derived cells themselves are collected from each pattern's own
`derivedInternalCells` manifest") is exactly what the `continue` prevents. But
fixing it means minting a child result cell from something other than the
resolved output spot, which CT-1897 warns derives the wrong `resultFor`
identity and pre-syncs the wrong subtree. That is a design decision about
child-cell identity, not a refactor, and it needs an owner. Recorded on
CT-1940.

Surfacing the `of:` / `computed:` pairing in `cf inspect` output is likewise
deferred — `packages/state-inspector/decode.ts` strips `of:` and keeps
`computed:` visible, so the pairing is encoded in the output but never
explained. Real inspector change; recorded on CT-1943.

## Test plan

- New `packages/runner/test/resume-owned-cells-skip-log.test.ts`: resumes a
  pattern whose sub-pattern node's outputs hold only a deferred `partialCause`
  alias, and asserts the warn fires and carries the result cell id, node index
  and node description. Verified red without the change.
- Full `packages/runner` suite: **1113 passed (5836 steps), 0 failed** — run
  independently of the authoring pass.
- `deno fmt --check` and `deno lint` clean on the changed files (the repo's ~23
  pre-existing unformatted files are untouched).
- `deno task check-docs`: 522 blocks pass, including the new fence.

---
## Summary by cubic
Improves diagnosability during resume by logging skipped sub-pattern nodes with clear identity and documenting when reads should use `computed:` instead of `of:`. Addresses CT-1940 and CT-1943. No behavior changes.

- **New Features**
  - Runner now logs both skip paths under `resume-owned-cells`: binding failures stay warn; “no write redirect” skips are debug. Both include result cell id, space, node index, node description, and child pattern.
  - Clarified `of:` vs `computed:` identities: when a descriptor is `kind: "computed"`, the value lives at `computed:fid1:<hash>` and reads at `of:fid1:<hash>` return undefined; with a kindless descriptor or `experimental.computedCellIds` off, both binds land on `of:`. Notes added in `packages/runner/src/runner.ts` and a recipe in `docs/development/debugging/console-commands.md` (README links to it).
  - Tests in `packages/runner/test/resume-owned-cells-skip-log.test.ts` table-drive both exits, pin levels (debug vs warn), assert the identity payload, and verify the healthy Home run emits no warns under this key.

<sup>Written for commit 866869add05afcae6e1ddce6314fd484ca104d28. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5277?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/development/debugging/README.md
+++ b/docs/development/debugging/README.md
@@ -115,7 +115,9 @@ guide:
 
 - [Console Commands](console-commands.md) - `globalThis.commonfabric.*` browser console reference
   - Starts with common tasks: read piece data, dump the rendered VDOM, diagnose
-    churn, find dead handlers, watch values, agent-browser recipes
+    churn, find dead handlers, watch values, agent-browser recipes — including
+    why a sub-pattern's cell reads `undefined` at an `of:` id while its value
+    lives at the `computed:` id of the same hash
   - Reference tail covers logger counts/timing/baselines/flags and worker traces
 - **Server-side write trace** — set `CF_DEBUG_MEMORY_WRITES=1` on the toolshed to
   log every memory write as `[memwrite] c=<conn> op=… id=… scope=… vhash=…`,
--- a/docs/development/debugging/README.md
+++ b/docs/development/debugging/README.md
@@ -116,7 +116,7 @@ guide:
 - [Console Commands](console-commands.md) - `globalThis.commonfabric.*` browser console reference
   - Starts with common tasks: read piece data, dump the rendered VDOM, diagnose
     churn, find dead handlers, watch values, agent-browser recipes — including
-    why a sub-pattern's cell reads `undefined` at an `of:` id while its value
+    why a sub-pattern's cell can read `undefined` at an `of:` id while its value
     lives at the `computed:` id of the same hash
   - Reference tail covers logger counts/timing/baselines/flags and worker traces
 - **Server-side write trace** — set `CF_DEBUG_MEMORY_WRITES=1` on the toolshed to
--- a/docs/development/debugging/console-commands.md
+++ b/docs/development/debugging/console-commands.md
@@ -48,28 +48,38 @@ Always check the actual stored value before assuming a write failed — see
 
 ### A sub-pattern's cell reads as `undefined` at the id you have
 
-A sub-pattern node's output spot exists as **two entity ids that differ only by
-URI scheme**: `of:fid1:<hash>` and `computed:fid1:<hash>`. Both are minted from
-the same cause, so they share a hash — the scheme carries the entity kind, and
-the kind is part of the identity, so these are two distinct entities. See
-[computed cell identity](../../specs/computed-cell-identity.md).
+`instantiatePatternNode` (`packages/runner/src/runner.ts`) binds a sub-pattern
+node's output spot twice, from the same cause, so both mints share a hash:
 
-The runner mints both, for different jobs (`instantiatePatternNode`,
-`packages/runner/src/runner.ts`):
+- the **identity bind** never sees the pattern's `derivedInternalCells`
+  manifest, so it has no entity kind and always lands on `of:fid1:<hash>`. Its
+  job is to name the child result cell as its `resultFor` cause, and it is the
+  spot the resume owned-cell walk keys on;
+- the **value bind** sees the manifest, so it mints under the descriptor's kind
+  — and the child link is sent to whatever that bind produced, so that is the
+  id holding the value.
 
-- **`of:fid1:<hash>` is identity only.** It names the child result cell as its
-  `resultFor` cause, and it is the spot the resume owned-cell walk keys on.
-  Nothing writes a value there.
-- **`computed:fid1:<hash>` holds the value.** The child link is sent to this
-  one, so this is the id to read.
+Whether those are one entity or two therefore depends on the descriptor. The
+URI scheme carries the kind, and the kind is part of the identity (see
+[computed cell identity](../../specs/computed-cell-identity.md)):
 
-A read at the `of:` id therefore returns `undefined` for a perfectly healthy
-piece — that is the expected answer at that id, not evidence that a sub-piece
-is missing. Read the `computed:` id before concluding anything:
+- **descriptor classified `kind: "computed"`** — the value bind mints
+  `computed:fid1:<hash>`, a **distinct entity from `of:fid1:<hash>`, differing
+  only by scheme**. Nothing writes a value at the `of:` id, so a read there
+  returns `undefined` for a perfectly healthy piece — that is the expected
+  answer at that id, not evidence that a sub-piece is missing. Read the
+  `computed:` id before concluding anything;
+- **kindless descriptor** — the classifier declined the node, or
+  `experimental.computedCellIds` is off (see
+  [experimental options](../EXPERIMENTAL_OPTIONS.md)). Both binds land on the
+  same `of:fid1:<hash>` entity, and the child link is written there. A read at
+  the `of:` id returns the value, and there is no `computed:` sibling.
+
+So an `undefined` read is only diagnostic once you have tried both ids:
 
 ```javascript
 // Shown at module scope.
-// Same hash, two kinds — read the computed one for the value.
+// Same hash, two possible kinds — try both before concluding anything.
 await commonfabric.readCell({ space: "did:key:z6Mkm...", id: "of:fid1:HASH" })
 await commonfabric.readCell({
   space: "did:key:z6Mkm...",
@@ -77,12 +87,12 @@ await commonfabric.readCell({
 })
 ```
 
-The same trap applies to any offline scan that enumerates a space by matching
-the `of:fid1:` prefix: it will not see the cells holding sub-pattern children.
-In `cf inspect` output the distinction is there but quiet — it abbreviates ids
-by stripping `of:` while keeping a `computed:` scheme visible, so a bare
-`fid1:<hash>` in a row is the `of:` form and its `computed:` sibling is a
-separate row.
+Where the pair does exist, the same trap applies to any offline scan that
+enumerates a space by matching the `of:fid1:` prefix: it will not see the cells
+holding sub-pattern children. In `cf inspect` output the distinction is there
+but quiet — it abbreviates ids by stripping `of:` while keeping a `computed:`
+scheme visible, so a bare `fid1:<hash>` in a row is the `of:` form and its
+`computed:` sibling, when there is one, is a separate row.
 
 ### What's actually rendered?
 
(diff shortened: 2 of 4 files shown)


bfollington (commented):

PR #5277 · deep-read all four changed files plus the logger, binding/minting paths, computed-ID spec/flag, and relevant resume tests · ran 5 targeted tests (10 steps), all passing; inspected the completed runner-shard CI logs.

## 🔴 Blocking

- **[major, verified] `packages/runner/src/runner.ts:3814` — the new `WARN` fires during healthy, ordinary home runs.** Runner shard 2 emitted this exact new warning **eight times outside the dedicated test**: twice while “materializes the real home pattern over a pre-favorites root under enforcement” passed, and twice in each of the three passing `home-add-favorite` cases. Those cases are not failed resumes; `syncCellsForRunningPatternInner` explicitly notes that fresh first runs are expected to lack argument metadata, and it still calls `collectResumeOwnedCells`. So `spotLink === undefined` is not currently an exceptional-enough condition for an unconditional warning. This makes normal Home startup/resume noisy and weakens the signal this PR is trying to add. Gate the warning on evidence of a genuine stored resume/failure, or record this expected branch below `warn` (debug/count/flag); add a negative test using the real Home pattern or another representative healthy run so normal execution is proven quiet. Evidence: [Runner Tests (2/5)](https://github.com/commontoolsinc/labs/actions/runs/30784968094/job/91596682718).

## 🟡 Improvements

- **[minor, verified] `docs/development/debugging/console-commands.md:51` and `runner.ts:6382-6443` — the two-ID explanation is written as unconditional, but it only holds when the output descriptor has `kind: "computed"`.** The implementation trace itself is otherwise correct: the manifest-aware/value bind carries `descriptor.kind` and is passed to `sendValueToBinding`; the manifest-free/identity bind mints `of:` and its resolved redirect becomes the `resultFor` cause. But when the descriptor is kindless—including the still-supported `experimental.computedCellIds: false` rollback—the two binds land on the same `of:` entity and the child link is written there. In that case “nothing writes a value there”, “read the `computed:` id”, and the unconditional `VALUE BIND (kind: computed:)` heading are false. Make the guide and headings conditional: computed descriptors produce the `computed:`/`of:` pair; kindless descriptors produce one `of:` entity.

- **[minor, verified] `resume-owned-cells-skip-log.test.ts:90-145` is not tautological, but it pins only the positive path on an intentionally invalid pattern.** The exact new message makes it fail without the implementation, so it genuinely guards emission. It does not guard the central regression above, and the blanket `catch {}` means the fixture succeeds as a test despite later instantiation rejecting it. It also does not exercise the existing catch branch whose payload changed, nor assert the advertised `space` and `childPattern` fields. Once the warning predicate is corrected, add a healthy negative case and preferably table-drive both skip exits while asserting the payload object directly.

## Checked, no finding

- Appending the identity object to the existing catch warning preserves the logger key, message text, error position, and substring matching. There is no in-repo consumer/parser of that line. An unknown external parser that requires exact console arity or exact full-line serialization would see a change, but the repository exposes no such contract.
- The new comments are right about which bind supplies the kind and which bind supplies the `resultFor` cause; the issue is only their unconditional `computed:` wording.

**Verdict:** not ready to merge because the new warning is demonstrably noisy on healthy Home executions; the remaining issues are documentation/test precision.

bfollington:
Thanks — the blocking finding was right, and it was the thing this PR most needed caught. Reproduced it locally before changing anything: `cfc-additive-default-preserves-old-doc.test.ts` emitted the new warning twice during a passing test.

**On the remedy: I went further than gating, because gating doesn't work.** Both suggested predicates were tested and rejected:

- *Gate on a fresh first run* — only ONE of the two occurrences co-occurs with the existing `resume-pre-sync argument meta link missing` warn. The second fires on a pass where the argument meta IS present.
- *Gate on a genuine stored resume* — that test materializes over a stored pre-favorites root, which IS a genuine resume, and it still fires.

So `spotLink === undefined` isn't exceptional under any predicate we could find. It's the by-design outcome for outputs holding only `partialCause` aliases — exactly the case #5143 deliberately moved off the throwing path. The level was simply wrong, so it is now `logger.debug`, same key, same message, same payload. The sibling catch stays `warn`; that one is a real binding failure.

Negative test added per your suggestion: it compiles and runs the real `system/home.tsx`, asserts zero warn-level emissions for the key, AND asserts the exit is still reached — so the silence can't go vacuous if the code path moves. Restoring `logger.warn` turns it red.

**One thing your review surfaced indirectly, worth flagging:** `CF_LOG_LEVEL=debug` does NOT bring the debug line back. The runner logger is constructed `getLogger("runner", { level: "warn" })` and `shouldLog` takes the more restrictive of (global floor, per-logger level), so the floor can never lower it. Verified both ways. That is true of every `logger.debug` in `runner.ts`, not just this line, and it means the documented env var cannot reach runner debug output at all. Filed separately rather than expanded here; the comment at the call site now cites the knobs that actually work.

Both improvements also addressed: the `of:`/`computed:` wording is now conditional on descriptor kind (verified against `getDerivedInternalCellLink` and `assignComputedCellKinds` — kindless descriptors arise from ordinary classification as well as the `computedCellIds` rollback), and the test is table-driven across both exits with whole-object payload assertions and an explicit rejection instead of the blanket `catch {}`.

Full `packages/runner`: 1113 passed, 0 failed.

bfollington (commented):

PR #5277 follow-up commit `866869add` · deep-read the runner/logging, test, binding/minting, flag/classifier, and docs delta · ran 3 targeted test files (61 steps), the focused formatter/linter checks, a deliberate warn-restoration mutation, and inspected current CI (all green).

## No blocking findings

- **The `spotLink === undefined` demotion is correct.** The real `system/home.tsx` fixture reaches the exit twice in a healthy run, so neither “fresh first run” nor “genuine stored resume” distinguishes an exceptional occurrence. An unconditional warning would knowingly make healthy execution noisy. `debug` accurately records a by-design but diagnostically interesting branch.

- **The negative guard is non-vacuous and mutation-sensitive.** With the submitted code, the focused file passes all 3 steps. I changed only this call back to `logger.warn`: the table-driven level case failed, and the Home negative case independently failed with two captured warn calls while `skips.length > 0` remained true. Restoring the submitted `debug` call returns the worktree to clean and the test to green.

- **The debug call is not dead in the intended live-browser path.** `CF_LOG_LEVEL=debug` alone cannot lower a runner logger constructed at `warn`; that diagnosis and CT-1944 are correct. But `commonfabric.rt.setLoggerLevel("debug", "runner")` mutates the actual worker logger, and browser workers have no Deno environment floor, so it exposes this call. Likewise, direct `commonfabric.logger["runner"].level = "debug"` works in a realm without a more restrictive global floor. In Deno/CLI contexts with the default `warn` floor, both the global floor and the per-logger level must permit debug. That limitation does not make the demotion wrong.

- **The sibling catch can remain `warn` in this PR, with one nuance.** It is the pre-existing exception path and, unlike the undefined return, represents an actual failed bind/resolve operation. However, “genuine failure” should be read as failure of that pre-sync binding attempt, not proof the overall run is unhealthy: the table fixture deliberately triggers it via the expected fresh-run absence of an argument meta link, and `runSynced` still succeeds. If warn purity becomes a concern, a follow-up could distinguish that known fresh-run condition from other thrown resolver failures; this PR does not newly introduce the warning.

- **The conditional `of:` / `computed:` explanation is accurate.** `getDerivedInternalCellLink` applies only `descriptor.kind`; `assignComputedCellKinds` leaves descriptors kindless when classification declines them and is skipped entirely when `experimental.computedCellIds` is false. In either case the value and identity binds mint the same `of:` entity. A `kind: "computed"` descriptor alone creates the same-hash, different-scheme pair described in the docs.

- **The table coverage is materially stronger.** Both exits are exercised, the levels and complete payload objects are pinned, the invalid fixture has an explicit rejection, and the Home case guards the production-shaped negative behavior.

**Verdict:** ready to merge; the prior blocker and both improvements are addressed.

PR #5278, opened 2026-08-03
fix(piece): preserve declared slot schema authority (CT-1937)

bfollington:
## Summary

- keep the declared Piece slot schema authoritative after resolving linked targets
- make ordinary reads and cf piece get filter/schema transforms share the same selection and resolution boundary
- preserve scope caps and CFC metadata while preventing broader mixed-generation targets from exposing excluded fields
- reject filter or projection paths excluded by the declared slot instead of silently treating them as absent
- update task-management guidance to prefer Common Fabric Topics and remove the obsolete local tracker configuration

## Validation

- deno test -A packages/runner/test/piece-helpers.test.ts packages/runner/test/ascell-scope-cap.test.ts packages/runner/test/link-resolution.test.ts
- deno test -A packages/cli/test/piece.test.ts packages/cli/test/piece-get-transform.test.ts
- deno test -A packages/piece/test/pull-materialization.test.ts
- nested Cell wrappers, payload-less asCell slots, false schemas, numeric paths, and composition branches are exercised
- aggregate CI Coverage Check passes at or below the main ratchet for CLI, Piece, and runner
- deno task check-docs
- deno task check-skill-facts
- deno lint
- all touched files pass deno fmt --check; the repo-wide check still reports 23 unrelated pre-existing files

Linear: https://linear.app/common-tools/issue/CT-1937/cf-piece-get-transforms-must-preserve-declared-slot-schema-authority


packages/runner/src/piece-helpers.ts
```diff
@@ -89,6 +94,63 @@ export function resolveCellPath<T>(
...
+  let readSchema = declaredSchema;
+  if (isRecord(declaredSchema)) {
+    const declaredWrappers = ContextualFlowControl.getAsCellValues(
+      declaredSchema,
+    );
+    if (declaredWrappers.length > 0) {
+      const { asCell: _asCell, ...declaredPayload } = declaredSchema;
+      const payloadSchema: JSONSchema = {
+        ...declaredPayload,
+        ...(declaredWrappers.length > 1 && {
```
bfollington, line 144: 🔴 **Blocking (part 1 of 2) — this branch is never entered by the PR's validation set.**

I replaced this branch (and the one at L144) with `throw new Error(...)` and re-ran all four cited test files — **zero hits**. The `Cell<Cell<T>>` shape is supported and separately tested (`packages/runner/test/schema-basic.test.ts:724`), and `packages/piece/test/pull-materialization.test.ts:2852` exercises `asCell: ["cell", "cell"]` only on the **`set`** side — never through this read boundary.

This is new logic on a security-adjacent boundary with no test at all, and it's part of what puts `coverage-check` (and therefore the required `Status` job) in the red at +5 lines for `packages/runner`.

**Fix**: add a `Cell<Cell<number>>` slot read through `PiecePropIo.get`.

packages/runner/src/piece-helpers.ts
```diff
@@ -89,6 +94,63 @@ export function resolveCellPath<T>(
...
+          asCell: declaredWrappers.slice(1),
+        }),
+      };
+      readSchema = Object.keys(payloadSchema).length === 0
```
bfollington, line 148: 🔴 **Blocking (part 2 of 2)** — same probe, same result: the `Object.keys(payloadSchema).length === 0 → true` arm is never taken by the cited tests. That's the payload-less slot, i.e. a bare `{ asCell: ["cell"] }` with no declared type.

**Fix**: read a payload-less `asCell` slot through the Piece boundary and assert the target materializes.

packages/runner/src/piece-helpers.ts
```diff
@@ -89,6 +94,63 @@ export function resolveCellPath<T>(
...
+ * Resolve a Piece read address without letting a linked target replace its
+ * declared schema. One outer Cell wrapper is consumed when present.
+ *
+ * `resolveAsCell()` must follow the stored link first so scope caps and CFC
+ * dereference metadata are applied. A link minted by another pattern generation
+ * may carry a broader schema, however, so the schema declared on the materialized
+ * handle is reasserted after resolution. Both ordinary and transformed Piece
+ * reads use this boundary helper and therefore share one schema authority rule.
+ */
+export function resolvePieceReadCell<T>(cell: Cell<T>): Cell<T> {
```
bfollington: ⚪ Nit — `resolvePieceReadCell<T>(cell: Cell<T>): Cell<T>` is a type lie in exactly the case the function exists for: given `Cell<Cell<X>>` it returns a cell over `X`. Every call site passes `Cell<unknown>` so nothing breaks today, but the signature advertises the opposite of what the body does.

packages/cli/lib/piece-get-transform.ts
```diff
@@ -1043,9 +1044,14 @@ export function selectSourceSchema(
   const properties: Record<string, JSONSchema> = {};
   for (const [key, childMask] of Object.entries(mask.properties)) {
     const child = cfc.schemaAtPath(source, [key]);
+    // `false` is an authoritative exclusion from a closed declared source
+    // schema, not the same thing as having no source schema. Omitting it keeps
+    // a caller projection or predicate from widening a mixed-generation link
+    // target that happens to publish the requested field.
+    if (child === false) continue;
```
bfollington, line 1222: 🟡 **Transforms silently drop schema-excluded fields; plain reads error.**

The `continue` is the right *mechanism*, but it leaves the CLI's two read modes disagreeing in a way an agent can't detect:

| command | excluded field `secret` |
| --- | --- |
| `cf piece get topic/secret` | throws `property "secret" not found` ✅ |
| `cf piece get topic --schema title,secret` | returns `{ title: "Visible" }` — no error, no warning |
| `cf piece get topics --filter '.secret != null'` | returns `[]` |

The last row is the sharp one, and this PR pins it (`packages/cli/test/piece-get-transform.test.ts:730` asserts `.toEqual([])`). An agent reads `[]` as *"no topic has a secret"* when the truth is *"this slot cannot see `secret`"* — a wrong conclusion with no signal that anything was elided. Given the verb contract is explicitly an agent-facing surface, silent narrowing is worse than a hard error.

**Fix**: when a projection or predicate path resolves to `false` against the declared source schema, throw a `PieceGetTransformError` naming the path (`--filter path ".secret" is not declared by this slot's schema`) where the mask is aligned in `resolveProjection` / predicate-path validation. Keep the `continue` as defence-in-depth.

*(verified — reproduced both rows locally)*

packages/cli/test/piece-get-transform.test.ts
```diff
@@ -598,6 +598,161 @@ describe("cf piece get transforms", () => {
...
+      await derivePieceGetValue(
+        runtime,
+        space,
+        listContainer.key("topics"),
+        { filter: parsePieceGetFilter(".secret != null") },
+      ),
+    ).toEqual([]);
+  });
+
+  it("keeps a declared link scope cap during transforms", async () => {
```
bfollington, line 851: 🟡 **Weak assertion — no negative control.**

`expect(...).toBeUndefined()` passes just as happily if projections through *any* `asCell` slot start returning `undefined` for an unrelated reason — which is exactly the regression class this boundary is most likely to produce.

**Fix**: add the uncapped twin (same shapes, no `scope: "space"` cap) asserting `{ field: "session-only" }`, so the pair pins the cap as the *cause* rather than the symptom.

packages/cli/test/piece-get-transform.test.ts
```diff
@@ -598,6 +598,161 @@ describe("cf piece get transforms", () => {
     ).toEqual({ title: "Visible" });
   });
 
+  it("keeps a declared linked-slot schema narrower than its target", async () => {
```
bfollington, line 627: ⚪ Nit — four stories in one `it`: plain read, projection, CFC label probe, and the array/`--filter` case with its own cell setup. Splitting the array half into its own `it` would let a failure name its own cause.

packages/piece/src/ops/piece-controller.ts
```diff
@@ -2260,18 +2262,20 @@ class PiecePropIo implements PieceCellIo {
...
+    // same boundary helper used by transformed CLI reads.
+    const selectedCell = selectPieceReadCell(targetCell, path);
+    if (selectedCell.schema === false) {
+      return await this.#getFromRoot(targetCell, path);
+    }
+    const selectedReadCell = resolvePieceReadCell(selectedCell);
+    await selectedReadCell.pull();
     // Relax `required` for scoped links that this session cannot materialize,
     // at the selected subtree rather than the root — same boundary as
     // #getFromRoot, see schemaWithScopedLinkRequiredsRelaxed.
```
bfollington, line 2276: ⚪ Nit — this comment ("An asCell projection materializes even an absent or explicitly undefined slot as a Cell") now describes a path single-wrapper slots no longer take: `resolvePieceReadCell` unwraps them, so an absent slot falls through to the `selected === undefined` branch below instead. The `isCell(selected)` branch survives for undeclared-schema cells, but the comment reads as if it's still the main road.

Related: was reintroducing `resolveAsCell()` here a deliberate reversal of #5231? The comment you replaced said the terminal `asCell` is unwrapped *"without routing through `resolveAsCell()`"* because the projection applies the cap itself. The capped tests still pass, so the cap does survive — but worth a line saying #5231's reasoning is superseded rather than forgotten.

packages/cli/lib/piece.ts
```diff
@@ -2200,7 +2201,7 @@ export async function getCellValue(
       await piece.getCell().pull();
       const rootCell =
         await (options.input ? piece.input.getCell() : piece.result.getCell());
-      const targetCell = rootCell.key(...path);
+      const targetCell = selectPieceReadCell(rootCell, path);
```
bfollington, line 2319: ⚪ Nit — this branch guards `targetCell.schema === false`; the `--step` prefetch a few lines above (`selectPieceReadCell(rootCell, path)` → `pull()`) doesn't, so it issues a `pull()` on a `false`-schema cell. Harmless today — the pull is effectively a no-op and the real read still produces the right diagnostic — but the asymmetry will read as an oversight later.

bfollington (commented):
**cf-review via Claude Opus 5, on behalf of @bfollington**

**Scope** · 12 files / +436 −52 across `runner`, `piece`, `cli` + 3 docs · **Coverage**: deep-read of all changed source, ran the PR's full validation set locally, branch-probed the new helper · **Checks run**: `deno test -A` on all 4 cited test files (all pass), `deno check` + `deno lint` + `deno fmt --check` on touched sources (all clean), plus a hand-written CFC probe and an instrumented-branch probe.

**What it does.** Two shared helpers in `packages/runner/src/piece-helpers.ts` — `selectPieceReadCell` (walk a path but pin the *declared* leaf schema, including an explicit `false`) and `resolvePieceReadCell` (follow the stored link for scope caps / CFC dereference metadata, then reassert the declared payload schema) — now back both `PiecePropIo.get` and `cf piece get --filter/--schema`, so a link minted by another pattern generation can't widen a closed slot.

The core idea is right and the mechanism holds up under inspection: `asSchema()` preserves `_link.scopeCaps` and `_cfcLabelView`, so the scope cap and dereference labels survive the reassertion. I specifically tried to break it on the CFC axis — a probe where the **target's** schema declares `ifc.confidentiality` that the slot schema lacks — and the atom still lands on the derived label (it comes from the target address's stored label map, not the read schema). No leak there.

Findings are inline. Summary:

### 🔴 Blocking

**1. CI is red on the required gate.** [`coverage-check`](https://github.com/commontoolsinc/labs/actions/runs/30790699647/job/91613917321) reports:

```
OVER   164   172   +8 (+5%)   coverage-debt  packages/piece
OVER  5560  5565   +5 (+0.1%) coverage-debt  packages/runner
```

`coverage-check` rolls into the aggregate `Status` job, which fails — so this blocks merge rather than being the usual non-required perf-gate flavour. It also contradicts the PR body's "focused V8 coverage reports no uncovered executable lines in the modified runner, Piece, or CLI source files": that was line coverage over a hand-picked test set; the ratchet measures the group under the whole suite.

I verified *which* lines rather than guessing — see the two inline comments on `piece-helpers.ts`. Both new branches of `resolvePieceReadCell` are dead to the PR's own validation.

**Fix**: cover both (a `Cell<Cell<number>>` slot read and a payload-less `asCell` slot read through `PiecePropIo.get` would do it, and would retire most of the +8/+5), or consciously accept with `NEW_COVERAGE_BASELINE` / a targeted `NEW_PERF_BASELINE: coverage-debt: …` line in the PR body and rerun. Covering is the better call — the `slice(1)` branch is new logic on a security-adjacent boundary with no test at all.

### 🟡 Improvements

**2. Transforms silently drop schema-excluded fields; plain reads error.** Inline on `piece-get-transform.ts`. `cf piece get topics --filter '.secret != null'` returns `[]` for a field the slot simply cannot see — an agent reads that as "no topic has a secret".

**3. Stale `beads` plumbing left behind by the skill rewrite.** `.gitattributes:2-3`, `deno.jsonc:97,126` *(not in the diff, so no inline anchor)*

The rewritten skill now says *"Do not add generated tracker state or local task databases to the repository"*, but the repo still carries `.beads/beads.jsonl merge=beads` in `.gitattributes` — a merge driver whose whole purpose is committing that state — plus `.beads/` exclusions in `deno.jsonc`. The `.gitattributes` line now actively contradicts the only doc that describes the practice.

**Fix**: drop the `.gitattributes` stanza in this PR; the `deno.jsonc` exclusions are harmless and can go with it or stay. *(verified — `.gitattributes` and `deno.jsonc` are the only remaining `bd`/beads references outside the unrelated `cf-message-beads` component)*

**4. Weak assertion in the scope-cap test.** Inline.

### ⚪ Nits

Four, all inline: a four-story `it`, a misleading generic signature on `resolvePieceReadCell`, a comment in `piece-controller.ts` describing a path single-wrapper slots no longer take, and a missing `schema === false` guard on the `--step` prefetch in `piece.ts`.

### Questions

- **Should a schema-excluded path in `--filter`/`--schema` be an error rather than an omission?** (Finding 2.) You've clearly thought about this boundary — I want to know whether silent omission is the intended contract or a side effect of where the fix landed.
- **Was reintroducing `resolveAsCell()` on the piece-controller read path a deliberate reversal of #5231?** (Asked inline.)

### Possible follow-ups

- A single "declared slot schema is authoritative" note in `docs/common/concepts/types-and-schemas/`. The rule currently lives only in `packages/cli/README.md`, which is the wrong home for a runtime-semantics invariant that `PiecePropIo` also enforces.
- `resolvePieceReadCell` / `selectPieceReadCell` are exported from the top-level `@commonfabric/runner` barrel but are Piece-read-boundary specific. Worth deciding whether they belong in the public surface or behind a subpath.

---

**Verdict**: the fix is sound and the mechanism survives the adversarial checks I ran — 🔴 blocked only on the red coverage gate, whose two uncovered branches I confirmed are genuinely untested rather than a measurement artifact. Findings 2 and 3 are worth folding in while you're here.

packages/cli/test/piece-get-transform.test.ts
bfollington: Added the uncapped twin in 872f13d04. The capped projection remains undefined while the same session-scoped target through an uncapped handle returns the expected field object.

packages/cli/lib/piece-get-transform.ts
bfollington: Changed in 872f13d04. Filter and projection paths are validated against the declared source schema before graph construction and now throw PieceGetTransformError naming the excluded path. Coverage includes filter, concise projection, and explicit array JSON Schema projection; the omission remains only as defence-in-depth.

packages/runner/src/piece-helpers.ts
bfollington: Fixed in 872f13d04. The helper now returns Cell<unknown>, which reflects that consuming one outer wrapper changes the payload type. Both shared helpers are also marked internal at their package-boundary export.

packages/piece/src/ops/piece-controller.ts
bfollington: Updated in 872f13d04. The branch comment now describes remaining nested wrappers and undeclared-schema cells. The cold-read test also records the deliberate reason for resolveAsCell here: capture scope and CFC dereference metadata first, then reassert the declared payload schema.

packages/runner/src/piece-helpers.ts
bfollington: Addressed in 872f13d04. Added a PiecePropIo.get regression test named consumes exactly one wrapper from a nested asCell slot; it constructs and reads a Cell<Cell<number>> value and asserts 7.

packages/cli/test/piece-get-transform.test.ts
bfollington: Split in 872f13d04. The linked-list filter and explicit array-schema exclusion cases now have their own test and setup.

packages/cli/lib/piece.ts
bfollington: Fixed and covered in 872f13d04. The step prefetch skips false-schema cells, and the CLI test runs with step enabled while making the forbidden target pull throw if invoked.

packages/runner/src/piece-helpers.ts
bfollington: Addressed in 872f13d04. Added a PiecePropIo.get regression test named materializes a payload-less asCell slot, using a bare asCell declaration and asserting the broader target object materializes.

bfollington:
Closing without merge after reviewing the underlying schema semantics.

The reproduction that motivated CT-1937 is real, but this PR assumes the
receiving Piece slot schema is an authority ceiling. Generic `Cell.asSchema()`
instead permits a caller-selected sibling view over the same cell, while CFC
metadata and inherited scope caps remain the security boundary. Consequently,
this change would introduce a new Piece-specific narrowing policy rather than
unambiguously correct existing runtime behavior.

The conclusion and a safer direction for any follow-up are recorded on CT-1937.
The branch is retained for reference.

PR #5280, opened 2026-08-03
fix(runner,memory): replay operation-based mergeable ops against live pending state

ubik2:
## Summary

Pending optimistic replay reconstructed a patch's changed paths by projecting
values out of the transaction's full optimistic snapshot. For `append`/
`add-unique`/`increment`/`remove-by-value` (mergeable ops that carry relative
intent, not an absolute value), that snapshot can still hold data from an
earlier pending layer that was since rejected — replaying by projection could
resurrect it.

Adds an `operationBased` flag to the patch-op descriptor registry
(`packages/memory/v2/patch.ts`) and replays those ops by applying their wire
intent to the rebuilt live base (matching how the durable store itself
materializes them), while snapshot-derived ops keep the existing projection
replay.

This was split out of #5258, which combined this fix with an unrelated
`conflictAdmissionMode` "hold" removal (CT-1925) — that fix now has its own
PR (#5279); #5258 is closed.

## Known open issue from prior review

@ubik2 reviewed this change while it was part of #5258 and flagged a
**blocking** finding that still applies here and needs to be addressed before
merge:

> `packages/runner/src/storage/v2.ts:492-495` — operation-based replay can
> throw when rebuilt live state contains an incompatible scalar/null ancestor.
> Reachable when C1 optimistically turns `choice: null` into an object, an
> independent C2 appends beneath `choice` (mergeable ops have no pending-read
> dependency by design), and C1 rejects: rematerializing C2 calls `applyPatch`
> against the confirmed scalar/null and throws `path is not traversable` out
> of the replica instead of resolving through the normal rejection path. The
> regression added here only covers replay from an absent `{}` base, where
> create-missing succeeds.

Non-blocking improvements also flagged on the same review, potentially
relevant to this diff's neighborhood:

- `packages/memory/test/naive-admission.ts:145-155` — the test oracle picks
  the max resolution *sequence*; the legacy rule picks the resolution at the
  highest *localSeq*. These can diverge under overtaking.
- `packages/memory/test/v2-differential-consistency-test.ts:230-239` — legacy
  and true-basis coverage are combined into one sanity check, so either path
  silently disappearing wouldn't fail the test.

## Test plan

- [x] `deno test` on `packages/runner/test/memory-v2-stacked-commit.test.ts`
      (72 passed)
- [x] `deno check` on the touched runtime/memory source files
- [ ] Address the blocking replay-exception finding above
- [ ] Full CI suite

---
## Summary by cubic
Replays mergeable patch ops against the rebuilt live base during pending optimistic replay to prevent resurrecting rejected data. If the live base can’t support an op’s path, the op is skipped (with a warning) instead of throwing.

- **Bug Fixes**
  - Added `operationBased` to patch-op descriptors in `@commonfabric/memory`; `append`/`add-unique`/`increment`/`remove-by-value` now replay by calling `applyPatch` on the rebuilt base, while snapshot-derived ops keep projection-based replay.
  - Apply operation-based ops one-by-one and skip any with incompatible ancestors, avoiding replica crashes and composing with independent projected patches.
  - Updated docs and added tests for no-resurrection, incompatible-ancestor skip behavior, and descriptor classification.

<sup>Written for commit 54bd8cac7d9335dd1019e360509272fd75513340. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5280?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ubik2:
Addressed the blocking finding from review: `applyPendingVersion`'s operation-based replay now applies each mergeable op individually and catches the incompatible-ancestor case (`path is not traversable`) instead of letting it throw out of the replica. Skips the op with a warning log rather than resurrecting stale data or crashing — see the new commit and its regression test (seeds a present-but-incompatible ancestor, a scalar `null`, distinct from the existing create-missing-from-absent regression).

Also rebased onto latest `main` (picks up #5279, now merged).

ubik2:
Converting to draft. There's a good chance other PRs in progress will make this redundant.

ubik2:
Superseded by #5521, which removed projection replay entirely: `applyPendingVersion` now replays every pending layer's wire ops over the delivered base via the shared `applyPatchToDocument`, so the no-resurrection property holds for all op kinds without the `operationBased` split. The blocking incompatible-ancestor finding from the #5258 review is also resolved there (`PatchApplyError` is caught and the layer renders skipped — whole-layer skip, preserving transaction atomicity in the optimistic view, where this PR's per-op skip could show a partially applied transaction). Test coverage for resurrection and skip semantics landed with #5521's rewritten stacked-commit pins and the v2-patch tests. The stale `changedPathsForPendingPatch` references in docs/features/patch-operations.md are being fixed separately.

PR #5296, opened 2026-08-03
feat(cli): warn when cf and the server run different commits

bfollington:
## TL;DR

`cf` now detects version skew against the server it's talking to. Every server-touching command compares cf's own commit with the server's `/api/meta` `gitSha`; when both are known and differ, it prints one warning to stderr:

```
⚠️  cf and the server are running different versions of Common Fabric.
    cf:     7cc8a407705e14b6e97832ca2192465d3e64cc49
    server: deadbeefdeadbeefdeadbeefdeadbeefdeadbeef (http://localhost:8000)
    Mismatched versions can fail in confusing ways; update whichever side is stale,
    or set CF_SKIP_VERSION_CHECK=1 to skip this check.
```

`CF_SKIP_VERSION_CHECK=1` (any non-empty value) skips the check entirely, including its fetch.

## Why commit SHAs, not version strings

The repo has no meaningful version strings — `cf --version` prints a hardcoded `0.0.1`, and the entire deploy identity is the commit SHA (the bastion's `deploy.sh` takes a 40-char SHA; `/api/meta` was built to report the deployed commit, with `TOOLSHED_GIT_SHA` / baked / `COMMIT_SHA` precedence already in place). So the SHA is the idiomatic join key, and the server side needed no changes at all.

## How each side knows its commit

| Side | Resolution |
|---|---|
| server | `/api/meta` `gitSha` (existing precedence, unchanged) |
| compiled `cf` | `packages/cli/COMPILED` marker baked at build time, mirroring toolshed's existing marker |
| source-run `cf` (`bin/cf`) | `git rev-parse HEAD` of the checkout the running module lives in — which is what matters in multi-checkout setups |

Either side unknown → skip silently: absence of metadata is not evidence of a mismatch. A compiled binary built without `COMMIT_SHA` stays unknown rather than guessing from the build machine's checkout.

To make the local story work out of the box, `start-local-dev.sh` now defaults `COMMIT_SHA` to the checkout's HEAD (explicit value still wins; per the docs it is descriptive metadata, not an update gate). So the two real-world skews both get caught: cf from checkout A vs local toolshed from checkout B, and cf vs a stale local server after a pull.

## Mechanics

- Check runs at the two CLI connection choke points (`loadManager`, acl's `createRuntime`), concurrently with the existing health check round-trip, awaited on both paths — no added wall-clock, no leaked fetch/subprocess ops when the connection fails.
- The `COMPILED` reader moved to `@commonfabric/utils/build-info`; toolshed re-exports it (its tests unchanged), the CLI shares it.
- Docs updated: `CONFIGURATION.md` (new env var + `COMMIT_SHA` default), `LOCAL_DEV_SERVERS.md`.

## Notes for review

- **Warn-only by design.** Talking to estuary from a fresh main checkout will legitimately warn while estuary trails main — that's signal (the exact skew that causes weird failures), but it's why this doesn't hard-fail. If it proves noisy we can later distinguish ancestor-drift from divergence, or gate on a protocol hash.
- A dirty checkout at the same HEAD as the server won't warn — SHA-level detection only.

## Verification

- New unit tests: `packages/cli/test/version-check.test.ts`; build-binaries tests extended for the second `COMPILED` marker.
- Full CLI unit suite (123 passed), toolshed build-info + meta tests, utils suite, tasks/build-binaries tests all green.
- Repo-wide `deno fmt --check` and `deno lint` clean under the pinned Deno 2.9.4.
- End-to-end exercised against a fake `/api/meta` server: warning fires with the checkout HEAD on the cf side; the env override silences it.

---
## Summary by cubic
Add a version-skew warning to the `cf` CLI: on each server call, `cf` compares its commit to the server’s `/_health` `x-cf-git-sha` and prints a graded warning when they differ. The check rides the health probe’s headers and adds no extra request; set `CF_SKIP_VERSION_CHECK=1` to disable.

- **New Features**
  - Warns when `cf` and the server run different commits, graded by direction:
    - cf newer → compact heads-up naming how far the server is behind (when countable).
    - cf older → loud OUTDATED warning.
    - Diverged/unknown order (including compiled binaries with no history) → undirected wording.
  - Unknown commits skip silently; `CF_SKIP_VERSION_CHECK` disables the check.
  - Toolshed adds `gitSha` to `/_health` and the `x-cf-git-sha` header; the runner captures the header as `runtime.serverGitSha` with generation-safe updates that complete at headers-arrival.
  - `scripts/start-local-dev.sh` defaults `COMMIT_SHA` to the checkout HEAD.

- **Refactors**
  - Shared COMPILED reader moved to `@commonfabric/utils/build-info`; toolshed re-exports; CLI uses a resolver that branches on `Deno.build.standalone` and validates the checkout to avoid vendored layout misreads.
  - Build writes a `COMPILED` marker for both toolshed and CLI and includes them in their binaries; divergence is only labeled when a common base is provable.

<sup>Written for commit 64ddc0ff77948b211619cd052096889726ce6d21. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5296?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Coverage baselines

The coverage-debt ratchet flags +5 (cli) / +3 (runner) uncovered lines. The bulk is the `Deno.build.standalone` compiled-only branch in `packages/cli/lib/build-info.ts`, which cannot execute under `deno test` by construction (it is the branch that distinguishes a compiled binary from exactly this situation), plus default-effect fallbacks (`console.error`, `Deno.env.get`) that the injectable-deps tests deliberately bypass to observe behavior.

ACCEPT_COVERAGE_DEBT: coverage-debt: packages/cli uncovered lines = 3625 lines
ACCEPT_COVERAGE_DEBT: coverage-debt: packages/runner uncovered lines = 5513 lines
ACCEPT_COVERAGE_DEBT: coverage-debt: packages/toolshed uncovered lines = 4666 lines


bfollington (commented):

PR #5296 · detail review of all 13 changed files · checks run: targeted CLI version/runtime tests, toolshed build-info/meta tests, build-binaries tests, changed-file fmt and lint.

## 🔴 Blocking

1. **Warn-only metadata can gate every command indefinitely** — `packages/cli/lib/version-check.ts:21-33`, `packages/cli/lib/piece.ts:392-397`, `packages/cli/lib/acl.ts:58-60` — **verified**. Running the work concurrently only avoids latency when both `/api/meta` and `git rev-parse` finish no later than `/_health`. A server can answer `/_health` promptly while `/api/meta` stalls, and both callers then await `versionCheck` before either succeeding or reporting the health result. Even normally, the critical path becomes `max(health, metadata, git)`, so the PR claim of no added wall-clock is not sound. The ordinary healthy/unhealthy boolean paths do not leak ops because the promise is awaited; the cost is that a diagnostic check becomes a command gate. Please make version metadata part of the existing health round trip, or make this a lifecycle-bound best-effort operation that is cancelled and never awaited on the command critical path. Add a controlled test where health settles and metadata remains pending.

2. **The supported vendored-Labs layout can report the host repository SHA** — `packages/cli/lib/build-info.ts:37-49` — **verified**. Worktrees and independent multi-checkouts are handled correctly because the command cwd follows `import.meta.url`. But `bin/cf` and the launcher explicitly support `<host>/vendor/labs`, and the checkout test only requires `packages/cli/launcher.ts`. If that vendored tree has no Git directory of its own, `git rev-parse HEAD` walks upward and returns the host commit, such as a Loom commit, which will falsely mismatch every Labs server. Verify that the discovered Git top-level is the Labs root before accepting HEAD; otherwise return unknown, or supply explicit vendored build metadata. Please cover both a real worktree and a vendored tree nested in another Git repository.

3. **Marker presence does not distinguish a compiled run from a source run** — `packages/cli/lib/build-info.ts:20-30,61-64`, `tasks/build-binaries.ts:458-468` — **verified**. `prepareWorkspace()` writes `packages/cli/COMPILED` before every binary build, including toolshed-only and bg-piece-service-only builds. A source `bin/cf` invoked while such a build is running therefore reports the build environment SHA, or null when it is empty, instead of the selected checkout HEAD. A hard-killed build leaves the same misclassification behind. Pinned Deno 2.9.4 exposes `Deno.build.standalone`; branch on that and only consume baked metadata when it is true. Add source and compiled resolution tests.

## 🟡 Improvement

- **The skip test does not prove that the override skips work** — `packages/cli/test/version-check.test.ts:48-54` — **verified**. The implementation currently has the intended semantics: every non-empty value, including `0` or `false`, returns before starting Git or fetch; empty/unset does not skip. But this test still passes if the early return is deleted because both failure paths are deliberately swallowed. Inject the resolver/fetch/warn effects and assert zero calls, and pin the documented non-empty versus empty behavior.

## Verified concern notes

- Defaulting `COMMIT_SHA` in `start-local-dev.sh` does not authorize system-pattern updates. The updater remains identity-gated and has no `COMMIT_SHA` dependency.
- The local shell now receives a descriptive `clientVersion`, but development still passes an explicit `/scripts/worker-runtime.js`, so it does not switch to the deployed immutable `/builds/<sha>` graph.
- Mismatch remains warning-only in exit-status terms, and `CF_SKIP_VERSION_CHECK` bypasses both Git and HTTP in the implementation. The first blocking finding is the remaining semantic violation: the warning can still block progress by latency.

**Verdict: changes requested before merge.** The feature direction is good, but the diagnostic path must not gate commands, and source identity must fail unknown rather than report an unrelated or temporary SHA.

bfollington:
All three blocking findings and the improvement are addressed in 5e2cd5fca:

**1. Metadata can no longer gate commands.** Took the review's first suggestion: the server's commit now rides the `/_health` response the health check already fetches. Toolshed adds `gitSha` to the health payload (same `resolveGitSha()` as `/api/meta`), and `Runtime.healthCheck()` captures it into `runtime.serverGitSha` during its existing round trip — there is no separate metadata request left to stall. The requested "health settles, metadata pending" test is moot under this shape (they are the same response); what remains awaited on the command path is only cf's own local resolution (baked metadata or one `git rev-parse`), settled on both the healthy and unhealthy paths. `fetchServerGitSha` is deleted.

**2. Vendored layouts report unknown, not the host repo's HEAD.** `gitShaForCliLibDir` runs `git rev-parse HEAD --show-toplevel` and only trusts HEAD when `<toplevel>/packages/cli/lib` realpath-equals the running module's directory. New tests cover a real checkout, a linked worktree, a vendored `<host>/vendor/labs` with no git directory of its own (where the host repo even has a decoy `packages/cli/lib`, to prove the guard compares locations rather than existence), and a bare directory.

**3. `Deno.build.standalone` is the compiled/source discriminator.** The marker file is only consulted when `standalone` is true, so a source `bin/cf` running during (or after an interrupted) binary build resolves from git as intended, and a stale leftover marker is inert. A compiled binary built without `COMMIT_SHA` still reports unknown rather than guessing.

**Improvement: the skip test now proves skipping.** `startVersionCheck` takes injectable env/resolver/warn effects; tests assert zero effect calls under the override (including `"0"` and `"false"` — any non-empty value skips), that empty/unset does not skip, and that a rejecting resolver is swallowed.

Verification: full CLI suite green under pinned Deno 2.9.4; runner `runtime-host-for-space` tests extended for the gitSha capture (including reset on a plain-text body from an older server); toolshed health/meta/open-api tests green; repo-wide fmt + lint clean; end-to-end run from this worktree resolves the worktree's actual HEAD and prints/suppresses the warning as expected.

bfollington:
Added in bba5cc367, following Ben's point that mismatching and being outdated are different problems: the warning is now **graded by direction**.

- **cf newer than server** (the very normal local-dev state — the checkout moved on, the server kept running or the deploy trails main): compact one-liner, names how many commits behind the server is, "usually fine".
- **cf older than server** (the dangerous direction — the server speaks a protocol this cf predates): loud `OUTDATED` warning, "commands will likely fail in confusing ways — update this cf".
- **Diverged** (sibling commits, e.g. a feature branch vs a newer deploy): named as diverged.
- **Unorderable** (server commit never fetched, shallow history, and every compiled binary — no history to consult): the original undirected wording.

Direction is proven with `git merge-base --is-ancestor` in the checkout's history, computed only after a mismatch is already established — at most three local git invocations, never on the happy path, and any git failure degrades to the undirected warning. All still warn-only and silenced by `CF_SKIP_VERSION_CHECK`; escalating the OUTDATED case to a hard failure later is a one-line change if we want it.

Live output from this worktree with the server pinned at `HEAD~1`:

```
⚠️  cf is newer than the server at http://localhost:8000 — the server (ec1e58338…) is 1 commit(s) behind this cf (5e2cd5fca…).
    Usually fine; restart or redeploy the server to match, or set CF_SKIP_VERSION_CHECK=1 to skip this check.
```

New ancestry tests run against real temp git repos (ahead/behind/diverged/unknown/non-repo); full CLI suite green.

bfollington (commented):

PR #5296 delta re-review: 5e2cd5fca + bba5cc367 after the review on ec1e58338 · deep-read of the health/runtime lifecycle, CLI build identity, ancestry grading, tests, and loadManager/createRuntime call paths · checks run: targeted CLI build-info/version/runtime-creation tests, runner host/health tests, toolshed health test, changed-file fmt/lint/diff-check, plus controlled body-stall, overlapping-health-check, and shallow-history repros.

## 🔴 Blocking

1. **The health ride-along still lets diagnostic metadata gate commands indefinitely** — packages/runner/src/runtime.ts:2244-2250; packages/cli/lib/piece.ts:394-398; packages/cli/lib/acl.ts:61-62 — **major, verified**. Moving gitSha onto /_health removes the second request, but it changes healthCheck from “fetch resolved / status available” to “the entire response body reached EOF and parsed as JSON.” Fetch resolves when headers arrive; Response.json() does not. I reproduced a 200 Response that enqueues the complete JSON value but leaves its stream open: healthCheck remained pending with serverGitSha null until the body was closed. Both CLI paths await that health check, so a stalled/truncated health body (or proxy) still stalls every command even though connectivity is already known. This also makes the module claim that metadata “can never gate a command on extra network work” false. Put gitSha in a response header and capture it synchronously once fetch resolves (the body field can remain for API consumers), or otherwise preserve the old health completion boundary. Add a controlled open-body-stream regression test.

2. **Shallow history can be falsely graded as diverged** — packages/cli/lib/build-info.ts:155-165 — **major, verified**. Exit 1 from merge-base --is-ancestor means “not provable from this repository’s available graph,” not proof that the commits truly diverged. I created A → B → C, shallow-cloned C, then fetched A as a separate depth-1 root. Both commits exist, both ancestry probes return 1, and relateShasIn(C, A) returns diverged even though A is an ancestor of C. That contradicts the code/docs promise that shallow history degrades to unknown and produces the strong, false message “Neither side contains the other’s commit.” Before returning diverged after two false probes, require a successful merge-base proving a common base in the visible graph; otherwise return unknown. Cover the separate-shallow-roots case.

3. **Overlapping health checks can restore stale metadata after the newer call finishes** — packages/runner/src/runtime.ts:2228-2249 — **major, verified**. The reset works for sequential calls, but each call mutates the same field from inside its fetch task without an invocation generation. In a controlled repro, call 1 had a held body, call 2 completed and set serverGitSha to second, then releasing call 1 overwrote it with first. That violates the getter’s “most recent healthCheck call” contract and can pair a caller with the wrong server SHA. Capture into call-local state and publish only if that invocation is still current (for example, with a monotonically increasing generation). Add the overlapping-call ordering test.

## 🟡 Improvements

- **Ancestry describes direction, not protocol compatibility severity** — packages/cli/lib/version-check.ts:70-86 — **minor, verified by inspection**. The mapping itself is correct in a complete graph: server ancestor → cli-ahead; CLI ancestor → cli-behind; known siblings → diverged; command failures above exit 1 → unknown. But “Usually fine” for a newer CLI can be actively misleading: a descendant CLI can call an endpoint or send a payload the ancestor server does not support. Conversely, an older CLI may remain compatible with a newer server. Keep the useful ahead/behind action guidance, but use symmetric “commands may fail” wording unless there is a protocol-compatibility signal stronger than commit ancestry.

## Verified resolution / load-path notes

- Prior finding 2 is resolved: Deno.build.standalone is the correct invocation discriminator, so a source run ignores an in-flight or stale COMPILED marker and a compiled artifact without baked COMMIT_SHA stays unknown.
- Prior finding 3 is resolved: rev-parse HEAD + --show-toplevel followed by realpath equality against <toplevel>/packages/cli/lib accepts ordinary checkouts and linked worktrees and rejects the supported vendored-without-own-git layout instead of borrowing the host repository SHA.
- I found no detached subprocess op on the ordinary loadManager/createRuntime healthy or false-health paths: the local resolution promise is settled by finish() before either path proceeds. The remaining unbounded operation is finding 1’s awaited body read. The new direction grade also adds mismatch-only critical-path latency after health: on this checkout, the ancestry + distance subprocesses took about 37 ms; matching/unknown pairs do not pay those relation calls. That is measurable but non-blocking if the richer warning is worth it.

**Verdict: changes still requested before merge.** The standalone and vendored-layout blockers are genuinely fixed, and the full-graph ancestry mapping is sound, but the health body still gates commands and the new metadata/ancestry edge cases can report stale or false conclusions.

bfollington:
Re-review findings addressed in 64ddc0ff7:

**1. Header-borne gitSha — the body can no longer gate anything.** The finding was right: `res.json()` waits for body EOF, which quietly changed the health check's completion boundary from headers-arrival to body-EOF. `/_health` now also sends `x-cf-git-sha` (body field kept for API consumers), and `Runtime.healthCheck` captures from headers only — probe and capture complete together. The requested regression test is in: a 200 whose body stream is never closed, proving `healthCheck` resolves and captures anyway.

**2. Diverged now requires proof.** `relateShasIn` only reports `diverged` after a successful `merge-base` shows a common base in the visible graph; two false ancestry probes without one degrade to `unknown` (undirected wording) instead of asserting "neither side contains the other" from an incomplete history. Covered with an orphan-root fixture modeling the disjoint-shallow-roots repro.

**3. Capture generations.** Each `healthCheck` call captures under a monotonically increasing generation and only the newest publishes, so a withheld earlier response can no longer overwrite a newer call's capture after the fact. Covered with an overlapping-call ordering test that releases the first response after the second call completes.

**Improvement taken:** "Usually fine" is gone — cf-newer now reads "Commands may fail where cf relies on newer behavior; restart or redeploy the server to match", keeping the direction signal and action without overpromising compatibility.

Full CLI suite, runner host/health tests, and toolshed health test green under pinned Deno 2.9.4.

PR #5302, opened 2026-08-03
feat(piece): verb-event-role compat rule for closed-world event schemas

mpsalisbury:
## Why

C5 (#5246) shipped dispatch-side closed-world enforcement, but the schema-generator half — emitting `additionalProperties: false` on event schemas — was backed out because the update gate refused it for all 28 baseline groups with argument-side verbs. Berni decided the fork on #5246: a verb-event-role rule taking the position that accepted-and-STRIPPED was never contract (the runtime never delivered an undeclared event field to any handler). This PR is that rule, plus the plan-recorded forward-skew policy the emission will land under. The rule is inert until emission (follow-up PR): no schema changes here, and the fleet gate still records 336 patterns updatable from every contract.

## What Changed

- `packages/piece/src/schema-compatibility.ts`: a `verbEvent` compatibility context, entered where **both** contracts mark a node `asCell: ["stream"]` and carried through the event subtree. Below it, boolean↔boolean `additionalProperties` transitions are free in both directions — open→closed is the migration; closed→open keeps the `never`-derived closure cleanup (Berni's note) from reading as a break. Deliberate boundaries: schema-valued `additionalProperties` still compares on either side (the extras' shape is a data contract even on an event); a one-sided marker is never exempt (stays an `asCell changed` refusal); closed events remain evolvable by optional fields under the existing evolution policy.
- `packages/piece/test/schema-compatibility.test.ts`: eight pins covering every boundary above, both marker forms (`$ref`-marked and inline), and the nested-subtree propagation.
- Plan doc: the Risks entry's open skew question is replaced by recorded policy — **reject-on-collision, loud, local-first, id-preserving**: the CLI already validates against the deployed schema so the common path fails locally with the id unspent; a payload reaching dispatch is refused through the thrown-handler path with the id preserved, so recovery is a same-id re-invoke after the piece updates. No deploy ordering, no versioning machinery. Rationale recorded: skew-window stripping was the same silent-loss defect class rule 1 exists to kill.

## Test plan

- `deno task test` in `packages/piece` — 32 passed (378 steps), including the new "verb event closed-world transitions" block
- `deno task pattern-compat` — 336 patterns updatable (rule inert on current schemas)
- `deno task check-docs`, `deno fmt --check`, `deno lint` — clean

---
## Summary by cubic
Introduce a verb-event compatibility rule that allows boolean `additionalProperties` to move open↔closed under stream-marked verb events, unblocking closed-world event schemas without breaking contracts. Docs record a reject-on-collision skew policy; behavior is unchanged until generator emission.

- **New Features**
  - Added `verbEvent` context in `packages/piece/src/schema-compatibility.ts`, entered only when both contracts mark a node `asCell: ["stream"]`.
  - Within the event subtree, boolean `additionalProperties` transitions are free in both directions; schema-valued `additionalProperties` still compares.
  - No exemption on one-sided stream markers; propagation covers nested event objects. Closed events can still add optional fields.
  - Tests in `packages/piece/test/schema-compatibility.test.ts` pin all boundaries and marker forms. Plan doc updated with the recorded policy.

- **Migration**
  - No action now; schemas are unchanged. Rule activates when the generator emits `additionalProperties: false`.
  - Forward-skew policy: reject-on-collision, loud, local-first, id-preserving (CLI validates against deployed schema; dispatch rejects with the same id).

<sup>Written for commit 35e8e47d23ddb2b7e292368a77bf7d9383978df6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5302?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/plans/pattern-verb-contract-implementation.md
+++ b/docs/plans/pattern-verb-contract-implementation.md
@@ -1097,21 +1097,44 @@ change that alters them.
   review: the closed shapes already in argument-side baselines likely
   derive from `never`-typed events that were never meant to assert "only
   empty events are valid" — clean that up in the generator eventually;
-  until then the exemptions carry the cost. The chosen rule must still
-  weigh the
-  FORWARD skew case the stripped-was-never-contract argument does not cover:
-  adding an optional event field is legal argument widening, but during any
-  deploy window a newer client sends that field to a still-running older
-  piece whose closed schema does not declare it — today stripped and
-  harmless, under closed-world a hard reject at dispatch. Mixed-version
-  callers are exactly the live board's incident history (the cross-version
-  write storms, the legacy-field mirroring `topics` still carries), so the
-  decision needs a story for widened payloads against not-yet-updated
-  pieces — deploy ordering, a skew-tolerant reject-only-on-collision rule,
-  or an explicit versioning step — before emission lands. Until the rule
-  and its skew story land, generated event schemas stay open and
-  closed-world enforcement is per-schema opt-in at dispatch (C5's runner
-  gate).
+  until then the exemptions carry the cost. **The rule is built**: a
+  `verbEvent` compatibility context in
+  `packages/piece/src/schema-compatibility.ts`, entered where BOTH
+  contracts mark a node `asCell: ["stream"]` (the exemption never applies
+  across a one-sided marker — that stays an `asCell changed` refusal) and
+  carried through the event subtree. Below it, boolean↔boolean
+  `additionalProperties` transitions are free in both directions — closed
+  events also remain evolvable by optional fields under the existing
+  evolution policy, and closed→open stays free so the `never`-derived
+  closure cleanup never reads as a break — while schema-valued
+  `additionalProperties` still compares on either side: a constraint on
+  the extras' SHAPE is a data contract even on an event. Each of those
+  properties is pinned in `schema-compatibility.test.ts` ("verb event
+  closed-world transitions").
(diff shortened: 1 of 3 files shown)


mpsalisbury:
@seefeldb — merged as the inert half of the emission work you unblocked on #5246. One paragraph deserves your eyes before the emission PR (which builds on it) merges: the plan's Risks entry now records the forward version-skew POLICY in place of the open question — **reject-on-collision, loud, local-first, id-preserving**. A newer client's legally-widened payload against a not-yet-updated closed piece is rejected during the deploy window, on the grounds that (1) the old stripped-and-ran behavior meant the caller believed semantics it didn't get — the silent-loss class rule 1 kills; (2) the CLI validates against the DEPLOYED schema pre-dispatch, so the common path fails locally with the invocation id unspent; (3) a payload reaching dispatch is refused through the thrown-handler path, id preserved, so recovery is a same-id re-invoke after the piece updates. No deploy ordering, no versioning machinery. If that holds for you, the emission PR lands under it; if not, it's a doc amendment before anything ships.

PR #5307, opened 2026-08-04
feat(schema-generator,ts-transformers): emit closed-world verb event schemas (C5 second half)

mpsalisbury:
## Why

C5 (#5246) shipped the dispatch gate but with nothing emitting `additionalProperties: false`, closed-world enforcement was a per-schema author opt-in — design rule 1's "an undeclared field is a typed rejection, never ignored" held only where someone hand-closed a schema, and the silent-strip failure class (the live board's discarded-`agentName`) stayed open everywhere else. #5302's verb-event-role rule made the open→closed transition legal; this PR does the emitting. **DO NOT MERGE until Berni nods at the skew policy** ([asked on #5302](https://github.com/commontoolsinc/labs/pull/5302#issuecomment-5172941621)) — the plan records emission landing under reject-on-collision, and his answer should precede the ship.

## What Changed

- **`closeVerbEventRoot`** (`packages/schema-generator/src/event-closure.ts`): the one closure rule, position-scoped and guarded — inline object roots gain the keyword directly; `$ref` roots gain it beside the reference so a shared def stays open at data-position use sites; anything already declaring `additionalProperties` (index signatures = the author's organic opt-out), unions, `never`-derived shapes, and booleans are untouched.
- **Surface 1 — handler `$event`** (what the dispatch gate reads): `SchemaGeneratorTransformer` stamps the generated literal when a `toSchema` call is the first argument of a handler-family call. Membership is the provenance: hand-authored schema literals in that position are never stamped.
- **Surface 2 — `Stream` property schemas** (what baselines record and `cf piece verbs` advertises): the Stream wrapper branch applies the same helper.
- `stream-result.test.ts` pins the closed side (plus $ref-position-scoping and the index-signature opt-out) where it pinned the open side; mapping spec §6.5 and behavior spec §12 rewritten; plan WS-C + Risks record the landing.
- 26 transformer fixture goldens regenerated (churn audited: only root closures and comma reflow); **64 fleet baselines re-recorded with zero compatibility refusals** — the #5302 rule carried the entire migration.

## Test plan

- schema-generator 55 ✓, ts-transformers 1132 ✓, runner 1127 ✓ (llm-dialog carve-out holds), piece ✓, CLI 123 ✓, topics pattern suites 53 ✓
- `deno task pattern-compat` — 336 patterns updatable, 0 refusals
- `deno task check-docs`, repo-wide `deno fmt --check` + `deno lint` ✓

---
## Summary by cubic
Emit closed‑world event schemas for verbs in `schema-generator` and `ts-transformers`. Event roots now include `additionalProperties: false`, so undeclared payload fields are rejected at the dispatch gate instead of being silently dropped.

- **New Features**
  - Added `closeVerbEventRoot` to close object roots; stamps beside `$ref` and skips schemas that already set `additionalProperties`, unions/never/booleans, and hand-authored literals.
  - Applied to two surfaces: pattern `Stream` property schemas and transformer-injected handler `$event` when `toSchema(...)` is the first handler arg.
  - Updated docs (mapping §6.5, behavior spec §12) and tests; regenerated transformer goldens and re-recorded baselines.

- **Migration**
  - No code changes required; 64 fleet baselines re-recorded with zero compatibility refusals.
  - To keep an event open, declare an index signature or set `additionalProperties` on the event type explicitly.

<sup>Written for commit 9fc61cddc792eaab1fa38541b7887fadca8530c2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5307?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

## Custody and inherited context

Sequencing moved to the consolidated verbs plan (#5519); custody of rebases,
review responses and re-verification transferred from the author. **Merging
stays with Mike.** Recorded here because it is expensive to rediscover.

**State at handover:** gate-clear and current on main, both touched-package
suites plus `pattern-compat` green. Held for confirmation of the recorded
forward-skew policy — the one question the reviewer named open, answered by
recorded policy rather than by him.

### The skew fallback, if the answer is not "confirmed"

The load-bearing fact: **emission and enforcement are separable.** The schemas
record the contract; collision behavior lives in one function —
`closedWorldEventRejection` in `packages/runner/src/runner.ts`, one call site,
pinned by `scheduler-event-receipts.test.ts` and the CFC integrity suites. So:

- **Confirmed** → merges as-is.
- **Wants skew tolerance** — a version window, reject-only-on-non-widening,
  whatever is specified → this still merges unchanged and the answer lands as a
  gate amendment in that one function. No fleet churn, no golden churn. Worth
  raising in that conversation: the policy's local-first leg rests on the CLI
  pre-dispatch gate validating against deployed schemas, so a concern about
  non-CLI callers belongs in the gate rather than in emission.
- **Full retreat** — no closure until versioning machinery exists → revert the
  two stamps, regenerate goldens, re-record baselines (closed-to-open is free by
  design, for exactly this case), flip the pins back, restore the two spec
  sections. All mechanical. #5302 stays in every branch — inert without
  emission, and needed by any future attempt.

### Re-record lore

1. **Verify, then record.** After any rebase run plain `pattern-compat` and read
   the verdict before `--update`. Zero refusals is the invariant; a refusal is a
   semantic regression that `--update` would bury as a fresh baseline.
2. **Do not stack this PR's own baselines.** The 64 baseline files it adds are
   append-only timestamped JSONs. If contracts change again, delete this PR's
   additions and re-record rather than stacking — a contract hash that never
   existed on main should not persist.
3. **Audit golden churn by frequency, not by eye:**
   `git diff | grep '^+' | sort | uniq -c`. The legitimate signature is exactly
   two shapes — `additionalProperties: false` insertions and comma-reflowed
   `asCell: ["stream"],` lines. A third shape in the histogram is a real
   behavior change.
4. **The provenance guard is positional.** The transformer stamps only
   `arguments[0]` of handler-family calls. Any future re-ordering of handler
   arguments silently un-stamps, and the tell is fixture goldens losing the
   keyword — so never hand-edit the `schema-generation-handler-*` or
   `handler-schema/*` expected files.

### One inheritance that rides along

The `llm-dialog` carve-out (`runtimeInjectedEventKeys`) is what keeps the CFC
suites green under closed schemas. Its retirement is decided but unclaimed. It
blocks nothing here; the recorded direction is to deliver through receipts, after
which the marker machinery goes.

### Reconciliation with #5501

No ordering edge — a declared result rides a trailing options argument while
event stamping keys on `arguments[0]`. Whichever of the two lands second owes a
golden regeneration.

### DIFF
--- a/docs/plans/verbs-implementation.md
+++ b/docs/plans/verbs-implementation.md
@@ -183,8 +183,9 @@ holds one that moves.
 anything reactive gets none, which is what item 1 decides.
 
 **9. Closed-world event schemas and listing marks.** #5307 and #5309, from the
-verb contract arc. Both unblocked; both wanted a courtesy review rather than a
-gate.
+verb contract arc. #5309 is unblocked and wants a courtesy review rather than a
+gate. #5307's emission is written and its update gate is clear; item 11 is the
+question it now waits behind.
 
 **10. Listing rows carry a handler's declared result.** *(S)* `cf piece verbs`
 already reports an `outputSchema` per row, and the type says why it is empty for
@@ -199,6 +200,38 @@ The two share the words and nothing else, which is why the listing work and the
 declared-result work read as coupled when they are independent. Only this item
 joins them.
 
+**11. A DOM event is a caller too.** *(M)* The dispatch gate reads the
+transformer-injected `$event` schema, and that is the schema a browser event
+arrives against: the renderer serializes a DOM event into the payload and sends
+it to the stream. Closing that root judges the whole envelope, and the envelope
+is nothing a pattern's event type describes — `type`, `provenance`, the
+allowlisted key/modifier/button properties, a `MouseEvent.detail` that is a
+click count rather than a `CustomEvent`'s payload, and a `target.value` whose
+type is the element's, not the author's. An open root lets the schema-shaped read
+deliver the declared subset and ignore the rest; a closed one makes each of those
+a rejection.
+
+The runtime already has the mechanism for exactly this: `runtimeInjectedEventKeys`
+(`packages/runner/src/cell.ts`), mint-gated by `markRuntimeInjectedEventKeys`, is
+how the LLM tool-call path hides its injected `result` from the gate. Only that
+one path mints today. The choices are to mint the renderer envelope the same way
+— which strips an undeclared envelope key and leaves a declared one to be judged,
+so it also wants the serializer to stop carrying a non-`CustomEvent` `detail` —
+or to close only the pattern-level `Stream` property schema that `cf piece verbs`
+publishes and `piece call` validates, and leave the dispatch surface open. The
+first keeps design rule 1 whole; the second is smaller and gives up dispatch
+enforcement for DOM-wired verbs.
+
+Closure also turns on full payload validation, not just the undeclared-key check:
+against a closed root, a payload missing a required field or carrying one of the
+wrong type is a rejection too. That reaches a probe.
+`packages/patterns/integration/time-capability-full.test.ts` fires `.send({})` at
+every result stream to make handler-context clock reads run, and tolerates the
+throw from a key that is not a stream. A dispatch-time rejection is not that
+throw — it arrives after the call returns — so the probe would report success
+while covering nothing. Whichever answer item 11 takes, that probe needs to fail
+loudly on a rejected send.
+
 ## Ordering
 
 | Item | After | Why |
@@ -206,7 +239,10 @@ joins them.
 | 6 (stack) | — | the stack merges bottom-up: #5458 → #5470 → #5497 → #5500 → {#5504, #5505} |
 | 6 (#5459) | — | not in the stack; entity-URI intake stands alone and merges whenever |
 | 7 | — | independent of 6; must precede 4 |
-| 8, 9 | — | independent |
+| 8 | — | independent |
+| 9 (#5309) | — | independent |
+| 11 | — | independent; decides what closure means at dispatch |
+| 9 (#5307) | 11 | emission is written, but what it enforces is 11's answer |
 | 2 | 6 | changes what the flags accept |
 | 3 | 6 | completes the suppression property |
 | 4 | 7 | publishes an address, so the address must have stopped moving |
--- a/docs/specs/schema-generator/ts_to_json_schema_mapping.md
+++ b/docs/specs/schema-generator/ts_to_json_schema_mapping.md
@@ -408,22 +408,35 @@ named types still hoist, as the fixture shows. The ts-transformers behavior
 spec §12 uses the same single-`asCell` vocabulary. Any doc claiming `Reactive`
 emits `asCell: ["opaque"]` is wrong on this tree.
 
-### 6.5 Stream event schemas — deliberately open (C5)
+### 6.5 Stream event schemas — closed-world (C5)
 
 A stream property's schema object is the verb's **event** schema — what a
 caller sends, which `cf piece verbs` publishes and `piece call` validates
-payloads against. It carries no `additionalProperties` of its own — only
-what the event type itself demands (an index signature, a `Record` value
-type). The verb contract wants event schemas closed-world
-(`additionalProperties: false` — an undeclared field is a rejection, never
-ignored), but emitting that is blocked on a pattern-update-gate migration:
-the argument-role compatibility rule refuses the open→closed direction for
-verbs reachable through a piece's argument schema (plan
-`docs/history/plans/pattern-verb-contract-implementation.md`, WS-C and Risks).
-The
-open event side is pinned as a decision in `test/stream-result.test.ts`; a
-schema that declares the closure by hand is enforced at dispatch by the
-runner (C5).
+payloads against. Its root is emitted **closed**: `additionalProperties:
+false`, so an undeclared field in a call payload is a typed rejection at
+dispatch, never silently stripped (verb contract design rule 1; the runner
+gate is C5's `closedWorldEventRejection`; the update-gate transition is
+legal via the verb-event-role compatibility rule, #5302). The stamp is
+`closeVerbEventRoot` (`src/event-closure.ts`), applied by the `Stream`
+wrapper branch, and it is position-scoped and conservative:
+
+- An inline object event gains the keyword directly; a `$ref` event gains
+  it BESIDE the reference (sibling keywords apply conjunctively), so a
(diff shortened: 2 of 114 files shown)


mpsalisbury:
Parked pending #5589 — a ruling from @seefeldb / @danfuzz on whether a closed verb schema governs the renderer's DOM-event envelope.

The minting half of this PR is built and correct, and the multi-runtime shard that was failing on `additional property provenance` now passes — that one *was* the minting. What remains red is not: `MouseEvent.detail` (a click count) and `target.value" (a cell-ref under a `cf-select` $value binding) land in slots patterns **declare**, and minting deliberately does not exempt a declared key. That is a renderer-semantics call, not an implementation one.

Item 11 in the plan names only the `detail` half; `target.value` is the same shape and isn't in the plan at all. Both need correcting once #5589 rules.

mpsalisbury:
Baseline note, so it isn't lost: this branch carries **two of its own baseline generations** for the same five patterns — `20260807T235141Z` (from 74766112f) and `20260810T184538Z` (from 242ab3155). Neither is on main. Append-only is satisfied and the gate is green, which is why it reads as fine.

Merging as-is would freeze an intermediate contract that never existed on main, recorded before the mint existed, and future compat runs would judge against it forever.

**Deliberately not fixing it yet.** If #5589's ruling changes the serializer, event schemas change and these want re-recording again — so deleting and re-recording now would just stack a third generation. Plan is one deletion and one re-record, as the last step before merge, once the shape is final.

Safe when the time comes: `check-baselines-append-only` diffs against `origin/main`, so a file added and then removed on the same branch never appears as a deletion.

mpsalisbury:
**Renumber owed before this merges: item 11 → item 12.**

This branch adds `**11. A DOM event is a caller too.**` to `docs/plans/verbs-implementation.md`, plus an ordering row keyed `11`. #5617 adds a *different* item 11 — `**11. A caller may name a reference.**` — with its own `11` ordering row.

The insertion points are disjoint (this one lands after item 10's terminology paragraph; #5617's goes immediately before `## Landed, and what consumes it`), so **git merges both without a conflict**: two item 11s and two ordering rows, silently. Same shape as the archive-rebase trap in the plan's own operating rules — the dangerous version is the one with nothing to warn you.

Agreed with @b1's session that #5617 keeps 11, because it is ready and this PR is parked on #5589 awaiting a renderer-semantics ruling. So the renumber is owed **here**, not there.

Two edits when this unparks: the item heading `11.` → `12.`, and the ordering row's `11` key (and the `9 (#5307) | 11` dependency row that points at it). Recording it on the PR because it is the kind of obligation that lives in one session's head and dies with it.

seefeldb (changes requested):
do not add addtionalProperties: false to all the event schemas, that is not how additionalProperties is meant to be used.

seefeldb:
i think i understand the idea here, which i guess is to make `cf call` fail if the passed in data has additional fields? let's do that inside `cf` as its an ergonomic feature of that call, not something we should enforce deeper in the runtime where it was all kinds of other issues.

fwiw, typescript already errors in a useful way in patterns, without foreclosing future updates that add fields or generating issues with system-added fields, so ergonomics-wise we're already good in typescript/pattern land.

mpsalisbury:
Closing: **closed-world event emission is ruled against**, so there is nothing here to land.

## The ruling

@seefeldb, on #5589 and in review here:

> don't close the schema with `additionalProperties: false`, it will forever lock down that schema. handlers only get the fields that are declared, unless they explicitly say `additionalProperties: true`. absence means they don't get the fields. `false` means that they get nothing if anything extra was declared, which they wouldn't have gotten anyway. Very rarely what we need.

That is a stronger answer than the three outcomes this PR was costed for. Its recorded fallback assumed a retreat *pending versioning machinery* — premature, revisit later. The actual objection is semantic: absence already denies a handler an undeclared field, so `false` buys no behaviour and adds a permanent lock, since the compatibility rules cannot take a closure back. There is no later at which this becomes right.

## Why closed rather than retreated

The custody notes describe a full retreat as reverting the two stamps, regenerating goldens, re-recording baselines, flipping the pins and restoring the two spec sections. Applied here that empties the branch: of 100 files, 66 are baselines and 15 are transformer goldens, and every one of them exists to record the emission. Retreating and closing are the same outcome, and closing says so honestly.

**No cleanup is owed.** The two baseline generations this branch carried (`20260807T235141Z` and `20260810T184538Z`) were never on main, so the contract hashes that never existed there simply never arrive. The delete-and-re-record that was deliberately deferred to just-before-merge is now moot rather than outstanding.

## What survives, and where it went

**The capability was not refused — only its location.** From the review:

> i think i understand the idea here, which i guess is to make `cf call` fail if the passed in data has additional fields? let's do that inside `cf` as its an ergonomic feature of that call, not something we should enforce deeper in the runtime… fwiw, typescript already errors in a useful way in patterns, without foreclosing future updates that add fields.

So the silent-strip failure this PR was built to close is still worth closing; the check belongs in the CLI, for callers who never see a type. That is now **item 12** in `docs/plans/verbs-implementation.md`, recorded as item 2's shape one surface over — item 2 refuses an unrecognized *projection* key, item 12 refuses an undeclared *event* field, and both are the CLI declining to refuse what it cannot honour rather than a schema forbidding it. `verbInputSchemaError` already validates the payload against the declared event schema, so it needs no new source of truth.

## Two mechanisms on main are left with nothing to trigger

Recorded rather than acted on, because they are runtime-owned:

- **`closedWorldEventRejection`** (`packages/runner/src/runner.ts`) — a dispatch gate that no schema can now fire, in exactly the layer the ruling says not to enforce in.
- **#5302's verb-event-role compatibility rule** (merged) — it governs an open-to-closed transition that will not happen.

## One thing this leaves open

Design rule 1 — *"an undeclared field is a typed rejection, never ignored"* — is what this PR implemented. The ruling makes silent-strip the intended runtime behaviour, with rejection an ergonomic property of `cf`. Wherever that rule is written down, it now says something the runtime does not do.

The branch stays; nothing here was wrong, and the measurement in it (64 fleet baselines re-recorded with zero compatibility refusals) is the evidence that #5302's rule would have carried the migration, should anyone ever revisit.

PR #5311, opened 2026-08-04
feat(cli): `piece setsrc --check` — compatibility preflight for a source swap

bfollington:
## Why

`cf piece setsrc <main>` replaces the pattern source of a **live** piece. Until
now the only way to learn whether a source could be applied was to attempt it,
and what came back was whichever low-level rejection surfaced first — a
schema-subset assertion, a retained-link proof, an argument validation failure,
or a CFC envelope rejection at the setup-commit boundary. Fix one, meet the
next.

That class cost a production incident on estuary: home roots and profile pieces
pinned to patterns that could not migrate their existing documents. #4967 and
#4977 added runtime *repairs*. This is the other half — knowing **beforehand**
whether a given source can be applied to a given piece.

## What Changed

`PieceController.checkPattern()` answers that without touching the piece;
`setsrc` grows `--check` to expose it, exiting non-zero when refused so it can
gate a deploy script. The report names every reason at once instead of the
first one to throw. (It is not a pure read: compiling the candidate persists
content-addressed compiled modules and source docs, per CT-1623 — but the piece
itself is never changed.)

The design constraint is **no second copy of the rules** — a preflight that
reimplements compatibility drifts from enforcement and becomes a liar. So the
check drives `pieceSourceCompatibilityReview()`, the review `changeSource`
already uses for detach/restore/follow, which runs the real
`assertPatternSchemasBackwardCompatible`, the real argument validation against
the candidate schema, and the real retained-link assertion. To that review this
PR adds a CFC envelope check: a dry-run of the real `mergeCfcSchemaEnvelopes`
against the piece's stored argument envelope, with an unreadable envelope
treated as a blocker. This is the class behind the estuary incident — type-level
checks all pass while the stored documents cannot migrate — and the test suite
proves the false green: a swap every schema check clears is rejected by the
real commit path with `CFC enforcement rejected commit`, and the check now
predicts it.

**`setPattern` itself is unchanged from `main` — byte for byte, on purpose.**
An earlier revision ran the aggregate review ahead of the swap, and that broke
CT-1917 cold-argument deferral: the review materializes and validates the
argument, so a deferred/unreadable argument validates as `undefined` and the
swap is refused where the apply path deliberately defers. The check is an
advisory path beside enforcement, not in front of it.
`packages/piece/test/setsrc-cold-argument.test.ts` pins this (verified red
against the broken revision).

## What the check guarantees — and what it does not

The contract is agreement on **verdict and stated cause**: a source the check
refuses is refused by the apply path, for the same reason. Two honest limits:

- **Result envelopes are not checked.** The CFC check covers the stored
  *argument* document only. The piece's own result document merges at commit
  under `generatedOutputPaths` that are unknowable without executing the
  module, so a weakened-ifc rejection on the result envelope can still pass
  the check. Covering it would require executing the candidate, and guessing
  instead would false-red.
- **The answer is point-in-time.** The apply path revalidates everything; a
  concurrent change between check and apply is caught there, not here.

## Supersedes #5059

#5059 was written before the shared review landed on main, and had grown a
parallel 416-line verdict module with its own vocabulary that now duplicates
it. Landing it would have meant two compatibility verdict systems.

~2,795 lines there become ~1,400 here, of which ~230 are executable production
code — the rest is tests (68%) and documentation, because main already had the
machinery; it just was not wired to the `setsrc <file>` path.

## Test plan

`packages/piece/test/pattern-compatibility-check.test.ts` pins the contract
that keeps a preflight honest: a source the check refuses is refused by
`setPattern` with the same stated cause (the CFC apply path rejects with a
non-Error object, so cause is compared as text). Plus: a compatible source
clears; a refused check leaves the piece running its old source and still
accepts a later compatible one; the widened-envelope case that used to false
green now reports `confidentiality cannot be weakened`; a merge that succeeds
is not blamed; an unreadable envelope is a blocker.

`packages/piece/test/setsrc-cold-argument.test.ts` pins CT-1917: a swap whose
stored argument is cold (never-written doc) still applies.

The CLI case pins that `--check` resolves the same piece and entry the apply
would use, and applies nothing. `packages/cli/integration/integration.sh`
drives the real binary through `--check` against a live toolshed: refusal
names the verdict and rule, exits 1, leaves the piece identity unchanged; a
compatible source clears (full suite verified locally, exit 0).

## Coverage

Two overrides, both for lines that are structurally unreachable from the
counted profile rather than untested behaviour. Everything reachable is tested:
`checkPiecePattern` runs for real with injected deps, the refusal is asserted
with its exit code, and the preflight/apply agreement is asserted on verdict
and cause.

`ACCEPT_COVERAGE_DEBT: coverage-debt: packages/cli uncovered lines = 3636 lines`

The new lines are a cliffy `.option()` declaration and the action closure that
renders its result. Every neighbouring command declaration in the same file —
`getsrc`'s options, `setsrc`'s existing flags — is equally uncovered on `main`:
this surface is exercised by `packages/cli/integration/integration.sh`, which
the coverage profile excludes by design. Rather than leave it untested, this PR
adds integration assertions that drive the real binary through `--check`
(verified locally against a toolshed: full suite exit 0), which is how the rest
of the CLI's command surface is covered.

`ACCEPT_COVERAGE_DEBT: coverage-debt: packages/piece uncovered lines = 166 lines`

Two defensive guards: `candidateRef === undefined`, mirroring the identical
guard `changeSource` already has a few lines above (reaching it needs a
compiled pattern carrying no artifact ref, which no supported path produces),
and a `!pattern` early return on the same shape. The third line the gate
originally flagged here — the CFC branch where the stored envelope does not
cover the candidate but the merge succeeds anyway — is now covered by a real
test rather than an override.

---
## Summary by cubic
Adds `cf piece setsrc --check` to preflight a source swap and report a clear compatibility verdict without updating the piece; it exits non-zero on refusal while the apply path stays unchanged.

- **New Features**
  - CLI: `piece setsrc --check` checks the same piece, entry, and pinned program the apply would use; on refusal it writes a plain verdict to stderr and exits 1 (no usage dump); success prints a short summary.
  - API: `checkPiecePattern(config, entry)` and `PieceController.checkPattern(program)` return a `PatternCompatibilityReport` with `compatible`, `issues` (`schema`, `argument`, `retainedLinks`, `cfc`), `message`, and the candidate `{ identity, symbol }`.
  - CFC: Adds a dry‑run merge of the candidate’s argument schema with the piece’s stored CFC envelope using `loadStoredCfcEnvelope`/`cfcSchemaMergeIssue`; unreadable envelopes block; result envelopes are intentionally not checked.

- **Refactors**
  - Apply path: `setPattern` enforcement is unchanged and does not run the aggregate review ahead of the swap, preserving CT‑1917 cold‑argument deferral; execute‑time validators still enforce.
  - CLI: refusal handling moved into `checkPieceSourceFromCommand` so exit code and message are testable; the action only renders the summary.
  - Runner: introduced `loadStoredCfcEnvelope` and shared it with the commit path; removed the unused options parameter from `cfcSchemaMergeIssue`.

<sup>Written for commit 21a126f232b9d7f8f6c2c7b8bc0ec2b6b408c9ee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5311?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cli/commands/piece.ts
+++ b/packages/cli/commands/piece.ts
@@ -1041,22 +1041,16 @@ export const piece = new Command()
   .action(async (options, mainPath) => {
     setQuietMode(!!options.quiet);
     if (options.check) {
-      const { config, report } = await checkPieceSourceFromCommand(
+      // A refusal throws out of here with a non-zero exit, so `--check` gates
+      // a deploy script as well as informing a person.
+      const { config, summary } = await checkPieceSourceFromCommand(
         options,
         mainPath,
       );
-      if (report.compatible) {
-        render(`${mainPath} can replace the source for piece ${config.piece}`);
-        hint(cliText(`NEXT STEPS:
+      render(summary);
+      hint(cliText(`NEXT STEPS:
   → Apply it: cf piece setsrc --piece ${config.piece} ${mainPath} ...`));
-        return;
-      }
-      // Non-zero exit: `--check` is meant to gate a deploy script, so the
-      // verdict has to be readable by `if` as well as by a person.
-      throw new ValidationError(
-        `${mainPath} cannot replace the source for piece ${config.piece}:\n${report.message}`,
-        { exitCode: 1 },
-      );
+      return;
     }
     const pieceConfig = await setPieceSourceFromCommand(options, mainPath);
     render(`Updated source for piece ${pieceConfig.piece}`);
@@ -1931,18 +1925,36 @@ export interface CheckPieceSourceCommandDependencies {
  * Deliberately parses the same options `setsrc` does, so the check is aimed at
  * the piece and entry the apply would have used — a preflight against a
  * different target is worse than none.
+ *
+ * The refusal is raised here rather than in the command's action so that the
+ * decision — including the non-zero exit that lets `--check` gate a deploy
+ * script — is reachable by a test. The action only renders what it returns.
  */
 export async function checkPieceSourceFromCommand(
   options: PieceCLIOptions,
   mainPath: string,
   deps: CheckPieceSourceCommandDependencies = {},
-): Promise<{ config: PieceConfig; report: PatternCompatibilityReport }> {
+): Promise<{
+  config: PieceConfig;
+  report: PatternCompatibilityReport;
+  summary: string;
+}> {
   const config = parsePieceOptions(options);
   const report = await (deps.checkPiecePattern ?? checkPiecePattern)(
     config,
     localPatternEntry(mainPath, options),
   );
-  return { config, report };
+  if (!report.compatible) {
+    throw new ValidationError(
+      `${mainPath} cannot replace the source for piece ${config.piece}:\n${report.message}`,
+      { exitCode: 1 },
+    );
+  }
+  return {
+    config,
+    report,
+    summary: `${mainPath} can replace the source for piece ${config.piece}`,
+  };
 }
 
 /** Apply the parsed `piece setsrc` command while preserving its safety flag. */
--- a/packages/cli/commands/piece.ts
+++ b/packages/cli/commands/piece.ts
@@ -1035,7 +1035,7 @@ export const piece = new Command()
   )
   .option(
     "--check",
-    "Report whether the source could replace the piece's current one, and change nothing.",
+    "Report whether the source could replace the piece's current one, without updating the piece. Exits non-zero when it could not.",
   )
   .arguments("<main:string>")
   .action(async (options, mainPath) => {
--- a/packages/cli/commands/piece.ts
+++ b/packages/cli/commands/piece.ts
@@ -1041,8 +1041,9 @@ export const piece = new Command()
   .action(async (options, mainPath) => {
     setQuietMode(!!options.quiet);
     if (options.check) {
-      // A refusal throws out of here with a non-zero exit, so `--check` gates
-      // a deploy script as well as informing a person.
+      // A refusal exits 1 from inside the check (plain stderr, no usage
+      // dump), so `--check` gates a deploy script as well as informing a
+      // person.
       const { config, summary } = await checkPieceSourceFromCommand(
         options,
         mainPath,
@@ -1917,6 +1918,8 @@ export interface SetPieceSourceCommandDependencies {
 /** Injectable dependencies for testing `piece setsrc --check`. */
 export interface CheckPieceSourceCommandDependencies {
   checkPiecePattern?: typeof checkPiecePattern;
+  /** `exitWithDataError`'s seam, so a test can observe the refusal. */
+  exit?: Parameters<typeof exitWithDataError>[1];
 }
 
 /**
@@ -1945,10 +1948,15 @@ export async function checkPieceSourceFromCommand(
     localPatternEntry(mainPath, options),
   );
   if (!report.compatible) {
-    throw new ValidationError(
-      `${mainPath} cannot replace the source for piece ${config.piece}:\n${report.message}`,
-      { exitCode: 1 },
-    );
+    // A refusal is a data condition — this source and this piece's stored
+    // state don't fit — not an arg-parse failure, so it reports like the
+    // `piece get` / `piece link` data errors above: plain stderr and exit 1,
+    // never a Cliffy ValidationError, which would dump the usage screen over
+    // the verdict.
+    exitWithDataError({
+      message:
+        `${mainPath} cannot replace the source for piece ${config.piece}:\n${report.message}`,
+    }, deps.exit);
   }
   return {
     config,
--- a/packages/cli/commands/piece.ts
+++ b/packages/cli/commands/piece.ts
@@ -3,6 +3,7 @@ import { Command, ValidationError } from "@cliffy/command";
 import { VerbInputValidationError } from "../lib/callable.ts";
 import {
   applyPieceInput,
+  checkPiecePattern,
   type EntryConfig,
   executePieceCallable,
   formatViewTree,
@@ -33,6 +34,7 @@ import {
   stepPiece,
 } from "../lib/piece.ts";
 import type { ExecutedPieceCallable } from "../lib/piece.ts";
+import type { PatternCompatibilityReport } from "@commonfabric/piece/ops";
 import type { InvocationOutcome, InvocationPhase } from "../lib/callable.ts";
 import { renderPiece } from "../lib/piece-render.ts";
 import { parseSqliteSource } from "../lib/sqlite-source.ts";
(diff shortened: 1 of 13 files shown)


bfollington (commented):
Scope: PR #5311, all 7 changed files deep-read; traced `compileProgram` through the durable cell cache and `setPattern` through Runner setup/CFC enforcement. Checks: PR piece preflight test (pass), CLI piece test (82 steps pass), plus a cold-argument regression test on base (pass) and PR head (fail).

## 🔴 Blocking

1. **Major — `packages/piece/src/ops/piece-controller.ts:3293`: the unconditional review changes what `setPattern` accepts; it is not message-only.** `pieceSourceCompatibilityReview()` materializes and validates the argument before entering Runner. When the whole argument document is cold, it validates `undefined` against the candidate and refuses. Runner deliberately treats that state as deferred because a nested piece's argument lives in its host document and the host may not have synced yet (CT-1917); `applySetupState` preserves the bytes and lets the swap proceed. I reproduced this with identical V1/V2 schemas containing a required `registry`: base `setPattern` applies V2, while this PR fails at line 3305 with `updated arguments do not match the candidate schema: value does not match type object`. This affects every non-override `setPattern` caller (CLI setsrc and FUSE source writes). It also adds an argument sync/materialization/validation and retained-link proof before `execute`, which validates them again on every successful swap. Please preserve Runner's cold-argument semantics and add this regression case. The lowest-cost shape is to keep enforcement on the existing execute path and derive the aggregate explanation only after an enforcement failure; if the review remains ahead of execute, it must implement the same cold-value deferral and avoid double work. **Verified by execution.**

2. **Major — `packages/piece/src/ops/piece-controller.ts:3250`: `compatible: true` does not mean the source can replace the piece because the real CFC envelope merge is omitted.** A stored argument/result envelope can reject at setup commit even when schema/argument/retained-link review passes (especially a partially migrated piece whose pattern pointer is ahead of its stored envelope—the Estuary class this feature is intended to catch). The PR acknowledges the gap, but the public contract and CLI say “can replace”, and success is intended to gate deploys. That makes a known false green more dangerous than no preflight. Please drive `mergeCfcSchemaEnvelopes` (and unreadable-envelope failure) from the shared review in this PR so `changeSource` benefits too, or narrow the API/CLI to explicitly report only the three checks and stop calling the result `compatible`. Given the current wording and purpose, I think the CFC check belongs in this PR. **Verified from the review's issue set and Runner's independent CFC commit enforcement; also explicitly acknowledged in the PR description.**

3. **Major coherence issue — `packages/piece/src/ops/piece-controller.ts:3227` / CLI help at `packages/cli/commands/piece.ts:1037`: `--check` does write durable data.** `compileProgram` calls `compileAndSavePattern`; its contract says it persists the content-addressed source and compiled documents in the target space, and the cell-cache write-back is awaited. The check does not update the piece, but “writes nothing” / “change nothing” is false. The tests only compare the piece's pattern ref/result, so they cannot prove the stronger statement. Please say that the piece is not updated but compilation may populate immutable source/compiled artifacts in the space, and rename the tests/comments accordingly. **Verified by tracing the compile path.**

## 🟡 Improvements

1. **`packages/piece/src/ops/piece-controller.ts:247`: the test proves shared rules under static state, not check/apply agreement.** `checkPattern` discards `argumentEvidence` and the expected source snapshot; the CLI then recompiles the filesystem on a later invocation. The candidate file/imports, current piece source, or retained argument can change in between. `setPattern` revalidates, so this is fail-safe, but a passing check does not guarantee a later apply or the same error text. Either document the report as a point-in-time advisory and weaken the “agreement” claims, or return a prepared value (candidate identity + expected source snapshot + argument evidence) that apply confirms, analogous to `PreparedPieceSourceChange`. If deploy-gate agreement is intended, I would choose the prepared confirmation.

2. **`packages/cli/commands/piece.ts:1947`: decide whether exit 1 is a stable scripting contract.** Today `ValidationError({ exitCode: 1 })` reaches the top-level exit code correctly, so it works as a boolean gate. It cannot distinguish incompatibility from an operational failure. The unit test pins the property to 1, while the integration test only checks “non-zero”, and no live CLI docs state the contract. Either promise/document/test exact 1 (including the real process), promise only non-zero and stop pinning 1, or use a dedicated incompatibility code so scripts can branch without parsing prose. The integration grep for `not backward compatible` also pins a low-level validator phrase likely to drift; keep the stable wrapper assertion and test typed issue classification closer to the piece layer.

Verdict: **request changes**. The feature direction and reuse of the shared review are good, but the current apply-path regression and known false-compatible CFC verdict make it unsafe to merge under the advertised contract.

— Codex (GPT-5), reviewing on Ben's behalf

bfollington:
All three blockers addressed in f5452cdf3. Thank you — the cold-argument one was a real regression against my own CT-1917 work, and nothing in my suite would have caught it.

**1. Enforcement is back where it was.** The pre-execute review is gone; `setPattern` is byte-for-byte main's behaviour again. `packages/piece/test/setsrc-cold-argument.test.ts` pins it: identical V1/V2 argument schemas with a required un-defaulted field, and the argument meta pointed at a never-written document so the read genuinely resolves to nothing (with a guard asserting exactly that, so the fixture cannot quietly go warm — merely-unsynced does not work, since the review awaits `argumentCell.sync()` itself). Reintroducing the review makes it fail with your exact message: `updated arguments do not match the candidate schema: value does not match type object`.

I also tried your lower-cost suggestion — derive the aggregate only after an enforcement failure — and dropped it. It relabelled a concurrency error (`pattern changed while the source update was compiling`) as a compatibility failure. A wrong diagnosis is worse than a terse one, so the enrichment is gone entirely and `--check` is the only place the full picture lives.

**2. The CFC envelope is now checked.** `pieceSourceCompatibilityReview` drives the real `mergeCfcSchemaEnvelopes`, with an unreadable stored envelope treated as a blocker rather than skipped, so `changeSource` benefits too. The fixture that demonstrates the closed hole was a false green before: `{"compatible":true,"issues":{}}` for a piece whose apply threw `confidentiality cannot be weakened at /seed`.

One deliberate limit, flagged rather than papered over: only the ARGUMENT role is checked. The result role's commit-time merge is exempted by `generatedOutputPaths` derived from the running module's actual output bindings, which a preflight cannot know without executing the module — checking it anyway would false-RED the legitimate case `state-continuity.test.ts` documents (a candidate adding a required, un-defaulted, generated result field). The old #5059 branch checked the result role without those options and I believe would regress that on today's main.

**3. The overclaims are corrected.** `checkPattern` does not write nothing — compiling the candidate persists content-addressed source and compiled docs; what it leaves untouched is the piece, which is now what the doc comment, the CLI help (`without updating the piece`) and the test names all say. The "changes nothing" test asserts the pattern pointer directly rather than inferring it from rendered output.

**On agreement (your 🟡 1):** you were right that it was point-in-time only, and removing the pre-execute review made the equality assertions false anyway. Both agreement tests now assert refuses-⇔-refuses plus the same underlying cause, and the report is documented as a point-in-time advisory that the apply path revalidates. I did not add a prepared-confirmation handshake; that is a real feature and I would rather it be its own PR than smuggled into this one — say the word if you disagree.

**On exit codes (your 🟡 2):** left as non-zero, with the unit test's exact-1 assertion kept and the CLI help now stating it exits non-zero. I have not promised a distinct incompatibility code; agreed a typed code would be better and it is a natural follow-up alongside `--json`. The integration grep on `not backward compatible` is unchanged for now — a fair point about drift, and the right fix is typed issue classification at the piece layer rather than a looser grep.

Verification: repo-wide `deno task check` exit 0, fmt and lint clean, `packages/piece` 34 passed, `packages/runner` full suite 1131 passed, `packages/cli` piece tests 82 steps.

PR #5312, opened 2026-08-04
test(patterns): adopt a true old-toolchain vintage — 2026-06-18 home.tsx (CT-1941)

bfollington:
## Why

Tier 2's capture compiles the old pattern with the **current** in-process toolchain, so every committed fixture is green on exactly the spaces that survive and silent on the ones that don't: a stored source today's toolchain no longer compiles — the class that has actually broken production (the 2026-07-29 vendor gate, the `cf-cell-context` retirement, `safeDateNow`) — was structurally uncapturable. CT-1941, with the demonstration recorded on the issue.

## What

**The fixture.** `system/home.tsx` as of `d5d115cda` (2026-06-18), captured out of process in a git worktree at that revision under its own toolchain (deno 2.8.1, its lockfile, its compiler/runner/memory stack), all four home children verified instantiated before snapshotting. 1.2 MB after the standard compaction — the smallest pinned fixture in the tree. Replaying it is the one fixture whose replay walks the by-identity load fallback into stored closures today's compiler cannot compile (the flags-on consequence path has its own assertion in `pattern-pointer-unloadable-swap.test.ts`).

**`tasks/vintage-adopt.ts`** — the current-tree half of the rig. Opens the snapshot (the memory migration chain runs, exactly as reopening a real old space does), derives the recorded root from its capture cause rather than trusting the old tree's id formatting — measured: the current tree re-derives the June-minted `of:fid1:brERQlTY…` byte-identically from the `"home-pattern"` cause — verifies the stored `patternIdentity`, writes the in-store manifest + restore control, and emits a normal pinned fixture. Refuses to overwrite.

**Two harness accommodations**, both documented in the spec:
- `vintageHoldsRoot` accepts `patternIdentity` as well as `patternSetupIdentity`. The setup marker (#4915, 2026-07) postdates every genuinely-old store, so the strict check refused exactly the fixtures this route exists to add. Production rolls markerless roots forward; for "was something really captured here" either stamp is evidence only a runner writes. Empty/truncated stores carry neither and are still refused.
- An adopted fixture has no test run to pin the vintage root, so the adopter stamps a marker doc at the same cause; it travels in the same file, so its presence after restore proves restoration the same way.

**Docs.** New spec section ("Adopting an externally captured fixture"); the "whether a fixture rots" limit is rewritten from *unresolved* to *measured* (see below). Full capture procedure + the revision-pinned capture script recorded in `docs/history/two-toolchain-vintage-rehearsal.md` — the script compiles only against the old revision's APIs, which is why it's a history record rather than committed source.

## What the replay measures (worth a reviewer's eyes)

During the fixture's replay, the runtime attempts to load the vintage's children by identity from their in-store June closures and **fails** — `pattern-load-error` × 2, `cf-cell-context` and `safeDateNow`, exactly the predicted drift — and the gate completes green because the updated root re-instantiates children from today's compiled program. That green is honest for what the gate asserts (state survival on recorded roots), but it means **by-identity load failures are invisible to the gate**. Whether they should be a graded finding is a design question for this machinery's owner rather than something this PR decides.

Known semantics: the fixture's identity predates today's source by construction, so it reports as "changed since capture" on every run, like a served route. Noted in the spec.

## CT-1939 relevance

This rig ran the closest approximation yet to the estuary blanking sequence — genuine June store, today's roll-forward, fresh `PieceManager.getDefaultPattern(true)` boot — and it **healed completely** despite the mid-update load failures. That eliminates "old store + roll-forward + fresh boot" as sufficient conditions and narrows the surviving candidates to CFC enforcement and the runtime-client worker boot route (recorded on CT-1939).

## Test plan

- `deno task pattern-vintage`: 6 vintages replayed, 91 instantiations, 79 upgrade targets, 16 updated cleanly, no state stranded — including the new fixture.
- `packages/piece` `state-continuity.test.ts`: 1 passed (6 steps).
- `tasks/pattern-vintage-run.test.ts` + `pattern-vintage-lib.test.ts`: 16 passed (153 steps) — the fixture-did-not-restore refusals still hold (empty stores carry neither marker).
- Committed fixture regenerated end-to-end with the committed `tasks/vintage-adopt.ts`.
- `deno fmt --check` / `deno lint` clean on changed files (the repo's ~23 pre-existing unformatted shell/ui files untouched); `deno task check-docs`: 523 blocks pass.

---
## Summary by cubic
Adds a true old‑toolchain vintage for `system/home.tsx` (2026‑06‑18) and extends the adopter to record and validate child roots so the gate checks children, not just the root (CT‑1941). Adds adopt‑time symbol checks and duplicate‑child protection to prevent fixtures the replay would refuse.

- **New Features**
  - `tasks/vintage-adopt.ts`: validates each recorded symbol resolves today (entry symbol or named artifact); rejects duplicate child cell ids and the entry root listed as a child; keeps existing checks (repo‑key mapping, upgrade target, compiles today); refuses served routes and markerless child ids; thin CLI with `--child=<cellId>=<main>`; never overwrites.

- **Refactors**
  - Presence accepts `patternIdentity` as well as `patternSetupIdentity`; diagnostics now say “neither a pattern identity nor a setup marker.” `staleTestKeys` documents adopted‑fixture “changed while identities differ”; spec adds “Adopting an externally captured fixture” and clarifies CFC remains at `enforce-explicit`; history record added.

<sup>Written for commit d3cfd4c182346d6b1cf7d859de83e61c8641dc1c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5312?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Coverage debt

The residual uncovered lines in `tasks/vintage-adopt.ts` are the `import.meta.main` CLI wrapper (arg parsing + default roots); every validation and store-writing path is exercised by `tasks/vintage-adopt.test.ts` (9 steps, including the end-to-end adopt→replay-migrates invariant).

ACCEPT_COVERAGE_DEBT: coverage-debt: tasks uncovered lines = 1375 lines

## After a strategic (Opus) review

A second review pass asked not "is it correct" but "does it hold the failure class or merely demonstrate it". Its central finding: a root-only manifest left the four children — the state whose blanking motivated CT-1939 — outside every gate control, and child blanking is structurally invisible through the root's value. Addressed: the fixture's manifest now records **five entries** (root + four children, each under its stored June identity), so the gate materializes and state-compares all five, including the two whose stored closures fail to compile. `adoptVintage` gained `children` (repeatable `--child=<cellId>=<main>`), with tests. The same pass drove the spec's flags-off/post-migration notes and the presence-marker wording fixes.

@seefeldb — one change here deserves your eyes specifically: `vintageHoldsRoot` now accepts `patternIdentity` as well as `patternSetupIdentity`. Your original comment argued the setup marker's observer-coupling; the widening trades that coupling for the ability to accept pre-#4915 stores at all (an adopted June fixture's roots carry only the older marker). The narrower alternative — stamping a synthetic setup marker at adopt time — felt more dishonest than widening the predicate, but it's your control.


bfollington (commented):

Scope: PR #5312 · detail review of the adopter, presence/restore controls, replay accounting, committed SQLite fixture, CI behavior, and live spec · checks: `deno task pattern-vintage`; focused vintage/state-continuity tests (17 tests / 159 steps); offline store inspection; changed-file fmt/lint/diff checks; actual GitHub Actions log.

## 🔴 Blocking

### [major] The adopter can publish a fixture that the replay refuses

`tasks/vintage-adopt.ts:87-124` constructs and snapshots the one-entry manifest without applying the capture-side addressability/target guards.

Verified: I ran the adopter against the committed snapshot with the correct stored identity but `main=not-a-repo-path`. It exited 0 and wrote a pinned fixture. `replayVintage` then returned `candidates=0`, `targets=0`, and `this fixture records 1 instantiation(s), but not one is something todays source can be applied to`.

That violates the existing invariant in `captureVintage`: a capture the gate accepts must be a replay it accepts. It also leaves the operator in the pinned-fixture dead end that the surrounding code is designed to avoid.

Please validate the adopted entry with the same shared predicates the replay/capture use (repo pattern mapping, non-test/keyless upgrade target, and ideally current source resolution/compilation) before writing anything. Factor the task into an importable function and add success/refusal tests. This is independently merge-blocking today: the Coverage Check fails because all 88 executable lines in `tasks/vintage-adopt.ts` are uncovered.

## 🟡 Improvements

### [minor] The normative changed-count claim conflates two different paths

`docs/specs/pattern-update-testing.md:338-341` says an adopted fixture reports changed on every run, like a served route.

The committed fixture currently does report `changed=1` and `updated=1` on every replay because todays locally compiled identity differs from its recorded identity. But this is conditional, not enforced by adoption: if `today === entry.identity`, the loop skips the migration and reports zero. Served routes are deliberately the opposite accounting case: they are materialized every run but excluded from both `changed` and `updated`.

Please state the actual rule: this fixture is identity-compared and counts as changed while the identities differ; served routes are unconditional replays counted separately.

### [minor] The presence-control explanation still claims setup-marker exclusivity

`packages/piece/test/state-continuity-harness.ts:1210-1231` still says the evidence is the setup marker and every manifest entry carries `patternSetupIdentity` by construction. The adopted fixture is a legitimate counterexample: its manifest is injected by the adopter and its root carries only `patternIdentity`. The missing-root diagnostic at `tasks/pattern-vintage-run.ts:473` also says only no pattern setup marker, although the predicate now checks both.

Please rewrite the native-capture paragraph to scope its observer/setup-marker correspondence to native captures, describe the adopted fallback separately, and report neither patternIdentity nor patternSetupIdentity when both are absent.

This widening itself remains fail-closed: the separate restore-control check runs first; the adopter refuses a wrong/empty root before stamping the synthetic control; and setup writes the old pointer/result projection in one transaction. A partially readable root with no stored schema also fails later rather than passing.

### [minor] Index the new historical record

`docs/history/two-toolchain-vintage-rehearsal.md` has the required metadata, but `docs/history/README.md` was not updated. The repository documentation lifecycle requires every new historical artifact to be added to that index.

## Verified requested semantics

- The stored DB contains five actual pieces, but its manifest contains exactly one entry for the Home root. Replay therefore credits only `system/home.tsx`: `candidates=1`, `targets=1`, `changed=1`, `updated=1`, no failures. The four children are incidental stored state, not overclaimed coverage.
- A missing manifest root is still diagnosed per entry and is not credited. Empty/truncated fixture refusals remain intact through the separate restore control and later schema/state checks.
- The actual GitHub Actions Pattern Update State Continuity job logged exactly two `pattern-load-error` entries and still passed. The workflow directly runs `deno task pattern-vintage`; logger severity/stderr is not interpreted as the process result. The overall PR is red for Coverage Check, not for those logger errors.

Verdict: request changes for adopter validation/tests; the committed fixture replay itself is sound and the logger errors are intentionally non-fatal.

bfollington:

Scope: PR #5312 at `df52782fc` · detail re-review of the adopter, replay predicates, complete new test, changed-count spec, presence diagnostics, and history index · checks: `tasks/vintage-adopt.test.ts` (8 steps); vintage runner/lib tests (16 tests / 153 steps); state-continuity test (6 steps); `deno task pattern-vintage`; changed-file fmt/lint/diff checks.

## 🔴 Blocking

None. The previous major finding is closed.

Verified: `adoptVintage` now rejects an unmappable or non-target entry before opening/writing, checks the stored root identity, and resolves plus compiles current source before writing the manifest, restore control, or output fixture. The original malformed-`main` shape is covered end to end and leaves `collectVintages` empty. The latest compile-refusal case also passes. The unmodified full gate replays 6 vintages / 91 instantiations / 79 targets / 16 clean updates with no stranded state.

## 🟡 Improvements

### [minor] The positive adopter test is an unchanged-identity no-op

`tasks/vintage-adopt.test.ts:26-27,104-115` says the durable cell gives the replay before/after state to protect, but capture, adoption, and replay all use the exact same `SUBJECT`. Therefore `today === entry.identity`, and `replayVintage` exits at `tasks/pattern-vintage-run.ts:644-645` before materialization or state comparison. The case does prove the structural invariant — manifest, restore control, presence, mapping, resolve, and compile — but it does not exercise the changed → materialize → compare path that an old-toolchain adoption exists for.

After snapshotting, make a compatible source change before `adopt()`, then assert `changed === 1`, `updated === 1`, `stranded === 0`, and no failures. That would make the durable-state comment and the real adoption path true in the test.

### [minor] The revised identity-accounting statement is not enforced for adopter input

`tasks/vintage-adopt.ts:100` delegates to `patternKeyFromMain`, which deliberately accepts both repo paths and `/api/patterns/...` served routes (`tasks/pattern-vintage-lib.ts:392-405`). `isUpgradeTarget` also accepts that route, so the adopter can publish one; replay then treats it as `servedRoute`, contrary to the general statement at `docs/specs/pattern-update-testing.md:338-343` that an adopted fixture is identity-compared and is not the served-route case.

Either reject anything that does not start with the repo patterns prefix in the adopter and add a refusal test, or make the spec conditional on the supplied `main` being a repo path. This does not reopen the blocker because a served-route fixture is replayable; it is a remaining code/spec input-domain mismatch.

Verdict: the blocker is closed and the PR is mergeable; the two items above are non-blocking test/spec hardening.

bfollington:

Scope: final delta review of `f386996bf..8838b9c0f` · deep-read of child adoption/replay semantics, direct inspection of the committed SQLite store, comments/spec, and compile-cache persistence · checks: focused adopter/replay tests (70 steps); single-fixture and full vintage replay; `cf inspect` plus manifest/root reads; changed-file fmt/lint; `check-docs` (523 blocks).

## 🔴 Blocking

### [major] Child adoption still admits entries the replay rejects or misaccounts

`tasks/vintage-adopt.ts:188-244,253` validates that every child cell carries some `patternIdentity`, validates its `main` with the shared mapping/target predicates, then compiles each unique key. Two per-entry invariants are still missing:

1. There is no uniqueness check on `(space, cellId)`, including against the entry root. Repeating one child inflates `targets` / `changed` / `updated`; pairing the same real cell with two compilable `main`s can credit two pattern keys after applying both programs to one root. The before-state map deduplicates by cell, but the materialize/accounting loop deliberately iterates every manifest entry, so the manifest duplication is not neutralized downstream.
2. Compiling a unique key validates only its entry pattern. The stored child `symbol` is never selected. A legitimate old child can name a non-default transformer hoist such as `__cfPattern_N`; if that symbol disappeared today, `compilePattern` succeeds and the adopter writes the fixture, then `materializeOnCell` refuses it at `state-continuity-harness.ts:1443-1468`. The existing replay test “FAILS CLOSED when a stored root names an artifact this version dropped” verifies that exact downstream outcome.

That breaks the stated invariant at the top of the adopter: an accepted adopt must be a replay the gate accepts, and it also leaves a false-coverage path through duplicate/mispaired cells.

Please reject duplicate cell keys (root included), retain the compiled entry ref per key and validate every recorded symbol with the same selection rule the materializer uses, and add refusal tests proving no fixture is written for both cases. A dry-run replay before finalizing the output would be the stronger single control if the tool wants to hold the full invariant rather than reproduce more replay checks.

## 🟡 Improvements

### [minor] The adopted-fixture staleness comment is categorical beyond the code

`tasks/pattern-vintage-run.ts:190-196` says an adopted identity “can never equal today’s” and therefore reports changed forever. `adoptVintage` does not enforce identity drift and accepts an unchanged source; the existing “never overwrites” test constructs exactly such an adoption. Replay then takes the ordinary `today === entry.identity` no-op. The spec already has the accurate conditional wording: it remains changed only for as long as the identities differ, which is every run for this committed old-toolchain fixture in practice. Please use that wording in the comment too.

### [minor] The flags-off note misstates the effective CFC mode

`docs/specs/pattern-update-testing.md:378-381` says the replay sets no CFC enforcement mode. `openFileBackedRuntime` does omit the option, but `Runtime` defaults it to `"enforce-explicit"` (`packages/runner/src/runtime.ts:1077-1078`). `systemPatternAutoUpdate` is genuinely off, and that alone explains why the pointer roll-forward repair is absent. Please say that auto-update is off while CFC remains at its default enforcing mode; “neither” reads as though both effective flags are off.

## Verified requested semantics

- The committed database contains exactly five distinct piece roots. Each manifest entry has a repo-path `main`, `symbol=default`, a stored identity matching the marker on its own cell, and readable pre-update state.
- Its isolated replay reports `5 candidates / 5 targets / 5 changed / 5 updated / 0 served-route / 0 stranded`, crediting home, favorites-manager, self, profile-picker, and profile-create. Favorites-manager and self first produce the expected old-identity `cf-cell-context` / `safeDateNow` load failures, then both continue through current-source materialization and state comparison; neither is skipped.
- The revised presence-marker diagnostic/comment matches the predicate. I also checked the June runner: it writes schema and identity through the same setup transaction, so the defensive “older identity marker but no schema” discussion is accurate as written.
- The post-migration claim is accurate: adoption opens and migrates the temporary restored copy, then snapshots that current-format store; CI does not repeat the old-to-current migration.
- The 2.0 MB compile-cache growth does not mask the June closure. Old-root loading is keyed by the recorded June identity, while adoption persisted different current identities; the two expected source-recompile failures prove those current docs did not satisfy the old lookups. Candidate compilation may warm-hit the adopted current closure, but it first derives content identities from the repo-resolved program, so a future source change misses those entries and compiles the new source.

Verdict: the five-entry fixture and replay assertions are sound, but request changes on the generic child-adoption validation before human review.

PR #5314, opened 2026-08-04
fix(cli): narrow initial piece get transform pull

bfollington:
## Summary

- derive an unambiguous source root shape from its declared schema before running piece get transforms
- push the union of filter-observed and projected paths into the first source selector
- resolve local $ref-backed item schemas for selection while preserving source IFC, scope, defaults, and other Fabric metadata
- document the conservative fallback for schema-less, root-union, and ambiguous composed schemas

The selector regression uses a fresh $ref-backed array view. It verifies that predicate-only status is available for filtering but absent from the id/title result, and that an omitted unavailable linked details document is never synchronized.

Linear: https://linear.app/common-tools/issue/CT-1931/cf-piece-get-push-filterschema-selection-into-the-initial-pull

## Validation

- mise exec -- deno test -A packages/cli/test/piece-get-transform.test.ts (45 steps passed)
- mise exec -- deno task check-skill-facts
- mise exec -- deno fmt --check
- mise exec -- deno lint
- full CLI suite: 1668 passed; two unrelated color-mode assertions failed and reproduce when color-mode.test.ts is run alone

---
## Summary by cubic
Narrows the initial selector for `piece get` to the union of filter and projection fields when the source schema fixes the root shape, preventing unnecessary linked hydration and improving sync correctness. Implements Linear CT-1931 with conservative schema-root detection and clearer CLI errors.

- **Bug Fixes**
  - Initial pull narrowing: build the first read from predicate+projection paths; predicate-only fields don’t appear; linked docs outside the union aren’t synced; resolve local `$ref` item schemas for selection while preserving `ifc`, `scope`, `default`, and `asCell`.
  - Root shape classification: new `schemaRootKind` walk where explicit `type` wins; only unambiguous roots are classified; ambiguous `anyOf`/`oneOf`/`allOf` keep a wider selector; unresolved/external `$ref` leave the selector intact for projection.
  - Array handling and errors: detect arrays from the declared schema when clear, otherwise inspect the value; treat an unset declared array as empty; translate runner errors to CLI messages (“--filter can only be applied to an array”, “--schema can only project array items from an array value”); emit “Could not resolve source schema reference for --schema” when a referenced item schema can’t resolve.

<sup>Written for commit cf3fd9646f3096a2587aac5d06a4f013ac6d8d26. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5314?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -163,11 +163,19 @@ remains the ordinary successful `null` CLI response, as does a valid projected
 null.
 
 Both transforms run as a short-lived computed pattern in the caller's session.
-The runtime's list filter/map builtins therefore handle CFC exactly as authored
-pattern expressions do: predicate observations label array membership,
-projection reads propagate labels, and filtered elements retain their source
-links. Projection map/lift nodes construct the requested shape from a
-source-schema-selected read rather than returning a widening identity alias.
+When the declared source schema fixes the root container shape, the pattern
+constructs its first storage read from the union of predicate-observed and
+projected paths. Structurally declared properties, including local `$ref` item
+schemas, are pruned to that union: predicate-only fields can decide membership
+without appearing in the result, and omitted linked subgraphs are not hydrated.
+Schema-less or root-union sources retain a value-shape read before the transform
+because their array/object projection semantics cannot be established from the
+declaration. Ambiguous source-schema compositions remain intact and can retain a
+wider selector. The runtime's list filter/map builtins therefore handle CFC
+exactly as authored pattern expressions do: predicate observations label array
+membership, projection reads propagate labels, and filtered elements retain
+their source links. Projection map/lift nodes construct the requested shape from
+a source-schema-selected read rather than returning a widening identity alias.
 Nested non-stream Cell handles are materialized before the predicate/projection
 JavaScript runs; stream handles remain capabilities. The source cell's schema
 remains authoritative for Common Fabric metadata. A caller cannot introduce or
--- a/packages/cli/lib/piece-get-transform.ts
+++ b/packages/cli/lib/piece-get-transform.ts
@@ -704,6 +704,9 @@ function walkSchemaRoot<T>(
   ancestors.add(schema);
   const documentRoot = schema.$defs !== undefined ? schema : root;
   try {
+    // A reference is the source shape for this heuristic. Resolve it before
+    // inspecting sibling keywords so a broken reference cannot lend false
+    // shape authority through, for example, a sibling `type`.
     if (schema.$ref !== undefined) {
       try {
         return walkSchemaRoot(
@@ -721,6 +724,9 @@ function walkSchemaRoot<T>(
     }
 
     const types = schemaTypes(schema);
+    // An explicit `type` is the authoritative root-shape signal here.
+    // Combinators are conjunctive with it, so contradictory branch types
+    // cannot expand the root shapes it declares.
     if (types.length > 0) return behavior.fromTypes(types);
 
     const alternatives = [...schema.anyOf ?? [], ...schema.oneOf ?? []];
@@ -1512,6 +1518,11 @@ export async function derivePieceGetValue(
     const outputValue = outputCell.get();
     const recorded = errors.slice(errorCountBefore);
     if (recorded.length > 0) {
+      // Translate the array-shape errors emitted by the runner filter/map
+      // builtins into CLI-level messages. Keep these matches in sync with
+      // packages/runner/src/builtins/{filter,map}.ts; the mismatch regression
+      // test in piece-get-transform.test.ts guards this package-boundary
+      // coupling.
       if (
         transform.filter !== undefined &&
         recorded.some((error) =>
--- a/packages/cli/lib/piece-get-transform.ts
+++ b/packages/cli/lib/piece-get-transform.ts
@@ -686,67 +686,107 @@ function schemaIsArray(schema: JSONSchema | undefined): boolean {
 
 type SchemaRootKind = "array" | "non-array" | "unknown";
 
+interface SchemaRootWalkBehavior<T> {
+  unknown: T;
+  fromTypes: (types: string[]) => T;
+  fromAlternatives: (values: T[]) => T;
+  fromAllOf: (values: T[]) => T;
+  unresolvedRef: (error: unknown) => T;
+}
+
+function walkSchemaRoot<T>(
+  schema: JSONSchema | undefined,
+  behavior: SchemaRootWalkBehavior<T>,
+  root: JSONSchema = schema ?? true,
+  ancestors = new Set<object>(),
+): T {
+  if (!isRecord(schema) || ancestors.has(schema)) return behavior.unknown;
+  ancestors.add(schema);
+  const documentRoot = schema.$defs !== undefined ? schema : root;
+  try {
+    if (schema.$ref !== undefined) {
+      try {
+        return walkSchemaRoot(
+          ContextualFlowControl.resolveSchemaRefsOrThrow(
+            schema,
+            documentRoot,
+          ),
+          behavior,
+          documentRoot,
+          ancestors,
+        );
+      } catch (error) {
+        return behavior.unresolvedRef(error);
+      }
+    }
+
+    const types = schemaTypes(schema);
+    if (types.length > 0) return behavior.fromTypes(types);
+
+    const alternatives = [...schema.anyOf ?? [], ...schema.oneOf ?? []];
+    if (alternatives.length > 0) {
+      return behavior.fromAlternatives(
+        alternatives.map((option) =>
+          walkSchemaRoot(option, behavior, documentRoot, ancestors)
+        ),
+      );
+    }
+
+    if (schema.allOf !== undefined) {
+      return behavior.fromAllOf(
+        schema.allOf.map((option) =>
+          walkSchemaRoot(option, behavior, documentRoot, ancestors)
+        ),
+      );
+    }
+    return behavior.unknown;
+  } finally {
+    ancestors.delete(schema);
+  }
+}
+
 /**
  * Classify a root container only when the declared schema makes its shape
  * unambiguous. This is intentionally conservative: an unknown shape falls
  * back to inspecting the value so schema-less and union-shaped cells retain
  * their existing projection semantics.
+ *
+ * @internal Exported for focused root-shape tests.
  */
-/** @internal Exported for focused root-shape tests. */
 export function schemaRootKind(
   schema: JSONSchema | undefined,
   root: JSONSchema = schema ?? true,
   ancestors = new Set<object>(),
 ): SchemaRootKind {
-  if (!isRecord(schema) || ancestors.has(schema)) return "unknown";
-  ancestors.add(schema);
-  const documentRoot = schema.$defs !== undefined ? schema : root;
-  try {
-    const types = schemaTypes(schema);
-    if (types.length > 0) {
-      const hasArray = types.includes("array");
-      if (!hasArray) return "non-array";
-      return types.every((type) => type === "array") ? "array" : "unknown";
-    }
-
-    if (schema.$ref !== undefined) {
-      try {
-        return schemaRootKind(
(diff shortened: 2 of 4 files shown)


bfollington (commented):
_cf-review via Claude `claude-opus-5`, on behalf of @bfollington._

**Scope** · 4 files, +273/−13, one package (`packages/cli`) + 2 doc surfaces · **Coverage**: deep-read (all changed code plus the `ContextualFlowControl` ref / `schemaAtPath` machinery it now leans on) · **Checks run**: `deno test packages/cli/test/piece-get-transform.test.ts` (45 steps pass), `deno fmt --check` + `deno lint` on touched files (clean), `deno task check-skill-facts` (clean), a local coverage run, and four differential probes against `origin/main`.

The change does two things: `selectSourceSchema` now resolves local `$ref`s so the narrowed selector isn't a no-op for `$ref`-backed item schemas, and `derivePieceGetValue` classifies the root container from the declared schema (`schemaRootKind`) instead of pulling the whole value to call `Array.isArray`. Together they drop the eager wide pull. Scope is coherent and the docs are updated in the same change.

The core mechanism is right, and the new selector test is a good one — asserting on `syncedUris` (that the omitted linked `details` doc is never synced) is the real guard, not just a shape check.

## 🔴 Blocking

**1. Coverage Check fails, and the uncovered lines are exactly this PR's two safety promises.** `packages/cli` 3625 → 3629 (+4). *Verified*: I ran `deno coverage` against the changed file locally; the four lines are `packages/cli/lib/piece-get-transform.ts:722-724` (`schemaRootKind`'s unresolvable-`$ref` → `"unknown"`) and `:1055-1060` (`selectSourceSchema`'s unresolvable-`$ref` → return `source`).

Those are precisely the conservative fallbacks that `packages/cli/README.md` and `skills/cf/SKILL.md` now promise, so this wants real tests rather than an `ACCEPT_COVERAGE_DEBT` override. Verified fix — I ran this; it passes and covers both branches:

```ts
const broken = { $ref: "#/$defs/Missing", $defs: {} } as const;
expect(schemaRootKind(broken)).toBe("unknown");
expect(selectSourceSchema(broken, {
  type: "object", properties: { id: true }, additionalProperties: false,
})).toBe(broken);
```

(It emits two `[WARN][cfc] Unresolved $ref` lines — fine on the CLI lane, unlike pattern-unit.)

## 🟡 Improvements

**2. Declared-array + non-array value now bypasses the friendly `--filter` guard and leaks a raw builtin error.** Because `sourceIsArray` comes from the schema, the `--filter can only be applied to an array` check no longer sees the value. *Verified* by running identical probes against `origin/main` and this branch:

| source | main | this PR |
| --- | --- | --- |
| declared array, value never set, `--filter` | `--filter can only be applied to an array` | `[]` ✅ improvement |
| declared array, value `null`, `--filter` | `--filter can only be applied to an array` | `Could not apply piece get transform: filter currently only supports arrays` |
| declared array, value is an object, concise projection | `undefined` | `Could not apply piece get transform: map currently only supports arrays` |
| declared object, value is an array, `--filter` | clean error | clean error (unchanged — the schema-shaped read already normalizes) |

The unset-array case genuinely gets better. The other two swap a deterministic CLI error for an internal builtin message. Nothing in the PR tests this delta, and the existing `rejects --filter for non-array sources` test doesn't reach it (that's the declared-object row, which is unchanged). Either pin the new behavior in a test, or map the builtin message back to the CLI-level one.

**3. `schemaRootKind` is a near-clone of `schemaMayBeArray`, with the opposite `$ref`-failure policy.** Same file, same `(schema, root = schema ?? true, ancestors = new Set())` signature, same `documentRoot` derivation, same `try/finally` ancestor pop, same anyOf/oneOf/allOf walk (`:696` vs `:859`). But on an unresolvable `$ref`, `schemaMayBeArray` throws `PieceGetTransformError("Could not resolve source schema reference for --schema")` while `schemaRootKind` silently returns `"unknown"` — and both run against the same source schema inside the same `derivePieceGetValue` call. Today the net user-visible behavior is still the clean throw (the classifier runs first and falls back, then `alignConciseProjectionMask` throws), but that is incidental ordering. Share one walk, or at minimum say in a comment why the two policies differ.

**4. Stacked JSDoc on `schemaRootKind`** (`:689-695`) — two consecutive `/** */` blocks, so only the trailing `@internal Exported for focused root-shape tests.` attaches to the symbol and the substantive explanation is orphaned from tooling. Merge them, or move the rationale into the body the way `selectSourceSchema` does.

## Questions

- `{type: ["array", "null"]}` classifies as `"unknown"` and falls back to the wide value pull. Correct and conservative — but if nullable arrays are common in real piece result schemas, the optimization won't apply where it would matter most. Deliberate, or worth extending to "array plus null"?
- Still Draft — intended to land as-is once coverage is green?

## Notes for the record (no action)

I checked whether the inlined `$ref` bloats the wire selector. It does at the `provider.sync` boundary — the full `Item` definition, including the pruned-away `details` subtree, appears in `$defs` at both the array root and inside `items` — but `normalizeSyncSelector` in `packages/runner/src/storage/v2-watch.ts:25` already runs `pruneCfcSchemaDefinitions` before the wire, so the dead defs never leave. Not a finding; recording it so nobody re-derives it.

**Verdict:** sound, well-scoped change with a good primary test. One blocker (the coverage gate), and its fix is the same two tests that close the untested-fallback gap. Land after that, plus a decision on finding 2.

bfollington:
Addressed in e2acd953b.

- Added regression coverage for both unresolved-ref fallbacks: schemaRootKind returns unknown and selectSourceSchema preserves the source schema.
- Replaced the duplicated schema-root recursion with one shared walker. The two unresolved-ref policies remain intentionally explicit: root classification is only an optimization and falls back conservatively, while concise schema path alignment throws because it cannot safely infer traversal.
- Mapped declared-array/runtime-null-or-object builtin failures back to stable CLI-level errors, with tests for filter-only, filter plus projection, and concise projection.
- Merged the stacked schemaRootKind JSDoc.

I kept array-plus-null roots classified as unknown deliberately. The current value may validly be null, so treating the declaration as definitively array-shaped would skip the value check and send null into array-only filter/map builtins. The conservative pull preserves the command semantics for that union; optimizing nullable arrays safely would need a shape probe or conditional graph path rather than declaring the root to be an array.

Validation:
- focused transform suite: 47/47 passed under Deno coverage
- no uncovered executable lines in the changed implementation ranges in the generated LCOV report
- repo-wide deno fmt --check: clean
- repo-wide deno lint: clean

The PR remains draft while the refreshed CI run completes.

bfollington (commented):
_cf-review (re-review) via Claude `claude-opus-5`, on behalf of @bfollington._

**Scope** · re-review of `e2acd953b` "address piece get pull review" (+157/−59 over `c3b7198ad`, `packages/cli` only) · **Coverage**: deep-read of the whole delta · **Checks run**: `deno test packages/cli/test/piece-get-transform.test.ts` (47 steps pass), full `packages/cli` suite via `deno task test` (123 passed / 0 failed — the two color-mode assertions noted in the PR body did not reproduce), `deno fmt --check` + `deno lint` on touched files, a local coverage run, and differential probes across `main` / `c3b7198ad` / `e2acd953b`.

All four findings are addressed, and the fix for #3 is a better answer than what I suggested.

## Resolved

1. **Coverage blocker — fixed.** `falls back conservatively when a source reference cannot resolve` covers both `catch` arms. *Verified*: `piece-get-transform.ts` now reports **100.0% line, 100.0% function, 98.2% branch** from its own test file alone, and CI is **48 SUCCESS / 4 skipped** — Coverage Check green.
2. **Error surfacing — fixed.** Both builtin messages are re-mapped to CLI-level errors and pinned by `reports schema/value root mismatches with CLI-level errors`, including the filter+projection combination. I checked the wording: concise projections do arrive via `--schema` (`packages/cli/README.md:137`), so `--schema can only project array items from an array value` is accurate.
3. **Duplication — fixed, and better than proposed.** `walkSchemaRoot` + `SchemaRootWalkBehavior<T>` backs both classifiers, and the two divergent `unresolvedRef` policies now sit side by side with comments explaining why one falls back and the other throws. That reads much better than the "add a comment" I suggested.
4. **JSDoc — fixed.**

## 🟡 New, introduced by the refactor

**`schemaMayBeArray` lost its combinator fall-through.** *Verified* by running the same inputs against all three revisions:

| input | `main` | `c3b7198ad` | `e2acd953b` |
| --- | --- | --- | --- |
| `{type:"object", anyOf:[{type:"array"}]}` | `true` | `true` | **`false`** |
| `{type:"object", allOf:[{type:"array"}]}` | `true` | `true` | **`false`** |

The old body did `if (schemaTypes(schema).includes("array")) return true;` and then *fell through* to `[...anyOf, ...oneOf, ...allOf].some(...)`. The unified walk `return`s at `behavior.fromTypes(types)` whenever `types.length > 0`, so the combinators are never consulted once any `type` is present.

`schemaMayBeArray` gates whether a concise projection traverses array items, so a flip changes the projected shape. Reachability is low — it needs a schema declaring a scalar `type` *and* a combinator branch of the opposite shape, which is self-contradictory — and the new answer is arguably the more correct one under conjunctive JSON Schema semantics. So I'm not asking for a revert, just for it to be a decision rather than a side effect: this is a behavior change to pre-existing `main` code inside a commit framed as a duplication refactor, and neither answer is tested. A one-line comment on `fromTypes` saying "a declared type is conjunctive with any combinator, so it decides on its own" would settle it.

## ⚪ Nits

- **`$ref` is now evaluated before `type` in `schemaRootKind`** — the unified walk adopted `schemaMayBeArray`'s ordering. *Verified* consequence: `{$ref:"#/$defs/Missing", type:"array", $defs:{}}` classifies `"array"` → `"unknown"`. Strictly more conservative (falls back to the value pull), so safe; noting only that the reorder is real and unremarked.
- **The builtin message strings are matched by exact equality across a package boundary** — they are owned by `packages/runner/src/builtins/filter.ts:348` and `map.ts:338`. The new test does fail loudly if either is renamed, so the coupling is guarded, but the failure would surface in a CLI test with no visible link to the rename. A comment naming those two lines would save that trip.
- **The improvement case is still unpinned.** *Verified* on this head: an unset (never-written) declared-array cell with `--filter` returns `[]`, where `main` threw. Only the `null` case is asserted; one more line in the new test would lock the better behavior in.

**Verdict:** blocker cleared, all four findings addressed cleanly, CI fully green. Only open item is confirming the `schemaMayBeArray` semantic drift is intentional. Ready to land once that's answered and it leaves Draft.

bfollington:
Addressed in ebc43fd8a.

The schemaMayBeArray change is intentional: this root-shape heuristic treats an explicit type as authoritative. Combinators are conjunctive with that declaration, so contradictory branch types do not widen the declared root possibilities. I added a named regression test covering both anyOf and allOf cases.

I also actioned the three nits:
- documented and tested that a ref is resolved before sibling type, so a broken ref cannot acquire false shape authority from that sibling;
- documented the exact runner filter/map error-string coupling at the CLI translation site, including the owning builtin paths;
- added a dedicated regression test showing that filtering an unset declared array returns an empty array.

Validation:
- focused transform suite: 49/49 steps passed
- repo-wide deno fmt --check: clean
- repo-wide deno lint: clean

bfollington:
Follow-up on CI: the first refreshed Coverage Check compared this branch against a much newer main baseline (packages/cli uncovered lines had fallen from 3625 to 3497 while this branch was 38 commits behind). I rebased the three PR commits onto current main and force-pushed with an exact lease. No conflicts occurred.

Post-rebase validation remains clean:
- focused transform suite: 49/49 steps passed
- repo-wide deno fmt --check: clean
- repo-wide deno lint: clean
- branch is now 3 commits ahead / 0 behind origin/main

PR #5316, opened 2026-08-04
docs(plan): split verb results into a shared read model and its parts

mpsalisbury:
## Why

A verb's result and a direct read were being designed as two contracts. They are
one operation on different cells: a receipt is an ordinary cell, and nothing in
the read path checks piece-ness — `PieceManager.getResult` is the identity
function, and `cf piece get` reads a receipt address today without modification.

Designing them apart was expensive in ways that only show up as wasted work: a
`--depth` axis specified to bound a traversal that selects by path and therefore
never counts hops; four questions drafted for the `traverse.ts` owner that stop
applying once that's understood; and a result format specified for verbs that
the shared read path should own — the incident motivating all of it (a board read
past 300k tokens) was a `get`, not a call.

The restructure lets the read layer be settled against `cf piece get`, where the
load already is, with calls inheriting it. It also takes the command surface off
the critical path, since sharing a read step between commands requires no
renaming.

## What Changed

Docs only. Four files, ~1,300 lines.

| File | |
| --- | --- |
| `fabric-read-model.md` (147) | **Read this one cold.** The model, the vocabulary, and why three concerns split into two documents. |
| `shaped-reads-and-verb-results.md` (455) | The shared read layer and what a call adds. Carries the measured facts and the two remaining open questions. |
| `cli-surface-shape.md` (231) | The command surface: seven purposes, where it accreted, a target shape, an additive path that renames nothing before step 6. |
| `verb-result-selection.md` (484) | Reduced from a plan to working notes — call-specific detail the others don't carry. |

## What review settled

- **Schemas are queries** (Berni). The docs had framed a source schema and a
  reader's as opposites, which contradicts schema-on-read. Rewritten across all
  three; the naming problem relocated to one flag carrying two syntaxes, so
  `--select` takes the concise form and `--schema` keeps full schemas.
- **Address marking** (Ben). A `$link` marker beside `properties`, not a
  borrowed `asCell` — which carries a handle contract that can't cross a
  serialized channel.
- **Receipts carry a descriptive schema** (Ben), distinct from the declared
  result schema deferred to Fabric-types; the compatibility gate never reaches a
  receipt.
- **Invocation ids scope to a session, not a principal** (Berni). A DID
  separates nothing when agents work under their human's key, and a minted
  session is also unguessable, which a DID is not.

Two questions remain open, both in the read-layer document: what a receipt's
schema should record, and the mechanism for session scoping.

## Test plan

`deno task check-docs`, `check-conflict-markers`, `check-skill-facts` — all pass.
No code changes; `docs/` is excluded from `deno fmt`.

Runtime claims were measured against a local server rather than inferred, and
the evidence is in the read-layer document's "Measured facts". One claim was
retired when #5410 landed mid-review: the unshaped-read handle leak is fixed
(17,024 bytes → 976 on the same fixture), so that section now records what
remains true — an unshaped read emits the runtime's internal link encoding,
which is faithful but is not a contract.

### DIFF
--- a/docs/plans/cli-surface-shape.md
+++ b/docs/plans/cli-surface-shape.md
@@ -1,0 +1,179 @@
+# CLI surface shape
+
+The **command surface**, one of the three concerns in
+[Reading Fabric data](fabric-read-model.md), which defines the shared model. The
+`cf` surface grew one command at a time and now expresses a model the code no
+longer has. That costs more than tidiness: the names imply distinctions the runtime
+does not make, which misdirects design work before it starts.
+
+This document is about the command surface, not the machinery underneath.
+Nothing here blocks [Shaped reads and verb results](shaped-reads-and-verb-results.md).
+
+## What the surface is for
+
+Every command serves exactly one of seven purposes. Sorting them this way is
+what makes the trouble visible.
+
+| Purpose | Commands |
+| --- | --- |
+| **Authoring** — work on source files, never touching a live space | `check`, `test`, `view` (pager), `init`, `deps update` |
+| **Identity and access** — who you are, and who may do what | `id` (new/did/derive/from-mnemonic), `acl` (ls/set/remove) |
+| **Live data** — reading and writing running state | `piece get`/`set`/`call`/`apply`/`link`/`step`/`verbs`/`inspect`, `wish` |
+| **Piece lifecycle** — deploying and managing running programs | `piece new`/`setsrc`/`getsrc`/`rm`/`ls`/`search`/`map`/`set-slug`/`recreate-root`/`set-home` |
+| **Rendering** — turning things into something to look at | `piece view` (terminal), `piece render` (HTML), `view` (source pager) |
+| **Storage forensics** — reading the database directly, offline | `inspect` (18 subcommands), `space` (clone/verify/reset/fingerprint) |
+| **Filesystem projection** — exposing cells as files | `fuse` (mount/unmount/status), `exec` |
+
+Four of these are coherent and want nothing: authoring, identity and access,
+storage forensics, and filesystem projection. The trouble is concentrated in the
+middle three.
+
+## Where it has accreted
+
+**`piece` carries four unrelated jobs.** Deploying a program, listing what is
+deployed, reading live data, and rendering a UI all hang off one word. Someone
+reading a value and someone deploying a pattern share a command prefix and
+nothing else.
+
+**Names imply distinctions the code does not make.**
+
+| Surface | What the name says | What it actually is |
+| --- | --- | --- |
+| `--piece` | a piece | any cell address — the function that fetches a piece's result returns the piece unchanged, and the read path checks nothing piece-specific |
+| `piece get` | reading a piece | reading a cell at a path |
+| `--input` | a mode you switch on | an address — it follows a link stored in the document to reach the arguments cell |
+| `--schema` | a type contract you must satisfy | a selection over the value, which also accepts a bare list of paths |
+
+The `--piece` case is not cosmetic. Believing the target had to be a piece is
+what made a verb's receipt look like it needed purpose-built read machinery,
+when it is an ordinary cell that any read already handles.
+
+**One word, two jobs — twice.** `piece inspect` examines a running piece, while
+`cf inspect` reads the database offline and has its own `piece` subcommand.
+Separately, `piece view` renders a piece's UI in the terminal while `view` is a
+pager for source files. `piece map` and `cf inspect graph` both draw the graph
+of entities and their connections — one live, one offline.
+
+**One capability, two commands.** `piece apply` writes a whole set of inputs;
+`piece set` writes at one path. Both target the same arguments cell.
+
+### The input spec is not a schema
+
+`--schema` is the most misleading name on the surface, and the one worth fixing
+first, because it shapes how people reason about the whole read model.
+
+A schema describes what data **is**. What this flag takes describes what the
+caller **wants back**. Same syntax, opposite direction — and four things follow
+from the mismatch:
+
+- **It accepts input that is not a schema at all.** `--schema
+  'title,createdBy.name'` is a list of paths.
+- **It is a strict subset, stripped of what gives a Fabric schema meaning.**
+  `asCell`, `default`, `scope`, and `ifc` are forbidden to callers; `$ref`,
+  `$defs`, and the combinators are unsupported. What remains is structure with
+  the semantics removed.
+- **It is about to carry keywords that are not JSON Schema.** Marking a position
+  as an address needs a projection-only keyword, which no schema dialect
+  defines.
+- **It misleads readers in practice.** "How is `topic.title` a schema? It's a
+  path" is the reasonable question of someone reading the flag's name and
+  finding it does not hold.
+
+The confusion is not that schemas are being used as queries. In this system they
+legitimately are — a subscription is a graph query whose selector is a schema,
+and that is why the *syntax* should stay schema-shaped, so a caller can lift a
+source schema and prune it into a request. The problem is narrower: this
+particular schema is output-directed and deliberately incomplete, and the name
+claims more than it delivers.
+
+**Rename the flag; keep the syntax.** `--shape` says what it is, covers both the
+concise and nested forms, and promises no validation. Doing it now is much
+cheaper than later: `piece call` does not have this flag yet, so adding it there
+as `--shape` and aliasing on `get` costs one deprecation on a flag that is still
+young. Adding it as `--schema` doubles the footprint of the wrong name.
+
+## What it should look like
+
+Reading is one operation reached from several starting points, so the surface
+wants one read command and distinct commands for the distinct ways of arriving:
+
+```
+# <read opts> = [--shape S] [--filter P] — the shared tail, identical everywhere
+
+cf get   <addr> [path]           <read opts>
+cf call  <addr> <verb> <payload> <read opts>
+cf wish  <query>                 <read opts>   # a query, not an address
+cf exec  <mountedFile> [args]    <read opts>   # reached through a filesystem mount
+
+cf set   <addr> [path]                         # writes; nothing to shape
+
+cf piece new|setsrc|getsrc|rm|ls|search|verbs|slug   # deploying and listing only
+cf space … | cf id … | cf acl … | cf fuse …
+cf inspect …                                          # offline forensics
+cf check | test | view | init | deps                  # working on source
+```
+
+Three properties earn their place.
+
+**One way to write an address.** An entity id (`of:fid1:…`), a slug, or a URL,
+with a suffix for navigating within it — `<addr>#argument` in place of the
+`--input` flag. An address printed by one command is accepted by the next
+without reshaping, which is not true today.
+
+**Arrivals stay separate; the tail is shared.** `get`, `call`, `wish`, and
+`exec` are genuinely different operations and none should absorb another.
+But all of them finish by turning a cell into structured output, so they take
+the same read options and an address renders the same way however you arrived at
+it. A command that returns data gets the whole tail, not a subset — a result
+worth shaping is a result worth filtering. `set` is the exception, because it
+writes rather than returns. Today each command grew its own output handling and
+they have drifted.
+
+**`piece` keeps only what is genuinely about pieces.** Deploying, updating,
+removing, listing, searching, slugs, and listing a piece's verbs. Reading and
+writing cells are not piece operations and stop presenting themselves as ones.
+
+`--shape` throughout, for the reasons above.
+
+## How to get there
+
+Additive, in dependency order. Nothing before step 5 removes or renames anything
+a caller depends on.
+
+1. **Factor out the shared read step** so a single implementation turns a cell
+   and a shape into structured output.
+2. **Give every arrival access to it** — `piece call` gains `--schema` and
+   `--filter`, `wish` gains them, and an address renders identically from each.
+3. **`--piece` accepts the `of:` address form**, so an emitted address composes
+   into the next command. This is where addressing stops being piece-flavoured
+   in practice.
+4. **Add positional addresses and the `#argument` suffix** beside the existing
+   flags, keeping both.
+5. **Add `cf get`/`set`/`call`** as aliases of the existing
+   implementations. Same code, honest names, both spellings working.
+6. **Deprecate the old spellings** once the new ones carry traffic.
+7. **Merge the duplicated nouns** — the two `inspect`s, the two `view`s, `piece
+   map` against `inspect graph`, and `apply` against `set`.
+
+Steps 1–5 are mechanical. Step 7 needs real decisions and belongs last, because
+each pair is two working commands whose merge changes behaviour rather than
+spelling.
+
+## Decisions this document does not make
+
+**`wish` stays its own command.** A wish is a query, not an address: it resolves
+to whatever satisfies it, and may match nothing. Folding it into a read over
+addresses would be a category error. It shares the output step, not the way you
+arrive.
+
+**`exec` runs something reached through a filesystem mount.** Whether that
+becomes an address form or stays its own command depends on how far the address
+grammar should reach into the filesystem projection.
+
+**`piece step` is not redundant with the `--step` flag.** A separate step
+process cannot carry its work into a later read in another process, so the
+standalone command and the flag are not interchangeable. Any merge has to
+preserve that.
+
+**`piece link` stays.** It writes a link with reactive-wiring meaning that a
+general value write does not express.
--- a/docs/plans/cli-surface-shape.md
+++ b/docs/plans/cli-surface-shape.md
@@ -81,7 +81,7 @@ The **concise** form is a different thing wearing the same flag:
 ```
 
 The concise form is a shorthand, and it exists because full schemas are verbose
-enough that nobody writes one to select two fields. It is modelled on
+enough that nobody writes one to select two fields. It is modeled on
 [`llm`'s schemas](https://llm.datasette.io/en/stable/schemas.html), which is
 worth recording so the next reader does not have to re-derive why it looks the
 way it does.
@@ -189,7 +189,7 @@ a caller depends on — steps 1–5 only add.
    map` against `inspect graph`, and `apply` against `set`.
 
 Steps 1–5 are mechanical. Step 7 needs real decisions and belongs last, because
-each pair is two working commands whose merge changes behaviour rather than
+each pair is two working commands whose merge changes behavior rather than
 spelling.
 
 **Each step carries its own documentation.** `--input`, `--piece`, and
--- a/docs/plans/cli-surface-shape.md
+++ b/docs/plans/cli-surface-shape.md
@@ -94,15 +94,20 @@ needing a justification as a schema dialect. A reader who sees
 `--schema title,createdBy.name@` and asks "how is that a schema?" is right, and
 will get righter.
 
-**Give the shorthand its own flag and leave `--schema` for full schemas.** That
-resolves the ambiguity rather than papering over it, and it leaves room for the
+**Give the concise syntax `--select` and leave `--schema` for full schemas.**
+That resolves the ambiguity rather than papering over it, and it leaves room for
+the
 shorthand to grow notation a schema does not need — a suffix meaning "give me
 the link at this path rather than its contents" is the obvious candidate, since
 that is the common case and spelling it in full is painful.
 
+`--select` is the name because it says what the syntax does, reads naturally
+with the address suffix (`--select 'topic@,topic.title'`), and leaves "shape"
+free as the word for what a caller asks for — covering both spellings rather
+than competing with one of them.
+
 Timing argues for doing it now: `piece call` does not have either form yet, so
-splitting them costs one deprecation on a flag that is still young. Naming the
-shorthand flag is open.
+splitting them costs one deprecation on a flag that is still young.
 
 **What a reader may not supply, in either syntax.** `asCell`, `default`,
 `scope`, and `ifc` stay the source's — they decide how a value is treated, not
@@ -123,7 +128,7 @@ Reading is one operation reached from several starting points, so the surface
 wants one read command and distinct commands for the distinct ways of arriving:
 
 ```
-# <read opts> = selection (--schema, or its shorthand) + --filter — identical everywhere
+# <read opts> = [--select S | --schema S] [--filter P] — identical everywhere
 
 cf get   <addr> [path]           <read opts>
 cf call  <addr> <verb> <payload> <read opts>
@@ -161,7 +166,7 @@ writing cells are not piece operations and stop presenting themselves as ones.
 they belong under `space`, and moving them is part of step 7 rather than
 something this shape decides.
 
-The same selection flags everywhere, split per the reasons above.
+`--select` and `--schema` everywhere, split per the reasons above.
 
 ## How to get there
 
@@ -170,8 +175,8 @@ a caller depends on — steps 1–5 only add.
 
 1. **Factor out the shared read step** so a single implementation turns a cell
    and a shape into structured output.
-2. **Give every arrival access to it** — `piece call` gains the selection and
-   filter flags, `wish` gains them, and an address renders identically from each.
+2. **Give every arrival access to it** — `piece call` gains `--select`,
+   `--schema` and `--filter`, `wish` gains them, and an address renders identically from each.
 3. **`--piece` accepts the `of:` address form**, so an emitted address composes
    into the next command. This is where addressing stops being piece-flavoured
    in practice.
--- a/docs/plans/cli-surface-shape.md
+++ b/docs/plans/cli-surface-shape.md
@@ -22,7 +22,7 @@ itself part of the problem.
 | **Live data** — reading and writing running state | `piece get`/`set`/`call`/`apply`/`link`/`step`/`verbs`/`inspect`, `wish` |
 | **Piece lifecycle** — deploying and managing running programs | `piece new`/`setsrc`/`getsrc`/`rm`/`ls`/`search`/`map`/`set-slug`/`recreate-root`/`set-home` |
 | **Rendering** — turning things into something to look at | `piece view` (terminal), `piece render` (HTML), `view` (source pager) |
-| **Storage forensics** — reading the database directly, offline | `inspect` (22 subcommands), `space` (clone/verify/reset/fingerprint) |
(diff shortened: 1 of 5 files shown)


docs/plans/cli-surface-shape.md
```diff
@@ -0,0 +1,186 @@
...
+
+**Two commands, overlapping but not equivalent.** `piece apply` validates a
+whole input against `argumentSchema` and re-executes the pattern with it; `piece
+set` writes at one path. Both target the arguments cell, and the overlap invites
+merging them, but they are not the same operation — which is why any merge
+belongs in the last step rather than among the renames.
+
+### The input spec is not a schema
+
+`--schema` is the most misleading name on the surface, and the one worth fixing
```
seefeldb: everywhere else we use the "schema-on-read" principle, i.e. effectively that schemas are queries: you describe the shape of the data you want. see also graphql, which is similar in philosophy.

so i would avoid this language, so it doesn't leak to other docs and starts confusing things.

docs/plans/cli-surface-shape.md
```diff
@@ -0,0 +1,186 @@
...
+first, because it shapes how people reason about the whole read model.
+
+A schema describes what data **is**. What this flag takes describes what the
+caller **wants back**. Same syntax, opposite direction — and four things follow
+from the mismatch:
+
+- **It accepts input that is not a schema at all.** `--schema
+  'title,createdBy.name'` is a list of paths.
```
seefeldb: or it can be seen as a shorthand to define a schema, since full schemas are very verbose.

docs/plans/cli-surface-shape.md
```diff
@@ -0,0 +1,186 @@
...
+- **It is a strict subset, stripped of what gives a Fabric schema meaning.**
+  `asCell`, `default`, `scope`, and `ifc` are forbidden to callers; `$ref`,
```
seefeldb: if you need these, provide a full schema as input.

of these, `asCell` is worth double clicking on for the short-hand syntax: that could be read as "as link", i.e. return that path, but instead of the contents, i just want the link pointing to the contents. maybe `title,createdBy.user@` means as link? idk, this is a shorthand.

docs/plans/cli-surface-shape.md
```diff
@@ -0,0 +1,186 @@
...
+  finding it does not hold.
+
+The confusion is not that schemas are being used as queries. In this system they
+legitimately are — a subscription is a graph query whose selector is a schema,
+and that is why the *syntax* should stay schema-shaped, so a caller can lift a
+source schema and prune it into a request. The problem is narrower: this
+particular schema is output-directed and deliberately incomplete, and the name
+claims more than it delivers.
+
+**Rename the flag; keep the syntax.** `--shape` says what it is, covers both the
```
seefeldb: sure, renaming it makes sense here, since it is a different syntax. leaves --schema for just full schemas.

mpsalisbury:
> **Updated.** The question at the bottom of this comment has since been answered
> by Ben, and the flag is named. Nothing here needs your time — the two things
> that still want you are in [the next comment](https://github.com/commontoolsinc/labs/pull/5316#issuecomment-5198354363).

Right on the framing, and it had spread further than that section — the umbrella
and the read-layer doc both defined "schema" and "shape" as opposites too. Fixed
in adc212fe: all three now say schemas are queries, a cell carries one and a
reader supplies one, they're the same kind of artifact, and what differs is
authority rather than direction. The source's schema governs how a value is
treated; the reader's selects among what's there.

That relocated the naming problem to where you put it. Not "the input isn't a
schema" — one flag takes two syntaxes. The full form is a schema in good
standing; the concise form is a shorthand that exists because full schemas are
too verbose to write for two fields. So the proposal is now yours: give the
concise syntax its own flag, leave `--schema` for full schemas. It is `--select`
in the docs now — it says what the syntax does, reads naturally with the address
suffix (`--select 'topic@,topic.title'`), and leaves "shape" free as the word for
what a caller asks for.

One correction, and it bears on your `asCell` idea:

> if you need these, provide a full schema as input

That doesn't work today. `FORBIDDEN_PROJECTION_KEYS` is checked against the
parsed projection whichever syntax produced it —
`packages/cli/lib/piece-get-transform.ts`, and the error reads "Invalid
--schema at <path>: "<key>" is controlled by the source schema and cannot be
supplied by a projection". `asCell`, `default`, `scope` and `ifc` are refused
in the full form as well as the shorthand.

<details>
<summary><b>Superseded</b> — the address-keyword question, answered by Ben below</summary>

So `title,createdBy.user@` meaning "as link" runs straight into it: reading `@`
as "as cell" is the natural intuition, and it's the one spelling the current
rules refuse. The question underneath it is whether a reader's schema can ever
say "link here", and if so under what keyword. Two ways I can see:

1. Relax the rule for `asCell` specifically. It's grouped with `ifc`, which is
   information-flow control, so whether `asCell` is separable from that group is
   your call rather than mine.
2. A projection-only keyword that means "give me the address at this path"
   without being `asCell`. No rule change, but a new word.

I've recorded the suffix as sugar over whichever wins, since the shorthand
desugars into a schema and needs something underneath it either way.

The four full-form spellings for the same question are laid out in
`shaped-reads-and-verb-results.md` ("which positions are addresses") if you want
to weigh in — two are ruled out, and the live choice is between a keyword beside
`properties` and a reserved name inside the selection set. The second is the
GraphQL selection-set shape, which you reached for independently, so that's
probably the one to beat.

**Resolved as:** a projection-only `$link`, not a relaxed `asCell`. Ben's
argument is better than either option I offered — `asCell` carries a handle
contract as well as a boundary, and a handle can't cross a serialized channel,
so a reader supplying it would be asking for something the channel silently
downgrades to an address. That's self-defeating rather than merely disallowed,
and it needs no carve-out in `FORBIDDEN_PROJECTION_KEYS`. The marker sits at the
position rather than inside the child's selection set, because the link is
stored in the parent document. Folded into the branch.

</details>

mpsalisbury:
> **Updated.** Ben has since endorsed the first question independently, and it is
> recorded as decided in the branch — your read on it would still be the one that
> de-risks it, since you made the deferral it distinguishes itself from. The
> second question is untouched and is the only thing on this PR still waiting on
> anyone.

<details>
<summary><b>Answered by Ben</b> — may a receipt carry a durable schema? (yes; kept for the reasoning)</summary>

Separate question, on the other half of this.

Receipts are read through a schema like any other cell — `piece get` does
`resultCell.asSchema(getMetaRaw("schema"))` when one is present. Receipts have
none, so two things don't work on them: the projection can't be pushed into the
fetch (Ben's narrowing needs the root container kind to build a selector before
reading), and a reader's schema matches a runtime value rather than a
declaration.

**May a receipt carry a durable schema?**

To be clear about what this isn't: it's not the result-schema emission for
streams you withdrew back in July. That would have put a `result` keyword into
`Pattern.resultSchema` — a published contract, compared across pattern versions
by `assertPatternSchemasBackwardCompatible`, permanent once shipped, and
hard-committing a shape the Fabric-types stream evolution is expected to
replace. Deferring it was right and I'm not reopening it.

A receipt's schema is a different artifact: per-invocation, describing what that
one receipt holds. The compatibility gate only ever compares `argumentSchema`
and `resultSchema` between two versions of a pattern, so it never reaches a
receipt and no permanence obligation attaches. When Fabric-types brings declared
results, the same field takes a better-sourced value — a different source, not a
migration.

Costed, since the shape of the ask isn't obvious:

- The durable schema is `setMetaRaw("schema", …)` (`runner.ts:1580`), read back
  via `getMetaRaw`. Existing mechanism — pattern result cells already get theirs
  this way.
- It is *not* `getCell`'s schema argument. That seeds the link scope
  (`schemaCellScope(schema) ?? "space"`) and the in-memory cell only; passing a
  scoped schema there would move the receipt's partition.
- No value walk needed. The narrowing only needs the root container kind, so
  `{"type": "object"}` is enough to make it engage.
- It doesn't have to be written at mint time — the receipt is minted before the
  handler runs, so there's no shape then. It can go in at result-write time, in
  the same create-only transaction.

One detail I can't settle: whether `{"type": "object"}` is honest for every
receipt. Value-less verbs write `{}`, which is fine. Is a non-record return
possible, or does the contract rule it out?

**Ben's answer:** yes, and the distinction the doc draws is what makes it safe
rather than an obstacle — the compatibility gate compares only a pattern's
argument and result schemas across versions and never reaches receipt cells, so
a receipt schema is per-invocation and carries no permanence obligation. He adds
that the runtime holds the value at write time, so a descriptive schema can be
derived there, and that it should be recorded as descriptive of what the receipt
holds, never authoritative for treatment.

What is left open is narrower and is now in the doc: whether a receipt should
declare only a root container kind, or the full structure including which
positions hold links.

</details>

---

**And a second, unrelated one: should an invocation id be namespaced by the
caller?**

`cf piece call --invocation <id>` lets a caller name a dispatch so that a retry
returns the original outcome instead of executing again. That id decides where
the receipt lands: `scopeCallerEventId` hashes `{caller, id, path, space}` —
where `caller` is the id *string*, not a principal — and the result becomes the
event id the receipt's address derives from.

So no identity enters the address. Two callers who pick the same id against the
same verb in the same space compute one address and read one receipt, with
nothing in the response indicating it. And an id is guessable: piece, verb and
id are all a third party needs to compute where someone else's outcome lives,
bounded by cell scope and CFC labels but never by authorship.

Pre-existing, but reading receipts deliberately is what makes it reachable — and
it isn't hypothetical, since the documented convention teaches human-friendly
ids like `add-comment-1` that two agents following it derive systematically.

Adding the caller's DID to that hash looks mechanically cheap — identity into a
hash that already exists rather than introducing one. The cost is elsewhere: it
changes dedup semantics, and it breaks deliberate id-sharing between agents, if
that's a thing we want to keep.

docs/plans/cli-surface-shape.md
bfollington: On whether `asCell` is separable from the forbidden group for reader shapes: I'd say no, and for a reason beyond the rule itself.

`asCell` bundles two things. One is a traversal boundary — stop realizing the subtree here. The other is a handle contract: what you get back is a live `Cell`, readable, writable, subscribable — and in a pattern, asking for `asCell` is often precisely how you signal intent to *mutate*. The read layer needs only the boundary plus an address rendering.

Over the CLI the handle half cannot be delivered at all. A `Cell` doesn't cross the process boundary; the only faithful serialization of "a cell" is its link. So a reader-supplied `asCell` would be a request the channel silently degrades — the name would promise handle semantics and deliver an address. That's the same name-claims-more-than-it-delivers shape this PR is otherwise removing.

There's also the measured fact that `asCell` is five variants (`cell`, `stream`, `opaque`, `readonly`, `writeonly`), only some of which are boundaries. Opening it to readers imports that whole surface into the projection rules — a reader asserting `writeonly` in a read is nonsense we'd then have to refuse case by case.

So my recommendation is the projection-only keyword (option 2), named `$link`: it names exactly what a serialized read can return, it leaves `asCell` meaning what it means in patterns, and it keeps the `FORBIDDEN_PROJECTION_KEYS` rule at its current strength — treatment keywords are never a caller's, no carve-out. It also dissolves the `@`-reads-as-asCell worry: with `$link` underneath, the suffix never has to mean `asCell` at all.

— Fable, on behalf of Ben

docs/plans/shaped-reads-and-verb-results.md
```diff
@@ -0,0 +1,410 @@
...
+- A sub-pattern instance cannot be declared as a cell — it lacks the cell
+  surface — while a cell handle cannot be declared as its plain type. So the
+  cell wrapper is expressible for exactly the values that are *not* child
+  pieces.
+- Of the shipped patterns, four verbs declare a result type and one returns a
+  reference.
+
+## Open questions
+
+**Which spelling marks addresses in a shape.** Two of the four are ruled out
```
bfollington: Weighing in on the address-marking spelling. There are two separable decisions here — the name and the placement — and I'd argue they shouldn't be decided as a bundle.

**The name should be `$link`, not `$id`, whichever placement wins.** Cells do have ids at one layer (the fid), but addresses are many-to-one over cells: a holder of an address cannot tell the canonical id from an alias, and nothing in normal use requires them to. What a caller needs from this feature is a link they can read next, not a claim about canonical identity. The rendered output already agrees — it returns a link object (`{id, space}`, per the measured facts a stored link also carries `path`/`scope`), and the request vocabulary should match the response vocabulary. `$id` would also teach something false: our documents have no id field — identity lives on the link.

**Placement: marked at the position.** The link is physically stored in the parent document, on the edge — that's the very reason a list of marked addresses costs one document read. `$link: true` beside `properties` annotates the position where the data actually is. The selection-set spelling puts a synthetic field inside the *child's* selection set, claiming the address is data of a target that is never touched. GraphQL gets away with the field fiction because its objects genuinely have `id` fields; ours don't. There's also an implementation asymmetry: the rejecting selector in `traverse.ts` takes a positional instruction, and `$link: true` maps onto it directly, while the field spelling has to infer "load nothing here" from the absence of real fields in a selection set — fetch-suppression by omission rather than by statement.

**The shorthand then desugars 1:1.** `createdBy.user@` rewrites the leaf in place to `{"user": {"$link": true}}`. Under the field spelling the same suffix must convert the leaf into a container selecting a synthetic member — a structural transformation rather than an annotation.

**Semantics of the marker: link *instead of* contents.** If I asked for the link, I shouldn't also get the properties. Address-plus-summary — the motivating case — is spelled explicitly as two paths, `topic@,topic.title`, which merge by union into `{"topic": {"$link": true, "properties": {"title": true}}}` and render as one result carrying both. Bare marker = address only; marker with explicit properties = both, because both were asked for. One rule, no special cases.

— Fable, on behalf of Ben

docs/plans/shaped-reads-and-verb-results.md
```diff
@@ -0,0 +1,410 @@
...
+implementation, since the rendered output is identical either way and only the
+request syntax differs.
+
+**Whether a caller-supplied shape may fix the root container kind.** Narrowing
+keys on the *source* schema today. A caller's shape states the root kind too,
+and honouring it would let schema-less sources narrow. The rule barring caller
+metadata covers `ifc`/`asCell`/`scope`/`default`; root container kind is
+structure rather than metadata, so this may be a clean exception.
+
+**Whether creating receipt cells with a schema is acceptable**, given the
```
bfollington: On whether creating receipt cells with a schema is acceptable: yes, and the distinction the doc draws is the reason it's safe rather than an obstacle. The compatibility gate compares only a pattern's argument and result schemas across versions and never reaches receipt cells, so a receipt schema is per-invocation and carries no permanence obligation. A verb *declaring* a result type is the thing with contract weight, and that's rightly deferred to Fabric-types.

On mechanism: at invocation time the runtime holds the value being written — including which positions are links — so a descriptive schema can be derived where the receipt is minted (`handleJavaScriptHandlerResult`). It should be recorded as descriptive of what this receipt holds, never authoritative for treatment. When declared result types arrive, the same slot takes the declared schema — a change of source, not a migration, exactly as the doc says.

— Fable, on behalf of Ben

docs/plans/cli-surface-shape.md
```diff
@@ -0,0 +1,194 @@
...
+shorthand flag is open.
+
+**What a reader may not supply, in either syntax.** `asCell`, `default`,
+`scope`, and `ifc` stay the source's — they decide how a value is treated, not
+which values come back. `$ref`, `$defs`, and the combinators are unsupported.
+Both checks run against the parsed projection whatever syntax produced it, so
+writing the full form does not unlock them. That is worth knowing before
+designing shorthand notation on top of `asCell`, since it is refused today.
+
+## What it should look like
```
bfollington, line 127: The target surface matches what Mike and I discussed — one read operation reached from several arrivals, addresses over piece-ness, `#argument` in place of `--input` — and the additive staging with renames held to step 6 is the right way to sequence it. Two notes:

For the record on the shorthand's lineage: the concise `--schema` syntax was directly inspired by `llm`'s schemas (https://llm.datasette.io/en/stable/schemas.html). Worth noting that `llm` keeps both syntaxes under one flag, so the precedent says one flag is livable — the reason to split anyway is that our shorthand is growing projection notation (`@`) that isn't schema syntax, and a separate flag gives it room without each addition needing a dialect justification. Recording the lineage in the doc would save the next reader from re-deriving why the concise form looks the way it does.

On migration: since `--input`, `--piece`, and `piece get` appear across the tutorial, the CLI README, and the pattern docs (and `verbs-over-the-cli.md` is already flagged stale), I'd like each migration step to carry its own docs-owed list rather than one line at the end, and the obsolete spellings maintained as redirects until step 6 has real traffic behind it.

— Fable, on behalf of Ben

seefeldb:
> **And a second, unrelated one: should an invocation id be namespaced by the caller?**
> 
> `cf piece call --invocation <id>` lets a caller name a dispatch so that a retry returns the original outcome instead of executing again. That id decides where the receipt lands: `scopeCallerEventId` hashes `{caller, id, path, space}` — where `caller` is the id _string_, not a principal — and the result becomes the event id the receipt's address derives from.
> 
> So no identity enters the address. Two callers who pick the same id against the same verb in the same space compute one address and read one receipt, with nothing in the response indicating it. And an id is guessable: piece, verb and id are all a third party needs to compute where someone else's outcome lives, bounded by cell scope and CFC labels but never by authorship.
> 
> Pre-existing, but reading receipts deliberately is what makes it reachable — and it isn't hypothetical, since the documented convention teaches human-friendly ids like `add-comment-1` that two agents following it derive systematically.
> 
> Adding the caller's DID to that hash looks mechanically cheap — identity into a hash that already exists rather than introducing one. The cost is elsewhere: it changes dedup semantics, and it breaks deliberate id-sharing between agents, if that's a thing we want to keep.

i'd make it local. in `cf-harness` we could provide a session id, and maybe the other harnesses provide that as well.

if we want to share these between two agents, the cleaner way is to just share the link to the result cell.

that all said: for reactive computes, i.e. a pattern tool, not a handler, there's an opportunity to dedupe by inputs+action, as by definition same inputs & action yield the same output. we don't even need the id here except for agent-side bookkeeping.

mpsalisbury:
Thanks — all three landed somewhere. Taking them in order.

**"I'd make it local."** Built, with one deviation worth naming: the session is minted explicitly rather than supplied by the harness. `cf session new`, carried in `CF_SESSION` or `--session`.

The reason for that shape is that an implicitly managed session is a session nobody can name when they need it — a caller who retries has to know which session their original call belonged to, and if a harness chose it silently they cannot. So minting is a visible act, and `--invocation` without a session is now an error rather than a silent fall back to unscoped. Falling back would have preserved exactly the collision the change exists to remove.

Mechanically: `scopeCallerEventId` takes the session as a **required** parameter and hashes it alongside the caller's id and the stream link. Required rather than optional so the compiler found every call site instead of leaving a default-to-unscoped path behind.

Worth adding why not a DID, since it was the obvious alternative and it fails for a reason specific to how we work: agents are directed to work under their human user's key rather than mint their own, so the collision that actually happens is two agents under one key. A session separates them; a DID does not. And a DID is public, so scoping by one would still leave an outcome's address computable by anyone who knows the piece, the verb, and a conventional id. The minted session closes that half too — it is the only unguessable component of the address.

**"Share the link to the result cell."** Agreed, and that is what the design says. It is also now expressible rather than aspirational: a `$link` marker in a selection renders an address a caller can pass straight to the next command, so sharing an outcome deliberately means handing over its address instead of two callers deriving one id from a convention.

**"For reactive computes, dedupe by inputs+action."** This is the one I had no answer for, and I do not think it should get lost in a thread, so it is recorded as an open question in the design document rather than folded into what shipped.

The reason I want to keep it visible is that it is not purely additive to what is built. If a pattern tool's outcome is addressed by what it computed, it needs neither an invocation id nor a session — the id survives only as agent-side bookkeeping, as you say. That means the session's scope of applicability narrows: it would scope handler calls and not reactive computes, and the two cases stop sharing one story about what names an invocation.

So the question I would want settled before either mechanism hardens: do the two share one mechanism, or stay deliberately separate? Separate seems right to me on the face of it — a handler is effectful and its id is a promise about not running twice, where a compute is a function and its address is a fact about its inputs. But that is an instinct, not an argument, and you have thought about the compute side more than I have.

mpsalisbury (commented):

**Scope:** the four design documents in this PR, read with the stacked #5435 and #5503. **Coverage:** all four deep-read; factual claims verified against `packages/runner`, `packages/cli`, `packages/piece`, and `packages/schema-generator` by read-only search agents. **Checks:** CI green here (49 checks).

## What held up

The "measured facts" verified with unusual precision — worth recording, since not re-deriving them is the section's whole point. Confirmed against code, exactly as written: `PieceManager.getResult` is the identity function and `getArgument` throws when the link is absent (`packages/piece/src/manager.ts:857`); `UNSUPPORTED_PROJECTION_KEYS` is exactly sixteen keys and `FORBIDDEN_PROJECTION_KEYS` is `asCell, default, ifc, scope`; the rejecting selector gates exactly nine sites in `traverse.ts` (the doc says "roughly nine"); `scopeCallerEventId` hashes `{caller, id, path, space}` with no principal anywhere in the hash; receipt minting is unconditional at the top of `handleJavaScriptHandlerResult` and the throw-vs-validation-failure table matches the code branch for branch; "four verbs declare a result type and one returns a reference" is exact (all in `topics`; `addTopic` is the reference); `verbs-over-the-cli.md` is stale exactly as described (`plainResultReceipts ??= true`, `runtime.ts:987`).

## 🔴 Blocking

**"Both are checked against the parsed projection regardless of which syntax produced it" is false.** (`shaped-reads-and-verb-results.md` §"What a shape can say today"; the same sentence in `cli-surface-shape.md` §"One flag, two syntaxes".) *Verified.* `normalizeProjectionSchema` — the only caller of both key sets (`packages/cli/lib/piece-get-transform.ts:487`) — runs only on the `@file` and inline-JSON branches. The concise branch calls `conciseProjectionSchema`, which never runs the checks; what protects the concise syntax today is its identifier grammar (`/^[A-Za-z_$][A-Za-z0-9_$-]*$/`), not a check on the parsed projection. Consequently `$link` in a full-form schema today is silently carried through and ignored — it is in neither key set — not refused.

Why it matters: both documents use this claim to argue the two syntaxes cannot diverge on what they unlock, and the `@` suffix (#5435's A3) desugars through exactly the unchecked path. An implementer building against this text will assume a guard that does not exist. The fix is two sentences (grammar guards the concise form; the checks guard the full forms) — plus, worth considering: reserve `$link` in `UNSUPPORTED_PROJECTION_KEYS` now, so its later introduction is refuse→meaningful rather than ignore→meaningful.

## 🟡 Improvements

**The plain-results-only bound on the descriptive schema lives only inside an open question.** §"What a receipt declares" states the derivation unqualified; §"Open questions" then reveals that a receipt for a verb returning anything reactive — including every verb returning a child piece, i.e. the flagship `createNote` case — carries no schema at all. A reader of the declaration section alone builds the wrong expectation (this also reaches #5435's C1, flagged there). *Verified* (both passages).

**Factual slips in `verb-result-selection.md`'s mechanism claims.** All *verified* against code:
- "announced … **before any network work**" (stated twice): the `invocation:` stderr line fires on the `dispatched` phase, after `initial_sync` (`packages/cli/lib/callable.ts:43-48`, announce at `commands/piece.ts:479`). The id survives commit loss, but not a sync hang — which slightly weakens the detached-recovery bullet leaning on this.
- "its exit returns `{id, status: "committed", …}`": the wire key is `invocation`, not `id` (`invocationJson`, `commands/piece.ts:622`) — the doc's own envelope sketch says `invocation`.
- `mintEventId`'s first form is `evt:<key>:<seq>:<linkid>` — the trailing link id is dropped on that branch in the doc.
- The cause is `{...inputs, $event: tx.dispatchedEventId ?? crypto.randomUUID()}` — the UUID fallback (non-dispatch invocations) is unrecorded and is another class of non-re-derivable receipt. Also "`tx.handlingReceiptLink` is published at commit time": it is set during handling, before commit, and only when receipts are enabled (`runner.ts:5231`).

**Question — `$link` key collision in rendered output.** The rendered form merges the address into the same namespace as selected content fields (`{"topic": {"$link": {…}, "title": …}}`). A document with a genuine field named `$link` — legal JSON — makes the declared contract ambiguous, and unlike the request side (where position disambiguates marker from property name) the response side has no escape story. `state-inspector`'s `annotate()` accepts this for debug output; this is a public contract. Consciously accepted? If so, the design should say so.

## ⚪ Nits

- Two British spellings survived this branch's own "American spellings throughout" commit: `honour` (`shaped-reads-and-verb-results.md:452`, §"What a receipt declares") and `piece-flavoured` (`cli-surface-shape.md:181`, §"How to get there"). `docs/README.md` §Spelling.
- `fabric-read-model.md`'s arrival table writes `cf piece get <addr>` — the positional doesn't exist until #5503's P1; today it's `--piece`. In a doc that "assumes no prior familiarity," an agent will type it.
- `verb-result-selection.md` is absent from `docs/plans/README.md`; its only inbound link is the umbrella.
- `cli-surface-shape.md` files all 22 `inspect` subcommands under "offline" — `inspect pull` fetches from a remote.

## Verdict

Sound and unusually well-evidenced. Fix the Blocking sentence in both documents, the working-notes slips, and the two spellings — all text fixes, nothing structural — then good to merge.

mpsalisbury (commented):
cf-review follow-up via Claude Fable 5, on behalf of @mpsalisbury — re-review of the response commit (`8df1f6b`).

Every finding from the earlier review is verified fixed, hunk by hunk: the two-mechanisms correction landed in both documents plus a reserve-before-meaningful paragraph; all four working-notes slips are corrected (the cause's UUID fallback, `mintEventId`'s full forms, `handlingReceiptLink`'s set-during-handling timing with the `receiptsEnabled` condition, the `invocation` wire key, and both "before any network work" claims — the second now states exactly what the id does and does not survive); the plain-results-only bound is stated in "What a receipt declares"; the spellings, the `inspect pull` caveat, the arrival-table spelling, and the working-notes index entry are all in.

Two items remain open:

**The `$link` output-collision question is still unanswered.** The rendered form merges the address into the same namespace as selected content fields (`{"topic": {"$link": {…}, "title": …}}`). A document with a genuine field named `$link` — legal JSON — makes the declared contract ambiguous, and unlike the request side (where position disambiguates marker from property name) the response side has no escape story. If the collision is consciously accepted, the design should say so; it is a public contract where `state-inspector`'s `annotate()` is debug output.

**The new suppress-fetch sentence is unqualified where A5 shows a boundary.** "A `$link` marker still renders an address and still suppresses the fetch" holds for a marker *on* the link position; #5435's A5 records that a marker *below* a link (`notes.title@`) loads every element document until the `projectionMask` reduction is symmetric — measured at four reads where one would do. The receipt's flagship case marks the link itself, so the sentence is true in context; one qualifying phrase ("a marker on a link position") would keep the two documents from disagreeing when read side by side.

PR #5327, opened 2026-08-04
feat(cf-harness): server-side compaction as a context guard

willkelly:
Adopts OpenAI's [server-side compaction](https://developers.openai.com/api/docs/guides/compaction) on the Responses path — the second of the two settings from [*How enabling two settings tripled our scores on the ARC-AGI-3 benchmark*](https://openai.com/index/how-two-settings-tripled-our-arc-agi-3-scores/). The first, retained reasoning, cf-harness already does.

## Background: reasoning tokens and compaction

OpenAI found GPT-5.6 Sol scoring **13.3%** on the ARC-AGI-3 public set, and **38.3%** with two API settings enabled — with **6× fewer output tokens**. The cause was the harness, not the model: it *"discarded private reasoning after every move and later removed older actions once the running history exceeded its window."* The model kept a record of *what* it had done, but lost *why*, so it re-derived each game's rules before acting.

The two settings address the two halves of that loss:

**1. Retained reasoning — already in cf-harness.** Reasoning tokens are the model's private thinking for a turn. By default they are discarded once the turn ends. Preserving them means the next turn builds on prior conclusions instead of re-deriving them. cf-harness does this the ZDR-compatible way: `store: false` with `include: ["reasoning.encrypted_content"]`, replaying the encrypted items through `providerContinuation`. The alternative in the docs — [`previous_response_id` chaining](https://developers.openai.com/api/docs/guides/conversation-state) — requires server-side persistence and is unavailable to a Zero Data Retention org.

**2. Compaction — this PR.** Rolling truncation drops the oldest turns outright; compaction *"summarizes older context instead of dropping it"*, folding prior context — including prior reasoning — into a single encrypted item.

The two mechanisms are the same shape, which matters for the implementation. A compaction item and a reasoning item are both `{id, type, encrypted_content}` and both replay through the same continuation seam. That is precisely why the retention filter here had to widen to cover both: it previously matched `type === "reasoning"` only, so a compaction item was dropped silently.

One consequence worth stating, because it cost me an investigation: **compaction spends output budget** to emit its item. With a small `max_output_tokens` there is no room for it and the API returns an opaque `server_error` rather than a budget error. cf-harness never sets `max_output_tokens`, so this does not affect us — but it makes compaction look broken upstream if you test it with a tight ceiling.

## Threshold: 75% of the input budget

Derived from the gateway registry and cached by `listModels()`; when the catalog has not been primed, the client discovers budgets lazily once the transcript is large enough to matter (a one-time fetch gated at 200k characters, so ordinary runs pay nothing). `--compact-threshold <n>` overrides it; `0` disables.

Deliberately **not** 75% of `contextWindow`. The registry reports that as input + output, and on the 400k models 75% of it is 300,000 against a hard 272,000 input ceiling — verified live:

```
gpt-5.2 @300k tokens -> Input tokens exceed the configured limit of 272000 tokens.
```

That threshold would sit past a wall the request can never reach, so the guard would never fire in the one case it exists for.

## Why 75% rather than something aggressive

This is a **guard, not an optimisation**. The compacting turn costs more than the turn it replaces, so a low threshold taxes every short run to benefit long ones. At 75% it stays dormant for ordinary runs and engages once a built-up context — large system prompt, preloaded skills, accumulated documents — approaches the limit.

The gap *above* the threshold is not slack either — it is the guard's working room. Verified live: compaction cannot **rescue** a request that has already overshot the input cap. Sending ~300k tokens into the 272k cap returns `400 context_length_exceeded` with or without `context_management` — the cap check runs first, so compaction only pre-empts. The crossing turn itself must still fit, and the 25% between threshold and cap (68k tokens on the 400k models) is the single-turn headroom that absorbs it. A higher fraction would shrink exactly the margin the guard depends on.

## Two pieces that make it work

**Compaction items are retained.** They are shaped `{id, type, encrypted_content}`, structurally identical to reasoning items, so the previous filter dropped them silently. Enabling `context_management` without this would pay to produce compaction and then discard the only thing that makes it useful.

**Input is pruned at the compaction boundary.** The [compaction guide](https://developers.openai.com/api/docs/guides/compaction) calls client-side pruning a *"latency tip"*: *"you can drop items that came before the most recent compaction item"*. For us it is mandatory — measured **+14%** without it, worse than not compacting at all. The tip reads as optional because the guide assumes `previous_response_id` chaining, which ZDR rules out for us; with a stateless input array, not pruning means re-sending the full transcript *and* paying for the compaction item.

Pruning starts at an assistant message so `function_call`/`function_call_output` pairs stay intact — slicing anywhere else can orphan a tool result from the call that produced it. Instructions are still built from the whole transcript, so the system prompt survives pruning.

## Verification

**Live, end to end**, against the stage gateway on a Loom-shaped transcript (84 KB system prompt built from Loom's real skills, document-scale tool output). Run exactly as shipped — no manual `listModels()` call; the client discovered budgets lazily, derived the threshold, and did its own pruning:

```
gpt-5.4-nano: contextWindow=400000 maxOut=128000   -> threshold 204,000
turn 11: in=206574  compaction_items=1     <- crosses threshold
turn 12: in=60899   compaction_items=0     <- pruned
```

Whole-run totals from the same transcript: **1,383,892 vs 1,708,676 baseline input tokens — −19%**, matching the −21.5% an offline replay predicted.

**Suite:** 594 passed, 0 failed. New tests cover the threshold policy (including the 400k trap), retention, pruning with tool-pair integrity, single emission of the boundary item, cross-model rejection, lazy budget discovery (one fetch, gated by transcript size, degrades to no compaction if the registry is unreachable), CLI/child-loop propagation (including the real environment projection, with no injected `env`), and the no-compaction case.

**Child loops:** an explicit positive threshold follows only children that inherit the parent model — a profile-overridden child (e.g. `web_search`'s chat-routed Gemini) keeps its own model's derived default instead of a number calibrated to the wrong budget. `0` propagates unconditionally: the off-switch is model-independent.

## Notes for review

- Benefit scales with run length; a run that ends immediately after compacting is net negative. That is the cost of it being a guard.
- Compaction cannot save a turn that jumps from below the threshold straight past the input cap (verified above). A single tool output adding >68k tokens (~270 KB of text) in one turn still fails; client-side truncation would be the remedy and is out of scope here.
- Measured on synthetic-but-real-shaped data: the content, prompt and message shapes come from real runs, but the 13-turn sequence was assembled because the model gave up after 3 documents.
- An earlier investigation of mine concluded compaction was broken upstream. That was wrong — it was an artifact of a test harness passing `max_output_tokens: 64`, which leaves no budget for the compaction item and yields an opaque `server_error`. cf-harness never sets `max_output_tokens`, so production was never affected.

## References

- [OpenAI — Compaction](https://developers.openai.com/api/docs/guides/compaction) — `context_management`, `compact_threshold`, and the ZDR note that server-side compaction is compatible with `store: false`
- [OpenAI — Conversation state](https://developers.openai.com/api/docs/guides/conversation-state) — `previous_response_id` chaining, the alternative we cannot use
- [OpenAI — How enabling two settings tripled our scores on the ARC-AGI-3 benchmark](https://openai.com/index/how-two-settings-tripled-our-arc-agi-3-scores/) — 13.3% → 38.3%, 6× fewer output tokens

---
## Summary by cubic
Adds server-side context compaction on the OpenAI Responses path in `cf-harness` to guard when input nears the model’s budget. Also improves `--compact-threshold` CLI validation with clearer errors.

- **New Features**
  - Sends `context_management: [{ type: "compaction", compact_threshold: N }]` once input reaches the threshold. Default N is 75% of the model’s input budget (contextWindow − maxOutputTokens) from the registry and cached by `listModels()`; if unknown, nothing is sent. Per-turn override via `compactThreshold`; `0` disables.
  - Retains `compaction` items and prunes at the newest assistant message that carries them, preserving `function_call`/`function_call_output` pairs. Compaction is emitted once; reasoning still replays; compaction from other models is ignored.
  - CLI and options: adds `--compact-threshold` and `CF_HARNESS_COMPACT_THRESHOLD`; programmatic `compactThreshold` on `CreateHarnessPromptLoopOptions` and `HarnessModelTurnRequest`. Tests cover threshold policy, pruning, retention, dedupe, cross-model rejection, and no-compaction.

- **Bug Fixes**
  - `--compact-threshold` parsing now names the requirement when rejected (e.g., `--compact-threshold -5`) and points to `--compact-threshold=<n>`. Tests cover valid counts, zero (disables), env/omitted cases, and each rejection path.

<sup>Written for commit c8b3abe9bdcc8f589b6d3fd0ceb8da37f4a96bf7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5327?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cf-harness/src/cli.ts
+++ b/packages/cf-harness/src/cli.ts
@@ -1276,6 +1276,9 @@ export const parseCfHarnessCliArgs = async (
       CF_HARNESS_PROMPT_CACHE_MODE: Deno.env.get(
         "CF_HARNESS_PROMPT_CACHE_MODE",
       ),
+      CF_HARNESS_COMPACT_THRESHOLD: Deno.env.get(
+        "CF_HARNESS_COMPACT_THRESHOLD",
+      ),
       CF_HARNESS_HOME: Deno.env.get("CF_HARNESS_HOME"),
       HOME: Deno.env.get("HOME"),
       CF_HARNESS_CFC_ENFORCEMENT_MODE: Deno.env.get(
--- a/packages/cf-harness/src/cli.ts
+++ b/packages/cf-harness/src/cli.ts
@@ -130,6 +130,7 @@ const CLI_STRING_FLAGS = [
   "model",
   "model-provider",
   "reasoning-effort",
+  "compact-threshold",
   "prompt-cache-mode",
   "skills-root",
   "skill",
@@ -197,6 +198,7 @@ export interface CfHarnessCliConfig {
   model?: string;
   modelProvider?: HarnessModelProviderId;
   reasoningEffort?: string;
+  compactThreshold?: number;
   promptCacheMode?: "implicit" | "explicit";
   gatewayConfigurationExplicit: boolean;
   harnessHome: string;
@@ -260,6 +262,7 @@ export interface CfHarnessCliCapabilities {
     modelUsage: true;
     promptCacheControls: true;
     reasoningEffort: true;
+    compactThreshold: true;
   };
 }
 
@@ -398,6 +401,8 @@ Options:
   --model <name>                Model name (default: ${DEFAULT_MODEL})
   --model-provider <provider>   openai-compatible-gateway | openai-codex
   --reasoning-effort <effort>   Provider reasoning effort (for example low, medium, high)
+  --compact-threshold <n>       Token threshold for server-side compaction
+                                (default: 75% of the model input budget; 0 disables)
   --prompt-cache-mode <mode>    implicit | explicit (GPT-5.6 API gateway only)
   --gateway-base-url <url>      OpenAI-compatible gateway URL
   --gateway-auth-mode <mode>    bearer | none (default: bearer)
@@ -433,6 +438,7 @@ Environment:
   CF_HARNESS_MODEL              Default value for --model (ignored on --resume-run)
   CF_HARNESS_MODEL_PROVIDER     Default value for --model-provider
   CF_HARNESS_REASONING_EFFORT   Default value for --reasoning-effort
+  CF_HARNESS_COMPACT_THRESHOLD  Default value for --compact-threshold
   CF_HARNESS_PROMPT_CACHE_MODE  Default value for --prompt-cache-mode
   CF_HARNESS_HOME               Local cf-harness credential/config directory
   CF_HARNESS_DOCKER_NETWORK_MODE none | bridge | host (default: bridge)
@@ -527,6 +533,7 @@ export const createCfHarnessCliCapabilities = (): CfHarnessCliCapabilities => ({
     modelUsage: true,
     promptCacheControls: true,
     reasoningEffort: true,
+    compactThreshold: true,
   },
 });
 
@@ -1316,6 +1323,22 @@ export const parseCfHarnessCliArgs = async (
   ) {
     throw new Error("--reasoning-effort requires a non-empty value");
   }
+  // 0 is meaningful (disables compaction), so an explicit 0 must survive.
+  const rawCompactThreshold = typeof args["compact-threshold"] === "string"
+    ? args["compact-threshold"].trim()
+    : nonEmptyEnvValue(env.CF_HARNESS_COMPACT_THRESHOLD);
+  let compactThreshold: number | undefined;
+  if (rawCompactThreshold !== undefined && rawCompactThreshold !== "") {
+    const parsedThreshold = Number(rawCompactThreshold);
+    if (!Number.isSafeInteger(parsedThreshold) || parsedThreshold < 0) {
+      throw new Error(
+        "--compact-threshold requires a non-negative integer token count",
+      );
+    }
+    compactThreshold = parsedThreshold;
+  } else if (args["compact-threshold"] !== undefined) {
+    throw new Error("--compact-threshold requires a non-empty value");
+  }
   const promptCacheMode = optionalStringValue(
     typeof args["prompt-cache-mode"] === "string"
       ? args["prompt-cache-mode"].trim()
@@ -1449,6 +1472,7 @@ export const parseCfHarnessCliArgs = async (
       : {}),
     ...(modelProvider !== undefined ? { modelProvider } : {}),
     ...(reasoningEffort !== undefined ? { reasoningEffort } : {}),
+    ...(compactThreshold !== undefined ? { compactThreshold } : {}),
     ...(promptCacheMode !== undefined ? { promptCacheMode } : {}),
     gatewayConfigurationExplicit,
     harnessHome,
@@ -2467,6 +2491,9 @@ export const runCfHarnessCli = async (
           ? { apiKey: parsed.apiKey, apiKeySource: parsed.apiKeySource }
           : {}),
         ...(modelClient !== undefined ? { modelClient } : {}),
+        ...(parsed.compactThreshold !== undefined
+          ? { compactThreshold: parsed.compactThreshold }
+          : {}),
         ...(parsed.reasoningEffort !== undefined
           ? { reasoningEffort: parsed.reasoningEffort }
           : {}),
--- a/packages/cf-harness/src/cli.ts
+++ b/packages/cf-harness/src/cli.ts
@@ -1337,7 +1337,14 @@ export const parseCfHarnessCliArgs = async (
     }
     compactThreshold = parsedThreshold;
   } else if (args["compact-threshold"] !== undefined) {
-    throw new Error("--compact-threshold requires a non-empty value");
+    // A bare flag lands here, and so does a value the parser read as another
+    // flag: `--compact-threshold -5` leaves the option set with no string.
+    // Name the requirement, and point at the form that survives parsing.
+    throw new Error(
+      "--compact-threshold requires a non-negative integer token count; " +
+        "pass values the parser would read as a flag as " +
+        "--compact-threshold=<n>",
+    );
   }
   const promptCacheMode = optionalStringValue(
     typeof args["prompt-cache-mode"] === "string"
--- a/packages/cf-harness/src/cli.ts
+++ b/packages/cf-harness/src/cli.ts
@@ -2498,11 +2498,11 @@ export const runCfHarnessCli = async (
           ? { apiKey: parsed.apiKey, apiKeySource: parsed.apiKeySource }
           : {}),
         ...(modelClient !== undefined ? { modelClient } : {}),
-        ...(parsed.compactThreshold !== undefined
-          ? { compactThreshold: parsed.compactThreshold }
-          : {}),
         ...(parsed.reasoningEffort !== undefined
           ? { reasoningEffort: parsed.reasoningEffort }
+          : {}),
+        ...(parsed.compactThreshold !== undefined
+          ? { compactThreshold: parsed.compactThreshold }
           : {}),
         ...(parsed.promptCacheMode !== undefined
           ? { promptCacheMode: parsed.promptCacheMode }
@@ -2633,6 +2633,9 @@ export const runCfHarnessCli = async (
         ...(modelClient !== undefined ? { modelClient } : {}),
         ...(parsed.reasoningEffort !== undefined
           ? { reasoningEffort: parsed.reasoningEffort }
+          : {}),
+        ...(parsed.compactThreshold !== undefined
+          ? { compactThreshold: parsed.compactThreshold }
           : {}),
         ...(parsed.promptCacheMode !== undefined
           ? { promptCacheMode: parsed.promptCacheMode }
--- a/packages/cf-harness/src/model/openai-compatible-gateway.ts
+++ b/packages/cf-harness/src/model/openai-compatible-gateway.ts
@@ -176,6 +176,28 @@ export const usesResponsesApi = (
  * `web_search`, which overrides the model to Gemini — so this fails loudly
  * rather than letting a future profile discover it as a provider error.
  */
+/**
+ * Rejects a compaction threshold on a turn that cannot honour it.
+ *
+ * `context_management` is only sent on the Responses path, so a chat-routed
+ * model would accept the option and silently ignore it — a user-supplied
+ * control that appears to work while doing nothing. `0` stays legal
+ * everywhere: its requested behaviour is already "do not compact".
+ */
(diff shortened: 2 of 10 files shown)


packages/cf-harness/src/model/openai-compatible-gateway.ts
```diff
@@ -234,9 +265,26 @@ const toModelAttempt = (
 
 export class OpenAICompatibleGatewayModelClient implements HarnessModelClient {
   readonly providerId = OPENAI_COMPATIBLE_GATEWAY_PROVIDER_ID;
+  /**
+   * Input budgets by model id, populated by `listModels()`. Compaction is a
+   * guard, so an unknown budget simply means no threshold is sent rather than
+   * a per-turn registry fetch on the hot path.
+   */
+  readonly #compactThresholds = new Map<string, number>();
```
mathpirate, line 306: Could we ensure this cache is populated on the normal run path? I can only find `listModels()` called by the separate `models openai-codex` command; regular gateway CLI and prompt-loop execution proceed directly to `complete()`.

As a result, when `compactThreshold` is omitted, `#resolveCompactThreshold()` returns `undefined` and no `context_management` is sent. That means the advertised default of 75% of the model input budget never activates in ordinary runs. The unit test currently masks this by explicitly calling `await client.listModels()` before completion.

Please either perform gateway model discovery once during run initialization or lazily memoize it before the first completion, and add an integration-level test that exercises the normal run path without an explicit threshold.

Sol on behalf of Gideon

packages/cf-harness/src/cli.ts
```diff
@@ -2467,6 +2498,9 @@ export const runCfHarnessCli = async (
           ? { apiKey: parsed.apiKey, apiKeySource: parsed.apiKeySource }
           : {}),
         ...(modelClient !== undefined ? { modelClient } : {}),
+        ...(parsed.compactThreshold !== undefined
```
mathpirate: `parsed.compactThreshold` is forwarded by this resume path, but it is missing from the fresh-run `createPromptLoop(...)` call below.

I reproduced this through `runCfHarnessCli`: a fresh run with `--compact-threshold 12000` exits successfully, but the constructed loop receives `compactThreshold: undefined`. `CF_HARNESS_COMPACT_THRESHOLD` is affected the same way.

Could we pass the same conditional `compactThreshold` option in the fresh-run path and add a `runCfHarnessCli` test that inspects the options received by `createPromptLoop`? The existing tests validate parsing but do not cover this handoff.

Sol on behalf of Gideon

packages/cf-harness/src/prompt-loop.ts
```diff
@@ -1736,6 +1738,7 @@ export class CfHarnessPromptLoop {
     this.#cacheAffinityKey = options.cacheAffinityKey;
     this.#promptCacheMode = options.promptCacheMode;
     this.#reasoningEffort = options.reasoningEffort;
+    this.#compactThreshold = options.compactThreshold;
```
mathpirate, line 1740: Could we also propagate `#compactThreshold` into the child `CfHarnessPromptLoop` alongside `#promptCacheMode` and `#reasoningEffort`?

Without that, a positive run-wide override applies only to the parent, not delegated work. More subtly, once default model discovery is wired up, `compactThreshold: 0` will disable compaction in the parent while children fall back to the derived default and compact anyway.

A delegated-loop test covering both a positive override and `0` would pin the intended run-wide behavior.

Sol on behalf of Gideon

packages/cf-harness/src/model/openai-compatible-gateway.ts
```diff
@@ -306,6 +354,7 @@ export class OpenAICompatibleGatewayModelClient implements HarnessModelClient {
       "drop",
     );
     const tools = toResponsesTools(request.tools);
+    const compactThreshold = this.#resolveCompactThreshold(request);
```
mathpirate: Could we fail explicitly when a positive `compactThreshold` is supplied to a path that cannot use it?

At present the option is only consumed by `#completeViaResponses()`. Chat-routed gateway models silently ignore it, and `openai-codex` also accepts it without adding `context_management`. This makes a user-supplied control appear successful while doing nothing.

The existing `promptCacheMode` behavior seems like a good precedent: reject unsupported combinations with a named error. Alternatively, if this is intentionally gateway-Responses-only, the CLI help and capability description should state that clearly. `0` can remain harmless on unsupported paths because its requested behavior is already “do not compact.”

Sol on behalf of Gideon

mathpirate (commented):
Posting four inline review comments as discussed.

Sol on behalf of Gideon

willkelly:
All four addressed — thanks, the first one meant the feature didn't work at all in ordinary use. **592 passing.**

## 1. The default never activated

Confirmed exactly as described: nothing on the run path calls `listModels()`, so `complete()` always saw an empty cache and sent no `context_management`. Worth naming how thoroughly this was masked — **my live end-to-end verification also called `listModels()` explicitly**, so both the unit test and the "proof" exercised a path the CLI never takes.

Budgets are now discovered lazily, once per client, tolerating failure (compaction is a guard — an unreachable registry should mean running without it, not aborting the run).

One wrinkle worth flagging: discovering unconditionally added a round trip before the first model turn of **every** gateway run, and shifted stub ordering in 38 existing tests. That's a real cost, not just fixture churn. So discovery is skipped below a transcript-size floor — the smallest 75% guard the registry can produce sits at 204,000 tokens, so a small turn cannot approach any threshold and shouldn't pay for discovery. Large runs still discover well before the threshold matters.

Tests now cover the normal path with no primed catalog, single-attempt discovery, the small-turn skip, and registry failure degrading rather than failing.

## 2. Fresh-run handoff

Fixed. Mutation-checked — reverting the fix fails the new `runCfHarnessCli` test, so it genuinely covers the handoff rather than re-testing parsing.

## 3. Delegation

Propagated alongside `promptCacheMode`/`reasoningEffort`. The test covers both a positive override and `0`, asserting every parent *and* child turn carries the same value — your point about `0` was the sharper one, since it becomes a real divergence now that the default actually resolves. Also mutation-checked.

## 4. Silent ignore on unsupported paths

Now throws a named error, following `assertPromptCacheModeSupported`. Applied to both chat-routed gateway models and `openai-codex`, which had the same silent-accept problem. `0` remains legal everywhere, as you suggested.

## Re-verified live

Same Loom-shaped transcript, **no manual `listModels()`**:

```
turn 10: in=186821  compaction=0
turn 11: in=206574  compaction=1     <- derived 204,000 threshold
turn 12: in=60899   compaction=0
```

packages/cf-harness/src/cli.ts
```diff
@@ -1316,6 +1323,29 @@ export const parseCfHarnessCliArgs = async (
   ) {
     throw new Error("--reasoning-effort requires a non-empty value");
   }
+  // 0 is meaningful (disables compaction), so an explicit 0 must survive.
+  const rawCompactThreshold = typeof args["compact-threshold"] === "string"
+    ? args["compact-threshold"].trim()
+    : nonEmptyEnvValue(env.CF_HARNESS_COMPACT_THRESHOLD);
```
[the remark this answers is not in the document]
willkelly: Confirmed and fixed in 05db6a4d0 — same disease as the review's finding 1: the variable worked in tests because they inject `env`, and the default projection built from `Deno.env.get` never read it. The projection now carries it, and a new test exercises the real projection (no injected `env`, variable set in the actual process environment); the test fails if the projection line is removed.

packages/cf-harness/src/prompt-loop.ts
```diff
@@ -1997,6 +2000,9 @@ export class CfHarnessPromptLoop {
           ...(this.#reasoningEffort !== undefined
             ? { reasoningEffort: this.#reasoningEffort }
             : {}),
+          ...(this.#compactThreshold !== undefined
```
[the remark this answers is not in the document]
willkelly: Fixed in 05db6a4d0, and it was worse than the wrong number: web_search overrides the child to chat-routed `gemini-3.5-flash`, so an explicit positive threshold would have tripped the new Responses-path assert and failed the subagent outright. A positive threshold now follows only children that inherit the parent model (`childModel.source === "parent"`); a profile-overridden child falls back to its own model's derived default, which is the per-model resolution you suggest — the shared client caches thresholds per model id. `0` still propagates unconditionally: the off-switch is model-independent. Mutation-checked test covers both cases.

Worth noting: `reasoningEffort` and `promptCacheMode` propagate unconditionally today with the same assert shape, so the explicit-option + web_search combination is a pre-existing trap on main for both. Follow-up material, not this PR.

packages/cf-harness/src/model/openai-compatible-gateway.ts
```diff
@@ -232,15 +285,72 @@ const toModelAttempt = (
...
 
+/**
+ * Transcript size below which registry discovery is not worth a round trip.
+ *
+ * The smallest input budget the registry advertises is 272,000 tokens, so the
+ * smallest 75% guard sits at 204,000 — roughly 800,000 characters. A turn
+ * under this floor cannot approach any derived threshold, so ordinary runs
+ * pay no extra request and only genuinely large contexts trigger discovery.
+ */
+const BUDGET_DISCOVERY_FLOOR_CHARS = 200_000;
```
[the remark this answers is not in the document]
willkelly: Keeping the floor at 200k characters deliberately. Two reasons: the cost asymmetry and the char→token variance. The discovery is one registry fetch per client lifetime (the promise is cached, success or failure), not per turn — so a false positive costs one GET once. A false negative means approaching the input cap unguarded, which compaction cannot rescue (verified live: ~300k tokens into the 272k cap returns 400 `context_length_exceeded` with or without `context_management` — the cap check runs first). And dense content (code, JSON tool output) runs ~2.5 chars/token, which puts the smallest 204k-token threshold near 510k characters, not the ~800k the 4-chars/token estimate suggests — a floor at the "documented boundary" would discover too late for exactly the transcripts most likely to need the guard. A token-aware estimate would still be an estimate; the conservative floor buys the same safety for one cheap request.

PR #5340, opened 2026-08-04
refactor(ts-transformers): drop the unused mode, debug, and logger op…

Hixie:
…tions

TransformationOptions carried three switches that nothing outside the test suite ever set. Every production caller of the pipeline lives in the runner's harness engine, and each of them passes only patternCoverage, moduleIdentities, and canonicalWriterIdentityFile.

The `mode` switch declared two pipeline modes, "transform" and "error". Only two branches anywhere read it, both in JSX expression-site lowering: in "error" mode they reported a `reactive:jsx-expression` diagnostic and returned the expression unrewritten, turning the pipeline into a validation pass that told you a rewrite would have happened without performing it. Nothing consumed that report. The tests that exercised it only checked that the diagnostic appeared, and the other tests that set `mode: "error"` were passing it to validation transformers that never looked at it. `debug` had no readers at all: it was declared, defaulted to false in DEFAULT_OPTIONS, and read nowhere. `logger` was threaded as a parameter through seven source files to emit two kinds of message, a warning when the type checker throws while resolving a node's type and a trace line naming each toSchema call, and the only caller that ever supplied one was a single test asserting that a logger receives log lines.

The dual-mode switch meant every reader of the lowering code had to hold two pipeline behaviors in mind, and the logger parameter widened the signature of getTypeAtLocationWithFallback and of the reactive-collection provenance options along the entire call chain that reaches them.

Removing all three: the TransformMode type and its two re-exports, the three option fields, DEFAULT_OPTIONS (which existed only to default mode and debug), the two error-mode branches, and the logger parameter and its threading through ast/utils.ts, ast/call-kind.ts, schema-generator.ts, reactive-variable-for.ts, expression-site-policy.ts, and array-method-policy.ts. The type-resolution catch now returns undefined without formatting a message, which is what every caller already relied on.

On the test side, the mode argument is dropped from the eight suites that set it, the test harness loses its own mode/logger plumbing, and the checkWouldTransform helper goes with the error mode it was built on — it had no callers beyond the two tests that tested the helper itself. That empties test/reactive/runtime-style.test.ts, so the file is deleted.

The current-behavior spec's "Global Modes" section becomes "Global Options": it now records that there is no mode switch and no validate-only configuration, and routes the reader to where each option that does change a compile is described. The mode: "error" bullet in the JSX expression-site routing section and the "does not branch on mode" clause in the reactive `.for()` section go with it.

---
## Summary by cubic
Removed the unused `mode`, `debug`, and `logger` options from `ts-transformers`, simplifying the pipeline to a single behavior and reducing API surface. No production behavior changes; docs and tests updated to match.

- **Refactors**
  - Removed `mode`, `debug`, and `logger` from `TransformationOptions`; deleted `TransformMode` export and default options.
  - Dropped error-mode branches in JSX expression-site lowering; the pipeline always rewrites and only reports real violations.
  - Removed logger plumbing; `getTypeAtLocationWithFallback` now returns `undefined` on type-checker errors without logging; removed `logger` from `ReactiveCollectionProvenanceOptions` and all call sites.
  - Pruned logging in schema generator and test utilities; removed validate-only helpers and deleted the now-empty runtime-style test.
  - Updated spec: “Global Modes” → “Global Options”; documented no validate-only mode and listed options that affect compiles (`patternCoverage`, `moduleIdentities`/`canonicalWriterIdentityFile`, `assertDiagnostics`, `state`/`diagnosticsCollector`).

- **Migration**
  - Stop passing `mode`, `debug`, or `logger` to `CommonFabricTransformerPipeline` and individual transformers.
  - Remove imports/uses of `TransformMode`, `DEFAULT_OPTIONS`, and any `context.options.logger` reads.
  - If calling `getTypeAtLocationWithFallback(node, checker, registry, logger)`, drop the `logger` argument.
  - Remove tests/helpers that relied on a validate-only path (e.g., `checkWouldTransform`).

<sup>Written for commit 294534816f60dd98200806e5a5de5d8c57bec5d6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5340?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mathpirate (approved):
lgtm thanks for the cleanup!

PR #5342, opened 2026-08-04
refactor(runner): the runtime serializes its own cells and artifacts

danfuzz:
A value on its way into the data model may hold two things the model has no
representation for: a builder artifact, and a `Cell`. The runtime replaces both
itself, at the boundaries where they arrive, rather than leaving either to be
recognized by the duck-typed `toJSON` protocol on the way through.

Artifacts already worked this way at the pattern-driven boundaries. This adds
the two that were missing.

## What changes

- **`diffAndUpdate()`** — the raw write funnel behind `Cell.set()` and the
  collection operations — now runs the artifact walk. `addUnique()`'s dedup
  comparison normalizes the same way, so a candidate is compared in the form a
  write would store it in.
- **`Runtime.getImmutableCell()`** now turns cells into links. Its entity id is
  derived from the bytes of what it is given, so a cell reaching the conversion
  unreplaced is a document that cannot be named. Its own comment already said
  as much: "`fabricFromNativeValue()` converts what callers actually pass --
  notably `Cell`s, which become sigil links via their `toJSON()`".

One walk covers both. `replaceArtifacts()` takes a `replaceOther` hook for the
values it does not descend into, and `getImmutableCell` supplies the cell
knowledge: `isCell()` lives in the runtime's core, and `encodable-form.ts` is a
leaf that the import graph reaches from the other direction.

A `Cell` also gains `toSigilLinkOrNull()`, and `hasEncodableForm()` asks for
THAT name rather than `toJSON`. Both names it now reads are the runtime's own,
so no value acquires an encodable form by carrying a member the JSON protocol
happens to share. `Cell.toJSON()` remains, delegating, for `JSON.stringify`.

## Why it is inert, measured

An A/B harness runs 25 probes across this tree and `main`: six
pattern-serialization encodings, seven raw cell writes, nine
`getImmutableCell` entity ids, plus liveness and provenance. **Every one is
byte-identical.**

Suites: full workspace passing; `packages/runner` 1133 passed / 0 failed, the
same count as `main`. `deno fmt --check`, `deno lint`, `deno task check`, and
the eight standalone gates are green.

## What this unblocks

`data-model` can stop honoring `toJSON()` in `fabricFromNativeValue()`, which
is the point of the exercise. That is the follow-on PR, stacked on this one.

## One thing found in passing, and left alone

`convertCellsToLinks()` is not the walk for this job. It treats a SHARED
reference as a cycle, so a pattern graph's repeated empty `path: []` comes back
rewritten as a back-link into the first occurrence. It also converts
query-result proxies that the conversion had been rejecting. Both are harmless
where it is used today, its one caller feeding it event values rather than a
graph, so this changes nothing about it.

---
## Summary by cubic
The runtime now serializes builder artifacts and `Cell`s at storage boundaries, replacing them with their encodable forms/links before values hit the data model. This removes reliance on duck-typed `toJSON` and keeps behavior unchanged.

- **Refactors**
  - `diffAndUpdate()` now flattens builder artifacts; `addUnique()` compares candidates in their normalized, stored form.
  - `Runtime.getImmutableCell()` replaces embedded `Cell`s with sigil links via a single walk that also flattens artifacts.
  - Added `Cell.toSigilLinkOrNull()`; `toJSON()` delegates for `JSON.stringify` only.
  - `hasEncodableForm()` now checks `toEncodableForm` or `toSigilLinkOrNull` (no `toJSON` fallback); the walk accepts a `replaceOther` hook to handle values it doesn’t descend into.
  - No behavior change: A/B probes are byte-identical and all tests in `packages/runner` pass.

<sup>Written for commit 367233c3fd06cb96f6c6127cb6bec4efcfff5dea. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5342?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

NEW_COVERAGE_BASELINE

The ratchet reports `packages/runner` uncovered +1. Attributed from CI's own
LCOV artifacts — every `coverage-profile-*` from this run and from the base
`main` run, merged by max hits, diffed per file:

- **427 lines newly uncovered, 407 newly covered** in `packages/runner` — a
  wash, and the same churn appears in `packages/llm`, `packages/memory`, and
  `packages/ts-transformers`, none of which this branch touches.
- Intersecting this branch's added source lines with that merged profile gives
  **zero** uncovered added lines.

The churn is shard re-partitioning. The runner suite is sharded five ways and
balanced on the `test-timing-*` artifacts; this branch changes per-file test
timings (a step removed from `doc-map.test.ts`, one added to
`cell-links.test.ts`), so the shards divide differently and incidental coverage
of untouched files moves with them.

(The reconstruction above does not reproduce the gate's arithmetic exactly —
it counts 4588 → 4608 where the gate counts 4783 → 4784 — so treat the wash and
the zero as the findings, not the absolute totals.)

### DIFF
--- a/packages/runner/src/cell.ts
+++ b/packages/runner/src/cell.ts
@@ -101,6 +101,7 @@ import {
 } from "./cfc/metadata.ts";
 import { toURI } from "./uri-utils.ts";
 import { createRef } from "./create-ref.ts";
+import { flattenBuilderArtifacts } from "./storage-preflight.ts";
 import {
   type SigilLink,
   type SigilWriteRedirectLink,
@@ -483,6 +484,19 @@ declare module "@commonfabric/api" {
     getAsReactiveProxy(
       boundTarget?: (...args: unknown[]) => unknown,
     ): Reactive<T>;
+    /**
+     * Returns the sigil link naming this cell, or `null` when the cell has no
+     * full link yet (one that has not been created). This is how a cell reaches
+     * storage: the link is what stands in for it, and `hasEncodableForm()` asks
+     * for this member by name.
+     */
+    toSigilLinkOrNull(): SigilLink | null;
+    /**
+     * Answers the JSON protocol with the same link `toSigilLinkOrNull()` gives,
+     * so a cell embedded in a value being stringified reads as what it names.
+     * `packages/html`'s VDOM key generation depends on this. Nothing on the way
+     * to storage consults it.
+     */
     toJSON(): SigilLink | null;
     runtime: Runtime;
     tx: IExtendedStorageTransaction | undefined;
@@ -541,6 +555,7 @@ const cellMethods = new Set<
   "filterWithPattern",
   "flatMap",
   "flatMapWithPattern",
+  "toSigilLinkOrNull",
   "toJSON",
   "for",
   "asSchema",
@@ -1803,7 +1818,7 @@ export class CellImpl<T extends FabricValue>
       // unconverted, and the strict conversion would reject their
       // non-string-keyed internals.
       const comparable = normalizeForComparison && !isCellLink(candidate)
-        ? fabricFromNativeValue(candidate)
+        ? fabricFromNativeValue(flattenBuilderArtifacts(candidate))
         : candidate;
       return existing.some((element) => valueEqual(element, comparable));
     };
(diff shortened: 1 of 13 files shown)


Hixie:
This apparently regressed the benchmarks; Claude suggested https://github.com/commontoolsinc/labs/pull/5482 as the fix.

PR #5339, opened 2026-08-04
feat: server-execution v2 stage A — flag, commit class, ON/OFF CI arms, stream-data disable

seefeldb:
## Why

Phase 1 stage A of the server-execution v2 plan
(`docs/plans/server-execution-v2.md`):
land the substrate every later stage builds on — the single flag, the commit
`class` vocabulary written in every arm, the flag-ON CI arm (v1's terminal
failure mode was that the flags-on branch never went through CI), and the
`stream-data` deferral — with the OFF arm byte-identical to today. The flag
stays OFF throughout; the ON arm is deliberately a behavioral no-op except the
`stream-data` error, matching Phase 0's success criterion ("a no-op ON arm
passes CI identically to OFF").

## What

**The flag** — `EXPERIMENTAL_SERVER_EXECUTION`, RuntimeOptions key
`serverExecution` (spec README §3.4; deliberately unlike v1's
`SERVER_PRIMARY_EXECUTION`):

- `ExperimentalOptions.serverExecution` + the canonical env mapping in
  `runtime-presets.ts`; ambient control point `setServerExecutionConfig` in
  `packages/memory/v2.ts` beside `persistentSchedulerState`'s (the memory
  server enforces the per-class admission rows under the flag, from stage B);
  shell build define for the browser side. Registered in
  `EXPERIMENTAL_OPTIONS.md` with both states defined; OFF is today
  byte-for-byte. Not a handshake capability — nothing is negotiated per
  connection.

**Commit `class`** (protocol.md §1, §7) — written in EVERY arm from this PR
on, enforced only under the flag (and the first enforceable row, the lease
equality check, lands with stage B):

- `CommitClass` (`authored | derived | system`, closed set) defined once in
  the wire-shape module `packages/memory/v2.ts`.
- A `class` column on the engine's `commit` table (migration backfills
  `'authored'`; fresh and migrated stores agree on column order), stamped by
  the admission path that processed the commit: the session-facing transact
  path stamps `authored`, the memory server's own direct-write path stamps
  `system`. Never a client-supplied field: `ClientCommit` cannot express a
  class, and a smuggled field is inert (pinned by test — FP15's derivation).
- `derived` is unclaimable until the `execution_lease` admission check exists:
  a discovery tripwire (`ProtocolError` naming serving-loop.md §2) refuses it
  in either arm, per the registry's flag-gated-tripwire convention.

**CI arms** (testing.md §2) — the pattern and package integration suites run
twice:

[...]

PR #5349, opened 2026-08-04
feat: server-execution v2 stage B — the execution lease and the derived-class admission check

seefeldb:
## Why

Phase 1 stage B of the server-execution v2 plan
(`docs/plans/server-execution-v2.md`): the single-deriver lease
(serving-loop.md §2). The invariant "derived state has exactly one deriving
committer" holds by construction against clients — no client code path can
even construct a `derived` commit — but against *other server processes*
(deploy overlap, partition) it holds via the lease this PR lands. Everything
lands dark: the flag stays OFF, no production path produces a `derived`
commit, and the lease is consumed by later stages — not exercised live until
Phase 2. Stacked on #5339 (stage A), whose commit `class` metadata this PR's
admission check reads.

## What

**The `execution_lease` table** (engine-v3 schema; created for existing
stores on reopen, since the whole schema runs `IF NOT EXISTS` at open):

- One row per space, `(space, holder, expires_at)` — **exactly three
  fields**. The v1 branch's richer shape (branch PK, `lease_generation`,
  `host_id`, `on_behalf_of`, a state enum) was prior art to REDUCE from
  (serving-loop.md §2's ruling, §8's tripwires); none of it carried over.
  No foreign keys: the table references nothing and nothing references it.
- The table joins the sqlite-builtin guard's core-table denylist, so pattern
  SQL can never reach it (the guard's schema-coverage test enforces this
  mechanically for every engine table).

**The DR1 per-process holder identity**
(`packages/memory/v2/execution-lease.ts`):

- `executionLeaseHolder(serviceIdentity)` = the service identity plus a
  process-instance component minted at PROCESS START (module init) — stable
  across every renew and reacquire within one process lifetime, fresh per
  genuinely-new process. This is what makes the one admission equality check
  itself the fence for every CROSS-process succession: a successor process's
  holder never equals its predecessor's.

**The acquire/renew/expire cycle** (same module — the holder side, on the
direct-engine plane, serving-loop.md §1 plane (c)):

- Acquire is ONE conditional write: take the row when it is free, expired,
  or already ours; never overwrite a live foreign lease. TTL 15 s.
- Renew is a DIRECT table update extending only a still-live own row — **a
  lease renewal is NEVER a commit**, pinned by test: the whole acquire/renew
  traffic leaves the commit table empty (the coverage register's OW2,
  implementation side). Renew cadence 5 s (`EXECUTION_LEASE_RENEW_INTERVAL_MS`);
  the timer that drives it belongs to stage F's SpaceServer, which
  serving-loop.md §1 lists as owning the lease "renewed on a timer" — this
  module owns the transitions, not the schedule.
- `ExecutionLeaseCycle` enforces the abort-before-reacquire discipline
  IN-PROCESS with the in-memory tenure counter the spec sanctions: a failed
  renewal ends the tenure BEFORE the caller can reacquire ("stop committing
  immediately" — in-flight work fails `isCurrentTenure` from that moment),
  and a reacquire mints the NEXT tenure, so work sealed under a lapsed
  tenure can never pass the check afterward. There is no path to a reacquire
  that does not first end the old tenure.

**The derived-class admission equality check** (protocol.md §2's `derived`
row), enforced under `EXPERIMENTAL_SERVER_EXECUTION`:

- The producer holds the space's live `execution_lease` — ONE equality check
  against the row's holder, not admission machinery. `holder` arrives as a
  server-internal `ApplyCommitOptions` field like `commitClass` itself;
  `ClientCommit` cannot express one, and a smuggled payload field is inert
  (pinned by test).
- Liveness is judged by the MEMORY SERVER's own clock: the admission read
  excludes expired rows, so **an expired row matches nobody** — a derived
  commit under an expired lease is rejected even before any successor
  acquires. Holder clocks never arbitrate liveness.
- Off the flag, `derived` stays unclaimable (stage A's posture — nothing
  outside the flag may claim the class, protocol.md §1); the stage-A
  discovery tripwire is replaced by the real rule.
- An admitted derived commit persists `class` + `holder` (protocol.md §7's
  closed metadata list: `holder`, derived only; NULL on every other class).

## OFF arm

Byte-identical: no production path passes `derived`, the lease table sits
empty, and acquire/renew have no callers until stage F. The only source
delta on an exercised path is the tripwire's message text.

## Tests

`packages/memory/test/v2-execution-lease-test.ts` mirrors the executable
spec model's C6/C7 properties against the real engine (the model itself is
untouched and stays green — `packages/spec-model` 18/18):

- **C6** — admission negatives: non-holder, missing holder, missing space,
  expired row, absent row; the positive case persists `class` + `holder`;
  authored/system ignore the lease and a stray holder never sticks to them.
- **C7a** — renewal failure ends the tenure before the reacquire; a probe
  sealed pre-expiry never commits; work sealed under the current tenure
  still flows (the fence is not vacuous).
- **C7b** — the residue the MUST exists for: WITHOUT the in-process check,
  the same-process stale probe WOULD be admitted (same holder by design);
  and cross-process succession is fenced by the equality check alone.
- Table shape: exactly three columns (the v1 fields' reappearance would fail
  here before any grep); pre-stage-B stores gain the table and the commit
  `holder` column on reopen.

No sleeps, no timers: expiry is constructed by acquiring at controlled
clock values (the cycle's injected clock stays anchored to the admission
clock — the co-hosted one-process premise — advanced by deterministic
deltas).

Gates run: memory `deno task test` (485 passed), spec-model (18 passed),
toolshed memory routes (27 passed), root `deno task check`, repo-wide
`deno fmt --check` + `deno lint`, `check-docs` (523 blocks),
`check-conflict-markers`. The ON-arm CI skip lists stay empty — no
integration test exercises a not-yet-landed surface.

## Docs (same change, per the lifecycle)

- Stage B ticked in the plan with its landing note.
- `EXPERIMENTAL_OPTIONS.md`: the `serverExecution` ON-state and status rows
  now include the stage-B surface.
- The two spec anchor statements claiming `execution_lease` "does not exist
  on main" (serving-loop.md §Anchors/§2, protocol.md §Anchors) updated to
  the landed state.

Prior PR in the series: #5339 (stage A). Next: stage C (main reduction — a
three-PR train), which this PR deliberately does not touch, and stage D
(sealing), which will thread `isCurrentTenure` through the wave commit step.

---
## Summary by cubic
Adds the execution lease and the derived-class admission check for server-execution v2 stage B. Lands the `execution_lease` table and in-process tenure fencing; admission is enforced only under `EXPERIMENTAL_SERVER_EXECUTION` and remains dark until the serving loop uses it.

- **New Features**
  - New `execution_lease` table: `(space, holder, expires_at)`; added to the SQLite guard denylist.
  - Per-process holder identity via `executionLeaseHolder(serviceIdentity)` (stable within a process; new per process).
  - Acquire/renew/release APIs and `ExecutionLeaseCycle` with in-process tenure fencing; renewals are direct table writes, never commits.
  - Derived-class admission: one equality check against the space’s live lease; liveness judged by the memory server clock; enforced only when `EXPERIMENTAL_SERVER_EXECUTION` is on.
  - Commit metadata adds `holder` for `derived` commits; ignored for other classes.

- **Migration**
  - On open, the engine creates `execution_lease` and adds `commit.holder`; existing stores migrate automatically.
  - Off the flag, `derived` stays unclaimable and behavior is unchanged; the lease sits unused until the serving loop lands.
  - No foreign keys to or from `execution_lease`.

<sup>Written for commit 1ac2077f089d19605c6fda33300d3dc4379740e2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5349?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seefeldb:
## Adversarial review (subagent panel + adjudication) — VERDICT: LANDABLE

No confirmed defects, and this review went to execution: all 22 touched tests run green, then the admission equality check was MUTATED (always-admit) — three tests failed as they must, proving the suite load-bearing, then restored. Verified: `execution_lease` is exactly `(space, holder, expiresAt)` with no v1 fields (grepped for lease_generation/fence/epoch/generation — only sanctioned prose hits); DR1's per-process holder mints ONCE at module load and is never re-minted; acquire is a correct SQLite UPSERT-with-WHERE conditional; renew/release are raw table ops that provably never touch the commit table ("never a commit" test passes); liveness uses the memory server's `Date.now()` fresh inside the transaction; the equality check is one check, flag-gated, and an expired row matches nobody; landed dark (no timers, no production importer of the lease module).

Flagged questions (non-blocking): (1) verification-coverage OW2's implementation half is satisfied by this PR's renewal-never-a-commit test — the register note is queued. (2) The engine relies on structure (no client path to `derived`) rather than an envelope `service:` check like the model's `admitDerived` — logged as a stage-F defense-in-depth question, not a stage-B change.

seefeldb (commented):
Verdict: changes required.

[major] In `packages/memory/v2/engine.ts:5172-5193`, live-lease admission runs before the idempotent replay lookup at 5225-5264. I reproduced the failure: an accepted derived commit returned seq 1; after releasing the lease, a byte-identical replay with the same session/localSeq was rejected for not holding the current lease instead of returning the stored seq. A network retry of an already accepted commit must not be re-admitted against current authority.

Move exact replay detection ahead of current-lease admission (while checking the stored class/holder as appropriate), and add replay-after-expiry/release/succession coverage.

seefeldb:
Fix ledger for the 2026-08-12 review (replay vs live-lease admission):

**Fixed in 959e074359.** Exact-replay detection now runs ahead of the live-lease equality check in `applyCommitTransaction` (flag-posture gate stays first). Replay identity also checks the stored class/holder, normalized as inserted; the lease check is byte-identical and fresh-commits-only, with pairwise check order preserved on the fresh path.

Evidence:
- Red-first: 5 new tests reproduced the reviewed failure on the unfixed engine — replay after release, after expiry, after succession, plus different-bytes refusal and envelope-mismatch refusal — and 2 preservation guards.
- Discrimination probe: reverse-applying the engine diff makes exactly the 5 reorder-dependent tests fail; restored, 22/22 green.
- Suites: memory package 494/494 with type-check; targeted files 91/91; fmt+lint clean.

Semantics now pinned (flagging, not silently deciding): replay of a commit on a deleted branch answers from the store; cross-class or cross-holder resubmission of the same (session, localSeq) refuses; off-flag derived replay still rejects; withdrawn-commit replay has no machinery and stays out of scope.

Descendant rule recorded for the train: any admission check added **above** the replay return re-creates this bug class for retries — new checks belong after it. The serving-loop §2 clarifying clause (exact replay answers from the store first) landed on the C.3 docs stage as b95479262.

seefeldb:

PR #5355, opened 2026-08-04
fix(runner): a shared reference is not a cycle in `convertCellsToLinks()`

danfuzz:
Last of the follow-ons from the `toJSON` removal arc; the rest have landed, so this one sits on `main`.

`convertCellsToLinks()` kept every value it had visited in `seen` and answered
a repeat with a back-link. That is what makes a cycle representable — but a
value reachable twice by *different paths* is not a cycle. It is shared, and
answering the second position with a pointer at the first rewrites content into
a reference to somewhere else in the same value.

A serialized pattern graph is full of such sharing: one empty `path: []` array
is aliased from every alias in the graph, so all but the first came back as
back-links into the first.

## The fix

`seen` now holds the **ancestors** of the value being converted, cleared on the
way back out. That is the idiom the sibling walk in this same file already
uses, comment included:

> `// Remove from ancestors when backtracking (shared refs at other paths are ok)`

## Impact

Four production call sites:

| Site | Carries |
| --- | --- |
| `runner/src/cell.ts:1420` | `Cell.send` event values |
| `html/src/worker/reconciler.ts:3058` | every VDOM prop crossing to the worker |
| `runtime-client/.../runtime-processor.ts:894` | `handleCellGet` — arbitrary cell values |
| `runtime-client/.../runtime-processor.ts:1013` | `handleCellSubscribe` — every cell update crossing to the main thread |

The last two carry arbitrary cell values, which is both where sharing is most
likely and where the cost below lands.

How this surfaced: reaching for this walk to convert cells at a storage intake
turned three runner tests red, and the diff showed a graph's
`outputs.$alias.path` rewritten into a back-link at
`["p","result","out","$alias","path"]`.

## The cost: a shared subtree is now converted once per position

Answering sharing as sharing means a node reachable by *k* paths is converted
*k* times, where a back-link converted it once. For a balanced DAG
(`node = {a: node, b: node}` nested *n* deep) that is **exponential** — measured
on this branch:

| depth | time |
| ---: | ---: |
| 12 | 5 ms |
| 16 | 51 ms |
| 20 | 831 ms |

`main` is O(nodes) on the same input. One mechanism cannot both preserve
sharing and dedupe it, so this is the trade the fix makes rather than an
oversight — but it is a real one on the two `runtime-client` paths above, and
worth a decision rather than a footnote.

## Tests

A new `convert-cells-to-links-sharing.test.ts`, with the differential run:
three of its four cases fail without the fix, and the fourth — that an ancestor
still comes back as a back-link — passes both before and after, so the cycle
handling is pinned as kept rather than removed.

`packages/runner` 1147 passed / 0 failed.

---
## Summary by cubic
Fixes shared references being treated as cycles in `convertCellsToLinks()`. Shared values are converted at each path; cycles still produce back-links, and ancestor tracking is cleared on every exit.

- **Bug Fixes**
  - Use an `ancestors` map (not `seen`) to track only the current path; clear it on every exit so only real cycles produce back-links — even when values convert to primitives (e.g., `Date` → `FabricEpochNsec`).
  - Keep empty `path: []` arrays at each alias in pattern graphs instead of collapsing into a back-link.
  - Tests pin shared vs cyclic behavior and assert the exact back-link shape for cycles, including the shared `Date` case.

<sup>Written for commit e985dde41fce726cc6a18baeb0ee5fb4e71b648a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5355?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/runner/src/cell.ts
ubik2: These early returns skip the finally that removes original from seen.

danfuzz: @ubik2 Good catch, and it was reachable — thank you.

Both exits that answer a value without descending into it (`!isRecord(value)`,
and the `FabricPrimitive` branch) sat outside the `try`, so a value leaving by
either stayed an ancestor for the rest of the walk. A shared `Date` is the
concrete case: it converts to a `FabricEpochNsec` and leaves by the second one,
so `{ a: shared, b: shared }` came back with `b` as a back-link to `a` — the
exact bug this PR set out to fix, still present on that path.

Fixed by opening the `try` immediately after the ancestor is recorded, so there
is no exit that can skip the clearing. Added a
`converts a twice-reachable value that converts to a primitive` case to the new
test file; verified red before the fix and green after. Runner suite 1147/0.

— Claude (agent, working with @danfuzz)

PR #5357, opened 2026-08-04
fix(runner): getOwnPropertyDescriptor must answer about OWN properties

bfollington:
## Why

The query-result proxy's `getOwnPropertyDescriptor` trap tested membership with `prop in current`, and **`in` walks the prototype chain**. Because the underlying value is an ordinary `Object.prototype`-rooted record, the trap synthesized a descriptor for every member of `Object.prototype` — `toString`, `valueOf`, `constructor`, `__proto__` — so `Object.hasOwn(proxy, "toString")` answered `true`. The `ownKeys` trap directly above it uses `Reflect.ownKeys()` and listed none of them.

Two traps describing the same value, disagreeing by construction. Every consumer that reasons about a read-back value's shape was being told it carries names it does not.

That isn't academic. `unsafeObjectKeyIn()` (#5264) refuses a `FabricValue` carrying own `__proto__`/`constructor`, and it asks with `Object.hasOwn()`. So **writing a read-back record back to a cell was refused for a key the record never had** — which breaks read-modify-write against a cell generally, for any consumer.

It took down every Loom state-cell push for three days ([CT-1949](https://linear.app/common-tools/issue/CT-1949)), and because the error names the data model rather than the proxy, it read as a data problem for most of that time. Loom's own defensive fix is [loom#4672](https://github.com/commontoolsinc/loom/pull/4672); this is the underlying cause.

Measured before the fix, against a live toolshed:

| name | `Object.hasOwn` | in `ownKeys` |
|---|---|---|
| `__proto__` | true | no |
| `constructor` | true | no |
| `toString` | true | no |
| `valueOf` | true | no |
| `hasOwnProperty` | true | no |
| `apiUrl` | true | **yes** ← a real own key |
| `definitelyNotAKeyAnywhere` | false | no |

## What Changed

One line: `prop in current` → `Object.hasOwn(current, prop)`, which is the question this trap owes its caller.

**The `has` trap keeps `in`**, and a test now pins that — `has` *is* the `in` operator's own trap, so inherited names *should* answer true there. Conflating the two is what went wrong, so the distinction is worth holding in place against a future cleanup that "fixes" both alike.

## Test plan

- Two cases added to `query-result-proxy-enumeration.test.ts`: one asserting inherited names are not reported as own (and that the two traps now agree), one pinning that `in` still sees them.
- **Verified non-vacuous** — the new case fails with the change backed out.
- **Supported suite is green**: `cd packages/runner && deno task test` — **1146 passed / 0 failed** (6012 steps), and all Runner Test CI shards pass.

  _Correction:_ an earlier revision of this description reported "1026 passed / 120 failed, unchanged with and without the commit" and called those 120 pre-existing timing-sensitive failures. That diagnosis was wrong. The runner suite requires `--preload=test/clock-preload.ts`; without it the global `clock` is undefined and the suite fails en masse. I had run it unsupported, so I was comparing two broken runs. The no-regression conclusion happened to hold, but the evidence for it did not — the green supported run above is the real result.

## Draft because

The fix direction seems clear to me, but it's a core proxy trap and the call on whether this is the right layer is yours. The alternative — having `unsafeObjectKeyIn()` consult `ownKeys` instead of `Object.hasOwn()` — closes this particular blast radius but leaves the proxy misreporting to every other consumer, so it looks like compensating rather than fixing. Happy to be told otherwise.

---
## Summary by cubic
Fix the query-result proxy to only report own properties in `getOwnPropertyDescriptor`, aligning with `Reflect.ownKeys` and restoring safe read-modify-write to cells. Resolves false positives for inherited names like `__proto__` and `constructor` (CT-1949).

- **Bug Fixes**
  - Switched `prop in current` to `Object.hasOwn(current, prop)` in `getOwnPropertyDescriptor`; the `has` trap still uses `in`.
  - Added tests pinning that inherited `Object.prototype` names aren’t reported as own, real own keys are, both traps agree, and `in` still sees inherited names.
  - Verified the runner test suite passes with the change.

<sup>Written for commit e33cbd683fa7b97614d47e96e27434c31564f728. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5357?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/query-result-proxy.ts
+++ b/packages/runner/src/query-result-proxy.ts
@@ -625,7 +625,25 @@ export function createQueryResultProxy<T>(
       }
       const readTx = runtime.readTx(tx);
       const current = readTx.readValueOrThrow(link, SHAPE_READ) as typeof value;
-      if ((isRecord(current) || Array.isArray(current)) && prop in current) {
+      // `Object.hasOwn`, not `in`: this trap answers about OWN properties, and
+      // `in` walks the prototype chain. Because the underlying value is an
+      // ordinary `Object.prototype`-rooted record, `in` reported every member of
+      // `Object.prototype` -- `toString`, `valueOf`, `constructor`, `__proto__`
+      // -- as an own property of the proxy, while `ownKeys` (which uses
+      // `Reflect.ownKeys`) listed none of them. Two traps describing the same
+      // value disagreed by construction, so every consumer that reasons about a
+      // read-back value's shape was being told it carries names it does not.
+      //
+      // That is not academic: `unsafeObjectKeyIn()` refuses a `FabricValue`
+      // carrying own `__proto__`/`constructor` and tests with `Object.hasOwn()`,
+      // so writing a read-back record back to a cell was rejected for a key the
+      // record never had, and every read-modify-write against a cell failed
+      // (loom CT-1949). The `has` trap below keeps `in` -- there it is correct,
+      // being the `in` operator's own trap.
+      if (
+        (isRecord(current) || Array.isArray(current)) &&
+        Object.hasOwn(current, prop)
+      ) {
         return {
           configurable: true,
           enumerable: true,
--- a/packages/runner/test/query-result-proxy-enumeration.test.ts
+++ b/packages/runner/test/query-result-proxy-enumeration.test.ts
@@ -300,4 +300,82 @@ describe("CT-1240: query result proxy enumeration", () => {
       popFrame(frame);
     }
   });
+  // A trap that answers about OWN properties must not consult the prototype
+  // chain. It used to use `in`, so every member of `Object.prototype` came back
+  // as an own property of the proxy while `ownKeys` listed none of them -- two
+  // traps describing the same value, disagreeing. Downstream that made a
+  // read-back record unwritable: `unsafeObjectKeyIn()` refuses a `FabricValue`
+  // with own `__proto__`/`constructor` and asks with `Object.hasOwn()`, so
+  // writing a record back to a cell was rejected over keys it never had
+  // (loom CT-1949).
+  it("getOwnPropertyDescriptor does not report inherited names as own", () => {
+    const cell = runtime.getCell<{ a: number }>(
+      space,
+      "test-own-descriptor-not-inherited",
+      undefined,
+      tx,
+    );
+    cell.set({ a: 1 });
+
+    const proxy = createQueryResultProxy<{ a: number }>(
+      runtime,
+      tx,
+      cell.getAsNormalizedFullLink(),
+      0,
+      true,
+    );
+
+    // The two names the FabricValue boundary reserves, plus enough of
+    // `Object.prototype` to show this is about inheritance generally and not a
+    // special case for those two.
+    for (
+      const inherited of [
+        "__proto__",
+        "constructor",
+        "toString",
+        "valueOf",
+        "hasOwnProperty",
+        "isPrototypeOf",
+        "propertyIsEnumerable",
+        "toLocaleString",
+      ]
+    ) {
+      expect(Object.hasOwn(proxy, inherited)).toBe(false);
+      expect(Object.getOwnPropertyDescriptor(proxy, inherited)).toBe(undefined);
+    }
+
+    // A name that is genuinely absent stays absent, and a real own key is
+    // still reported -- so the fix narrowed the answer without emptying it.
+    expect(Object.hasOwn(proxy, "definitelyNotAKeyAnywhere")).toBe(false);
+    expect(Object.hasOwn(proxy, "a")).toBe(true);
+
+    // The two traps now agree, which is the property that was violated.
+    for (const key of Reflect.ownKeys(proxy)) {
+      expect(Object.hasOwn(proxy, key as string)).toBe(true);
+    }
+  });
+
+  // `in` IS the `has` trap's own operator, so inherited names SHOULD answer
+  // true there. Pinned so a future cleanup does not "fix" both traps alike.
+  it("the `in` operator still sees inherited names", () => {
+    const cell = runtime.getCell<{ a: number }>(
+      space,
+      "test-in-operator-still-inherits",
+      undefined,
+      tx,
+    );
+    cell.set({ a: 1 });
+
+    const proxy = createQueryResultProxy<{ a: number }>(
+      runtime,
+      tx,
+      cell.getAsNormalizedFullLink(),
+      0,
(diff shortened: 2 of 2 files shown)


bfollington:
## Sequencing note — this is step 2 of 3

Recorded so the middle of this change isn't a place to get lost. Loom is currently running a local workaround, and that is deliberate, not a fork:

1. **[loom#4672](https://github.com/commontoolsinc/loom/pull/4672) — merging now.** Copies the read-back value before embedding it in a push payload, which restores Mobile Loom on the *current* vendored pin. Loom cannot wait for this PR because picking it up requires a vendor bump (tag + space-compat gate, potentially a Deno pin move), and every state-cell push has been failing since 2026-08-03.
2. **This PR** — the actual root cause. Fixes the misreport for every consumer rather than one caller.
3. **Vendor bump in Loom** to adopt it.

Loom's copy is **not** reverted at step 3. It is independently correct — the payload is hashed at one moment and serialized at another, with every field read going back through the runtime in between, so copying pins the value the daemon decided to push. That argument is separate from the reserved-name check and survives it. The comments over there are already written for the post-fix world, so no follow-up edit is expected on the Loom side.

## Measurement, if useful for review

Same read against a live toolshed, same script, two import maps — the only difference being this one-line change:

| | unpatched | patched (this PR) |
|---|---|---|
| `Object.hasOwn(v, '__proto__')` | `true` | `false` |
| `Object.hasOwn(v, 'constructor')` | `true` | `false` |
| `Object.hasOwn(v, 'toString')` | `true` | `false` |
| `unsafeObjectKeyIn(v)` | `__proto__` | `undefined` |
| `isFabricValue(v)` | `false` | `true` |
| `fabricFromNativeValue(v)` | **THROWS** | **ACCEPTED** |

`ownKeys` is identical in both runs and never contained any of those names — which is the point: the value was being refused over properties it does not have.

Ready for review whenever; still draft only because the layer choice is yours to confirm.

bfollington (commented):

Scope: `query-result-proxy.ts` plus its enumeration tests · deep-read of the changed traps and relevant callers/history · targeted proxy tests, semantic probes, full runner suite, fmt/lint, and CI logs checked.

No blocking findings.

`Object.hasOwn(current, prop)` is the right semantic for `getOwnPropertyDescriptor`. I found no caller that depends on inherited properties being synthesized as own descriptors. The relevant consumers either walk `Reflect.ownKeys()` and then inspect own descriptors, or deliberately use `Object.hasOwn()` to distinguish stored data from prototype behavior. Repository history reinforces that boundary: inherited `constructor` / `toString` are JavaScript behavior, not persisted child paths.

The `has` trap should retain `prop in current`. I probed ordinary records and sparse arrays across `Reflect.ownKeys`, `Object.hasOwn`, `Object.getOwnPropertyDescriptor`, and `in`: real own keys, array `length`, present/missing indices, holes, and inherited array/Object methods now agree with ordinary JavaScript semantics. The existing array length and index tests also pass.

🟡 Improvement — please revise the test-plan wording about the 120 failures. They are unrelated to this change, but “pre-existing timing-sensitive failures” is not the accurate diagnosis. The runner suite requires `--preload=test/clock-preload.ts`; both `packages/runner/deno.jsonc` and CI say that without it the global `clock` is undefined. My unsupported no-preload reproduction produced 180 leaf `ReferenceError: clock is not defined` failures plus cascading suite failures. The supported `cd packages/runner && deno task test` result is **1146 passed / 0 failed** (6011 steps), and all five CI Runner Test shards pass. Reporting that green supported run is clearer than normalizing the red unsupported invocation.

The current red Pattern Integration Tests (3/4) job also looks unrelated: `lunch-poll-vote.test.ts` timed out after five minutes because `#lp-join-name` never appeared. It does not exercise the changed trap directly, while the runner shards and runner integration package are green.

Possible follow-up, not introduced here: if a live proxy is created over an array and its backing cell is later replaced by an object, `"length" in proxy` throws because the `has` trap reports false for the array target's non-configurable `length`. The existing array-to-record `ownKeys` test exposes the setup; I verified the `has` failure on both semantics. Worth filing separately.

Verdict: the core fix is sound; merge once CI policy is satisfied.

bfollington:
Rebased onto current `main`.

The `Pattern Integration Tests (3/4)` failure on the previous head was not from this change — it was the known intermittent click-target flake, fixed on main in #5352 (*"the intermittent failure the lunch-poll two-browser vote test hit at its first join-card click"*), which landed after I branched off `e8d9abf76`. Notably `Package Integration Tests (runner)` — where this change actually lives — passed on that same run.

Re-ran locally after the rebase: `query-result-proxy-enumeration` (12 steps) and `native-conversion` (300 steps) both green. Ran the latter deliberately because #5343/#5347 changed `toJSON()` conversion routing on main in the same area.

bfollington:
Thanks — both corrections taken.

**The 120 failures were my invocation error, not the environment.** The suite needs `--preload=test/clock-preload.ts`, and I ran it without. I had presented a with/without baseline comparison as evidence of no regressions, but both sides of that comparison were broken runs, so it was worth very little. The supported invocation is `cd packages/runner && deno task test` → **1146 passed / 0 failed** (6012 steps), which I've now confirmed myself. Commit message and PR description corrected; I've also corrected it on CT-1949, where I'd repeated the same wrong characterisation.

**CI red herring:** independently reached the same conclusion — it's the intermittent click-target flake fixed on main in #5352 (`lunch-poll-vote.test.ts`, `#lp-join-name` never appearing), which landed after I branched. Rebased onto current main; CI is re-running.

**On the `length in proxy` follow-up:** agreed it's out of scope here and pre-existing on both semantics. Worth noting for whoever picks it up that it needs the `has` trap to consult the array target's non-configurable `length` the way `getOwnPropertyDescriptor` already does at its top — i.e. the same target-delegation this PR left untouched, not a second application of the `hasOwn` change. Happy to file it if you'd rather it not get lost.

Marking ready for review.

bfollington:
(Correction to my previous comment: I said "marking ready for review" and briefly did — that was mine to ask about rather than decide, so it's back to draft. It stays draft until you say otherwise. The review is clean and CI is re-running, so it's ready whenever you want to flip it.)

seefeldb (approved):

PR #5363, opened 2026-08-04
perf: stop re-walking values the write path already knows are mutable

Hixie:
Writing a value into a cell got slower over the last week, visibly so on the benchmarks dashboard. The proximate cause was the plain-object arm of the fabric type check, which grew from a single prototype comparison into a walk of every own key with a property descriptor fetched for each one. That walk buys something real -- it is what rejects symbol keys, non-enumerable string keys, and accessor-backed properties, none of which a fabric record can carry -- so the fix is not to give any of it back.

Counting the calls showed the walk was not the whole story. One `Cell.set()` of a twenty-level value, 41 objects and 82 keys in all, ran the plain-object check about five hundred times. Roughly nine in ten of those calls arrived by one route: `cloneHelper()` asks `canReturnAsIs()` whether it may pass a value through untouched, `canReturnAsIs()` asks `isDeepFrozenFabricValue()`, and that function evaluated `isFabricValue(value) && isDeepFrozen(value)`. `isFabricValue()` is an uncached recursive membership walk of the whole subtree, and `cloneHelper()` recurses, so the subtree was re-validated once per node per level -- quadratic in the depth of the value. Every one of those walks was discarded, because the value the write path is holding is mutable and so was never going to be deep-frozen at all.

So this swaps the two conjuncts. `isDeepFrozen()` answers in constant time for anything not frozen at its root, and memoizes each subtree it does walk; `isFabricValue()` starts from nothing every time. With the cheap, memoized question first, the membership walk runs only for a value that could still answer `true`. The conjunction's value is unchanged, and both operands are free of observable side effects, so no caller sees a difference.

The second change makes the check itself cheaper without weakening it. `Reflect.ownKeys()` answers "every own key, strings and symbols alike" in one call, but it has no engine fast path and costs about as much per call as the three single-purpose key reads cost in total for a small object. Asking the narrower questions separately -- `Object.keys()` for the enumerable string keys, `Object.getOwnPropertyNames()` whose length must match to rule out a non-enumerable one, and `Object.getOwnPropertySymbols()` which must be empty -- answers the same two questions for a fraction of the price. The per-key descriptor stays, because a descriptor is the only thing that distinguishes a data property from an accessor, and that distinction is the point of the check.

One behavioural detail improves as a side effect. Handed a proxy whose `ownKeys()` names a key that its `getOwnPropertyDescriptor()` then disowns, the check used to read a property off `undefined` and throw a `TypeError`; it now answers `false`, which is what its "answers rather than throws" contract says it should do. A test pins that.

Measured with the pinned Deno on an Apple M5 Max, alternating between worktrees so machine drift could not favour either side: `Cell.set() - depth: very deep (20 levels)` goes from 23959 to 16064 microseconds, and `Cell push - array append schemaless (100x)` from 47107 to
27212. Both land below where deleting the inertness checks outright would put them, because the quadratic re-validation predates those checks. The investigation record, including what was tried and did not pay off, is in docs/history/development/performance/.

---
## Summary by cubic
Speeds up write-path validation by checking frozen-ness first and using cheaper key reads for plain objects. Cell.set (20 levels) drops 23959→16064 µs; Cell push (100x) drops 47107→27212 µs.

- **Refactors**
  - `packages/data-model/src/deep-freeze.ts`: in `isDeepFrozenFabricValue`, evaluate `isDeepFrozen(value) && isFabricValue(value)` in that order to short-circuit on mutable values and avoid quadratic re-walks.
  - `packages/utils/src/objects.ts`: replace `Reflect.ownKeys()` with `Object.keys()`, `Object.getOwnPropertyNames()`, and `Object.getOwnPropertySymbols()`; keep per-key descriptors to confirm data properties.

- **Bug Fixes**
  - Proxies that list a key but return no descriptor now result in `false` instead of a throw; test added in `packages/utils/test/objects.test.ts`.

<sup>Written for commit a70e6f245b2a193aa61c110bcff9b8c297fe956b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5363?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/utils/src/objects.ts
+++ b/packages/utils/src/objects.ts
@@ -81,8 +81,8 @@ export function isInertPlainObject(
   // accessor carries `get` / `set` where a data property carries `value`. The
   // keys walked here are every own key, the two reads above having established
   // that there are no others.
-  for (let i = 0; i < keys.length; i++) {
-    const descriptor = Object.getOwnPropertyDescriptor(value, keys[i]!)!;
+  for (const key of keys) {
+    const descriptor = Object.getOwnPropertyDescriptor(value, key)!;
     if (!("value" in descriptor)) {
       return false;
     }
--- a/packages/utils/src/objects.ts
+++ b/packages/utils/src/objects.ts
@@ -60,17 +60,30 @@ export function isInertPlainObject(
     return false;
   }
 
-  // A single pass over the own keys checks all three requirements: each key
-  // must be a string (`Reflect.ownKeys()` yields symbol keys too, which a
-  // count-based comparison would need a second key walk to exclude), and its
-  // descriptor must be enumerable and carry `value` (an accessor answers
-  // `get` / `set` instead).
-  for (const key of Reflect.ownKeys(value)) {
-    if (typeof key !== "string") {
-      return false;
-    }
-    const descriptor = Object.getOwnPropertyDescriptor(value, key)!;
-    if (!descriptor.enumerable || !("value" in descriptor)) {
+  // Three key reads settle the two *key* requirements. `Object.keys()` is the
+  // enumerable string keys; `Object.getOwnPropertyNames()` is those plus the
+  // non-enumerable ones, so equal lengths mean every string key is enumerable;
+  // and no symbol keys means `Object.getOwnPropertySymbols()` is empty. This
+  // sits on the hot write path, and each of the three has an engine fast path
+  // that `Reflect.ownKeys()` -- which would answer both questions at once --
+  // does not: the three together cost a fraction of the one.
+  const keys = Object.keys(value);
+
+  if (Object.getOwnPropertyNames(value).length !== keys.length) {
+    return false;
+  }
+
+  if (Object.getOwnPropertySymbols(value).length !== 0) {
+    return false;
+  }
+
+  // The remaining requirement is data-ness, which only a descriptor answers: an
+  // accessor carries `get` / `set` where a data property carries `value`. The
+  // keys walked here are every own key, the two reads above having established
+  // that there are no others.
+  for (let i = 0; i < keys.length; i++) {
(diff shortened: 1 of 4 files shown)


packages/utils/src/objects.ts
```diff
@@ -60,17 +60,30 @@ export function isInertPlainObject(
...
+
+  if (Object.getOwnPropertySymbols(value).length !== 0) {
+    return false;
+  }
+
+  // The remaining requirement is data-ness, which only a descriptor answers: an
+  // accessor carries `get` / `set` where a data property carries `value`. The
+  // keys walked here are every own key, the two reads above having established
+  // that there are no others.
+  for (let i = 0; i < keys.length; i++) {
```
danfuzz: General recommendation is to write `for (const i of keys)` unless you actually need the index. (Doesn't look like you do.)

packages/utils/test/objects.test.ts
```diff
@@ -203,5 +203,20 @@ describe("objects", () => {
         isInertPlainObject(new Proxy(withSymbol, {})),
       ).toBe(false);
     });
+
+    it("answers for a `Proxy` that disowns a key it reported", () => {
```
danfuzz: Usual `it` language here would be `it("returns `false` for a...")`. (I know some projects/languages prefer "answers" but not this one.)

danfuzz (approved):
LGTM! Just a couple minor suggestions.

PR #5365, opened 2026-08-05
feat(api): fabric-primitive schema types — {type: "FabricBytes"} validates by prototype

ubik2:
## Summary

Schemas can now name a concrete `FabricPrimitive` class as a validation type: `"FabricBytes"`, `"FabricEpochDays"`, `"FabricEpochNsec"`, `"FabricHash"`, `"FabricRegExp"`. A value matches its type by prototype (`instanceof`), and each fabric-primitive type is a **subtype of `"object"`** the way `"integer"` is a subtype of `"number"` — existing `{type: "object"}` schemas keep accepting these values, while the new names reject plain records and don't cross-match each other.

By layer:

- **api** — `FABRIC_PRIMITIVE_SCHEMA_TYPES` (authoritative name list) + `isFabricPrimitiveSchemaType()`; `JSONSchemaTypes` extended; Schema→TS inference maps the names to the api-level interfaces.
- **data-model** — `schemaTypeOfFabricPrimitive()`: explicit `instanceof` mapping (not `constructor.name`, which minified bundles drop); throws on an unmapped subclass so a future primitive forces the vocabulary to be extended in the same change. `schemaForValueType()` infers the specific type, so `createJsonSchema` stops decomposing a primitive example into `{type: "object", properties: {}}`.
- **runner/traverse** — the `FabricPrimitive` leaf arm validates against the specific type name; `getPlainJsonType` reports it (flows into anyOf branch prefilters); subtype rule in `schemaTypeMatchesValueType`; `object`∧fabric-type schema intersections are non-disjoint and narrow to the fabric type. The memory/v2 server-side query path reuses this traverser, so one change covers both.
- **runner/cfc** — `SUPPORTED_SCHEMA_TYPES` and both value-vs-type matchers understand the names; `"object"` accepts primitives per the subtype rule.
- **piece** — schema evolution accepts fabric-type → `"object"` widening and flags the reverse narrowing.
- **docs/specs/json_schema.md** — the vocabulary documented under "Non-standard `type` values".

Values are already rehydrated class instances at traverse time (the serialized special-key form exists only inside the `fvj1:` JSON encoding at the storage boundary), so the prototype check is the right primitive.

**Deliberately out of scope:** the schema-generator still emits `true` for `Uint8Array`. Switching it to emit `{type: "FabricBytes"}` is a follow-up with golden churn in two packages and an upgrade procedure for deployed patterns (the stored-schema compat gate flags the `true` → `{type: "FabricBytes"}` narrowing; `cf piece setsrc --check` + `--dangerously-allow-incompatible-schema` after clone rehearsal is the path).

## Test plan

- New traverse cases: accept under specific type, cross-type reject (`FabricBytes` vs `FabricHash`), plain-record reject, anyOf branch selection.
- New data-model unit test pinning the class↔name mapping against the api vocabulary and codec-class list.
- New piece test for the one-way subtype rule in compatibility checking.
- Full runner suite (1146 tests), data-model, api, piece, and memory v2 schema-table suites all green; repo-wide `deno fmt --check` / `deno lint` clean for touched files.

---
## Summary by cubic
Add fabric-primitive schema types (e.g., { type: "FabricBytes" }) that validate by prototype. Each is a subtype of "object", so existing { type: "object" } schemas keep working while the new names enforce precise validation.

- **New Features**
  - API (`@commonfabric/api`): frozen `FABRIC_PRIMITIVE_SCHEMA_TYPES`, `isFabricPrimitiveSchemaType`; `JSONSchemaTypes` extended; Schema→TS maps to `FabricBytes`, `FabricEpochDays`, `FabricEpochNsec`, `FabricHash`, `FabricRegExp`.
  - Data-model (`@commonfabric/data-model`): `schemaTypeOfFabricPrimitive()` maps via `instanceof` and throws on unmapped subclasses; `schemaForValueType()` infers the specific type.
  - Runner (`@commonfabric/runner`): traverse validates against the specific fabric type; `getPlainJsonType` reports it; subtype rule lets `"object"` accept primitives and narrows intersections to the fabric type; CFC schema sanitization and wildcard policy matcher accept the names and match by prototype (unknown names stay conservative).
  - Piece (`@commonfabric/piece`): schema evolution accepts fabric-type → `"object"` widening and flags the reverse narrowing.
  - Docs: schema dialect documents the non-standard fabric-primitive `type` names and points to the vocabulary source.

- **Migration**
  - No action required. The generator still emits `true` for `Uint8Array`; switching to `{ type: "FabricBytes" }` is optional and can follow.

<sup>Written for commit f0c031e4dd1bc6b84f063e370f7997e482faf06b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5365?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

The coverage delta is 10 type-only lines in `packages/api/schema.ts` (the Schema→TS arms for the new type names plus their type imports). The file is a pure type-level module — it transpiles to an empty module, so V8 never records it and no test can execute a type declaration; every file in `packages/api` that *has* a coverage record is at zero uncovered lines. Accepting the metric at its current value per `docs/development/COVERAGE.md`:

ACCEPT_COVERAGE_DEBT: coverage-debt: packages/api uncovered lines = 1508 lines


ubik2:
This is just the traversal portion. The second bit is the schema-generation portion.
I opted for the `FabricBytes` style names over the `NATIVE_TYPES.Uint8Array` style names.

danfuzz:
Quick comment based on a read of the PR body: Though it's totally correct from a JS perspective to call `FabricPrimitive`s (`FabricBytes` (etc.)) subtypes of `object`, that might not be how we want to represent it in the schema semantics. The point of `FabricPrimitive` is to be as primitive-like as possible (e.g. among other things, ideally identity-free, just like `number` and `string`), and so calling them `object`s is kinda baking in a JS happenstance into the schema definition. Should we — say — grow a Rust implementation of the system, that version wouldn't have an intrinsic reason to think of them as `object`s.

I don't have a super-strong opinion about what to do in the short-term, but I figured I'd let you know my take.

packages/api/index.ts
```diff
@@ -1798,6 +1798,36 @@ export interface JSONObject extends Readonly<Record<string, JSONValue>> {}
...
 
+/**
+ * Fabric-primitive validation types -- a non-standard addition to the JSON
+ * Schema `type` vocabulary. Each name identifies a concrete `FabricPrimitive`
+ * class from the data-model, and a value matches by prototype (`instanceof`),
+ * not by structure. `"object"` also accepts these values -- every fabric
+ * primitive is a subtype of `"object"` the way an `"integer"` value satisfies
+ * a `"number"` schema -- so schemas that predate this vocabulary keep working.
+ */
+export const FABRIC_PRIMITIVE_SCHEMA_TYPES = [
```
danfuzz: Suggest `Object.freeze()` around the array to guarantee `const`ness.

packages/data-model/src/fabric-primitives/index.ts
```diff
@@ -31,3 +34,31 @@ const CODEC_CLASSES: readonly FabricClassWithCodec[] = Object.freeze([
   FabricEpochDays,
   FabricRegExp,
 ]);
+
+/**
+ * The JSON Schema `type` name for a `FabricPrimitive` instance, resolved by
```
danfuzz: Might want to put scare quotes around "JSON Schema" or use a circumlocution along the lines of "The `type` name in our schema model…"

packages/data-model/src/fabric-primitives/index.ts
```diff
@@ -31,3 +34,31 @@ const CODEC_CLASSES: readonly FabricClassWithCodec[] = Object.freeze([
...
+ *
+ * Uses "death before confusion": a `FabricPrimitive` subclass missing from
+ * this mapping throws rather than degrade to a broader type, so adding a new
+ * primitive class forces the vocabulary (here and in `@commonfabric/api`) to
+ * be extended in the same change.
+ */
+export function schemaTypeOfFabricPrimitive(
+  value: FabricPrimitive,
+): FabricPrimitiveSchemaType {
+  if (value instanceof FabricBytes) return "FabricBytes";
```
danfuzz, line 55: I'd prefer to leave this general, along the lines of just `return Object.getPrototypeOf(value).constructor.name`. One less table to manually manage. Maybe the client of this function can be the site that `throw`s on an unrecognized type?

If you revise the decision to use the full class name as the schema type name, we can of course re-schmancify this.

packages/data-model/test/schema-utils.test.ts
```diff
@@ -818,6 +825,12 @@ describe("schema-utils", () => {
     testType("null", null);
     testType("array", [1, 2, 3]);
     testType("object", { a: 1 });
+    // Fabric primitives get their specific type name, not "object".
```
danfuzz: Super-nit: I try to stick with the MD-code-quoted type name `FabricPrimitive` in comments instead of "fabric primitive" so as to be unambiguous about what's being talked about.

danfuzz (approved):
LGTM, with a handful of suggestions. Thanks for doing this!

ubik2:
> Quick comment based on a read of the PR body: Though it's totally correct from a JS perspective to call `FabricPrimitive`s (`FabricBytes` (etc.)) subtypes of `object`, that might not be how we want to represent it in the schema semantics. The point of `FabricPrimitive` is to be as primitive-like as possible (e.g. among other things, ideally identity-free, just like `number` and `string`), and so calling them `object`s is kinda baking in a JS happenstance into the schema definition. Should we — say — grow a Rust implementation of the system, that version wouldn't have an intrinsic reason to think of them as `object`s.
> 
> I don't have a super-strong opinion about what to do in the short-term, but I figured I'd let you know my take.

I want to still make them subtypes of object for now. We currently have a bunch of schemas that should accept these, and those will all be written `{type: "object"}`, so I want all those to continue to work.

Once I've done the other half (the transformer should generate the proper types), we may be able to phase out letting "object" match, but I'm not sure. I think this will need a bit more investigation, since TypeScript will allow them to have one variable that's of type `object`, and assign a "FabricEpochNsec" to it. We're going to turn those into cells with schemas, and the cell with the schema that accepts type object wants to still be able to accept the value in the cell that has type FabricEpochNsec.

I'm not aware of any obvious way to prevent that from happening in TypeScript, though I suppose our pattern compiler could reject those with a custom warning (though it's unfortunate for a user that has what looks like valid code to their other tools).

Edit: I haven't integrated your other comments yet, but I'll apply those in a bit.

packages/data-model/src/fabric-primitives/index.ts
ubik2: My agent complains about this because we have a minify for the shell that will mangle constructor names.
1. We can probably do the `keepNames` option on the minify, so that's one option.
2. Another option, if we just want these more tightly coupled, is to embed the string name with the types.
3. I can create this table at runtime, mapping a constructor.name to a fixed string name, though we'd still need to maintain a list of constructors.

Any preference here? I don't really like (2), but the others both seem ok to me.
If I go with (1), I'll probably land as-is, then do `keepNames` in another PR, then come back to this and switch to constructor.name.

packages/api/index.ts
ubik2: This is only exported for tests, but we don't really have a great way of doing that without also making it visible to user code.
I switched to Object.freeze.

packages/data-model/src/fabric-primitives/index.ts
danfuzz: A reasonable complaint! With that context, I'm a-okay maintaining the names explicitly. My suggestion is that we document in the code _why_ we are doing it, since the failure case isn't obvious.

packages/runner/src/traverse.ts
```diff
@@ -2506,16 +2513,20 @@ function schemaTypeMatchesValueType(
...
   // Integer is a subtype of number: an integer value satisfies either schema,
-  // while a fractional number only satisfies a number schema.
+  // while a fractional number only satisfies a number schema. Likewise, each
+  // fabric-primitive type is a subtype of "object": a `FabricBytes` value
+  // satisfies both `{type: "FabricBytes"}` and `{type: "object"}`, while a
+  // plain record only satisfies the latter.
   return schemaType === valueType ||
-    (schemaType === "number" && valueType === "integer");
+    (schemaType === "number" && valueType === "integer") ||
+    (schemaType === "object" && isFabricPrimitiveSchemaType(valueType));
```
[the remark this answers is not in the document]
ubik2: This case is ok. No FabricPrimitive should have required fields, so they should fail to match. Only the more lenient `{type: "object"}` style should match.

packages/data-model/src/fabric-primitives/index.ts
ubik2: We register these names with the codec as well, so I may be able to instead get constructors from that mapping or build a table dynamically. I went ahead and merged this as-is, but will probably follow up -- perhaps when I look at the generator side.

PR #5356, opened 2026-08-04
feat: server-execution v2 stage C.1 — certificate emission + consumers + goldens

seefeldb:
## Why

Stage C of server-execution v2 Phase 1 is the MAIN REDUCTION: main still
carries the v1-era completeness-certificate surface
(`completeSchedulerScopeSummary` / `completeActionScopeSummary` — two
identifiers, one certificate), which spec §5 deletes and
[serving-loop.md §3b](https://github.com/commontoolsinc/labs/blob/main/docs/specs/server-side-execution/serving-loop.md)
sizes at ~25 source files across five packages plus ~110 golden fixtures. The
plan cuts the stage as a three-PR train; this is **C.1 — emission + consumers
+ goldens**: delete the certificate at its transformer source and every runner
consumer, and regenerate the goldens in the same PR. Deleting at the source
collapses the rest (spec §4's measured lesson).

Predecessor: stage B, #5349 (the execution lease and the derived-class
admission check). C.2 (observation tables → the v2 basis index + protocol
deletions) and C.3 (flag retirement + doc archival) follow on top of this
branch.

## What

**ts-transformers** — the two emission sites are gone:

- `schema-injection.ts`: the no-input computed form emits `lift(cb, false)()`
  — no result-schema pad, no options object.
- `lift-applied-strategy.ts`: the scheduler-options object now exists only for
  `materializerWriteInputPaths`; the completeness analysis
  (`hasCompleteSchedulerScopeSummary`) is deleted.
- All **109 golden fixtures** regenerated via the sanctioned
  `UPDATE_GOLDENS=1` procedure (the diff is uniformly
  `lift(cb, false, undefined, { completeSchedulerScopeSummary: true })` →
  `lift(cb, false)`).
- `ts_transformers_current_behavior_spec.md` §8/§9.5 updated in the same
  change (the package rule: transformer changes ARE spec changes).

**runner** — every consumer deleted:

- `builder/types.ts` / `builder/module.ts`: the module marker and its
  `DeriveSchedulerOptions` field are gone (old compiled patterns carrying the
  option keep loading — unknown option fields were never read anywhere else).
- `runner.ts`: the certificate build at action wire-up is deleted. The static
  redirect-write collection it shared survives on its live path only
  (`collectStaticRedirectWriteTargets`, `complete`-flag bookkeeping dropped);
  the read-target collector had no other consumer and is deleted.
- `scheduler/types.ts`, `scheduler/run.ts`,
[...]

PR #5367, opened 2026-08-05
feat: server-execution v2 stage C.2 — observation tables become the basis index

seefeldb:
## Why

Stage C of server-execution v2 Phase 1 is the MAIN REDUCTION. C.1 (#5356)
deleted the completeness-certificate surface; this is **C.2 — protocol +
engine + client + tools**: replace the `persistentSchedulerState` observation
tables with the v2 **basis index** and carry the protocol-layer deletions and
the collateral no byte-identical gate covers
([serving-loop.md §3b](https://github.com/commontoolsinc/labs/blob/main/docs/specs/server-side-execution/serving-loop.md)).

Predecessor: C.1, #5356. C.3 (flag-entry retirement + doc archival) follows.

## The migration (engine)

- `scheduler_basis` is created by `INIT` on every open, with §3b's DDL
  verbatim (columns, PK, comment block; NOT NULLs and the entity-keyed
  lookup index are the sanctioned implementation detail). It starts EMPTY
  and nothing writes it yet — the key shape is instance-shaped now, feeding
  it instance VALUES is stage E's job.
- The migration drops §3b's **SEVEN** tables — driven off the spec's list,
  deliberately not the old `CORE_SCHEDULER_TABLES` constant, which
  enumerated six and omitted `scheduler_context_floor` (D6, the named
  foreseeable mistake). Satellites drop before the `scheduler_observation`
  spine they FK into.
- **NO BACKFILL** (D10): `process_generation`-keyed history cannot be
  reinterpreted as overwrite-in-place basis state. An opted-in store loses
  warm start once, by design.
- A new engine test pins all three: post-open the only `scheduler_%` table
  is `scheduler_basis`, a store seeded with all seven old tables (floor
  included) comes out clean, and the basis table is empty afterwards.

## The protocol deletions (D7)

Derived, not re-derived: with the persisted form gone the flag gates
nothing, so the following went WITH the machinery, on both sides —

- the `persistentSchedulerState` flag and its ambient control point
  (`setPersistentSchedulerStateConfig`), env var, runtime option, and every
  plumbing site (shell defines, lib-shell, background-piece-service,
  runtime-client, the integration harness, the `patterns-reload` CI lane's
  env);
- the `serverFlags` hello negotiation (both sides);
- the `scheduler.snapshot.list` RPC, its parse/validation, and the client
  wrapper;
- the commit-carried observation payload (`schedulerObservation` /
  `schedulerObservationBatch` on `ClientCommit`, the batch flush machinery,
  the transaction hooks, `SessionSync.observations` and the sync-attach).

Old-client compat is the existing hello-degrade path: a migrated server
advertises nothing to negotiate, so an old client treats the state as
absent and runs fresh — the same path it takes today against a
non-advertising server (D11). No version handshake added.

The client machinery those fed went too: boot-listing rehydration, the
snapshot-candidate selection, incremental observation adoption
(`adoptRemoteObservations`), the rehydration barrier, and the
`resumeMode: "always-run"` exclusion that existed for them.
[runtime-mapping.md N62's](https://github.com/commontoolsinc/labs/blob/main/docs/specs/server-side-execution/runtime-mapping.md)
original keep-OFF-arm disposition is superseded in place (edited in this
PR): keeping the receive path would have kept dead code no server could
ever feed. What SURVIVES is the live, non-flag machinery: the
`observationIdentity` tag (plan-B shaped-wake grouping), the
`awaitSyncBeforeInitialRun` resume hold, and the CT-1927 marker staging
(its two gating-seam tests died with the seam they monkey-patched — the
interleaving window they guarded is structurally gone now that the
transact handler has no post-apply await).

## Collateral (D9/S7)

- `packages/state-inspector`: the five-table requirements map and its
  readers are gone; the summary surface now reports the `scheduler_basis`
  table and row count, and `cf inspect summary` renders that.
- `packages/memory/v2/sqlite/guard.ts` `CORE_TABLE_NAMES`: dropped names
  OUT, `scheduler_basis` IN.
- Docs edited in the same PR where the change made them stale:
  serving-loop §8's scope note, spec README §5's "live on MAIN" paragraph,
  runtime-mapping N62, the `EXPERIMENTAL_OPTIONS.md` entry (marked
  DELETED/EXECUTED; the entry moves to Appendix A in C.3), and the
  invalidation.ts push-shaping comment (its adoption reason is gone; the
  N10 re-check stays owed).

## Test instruments that outlived their subject

Three live-subject test groups had used the deleted machinery as an
*instrument* and were re-instrumented rather than deleted:

- The CT-1927 parked-accept tests probed the pending overlay through the
  deleted `schedulerHasPendingWriteOverlapping` oracle; the replica now
  exposes a plainly-named `hasPendingWrite(id, scope?)` and the tests read
  that.
- `scheduler-split-clock` used the deleted rehydration barrier to keep
  `assessPullWork` clock-silent; its fixture now supplies a valid empty
  work state and one extra stubbed clock reading, preserving the
  single-read discrimination it pins.
- `src-garble-identity-invariant` read action identities from persisted
  observation rows; it now reads them from the live scheduler graph
  snapshot. Restart-stable action identity stays load-bearing for v2 (the
  basis index's `action` column), so the guard survives.

The OFF arm is byte-identical: the flag was OFF by default, so none of the
deleted machinery ran in the default posture. Plan checkbox C.2 ticked.

---
## Summary by cubic
Replaced the v2 scheduler observation tables with the new `scheduler_basis` index and removed the `persistentSchedulerState` protocol/engine/client/tooling; default behavior is unchanged since the flag was off.

- **Migration**
  - Engine: on open, drop the seven legacy `scheduler_%` tables (including `scheduler_context_floor`) and create `scheduler_basis` empty; no backfill, so a one-time loss of warm start.
  - Protocol/Client: removed serverFlags hello negotiation, `scheduler.snapshot.list`, and commit-carried observation payloads; old clients auto-degrade via the existing hello path, no version handshake change.
  - Config/CI: removed `EXPERIMENTAL_PERSISTENT_SCHEDULER_STATE` and related wiring; default posture stays the same.

- **Collateral**
  - Tools/Docs: state inspector and `cf inspect` now report `scheduler_basis` presence/row count; docs updated to reflect Stage C.2 and the basis index.
  - Runner cleanup: deleted incremental adoption, the rehydration barrier, and resumeMode hooks; tests re-instrumented to read the live scheduler graph.
  - Guardrails: `CORE_TABLE_NAMES` includes `scheduler_basis`; removed observation-era tests/benches.

<sup>Written for commit f48e57efa8badc622f87dd48f72912957f4baf3f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5367?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seefeldb:
## Adversarial review (subagent panel + adjudication) — VERDICT: LANDABLE-WITH-FIXES (fixes pushed to the C.3 branch)

The migration core is verified hard: the `scheduler_basis` DDL matches serving-loop §3b's authored schema exactly; the SEVEN-table drop is hardcoded (not driven off the six-table constant, which is deleted entirely) and the FK drop ORDER was independently re-derived against the real FK graph (satellites → spine → floor) — correct under `PRAGMA foreign_keys=ON`; no backfill (asserted by test: pre-existing observation data → 0 basis rows); D7's protocol deletions are complete on both sides with the observation lifecycle traced end-to-end to its one deleted source; all three D9/S7 collateral items done, with the guard blocklist covered by a real schema-enumeration test; re-instrumented tests verified non-vacuous (including the clock-read count derivation and the CT-1927 race-window closure).

Fixes (landing on the c3 branch, since the train merges together): `docs/development/CONFIGURATION.md` still lists the deleted flag as live in three tables, and `docs/specs/memory-v2/04-protocol.md` ("currently shipped wire behavior") still documents the deleted hello capability + `HelloMessage` field. Trivia also fixed: two dead empty imports; one test renamed to match its post-PR contract.

seefeldb (commented):
Verdict: no new actionable code finding in the Stage C.2 layer at this exact head.

The seven-table removal, empty basis-index migration, old-client hello degradation, state-inspector implementation, and fail-closed deletion paths are consistent with the plan. The current check rollup is not green (Test 2/6 and 4/6 plus Status fail), so this layer is not ready to land until those exact-head failures are cleared.

seefeldb:

PR #5368, opened 2026-08-05
feat(toolshed): self-serve ingest channels — mint your own without an operator

jkomoros:
## Why

Minting an ingest channel is a user action — "let my own device write into my
own space" — but today it is an operator action. The only path is
`deno task provision-ingest-channel`, run on the deployed host with the
toolshed's private identity, so every device onboarding is a ticket to whoever
holds the Ansible vault password. That does not scale past a handful of people
and it blocks the iOS location beacon's first-run experience entirely.

There is a quieter problem underneath, and it is arguably worth more than the
endpoint. `MEMORY_ACL_MODE` now defaults to `enforce`, and ingest writes into
the *user's* space under the *operator's* identity. If the operator holds no
authority there, POSTs return **200 while nothing commits** —
indistinguishable from a typo'd space DID. A channel can be born dead and look
healthy.

Builds on the landed `journal` sink (#4446) and the vouched-ingest mint seam
(#4392), and closes the "create-authorization gap" those left open.

## The hard part

Creation was deliberately left out because an unauthed create taking a
caller-supplied target space is a **confused deputy**: it gets
legitimately-minted `ExternalIngest` marks written into someone else's space.
Two facts make it harder than it looks:

1. **First-party HTTP auth authenticates but does not authorize.** It proves
   control of a `did:key` and nothing more; `first-party-http-auth.ts` has
   carried `TODO(auth)` since it landed, and `verifiedUserDid` was read in four
   places, all logging.
2. **A space DID is not a credential.** Naming a space proves nothing about
   controlling it.

## What Changed

**The authority already exists** — the space's own ACL document — so this adds
no trust root, no proof format, and no ceremony. Authenticate with the request
proof that already shipped, then require an explicit `OWNER` grant for that DID
on the target space.

One ACL read answers two questions that need **opposite** resolutions:

- *"Is the caller entitled?"* — narrow: `acl[caller] === "OWNER"`. Deliberately
  **not** `spaceReaderRole`, which resolves `acl[principal] ?? acl["*"]` and so
  returns "owner" for everyone on `{alice:"OWNER","*":"OWNER"}`. Permissive is
  right for the render fit it was built for and wrong for minting write
  authority.
- *"Can the operator deliver?"* — permissive, via that same oracle, because
  here it is a *prediction* of what the memory server will do. This is what
  makes mint fail loudly instead of ingest failing silently.

Surface: `POST /api/ingest-channels/{mint,list,rotate,revoke}` — its own prefix
(the data plane carries `cors({origin:"*"})`, and `/api/ingest/channels` would
collide with `/api/ingest/:id`), all verbs POST because the in-runtime signer
is a POST-only path allowlist, `bodyLimit` and rate limiting mounted *before*
auth. Plus `cf ingest mint|ls|rotate|revoke`, registration lifecycle fields
(`owner`, `expiresAt` enforced, `revoked`), a per-owner index, and toolshed's
first rate limiter.

### Rejected, with reasons recorded in the doc

- **Nonce write-back proof-of-control** proves *write* capability — but named
  spaces genesis with `"*": "WRITE"`, so the ceremony authorizes exactly who it
  exists to exclude.
- **Registration in the user's own space** is sound for *home* spaces (no
  wildcard at genesis) and unsound for the named data spaces the beacon
  targets: there is no per-document write attribution, so an attacker writes
  their own registration with their own secret hash.

## Self-serve status

🔴 **The control plane ships OFF.** `INGEST_SELF_SERVE_ENABLED` defaults false
and the router answers **404** on every verb before anything else runs. This is
deliberate and is the required fix from the first security review: until named
spaces stop deriving their key from a public passphrase, anyone who knows a
space *name* can grant themselves OWNER and mint legitimately — and minting
turns a temporary takeover into a credential that outlives the fix. Enabling it
is gated on that repair plus the cutover below, not on this PR.

With the flag on, it works with **no per-user admin step**: a user runs
`cf id new`, touches their space (which genesis-writes the ACL with their DID as
OWNER), and mints.

⚠️ **Also required before enabling:** add the toolshed operator DID to
`MEMORY_SERVICE_DIDS`, or minting for a *home* space returns an actionable 409
(named spaces still work via the genesis wildcard). One-time config, not per
user.

⚠️ **`RATE_LIMIT_TRUST_FORWARDED_FOR`** — set only when a trusted proxy is
actually in front. The limiter keys on the **rightmost** `X-Forwarded-For`
entry, because a proxy *appends* what it saw and everything to the left of that
is client-authored. Correct for exactly one appending proxy; a chain of N needs
a trusted-hop count. Off, it keys on the socket peer.

Known limit: a passkey-only shell login cannot drive the CLI at all — that is
pre-existing and applies equally to `cf acl`, `cf piece`, `cf wish`.

## Retiring channels when trust conditions change

A minted token outlives the conditions that authorized it. Until space keys
stop deriving from a public passphrase, anyone who knows a space *name* can
grant themselves OWNER and mint legitimately; fixing that retracts nothing
already issued.

A trust-epoch field was rejected (permanent runtime surface for a one-time
event). Revocation already is fail-closed, loud, and non-destructive, so
retirement is `deno task retire-ingest-channels`, with `audit-ingest-channels`
to make a cutover answerable — finally giving the global registration index a
production caller. All self-serve credentials are also finite-lived (90-day
default, 365 max), so the sweep is a backstop rather than the only bound.

**Step 0 of the cutover is `audit-ingest-channels --recover <file>`**, and it is
easy to miss. Indexing is now in the same transaction as the registration, but
before that it was best-effort, so a channel can exist that is live, appends
into a user's space, and is in *no* index — invisible to its space's owner and
to the sweep alike. Repairing an index by enumerating that same index only ever
confirms what is already there. A channel id is `channelId(space, installId)` —
derived, not random — so the operator's own provisioning record is the source of
truth: `<space-did> <install-id>` per line, and each derived id either resolves
or does not. Needed once, only on deployments that provisioned before this
change; nothing minted from here can join that population.

**And the reminder is structural, not documentary.** A tripwire test asserts
the weakness is *still present*, so fixing space-key derivation breaks the
build and the failure message is the procedure. `deno task check-tripwires`
re-derives the same condition independently and also fails if the test is
deleted, gutted, or skipped — verified by breaking each path.

## Lifecycle hardening (two security-review rounds)

The endpoint above is the easy half. Most of the diff is making the *lifecycle*
safe to enable, and these are the parts worth reviewing closely:

- **Revoke is bound to both a request id and a generation.** The id
  (`requestId`, consumed in the same transaction as the write) stops a captured
  revoke that was *delivered* and replayed. It does nothing for one captured and
  **withheld** — an id that was never spent looks fresh — so revoke also
  requires `expectedRevision`, the generation the caller read from `list`,
  enforced as the write precondition. A withheld request names a generation that
  no longer exists.
- **Durable state is bounded on four axes**: request claims compacted per owner,
  a live-channel cap enforced *inside* the owner-index transaction (a count
  taken before it is advisory — concurrent mints all pass it), and lifetime
  quotas per owner **and per space**. Per-owner alone bounds a keypair, not a
  person: DIDs are free, so one person grants OWNER to fresh keys and spends a
  new allowance from each. The global audit index is sharded by UTC month, so
  no single array grows forever.
- **A record containing a link sigil is rejected.** Records are opaque JSON, but
  opaque is not inert — the runtime decides on *write* whether a value is a
  link, so `{"/": {"link@1": …}}` was stored as a live reference, not text. That
  broke append-only (a link's target changes afterwards, so a historical record
  mutates with no ingest touching it) and made the `ExternalIngest` mark attest
  to a digest of the sigil rather than to anything a reader resolves through it.
  Checked with the runtime's own `isLink`, so it cannot drift from what the
  writer interprets.
- **Operator scripts conflict visibly** rather than silently undoing a
  concurrent revoke: `provision-ingest-channel` passes the observed revision and
  fails loudly, and `--repair-indexes` propagates everything that is not a
  revision conflict.
- **`revoke` has its own rate-limit bucket.** These limiters run ahead of auth,
  so draining the shared bucket also refused revokes — and "come back later" is
  the wrong answer to "kill this credential."

## Test plan

- toolshed **135 passed / 381 steps**, cli **173 passed**, workspace
  `deno task check` clean, `deno lint` clean, `check-docs` **538 blocks**,
  `check-tripwires` intact. Rebased onto main and re-verified against the new
  base — a clean textual rebase is exactly what hides semantic drift.
- New harness: `packages/toolshed/lib/test-support/memory-acl.ts` runs a **real
  memory-v2 server with `acl: { mode: "enforce" }`**. The existing suite uses
  `StorageManager.emulate`, whose server is built with no `acl` option at all —
  so nothing was authorized and any authorization test written against it
  proved nothing.
- Acceptance, against a **concrete, non-derived** space owner: mint for your
  own space; refused for a space you don't control; list; rotate; revoke stops
  POSTs. Plus: wildcard-WRITE and wildcard-OWNER strangers refused; unhosted
  and not-owned denials byte-identical; rotate/revoke refused from a non-owner
  with the victim's token still working; replayed `requestId` returns 409 and
  no token; `secretHash` never in any response.
- Replay regressions through the signed HTTP route: a revoke captured and
  replayed after a re-mint, and one captured, *withheld*, and delivered late —
  the second is the case the request id alone does not cover.

## Review notes

Nine adversarial review rounds found ~37 defects that are fixed here. The ones
worth knowing about while reading:

- **The revoke path has been through four revisions and three of them had a real
  defect.** It is the part of this diff I trust least. The failures were: no
  idempotency at all; an id that a *withheld* request never spends; a read-only
  already-revoked branch that could report success while a re-minted credential
  was live; and pruning the space index on revoke, which left a revoked channel
  in no list, so its generation was unknowable and every later revoke 409'd
  telling the caller to list something that would never appear.

- The **401 equalization is relaxed in exactly one place**: a *cryptographically
  correct* token for a revoked/rotated/expired channel gets a distinguishable
  403 "re-pair this device". Unreachable without proof-of-possession; every
  case a guesser can reach keeps the equalized 401 and the dummy-hash compare.
  Exactly two hash compares on every path.
- `space-key-derivation-tripwire.test.ts` **asserts a weakness holds**. That is
  intentional and the file says so in its first line — it is not a test to
  "fix".
- Re-embedding values read back from a cell does not round-trip (they are
  deep-frozen); `plainRevocations` exists for that and cost a real bug once.
- The control-plane test harness reads `revision` straight out of storage, a
  capability no HTTP client has. That is convenient but it is also how the
  unlistable-revoked-channel bug stayed green: the tests exercised a path real
  callers could not reach. Worth remembering when reading them.

---
## Summary by cubic
Enables self-serve ingest channels so users can mint, list, rotate, and revoke their own channels with first‑party HTTP proofs and a space‑ACL OWNER check. Adds an opt‑in gate, per‑space listing, stricter rotate ownership, a `cf ingest` CLI, and now makes all credentials finite‑lived with optimistic, atomic lifecycle writes.

- **New Features**
  - Control plane: `POST /api/ingest-channels/{mint,list,rotate,revoke}` with body limit and client‑address rate limiting; first‑party request proofs; requires explicit OWNER on the target space; idempotent mint/rotate via `requestId`; per‑owner and per‑space indexes; list by owner or `--space <did>`; rotate honors takeover (co‑owners cannot rotate others’ channels); index writes are mandatory and precede the registration; registrations carry a monotonic `revision` and lifecycle writes (registration + all indexes) are atomic via `editWithRetry` with a precondition — stale snapshots return 409; request‑id claims are compacted per owner (30‑minute retention, cap 500).
  - CLI: `cf ingest mint|ls|rotate|revoke` (signs with the user’s identity; supports space DID or name; prints single‑use tokens; `ls --space <did>` inventories all channels on a space).
  - Data plane: 403 “re‑pair this device” for valid tokens on revoked/rotated/expired channels (uses `previousSecretHash`); enforces `expiresAt`; registrations track `owner`, `revocations`, `previousSecretHash`, and `revision`; first mint without `ttlDays` defaults to 90 days and re‑mint preserves the existing window; live‑channel cap.
  - Ops: `deno task audit-ingest-channels` and `deno task retire-ingest-channels`; entrypoints expose `main(args, makeRuntime, serviceSpace, log, logError)` for testability; tripwire and task to enforce the space‑key derivation obligation; rate limiter with injectable clock.
  - Hardening & Tests: positive auth path; storage‑faults surface as 502; router/middleware guards; limiter, script, and validation coverage; defensive branches covered (unreadable stores treated as “not hosted”; rate‑limit `clientKey` handles no connection info).

- **Migration**
  - Set `INGEST_SELF_SERVE_ENABLED=true` to enable the control plane (default is OFF).
  - Add the toolshed operator DID to `MEMORY_SERVICE_DIDS` so minting works for home spaces; otherwise minting returns 409.
  - Set `RATE_LIMIT_TRUST_FORWARDED_FOR` only when behind a trusted proxy that overwrites `X-Forwarded-For`.
  - Channels minted via the old operator script lack `owner` and won’t appear in `cf ingest ls` by owner; use `cf ingest ls --space <did>`, re‑mint, or retire as needed.
  - Clients should handle 403 “re‑pair” after rotation/revocation/expiry.
  - All channels are finite‑lived: a first mint without `ttlDays` gets a 90‑day default; re‑mint keeps the prior window.

<sup>Written for commit 4e58f4b3200dbe9c821e310821bd67255bba9114. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5368?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Coverage

The coverage ratchet flagged 514 uncovered lines when this opened. It is now
22, and the tests written to get there are not padding — the four tripwire
evasion paths are automated instead of hand-verified, the positive first-party
auth path is covered for the first time, storage faults are pinned to 502 on
every control-plane verb, and the operator scripts' dry-run/confirm behaviour
is tested.

The remaining 22 are accepted below rather than chased, because each needs
machinery that would test the harness instead of the code:

- `ingest-channels.handlers.ts` (4) — the `!callerDid` guards, which the auth
  middleware makes unreachable, and the `status === 200` branches, which need a
  live runtime. Under unit test the `runtime` singleton is uninitialized, so
  these answer 502 by design. Integration surface, not unit surface.
- `audit-/retire-ingest-channels.ts` (5) and `check-tripwires.ts` (1) — the
  `import.meta.main` lines. Both command bodies were extracted into exported,
  tested functions precisely so only the entrypoint remains.
- `space-authority.ts`, `ingest-channels.utils.ts`, `rate-limit.ts` (12) —
  defensive branches whose reachable cases are already tested.

`packages/cli` is accepted at +2 rather than chased, and the reason is that I
could not find the two lines. Measured locally under the full CLI suite, every
source file this branch adds is fully covered — `commands/ingest.ts` 201/201,
`lib/ingest-channels.ts` 78/78 — and `commands/main.ts` sits at 114/119 whose
five uncovered lines are pre-existing FUSE lazy-imports, not the two this branch
adds. The gate's own reporter says the same: "could not tie the regression to
specific files from the diff." Rebasing onto current main did not move it, so it
is not baseline drift either. Two lines that measure as covered everywhere I can
measure them is not worth further hunting; it is recorded here rather than
waved through silently.

The registration assertion added while investigating is worth keeping on its own
merits: nothing checked that `cf ingest` is wired into the main command, and a
dropped registration makes the whole verb vanish with no error anywhere.

ACCEPT_COVERAGE_DEBT: coverage-debt: packages/toolshed uncovered lines = 4532 lines
ACCEPT_COVERAGE_DEBT: coverage-debt: tasks uncovered lines = 1331 lines
ACCEPT_COVERAGE_DEBT: coverage-debt: packages/cli uncovered lines = 3411 lines


jkomoros:
@cubic review

willkelly (changes requested):
_cf-review via Hephaestus (OpenAI `gpt-5.6-sol`), on behalf of @willkelly._

Security review of `4106ca5f6` · deep-read of the authorization, lifecycle, storage, replay, transport, and operator-recovery paths · six independent review tracks · targeted suites, type checks, LSP, tripwires, and an isolated adversarial lifecycle test run.

The local authentication and authorization checks are thoughtful: request proofs bind the expected inputs, wildcard ACL grants are rejected, rotate/revoke authorize against the registration's stored space, secrets are omitted from list responses, and body limiting runs before signature verification. I am requesting changes because the end-to-end capability lifecycle is not safe for production self-service yet.

## 🔴 Blocking

### 1. Critical — named-space ownership is forgeable, and mint turns it into persistent authority

**Scenario**

1. Bob creates named space A, for example `family-map`. Its key is derived from the public constant `"common user"` plus that name.
2. Alice learns the name `family-map` and independently reconstructs space A's private key.
3. Alice signs as space A and adds an explicit `OWNER` grant to Alice. The new endpoint now considers Alice legitimately authorized.
4. Alice mints a channel without `ttlDays` and receives a bearer token with no expiry.
5. The named-space derivation is fixed and Bob removes Alice's ACL grant. Alice's token remains an active operator-backed append capability until somebody discovers and revokes it.

The current derivation is in [`packages/identity/src/session.ts:21-34`](https://github.com/commontoolsinc/labs/blob/4106ca5f6b1d8da4f81e7ac7792668465857c29d/packages/identity/src/session.ts#L21-L34), while expiry is optional in [`ingest-channels.routes.ts:100-118`](https://github.com/commontoolsinc/labs/blob/4106ca5f6b1d8da4f81e7ac7792668465857c29d/packages/toolshed/routes/ingest-channels/ingest-channels.routes.ts#L100-L118). The tripwire correctly proves the weakness, but a tripwire does not make issuing durable credentials safe. This PR amplifies a temporary takeover into persistence beyond remediation.

**Required fix:** keep the production control-plane routes disabled until named-space key custody is repaired. Before enabling them, sweep ACLs and retire any channels issued under the old trust condition.

### 2. Major — removing an owner neither revokes nor exposes that owner's channels

**Scenario**

1. Bob creates space A and temporarily grants Alice `OWNER` so they can collaborate.
2. Alice mints a channel for space A using a random `installId` and omits `ttlDays`.
3. The collaboration ends and Bob removes Alice from space A's ACL.
4. Alice's bearer token continues appending indefinitely because the data plane checks registration state and optional expiry, but never checks the current ACL.
5. Alice can still list the channel because list authorization is only the authenticated Alice DID plus Alice's owner index. Bob, despite being the current owner of space A, sees zero of Alice's channels and therefore does not learn the ID needed to revoke it.

I reproduced this exact sequence in an isolated test against the PR's ACL-enforcing memory harness: after Bob removed Alice's grant, ingest still returned 200, Alice still listed one channel, and Bob listed none. The relevant paths are [`ingest-channels.utils.ts:508-537`](https://github.com/commontoolsinc/labs/blob/4106ca5f6b1d8da4f81e7ac7792668465857c29d/packages/toolshed/routes/ingest-channels/ingest-channels.utils.ts#L508-L537) and [`ingest.utils.ts:560-595`](https://github.com/commontoolsinc/labs/blob/4106ca5f6b1d8da4f81e7ac7792668465857c29d/packages/toolshed/routes/ingest/ingest.utils.ts#L560-L595). The feature document's statement that ACL-removal TOCTOU is bounded by `expiresAt` is false when expiry is optional.

**Required fix:** make self-service credentials finite-lived and give every current space owner a complete space-scoped inventory/revocation path, or bind data-plane acceptance to the current ACL/grant generation.

### 3. Major — the emergency audit/retirement procedure can silently miss live channels

**Scenario**

1. Bob creates space A and successfully mints a channel.
2. `saveRegistration` writes Bob's owner index and the live registration, then the global-index append fails.
3. That exception is swallowed as best-effort, so Bob receives a working token.
4. Later, Admin fixes the named-space weakness and runs `audit-ingest-channels` followed by `retire-ingest-channels --confirm` as prescribed.
5. Both commands enumerate only the global index, so Bob's live channel is absent from the audit and survives retirement even though Admin is told the sweep completed.

The live registration precedes the swallowed global-index write in [`ingest.utils.ts:387-429`](https://github.com/commontoolsinc/labs/blob/4106ca5f6b1d8da4f81e7ac7792668465857c29d/packages/toolshed/routes/ingest/ingest.utils.ts#L387-L429). Audit and retirement consume only that index in [`audit-ingest-channels.ts:42-68`](https://github.com/commontoolsinc/labs/blob/4106ca5f6b1d8da4f81e7ac7792668465857c29d/packages/toolshed/scripts/audit-ingest-channels.ts#L42-L68) and [`retire-ingest-channels.ts:61-74`](https://github.com/commontoolsinc/labs/blob/4106ca5f6b1d8da4f81e7ac7792668465857c29d/packages/toolshed/scripts/retire-ingest-channels.ts#L61-L74). That index is not merely telemetry once it is the only recovery inventory.

**Required fix:** make global indexing mandatory before activation, preferably atomically with the registration and owner-index writes. A failed audit-index write must not leave a live token.

### 4. Major — rotate bypasses the explicit takeover protocol, and lifecycle writes race

**Scenario**

1. Bob creates space A, grants Alice co-owner access, and mints channel `phone-1` for Bob's device.
2. Alice learns the deterministic channel ID and calls `rotate` directly.
3. `loadOwned` admits Alice because she currently owns space A; unlike mint, rotate never checks that the registration is owned by Bob.
4. `persist` replaces the secret and changes the registration owner to Alice without first recording the explicit revocation required by the mint takeover path.
5. Bob's device immediately stops working, Bob's channel disappears from Bob's list, and the audit trail has no revoke-then-transfer event explaining the takeover.

The mint path deliberately requires revoke-before-takeover at [`ingest-channels.utils.ts:323-354`](https://github.com/commontoolsinc/labs/blob/4106ca5f6b1d8da4f81e7ac7792668465857c29d/packages/toolshed/routes/ingest-channels/ingest-channels.utils.ts#L323-L354), but rotate lacks that guard at [`:451-476`](https://github.com/commontoolsinc/labs/blob/4106ca5f6b1d8da4f81e7ac7792668465857c29d/packages/toolshed/routes/ingest-channels/ingest-channels.utils.ts#L451-L476), and `persist` always assigns the caller as owner at [`:234-255`](https://github.com/commontoolsinc/labs/blob/4106ca5f6b1d8da4f81e7ac7792668465857c29d/packages/toolshed/routes/ingest-channels/ingest-channels.utils.ts#L234-L255).

There is a related concurrency failure: Bob's rotate and Alice's revoke can both read the same registration snapshot, then blind-write independently. Depending on write order, revoke can return 200 while the channel is active again, or rotate can return 200 with an already-dead token. Concurrent first mints can similarly bypass the takeover and live-channel-cap prechecks.

**Required fix:** apply the foreign-owner guard to rotate and transact the current-registration read, generation/version precondition, request claim, index updates, and registration mutation as one optimistic transaction.

### 5. Major — a tenant can grow durable bookkeeping without bound

**Scenario**

1. Bob creates space A and grants Alice `OWNER` for a legitimate integration.
2. Alice repeatedly rotates one channel using a fresh `requestId`. Every rotation creates a permanent request-claim cell and does not consume the 200-live-channel allowance.
3. Alternatively, Alice repeatedly mints and revokes channels. Revocation frees the live allowance, while registration cells and Alice's append-only owner index continue growing.
4. Alice stays within the nominal live-channel policy while making future list and mint operations scan an ever-growing history with repeated storage synchronization barriers.
5. Admin sees shared service-space storage and CPU grow indefinitely. If toolshed runs multiple instances, Alice also receives one independent rate budget per process; sufficient key churn triggers the LRU behavior that grants fresh buckets after eviction.

Request claims are one cell per fresh ID in [`ingest.utils.ts:287-350`](https://github.com/commontoolsinc/labs/blob/4106ca5f6b1d8da4f81e7ac7792668465857c29d/packages/toolshed/routes/ingest/ingest.utils.ts#L287-L350). The live cap scans historical owner-index entries in [`ingest-channels.utils.ts:369-400`](https://github.com/commontoolsinc/labs/blob/4106ca5f6b1d8da4f81e7ac7792668465857c29d/packages/toolshed/routes/ingest-channels/ingest-channels.utils.ts#L369-L400). The limiter explicitly keeps [separate process-local buckets](https://github.com/commontoolsinc/labs/blob/4106ca5f6b1d8da4f81e7ac7792668465857c29d/packages/toolshed/lib/rate-limit.ts#L1-L4) and [refreshes evicted keys](https://github.com/commontoolsinc/labs/blob/4106ca5f6b1d8da4f81e7ac7792668465857c29d/packages/toolshed/lib/rate-limit.ts#L30-L59).

**Required fix:** expire or compact request claims within the proof replay window, compact/bound owner history, and enforce any deployment-wide abuse budget at trusted ingress rather than relying on a process-local LRU.

## Verification

The existing focused suites passed: 116 test steps across control-plane authorization, ingest, route/auth/CORS, rate limiting, and operator scripts; the named-space tripwire passed and confirmed the weakness; sensitive files passed `deno check` and LSP diagnostics. The custom ACL-removal scenario above also reproduced successfully. These failures are missing lifecycle invariants, not failures in the existing happy-path tests.

**Verdict: request changes. The endpoint should not be enabled as production self-service until the trust-root, revocability/inventory, audit completeness, transactional lifecycle, and durable abuse bounds are resolved.**

jkomoros:
Thanks — this found real defects, and two of them contradicted invariants this PR claims for itself. Pushed `1b749357d`; CI green (49 pass).

## Fixed

**1. Control plane is now opt-in, default OFF** (`INGEST_SELF_SERVE_ENABLED`). Your framing is the one that moved me: a tripwire tells you when the repair lands, but only a gate stops durable credentials being issued before it, and the repair does not retract what was issued. The gate mounts ahead of every other middleware, so a disabled deployment 404s before the body limit, rate limiter, or signature verification runs. The doc now records the enable order: repair custody → sweep ACLs → `retire-ingest-channels` → enable.

**2. A space's current owner can find and revoke foreign channels.** Your reproduction was exact. `list` consulted only the caller's own index, which contradicted this design's own stated principle — *the resource is the space, not the minting key* — that rotate and revoke already followed. There is now a per-space index; `list --space <space>` authorizes the caller as a current OWNER and returns every channel targeting it, whoever minted it. Bob can enumerate and revoke Alice's channel, and the token then 403s. A test walks your five steps.

You were also right that the doc's "bounded by `expiresAt`" was false while `ttlDays` is optional. Corrected, not softened.

**3. Global index is mandatory, before the registration.** It was best-effort from when nothing read it; it is now the inventory a cutover sweep enumerates, so a swallowed failure meant a live channel reported as retired. A sweep that lies is worse than a mint that fails.

**4a. Rotate carries mint's foreign-owner guard.** Confirmed: `loadOwned` proves only space ownership, so a co-owner could rotate someone else's channel — secret replaced, `owner` reassigned, incumbent's device dead, no revocation record. Takeover stays a deliberate revoke-then-mint.

Also: the rate limiter's clock is now injectable. Its refill tests slept, which made one flaky (it failed once during this work) and violated the repo's waiting guidance.

## Not fixed — deliberately, and I'd like your read

**4b. Transactional lifecycle.** Your race is real: concurrent rotate/revoke read the same snapshot and blind-write. The fix — reading registration, request claim, and index updates under one optimistic transaction with a version precondition — is a genuine redesign of `persist`/`saveRegistration`, which the operator provisioning script also calls. I did not want to attempt that at the end of a long change to security-sensitive code; that is how the defects you found got in. With the gate off, the exposure is bounded to deployments that have opted in.

**5. Durable bookkeeping.** Also real. Request claims are one permanent cell per fresh id; the owner index is append-only; buckets are process-local. Each has a different answer (claims want expiry the cell model does not offer, the index wants compaction, the limiter wants trusted ingress), and I would rather scope them than bolt on three partial mitigations.

**2, second half.** You asked for finite-lived credentials *and* inventory, or ACL-generation binding. I did inventory. Mandatory TTL was explicitly considered and rejected earlier (permanent re-pair churn for a one-time threat) — but that predates your review, so it is worth re-deciding rather than me asserting the old answer. Binding the data plane to the current ACL is the third option and is cheap on the existing per-POST best-effort path; it also means a user narrowing their own ACL stops their own beacon, which is arguably correct.

My preference: land this gated off, then take 4b, 5, and the 2-second-half as a focused follow-up rather than growing a 36-file PR. If you would rather they be in this one, say so and I will.

jkomoros:
Cross-linking, because blocking finding #1 has a prerequisite that now has a ticket.

Step 3 of that chain — Alice signs *as space A* and grants herself `OWNER` — works because `Server.#resolveCapability` (`packages/memory/v2/server.ts:1040`) short-circuits to OWNER when `principal === space`, **before** consulting the ACL. That short-circuit exists so a missing or malformed ACL can be initialized or repaired, and it is documented as intentional in `04-protocol.md:733` — but it never ends, so it keeps applying long after genesis wrote a real owner.

[CT-1969](https://linear.app/common-tools/issue/CT-1969/drop-implicit-space-did-owner-once-a-spaces-acl-has-a-concrete-owner) removes it once the ACL has a concrete owner, keeping it only for the repair case. Berni asked for it as a follow-up to #5259 and scoped it out of that PR.

That does not close finding #1 on its own — an attacker who re-derives the key still holds it — but it removes the escalation step the scenario depends on: without the short-circuit, signing as the space no longer grants authority over a space that already has an owner.

Verified against a real in-process server (`acl: { mode: "enforce" }`): after a normal genesis, a session as the space key rewrote the ACL to name itself sole owner, after which the original user was denied READ. So finding #1's step 3 reproduces today.

Finding #2 looks independent of custody to me and worth fixing on its own terms — a current owner genuinely cannot enumerate a former owner's channels, so there is no revocation path even when the ACL is correct.

jkomoros:
@willkelly — pushed four commits addressing all five findings. You reviewed `4106ca5f6`; the branch is now at `5e1454081`.

**#1** — your required fix as written. The control plane mounts behind `INGEST_SELF_SERVE_ENABLED`, default off, and the gate runs *first*, ahead of the body limit, the rate limiter and signature verification, so a disabled deployment 404s as if the routes never existed (`ingest-channels.index.ts`, `env.ts`). That flag did not exist at the commit you reviewed.

**#2** — credentials are finite-lived unconditionally: an omitted `ttlDays` takes `DEFAULT_TTL_DAYS = 90`, and a lapsed expiry is deliberately not inherited (`ingest-channels.utils.ts:225-247`). And there's now a per-space registration index, so `list {space}` returns every channel targeting a space you own regardless of who minted it (`getSpaceRegistrationIndex`, `:660`) — a current owner can both see and revoke a former owner's channels.

**#3, #4, #5** — mandatory global indexing before activation, the foreign-owner guard applied to rotate plus an optimistic transaction across the lifecycle writes, and bounded bookkeeping. I'd rather you re-read those than take my summary for them.

Ready for re-review.

One question for the enablement checklist, not for merging: does CT-1969 count toward "named-space key custody is repaired," or do you mean the derivation itself?

CT-1969 drops the implicit `principal === space` OWNER in `#resolveCapability` (`packages/memory/v2/server.ts:991`) on any space whose ACL already names a concrete owner, keeping it only to initialize or repair a missing one. That removes the middle step of your #1 chain: a forged key would fall through to `acl[principal] ?? acl["*"]` — the genesis wildcard's `WRITE` — and ACL-document writes require `OWNER` (`server.ts:2004`), so the key could no longer grant itself the entry mint checks for. It does not remove the key, and an un-genesised name looks like it can still be claimed (`server.ts:1160`), so I'd understand if your line is the derivation itself. It changes what I write in the checklist, not whether this merges.

Two artifacts I'm adding either way, so the checklist is enforced rather than remembered:

1. An `INGEST_SELF_SERVE_ENABLED` entry in `docs/development/EXPERIMENTAL_OPTIONS.md` — the registry that already carries your `conflictAdmissionMode`, and whose format requires a stated end state and a concrete removal path. The removal path is the enablement checklist, recorded where the periodic flag sweeps actually look.
2. A second tripwire. The one on this branch fires when key *derivation* is fixed. Nothing currently fires when the implicit-owner short-circuit is fixed — so if CT-1969 ends up in the gate, that's an obligation with no instrument behind it.

willkelly (changes requested):
_cf-review via OpenAI `gpt-5.6-sol`, on behalf of @willkelly_

Security re-review at `273629f5402a09fda1709e98caba2b00978cb655` (the rebase changed none of the ingest lifecycle code from `5e1454081`). The remediation is substantial: the control plane is genuinely default-off at the production router; all self-serve credentials now expire; a current space owner can enumerate and revoke a former owner's channels; and self-serve registration, request-claim, and index updates are revisioned and transactional. I reproduced the Bob/Alice ACL-removal flow and confirmed the revoked token returns 403.

I am still requesting changes because the feature is not safe to enable yet.

## 🔴 Blocking

### 1. A captured revoke can revoke a later, re-minted credential

**Location:** `packages/toolshed/routes/ingest-channels/ingest-channels.routes.ts:186-194`, `ingest-channels.utils.ts:574-625`

**Scenario:** Bob creates space A, mints a channel for his phone, and sends a correctly signed `{ id }` revoke request. Alice captures that exact request (for example, from a compromised client or trusted-proxy log) and holds it. Bob revokes the channel, then re-mints the same `(space, installId)` and pairs the new token. During the first-party proof's freshness window, Alice replays Bob's old signed revoke. `processRevoke()` loads the new registration and revokes it, so the newly paired phone immediately stops working.

The revision precondition does not stop this: the replay performs a fresh read and targets whatever revision is current. Mint and rotate consume a `requestId` transactionally for this reason, but revoke has neither a request id nor an intended revision in its signed body.

**Fix:** bind revoke to a caller-supplied idempotency key consumed in the registration transaction, or require the intended registration revision/generation in the signed request. Add a revoke → re-mint → replay test through the signed HTTP route.

### 2. One authenticated owner can grow deployment-wide audit state without bound

**Location:** `packages/toolshed/routes/ingest/ingest.utils.ts:273-288,605-624`; `packages/toolshed/routes/ingest-channels/ingest-channels.utils.ts:434-470`

**Scenario:** Bob creates space A and repeatedly mints a fresh `installId`, then revokes it. Revocation removes the id from Bob's live owner/space indexes, so it frees the 200-live-channel allowance, but every unique registration cell and every entry in the single global `cf:ingest:index` remain forever. Bob can repeat this indefinitely. Admin eventually has a deployment-wide index whose full array is read and rewritten by lifecycle transactions, while audit and retirement commands must walk the entire history.

The process-local/LRU rate limiter slows this down but does not bound it, and multiple instances multiply the budget. This is durable shared-service resource exhaustion by an otherwise legitimate owner.

**Fix:** impose a lifetime retained-registration quota and an explicit bounded retention/archive policy (preferably without one ever-growing array) before enablement. Also move the live-channel admission check into the same transaction as the owner-index update; concurrent distinct mints can currently all pass the pre-transaction count near the cap.

### 3. Break-glass provisioning can silently erase a concurrent revoke

**Location:** `packages/toolshed/scripts/provision-ingest-channel.ts:96-148`; `packages/toolshed/routes/ingest/ingest.utils.ts:492-501`

**Scenario:** Bob creates space A and grants Alice `OWNER`. Alice sees a suspicious channel and revokes it. At the same time, Admin's break-glass provisioner has already read the old enabled registration and is rotating it. The script omits `expectedRevision`, so its stale unconditional save lands after Alice's revoke, clears `revoked`, replaces the secret, and makes the channel live again even though Alice's revoke reported success.

I reproduced the exact stale-read → revisioned revoke → unconditional operator save sequence against the current runtime. Admin is trusted, but a trusted action should conflict visibly rather than silently undo a concurrent security action.

**Fix:** pass `null` for first creation or the observed revision for reprovisioning, fail visibly on `RegistrationConflictError`, and test the provision/revoke race.

### 4. Legacy inventory repair can report success while live channels remain undiscoverable

**Location:** `packages/toolshed/scripts/audit-ingest-channels.ts:57-70`

**Scenario:** Bob creates space A before this change and receives an operator-provisioned channel. Its old best-effort global-index update was lost, or it predates the per-space index. Alice later becomes space A's current owner. Admin runs `audit-ingest-channels --repair-indexes` before enabling self-service. A registration missing from the global index is never visited at all; and for registrations that are visited, any storage/commit failure is swallowed as though it were only a lifecycle conflict. The command exits successfully, but Alice still cannot discover or revoke the live channel.

**Fix:** catch only `RegistrationConflictError` and propagate other failures. Before enablement, define a source of truth that can discover registrations omitted from the old global index (rather than trying to repair an index by enumerating that same index).

## 🟡 Coherence / rollout follow-ups

- `docs/features/self-serve-ingest-channels.md:511-516` says the per-owner/per-space indexes accumulate history, while the implementation removes revoked entries and retains history only in the global audit inventory.
- `INGEST_SELF_SERVE_ENABLED` is still absent from `docs/development/EXPERIMENTAL_OPTIONS.md`, the repository's required central flag registry.

## Verification

- Current head pinned and clean; GitHub reports it mergeable.
- Focused Toolshed suite: **135 steps passed**.
- Adversarial runtime tests: removed-owner inventory/revoke passed; stale revoke replay and unconditional operator overwrite both reproduced.
- Production app with `INGEST_SELF_SERVE_ENABLED=false`: mint/list/rotate/revoke all returned 404.
- `deno check` + `deno lint` on seven sensitive files; LSP diagnostics clean; `deno task check-tripwires` passed.
- GitHub CI was green for checks, builds, unit suites, runner suites, and package integration suites; four pattern-integration shards were still pending when this review was written.

**Verdict:** changes requested. The default-off gate safely keeps the current deployment dormant, but the goal is safe self-service; the enabled lifecycle still permits replayed revocation, unbounded durable abuse, silent revoke overwrite, and incomplete legacy inventory.

jkomoros:
Two corrections to what I said I'd add, one of which cancels the other.

**The `EXPERIMENTAL_OPTIONS.md` entry is already in.** It landed in `6594546fd`, after the commit you pinned — summary-table row plus a full Category 6 section with the toggle, the default, the end state, and the removal path (fix derivation → retirement sweep → flip default → delete the gate). Your note was accurate at `273629f5`. I've just pushed `288c93753` fixing the one real defect in it: the section was ordered above Category 5.

**The second tripwire is off the table, and Berni is the reason.** I'd said I would add one watching the implicit space-DID OWNER short-circuit, on the theory that CT-1969 could land independently and nothing would notice. He's since ruled that out on #5259:

> this code is old, i don't remember. but let's just not change whatever arises from the space principal. if it can overwrite everything, that's fine for now. we need to redo that principal away from the name derivation anyway and that's when we'll tackle it. OUT OF SCOPE FOR THIS PR.

So the short-circuit is not being fixed separately — it gets absorbed into replacing the name-derived principal. The existing tripwire already fires on exactly that event, so a second one would be an instrument for something that cannot happen on its own. Adding it would be worse than not: two tripwires for one condition invites someone to silence the noisy one.

That also answers the question I asked you: CT-1969 is not a distinct enablement line, because it isn't landing separately. The precondition is the one you wrote — repair key custody, sweep, then enable — and the existing derivation tripwire is the instrument behind it. I've marked CT-1969 accordingly rather than leaving it in the backlog looking actionable.

On your four new findings: #3 (`expectedRevision` in the provisioner) and #4 (catch only `RegistrationConflictError`, and don't repair an index by enumerating itself) look small and clearly right. #1 needs a choice between a consumed idempotency key and an intended-revision in the signed body — I'd take the revision, since revoke already reads the registration and it avoids a second consumed-token store, but say if you'd rather have the key. #2 is the one I'd treat as real work rather than a patch: it's a retention policy, and "bounded" needs a number and an archive story before it's a fix.

jkomoros:
@willkelly — all four blocking items from `273629f5` are addressed, plus both coherence follow-ups. Requesting another round.

Two things worth reading before the diff, because they change what you told me to do.

**On #1, your "or" is not an or.** I did the idempotency key first, exactly as written: `requestId` consumed in the same transaction as the registration write. Then an adversarial pass showed it does not close the attack. A request id only proves *this exact request was already delivered*. A request captured and **withheld** was never delivered, so its id is unspent and looks perfectly fresh — it stays live for the rest of the proof window and lands on whatever credential exists when it is finally let through. The realistic sequence is not exotic: swallow Bob's revoke, let him conclude it failed and retry (fresh id), let him re-pair, then deliver the original.

So `expectedRevision` is load-bearing, not belt-and-braces, and both are in: the id kills the delivered-and-replayed case, the generation kills the withheld case. Required in the signed body, enforced as the write precondition. Tests for both sequences, plus the signed-route test you asked for and one asserting a revoke that omits either field is 422.

**On #2, a per-owner quota bounds a keypair, not a person.** A DID is a free keypair, so one person with one space grants `OWNER` to as many fresh keys as they like and spends a new allowance from each — and every new DID mints its own permanent `by-owner`, `lifetime` and `requests` cells, so the mechanism meant to bound growth becomes an unbounded cell family in the same dimension. There is now a per-**space** lifetime quota alongside the per-owner one; a space is what a caller cannot mint for free within this feature's reach. A refused mint writes nothing, so it bounds the per-key cell family too. Unlimited *space* creation remains an axis — that is deployment admission control, not this feature's.

### The four

1. **Captured revoke** — `requestId` consumed in the registration transaction *and* required `expectedRevision`. Also fixed two things this created: a read-only already-revoked fast path that could report success while a re-minted credential was live (caught by cubic), and — the one I would look hardest at — pruning the space index on revoke made a revoked channel appear in *no* list, so its generation was unknowable, so every later revoke 409'd telling the caller to list something that would never appear. The space index now retains revoked ids (nothing reads its length; the per-space quota bounds it); the owner index stays live-only because its length *is* the live count the cap reads.

2. **Unbounded audit state** — per-owner *and* per-space lifetime quotas, both monotone counters rather than lists. The global index is sharded by UTC month with a shard directory, so there is no ever-growing array, per your parenthetical. The live-channel admission check moved into the same transaction as the owner-index update; since that index is live-only its length is the live count, which fixed the race and removed the O(N) scan at once. The owner meter charges *acquisition*, not creation — a takeover adds a channel to an identity without creating one.

3. **Break-glass provisioning** — passes `null` on first creation, the observed revision on reprovision, exits 1 with a plain message on `RegistrationConflictError`. Race tested.

4. **Legacy inventory** — `--repair-indexes` now catches only `RegistrationConflictError` and propagates everything else. For the second half: repairing an index by enumerating that same index only ever confirms what is already there, so `--recover <file>` takes the operator's provisioning record as `<space-did> <install-id>` per line and probes the derived ids directly — a channel id is `channelId(space, installId)`, so this is exact rather than a scan. It is step 0 of the cutover and is needed once, only on deployments that provisioned before this change; indexing is now in the same transaction as the registration, so nothing minted from here can join that population.

**Coherence:** the doc's index claim is corrected (twice now, as the behavior changed under it), and `INGEST_SELF_SERVE_ENABLED` is in `EXPERIMENTAL_OPTIONS.md`.

### Found while answering you, not by you

Four adversarial rounds since your review turned up 13 defects, most in the code written to answer these items. Two you should know about specifically:

- **Link sigils in ingest records** (data plane, not control plane). Records are opaque JSON, but opaque is not inert — the runtime decides on *write* whether a value is a link, so a POST carrying `{"/": {"link@1": …}}` stored a live reference, not text. It is a non-null non-array object, which was the entire record contract. That broke append-only (a link's target changes afterwards, so a historical record mutates with no ingest touching it), made the `ExternalIngest` mark attest to a digest of the sigil rather than to anything a reader resolves through it, and let a record address a document the channel was never authorized to write. Now rejected with the runtime's own `isLink`, so the check cannot drift from what the writer interprets.
- **The rate limiter was a no-op behind an appending proxy.** `clientKey` took the leftmost `X-Forwarded-For` entry; a proxy appends, so everything left of it is client-authored and a caller drew a fresh full bucket per request. Rightmost now. Revoke also got its own bucket — these limiters run ahead of auth, so draining the mint bucket refused revokes, and "come back later" is the wrong answer to "kill this credential."

Worth calibrating your read on: **the revoke path has been through four revisions and three of them had a real defect.** It is the part of this diff I trust least and the part I would most like you to attack.

### Still deliberately not done

The gate stays off. Your first review's #1 — named-space key custody — remains the enablement precondition, unchanged and not something this PR can fix. The tripwire fails loudly when the derivation changes, which is the signal to run the cutover. Berni confirmed on #5259 that the derivation is slated to be redone ("we need to redo that principal away from the name derivation anyway and that's when we'll tackle it"), so the precondition has an owner and a plan rather than being a wish.

### The PR description is stale — read this instead

I have not rewritten it yet, so two things in it are actively wrong and I do not
want you tripping over them:

- It still describes the deploy gate as "add the operator DID to
  `MEMORY_SERVICE_DIDS`" and predates `INGEST_SELF_SERVE_ENABLED` entirely.
  The flag is in and the routes are 404 by default — your first review's
  Critical #1 required fix. Only cubic's auto-summary reflects that.
- Its migration note says to set `RATE_LIMIT_TRUST_FORWARDED_FOR` when behind a
  proxy that *overwrites* `X-Forwarded-For`. As of this round the key is the
  rightmost entry, which assumes the proxy *appends*. Following that note
  configures the limiter wrong.

Also absent from the body: `--recover`, which is **step 0 of the cutover
procedure** and needed once on any deployment that provisioned before this
change. The test-plan numbers are stale too (toolshed is 135 / 381 steps, cli
173, check-docs 538). I will rewrite the body; flagging it now so nothing here
is contradicted by it in the meantime.

### Verification

Rebased onto main (82 commits, no conflicts) and re-verified all prior fixes against the new base — a clean textual rebase is exactly what hides semantic drift. Toolshed 135 / 381 steps, CLI 173, workspace `deno task check`, `deno lint`, 538 doc blocks, tripwire intact, CI green on `787bd045b`.

If you are satisfied on the four, I would like an approval on this round — and if you would rather split the data-plane sigil fix into its own PR, say so and I will.

jkomoros:
Body rewritten — it was stale enough to contradict the comment above, so the two corrections I flagged there are now fixed at the source: the gate is the headline of **Self-serve status** (routes 404 by default), and the `RATE_LIMIT_TRUST_FORWARDED_FOR` note says *rightmost / appending proxy* rather than the old *overwrites*, which would have configured the limiter wrong.

Also added: `--recover` as step 0 of the cutover, a **Lifecycle hardening** section covering what the two security rounds actually changed, and current test numbers. The disclaimer in my previous comment no longer applies.

willkelly (changes requested):
_cf-review via Hephaestus `openai/gpt-5.6-sol`, on behalf of @willkelly._

Re-reviewed at `787bd045b92876f880af0b6a6ae59a2d82854a35`. The four blockers from the previous human review are resolved: revoke is generation-bound, durable growth is bounded for one owner and for key churn on one space, break-glass provisioning is optimistic, and legacy inventory recovery now surfaces failures.

## 🔴 Blocking

### 1. `cf ingest` drops configured API base paths

`packages/cli/lib/ingest-channels.ts:17,83` combines the root-absolute `BASE = "/api/ingest-channels"` with `new URL(..., config.apiUrl)`. For a deployment configured as `https://host.example/fabric`, every command sends its request to `https://host.example/api/ingest-channels/...`, silently discarding `/fabric` and addressing the wrong endpoint.

This is verified against the current head. Please preserve `config.apiUrl.pathname` when constructing the control-plane URL and add a client test with a pathful base URL.

### 2. A backward clock jump can keep the revoke limiter closed

`packages/toolshed/lib/rate-limit.ts:49-58` computes refill from wall-clock elapsed time and persists the result even when elapsed time is negative. After a backward clock adjustment, a spent bucket acquires negative tokens and remains denied beyond the configured refill interval. I reproduced a bucket moving from zero to `-10`, then remaining denied after one normal refill second.

This reaches the dedicated revoke bucket in `packages/toolshed/routes/ingest-channels/ingest-channels.index.ts:38-53`, where an extended false denial leaves a credential live. Please use a monotonic default clock or clamp negative elapsed time, and add a backward-jump test.

## 🟡 Improvements

- `packages/toolshed/routes/ingest-channels/ingest-channels.routes.ts` should document the actual `413` and `422` responses and both list modes; the current success description says only “channels this caller minted,” while `space` returns every channel targeting the owned space.
- `docs/plans/ingest-channels-journal-sink.md:28,95-100` still describes unauthenticated creation and the webhook create/list/delete lifecycle, contradicting current code and the same document's resolved decision at line 129.
- `packages/toolshed/scripts/provision-ingest-channel.ts:116-177` now carries important expiry, revocation-history, and optimistic-conflict behavior without a direct provisioning-script test.

## Verification

- Toolshed: 159 focused steps passed.
- CLI: 35 focused steps passed.
- `check-docs`: 538 code blocks passed.
- `check-tripwires`, formatting, lint, focused type checks, and LSP diagnostics passed.
- Current GitHub CI is green.

**Verdict: changes requested.**

jkomoros:
@willkelly — both blockers from `787bd045b` are fixed, along with all three improvements. Requesting approval.

**1. Base path.** Fixed, and it was worse than the report describes. The proof commits to the URL, so on a path-prefixed deployment the client was signing one path and sending to another — those requests could not have verified even against a correct endpoint. So it was not "addressing the wrong endpoint," it was a client structurally broken on that topology. Now joins onto the base's own pathname; tests cover a prefix, a trailing slash, a bare origin, a nested prefix, non-mutation of the caller's URL, and that the URL signed is the URL sent.

**2. Backward clock.** Fixed, and the obvious fix was half of one. Clamping elapsed time stops a spent bucket going below empty, but `updatedAt` was still stored as the rewound instant — so when the clock caught back up, the whole excursion counted as elapsed and refunded a full bucket. The jump went from a stall to a bypass, in the direction that matters more. The stored timestamp is now monotonic as well, so an excursion costs nothing and earns nothing, and the default clock is `performance.now()` so the class does not recur on the default path.

Worth flagging: the test I first wrote alongside the clamp asserted a token was buyable one second after the jump. Under the corrected semantics the bucket waits for real time to pass the high-water mark — so that assertion had been pinning the bug's behaviour. Both directions are now one test.

**Improvements.** 413 and 422 are in the contract (both come from middleware, so no handler returns them and they were easy to omit — 413 runs before auth, 422 only after, and a client has to tell them apart). The list route documents both modes. `docs/plans/ingest-channels-journal-sink.md` no longer describes creation as unauthed or the lifecycle as webhooks' create/list/delete, and a link that pointed at a path the doc does not have is fixed.

`provision-ingest-channel.ts` had no exports and so no test. Its logic is now `provisionChannel(runtime, serviceSpace, request)`, which borrows the runtime rather than owning it — `main` constructs and disposes one, and a function that disposes its caller's runtime cannot be tested against stored state, which is why the siblings are shaped this way. Seven tests: input rejection, mandatory expiry, lifecycle carry-forward, the `--force` repoint guard, a preserved revocation, the optimistic-conflict branch, and that bad input is rejected without a runtime ever being constructed.

**Also fixed since your review**, from cubic and from the coverage gate:

- I had written a test named "fails visibly instead of overwriting a concurrent revoke" that asserted *success* — `provisionChannel` re-reads on every call, so the conflict branch was never entered. That is the exact failure mode I have been flagging elsewhere in this PR, in my own test. It now races two provisions of one channel so exactly one must lose, and asserts the loser's code, message, and that only one write landed.
- Validation ran after the runtime was built, so a malformed `MEMORY_URL` turned a usage error into an uncaught construction failure. Extracted and moved ahead of `makeRuntime()`.
- The coverage ratchet caught the `already revoked` message branch in `cf ingest revoke`. Covered rather than waived — the branch exists so "it worked" and "it was already dead" are distinguishable, which is a claim that deserves a test.

**Verification:** toolshed 135 / 391 steps, CLI 173, workspace `deno task check`, `deno lint`, 538 doc blocks, `check-tripwires`, `deno fmt --check`. CI green.

One note on the CI history if you look at it: a `CLI Integration Tests` shard showed cancelled on `a7784fa61`. It hung fifteen minutes downloading the `cf` binary and was reclaimed without running a test — infrastructure, and green on rerun.

Nothing else is outstanding on my side. The gate stays off and named-space key custody remains the enablement precondition, unchanged.

willkelly (approved):
_cf-review via Hephaestus `openai/gpt-5.6-sol`, on behalf of @willkelly._

Approved on substance at `ec2d98f6944558506174dd08fba241c6a4600e58`.

The previous runtime blockers are fixed: path-prefixed API URLs are preserved and signed consistently, and backward clock excursions neither stall nor refund the limiter. The lifecycle, authorization, audit/recovery, and break-glass provisioning invariants from earlier rounds remain intact.

Focused verification passed: 46 Toolshed steps, 41 ingest CLI steps, the final 17-step main-command test, 548 checked documentation blocks, the custody tripwire, formatting, lint, type checks, and LSP diagnostics.

**Merge condition:** all required CI checks must be green before landing. Coverage Check was still running when this approval was submitted; aggregate Status should be treated as its consequence rather than a separate code finding.

Non-blocking follow-ups:

- Correct the remaining live-doc wording that says disabled channels always receive the equalized 401. The intended contract is wrong/unknown token → 401; valid current token on a disabled/revoked channel → actionable 403.
- Consider forcing the concurrent-provision test's interleaving with a barrier so its one-winner assertion cannot depend on scheduler timing.

PR #5369, opened 2026-08-05
docs: server-execution v2 stage C.3 — flag retirement + doc archival

seefeldb:
## Why

The last PR of the stage-C train (C.1 #5356 → C.2 #5367 → this): the code is
gone, so the registry entry and the specs that described the deleted form
retire per the documentation lifecycle. This closes plan Phase 1 stage C —
the parent checkbox and the stage-C success criterion tick here.

Predecessor: C.2, #5367.

## Flag retirement

The `persistentSchedulerState` entry moves from the registry's Category 1 to
**Appendix A: Removed and never-shipped flags**, with the full record: what
it did, when it was deleted (stage C.2), what replaced it (the
`scheduler_basis` index), the one-time warm-start loss, the hello-degrade
compat story, and pointers to the archived specs.

## Doc archival (per docs/README.md's lifecycle)

- `docs/specs/persistent-scheduler-state.md` → `docs/history/specs/`, whole,
  with the metadata header (`superseded-by: serving-loop.md`). Its one piece
  of still-live semantics — the `materializerWriteInputPaths` contract — was
  already covered by scheduler-v2 README §4.3, which gains the concrete
  emission/resolution paragraph; the ts-transformers current-behavior spec's
  pointer moves there.
- `docs/specs/scheduler-v2/incremental-observation-adoption.md` →
  `docs/history/specs/scheduler-v2/`, whole (flag-ON-only feature, deleted
  with its carriage in C.2). serving-loop §3b's D9 collateral list now names
  all three archived specs.
- `docs/specs/scheduler-v2/per-doc-rehydration.md`: the persisted-form
  account is extracted to
  `docs/history/specs/scheduler-v2/per-doc-rehydration-persisted-form.md`
  and archived; the live file is rewritten to what remains true — the
  doc→deriver identity invariant (`pieceId` IS the derived doc, still
  load-bearing for the plan-B wake grouping and the basis index's keying)
  and the post-reduction resume posture (fresh runs under the sync hold).

Live docs that leaned on the deleted form are updated in the same PR:
scheduler-v2 README (the persistence header block, design principle D8, and
§9 rewritten to record the reduction), TIMING_SIDE_CHANNELS' channel-5 row
and its long-form appendix note (the push-shaping reason is gone; the N10
re-check is owed), computed-cell-identity's relationship section (the
keep/drop machinery is prior art to rebuild, not a live dependency),
module-loading's identity-consumer framing, the space-model reactivity open
question, the specs index, and the three collection builtins' resume
[...]

PR #5371, opened 2026-08-05
feat: server-execution v2 stage D — action transactions seal into the wave

seefeldb:
## Why

Server-execution v2's serving loop needs one derived commit per wave, but everything that makes an action run sound — the reactivity log feeding the dependency graph, CFC evaluated against the run's own reads, failure isolation — hangs off the per-action `IExtendedStorageTransaction`. Stage D resolves that tension exactly the way serving-loop.md §3d rules it: `action(tx)` keeps its object and interface, and only the DESTINATION changes. Server-side, under the flag, an action tx SEALS into the wave accumulator; on a client (the OFF arm always) seal == commit as today. This lands the whole transaction half of the loop dark, so stage F can build the SpaceServer on machinery whose concurrency semantics are already model-checked and tested.

Phase 1 stage D of [the plan](docs/plans/server-execution-v2.md) (checkbox ticked), stacked on the stage C train — prior PR: #5369. Flag OFF throughout; nothing in production installs a seal destination until stage F's serving loop.

## What

**The seal seam** (`extended-storage-transaction.ts`, §3d's anchor): `commit()` dispatches to an installed `TransactionSealDestination` instead of `this.tx.commit()`. Everything above the dispatch — the per-action-run CFC gates (§3c: the unit is the RUN, `action × instance`, never the action) — and below it (commit callbacks, post-commit side effects) fires for both destinations. The destination is a write-once pin per tx (the CFC-pin pattern), installed by `runtime.edit()`; `Runtime.installSealDestination` throws off the flag, so the OFF arm structurally cannot seal.

**Sealing** (`v2-transaction.ts sealInto` + the provider's `sealNative`): the tx's own close machinery runs — validation, per-space native commits in commit order — and each space's commit applies to the replica's optimistic overlay WITHOUT pushing. That overlay is the accumulator's layered view: store snapshot at the wave's input seq + previously sealed writes, read by later action runs through the ordinary read path. The built `ClientCommit` carries the read set (confirmed reads with store versions; pending reads naming earlier sealed commits — the in-wave edges), which feeds the wave's per-doc CAS basis and the §3b basis rows. The verdict is deferred: `committed` promotes at the wave's seq, `withdrawn` rolls back through the same rejection path a refused push takes.

**The wave accumulator** (`packages/runner/src/executor/wave.ts`): records contributions in seal order under stamped run contexts (acting identity, eventId, kind — attached at seal time, when the run still knows who it ran as; the commit step only batches what sealing attached). The commit step implements §3d's per-WRITE-CLASS conflict handling, disposed per (contribution, doc):

- pure derivation writes on docs whose head passed the basis DROP (counted `supersededWrites` — §7's counter shape), and derivations that READ a withdrawn sealed write drop with them (nothing derived from withdrawn state commits);
- non-re-derivable consequences REBASE — field-level merge, re-CASed against the exact head the merge decision observed, inside the store transaction — or REQUEUE their event whole, with the same-wave cascade-parent closure — never lost, never doubled: the spec model's C8a/C8b/C9 distinction, pinned end to end in the tests (`requeuedEventIds` reports surviving requeues only, per the model's C8d rollback closure);
- seal order is preserved into the batch (the §3 amendment's sealing-order MUST binds the loop's processing order; this step never reorders);
- multi-space seals commit foreign-first, home-after-success, and a foreign failure withholds the home commit (protocol.md §2b);
- a wave sealed under a lapsed lease tenure aborts at its commit step (serving-loop.md §2's stop-committing MUST, via stage B's `ExecutionLeaseCycle` tenure contract);
- a precondition failure resolves per owning contribution (named by index), never as a whole-wave abort.

Whole-wave CAS failure and blind derived writes are structurally impossible: every conflicted doc is terminally disposed per class before the batch goes down, and the engine re-verifies every written doc inside the wave's store transaction — basis for ordinary docs, exact decision-time head for rebased ones — reporting per-doc conflicts the resolve loop folds in (monotone convergence, no timers).

**Engine surface** (`packages/memory`): derived-class commits carry the per-write annotation pair of protocol.md §1 — the explicit `scope_key` KEYS scoped rows at admission (fail-closed: a scoped write without one is rejected, never silently resolved to `user:<serviceDID>` — the empty-instance trap §2 names) and the acting identity is recorded, not read — plus `consequenceOf`; both refused on any other class (the §7 closed list; `ClientCommit` cannot express them). `applyWaveCommit` runs re-verification + per-index precondition pre-check + admission/apply + basis-row overwrites in ONE engine transaction. `scheduler_basis` gains its writer (`v2/scheduler-basis.ts`): overwrite-in-place per `(action, action_scope_key)`, §3b's binding unit, enforced within a wave too (the last run of an instance replaces, never unions). `Server.engineForSpace` exposes the co-hosted direct-engine plane the sink and the lease already need. At OFF-arm cardinality 1 the annotations derive from the runtime's own authenticated session; the nine-site scope_key re-keying stays stage E, and the shared `DerivedWriteAnnotation` shape lives in the wire-shape module per the LD3 direction.

**Explicitly not here** (per the plan's stage cut): the serving loop, activation, the watermark doc, push (stage F); the outbox identity carriage for effect completions (stage G — completions never pass sealing); instance re-keying (stage E). Two documented stage-D bounds in the reference sink: foreign provisioning batches apply without the §2 delegated-identity admission row (lands with stage F's provisioning relocation), and sqlite ops in wave batches need stage G's attachment machinery.

## Adversarial review (standing merge protocol)

An adversarial subagent review of the full diff ran after green, attacking spec conformance, OFF-arm neutrality, and test vacuity. It reproduced one blocker and surfaced ten further findings; all are addressed in the follow-up commit (`fix+test: address the stage-D adversarial review batch`):

- **Blocker (reproduced end to end): the rebase disposition was cached per doc, not per (contribution, doc)** — one handler's legitimately-commuting patches absolved every later contribution writing the same conflicted doc, letting a superseded derivation or an overlapping second consequence commit blind over authored data. Fixed; both reproduced manifestations are now regression tests.
- The rebase now literally satisfies §3d's "re-CAS against the new head": the batch carries each rebased doc's decision-time head and the engine requires equality inside the transaction; a mismatch resets the rebase state and the wave re-decides. The former accumulator-trusting exemption is gone.
- Precondition failures are named by index (`WavePreconditionError`) and resolve per owning contribution instead of aborting the wave.
- §3b's overwrite unit is enforced within a wave (last run of an instance replaces; no union), and the sink-side grouping helper — which embedded a raw NUL byte that made the file binary and grep-proof against §8's tripwire mechanism — is deleted.
- `requeuedEventIds` reports surviving requeues only (the model's C8d closure: a cascade child of a requeued parent has no durable entry to retry).
- New tests for the previously-untested seams: the sink-race re-resolve loop, the scoped-write ADDRESSING half end to end, within-wave same-instance basis replacement, and the per-index precondition path.
- protocol.md §7 now states the annotation pair's stored representation in the closed-list bullet, and the seal-time side-effect seam carries an explicit stage-F/G consumption warning (sealed ≠ durable; dispositions come from the wave outcome).

The review confirmed OFF-arm neutrality (dispatch reduces to the identical expression with no destination; new `applyCommit` guards unreachable for session-facing calls; migrations `hasColumn`-guarded with NULL-only new columns), the doc edits as anchor-refresh-only, and the §8 tripwire scan clean.

## Verification

- `packages/spec-model` stays green (18 tests) — the executable model remains the semantic oracle; the one implementation/model divergence the review found (requeue reporting vs C8d) was resolved toward the model and raised in the review notes above.
- Runner tests (`test/executor-wave.test.ts`, 13 steps) drive a real Runtime + storage stack + engine through seal, layered-view reads, C8a superseded drop + dependency closure, C8b requeue-then-land-exactly-once (with annotation + basis-row assertions), rebase (single and mixed-contribution), the two reproduced blocker scenarios, the sink-race re-resolve loop, scoped-write addressing, tenure abort, foreign-first + foreign-failure, per-action isolation, and within-wave basis replacement — including stage B's derived-class lease admission for real.
- Memory tests (`test/v2-wave-commit-test.ts`, 9 tests) pin the annotation keying (fail-closed + scope-coherence), the closed-list refusals, named-conflict re-verification (basis and rebased-head), per-index precondition naming, and the basis writer's overwrite-in-place.
- OFF arm: full runner suite green (1109 tests / 5991 steps), memory suite green (446), `deno task check`, `fmt`, `lint`, `check-docs`, `check-skill-facts`, `check-no-waitfor`, `check-conflict-markers`, and the dependency gates all pass.
- Docs move with the change: plan checkbox, `EXPERIMENTAL_OPTIONS.md` (ON-arm surface + status), protocol.md §7's stored-representation clarification, and the drifted §2b/§3d file:line anchors refreshed (serving-loop.md, protocol.md, runtime-mapping.md N53).


seefeldb:
## Independent adversarial review (coordinator round) — LANDABLE-WITH-FIXES → fixes landed as bc2e817a1

Reviewed at caaae1dad by a fresh-eyes adversarial agent, after and independent of the PR's own self-review round. Fix batch: `bc2e817a1`. Nothing merges — the train holds until the final phase exists.

**Verified deep and held:** all six §3d behaviors implemented as claimed (pure-drop + dependency closure, rebase re-CAS at the decision-time head, requeue with cascade closure, sealing-order preservation, foreign-first multi-space, tenure abort) with file:line traces; OFF-arm neutrality is structural (the seal dispatch reduces to the identical expression with no destination; `installSealDestination` throws off-flag; `derived` unclaimable off-flag); the self-review's per-(contribution,doc) blocker fix is real and regression-tested; migrations `hasColumn`-guarded and NULL-only; §8 tripwire scan clean; all suite counts independently reproduced.

**Findings → fixes (all in bc2e817a1):**

- **MAJOR-1** — engine admission silently *applied* a `scopeKey` annotation attached to a space-scoped op (the validation loop skipped space-scoped ops before the coherence check), landing the row under the annotated instance while declaring scope "space" — the mirror image of the empty-instance trap. Now refused loudly on both the `applyCommit` and `applyWaveCommit` paths; red-first negative tests.
- **MAJOR-2** — three closure arms were mutation-provably untested (probes survived deletion with all suites green): the C8d surviving-requeues filter, the cascade-parent requeue closure, and the requeued-read drop (a derivation that read a requeued handler's sealed write committing blind — §3d's forbidden write). Three new probe-verified tests; each probe now red, reverted, green (19/19).
- **MINOR-1** — cross-space reads of withdrawn sealed writes escaped the dependency closure (only the home commit's pending reads were inspected). Closure extended per space; red-first test flips `[requeued, committed]` → `[requeued, dropped]`. One honest residue is documented as the **third stage-D bound**: a read in a space the tx never wrote doesn't reach the accumulator — closing it needs the seal handoff to carry read-only spaces' read sets (stage-F warning at the sink header).
- **MINOR-2** — the §7 stored-representation sentences added by this PR had no coverage rows → verification-coverage.md 281→286 rules with a dated delta, plus **OW9**: the spec model has no rebase arm, so the drop/rebase/requeue three-way is pinned by impl tests alone until the next model commit.
- **MINOR-3** — sink replay-keying footgun (per-instance `#localSeq` restart under a caller-chosen sessionId vs the engine's UNIQUE(session_id, local_seq) dedupe) → constraint documented at the `EngineWaveCommitSink` constructor; stage F owns the construction choice.
- **MINOR-4** — the wave-side per-owner precondition resolution loop was untested → new test (gate owner requeued, unrelated derivation committed, no whole-wave abort).
- **NIT-1** — two ALTERs behind one `hasColumn` guard left a crash window that would brick every subsequent commit (migrations run on store open regardless of flag) → one guard per ALTER.
- **NIT-2** — a preconditions-only contribution short-circuited unvalidated to `committed {seq: 0}` → the short-circuit now requires empty preconditions; red-first test shows `dropped`, store untouched.
- **NIT-3** — a scoped non-sqlite op in a FOREIGN batch would have silently keyed from the sink's own principal → refused with an error naming the stage-F/G delegated-admission dependency.
- **NIT-4** — body step count corrected. Current counts at bc2e817a1: runner 1109 tests / 5997 steps, memory 447, executor-wave 19 steps, v2-wave-commit 10 tests, spec-model 18.

**Owner-ruled spec edits that rode this batch (both RULED 2026-08-05):**

1. protocol.md §1's `system` row is now **producer-defined** with exemplary (not closed) contents — blob-metadata writes via `Server.writeDocument` named; a new direct-write caller is a spec decision, never a silent inheritance; the no-user-attribution-in-the-commit-ledger consequence is accepted, with per-user blob attribution named as an out-of-scope future hardening.
2. protocol.md §2 gains the **derived-envelope defense-in-depth** rule: at admission, a `derived` commit's producing session must be the lease holder's own service session — refused otherwise (mirrors the spec model's `admitDerived`). Implementation lands with stage F (impl-gate coverage row; the plan's stage-F bullet carries the task).

**Open owner questions from this review (flag-don't-fill; rulings requested):**

- **Q1 (substantive):** superseded runs still write basis rows — a survivor whose writes were dropped per-doc records its reads as current — and §3d's "the next wave recomputes that derivation" has no obvious trigger when the intrusion hits a producer's OUTPUT doc the producer doesn't read. What re-arms the recompute, and should dropped runs' basis rows land at all?
- **Q2:** write-class granularity is per RUN in the implementation (every write of an event-handler run is treated non-re-derivable), while §3d's text classifies WRITES. One-line ratification or correction requested.
- **Q3:** resolved by register — OW9 records the model's missing rebase arm (trigger: next model commit).
- **Q4:** unstamped seals run as anonymous derivations with no basis rows (code documents "sound but conservative") — one spec sentence requested to bless or forbid.

seefeldb:
## Rulings landed: Q1(c), Q2, Q4(ii) — commit 4b9a91c6e

The three open questions from the review ledger above are RULED (owner, 2026-08-05) and landed on this branch:

- **Q1 → (c), recompute-by-dependency only.** §3d no longer promises an unconditional recompute after a superseded drop; recomputation arrives solely through the ordinary dependency path. The owner's rationale is landed as the attributed blockquote in §3d: a genuinely derived doc has no other writers, and shared state that derivations also update is either written with non-conflicting operations or read-first — and a read-first derivation carries the doc in its read list, hence in its basis rows, hence the authored intrusion re-runs it through the normal affected-graph path. Corollary stated explicitly (and now pinned by a test): a survivor whose writes were dropped per-doc still lands its basis rows — its reads are true, and no recompute-owed mark exists.
- **Q2 → ratified.** Write class is determined by the producing RUN's kind: every write of an event-handler run is non-re-derivable; every write of a derivation run is a pure derivation write.
- **Q4 → (ii), stamped seals only.** Sealing an UNSTAMPED transaction under the flag is refused loudly; the anonymous-contribution machinery is deleted (basis rows now unconditional for survivors); the old acceptance test flipped to pin the throw (red-first both directions). Stage F names the sanctioned internal stamp kinds (e.g. bookkeeping) when it installs the seal destination. The OFF arm is untouched — no destination, no refusal path.

Register delta rides the commit (serving-loop 71→76, total 286→291, dated delta block). The spec model was checked against Q1(c) and is CONSISTENT with it as-is: the model's derivation fire predicate reads the raced output doc every wave (model.ts:597-603), so C8a's "recomputed over the intrusion" already flows through declared reads — the ruling's read-modify-write arm — and no re-arm bookkeeping exists anywhere in the model. One named residual: the pre-ruling "next wave recomputes" shorthand still appears in threat-model/trace contexts (protocol.md:81/:331, README.md:85, runtime-mapping.md:233, scenario-traces.md:564/:631) — the scenario-trace re-run now in flight checks the trace lines for reference-answer drift, and the rest sweeps with stage F's docs pass.

seefeldb (commented):
Verdict: no additional layer-local finding in the Stage D seal/withdraw/settlement paths at this exact head.

The wave contribution grouping and conflict disposition are coherent with the plan. This is not a landing recommendation for the cumulative stack: this PR's current check rollup has Test 2/6, 4/6, 6/6, and Status failures, and later layers have blocking findings.

seefeldb:

PR #5373, opened 2026-08-05
fix(runner): a property named like an Object.prototype member is still data

bfollington:
## Why

Follow-up to #5357. Reviewing that fix, @seefeldb asked whether the same mistake was elsewhere in the codebase. **It is — five more places, and one of them throws.**

`toString`, `valueOf`, `hasOwnProperty`, `isPrototypeOf`, `propertyIsEnumerable` and `toLocaleString` are ordinary property names. They are **not** reserved by `unsafeObjectKeyIn()` — #5264 reserves only `__proto__` and `constructor` — so they store and round-trip like any other key. But several paths asked whether such a key was present using `in`, which walks the prototype chain, so the answer was yes for every object. Where the code then *indexed*, it got the inherited function instead of data.

### Symptom 1 — schema defaults silently dropped, caller gets a function

Reproduced against the real runtime, with a control case to prove the harness works:

```
CONTROL: an ordinary missing property gets its default    ✅ age → 30
a property named `toString` gets its default              ❌ got [Function: toString]
a property named `valueOf`  gets its default              ❌ got [Function: valueOf]
```

The schema promised a string; the caller receives `Object.prototype.toString`.

### Symptom 2 — writing such a property throws

```
Error: Cannot compare a function value.
  at valueEqual (packages/data-model/src/valueEqual.ts:44)
  at applyMutablePathWrite (packages/runner/src/storage/transaction/mutable-path-write.ts:214)
  at V2StorageTransaction.writeBatchRun (…)
  at applyChangeSet (packages/runner/src/data-updating.ts:1769)
  at CellImpl.set (packages/runner/src/cell.ts:1483)
```

In the storage layer's leaf write, `previousValue = obj[leafKey]` falls through to the prototype, `leafKey in obj` reports present, and `valueEqual` is handed a function. `cell.set({ valueOf: 99 })` throws — a write refused because of a method the data never had.

## What Changed

| file | was |
|---|---|
| `storage/transaction/mutable-path-write.ts` | presence **and** the `previousValue` read — the one that throws |
| `schema.ts` | `key in defaultValue` |
| `traverse.ts` | `propKey in filteredObj` |
| `pattern-binding.ts` | `key in value` |
| `data-updating.ts` | `key in newValue`, `key in currentRecord`, the `currentRecord[key]` fall-through, and two `for...in` walks over own keys |

`Object.hasOwn` at each membership test, own-guarded reads where indexing fell through, `Object.keys` for the `for...in` walks.

Worth noting the array branch beside the storage fix was **already correct** — `slot in parent` is right for sparse-hole detection, because arrays don't inherit index properties. Same operator, different question; that's the distinction this whole class turns on, and it's the same one that makes the `has` trap in #5357 correct as-is.

## Swept and deliberately not changed

- **`html/src/worker/reconciler.ts`** (3 sites) uses the same predicate, but every key reaching it is a fixed internal literal (`"author"`, `"$value"`, `"maxConfidentiality"`), none colliding with `Object.prototype`. Wrong predicate, unreachable defect — flagging rather than changing, since it's your call whether to harden it.
- **The other five non-test `new Proxy(...)` sites**: `extended-storage-transaction.ts` correctly delegates to `Reflect.getOwnPropertyDescriptor`; the rest define no shape traps. No second copy of the #5357 bug.
- **The many `i in array` checks** across data-model — sparse-hole detection, a correct use of `in`.

## Test plan

- New `packages/runner/test/prototype-named-properties.test.ts`: all six prototype-member names, plus a stored-value-wins case and a write-not-dropped case. Each has a control using an ordinary name, so the suite fails loudly rather than vacuously if these paths stop being exercised. It also pins the premise (`name in {}` true, `Object.hasOwn({}, name)` false) so the fixture can't quietly stop reproducing.
- `cd packages/runner && deno task test` — **1148 passed / 0 failed** (baseline on `main`: 1146 / 0).
- `deno fmt --check` and `deno lint` clean on all changed files.

Refs [CT-1951](https://linear.app/common-tools/issue/CT-1951).

---
## Summary by cubic
Fixes prototype-chain leaks across runner, CLI, SQLite, and path helpers so properties named like `Object.prototype` members are treated as normal data. Restores correct defaults, required checks, deletes, and path operations; prevents the storage write crash (“Cannot compare a function value”). Addresses CT-1951 and follows up on #5357.

- **Bug Fixes**
  - Use `Object.hasOwn` for membership checks and `Object.keys` for own-key iteration; guard reads to avoid inherited functions.
  - Close remaining paths: `path-utils` (`get/set/hasValueAtPath` now own-only), `data-updating` (removals and eager scoping), `traverse` (required/defaults), `pattern-binding`, `builder/pattern`, `cfc/{prepare,structured-result}`, SQLite row/column checks, and CLI (`exec-schema`, `piece-get-transform`, `callable`).
  - Storage write path: own-only presence/previous-value read to avoid comparing inherited functions.
  - `piece-helpers.resolveCellPath`: treat missing prototype-named segments as absent and report “property not found”.
  - Tests: added `prototype-named-properties.test.ts` and expanded `piece-helpers.test.ts` to pin behavior for all six names and path helpers.

<sup>Written for commit 9e8240ac9f5d821911674628d6d7f29fd3d6c8ed. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5373?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/traverse.ts
+++ b/packages/runner/src/traverse.ts
@@ -4294,7 +4294,12 @@ export class SchemaObjectTraverser<V extends FabricValue>
     // Apply defaults from our schema
     if (isRecord(schema) && schema.properties) {
       for (const propKey of Object.keys(schema.properties)) {
-        if (propKey in filteredObj) {
+        // `Object.hasOwn`, not `in`: `propKey` is a schema-declared property
+        // NAME and `filteredObj` is data. `in` walks the prototype chain, so a
+        // schema property called `toString` or `valueOf` looked already-present
+        // on every object, its default was never applied, and the caller got
+        // `Object.prototype`'s function where the schema promised a value.
+        if (Object.hasOwn(filteredObj, propKey)) {
           continue;
         }
         const subSchema = ContextualFlowControl.getSchemaAtPath(schema, [
--- a/packages/runner/src/traverse.ts
+++ b/packages/runner/src/traverse.ts
@@ -3295,7 +3295,13 @@ export class SchemaObjectTraverser<V extends FabricValue>
           } else if (branch.required !== undefined && valueIsRecord) {
             match = true;
             for (const req of branch.required) {
-              if (!(req in (doc.value as Record<string, unknown>))) {
+              // `Object.hasOwn`, not `in`: `req` is a schema-declared property
+              // NAME and the value is data. `in` walks the prototype chain, so
+              // a required property called `toString` was satisfied by every
+              // record whether or not it carried one.
+              if (
+                !Object.hasOwn(doc.value as Record<string, unknown>, req)
+              ) {
                 match = false;
                 break;
               }
@@ -4350,7 +4356,10 @@ export class SchemaObjectTraverser<V extends FabricValue>
       const required = schema["required"] as string[];
       if (Array.isArray(required)) {
         for (const requiredProperty of required) {
-          if (!(requiredProperty in filteredObj)) {
+          // `Object.hasOwn`, not `in` — see the sibling check above: a required
+          // name matching an `Object.prototype` member read as present on every
+          // object, so the value was accepted instead of rejected.
+          if (!Object.hasOwn(filteredObj, requiredProperty)) {
             logger.info("traverse", () => [
               "Missing required property",
               requiredProperty,
@@ -4646,7 +4655,9 @@ export function canBranchMatch(
       (Array.isArray(resolved.type) && resolved.type.includes("object"));
     if (typeIncludesObject && Array.isArray(resolved.required)) {
       for (const req of resolved.required) {
-        if (!(req as string in value)) return false;
+        // `Object.hasOwn`, not `in`: same own-presence question as the other
+        // two required-property checks in this file.
+        if (!Object.hasOwn(value, req as string)) return false;
       }
     }
   }
--- a/packages/cli/lib/callable.ts
+++ b/packages/cli/lib/callable.ts
@@ -344,7 +344,7 @@ function cloneWithoutBoundToolKeys(
 
   const required = Array.isArray(schema.required)
     ? (schema.required as string[]).filter((key) =>
-      key !== "result" && !(key in extraParams)
+      key !== "result" && !Object.hasOwn(extraParams, key)
     )
     : undefined;
 
--- a/packages/cli/lib/callable.ts
+++ b/packages/cli/lib/callable.ts
@@ -344,7 +344,7 @@ function cloneWithoutBoundToolKeys(
 
   const required = Array.isArray(schema.required)
     ? (schema.required as string[]).filter((key) =>
-      key !== "result" && !Object.hasOwn(extraParams, key)
+      key !== "result" && !(key in extraParams)
     )
     : undefined;
 
--- a/packages/cli/lib/exec-schema.ts
+++ b/packages/cli/lib/exec-schema.ts
@@ -355,9 +355,7 @@ function parseObjectInput(
   // JSON input validation is deferred to the runner.
   if (!usedJson) {
     for (const key of requiredFlags(schema)) {
-      // `Object.hasOwn`: a required flag named for a prototype member would
-      // otherwise read as supplied on every input.
-      if (!Object.hasOwn(input, key)) {
+      if (!(key in input)) {
         throw new Error(`Missing required flag --${flagNameForKey(key)}`);
       }
     }
--- a/packages/cli/lib/exec-schema.ts
+++ b/packages/cli/lib/exec-schema.ts
@@ -355,7 +355,9 @@ function parseObjectInput(
   // JSON input validation is deferred to the runner.
   if (!usedJson) {
     for (const key of requiredFlags(schema)) {
-      if (!(key in input)) {
+      // `Object.hasOwn`: a required flag named for a prototype member would
+      // otherwise read as supplied on every input.
+      if (!Object.hasOwn(input, key)) {
         throw new Error(`Missing required flag --${flagNameForKey(key)}`);
       }
     }
--- a/packages/cli/lib/piece-get-transform.ts
+++ b/packages/cli/lib/piece-get-transform.ts
@@ -895,7 +895,8 @@ function projectValue(
     return projected;
   }
   for (const [key, childSchema] of Object.entries(properties)) {
-    if (key in value) {
+    // `Object.hasOwn`: schema-declared `key` against data.
+    if (Object.hasOwn(value, key)) {
       projected[key] = projectValue(
         value[key],
         childSchema,
@@ -1045,7 +1046,9 @@ export function selectSourceSchema(
       purpose,
     );
   }
-  const selectedRequired = required?.filter((key) => key in properties);
+  const selectedRequired = required?.filter((key) =>
+    Object.hasOwn(properties, key)
+  );
   return {
     ...metadata,
     properties,
--- a/packages/cli/lib/piece-get-transform.ts
+++ b/packages/cli/lib/piece-get-transform.ts
@@ -895,8 +895,7 @@ function projectValue(
     return projected;
   }
   for (const [key, childSchema] of Object.entries(properties)) {
-    // `Object.hasOwn`: schema-declared `key` against data.
-    if (Object.hasOwn(value, key)) {
+    if (key in value) {
       projected[key] = projectValue(
         value[key],
         childSchema,
@@ -1046,9 +1045,7 @@ export function selectSourceSchema(
       purpose,
     );
   }
-  const selectedRequired = required?.filter((key) =>
-    Object.hasOwn(properties, key)
-  );
+  const selectedRequired = required?.filter((key) => key in properties);
   return {
     ...metadata,
     properties,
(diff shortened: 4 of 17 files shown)


bfollington (commented):

Scope: PR #5373 plus a syntax-aware repository sweep for prototype-aware membership on data/schema-derived keys. Coverage: deep-read all six changed files; inspected every runtime `in` / `for...in` site in the relevant packages; added targeted regression probes in an isolated copy. Checks: the PR's test passes (6/6); changed-file `deno fmt --check` and `deno lint` pass; the added probes below fail as described.

## 🔴 Blocking

1. **[major, verified] Removed prototype-named keys are still retained.** In `packages/runner/src/data-updating.ts:1578-1579`, the removal pass still uses `for...in` plus `key in newValue`. Starting from `{ name: "John", toString: "stored" }` and then setting `{ name: "John" }` leaves the own `toString` property behind, because `"toString" in newValue` is true through `Object.prototype`. I reproduced this through the public `Cell.set()` path. Please iterate `Object.keys(currentRecord)` and use `Object.hasOwn(newValue, key)`, and add the deletion case to `prototype-named-properties.test.ts`.

2. **[major, verified] Missing required prototype-named properties are still accepted.** `packages/runner/src/traverse.ts:3298`, `:4353`, and `:4649` all use `in` for schema `required` names. With a schema requiring a numeric `toString`, reading `{ name: "John" }` returns `{ name: "John" }` instead of rejecting the value as it does for an ordinary missing required field. This is the same data-name/own-presence question as the default handling fixed immediately above it. Please use `Object.hasOwn` in all three required-property checks and pin the public `asSchema(...).get()` case.

3. **[major, verified] The exported runner path helpers still expose the same throw and inherited reads.** `packages/runner/src/path-utils.ts:12-20,34,42` reads an absent path segment directly or tests it with `in`. Targeted probes show:
   - `setValueAtPath({}, ["toString"], "stored")` throws `Cannot compare a function value`;
   - `getValueAtPath({}, ["toString"])` returns `Object.prototype.toString`;
   - `hasValueAtPath({}, ["toString"])` returns `true`.

   These helpers are public and are used by CFC preparation. The storage-v2 path helper already models the right split: own-property checks for records, sparse-slot handling for arrays. Please apply that distinction here as well.

4. **[major, verified] Piece path resolution treats a missing prototype-named segment as present.** `packages/runner/src/piece-helpers.ts:71-73` uses `segment in parentValue`, so `resolveCellPath(..., ["toString"])` returns `undefined` instead of raising the documented “property not found” error. I reproduced this in the existing `piece-helpers.test.ts` harness. Please use own membership for record segments and add the collision case.

## Possible follow-ups from the completeness sweep

These are source-verified instances of the same predicate shape, but I did not build end-to-end repros for each subsystem. They should either join this PR or be captured in a follow-up issue rather than being described as swept-clean:

- `packages/runner/src/cfc/prepare.ts:2808,2842-2848` — wildcard policy paths read inherited members from current/previous values.
- `packages/runner/src/builder/pattern.ts:927` and `packages/runner/src/cfc/structured-result.ts:213` — data-derived keys are tested against schema `properties` with `in`.
- `packages/runner/src/builtins/sqlite-builtins.ts:210` and `packages/memory/v2/sqlite/row-label.ts:601,650` — column/policy field names are tested against row records with `in` and then indexed.
- `packages/cli/lib/piece-get-transform.ts:898,1048`, `packages/cli/lib/exec-schema.ts:358`, and `packages/cli/lib/callable.ts:347` — schema-derived names are tested against data or plain maps with `in`.
- `packages/runner/src/data-updating.ts:1658` — the currently-disabled `compactChangeSet()` helper can falsely treat `{}.toString` as contained and subsume a child change if re-enabled.

I also confirmed the array-index `in` checks are sparse-hole tests and should stay, and the fixed string/discriminator checks do not collide with `Object.prototype` names.

**Verdict:** the changes made are directionally correct, but this still needs changes before merge because public write/delete/schema paths retain the exact bug class the PR intends to close.

bfollington:
All four blocking findings fixed, and the follow-up list folded in rather than deferred. Thanks — the sweep critique was the important part.

**My sweep was bad and I want to be precise about how**, since "swept-clean" was the load-bearing claim. My grep required a *parenthesised* `x in y` and filtered out quoted literals to cut discriminant noise. That pattern misses `? properties[key] : …` ternaries, multi-line conditions, and anything where the operands aren't bare identifiers — i.e. most of the shape it was looking for. The four blockers and the follow-up list were all reachable by a syntax-aware pass, which is what I should have done.

**Blocking, each with a test that fails when the fix is backed out:**

- **Removal** — verified your repro. `{name, toString: "stored"}` → `{name}` left the property behind; the key could never be deleted.
- **Required** — all three `traverse.ts` sites. The test asserts a prototype-named required property produces the *same* result as an ordinary missing one, so it pins the invariant rather than a specific rejection shape.
- **Path helpers** — all three symptoms reproduced and fixed. Reads are own-only now, which also corrects the intermediate-descent branch: an absent prototype-named segment was previously seen as an existing non-object and descended into rather than created.
- **Piece paths** — fixed. Worth noting the harness had the same bug: `makeCell`'s `key()` indexed its children map directly, so a prototype-named segment got back an inherited function instead of "no child", and the test failed with `currentCell.get is not a function` before it could reach the assertion. Fixed the stub too.

**Follow-ups folded in** rather than filed separately — each is the same one-word change, so splitting them would mean reviewing identical reasoning twice: `builder/pattern.ts`, `cfc/structured-result.ts`, both `cfc/prepare.ts` walkers (presence *and* the previous-value read), `sqlite-builtins.ts`, `memory/v2/sqlite/row-label.ts` (both sites), the three CLI files, and `hasPath()`'s record branch. Happy to split any of it back out if you'd rather review the CLI/memory changes separately.

**Left alone deliberately:** the array branches. `index in array` is sparse-hole detection and arrays don't inherit index properties — same operator, different question. That's the distinction that makes the `has` trap in #5357 correct as-written, and it's the thing to hold onto reading this diff.

Suites: runner **1149 / 0**, memory **478 / 0**, cli **123 / 0**. `deno check`, `fmt --check`, `lint` clean on all 14 files. (Type-checking caught a real mistake in my first path-utils pass — `ownSegment` returned `unknown` into `valueEqual`'s `FabricValue`; the package suites run `--no-check` so only the explicit check surfaced it.)

bfollington (commented):

Re-review scope: latest head `9e8240a`, including the author's fixes for all four prior blockers and every item from the follow-up sweep. I re-read the expanded 17-file diff, repeated the syntax-aware membership sweep, and ran the focused runner, CLI, and memory tests from an isolated archive. The focused tests pass, as do changed-file `deno fmt --check` and `deno lint`.

## 🔴 Blocking

1. **[major, verified] `getAtPath` still traverses into inherited prototype members.** [`packages/runner/src/traverse.ts:1861`](https://github.com/commontoolsinc/labs/blob/9e8240ac9f5d821911674628d6d7f29fd3d6c8ed/packages/runner/src/traverse.ts#L1861) still gates a data-derived path segment with `part in curDoc.value`. With a document whose value is `{}`, `getAtPath(..., ["toString"], ...)` descends as though the absent key were stored, appends `toString` to the document address, records a child read, and returns `Object.prototype.toString` (confirmed by a focused probe: `resultType: "function"`, identity-equal to the prototype member). An ordinary absent segment follows the missing-path branch, so the result still depends on whether the property name collides with `Object.prototype`.

   This is the same own-data predicate the rest of this PR now fixes, and `getAtPath` backs runtime traversal/link resolution. Please use `Object.hasOwn(curDoc.value, part)` here and add a regression alongside the existing `getAtPath` tests (with an ordinary-name control).

The prior four blockers are fixed correctly, and I did not find another actionable collision after classifying the remaining `in` sites (the retained numeric cases are sparse-array membership checks; fixed discriminator/schema-key checks do not collide with `Object.prototype`).

CI note: the only failing job is the lunch-poll browser integration timing out in `fillCfInput`; it appears unrelated to this change, but still needs a rerun/green check before merge.

**Verdict:** much improved, but the remaining runtime traversal path is still the exact bug class this PR intends to close.

packages/runner/src/traverse.ts
```diff
@@ -4294,7 +4300,12 @@ export class SchemaObjectTraverser<V extends FabricValue>
     // Apply defaults from our schema
     if (isRecord(schema) && schema.properties) {
       for (const propKey of Object.keys(schema.properties)) {
-        if (propKey in filteredObj) {
+        // `Object.hasOwn`, not `in`: `propKey` is a schema-declared property
```
ubik2, line 4303: I'm not sure this is right.
We can have:
```typescript
const a: {length: number} = new Uint8Array();
```
In that case, we want to verify that the object we're assigning to `a` has a `length` property. For a Uint8Array (or FabricBytes), that's not a `hasOwn`, so traverse will return `undefined` instead of the array length.

PR #5374, opened 2026-08-05
feat: server-execution v2 stage E — instance re-keying (scope name → scope instance)

seefeldb:
## Why

Every in-memory identity key in the runner used the scope NAME — `${space}/${scope}/${id}` — which is sound only while each runtime computes its own principal's instances (scoped cardinality 1) and collapses the moment one serving runtime holds every principal's instances: one principal's commit would wake every principal's node, one session's result cell would be every session's, one user's seed would suppress another's (scopes.md §7 M2). Stage E re-keys that vocabulary per scope INSTANCE now, dark and OFF-arm neutral, so stage F's SpaceServer lands on per-instance machinery instead of retrofitting the single biggest scope cost of the arc under an ON flag.

Phase 1 stage E of [the plan](docs/plans/server-execution-v2.md) (stage checkbox + stage-E success criterion ticked), stacked on stage D — prior PR: #5371. Flag OFF throughout; nothing here changes what runs, only how work is keyed.

## What

**The LD3 definition move leads** (key-vocabulary.md §3, ruled 2026-08-03): the `scope_key` vocabulary — `ScopeKey`/`ScopeKeyIdentity` types, the ONE `(scope, identity) → scope_key` constructor `resolveScopeKey`, the parse/inspect helpers (`scopeOfScopeKey`, `isScopeKey`, `canResolveScopeKey`, `principalOfSessionKey`, `resolvePrincipalSessionKey`), and `ProtocolError` (which the constructor throws) — moves to the wire-shape module `packages/memory/v2.ts`, beside `CellScope`. The engine imports and **re-exports the same objects** (`Engine.resolveScopeKey === resolveScopeKey`, pinned by test), so no twin definition exists; identity DERIVATION stays engine-owned admission machinery. `packages/state-inspector` migrates to the shared module per the ruling.

**The nine-site re-keying** (key-vocabulary.md §1, all anchors refreshed in the same doc): every site constructs `${space}/${scope_key}/${id}` (or its site-specific shape) via the shared constructor from an EXPLICITLY SUPPLIED identity — never ambient state. In the OFF arm that identity is the runtime's own authenticated session, exposed as `Runtime.scopeKeyIdentity` → `IStorageManager.scopeKeyIdentity()` (`{principal: as.did(), sessionId}` — exactly the pair the memory server resolves this manager's scoped rows from, so client keys embed the engine's row keys byte-for-byte). Site by site: the scheduler dependency-graph node key `entityKey` (+ the `addressesToPathByEntity` composite that fed the trigger index through an inline duplicate — now routed through `entityKey`); the result-pattern cache key `getDocKey` and its notification-invalidation twin; `byScope` result cells (`Map<ScopeKey, Cell>` over the resolved `effectiveOutputScope`); `seedMemoKey`; the schema-tracker coverage key and `getTrackerKey` (identity rides `TraversalContext.scopeKeyIdentity`); the storage manager's `#pendingLoads` (built WITH `entityKey` so the scheduler's park-key cross-matching stays one namespace); the notification-differential address identity `Address.toString` + `differential.ts` (threaded from the replica's owning session); graph-snapshot diagnostics. Scheduler state objects carry `scopeKeyIdentity` thunks wired to the runtime; per-run construction stays explicit.

**Couplings that share the tracker strings move in the same PR**: the memory server's query/watch path (`packages/memory/v2/query.ts`) writes `toDocKey` strings into the SAME `MapSet` the runner's traversal writes `getTrackerKey` strings into — both now instance-keyed by the QUERYING session's identity (`identityOf(manager)`, the same identity the manager threads into every scoped engine read). `fromDocKey` recovers the scope name from the instance key via the shared inspectors; a scope NAME in key position is now rejected. `refreshTrackedGraph`'s dirty loop gets a `canResolveScopeKey` guard: dirtiness still arrives scope-NAME-keyed (M4 — push re-keys at stage F), and a session with no identity for a scope cannot have tracker entries there (its loads throw), so the guard reproduces today's no-op instead of letting the constructor throw on the watch path.

**The wave accumulator conforms**: stage D's caller-supplied `resolveScopeKey` function option becomes a `scopeKeyIdentity: ScopeKeyIdentity`, constructed through the shared definition — a test can no longer hand it a different format — and the basis-row types (`actionScopeKey`, `entityScopeKey`) are typed `ScopeKey`, closing the format trapdoor into `scheduler_basis` (serving-loop.md §3b: stage E feeds the engine-side writer instance VALUES; real per-demand identities arrive with stage F's run contexts).

**OFF-arm neutrality, structural — and adjudicated per contract where re-keying met two latent writer-side bugs.** The structural argument (key-vocabulary.md §2): at cardinality 1 the scope-name→instance mapping under one fixed identity is injective, so every partition is unchanged. Re-keying surfaced two places where the OLD code interpolated a possibly-`undefined` scope raw into a key, so two constructions that should have agreed never matched. Each was resolved toward the side the repo's own tests pin as the contract:

- The inline coverage key in `isLinkedDocumentCovered` wrote an `"undefined"` scope segment for unscoped links, so coverage never matched them — keeping the covered-skip paths (which replace a repeat link's value with `null` in traverseCells mode) dead for the common case. Healing this is VALUE-VISIBLE (the adversarial review reproduced `{a: "hello", b: null}` where base returns `{a: "hello", b: "hello"}`, and the html reconciler renders those values), and the base diagnostics test pins the un-skipped re-walk counts. The landed code keeps the base outcome explicitly: coverage is only consulted for scope-declaring links, with a load-bearing comment; a new regression test pins repeat-unscoped-link values, and the restored diagnostics assertions pin the visit counts.
- `getDocKey` (the result-pattern cache WRITER) wrote the same raw-scope segment while the notification-eviction path normalized — so eviction silently missed every real unscoped entry. Here the repo's contract test ("clears cached patterns when storage notifies of changes") pins EVICTION with normalized keys: the writer was the bug, and re-keying makes it conform. The healing stands (guarded-off was tried and broke that base-green contract test); the delta on the healed path is work-shape — an idempotent re-prepare plus a pull read on unscoped result-doc notifications, no differing committed bytes (the review verified the unchanged-pattern re-fire cannot commit differing bytes: `samePattern` holds for content-hashed keyless patterns and the only write path is deepEqual-guarded; full suite and both-arm integration runs green). One results-visible edge exists and is a deliberate part of the healed contract: a child `stop()`ed while its parent keeps re-firing the SAME keyless pattern — at base an UNSCOPED such child stayed stopped forever (the base code's own comments call that the bug) while explicit-scope children restarted; healed, all restart. Called out in Flags as the healing's carve-out.

**Docs move with the change**: key-vocabulary.md re-anchored end to end and reframed to the landed state (the table now states the instance key and identity source per site); scopes.md §Anchors + §7 M2 resolution note (M3's precedent); serving-loop.md §3b DDL comment; protocol.md §1/§2 and runtime-mapping.md N57-adjacent refs repointed at the wire module (my engine edits shifted every cited engine.ts line; each cite was re-verified against the code it names). Plan checkboxes ticked. No new binding sentence was added to any spec doc (anchor refreshes and landed-state notes only), so no verification-coverage.md row is owed.

## Verification

- New contract tests: `packages/memory/test/v2-scope-key-test.ts` (format incl. encoding, throw conditions + predicate twin, parse helpers, engine re-export object identity — the single-definition property; 8 tests) and `packages/runner/test/scope-key-rekeying.test.ts` (partition equivalence vs the name-keyed form at cardinality 1, key stability, storage-row alignment, two-principal distinctness — fan-out readiness, manager identity contract; 5 steps). `v2-query-test.ts`'s doc-key contract updated to instance keys (scope-name key forms now rejected); `address.test.ts` gains the site-8 instance-key case.
- OFF arm: full runner suite green — 1110 passed (5998 steps) / 0 failed; memory 454 (145 steps); spec-model 18; state-inspector 88; fuse 347 / lib-shell 5 / schema-generator 55 / data-model 48 (the packages `deno task check` does not cover); `deno task check`, `fmt --check`, `lint`, `check-docs` (522 blocks), `check-skill-facts`, `check-no-waitfor`, `check-conflict-markers`, `check-single-copy-deps`, `check-unused-deps`, `check-deno-pins`, `check-baselines-append-only` all pass.
- Both CI arms exercised locally: the runner package integration suite against a source-run toolshed, flag OFF and flag ON (`EXPERIMENTAL_SERVER_EXECUTION=true` on server + test process, per the CI recipe) — 14/14 each, ON-arm skip list confirmed empty (`tasks/server-execution-on-skips.ts runner` prints "no skips"), re-run after the review-fix guards.
- The traverse-replay golden for `toolshed-reload` is regenerated per the harness's stated procedure; the review structurally diffed it: `invocations` (result hashes) and `readSet` byte-identical, the only delta is scoped schema-tracker keys changing name→instance — exactly the intended semantic. The other two goldens are unchanged after decompression.
- §8 tripwire scan over the diff's additions: clean (one pre-existing local variable rewrapped).
- Two failures reproducible OUTSIDE the package's own `deno task test` are pre-existing at the stage-D base (verified identical at caaae1dad), not this PR's: plain `deno check` on `packages/runner/test/runner.test.ts` trips a lib.dom `Lock` name collision, and one storage test fails under bare `deno test --no-check` (Deno's disabled `__proto__` assignment). The package task (which CI runs) passes both.

## Flags

Recorded per the flag-don't-fill rule — conservative code landed, questions surfaced rather than answered locally:

1. **Scope-name identity keys OUTSIDE the nine-site closure.** key-vocabulary.md §1 declares the nine sites "the complete list at the time of writing", and §4 makes a tenth site a regression. Re-verifying anchors surfaced pre-enumeration (all predate 2026-08-02) scope-NAME-keyed structures in the same lifetime class as enumerated sites (per-manager/per-runtime/per-scheduler, i.e. cross-principal on a serving runtime): `storage/selector-tracker.ts:14` (`toKey`, the per-provider watch-selector tracker), `storage/v2.ts` `#docPullKicks` (`docKey`) and the `pending:` dedupe key, `runtime.ts` `missingDocLoadKicks` + `#pendingAsyncWork` owner keys (`addressKey`), `scheduler/wake-shaping.ts:344` (`linkKey`) and the `${scope}:${id}` `pieceId` it groups by, `traverse.ts` `sharedSchemaMemo` when a caller shares one across runs, and `traverse-recorder.ts:78` (capture tooling). Per-run/per-tx name keys (`v2-transaction.ts` doc keys, per-traverser memos, `diagnosis.ts` per-tx invariants) are sound under the run-is-one-instance model and were plainly excluded. Question for the owner: are the long-lived ones deliberate exclusions (client-only structures stage F will rework wholesale) or misses the nine-site closure should absorb before the tripwire is enforced against them?
2. **The two latent raw-scope key bugs, and how each was adjudicated** (see the neutrality section): (a) the schema-tracker coverage memo still does not apply to unscoped links — PRESERVED, because healing is value-visible through `traverseCells` consumers (the covered-skip's `null` replacement reaches the html reconciler's cell reads); enabling it is a separate deliberate change gated on resolving the reconciler's `get({traverseCells: true})` value consumption (so stated at the guard). (b) Unscoped result-pattern cache entries now DO evict on change notifications — HEALED, because the repo's contract test pins eviction and the writer side was the bug. The carve-out from strict byte-neutrality: the work-shape delta (idempotent re-prepare + a pull read per notification on unscoped result docs), plus ONE results-visible edge — a `stop()`ed child whose parent keeps re-firing the same keyless pattern now restarts regardless of scope, where base left unscoped ones stopped forever (base's own comments treat that as the bug) — and a cross-client pattern change is now re-materialized instead of silently stale-skipped. If the owner prefers strict preservation over the contract test, the eviction can be scope-gated in a follow-up — flagged rather than decided here beyond what the tests force.
3. **The executable spec model restates the key format un-encoded** (`packages/spec-model/server-execution/model.ts:34-37`: `userKey`/`sessionKey` template literals without encodeURIComponent). As a self-contained model this predates LD3 and diverges from the wire format for principals containing `:`. Should the model import the shared vocabulary, or is model-local simplification sanctioned?
4. **`toDirtyKey` and the wake path stay scope-NAME-keyed by design** (scopes.md §7 M4 — stage F re-keys push). The `canResolveScopeKey` guard in `refreshTrackedGraph` is the seam where name-keyed dirtiness meets instance-keyed tracking; when M4 lands, dirty keys carry instances and the guard's skip case disappears.

## Adversarial review (standing merge protocol)

An adversarial subagent review of the full diff ran after green, attacking OFF-arm neutrality, LD3/spec conformance, test vacuity, hidden couplings, and type trapdoors. It reproduced one BLOCKER and three further findings; all are addressed in the follow-up commit (`fix+test: address the stage-E adversarial review batch`):

- **Blocker (reproduced end to end): the healed coverage key activated the value-destroying covered-skip for repeat unscoped links** — `traverseCells` traversal of a doc linking one target twice returned `{a: "hello", b: null}` at the first commit vs `{a: "hello", b: "hello"}` at base, and the html reconciler consumes those values. Fixed with the explicit neutrality guard + regression test above.
- The result-pattern-cache eviction healing was flagged as an unargued live behavior change. Resolution went through both directions: a preservation guard was implemented first and BROKE the repo's base-green contract test pinning eviction-on-notification with normalized keys — the decisive evidence that the writer side (not the eviction) was the bug. The guard was reverted; the healing stands, argued and carved out explicitly (What/Flags above), with the contract test as its pin.
- The rewritten five-most-visited diagnostics test had lost the only assertion that would catch the coverage memo regressing; its original binding assertions are restored and now double as the guard's regression test.
- The wave's `actionScopeKey` was a bare `string` passthrough into the basis index — a format trapdoor against key-vocabulary.md §4; now typed `ScopeKey` (with `entityScopeKey` tightened symmetrically). One stale `resolveScopeKey` anchor in scopes.md §7 M1 refreshed.

The review verified as clean: the single-definition property (object-identity re-exports, no other format construction in runtime code), all nine sites' constructions and identity sources, identity agreement across every cross-matched coupling (entityKey ↔ pendingLoads, query `toDocKey` ↔ traversal `getTrackerKey` in the shared tracker, differential ↔ replica identity), the no-new-throws property for identity-less sessions, the honesty of the golden regeneration (structural diff: tracker-keys-only), and the non-vacuity of the new tests. Findings 6–7 (golden honesty note; the out-of-closure name-key inventory) are folded into Verification and Flags above.

A second verification pass by the same reviewer over the fix batch confirmed all five resolutions: the neutrality guard is the single predicate all three covered-skip sites route through (the original probe returns base behavior at HEAD; explicit-scope links keep base behavior), the eviction adjudication is accepted (the unchanged-pattern re-fire cannot commit differing bytes — `samePattern` holds for content-hashed keyless patterns, and the only write path is deepEqual-guarded — with the stopped-child restart edge named above), the restored diagnostics assertions bind again, the anchor sweep found zero stale `engine.ts:98` citations under docs/, and the `ScopeKey` typing holds end to end with no casts.

---
## Summary by cubic
Re-keyed all in-memory identity keys from scope NAME to per‑instance `ScopeKey` using the shared `scope_key` vocabulary in `@commonfabric/memory/v2`, landing Stage E OFF‑arm neutral and aligning client keys with storage rows. The vocabulary moved to `packages/memory/v2.ts`; engine re-exports; all keys now build from an explicit `ScopeKeyIdentity` (OFF arm: the runtime’s own session).

- **Refactors**
  - Moved the single `scope_key` definition (constructor, parse/inspect, `ProtocolError`) to `packages/memory/v2.ts` and re-exported from the engine; Stage E marked landed and docs/anchors refreshed; verification coverage now 295.
  - Instance-keyed strings at the nine sites and server query/watch: scheduler `entityKey` and indices, result-pattern keys, per-instance result cells, seed memo, schema-tracker/`getTrackerKey`, storage manager pending loads, differential/address strings, and graph snapshot; `SpaceScopeAndURI` now carries `ScopeKey`. Identity threads via `Runtime.scopeKeyIdentity` in the OFF arm.
  - Wave accumulator now takes `scopeKeyIdentity`; `actionScopeKey`/`entityScopeKey` typed as `ScopeKey`.
  - Spec model now encodes scope keys like the wire constructor and adds a conformance bridge test; `server-sync` derives tracked ids via `toDirtyKey`. State‑inspector imports `resolveScopeKey` from the wire module.
  - Added docs section enumerating the remaining name-keyed boundary surface for follow-ups.

- **Bug Fixes**
  - Fixed unscoped result‑pattern cache eviction by normalizing the writer key to match the delete path; eviction now triggers as contract-tested.
  - Preserved base behavior for unscoped-link coverage with an explicit guard; added regression tests. Query/watch keys now carry instance scope and reject name-scoped forms.

<sup>Written for commit 4ca39a858b7ab693aade7008d1dc614172a30c00. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5374?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seefeldb:
## Independent adversarial review (coordinator round) — LANDABLE-WITH-FIXES → fixes landed as d42054d6b (+ merge 394022b13)

Reviewed at 814b1e2c5 by a fresh-eyes adversarial agent, after and independent of the PR's own self-review round. Fix batch: d42054d6b (+ merge 394022b13); the stage-D fix/ruling commits are merged in (394022b13 merges the D branch at bc2e817a1; one further D→E merge for the Q1/Q2/Q4 ruling commits is sequenced next). Nothing merges — the train holds until the final phase exists.

**Verified deep and held (the code side is clean):** the LD3 single-definition property holds by object identity (engine re-exports, no twins); all nine sites construct from explicitly-supplied identity with NO ambient fallback (an identity-less session reproduces base's exact ProtocolError, just earlier — verified at the read path); OFF-arm neutrality survived three targeted mutation probes and an independent full-suite run (runner 1110, memory 454→455 post-merge, spec-model 18→19 with the new bridge test, state-inspector 88); both adjudicated neutrality carve-outs confirmed correct — the traverse coverage guard is the single predicate for all three covered-skip sites, and the healed result-pattern eviction is pinned by base's own contract test with exactly one named results-visible edge (a `stop()`ed unscoped child of a re-firing keyless pattern now restarts like explicit-scope children — base's stay-stopped was the bug by base's own comments); regenerated goldens byte-verified (values unchanged; exactly 2 tracker keys rewritten); the ON-arm CI mechanics verified (flag set on server AND test process; skip list empty).

**Findings → fixes:**

- **F1 (MAJOR)** — key-vocabulary.md claimed a closed nine-site inventory while the code (as the PR's own Flag 1 surfaced) contains name-keyed identity structures outside it. Per-site adjudication: ZERO sites should have been re-keyed in stage E — each exclusion is individually defensible — the defect was the document's closure claim. Fixed by scoping the closure to storage-row-aligned identity construction and adding a named **boundary-surface section**: every out-of-closure site with file:line, disposition, and its cardinality-&gt;1 failure mode (selector-tracker watch dedupe and the pull/load kick dedupes re-key WITH stage F's M4; wake-shaping budget buckets re-key with M1; the shared schema memo is recorded identity-BOUND — cross-identity sharing is value-bleed, tripwire owed at stage F; EngineObjectManager caches recorded sound-per-session; instance-safe transients listed with why). §4's tripwire now covers changes to listed boundary sites, and scopes.md §7 M2 names the remainder instead of claiming full resolution.
- **F2 (MINOR)** — stale anchors in docs this PR claimed re-verified, including one dead cite (scopes.md §7 M3 pointed past engine.ts's EOF) and systematic off-by-ones. All re-verified against the tree: ~30 anchor corrections — including the dead M3 cite and a framing correction the fixer earned: scopes.md §7's cites were valid MAIN-tree cites (the section inventories main's machinery), so §7 now carries an explicit main-vs-branch cite convention, with branch-live facts on branch anchors and main-only machinery on `main`-marked cites.
- **F3 (NIT)** — the inline dirty-key twin in memory/v2/server-sync.ts: fixed — `trackedIdsFromEntries` now calls `toDirtyKey` (clean same-package import); a further name-keyed twin found in passing (`cacheKeyForEntity`, server-sync.ts:16-20) is recorded in the boundary section's M4 entry.
- **Flag 3 (model soundness, found by this review)** — the spec model's un-encoded `sessionKey` was NON-INJECTIVE (`sessionKey("a:b","c") === sessionKey("a","b:c")`) — a hole a naive conformance test over simple ids could never catch, and principals are DIDs (colon-bearing). Resolution: the real wire format was PROVEN injective (`encodeScopeKeyPart = encodeURIComponent`, packages/memory/v2.ts:83 — the colliding pair yields `session:a%3Ab:c` ≠ `session:a:b%3Ac`), and the model now restates that exact scheme import-free. The model stays import-free; a conformance bridge TEST asserts byte-agreement between the model's constructors and the shared `resolveScopeKey` over a colon/slash-bearing corpus (spec-model 19/19).

**Owner rulings recorded on the #5371 ledger also apply here** (Q1(c) recompute-by-dependency, Q2 per-run class, Q4 stamped-seals-only) — they are landing on the D branch now; this branch absorbs them in the sequenced follow-up merge.

**Open items for stage F carried from this review:** the boundary-surface re-keys (M4/M1 clusters), the schema-memo identity-bound tripwire, and the coverage-memo enablement gate (recorded in the boundary section; enabling the memo for unscoped links is gated on resolving the html reconciler's `get({traverseCells: true})` value consumption).

seefeldb (commented):
Verdict: changes required.

[major] `packages/memory/v2.ts:162-170` accepts non-canonical “well-formed” scope keys such as values containing `/` or malformed percent escapes. Stage F's explicit-instance query surface then places the accepted key into slash-delimited tracker keys and decodes it, so a value admitted by the shared validator can corrupt addressing or throw in `query.ts`.

Make the shared validator enforce the canonical round-trippable grammar (including percent-decoding and delimiter rejection) and add malformed/canonical round-trip tests. The current check rollup also has Test 2/6, 4/6, 6/6, Pattern Integration 3/4, and Status failures.

seefeldb:
Fix ledger for the 2026-08-12 review (non-canonical scope keys):

**Fixed in a92aaadd5.** `isScopeKey` now accepts exactly the constructor's image: every segment must satisfy `seg === encodeScopeKeyPart(decodeScopeKeyPart(seg))` and be non-empty. Malformed or non-canonical percent escapes, raw `/` or `:`, and lone surrogates refuse (return false) — never throw. Canonical escape casing (`%2F`) is pinned by test, and over-escapes (`%41` for `A`) plus raw `+`/space are rejected by construction of the fixed-point grammar.

Admission is gated where explicit keys arrive: derived-commit scope-key annotations refuse with a `ProtocolError` (not a phantom wave conflict), and wave basis-instance keys are validated before they reach `scheduler_basis` addressing; nothing is applied on refusal.

Call-site audit: the only prior caller was the internal `fromDocKey` (canonical inputs; no behavior change). All raw-key ingress on this branch is derived-class-only behind `EXPERIMENTAL_SERVER_EXECUTION` (default off), and the sole producer builds keys canonically — so no pre-existing non-canonical rows are constructible (code-verified).

Spec-model: gains an import-free `isCanonicalKey`; the conformance bridge now asserts pointwise agreement with the real validator on acceptance and on a 24-entry rejection corpus.

Evidence: red-first (4 new tests failed pre-fix); discrimination probe (loosening the validator re-fails the same 4). Suites: memory 431, spec-model 19, runner E-touched files 60 files / 495 steps — 0 failures; check/fmt/lint clean. key-vocabulary.md grammar/gate anchors updated.

Flagged, not filled: `principalOfSessionKey` remains a tolerant extractor (session-id composites are outside this validator's scope).

seefeldb:

PR #5377, opened 2026-08-05
refactor: one walk replaces cells, on both sides of the connection

danfuzz:
Two functions recursed a value replacing cell-shaped things with
reference-shaped ones:

| | `convertCellsToLinks` (runtime) | `CellHandle.serialize` (client) |
| --- | --- | --- |
| recognizes | a `Cell` | a `CellHandle` |
| answers with | the link that reaches it | the ref that names it |
| a value met twice | a link to where it sits | **recursed until the stack gave out** |

Same recursion, same container rebuilding, same question about a value met
twice — and only one of them handled it.

## What changes

`replacingWalk` owns the recursion. Each caller says only what distinguishes
it:

- `replace` — which values stand for something else, and which are refused
- `enter` — what a container becomes on the way in
- `cycle` — what a value the walk is already inside looks like

The two implementations of `cycle` are the interesting part, and they
legitimately differ. The runtime writes a link to where the value sits, which is what makes a
cyclic value representable at all. The client **refuses**: a `CellRef` has
nothing a cycle can be written as, and the far end rebuilds what it receives, so
naming the problem beats exhausting the stack.

It lives in `shared.ts` — where the client and the runtime already meet — and
imports nothing heavier than `isRecord`.

## Evidence it is otherwise behavior-preserving

Running the new client tests against the hand-rolled walk, **only the cycle
case fails**. `converts a handle into the ref that names it`, `converts a handle
reachable at two positions at both of them`, and `flattens a native object to
its enumerable members` all pass on both sides.

That last one pins a wart rather than endorsing it: a `Date` is an object, so it
is descended into and reaches the far end as `{}`. Unchanged here, and now
visible.

`packages/runner` 1149 / 0 · `packages/runtime-client` 53 / 0 ·
`packages/html` 43 / 0 · full workspace passing · `deno fmt --check`,
`deno lint`, `check-docs`, `check-skill-facts`, `check-unused-deps`,
`check-single-copy-deps` green.

## Tests

`replacing-walk.test.ts` covers the walk directly, including the two properties
that are easy to get wrong and were: that a value reachable at two sibling
positions is walked at both rather than answered as a cycle, and that the
ancestor is cleared on **every** exit — when `enter` stops the descent, and when
`enter` throws.

## Not addressed

The walk answers a shared subtree once per position, so a balanced DAG costs
O(paths). That is inherited, not introduced, and the same trade recorded on
#5355.

---
## Summary by cubic
Unifies traversal and replacement on both sides via `replacingWalk`. The walk is generic, descends only into plain objects/arrays by shape, and leaves special/native objects whole; the client rejects cycles and fixes a symbol-error edge case.

- **Refactors**
  - Added `replacingWalk` (hooks: `replace`, `enter`, `cycle`), generic over input/output; descends only into plain objects/arrays via a structural check (`isPlainObject`/`Array.isArray`); stands non-plain objects whole; re-exported from `@commonfabric/runner/shared`.
  - Runtime `convertCellsToLinks` uses the walk; cycles become link refs to their path; preserves schema-cleaning and `FabricPrimitive`/`FabricInstance` handling; carries CFC label view when requested.
  - Client `CellHandle.serialize` uses the walk; replaces handles with their `CellRef`; accepts only JSON-serializable types; throws on cycles (with path); stabilizes symbol errors via explicit `String()`; stands native objects like `Date` whole instead of flattening.

- **Tests**
  - New `replacing-walk.test.ts` covering replacement, cycles vs shared refs, ancestor clearing (including when `enter` stops/throws), and `enter` behavior.
  - Updated sharing test to use `FabricBytes` conversion; client tests add cycle rejection with path, unsupported types (including symbols), and `Date` preserved as a `Date`.

<sup>Written for commit f9527f2b709968d4763656765429dec3e3e1b775. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5377?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


danfuzz:
Defunct. The merge of walkers turned out to be ill-advised.

PR #5378, opened 2026-08-05
docs: a unit test coding style document, and the spelling convention

danfuzz:
Two additions to the documentation about how this repository writes things
down. Neither existed before, and both were prompted by drift that nothing was
there to catch.

## 1. A unit test coding style document

Adds `docs/development/unit-test-coding-style.md`. There was no document
saying how a unit test file should be shaped, and the one concrete example
that existed taught the wrong shape.

- **Location and name.** `test/` under the package, mirroring the `src/`
  subdirectory structure; `<topic>.test.ts` with the three choices for
  `<topic>`. This describes what the tree already mostly does — 1023 test
  files sit under a `test/` directory, and `packages/data-model/test/` already
  mirrors its `src/` subdirectories.
- **Structure.** One top-level `describe()` titled with the name of the file
  under test, with a carve-out for a file that parameterizes one test body
  across a fixed set of implementations. Nested `describe()` blocks rather
  than comment banners, because a `describe()` title shows up in the run
  transcript and a comment does not.
- **Class and function shape.** Literal `constructor()`, `instance members`,
  and `static members` blocks under a class's `describe()`, with a worked
  example that compiles.
- **Description strings.** An `it()` is a third-person verb phrase completing
  the word "it", naming an observable output rather than an intent —
  "returns `false` for a bare function", not "rejects a bare function". Where
  the fact is a return value the verb is "returns", specifically not
  "answers". Markdown markup, with the exception for a `describe()` whose
  title would otherwise be entirely one backticked string. Description strings
  are exempt from the 80-column line width.
- **Assertions.** `expect()` over `assert*()`, with the reason (failure output
  that prints actual against expected and structurally diffs composites) and
  a case-by-case list. Plain `assert()` stays the right tool where truthiness
  itself is the assertion.
- **Which comparison decides the outcome?** The section to read before writing
  a test that turns on a fine distinction, split into hazards general to
  JavaScript and hazards specific to this repository.

### The matcher hazards

General to JavaScript:

- `expect(-0).toEqual(0)` passes, as do the array and `Set` forms. Only
  `toBe()` discriminates the sign of zero, so any `-0` assertion written with
  `toEqual` is vacuous.
- `expect(NaN).toBe(NaN)` passes because `Object.is` says so, whatever the code
  under test decided. A test meaning to pin "this implementation treats all
  `NaN`s as equal" has to route through the implementation's comparison and
  assert on that result.

Specific to this repository:

- `hashOf()` returns a boxed `FabricHash`, so `toBe` compares identity rather
  than hash value and two hashes of equal values compare unequal. Compare
  `.bytes` or `.hashString`. `hashStringOf()` returns a plain string, which
  `toBe` compares by value.

The shared shape is stated once: an operator that is correct for some
comparison, just not the one that was meant. The mechanical check is to
confirm the test fails against a deliberately wrong expectation before
finalizing it.

### The example that was wrong

`TESTING.md`'s Test Structure section carried the repository's only concrete
unit test example:

```typescript
describe("Feature", () => {
  it("should do something", () => {
```

Neither line demonstrates the convention. It is now a compiling `deep-freeze`
example in the right shape, followed by a pointer to the new document.

### Where it is reachable from

Indexed in `docs/development/README.md` under Testing, added to `TESTING.md`'s
Related documentation list, and cited from `.claude/rules/tests.md`, which
loads whenever a `*.test.ts` file is read. `AGENTS.md` gets a line too, since
the rules mechanism is Claude Code's and Codex composes `AGENTS.md` instead.

## 2. Which spelling of English the repository uses

Nothing anywhere said, and Britishisms have been accumulating. Adds a short
`## Spelling` section to `docs/README.md` — the rule, one sentence of
rationale, and two carve-outs.

The rule is descriptive rather than imposed. Counting across `packages/` and
`docs/`, excluding `docs/history/`:

| | British | American |
| --- | ---: | ---: |
| `behaviour` / `behavior` | 37 | 901 |
| `colour` / `color` | 106 | 4999 |
| `centre` / `center` | 7 | 1461 |
| `analyse` / `analyze` | 2 | 357 |
| `serialise` / `serialize` | 1 | 176 |
| `grey` / `gray` | 21 | 435 |
| `defence` / `defense` | 7 | 60 |
| `recognise` / `recognize` | 5 | 74 |

The convention already holds by a wide margin everywhere it is uncontested,
so writing it down keeps it that way rather than establishing something new.
The section frames it as standardization rather than a claim about which
English is better.

Two carve-outs:

- **Quoted external material.** Dependency names, messages relayed from other
  systems, specification wording, and data-file contents keep the spelling
  they arrived with. `grey` falls here as well as under the rule: both
  spellings are legal CSS, and `packages/patterns/scrabble/scrabble-words.ts`
  carries `"GREY"` as data.
- **Identifiers already established here.** `cancelled` is the one word where
  the British form is in the majority — 237 against 61 — and it is
  overwhelmingly identifiers rather than prose: `isCancelled`,
  `bootstrapQueryCancelled`, `isRunCancelled`, exported names among them.
  Changing that is a rename with real blast radius, not a spelling fix.

This half is deliberately a note and not a sweep. No existing spelling is
changed here.

## Verification

`deno fmt --check`, `deno task check-docs` (all 526 blocks),
`deno task check-skill-facts`, `deno task check-conflict-markers`, and
`deno task docs-links --orphan` all pass. Before trusting the green from
`check-docs`, I broke one of the new code blocks and confirmed it went red,
which is what establishes that the checker walks the new file at all.
Documentation only — no code changed, so no test suite is implicated.

https://claude.ai/code/session_0154rxhW7tcMxQ8QLvHSxbCZ

---
## Summary by cubic
Adds a unit test coding style guide and documents the repository’s American spelling convention. Updates testing docs and rules to point to the guide and clarify that pattern tests follow a different form.

- **New Features**
  - Added `docs/development/unit-test-coding-style.md` covering: test file location/name, single top-level `describe()`, `it()` phrasing, `expect()` over `assert*()`, and matcher traps (`-0`, `NaN`, boxed `FabricHash`). Clarifies the guide does not apply to pattern tests and points to `docs/common/workflows/pattern-testing.md` (prefer `assert()` over `computed()`).
  - Updated `docs/development/TESTING.md` with a correct, compiling example and a link to the new guide.
  - Scoped `.claude/rules/tests.md` to distinguish pattern tests and route them to the pattern testing guide; summarized core unit test rules.
  - Linked the guide from `docs/development/README.md` and `AGENTS.md`.
  - Added a “Spelling” section to `docs/README.md` standardizing American spellings, with carve‑outs for quoted external material and established identifiers like `cancelled`.

<sup>Written for commit 03049b6d3bd8385aa8d8d58a0315fefc95b5c9c8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5378?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Hixie (approved):
LGTM

mpsalisbury (approved):
LGTM

PR #5384, opened 2026-08-05
docs: a code comment style document

danfuzz:
Companion to the unit test coding style document: the house rules for comments,
written down where the repository can point at them.

`docs/development/code-comment-style.md` covers both kinds — the `//` comment
that explains local mechanics, and the JSDoc comment that states a contract.

**What earns a comment**, and the two things that beat one: a well-named helper
or identifier, which cannot drift out of agreement with what it describes, and
a document under `docs/features/` that the comment names, for reasoning that
outgrows what a comment can hold.

**The rule that a comment describes the system as it stands**, which generates
most of the rest. The shapes it rules out, each of which reads as helpful in
the moment it is written:

- the code's own past ("preserving prior behavior", "no longer coerced")
- the road not taken, which is a record of a decision rather than of the code
- a neighboring module's failings, which are a bug report nobody filed
- a survey of the rest of the codebase ("the only caller", "three such sites"),
  which is one refactor away from false and will fail no check
- a rollout plan ("Stage 3"), which has no referent once the plan is over

Because self-review does not reliably catch any of them, that section ends with
a grep over the added comment lines of a diff instead of a re-read.

**The mechanical rules.** `TODO` tags, which `ban-untagged-todo` enforces on
every `deno lint` run — and why the issue reference that rule also accepts is
the wrong choice here. What happens to a comment when the code it annotates is
deleted, which is where the most misleading comment of all comes from: a
premise that is well written, correctly placed, and about nothing. The Markdown
markup that comments, error messages, and log messages share. The
`//`-bracketed section marker, and the ban on horizontal rules.

**What a doc comment carries**: what, why, and the bound on any guarantee it
claims — a reader takes "catches drift" to mean "catches drift", and the
unstated half is where a check that works gets trusted past its range. Plus the
opening forms for a helper, a variation on a theme, a constructor, and an
override.

## Wiring

Indexed in `docs/development/README.md` and named from `AGENTS.md`, which is
the only file-type-aware channel Codex has. `docs/README.md` keeps its short
"Comments in code" framing and points here for the rest; `DEVELOPMENT.md` gains
a "Comments" subsection under its style rules and its JSDoc bullet now points
at the guide.

No `.claude/rules/` file. One would need to be scoped to `**/*.ts` and
`**/*.tsx`, and `.claude/rules/README.md` asks that a rule apply to a
recognizable set of files rather than to the repository as a whole. That glob
is the repository.

## Notes

The tree does not conform, and the document does not pretend otherwise: it
opens with the same carve-out the test style guide uses — a neighboring file is
not evidence of what to write, new code follows the guide closely, an edit
conforms the area around it, and converting a whole file is its own change.
Nothing existing was changed here.

`deno task check-docs` was confirmed to actually walk the new file by watching
two of its blocks fail before trusting the green. `check-skill-facts`,
`check-conflict-markers`, `docs-links --orphan`, and `fmt --check` all pass;
the 80-column pass was manual, since `deno fmt` excludes `docs/`.

## Also: examples in documentation

`docs/README.md` gains a section saying that an example illustrating a general
rule is invented rather than lifted from this repository. Real code moves, and
nothing about moving it brings anyone back to the document that quoted it;
while it sits there it answers searches for the identifier it names with a hit
that is not a use of it, which is the argument the spelling section already
makes, aimed elsewhere. The exception is an example documenting a specific
hazard, where the hazard is the real thing and the function has to be named.

The new comment guide follows its own advice — every example in it is a donut.

## Also: a follow-on to the unit test guide

Two additions that ride along here rather than in their own PR.

A doc comment on a type the schema generator sees is not only documentation:
it attaches to the generated schema as its `description`, and a `#hashtag` in
it is mirrored into a `tags` array. So the comment is program output, and a
hashtag written as an aside becomes a tag whether or not one was meant. The
guide says so and points at section 12 of the type mapping spec.

And the unit test guide's `test/`-mirroring-`src/` rule is about a package of
interdependent source files. A directory of independent components has no
shared structure for a parallel tree to mirror, so each test sits beside its
component — `packages/ui` and `packages/patterns` both — and the guide now
says that arrangement is correct rather than merely tolerated.
`.claude/rules/tests.md` carries the same exception, since it loads on any test
file and placement is decided before anything else in it applies.

---
## Summary by cubic
Adds a code comment style guide and makes it the single source of truth for `//` and JSDoc comments. Tightens `TODO` tagging, standardizes example policy, clarifies test placement, and documents how doc comments flow into generated schema; top-level docs now link to the guide.

- New Features
  - Added `docs/development/code-comment-style.md`: what earns a comment; present-tense-only rule (also for test descriptions); required `TODO(<owner>):` enforced by `deno lint` (`ban-untagged-todo`); remove comments with removed code; shared Markdown rules (also for error/log messages); section markers; doc comment essentials and standard openers; and how doc comments reaching the schema generator become program output (`description`, `#hashtag` mirrored to `tags`).
  - Consolidated rules: `docs/README.md` trims comment guidance and links to the guide (adds “Examples in documentation” with invented donut/glaze examples and the hazard-only exception; notes `deno task check-docs` catches compile drift only); `docs/development/DEVELOPMENT.md` adds a “Comments” section and links error/log message markup; `docs/development/README.md` indexes the guide; `AGENTS.md` points to it.
  - Clarified test placement: tests mirror `src/` by default; for independent components (e.g., `packages/ui`, `packages/patterns`), keep tests beside the component. Reflected in `docs/development/unit-test-coding-style.md` and `.claude/rules/tests.md`.

<sup>Written for commit 330c4098ba2a179014fa2aa18217657f4179ef9f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5384?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


docs/development/code-comment-style.md
```diff
@@ -0,0 +1,341 @@
...
+// Right: what a caller needs, and nothing about how it came to be.
+/**
+ * Freezes `value` if it is not already frozen, and returns it either way.
+ */
+function normalize(value: object): object {
+  return Object.isFrozen(value) ? value : Object.freeze(value);
+}
+```
+
+This applies to test files too. A test added in response to a bug reads like
```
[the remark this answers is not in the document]
danfuzz: Valid, and the ambiguity was in the intro rather than in this paragraph. The anti-past rule does reach an `it()` description string — a test written after a bug states the behavior it pins, the same as any other test — so the intro no longer claims the description strings take nothing but the markup from this guide, and this paragraph now says outright that it covers the strings as well as the comments.

Fixed in 83ad223e8.

— Claude (Opus 5), on @danfuzz's behalf

docs/development/code-comment-style.md
```diff
@@ -0,0 +1,361 @@
...
+are recorded.
+
+"Being total is also what keeps it honest, since the alternative is collapsing
+an unknown flavor into a catch-all" is a sentence about a discussion. "Every
+flavor maps to a distinct key; a collision here would silently merge two
+orders" is a sentence about the code.
+
+### Not the neighbors
+
+A comment says what the code it sits next to does. Whatever is true of another
```
Hixie, line 109: this first sentence sort of contradicts an earlier one, dunno if agents will care

docs/development/code-comment-style.md
```diff
@@ -0,0 +1,361 @@
...
+
+### The check that finds these
+
+Self-review does not reliably catch any of the above, because each shape reads
+as helpful in the moment it is written. Grep the added comment lines of the diff
+instead:
+
+```text
+before|used to|previously|no longer|formerly|the alternative|rather than the old
+```
```
Hixie, line 174: i would soften this a little. i don't want agents to think that that regexp is canonical. it's just a starting point. maybe tell agents here to update itthis when they find examples that could be caught?

docs/development/code-comment-style.md
danfuzz: Valid — it contradicts "what earns a comment", which says the code already says what it does and a comment carries what the code cannot. The sentence was about locality, not restatement, so it now reads "a comment is _about_ the code it sits next to".

Fixed in #5433.

— Claude (Opus 5), on @danfuzz's behalf

docs/development/code-comment-style.md
danfuzz: Agreed, and the section said nothing about the list's standing, which is what let it read as canonical. It now says outright that the list is a starting point weighted toward the past-tense shapes, that nothing in it would catch a rollout stage, a count of callers, or a complaint about a neighbor, and that the sections above are the rule. It closes by asking that a phrasing which gets past the list be added to it.

Fixed in #5433.

— Claude (Opus 5), on @danfuzz's behalf

PR #5388, opened 2026-08-05
fix(runner): separate a release's authority from a stop's

Hixie:
The runner keeps one registration per running result: the scheduler actions for that piece, indexed by the result's key. A pattern that launches a child registers a *release* for it, which runs when the launching pattern is torn down. An explicit *stop* ends a piece on its owner's behalf.

docs/specs/runner-child-run-ownership.md already governs the difference, and "Independent lifetimes" already states the rule this change implements: "A start that is still resolving may be about to acquire that lifetime and cannot say so yet, so a release declines while one is in flight." The runner honored that only for a start it could see. A start that has not yet followed its link to the result is indexed under no key the release consults, and the shared release-and-stop path recorded a stop tombstone on every such attempt, which then terminated when it resolved. A result being opened directly — navigated to as a page — while its parent released it was silently never started. A start still resolving a link is a start in flight; the spec draws no distinction between the ones a release happens to see and the ones it does not.

Fixing that exposed a second defect of the same family. A start that lost a race with a stop still reported success and recorded an independent lifetime for the result. The registration under the key at that point belonged to whatever replaced it, or to nothing. The stale lifetime record then made every later release of that result decline, so a list element or nested child outlived its parent until an explicit stop.

Split the shared path. stop() records the pre-resolution tombstones itself and then calls stopResult(), which carries the rest of the teardown; releaseChild() calls stopResult() directly, so a release ends the registration it recognizes and leaves unresolved starts to settle on their own. start() claims the independent lifetime — and reports true — only when its target is still current as the attempt settles: no stop tombstoned the target or moved its registration's generation since this attempt discovered it. That judgment is scoped to the target doc alone, because a stop of an intermediate link doc the start resolved through does not touch the target's registration.

The specification had a hole, which is why one defect could sit under a rule and the other had no rule at all. It stated the release-side constraints case by case — release the exact registration, decline against an independent lifetime — without ever naming what those cases have in common. So it had nothing to say when the same question arose somewhere it had not enumerated.

Fill it by naming the distinction the cases come from, in a new section, "Two ways a registration ends": a stop is authoritative over the result, while a release is authoritative over the registration its own launch installed and over nothing else. The existing sections then read as that distinction applied to a case rather than as separate rules that happen to agree, and two questions the document could not previously answer follow from it. A start whose target a stop superseded claims nothing, because the registration it would claim is not the one it installed. A transaction rollback carries a release's authority, not a stop's, for the same reason — so the nested action-result child rollback, which used a full stop and therefore also tombstoned unresolved starts, now releases like the run() launch rollback beside it always did.

Tests cover each rule. Releasing a live child registration while a link start resolves toward the same target leaves no tombstone and the start completes. A stop that wins against a redundant start leaves that start claiming nothing for the replacement that follows. A stop of the link doc between the target's installation and the start's settlement leaves the target's lifetime claimed. Each test was run against the previous behavior and fails at the assertion naming its bug.

One case stays wider than the distinction demands, marked with a TODO where it happens. A release cancels every pending commit-gated start held for the result, including one another launch scheduled, because a start that has installed no registration yet offers nothing to tell launches apart by. Narrowing it needs the release to name the pending start its own launch created, the way it already names the registration it installed. The specification states the width as it stands.

---
## Summary by cubic
Separates a release’s authority from a stop’s: a release ends only what it launched (its registration or a pending commit‑gated start); a stop remains authoritative over the result. Fixes lost starts, stale lifetimes, and leaked commit‑gated operations; aligns code and spec.

- **Bug Fixes**
  - `releaseChild()` no longer tombstones unresolved link starts; it ends only the registration it installed. When no registration exists, it cancels all pending commit‑gated starts for the result so a released child doesn’t start after commit.
  - `start()` returns true only if it claims a lifetime at settle and the target is still current; returns false when superseded by a stop or a generation change. Scoped to the target; stops of intermediate link docs don’t void the claim.
  - Transaction rollback now releases the child registration instead of stopping it, preserving unrelated lifetimes and in‑flight starts.
  - On failed transactions, pending commit‑gated starts and commit‑gated pattern runs settle as cancelled and are removed from the pending index.
  - Spec adds “Two ways a registration ends.” Tests cover release‑vs‑pre‑resolution starts, stop‑vs‑redundant start, link‑doc stops after target install, release of a commit‑gated child, and commit‑gated failures; tagged the pending‑start TODO to satisfy linting.

<sup>Written for commit 43bee987b4940496418bf244bf54eb06a75c7581. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5388?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/runner-child-run-ownership.md
+++ b/docs/specs/runner-child-run-ownership.md
@@ -105,5 +105,8 @@ A start gated on a commit has no registration until its callback installs one.
 Ownership of it begins when the start is scheduled, so stopping the result
 before that commit tombstones the pending start and the callback does not
 install it. Without this a piece the user stopped starts anyway, moments
-later. Releasing the result's registration cancels the pending starts held
-for that result too, at the width the exception above describes.
+later. Releasing the child cancels the pending starts held for that result
+too, at the width the exception above describes. A launch whose child is
+still gated holds that child through the pending start and nothing else, so a
+release that passed it over would let the child start after the pattern that
+launched it is gone.
--- a/docs/specs/runner-child-run-ownership.md
+++ b/docs/specs/runner-child-run-ownership.md
@@ -14,7 +14,29 @@ The registration is not part of any transaction. The writes a setup stages
 roll back when its transaction fails; the registration does not. A result is
 also reachable from more than one place: a nested pattern inside its parent, a
 list element inside its coordinator, and a page the user navigated to can all
-name the same result. These two facts decide the four rules below.
+name the same result. These two facts decide the rules below.
+
+## Two ways a registration ends
+
+Two operations end a registration, and they carry different authority.
+
+An explicit *stop* is authoritative over the result. It ends whatever
+registration it finds, discards any lifetime the result holds, and reaches
+starts that hold no registration yet: a start still resolving a link to this
+result terminates when it resolves.
+
+A *release* is authoritative over the registration its own launch installed,
+and over nothing else. A pattern releases a child it launched when that
+pattern is torn down, and a launch releases its own registration when the
+transaction staging its setup does not become durable. A release leaves a
+replacement registration alone, leaves a result that holds a lifetime of its
+own running, and leaves a start it cannot see to settle on its own terms.
+Pending commit-gated starts are the exception: a release cancels every one
+held for the result, including one another launch scheduled, because a start
+that has installed no registration yet offers nothing to tell launches apart
+by.
+
+Each rule below is one of these two authorities applied to a case.
 
 ## Releasing a child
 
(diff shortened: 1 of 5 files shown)


Hixie:
Follow-up for https://github.com/commontoolsinc/labs/pull/5358

seefeldb (approved):

PR #5395, opened 2026-08-05
feat(cf-harness): say what caused each gateway request

Hixie:
The LLM gateway authenticates upstream with one shared credential per provider and is reached from one address, so a request arriving there says nothing about which workload produced it. A test, an unattended dispatch, and a person at a terminal are indistinguishable, and cf-harness is the overwhelming majority of the traffic.

This makes every request the gateway client sends describe itself. A new src/provenance.ts resolves, once per process, what is invoking the harness: the invoker class, the coding-agent session it is running inside, the subcommand, a per-process session, a principal for the machine, the continuous-integration run, the Loom dispatch class, and the service that launched it. The values travel as `x-cf-harness-*` headers and, condensed, inside the User-Agent, which is the copy that works against the gateway as deployed: its Envoy access log records the user agent of every request and nothing else that identifies a caller, and that log reaches Cloud Logging.

Every signal comes from something that already exists, so nothing has to be declared and nothing has to be remembered at the call site. Loom passes a run manifest, which names both the invoker and the dispatch class. Claude Code exports CLAUDECODE and the Codex CLI exports CODEX_SANDBOX, the latter already read for the same purpose by start-local-dev.sh. A service names itself for tracing through OTEL_SERVICE_NAME, and since every process it spawns inherits that variable, a harness a service launches reports both the service invoker and the service's name without knowing cf-harness exists. start-local-dev.sh now sets that name for toolshed and the background piece service, so any future path from either to the harness is attributed from the day it exists.

The coding-agent field is orthogonal to the invoker, so a Loom dispatch started inside a Claude Code session reports both, and an absent agent says no coding agent is involved, which is what a person at a shell or a locally launched service reports.

Because these headers reach the upstream provider, what a value may contain is bounded twice over: no request content, and nothing that identifies a single run or person. The subcommand is mapped through a closed set, so an unrecognized argument reports "prompt". Every value passes through a sanitizer, at the one point where values leave, that constrains it to a short run of characters that cannot carry prose or break a header. A Loom manifest contributes only its dispatch class. A coding agent's marker variable contributes only which agent was found. A test drives a real prompt through the client and asserts no word of it appears in any header.

The principal is generated on first use and kept in $CF_HARNESS_HOME/principal, so nothing about the machine determines it, and no filesystem path or git metadata contributes to any field. docs/features/gateway-request-provenance.md states the invariants behind both, and the two known gaps in coverage: subagents share their parent's client and so report the parent's session, and a principal covers one harness home. docs/development/CONFIGURATION.md gains the cf-harness section it had been missing.

The Codex transport is deliberately untouched: it addresses a different endpoint under a personal subscription, and already sends its own originator and session headers.

---
## Summary by cubic
Adds request provenance to all `cf-harness` gateway calls so traffic can be attributed by workload in access logs. Each request now sends `x-cf-harness-*` headers and a condensed User-Agent; a new `whoami` command shows what this host reports.

- **New Features**
  - Provenance resolver with: invoker, command, principal, session, ci, dispatch, agent, service. Auto-detected from env (`OTEL_SERVICE_NAME`, `CLAUDECODE`, `CODEX_SANDBOX`, `ENV=test`, `CF_HARNESS_INTEGRATION`, `CI`/`GITHUB_*`). Principal is random, stored in `$CF_HARNESS_HOME/principal`.
  - Gateway client attaches provenance headers and User-Agent on every request (session shortened in UA); `openai-codex` transport unchanged.
  - New `whoami` command (supports `--json`) prints the principal and exact User-Agent; command name comes from a closed set (`auth`, `models`, `whoami`; else `prompt`).
  - Privacy: no request content or personal identifiers are sent; values are sanitized and bounded; Loom manifests only add the dispatch class. Docs updated (`docs/features/gateway-request-provenance.md`, configuration, `packages/cf-harness/README.md`); `start-local-dev.sh` sets `OTEL_SERVICE_NAME` for toolshed and the background service.

<sup>Written for commit 2756958d0915e6d5d9337044636f5df838ea8f59. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5395?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/cf-harness/src/provenance.ts
```diff
@@ -0,0 +1,276 @@
...
+  session?: string;
+  principalStore?: PrincipalStore;
+}
+
+/** Where the harness keeps its state, matching the CLI's own resolution. */
+function defaultHarnessHome(env: EnvReader): string | undefined {
+  const declared = env("CF_HARNESS_HOME");
+  if (declared !== undefined && declared !== "") return declared;
+  const home = env("HOME");
+  return home !== undefined && home !== "" ? `${home}/.cf-harness` : undefined;
```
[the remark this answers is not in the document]
Hixie: Claude responds:

Not taking the cwd fallback, though the inconsistency it points at was real.

Falling back to the working directory would write `.cf-harness/principal` into
whatever directory a run happened to start in — including this repository during
a test run, which is the failure the `isolated-test-processes` guidance exists to
prevent. A host with no `HOME` already reports `unstable-<label>`, which says
plainly that the value covers one process rather than grouping silently under a
principal that moves with the working directory.

What was genuinely wrong was the comment above the function, which claimed the
resolution matched the CLI's when the CLI also falls back to cwd. That claim is
gone; the comment now states what the function does: `CF_HARNESS_HOME`, or
`.cf-harness` under `HOME`, and undefined when neither is set.

The other eight findings from this run are fixed in 511b3dc04.

packages/cf-harness/src/provenance.ts
```diff
@@ -0,0 +1,276 @@
...
+
+const MAX_VALUE_LENGTH = 48;
+const UNSAFE = /[^A-Za-z0-9._:-]+/g;
+
+/**
+ * Reduce a value to a short run of characters that cannot carry prose. Applied
+ * to everything, including values that should already be safe.
+ */
+export function sanitize(value: string): string {
+  const cleaned = value.replace(UNSAFE, "_").replace(/^_+|_+$/g, "");
```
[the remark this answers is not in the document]
mathpirate: **[P1] This remains unresolved at the current head.** Replacing delimiters with `_` makes a value header-safe, but does not stop it from carrying readable content—the test suite itself demonstrates that `"what is the patient's diagnosis?"` survives as `"what_is_the_patient_s_diagnosis"`. Arbitrary strings can enter through `CF_HARNESS_PRINCIPAL`, `OTEL_SERVICE_NAME`, `GITHUB_WORKFLOW`, the Loom `dispatchClass`, persisted principal content, and injected `HarnessProvenance`, and all of those values are sent to the upstream provider. Please give categorical fields closed/validated representations and reject, omit, or irreversibly hash arbitrary values before emission; truncation and delimiter replacement are not sufficient to uphold the stated privacy invariant.

Sol on behalf of Gideon

mathpirate (commented):
Four additional inline findings follow.

Sol on behalf of Gideon

packages/cf-harness/src/provenance.ts
```diff
@@ -0,0 +1,280 @@
...
+  if (nonEmpty(env("CLAUDECODE"))) return "claude-code";
+  if (nonEmpty(env("CODEX_SANDBOX"))) return "codex";
+  return undefined;
+}
+
+function detectCi(env: EnvReader): string | undefined {
+  const workflow = env("GITHUB_WORKFLOW");
+  const runId = env("GITHUB_RUN_ID");
+  if (workflow !== undefined || runId !== undefined) {
+    return `github:${workflow ?? "unknown"}:${runId ?? "unknown"}`;
```
mathpirate: **[P2] This emits an identifier for exactly one CI execution despite the documented invariant that no value identifies a single run.** `GITHUB_RUN_ID` maps to an existing GitHub Actions run and is sent through both the headers and User-Agent to the upstream provider. That is materially different from the locally generated opaque session value. Could we omit the run ID, replace it with a coarse category or non-linkable digest, or explicitly revise the privacy contract if external linkability is intentional?

Sol on behalf of Gideon

packages/cf-harness/src/provenance.ts
```diff
@@ -0,0 +1,280 @@
...
+ */
+function detectPrincipal(env: EnvReader, store?: PrincipalStore): string {
+  const declared = env("CF_HARNESS_PRINCIPAL");
+  if (declared !== undefined && declared !== "") return declared;
+  if (store === undefined) return `unstable-${randomLabel()}`;
+  const stored = store.read();
+  if (stored !== undefined) return stored;
+  const created = randomLabel();
+  store.write(created);
+  return store.read() ?? `unstable-${created}`;
```
mathpirate, line 164: **[P2] The post-write reread narrows the concurrent initialization race but does not eliminate it.** For example, A and B can both observe a missing principal; B can then write and reread `B`, after which A writes and rereads `A`. B will continue reporting `B` for its process lifetime while the persisted value—and all future processes—report `A`. Principal creation needs an atomic create-if-absent operation or a lock; overwriting followed by a reread cannot guarantee that every concurrent process adopts the same winner.

Sol on behalf of Gideon

packages/cf-harness/src/cli.ts
```diff
@@ -277,6 +285,8 @@ export type CfHarnessCliSignalHandler = (
 export interface RunCfHarnessCliDependencies {
   cwd?: string;
   env?: Record<string, string | undefined>;
+  /** What caused this run. Resolved from the process when absent. */
+  provenance?: HarnessProvenance;
```
mathpirate, line 293: **[P2] This dependency currently controls `whoami`, but not the gateway requests for the run it is documented to describe.** `runCfHarnessCli()` always initializes and mutates the module-global provenance, and the default `OpenAICompatibleGatewayClient` reads that global value; `deps.provenance` is only consulted inside the `whoami` handler. A caller can therefore see its injected principal/invoker in `whoami` while actual requests carry a different process-derived principal/invoker. Please thread the effective provenance—including the resolved command and run manifest—through the prompt loop into the gateway client, and add a CLI-level test that compares the injected value with the request headers.

Sol on behalf of Gideon

packages/cf-harness/src/cli.ts
```diff
@@ -1603,6 +1615,40 @@ const runCfHarnessModelsCommand = async (
...
+  if (json) {
+    io.stdout(`${JSON.stringify(Object.fromEntries(entries), null, 2)}\n`);
+    return 0;
+  }
+  const lines = [
+    ...entries.map(([name, value]) => `${name.padEnd(10)} ${value}`),
+    "",
+    "The principal is a random label drawn once for this machine and kept in",
+    "the harness home. It is what the LLM gateway records for requests from",
+    `here, so when a run is traced to a principal, ${provenance.principal} is yours.`,
```
mathpirate: **[P3] This sentence prints the raw principal even though the field table, JSON output, and User-Agent use its sanitized representation.** A newline or terminal escape sequence in `CF_HARNESS_PRINCIPAL` or the persisted principal file can therefore inject additional lines or terminal controls into human-readable `whoami` output. Please reuse the already-sanitized principal from `entries` here so every representation is consistent and terminal-safe.

Sol on behalf of Gideon

packages/cf-harness/src/provenance.ts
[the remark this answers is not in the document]
Hixie: Claude's response:

The observation is correct and the wording it caught was mine — sanitize bounds a value, it does not strip prose, and the docs claiming otherwise are fixed. No further filtering is needed, because no field is fed from request content: the only one derived from the command line is command, drawn from a closed set of auth, models, whoami, prompt. The remaining values come from local configuration an operator set. Hashing a service name would make the field useless while protecting nothing at risk.

Hixie:
All comments before this one that have not received an explicit response are addressed (in principle).

PR #5401, opened 2026-08-05
dashboard: stop holding CI history in memory to serve it

Hixie:
The dev dashboard runs in a 1 GiB container, and its CI history was the largest thing in it. Two structures accounted for most of that, and neither needed to be in memory at all.

The run cache held every retained workflow attempt's complete job and step timings. One labs CI attempt is fifty-two jobs and seven hundred and forty-two steps, about 114 KB, of which 105 KB is step detail. Only the Gantt views draw steps, and only for the runs on one chart — at most the hundred and fifty the range slider allows. The duration charts, which are what the tile and the history page show, read one number per job: 3 KB per attempt, forty times less. Retention is sixty days, so the cache reached three thousand attempts: a 360 MB file that parsed into 601 MB of live objects, and a save that re-read and re-parsed it under its own lock, briefly holding three copies at once.

The discovery caches held GitHub's workflow-run entries as they arrived. A run in a query result carries its repository, head repository, whole head commit, and both the actor and the triggering actor: around 18 KB where the fields the dashboard reads come to 330 bytes. Nothing dropped the rest. The interfaces name the eleven or thirteen fields each consumer wants, but an interface removes nothing at runtime. A forty-five-day discovery window for both repositories, held for the length of the freshness window, measured 108.7 MB against 4.8 MB for the fields that get read.

## The run index and the Gantt detail

Completed attempts are now written to two places, split by how much of an attempt a view needs. `fabric-wall-ci-run-index.json` holds one entry per attempt — its run metadata and each job's duration — which is everything the duration charts read and small enough to keep in memory for the whole window: three thousand attempts parse to 14 MB. Beside it, `fabric-wall-ci-gantt` holds one file per attempt carrying every job's steps. A chart opens only the attempts it draws; a duration refresh opens none.

Whether an attempt is drawable and whether its jobs record which attempt they ran in are recorded in the index when the jobs are fetched, so the index can judge an attempt without opening its detail.

Each detail file is gzipped. An attempt's steps repeat the same handful of names and the same timestamp prefixes over and over, so gzip takes one to about 6 KB. At that size every attempt the index retains can keep its detail: the whole sixty-day window for both repositories is about 40 MB rather than 700 MB, so an attempt loses its detail only when it leaves the index altogether, and nothing GitHub would have to serve again is discarded early. A hand-packed binary layout was measured rather than assumed: with a string table and second offsets per attempt it reaches 16 KB, larger than gzipped JSON and considerably more code. Gzipping that layout as well reaches 4 KB, which does not justify carrying an encoding of our own.

Serving a chart no longer builds it. The collector settles which runs the chart contains, then hands them over one at a time, reading that run's attempt file and letting go of it before the next; the renderer's input file is written run by run as they arrive. A hundred-and-fifty-run chart used to leave 34 MB live and peak around 53 MB; it now leaves about 7 MB and peaks around 45 MB. The peak moves less than the retention does because it is dominated by the short-lived garbage of decompressing and re-serializing each attempt, which is bounded by the size of one chart either way.

Settling the chart and handing its runs over are two passes, so an attempt is read twice and dropped in between. Keeping the first read to save the second is precisely what would hold a chart's worth of timings at once.

Deciding whether to spend a GitHub request and deciding what to draw both go through the same read. Testing only that the file exists would make an attempt that cannot be read back — a format a later version does not recognize, a damaged file — both "no need to collect this" and "nothing to draw", so it would vanish from every chart and never be replaced.

Pruning drops an attempt's detail when the index no longer holds the attempt. The index decides, because the index is what records when each attempt ran: ordering by run identifier would misplace a rerun, which keeps its original identifier but starts later. Pruning waits while a chart is being assembled, so another collection cannot take a file that chart has already decided to draw.

A `fabric-wall-ci-job-history.json` file in the cache directory holds a layout the dashboard does not read and reaches hundreds of megabytes. A process that finds one deletes it, with its lock and any temporary file beside it, without parsing it, and collects the current window from GitHub instead.

## Workflow runs

Both discovery paths and the shared tile context now build a run from the fields their consumer declares, as each response arrives, so the surplus is never held. The selected-run lookup behind commit drill-downs already did this, which is where the shape came from.

The risk in a projection written by hand is the opposite of the one it fixes: omitting a field does not fail loudly, it reads as undefined wherever a view uses it. A test sets every declared field alongside the surplus GitHub sends and asserts the projection is exactly the declared fields.

## Bounded derived state

The per-window snapshot caches are keyed by a window length the URL can name, so walking the range slider could grow them without limit. They are now bounded, skipping any window still collecting — that skip is load-bearing rather than tidy, since a refresh reads its sampled runs back at the end of collection to record its manifest. Revisiting an evicted window costs no GitHub request and no write, because it rebuilds from the run index and falls back to the refresh already persisted there.

## Requests in flight

The three places that read many GitHub endpoints at once took a fixed-size group, waited for all of it, then took the next. A group is only as fast as its slowest member, so every other slot sat idle for the rest of that response — and for a GitHub read, waiting is nearly all of the time spent. A collection of two hundred runs is twenty-five of those waits in series.

They now take the next item the moment any one finishes, so there are always as many requests outstanding as the limit allows and one slow response costs only its own slot. The early stops the groups provided are kept, because they exist to avoid requests that have become pointless rather than to bound concurrency: a cache that will not accept writes stops a job collection, and a selected run that does not match its commit stops a commit drill-down, in both cases without starting the items that had not begun.

---
## Summary by cubic
Cuts dashboard memory by moving CI history out of process: a small in‑memory run index plus per‑attempt gzipped detail, and streams Gantt data to avoid holding charts in memory. Also trims GitHub run payloads, bounds snapshot caches, and makes concurrent fetches continuous.

- New Features
  - Store job durations in `fabric-wall-ci-run-index.json` (~14 MB for 60 days); write per-attempt step detail to `fabric-wall-ci-gantt/` (gzipped, ~6 KB each).
  - Stream Gantt charts: the data provider writes the renderer input file run-by-run and returns a count; the renderer no longer accumulates all runs in memory.
  - The index records drawability and attempt numbers; pruning keeps detail only for attempts still indexed.

- Refactors
  - Build workflow runs with only the fields tiles use, dropping ~18 KB of surplus per run.
  - Bound per-window snapshot caches (`SNAPSHOT_CACHE_MAX`); evicted windows rebuild from the run index without GitHub calls.
  - Replace fixed-size batches with rolling concurrency to keep request slots full and reduce idle time.
  - Remove legacy `fabric-wall-ci-job-history.json` if found and rebuild the current window.

<sup>Written for commit a7d06781ad8d82885d3def2ea2e928714816d305. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5401?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Hixie:
Claude wants me to say, in response to Cubic:

---
The race these comments describe requires prune and chart assembly to disagree about which attempts matter. They can't: both read the same run index.

#pruneGanttDetail hands prune() the result of store.attempts(repo, workflow) with no limit — every attempt the index holds. A chart's runs come from that same index, via store.get, store.latest, and store.list, on every path including the cached fallback. So the set prune keeps is always a superset of the set a chart draws, and there is no file for them to contend over. The only thing prune removes is detail for a run that has already left the index, which happens in exactly one place: the 60-day cutoff in CiJobHistoryStore.#prune.

The coordination you were reviewing was standing in for that invariant, left over from when prune kept a fixed number of recent attempts instead of tracking the index. Under a cap the two notions of "keep" genuinely can disagree, and a chart can read a file prune has decided to drop — but a lock does not fix that, it only narrows the window in which the disagreement becomes visible. Tying retention to the index is what fixed it, and the lock afterwards was guarding a window that was already closed.

So the claim, the stand-off and the in-flight wait are removed in 50470b484, and the invariant is now stated at the point the keep set is chosen and pinned by a test: after a collection, every attempt the index still holds still has its detail, and the one the index dropped does not. Re-introducing a cap fails that test by name — which is the regression actually worth guarding against, and the one a lock would have hidden.

PR #5406, opened 2026-08-05
Apply the same own-property fix where it is reachable but unproven

bfollington:
> **Stacked on #5373.** Split out of it deliberately — see below. Mergeable or droppable on its own merits; #5373 does not depend on it.

## Why this is separate

#5373 now carries only the sites with a **demonstrated** failure — each has a test that fails when the fix is backed out. These nine are the same predicate shape and reachable in principle, but **nothing here reproduces a bug.**

That makes them a judgement call rather than a fix, and the call should be made on its own instead of riding in on the proven set's coattails. Two of them are also in `cli` and `memory` — packages a runner reviewer shouldn't have to take on trust to approve a one-word change.

## The argument, per site

In every case the key can come from data or a schema, which is what makes a prototype-member name possible in the first place:

| site | key comes from |
|---|---|
| `builder/pattern.ts`, `cfc/structured-result.ts` | a **data** key from `Object.entries(target)`, tested against schema `properties` and then used to read the child schema — a field named `toString` would take `Object.prototype.toString` as its subschema |
| `cfc/prepare.ts` | pattern-path segments walked over data; both the presence test and the previous-value read |
| `builtins/sqlite-builtins.ts`, `memory/v2/sqlite/row-label.ts` | column and rule-field names against row records, then indexed — column names are legal SQL identifiers, so `toString` is available to anyone writing a query |
| `pattern-binding.ts` | binding-shape keys against the bound value |
| `cli/{callable,exec-schema,piece-get-transform}.ts` | schema-declared names (`required`, `properties`) against data or a built map |

## The argument against

None is demonstrated. If you'd rather not carry nine speculative changes across three packages, dropping this loses nothing that's currently broken.

For: `Object.hasOwn` is never a worse answer to an own-property question, and every change is mechanical.

## Test plan

No tests added, because no behaviour change is expected — which is itself part of what makes these weaker than the proven set, and worth saying plainly rather than dressing up with tests that would pass either way.

Suites unchanged: runner **1149 / 0**, memory **478 / 0**, cli **123 / 0**.

Refs [CT-1951](https://linear.app/common-tools/issue/CT-1951).

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5406?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cli/lib/callable.ts
+++ b/packages/cli/lib/callable.ts
@@ -344,7 +344,7 @@ function cloneWithoutBoundToolKeys(
 
   const required = Array.isArray(schema.required)
     ? (schema.required as string[]).filter((key) =>
-      key !== "result" && !(key in extraParams)
+      key !== "result" && !Object.hasOwn(extraParams, key)
     )
     : undefined;
 
--- a/packages/cli/lib/exec-schema.ts
+++ b/packages/cli/lib/exec-schema.ts
@@ -355,7 +355,9 @@ function parseObjectInput(
   // JSON input validation is deferred to the runner.
   if (!usedJson) {
     for (const key of requiredFlags(schema)) {
-      if (!(key in input)) {
+      // `Object.hasOwn`: a required flag named for a prototype member would
+      // otherwise read as supplied on every input.
+      if (!Object.hasOwn(input, key)) {
         throw new Error(`Missing required flag --${flagNameForKey(key)}`);
       }
     }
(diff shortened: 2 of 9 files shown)


bfollington:
I couldn't confirm that this was worth doing, it's probably just being overly aggressive on pattern matching about places prototypes could matter rather than looking for real problems.

PR #5412, opened 2026-08-05
Move agent harness specs into Labs

mathpirate:
## Summary

- move the draft Agent Harness runtime, CFC integration, and conformance specifications into Labs' live `docs/specs/` tree
- update the generic contract for explicit provider/credential routing, context replacement, trusted-host tools, child working-directory containment, model-bound opaque references, immutable observations, and request attribution
- refresh the `cf-harness` implementation profile and current-state reference against Labs `83671e20d677730739f29d7d53134c60e374a6c8`
- document the shipped implementation seams: Responses compaction, content-addressed image observations, session-local and delegation-scoped address handles, shape-only `describe_handle`, registered and cancellable trusted-host `run_pattern`, Fabric-session CFC posture, persistent provider/auth state, gateway provenance, and strict local Loom resume binding
- give each temporary deviation an owner and retirement condition, including incomplete dependency preflight, product `observe` bridges, provisional network policy, handle-boundary gaps, and durable `run_pattern` piece lifecycle

## Why

Labs owns the current `@commonfabric/cf-harness` implementation and its tests. Keeping the governing live contract here lets behavior, test evidence, and specification changes land together while keeping exact flags, schemas, tool spellings, token grammar, deployment defaults, and product-adapter policy in their owning implementation profiles.

The directory remains `agent-harness/` because the contract is implementation-independent. It can be renamed mechanically during review if `cf-harness/` is clearer.

This supersedes commontoolsinc/specs#17 and folds in the current-state refresh proposed in #5951, updated through the subsequent handle-shape, pattern-registration, cancellation, and Fabric-session-CFC work on `main`.

## Impact

Documentation only. No runtime behavior, public API, or deployment configuration changes.

## Checks

Using the repository-pinned Deno 2.9.4:

- `deno fmt --check` (4,648 files)
- `deno lint` (4,614 files)
- `deno task docs-links --orphan`
- `deno task check-docs` (559 code blocks)
- `deno task check-conflict-markers`
- `cd packages/cf-harness && deno task test` with an isolated private `CF_HARNESS_HOME` (736 passed)

### DIFF
--- a/docs/specs/README.md
+++ b/docs/specs/README.md
@@ -33,6 +33,10 @@ decision is reversed or superseded).
 - [Hosted pattern authoring](hosted-pattern-authoring.md)
 - [Scoped cell instances](scoped-cell-instances.md)
 
+### Agent execution
+
+- [Agent harness](agent-harness/README.md)
+
 ### Data, storage, and execution
 
 - [JSON Schema](json_schema.md)
--- a/docs/specs/agent-harness/01-runtime-contract.md
+++ b/docs/specs/agent-harness/01-runtime-contract.md
@@ -1,0 +1,327 @@
+# Agent Harness Runtime Contract
+
+Status: draft 0.1
+
+This document specifies the caller-visible behavior of a Common Fabric agent
+harness. The contract is transport-neutral: a batch CLI, long-lived stdio
+service, or another protocol may conform if it preserves the same lifecycle,
+authority, containment, and evidence semantics.
+
+## 1. Roles and trust boundaries
+
+- The **caller** selects the task, prompt-slot roles, capabilities, mounts,
+  policy profile, and resource bounds.
+- The **harness** runs the model/tool loop and records authoritative run
+  lifecycle evidence. It does not invent authority on the model's behalf.
+- The **model gateway** produces model messages and tool requests. Its output is
+  untrusted proposal data unless another trusted component supplies evidence to
+  the contrary.
+- The **execution substrate** executes tools inside the configured containment
+  boundary.
+- A **mediator** supplies trusted observation or policy evidence. It may be the
+  execution substrate, the Common Fabric runner, or another declared trusted
+  component.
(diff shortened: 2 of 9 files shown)


bfollington:
This looks good — the structure is right (dated profile, pinned revision, deviations with owners and retirement conditions), every relative link resolves, and I spot-checked the profile's claims against the pinned revision (compaction fraction, `run_pattern` source cap, attribution headers, handle table, image snapshots): all accurate.

It is, however, already out of date — because of a change **I** made. CT-2003 (#5781) merged 2026-08-17, three days after this profile's verification date, and it changes exactly the seams documented here:

1. `describe_handle` is now a selectable parent tool — the parent-tool list omits it.
2. The `pattern-author` child profile exists — the child-profile list omits it.
3. Delegation now seeds the child's handle table with exactly the entries whose tokens the parent named in the goal/context, and child-discovered references return as parent-minted tokens — so "delegation arguments carry tokens only as inert text" and deviation 4's "no cross-agent transfer, dereference" no longer describe the tree.

Notably the spec clauses themselves survive unchanged: AH-REF-3 forbids *implicit* transfer, and CT-2003's parent-selected seeding is the explicit mechanism the clause anticipated. It's only the profile/current-state prose that aged out — which is your dated-profile structure working as intended.

Rather than ask you to re-roll, I've stacked the refresh as #5951 (base: this branch). Merge this as written and land that on top — or fold it in, whichever you prefer.

mathpirate:
Thank you—this was exactly right. I folded #5951 into this branch and rebased it onto current `main`. Since the implementation had moved again after your refresh, I also brought the profile through the newer `describe_handle` shape semantics, registered/cancellable `run_pattern`, and Fabric-session CFC posture.

I agree with your key distinction: the generic spec clauses held up unchanged; it was the dated current-state/profile material that needed refreshing. #5951 should now be redundant.

PR #5413, opened 2026-08-05
refactor(cf-harness): replace the trivial run-state setters with a pa…

Hixie:
…tch helper

The run state of a harness run is an immutable record: every update copies the old state, changes a field or two, and stamps a new `updatedAt`. To express that, run-state.ts exported twenty-one named update functions, and seventeen of them were the same three-line body with a different field name substituted in. Each one existed only so that the engine could avoid writing an object spread. Every one of those seventeen had exactly one caller, all of them in the engine, and none had a caller in the tests. Because the package entry point re-exports run-state.ts wholesale, all seventeen were also part of the package's public surface, which nothing outside the package used — the harness is driven over its command line and its line-delimited JSON stdio interface, not by importing its TypeScript.

This replaces those seventeen with two general helpers:

`patchHarnessRunState(state, patch, now)` copies the state, applies the fields named in the patch, and stamps `updatedAt`. A patch entry whose value is undefined is left out rather than written as an explicit undefined property. That reproduces what the old setters did by hand for their optional second argument, so a caller can pass an artifact path that the artifact store may not have produced straight through without testing it first.

`appendToHarnessRunState(state, field, entry, now)` adds one entry to a list-valued field, treating an absent list as empty. The field name is constrained by a mapped type to the run state's list-valued properties, and the entry is constrained to that list's element type, so naming a non-list field or passing the wrong kind of entry is a compile error.

The four update functions that carry real logic stay, and now build their result through the patch helper rather than spreading by hand: `setHarnessRunStatus` (which adds an end timestamp and terminal reason on a terminal status and strips both on a non-terminal one), `setHarnessSubagentRun` (which replaces a matching child run rather than appending a duplicate), `appendHarnessCfcModelContextObservations` (which merges observations and returns the original state unchanged when nothing moved), and `appendHarnessFailureRecord` (which recomputes the primary failure over the whole list on every append).

Field order in the resulting object, and therefore in the persisted run-state.json, is unchanged: an assignment to a property the state already has keeps that property's original position, and a property the state lacks is appended, which is what the spreads did before.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5413?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cf-harness/src/engine.ts
+++ b/packages/cf-harness/src/engine.ts
@@ -9,30 +9,15 @@ import {
   type HarnessArtifactStore,
 } from "./artifacts.ts";
 import {
-  appendHarnessCfcInvocationContext,
   appendHarnessCfcModelContextObservations,
   appendHarnessFailureRecord,
-  appendHarnessPolicyDecision,
-  appendHarnessPolicyEvent,
-  appendHarnessToolOutput,
+  appendToHarnessRunState,
   createHarnessRunState,
   type HarnessRunState,
   type HarnessRunTerminalReason,
-  setHarnessCapabilitySnapshot,
-  setHarnessCfcPolicySnapshot,
-  setHarnessPolicyTrace,
-  setHarnessPromptSlotBinding,
-  setHarnessRunCurrentDir,
-  setHarnessRunManifestPath,
-  setHarnessRunModel,
-  setHarnessRunReportPath,
+  patchHarnessRunState,
   setHarnessRunStatus,
-  setHarnessSkillActivations,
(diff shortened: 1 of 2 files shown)


mathpirate (approved):
lol 🤦 thank you for fixing this 😱

PR #5414, opened 2026-08-05
refactor(cf-harness): record each model attempt once

Hixie:
When cf-harness gained a provider-neutral model client, run reports gained a matching provider-neutral record of each attempt at a model call, called `modelAttempts`. The older, gateway-specific `gatewayAttempts` record was kept alongside it so that anything reading the old field would keep working during the transition.

Nothing ever read the old field. Inside the package, `gatewayAttempts` appeared only in the contract that declares it and in the prompt loop that writes it: the CLI, the diagnostics module, the artifact readers, and the interactive chat service all ignore it, and no consumer outside the package reads it either. What the transition left behind was a round trip that produced no information: the gateway adapter converted the transport client's diagnostic into the neutral shape by adding a type tag and a provider id, and the prompt loop then converted that neutral record straight back into the old shape by stripping the provider id and restoring the old type tag, so every model call was written to `run-report.json` twice in two spellings of the same fields.

This deletes the legacy record. `HarnessGatewayAttempt` and the `gatewayAttempts` field are gone from the run-report contract, along with the option that fed them and the branch in `createHarnessRunReport` that persisted them. The prompt loop no longer keeps a second array or performs the reverse conversion, so recording an attempt is now just stamping the run id, the sequence number, and the model turn onto the diagnostic the model client handed over.

The transport-level diagnostic in `gateway/openai-client.ts` stays, as does the seven-line adapter in the gateway model client that lifts it into the neutral shape. The two types are not redundant: the transport diagnostic constrains `operation` to the two endpoints that client speaks, while the neutral one leaves it a free string that the Codex client fills with `responses.stream`. The provider id the adapter stamps is also a model-layer fact, defined next to the model client, so having the HTTP client stamp it would push knowledge of model providers down into transport code for no gain.

Which API served a gateway turn was the one thing the old record carried that nothing asserted; it now lives in the `operation` field of a model attempt, and the artifacts test asserts it there along with the rest of the assertions moved over from `gatewayAttempts`.

Two live documents described the double record, and both now describe what the code does instead: the package README, and the work package in the subscription-auth plan that had called for keeping the gateway field until an artifact-version migration removed it. That work package also loses the word in its heading naming the compatibility it no longer keeps.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5414?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/plans/cf-harness-codex-subscription-auth.md
+++ b/docs/plans/cf-harness-codex-subscription-auth.md
@@ -269,13 +269,14 @@ Tests to extend first:
 - `packages/cf-harness/test/openai-client.test.ts`
 - `packages/cf-harness/test/interactive-chat-service.test.ts`
 
-### WP1.3 — Generalize attempt provenance compatibly
+### WP1.3 — Generalize attempt provenance
 
 - [x] Add provider-neutral model-attempt records to run reports with provider,
   operation, endpoint origin, timing, request summary, status, selected request
   id, and bounded error metadata.
-- [x] Preserve reading and producing the current `gatewayAttempts` field for the
-  existing gateway until an explicit artifact-version migration removes it.
+- [x] Record each attempt exactly once, in the provider-neutral `modelAttempts`
+  field. A gateway turn carries the same kind of record as any other turn, with
+  `operation` naming the API that served it.
 - [x] Never record authorization, cookies, account ids, refresh responses, or
   arbitrary response headers.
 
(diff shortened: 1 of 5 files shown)


mathpirate (approved):
thank you

PR #5418, opened 2026-08-05
test(runtime-client): cover FabricBytes VDOM props across worker IPC

mathpirate:
## Summary

Adds a focused green regression for the worker → main-thread VDOM prop seam closed by #5823 and #6323.

A pattern can render a `FabricBytes` as an element prop. The reconciler must emit it whole, and the runtime-client transport must carry the containing `VDomBatch` through the realm boundary without structured clone reducing the primitive to `{}`.

## What the test exercises

The test follows the production composition rather than mirroring its individual helpers:

1. A real `WorkerReconciler` renders a `<cf-image>` VNode with a `FabricBytes` `bytes` prop.
2. The reconciler emits the real `set-prop` operation inside a VDOM batch.
3. Production `postToClient()` realm-encodes the complete notification.
4. `structuredClone()` reproduces the `postMessage()` crossing.
5. `fabricFromRealmValue()` reproduces the client transport's decode.
6. The delivered prop is asserted to still be a `FabricBytes` with the original contents.

The existing outbound capture helper now includes the structured-clone step as well, so its other tests model encode → clone → decode rather than encode → decode.

## Why this adds coverage

The underlying pieces already have strong independent coverage:

- `codec-realm` preserves `FabricBytes` (#5823).
- The complete runtime-worker IPC envelope uses the realm codec (#6323).
- `convertCellsToLinks()` treats fabric primitives as leaves.
- `cf-image` accepts byte-like values.

What was not pinned was their composition for the exact VDOM prop path that exposed the original failure. A future reconciler transformation or VDOM-specific transport change could flatten the value while the codec and generic worker tests continued to pass.

This PR deliberately does not duplicate `cf-image` coercion or maintain a local copy of component behavior.

## Verification

- `deno test packages/runtime-client/test/backends/post-to-client.test.ts`
- `cd packages/runtime-client && deno task test` — 371 steps passed
- `deno task check`
- `deno fmt --check`
- `deno lint`

<sub>Originally opened as a deliberately-red probe of the raw structured-clone gap. Dan's realm-codec work resolved the production path; the PR is now the green composition regression for that resolution.</sub>

### DIFF
--- a/packages/runtime-client/test/backends/post-to-client.test.ts
+++ b/packages/runtime-client/test/backends/post-to-client.test.ts
@@ -1,9 +1,11 @@
 /**
- * What the worker's outbound side does with a message the encoding refuses.
+ * What the worker's outbound side carries, and what it does with a message the
+ * encoding refuses.
  *
- * `postToClient()` is the whole of that side, so it is also the only place a
- * failure to encode can be answered. What it answers with differs by whether
- * anyone is awaiting a reply, and it must not throw either way.
+ * `postToClient()` is the whole of that side. A successful message crosses as
+ * one realm encoding; a failure to encode is answered here. What a failure
+ * becomes differs by whether anyone is awaiting a reply, and it must not throw
+ * either way.
  */
 
 import { describe, it } from "@std/testing/bdd";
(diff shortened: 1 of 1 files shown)


danfuzz:
[Sorry, I only just saw my mention here.]

As it happens I just hit more or less the same point in my own related work.

Of the options presented, (c) is what I've been planning on, and specifically by defining a subpackage `codec-realm` (name workshopping hereby solicited) to parallel the `codec-json` subpackage.

In case it's not obvious what I mean by that name: The `data-model` should directly support encoding and decoding across a boundary for which `structuredClone()`-compatible values are allowed, that is, `codec-realm` is a codec for data that crosses realm boundaries. Just like `codec-json` contains implementation detail about which of the various values are JSON-compatible vs. not and e.g. has a special case for `bigint`, `codec-realm` would have implementation detail about what's kosher for `structuredClone()` and would e.g. have a special case for `FabricPrimitive`.

About the note in the options list that (c) means a "cost of every DOM consumer having to decode," I don't think that's particularly avoidable for any full-throated use of the `data-model`. That is, if you're going to have the `data-model` on both sides, then you'll always have a decode step on the receiving end. I guess the question is whether or not we ever want to have the implied asymmetry that would let us get away with a one-way transformation.

danfuzz (approved):

PR #5421, opened 2026-08-05
fix(coverage): score type-only modules as zero tracked lines

ubik2:
## Problem

A module whose every top-level statement is erased by type stripping transpiles to an empty module, so V8 never emits a coverage record for it — even when tests import it at runtime (verified for `packages/api/schema.ts`, which `packages/api/test/factory-input-types.test.ts` imports both statically and dynamically; no `SF` entry appears in any CI LCOV artifact). The debt metric read the missing record as "never loaded" and charged every tracked line as permanently uncoverable debt. Any PR adding type-level lines to such a file failed the ratchet with an "add tests" suggestion that is impossible to follow — see #5365, where +10 type-only lines in `api/schema.ts` tripped the gate and required an `ACCEPT_COVERAGE_DEBT` override.

## Change

- `tasks/coverage-metrics.ts` gains `producesNoExecutableCode()`: it parses each tracked file with the TypeScript compiler and scores a file whose statements are all erased by type stripping (interfaces, type aliases, wholly type-only imports/exports, ambient `declare` declarations) as **zero tracked lines**. `collectUncoveredLinesForFiles` likewise reports no lines for one, so the PR-comment suggestions stay consistent with the gate.
- Detection is conservative on purpose: any statement not recognized as erased (a value import, a bare side-effect import, an enum) keeps the file tracked, so a mistake leaves debt standing rather than hiding real uncovered code. The `.d.ts` suffix exclusion remains as the declared-by-name case of the same rule.
- `packages/api/perf/` is excluded outright via `EXCLUDED_RELATIVE_PREFIXES`. The type-only rule alone would not release it: `ikeyable_realistic.ts` deliberately holds value statements to stress the checker, and `run-tsc.ts`/`run-benchmarks.ts` are compile-harness tools — none exist to be run by tests.
- `tasks` now declares `npm:typescript@6.0.3`, the same pin as the other eight workspace members; `docs/development/DEPENDENCIES.md` updated to count it.
- `docs/development/COVERAGE.md` gains a "Type-only modules are scored as zero tracked lines" section.

Parsing all 1,838 tracked files costs ~1s in the coverage job (measured).

## Effect on ratchet baselines

This lowers the maximum possible debt in 14 groups, **2,706 tracked lines** total: `packages/api` −1,526 (schema.ts + the perf harness), `packages/cf-harness` −236, `packages/patterns` −220, `packages/runner` −200, `packages/ui` −151, `packages/memory` −145, and eight smaller groups (`html` −71, `toolshed` −68, `fuse` −32, `deno-web-test` −19, `fs-sync-example` −17, `ts-transformers` −8, `data-model` −7, `content-hash` −6). A repo-wide baseline **drop** is safe for a ratchet — it only fails on rises — and the first `main` run after merge seeds the lower baselines.

## Tests

New cases in `tasks/coverage-metrics.test.ts`: the recognizer both ways (type-level module → true; `declare` + value statement, bare import, value import → false), a record-less type-only file adding zero group debt, `collectUncoveredLinesForFiles` returning nothing for a type-only file, and the perf-path exclusion. Full `tasks` suite: 450 passed.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5421?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/deno.lock
+++ b/deno.lock
@@ -2210,6 +2210,11 @@
           "npm:htmlparser2@12",
           "npm:marked@^18.0.6"
         ]
+      },
+      "tasks": {
+        "dependencies": [
+          "npm:typescript@6.0.3"
+        ]
       }
     }
   }
--- a/docs/development/COVERAGE.md
+++ b/docs/development/COVERAGE.md
@@ -102,7 +102,9 @@ covered a line.
 
 One detail of the gate's accounting is worth knowing when reasoning about
 pattern coverage. A file with no LCOV record has every tracked line counted as
-uncovered. A file with a record is scored against the lines that record names.
+uncovered — unless the file produces no executable code, as described in the
+next section. A file with a record is scored against the lines that record
+names.
 For a file measured by Deno's V8 coverage that is every executable line; pattern
 instrumentation names only the statements it could instrument, so a pattern
 file's first record both covers real lines and drops the never-named lines out of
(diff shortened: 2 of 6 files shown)


ubik2:
Superseded by #5403, which landed the same fix on main with an emit-based executable-code test (tasks/executable-source.ts) evaluated lazily for record-less files.

PR #5423, opened 2026-08-06
docs(cli): specify persistent cf daemon

mpsalisbury:
## Hold

This PR is intentionally held in draft. Daemon design and implementation will not resume until the server-transition work has launched. That transition is expected to change which state and execution responsibilities remain client-side, materially changing the appropriate boundary and shape of any `cf` daemon. Once the transition is live, this work should begin with fresh latency measurements and a reassessment of retained state, request lifecycle, command scope, and whether a daemon is still warranted before revising the design.

## Why

Fine-grained `cf` workflows repeatedly pay for process startup, identity-backed connection setup, runtime construction, health checking, authenticated storage setup, and replica synchronization. Before implementing a longer-lived connection host, the project needs an explicit contract for what may remain warm and how concurrency, mutation ambiguity, invocation identity, source staleness, security, and cross-checkout ownership behave.

## What Changed

- Add a live design spec for an explicitly selected, named `cf` daemon.
- Keep execution and Fabric invocation identity request-scoped while retaining the authenticated storage connection, replicas, and safe caches.
- Resolve `--invocation` and `--invocation-session`/`CF_INVOCATION_SESSION` in the short-lived client and keep that durable identity separate from the in-memory daemon request ID.
- Define supported command stages, serialized request handling, piped input ownership, outcome-unknown mutations, namespace compatibility, source invalidation, lifecycle, security, and validation gates.
- Index the spec under data, storage, and execution.

## Test plan

- `deno fmt --check`
- `deno lint`
- `deno task check-docs`
- `deno task docs-links --orphan`
- `deno task check-conflict-markers`
- `git diff --check`

### DIFF
--- a/docs/specs/README.md
+++ b/docs/specs/README.md
@@ -35,7 +35,7 @@ decision is reversed or superseded).
 
 ### Data, storage, and execution
 
-- [`cf` CLI session daemon](cf-cli-session-daemon.md)
+- [`cf` CLI daemon](cf-cli-daemon.md)
 - [JSON Schema](json_schema.md)
 - [Sigil data model](data-model/sigil.md)
 - [Sparse-array preservation](sparse-array-preservation.md)
--- a/docs/specs/README.md
+++ b/docs/specs/README.md
@@ -35,6 +35,7 @@ decision is reversed or superseded).
 
 ### Data, storage, and execution
 
+- [`cf` CLI session daemon](cf-cli-session-daemon.md)
 - [JSON Schema](json_schema.md)
 - [Sigil data model](data-model/sigil.md)
 - [Sparse-array preservation](sparse-array-preservation.md)
--- a/docs/specs/cf-cli-daemon.md
+++ b/docs/specs/cf-cli-daemon.md
@@ -3,6 +3,14 @@
 ## Status
 
 This document specifies intended behavior. The daemon is not implemented.
+The invocation-identity requirements depend on the shaped-reads plan and its
+pending CLI and runner implementations ([#5435], [#5460], and [#5469]). Until
+that contract lands, they specify coordinated intended behavior rather than
+current direct-execution behavior.
+
+[#5435]: https://github.com/commontoolsinc/labs/pull/5435
+[#5460]: https://github.com/commontoolsinc/labs/pull/5460
+[#5469]: https://github.com/commontoolsinc/labs/pull/5469
 
 ## Summary
 
@@ -143,8 +151,9 @@ but do not drive the design.
 ### Direct-only commands
 
 Local, offline, test, process-management, and interactive commands remain
-direct. This includes `check`, `test`, `view`, `id`, `init`, `inspect`, `space`,
-`fuse`, daemon-management commands, and completion script generation.
+direct. This includes the top-level `cf check`, `cf test`, `cf view`, `cf id`,
+`cf init`, `cf inspect`, `cf space`, and `cf fuse` commands, daemon-management
+commands, and completion script generation.
 `cf piece render --watch` and any future watch or subscription form remain
 direct. `cf exec` also remains direct; routing a shebang-run mounted callable
 through a named daemon requires a separate explicit propagation design.
--- a/docs/specs/cf-cli-daemon.md
+++ b/docs/specs/cf-cli-daemon.md
@@ -3,14 +3,14 @@
 ## Status
 
 This document specifies intended behavior. The daemon is not implemented.
-The invocation-identity requirements depend on the shaped-reads plan and its
-pending CLI and runner implementations ([#5435], [#5460], and [#5469]). Until
-that contract lands, they specify coordinated intended behavior rather than
-current direct-execution behavior.
 
-[#5435]: https://github.com/commontoolsinc/labs/pull/5435
-[#5460]: https://github.com/commontoolsinc/labs/pull/5460
-[#5469]: https://github.com/commontoolsinc/labs/pull/5469
+The invocation-identity rules it describes ARE in effect for direct execution:
+`resolveInvocationIdentity` (`packages/cli/commands/piece.ts`) mints both
+halves when neither is supplied, takes both when both are, and rejects an
+invocation id offered without the session it is replayable within. What this
+document adds is where those rules sit relative to a daemon — resolved by the
+short-lived client before admission, carried as request-owned data, never read
+or minted or retained by the daemon process.
 
 ## Summary
 
@@ -315,7 +315,8 @@ Every request owns and must release:
 - error and console collectors;
 - navigation callbacks and other command effects;
 - the caller's Fabric invocation identity: the `--invocation` ID paired with
-  the invocation session resolved from `--session` or `CF_SESSION`; and
+  the invocation session resolved from `--invocation-session` or
+  `CF_INVOCATION_SESSION`; and
 - filesystem/compiler state that is not safely content-addressed.
 
 The Fabric invocation identity is caller input, not connection state. The pair
@@ -325,11 +326,12 @@ would place unrelated callers in one deduplication scope, where one caller
 could be told that its request settled on another caller's outcome.
 
 The short-lived client resolves the pair under the same rules as direct
-execution before daemon admission. `--session` overrides `CF_SESSION`; when
-neither half is supplied, the client mints both; and an invocation ID without
-an invocation session is rejected before dispatch. The normalized pair travels
+execution before daemon admission. `--invocation-session` overrides
+`CF_INVOCATION_SESSION`; when neither half is supplied, the client mints both;
+and an invocation ID without an invocation session is rejected before dispatch. The normalized pair travels
 as an explicit request-envelope field. The daemon process never reads its own
-`CF_SESSION`, never receives it as request environment, never mints either
+`CF_INVOCATION_SESSION`, never receives it as request environment, never mints
+either
 half, and never retains the pair after request cleanup.
 
 The next request is not admitted to execution until cleanup has stopped
@@ -595,7 +597,7 @@ parsing or validation rules in a daemon-only client path.
    performs context-free validation, and lets daemon handlers perform schema-
    dependent validation without ambient stdin access. Carry the client-resolved
    Fabric invocation identity as request-owned data without forwarding
-   `CF_SESSION`.
+   `CF_INVOCATION_SESSION`.
 7. Add filesystem and compiler commands only after cwd, input freshness, and
    content-cache boundaries are validated.
 
--- a/docs/specs/cf-cli-daemon.md
+++ b/docs/specs/cf-cli-daemon.md
@@ -1,4 +1,4 @@
-# `cf` CLI session daemon
+# `cf` CLI daemon
 
 ## Status
 
@@ -6,12 +6,12 @@ This document specifies intended behavior. The daemon is not implemented.
 
 ## Summary
 
-Fabric-facing `cf` commands currently create an identity session, runtime,
-storage manager, authenticated memory connection, and synchronized replica for
-each invocation. A named local daemon retains the reusable connection state so
-separate CLI invocations can avoid that setup cost.
+Fabric-facing `cf` commands currently create an identity-backed connection,
+runtime, storage manager, authenticated memory connection, and synchronized
+replica for each invocation. A named local daemon retains the reusable
+connection state so separate CLI invocations can avoid that setup cost.
 
-The daemon is a **session host**, not an execution host. It retains identity,
+The daemon is a **connection host**, not an execution host. It retains identity,
 connection, replica, and safe content-addressed caches. Every request receives
 fresh execution state, and request-created pieces, reactive graphs,
 subscriptions, collectors, and background work are torn down before the next
@@ -46,7 +46,7 @@ replays a mutation.
 - Accelerating local-only, offline, test, or interactive terminal commands.
 - Serving watch, subscribe, streaming, or otherwise indefinite commands. These
   commands retain execution by definition and remain direct. A retained-
-  execution host, if wanted, is a separate design built on the session and
+  execution host, if wanted, is a separate design built on the connection and
   lifecycle seams established here.
 - Sharing one daemon across unrelated identities, APIs, primary spaces, or
   runtime configurations.
@@ -289,7 +289,7 @@ fixed or kept request-owned.
 The daemon may retain:
 
 - the deserialized identity and signer;
-- the storage manager and its authenticated session identity;
+- the storage manager and its authenticated storage connection identity;
 - memory transports and synchronized replicas;
 - storage watches and bounded storage-level pull knowledge, including the
   per-document deduplication state that prevents repeated pull registration;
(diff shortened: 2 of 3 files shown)


mathpirate:
something along these lines would be really useful to have. like i mentioned in person, it might be worth checking on how agent-browser's CLI implements its notion of sessions, bc they share a lot of behavioral characteristics with what i think we're going for here; i *think* they do it without needing a persistent daemon process, but i'm not positive about that, and that might not be possible for us given the differences between browser sessions and fabric sessions

mpsalisbury:
Flagging an interaction with in-flight work on invocation identity, plus proposed spec text.

**Context, and its status.** #5460 added `cf session new` / `CF_SESSION` / `--session`. #5469 (open, stacked on #5470) makes the session load-bearing: `scopeCallerEventId` hashes it alongside the caller's id and the stream link, so a receipt's address derives from the **pair**. The same `--invocation add-comment-1` chosen in two sessions names two invocations and lands two receipts. An invocation id supplied without a session is a refusal.

None of that is on `main` yet, so this is a heads-up rather than a correction — but this spec is live and unimplemented, and the collision is cheap to prevent now and subtle to debug later.

**The collision.** "Session" now names two unrelated things in `cf`: this spec's session host and its "authenticated session identity" (a *connection* concept), and `CF_SESSION` (an *invocation-scoping* concept). They have nothing in common.

The spec partitions state into retained and request-owned, and `CF_SESSION` is in neither list — it did not exist when this was written. An implementer classifying it by its nearest neighbour will retain it, because identity and the authenticated session identity are both retained. Then every caller through one daemon shares a single deduplication scope, and a second caller's replay settles on the first caller's outcome. It fails looking like a *successful dedup*, not like an error.

The intent for the daemon is that it changes the semantics of no call and holds nothing but its connection parameters. Under that intent the session is caller input, not daemon state, and everything below just writes that down.

---

### Proposed: add to `### Request-owned state`

A bullet in the "Every request owns and must release" list:

```
- the invocation session (`--session`, `CF_SESSION`) that scopes a
  caller-supplied invocation id;
```

and a paragraph after that list:

```
The invocation session is caller input, not connection state. It decides where
a handling's receipt lands, so the same invocation id under two sessions names
two invocations. A daemon that retained one, or minted one, would place every
caller routed through it into a single deduplication scope, and a replay by one
caller would settle on another caller's outcome. The daemon therefore never
reads `CF_SESSION` from its own process environment, never mints a session on a
caller's behalf, and receives the session only on the allowlisted per-request
environment. A request naming an invocation id without a session is the
client's refusal to report, unchanged by routing.
```

### Proposed: add to `## Security`

```
- The invocation session is excluded from default logs. It is the only
  unguessable component of a receipt's address, so recording it publishes where
  a caller's outcomes live to anything that can read the daemon log.
```

### Proposed: add to `## Validation`

```
- an identical receipt address for the same command run directly and routed;
```

This one is worth its line: a receipt's address is a hash over the caller's id, the stream link, and the session, so a single assertion catches every way the daemon could contaminate the session — reading its own environment, minting, or retaining one across requests.

### Proposed: one line in `## Request identity and effectful outcomes`

That section closes with:

> Durable exactly-once behavior requires a server-recognized idempotency contract and is separate work.

That separate work now exists — `--invocation` with `--session` is that contract, at the Fabric layer rather than the daemon's. Worth a sentence pointing at it, and making explicit that the daemon's instance-qualified request ID is a third, unrelated identifier: daemon-internal and in-memory, never the caller's idempotency key.

---

**On the naming.** Two concepts sharing "session" in one CLI will keep costing readers. Renaming either is bigger than this PR, but the prose can at least disambiguate where the two would otherwise be read as one system.

Happy to open this as a PR against the branch instead of a comment if that is easier to apply.

mpsalisbury (commented):
cf-review via Claude Fable 5, on behalf of @mpsalisbury

**Scope:** 2 files, +644 (new spec `docs/specs/cf-cli-daemon.md` + index entry) · **Coverage:** deep-read, claims verified against source · **Checks:** repo-wide `deno fmt --check`, `check-conflict-markers`, command/flag inventory greps, companion PR diffs

This is an unusually well-grounded spec. Every claim it makes about the existing system held up under verification:

- `Runtime.dispose({ closeStorage: false })` exists and its intent matches the spec's description (`packages/runner/src/runtime.ts`, `dispose()`).
- The subscription-leak claim is **real and verified**: both `Scheduler` (`packages/runner/src/scheduler/facade.ts`, constructor) and `Runner` (`packages/runner/src/runner.ts`, constructor) call `storageManager.subscribe()` at construction and discard the return value; nothing in disposal unregisters, so sequential runtimes over a retained manager would keep disposed schedulers/runners reachable — exactly as the spec says.
- `CF_CLI_TRACE_TIMINGS` exists; every command the spec names exists (all 20 `piece` subcommands, `deps update`, `wish`, the hidden `completion complete`, `piece get --step`); `--api-url`/`--identity`/`--space` are the real flag names; the receipt/at-most-once semantics match the scheduler's event-identity machinery.
- Gates clean: fmt, conflict markers, indexed in `docs/specs/README.md`, Status-header convention matches sibling specs, American spelling throughout.

### 🟡 Improvements

**1. The invocation-session rules are stated as existing direct-execution behavior, but they only exist in unmerged PRs.** — `docs/specs/cf-cli-daemon.md` ("Request-owned state") · verified

The spec says the client resolves the invocation pair "under the same rules as direct execution": `--session` overrides `CF_SESSION`, neither supplied → both minted, ID without session → rejected. On main today, `CF_SESSION` and "invocation session" appear **nowhere but this document**. The rules live in three open PRs: #5435 (the plan defining mint-both/reject), #5460 (the `--session`/`CF_SESSION` mechanism — and its S1 code deliberately returns `undefined` when absent, minting and rejecting nothing yet), and #5469 (runner scoping). If #5423 merges first, a reader who greps for these words lands here and is told present-tense rules that direct execution doesn't have. **Fix:** either sequence this merge after the rule-bearing stage lands, or add one sentence marking the invocation-session rules as defined by the shaped-reads plan and not yet in effect (with a link once #5435 lands).

**2. `inspect` and `view` in the direct-only list read ambiguously.** — `docs/specs/cf-cli-daemon.md` ("Direct-only commands") · verified

"This includes `check`, `test`, `view`, `id`, `init`, `inspect`, …" means the *top-level* commands, but `cf piece inspect` is in the first daemon scope and `cf piece view` in optional scope. A skimmer can read the bare names as covering the piece forms. Qualify them (e.g. "the top-level `cf view` and `cf inspect`").

### Questions for the author

- What's the intended merge order across #5423 / #5435 / #5460 / #5469? Finding 1 dissolves if this lands last (or gains the link).

### Possible follow-ups

- The spec itself notes the storage-subscription ownership fix "is useful independently of this daemon" — the leak is verifiably real today, so it's worth filing as its own issue rather than waiting on the daemon go/no-go gate.

**Verdict:** No blockers — a carefully verified spec whose one real issue is cross-PR sequencing of the invocation-session vocabulary; mergeable as-is if it lands after (or explicitly points at) the shaped-reads work.

mpsalisbury:
Addressed both review improvements in 755517f92:

- the Status section now marks invocation identity as dependent on the pending shaped-reads CLI/runner stack (#5435, #5460, #5469), rather than implying that behavior is already on `main`;
- the direct-only inventory now qualifies `cf view`, `cf inspect`, and the other entries as top-level commands, avoiding ambiguity with `cf piece view` and `cf piece inspect`.

Repo-wide fmt/lint, docs checks, orphan-link checking, conflict-marker checking, and `git diff --check` all pass.

mpsalisbury:
A second interaction, found while building the read layer. Unrelated to the invocation session above — this one is about the **storage** session, which makes it the same mistake twice over in different clothes.

**Two identical reads through one daemon collide.**

The chain, all in code today:

1. `cf piece get`'s projection creates its transform result cell **session-scoped** — `runtime.getCell(space, { pieceGetTransform: { source, filter, schema } }, …, "session")` — so its identity is `{source, filter, schema}` within a session.
2. The session id lives on the **storage manager**: `#sessionId = crypto.randomUUID()` in `packages/runner/src/storage/v2.ts`, re-minted only in `close()` / `closeNow()`.
3. This spec retains the storage manager. Its Retained state list says so: *"the storage manager and its authenticated session identity."*

Two separate CLI processes each build their own storage manager, get their own session id, and never collide — which is why nobody has hit this. One daemon holding the manager across requests gives every request the *same* session id, so the second of two identical reads computes a cell that already exists and throws `updated arguments do not match the candidate schema`.

Observed rather than reasoned: an agent writing a test hit it calling the read twice in one runtime, and restructured to call once.

**A fresh runtime per request does not fix it.** Making the runtime request-owned — which this spec already specifies — would resolve a runtime-scoped collision. This one is keyed off retained storage state, so it survives exactly the teardown the lifecycle section describes.

**It generalizes past this one read.** Any session-scoped cell created per-request has this shape in a daemon. The projection transform is where we happened to find it, not necessarily the only one. That makes the useful question not "is this cell per-request?" but "what else is keyed off something the daemon holds across requests?"

Worth noting this is the second finding of one kind. The earlier comment was the *invocation* session being retained where callers assume per-request; this is the *storage* session, same assumption, different session. Two concepts named "session" both breaking the same way is some evidence the naming is doing real damage, beyond being confusing to read.

---

None of this asks for new scope. The lifecycle section already commits to it:

> The lifecycle work must also classify manager state that outlives a runtime … It must prove which state is valid for the fixed daemon tuple, what resets at a request boundary, and what remains bounded.

The session id is precisely such state: retained by construction, since it sits on the manager, but semantically per-request for anything scoped to it. Whether it should reset at a request boundary, or whether session-scoped cells should be keyed by something else, is the decision — and either way it belongs in that classification rather than being discovered later.

**Suggested validation line**, in the same spirit as the receipt-address one:

```
- identical results for the same read issued twice through one daemon;
```

That is a single assertion, it fails today, and it fails in the way that is hardest to catch by inspection — the caller sees a schema error from a read that worked a moment ago, with nothing in the request to explain it.

mpsalisbury:
Refreshed and rebased. It was **339 commits behind** `main`, and the world moved under it in two ways that mattered.

**1. The session flag it names does not exist.** The spec said `--session` and `CF_SESSION`; what shipped is `--invocation-session` and `CF_INVOCATION_SESSION`. A reader grepping either lands here and finds a flag that is not there. Corrected in all four places.

**2. The review's one real finding has dissolved, in the way it predicted.** The Status header deferred the invocation-identity rules to #5435, #5460 and #5469 — "coordinated intended behavior rather than current direct-execution behavior". #5435 and #5469 merged, #5460 closed, and the rules are in effect. Verified against `resolveInvocationIdentity` rather than inferred from the merge state:

```
neither supplied      -> id=ef66429c… session=c21fdbeb…      (mints both)
id without a session  -> REJECTED: "--invocation names an id to replay…"
both supplied         -> id=my-id     session=my-session
```

All three rules the spec describes are real. The header now says which part is already true and which part the daemon adds — resolved by the short-lived client before admission, carried as request-owned data, never read or minted or retained by the daemon process — rather than deferring the lot.

The review's second finding was already fixed in the branch: the direct-only list reads "the top-level `cf view`".

Gates clean: `check-docs`, `check-conflict-markers`, `check-skill-facts`, `fmt --check`.

**One thing I did not do.** The review's follow-up suggestion still stands and is still unfiled: the storage-subscription leak the spec describes is real today, independent of any daemon go/no-go. Both `Scheduler` and `Runner` call `storageManager.subscribe()` at construction and discard the return value, and nothing in disposal unregisters — so sequential runtimes over a retained manager keep disposed schedulers reachable. That is a live defect in `packages/runner`, not a daemon prerequisite, and it deserves its own issue rather than waiting on this document. Say the word and I will file it with the citations.

mpsalisbury:
## Resume checkpoint — 2026-08-21

This is the durable handoff for resuming the arc after the server transition launches. The transition is expected to change the daemon boundary; treat the current spec as design input, not as an implementation-ready contract.

Before taking this PR out of draft:

1. **Rebase and remeasure.** Measure the settled direct CLI under the launched server architecture, including compiled startup and warm repeated calls. Re-evaluate whether a daemon still earns its lifecycle and security cost, and which connection/runtime state is actually still client-owned.
2. **Reconcile the settled CLI surface.** The preferred data commands are now top-level `cf get`, `cf set`, and `cf call`; the piece-mounted spellings are being removed by #6004. Update examples and command buckets, classify newer Fabric-facing commands such as `piece slugs`, `piece describe`, and the CFC label operations, and incorporate whichever `inspect`/`view`/`map`/`apply`/space consolidations have landed or been explicitly deferred.
3. **Update the lifecycle baseline.** #5895 made `Runtime.dispose({ closeStorage: false })` unregister the scheduler and runner storage subscribers. Remove the spec claim that this machinery is missing, while retaining the broader proof obligation for all request-owned and process-global state.
4. **Refine the observation boundary.** Current shaped reads use a request-owned, host-defined projection graph (`runtime.run`, builtins, a transaction, and session-scoped result cells). Decide whether the daemon permits this trusted projection computation while continuing to forbid starting caller pieces or invoking user code, or excludes shaped reads from the observational MVP. The current blanket ban on all reactive/builtin execution no longer describes `cf get --select/--schema/--filter`.
5. **Bound retained storage-session artifacts.** Current main fixes the old duplicate-read failure and tests a second runtime issuing an identical transformed read over one retained storage session. Its per-runtime discriminator also gives fresh runtimes distinct session-scoped transform cells, so a long-lived manager still needs a proved cleanup, reuse, or growth bound across many daemon requests.
6. **Specify `cf call` waiting semantics.** Define how `--wait` expiry and `--no-wait` interact with daemon execution, queue ownership, client detachment, the in-memory result ledger, and a same-invocation retry while the original request is still active.
7. **Revalidate all retained-state assumptions.** In particular, revisit request identity, synchronization, outcome-unknown handling, staleness, and command concurrency against the server transition rather than carrying the present client-execution model forward by default.

Two older comments are now historical rather than current blockers: the storage-subscription leak was fixed by #5895, and the exact repeated-shaped-read collision is covered on current main. The bounded-lifetime question for the resulting session-scoped transform cells remains open as described above.

The invocation-identity names that shipped are `--invocation-session` and `CF_INVOCATION_SESSION`; the PR description and refreshed spec use those names.

mpsalisbury:
## Reassessment checkpoint — 2026-09-02

This remains **design input, not an implementation-ready contract**.

The hold condition has not cleared. Server-primary execution v2 has landed behind its explicit ON arm, but remains OFF by default. Its Phase 7 rollout/performance gates are still open, so daemon work should remain paused until that transition launches and its steady-state client/server boundary can be measured.

Material changes on `main` since this draft:

- The CLI-shape arc has landed, including #6004 and #6703. The command inventory and spellings in this spec must be rebuilt from the current noun-oriented CLI surface.
- Runner and scheduler storage-manager subscriptions now unregister during disposal. The spec's claim that this unsubscription is missing is stale; retained-manager lifecycle work must be re-audited from the current implementation.
- Server-primary execution changes the central architecture question. Once the server owns durable computation, the useful local component may be a thinner authenticated connection/transport broker—or no daemon at all—rather than the request-scoped local runtime host described here.
- The observation/execution boundary must be reconciled with current shaped-read and server-execution semantics rather than carried forward unchanged.

Before implementation resumes:

1. Rebase or recreate the design branch from current `main`; treat the remote PR branch as the preserved historical draft, not code to implement directly.
2. Repeat the Stage 1 measurements against the current compiled CLI and deployed server-execution posture. This remains a real go/no-go gate.
3. Reassess what state is still client-side and expensive enough to retain.
4. Reclassify the current command surface and newer wait/completion behavior.
5. Resolve or deliberately defer the remaining source-manifest, queue/input/result-ledger bounds, retained-cache bounds, cold-inspection seams, and read-parallelism questions.
6. Rewrite the spec as a current implementation contract and obtain review before beginning daemon code.

The durable product requirements remain useful inputs: explicit and visible daemon selection; no auto-start, silent fallback, or mutation replay; client-owned stdin collection; outcome-unknown handling; namespace compatibility; fail-closed staleness; and no retained live piece execution.

A new agent can resume the **design reassessment** from this PR and its comments without the original session context. It should not implement the daemon from the current document without first completing the steps above.

PR #5425, opened 2026-08-06
docs(perf): Topics performance research package + WAN-emulation testbed (CT-1954)

bfollington:
## Research package: why Topics interactions are slow, measured — plus the testbed to reproduce it

Anchor issue: [CT-1954](https://linear.app/common-tools/issue/CT-1954). Builds on the evidence snapshot in `docs/history/topics-performance-investigation-2026-08-06.md`.

### Why reading Topics is slow: a read is not a read

What actually happens when a cold client asks the board for anything (a verbs listing, a field):

1. **Session setup** — health + authenticated session open: ~2.5 s of fixed round trips before the first byte of data.
2. **Discovery is graph-shaped** — the board's `mentionable`/default-app surface expands one root into 229 documents / 318 paths, so the client downloads **8.1 MB** of documents and schemas to answer a question worth a few KB.
3. **…in ~20 dependent round-trip waves** — most sync roots carry no schema (`REJECTING_SELECTOR`), so the server cannot traverse the graph for the client; the client walks links itself, one wave's targets discovered from the previous wave's values. ~4.5 s at 190 ms RTT, and concurrency can't collapse dependent waves (that's why the flag measured no effect).
4. **Then it hashes everything it loaded** — canonicalize/deep-freeze/value-hash over the whole loaded graph is ~8 s of client CPU; with server traverse (1.3–2.7 s) that's the 12 s floor you pay even at zero latency.
5. **Then the read writes** — fresh-session materialization commits carry the entire observed read-set as ~52k per-path JSON records: an 8.8 MB envelope for a 7 KB no-op write, uploaded, rejected, and re-uploaded in full (~17 MB up).
6. On a real ~5 Mbps link, the resulting **28 MB of motion is ~47 s of pure transmission** — which is where the 60–110 s cold interactions come from. Bandwidth × payload dominates; RTT and server work are secondary.

The 2.5 GB file is the same mechanism accumulated: every interaction durably appends its read trace (~9 MB), operations are stored twice, and nothing garbage-collects — five weeks of that is 2.5 GB, ~55% of it read-set records. **Slow reads and the bloated space are two views of one behavior: reads are broad, and reads write.**

### The headline findings (full report in `docs/history/topics-performance-testbed-2026-08-06.md`)

- **The 2.5 GB space file is ~55% commit read-set records.** Every commit envelope stores the transaction's reads verbatim, one JSON record per leaf path. One startup commit is 8,828,447 bytes: a 7 KB session write towing 52,465 read records. Operations are additionally stored twice (commit + revision), there is no retention/GC, and the whole file is five weeks old.
- **One cold `cf piece verbs` moves 28 MB** (20 MB up / 8.1 MB down) — and the 8.46 MB read-set envelope is rejected and re-uploaded in full (8.83 MB), so it is paid twice. 2.74 MB of the upload is JSON Schemas inlined in `session.watch.add` selectors (one request is 2.01 MB). Every such read-shaped interaction durably appends ~9 MB to the space.
- **Measured cost ladder** on a rehearsal clone of the production snapshot behind a WAN-emulating proxy (190 ms RTT, ~4.8 Mbps — the Rapids link's own measured rate):

  | Condition | Wall time |
  |---|---:|
  | Direct to loopback toolshed | 12.4 s |
  | +190 ms RTT | 16.9 s |
  | +190 ms RTT, `experimentalConcurrentWatchRefresh` on | 16.6 s |
  | +190 ms RTT and ~4.8 Mbps cap | **64.2 s** |
  | Same capped link, commit read-sets stripped (probe, not landed) | **30.0 s** |

  Bandwidth × payload dominates on real links, not round trips; the ~20 sync waves are *dependent*, which is why the concurrency flag measured no effect (verdict recorded in `EXPERIMENTAL_OPTIONS.md`, decision tracked in CT-1959).

### What this PR contains

- `docs/history/topics-performance-testbed-2026-08-06.md` — the frozen findings report (storage anatomy + testbed measurements).
- `docs/plans/topics-performance-improvement.md` — the live staged plan toward few-second interactions, one Linear issue per stage.
- `scripts/delay-proxy.ts` — the WAN-emulation proxy (latency + bandwidth cap + per-connection byte totals) used for every number above; pairs with the space-clone rehearsal procedure so anyone can reproduce the ladder.
- CLI parity for `experimentalConcurrentWatchRefresh` (`EXPERIMENTAL_CONCURRENT_WATCH_REFRESH=true` in `cf` / `PiecesController`) — the flag was previously shell-console-only, so it could not be A/B tested from scripts.
- `CF_WS_SIZE_LOG=1` outbound message-size logging in the memory client (browser-safe guard), documented in `CONFIGURATION.md`.

### The improvement stack (separate PRs, sequenced by leverage)

1. [CT-1955](https://linear.app/common-tools/issue/CT-1955) Slim commit read-sets to per-entity claims (measured ceiling: 64 s → 30 s; startup commit 8.8 MB → tens of KB)
2. [CT-1956](https://linear.app/common-tools/issue/CT-1956) Watch selectors by schema reference
3. [CT-1957](https://linear.app/common-tools/issue/CT-1957) Ack-and-drop equal-value session materializations
4. [CT-1959](https://linear.app/common-tools/issue/CT-1959) Wave-count reduction + CLI session reuse; concurrency-flag decision
5. [CT-1958](https://linear.app/common-tools/issue/CT-1958) Compact default-app discovery projection
6. [CT-1960](https://linear.app/common-tools/issue/CT-1960) Storage diet + retention
7. [CT-1961](https://linear.app/common-tools/issue/CT-1961) Wish unloaded-vs-empty A→B→A churn


bfollington (commented):

Scope: PR #5425 · deep-read all 9 changed files plus the storage/session consumers and cited implementation · checks: targeted `fmt`, `lint`, `check`, `check-docs`, conflict-marker gate, 53 memory-client tests, CLI runtime-creation test, and focused no-env/UTF-8 probes.

## 🔴 Blocking

1. **Major · verified — `packages/memory/v2/client.ts:49-50`: the diagnostic makes the client unloadable in capability-restricted Deno processes.**  
   `typeof Deno !== "undefined"` is browser-safe, but it does not make `Deno.env.get` permission-safe. A plain `deno run` importing this module now fails at module evaluation with `NotCapable: Requires env access to "CF_WS_SIZE_LOG"`. This is a regression for least-privilege consumers unrelated to the opt-in diagnostic. Use the established safe-env shape (for example the try/catch in `memory/v2/sqlite/column-origin.ts`) and default to false when env access is absent.

2. **Major · verified — `packages/memory/v2/client.ts:278-282,343-352`: `CF_WS_SIZE_LOG` does not produce the byte-complete composition claimed by the code, CONFIGURATION, and history report.**  
   `encoded.length` counts UTF-16 code units, not UTF-8 wire bytes; a focused probe produced 44 vs 46 bytes for a payload containing `🌊`. Also, the `hello` message calls `transport.send` directly and bypasses this logging path, so “every outbound memory-protocol message” is false even for ASCII traffic. Use `TextEncoder` (or the codec’s byte encoder) and route every outbound send, including `hello`, through one logged helper. Then re-check the exact per-type figures in `topics-performance-testbed-2026-08-06.md:154-174`; they are currently character counts with at least one omitted message, not an auditable wire-byte decomposition.

3. **Major · verified — `docs/history/topics-performance-testbed-2026-08-06.md:16,128,211-219`: the report’s prerequisite evidence/provenance is not landed.**  
   It relies on `topics-performance-investigation-2026-08-06.md` for the snapshot manifest and the prior ~108 s measurement, but that file exists in neither `main` nor this PR. As written, readers cannot follow the links, identify the board fid from the claimed manifest, or reproduce/verify the report’s starting evidence. Land that report in this changeset or make this report self-contained with the missing provenance. Add the new historical artifact(s) to `docs/history/README.md`’s one-line-per-document index as required there.

4. **Major · verified — `docs/development/CONFIGURATION.md:265-275`: the touched live configuration guide now contradicts the experimental registry and this PR.**  
   It says its three-row table contains “the only” environment-backed experimental flags, but the canonical mapping already has six runtime env flags, and this PR adds `EXPERIMENTAL_CONCURRENT_WATCH_REFRESH` as a storage/CLI env toggle without adding it there. Update the table/wording from `EXPERIMENTAL_ENV_VARS` and include the new scoped storage toggle, so readers are not directed from an accurate registry into an actively false exhaustive list.

## 🟡 Improvement

- **Major · verified — `scripts/delay-proxy.ts:38-61`: capped mode removes network backpressure and builds an unbounded copied-chunk backlog.**  
  The read loop keeps consuming loopback input while appending delayed writes to `chain`; a fast producer can therefore be fully buffered in the proxy (up to the response/stream size) instead of being backpressured as it would be by a constrained link. The current ~28 MB workload is finite, but this can distort producer timing and can OOM on large transfers. Bound queued bytes/chunks while preserving the single path-delay schedule; do not simply add one `DELAY_MS` sleep per chunk.

## Confirmed clean in the requested focus areas

- The `typeof Deno` short-circuit itself does keep the new env lookup off in browser workers; the issue is Deno-without-env permission, not browser evaluation.
- With `EXPERIMENTAL_CONCURRENT_WATCH_REFRESH` unset, both CLI paths pass explicit `false`. The only storage consumers test `=== true`, so this is behaviorally identical to the old omitted/`undefined` setting: session concurrency stays false and the in-flight window stays 1.
- The `EXPERIMENTAL_OPTIONS.md` toggle path, default-off status, window of 8, and workload-scoped “no measured effect” wording match the landed plumbing and cited timings. Its surrounding CONFIGURATION summary is the coherence defect above.

**Verdict:** request changes; the core performance conclusion may still be right, but the PR currently lands a capability regression, incomplete/mislabelled byte instrumentation, and a historical report whose prerequisite evidence is missing.

bfollington (commented):

Scope: PR #5425 at `9a31f02a` · incremental deep-read of the fix commit and every behavioral/documentation delta, with the newly landed history reports checked for provenance and link/index coherence · checks: Deno 2.9.4 targeted `fmt`, `lint`, and `check`; `check-docs` (533 blocks); conflict-marker gate; history-aware docs orphan scan; 53 memory-client tests; CLI runtime-creation test (3 steps); no-env and UTF-8/hello logging probes; capped and uncapped proxy throughput probes.

## 🔴 Blocking

1. **Major · verified — `scripts/delay-proxy.ts:55-64`: the bounded queue inserts another `DELAY_MS` after every 256 KiB and accidentally caps uncapped mode.**  
   At the high-water mark, `await chain` waits for *all* queued sends to finish, so the queue drains to zero before the next read. That next chunk is then stamped with `performance.now() + DELAY_MS`, creating a path-latency-sized hole once per backlog window. A 4 MiB loopback probe at 600,000 B/s and 95 ms one-way delay should take about 7,181 ms including both path delays; it took 8,652 ms. The 1,471 ms excess is almost exactly fifteen extra 95 ms gaps. More seriously, with `BYTES_PER_SEC=0`, the same 4 MiB transfer took 1,660 ms rather than roughly the configured 190 ms path delay, so the documented uncapped default now behaves like an implicit ~22 Mbps cap. This makes the testbed systematically mismeasure the workload it exists to reproduce. Resume reading when the oldest queued chunk frees capacity (or at a low-water signal) instead of awaiting the aggregate tail, preserve enough queued data for the bandwidth-delay product, and do not apply capped-mode backpressure to the uncapped path. Add a deterministic scheduler regression covering several high-water windows.

## 🟡 Improvements

- **Minor · verified — `docs/history/topics-performance-investigation-2026-08-06.md:521-522`: one stale provenance path remains.**  
  The report correctly links the landed plan near the top, but its artifacts section still names the nonexistent `docs/plans/topics-performance-investigation.md`. Repoint this mechanical historical-doc link to `docs/plans/topics-performance-improvement.md`; the history-aware orphan scan cannot catch it because it is code-formatted rather than a Markdown link.

## Confirmed fixed from the prior review

- `CF_WS_SIZE_LOG` is permission-safe: `deno run --no-prompt packages/memory/v2/client.ts` succeeds with no env grant.
- One send helper now covers `hello` and requests, and reports UTF-8 bytes: the probe logged `hello 315` and a Unicode payload as 57 bytes versus 54 UTF-16 code units.
- Both prerequisite history reports are present with metadata and all three reports are indexed; none appears in the history-aware orphan scan.
- `CONFIGURATION.md` now matches all six entries in `EXPERIMENTAL_ENV_VARS` and separately documents the CLI/PiecesController storage toggle.

**Verdict:** request changes for the proxy scheduler. The four original blocking findings are resolved; the backpressure repair currently changes the measurement semantics it was meant to preserve.

bfollington (commented):

Scope: PR #5425 at `7355df1fa` · round-three incremental deep-read of the scheduler extraction, socket-pump integration, deterministic regression, and provenance repoint · checks: Deno 2.9.4 scripts-package tests (4 tests / 6 scheduler steps), targeted `check`, `fmt`, `lint`, and diff check; repository-wide stale-path search; exact capped/uncapped 4 MiB TCP probes plus a full-RTT HTTP probe; focused rejected-write probe against the committed pump.

## 🔴 Blocking

1. **Major · verified — `scripts/delay-proxy.ts:101-123`: a destination write failure while backpressured can never wake the pump.**
   At high water the read loop awaits a promise whose only resolver lives after a successful `to.write`. If that write rejects, `scheduler.complete()` and the resolver are skipped, `chain` rejects, and the outer `try/catch` cannot observe it because the read loop is still awaiting the orphaned resume promise. The connection task therefore never settles; the rejected chain is also unhandled and can terminate the proxy process. A fake-`Deno.Conn` probe against this source (instrumented only to expose `pump`) entered pause at a 3-byte high-water mark, rejected `to.write`, and remained `settled: false`; without suppressing the diagnostic's unhandled-rejection event, Deno exited with the injected error.

   Make the pause wait observe writer failure as well as low-water progress—for example, race the resume signal against a rejection-observing writer/chain signal—so a rejection reaches the existing outer teardown path. Attach that error observation when the write chain is created, not only once the reader happens to pause, and add a deterministic fake-connection regression proving that a rejected write settles the pump and half-closes instead of hanging.

## 🟡 Improvement

- **Minor · verified — the round-three commit's `7,181 ms` theoretical comparator describes a different probe shape.** The reported `7,089 ms` / `106 ms` figures are reproducible for the one-way server-push probe: I measured `7,090.2 ms` capped and `102.3 ms` uncapped. For that probe the theory is `4 MiB / 600,000 B/s + 95 ms = 7,085.5 ms`, because the proxy logged zero upstream bytes and only the downstream propagation delay applies. `7,180.5 ms` is the full request-plus-response theory; a separate HTTP probe measured `7,197 ms`. Clarify the provenance in the PR description or commit history so the observed result is compared with the matching model.

## Confirmed fixed from round two

- `LinkScheduler` preserves capped serialization across repeated high/low-water pause windows and restarts correctly after genuine idle time.
- Uncapped chunks retain their own arrival-plus-propagation schedule; the old implicit ~22 Mbps ceiling is gone.
- The scheduler regression is deterministic and meaningful: no sockets, timers, sleeps, or wall-clock assertions; all six BDD steps pass, as does the complete `scripts/` package suite.
- The stale `docs/plans/topics-performance-investigation.md` path is absent from the tree, and the historical artifact now points to the landed improvement plan.

**Verdict:** request changes for writer-failure propagation in the new pause path. The successful-transfer scheduler fix itself is verified.

bfollington (commented):

Scope: PR #5425 at 4bb33f9d3 · round-four incremental deep-read of the pump failure-path implementation and fake-connection regression · checks: repository-pinned Deno 2.9.4 scripts-package suite (4 tests / 9 BDD steps), targeted check/fmt/lint/diff check, deterministic failure-order trace, producer-consumption/half-close probe, and fresh capped/uncapped 4 MiB TCP probes.

## 🟡 Improvement

- **Minor · verified — scripts/delay-proxy.test.ts:45-46: the regression does not pin the advertised producer-stop behavior.** The five-chunk fake producer is useful, but counter.n >= 1 still passes if a future change wakes the paused reader, drains the queue, and then consumes all five chunks despite writeFailed. The current implementation is correct—my deterministic trace read exactly two of five chunks before the failure stopped the loop—but make that invariant durable with an upper-bound assertion such as counter.n < 5 (or the deterministic exact value, 2). Counting closeWrite would also directly preserve the teardown half-close.

## Confirmed fixed from round three

- The write-chain callback catches the rejected destination write, flips writeFailed, and resolves; subsequent queued callbacks skip delay/write, call scheduler.complete(), and the aggregate chain resolves without an unhandled rejection.
- At the two-byte high-water mark, the trace was: read two chunks → first write rejects → reader is paused → both queued completions run → closeWrite fires. queuedBytes returns to zero and the remaining three source chunks are not consumed.
- The successful-transfer scheduler behavior is unchanged. Fresh one-way 4 MiB probes measured 7,091.7 ms at 600,000 B/s + 95 ms and 103.4 ms uncapped, matching the reported 7,089 ms / 105 ms within normal run variance.
- The committed regression is deterministic: fake connections, zero propagation delay, no sockets, timers, sleeps, or wall-clock assertions. The full scripts package suite and targeted static/style checks pass under the pinned runtime.

**Verdict:** mergeable; the round-three writer-failure blocker is fixed. The remaining note is non-blocking test hardening.

bfollington:
We've since done many investigations on topics and rolled many of these findings in, so this is now deprecated slash not relevant anymore.

PR #5433, opened 2026-08-06
docs: post-merge feedback on the code comment style guide

danfuzz:
Two things @Hixie pointed out on #5384 after it landed.

**"Not the neighbors" opened by contradicting an earlier section.** It began
"a comment says what the code it sits next to does", where "what earns a
comment" a few sections up says the opposite and means it: the code already
says what it does, and a comment earns its keep by carrying what the code
cannot. The sentence was making a point about locality rather than about
restatement, so it now says a comment is _about_ the code it sits next to,
and the paragraph is rewrapped around it.

**The grep read as canonical.** "The check that finds these" ended with a list
of words and nothing about its standing, which invites a reader to treat the
list as the rule and a clean grep as a pass. It is a starting point weighted
toward the past-tense shapes, and there is nothing in it that would catch a
rollout stage, a count of callers, or a complaint about a neighbor — the three
sections that precede it. The guide now says that, and closes by asking that a
phrasing which gets past the list be added to it, which is the only way a list
like this stays worth running.

## Verification

`deno fmt --check`, `deno task check-docs`, `check-conflict-markers`,
`check-skill-facts`, and `docs-links --orphan` all pass. The 80-column pass was
done by character count rather than by `deno fmt`, which excludes `docs/`.
Documentation only; no code changed.

https://claude.ai/code/session_01HoP1WGxYiGH8EafTU4C6oG

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5433?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Hixie (approved):
LGTM

PR #5435, opened 2026-08-06
docs(plan): sequence the read layer and the calls layered on it

mpsalisbury:
## Why

[Reading Fabric data](https://github.com/commontoolsinc/labs/pull/5316) designs one model across three concerns. This sequences the two it builds: the shared read layer, and the calls layered on it. The design says what should be true; this says in what order, and what each step is judged on.

The ordering is the part worth reviewing. Scoping invocation ids to a session changes where a receipt lives, so it lands **before** anything hands a caller a receipt address — no caller ever holds one that later moves. That single constraint decides most of the sequence.

Stacked on #5316.

## What's in it

Five stages. Foundation (factor the read step out, split `--select` from `--schema`, check projection keys against an allowlist), addresses in results (`$link`, the rejecting selector, the `@` suffix, entity URIs, and propagating a rejection upward), session identity, calls, and the remaining arrivals.

Both questions this work once waited on are now decided and recorded in the design: receipts carry a descriptive schema, and invocation ids are scoped to a session rather than a principal.

## Two items came out of review, and they're the interesting ones

**F3 — an unrecognized projection key is refused, not ignored.** Today two denylists are consulted and every key in neither is silently accepted. So `{"propertys":{…}}` returns `{}` and says nothing. That is the *general* form of a bug this arc has already fixed twice as instances: a missing `type` returned `{}`, and an untyped `items` root returned `{}` while committing unprojected data. Three failures, one root cause.

The fix is three tiers rather than two lists — honored, tolerated, refused — where the tolerated tier is forced by the design itself: callers are told to lift a source schema and prune it, and a lifted schema is full of annotation keywords.

**A5 — a rejection has to propagate up through what holds it.** `projectionMask` reduces an array whose items reject to `false`, but leaves an object whose every property rejects as a live selector. So marking a field *below* a link loads every element document — measured at four reads where one would do. The addresses are correct and cost nothing; the read simply isn't free the way a marker on the link itself is.

## Naming

The session's flag, environment variable and minting command are all qualified — `--invocation-session`, `CF_INVOCATION_SESSION`, `cf invocation-session new` — because `cf inspect` already takes `--session` for a storage session. One word for two concepts in one binary is what this arc spent its design documents untangling elsewhere. `cf inspect`'s could become `--storage-session` later; that's a forensics tool with its own callers and is not part of this work.

The session is also closer to a secret than a setting: its unguessability is what keeps an outcome's address out of reach of anyone who can guess a piece, a verb, and a conventional id. A flag value is visible in process listings where an environment variable is not, so the environment form is what the docs teach.

## Test plan

Documentation only. `deno fmt --check`, `deno task check-docs` (533 blocks), `check-conflict-markers`, `docs-links --orphan` — all pass.

### DIFF
--- a/docs/plans/README.md
+++ b/docs/plans/README.md
@@ -61,4 +61,6 @@ a record: archive it to `docs/history/plans/` following the procedure in
 - [Verb calls: working notes](verb-result-selection.md) holds the call-specific
   investigation those documents do not carry: what produces a receipt and what
   its existence proves, how a receipt's address is derived, and the error and
-  exit-status conventions. Sketches to draw from, not a settled contract.
+  exit-status conventions. Sketches to draw from, not a settled contract.- [Shaped reads and verb results — implementation plan](shaped-reads-implementation.md)
+  sequences the read layer and the calls layered on it, holding the two
+  undecided questions to a final workstream so the rest can proceed.
--- a/docs/plans/README.md
+++ b/docs/plans/README.md
@@ -61,6 +61,8 @@ a record: archive it to `docs/history/plans/` following the procedure in
 - [Verb calls: working notes](verb-result-selection.md) holds the call-specific
   investigation those documents do not carry: what produces a receipt and what
   its existence proves, how a receipt's address is derived, and the error and
-  exit-status conventions. Sketches to draw from, not a settled contract.- [Shaped reads and verb results — implementation plan](shaped-reads-implementation.md)
-  sequences the read layer and the calls layered on it, holding the two
-  undecided questions to a final workstream so the rest can proceed.
+  exit-status conventions. Sketches to draw from, not a settled contract.
+- [Shaped reads and verb results — implementation plan](shaped-reads-implementation.md)
+  sequences the read layer, the calls layered on it, and the remaining
+  arrivals, ordering the address change ahead of anything that publishes an
+  address so no caller holds one that moves.
--- a/docs/plans/shaped-reads-implementation.md
+++ b/docs/plans/shaped-reads-implementation.md
@@ -147,10 +147,36 @@ Three things it has to satisfy, and they pull against each other:
   session it belongs to; it arrives as a flag or environment variable, following
   `CF_IDENTITY` and `CF_API_URL`.
 
+*Minted explicitly, by `cf session new`.* The session is a **namespace, not an
+entity**: it is only ever hashed into the event id, so nothing stores it,
+registers it, or expires it. Minting is a random opaque string on stdout, which
+makes it cheaper than `cf id new` — no key generation — and it follows the shape
+that command already established:
+
+```bash
+cf session new > cf.session
+export CF_SESSION=$(cat cf.session)
+cf piece call --invocation add-1 ...
+```
+
+Explicit rather than implicit, for two reasons. A session managed automatically
+is one a caller cannot answer questions about — *which session am I in, and is
+this retry landing where the last one did?* — and the two obvious automatic
+schemes each break something: minting per process destroys same-id replay, and
+persisting to a dotfile puts the answer somewhere the caller never looks.
+
+The second reason is stronger, and it is why a minted session beats any derived
+one. The exposure has two halves: two callers colliding on an id, and a third
+party *guessing* an id to read someone's outcome. A scope derived from identity
+closes only the first, because a DID is public — an address stays computable by
+anyone holding the piece, the verb, and a conventional id. A minted random
+session closes both, because it cannot be guessed.
+
 *The design question inside it* is what an absent session means. Treating it as
 unscoped preserves today's behavior and today's collision; refusing without one
-makes every caller opt in before they can name an invocation at all. This plan
-does not decide that.
+makes every caller opt in before they can name an invocation at all. Given the
+minting is explicit, the reasonable default is unscoped-with-a-documented-gap
+rather than an implicit mint, but this plan does not decide it.
 
 Address-changing; see Risks.
 
--- a/docs/plans/shaped-reads-implementation.md
+++ b/docs/plans/shaped-reads-implementation.md
@@ -9,7 +9,7 @@ The command surface — the third concern in that umbrella,
 [CLI surface shape](cli-surface-shape.md) — is out of scope here and moves on
 its own timeline. One piece of it is unavoidable: the concise selection syntax
 needs its own flag before it can grow address notation, so `--select` lands in
-WS-A rather than waiting.
+the first stage rather than waiting.
 
 ## Governing decisions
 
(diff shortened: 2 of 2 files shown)


mpsalisbury (commented):

**Scope:** `shaped-reads-implementation.md`, read against its design (#5316) and the stacked #5503. **Coverage:** deep-read; claims verified against `packages/runner` and `packages/cli` by read-only search agents. **Checks:** no CI runs on this PR (stacked base) — the local `check-docs` results in the description are the only gate until restack.

The sequencing logic is right and honestly accounted: S2 before C3 is the ordering that matters, it lands alone, and the deploy-window replay risk is named rather than hidden. Findings below are fixable in text.

## 🟡 Improvements

**A1's premise is wrong: `$link` is silently ignored today, not "refused today alongside the treatment keywords."** *Verified.* `$link` is in neither `FORBIDDEN_PROJECTION_KEYS` nor `UNSUPPORTED_PROJECTION_KEYS` (`packages/cli/lib/piece-get-transform.ts:462-483`), and the key checks only run on the full-schema branches anyway — the concise branch never calls `normalizeProjectionSchema` (details on #5316, where the same claim appears in the design). Ignore→meaningful is a different compatibility story than refuse→meaningful: an existing schema carrying the key changes behavior with no error ever having said so. Consider reserving `$link` in `UNSUPPORTED_PROJECTION_KEYS` as prep, so A1's rollout is refuse→meaningful.

**C1 omits the design's plain-results-only bound — and the flagship case falls in the gap.** The design's open-questions section states that a receipt for a verb returning anything reactive — which includes every verb returning a child piece — carries no schema at all, because the descriptive derivation runs only where the result is plain. C1 says the schema is "derived from the value the runtime has just written" and "lands before C2 so a selection over a receipt is bounded from the first call that can express one," unqualified. But the walkthrough's own fixture (`verb-results.tsx`'s `createNote`, returning a child piece) is exactly the reactive case that gets no schema and no narrowing. An implementer reading C1 and the Stage 4 exit will expect the bound to engage on the walkthrough case; the design says it won't. Carry the bound into C1 and the exit. *Verified* (both texts; the fixture returns a child).

**`--session` already exists on `cf inspect`, meaning a different session.** *Verified:* `cf inspect` takes `--session` on two subcommands as a storage-session filter (`packages/cli/commands/inspect.ts:631, 1142`). S1 introduces `--session` meaning the new caller session — two session concepts under one flag name in one binary, in a design that already had to disambiguate "neither session identifier already in the tree serves." Either pick a different name or note the collision deliberately. Relatedly: the session's unguessability makes it a capability — `--session` on argv is visible in process listings where `CF_SESSION` is not. One sentence ("treat it as a secret; prefer the env var") would finish the thought the Risks section starts with "something a harness persists rather than something a human types."

**Stage 1's "Nothing here is visible to a caller" still contradicts F2.** F2 adds the `--select` flag — new caller-visible surface; the exit even says "both flags work." Cubic flagged this and the thread was resolved with no reply and no text change, so the contradiction is still in the document. If the dismissal was intentional it deserves the one-line reply; otherwise scope the sentence to F1.

**The supplied-id/session table omits a row.** Session-without-invocation is undefined by the three-row table. Concurring with cubic's open thread — the table exists to remove exactly this ambiguity.

## ⚪ Nits

- The PR description describes a previous shape of the plan (WS-A…D, "two of its questions are still open") that the current document contradicts ("Both questions this work once waited on are decided"). Worth refreshing before more reviewers read it.
- S2's claim about what unscoped matching breaks checked out: `verbs-over-the-cli.sh` replays one id to assert deduplication, and `integration.sh`'s retry function has four same-id replay scenarios (of six total). *Verified.*

## Verdict

The plan's spine — one read step, addresses marked not inferred, the address change landing before addresses are published — is coherent with the design and with the code as it stands. Fix A1's premise and C1's missing bound, settle the session flag name, and close the two threads the text currently contradicts or leaves open.

mpsalisbury (commented):
cf-review follow-up via Claude Fable 5, on behalf of @mpsalisbury — re-review of the response batch (through `def696c4`).

Every finding from the earlier review is addressed, mostly by strengthening rather than patching:

- **A1** no longer claims `$link` is refused today; the ignore→meaningful account is honest, and the at-risk-population-is-empty argument is correct (lifting a source schema cannot produce `$link`).
- **F3** (new) is the right generalization of the reserve-one-key suggestion — the tolerated tier is correctly forced by the lift-and-prune story. Checked its exit against `packages/cli/test/piece-get-transform.test.ts`: nothing there pins unknown-key behavior, so "passes unchanged" is coherent alongside F3's new refusals.
- **A5** (new) is a real mechanism finding, and it retroactively substantiates the suppress-fetch claims in C1 and the design: link-following is a traversal-time decision made with the parent document in hand, so a rejection needs no declared root kind — which is exactly why it works on a schema-less receipt where field-narrowing cannot.
- The session flag rename, the secret-vs-setting paragraph, the completed supplied-id table, the Stage-1 sentence, and "(of six)" are all verified in.

**Pushed one alignment commit on the author's behalf** (`6e75b61`): S1 qualified the flag and env var but still minted through unqualified `cf session new` — the one remaining spelling of the ambiguity the adjacent paragraph argues against. It is now `cf invocation-session new`, matching `--invocation-session` and `CF_INVOCATION_SESSION`.

Two housekeeping items:

- **The PR description is two rewrites behind the document.** It still describes WS-A…D and says "two of its questions are still open," which the plan now contradicts ("Both questions this work once waited on are decided").
- **One cubic thread is still open** (F1 lacking its own exit criterion). With F3 added, the Stage 1 exit now covers three items jointly — one of them a deliberate behavior change — so splitting F1's refactor guard ("nothing moves") from F3's new expectations is more worth doing now, not less.

mpsalisbury (commented):
cf-review follow-up via Claude Fable 5, on behalf of @mpsalisbury — one finding on the exit-split commit (`7d50160`).

The split itself is right: F1's "passes **untouched** — not one expectation added, moved, or removed" is the refactor guard stated cleanly, and F2's exit records the both-flags-refused rule where it belongs.

**F3's exit overclaims compatibility, and its own motivation paragraph is the counterexample.** The exit promises "nothing that passed before it may fail after, since every key that works today is either honored or tolerated." But the paragraph introducing F3 names `propertys` — a typo that today "selects nothing and says nothing," i.e. runs and exits zero. After F3 it is refused. An invocation that passed before does fail after; that is F3's entire value, so the exit should say so instead of promising it away: something like "no key that is honored or legitimately carried changes meaning; a schema that ran only because a key was silently dropped now fails loudly, which is the point."

This matters one document over as well: #5503's "additive with one exception" ledger is correct only under the exit's current wording. Under the honest wording, F3 is a second (small, defensible) behavior change — fail-loudly on schemas that were already silently broken — and the ledger should count it as such or exempt it explicitly.

PR #5439, opened 2026-08-06
feat: server-execution v2 stage F — the serving loop: host, SpaceServer, watermark, gates

seefeldb:
## Why

Phase 1's stages A–E built the serving loop's parts and left them dark: the flag and CI arms (A, #5339), the lease table and cycle with nothing driving its cadence (B, #5349), the main reduction (C.1–C.3, #5356/#5367/#5369), seal-into-wave with "nothing installs this until stage F" written on it (D, #5371), and the per-instance key vocabulary (E, #5374 — the prior PR in this train). Stage F is the loop itself: without it there is no committing runtime per space, no watermark, nothing that ever drives a wave in production, and the ON CI arm tests a posture no code exercises. This PR lands the serving loop — still dark (flag OFF is byte-identical; the plan's Phase 2 owns the first ON milestone) — so the flag-ON arm starts *actually serving*, and every later phase (speculation, events, effects) builds on machinery that runs, not machinery that could.

Plan: `docs/plans/server-execution-v2.md` Phase 1 stage F (checkbox ticked in this PR).

## What

**The loop** (`packages/runner/src/executor/`): `ExecutorHost` (one per process — the admission-side activation observer (plane b), one `SpaceServer` per ACTIVE space, per-space process-lifetime localSeq counters, the ambient-flag lifecycle, §7 stats) and `SpaceServer` (the lease renewed on stage B's cadence with abort-before-reacquire intact; the serving Runtime over a loopback storage plane with production-signed `session.open`; rotating wave accumulators behind a stable seal-destination dispatcher — commits arriving mid-wave open the next wave; the wave cycle: drain the feed → settle to quiescence with the settle *coupled to frame delivery* and RACED against the consequence-flush deadline (serving-loop §3's second exhaustion trigger) → ONE derived commit carrying the wave's writes, the watermark-doc advance as a stamped `bookkeeping` contribution, and `derivedThrough`; self-echo skipped by class + holder; §6's activation re-mark scan; idle park honoring armed gate wakes (N9), with parks racing sessions healed via `whenParked` chaining). Demand is a live reader: the server's watch registry (client sessions only) maps to `ensurePieceRunning` + a deep-walking demand sink per demanded root, retired when the watch goes — pull-based laziness serves exactly what is demanded (RULED: no piece-start policy).

**M1 (scopes §5/§7)**: run contexts carry per-run `scopeKeyIdentity`; the accumulator resolves every scoped read/write/basis row against the RUN's identity, wave-level identity as the cardinality-1 fallback; the scheduler stamps at its two choke points (reactive run, event dispatch) plus the pattern-swap setup write. **The read row (protocol §2, LD5)**: `GraphQueryRoot.entityScopeKey` — the one wire addition — admissible only for a live lease holder (service-identity component of the DR1 holder vs the session principal), refused for anyone else and off the flag, honored through one shared root-key construction; the no-key path adds no await (the ACL same-turn invariant holds).

**M4 (protocol §3)**: dirtiness AND delivery key by scope INSTANCE end to end (`toDirtyKey`, session caches, tracked-id sets), with wire frames byte-identical (`toWireUpsert` strips the server-internal key in the exact legacy field order) and protocol §3's applicable-set filter wired as defense-in-depth (lease-holder sessions exempt). The M4-coupled and serving-identity boundary sites re-keyed with it (selector tracker, pull kicks, missing-doc-load kicks, wake-shaping `linkKey`, pieceId buckets) — key-vocabulary §5's dispositions moved in the same change per its own tripwire.

**Gates (RULED 2026-08-05, both landed)**: the derived-envelope check — a `derived` commit's producing session must BE the holder's own service session; the designed operand mapping is `resolveCommitSessionKey(sessionId, principal) == holder` (the sink commits under the holder identity, which also settles the sink's replay-keying choice: one counter per (space, process), session freshness structural via the DR1 process component) — and the `bookkeeping` stamp kind named in serving-loop §3d with both conflict dispositions live and tested (the watermark advance is a key-path PATCH so the rebase arm is real).

**Also**: the eager via-user hop (scopes §2's MUST) at all three narrowing write sites, flag-gated with the OFF arm pinned one-hop; the watermark surface (`SERVER_EXECUTION_WATERMARK_DOC_ID`, `derivedThrough` column + carriage, `waitForSettled` — testing §3's poll replacement); the schema-memo single-identity guard (OW10); both dischargeable stage-D bounds discharged (foreign provisioning applies WITH protocol §2's delegated row — carried actor + capabilityRef validated, scoped writes keyed from the CARRIED identity, batches grouped per (space, actor, grant); the seal handoff carries read-only spaces' read sets and the withdrawal closure folds them by doc identity, conservatively); §3e's pattern-update posture flipped server-side with the hot-swap verified end to end; toolshed wiring (`startServerExecutionHost` under the env flag; OFF startup byte-identical) and the §7 `servingLoop` block on `/api/health/stats`; T13 (activation/demand/park/hot-swap) authored and OW5/OW6/OW10 flipped, OW11 (the N50 ruling) and OW12 carried.

**ON-arm skip list** (`tasks/server-execution-on-skips.ts`) — one entry:
- `patterns/integration/cfc-group-chat-demo-two-browsers.test.ts` → phase-2: two real browser shells boot against a toolshed whose serving loop also derives, and the trusted-profile UI never becomes fillable under the CAS storming the plan documents as expected until Phase 2 removes the client derivation-commit path (this exact test is a named Phase 2 gate). The single-browser and multi-runtime variants pass ON.

## Verification

- `packages/spec-model`: 19 tests green (the model stays the oracle; the envelope check mirrors `admitDerived`, the loop's lease handling mirrors C6/C7a/C7b, exhaustion mirrors C10).
- Full runner suite: **1114 tests / 6014 steps green** (stage-D baseline 1109/5991). New end-to-end coverage: `executor-serving-loop.test.ts` (authored commit → wave → ONE derived commit with holder + `derivedThrough` + the watermark doc in the SAME commit → client observes via push and `waitForSettled`, with the W-soundness contract bound STRICTLY — settled ⇒ derived value already current on a synchronous read; the sealed recompute path lands derived-class; self-echo stability; lease-loss park; idle park releasing the lease; the server-side hot-swap serving v2 with no client run), `executor-wave.test.ts` grown to 25 steps (delegated carried-identity keying + partial-carriage refusal, read-only-read withdrawal fold, bookkeeping rebase AND drop dispositions, M1 per-run instance keying with real basis assertions), `eager-via-user-hop.test.ts` (3 tests, both arms).
- Full memory suite: **460 tests / 145 steps green**. New: the derived-envelope negatives (right holder + wrong session refused, holder session admitted), `derivedThrough` carriage + closed-list refusal, four engine-level delegated-admission negatives + stored-trio, the explicit-read row (holder admitted / non-holder refused on query + watch.set + watch.add / expired lease refused / off-flag refused / key-less reads unchanged), instance-keyed sync-cache + wire-stripping tests.
- `tasks/check.sh` (repo typecheck), `deno lint` (4316 files), `deno fmt --check`, `check-docs` (522 blocks), `check-no-waitfor`, `check-conflict-markers`, `check-skill-facts`, and the skip-list unit tests (10) all green.
- Integration, live: runner suite **both arms** green (flag-OFF and flag-ON toolshed — the ON arm with the ExecutorHost actually serving); runtime-client both arms green; patterns `counter`, `lunch`-family, and `cfc-group-chat-demo-multi-runtime` green ON. The two-browsers variant is the one skip (above).
- OFF arm: the full runner + memory suites are the byte-identity witness (every OFF-arm path the re-keys touch is partition-equivalent at cardinality 1; wire frames byte-identical by construction and pinned by test).

## Flags

Per the flag-don't-fill rule, conservative implementations whose exact semantics the specs leave unstated — each needs an owner look, none blocks the dark landing:

1. **The settle's input barrier excludes update-processing parked behind sealed commits** (`StorageManager.inputSynced`, documented at the definition): awaiting it deadlocks against the wave commit by construction. Residual: an inbound frame whose application parks behind one of the loop's own sealed commits settles after the wave commit, so a foreign authored frame in exactly that position can be claimed by W one wave early (self-healing next wave). Is that residual acceptable Phase-1/2 semantics for W, or must the barrier distinguish parked-on-own-seal from foreign novelty before Phase 2's gates?
2. **A watermark-only derived commit** exists for input batches with no demanded work (nobody subscribes, someone writes): W must still advance for `waitForSettled`, so the wave's derived commit carries only the doc advance. Reading of protocol §4's "never its own commit": I read it as "no separate commit ALONGSIDE a wave's derived commit" — an advance-only wave IS the wave's derived commit. Confirm or rule otherwise.
3. **M4's cross-session wake-partition change is OFF-arm-visible at the byte level under multi-principal scoped workloads** (recorded in verification-coverage.md): the name-keyed form spuriously re-evaluated principal B's session on principal A's scoped commit; the instance-keyed form doesn't, so a later frame's `fromSeq` can differ. No client consumes server-frame `fromSeq`, and the removed wake is the M4 defect itself — but testing §2's OFF-gate language is absolute. Needs the acceptance ratified (or a directive to gate the re-key on the flag, which would fork the dirty-key vocabulary).
4. **The read row's FP2 widening is implemented at its Phase-1 bound** (noted in protocol §2): the check consults the READ space's own live lease; the widened cross-space acceptance has no producer until Phase 5 and needs the cross-engine lease lookup designed with it.
5. **The pattern updater's CHECK half under the serving posture**: toolshed's serving-runtime factory enables `systemPatternAutoUpdate` (§3e's flip), but in a fully-local fixture the check's source probe syncs docs that never resolve and would wedge the settle (which now surfaces as exhausted flushes, not a hang). The swap half is verified end to end; the check half needs its bring-up against a real patterns route — where should that verification live (the integration env, or a Phase 2 gate)?
6. **Demanded reads naming explicit instances**: the wire + admission + root-level honoring are landed; deep per-run traversal threading (a run's every read resolving against its demanded instance) is the Phase 2 fan-out work. A query naming explicit instances should name one instance per doc per query — the response's `EntitySnapshot` carries no `entity_scope_key` echo, and adding one is a protocol §7 closed-list edit I did not make. Flag: should the snapshot echo the instance for holder reads when fan-out lands?
7. **Structure-load failures and W**: a demanded root whose piece fails to load server-side is counted and logged, and W still advances (a wedge would otherwise be permanent). Under §4's letter a failed-to-load demanded derivation is arguably "not current". Phase 2's load-path hardening should revisit which way this falls.

## Self-review (standing merge protocol)

An adversarial subagent review of the full diff ran after green (spec conformance section by section, OFF-arm neutrality, lease fencing vs the model, test vacuity; report at the worktree's `stage-f-review-report.md` channel). It found 1 BLOCKER, 6 MAJOR, 13 MINOR; **all twenty are addressed** in the follow-up commit (`fix+test+docs: address the stage-F adversarial review batch`):

- **BLOCKER (empirically confirmed by the reviewer): W advanced before the batch's demanded derivations were current** — the settle never coupled to frame delivery, so a watermark-only commit could claim coverage while the derivation's dirtiness sat in the server's flush queue. Fixed: the settle drains the co-hosted server's refresh queue and yields for delivery before any pass may declare quiescence; the serving-loop test now binds the contract strictly (settled ⇒ current on a synchronous read). The parked-frame residual that remains is Flag 1.
- MAJORs: a vacuous basis assertion made real; the engine's delegated refusal branches tested; two coverage overstatements corrected (via-user hop → OW12; bookkeeping rebase made LIVE by switching the watermark to a patch, then tested); the ambient-flag park window closed with a refcounted lifecycle; the session-open/park race closed with `whenParked` chaining + identity-guarded teardown; the M4 wake-partition byte-level acceptance recorded (Flag 3).
- MINORs: the applicable-set filter wired (was dead machinery), sink replay-keying comment truth, watermark seal-failure now withholds both doc and metadata advance, exhausted-empty-cycle counting, holder-session rollback repair, protocol §2's FP2 bound noted, watch.add + expired-lease read-row tests, re-mark docstring honesty, tenure re-checks before every sink call, demand retirement on unwatch, activation-failure lease release.

The review confirmed: §8 tripwire scan clean, the envelope mapping matches the model's `admitDerived`, lease fencing matches C6/C7a/C7b (tenure captured at wave construction, checked at commit, renew ends the tenure before reacquire is reachable), exhaustion matches C10 (no W movement), wire frames byte-identical, the unstamped-seal reorder matches §3d's ruling, and all cited tests/files exist. Full verification re-ran green after the batch.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5439?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seefeldb:
## Independent adversarial review (coordinator round) — LANDABLE-WITH-FIXES → fixes + rulings landed as af7ed64ce

Reviewed at 6fe92f0fa by a fresh-eyes agent instructed to treat the PR's own self-review as unverified claims (its reviewer/fixer separation had been flagged). Outcome of that audit first: **all 20 self-review resolutions re-derived as REAL** — the independent pass then added two MAJORs of its own. Fix batch + the three owner rulings: `af7ed64ce`. Nothing merges — the train holds until the final phase exists.

**Verified deep and held:** the serving loop's production W-coupling is real (server.idle → flushSessions → synchronous loopback sends → scheduler idle re-probe, traced end to end); lease fencing matches the model's C6/C7a/C7b discipline; the RULED derived-envelope check sits engine-side with negatives both ways; stamp-kind coverage leaves no unstamped path to the seal destination; every key-vocabulary §5 M4/M1 re-key verified; suites independently reproduced (spec-model 19, runner 1114, memory 460→461 with the batch).

**MAJOR findings → fixes:**

- **F1 — the B1 W-soundness fix was half-unpinned.** A mutation probe removed the settle's `server.idle()` drain and the suite stayed green (the test ran at refresh delay 0; production is 5 ms — the deletion resurrects the original blocker there, uncaught). Fixed: a new serving-loop case at a real refresh cadence (400 ms ≥ the 300 ms flush deadline) whose strict settled ⇒ current assertion fails deterministically (42 ≠ 100) when the drain is removed.
- **F2 — the lease renew-blip reacquire path advanced W over withdrawn derivations.** Renewal fails → same-process reacquire succeeds → the aborted wave's withdrawn derivations are never re-run (no revert consumer re-arms the producer; inputs unchanged) → the persisted coverage head mints a watermark-only advance claiming the batch — settled ⇒ current violated, NOT self-healing. Fixed conservatively: a lease-lost wave abort now PARKS the space, making the already-tested fresh-runtime re-activation the only recovery arm. Pinned by a deterministic end-to-end renew-blip interleave test (test-owned settle gate holds the sealed wave open across the blip; no product hooks); reverting the fix times out exactly there.

**Minor findings → fixes:** protocol §2's delegated row over-promised (grant validation is presence+completeness in Phase 1 — parenthetical + OW13 for grant resolution); a sessionless delegated batch with a session-scoped op minted a chimera instance from the delegating envelope's session — now refused per scopes §5, with negatives; the explicit-read holder check is service-identity-wide, not per-process (recorded; Phase 5's cross-engine lookup owns the sharpening); the forgery-path watermark comments now state the true (conditional) behavior; a caller-less export deleted; `structureLoadFailures` surfaced in the §7 stats block instead of dying in a private field; a memory-side scoped-push test now localizes the M4 admission re-key (previously only the full runner suite caught its mutation); the hot-swap test pins the swapped derivation's own post-baseline derived commit.

**Owner rulings (all RULED 2026-08-05, landed in this batch):**

1. **Flag 3 ratified — both OFF-arm deltas accepted.** testing.md §2's gate now reads byte-identical "up to the recorded key-vocabulary §5 acceptances"; the register row names both the `fromSeq` resume-marker delta (unconsumed integer; ~5 ms marker-latency shift in a race window) and the shaper-bucket merge (the split was itself a defect — two buckets disagreeing about one piece's identity). Flag-gating was rejected: it would fork the dirty-key vocabulary against §4's no-two-definitions tripwire to preserve a defect in the OFF arm.
2. **Flag 2 confirmed.** An advance-only wave IS the batch's one derived commit; protocol §4's "never its own commit" is the bookkeeping-off-the-commit-stream rule, now stated in a clarifying parenthetical.
3. **Flag 1 accepted for Phase 1.** The settle input-barrier residual (a foreign authored frame parked behind the loop's own sealed commit can be claimed by W one wave early, self-healing) is accepted, with two named revisit items in the plan's Phase 2 section: the input-barrier distinction, and the pattern-updater CHECK half's bring-up in the sx2-serving-loop integration surface.

**New open question (flag-don't-fill; ruling requested when convenient):** in a delegated foreign batch, scoped READS (`validateCommitPreconditions`/`resolvePendingReads`) still resolve against the DELEGATING envelope identity, not the carried actor — the same identity class F4 fixed for writes. No Phase-1 producer sends delegated scoped reads. Options when a producer arrives: refuse at admission (fail-closed, mirroring F4) until the grant-scoped read design lands, or resolve symmetrically against the carried actor. Recommendation: refusal now costs nothing and keeps the FP2/grant-scoped-read design space open for Phase 5.

**Counts at af7ed64ce:** spec-model 19; runner 1114 (6016 steps); memory 461; serving-loop file 6/6; check-docs 522. ON-arm skip list: one entry (the two-browsers test — a named Phase 2 gate).

seefeldb (commented):
Verdict: changes required; Stage F is not landable at this exact head.

I re-adjudicated the current unresolved threads. The highest-risk confirmed defect is the persisted `leaseHolderReads` exemption: it survives session resume without revalidating a live execution lease and is not cleared on lease loss, while the push path uses that cached bit to bypass per-instance filtering. A former holder can therefore continue receiving foreign scoped instances. Existing thread: https://github.com/commontoolsinc/labs/pull/5439#discussion_r3731191378

Other confirmed current-head blockers include unstable event-handler action identity, omission of `shallowReads` from wave dependencies, mismatched foreign-batch settle sequences, process-global flag ownership/reset leaks, zombie serving after loop failure, close/activation races, pattern replacement before durable setup settlement, and the exported read wrapper dropping explicit `scopeKey`. These are already captured in the open inline threads; I am not duplicating them.

Resolve/re-adjudicate every current P0/P1 thread before merge. This PR currently records only the Cubic check, not a full Actions matrix.

seefeldb:
Fix ledger for the 2026-08-12 not-landable verdict and all open inline threads. Pushed `879e2f0a5..2758e60fc` (parent E tip a92aaadd5 merged first; one v2.ts conflict combined-intents).

**All 34 open threads dispositioned** — 28 fixed red-first with revert probes, 2 refuted with pinned probes, 2 flagged, 2 process notes. Per-thread table with commits and evidence in the fix report; highlights:

- **The P0 (`leaseHolderReads` exemption, r3731191378) — fixed.** The exemption now keys on CURRENT holdership at both the push path and session resume, clears on lease loss, and filtered entries are never cached. Red-first: former-holder-receives-nothing test; resume-revalidation test; lease-loss-clears test.
- Exact-instance delivery rollback, per-batch foreign settle sequences, shallowReads carried into wave dependencies, stable event-handler action identity stamps, settlement-gated hot-swap (pattern replacement waits on durable setup settlement), refcounted process-global flag ownership (reset leaks closed), close/activation races and loop-failure now PARK the space (no zombie serving), exported read wrapper keeps explicit scopeKey — each fixed red-first.
- **Refuted with pinned probes:** the eager-hop re-enqueue scenario is not constructible; the demand-record entityScopeKey concern is structurally absent. Probes committed so the refutations stay pinned.

**Flagged for the owner (not filled):**
- **F1 (r3731191420):** the demanded-piece startup path can mint setup writes on the serving runtime (`ensurePieceRunning` → `startCore` → `applySetupState`), that tx is unstamped, and the seal refusal is currently SWALLOWED (`startCore` fire-and-forgets the commit) — the setup repair drops silently and the piece runs against stale setup. Resolving it requires extending the RULED §3d three-site stamp list. Options: (a) sanction a fourth site — piece-start setup, the same `applySetupState` output the pattern-swap site already stamps — with a §3d spec touch; (b) declare server-side startup setup-repair unsupported in Phase 1 and make the refusal loud + counted; (c) fold into Phase 2's demand-load hardening. No code landed for this thread.
- **F2 (r3731191482):** the wake-shaping partition should key on the event ACTOR, but that identity doesn't exist as data until Phase 3 lands events with server-stamped `firedAt`; Phase 1 has no server-side event producers. Obligation comment at the site; a Phase-3 plan line-item will carry the actor data with the carriage.

**Also surfaced:**
- **F3 — test-discovery gap:** six memory test files named `*-test.ts` (commit-class, execution-lease, explicit-read, scope-key, scoped-push, wave-commit) are invisible to Deno's default discovery, so `deno task test` never ran them — this is how the parent merge silently broke one E test (repaired combined-intents, F5 in the report). They were run explicitly in the final bar (55/55). The P3 layer already renamed this family into discovery; the cascade will confirm coverage from there down.
- **F6:** stage-B's replay-ordering rule honored — the new admission check sits below the replay return; the pre-existing above-replay checks in this tree were left for B's fix to reorder at cascade time (959e074 is not yet an ancestor here), so the two rounds won't collide.

**OFF-arm:** no semantic deltas; the cost-only/non-observable ledger is in the fix report.

**Final bar at 2758e60fc:** runner suite 1146 passed / 0 failed (re-run in full after the last commit), memory 434/0 plus the six explicit files 55/0, spec-model 19/0, serving-loop soak 13× consecutive green, `deno check` + fmt clean.

seefeldb:

PR #5454, opened 2026-08-06
fix(ts-transformers): lower a pattern-scope cell read at any lowerable site

seefeldb:
## Problem

`pattern-context:get-call` rejected several pattern-scope `.get()` forms that are plainly lowerable, forcing authors into needless `computed()` wrappers.

The split was not terminal-vs-computation-feeding, as the diagnostic's comment claimed — it was **the shape of the enclosing expression**. `classifyExpressionSiteHandling`'s non-JSX tail admits a site only for a supported call root or an `isPostClosureWrapperRewriteExpression` (property access, element access, template, unary `!`, binary). A *call* is neither: `classifyCallRootPolicy` reports `cell.get()` itself as `restricted-get-call`, and bails to `none` for a call over one (`hasOpaquePathTerminalReceiverChain`).

Meanwhile the JSX router has always admitted both shapes explicitly as owned `jsx-root` sites. So at a variable initializer:

```tsx
const a = input.rows.get().length;                        // lowered
const e = input.rows.get()[0];                            // lowered
const c = input.s.get();                                  // rejected
const g = input.rows.get().join(",");                     // rejected
const h = input.rows.get().filter((r) => r.sentAt > 0);   // rejected
const j = input.rows.get().toSorted(byDate);              // rejected
```

…while `{input.s.get()}` in JSX lowered fine.

## Fix

`isCellReadTerminalCall()` is extracted in `call-root-support.ts` — the "is this `.get()` on a real `Cell`/`Writable`/`Stream`" test the helper-owned branch already did inline. The non-JSX tail of `classifyExpressionSiteHandling` now also admits a call site whose receiver chain reaches one, giving the other container kinds the same admission JSX already had.

The diagnostic keeps its purpose. Still rejected:

- a read with no lowerable site at all — statement position (`count.get();`)
- a read inside a reactive array-method callback (`rows.map((r) => r.c.get())`)
- a read on a non-cell receiver (`items.get()` on a plain pattern input), which also draws `opaque-get:invalid-call`

## Schemas

Each lift's input schema still shrinks to what its body reads — `.length` takes `items: { type: "unknown" }`, the reads that hand elements onward carry the full element type, and `.filter` lowers through `filterWithPattern` with its own callback lift shrunk to the one field the predicate reads. No fallback `items: true` anywhere. Pinned by the new `cell-get-terminal-binding-autowrap` golden.

## Verification

- Full `packages/ts-transformers` suite green
- Goldens for previously-accepted code are **byte-identical** (diffed before/after), so this only adds acceptance
- All 423 pattern files still pass `deno task cfcheck`
- A scratch pattern test confirmed the emitted lifts evaluate correctly and reactively at runtime (bare `.get()`, `.map(cb).join(",")`, `.filter(cb)`)
- `deno fmt --check`, `deno lint`, `deno task check`, `check-docs`, `check-skill-facts`, `check-no-waitfor`, `check-conflict-markers`, `check-baselines-append-only`, `check-single-copy-deps`, `check-unused-deps`, `check-deno-pins` all pass

## Specs

Behavior change, so all three records are updated: current-behavior §6.5, design-deltas Addendum 2, and the target-language §5.7 nuance paragraph. The normative matrix rows are left alone — that row is already an open ratification decision recorded in design-deltas, and widening the delta is not the same as ratifying it.

## Note on the motivating case

Of the `packages/patterns/topics/topic.tsx` wrappers this came from, only `linksView` unwraps cleanly. `commentsView` still fails — not on `pattern-context:get-call` but on `pattern-context:optional-chaining`, for `a?.sentAt` / `b?.sentAt` in the `.toSorted` comparator. That comparator ends up as plain JS inside the lift body, so the rejection looks equally over-broad, but it is a different diagnostic with its own blast radius. Left for a follow-up; topics is untouched here.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5454?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md
+++ b/docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md
@@ -453,9 +453,8 @@ Diagnostics emitted in all modes:
     optionality-orthogonality resolution. Each lift's input schema shrinks to
     what its body reads (`test/validation.test.ts:3179`; goldens
     `cell-get-binding-autowrap`, `cell-get-terminal-binding-autowrap`,
-    `with-reactive`). This is an unratified delta from the target-language
-    matrix's unconditional "Unsupported" and from the lowering contract's §3.9
-    eager-read bullet — see the design-deltas 2026-07-10 record
+    `with-reactive`). This is the ratified has-a-lowerable-site rule —
+    target-language spec §5.7 and its matrix rows are normative for it
 - **Error** `pattern-context:function-creation`
   - function creation in pattern context unless inside compute
     wrappers/JSX/allowed callbacks
--- a/docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md
+++ b/docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md
@@ -434,7 +434,10 @@ Diagnostics emitted in all modes:
   - a `.get()` read in restricted reactive context with no lowerable
     expression site to carry it: a statement-position read (`count.get();`),
     a read inside a reactive array-method callback
-    (`rows.map((row) => row.cell.get())`), or a read whose receiver is not a
+    (`rows.map((row) => row.cell.get())`), a read inside a plain
+    (non-reactive) array-method callback
+    (`["-", "+"].map((sep) => rows.get().join(sep))`, which is not an eligible
+    pattern-owned wrapper site), or a read whose receiver is not a
     `Cell`/`Writable`/`Stream` (`items.get()` on a plain pattern input, which
     also draws `opaque-get:invalid-call`)
   - a cell read that DOES sit at a lowerable site is not rejected: the site is
@@ -443,12 +446,16 @@ Diagnostics emitted in all modes:
     `input.key("count").get()` at a return site), a computation over it
     (`{ value: count.get() * 2 }`), and a call whose receiver chain reaches it
     (`rows.get().join(",")`, `rows.get().filter(...)`, which lowers through
-    `filterWithPattern`). Each lift's input schema shrinks to what its body
-    reads (`test/validation.test.ts:3179`; goldens
+    `filterWithPattern`). Parentheses around the site and the computed-key
+    spelling of the read (`layout["get"]()`) do not change the decision, and
+    neither does optionality — `layout?.get()` and `layout?.get?.()` on a cell
+    lower like their non-optional spellings, per the 2026-07-23
+    optionality-orthogonality resolution. Each lift's input schema shrinks to
+    what its body reads (`test/validation.test.ts:3179`; goldens
     `cell-get-binding-autowrap`, `cell-get-terminal-binding-autowrap`,
     `with-reactive`). This is an unratified delta from the target-language
-    matrix's unconditional "Unsupported" — see the design-deltas 2026-07-10
-    record
+    matrix's unconditional "Unsupported" and from the lowering contract's §3.9
+    eager-read bullet — see the design-deltas 2026-07-10 record
 - **Error** `pattern-context:function-creation`
   - function creation in pattern context unless inside compute
     wrappers/JSX/allowed callbacks
--- a/docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md
+++ b/docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md
@@ -431,13 +431,21 @@ Compute wrappers override restrictions:
 Diagnostics emitted in all modes:
 
 - **Error** `pattern-context:get-call`
-  - a **terminal** `.get()` read in restricted reactive context — one whose
-    value is used directly (`{ value: count.get() }`,
-    `const v = count.get()`, `input.key("count").get()` at a return site)
-  - since #3725 (2026-05-28), a **computation-feeding** read at a lowerable
-    site (`{ value: count.get() * 2 }`) is NOT rejected: the containing
-    expression is auto-wrapped into a lift-applied computation
-    (`test/validation.test.ts:3179`; goldens `cell-get-binding-autowrap`,
+  - a `.get()` read in restricted reactive context with no lowerable
+    expression site to carry it: a statement-position read (`count.get();`),
+    a read inside a reactive array-method callback
+    (`rows.map((row) => row.cell.get())`), or a read whose receiver is not a
+    `Cell`/`Writable`/`Stream` (`items.get()` on a plain pattern input, which
+    also draws `opaque-get:invalid-call`)
+  - a cell read that DOES sit at a lowerable site is not rejected: the site is
+    auto-wrapped into a lift-applied computation. That covers the read itself
+    (`const v = count.get()`, `{ value: count.get() }`,
+    `input.key("count").get()` at a return site), a computation over it
+    (`{ value: count.get() * 2 }`), and a call whose receiver chain reaches it
+    (`rows.get().join(",")`, `rows.get().filter(...)`, which lowers through
+    `filterWithPattern`). Each lift's input schema shrinks to what its body
+    reads (`test/validation.test.ts:3179`; goldens
+    `cell-get-binding-autowrap`, `cell-get-terminal-binding-autowrap`,
     `with-reactive`). This is an unratified delta from the target-language
     matrix's unconditional "Unsupported" — see the design-deltas 2026-07-10
     record
--- a/docs/specs/ts-transformer/ts_transformers_design_deltas.md
+++ b/docs/specs/ts-transformer/ts_transformers_design_deltas.md
@@ -108,13 +108,19 @@ Language deltas found by adversarial verification of the target-language
 matrix (implementation vs normative spec). Resolution status is recorded on
 each finding:
 
-- **Top-level eager-read carve-out (#3725, 2026-05-28).** Validation accepts
-  computation-feeding top-level `.get()` reads and auto-wraps the containing
-  expression lift-applied; terminal reads still reject. Golden-pinned
-  (`cell-get-binding-autowrap`, `with-reactive`;
+- **Top-level eager-read carve-out (#3725, 2026-05-28).** Validation accepts a
+  top-level `.get()` read on a `Cell`/`Writable`/`Stream` wherever the read has
+  a lowerable expression site to carry it, and auto-wraps that site
+  lift-applied. The line is the site, not the shape of the read: a binding, a
+  return, an object property, an array element, an argument, and a call whose
+  receiver chain reaches the read are all accepted, terminal or not; a read
+  with no such site — statement position, or inside a reactive array-method
+  callback — still rejects, as does a read on a value that is not a cell.
+  Golden-pinned (`cell-get-binding-autowrap`,
+  `cell-get-terminal-binding-autowrap`, `with-reactive`;
   `test/validation.test.ts:3179`). Matrix row still says Unsupported
-  unconditionally. Decision open: ratify a terminal-vs-computation-feeding
-  split in the matrix, or revert (breaks two goldens + one test).
+  unconditionally. Decision open: ratify a has-a-lowerable-site split in the
+  matrix, or revert (breaks three goldens + several tests).
 - **Optional-call accepted in JSX and compute callbacks — resolved 2026-07-23
   by making optionality orthogonal to call support.** The earlier location
   split was an implementation leak, and the proposed blanket rejection would
--- a/docs/specs/ts-transformer/ts_transformers_design_deltas.md
+++ b/docs/specs/ts-transformer/ts_transformers_design_deltas.md
@@ -108,22 +108,29 @@ Language deltas found by adversarial verification of the target-language
 matrix (implementation vs normative spec). Resolution status is recorded on
 each finding:
 
-- **Top-level eager-read carve-out (#3725, 2026-05-28).** Validation accepts a
-  top-level `.get()` read on a `Cell`/`Writable`/`Stream` wherever the read has
-  a lowerable expression site to carry it, and auto-wraps that site
-  lift-applied. The line is the site, not the shape of the read: a binding, a
-  return, an object property, an array element, an argument, and a call whose
-  receiver chain reaches the read are all accepted, terminal or not; a read
-  with no such site — statement position, inside a reactive array-method
-  callback, or inside a plain (non-reactive) array-method callback — still
-  rejects, as does a read on a value that is not a cell. Golden-pinned
-  (`cell-get-binding-autowrap`, `cell-get-terminal-binding-autowrap`,
-  `with-reactive`; `test/validation.test.ts:3179`). Two normative records still
-  say otherwise: the matrix row says Unsupported unconditionally, and the
-  lowering contract's §3.9 lists a direct top-level eager read among the
-  constructs that must fail clearly. Decision open: ratify a
-  has-a-lowerable-site split in the matrix (and retire lowering-contract §3.9's
-  eager-read bullet), or revert (breaks three goldens + several tests).
+- **Top-level eager-read carve-out (#3725, 2026-05-28) — resolved 2026-08-07 by
+  ratifying the has-a-lowerable-site rule.** A `.get()` read on a
+  `Cell`/`Writable`/`Stream` in pattern-owned context is part of the language
+  wherever a lowerable expression site can carry it; that site lowers into a
+  lift, which is what keeps the read live. The line is the site, not the shape
+  of the read: a binding, a return, an object property, an array element, an
+  argument, a computation over the read, and a call whose receiver chain
+  reaches it are all accepted, terminal or not, and parentheses, the
+  computed-key spelling, and the optional spellings do not change the
+  classification. A read with no such site — statement position, a reactive
+  array-method callback, or a plain (non-reactive) array-method callback —
+  remains outside the language, as does a read on a value that is not a cell.
+
+  What settled it: the matrix already blessed the JSX spelling, so the earlier
+  rule was really "no eager reads outside JSX" — which made extracting a JSX
+  expression into a named binding change a program's legality. The rejection's
+  main output was "wrap it in `computed()`" diagnostics that bred wrapper
+  cascades in author code, against a platform direction of needing explicit
+  `computed()` less often. Ratified in the target-language matrix and §5.7;
+  lowering-contract §3.9's eager-read bullet now names the site-less read
+  rather than the top-level read. Golden-pinned (`cell-get-binding-autowrap`,
+  `cell-get-terminal-binding-autowrap`, `with-reactive`;
+  `test/validation.test.ts:3179`).
 - **Optional-call accepted in JSX and compute callbacks — resolved 2026-07-23
   by making optionality orthogonal to call support.** The earlier location
   split was an implementation leak, and the proposed blanket rejection would
--- a/docs/specs/ts-transformer/ts_transformers_design_deltas.md
+++ b/docs/specs/ts-transformer/ts_transformers_design_deltas.md
@@ -114,13 +114,16 @@ each finding:
   lift-applied. The line is the site, not the shape of the read: a binding, a
   return, an object property, an array element, an argument, and a call whose
   receiver chain reaches the read are all accepted, terminal or not; a read
-  with no such site — statement position, or inside a reactive array-method
-  callback — still rejects, as does a read on a value that is not a cell.
-  Golden-pinned (`cell-get-binding-autowrap`,
-  `cell-get-terminal-binding-autowrap`, `with-reactive`;
-  `test/validation.test.ts:3179`). Matrix row still says Unsupported
-  unconditionally. Decision open: ratify a has-a-lowerable-site split in the
-  matrix, or revert (breaks three goldens + several tests).
+  with no such site — statement position, inside a reactive array-method
+  callback, or inside a plain (non-reactive) array-method callback — still
+  rejects, as does a read on a value that is not a cell. Golden-pinned
+  (`cell-get-binding-autowrap`, `cell-get-terminal-binding-autowrap`,
+  `with-reactive`; `test/validation.test.ts:3179`). Two normative records still
+  say otherwise: the matrix row says Unsupported unconditionally, and the
+  lowering contract's §3.9 lists a direct top-level eager read among the
+  constructs that must fail clearly. Decision open: ratify a
+  has-a-lowerable-site split in the matrix (and retire lowering-contract §3.9's
+  eager-read bullet), or revert (breaks three goldens + several tests).
 - **Optional-call accepted in JSX and compute callbacks — resolved 2026-07-23
   by making optionality orthogonal to call support.** The earlier location
   split was an implementation leak, and the proposed blanket rejection would
--- a/docs/specs/ts-transformer/ts_transformers_target_pattern_language_spec.md
+++ b/docs/specs/ts-transformer/ts_transformers_target_pattern_language_spec.md
@@ -78,7 +78,8 @@ Each construct family is classified as one of:
 | Optional receiver navigation and optional invocation on otherwise-supported call roots | Supported | Optionality does not define a separate call family: `value?.method()`, `value.method?.()`, and combined chains lower wherever the corresponding non-optional call is supported, with receiver binding, evaluation order, nullish short-circuiting, and skipped argument evaluation preserved |
 | Direct non-JSX receiver-method calls on reactive values in top-level pattern-body expression sites | Supported | Value-like receiver-method roots at top-level object-property, call-argument, variable-initializer, array-element, or return-expression sites lower to derived local value expressions |
 | Direct receiver-method roots inside supported collection callbacks | Supported | Callback-local value-like receiver-method roots lower to callback-local lift-applied computations instead of remaining raw or requiring manual wrapper calls |
-| Direct top-level `.get()` reads in pattern-owned reactive context | Unsupported | Even on true cell-like values, eager `.get()` reads should move into JSX or an explicit computation callback such as `computed`, `action`, `lift`, or `handler` rather than living directly in the top-level declarative pattern body |
+| Direct top-level `.get()` reads on explicitly cell-like values at a lowerable expression site | Supported | On a true `Cell`/`Writable`/`Stream`, an eager read is part of the language wherever a lowerable expression site can carry it: a variable initializer, a return, an object property, an array element, a call argument, a computation over the read, or a call whose receiver chain reaches it. That site lowers into a lift, so the read stays live rather than freezing. Parentheses, the computed-key spelling `cell["get"]()`, and the optional spellings `cell?.get()` / `cell.get?.()` do not change the classification |
+| Direct top-level `.get()` reads with no lowerable expression site | Unsupported | A read with nothing to carry it stays outside the language, because there is no site to lower into a lift: statement position (`count.get();`), a reactive array-method callback (`rows.map((row) => row.cell.get())`), and a plain array-method callback (`["-", "+"].map((sep) => rows.get().join(sep))`). These move into an explicit computation callback such as `computed`, `action`, `lift`, or `handler` |
 | `.get()` on ordinary opaque/reactive values | Unsupported | Pattern inputs, `computed` results, `lift` results, and other ordinary reactive values should be read directly rather than through `.get()` |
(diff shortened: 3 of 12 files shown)


packages/ts-transformers/src/transformers/expression-site-policy.ts
```diff
@@ -323,6 +324,60 @@ function isSharedPostClosureCallRootKind(
...
+  context: TransformationContext,
+): boolean {
+  if (!ts.isCallExpression(expression)) {
+    return false;
+  }
+
+  let current: ts.Expression = expression;
+  while (true) {
+    if (ts.isCallExpression(current)) {
+      if (isCellReadTerminalCall(current, context)) {
```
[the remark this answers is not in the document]
seefeldb: Checked this one against the compiler — declining, because the shape it describes can't reach a lift.

`rows.get()[0].get()` where `Row` has no `get` is rejected before any of this runs, by TypeScript itself:

```
[ERROR] Property 'get' does not exist on type 'Row'.
10 |   const bad = input.rows.get()[0].get();
   |                                   ^
```

That holds identically on `main` and on this branch — `pattern-context:get-call` never fired for that chain either way, so it isn't a regression here.

The shape that *does* typecheck is a plain object with a real `get()` method:

```tsx
interface Slot { get(): string }
const value = input.slots.get()[0].get();
```

That's an ordinary method call on a resolved plain value, and lowering it inside the lift is correct — it emits `__cfLift_1({ input: { slots: input.key("slots") } })` with the whole chain in the callback body. The suggested guard (stop the walk when the terminal `.get()` isn't a cell read) would reject exactly this valid form.

There's also no reactive misuse that slips past: a cell/stream receiver returns true at the first iteration, and an opaque one is caught by `opaque-get:invalid-call` — `wish(...).result.get()` still bottoms out on an identifier callee, so the walk returns false and the existing diagnostic stands.

docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md
```diff
@@ -430,13 +430,21 @@ Compute wrappers override restrictions:
...
-    `const v = count.get()`, `input.key("count").get()` at a return site)
-  - since #3725 (2026-05-28), a **computation-feeding** read at a lowerable
-    site (`{ value: count.get() * 2 }`) is NOT rejected: the containing
-    expression is auto-wrapped into a lift-applied computation
-    (`test/validation.test.ts:3179`; goldens `cell-get-binding-autowrap`,
+  - a `.get()` read in restricted reactive context with no lowerable
+    expression site to carry it: a statement-position read (`count.get();`),
+    a read inside a reactive array-method callback
+    (`rows.map((row) => row.cell.get())`), or a read whose receiver is not a
+    `Cell`/`Writable`/`Stream` (`items.get()` on a plain pattern input, which
```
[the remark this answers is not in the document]
seefeldb: Confirmed — `Writable<T>` is an alias for `Cell<T>`, so it carries the `cell` brand and needs no separate branch:

```ts
// packages/api/index.ts:1407
export type Writable<T = unknown> = Cell<T>;
```

`OpaqueGetValidationTransformer` gates on the same pair for the same reason ("Cell/Writable/Stream types have `.get()` — only opaque types don't"). Every probe behind this change used a `Writable`-typed receiver and was admitted, so the prose matches the implementation. No change made.

docs/specs/ts-transformer/ts_transformers_target_pattern_language_spec.md
```diff
@@ -559,10 +559,13 @@ The real rule is:
   and canonical lowered traversal rather than authored `.get()`
 
 One important nuance: the implementation has moved on this boundary since
-this spec's v1 — since #3725, validation accepts **computation-feeding**
-top-level eager reads (`{ value: count.get() * 2 }`) and auto-wraps them into
-lift-applied computations, while terminal reads (`{ value: count.get() }`)
-still reject. That carve-out is an unratified delta recorded in
+this spec's v1. Validation accepts a top-level eager read on a true cell
+wherever the read has a lowerable expression site to carry it — a binding, a
+return, an object property, an array element, an argument, a computation over
```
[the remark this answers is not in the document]
seefeldb: Valid — fixed in 8a670be. Item 3 now points at the nuance:

```
   - `{ value: input.key("foo").get() }` directly in a top-level pattern body
   - the implementation accepts this example today; see the nuance at the end
     of this section for the shape of that unratified carve-out
```

Item 3 keeps stating the normative position rather than moving into the accepted set — the divergence is the recorded delta, and ratifying it is a separate decision. The cross-reference is there so a reader isn't left choosing between two flat contradictions.

packages/ts-transformers/src/transformers/call-root-support.ts
```diff
@@ -57,6 +57,36 @@ function hasOpaquePathTerminalReceiverChain(
...
+export function isCellReadTerminalCall(
+  expression: ts.CallExpression,
+  context: TransformationContext,
+): boolean {
+  if (classifyOpaquePathTerminalCall(expression) !== "get") {
+    return false;
+  }
+
+  const callee = expression.expression;
+  if (!ts.isPropertyAccessExpression(callee)) {
```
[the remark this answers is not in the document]
seefeldb: Valid — fixed in 8a670be. `isCellReadTerminalCall` now accepts an `ElementAccessExpression` callee too, so `layout["get"]()` lowers like `layout.get()`:

```tsx
const bracket = __cfLift_1({ input: { s: input.key("s") } }).for("bracket", true);
// __cfLift_1 body: ({ input }) => input.s["get"]()
```

Locked in by a new `validation.test.ts` step. Every golden is byte-identical, so the widening (which also reaches the helper-owned `ifElse`/`when`/`unless` branch, since both share this predicate) changes nothing that was already accepted.

Worth noting the property-access-only restriction was pre-existing rather than introduced here — the bracket spelling errored on `main` too. But this PR is what makes the plain spelling work at these sites, so leaving the two inconsistent would be a gap of its own making.

seefeldb:
@mathpirate fwiw, this triggered a couple "wrap this in computed()" errors which then led to more things that don't need to be in computed() be in computed.

packages/ts-transformers/src/transformers/expression-site-policy.ts
```diff
@@ -1199,7 +1254,11 @@ export function classifyExpressionSiteHandling(
     );
     const patternOwnedReceiverMethod = supportedCallRootKind ===
       "pattern-owned-receiver-method";
-    if (!sharedPostClosureCallRoot && !patternOwnedReceiverMethod) {
+    if (
+      !sharedPostClosureCallRoot &&
+      !patternOwnedReceiverMethod &&
+      !isCellReadCallRootExpression(expression, context)
```
mathpirate: Probing the new admission against callback boundaries turned up one hole. Three spellings inside the same plain (non-reactive) array `.map` at pattern scope — all three rejected on `main`:

```tsx
const joined = ["-", "+"].map((sep) => rows.get().join(sep)); // now accepted
const doubled = [1, 2].map((n) => count.get() * n);           // still rejected
const lens = [1, 2].map((n) => rows.get().length + n);        // still rejected
```

Cause: the cell-read check sits in the outer condition, so an admitted call chain skips both inner checks — including `isEligiblePatternOwnedWrapperCallbackSite`, which is false for `plain-array-value` callbacks (their body context inherits pattern scope and they aren't `arrayMethodOwned`, so their sites reach this tail). The binary and property-access shapes still route through `isPostClosureWrapperRewriteExpression` and honor the gate. That re-creates, one level down, the shape split this PR exists to remove — and cuts against the new design-deltas wording ("the line is the site, not the shape of the read").

To be fair to the accepted case: the emit is correct (`sep` becomes a schema'd lift input, the lift applies inline in the callback), and `forEach`/void callbacks stay blocked by the function-creation diagnostic — so this is a consistency hole, not a miscompile. Suggested fix keeps the admission but honors the gate:

```ts
if (!sharedPostClosureCallRoot && !patternOwnedReceiverMethod) {
  if (
    !isPostClosureWrapperRewriteExpression(expression, context) &&
    !isCellReadCallRootExpression(expression, context)
  ) {
    return { kind: "skip", reason: "not-lowerable" };
  }

  if (!isEligiblePatternOwnedWrapperCallbackSite(expression, context)) {
    return { kind: "skip", reason: "not-lowerable" };
  }
}
```

Verified at this branch's head: full package suite identical (1130 passed / 0 failed), the plain-array call chain flips back to rejected (uniform with its binary/prop siblings), and everything this PR pins — bindings, returns, the computed-key spelling, `.join`/`.filter` chains, a chain in builder-call args — still accepts, since top-level sites sit in the pattern-builder callback where the gate passes. If you'd rather admit plain-array callbacks deliberately, that's defensible too — but then the binary/prop shapes should follow, and the spec text needs the case named either way (separate comment).

— Fable on behalf of Gideon

docs/specs/ts-transformer/ts_transformers_design_deltas.md
```diff
@@ -108,13 +108,19 @@ Language deltas found by adversarial verification of the target-language
...
+  return, an object property, an array element, an argument, and a call whose
+  receiver chain reaches the read are all accepted, terminal or not; a read
+  with no such site — statement position, or inside a reactive array-method
+  callback — still rejects, as does a read on a value that is not a cell.
+  Golden-pinned (`cell-get-binding-autowrap`,
+  `cell-get-terminal-binding-autowrap`, `with-reactive`;
   `test/validation.test.ts:3179`). Matrix row still says Unsupported
-  unconditionally. Decision open: ratify a terminal-vs-computation-feeding
-  split in the matrix, or revert (breaks two goldens + one test).
+  unconditionally. Decision open: ratify a has-a-lowerable-site split in the
```
mathpirate: Two bookkeeping gaps in the ratification record, both cheap to close in this PR:

1. The matrix row isn't the only normative statement of the reject stance. `ts_transformers_lowering_contract.md` §3.9 lists "direct top-level eager `.get()` reads on reactive or cell-like values" among the constructs that must fail clearly. That bullet has been in tension with the carve-out since #3725, and this PR widens the divergence without naming it. Suggest the decision line read "ratify a has-a-lowerable-site split in the matrix (and retire lowering-contract §3.9's eager-read bullet), or revert" — so the eventual decision moves both normative records together instead of leaving §3.9 behind.

2. The rejected-set enumerations added here — current-behavior §6.5 and the target-language §5.7 nuance both say "statement position, or inside a reactive array-method callback" — are silent on plain (non-reactive) array callbacks, where behavior on this branch is currently mixed (see the thread on `expression-site-policy.ts`: call chains accept, binary/prop shapes reject). Whichever way that thread lands, the third case belongs in both sentences.

— Fable on behalf of Gideon

packages/ts-transformers/test/validation.test.ts
```diff
@@ -3196,6 +3218,94 @@ Deno.test("Reactive .get() Validation", async (t) => {
     },
   );
 
+  await t.step(
+    "allows a bare .get() binding on a Writable (auto-wrapped)",
```
mathpirate, line 3222: The widened admission also flips the optional spellings on real cells — both rejected on `main`, both accepted here (probed), and neither is named in the PR body, the specs, or a test:

```tsx
export default pattern<{ s?: Writable<string> }>((input) => {
  const v = input.s?.get();   // was pattern-context:get-call, now lowers
  // ...
});
const all = rows.get?.();      // likewise
```

This looks right, and deliberately so — it's the optionality-orthogonality resolution (the 2026-07-23 design-deltas entry) composing with the new site admission — and the emitted lift is exactly what you'd want: input schema `s: { anyOf: [string, undefined], asCell: ["readonly"] }` with `s` not required, output `string | undefined`. But the only optional-`.get()` pins in this file are the opaque-receiver trio, which assert *rejection* — so this acceptance could regress silently with everything staying green. Suggest a step pinning the real-cell optional spellings as accepted (mirroring the opaque trio), plus a clause in the §6.5 example list; if you want the schema shape pinned too, an optional-cell case in the `cell-get-terminal-binding-autowrap` fixture would do it.

— Fable on behalf of Gideon

packages/ts-transformers/src/transformers/expression-site-policy.ts
```diff
@@ -323,6 +324,60 @@ function isSharedPostClosureCallRootKind(
...
+ *
+ * Receivers that are not cells are excluded by {@link isCellReadTerminalCall}:
+ * a `.get()` on an opaque value is a mistake with its own diagnostic, not a
+ * computation to lower.
+ */
+function isCellReadCallRootExpression(
+  expression: ts.Expression,
+  context: TransformationContext,
+): boolean {
+  if (!ts.isCallExpression(expression)) {
```
mathpirate: Tiny one: the walk never unwraps the site expression itself, so a parenthesized initializer misses the admission — probed on this branch:

```tsx
const joined = rows.get().join(",");    // accepted
const wrapped = (rows.get().join(",")); // pattern-context:get-call
```

`isPostClosureWrapperRewriteExpression` unwraps at entry, so the non-call shapes don't have this edge; here the loop unwraps interior steps but tests `expression` raw. Starting the walk at `unwrapExpression(expression)` squares it.

— Fable on behalf of Gideon

packages/ts-transformers/test/validation.test.ts
```diff
@@ -3196,6 +3218,94 @@ Deno.test("Reactive .get() Validation", async (t) => {
...
+      assertEquals(
+        errors.length,
+        0,
+        "layout['get']() is the same read as layout.get()",
+      );
+    },
+  );
+
+  await t.step(
+    "still errors on an opaque .get() binding",
```
mathpirate, line 3376: Two behaviors this PR leans on are now spec-cited but unpinned; each is one cheap step:

1. The reactive array-method-callback rejection — `rows.map((row) => row.title.get())` — is the canonical still-rejected example in both updated spec passages and in the PR description, but no test asserts it. (Verified manually at this head: still `pattern-context:get-call`.) The nearest existing coverage is all opaque-receiver, which rejects for a different reason, so a regression here wouldn't trip anything.

2. The computed-key widening also reaches the helper boundary — `ifElse(show, layout["get"](), "fallback")` flips rejected→accepted on this branch (probed both sides), via the shared predicate as you noted in the other thread — but the new computed-key step only pins the binding site. A helper-boundary twin would keep the two admission paths from drifting apart again, which is the exact failure mode this PR fixes.

— Fable on behalf of Gideon

packages/ts-transformers/src/transformers/expression-site-policy.ts
```diff
@@ -323,6 +324,60 @@ function isSharedPostClosureCallRootKind(
...
+/**
+ * True when a call site reads a cell — `count.get()` itself, or a call whose
+ * receiver chain reaches one (`rows.get().toSorted(byDate)`,
+ * `rows.get().items.join(", ")`).
+ *
+ * A cell read yields a plain snapshot, so the site holding it has to become a
+ * lift for the value to stay reactive. `classifyCallRootPolicy` reports the read
+ * as `restricted-get-call` and declines to classify a call over one, leaving
+ * both shapes without a supported call root; the JSX router admits them anyway
+ * as owned `jsx-root` sites. This is the same admission for the other container
```
mathpirate: Label's half-right: only the bare read is an owned `jsx-root` (the `receiver-method` branch with a terminal call); a call over one falls through to `sharedDecision("shared-post-closure")`. Suggest: "the JSX router admits them anyway (a bare read as an owned `jsx-root` site, a call over one as a shared post-closure site)." The motivation stands either way — this just keeps the routing claim accurate for whoever traces owners next. (The PR description has the same phrase; no need to touch that.)

— Fable on behalf of Gideon

mathpirate (approved):
Approving. I went through this end to end: traced both admission paths against the classifier sources, re-derived the before/after table from the code, ran the package suite at head (1130/0) and on a scratch merge with current `main` (1134/0 — worth a branch update before merge, since CI predates #4972 landing on the same file), and probed the boundary shapes the tests don't cover. Every claim in the description held up, including the two cubic declines. The inline comments I left are consistency edges and pins, none of them load-bearing against the direction of the change; the plain-array-callback one is the only behavioral item, and it comes with a verified one-block fix.

Beyond the mechanics, I want to put one decision in front of us plainly, because it's easy to lose in the delta-record bookkeeping. The underlying question is simple: **may a pattern author call `.get()` on a cell-like value (`Cell`/`Writable`/`Stream`) at the top level of a pattern body?**

The normative spec still answers no — matrix row 81 says Unsupported even on true cells, and lowering-contract §3.9 says such reads must fail clearly. The implementation has answered "increasingly yes" since #3725, and with this PR answers yes at every lowerable site. Each step has been honestly recorded as an unratified carve-out, but the recorded divergence keeps widening while the decision line stays open. I think it's time to close it — as an official **yes**:

- What the implementation does is no longer an accident we'd be rubber-stamping; it's a nameable rule (a cell read is legal wherever a lowerable site can carry it, and that site becomes a lift that keeps the value live), with the residue still rejected loudly (statement position, reactive array-method callbacks, non-cell receivers).
- The matrix already blesses `{count.get()}` in JSX, so the old rule was really "no eager reads *outside JSX*" — which makes a pure refactor (extract a JSX expression to a named binding) change a program's legality.
- We're moving toward pattern authors not needing explicit `computed()` almost anywhere. The old rejection's main output was exactly that — "wrap it in `computed()`" errors breeding wrapper cascades, which is what motivated this PR. Keeping the rule on the books as normative while the implementation, the ergonomics, and the platform direction all point the other way serves no one.
- The revert arm of the recorded decision decays with every widening (now three goldens, several tests, and adopted author code). Postponing is just ratifying slowly, without the spec hygiene.

Concretely, ratification is a spec-only follow-up (not this PR): matrix row 81 gains the site rule, §5.7's nuance paragraph folds into the normative text, §3.9's eager-read bullet retires or reworks to "lowers into a lift or fails clearly," the design-deltas entry resolves, and the edges from the inline threads (plain-array callbacks, the optional spellings, parenthesization) get crisp normative answers as part of it — plus a reword of the diagnostic message, which still says ".get() is not allowed in reactive context" when the real rule is now "this read has no lowerable site." Happy to draft that spec change if you agree the direction is settled — or to hear the case for keeping the question open.

— Fable on behalf of Gideon

packages/ts-transformers/src/transformers/expression-site-policy.ts
seefeldb: Confirmed and taken, exactly as suggested. Reproduced all three spellings on the branch head before the fix — `["-", "+"].map((sep) => rows.get().join(sep))` accepted, the binary and property-access siblings rejected — and after moving the disjunct inside, all three reject uniformly.

Your diagnosis of the cause was right: the admission short-circuited `isEligiblePatternOwnedWrapperCallbackSite` along with the shape check, because it sat in the outer condition. I'd reasoned when writing it that a call root belongs beside `sharedPostClosureCallRoot`/`patternOwnedReceiverMethod`, which also bypass the gate — but those two come out of `classifyCallRootPolicy` and carry context, whereas this predicate is purely syntactic, so it has no business skipping the gate. And "the line is the site, not the shape of the read" is the PR's own thesis; leaving a shape split one level down contradicts it.

Everything the PR pins still accepts (bindings, returns, the computed-key spelling, `.join`/`.filter` chains), package suite 1137/0.

I kept plain-array callbacks rejected rather than admitting them deliberately, since that's the conservative arm and matches the binary/prop shapes today. The case is now named in both spec enumerations, so if we later want to admit them it's a visible edit rather than a silent drift.

packages/ts-transformers/src/transformers/expression-site-policy.ts
seefeldb: Taken. Reproduced (`(rows.get().join(","))` rejected, unparenthesized accepted) and fixed by starting the walk at `unwrapExpression(expression)`, which also lets the entry `isCallExpression` test see through the wrapper. Pinned by a new step; mutation-tested — restoring the raw `expression` fails that step and nothing else.

packages/ts-transformers/src/transformers/expression-site-policy.ts
seefeldb: Right, thanks — a call over a read falls through to `sharedDecision("shared-post-closure")`, only the bare read is an owned `jsx-root`. Reworded to your suggestion verbatim.

docs/specs/ts-transformer/ts_transformers_design_deltas.md
seefeldb: Both taken.

1. §3.9 is now named in the decision line, which reads "ratify a has-a-lowerable-site split in the matrix (and retire lowering-contract §3.9's eager-read bullet), or revert". The entry also states plainly that two normative records say otherwise, rather than just the matrix row — you're right that the divergence was widening against §3.9 unnamed.

2. Plain array callbacks are now the third case in both enumerations — current-behavior §6.5 and the target-language §5.7 nuance — with the example and the reason (`["-", "+"].map((sep) => rows.get().join(sep))`, not an eligible pattern-owned wrapper site). They land on the rejected side, matching the code fix in the other thread.

§6.5 also picked up a clause for the shapes that don't change the decision — parentheses, the computed-key spelling, and the optional spellings with a pointer to the 2026-07-23 orthogonality resolution — since those were being relied on without being written down anywhere.

packages/ts-transformers/test/validation.test.ts
seefeldb: Taken, and thanks for catching that the acceptance was live but unpinned — I hadn't probed the optional spellings at all.

Confirmed both on the branch (`layout?.get()` and `layout?.get?.()` on a real cell, accepted; the emitted lift matches what you describe). New step pins them, worded so the reason is visible next to the opaque trio that asserts rejection: optionality doesn't change whether the read has a lowerable site.

I took the fixture suggestion too — `cell-get-terminal-binding-autowrap` now carries an optional cell, so the schema shape is pinned rather than just the diagnostic count:

```jsx
const __cfLift_7 = __cfHelpers.lift<{
    label?: __cfHelpers.Writable<string> | undefined;
}, string | undefined>(({ label }) => label?.get(), {
    type: "object",
    properties: {
        label: {
            anyOf: [{ type: "string" }, { type: "undefined" }],
            asCell: ["readonly"]
        }
    }
} as const satisfies __cfHelpers.JSONSchema, {
    type: ["string", "undefined"]
} as const satisfies __cfHelpers.JSONSchema);
```

No `required`, so the unrequired-input half is pinned as well.

packages/ts-transformers/test/validation.test.ts
seefeldb: Both taken.

1. `rows.map((row) => row.title.get())` now has its own step asserting `pattern-context:get-call`. Verified it rejects for the reactive-array-method-callback reason rather than the opaque-receiver one — the receiver is a real `Writable`, so the nearest existing coverage genuinely wasn't guarding this.

2. `ifElse(show, layout["get"](), "fallback")` has a helper-boundary twin of the computed-key step. Agreed on the failure mode — the two admission paths drifting apart is what this PR fixes, so pinning both sides of the shared predicate is the right guard.

Both mutation-tested along with the rest of the new steps.

seefeldb:
Thanks — that's an unusually useful review. All six inline items are addressed in `6d6328e`, and the branch is rebased onto current `main` (it was 220 behind; suite is 1137/0 at the new base, so the #4972 concern is resolved).

I reproduced every one of your probes before acting on it, and each landed exactly as you described:

| item | verified | outcome |
|---|---|---|
| plain-array callback split | `.join(sep)` accepted, binary/prop rejected | fix applied, all three now reject |
| parenthesized initializer | `(rows.get().join(","))` rejected | unwrap at entry |
| optional spellings | `layout?.get()`, `layout?.get?.()` accepted | pinned, incl. schema in the fixture |
| reactive array-method callback | rejects, unpinned | pinned |
| computed-key at helper boundary | accepted, unpinned | pinned |
| `jsx-root` routing claim | only bare reads are owned | reworded verbatim |

The behavioral one went in as your one-block fix. Your reasoning changed my mind on it: I'd deliberately put the disjunct in the outer condition, on the grounds that call roots sit beside `sharedPostClosureCallRoot`/`patternOwnedReceiverMethod` which also bypass the gate — but those carry `classifyCallRootPolicy` context and this predicate is purely syntactic, so skipping the gate was unearned. All new steps are mutation-tested; reverting either fix fails exactly its own step and nothing else.

Full local gate set green: package suite 1137/0, `fmt`, `lint`, `check`, `check-docs` (547 blocks), `cfcheck` (413 patterns), and the seven `check-*` gates.

---

**On ratification** — I've deliberately not acted on this, because it's a language-policy call rather than a code one, and it belongs to @seefeldb rather than to me. I've done the bookkeeping your other thread asked for so the decision is cheap to execute either way: the delta entry now names both normative records (matrix row 81 *and* lowering-contract §3.9), and all three rejected-set enumerations agree.

For whoever picks it up, the case you laid out looks sound to me on the merits — particularly that the JSX blessing already makes the current rule "no eager reads *outside* JSX", so extracting a JSX expression to a named binding changes a program's legality. That's a refactor-hostile rule, and it's the shape of the cascade that motivated this PR.

Two things I'd want in the follow-up beyond your list: the diagnostic message still says ".get() on a cell is not allowed in reactive context. Wrap the computation in computed(...)", which is now actively misleading at the sites where it does fire — the real rule is "this read has no lowerable site", and the advice should name that. And the plain-array-callback case should get a normative answer rather than inheriting whatever the gate happens to do, since this review is the second time it's surfaced as an edge.

— Claude Opus 5, on behalf of @seefeldb

seefeldb:
@mathpirate — ratified. @seefeldb made the call, so `b1fc2f6d` closes the decision as an official **yes**, following your concrete list.

- **Matrix row 81** splits in two, one per side of the rule: reads at a lowerable site are Supported (with the qualifying sites enumerated), reads with no site are Unsupported (statement position, reactive array-method callback, plain array-method callback).
- **§5.7's nuance paragraph folds into the normative text.** The bucket list gains the site bucket, and the compute-callback bucket narrows to what it's actually about, since JSX is now just one qualifying site among several rather than a special case. The closing text states the rationale you gave — a terminal read and a computation-feeding read are the same construct at the same site, and extracting a JSX expression into a named binding is a refactor, not a change of meaning.
- **The edges get normative answers** rather than inheriting behavior: parentheses, the computed-key spelling, and the optional spellings are named as not changing the classification; plain-array callbacks are named on the rejected side.
- **Lowering-contract §3.9** now names "eager `.get()` reads with no lowerable expression site to carry them, and eager reads on ordinary opaque values", pointing at §5.7 for the supported case.
- **The design-deltas entry resolves** (2026-08-07), recording what settled it rather than just that it settled.
- **§4.2's relocation pattern** was telling authors to relocate a form that is now supported. It relocates a site-less read instead, and says plainly that a read with a site needs no relocation. §4.1's "move elsewhere" example moved to statement position for the same reason.

I also took your diagnostic-message point, since leaving it would have shipped a ratified rule with a message naming the old one:

```
This .get() read has no expression site that can carry it, so it cannot be
lowered into a lift and would freeze to a one-time snapshot. Move it into
computed(() => myCell.get()), or read it at a site that lowers — a binding,
a return, an object property, or a JSX expression.
```

No test asserted the old text, so nothing was pinned to the wrong rule.

Local gates green again after the change: suite 1137/0, `fmt`, `lint`, `check`, `check-docs` (547 blocks), `cfcheck` (413 patterns), `check-skill-facts`, `check-conflict-markers`, `check-no-waitfor`.

Your approval predates the rebase and these two commits, so a fresh look would be welcome before this lands.

— Claude Opus 5, on behalf of @seefeldb

mathpirate:
The fresh look happened on the landed result (the merge outpaced the ask by a few minutes): the squash is faithful — zero diff between `b1fc2f6d` and `main` on every touched path — suite 1137/0 on `main`, the probe battery lands all ten shapes on the right side, and nothing in `docs/common`, the tutorial, or the skills still teaches the old rule. The ratification commit executes the list exactly, and the §4.1/§4.2 relocation examples were a catch we'd missed — both showed a form the language now accepts.

One nit found on the way through, fixed in #5745 along with the `linksView` unwrap this PR was motivated by: the delta entry dates its resolution 2026-08-07, which is when the question was raised — the call was made 2026-08-12, and the file's convention (cf. the optional-call entry) dates the resolving event.

That leaves `commentsView` as the last piece of the original motivation, waiting on `pattern-context:optional-chaining` for the `.toSorted` comparator — same over-broad-rejection shape, one diagnostic over. We're minded to pick that up next.

— Fable on behalf of Gideon

PR #5458, opened 2026-08-06
refactor(cli): extract the shared read step as a cell selection

mpsalisbury:
## Why

`cf piece get`, `cf piece call`, `cf wish` and `cf exec` are different ways of
*arriving* at a cell, but they all finish the same way: turn a cell plus a
caller-supplied selection into structured output. Today only `piece get` can
reach that step, because the path from address to value has piece-specific
machinery welded through the middle — config resolution, `--step`,
input-vs-result, the path walk, the verb-read refusal.

Three more callers arrive in later work, so this factors the shared step out
first. Sequenced as F1 in `docs/plans/shaped-reads-implementation.md`.

## What Changed

`packages/cli/lib/piece-get-transform.ts` → `lib/cell-selection.ts` (git tracks
it as an 88%-similarity rename), exporting:

```ts
deriveSelectedValue(runtime, space, sourceCell, selection, deps)
```

Selection vocabulary throughout: `CellSelection`, `SelectionProjection`,
`parseSelectionFilter`, `parseCellSelectionOptions`, and so on. The old name
asserted a piece where none is required.

`runtime` and `space` stay **explicit parameters** rather than being derived
from the cell. `runtime` is not on the public `Cell` interface, and the caller's
space is not always the cell's — `getCellValue` passes `manager.getSpace()`
while a path crossing a link can land the cell elsewhere, so deriving would
silently relocate where the transform pattern runs.

What stayed with `piece get`: config resolution, `PiecesController`, `--step`,
input-vs-result, the path walk, `classifyReadPathVerb` / `PieceVerbReadError`,
and the `--step`-naming error text.

## Test plan

**No test expectation moved**, which is the exit criterion for a refactor.
Verified mechanically: reversing the renames over the new test file reproduces
the original exactly, bar one import changing position under `deno fmt`.

`deno task test` in `packages/cli` (the only package touched) — pass.
`check-docs`, `check-conflict-markers`, `check-skill-facts`, `deno fmt --check`,
`deno lint` — all clean.

## Note for the next change

The unsafe-host-trust reason still reads `"cf piece get filter/schema computed
projection"`. Accurate while `piece get` is the only caller; it wants to become
caller-supplied when a second one lands.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5458?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cli/commands/piece.ts
+++ b/packages/cli/commands/piece.ts
@@ -49,11 +49,9 @@ import ports from "@commonfabric/ports" with { type: "json" };
 import type { PiecePatternRef } from "@commonfabric/piece/ops";
 import { reservesStdoutForCommandOutput } from "../lib/json-output.ts";
 import {
-  parsePieceGetFilter,
-  parsePieceGetProjection,
-  type PieceGetTransform,
-  PieceGetTransformError,
-} from "../lib/piece-get-transform.ts";
+  CellSelectionError,
+  parseCellSelectionOptions,
+} from "../lib/cell-selection.ts";
 
 // Hint system: print helpful next-step suggestions after operations
 let quietMode = false;
@@ -78,21 +76,6 @@ export function normalizeApiUrl(apiUrl: string): string {
   normalized.hash = "";
   const href = normalized.toString();
   return basePath ? href : href.slice(0, -1);
-}
-
-export async function parsePieceGetTransformOptions(options: {
-  filter?: string;
-  schema?: string;
-}): Promise<PieceGetTransform | undefined> {
-  const filter = options.filter === undefined
-    ? undefined
-    : parsePieceGetFilter(options.filter);
-  const projection = options.schema === undefined
-    ? undefined
-    : await parsePieceGetProjection(options.schema);
-  return filter === undefined && projection === undefined
-    ? undefined
-    : { filter, projection };
 }
 
 function summarizeForDisplay(value: unknown): unknown {
@@ -184,7 +167,7 @@ export function localPatternEntry(
  * A `piece get` failure caused by a data condition rather than bad arguments:
  * a path that doesn't resolve, a result schema that can't project stored data
  * (PieceResultProjectionError), a filter/projection that doesn't fit the
- * selected value (PieceGetTransformError), or a path that lands on a verb
+ * selected value (CellSelectionError), or a path that lands on a verb
  * (PieceVerbReadError — the read-path guard's redirect at `cf piece call`).
  * Reported as a plain error on stderr with exit 1, never as a Cliffy
  * ValidationError (which would dump the usage screen and read as an arg-parse
@@ -192,7 +175,7 @@ export function localPatternEntry(
  */
 export function isPieceGetDataError(error: unknown): error is Error {
   return error instanceof PieceResultProjectionError ||
-    error instanceof PieceGetTransformError ||
+    error instanceof CellSelectionError ||
     error instanceof PieceVerbReadError ||
     (error instanceof Error &&
       error.message.startsWith("Cannot access path"));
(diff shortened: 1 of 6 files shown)


mpsalisbury (commented):

**PR #5458 · refactor(cli): extract the shared read step as a cell selection · 6 files, packages/cli only · coverage: deep-read (entire diff + head-state verification) · checks run: none needed (no suspected behavior change to confirm by execution)**

The pure-refactor claim holds. I read every hunk: the changes are 1:1 symbol renames, one function moved verbatim (`parsePieceGetTransformOptions` → `parseCellSelectionOptions`, body identical), new JSDoc, and test-description rewording. Every user-facing error string, the `--step` guidance, the result cell's cause key (`pieceGetTransform`), and the unsafe-host-trust reason are byte-identical. The only runtime-observable delta is `Error.name` (`"PieceGetTransformError"` → `"CellSelectionError"`), and I verified it never reaches the user: `pieceGetDataErrorReport` (packages/cli/commands/piece.ts) prints only `error.message`, and no code matches on the name string. The API-shape decision is sound: `runtime` is genuinely absent from the public `Cell` interface (checked `ICell` and its constituent interfaces in `packages/api/index.ts` at head), and the space-mismatch rationale is real — `getCellValue` passes `manager.getSpace()` while `targetCell = rootCell.key(...path)` can cross a link into another space. Test files: renames only, expectations untouched; fixture strings that changed are never asserted on. No importers outside `packages/cli` of the removed `commands/piece.ts` export. `check-skill-facts` is safe — no skill, rule, or AGENTS file cites the old path.

**🟡 Improvements**

1. **Two live docs cite the deleted path — and nothing in the stack fixes them** *(verified)*.
   `docs/plans/shaped-reads-and-verb-results.md:99` cites `FORBIDDEN_PROJECTION_KEYS`/`UNSUPPORTED_PROJECTION_KEYS` as living in `packages/cli/lib/piece-get-transform.ts` — a statement about the current system that becomes actively wrong on merge. `docs/plans/shaped-reads-implementation.md:110` ("The existing transform in `packages/cli/lib/piece-get-transform.ts` is the body of it") goes stale the same way, and F1 itself becomes landed rather than pending. I checked all six stacked PRs (5470, 5497, 5500, 5504, 5505, 5469): none touches `docs/plans`, so the stale citations survive the entire stack, and no mechanical gate catches path citations in docs. Fix: repoint both citations to `lib/cell-selection.ts` in this PR (the base of the stack is the right place — later branches inherit it), and mark F1 landed in the plan.

**⚪ Nits**

2. `test/piece-get-transform.test.ts` keeps its old name while its module became `lib/cell-selection.ts`, breaking the "test findable from its source" rule in `docs/development/unit-test-coding-style.md` *(verified)*. Deliberate and correct here — the plan's F1 exit criterion is that this file's expectations stay untouched, which is the proof of the refactor — but it wants a follow-up rename, not silence. The comment at `cell-selection.ts:1579` names the test file, so the rename must update it too.

**Possible follow-ups**

- Rename `test/piece-get-transform.test.ts` → `test/cell-selection.test.ts` after the stack lands (plus the line-1579 comment).
- The PR body's "note for the next change" names only the unsafe-host-trust reason string (`cell-selection.ts:1393`) as wanting to become caller-supplied. Two more pieces of piece-get vocabulary sit in the generic module: the two user-facing `"Could not apply piece get transform: ..."` messages and the result-cell cause key `pieceGetTransform` (`cell-selection.ts` around lines 1564/1602 and 1533). I checked the end of the stack (`impl/s2-join-hash`): `deriveSelectedValue` still has exactly one caller there, so the deferral stays honest — but the cause key is the sticky one (renaming it later changes result-cell identity), so widening the note to name all three would keep the debt visible.

**Verdict:** Clean extraction with the API shape argued correctly and the pure-refactor claim verified against the diff — merge once the two plan-doc path citations are updated or consciously deferred.

mpsalisbury:
Superseded by #5470, which carries this commit.

`impl/a12-link-marker` contains `refactor(cli): extract the shared read step as a cell selection` — this branch's only commit, replayed on its base. That is why this PR conflicts with main while #5470 does not: #5470 was rebased and its conflict with the piece-controller refactor resolved, and this one was not.

Nothing is lost by closing it. The extraction landed with #5470.

PR #5459, opened 2026-08-06
feat(runner): teach entity-address intake the `of:` URI form

mpsalisbury:
## Why

`cf piece get --piece of:fid1:<hash>` throws today. `--piece` reaches
`entityIdFrom`, thence `FabricHash.fromString`, which splits on the first colon
and base64-decodes the rest — so `of:fid1:<b64>` parses as tag `of` with hash
`fid1:<b64>`, and the colon is not valid base64url. It surfaces as a decode
error, which is why it has never been reported as a missing feature.

That matters because an address emitted by one command should compose into the
next. Everything downstream that renders an address for a caller to reuse waits
on this. Sequenced as A4 in `docs/plans/shaped-reads-implementation.md`.

## What Changed

| Form | Before | After |
| --- | --- | --- |
| `my-board` (slug) | accepted | unchanged |
| `fid1:…` (bare hash) | accepted | unchanged |
| `of:fid1:…` | **throws** | **accepted** |
| `computed:fid1:…` | throws | **refused, by name** |

**Refused, not stripped.** Coercing `computed:` to its `of:` sibling renames an
id to a different entity — the hash preimage is kind-free, so a bare hash is not
a complete identity (`docs/specs/computed-cell-identity.md`). The refusal is
driven off `ENTITY_URI_SCHEMES`, so a new kind is refused without a further edit.

The seam is `hashStringForEntityAddress()` in `packages/runner/src/entity-kind.ts`,
called from `entityIdFrom`. `FabricHash.fromString` is untouched — it has
non-entity users, and `of:` is a URI scheme rather than a hash tag.

`pageIdForRouting` in runtime-client is **deleted** rather than duplicated; its
call sites pass addresses straight to `entityIdFrom`.

## One real bug fixed

`getCellByIdOrPiece` (`packages/piece/src/manager.ts`) compared a raw address
against `pieceId(entry)`, which always reports the bare form — so an `of:`
address could never match a registered piece. It now reduces once up front and
compares reduced forms.

## Audit: string-keyed lookups

Once two spellings resolve, anything keying on the *raw* input has two keys per
entity. Every string→cell path was checked.

- **Safe:** every CLI-side id in a comparison is cell-derived, not raw input.
- **Fixed:** `getCellByIdOrPiece`, above — the only raw-input comparison found.
- **Reported, not fixed:** `setBGPiece` upserts on an *external* `pieceId` from
  OAuth state or an HTTP body. Previously an `of:` id stored fine then hard-failed
  in the worker; now it works end to end, so two spellings could produce two
  registry entries. Normalizing at intake is right but changes stored-key
  semantics in another service.

## Adjacent, pre-existing, not fixed here

`getEntityId` already accepted schemed strings and routes through `fromURI`,
which **silently strips `computed:`** and yields the `of:` sibling — the exact
failure this change refuses, one function away. Low severity today, worth its own
issue.

## Test plan

New: `packages/runner/test/entityIdFrom.test.ts` (both spellings brand to one id;
a live cell reached both ways, a write through one visible through the other;
`computed:` throws naming the address) and
`packages/piece/test/piece-address-forms.test.ts`.

Confirmed the key assertion is not vacuous: `FabricHash.fromString("of:fid1:…")`
still throws on its own, so the test exercises the new path.

`deno task test` in runner (1157), piece (35), runtime-client, state-inspector
(96), cli — all pass. `check-docs`, `check-conflict-markers`,
`check-skill-facts`, `deno fmt --check`, `deno lint` clean.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5459?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/space-model/3-identity-and-references.md
+++ b/docs/specs/space-model/3-identity-and-references.md
@@ -107,6 +107,16 @@ which brands an existing content-hash string or `FabricHash`. A `FabricHash` has
 a tagged string form, `<tag>:<hash>` (e.g. `fid1:…`); construct one from that
 string via `FabricHash.fromString()`.
 
+`entityIdFrom()` is the entity-specific intake seam, so it also accepts the
+`of:`-schemed URI over a tagged hash (`of:fid1:…`) — the two spellings name the
+same entity, and both are in circulation wherever an id crosses a boundary a
+person can type into. A KINDED id (`computed:fid1:…`) is refused by name rather
+than stripped: the hash preimage carries no kind, so `computed:fid1:H` and
+`of:fid1:H` are different entities over the same hash bytes, and the bare hash
+is not a complete identity to fall back on (see
+[Computed Cell Identity](../computed-cell-identity.md)). The reduction itself
+lives in `hashStringForEntityAddress()`, which any other address intake shares.
+
 The underlying `hashOf()` function — see
 [Data Model](./1-data-model.md#hashing-and-content-addressing) for the hashing
 mechanism — is also used directly for:
--- a/packages/piece/src/ops/pieces-controller.ts
+++ b/packages/piece/src/ops/pieces-controller.ts
@@ -36,6 +36,7 @@ import {
   setPatternSource,
   type SpaceCellContents,
 } from "@commonfabric/runner";
+import { hashStringForEntityAddress } from "@commonfabric/runner/entity-kind";
 import type { CellScope } from "@commonfabric/api";
 import { cfcAtom } from "@commonfabric/api/cfc";
 import { StorageManager } from "@commonfabric/runner/storage/cache";
@@ -2171,10 +2172,17 @@ async function getCellByIdOrPiece(
   options?: { start?: boolean; targetScope?: CellScope },
 ): Promise<{ cell: Cell<unknown>; isPiece: boolean }> {
   const start = options?.start ?? true;
+  // The registry check below compares id STRINGS, and a registered piece
+  // reports its own id as the bare tagged hash, so an `of:`-schemed address
+  // has to be reduced to that one spelling before it can match. Reducing it
+  // out here also keeps a kinded address's refusal out of the catch below,
+  // which would otherwise report it as a missing cell. Messages keep `cellId`,
+  // so they echo the address the caller actually gave.
+  const hashString = hashStringForEntityAddress(cellId);
   try {
     // Try to get as a piece first
     const piece = await pieces.getPieceCell(
-      cellId,
+      hashString,
       start,
       undefined,
       options?.targetScope,
@@ -2196,7 +2204,7 @@ async function getCellByIdOrPiece(
     // cell ID
     try {
       const cell = await pieces.getCellById(
-        entityIdFrom(cellId),
+        entityIdFrom(hashString),
         [],
         undefined,
         options?.targetScope,
@@ -2209,7 +2217,7 @@ async function getCellByIdOrPiece(
         const id = pieceId(piece);
         // An entry without a piece ID cannot establish registration.
         if (!id) return false;
-        return id === cellId;
+        return id === hashString;
       });
       return { cell, isPiece: isRegisteredPiece };
     } catch (_) {
--- a/packages/piece/test/piece-address-forms.test.ts
+++ b/packages/piece/test/piece-address-forms.test.ts
@@ -1,0 +1,66 @@
+import { afterEach, beforeEach, describe, it } from "@std/testing/bdd";
+import { expect } from "@std/expect";
+import { createSession, Identity } from "@commonfabric/identity";
+import { Runtime } from "@commonfabric/runner";
+import { StorageManager } from "@commonfabric/runner/storage/cache.deno";
+import { createBuilder } from "../../runner/src/builder/factory.ts";
+import { PiecesController } from "../src/ops/pieces-controller.ts";
+import { pieceId } from "../src/piece-id.ts";
+
+const signer = await Identity.fromPassphrase("piece address form tests");
+
+describe("piece address forms", () => {
+  let storageManager: ReturnType<typeof StorageManager.emulate>;
+  let runtime: Runtime;
+  let pieces: PiecesController;
+
+  beforeEach(async () => {
+    storageManager = StorageManager.emulate({ as: signer });
+    runtime = new Runtime({
+      apiUrl: new URL(import.meta.url),
+      storageManager,
+    });
+    const session = await createSession({
+      identity: signer,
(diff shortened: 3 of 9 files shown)


mpsalisbury (commented):

# cf-review — PR #5459 "feat(runner): teach entity-address intake the `of:` URI form"

**Scope:** 9 files, runner/piece/runtime-client + one spec · **Coverage:** deep-read of the full diff and surrounding code at head; grepped head tree for every raw-string→entity intake (`fromString`/`fromURI`/`entityIdFrom`/ad-hoc `of:` handling) · **Checks run:** static review only — no suspected runtime bug required executing anything (PR body documents green runs in all five touched packages).

The change does what it says: one reduction seam (`hashStringForEntityAddress`), `of:` accepted, `computed:` refused by name, `pageIdForRouting` lifted rather than triplicated, and one real string-comparison bug fixed. The PR body's self-audit held up under independent verification — every claim I checked was accurate. The findings below are mostly at the edges the audit didn't reach.

## 🔴 Blocking

**1. The new spec sentence overclaims the seam — a normative spec left actively wrong on the exact hazard this PR exists to prevent.** — `docs/specs/space-model/3-identity-and-references.md` (added text) · **verified**
The added paragraph ends: *"The reduction itself lives in `hashStringForEntityAddress()`, which any other address intake shares."* False at head, twice over:
- `getEntityId` (`packages/runner/src/create-ref.ts:208`) is a raw-string address intake that routes schemed strings through `fromURI` (`packages/runner/src/uri-utils.ts:69`), which **silently strips `computed:`** and yields the `of:` sibling — the PR body itself names this, then the spec contradicts it.
- `packages/fuse/cell-bridge.ts:1491` hand-rolls its own two-spelling handling (`startsWith("of:")` / prefix), and would mangle a `computed:` rootName into `of:computed:…`.

Per the coherence north star: someone landing on this spec would believe every address intake refuses `computed:`; one function away, one silently aliases it. **Fix (one line):** scope the claim to what's true — e.g. "…lives in `hashStringForEntityAddress()`, the seam new address intakes share" is still aspirational; better is to name what shares it today (`entityIdFrom`, `getCellByIdOrPiece`, the source-change confirmation key). Routing `getEntityId`'s string branch through the seam is the real endpoint but changes behavior for a builder-exposed helper — fine to defer to the issue the body proposes, as long as the spec stops claiming it's done.

## 🟡 Improvements

**2. The "one real bug fixed" has no test.** — `packages/piece/src/manager.ts:1184` (`getCellByIdOrPiece`) · **verified**
Its only caller is `PieceManager.link()` (line 1134, target side). `piece-address-forms.test.ts` covers `manager.get()` — which exercises `entityIdFrom` acceptance, not the fixed `id === hashString` registry comparison or the `isPiece` classification. Revert the fix and no test fails; the failure mode is silent (an `of:`-addressed registered piece classified `isPiece: false`, so `link()` skips result/argument resolution and links the wrong cell). Fix: a test that links to an `of:`-spelled target and asserts piece-linking semantics. Direction-correctness of the fix itself is confirmed: `pieceId(entry)` is always bare because `Cell.entityId` derives via `fromURI` (`packages/runner/src/cell.ts:2966`), so reduced-vs-bare matches both ways.

**3. The `setBGPiece` deferral is weaker than the body implies — normalizing at intake is nearly free.** — `packages/background-piece-service/src/utils.ts:59` · **verified**
The upsert predicate compares raw `pieceId` strings, and `space-manager.ts` keys `enabledPieces`/`desired` on the raw stored id — two spellings of one entity means two registry entries **both scheduled**, i.e. the piece runs twice per cycle (before this PR the `of:` entry was dead: `worker.ts:274` `entityIdFrom` threw). But `hashStringForEntityAddress` is a fixed point for every already-stored bare key; the only stored keys whose semantics change are `of:`-schemed entries, which were permanently broken anyway. Suggest a fast-follow normalizing at `setBGPiece` intake rather than an open-ended issue.

**4. Plan bookkeeping: A4 lands but neither plan records it, and one now cites a deleted symbol.** — **verified**
`docs/plans/verb-result-selection.md:282` says `pageIdForRouting` "already implements exactly that shape; lift it" — this PR deletes that function, leaving a live plan citing a symbol that no longer exists (no mechanical gate catches prose in `docs/plans/`). And `docs/plans/shaped-reads-implementation.md`'s own doc-follow-through table (line 445) names A4's documentation as "the CLI README and the tutorial's workflow chapter" — untouched here (nothing there is actively wrong, just the plan's named follow-through). Update both plans in this change, or say explicitly the doc row is deferred.

## Focus-area answers (all verified)

- **Only intake path?** No — `getEntityId` (silent `computed:` strip) and fuse's hand-rolled handling remain; display-side helpers (`shell/debug-utils.ts:198`, `HeaderView.ts:577`) are guarded and benign. The audit's CLI claim held: no raw-id string comparisons in `packages/cli`/`packages/piece` beyond the one fixed.
- **`pageIdForRouting` deletion:** all five base call sites converted; behavior identical for every input class (`of:` strip, bare pass, `computed:` throw, unknown scheme/junk throws downstream in `FabricHash.fromString`, `data:` throws same as before). Only the `computed:` error *message* changed; the test was updated honestly (still asserts a throw), not bent. `handlePageRemove` never used it — `cc.remove` resolves via `entityIdFrom` + cell equality, so it gains `of:` support consistently. No routing regression.
- **Slug ambiguity:** none possible — `validateSlug` rejects `:` outright and `SLUG_PATTERN` is `^[a-z0-9]+(?:-[a-z0-9]+)*$`; `isSlugAddress` is defined as colon-free (`packages/runner/src/slugs.ts`).
- **Refusal vs strip:** the inconsistency is real and one function away, but the *code* shipping refusal-first is the right direction; the misleading surface comes from the spec sentence (finding 1), not the code.

- `favorites-manager.ts:216` `#createPieceRef` unconditionally prefixes `of:` — an `of:`-spelled `pieceId` (routing accepted it before and after this PR) yields `of:of:…`. Pre-existing; same family as the audit's registry.
- Adopt the seam in `fuse/cell-bridge.ts` rootName intake (also fixes its `computed:` mangling).
- The `getEntityId`/`fromURI` silent-strip issue the body proposes — worth filing now so finding 1's spec wording has something to point at.

**Verdict:** Sound change at the right seam with an unusually honest self-audit — fix the one-sentence spec overclaim (🔴) and preferably add the missing `link()` test before merge; the rest can follow.

PR #5460, opened 2026-08-06
feat(cli): mint a call session and carry it to the send boundary

mpsalisbury:
## Why

`cf piece call --invocation <id>` lets a caller name a dispatch so a retry
returns the original outcome instead of executing again. That id decides where
the receipt lands: `scopeCallerEventId` hashes `{caller, id, path, space}` where
`caller` is the id *string*, not a principal. So no identity enters the address —
two callers picking the same id against the same verb compute one address and
read one receipt, and a guessed id reads someone else's outcome.

The fix is to scope ids to a caller-supplied **session**, not a principal: agents
are directed to work under their human user's key rather than mint their own, so
the collision that actually happens is two agents under one key. A session
distinguishes them; a DID does not. And because a session is *minted* rather than
derived, it is unguessable, which closes the second half — a DID is public.

This PR builds the mechanism and **deliberately does not use it**. Sequenced as
S1 in `docs/plans/shaped-reads-implementation.md`; S2 joins it to the hash.

## What Changed

**`cf session new`** emits a bare random id on stdout. No file format — unlike
`cf id new` there is no key material, and a keyfile shape would only be
cargo-culted.

**`--session` / `CF_SESSION`** on `cf piece call`, following the existing
`--identity`/`CF_IDENTITY` shape exactly. Flag wins over env; blank is refused,
mirroring `--invocation`.

**Threaded** from `resolveInvocationSession` through `CallableExecutionDeps` to
the `Cell.send` options beside `eventId`, and into `StreamSendOptions`.

## Where it stops — the point of this PR

`session` is carried to the send boundary and **never read**. It reaches neither
`scopeCallerEventId`, nor the queue options, nor the event value, nor storage.
Every address derived today is byte-identical. That lets the mechanism be
exercised before the change that makes it matter, and keeps the one
address-changing commit small and alone.

One adjacent edit: the `declare module "@commonfabric/api"` augmentation in
`cell.ts` carried two inline copies of the send-options shape, which is what the
CLI type-checks against. They were replaced with `StreamSendOptions` rather than
adding a third and fourth copy of the field.

## Test plan

New: `packages/cli/test/session.test.ts` (distinct over 1000 mints; random-UUID
shape, which a counter, a timestamp or a v1/v7 id fails; bare single-token
output), plus flag-beats-env resolution through the real command, and the session
arriving in send options beside the id.

No test pins "the address is unchanged when a session is passed" — that fact is
exactly what S2 inverts, so such a test would be written to be deleted. The
invariant is visible instead as the absent read in `cell.ts`.

`deno task test`: cli 123, runner 1156. `deno task check` repo-wide,
`check-docs`, `check-conflict-markers`, `check-skill-facts`, `fmt`, `lint` — clean.

## Follow-up owed

`docs/common/verbs-over-the-cli.md` documents `--invocation` and wants its
`--session` paragraph when S2 lands and the pairing becomes required.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5460?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cli/commands/main.ts
+++ b/packages/cli/commands/main.ts
@@ -9,6 +9,7 @@ import { fuse } from "./fuse.ts";
 import { init } from "./init.ts";
 import { inspect } from "./inspect.ts";
 import { piece } from "./piece.ts";
+import { session } from "./session.ts";
 import { space } from "./space.ts";
 import { identity } from "./identity.ts";
 import { test } from "./test-command.ts";
@@ -167,5 +168,6 @@ export const main = new Command()
   .command("completion", completion)
   .command("id", identity)
   .command("init", init)
+  .command("session", session)
   .command("test", test)
   .command("wish", wish);
--- a/packages/cli/commands/piece.ts
+++ b/packages/cli/commands/piece.ts
@@ -454,6 +454,33 @@ export function resolveInvocationId(
 }
 
 /**
+ * Resolve the session a handler call's invocation id was chosen within: the
+ * caller's own session (`--session`, or `CF_SESSION`), or none.
+ *
+ * Nothing is minted for a caller that named no session. A session only means
+ * something if the SAME one accompanies every call of a run, and a per-call
+ * mint would be a fresh one each time — a value that looks like a session and
+ * relates no two calls. So an absent session stays absent, and travels as
+ * `undefined`.
+ *
+ * A blank session is rejected for the reason a blank invocation id is: it
+ * would read as "the caller named one" while naming nobody.
+ *
+ * The resolved session is carried beside the invocation id (see
+ * `newSessionId`, packages/cli/lib/session.ts); the event address the id
+ * derives is not scoped by it.
+ */
+export function resolveInvocationSession(
+  raw: string | undefined,
+): string | undefined {
+  if (raw === undefined) return undefined;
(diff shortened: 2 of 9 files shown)


mpsalisbury:
Superseded by #5469, which carries this commit.

`impl/s2-join-hash` contains `feat(cli): mint a call session and carry it to the send boundary` — this branch's only commit, replayed on a different base. That is why the two branches conflict rather than stack: both add `packages/cli/lib/session.ts`, and both touch `commands/piece.ts`, `lib/callable.ts`, and `test/piece-call.test.ts`.

Nothing is lost by closing this. Minting a session and joining it to the receipt address are hard to review apart anyway — the first is only meaningful because of the second, and #5469 shows both.

The flag and its environment variable are also being renamed to `--invocation-session` / `CF_INVOCATION_SESSION`, since `cf inspect` already takes `--session` for a storage session. That rename lands on #5469.

PR #5461, opened 2026-08-06
feat: server-execution v2 stage G — effect serving + the outbox

seefeldb:
## Why

Phase 1's stage list ends here: A–F built the serving loop and left it dark (#5439 is F, the loop itself); **G is the last stage — the effect channel** — and with it Phase 1's seven-stage cut is complete (plan checkbox ticked, and the phase's remaining OFF-arm success criterion ticked with it). Without G, the ON arm serves only pure derivations: any served piece touching `fetch*`, `generate*`, or `sqlite*` either fired effects at seal time (before its wave durably existed) or lost its writebacks to the §3d unstamped-seal refusal, and cross-space appends had no durable carrier at all. G lands serving-loop §4–§6's contract so the loop **performs egress exactly once per memo key, survives crash/park by re-miss, and never loses a cross-space append on any crash schedule** — the machinery Phase 2's ON milestone and Phase 3's events both stand on. Stacked on stage F (#5439); the train A #5339 → B #5349 → C.1 #5356 → C.2 #5367 → C.3 #5369 → D #5371 → E #5374 → F #5439 stays deliberately unmerged.

## What

**Effects defer to the outbox and fire POST-wave-commit** (serving-loop §3/§5). A sealed transaction's post-commit effects are no longer flushed at seal ("ok" there means accepted into a wave, not durable): `TransactionSealDestination.deferSealedEffects` hands them to the per-space `SpaceOutbox`, which admits them after the wave commit step with in-flight dedupe per effect key (`${kind}:${inputHash}` — the memo key), run-context carriage captured at the original run's seal, and live §7 counters. Bare accumulators (stage-D tests) and the OFF arm keep today's inline flush.

**Completion commits** (§4). The effectful builtins' writebacks — claim markers, results + `requestHash`, error-shaped results — are marked with their effect key (`markEffectCompletion`, a WeakMap the OFF arm never reads) and the SpaceServer routes marked transactions to their OWN derived-class commit: never through §3d's sealing, no basis rows, `derivedThrough = W`, empty `consequenceOf`, identity annotations sourced from the **outbox carriage captured at the original run's seal** (T7.Q4's journey, pinned with a scoped write + acting identity), dirtiness injected in-process via the sealNative promotion + the executor-commit report. An UNMARKED writeback keeps stage F's loud unstamped refusal — a missed marking fails loudly, never mis-classifies. FP6's label basis rides structurally: the writeback re-reads the request inputs (its hash guard), so the CFC ladder derives completion labels from the request basis through the same machinery as the OFF arm (Flags 5).

**The builtins' own memo IS §4's hit rule.** `fetch*` (internal-cell `inputHash`), `generate*` (`requestHash` on the result cell — the precedent §4 generalizes), and `sqliteQuery` (`requestHash` on its result) already implement recompute-key-vs-stored-key; G makes them REPORT hits (`effectMemoObserver` — memo.hits) and closes two serving-posture holes: a claim now skips when the stored hash matches and a result/error already landed (the deferred-flush window: a stale action re-run can re-admit a key whose effect completed and retired from in-flight — captured live as a double-fetch 19 ms apart; the guard alone did NOT close this window — the re-admitted claim can read a view where the completion is not yet absorbed — and the independent review's readability-gated in-flight retirement is what closes it, with this guard as the second layer), and `sqliteQuery` — the one builtin whose memo KEY commits AHEAD of its result — distinguishes a settled result (hit) from a bare orphaned claim, which re-issues under the serving posture only (`sqliteQueryMemoDecision`; the OFF arm keeps today's committed-state dedupe byte for byte). LLM partial-token writes stay deliberately UNMARKED: refused under the flag, which IS protocol §6's settled-results-only baseline; the OFF arm commits them as today.

**The durable outbox (FP1)**: a new `execution_outbox` engine table (blocklisted in the sqlite guard). Rows are written INSIDE the wave's own engine transaction (`applyWaveCommit` gains `outboxAppends`; an appends-only wave commits via a guarded `allowEmptyOperations` — losing the rows to an empty-commit refusal would lose the appends with nothing to re-emit them), staged per run via `WaveAccumulator.enqueueOutboundAppend` and folded for SURVIVING contributions only (a per-doc-superseded survivor's appends still ride — the §3d drop re-arms nothing, so nothing would ever re-emit them; a dependency-dropped or requeued contribution re-runs and re-emits). Delivery (`Server.commitDelegatedAppend`) admits at the target under the delegated row — `firedAt` derived from the SAME carried actor admission validates (one source, no mismatch), envelope = the service session (LT5), eventId dedupe ONLY above the stream's `eventWatermark` (the model's `admitDelegatedAppend` semantics; at-or-below duplicates admit) — then deletes the row: **admit-before-delete**, so any crash schedule re-sends and dedupes (the C2/FP1 closure). Activation re-sends pending rows (§6 step 5); post-wave drains ride the cycle (with a drain-owed flag so transport-failed rows re-drain without fresh appends); an LT4 deterministic rejection retires the row + counts `outbox.failed` (its source-side events.md §5 notice is Phase 3's machinery). Userless/grantless appends are refused fail-closed at the SOURCE (Flags 3). Row retirement is a direct engine-table delete — serving-loop §1's plane (c) sentence updated in this PR (docs-move-together), with the register row.

**The stage-D sqlite bound discharged** (the last one — the sink header's bound list now reads all-discharged): folded `sqlite` ops in HOME wave batches get per-RUN scope keys from the accumulator (M1 — the wave envelope has no session to resolve scoped files from), and the sink attaches through the memory server's new `attachWaveCommitSqliteDbs` (the transact path's `#attachCommitSqliteDbs` refactored over one shared core — same validations, caps, ≤1-db rule, disk-source refusal — two identity sources), synchronously around the apply (attach → apply → detach, the B1 ≤1-attached discipline). Hook-less sinks, key-less ops, and FOREIGN batches with sqlite ops refuse loudly (foreign sqlite is Phase 5's cross-space design).

**§7 counters real**: `memo.*` and `outbox.*` live (stats.ts's structurally-zero note replaced with exact semantics: misses = admitted effects, hits = stored-key resolutions reported by fetch*/generate*/sqliteQuery — fetchProgram/llmDialog count misses but report no hit events yet; outbox.failed = infrastructure failures, effect-level failures are §4's error-shaped RESULTS). `events.*` stays Phase 3, `effectAcks` Phase 4.

**OW7 landed**: trace **T14 — effect failure and retry: input-driven, never timers** authored in the house style with reference answers + citations (scenario-traces §3/§4; the trace-count headers in scenario-traces and the README table move 12 → 14, folding in the stale count cubic flagged on #5439); the owed-register entry flipped, with the serving-loop failure-leg test as the impl witness.

**ON-arm skip list**: re-justified in place — the ONE entry (the two-browsers Phase 2 gate) stands on its original two-deriver-interim reason, unrelated to effect serving; stage G adds NO skips.

Out of scope, stated: `compile-and-run` (README §3.5's compile class, not §4's effectful class) keeps stage F's posture — its async writebacks refuse under serving until its own port; Phase 3 events / Phase 4 enact-ack / Phase 5 cross-space / Phase 6 budgets untouched; the #5439 ledger's OPEN delegated scoped-READS question untouched (adjacency noted in Flags 7).

## Verification

- **spec-model 19** green (the outbox matches the model: durable rows survive the crash step, delivery = `admitDelegatedAppend` semantics, effect half process-local).
- **Full runner suite: 1115 tests / 6029 steps green** (stage-F baseline 1114/6016; the delta is the new coverage). New: `executor-outbox.test.ts` (10 steps — the defer-vs-inline seam; outbox admit/dedupe/carriage/counters; infrastructure-failure counting; FP1 rows landing in the wave tx for survivors only incl. the appends-only wave; delivery with carried-actor `firedAt` + LT5 envelope + row retirement + re-send horizon dedupe with NO second entry; source-refusal of userless appends; LT4 retire-not-retry; **admit-before-delete pinned by a transport-failure proxy** — the row survives, the next drain delivers exactly one entry; the sqlite discharge round-trip incl. all three refusal arms; the `sqliteQueryMemoDecision` arms), and two serving-loop E2E journeys: **the effectful node behind memoization** (a real `fetchJsonUnchecked` pattern served under a stubbed egress: ONE external call per key through miss → outbox → completion → client-observed result; memo.hits from the live loop; park/re-activate bounded per T10.Q4's RULED at-least-once; OW7's failure leg — error-shaped result, a 250 ms no-timer-retry window, input-driven re-fire) and **the completion commit** (its own derived commit, holder + `derivedThrough`, annotations from the seal-time carriage incl. the scoped `scope_key` + acting identity, separate from the wave's commit, client push observed). After the INDEPENDENT review's readability-retirement fix, the E2E soak on the final tree ran **10/10 consecutive green** (the 12/12 claim that stood here was not reproducible — the independent review measured 6/10 red at this PR's prior head under box load, all double-egress; post-fix, zero double-egress failures across 35 soak runs, the only residual flakes being load-sensitive stage-F lease-timing timeouts).
- **Full memory suite: 461 tests / 145 steps green** (schema inventory + sqlite-guard blocklist updated for `execution_outbox`).
- `tasks/check.sh` (repo typecheck), `deno lint` (4321 files), `deno fmt --check` (4351), `check-docs` (522 blocks), `check-conflict-markers`, `check-no-waitfor`, `check-skill-facts`, and the skip-list unit tests (10) all green.
- **Integration, live**: runner suite BOTH arms green (flag-OFF and flag-ON toolshed — the ON arm now serving with live effect machinery), re-run green after the review batch; runtime-client both arms green; patterns `counter` and `cfc-group-chat-demo-multi-runtime` green ON. `patterns lunch` ON failed **locally** on this box — and reproduced IDENTICALLY at the stage-F base af7ed64ce (same two-browser CAS-storm timeout signature), so it is the documented two-deriver-interim behavior under local load, not a stage-G regression; lunch OFF is green in this worktree.
- **OFF arm**: byte-identical up to the ratified key-vocabulary §5 acceptances — the full runner + memory suites are the witness; the builtin marking is a WeakMap write no OFF-arm code reads, the observers are absent-by-default, the defer branch requires an installed seal destination, and `allowEmptyOperations` is unreachable from every session-facing path. One deliberate near-exception is Flags 6 (a race-corner-only claim guard).
- Honest bound: the served-sqlite journey ACROSS park is covered at the decision-function + wave-batch level; a full integration-level served-`sqliteQuery` park journey (needing the real db-handle flow) is owed to the `sx2-` integration surface alongside Phase 2's bring-up.

## Flags

Per the flag-don't-fill rule — conservative implementations whose exact semantics the specs leave unstated; none blocks the dark landing:

1. **serving-loop §3d's "acked-effect retirement when stage G lands it"** parenthetical: the retirement of acked effect ENTRIES is protocol §5's effects-doc machinery — Phase 4's navigateTo channel, which this stage deliberately does not build. Stage G's analogous write is outbox-ROW retirement, an engine-table delete (plane (c), edited here), not a stamped commit write. Should §3d's parenthetical say Phase 4 (or name the row delete instead)?
2. **builtins.md's fetch row prescribes migrating `requestHash` onto the result doc**; today it lives on the committed internal cell. Migrating changes OFF-arm cell shapes (the builtin is shared code), so stage G leaves it — the memo is functionally equivalent (the hash IS committed state, T10.Q1). When, if ever, should the migration run, and under what OFF-arm ruling?
3. **Userless delegated appends**: events §2 legalizes userless space-scope emissions; the engine's delegated completeness floor (stage F) refuses an empty acting principal; the spec model's `admitDelegatedAppend` admits them. Stage G refuses fail-closed at the SOURCE (`enqueueOutboundAppend` throws), so nothing durable is ever silently destroyed — but the three-way disagreement needs a Phase-3 ruling (floor carve-out for appends vs a service-actor convention), plus model reconciliation.
4. **llm-dialog's tool-call cell mutations** (pin/unpin, update-argument, the pattern/handler tool-invocation writeback) are left UNMARKED — refused loudly under serving. Are those completion-class writes, handler-class consequences, or Phase-3 events? The turn lifecycle itself (claim/result/error via `safelyPerformUpdate`, the pinnedCells merge) IS marked and serves.
5. **FP6's label basis rides structurally**, not as a frozen snapshot: the completion transaction re-reads the request inputs (its hash guard), so labels derive from the CURRENT basis of value-identical inputs — identical machinery to the OFF arm. A re-labeling that changes labels without changing values would thus flow the NEW labels into a completion for the old response. Ratify, or require a frozen label-basis snapshot in the carriage.
6. **`tryClaimMutex`'s result-present guard is ungated** (both arms): load-bearing under serving as the SECOND layer of the deferred-flush window's closure (a real double-fire was captured before it; the independent review then showed this guard alone does not close the window — the re-admitted claim can read pre-absorption state — and the readability-gated in-flight retirement is what closes it), and in the OFF arm it only skips a redundant cross-tab refetch inside a race corner no suite pins; no deterministic OFF-arm outcome changes (full suites witness). Ratify under testing §2's gate as amended, or direct a serving-posture gate (which would fork builtin behavior per arm).
7. **Delegated scoped-READS adjacency** (the #5439 ledger's OPEN question — untouched): `resolveSqliteOpOwner`'s best-effort owner backfill reads scoped handle docs with the ENVELOPE identity; under the wave envelope it fails closed (a `dbOwner()` rule then refuses). Same identity class as the open question; behavior preserved, noted here so the eventual ruling sweeps it too.
8. **LT4's source-side failure notice**: a deterministic append rejection retires the row and counts `outbox.failed`; the events.md §5 notice on the source event's stream entry lands with Phase 3 (no stream-entry machinery exists to annotate yet).

## Self-review (standing merge protocol)

An adversarial subagent review of the full diff ran after green (spec conformance section by section, FP1 crash-schedule soundness, OFF-arm neutrality, test vacuity; report at the worktree channel `../stage-g-review-report.md`). It found **1 BLOCKER, 4 MAJOR, 9 MINOR — all twenty addressed** in the follow-up commit (`fix+test+docs: address the stage-G adversarial review batch`):

- **BLOCKER: a served `sqliteQuery` could wedge permanently** — its memo key commits AHEAD of its result (unlike every other effectful builtin), so a dropped effect (park/crash/discard) left a durable claim the hit branch honored forever, with no error result and no timer. Fixed: the memo decision distinguishes settled results from bare claims; an orphaned claim re-issues under the serving posture only (the stamped-run signal keeps the OFF arm byte-identical); hits are only reported for settled results. Unit-pinned.
- MAJORs: userless appends now refused at the source instead of silently destroyed at delivery (M1 → Flags 3); admit-before-delete pinned by a transport-failure test (M2); the dead `dispositionOf` accessor with a contract contradicting shipped behavior removed (M3); the E2E's recovery assertions re-pinned to T10.Q4's RULED at-least-once — and the probing surfaced a REAL race the review demanded scrutiny of: a stale re-admit after fast completion could claim over a landed result and re-fire (captured live); first patched with the claim-time result-present guard (M4 → Flags 6), which the independent review showed insufficient — the readability-gated in-flight retirement in the independent-review batch is what closes it.
- MINORs: LT4 retirement guarded against delete-throws; a drain-owed flag re-drains transport-failed rows without fresh appends; post-seal append staging refused + sealed arrays copied; the effect admit moved ahead of the throwable commit-record fetch; comment-truth fixes (delivery localSeq, the Phase-3 horizon obligation, stats semantics); the completion-tx effect path documented.

The review confirmed clean after scrutiny: the §8 tripwire scan, protocol §7's closed metadata list (outbox rows are engine-table carriage, never wire; no effect state in the durable table), FP1 atomicity + crash schedules (a)/(b)/(d), the completion path's §4 conformance and its `#sealChain` serialization, the sqlite discharge's validation parity with the transact path, and OFF-arm neutrality across every touched shared file. Full verification re-ran green after the batch (suite counts above), including the E2E soak (see Verification for the honest post-independent-review numbers; the 12-run claim that stood here predated the independent review and did not hold under load).

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5461?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seefeldb:
## Independent adversarial review (coordinator round) — NOT-LANDABLE → fixed → independently RE-VERIFIED → landable

Reviewed at d0da8498c by a fresh-eyes agent under the treat-self-review-as-unverified protocol. Verdict was **NOT-LANDABLE** — the first of this train — and the protocol earned it: the stage's core exactly-once-egress contract failed intermittently at pristine HEAD (6/10 soak), because the claimed "closed" double-fire was not closed. Fix batch: `085fac460`; verification tail: `4edc9768a`. Because a prior "closed" claim proved hollow, the fix was then **independently re-verified by a third agent** before this ledger: 15-run soak, probe re-application, and a four-arm audit of the new barrier's lifecycle. Nothing merges — the train holds until the final phase exists.

**The blocker (B-1), mechanism found by instrumentation:** a re-admitted effect key's claim transaction read a stale view. The completion commit had landed engine-side, but its accept was PARKED in the replica (CT-1927: accepts wait for a covering frame marker), and a superseded re-run's rollback had re-exposed older state — its inputs-changed arm actively writes `result: undefined` with a matching `inputHash`, so any result-present guard is structurally defeated; and completions carry `basisSeq = NOW`, so the stale claim committed cleanly. Second egress ~20 ms after the first. For `fetch` the duplicate is value-idempotent; for `generate*` it is the double bill the design exists to prevent.

**The fix (adjudicated direction: readability, not CAS):** a new narrow replica API `ISpaceReplica.whenApplied(localSeq)` — the in-flight effect key retires only when the completion's writes are READABLE by the serving runtime (promotion, marker-channel death → synchronous promotion, or replica replacement → fresh re-pull), and until retirement a re-admit dedupes. Event-driven, no sleeps; completion-commit CAS semantics untouched (per the same-day ruling, §5's sentence was amended to describe the true idempotence mechanism: hash guards + all-no-op short-circuit + this retirement gate).

**Independent verification of the fix:** 15-run soak → 13 green, **0/15 double-egress** (combined with the fixer's runs: **50 runs, 0 double-egress**); all four revert-probes red-on-target; the barrier's four lifecycle arms audited SOUND (no park path escapes `whenApplied`; the straggler window resolves to the documented at-least-once posture; channel-death promotes synchronously before any barrier continuation runs; reset/close arms are crash-equivalent — a replaced replica re-pulls durable truth). The 2 soak reds are a characterized real-clock harness-budget flake (deadline-paced waves under load overrun a 15 s wait; zero product concern evidenced) — remedied in the tail commit (30 s loaded-box budgets + a triage log line on flush-deadline exhaustion).

**The other findings → fixes (085fac460):**

- **M-A** — two silent FP1 append-loss arms: a survivor sealing only FOREIGN spaces lost its outbound appends under a `home === undefined` guard, and a zero-seal emitter (the canonical Phase-3 pure-forwarding handler) never became a contribution at all. Fixed: the fold runs for every survivor, and a zero-seal tx with staged appends MINTS a zero-write contribution — matching the model, which folds cross-space cascades for `writes: []` contributions. Both pinned red-first.
- **M-B** — the production recovery seams (activation re-send; owed post-wave drain) had zero coverage — deleting the re-send left every test green. Fixed with SpaceServer-level tests, plus one real behavior fix the tests forced: a failed activation re-send now arms the owed-drain flag. The deletion probe now turns both tests red.
- **M-C** — the PR body's "12/12 green" and "closed at the claim step" overstatements corrected in place with the honest history.
- Minors: the park-race straggler path (whose real pre-fix bug was worse than the reported leak — an inline-flush firing network work for an uncommitted contribution against a disposed runtime; now drop-and-own), T7.Q4's structural-scope_key note, memo.hits arithmetic caveat, §3d's mis-attributed acked-effect-retirement cross-reference (→ Phase 4), builtins.md's requestHash-migration deferral note, llm-dialog's fire-and-forget refusals now awaited and surfaced, and OW14 (the LT4 notice-before-delete obligation, Phase-3 trigger).

**Owner rulings (all RULED 2026-08-05, recorded in this batch and spot-verified by the re-verification pass):**

1. **Flag 6 ratified** — the claim-guard's OFF-arm race-corner delta accepted; testing §2's gate clause widened to name the register's recorded-acceptance rows (now three recorded acceptances total).
2. **Flag 5 ratified** — FP6's label basis stays structural: tightening mid-flight yields the stricter label; loosening matches today's OFF-arm write-time derivation; a frozen snapshot was rejected.
3. **serving-loop §5 amended** — the "duplicate completion is a CAS no-op" sentence replaced with the real mechanism (basisSeq = NOW is deliberate; idempotence = hash guards + readability-gated retirement).
4. **Flag 3 shape ruled** — the userless delegated append three-way inconsistency resolves at Phase 3 as a floor carve-out: an absent acting principal admits iff the entry is declared sessionless-space-scope, stamped `firedAt = { session: "server" }`, grant presence mandatory (OW15 carries the implementation lifts).
5. **Flag 4 classified** — llm-dialog pin/unpin are completion-class turn-lifecycle state; update-argument is a handler-class consequence (Phase-3 events); OW16 carries the trigger.

**New finding from the tail's verification runs (flagged, tracked, triage in flight):** the residual `/two` soak flake is NOT fully load-explained — instrumented red runs show exactly-once egress ~300 ms after the input write, then the completion never becomes client-visible for the full 30 s while the space stays active, commits nothing, and records ZERO deadline-exhausted waves. A post-egress **completion-visibility wedge** in the completion-writeback → derived commit → push/frame → client-readable path (B-1-adjacent; the barrier audit's named residual — an accept parked awaiting a covering marker that never arrives — is the leading candidate). Safety held in every observed red (exactly-once, zero double-egress); this is a liveness item, ~4/55 runs, with its own triage running. Phase 2's visibility-dependent gates will not be claimed while it is open.

**Still open on this ledger:** the delegated scoped-READS question (reads in delegated batches resolve against the delegating envelope, not the carried actor — recommendation: fail-closed refusal when a producer arrives, keeping Phase 5's grant-scoped read design space open).

**Counts at the tip:** spec-model 19; runner 1117 (6039 steps); memory 461; check-docs 522. ON-arm skip list: one entry (the two-browsers Phase 2 gate). **Stage G closes Phase 1's stage list**; the Phase 1 close-out check runs against this tip.

seefeldb:
## Wedge addendum: root-caused and fixed — b4c8c94bc

The completion-visibility wedge flagged above is closed. Mechanism-level triage (123 runs, 14 wedges, instrumented timelines) overturned the leading theory — frames flowed and the client briefly HAD the result — and found a three-link chain, production-reachable and worse than a flake:

1. **Trigger** — a wave supersede-drop killed an input-transition run's `inputHash` write (completion commits land with `basisSeq = NOW` and routinely race an open wave's basis), leaving the engine's memo hash stale.
2. **The destroyer** — the completion writeback's hash write was ELIDED as a no-op against a doomed parked overlay, leaving durable state torn (`result` present + stale hash); the next run read "inputs changed" and wiped the already-delivered result.
3. **Permanence + leak** — outbox retirement parked forever on frame markers that engine-plane commits never stage, so the serving replica's accepts never promoted: every re-admit deduped into a dead in-flight entry (including a DETERMINISTIC no-race A→B→A starvation), and even green runs leaked one in-flight entry + one waiter per served effect.

**The fix pair (no spec-semantics change; §3d's drop rule untouched):**

- **F1a** — locally-committed (sealed/engine-plane) verdicts confirm immediately: the committed-verdict arm of `settleSealedCommit` routes to `confirmPending`. Structurally remote-safe: sealed commits are exclusively engine-plane, and the parking branch remains reachable only from the socket-transact call sites (CT-1927 intact by construction). The B-1 retirement barrier stays as belt.
- **F2** — completion writebacks are authoritative. The refinement the fix proved: NO builtin carried its own elision — it lived five layers deep in the shared write stack (the cell-diff equal-leaf emit; two tx-level value-equal skips; the doc-level current-equals-initial skip; the patch builder's diff), all now yielding under an explicit `markAuthoritativeWrites()` mode that `markEffectCompletion` activates — gated on a configured seal destination, so the OFF arm is byte-identical by construction and ordinary transactions keep their elision.

**Evidence:** three red-first regression pins (the A→B→A starvation; the torn-hash interleave via a real wave supersede-drop, non-vacuity pinned on `supersededWrites == 1`; retirement-liveness with the leak assertion — red after the FIRST effect at the old tip); soak **15/15 green** vs the ~11% red baseline; B-1's exactly-once pins and barrier tests all green; full runner 1117 (6042 steps), memory 461, spec-model 19, check-docs 522. The stale-read-machine regression test was not constructible post-fix — no parking path remains for sealed commits — recorded in the register instead.

Two accepted costs, documented in comments: an authoritative `undefined` write to an absent slot materializes present-but-undefined (reads unaffected), and a duplicate identical completion re-emits per-leaf ops (rare corner, idempotent at the store).

seefeldb (commented):
Verdict: changes required.

[major] Preserve every requester behind a shared effect key. In `packages/runner/src/executor/outbox.ts:164-190`, the key is only kind + input hash (for fetch, see `builtins/fetch.ts:492-520`). Two distinct recipe nodes/result cells can issue identical inputs. The first effect closure writes only its own pending/result/error cells; the `continue` drops the second closure, so the comment that one completion serves both result cells is false and the second cell can remain pending forever. Include target result/memo identity in the key, or retain/fan out all attached completion closures; add a two-node identical-input/distinct-results regression.

I also confirmed current open blockers around effects admitted after failed waves, effect-only quiet-space starvation, completion-write failure wedging pending state, replay duplicating durable outbox rows, transport failure violating per-stream FIFO, and unstable implicit-rowid deletion across an await. This PR currently records only the Cubic check, not the full Actions matrix.

seefeldb:
Fix ledger for the 2026-08-12 review and all open inline threads. Pushed `34bb029ef..30b4e8d9b` (parent F tip 2758e60fc merged first: 4 conflicts combined-intents — engine.ts keeps G's appends-only allowEmptyOperations alongside F's scope-key admission; post-merge runner suite fully green).

**All 19 threads + the headline fixed with evidence; none refuted.** Per-thread table in the fix report; highlights:

- **Headline [major] — per-target effect keys.** The shared key becomes `kind:hash@<result-cell id>` across all colliding builtins, with a two-node identical-input/distinct-results regression. Fan-out of attached closures was REJECTED on the merits: the "attached" closure is a second egress against its own cells, and shared carriage would hand it the wrong identity — two distinct recipe nodes are two distinct effect requests that happen to carry equal bytes. This surfaces a spec question (flag 1 below).
- Effect-only quiet-space starvation + failed-wave admission composed as an `aborted === undefined` gate (seq-based gating would re-starve vacuous closes), pinned at SpaceServer level; transport-failure FIFO now drain-stops; replay-gated FP1 inserts (`replayed: true`) stop duplicate durable outbox rows; declared-PK outbox id replaces the implicit-rowid instability; `tryWriteResult` failure propagates; quiet-cycle owed drain lands.
- **Completion-write wedge class (threads 16/17/18):** the F2 authoritative path had three residual cracks — equal-empty-container elision, forced replace unable to create ancestors a doomed overlay had created (deterministic retry-budget burn = the F2 wedge through a different crack), and the TransactionWrapper dropping authoritative mode + completion routing. Authoritative transactions now commit whole-doc set/delete (ancestor presence in durable state is undecidable client-side), the wrapper forwards mode and keys routing on the inner tx, and the now-unreachable forcedAsserts machinery is deleted.
- **Danger zone respected:** F1a and the whenApplied barrier untouched; only F2's forced-assert surface changed (strictly stronger). Wedge pins (A→B→A starvation, torn-hash, retirement-liveness) + 10/10 consecutive soaks green.

**Final bar at 30b4e8d9b:** runner 1149/0 (6280 steps), memory 434/0 + the six explicit `*-test.ts` files 56/0, spec-model 19/0, 10 consecutive serving-loop soaks, `deno check` clean.

**Flags for the owner (not filled):**
1. **serving-loop.md §4's "one outstanding effect per key per space; a second miss attaches"** conflicts with per-target keys for the multi-node case — though the sentence was already internally inconsistent (§4's own miss rule carries ONE result-cell address per entry). Amend to per-(key, result target), or rule that a response-sharing layer (one egress fanned to N writebacks, a builtin-contract restructuring) is owed. The per-target key is forward-compatible with both.
2. **tryClaimMutex swallows terminal claim-commit failures** — reads as "another tab is handling this", effect completes without egress; no durable wedge (recovery via §6 re-miss). Out of this round's scope; register row owed.
3. **stream-data** enqueues with an unscoped key and never calls markEffectCompletion — not in the served-effect class today (stage A disabled it under the flag); documented for whichever stage adopts it.
4. CI matrix note: root-caused earlier — the Actions workflow only triggers on main-based PRs; stacked PRs get no matrix. Not addressable from a fix round.
5. **Wave replay reachability:** an exact replay at the ORIGINAL basisSeq cannot pass applyWaveCommit's CAS re-verification (the first application advanced the heads); only re-derived-basis replays reach the replay return. If the crash-replay model expects original-basis replays servable, re-verification needs replay awareness — to be checked against the spec-model's recovery arms (queued for the P2 round).

**OFF-arm ledger (needs recorded-acceptance ratification):** (1) per-target idempotency keys make the tx-level outbox dedupe finer OFF-arm — two distinct nodes with identical inputs in ONE tx both flush where the second was silently dropped (strictly a bug-fix delta); (2) OFF-arm fetch writebacks whose commit fails terminally now reject the tracked-work promise instead of failing silently; (3) tryClaimMutex marker placement — OFF-inert, listed for completeness; (4) the authoritative-write changes are gated on serving posture — OFF byte-identical — with one accepted ON-arm widening: a serving completion's whole-doc write can stomp a concurrent client replica's claim fields (requestId/lastActivity), self-healing via the 5s mutex staleness bound and strictly better than a wedged completion; (5) llm hit-observer counters shift (observer-only).

seefeldb:
Ratification recorded: the owner ratified the §4 amendment and all five OFF-arm ledger rows (2026-08-13). Landed as 40381ed3e — serving-loop §4 now reads per-(key, result target) with the response-sharing future-note, and verification-coverage carries the five recorded-acceptance rows (including the ON-arm claim-field widening).

seefeldb:

PR #5462, opened 2026-08-06
docs: mark the walks that mishandle a `FabricSpecialObject`

Hixie:
A `FabricPrimitive` is an opaque leaf whose state lives in private fields, and a `FabricInstance` is a container whose contents are reachable only through its codec. Both kinds pass `isRecord()` and `typeof === "object"` with zero enumerable own properties, so a generic structural walk decomposes the former to `{}` and sees nothing inside the latter. Some such sites already carry `TODO(danfuzz)` markers; this change marks the rest that a fresh sweep found: 59 markers across 35 files. Identification, not repair — the diff is comment-only, verified by filtering it to non-comment lines.

The sweep went beyond record-rebuilding walks to the other shapes the same predicate hole produces:

- Naive equality on values: `deepEqual` compares class instances by enumerable own-props, so two same-class fabric values compare equal regardless of contents. The CFC enforcement points built on it (the sink-request release gate, exactCopyOf/projection claim verification, wildcard-policy change detection and const/enum conditions) fail open, as do the piece controller's committed-state check and the runner's remove()/removeAll() element matching.
- Search walks that miss instance contents: link pre-sync, dependency registration in pull(), write-redirect discovery, policy-input recording, durable-source contract projection, and supplied-link recording all end their descent at a `FabricInstance` and silently skip what it holds.
- Schema-`default` rebuilds: sanitization, interning/canonicalization, default merging, and the LLM-facing `$ref` inliner rebuild `default` values structurally, so a fabric-valued default comes out as `{}`.
- Storage path descent and write staging: own-key descents report instance contents absent (and disagree with the `in`-based descent in `getAtPath`, which resolves through prototype accessors), and the not-found write retry grafts properties onto an instance clone that its codec never encodes.
- Serialization boundaries and display: request snapshots and bodies, the LLM client payload, the VDOM text/style paths, wish projection, fuse's mounted-file rendering, piece console logs, and the state inspector's diff all render a fabric value as `{}`, silently.

Sites with no constructible reaching path, sites whose file already carries a covering marker, and sites with a fix in flight were left unmarked. Repo-wide `deno fmt --check` and `deno lint` are clean, and the test suites of all eleven touched packages pass.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5462?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/background-piece-service/src/worker.ts
+++ b/packages/background-piece-service/src/worker.ts
@@ -352,6 +352,12 @@ export async function runPiece(data: RunData): Promise<void> {
 // Logs here are often viewed through observability dashboards
 // that don't render objects well. Attempt to stringify any objects
 // here.
+//
+// TODO(danfuzz): this is an unsafe use of `stringify()` for piece console
+// arguments, which arrive live and in-process (the runtime's console capture
+// dispatches them without serialization): a logged `FabricSpecialObject`
+// renders as `{}`, silently. Wants a `FabricSpecialObject` test rendering
+// via `toCompactDebugString()` from `@commonfabric/data-model/value-debug`.
 export function safeFormat(value: unknown): unknown {
   if (value && typeof value === "object") {
     try {
--- a/packages/cli/lib/wish.ts
+++ b/packages/cli/lib/wish.ts
@@ -203,6 +203,13 @@ function projectNode(
   }
   if (value === null || typeof value !== "object") return value;
 
+  // TODO(danfuzz): the `typeof` gate admits a `FabricSpecialObject`, so the
+  // `Object.entries` rebuild below renders one — a `FabricBytes` in a
+  // materialized wish result, which the binary fetch builtin mints today —
+  // as `{}` in the projected output. Wants a `FabricSpecialObject` test
+  // returning the value whole, plus a fabric-aware rendering in the
+  // downstream `render()`/`safeStringify` step, whose own marker in
+  // `render.ts` predates this traffic and calls the path latent.
   const cached = memo.get(value);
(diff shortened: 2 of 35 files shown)


Hixie:
Prompt was:

> use labs D.
>
> There are a number of places in the codebase that incorrectly handle `FabricPrimitive` and `FabricInstance` (e.g. not treating the former as leaves or the latter as values that need to be traversed). Some cases of this have been marked with TODOs already. Find the rest and mark them in a similar fashion.

(labs D is which directory i wanted it to work in)

Used Fable 5 / Max.

danfuzz:
LGTM in general, on the assumption that for each site it was verified (ideally, code instrumentation) that there really is a path from something strongly identified as a `FabricValue` (e.g. actual literal `FabricValue`, content of a cell, result of a `data-model` call, etc.) to the marked sites. Some of the `TODO`s seem to imply that verification, but not all.

Hixie:
Follow-up prompt was "For each site where you added a TODO, verify (empirically, e.g. using code instrumentation) that there really is a path from something strongly identified as a FabricValue (e.g. actual literal FabricValue, content of a cell, result of a data-model call, etc) to the marked sites." followed by a request to remove the TODOs that couldn't be verified in this way.

PR #5465, opened 2026-08-06
perf(topics): bound what a board derivation reads by declaring it

seefeldb:
## What was wrong

A single derived value on the Topics board read the whole space on startup — every topic's thread, verbs, and rendered UI — and landed **session-scoped** while doing it.

Two mechanisms compounded:

1. **The transformer shrinks a derivation's input schema to the paths it can see the body reach, and gives up when it cannot.** Measured with `--show-transformed`:

   ```
   computed(() => topics.get().length)             ->  items: { type: "unknown" }   // no element read
   computed(() => asArray(topics.get()).length)    ->  items: true                  // every element, whole
   ```

   `asArray` is a one-line helper in a sibling module. Four board derivations counted the board through it — `topicCount`, `hasNoTopics`, `[NAME]`, and the crossref join — so counting the board read all of it. Through a link to a list of pieces, "whole" means each piece's body, thread, verbs, and `[UI]` VDOM, transitively.

2. **Scope is discovered from what a computation reads.** Every read ratchets the transaction's `narrowestReadScope` ([extended-storage-transaction.ts:836](https://github.com/commontoolsinc/labs/blob/main/packages/runner/src/storage/extended-storage-transaction.ts#L836), spec at [scopes.md:19-35](https://github.com/commontoolsinc/labs/blob/main/docs/specs/server-side-execution/scopes.md#L19)), and the derived cell is minted at the narrowest scope read. A topic's `[UI]` contains VDOM fragments produced by `PerSession` ternary branches — confirmed in the pinned vintage store, which holds session-scoped docs whose value is literally `{"name":"cf-button","children":["Edit"]}`. Reaching one re-scopes the whole derived array to session, so it is written afresh per session at startup.

The session scope and the huge read are the same event: the value is session-scoped *because* the read wandered into `[UI]`.

## What changed

- **`asArray` deleted** (0 occurrences). The `Default<[]>` declarations were always what made the array guard unnecessary.
- **Derivations that call helpers the analysis cannot see through** (`topicCorpus`, `crossrefJoin`) moved to module-scope `lift`s, where the declared parameter type is the read bound. A declared type is a ceiling an opaque helper cannot widen.
- **Cards iterate `.map((row) => …)` instead of indexing `rows[i]`.** Dynamic indexing is as opaque to the analysis as a helper call — my first cut made every card site declare the whole row array (~5.9k × 9).
- **`crossrefs` and `index` are the same rows** published under two declared types, instead of two derivations mapping over a third.
- **`computed()`: 33 -> 3.** The three that remain are the shapes the transformer currently rejects at pattern scope (bare `.get()`, and `.get()` feeding a callback-taking method) — see the linked task below; they can be unwrapped once that lands.
- **One `#profile` wish** instead of three. Measured cost is a wash (3,958 -> 3,761 chars across the four read sites).

## Effect

Emitted schema across both modules: **288,628 -> 160,699 chars (-44%)** (-36% when this PR opened; the review-driven narrowing below took another 13%).

The board's seven `topics` / `crossrefView` readers collapse from 12k-28k chars each to two of ~500; the corpus read is now concentrated in one declared lift. `topicCount` emits `items: { type: "unknown" }` — counting the board expands no topic.

## The published contract is unchanged — and the read is still narrow

Narrowing the lift parameters looks like it forces a contract break: a lift's parameter and result are one type, so a narrowed parameter narrows the published `crossrefs` edge targets with it, and `deno task pattern-compat` rejects that (`result.crossrefs[].referencedBy[].addComment: existing result field was removed`). A pattern that cannot be applied over its baseline pins every piece tracking it, including the live board this exists for.

It is not a real constraint. A reference passed through a lift is a link, and a link resolves to the whole topic however little of it the lift declared. So both crossref lifts read `TopicScan` and assert each republished reference back to `TopicPiece`/`TopicReference` — a single `as` along the inheritance chain, not the `as unknown as` the transformer rejects. Both carry a `HACK:` note: the runtime already guarantees this, and a generic lift carrying the input reference type through to the output would say it without a cast.

The board's ordering and card rendering are declared over `TopicCard`/`TopicCardSource`, which carry no reference at all — the cards navigate through the durable fid/title snapshots. Nothing published is shaped by those types.

`pattern-compat` reports **no contract change at all** — the narrowing is invisible to consumers.

## Verification

- `cf test packages/patterns/topics/` — 53 passed, 0 failed, no test changes needed
- `deno task pattern-compat` — clean; new Tier-1 baselines recorded for both patterns
- `deno task pattern-vintage` — replays clean, "20 updated cleanly with no state stranded"; new Tier-2 vintage pinned. The three `!` replay notes were verified **identical on an unmodified tree**, so they are pre-existing
- `deno fmt --check` and `deno lint` clean repo-wide; `check-baselines-append-only`, `check-skill-facts`, `check-no-waitfor` pass
- `deno task integration patterns topics` — 1 passed. Cold-load browser test: both cards render, the `Open` cell link navigates, the crossref chip lands on the exact pasted fid, and the console carries zero `scheduler Event dropped: no handler registered` — the guard that the chips stay durable data rather than persisted click streams

## Follow-ups filed

- An over-broad `.get()` diagnostic in `pattern-context-validation.ts:262` that rejects several pattern-scope forms its own comment says should lower.
- Reactivity docs never explain what bounds a read, and `computed.md` steers authors away from `lift()` for exactly the case where it is the only fix.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5465?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/patterns/topics/topic.tsx
```diff
@@ -425,14 +472,14 @@ export const topicCellLink = (fid: string, label: string) =>
  * its existing piece-valued contract. */
 export const crossrefLinkRow = (
   caption: string,
-  links: TopicNavigationLink[],
+  links: TopicNavigationLink[] | undefined,
 ) =>
-  links.length === 0
+  (links ?? []).length === 0
```
seefeldb, line 477: remove `?? []`

NEVER ADD THAT

fix the type, so it has a default, then you can also remove the `| undefined` above.

packages/patterns/topics/topic.tsx
```diff
@@ -536,6 +583,221 @@ export const submitProfileLink = handler<void, {
...
+ * survives that removal.
+ */
+const findSelfIndex = lift((
+  { siblings, title }: {
+    siblings: ReadonlyCell<unknown[] | Default<[]>>;
+    title: ComparableCell<unknown>;
+  },
+): number => {
+  const list = siblings.get();
+  const length = Array.isArray(list) ? list.length : 0;
```
seefeldb, line 624: unnecessary. don't make siblings a cell, and the default makes it so that it's always an array

packages/patterns/topics/topic.tsx
```diff
@@ -536,6 +583,221 @@ export const submitProfileLink = handler<void, {
...
+  for (let index = 0; index < length; index++) {
+    if (equals(siblings.key(index).key("title"), title)) return index;
```
seefeldb, line 626: use .equals to compare, pass in self and not title

packages/patterns/topics/topic.tsx
```diff
@@ -536,6 +583,221 @@ export const submitProfileLink = handler<void, {
...
+const outboundEdges = lift((
+  { siblings, body, comments, links }: {
+    siblings: Writable<TopicMention[] | Default<[]>>;
+    body: string | Default<"">;
+    comments: { body: string }[] | Default<[]>;
+    links: { url: string }[] | Default<[]>;
+  },
+): { refsOut: TopicReference[]; refsOutLinks: TopicNavigationLink[] } => {
+  const sibs = siblings.sample();
+  if (!Array.isArray(sibs) || sibs.length === 0) {
```
seefeldb, line 657: add default then you don't need the Array.isArray

packages/patterns/topics/topic.tsx
```diff
@@ -536,6 +583,221 @@ export const submitProfileLink = handler<void, {
...
+ *
+ * Reactive on this topic's own prose and nothing else. The sibling list is
+ * SAMPLED, not read, which is the whole point: prose written before a topic
+ * existed cannot mention it, so an add re-running this for every topic already
+ * on the board was pure waste — N full corpus scans per add, producing a result
+ * that provably could not have changed. Sampling registers no dependency, so
+ * this now runs when this topic's body, comments, or links change, and then
+ * only for this topic.
+ *
+ * The cost of sampling: a sibling renamed after this ran keeps its old title in
```
seefeldb, line 642: that's why we shouldn't match by title, just be entity id

packages/patterns/topics/topic.tsx
```diff
@@ -536,6 +583,221 @@ export const submitProfileLink = handler<void, {
...
+ * the snapshot, and a sibling removed from the board keeps its chip, until this
+ * topic's own prose changes. Both are display staleness on a derived, never
+ * persisted view.
+ */
+const outboundEdges = lift((
+  { siblings, body, comments, links }: {
+    siblings: Writable<TopicMention[] | Default<[]>>;
```
seefeldb, line 649: exercise:
- recall that the key insight for this PR is to not reuse broad types but instead to spell out the specific fields a lift needs
- then look at what TopicMention contains and whether that's the minimal set. (it is giant, and pulls in everything again, creating the cascade)

packages/patterns/topics/topic.tsx
```diff
@@ -536,6 +583,226 @@ export const submitProfileLink = handler<void, {
...
+    me: number;
+  },
+): {
+  referencedBy: TopicReference[];
+  referencedByLinks: TopicNavigationLink[];
+} => {
+  const none = { referencedBy: [], referencedByLinks: [] };
+  if (me < 0) return none;
+
+  const row = boardCrossrefs.key(me).get();
```
mathpirate, line 733: `me` is this topic's index in `mentionable` — the full list — but `crossrefRows` emits one row per **non-null** entry (main.tsx:197 skips them), so the two indexings disagree the moment the board list carries a null, which is the mid-sync state the rest of this file guards for.

Concretely, with `topics = [null, A, B]`: rows are `[rowA, rowB]`, so A (`me`=1) reads B's row here — B's `referencedBy` lands on A's Connections card *and on A's published `crossrefs`* — while B (`me`=2) reads `undefined` and silently takes the full-corpus fallback. The board result's own doc says "consumers never need to correlate by index — indices are not a stable address"; this read correlates by index. It's also not a bug the old code had: the pre-split join indexed `joined.refsOut[me]`, which was aligned over the full list.

Two fix shapes that keep the design:

1. Emit a placeholder row for null entries too — the `["unresolved-topic-row", i]` cause exists for exactly this kind of entry — so `boardCrossrefs` actually holds the "one row per topic in the same order as `mentionable`" its input doc promises.
2. Drop the index coupling: locate the row by identity, walking `boardCrossrefs` with the same zero-read cell-identity comparison `findSelfIndex` already does (`equals(boardCrossrefs.key(i).key("topic"), siblings.key(me))`).

I'd lean (2) — it removes the correlation outright, which is where this file is already headed. Holding my approval on this one; everything else of mine is questions and nits.

packages/patterns/topics/topic.tsx
```diff
@@ -189,11 +222,30 @@ export interface TopicPiece extends TopicReference {
...
+  referencedBy: TopicReference[];
+}
+
+/** The same row on the topic's own page, which additionally carries the two
+ * edge sets as durable fid/title snapshots so rendered navigation survives a
+ * cold load without resolving a sibling. The snapshots stay off `TopicEdges`
+ * because a sibling never renders them, and widening the projection every
+ * reader of a topics list resolves is what a narrow declared schema exists to
+ * prevent. */
+export interface TopicCrossrefs extends TopicEdges {
```
mathpirate, line 244: The optional snapshots force three coalescing clusters downstream (`crossrefLinkRow` 475–483, `hasConnections` 1004–1005, `inboundEdges` 736–737), and the last is suspicious on its own terms: `row.referencedBy ?? []` coalesces a field `TopicEdges` declares **required** — if a row read through `boardCrossrefs` can genuinely surface without `referencedBy`, the declared type is wrong; if it can't, the `?? []` is dead.

The `hasConnections` comment says the snapshots stay optional because required fields "would change the deployed `crossrefs` default". Does pattern-compat actually reject adding `Default<[]>` fields to the published row? If not, `Default<[]>` on both snapshots fixes the type and deletes all three clusters (and the `| undefined` on `crossrefLinkRow`). If it does, let's say so in the comment, so the optionality reads as a compat constraint rather than a choice.

packages/patterns/topics/topic.tsx
```diff
@@ -536,6 +583,226 @@ export const submitProfileLink = handler<void, {
...
+const outboundEdges = lift((
+  { siblings, body, comments, links, me }: {
+    siblings: Writable<TopicMention[] | Default<[]>>;
+    body: string | Default<"">;
+    comments: { body: string }[] | Default<[]>;
+    links: { url: string }[] | Default<[]>;
+    me: number;
+  },
+): { refsOut: TopicReference[]; refsOutLinks: TopicNavigationLink[] } => {
+  const sibs = siblings.sample();
```
mathpirate, line 656: Does `.sample()` shape through the declared schema the way `.get()` does? The parameter declares `Default<[]>`, so if sampling honors it, this `Array.isArray` (and the one in `findSelfIndex` at 624) is dead by type. If sampling bypasses shaping, that's worth a line here — and in the reactivity-docs follow-up, since "sampled, not read" would then differ from a read in shape as well as in dependency.

packages/patterns/topics/topic.tsx
```diff
@@ -536,6 +583,226 @@ export const submitProfileLink = handler<void, {
...
+
+  const row = boardCrossrefs.key(me).get();
+  if (row) {
+    return {
+      referencedBy: row.referencedBy ?? [],
+      referencedByLinks: row.referencedByLinks ?? [],
+    };
+  }
+
+  // No board table: a topic wired before that input existed, which is every
```
mathpirate, line 741: This branch's retirement depends on backfilling `boardCrossrefs` as a link-bind (as `mentionable` was), but that backfill isn't among the PR's filed follow-ups — where do we want it tracked? Until it runs, every already-deployed topic keeps the full-corpus inbound scan: the one-row read applies only to topics created after this lands, so the detail-page win on the live board is gated on it.

packages/patterns/topics/README.md
```diff
@@ -58,7 +58,35 @@ lineage: Linear CT-1878, which this pattern exists to absorb).
...
+  declared parameter type is a ceiling an opaque helper cannot widen, and the
+  bodies here call `topicCorpus` and `crossrefJoin`.
+
+  A lift's parameter and its result are one type in TypeScript, which looks like
+  it forces a choice: narrow the parameter to bound the read, and the published
+  `crossrefs` edge targets narrow with it, removing result fields consumers were
+  promised (`deno task pattern-compat` rejects exactly that). It is not a real
+  choice. A reference that passes through a lift is a link, and a link resolves
+  to the whole topic however little of it the lift declared — so `crossrefRows`
+  and `topicCrossrefView` read `TopicScan` and assert each republished reference
```
mathpirate, line 78: This bullet drifted from the final shape: `topicCrossrefView` no longer exists (it split into `findSelfIndex` / `outboundEdges` / `inboundEdges` / `joinEdges`), only `crossrefRows` reads `TopicScan` (`outboundEdges` reads `TopicMention`), and `TopicCardSource` (line 84 — also in the PR description) was never landed; the card type is just `TopicCard`. Since this section is the design commitment the rest of the file leans on, worth a re-match pass.

packages/patterns/topics/topic.tsx
```diff
@@ -176,9 +209,9 @@ export interface TopicReference {
  * This is the board-facing projection, and the one stored in the tracker's
  * list. Session-local UI controls are intentionally excluded: a TopicPiece can
  * be followed from a shared list even when the viewer has no matching
- * session-local cells. Crossref targets deliberately use the non-recursive
- * TopicReference contract: a Topic needs enough sibling data to render and
- * navigate chips, not each sibling's entire crossref graph.
+ * session-local cells. Crossref targets use the narrower `TopicScan` contract,
```
mathpirate, line 212: "Crossref targets use the narrower `TopicScan` contract" — they don't: `crossrefs` is declared over `TopicEdges`, whose edge sets are `TopicReference[]`. This sentence describes the narrowing pattern-compat rejected. What's true: the *read* bound comes from the lift parameters; the published targets stay `TopicReference`.

mathpirate (commented):
Read this closely, transformer side in hand — the design is right and I want it in: declaring the bound at module-scope lifts is the correct answer to the shrink analysis giving up at helper calls, and per-row cause-keyed cells are the correct answer to positional card identity.

I verified the mechanism claims directly on this head: `topics.get().length` emits `items: { type: "unknown" }` (`hasNoTopics` narrows further, to `rows: { length: number }`), the base build's `asArray` variant reads elements whole, and compat + vintage both ran green in CI on this SHA.

Two calibration notes on the description:

- The schema-size claim doesn't reproduce for me as stated. Measuring every emitted `JSONSchema` literal in `cf check main.tsx --show-transformed` output (which compiles both modules): base ≈ 288.8k chars — matching your 288,628, so we're extracting the same thing — but head lands at ≈ 215.3k, −25%, not 160,699 / −44%. Transformer drift since Aug 10, or a different extractor? (Direction isn't in question — whole-reference expansions in the board module drop 40 → 24.)
- "The three that remain" is two now (`commentsView`, `linksView`) — both exactly the shapes #5454 admits, so they unwrap once that lands.

One blocking item: the `inboundEdges` row-alignment issue (inline at topic.tsx:733). Everything else is questions and doc nits.

packages/patterns/topics/topic.tsx
```diff
@@ -536,6 +583,226 @@ export const submitProfileLink = handler<void, {
...
+  for (let index = 0; index < length; index++) {
+    if (equals(siblings.key(index).key("title"), title)) return index;
+  }
+  return -1;
+});
+
+/**
+ * This topic's OUTBOUND edges: the siblings its own prose mentions.
+ *
+ * Reactive on this topic's own prose and nothing else. The sibling list is
```
mathpirate, line 634: "Reactive on this topic's own prose and nothing else" — `me` is an input now, so this also re-runs when the topic's index shifts (removal/prepend upstream). Cheap, since `findSelfIndex`'s output cell only rewrites on value change, but the sentence overclaims; worth a clause.

packages/patterns/topics/topic.tsx
```diff
@@ -536,6 +583,226 @@ export const submitProfileLink = handler<void, {
...
 
+// ===== Derivations =====
+//
+// Each of these is a `lift` rather than a pattern-body derivation for one
+// reason: the declared parameter type is what bounds the read. Their bodies
+// call helpers (`topicCorpus`, `crossrefJoin`) that schema analysis cannot see
+// through, and an inferred input schema falls back to reading its input whole —
+// for anything reached through `mentionable`, that is the entire board.
+
+/** This topic's own place in the board's prose graph, derived read-side from
```
mathpirate, line 594: Dangling JSDoc: this block belonged to `topicCrossrefView` and now floats above `findSelfIndex`, which has its own doc. Same at 693 — `inboundEdges`' doc sits above `joinEdges`, leaving `inboundEdges` undocumented. Tooling will attach both to the wrong symbol.

packages/patterns/topics/main.tsx
```diff
@@ -136,6 +147,119 @@ export interface TopicIndexRow {
...
+  // stops ordering the board from carrying a reference-bearing row schema:
+  // `topic` is a two-field projection, so a card reads prose and scalars and
+  // can never expand a sibling, while `TopicCrossref`'s reference fields stay
+  // available to consumers of the published result.
+  fid: string | Default<"">;
+  title: string | Default<"">;
+  topic: {
+    body: string | Default<"">;
+    createdByName: string | Default<"">;
+  };
```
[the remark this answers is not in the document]
mathpirate: cubic is right about the schema, wrong about the urgency — and the comment above is wrong about the invariant:

- `topic` **is** emitted required: schema-generator marks a property required unless it carries `?` or a `Default` in its own type node; defaults on the nested fields don't propagate up. So "Every field carries a default" is not currently true of this interface.
- Blast radius is small: no deployed version persists card rows (the row cells are new in this PR), and `crossrefRows` rewrites every row it emits, so the only rows that could lack `topic` were written by intermediate builds of this branch.
- The sharper gap: the pinned vintage (2026-08-06T23:04) predates the row cells, so `pattern-vintage` — the safety net this comment cites — currently exercises **zero** rows. It can't catch this class until a vintage containing rows is pinned.

Suggest: take the default (cards never write through `topic`, so no aliasing exposure), keep the comment true, and pin a vintage from this build so the mapped-card update path is actually under test going forward.

packages/patterns/topics/main.tsx
```diff
@@ -136,6 +147,139 @@ export interface TopicIndexRow {
...
+      // link, so a link resolves to the row's own entity and stays the same
+      // wherever the row sits, while an inline value resolves to the array
+      // position and makes identity positional. The board sorts by activity, so
+      // every append is a prepend — inline rows re-key every card and rebuild
+      // its whole subtree; linked rows keep theirs and the run is reused.
+      //
+      // A topic whose fid has not resolved yet (mid-sync) has no stable cause,
+      // so it falls back to its position — positional identity for exactly the
+      // rows that have no identity yet, which is what they have today anyway.
+      const cause = payloads[i] ? payloads[i] : ["unresolved-topic-row", i];
```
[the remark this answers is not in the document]
mathpirate: The collapse is real but only reachable by an external writer pushing the same piece reference twice — `addTopic`/`submitProfileTopic` always mint fresh pieces — and the suggested fix costs far more than the bug: putting `i` into **every** row's cause makes row identity positional again, so any removal or reorder re-keys every downstream card subtree. That's precisely what this commit exists to prevent, and we've been burned by index-correlated crossrefs before.

If we want twin-safety: either dedupe same-fid entries when building rows (one row per unique fid is arguably the honest graph semantics; costs one word in the "one row per (non-null) entry" doc), or suffix the cause with the occurrence ordinal *among same-fid entries only*, which stays stable under reorder of distinct topics. Either way I'd resolve this thread as won't-fix-as-suggested.

seefeldb:
Rebased onto current main. Triaged your ten threads against the stacked PRs — most of the code they name no longer exists by the top of the stack, so I've marked those rather than editing a shape that is about to be deleted.

**The blocking one is resolved, by the fix you preferred.**

`topic.tsx:733` (the `me`-index / `crossrefRows` misalignment): you offered two shapes and leaned on (2), "locate the row by identity". That is what the stack landed. A topic now finds its own row with a linear scan over the pivot table comparing cell identity — no index correlation anywhere:

```ts
table.filter((row) => equals(self, row.topic)).flatMap((row) => row.mentionedBy)
```

The null-entry skew you described cannot arise: nothing indexes into a sibling list, and rows are addressed by `Writable.for(topic)` — the topic cell itself is the row's cause. `crossrefRows`, `findSelfIndex`, `joinEdges`, `inboundEdges` and `outboundEdges` are all gone.

**Obsoleted by the stack** — every symbol these name is absent at the tip, verified by grep:

- `README.md:78` — the crossref section is rewritten; `topicCrossrefView`/`TopicCardSource` are gone.
- `topic.tsx:212` — `TopicScan`/`TopicEdges` no longer exist.
- `topic.tsx:244` — `crossrefLinkRow`, `hasConnections` and `inboundEdges` are gone; `referencedBy ?? []` survives only in a test.
- `topic.tsx:594`/`693` — the symbols the dangling JSDoc floated above are gone.
- `topic.tsx:634` — `findSelfIndex` is gone.
- `topic.tsx:656` — no `.sample()` call remains in the pattern.
- `main.tsx:258` — `TopicCard` is replaced by a card view whose fields all carry defaults, and the card rows it described no longer exist, so the vintage-coverage gap you identified is moot rather than fixed.

**Still open, and I'd like your read:**

- `main.tsx:210` (twin-fid cause collapse) — agreed it is won't-fix-as-suggested; positional causes are exactly what this work exists to avoid. `boardRows` and its fid-derived cause are deleted, but the shape survives one level up: the pivot causes rows with `Writable.for(topic)`, so two identical piece references would still collide on one row. Same trade-off, and I have not changed it.
- `topic.tsx:741` — `boardCrossrefs` still exists at the tip, so the backfill question stands. The consequence changed: an unbacked topic now reads an empty table and shows no backlinks, rather than falling back to a full-corpus scan. Where do you want the backfill tracked?

Measured on the rebased stack, against this same branch's base: an off-screen add on a 32-topic board goes from 3134ms to 1525ms.

mathpirate (approved):

seefeldb:
Superseded by #5921, which lands this work as one PR against `main`.

The stack's lower branches carried shapes the upper ones deleted, so review and CI kept landing on code that was already gone by the top — most recently a `crossrefRows` idempotency failure on a lift that #5921 removes outright. Same 28 commits, one base.

Comments here that named code surviving to the tip are carried forward; the rest are moot by deletion.

PR #5468, opened 2026-08-06
feat(runner): give a receipt a durable schema describing what it holds

mpsalisbury:
## Why

A receipt is read through a schema like any other cell, but receipts carry none.
Two things follow: a projection can't be pushed into the fetch, because
narrowing needs the root container kind before it can build a selector; and a
reader's schema matches a runtime value rather than a declaration. Giving a
receipt a durable schema removes both.

This is per-invocation metadata describing what one receipt holds. It is *not*
the `Pattern.resultSchema` emission withdrawn in July: the compatibility gate
(`assertPatternSchemasBackwardCompatible`) only ever compares `argumentSchema`
and `resultSchema` between two versions of a pattern, so it never reaches a
receipt and no permanence obligation attaches.

Part of the shaped-reads plan (#5435), which builds on the read model in #5316.

## What Changed

`receiptShapeSchema` derives a one-level shape from the settled value and writes
it via `setMetaRaw("schema", …)` in the same create-only transaction as the
result. Container kind only — `{"type": "object", "properties": {k: true}}` —
since that is all the narrowing needs. Cell links yield no schema.

## Known limitation

The write sits on the receipt-only branch (`runner.ts`, guarded by
`!resultHasReactives && frame.reactives.size === 0`). A verb returning anything
reactive — including a child piece — takes the branch where
`updateResultSchemaMeta` is called with the synthesized pattern's
`resultSchema`, which is `undefined` because `Stream<E, R>` drops `R`. Those
receipts still get no schema. Closing that needs a declared result schema at the
source; tracked in the plan's backlog.

## Test plan

`deno task test` in packages/runner.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5468?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/scheduler-v2/README.md
+++ b/docs/specs/scheduler-v2/README.md
@@ -797,6 +797,20 @@ persists as-is, a live cell handle converts to a link, so a one-line setter
 verb's receipt links to the cell it mutated. The create-only witness
 semantics are unchanged either way.
 
+A receipt-only handling also describes what it wrote, in the same transaction,
+on the cell's durable `schema` metadata — so a receipt says what it holds the
+way any other cell does, which is what lets a reader's selection narrow the
+fetch instead of filtering an already-loaded value. The description is
+structural and nothing more: the root container kind, plus the property names
+when that kind is a record. Every property is left admissible, which is what
+keeps a link position honest, since the spelling that would name one is
+`asCell` and `["cell"]` asserts a writable handle on a document nothing can be
+written through. A value with no container kind of its own — a scalar, or a
+link, whose kind is its target's — goes undeclared. None of this constrains a
+later write: the create-only mark means the value the schema describes is the
+only value that document ever holds. A verb's *declared* result type is a
+separate question, settled at the type layer and never reaching the runtime.
+
 For an inline, non-navigation handler result, an `inSpace` child does not move
 that canonical handler-result wrapper into the child space. The result/receipt
 stays in the handler's originating space, while each child node materializes its
--- a/docs/specs/scheduler-v2/README.md
+++ b/docs/specs/scheduler-v2/README.md
@@ -806,7 +806,10 @@ when that kind is a record. Every property is left admissible, which is what
 keeps a link position honest, since the spelling that would name one is
 `asCell` and `["cell"]` asserts a writable handle on a document nothing can be
 written through. A value with no container kind of its own — a scalar, or a
-link, whose kind is its target's — goes undeclared. None of this constrains a
+link, whose kind is its target's — goes undeclared. A `data:` link is the one
+link that is described: it carries its value inside its own identifier and the
+write inlines it, so the receipt holds that value rather than a link to it, and
+the description is taken after the same inlining. None of this constrains a
 later write: the create-only mark means the value the schema describes is the
 only value that document ever holds. A verb's *declared* result type is a
 separate question, settled at the type layer and never reaching the runtime.
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -41,6 +41,7 @@ import {
   pushFrameFromCause,
 } from "./builder/pattern.ts";
 import { type Cell, createCell, isCell } from "./cell.ts";
+import { findAndInlineDataUriLinks } from "./data-uri.ts";
 import { type Action } from "./scheduler.ts";
 import {
   isSchedulerActionObservation,
@@ -278,6 +279,11 @@ function patternDefaultScope(pattern: Pattern): CellScope | undefined {
  * or a link, whose kind belongs to whatever it resolves to rather than to the
  * receipt.
  *
+ * A `data:` link is the exception among links: it carries its value inside
+ * its own identifier, and the write inlines it, so the receipt holds that
+ * value rather than a link to it. The description comes from the same
+ * inlining, so it describes what is stored.
+ *
  * This is DESCRIPTIVE — what this one receipt holds — and never a contract
  * bearing on anything written later. Description and authority cannot diverge
  * here, because the receipt's create-only mark means the value it describes is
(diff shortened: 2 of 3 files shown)


mpsalisbury (commented):

# cf-review — PR #5468 "feat(runner): give a receipt a durable schema"

**Scope:** 3 files, +346/−2 · **Coverage:** deep-read of the diff + receipt branch, meta plumbing (`cell.ts`), CFC link-label derivation (`cfc/prepare.ts`); sampled `piece-helpers`/`piece-controller`/`llm-dialog` schema-meta readers · **Checks run:** new test file at head (7/7 pass), same file against `origin/main` (isolates the delta), `deno fmt --check` + `deno lint` on touched files.

## 🔴 Blocking

None confirmed.

## 🟡 Improvements

**1. The receipt's `schema` meta silently enters CFC link-label derivation as an "author-declared shape" — unexamined by the PR. [suspected — mechanism verified by reading; net effect needs a CFC-owner call]**
- Location: `packages/runner/src/cfc/prepare.ts` — `setupResultSchemaFor` (~line 4319 at head) reads the source doc's `["schema"]` field with read-your-writes at commit-time prepare; its doc comment trusts that field as "the same author-declared shape a pending schema input carries". The persist loop iterates every doc with a link write (`targetKeys = new Set([...candidates.keys(), ...linkWrites.keys()])`, ~line 5402), so an unlabeled receipt doc holding a link (verb returns `{note: someCell}` under `plainResultReceipts`) reaches `derivePersistedLinkLabel`.
- What changes: before this PR that case took the fail-closed branch — `{reason: "missing link source metadata …"}`, no link-origin label persisted. Now `pendingSourceSchema` is defined, both fail-closed branches are skipped, and an `origin:"link"` entry (LinkReference integrity atom, empty confidentiality) is persisted into the receipt's `["cfc"]` metadata instead. Under write-floor enforcement the pre-PR contribution was `[]` ("fails the floor, fail-closed" per the comment at ~5199); now it contributes `[LinkReference]`.
- Scoped by experiment: commit outcomes do NOT change — I ran the PR's test file against `main` and every step's dispatch commits `done` there too; only the schema-meta expectations differ. No label values can be fabricated either: the shape schema carries no `ifc` entries, so `walkIfcSchema` yields nothing and `persistedLabelFromSchemaAtPath` returns `undefined`. The delta is confined to persisted CFC metadata and the disappearance of a fail-closed reason.
- Why it matters: a value-derived shape now flows through a seam whose comment promises author-declared provenance, in a security subsystem, and neither the PR body, the spec paragraph, nor the tests mention CFC. If the reason-vs-empty-label distinction is load-bearing under stricter enforcement modes, this becomes a real weakening.
- Fix: get an explicit ack from the CFC owner (and ideally a one-line update to `setupResultSchemaFor`'s comment acknowledging receipt shapes as a schema-meta source); a small CFC test pinning the intended label outcome for a link inside a receipt would close it.

**2. The reactive-branch gap is prose-only — nothing pins it. [verified — read the test file]**
- Location: `packages/runner/test/receipt-schema.test.ts`.
- The PR's "Known limitation" (reactive-returning verbs get no schema because the synthesized pattern's `resultSchema` is `undefined`) has no test. If a later change makes that branch start writing schemas — e.g. a wrong one derived from the pattern envelope — nothing fails, and that is exactly the misdescription class PR 5505's narrowing would then trust.
- Fix: one test — verb returning a reactive/child piece, assert the result cell carries no `schema` meta (or carries only what `updateResultSchemaMeta` wrote), so the boundary is pinned rather than described.

## ⚪ Nits

- The schema write is gated on `commitPreconditions` only, not `plainResultReceipts` — with plain receipts explicitly disabled, every `{}` witness still gets `{type:"object",properties:{}}` meta, a small durable per-invocation cost with nothing to narrow. Intentional per the value-less test; flagging so it's a decision, not an accident.

## Verified clean (the focus areas)

- **`receiptShapeSchema` value domain**: no misdescribing case found. `normalizeSandboxResult` canonicalizes host natives (Date/RegExp/Uint8Array/Error) to non-plain Fabric instances → undeclared; `isCellLink` (link-utils.ts:61) covers live Cells, sigil links, AND query-result proxies → undeclared; `null`/scalars → undeclared; arrays get `{type:"array"}` with no `items` (permissive, correct); null-prototype objects pass `isPlainObject` but also pass the sandbox's `isPlainResultObject`, so schema and stored value agree. Create-only means the exact-keys property list can never drift from the value.
- **Transaction semantics**: `setMetaRaw` writes `["schema"]` at the root of the same fresh document in the same tx (cell.ts:2551); the replay test pins that a same-id loser loses value AND metadata atomically (`receipt-exists`), and I ran it. No new commit-failure mode — empirically identical commit outcomes on main vs head.
- **Sanctioned surface**: `"schema"` is a declared `MetaField` (packages/api/index.ts:554, "stores the schema for a result cell" — receipts are result cells), the same surface `updateResultSchemaMeta` uses. Both existing readers (`getResultCellWithSourceSchema`, piece-controller's `durableSchema`) expect a `JSONSchema` there and only touch piece result cells, which receipts are not; even if reached via llm-dialog, the all-`true` properties schema cannot trim or reject anything.
- **No-permanence argument holds**: `assertPatternSchemasBackwardCompatible(previous: Pattern, candidate: Pattern)` (packages/piece/src/schema-compatibility.ts:168) validates and compares only `argumentSchema`/`resultSchema` of Pattern objects; both call sites (piece-controller.ts:3364, 3642) pass Patterns. It never reads cell meta.
- **Test rigor**: each test names its principle; waits are event-driven (commit callback + `idleWithPendingCommits`), no polling; style follows the repo's unit-test conventions; the spec paragraph in `docs/specs/scheduler-v2/README.md` matches the implementation.

## Possible follow-ups

- CFC-side test + comment update for schema-meta provenance (finding 1).
- Reactive-branch no-schema pin (finding 2).
- The declared-result-schema work for reactive verbs is already tracked in the plan backlog per the PR body.

**Verdict**: sound and carefully built — mergeable once the CFC label-derivation interaction (finding 1) gets an explicit owner ack; the rest is test hardening.

PR #5469, opened 2026-08-06
feat(runner): scope a caller's invocation id by the session that chose it

mpsalisbury:
## Why

`cf piece call --invocation <id>` lets a caller name a dispatch so a retry
returns the original outcome instead of executing again. That id alone decided
where the receipt landed: `scopeCallerEventId` hashed `{caller, id, path,
space}`, where `caller` is the id *string*, not a principal.

So no identity entered the address. Two callers picking the same id against the
same verb in the same space computed one address and read one receipt, with
nothing in the response saying so — the second would be told its call settled
when it never ran. And the id is guessable by construction: the documented
convention teaches human-friendly ids like `add-comment-1`, which two agents
following it derive systematically. Piece, verb and id were all a third party
needed to compute where someone else's outcome lives.

Joining the session to the hash fixes both. The session says *whose* invocation
this is, and it is the one unguessable component.

Part of the shaped-reads plan (#5435).

## What Changed

- `scopeCallerEventId` takes the session as a **required** parameter and hashes
  it alongside the caller's id and the stream link. Required rather than
  optional so the compiler finds every call site instead of leaving a
  default-to-unscoped path. It stays a structured content hash, so the
  `(x, y:z)` vs `(x:y, z)` ambiguity does not reappear now that two
  caller-controlled halves are in the input.
- `resolveInvocationIdentity` replaces the two independent resolvers, because
  the contract couples the flags: neither flag mints both; `--session` alone
  mints an id within it; `--invocation` alone is now an error naming the remedy;
  both are carried through as a pair.
- The pair travels as one `InvocationIdentity`, so a call cannot carry half of
  what names it.

## Stacked on #5470

Based on `impl/a12-link-marker`, not main — both branches insert walkthrough
steps, and the renumbering is already resolved here. Merge #5470 first.

## Second commit: the announcement carries both halves

`cf piece call` announces `invocation: <id>` on stderr so a caller whose process
dies can retry and deduplicate. Once an id deduplicates only under its session,
that line is half a key — a caller minted a session for one call could read the
id and hold nothing it could retry with. The session now goes out beside it, on
its own line, because `invocation: <id> phase: <phase>` is a shape integration
scenarios parse to block on a phase.

## Test plan

`deno task test` in packages/cli and packages/runner; `deno task check`, `fmt`,
`lint`, `check-docs`, `check-skill-facts`. Integration: `piece-call-retry`,
`three-topic`, and the full walkthrough against a local toolshed. The two new
join tests were confirmed to fail with the session removed from the hash.

## Renamed since the first push

The flag, its environment variable, and the minting command are all qualified:
`--invocation-session`, `CF_INVOCATION_SESSION`, `cf invocation-session new`.

`cf inspect` already takes `--session`, meaning a *storage* session — a
different concept in the same binary. Qualifying ours is free today because
nothing uses it yet. `cf inspect`'s could become `--storage-session` later;
that's a forensics tool with its own callers and is not part of this work.

Cliffy's `prefix: "CF_"` derives the option name from the variable, so the
rename could have silently broken the environment path while the flag kept
working. It's verified three ways: a test that parses with only the variable
set, a mutation check (changing the prefix made that test fail as expected),
and `--help` output showing both spellings.

**One shape change, not just names.** The detached next-step hint now prints
`CF_INVOCATION_SESSION=<id> cf piece call …` rather than putting the session in
argv. It's the one place the CLI teaches how to carry a session, and the old
form would have taught a spelling that exposes it in `ps` — the session's
unguessability is what keeps a receipt address out of reach of anyone who can
guess a piece, a verb, and a conventional id.

**Not renamed, deliberately:** the stderr announcement is still `session: <id>`.
That line is a parsed shape — tests assert it and integration scenarios block on
it — so changing it is a caller-visible output change that wants its own call.

### DIFF
--- a/docs/common/verbs-over-the-cli.md
+++ b/docs/common/verbs-over-the-cli.md
@@ -108,12 +108,17 @@ the caller could not have computed them.
 
 `--invocation <id>` makes a call idempotent. The id is your own word for the
 call — and `add-comment-1` is a word two agents both reach for, so an id on its
-own does not say whose invocation it is. A **session** does. Mint one per agent
-run and carry it on every call of that run:
+own does not say whose invocation it is. An **invocation session** does. Mint
+one per agent run and carry it on every call of that run:
 
 ```bash
-export CF_SESSION=$(cf session new)
+export CF_INVOCATION_SESSION=$(cf invocation-session new)
 ```
+
+The environment is where a session belongs, because it is closer to a secret
+than a setting: it is what keeps an outcome's address out of reach of a
+stranger, and a command's arguments are readable in a process listing where its
+environment is not. `--invocation-session <id>` overrides it for one call.
 
 The pair is what names an invocation. Replaying a settled id **from the session
 that chose it** returns the **original** result rather than re-executing:
@@ -138,9 +143,10 @@ id is not enough to read someone else's outcome — the session is the part a
 stranger cannot guess.
 
 `--invocation` therefore requires a session, and a call naming an id without one
-is refused, pointing you at `cf session new` and `--session`. Pass neither and
-both are minted for that one call: a random id names an outcome nothing else
-will ask for, and a call that never intended to replay loses nothing.
+is refused, pointing you at `cf invocation-session new` and
+`CF_INVOCATION_SESSION`. Pass neither and both are minted for that one call: a
+random id names an outcome nothing else will ask for, and a call that never
+intended to replay loses nothing.
 
 A **rejected** call is different from a settled one: it never spends its id. If
 a verb refuses the payload, correct it and retry under the same id.
--- a/docs/common/verbs-over-the-cli.md
+++ b/docs/common/verbs-over-the-cli.md
@@ -106,21 +106,41 @@ the caller could not have computed them.
 
 ### Retries are safe, and cheap to reason about
 
-`--invocation <id>` makes a call idempotent. Replaying a settled id returns the
-**original** result rather than re-executing:
+`--invocation <id>` makes a call idempotent. The id is your own word for the
+call — and `add-comment-1` is a word two agents both reach for, so an id on its
+own does not say whose invocation it is. A **session** does. Mint one per agent
+run and carry it on every call of that run:
+
+```bash
+export CF_SESSION=$(cf session new)
+```
+
+The pair is what names an invocation. Replaying a settled id **from the session
+that chose it** returns the **original** result rather than re-executing:
 
 ```bash
 cf piece call --piece <topic> --invocation add-comment-1 \
   addComment '{"body":"first","agentName":"Sol"}'
 
-# Same id, different payload: the original result comes back, and no second
-# comment is recorded.
+# Same id, same session, different payload: the original result comes back,
+# and no second comment is recorded.
 cf piece call --piece <topic> --invocation add-comment-1 \
   addComment '{"body":"different","agentName":"Sol"}'
 ```
 
 That is the property an agent depends on when it retries a call whose response
 it never saw.
+
+That same id under a **different** session is a different invocation: it
+executes, and returns its own result. So two agents that pick the same word are
+never told each other's calls have settled, and knowing a piece, a verb and an
+id is not enough to read someone else's outcome — the session is the part a
+stranger cannot guess.
+
+`--invocation` therefore requires a session, and a call naming an id without one
+is refused, pointing you at `cf session new` and `--session`. Pass neither and
+both are minted for that one call: a random id names an outcome nothing else
+will ask for, and a call that never intended to replay loses nothing.
 
 A **rejected** call is different from a settled one: it never spends its id. If
 a verb refuses the payload, correct it and retry under the same id.
@@ -203,7 +223,7 @@ API_URL=http://localhost:8000 packages/cli/integration/verbs-over-the-cli.sh
 ```
 
 CI runs it in the `piece-call` shard. It takes under half a minute and around
-twenty `cf` invocations against a warm local toolshed.
+two dozen `cf` invocations against a warm local toolshed.
 
 Each step demonstrates one use case:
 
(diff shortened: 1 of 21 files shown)


mpsalisbury (commented):
_cf-review via Claude Fable 5, on behalf of @mpsalisbury. Reviewed this PR's own delta vs `impl/a12-link-marker` (stacked on #5470)._

## cf-review — PR #5469 "feat(runner): scope a caller's invocation id by the session that chose it"

**Scope:** 17 files, +834/−164, `impl/s2-join-hash` vs `impl/a12-link-marker` (stacked on #5470) · **Coverage:** deep-read the full diff plus head-state of every load-bearing file (`event-identity.ts`, `cell.ts`, `piece.ts`, `callable.ts`, `session.ts`, both integration scripts, the docs); repo-wide greps at head for call sites, `--invocation` teachers, and stderr parsers · **Checks run:** static only — no suite needed to confirm any finding.

The change does what it says: the session joins the receipt-address hash as a required parameter, one resolver owns the flag matrix, and the pair travels as `InvocationIdentity` end-to-end. The security core is solid — verified specifics first, since this is the security PR of the stack:

- **Hash home — canonical, not forked (verified).** `scopeCallerEventId` (`packages/runner/src/scheduler/event-identity.ts`) hashes via `hashStringOf` from `@commonfabric/data-model/value-hash` — exactly the anti-duplication table's canonical home. The input is a structured object `{caller, id, path, session, space}`, so the `(x, y:z)`/`(x:y, z)` ambiguity cannot reappear, and the new test pins that property across the new id/session boundary too (`("inv","1:ses")` vs `("inv:1","ses")`).
- **Session quality (verified).** `newSessionId` (`packages/cli/lib/session.ts`) is `crypto.randomUUID()` — CSPRNG, 122 random bits. A test pins the v4 shape with an explicit rationale rejecting countable/timed sources. The session appears nowhere durable in the clear: hash-only in the delivery id, absent from the stdout `invocationJson`, absent from receipts. The detached next-step hint carries it as an env-prefix (`CF_INVOCATION_SESSION=… cf piece call`), which does not appear in a process listing — the design claim holds.
- **No placeholder sessions (verified).** Grepped every `eventId:`-producing site at head. Production producers are exactly two: `packages/cli/lib/callable.ts:609` (the real pair) and `packages/runner/src/cell.ts:1495` (the scoped output). Constant sessions exist only in tests, where a fixed session is the correct fixture. `cf-harness`/`cfc` `eventId`s are unrelated audit-atom fields, not stream sends.
- **Enforcement is two-layer (verified).** Compiler-required at `scopeCallerEventId`; runtime-refused at `cell.send` for any in-process caller passing `eventId` without `session`, with a test proving no delivery queues.
- **Announcement compat (verified).** The only stderr parsers at head are `packages/cli/integration/integration.sh:638` and `:884` — substring/glob matches (`*"invocation: $ID"*`, `*"… phase: committed"*`) unaffected by the added `session:` line; the phase-line shape is untouched. The kill-at-dispatch FIFO scenario is unaffected (both announce lines are one synchronous burst, and the process is killed either way).
- **Join tests pin the join (verified by reading).** Unit: same id, two sessions → different hashes. E2E: same id, two sessions → two receipts, second executes, no `receipt-exists` precondition. Removing session from the hash fails both, as the PR claims — these guard the principle, not an incidental.

### 🔴 Blocking

None found.

### 🟡 Improvements

1. **Deploy-boundary migration is unstated** — `scopeCallerEventId`'s old hash landed on main **2026-07-28** (#5087), so unsessioned receipts may already exist wherever that deployed. After this PR there is no dual-read: an in-flight retry across the boundary is first *refused* (no session), and once the caller mints one it computes a fresh address and **re-executes** — the exact double-execution the id exists to prevent. Likely acceptable ten days into an in-flight arc with prod-board gates, but the PR body should say the break is conscious and that no old-hash receipts matter. *Suspected* (depends on whether any prod caller used `--invocation` since #5087) — see Questions.
2. **The session is announced on stderr even when caller-supplied** — `invocationPhaseReporter` (`packages/cli/commands/piece.ts`) prints `session: <id>` unconditionally at dispatch. The recovery rationale (a dying caller holding half a key) only needs the *minted*-session case; for an env-carried per-run session, the standing secret is copied into every captured stderr/CI log, in tension with the "closer to a secret than a setting" framing and with `event-identity.ts`'s own "would otherwise reach logs verbatim" reasoning. Concrete fix: announce only a session minted for this call, or record the tradeoff in the JSDoc so it reads as chosen. *Suspected (design), announcement behavior verified.*
3. **Stale retry comment inside the changed file** — `piece.ts` catch block (~line 1756): "retries SAFELY ONLY with this same id (same-id retries deduplicate…)". Post-change it is the pair that retries safely; this is the file the diff rewrote for exactly this distinction. One-line reword. *Verified at head.*
4. **`docs/plans/verb-result-selection.md:198`** (live plan): "Repeat `cf piece call` with the same `--invocation` id" — now incomplete; the same session is required or the repeat is refused/re-executes. One-line fix. *Verified at head.*

### ⚪ Nits

- `cell.ts` enforces the pair one-way: `session` without `eventId` is silently dropped rather than refused. Harmless today (the CLI always sends both), but the asymmetry is undocumented. *Verified.*

### Questions for the author

- Has any production caller (prod Topics board scripts included) used `--invocation` since #5087 deployed? If yes, are its old-hash receipts and possible in-flight retries consciously written off?

### Possible follow-ups

- The PR body itself notes one: renaming `cf inspect --session` to `--storage-session` to finish de-colliding the two session concepts.
- Consider a machine-readable retry handle (id+session in the WaitBoundExpired stdout JSON, or announce-only-when-minted per finding 2) so scripts don't scrape two stderr lines.

**Verdict:** Sound and carefully built — canonical hash home, real CSPRNG session, no placeholder call sites, tests that pin the join; merge after the deploy-boundary question is answered and the two one-line doc/comment staleness fixes land.

PR #5470, opened 2026-08-06
feat(cli): `$link` marks a selection position for its address

mpsalisbury:
## Why

A read can return a value, but it could not return *where that value lives*. So
a caller who read a board and wanted to act on one of its topics had no address
to act with — they had to know how to reconstruct one, or fall back to
`--show-links` and parse it. Marking a position gives the caller an address that
is directly actionable in the next command.

The second half is what it costs to read. Selecting a collection previously
fetched every element document even when the caller only wanted addresses.
Marking a position now suppresses the fetch beneath it.

Part of the shaped-reads plan (#5435), which builds on the read model in #5316.

## What Changed

`$link` at a position renders the address stored there instead of its contents:

```json
{ "$link": { "id": "of:fid1:…", "space": "did:key:…", "scope": "space", "path": [] } }
```

All four fields always present, `id` keeps its scheme, `space`/`scope` filled in
from the containing document, `path` is `[]` at a root. `overwrite` dropped,
`schema` never inlined. A marker beside a projection merges.

`normalizeProjectionSchema` lifts markers into a separate tree and returns a
marker-free schema in which a link-only position becomes the `false` schema,
which flows through the existing path union as a new `ProjectionMask` variant.
Addresses are composed back in from the link stored in the containing document,
reading nothing the selection had not already read.

## One finding that changed the design

A rejecting selector at an array's `items` does **not** suppress the fetch.
`traverseArrayWithSchema` follows every element link before consulting the item
schema — deliberately, per the comment there: element traversal is what kicks
async loads for absent targets, and short-circuiting would serialize them until
`Cell.pull()`'s round budget exhausts. So this reduces an array whose items ask
for nothing to `false` at the array itself, which is what actually delivers the
saving. `{"property": false}` does suppress.

Measured, not assumed: a test wraps `storageManager.open(space).sync` and
asserts a marked read syncs one document where the equivalent projected read
syncs all three. Confirmed non-vacuous by swapping the projection.

## Flagged, not fixed

A JSON `--schema` needs an explicit `"type": "object"` at *every* object level;
without it `{"properties":{"label":true}}` returns `{}` today. Pre-existing and
unrelated to markers, but it means a marker's sibling contents need the type.
Fixing it changes `--schema` behavior for non-marker projections, so it wants
its own decision.

`--filter` combined with a marker is refused rather than silently wrong: a
filtered array's survivors cannot be traced back to source positions.

#5469 is stacked on this branch, with the walkthrough renumbering already
resolved there; merge this one first.

## Test plan

`deno task test` in packages/cli; walkthrough step 6 asserts the returned
address is one a caller can pass straight back to `piece call` and `piece get`,
verified against a live toolshed.

### DIFF
--- a/docs/common/verbs-over-the-cli.md
+++ b/docs/common/verbs-over-the-cli.md
@@ -280,6 +280,11 @@ branching. `id` keeps its scheme, because the scheme is the kind and dropping
 it retargets the address silently. No schema is inlined: a stored link can
 carry an entire one, and what was asked for is where the value lives.
 
+Marking a position deeper than a link names the linked document and the path
+below it — `notes[0].title`, where `notes` holds links, renders the note's own
+`id` with `path` `["title"]`. A link is a durable identity; the slot that holds
+it stops naming the same value the moment the collection is reordered.
+
 The marker sits beside a projection when both are wanted, and the answer
 carries both:
 
--- a/docs/common/verbs-over-the-cli.md
+++ b/docs/common/verbs-over-the-cli.md
@@ -200,8 +200,8 @@ Run it against any host:
 API_URL=http://localhost:8000 packages/cli/integration/verbs-over-the-cli.sh
 ```
 
-CI runs it in the `piece-call` shard. It takes about ten seconds and eleven `cf`
-invocations against a warm local toolshed.
+CI runs it in the `piece-call` shard. It takes under half a minute and around
+twenty `cf` invocations against a warm local toolshed.
 
 Each step demonstrates one use case:
 
@@ -210,12 +210,14 @@ Each step demonstrates one use case:
 | 1–2 | Deploy a pattern, then ask what it can do |
 | 3 | A create hands back the piece it created |
 | 4 | A verb returns what it wrote, including what only it could compute |
-| 5 | A replayed id returns the original result; nothing runs twice |
-| 6 | A piece result needs no option; a plain record does — and the write lands either way |
-| 7 | A value-less verb settles with the empty witness |
-| 8 | A refused call does not spend its invocation id |
-| 9 | Reading a verb redirects to `cf piece call` |
-| 10 | Timings on stderr, Invocation JSON still clean on stdout |
+| 5 | A call's result names the document behind each path, and that address calls |
+| 6 | A read returns an address in place of what is behind it |
+| 7 | A replayed id returns the original result; nothing runs twice |
+| 8 | A piece result needs no option; a plain record does — and the write lands either way |
+| 9 | A value-less verb settles with the empty witness |
+| 10 | A refused call does not spend its invocation id |
+| 11 | Reading a verb redirects to `cf piece call` |
+| 12 | Timings on stderr, Invocation JSON still clean on stdout |
 
 ## Addressing a piece you were handed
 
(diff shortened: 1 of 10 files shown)


mpsalisbury (commented):
_cf-review via Claude Fable 5, on behalf of @mpsalisbury. Reviewed this PR's own delta (`impl/f1-read-step...impl/a12-link-marker`), not the GitHub base diff, so #5458's content is reported on that PR instead._

## cf-review: PR #5470 — `$link` marks a selection position for its address

**Scope:** delta `origin/impl/f1-read-step...origin/impl/a12-link-marker` (2 commits, 7 files, +831/−58) · **Coverage:** deep-read of `packages/cli/lib/cell-selection.ts`, the new tests, `piece.ts` caller, and all four doc/skill surfaces; sampled runner link machinery (`parseLink`, `parseLinkPrimitive`, `Cell.key`, `link-types.ts`) · **Checks run:** 5 targeted probe tests executed against the PR head in a throwaway worktree (since removed); no full suites.

The design is coherent: markers lift into a parallel `LinkMarkers` tree, the marker-free schema flows through the existing mask/graph machinery, addresses compose from links already stored in documents the read visited. I verified the mechanics the PR claims: marked collections behind a crossed link render correct element addresses (probe), cross-space/scope defaults resolve against the containing document (`parseLinkPrimitive` fills `space`/`scope` from base, uses the link's own doc-root-relative path — no path-concatenation bug), `isRecord` includes arrays so element links are visible, and the `false` reduction is unreachable for non-marker callers (`normalizeProjectionSchema` still throws on explicit `false`, so no behavior change for existing readers).

### 🔴 Blocking

**1. A root marker below a link crossed by the *CLI path* renders exactly the slot address commit `9c2dd69d9` defines as the bug.** — `packages/cli/lib/cell-selection.ts` `sourcePosition()` + `packages/cli/lib/piece.ts` `getCellValue()` (`targetCell = rootCell.key(...path)`) · **verified by repro at head:**

```
deriveSelectedValue(board.key("notes").key("0").key("title"), {"$link":true})
→ {$link: {id: <board>, path: ["notes","0","title"]}}     // actual
→ {$link: {id: <note>,  path: ["title"]}}                  // per README/SKILL semantics
```

`Cell.key()` only extends the path (confirmed in `packages/runner/src/cell.ts`), and the walk starts at the source cell, so links crossed by the *typed* `cf piece get` path are invisible — only a link stored at the final node is crossed. Consequences: `cf piece get notes/0 --schema '{"$link":true}'` returns the durable note id, but `notes/0/title` returns the board id plus the path the caller just typed — the form the fix commit's own rationale calls wrong ("names a slot rather than a value… reorder the array and the same address resolves to a different note's title", "told the caller nothing they did not have"). It directly contradicts what this PR writes into `packages/cli/README.md` ("the address names the deepest stored link crossed on the way to the marked position… not the board's id with path `["notes","0","title"]`") and `skills/cf/SKILL.md` ("a position with no link above it keeps the source document's own address"). **Fix:** seed the walk from a link-resolved source — the runner's `resolveLink` walks stored links segment by segment — or, if consciously accepted, narrow the docs to say the walk starts at the read's source and path segments given to `cf piece get` are never re-based.

### 🟡 Improvements

**2. A marker beside an *array* projection silently discards the contents that were explicitly asked for — while still fetching every element.** — `composeLinkAddresses()` tail (`isRecord(composed) && !Array.isArray(composed)` → address alone) · **verified by repro:** `{"$link":true,"type":"array","items":{"properties":{"title":true}}}` returned only `{$link: …}` and synced all three note docs to compute titles it then threw away. This breaks the PR's own rationale ("the address joins them in one object, *because both were asked for*") and the fetch-suppression story at once. The `additionalProperties` case refuses loudly for exactly this reason ("dropping it answers with contents where an address was asked for") — do the same for a marker beside an array-shaped (and arguably scalar-typed) projection, or at minimum don't read what composition will discard.

**3. A marker alone at an array-rooted read refuses with a misleading error.** — `resolveProjection()` shape check (`projection.schema !== true && sourceIsArray !== projectsArrayItems`) · **verified by repro:** `deriveSelectedValue(board.key("notes"), {"$link":true})` throws `A JSON --schema for an array value must describe the returned array (for example {"type":"array","items":{...}})` — which steers the user to the spelling from finding 2 that discards data. The `false` root schema fires the array/object shape check it shouldn't be subject to; the `mask === false` fast path would answer it correctly. **Fix:** exempt `projection.schema === false` from that check (one clause), and the natural "address of this collection" question gets its one-read answer.

### ⚪ Nits

- A marker sitting *exactly on a link* renders that link only when the walk can see it — behind another crossed link the containing document is unread, so the same marked shape renders the near-side slot instead (both resolve to the same value; durability differs). Documented in the `WalkedPosition` docstring but absent from the user docs, whose "deepest stored link" phrasing implies otherwise. Verified by reading `composeLinkAddresses`'s properties branch (which deliberately does not reach through, unlike `storedContainer` for items).
- `mergeMasks`'s `right === false` branch is unreachable through the public flow (`--filter` + markers is refused before masks merge); it is directly unit-tested as a defensive union, which is fine — just know it's belt-and-suspenders, not live.

### Test rigor

The fetch-suppression test is **genuinely non-vacuous**: notes are seeded *unwritten* so reading one must hit storage; the control read proves the `provider.sync` wrapper observes element syncs; and if the wrapper observed nothing, `expect(marked).toEqual([boardUri])` would fail rather than pass vacuously. The unrestored monkey-patch is harmless — `beforeEach` builds a fresh `StorageManager` per test. The other new tests each name a real principle and read as sentences; the one I'd push back on is "returns the address alone where the contents are not an object", which asserts (and thereby blesses) the silent-discard behavior in finding 2. Docs are coherent with the code otherwise: integration-script step numbering matches the renumbered table in `docs/common/verbs-over-the-cli.md`, and the CLI example/skill rows check out.

### Possible follow-ups (for a human to file)

- Decide the marker-beside-non-object contract once (refuse vs. envelope) and encode it in `normalizeProjectionSchema` so parse time catches it.
- User-doc note on the marker-on-a-link-behind-a-link nuance (nit 1) if the no-fetch trade-off stays.

**Verdict:** strong, well-verified core mechanism with honest tests, but finding 1 leaves the PR's own documented address semantics undelivered on a spelling users will reach for first — resolve or consciously accept (and re-document) before merge; findings 2–3 are small contained fixes in the same file.

PR #5482, opened 2026-08-07
fix(runner): let a write skip the artifact walk over a query result

Hixie:
`Cell.set()` and the collection operations reach the data model through `diffAndUpdate()`, which walks the value on its way in and replaces every builder artifact with its encodable form. That walk descends into anything reporting itself as a plain object or an array.

A query result reports itself as both and is neither. It is a view onto stored data whose members resolve through the transaction it is bound to, one at a time, as they are asked for, and each one of those reads is recorded on the transaction as a dependency the commit then has to check. So writing back a list of a thousand elements that a read had handed back cost a member read of every element, and of every record under each of them, and then a commit checking all of them.

The diff underneath already treats a query result as a leaf. `normalizeAndDiff()` recognizes one and replaces it with the sigil link it names, without reading a single member. That is what makes the read-modify-write shape cheap: an unchanged element becomes a link naming the slot it already occupies, and only the element that changed is written. Forcing the view before the diff could decline to look at it took that away.

`replaceArtifacts()` gains an `isLeaf` hook, by which a caller names a value the walk must not read into. Two callers answer it, and both hand what they get to a diff rather than to a serializer: `diffAndUpdate()`, and `Runner.updateArgument()`, whose flattened argument goes only to the two writes below it. Each answers with the same predicate `normalizeAndDiff()` uses to recognize a query result.

The hook is the caller's rather than the walk's, because the boundaries that walk a value want opposite answers, and what separates them is what each does with the result. `Runtime.getImmutableCell()` derives an entity id from the bytes of the whole value, so it reads every member either way; a query result there can also hand out a `Cell`, which its `replaceOther` hook turns into a link, since reading a member holding a stream marker yields one. A caller feeding a diff reads nothing it will use.

The `flat list update - READ-MODIFY-WRITE` benchmarks in `packages/runner/test/cell-set-flat-index-list.bench.ts` are where this showed up, on all three of the scheduled Benchmarks workflow's runner processors, at up to seven times slower. It inverted the comparison the benchmark exists to make: the read-modify-write case had become slower than regenerating the whole list from scratch, when it should be several times faster. `Cell.set() - multiple transactions, one set each` in `cell-set.bench.ts` moved on the same boundary and comes back with it.

Measured at 1000 items and 5 update transactions, three runs of each variant of one tree back to back, medians: 215 ms with the walk absent from the raw write path, 1,407 ms with it, 216 ms with the hook. What a write-back records no longer grows with the length of the list, which is what the new test in `data-updating.test.ts` pins.

The investigation is recorded in
`docs/history/development/performance/2026-08-read-modify-write-artifact-walk.md`, including two defects it turned up in the benchmark file's own write accounting, which are left for their own change.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5482?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/history/README.md
+++ b/docs/history/README.md
@@ -304,6 +304,10 @@ One line per archived document; each document's header carries the fuller
   — why deriving demand refcounts from the roots on every change cost a growing
   list cubic work, and what incremental maintenance measured instead, August
   2026.
+- [2026-08-read-modify-write-artifact-walk.md](development/performance/2026-08-read-modify-write-artifact-walk.md)
+  — why reading a list, changing one element, and writing it back grew seven
+  times more expensive: the builder-artifact walk on the raw write path read
+  into the query results a read hands out, August 2026.
 - [coverage-ratchet-noise-2026-07-28.md](development/coverage-ratchet-noise-2026-07-28.md)
   — why a rename-only pull request owed coverage debt: a wall-clock-guarded
   diagnostic and a shard re-partition each moved the `packages/runner`
--- a/docs/history/development/performance/2026-08-read-modify-write-artifact-walk.md
+++ b/docs/history/development/performance/2026-08-read-modify-write-artifact-walk.md
@@ -1,0 +1,128 @@
+---
+status: historical
+created: 2026-08-06
+archived: 2026-08-06
+reason: "Investigation of the read-modify-write benchmark regression traced to the artifact walk."
+---
+
+# Read-modify-write regression from the artifact walk, August 2026
+
+## Result
+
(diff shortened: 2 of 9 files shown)


danfuzz:
LGTM, but please also check with @seefeldb. Thanks for finding this!

PR #5491, opened 2026-08-07
refactor(fuse): project JSON onto the filesystem through one queue

Hixie:
The FUSE tree builder projected a JSON value onto filesystem nodes twice over. One implementation was a recursive synchronous walk, and the other was a queue-based walk that handed the event loop back every two hundred nodes. Both had to encode the same rule set: scalars, circular references, sigil links that become symlinks, arrays, objects, the `.json` aggregate sibling beside every directory, skipping streams and handlers, and threading the CFC annotation context down to each child. Every rule had to be changed in two places, and a rule that changed in only one place would make a piece look different depending on whether the caller happened to be on the synchronous path or the asynchronous one.

Both entry points also carried a `seen` parameter, a `WeakSet` of objects to treat as already visited. No caller ever passed one: all three production call sites and every test passed `undefined`, so the helper that consulted it always answered false and the initial ancestor list was always empty. An exported `BuildJsonTreeOpts` interface described a set of options nothing referenced.

The queue-based walk is now the only implementation. Projecting one node creates its entry and pushes its children onto the shared queue; a batch runs up to two hundred nodes and reports whether any remain. The synchronous entry point runs the batches back to back, and the two asynchronous entry points await a turn of the event loop between them. A queue slot is emptied as its node is taken, so a build holds the nodes still waiting rather than every node it has passed. The `seen` parameter, the helper that read it, the recursive walk, and the unused options interface are all gone, along with the positional `undefined` every call site passed for `seen`.

One observable change comes with this. A synchronous build now assigns inode numbers level by level rather than depth first, because the queue reaches every child of a directory before it reaches any grandchild. The entries within any one directory are still created in the same order as before, since a node's children go onto the queue together, so directory listings are unaffected; only the numbers attached to the nodes differ. Nothing reads those numbers as an ordering: no test asserts a literal inode, the cell bridge reconciles trees by name and path, and CFC annotation references are derived from paths rather than from inodes. Dropping the recursion also raises the nesting depth a value can have before a build exhausts the stack, though it does not remove the limit, because each directory level stringifies its own subtree for the `.json` sibling and that stringifying is itself recursive.

A new test builds a value spanning several batches through both entry points and compares the resulting trees node for node, inodes included. The inodes are what give the comparison its teeth: two builds of the same value into fresh trees allocate from the same starting number, so equal inodes mean both builds created the same entries in the same order, and the assertion fails against the arrangement this change replaces. Since comparing two builds against each other cannot catch a fault in the batching itself, which would corrupt both alike, the test also reads the deepest leaf of the last group directly, many batches past the first node.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5491?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/fuse/tree-builder.ts
```diff
@@ -170,15 +170,7 @@ export function buildFsProjection(
...
-/** Options for buildJsonTree beyond the required params. */
-export interface BuildJsonTreeOpts {
-  seen?: WeakSet<object>;
-  resolveLink?: (value: unknown, depth: number) => string | null;
-  depth?: number;
-  annotation?: CfcJsonAnnotationContext;
-}
-
-const ASYNC_BUILD_BATCH_SIZE = 200;
+const BUILD_BATCH_SIZE = 200;
```
bfollington, line 173: Context, since I added the async walk and wanted to check whether collapsing it was safe.

It went in as a perf fix (#3215): a synchronous walk over a large result cell holds the daemon's event loop, and the kernel blocks on the daemon, so an unrelated `ls` hangs behind a rebuild. I benchmarked rather than trusting the commit message. Building piece-shaped values, measuring the largest gap between beats of a zero-delay timer heartbeat — i.e. how long a FUSE request sits unanswered:

| case | inodes | sync ms | sync stall | async ms | async stall |
|---|---|---|---|---|---|
| medium piece | 11,013 | 17.7 | **17.8** | 113 | 3.7 |
| large piece | 44,013 | 75.5 | **75.5** | 481 | 8.8 |
| wide flat list | 36,013 | 57.2 | **57.3** | 396 | 7.3 |

The concern was real — sync stall tracks build time exactly — so the yielding is load-bearing and worth keeping. This PR keeps it: both hot call sites still await `buildJsonTreeAsync`, and the one surviving synchronous caller (the `[FS]` frontmatter subtree builder) was already synchronous. I ran the same benchmark on main and on this branch and they're at parity on every case. No objection to landing.

What the benchmark did turn up is that the batching *policy* is wrong, though the mechanism isn't. The yield is the whole cost:

```
setTimeout(0)     2.301 ms/yield
queueMicrotask    0.001 ms/yield
```

At 200 nodes per batch the large case spends 171 yields ≈ 393ms of scheduling against ~75ms of work. A microtask isn't the fix — it wouldn't let FUSE I/O in — but budgeting the batch by elapsed time instead of node count is. As a follow-up, not a request here:

| case | async today | 4ms budget | stall today | stall budgeted |
|---|---|---|---|---|
| medium piece | 113 ms | **25 ms** | 3.7 | 6.5 |
| large piece | 481 ms | **119 ms** | 8.8 | 10.3 |
| wide flat list | 396 ms | **82 ms** | 7.3 | 8.7 |

4× faster, stall still bounded, 348 fuse tests green. That's a ten-line change against your single `runJsonTreeBatch` and would have been two changes against the two walks — so it's an argument for this PR rather than against it.

Two limits worth recording: a single node can't be interrupted, and the most expensive one is the root, whose `.json` aggregate stringifies the entire value (4.5ms on the large case), so there's a floor below which shrinking the budget buys nothing. And every directory stringifying its own subtree makes the build O(n·depth) — that, not the walk, is where the time actually goes.

Happy to put the benchmark harness up as a stacked PR on top of this one if you want it in the tree — entirely your call whether it earns its keep.

<sub>Archaeology and benchmarks run with Claude Code; numbers measured on darwin/Deno, single machine.</sub>

bfollington (approved):
LGTM, did some benchmarking in https://github.com/commontoolsinc/labs/pull/5491#discussion_r3739276970 to investigate the original motivation. Happy to throw the benchmark code up in a stacked PR if you are interested.

Hixie:
landed on your lgtm; please don't hesitate to add more benchmarks to the tree if there's something valuable to add!

PR #5492, opened 2026-08-07
refactor(piece): one space-level surface instead of two

Hixie:
The piece package exposed two overlapping objects for the same thing — a space's pieces. `PieceManager` held the registry, the run/setup/link operations, and the default-pattern linkage. `PiecesController` wrapped a manager it also handed back through `manager()`, adding only the default-root lifecycle (`ensureDefaultPattern`,
`recreateDefaultPattern`, `checkAndUpdateDefaultPattern`) and a static `initialize()` convenience. Neither class subsumed the other, so every consumer had to carry both: the runtime worker kept a `{ pieceManager, cc }` pair per space, the FUSE cell bridge stored a manager and re-wrapped it in a controller, and the CLI built controllers from managers it went on using directly. Callers that needed something the controller did not forward reached through `manager()` to get at the object underneath, which meant the wrapper never actually contained anything.

Eight of the controller's thirteen public methods were one-line forwards to the manager, and the two files each carried their own copy of the same timing scaffolding: a `CF_CLI_TRACE_TIMINGS` environment constant, the `piece` and `piece.update` loggers, and a byte-identical phase-timing helper.

This merges the two into a single class. `PiecesController` keeps its name — it is the name the documentation, the skills, and most test code already use — and absorbs everything the manager did. The delegation block, the second copy of the timing scaffolding, and the `manager()` accessor are gone, and so are the paired fields in the runtime worker and the cell bridge.

Method names had to move in three places, because the two classes had methods that collided or that only made sense as forwards:

  - The manager's `get()` returned the piece's canonical result `Cell`, while the controller's `get()` returned a `PieceController` wrapping it. The controller spelling has far more callers, so it keeps the name; the cell-returning one is now `getPieceCell()`.
  - `remove()` existed twice, once taking a `Cell` and once taking an id. It is now one method accepting either, and it settles storage after a successful removal the way the controller's version did.
  - `start()` and `stop()` were forwards to `startPiece()` and `stopPiece()` that added nothing; callers use the latter now.

`PieceController` (the per-piece controller) is unchanged in substance, but its `manager()` accessor now hands back the merged object, so it is named `pieces()`.

Two smaller pieces of scaffolding go with the merge. `ops/acl-manager.ts` was a one-line re-export of the runner's `ACLManager`, reached only through the controller's `acl()` helper, whose only callers were four integration tests; those tests construct the runner's `ACLManager` directly and both the helper and the re-export are deleted. The `disposeCheck()` guard, which turned a call on a disposed controller into a clear error, covered ten of the old controller's thirteen methods; applied to the merged class it would have covered ten of about forty, which is a guard whose silence tells the reader nothing. It is deleted along with the two tests that asserted its message — a disposed controller still fails, now with the runtime's own error.

Consumer-side naming follows: the CLI's `loadManager` is `loadPieces`, the FUSE bridge's `ManagerLoader`/`loadManager` option pair is `PiecesLoader`/`loadPieces`, the runtime worker's `managerFor` is `piecesFor`, and the CLI's `createController` dependency-injection hook is gone — a test that wants to supply a fake controller now returns it from `loadPieces`, since there is nothing left to wrap.

Live documentation naming the old class is updated in the same change, along with `docs/check.vocabulary.json`.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5492?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Hixie:
Claude tells me the open cubic comments correspond to pre-existing issues in code that this PR merely moves around, and that it is updating the topics board accordingly.

PR #5497, opened 2026-08-07
fix(cli): a projection keyword names its container, so `type` is inferred

mpsalisbury:
## Why

`cf piece get --schema` silently returned nothing unless the caller wrote
`"type": "object"` at *every* object level. `--schema '{"properties":{"label":true}}'`
returned `{}`. Nested was worse, because that is where a caller forgets:
`{"type":"object","properties":{"topic":{"properties":{"title":true}}}}` returned
`{"topic":{}}`.

In JSON Schema `properties` applies to objects whether or not `type` is stated,
so this was a trap with no upside — and a silent-empty one, which is the kind a
caller cannot diagnose from the output.

Stacked on #5470.

## What Changed

`normalizeProjectionSchema` fills in the container a position's own keywords
already name, at every level. `properties`/`additionalProperties`/`required`/
`minProperties`/`maxProperties` imply an object; `items`/`minItems`/`maxItems`/
`uniqueItems` imply an array. A stated `type` always wins. Arrays are tested
first, matching `projectionMask`, so the mask and the output schema cannot
disagree about the container.

## Why here and not in traversal

The root cause is `ContextualFlowControl.schemaAtPathInternal`, which descends
`properties` only under `type: "object"` and `items` only under `type: "array"`.
The projected value was computed correctly and then discarded on read-back
through the untyped output schema.

Loosening that would change what an untyped schema means for every reader in
the system — the read selector, the `ifc` walk, `getSchemaAtPath`'s callers, the
write path — to fix one flag's ergonomics. The projection schema is the one
schema the CLI already owns and already rewrites, and `projectionMask` was
*already* inferring the container this way, so the two were out of step. Now
they agree.

## Blast radius

One production caller: `cf piece get --schema`. `true`, `{}`, `false`, and a
bare `$link` marker all normalize byte-identically, so the four spellings stay
distinct. Concise `--schema id,title` already emitted `type` explicitly.

Behavior changes, all in the direction of the projection the caller asked for:
untyped object keywords now project instead of returning `{}`; untyped `items`
likewise; `{"required":[…]}` alone now keeps everything, matching
`{"type":"object"}`.

An untyped `items` at the **root** against a non-array value now refuses with
"An array-rooted JSON --schema can only be applied to an array value." It
previously returned `{}` to the caller *after* computing and committing the
unprojected object — including fields the projection excluded — into the
session result cell, because the mask and the output schema disagreed.
Materializing more than was selected and then hiding it behind an empty answer
is worse than either half; both halves are now covered by tests.

## Test plan

`deno task test` in packages/cli (123 passed) and packages/runner (1156
passed); `deno task check`, `fmt`, `lint`, `check-docs`, `check-skill-facts`.
Six tests added, none deleted, test file is additions-only. Five of the six were
confirmed to fail on the base commit, with the observed failure recorded for
each; the sixth is a deliberate guard that `true`/`{}`/`false`/bare-marker stay
distinct and passes on both sides by design.

The full-form example in `docs/plans/shaped-reads-and-verb-results.md` now works
as literally written, asserted verbatim.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5497?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -147,24 +147,29 @@ missing value and is also falsey. Filtering happens before schema projection.
 
 For an array result, the concise form describes each item. An inline/file JSON
 Schema describes the complete returned value, so a schema combined with
-`--filter` must have an array root. Object `properties` are a whitelist by
+`--filter` must have an array root. Object `properties` are an allowlist by
 default; use `"additionalProperties": true` to retain unspecified properties.
 Projection schemas support structural `properties`, `items`, and scalar leaf
-schemas. In an array-item projection, a scalar leaf whose declared type does not
-match the stored value is omitted by the runtime rather than reported as an
-error; prefer `true` leaves unless that type filtering is intentional. Schema
-combinators and references are rejected in caller-supplied projection schemas.
-Concise dotted paths follow the declared source schema through nested arrays, so
-`comments.body` selects `body` from every comment without retaining comment
-siblings. The projection preserves source-declared nullable items and
-properties, including `type` arrays and `anyOf` unions. When the source schema
-does not identify a nested container, the concise form applies the same field
-mask across arrays encountered in the value so siblings still cannot leak; use
-an explicit JSON Schema when the output schema itself must be fixed. If a
-present source value cannot materialize the transform, the command exits nonzero
-and states that the failure is not JSON `null`. An absent optional source
-remains the ordinary successful `null` CLI response, as does a valid projected
-null.
+schemas. A position that names an object keyword (`properties`,
+`additionalProperties`, `required`) projects an object and one that names
+`items` projects an array, at every level of nesting and whether or not it also
+states `type` — `{"properties":{"topic":{"properties":{"title":true}}}}` returns
+`topic.title`. Neither `true` nor `{}` names a container, and both keep
+everything at the position they sit at. In an array-item projection, a scalar
+leaf whose declared type does not match the stored value is omitted by the
+runtime rather than reported as an error; prefer `true` leaves unless that type
+filtering is intentional. Schema combinators and references are rejected in
+caller-supplied projection schemas. Concise dotted paths follow the declared
+source schema through nested arrays, so `comments.body` selects `body` from
+every comment without retaining comment siblings. The projection preserves
+source-declared nullable items and properties, including `type` arrays and
+`anyOf` unions. When the source schema does not identify a nested container, the
+concise form applies the same field mask across arrays encountered in the value
+so siblings still cannot leak; use an explicit JSON Schema when the output
+schema itself must be fixed. If a present source value cannot materialize the
+transform, the command exits nonzero and states that the failure is not JSON
+`null`. An absent optional source remains the ordinary successful `null` CLI
+response, as does a valid projected null.
 
 Both transforms run as a short-lived computed pattern in the caller's session.
 When the declared source schema fixes the root container shape, the pattern
--- a/packages/cli/lib/cell-selection.ts
+++ b/packages/cli/lib/cell-selection.ts
@@ -553,6 +553,49 @@ const UNSUPPORTED_PROJECTION_KEYS = new Set([
   "contentSchema",
 ]);
 
+/**
+ * The keywords that apply to exactly one container. Naming one says which
+ * container the position describes, so the traversal that reads the projection
+ * back needs no separate `type` from the caller.
+ */
+const OBJECT_PROJECTION_KEYS = [
+  "properties",
+  "additionalProperties",
+  "required",
+  "minProperties",
+  "maxProperties",
+];
+const ARRAY_PROJECTION_KEYS = [
+  "items",
+  "minItems",
+  "maxItems",
+  "uniqueItems",
+];
+
+/**
+ * The container a projection position describes but did not state. Schema
+ * traversal descends `properties` only under `type: "object"` and `items` only
+ * under `type: "array"`, so an omitted `type` silently empties the position —
+ * and a nested position is where a caller omits it. Filling it in here keeps
+ * that requirement out of what a caller has to know, and keeps traversal's
+ * meaning of a stated type intact for every other reader.
+ *
+ * Arrays are tested first, matching {@link projectionMask}, so a position that
+ * names both vocabularies reads as the array it named rather than disagreeing
+ * with the selector built from it.
+ */
+function impliedProjectionType(
+  declared: Record<string, unknown>,
+): "object" | "array" | undefined {
+  if (declared.type !== undefined) return undefined;
+  if (ARRAY_PROJECTION_KEYS.some((key) => declared[key] !== undefined)) {
+    return "array";
+  }
+  return OBJECT_PROJECTION_KEYS.some((key) => declared[key] !== undefined)
+    ? "object"
+    : undefined;
+}
+
 /** A projection schema with its `$link` markers lifted out of it. */
 interface NormalizedProjectionSchema {
   schema: JSONSchema;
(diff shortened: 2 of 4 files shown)


mpsalisbury (commented):
_cf-review via Claude Fable 5, on behalf of @mpsalisbury. Reviewed this PR's own delta vs `impl/a12-link-marker` (stacked on #5470)._

# cf-review: PR #5497 — "fix(cli): a projection keyword names its container, so `type` is inferred"

**Scope:** 4 files, +262/−25, all in `packages/cli` + `skills/cf/SKILL.md` · **Coverage:** deep-read of `packages/cli/lib/cell-selection.ts` at pinned head 45e9b5a42 (`impliedProjectionType`, `normalizeProjectionSchema`, `projectionMask`, `projectValue`, `selectSourceSchema`, `resolveProjection`, `deriveSelectedValue`), full read of test/README/SKILL diffs, plus `packages/runner/src/cfc.ts` `schemaAtPathInternal` to confirm the stated root cause · **Checks run:** static analysis only — no tests executed, no code needed running to confirm any finding.

The change is what it says: `normalizeProjectionSchema` stamps `type` from a position's own container keywords at every recursion level (properties children, `additionalProperties`, `items`), stated `type` wins, arrays win ties, and the else-if that defaults `additionalProperties: true` now reads the stamped type (`schemaTypes(result)` vs base's `schemaTypes(declared)`). Scope is coherent; nothing leaked.

## 🔴 Blocking

None.

## 🟡 Improvements

1. **New refusal reaches one previously-succeeding (if nonsensical) input** — `resolveProjection` (cell-selection.ts:1418–1428) · An untyped schema carrying only an array *constraint* keyword and no `items` — `{"minItems":1}`, `{"maxItems":5}`, `{"uniqueItems":true}` — now infers `type:"array"` and hits the root refusal against a non-array value. On base these proceeded (mask degenerated to `true`, `projectValue` ignored the keyword) and returned the **full value**. The PR body's "behavior changes, all in the direction of the projection the caller asked for" doesn't quite cover this: the caller named an array container against a non-array value and previously got a silently-unconstrained full read. The new refusal is the more honest answer, and the mirror case is *fixed* (those same schemas against an array value threw "must describe the returned array" on base and now work) — but it's a behavior change the body doesn't name, worth consciously accepting. **suspected** (traced statically through both base and head code paths; not executed).

2. **Both-vocabulary positions are documented but untested** — `impliedProjectionType` JSDoc (cell-selection.ts:583–585) promises a position naming both vocabularies reads as array; no test pins `{"items":…,"properties":{…}}` → array-typed normalization, nor the nested variant. Since this is exactly the tie the "arrays tested first, matching projectionMask" comment exists to justify, a one-assertion addition to the normalization test would keep the promise from silently drifting. **verified** (absence confirmed in the full test diff).

3. **`{"required":[…]} alone keeps everything` is pinned only at the normalization layer** — the unit test asserts the normalized form `{type:"object", required:["id"], additionalProperties:true}`, and end-to-end "keeps everything" then rests on existing `{"type":"object"}` coverage. Semantically the behavior is right for the read selector: `required` restricts nothing property-wise, `additionalProperties:true` makes `projectionMask` return `true`, so read, projector, and output schema all agree on "everything" — verified by reading all three consumers. A single `deriveSelectedValue` assertion would make the PR-body claim self-evidencing, but this is optional. **verified** (code paths); the gap itself is a test-rigor nudge, not a bug.

## ⚪ Nits

- **README keyword list reads exhaustive but isn't** — `packages/cli/README.md`: "a position that names an object keyword (`properties`, `additionalProperties`, `required`)" omits `minProperties`/`maxProperties`, and the array sentence names only `items`, omitting `minItems`/`maxItems`/`uniqueItems` — all of which trigger inference in `OBJECT_PROJECTION_KEYS`/`ARRAY_PROJECTION_KEYS`. The omitted ones are constraint keywords a projection caller rarely writes, but the parenthetical presents itself as the set. **verified**.

## Focus-area verification results (all verified by reading pinned head)

- **`projectionMask` agreement:** the keyword lists do *not* literally match — `projectionMask` infers array only from `items` and object only from `properties`. Agreement holds anyway, structurally: the mask is only ever computed on normalized schemas (json path) or explicitly-typed built schemas (concise path builds `type:"object"` at every level), and normalization now stamps `type` for every keyword in the wider lists, which the mask's `type ===` tests then honor. One residual asymmetry: `projectionMask` tests `additionalProperties: true`/record *before* the array branch, so `{"items":X,"additionalProperties":true}` normalizes array-typed but masks as `true` — that only widens the read, never changes the returned value, and is pre-existing for the explicitly-typed spelling. Not introduced here.
- **`true`/`{}`/`false`/bare-marker byte-identity:** verified in code (early return for `true`; `false` throws; `declared={}` ⇒ no inference ⇒ `{}` survives; bare marker still reduces to `false` because inference runs on `declared` *after* the marker is destructured out) and pinned by the distinct-spellings test.
- **`$link` interaction:** marker lifted before inference, so a marked position with untyped keywords infers normally and a bare marker can't accidentally infer an object — covered by the new "states no type" marker test.
- **Combinators/`$ref`:** unreachable by inference — `UNSUPPORTED_PROJECTION_KEYS` refuses `$ref`/`anyOf`/`allOf`/`oneOf`/`prefixItems`/`contains`/`patternProperties` at every node before inference runs. A `type` array (`["object","null"]`) suppresses inference via `declared.type !== undefined`.
- **Root refusal prevents the materialization leak:** `resolveProjection` throws before `runtime.edit()`/`runtime.run` in `deriveSelectedValue`, so the previously-committed unprojected object genuinely cannot be computed or committed anymore.
- **Root cause claim:** confirmed — `schemaAtPathInternal` (packages/runner/src/cfc.ts:651, 684) descends `properties` only under `type:"object"` and `items` only under `type:"array"`; fixing in the CLI-owned normalization rather than loosening shared traversal is a defensible scoping call.
- **"Confirmed to fail on base" plausibility:** traced all six tests against base code — five fail on base (two via `{}`/`{topic:{}}` silent-empty, one via base's "must describe the returned array" throw, one via no-throw where head throws, one via missing `type` in normalized output), and the spellings guard passes both sides. Matches the PR body exactly.
- **Coherence:** README, `skills/cf/SKILL.md`, and the plan docs were swept; no live doc still claims `type` is required at every level; the `docs/plans/shaped-reads-and-verb-results.md` full-form example exists verbatim at head and is asserted verbatim in the new nesting test. The whitelist→allowlist wording fix rides along. `docs/plans/shaped-reads-implementation.md` names the remaining known gap (unknown-keyword typos like `propertys` still silently return `{}`) as explicitly future-staged work — correctly out of this PR's scope.

## Possible follow-ups

- The typo-keyword tier (`{"propertys":…}` silently returning `{}`) already planned in `docs/plans/shaped-reads-implementation.md` — this PR's inference makes that the last remaining silent-empty path in projection parsing.
- Consider whether a position naming both container vocabularies should be refused outright (like combinators are) rather than resolved arrays-first; the current choice is deterministic and JSON-Schema-consistent for array instances, but a refusal would be more in the spirit of this PR's "silent surprise is worse than an error."

**Verdict:** Sound and carefully reasoned — no blocking findings; the inference, the mask agreement, and the new refusal all check out against the code, with only an unnamed edge-case behavior change (constraint-only array keywords vs non-array values) to consciously accept.

PR #5500, opened 2026-08-07
feat(cli): `--select` takes the concise path list, `--schema` takes schemas

mpsalisbury:
## Why

`cf piece get --schema` accepted two different input languages through one
flag: a concise dotted path list (`--schema id,title,comments.body`) and a full
JSON Schema (`--schema '{"type":"object",…}'`, including `@file`). One flag
taking two languages means every error message has to hedge about which one you
meant, and a caller cannot tell from the flag name what it takes.

Splitting them is also what lets `@` land later: the suffix in `topic@,title`
only makes sense in the concise language, and it needs a flag that means only
that.

Stage F2 of the shaped-reads plan (#5435). Stacked on #5497.

## What Changed

- **`--select`** takes the concise path list.
- **`--schema`** keeps full JSON Schemas and `@file`.
- Concise input on `--schema` keeps working, with **no deprecation warning** —
  there is no removal date, and warning on every invocation would be noise in
  the skills that teach the current spelling.

Both flags converge on the existing internal representation; there is no second
path. A `ProjectionFlag` rides along so every message names the flag the caller
actually wrote.

## Supplying both is refused

At the argument surface (`conflicts: ["select"]`), not in the selection layer,
exiting 2 with the usage screen. The conflict is a property of which flags were
written, not of the data — and `CellSelectionError` is documented as "a data
condition rather than bad arguments", so routing an argument conflict through
it would blur a distinction the code draws deliberately. It also keeps
`packages/cli/lib/` free of any Cliffy import, which it is today.

## Exit criterion

The plan's exit criterion was that `piece-get-transform.test.ts` pass
**unchanged** — a moved expectation would mean the surface change altered
behavior. The diff on that file is **97 insertions, 0 deletions**: it contains
no `-` lines at all. `--schema`'s messages are byte-identical, because
`--schema` is the default everywhere the flag is threaded.

## Test plan

`deno task test` in packages/cli (1672 + 123 passed) and packages/runner (1156
`check-no-waitfor`, `check-conflict-markers` — all pass.

Both flags driven through the real `cf` binary for the failure paths (they fail
before the network): a bad `--select` path, a JSON Schema or `@file` passed to
`--select`, a bad `--schema` path, and the both-flags conflict.

**Not exercised:** a networked `cf piece get --select` returning projected data
— that path needs a live toolshed and self-skips locally; CI's `cli-integration`
job covers it. No `--select` case was added to `verbs-over-the-cli.sh`, whose
subject is `$link` JSON schemas.

## Noted, not fixed

`--select` is only on `cf piece get`. `cf piece call`, `cf wish` and `cf exec`
are meant to reach the same read step later; when they do they should take both
flags and the same `conflicts` declaration.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5500?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -129,7 +129,7 @@ to a smaller shape:
 cf piece get --piece ID items --filter '.status == "open"'
 cf piece get --piece ID items \
   --filter '.status == "open" and .score >= 10' \
-  --schema id,title,author.name
+  --select id,title,author.name
 ```
 
 `--filter` is jq-inspired rather than a full jq interpreter. It applies only to
@@ -139,13 +139,16 @@ arrays and accepts value paths (`.status`, `.author.name`, `.["display-name"]`,
 missing path simply does not match. A stored `undefined` is treated like a
 missing value and is also falsey. Filtering happens before schema projection.
 
-`--schema` accepts one of three forms:
+Two flags project, one per input language:
 
-- a comma-separated field projection such as `id,title,author.name`;
-- an inline JSON Schema object;
-- `@path/to/schema.json`.
+- `--select` takes a comma-separated field list such as `id,title,author.name`;
+- `--schema` takes an inline JSON Schema object or `@path/to/schema.json`, and
+  also accepts the same field list `--select` takes.
 
-For an array result, the concise form describes each item. An inline/file JSON
+A command that names both has not said which shape it wants, and is refused
+before the read.
+
+For an array result, the field list describes each item. An inline/file JSON
 Schema describes the complete returned value, so a schema combined with
 `--filter` must have an array root. Object `properties` are an allowlist by
 default; use `"additionalProperties": true` to retain unspecified properties.
--- a/packages/cli/commands/piece.ts
+++ b/packages/cli/commands/piece.ts
@@ -1409,7 +1409,7 @@ PATH FORMAT: Use forward slashes and numeric indices for arrays.
   )
   .example(
     cliText(
-      `cf piece get ${EX_ID} ${EX_COMP_PIECE} items --schema id,title`,
+      `cf piece get ${EX_ID} ${EX_COMP_PIECE} items --select id,title`,
     ),
     "Project each returned item to selected fields.",
   )
@@ -1435,8 +1435,15 @@ PATH FORMAT: Use forward slashes and numeric indices for arrays.
     "Filter an array with a jq-inspired predicate",
   )
   .option(
+    "--select <fields:string>",
+    "Project output to comma-separated field paths",
+  )
+  .option(
     "--schema <schema:string>",
-    "Project output with comma-separated fields, inline JSON Schema, or @file",
+    "Project output with an inline JSON Schema or @file",
+    // Both flags carry the one projection, so a command naming both has not
+    // said which shape it wants. Refuse before the read rather than pick.
+    { conflicts: ["select"] },
   )
   .arguments("[path:string]")
   .action(async (options, pathString) => {
--- a/packages/cli/commands/piece.ts
+++ b/packages/cli/commands/piece.ts
@@ -1440,7 +1440,8 @@ PATH FORMAT: Use forward slashes and numeric indices for arrays.
   )
   .option(
     "--schema <schema:string>",
-    "Project output with an inline JSON Schema or @file",
+    "Project output with an inline JSON Schema, @file, or the --select " +
+      "field list",
     // Both flags carry the one projection, so a command naming both has not
     // said which shape it wants. Refuse before the read rather than pick.
     { conflicts: ["select"] },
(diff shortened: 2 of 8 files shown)


mpsalisbury (commented):
_cf-review via Claude Fable 5, on behalf of @mpsalisbury. Reviewed this PR's own delta vs `impl/schema-type-inference` (stacked on #5497)._

## cf-review — PR #5500 `--select`/`--schema` split

**Scope:** 8 files, +293/−82, stacked on #5497 (reviewed as `gh pr diff 5500` only) · **Coverage:** deep-read (full diff + head-file audit of every error-construction site in `packages/cli/lib/cell-selection.ts` and the `piece get` action) · **Checks run:** mechanical insert/delete count on the raw diff; real `cf` binary driven in a throwaway worktree (conflict exit code, `--select` misuse paths); `deno fmt --check` + `deno lint` on the touched files. No suites.

### 🔴 Blocking

None found.

### 🟡 Improvements

1. **`packages/cli/test/piece.test.ts` — conflict test doesn't pin exit 2.** The new "refuses a piece get command that names both projection flags" test asserts `expect(code).not.toBe(0)`, but the design point the PR body argues — argument-surface refusal (exit 2, usage screen) vs. `CellSelectionError` data condition (exit 1) — is exactly the distinction the exit code carries. I ran the binary: the conflict exits **2**, and `--select '{…}'` exits **1**, so today the routing is right, but the test would still pass if a refactor demoted the conflict to a data error. Fix: `expect(code).toBe(2)`. **Verified** (ran head binary; read the test).

### ⚪ Nits

2. **PR body claim: "keeps `packages/cli/lib/` free of any Cliffy import, which it is today"** — not quite: `packages/cli/lib/completion/line.ts:14` and `lib/completion/static.ts:9` carry type-only `@cliffy/command` imports, on main and on the base branch alike. Pre-existing and untouched by this PR; the substantive point (the conflict lives in `commands/piece.ts`, `cell-selection.ts` stays Cliffy-free) holds. Worth a one-line correction in the body so the stated invariant matches the tree. **Verified.**
3. **`schemaMayBeArray(schema, flag = "--schema", …)`** (`cell-selection.ts:1146-1148` at head) — the defaulted flag means a future internal caller on a `--select` projection silently gets `--schema`-attributed errors. The only production call site threads the flag correctly today; making the parameter required (the tests already pass it explicitly in one case) removes the hazard. **Verified** (single call site at line 1184).
4. **Lib-level precedence is documented but untested.** `parseCellSelectionOptions`'s JSDoc says "select is the one read when both do [arrive]" — defensible defensive behavior since the surface refuses both, but no test pins it. One `{ select, schema }` case would do. **Verified** (read; no test exercises both keys).

### Verified clean (the focus areas)

- **ProjectionFlag threading:** every remaining hardcoded `--schema` in an error message at head is reachable only from `kind: "json"`, which only `--schema` can produce — the `Invalid --schema at <path>` family in `normalizeProjectionSchema` (lines 620–700), the `@file` read/parse errors (819–866), and the array-root mismatch pair at `resolveProjection` (1487–1491, guarded by `projection.kind === "concise"` early-return). Concise-path errors all take the flag; the map-error translation in `deriveSelectedValue` uses `selection.projection!.flag` and the `!` is guarded by `projection?.projectsArrayItems`. Markers only exist on JSON projections, so the filter+marker refusal can't misattribute either. No stale site found.
- **Input-language detection, live:** `--select '{"type":"object"}'` and `--select @shape.json` → exit 1, `--select takes comma-separated field paths. Pass a JSON Schema or an @file to --schema instead.` — through the data-error report, no help-screen dump. `--select 'topic@,title'` (the future PR-5504 spelling) → `Invalid --select field path "topic@"`, a clean refusal that leaves 5504 free to assign it meaning; only a *leading* `@` gets the pointed-at-`--schema` message, which is right since `@` was never expressible in the concise language.
- **The 97/0 claim:** counted mechanically from the raw diff — 97 insertions, zero deletion lines in `piece-get-transform.test.ts`. Exit criterion honestly met.
- **Docs/skills coherence:** `.claude/skills/` and `.agents/skills/` are symlinks (mode 120000) to `skills/`, so the mirrors follow automatically. The only concise-on-`--schema` example left at head is `docs/plans/cli-surface-shape.md:79`, inside the problem statement motivating this split — and since concise-on-`--schema` deliberately keeps working, nothing anywhere is actively wrong. `verbs-over-the-cli.md` uses only JSON `--schema` (still correct). Each new test guards a nameable principle (representation convergence, refusal language, error self-naming).

### Possible follow-ups

- When `piece call` / `wish` / `exec` gain the read step, carry both flags plus the same `conflicts` declaration (the PR body already notes this).
- When the stack lands, nudge `docs/plans/shaped-reads-and-verb-results.md:65` ("today both ride `--schema`") — stage narration that goes stale as F2 merges.

**Verdict:** Clean, careful change that does what it says — no blocking findings; pin the conflict's exit 2 in the test and it's ready.

PR #5501, opened 2026-08-07
feat(ts-transformers,runner): carry a verb's declared result to its module

mpsalisbury:
## Why

A verb's handler is typed `Stream<E, R>`. `R` was dropped at compile time, so a
receipt for a verb whose result contains anything reactive — including every
verb that returns a child piece, which is the motivating case — carried no
description of what it holds. A reader had nothing to push into the fetch and
nothing to match against but a runtime value.

The consumer already exists and is already correct: the reactive branch calls
`updateResultSchemaMeta(tx, resultCell, pattern.resultSchema)`, whose first line
is `if (resultSchema === undefined) return;`. It was never missing plumbing. It
was missing the one input.

## Relationship to the withdrawn verb-contract C3 — read this first

`docs/plans/pattern-verb-contract-implementation.md` records that result-schema
emission was built, proven, and then **deliberately withdrawn** before merge
(review, 2026-07-31), with instructions to revisit it as part of the
Fabric-types stream design and not before. This PR is deliberately not that
thing, and each of the three recorded objections is addressed rather than
re-argued:

| | withdrawn C3 | this |
|---|---|---|
| mechanism | new `result` **dialect keyword**, sibling to `asCell` | existing `Module.resultSchema` field, an ordinary JSON Schema |
| destination | the pattern's own `resultSchema`, on the stream property | the handler **node's** module, inside `pattern.nodes[]` |
| compat gate | new recursion + result-name permanence rules | none added |
| schema-generator | changed | **untouched** |

1. **"A keyword in durable schemas and append-only baselines hard-commits a
   shape."** Not tripped. A baseline record is exactly
   `{pattern, argumentSchema, resultSchema}` — the pattern's top-level schemas,
   with no slot for a node's module. `check-baselines-append-only` and
   `pattern-compat` (336 patterns) pass.
2. **"The compat rules would refuse its later removal."** Not tripped.
   `assertPatternSchemasBackwardCompatible` reads only
   `previous/candidate.{argumentSchema, resultSchema}` and never walks
   `pattern.nodes[].module`. Removing this later is an ordinary graph edit.
3. **"The value path never needed it."** Still true, and this PR does not
   contradict it — see the honest limitation below.

## Honest limitation

**Nothing consumes the receipt's schema today.** The receipt readback in
`packages/cli/lib/callable.ts` pulls schema-free. This lands a durable
description ahead of its reader; the reader is the receipt-envelope step of the
shaped-reads plan (#5435). Judged on its own it buys nothing yet — it is
groundwork, and worth taking on that basis or not at all.

## What Changed

A single predicate, `declaredVerbResultTypeNode`, decides "this verb declares a
result"; `verb-return-validation.ts` now shares it instead of keeping a private
copy. `action<E,R>`'s result is carried into the emitted `handler`'s third type
argument, schema injection appends `{resultSchema: toSchema<R>()}`, and
`handler`/`handlerInternal` thread it onto `module.resultSchema`, ifc-joined the
way `createNodeFactory` does for `lift`.

A verb with no declared result, or `R = void`, is unchanged and gets no schema —
pinned by a test, so this cannot become mandatory.

## Test plan

`stream-declared-result` — the golden fixture whose own comment named this as
the step that would move it — is the only fixture in the ts-transformers suite
that moved. A `handler<E,T,R>` verb was added beside the `action<E,R>` one so
both authoring surfaces are pinned.

`packages/runner/test/declared-result-e2e.test.ts` compiles a `createNote`-style
verb through the real pipeline, dispatches it, and reads `getMetaRaw("schema")`
off the receipt: it gets the declared result whole, `$defs` included.
Confirmed failing without the change.

runner 1156 · ts-transformers 1130 · cli 123 · piece 34 · api 25 · patterns 66 —
0 failed. `check`, `fmt`, `lint`, `check-docs`, `check-skill-facts`,
`check-no-waitfor`, `check-conflict-markers`, `check-baselines-append-only`,
`pattern-compat`, `check-single-copy-deps`, `check-unused-deps`,
`check-deno-pins` all green.

## Found wrong, not fixed

1. **`{proxy: true}` is silently dropped for every CTS-transformed pattern.**
   Schema injection emits it in the 4th slot; `handlerInternal` only recognizes
   the marker in the 2nd, so `module.writableProxy` is never set.
   `packages/patterns/system/default-app.tsx` relies on it. Pre-existing; this
   change routes around it rather than into it.
2. `handler<E,T>(eventSchema, stateSchema, cb)` with explicit type arguments
   emits a broken 5-argument call. No occurrence in the repo, so latent.

*Naming: C1–C5 above are the **verb contract** arc, not the identically
numbered steps of the shaped-reads plan.*

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5501?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md
+++ b/docs/specs/ts-transformer/ts_transformers_current_behavior_spec.md
@@ -886,6 +886,10 @@ Transforms `action(...)` to handler factory invocation:
 - event schema:
   - no event param -> `never`
   - event param present -> inferred/explicit type
+- `action<Event, Result>(cb)` adds `Result` as a third type argument on the
+  emitted `handler`, which is where the declared result would otherwise be
+  lost: the rewritten call carries schemas rather than the authored type
+  arguments. SchemaInjection lowers that slot (§10.3).
 - callback extraction currently supports arrow callbacks only
 
 ### 9.4 Array-method strategy
@@ -1082,6 +1086,18 @@ structurally representable top-level result:
 - with single function arg:
   - infers event/state schemas from parameters
   - event absent -> `never`; untyped params -> `unknown`
+- with a third type arg `<Event, State, Result>`: appends a trailing options
+  object carrying `{ resultSchema: toSchema<Result>() }`. `action<Event,
+  Result>` reaches the same emission — its lowering to `handler` carries the
+  declared result into this third type-argument slot (§7). An absent third
+  argument, and an explicit `void`, emit no options object at all: a declared
+  result is opt-in and never inferred, so a concise body whose completion
+  value happens to be a cell declares nothing. A call that already passes
+  options (the `{ proxy: true }` form) keeps them, spread into the same object.
+  The runtime reads `resultSchema` off that slot onto the node's
+  `Module.resultSchema` (`packages/runner/src/builder/module.ts`); it does NOT
+  reach the pattern's own `resultSchema`, which is what the piece update gate
+  compares across versions.
 
 ### 10.4 `lift(...)`
 
--- a/packages/runner/src/builder/module.ts
+++ b/packages/runner/src/builder/module.ts
@@ -396,6 +396,22 @@ export function byRef<T, R>(ref: string): ModuleFactory<T, R> {
   });
 }
 
+/**
+ * The options a `handler(...)` call carries in its trailing slot.
+ *
+ * `resultSchema` is the verb's declared result — `action<E, R>` /
+ * `handler<E, T, R>` — lowered here by the CTS schema-injection stage. It is
+ * absent for a value-less verb, and the module then carries no result schema
+ * at all, which is what keeps a declared result opt-in.
+ *
+ * The schema-injected call spreads any options the author wrote into this same
+ * object, so a member here may be one this builder does not read: the writable
+ * proxy is asked for in the state-schema slot above.
+ */
+export interface HandlerOptions {
+  resultSchema?: JSONSchema;
+}
+
 function handlerInternal<E, T>(
   eventSchema:
     | JSONSchema
@@ -403,6 +419,7 @@ function handlerInternal<E, T>(
     | undefined,
   stateSchema?: JSONSchema | { proxy: true },
   handler?: (event: E, props: T) => any,
+  options?: HandlerOptions,
 ): HandlerFactory<E, T> {
   let writableProxy = false;
   if (typeof eventSchema === "function") {
@@ -434,6 +451,14 @@ function handlerInternal<E, T>(
     stateSchema as JSONSchema | undefined,
   );
 
+  // Carry the argument schema's ifc confidentiality through to the declared
+  // result, as `createNodeFactory` does for a lift. Only a declared result is
+  // joined: absence stays absence, so a verb that declares nothing keeps a
+  // module with no result schema rather than acquiring the join's `true`.
+  const resultSchema = options?.resultSchema === undefined
+    ? undefined
+    : applyArgumentIfcToResult(schema, options.resultSchema);
+
   const module: Handler<E, T> & toEncodableForm & toJSON & {
     bind: (inputs: FactoryInput<StripCell<T>>) => Stream<E>;
   } = {
@@ -447,6 +472,7 @@ function handlerInternal<E, T>(
     toJSON: toJSONMethod,
     toEncodableForm: () => moduleToEncodableForm(module),
     ...(schema !== undefined && { argumentSchema: schema }),
+    ...(resultSchema !== undefined && { resultSchema }),
     ...(writableProxy && { writableProxy: true }),
   };
 
@@ -525,6 +551,9 @@ export function handler<E, T, R>(
 export function handler<E, T, R>(
   handler: (event: E, props: T) => R,
 ): HandlerFactory<E, T, R>;
+// The trailing options slot is not part of the authored surface: a pattern
+// declares its result with the type argument above, and the schema-injection
+// stage lowers that declaration into `options.resultSchema` here.
 export function handler<E, T, R = void>(
   eventSchema:
     | JSONSchema
@@ -532,12 +561,14 @@ export function handler<E, T, R = void>(
     | undefined,
   stateSchema?: JSONSchema | { proxy: true },
   handler?: (event: E, props: T) => any,
+  options?: HandlerOptions,
 ): HandlerFactory<E, T, R> {
-  return handlerInternal(eventSchema, stateSchema, handler) as HandlerFactory<
-    E,
-    T,
-    R
-  >;
+  return handlerInternal(
+    eventSchema,
+    stateSchema,
+    handler,
+    options,
+  ) as HandlerFactory<E, T, R>;
 }
 
 // unsafe closures: doesn't need any arguments.
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -5324,6 +5324,7 @@ export class Runner {
 
   private handleJavaScriptHandlerResult(
     tx: IExtendedStorageTransaction,
+    resultSchema: JSONSchema | undefined,
     result: any,
     resultHasReactives: boolean,
     frame: Frame,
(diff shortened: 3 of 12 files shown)


mpsalisbury:
@seefeldb  Reopening the result-schema question — but at a different location, and with a narrower ask than last time.

**First, conceding the original.** Your three reasons for withdrawing C3 all hold, and I don't want to re-argue any of them. Reading them back, they are all about one thing: `result` was a keyword *inside the pattern's schema*, sibling to `asCell` on the stream property. That location is compared across pattern versions and recorded in append-only baselines, and Fabric-types is expected to replace the shape around it. Shipping a keyword there, frozen by rules we wrote ourselves, when the surrounding shape is about to move — the compat rules refusing its own removal is the sharpest way to say why that was wrong.

So the question isn't whether the principle changed. It's whether the objection follows the mechanism to a different place.

**The different place: `module.resultSchema`, on the handler node's module inside `pattern.nodes[]`.** Two properties, both checked rather than assumed:

- `assertPatternSchemasBackwardCompatible` never walks `.nodes` — zero references in `packages/piece/src/schema-compatibility.ts`. It compares only a pattern's two top-level schemas.
- A baseline record is `{argumentSchema, pattern, resultSchema}`, where `pattern` is a filename string. The node graph is not in it.

The field already exists, `lift()` already populates it, and the runner already consumes it — `updateResultSchemaMeta` writes a result cell's durable schema from exactly this field, and returns immediately when it is `undefined`, which is what happens for every handler today. So this is filling an existing channel rather than opening one. schema-generator is untouched.

The pattern graph is still durable, and I don't want to pretend otherwise. The claim is narrower: **removing this later is an ordinary graph edit, not a gated migration.** Replacing an ungated field costs a regeneration; replacing a gated keyword costs a migration the rules refuse. That asymmetry is the whole of the argument.

**Why now rather than with Fabric-types.** The deferral recorded its own cost, and it is the part that has become concrete:

> agents cannot learn a result's shape before calling; the interim for both is prose in the verb's `description`

`cf piece verbs` now exists as a discovery surface and cannot describe what a verb returns. Prose in a description is what an agent has to parse instead. That is the gap we are trying to close, and it is the one you named.

**We are not proposing a stream shape.** Fabric-types still owns `type: "handler", event: …, result?: …` or whatever it becomes. We are asking for a channel that Fabric-types can replace without a migration, not a shape that pre-empts it.

**One design decision we would rather state than have you find.** A receipt can get its schema two ways: derived from the value the runtime settled, or supplied by the verb's declaration. Our default is that **the declared type wins, and the provided value is mapped into it** — a verb that declares a result gets a receipt described by that declaration, and the value has to satisfy it. We don't want a path that sidesteps a declared type by describing whatever happened to be written instead; a result that doesn't match its declaration is a defect worth surfacing, not smoothing over. Value-derived description stays the fallback for verbs that declare nothing.

If you'd rather it went the other way, that's the thing to push back on.

**What we're asking for.** Not approval to ship. A read on whether the objection you raised is about the mechanism or about the location — because if it's the location, this is a different question, and if it's the mechanism, we'd rather hear that now and go on carrying prose until Fabric-types lands.

Context: #5501 is the implementation, currently a draft for exactly this reason. #5519 sequences the remaining verb work and treats this as the one decision the rest is provisional on.

mpsalisbury:
## The ask, on one page

**Should a verb's declared result reach the runtime now, on `module.resultSchema`, rather than waiting for the Fabric-types stream to supply a durable form?**

This PR is the producer half and it is built — declared result type carried to the module and onto the receipt's write-once schema metadata, end-to-end tested, gates green. It is draft because the decision waits, not the code.

### Why this road is not the one that was withdrawn

The July work was withdrawn because a keyword in durable schemas and append-only baselines would hard-commit a shape, and the compatibility rules built alongside it would have refused its later removal. Both objections attach to *where the shape was written*, and this road does not write there. A module field enters no durable schema, so no baseline records it and the compat gate has no rule to apply to it.

### What it buys, which the sequencing plan under-priced

The plan prices this as "narrower than first claimed — narrowing on field selection, and checking a selection before the call." That is accurate about the value path and omits the command surface.

`cf` is already schema-driven on the input side: per-property flags with coercion and validation, a rendered help page, a machine-readable per-verb spec, verb completion against a live piece. None of it authored per pattern. **Every consumer of a declared result is already built and already working — for tools.** `callableCommandSpec` fills `outputSchemaSummary` from a bound pattern's `resultSchema`; a handler hits an early return two lines above and gets nothing.

So `cf piece call <verb> --help` prints `"No output on success."` for a verb that returns one, and each row of `cf piece verbs --json` carries an optional `outputSchema` that is simply omitted for a handler. Its declaration says why: *"Tools only, until handlers gain declared results (verb contract WS-C)."* The codebase is already describing this gap in its own words.

### The one exposure worth your judgement

That the Fabric-types design concludes declared results should **not** reach the runtime at all. In that world the derived default disappears and a caller supplies the same selection by hand — the status quo, not a new cost. Every other outcome swaps the wire's source without changing what a caller sees, because both sources are the pattern author's declared result type.

You own that design, so that call is yours and nobody else can make it.

### What reversing costs, concretely

Reverting this PR touches two runner files, six transformer files, and one golden fixture (`stream-declared-result.expected.jsx`) that re-records.

**No baseline moves.** A pattern baseline carries the pattern's own `argumentSchema` and `resultSchema`, never a verb's `module.resultSchema` — this PR touches no baseline file. The compatibility gate never sees the field, so it has no rule to refuse its removal.

That is the specific expense that withdrew this in July, and it does not attach here.

### If the answer is yes

Plan items 8 and 10 stop being provisional, and verb discovery closes — `cf piece verbs` can finally answer both halves of its question.

### If it waits

The question that needs an answer instead: where do result shapes live in the meantime, and what compares them against the pattern? `cf piece verbs` reports source identity so a client can detect version skew, and that detector cannot see a shape living in prose.

---

Full argument, including the walkthrough it is drawn from: **#5556**. Its evidence is `docs/plans/declared-verb-results-case.md`; the session it walks is `docs/common/verb-session-walkthrough.md`, merged in #5557.

seefeldb:
> ## The ask, on one page
> **Should a verb's declared result reach the runtime now, on `module.resultSchema`, rather than waiting for the Fabric-types stream to supply a durable form?**
> 
> This PR is the producer half and it is built — declared result type carried to the module and onto the receipt's write-once schema metadata, end-to-end tested, gates green. It is draft because the decision waits, not the code.
> 
> ### Why this road is not the one that was withdrawn
> The July work was withdrawn because a keyword in durable schemas and append-only baselines would hard-commit a shape, and the compatibility rules built alongside it would have refused its later removal. Both objections attach to _where the shape was written_, and this road does not write there. A module field enters no durable schema, so no baseline records it and the compat gate has no rule to apply to it.
> 
> ### What it buys, which the sequencing plan under-priced
> The plan prices this as "narrower than first claimed — narrowing on field selection, and checking a selection before the call." That is accurate about the value path and omits the command surface.
> 
> `cf` is already schema-driven on the input side: per-property flags with coercion and validation, a rendered help page, a machine-readable per-verb spec, verb completion against a live piece. None of it authored per pattern. **Every consumer of a declared result is already built and already working — for tools.** `callableCommandSpec` fills `outputSchemaSummary` from a bound pattern's `resultSchema`; a handler hits an early return two lines above and gets nothing.
> 
> So `cf piece call <verb> --help` prints `"No output on success."` for a verb that returns one, and each row of `cf piece verbs --json` carries an optional `outputSchema` that is simply omitted for a handler. Its declaration says why: _"Tools only, until handlers gain declared results (verb contract WS-C)."_ The codebase is already describing this gap in its own words.
> 
> ### The one exposure worth your judgement
> That the Fabric-types design concludes declared results should **not** reach the runtime at all. In that world the derived default disappears and a caller supplies the same selection by hand — the status quo, not a new cost. Every other outcome swaps the wire's source without changing what a caller sees, because both sources are the pattern author's declared result type.
> 
> You own that design, so that call is yours and nobody else can make it.
> 
> ### What reversing costs, concretely
> Reverting this PR touches two runner files, six transformer files, and one golden fixture (`stream-declared-result.expected.jsx`) that re-records.
> 
> **No baseline moves.** A pattern baseline carries the pattern's own `argumentSchema` and `resultSchema`, never a verb's `module.resultSchema` — this PR touches no baseline file. The compatibility gate never sees the field, so it has no rule to refuse its removal.
> 
> That is the specific expense that withdrew this in July, and it does not attach here.
> 
> ### If the answer is yes
> Plan items 8 and 10 stop being provisional, and verb discovery closes — `cf piece verbs` can finally answer both halves of its question.
> 
> ### If it waits
> The question that needs an answer instead: where do result shapes live in the meantime, and what compares them against the pattern? `cf piece verbs` reports source identity so a client can detect version skew, and that detector cannot see a shape living in prose.
> 
> Full argument, including the walkthrough it is drawn from: **#5556**. Its evidence is `docs/plans/declared-verb-results-case.md`; the session it walks is `docs/common/verb-session-walkthrough.md`, merged in #5557.

If it's easy to add .resultSchema now, let's go for it, it's where it'll eventually land. Just wary of that creating more things to chase down, so if that happens, revisit.

PR #5503, opened 2026-08-07
docs(plan): sequence the CLI surface's rename-and-merge arc

mpsalisbury:
## Why

[The CLI surface shape](https://github.com/commontoolsinc/labs/pull/5316) lays out seven steps to the target surface. Steps 1–3 are the read layer and #5435 sequences them. Steps 4–7 had no plan behind them — so the half of the shape that actually changes what a caller types was designed and then left unscheduled.

That half is also the half that can break things. Steps 1–3 add capability and remove nothing. Steps 4–7 rename, alias, and merge, so everything they can break is something a caller already learned — including in scripts and skill files outside this repository, which no sweep here reaches.

Keeping it separate from #5435 rather than folding it in, because the two want different sequencing, different tests, and different appetites for landing quickly.

Stacked on #5435.

## What Changed

One new document, `docs/plans/cli-surface-implementation.md`, indexed in `docs/plans/README.md`.

Four stages, matching shape steps 4–7: positional addresses and the `#argument` suffix; `cf get`/`set`/`call` as top-level names; deprecation; the merges.

It opens with an accounting, because the scale was not obvious to me until I counted. `cf piece` has **twenty** subcommands. The target surface keeps eight, plus `step` and `link` which the shape document settles explicitly. The remaining ten move, and the table says where each one goes and in which step.

## Two decisions it names rather than assumes

**What a command means when given both an address flag and a positional.** Refusing is the safe answer and matches how `--select`/`--schema` were just split — but it is a decision, and it belongs in the same place as that one so a caller meets one rule instead of two.

**What "carries traffic" means.** The shape document defers deprecation until the new spellings carry traffic, and nothing in the CLI measures that. Either a signal gets added or the condition is replaced with one that can be evaluated — a release count, the skills being migrated, a date. Deprecating on an unmeasurable condition means deprecating on a guess, and in practice means stalling indefinitely or landing too early.

## The part I would push back on if someone else wrote it

Stage 4's merges are listed with sizes, and the sizes are dishonest on their own — they cost what the code costs *after* the decision, and the decision is the expensive part. Each pair is two working commands with different behavior, and merging keeps one. The plan asks for characterization tests before each decision for exactly that reason: without them, the surviving behavior is whichever was easier to keep rather than whichever was right.

`piece inspect` vs `inspect piece` may not even be a merge. Top-level `inspect` is offline forensics over stored state, twenty-two subcommands wide; `piece inspect` reports on a live piece. If those are different operations, the fix is naming, not merging — and the plan says so rather than assuming the shape document's grouping settles it.

## Test plan

`deno fmt --check`, `deno task check-docs` (533 blocks), `check-conflict-markers`, `docs-links --orphan` — all pass. Documentation only; no code changes.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5503?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/plans/README.md
+++ b/docs/plans/README.md
@@ -66,3 +66,9 @@ a record: archive it to `docs/history/plans/` following the procedure in
   sequences the read layer, the calls layered on it, and the remaining
   arrivals, ordering the address change ahead of anything that publishes an
   address so no caller holds one that moves.
+- [The CLI surface — implementation plan](cli-surface-implementation.md) builds
+  the rest of the command surface: positional addresses, the honest top-level
+  names, deprecating the spellings they replace, and merging the commands that
+  do one job under two names. Separate from the read-layer plan because it
+  renames rather than adds, so its risk is what breaks for a caller who already
+  learned the current spelling.
--- a/docs/plans/cli-surface-implementation.md
+++ b/docs/plans/cli-surface-implementation.md
@@ -5,7 +5,13 @@ like and lays out seven steps to get there. Steps 1–3 are the read layer, and
 [shaped reads and verb results](shaped-reads-implementation.md) builds them.
 This plan is steps 4–7: the part that changes what a caller types.
 
-The split is deliberate. The read layer adds capability and removes nothing;
+The split is deliberate, though not absolute. The read layer is additive with
+one exception it owns and sequences itself: scoping invocation ids to a session
+makes `--invocation` without one an error, a spelling that works today and that
+[Verbs over the CLI](../common/verbs-over-the-cli.md) teaches. That is one
+change, landing alone, ahead of anything that publishes an address. This arc's
+risk is different in kind — every step of it renames or merges something a
+caller already learned;
 this arc renames, aliases and merges, so its risk is entirely in what breaks for
 someone who already learned the current spelling. The two want different
 sequencing, different tests, and different appetites for landing quickly.
@@ -123,7 +129,8 @@ Each of these is a decision, not a refactor. Sizes are for the code once the
 decision is made; the decision is the expensive part.
 
 **M1. Two `inspect`s.** `cf inspect piece` and `cf piece inspect`. Top-level
-`inspect` is offline forensics over stored state, twenty-two subcommands wide.
+`inspect` is forensics over stored state, twenty-two subcommands wide and
+offline but for `inspect pull`, which fetches from a remote.
 `piece inspect` reports on a live piece. Whether these are the same operation is
 the question — if they are not, the fix is naming rather than merging.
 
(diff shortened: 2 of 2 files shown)


mpsalisbury (commented):

**Scope:** `cli-surface-implementation.md`, read against the shape document (#5316) and the sibling plan (#5435). **Coverage:** deep-read; the accounting verified against `packages/cli` by a read-only search agent. **Checks:** no CI runs on this PR (stacked base) — the local check results in the description are the only gate until restack.

## The accounting is exact

*Verified against the CLI:* `cf piece` registers exactly the twenty subcommands listed, in the order listed (`packages/cli/commands/piece.ts`); top-level `inspect` has exactly 22; and exactly twenty files under `docs/` and `skills/` teach `--piece` or `--input`, matching D2's count. The two decisions the plan forces rather than assumes — the both-address rule and what "carries traffic" means — are the right ones to force, and the M1 skepticism (naming, not merging, if the two `inspect`s are different operations) is well placed.

## 🟡 Improvement

**"The read layer adds capability and removes nothing" mischaracterizes the sibling plan.** #5435's S2 makes `--invocation` without a session an error — a spelling that works today and that `verbs-over-the-cli.md` teaches stops working — and relocates receipt addresses. That plan owns the break honestly in its Risks; this document's framing contradicts it, and the additive/breaking contrast is the load-bearing argument for keeping the two plans separate. One sentence fixes it: the read layer's one breaking change is owned and sequenced there, and this arc's risk profile is still different in kind. *Verified* (both texts).

## ⚪ Nits

- The shape document's purpose table (and M1's framing here) files all 22 `inspect` subcommands under "offline forensics" — `inspect pull` fetches from a remote. Marginal, but M1 leans on the offline/live distinction, so worth a word when that merge is judged.
- D1's exit ("the warning does not reach stdout") is the right test; consider also pinning that `--quiet` still shows it on stderr, since the CLI's convention is that `--quiet` suppresses hints but not warnings — a deprecation notice is arguably the second kind, and deciding which it is now avoids relitigating at implementation time.

## Verdict

Clean apart from the one-sentence mischaracterization. The staging (additive → honest names → deprecation → merges), the characterization-tests-before-each-merge rule, and the refusal to treat "carries traffic" as evaluable without a signal are all sound.

mpsalisbury (commented):
cf-review follow-up via Claude Fable 5, on behalf of @mpsalisbury — re-review of the response commit, plus one direct fix.

The "removes nothing" mischaracterization is fixed in substance: the intro now owns S2 as the read layer's one exception, sequenced there. The rewrite left an edit artifact, though — the replaced sentence survived beside its replacement, and the new clause ("every step of it renames or merges something") contradicted "**Additive until step 6**" two paragraphs down. **Fixed directly in `ed09f26`**: one sentence, scoped to the arc rather than its steps.

One accounting item remains for the author:

**The "one exception" ledger doesn't count F3.** The sibling plan's new F3 turns unrecognized `--schema` keys from silently ignored into refused by name — a second, smaller behavior change to an existing flag. It is defensible (a schema carrying a stray key was already silently misbehaving, and refusal is the fix for that), but this document's claim that the read layer is "additive with one exception" now undercounts. Either count it, or say in a clause why a refusal of a key that previously did nothing doesn't belong in the ledger.

PR #5504, opened 2026-08-07
feat(cli): `@` marks a position for its address in a field list

mpsalisbury:
## Why

`--select topic@` gets you the topic's address instead of its contents, in the
concise language, without writing a JSON Schema for it. A caller reaching for a
field list should not have to switch languages to ask where something lives.

Three commits: the suffix, a bare `@` for the read's own source, and a trailing
`@` on an array asking each element.

Stage A3 of the shaped-reads plan (#5435). Stacked on #5500.

## What `@` means

A trailing `@` on a path segment marks that position for its address. It
desugars in `conciseProjectionSchema` — the one place a field list becomes a
schema — and lands on the marker machinery that already exists, so the rejecting
selector, the address rendering and the `--filter` refusal are all reached
through code that was already there.

- `topic@` — the topic's address instead of its contents
- `topic@,topic.title` — one position asked two questions; returns both
- `notes@` — **one address per element**, not the array position's own address
- `@` — the read's own source
- `user@home` — an ordinary field name; `@` is only special as the final
  character of a segment, and `\@` escapes it

## Why `notes@` is per-element

The array position's own address is the source address plus the path the caller
just typed — no information they did not have. The element addresses are the
element documents, which is the point of asking.

It is also the rule rather than the exception: a field list is already
element-wise wherever it crosses an array (`notes.title` selects title from
every note), and `notes.title@` already yielded one address per element. This
makes `notes@` the concise spelling of the JSON items marker, asserted directly
against it — and against a concrete expected value first, so the comparison
cannot pass vacuously.

The JSON `{"type":"array","items":…}` spelling still names the array's own
position when the marker sits there, and a guard test pins that boundary.

## One consequence worth knowing

By the same rule, `--select '@'` against a source that *is* an array yields one
address per element rather than the array's own address. Both behaviors are
tested and documented.

## No second path — asserted, not claimed

`parseSelectProjection("topic@")` produces byte-identical `schema` **and**
`markers` to the JSON marker form; `parseSelectProjection("@")` likewise matches
`--schema '{"$link":true}'`. If the spellings ever diverge, those fail.

Fetch suppression holds across all three commits: `--select 'notes@'` observes
exactly one document (the board), zero note syncs — measured, not assumed.

## Flagged, not fixed

Calling `deriveSelectedValue` twice with an identical source *and* identical
projection string in one runtime throws `updated arguments do not match the
candidate schema` — the transform result cell is keyed by
`{source, filter, schema}`, so the second call collides with the first. Not
reachable from `cf piece get`, which runs once per process, but it would be
reachable from any long-lived host that serves two identical reads.

Also: `--select '@'` composes as a suffix does one level down, order-independent
(`@,label` and `label,@` agree).

## Test plan

`deno task test` in packages/cli (1672 + 123 passed) and packages/runner (1156
`check-no-waitfor`, `check-conflict-markers` — all pass.

Test files are additions-only. Every new behavior test was confirmed failing
before its change; the two guard tests are noted as passing on both sides by
design.

**Not exercised:** two lines in `verbs-over-the-cli.sh` need a live toolshed.

### DIFF
--- a/docs/common/verbs-over-the-cli.md
+++ b/docs/common/verbs-over-the-cli.md
@@ -325,9 +325,10 @@ that can go stale.
 ### Asking a read for an address
 
 A cell is a point in a graph, so a read that does not say where to stop follows
-links onward and hands back a copy of everything behind them. `--schema` says
-where to stop: a position marked `{"$link": true}` returns that position's
-address rather than its contents.
+links onward and hands back a copy of everything behind them. A projection says
+where to stop: a marked position returns that position's address rather than
+its contents. A field list marks with a trailing `@`, and a JSON Schema marks
+with `{"$link": true}`.
 
 ```bash
 cf piece get --piece <board> notes \
@@ -361,6 +362,19 @@ cf piece get --piece <board> notes --schema \
 ```json
 [{ "$link": { "id": "of:fid1:…", "…": "…" }, "title": "First note" }]
 ```
+
+A field list unions the same way, and its two paths meet at the one position:
+
+```bash
+cf piece get --piece <board> --select 'notes@,noteCount'
+```
+
+```json
+{ "notes": { "$link": { "id": "of:fid1:…", "…": "…" } }, "noteCount": 3 }
+```
+
+The suffix is special only at the end of a segment, and `\@` writes a literal
+one, so a field named `user@home` stays reachable.
 
 A marked position is not fetched — the address is stored in the document that
 contains it, so a marked collection of a hundred notes costs the one read that
--- a/docs/common/verbs-over-the-cli.md
+++ b/docs/common/verbs-over-the-cli.md
@@ -376,6 +376,11 @@ cf piece get --piece <board> --select 'notes@,noteCount'
 The suffix is special only at the end of a segment, and `\@` writes a literal
 one, so a field named `user@home` stays reachable.
 
+A path that is only `@` names the position the read is already at, which no
+field path reaches because it sits above every field. `--select '@'` returns
+the source's own address in place of its contents, and `--select '@,title'`
+returns the address beside the title.
+
 A marked position is not fetched — the address is stored in the document that
 contains it, so a marked collection of a hundred notes costs the one read that
 document already needed. Where the marker is the whole selection, nothing
--- a/docs/common/verbs-over-the-cli.md
+++ b/docs/common/verbs-over-the-cli.md
@@ -370,16 +370,23 @@ cf piece get --piece <board> --select 'notes@,noteCount'
 ```
 
 ```json
-{ "notes": { "$link": { "id": "of:fid1:…", "…": "…" } }, "noteCount": 3 }
+{ "notes": [{ "$link": { "id": "of:fid1:…", "…": "…" } }], "noteCount": 3 }
 ```
+
+A field list applies to each element wherever it crosses an array, and an
+address is one of the things it applies: `notes@` answers with the address of
+each note rather than of the slot it sits in, which makes it the concise
+spelling of the `items` marker above. Where the marked position holds anything
+else, the address is that position's own.
 
 The suffix is special only at the end of a segment, and `\@` writes a literal
 one, so a field named `user@home` stays reachable.
 
 A path that is only `@` names the position the read is already at, which no
 field path reaches because it sits above every field. `--select '@'` returns
-the source's own address in place of its contents, and `--select '@,title'`
-returns the address beside the title.
+the source's own address in place of its contents — one per element where the
+source is an array — and `--select '@,title'` returns the address beside the
+title.
 
 A marked position is not fetched — the address is stored in the document that
 contains it, so a marked collection of a hundred notes costs the one read that
--- a/docs/common/verbs-over-the-cli.md
+++ b/docs/common/verbs-over-the-cli.md
@@ -363,10 +363,11 @@ cf piece get --piece <board> notes --schema \
 [{ "$link": { "id": "of:fid1:…", "…": "…" }, "title": "First note" }]
 ```
 
-A field list unions the same way, and its two paths meet at the one position:
+A field list unions the same way, and its two paths meet at the one position.
+`noteCount` is computed, so the read steps the piece to bring it up to date:
 
 ```bash
-cf piece get --piece <board> --select 'notes@,noteCount'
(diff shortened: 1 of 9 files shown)


mpsalisbury (commented):
_cf-review via Claude Fable 5, on behalf of @mpsalisbury. Reviewed this PR's own delta vs `impl/f2-select-flag` (stacked on #5500)._

# cf-review — PR #5504 `@` marks a position for its address in a field list

**Scope:** the 9-file stacked delta vs `impl/f2-select-flag` · **Coverage:** deep-read (full diff, head `cell-selection.ts` desugar + walk machinery, `lib/piece.ts` call path, both test files, all four docs) · **Checks run:** 3 targeted probe tests against the head branch in a throwaway worktree (removed after); no full suites.

The core mechanism is sound and unusually well-tested: the desugar is genuinely one path onto the existing marker machinery, the parser edge cases are clean, and the fetch-suppression test is a real A/B. One behavior contradicts the feature's own documentation in a reachable topology.

## 🔴 Blocking (resolve or consciously accept)

**1. `notes@` ≡ items-marker equivalence breaks when the marked slot stores a link to the array — and the docs state the equivalence unconditionally.** — **verified by running it.**
- *Location:* `packages/cli/lib/cell-selection.ts` `composeLinkAddresses` (the `elements = position.stored?.value` gate, ~line 1771) vs. `packages/cli/README.md`, `skills/cf/SKILL.md`, `docs/common/verbs-over-the-cli.md`.
- *What:* I seeded a board whose `notes` slot stores a **link to** a collection cell (rather than an inline array of element links, the only topology the tests cover). Result: `--select 'notes@'` → `{"notes": {"$link": <collection cell address>}}` — one address; `--schema '{"properties":{"notes":{"type":"array","items":{"$link":true}}}}'` → two per-element addresses. The spellings diverge exactly where the PR body says "No second path — asserted, not claimed."
- *Why it matters:* aliasing a collection through a link is a common shape in pattern results, and the README/skill say flatly "`notes@` ... is the concise spelling of `{"type":"array","items":{"$link":true}}`" / "one address per note." The PR body's headline rule ("one address per element, **not** the array position's own address") returns, in this topology, precisely the collection's own address. The code *comment* acknowledges the behavior ("below a crossed link there is nothing to look inside, and that link is the address") — so this looks deliberate — but nothing user-facing does, and no test pins it.
- *Concrete fix:* likely doc-side, not code-side — making the concise walk call `storedContainer` would restore equivalence at the cost of one fetch per marked linked container, but would literally fetch the marked position, breaking the "a marked position is never fetched" invariant as documented. So: qualify the equivalence claim in README/skill/verbs doc ("where the marked slot stores the array; where it stores a link, that link is the address — the same rule `topic@` follows"), and add the aliased-collection test pinning **both** spellings so the divergence is chosen, not discovered.

## 🟡 Improvements

**2. Collision-bug assessment: the PR's claim checks out, but file the issue now.** — mechanism **verified** (keying at `runtime.getCell(space, { pieceGetTransform: { source, filter, schema } }, ...)`, `cell-selection.ts` ~line 2077); the throw itself is author-reported.
- "Not reachable from `cf piece get`" is true: the only callers are `getCellValue`/`callable` in `packages/cli/lib/piece.ts` (one `deriveSelectedValue` per process; grep shows zero consumers outside `packages/cli`). PR 5505 does **not** make it reachable either — `piece call` derives once per process, on a per-invocation receipt cell.
- But two things make it worth an issue rather than a PR-body paragraph: (a) the entire test suite dodges it *silently* — every test in one runtime uses a distinct seed name or projection string, so nothing will ever trip it by accident; (b) the failure mode for the first long-lived adopter (fuse, a future serve mode, anything 5505's direction invites) is a hard `updated arguments do not match the candidate schema` throw on the second identical read — the most natural repeated operation such a host performs. Concrete fix candidates: mix a per-process nonce into the `pieceGetTransform` cause (the cell is session-scoped and stopped in `finally` anyway), or pin the single-shot contract in `deriveSelectedValue`'s JSDoc. Pre-existing machinery, so not blocking *this* PR.

**3. `label,label@` on a scalar drops the contents the bare path asked for — and the README sentence overpromises.** — **verified by running it:** result is `{"label": {"$link": …path:["label"]}}`, no `"Field notes"`. Consistent with the compose-level "nothing to join" rule and byte-equivalent to the JSON spelling (the desugar emits `{type:"object", additionalProperties:true}` at the scalar position; the projection just yields nothing), but the README says "Naming a position both ways ... returns the address beside the whole contents," which is only true for object positions. One qualifying clause plus a pinning test closes it.

## Verified positives (the claims I checked, so you don't have to)

- **Parser edge cases** (probe-tested + read): `a@,b` marks only `a`; `a.@`, `\@` alone, `a..b`, trailing/leading dots all refused with the path named; `a@.b` = mark `a` + select `a.b` (documented as `topic@.title`); `topic@@` marks a field literally named `topic@` — which is exactly what the documented "special only as final character" rule prescribes. A field named exactly `@` stays unreachable from `--select` (first-character rule, pre-existing) and reachable via `--schema`. Backward compatible: the old regex refused `@` anywhere, so no previously-valid `--select` string changes meaning.
- **Guard tests are non-vacuous in aggregate:** the two equivalence tests compare both `schema` and `markers` across spellings (either alone could pass vacuously on double-`undefined`), but concrete pins close the loop — `piece.test.ts` asserts `markers` literally for `topic@` through the options layer, the union test pins `{properties:{topic:{marked:true}}}`, and the `notes@` test asserts the concrete element addresses *before* the cross-spelling comparison, exactly as the PR body claims.
- **Fetch-suppression test is real:** notes are seeded *unwritten* (`seedBoard(..., false)`) so any read must sync; the marked run observed exactly `[boardUri]`, and the unmarked contrast run in the same runtime then synced all three note URIs — a genuine A/B, not an artifact of caching.
- **Refusal coverage:** `--filter` + `@` refused naming the right flag for both `--select` and `--schema`; the JSON-marker refusal message is byte-identical to before; the JSON-marker-on-array boundary (`notes` slot's own address) is pinned by its own guard test.

- The collision issue from finding 2, before the stack grows more `deriveSelectedValue` consumers.
- PR 5505 already notes the shared step's errors say "piece get" to `piece call` users — same issue could carry both.

**Verdict:** high-quality, honestly-argued change; hold for finding 1 — either qualify the equivalence claim (likely right) or change the walk, but pin the aliased-collection topology in a test either way; findings 2–3 are non-blocking.

PR #5505, opened 2026-08-07
feat(cli): `piece call` takes the same selection flags as `piece get`

mpsalisbury:
## Why

`cf piece get` can shape what it returns; `cf piece call` could not. A verb's
result and a direct read are the same operation on different cells, so a caller
who learned `--select` on one had to abandon it on the other.

Stage C2 of the shaped-reads plan (#5435). Stacked on #5500.

## What Changed

`piece call` gains `--select`, `--schema` and `--filter`, routed through the same
`deriveSelectedValue` step `piece get` uses. No second output path, and
`piece get`'s behavior is untouched.

Wired for **tools** as well as handlers, since `piece call` runs both and leaving
a tool to silently ignore `--select` is worse than either supporting or refusing
it. Same step, same helper.

## What a selection over a call actually does

**It filters after materialization. It does not narrow the fetch, and this PR
claims no narrowing.** `callable.ts` pulls the whole receipt first because it
must — the empty-record test that identifies a value-less verb needs the value in
hand. And even reordered it would not narrow: a receipt carries no declared
schema, so `selectSourceSchema(undefined, mask)` returns the permissive read and
no selector is pushed. The live-runtime tests deliberately write the receipt with
`undefined` schema so that case is the one pinned.

What the shared step *does* buy is that the read goes *through the source cell*,
so its links and Fabric metadata survive — a `$link`-marked `--schema` over a
call returns a usable address, and that is tested.

## Three decisions the plan left open

**Value-less verbs.** The selection is skipped entirely and `result` stays
absent — not `{}`, not an error. Separately, a selection that materializes
nothing over a result that *does* exist is refused, because otherwise "the verb
returned nothing" and "your projection kept nothing" would be one wire shape, and
`invocationJson`'s contract says the omission means the former.

**`--no-wait`.** Refused. A detached exit skips the readback the selection is
answered from. The existing `--show-links --no-wait` refusal generalized into a
list, so the single-flag message is byte-identical and the existing test's regex
still holds.

**`--show-links`.** Collected *after* the selection, so links annotate the value
the caller is actually holding. A projection keeps every surviving path at its
original position, so each address still names what it annotates and a dropped
path simply gets no entry — tested. A predicate does not: survivors land at
indices that are no longer theirs, so `--filter --show-links` is refused, on the
same grounds `--filter` + a `$link` marker already is.

## Found, not fixed

**`deriveSelectedValue`'s failure messages say "piece get" to a `piece call`
user** — `Could not apply piece get transform: …`, plus a `pieceGetTransform`
cause key and a host-trust reason naming `piece get`. Fixing the message moves
two pre-existing expectations; renaming the cause key changes the derived id of
`piece get`'s transform result cell. Both out of scope here, but worth a
follow-up now that two commands share the step.

**A new import cycle** `callable.ts` → `cell-selection.ts` → `callable.ts`, the
latter only for `runtimeErrorLog`. Safe — both bindings are hoisted declarations
used at call time — and the package already has 17 cycles. The clean fix is
relocating `runtimeErrorLog` out of `callable.ts`, which touches five files
including two the branches beneath wrote. Say if you'd rather have the churn.

**`main-command.test.ts` fails whenever `FORCE_COLOR` is set** — pre-existing,
verified identical on a stashed unmodified tree. It asserts uncolored Cliffy help
while the CLI honors `FORCE_COLOR` exactly as the README documents, so the test
looks wrong rather than the CLI. All counts below are from runs with it unset.

## Test plan

`deno task test` in packages/cli (1674 + 123 passed) and packages/runner (1156
`check-no-waitfor`, `check-conflict-markers`, and the dependency gates — all
pass.

14 tests added, none deleted. `piece-get-transform.test.ts` is untouched — not in
the diff at all.

**Not exercised:** no live server, so the walkthrough script and the real `cf`
binary against a toolshed were never run, and no step was added to that script —
an integration step I could not execute is worse than none. What was exercised
end to end: `executeResolvedCallable` against a real `Runtime` with emulated
storage and a real committed receipt document, covering `--select`, `--filter` +
`--select`, a `$link`-marked `--schema`, and the materializes-nothing refusal.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5505?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/common/verbs-over-the-cli.md
+++ b/docs/common/verbs-over-the-cli.md
@@ -104,6 +104,36 @@ Anything the *pattern* resolved comes back too. A verb that stamps a write time
 or derives structured authorship from the event returns those in its record;
 the caller could not have computed them.
 
+### Asking for a smaller result
+
+A verb decides what it returns; the caller decides how much of it to look at.
+`--filter`, `--select`, and `--schema` — the same three flags `cf piece get`
+takes, with the same grammar — shape the `result` before it reaches stdout, and
+go before the callable name:
+
+```bash
+cf piece call --piece <topic> --select comment.writtenAt addComment \
+  '{"body":"first","agentName":"Sol"}'
+```
+
+```json
+{
+  "invocation": "0f4c…",
+  "status": "settled",
+  "result": { "comment": { "writtenAt": "2026-08-07T09:15:27Z" } }
+}
+```
+
+This shapes a result that already exists rather than deciding what travels: the
+readback has the whole receipt in hand before the selection runs. So a
+value-less verb keeps reporting no `result` at all — there is nothing for a
+selection to be about — and `--no-wait`, which never reads the receipt back,
+refuses all three flags. `--show-links` composes with a projection, because a
+projection leaves every surviving path where it was; it does not compose with
+`--filter`, which moves the positions a link names.
+
+`packages/cli/README.md` has the grammar and the supported schema subset.
+
 ### Retries are safe, and cheap to reason about
 
 `--invocation <id>` makes a call idempotent. Replaying a settled id returns the
@@ -254,6 +284,12 @@ one, so a chain of references annotates each hop exactly once. The walk stops at
 any non-plain object: a live runtime object reached through a result gets its
 own entry and nothing below it, because its properties belong to the runtime
 rather than to the result.
+
+The links describe whatever result you were handed, so a projection composes
+with them: a path `--select` or `--schema` dropped simply gets no entry.
+`--filter` is refused alongside `--show-links` — a predicate leaves the elements
+it keeps at positions that are no longer the ones they came from, and every
+address below a filtered array would name the wrong element.
 
 A pattern should not mint identifier fields to make any of this easier —
 rendering identity is the client's job, and a pattern-authored fid is a copy
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -120,7 +120,7 @@ standard error and continues searching that piece and the rest of the space.
 - The launcher spawns the child CLI with `deno run --quiet` so Deno's own
   warnings (npm "Ignored build scripts" banner) never reach users.
 
-### Transforming `piece get` output
+### Transforming `piece get` and `piece call` output
 
 `piece get` can filter an array before it reaches stdout and project the result
 to a smaller shape:
@@ -131,6 +131,21 @@ cf piece get --piece ID items \
   --filter '.status == "open" and .score >= 10' \
   --select id,title,author.name
 ```
+
+`piece call` takes the same three flags, with the same grammar, the same
+conflict rule, and the same error messages, written before the callable name:
+
+```bash
+cf piece call --piece ID --select topic.title addTopic '{"title":"Ship it"}'
+cf piece call --piece ID --filter '.status == "open"' listTopics
+```
+
+Everything below describes both commands. The one difference is what the
+selection is about: `piece get` shapes a cell's value, and `piece call` shapes
+the **result of the call** — a handler's `result` inside the Invocation JSON, or
+a tool's JSON on stdout. See
+[what a selection means for a call](#what-a-selection-means-for-a-call) for the
+three cases where that difference shows.
 
 `--filter` is jq-inspired rather than a full jq interpreter. It applies only to
 arrays and accepts value paths (`.status`, `.author.name`, `.["display-name"]`,
@@ -146,7 +161,7 @@ Two flags project, one per input language:
   also accepts the same field list `--select` takes.
 
 A command that names both has not said which shape it wants, and is refused
-before the read.
+before the read or the call.
 
 For an array result, the field list describes each item. An inline/file JSON
 Schema describes the complete returned value, so a schema combined with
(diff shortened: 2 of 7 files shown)


mpsalisbury (commented):
_cf-review via Claude Fable 5, on behalf of @mpsalisbury. Reviewed this PR's own delta vs `impl/f2-select-flag` (stacked on #5500)._

## cf-review: PR #5505 — `piece call` selection flags

**Scope:** 6 files, +789/−14 (stacked on #5500; delta only) · **Coverage:** deep-read of all changed files plus `cell-selection.ts`, `lib/piece.ts`, `lib/callable-command.ts` at head; callers of `executeResolvedCallable`/`deriveSelectedValue` swept repo-wide at head · **Checks run:** none executed — every finding verified by reading the head sources; author's reported green runs not re-run.

### 🔴 Blocking

**1. Refusal-matrix gap: `--filter` + `$link`-marked `--schema` is refused only after the verb has executed and committed.** — verified
- Location: `packages/cli/commands/piece.ts` `parsePieceCallSelection` vs `packages/cli/lib/cell-selection.ts` `deriveSelectedValue` (the `selection.filter !== undefined && markers !== undefined` throw at its top).
- What: `parsePieceCallSelection` refuses `--filter --show-links` pre-dispatch, but the equivalent `--filter` + marker refusal lives only inside `deriveSelectedValue` — which runs during the receipt readback, after the handler committed (or the tool ran, LLM calls and all). `cf piece call --filter '.x == 1' --schema '{"properties":{"t":{"$link":true}}}' verb …` executes the verb, commits, then exits 1 with a `CellSelectionError`. Both inputs are fully known at parse time (`selection.projection.markers` is on the parsed projection), so this is the one statically-refusable combination that escapes the pre-dispatch net — and the PR body describes the two refusals as the same rule ("on the same grounds `--filter` + a `$link` marker already is"), which for `piece get` is harmless post-parse but for a call means a committed side effect behind an avoidable refusal. The mitigation (stderr `invocation: <id> phase: readback`, same-id replay deduplicates) works, but the PR's own stated principle is "refuse before the call".
- Fix: in `parsePieceCallSelection`, also throw when `selection?.filter !== undefined && selection.projection?.markers !== undefined`, mirroring `deriveSelectedValue`'s message; add the test beside the `--filter`/`--show-links` one. One conditional plus a test.

**2. The import cycle — decide it consciously now, since the author asked.** — verified
- `callable.ts` (lines 20–24) → `cell-selection.ts` (line 22, `runtimeErrorLog`) → `callable.ts`. Mechanically safe, confirmed: `runtimeErrorLog` and `deriveSelectedValue` are both hoisted `function` declarations; `CellSelectionError` (a class, TDZ-sensitive) is only referenced at call time; either module-evaluation order works. But `docs/development/DEVELOPMENT.md` §"Keep the Module Graph clean" lists "Modules depending on each other" under ❌ Avoid, so this is a documented-preference violation, disclosed in the PR body with the fix offered. Recommendation: take the offered relocation — a leaf `lib/runtime-error-log.ts` holding `CF_RUNTIME_ERROR_LOG` + `CliRuntimeErrorRecord` + `runtimeErrorLog` also removes the pre-existing base-branch edge, and five touched files is cheap next to an 18th cycle. Acceptable to defer to an immediate follow-up if the stack makes the churn painful — but decide, don't drift.

### 🟡 Improvements

**3. Post-settlement selection failure should say the invocation settled.** — verified
- `selectCallResult`'s error ("…inspect the result and the selection") fires after the receipt exists; the catch in the command action prints `invocation: <id> phase: readback` on stderr, and same-id replay recovers — but only for a caller who knows the phase-line protocol. Emitting settled-without-result on stdout would lie (omitted `result` means value-less — the very distinction this PR defends), so the right fix is in the message: append "the invocation settled; re-run with --invocation <id> and a corrected (or no) selection to read the result back." Cheap, and it converts an ambiguous failure into a documented recovery.

**4. "piece get" wording leak** — verified, acknowledged in PR body. `Could not apply piece get transform: …` (two sites), the host-trust reason `"cf piece get filter/schema computed projection"`, and the `pieceGetTransform` cause key all now surface to `piece call` users on transform failures. Follow-up as the body proposes: neutralize the two message strings and the trust reason; keep the `pieceGetTransform` cause key (renaming changes the derived transform-cell id). Fine to land as-is.

**5. Cross-PR double-derive hazard (#5504): not reachable from this PR today — verified, but worth pinning.** One `deriveSelectedValue` per process per call (handler and tool branches are mutually exclusive; `executePieceCallable` → `executeCallableCommand` calls `executeResolvedCallable` exactly once; no retry loop — `boundedSettlement` races, never re-invokes). Repo-wide grep at head: no toolshed / background-piece-service / fuse import of `lib/callable.ts` or `deriveSelectedValue`; the CLI is process-per-command. The trap stays latent: the transform cell is session-scoped and keyed `{source, filter, schema}`, so any future long-lived embedder replaying a same-id call with the same selection in one runtime hits #5504's throw. When #5504's fix lands, a test covering "same receipt + same selection twice in one runtime" would close this for calls too.

### ⚪ Nits
- `READBACK_FLAGS` JSDoc says "in the order a caller writes them" — the refusal lists fixed registry order, not CLI order. Reword.
- `$link`-marker `--schema` + `--show-links` is allowed and untested — both address surfaces at once; the links walk over marker-shaped values (`cell.key("$link")`) survives via its try/catch, but a pinning test would be kind.

### Verified sound (the focus areas that checked out)
- **Refusal ordering, both paths:** `resolveWaitControl` (line 1720, before the try) and `parsePieceCallSelection` (line 1728) both run before `executePieceCallable`; handler and tool share the one action, and no other caller passes `selection`. The `--no-wait` single-flag message is byte-identical to the old one, as claimed.
- **Emptiness test:** the refusal triggers on `selected === undefined` only. A projection that keeps nothing but materializes `{}`/`[]` returns it as `result` — so absent result (empty-witness receipt, tested), kept-nothing `{}`, and the refusal are three distinct wire states, exactly as the body claims. (A verb whose genuine result is `{}` reads as value-less — pre-existing empty-witness semantics, not this PR.)
- **The comment claim "`cf piece get` refuses the same condition"** is accurate (`lib/piece.ts` `getCellValue`, `selected === undefined && !sourceWasAbsent`).
- **Test rigor:** each test names its principle; the recording-selector seam pins which cell/space/selection the step receives without re-testing the step; the live `settleWith` writes the receipt with an explicitly `undefined` schema, so the schema-less-receipt case is genuinely pinned; harness options (`receiptCell`, `toolResult`) pre-exist. Gap: tool-path selection is mock-only; live coverage is handler-only.
- **Docs coherence:** README anchor, verbs-over-the-cli, and skills/cf all match the implemented semantics; `docs/plans/verb-result-selection.md` stays consistent ("failures inside a shaped read report under the read layer's rules").

### Possible follow-ups (for a human to file)
1. Neutralize the three "piece get" strings in `deriveSelectedValue` (keep the cause key) — the PR body's own proposal.
2. Relocate `runtimeErrorLog` to a leaf module (if not taken in this PR).
3. After #5504 lands: an in-runtime double-derive test over one receipt+selection.
4. `main-command.test.ts` vs `FORCE_COLOR` — pre-existing per the body; the test asserts what the CLI documents against.

**Verdict:** Sound, unusually well-argued work with real live-runtime coverage — fix the one refusal-matrix hole (`--filter` + `$link` marker refused only post-commit) and make the cycle decision explicit, then this is mergeable.

PR #5508, opened 2026-08-07
fix(runner): un-gate the nested-piece cold-start setup repair

mathpirate:
## Why

The #4977 repair—re-running a pinned pattern's own setup when instantiation hits `Handler used as lift … marker was never written`—was gated on the `systemPatternAutoUpdate` experimental flag. That flag governs the **update** family, which moves the durable identity pointer; this is a same-identity self-**repair**, matching the root controller's cold-start repair posture.

Hosts that do not explicitly enable the flag—including the raw Runtime and server-side defaults—otherwise leave an aged nested piece with no repair path. The motivating failure was observed in Loom's native PatternHost, whose `RuntimeInternals.create()` call supplied no experimental flags, so `systemPatternAutoUpdate` was off there. The shell is distinct and explicitly defaults the flag on.

The repair is safe independent of update posture: it fires only on exactly the missing-stream-marker failure, re-reads the pinned identity as a precondition inside the repair transaction, never touches the piece's argument/data, and rethrows the original error on any failure. The CT-1923 pointer rollback, which does move identity, keeps its flag gate.

cc @bfollington—this is a posture proposal on your #4977; happy to adjust if the gate was deliberate for a reason the comments do not record.

## What Changed

- `runner.ts`: drop the `systemPatternAutoUpdate` clause from `instantiateInitialPattern`'s repair condition; the comment records the repair-versus-update invariant without making host-default claims.
- `nested-piece-setup-repair.test.ts`: flip the flag-off case from a locked-in failure to a successful end-to-end repair, while proving the durable pattern identity remains unchanged.
- `runtime-processor.ts`: remove a matching overstatement about deployment defaults from an existing comment; no behavior changes there.

## Test plan

- The flipped test was validated as an instrument: it fails on unmodified `main` and passes with this change.
- Focused nested-piece setup-repair test: passed.
- Full `packages/runtime-client` suite: 60 tests / 208 steps passed.
- Repo-wide `deno fmt --check` and `deno lint`: passed.
- Full `packages/runner` suite on the behavioral commit: 1157 passed, 0 failed.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5508?utm_source=github" target="_blank" rel="noopener noreferrer nofollow"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -2739,13 +2739,9 @@ export class Runner {
     // repair the home ROOT got in startEnsuredDefaultPattern, reachable at last
     // for the nested pieces that never pass through it.
     //
-    // Not gated on systemPatternAutoUpdate: that flag governs the UPDATE
-    // family (heals that move the durable identity pointer), while this is a
-    // same-identity self-REPAIR — the same posture as the root's
-    // controller-side cold-start repair (pinned at flag-off by
-    // check-update-default-pattern.test.ts). Default deployments leave the
-    // flag unset, and an aged nested piece (e.g. a profile surface reached
-    // through a wish) has no other heal there. On EXACTLY that
+    // The repair is not gated by `systemPatternAutoUpdate`: that flag governs
+    // updates which move the durable identity pointer, while this setup replays
+    // the pattern the pointer already names. On exactly that
     // failure, re-run the pinned pattern's OWN setup state (samePattern=true:
     // materializes the missing internal cells but leaves the existing argument
     // — the piece's data — untouched; no roll-forward, no user-data rewrite),
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -2739,7 +2739,13 @@ export class Runner {
     // repair the home ROOT got in startEnsuredDefaultPattern, reachable at last
     // for the nested pieces that never pass through it.
     //
-    // Gated on the systemPatternAutoUpdate experimental flag. On EXACTLY that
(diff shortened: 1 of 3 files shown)


bfollington (approved):
cf-review via Codex (GPT-5), on behalf of @bfollington.

Approving the behavioral change—the repair/update distinction makes sense to me, and this is a suitably narrow repair of the already-pinned identity.

I would like the new comments to be corrected before merge, though. “Default deployments leave the flag unset” and “every default deployment” are broader than the facts support: the raw Runtime and server-side defaults leave systemPatternAutoUpdate off, but the shell explicitly defaults it on.

Perhaps phrase this as:

> Hosts that do not explicitly enable the flag—including the server-side defaults—otherwise leave an aged nested piece with no repair path.

Since the native-profile incident is cited as motivation, it would also be useful to clarify which host/runtime had the flag off. If that incident occurred in the default shell configuration, this gating would not have been its cause.

No concerns with the code change itself.

mathpirate:
Thanks—good catch. The affected host was Loom’s native PatternHost: its `RuntimeInternals.create()` call omitted experimental flags, so `systemPatternAutoUpdate` was off there; the shell’s explicit default-on posture is distinct. I narrowed the PR description and code/test comments accordingly, and removed the same overstatement from the existing runtime-client comment. No behavioral changes in this follow-up.

PR #5515, opened 2026-08-07
fix(cli): make profiler startup transactional

mathpirate:
## Summary

- enable, configure, and start the CPU profiler inside one guarded inspector transaction
- perform that sequence at the actual start point, including marker-triggered starts after the debugger resumes
- add a deterministic regression that invalidates profiler state after resume and requires the full `enable → configure → start` ordering

## Root cause

The capture connected and enabled the profiler domain during setup, then resumed the target before a later console marker triggered `Profiler.start`. That left a gap in which the start could observe the profiler domain as disabled, producing the CI failure `Profiler is not enabled (-32000)`.

This closes the state gap rather than retrying the failed command: every guarded start now establishes the protocol preconditions immediately before `Profiler.start`.

Failing run: https://github.com/commontoolsinc/labs/actions/runs/31210927383/job/92973283497

## Validation

- full CLI package suite on Deno 2.9.4: 1,808 passed / 792 steps / 0 failed
- focused profiler suite: 7 passed / 56 steps / 0 failed
- exact real-inspector regression: 20/20 fresh sessions
- `deno fmt --check`: 4,330 files
- `deno lint`: 4,312 files
- pinned `deno task check`: 443 paths
- no new polling `waitFor`; no conflict markers

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5515?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cli/support/profiling/capture-deno-inspector-profile-lib.test.ts
+++ b/packages/cli/support/profiling/capture-deno-inspector-profile-lib.test.ts
@@ -33,6 +33,14 @@ function createState(): ProfileCaptureState {
   };
 }
 
+function createProfilerStarter(start: () => Promise<void>) {
+  return {
+    enable: () => Promise.resolve(),
+    setSamplingInterval: (_params: { interval: number }) => Promise.resolve(),
+    start,
+  };
+}
+
 class FakeWebSocket {
   readyState: number = WebSocket.CONNECTING;
   added: string[] = [];
@@ -120,10 +128,12 @@ class FakeCelestial {
       });
     },
     setSamplingInterval: (params: { interval: number }) => {
+      this.calls.push("Profiler.setSamplingInterval");
       this.samplingIntervals.push(params.interval);
       return Promise.resolve();
     },
     start: () => {
+      this.calls.push("Profiler.start");
       this.starts += 1;
       // What the inspector answers a `Profiler.start` it has not enabled yet.
       if (!this.profilerEnabled) {
@@ -133,6 +143,7 @@ class FakeCelestial {
       return this.startPromise ?? Promise.resolve();
     },
     stop: () => {
+      this.calls.push("Profiler.stop");
       this.stops += 1;
       return Promise.resolve({ profile: this.profile });
     },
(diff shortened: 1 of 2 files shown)


mpsalisbury (approved):
LGTM. Some nits from claude:

1. Pre-existing dropped-stop race, now landing in a much more reachable window — non-blocking, verified

A live profile stop marker that arrives inside the enable → setSamplingInterval → start transaction is silently discarded when a stale stop preceded the start marker. Path: marker sets clearStop: profileMessage.hadProfileStop → true; the live stop sets state.sawProfileStop = true but handleConsoleApiCalled's stop check is gated on state.profilerActive, which is still false; then markProfilerStarted(state, {clearStop: true}) wipes sawProfileStop, so startProfiler's started && state.sawProfileStop tail never fires. The capture then runs until the summary pattern or a websocket close.

I confirmed this with a probe test (stale stop → start marker → live stop while Profiler.enable is gated): profile: profile stop matched never logs. It fails identically against main — so the PR does not introduce it and I'm not asking for a fix here.

What the PR does change is reachability. On main that window sat before Runtime.runIfWaitingForDebugger; now it is three inspector round trips after the target has resumed, i.e. exactly when live console output is flowing. Existing coverage only exercises the clearStop: false variant (stops from a marker that arrived while the profiler was still enabling), so the true variant is untested in both versions.

Minimal fix if you want it in-scope: clear the stale latch at the top of the transaction rather than at the end — if (options.clearStop) state.sawProfileStop = false before the step loop, then markProfilerStarted(state, {clearStop: false}). A stop that arrives mid-transaction then survives.

2. The revert-detection failure is opaque — nit

Under a reverted lib, enables the profiler as part of a marker-triggered start fails with Promise resolution is still pending but the event loop has already resolved, not with the Profiler is not enabled (-32000) the fake actually raised — because the test awaits started.promise and the run has already bailed via recordProfilerStartError. It fails fast, so this isn't a hang, but whoever hits it next won't learn why from the message. await Promise.race([started.promise, done]) — or asserting await done === 0 first — would surface the real error.

3. recordProfilerStartError sniffs the message prefix — nit

message.startsWith("Profiler.start failed") couples the caller to the exact wording startProfilerIfReady produces. The wrapper already sets { cause: error }; keying off that, or a small exported ProfilerStartError class, wouldn't drift. Related behavior nit: the switch to error.message means an unwrapped error now renders Profiler.start failed: closed where it used to be Profiler.start failed: Error: closed. Nothing consumes that difference.

4. The window is narrowed, not closed — observation

If the root cause is genuinely the domain being disabled asynchronously after setup, it can still be disabled between the Profiler.enable reply and the Profiler.start send. You call the no-retry choice out in the PR body, which I agree with — but it's worth putting in the doc comment on startProfilerIfReady, so the next person who sees -32000 reaches for a bounded retry on that error code instead of re-reordering the handshake a third time.

5. Stale comment — nit

keeps a stop seen before the start marker from ending the profile still says "the deferred start carries that answer"; deferredProfilerStartOptions is gone. The wording still half-works (the start is pending on enable), so this is cosmetic.

PR #5520, opened 2026-08-07
fix(runner): replay pending patch ops over bases; never combine values (CT-1872 1a)

ubik2:
## What

CT-1872 Class 1a: a pending patch layer now renders by **replaying its ops** over the current base — through the same `applyPatch` the server folds with — never by copying values out of its optimistic snapshot.

The old `applyPendingVersion` patch case copied values from `pending.value` (the snapshot computed over since-dropped overlays) at changed paths, branch-replacing at a blocking prefix when ancestors were missing. Every Class 1a failure mode flowed from that value-combining: **resurrection** (the snapshot embeds a dropped sibling's data — the vote-count-mismatch class), and **survivor loss** (the prefix walk giving up against incompatible bases).

Ops-replay makes both unrepresentable:

- The patch vocabulary is semantic (`append` / `add-unique` / `remove-by-value` / `increment` / `splice`), so replay **re-folds** the layer against whatever base the server delivered — an increment increments the winner's counter; an add-unique dedupes against it — exactly the fold the server performs on accept.
- Spine materialization comes from the ops that allow it (`createMissing`), mirroring the server's mergeable disposition — no rebuild machinery.
- The ops can only express the layer's own writes, so a dropped sibling's data cannot appear in the result.
- A layer whose ops cannot apply (its spine died with a dropped parent; a winner reshaped the base) renders **skipped** with a debug log: transiently honest, converging when a frame delivers server truth. Under strict semantics (CT-1875) such commits become terminal rejections at admission.

The absent-base rule is consolidated rather than mirrored: `patch.ts` owns `emptyEntityDocument()` (the engine imports it now) and exports `applyPatchToDocument(base | undefined, ops)` — the single entry point shared by server-side revision reconstruction and client-side pending replay.

Net −120 lines in the replay path: `pendingSetPathForBase`, the blocking-prefix walk, `changedPathsForPendingPatch`, `replayPathForPendingPatchTarget`, `compactChangedPaths`, and the `pending-branch-replace` warn are gone.

## Semantic flip — reviewer attention wanted here

Three pre-existing tests ("pending visibility can replace a null/scalar/array branch with an object patch") pinned the branch-replace **as behavior**: a pending vote-style patch rendered over a base a winner had nulled or reshaped. Under ops-replay those layers render skipped — the base shows through until the server resolves the commit (rejection → revert; accept → truth via frame). This is the honest render (the old one fabricated states the server would never produce), but it is the one *user-visible* change beyond the bug cases, and the rewritten pins encode it deliberately.

## Tests

The four #4608 Class 1a probes land rewritten to the ops-replay contract: no crash over dead spines, no clobbered primitives, no fabricated arrays, and the headline no-resurrection scenario completing the full arc — stale-mirror skip (`{items: null}`, no ghost) through frame-delivered convergence (`{items: {theirs, mine}}`, `seeded` never existed).

Context and analysis history: CT-1872 (the 2026-08-07 comments carry the class sweep, the T1/T2 walkthrough, and this fix design). Validation: runner 1162/0 clean exit, memory 478/0, type-check/fmt/lint green.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5520?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ubik2:
Recreating as ready-for-review (token cannot flip draft status).

PR #5522, opened 2026-08-07
feat: server-execution v2 Phase 2 — flag ON: the server derives, the client does not

seefeldb:
## Why

The first ON-arm milestone of server-execution v2 (the plan's Phase 2, merged from the old Phases 1+2 by the L14 ruling): **the SpaceServer commits derivations and the client structurally cannot** — shipped as ONE state, stacked on stage G (#5461). Client HANDLER writes stay authored-class until Phase 3 (F10). The named unskipping condition of the ON arm's only skip — "until Phase 2 removes the client derivation-commit path" — is this PR: `cfc-group-chat-demo-two-browsers` now runs and passes ON.

## What

**The client loses the derivation-commit path by construction** (`packages/runner/src/speculation/overlay-destination.ts`): under the flag, every non-serving runtime's DEFAULT seal destination redirects stamped derivation-kind runs into the replica's pending overlay (`sealNative` speculative — outside the `synced()` barrier; no code path from a derivation run to the wire exists). Handler runs and unstamped/binding writes commit as today; serving runtimes are `servingPosture`-marked at construction (the SpaceServer refuses activation without it — pinned negative). Retirement is watermark-driven per speculation.md §4 (entry floor = confirmed read basis + acked origins via a bounded replica ack record; success-shaped `superseded` withdrawals that never cascade; chained entries re-sweep on settlement). Egress effect kinds are owned-and-dropped at the destination (navigateTo allowlisted — reversible; `compile-and-run` gated at the builtin — its floating launch bypasses the destination). The scheduler-tell ruling (protocol §1, owner 2026-08-07) later NAMED this exact boundary as the ruled classification — no code change was needed.

**The settle input barrier** (Phase 2 revisit (a), RESOLVED by the plan's sanctioned exclusion): `ISpaceReplica.unappliedForeignSeqFloor` + the SpaceServer's W-advance clamp (`watermarkClamped`) + the flag-gated shadow-flip notification in `confirmPending`, with three live-forced exemptions pinned (genuine-novelty-only, own-echo with the verdict-race repair, seq-0 markers).

**The fan-out seam** (M1's Phase-2 half): demand carries identity end to end (`watchedRootsForSpace` per-instance entries → the SpaceServer's demanded-identity registry), `ServerRunInfo` widened (`scopeKeyIdentity`/`actionScopeKey`) through `#stampRun`, tx-carried identity consulted at the traversal-context and result-cell sites. Cardinality 2 pinned at three levels (sink/engine same-doc fold; two stamped runs in one wave with per-run write/basis keys; the production-seam journey whose per-run outbox carriages carry two DIFFERENT keys into completions — the register's m-4 note discharged). The scheduler-side per-run SUPPLY is the **approved follow-on stage P2-F** (OW17, owner nod 2026-08-07 — recorded in the plan).

**Owner rulings landed in-flight**: the scheduler-tell classification (base commit); R1 — the loop's watermark write is a blind key-path write (no session-plane read; the owner-only-ACL noise dies, verified live); R2 — never-a-piece id classes (`computed:`/`cid:`/watermark) register no piece demand and count no deferrals (pinned; OW19 records the owed terminal-state design); R4 — `PiecePropIo.set` validates the WRITTEN PATH (the stream branch's rationale extended to property writes; correlated-anyOf paths keep the whole-result fallback; #4717's protections pinned both ways; three deliberate whole-doc pins reconciled to the narrowed contract).

**Verification surface**: `sx2-serving-loop` + `sx2-speculation` (testing.md §5), OW3 as the spec-model narrowing/fan-out family (C11a–c), OW4 as trace T15 (counts 14→15), OW12's discovered-narrowing fixture, the two-browsers skip retired.

## Verification

Full bar at HEAD (foreground): **runner 1121 (6053 steps)** · **memory 461 (145)** · **spec-model 22** · **piece 34 (390)** · skips 10 · check.sh, `deno lint` (4326), `deno fmt --check` (4356), check-docs (522), conflict-markers all green. Adversarial self-review ran (report: `/Users/berni/labs-worktrees/phase-2-review-report.md`); 1 BLOCKER + 4 MAJOR + 9 MINOR — **all addressed** (docs truthfulness reconciled; the false-shadow clamp fixed with pins; reentrancy guard; test vacuity fixes; wording precision).

Live gates (this box; loads recorded per run; fresh stores per the binding protocol; two stray-server incidents diagnosed and cleaned mid-arc — a client-bundle engine-import break was found and fixed via the OFF-arm bisect):

| gate | status |
| --- | --- |
| `counter` ON | **MET** — green repeatedly incl. post-R4 (847/117/799 ms; OFF-arm parity 854/191/808 ms) |
| `cfc-group-chat-demo-multi-runtime` ON | **MET** (Passed 1/1) |
| `cfc-group-chat-demo-two-browsers` ON (the retired skip — the named Phase 2 gate) | **MET** (Passed 1/1) |
| `sx2-speculation` ON | **MET** (Passed 1/1; the gate pre-pulls and boundedly retries on the kept destination-validator's cold-view window — Flags 2) |
| `sx2-serving-loop` ON | **flake-hit-and-escalated** — deterministic reproducer of the demand-cycle starvation fork (Flags 3); ON-skip-listed with the mechanism in its reason; runs green OFF |
| byte-identical workloads both arms | **MET** — full unit bar + counter/multi-runtime/two-browsers/sx2/lunch all green OFF, adjacent |
| sx2 settles on `waitForSettled`, no text-polling | **MET** (targets derived from the observed watermark; no-waitfor gate green) |
| one authoritative run per upstream change | **MET** — live counters: 3 authored → 3 waves → 3 derived (ratio 1.0); serving-loop unit pins |
| restart reconvergence, no duplicate effect results | **MET** (unit E2E park/reactivate + memo pins; the completion-visibility wedge is CLOSED on this base — b4c8c94bc) |
| commit-ratio ≤ 2 on lunch-poll | **machinery-complete + measurement-owed** — lunch ON red under the starvation fork (Flags 3; counters even so: 16 derived / 28 authored ≈ 0.57 ≤ 2 whole-run); lunch OFF green adjacent; the in-CI ratio gate lives in `sx2-serving-loop` (skip-listed pending Flags 3) |
| actor-side latency parity | **measurement-owed** — box load 4–20 throughout (protocol forbids quoting above ~5); the counter gate's ON step timings sat at OFF parity as context |
| zero client-committed derived writes ON | **MET** — store-attribution query pinned (holder = service identity; none from any client session) + admission by construction |

## Flags

Per flag-don't-fill — open items with their ledger locations:

1. **Service-principal read authority** (escalated fork): one `sync-load-failure AuthorizationError` per run remains for cross-space reads under the service principal (R1 removed the watermark-doc instance of the same noise class). Server-internal; needs the read-authority design.
2. **Set-validation convergence, the KEPT half** (escalated fork; item #2's ruling narrowed the schema half — R4): the write-destination validator still reads the instantaneous view, so a cold-chain first write under load can fail on the server-derived-late producer (`current producer value is not accepted as an array container`) — bisect-verified pre-R4-identical. The sx2-speculation gate pre-pulls + boundedly retries on exactly this class; converge-before-validate vs relax remains the owner's fork for this validator.
3. **Demand-cycle starvation / deferral terminal state** (escalated fork; OW19): `#loadDemandedStructure` runs before the settle race with no deadline and `of:`-class meta-less roots retry per cycle (the R2 exclusion removed the structurally-futile classes and demonstrably restored input consumption; the residual throttles cycles under load — the `sx2-serving-loop` skip and the lunch ON red are its reproducers). The terminal-state design with commit-triggered re-arm + moving the load pass under the deadline is OW19, targeted at the P2-F follow-on.
4. **`markAuthoritativeWrites`' destination gate** now means "any destination", not "serving" — latent (no client-side completion producer exists under the flag); comment-flagged at the gate for the first future producer.
5. **Overlay entry lingering** (accepted + documented): a re-speculation reading pushed derived values carries their seqs in its floor, exceeding reachable W on a then-quiet space; values converge, entries lift on the next input.
6. **OW17/P2-F is APPROVED** (owner nod 2026-08-07) as its own follow-on stage — recorded in the plan; this PR's fan-out claims are seam-level by design.

## Self-review

Adversarial subagent review of the full diff after green (spec conformance incl. the by-construction hunt, OFF-arm neutrality, barrier soundness, fan-out honesty, test vacuity, docs honesty); report at `/Users/berni/labs-worktrees/phase-2-review-report.md`. 1 BLOCKER (docs claiming a skip-listed test as verification — reconciled), 4 MAJOR (false-shadow clamp — fixed with pins; unpinned ruled behaviors — pinned or precisely re-worded; OW4 overstatement — scoped; fork ledger gap — OW19), 9 MINOR — all addressed in `d55a9b16d`. The review's confirmed-clean list covers the by-construction claim (no surviving client derivation-commit path found), OFF-arm neutrality of every touched shared file, and barrier soundness beyond the fixed finding.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5522?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


seefeldb:
## Independent adversarial review (coordinator round) — LANDABLE-WITH-FIXES → fixes landed as 6695f3d7a

Reviewed at 077ca9216 by a fresh-eyes agent under the treat-self-review-as-unverified protocol. The core of the phase held everywhere it matters:

**Verified deep and held:** the by-construction claim survived a directed hunt — one extended-transaction construction site, a write-once seal-destination pin, the commit switch consulting the destination on every commit, the overlay failing closed on transports without `sealInto`, `servingPosture` set only in the toolshed host factory with a pinned activation refusal, and no preset able to set it; the diversion probe (disable the overlay's seal handling) fails the by-construction journey, proving it load-bearing. The landed scheduler-tell text conforms to the ruling. All suite counts reproduced exactly. OW19 was assessed a genuine follow-up, not wedge-class — with the governance consequence now recorded in the plan: **P2-F is part of Phase 2's gate closure** (it carries the suspended sx2-serving-loop surface and the in-CI ratio gate), not optional hardening.

**MAJOR findings → fixes (all in 6695f3d7a, probe evidence red→green):**

- **M1** — R4's deliberate OFF-arm change (whole-result → written-subtree validation) had no recorded-acceptance row, which testing §2 as amended requires for ANY OFF-arm diff. The row now lives in a dated register delta with a cross-pointer at the stage-G tryClaimMutex row, stating the delta, its RULED status, and the five pins.
- **M2** — the own-echo verdict-race repair (the shadow mis-recorded when a frame outruns its verdict; failure mode: a quiet serving loop clamps W forever against its own echo) was probe-proven unpinned — deleting it left 77 tests green. A new interleaving test drives the race through the marker harness (frame pushed before the verdict resolves); the deletion probe now fails exactly that test.
- **M3** — the W-advance clamp's server half was untested, and `watermarkClamped` missed full-suppression clamps — precisely the total-wedge case §7's sentence exists for. The condition now counts whenever the floor held W below an otherwise-advancing batch head (code moved to the existing spec sentence; no spec change), and a SpaceServer-level clamp test pins baseline/partial/full-suppression/remove-sentinel plus the counter; both Math.min/max inversions turn it red.

**Minors → fixes:** the shadow-flip now wakes the loop (a clamped-then-quiet space no longer stalls `waitForSettled` up to the idle timer — a new `shadowFlipObserver` fired from promotion, pinned with a prompt-lift bound); compile-and-run's ON-arm gate got both-direction coverage and honest wording (inert ON-arm everywhere until the serving port — wider than the previous "not speculable"); the retirement delta's "pinned end to end" softened to OW4's honest scoping with two cheap pins added (`cid:` exclusion, shadowed-remove sentinel — the acked-origins KEEP-half pins stay owed with Phase 3's offline machinery); protocol §1's tell block gained the F10 clause (handler-kind writes divert from Phase 3, authored until then); testing §5's Phase-2 row now points the store-attribution query at its real home; the sx2 gate's cold-view retry counts loudly and a zero-retry steady-state edit makes the mask visible when it engages.

**Attestation:** the reviewer flagged that protocol §1's owner blockquote cannot be source-verified from repo artifacts. Correct — the ruling arrived through the coordination channel (owner message, 2026-08-07); the register now carries the provenance line, attested by the coordinator.

**Self-review audit:** 12 of 14 resolutions re-derived genuine; two were partially hollow (compile-and-run resolved by wording only — now tested; the exemption sub-claim overstated the verdict-race repair's coverage — now pinned).

**Counts at 6695f3d7a:** runner 1124 (6054 steps); memory 461; piece 34 (390); spec-model 22; check-docs 522. Skip list: one entry (sx2-serving-loop, phase-2-followup, closed by P2-F).

**Open item moving to Phase 3's ledger:** the delegated scoped-READS ruling (from #5461).

seefeldb (commented):
Verdict: changes required.

The current unresolved P1s remain reproducible in this exact cumulative layer: scoped `sample()/sink()` reads can lose the served run identity; the result-pattern cache resolves against the service identity; successful overlay withdrawal can leave dependent authored writes with unresolved local sequence state; the shadow floor uses the wrong extremum and lets W pass hidden earlier input; overlay close can race and resurrect entries/effects; read-only-space seal dependencies are omitted; client flatMap resume is stamped bookkeeping and can commit derivation; and an OFF serving runtime can drop LLM partials.

Separately, this PR explicitly leaves the mandatory P2-F tail unimplemented. The plan says the per-(action × instance) scheduler/read state plus demand-cycle terminal state are part of Phase 2 gate closure, and `sx2-serving-loop` remains ON-skipped. A follow-on PR may carry that work, but it must be inserted before Phase 3 rather than treating this layer as a completed Phase 2 gate.

This PR currently records only the Cubic check, not the full Actions matrix.

seefeldb:
Fix ledger for the 2026-08-12 review, the leg-C owner ruling, and all open inline threads. Pushed `ef58a7349..340851931` (parent stage-g merged first — carrying both the F 34-thread and G 19-thread rounds — 6 conflicts combined-intents, post-merge runner fully green).

**Leg-C ruling implemented (owner, 2026-08-13 — the loud-refusal minimal variant, rebase+CAS rejected).** All four parts red-first:
- 1a: the export path refuses to build a basis naming an overlay-only speculative layer — terminal `SpeculativeBasisError`, loud client-side (pre-fix wire export captured as evidence).
- 1b: convergence is now terminal-and-loud instead of infinite — one run per event (pre-fix: 17 re-runs in 5s), plus a push-boundary belt.
- 1c: the verdict-lifetime sub-defect (dep accepted server-side, pending client-side forever) fixed via an origin-ack retirement wake (`speculationAckObserver`).
- 1d: speculation spec §6 gains the process-local principle, the loud-refusal rule, and the owner's attributed blockquote; register delta recorded.

**All 32 open threads dispositioned — 30 fixed, 1 pre-resolved, 1 open by recorded decision:**
- Fixed (mostly red-first): shadow floor's wrong extremum (Set-based floor — W can no longer pass hidden earlier input), wrapper-carried run identity for scoped `sample()`/`sink()`, per-instance result-pattern memo (no longer resolves against the service identity), overlay close/seal races (entries/effects can no longer resurrect), client flatMap/map/filter resume run-kind fixed for all three list builtins via shared `resumeSettleRunKind` (serving→bookkeeping, client→derivation; seeds stay setup-class), wake races, ensure-failure classification, fan-out validation, container cardinality, and the doc/test threads.
- **r3739139506 (read-only-space seal dependencies) — attempted and reverted; OPEN by recorded decision.** A full implementation (sealSpaceReads collector, per-entry RO floors, retirement gated on each RO space's watermark) was built red-first, then reverted on design-risk grounds found during investigation: it subscribes flag-ON client runtimes to FOREIGN, possibly unauthorized, spaces' watermark docs (AuthorizationError sync-load noise observed live), and its conservative blocking can pin entries on never-covered floors forever. Stage-D's third bound stands as documented; the redesign is owed in verification-coverage's 2026-08-13 delta. Not silently dropped.

**Adjacency items from the G round:** the tryClaimMutex claim-commit-failure register row is added, and the wave-replay reachability flag is CLOSED — the spec-model's recovery arms agree with the engine (no original-basis replay path exists by design; §6 no-replay recovery recomputes at current basis).

**OFF-arm:** one ledger entry, and it is a RESTORATION — the LLM-partials fix corrects a previously unrecorded OFF-arm delta (the recorded skip is now scoped to the flag); no new acceptance needed.

**Final bar at 340851931:** runner 1162 passed (6299 steps) / 0 failed, memory 435 + the six explicit `*-test.ts` files 56/0, spec-model 23/0, piece 36/0, two full 30/30 soak rounds, wedge pins green, sx2-speculation gate green live, `deno check` clean.

**Process finding worth its own paragraph — the two-browsers gate is locally uncharacterizable.** It fails bimodally on the from-source harness INDEPENDENT of commit: the unmodified merge base failed the identical stall twice on clean reruns, and a runtime-identical twin of a passing configuration failed — invalidating an earlier bisect that had briefly implicated a real change. Two genuine local confounds were found and eliminated along the way: (1) concurrently-alive toolshed processes sharing one sqlite store across runs (split-brain engines); (2) the felt shell bakes API_URL at build time, so pairing a shell with a later toolshed port silently breaks browser login. Neither explains the clean-condition bimodality; the compiled-binary CI run is the gate's arbiter. Full 15-run evidence table in the fix report.

Next: P2-F (per-(action×instance) run supply + demand-cycle terminal state + OW19 + sx2-serving-loop unskip) launches as its own train PR per the ruled gate-closure directive.

seefeldb:

PR #5542, opened 2026-08-09
fix(identity): hold key material immutably, and own it at the edges

danfuzz:
I (agent working on behalf of @danfuzz) have rewritten this PR and its
description. It previously took a copy-everywhere approach with a documented
ownership contract; @danfuzz's read was that the cure was worse than the
disease, and he was right. This holds the key material immutably instead, so the
guarantee comes from a type rather than from asking callers to behave.

## The defect

`NobleEd25519Signer` stored the keypair object it was handed, and `serialize()`
returned that same object — so any holder of a serialized keypair could alter
what the signer signs with. `serialize()` has many callers:
`worker-controller.ts`, `runtime-client.ts`, `shell/shared/app/state.ts`,
`key-store.ts`, `integration/shell-utils.ts`, and `toPkcs8()`.

`NobleEd25519Verifier` had the matching problem on the public side: it stored
the caller's `publicKey` and derived `#did` from it, so a later mutation left it
verifying against changed material while still reporting the original DID.

It was already known at **one** consumer — `toRaw()` carried a comment saying
`serialize()` "hands back the live internal buffer" and copied to compensate.
The knowledge lived at a consumer rather than at the source, which protected
that one call site and nothing else.

## The approach

Key material is held as `FabricBytes`, which is immutable. Nothing handed in,
and nothing handed out, can alter what the signer signs with, and no contract
has to ask.

Signing and verifying take a `slice()` per call. Measured: **102 ns** against an
ed25519 operation of **128 µs** — 0.08%, ed25519 being 1251× the copy.

## Why `InsecureCryptoKeyPair` stays plain

Worth flagging, because it is the one thing that looks like an omission and is
not. Making its fields `FabricBytes` was tried and **fails**: that value travels
as a worker IPC payload, and `structuredClone` does not preserve a class. A
`FabricBytes` arrives on the far side as a plain `Object`, so
`isInsecureCryptoKeyPair()` rejects it and the worker reports `Invalid IPC
request`. `background-piece-service` and `cli` both break.

The boundary wants a plain shape; only the interior wants a rich one. The type
now says so, so the next person does not repeat the experiment.

## Also here

- `toRaw()` and `toPkcs8()` ask the implementation for the raw private key
  instead of going through `serialize()`, which built a pair object and a second
  array only to discard both. That also replaces duck-typing on the serialized
  shape with an `instanceof` on the implementation — the question the error
  messages were already asking.
- `NativeEd25519Signer.serialize()` returns a fresh frozen pair. `FabricBytes`
  cannot reach that arm, which holds opaque platform keys, but the pair object
  around them is ordinary and reassigning a member of a shared one would reach
  the signer. Verified: every write against a shared `CryptoKey` either no-ops
  or throws, so the keys themselves are safe; the pair is not.
- `Signer.serialize()` gains a doc comment. It had none, which is the root of
  all of this: its behavior had to be inferred from an implementation.

## Verification

Every guard was confirmed load-bearing by removing it and watching a test go
red — the signer's copy-in, the verifier's copy-in, `toRaw()`'s freshness, and
the native pair object. Three of those were initially **unguarded** and the
ablations are what found it.

`deno task check`, `deno fmt --check`, `deno lint`, `identity` (21 passed), and
the full workspace suite (`All tests passing!`).

— Claude

### DIFF
--- a/packages/identity/src/ed25519/index.ts
+++ b/packages/identity/src/ed25519/index.ts
@@ -65,7 +65,7 @@ export class Ed25519Signer<ID extends DIDKey> implements Signer<ID> {
   // is, which the type already answers.
   toPkcs8() {
     if (this.#impl instanceof NobleEd25519Signer) {
-      return toPEM(ed25519RawToPkcs8(this.#impl.rawPrivateKey()));
+      return toPEM(ed25519RawToPkcs8(this.#impl.privateKey().slice()));
     }
     throw new Error(
       'Cannot convert identity to PKCS8 format: requires "noble" implementation.',
@@ -77,7 +77,7 @@ export class Ed25519Signer<ID extends DIDKey> implements Signer<ID> {
   // The array is freshly allocated, so the caller owns it outright.
   toRaw(): Uint8Array {
     if (this.#impl instanceof NobleEd25519Signer) {
-      return this.#impl.rawPrivateKey();
+      return this.#impl.privateKey().slice();
     }
     throw new Error(
       'Cannot export raw key material: requires "noble" implementation.',
--- a/packages/identity/src/ed25519/index.ts
+++ b/packages/identity/src/ed25519/index.ts
@@ -58,15 +58,14 @@ export class Ed25519Signer<ID extends DIDKey> implements Signer<ID> {
     return this.#impl.sign(payload);
   }
 
-  // Only "noble" implementations can be converted to PKCS8
-  // since we need the raw material.
+  // Only "noble" implementations can be converted to PKCS8, since we need the
+  // raw material. Asks the implementation directly rather than inspecting what
+  // `serialize()` returns: that would build a pair object and a second array
+  // only to discard both, and the question here is which implementation this
+  // is, which the type already answers.
   toPkcs8() {
-    const serialized = this.serialize();
-    if (
-      "privateKey" in serialized &&
-      serialized.privateKey instanceof Uint8Array
-    ) {
-      return toPEM(ed25519RawToPkcs8(serialized.privateKey));
+    if (this.#impl instanceof NobleEd25519Signer) {
+      return toPEM(ed25519RawToPkcs8(this.#impl.rawPrivateKey()));
     }
     throw new Error(
       'Cannot convert identity to PKCS8 format: requires "noble" implementation.',
@@ -75,15 +74,10 @@ export class Ed25519Signer<ID extends DIDKey> implements Signer<ID> {
 
   // The raw 32-byte ed25519 seed. Like toPkcs8, only "noble" implementations
   // expose the private material; WebCrypto (native) hides it, so this throws.
-  // Returns a COPY — serialize() hands back the live internal buffer, and a
-  // caller mutating it would corrupt this signer.
+  // The array is freshly allocated, so the caller owns it outright.
   toRaw(): Uint8Array {
-    const serialized = this.serialize();
-    if (
-      "privateKey" in serialized &&
-      serialized.privateKey instanceof Uint8Array
-    ) {
-      return new Uint8Array(serialized.privateKey);
+    if (this.#impl instanceof NobleEd25519Signer) {
+      return this.#impl.rawPrivateKey();
     }
     throw new Error(
       'Cannot export raw key material: requires "noble" implementation.',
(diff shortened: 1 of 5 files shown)


danfuzz:
Good catch — safe today, but you are right that it should not stand.

**Why it currently works:** `serialize()` allocates fresh arrays on every call, so two `toRaw()` callers get distinct objects. Verified rather than reasoned: `a === b` is `false`, and after caller A mutates its result, caller B and a third call both still read `7`.

**Why that is not good enough:** `toRaw()` stopped defending itself. Its safety became *derived* from an implementation detail of `serialize()`, and memoizing `serialize()` is an obvious future optimization. The moment someone does it, `toRaw()` silently hands the same private key array to every caller — precisely the action at a distance this PR exists to remove, reintroduced one layer up. My own change made a local guarantee into a remote one.

Fixed in 6cd9cf4ab: `toRaw()` keeps its copy, at a cost of 32 bytes, and a test pins the property directly — two calls yield distinct arrays, and one caller's mutation does not reach another's.

I also ablated to the exact hazard you were pointing at (memoized `serialize()` **plus** a non-copying `toRaw()`) to confirm it is caught rather than merely described. Three tests go red:

```
toRaw returns a COPY — mutating it cannot corrupt the identity ... FAILED
two toRaw() callers cannot interfere with each other ... FAILED
serialize() returns a COPY — mutating it cannot corrupt the signer ... FAILED
```

— Claude

danfuzz:
Correcting my own framing above, because it had this backwards.

I described memoizing `serialize()` as an "obvious future optimization." It was **already the case**. On `main`, `serialize()` returns `this.#keypair` — measured, the identical object *and* identical `privateKey` array on every call. Returning the stored object is memoization, trivially.

So two things I said need restating:

1. **This PR de-memoizes `serialize()`**, and that is the fix rather than a side effect. The cached object it was handing out *is* the signing secret.
2. **`toRaw()`'s copy on `main` was load-bearing, not redundant.** The comment there — "serialize() hands back the live internal buffer" — was exactly accurate. I removed that copy only because I had changed the property it depended on, in this same PR. Invalidating a guard's premise and then deleting the guard is a worse error than the one I described earlier, and 6cd9cf4ab restores it.

On the cost of de-memoizing: `serialize()` has no repeated callers. Every site is a one-off boundary crossing — `worker-controller.ts:98`, `runtime-client.ts:137-138`, `key-store.ts:29`, and `toPkcs8()`/`toRaw()` — none in a loop or per-request path. So it is 64 bytes at roughly six call sites, which seems the right price for not handing out the live private key.

— Claude

danfuzz:
Both done in dd087233f, and this reframing produced a better design than what I had.

**(a) Noble is now equivalently safe.** You were right that the two were not on the same footing. Native returns something a caller *cannot* corrupt — I checked every write path against a shared `CryptoKey`, and each either silently no-ops (`type`, `extractable`) or throws (`usages` push, `usages[0]` assignment, `algorithm.name`), because `usages` is frozen and sealed. Noble was safe only by arithmetic about who held which array.

Both now `Object.freeze()` the returned object, which closes the member-reassignment hole that copying alone leaves open. Worth recording why freezing is not a substitute for copying: `Object.freeze()` does **not** reach `ArrayBuffer` contents — on a frozen keypair, `pair.privateKey[0] = 0xff` succeeds and the signing secret changes in place. So freezing closes the cheap hole; fresh allocation closes the dangerous one. Noble needs both; native gets the second from the platform.

**(b) The contract is on the interface**, where it binds both implementations, and it states its own bound — that byte arrays inside an `InsecureCryptoKeyPair` cannot be frozen and are unshared instead.

With it written down, `ed25519/index.ts` relies on it rather than defending itself: `toRaw()` returns `serialized.privateKey` directly, and `toPkcs8()` needs nothing. That reliance is enforced, not assumed — ablating `serialize()`'s copy turns three tests red:

```
toRaw returns a COPY — mutating it cannot corrupt the identity ... FAILED
two toRaw() callers cannot interfere with each other ... FAILED
serialize() returns a COPY — mutating it cannot corrupt the signer ... FAILED
```

The tests pin the observable properties rather than where the copy happens, so they survived moving it.

The general lesson, which is really the root of this whole PR: the original defect was not the missing copy but the **missing contract**. `Signer.serialize()` had no documentation at all, so its behaviour had to be inferred from an implementation — and every error around it followed from that, including the original author's one-consumer workaround and both of mine.

Full workspace suite green (`All tests passing!`).

— Claude

PR #5544, opened 2026-08-09
perf: cede fresh byte arrays, and stop boxing blob uploads

danfuzz:
I (agent working on behalf of @danfuzz) started this as a narrow application of
the `transfer` parameter added by #5539 (`BL-007`). Review turned up two larger
things in the same area, so it now carries three related changes.

## 1. Cede where it earns its place

Ceding is worth it only where the array is **already fresh** and the alternative
is a real extra pass over the payload:

| site | source | what ceding saves |
| --- | --- | --- |
| `runner/src/builtins/fetch.ts` | `await response.arrayBuffer()` | `new Uint8Array(buf)` over that buffer is a **view, not a copy**, so ceding costs **zero** copies of an unbounded response body. Not ceding costs one full copy. |
| `runtime-client/backends/runtime-processor.ts` | the request's own payload | avoids a second pass over the blob |

`result-normalization.ts` moves the *other* way. An earlier version of this
description claimed it went "from two copies to one"; that was **wrong**.
`new FabricBytes(new Uint8Array(value), true)` and `new FabricBytes(value)` cost
exactly one copy either way — the constructor makes it when nothing is ceded —
so the first merely spent the same copy outside the constructor and added a
transferred buffer and a wrapper on top. It now reads `new FabricBytes(value)`.

## 2. A blob upload no longer boxes every byte

`UploadBlobRequest.body` was `number[]`. Nothing needed it to be: this protocol
never leaves one runtime instantiation, and its transport is structured cloning,
which carries a `Uint8Array` intact. The public API already took a `Uint8Array`
and converted *down* purely to cross the boundary.

Measured on a 1 MiB blob:

| step | `number[]` (before) | `Uint8Array` (after) |
| --- | --- | --- |
| sender `Array.from()` | 10.37 ms | — |
| transport `structuredClone` | 8.63 ms | 0.08 ms |
| receiver `Uint8Array.from()` | 0.34 ms | — |
| **total** | **19.34 ms** | **0.08 ms** |

`Uint8Array`, not `FabricBytes` — a class does not survive structured cloning,
which is exactly what broke `background-piece-service` and `cli` when tried on
`InsecureCryptoKeyPair` during #5542.

## 3. The ownership rule the cede rests on

The receiver cedes its request payload, which is only sound if nothing else
holds it. That was true because `postMessage` clones — an **implementation
detail of the one transport that exists**, not anything the interface required.
A future transport passing messages by reference would have satisfied
`RuntimeTransport` as written and silently broken the handler.

So it is now stated as a rule rather than relied on as a coincidence:

- **`BaseRequest`** — any value reaching a handler is owned outright by the
  receiver: not already shared, and not becoming shared later, except by the
  receiver's own action. A handler may retain, mutate, or cede without
  defending itself.
- **`RuntimeTransport.send()`** — the matching obligation on whatever delivers a
  request. Structured cloning satisfies it for free; a transport handing one
  object to both ends does not, and cannot be used as-is.

Stating it once at that level, rather than per payload, is what lets the next
binary field inherit it.

The test built its request with a `number[]`, so it did not model what a
transport delivers. It now allocates the array the handler is entitled to
consume — the code should not bend to an unrealistic test.

## Verification

`deno task check`, `deno fmt --check`, `deno lint`, `runner`, `runtime-client`,
and the full workspace suite, with current `main` merged in. `runtime-client`'s
test task runs `--no-check`, so `deno check` was also run directly against the
files it touches.

— Claude

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5544?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/runner/src/builtins/fetch.ts
+++ b/packages/runner/src/builtins/fetch.ts
@@ -153,9 +153,10 @@ async function processJsonResponse(
 async function processBinaryResponse(
   response: Response,
 ): Promise<FetchBinaryResult> {
-  // `arrayBuffer()` yields a buffer owned by nobody else, and the view over
-  // it is a temporary, so it is ceded rather than copied. A response body is
-  // unbounded, which is what makes the copy worth avoiding.
+  // The `true` below cedes the array to the `FabricBytes` rather than having
+  // it copied. `arrayBuffer()` yields a buffer nobody else holds, and the view
+  // over it is a temporary, so there is nothing left to protect. A response
+  // body is unbounded, which is what makes the copy worth avoiding.
   const bytes = new FabricBytes(
     new Uint8Array(await response.arrayBuffer()),
     true,
--- a/packages/runner/src/builtins/fetch.ts
+++ b/packages/runner/src/builtins/fetch.ts
@@ -153,7 +153,13 @@ async function processJsonResponse(
 async function processBinaryResponse(
   response: Response,
 ): Promise<FetchBinaryResult> {
-  const bytes = new FabricBytes(new Uint8Array(await response.arrayBuffer()));
+  // `arrayBuffer()` yields a buffer owned by nobody else, and the view over
+  // it is a temporary, so it is ceded rather than copied. A response body is
+  // unbounded, which is what makes the copy worth avoiding.
+  const bytes = new FabricBytes(
+    new Uint8Array(await response.arrayBuffer()),
+    true,
+  );
   const contentType = response.headers.get("content-type");
   const mediaType = contentType?.split(";")[0].trim().toLowerCase() ||
     "application/octet-stream";
--- a/packages/runner/src/sandbox/result-normalization.ts
+++ b/packages/runner/src/sandbox/result-normalization.ts
@@ -105,9 +105,10 @@ function normalizeSandboxNativeLeaf(value: unknown): unknown {
     brand === "[object Uint8Array]" && ArrayBuffer.isView(value) &&
     (value as Uint8Array).BYTES_PER_ELEMENT === 1
   ) {
-    // The inner copy is required -- `value` belongs to the sandbox -- but it
-    // is then unshared, so it is ceded rather than copied a second time.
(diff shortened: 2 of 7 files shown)


danfuzz:
Re-evaluated through the `FabricBytes` lens, per @danfuzz, with `main` merged in. One of the three sites was wrong and is now simplified.

I (agent working on behalf of @danfuzz) claimed in the description above that `result-normalization.ts` went "from two copies to one". **That was wrong.** `new FabricBytes(new Uint8Array(value), true)` and `new FabricBytes(value)` cost *exactly one copy either way* — the constructor makes it when not ceded. My version merely spent the same copy outside the constructor and added a transferred buffer and a wrapper object on top. It now reads `new FabricBytes(value as Uint8Array)`.

The other two sites stand, and for a reason worth stating precisely:

| site | why ceding earns its place |
| --- | --- |
| `fetch.ts` | `arrayBuffer()` yields a buffer nobody else holds, and `new Uint8Array(buf)` over it is a **view, not a copy** — so ceding costs **zero** copies of an unbounded response body, where not ceding costs one full copy |
| `runtime-processor.ts` | `Uint8Array.from()` allocates once; ceding avoids a second pass over the blob payload |

So the rule that fell out: ceding is worth it only where the array is *already* fresh and the alternative is a genuine extra pass. Where a copy has to happen regardless — a value owned by someone else — the constructor should just make it.

On `FabricBytes` more broadly: all three sites already construct one, and the surrounding types cannot usefully hold `FabricBytes` instead. `runtime-processor.ts` takes `request.body` across an IPC boundary, where `structuredClone` does not preserve a class.

— Claude

PR #5546, opened 2026-08-09
feat(memory,runner): own-write echo — deliver own patch heads on sync frames (CT-1965)

ubik2:
## Summary

Implements CT-1965: a sync frame includes a doc **unless the writing session provably holds it**. Own accepted `set`- and `delete`-produced heads stay elided exactly as today — the client supplied the bytes (or the absence), and the verdict plus marker promote them. Own accepted **`patch`-produced heads now ride the covering frame as full post-apply documents**: merged state (the Alice/Bob blind-patch case, CT-1872 Class 1a(b)) is truth the writer cannot extrapolate, so the frame is the truth channel.

No new wire surface and no negotiated capability: full-doc echoes already occur on the full-resync and watchAdd paths, and both client generations handle them — parking clients retire the overlay against the delivered base (`confirmPending`'s seq guard skips extrapolation), verdict-promoting clients receive an equal-seq swap that is identical bytes or the correction. Rollout control is a server-side `ownWriteEcho` config trio (default on), registered in `EXPERIMENTAL_OPTIONS.md`.

## Design keys

- **No reconstruction needed**: under #5529's per-space publication lock, an origin matching the head seq (`origin.sessionId === sessionId && origin.seq === entry.seq`) means the flush-time head **is** that commit's post-apply document. The CT-1926 seq-pinned engine slice stays unused.
- **Per-doc op classification**: `DirtyOrigin` gains `op`, stamped at the transact site as the **last** non-sqlite op per dirty key (a commit can set A and patch B; a multi-op commit ending in a set leaves a head the writer holds). Mixed provenance and the requeue merge are unchanged — they compare `sessionId` only.
- **Client extrapolation stays** as the set-promotion mechanism and the unwatched/old-server fallback.
- Deferred (recorded on CT-1965): exact-replay patch elision via `origin.baseSeq` + session-cache seq — implement if wire measurements show the O(doc) patch echo mattering.

## Runner: identity guard — removed after review

An earlier commit added a `valueEqual` identity guard on value-equal deliveries; review flagged it as unconditional content hashing on the hot path, contrary to `10-implementation-guidance.md` ("compare sync cache entries by revision identity, not payload serialization"). It is reverted — plain authoritative replacement stands, and the payoff cases (re-adds, retargets) mostly never re-deliver anyway since `sameSnapshot` dedupes by revision server-side.

## Tests

- **memory** (491 passed): the "does not echo" pin became the elision/echo split — own set → marker-only frame (original assertions intact), own patch → full post-apply doc on the writer's frame, observer unchanged. New: classification pins (patch-then-set elides, set-then-patch echoes, coalesced commits echo at latest seq, own delete elides) and a config-off pin of the pre-CT-1965 suppression.
- **runner** (1168 passed): new `own-write-echo-live.test.ts` over a live in-process server with held flushes — the pure patch echo applies **exactly once** and does not re-notify its writer (the double-apply risk pin); a writer converges to the merged list **including elements it never observed**. New server pin: a foreign-moved head is delivered to the earlier own-set writer. The mixed-provenance pin now runs with `ownWriteEcho` off, where the clearing rule is load-bearing.
- `deno task check` (memory, runner), `deno task check-docs`, `deno fmt --check`, `deno lint` — all green.

## Spec

`04-protocol.md` §4.11.2: the echo-suppression clause becomes the include-unless-provably-held rule; the parking bullet now states frame-retirement promotion for patch heads with extrapolation as fallback.

### DIFF
--- a/docs/development/EXPERIMENTAL_OPTIONS.md
+++ b/docs/development/EXPERIMENTAL_OPTIONS.md
@@ -44,6 +44,7 @@ was last checked against the code.
 | [`cfcLabelMetadataProtection`](#cfclabelmetadataprotection) | `RuntimeOptions.cfcLabelMetadataProtection` | `off` | Bernhard Seefeld (#4638) | `observe` (divergence counting) first, then `enforce` | implemented, staged rollout |
 | [`conflictAdmissionMode`](#conflictadmissionmode) | `CF_CONFLICT_ADMISSION` env, or `setConflictAdmissionMode()` | `off` | William Kelly (#4237); `hold` removed CT-1925 (#5110) | keep `preempt` as a tuning dial or remove after re-measurement | implemented, off by default, measured net-negative |
 | [`syncSchemaTableV2`](#syncschematablev2) | `setSyncSchemaTableConfig()` (negotiated per connection) | on | Ben Follington (#4292) | retire the negotiation once every peer speaks v2 | implemented, on by default |
+| [`ownWriteEcho`](#ownwriteecho) | `setOwnWriteEchoConfig()` (server-side only, not negotiated) | on | Robin McCollum (CT-1965) | remove the switch once the echo has field-soaked | implemented, on by default |
 | [`experimentalConcurrentWatchRefresh`](#experimentalconcurrentwatchrefresh) | `IRemoteStorageProviderSettings`; in the shell, the `commonfabric.concurrentWatchRefresh()` console command (localStorage, per browser profile) | off | Ben Follington (#4937; shell toggle #4974) | graduate to always-on after live measurement, or remove if superseded | implemented behind the flag, off by default, not yet measured over real latency |
 | [`cfcRenderCeiling`](#cfcrenderceiling) | `commonfabric.cfcRenderCeiling()` in the browser (localStorage) | off | Bernhard Seefeld (#4550) | graduate once exchange resolution lands | implemented, off by default, dogfood only |
 | [`fuseNfsCacheTuning`](#fusenfscachetuning) | `cf fuse mount --attrcache-timeout <whole seconds; 0 = untuned>` or `--noattrcache` | cf adds `attrcache-timeout=1` (one second) to FUSE-T mounts | Ian Hickson | keep the default; shrink the exec.ts listing-recheck delay once the default has field-soaked | implemented, on by default for FUSE-T, soak-validated |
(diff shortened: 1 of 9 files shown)


ubik2 (commented):

Re-review at d5312771be91908169d08ef33175408d07ebaf07. I reviewed all 9 changed files plus the affected engine, publication-lock, equality, and materialization paths. CI is green, git diff --check is clean, and focused local runs passed 57 memory tests plus 4 runner test steps.

🔴 Blocking — major, verified

packages/runner/src/storage/v2.ts:4070-4072 adds a valueEqual comparison to every non-deleted sync upsert. For distinct objects, valueEqual computes canonical SHA-256 content hashes by traversing both full documents. This conflicts directly with docs/specs/memory-v2/10-implementation-guidance.md:232-234, which says sync cache entries must be compared by revision identity rather than document payloads on hot paths.

This adds O(document size) work to every delivered upsert, including the new full-document patch echoes, even when there are no notification subscribers. The echo correction itself does not require cross-sequence object identity preservation.

Concrete fix: sever the runner identity commit and retain normal authoritative replacement, or introduce a cheap content identity that does not rehash the payload client-side. If cross-sequence Object.is stability is intentionally worth this cost, provide measurements and update the hot-path guidance explicitly before merge.

🟡 Improvement — minor, verified

packages/memory/test/v2-verdict-catchup.test.ts:332-356 claims to prove that mixed provenance clears suppression, but its write is a patch. With ownWriteEcho enabled, every own patch head is delivered regardless of provenance, so the test passes even if the mixed-provenance machinery is removed.

Run this case with ownWriteEcho disabled, resetting it afterward, so delivery genuinely depends on mixed provenance clearing. Alternatively, remove or reframe the test if that fallback invariant is intentionally retired.

The per-space lock, head-producing-op classification, coalescing behavior, and runner double-apply prevention otherwise look sound.

Verdict: changes requested because of the unconditional content hashing.

ubik2:
Both findings addressed in b21146805:

**Blocking (unconditional content hashing)**: agreed and taken via your primary suggestion — the identity guard is severed. `applySessionSync` restores plain authoritative replacement (revision-identity semantics per `10-implementation-guidance.md`), the canonical key-order pin in `pattern-object-member-storage` is back to its original assertion, and the live `Object.is` pin is removed. Beyond the hot-path cost, the payoff cases mostly never re-deliver — `sameSnapshot` already dedupes by revision server-side — so the guard bought little even where it fired. The cheap same-seq skip variant was considered and rejected: a client that promoted a wrong extrapolation at seq N (old-server fallback world) must not discard the same-seq correction a later frame delivers.

**Minor (vacuous mixed-provenance pin)**: fixed per your suggestion — the test now runs with `setOwnWriteEchoConfig(false)` (try/finally reset), so delivery genuinely depends on the clearing rule. Considered converting the own write to a `set` to pin the machinery in the default configuration instead, but rejected it: an own set clobbers the doc, so eliding a set-over-foreign head is value-correct and the delivery isn't load-bearing there — suppression mode is where the invariant lives.

Suites re-run green: memory (58 in the two touched files, 491 full), runner (targeted + `deno task check`).

PR #5547, opened 2026-08-09
feat(memory): manual subscription-flush mode for explicit fan-out gating (CT-1962)

ubik2:
## Summary

First of four CT-1962 PRs (timer-turn loopback delivery). `subscriptionRefreshDelayMs` widens to `number | "manual"`: manual mode never arms the refresh timer, so dirty spaces accumulate and fan out only through explicit `flushSessions()` calls, and a partial `flushSessions(spaces)` leaves the other dirty spaces held.

Why a sentinel and not a large delay: under the runner's fake-clock auto-advance, the pump fires the earliest armed prod timer regardless of its nominal delay — a 60s delay is not a hold once loopback delivery yields real timer turns. The gate has to be structural: no timer exists to fire. (`null` was rejected because `?? SUBSCRIPTION_REFRESH_DELAY_MS` would swallow it silently.)

## Tests

New `v2-manual-flush.test.ts` (BDD): manual mode holds fan-out across two 60s all-timer advances and delivers on explicit flush; a timed-mode control delivers under the identical drive (the negative's paired control); a partial flush keeps other spaces held with nothing re-armed.

Memory suite: 493 passed. `deno task check`, fmt, lint green.

## Follow-ups in this arc

PR 2: shared-harness extraction (32 copy-pasted `SharedServerStorageManager` sites). PR 3: staleness-family conversions to manual flush. PR 4: the loopback timer-turn flip itself.

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5547?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/memory/test/v2-manual-flush.test.ts
+++ b/packages/memory/test/v2-manual-flush.test.ts
@@ -1,0 +1,177 @@
+import { afterEach, beforeEach, describe, it } from "@std/testing/bdd";
+import { expect } from "@std/expect";
+import { FakeTime } from "@std/testing/time";
+import {
+  encodeMemoryBoundary,
+  getMemoryProtocolFlags,
+  MEMORY_PROTOCOL,
+  type ResponseMessage,
+  type ServerMessage,
+  type SessionEffectMessage,
+  type SessionOpenAuthMetadata,
+  type SessionSync,
+} from "../v2.ts";
+import { Server } from "../v2/server.ts";
+import { testSessionOpenServerOptions } from "./v2-auth-test-helpers.ts";
+
+// `subscriptionRefreshDelayMs: "manual"` is the fan-out gate for
+// controlled-staleness tests: the refresh timer is never armed, so dirty
+// spaces accumulate and fan out only through an explicit `flushSessions()`.
+// The FakeTime advances below are the structural negative — firing every
+// armed timer in the system delivers nothing, because there is no timer to
+// fire — paired with a timed-mode control that delivers through the same
+// advance.
+
+const HELLO = {
+  type: "hello",
+  protocol: MEMORY_PROTOCOL,
+  flags: getMemoryProtocolFlags(),
+} as const;
+
+const shiftMessage = (messages: ServerMessage[]): ServerMessage => {
+  const message = messages.shift();
+  expect(message).toBeDefined();
+  return message!;
+};
+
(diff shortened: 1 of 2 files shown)


ubik2 (commented):
_cf-review via Codex GPT-5, on behalf of @ubik2_

🔴 **Blocking — `packages/memory/v2/server.ts:1417` and `:3624`**

Manual mode makes `Server.idle()` return while dirty subscription fan-out remains pending. `scheduleRefresh()` deliberately creates no timer, but `idle()` only calls `flushSessions()` when `#refreshTimer` or `#refreshing` is non-null. This contradicts `idle()`'s documented guarantee to return with no pending work.

**Verified locally:** after a watched `writeDocument()` in manual mode, `await server.idle()` returns without delivering the upsert.

I think the clean semantic is that `"manual"` means no timer-driven fan-out. `flushSessions()` and `idle()` are both explicit synchronization points, so `idle()` should count as a manual flush rather than wait for another caller. Please make `idle()` notice `#dirtySpaces.size > 0`, add a regression test showing that it drains manual-mode work, and update the option JSDoc to say that explicit `flushSessions()` or `idle()` drains held fan-out.

**Verdict:** request changes until manual mode preserves `idle()`'s quiescence contract.

ubik2:
Fixed in 822439c9a, adopting your proposed semantic exactly: `"manual"` gates the TIMER, not the explicit synchronization points — `idle()` now notices dirty spaces with no timer armed and flushes, same as an explicit `flushSessions()`. Verified before adopting that no current `Server.idle()` caller uses a manual-mode server and no manual-mode staleness test calls `server.idle()`, so the quiescence contract stays uniform with no premise destroyed. Added the regression pin (manual server + watched `writeDocument` + `tickAsync` delivers nothing, then `idle()` alone delivers) and updated the option jsdoc to name both explicit drains; the `waiting-in-tests.md` gating section on the flip branch (#5550, ea810a335) says the same. Memory suite 493+1 green; the fix is merged up the stack (#5548/#5549/#5550).

PR #5548, opened 2026-08-09
refactor(runner,piece,cli): extract the shared loopback-server harness (CT-1962)

ubik2:
## Summary

Second of four CT-1962 PRs (stacked on #5547). The 32 byte-identical copy-pasted `SharedServerStorageManager` subclasses (29 runner, 2 piece, 1 cli) collapse into one static: `EmulatedStorageManager.connectTo(server, options)`, plus `newLoopbackServer(options?)` (re-exported via `cache.deno`) and an audience-pinned `newSharedServer(options?)` in the runner test utils. Both accept `subscriptionRefreshDelayMs`, including `"manual"` from #5547.

Two behavior-relevant details, both improvements:
- `connectTo` sets caller-owned server lifecycle (`#ownsServer = false`) — the copies achieved this implicitly by overriding `server()` so the memoized close path never saw the shared server.
- `own-write-echo-live` converts from a 60s held delay to `"manual"`: under the runner's fake-clock auto-advance, a large delay is not a hold — the pump fires any armed timer regardless of nominal delay once macrotask turns exist. The archetypal controlled-staleness fixture now gates fan-out structurally.

Deviations preserved as-is: fixed-principal servers stay local (`action-result-fabric-values`, `scheduler-cold-replica`); `effect-conflict-recovery` keeps its `subscribe()` override as a minimal subclass; `list-builtin-edge-paths`' separate plain-transport harness untouched; `sqlite-handle-multi-runtime`'s audience literal unifies to the shared test audience (inert — the client echoes whatever the server advertises).

## Verification

- runner: full suite 1168 passed / 0 failed (with preload); `deno task check` green
- piece: both converted files, 4 passed (135 steps)
- cli: piece.test.ts, 83 steps green
- fmt/lint clean across all 35 touched files

<a href="https://cubic.dev/pr/commontoolsinc/labs/pull/5548?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/cli/test/piece.test.ts
+++ b/packages/cli/test/piece.test.ts
@@ -11,9 +11,9 @@ import {
 } from "@commonfabric/runner";
 import {
   EmulatedStorageManager,
+  newLoopbackServer,
   StorageManager,
 } from "@commonfabric/runner/storage/cache.deno";
-import * as MemoryV2Server from "@commonfabric/memory/v2/server";
 import { FabricError } from "@commonfabric/data-model/fabric-instances";
 import { FabricSpecialObject } from "@commonfabric/data-model/fabric-value";
 import { FabricHash } from "@commonfabric/data-model/fabric-primitives";
@@ -2727,32 +2727,7 @@ describe("cli piece parsing", () => {
       "cf piece search cold runtime test",
     );
     const space = signer.did();
-    const audience = "did:key:z6Mk-runner-emulated-memory";
-    const server = new MemoryV2Server.Server({
-      authorizeSessionOpen(message) {
-        const principal = (message.authorization as { principal?: unknown })
-          ?.principal;
-        return typeof principal === "string" ? principal : undefined;
-      },
-      sessionOpenAuth: { audience },
-    });
-
-    class SharedStorage extends EmulatedStorageManager {
-      static connect(server: MemoryV2Server.Server): SharedStorage {
-        const manager = new SharedStorage(
-          { as: signer, memoryHost: new URL("memory://") } as any,
-          () => server,
-        );
-        manager.#server = server;
-        return manager;
-      }
-
-      #server!: MemoryV2Server.Server;
-
-      protected override server(): MemoryV2Server.Server {
-        return this.#server;
-      }
-    }
+    const server = newLoopbackServer();
 
     const leafSchema = {
       type: "object",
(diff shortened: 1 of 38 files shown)


ubik2 (commented):

No blocking findings. Three non-blocking improvements:

1. **Assert the manual fan-out premise** — `packages/runner/test/own-write-echo-live.test.ts:138-148`

   The test says session 2 remains stale until the explicit flush, but never checks it. The unchanged test still passed when `"manual"` was experimentally replaced with a numeric timer, so broken option forwarding would escape. Assert that `cell2` is still `["seed"]` before B writes; that assertion passed in manual mode and failed with the timer.

2. **Pin server ownership directly** — `packages/runner/src/storage/v2-emulate.ts:124-168`

   `connectTo()` correctly leaves the shared server open while `emulate()` closes its owned server, but the migrated cleanup paths do not directly assert which object performed the close. Add a focused ownership test. Instrumentation verified the current behavior as `{ sharedCloses: 0, ownedCloses: 1 }`.

3. **Mention `idle()` in the manual-mode JSDoc** — `packages/runner/src/storage/v2-emulate.ts:8-13` and `packages/runner/test/memory-v2-test-utils.ts:52-57`

   Both comments say manual fan-out occurs entirely through `flushSessions()`, but `Server.idle()` also drains held fan-out. Describe manual mode as disabling timer-driven fan-out, with either explicit synchronization point draining it.

Verdict: safe to merge; no behavioral migration regressions found. These are minor test-rigor and documentation-coherence gaps.

ubik2:
All three suggestions applied in a262b36d8: own-write-echo-live now asserts the staleness premise directly right after the concurrent append (your validation that it fails under any timed cadence made this an easy yes — it pins the option forwarding, not just the merge outcome); a new emulated-storage-ownership.test.ts pins the lifecycle split (connectTo borrows — manager close leaves the shared server open, its own close still works; emulate() owns — manager close closes it); and both runner-side manual-mode JSDocs now describe manual as disabling TIMER-driven fan-out with either explicit synchronization point (flushSessions() or idle()) draining it, matching the #5547 idle() fix.

